PR #394, opened 2014-12-11
[WFCORE-309] : system properties are trim()'d and loose whitespace.

ehsavoie:
When the ModelType is STRING or PROPERTY the attribute value is set without trimming.

Jira: https://issues.jboss.org/browse/WFCORE-309

### DIFF
--- a/controller/src/main/java/org/jboss/as/controller/AttributeParser.java
+++ b/controller/src/main/java/org/jboss/as/controller/AttributeParser.java
@@ -24,13 +24,13 @@
 
 package org.jboss.as.controller;
 
+
 import javax.xml.stream.XMLStreamException;
 import javax.xml.stream.XMLStreamReader;
 
 import org.jboss.as.controller.operations.validation.ParameterValidator;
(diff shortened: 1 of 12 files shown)


controller/src/main/java/org/jboss/as/controller/AttributeParser.java
```diff
@@ -89,10 +92,14 @@ private ModelNode parse(final AttributeDefinition attribute, final String value)
         final String trimmed = value == null ? null : value.trim();
         ModelNode node;
         if (trimmed != null) {
-            if (attribute.isAllowExpression()) {
+            if (attribute.isAllowExpression() && ParseUtils.isExpression(trimmed)) {
                 node = ParseUtils.parsePossibleExpression(trimmed);
```
bstansberry: This (and the others) can just be

```
node = new ModelNode(new ValueExpression(value));
```

That's all that parsePossibleExpression does, except for an isExpression() call that's now redundant.

bstansberry:
Agreed; a utility in ParseUtils would be better.

PR #400, opened 2014-12-15
[WFCORE-371/WFCORE-419] : Expose mgmt client socket binding config through ModelControllerClient Factory.

ehsavoie:
Now you can specify the bind address for the ModelControllerClient.
You may set it using  client socket binding config through CLI connect with the --bind= param.

Jira: https://issues.jboss.org/browse/WFCORE-371 and https://issues.jboss.org/browse/WFCORE-419

### DIFF
--- a/cli/src/main/java/org/jboss/as/cli/impl/CliLauncher.java
+++ b/cli/src/main/java/org/jboss/as/cli/impl/CliLauncher.java
@@ -63,6 +63,7 @@ public class CliLauncher {
             char[] password = null;
             boolean noLocalAuth = false;
             int connectionTimeout = -1;
+            String clientBindAddress = null;
 
             for(String arg : args) {
                 if(arg.startsWith("--controller=") || arg.startsWith("controller=")) {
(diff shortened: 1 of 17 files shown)


bstansberry:
Looks good. @aloubyansky, please approve or comment. Thanks!

cli/src/main/java/org/jboss/as/cli/impl/CliLauncher.java
```diff
@@ -174,6 +175,12 @@ public static void main(String[] args) throws Exception {
                         argError = "'=' is missing after --timeout";
                         break;
                     }
+                } else if(arg.startsWith("--bind=") || arg.startsWith("bind=")) {
```
aloubyansky: Let's not support "bind=". That will be consistent with all the other supported arguments except for the controller.

aloubyansky:
That's all I had. Otherwise, nicely done. Thanks!

cli/src/main/java/org/jboss/as/cli/impl/CommandContextConfiguration.java
```diff
@@ -0,0 +1,100 @@
...
+        this.controller = controller;
+        this.username = username;
+        this.password = password;
+        this.clientBindAddress = clientBindAddress;
+        this.consoleInput = consoleInput;
+        this.consoleOutput = consoleOutput;
+        this.initConsole = initConsole;
+        this.disableLocalAuth = disableLocalAuth || username != null;
+        this.connectionTimeout = connectionTimeout;
+    }
```
ehsavoie: Done

PR #402, opened 2014-12-16
Few JDK9 related & general build fixes

ctomc:
Changes apply to older JDKs as well.

### DIFF
--- a/host-controller/src/main/java/org/jboss/as/host/controller/model/jvm/JvmOptionsBuilderFactory.java
+++ b/host-controller/src/main/java/org/jboss/as/host/controller/model/jvm/JvmOptionsBuilderFactory.java
@@ -27,6 +27,8 @@ import java.util.Collections;
 import java.util.HashMap;
 import java.util.List;
 import java.util.Map;
+import java.util.regex.Matcher;
+import java.util.regex.Pattern;
 
 import org.jboss.as.controller.parsing.Attribute;
 import org.jboss.as.controller.parsing.Element;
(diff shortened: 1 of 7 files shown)


ctomc:
Rebased & changed to use Integer instead of double same as we do in PlatformMBeanUtil.

host-controller/src/main/java/org/jboss/as/host/controller/model/jvm/JvmOptionsBuilderFactory.java
```diff
@@ -167,6 +168,19 @@ boolean checkOption(boolean condition, String jvm, String option, String schemaE
     }
 
     private static class SunJvmOptionsBuilder extends JvmOptionsBuilder {
+        public static final int JVM_MAJOR_VERSION;
+
+        static {
+            int vmVersion;
+            try {
+                String vmVersionStr = WildFlySecurityManager.getPropertyPrivileged("java.specification.version", null);
+                vmVersion = Integer.valueOf(vmVersionStr.substring(2));
```
dmlloyd: You are making an assumption here.  You should validate the assumption (using a regex is probably simplest) before extracting the minor version number.  If the assumption cannot be validated, then you need to have some sane recourse (even if it's just throwing an exception), because otherwise this will become one of those impossible-to-figure-out bugs when Oracle suddenly decides that it's time for Java 2.0.

ctomc: Yes it is an assumption that might break in a future.
I just copied existing code :) will add some more safeguards.

jamezp:
RE: b3c269d6913c4cfd8c95ee007b56ea10043b0877. I vote we just remove the `@ServerControl` annotation and use a system property to control the `automaticServiceControl` boolean.

ctomc:
@jamezp that would prevent us to have mixed type of tests as part of same test execution.

dmlloyd:
Seems OK

jamezp:
We don't have mixed tests and I don't see it in our future, but if I'm the only one voting that way then... :)

jamezp:
I'm going to back peddle here a bit. We should remove that last commit with the `@ServerControl` annotation change. Then a separate PR to remove it. I'm not sold we need it at all. It's only used in manual mode tests and even they don't use the default. It seems to be more trouble than it's worth.

ctomc:
@jamezp done, removed last commit, will send new PR just about testrunner once this is merged

PR #421, opened 2015-01-09
Undertow 1.2.0.Beta7

stuartwdouglas:


stuartwdouglas:
Seems to be some kind of an issue with the latest release

PR #422, opened 2015-01-09
Copywrite Header to SameTypeOperation

Gonthim:


cli/src/main/java/org/jboss/as/cli/handlers/ifelse/SameTypeOperation.java
```diff
@@ -1,3 +1,24 @@
+/*
+ * JBoss, Home of Professional Open Source.
+ * Copyright 2011, Red Hat, Inc., and individual contributors
```
kabir: Back to the future? :-)

Gonthim:
Well if I'm going to update one of them... Guess I'll update everything in the CLI!

bstansberry:
Please don't make any further such bulk changes.

PR #423, opened 2015-01-09
[WFCORE-302] : Enable CORS

ehsavoie:
List of allowed-origins is parametrable and is filtering requests.
Security headers are set to enable authentication for all CORS requests.

Jira: https://issues.jboss.org/browse/WFCORE-302

### DIFF
--- a/server/src/main/resources/schema/wildfly-config_3_0.xsd
+++ b/server/src/main/resources/schema/wildfly-config_3_0.xsd
@@ -2127,6 +2127,15 @@
                 </xs:choice>
                 <xs:attribute name="console-enabled" type="xs:boolean" use="optional" default="true"/>
                 <xs:attribute name="http-upgrade-enabled" type="xs:boolean" use="optional" default="false"/>
+                <xs:attribute name="allowed-origins" type="urlListType" use="optional" >
+                    <xs:annotation>
+                        <xs:documentation>
+                            A space separated list of Origins that will be trusted to send request to the management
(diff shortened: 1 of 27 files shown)


kabir:
This could do with a core model test to 'prove' the parsing of the new element works

ehsavoie:
Test added

bstansberry:
@darranl this needs your approval

server/src/main/resources/schema/wildfly-config_3_0.xsd
```diff
@@ -2127,6 +2127,15 @@
                 </xs:choice>
                 <xs:attribute name="console-enabled" type="xs:boolean" use="optional" default="true"/>
                 <xs:attribute name="http-upgrade-enabled" type="xs:boolean" use="optional" default="false"/>
+                <xs:attribute name="allowed-origins" type="xs:string" use="optional" >
```
darranl: I would prefer it if one of the list types was used for the attribute type which would make it a space separated list.  Any form of comma separated list requires reading of the documentation to understand how to format the content meaning tools can not do this automatically.

We already have a urlListType and stringListType in the schema that should be suitable for this.

darranl:
Happy with the changes in principal, however as commented would prefer one of the list types is used instead of relying on documentation to let end users know how to format the attribute value.

The list types are already used elsewhere where users are configuring security for the management interfaces.

ehsavoie:
Coma separated list string replaced by a space separated list.

PR #426, opened 2015-01-12
WFLY-3723: Ensure ROOT Locale (en) is loaded whenever user request for en_* locale

rpelisse:
Also,
    - add extra test to ensure provided locale tag string is neither too small or too big
    - code cleanup to allow unit test
    - create a protected static method implementing the desired Locale resolution, but separated from Wildfly data model
    - methods throw exception to be catched by wfly specific code, which handle proper logging/display of format error
    - this ultimately allowed to add test case to ensure LocaleResolution behave as desired
    - cleanup, extract Locale, low-level parsing to separate utility class


rpelisse:
Matching PR for the Wildfly 8.x previous [PR7086](https://github.com/wildfly/wildfly/pull/7086) fixing the issue on 8.x branch.

rpelisse:
Appears unrelated - can you retest please ?

Thanks !

rpelisse:
Grr, just realized there is a missing header on the test file. Close, will open a new PR

PR #427, opened 2015-01-12
[WFCORE-497] Do cross-model validation in a separate step

bstansberry:
While at it I made a couple minor changes to how data is obtained during validation.


ctomc:
Just thinking out loud, couldn't we do validation like this also in ValidateModelStepHandler? which is already ran as separate step.

kabir:
The concern here seems to be very different wrt to the ValidateModelStepHandler question.

bstansberry:
@ctomc Eventually much of this stuff should be done in something like ValidateModelStepHandler, or perhaps internally in the OC the way that capabilities and requirements are resolved. I'm talking here about the "validating model ref" part. These particular OSHs also do some pretty-custom validation that I don't envision us handling centrally -- the HandlerUtil.checkNoOtherHandlerWithTheSameName bit.

I did this as a one-off because it is blocking Alexey's work.

ctomc:
makes sense, tnx @bstansberry

PR #428, opened 2015-01-12
[WFCORE-263] Detect and react to domain rollout issues in invocations of cancel|find-non-progressing-operation on a slave HC

bstansberry:
Also fixes WFCORE-489

The convenience methods find-non-progressing-operation and cancel-non-progressing operation look for ops that are holding the exclusive lock for an excessive amount of time. When invoked on a slave HC, if the real problem preventing completion is the rollout of the op to one of the slave's servers, these ops will instead see the request rolling out the op from the master to the HC itself as the problematic one, as that will be the one holding the exclusive lock (taken to update the slave HC's copy of the model). But that request is just waiting for a commit from the master once the server rollout is done, and isn't actually misbehaving.

The second commit in this PR checks for this situation, and if found checks if the server ops associated with the same domain-uuid are taking too long to complete. If so:

1) if there is only one such server op, that is the one found/cancelled. If the slave HC the user targeted is the only one with a server not completing, this should free up the entire system. 

If there is more than one slave HC with this condition, it won't free up the entire system, but my assumption here is if the user is targeting this op to a particular slave and not to the master, they likely have done some sort of analysis to lead them to do that.

2) if there is more than one such server op, there is no single op that can be identified/cancelled and the request fails with a message instructing the user to target the request against the master HC. Only the master can trigger cancellation across multiple processes.


bstansberry:
retest this please

bstansberry:
retest this please

baranowb:
retest this please

bstansberry:
retest this please

baranowb:
retest this please

bstansberry:
I can't see how that last "org.jboss.as.jmx.rbac.JmxRbacEnabledTestCase.testAdministratorSensitiveMBeans" failure could have anything to do with this PR. It's a failure to register or remove a JMX NotificationListener for mbean test:name=bean which is not related at all to the DMR management layer stuff this PR changes.

PR #430, opened 2015-01-13
https://issues.jboss.org/browse/WFCORE-501 ServerGroupAddHandler should ...

aloubyansky:
...add a separate step to validate profile and socket-binding-group refs

Thanks,
Alexey


aloubyansky:
Needs work.

PR #438, opened 2015-01-16
[WFCORE-507] :  Unit test FileSystemDeploymentServiceUnitTestCase fails on IBM JDK.

ehsavoie:
As ByteMan can't attach itslef to a running process which is what is done when using @BMRule then the test fails.
So the test simply returns when an IBM jvm is used.

Jira: https://issues.jboss.org/browse/WFCORE-507

### DIFF
--- a/deployment-scanner/pom.xml
+++ b/deployment-scanner/pom.xml
@@ -37,6 +37,33 @@
 
     <name>WildFly: Deployment Scanner</name>
 
+    <build>
+        <plugins>
+            <plugin>
+                <groupId>org.apache.maven.plugins</groupId>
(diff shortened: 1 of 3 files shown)


bstansberry:
http://lightning.mw.lab.eng.bos.redhat.com/viewLog.html?buildId=8101&tab=buildResultsDiv&buildTypeId=EAP_6xLinuxIbmJ is a custom run of the EAP IBM JDK job against the EAP version of this branch, to try and see if this resolves the problem.

ehsavoie:
The failing tests don"t seem to be related

ctomc:
yes failed tests fail also on 6.x branch without this patch.

PR #440, opened 2015-01-18
WFCORE-509 Add listener to restart the whole deployment on service restart

stuartwdouglas:
**Do not merge without ok from Jason**

This deals with the problem of deployment service restartability. As many deployment
unit services are not 100% restartable this change will force the whole deployment to restart
if a restart is detected.


server/src/main/java/org/jboss/as/server/deployment/DeploymentRestartListener.java
```diff
@@ -0,0 +1,66 @@
+package org.jboss.as.server.deployment;
```
bstansberry: Copyright.

PR #441, opened 2015-01-19
[WFCORE-495] don't write to config file if attribute PERSISTENT is false

soul2zimate:
https://issues.jboss.org/browse/WFCORE-495

The PERSISTENT checking was removed in https://github.com/wildfly/wildfly-core/commit/673ada6dbfbf4341acff29f80772d053612343cc which fixed incorrect isRuntime() return value.

### DIFF
--- a/server/src/main/java/org/jboss/as/server/deployment/DeploymentFullReplaceHandler.java
+++ b/server/src/main/java/org/jboss/as/server/deployment/DeploymentFullReplaceHandler.java
@@ -44,6 +44,7 @@ import org.jboss.as.controller.OperationFailedException;
 import org.jboss.as.controller.OperationStepHandler;
 import org.jboss.as.controller.PathAddress;
 import org.jboss.as.controller.PathElement;
+import org.jboss.as.controller.registry.Resource;
 import org.jboss.as.protocol.StreamUtils;
 import org.jboss.as.repository.ContentReference;
 import org.jboss.as.repository.ContentRepository;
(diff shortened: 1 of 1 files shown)


bstansberry:
I want us to find a different solution to this. I don't want persistence controlled by an arbitrary attribute of the model, which can only be understood by custom code. It needs to be exposed via the Resource or ImmutableManagementResourceRegistration API. That is what the WFCORE-140 change to relying on Resource.isRuntime() was about.

soul2zimate:
Rebase to work with Resource.isRuntime() value. runtimeOnly is *final* inside AbstractModelResource.
Unlike DeploymentAddHandler, DeploymentFullReplaceHandler needs to remove and add new resource with updated runtimeOnly value for the replacement.

server/src/main/java/org/jboss/as/server/deployment/DeploymentFullReplaceHandler.java
```diff
@@ -139,6 +140,13 @@ public void execute(OperationContext context, ModelNode operation) throws Operat
...
             operation.get(PERSISTENT.getName()).set(true);
         }
+
+        boolean persistent = PERSISTENT.resolveModelAttribute(context, operation).asBoolean();
+        context.removeResource(PathAddress.pathAddress(deploymentPath));
+        final Resource resource = Resource.Factory.create(!persistent);
+        resource.writeModel(deploymentModel);
+        context.addResource(PathAddress.pathAddress(deploymentPath), resource);
+
         PERSISTENT.validateAndSet(operation, deploymentModel);
```
bstansberry: Thanks; I like this approach much better.

Looking at this, I see something that needs to change, and a recommended improvement.

1) Needs to change: the resource.writeModel() call at L147 needs to move below the "if (operation.hasDefined(ENABLED.getName())) {" block further down, before the "// Do the runtime part if the deployment is enabled" section.

Otherwise the PERSISTENT.validateAndSet and ENABLED.validateAndSet calls will have no impact, as once writeModel is called the deploymentModel var is referencing a different ModelNode from what's inside the resource.

Hmm, actually these calls need to happen as well before the resource.writeModel() call:

DeploymentUtils.enableAttribute(deploymentModel); (L154 in master)
DeploymentUtils.disableAttribute(deploymentModel); (L157 in master)

but the other stuff in that if/else block must happen *after*. So a bit of reworking is needed.

I say L147 needs to move down, but it's clearer to move down your entire block, except...

2) Recommended improvement: L102 is reading the resource, and then your L145 is removing it. L102 can just remove it directly and save a second lookup. A call to context.removeResource() returns a ref to the Resource, so you can still get the deploymentModel.

A comment as to why the remove is happening would be good, as this is all about a corner case.

soul2zimate:
thanks for review.

PR #442, opened 2015-01-19
JBIDE-19029 - Error: 'Attribute name is not allowed to appear in element...

robstryker:
... sub-deployment' in two Seam examples


ctomc:
this is ok to test

PR #444, opened 2015-01-19
Use a static-discovery element for our default slave HC config. It segue...

bstansberry:
...s better into listing multiple options


kabir:
Related failure, xsd needs tweaking to allow for expressions for the protocol

bstansberry:
Once https://github.com/wildfly/wildfly/pull/7117 is merged the test failure should go away. The xsd is ok; there are lots of places where an expression violates the xsd.

jamezp:
retest this please

PR #445, opened 2015-01-20
[WFCORE-297] : HC should remember the 'run' state of the server instances after crash or shutdown

ehsavoie:
Since starting or stopping a server is a read-only operation this can't be done as proposed.
When the update-auto-start-with-server-status attribute is set to true on the server declaration then when the server is started a file servername.started will be created in ${jboss.datahome}/auto-start.
When the server is stopped the servername.started will be deleted and a servername.stopped will be created.
Those files are used at startup to determine the auto-start value : if the  update-auto-start-with-server-status is set then the exisitng file will be used instead of the value of the auto-start attribute in the XML.

Jira: https://issues.jboss.org/browse/WFCORE-297

### DIFF
--- a/host-controller/src/main/java/org/jboss/as/host/controller/operations/ServerStartHandler.java
+++ b/host-controller/src/main/java/org/jboss/as/host/controller/operations/ServerStartHandler.java
@@ -96,7 +96,7 @@ public class ServerStartHandler implements OperationStepHandler {
         final PathAddress address = PathAddress.pathAddress(operation.require(OP_ADDR));
         final PathElement element = address.getLastElement();
         final String serverName = element.getValue();
-        final boolean blocking = operation.get("blocking").asBoolean(false);
+        final boolean blocking = operation.get(ModelDescriptionConstants.BLOCKING).asBoolean(false);
 
         final ModelNode model = Resource.Tools.readModel(context.readResourceFromRoot(PathAddress.EMPTY_ADDRESS, true));
         context.addStep(new OperationStepHandler() {
@@ -108,20 +108,29 @@ public class ServerStartHandler implements OperationStepHandler {
                 final ServerStatus origStatus = serverInventory.determineServerStatus(serverName);
                 if (origStatus != ServerStatus.STARTED && origStatus != ServerStatus.STARTING) {
                     final ServerStatus status = serverInventory.startServer(serverName, model, blocking);
+                    persistAutoStart(context);
                     context.getResult().set(status.toString());
                 } else {
                     context.getResult().set(origStatus.toString());
                 }
-
                 context.completeStep(new OperationContext.RollbackHandler() {
                     @Override
                     public void handleRollback(OperationContext context, ModelNode operation) {
                         if (origStatus != ServerStatus.STARTED && origStatus != ServerStatus.STARTING) {
                             serverInventory.stopServer(serverName, -1);
+                            persistAutoStart(context);
                         }
                     }
                 });
             }
+            /**
+             * By reading the model we ensure that ServerConfigResource.persistAutoStart will get called.
+             * @param context the current operation context.
+             */
+            private void persistAutoStart(OperationContext context) {
+                context.readResource(PathAddress.EMPTY_ADDRESS, false).getModel();
+            }
         }, OperationContext.Stage.RUNTIME);
     }
+
 }
--- a/host-controller/src/main/java/org/jboss/as/host/controller/resources/ServerConfigResource.java
+++ b/host-controller/src/main/java/org/jboss/as/host/controller/resources/ServerConfigResource.java
@@ -1,0 +1,129 @@
+/*
+ * Copyright (C) 2015 Red Hat, inc., and individual contributors
+ * as indicated by the @author tags. See the copyright.txt file in the
+ * distribution for a full listing of individual contributors.
+ *
+ * This library is free software; you can redistribute it and/or
+ * modify it under the terms of the GNU Lesser General Public
+ * License as published by the Free Software Foundation; either
+ * version 2.1 of the License, or (at your option) any later version.
+ *
+ * This library is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the GNU
+ * Lesser General Public License for more details.
+ *
+ * You should have received a copy of the GNU Lesser General Public
+ * License along with this library; if not, write to the Free Software
+ * Foundation, Inc., 51 Franklin Street, Fifth Floor, Boston,
+ * MA 02110-1301  USA
+ */
+package org.jboss.as.host.controller.resources;
+
+import static org.jboss.as.controller.descriptions.ModelDescriptionConstants.AUTO_START;
+import static org.jboss.as.controller.descriptions.ModelDescriptionConstants.UPDATE_AUTO_START_WITH_SERVER_STATUS;
+
+import java.io.File;
+import java.io.IOException;
+import java.nio.file.Files;
+import java.nio.file.Path;
+import org.jboss.as.controller.ControlledProcessState;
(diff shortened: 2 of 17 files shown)


ehsavoie:
retest this please

host-controller/src/main/java/org/jboss/as/host/controller/resources/ServerConfigResource.java
```diff
@@ -0,0 +1,132 @@
...
+
+    private static final String STOPPED_EXT = ".stopped";
+    private static final String STARTED_EXT = ".started";
+
+    public ServerConfigResource(ServerInventory serverInventory, ControlledProcessState processState, String serverName, File domainDataDir, Resource delegate) {
+        super(delegate);
+        this.serverInventory = serverInventory;
+        this.serverName = serverName;
+        this.processState = processState;
+        this.autoStartDataDir = Paths.get(domainDataDir.toURI()).resolve("auto-start");
```
jamezp: Why not just use `domainDataDir.toPath()`?

ehsavoie: Good question ;)

host-controller/src/main/java/org/jboss/as/host/controller/resources/ServerConfigResource.java
```diff
@@ -0,0 +1,132 @@
...
+        if (serverInventory != null && shouldUpdateAutoStart(model)) {
+            ServerStatus status = serverInventory.determineServerStatus(serverName);
+            try {
+                if (status == ServerStatus.STARTED || status == ServerStatus.STARTING) {
+                    if (Files.notExists(startedFile)) {
+                        Files.createFile(startedFile);
+                    }
+                    Files.deleteIfExists(stoppedFile);
+                    model.get(AUTO_START).set(true);
+                    HostControllerLogger.ROOT_LOGGER.warn("We are updating the auto-start attribute to TRUE for " + serverName);
```
jamezp: Needs to be internationalized.

host-controller/src/main/java/org/jboss/as/host/controller/resources/ServerConfigResource.java
```diff
@@ -0,0 +1,132 @@
...
+                } else if (status == ServerStatus.STOPPED || status == ServerStatus.STOPPING) {
+                    if (Files.notExists(stoppedFile)) {
+                        Files.createFile(stoppedFile);
+                    }
+                    Files.deleteIfExists(startedFile);
+                    model.get(AUTO_START).set(false);
+                    HostControllerLogger.ROOT_LOGGER.warn("We are updating the auto-start attribute to FALSE for " + serverName);
```
jamezp: Needs to be internationalized.

host-controller/src/main/java/org/jboss/as/host/controller/resources/ServerConfigResource.java
```diff
@@ -0,0 +1,132 @@
...
+                }
+            } catch (IOException ex) {
+                HostControllerLogger.ROOT_LOGGER.error("Error persisting server autostart status", ex);
```
jamezp: Needs to be internationalized.

testsuite/domain/src/test/java/org/jboss/as/test/integration/domain/management/ServerAutoStartTestCase.java
```diff
@@ -0,0 +1,281 @@
...
+    private static final ModelNode mainOne = new ModelNode();
+    private static final ModelNode mainTwo = new ModelNode();
+    private static final ModelNode mainThree = new ModelNode();
+    private static final ModelNode mainFour = new ModelNode();
+    private static final ModelNode otherOne = new ModelNode();
+    private static final ModelNode otherTwo = new ModelNode();
+    private static final ModelNode otherThree = new ModelNode();
+    private static final ModelNode otherFour = new ModelNode();
+    private static final Path autoStartMasterDataDir
+            = Paths.get(DomainTestSupport.getHostDir(ServerAutoStartTestCase.class.getSimpleName(), "master").toURI())
```
jamezp: `File.toPath()` would be a better approach than creating a new `Path` based on the URI.

testsuite/domain/src/test/java/org/jboss/as/test/integration/domain/management/ServerAutoStartTestCase.java
```diff
@@ -0,0 +1,281 @@
...
+            .resolve("data")
+            .resolve("auto-start");
+    private static final Path autoStartSlaveDataDir
+            = Paths.get(DomainTestSupport.getHostDir(ServerAutoStartTestCase.class.getSimpleName(), "slave").toURI())
```
jamezp: `File.toPath()` would be a better approach than creating a new `Path` based on the URI.

host-controller/src/main/java/org/jboss/as/host/controller/resources/ServerConfigResource.java
ehsavoie: removed

host-controller/src/main/java/org/jboss/as/host/controller/resources/ServerConfigResource.java
ehsavoie: removed

host-controller/src/main/java/org/jboss/as/host/controller/resources/ServerConfigResource.java
ehsavoie: Fixed

host-controller/src/main/java/org/jboss/as/host/controller/resources/ServerConfigResource.java
```diff
@@ -0,0 +1,129 @@
...
+        this.serverInventory = serverInventory;
+        this.serverName = serverName;
+        this.processState = processState;
+        this.autoStartDataDir = domainDataDir.toPath().resolve("auto-start");
+        try {
+            if (Files.notExists(autoStartDataDir)) {
+                Files.createDirectory(autoStartDataDir);
+            }
+        } catch (IOException ex) {
+            HostControllerLogger.ROOT_LOGGER.couldNotCreateDomainAutoStartDirectory(autoStartDataDir, ex);
```
jamezp: It seems a little odd here to log the error as errors in the `getAutoStart` method as well. I'm not sure that it's a better idea to throw an error however. Just wanted to make a comment for a management reviewer to make the final decision.

ehsavoie: Hum, this one is throwing an IllegalStateException.
The "couldNotPersistAutoStartServerStatus" is just an error message as it will work as degraded mode but I don't think it should fail the booting/running of an HC.

jamezp: Actually I should have looked at the logger. You need to add the `throw` here, then it will throw it :)

host-controller/src/main/java/org/jboss/as/host/controller/operations/ServerStartHandler.java
```diff
@@ -108,16 +108,17 @@ public void execute(OperationContext context, ModelNode operation) throws Operat
                 final ServerStatus origStatus = serverInventory.determineServerStatus(serverName);
                 if (origStatus != ServerStatus.STARTED && origStatus != ServerStatus.STARTING) {
                     final ServerStatus status = serverInventory.startServer(serverName, model, blocking);
+                    context.readResource(PathAddress.EMPTY_ADDRESS, false);
```
bstansberry: AIUI you call this to trigger a call to ServerConfigResource.persistAutoStart()? If so you'd need to add a getModel() call. Also, there should be a comment explaining what this is about, or someone will likely delete it.

Same comment applies to other similar spots; e.g. L121 below.

PR #448, opened 2015-01-22
ConfigurationFile fixes

bstansberry:
1) General housekeeping -- comments and such
2) WFCORE-513 -- use the correct boot file in reload in the special cases
3) WFCORE-514 -- use the correct file for taking snapshots when we don't persist changes to the main file

### DIFF
--- a/controller/src/main/java/org/jboss/as/controller/persistence/ConfigurationFile.java
+++ b/controller/src/main/java/org/jboss/as/controller/persistence/ConfigurationFile.java
@@ -68,21 +68,38 @@ public class ConfigurationFile {
     private final File configurationDir;
     private final String rawFileName;
     private final String bootFileName;
+    // File from which boot operations should be parsed; null if currently undetermined
     private volatile File bootFile;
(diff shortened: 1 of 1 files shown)


controller/src/main/java/org/jboss/as/controller/persistence/ConfigurationFile.java
```diff
@@ -68,21 +68,40 @@
     private final File configurationDir;
     private final String rawFileName;
     private final String bootFileName;
+    // File from which boot operations should be parsed; null if currently undetermined
     private volatile File bootFile;
+    /* Whether the next determination of the bootFile should use the .last file in history
+       instead of the {@link #mainFile}. Only relevant if {@link #persistOriginal} is false */
     private volatile boolean reloadUsingLast;
+    // Whether {@link #bootFile has been reset from its first value
+    private volatile boolean bootFileReset;
```
jamezp: Probably not any huge performance hit, but it likely doesn't need to be volatile since all access is guarded by `this`.

PR #449, opened 2015-01-22
Add option to prevent manual mode tests to automaticly run server

ctomc:


jamezp:
I'm still voting we just get rid of the `@ServerControl` annotation. It's only used in manual mode tests and they all have to override the default value for the annotation. The annotation does nothing if the `manual` attribute is set to `false` which is the default. Having the annotation and a system property just doesn't make much sense to me.

ctomc:
@stuartwdouglas wdyt about this?

stuartwdouglas:
I am fine with getting rid of the annotation. Manual and automatic tests should not be in the same maven module anyway

ctomc:
will rework this to have system property to control in what way servers are started, manual or automatic.

PR #451, opened 2015-01-22
Host controller subsystems

kabir:
This is the first iteration of HC subsystems, which allows you to define extensions and subsystems in the host model. In my original approach (which I have not shared with anybody) I was going down a slightly wrong path, attempting to share the extension registry between the host and domain models, and dealing with all the problems resulting from that. Yesterday when dealing with the last problems I had an 'aha' moment and have reimplemented it with a separate extension registries for the domain extensions and the host extensions, which is a lot simpler and means all the stuff I had in there to deal with the shared registry could be removed.

I had to make some changes to how the MutableRootResourceRegistrationProvider is passed in to the parallel extension add handler to make sure that the correct extension registry is used when adding the extensions at boot.

In the JMX subsystem, which is now host capable, I have a RuntimeHostControllerInfoAccessor to defer getting the information about whether we are a master or a slave domain controller until we reach the runtime stage of the boot process. The reason for this is that during boot of the host model, the extension gets initialised before LocalHostControllerInfoImpl.setMasterDomainController() gets called when handling the local/remote domain controller elements.

https://issues.jboss.org/browse/WFCORE-452 (leave open, there is more work coming)

Some outstanding things which are on my list are:
* Adding 'assembly' of host.xml subsystems (similar to how we construct domain.xml and standalone.xml from a template)
* Adding socket binding groups to host.xml (since some subsystems might require that)


controller/src/main/java/org/jboss/as/controller/extension/ExtensionRegistry.java
```diff
@@ -258,10 +302,18 @@ public ExtensionContext getExtensionContext(final String moduleName, ManagementR
      * @throws IllegalStateException if the extension still has subsystems present in {@code rootResource} or its children
      */
     public void removeExtension(Resource rootResource, String moduleName, ManagementResourceRegistration rootRegistration) throws IllegalStateException {
-        // Can't use processType.isServer() to determine where to look for profile reg because a lot of test infrastructure
-        // doesn't add the profile mrr even in HC-based tests
-        ManagementResourceRegistration profileReg = rootRegistration.getSubModel(PathAddress.pathAddress(PathElement.pathElement(PROFILE)));
-        if (profileReg == null) {
+        final ManagementResourceRegistration profileReg;
+        if (rootRegistration.getPathAddress().size() == 0) {
```
ctomc: that is bit odd condition, why not something like getPathAddress() == PathAddress.EMPTY_ADDRESS

jamezp: That makes the assumption that `PathAddress.EMPTY_ADDRESS` is and will always be used when it's truly empty. Using the size seems safer to me.

core-feature-pack/src/main/resources/content/domain/configuration/host-master.xml
```diff
@@ -5,6 +5,9 @@
    and does not itself directly control any servers.
 -->
 <host name="master" xmlns="urn:jboss:domain:3.0">
+    <extensions>
+        <extension module="org.jboss.as.jmx"/>
```
ctomc: please don't use jmx testing  / default configs as i think we still want to move it out of core repo.

kabir: If that happens the tests/config will have to move. We already use it although in other places, e.g. standalone.xml and host.xml

ctomc: I know, and we still keep adding new stuff, there was a PR already for removing jmx subsystem from core. @bstansberry at the time didn't want to merge it because of some bigger refactoring needed in core that included jmx code. 
I don't mind you add this, it will just mean more work when we do move code back to full repo.

kabir:
Unrelated failure

Retest this please!

kabir:
Retest this please

testsuite/domain/src/test/java/org/jboss/as/test/integration/domain/extension/RootResourceDefinition.java
```diff
@@ -69,9 +71,11 @@ public void registerOperations(ManagementResourceRegistration resourceRegistrati
         @Override
         protected void populateModel(ModelNode operation, ModelNode model) throws OperationFailedException {
             NAME.validateAndSet(operation, model);
+        }
 
+        @Override
+        protected void performRuntime(OperationContext context, ModelNode operation, Resource resource)
+                throws OperationFailedException {
```
ctomc: probably added by mistake

bstansberry:
retest this please

PR #453, opened 2015-01-23
[WFCORE-491] Add fast path for disabled audit logging

doctau:
Removes the need for taking the audit logging configuration lock when
audit logging is disabled. Since that is a fair lock, it is heavy weight
when under high contention.


doctau:
In non-scientific benchmarking, with a few threads performing a series of read-resource operations, this reduces the time to completion by 10% in the common case of audit logging being disabled, and had no measurable impact when audit logging was enabled.

It would be good for someone to check over the changes to ensure I haven't introduced any thread safety issues (I don't think I have), and that there aren't any other situations like handler updates that need to disable the fast path.

PR #457, opened 2015-01-27
new jboss-parent & checkstyle

ctomc:
+ fixes in codestyle new checkstyle found

### DIFF
--- a/cli/src/main/java/org/jboss/as/cli/parsing/GlobalCharacterHandlers.java
+++ b/cli/src/main/java/org/jboss/as/cli/parsing/GlobalCharacterHandlers.java
@@ -56,7 +56,9 @@ public class GlobalCharacterHandlers {
         public void handle(ParsingContext ctx)
                 throws OperationFormatException {
         }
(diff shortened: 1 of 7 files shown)


kabir:
Retest this please

jamezp:
The new parent brings in version 3.2 of the `maven-compiler-plugin` which can cause compile issues with annotation processors. https://jira.codehaus.org/browse/MCOMPILER-236 I want to look at this further before allowing that version of the compiler in.

jamezp:
Setting the `useIncrementalCompilation` to `false` in the compiler configuration seems to workaround this error. Is there a reason we need 3.2? I'd be more comfortable leaving it at 3.1 unless we have a compelling reason to upgrade.

ctomc:
3.2 fixes problem with not properly displayed logging/output from annotation processors. Fix that David contributed.
I don't mind disabling incremental compilation if that causes problems. Did you saw this problem manifest also in any of our projects?

jamezp:
Yeah. I can replicate it quite easily by doing a `mvn clean install && mvn install`. The second one fails.

PR #458, opened 2015-01-27
[build-fix] : Fixing build

ehsavoie:
manualmode tests are run with -DallTests
Fixing Log4jAppenderTestCase as it tries to connect before starting the server.

No Jira


jamezp:
We need to hold off enabling the manualmode tests for now. The `org.jboss.as.test.integration.management.util.CustomCLIExecutor` is leaving the CLI process running. I'm going to try to have a look at it now.

ctomc:
@jamezp i think we should still merge this asap. Extra fixes are welcome but can come as part of separate PR

jamezp:
I'm closing this in favor of https://github.com/wildfly/wildfly-core/pull/459

PR #459, opened 2015-01-28
[WFCORE-498] Ensure the server has been started before executing operations

jamezp:
Fix broken logging manual mode test. The test was attempting to use the `ModelControllerClient` before the server has been started.

Moved the `manualmode` module to run with the `allTests` property.

[WFCORE-524] Send a response to the hanging CLI processes to ensure it terminates.


ctomc:
:+1:

PR #460, opened 2015-01-28
Updated jboss-cli to latest version of æsh using low level api

stalep:


stalep:
currently wildfly & eap6 is using a very old version of æsh that is only maintained since its used in those projects. using the latest version also adds several new functionalities we could enable if wanted. aliasing, export, etc.
it would also make it a bit easier if we in the future would be interested to use the new api in æsh.

pom.xml
```diff
@@ -91,7 +91,7 @@
         <version.org.codehaus.woodstox.stax2-api>3.1.4</version.org.codehaus.woodstox.stax2-api>
         <version.org.codehaus.woodstox.woodstox-core-asl>4.4.1</version.org.codehaus.woodstox.woodstox-core-asl>
         <version.org.fusesource.jansi>1.9</version.org.fusesource.jansi>
-        <version.org.jboss.aesh>0.33.14</version.org.jboss.aesh>
+        <version.org.jboss.aesh>0.57-SNAPSHOT</version.org.jboss.aesh>
```
ctomc: don't use snapshot versions

cli/src/main/java/org/jboss/as/cli/impl/CliLauncher.java
```diff
@@ -269,19 +272,33 @@ public static void main(String[] args) throws Exception {
...
             System.out.println(Util.getMessagesFromThrowable(t));
-            exitCode = 1;
-        } finally {
-            if((cmdCtx != null) && !gui) {
+           exitCode = 1;
+       }
+        finally {
+            if(cmdCtx != null && fileMode) {
+                //since the cli is async we give it a bit to finish
+                Thread.sleep(500);
```
ctomc: yuck

cli/src/main/java/org/jboss/as/cli/impl/CliLauncher.java
```diff
@@ -269,19 +272,33 @@ public static void main(String[] args) throws Exception {
...
                 cmdCtx.terminateSession();
                 if(cmdCtx.getExitCode() != 0) {
                     exitCode = cmdCtx.getExitCode();
                 }
+                System.exit(exitCode);
             }
-            if (!gui) {
+            else if(cmdCtx != null && commandMode) {
+                //since the cli is async we give it a bit to finish
+                Thread.sleep(500);
```
ctomc: yuck

stalep:
removed the sleep, something i added to test and forgot to remove before the pr.

stalep:
the plan is to release a final of aesh 0.57 by this week, ill update the pr with the latest version.

bstansberry:
Closing, as there is a different PR in the works.

PR #461, opened 2015-01-28
[WFCORE-232] : Unclean shutdown of deployment scanner

ehsavoie:
Avoiding tracing the exception when the Executor is shutting down.
Adding a test to check that reload does occurs without blocking.

Jira: https://issues.jboss.org/browse/WFCORE-232

### DIFF
--- a/deployment-scanner/src/main/java/org/jboss/as/server/deployment/scanner/DeploymentScannerService.java
+++ b/deployment-scanner/src/main/java/org/jboss/as/server/deployment/scanner/DeploymentScannerService.java
@@ -49,16 +49,16 @@ import org.jboss.msc.value.InjectedValue;
  */
 public class DeploymentScannerService implements Service<DeploymentScanner> {
 
-    private long interval;
+    private final long interval;
     private TimeUnit unit = TimeUnit.MILLISECONDS;
-    private boolean enabled;
-    private boolean autoDeployZipped;
-    private boolean autoDeployExploded;
-    private boolean autoDeployXml;
-    private Long deploymentTimeout;
+    private final boolean enabled;
+    private final boolean autoDeployZipped;
+    private final boolean autoDeployExploded;
+    private final boolean autoDeployXml;
+    private final Long deploymentTimeout;
     private final String relativeTo;
     private final String path;
-    private boolean rollbackOnRuntimeFailure;
+    private final boolean rollbackOnRuntimeFailure;
 
     /**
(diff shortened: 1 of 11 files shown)


deployment-scanner/src/main/java/org/jboss/as/server/deployment/scanner/DeploymentScannerService.java
```diff
@@ -178,6 +178,7 @@ public DeploymentOperations create() {
      */
     @Override
     public synchronized void stop(StopContext context) {
+        context.asynchronous();
```
kabir: I am not super familiar with this but the usage pattern elsewhere seems to be
1) inject an executor
2) Call context.asynchronous()
3) Perform the work which may block in a Runnable
4) Call context context.complete() at the end of the Runnable

i.e. the heavy work should happen in a different thread.

ehsavoie: I'm not also but the scanner.stop is happening in another thread I understand correctly.

deployment-scanner/src/main/java/org/jboss/as/server/deployment/scanner/FileSystemDeploymentService.java
```diff
@@ -521,6 +529,7 @@ private void executeScannerTasks(List<ScannerTask> scannerTasks, DeploymentOpera
             for (ScannerTask task : scannerTasks) {
                 task.recordInProgress(); // puts down .isdeploying, .isundeploying
                 final ModelNode update = task.getUpdate();
+                ROOT_LOGGER.infof("Deployment scan of [%s] found update action [%s]", deploymentDir, update);
```
bstansberry: This should be debug.

bstansberry:
The fix looks fine (except for the INFO logging) but the test case is really fragile. For example,

https://github.com/wildfly/wildfly-core/pull/461/files#diff-714e40ed01d8912c0ea7422845e5944dR5

will break as soon as that class is edited. And the byteman poking into all sorts of internal to execute the reload is really clever, but very fragile. :)

We could keep that, but @Ignore it.

Since this isn't adding any new blocking behavior into DeploymentScannerService.stop, this seems like behavior that can be tested in a unit test. For example pass a custom DeploymentOperations that blocks in deploy into FileSystemDeploymentService.startScanner and then in another thread call stopScanner.

PR #463, opened 2015-01-30
[WFCORE-499] : System properties are broken in domain mode.

ehsavoie:
Allowing the boot-time attribute for managed servers.

Jira: https://issues.jboss.org/browse/WFCORE-499

### DIFF
--- a/testsuite/domain/src/test/java/org/jboss/as/test/integration/domain/suites/CoreResourceManagementTestCase.java
+++ b/testsuite/domain/src/test/java/org/jboss/as/test/integration/domain/suites/CoreResourceManagementTestCase.java
@@ -22,9 +22,11 @@
 
 package org.jboss.as.test.integration.domain.suites;
 
+import static org.hamcrest.CoreMatchers.is;
 import static org.jboss.as.controller.descriptions.ModelDescriptionConstants.ADD;
 import static org.jboss.as.controller.descriptions.ModelDescriptionConstants.ANY_ADDRESS;
 import static org.jboss.as.controller.descriptions.ModelDescriptionConstants.AUTO_START;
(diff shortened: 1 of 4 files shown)


bstansberry:
The "boot-time" attribute doesn't belong on a server. It's an instruction to the HC to apply the property as part of the command used to launch the java.lang.Process; i.e. used in ManagedServerBootCmdFactory. So any system-property resource with that set to 'true' should not become part of the server's configuration model.

bstansberry:
My guess is this was the problem:

https://github.com/wildfly/wildfly-core/pull/387

and this is the fix:

https://github.com/bstansberry/wildfly-core/commit/0065d42f72623690f8366368d68ae2f8a0847a78

bstansberry:
That commits fixes the https://issues.jboss.org/browse/WFLY-4299 problem, which looks to be just a specific example of WFCORE-499.

testsuite/domain/src/test/java/org/jboss/as/test/integration/domain/suites/CoreResourceManagementTestCase.java
```diff
@@ -369,6 +389,50 @@ public void testSystemPropertyManagement() throws IOException {
         Assert.assertEquals(origPropCount, returnVal.asList().size());
     }
 
+    @Test
+    public void testSystemPropertyBootTime() throws IOException {
```
bstansberry: I think for a test like this it would be good to use the core-service=platform-mbean stuff to validate the actual system property in the server VM. org.jboss.as.test.deployment.trivial.ServiceActivatorDeploymentUtil has some utility code related to reading system properties that way.

ehsavoie:
Fixing missing code and covering with a test.

PR #464, opened 2015-02-02
WFCORE-493 Modules imported with "annotations" flag have their indexes c...

tadamski:
...alculated on the fly if no static index is attached to them.


stuartwdouglas:
Looks good to me, although we also need some way to do this if the pre-computed index is the wrong version.

PR #465, opened 2015-02-02
Make setting deprecated data bit easier for resources

ctomc:


bstansberry:
This kind of ease of use should be done with a builder API.

ctomc:
@bstansberry agreed but we don't really have(ResourceBuilder kinda is) proper builder API for ResourceDefinition or at least almost no one uses it. 
This change will allow to add deprecated data to resource definitions without too much impact on existing code.

PR #270, opened 2014-10-27
WFCORE-135: remove custom runtime permissions

emmartins:

PR #466, opened 2015-02-02
WFCORE-135: replace jdk runtime permissions with controller runtime perm...

bstansberry:
...issions, rework permission names to follow similar format

This is just a rebase of #270 to resolve a conflict in ControllerLogger.


bstansberry:
I'll merge this once the PR run is done.

PR #468, opened 2015-02-03
[WFLY-4188] resolve expression values for client-mapping

soul2zimate:
[WFLY-4188] resolve expression values for client-mapping
"source-network" "destination-address" and "destination-port" are expression allowed attributes.
client-mappings change requires a reload operation as well.


bstansberry:
The resolveModelAttribute change looks good; thanks!

Please remove the change to ClientMappingsHandler, as it is inconsistent with other uses of a SocketBinding. If SocketBinding being mutable is a problem, please start a wildfly-dev list thread so we can discuss it.

soul2zimate:
OK, it's not really runtime, it writes for unbound socket-binding. 
I have rebased to add OperationContext in parameters list to resolve expression values. Thanks.

bstansberry:
Ah, ok, so the ClientMappingsHandler change wasn't because mutating the SocketBinding was a problem. This looks fine; thanks!

PR #473, opened 2015-02-05
[WFCORE-531] Update subsystem registration to use a ModelVersion instead of the component parts of the version.

darranl:


controller/src/main/java/org/jboss/as/controller/extension/ExtensionRegistry.java
```diff
@@ -431,18 +436,20 @@ public SubsystemRegistration registerSubsystem(String name, int majorVersion, in
...
+        }
+
+        public SubsystemRegistration registerSubsystem(String name, ModelVersion version, boolean deprecated) {
             assert name != null : "name is null";
             checkNewSubystem(extension.extensionModuleName, name);
             SubsystemInformationImpl info = extension.getSubsystemInfo(name);
-            info.setMajorVersion(majorVersion);
-            info.setMinorVersion(minorVersion);
-            info.setMicroVersion(microVersion);
+            info.setVersion(version);
```
kabir: I think I would throw an exception if version is null, and get rid of the null checks in SubsystemInformationImpl

controller/src/main/java/org/jboss/as/controller/extension/ExtensionRegistry.java
```diff
@@ -521,29 +526,21 @@ void addParsingNamespace(final String namespace) {
...
         }
 
         @Override
         public Integer getManagementInterfaceMinorVersion() {
-            return minorVersion;
-        }
-
-        private void setMinorVersion(Integer minorVersion) {
-            this.minorVersion = minorVersion;
+            return version != null ? version.getMinor() : null;
```
kabir: Not a big deal since this is public API, but we could replace these getters with a simple getVersion().
If not, remove the null check

controller/src/main/java/org/jboss/as/controller/extension/ExtensionRegistry.java
darranl: IMO that is established API that I don't want to be touching here.  I just want to be ensuring I can use a ModelVersion not revisiting the mandating all subsytems have a version.

controller/src/main/java/org/jboss/as/controller/extension/ExtensionRegistry.java
darranl: This is established behaviour having a null check here, these changes are to retain compatibility.

controller/src/main/java/org/jboss/as/controller/extension/ExtensionRegistry.java
kabir: All the exisiting registerSubsystem() methods take one or more int parameters which cannot be null, so you end up with a version whether you like it or not

bstansberry:
Looks good; I agree that changing the null semantics shouldn't be part of this work. 

Probably shouldn't happen at all.

I believe the reason null is allowed is because ExtensionParsingContextImpl.setSubsystemXmlMapping can establish a SubsystemInformationImpl and it's only later when ExtensionContextImpl.registerSubsystem gets called that the version gets set. So there's a tiny window where it is null. And we expose this data structure via the management API, so it's possible a read of the data can happen during that window.

PR #479, opened 2015-02-10
Offline CLI -- standalone mode

bstansberry:
Not ready for merging; let any discussion at http://lists.jboss.org/pipermail/wildfly-dev/2015-February/003611.html play out a bit.

It works though. :)

A wiki describing it is at https://developer.jboss.org/docs/DOC-53050


bstansberry:
retest this please

embedded/src/main/java/org/jboss/as/embedded/EmbeddedStandAloneServerFactory.java
```diff
@@ -0,0 +1,479 @@
...
+    private static File getTempRoot(Properties props) {
+        String tempRoot = props.getProperty(JBOSS_EMBEDDED_ROOT, null);
+        if (tempRoot == null) {
+            return null;
+        }
+
+        File root = new File(tempRoot);
+        if (!root.exists()) {
+            //Attempt to try to create the directory, in case something like target/embedded was specified
+            root.mkdirs();
```
ctomc: This is really minor, but could you use Path api here? As it will return better error message than File will. giving some context on what is going on...

embedded/src/main/java/org/jboss/as/embedded/EmbeddedStandAloneServerFactory.java
```diff
@@ -0,0 +1,479 @@
...
+        for (String current : src.list()) {
+            final File srcFile = new File(src, current);
+            final File destFile = new File(dest, current);
+
+            if (srcFile.isDirectory()) {
+                destFile.mkdir();
+                copyDirectory(srcFile, destFile);
+            } else {
+                try {
+                    final InputStream in = new BufferedInputStream(new FileInputStream(srcFile));
```
ctomc: Files#copy(....)

embedded/src/main/java/org/jboss/as/embedded/StandaloneServer.java
```diff
@@ -0,0 +1,65 @@
...
+ */
+public interface StandaloneServer {
+
+    // TODO: use a DeploymentPlan
+    @Deprecated
+    void deploy(File file) throws IOException, ExecutionException, InterruptedException;
+
+    // TODO: use a DeploymentPlan
+    @Deprecated
+    void undeploy(File file) throws ExecutionException, InterruptedException;
```
ctomc: why adding deprecated methods to interface if it is new impl... it could probably be removed

bstansberry: I didn't really intend this to be a new impl, just a port of the existing one. I gave it a new module name just so it could temporarily co-exist with the one in WF full.

This "port vs ???" question is one of the areas to discuss about this work though. If it's *not* just a port, sure, deprecated methods should not come in. If it is just a port, getting rid of them is a valid question too, but it's a separate issue.

bstansberry: The Path/Files comments you made though are valid either way; those are internal details we should change.

PR #481, opened 2015-02-10
[WFCORE-536] : Sanitize origins ports in CORS.

ehsavoie:
Sanitizing default port from urls.
Fixing double port issue when defining the default origin.

Jira: https://issues.jboss.org/browse/WFCORE-536

### DIFF
--- a/domain-http/interface/src/main/java/org/jboss/as/domain/http/server/cors/CorsUtil.java
+++ b/domain-http/interface/src/main/java/org/jboss/as/domain/http/server/cors/CorsUtil.java
@@ -30,14 +30,14 @@ import io.undertow.server.handlers.ResponseCodeHandler;
 import io.undertow.util.HeaderMap;
 import io.undertow.util.Headers;
 import io.undertow.util.Methods;
+import io.undertow.util.NetworkUtils;
 import java.util.Collection;
 
 /**
- *
+ * Utility class for CORS handling.
  * @author <a href="mailto:ehugonne@redhat.com">Emmanuel Hugonnet</a> (c) 2014 Red Hat, inc.
  */
 public class CorsUtil {
-
     public static boolean isCoreRequest(HeaderMap headers) {
         return headers.contains(ORIGIN)
                 || headers.contains(ACCESS_CONTROL_REQUEST_HEADERS)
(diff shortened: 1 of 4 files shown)


ehsavoie:
retest this please

domain-http/interface/src/main/java/org/jboss/as/domain/http/server/cors/CorsUtil.java
```diff
@@ -70,17 +70,35 @@ public static String matchOrigin(HttpServerExchange exchange, Collection<String>
...
+    }
+
+    /**
+     * Removes the port from a URL if this port is the default one for the URL's scheme.
+     * @param url the url to be sanitized.
+     * @return  the sanitized url.
+     */
+    public static String sanitizeDefaultPort(String url) {
+        String scheme = url.substring(0, url.indexOf("://"));
+        int portIndex = url.indexOf(':', scheme.length() + 3);
```
bstansberry: This doesn't account for IPv6 addresses in URLs.

domain-http/interface/src/main/java/org/jboss/as/domain/http/server/cors/CorsUtil.java
```diff
@@ -53,10 +53,10 @@ public static boolean isCoreRequest(HeaderMap headers) {
      */
     public static String matchOrigin(HttpServerExchange exchange, Collection<String> allowedOrigins) throws Exception {
         HeaderMap headers = exchange.getRequestHeaders();
-        String origin = headers.getFirst(Headers.ORIGIN);
+        String origin = sanitizeDefaultPort(headers.getFirst(Headers.ORIGIN));
```
bstansberry: I believe returning this sanitized value will violate the CORS spec. See Section 5.1 of http://www.w3.org/TR/cors/#access-control-allow-origin-response-header:

```The Access-Control-Allow-Origin header indicates whether a resource can be shared based by returning the value of the Origin request header, "*", or "null" in the response.```

The sanitized value is not the value of the Origin header.

I think you need to retain the origin value and only used the sanitized one in the matching checks.

Comment (ehsavoie) [inline domain-http/interface/src/main/java/org/jboss/as/domain/http/server/cors/CorsUtil.java]: You're right.
I have also added the possibility to define '*' as allowed-origin.

File: domain-http/interface/src/main/java/org/jboss/as/domain/http/server/cors/CorsUtil.java [hunk 3]
```diff
@@ -45,7 +44,9 @@ public static boolean isCoreRequest(HeaderMap headers) {
     }
 
     /**
-     * Match the Origin header with the allowed origins. If it doesn't match then a 403 response code is set on the response.
+     * Match the Origin header with the allowed origins.
+     * If the allowed-origins contains '\*' then it always match and returns '\*'.
```
bstansberry: I don't see this part.

bstansberry:
This is only tangentially related to this PR but is something we should address before Final:

The Origin header can continue multiple values, separated by a space:

http://tools.ietf.org/html/rfc6454#section-7.1

I'm not sure what the semantics are for the response header, but I'm sure the spec spells it out.

domain-http/interface/src/main/java/org/jboss/as/domain/http/server/cors/CorsUtil.java
ehsavoie: yes I removed it but forgot to update the doc

ehsavoie:
According to the specs we need to return the Origin header that means returning the whole list:

The Access-Control-Allow-Origin header indicates whether a resource can be shared based by returning the value of the Origin request header, "*", or "null" in the response. ABNF:
Access-Control-Allow-Origin = "Access-Control-Allow-Origin" ":" origin-list-or-null | "*"

PR #482, opened 2015-02-11
[WFCORE-537] : Add path support into the management API.

ehsavoie:
Allow the use of String to pass an operation address.

Jira: https://issues.jboss.org/browse/WFCORE-537

### DIFF
--- a/controller/src/main/java/org/jboss/as/controller/PathAddress.java
+++ b/controller/src/main/java/org/jboss/as/controller/PathAddress.java
@@ -54,8 +54,7 @@ public class PathAddress implements Iterable<PathElement> {
     public static final PathAddress EMPTY_ADDRESS = new PathAddress(Collections.<PathElement>emptyList());
 
     /**
-     * Creates a PathAddress from the given ModelNode address.  The given node is expected
-     * to be an address node.
+     * Creates a PathAddress from the given ModelNode address. The given node is expected to be an address node.
      *
      * @param node the node (cannot be {@code null})
      *
@@ -73,11 +72,9 @@ public class PathAddress implements Iterable<PathElement> {
                 Property prop = null;
                 if (element.getType() == ModelType.PROPERTY || element.getType() == ModelType.OBJECT) {
                     prop = element.asProperty();
-                }
-                else if (key == null) {
+                } else if (key == null) {
                     key = element.asString();
-                }
-                else {
+                } else {
                     prop = new Property(key, element);
                 }
                 if (prop != null) {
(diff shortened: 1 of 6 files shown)


controller/src/main/java/org/jboss/as/controller/operations/common/Util.java
```diff
@@ -179,16 +178,83 @@ public static ModelNode validateOperation(ModelNode operation) {
             try {
                 operation.get(OP_ADDR).asList();
             } catch (IllegalArgumentException ex) {
-                if(errors.length() > 0) {
-                    errors.append(System.lineSeparator());
+                try {
+                    ModelNode address = parseAddress(operation.get(OP_ADDR).asString());
```
ctomc: this looks mighty odd.
why not check the type of OP_ADDR attribute? something like if (operation.get(OP_ADDR).getType() == ModelType.STRING and if it is string try to parse it as such, doing this with try |catch is just wrong

controller/src/main/java/org/jboss/as/controller/operations/common/Util.java
```diff
@@ -179,16 +178,83 @@ public static ModelNode validateOperation(ModelNode operation) {
...
+        }
+        String trimmedAddress = address.trim();
+        if (trimmedAddress.charAt(0) != '/' || !Character.isAlphabetic(trimmedAddress.charAt(1))) {
+            throw new IllegalArgumentException();
+        }
+        char[] characters = address.toCharArray();
+        boolean escaped = false;
+        StringBuffer keyBuffer = new StringBuffer();
+        StringBuffer valueBuffer = new StringBuffer();
+        StringBuffer currentBuffer = keyBuffer;
```
ctomc: StringBuilder would be sufficient for all 3

controller/src/main/java/org/jboss/as/controller/operations/common/Util.java
```diff
@@ -179,16 +178,83 @@ public static ModelNode validateOperation(ModelNode operation) {
...
         }
         return responseNode;
     }
+
+    public static ModelNode parseAddress(String address) throws IllegalArgumentException {
+        ModelNode parsedAddress = new ModelNode().setEmptyList();
+        if (address == null || address.isEmpty()) {
+            return parsedAddress;
+        }
+        String trimmedAddress = address.trim();
```
ctomc: you could trim it above already and if it is empty return empty address.

controller/src/main/java/org/jboss/as/controller/operations/common/Util.java
```diff
@@ -179,16 +178,83 @@ public static ModelNode validateOperation(ModelNode operation) {
...
         }
-        if(errors.length() > 0) {
+        if (errors.length() > 0) {
             responseNode.get(OUTCOME).set(FAILED);
             responseNode.get(FAILURE_DESCRIPTION).set(errors.toString());
         }
         return responseNode;
     }
+
+    public static ModelNode parseAddress(String address) throws IllegalArgumentException {
```
ctomc: This parse address belongs more in PathAddress than into Util class

controller/src/main/java/org/jboss/as/controller/operations/common/Util.java
ehsavoie: I agrre, forgot to do that

controller/src/main/java/org/jboss/as/controller/operations/common/Util.java
ehsavoie: I'm not sure about that since I return a ModelNode instead of a PathAddress

controller/src/main/java/org/jboss/as/controller/PathAddress.java
```diff
@@ -150,6 +147,68 @@ public static PathAddress pathAddress(PathAddress parent, PathElement... element
         return pathAddress(list);
     }
 
+    public static PathAddress parseAddress(String address) throws IllegalArgumentException {
```
bstansberry: This is actually parsing a CLI-style address, so the method name should reflect that; something like parseCLIStyleAddress. This names corresponds to the toCLIStyleString method.

The other style is HTTP style, where the address is part of a GET URL. That style uses '/' instead of '='.

controller/src/main/java/org/jboss/as/controller/operations/common/Util.java
```diff
@@ -177,15 +176,20 @@ public static ModelNode validateOperation(ModelNode operation) {
...
+                    operation.get(OP_ADDR).set(address);
+                } else {
+                    operation.get(OP_ADDR).asList();
+                }
             } catch (IllegalArgumentException ex) {
-                if(errors.length() > 0) {
+                if (errors.length() > 0) {
                     errors.append(System.lineSeparator());
                 }
                 errors.append(ControllerLogger.ROOT_LOGGER.attributeIsWrongType(OP_ADDR, ModelType.LIST, operation.get(OP_ADDR).getType()).getMessage());
```
bstansberry: This isn't the correct message in the parsing case.

controller/src/main/java/org/jboss/as/controller/PathAddress.java
```diff
@@ -150,6 +147,68 @@ public static PathAddress pathAddress(PathAddress parent, PathElement... element
...
+        parsedAddress = addpathAddressElement(parsedAddress, keyBuffer, valueBuffer);
+        return parsedAddress;
+    }
+
+    private static PathAddress addpathAddressElement(PathAddress parsedAddress, StringBuilder keyBuffer, StringBuilder valueBuffer) {
+        if (keyBuffer.length() > 0) {
+            if (valueBuffer.length() > 0) {
+                return parsedAddress.append(PathElement.pathElement(keyBuffer.toString(), valueBuffer.toString()));
+            }
+            return parsedAddress.append(PathElement.pathElement(keyBuffer.toString()));
```
bstansberry: I don't think we should allow this. PathElement.pathElement(String) is a convenient mechanism for server side developers, but an incompletely specified String is not valid as an end user API. If they want a wildcard, they need to use key=*.

ehsavoie:
retest this please

bstansberry:
Looks good but needs a rebase. Please ping me when you push it so I can get it in.

PR #485, opened 2015-02-12
WFCORE-538: Add GZIP handler for the management endpoint

hpehl:
https://issues.jboss.org/browse/WFCORE-538

### DIFF
--- a/domain-http/interface/src/main/java/org/jboss/as/domain/http/server/DomainApiCheckHandler.java
+++ b/domain-http/interface/src/main/java/org/jboss/as/domain/http/server/DomainApiCheckHandler.java
@@ -27,12 +27,15 @@ import static org.jboss.as.domain.http.server.logging.HttpServerLogger.ROOT_LOGG
 import io.undertow.server.HttpHandler;
 import io.undertow.server.HttpServerExchange;
 import io.undertow.server.handlers.BlockingHandler;
+import io.undertow.server.handlers.encoding.EncodingHandler;
 import io.undertow.util.HeaderMap;
 import io.undertow.util.Headers;
 import io.undertow.util.HttpString;
 import io.undertow.util.Methods;
 import java.util.ArrayList;
 import java.util.Collection;
+import java.util.Collections;
+
 import org.jboss.as.controller.ControlledProcessState;
 import org.jboss.as.controller.ControlledProcessStateService;
 import org.jboss.as.controller.ModelController;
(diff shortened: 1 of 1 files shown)


domain-http/interface/src/main/java/org/jboss/as/domain/http/server/DomainApiCheckHandler.java
```diff
@@ -57,9 +60,9 @@
 
     DomainApiCheckHandler(final ModelController modelController, final ControlledProcessStateService controlledProcessStateService, final Collection<String> allowedOrigins) {
         this.controlledProcessStateService = controlledProcessStateService;
-        domainApiHandler = new BlockingHandler(new SubjectDoAsHandler(new DomainApiHandler(modelController)));
+        domainApiHandler = new BlockingHandler(new SubjectDoAsHandler(new EncodingHandler.Builder().build(Collections.<String,Object>emptyMap()).wrap(new DomainApiHandler(modelController))));
         addContentHandler = new BlockingHandler(new SubjectDoAsHandler(new DomainApiUploadHandler(modelController)));
-        genericOperationHandler = new BlockingHandler(new SubjectDoAsHandler(new DomainApiGenericOperationHandler(modelController)));
+        genericOperationHandler = new BlockingHandler(new SubjectDoAsHandler(new EncodingHandler.Builder().build(Collections.<String,Object>emptyMap()).wrap(new DomainApiGenericOperationHandler(modelController))));
```
ctomc: wont that enable gzip for all requests not just ones we want them to?

hpehl:
Yes that's right. This will gzip all requests. What kind of requests would you like to exclude?

bstansberry:
I believe this is fine, because EncodingHandler will correctly limit this based on the presence and contents of the Accept-Encoding HTTP header. At a glance that class follows the instructions of Section 14.3 of RFC 2616:

http://www.w3.org/Protocols/rfc2616/rfc2616-sec14.html

Particularly

```
If an Accept-Encoding field is present in a request, and if the server cannot send a response which is acceptable according to the Accept-Encoding header, then the server SHOULD send an error response with the 406 (Not Acceptable) status code.

If no Accept-Encoding field is present in a request, the server MAY assume that the client will accept any content coding. In this case, if "identity" is one of the available content-codings, then the server SHOULD use the "identity" content-coding, unless it has additional information that a different content-coding is meaningful to the client.
```

ctomc:
i forgot to reply to this. Change is fine, as Brian points out if request doesn't contain accept header it won't be compressed.

PR #487, opened 2015-02-12
Minor fixes in WrapperSSLContext

ceharris:
`HttpsURLConnection` invokes `createSocket(Socket, String, int, boolean)`. Extension modules that use `SSLContextService` to inject an `SSLContext` get an NPE when trying to use the socket factory from the
injected context to open a connection to a URL.

The fix here simply invokes the corresponding method on the wrapped socket factory.

### DIFF
--- a/domain-management/src/main/java/org/jboss/as/domain/management/security/WrapperSSLContext.java
+++ b/domain-management/src/main/java/org/jboss/as/domain/management/security/WrapperSSLContext.java
@@ -161,7 +161,7 @@ class WrapperSSLContext extends SSLContext {
             @Override
             public Socket createSocket(InetAddress address, int port, InetAddress localAddress, int localPort)
                     throws IOException {
-                Socket socket = createSocket(address, port, localAddress, localPort);
+                Socket socket = wrapped.createSocket(address, port, localAddress, localPort);
                 setSslParams(socket);
                 return socket;
             }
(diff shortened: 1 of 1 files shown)


kabir:
@ceharris Welcome :-) 

ceharris:
I agree to the terms of the MIT License for this (very minor :-)  contribution.

ctomc:
this is ok to test

domain-management/src/main/java/org/jboss/as/domain/management/security/WrapperSSLContext.java
```diff
@@ -122,8 +122,7 @@ private WrapperSSLSocketFactory(final SSLSocketFactory wrapped) {
 
             @Override
             public Socket createSocket(Socket s, String host, int port, boolean autoClose) throws IOException {
-                // TODO Auto-generated method stub
-                return null;
+                return wrapped.createSocket(s, host, port, autoClose);
```
bstansberry: Before returning it, should pass the created socket to the setSslParams(Socket) method, like the other variants do.

While looking at this, I (aka my IDE) noticed that this variant recurses infinitely:

{code}
            public Socket createSocket(InetAddress address, int port, InetAddress localAddress, int localPort)
                    throws IOException {
                Socket socket = createSocket(address, port, localAddress, localPort);
                setSslParams(socket);
                return socket;
            }
{code}

That call createSocket call should be changed to wrapped.createSocket(...)

Do you mind fixing that in this pull request?

bstansberry:
Thanks so much!

ceharris:
Glad to give back in very small way.  You guys rock.

PR #488, opened 2015-02-13
[WFCORE-542] Clarify JavaDoc for deployment sub-models

jamezp:
Note the change in the implementation to not just get resources if they've already been created. It doesn't seem to fit the `create` prefix if it doesn't actually create the resource. Also a `null` `PathElement` or `PathAddress.EMPTY_ADDRESS` are no allowed so the subsystem resource can be created or a custom resource an be used.

The JavaDoc could definitely use a good review as well to ensure it make sense. I tried to clarify what the expected results should be. The result of each method is a `ModelNode`, but the JavaDoc uses resource quite a bit as that's what's actually being created.


jamezp:
I broke full. I'll close this until I can do some testing.

PR #491, opened 2015-02-13
[WFCORE-353] : CLI command for executing the deployment scanner.

ehsavoie:
Adding a new operation that will run a single scan.
This scan will run asynchronously and its failures won't make a composite op rollback but if the composite op rollback then the result of the scan will rollback too.

Jira: https://issues.jboss.org/browse/WFCORE-353

### DIFF
--- a/deployment-scanner/src/main/java/org/jboss/as/server/deployment/scanner/DeploymentScannerDefinition.java
+++ b/deployment-scanner/src/main/java/org/jboss/as/server/deployment/scanner/DeploymentScannerDefinition.java
@@ -125,4 +125,11 @@ public class DeploymentScannerDefinition extends SimpleResourceDefinition {
         resourceRegistration.registerReadWriteAttribute(DEPLOYMENT_TIMEOUT, null, WriteDeploymentTimeoutAttributeHandler.INSTANCE);
         resourceRegistration.registerReadWriteAttribute(RUNTIME_FAILURE_CAUSES_ROLLBACK, null, WriteRuntimeFailureCausesRollbackAttributeHandler.INSTANCE);
     }
+
+    @Override
+    public void registerOperations(ManagementResourceRegistration resourceRegistration) {
+        super.registerOperations(resourceRegistration);
+        resourceRegistration.registerOperationHandler(FileSystemDeploymentScanHandler.DEFINITION, FileSystemDeploymentScanHandler.INSTANCE);
+    }
+
 }
--- a/deployment-scanner/src/main/java/org/jboss/as/server/deployment/scanner/FileSystemDeploymentScanHandler.java
+++ b/deployment-scanner/src/main/java/org/jboss/as/server/deployment/scanner/FileSystemDeploymentScanHandler.java
@@ -1,0 +1,79 @@
+/*
+ * Copyright (C) 2014 Red Hat, inc., and individual contributors
+ * as indicated by the @author tags. See the copyright.txt file in the
+ * distribution for a full listing of individual contributors.
+ *
+ * This library is free software; you can redistribute it and/or
+ * modify it under the terms of the GNU Lesser General Public
+ * License as published by the Free Software Foundation; either
+ * version 2.1 of the License, or (at your option) any later version.
+ *
+ * This library is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the GNU
+ * Lesser General Public License for more details.
+ *
+ * You should have received a copy of the GNU Lesser General Public
+ * License along with this library; if not, write to the Free Software
+ * Foundation, Inc., 51 Franklin Street, Fifth Floor, Boston,
+ * MA 02110-1301  USA
+ */
+package org.jboss.as.server.deployment.scanner;
+
+import static org.jboss.as.controller.descriptions.ModelDescriptionConstants.OP_ADDR;
+
+import org.jboss.as.controller.OperationContext;
+import org.jboss.as.controller.OperationDefinition;
+import org.jboss.as.controller.OperationFailedException;
+import org.jboss.as.controller.OperationStepHandler;
+import org.jboss.as.controller.PathAddress;
+import org.jboss.as.controller.SimpleOperationDefinitionBuilder;
+import org.jboss.dmr.ModelNode;
+import org.jboss.msc.service.ServiceController;
+
+/**
+ * Operation handler to run a single scan by a DeploymentScanner.
+ * @author <a href="mailto:ehugonne@redhat.com">Emmanuel Hugonnet</a> (c) 2014 Red Hat, inc.
+ */
+public class FileSystemDeploymentScanHandler implements OperationStepHandler {
+
+    public static final String OPERATION_NAME = "run-scan";
+    public static final OperationDefinition DEFINITION = new SimpleOperationDefinitionBuilder(
+            OPERATION_NAME, DeploymentScannerExtension.getResourceDescriptionResolver("deployment.scanner")).build();
+
+    public static final FileSystemDeploymentScanHandler INSTANCE = new FileSystemDeploymentScanHandler();
+
+    private FileSystemDeploymentScanHandler() {
+    }
+
+    @Override
+    public void execute(OperationContext context, ModelNode operation) throws OperationFailedException {
+        context.addStep(new OperationStepHandler() {
+            @Override
+            public void execute(OperationContext context, ModelNode operation) throws OperationFailedException {
+                final FileSystemDeploymentService scanner = getExistingScanner(context, operation);
+                if (scanner != null) {
+                    context.completeStep(new OperationContext.ResultHandler() {
+                        @Override
+                        public void handleResult(OperationContext.ResultAction resultAction, OperationContext context, ModelNode operation) {
+                            if (resultAction == OperationContext.ResultAction.KEEP) {
+                                scanner.singleScan();
+                            }
+                        }
+                    });
+                }
+            }
+        }, OperationContext.Stage.RUNTIME);
+    }
+
+    private FileSystemDeploymentService getExistingScanner(OperationContext context, ModelNode operation) throws OperationFailedException {
+        final PathAddress address = PathAddress.pathAddress(operation.get(OP_ADDR));
+        final String name = address.getLastElement().getValue();
+        ServiceController<?> serviceController = context.getServiceRegistry(true).getService(DeploymentScannerService.getServiceName(name));
+        if(serviceController != null && serviceController.getState() == ServiceController.State.UP) {
+            DeploymentScannerService service =  (DeploymentScannerService) serviceController.getService();
+            return (FileSystemDeploymentService) service.getValue();
+        }
+        return null;
+    }
+}
--- a/deployment-scanner/src/main/java/org/jboss/as/server/deployment/scanner/FileSystemDeploymentService.java
+++ b/deployment-scanner/src/main/java/org/jboss/as/server/deployment/scanner/FileSystemDeploymentService.java
@@ -385,6 +385,13 @@ class FileSystemDeploymentService implements DeploymentScanner {
         }
     }
 
+    /** Perform a one-off scan to establish deployment tasks */
+    void singleScan() {
+        assert deploymentOperationsFactory != null : "deploymentOperationsFactory is null";
+        ManualScanCallable manualScan = new ManualScanCallable();
+        scheduledExecutor.submit(manualScan);
+    }
+
     /** Perform a normal scan */
     void scan() {
         if (acquireScanLock()) {
@@ -415,7 +422,7 @@ class FileSystemDeploymentService implements DeploymentScanner {
 
         if (acquireScanLock()) {
             try {
-                ROOT_LOGGER.tracef("Performing a post-boot forced undeploy scan for scan directory %s", deploymentDir.getAbsolutePath());
+                ROOT_LOGGER.tracef("Performing a post-boot forced undeploy scan for scan directory %oManu", deploymentDir.getAbsolutePath());
                 ScanContext scanContext = new ScanContext(deploymentOperations);
 
                 // Add remove actions to the plan for anything we count as
@@ -523,6 +530,7 @@ class FileSystemDeploymentService implements DeploymentScanner {
                     scannerTasks.add(new UndeployTask(missing.getKey(), missing.getValue().parentFolder, scanContext.scanStartTime, false));
                 }
                 try {
+                    scanResult.tasks = scannerTasks;
                     executeScannerTasks(scannerTasks, deploymentOperations, oneOffScan, scanResult);
                 } catch (InterruptedException e) {
                     Thread.currentThread().interrupt();
@@ -616,6 +624,26 @@ class FileSystemDeploymentService implements DeploymentScanner {
                     for (ScannerTask current : scannerTasks) {
                         current.handleFailureResult(results);
                     }
+                }
+            }
+        }
+    }
+
+    private class ManualScanCallable implements Runnable {
+        @Override
+        public void run() {
+            DeploymentOperations operations = deploymentOperations;
+            if (operations == null) {
+                operations = deploymentOperationsFactory.create();
+            }
+            if (acquireScanLock()) {
+                try {
+                    DeploymentScannerLogger.ROOT_LOGGER.debug("Manual scan launched");
+                    scan(true, operations);
+                } catch (Exception e) {
+                    ROOT_LOGGER.scanException(e, deploymentDir.getAbsolutePath());
+                } finally {
+                    releaseScanLock();
                 }
             }
         }
@@ -1309,10 +1337,8 @@ class FileSystemDeploymentService implements DeploymentScanner {
 
         @Override
         protected void handleFailureResult(final ModelNode result) {
-
             // Remove the in-progress marker
             removeInProgressMarker();
-
             writeFailedMarker(deploymentFile, result.get(FAILURE_DESCRIPTION).toString(), doDeployTimestamp);
         }
     }
@@ -1521,6 +1547,7 @@ class FileSystemDeploymentService implements DeploymentScanner {
     private static class ScanResult {
         private boolean scheduleRescan;
         private boolean requireUndeploy;
+        private List<ScannerTask> tasks;
     }
 
     /**
--- a/testsuite/manualmode/src/test/java/org/jboss/as/test/manualmode/deployment/DeploymentScannerOperationRollbackTestCase.java
+++ b/testsuite/manualmode/src/test/java/org/jboss/as/test/manualmode/deployment/DeploymentScannerOperationRollbackTestCase.java
@@ -1,0 +1,121 @@
+/*
+ * JBoss, Home of Professional Open Source
+ * Copyright 2014, JBoss Inc., and individual contributors as indicated
+ * by the @authors tag.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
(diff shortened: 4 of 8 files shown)


ehsavoie:
retest this please

deployment-scanner/src/main/java/org/jboss/as/server/deployment/scanner/DeploymentScannerDefinition.java
```diff
@@ -37,12 +37,13 @@
  * @created 25.1.12 17:24
  */
 public class DeploymentScannerDefinition extends SimpleResourceDefinition {
-
+    private final PathManager pathManager;
```
jamezp: It looks like we're not using this so we should remove it.

deployment-scanner/src/main/java/org/jboss/as/server/deployment/scanner/FileSystemDeploymentService.java
```diff
@@ -147,6 +148,13 @@
     private final String relativePath;
     private final PropertyChangeListener propertyChangeListener;
 
+    void rollBack(List tasks, ModelNode failure) {
```
jamezp: This should be `List<ScannerTask>`.

deployment-scanner/src/main/java/org/jboss/as/server/deployment/scanner/FileSystemDeploymentService.java
```diff
@@ -367,7 +379,17 @@ void bootTimeScan(final DeploymentOperations deploymentOperations) {
         }
     }
 
-    /** Perform a normal scan */
+    /**
+     * Perform a one-off scan to establish deployment tasks
+     */
+    Future<List<?>> singleScan() {
```
jamezp: Should return `List<ScannerTask>`.

deployment-scanner/src/main/java/org/jboss/as/server/deployment/scanner/FileSystemDeploymentService.java
```diff
@@ -590,6 +612,27 @@ private void executeScannerTasks(List<ScannerTask> scannerTasks, DeploymentOpera
         }
     }
 
+    private class ManualScanCallable implements Callable<List<?>> {
+        @Override
+        public List<?> call() throws Exception {
```
jamezp: Should return `List<ScannerTask>`.

deployment-scanner/src/main/java/org/jboss/as/server/deployment/scanner/FileSystemDeploymentService.java
```diff
@@ -590,6 +612,27 @@ private void executeScannerTasks(List<ScannerTask> scannerTasks, DeploymentOpera
         }
     }
 
+    private class ManualScanCallable implements Callable<List<?>> {
```
jamezp: Use the typed generic rather than a wildcard; `Callable<List<ScannerTask>>`

deployment-scanner/src/main/java/org/jboss/as/server/deployment/scanner/FileSystemDeploymentService.java
```diff
@@ -590,6 +612,27 @@ private void executeScannerTasks(List<ScannerTask> scannerTasks, DeploymentOpera
...
+        @Override
+        public List<?> call() throws Exception {
+            DeploymentOperations operations = deploymentOperations;
+            if (operations == null) {
+                operations = deploymentOperationsFactory.create();
+            }
+            if (acquireScanLock()) {
+                try {
+                    DeploymentScannerLogger.ROOT_LOGGER.info("Manual scan launched");
```
jamezp: This should probably be a debug message. If not then it needs to be localized.

deployment-scanner/src/main/java/org/jboss/as/server/deployment/scanner/FileSystemDeploymentService.java
```diff
@@ -590,6 +612,27 @@ private void executeScannerTasks(List<ScannerTask> scannerTasks, DeploymentOpera
...
+                    return scan(true, operations).tasks;
+                } catch (Exception e) {
+                    ROOT_LOGGER.scanException(e, deploymentDir.getAbsolutePath());
+                } finally {
+                    releaseScanLock();
+                }
+            }
+            return Collections.EMPTY_LIST;
```
jamezp: Should be `return Collections.emptyList()`

testsuite/manualmode/src/test/java/org/jboss/as/test/manualmode/deployment/DeploymentScannerOperationRollbackTestCase.java
```diff
@@ -0,0 +1,248 @@
...
+        try {
+            client = TestSuiteEnvironment.getModelControllerClient();
+            try {
+                final File deploymentOne = new File(deployDir, "deployment-one.jar");
+                createDeployment(deploymentOne, "org.jboss.modules");
+                // Add a new de
+                addDeploymentScanner();
+                try {
+                    Thread.sleep(500);
+                    assertThat(exists(DEPLOYMENT_ONE), is(false));
```
jamezp: Any reason not to use `assertTrue(exists(DEPLOYMENT_ONE))`?

testsuite/manualmode/src/test/java/org/jboss/as/test/manualmode/deployment/DeploymentScannerOperationRollbackTestCase.java
```diff
@@ -0,0 +1,248 @@
...
+        container.start();
+        try {
+            client = TestSuiteEnvironment.getModelControllerClient();
+            try {
+                final File deploymentOne = new File(deployDir, "deployment-one.jar");
+                createDeployment(deploymentOne, "org.jboss.modules");
+                // Add a new de
+                addDeploymentScanner();
+                try {
+                    Thread.sleep(500);
```
jamezp: What are we waiting for here?

testsuite/manualmode/src/test/java/org/jboss/as/test/manualmode/deployment/DeploymentScannerOperationRollbackTestCase.java
```diff
@@ -0,0 +1,248 @@
...
+                    assertThat(exists(DEPLOYMENT_ONE), is(false));
+                    String[] files = deployDir.list();
+                    assertThat(files.length, is(1));
+                    assertThat(files[0], is("deployment-one.jar"));
+                    deploymentOne.delete();
+                } finally {
+                    removeDeploymentScanner();
+                }
+            } finally {
+                StreamUtils.safeClose(client);
```
jamezp: Not a huge deal, but we can always use try-with-resources here.

testsuite/manualmode/src/test/java/org/jboss/as/test/manualmode/deployment/DeploymentScannerOperationRollbackTestCase.java
```diff
@@ -0,0 +1,248 @@
...
+        builder.addStep(Operations.createWriteAttributeOperation(PathAddress.pathAddress(SYSTEM_PROPERTY, "my.property").toModelNode(), "value", "test"));
+        builder.addStep(Util.createEmptyOperation("run-scan", getTestDeploymentScannerResourcePath()));
+        builder.addStep(Operations.createReadResourceOperation(DEPLOYMENT_ONE.toModelNode(), true));
+        final ModelNode result = executeOperation(builder.build().getOperation());
+        assertThat(result.get(OUTCOME).asString(), is(FAILED));
+    }
+
+    private void addDeploymentScanner() throws Exception {
+        ModelNode addOp = Util.createAddOperation(PathAddress.pathAddress(PathElement.pathElement(EXTENSION, "org.jboss.as.deployment-scanner")));
+        ModelNode result = executeOperation(addOp);
```
jamezp: We might want to consider a composite operation here. Manual mode tests are already slow and any extra speed we can get out of them the better. In the `org.jboss.as.controller.client.helpers.Operations` there is a `CompositeOperationBuilder` that can assist with creating the composite operation.

testsuite/manualmode/src/test/java/org/jboss/as/test/manualmode/deployment/DeploymentScannerOperationRollbackTestCase.java
```diff
@@ -0,0 +1,248 @@
...
+        operation.get(NAME).set("status");
+
+        final ModelNode result = executeOperation(operation);
+        if (SUCCESS.equals(result.get(OUTCOME).asString())) {
+            return result.get(RESULT).asString();
+        }
+        return "failed";
+    }
+
+    protected boolean isRunning() throws IOException {
```
jamezp: I don't see this used anywhere so it should probably be removed.

testsuite/manualmode/src/test/java/org/jboss/as/test/manualmode/deployment/DeploymentScannerOperationTestCase.java
```diff
@@ -0,0 +1,253 @@
...
+        try {
+            client = TestSuiteEnvironment.getModelControllerClient();
+            try {
+
+                final File deploymentOne = new File(deployDir, "deployment-one.jar");
+                createDeployment(deploymentOne, "org.jboss.modules");
+                // Add a new de
+                addDeploymentScanner();
+                try {
+                    Thread.sleep(500);
```
jamezp: Why are we sleeping here?

testsuite/manualmode/src/test/java/org/jboss/as/test/manualmode/deployment/DeploymentScannerOperationTestCase.java
```diff
@@ -0,0 +1,253 @@
...
+    }
+
+    private void runScan() throws Exception {
+        ModelNode runScanOp = Util.createEmptyOperation("run-scan", getTestDeploymentScannerResourcePath());
+        ModelNode result = executeOperation(runScanOp);
+        assertEquals("Unexpected outcome of running the scanner: " + runScanOp, SUCCESS, result.get(OUTCOME).asString());
+    }
+
+    private void addDeploymentScanner() throws Exception {
+        ModelNode addOp = Util.createAddOperation(PathAddress.pathAddress(PathElement.pathElement(EXTENSION, "org.jboss.as.deployment-scanner")));
```
jamezp: Same comment about a composite op. Not a huge deal either way though.

testsuite/manualmode/src/test/java/org/jboss/as/test/manualmode/deployment/DeploymentScannerOperationTestCase.java
```diff
@@ -0,0 +1,253 @@
...
+        operation.get(NAME).set("status");
+
+        final ModelNode result = executeOperation(operation);
+        if (SUCCESS.equals(result.get(OUTCOME).asString())) {
+            return result.get(RESULT).asString();
+        }
+        return "failed";
+    }
+
+    protected boolean isRunning() throws IOException {
```
jamezp: I don't see this used anywhere so it should probably be removed.

ehsavoie: line 125 and 118

deployment-scanner/src/main/java/org/jboss/as/server/deployment/scanner/FileSystemDeploymentService.java
ehsavoie: ScannerTask is not part of the API, I don't want to expose it

testsuite/manualmode/src/test/java/org/jboss/as/test/manualmode/deployment/DeploymentScannerOperationRollbackTestCase.java
ehsavoie: Because it breaks ?

deployment-scanner/src/main/java/org/jboss/as/server/deployment/scanner/FileSystemDeploymentService.java
jamezp: Keeping it package private is better than using a wildcard lists IMO.

deployment-scanner/src/main/java/org/jboss/as/server/deployment/scanner/FileSystemDeploymentScanHandler.java
```diff
@@ -0,0 +1,89 @@
...
+            public void execute(OperationContext context, ModelNode operation) throws OperationFailedException {
+                final FileSystemDeploymentService scanner = getExistingScanner(context, operation);
+                if (scanner != null) {
+                    final Future<List<?>> tasks = scanner.singleScan();
+                    context.completeStep(new OperationContext.RollbackHandler() {
+
+                        @Override
+                        public void handleRollback(OperationContext context, ModelNode operation) {
+                            try {
+                                scanner.rollBack(tasks.get(), context.getResult());
```
bstansberry: I don't get why this doesn't deadlock, although your test case implies it doesn't. The tasks.get() call should block until ManualScanCallable completes, and it won't complete because this op is holding the exclusive lock.

Oh, I see why the test passes and doesn't deadlock. I'll comment there.

testsuite/manualmode/src/test/java/org/jboss/as/test/manualmode/deployment/DeploymentScannerOperationRollbackTestCase.java
```diff
@@ -0,0 +1,97 @@
...
+            container.stop();
+        }
+    }
+
+    private void runScan(ModelControllerClient client) throws Exception {
+        final Operations.CompositeOperationBuilder builder = Operations.CompositeOperationBuilder.create();
+        builder.addStep(Operations.createAddOperation(PathAddress.pathAddress(SYSTEM_PROPERTY, "my.property").toModelNode()));
+        builder.addStep(Operations.createWriteAttributeOperation(PathAddress.pathAddress(SYSTEM_PROPERTY, "my.property").toModelNode(), "value", "test"));
+        builder.addStep(Util.createEmptyOperation("run-scan", getTestDeploymentScannerResourcePath()));
+        builder.addStep(Operations.createReadResourceOperation(DEPLOYMENT_ONE.toModelNode(), true));
```
bstansberry: I believe the intent of this op is to trigger rollback, due to the resource not existing?

This doesn't work because it fails in Stage.MODEL and your run-scan stuff is implemented in Stage.RUNTIME. So the run-scan work never really happens and the deadlock I talk about doesn't occur. It would occur if the op failed in RUNTIME.

deployment-scanner/src/main/java/org/jboss/as/server/deployment/scanner/FileSystemDeploymentService.java
```diff
@@ -147,6 +149,13 @@
     private final String relativePath;
     private final PropertyChangeListener propertyChangeListener;
 
+    void rollBack(List tasks, ModelNode failure) {
+        List<ScannerTask> scannerTasks = (List<ScannerTask>) tasks;
+        for(ScannerTask task : scannerTasks) {
+            task.handleFailureResult(failure);
+        }
```
bstansberry: This isn't a rollback. It just puts unrelated failure info in the deployments dir. The ManualScanCallable is still running.

bstansberry:
@ehsavoie here's a possible way to abort the scan if the op rolls back. If there's a problem with this we should give up on trying to integrate the rollback with the scan in any fashion. Either way the scan results recorded by the scan should not be affected but what happens to the run-scan op.

```
diff --git a/deployment-scanner/src/main/java/org/jboss/as/server/deployment/scanner/FileSystemDeploymentScanHandler.java b/deployment-scanner/src/main/java/org/jboss/as/server/deployment/scanner/FileSystemDeploymentScanHandler.java
index 4da1d35..11706eb 100644
--- a/deployment-scanner/src/main/java/org/jboss/as/server/deployment/scanner/FileSystemDeploymentScanHandler.java
+++ b/deployment-scanner/src/main/java/org/jboss/as/server/deployment/scanner/FileSystemDeploymentScanHandler.java
@@ -57,17 +57,11 @@ public class FileSystemDeploymentScanHandler implements OperationStepHandler {
                 final FileSystemDeploymentService scanner = getExistingScanner(context, operation);
                 if (scanner != null) {
-                    final Future<List<?>> tasks = scanner.singleScan();
-                    context.completeStep(new OperationContext.RollbackHandler() {
-
+                    context.completeStep(new OperationContext.ResultHandler() {
                         @Override
-                        public void handleRollback(OperationContext context, ModelNode operation) {
-                            try {
-                                scanner.rollBack(tasks.get(), context.getResult());
-                            } catch (InterruptedException ex) {
-                                Thread.currentThread().interrupt();
-                            } catch (ExecutionException ex) {
-                                DeploymentScannerLogger.ROOT_LOGGER.error("Exception during rollback", ex);
+                        public void handleResult(OperationContext.ResultAction resultAction, OperationContext context, ModelNode operation) {
+                            if (resultAction == OperationContext.ResultAction.KEEP) {
+                                scanner.singleScan();
                             }
                         }
                     });
``` 

Basically, this moves the scanner.singleScan() call into the ResultHandler.

testsuite/manualmode/src/test/java/org/jboss/as/test/manualmode/deployment/DeploymentScannerOperationRollbackTestCase.java
```diff
@@ -0,0 +1,121 @@
...
+                addDeploymentScanner(client);
+                prepareRollback(client);
+                try {
+                    assertThat(exists(client, DEPLOYMENT_ONE), is(false));
+                    runFailingScan(client);
+                    // Wait until deployed ...
+                    long timeout = System.currentTimeMillis() + TimeoutUtil.adjust(30000);
+                    while (!exists(client, DEPLOYMENT_ONE) && System.currentTimeMillis() < timeout) {
+                        Thread.sleep(100);
+                    }
```
bstansberry: This will always take 30 secs to prove a scan didn't happen. I don't have any clever ideas for proving a negative, but how about 2 or 3 secs?

deployment-scanner/src/main/java/org/jboss/as/server/deployment/scanner/FileSystemDeploymentService.java
bstansberry: The singleScan method can be a void return now, so there's no issue re: ScannerTask becoming API.

deployment-scanner/src/main/java/org/jboss/as/server/deployment/scanner/FileSystemDeploymentService.java
```diff
@@ -1200,6 +1231,10 @@ protected void recordInProgress() {
 
         protected abstract void handleFailureResult(final ModelNode result);
 
+        protected void rollBack() {
+            System.out.println("Rollback");
+        }
+
```
bstansberry: This can be removed. Sorry I didn't notice it before.

deployment-scanner/src/main/java/org/jboss/as/server/deployment/scanner/FileSystemDeploymentService.java
bstansberry: If you change this to remove the rollBack() method from ScannerTask, might as well make this implement Callable<List<ScannerTask>>.

Actually it could just implement Runnable, as nothing is done with the return value from call(). (I think doing nothing is correct. I think all that could be done is to schedule a rescan if scanResult.scheduleRescan == true, but doing that is wrong IMHO. If the scanner isn't enabled for automatic scanning, it shouldn't do any scanning without an explicit command.)

PR #493, opened 2015-02-13
[WFLY-3528] Upgrade Xerces from 2.9.1-jbossas-2 to 2.11.0.

langfr:


bstansberry:
Presumably the failures are related.

kabir:
A bit of context for why the upgrade is needed would be good

asoldano:
Kabir, generally speaking, the reason why we should upgrade Xerces is basically that we're relying on a really old version, which has bugs, for instance the one mentioned in WFLY-3528.
This said, as Tomaz already explained in https://github.com/wildfly/wildfly/pull/7165 , we can't upgrade to vanilla 2.11.0, because we need a patched version that works in modular environment. The jira is currently assigned to Carlo for this reason (and others, EAP related).

wolfc:
https://issues.jboss.org/browse/WFLY-3528 which is apparently assigned to me.
We should be upgrading to https://repository.jboss.org/nexus/content/groups/public/xerces/xercesImpl/2.11.0.SP1/

wolfc:
Superseded by https://github.com/wildfly/wildfly-core/pull/522

PR #494, opened 2015-02-16
Deprecate threading subsystem & security module

ctomc:


dmlloyd:
That's a lot of failures.

ctomc:
@dmlloyd yeah, they are fixed by https://github.com/wildfly/wildfly/pull/7171

ctomc:
retest this please

bstansberry:
The PR failures on core look related to this.

scottmarlow:
@jamezp mentioned that he was able to run the TCK locally after removing the threading subsystem from his standalone.xml.  Did this bz 1190981 change make it into EAP 6.4 yet?

ctomc:
@scottmarlow threads subsystem was deprecated in EAP already.
In upstream it there is additional change that prevents threads subsystem to be used for anything else but managing legacy model (for mixed domain mode scenarios) similar as we have with web and jaxr subsystems.
Upstream code & configurations should not use threads subsystem anymore.
So proper fix for TCK is to remove threads subsystem from all its configs

PR #495, opened 2015-02-16
https://issues.jboss.org/browse/WFCORE-508 Remove need to maintain disco...

aloubyansky:
...very-option ordering separately from the option

This commit introduces a new attribute 'options' for discovery-options resource, which contains a LIST of various discovery options as described on the jira.
The resources discovery-option and static-discovery are still preserved in the tree, although they are PlaceholderResource's now, all the actual data is kept in the newly introduced attribute and all the resource manipulation will be translated into the corresponding manipulation of the attribute.
discovery-options attribute is also preserved for now, although it's a read-only attribute now and the read-attribute handler will pull the data from the options attribute.

Thanks,
Alexey


aloubyansky:
I've identified some issues with the description generation and validation. I'll re-open the PR when those are fixed.

PR #497, opened 2015-02-17
Management deprecation

bstansberry:
Deprecate a number of core management resources that may be converted to subsystems in WildFly 10.

Upstream for the equivalent EAP 6.4 change.


bstansberry:
Merging my own since it's already in EAP.

PR #500, opened 2015-02-18
[WFCORE-550] :  [WFCORE-550] : When testing with WildFly Full, the WSTestCase is randomly unstable.

ehsavoie:
WSTestCase is using HttpRequest that doesn't disconnect after executing HTTP requests this may lead to some strange behaviour as pre doc:
"Each HttpURLConnection instance is used to make a single request but the underlying network connection to the HTTP server may be transparently shared by other instances. Calling the close() methods on the InputStream or OutputStream of an HttpURLConnection after a request may free network resources associated with this instance but has no effect on any shared persistent connection. Calling the disconnect() method may close the underlying socket if a persistent connection is otherwise idle at that time."

Jira: https://issues.jboss.org/browse/WFCORE-550


bstansberry:
retest this please

ehsavoie:
So it doesn't fix the issue :(

PR #501, opened 2015-02-18
https://issues.jboss.org/browse/WFCORE-508 Remove need to maintain discovery-option ordering separately from the option

aloubyansky:
The first commit introduces a new attribute 'options' for discovery-options resource, which contains a LIST of various discovery options as described on the jira.
The resources discovery-option and static-discovery are still preserved in the tree, although they are PlaceholderResource's now, all the actual data is kept in the newly introduced attribute and all the resource manipulation will be translated into the corresponding manipulation of the attribute.
discovery-options attribute is also preserved for now, although it's a read-only attribute now and the read-attribute handler will pull the data from the options attribute.

The second commit fixes value type initialization for complex types with map attributes during the generation of descriptions.

Thanks,
Alexey


host-controller/src/main/java/org/jboss/as/host/controller/discovery/DiscoveryOptionsResource.java
```diff
@@ -0,0 +1,407 @@
...
+    private final Resource delegate;
+
+    public DiscoveryOptionsResource() {
+        this(Resource.Factory.create());
+    }
+
+    DiscoveryOptionsResource(Resource delegate) {
+        if (delegate == null) {
+            throw HostControllerLogger.ROOT_LOGGER.nullVar("delegate");
+        }
```
bstansberry: No need to change this; just an FYI.

We put in quite a few of these early in AS7 days, but they can be just an assert delegate != null; If this is null it's a programming error, not a mistake in user input.

host-controller/src/main/java/org/jboss/as/host/controller/operations/DiscoveryOptionAddHandler.java
```diff
@@ -65,7 +64,7 @@ protected boolean requiresRuntime(OperationContext context) {
     @Override
     protected void performRuntime(final OperationContext context, final ModelNode operation, final ModelNode model) throws OperationFailedException {
         if (context.isBooting()) {
-            populateHostControllerInfo(hostControllerInfo, context, operation, model);
+            populateHostControllerInfo(hostControllerInfo, context, operation, operation);
```
bstansberry: This would be clearer if the "model" param was just removed.

bstansberry:
FYI, I found this while looking at what this branch does:

https://issues.jboss.org/browse/WFCORE-565

Not something that needs to be fixed in this PR or by you; just an FYI.

host-controller/src/main/java/org/jboss/as/host/controller/operations/AbstractDiscoveryOptionAddHandler.java
```diff
@@ -43,22 +49,42 @@
...
         // this list to maintain the order
-        ModelNode discoveryOptionsOrdering = discoveryOptions.get(DISCOVERY_OPTIONS).clone();
-        if (!discoveryOptionsOrdering.isDefined()) {
-            discoveryOptionsOrdering.setEmptyList();
+        final ModelNode list = discoveryOptions.get(ModelDescriptionConstants.OPTIONS).clone();
+        if (!list.isDefined()) {
+            list.setEmptyList();
         }
-        discoveryOptionsOrdering.add(type, Util.getNameFromAddress(operation.get(OP_ADDR)));
+        list.add(type, node);
```
bstansberry: This produces a list of ModelType.PROPERTY instead of the specified list of ModelType.OBJECT

```
ModelNode element = new ModelNode();
element.get(type).set(node);
list.add(element);
```

will produce the expected structure.

host-controller/src/main/java/org/jboss/as/host/controller/discovery/DiscoveryOption.java
```diff
@@ -52,6 +52,7 @@
 
     /**
      *  Gets the host name of the remote domain controller.
+     *  This method may return null.
```
aloubyansky: I was not sure about this. It looks to me it's not expected to return null though.

bstansberry: It's messy. It can return null until the discover() method is invoked. I commented more on WFCORE-565.

I wanted to play with this yesterday so I could see what the read-resource and read-resource-description responses looked like, including with custom-discovery set up. I found a lot of problems, but not due to this patch.

https://github.com/bstansberry/wildfly-core/commits/WFCORE-565 had some quick changes just to get things so I could do my experiments.

aloubyansky:
retest this please

PR #504, opened 2015-02-18
Remove unused common-core dep from logging

jamezp:


jamezp:
Apparently logging does use it and I was scanning for the wrong package.

PR #508, opened 2015-02-19
Move model validation into controller

kabir:
https://issues.jboss.org/browse/WFCORE-543
Also fix the tests affected by the improved validation. I pushed a few fixes to wildfly already, so hopefully this should pass there as well.

### DIFF
--- a/controller/src/main/java/org/jboss/as/controller/ModelControllerImpl.java
+++ b/controller/src/main/java/org/jboss/as/controller/ModelControllerImpl.java
@@ -342,7 +342,7 @@ class ModelControllerImpl implements ModelController {
             final OperationContextImpl context = new OperationContextImpl(operationID, operation.get(OP).asString(),
                     operation.get(OP_ADDR), this, processType, runningModeControl.getRunningMode(),
                     contextFlags, handler, attachments, managementModel.get(), originalResultTxControl, processState, auditLogger,
-                    bootingFlag.get(), hostServerGroupTracker, blockingTimeoutConfig, accessMechanism, notificationSupport);
+                    bootingFlag.get(), hostServerGroupTracker, blockingTimeoutConfig, accessMechanism, notificationSupport, false);
             // Try again if the operation-id is already taken
             if(activeOperations.putIfAbsent(operationID, context) == null) {
                 CurrentOperationIdHolder.setCurrentOperationID(operationID);
@@ -400,17 +400,19 @@ class ModelControllerImpl implements ModelController {
     }
 
     boolean boot(final List<ModelNode> bootList, final OperationMessageHandler handler, final OperationTransactionControl control,
-              final boolean rollbackOnRuntimeFailure, MutableRootResourceRegistrationProvider parallelBootRootResourceRegistrationProvider) {
+              final boolean rollbackOnRuntimeFailure, MutableRootResourceRegistrationProvider parallelBootRootResourceRegistrationProvider, boolean skipModelValidation) {
 
         final Integer operationID = random.nextInt();
 
         EnumSet<OperationContextImpl.ContextFlag> contextFlags = rollbackOnRuntimeFailure
                 ? EnumSet.of(AbstractOperationContext.ContextFlag.ROLLBACK_ON_FAIL)
                 : EnumSet.noneOf(OperationContextImpl.ContextFlag.class);
+
+        //For the initial operations the model will not be complete, so defer the validation
         final AbstractOperationContext context = new OperationContextImpl(operationID, INITIAL_BOOT_OPERATION, EMPTY_ADDRESS,
                 this, processType, runningModeControl.getRunningMode(),
                 contextFlags, handler, null, managementModel.get(), control, processState, auditLogger, bootingFlag.get(),
-                hostServerGroupTracker, null, null, notificationSupport);
+                hostServerGroupTracker, null, null, notificationSupport, true);
 
         // Add to the context all ops prior to the first ExtensionAddHandler as well as all ExtensionAddHandlers; save the rest.
         // This gets extensions registered before proceeding to other ops that count on these registrations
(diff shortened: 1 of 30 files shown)


controller/src/main/java/org/jboss/as/controller/OperationContextImpl.java
```diff
@@ -1687,6 +1688,11 @@ ManagementModel getManagementModel() {
         return managementModel;
     }
 
+    @Override
+    boolean isSkipModelValidation() {
+        return false;
```
ctomc: is this ok? should be this somewhat context dependent?

controller/src/main/java/org/jboss/as/controller/ValidateModelStepHandler.java
```diff
@@ -113,5 +161,22 @@ boolean hasAlternative(final String[] alternatives, ModelNode operationObject) {
         return false;
     }
 
+    private Resource loadResource(OperationContext context) {
+        final PathAddress address = context.getCurrentAddress();
+        PathAddress current = PathAddress.EMPTY_ADDRESS;
+        Resource resource = context.readResourceFromRoot(PathAddress.EMPTY_ADDRESS, false);
+        for (PathElement element : address) {
```
ctomc: what is the point of this? try to find last resource that is still valid as part of path address?
What does that give us?

core-model-test/framework/src/main/java/org/jboss/as/core/model/test/TestModelControllerService.java
```diff
@@ -442,6 +488,11 @@ public void initCoreModel(Resource rootResource, ManagementResourceRegistration
         final HostFileRepository hostFileRepository = createHostFileRepository();
         final DomainController domainController = createDomainController(env, info);
 
+        public HostInitializer() {
+            // FIXME HostInitializer constructor
```
ctomc: forgot something?

controller/src/main/java/org/jboss/as/controller/ValidateModelStepHandler.java
kabir: To be more forgiving if a resource disappears for whatever reason. The thing that led me to this was finding that a composite to
    add a=a
    remove a=b
would cause an exception failing validation. Also exceptions are very expensive to create, so I'm just playing safe.

controller/src/main/java/org/jboss/as/controller/OperationContextImpl.java
kabir: Fixing, thanks!

kabir:
I need to look into the full integration failures

kabir:
Unrelated failure

controller/src/main/java/org/jboss/as/controller/ModelControllerImpl.java
```diff
@@ -449,6 +450,19 @@ boolean boot(final List<ModelNode> bootList, final OperationMessageHandler handl
...
+                Set<PathAddress> validateAddresses = new HashSet<PathAddress>();
+                validateAddresses.addAll(context.getModifiedResourcesForModelValidation());
+                validateAddresses.addAll(postExtContext.getModifiedResourcesForModelValidation());
+                final AbstractOperationContext validateContext = new OperationContextImpl(operationID, POST_EXTENSION_BOOT_OPERATION,
+                        EMPTY_ADDRESS, this, processType, runningModeControl.getRunningMode(),
+                        contextFlags, handler, null, managementModel.get(), control, processState, auditLogger,
+                                bootingFlag.get(), hostServerGroupTracker, null, null, notificationSupport, false);
+                resultAction = validateContext.executeOperation();
+
+            }
```
bstansberry: I don't get this part. It doesn't add any steps so I guess it's just supposed to validate. But the 'validateAddresses' set isn't used.

controller/src/main/java/org/jboss/as/controller/ParallelBootOperationContext.java
```diff
@@ -434,4 +436,14 @@ Resource getModel() {
...
     }
+
+    @Override
+    boolean isSkipModelValidation() {
+        return primaryContext.isSkipModelValidation();
+    }
+
+    @Override
+    void addModelValidationSteps() {
+        primaryContext.addModelValidationSteps();
```
bstansberry: This will result in the steps getting added multiple times, as each subsystem runs, as the primary context is maintaining a global modifiedResourcesForModelValidation set.

It probably isn't causing problems because it's not getting called during boot, since skipModelValidation is true. But it's fragile. If it shouldn't ever be called perhaps its better to have it throw UnsupportedOperationException.

controller/src/main/java/org/jboss/as/controller/ReadOnlyContext.java
```diff
@@ -367,4 +369,13 @@ public void deregisterCapability(String capability) {
...
     }
 
+    @Override
+    boolean isSkipModelValidation() {
+        return primaryContext.isSkipModelValidation();
+    }
+
+    @Override
+    void addModelValidationSteps() {
+        primaryContext.addModelValidationSteps();
```
bstansberry: This is a read-only context and should not do work that is associated with writes. Probably skipModelValidation should be true.

controller/src/main/java/org/jboss/as/controller/ValidateModelStepHandler.java
```diff
@@ -38,33 +44,48 @@ private ValidateModelStepHandler() {
...
+                                throw ControllerLogger.ROOT_LOGGER.requiredAttributeNotSet(required, attr.getName());
+                            }});
                     }
                 }
             }
 
             if (!isAllowed(attr, model)) {
+                //TODO should really use attemptReadMissingAttributeValueFromHandler() to make this totally good, but the
+                //overhead might be bigger than is worth at the moment since we would have to invoke the extra steps for
+                //every single attribute not found (and not found should be the normal).
```
bstansberry: I agree with the thinking in the comment.

This isn't an issue with this PR, but it seems that isAllowed and this block guarded by the call to isAllowed are doing redundant work. There could be a single loop to create the StringBuilder, and fail if sb != null.

controller/src/main/java/org/jboss/as/controller/ModelControllerImpl.java
bstansberry: I think the intent was to call addModifiedResourcesForModelValidation.

controller/src/main/java/org/jboss/as/controller/ValidateModelStepHandler.java
bstansberry: Is this still needed now that you are using notifications to populate modifiedResourcesForModelValidation?

controller/src/main/java/org/jboss/as/controller/ModelControllerImpl.java
kabir: ooops :-(

kabir:
@bstansberry I've updated to make ParallelBoot- and ReadOnlyContext skip validation. Rather than try to gather all the modified resources from the various parts of the boot, I now simply iterate the model as input for the additional validation step.

PR #510, opened 2015-02-19
[WFCORE-560] Don't reassign 'model' until its children are read

bstansberry:


bstansberry:
Closing, as I don't really understand what this code is doing. I don't think it's right, but I don't know if this PR's change just makes things worse.

The comment above this block says "// Allow wildcard navigation for the last element" so it seems the intent is to only deal with the final PathElement in the address, but the code isn't checking that.

PR #511, opened 2015-02-20
fix parsing of attribute groups

jmesnil:


controller/src/test/java/org/jboss/as/controller/persistence/PersistentResourceXMLParserTestCase.java
```diff
@@ -0,0 +1,209 @@
...
+                    )
+                    .build();
+        }
+
+        @Override
+        public void readElement(XMLExtendedStreamReader reader, List<ModelNode> list) throws XMLStreamException {
+            xmlDescription.parse(reader, PathAddress.EMPTY_ADDRESS, list);
+        }
+
+        @Override
```
ctomc: no need for this anymore, you should rather extend PersistentResourceXMLParser, where read/write Element methods are hidden.

ctomc: take a look at IOSubsystemParser_1_1 for example

jmesnil: using `PersistentResourceXMLParser` requires to create a `SubsystemMarshallingContext` that itself requires a `XMLExtendedStreamWriter` which can not be created in a simple way using the staxmapper API.

For the purpose of this tests, implementing directly `XMLElementReader` and `XMLElementReader` should be sufficient.

bstansberry:
Thanks!

PR #515, opened 2015-02-23
 [WFCORE-293] : Add EAP 6.2 and 6.3 to versions tested in core-model-test.

ehsavoie:
Updating wildfly-legacy-test version.
Updating transformers for new versions.
Fixing some legacy transformers issues

Jira: https://issues.jboss.org/browse/WFCORE-293


bstansberry:
I had to revert this as the 1.0.0.Alpha5 release was broken:

https://github.com/wildfly/wildfly-legacy-test/blob/1.0.0.Alpha5/spi/interfaces/src/main/java/org/wildfly/legacy/test/spi/Version.java#L29

I tried to whip together an Alpha6 to fix it, but it won't build, as it seems to need a non-existent WildFly core Alpha19.

PR #516, opened 2015-02-23
WELD-1853 Add path com/sun/org/apache/bcel/internal/classfile to sun-jdk system dependency

mkouba:
To improve some error messages Weld attempts to get the line number
associated with a particular class member. Reflection API does not
expose such an info so Weld is analysing the bytecode using the
Apache BCEL inluded in Oracle JDK and OpenJDK.


bstansberry:
What happens on the IVM JDK?

jharting:
(02:39:51 PM) dmlloyd: if the package isn't there in a system dep, it is treated as empty
(02:40:01 PM) dmlloyd: so it won't fail, the module will just have no content

bstansberry:
Sure. I meant what does Weld do when these classes are not available.

My assumption is it handles it gracefully but I figure I should ask.

jharting:
Ah, I see. Yes, Weld uses those classes optionally for better error reporting (member line number in the source file). If those classes are not available, simply less information is given in the error report.

PR #518, opened 2015-02-23
Revert "Use wildfly-legacy-test 1.0.0.Alpha5"

ctomc:
This breaks the build


bstansberry:
The merge history was weird here. This commit was merged after Emmanuel's, and the merge didn't really change anything as Emmanuel's did the upgrade as well.

#519 restores things.

Thanks for figuring out where this Alpha4-SNAPSHOT came from!

PR #523, opened 2015-02-24
WFCORE-552 SaslResource and SaslPolicyResource don't deal with the paren...

ctomc:
...t resource correctly


bstansberry:
Looks good but needs a rebase. Weird; I don't recall anything in this area recently.

PR #528, opened 2015-02-27
https://issues.jboss.org/browse/WFCORE-554 Unintuitive CLI behavior with...

aloubyansky:
... 'rbac' enabled and no security-realm configured on mangement interface

@bstansberry please, review whether that's good enough.

Thanks,
Alexey


bstansberry:
Replaced by #531 which is just this plus a minor edit to the failure message. #531 is merged.

PR #536, opened 2015-03-03
Minor javadoc cleanup.

darranl:


controller/src/main/java/org/jboss/as/controller/registry/Resource.java
```diff
@@ -280,6 +291,8 @@ static ModelNode readModel(final PathAddress address, final Resource resource, f
          * @param address the address the address relative to {@code resource}'s address. Cannot be {@code null}
          * @return the resource the descendant resource. Will not be {@code null}
          * @throws java.util.NoSuchElementException if there is no descendant resource at {@code address}
+         *
+         * @see NoSuchResourceException
```
bstansberry: This PR needs a rebase; probably because of this. It probably conflicts with https://github.com/wildfly/wildfly-core/pull/537 which should have removed the need for this @see tag.

darranl:
Am going to close this PR, it was just intended to be a quick change to get some javadoc comments added.

PR #524, opened 2015-02-24
[WFCORE-542] Added a new utility to work with registering resources and ...

jamezp:
...retrieving models from a deployment resource. Deprecated the previous methods on the DeploymentUtil.

Added a new API to assist with registering and creating resources for a deployments subsystem model. The new API was added as an attachment to the `DeploymentUnit` and the old `DeploymentUnit` methods were deprecated.

Deprecated the `DeploymentModelUtils` as there was a `TODO` to make it package private. Moved the needed methods within the package to the `DeploymentResourceSupport`.

PR #540, opened 2015-03-05
[WFCORE-542] Add ability to check for already registered resources.

jamezp:
An addition to #524 to allow deployments to check whether or not resources have been registered to the deployments subsystem.

### DIFF
--- a/server/src/main/java/org/jboss/as/server/deployment/DeploymentResourceSupport.java
+++ b/server/src/main/java/org/jboss/as/server/deployment/DeploymentResourceSupport.java
@@ -50,6 +50,19 @@ public final class DeploymentResourceSupport {
     }
 
     /**
+     * Checks to see if a subsystem resource has already been registered for the deployment.
+     *
+     * @param subsystemName the name of the subsystem
+     *
+     * @return {@code true} if the subsystem exists on the deployment otherwise {@code false}
+     */
+    public boolean hasDeploymentSubsystemModel(final String subsystemName) {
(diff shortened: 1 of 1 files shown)


server/src/main/java/org/jboss/as/server/deployment/DeploymentResourceSupport.java
```diff
@@ -82,6 +82,47 @@ public ModelNode registerDeploymentSubsystemResource(final String subsystemName,
...
+     *
+     * @param subsystemName the name of the subsystem
+     * @param address       the address to check
+     *
+     * @return {@code true} if the address exists on the subsystem otherwise {@code false}
+     */
+    public boolean hasDeploymentSubModel(final String subsystemName, final PathElement address) {
+        final Resource root = deploymentUnit.getAttachment(DEPLOYMENT_RESOURCE);
+        final PathElement subsystem = PathElement.pathElement(SUBSYSTEM, subsystemName);
+        return root.hasChild(subsystem) && root.getChild(subsystem).hasChild(address);
```
bstansberry: Should this support a null address? Same with the other one, which make support a null or empty address.

jamezp: Probably actually. Since you can register subsystem resources there should be a check for that as well.

bstansberry:
@jperkins go ahead and merge this if the run against full looks ok. (Seems fine so far.)

PR #543, opened 2015-03-06
WFCORE-564 Fix PersistentResourceXMLDescription parser

ctomc:
- it now properly supports mixing attribute groups & wrappers
- add support for multiple resources inside wrapper element
- add option to override also marshaller inside parser definition
- add tests to make sure this works properly

rework internals to not use map<attribute-name, attribute> to properly support multiple groups with same xml names for different attributes

### DIFF
--- a/controller/src/test/java/org/jboss/as/controller/persistence/PersistentResourceXMLParserTestCase.java
+++ b/controller/src/test/java/org/jboss/as/controller/persistence/PersistentResourceXMLParserTestCase.java
@@ -26,73 +26,97 @@ import static org.jboss.as.controller.PersistentResourceXMLDescription.builder;
 import static org.jboss.as.controller.SimpleAttributeDefinitionBuilder.create;
 import static org.junit.Assert.assertEquals;
 
+import java.io.BufferedReader;
+import java.io.IOException;
+import java.io.InputStreamReader;
 import java.io.StringReader;
 import java.io.StringWriter;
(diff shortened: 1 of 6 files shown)


ctomc:
retest this please

ctomc:
unreleated failure.

ctomc:
retest this please

controller/src/test/java/org/jboss/as/controller/persistence/PersistentResourceXMLParserTestCase.java
```diff
@@ -26,73 +26,131 @@
...
+            while ((line = reader.readLine()) != null) {
+                writer.write(line + "\n");
+            }
+        }
+        return writer.toString();
+    }
+
+    public static XMLExtendedStreamWriter createXMLStreamWriter(XMLStreamWriter writer) throws Exception {
+        // Use reflection to access package protected class FormattingXMLStreamWriter
+        // TODO: at some point the staxmapper API could be enhanced to make this unnecessary
```
bstansberry: This shouldn't be necessary any more. But it's fine to clean this up in a separate PR, as there's other code like this.

ctomc: Yeah I know, I have follow up work that contains fix for this, but wanted to get this merged first :)

PR #544, opened 2015-03-06
https://issues.jboss.org/browse/WFCORE-582 CLI: ability to load command ...

aloubyansky:
...handlers from the available in the model extensions

This commit is trying to load CLI command handlers using Java service loader mechanism from the modules of the management model extensions when the CLI (re-)connects to the controller.
It also includes a simple test.

Thanks,
Alexey

### DIFF
--- a/cli/src/main/java/org/jboss/as/cli/CommandRegistry.java
+++ b/cli/src/main/java/org/jboss/as/cli/CommandRegistry.java
@@ -21,8 +21,11 @@
  */
 package org.jboss.as.cli;
 
+import java.util.ArrayList;
+import java.util.Collections;
 import java.util.HashMap;
 import java.util.HashSet;
+import java.util.List;
 import java.util.Map;
 import java.util.Set;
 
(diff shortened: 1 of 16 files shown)


aloubyansky:
retest this please

testsuite/standalone/src/test/java/org/jboss/as/test/integration/management/cli/extensions/CliExtCommandsTestCase.java
```diff
@@ -0,0 +1,150 @@
...
+import org.wildfly.core.testrunner.WildflyTestRunner;
+import org.xnio.IoUtils;
+
+/**
+ * Testing commands loaded from the available management model extensions.
+ *
+ * @author Alexey Loubyansky
+ */
+@RunWith(WildflyTestRunner.class)
+@ServerControl(manual = true)
```
bstansberry: You need to move this to testsuite/manualmode.

PR #545, opened 2015-03-06
[WFCORE-584] Support embedding a standalone server in a CLI process

bstansberry:
This is a follow-up to #479, with quite a lot of tests added, issues revealed by those tests fixed, and some changes to the embedded module. I feel it's quite solid now.

See https://developer.jboss.org/docs/DOC-53050 for background.

A couple significant points that are different from what I had before:

1) I am treating the "embedded" module I've added to core as being there to support this usage, not as general purpose embedding solution. Therefore, it uses a different maven GAV from the embedded module that's in full, the classes have been repackaged, and the StandaloneServer API has been changed to remove methods unrelated to this use case.

This use case needs an embedding solution, and this module works for this use case. But there's a fair amount of work going on related to embedding WildFly, and I see little benefit of turning this into a general purpose API. If we develop something better, this use case can just switch to it; it's needs are fairly simple.

2) When the embedding app is non-modular, the org.jboss.logging and org.jboss.logmanager packages are configured as system packages in the embedded server's modular classloading environment. The org.jboss.logmanager.LogManager class is installed as the JDK log manager, and it uses its LogContext class and the static fields therein heavily. If the CLI-side LogContext class and the modular side LogContext class are different, nothing works properly. Since logmanager is a system pkg, the org.jboss.logging API stuff must be too.

3) The CLI configures the StdioContextSelector and the LogContextSelector (using a ThreadLocal) before making any call into the embedded server. This ensures that the System.out handling is appropriate when the embedded server writes to the console log. This is currently done on the CLI side. It could potentially be moved to the embedded server side. Whether or not that is appropriate is one reason for 1) above. The CLI side knows what setup it wants; if it were moved to the embedded server side, there would need to be some mechanism for the CLI side to indicate what the desired settings are.

The test coverage I added is for the case where the embedding app is non-modular. That's because a JUnit test is naturally non-modular. It would be nice to set up some sort of test fixture where the test fixture is modular so regression tests for that case (equivalent to what users will see using jboss-cli.sh) can be added. But that's a fairly big undertaking. Manual tests show the behavior is as expected.


logging/src/main/java/org/jboss/as/logging/LoggingExtension.java
```diff
@@ -71,7 +71,9 @@
 
     static final PathElement LOGGING_PROFILE_PATH = PathElement.pathElement(CommonAttributes.LOGGING_PROFILE);
 
-    static final WildFlyLogContextSelector CONTEXT_SELECTOR = WildFlyLogContextSelector.Factory.create();
+    /** @deprecated only present for use by an integration test hack */
+    @Deprecated
+    static WildFlyLogContextSelector CONTEXT_SELECTOR = WildFlyLogContextSelector.Factory.create();
```
jamezp: Just an FYI I filed a JIRA to remove the hacky test that uses this https://issues.jboss.org/browse/WFLY-4410

logging/src/main/java/org/jboss/as/logging/logmanager/WildFlyLogContextSelectorImpl.java
```diff
@@ -41,7 +42,15 @@
...
-        contextSelector = new ClassLoaderLogContextSelector(true);
+        // Use the current log context as the default, not LogContext.DEFAULT_LOG_CONTEXT_SELECTOR
+        // This allows embedding use cases to control the log context
+        final LogContext defaultLogContext = LogContext.getLogContext();
+        contextSelector = new ClassLoaderLogContextSelector(new LogContextSelector() {
+            @Override
+            public LogContext getLogContext() {
+                return defaultLogContext;
+            }
+        }, true);
```
jamezp: Is there a reason to wrap this in a new `LogContextSelector` rather than just using `new ClassLoaderLogContextSelector(defaultLogContext, true)`? It doesn't really matter all that much really.

jamezp:
Just kind of a general note and nothing to hold the PR up. With the new `InteractionPolicy` we might need to look at how/when the logging extension persists the `logging.properties` file.

logging/src/main/java/org/jboss/as/logging/logmanager/WildFlyLogContextSelectorImpl.java
bstansberry: I don't see that constructor variant.

bstansberry:
@jamezp 
Re: "Just kind of a general note and nothing to hold the PR up. With the new InteractionPolicy we might need to look at how/when the logging extension persists the logging.properties file."

Do you mean for InteractionPolicy.READ_ONLY?

If so, that's a fair point. It's not really a new thing from this PR though -- the read only behavior has been there for a while. I just changed the way the state related to it was tracked to make it a bit easier to track the other flavors I added here.

logging/src/main/java/org/jboss/as/logging/logmanager/WildFlyLogContextSelectorImpl.java
jamezp: Sorry you're right, I'm being thick. I'll blame it one being early :)

jamezp:
@bstansberry Yes the `READ_ONLY` thing. It should be happening already, I should just review it to make sure it's still working. I guess the comment is for me on your PR and doesn't have a lot to do with this PR other than I thought abut it :)

cli/src/main/java/org/jboss/as/cli/embedded/EmbedServerHandler.java
```diff
@@ -0,0 +1,316 @@
...
+            }
+        }
+    }
+
+    private void configureLogContext(LogContext embeddedLogContext, File jbossHome, CommandContext ctx) {
+        File standaloneDir =  new File(jbossHome, "standalone");
+        File configDir =  new File(standaloneDir, "configuration");
+        File logDir =  new File(standaloneDir, "log");
+        File bootLog = new File(logDir, "boot.log");
+        File loggingProperties = new File(configDir, "logging.properties");
```
jamezp: Thinking out loud here; I wonder if we should create a `logging.properties` we could package in the JAR without a console handler. I suppose though the logging subsystem might just add it back.

cli/src/main/java/org/jboss/as/cli/impl/CLIPrintStream.java
```diff
@@ -0,0 +1,435 @@
...
+        }
+        try {
+            entered.set(Boolean.TRUE);
+            delegate.print(obj);
+        } finally {
+            entered.remove();
+        }
+    }
+
+    public void println() {
```
jamezp: Do we need the entered check here and the boolean variant below?

bstansberry: Yes; thanks. I'll fix  that.

cli/src/main/java/org/jboss/as/cli/embedded/EmbedServerHandler.java
bstansberry: My thinking is the behavior should be the same as if the user had started the server normally -- whatever is in logging.properties is used until the subsystem runs. The deviation from that is just whatever setting the user uses for the embed-server --std-out=??? param. That control over stdout is the only aspect that is distinctive for the embedded case.

jamezp: Makes sense. I didn't notice the --std-out stuff until after I made the comment :)

core-feature-pack/src/main/resources/modules/system/layers/base/org/wildfly/embedded/main/module.xml
```diff
@@ -0,0 +1,54 @@
...
+        <module name="org.jboss.as.controller-client"/>
+        <!--module name="org.jboss.as.naming"/-->
+        <module name="org.jboss.jandex"/>
+        <module name="org.jboss.logging"/>
+        <module name="org.jboss.modules"/>
+        <module name="org.jboss.staxmapper"/>
+        <module name="org.jboss.stdio"/>
+        <module name="org.jboss.vfs"/>
+        <module name="org.jboss.as.protocol"/>
+    </dependencies>
```
jamezp: There should probably be a dependency on `org.jboss.logmanager` as well. Possibly optional as it's optional in the pom.

embedded/src/main/java/org/wildfly/core/embedded/EmbeddedStandAloneServerFactory.java
```diff
@@ -0,0 +1,417 @@
...
+            return null;
+        }
+
+        try {
+            File root = new File(tempRoot);
+            if (!root.exists()) {
+                //Attempt to try to create the directory, in case something like target/embedded was specified
+                Files.createDirectories(root.toPath());
+            }
+            validateDirectory("jboss.test.clean.root", root);
```
jamezp: This probably just came over from WildFly, but the property doesn't seem right.

embedded/src/main/java/org/wildfly/core/embedded/EmbeddedStandAloneServerFactory.java
```diff
@@ -0,0 +1,417 @@
...
+            root = new File(root, "configs");
+            Files.createDirectories(root.toPath());
+            SimpleDateFormat format = new SimpleDateFormat("yyyyMMddHHmmssSSS");
+            root = new File(root, format.format(new Date()));
+            Files.createDirectory(root.toPath());
```
jamezp: We could just use `Files.createTempDirectory(root, "wildfly-embedded")` or something similar rather than using a date.

cli/src/main/java/org/jboss/as/cli/impl/CLIPrintStream.java
bstansberry: This is fixed.

jamezp:
This looks good to me.

core-feature-pack/src/main/resources/modules/system/layers/base/org/wildfly/embedded/main/module.xml
jamezp: I'm wrong here, wildfly-embedded only seems to use a string constant so the module dependency is not needed.

embedded/src/main/java/org/wildfly/core/embedded/EmbeddedStandAloneServerFactory.java
bstansberry: Timestamps are kind of nice though.

jamezp: Yeah, doesn't really matter all that much. I just cringe a little when I see `SimpleDateFormat`, but really it's not like it will be used a lot here so it's fine :)

PR #546, opened 2015-03-09
[WFCORE-587] Wildfly does not log stack trace of an exception

TomasHofman:
Issue: https://issues.jboss.org/browse/WFCORE-587

### DIFF
--- a/controller/src/main/java/org/jboss/as/controller/AbstractOperationContext.java
+++ b/controller/src/main/java/org/jboss/as/controller/AbstractOperationContext.java
@@ -1364,8 +1364,10 @@ abstract class AbstractOperationContext implements OperationContext {
                     }
                 }
             } catch (Exception e) {
-                report(MessageSeverity.ERROR,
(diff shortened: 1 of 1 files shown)


bstansberry:
This doesn't fix the problem (the "report" method doesn't log anything) and it unnecessarily rearranges imports to not follow the WildFly standard.

dmlloyd:
This fix is completely wrong anyway.  If you want the stack trace to be logged, just add ``@Cause`` to tge exception parameter and drop the corresponding part from the format string.

bstansberry:
I don't want the stack trace in the output to report(...) though. That's an out-of-band message theoretically back to the client, which currently goes nowhere. I don't want to be sending server side stack traces to the client.

There just needs to be a new log method to log this.

TomasHofman:
OK, I will direct it into log instead. Thanks for comments!

TomasHofman:
So I left the original report message without change but also logged it with a stack trace.

PR #551, opened 2015-03-10
[WFCORE-590]  SSL secured communication between master and slave in domain mode test cases added

OndrejKotek:


testsuite/domain/src/test/java/org/jboss/as/test/integration/domain/AbstractSSLMasterSlaveTestCase.java
```diff
@@ -0,0 +1,188 @@
...
+
+            HttpEntity entity = response.getEntity();
+            if (entity == null) {
+                return null;
+            }
+            responseNode = ModelNode.fromJSONStream(response.getEntity().getContent());
+            EntityUtils.consume(entity);
+        } catch (IOException ex) {
+            ex.printStackTrace();
+            return null;
```
bstansberry: Why?

testsuite/domain/src/test/java/org/jboss/as/test/integration/domain/SSLMasterSlaveTwoWayTestCase.java
```diff
@@ -0,0 +1,128 @@
...
+    @AfterClass
+    public static void tearDownDomain() throws Exception {
+        setOriginMasterManagementNativeInterface();
+        reloadMaster();
+        checkHostStatusOnMaster("master");
+        masterManagementRealmSetup.tearDown(domainMasterLifecycleUtil.getDomainClient());
+
+        testSupport.stop();
+        testSupport = null;
+        domainMasterLifecycleUtil = null;
```
bstansberry: These last 3 lines should be in a finally block. If there's a problem with this test, it must still shut down the processes or the rest of the testsuite will be screwed up.

bstansberry:
The two test classes are doing a lot of duplicative setup/teardown stuff that could be handled in static methods in the parent class.

testsuite/domain/src/test/java/org/jboss/as/test/integration/domain/AbstractSSLMasterSlaveTestCase.java
OndrejKotek: Why over HTTP? Because I do not know how to use native interface with SSL here.
Or, another why?

OndrejKotek:
Thank you very much for feedback. Next commit tries to provide fixes.

testsuite/domain/src/test/java/org/jboss/as/test/integration/domain/AbstractSSLMasterSlaveTestCase.java
bstansberry: Why dump the exception to System.err and return null?

testsuite/domain/src/test/java/org/jboss/as/test/integration/domain/AbstractSSLMasterSlaveTestCase.java
```diff
@@ -0,0 +1,201 @@
...
+        op.get(ADMIN_ONLY).set(false);
+        executeOverHttp(getMasterMgmtUri(), op.toJSONString(true));
+    }
+
+    protected static void checkHostStatusOnMaster(String host) throws Exception {
+        final URI mgmtURI = getMasterMgmtUri();
+        final String operation = getHostStateRunningOperationJson(host);
+        final long time = System.currentTimeMillis() + TIMEOUT;
+        do {
+            Thread.sleep(1000);
```
bstansberry: Please move the sleep below the "if (isHostStateRunning(mgmtURI, operation)) {" block and decrease the sleep time to 100ms at most. The testsuite should run fast.

OndrejKotek:
Thanks again for your feedback.

PR #553, opened 2015-03-10
WFCORE-85 Remove non socket-binding attributes from management interfaces

ctomc:


bstansberry:
retest this please

bstansberry:
This looks good, except it needs the xml aspects. The attributes should be gone from the latest xsd, and I think the parsing for legacy xsds should fail immediately with a useful message.

ctomc:
I was thinking more in lines of removing old parsers & XSDs completely and in newer parser fail. wdyt?

bstansberry:
I don't want to mix the issue of dropping old parsers (which I doubt we will do) with this work.

darranl:
The way the code is currently implemented dropping old parsers would be very complex, the best you could really do it not register the older parsers.

The task I am currently undertaking is to split out the code so that the next major version is independent of the previous parsers, this would technically make it easier to delete versions 1, 2, and 3 in a single sweep if we do get to the point we no longer want to support them.

ctomc:
@bstansberry PR is updated to remove xsd stuff and modify the parser. For the old parsers where we just ignore the values set for this attributes. I think it is better than failing.

server/src/main/java/org/jboss/as/server/parsing/StandaloneXml.java
```diff
@@ -528,16 +527,16 @@ private void parseHttpManagementInterface1_0(final XMLExtendedStreamReader reade
                 final Attribute attribute = Attribute.forName(reader.getAttributeLocalName(i));
                 switch (attribute) {
                     case INTERFACE: {
-                        HttpManagementResourceDefinition.INTERFACE.parseAndSetParameter(value, mgmtSocket, reader);
+                        //ignore
                         hasInterfaceName = true;
```
bstansberry: Shouldn't this fail?

AIUI, parse will pass, since hasInterfaceName=true, but presumably it will fail later due to no socket-binding setting.

The upshot of all this is a 1.0 config is likely useless, since there's no socket-binding alternative available. But we should fail with a descriptive failure message.

server/src/main/java/org/jboss/as/server/parsing/StandaloneXml.java
```diff
@@ -584,12 +583,12 @@ private void parseNativeManagementInterface1_0(final XMLExtendedStreamReader rea
                 final Attribute attribute = Attribute.forName(reader.getAttributeLocalName(i));
                 switch (attribute) {
                     case INTERFACE: {
-                        NativeManagementResourceDefinition.INTERFACE.parseAndSetParameter(value, mgmtSocket, reader);
+                        //ignore as we dont support it anymore
                         hasInterface = true;
```
bstansberry: Same as above.

server/src/main/java/org/jboss/as/server/parsing/StandaloneXml.java
```diff
@@ -897,12 +900,14 @@ private void parseNativeManagementSocket(XMLExtendedStreamReader reader, ModelNo
                 final Attribute attribute = Attribute.forName(reader.getAttributeLocalName(i));
                 switch (attribute) {
                     case INTERFACE: {
-                        NativeManagementResourceDefinition.INTERFACE.parseAndSetParameter(value, addOp, reader);
+                        //todo drop old parser completely
+                        //NativeManagementResourceDefinition.INTERFACE.parseAndSetParameter(value, addOp, reader);
                         hasInterface = true;
```
bstansberry: Same as above.

server/src/main/java/org/jboss/as/server/parsing/StandaloneXml.java
```diff
@@ -931,16 +936,16 @@ private void parseHttpManagementSocket(XMLExtendedStreamReader reader, ModelNode
                 final Attribute attribute = Attribute.forName(reader.getAttributeLocalName(i));
                 switch (attribute) {
                     case INTERFACE: {
-                        HttpManagementResourceDefinition.INTERFACE.parseAndSetParameter(value, addOp, reader);
+                        //ignore
                         hasInterface = true;
```
bstansberry: Same as above.

server/src/main/java/org/jboss/as/server/parsing/StandaloneXml.java
ctomc: That is the same debate we had with Darren on the subject, but agreed that no solution is perfect. 
If you want to go with failing instead of ignoring, i will fix it so.

bstansberry: Yes, please fail. Ignoring either means a failure later with a less useful error message, or somehow it doesn't fail but the server isn't manageable.

ctomc:
retest this please

bstansberry:
This needs to provide helpful exceptions from the parser if people use these attribtutes/elements. The standard "unexpectedAttribute/Element" isn't very helpful, since the xsds in question say these are ok.

ctomc:
@bstansberry rebased and updated with better error message

bstansberry:
retest this please

PR #559, opened 2015-03-12
[WFCORE-591] NPE in revertReloadRequired

jmesnil:
Prevent revertReloadRequired (resp. revertRestartReloadRequired) to
throw a NPE if reloadRequired (resp. restartRequired) as not been called
beforehands.

JIRA: https://issues.jboss.org/browse/WFCORE-591


bstansberry:
I'm tempted to say we should have asserts there, as this sounds like a programming error. But this is fine.

PR #560, opened 2015-03-12
Introduce org.jboss.as.controller.AbstractAttributeDefinitionBuilder#set...

rhusar:
...Name

Because if using SimpleAttributeDefinitionBuilder#create(org.jboss.as.controller.SimpleAttributeDefinition) you are never given a change to change the name.

### DIFF
--- a/controller/src/main/java/org/jboss/as/controller/AbstractAttributeDefinitionBuilder.java
+++ b/controller/src/main/java/org/jboss/as/controller/AbstractAttributeDefinitionBuilder.java
@@ -44,7 +44,7 @@ import org.jboss.dmr.ModelType;
 @SuppressWarnings("unchecked")
 public abstract class AbstractAttributeDefinitionBuilder<BUILDER extends AbstractAttributeDefinitionBuilder, ATTRIBUTE extends AttributeDefinition> {
 
-    protected final String name;
+    protected String name;
     protected ModelType type;
     protected String xmlName;
     protected boolean allowNull;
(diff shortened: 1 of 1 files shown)


controller/src/main/java/org/jboss/as/controller/AbstractAttributeDefinitionBuilder.java
```diff
@@ -139,6 +139,20 @@ public AbstractAttributeDefinitionBuilder(final String attributeName, final Attr
...
     /**
+     * Sets the {@link AttributeDefinition#getName() name} for the attribute, which is only needed
+     * if the attribute was created from an existing {@link SimpleAttributeDefinition} using
+     * {@link SimpleAttributeDefinitionBuilder#create(org.jboss.as.controller.SimpleAttributeDefinition)}
+     * method.
+     *
+     * @param name the attribute's name. {@code null} is not allowed
+     * @return a builder that can be used to continue building the attribute definition
+     */
+    public BUILDER setName(String name) {
```
bstansberry: assert name != null;

rhusar:
Actually, there is a well hidden contructor that allows for changing the name of the attribute:

    org.jboss.as.controller.SimpleAttributeDefinitionBuilder#SimpleAttributeDefinitionBuilder(java.lang.String, org.jboss.as.controller.SimpleAttributeDefinition)

So would it still be desirable to have org.jboss.as.controller.AbstractAttributeDefinitionBuilder#setName or should I close it? WDYT?

ctomc:
@rhusar it doesn't hurt to have it on builder as well

rhusar:
Alright, lets do it then.

PR #565, opened 2015-03-13
Allow list attributes to be marshalled as elements.

pferraro:
Other cleanup:
AttributeDefinition.marshallAsElement(...) should defer to the AttributeMarshaller to determine if the value can be marshalled as an element.
Fix return value of DefaultAttributeMarshaller.isMarshallableAsElement().

### DIFF
--- a/controller/src/main/java/org/jboss/as/controller/DefaultAttributeMarshaller.java
+++ b/controller/src/main/java/org/jboss/as/controller/DefaultAttributeMarshaller.java
@@ -32,13 +32,17 @@ import org.jboss.dmr.ModelNode;
  */
 public class DefaultAttributeMarshaller extends AttributeMarshaller {
 
-
+    @Override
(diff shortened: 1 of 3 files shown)


pferraro:
core-full integration failure is unrelated.

controller/src/main/java/org/jboss/as/controller/DefaultAttributeMarshaller.java
```diff
@@ -32,13 +32,17 @@
...
         if (isMarshallable(attribute, resourceModel, marshallDefault)) {
-            writer.writeAttribute(attribute.getXmlName(), resourceModel.get(attribute.getName()).asString());
+            writer.writeAttribute(attribute.getXmlName(), this.asString(resourceModel.get(attribute.getName())));
         }
     }
 
+    /**
+     * Use {@link #marshallAsElement(AttributeDefinition, ModelNode, boolean, XMLStreamWriter)} instead.
+     */
+    @Deprecated
```
bstansberry: I was going to object to deprecating; but you're right. It's not in the AttributeMarshaller interface. Good call.

PR #568, opened 2015-03-16
[WFCORE-599] Have the module command handler to check the jboss.home.dir...

jamezp:
... system property. This allows clients that use the CLI command API to not need the JBOSS_HOME environment variable set.

@aloubyansky could you have a look at this? I'm checking the system property first as a way to override the environment variable. Though I'd like your opinion on whether that order is correct or not.

### DIFF
--- a/cli/src/main/java/org/jboss/as/cli/handlers/module/ASModuleHandler.java
+++ b/cli/src/main/java/org/jboss/as/cli/handlers/module/ASModuleHandler.java
@@ -81,6 +81,7 @@ public class ASModuleHandler extends CommandHandlerWithHelp {
     }
 
     private static final String JBOSS_HOME = "JBOSS_HOME";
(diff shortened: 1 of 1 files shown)


aloubyansky:
I don't have a strong opinion here. Especially with this specific command, which we don't want to improve significantly. I won't mind this change.

For reference, patch command handler does the opposite. See getJBossHome(args) method in https://github.com/wildfly/wildfly-core/blob/master/patching/src/main/java/org/jboss/as/patching/cli/PatchHandler.java
Although, it allows a user to point to the distribution using an argument too.

jamezp:
Hmm... ...it does seem it should be consistent. I don't mind checking the environment variable first. My thought was just providing a way to override the environment property. It really might not matter though. I'd imagine it's a rare case the environment would be set and the user wants to override that though.

I'll go ahead and switch it to check the `JBOSS_HOME` first and only check `jboss.home.dir` if `JBOSS_HOME` was `null`.

PR #579, opened 2015-03-20
Support for WFLY-334 HA Singleton deployer for applications

pferraro:
Introduces configurable strategy for building the next deployment phase service.

https://issues.jboss.org/browse/WFLY-334

### DIFF
--- a/server/src/main/java/org/jboss/as/server/deployment/DeploymentUnitPhaseService.java
+++ b/server/src/main/java/org/jboss/as/server/deployment/DeploymentUnitPhaseService.java
@@ -126,7 +126,8 @@ final class DeploymentUnitPhaseService<T> implements Service<T> {
         if (nextPhase != null) {
             final ServiceName serviceName = DeploymentUtils.getDeploymentUnitPhaseServiceName(deploymentUnit, nextPhase);
             phaseService = DeploymentUnitPhaseService.create(deploymentUnit, nextPhase);
-            phaseServiceBuilder = serviceTarget.addService(serviceName, phaseService);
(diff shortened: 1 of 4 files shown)


pferraro:
Can someone review this?  I'm waiting to submit the wildfly full PR for singleton deployments since it depends on this.

server/src/main/java/org/jboss/as/server/deployment/DeploymentUnitPhaseService.java
```diff
@@ -126,7 +126,8 @@ public void transition(final ServiceController<?> controller, final ServiceContr
         if (nextPhase != null) {
             final ServiceName serviceName = DeploymentUtils.getDeploymentUnitPhaseServiceName(deploymentUnit, nextPhase);
             phaseService = DeploymentUnitPhaseService.create(deploymentUnit, nextPhase);
-            phaseServiceBuilder = serviceTarget.addService(serviceName, phaseService);
+            DeploymentUnitPhaseBuilder builder = this.deploymentUnit.getAttachment(Attachments.DEPLOYMENT_UNIT_PHASE_BUILDER);
```
stuartwdouglas: So this happens before the deployers for the current phase are run. So if you attach this in a given phase it won't affect the next phase, but the one after. Also because it is a DeploymentUnit attachment and not a phase context attachment it will then affect every phase from then on. Is this the behaviour you really want?

pferraro: Good points.  If we process the DUPs before building the service for the next phase, we only need to attach the singleton DeploymentUnitPhaseBuilder to the phase context instead of the DeploymentUnit context - which makes everything simpler.

pferraro:
Closing while I incorporate stewart's feedback.

PR #582, opened 2015-03-26
Don't validate runtime resources

ctomc:


controller/src/main/java/org/jboss/as/controller/ValidateModelStepHandler.java
```diff
@@ -43,7 +43,7 @@ private ValidateModelStepHandler() {
     @Override
     public void execute(OperationContext context, ModelNode operation) throws OperationFailedException {
         final Resource resource = loadResource(context);
-        if (resource == null) {
+        if (resource == null || resource.isRuntime()) {
```
kabir: Perhaps it would be better to have the loadResource() method return null if a resource is runtime somewhere in the tree? Although it is a user error to define a non-runtime child of a runtime resource, I think it could still happen?

ctomc: No it cannot happen, we fixed that not long a go, we make sure that if parent is runtime so is the child. I don't remember the exact commit, but it wasn't long ago. It might have been 752ccd02e6cb7ab1b292ed84f096a2c9ff8d35a9

jamezp:
This still reads the resource children which may be expensive. I think an approach like #586 might be better. Thoughts?

kabir:
+1 loadResource() should return null on the first runtime child in the tree

ctomc:
changed to check for isRuntime() in loadResource()

PR #584, opened 2015-03-26
Support a SELF_CONTAINED mode of operation:

bobmcwhirter:
- Add a SELF_CONTAINED ProcessType and LaunchType
- Configure ServerEnvironment differently if LaunchType.SELF_CONTAINED is used
- Adjust all path-related operations (properties, ServerPathManagerService) to accomodate
  possibly null paths.
- Add a self-contained module
  - ConfigurationPersister
  - ContentRepository
- Add a SelfContainedContainer to be used by wildfly-boot to launch a self-contained
  Wildfly from within an existing and external main().

### DIFF
--- a/self-contained/src/main/java/org/jboss/as/selfcontained/SelfContainedConfigurationPersister.java
+++ b/self-contained/src/main/java/org/jboss/as/selfcontained/SelfContainedConfigurationPersister.java
@@ -1,0 +1,103 @@
+/*
+ * JBoss, Home of Professional Open Source.
+ * Copyright 2011, Red Hat, Inc., and individual contributors
+ * as indicated by the @author tags. See the copyright.txt file in the
+ * distribution for a full listing of individual contributors.
+ *
+ * This is free software; you can redistribute it and/or modify it
+ * under the terms of the GNU Lesser General Public License as
+ * published by the Free Software Foundation; either version 2.1 of
+ * the License, or (at your option) any later version.
+ *
+ * This software is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the GNU
+ * Lesser General Public License for more details.
+ *
+ * You should have received a copy of the GNU Lesser General Public
+ * License along with this software; if not, write to the Free
+ * Software Foundation, Inc., 51 Franklin St, Fifth Floor, Boston, MA
+ * 02110-1301 USA, or see the FSF site: http://www.fsf.org.
+ */
+
+package org.jboss.as.selfcontained;
+
+import org.jboss.as.controller.PathAddress;
+import org.jboss.as.controller.persistence.ConfigurationPersistenceException;
+import org.jboss.as.controller.persistence.ConfigurationPersister;
+import org.jboss.as.controller.persistence.ExtensibleConfigurationPersister;
+import org.jboss.as.controller.persistence.SubsystemMarshallingContext;
+import org.jboss.dmr.ModelNode;
+import org.jboss.staxmapper.XMLElementWriter;
+
+import java.io.OutputStream;
+import java.util.List;
+import java.util.Set;
+
+/** A read-only configuration persister that provides configuration
+ *  from a pre-computed list of operations.
+ *
+ * @author Bob McWhirter
+ */
+public class SelfContainedConfigurationPersister implements ExtensibleConfigurationPersister {
+
+    /** Pre-computed configuration */
+    private final List<ModelNode> containerDefinition;
+
+    /** Construct with a pre-computed list of operations.
+     *
+     * @param containerDefinition The list of configuration operations.
+     */
+    public SelfContainedConfigurationPersister(List<ModelNode> containerDefinition) {
+        this.containerDefinition = containerDefinition;
+    }
+
+    @Override
+    public ConfigurationPersister.PersistenceResource store(ModelNode model, Set<PathAddress> affectedAddresses) throws ConfigurationPersistenceException {
+        return new ConfigurationPersister.PersistenceResource() {
+            @Override
+            public void commit() {
+            }
+
+            @Override
+            public void rollback() {
+            }
+        };
+    }
+
+    @Override
+    public void marshallAsXml(ModelNode model, OutputStream output) throws ConfigurationPersistenceException {
+    }
+
+    @Override
+    public List<ModelNode> load() throws ConfigurationPersistenceException {
+        return this.containerDefinition;
+    }
+
+    @Override
+    public void successfulBoot() throws ConfigurationPersistenceException {
+    }
+
+    @Override
+    public String snapshot() throws ConfigurationPersistenceException {
+        return null;
+    }
+
+    @Override
+    public SnapshotInfo listSnapshots() {
+        return ConfigurationPersister.NULL_SNAPSHOT_INFO;
+    }
+
+    @Override
+    public void deleteSnapshot(String name) {
+    }
+
+    @Override
+    public void registerSubsystemWriter(String name, XMLElementWriter<SubsystemMarshallingContext> writer) {
+        // writing is not supported
+    }
+
+    @Override
+    public void unregisterSubsystemWriter(String name) {
+    }
+}
--- a/self-contained/src/main/java/org/jboss/as/selfcontained/SelfContainedContentRepository.java
+++ b/self-contained/src/main/java/org/jboss/as/selfcontained/SelfContainedContentRepository.java
@@ -1,0 +1,124 @@
+/*
+ * JBoss, Home of Professional Open Source.
+ * Copyright 2011, Red Hat, Inc., and individual contributors
+ * as indicated by the @author tags. See the copyright.txt file in the
+ * distribution for a full listing of individual contributors.
+ *
+ * This is free software; you can redistribute it and/or modify it
+ * under the terms of the GNU Lesser General Public License as
+ * published by the Free Software Foundation; either version 2.1 of
+ * the License, or (at your option) any later version.
+ *
+ * This software is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the GNU
+ * Lesser General Public License for more details.
+ *
+ * You should have received a copy of the GNU Lesser General Public
+ * License along with this software; if not, write to the Free
+ * Software Foundation, Inc., 51 Franklin St, Fifth Floor, Boston, MA
+ * 02110-1301 USA, or see the FSF site: http://www.fsf.org.
+ */
+
+package org.jboss.as.selfcontained;
+
+import org.jboss.as.repository.ContentReference;
+import org.jboss.as.repository.ContentRepository;
+import org.jboss.msc.inject.Injector;
+import org.jboss.msc.service.Service;
+import org.jboss.msc.service.ServiceTarget;
+import org.jboss.msc.service.StartContext;
+import org.jboss.msc.service.StartException;
+import org.jboss.msc.service.StopContext;
+import org.jboss.msc.value.InjectedValue;
+import org.jboss.vfs.VirtualFile;
+
+import java.io.IOException;
+import java.io.InputStream;
+import java.util.Collections;
+import java.util.HashMap;
+import java.util.Map;
+import java.util.Set;
+
+/** A content-repository capable of providing a static bit of content.
+ *
+ * @see org.jboss.as.selfcontained.SelfContainedContentService
+ *
+ * @author Bob McWhirter
+ */
+public class SelfContainedContentRepository implements ContentRepository, Service<ContentRepository> {
+
+    /** Install the service. */
+    public static void addService(ServiceTarget serviceTarget) {
+        SelfContainedContentRepository contentRepository = new SelfContainedContentRepository();
+        serviceTarget.addService(SERVICE_NAME, contentRepository)
+                .addDependency( SelfContainedContentService.NAME, VirtualFile.class, contentRepository.getContentInjector() )
+                .install();
+    }
+
+    private InjectedValue<VirtualFile> contentInjector = new InjectedValue<>();
+
+    public Injector<VirtualFile> getContentInjector() {
+        return this.contentInjector;
+    }
+
+    @Override
+    public byte[] addContent(InputStream stream) throws IOException {
+        return new byte[0];
+    }
+
+    @Override
+    public void addContentReference(ContentReference reference) {
+    }
+
+    @Override
+    public VirtualFile getContent(byte[] hash) {
+        // the [0] array is a sentinal for the self-contained content.
+        if ( hash.length == 1 && hash[0] == 0 ) {
+            return this.contentInjector.getValue();
+        }
+        return null;
+    }
+
+    @Override
+    public boolean hasContent(byte[] hash) {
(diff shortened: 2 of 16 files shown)


bobmcwhirter:
ServerEnvironment has the most changes, and one in ApplicationServerService dealing with content-repositories may be sketchy.

Otherwise, it's mostly code to avoid NPEs if certain paths are null, in the setting of properties and in the PathManager.

All the code-paths I've added should basically be avoided in normal execution and just makes them available to the SelfContainedContainer and ultimately wildfly-boot (or whatever it'll end up being called).

ctomc:
this is ok to test
@bobmcwhirter congrats on first WildFly PR :)

server/src/main/java/org/jboss/as/server/SelfContainedContainer.java
```diff
@@ -0,0 +1,154 @@
...
+                            }
+                        });
+                configuration.setModuleLoader(Module.getBootModuleLoader());
+
+                bootstrap.bootstrap(configuration, Collections.<ServiceActivator>singletonList(new SelfContainedContentServiceActivator(content))).get();
+                return;
+            }
+        } catch (Throwable t) {
+            t.printStackTrace();
+            abort(t);
```
ctomc: printing stacktrace twice?

server/src/main/java/org/jboss/as/server/SelfContainedContainer.java
```diff
@@ -0,0 +1,154 @@
...
+        ProductConfig productConfig;
+        if (serverConfig != null && initialServerConfig != null) {
+            throw ServerLogger.ROOT_LOGGER.cannotHaveBothInitialServerConfigAndServerConfig();
+        }
+        String hostControllerName = null; // No host controller unless in domain mode.
+        productConfig = new ProductConfig(Module.getBootModuleLoader(), WildFlySecurityManager.getPropertyPrivileged(ServerEnvironment.HOME_DIR, null), systemProperties);
+        return new ServerEnvironment(hostControllerName, systemProperties, systemEnvironment, serverConfig, null, launchType, runningMode, productConfig);
+    }
+
+    private static String fixPossibleIPv6URL(String val) {
```
ctomc: where is this used?

bobmcwhirter:
Ah, abort() prints it?

server/src/main/java/org/jboss/as/server/SelfContainedContainer.java
ctomc: yup, abort prints it as well

server/src/main/java/org/jboss/as/server/SelfContainedContainer.java
bobmcwhirter: SelfContainedContainer is used by wildfly-boot, an external repository/project.

SelfContainedContainer itself needs to be in the server module, though, best I can tell, given classloaders.  I could try to move it outboard, though, I guess.

bobmcwhirter:
Fixed the double stack-print, and also renamed our tmpdir stuff.  Added a commit.

server/src/main/java/org/jboss/as/server/SelfContainedContainer.java
ctomc: i mean the method fixPossibleIPv6URL() doesn't look to be used anywhere

bobmcwhirter: Ah, that's probably just copy-pasta cruft.  This started out as a copy of the existing Main.java and then modified.

self-contained/src/main/java/org/jboss/as/selfcontained/SelfContainedConfigurationPersister.java
```diff
@@ -0,0 +1,73 @@
+package org.jboss.as.selfcontained;
```
jamezp: Missing copyright

self-contained/src/main/java/org/jboss/as/selfcontained/SelfContainedConfigurationPersister.java
```diff
@@ -0,0 +1,73 @@
...
+        return this.containerDefinition;
+    }
+
+    @Override
+    public void successfulBoot() throws ConfigurationPersistenceException {
+    }
+
+    @Override
+    public String snapshot() throws ConfigurationPersistenceException {
+        return null;
```
jamezp: This should return [`NULL_SNAPSHOT_INFO`](https://github.com/wildfly/wildfly-core/blob/master/controller/src/main/java/org/jboss/as/controller/persistence/ConfigurationPersister.java#L132). The contract itself states `null` will never be returned.

self-contained/src/main/java/org/jboss/as/selfcontained/SelfContainedConfigurationPersister.java
```diff
@@ -0,0 +1,73 @@
...
+    }
+
+    @Override
+    public SnapshotInfo listSnapshots() {
+        return null;
+    }
```
jamezp: This should return [`NULL_SNAPSHOT_INFO`](https://github.com/wildfly/wildfly-core/blob/master/controller/src/main/java/org/jboss/as/controller/persistence/ConfigurationPersister.java#L132). The contract itself states `null` will never be returned.

self-contained/src/main/java/org/jboss/as/selfcontained/SelfContainedContentRepository.java
```diff
@@ -0,0 +1,94 @@
+package org.jboss.as.selfcontained;
```
jamezp: missing copyright

self-contained/src/main/java/org/jboss/as/selfcontained/SelfContainedContentService.java
```diff
@@ -0,0 +1,36 @@
+package org.jboss.as.selfcontained;
```
jamezp: missing copyright

self-contained/src/main/java/org/jboss/as/selfcontained/SelfContainedContentServiceActivator.java
```diff
@@ -0,0 +1,36 @@
+package org.jboss.as.selfcontained;
```
jamezp: missing copyright

self-contained/src/main/java/org/jboss/as/selfcontained/SelfContainedContentServiceActivator.java
```diff
@@ -0,0 +1,36 @@
...
+
+    @Override
+    public void activate(ServiceActivatorContext serviceActivatorContext) throws ServiceRegistryException {
+        VirtualFile mountPoint = VFS.getRootVirtualFile().getChild( "ROOT.war" );
+        try {
+            VFS.mountReal( content, mountPoint );
+            SelfContainedContentService service = new SelfContainedContentService( mountPoint );
+            serviceActivatorContext.getServiceTarget().addService( SelfContainedContentService.NAME, service ).install();
+        } catch (IOException e) {
+            e.printStackTrace();
```
jamezp: Do we really want to print the stack trace and throw an exception?

server/src/main/java/org/jboss/as/server/SelfContainedContainer.java
```diff
@@ -0,0 +1,134 @@
+/**
+ * @author Bob McWhirter
```
jamezp: This should be removed the author tag should likely be appended below.

server/src/main/java/org/jboss/as/server/SelfContainedContainer.java
```diff
@@ -0,0 +1,134 @@
...
+    public SelfContainedContainer() {
+    }
+
+    /**
+     * The main method.
+     *
+     * @param containerDefinition The container definition.
+     */
+    public void start(final List<ModelNode> containerDefinition, File content) {
+        System.setProperty("java.util.logging.manager", "org.jboss.logmanager.LogManager");
```
jamezp: This should use the `WildFlySecurityManager.WildFlySecurityManager.setPropertyPrivileged("java.util.logging.manager", "org.jboss.logmanager.LogManager")`.

jamezp: A side note too we should verify this isn't too late. If any logger is created before this like is invoked the property won't work.

server/src/main/java/org/jboss/as/server/ServerEnvironment.java
```diff
@@ -332,154 +335,187 @@ public ServerEnvironment(final String hostControllerName, final Properties props
...
+                File tmpDir = File.createTempFile( "wildfly-self-contained", ".d" );
+                if ( tmpDir.exists() ) {
+                    while (!tmpDir.delete()) {
+                        try {
+                            Thread.sleep(50);
+                        } catch (InterruptedException e) {
+                            break;
+                        }
+                    }
+                }
```
jamezp: Why is the sleep needed here? If an error occurs during the delete this could loop forever.

server/src/main/java/org/jboss/as/server/ServerEnvironment.java
```diff
@@ -507,14 +543,14 @@ else if (!tmp.mkdirs()) {
...
-        WildFlySecurityManager.setPropertyPrivileged(SERVER_LOG_DIR, serverLogDir.getAbsolutePath());
-        WildFlySecurityManager.setPropertyPrivileged(SERVER_TEMP_DIR, serverTempDir.getAbsolutePath());
+        setPathProperty(HOME_DIR, homeDir);
+        setPathProperty(MODULES_DIR, modulesDir);
+        setPathProperty(SERVER_BASE_DIR, serverBaseDir);
+        setPathProperty(SERVER_CONFIG_DIR, serverConfigurationDir);
+        setPathProperty(SERVER_DATA_DIR, serverDataDir);
+        setPathProperty(SERVER_DEPLOY_DIR, serverContentDir);
+        setPathProperty(SERVER_LOG_DIR, serverLogDir);
+        setPathProperty(SERVER_TEMP_DIR, serverTempDir);
```
jamezp: The [documentation](https://docs.jboss.org/author/display/WFLY8/General+configuration+concepts#Generalconfigurationconcepts-Paths) indicates these should be set automatically. We should either set them or we'll need to update the documentation.

server/src/main/java/org/jboss/as/server/ServerPathManagerService.java
```diff
@@ -37,13 +39,13 @@
...
-        service.addHardcodedAbsolutePath(serviceTarget, ServerEnvironment.SERVER_TEMP_DIR, serverEnvironment.getServerTempDir().getAbsolutePath());
-        service.addHardcodedAbsolutePath(serviceTarget, ServerEnvironment.CONTROLLER_TEMP_DIR, serverEnvironment.getControllerTempDir().getAbsolutePath());
+        addAbsolutePath(service, serviceTarget, ServerEnvironment.HOME_DIR, serverEnvironment.getHomeDir());
+        addAbsolutePath(service, serviceTarget, ServerEnvironment.SERVER_BASE_DIR, serverEnvironment.getServerBaseDir());
+        addAbsolutePath(service, serviceTarget, ServerEnvironment.SERVER_CONFIG_DIR, serverEnvironment.getServerConfigurationDir());
+        addAbsolutePath(service, serviceTarget, ServerEnvironment.SERVER_DATA_DIR, serverEnvironment.getServerDataDir());
+        addAbsolutePath(service, serviceTarget, ServerEnvironment.SERVER_LOG_DIR, serverEnvironment.getServerLogDir());
+        addAbsolutePath(service, serviceTarget, ServerEnvironment.SERVER_TEMP_DIR, serverEnvironment.getServerTempDir());
+        addAbsolutePath(service, serviceTarget, ServerEnvironment.CONTROLLER_TEMP_DIR, serverEnvironment.getControllerTempDir());
```
jamezp: Same note as above about the documentation.

server/src/main/java/org/jboss/as/server/ServerEnvironment.java
bobmcwhirter: I'll make it less dangerous, but the JVM may not immediate delete a file after successful delete().  Apparently.

server/src/main/java/org/jboss/as/server/ServerPathManagerService.java
bobmcwhirter: Yah, one of the points of SELF_CONTAINED is a removal of some assumptions around filesystem. While we could conceivably set some of them (such as SERVER_CONFIG_DIR to $PWD), it would be somewhat meaningless.  I think ultimately safe is to not set them. 
Then, if a service *does* attempt to grab a path from the PathManager, we get a nice exception so we can know that the service needs to be investigated, instead of it silenting trudging forward with an invalid/bogus path.

My vote would be a documentation update.

bobmcwhirter:
Rebased and notes from James have been applied.

server/src/main/java/org/jboss/as/server/ServerEnvironment.java
jamezp: Okay the sleep is fine then. Maybe use [`if (Files.deleteIfExists(tmpDir.toPath())) { Thread.sleep(50L);}`](http://docs.oracle.com/javase/7/docs/api/java/nio/file/Files.html#deleteIfExists(java.nio.file.Path)).

server/src/main/java/org/jboss/as/server/ServerEnvironment.java
```diff
@@ -332,154 +335,183 @@ public ServerEnvironment(final String hostControllerName, final Properties props
...
-        }
-        if (standalone) {
-            if (!tmp.exists()) {
-                throw ServerLogger.ROOT_LOGGER.serverBaseDirectoryDoesNotExist(tmp);
-            } else if (!tmp.isDirectory()) {
-                throw ServerLogger.ROOT_LOGGER.serverBaseDirectoryIsNotADirectory(tmp);
+            try {
+                File tmpDir = File.createTempFile( "wildfly-self-contained", ".d" );
+                if ( tmpDir.exists() ) {
+                    if ( ! tmpDir.delete() ) {
```
jamezp: I just realized this might not work. It will error if the directory is not empty. We'd likely need a recursive delete.

server/src/main/java/org/jboss/as/server/ServerEnvironment.java
```diff
@@ -332,154 +335,183 @@ public ServerEnvironment(final String hostControllerName, final Properties props
...
+                        throw new RuntimeException( "Unable to delete temporary file" );
```
bstansberry: This will need i18n; i.e. add a method to org.jboss.as.server.logging.ServerLogger.

server/src/main/java/org/jboss/as/server/ServerPathManagerService.java
bstansberry: I think this will end up as a documentation issue as well. Not just docs though; the path stuff is in the xsd too:

https://github.com/wildfly/wildfly-core/blob/master/server/src/main/resources/schema/wildfly-config_3_0.xsd#L3526

If there's no meaningful value for a path, I don't think we should provide one, but for sure we'll need to update docs.

Just for the sake of being robust, +1 to limiting the null check to SELF_CONTAINED servers.

bobmcwhirter: SELF_CONTAINED doesn't use any config XML, so I dunno if the .xsd needs touching?

bobmcwhirter:
Okay, integrated recommendations, and rebased to HEAD.

jamezp:
retest this please

PR #586, opened 2015-03-27
Skip validation of runtime resources

jamezp:
In PR #582 the resources children are still read which in some cases could be expensive. This stops the resource from being validated as well as being read.


bstansberry:
Note this doesn't supercede #582, as this is about eliminating superfluous work during boot, while the other applies at any time.

PR #590, opened 2015-03-31
Nested attribute groups

tadamski:
Nested attribute groups are introduced. XML parsing extension enables descriptive XML creation without creating arfiticial child resources. Attribute groups are metadata only, they are not used in addressing and operations (except read-attribute-group).

I have abstracted parsing/persisting xml attributes from PersistentResourceXMLDescription class -
AttributesXMLDescription interface is introduced.

I have provided two implementations of it:
NestGroupedAttributesXMLDescription - creates nested xml tags based on nested elements of the group; if groups consist only of one element then this class works the same as current implementation;
IgnoreGroupsAttributesXMLDescription - attribute groups are ignored during parsing - equivalent of running current code with useElementsForGroups flag set to false

Both classes work correctly with any combination and XML order of root attributes, grouped attributes, property attributes (with wrapping possibility) and children resources (with wrapping possibility).

Passes: widlfy-core tests, wildfly tests, integration tests.

### DIFF
--- a/controller/src/main/java/org/jboss/as/controller/AttributeDefinition.java
+++ b/controller/src/main/java/org/jboss/as/controller/AttributeDefinition.java
@@ -58,6 +58,7 @@ import org.jboss.dmr.ModelType;
  * methods for validation.
  *
  * @author Brian Stansberry (c) 2011 Red Hat Inc.
+ * @author <a href=mailto:tadamski@redhat.com>Tomasz Adamski</a>
  */
 public abstract class AttributeDefinition {
 
@@ -84,7 +85,7 @@ public abstract class AttributeDefinition {
     private final List<AccessConstraintDefinition> accessConstraints;
     private final Boolean nilSignificant;
     private final AttributeParser parser;
-    private final String attributeGroup;
+    private final AttributeGroup attributeGroup;
 
     // NOTE: Standards for creating a constructor variant are:
     // 1) Expected to be a common use case; no one-offs.
@@ -167,7 +168,7 @@ public abstract class AttributeDefinition {
                                 final ParameterCorrector valueCorrector, final ParameterValidator validator, final boolean validateNull,
                                 final String[] alternatives, final String[] requires, AttributeMarshaller attributeMarshaller,
                                 boolean resourceOnly, DeprecationData deprecationData, final List<AccessConstraintDefinition> accessConstraints,
-                                Boolean nilSignificant, AttributeParser parser, final String attributeGroup, ModelNode[] allowedValues, final EnumSet<AttributeAccess.Flag> flags) {
+                                Boolean nilSignificant, AttributeParser parser, final AttributeGroup attributeGroup, ModelNode[] allowedValues, final EnumSet<AttributeAccess.Flag> flags) {
 
         this.name = name;
         this.xmlName = xmlName == null ? name : xmlName;
@@ -290,7 +291,7 @@ public abstract class AttributeDefinition {
      *
      * @return the name of the group, or {@code null} if the attribute is not associated with a group
      */
-    public String getAttributeGroup() {
+    public AttributeGroup getAttributeGroup() {
         return attributeGroup;
     }
 
@@ -781,7 +782,11 @@ public abstract class AttributeDefinition {
         result.get(ModelDescriptionConstants.TYPE).set(type);
         result.get(ModelDescriptionConstants.DESCRIPTION); // placeholder
         if (attributeGroup != null) {
-            result.get(ModelDescriptionConstants.ATTRIBUTE_GROUP).set(attributeGroup);
+            final ModelNode attributeGroupNode = new ModelNode();
+            for (final String component : attributeGroup) {
+                attributeGroupNode.add(component);
+            }
+            result.get(ModelDescriptionConstants.ATTRIBUTE_GROUP).set(attributeGroupNode);
         }
         result.get(ModelDescriptionConstants.EXPRESSIONS_ALLOWED).set(isAllowExpression());
         if (forOperation) {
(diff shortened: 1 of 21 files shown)


ctomc:
:-1:

controller/src/main/java/org/jboss/as/controller/transform/AttributeGroup.java
```diff
@@ -0,0 +1,102 @@
...
+ * Lesser General Public License for more details.
+ *
+ * You should have received a copy of the GNU Lesser General Public
+ * License along with this software; if not, write to the Free
+ * Software Foundation, Inc., 51 Franklin St, Fifth Floor, Boston, MA
+ * 02110-1301 USA, or see the FSF site: http://www.fsf.org.
+ */
+
+
+package org.jboss.as.controller.transform;
```
bstansberry: Why in this package?

controller/src/main/java/org/jboss/as/controller/PersistentResourceXMLDescription.java
```diff
@@ -122,68 +99,42 @@ public PathElement getPathElement() {
...
-        String name = parseAttributeGroups(reader, op, wildcard);
+        String name = parseRootAttributes(reader, op, wildcard);
         if (wildcard && name == null) {
             throw ControllerLogger.ROOT_LOGGER.missingRequiredAttributes(new StringBuilder(NAME), reader.getLocation());
         }
+
         PathElement path = wildcard ? PathElement.pathElement(resourceDefinition.getPathElement().getKey(), name) : resourceDefinition.getPathElement();
         PathAddress address = parentAddress.append(path);
-        if (!noAddOperation) {
+        if(!noAddOperation) {
```
bstansberry: Please adjust your IDE to not make whitespace changes to lines otherwise unedited. Also, removing the space is incorrect.

controller/src/main/java/org/jboss/as/controller/AttributeDefinition.java
```diff
@@ -290,11 +292,24 @@ public ModelNode getDefaultValue() {
...
+    public AttributeGroup getAttributeGroup() {
         return attributeGroup;
     }
 
     /**
+     * Gets the stringified name of the attribute group with which this attribute is associated, if any.
+     *
+     * @return the stringified name of the group, or {@code null} if the attribute is not associated with a group
+     */
+    public String getStringifiedAttributeGroup() {
```
bstansberry: I don't like this here; it's bloat to the already large AttributeDefinition API just to save a null check in one place.

controller/src/main/java/org/jboss/as/controller/AttributesXMLDescription.java
```diff
@@ -0,0 +1,100 @@
...
+import org.jboss.staxmapper.XMLExtendedStreamReader;
+import org.jboss.staxmapper.XMLExtendedStreamWriter;
+
+/**
+ * XML description of attributes in a {@link org.jboss.as.controller.registry.Resource}. Allows parsing and persisting
+ * Resource's attributes XML configuration.
+ *
+ * @author <a href=mailto:tadamski@redhat.com>Tomasz Adamski</a>
+ */
+public interface AttributesXMLDescription {
```
bstansberry: I'm not enthusiastic about adding this interface at this time as a publicly available extension point. It becomes an API we may have to support for a decade, and I don't believe that over the next 6 months we're going to have the bandwidth to properly evaluate whether it's a good fit for such a long commitment.

Do we have a specific, known use case for an impl of this other than the Ignore and Nest impls you have here? If not we should consider using an abstract class with a package-protected constructor instead of an interface (or some other approach with similar semantics).

controller/src/main/java/org/jboss/as/controller/operations/global/GlobalOperationAttributes.java
```diff
@@ -90,4 +93,11 @@
     .setAllowNull(true)
     .build();
 
+    static final AttributeDefinition GROUP_ELEMENT = new SimpleAttributeDefinitionBuilder(ModelDescriptionConstants.GROUP_ELEMENT, ModelType.STRING, true)
+    .setAllowNull(false)
+    .build();
+
+    static final SimpleListAttributeDefinition ATTRIBUTE_GROUP = new SimpleListAttributeDefinition.Builder(ModelDescriptionConstants.NAME, GROUP_ELEMENT)
+    .setAllowNull(true)
```
bstansberry: This should have a ParameterCorrector associated such that it is tolerant of a value of ModelType.STRING instead of ModelType.LIST, with the correcting converting the string to a single element list. For example, see

https://github.com/wildfly/wildfly-core/blob/master/server/src/main/java/org/jboss/as/server/controller/resources/DeploymentAttributes.java#L166

and 

https://github.com/wildfly/wildfly-core/blob/master/server/src/main/java/org/jboss/as/server/controller/resources/DeploymentAttributes.java#L415

ctomc:
I should probably explain my :-1: for this, as we discussed on mailing lists and other conversations. 
Notion of nested attribute groups is completely useless as it doesn't add anything useful beyond some cosmetics in XML, which are not that useful also considering that we will be adding also support other non XML persistent formats.
Developers shouldn't bother with how XML looks beyond that it is readable. They should however bother with how management model looks like and nested attribute groups don't help with that at all.
Even if argument would be that it "helps" with GUI in console, lets be fair that having fieldsets is sufficient, having nested fieldsets brings no real improvements.

tadamski:
My rationale for that change was that we are getting more descriptive presentation metadata at low cost. Usefulness of it is not restricted to XML. If it is useful to group something in XML for presentation reasons then it is likely that it would be useful in other formats.

PR #593, opened 2015-04-03
[WFCORE-623] add missing methods to ModelControllerClient.Factory

Ladicek:


bstansberry:
Instead of adding more overloaded methods, I'd rather deprecate a bunch of the existing variants and refer people to public static ModelControllerClient create(final ModelControllerClientConfiguration configuration) and ModelControllerClientConfiguration.Builder.

Ladicek:
Having a builder is certainly preferable, but I believe that these should be added too. The reason is EAP compatibility, as explained in the JIRA.

bstansberry:
I see, so the first method you added is actually restoring one that got dropped, which we need for sure. I don't see the need for or like adding the second one, but I don't care enough to argue about it. :)

I've superceded this with #600 though, as I want all these convoluted factory methods deprecated. And then I did some other clean up while adding the @Deprecated stuff.

Ladicek:
Thank you, Brian :-) I added the second one purely for symetry, no other reason.

PR #594, opened 2015-04-03
[WFCORE-569] : Clean the DiscoveryOption API

ehsavoie:
DiscoveryOption discover should return a list of connection parameters to a DC instead of only one value set into the DiscoveryOption.
Adding RemoteDomainControllerConnectionConfiguration as a holder class for such parameters.

Jira: https://issues.jboss.org/browse/WFCORE-569


host-controller/src/main/java/org/jboss/as/host/controller/discovery/S3Discovery.java
```diff
@@ -117,21 +111,7 @@ public void allowDiscovery(List<DomainControllerManagementInterface> interfaces)
...
-                        return 1;
-                    }
-                    if (Protocol.HTTPS_REMOTING.toString().equals(data.getProtocol())) {
-                        if (Protocol.REMOTE.toString().equals(otherData.getProtocol())) {
-                            return -1;
-                        }
-                        return 1;
-                    }
-                    return -1;
-                }});
```
bstansberry: Why is this sorting being removed? This is what ensures that legacy hosts see the remote protocol that they understand.

ehsavoie: I misunderstood your remark about https://github.com/bstansberry/wildfly-core/commit/c873c41eff7331f432a664fffe928d94fc8819dd , I thought that this was no longer required at all. Re-reading the discussion I reverting this.

ehsavoie:
retest this plese

PR #595, opened 2015-04-06
WFCORE-626 Global list-get operation can inadvertently create list elements

pferraro:
https://issues.jboss.org/browse/WFCORE-626


ctomc:
:+1:

PR #597, opened 2015-04-07
[WFCORE-282] Wildcard read support across domain processes

bstansberry:
Support for variations of /host=*/**, /host=*/server=* and /host=x/server=* addresses in the ops that support wildcard reads (read-resource, read-attribute, read-attribute-group, and the upcoming WFCORE-287 'query' op.)

Also includes commits fixing or working around issues found while testing this: WFCORE-619, WFCORE-622 and WFCORE-627.

Compatibility note: Previously, wildcard reads would store any RBAC access-control response header data in the items in the response's "result" node. This change strips them out from there and places them in the top level response, consistent with non-wildcard reads. So instead of

```
{
    "outcome"=>"success",
    "result"=> [{
        "address"=>[("foo","bar")],
        "result"=>{ some stuff},
        "response-headers"=> {
            "access-control"=>[{
                "absolute-address"=>[("foo","bar")]
                ....
            }]
        }
    }]
}
```

It is now

```
{
    "outcome"=>"success",
    "result"=> [{
        "address"=>[("foo","bar")],
        "result"=>{ some stuff}
    }],
    "response-headers"=> {
        "access-control"=>[{
            "absolute-address"=>[("foo","bar")]
            ....
        }]
    }
}
```

I regard the previous structure as being a bug. But input from @heiko-braun or @hpehl on whether that will cause problems in the console is needed before this can be merged.

Note also there is an open issue related to this work in a mixed domain: see WFCORE-621.

### DIFF
--- a/controller/src/main/java/org/jboss/as/controller/ProxyStepHandler.java
+++ b/controller/src/main/java/org/jboss/as/controller/ProxyStepHandler.java
@@ -55,9 +55,11 @@ import org.jboss.dmr.ModelNode;
 public class ProxyStepHandler implements OperationStepHandler {
 
     private final ProxyController proxyController;
+    private final boolean forServer;
 
     public ProxyStepHandler(final ProxyController proxyController) {
         this.proxyController = proxyController;
+        this.forServer = proxyController.getProxyNodeAddress().getLastElement().getKey().equals(RUNNING_SERVER);
     }
 
     @Override
@@ -239,6 +241,9 @@ public class ProxyStepHandler implements OperationStepHandler {
                             if (context.getProcessType() == ProcessType.HOST_CONTROLLER && responseNode.has(SERVER_GROUPS)) {
                                 context.getServerResults().set(responseNode.get(SERVER_GROUPS));
                             }
+                            if (responseNode.hasDefined(RESPONSE_HEADERS)) {
+                                context.getResponseHeaders().set(processResponseHeaders(responseNode.get(RESPONSE_HEADERS)));
+                            }
 
                             // Make sure any streams associated with the remote response are properly
                             // integrated with our response
@@ -264,6 +269,20 @@ public class ProxyStepHandler implements OperationStepHandler {
             if (!completeStepCalled && txRef.get() != null) {
                 txRef.get().rollback();
             }
+        }
+    }
+
+    private ModelNode processResponseHeaders(ModelNode responseHeaders) {
+        if (!responseHeaders.hasDefined(ACCESS_CONTROL) || !forServer) {
+            return responseHeaders;
+        } else {
+            ModelNode result = responseHeaders.clone();
+            for (ModelNode accItem : result.get(ACCESS_CONTROL).asList()) {
+                ModelNode itemAddrNode = accItem.get("absolute-address");
+                PathAddress itemAddr = PathAddress.pathAddress(itemAddrNode);
+                itemAddrNode.set(proxyController.getProxyNodeAddress().append(itemAddr).toModelNode());
+            }
+            return result;
         }
     }
 
--- a/controller/src/main/java/org/jboss/as/controller/ProxyStepHandler.java
+++ b/controller/src/main/java/org/jboss/as/controller/ProxyStepHandler.java
@@ -22,6 +22,7 @@
 
 package org.jboss.as.controller;
 
+import static org.jboss.as.controller.descriptions.ModelDescriptionConstants.ACCESS_CONTROL;
 import static org.jboss.as.controller.descriptions.ModelDescriptionConstants.FAILED;
 import static org.jboss.as.controller.descriptions.ModelDescriptionConstants.FAILURE_DESCRIPTION;
 import static org.jboss.as.controller.descriptions.ModelDescriptionConstants.OP;
@@ -29,6 +30,7 @@ import static org.jboss.as.controller.descriptions.ModelDescriptionConstants.OP_
 import static org.jboss.as.controller.descriptions.ModelDescriptionConstants.OUTCOME;
 import static org.jboss.as.controller.descriptions.ModelDescriptionConstants.RESPONSE_HEADERS;
 import static org.jboss.as.controller.descriptions.ModelDescriptionConstants.RESULT;
+import static org.jboss.as.controller.descriptions.ModelDescriptionConstants.RUNNING_SERVER;
 import static org.jboss.as.controller.descriptions.ModelDescriptionConstants.SERVER_GROUPS;
 
 import java.io.IOException;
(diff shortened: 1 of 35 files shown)


bstansberry:
#596 showed the same failures on the build-45 CI slave: 

http://brontes.lab.eng.brq.redhat.com/viewLog.html?buildId=49339&tab=buildResultsDiv&buildTypeId=WildFlyCore_PullRequest_WildFlyCoreFullIntegration

Log shows JGroups discovery issues during server boot:

&amp#27;[0m&amp#27;[31m14:39:10,325 SEVERE [org.jgroups.protocols.MPING] (MSC service thread 1-3) failed sending discovery request: java.io.IOException: Network is unreachable
    at java.net.PlainDatagramSocketImpl.send(Native Method)
    at java.net.DatagramSocket.send(DatagramSocket.java:693)
    at org.jgroups.protocols.MPING.sendMcastDiscoveryRequest(MPING.java:295)

So my guess is this is unrelated to this PR.

bstansberry:
retest this please

hpehl:
Moving the `response-headers` element should not affect the current RBAC processing. We do not use the information inside this element at the moment. We read the `access-control` data inside the `result` node, which I expect will remain unchanged:

```
{
    "outcome" => "success",
    "result" => [{
        "address" => [
            ("subsystem" => "resource-adapters"),
            ("resource-adapter" => "*")
        ],
        "outcome" => "success",
        "result" => {
            "description" => ...,
            ...
            "access-control" => {
                ...
            }
        }
    }]
}
```

bstansberry:
@hpehl -- Thanks. And yes, the output of something like /subsystem=resource-adapters/resource-adapter=*:read-resource-description(access-control=[combined|trim]-descriptions) is not affected. For read-resource-description the access-control data is part of the intended result of the request, it's not "out of band" extra information.

heiko-braun:
+1 on what Harald said. Furthermore your changes look like a consolidation of the response structures, which is a good thing.

controller/src/main/java/org/jboss/as/controller/ProxyStepHandler.java
```diff
@@ -267,6 +276,20 @@ public void handleResult(OperationContext.ResultAction resultAction, OperationCo
         }
     }
 
+    private ModelNode processResponseHeaders(ModelNode responseHeaders) {
+        if (!responseHeaders.hasDefined(ACCESS_CONTROL) || !forServer) {
+            return responseHeaders;
+        } else {
+            ModelNode result = responseHeaders.clone();
+            for (ModelNode accItem : result.get(ACCESS_CONTROL).asList()) {
+                ModelNode itemAddrNode = accItem.get("absolute-address");
```
jamezp: I wouldn't say this is a blocker, but should `"absolute-address"` be a constant since it's used elsewhere as well? I don't think it's  a big deal either way.

controller/src/main/java/org/jboss/as/controller/operations/global/GlobalOperationHandlers.java
```diff
@@ -255,33 +258,40 @@ public void handleResult(OperationContext.ResultAction resultAction, OperationCo
...
+                final ImmutableManagementResourceRegistration registration, boolean ignoreMissing) {
             try {
-                execute(address, base, context, ignoreMissing);
+                ControllerLogger.MGMT_OP_LOGGER.tracef("safeExecute for %s, remaining is %s", base, remaining);
+                execute(base, remaining, context, registration, ignoreMissing);
             } catch (UnauthorizedException e) {
                 // equivalent to the resource not existing
                 // Just report the failure to the filter and complete normally
                 filteredData.addReadRestrictedResource(base);
+                ControllerLogger.MGMT_OP_LOGGER.tracef("Caught UnauthorizedException in %s", this);
```
jamezp: Should `toString()` be overridden so when `this` is logged it's something more meaningful than the default `toString()`?

controller/src/main/java/org/jboss/as/controller/operations/global/GlobalOperationHandlers.java
```diff
@@ -296,28 +306,178 @@ private void safeExecute(final PathAddress address, final PathAddress base, fina
...
+                                @Override
+                                public void execute(OperationContext context, ModelNode operation) throws OperationFailedException {
+                                    try {
+                                        delegate.execute(context, operation);
+                                    } catch (UnauthorizedException e) {
+                                        // equivalent to the resource not existing
+                                        // Just report the failure to the filter and complete normally
+                                        filteredData.addReadRestrictedResource(base);
+                                        filtered.set(true);
+                                        ControllerLogger.MGMT_OP_LOGGER.tracef("Caught UnauthorizedException in %s from %s", this, delegate);
```
jamezp: `toString()` probably should be overridden here otherwise you'll end up with something like `org.jboss.as.controller.operations.global.GlobalOperationHandlers.ModelAddressResolver.$1@xxxxx`. Maybe not super important I guess. I'll keep reading further.

host-controller/src/main/java/org/jboss/as/domain/controller/operations/coordination/DomainFinalResultHandler.java
```diff
@@ -188,38 +216,109 @@ private boolean collectHostFailures(final OperationContext context, final boolea
...
-                result = getHostControllerResult(hostResponse, stepLabels);
+
+            if (hostName.equals(multiphaseContext.getLocalHostInfo().getLocalHostName())) {
+                //ModelNode hostFinalResult = multiphaseContext.getLocalContext().getLocalResponse();
+                //result = getFromFinalHCResult(hostFinalResult, stepLabels);
+                result = getHostControllerResult(localDomainFormatted, stepLabels);
+            } else {
+                ModelNode hostFinalResult = multiphaseContext.getHostControllerFinalResults().get(hostName);
+                if (hostFinalResult != null) {
+                    //result = getFromFinalHCResult(hostFinalResult, stepLabels);
```
jamezp: Any reason to keep these commented lines; 249, 250 and 255, around?

bstansberry: Nope.

controller/src/main/java/org/jboss/as/controller/operations/global/GlobalOperationHandlers.java
bstansberry: That was sufficient info when I added these; enough to distinguish what code point was logging vs the couple other cases. It is cryptic though.

controller/src/main/java/org/jboss/as/controller/ProxyStepHandler.java
bstansberry: I actually have that on my TODO list, but I wanted to get this done and also figured this patch was big enough! I'll do it at some point pretty soon though.

bstansberry: I'll do it now; this affects fewer spots than I thought.

controller/src/main/java/org/jboss/as/controller/operations/global/GlobalOperationHandlers.java
bstansberry: This one is fine; it will be the class name that is logging, which is the useful data.

PR #598, opened 2015-04-07
https://issues.jboss.org/browse/WFCORE-625 Error message in CLI if exten...

aloubyansky:
...sion existing on the server is not present in the CLI instalaltion

This commit changes the error to a simple message on the terminal and rewords the message to indicate it's not critical to the CLI session in general.

Thanks,
Alexey


aloubyansky:
I've decided to re-work this. Sorry for the noise.

PR #599, opened 2015-04-07
https://issues.jboss.org/browse/WFCORE-625 Error message in CLI if exten...

aloubyansky:
...sion existing on the server is not present in the CLI instalaltion

With this commit, it collects the errors while loading commands from extensions and displays a one line warning saying that there were problems, for more details the user can execute extension-commands --errors.

Thanks,
Alexey

### DIFF
--- a/cli/src/main/java/org/jboss/as/cli/impl/ExtensionsLoader.java
+++ b/cli/src/main/java/org/jboss/as/cli/impl/ExtensionsLoader.java
@@ -35,6 +35,7 @@ import org.jboss.as.cli.CommandLineException;
 import org.jboss.as.cli.CommandRegistry;
 import org.jboss.as.cli.ControllerAddress;
 import org.jboss.as.cli.Util;
+import org.jboss.as.cli.handlers.CommandHandlerWithHelp;
 import org.jboss.as.controller.client.ModelControllerClient;
 import org.jboss.dmr.ModelNode;
 import org.jboss.dmr.Property;
@@ -60,7 +61,7 @@ class ExtensionsLoader {
     /** error messages collected during loading the commands */
     private List<String> errors = Collections.emptyList();
 
-    ExtensionsLoader(CommandRegistry registry, CommandContext ctx) {
+    ExtensionsLoader(CommandRegistry registry, CommandContext ctx) throws CommandLineException {
         assert registry != null : "command registry is null";
         assert ctx != null : "command context is null";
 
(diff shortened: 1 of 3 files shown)


aloubyansky:
retest this please!

aloubyansky:
retest this please!

bstansberry:
retest this please

cli/src/main/java/org/jboss/as/cli/impl/ExtensionsLoader.java
```diff
@@ -97,6 +100,7 @@ void loadHandlers(ControllerAddress address) throws CommandLineException {
         assert client != null : "client is null";
 
         if(moduleLoader == null) {
+            ctx.printLine("Wanring! The CLI is running in a non-modular environment and cannot load commands from management extensions.");
```
kabir: Typo

cli/src/main/resources/help/extension-commands.txt
```diff
@@ -0,0 +1,19 @@
+SYNOPSIS
+
+    extension-commands (--errors | --help)
+
+DESCRIPTION
+
+    In case --errors argument is present the command will print detailed error
```
kabir: Since there was another typo..."In case the --errors..."

bstansberry:
This "org.jboss.as.test.manualmode.management.cli.CLIEmbedServerTestCase.testStdOutEcho" failure is quite strange. It only fails when testing this PR branch. It's been run 386 times and has failed 6, all on tests of this PR. The failure happens when the embedded server tries to boot and MSC tries to register mbeans, failing because said mbeans are already registered. Which would imply something is not getting cleaned up in the test driver process.

But this PR doesn't add any new tests and I don't see anything in it that should trigger MSC to register mbeans.

Perhaps something else is failing and then the MSC mbean thing is just later noise when the test doesn't deal well with the original failure.

bstansberry:
Ah, maybe that logging is a different problem but it doesn't fail the test.

Looking at the log output in the test failure details, it looks garbled.

bstansberry:
I think the problem is in EmbedServerHandler. It binds the MCC to the CommandContext before the server is completely started. This patch results in the CLI writing to the console as part of handling the bind of the MCC. This is producing garbage output.

I think the binding can just be moved down in the method; I'll see.

bstansberry:
JIRA for the unrelated mbean issue is https://issues.jboss.org/browse/WFCORE-657

I'm closing this in favor of #632, which is this plus an additional commit to fix the test failure a la my last comment above.

PR #604, opened 2015-04-09
[WFCORE-65][WFCORE-632] Use an 'owner' attribute to determine if a scanner controls a deployment

bstansberry:


bstansberry:
Reviewed by Emmanuel (thanks!)

PR #606, opened 2015-04-09
Add PersistentResourceXMLBuilder#addAttributes method with Collection pa...

tadamski:
...rameter

PersistentResourceDefinition class encourages using collection for storing attributes and children. On the other hand XML description has varargs methods. Adding this convenience method will remove the need of converting between collection and arrays in subsystem parsers.


ctomc:
*NO*
This was originally part of API but was internationally removed to make sure developers don't do mistakes with versioned parsers, where they would use addAttributes(resource.getAttributes())
which would mistakenly also have all new attributes for old parser.

ctomc:
see https://github.com/wildfly/wildfly-core/commit/765a60fe445ca8986937ce12957ed632aa73e94a

tadamski:
OK. Closing it then.

PR #610, opened 2015-04-10
[WFCORE-548] Remove the reload requirement if the append value is change...

jamezp:
...d for a file handler.


bstansberry:
retest this please

PR #613, opened 2015-04-10
register notifications for the ModelController server root resource

jmesnil:
* this prevents warnings in tests using this class about resource
  emitting unregistered notifications.


jmesnil:
retest this please

PR #615, opened 2015-04-10
[WFCORE-638] :Deployment runtime-name mismatch between content repo and server-group and logs

ehsavoie:
The server-group reference is now correctly updated.

Jira: https://issues.jboss.org/browse/WFCORE-638


bstansberry:
looks good, assuming tests pass

PR #619, opened 2015-04-14
Include the original cause for failure to create a temp file provider

jaikiran:
The commit here enhances the logging to include the original exception that caused the temp file provider creation to fail. This should help in cases like the one reported in the forum thread here https://developer.jboss.org/thread/255634


bstansberry:
Just thought of something... :)

jaikiran:
Ha ha! Of course, I do agree to those terms :)

P.S: Sorry did not receive any notification from github about your comment Brian, hence the delay in response.

PR #620, opened 2015-04-14
[WFCORE-114] : Get RuntimeVaultReader into WildFly Core

ehsavoie:
Moving RuntimeVaultReader from WildFly to WildFly-Core

Jira: https://issues.jboss.org/browse/WFCORE-114


ctomc:
Does wildfly-core still boot with this in?
As by design we don't include picketbox as part of wildfly-core distro but only wildfly-full ships PB modules.

ctomc:
as discussed with @ehsavoie this works fine without PB, it is safe to merge

bstansberry:
@ehsavoie I don't see how it can work fine. If you configure a vault, it won't work as the classes are not present.

There are tests in core-model-test of the vault config, but they don't involve the runtime stuff, just Stage.MODEL.

ehsavoie:
To be reworked

PR #621, opened 2015-04-14
[WFCORE-460 Enhanced syntax support

ctomc:
https://issues.jboss.org/browse/WFCORE-460

### DIFF
--- a/controller/src/main/java/org/jboss/as/controller/operations/global/EnhancedSyntaxSupport.java
+++ b/controller/src/main/java/org/jboss/as/controller/operations/global/EnhancedSyntaxSupport.java
@@ -41,7 +41,7 @@ import org.jboss.dmr.ModelType;
 class EnhancedSyntaxSupport {
 
     static final Pattern ENHANCED_SYNTAX_PATTERN = Pattern.compile("(.*[\\.\\[\\]].*)");
-    static final Pattern BRACKETS_PATTERN = Pattern.compile("(.*)\\[(\\d+)\\]");
+    static final Pattern BRACKETS_PATTERN = Pattern.compile("(.*)\\[(-?\\d+)\\]");
     static final Pattern EXTRACT_NAME_PATTERN = Pattern.compile("^(.*?)[\\.\\[].*");
 
     static void resolveAttribute(OperationContext context, ModelNode operation, String attributeName, boolean defaults, ImmutableManagementResourceRegistration registry, boolean enhancedSyntax) throws OperationFailedException {
@@ -92,7 +92,7 @@ class EnhancedSyntaxSupport {
                 int index = Integer.parseInt(matcher.group(2));
                 ModelNode list = attribute.isEmpty()?result:result.get(attribute); // in case we want to additionally resolve already loaded attribute, this usually applies to attributes that have custom read handlers
                 if (list.isDefined() && list.getType() == ModelType.LIST){
-                    result = list.asList().get(index);
+                    result = list.get(index);
                 }else{
                     throw new OperationFailedException(String.format("Could not resolve attribute expression: '%s', type is not a list",attributeExpression));
                 }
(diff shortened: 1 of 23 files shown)


controller/src/main/java/org/jboss/as/controller/operations/global/EnhancedSyntaxSupport.java
```diff
@@ -0,0 +1,146 @@
...
+                    if (!result.isDefined() || result.getType() == ModelType.OBJECT){
+                        result = result.get(part);
+                    }else{
+                        throw ControllerLogger.SERVER_LOGGER.couldNotResolveExpression(attributeExpression);
+                    }
+                }
+            }
+            result.set(newValue);
+            return oldValue;
+        }
```
bstansberry: This method's whitespace is messed up.

controller/src/main/java/org/jboss/as/controller/operations/global/EnhancedSyntaxSupport.java
```diff
@@ -0,0 +1,146 @@
...
+/**
+ * @author Tomaz Cerar (c) 2015 Red Hat Inc.
+ */
+class EnhancedSyntaxSupport {
+
+    static final Pattern ENHANCED_SYNTAX_PATTERN = Pattern.compile("(.*[\\.\\[\\]].*)");
+    static final Pattern BRACKETS_PATTERN = Pattern.compile("(.*)\\[(-?\\d+)\\]");
+    static final Pattern EXTRACT_NAME_PATTERN = Pattern.compile("^(.*?)[\\.\\[].*");
+
+    static void resolveAttribute(OperationContext context, ModelNode operation, String attributeName, boolean defaults, ImmutableManagementResourceRegistration registry, boolean enhancedSyntax) throws OperationFailedException {
```
bstansberry: This method is only called from ReadAttributeHandler, so shouldn't it stay there?

controller/src/main/java/org/jboss/as/controller/operations/global/EnhancedSyntaxSupport.java
```diff
@@ -0,0 +1,146 @@
...
+        final Resource resource = context.readResource(PathAddress.EMPTY_ADDRESS, false);
+        final ModelNode subModel = resource.getModel();
+        if (enhancedSyntax){
+            context.getResult().set(resolveEnhancedSyntax(attributeName,subModel));
+        }else if (subModel.hasDefined(attributeName)) {
```
bstansberry: There's a problem here re the " || !defaults" check from the old L194 of ReadAttributeHandler. That check is not present in the old L170 of ReadAttributeHandler, so there's a difference in the handling between those two cases. The difference is valid, as in the L170 case, there is no AttributeAccess, so if we need to deal with the fact that the attribute name might be bogus. If we added the " || !defaults" check to L170, then when defaults == false, we'd just return undefined instead of dropping into the code that detects the fact that the attribute name is bogus.

My proposal:

1) We change your current L171 of ReadAttributeHandler to no longer call this method, and just do

```
throw new OperationFailedException(ControllerLogger.ROOT_LOGGER.unknownAttribute(attributeName));
```

Basically, we no longer support attributes not registered in the ManagementResourceRegistration.

2) Then L52 ^^^ has the " || !defaults" restored.

controller/src/main/java/org/jboss/as/controller/operations/global/EnhancedSyntaxSupport.java
```diff
@@ -0,0 +1,146 @@
...
+            final ModelNode result = subModel.get(attributeName);
+            context.getResult().set(result);
+        } else {
+            // No defined value in the model. See if we should reply with a default from the metadata,
+            // reply with undefined, or fail because it's a non-existent attribute name
```
bstansberry: If AttributeAccess.getAttributeDefinition() returns non-null, it would sure be simpler to just get the default from the AD. :)

It might return null though, unless we remove this from ManagementResourceRegistration:

```
    @Deprecated
    void registerReadOnlyAttribute(String attributeName, OperationStepHandler readHandler, AttributeAccess.Storage storage);
```

controller/src/main/java/org/jboss/as/controller/operations/global/EnhancedSyntaxSupport.java
```diff
@@ -0,0 +1,146 @@
...
+
+    static final Pattern ENHANCED_SYNTAX_PATTERN = Pattern.compile("(.*[\\.\\[\\]].*)");
+    static final Pattern BRACKETS_PATTERN = Pattern.compile("(.*)\\[(-?\\d+)\\]");
+    static final Pattern EXTRACT_NAME_PATTERN = Pattern.compile("^(.*?)[\\.\\[].*");
+
+    static void resolveAttribute(OperationContext context, ModelNode operation, String attributeName, boolean defaults, ImmutableManagementResourceRegistration registry, boolean enhancedSyntax) throws OperationFailedException {
+        final Resource resource = context.readResource(PathAddress.EMPTY_ADDRESS, false);
+        final ModelNode subModel = resource.getModel();
+        if (enhancedSyntax){
+            context.getResult().set(resolveEnhancedSyntax(attributeName,subModel));
```
bstansberry: I think this L50-51 block is in the wrong place, and that this method could use both the true attribute name and the expression as params.

Line 52-70 deal with reading from the model, providing defaults if requested, and throwing a standard exception if the basic attribute name is wrong. So we shouldn't bypass all that (particularly the exception throwing.) Better is to take output from all that, and then pass that into resolveEnhancedSyntax. That's basically what you are doing with the new L175-186 in ReadAttributeHandler, so having an equivalent call pattern for this branch of the logic seems more robust.

bstansberry:
In general, it's looking good. I'm too lazy to examine the tests in detail. :) Is there any coverage of a rollback of an enhanced syntax write-attribute? That's one case I could imagine getting messed up.

controller/src/main/java/org/jboss/as/controller/operations/global/EnhancedSyntaxSupport.java
ctomc: You are right, at the time I moved it it had some regexes in, doesn't make any sense now.

controller/src/main/java/org/jboss/as/controller/operations/global/EnhancedSyntaxSupport.java
ctomc: Good point, also I was sure that this method was already removed. Will make sure it gets removed :)

PR #624, opened 2015-04-15
Keep VirtualFile throughout.

bobmcwhirter:
Mistakenly made some things java.io.File, which is limiting and not necessary.  VirtualFiles suffice throughout.


bobmcwhirter:
Not yet ready...

PR #626, opened 2015-04-16
WFCORE-640 - Add file permissions checks during execution of add-user.bat/add-user.sh

bgaisford:
Corrected the issue in the unit test case, this is good to go now.

### DIFF
--- a/domain-management/src/test/java/org/jboss/as/domain/management/security/adduser/PropertyFileFinderPermissionsTestCase.java
+++ b/domain-management/src/test/java/org/jboss/as/domain/management/security/adduser/PropertyFileFinderPermissionsTestCase.java
@@ -23,26 +23,20 @@
 package org.jboss.as.domain.management.security.adduser;
 
 import org.jboss.as.domain.management.security.adduser.AddUser.FileMode;
-import org.jboss.msc.service.StartException;
 import org.junit.Before;
 import org.junit.Test;
 
-import java.io.BufferedWriter;
-import java.io.File;
-import java.io.FileOutputStream;
-import java.io.IOException;
-import java.io.OutputStreamWriter;
+import java.io.*;
 import java.nio.charset.StandardCharsets;
 import java.util.Properties;
 
 import static java.lang.System.getProperty;
-import static java.lang.System.setProperty;
 import static org.junit.Assert.assertTrue;
 
 /**
- * Test the property file finder.
+ * Test the property file finder permissions tests.
  *
- * @author <a href="mailto:flemming.harms@gmail.com">Flemming Harms</a>
+ * @author <a href="mailto:bgaisford@punagroup.com">Brandon Gaisford</a>
  */
 public class PropertyFileFinderPermissionsTestCase extends PropertyTestHelper {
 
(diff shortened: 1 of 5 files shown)


darranl:
Generally I think this looks good, the only thing I would say is that the StateValues class is more for the propagation of values from state to state as the utility executes - when we are within a single state an alternative should possibly be used even if that does mean some refactoring of the method calls within the state.

domain-management/src/test/java/org/jboss/as/domain/management/security/adduser/PropertyFileFinderPermissionsTestCase.java
```diff
@@ -0,0 +1,133 @@
...
+ * 02110-1301 USA, or see the FSF site: http://www.fsf.org.
+ */
+
+package org.jboss.as.domain.management.security.adduser;
+
+import org.jboss.as.domain.management.security.adduser.AddUser.FileMode;
+import org.junit.Before;
+import org.junit.Test;
+
+import java.io.*;
```
darranl: Please fully expand the import to cover the classes actually required.

bgaisford:
Would you like me to refactor PropertyFileFinder as per your comments above?  Happy to do so.

darranl:
Yes please, try and keep it so StateValues is only used for values that need to transition to other states.

bstansberry:
this is ok to test

bgaisford:
This is good to go Darran.  It was as simple as moving the permissions checks state from StateValues to internal members of PropertyFileFinder.  Thanks for the excellent feedback.  Please let me know if there is anything else you would like changed based on the latest updates.

kabir:
@bgaisford before this can be merged, please answer the following question:

bgaisford:
Yes, I agree to the terms of the MIT License.

bgaisford:
**@kabir** During development I did create a new source file and I copied the license from an existing similar source file.  Upon closer inspection, the license text I used (see below) doesn't match the license you sent in the link above.  Should I update the source file with the MIT license?
"/*
 * JBoss, Home of Professional Open Source.
 * Copyright 2012, Red Hat, Inc., and individual contributors
 * as indicated by the @author tags. See the copyright.txt file in the
 * distribution for a full listing of individual contributors.
 *
 * This is free software; you can redistribute it and/or modify it
 * under the terms of the GNU Lesser General Public License as
 * published by the Free Software Foundation; either version 2.1 of
 * the License, or (at your option) any later version.
 *
 * This software is distributed in the hope that it will be useful,
 * but WITHOUT ANY WARRANTY; without even the implied warranty of
 * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the GNU
 * Lesser General Public License for more details.
 *
 * You should have received a copy of the GNU Lesser General Public
 * License along with this software; if not, write to the Free
 * Software Foundation, Inc., 51 Franklin St, Fifth Floor, Boston, MA
 * 02110-1301 USA, or see the FSF site: http://www.fsf.org.
 */"

kabir:
@bgaisford The license header is the correct one, although set the year to 2015 rather than 2012.

The MIT license agreement is more about your contributions, and being able to relicense them:
"Until now, contributor agreements applicable to WildFly have generally limited outbound licensing of contributions under LGPL (or  GPL) only. We would now like to have the flexibility to consider  other outbound open source licensing options for components of WildFly that may be incompatible with or less restrictive than LGPL  licensing."

bgaisford:
@kabir Understood.  I have verified licenses and rev'd copyright year in touched and new files.

kabir-ncl:
@bgaisford sorry to be a pain. But you should only update the year in your new files, not in the existing ones which you have modified.

kabir:
@bgaisford sorry to be a pain. But you should only update the year in your new files, not in the existing ones which you have modified.

bgaisford:
@kabir Done, no worries.

PR #627, opened 2015-04-17
JBug preparation

kabir:
This PR undoes the work done for WFLY-3371/https://github.com/wildfly/wildfly/pull/6298 (which incidentally was done for another JBug).

The JBug is on Tuesday 21st, and assuming tests pass, I will merge it before then, and reintroduce the fixes during the JBug.


kabir:
Putting the hold label. I will merge this on Tuesday before the JBug, so I can look all clever and fix again

PR #629, opened 2015-04-17
Support --debug flag without port argument

briehman:
Add a dummy positional parameter in the instance where --debug is
specified but no port is provided. This dummy parameter will be
swallowed instead of the next true parameter.


bstansberry:
this is ok to test

jamezp:
Which OS seems to have this issue?

jamezp:
FWIW this looks like it would fix the issue as well https://github.com/wildfly/wildfly-core/pull/638.

briehman:
Yes, this is fixed by #638. Closing this PR.

jamezp:
Excellent, thank you!

PR #631, opened 2015-04-20
Misc changes for Wildfly Swarm

bobmcwhirter:
* Adjust how content is provided in order to support multiple deployments.
* Use start() instead of bootstrap() so that quiesence is achieved.
* Rename (simplify) some items.
* Throw exceptions instead of abort()'ing, printing and exiting.
* Do not install StdioContext, to allow user's own
  System.err/System.out usage to continue as expected.

### DIFF
--- a/server/src/main/java/org/jboss/as/server/SelfContainedContainer.java
+++ b/server/src/main/java/org/jboss/as/server/SelfContainedContainer.java
@@ -24,26 +24,27 @@ package org.jboss.as.server;
 
 import org.jboss.as.controller.RunningMode;
 import org.jboss.as.controller.persistence.ExtensibleConfigurationPersister;
-import org.jboss.as.process.ExitCodes;
+import org.jboss.as.selfcontained.ContentProvider;
+import org.jboss.as.selfcontained.ContentProviderServiceActivator;
 import org.jboss.as.selfcontained.SelfContainedConfigurationPersister;
-import org.jboss.as.selfcontained.SelfContainedContentServiceActivator;
 import org.jboss.as.version.ProductConfig;
 import org.jboss.dmr.ModelNode;
 import org.jboss.modules.Module;
 import org.jboss.modules.ModuleIdentifier;
+import org.jboss.modules.ModuleLoadException;
 import org.jboss.msc.service.ServiceActivator;
+import org.jboss.msc.service.ServiceContainer;
 import org.jboss.stdio.LoggingOutputStream;
 import org.jboss.stdio.NullInputStream;
 import org.jboss.stdio.SimpleStdioContextSelector;
 import org.jboss.stdio.StdioContext;
 import org.wildfly.security.manager.WildFlySecurityManager;
 
-import java.io.File;
-import java.io.PrintStream;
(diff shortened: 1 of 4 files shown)


server/src/main/java/org/jboss/as/server/SelfContainedContainer.java
```diff
@@ -67,58 +62,31 @@ public SelfContainedContainer() {
...
-                } catch (Throwable ignored) {
-                }
-                // Install JBoss Stdio to avoid any nasty crosstalk, after command line arguments are processed.
-                StdioContext.install();
-                final StdioContext context = StdioContext.create(
-                        new NullInputStream(),
-                        new LoggingOutputStream(org.jboss.logmanager.Logger.getLogger("stdout"), org.jboss.logmanager.Level.INFO),
-                        new LoggingOutputStream(org.jboss.logmanager.Logger.getLogger("stderr"), org.jboss.logmanager.Level.ERROR)
-                );
-                StdioContext.setStdioContextSelector(new SimpleStdioContextSelector(context));
```
jamezp: Is there a reason to remove the stdout and stderr from redirecting to a logger? Anything written to `System.out` or `System.err` will not be redirected to the `server.log` if we remove this section.

bobmcwhirter: Since this is used from a user-provided main(), I figured the user may end up being surprised if stdout/stderr end up going through logging, especially if those log categories are suppressed.

jamezp: The logging subsystem also [installs STDIO](https://github.com/wildfly/wildfly-core/blob/master/logging/src/main/java/org/jboss/as/logging/stdio/LogContextStdioContextSelector.java#L48-L64). We might need to do something there as well. We should note too anything that writes to `System.out` and `System.err` will not behave the same as a normal server as well.

bobmcwhirter: fwiw, to be perfectly honest, when I installed StdioContext, in my use-case, stdout/stderr all went to /dev/null, even with the categories turned on.  Maybe I should dig deeper and figure out why they ended up in the ether.

jamezp: Not even to the server.log? Though if you printed to it before WildFly was launched it should work. We could potentially get into some interesting scenarios with logging too. For example if the uses logging before they launch WildFly then the logging subsystem won't work.

bobmcwhirter: Before launching (or the installation of StdioContext in this method), it'd work.

After the installation of the StdioContext, then it stopped entirely.  My logging subsystem is purely CONSOLE-centric and things would still go awol.  Might be misconfiguration on my part, though. I'll continue to investigate.

bobmcwhirter: So, I'm an idiot.  Everything works as expected.  I'll append to the PR.

PR #634, opened 2015-04-21
[WFCORE-653] Make json.pretty=1 possible

kabir-ncl:


kabir-ncl:

kabir:

kabir:
This is ok to test

PR #637, opened 2015-04-22
[WFCORE-428] Enabled module add/remove from user-provided module paths

Gonthim:
https://issues.jboss.org/browse/WFCORE-428


Gonthim:
retest this please

Gonthim:
ggrrrr

Retest this please

bstansberry:
Do not merge until we decide how this new capability fits into our long term strategy for this command. The current implementation of the module command is broken as it relies on local filesystem access and doesn't work in the domain.

Unfortunately, where I'm working today I don't seem to be able to access jboss.org so I can't look at the JIRA to see or participate in any discussion that may be referenced there.

Gonthim:
I'll withdraw this then and start a conversation on the dev list. Seems there are a few issues that have been waiting on such a discussion, so doing the refactor first would be better than adding more functionality to refactor later.

PR #638, opened 2015-04-23
[WFCORE-659] standalone.sh fails to start server with debug argument

TomasHofman:
...on solaris 10 and HP-UX hosts

Solaris sh doesn't seem to support variety of substitutions.

### DIFF
--- a/core-feature-pack/src/main/resources/content/bin/standalone.sh
+++ b/core-feature-pack/src/main/resources/content/bin/standalone.sh
@@ -13,9 +13,9 @@ do
     case "$1" in
       --debug)
           DEBUG_MODE=true
-          shift
-          if [ -n "$1" ] && [ "${1#*-}" = "$1" ]; then
-              DEBUG_PORT=$1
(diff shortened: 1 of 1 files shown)


ctomc:
retest this please

ctomc:
@jamezp can you take a look at this?

jamezp:
I'm kind of inclined to not worry about this fix. While it's convenient I'm not a big fan of the `--debug` argument.

If we do fix it I think we should just remove the shift from `--debug` case and rely on one at the bottom. Then change the port check to use `$2` and that should fix it.

TomasHofman:
There must be an extra shift, otherwise the optional port parameter will get included in SERVER_OPTS and cause an exception.

However, it could be moved into the if-block that is checking whether next parameter is port number:

case "$1" in
      --debug)
          DEBUG_MODE=true
          if [ -n "$2" ] && [ "${2#*-}" = "$2" ]; then
              DEBUG_PORT=$2
              shift
          fi
          ;;

and then the last shift could be left without checking as it was.

TomasHofman:
PR updated accordingly.

PR #639, opened 2015-04-23
Launcher changes

jamezp:
Some general clean up and checking before adding the `XX:MaxPermSize` by default.

### DIFF
--- a/launcher/src/main/java/org/wildfly/core/launcher/AbstractCommandBuilder.java
+++ b/launcher/src/main/java/org/wildfly/core/launcher/AbstractCommandBuilder.java
@@ -27,6 +27,7 @@ import java.nio.file.Files;
 import java.nio.file.Path;
 import java.nio.file.Paths;
 import java.util.ArrayList;
+import java.util.Collection;
 import java.util.Iterator;
 import java.util.List;
 import java.util.Locale;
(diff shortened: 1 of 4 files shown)


launcher/src/main/java/org/wildfly/core/launcher/AbstractCommandBuilder.java
```diff
@@ -44,18 +46,28 @@
...
+        // Default JVM parameters for all versions
+        javaOpts.add("-Xms64m");
+        javaOpts.add("-Xmx512m");
+        javaOpts.add("-Djava.net.preferIPv4Stack=true");
+        javaOpts.add("-Djava.awt.headless=true");
+        javaOpts.add("-Djboss.modules.system.pkgs=org.jboss.byteman");
+
+        // Versions below 8 should add a MaxPermSize
+        if (VersionComparator.compareVersion(jvmVersion, "1.8") < 0) {
+            javaOpts.add("-XX:MaxPermSize=256m");
```
ctomc: :+1:

launcher/src/main/java/org/wildfly/core/launcher/VersionComparator.java
```diff
@@ -0,0 +1,99 @@
...
+/**
+ * Compares 2 different version strings.
+ * <p/>
+ * The version strings must be valid integers separated by {@code .} (dots).
+ * <p/>
+ * Date: 09.11.2011
+ *
+ * @author <a href="mailto:jperkins@redhat.com">James R. Perkins</a>
+ */
+public class VersionComparator implements Comparator<String> {
```
ctomc: take a look at JvmOptionsBuilderFactory.SunJvmOptionsBuilder that solution satisfied also David :)

bstansberry:
Re: the first commit, have you checked with whoever the known users are re: potential breakage from this change?

I think it's good and we don't want to leave it out of 9 and then make a breaking change in 10. But if anyone will be affected, giving them a heads up would be good.

TBH I doubt anyone uses this. :)

jamezp:
Yeah I should have linked the [JIRA](https://issues.jboss.org/browse/WFLY-4552). I'm not sure why I didn't clarify in my comments, but 9 is the first release with the launcher so I think we should remove them before final.

And yeah I doubt anyone uses it except for me :)

launcher/src/main/java/org/wildfly/core/launcher/VersionComparator.java
jamezp: Yeah. That only checks for version 1.x. Which for now is fine, but this will compare any versions in the future. Who knows when we'll see Java 2.0, but you never know :)

PR #641, opened 2015-04-24
Add EAP 6.4 as transformer target

ctomc:


kabir:
https://github.com/wildfly/wildfly-core/pull/642 superseded by https://github.com/wildfly/wildfly-core/pull/642

PR #646, opened 2015-04-27
[WFCORE-665] When host jvm name and server jvm name are same then the ho...

jaysensharma:
...st jvm settings are ignored

https://issues.jboss.org/browse/WFCORE-665
When host jvm name and server jvm name are same then the host jvm settings are ignored


jaysensharma:
retest this please

bstansberry:
Thanks for the patch. I prefer that any merging of data from different <jvm> elements be done in the JVMElement class, with ManagedServerBootCmdFactory just concerned with identifying what elements are relevant and determining precedence.

I'm also not convinced this is a bug. AIUI the software works as designed. It seems that what's wanted it a change in behavior. There's certainly a case for that though.

jamezp:
Given the failures, Brian's comments and the fact we're getting close to a core release I'm closing this.

PR #647, opened 2015-04-27
Attr group

tadamski:
I have refactored iiop subsystem to use flat attribute groups, but encoutered following problems:
Bugs: subsystem tag is closed if there are grouped attributes but no children, no exception is thrown when invalid task is read but there are no children. I am proposing fix to those two in first commit.
Properties: if there is properties argument it has to be defined in the subsystem. There is property argument in iiop subsystem but it not mandatory and is not used in standard configuration. I am proposing the change in which properties are parsed/persisted separately and as a result of that: the don't have to be present in subsystem, they can be put in arbitrary place within subsystem argument tags, wrapper properties are parsed correctly. This is a second commit.


ctomc:
I agree with all this fixes, but the property fix is really just one off, as we need proper solution for all multi element attributes. But anyhow, for time being this is more than welcome.

However could you please add tests for this. Easiest way to do that is to add extra checks to PersistentResourceXMLParserTestCase

tadamski:
Tests added.

kabir:
Merged

PR #652, opened 2015-04-28
Update default logging subsystem config to be more in line with current code

ctomc:
no need to have mgmt config logger set to debug,
we also don't use jacorb & tomcat/jbossweb anymore.

### DIFF
--- a/logging/src/main/resources/subsystem-templates/logging.xml
+++ b/logging/src/main/resources/subsystem-templates/logging.xml
@@ -35,24 +35,12 @@
    <supplement name="default">
       <replacement placeholder="@@append@@" attributeValue="true"/>
       <replacement placeholder="LOGGERS">
-	      <logger category="com.arjuna">
-	          <level name="WARN"/>
-	      </logger>
-	      <logger category="org.apache.tomcat.util.modeler">
-	          <level name="WARN"/>
-	      </logger>
-	      <logger category="org.jboss.as.config">
-	          <level name="DEBUG"/>
-	      </logger>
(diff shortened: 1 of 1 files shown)


logging/src/main/resources/subsystem-templates/logging.xml
```diff
@@ -35,24 +35,12 @@
    <supplement name="default">
       <replacement placeholder="@@append@@" attributeValue="true"/>
       <replacement placeholder="LOGGERS">
-	      <logger category="com.arjuna">
-	          <level name="WARN"/>
-	      </logger>
-	      <logger category="org.apache.tomcat.util.modeler">
-	          <level name="WARN"/>
-	      </logger>
-	      <logger category="org.jboss.as.config">
```
jamezp: This one should probably stay.

logging/src/main/resources/subsystem-templates/logging.xml
```diff
@@ -35,24 +35,12 @@
...
-	          <level name="DEBUG"/>
-	      </logger>
-	      <logger category="sun.rmi">
-	          <level name="WARN"/>
-	      </logger>
-	      <logger category="jacorb">
-	          <level name="WARN"/>
-	      </logger>
-	      <logger category="jacorb.config">
```
jamezp: For both jacorb ones we need to determine if we should add new categories.

jamezp:
The various logging.properties files also need to be updated too.

ctomc:
will work on this when we move to 2.x

PR #656, opened 2015-04-28
[WFCORE-558] Drop common-core dependency in org.jboss.vfs module.xml

bstansberry:
@ctomc I noticed this dependency is still in the module.xml but I don't think it's needed.


ctomc:
@bstansberry you are right. Didn't even check as we are dropping common-core early in 10 lifecycle.

ctomc:
retest this please

PR #657, opened 2015-04-29
[WFCORE-230] : Failure by ProcessController to launch a server vm is not logged

ehsavoie:
Tracing such failures.

Jira: https://issues.jboss.org/browse/WFCORE-230


bstansberry:
This looks good, but I'd prefer it if the test were part of one of our existing suites, probably DomainTestSuite. I easily could be wrong, but other than the debug stuff in the HC configs (which doesn't seem to be used), the tests aren't doing anything that couldn't be done in a @BeforeClass / @AfterClass.

I prefer using the suites because starting and stopping HCs is time consuming.

bstansberry:
As we discussed in chat, the test stuff can wait.

PR #661, opened 2015-04-29
[WFCORE-668] Allow devs to turn of INFO logging re specific deprecated a...

bstansberry:
...ttributes and operations

This relates to https://bugzilla.redhat.com/show_bug.cgi?id=1213421, where we're spamming the EAP logs about deprecated attributes that the user has no choice but to use in those releases. We even use them in our standard configs. This PR lets devs selectively turn off such logging when it adds no value for the attribute/operation in question. The EAP fix would backport this and then turn off the logging for the noisy attributes.

This also improves the log messages.

### DIFF
--- a/controller/src/main/java/org/jboss/as/controller/AbstractAttributeDefinitionBuilder.java
+++ b/controller/src/main/java/org/jboss/as/controller/AbstractAttributeDefinitionBuilder.java
@@ -462,12 +462,34 @@ public abstract class AbstractAttributeDefinitionBuilder<BUILDER extends Abstrac
     }
 
     /**
-     * Marks the attribute as deprecated since the given API version.
+     * Marks the attribute as deprecated since the given API version. This is equivalent to calling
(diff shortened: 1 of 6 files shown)


jamezp:
Not sure this is needed here, but there is a new `@LogOnce` annotation for logging that will only a log message once.

bstansberry:
@jamezp once for a given message method, or once per set of params?

The latter sounds a bit unlikely as that would imply caching param values, possible classloader leaks, etc.

jamezp:
@bstansberry Sorry just noticed this for some reason. Once per message. So once it's logged once it won't be logged again until a restart.

bstansberry:
@jamezp ok, thanks, but that wouldn't be helpful here. This particular message can validly be logged multiple times with different parameters.

In general too WildFly devs should be very careful with that feature, as a reload is not a full vm restart, but for lots of behaviors people would expect it to act like a full restart. Including IMO logging.

PR #662, opened 2015-04-29
[WFCORE-639] : ManagementPermissionAuthorizer is limited to the standard roles for its authorizeJmxOperation impl

ehsavoie:
Removing the JmxPermissionFactory as it has no real value.
Adding tests in domain mode.

Jira: https://issues.jboss.org/browse/WFCORE-639

https://github.com/wildfly/wildfly/pull/7401 is required for full integration


controller/src/main/java/org/jboss/as/controller/access/JmxAction.java
```diff
@@ -47,11 +79,50 @@ public Impact getImpact() {
...
+
+    public ModelNode getCurrentValue() {
+        return currentValue;
+    }
+
+    public List<AccessConstraintDefinition> getConstraints() {
+        return constraints;
+    }
+
+    public boolean isIsNonFacadeMBeansSensitive() {
```
bstansberry: The method name has an extra "Is".

controller/src/test/java/org/jboss/as/controller/access/permission/ManagementPermissionAuthorizerTestCase.java
```diff
@@ -155,13 +157,18 @@ public PermissionCollection getRequiredPermissions(Action action, TargetResource
             return getRequiredPermissions(action);
         }
 
+        //@Override
+        public Set<String> getUserRoles(Caller caller, Environment callEnvironment, Action action, TargetResource target) {
+            return null;
+        }
```
bstansberry: Is this needed any more?

jmx/src/main/java/org/jboss/as/jmx/PluggableMBeanServerImpl.java
```diff
@@ -1229,11 +1280,10 @@ void log(boolean readOnly, Throwable error, String methodName, String[] methodSi
         }
     }
 
-    boolean authorizeSensitiveOperation(String methodName, boolean readOnly, boolean exception) throws MBeanException {
+    boolean authorizeSensitiveOperation(JmxTarget target, JmxAction action, boolean exception) throws MBeanException {
```
bstansberry: Not caused by this PR, but this method should be private.

jmx/src/main/java/org/jboss/as/jmx/PluggableMBeanServerImpl.java
```diff
@@ -1245,10 +1295,10 @@ boolean authorizeSensitiveOperation(String methodName, boolean readOnly, boolean
         return true;
     }
 
-    boolean authorizeSuperUserOrAdministrator(String methodName) throws MBeanException {
+    boolean authorizeSuperUserOrAdministrator(JmxTarget target , String methodName) throws MBeanException {
```
bstansberry: Not caused by this PR, but this method should be private.

It should be renamed too, as the "SuperUserOrAdministrator" bit is an implementation detail of a particular rbac provider.

server/src/main/java/org/jboss/as/server/jmx/MBeanServerPlugin.java
```diff
@@ -51,4 +53,14 @@
...
+
+    /**
+     * Return the operation target for calling this method on this object.
+     * @param name
+     * @param method
+     * @return
+     */
+    JmxTarget getJmxTarget(final ObjectName name, final String method);
+
+    JmxAction getJmxAction(final ObjectName name, final String method, String operation, JmxAction.Impact impact, boolean isNonFacadeMBeansSensitive, String attribute);
```
bstansberry: The javadoc should include this

* @throws IllegalStateException if {@link #shouldAuthorize()} returns {@code false}

Also, please fill in the rest of the javadoc.

server/src/main/java/org/jboss/as/server/jmx/MBeanServerPlugin.java
```diff
@@ -51,4 +53,14 @@
...
      */
     boolean shouldAuthorize();
+
+    /**
+     * Return the operation target for calling this method on this object.
+     * @param name
+     * @param method
+     * @return
+     */
+    JmxTarget getJmxTarget(final ObjectName name, final String method);
```
bstansberry: The javadoc should include this

* @throws IllegalStateException if {@link #shouldAuthorize()} returns {@code false}

Also, please fill in the rest of the javadoc.

bstansberry: Actually, UnsupportedOperationException is better.

server/src/main/java/org/jboss/as/server/jmx/MBeanServerPlugin.java
bstansberry: Actually, UnsupportedOperationException is better.

jmx/src/main/java/org/jboss/as/jmx/model/ModelControllerMBeanServerPlugin.java
```diff
@@ -246,20 +266,47 @@ public void removeNotificationListener(ObjectName name, NotificationListener lis
...
         throw new RuntimeOperationsException(JmxLogger.ROOT_LOGGER.removeNotificationListerWithObjectNameNotSupported(listener));
     }
 
+    @Override
     public void removeNotificationListener(ObjectName name, ObjectName listener) throws InstanceNotFoundException, ListenerNotFoundException {
         throw new RuntimeOperationsException(JmxLogger.ROOT_LOGGER.removeNotificationListerWithObjectNameNotSupported(listener));
     }
 
+    @Override
+    public JmxTarget getJmxTarget(final ObjectName name, final String method) {
```
bstansberry: This should just throw UnsupportedOperationException. This class returns false from shouldAuthorize(), so it shouldn't attempt to do authorization tasks.

jmx/src/main/java/org/jboss/as/jmx/model/ModelControllerMBeanServerPlugin.java
```diff
@@ -246,20 +266,47 @@ public void removeNotificationListener(ObjectName name, NotificationListener lis
...
+        ModelControllerMBeanHelper helper = getHelper(name);
+        PathAddress address = name != null ? helper.resolvePathAddress(name) : PathAddress.EMPTY_ADDRESS;
+        return new JmxTarget(method, name, address, helper.resolveConstraints(address));
+    }
+
+    @Override
+    public JmxAction getJmxAction(final ObjectName name, final String method, String operation, JmxAction.Impact impact, boolean isNonFacadeMBeansSensitive, String attributeName) {
```
bstansberry: This should just throw UnsupportedOperationException. This class returns false from shouldAuthorize(), so it shouldn't attempt to do authorization tasks.

For the JMX domains handled by this MBeanServerPlugin, the standard authz from the ModelController, OperationContext etc will happen, so there is no anything in the JMX layer.

With that change, a whole lot of stuff can presumably come out.

controller/src/main/java/org/jboss/as/controller/access/JmxTarget.java
```diff
@@ -0,0 +1,62 @@
...
+ * @author <a href="mailto:ehugonne@redhat.com">Emmanuel Hugonnet</a> (c) 2015 Red Hat, inc.
+ */
+public class JmxTarget {
+
+    private final String method;
+    private final ObjectName objectName;
+    private final PathAddress address;
+    private final List<AccessConstraintDefinition> accessConstraintDefinitions;
+
+    public JmxTarget(String method, ObjectName objectName, PathAddress address, List<AccessConstraintDefinition> accessConstraintDefinitions) {
```
bstansberry: A JMXTarget doesn't have a PathAddress, as it never refers to a management model Resource.

bstansberry: In theory, it could have something like an AccessConstraintDefinition, but I don't believe this will ever be anything but an empty list once ModelControllerMBeanServerPlugin is changed to not try and support the JMX-layer authz. And if we ever do try to configure some sort access constraint configurability for mbeans, it's likely the interfaces involved would not exactly be AccessConstraintDefinition. So the accessConstraintDefinitions param/getter isn't needed now and isn't a good candidate for including for future use.

controller/src/main/java/org/jboss/as/controller/access/JmxAction.java
```diff
@@ -28,14 +37,37 @@
...
+    }
+
+    public JmxAction(String methodName, String operation, Impact impact, boolean isNonFacadeMBeansSensitive, Attribute attribute) {
+        this(methodName, operation, impact, isNonFacadeMBeansSensitive, null, attribute);
+    }
+
+    public JmxAction(String methodName, String operation, Impact impact, boolean isNonFacadeMBeansSensitive, List<AccessConstraintDefinition> constraints, Attribute attribute) {
         this.methodName = methodName;
         this.impact = impact;
+        this.operation = operation == null ? "test" : operation;
```
bstansberry: Is operation ever different from methodName once ModelControllerMBeanServerPlugin is changed to no longer support authz?

Re: the AccessConstraintDefinitions, same comment as JMXTarget applies.

I think isNonFacadeMBeanSensitive fits better in JMXTarget. It relates to targets more than actions.

testsuite/domain/src/test/java/org/jboss/as/test/integration/domain/rbac/JmxRBACProviderServerGroupScopedRolesTestCase.java
```diff
@@ -0,0 +1,311 @@
...
+import org.junit.BeforeClass;
+import org.junit.Test;
+import org.wildfly.test.jmx.JMXServiceDeploymentSetupTask;
+
+/**
+ * Tests of server group scoped roles using the "rbac" access control provider.
+ *
+ * @author Brian Stansberry (c) 2013 Red Hat Inc.
+ */
+public class JmxRBACProviderServerGroupScopedRolesTestCase extends AbstractServerGroupScopedRolesTestCase {
```
bstansberry: This test validates the expected behavior of server-group-scoped roles when the user is in a role that allows access to the server (i.e. it's in the group). But it's not validating that users in scoped roles whose server groups don't include the target server are rejected.

We should add a test for host-scoped roles as well, as the same basic pattern should be covered.

controller/src/main/java/org/jboss/as/controller/access/constraint/ApplicationTypeConstraint.java
```diff
@@ -143,5 +149,10 @@ protected int internalCompare(AbstractConstraintFactory other) {
             // We have no preference
             return 0;
         }
+
+        @Override
+        public Constraint getRequiredConstraint(Action.ActionEffect actionEffect, JmxAction action, JmxTarget target) {
+            return (isApplicationType(action.composeAction()) || isApplicationType(target)) ? APPLICATION : NON_APPLICATION;
```
bstansberry: This can simply return NON_APPLICATION and isApplicationType() can be removed. The methods will always return false.

controller/src/main/java/org/jboss/as/controller/access/constraint/NonAuditConstraint.java
```diff
@@ -85,7 +87,11 @@ private boolean isAuditResource(TargetResource target) {
         protected int internalCompare(AbstractConstraintFactory other) {
             // We prefer going ahead of anything except a ScopingConstraint or AuditConstraint
             return (other instanceof ScopingConstraintFactory || other instanceof AuditConstraint.Factory)  ? 1 : -1;
+        }
 
+        @Override
+        public Constraint getRequiredConstraint(Action.ActionEffect actionEffect, JmxAction action, JmxTarget target) {
+            return isAuditOperation(action.composeAction()) ? AUDIT : NOT_AUDIT;
```
bstansberry: This can simply return NOT_AUDIT.

controller/src/main/java/org/jboss/as/controller/access/constraint/SensitiveTargetConstraint.java
```diff
@@ -149,5 +155,10 @@ protected int internalCompare(AbstractConstraintFactory other) {
             // We have no preference
             return 0;
         }
+
+        @Override
+        public Constraint getRequiredConstraint(Action.ActionEffect actionEffect, JmxAction action, JmxTarget target) {
+            return (action.getImpact() == JmxAction.Impact.EXTRA_SENSITIVE || target.isNonFacadeMBeansSensitive() || isSensitiveAction(action.composeAction(), actionEffect) || isSensitiveResource(target, actionEffect)) ? SENSITIVE : NOT_SENSITIVE;
```
bstansberry: isSensitiveAction() and isSensitiveResource() will always return false, so the calls can be dropped.

controller/src/main/java/org/jboss/as/controller/access/constraint/SensitiveVaultExpressionConstraint.java
```diff
@@ -139,6 +141,19 @@ protected int internalCompare(AbstractConstraintFactory other) {
             // We have no preference
             return 0;
         }
+
+        @Override
+        public Constraint getRequiredConstraint(Action.ActionEffect actionEffect, JmxAction action, JmxTarget target) {
+            //TODO fix me
+            if (VaultExpressionSensitivityConfig.INSTANCE.isSensitive(actionEffect)) {
+                if (actionEffect == Action.ActionEffect.WRITE_RUNTIME || actionEffect == Action.ActionEffect.WRITE_CONFIG) {
+                    if (action.getCurrentValue() != null && isSensitiveValue(action.getCurrentValue())) {
```
bstansberry: It looks like action.getCurrentValue() will always return null.

Trying to do this check is a good idea, but I think it should be commented out with a TODO until we can sort out how to cleanly deal with the value (which could be any type). That can be separate work.

controller/src/main/java/org/jboss/as/controller/access/JmxAction.java
```diff
@@ -47,11 +65,42 @@ public Impact getImpact() {
...
+        return attribute != null;
+    }
+
+    public String getAttributeName() {
+        return attribute;
+    }
+
+    public ModelNode getCurrentValue() {
+        return currentValue;
+    }
```
bstansberry: I'm not sure we should have this method if we can't properly populate the value. (AFAICT it's always 'null' now.)

In any case, it should return Object. We have no mechanism to consistently convert the arbitrary types JMX supports to ModelNode. The only use of this now is SensitiveVaultExpressionConstraint, and that code can do an "instanceof String" check and if true wrap the string in a ModelNode for the isSensitiveValue check.

controller/src/main/java/org/jboss/as/controller/access/rbac/DefaultPermissionFactory.java
```diff
@@ -109,12 +109,13 @@ public PermissionCollection getUserPermissions(Caller caller, Environment callEn
...
+    public PermissionCollection getUserPermissions(Caller caller, Environment callEnvironment, JmxAction action, JmxTarget target) {
+        return getUserPermissions(roleMapper.mapRoles(caller, callEnvironment, action == null ? null: action.composeAction(),(TargetResource) null));
     }
 
-    public boolean isNonFacadeMBeansSensitive() {
-        return authorizerConfiguration.isNonFacadeMBeansSensitive();
+    //@Override
+    public Set<String> getUserRoles(Caller caller, Environment callEnvironment, Action action, TargetResource target){
+        return roleMapper.mapRoles(caller, callEnvironment, action, target);
     }
```
bstansberry: This method is no longer used and should be removed.

jmx/src/main/java/org/jboss/as/jmx/model/ModelControllerMBeanHelper.java
```diff
@@ -172,11 +176,26 @@ public boolean onResource(ObjectName resourceName) {
         }).iterate();
     }
 
-
```
bstansberry: I believe all changes to this class can be reverted.

server/src/main/java/org/jboss/as/server/jmx/MBeanServerPlugin.java
```diff
@@ -51,4 +53,25 @@
...
      */
     boolean shouldAuthorize();
+
+    /**
+     * Return the operation target for calling this method on this object.
+     * @param name the name of the object target of the call.
+     * @param method the name of the method called.
+     * @param isNonFacadeMBeansSensitive
+     * @return a JmxTarget for the method inovkation on the specified MBean.
+     * @throws UnsupportedOperationException
```
bstansberry: @throws UnsupportedOperationException if {@link #shouldAuthorize()} returns {@code true}

server/src/main/java/org/jboss/as/server/jmx/MBeanServerPlugin.java
```diff
@@ -51,4 +53,25 @@
...
+    JmxTarget getJmxTarget(final ObjectName name, final String method, boolean isNonFacadeMBeansSensitive) throws UnsupportedOperationException;
+
+    /** Return the operation action for calling this method on this object.
+     * @param name the name of the object target of the call.
+     * @param method the name of the method called.
+     * @param operation the model operation name corresponding to this call.
+     * @param impact impact of this method call.
+     * @param attribute name of the object attribute subject of this call.
+     * @return
+     * @throws UnsupportedOperationException
```
bstansberry: @throws UnsupportedOperationException if {@link #shouldAuthorize()} returns {@code true}

controller/src/main/java/org/jboss/as/controller/access/JmxTarget.java
```diff
@@ -0,0 +1,52 @@
...
+    }
+
+    public ObjectName getObjectName() {
+        return objectName;
+    }
+
+    public boolean isNonFacadeMBeansSensitive() {
+        return isNonFacadeMBeansSensitive;
+    }
+}
```
bstansberry: This should provide these:

HostEffect getHostEffect()

ServerGroupEffect getServerGroupEffect()

Those can then drive HostEffectConstraint and ServerGroupEffectConstraint.

The trick then is to get the name of the host and the name of the server group in to PluggableMBeanServerImpl.

Possibly the HostEffect and ServerGroupEffect themselves can be provided instead of the details used to create them. Passing them via JmxAuthorizer is a simple solution.

On a domain server, the host name and server-group name are recorded in the root management resource. They are undefined of course in a standalone server. We should also check with Kabir Khan as to whether this code should function in an HC process as well (IIRC, his plan is to have the JMX subsystem function as a HC subsystem too), in which case, the host name is known in the model too, while the server group name is undefined.

jmx/src/main/java/org/jboss/as/jmx/JMXSubsystemAdd.java
```diff
@@ -72,6 +76,18 @@ static void launchServices(OperationContext context, ModelNode model, ManagedAud
         } else {
             isMasterHc = false;
         }
+        if (context.getProcessType() == ProcessType.HOST_CONTROLLER) {
```
bstansberry: I'm going to change this some. We need to handle ProcessType.DOMAIN_SERVER; this block looks about right for that case. Handling HOST_CONTROLLER is good if this is a host-controller subsystem. But there, server-group will always be null (as an HC is not in a server group).

controller/src/main/java/org/jboss/as/controller/access/rbac/DefaultPermissionFactory.java
```diff
@@ -109,12 +109,8 @@ public PermissionCollection getUserPermissions(Caller caller, Environment callEn
...
 
     @Override
-    public Set<String> getUserRoles(Caller caller, Environment callEnvironment, Action action, TargetResource target){
-        return roleMapper.mapRoles(caller, callEnvironment, action, target);
-    }
-
-    public boolean isNonFacadeMBeansSensitive() {
-        return authorizerConfiguration.isNonFacadeMBeansSensitive();
+    public PermissionCollection getUserPermissions(Caller caller, Environment callEnvironment, JmxAction action, JmxTarget target) {
+        return getUserPermissions(roleMapper.mapRoles(caller, callEnvironment, action == null ? null: action.composeAction(),(TargetResource) null));
```
bstansberry: I'm going to add a method to RoleMapper for this.

jmx/src/main/java/org/jboss/as/jmx/JMXSubsystemAdd.java
ehsavoie: Yes this was for DOMAIN_SERVER not HC (i did something wrong with cleaning the code)

bstansberry:
This looks good; just needs a JmxRBACProviderHostScopedRolesTestCase.

PR #663, opened 2015-04-29
Testsuite code cleanup & stop using deprecated api

ctomc:


ctomc:
do not merge! errors are valid.

PR #668, opened 2015-05-01
update powershell scripts to be more in line of others

ctomc:
far from perfect, but on a good way to replace our ratchet .bat scripts

It does passes much better chunk of scripts testsuite than batch scripts do.

### DIFF
--- a/core-feature-pack/src/main/resources/content/bin/jboss-cli.ps1
+++ b/core-feature-pack/src/main/resources/content/bin/jboss-cli.ps1
@@ -1,6 +1,6 @@
-### -*- Power Shell file -*- ################################################
+#############################################################################
 #                                                                          ##
-#    JBoss CLI Script for interacting with the standalone server           ##
+#    WildFly CLI Script for interacting with the server                    ##
 #                                                                          ##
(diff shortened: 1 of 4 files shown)


core-feature-pack/src/main/resources/content/bin/jboss-cli.ps1
```diff
@@ -1,6 +1,6 @@
-### -*- Power Shell file -*- ################################################
+#############################################################################
 #                                                                          ##
-#    JBoss CLI Script for interacting with the standalone server           ##
+#    WildFly CLI Script for interacting with the server                    ##
```
bstansberry: This kind of thing just makes maintaining EAP a PITA, as we are certain to get complaints leading to silly code differences between branches. Let's just get rid of the name altogether in these. If people are confused about what software the script is for, they probably shouldn't touch them anyway. :)

bstansberry:
Please send up a separate PR to get rid of the "WildFly" stuff. Thanks for giving these some love!

ctomc:
will do, scripts need some more work as well, as proper domain.ps1 script is still missing

PR #675, opened 2015-05-05
https://issues.jboss.org/browse/WFCORE-672 CLI tab-completion for attrib...

aloubyansky:
...ute name path syntax

Thanks,
Alexey


cli/src/main/java/org/jboss/as/cli/impl/AttributeNamePathCompleter.java
```diff
@@ -0,0 +1,478 @@
...
+                return Collections.emptyList();
+            }
+
+            ModelNode typeDescr = attrsDescr;
+            for(String name : path) {
+                if(!typeDescr.has(name)) {
+                    return Collections.emptyList();
+                }
+                final ModelNode descr = typeDescr.get(name);
+                if(writeOnly && isWritable(descr)) {
```
bstansberry: Is this correct? Should it be this?

```
if(writeOnly && !isWritable(descr)) {
```

aloubyansky: You are right, thanks a lot! Fixed it.

PR #676, opened 2015-05-05
https://issues.jboss.org/browse/WFCORE-670 Node type is not complete at ...

aloubyansky:
...index 0 after tab completion on an empty resource

Thanks,
Alexey


jamezp:
Tested locally and this works perfectly.

PR #678, opened 2015-05-05
Capabilities and requirements improvements

bstansberry:
Make the API more user-friendly for the common case of a capability providing a service and the requirer needing an injection.

Support capabilities with dynamic components to their names.

Allow declaration of capability requirements in the AttributeDefinition; use that data in the add and write-attribute handlers to automatically wire up the requirements.

Use the above for a couple use cases -- JMX, Remoting and IO.

Next step on this is dealing with socket binding capabilities, which have a bit of a twist due to the presence of multiple socket binding groups in a domain config and the need to verify that all refs in a profile can be resolved against at least on socket-binding-group. But I'll do that as a later PR as what I have here has been useful to Darran in the Elytron work.


bstansberry:
Sweet! The test failures prove the need for this work! 3 places in full, subsystems that need JMX are doing stuff like this:

```
private static final ServiceName MBEAN_SERVER_NAME = ServiceName.JBOSS.append("mbean", "server");
```

Hard coding assumed service naming patterns from another subsystem into their own code.

I'll add a commit registering under the old name as an alias, to be removed once this is integrated into full and my branch converting the offending subsystems is merged too.

darranl:
Have been making use of this with the new Elytron subsystem for dynamic capabilities and requirements and so far it is working nicely.

ctomc:
This is a big improvement over what we had currently. I am not completely sold on the on AttributeDefinition <-- --> capability relation, but cannot really find anything strongly against it.

There is still room for improvement in area of injecting services via capabilities as it is still to verbose with this PR. More verbose == more problems with external users trying to write extension that use this.
But this can be improved upon in later PRs.
In short go ahead with merging this so we can get Elytron in, which will bring another case where we can see how it all behaves and how can be improved.

darranl:
I have added this PR to Brian's branch https://github.com/bstansberry/wildfly-core/pull/2, either Brian can merge it with this PR or I will send in a PR to core after this is merged.

PR #679, opened 2015-05-05
[WFCORE-114] : Get RuntimeVaultReader into WildFly Core

bstansberry:
Moving RuntimeVaultReader from WildFly to WildFly-Core

Reopening #620 as having this does good and doesn't mean the picketbox lib must be present in core.

Right now (w/o this PR) the way it works is the service loader is in org.jboss.as.security which is visible because HC and server module.xml optionally depend on org.jboss.as.security, with services=import.

With this PR the service loader is in server, which HC already depends on, but now I added services=import too. So the core modules no longer need to depend on org.jboss.as.security as they get nothing from it.

The vault doesn't actually work in core, because there is no picketbox module but server module has an optional dep on picketbox, so if it is present (i.e. in full) then it just works.

I will need to follow up with a PR for full to delete the old org.jboss.as.security RuntimeVaultReader and its service loader file but that's just a cleanup of cruft; it's dead code now because core modules no longer make that stuff visible.

### DIFF
--- a/core-feature-pack/src/main/resources/modules/system/layers/base/org/jboss/as/host-controller/main/module.xml
+++ b/core-feature-pack/src/main/resources/modules/system/layers/base/org/jboss/as/host-controller/main/module.xml
@@ -59,8 +59,7 @@
         <module name="org.jboss.as.protocol"/>
         <module name="org.jboss.as.remoting"/>
         <module name="org.wildfly.security.manager" services="import"/>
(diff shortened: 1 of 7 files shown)


bstansberry:
This doesn't really help, as this class is loaded via ServiceLoader and there's no hook for that. So a user couldn't add support simply by installing the picketbox module.

PR #691, opened 2015-05-06
[WFCORE-677] Tighten up the EmbeddedServerFactory API

bstansberry:
1.x variant.

### DIFF
--- a/embedded/src/main/java/org/wildfly/core/embedded/EmbeddedServerFactory.java
+++ b/embedded/src/main/java/org/wildfly/core/embedded/EmbeddedServerFactory.java
@@ -22,27 +22,25 @@
 
 package org.wildfly.core.embedded;
 
-import org.wildfly.core.embedded.logging.EmbeddedLogger;
+import java.io.File;
+import java.lang.reflect.InvocationTargetException;
+import java.lang.reflect.Method;
+import java.util.Map;
+import java.util.Properties;
+import java.util.logging.LogManager;
+
 import org.jboss.modules.Module;
 import org.jboss.modules.ModuleClassLoader;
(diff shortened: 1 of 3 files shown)


embedded/src/main/java/org/wildfly/core/embedded/EmbeddedServerFactory.java
```diff
@@ -71,17 +69,53 @@
...
+    /**
+     * Create an embedded standalone server with modules loaded from the standard location in the server installation.
+     *
+     * @param jbossHomePath the location of the root of server installation. Cannot be {@code null} or empty.
+     * @param systemPackages names of any packages that must be treated as system packages, with the same classes
+     *                       visible to the caller's classloader visible to server-side classes loaded from
+     *                       the server's modular classlaoder
+     * @return the server. Will not be {@code null}
+     */
+    public static StandaloneServer create(String jbossHomePath, String... systemPackages) {
```
jamezp: This one should probably be removed. Using two String parameters here will result in the `public static StandaloneServer create(String jbossHomePath, String modulePath, String... systemPackages` being invoked.

bstansberry: +1. I need to listen to the little voice in my head that says "don't add unused API."

ctomc: (in low whisper) Brian: "don't add unused API" 
 "don't add unused API" 
 "don't add unused API" 
 "don't add unused API" 
 "don't add unused API" 
 "don't add unused API" 
 "don't add unused API" 

:)

PR #694, opened 2015-05-06
[WFCORE-679] Be able to check at runtime whether resource transformation

kabir:
should be discarded or rejected.

https://issues.jboss.org/browse/WFCORE-679


controller/src/main/java/org/jboss/as/controller/transform/description/DynamicDiscardPolicy.java
```diff
@@ -0,0 +1,41 @@
...
+
+import org.jboss.as.controller.PathAddress;
+import org.jboss.as.controller.transform.TransformationContext;
+
+/**
+ * @author <a href="mailto:kabir.khan@jboss.com">Kabir Khan</a>
+ */
+public interface DynamicDiscardPolicy {
+    /**
+     * Checks whether the child should be added
```
bstansberry: s/added/discarded/g ?

bstansberry:
My javadoc comment is a nit. Looks good.

PR #698, opened 2015-05-07
WFCORE-678 console-enabled="false" does not work in domain mode

luck3y:
Fixes WFCORE-678, and allows the console to be disabled in domain mode....


bstansberry:
This one also needs the context.getServiceRegistry(true) change.

bstansberry:
Closing in favor of #702.

PR #707, opened 2015-05-09
WFCORE-685 Speed up PathElement Validation

marschall:
During benchmarking of our WildFly instance creating the regex matcher for
controller path element validation has popped up.

 * implement path element validation using a one pass iteration over all
   characters rather than a regex
 * do not call #toCharArray as it allocates a new char[]

We did some microbenchmarks with JMH and doing the validation this way
is more than an order of magnitude faster.

Issue: WFCORE-685
https://issues.jboss.org/browse/WFCORE-685

### DIFF
--- a/controller/src/main/java/org/jboss/as/controller/PathElement.java
+++ b/controller/src/main/java/org/jboss/as/controller/PathElement.java
@@ -22,8 +22,6 @@
 
 package org.jboss.as.controller;
 
-import java.util.regex.Pattern;
-
 import org.jboss.as.controller.logging.ControllerLogger;
 import org.jboss.dmr.ModelNode;
 import org.jboss.dmr.Property;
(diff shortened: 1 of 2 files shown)


kabir:
Retest this please

dmlloyd:
Does the performance of this code cause some real-world issue?  If so, under what circumstances?

marschall:
It does not cause a real world performance issue. It just popped up during a load test and was easy to fix.

controller/src/main/java/org/jboss/as/controller/PathElement.java
```diff
@@ -110,6 +102,53 @@ public static PathElement pathElement(final String key, final String value) {
...
+    private static boolean isValidKey(final String s) {
+        // Equivalent to this regex \*|[_a-zA-Z](?:[-_a-zA-Z0-9]*[_a-zA-Z0-9]) but faster
+        if (s == null) {
+            return false;
+        }
+        if (s.equals(WILDCARD_VALUE)) {
+            return true;
+        }
+        int lastIndex = s.length() - 1;
+        if (!isValidKeyStartCharacter(s.charAt(0))) {
```
bstansberry: This will fail if the key is an empty string.

marschall:
I fixed the empty string issue and added a test case.

marschall:
We did some more investigation. Method seems to be mostly called by the deployment scanner. In production you'll likely have the deployment scanner disabled or you spend most of the work in business code.

The following is an extreme example with a WildFly that is mostly idling and almost 6% of all allocations made are Matchers in PathElement. As I said this would be much lower if you actually did any work.

pattern-usage

dmlloyd:
I think that we can probably consider that to be substantial & relevant.

bstansberry:
Thanks for the chart; it's very helpful in pointing out a good area to look to get rid of calls to PathAddress.pathAddress.

bstansberry:
Thanks!

PR #709, opened 2015-05-11
[WFCORE-645] Version 4 of the schema and management API.

darranl:


darranl:
This failure appears to be on the WildFly side which is going to require a core update followed by this PR https://github.com/wildfly/wildfly/pull/7444

domain-management/src/main/java/org/jboss/as/domain/management/logging/DomainManagementLogger.java
```diff
@@ -1127,6 +1127,15 @@
     String filePermissionsProblemsFound(String file);
 
     /**
+     * A message indicating an unsupported resource in the model during marshalling.
+     *
+     * @param name the name of the resource.
+     * @return The exception for the error.
+     */
+    @Message(id = 108, value = "Unsupported resource '%s'")
+    IllegalStateException unsupportedResource(String name);
```
bstansberry: I don't see this being used anywhere.

bstansberry:
retest this please

I've disabled StandardConfigsXMLValidationUnitTestCase until we get this processed.

domain-management/src/main/java/org/jboss/as/domain/management/logging/DomainManagementLogger.java
darranl: It's in ManagementXml_4.java - for some reason GitHub is not expanding that one by default.

PR #723, opened 2015-05-13
Drop wildfly-security-manager, use Elytron version instead

dmlloyd:
Drop the wildfly-security-manager artifact.  Redefine security manager module to re-export Elytron APIs.  Move all internal modules to point at the Elytron module directly.


dmlloyd:
Updated to keep the artifact around.  We will drop the artifact after the next core release, so that WildFly proper can continue to compile until then.

PR #728, opened 2015-05-14
[master] WFCORE-692 read-children-names(include-aliases=true) can return phantom results

pferraro:
… if another resource exists with the same key as tje target address of a resource alias.

https://issues.jboss.org/browse/WFCORE-692

### DIFF
--- a/controller/src/main/java/org/jboss/as/controller/operations/global/GlobalOperationHandlers.java
+++ b/controller/src/main/java/org/jboss/as/controller/operations/global/GlobalOperationHandlers.java
@@ -683,12 +683,16 @@ public class GlobalOperationHandlers {
                     }
                 }
             } else {
-                //PathAddress target = aliasEntry.getTargetAddress();
                 PathAddress target = aliasEntry.convertToTargetAddress(addr.append(element));
                 PathAddress targetParent = target.subAddress(0, target.size() - 1);
                 Resource parentResource = context.readResourceFromRoot(targetParent, false);
-                if (parentResource != null && parentResource.hasChildren(target.getLastElement().getKey())) {
-                    set.add(element.getValue());
+                if (parentResource != null) {
+                    PathElement targetElement = target.getLastElement();
+                    if (targetElement.isWildcard()) {
+                        set.addAll(parentResource.getChildrenNames(targetElement.getKey()));
+                    } else if (parentResource.hasChild(targetElement)) {
+                        set.add(element.getValue());
+                    }
                 }
             }
             if (!element.isWildcard()) {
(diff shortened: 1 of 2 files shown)


controller/src/main/java/org/jboss/as/controller/operations/global/GlobalOperationHandlers.java
```diff
@@ -687,7 +687,7 @@ public void execute(OperationContext context, ModelNode operation) throws Operat
                 PathAddress target = aliasEntry.convertToTargetAddress(addr.append(element));
                 PathAddress targetParent = target.subAddress(0, target.size() - 1);
                 Resource parentResource = context.readResourceFromRoot(targetParent, false);
-                if (parentResource != null && parentResource.hasChildren(target.getLastElement().getKey())) {
+                if (parentResource != null && parentResource.hasChild(target.getLastElement())) {
```
bstansberry: I believe this needs logic to deal with the possibility that target.getLastElement().isWildcard() is true. If so there won't be a child at that address and it will need to be handled more like the "if" block above this "else" block.

I assume it's because of that case that the code was written the way it is, but that left the WFCORE-692 problem.

pferraro: Good point - I'll adjust this.

controller/src/main/java/org/jboss/as/controller/operations/global/GlobalOperationHandlers.java
```diff
@@ -683,12 +683,14 @@ public void execute(OperationContext context, ModelNode operation) throws Operat
...
-                //PathAddress target = aliasEntry.getTargetAddress();
                 PathAddress target = aliasEntry.convertToTargetAddress(addr.append(element));
                 PathAddress targetParent = target.subAddress(0, target.size() - 1);
                 Resource parentResource = context.readResourceFromRoot(targetParent, false);
-                if (parentResource != null && parentResource.hasChildren(target.getLastElement().getKey())) {
-                    set.add(element.getValue());
+                if (parentResource != null) {
+                    PathElement targetElement = target.getLastElement();
+                    if (targetElement.isWildcard() ? parentResource.hasChildren(targetElement.getKey()) : parentResource.hasChild(targetElement)) {
+                        set.add(element.getValue());
```
bstansberry: If targetElement.isWildcard(), then the names of all children of that type need to be added. This will add "*'. (The old code did too.) The block starting here does this:

https://github.com/wildfly/wildfly-core/blob/master/controller/src/main/java/org/jboss/as/controller/operations/global/GlobalOperationHandlers.java#L678

This code could do the same thing, but using 'parentResource' instead of 'resource' and 'targetElement.getKey()' instead of 'childType'.

PR #729, opened 2015-05-14
[1.x] WFCORE-692 read-children-names(include-aliases=true) can return phantom results

pferraro:
if another resource exists with the same key as tje target address of a resource alias.

https://issues.jboss.org/browse/WFCORE-692

### DIFF
--- a/controller/src/main/java/org/jboss/as/controller/operations/global/GlobalOperationHandlers.java
+++ b/controller/src/main/java/org/jboss/as/controller/operations/global/GlobalOperationHandlers.java
@@ -683,12 +683,16 @@ public class GlobalOperationHandlers {
                     }
                 }
             } else {
-                //PathAddress target = aliasEntry.getTargetAddress();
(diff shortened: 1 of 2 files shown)


bstansberry:
Is this affecting you in 9.x? The method you are changing affects a lot of calls, so I'm reluctant to change it at the last minute in 9 to fix a problem that you don't see until 10.

That said, I think with the tweak I mention in my comment on the master PR, this looks pretty safe.

pferraro:
Technically, I don't need this until 10.  I opened a PR against 1.x since I figured that 1.0.0.Final gets integrated into WF10 before 2.0.0.Alpha1.

bstansberry:
2.0.0.Alpha1 will probably end up in full's master branch tomorrow. The 1.x branch is only being used for full's 9.x.

pferraro:
I've updated this PR with your comments from the corresponding master PR.  Technically, I don't need this fix until WF10, so I'll leave it to you to decide to merge or not.

PR #731, opened 2015-05-14
[1.x] - [WFCORE-694] Retrieve the model from the resource rather than attempt…

jamezp:
…ing to recursively read the resource. This also allows runtime-only deployments to be redeployed.


jamezp:
Master PR https://github.com/wildfly/wildfly-core/pull/730

PR #734, opened 2015-05-15
[WFCORE-696] Reduce PathAddress creation in access control checks

bstansberry:
Kabir counted 30K PathElement creations during a boot of full WildFly. I bet a lot of them come from here, as this creates 2 addresses per authorization, we authorize every attribute, and there are probably on average at least 2 PathElements per attribute.


bstansberry:
In unscientific testing on my laptop, this knocks about 61 ms off the boot time of the standalone-full.xml config. That was after #707 had knocked off about 18 ms.

PR #736, opened 2015-05-16
Drop the now-unused WildFly Security Manager artifact

dmlloyd:


dmlloyd:
It was only left in to allow the core upgrade to go smoothly until https://github.com/wildfly/wildfly/pull/7465 was merged.

PR #754, opened 2015-05-21
[WFCORE-700] jboss deployment structure error when two deployments share the same dependency

elguardian:
BZ: https://bugzilla.redhat.com/show_bug.cgi?id=1120628
Jira: https://issues.jboss.org/browse/WFCORE-700

This could cause race condition during deployment causing not being able
to deploy. During undeployment this problem could cause making apps
unusable (if the first app using the resource is undeployed, the other
sharing the resource cannot access those resourcees anymore)

The VFS mount point needs to be shared as several deployments
could share the same resource.

### DIFF
--- a/server/src/main/java/org/jboss/as/server/deployment/module/MountHandle.java
+++ b/server/src/main/java/org/jboss/as/server/deployment/module/MountHandle.java
@@ -36,7 +36,7 @@ import org.jboss.vfs.VFSUtils;
  * @author Stuart Douglas
  */
 public class MountHandle implements Closeable {
-    private final Closeable handle;
+    private Closeable handle;
 
     /**
      * Construct new instance with the mount handle to close.
@@ -53,6 +53,7 @@ public class MountHandle implements Closeable {
     public void close() {
         if (handle != null) {
             VFSUtils.safeClose(handle);
+            handle = null; // it is already closed. not need to close again during gc.
         }
     }
 
--- a/server/src/main/java/org/jboss/as/server/deployment/module/descriptor/SharedMountHandle.java
+++ b/server/src/main/java/org/jboss/as/server/deployment/module/descriptor/SharedMountHandle.java
@@ -1,0 +1,55 @@
+package org.jboss.as.server.deployment.module.descriptor;
+
+import java.io.Closeable;
+import java.io.IOException;
+import java.util.HashMap;
+import java.util.Map;
+
+import org.jboss.as.server.deployment.module.TempFileProviderService;
+import org.jboss.vfs.VFS;
+import org.jboss.vfs.VirtualFile;
+
+class SharedMountHandle implements Closeable {
+    private static Map<VirtualFile, SharedMountHandle> mountPoints = new HashMap<VirtualFile, SharedMountHandle>();
+    private VirtualFile file;
+    private Closeable closeable;
+    private volatile int numberOfReferences;
+
+    private SharedMountHandle(VirtualFile file, Closeable closeable) {
+        this.file = file;
+        this.closeable = closeable;
+    }
+
+    @Override
+    public void close() throws IOException {
+        synchronized (mountPoints) {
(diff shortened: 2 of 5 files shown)


elguardian:
retest this please.

server/src/main/java/org/jboss/as/server/deployment/module/descriptor/SharedMountHandle.java
```diff
@@ -0,0 +1,55 @@
...
+import java.util.HashMap;
+import java.util.Map;
+
+import org.jboss.as.server.deployment.module.TempFileProviderService;
+import org.jboss.vfs.VFS;
+import org.jboss.vfs.VirtualFile;
+
+class SharedMountHandle implements Closeable {
+    private static Map<VirtualFile, SharedMountHandle> mountPoints = new HashMap<VirtualFile, SharedMountHandle>();
+    private VirtualFile file;
```
bstansberry: This can be final.

server/src/main/java/org/jboss/as/server/deployment/module/descriptor/SharedMountHandle.java
```diff
@@ -0,0 +1,55 @@
...
+    private Closeable closeable;
```
bstansberry: This can be final/

server/src/main/java/org/jboss/as/server/deployment/module/descriptor/SharedMountHandle.java
```diff
@@ -0,0 +1,55 @@
...
+import java.io.IOException;
+import java.util.HashMap;
+import java.util.Map;
+
+import org.jboss.as.server.deployment.module.TempFileProviderService;
+import org.jboss.vfs.VFS;
+import org.jboss.vfs.VirtualFile;
+
+class SharedMountHandle implements Closeable {
+    private static Map<VirtualFile, SharedMountHandle> mountPoints = new HashMap<VirtualFile, SharedMountHandle>();
```
bstansberry: This static map makes me uncomfortable. In most cases such things are wrong, and something that has a more limited scope is better; i.e. something controlled by a service that can be cleaned up when that service stops, or simple gc'd when the service is no longer referenced.

I'm not familiar enough with all this to suggest a service that can do this though.

dmlloyd: I agree, we cannot have all this global state hanging around.

server/src/main/java/org/jboss/as/server/deployment/module/MountHandle.java
```diff
@@ -53,6 +53,7 @@ public MountHandle(final Closeable handle) {
     public void close() {
         if (handle != null) {
             VFSUtils.safeClose(handle);
+            handle = null; // it is already closed. not need to close again during gc.
```
dmlloyd: A better fix here is to keep it final and delete the finalize method.

elguardian: I was trying to keep the finalize method because I gueess it is there for proper clean up when some subsystems are not closing the mount handle (not calling the close method). I'm not sure if there is any sceneario where this could happen (that is why I kept it). If this cannot happen for sure I will remove it.

server/src/main/java/org/jboss/as/server/deployment/module/descriptor/SharedMountHandle.java
elguardian: The shared stated is based on the MountHandle finalize (gc)/close method.

ctomc:
As discussed in jira this is as wrong approach to fix this non-problem that is caused by miss using how deployment dependencies are done.
If anything we should add validation that resource-root defined in jboss-deployment-structure.xml doesn't allow paths that are outside of "current" deployment.

PR #755, opened 2015-05-21
adding a forcedName attribute to PersistentResourceXMLDescription

maeste:


bstansberry:
retest this please

bstansberry:
I think this approach is ok for this issue. If @ctomc comes up with something better before 10.0.0.Beta1 we can revisit it.

PR #756, opened 2015-05-21
WFCORE-287: Provide an operation to execute domain wide queries on the server side

heiko-braun:
Also includes:

- https://issues.jboss.org/browse/WFCORE-510 
- https://issues.jboss.org/browse/WFCORE-612
- https://issues.jboss.org/browse/WFCORE-287

### DIFF
--- a/controller/src/main/java/org/jboss/as/controller/descriptions/ModelDescriptionConstants.java
+++ b/controller/src/main/java/org/jboss/as/controller/descriptions/ModelDescriptionConstants.java
@@ -162,6 +162,7 @@ public class ModelDescriptionConstants {
     public static final String FILE = "file";
     public static final String FILE_HANDLER = "file-handler";
     public static final String FILTER = "filter";
+    //public final static String FILTER_CONJUNCT = "conjunct";
     public static final String FILTERED_ATTRIBUTES = "filtered-attributes";
     public static final String FILTERED_CHILDREN_TYPES = "filtered-children-types";
     public static final String FILTERED_OPERATIONS = "filtered-operations";
@@ -277,6 +278,7 @@ public class ModelDescriptionConstants {
     public static final String OPERATION_NAME = "operation-name";
     public static final String OPERATIONS = "operations";
     public static final String OPTIONS = "options";
+    public static final String OPERATOR = "operator";
     public static final String OUTBOUND_CONNECTION = "outbound-connection";
     /** Use this standard operation address field in the operation *description* ModelNode */
     public static final String OUTCOME = "outcome";
@@ -308,6 +310,7 @@ public class ModelDescriptionConstants {
     public static final String PROTOCOL = "protocol";
     public static final String PROVIDER = "provider";
     public static final String PROXIES = "proxies";
+    public static final String QUERY = "query";
     public static final String READ = "read";
     public static final String READ_ATTRIBUTE_OPERATION = "read-attribute";
     public static final String READ_ATTRIBUTE_GROUP_OPERATION = "read-attribute-group";
@@ -408,6 +411,7 @@ public class ModelDescriptionConstants {
     public static final String RESUME_SERVERS = "resume-servers";
     public static final String RESUME = "resume";
     public static final String SASL_PROTOCOL = "sasl-protocol";
+    public static final String SELECT = "select";
     public static final String SERVER_LOGGER = "server-logger";
     public static final String SERVER_NAME = "server-name";
     public static final String SHUTDOWN = "shutdown";
@@ -484,6 +488,7 @@ public class ModelDescriptionConstants {
     public static final String VAULT_EXPRESSION = "vault-expression";
     public static final String VAULT_OPTION = "vault-option";
     public static final String VAULT_OPTIONS = "vault-options";
+    public static final String WHERE = "where";
     public static final String WILDCARD = "wildcard";
     public static final String WRITE = "write";
     public static final String WRITE_ATTRIBUTE_OPERATION = "write-attribute";
--- a/controller/src/main/java/org/jboss/as/controller/operations/global/QueryOperationHandler.java
+++ b/controller/src/main/java/org/jboss/as/controller/operations/global/QueryOperationHandler.java
@@ -1,0 +1,261 @@
+/*
+ * JBoss, Home of Professional Open Source.
+ * Copyright 2011, Red Hat, Inc., and individual contributors
+ * as indicated by the @author tags. See the copyright.txt file in the
+ * distribution for a full listing of individual contributors.
+ *
+ * This is free software; you can redistribute it and/or modify it
+ * under the terms of the GNU Lesser General Public License as
+ * published by the Free Software Foundation; either version 2.1 of
+ * the License, or (at your option) any later version.
+ *
+ * This software is distributed in the hope that it will be useful,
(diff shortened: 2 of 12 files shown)


controller/src/main/java/org/jboss/as/controller/descriptions/ModelDescriptionConstants.java
```diff
@@ -162,6 +163,7 @@
     public static final String FILE = "file";
     public static final String FILE_HANDLER = "file-handler";
     public static final String FILTER = "filter";
+    //public final static String FILTER_CONJUNCT = "conjuinct";
```
bstansberry: Can you remove this? Or fix the spelling of "conjuinct" so if it gets uncommented later it doesn't slip in misspelled.

controller/src/main/java/org/jboss/as/controller/operations/global/QueryOperationHandler.java
```diff
@@ -0,0 +1,253 @@
...
+                                isEqual = filterValue.asInt() == resource.get(filterName).asInt();
+                                break;
+                            case DOUBLE:
+                                isEqual = filterValue.asDouble() == resource.get(filterName).asDouble();
+                                break;
+                            default:
+                                isEqual = filterValue.equals(resource.get(filterName));
+                        }
+                    } catch (IllegalArgumentException e) {
+                        throw new OperationFailedException("Illegal argument for attribute '"+filterName+"'. Expected type "+targetValueType);
```
bstansberry: This needs i18n handling, via a new method in ControllerLogger, a la ControllerLogger.invalidKeyForObjectType or various other ones that return OperationFailedException.

controller/src/main/java/org/jboss/as/controller/descriptions/ModelDescriptionConstants.java
heiko-braun: sure

controller/src/main/java/org/jboss/as/controller/operations/global/QueryOperationHandler.java
heiko-braun: ok, i'll update it accordingly

controller/src/main/java/org/jboss/as/controller/operations/global/QueryOperationHandler.java
```diff
@@ -0,0 +1,253 @@
...
+                for (Boolean match : matches) {
+                    if (match) {
+                        isMatching = true;
+                        break;
+                    }
+                }
+            }
+            else {
+                throw new OperationFailedException("Invalid operator"+operator);
+            }
```
bstansberry: This would need i18n if it remained, but it can go. L99's OPERATOR_ATT.validateOperation(operation); has already validated the input, and you've already done Operator.valueOf.

If you want to leave this block here as a guard against adding new Operator values but forgetting to handle them, please throw IllegalStateException or some other RuntimeException. OperationFailedException indicates a user mistake, not a server fault. Places I do that I don't include any text in the exception message, as that would require i18n and that's overkill. The stack trace points to where the code is broken.

host-controller/src/main/java/org/jboss/as/host/controller/HostModelUtil.java
```diff
@@ -97,6 +97,7 @@ public static void createRootRegistry(final ManagementResourceRegistration root,
 
         // Other root resource operations
         root.registerOperationHandler(CompositeOperationHandler.DEFINITION, CompositeOperationHandler.INSTANCE);
+        //root.registerOperationHandler(MapReduceOperationHandler.DEFINITION, MapReduceOperationHandler.INSTANCE);
```
bstansberry: Please remove. The original ordering of the imports was the correct one as well.

server/src/main/java/org/jboss/as/server/controller/resources/ServerRootResourceDefinition.java
```diff
@@ -259,6 +259,8 @@ public void registerOperations(ManagementResourceRegistration resourceRegistrati
 
         // Other root resource operations
         resourceRegistration.registerOperationHandler(CompositeOperationHandler.DEFINITION, CompositeOperationHandler.INSTANCE, false);
+        //resourceRegistration.registerOperationHandler(MapReduceOperationHandler.DEFINITION, MapReduceOperationHandler.INSTANCE, false);
```
bstansberry: Please remove. The original ordering of the imports was the correct one as well.

testsuite/domain/src/test/java/org/jboss/as/test/integration/domain/suites/CoreResourceManagementTestCase.java
```diff
@@ -910,6 +915,22 @@ public void testAddRemoveServerConfig() throws Exception {
...
+
+        final ModelNode operation = new ModelNode();
+        operation.get(OP).set(QUERY);
+        operation.get(ADDRESS).setEmptyList();
+        operation.get(ADDRESS_TEMPLATE).set("/host=master/server=*");
+        operation.get(FILTER).add("foo", "bar");
+        operation.get(SELECT).add("name");
+
+        ModelNode response = masterClient.execute(operation);
+        System.out.println(response);
```
bstansberry: No assertions on the response?

heiko-braun: whoops! Unfinished business :)

heiko-braun:
Increased test coverage.

heiko-braun:
the latest commit squashes the improvements based on @bstansberry comments on removes the reference to WFCORE-442 (which isn't fixed)

heiko-braun:
Are we good now?

controller/src/main/java/org/jboss/as/controller/operations/global/QueryOperationHandler.java
```diff
@@ -0,0 +1,261 @@
...
+                                isEqual = filterValue.asInt() == resource.get(filterName).asInt();
+                                break;
+                            case DOUBLE:
+                                isEqual = filterValue.asDouble() == resource.get(filterName).asDouble();
+                                break;
+                            default:
+                                isEqual = filterValue.equals(resource.get(filterName));
+                        }
+                    } catch (IllegalArgumentException e) {
+                        throw ControllerLogger.MGMT_OP_LOGGER.validationFailedCouldNotConvertParamToType(filterName, targetValueType, ModelDescriptionConstants.QUERY);
```
bstansberry: This method throws IllegalArgumentException, not OperationFailedException. But your catch block at L144 is counting on catching OperationFailedException.

I was testing the WFCORE-442 fix and suddenly /path=*:query(where={read-only=-1}) wouldn't fail properly any more. This is why.

heiko-braun: I don't understand your comment. It fails the same way as before, just with proper message

heiko-braun: ah, now i see it. not all ControllerLogger messages create OperationFailedExceptions

heiko-braun:
ok, local test pass and the response to the tets query for WFCORE-442 fails with a proper message.

bstansberry:
:-)

PR #757, opened 2015-05-21
[WFCORE-704] Adding initial Elytron capability definitions.

darranl:
For the Elytron integration it should be possible to completely remove the subsystem and provide an alternative that satisfies the same set of capabilities.  This PR is to get the initial class in place to hold the capability definitions (more to be defined later).

### DIFF
--- a/controller/src/main/java/org/jboss/as/controller/security/Capabilities.java
+++ b/controller/src/main/java/org/jboss/as/controller/security/Capabilities.java
@@ -1,0 +1,73 @@
+/*
+ * JBoss, Home of Professional Open Source.
+ * Copyright 2015, Red Hat, Inc., and individual contributors
+ * as indicated by the @author tags. See the copyright.txt file in the
+ * distribution for a full listing of individual contributors.
+ *
+ * This is free software; you can redistribute it and/or modify it
+ * under the terms of the GNU Lesser General Public License as
+ * published by the Free Software Foundation; either version 2.1 of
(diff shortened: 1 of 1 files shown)


bstansberry:
I don't like this at all, particularly in controller.

darranl:
Yes that was the reason for getting this PR in - I assumed it would be one to drive some debate ;-)

So here is the problem - we have multiple subsystems providing the capabilities and they are used in multiple places - both within our core codebase and within other subsystems, some of those subsystems are in the WildFly / Core codebase and others are not.

Essentially there are three pieces of information that everything needs to be in agreement with when it comes to the capability definitions: -
 - The base name.
 - Whether the capability is dynamic.
 - The Service type

Regarding location of the definitions the reason I selected controller was really driven by maven dependencies and controller did already contain a 'security' package.  To define a RuntimeCapability I need a dependency on controller as that is where that class lives, I did consider the existing security and security-api modules but controller already had a dependency on them so would have created a cycle - this is also more than domain management so did not really fit in that module either.  

So we need two discussions: -
 1 - How we coordinate the three requirements described above between different components.
 2 - Where that coordination lives.

Number 2 probably only makes sense after 1.

darranl:
It has been decided that the capabilities will be co-ordinated using documentation: -
  https://developer.jboss.org/wiki/WildFlySecurityElytron-CapabilitiesAndRequirements-Reference

Static definitions can still be used within code but these would be for the local module they are defined within.

PR #758, opened 2015-05-21
Fix CLI Vault testsuite failures on windows (or any other machine wit…

ctomc:
…h empty maven repo)


bstansberry:
I think this is a problem on 1.x as well.

ctomc:
ah yes, let me send PR against 1.x as well

PR #762, opened 2015-05-27
[WFCORE-561] Remove the JDK version of javax.sql from the visible class path

dmlloyd:


jamezp:
Failure seems related.

dmlloyd:
Hmph.

dmlloyd:
That one is definitely not related. :-)

PR #764, opened 2015-05-28
[WFCORE-718] Add CLI Command Builder

jamezp:
Added a `CliCommandBuilder` and converted the CLI tests to use it.

The main idea behind the launcher is to allow consumers like the `wildfly-maven-plugin` to be execute CLI commands in a modular environment. Some CLI commands seem to prefer, maybe require, a modular environment. This just eases the pain of manually launching the process with some defaults.

### DIFF
--- a/launcher/src/main/java/org/wildfly/core/launcher/CliCommandBuilder.java
+++ b/launcher/src/main/java/org/wildfly/core/launcher/CliCommandBuilder.java
@@ -1,0 +1,688 @@
+/*
+ * JBoss, Home of Professional Open Source.
+ * Copyright 2015, Red Hat, Inc., and individual contributors
+ * as indicated by the @author tags. See the copyright.txt file in the
+ * distribution for a full listing of individual contributors.
+ *
+ * This is free software; you can redistribute it and/or modify it
+ * under the terms of the GNU Lesser General Public License as
+ * published by the Free Software Foundation; either version 2.1 of
+ * the License, or (at your option) any later version.
+ *
+ * This software is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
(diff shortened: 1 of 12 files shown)


jamezp:
Introduces a binary incompatibility. It will be better to wait until a core release until the tests are converted. I'll fix this in the morning after I've thought about it.

jamezp:
These aren't failing locally. retest this please

bstansberry:
This seems like a pretty big scope change for the launcher. I'm not really opposed to this in principle, as the launcher lib looks to be more of a general lib for creating modular processes  but perhaps a dev list post would be good?

launcher/src/main/java/org/wildfly/core/launcher/CliCommandBuilder.java
```diff
@@ -0,0 +1,654 @@
...
+     * Sets the protocol, hostname and port to connect to.
+     *
+     * @param protocol the protocol to use
+     * @param hostname the host name
+     * @param port     the port
+     *
+     * @return the builder
+     */
+    public CliCommandBuilder setController(final String protocol, final String hostname, final int port) {
+        setController(String.format("%s://%s:%d", protocol, hostname, port));
```
bstansberry: This is a bit weak when it comes to IPv6 addresses, as it assumes "hostname" is properly enclosed in [ ].

jamezp:
I'll mail the dev mailing list. If I'm the only one that will use it maybe it won't make sense to add it here.

launcher/src/main/java/org/wildfly/core/launcher/CliCommandBuilder.java
jamezp: Do we want to fully support parsing it like the `org.jboss.as.network.NetworkUtils` does?

jamezp:
For completeness here's the link to the start of the thread http://lists.jboss.org/pipermail/wildfly-dev/2015-June/004049.html

PR #767, opened 2015-05-29
WFLY-4670 Remove jacorb logging configuration

tadamski:


jamezp:
We'll need to remove the categories from https://github.com/wildfly/wildfly-core/blob/master/core-feature-pack/src/main/resources/content/standalone/configuration/logging.properties#L44-L45 and https://github.com/wildfly/wildfly-core/blob/master/core-feature-pack/src/main/resources/content/domain/configuration/default-server-logging.properties#L44-L45 too.

jamezp:
Sorry I should have been more specific. They need to be removed from https://github.com/tadamski/wildfly-core/blob/WFLY-4670/core-feature-pack/src/main/resources/content/domain/configuration/default-server-logging.properties#L24 and https://github.com/tadamski/wildfly-core/blob/WFLY-4670/core-feature-pack/src/main/resources/content/domain/configuration/default-server-logging.properties#L32-L33 on both files as well.

tadamski:
Removed the line + jacorb logger configuration. No more jacorbs left in both files :)

PR #768, opened 2015-05-30
[WFCORE-717] Remove duplicate code in standalone.sh and disable setti…

langfr:
…ng of option -XX:+UseCompressedOops as it's enabled by default in most cases.

### DIFF
--- a/core-feature-pack/src/main/resources/content/bin/standalone.sh
+++ b/core-feature-pack/src/main/resources/content/bin/standalone.sh
@@ -157,15 +157,6 @@ if [ "$PRESERVE_JAVA_OPTS" != "true" ]; then
         fi
     fi
 
-    # Enabled by default in Java SE 6u23 and later.
-    # In Java SE 7, use of compressed oops is the default for 64-bit JVM processes when -Xmx isn't specified and for values of -Xmx less than 32 gigabytes.
-    #if [ $CLIENT_VM = false ]; then
-    #    NO_COMPRESSED_OOPS=`echo $JAVA_OPTS | $GREP "\-XX:\-UseCompressedOops"`
-    #    if [ "x$NO_COMPRESSED_OOPS" = "x" ]; then
-    #        "$JAVA" $JVM_OPTVERSION -server -XX:+UseCompressedOops -version >/dev/null 2>&1 && PREPEND_JAVA_OPTS="$PREPEND_JAVA_OPTS -XX:+UseCompressedOops"
-    #    fi
-    #fi
-
     # EAP6-121 feature disabled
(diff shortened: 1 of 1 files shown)


core-feature-pack/src/main/resources/content/bin/standalone.sh
```diff
@@ -157,12 +157,28 @@ if [ "$PRESERVE_JAVA_OPTS" != "true" ]; then
...
 
-    if [ $CLIENT_VM = false ]; then
-        NO_COMPRESSED_OOPS=`echo $JAVA_OPTS | $GREP "\-XX:\-UseCompressedOops"`
-        if [ "x$NO_COMPRESSED_OOPS" = "x" ]; then
-            "$JAVA" $JVM_OPTVERSION -server -XX:+UseCompressedOops -version >/dev/null 2>&1 && PREPEND_JAVA_OPTS="$PREPEND_JAVA_OPTS -XX:+UseCompressedOops"
-        fi
-    fi
+    # Enabled by default in Java SE 6u23 and later.
+    # In Java SE 7, use of compressed oops is the default for 64-bit JVM processes when -Xmx isn't specified and for values of -Xmx less than 32 gigabytes.
+    #if [ $CLIENT_VM = false ]; then
```
ctomc: you can remove this whole block of script as I doubt it will ever be used again.

core-feature-pack/src/main/resources/content/bin/standalone.sh
```diff
@@ -157,12 +157,19 @@ if [ "$PRESERVE_JAVA_OPTS" != "true" ]; then
         fi
     fi
 
-    if [ $CLIENT_VM = false ]; then
-        NO_COMPRESSED_OOPS=`echo $JAVA_OPTS | $GREP "\-XX:\-UseCompressedOops"`
-        if [ "x$NO_COMPRESSED_OOPS" = "x" ]; then
-            "$JAVA" $JVM_OPTVERSION -server -XX:+UseCompressedOops -version >/dev/null 2>&1 && PREPEND_JAVA_OPTS="$PREPEND_JAVA_OPTS -XX:+UseCompressedOops"
-        fi
-    fi
```
jamezp: This seems okay to remove, but we should also remove the `CLIENT_VM` variable and the parts where it's set.

langfr: Done.

jamezp:
Thanks for the fix here! I know you've submitted several other PR's and it looks like we just plain forgot to ask you the following :).

langfr:
Yes, I agree.

PR #769, opened 2015-06-01
WFCORE-725 Initial work on removing ClassIndex

stuartwdouglas:
This moves similar functionality found in ClassIndex into DeploymentClassIndex. Once
a core release with these changes has been released then full can remove all references
to ClassIndex, and it can be removed from the code base.


jamezp:
retest this please

PR #771, opened 2015-06-02
[WFCORE-729] Add smoke test for showing the real value of expression in CLI

marekkopecky:
CLI should show the real value of expression properties in addition to the expression
https://issues.jboss.org/browse/WFCORE-729


kabir:
This is ok to test

Gonthim:
Looks good to me. Thanks!

PR #772, opened 2015-06-02
Domain Ops and ordered child resources

kabir:
This is to see if the tests pass or not. The full integration one will definitely fail. For that to pass https://github.com/wildfly/wildfly/pull/7555 will need to be merged first, as it contains some temporary fixes for compatibility which will later need to be removed as part of https://issues.jboss.org/browse/WFLY-4716 and https://issues.jboss.org/browse/WFCORE-726.

https://github.com/wildfly/wildfly-core/pull/772 contains removal of the workarounds for backwards compatibility following a core release.

http://brontes.lab.eng.brq.redhat.com/viewLog.html?buildId=55530&tab=buildResultsDiv&buildTypeId=WF_WildFlyCoreIntegrationExperiments and http://brontes.lab.eng.brq.redhat.com/viewLog.html?buildId=55532&tab=buildResultsDiv&buildTypeId=WF_WildFlyCoreIntegrationExperiments test core against a snapshot containing this patch.


kabir:
http://brontes.lab.eng.brq.redhat.com/viewLog.html?buildId=55249&tab=buildResultsDiv&buildTypeId=WF_WildFlyCoreIntegrationExperiments is the job with this snapshot against #7555

kabir:
Retest this please

kabir:
Retest this please

kabir:
The ParseAndMarshalModelsTestCase failures are due to the experiments job somehow picking up the wrong version of wildfly-core-testsuite-shared. Although 2.0.0.Alpha4-SNAPSHOT is specified, it is still using 2.0.0.Alpha3. I've reproduced this locally, but don't see the root of the problem, as it looks like it should use the specified version https://github.com/wildfly/wildfly/blob/master/testsuite/shared/pom.xml#L32

PR #773, opened 2015-06-03
DISCARD_AND_WARN claims to reject instead of discard

rhusar:
Kabir, maybe something like this then?

### DIFF
--- a/controller/src/main/java/org/jboss/as/controller/transform/description/TransformingDescription.java
+++ b/controller/src/main/java/org/jboss/as/controller/transform/description/TransformingDescription.java
@@ -137,14 +137,15 @@ class TransformingDescription extends AbstractDescription implements Transformat
 
     @Override
     public void transformResource(final ResourceTransformationContext ctx, final PathAddress address, final Resource original) throws OperationFailedException {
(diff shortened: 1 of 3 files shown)


rhusar:
Retest this please.

controller/src/main/java/org/jboss/as/controller/transform/description/TransformingDescription.java
```diff
@@ -137,14 +137,15 @@ void invokeNext(OperationTransformer.TransformedOperation transformedOperation)
...
         final DiscardPolicy discardPolicy = determineDiscardPolicy(ctx, address);
 
         switch (discardPolicy) {
-            case DISCARD_AND_WARN:
             case REJECT_AND_WARN:
                 ctx.getLogger().logRejectedResourceWarning(address, null);
                 return;
+            case DISCARD_AND_WARN:
+                // don't return yet, just log a warning first and then discard
+                ctx.getLogger().logDiscardedResourceWarning(address);
```
kabir: I would also add a parameter for the host name:
 ctx.getTarget().getHostName()
Apart from that it looks good

rhusar: OK; added host name to logging.

PR #777, opened 2015-06-04
[WFCORE-732] Extend tests for CLI commands extensibility

pkremens:
https://issues.jboss.org/browse/WFCORE-732 - Extend tests for CLI commands extensibility

Alexey added test as part of https://issues.jboss.org/browse/WFCORE-582. Adding two more tests.

**command name conflict**
* try to register command with name of some standard CLI command
* command is not registered

**help message for custom command**
* add help message to the command from extension
* verify, that help message is available

Adding tests as part of https://issues.jboss.org/browse/EAP7-91 verification procedure.


pkremens:
I'll try to hunt down the Windows failure...

bstansberry:
I don't see why these test classes (including the existing stuff in CliExtCommandsTestCase) needs to be in manual mode (which is slower.) Unless you're aware of a good reason for this, can you file a JIRA to move them please? Thanks.

Otherwise, looks good.

pkremens:
You are right, these can be moved. I'll move them once this is merged.

https://issues.jboss.org/browse/WFCORE-766

PR #780, opened 2015-06-04
[WFCORE-718] Add CLI command builder.

jamezp:
The 2.0 PR has some details about why this is desired https://github.com/wildfly/wildfly-core/pull/764.

This is not required for 1.0, but it would be nice to be able to use it in the `wildfly-maven-plugin` which will target 1.0 to stick with a Java 1.7 requirement.

I did not change bring the change for the tests over to the 1.x PR as I didn't want make potential test breaking changes. For 1.x this API will not be used anywhere only to be consumed from other projects.


jamezp:
retest this please

jamezp:
I tested this locally with `-Dipv6` and all the tests passed.

PR #781, opened 2015-06-04
[WFCORE-723] Allow multiple server arguments with the same key.

jamezp:
In some cases multiple arguments with the same key should be allowed to be added. 

In the system property `-Dkey=value` the key is defined as `key`. With all other argument types the key is defined as any string before an `=`. If the argument does not contain an `=` then they key is the argument and the value is `null` for the argument.


bstansberry:
I'm not merging, but +1.

PR #784, opened 2015-06-05
[WFCORE-579]:ProxyMetadataSource does not return JDK8 'default' metho…

jimma:
…ds fro minterfaces


jamezp:
This will be fixed as part of https://github.com/wildfly/wildfly-core/pull/769

PR #786, opened 2015-06-08
WFLY-4570 Add a systemd service

rarguello:
An example systemd service.

### DIFF
--- a/core-feature-pack/src/main/resources/content/bin/systemd/README
+++ b/core-feature-pack/src/main/resources/content/bin/systemd/README
@@ -7,17 +7,16 @@
 
 == Install WildFly
 
-    # mkdir /opt/wildfly
-    # tar xvzf wildfly-9.0.0.Final.tar.gz -C /opt
-    # ln -s /opt/wildfly-9.0.0.Final /opt/wildfly
+    # tar xvzf wildfly-9.0.1.Final.tar.gz -C /opt
+    # ln -s /opt/wildfly-9.0.1.Final /opt/wildfly
     # chown -R wildfly:wildfly /opt/wildfly
 
 == Configure systemd
 
     # mkdir /etc/wildfly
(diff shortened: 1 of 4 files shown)


core-feature-pack/src/main/resources/content/bin/systemd/README
```diff
@@ -0,0 +1,26 @@
+= How to configure WildFly as a systemd service
+
+== Create a wildfly user
+
+    # groupadd -r wildfly
+    # useradd -r -g wildfly -d /usr/share/wildfly -s /sbin/nologin wildfly
+
+== Install WildFly
+
+    # mkdir /usr/share/wildfly/
```
dmlloyd: Installation should probably be in /opt.  Data files should be (I guess) in /var/opt.  I would at least use http://www.linuxbase.org/betaspecs/fhs/fhs.html as a reference.

dmlloyd:
OK but I think we need different default configurations for this then, because we don't want to put data files under /opt, they should be under /var, in particular the logging and data directories need to be separated.  So maybe this should continue to just be an example rather than something we expect people to typically use.

rarguello:
This should be considered an example only, like the init.d scripts:
https://github.com/wildfly/wildfly-core/tree/master/core-feature-pack/src/main/resources/content/bin/init.d

jamezp:
I'm placing a hold on this for now until we can determine how we want to do this or if we want to do this. I personally don't like any of the extra scripts in the code base.

ctomc:
@dmlloyd @jamezp i think we should merge this, as everything is moving to systemd and even having a starting point is a good thing.

ctomc:
this is ok to test

ctomc:
I squashed your commits in this PR and included them in #1132

rarguello:
Thanks

PR #787, opened 2015-06-08
[WFCORE-747] Use read-attribute server-state to exam EAP service startup result

soul2zimate:
see description and comment in https://issues.jboss.org/browse/WFCORE-747


bstansberry:
This isn't reliable. There's no guarantee the config supports remote management. It probably does, but then there's no guarantee the management interface allows the local authentication protocol, without which your CLI command will not work. Remove the following from the config file and it won't work any more:

```
<local default-user="$local"/>
```

Also, 9999 is not the default management port any more (9990 is), and the protocol to use would need to be configurable.

soul2zimate:
OK, thanks for review. I was not aware of the authentication of CLI connection since I usually connect it without any password required in dev. In such case, I think checking from CLI command is not right approach as non encrypted password in any configuration files is not advised. I will go back into logs and try alternative way.

PR #791, opened 2015-06-08
[WFCORE-746] :  [WFCORE-746] : Provides a UUID per server instance

ehsavoie:
Provides a type 4 UUID for a server instance.
The UUID will be saved in $JBOSS_DATA_DIR/kernel/process-uuid and will be (re)created on server instance startup if it doesn't exist.
This UUID should be accessible through the Management API.

Jira: https://issues.jboss.org/browse/WFCORE-746

### DIFF
--- a/controller/src/main/java/org/jboss/as/controller/operations/common/ProcessEnvironment.java
+++ b/controller/src/main/java/org/jboss/as/controller/operations/common/ProcessEnvironment.java
@@ -24,9 +24,15 @@ package org.jboss.as.controller.operations.common;
 
 import static org.jboss.as.controller.descriptions.ModelDescriptionConstants.VALUE;
 
+import java.io.IOException;
 import java.net.InetAddress;
 import java.net.UnknownHostException;
+import java.nio.file.Files;
+import java.nio.file.Path;
+import java.nio.file.StandardOpenOption;
+import java.util.Collections;
 import java.util.UUID;
+import java.util.stream.Stream;
 
 import org.jboss.as.controller.AttributeDefinition;
 import org.jboss.as.controller.OperationContext;
@@ -47,6 +53,11 @@ import org.jboss.dmr.ModelType;
  * @author Brian Stansberry (c) 2011 Red Hat Inc.
  */
 public abstract class ProcessEnvironment {
+    /** The name of the file used to store the process UUID */
+    protected static final String UUID_FILE = "process-uuid";
+
+    /** The name of the directory used to store WildFly kernel specific files */
+    protected static final String KERNEL_DIR = "kernel";
 
     /** The special process name value that triggers calculation of a UUID */
     public static final String JBOSS_DOMAIN_UUID = "jboss.domain.uuid";
@@ -116,6 +127,27 @@ public abstract class ProcessEnvironment {
      * @param propertyValue the value of the property. May be {@code null}
      */
     protected abstract void systemPropertyUpdated(String propertyName, String propertyValue);
+
+    /**
+     * Get the UUID of this process.
+     * @return the UUID of this process.
+     */
+    public abstract UUID getInstanceUuid();
+
+    protected UUID obtainProcessUUID(final Path filePath) throws IOException {
+        UUID uuid = null;
+        if (Files.exists(filePath)) {
+            try (Stream<String> lines = Files.lines(filePath)) {
+                uuid = UUID.fromString(lines.findFirst().get());
+            }
+        }
+        if (uuid == null) {
+            uuid = UUID.randomUUID();
+            Files.createDirectories(filePath.getParent());
+            Files.write(filePath, Collections.singletonList(uuid.toString()), StandardOpenOption.CREATE);
+        }
+        return uuid;
+    }
 
     protected static String resolveGUID(final String unresolvedName) {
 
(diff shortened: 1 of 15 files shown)


controller/src/main/java/org/jboss/as/controller/operations/common/ProcessEnvironment.java
```diff
@@ -117,6 +127,26 @@ protected abstract boolean isRuntimeSystemPropertyUpdateAllowed(String propertyN
...
+    protected UUID createProcessUUID(final Path filePath) throws IOException {
+        UUID uuid = null;
+        if (Files.exists(filePath)) {
+            String line = Files.lines(filePath).findFirst().get();
+            uuid = UUID.fromString(line);
+        }
+        if (uuid == null) {
+            uuid = UUID.randomUUID();
+            Files.createDirectories(filePath.getParent());
+            Files.write(filePath, Collections.singletonList(uuid.toString()), StandardOpenOption.CREATE);
```
jamezp: These two lines, 144 and 145, should likely be done in a privileged block.

jamezp: Actually I should get @dmlloyd opinion on that. It is probably okay not to use a privileged block.

controller/src/main/java/org/jboss/as/controller/operations/common/ProcessEnvironment.java
```diff
@@ -117,6 +127,26 @@ protected abstract boolean isRuntimeSystemPropertyUpdateAllowed(String propertyN
...
+    /**
+     * Get the UUID of this process.
+     * @return the UUID of this process.
+     */
+    public abstract UUID getInstanceUuid();
+
+    protected UUID createProcessUUID(final Path filePath) throws IOException {
+        UUID uuid = null;
+        if (Files.exists(filePath)) {
+            String line = Files.lines(filePath).findFirst().get();
```
jamezp: This should be done in a try-with-resources block to ensure the stream is closed.

host-controller/src/test/java/org/jboss/as/host/controller/HostControllerEnvironmentTestCase.java
```diff
@@ -0,0 +1,112 @@
...
+        }
+    }
+
+    @Test
+    public void testUUIDLifeCycle() throws IOException {
+        Map<String, String> hostProperties = new HashMap();
+        Path domainDir = homeDir.resolve("domain");
+        Files.createDirectories(domainDir.resolve("configuration"));
+        Path uuidPath = domainDir.resolve("data").resolve("kernel").resolve("process-uuid");
+        assertThat(Files.notExists(uuidPath), is(true));
```
jamezp: More curiosity  than anything, is there a reason not to use `assertTrue(Files.notExists(uuidPath))`?

ehsavoie: I prefer to use the same assertion everywhere

controller/src/main/java/org/jboss/as/controller/operations/common/ProcessEnvironment.java
ehsavoie: Yes this was lost in refactoring

host-controller/src/test/java/org/jboss/as/host/controller/HostControllerEnvironmentTestCase.java
stuartwdouglas: I am not sure what you mean? The majority of the test suite uses Assert.assert(True|False|Equals), and this is basically just a more verbose way of saying Assert.assertTrue()

controller/src/main/java/org/jboss/as/controller/operations/common/ProcessEnvironment.java
```diff
@@ -117,6 +130,34 @@ protected abstract boolean isRuntimeSystemPropertyUpdateAllowed(String propertyN
...
+
+    protected UUID obtainProcessUUID(final Path filePath) throws Exception {
+        UUID uuid = null;
+        if (Files.exists(filePath)) {
+            try (Stream<String> lines = Files.lines(filePath)) {
+                uuid = UUID.fromString(lines.findFirst().get());
+            }
+        }
+        if (uuid == null) {
+            PrivilegedExceptionAction<UUID> action = new PrivilegedExceptionAction<UUID>() {
```
jamezp: I was wrong about the needing of a privileged block. I should have looked more closely at where this was used. Sorry about that.

controller/src/main/java/org/jboss/as/controller/operations/common/ProcessEnvironment.java
```diff
@@ -117,6 +130,34 @@ protected abstract boolean isRuntimeSystemPropertyUpdateAllowed(String propertyN
      */
     protected abstract void systemPropertyUpdated(String propertyName, String propertyValue);
 
+    /**
+     * Get the UUID of this process.
+     * @return the UUID of this process.
+     */
+    public abstract UUID getInstanceUuid();
+
+    protected UUID obtainProcessUUID(final Path filePath) throws Exception {
```
jamezp: It probably makes more sense to throw an `IllegalStateException` here as each usage of this method just catches an exception and throws an `IllegalStateException`.

host-controller/src/test/java/org/jboss/as/host/controller/HostControllerEnvironmentTestCase.java
bstansberry: I don't personally like this style, but as long as it's not getting mixed in with the usual assertTrue style in existing classes, it's fine.

ehsavoie: with assertTrue you have : an exception AssertionError with a stack trace
with assertThat you have : Expected is <true> but was <false>
I find this to be more coherent with the other assertion messages and behaviour

PR #792, opened 2015-06-08
WFCORE-749 Add SELF_CONTAINED to ServerService.getProcessType()

luck3y:
WFCORE-749 1.x version


jamezp:
I think this should be in 1.x as WildFly Swarm uses the `SELF_CONTAINED` server type and depends on WildFly Core 1.x.

luck3y:
Yep, agree. I sent another PR for 1.x.

PR #804, opened 2015-06-11
Get framework set up for testing the current client against old versions

kabir:
Refactor the Server in test-runner to not use static fields, so that we
can change the values.

As mentioned in my comment https://issues.jboss.org/browse/JBIDE-19641?focusedCommentId=13077135&page=com.atlassian.jira.plugin.system.issuetabpanels:comment-tabpanel#comment-13077135 this needs some setup:
-Djboss.test.client.old.server.dir=<path-to-dir-containing zips> 
-Djboss.test.client.old.server.jdk7=%env.JDK_17%

TeamCity currently only has the EAP zips, so I want to merge this first and to explore what JVM options are needed before uploading all the other versions. We probably don't want this as part of normal PR/master-ignore etc. runs, so I will create a custom job (similar to the mixed domain job) and see where that gets me.

### DIFF
--- a/testsuite/client-old-server/server-provisioning.xml
+++ b/testsuite/client-old-server/server-provisioning.xml
@@ -1,0 +1,26 @@
+<!--
+  ~ JBoss, Home of Professional Open Source.
+  ~ Copyright 2014, Red Hat, Inc., and individual contributors
+  ~ as indicated by the @author tags. See the copyright.txt file in the
+  ~ distribution for a full listing of individual contributors.
+  ~
+  ~ This is free software; you can redistribute it and/or modify it
+  ~ under the terms of the GNU Lesser General Public License as
+  ~ published by the Free Software Foundation; either version 2.1 of
+  ~ the License, or (at your option) any later version.
+  ~
+  ~ This software is distributed in the hope that it will be useful,
+  ~ but WITHOUT ANY WARRANTY; without even the implied warranty of
+  ~ MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the GNU
+  ~ Lesser General Public License for more details.
+  ~
+  ~ You should have received a copy of the GNU Lesser General Public
+  ~ License along with this software; if not, write to the Free
+  ~ Software Foundation, Inc., 51 Franklin St, Fifth Floor, Boston, MA
+  ~ 02110-1301 USA, or see the FSF site: http://www.fsf.org.
+  -->
+<server-provisioning xmlns="urn:wildfly:server-provisioning:1.0" copy-module-artifacts="true">
+    <feature-packs>
+        <feature-pack groupId="org.wildfly.core" artifactId="wildfly-core-feature-pack" version="${project.version}" />
+    </feature-packs>
+</server-provisioning>
--- a/testsuite/client-old-server/src/test/java/org/wildfly/client/old/server/ClientAgainstOldServerTestCase.java
+++ b/testsuite/client-old-server/src/test/java/org/wildfly/client/old/server/ClientAgainstOldServerTestCase.java
@@ -1,0 +1,137 @@
+/*
+ * JBoss, Home of Professional Open Source.
+ * Copyright ${year}, Red Hat, Inc., and individual contributors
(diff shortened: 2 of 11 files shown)


testsuite/client-old-server/src/test/java/org/wildfly/client/old/server/ClientAgainstOldServerTestCase.java
```diff
@@ -0,0 +1,137 @@
...
+
+    @Test
+    public void testDeployment() throws Exception {
+        final ModelControllerClient client = managementClient.getControllerClient();
+        final ServerDeploymentManager manager = ServerDeploymentManager.Factory.create(client);
+
+        //Deploy
+        try (final InputStream is = testDeploymentFile.getInputStream()){
+            Future<?> future = manager.execute(
+                    manager.newDeploymentPlan().add(testDeploymentFile.getName(), is).deploy(testDeploymentFile.getName()).build());
```
bstansberry: Just removing the ".deploy(testDeploymentFile.getName())" should eliminate the DUPs from the test.

testsuite/client-old-server/src/test/java/org/wildfly/client/old/server/ClientAgainstOldServerTestCase.java
```diff
@@ -0,0 +1,137 @@
...
+        }
+
+        //Check there
+        ModelNode findDeployment = Util.createEmptyOperation(READ_RESOURCE_OPERATION,
+                PathAddress.pathAddress(DEPLOYMENT, testDeploymentFile.getName()));
+        Assert.assertTrue(
+                managementClient.executeForResult(findDeployment).isDefined());
+
+        //Undeploy
+        Future<?> future = manager.execute(manager.newDeploymentPlan().undeploy(testDeploymentFile.getName()).remove(testDeploymentFile.getName()).build());
```
bstansberry: If you remove the ".deploy(testDeploymentFile.getName())" then the ".undeploy(testDeploymentFile.getName())" goes too.

testsuite/client-old-server/src/test/java/org/wildfly/client/old/server/util/Version.java
```diff
@@ -0,0 +1,95 @@
...
+/**
+ * @author Kabir Khan
+ */
+public class Version {
+
+    static final String AS = "jboss-as-";
+    static final String WILDFLY = "wildfly-";
+    static final String EAP = "jboss-eap-";
+
+    public enum AsVersion {
```
bstansberry: If JBDS supports them, we should cover 7.0.0, 7.0.1, 7.0.2, 7.1.0, 7.1.1. That's not needed for merging this though.

ctomc: well 7.0.x stuff we should stop testing.
On related problem, 7.x stuff should be tested with JDK6 not 7....

bstansberry: What we test depends on what JBDS supports. The goal of this is to identify a set of client versions they can use against all releases they support.

If our client won't work against some release we can stop testing it once we identify the branch that has a client that will work, and start testing that client.

testsuite/client-old-server/server-provisioning.xml
```diff
@@ -0,0 +1,26 @@
...
+  ~ but WITHOUT ANY WARRANTY; without even the implied warranty of
+  ~ MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the GNU
+  ~ Lesser General Public License for more details.
+  ~
+  ~ You should have received a copy of the GNU Lesser General Public
+  ~ License along with this software; if not, write to the Free
+  ~ Software Foundation, Inc., 51 Franklin St, Fifth Floor, Boston, MA
+  ~ 02110-1301 USA, or see the FSF site: http://www.fsf.org.
+  -->
+<server-provisioning xmlns="urn:wildfly:server-provisioning:1.0" copy-module-artifacts="true">
```
ctomc: we could probably go with thin provisioning here aka copy-module-artifacts=false

PR #813, opened 2015-06-16
[WFCORE-759] Creating ordered child resources driven by resource

kabir:
registration. All that is needed now is to call the SimpleResourceDefinition constructor with ordered=true. This ends up as ResourceDefinition.isOrderedChild() = true, and then the framework can figure out the rest.

~~https://github.com/wildfly/wildfly/pull/7611 needs merging before this can pass the full integration tests.~~ MERGED

https://issues.jboss.org/browse/WFCORE-759
https://issues.jboss.org/browse/WFCORE-764 is for getting rid of the deprecated methods once this is in a release and full can be adjusted to not use them

### DIFF
--- a/controller/src/main/java/org/jboss/as/controller/SimpleResourceDefinition.java
+++ b/controller/src/main/java/org/jboss/as/controller/SimpleResourceDefinition.java
@@ -56,7 +56,7 @@ public class SimpleResourceDefinition implements ResourceDefinition {
     private final OperationEntry.Flag removeRestartLevel;
     private final boolean runtime;
     private volatile DeprecationData deprecationData;
-    private final boolean orderedChildResource;
+    private final boolean orderedChild;
 
     /**
      * {@link ResourceDefinition} that uses the given {code descriptionProvider} to describe the resource.
@@ -80,7 +80,7 @@ public class SimpleResourceDefinition implements ResourceDefinition {
         this.removeRestartLevel = null;
         this.deprecationData = null;
         this.runtime = false;
-        this.orderedChildResource = isOrderedChildResource();
+        this.orderedChild = false;
     }
 
     /**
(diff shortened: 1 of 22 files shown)


controller/src/main/java/org/jboss/as/controller/SimpleResourceDefinition.java
```diff
@@ -225,6 +261,29 @@ public SimpleResourceDefinition(final PathElement pathElement, final ResourceDes
...
+     *                            Can be {null}
+     * @param deprecationData     Information describing deprecation of this resource. Can be {@code null} if the resource isn't deprecated.
+     * @param runtime             Whether this is a runtime resource
+     * @param orderedChild  Whether this child type is ordered within the parent
+     * @throws IllegalArgumentException if {@code descriptionResolver} is {@code null}.
+     */
+    public SimpleResourceDefinition(final PathElement pathElement, final ResourceDescriptionResolver descriptionResolver,
+                                    final OperationStepHandler addHandler, final OperationStepHandler removeHandler,
+                                    final OperationEntry.Flag addRestartLevel, final OperationEntry.Flag removeRestartLevel,
+                                    final DeprecationData deprecationData, final boolean runtime, final boolean orderedChild) {
```
bstansberry: This is starting to look ripe for a builder API instead of more overloaded constructors.

bstansberry: Lets talk about a builder API. If it makes sense (kind of doubtful as this class is extensively subclassed) we can deprecate these new variants, add the builder and remove them by WFCORE 2.0 Beta1.

ctomc: It would make sense but in most cases we extend the SimpleResourceDefinition and pass params in super constructor.

kabir: @ctomc See https://github.com/wildfly/wildfly-core/pull/821

PR #817, opened 2015-06-17
[WFCORE-690] Add logging configuration information to deployments. This helps show runtime information for the configuration deployments are using.

jamezp:
Adds a runtime resource to a deployment in an attempt to display the logging configuration a deployment is using. Note that the resulting configuration may not be used on deployments that provide their own log manager, such as deployments that use logback. The resource displayed for these deployments will likely be the `default` resource.

Example:
```
        {
            "address" => [
                ("deployment" => "ear.ear"),
                ("subdeployment" => "logging-profile.war"),
                ("subsystem" => "logging")
            ],
            "outcome" => "success",
            "result" => {"configuration" => {"ear.ear/META-INF/logging.properties" => {
                "error-manager" => undefined,
                "filter" => undefined,
                "formatter" => {"PATTERN" => {
                    "class-name" => "org.jboss.logmanager.formatters.PatternFormatter",
                    "module" => undefined,
                    "properties" => {"pattern" => "[DEPLOYMENT] %d{HH:mm:ss,SSS} %-5p [%c] %s%e%n"}
                }},
                "handler" => {"CONSOLE" => {
                    "class-name" => "org.jboss.logmanager.handlers.FileHandler",
                    "encoding" => undefined,
                    "error-manager" => undefined,
                    "filter" => undefined,
                    "formatter" => "PATTERN",
                    "handlers" => [],
                    "level" => "INFO",
                    "module" => undefined,
                    "properties" => {
                        "autoFlush" => "true",
                        "append" => "false",
                        "fileName" => "/home/jperkins/projects/jboss/wildfly/wildfly/build/target/wildfly-10.0.0.Alpha5-SNAPSHOT/standalone/log/test.log"
                    }
                }},
                "logger" => {"ROOT" => {
                    "filter" => undefined,
                    "handlers" => ["CONSOLE"],
                    "level" => "INFO",
                    "use-parent-handlers" => undefined
                }},
                "pojo" => undefined
            }}}
        },
        {
            "address" => [
                ("deployment" => "ear.ear"),
                ("subdeployment" => "log4j.war"),
                ("subsystem" => "logging")
            ],
            "outcome" => "success",
            "result" => {"configuration" => {"log4j.war/WEB-INF/classes/log4j.properties" => {
                "error-manager" => undefined,
                "filter" => undefined,
                "formatter" => undefined,
                "handler" => undefined,
                "logger" => undefined,
                "pojo" => undefined
            }}}
        }
```

In the above example the deployment, `ear.ear`, has defined a logging-profile. The `logging-profile.war` sub-deployment uses this configuration where as the `log4j.war` uses it's own. 

The `configuration` value also shows which configuration the deployment, or sub-deployment, is using. The values are 
  * `default` if the deployment is using the logging subsystem configuration
  * `profile-$PROFILE_NAME` if the deployment is using a logging profile
  * Or the path the configuration file being used

It may make sense to remove the `profile-` prefix but seemed it would be useful at the same time to see which deployments may be using a logging-profiles.

### DIFF
--- a/logging/src/main/java/org/jboss/as/logging/LoggingExtension.java
+++ b/logging/src/main/java/org/jboss/as/logging/LoggingExtension.java
@@ -50,6 +50,7 @@ import org.jboss.as.controller.transform.description.ChainedTransformationDescri
 import org.jboss.as.controller.transform.description.ResourceTransformationDescriptionBuilder;
 import org.jboss.as.controller.transform.description.TransformationDescriptionBuilder;
 import org.jboss.as.logging.LoggingProfileOperations.LoggingProfileAdd;
+import org.jboss.as.logging.deployments.resources.LoggingDeploymentResources;
 import org.jboss.as.logging.logging.LoggingLogger;
 import org.jboss.as.logging.logmanager.WildFlyLogContextSelector;
 import org.jboss.as.logging.stdio.LogContextStdioContextSelector;
@@ -69,7 +70,7 @@ public class LoggingExtension implements Extension {
 
     private static final String RESOURCE_NAME = LoggingExtension.class.getPackage().getName() + ".LocalDescriptions";
 
-    static final String SUBSYSTEM_NAME = "logging";
+    public static final String SUBSYSTEM_NAME = "logging";
 
     static final PathElement LOGGING_PROFILE_PATH = PathElement.pathElement(CommonAttributes.LOGGING_PROFILE);
 
@@ -104,7 +105,7 @@ public class LoggingExtension implements Extension {
      *
      * @return the resolver
      */
-    static ResourceDescriptionResolver getResourceDescriptionResolver(final String... keyPrefix) {
+    public static ResourceDescriptionResolver getResourceDescriptionResolver(final String... keyPrefix) {
         StringBuilder prefix = new StringBuilder(SUBSYSTEM_NAME);
         for (String kp : keyPrefix) {
             prefix.append('.').append(kp);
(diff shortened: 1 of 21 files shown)


bstansberry:
This address isn't ideal:

"configuration" => {"/content/ear.ear/META-INF/logging.properties"

First, it's not an absolute path.

Second, the "/content/" part is an internal detail that can't escape into API like this. (People may come to count on the addressing pattern, so it becomes API.)

LoggingConfigDeploymentProcessor.processDeployment has a "ResourceRoot" param. Looking at it briefly it seems like it should be possible to work out a path that starts with the deployment. The findConfigFile(ResourceRoot) method looks like it assumes the config will be in a META-INF child of resourceRoot.getRoot().

bstansberry:
Please check the attribute definitions you've added. I believe in some cases "undefined" may be a valid value but the "allowNull" property on the AD is not set to true.

dmlloyd:
I think that only one logging configuration can be active per deployment unit, right?  If so, you really only need to identify the deployment itself.  The relative path could just be an attribute.

jamezp:
I'm using the path to the configuration file as the address value purely for convenience. It could be an attribute with some default value for the address or just the name of the configuration file with the relative path.

WRT the `content` portion of the path I mainly left it because I couldn't think of a good way to remove it. It's easy enough to just remove the first parent, but if that ever changes it would be problematic. I guess though I could just examine the address and look for a deployment name. I think it's important to have the deployment name in cases like the example where a WAR is getting the logging configuration from it's parent EAR.

Long story sort, I'll look at removing the unimportant parts of the path to the configuration file.

jamezp:
I've updated the original comment to reflect the new resource name.

PR #819, opened 2015-06-17
[WFCORE-765] ServerDomainProcessShutdownHandler shouldn't fail if shu…

bstansberry:
…tdown is already underway


bstansberry:
retest this please

PR #821, opened 2015-06-17
[WFCORE-767] Introduce parameters object for SimpleResourceDefinition

kabir:
As Brian points out on the PR comment referenced in the description of https://issues.jboss.org/browse/WFCORE-767.

### DIFF
--- a/controller/src/main/java/org/jboss/as/controller/SimpleResourceDefinition.java
+++ b/controller/src/main/java/org/jboss/as/controller/SimpleResourceDefinition.java
@@ -64,10 +64,11 @@ public class SimpleResourceDefinition implements ResourceDefinition {
      * @param pathElement         the path. Can be {@code null}.
      * @param descriptionProvider the description provider. Cannot be {@code null}
      * @throws IllegalArgumentException if {@code descriptionProvider} is {@code null}.
-     * @deprecated
+     * @deprecated Use {@link #SimpleResourceDefinition(Parameters)}
      */
     @Deprecated
     public SimpleResourceDefinition(final PathElement pathElement, final DescriptionProvider descriptionProvider) {
+        //Can be removed when we get to 3.0.0
         if (descriptionProvider == null) {
             throw ControllerLogger.ROOT_LOGGER.nullVar("descriptionProvider");
         }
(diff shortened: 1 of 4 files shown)


controller/src/main/java/org/jboss/as/controller/SimpleResourceDefinition.java
```diff
@@ -64,10 +64,11 @@
      * @param pathElement         the path. Can be {@code null}.
      * @param descriptionProvider the description provider. Cannot be {@code null}
      * @throws IllegalArgumentException if {@code descriptionProvider} is {@code null}.
-     * @deprecated
+     * @deprecated Use {@link #SimpleResourceDefinition(Parameters)}
      */
     @Deprecated
     public SimpleResourceDefinition(final PathElement pathElement, final DescriptionProvider descriptionProvider) {
+        //Can be removed when we get to 3.0.0
```
ctomc: I think we should deprecate it in 1.x first so we can remove it in 2.x already. otherwise we will be stuck with it for 10 years.
we haven't done 1.0.0.Final yet, so it is not too late.

controller/src/main/java/org/jboss/as/controller/SimpleResourceDefinition.java
```diff
@@ -462,4 +461,136 @@ protected boolean isOrderedChildResource() {
...
+         * Call to indicate that a resource is of a type where ordering matters amongst the siblings of the same type.
+         * If not called, the default is {@code false}.
+         *
+         * @return this Parameters object
+         */
+
+        public Parameters setOrderedChild() {
+            this.orderedChildResource = true;
+            return this;
+        }
```
ctomc: we could probably add build() method here as well, that could help create resource definition directly.
it could be useful when you don't extend the class

ctomc:
Maybe this build should be on ResourceDefinition directly so it could be easier to use it in all implementations.

ctomc:
Also while you are here, can you also add builder variant to constructor of PersistentResourceDefinition?

controller/src/main/java/org/jboss/as/controller/SimpleResourceDefinition.java
bstansberry: That will require rewriting a ton of existing subclasses, and we don't have time for that.

ctomc: sure, but having possibility to remove it is always better than not having it.
If we are deprecating it anyway lets just do it in 1.x branch as well.

ctomc:
retest this please

kabir:
@ctomc This has been merged as-is, if you want to implement your ideas, please open a PR

PR #825, opened 2015-06-17
Add wildfly-common as a module in the core feature pack

dmlloyd:


ctomc:
and i worked so hard to remove jboss-common-core damn :)

PR #828, opened 2015-06-19
Get framework set up for testing the current client against old versions

kabir:
Refactor the Server in test-runner to not use static fields, so that we
can change the values.

I have set up a job http://brontes.lab.eng.brq.redhat.com/viewType.html?buildTypeId=WildFlyCore_ClientOldServerLinux which passes using my branch. Once merged, that job needs changing to run off master.


bstansberry:
Looks good to me (although I still don't get why you want to actually deploy/undeploy). If you're comfortable with CI being ready for it, please feel free to merge.

PR #829, opened 2015-06-19
Update README.md (current build need java 8)

kylinsoong:
current build need java 8, rather than java 7


bstansberry:
this is ok to test

PR #831, opened 2015-06-20
WFCORE-570 Provide ability to display module dependencies via mgmt API

ctomc:
https://issues.jboss.org/browse/WFCORE-570

### DIFF
--- a/server/src/main/java/org/jboss/as/server/controller/resources/ModuleInfoHandler.java
+++ b/server/src/main/java/org/jboss/as/server/controller/resources/ModuleInfoHandler.java
@@ -1,0 +1,214 @@
+/*
+ * JBoss, Home of Professional Open Source
+ * Copyright 2015, Red Hat, Inc., and individual contributors as indicated
+ * by the @authors tag.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.jboss.as.server.controller.resources;
+
+import java.lang.reflect.Field;
+
+import org.jboss.as.controller.AttributeDefinition;
+import org.jboss.as.controller.ObjectListAttributeDefinition;
+import org.jboss.as.controller.ObjectTypeAttributeDefinition;
+import org.jboss.as.controller.OperationContext;
+import org.jboss.as.controller.OperationDefinition;
+import org.jboss.as.controller.OperationFailedException;
+import org.jboss.as.controller.OperationStepHandler;
+import org.jboss.as.controller.SimpleAttributeDefinitionBuilder;
+import org.jboss.as.controller.SimpleOperationDefinitionBuilder;
+import org.jboss.as.controller.StringListAttributeDefinition;
+import org.jboss.as.server.logging.ServerLogger;
+import org.jboss.dmr.ModelNode;
+import org.jboss.dmr.ModelType;
+import org.jboss.modules.Module;
+import org.jboss.modules.ModuleIdentifier;
+import org.jboss.modules.ModuleLoader;
+import org.jboss.modules.management.DependencyInfo;
+import org.jboss.modules.management.ModuleInfo;
+import org.jboss.modules.management.ModuleLoaderMXBean;
+import org.jboss.modules.management.ResourceLoaderInfo;
+
+/**
+ * @author Tomaz Cerar (c) 2015 Red Hat Inc.
+ */
+public class ModuleInfoHandler implements OperationStepHandler {
+
+
+    static final AttributeDefinition MODULE_NAME = SimpleAttributeDefinitionBuilder.create("name", ModelType.STRING)
+            .setAllowNull(false)
+            .build();
+    static final AttributeDefinition MODULE_SLOT = SimpleAttributeDefinitionBuilder.create("slot", ModelType.STRING)
+            .setDefaultValue(new ModelNode("main"))
+            .setAllowNull(true)
+            .build();
+
+    //reply parameters
+    static final AttributeDefinition MAIN_CLASS = SimpleAttributeDefinitionBuilder.create("main-class", ModelType.STRING, true)
+            .setStorageRuntime()
+            .build();
+
+    static final AttributeDefinition FALLBACK_LOADER = SimpleAttributeDefinitionBuilder.create("fallback-loader", ModelType.STRING, true)
+            .setStorageRuntime()
+            .build();
+
+    static final ObjectTypeAttributeDefinition DEPENDENCY = ObjectTypeAttributeDefinition.Builder.of("dependency",
+            SimpleAttributeDefinitionBuilder.create("dependency-name", ModelType.STRING).build(),
+            SimpleAttributeDefinitionBuilder.create("module-name", ModelType.STRING).build(),
+            SimpleAttributeDefinitionBuilder.create("export-filter", ModelType.STRING).build(),
+            SimpleAttributeDefinitionBuilder.create("import-filter", ModelType.STRING).build(),
+            SimpleAttributeDefinitionBuilder.create("optional", ModelType.BOOLEAN).build(),
+            SimpleAttributeDefinitionBuilder.create("local-loader-class", ModelType.STRING, true).build(),
+            new StringListAttributeDefinition.Builder("local-loader-paths").build()
+    )
+            .setStorageRuntime()
+            .build();
+
+    static final AttributeDefinition DEPENDENCIES = ObjectListAttributeDefinition.Builder.of("dependencies", DEPENDENCY)
+            .setStorageRuntime()
+            .build();
+
+    static final ObjectTypeAttributeDefinition RESOURCE_LOADER = ObjectTypeAttributeDefinition.Builder.of("resource-loader",
+            SimpleAttributeDefinitionBuilder.create("type", ModelType.STRING).build(),
+            new StringListAttributeDefinition.Builder("paths").build()
+
+    )
+            .setStorageRuntime()
+            .build();
+
+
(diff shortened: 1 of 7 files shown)


controller/src/main/java/org/jboss/as/controller/AbstractOperationStepHandler.java
```diff
@@ -0,0 +1,54 @@
...
+
+    public AbstractOperationStepHandler(AttributeDefinition... parameters) {
+        this.attributes = Arrays.asList(parameters);
+    }
+
+    public AbstractOperationStepHandler(OperationDefinition operationDefinition) {
+        this.attributes = Arrays.asList(operationDefinition.getParameters());
+    }
+
+    protected void populateModel(final ModelNode operation, final ModelNode model) throws OperationFailedException {
```
kabir: To be nit-picky, perhaps something like validateModel instead? Since we're not actually populating anything really

server/src/main/java/org/jboss/as/server/controller/resources/ModuleInfoHandler.java
```diff
@@ -0,0 +1,202 @@
...
+            ModuleInfo moduleInfo = mxBean.getModuleDescription(id.toString());
+            context.getResult().set(displayModuleInfo(moduleInfo));
+        } catch (Exception e) {
+            throw new OperationFailedException("Could not get module info", e);
+        }
+    }
+
+
+    private ModuleLoaderMXBean getMxBean(ModuleLoader loader) throws ReflectiveOperationException {
+        Field mxBeanField = ModuleLoader.class.getDeclaredField("mxBean");
```
kabir: I've not looked too carefully, but is there any reason not to call this via the platform mbean server rather than relying on a private field?

controller/src/main/java/org/jboss/as/controller/AbstractOperationStepHandler.java
ctomc: We are populating the model not just validating it. This is the same behavior we have in AbstractAddStepHandler #populateModel() which is what i was going for here.
OOTB we do it for you, or you can override it if you need some special module population.

server/src/main/java/org/jboss/as/server/controller/resources/ModuleInfoHandler.java
ctomc: Will that work even when there are no subsystems configured?

server/src/main/java/org/jboss/as/server/controller/resources/ModuleInfoHandler.java
```diff
@@ -0,0 +1,203 @@
...
+
+     */
+    private ModelNode displayModuleInfo(ModuleInfo module) throws Exception {
+        ModelNode result = new ModelNode();
+        result.get("name").set(module.getName());
+
+        if (module.getMainClass() != null) {
+            result.get("main-class").set(module.getMainClass() == null ? "<not defined>" : module.getMainClass());
+        }
+        result.get("fallback-loader").set(module.getFallbackLoader() == null ? "<not defined>" : module.getFallbackLoader());
```
jamezp: Should these `<not defined>` messages be internationalized?

controller/src/main/java/org/jboss/as/controller/AbstractOperationStepHandler.java
bstansberry: Where else is this abstraction going to be used? The subclass in this PR isn't actually using the abstraction, as everything it does could just go in execute(...).

bstansberry: Please get rid of this. It's not used, isn't documented, and the implementation doesn't work the same way as other abstract OSHs in the package that use the same method names, e.g. performRuntime here is called in Stage.MODEL.

server/src/main/java/org/jboss/as/server/controller/resources/ModuleInfoHandler.java
ctomc: fixed

controller/src/main/java/org/jboss/as/controller/AbstractOperationStepHandler.java
ctomc: @bstansberry Removed the whole abstraction. I thought it was a good a idea at the time but it really wasn't.

server/src/main/java/org/jboss/as/server/controller/resources/ModuleInfoHandler.java
```diff
@@ -0,0 +1,209 @@
...
+        The deployment module loader
+
+
+     */
+    private ModelNode displayModuleInfo(ModuleInfo module) throws Exception {
+        ModelNode result = new ModelNode();
+        result.get("name").set(module.getName());
+
+        if (module.getMainClass() != null) {
+            result.get("main-class").set(module.getMainClass());
```
bstansberry: MAIN_CLASS should be configured to allowNull.

server/src/main/java/org/jboss/as/server/controller/resources/ModuleInfoHandler.java
```diff
@@ -0,0 +1,209 @@
...
+        }
+        if (module.getFallbackLoader() != null) {
+            result.get("fallback-loader").set(module.getFallbackLoader());
```
bstansberry: FALLBACK_LOADER should be configured to allowNull.

server/src/main/java/org/jboss/as/server/controller/resources/ModuleInfoHandler.java
```diff
@@ -0,0 +1,209 @@
...
+                dependency.get("local-loader-class").set(dependencySpec.getLocalLoader());
+            }
+            if (dependencySpec.getLocalLoaderPaths() != null) {
+                ModelNode paths = dependency.get("local-loader-paths");
+                for (String path : dependencySpec.getLocalLoaderPaths()) {
+                    paths.add(path);
+                }
+            }
+        }
+        ModelNode resourceLoaders = result.get("resource-loaders");
```
bstansberry: This should be ModelNode resourceLoaders = result.get("resource-loaders").setEmptyList(), or RESOURCE_LOADERS should be configured to allowNull.

I think the former is better.

server/src/main/java/org/jboss/as/server/controller/resources/ModuleInfoHandler.java
```diff
@@ -0,0 +1,209 @@
...
+        result.get("name").set(module.getName());
+
+        if (module.getMainClass() != null) {
+            result.get("main-class").set(module.getMainClass());
+        }
+        if (module.getFallbackLoader() != null) {
+            result.get("fallback-loader").set(module.getFallbackLoader());
+        }
+
+        ModelNode dependencies = result.get("dependencies");
```
bstansberry: This should be ModelNode dependencies = result.get("dependencies").setEmptyList(), or DEPENDENCIES should be configured to allowNull.

I think the former is better.

server/src/main/java/org/jboss/as/server/controller/resources/ModuleInfoHandler.java
```diff
@@ -0,0 +1,209 @@
...
+    static final AttributeDefinition FALLBACK_LOADER = SimpleAttributeDefinitionBuilder.create("fallback-loader", ModelType.STRING)
+            .setStorageRuntime()
+            .build();
+
+    static final ObjectTypeAttributeDefinition DEPENDENCY = ObjectTypeAttributeDefinition.Builder.of("dependency",
+            SimpleAttributeDefinitionBuilder.create("dependency-name", ModelType.STRING).build(),
+            SimpleAttributeDefinitionBuilder.create("module-name", ModelType.STRING).build(),
+            SimpleAttributeDefinitionBuilder.create("export-filter", ModelType.STRING).build(),
+            SimpleAttributeDefinitionBuilder.create("import-filter", ModelType.STRING).build(),
+            SimpleAttributeDefinitionBuilder.create("optional", ModelType.STRING).build(),
```
bstansberry: This should be ModelType.BOOLEAN.

server/src/main/java/org/jboss/as/server/controller/resources/ModuleInfoHandler.java
```diff
@@ -0,0 +1,209 @@
...
+                continue; //todo check why it returns empty dependancy
+            }
+            ModelNode dependency = dependencies.add();
+            dependency.get("dependency-name").set(dependencySpec.getDependencyType());
+            dependency.get("module-name").set(dependencySpec.getModuleName());
+            dependency.get("export-filter").set(dependencySpec.getExportFilter());
+            dependency.get("import-filter").set(dependencySpec.getImportFilter());
+            dependency.get("optional").set(dependencySpec.isOptional());
+            if (dependencySpec.getLocalLoader() != null) {
+                dependency.get("local-loader-class").set(dependencySpec.getLocalLoader());
```
bstansberry: The field in DEPENDENCY for this should be configured to allowNull.

server/src/main/java/org/jboss/as/server/controller/resources/ModuleInfoHandler.java
```diff
@@ -0,0 +1,209 @@
...
+            }
+            if (dependencySpec.getLocalLoaderPaths() != null) {
```
bstansberry: The field in DEPENDENCY for this should be configured to allowNull.

server/src/main/java/org/jboss/as/server/controller/resources/ModuleInfoHandler.java
bstansberry: @kabir please confirm:

Working with the platform mbean server via the JDK APIs should be fine regardless of the presence of the jmx subsystem. There's a platform mbean server either way, and the additional capabilities added by the jmx subsystem (management resource facade, RBAC, JSR-77 integration) are not needed for this.

server/src/main/java/org/jboss/as/server/controller/resources/ModuleInfoHandler.java
ctomc: problem is more in how ModelNode works, you cannot say .set(null) as it i will complain with IAE.
I could have it undefined if it is null.

server/src/main/java/org/jboss/as/server/controller/resources/ModuleInfoHandler.java
bstansberry: In today's Domain Management call David said he's fine with the reflection approach.

PR #832, opened 2015-06-22
WFCORE-620 : JBoss Product Installation Report

ehsavoie:
Making WFCORE a 'product'
Using server environment
Fixing tests
Adding some infos about the last update date.
Adding Organization Identifier field in model.
Making the whole operation running on /host=*

The operation from the cli allowing to download the report(s) has been deactivated.

Jira: https://issues.jboss.org/browse/WFCORE-620


bstansberry:
I made a couple comments on the 1st commit; I'll redo those on the complete diff instead and go from there.

cli/src/main/java/org/jboss/as/cli/handlers/BaseOperationCommand.java
```diff
@@ -260,11 +269,16 @@ protected void addHeaders(CommandContext ctx, ModelNode request) throws CommandF
...
+    protected void handleResponse(CommandContext ctx, OperationResponse response, boolean composite) throws CommandLineException {
         displayResponseHeaders(ctx, response);
+        handleAttachedFile(ctx, response);
+    }
+
+    protected void handleAttachedFile(CommandContext ctx, OperationResponse operationResponse) throws CommandLineException {
     }
 
-    protected void displayResponseHeaders(CommandContext ctx, ModelNode response) {
+    protected void displayResponseHeaders(CommandContext ctx, OperationResponse operationResponse) {
```
bstansberry: Is it necessary to change this signature? This is somewhat a public API, as external extensions can write handlers, so we shouldn't change API if it's not required. I don't think any impl of this would want to do anything with the operationResponse besides call getResponseNode().

cli/src/main/java/org/jboss/as/cli/handlers/DeploymentInfoHandler.java
```diff
@@ -226,8 +227,6 @@ public ModelNode buildRequestWithoutHeaders(CommandContext ctx) throws CommandFo
                         steps.add(step);
                     }
                 }
-            } else if(serverGroupName.indexOf('*') >= 0) {
-                  throw new CommandFormatException("Wildcards are not permited here, please give a valid server-group name.");
```
bstansberry: Why? Possible rebasing mistake?

controller/src/main/java/org/jboss/as/controller/AttributeMarshaller.java
```diff
@@ -105,6 +105,33 @@ protected String asString(ModelNode value) {
...
+            this.marshallSimpleTypeAsAttribute = marshallSimpleTypeAsAttribute;
+        }
+
+        @Override
+        public void marshallAsElement(AttributeDefinition attribute, ModelNode resourceModel, boolean marshallDefault, XMLStreamWriter writer) throws XMLStreamException {
+            if (resourceModel.hasDefined(attribute.getName()) && attribute instanceof ObjectTypeAttributeDefinition) {
+                AttributeDefinition[] valueTypes = ((ObjectTypeAttributeDefinition) attribute).getValueTypes();
+                writer.writeStartElement(attribute.getXmlName());
+                for (AttributeDefinition valueType : valueTypes) {
+                    if(resourceModel.get(attribute.getName()).hasDefined(valueType.getName())) {
```
bstansberry: Use the new ModelNode op:

```
if (resourceMode.hasDefined(attribute.getName(), valueType.getName())) {
```

That call has no chance of mutating a ModelNode where doing so isn't wanted, so I want us to start using it.

controller/src/main/java/org/jboss/as/controller/AttributeMarshaller.java
```diff
@@ -105,6 +105,33 @@ protected String asString(ModelNode value) {
...
+                        ModelNode handler = resourceModel.get(attribute.getName());
+                        if(!marshallSimpleTypeAsAttribute) {
```
bstansberry: I find this variable name confusing given it results in a call to "marshallAsElement". It needs comments, and the Javadoc for ELEMENT_ONLY_OBJECT and ATTRIBUTE_OBJECT needs to be a lot more explicit about what results will happen if those marshallers are used.

controller/src/main/java/org/jboss/as/controller/ObjectTypeAttributeDefinition.java
```diff
@@ -108,6 +90,10 @@ protected ModelNode convertParameterExpressions(ModelNode parameter) {
         return result;
     }
 
+    public AttributeDefinition[] getValueTypes() {
```
bstansberry: This doesn't need to be public; package protected is fine.

controller/src/main/java/org/jboss/as/controller/AttributeMarshaller.java
```diff
@@ -105,6 +105,33 @@ protected String asString(ModelNode value) {
...
+    private static class ObjectMarshaller extends DefaultAttributeMarshaller {
+        private final boolean marshallSimpleTypeAsAttribute;
+
+        private ObjectMarshaller(boolean marshallSimpleTypeAsAttribute) {
+            this.marshallSimpleTypeAsAttribute = marshallSimpleTypeAsAttribute;
+        }
+
+        @Override
+        public void marshallAsElement(AttributeDefinition attribute, ModelNode resourceModel, boolean marshallDefault, XMLStreamWriter writer) throws XMLStreamException {
+            if (resourceModel.hasDefined(attribute.getName()) && attribute instanceof ObjectTypeAttributeDefinition) {
```
bstansberry: Should there be an assert here re: the type of 'attribute'? Documented in the javadoc for the constants?

Perhaps this class and the constants should move to ObjectTypeAttributeDefinition, since the impls are tightly coupled. The ListMarshaller class in AttributeMarshaller isn't tied that way. It probably wouldn't be used except with some variant of ListAttributeDefinition, but it could be.

controller/src/main/java/org/jboss/as/controller/logging/ControllerLogger.java
```diff
@@ -3349,4 +3349,7 @@ XMLStreamException discoveryOptionsMustBeDeclared(String adminOnlyCmd, String po
 
     @Message(id = 404, value = "The binding name '%s' in socket binding group '%s' is not unique. Names must be unique across socket-binding, local-destination-outbound-socket-binding and remote-destination-outbound-socket-binding")
     OperationFailedException bindingNameNotUnique(String name, String groupName);
+
+    @Message(id = 405, value = "Couldn't build the report")
+    OperationFailedException failedToBuildReport(@Cause Throwable t);
```
bstansberry: This should return RuntimeException. An OFE is a form of OperationClientException and shouldn't be thrown except in cases where the client has made a mistake in their request. An OSH should throw a RuntimeException if it encounters an unexpected server fault not due to client error.

AbstractOperationContext.executeStep counts on this in its error handling; e.g. after boot it will only log an OFE in the server log at DEBUG.

cli/src/main/java/org/jboss/as/cli/handlers/BaseOperationCommand.java
```diff
@@ -260,11 +269,16 @@ protected void addHeaders(CommandContext ctx, ModelNode request) throws CommandF
         opHeaders.set(headersNode);
     }
 
-    protected void handleResponse(CommandContext ctx, ModelNode response, boolean composite) throws CommandLineException {
+    protected void handleResponse(CommandContext ctx, OperationResponse response, boolean composite) throws CommandLineException {
```
bstansberry: Please ping Alexey or Joe Wertz about this. I have the same concern here about an API change as I have with displayResponseHeaders, and this looks to be a bigger issue as the method is overridden quite a bit.

Perhaps add this method and leave the existing one, and then call to the existing one before calling handleAttachedFile. That's a more complicated API but doesn't break anyone. That's what I've done to evolve controller classes like AbstractAddStepHandler.

controller/src/main/resources/org/jboss/as/controller/descriptions/common/LocalDescriptions.properties
```diff
@@ -531,6 +531,30 @@ global.query.where=a list of filter criteria tuples (i.e. 'running=true')
...
+global.summary.java-version=The Java language version.
+global.summary.jvm=The JVM configuration for the server.
+global.summary.jvm-version=The complete build version of the JVM.
+global.summary.jvm-vendor=The vendor of the JVM.
+global.summary.last-update-date=Date of the last update of the current product.
+global.summary.node-name=The name of the instance node.
+global.summary.product-community-identifier=Indicates the type of distribution, if it is a Product or a Project.
+global.summary.product-home=Installation directory of the current server instance.
+global.summary.product-name=The name of the WildFly based product that is being run by this server.
+global.summary.product-version=The version of the WildFly based product release that is being run by this server.
```
bstansberry: Please remove "WildFly based". The info it adds isn't worth the possible hassle to remove it if someone wants it gone.

host-controller/src/main/java/org/jboss/as/host/controller/model/host/HostResourceDefinition.java
```diff
@@ -160,6 +162,9 @@
             .setValidator(new StringLengthValidator(1, true))
             .setStorageRuntime()
             .build();
+    public static final SimpleAttributeDefinition ORGANIZATION_IDENTIFIER = SimpleAttributeDefinitionBuilder.create(ModelDescriptionConstants.ORGANIZATION, ModelType.STRING, true)
+            .setValidator(new StringLengthValidator(1, true))
+            .build();
```
bstansberry: Should this be on DomainRootDefinition?

In theory different hosts could have different "organizations" since it's a term that means what the user wants. It would be annoying to have to set it in every host.xml though. I'm not sure how hard it would be to have both here and DomainRootDefinition, with the DomainRootDefinition value only being used in a host/domain server report if the host level value wasn't set.

host-controller/src/main/java/org/jboss/as/domain/controller/resources/GlobalHostResourceDefinition.java
```diff
@@ -0,0 +1,48 @@
...
+        super(PathElement.pathElement(HOST), HostModelUtil.getResourceDescriptionResolver());
+    }
+
+    @Override
+    public void registerOperations(ManagementResourceRegistration resourceRegistration) {
+        super.registerOperations(resourceRegistration);
+        resourceRegistration.registerOperationHandler(GlobalInstallationReportHandler.DEFINITION, GlobalInstallationReportHandler.INSTANCE);
+    }
+
+}
```
bstansberry: I'm quite certain this isn't something we want to do. Why doesn't the existing stuff for registering host stuff work?

server/src/main/java/org/jboss/as/server/operations/ProductEnvironment.java
```diff
@@ -0,0 +1,73 @@
...
+
+import java.util.UUID;
+import org.jboss.as.controller.RunningModeControl;
+import org.jboss.as.version.ProductConfig;
+
+/**
+ * Encapsulates the runtime environment for a product.
+ * @author <a href="mailto:ehugonne@redhat.com">Emmanuel Hugonnet</a> (c) 2015 Red Hat, inc.
+ */
+public interface ProductEnvironment {
```
bstansberry: It's tempting to just add these as abstract methods in ProcessEnvironment.

host-controller/src/main/java/org/jboss/as/host/controller/operations/InstallationReportHandler.java
```diff
@@ -0,0 +1,92 @@
...
+
+    @Override
+    public void execute(OperationContext context, ModelNode operation) throws OperationFailedException {
+        final ModelNode patchingInfo = new ModelNode();
+        PathAddress patchingAddress = PathAddress.pathAddress(
+                PathElement.pathElement(HOST, environment.getHostControllerName()),
+                PathElement.pathElement(CORE_SERVICE, "patching"));
+        OperationEntry opEntry = context.getRootResourceRegistration().getOperationEntry(patchingAddress, "show-history");
+        if(opEntry != null) {
+            context.addStep(patchingInfo, Util.createOperation("patch-info", patchingAddress),
```
bstansberry: Shouldn't this be "show-history"? Or L77 should be "patch-info"?

host-controller/src/main/java/org/jboss/as/host/controller/operations/InstallationReportHandler.java
```diff
@@ -0,0 +1,92 @@
...
+                opEntry.getOperationHandler(), OperationContext.Stage.RUNTIME);
+        }
+        final Path installationDir = environment.getHomeDir().toPath();
+        context.addStep(new OperationStepHandler() {
+            @Override
+            public void execute(OperationContext context, ModelNode operation) throws OperationFailedException {
+                System.out.println("Here is what we have as infos " + patchingInfo);
```
bstansberry: :)

server/src/main/java/org/jboss/as/server/operations/AbstractInstallationReporter.java
```diff
@@ -0,0 +1,303 @@
...
+            } else {
+                product.get(PRODUCT_COMMUNITY_IDENTIFIER).set(PROJECT_TYPE);
+            }
+        }
+        if(installationDir != null && Files.exists(installationDir)) {
+            product.get(PRODUCT_HOME).set(installationDir.toAbsolutePath().toString());
+            BasicFileAttributeView view = Files.getFileAttributeView(installationDir, BasicFileAttributeView.class);
+            try {
+                BasicFileAttributes attributes = view.readAttributes();
+                product.get(PRODUCT_INSTALLATION_DATE).set(attributes.creationTime().toString());
```
bstansberry: This may not be accurate. Per the javadoc for it, some filesystems may report last-modified-time instead, others may report something like the epoch.

Probably not worth worrying about though. Only thing I could think of is checking if attributes.creationTime().toMillis() == 0 and not reporting that.

server/src/main/java/org/jboss/as/server/operations/AbstractInstallationReporter.java
```diff
@@ -0,0 +1,303 @@
...
+            if(environment.getProductConfig().getProductVersion() != null) {
+                product.get(PRODUCT_VERSION).set(environment.getProductConfig().getProductVersion());
+            }
+            if (environment.getProductConfig().isProduct()) {
+                product.get(PRODUCT_COMMUNITY_IDENTIFIER).set(PRODUCT_TYPE);
+            } else {
+                product.get(PRODUCT_COMMUNITY_IDENTIFIER).set(PROJECT_TYPE);
+            }
+        }
+        if(installationDir != null && Files.exists(installationDir)) {
```
bstansberry: This needs to be dealt with in the callers to this method, but the wildfly-swarm case needs to be handled. See ServerEnvironment L342 for what I'm getting at:

```

        if ( launchType.equals( LaunchType.SELF_CONTAINED ) ) {
            homeDir = new File(WildFlySecurityManager.getPropertyPrivileged("user.dir", "."));
```

With LaunchType.SELF_CONTAINED (also visible as ProcessType.SELF_CONTAINED) there really is no meaningful installation dir.

server/src/main/java/org/jboss/as/server/operations/AbstractInstallationReporter.java
```diff
@@ -0,0 +1,303 @@
...
+     * @param installationDir installation directory of the current server instance.
+     * @return a ModelNode representing the instance product and the plateform it is running on.
+     * @throws OperationFailedException
+     */
+    protected ModelNode createProductNode(OperationContext context, ProductEnvironment environment,
+            ModelNode patchingInfo, Path installationDir) throws OperationFailedException {
+        assert environment != null;
+        ModelNode product = new ModelNode().setEmptyObject();
+        product.get(REPORT_VERSION).set("1.0");
+        product.get(NODE_NAME).set(environment.getHostName());
```
bstansberry: I don't see this in WFCORE-620.

There is a system property jboss.node.name that has a particular meaning (see ServerEnvironment.getNodeName() and we need to be careful about not confusing things.

server/src/main/java/org/jboss/as/server/operations/AbstractInstallationReporter.java
```diff
@@ -0,0 +1,303 @@
...
+        product.get(HOSTNAME).set(environment.getQualifiedHostName());
```
bstansberry: environment.getQualifiedHostName() may return the values of system properties jboss.qualified.host.name or jboss.host.name, and not necessarily some OS type concept such as the HOSTNAME or COMPUTERNAME environment variables or the result of something like InetAddressUtil.getLocalHost().getHostName().

I'm not sure if that WildFly-specific deviation from any "cross-product" meaning of "Host Name" matters or not.

Using -Djboss.host.name is a common way to assign the name for a host in the management model without having to hard code it in each host's host.xml, so this isn't an edge case.

server/src/main/resources/schema/wildlfy-product-info_1_0.xsd
```diff
@@ -0,0 +1,193 @@
...
+MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the GNU
+Lesser General Public License for more details.
+
+You should have received a copy of the GNU Lesser General Public
+License along with this library; if not, write to the Free Software
+Foundation, Inc., 51 Franklin Street, Fifth Floor, Boston,
+MA 02110-1301  USA
+-->
+
+<xs:schema xmlns:xs="http://www.w3.org/2001/XMLSchema"
```
bstansberry: I haven't checked, but this can't get into any build. Better if it gets deleted in the 2nd commit you have.

controller/src/main/java/org/jboss/as/controller/operations/global/GlobalOperationHandlers.java
```diff
@@ -84,6 +84,7 @@ public static void registerGlobalOperations(ManagementResourceRegistration root,
             root.registerOperationHandler(org.jboss.as.controller.operations.global.ReadAttributeHandler.RESOLVE_DEFINITION,
                     org.jboss.as.controller.operations.global.ReadAttributeHandler.RESOLVE_INSTANCE, true);
             root.registerOperationHandler(ReadAttributeGroupHandler.RESOLVE_DEFINITION, ReadAttributeGroupHandler.RESOLVE_INSTANCE, true);
+            root.registerOperationHandler(GlobalInstallationReportHandler.DEFINITION, GlobalInstallationReportHandler.INSTANCE, false);
```
bstansberry: This doesn't belong here. I don't think the op is meant to be registered (i.e. second commit is to disable the file report) and if it should be registered, ServerRootResourceRegistration is where it should happen.

I think some confusion probably comes from GlobalOperationHandlers.AbstractMultiTargetHandler. That class' functionality can be used for non "global" ops; it just hasn't been.. A "global" op means it exists on every resource.

host-controller/src/main/java/org/jboss/as/host/controller/operations/InstallationReportHandler.java
```diff
@@ -0,0 +1,92 @@
...
+                ModelNode result = context.getResult();
+                result.get(SUMMARY_DEFINITION.getName()).set(createProductNode(context, environment, patchingInfo, installationDir));
```
bstansberry: The .get(SUMMARY_DEFINITION.getName()) shouldn't be necessary. If it is we need to figure out how to fix that.

server/src/main/java/org/jboss/as/server/operations/ProductEnvironment.java
ehsavoie: Ok I gave into the temptation

server/src/main/java/org/jboss/as/server/operations/AbstractInstallationReporter.java
ehsavoie: this value is overloaded in domain mode. Not sure i should keep it there

server/src/main/java/org/jboss/as/server/operations/AbstractInstallationReporter.java
ehsavoie: Yes I agree, but it's a best effort as we don't have this metadata anywhere.

bstansberry:
retest this please

ehsavoie:
retest this please

PR #837, opened 2015-06-24
Use JDK's JAXP / WFLY-4704 Remove Xalan from WildFly

ctomc:
It upgrades jboss-modules to 1.5.0.Beta1
needs https://github.com/wildfly/wildfly/pull/7652 for full


core-feature-pack/src/main/resources/modules/system/layers/base/sun/jdk/main/module.xml
```diff
@@ -52,8 +52,145 @@
                 <path name="com/sun/jndi/url/rmi"/>
                 <path name="com/sun/media/sound"/>
                 <path name="com/sun/crypto/provider"/>
-                <path name="com/sun/org/apache/bcel/internal/classfile"/>
+                <path name="com/sun/org/apache/xml/internal/security/algorithms"/>
```
n1hility: Nothing should be depending on these com/sun classes. Why are they needed?

ctomc: via JAXP factories are resolved and loaded from jdk, and if they are not included in this file they wont work.
I just added all com/sun/org/apache/xml paths here.

ctomc:
As discussed this cannot happen reliably in time for 10. 
But should be revisited after JDK9 comes out.

PR #838, opened 2015-06-25
[WFCORE-779] AbstractAddStepHandler should let AD record default requ…

bstansberry:
…irements


jamezp:
Not really sure why or how, but these test failures on full seem consistent.

bstansberry:
Replaced by #869

PR #843, opened 2015-06-26
Support for WFLY-334 HA Singleton deployer for applications

pferraro:
Introduces configurable strategy for building the next deployment phase service.  A DUP can attach an implementation of DeploymentUnitPhaseBuilder to the DeploymentPhaseContext to manipulate how the service for the next phase is built.

https://issues.jboss.org/browse/WFLY-334

This PR contains requisite changes to wildfly-core for this branch:
https://github.com/pferraro/wildfly/tree/singleton

### DIFF
--- a/server/src/main/java/org/jboss/as/server/deployment/Attachments.java
+++ b/server/src/main/java/org/jboss/as/server/deployment/Attachments.java
@@ -53,6 +53,7 @@ import org.jboss.vfs.VirtualFile;
 /**
  * @author <a href="mailto:david.lloyd@redhat.com">David M. Lloyd</a>
  */
+@SuppressWarnings("deprecation")
 public final class Attachments {
 
     //
@@ -99,11 +100,16 @@ public final class Attachments {
      * @deprecated the object attached under this key does nothing
      */
     @Deprecated
-    @SuppressWarnings("deprecation")
     public static final AttachmentKey<ServiceVerificationHandler> SERVICE_VERIFICATION_HANDLER = AttachmentKey.create(ServiceVerificationHandler.class);
 
 
     public static final AttachmentKey<Boolean> ALLOW_PHASE_RESTART = AttachmentKey.create(Boolean.class);
+
+    /**
+     * A builder used to install a deployment phase
+     */
+    public static final AttachmentKey<DeploymentUnitPhaseBuilder> DEPLOYMENT_UNIT_PHASE_BUILDER = AttachmentKey.create(DeploymentUnitPhaseBuilder.class);
+
     //
     // STRUCTURE
     //
--- a/server/src/main/java/org/jboss/as/server/deployment/DeploymentPhaseContextImpl.java
+++ b/server/src/main/java/org/jboss/as/server/deployment/DeploymentPhaseContextImpl.java
@@ -22,6 +22,8 @@
 
 package org.jboss.as.server.deployment;
 
+import java.util.List;
+
 import org.jboss.msc.inject.Injector;
 import org.jboss.msc.service.ServiceBuilder;
 import org.jboss.msc.service.ServiceName;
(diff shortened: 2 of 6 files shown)


n1hility:
Sorry for the delay on this one. Can you do a design post to wildly-dev for this feature?  One question I have is why intercepting the deployment process is necessary at all, instead of just having the whole deployment in stopped state. It won't improve deployment start by much, since most of the cost is initialization, and it adds a lot of complexity.

pferraro:
That was my initial design (i.e. from 1+ years ago) - to install the parent deployment service as a singleton service, based on a "deployment policy" parameter added to the deployment operation.  I thought that was the cleanest approach, but it required a change to the deployment API, which wasn't seen as welcome at the time.  In response, it was generally agreed by Brian, David, et al (including yourself, as I recall), singleton deployments were best done instrumented via a DUP that reads a deployment descriptor which specifies the deployment policy - which can be either attached as an overlay (as recommended by Stuart at the last F2F meeting) (i.e. singleton-deployment.xml) or embedded within an existing jboss.xml.  The deployment overlay construct fits nicely (i.e. like an aspect), since we can take an existing deployment and change how it is deployed externally. 
If we instead instrumented the singleton service at a coarser level, we still need a mechanism to specify "how" to deploy.  If not a deployment operation parameter, where?
That said, I can certainly reopen the discussion.  As you can see - the requirements affecting wildfly-core are minimal - so it can't hurt to rethink this..

n1hility:
Now that you mention it I have a vague recollection of it. Let me look at your approach in more detail and i will get back to you.

bstansberry:
IIRC there were various concepts tossed around, but a more recent one involved adding a 'policy' attribute to the deployment resource. Simple string, with a policy name. Whatever code could then implement the policy; in this case it would be a clustering subsystem. IIRC how the policy would be implemented looked a lot like this; via a custom builder for the DeploymentPhaseUnitService.

I recall a couple concerns with that:

1) Management clients would have to be adapted to use it. CLI, web console, JBDS, scanner. That's not an issue with a deployment descriptor.

2) The 'policy' attribute would kind of be 'floating in space'. It would be a value pointing to ???. The capabilities and requirements stuff can somewhat help with this, as now the attribute could be declared as configuring a requirement for an instance of dynamic resource 'org.wildfly.deployment.deployment-policy' or something. The clustering subsystem could provide an impl of that capability, and the capability registry, once we expose it, would help show the user how things tie together.

Adding a requirement from a global resource like a deployment to a profile-specific capability will be a new capability resolution challenge though.

A key question I think is how this 'policy' API should work, however it is configured. If the correct API is via DeploymentUnitPhaseBuilder, then it doesn't much matter whether the policy attribute is in the management model or a deployment-descriptor. Either could work, both can work, and we can do one now and if we want, another later.

pferraro:
@bstansberry The approach I've taken in this PR (and the subsequent wildfly PR) does away with any requirements for global resources. Instead, any DUP can manipulate how the next phase of a deployment is built by attaching a DeploymentUnitPhaseBuilder implementation to the deployment phase context.
The intention is to use this here:
https://github.com/pferraro/wildfly/blob/singleton/clustering/singleton/extension/src/main/java/org/wildfly/extension/clustering/singleton/deployment/SingletonDeploymentProcessor.java
where the DeploymentUnitPhaseBuilder will look like:
https://github.com/pferraro/wildfly/blob/singleton/clustering/singleton/extension/src/main/java/org/wildfly/extension/clustering/singleton/deployment/SingletonDeploymentUnitPhaseBuilder.java

bstansberry:
@pferraro Yes, this approach gets rid of any global resources. Sorry, most of my response above was documenting the history of the discussion re: putting this config in the server management model vs putting it in a deployment descriptor. There were some offline chats about that last week and I wanted to write it down formally.

I think the main question is how comfortable Jason is with opening up internal manipulation of the deployment chain the way this PR does with DeploymentUnitPhaseBuilder. If there was no requirement for DUP-based configuration, then in theory things could be a bit more limited; e.g. just a "RootDeploymentUnitServiceBuilder" that doesn't put as much flexibility in the hands of the policy impl.

I doubt that's a big improvement though, and if we want to be conservative we can always deprecate this API or otherwise mark it as "experimental", limiting its use to this HASingleton case.

PR #846, opened 2015-06-29
Remove uses of the legacy service names for SocketBinding and OutboundSocketBinding

bstansberry:
Switch these to the capability-based service names.


bstansberry:
I'll rework this a bit once #856 is merged.

PR #847, opened 2015-06-30
[WFCORE-781] Add ability to use capability-provided integrations from…

bstansberry:
… services and DUPs


bstansberry:
Closing for now while the discussion on http://lists.jboss.org/pipermail/wildfly-dev/2015-June/004159.html plays out.

PR #848, opened 2015-06-30
Introduce threads xsd 2.0 which is using PRDefinition and new parsers

ctomc:
this makes it easier for inclusion by other subsystems that use resources from threads.

this version doesn't break wildfly-full

### DIFF
--- a/threads/src/main/resources/schema/wildfly-threads_2_0.xsd
+++ b/threads/src/main/resources/schema/wildfly-threads_2_0.xsd
@@ -1,0 +1,291 @@
+<?xml version="1.0" encoding="UTF-8"?>
+
+<!--
+  ~ JBoss, Home of Professional Open Source
+  ~ Copyright 2015, Red Hat, Inc., and individual contributors as indicated
+  ~ by the @authors tag.
+  ~
+  ~ Licensed under the Apache License, Version 2.0 (the "License");
+  ~ you may not use this file except in compliance with the License.
+  ~ You may obtain a copy of the License at
+  ~
+  ~ http://www.apache.org/licenses/LICENSE-2.0
+  ~
+  ~ Unless required by applicable law or agreed to in writing, software
+  ~ distributed under the License is distributed on an "AS IS" BASIS,
+  ~ WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+  ~ See the License for the specific language governing permissions and
+  ~ limitations under the License.
+  -->
+
+<xs:schema xmlns:xs="http://www.w3.org/2001/XMLSchema"
+           targetNamespace="urn:jboss:domain:threads:2.0"
+           xmlns="urn:jboss:domain:threads:2.0"
+           elementFormDefault="qualified"
+           attributeFormDefault="unqualified"
+           version="1.0">
+
+    <!-- The threads subsystem root element -->
+    <xs:element name="subsystem" type="subsystem"/>
+
+    <xs:complexType name="subsystem">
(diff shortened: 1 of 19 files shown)


bstansberry:
Why is the registerRuntimeOnly stuff dropped?

ctomc:
@bstansberry because it was wrongly used, attributes should be registered in any case.
This way metrics attributes ware not even present on resources unless state of server was runtime.
Original problem probably originates from PoolAttributeDefinitions metric attributes not being marked as storage runtime.

bstansberry:
Whether a metric attribute should be registered on an admin-only server is a good question. I believe not registering them is common. If you register them, they need to state they allow null, since now they do.

Unless I'm mistaken and the pattern of not registering them isn't followed elsewhere, it's a good topic for the dev list.

OTOH, I see no reason to register a metric attribute for a domain /profile=x/.... resource. Those are configuration-only resources.

ctomc:
It is a interesting topic for sure. We used to use this pattern of registering metrics only when runtime registration is needed.
But lately most of the code registers all attributes no matter what is the registration stage. Which is something i prefer as well.
Will send mail to dev list about this.

bstansberry:
WFCORE-831 describes what came out of that thread. Given that, a parameter to control whether metrics should be registered is still needed, as they should not be registered in a typical Host Controller /profile=*/subsystem=y situation. The value for that parameter would be driven according to the WFCORE-831 solution.

ctomc:
@bstansberry changes required for WFCORE-831 ware merged and PR rebased. 
I think this should do it now.

bstansberry:
My last comment still applies.

These metrics should not be registered on a Host Controller, because they are not a valid part of the API of a /profile=*/subsystem=x resource. A /profile=*/subsystem=x resource on an HC is not the same thing a a /subsystem=x resource on a server.

Kabir's WFCORE-831 work changes what result you get from ExtensionContextisRuntimeOnlyRegistrationValid() in the --admin-only server case, but it does not always return 'true'.

ctomc:
Well, I assumed wrongly that changes done by WFCORE-831 would also take care of what you are talking about.
That way subsystem developer wouldn't need to worry about conditionally registering attributes/metrics.

In any case it is probably too late to change that, so I will rework this PR to still consider runtimeRegistration

ctomc:
@bstansberry reworked to still use runtimeRegistration check where needed.

threads/src/main/resources/schema/wildfly-threads_2_0.xsd
```diff
@@ -0,0 +1,310 @@
...
+            ]]>
+            </xs:documentation>
+        </xs:annotation>
+        <xs:restriction base="xs:integer">
+            <xs:minInclusive value="1"/>
+            <xs:maxInclusive value="10"/>
+        </xs:restriction>
+    </xs:simpleType>
+
+    <xs:complexType name="countType">
```
bstansberry: This isn't used.

threads/src/main/resources/schema/wildfly-threads_2_0.xsd
```diff
@@ -0,0 +1,310 @@
...
+            <xs:documentation>
+                <![CDATA[
+                A size designation.
+            ]]>
+            </xs:documentation>
+        </xs:annotation>
+        <xs:attribute name="count" type="xs:int" use="required"/>
+    </xs:complexType>
+
+    <xs:complexType name="ref">
```
bstansberry: This isn't used.

ctomc:
thank you!

PR #849, opened 2015-06-30
Errors on CLI server reload reconnection

Gonthim:
https://issues.jboss.org/browse/WFCORE-783

Also fixed a problem with the timeout argument not overriding the default or file config timeout value.

### DIFF
--- a/cli/src/main/java/org/jboss/as/cli/handlers/ReloadHandler.java
+++ b/cli/src/main/java/org/jboss/as/cli/handlers/ReloadHandler.java
@@ -160,20 +160,7 @@ public class ReloadHandler extends BaseOperationCommand {
             }
         }
 
-        // if I try to reconnect immediately, it'll hang for 5 sec
-        // which the default connection timeout for model controller client
-        // waiting half a sec on my machine works perfectly
-        try {
-            Thread.sleep(500);
-        } catch (InterruptedException e) {
-            throw new CommandLineException("Interrupted while pausing before reconnecting.", e);
-        }
-        try {
-            cliClient.ensureConnected(ctx.getConfig().getConnectionTimeout() + 1000);
-        } catch(CommandLineException e) {
-            ctx.disconnectController();
-            throw e;
-        }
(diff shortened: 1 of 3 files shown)


bstansberry:
WildFly Core failures look related.

Please ping Alexey to have a look once the failures are sorted. This looks ok to me but he's more familiar with the existing remote reconnect logic and may notice something I don't.

cli/src/main/java/org/jboss/as/cli/handlers/ReloadHandler.java
```diff
@@ -220,17 +211,8 @@ private void doHandleEmbedded(CommandContext ctx, ModelControllerClient client)
                     ctx.disconnectController();
```
bstansberry: A nit, but the comment at L203 above here is slightly out of date now. The catch of IOException is no longer just to comply with the compiler now that the call may go remote.

Gonthim:
The server can reload into 'running' or 'restart-required' states. The persistent restart-required test wasn't running against the embedded server reload so it wasn't caught before.

cli/src/main/java/org/jboss/as/cli/handlers/ReloadHandler.java
```diff
@@ -190,6 +177,10 @@ private void doHandleEmbedded(CommandContext ctx, ModelControllerClient client)
             throw new CommandLineException("Failed to execute :reload", e);
         }
 
+        ensureServerRebootComplete(ctx, client);
+    }
+
+    private void ensureServerRebootComplete(CommandContext ctx, ModelControllerClient client) throws CommandLineException {
```
aloubyansky: Note, the same "making sure it's up" issue affects ShutdownHandler when it's handling a restart.

aloubyansky:
Looks ok to me.

Gonthim:
Retest this please

Gonthim:
ConnectionFactoryManagementTestCase failure seems unrelated. Can replicate locally with and without this commit. Unknown cause.

PR #852, opened 2015-07-01
Host controller subsystem socket bindings

kabir:
https://issues.jboss.org/browse/WFCORE-777  - Add ability to use socket bindings from host subsystems. This also introduces a HostCapabilityContext for wiring the socket-binding capabilities and requirements for the bindings and the host subsystems.

https://issues.jboss.org/browse/WFCORE-786 (clean up)


controller/src/main/java/org/jboss/as/controller/OperationContextImpl.java
```diff
@@ -2524,4 +2540,38 @@ public int hashCode() {
         }
     }
 
+    private static class HostCapabilityContext implements CapabilityContext {
+        @Override
+        public boolean canSatisfyRequirements(CapabilityContext dependentContext) {
+            return true;
```
bstansberry: This should be:

```
return dependentContext == CapabilityContext.GLOBAL || dependentContext instanceof HostCapabilityContext;
```

Otherwise capabilities from domain level profile and socket-binding-group resources will be able to find requirements from the host context.

PR #853, opened 2015-07-01
[WFCORE-787] Discard any management model changes before beginning ro…

bstansberry:
…llback

See http://lists.jboss.org/pipermail/wildfly-dev/2015-July/004170.html

After James Perkins reported WFCORE-788, which looks to be a flaw with the WFCORE-787 fix, I debugged how operations were executing and realized the support in AbstractOperationContext  for the no-longer used recursive execution mode was making the logic far too error prone. So I decided to clean that up, and filed WFCORE-792. The second commit in this PR is for that.

Finally, the 3rd commit fixes WFCORE-788 itself by fixing a problem in the CapabilityRegistry copy-on-write behavior where the copy wasn't deep enough, allowing state mutation to leak out of the OperationContext making modifications even if the copy is never published.


bstansberry:
Hold while I sort WFCORE-788.

PR #854, opened 2015-07-01
WFCORE-219 Implement transformers for graceful shutdown operations

ctomc:


core-model-test/tests/src/test/java/org/jboss/as/core/model/test/servergroup/DomainServerGroupTransformersTestCase.java
```diff
@@ -112,19 +111,43 @@ public void testRejectTransformers71x() throws Exception {
                 .createContentRepositoryContent("09876543210987654321");
 
         // Add legacy subsystems
-        LegacyKernelServicesInitializer legacyInitializer =
-                StandardServerGroupInitializers.addServerGroupInitializers(builder.createLegacyKernelServicesBuilder(modelVersion, testControllerVersion));
-        ExcludeCommonOperations.excludeBadOps_7_1_x(legacyInitializer);
+        StandardServerGroupInitializers.addServerGroupInitializers(builder.createLegacyKernelServicesBuilder(modelVersion, testControllerVersion));
         KernelServices mainServices = builder.build();
 
-        List<ModelNode> ops = builder.parseXmlResource("servergroup_1_3-with-expressions.xml");
```
kabir: This file can be removed

core-model-test/tests/src/test/java/org/jboss/as/core/model/test/servergroup/DomainServerGroupTransformersTestCase.java
```diff
@@ -77,18 +80,14 @@ public DomainServerGroupTransformersTestCase(TransformersTestParameter params) {
     @Test
     public void testServerGroupsTransformer() throws Exception {
 
-        boolean is71x = modelVersion.getMajor() == 1 && modelVersion.getMinor() < 4;
         KernelServicesBuilder builder = createKernelServicesBuilder(TestModelType.DOMAIN)
                 .setModelInitializer(StandardServerGroupInitializers.XML_MODEL_INITIALIZER, StandardServerGroupInitializers.XML_MODEL_WRITE_SANITIZER)
                 .createContentRepositoryContent("12345678901234567890")
                 .createContentRepositoryContent("09876543210987654321")
-                .setXmlResource(is71x ? "servergroup_1_3.xml" : "servergroup-with-expressions.xml");
```
kabir: servergroup_1_3.xml can be removed

host-controller/src/main/java/org/jboss/as/domain/controller/operations/DomainServerLifecycleHandlers.java
```diff
@@ -391,4 +396,27 @@ public void execute(OperationContext context, ModelNode operation) throws Operat
         }
     }
 
+    public static void registerServerLifeCycleOperationsTransformers(ResourceTransformationDescriptionBuilder builder) {
+        builder.addOperationTransformationOverride(SUSPEND_SERVERS).setReject().end()
+                .addOperationTransformationOverride(RESUME_SERVERS).setReject().end()
+                .addOperationTransformationOverride(STOP_SERVERS)
+                .setCustomOperationTransformer((context, address, operation) -> {
+                    if (operation.hasDefined(ModelDescriptionConstants.TIMEOUT)) { //timeout is new parameter, reject if present, otherwise it is fine
```
kabir: You can use the builder for this as well

    .addRejectCheck() 

It works similar to for attributes, but called on the OperationTransformationOverrideBuilder applies to operation parameters.

kabir:
Apart from the nit-picks this looks ok to me, however I’d like Brian to give the ok on this. It could be that I am making problems out of nothing, but Brian needs to decide that.

One issue I see is if you for example have a domain with everything the same version and suspend the servers it will work fine. But while suspended, if you then add a legacy HC, you will not be able to resume the servers again. 

It might be worth checking if the legacy hosts have any servers running for the targetted server groups, although I don’t know off the top of my head if this will actually work in practise or not. If it is possible, I added some functionality to be able to propagate information from the OperationStepHandlers to the transformers via attachments. e.g. https://github.com/wildfly/wildfly-core/blob/master/subsystem-test/tests/src/test/java/org/jboss/as/subsystem/test/transformers/subsystem/VersionedExtensionCommon.java#L153 gets propagated to the transformers and are used at https://github.com/wildfly/wildfly-core/blob/master/subsystem-test/tests/src/test/java/org/jboss/as/subsystem/test/transformers/subsystem/VersionedExtension2.java#L68.

host-controller/src/main/java/org/jboss/as/domain/controller/operations/DomainServerLifecycleHandlers.java
ctomc: totally forgot about it :) yeah that is much better.

core-model-test/tests/src/test/java/org/jboss/as/core/model/test/servergroup/DomainServerGroupTransformersTestCase.java
ctomc: removed

core-model-test/tests/src/test/java/org/jboss/as/core/model/test/servergroup/DomainServerGroupTransformersTestCase.java
ctomc: removed

bstansberry:
Re @kabir 's comment above, the "resume" op probably shouldn't be rejected, it should be ignored. If the legacy slave was not able to suspend a server, then nothing is suspended and the "resume" can be interpreted as having worked.

By "ignored" I mean the DC either doesn't send it to the slave at all or if that's not possible the DC suppresses the failure that comes back. I forget which of those we have facilities to handle. :( Hopefully one or the other!

kabir:
ignored == discard

ctomc:
@bstansberry @kabir it makes sense to discard resume, but same thing could be said for suspend.
both are essentially no-op for servers that don't support it.

PR #858, opened 2015-07-03
[WFCORE-795] Copy attribute group and allowed values in the copy attr…

kabir:
…ibute builder part

https://issues.jboss.org/browse/WFCORE-795


kabir:
Retest this please

PR #859, opened 2015-07-03
Wfcore 747 master

soul2zimate:
https://issues.jboss.org/browse/WFCORE-747

Add a marker file to standalone or domain server tmp directory to record server startup state once server is started. It will be removed when server is stopped.
Inside wildfly-init-redhat.sh and wildfly-init-debian.sh, it will read this temporary marker file to exam server startup result before timeout instead of checking volatile log message from console log(key log id changed, log handler removed, etc.)

### DIFF
--- a/core-feature-pack/src/main/resources/content/bin/init.d/wildfly-init-debian.sh
+++ b/core-feature-pack/src/main/resources/content/bin/init.d/wildfly-init-debian.sh
@@ -100,6 +100,7 @@ if [ "$JBOSS_MODE" = "standalone" ]; then
 	if [ -z "$JBOSS_CONFIG" ]; then
 		JBOSS_CONFIG=standalone.xml
 	fi
+	JBOSS_MARKERFILE="$JBOSS_HOME/standalone/tmp/wildfly-startup-marker"
 else
 	JBOSS_SCRIPT="$JBOSS_HOME/bin/domain.sh"
 	if [ -z "$JBOSS_DOMAIN_CONFIG" ]; then
@@ -108,6 +109,7 @@ else
 	if [ -z "$JBOSS_HOST_CONFIG" ]; then
 		JBOSS_HOST_CONFIG=host.xml
 	fi
+	JBOSS_MARKERFILE="$JBOSS_HOME/domain/tmp/wildfly-startup-marker"
 fi
 
 # Check startup file
@@ -162,6 +164,7 @@ case "$1" in
 		mkdir -p $(dirname "$JBOSS_CONSOLE_LOG")
 		chown $JBOSS_USER $(dirname "$JBOSS_PIDFILE") || true
 		cat /dev/null > "$JBOSS_CONSOLE_LOG"
+		currenttime=$(date +%s%N | cut -b1-13)
 
 		if [ "$JBOSS_MODE" = "standalone" ]; then
 			start-stop-daemon --start --user "$JBOSS_USER" \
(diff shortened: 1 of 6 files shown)


jamezp:
We should discuss this on the wildfly-dev mailing list. I'm not sure if we want to go here or not.

core-feature-pack/src/main/resources/content/bin/init.d/wildfly-init-debian.sh
```diff
@@ -108,6 +109,7 @@ else
 	if [ -z "$JBOSS_HOST_CONFIG" ]; then
 		JBOSS_HOST_CONFIG=host.xml
 	fi
+	JBOSS_MARKERFILE="$JBOSS_HOME/domain/tmp/jboss-as-startup-marker"
```
jamezp: All the new lines have tabs and we're a no tabs code base :)

core-feature-pack/src/main/resources/content/bin/init.d/wildfly-init-redhat.sh
```diff
@@ -62,6 +62,7 @@ if [ "$JBOSS_MODE" = "standalone" ]; then
 	if [ -z "$JBOSS_CONFIG" ]; then
 		JBOSS_CONFIG=standalone.xml
 	fi
+	JBOSS_MARKERFILE=$JBOSS_HOME/standalone/tmp/jboss-as-startup-marker
```
jamezp: All the new lines have tabs and we're a no tabs code base :)

host-controller/src/main/java/org/jboss/as/host/controller/HostControllerService.java
```diff
@@ -173,6 +176,21 @@ public void stop(StopContext context) {
         String prettyVersion = environment.getProductConfig().getPrettyVersionString();
         processState.setStopping();
         ServerLogger.AS_ROOT_LOGGER.serverStopped(prettyVersion, Integer.valueOf((int) (context.getElapsedTime() / 1000000L)));
+        deleteStartupMarker();
+    }
+
+    private void deleteStartupMarker() {
+        String tmpDir = System.getProperty("jboss.domain.temp.dir");
```
bstansberry: This is available from the HostControllerEnvironment (the 'environment' field.)

server/src/main/java/org/jboss/as/server/ApplicationServerService.java
```diff
@@ -200,6 +203,21 @@ public synchronized void stop(final StopContext context) {
         CurrentServiceContainer.setServiceContainer(null);
         String prettyVersion = configuration.getServerEnvironment().getProductConfig().getPrettyVersionString();
         ServerLogger.AS_ROOT_LOGGER.serverStopped(prettyVersion, Integer.valueOf((int) (context.getElapsedTime() / 1000000L)));
+        deleteStartupMarker();
+    }
+
+    private void deleteStartupMarker() {
+        String tmpDir = System.getProperty("jboss.server.temp.dir");
```
bstansberry: This is available from the ServerEnvironment, which is available from the 'configuration' field.

server/src/main/java/org/jboss/as/server/BootstrapListener.java
```diff
@@ -92,11 +101,36 @@ protected void done(final long bootstrapTime, final StabilityStatistics statisti
...
             ServerLogger.AS_ROOT_LOGGER.startedClean(prettyVersion, bootstrapTime, started, active + passive + onDemand + never + lazy, onDemand + passive + lazy);
+            createStartupMarker("success", startTime);
         } else {
             ServerLogger.AS_ROOT_LOGGER.startedWitErrors(prettyVersion, bootstrapTime, started, active + passive + onDemand + never + lazy, failed + problem, onDemand + passive + lazy);
+            createStartupMarker("error", startTime);
         }
     }
 
+    private void createStartupMarker(String result, long startTime) {
+        String tmpDir = isStandalone ? System.getProperty("jboss.server.temp.dir") : System.getProperty("jboss.domain.temp.dir");
```
bstansberry: No system properties. See my comments above about where to find the location of these dirs.

bstansberry:
Please post on the wildfly-dev thread about this design, so anyone interested is aware.

core-feature-pack/src/main/resources/content/bin/init.d/wildfly-init-debian.sh
soul2zimate: These two script files use tabs in new lines, do you want me to reformat them all?

n1hility:
I think this is fine, but i think a better approach would be use tail -f to wait for the write, so that you don't have to sleep loop. That can be a subsequent enhancement.

soul2zimate:
rebased to resolve conflict, retest this please.

PR #863, opened 2015-07-07
[WFCORE-797] Add smoke test for cloning profiles in CLI

marekkopecky:
Provide profile cloning ability (runtime (CLI)) to create new profiles based on existing JBoss profiles.
https://issues.jboss.org/browse/WFCORE-797

This test is similar as CloneProfileTestCase from WildFly (https://issues.jboss.org/browse/WFLY-4838). WildFly has more complex default profile in domain part of TS than WildFly Core. But there should be a variant of this in WildFly Core too (https://github.com/wildfly/wildfly/pull/7699#issuecomment-118884646).


marekkopecky:
retest this please

PR #865, opened 2015-07-08
WFLY-4885 Include the original cause why loading of console's module failed

jaikiran:
The commit here changes the error logging to include the original cause which prevented the console's module from being loaded. This will help debug issues like the one reported in https://issues.jboss.org/browse/WFLY-4885


ctomc:
we don't want this. as if console module is missing that is completely valid scenario. better to log warn instead of error for case like this

jaikiran:
> we don't want this. as if console module is missing that is completely valid scenario.

Hi Tomaz, you are right. Doesn't make sense to log the stacktrace when the absence of a console is a valid case.

Closing this PR.

PR #866, opened 2015-07-08
[WFCORE-774] Memory leak in JBoss CLI

TomasHofman:
https://issues.jboss.org/browse/WFCORE-774

### DIFF
--- a/cli/src/main/java/org/jboss/as/cli/impl/CommandContextImpl.java
+++ b/cli/src/main/java/org/jboss/as/cli/impl/CommandContextImpl.java
@@ -264,6 +264,9 @@ class CommandContextImpl implements CommandContext, ModelControllerClientFactory
     private InputStream stdIn = Settings.getInstance().getInputStream();
     private boolean uninstallIO;
 
+    private static JaasConfigurationWrapper jaasConfigurationWrapper; // we want this wrapper to be only created once
+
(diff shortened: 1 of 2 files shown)


aloubyansky:
There is a method in CommandContext called terminateSession() which is where the clean up should take place when the session is over. Is there anything that could also be added to that method?

TomasHofman:
Well, the only problem there was that every instance of CommandContextImpl added new JaasConfigurationWrapper to global Configuration (so it was creating longer and longer chain). JaasConfigurationWrappers are also holding references back to CommandContextImpl instances, which prevents those from being garbage collected.

So I see two options:
1) only add JaasConfigurationWrapper to global Configuration in the first CommandContext instance being created (shouldn't affect the functionality as I believe),
2) or remove JaasConfigurationWrapper from the Configuration in terminateSession() and restore original configuration.

Neither option is bullet proof, and I thought 1) is simpler.

aloubyansky:
@darranl what's your opinion on this?

aloubyansky:
BTW, Thomas, thanks for your effort here!

darranl:
Looks fine.

aloubyansky:
Ok, given that no matter which option we choose it's still a static config at the end, I'm fine with this too.

PR #868, opened 2015-07-08
WFLY-4885 Log warn instead of error in case console module is not present

ctomc:


domain-http/interface/src/main/java/org/jboss/as/domain/http/server/logging/HttpServerLogger.java
```diff
@@ -51,11 +52,11 @@
...
     void uploadError(@Cause Throwable cause);
 
-    @LogMessage(level = ERROR)
+    @LogMessage(level = WARN)
     @Message(id = 3, value = "Unable to load console module for slot %s, disabling console")
     void consoleModuleNotFound(String slot);
 
-    @LogMessage(level = ERROR)
+    @LogMessage(level = WARN)
     @Message(id = 4, value = "Unable to load error context for slot %s, disabling error context.")
```
bstansberry: The org.jboss.as.domain-http-error-context is part of the core feature pack and its use isn't just for console use cases. Not having it present is more of an ERROR condition.

ctomc:
fixed

PR #875, opened 2015-07-10
[WFCORE-755] correctly parse --timeout on embed-server

baranowb:
https://issues.jboss.org/browse/WFCORE-755


bstansberry:
L232 is checking nanos:

```
                } while (System.nanoTime() < expired);
```

So L149 and L239 need to be fixed to convert to/from nanos, not millis.

baranowb:
Im not sure why it needs nano resolution, but its up.

PR #877, opened 2015-07-10
[WFCORE-810] ProtocolConfigRemoveHandler should check for handler res…

bstansberry:
…ource existence

I didn't add a test for this, as an existing test (AuditLogToTCPSyslogTestCase) fails in tearDown because of this in my WFCORE-808 branch. This fixes the problem.


bstansberry:
retest this please

PR #881, opened 2015-07-13
Trigger validation of includes also when a profile/s-g-b is empty

kabir:
https://issues.jboss.org/browse/WFCORE-815 / https://issues.jboss.org/browse/JBEAP-401

https://issues.jboss.org/browse/WFCORE-816

### DIFF
--- a/controller/src/main/java/org/jboss/as/controller/logging/ControllerLogger.java
+++ b/controller/src/main/java/org/jboss/as/controller/logging/ControllerLogger.java
@@ -36,6 +36,7 @@ import java.util.Collection;
 import java.util.List;
 import java.util.Set;
 import java.util.concurrent.CancellationException;
+
 import javax.xml.namespace.QName;
 import javax.xml.stream.Location;
 import javax.xml.stream.XMLStreamException;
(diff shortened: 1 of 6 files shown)


controller/src/main/java/org/jboss/as/controller/logging/ControllerLogger.java
```diff
@@ -3332,35 +3333,17 @@ XMLStreamException discoveryOptionsMustBeDeclared(String adminOnlyCmd, String po
     @Message(id = 398, value = "An attempt was made to rename the resource found at %s to %s. However, '%s' is one of the resource types defined to be ordered on the parent resource %s")
     OperationFailedRuntimeException orderedChildTypeRenamed(PathAddress read, PathAddress transformed, String type, Set<String> parentOrderedChildren);
 
-    @Message(id = 399, value = "Model references of type '%s' are missing: %s")
-    OperationFailedException missingReferences(String type, Set<String> missing);
```
bstansberry: Did we ship any of these ids in WildFly 9? If so, they can't be repurposed and should just be commented out if no longer used.

bstansberry:
Just an FYI, I expect a lot of the validation is going to end up being done via capability resolution.

controller/src/main/java/org/jboss/as/controller/logging/ControllerLogger.java
kabir: In 1.0.0.Final (i.e. WF 9) the ids end at 395, so we should be good.
https://github.com/wildfly/wildfly-core/blob/1.0.0.Final/controller/src/main/java/org/jboss/as/controller/logging/ControllerLogger.java

(Not that I actually checked beforehand :-) )

PR #884, opened 2015-07-13
WFCORE-804 Display attribute definition's capabilty references

ctomc:
I refactored the way CapabilityRecorder is set in builder to have more generic code.
I am also tempted to remove get/setCapabiltyRecorder() as this is impl detail more than anything else.

### DIFF
--- a/controller/src/main/java/org/jboss/as/controller/AttributeDefinition.java
+++ b/controller/src/main/java/org/jboss/as/controller/AttributeDefinition.java
@@ -84,6 +84,7 @@ public abstract class AttributeDefinition {
     private final Boolean nilSignificant;
     private final AttributeParser parser;
     private final String attributeGroup;
+    protected final CapabilityReferenceRecorder referenceRecorder;
 
     // NOTE: Standards for creating a constructor variant are:
     // 1) Expected to be a common use case; no one-offs.
@@ -99,7 +100,7 @@ public abstract class AttributeDefinition {
                 toCopy.isValidateNull(), toCopy.getAlternatives(), toCopy.getRequires(), toCopy.getAttributeMarshaller(),
                 toCopy.isResourceOnly(), toCopy.getDeprecated(),
                 wrapConstraints(toCopy.getAccessConstraints()), toCopy.getNullSignificant(), toCopy.getParser(),
-                toCopy.getAttributeGroup(), toCopy.getAllowedValues(), wrapFlags(toCopy.getFlags()));
+                toCopy.getAttributeGroup(), toCopy.referenceRecorder, toCopy.getAllowedValues(), wrapFlags(toCopy.getFlags()));
     }
 
     protected AttributeDefinition(String name, String xmlName, final ModelNode defaultValue, final ModelType type,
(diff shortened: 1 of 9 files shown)


bstansberry:
+1 to getters. This information needs to be always available, regardless of the CRR impl.

controller/src/main/java/org/jboss/as/controller/CapabilityReferenceRecorder.java
```diff
@@ -55,9 +59,9 @@
      */
     public static class DefaultCapabilityReferenceRecorder implements CapabilityReferenceRecorder {
 
-        private final String baseRequirementName;
-        private final String baseDependentName;
-        private final boolean dynamicDependent;
+        protected final String baseRequirementName;
+        protected final String baseDependentName;
+        protected final boolean dynamicDependent;
```
bstansberry: These can be private now.

controller/src/main/java/org/jboss/as/controller/CapabilityReferenceRecorder.java
```diff
@@ -47,6 +47,10 @@
      */
     void removeCapabilityRequirements(OperationContext context, String attributeName, String... attributeValues);
 
+    String getBaseDependentName();
+
+    String getBaseRequirementName();
```
bstansberry: These need javadoc.

controller/src/main/java/org/jboss/as/controller/CapabilityReferenceRecorder.java
```diff
@@ -47,6 +47,10 @@
...
+
```
bstansberry: You might as well expose isDynamicDependent() as well. I believe Elytron branch has a custom impl of this interface, which this will break, so we might as well get everything in and only break them once. ;)

controller/src/main/java/org/jboss/as/controller/AttributeDefinition.java
```diff
@@ -812,6 +815,10 @@ public ModelNode getNoTextDescription(boolean forOperation) {
                 result.get(ModelDescriptionConstants.REQUIRES).add(required);
             }
         }
+        if (referenceRecorder != null) {
+            result.get(ModelDescriptionConstants.CAPABILITY_REFERENCE).set(referenceRecorder.getBaseRequirementName());
```
bstansberry: Please ping @heiko-braun as to whether he'd like to see all 3 pieces of data here: requirement name, dependent name, whether dependent is dynamic. All that info might be overkill, but OTOH once we ship with a data structure we are stuck with it. My instinct is to report more.

controller/src/main/java/org/jboss/as/controller/CapabilityReferenceRecorder.java
ctomc: I will check with elytron guys to see how their custom impl looks like and if we can abolish the CapabilityReferenceRecorder interface and hide all its stuff internally

controller/src/main/java/org/jboss/as/controller/AttributeDefinition.java
ctomc: I don't think we need anything more here.
On resource that itself exposes capability there should be all that extra info.
In this case, dependent name is name of the attribute on which this information is present.
Info about dynamic capability i am not sure if that helps in any way. 
Given that later on when there will be need for resolving that capability that will be done by capability registry and not directly.

controller/src/main/java/org/jboss/as/controller/CapabilityReferenceRecorder.java
ctomc: Elytron guys don't use custom impl. So we have bit more free hands here.

heiko-braun:
With the backwards compatibility requirement on mind, I would suggest we reduce the attribute info the absolute minimum for now, unitl I had a chance to actually swork through more sceanrios that make use of the new meta data. 

From my point of view it would be sufficient to currently expose just the referenced capability.

PR #885, opened 2015-07-13
[WFCORE-818] OperationContext shouldn't require a registered capabili…

bstansberry:
…ty to provide a ServiceName

I'm inclined to remove the Class serviceValueType params from this API altogether but if I do I'll use a different JIRA. The now purely optional type checking isn't worth the API bloat IMO.


bstansberry:
retest this please

PR #886, opened 2015-07-14
[WFCORE-820] Attribute for-hosts for Security Realm with Kerberos doe…

TomasHofman:
…s not select keytab in defined order

https://issues.jboss.org/browse/WFCORE-820
https://bugzilla.redhat.com/show_bug.cgi?id=1228228

### DIFF
--- a/domain-management/src/test/java/org/jboss/as/domain/management/security/KeytabIdentityFactoryServiceTestCase.java
+++ b/domain-management/src/test/java/org/jboss/as/domain/management/security/KeytabIdentityFactoryServiceTestCase.java
@@ -1,0 +1,174 @@
+/*
+ * JBoss, Home of Professional Open Source.
+ * Copyright 2012, Red Hat Middleware LLC, and individual contributors
+ * as indicated by the @author tags. See the copyright.txt file in the
+ * distribution for a full listing of individual contributors.
+ *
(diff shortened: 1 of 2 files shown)


TomasHofman:
@darranl Would you review please?

domain-management/src/test/java/org/jboss/as/domain/management/security/KeytabIdentityFactoryServiceTestCase.java
```diff
@@ -0,0 +1,174 @@
+/*
+ * JBoss, Home of Professional Open Source.
+ * Copyright 2012, Red Hat Middleware LLC, and individual contributors
```
darranl: It is 2015

darranl:
The actual code changes look fine - however IMO when it comes to backwards compatibility unless the current behaviour is so broken it is impossible for it to actually be used by anyone the current behaviour has to take priority over documentation.

In this case I think it will be better to just update the documentation.

Should the issue described here actually occur users do have the opportunity to add an additional for-hosts value to restore the desired behaviour.

bstansberry:
Given the comments above I'm closing this.

TomasHofman:
I opened new PR with modified documentation: https://github.com/wildfly/wildfly-core/pull/887

PR #887, opened 2015-07-15
[WFCORE-820] Attribute for-hosts for Security Realm with Kerberos doe…

TomasHofman:
…s not select keytab in defined order

https://bugzilla.redhat.com/show_bug.cgi?id=1228228
https://issues.jboss.org/browse/WFCORE-820

### DIFF
--- a/domain-management/src/test/java/org/jboss/as/domain/management/security/KeytabIdentityFactoryServiceTestCase.java
+++ b/domain-management/src/test/java/org/jboss/as/domain/management/security/KeytabIdentityFactoryServiceTestCase.java
@@ -1,0 +1,174 @@
+/*
+ * JBoss, Home of Professional Open Source.
+ * Copyright 2015, Red Hat Middleware LLC, and individual contributors
+ * as indicated by the @author tags. See the copyright.txt file in the
+ * distribution for a full listing of individual contributors.
+ *
+ * This is free software; you can redistribute it and/or modify it
+ * under the terms of the GNU Lesser General Public License as
+ * published by the Free Software Foundation; either version 2.1 of
+ * the License, or (at your option) any later version.
+ *
+ * This software is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the GNU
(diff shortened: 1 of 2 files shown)


TomasHofman:
Retest this please.

server/src/main/resources/schema/jboss-as-config_1_7.xsd
```diff
@@ -461,8 +461,8 @@
                             
                             At the time authentication is going to be handled the keytab will be selected as follows: -
                              1 - Iterate the list of keytabs and identity one where the for-hosts attribute contains an entry matching protocol/hostname.
-                             2 - Iterate the list of keytabs and identity one where the for-hosts attribute contains an entry matching hostname.
-                             3 - Iterate the list of keytabs and identify one where the name of the principal matches matches protocol/hostname.
+                             2 - Iterate the list of keytabs and identify one where the name of the principal matches matches protocol/hostname.
+                             3 - Iterate the list of keytabs and identity one where the for-hosts attribute contains an entry matching hostname.
                              4 - Iterate the list of keytabs and identify one where the hostname portion of the principal matches the hostname of the request.
                              5 - Use the keytab where for-hosts is set to '*'.
```
bstansberry: The current schema is https://github.com/wildfly/wildfly-core/blob/master/server/src/main/resources/schema/wildfly-config_4_0.xsd

TomasHofman:
Ah, thanks Brian. Didn't realize there may be newer version. So probably only fix the latest one...?

Can I actually change XSD documentation just like that? Should I involve documentation team for some reason?

bstansberry:
Hi Tomas,

Sorry, I didn't check back to see your comment.

Generally we don't update schemas once they are published in a final release. It's confusing to have different variations of the same schema out there on the internet. So just the latest is ok.

If you need to fix BZ 1228228 (currently it's not triaged for inclusion in a CP and I doubt it ever will be) and someone insists on updating the 1.7 schema as the fix, we won't object to changing it in WildFly Core as well.

Re: docs, probably the simplest thing to do is change the "Doc Type" on the BZ to Known Issue and fill in the Doc Text. I believe that will provide the info needed to create an entry in the EAP Release Notes if docs wants one.

bstansberry:
retest this please

PR #889, opened 2015-07-16
[WFCORE-822] Use a test specific name for the configuration.

darranl:
Also wrap the DirectoryService to enable shutdown of the CacheManager.


jamezp:
@darranl there seems to be a test compilation issue.

asoldano:
Can we also make sure the testsuite works with a bit more recent ehcache versions (at least 2.7.2) ?

darranl:
That would most likely be out of our control as the tests are depending on the version Apache DS is depending on.

PR #890, opened 2015-07-16
WFCORE-671 : Configuration change reporting.

ehsavoie:
Adding new collector
Creating a resource for configuration-changes configuration.
Securing operations :
        * if the user can't access the resource addressed by the operation then the operation information are not displayed but the entry point is kept (aka date, result).
        * if the user can access the address but doesn't have the rights to run the operation then the parameters are not accessible in the operation description.

 Providing an in-memory audit-logger to track configuration changes easily.

Jira: https://issues.jboss.org/browse/WFCORE-671


ehsavoie:
Retest this please

bstansberry:
retest this please

domain-management/src/main/java/org/jboss/as/domain/management/ConfigurationChangeResourceDefinition.java
```diff
@@ -0,0 +1,241 @@
...
+        private static final MaxHistoryWriteHandler INSTANCE = new MaxHistoryWriteHandler(ConfigurationChangesCollector.INSTANCE);
+        private final ConfigurationChangesCollector collector;
+
+        private MaxHistoryWriteHandler(ConfigurationChangesCollector collector) {
+            this.collector = collector;
+        }
+
+        @Override
+        public void execute(OperationContext context, ModelNode operation) throws OperationFailedException {
+            ModelNode maxHistory = MAX_HISTORY.resolveModelAttribute(context, operation);
```
bstansberry: This isn't correct for a write-attribute op handler, which has params (name=max-history,value=15), not (max-history=15).

core-model-test/tests/src/test/resources/org/jboss/as/core/model/test/auditlog/standalone.xml
```diff
@@ -40,6 +41,7 @@
                <syslog-handler name="syslog-tls-default" formatter="json-two" >
                   <tls/>
                </syslog-handler>
+               <configuration-changes-handler name="in-memory" max-history="10"/>
```
bstansberry: I think this should just be "in-memory-handler". The semantics should be the same as any other handler, except for the max-history behavior. It should record whatever the other handlers record, not limited to configuration changes, not disabled at boot, etc.

ehsavoie: Should it use a formatter ?

domain-management/src/main/java/org/jboss/as/domain/management/audit/InMemoryAuditLogHandlerResourceDefinition.java
```diff
@@ -0,0 +1,236 @@
...
+            SimpleAttributeDefinitionBuilder.create(REMOTE_ADDRESS, ModelType.STRING, false).setStorageRuntime().build(),
+            SimpleAttributeDefinitionBuilder.create(OUTCOME, ModelType.STRING, false).setStorageRuntime().build(),
+            SimpleListAttributeDefinition.Builder.of(OPERATIONS,
+                    SimpleAttributeDefinitionBuilder.create(OP, ModelType.STRING, false).setStorageRuntime().build())
+            .build())
+            .build();
+
+    public static final String OPERATION_NAME = "show-logs";
+    protected static final AttributeDefinition[] ATTRIBUTES = new AttributeDefinition[]{MAX_OPERATION_COUNT};
+    private static final Map<String, InMemoryAuditLogHander> histories = Collections.synchronizedMap(new HashMap<String, InMemoryAuditLogHander>());
```
bstansberry: I don't want to delay merging this PR for this, but once it's merged please send up a PR to make this a non-static field and have the InMemoryAuditLogHandlerResourceDefinition pass a ref to it to the handlers when it constructs them.

Having this be static is no big deal in this case, but I don't want anyone looking at kernel code and thinking that using static fields to store state is a good practice.

PR #891, opened 2015-07-16
WFCORE-809 Can not parse an ObjectListAttributeDefinition of a Persis…

ctomc:
…tentResourceDefinition


ctomc:
something is wrong with this PR, code passes testsuite fine locally. Will try opening new PR from same branch.

PR #899, opened 2015-07-17
Register notifications in model parse test utils so we spam the logs …

bstansberry:
…less


bstansberry:
This doesn't appear to help.

PR #900, opened 2015-07-17
WFCORE-775 domain mode - server fails to start intermittently

luck3y:
WFCORE-775
Forgot to send-pr.

This changes the authKey generation from a 16 byte array to a base 64 encoded string. The amount  of information present remains the same, the string will be 24 bytes in length, however.

java.utils.Base64 is used, JDK8 only, backports would need to utilize an alternative encoder.

### DIFF
--- a/process-controller/src/main/java/org/jboss/as/process/ManagedProcess.java
+++ b/process-controller/src/main/java/org/jboss/as/process/ManagedProcess.java
@@ -34,6 +34,7 @@ import java.io.InputStreamReader;
 import java.io.OutputStream;
 import java.io.OutputStreamWriter;
 import java.io.PrintStream;
+import java.nio.charset.Charset;
 import java.util.ArrayList;
 import java.util.List;
 import java.util.Map;
@@ -178,7 +179,7 @@ final class ManagedProcess {
             StreamUtils.writeUTFZBytes(base64, hostName);
             StreamUtils.writeInt(base64, port);
             StreamUtils.writeBoolean(base64, managementSubsystemEndpoint);
-            base64.write(asAuthKey.getBytes());
+            base64.write(asAuthKey.getBytes(Charset.forName("US-ASCII")));
             base64.close(); // not flush(). close() writes extra data to the stream allowing Base64 input stream
                             // to distinguish end of message
         } catch (IOException e) {
@@ -229,7 +230,7 @@ final class ManagedProcess {
         try {
             // WFLY-2697 All writing is in Base64
             OutputStream base64 = getBase64OutputStream(stdin);
-            base64.write(authKey.getBytes());
+            base64.write(authKey.getBytes(Charset.forName("US-ASCII")));
             base64.close(); // not flush(). close() writes extra data to the stream allowing Base64 input stream
                             // to distinguish end of message
             ok = true;
--- a/process-controller/src/main/java/org/jboss/as/process/ManagedProcess.java
+++ b/process-controller/src/main/java/org/jboss/as/process/ManagedProcess.java
@@ -61,7 +61,7 @@ final class ManagedProcess {
     private final Object lock;
 
     private final ProcessController processController;
-    private final byte[] authKey;
+    private final String authKey;
     private final boolean isPrivileged;
     private final RespawnPolicy respawnPolicy;
 
@@ -72,7 +72,7 @@ final class ManagedProcess {
     private boolean stopRequested = false;
     private final AtomicInteger respawnCount = new AtomicInteger(0);
 
-    public byte[] getAuthKey() {
+    public String getAuthKey() {
         return authKey;
     }
 
@@ -95,7 +95,7 @@ final class ManagedProcess {
         ;
     }
 
-    ManagedProcess(final String processName, final List<String> command, final Map<String, String> env, final String workingDirectory, final Object lock, final ProcessController controller, final byte[] authKey, final boolean privileged, final boolean respawn) {
+    ManagedProcess(final String processName, final List<String> command, final Map<String, String> env, final String workingDirectory, final Object lock, final ProcessController controller, final String authKey, final boolean privileged, final boolean respawn) {
         if (processName == null) {
             throw ProcessLogger.ROOT_LOGGER.nullVar("processName");
         }
@@ -117,7 +117,7 @@ final class ManagedProcess {
         if (authKey == null) {
             throw ProcessLogger.ROOT_LOGGER.nullVar("authKey");
         }
-        if (authKey.length != 16) {
+        if (authKey.length() != ProcessController.AUTH_BYTES_ENCODED_LENGTH) {
             throw ProcessLogger.ROOT_LOGGER.invalidLength("authKey");
         }
         this.processName = processName;
@@ -169,7 +169,7 @@ final class ManagedProcess {
         }
     }
 
-    public void reconnect(String scheme, String hostName, int port, boolean managementSubsystemEndpoint, byte[] asAuthKey) {
+    public void reconnect(String scheme, String hostName, int port, boolean managementSubsystemEndpoint, String asAuthKey) {
         assert holdsLock(lock); // Call under lock
         try {
             // WFLY-2697 All writing is in Base64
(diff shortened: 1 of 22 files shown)


process-controller/src/main/java/org/jboss/as/process/ProcessController.java
```diff
@@ -115,7 +117,7 @@ public void addProcess(final String processName, final byte[] authKey, final Lis
             }
             final ManagedProcess process = new ManagedProcess(processName, command, env, workingDirectory, lock, this, authKey, isPrivileged, respawn);
             processes.put(processName, process);
-            processesByKey.put(new Key(authKey), process);
+            processesByKey.put(new Key(authKey.getBytes()), process);
```
n1hility: This needs 'US-ASCII' if it is going to be encoded like this.

process-controller/src/main/java/org/jboss/as/process/ProcessControllerClient.java
```diff
@@ -110,11 +110,11 @@ public void handleMessage(final Connection connection, final InputStream dataStr
...
                         for (int i = 0; i < cnt; i++) {
                             final String processName = readUTFZBytes(dataStream);
-                            final byte[] processAuthCode = new byte[16];
-                            readFully(dataStream, processAuthCode);
+                            final byte[] processAuthBytes = new byte[ProcessController.AUTH_BYTES_ENCODED_LENGTH];
+                            readFully(dataStream, processAuthBytes);
                             final boolean processRunning = StreamUtils.readBoolean(dataStream);
                             final boolean processStopping = StreamUtils.readBoolean(dataStream);
-                            inventory.put(processName, new ProcessInfo(processName, processAuthCode, processRunning, processStopping));
+                            inventory.put(processName, new ProcessInfo(processName, new String(processAuthBytes), processRunning, processStopping));
```
n1hility: Also needs a charset if its decoded like this.

process-controller/src/main/java/org/jboss/as/process/ProcessControllerClient.java
```diff
@@ -304,7 +305,7 @@ public void reconnectProcess(final String processName, final URI managementURI,
             writeUTFZBytes(os, managementURI.getHost());
             writeInt(os, managementURI.getPort());
             writeBoolean(os, managementSubsystemEndpoint);
-            os.write(authKey);
+            os.write(authKey.getBytes());
```
n1hility: same as earlier comment

process-controller/src/main/java/org/jboss/as/process/ProcessController.java
luck3y: Ah, good catch thanks!

process-controller/src/main/java/org/jboss/as/process/ManagedProcess.java
```diff
@@ -178,7 +178,7 @@ public void reconnect(String scheme, String hostName, int port, boolean manageme
             StreamUtils.writeUTFZBytes(base64, hostName);
             StreamUtils.writeInt(base64, port);
             StreamUtils.writeBoolean(base64, managementSubsystemEndpoint);
-            base64.write(asAuthKey);
+            base64.write(asAuthKey.getBytes());
```
n1hility: needs US-ASCII charset, or some other simple ascii encoding conversion

process-controller/src/main/java/org/jboss/as/process/ManagedProcess.java
```diff
@@ -229,7 +229,7 @@ void doStart(boolean restart) {
         try {
             // WFLY-2697 All writing is in Base64
             OutputStream base64 = getBase64OutputStream(stdin);
-            base64.write(authKey);
+            base64.write(authKey.getBytes());
```
n1hility: needs US-ASCII charset, or some other simple ascii encoding conversion

process-controller/src/main/java/org/jboss/as/process/ProcessController.java
```diff
@@ -183,7 +185,7 @@ public void removeProcess(final String processName) {
                 return;
             }
             boolean removed = processes.remove(processName) != null;
-            processesByKey.remove(new Key(process.getAuthKey()));
+            processesByKey.remove(new Key(process.getAuthKey().getBytes()));
```
n1hility: needs US-ASCII charset, or some other simple ascii encoding conversion

process-controller/src/main/java/org/jboss/as/process/ProcessController.java
```diff
@@ -339,7 +341,7 @@ void sendInventory() {
                         StreamUtils.writeInt(os, processCollection.size());
                         for (ManagedProcess process : processCollection) {
                             StreamUtils.writeUTFZBytes(os, process.getProcessName());
-                            os.write(process.getAuthKey());
+                            os.write(process.getAuthKey().getBytes());
```
n1hility: needs US-ASCII charset, or some other simple ascii encoding conversion

process-controller/src/main/java/org/jboss/as/process/ProcessControllerClient.java
```diff
@@ -211,14 +211,15 @@ public void addProcess(String processName, byte[] authKey, String[] cmd, String
...
+        // this is Base64 encoded, and padded up to 24 bytes
+        if (authKey.length() != ProcessController.AUTH_BYTES_ENCODED_LENGTH) {
             throw ProcessLogger.ROOT_LOGGER.invalidAuthKeyLen();
         }
         final OutputStream os = connection.writeMessage();
         try {
             os.write(Protocol.ADD_PROCESS);
             writeUTFZBytes(os, processName);
-            os.write(authKey);
+            os.write(authKey.getBytes());
```
n1hility: needs US-ASCII charset, or some other simple ascii encoding conversion

n1hility:
haha ill stop spamming you, you get the point

luck3y:
Yah, fix incoming, thanks

bstansberry:
I don't think the change to process-controller is needed. I'm talking about the byte[] that's sent by the PC to a process via stdin and then gets passed back in ProcessControllerClient calls. Those bytes weren't getting converted to a string on either side, so they were not part of the WFCORE-775 problem. Now they are getting converted to String, hence Jason's many comments. AIUI the WFCORE-775 problem was limited to the HC<->server comm channel, which doesn't involve the process controller. 

I don't particularly object to the PC change, but if I'm right about the above, any EAP fix can't include it, as it's inappropriate for a CP. (Changes something not broken.) So Ken you'd have to do any backport, as I don't want someone else trying to separate the PC part from the HC<->server comm part.

luck3y:
Thanks for the feedback Brian. I agree with your comments about the necessity of the PC changes, those were made for consistency more than anything else. I can back port just the fix to EAP quickly if i omit these, so I think that should work ok.

bstansberry:
Ok, sounds good.

process-controller/src/main/java/org/jboss/as/process/ProcessControllerClient.java
dmlloyd: I believe IDEA has an inspection for this that you can enable.  Also for locales too, which can cause similar issues.

luck3y:
@All : Updated with corrections for encoding / decoding to US-ASCII. Any more thoughts / comments?

n1hility:
In regards to the PC changes @bstansberry was referring to ,I was wondering about the double base64 encoding, since the I/O stream is itself already encoded. However, I think it would end up being six of one and half dozen of the other because you would have to encode/decode around every auth in a server start.

luck3y:
Yeah, I guess it could be completely redundant there. I'll double check.

luck3y:
Ah, so thats the only value ever written to that Base64OutPutStream, I'd mistakenly assumed that there might have been other data too. 

I'm not sure this will make much of a difference here, since we're never looking to actually use the decoded value, we want the base 64 encoded string as they authKey after we generate it.

bstansberry:
Re: "Ah, so thats the only value ever written to that Base64OutPutStream, I'd mistakenly assumed that there might have been other data too."  I'm not sure if you meant that specific call, but if not Base64OutPutStream is also used by ProcessControllerClient.sendStdIn and ManagedProcess.sendStdIn to send the serialized ServerStartTask from the HC to the PC and on to the server.

Anyway, this looks fine.

luck3y:
Yeah sorry I noticed that afterwards right above where I was looking, forgot to edit my comment. Thanks for merge :)

PR #901, opened 2015-07-17
WFCORE-760: Validate that the "category" attribute for a logger is no…

brunolitman:
…t an empty string. 
Without this fix, such a config fails to start and throws a StringOutOfBoundsException. Now instead, it fails with an invalidAttributeValue error which generates a more useful error message as follows:

javax.xml.stream.XMLStreamException: ParseError at [row,col]:[1,5167]
Message: WFLYCTL0106: Invalid value '' for attribute 'category'

To validate the error manually, i simply added a logger with an empty category to the src/test/resources/logging.xml file and ran the tests, which throws the above error, as expected. A test cannot be written as we cannot add a logger dynamically with an empty category, this can only happen in a config file.

### DIFF
--- a/logging/src/main/java/org/jboss/as/logging/LoggingSubsystemParser_3_0.java
+++ b/logging/src/main/java/org/jboss/as/logging/LoggingSubsystemParser_3_0.java
@@ -85,6 +85,7 @@ import java.util.Set;
 import org.jboss.as.controller.PathAddress;
 import org.jboss.as.controller.operations.common.Util;
 import org.jboss.as.controller.parsing.ParseUtils;
+import static org.jboss.as.controller.parsing.ParseUtils.invalidAttributeValue;
 import org.jboss.dmr.ModelNode;
 import org.jboss.staxmapper.XMLElementReader;
 import org.jboss.staxmapper.XMLExtendedStreamReader;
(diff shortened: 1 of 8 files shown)


brunolitman:
https://issues.jboss.org/browse/WFCORE-760

jamezp:
this is okay to test

brunolitman:
@luck3y thank you for the feedback! I have removed the else blocks. As for null checking, I don't believe it is possible to get to that point with a null value. Briefly discussed with @jamezp in hipchat and he recommended we leave the null check out, as an NPE here would denote a valid bug.

luck3y:
@brunolitman Yep, saw that in hipchat :) Looks good to me.

brunolitman:
Just for the record, more discussion led to `if (value == null || value.trim().isEmpty())` to remove blank spaces as well, which are not allowed by the log manager either.

logging/src/main/java/org/jboss/as/logging/LoggingSubsystemParser_3_0.java
```diff
@@ -209,6 +210,9 @@ private static void parseLoggerElement(final XMLExtendedStreamReader reader, fin
             required.remove(attribute);
             switch (attribute) {
                 case CATEGORY: {
+                    if("".equals(value)) {
```
jamezp: Missed one for the empty check :)

brunolitman: Noticed that just now. Fixed - sorry

PR #906, opened 2015-07-19
https://issues.jboss.org/browse/WFCORE-830 support multiple patch str…

aloubyansky:
…eams for the same installation

split installation manager and installed identity

multiple identities per installation manager

added patch-stream runtime resource - a child of core-service=patching, added stream name arg to the tool API and PatchingException, added optional --patch-stream arg to the CLI patch command, made patch-info OSH delegate to read-resource in case patch-id was not specified


patching/src/main/java/org/jboss/as/patching/management/LocalPatchRollbackHandler.java
```diff
@@ -34,38 +34,38 @@
...
+    protected void execute(final OperationContext context, final ModelNode operation, final InstallationManager installationManager, final String patchStream) throws OperationFailedException {
 
         if (installationManager.requiresRestart()) {
             throw PatchLogger.ROOT_LOGGER.serverRequiresRestart();
         }
 
+        final ModelNode patchIdNode = PatchResourceDefinition.PATCH_ID.resolveModelAttribute(context, operation);
+        if(!patchIdNode.isDefined()) {
+            throw ControllerLogger.ROOT_LOGGER.nullNotAllowed(PatchResourceDefinition.PATCH_ID.getName());
+        }
```
bstansberry: This doesn't seem right. If undefined is not allowed the AttributeDefinition should be configured that way and L52 will throw the exception, removing the need for this if block. But PatchResourceDefinition.PATCH_ID has allowNull set to 'true' which means the operation description will incorrectly say undefined is ok.

Perhaps the same AttributeDefinition is being used for two purposes?

patching/src/main/java/org/jboss/as/patching/management/PatchResource.java
```diff
@@ -109,22 +117,57 @@ public Resource clone() {
...
+        public Set<String> children() {
+            final InstallationManager manager = getInstallationManager();
+            if (manager == null) {
+                return Collections.emptySet();
+            }
+            List<InstalledIdentity> installedIdentities;
+            try {
+                installedIdentities = manager.getInstalledIdentities();
+            } catch (PatchingException e) {
+                e.printStackTrace(); // TODO
```
bstansberry: :)

patching/src/main/java/org/jboss/as/patching/management/PatchResourceDefinition.java
```diff
@@ -173,7 +173,59 @@ private PatchResourceDefinition() {
...
+
+        final StandardResourceDescriptionResolver resolver = new StandardResourceDescriptionResolver("patching.patch-stream", RESOURCE_NAME, PatchResourceDefinition.class.getClassLoader());
+        registry.registerSubModel(new SimpleResourceDefinition(PathElement.pathElement("patch-stream"), resolver) {
+
+            @Override
+            public void registerAttributes(final ManagementResourceRegistration resource) {
+                super.registerAttributes(resource);
+                registerPatchingAttributes(resource);
+                // HACK -- workaround WFCORE-17
+                resource.setRuntimeOnly(true);
```
bstansberry: WFCORE-17 is resolved now.

patching/src/main/java/org/jboss/as/patching/management/PatchResourceDefinition.java
```diff
@@ -173,7 +173,59 @@ private PatchResourceDefinition() {
...
+            }
+
+            @Override
+            public List<AccessConstraintDefinition> getAccessConstraints() {
+                return sensitivity;
+            }
+        });
+
+        // HACK -- workaround WFCORE-17
+        registry.setRuntimeOnly(true);
```
bstansberry: WFCORE-17 is resolved now.

patching/src/main/java/org/jboss/as/patching/tool/PatchOperationTarget.java
```diff
@@ -244,58 +292,122 @@ public RemotePatchOperationTarget(PathAddress address, ModelControllerClient cli
...
             }
-            return client.execute(operation);
+            return executeOp(operation);
+        }
+
+        protected ModelNode executeOp(ModelNode operation) throws PatchingException {
+            try {
+                return client.execute(operation);
+            } catch (IOException e) {
+                throw new PatchingException("Failed to execute operation " + operation, e);
```
bstansberry: Do we care about i18n in the tool? I suppose it's governed by the same rules as the CLI.

patching/src/main/java/org/jboss/as/patching/management/LocalAgeoutHistoryHandler.java
```diff
@@ -68,20 +66,31 @@ public boolean accept(File dir, String name) {
...
+            if (patchStream != null) {
+                final InstalledIdentity installedIdentity = instMgr.getInstalledIdentity(patchStream, null);
+                ageOutHistory(installedIdentity);
+            } else {
+                for (InstalledIdentity installedIdentity : instMgr.getInstalledIdentities()) {
+                    ageOutHistory(installedIdentity);
+                }
+            }
+        } catch (PatchingException e) {
+            throw new OperationFailedException(PatchLogger.ROOT_LOGGER.failedToLoadIdentity(), e);
```
bstansberry: Is this failure due to a client mistake? If not, a RuntimeException of some sort should be thrown, as OFE represents a client mistake and is only logged in the server.log at DEBUG. Also, at some point we'll use exceptions to trigger proper HTTP response codes (versus current 500) in the http interface, and an OFE would result in a 4xx.

Don't worry about this for this patch; this doesn't look any different from other places in the patching code. My question is intended to trigger a general review at some point of throwing OFE in the patching resources.

bstansberry:
Small comments aside, it looks ok.

patching/src/main/java/org/jboss/as/patching/management/LocalPatchRollbackHandler.java
aloubyansky: Right. I missed the description part of it.

patching/src/main/java/org/jboss/as/patching/tool/PatchOperationTarget.java
aloubyansky: Yes, it affects only the CLI which is why I left it hardcoded.

patching/src/main/java/org/jboss/as/patching/management/LocalAgeoutHistoryHandler.java
aloubyansky: That won't be a client mistake.

bstansberry:
Interesting. I've seen that set of the same 4 failures on at least a couple other PR runs in the last day or so.

bstansberry:
The 4 failures were due to a problem on one of the agents; zombie process I believe.

bstansberry:
retest this please

PR #907, opened 2015-07-20
[WFCORE-654] : Configuration files get default permissions when written.

ehsavoie:
Using Nio2 API to try to create the temporary configuration file with the same attributes as the original configuration file.

Jira: https://issues.jboss.org/browse/WFCORE-654


bstansberry:
retest this please

PR #908, opened 2015-07-20
Improve parser to support properly persiting / parsing of multi element attributes

ctomc:
WFCORE-809 Can not parse an ObjectListAttributeDefinition of a PersistentResourceDefinition
WFCORE-691 introduce general parsing support for attributes that marshall as elements
WFCORE-691 Can not parse a wrapped properties attribute inside a group

needs https://github.com/wildfly/wildfly/pull/7790 to be merged first


ctomc:
^^ fixed by https://github.com/wildfly/wildfly/pull/7817

PR #909, opened 2015-07-21
WFCORE-805 Add capabilty information to ResourceDefinition / MRR

ctomc:


controller/src/main/java/org/jboss/as/controller/registry/ImmutableManagementResourceRegistration.java
```diff
@@ -252,4 +253,11 @@
      * @return the set of ordered child types, and and empty set if there are none
      */
     Set<String> getOrderedChildTypes();
+
+    /**
+     * Returns capability definition for this resource.
+     *
+     * @return RuntimeCapability if any registered otherwise null
```
bstansberry: This will not return null but may return an empty set.

controller/src/main/java/org/jboss/as/controller/registry/ImmutableManagementResourceRegistration.java
```diff
@@ -252,4 +253,11 @@
...
+     */
+    Set<RuntimeCapability<?>> getCapabilities();
```
bstansberry: I'm a bit uncomfortable about exposing the actual RuntimeCapability instances here. A capability provider's RuntimeCapability instances are really not meant to leak outside that provider's own code and the kernel.

WDYT about using AbstractCapability? That includes the metadata type stuff without including any of the service discovery and runtime API discovery stuff that shouldn't leak out. That kind of use-case distinction is the (vague) reason why I used the superclass.

bstansberry:
Other than the 2 comments, looks good.

controller/src/main/java/org/jboss/as/controller/registry/ImmutableManagementResourceRegistration.java
bstansberry: BTW, I'm fine with renaming AbstractCapability too. ManagedCapability? Just Capability? I'm normally not a fan of using 'AbstractXXX' as a return value or param type in widely used APIs.

ctomc: I think having just Capabilty would be fine, could even be new interface just to properly abstract this.

ctomc: Looking at OperationContext#registerCapability it also accepts RuntimeCapability as parameter, should we change that as well?

ctomc:
fixed the PR based on the comments.
I went with introducing interface Capability that AbstractCapability implements

controller/src/main/java/org/jboss/as/controller/ResourceDefinition.java
```diff
@@ -81,6 +81,15 @@
     void registerChildren(final ManagementResourceRegistration resourceRegistration);
 
     /**
+     * Register capabilities associated with this resource.
+     *
+     * @param resourceRegistration a {@link ManagementResourceRegistration} created from this definition
+     */
+    default void registerCapabilities(final ManagementResourceRegistration resourceRegistration){
```
ctomc: just heads up, this is default method just so wildfly full code wont get angry.

controller/src/main/java/org/jboss/as/controller/registry/ImmutableManagementResourceRegistration.java
bstansberry: No, that's not a problem as that's not a leak of the instance to someone other than the provider or the kernel. It passes the instance from provider to kernel. What I wanted to avoid was other subsystems seeing an API call to get a ref to someone else's RuntimeCapability and deciding it was appropriate to use that.

bstansberry:
Assuming the tests come out ok, this looks good.

ctomc:
tests should be fine, the previous two full had 4 failed tests as result of handed processes on agent. (agent has been rebooted since)

bstansberry:
Ah, ok that explains why I saw the same 4 failures on a couple other runs. Thanks.

PR #915, opened 2015-07-24
https://issues.jboss.org/browse/WFCORE-838 Decrease Patch Size Per CP

aloubyansky:
An improvement of the patch bundle concept currently used for CPs where a bundle contains a number of CP patches from which only those relevant to the target installation will be applied.
The structure and format of the patched data, checksum verification and all the rest of the mechanism of application the changes stays the same.
This commit introduces support for multiple patch.xml files - one per identity version the patch can be applied to and update that identity to the version of the latest CP included in the patch.

Thanks,
Alexey

### DIFF
--- a/patching/src/main/java/org/jboss/as/patching/Constants.java
+++ b/patching/src/main/java/org/jboss/as/patching/Constants.java
@@ -40,6 +40,7 @@ public class Constants {
     public static final String NAME = "name";
     public static final String VERBOSE = "verbose";
     public static final String VERSION = "version";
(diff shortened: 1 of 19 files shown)


aloubyansky:
Just wanted to add, there are hardcoded exception messages in PatchMerger and PatchToolImpl, these won't be exposed through the management interface, they will only appear in the CLI (that doesn't have any localization) and the patchgen tool (that is a separate internal tool at this point).

aloubyansky:
I'll re-open after fixing the failure.

PR #917, opened 2015-07-24
Capability resolution does not properly deal with 'included' profiles and socket-binding-groups

bstansberry:
https://issues.jboss.org/browse/WFCORE-834

Summary of the design here:

When a capability is registered, the registration is associated with a scope, the "CapabilityContext".[1] A given capability name can only exist once within a given scope, but can exist in different scopes. An example scope is a profile within a domain.

(CapabiltyContext is a bad name. I'll rename it to CapabilityScope. https://issues.jboss.org/browse/WFCORE-842)

A big part of capability resolution is understanding the various scopes and how they can relate to each other. For example, a capability in a profile scope can find required capabilities in a socket-binding-group scope, but not vice versa.

From WFCORE-834 and other domain resolution cases (e.g. WFCORE-833), it's clear that the existing resolution algorithm is inadequate. This PR modifies it in two basic ways:

1) The CapabilityRegistry, when performing a resolution asks the CapabilityContext instances whether they are capable of resolving a particular capability requirement from a different CapabilityContext. The context implementation can then use whatever logic it wants to make that decision. The rules about what context can work with what other context is encapsulated in the context impl itself.

2) For each resolution (i.e. at the end of Stage.MODEL for any op that modifies the CapabilityRegistry) a CapabilityResolutionContext object is created with a ref to the current management model. This is passed to each CapabilityContext when it is asked if it can provide the required capability. The CapabilityContext can attach things to the CapabilityResolutionContext as to share data across calls.

The WFCORE-834 solution then builds on this by having the CapabilityContext impls for profiles and socket-binding-groups use the CapabilityResolutionContext to inspect the current management model and determine what scopes include other scopes. This data is calculated once per resolution run and attached to the CapabilityResolutionContext so it doesn't need recalculation for that run. The result is a CapabilityContext for profile "a" understands that it can provide capabilities for the profile "b" context if "b" includes "a".


controller/src/main/java/org/jboss/as/controller/OperationContextImpl.java
```diff
@@ -300,14 +310,23 @@ private boolean validateCapabilities() {
...
                             }
-                            set.add(required);
+                        } else {
+                            if (unexplainedProblem == null) {
+                                unexplainedProblem = new StringBuilder(ControllerLogger.ROOT_LOGGER.requiredCapabilityMissing());
+                            }
+                            String formattedCapability = ignoreContext
+                                    ? ControllerLogger.ROOT_LOGGER.formattedCapabilityName(reqReq.getRequiredName())
+                                    : ControllerLogger.ROOT_LOGGER.formattedCapabilityId(reqReq.getRequiredName(), reqReq.getDependentContext().getName());
+                            unexplainedProblem = unexplainedProblem.append('\n').append(formattedCapability);
```
jamezp: We should probably use `append(System.lineSeparator())` so Windows will get a nice looking message as well.

controller/src/main/java/org/jboss/as/controller/OperationContextImpl.java
```diff
@@ -347,33 +367,50 @@ private boolean validateCapabilities() {
...
+                            }
                         }
-                        if (isBooting()) {
-                            ControllerLogger.ROOT_LOGGER.inconsistentCapabilityContexts(reg.getDependentName(),
-                                    depConName, step.address.toCLIStyleString(),reg.getRequiredName(), depConName);
+                    } else {
+                        if (unexplainedProblem == null) {
+                            unexplainedProblem = new StringBuilder();
+                        } else {
+                            unexplainedProblem.append('\n');
```
jamezp: Same note here with the `System.lineSeparator()`.

jamezp:
I'm going to mark this as ready for merge and we can worry about the message formatting later.

ctomc:
@jamezp agreed, that msg formatting is not important in this context. 
As part of model references this message will be improved/changed a lot. 
Than we should be more careful about formatting

PR #918, opened 2015-07-28
[WFCORE-844] Add smoke test for hierarchical composition in CLI

marekkopecky:
Test for logical hierarchical composition of Profies and Socket Binding Groups.
https://issues.jboss.org/browse/WFCORE-844

This test serves as basic smoke tests to be run within the testsuite.

### DIFF
--- a/testsuite/domain/src/test/java/org/jboss/as/test/integration/domain/management/cli/HierarchicalCompositionTestCase.java
+++ b/testsuite/domain/src/test/java/org/jboss/as/test/integration/domain/management/cli/HierarchicalCompositionTestCase.java
@@ -121,7 +121,7 @@ public class HierarchicalCompositionTestCase extends AbstractCliTestBase {
         try {
             output = (String) cli.readAllAsOpResult().getResult();
         } catch (IOException e) {
-            Assert.fail("Fail to get result of CLI command: " + line);
+            Assert.fail("Fail to get result of CLI command: " + line + "\n" + e.getMessage() + "\n" + e.getStackTrace());
         }
         return output;
     }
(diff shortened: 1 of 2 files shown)


testsuite/domain/src/test/java/org/jboss/as/test/integration/domain/management/cli/HierarchicalCompositionTestCase.java
```diff
@@ -0,0 +1,236 @@
...
+     * @return CLI output
+     */
+    private String cliGetResult(String line) {
+        log.info(line);
+        cli.sendLine(line);
+        String output = null;
+        try {
+            output = (String) cli.readAllAsOpResult().getResult();
+        } catch (IOException e) {
+            Assert.fail("Fail to get result of CLI command: " + line);
```
jamezp: We should add the stack trace to the failure message. It could be important to the cause of the failure.

marekkopecky:
retest this please

testsuite/domain/src/test/java/org/jboss/as/test/integration/domain/management/cli/HierarchicalCompositionTestCase.java
```diff
@@ -0,0 +1,236 @@
...
+     * @return CLI output
+     */
+    private String cliGetResult(String line) {
+        log.info(line);
+        cli.sendLine(line);
+        String output = null;
+        try {
+            output = (String) cli.readAllAsOpResult().getResult();
+        } catch (IOException e) {
+            Assert.fail("Fail to get result of CLI command: " + line + "\n" + e.getMessage() + "\n" + e.getStackTrace());
```
jamezp: One small note here, but it doesn't need to be changed immediately. Instead of using `\n` we should use `System.lineSeparator()` so errors will look nice for Windows users as well. It's not a huge deal in a test really, just something to keep consistent on for good practice :)

Another option would be to use a string format like:
```java
String.format("Fail to get result of CLI command: %s%n%s%n%s", line, e.getMessage(), e.getStackTrace())
```

jamezp:
@Gonthim Does this look okay to you?

PR #920, opened 2015-07-29
WFLY-2422 Introduction of remote-outbound-connection-group concept

tadamski:
Connection group is a tool that is suppose to simplify configuration in cases in which many connections share the same properties. Connections group references will be configured in remoting profile. They are also planned to be used in cluster configuration.

Forum references:
Problems with current implementation:
http://lists.jboss.org/pipermail/wildfly-dev/2015-March/003674.html
Proposed solution:
http://lists.jboss.org/pipermail/wildfly-dev/2015-July/004235.html

### DIFF
--- a/remoting/subsystem/src/main/java/org/jboss/as/remoting/Namespace.java
+++ b/remoting/subsystem/src/main/java/org/jboss/as/remoting/Namespace.java
@@ -39,12 +39,13 @@ public enum Namespace {
     REMOTING_1_2("urn:jboss:domain:remoting:1.2"),
     REMOTING_2_0("urn:jboss:domain:remoting:2.0"),
     REMOTING_3_0("urn:jboss:domain:remoting:3.0"),
+    REMOTING_3_1("urn:jboss:domain:remoting:3.1")
     ;
 
     /**
      * The current namespace version.
      */
-    public static final Namespace CURRENT = REMOTING_3_0;
+    public static final Namespace CURRENT = REMOTING_3_1;
 
     private final String name;
 
--- a/remoting/tests/src/test/resources/org/jboss/as/remoting/remoting-with-outbound-connection-group.xml
+++ b/remoting/tests/src/test/resources/org/jboss/as/remoting/remoting-with-outbound-connection-group.xml
@@ -1,0 +1,12 @@
+<subsystem xmlns="urn:jboss:domain:remoting:3.1">
+    <connector name="test-connector" socket-binding="test"/>
+    <outbound-connections>
+        <remote-outbound-connection-group name="group" outbound-socket-binding-refs="dummy-outbound-socket, other-outbound-socket">
+            <properties>
+                <property name="org.xnio.Options.SASL_POLICY_NOANONYMOUS" value="false"/>
+                <property name="org.xnio.Options.SSL_ENABLED" value="false"/>
+            </properties>
+        </remote-outbound-connection-group>
+    </outbound-connections>
+</subsystem>
+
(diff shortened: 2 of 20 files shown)


jamezp:
We should create a [WFCORE JIRA](https://issues.jboss.org/browse/WFCORE) for this.

tadamski:
OK. Will create a child jira of WFLY-2422.

kabir:
@tadamski I have not looked at all, beyond seeing that no new transformers are added. Does it need some?

Also, is this something which should be in product?

tadamski:
@kabir transformers are not needed. This PR is an attempt to provide tool that will enable simpler configuration in common use cases as requested in WFLY-2422 but all configurations valid before it's introduction are valid and don't have to be changed. I also believe that it shouldn't be a part of product yet - IMO it has to be tested in community edition.

rhusar:
Is this going to get to the product? We would be happy to use it in mod_cluster and infinispan subsystem for proxies and hot rod, respectively.

Also, if and how is this transformable?

remoting/subsystem/src/main/java/org/jboss/as/remoting/Namespace.java
```diff
@@ -39,12 +39,13 @@
...
     REMOTING_2_0("urn:jboss:domain:remoting:2.0"),
     REMOTING_3_0("urn:jboss:domain:remoting:3.0"),
+    REMOTING_3_1("urn:jboss:domain:remoting:3.1")
     ;
 
     /**
      * The current namespace version.
      */
-    public static final Namespace CURRENT = REMOTING_3_0;
+    public static final Namespace CURRENT = REMOTING_3_1;
```
rhusar: Why 3_1? Sounds like this should be 4_0 to align with other subsystems that have been bumped with every release..

remoting/tests/src/test/resources/org/jboss/as/remoting/remoting-with-outbound-connection-group.xml
```diff
@@ -0,0 +1,12 @@
+<subsystem xmlns="urn:jboss:domain:remoting:3.1">
+    <connector name="test-connector" socket-binding="test"/>
+    <outbound-connections>
+        <remote-outbound-connection-group name="group" outbound-socket-binding-refs="dummy-outbound-socket, other-outbound-socket">
```
rhusar: This is defined in xsd as xs:list which is *whitespace* delimited, so why the commas...

http://www.w3schools.com/schema/el_list.asp

remoting/subsystem/src/main/java/org/jboss/as/remoting/Namespace.java
bstansberry: Yes. Schema version bumps should jump the same degree as the release that contains them. So major bumps in majors.

This ensures that any branches created from those release tags can add new schema versions without concern about conflicting with a schema that master or the branch from a later release has added.

bstansberry:
@n1hility we need to decide if this is acceptable in WF 10.

I haven't looked at it.

tadamski:
@bstansberry I will fix the schema issues. There would be also changes in widfly repository which are part of the same functionality. I'm currently implementing them. Hopefully will be able to create a wildfly PR within a week.

remoting/subsystem/src/main/java/org/jboss/as/remoting/Namespace.java
rhusar: @bstansberry @tadamski ...and same for the model version.

bstansberry:
There is a requirement for transformation, since a new resource type, remote-outbound-connection-group, is added. The transformer basically ensures the user gets a comprehensible error message if they try to use this in a profile used by a legacy slave. Without it the slave will fail producing the same message it produces when it gets a bogus operation with a bad address, which is a confusing message.

tadamski:
OK. Will add such transformer then.

tadamski:
The solution requires further discussion about changes on wildfly side regarding integration with new ejb-client discovery mechanism and as a result of that there may be further modifications on wildfly-core side. Because of that, I'm closing this PR now. Will return to it after PTO.

PR #921, opened 2015-07-30
https://issues.jboss.org/browse/WFCORE-838 Decrease Patch Size Per CP

aloubyansky:
An improvement of the patch bundle concept currently used for CPs where a bundle contains a number of CP patches from which only those relevant to the target installation will be applied.
The structure and format of the patched data, checksum verification and all the rest of the mechanism of application the changes stays the same.
This commit introduces support for multiple patch.xml files - one per identity version the patch can be applied to and update that identity to the version of the latest CP included in the patch.

Thanks,
Alexey

### DIFF
--- a/patching/src/main/java/org/jboss/as/patching/metadata/PatchMerger.java
+++ b/patching/src/main/java/org/jboss/as/patching/metadata/PatchMerger.java
@@ -1,0 +1,512 @@
+/*
+ * JBoss, Home of Professional Open Source.
+ * Copyright 2015, Red Hat, Inc., and individual contributors
+ * as indicated by the @author tags. See the copyright.txt file in the
+ * distribution for a full listing of individual contributors.
+ *
+ * This is free software; you can redistribute it and/or modify it
+ * under the terms of the GNU Lesser General Public License as
+ * published by the Free Software Foundation; either version 2.1 of
+ * the License, or (at your option) any later version.
+ *
+ * This software is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the GNU
+ * Lesser General Public License for more details.
+ *
+ * You should have received a copy of the GNU Lesser General Public
+ * License along with this software; if not, write to the Free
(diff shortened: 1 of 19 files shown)


patching/src/main/java/org/jboss/as/patching/metadata/PatchMerger.java
```diff
@@ -0,0 +1,512 @@
...
+import org.jboss.as.patching.metadata.Patch.PatchType;
+import org.jboss.as.patching.runner.PatchUtils;
+
+/**
+ *
+ * @author Alexey Loubyansky
+ */
+public class PatchMerger {
+
+    private static final String DIRECTORY_PREFIX = "wildfly-patch-";
```
jamezp: There's no fix needed for this PR, but should this use some kind of configurable property so it will work with EAP as well?

patching/src/main/java/org/jboss/as/patching/metadata/PatchMerger.java
```diff
@@ -0,0 +1,512 @@
...
+        private Integer getKey(ContentItem item) {
+            final int prime = 31;
+            int result = 1;
+            result = prime * result + item.getContentType().hashCode();
+            result = prime * result + item.getRelativePath().hashCode();
+            return result;
+        }
+    }
+
+    private static void ls(final File f) {
```
jamezp: Both the `ls` methods don't appear to be used.

patching/src/main/java/org/jboss/as/patching/metadata/PatchMerger.java
```diff
@@ -0,0 +1,512 @@
...
+
+/**
+ *
+ * @author Alexey Loubyansky
+ */
+public class PatchMerger {
+
+    private static final String DIRECTORY_PREFIX = "wildfly-patch-";
+    public static final String PATCH_XML_SUFFIX = "-patch.xml";
+    private static final File TEMP_DIR = new File(SecurityActions.getSystemProperty("java.io.tmpdir"));
```
jamezp: Any reason not to just use `WildFlySecurityManager.getPropertyPrivileged("java.io.tmpdir", null)` and remove the `SecurityActions` class?

If this doesn't need to be back ported to Java 6 you could use `Files.createTempDirectory()` where this static instance is used.

PR #924, opened 2015-07-31
[WFCORE-839] : Don't store state in InMemoryAuditLogHandlerResourceDefinition.

ehsavoie:
Updating ManagedAuditLogger to be able to access the last log entries with only a real implemntation in the InMemoryAuditLogHander.

Jira: https://issues.jboss.org/browse/WFCORE-839

### DIFF
--- a/controller/src/main/java/org/jboss/as/controller/audit/InMemoryAuditLogHander.java
+++ b/controller/src/main/java/org/jboss/as/controller/audit/InMemoryAuditLogHander.java
@@ -61,33 +61,29 @@ public final class InMemoryAuditLogHander extends AuditLogHandler {
     private int maxHistory;
     private final AuditLogItemFormatter myFormatter = new InMemoryFormatter();
 
+
     public InMemoryAuditLogHander(String name, int maxHistory) {
         super(name, IN_MEMORY_FORMATTER_NAME, maxHistory);
-        items = new ArrayList<>(maxHistory);
+        this.items = new ArrayList<>(maxHistory);
         this.maxHistory = maxHistory;
         setFormatter(myFormatter);
     }
 
-    public List<ModelNode> getItems() {
+    @Override
+    public List<ModelNode> listLastEntries() {
         return Collections.unmodifiableList(items);
     }
 
     public void setMaxHistory(int maxHistory) {
-        synchronized (items) {
-            this.maxHistory = maxHistory;
-            while (maxHistory < items.size()) {
-                items.remove(0);
-            }
+        this.maxHistory = maxHistory;
+        while (maxHistory < items.size()) {
+            items.remove(0);
         }
     }
 
     @Override
     void setFormatter(AuditLogItemFormatter formatter) {
-        if (formatter != null) {
-            super.setFormatter(formatter);
-        } else {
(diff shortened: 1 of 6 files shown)


controller/src/main/java/org/jboss/as/controller/audit/InMemoryAuditLogHander.java
```diff
@@ -58,36 +58,34 @@
     }
     private static final String IN_MEMORY_FORMATTER_NAME = "in-memory-formatter";
     private final List<ModelNode> items;
-    private int maxHistory;
+    private volatile int maxHistory;
```
jamezp: The rest of this class isn't thread safe. I don't think `volatile` is needed here.

controller/src/main/java/org/jboss/as/controller/audit/ManagedAuditLogger.java
```diff
@@ -182,6 +184,14 @@
     JsonAuditLogItemFormatter getJsonFormatter(String name);
 
     /**
+     * Gets the last log entries
+     *
+     * @param name the name of the handler
+     * @return the last log entries of the handler
+     */
+    List<ModelNode> listLastEntries(String name);
```
jamezp: Why not use a default method here?
```java
default List<ModelNode> listLastEntries(String name) {
    return Collections.emptyList();
}
```

This would allow you not to have to define `Collections.emptyList()` everywhere else.

ehsavoie: because I'm not used to using java 8 features ;)

jamezp: Cool it's not a huge deal, but just better for compatibility. Even though in this case it won't really matter.

bstansberry:
We're all these changes necessary for WFCORE-839, or was there some other problem that had to be fixed? This is a lot bigger than I expected; I thought it would be trivial.

The handler doesn't seem thread safe now.

ehsavoie:
The handler isn't thread safe as all handlers. Thread safety is managed by ManagedAuditLoggerImpl

bstansberry:
Thanks; I'll have another look now that I get the idea. :)

controller/src/main/java/org/jboss/as/controller/audit/InMemoryAuditLogHander.java
```diff
@@ -61,33 +61,31 @@
...
-                items.remove(0);
-            }
+    @Override
+    public void setMaxFailureCount(int maxHistory) {
+        super.setMaxFailureCount(maxHistory);
+        this.maxHistory = maxHistory;
+        while (maxHistory < items.size()) {
+            items.remove(0);
         }
     }
```
bstansberry: Looking at ManagedAuditLogger, it already has handler-specific modification methods, e.g. updateSyslogHandlerAppName, updateSyslogHandlerFacility, updateSyslogHandlerReconnectTimeout. That's a bit yuck, but since they are already there I don't see why another one can't be added for setting the max history.

ManagedAuditLogger isn't usable outside the kernel so we can be a bit casual about its API.

PR #926, opened 2015-07-31
[WFCORE-849] : XmlCompletionScannerTest fails with IBM JDK 8.

ehsavoie:
Because of IBM Jdk aggressive gc the logger gets gced so the handler is no longer configured.
keeping a strong reference on the logger did the trick.

Jira: https://issues.jboss.org/browse/WFCORE-849

### DIFF
--- a/deployment-scanner/src/test/java/org/jboss/as/server/deployment/scanner/XmlCompletionScannerTest.java
+++ b/deployment-scanner/src/test/java/org/jboss/as/server/deployment/scanner/XmlCompletionScannerTest.java
@@ -33,23 +33,39 @@ import static org.hamcrest.CoreMatchers.not;
 import static org.hamcrest.CoreMatchers.nullValue;
 import static org.junit.Assert.assertThat;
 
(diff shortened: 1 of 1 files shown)


deployment-scanner/src/test/java/org/jboss/as/server/deployment/scanner/XmlCompletionScannerTest.java
```diff
@@ -40,16 +40,18 @@
 public class XmlCompletionScannerTest {
         
     private final FakeHandler handler;
+    private final org.jboss.logmanager.Logger lmLogger;
```
jamezp: I don't see any reason to use an `org.jboss.logmanager.Logger` here. Using a `java.util.logging.Logger` should work the same.

That said I'm somewhat surprised this test works at all. I would think a simple reference to the `DeploymentScannerLogger.ROOT_LOGGER`  should avoid the GC. If not then I'd say we need to change the test as the test container is using a different log context.

PR #931, opened 2015-08-03
Add request controller capabiity

stuartwdouglas:


request-controller/src/main/java/org/wildfly/extension/requestcontroller/RequestController.java
```diff
@@ -59,7 +59,7 @@
  */
 public class RequestController implements Service<RequestController>, ServerActivity {
 
-    public static final ServiceName SERVICE_NAME = ServiceName.JBOSS.append("server", "global-request-controller");
+    public static final ServiceName SERVICE_NAME = RequestControllerRootDefinition.REQUEST_CONTROLLER_CAPABILITY.getCapabilityServiceName();
```
jamezp: I think we should keep the old name for now, maybe deprecate it, and set it as an alias on the service.

The service should register the name with `OperationContext.getCapabilityServiceName("org.wildfly.request-controller", RequestController.class)`.

stuartwdouglas: Why? Service names are internal, and anyone referencing the request controller service name should have done it through the SERVICE_NAME constant anyway. If we keep the old name it will not be used anywhere.

jamezp: In this case it's probably okay, but generally service names are retrieved from the [`RuntimeCapabilityRegistry`](https://github.com/wildfly/wildfly-core/blob/master/controller/src/main/java/org/jboss/as/controller/capability/registry/RuntimeCapabilityRegistry.java#L47-47). This is implemented in the [`ModelControllerImpl`](https://github.com/wildfly/wildfly-core/blob/master/controller/src/main/java/org/jboss/as/controller/ModelControllerImpl.java#L1420-L1423). Currently I don't think the `CapabilityContext` will be an issue with this use-case, but I don't know where Brian is going with it.

jamezp: Just to add to Brian is the real authority here. I'm just suggesting this based on what was done elsewhere. If this can wait until after Beta1 when he's back that's fine too. I don't know that this is wrong, it's just not what we've done else where that I know of.

request-controller/src/main/java/org/wildfly/extension/requestcontroller/RequestControllerRootDefinition.java
```diff
@@ -59,6 +62,10 @@
             .build();
     public static final RequestControllerRootDefinition INSTANCE = new RequestControllerRootDefinition(true);
 
+    static final RuntimeCapability<Void> REQUEST_CONTROLLER_CAPABILITY =
+            RuntimeCapability.Builder.of(REQUEST_CONTROLLER_CAPABILITY_NAME, false, RequestController.class)
+                    .build();
+
     private final boolean registerRuntimeOnly;
 
     RequestControllerRootDefinition(boolean registerRuntimeOnly) {
```
ctomc: While you are here, 
can you also register capability on resource so we have info of what resource provides what.
it is a new api not used much yet but everything will soon move to this, as it is needed to make proper model reference impl.

just add 
@Override
    public void registerCapabilities(ManagementResourceRegistration resourceRegistration) {
        resourceRegistration.registerCapability(REQUEST_CONTROLLER_CAPABILITY);
    }

request-controller/src/main/java/org/wildfly/extension/requestcontroller/RequestController.java
bstansberry: This is fine, although it would be better if this constant was deprecated and then moved to internal use only once the uses of it in full have shifted to using the capability instead.

In a case or two we've done as James suggested kept the old ServiceName around and registered the service under that name as an alias as a help for folks who may not have used the old constant. But I don't think that's needed here; the request controller stuff is new code without time for old crufty external user code to build up around it.

PR #934, opened 2015-08-04
Removing repeated disconnected messages.

Gonthim:
https://issues.jboss.org/browse/WFCORE-863

### DIFF
--- a/cli/src/main/java/org/jboss/as/cli/impl/CommandContextImpl.java
+++ b/cli/src/main/java/org/jboss/as/cli/impl/CommandContextImpl.java
@@ -387,10 +387,6 @@ class CommandContextImpl implements CommandContext, ModelControllerClientFactory
                         terminateSession();
                     else {
                         handleSafe(output.getBuffer().trim());
-                        if (client == null && !terminate) {
-                            printLine("You are disconnected at the moment. Type 'connect' to connect to the server or" +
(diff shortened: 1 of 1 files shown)


jamezp:
This stops the message from being printed at all. Instead each line gets the following repeated after a disconnect.
```
The command is not available in the current context (e.g. required subsystems or connection to the controller might be unavailable).
```

The `You are disconnected at the moment. Type 'connect' to connect to the server or 'help' for the list of supported commands.` is no longer printed at all which I think is more valuable than the other repeated message.

Gonthim:
This replicates the behavior before the AESH upgrade. The disconnected message is printed once when the CLI starts and the context message is printed if commands that require a connection are attempted.

The benefit is that commands that don't require a connection, like 'echo' or alias or module, won't get hit with disconnected messages that aren't relevant.

I think this should go in. Changing the error message for commands that require a connection might be useful, but that's a different problem.

Gonthim:
This replicates the behavior before the AESH upgrade. The disconnected message is printed once when the CLI starts and the context message is printed if commands that require a connection are attempted.

The benefit is that commands that don't require a connection, like 'echo' or alias or module, won't get hit with disconnected messages that aren't relevant.

I think this should go in. Changing the error message for commands that require a connection might be useful, but that's a different problem.

PR #936, opened 2015-08-04
[WFCORE-867] Delegate to the AbstractAddStepHandler to discover if a runtime step is required.

darranl:


darranl:
This needs to be tested again after the following PR has been merged - https://github.com/wildfly/wildfly/pull/7877

For some reason ServiceRemoveStepHandler was extended and most of it's methods overidden.

darranl:
Retest this please

controller/src/main/java/org/jboss/as/controller/ServiceRemoveStepHandler.java
```diff
@@ -1,3 +1,25 @@
+/*
+ * JBoss, Home of Professional Open Source.
+ * Copyright 2013, Red Hat, Inc., and individual contributors
```
jamezp: If we're adding a copyright we might as well use the current year :)

jamezp:
@bstansberry WDYT?

controller/src/main/java/org/jboss/as/controller/ServiceRemoveStepHandler.java
darranl: No this is the year the class was written - 2013 is not an accident ;-)

bstansberry:
@jamezp it's fine. It's a behavior change but a totally logical one and WF 10 is a major.

When I saw this JIRA this morning I was quite nervous about scope, but this particular case is quite limited in scope.

PR #937, opened 2015-08-05
[WFCORE-802] Allow log files in subdirectories of the relative jbossserver.log.dir to be listed as log file resources.

jamezp:
Currently subdirectories of the `jboss.server.log.dir` are ignored when listing log files as resources. This change allows log files in subdirectories to be listed under the `subsystem=logging/log-file=*` resource. The deprecated `subsystem=logging:list-log-files` and `subsystem=logging:read-log-file` operations do not files in the subdirectories.

### DIFF
--- a/logging/src/main/java/org/jboss/as/logging/logging/LoggingLogger.java
+++ b/logging/src/main/java/org/jboss/as/logging/logging/LoggingLogger.java
@@ -832,12 +832,16 @@ public interface LoggingLogger extends BasicLogger {
     IllegalStateException extensionNotInitialized();
 
     /**
-     * Creates an exception indicating a failure to read the log file
+     * Creates an exception indicating a failure to read the log file.
+     * <p>
+     * Thrown as a {@link RuntimeException} because the operation did not fail. The cause is likely a failure to read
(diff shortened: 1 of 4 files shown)


logging/src/main/java/org/jboss/as/logging/logging/LoggingLogger.java
```diff
@@ -840,7 +840,7 @@
      * @return an {@link OperationFailedException} for the error
      */
     @Message(id = 79, value = "Failed to read the log file '%s'")
-    RuntimeException failedToReadLogFile(@Cause Throwable cause, String name);
+    OperationFailedException failedToReadLogFile(@Cause Throwable cause, String name);
```
bstansberry: Why the change? If the failure is due to user error (bad input) the OFE is fine, but if it's a server fault, throw a RuntimeException.

jamezp: Maybe that's why I had it a `RuntimeException` to begin with. I was kind of kicking myself for not making a note of why I changed that in the first place to a `RuntimeException`.

PR #938, opened 2015-08-05
[WFCORE-868] CLI if/else is not handling special characters strings w…

TomasHofman:
…ith parenthesis

https://issues.jboss.org/browse/WFCORE-868
https://bugzilla.redhat.com/show_bug.cgi?id=1249773

### DIFF
--- a/cli/src/main/java/org/jboss/as/cli/handlers/ifelse/ExpressionParser.java
+++ b/cli/src/main/java/org/jboss/as/cli/handlers/ifelse/ExpressionParser.java
@@ -138,6 +138,8 @@ public class ExpressionParser {
                     } else {
                         --depthCount;
                     }
+                } else if (input.charAt(endIndex) == '"') { // traverse over string operands surrounded by quotes
+                    endIndex = traverseStringInQuotes(endIndex);
                 }
             }
             throw new IllegalStateException("Failed to locate closing ')' for '(' at " + pos + " in '" + input + "'");
@@ -179,6 +181,26 @@ public class ExpressionParser {
         }
 
         return operand;
+    }
+
+    /**
+     * Traverses over string surrounded by quotation marks.
+     *
+     * @param index index of opening quote
+     * @return index of closing quote
+     */
+    private int traverseStringInQuotes(int index) {
+        if (input.charAt(index) == '"') {
+            while (++index < input.length()) {
+                if (input.charAt(index) == '\\') {
+                    index++;
+                } else if (input.charAt(index) == '"') {
+                    return index;
+                }
+            }
+            throw new IllegalStateException("The closing '\"' is missing.");
+        }
+        return index;
     }
 
     protected ComparisonOperation tryComparison() {
--- a/cli/src/test/java/org/jboss/as/cli/parsing/ifelse/test/IfExpressionParsingTestCase.java
+++ b/cli/src/test/java/org/jboss/as/cli/parsing/ifelse/test/IfExpressionParsingTestCase.java
@@ -33,6 +33,7 @@ import static org.jboss.as.cli.handlers.ifelse.ExpressionParser.GT;
 import static org.jboss.as.cli.handlers.ifelse.ExpressionParser.LT;
 import static org.jboss.as.cli.handlers.ifelse.ExpressionParser.NGT;
 import static org.jboss.as.cli.handlers.ifelse.ExpressionParser.NLT;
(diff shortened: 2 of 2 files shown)


bstansberry:
@Gonthim or @aloubyansky please approve

cli/src/main/java/org/jboss/as/cli/handlers/ifelse/ExpressionParser.java
```diff
@@ -138,6 +138,12 @@ protected Operand parseOperand() throws CommandLineException {
                     } else {
                         --depthCount;
                     }
+                } else if (input.charAt(endIndex) == '"') { // traverse over string operands
+                    endIndex = input.indexOf('"', endIndex + 1);
```
jamezp: I'll leave the approval to @aloubyansky or @Gonthim but if escaped quotes, `\"`, are allowed in strings then this won't work.

TomasHofman: True, I will fix that.

TomasHofman:
Updated to handle escaped quotes.

cli/src/main/java/org/jboss/as/cli/handlers/ifelse/ExpressionParser.java
```diff
@@ -138,6 +138,8 @@ protected Operand parseOperand() throws CommandLineException {
                     } else {
                         --depthCount;
                     }
+                } else if (input.charAt(endIndex) == '"') { // traverse over string operands surrounded by quotes
```
aloubyansky: If we introduce escaping here then any character could be escaped including the character checked in this loop, which means proceeding handling '(', ')' or '"' we have to make sure there is no '\' (or an uneven number of '\') in front.

aloubyansky: Give me some time to look into this, please. Thanks!

bstansberry:
I've added the hold label to help prevent anyone merging it prior to Alexey's ok. I have no objection to it otherwise (haven't looked) so feel free to merge once @aloubyansky approves.

cli/src/main/java/org/jboss/as/cli/handlers/ifelse/ExpressionParser.java
TomasHofman: @aloubyansky any thoughts on this?

My intention was to allow escaping only _inside_ quote marks - everything properly enclosed in '"' marks would be skipped by this loop. If someone tries to escape '(' or ')' outside of "string" or a starting quote, then I believe behaviour will not change from current one.

aloubyansky: @TomasHofman sorry for this huge delay. I'm going to look into this today.

aloubyansky: @TomasHofman this issue is not that simple. We shouldn't be adding escaping here at all. When the CLI receives an input, it parses the line which already includes processing of escaped characters and resolution of expressions (system properties and variables). The command handler then receives arguments already resolved (with escaping processed too). Unresolved argument value will still be available to the handler too.

So, if we introduce character escaping here, it'll be the second one already. On the other hand, if we ask for the unresolved argument value then we also have to take into account that the line might contain unresolved expressions.

This commit unfortunately doesn't solve the general issue. E.g.
if (outcome == success && result.value != "a64180c0)e7187af86b435229de904104\"") of /system-property=test:read-resource

In the line above my intention was to include the quote into the value using \. But the condition string parsed by ExpressionParser will be (outcome == success && result.value != "a64180c0)e7187af86b435229de904104"")

I wish we could just take your commit in and be done with it but we need a proper parser here. Sorry but thanks for bringing this up!

bstansberry:
I'm closing this, given the discussion above. Thanks for this effort, though. I'm closing instead of leaving it open for further changes because it sounds like any such changes would be non-trivial, in which case it's best to exchange pull requests privately between yourself, Alexey and Joe Wertz.

cli/src/main/java/org/jboss/as/cli/handlers/ifelse/ExpressionParser.java
TomasHofman: I see, thanks for explanation!

PR #941, opened 2015-08-05
WFLY-3854 Make build work on JDK9

ctomc:


cli/src/main/java/org/jboss/as/cli/gui/ManagementModel.java
```diff
@@ -82,9 +82,9 @@ public ManagementModel(ManagementModelNode root, CliGuiContext cliGuiCtx) {
      */
     public ManagementModelNode findNode(String address) {
         ManagementModelNode root = (ManagementModelNode)tree.getModel().getRoot();
-        Enumeration<ManagementModelNode> allNodes = root.depthFirstEnumeration();
+        Enumeration<javax.swing.tree.TreeNode> allNodes = root.depthFirstEnumeration();
```
jamezp: What's the reason for this?

ctomc: Swing API got generics :)

PR #943, opened 2015-08-05
Remove elements in child pom's that are inherited from the parent pom.

jamezp:


jamezp:
This failure should be fixed in https://github.com/wildfly/wildfly/pull/7884

PR #944, opened 2015-08-06
[WFCORE-869] CLI never asks to accept cert in non-interactive mode

spyrkob:


darranl:
:warning: DO NOT MERGE - behaviour should only be changed in the case of an additional flag to disable all security prompts.

A valid change will also have to take into account other possible security prompts.

bstansberry:
I'm closing this as it sounds like a design discussion is needed, and any code sharing needed for that can happen external to the PR queue.

PR #946, opened 2015-08-06
CLI Testsuite Refactoring

Gonthim:
Further refactoring to get all of the CLI tests using the structure to follow.


cli/src/main/java/org/jboss/as/cli/impl/CommandContextImpl.java
```diff
@@ -1641,6 +1641,11 @@ private void dohandle(Callback[] callbacks) throws IOException, UnsupportedCallb
                             password = temp.toCharArray();
                         }
                     }
+                    try {
+                        Thread.sleep(5000);
```
ctomc: 5 second sleep? that sounds excessive.

Gonthim: Excessive indeed. Was only for testing.

testsuite/shared/src/main/java/org/jboss/as/test/integration/management/cli/CliProcessWrapper.java
```diff
@@ -0,0 +1,372 @@
...
+        InputStream cliStream = null;
+
+        public CliErrorsReader(InputStream inputStream){
+            cliStream = inputStream;
+        }
+
+        @Override
+        public void run() {
+            java.io.InputStreamReader isr = new java.io.InputStreamReader(cliStream);
+            java.io.BufferedReader br = new java.io.BufferedReader(isr);
```
ctomc: this two are never closed,
probably best if you do add it to try-with-resoruces block

testsuite/shared/src/main/java/org/jboss/as/test/integration/management/cli/CliProcessWrapper.java
```diff
@@ -0,0 +1,372 @@
...
+        }
+
+    }
+
+    private void logErrors() {
+        //System.out.println("Failed to execute '" + cmd + "'");
+        System.out.println("Command's output: '" + cliOutputBuffer + "'");
+
+        java.io.InputStreamReader isr = new java.io.InputStreamReader(cliProcess.getErrorStream());
+        java.io.BufferedReader br = new java.io.BufferedReader(isr);
```
ctomc: same here, readers not closed.

bstansberry:
Superceded by #959, which is merged.

PR #951, opened 2015-08-11
 [WFCORE-74] : Use 'filesystem-path' arbitrary descriptor for attributes/params that represent a path.

ehsavoie:
Adding support for arbitrary descriptors metadata.
Adding FILESYSTEM_PATH and WEB_URL arbitrary descriptors.
Adding support in CLI for FILESYSTEM_PATH during completion.
Adding CLI operation parameter validation before sending operation.

Jira: https://issues.jboss.org/browse/WFCORE-74


model-test/src/main/java/org/jboss/as/model/test/ModelTestModelDescriptionValidator.java
```diff
@@ -113,6 +115,8 @@
         validChildTypeKeys.put(MIN_OCCURS, SimpleIntDescriptorValidator.INSTANCE);
         validChildTypeKeys.put(MAX_OCCURS, SimpleIntDescriptorValidator.INSTANCE);
         validChildTypeKeys.put(ALLOWED, StringListValidator.INSTANCE);
+        validChildTypeKeys.put(FILESYSTEM_PATH, BooleanDescriptorValidator.INSTANCE);
```
ctomc: boolean descriptor validator?

ehsavoie: What is wrong with that ?

ctomc: nothing, i wrongly read it above and tough it was a string and not a boolean. 
sorry for confusion

bstansberry:
Please remove the validation from this PR. It is not required.

bstansberry:
My apologies; I thought I'd posted this last week but it seems I didn't.

Please remove the validation stuff. It's not really part of WFCORE-74, it's too late in the cycle to deal with this kind of behavior change, and the general type checking is something I very much oppose.

cli/src/main/java/org/jboss/as/cli/Util.java
```diff
@@ -181,7 +182,7 @@
     public static final String WRITE_ATTRIBUTE = "write-attribute";
 
     public static boolean isWindows() {
-        return WildFlySecurityManager.getPropertyPrivileged("os.name", null).toLowerCase(Locale.ENGLISH).indexOf("windows") >= 0;
+        return WildFlySecurityManager.getPropertyPrivileged("os.name", null).toLowerCase(Locale.ENGLISH).contains("windows");
```
jamezp: Why the change? This does the same thing as the original code.

ehsavoie: Because I rewrote some code around and might have click an 'old code' hint

cli/src/main/java/org/jboss/as/cli/operation/impl/DefaultOperationCandidatesProvider.java
```diff
@@ -336,6 +310,73 @@ public CommandLineCompleter getValueCompleter() {
         return result;
     }
 
+    private CommandLineCompleter getCompleter(final Map<String, CommandLineCompleterFactory> globalOpProps, final Property prop, CommandContext ctx, String operationName, OperationRequestAddress address) throws IllegalArgumentException {
+        CommandLineCompleter propCompleter = null;
+        final CommandLineCompleterFactory factory = globalOpProps == null ? null : globalOpProps.get(prop.getName());
+        if (factory != null) {
+            final String propName = ctx.getParsedCommandLine().getPropertyValue(Util.NAME);
+            if (Util.WRITE_ATTRIBUTE.equals(operationName) && propName != null) {
```
bstansberry: Why not "add" as well?

bstansberry: Never mind; I assume the block at L325 is meant to handle the "add" case.

cli/src/main/java/org/jboss/as/cli/operation/impl/DefaultOperationCandidatesProvider.java
```diff
@@ -336,6 +310,73 @@ public CommandLineCompleter getValueCompleter() {
...
+        }
+        req.get(Util.OPERATION).set(Util.READ_RESOURCE_DESCRIPTION);
+        try {
+            ModelNode response = ctx.getModelControllerClient().execute(req);
+            final ModelNode result = response.get(Util.RESULT);
+            if (result.isDefined()) {
+                ModelNode allAttrs = result.get(Util.ATTRIBUTES);
+                if (allAttrs.isDefined()) {
+                    final ModelNode propDescr = allAttrs.get(propName);
+                    if (propDescr.isDefined() && propDescr.get(Util.FILESYSTEM_PATH).asBoolean()) {
```
bstansberry: If you want, L365 - 370 can be simplified down to

if (response.get(Util.RESULT, Util.ATTRIBUTES, propName, Util.FILESYSTEM_OATH).asBoolean()) {

controller/src/main/java/org/jboss/as/controller/AttributeDefinition.java
```diff
@@ -849,7 +860,11 @@ public ModelNode getNoTextDescription(boolean forOperation) {
             }
         }
         addAllowedValuesToDescription(result, validator);
-
+        arbitraryDescriptors.entrySet().stream().forEach((arbitraryDescriptor) -> {
+            if(!result.hasDefined(arbitraryDescriptor.getKey())) {//You can't override an arbitrary descriptor set through other properties.
```
bstansberry: There should be an assert for this, as it's a programming mistake that should fail tests.

cli/src/main/java/org/jboss/as/cli/operation/impl/DefaultOperationCandidatesProvider.java
bstansberry: I think this logic should be encapsulated within the CommandLineCompleterFactory created here:

https://github.com/wildfly/wildfly-core/blob/master/cli/src/main/java/org/jboss/as/cli/operation/impl/DefaultOperationCandidatesProvider.java#L384

It's the job of the CommandLineCompleterFactory to provide the completer.

@Gonthim / @aloubyansky WDYT?

ehsavoie: We could have the FilenameTabCompleter.translatePath take the CommandContext as argument, thus we could have a constructor without CommandContext, and some changes in FileSystemPathArgument which would then requires the CommandContext in its constructor.

aloubyansky:
The CLI part looks good. Thanks!

ehsavoie:
retest this please

PR #952, opened 2015-08-12
[WFCORE-833] Capability-based model reference validation for profiles…

bstansberry:
…, socket-binding-groups, server-groups and server-configs

Involves changes to AbstractControllerService/ModelControllerImpl/OperationContextImpl to allow partial deferral of capability resolution to account for the fact that during HostController boot the entire set of configured capabilities is not known until the last of several operation executions is complete (i.e. domain.xml content is not known when host.xml content is installed.)

Also includes a fair bit of test fixture tweaks related to supporting capability resolution testing in core-model-test.

This PR includes a fair amount of commented out logic in DomainModelReferenceValidator and in some unit tests of OSHs in the host-controller module. My intent in doing that was to perhaps make it easier for @kabir to see what I was doing here. Once he's had a chance to have a look I can remove the commented code, either in a change to this PR or another one. The unit test changes are because the handlers being directly tested no longer do reference validation. The validation is part of capability resolution. Equivalent testing is now done in core-model-test.

### DIFF
--- a/host-controller/src/test/java/org/jboss/as/domain/controller/operations/ProfileIncludesHandlerTestCase.java
+++ b/host-controller/src/test/java/org/jboss/as/domain/controller/operations/ProfileIncludesHandlerTestCase.java
@@ -94,15 +94,16 @@ public class ProfileIncludesHandlerTestCase extends AbstractOperationTestCase {
         operationContext.executeNextStep();
     }
 
-    @Test(expected=OperationFailedException.class)
-    public void testBadProfileIncludesWrite() throws Exception {
-        PathAddress addr = getProfileAddress("profile-one");
-        ModelNode list = new ModelNode().add("bad-profile");
-        ModelNode op = Util.getWriteAttributeOperation(addr, INCLUDES, list);
-        MockOperationContext operationContext = getOperationContext(addr);
-        ProfileResourceDefinition.createReferenceValidationHandler().execute(operationContext, op);
-        operationContext.executeNextStep();
-    }
+//    // WFCORE-833 Replaced with test in core-model-test ProfileTestCase
+//    @Test(expected=OperationFailedException.class)
+//    public void testBadProfileIncludesWrite() throws Exception {
+//        PathAddress addr = getProfileAddress("profile-one");
+//        ModelNode list = new ModelNode().add("bad-profile");
+//        ModelNode op = Util.getWriteAttributeOperation(addr, INCLUDES, list);
+//        MockOperationContext operationContext = getOperationContext(addr);
+//        ProfileResourceDefinition.createReferenceValidationHandler().execute(operationContext, op);
+//        operationContext.executeNextStep();
+//    }
 
     @Test(expected=OperationFailedException.class)
     public void testCyclicProfileIncludesWrite() throws Exception {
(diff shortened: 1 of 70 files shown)


host-controller/src/test/java/org/jboss/as/domain/controller/operations/SocketBindingGroupIncludesHandlerTestCase.java
```diff
@@ -100,15 +101,16 @@ public void testGoodSocketBindingGroupIncludesWrite() throws Exception {
...
+//        PathAddress addr = getSocketBindingGroupAddress("binding-one");
+//        ModelNode list = new ModelNode().add("bad-SocketBindingGroup");
+//        ModelNode op = Util.getWriteAttributeOperation(addr, INCLUDES, list);
+//        MockOperationContext operationContext = getOperationContext(addr);
+//        SocketBindingGroupResourceDefinition.createReferenceValidationHandler().execute(operationContext, op);
+//        operationContext.executeNextStep();
+//    }
 
     @Test(expected=OperationFailedException.class)
     public void testCyclicSocketBindingGroupIncludesWrite() throws Exception {
```
kabir: Is there any reason this could not be moved as well?

host-controller/src/test/java/org/jboss/as/domain/controller/operations/ProfileIncludesHandlerTestCase.java
```diff
@@ -94,15 +94,16 @@ public void testGoodProfileIncludesWrite() throws Exception {
...
+//        PathAddress addr = getProfileAddress("profile-one");
+//        ModelNode list = new ModelNode().add("bad-profile");
+//        ModelNode op = Util.getWriteAttributeOperation(addr, INCLUDES, list);
+//        MockOperationContext operationContext = getOperationContext(addr);
+//        ProfileResourceDefinition.createReferenceValidationHandler().execute(operationContext, op);
+//        operationContext.executeNextStep();
+//    }
 
     @Test(expected=OperationFailedException.class)
     public void testCyclicProfileIncludesWrite() throws Exception {
```
kabir: Is there any reason this could not be moved as well?

kabir:
Perhaps for the core-model tests some ‘good’ adds/changes, although I guess the booting of the xml handles that. Still there are some of those remaining in the host-controller tests, and it is not obvious to me why they cannot be removed (this is not just the ‘cyclic’ ones I commented on in-line).

For all the ModelInitializers for the core model tests perhaps there could be a better way to centralize the capability setup stuff, but that can come later (if it is even possible).

+1 for leaving the commented test methods in and fixing in a follow-up commit/PR. It makes it clear what has moved where.

bstansberry:
@kabir The others could be moved, although as you say the adds are already covered (although there may be some corner case not covered.) The existing tests in host-controller are unit tests of the handler classes themselves. So they should be faster than an integration test in core-model-test, and in some cases like the "OverriddenSubsystems" ones, they look easier to set up as unit tests. There also may be some where some of the assertions possible because of the mock OperationContext could not be reproduced.

I just moved the ones that failed because the handler alone could no longer produce the expected error. Granted, leaving some others was just laziness. ;)

+1 on improving the ModelInitializer stuff. As I started I figured I'd need some utilities or something, but I never had any great ideas and the pain level never got real high. But I figure as we go along something will get done.  A constant no-op ModelInitialization is kind of obvious, and I speculated some sort of chaining MI would be useful (to allow combination of other "standard" ones to fit the needs of particular tests.) But I never really needed that for this PR.

bstansberry:
Rebased. Plus I added a NO_OP ModelInitializer. Kabir approved it before, so I'll merge it once the tests run.

PR #954, opened 2015-08-13
WFCORE-879 Allow AbstractLegacyExtension to boot in ADMIN_ONLY mode i…

stuartwdouglas:
…n standalone servers

Also allows a server to boot in admin only mode when there are unsatisfied requirements (and will continue to allow operations to ignore them until the server is no longer in an error state).

This is necessary for EAP6 migrations, as the remoting services will have an unsatisfied requirement on the XNIO worker that is provided by the new IO subsystem (and added by the Undertow migration op)

### DIFF
--- a/controller/src/main/java/org/jboss/as/controller/ModelControllerImpl.java
+++ b/controller/src/main/java/org/jboss/as/controller/ModelControllerImpl.java
@@ -1242,8 +1242,8 @@ class ModelControllerImpl implements ModelController {
          * @return a map whose keys are missing capabilities and whose values are the names of other capabilities
          *         that require that capability. Will not return {@code null} but may be empty
          */
-        CapabilityValidation validateCapabilityRegistry() {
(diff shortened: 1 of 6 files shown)


stuartwdouglas:
should not be merged without @bstansberry say so

ctomc:
i think some of this changes clash / duplicate changes in https://github.com/wildfly/wildfly-core/pull/953

bstansberry:
WFCORE-879 is a dup of WFCORE-875, so the first commit duplicates #953. But we can go with this one, as I like your error wording better. Plus you have the other commit, which is valid. Even if it WILL force me to rebase #952. ;) 

In an ideal world, that "capabilitiesAlreadyBroken" handling would be a bit more sophisticated such that we could compare the new validation result to the "capabilitiesAlreadyBroken" one and not ignore new problems. But that's a nice to have that can come later, or never.

PR #957, opened 2015-08-14
WFCORE-874 Handle include-aliases=false properly for singleton resour…

kabir:
…ces in read-children-types

https://issues.jboss.org/browse/WFCORE-874


bstansberry:
retest this please

bstansberry:
The domain hangs are due to https://github.com/wildfly/wildfly/pull/7930

PR #958, opened 2015-08-14
WFCORE-806 Collect capabilty data & references

ctomc:
- make capability registry  copy on commit/publish and rollback aware
 - make capability registry standalone class
 - expose registry to server entry classes
 - simply registration classes

needs review by @bstansberry


ctomc:
retest this please

PR #961, opened 2015-08-17
Add missing constants to ModelDescriptionConstants

stuartwdouglas:


bstansberry:
Thanks, Stuart. Having that other ModelDescriptionConstants class was a wrong turn. This one should have everything in the kernel management API.

PR #964, opened 2015-08-18
WFCORE-806 Collect capabilty data & references 

ctomc:
- make capability registry  copy on commit/publish and rollback aware
 - make capability registry standalone class
 - expose registry to server entry classes
 - simply registration classes

needs review by @bstansberry


controller/src/main/java/org/jboss/as/controller/AbstractAddStepHandler.java
```diff
@@ -274,6 +275,11 @@ protected void recordCapabilitiesAndRequirements(OperationContext context, Model
                 context.registerCapability(capability, null);
             }
         }
+
+        //just here until we fix the registration across whole server
+        if (context.getResourceRegistration().getCapabilities().size() != capabilities.size()){
+            ControllerLogger.ROOT_LOGGER.warnf("Number of possible capabilities registered on resource and number of runtime capabilities don't match for path: %s", context.getCurrentAddress());
+        }
```
bstansberry: This is ok, but please file a Blocker JIRA to get rid of this by CR1 so it doesn't fall into a crack.

controller/src/main/java/org/jboss/as/controller/CapabilityRegistry.java
```diff
@@ -0,0 +1,621 @@
...
+        return getPublishedRegistry().getPossibleProviderPoints(capabilityId);
+    }
+
+    /**
+     * Publish the changes to main registry
+     */
+    synchronized void publish() {
+        if (publishedFullRegistry == null) {
+            throw new RuntimeException("Cannot write directly to main registry");
+        }
```
bstansberry: Nit: can this be an assert?

host-controller/src/main/java/org/jboss/as/host/controller/DomainModelControllerService.java
```diff
@@ -888,6 +893,7 @@ public void stop(final StopContext context) {
             }
         }
         extensionRegistry.clear();
+        capabilityRegistry.clear();
```
bstansberry: Can this move to AbstractControllerService? It would be nice if the clear() method wasn't public.

controller/src/main/java/org/jboss/as/controller/ModelControllerImpl.java
```diff
@@ -1212,27 +1192,8 @@ ManagementModelImpl cloneRootResource() {
...
-                mrr = resourceRegistration;
-                currentResource = rootResource;
-                currentCaps = capabilityRegistry;
-            }
-            CapabilityRegistryImpl clone = currentCaps.copy();
-            ManagementModelImpl result = new ManagementModelImpl(mrr, currentResource, clone);
-            ControllerLogger.MGMT_OP_LOGGER.tracef("cloned to %s to create %s and %s", currentCaps, clone, result);
-            return result;
+            //todo why is this even needed, return this; should be fine.
+            return new ManagementModelImpl(resourceRegistration, rootResource, capabilityRegistry);
```
bstansberry: Please resolve this TODO. If cloneCapabilityRegistry then essentially becomes a no-op, the method should be removed.

controller/src/main/java/org/jboss/as/controller/ModelControllerImpl.java
```diff
@@ -1259,6 +1220,7 @@ private void publish() {
             ModelControllerImpl.this.managementModel.set(this);
             ControllerLogger.MGMT_OP_LOGGER.tracef("published %s", this);
             published = true;
+            capabilityRegistry.publish();
```
bstansberry: The ordering here seems wrong. It may not matter, but it seems unintuitive to log and set the flag, and then continue with the publishing process.

controller/src/main/java/org/jboss/as/controller/ModelControllerImpl.java
```diff
@@ -1731,25 +1286,17 @@ public void close() throws IOException {
...
-                return result;
+        ResponseConverter<ModelNode> TO_MODEL_NODE = or -> {
+            ModelNode result = or.getResponseNode();
+            try {
+                or.close();
+            } catch (IOException e) {
+                ROOT_LOGGER.debugf(e, "Caught exception closing %s whose associated streams, "
+                        + "if any, were not wanted", or);
             }
+            return result;
```
bstansberry: Ugh. Please resist the urge to do this. :)

controller/src/main/java/org/jboss/as/controller/capability/registry/CapabilityRegistration.java
```diff
@@ -31,15 +38,30 @@
...
 
     public CapabilityRegistration(C capability, CapabilityContext context) {
         this.capability = capability;
         this.id = new CapabilityId(capability.getName(), context);
+    }
+
+    public CapabilityRegistration(C capability, CapabilityContext context, RegistrationPoint registrationPoint) {
+            this(capability, context);
+        this.registrationPoints.put(registrationPoint.getAddress(), registrationPoint);
+        }
```
bstansberry: Nit. Some whitespace funkiness here.

controller/src/main/java/org/jboss/as/controller/capability/registry/ImmutableCapabilityRegistry.java
```diff
@@ -0,0 +1,67 @@
...
+     * @param capabilityId id of capability
+     * @return run set of {@link RegistrationPoint} or empty set if no registration points are found
+     */
+    Set<RegistrationPoint> getPossibleProviderPoints(CapabilityId capabilityId);
+
+    Set<CapabilityRegistration> getCapabilities();
+
+    Set<CapabilityRegistration> getPossibleCapabilities();
+
+    ServiceName getCapabilityServiceName(String capabilityName, CapabilityContext context, Class<?> serviceType);
```
bstansberry: Everything needs complete javadoc.

controller/src/main/java/org/jboss/as/controller/capability/registry/RequirementRegistration.java
```diff
@@ -29,6 +37,7 @@
  */
 public class RequirementRegistration {
 
+    protected final Map<PathAddress, RegistrationPoint> registrationPoints = new LinkedHashMap<>();
```
bstansberry: Can this be private? I didn't see an obvious external use, although checking that in github is not an exact thing. ;)

host-controller/src/main/java/org/jboss/as/host/controller/DomainModelControllerService.java
ctomc: yes this can move to AbstractControllerService

controller/src/main/java/org/jboss/as/controller/CapabilityRegistry.java
ctomc: yeah it should be.

server/src/main/java/org/jboss/as/server/controller/resources/ServerRootResourceDefinition.java
```diff
@@ -444,6 +449,9 @@ public String getProductName() {
         // Paths
         resourceRegistration.registerSubModel(PathResourceDefinition.createSpecified(pathManager));
 
+        //capability registry
+        resourceRegistration.registerSubModel(new CapabilityRegistryResourceDefinition(capabilityRegistry));
+
```
bstansberry: We need to work this out for the HC as well. It belongs under host=<localhostname>

controller/src/main/java/org/jboss/as/controller/ModelControllerImpl.java
ctomc: ah, I forgot I did that. IDE tempted me :) 
I can revert this change if you don't feel comfortable with it.

controller/src/main/java/org/jboss/as/controller/capability/registry/RequirementRegistration.java
ctomc: yeah it should be private. It needed to be protected before refactoring, that is no longer the case.

controller/src/main/java/org/jboss/as/controller/ModelControllerImpl.java
bstansberry: Yeah, I'd prefer it. Thanks.

controller/src/main/java/org/jboss/as/controller/capability/registry/ImmutableCapabilityRegistry.java
ctomc: added

controller/src/main/java/org/jboss/as/controller/capability/registry/CapabilityRegistration.java
ctomc: fixed

controller/src/main/java/org/jboss/as/controller/ModelControllerImpl.java
ctomc: reverted

controller/src/main/java/org/jboss/as/controller/ModelControllerImpl.java
ctomc: this was a tricky one, but got to fix it.

bstansberry:
I'll try and summarize how I understand this. Goal is to help ensure I'm on the right track.

1) A kind of "primordial" CR is created, and gets passed into the AbstractControllerService and by it to the root MRR.

2) The MRRs write directly to that primordial MRR.

3) When the ModelController creates the ManagementModel, it creates a "shadow copy" of the primordial CR which is then used by any OperationContext. The shadow copy is able to publish to the primordial one or to restore its state from the primordial one. The OC tells it to do this as part of commit or rollback of write ops.

4) The ImmutableCapabilityRegistry is just a read-only view into either the primordial CR or the shadow copy. Which its a view onto depends on which one the caller doing a read had access to.

a) CapabilityRegistryResourceDefinition gets a read-only view of the primordial MRR.
b) OCs get a read-only view of the shadow copy MRR.

As a result of this, CapabilityRegistryResourceDefinition and related OSHs handling reads will not see most unpublished changes made by concurrently executing writes. This is because they satisfy requests by reading from the primordial MRR. So this is good.

Other OSHs that do reads by reading from the OC will see unpublished changes made to the shadow copy by concurrently executing writes. This isn't real good, although I'm not sure if there's any real harm. The OSHs that use the CR outside of write ops are the ones related to CapabilityRegistryResourceDefinition, and they don't read from the OC; they read from the read-only view of the primordial CR.

One thing to look into, and I'll look tomorrow too, is the whole CapabilityServiceSupport thing. I suspect those classes will be able to see unpublished writes to the shadow copy. I haven't thought through the implications of that. :)

One concern I have is that the MRRs are writing to the primordial CR. This will happen in the course of management ops (e.g. extension=foo:add). Two concerns:

a) (minor) Ops that are reading the CapabilityRegistryResourceDefinition stuff will see changes from ops that are not yet committed.
b) (MAJOR) I think that in the course of that op that's adding an extension, the MRRs will be writing to the primordial CR while any other changes will be made to the shadow copy. Then when the op commits, the  changes made by the MRRs will be lost.

Is a solution to b) to give the MRRs a ref to the shadow copy instead of the primordial CR? Any writes made by the MRRs should be being done within the scope of a management op, with the commit/rollback of that op determining whether the changes are kept.

The other concern I have is I believe there is inconsistent synchronization. The publish and rollback ops are synchronized, but the other reads and write are not. The ImmutableCapabilityRegistry makes protecting the data harder, because refs to the internal maps are provided to the ICRImpl. Perhaps CR will need to use a lock object instead of "synchronized" and then ICRImpl will need to be given a ref to that lock so it can participate in the locking.

ctomc:
I was actually trying to figure out this exact problem that you describe with MRRs writing directly to primordial CR.
I was able to reproduce it when testing the after boot registration of capabilities, but couldn't narrow it down completely. 
So thank you for in depth analysis.

controller/src/main/java/org/jboss/as/controller/AbstractAddStepHandler.java
ctomc: added https://issues.jboss.org/browse/WFCORE-912

server/src/main/java/org/jboss/as/server/controller/resources/ServerRootResourceDefinition.java
ctomc: Implemented also this.

controller/src/main/java/org/jboss/as/controller/CapabilityRegistry.java
```diff
@@ -0,0 +1,745 @@
...
+        try {
+            registerRequirement(requirement);
+        } finally {
+            writeLock.unlock();
+        }
+    }
+
+    private void registerRequirement(RuntimeRequirementRegistration requirement) {
+        try {
+            writeLock.lock();
```
bstansberry: This should be outside the try if the release is in the finally.

But you don't need to lock/unlock in this method, as it is called with the lock held. You can document it as "must be called with the write lock held".

controller/src/main/java/org/jboss/as/controller/CapabilityRegistry.java
```diff
@@ -0,0 +1,745 @@
...
+     * Publish the changes to main registry
+     */
+    void publish() {
+        assert publishedFullRegistry != null : "Cannot write directly to main registry";
+
+        if (!modified) {
+            return;
+        }
+        try {
+            writeLock.lock();
```
bstansberry: This protects the data of 'this' but a lot of the data that's being modified is 'publishedFullRegistry'.

Since this.modified is being changed, protecting that with this.writeLock is ok, but then publishedFullRegistry.writeLock needs to be taken to protect the publishedFullRegistry data.

I don't see any case where publishedFullRegistry would take its own writeLock and then try to do something that results in trying to take this.writeLock, so there's no risk of deadlock from taking the two locks.

controller/src/main/java/org/jboss/as/controller/CapabilityRegistry.java
```diff
@@ -0,0 +1,745 @@
...
+     * Discard the changes.
+     */
+    void rollback() {
+        if (publishedFullRegistry == null) {
+            return;
+        }
+        try {
+            writeLock.lock();
+            clear();
+            copy(publishedFullRegistry, this);
```
bstansberry: Here, publishedFullRegistry.readLock needs to be taken.

controller/src/main/java/org/jboss/as/controller/CapabilityRegistry.java
```diff
@@ -0,0 +1,745 @@
...
+            modified = false;
+        } finally {
+            writeLock.unlock();
+        }
+    }
+
+    boolean isModified() {
+        return modified;
```
bstansberry: Needs readLock.

controller/src/main/java/org/jboss/as/controller/CapabilityRegistry.java
```diff
@@ -0,0 +1,745 @@
...
+     *
+     * @param requirementRegistration the requirement. Cannot be {@code null}
+     * @see #registerAdditionalCapabilityRequirement(org.jboss.as.controller.capability.registry.RuntimeRequirementRegistration)
+     */
+    @Override
+    public void removeCapabilityRequirement(RuntimeRequirementRegistration requirementRegistration) {
+        // We don't know if this got registered as an runtime-only requirement or a hard one
+        // so clean it from both maps
+        try {
+            writeLock.lock();
```
bstansberry: All the lock acquisitions need to be moved outside the try/finally that releases them. See Java Concurrency in Practice listing 13.2, - 13.5.

controller/src/main/java/org/jboss/as/controller/CapabilityRegistry.java
ctomc: Does it really? modifed field only gets changed on write operations and should affect read ops.

controller/src/main/java/org/jboss/as/controller/CapabilityRegistry.java
ctomc: Makes complete sense, fixed it that way.

controller/src/main/java/org/jboss/as/controller/CapabilityRegistry.java
ctomc: you are right it does need it. fixed it.

controller/src/main/java/org/jboss/as/controller/CapabilityRegistry.java
```diff
@@ -0,0 +1,755 @@
...
+     * @param context           the context in which the capability is registered. Cannot be {@code null}
+     * @param registrationPoint the specific registration point that is being removed
+     * @return the capability that was removed, or {@code null} if no matching capability was registered or other
+     * registration points for the capability still exist
+     */
+    @Override
+    public RuntimeCapabilityRegistration removeCapability(String capabilityName, CapabilityContext context,
+                                                          PathAddress registrationPoint) {
+        try {
+            writeLock.lock();
```
bstansberry: All of the lock acquisitions need to be outside the try/finally.

controller/src/main/java/org/jboss/as/controller/CapabilityRegistry.java
```diff
@@ -0,0 +1,755 @@
...
+
+    /**
+     * Publish the changes to main registry
+     */
+    void publish() {
+        assert publishedFullRegistry != null : "Cannot write directly to main registry";
+
+        if (!modified) {
+            return;
+        }
```
bstansberry: This goes in the writeLock, as 'modified' is guarded by it.

controller/src/main/java/org/jboss/as/controller/CapabilityRegistry.java
```diff
@@ -0,0 +1,755 @@
...
+        writeLock.lock();
+        publishedFullRegistry.writeLock.lock();
```
bstansberry: Use a nested try/finally for this publishedFullRegistry lock so there is no possibility of this.writeLock not getting released.

controller/src/main/java/org/jboss/as/controller/CapabilityRegistry.java
```diff
@@ -0,0 +1,755 @@
...
+
+    /**
+     * Discard the changes.
+     */
+    void rollback() {
+        if (publishedFullRegistry == null) {
+            return;
+        }
+        writeLock.lock();
+        publishedFullRegistry.readLock.lock();
```
bstansberry: Use a nested try/finally for this publishedFullRegistry lock so there is no possibility of this.writeLock not getting released.

controller/src/main/java/org/jboss/as/controller/ParallelBootOperationContext.java
```diff
@@ -324,6 +324,11 @@ void awaitServiceContainerStability() throws InterruptedException {
     }
 
     @Override
+    void publishCapabilityRegistry() {
+        primaryContext.publishCapabilityRegistry();
```
bstansberry: This should be a no-op. The primary context controls publication when everything is ready; the ParallelBootOperationContexts only know what has happened with their bit.

controller/src/main/java/org/jboss/as/controller/AbstractOperationContext.java
```diff
@@ -707,6 +712,7 @@ public void rollback() {
                     persistenceResource.commit();
                 }
             }
+            publishCapabilityRegistry();
```
bstansberry: There are 3 cases here:

1) persistenceResource == null
2) persistenceResource != null && resultAction == ResultAction.KEEP
3) persistenceResource != null && resultAction == ResultAction.ROLLBACK

The publishCapabilityRegistry(); should only happen in the first two cases.

The 3rd case already gets handled by the subsequent call to executeResultHandlerPhase, which calls operationRollingBack if resultAction != ResultAction.KEEP.

controller/src/main/java/org/jboss/as/controller/capability/registry/ImmutableCapabilityRegistry.java
```diff
@@ -0,0 +1,82 @@
...
+import org.jboss.msc.service.ServiceName;
+
+/**
+ * A read-only view of {@link CapabilityRegistry}
+ *
+ * @author Tomaz Cerar (c) 2015 Red Hat Inc.
+ */
+public interface ImmutableCapabilityRegistry {
+    /**
+     * Gets whether a capability with the given name is registered.
```
bstansberry: This can come after merge, but the javadoc in this interface could use clarifiication re: the "possible capabilities" vs the "actual" capabilities. Perhaps an explanation in the class level javadoc of the distinction between the two concepts, and then in each method a clear statement re: which of the two is relevant.

controller/src/main/java/org/jboss/as/controller/capability/registry/PossibleCapabilityRegistry.java
```diff
@@ -0,0 +1,50 @@
...
+
+/**
+ * @author Tomaz Cerar (c) 2015 Red Hat Inc.
+ */
+public interface PossibleCapabilityRegistry {
+
+    /**
+     * Registers a capability with the system. Any
+     * {@link org.jboss.as.controller.capability.AbstractCapability#getRequirements() requirements}
+     * associated with the capability will be recorded as requirements.
```
bstansberry: This second sentence should be removed.

controller/src/main/java/org/jboss/as/controller/capability/registry/PossibleCapabilityRegistry.java
```diff
@@ -0,0 +1,50 @@
...
+ * limitations under the License.
+ */
+
+package org.jboss.as.controller.capability.registry;
+
+import org.jboss.as.controller.PathAddress;
+import org.jboss.as.controller.capability.Capability;
+
+/**
+ * @author Tomaz Cerar (c) 2015 Red Hat Inc.
```
bstansberry: Can come later, but an explanation of what a "possible capability" is should be added.

server/src/main/java/org/jboss/as/server/controller/resources/ServerRootResourceDefinition.java
bstansberry: I *may* want to disable this before CR1 just because there's no client that will use it yet, and committing to support the end-user API for 10 years is a big step to take if we don't have to. What you have seems ok, so this is really just me being cautious. I'm a bit iffy about the point.getAddress().toCLIStyleString() for the reg points. We do that elsewhere but this data may be more directly consumed by the console, while other spots we do that (I think some RBAC config data) the console at most would just present the info.

controller/src/main/java/org/jboss/as/controller/capability/registry/ImmutableCapabilityRegistry.java
ctomc: improved javadoc a bit, but could use some more tuning in future.

controller/src/main/java/org/jboss/as/controller/capability/registry/PossibleCapabilityRegistry.java
ctomc: removed

controller/src/main/java/org/jboss/as/controller/capability/registry/PossibleCapabilityRegistry.java
ctomc: added some crude definition of what it is.

controller/src/main/java/org/jboss/as/controller/AbstractOperationContext.java
ctomc: change logic to only publish registry 
in case ```resultAction != ResultAction.ROLLBACK```

ctomc:
woohoo! :-)

PR #966, opened 2015-08-18
WFLY-5112 Set map attribute to undefined if it is empty after remove …

tadamski:
…operation

### DIFF
--- a/controller/src/main/java/org/jboss/as/controller/operations/global/MapOperations.java
+++ b/controller/src/main/java/org/jboss/as/controller/operations/global/MapOperations.java
@@ -155,7 +155,9 @@ public class MapOperations {
         void updateModel(final OperationContext context, ModelNode model, AttributeDefinition attributeDefinition, ModelNode attribute) throws OperationFailedException {
             String key = KEY.resolveModelAttribute(context, model).asString();
             attribute.remove(key);
+            if(attribute.asList().isEmpty()){
+                attribute.clear();
+            }
         }
     }
-
 }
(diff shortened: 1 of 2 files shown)


controller/src/main/java/org/jboss/as/controller/operations/global/MapOperations.java
```diff
@@ -155,6 +157,9 @@ private MapRemoveHandler() {
         void updateModel(final OperationContext context, ModelNode model, AttributeDefinition attributeDefinition, ModelNode attribute) throws OperationFailedException {
             String key = KEY.resolveModelAttribute(context, model).asString();
             attribute.remove(key);
+            if(attribute.asPropertyList().isEmpty()){
```
ctomc: bit faster way would be attribute.asList() also would avoid some possible issues with type of model

tadamski:
Fixed. I have also added ability to parse empty property list to PropertiesParser. Such configuration is redundant but technically valid IMO.

ctomc:
:+1:

controller/src/main/java/org/jboss/as/controller/operations/global/MapOperations.java
```diff
@@ -155,6 +157,9 @@ private MapRemoveHandler() {
         void updateModel(final OperationContext context, ModelNode model, AttributeDefinition attributeDefinition, ModelNode attribute) throws OperationFailedException {
             String key = KEY.resolveModelAttribute(context, model).asString();
             attribute.remove(key);
+            if(attribute.asList().isEmpty()){
+                attribute.set(UNDEFINED);
```
jamezp: This should be `attribute.clear()`. The static `UNDEFINED` node is mutable.

jamezp:
If this fixes the WFLY-5112 issue then the JIRA needs to be moved to WFCORE. If this is a separate fix then we need a new WFCORE PR.

tadamski:
Fixed. I have created new https://issues.jboss.org/browse/WFCORE-894 for this PR.

jamezp:
retest this please

jamezp:
One more time retest this please

bstansberry:
retest this please

The "Core - Full Integration" failure looks like the pattern we've seen before when some zombie process was sitting around on a CI slave.

PR #969, opened 2015-08-19
Undertow 1.3.0.Beta9

stuartwdouglas:


stuartwdouglas:
This seems to be an issue with an upstream test that is sending an entity body with a 204 (no content) response. This version of Undertow tightened RFC compliance with regard to this, so I will fix the upstream test.

stuartwdouglas:
Retest this please

PR #971, opened 2015-08-20
[WFCORE-891] jboss-deployment-structure.xml resource root allows bad references

elguardian:
JIRA https://issues.jboss.org/browse/WFCORE-891

I tried to cover these two use cases:
1) If the project is deployed exploded, then you can reference
outside the deployment; check whether the reference is a child
of the deployment root

2) the project is an archive. In this case you cannot reference outside
the deployment (VFS Zip file system) but you can reference a
non-existant resource-root; check if the file exist or not.

Note: when the JAR is inside the deployment, this is processed by
EarStructureProcessor. So if it is a jar/sar is mounted in that stage.

### DIFF
--- a/server/src/main/java/org/jboss/as/server/deployment/module/descriptor/JBossDeploymentStructureParser10.java
+++ b/server/src/main/java/org/jboss/as/server/deployment/module/descriptor/JBossDeploymentStructureParser10.java
@@ -522,8 +522,22 @@ public class JBossDeploymentStructureParser10 implements XMLElementReader<ParseR
                             Closeable closable = null;
                             if(overlay != null) {
                                 overlay.remountAsZip(false);
-                            } else if(child.isFile()) {
-                                closable = VFS.mountZip(child, child, TempFileProviderService.provider());
+                            } else {
+                                if(!child.exists()) {
+                                    // the reference does not exist
+                                    throw ServerLogger.ROOT_LOGGER.illegalResoureRootReference(child.getName());
+                                } else if(child.isFile()) {
+                                    try {
+                                        // check if the resource-root is within the deployment or not
+                                       if(child.getPathNameRelativeTo(deploymentRootFile) != null) {
+                                          closable = VFS.mountZip(child, child, TempFileProviderService.provider());
+                                       }
+                                    } catch(IllegalArgumentException err) {
+                                       throw ServerLogger.ROOT_LOGGER.illegalResoureRootReference(child.getName());
+                                    }
+                                }
+                                // the only thing left is either a directory or a jar/sar mounted already by EarStructureProcessor
+                                // not need to do any action
                             }
                             final MountHandle mountHandle = new MountHandle(closable);
                             ResourceRoot resourceRoot = new ResourceRoot(name, child, mountHandle);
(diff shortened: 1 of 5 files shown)


ctomc:
@elguardian can you please rebase this?

server/src/main/java/org/jboss/as/server/logging/ServerLogger.java
```diff
@@ -1159,4 +1159,14 @@ OperationFailedException illegalCombinationOfHttpManagementInterfaceConfiguratio
 
     @Message(id = 232, value = "Could not get module info for module name: %s")
     OperationFailedException couldNotGetModuleInfo(String moduleName, @Cause Throwable cause);
+
+    @Message(id = 233, value = "Undeployed \"%s\" (runtime-name: \"%s\")")
+    String deploymentUndeployedNotification(String managementName, String deploymentUnitName);
```
jamezp: This is never used and be removed

server/src/main/java/org/jboss/as/server/logging/ServerLogger.java
```diff
@@ -1159,4 +1159,14 @@ OperationFailedException illegalCombinationOfHttpManagementInterfaceConfiguratio
...
+
+    @Message(id = 234, value = "Deployed \"%s\" (runtime-name : \"%s\")")
+    String deploymentDeployedNotification(String managementName, String deploymentUnitName);
```
jamezp: This is never used and should be removed.

server/src/main/java/org/jboss/as/server/deployment/module/descriptor/JBossDeploymentStructureParser10.java
```diff
@@ -522,8 +522,22 @@ private static void parseResourceRoot(final DeploymentUnit deploymentUnit, final
...
-                            } else if(child.isFile()) {
-                                closable = VFS.mountZip(child, child, TempFileProviderService.provider());
+                            } else {
+                                if(!child.exists()) {
+                                    // the reference does not exist
+                                    throw ServerLogger.ROOT_LOGGER.illegalResoureRootReference(child.getName());
+                                } else if(child.isFile()) {
+                                    try {
+                                        // check if the resource-root is within the deployment or not
+                                       if(child.getPathNameRelativeTo(deploymentRootFile) != null) {
```
jamezp: This check won't work. The JavaDoc for `VirtualFile.getPathNameRelativeTo()` states:
> Get the path name relative to a parent virtual file. If the given virtual file is not a parent of this virtual file, then an IllegalArgumentException is thrown.

Note this is true for each parser, but I'll just comment on this one.

jamezp: Let me rephrase that. What I mean is that `null` will never be returned. It doesn't seem quite right to use the exception thrown as the determining factor of whether it's outside the deployment or not.

jamezp:
Just a general note for reviewers. @stuartwdouglas stated in [WFCORE-700](https://issues.jboss.org/browse/WFCORE-700?focusedCommentId=13086848&page=com.atlassian.jira.plugin.system.issuetabpanels:comment-tabpanel#comment-13086848)
> &lt;resource-root&gt; was only intended to be used to access files inside the deployment, it was not intended to allow access to external libraries.

So some kind of fix is likely needed. I do wonder if users are relying on this behavior as it's been like this since the beginning I believe.

server/src/main/java/org/jboss/as/server/deployment/module/descriptor/JBossDeploymentStructureParser10.java
elguardian: @jamezp I agree that the solution regarding the check is a hack but I did not find any other way to rely on VFS to check a path being relative to another path. I mean there is no child.isPathRelativeTo(parent). Maybe @ctomc knows if there is anything in the API regarding this kind of checks.

elguardian:
That's exactly the validation this fix is trying to do.

PR #972, opened 2015-08-20
[WFCORE-898] Upgrade to WildFly Elytron 1.0.0.CR1.

darranl:
This is a slimmed version of the library containing just the security manager implementation.


bstansberry:
retest this please

The "Core - Full Integration" failure seems related but the Linux one probably isn't.

PR #973, opened 2015-08-20
[WFCORE-893] Add a default io subsystem to profiles using a legacy re…

bstansberry:
…moting config

This also tweaks the model test validation code to defer validation of boot ops until after the ModelController's initModel method gets a chance to run. That method is where tests can hook in and register resource definitions that allow the validation to complete.

### DIFF
--- a/model-test/src/main/java/org/jboss/as/model/test/ModelTestOperationValidatorFilter.java
+++ b/model-test/src/main/java/org/jboss/as/model/test/ModelTestOperationValidatorFilter.java
@@ -46,13 +46,23 @@ public class ModelTestOperationValidatorFilter implements Serializable {
     private final List<OperationEntry> entries;
 
     private ModelTestOperationValidatorFilter(List<OperationEntry> entries) {
-        this.entries = entries;
+        this.entries = createStandardEntries(entries);
         validateNone = false;
     }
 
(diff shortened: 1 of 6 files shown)


model-test/src/main/java/org/jboss/as/model/test/ModelTestOperationValidatorFilter.java
```diff
@@ -46,13 +46,23 @@
...
+        entries = validateNone ? null : createStandardEntries(null);
+    }
+
+    private static List<OperationEntry> createStandardEntries(List<OperationEntry> provided) {
+        if (provided == null) {
+            provided = new ArrayList<>();
+        }
+        // Don't check the private internal op that deregisters itself before we get a chance to validate it
+        OperationEntry oe = new OperationEntry(PathAddress.EMPTY_ADDRESS, "boottime-controller-initializer-step", Action.NOCHECK, null);
+        provided.add(oe);
```
bstansberry: An alternative to this is to change this OSH to have it not deregister itself and instead just throw an ISE if invoked post-boot. It's private so it's not visible to end users regardless.

kabir:
@bstansberry I am closing this since I believe it is contained in https://github.com/wildfly/wildfly-core/pull/991

PR #976, opened 2015-08-21
[WFCORE-873] Add attachment to propagate the original address for an

kabir:
aliased operation

https://issues.jboss.org/browse/WFCORE-873

Testing with a composite shows that each step gets the correct alias attachment. However, there were some problems with the alias attachment not being removed, hence the hack having RegistrationAddressResolver (currently only used by r-r-d) removing it.

I think this will do for now, since only the top level resource description gets the address. In the recursive case, children don't get the address.

### DIFF
--- a/controller/src/main/java/org/jboss/as/controller/operations/global/GlobalOperationHandlers.java
+++ b/controller/src/main/java/org/jboss/as/controller/operations/global/GlobalOperationHandlers.java
@@ -60,6 +60,7 @@ import org.jboss.as.controller.registry.AliasEntry;
 import org.jboss.as.controller.registry.ImmutableManagementResourceRegistration;
 import org.jboss.as.controller.registry.ManagementResourceRegistration;
 import org.jboss.as.controller.registry.Resource;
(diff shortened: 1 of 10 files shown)


controller/src/main/java/org/jboss/as/controller/operations/global/GlobalOperationHandlers.java
```diff
@@ -627,17 +628,20 @@ public void handleResult(OperationContext.ResultAction resultAction, OperationCo
         private final ModelNode operation;
         private final ModelNode result;
         private final OperationStepHandler handler; // handler bypassing further wildcard resolution
+        private final boolean checkAlias;
```
bstansberry: This is unused.

bstansberry: I merged it anyway, but please delete it later. No one else will be brave enough to touch this hack in order to delete it. ;)

kabir: oh, I was using it before cleaning up

PR #980, opened 2015-08-21
[WFCORE-900] Be forgiving if the 'dependent' param to OperationContex…

bstansberry:
…t.hasOptionalCapability is null

This also fixes some incorrect, although harmlessly so, use of CapabilityContext.


ctomc:
I am going to enjoy rebasing to this :)

PR #982, opened 2015-08-22
[WFCORE-873] Followup - Reinstate use of check alias

kabir:
This is a follow up commit from Brian's comment at https://github.com/wildfly/wildfly-core/pull/976#discussion-diff-37642395

Although only r-r-d uses the RegistrationAddressResolver, I think the checkAlias parameter/check makes it clear that the hack is happening. Alternatively, I have removed it completely in https://github.com/wildfly/wildfly-core/pull/985. If this one is merged, that one should be closed.


kabir:
Merging the other one

PR #988, opened 2015-08-24
[WFCORE-904] Signal the future when boot fails

bstansberry:


kabir:
@bstansberry I am closing this since I believe it is superseded by https://github.com/wildfly/wildfly-core/pull/991

PR #989, opened 2015-08-24
[WFCORE-903] : Use notifications to help the scanner get a proper status of all deployments between scans

ehsavoie:
When a deployment is deployed or undeployed from out of a scanner, a notification is sent to the scanner so it can update the status.
Using owner instead of notification to send notification

DO NOT MERGE
USED for testing that WFCORE-860 is fixed by this


ehsavoie:
retest this please

ehsavoie:
Creating effective new PR

PR #994, opened 2015-08-25
[WFCORE-910] If an attribute looks extended but does not exist in the…

kabir:
… registry, fall back to the full name

https://issues.jboss.org/browse/WFCORE-910

### DIFF
--- a/controller/src/main/java/org/jboss/as/controller/operations/global/ReadAttributeHandler.java
+++ b/controller/src/main/java/org/jboss/as/controller/operations/global/ReadAttributeHandler.java
@@ -149,24 +149,27 @@ public class ReadAttributeHandler extends GlobalOperationHandlers.AbstractMultiT
 
     private void doExecuteInternal(OperationContext context, ModelNode operation) throws OperationFailedException {
         validator.validate(operation);
+        final ImmutableManagementResourceRegistration registry = context.getResourceRegistration();
         String attributeName = GlobalOperationAttributes.NAME.resolveModelAttribute(context, operation).asString();
         final boolean defaults = GlobalOperationAttributes.INCLUDE_DEFAULTS.resolveModelAttribute(context,operation).asBoolean();
-        final boolean useEnhancedSyntax = containsEnhancedSyntax(attributeName);
-        String attributeExpression = attributeName;
+        boolean useEnhancedSyntax = containsEnhancedSyntax(attributeName);
+        final String fullAttributeName = attributeName;
         if (useEnhancedSyntax){
             attributeName = extractAttributeName(attributeName);
         }
-
-        final ImmutableManagementResourceRegistration registry = context.getResourceRegistration();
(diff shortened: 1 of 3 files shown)


controller/src/main/java/org/jboss/as/controller/operations/global/ReadAttributeHandler.java
```diff
@@ -149,24 +149,27 @@ void doExecute(OperationContext context, ModelNode operation, FilteredData filte
 
     private void doExecuteInternal(OperationContext context, ModelNode operation) throws OperationFailedException {
         validator.validate(operation);
+        final ImmutableManagementResourceRegistration registry = context.getResourceRegistration();
         String attributeName = GlobalOperationAttributes.NAME.resolveModelAttribute(context, operation).asString();
         final boolean defaults = GlobalOperationAttributes.INCLUDE_DEFAULTS.resolveModelAttribute(context,operation).asBoolean();
-        final boolean useEnhancedSyntax = containsEnhancedSyntax(attributeName);
-        String attributeExpression = attributeName;
+        boolean useEnhancedSyntax = containsEnhancedSyntax(attributeName);
```
ctomc: I think cleaner solution would be to pass AttributeAccess to ```containsEnhancedSyntax``` and it would return true based on this new requirement

ctomc: Also with that kind of change you would also address AbstractCollectionHandler that uses EnhancedSyntaxSupport as well. and would need similar change otherwise.

controller/src/test/java/org/jboss/as/controller/operation/global/EnhancedSyntaxTestCase.java
```diff
@@ -133,6 +137,7 @@ protected void revertUpdateToRuntime(OperationContext context, ModelNode operati
                 }).addReadWriteAttribute(COMPLEX_ATTRIBUTE, null, new ModelOnlyWriteAttributeHandler(COMPLEX_ATTRIBUTE))
                 .addReadWriteAttribute(OBJECT_LIST, null, new ModelOnlyWriteAttributeHandler(OBJECT_LIST))
                 .addReadWriteAttribute(COMPLEX_ATTRIBUTE2, null, new ModelOnlyWriteAttributeHandler(COMPLEX_ATTRIBUTE2))
+                .addReadWriteAttribute(NORMAL_LOOKING_EXTENDED, null, new ModelOnlyWriteAttributeHandler(COMPLEX_ATTRIBUTE2))
```
ctomc: should probably be .addReadWriteAttribute(NORMAL_LOOKING_EXTENDED, null, new ModelOnlyWriteAttributeHandler(NORMAL_LOOKING_EXTENDED))

kabir:
Replaced by https://github.com/wildfly/wildfly-core/pull/995

PR #995, opened 2015-08-25
WFCORE-910 If an attribute looks extended but does not exist in the r…

ctomc:
…egistry, fall back to the full name

Just bit reworked version from @kabir 's proposal #993 

https://issues.jboss.org/browse/WFCORE-910


bstansberry:
@kabir please approve

PR #997, opened 2015-08-25
WFCORE-918 Allow embedded server logging to be overridden by properties

luck3y:
@bstansberry Can you have a look at this? I think this covers the cases we discussed yesterday. I had some challanges testing the non-modular version, but I think this works.

Thoughts?

### DIFF
--- a/cli/src/main/java/org/jboss/as/cli/embedded/EmbedHostControllerHandler.java
+++ b/cli/src/main/java/org/jboss/as/cli/embedded/EmbedHostControllerHandler.java
@@ -155,7 +155,7 @@ class EmbedHostControllerHandler extends CommandHandlerWithHelp {
 
             // Create our own LogContext
             final LogContext embeddedLogContext = LogContext.create();
-            // Set up logging from standalone/configuration/logging.properties
+            // Set up logging from domain/configuration/logging.properties
             configureLogContext(embeddedLogContext, jbossHome, ctx);
 
             Contexts localContexts = new Contexts(embeddedLogContext, discardStdoutContext);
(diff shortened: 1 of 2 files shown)


cli/src/main/java/org/jboss/as/cli/embedded/EmbedHostControllerHandler.java
```diff
@@ -258,9 +258,14 @@ public void cliEvent(CliEvent event, CommandContext ctx) {
     }
 
     private void configureLogContext(LogContext embeddedLogContext, File jbossHome, CommandContext ctx) {
-        File standaloneDir =  new File(jbossHome, "domain");
-        File configDir =  new File(standaloneDir, "configuration");
-        File logDir =  new File(standaloneDir, "log");
+
+        final String domainBaseDir = WildFlySecurityManager.getPropertyPrivileged("jboss.server.base.dir", jbossHome.getAbsolutePath() + File.separator + "domain");
+        final String domainConfigDir = WildFlySecurityManager.getPropertyPrivileged("jboss.server.configuration.dir", "configuration");
+        final String domainLogDir = WildFlySecurityManager.getPropertyPrivileged("jboss.server.log.dir", "log");
```
jamezp: The `domainConfigDir` and `domainLogDir` are never used. Depending on what we're doing too all these properties should probably be `jboss.domain.*`. The `domainBaseDir` and `domainConfigDir` definitely should be.

cli/src/main/java/org/jboss/as/cli/embedded/EmbedServerHandler.java
```diff
@@ -267,9 +267,14 @@ public void cliEvent(CliEvent event, CommandContext ctx) {
...
-        File configDir =  new File(standaloneDir, "configuration");
-        File logDir =  new File(standaloneDir, "log");
+
+        final String serverBaseDir = WildFlySecurityManager.getPropertyPrivileged("jboss.server.base.dir", jbossHome.getAbsolutePath() + File.separator + "standalone");
+        final String serverConfigDir = WildFlySecurityManager.getPropertyPrivileged("jboss.server.configuration.dir", "configuration");
+        final String serverLogDir = WildFlySecurityManager.getPropertyPrivileged("jboss.server.log.dir", "log");
+
+        File standaloneDir =  new File(serverBaseDir);
+        File configDir =  new File(standaloneDir, serverConfigDir);
+        File logDir =  new File(standaloneDir, serverLogDir);
```
jamezp: This doesn't seem right. If the `jboss.server.log.dir` is set to something like `/var/log/wildfly` this will result in `$JBOSS_HOME/standalone/var/log/wildfly`.

cli/src/main/java/org/jboss/as/cli/embedded/EmbedServerHandler.java
```diff
@@ -267,9 +267,14 @@ public void cliEvent(CliEvent event, CommandContext ctx) {
     }
 
     private void configureLogContext(LogContext embeddedLogContext, File jbossHome, CommandContext ctx) {
-        File standaloneDir =  new File(jbossHome, "standalone");
-        File configDir =  new File(standaloneDir, "configuration");
-        File logDir =  new File(standaloneDir, "log");
+
+        final String serverBaseDir = WildFlySecurityManager.getPropertyPrivileged("jboss.server.base.dir", jbossHome.getAbsolutePath() + File.separator + "standalone");
+        final String serverConfigDir = WildFlySecurityManager.getPropertyPrivileged("jboss.server.configuration.dir", "configuration");
```
jamezp: This should be `jboss.server.config.dir`.

jamezp:
An additional node if they can be used there are property constants in the `org.jboss.as.server.ServerEnvironment` and `org.jboss.as.host.controller.HostControllerEnvironment` to help keep the property names consistent.

luck3y:
Great, thanks for the feedback James, I'll review and update.

bstansberry:
I'm going to close this one to clean up the queue. It's not needed for 2.0; just a nice enhancement that can happen later. Please open a new PR when you get back to this.

PR #998, opened 2015-08-25
WFCORE-693 Do not require the exclusive management lock to run the "kill" or "de…

luck3y:
Don't aquire exclusive lock for server kill / shutdown ops.

### DIFF
--- a/host-controller/src/main/java/org/jboss/as/host/controller/operations/ServerProcessHandlers.java
+++ b/host-controller/src/main/java/org/jboss/as/host/controller/operations/ServerProcessHandlers.java
@@ -22,6 +22,8 @@
 
 package org.jboss.as.host.controller.operations;
 
+import java.util.EnumSet;
+
 import org.jboss.as.controller.OperationContext;
 import org.jboss.as.controller.OperationDefinition;
 import org.jboss.as.controller.OperationFailedException;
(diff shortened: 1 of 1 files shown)


luck3y:
retest this please

host-controller/src/main/java/org/jboss/as/host/controller/operations/ServerProcessHandlers.java
```diff
@@ -68,11 +71,13 @@ public void execute(final OperationContext context, final ModelNode operation) t
...
-                // WFLY-2189 trigger a write-runtime authz check
-                context.getServiceRegistry(true);
-
-                doExecute(serverName);
-                context.completeStep(OperationContext.RollbackHandler.NOOP_ROLLBACK_HANDLER);
+                context.authorize(operation, EnumSet.of(Action.ActionEffect.WRITE_RUNTIME));
+                context.completeStep(new OperationContext.ResultHandler() {
+                    @Override
+                    public void handleResult(OperationContext.ResultAction resultAction, OperationContext context, ModelNode operation) {
+                        doExecute(serverName);
```
bstansberry: Feel free to ignore this comment; I'll mark this "Ready for merge" regardless.

Why the move? I'm just curious.

If this step was part of a composite, this will alter when in the sequence of events the kill/destroy happens; e.g. if the next step was then some config mod that will fail because the server isn't dead yet. But that's a silly edge case, so moving it is fine.

Since this is in a ResultHandler now, the doExecute() call can be limited to the ResultAction.KEEP case. But this op executing in a composite with something else that will trigger ResultAction.ROLLBACK is also an edge case.

luck3y: Thanks for the feedback. I ended up doing it like this as the other
examples seemed similar. Would this be more correct w.r.t. the cases you
list above:

context.authorize(operation, EnumSet.of(Action.ActionEffect.WRITE_RUNTIME));

context.completeStep(OperationContext.RollbackHandler.NOOP_ROLLBACK_HANDLER);

doExecute(serverName);


?

Thanks,

Ken

PR #999, opened 2015-08-26
[WFCORE-892] Increased timeout for CLI executor to avoid test errors

fbogyai:
Manual tests that are using CustomCLIExecutor are failing on IBM JDK 8 due to low timeouts. I have increased timeout for command execution and changed initial value of exitCode to non-zero for better testability.

JIRA: https://issues.jboss.org/browse/WFCORE-892

### DIFF
--- a/testsuite/shared/src/main/java/org/jboss/as/test/integration/management/util/CustomCLIExecutor.java
+++ b/testsuite/shared/src/main/java/org/jboss/as/test/integration/management/util/CustomCLIExecutor.java
@@ -62,8 +62,9 @@ public class CustomCLIExecutor {
             + MANAGEMENT_HTTPS_PORT;
 
     private static Logger LOGGER = Logger.getLogger(CustomCLIExecutor.class);
-    private static final int CLI_PROC_TIMEOUT = 5000;
-    private static final int STATUS_CHECK_INTERVAL = 2000;
+
+    private static final int CLI_PROC_TIMEOUT = TimeoutUtil.adjust(30000);
+    private static final int STATUS_CHECK_INTERVAL = TimeoutUtil.adjust(2000);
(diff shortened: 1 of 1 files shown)


testsuite/shared/src/main/java/org/jboss/as/test/integration/management/util/CustomCLIExecutor.java
```diff
@@ -62,8 +62,9 @@
             + MANAGEMENT_HTTPS_PORT;
 
     private static Logger LOGGER = Logger.getLogger(CustomCLIExecutor.class);
-    private static final int CLI_PROC_TIMEOUT = 5000;
-    private static final int STATUS_CHECK_INTERVAL = 2000;
+
+    private static final int CLI_PROC_TIMEOUT = TimeoutUtil.adjust(30000);
+    private static final int STATUS_CHECK_INTERVAL = TimeoutUtil.adjust(2000);
```
bstansberry: A 2000 ms check interval is already a long time. Unless what happens after the interval somehow influences the test result, these intervals should be quite short (200 ms or less) so tests execute as rapidly as possible, and they don't need to be adjusted. The only case where having this one be longer might matter is if the "failureResponse" param is passed to "execute". If it's not there's no reason not to make the test execute as fast as possible. AFAICT that param is never used. (I think any test that used it would be unstable, since it would mean the test was counting on some effect from writing the to CLI after some unstable amount of time.)

bstansberry: Apologies; my IDE search was inaccurate. The "failureResponse" response is used in one test.

bstansberry:
I think the check interval handling should be improved to speed these tests, but I'm not going to hold this up over it.

PR #1000, opened 2015-08-26
WFCORE-831 Revisit metrics handling

kabir:
https://issues.jboss.org/browse/WFCORE-831

This will fail on the core full integration job until https://github.com/wildfly/wildfly/pull/7991 has been merged

### DIFF
--- a/controller/src/main/java/org/jboss/as/controller/AbstractAttributeDefinitionBuilder.java
+++ b/controller/src/main/java/org/jboss/as/controller/AbstractAttributeDefinitionBuilder.java
@@ -69,6 +69,7 @@ public abstract class AbstractAttributeDefinitionBuilder<BUILDER extends Abstrac
     protected AttributeParser parser;
     protected String attributeGroup;
     protected CapabilityReferenceRecorder referenceRecorder;
+    private ModelNode undefinedMetricValue;
 
     /**
      * Creates a builder for an attribute with the give name and type. Equivalent to
@@ -661,6 +662,19 @@ public abstract class AbstractAttributeDefinitionBuilder<BUILDER extends Abstrac
         return (BUILDER)this;
     }
 
+    /**
+     * Sets a {@link AttributeDefinition#getUndefinedMetricValue()  default value} to use for the
+     * metric if no runtime value is available (e.g. we are a server running in admin-only mode).
+     *
+     * @param undefinedMetricValue the default value, or {@code null} if no default should be used
+     * @return a builder that can be used to continue building the attribute definition
+     */
+    public BUILDER setUndefinedMetricValue(ModelNode undefinedMetricValue) {
+        this.undefinedMetricValue = (undefinedMetricValue == null || !undefinedMetricValue.isDefined()) ? null : undefinedMetricValue;
+        return (BUILDER) this;
+    }
+
+
     public String getName() {
         return name;
     }
@@ -741,6 +755,10 @@ public abstract class AbstractAttributeDefinitionBuilder<BUILDER extends Abstrac
         return nullSignificant;
     }
 
+    public ModelNode getUndefinedMetricValue() {
+        return undefinedMetricValue;
+    }
+
     /**
      * @deprecated Use {@link #getNullSignificant()}.
      */
@@ -787,4 +805,5 @@ public abstract class AbstractAttributeDefinitionBuilder<BUILDER extends Abstrac
         System.arraycopy(toCopy, 0, result, 0, toCopy.length);
         return result;
     }
+
 }
--- a/controller/src/main/java/org/jboss/as/controller/AttributeDefinition.java
+++ b/controller/src/main/java/org/jboss/as/controller/AttributeDefinition.java
@@ -30,6 +30,7 @@ import java.util.Locale;
 import java.util.Objects;
 import java.util.ResourceBundle;
 import java.util.Set;
+
 import javax.xml.stream.XMLStreamException;
 import javax.xml.stream.XMLStreamWriter;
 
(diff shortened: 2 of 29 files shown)


controller/src/main/java/org/jboss/as/controller/SimpleAttributeDefinition.java
```diff
@@ -61,10 +60,8 @@
      * @param allowNull {@code true} if {@link org.jboss.dmr.ModelType#UNDEFINED} is a valid type for the value
      */
     public SimpleAttributeDefinition(final String name, final ModelType type, final boolean allowNull) {
-        super(name, name, null, type, allowNull, false, (MeasurementUnit) null,
-                (ParameterCorrector) null, (ParameterValidator) null, true, (String[]) null, (String[]) null,
-                (AttributeMarshaller) null, false, (DeprecationData) null, (AccessConstraintDefinition[]) null,
-                (Boolean) null, (AttributeParser) null);
+        super(name, name, null, type, allowNull, false, null, null, null, true, null, null,
+                null, false, null, null, null, null, null);
```
bstansberry: The casts were there as a lame form of code comment because I was getting lost trying to figure out the param mapping. :) I figured the compiler would eliminate them anyway.

This would have been better I suppose:

super(name, name, null, type, allowNull, false, /*MeasurementUnit*/ null,

Anyway, I don't mind them going away.

controller/src/main/java/org/jboss/as/controller/AttributeDefinition.java
```diff
@@ -108,12 +111,57 @@ protected AttributeDefinition(String name, String xmlName, final ModelNode defau
                                   final ParameterCorrector valueCorrector, final ParameterValidator validator,
                                   boolean validateNull, final String[] alternatives, final String[] requires, AttributeMarshaller attributeMarshaller,
                                   boolean resourceOnly, DeprecationData deprecationData, final AccessConstraintDefinition[] accessConstraints,
-                                  Boolean nilSignificant, AttributeParser parser, final AttributeAccess.Flag... flags) {
+                                  Boolean nilSignificant, AttributeParser parser, final ModelNode undefinedMetricValue, final AttributeAccess.Flag... flags) {
```
bstansberry: This is a breaking API change. This multi-param protected method only exists to not break any external code that might have used it or the subclass constructors that use it, so it shouldn't be changed. Any code that wants to use an undefinedMetricValue should use the builder API.

Dropping this param lets you not change List/Map/SimpleAttributeDefinition.

platform-mbean/src/main/java/org/jboss/as/platform/mbean/BufferPoolResourceDefinition.java
```diff
@@ -45,16 +45,16 @@
 class BufferPoolResourceDefinition extends SimpleResourceDefinition {
 
 
-    private static AttributeDefinition MEMORY_USED_NAME = SimpleAttributeDefinitionBuilder.create(PlatformMBeanConstants.MEMORY_USED_NAME, ModelType.LONG, false)
+    private static AttributeDefinition MEMORY_USED_NAME = SimpleAttributeDefinitionBuilder.create(PlatformMBeanConstants.MEMORY_USED_NAME, ModelType.LONG, true)
```
bstansberry: Why are the platform mbean attributes being changed? I don't think there's a situation where these will be undefined.

java.lang.management.BufferPoolMXBean won't return null from any of its methods.

controller/src/main/java/org/jboss/as/controller/AttributeDefinition.java
```diff
@@ -108,12 +111,57 @@ protected AttributeDefinition(String name, String xmlName, final ModelNode defau
...
+        this.resourceOnly = resourceOnly;
+        this.accessConstraints = accessConstraints;
+        this.deprecationData = deprecationData;
+        this.nilSignificant = nilSignificant;
+        this.attributeGroup = attributeGroup;
+        this.allowedValues = allowedValues;
+        this.undefinedMetricValue = new ModelNode();
+        if (undefinedMetricValue != null) {
+            this.undefinedMetricValue.set(undefinedMetricValue);
+        }
```
bstansberry: getUndefinedMetricValue() is going to return null if the constructor param is null or undefined, so in that case we can skip constructing a ModelNode and just store null. Same applies for the defaultValue treatment above (which I wrote.)

I've been working a managment perf support case, so I have perf on my mind today. ;)

threads/src/main/java/org/jboss/as/threads/PoolAttributeDefinitions.java
```diff
@@ -78,11 +78,11 @@
...
-    AttributeDefinition TASK_COUNT = new SimpleAttributeDefinition(CommonAttributes.TASK_COUNT, ModelType.INT, false);
-    AttributeDefinition QUEUE_SIZE = new SimpleAttributeDefinition(CommonAttributes.QUEUE_SIZE, ModelType.INT, false);
+    AttributeDefinition CURRENT_THREAD_COUNT = new SimpleAttributeDefinition(CommonAttributes.CURRENT_THREAD_COUNT, ModelType.INT, true);
+    AttributeDefinition LARGEST_THREAD_COUNT = new SimpleAttributeDefinition(CommonAttributes.LARGEST_THREAD_COUNT, ModelType.INT, true);
+    AttributeDefinition REJECTED_COUNT = new SimpleAttributeDefinition(CommonAttributes.REJECTED_COUNT, ModelType.INT, true);
+    AttributeDefinition ACTIVE_COUNT = new SimpleAttributeDefinition(CommonAttributes.ACTIVE_COUNT, ModelType.INT, true);
+    AttributeDefinition COMPLETED_TASK_COUNT = new SimpleAttributeDefinition(CommonAttributes.COMPLETED_TASK_COUNT, ModelType.INT, true);
+    AttributeDefinition TASK_COUNT = new SimpleAttributeDefinition(CommonAttributes.TASK_COUNT, ModelType.INT, true);
+    AttributeDefinition QUEUE_SIZE = new SimpleAttributeDefinition(CommonAttributes.QUEUE_SIZE, ModelType.INT, true);
```
bstansberry: Can't all of these have an undefinedMetricValue of 0?

controller/src/main/java/org/jboss/as/controller/extension/ExtensionRegistry.java
```diff
@@ -527,7 +527,7 @@ public RunningMode getRunningMode() {
 
         @Override
         public boolean isRuntimeOnlyRegistrationValid() {
-            return processType.isServer() && runningModeControl.getRunningMode() != RunningMode.ADMIN_ONLY;
+            return processType.isServer();
```
bstansberry: What about Host Controller extensions?

controller/src/main/java/org/jboss/as/controller/AbstractAttributeDefinitionBuilder.java
```diff
@@ -661,6 +662,19 @@ public BUILDER setCapabilityReference(CapabilityReferenceRecorder referenceRecor
         return (BUILDER)this;
     }
 
+    /**
+     * Sets a {@link AttributeDefinition#getUndefinedMetricValue()  default value} to use for the
+     * metric/runtime-attribute if no runtime value is available (e.g. we are a server running in admin-only mode.
```
bstansberry: We aren't using this for non-metrics at this point, so I don't think the javadoc so refer to anything but metrics. If we want to use it for non-metrics we should change the method name, e.g. getUndefinedRuntimeValue() or something.

bstansberry:
Superceded by #1011

PR #1002, opened 2015-08-27
Perf improvements in large configs

bstansberry:
https://issues.jboss.org/browse/WFCORE-921
https://issues.jboss.org/browse/WFCORE-922

The first is a big deal in a domain, cutting the time to add 8000 datasources to a config from many 10s of minutes to less than a minute.

### DIFF
--- a/host-controller/src/main/java/org/jboss/as/domain/controller/operations/coordination/ServerOperationResolver.java
+++ b/host-controller/src/main/java/org/jboss/as/domain/controller/operations/coordination/ServerOperationResolver.java
@@ -99,6 +99,7 @@ import org.jboss.dmr.Property;
 public class ServerOperationResolver {
 
     public static final AttachmentKey<Set<ModelNode>> DONT_PROPAGATE_TO_SERVERS_ATTACHMENT = AttachmentKey.create(Set.class);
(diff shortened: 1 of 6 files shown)


host-controller/src/main/java/org/jboss/as/domain/controller/operations/coordination/ServerOperationResolver.java
```diff
@@ -287,6 +288,15 @@ private static ModelNode cleanOpForDontPropagate(ModelNode op) {
         }
     }
 
+    private static ModelNode getDomainModel(OperationContext context) {
+        ModelNode model = context.getAttachment(DOMAIN_MODEL_ATTACHMENT);
+        if (model == null) {
+            model = Resource.Tools.readModel(context.readResourceFromRoot(PathAddress.EMPTY_ADDRESS, true));
+            context.attach(DOMAIN_MODEL_ATTACHMENT, model);
```
jamezp: Should this do `model.protect()` to be safe it won't be modified?

jamezp: Actually maybe it doesn't matter since it's only used here.

bstansberry: Yeah, it's only used here.

It's a clone of the original model, so the only risk is this class modifying it and messing itself up.

PR #1003, opened 2015-08-27
Remove depecated methods from ManagementResourceRegistration

ctomc:
This methods ware deprecated in 7.x times. It is time to remove them.


controller/src/main/java/org/jboss/as/controller/OperationDefinition.java
```diff
@@ -64,6 +65,7 @@ protected OperationDefinition(String name,
                                final DeprecationData deprecationData,
                                AttributeDefinition[] replyParameters,
                                AttributeDefinition[] parameters,
+                               final DescriptionProvider descriptionProvider,
                                AccessConstraintDefinition... accessConstraints
```
bstansberry: This and the subclass constructors are public API, so I prefer not to change their signatures. I'd rather deprecate them and move to passing the builder as the param. See #1009.

controller/src/main/java/org/jboss/as/controller/SimpleOperationDefinition.java
```diff
@@ -79,12 +79,12 @@ protected SimpleOperationDefinition(final String name,
                                      final DeprecationData deprecationData,
                                      final AttributeDefinition[] replyParameters,
                                      final AttributeDefinition... parameters) {
-        super(name, entryType, flags, replyType, replyValueType, replyAllowNull, deprecationData, replyParameters, parameters);
+        super(name, entryType, flags, replyType, replyValueType, replyAllowNull, deprecationData, replyParameters, parameters, null);
```
bstansberry: The 'null' is in the wrong position. #1009 fixes that.

bstansberry:
Superceded by #1009.

PR #1004, opened 2015-08-27
[WFCORE-925] Ability to use proxy in MavenUtil

marekkopecky:
WildFly Core testsuite can be run on pure-IPv6 machines. Proxy could be used in that case.

But some tests use MavenUtil class for download some dependencies. And this class can not use proxy server. So these tests fail.

https://issues.jboss.org/browse/WFCORE-925

### DIFF
--- a/model-test/src/main/java/org/jboss/as/model/test/MavenUtil.java
+++ b/model-test/src/main/java/org/jboss/as/model/test/MavenUtil.java
@@ -59,6 +59,7 @@ import org.eclipse.aether.impl.MetadataGeneratorFactory;
 import org.eclipse.aether.impl.VersionRangeResolver;
 import org.eclipse.aether.impl.VersionResolver;
 import org.eclipse.aether.repository.LocalRepository;
+import org.eclipse.aether.repository.Proxy;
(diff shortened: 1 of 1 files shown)


ctomc:
this should use java standard system properties, which I think would't even require any changes in code to make it work.
see http://docs.oracle.com/javase/8/docs/api/java/net/doc-files/net-properties.html

marekkopecky:
@ctomc, MavenUtil class uses Aether library. This library does not use standard system properties like "http.proxyHost", etc. So it is necessary to set proxy directly in code.

I tried to use these properties before I create this pull request ...

marekkopecky:
retest this please

ctomc:
In that case, it would make sense to use same system properties, instead of having special ones only for this class.
maybe it would be worth looking up also System.getEnv("http_proxy") to lookup linux's default http_proxy config.

marekkopecky:
"http.proxyHost", "http.proxyPort", "https.proxyHost", "https.proxyPort" system properties is used.

I'm not sure if System.getEnv("http_proxy") breaks someone's automation ...

ctomc:
:+1:

PR #1006, opened 2015-08-27
Move some transformers code to testsuite as it is only used there

ctomc:


jamezp:
This doesn't even compile.

ctomc:
retest this please

ctomc:
failure unrelated

PR #1023, opened 2015-09-03
[WFCORE-885] [BZ1063649] Fixing audit log creation with world readabl…

istudens:
…e file permissions.

Peter's commit from https://github.com/jbossas/jboss-eap/pull/1320 restructured for the current WFCORE.

https://issues.jboss.org/browse/WFCORE-885
https://issues.jboss.org/browse/JBEAP-721

### DIFF
--- a/controller/src/main/java/org/jboss/as/controller/audit/AbstractFileAuditLogHandler.java
+++ b/controller/src/main/java/org/jboss/as/controller/audit/AbstractFileAuditLogHandler.java
@@ -71,11 +71,7 @@ public abstract class AbstractFileAuditLogHandler extends AuditLogHandler {
             initializeAtStartup(file);
 
             if (!file.exists()) {
-                try {
-                    file.createNewFile();
-                } catch (IOException e){
-                    throw new RuntimeException(e);
-                }
+                createNewFile(file);
             }
 
             this.file = file;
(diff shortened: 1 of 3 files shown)


bstansberry:
@darranl Please approve.

darranl:
Looks fine as in porting the original approach to fix this - however as we are now in Java 8 as minimum we should have additional APIs available to us to control permissions.

istudens:
@darranl  You mean PosixFilePermissions, right? I will rework the PR.

darranl:
At this point this PR should be fine as we have used this approach for a while now.  The file permission approach is something however we should switch to subsequently in all cases where we control file permissions.

istudens:
I can easily update the PR to something like the following snippet if you like:
https://gist.github.com/istudens/32884c81249d001c59b3

ehsavoie:
I'm not sure PosixFilePermissions are available to set on all OS (Windows and Solaris might be failing).

ehsavoie:
@istudens I'm afraid this will not work on Windows :( so you will have to go to ACL there or back to the current solution.

istudens:
Ok, so let's keep this PR as is and then we can prepare a JDK8-like general solution and update the entire codebase with it later.

PR #1027, opened 2015-09-03
[WFCORE-155] : Evaluate robustness of decisions about target versions for transformation.

ehsavoie:
When a target version is not registered : look for any previous version which differs from the unregistered version by micro version value.

Jira: https://issues.jboss.org/browse/WFCORE-155


kabir:
Clashes with other things on master-ignore

kabir:
Merged a rebased version via command line

PR #1039, opened 2015-09-04
CapabilityRegistry management view fixes

bstansberry:
1) WFCORE-961 -- get stable sorting of the data

2) Disable exposing this for WF 10, pending fix of WFCORE-960 and API discussions with the HAL team.


ctomc:
:+1:

PR #1040, opened 2015-09-06
WFCORE-963 Add capability for IO worker thread pool size

n1hility:


io/subsystem/src/main/java/org/wildfly/extension/io/WorkerAdd.java
```diff
@@ -189,7 +221,14 @@ protected void performRuntime(OperationContext context, ModelNode operation, Mod
             }
         }
 
-        final WorkerService workerService = new WorkerService(builder.getMap());
+        OptionMap options = builder.getMap();
+        final ThreadsService threadsService = new ThreadsService(options.get(Options.WORKER_TASK_MAX_THREADS));
```
ctomc: What is idea behind this? just have a capability that provides read-only view of max threads that worker has?

n1hility: Yes. One example I have in mind is to provide the EJB subsystem with the info so that strict max pools can auto-configure to match the sizing computed by the io worker pool.

ctomc: Makes sense, but to make this work EJB subsystem will need to have reference to name of the worker. 
Which is probably not an issue.

n1hility:
I will most likely nix this approach. Just leaving it open a bit longer.

PR #1041, opened 2015-09-07
[WFCORE-949] : Removing http-interface autocompletes but fails.

ehsavoie:
Removing the management interfaces will put the server in reload mode instead of stopping (and removing) those interfaces at runtime.

Jira: https://issues.jboss.org/browse/WFCORE-949

### DIFF
--- a/testsuite/manualmode/src/test/java/org/wildfly/core/test/standalone/mgmt/RemoveManagementInterfaceTestCase.java
+++ b/testsuite/manualmode/src/test/java/org/wildfly/core/test/standalone/mgmt/RemoveManagementInterfaceTestCase.java
@@ -1,0 +1,200 @@
+/*
+ * JBoss, Home of Professional Open Source
+ * Copyright 2014, JBoss Inc., and individual contributors as indicated
+ * by the @authors tag.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.wildfly.core.test.standalone.mgmt;
+
+import static org.hamcrest.CoreMatchers.is;
+import static org.jboss.as.controller.descriptions.ModelDescriptionConstants.FAILED;
+import static org.jboss.as.controller.descriptions.ModelDescriptionConstants.OUTCOME;
+import static org.jboss.as.controller.descriptions.ModelDescriptionConstants.SUCCESS;
+import static org.jboss.as.test.integration.management.util.CustomCLIExecutor.MANAGEMENT_HTTP_PORT;
+import static org.jboss.as.test.integration.management.util.CustomCLIExecutor.MANAGEMENT_NATIVE_PORT;
+import static org.jboss.as.test.integration.management.util.ModelUtil.createOpNode;
+import static org.junit.Assert.assertThat;
+
+import java.io.IOException;
+import java.net.InetAddress;
+import java.net.UnknownHostException;
+
+import javax.inject.Inject;
+import org.hamcrest.CoreMatchers;
+
+import org.jboss.as.cli.CommandContext;
+import org.jboss.as.controller.client.ModelControllerClient;
+import org.jboss.as.controller.descriptions.ModelDescriptionConstants;
+import org.jboss.as.test.categories.CommonCriteria;
+import org.jboss.as.test.integration.domain.management.util.DomainTestSupport;
+import org.jboss.as.test.integration.management.util.CLITestUtil;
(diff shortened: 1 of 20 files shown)


testsuite/manualmode/src/test/java/org/wildfly/core/test/standalone/mgmt/RemoveManagementInterfaceTestCase.java
```diff
@@ -0,0 +1,206 @@
...
+public class RemoveManagementInterfaceTestCase {
+
+    public static Logger LOGGER = Logger.getLogger(RemoveManagementInterfaceTestCase.class);
+    @Inject
+    protected static ServerController controller;
+
+    @BeforeClass
+    public static void startAndSetupContainer() throws Exception {
+
+        LOGGER.info("*** starting server");
```
ctomc: trace

testsuite/manualmode/src/test/java/org/wildfly/core/test/standalone/mgmt/RemoveManagementInterfaceTestCase.java
```diff
@@ -0,0 +1,206 @@
...
+        controller.start();
+        ManagementClient managementClient = controller.getClient();
+
+        LOGGER.info("*** will configure server now");
```
ctomc: trace

testsuite/manualmode/src/test/java/org/wildfly/core/test/standalone/mgmt/RemoveManagementInterfaceTestCase.java
```diff
@@ -0,0 +1,206 @@
...
+        serverSetup(managementClient.getControllerClient());
+
+        LOGGER.info("*** reloading server");
```
ctomc: trace

testsuite/manualmode/src/test/java/org/wildfly/core/test/standalone/mgmt/RemoveManagementInterfaceTestCase.java
```diff
@@ -0,0 +1,206 @@
...
+        } catch (IOException ioex) {
+            ioex.printStackTrace();
+        } finally {
+            client.close();
+        }
+    }
+
+    @AfterClass
+    public static void stopContainer() throws Exception {
+        LOGGER.info("*** reseting test configuration");
```
ctomc: trace

testsuite/manualmode/src/test/java/org/wildfly/core/test/standalone/mgmt/RemoveManagementInterfaceTestCase.java
```diff
@@ -0,0 +1,206 @@
...
+        ModelControllerClient client = getNativeModelControllerClient();
+        serverTearDown(client);
+        LOGGER.info("*** stopping container");
```
ctomc: trace

testsuite/manualmode/src/test/java/org/wildfly/core/test/standalone/mgmt/RemoveManagementInterfaceTestCase.java
```diff
@@ -0,0 +1,206 @@
...
+        response = client.execute(operation);
+        if (response.hasDefined(OUTCOME) && FAILED.equals(response.get(OUTCOME).asString())) {
+            // create http interface to control server
+            operation = createOpNode("core-service=management/management-interface=http-interface", ModelDescriptionConstants.ADD);
+            operation.get("security-realm").set("ManagementRealm");
+            operation.get("socket-binding").set("management-http");
+            operation.get("http-upgrade-enabled").set(true);
+            CoreUtils.applyUpdate(operation, client);
+        }
+        LOGGER.info("*** reloading server");
```
ctomc: trace

bstansberry:
These should extend ReloadRequiredRemoveStepHandler, as that class properly handles doing/reverting the setting of reload-required in Stage.RUNTIME. Doing it in performRemove results in it happening prematurely, in Stage.MODEL, and doesn't allow it to be reverted in rollback.

These classes have special behavior in requiresRuntime that needs to be maintained though. Basically, the same evaluation that happens in the matching "add" handler should happen in these classes. If the "add" handler would return 'true' and thus install services in performRuntime, then these remove handlers should return true and thus performRuntime will get called to set reload-required. Otherwise they should return 'false' as there were no services added and thus no need for reload.

There are 2 other classes that also need to be changed:

org.jboss.as.server.operations.NativeRemotingManagementRemoveHandler
org.jboss.as.host.controller.operations.HttpManagementRemoveHandler

Finally, org.jboss.as.host.controller.resources.NativeManagementResourceDefinition isn't even registering a handler for "remove". While that's off-topic to this JIRA, let's chat about it. It made sense to not allow "remove" in EAP 6 as a native interface was absolutely required in order for a slave to talk to the master. But IIRC we no longer have such a requirement, as the slave can use HTTP Upgrade.

bstansberry:
Needs checkstyle corrections.

bstansberry:
Adding the "hold" label as this removal will break the jmx remoting connector (if configured) and that won't be detected until the reload. Correcting that is not a trivial thing.

bstansberry:
I'm going to close this so it's more obvious if you do an update. So please open a new PR if you're able to get back to this.

PR #1042, opened 2015-09-08
Small mgmt api changes / cleanup (multiple commits)

ctomc:


controller/src/main/java/org/jboss/as/controller/logging/ControllerLogger.java
```diff
@@ -1709,20 +1709,6 @@
...
-     *
-     * @param stepOpName the step operation name.
-     * @param address    the address.
-     *
-     * @return the message
-     *
-     * @deprecated use {@link #noSuchResourceType(PathAddress)} or {@link #noHandlerForOperation(String, PathAddress)}
-     */
-//    @Message(id = 148, value = "No handler for %s at address %s")
-//    String noHandler(String stepOpName, PathAddress address);
```
bstansberry: Please don't do this. The commented code clarifies why there's a gap in the id sequence and makes it easy to use git-annotate to figure out what change made the message unnecessary.

ctomc: it confused the compiler as it thinks that  * @deprecated marks the method below it as deprecated.
will change it in a way both things can be satisfied.

bstansberry: Ah, cool. Thanks.

PR #1047, opened 2015-09-09
WFCORE-807 Improve error reporting for missing capabilties

ctomc:
messages like this now:
```
WFLYCTL0369: Required capabilities are not available:
        cap_a in context 'profile=a'; There are no known registration points which can provide this capability.

```
and
```
WFLYCTL0369: Required capabilities are not available:
     cap_a in context 'profile=a'; Possible registration points for this capability:
       /subsystem=bah/
```


bstansberry:
I assume some of the core testsuite failures are related.

ctomc:
@bstansberry yeah, I've changed the message but didn't update tests to follow that.
fixed.

PR #1049, opened 2015-09-10
WFCORE-597 Where an ObjectTypeAttributeDefinition is in use respect t…

ctomc:
…he ResourceOnly setting on contained types for add operations.


controller/src/main/java/org/jboss/as/controller/ObjectTypeAttributeDefinition.java
```diff
@@ -195,9 +194,13 @@ public ModelNode resolveValue(ExpressionResolver resolver, ModelNode value) thro
     }
 
     protected void addValueTypeDescription(final ModelNode node, final String prefix, final ResourceBundle bundle,
```
bstansberry: Since this was protected, please restore and deprecate the old variant, having it pass 'false' into this new one.

PR #1052, opened 2015-09-10
WFLY-5322 Skip CLEANUP phase if we allow restart

pferraro:
... since the associated processors don't restore their state on undeploy

Fixes singleton deployment issue:
https://issues.jboss.org/browse/WFLY-5322

### DIFF
--- a/server/src/main/java/org/jboss/as/server/deployment/DeploymentUnitPhaseService.java
+++ b/server/src/main/java/org/jboss/as/server/deployment/DeploymentUnitPhaseService.java
@@ -155,7 +155,11 @@ final class DeploymentUnitPhaseService<T> implements Service<T> {
             }
         }
 
-        final Phase nextPhase = phase.next();
+        Phase nextPhase = phase.next();
+        // Skip CLEANUP phase if we allow restart, since the processors associated with this phase can't restore their state on undeploy
+        if (allowRestart && (nextPhase == Phase.CLEANUP)) {
+            nextPhase = nextPhase.next();
+        }
         if (nextPhase != null) {
             final ServiceName serviceName = DeploymentUtils.getDeploymentUnitPhaseServiceName(deploymentUnit, nextPhase);
             final DeploymentUnitPhaseService<?> phaseService = DeploymentUnitPhaseService.create(deploymentUnit, nextPhase);


stuartwdouglas:
This is not really sufficient to guarantee restartability. Because DUP's work by modifying mutable data if they are re-ran the modification will essentially be made twice, which can result in problems in some situations. 

https://issues.jboss.org/browse/WFCORE-509

https://github.com/wildfly/wildfly-core/pull/440/files attempted to deal with by forcing a complete restart but there were some issues with clustered singletons.

pferraro:
This PR isn't intended to guarantee restartability.  To your point, DUPs that modify state should use undeploy() to restore any state modified during deploy() - and in that way, be restartable (assuming that the transient state on which they might rely is available).

The change I've proposed here seems harmless enough.  All CLEANUP phase DUPs, by definition, aren't restartable - so we can effectively skip that whole phase, if and only if, we allow deployment restarts (e.g. for singleton deployments).  Also, this does effectively fix the attached issue.

Re: WFCORE-509 Forcing a complete deployment restart isn't going to work for singleton deployments (as they are currently implemented), since a deployment doesn't know that it is is a singleton deployment until the PARSING phase.

n1hility:
> This PR isn't intended to guarantee restartability. To your point, DUPs that modify state should use > undeploy() to restore any state modified during deploy() - and in that way, be restartable (assuming that the transient state on which they might rely is available).

n1hility:
> Re: WFCORE-509 Forcing a complete deployment restart isn't going to work for singleton deployments (as they are currently implemented), since a deployment doesn't know that it is is a singleton deployment until the PARSING phase.

All it attempted to do was trigger a full redeploy when it detected the deployment was restarting, so it seems possible to work with singleton deployments. A singleton deployment is really just a paused initial start, so it seems just a matter of differentiating between an initial start, and a restart.

It might also be possible to roll back to where you as long as it can be validated that all deployers in the phase in hold on to their state, but I'm not sure its worth it. Also enabling that ALLOW_RESTART flag means that bugs will present themselves if part of the deployment bounces. So things like https://issues.jboss.org/browse/WFLY-975 will happen with singleton deployments. 

So really I think we just need to find a way to make WFCORE-509 singleton compatible.

pferraro:
OK - let me close this and use @stuartwdouglas 's branch as a starting point.

PR #1058, opened 2015-09-11
WFCORE-826 List-* operations should accept also objects as value

ctomc:


controller/src/main/java/org/jboss/as/controller/AbstractWriteAttributeHandler.java
```diff
@@ -178,31 +176,6 @@ protected abstract void revertUpdateToRuntime(OperationContext context, ModelNod
...
     /**
-     * Does nothing. Subclasses can alter this behavior.
-     *
-     * @param attributeName the name of the attribute being updated
-     * @param unresolvedValue the unresolved value
-     *
-     * @deprecated the {@link AttributeDefinition} provided to the constructor will validate
-     */
-    @Deprecated
-    protected void validateUnresolvedValue(final String attributeName, final ModelNode unresolvedValue) throws OperationFailedException {
```
ctomc: removing this as they ware deprecated back in march 2013 and not used anywhere anymore.

PR #1062, opened 2015-09-15
Revert PR #1048 and just upgrade aesh

bstansberry:
This is just an experiment to see how CI likes two of the 3 commits in #1061 while Alexey looks into the failures on #1061.


bstansberry:
retest this please

The Full Integration - Windows failure is some utility cleanup noise; the actual testsuite passed.

The Full Integration - Linux failures are what we see when a leftover process from another test is running on the agent, causing port conflicts.

PR #1063, opened 2015-09-15
Profile clone transformer and ignores

kabir:
This currently has no test, I am planning on adding one tomorrow.

When executing /profile-clone=ignored(to-profile=new) in a mixed domain, if 
a) the slave has set ignore-unused-configuration=true (and the ignored profile is not used) the operation goes through the transformer.
b) the slave has not set ignore-unused-configuration=true (or does not support it) the operation should be rejected with a message (this still needs to be improved) saying they need to 
-reload the DC into admin-only mode 
-perform the clone 
-ignore the profile on slaves which cannot support, or are not interested, in the new profile 
-reload the DC into normal mode 
-reload any slaves needing reload

### DIFF
--- a/host-controller/src/main/java/org/jboss/as/domain/controller/operations/ProfileCloneHandler.java
+++ b/host-controller/src/main/java/org/jboss/as/domain/controller/operations/ProfileCloneHandler.java
@@ -44,7 +44,9 @@ import org.jboss.as.controller.PathElement;
 import org.jboss.as.controller.SimpleAttributeDefinitionBuilder;
 import org.jboss.as.controller.SimpleOperationDefinitionBuilder;
 import org.jboss.as.controller.access.management.SensitiveTargetAccessConstraintDefinition;
+import org.jboss.as.domain.controller.LocalHostControllerInfo;
 import org.jboss.as.domain.controller.resources.DomainResolver;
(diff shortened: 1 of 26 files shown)


host-controller/src/main/java/org/jboss/as/domain/controller/operations/ProfileCloneHandler.java
```diff
@@ -53,21 +59,46 @@
...
+                    RemoteDomainControllerAddHandler.IGNORE_UNUSED_CONFIG.resolveModelAttribute(context,
+                            host.getModel().get(
+                                    HostResourceDefinition.DOMAIN_CONTROLLER.getName(),
+                                    HostResourceDefinition.DC_REMOTE.getName())).asBoolean();
+
+            if (ignoreUnusedConfiguration) {
+                //Since the cloned profile is a new one, we don't need to do anything. It is guaranteed to be unused,
+                //and thus should be ignored
+                return;
+            }
```
bstansberry: I'm going to send up a new PR changing this to just call LocalHostControllerInfo.isRemoteDomainControllerIgnoreUnaffectedConfiguration() as that will be necessary for this to work properly once the WFCORE-310 work comes in. There will no longer be a default value for IGNORE_UNUSED_CONFIG; the behavior when undefined will depend on whether --backup is used.

kabir:
Superseded by #1064

PR #1070, opened 2015-09-16
WFCORE-310 - change ignore-unused-configuration default true

luck3y:
- if nothing is specified, then ignore-unused-configuration = true, if --backup is used and host.xml has unspecified value, then  ignore-unused-configuration is false, otherwise the value configured in host.xml is used.
- WFCORE-981 - RemoteDomainConnectionService uses out of date HostInfo data when reconnecting, this is now calculated on-demand, for example during reconnection to the DC.


host-controller/src/main/java/org/jboss/as/host/controller/operations/LocalHostControllerInfoImpl.java
```diff
@@ -153,6 +153,12 @@ public boolean isRemoteDomainControllerIgnoreUnaffectedConfiguration() {
         return remoteIgnoreUnaffectedConfiguration;
     }
 
+    @Override
+    // started with --backup
```
kabir: Nitpick, I'd prefer javadoc or putting this comment inside the method body

server/src/main/resources/schema/wildfly-config_4_0.xsd
```diff
@@ -2337,11 +2337,17 @@
         </xs:attribute>
         <xs:attribute name="security-realm" type="xs:string" use="optional" />
         <xs:attribute name="username" type="xs:string" use="optional" />
-        <xs:attribute name="ignore-unused-configuration" type="xs:boolean" default="false">
+        <xs:attribute name="ignore-unused-configuration" type="xs:boolean" use="optional">
             <xs:annotation>
                 <xs:documentation>
-                    Set to true to instruct the master Host Controller to automatically not forward configuration and
+                    When set to true, this instructs the master Host Controller to not forward configuration and
```
kabir: Although it is quite a mouthful, we should probably put this into the correct LocalDescriptions.properties?

testsuite/domain/src/test/resources/host-configs/host-slave.xml
```diff
@@ -141,6 +141,8 @@
         </server>
         <server name="reload-two" group="reload-test-group"  auto-start="false">
             <jvm name="default" />
+            <!-- this is for ServerRestartRequiredTestCase.testSocketBinding() unless the socket-binding-group is actually used for something it is ignored on the slaves, this is a quick fix before correcting the test -->
+            <socket-bindings socket-binding-group="reload-sockets" port-offset="1550"/>
```
bstansberry: We need to correct this problem as part of this work. Unless my understanding of what is going on with this test is incorrect, this isn't a case of a incorrecdt test, it's the software itself that is not working correctly.

AIUI, in the domain config this test uses (domain-standard.xml), the "reload-sockets" s-b-g is used by the "reload-test-group" server group. Therefore this "reload-two" server-config means the "reload-sockets" is used by this server even without this element.

luck3y: Yep, opened https://issues.jboss.org/plugins/servlet/mobile#issue/WFCORE-982
for this test.

I think we understand the issue though. The sbg is unused on the slave, so
it ignores until its actually used for something. Am I missing something
there in how that should work?

server/src/main/resources/schema/wildfly-config_4_0.xsd
luck3y: That already has a slighty more concise version: 

host.write-remote-domain-controller.ignore-unused-configuration=Set to true to instruct the master Host Controller to automatically not forward configuration and operations for profiles, socket binding groups and server groups which do not affect our servers.

bstansberry: Perhaps add:

"If undefined the default behavior is 'true' unless the process was started with the '--backup' command line parameter used, in which case the default behavior is 'false'.

testsuite/domain/src/test/resources/host-configs/host-slave.xml
bstansberry: Unless I've missed something, the sbg is not unused.

server-config=reload-two points to server-group=reload-test-group which points to socket-binding-group=reload-sockets

That chain is sufficient to establish use of the sbg. (It's analogous to the chain that establishes uses of a profile.)

Adding a socket-bindings child under the server-config is solely to allow overriding of the configuration obtained from the server-group. Its presence is not necessary to establish "use" of an sbg.

bstansberry: That chain is what's there when the domain is first started; i.e. it's what the xml files show. It's possible that it's no longer there by the time this test runs, due to some other change made by a test after the domain starts. But I don't see that being done in ServerRestartRequiredTestCase. I know ServerRestartRequiredTestCase will fail even if it is run by itself with -Dtest (i.e. not part of a larger test suite with other tests that may alter the config).

luck3y:
Thanks for the feedback. Closing this, will reopen with separate commits.

PR #1071, opened 2015-09-16
[WFCORE-393] Better error message for rejection of profile clone

kabir:


host-controller/src/main/java/org/jboss/as/domain/controller/logging/DomainControllerLogger.java
```diff
@@ -741,6 +741,11 @@
     @Message(id = 77, value = "Duplicate included socket binding group '%s'")
     XMLStreamException duplicateSocketBindingGroupInclude(String s);
 
-    @Message(id = 78, value = "The profile clone operation is not available on the host '%s'")
+    @Message(id = 78, value = "The profile clone operation is not available on the host '%s'. To be able to use it in a " +
+            "mixed domain, you need to either: " +
```
bstansberry: "mixed version domain" ?

host-controller/src/main/java/org/jboss/as/domain/controller/logging/DomainControllerLogger.java
```diff
@@ -741,6 +741,11 @@
...
+            "a) Make sure that all older slaves with a model version smaller than 4.0.0 " +
+            "ignore the cloned profile and the profile specified in the 'to-profile' parameter. " +
```
bstansberry: Do we need to require that they ignore the cloned profile? Since we are rejecting, I don't think it matters. The clone op doesn't alter the cloned profile.

kabir: If the user does:

    /profile=original:clone(to-profile=new)

They will get the message. Now if they ignore just the 'original' profile on the slave the clone will work

    /profile=original:clone(to-profile=new)
    
But if they now try to do something on the 'new' profile it will fail since there is no 'new' profile on the slave, i.e. the slave is out of sync until it is reloaded. Ignoring 'new' on the slaves is one workaround, reloading the affected slaves is another.

PR #1072, opened 2015-09-16
https://issues.jboss.org/browse/WFCORE-876 Reload or Shutdown inside …

aloubyansky:
…IF statement is performed before the if/else block batch is executed

(this commit removes the logic starting batches automatically for every control flow block of if-else and try-catch-finally)

Thanks,
Alexey


bstansberry:
@Gonthim please have a look.

Gonthim:
Looks good to me

PR #1073, opened 2015-09-16
[WFCORE-980]:  The capability registry fails to rollback properly

ehsavoie:
The issue is not in the rollback itself but in the publish where we don't deep copy the registry content into the publishedRegistry thus sharing references afterwards.

Jira: https://issues.jboss.org/browse/WFCORE-980


ctomc:
:+1:

PR #1075, opened 2015-09-17
Validators master

soul2zimate:
For some subsystem properties values, it's legal to set -1 (means no limit) or a positive value, but zero is not a valid value. For example "thread-pool-max-size". It's helpful to add InfiniteOrPositiveValidators from messaging subsystem to wildfly-core and share it for different subsystems.

Also,  set IntRangeValidator minimum value -1 to several timeout values.


controller/src/main/java/org/jboss/as/controller/operations/validation/InfiniteOrPositiveValidators.java
```diff
@@ -0,0 +1,59 @@
...
+import org.jboss.as.controller.logging.ControllerLogger;
+import org.jboss.dmr.ModelNode;
+import org.jboss.dmr.ModelType;
+
+/**
+ * Validates a given parameter is a legal value (-1 or > 0) where -1 means "forever"
+ *
+ * @author Jeff Mesnil (c) 2012 Red Hat Inc.
+ */
+public interface InfiniteOrPositiveValidators {
```
bstansberry: Please make this a final class with a no-op private constructor. A constant interface is an anti-pattern

bstansberry:
Besides the constant interface issue I noted, the general concept of such a validator is ok. But use of this is to be avoided in the WildFly management API. (Note: I'm not suggesting changing existing API -- API compatibility outweighs this.)

The only valid use case for this is if the attribute has no default value or a default value other than -1. The "no default value" case is likely a poor API, since why shouldn't there be a default value? The default value other than -1 case is valid, because in that case a user-configured value of 'undefined' means whatever that default is.

A value of -1 should not mean "infinite". Either undefined or Integer.MAX_VALUE suffice for that.

host-controller/src/main/java/org/jboss/as/domain/controller/operations/DomainServerLifecycleHandlers.java
```diff
@@ -72,6 +73,7 @@
             .build();
 
     private static final AttributeDefinition TIMEOUT = SimpleAttributeDefinitionBuilder.create(ModelDescriptionConstants.TIMEOUT, ModelType.INT, true)
+            .setValidator(new IntRangeValidator(-1))
```
bstansberry: This is incorrect. The value of zero is legal. It's the default.

I see no reason to make an API change to disallow values < -1. If this were a new API I'd agree that limiting to -1 is better, but this API was part of WildFly 9.

bstansberry:
The second commit is not ok, for the reasons explained in my comment on DomainServerLifecycleHandlers.

PR #1078, opened 2015-09-17
https://issues.jboss.org/browse/WFCORE-970 Deploying a CLI archive th…

aloubyansky:
…at contains 'module add' commands results in 'WFLYCTL0155: steps may not be null' when using the CLI GUI

This commit instead of sending an empty composite request and displaying the error, logs a warning explaining that the request was not sent because there were no server-side operations to execute.

@ssilvert could you please review this too?

Thanks,
Alexey


ssilvert:
The fix looks good.  I kind of wonder why we had an empty composite request, but that's unrelated to the patch.

aloubyansky:
Because the script included into the archive does not contain any operation request (or command that translates into an operation request) but module add instructions, those are local file manipulations.

Thanks!

aloubyansky:
retest this please

PR #1080, opened 2015-09-18
WFCORE-981 - RemoteDomainConnectionService uses out of date HostInfo

luck3y:


luck3y:
Going to just add as a commit to WFCORE-310, closing

PR #1082, opened 2015-09-18
[WFCORE-939] Use appropriate launch-type value for the embedded-host-controller

jamezp:
@bstansberry please let me know if this isn't what you had in mind. This just keeps the change contained to the host-controller.


bstansberry:
How about adding a getProcessType() to HostControllerEnvironment? The type of the process is a legit part of the environmental info. That would require a param added to Main.determineEnvironment, but would allow us to get rid of the 'embedded' param to the HostControllerService constructor.

jamezp:
That's a good idea. I think that makes sense.

PR #1085, opened 2015-09-20
[WFCORE-393] Use a runtime registry on the slave to track if an ignored

kabir:
profile is being cloned. In this case, the clone operation does not get
executed on the slave. Track this, and ignore any subsequent operations
on that profile on the slave, and put the slave in the requires-reload
state.

### DIFF
--- a/host-controller/src/main/java/org/jboss/as/domain/controller/operations/coordination/HostControllerExecutionSupport.java
+++ b/host-controller/src/main/java/org/jboss/as/domain/controller/operations/coordination/HostControllerExecutionSupport.java
@@ -44,6 +44,7 @@ import java.util.List;
 import java.util.Map;
 import java.util.Set;
 
+import org.jboss.as.controller.OperationContext;
 import org.jboss.as.controller.OperationFailedException;
 import org.jboss.as.controller.PathAddress;
 import org.jboss.as.controller.PathElement;
@@ -90,6 +91,21 @@ interface HostControllerExecutionSupport {
     ModelNode getFormattedDomainResult(ModelNode resultNode);
 
     /**
+     * Returns whether the operation puts the host in the reload-required state.
+     *
+     * @return {@code true} if the operation puts the host in the reload-required state
+     */
+    boolean isReloadRequired();
+
+    /**
+     * Callback for then the controller transaction has completed.
+     *
+     * @param rollback Whether the transaction rolled back or not. {@code true} indicates it was rolled back; {@code false}
+     *                 indicates it was committed.
+     */
+    void complete(boolean rollback);
+
+    /**
      * Provider of server level operations necessary to effect a given domain or host level operation on the servers
      * managed by this host controller.
      */
@@ -121,6 +137,8 @@ interface HostControllerExecutionSupport {
         /**
          * Create a HostControllerExecutionSupport for a given operation.
          *
+         *
+         * @param context
          * @param operation the operation
          * @param hostName the name of the host executing the operation
          * @param domainModelProvider source for the domain model
@@ -129,7 +147,7 @@ interface HostControllerExecutionSupport {
          *
          * @return the HostControllerExecutionSupport
          */
-        public static HostControllerExecutionSupport create(final ModelNode operation,
+        public static HostControllerExecutionSupport create(OperationContext context, final ModelNode operation,
                                                             final String hostName,
                                                             final DomainModelProvider domainModelProvider,
                                                             final IgnoredDomainResourceRegistry ignoredDomainResourceRegistry,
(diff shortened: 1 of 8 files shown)


testsuite/domain/src/test/java/org/jboss/as/test/integration/domain/suites/IgnoredResourcesProfileCloneTestCase.java
```diff
@@ -0,0 +1,317 @@
...
+                    Util.getReadAttributeOperation(SLAVE_ADDR, DOMAIN_CONTROLLER),
+                    masterClient).get(REMOTE);
+            originalSlaveDc.protect();
+            Assert.assertFalse(originalSlaveDc.hasDefined(IGNORE_UNUSED_CONFIG));
+            try {
+                //TODO Remove this - Once the default for ignore-unused-configuration is true, we don't need to set it to true
+                ModelNode newSlaveDc = originalSlaveDc.clone();
+                newSlaveDc.get(IGNORE_UNUSED_CONFIG).set(true);
+                writeSlaveDomainController(slaveClient, newSlaveDc);
+                reloadSlave(slaveLifecycleUtil);
```
bstansberry: I plan to ensure Ken's work is merged before dealing with this, as it's more critical, so you should assume you'll remove this before this PR gets merged.

host-controller/src/main/java/org/jboss/as/host/controller/ignored/IgnoredDomainResourceRegistry.java
```diff
@@ -106,4 +123,75 @@ public void registerChildren(ManagementResourceRegistration resourceRegistration
             resourceRegistration.registerSubModel(new IgnoredDomainTypeResourceDefinition());
         }
     }
+
+    public class IgnoredClonedProfileRegistry {
```
bstansberry: Needs javadoc.

host-controller/src/main/java/org/jboss/as/host/controller/ignored/IgnoredDomainResourceRegistry.java
```diff
@@ -106,4 +123,75 @@ public void registerChildren(ManagementResourceRegistration resourceRegistration
...
+        private volatile Set<String> ignoredClonedProfiles = Collections.synchronizedSet(new HashSet<>());
+        private volatile Set<String> currentTxIgnoredClonedProfiles;
+        private volatile boolean reloadRequired;
+
+        private IgnoredClonedProfileRegistry() {
+        }
+
+        public boolean checkIgnoredProfileClone(OperationContext context, ModelNode operation) {
```
bstansberry: Needs javadoc.

host-controller/src/main/java/org/jboss/as/host/controller/ignored/IgnoredDomainResourceRegistry.java
```diff
@@ -106,4 +123,75 @@ public void registerChildren(ManagementResourceRegistration resourceRegistration
...
+                        }
+                        //Ignore depending on if it is in the runtime registry
+                        return getIgnoredClonedProfiles(false).contains(first.getValue());
+                    }
+                }
+            }
+            return false;
+        }
+
+        public void initializeModelSync() {
```
bstansberry: Needs javadoc.

host-controller/src/main/java/org/jboss/as/host/controller/ignored/IgnoredDomainResourceRegistry.java
```diff
@@ -106,4 +123,75 @@ public void registerChildren(ManagementResourceRegistration resourceRegistration
...
+            if (!localHostControllerInfo.isMasterDomainController()) {
+                //We are resyncing the model, so clear the runtime registry
+                getIgnoredClonedProfiles(true).clear();
+            }
+        }
+
+        public void complete(OperationContext context, boolean rollback) {
```
bstansberry: Needs javadoc.

host-controller/src/main/java/org/jboss/as/host/controller/ignored/IgnoredDomainResourceRegistry.java
```diff
@@ -106,4 +123,75 @@ public void registerChildren(ManagementResourceRegistration resourceRegistration
...
+            if (!localHostControllerInfo.isMasterDomainController()) {
+                if (!rollback) {
+                    if (currentTxIgnoredClonedProfiles != null) {
+                        ignoredClonedProfiles = currentTxIgnoredClonedProfiles;
+                    }
+                } else {
+                    if (reloadRequired) {
+                        context.revertReloadRequired();
```
bstansberry: This won't work. There are two steps that trigger a call to this method: SyncDomainModelOperationHandler and an inner class of OperationSlaveStepHandler. Neither are the steps that will trigger the context.reloadRequired() above. Therefore AbstractOperationContext.revertReloadRequired() will ignore their attempt to revert the reload required. (This is all part of ensuring that if multiple steps in a composite record reload-required, it's not reverted unless all of them revert it.)

host-controller/src/main/java/org/jboss/as/domain/controller/operations/coordination/HostControllerExecutionSupport.java
```diff
@@ -121,6 +124,8 @@
         /**
          * Create a HostControllerExecutionSupport for a given operation.
          *
+         *
+         * @param context
```
bstansberry: (nit) javadoc

host-controller/src/main/java/org/jboss/as/domain/controller/operations/coordination/OperationSlaveStepHandler.java
```diff
@@ -135,6 +135,19 @@ HostControllerExecutionSupport addSteps(final OperationContext context, final Mo
...
         context.addStep(sorh, OperationContext.Stage.DOMAIN);
 
+        context.addStep(new OperationStepHandler() {
+            @Override
+            public void execute(OperationContext context, ModelNode operation) throws OperationFailedException {
+
+                context.completeStep(new OperationContext.ResultHandler() {
+                    @Override
+                    public void handleResult(OperationContext.ResultAction resultAction, OperationContext context, ModelNode operation) {
+                        hostControllerExecutionSupport.complete(context, resultAction == OperationContext.ResultAction.ROLLBACK);
```
bstansberry: I believe this can be moved to the existing ResultHandler passed to completeStep by OperationSlaveStepHandler's execute method.

bstansberry:
The following comment doesn't need to block this, but I'd like a JIRA filed to look into it ASAP (before we forget the details.)

The stuff with IgnoredDomainResourceRegistry calling context.reloadRequired() / revertReloadRequired() makes me very uncomfortable. It's too far removed from the step handlers and is a potential source of bugs, or just plain old head scratching from future maintainers. It also results in passing around the context through different classes, opening up the chances that someone in the future will misuse it. (See WFCORE-988 for an example of where that can lead.)

Perhaps HostControllerExecutionSupport itself can expose a getter that tells OperationSlaveStepHandler whether it needs to set reloadRequired. IgnoredClonedProfileRegistry.checkIgnoredProfileClone would need to return some kind of simple object instead of just a boolean, to encapsulate both the "ignore it?" and "reload-required?" values.

bstansberry:
BTW, thanks for this. I'm in that review mode where I go find small problems, so I forget to say the overall idea is a good one. :)

testsuite/domain/src/test/java/org/jboss/as/test/integration/domain/suites/IgnoredResourcesProfileCloneTestCase.java
kabir: This should still work with Ken's work merged, it is more a cleanup task

kabir:
Superseded by https://github.com/wildfly/wildfly-core/pull/1091

PR #1086, opened 2015-09-20
Fix problems revealed when domain testsuite runs with assertions enabled.

bstansberry:
This PR does two main things:

1) Turns on assertions in the build.
2) Fixes domain mode issues that are revealed thereby.

We can skip part 1) for now if it's considered too risky for CR2. But I wanted it on for now to get CI tests of this work with assertions now. I can back out those commits before merging if that's the decision.

The details on 2) are:

a) Fix WFCORE-988
b) Fix WFCORE-989
c) Fix an issue where in some situations request tracking wasn't recording that an operation had been 'prepared' even though it had, plus some related code cleanup to help make the intended behavior clearer.

### DIFF
--- a/testsuite/shared/src/main/java/org/jboss/as/test/integration/domain/management/util/DomainLifecycleUtil.java
+++ b/testsuite/shared/src/main/java/org/jboss/as/test/integration/domain/management/util/DomainLifecycleUtil.java
@@ -149,6 +149,13 @@ public class DomainLifecycleUtil {
                         .addProcessControllerJavaOptions(javaOpts);
             }
 
+            if (configuration.isEnableAssertions()) {
+                commandBuilder.addProcessControllerJavaOption("-ea");
+                commandBuilder.addHostControllerJavaOption("-ea");
+                // This doesn't work; the HC Main class treats this as an invalid param
+                //commandBuilder.addServerArgument("-ea");
+            }
+
             final String jbossArgs = System.getProperty("jboss.domain.server.args");
             if (jbossArgs != null) {
                 commandBuilder.addServerArguments(jbossArgs.split("\\s+"));
--- a/controller/src/main/java/org/jboss/as/controller/ModelControllerImpl.java
+++ b/controller/src/main/java/org/jboss/as/controller/ModelControllerImpl.java
@@ -388,6 +388,7 @@ class ModelControllerImpl implements ModelController {
                     }
 
                     context.addStep(responseNode, operation, prepareStep, OperationContext.Stage.MODEL);
+                    ControllerLogger.MGMT_OP_LOGGER.tracef("Executing %s", operation);
                     context.executeOperation();
                     responseStreams = context.getResponseStreams();
                 } finally {
--- a/controller/src/main/java/org/jboss/as/controller/TransformingProxyController.java
+++ b/controller/src/main/java/org/jboss/as/controller/TransformingProxyController.java
@@ -69,13 +69,45 @@ public interface TransformingProxyController extends ProxyController {
      */
     OperationTransformer.TransformedOperation transformOperation(OperationContext context, ModelNode operation) throws OperationFailedException;
 
-    public static class Factory {
+    /**
+     * Transform the operation.
+     *
+     * @param parameters parameters that drive the transformation
+     * @param operation the operation to transform.
+     * @return the transformed operation
+     * @throws OperationFailedException
+     */
+    OperationTransformer.TransformedOperation transformOperation(Transformers.TransformationInputs parameters, ModelNode operation) throws OperationFailedException;
 
+
+    /**
+     * Factory methods for creating a {@link TransformingProxyController}
+     */
+    class Factory {
+
+        /**
+         * Creates a {@link TransactionalProtocolClient} based on the given {@code channelAssociation} and then
+         * uses that to create a {@link TransformingProxyController}.
+         *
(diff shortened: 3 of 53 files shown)


bstansberry:
The WildFly full mixed domain tests are failing because they are running with assertions enabled. I don't see why this PR causes that though.

The core problem on Linux though is presumably related. I've seen intermittent hangs of OperationCancellationTestCase. If it's happening reproducibly that's actually kind of good.

bstansberry:
retest this please

bstansberry:
retest this please

bstansberry:
retest this please

bstansberry:
retest this please

kabir:
https://github.com/wildfly/wildfly-core/pull/1102 attempts to fix the transformer problem, which is causing the mixed domain tests to fail

testsuite/shared/src/main/java/org/jboss/as/test/integration/domain/management/util/DomainLifecycleUtil.java
```diff
@@ -149,6 +149,13 @@ public void start() {
                         .addProcessControllerJavaOptions(javaOpts);
             }
 
+            if (configuration.isEnableAssertions()) {
+                commandBuilder.addProcessControllerJavaOption("-ea");
+                commandBuilder.addHostControllerJavaOption("-ea");
+                // This doesn't work; the HC Main class treats this as an invalid param
+                //commandBuilder.addServerArgument("-ea");
```
jamezp: Should this work?

jamezp:
This looks good to me though I think @kabir should probably have a look too since he's a lot more familiar with how transformers work.

testsuite/shared/src/main/java/org/jboss/as/test/integration/domain/management/util/DomainLifecycleUtil.java
bstansberry: Not really, no. This ends up as a command line param to the HC Main class and it doesn't know what to do with it. It only knows how to handle specific things.

The way to control server jvm settings is via the <jvm> elements in domain/host.xml, which I update in one of the commits here. For this command line thing to work, the HC would have to understand "-ea" as meaning to add that to each JVM launch. Not worth implementing IMHO.

bstansberry:
retest this please

bstansberry:
retest this please

kabir:
I think it looks fine, but it looks like OperationCancellationTestCase is being difficult again

bstansberry:
Having these assertions enabled is very helpful. They are doing what they were meant to do, showing much more clearly in the logs what is going wrong when OperationCancellationTestCase fails.

bstansberry:
retest this please

bstansberry:
retest this please

@kabir i'm retesting again to get more reps, but it's looking good so far. Both of the core runs have passed 3 times with my latest.

What you looked at before is unchanged, but I've added another commit.

bstansberry:
retest this please

Had a different hang of some sort. Investigating

bstansberry:
retest this please

jamezp:
I'm good with this whenever you are @bstansberry.

bstansberry:
@jamezp now there have twice been hangs in HcExtensionAndSubsystemManagementTestCase. For some reason an HC reload isn't completing. Hard to see how it relates to this, but I haven't noticed it elsewhere.

#1129 aims to tighten up that test a bit, but what I did in that PR should just mean things happen a bit more cleanly. If the reload doesn't complete the test run will still be screwed up.

bstansberry:
retest this please

ctomc:
retest this please

bstansberry:
retest this please

bstansberry:
retest this please

bstansberry:
retest this please

bstansberry:
retest this please

bstansberry:
retest this please

bstansberry:
retest this please

bstansberry:
retest this please

bstansberry:
retest this please

bstansberry:
retest this please

bstansberry:
@rhusar @kabir the "Full integration - Windows" job that timed out in the previous comment hung here:

[03:08:57][org.wildfly:wildfly-clustering-infinispan-extension] Running org.jboss.as.clustering.infinispan.subsystem.TransformersTestCase
[04:17:32][org.wildfly:wildfly-clustering-infinispan-extension] The build Pull Request::WildFly Core Full - Integration - Windows #524 {buildId=74614} has been running for more than 75 minutes. Terminating...

Unfortunately on the Windows runs the thread dump captured in the Team City build log doesn't include the relevant processes.

That hang is very unlikely to be related to this PR, as this PR is about remote communication between separate processes and the subsystem unit tests do not rely on that.

bstansberry:
retest this please

rhusar:
The only known hang now is on https://issues.jboss.org/browse/WFLY-5290 but I don't see how that would be now triggerd with this change. Anyway any additional info would be very helpful.

bstansberry:
@rhusar I don't believe the hang I reported has anything to do with this PR. I think it was just my bad luck that it happened to this one.

bstansberry:
retest this please

bstansberry:
retest this please

bstansberry:
retest this please

bstansberry:
retest this please

bstansberry:
retest this please

bstansberry:
retest this please

PR #1088, opened 2015-09-21
WFCORE-933 - argument validation starting embedded server

luck3y:


luck3y:
retest this please

cli/src/main/java/org/jboss/as/cli/embedded/EmbedServerHandler.java
```diff
@@ -176,6 +168,9 @@ protected void doHandle(CommandContext ctx) throws CommandLineException {
 
             List<String> cmdsList = new ArrayList<>();
             if (xml != null && xml.trim().length() > 0) {
+                if (!xml.endsWith(".xml")) {
+                    throw new CommandFormatException("The --server-config filename must end with .xml.");
+                }
```
bstansberry: This is fine as a workaround, but it smells funny that this is necessary. Can you file a JIRA to sort out why it's needed? Just a comment on the existing JIRA explaining why is fine if you know the answer and it's reasonable.

The check should be if (!xml.trim().endsWith(".xml")) or the trimming can just be done once and the result kept. :)

bstansberry:
Other than the nit about trim(), looks good, but please squash the two commits.

cli/src/main/java/org/jboss/as/cli/embedded/EmbedServerHandler.java
luck3y: Yep, I'd already opened https://issues.jboss.org/browse/WFCORE-991 for this. There's some suspicious code in ConfiguratioinFile.addSuffixToFile(historyBase, INITIAL) assuming that the rightmost . is part of a file suffix, and without it being present it appends the config name to the next . it encounters to create the path for the initial config file backup. This is almost certainly a directory that doesn't exist.

PR #1089, opened 2015-09-21
[WFCORE-949] : Removing http-interface autocompletes but fails.

ehsavoie:
Removing the management interfaces will put the server in reload mode instead of stopping (and removing) those interfaces at runtime.
Until the introduction a new capability by refactoring the management we check manually if JMX requires a management interface for remoting access and if one is still available after removing.
Fixing the sybsystem test case capability registration so that we may define a serviceValueType.

Jira: https://issues.jboss.org/browse/WFCORE-949

### DIFF
--- a/host-controller/src/main/java/org/jboss/as/host/controller/operations/NativeManagementRemoveHandler.java
+++ b/host-controller/src/main/java/org/jboss/as/host/controller/operations/NativeManagementRemoveHandler.java
@@ -1,30 +1,28 @@
 /*
- * JBoss, Home of Professional Open Source.
- * Copyright 2011, Red Hat, Inc., and individual contributors
+ * Copyright (C) 2015 Red Hat, inc., and individual contributors
  * as indicated by the @author tags. See the copyright.txt file in the
  * distribution for a full listing of individual contributors.
  *
- * This is free software; you can redistribute it and/or modify it
- * under the terms of the GNU Lesser General Public License as
- * published by the Free Software Foundation; either version 2.1 of
- * the License, or (at your option) any later version.
+ * This library is free software; you can redistribute it and/or
+ * modify it under the terms of the GNU Lesser General Public
+ * License as published by the Free Software Foundation; either
+ * version 2.1 of the License, or (at your option) any later version.
  *
- * This software is distributed in the hope that it will be useful,
+ * This library is distributed in the hope that it will be useful,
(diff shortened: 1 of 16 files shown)


host-controller/src/main/java/org/jboss/as/host/controller/operations/NativeManagementRemoveHandler.java
```diff
@@ -0,0 +1,59 @@
...
+
+    @Override
+    protected boolean requiresRuntime(OperationContext context) {
+        return (context.getProcessType() != ProcessType.EMBEDDED_HOST_CONTROLLER);
+    }
+
+    @Override
+    protected void performRemove(OperationContext context, ModelNode operation, ModelNode model) throws OperationFailedException {
+        PathAddress nativeAddress = context.getCurrentAddress().getParent().append(PathElement.pathElement(MANAGEMENT_INTERFACE, HTTP_INTERFACE));
+        ManagementRemotingServices.isManagementResourceRemoveable(context, nativeAddress);
```
bstansberry: This should be done in an added step. Always do cross-resource validation in an added step, as that allows composites to function correctly (e.g. a composite that has one step removing http and another adding native.) The validation-in-an-added-step approach defers the validation check until both steps have run.

remoting/subsystem/src/main/java/org/jboss/as/remoting/logging/RemotingLogger.java
```diff
@@ -129,4 +130,7 @@
     @Message(id = 24, value = "The remoting subsystem is present but no io subsystem was found. An io subsystem " +
             "was not required when remoting schema '%s' was current but now is, so a default subsystem is being added.")
     void addingIOSubsystem(String legacyNS);
+
+    @Message(id = 25, value = "Can't remove %s as JMX uses it as remoting endpoint")
```
bstansberry: "uses it as a remoting endpoint"

Missing "a".

host-controller/src/main/java/org/jboss/as/host/controller/operations/NativeManagementRemoveHandler.java
bstansberry: Please assume this comment applies to all the remove handlers.

bstansberry:
Other than the added step thing, looks good. Thanks!

PR #1094, opened 2015-09-21
WFCORE-931 - Embedded host-controller throws NPE on regular shutdown

luck3y:


bstansberry:
Can this be fixed by not registering the operation at all? That is how it works in the standalone server case. We should avoid exposing an API if it can't be invoked.

PR #1098, opened 2015-09-22
[WFLY-5227] Move the security manager subsystem to wildfly-core

sguilhen:
Process I've followed to move the security subsystem from wildfly to wildfly-core:
- used filter-branch to select only the security-manager folder, preserving all the history
  git filter-branch --subdirectory-filter security-manager -- --all

- in wildfly-core, I've added a remote repo pointing to the new local repo that resulted from filter-branch. Then pulled (merged) the changes to import all the files along with the history.
  git pull wildfly security-manager

- finally removed the dependency on jboss-metadata and updated the core-feature-pack to include the new subsystem. Build was fine and I was able to boot the core server with the subsystem enabled.

I want to get some feedback on this as preserving the history ended up creating a merge commit. If the history in this case is not important I can just copy everything over and add a new commit that can be rebased like any other commit.

I got the changes in wildfly ready (removal of the subsystem) and of course it depends on a new release of wildfly-core. Once that happens it is probably best if I include the core upgrade in my commit so that both the upgrade and subsystem removal are handled in a single step.


sguilhen:
Retest this please

ctomc:
you could probably rebase -i and removed all commits that are "release xx" / "next is xx" as all they did was change version in pom.xml and as such should really be a factor.

ctomc:
retest this please

core-feature-pack/src/main/resources/configuration/standalone/subsystems.xml
```diff
@@ -5,5 +5,6 @@
       <subsystem>logging.xml</subsystem>
       <subsystem>jmx.xml</subsystem>
       <subsystem>request-controller.xml</subsystem>
+      <subsystem>security-manager.xml</subsystem>
```
ctomc: Having this here will make security manager subsystem enabled by default for all configuration build from this core.
Which would mean that we enable SM by default in all our shipped configurations.
I think this should be removed.

dmlloyd: Actually this is not accurate.  The security manager itself will only be enabled if the -secmgr flag is given at boot.

ctomc: ah yes, the -secmgr. So I guess this should be fine as for SM enablement goes.
but still I don't think we should have this in default standalone.xml of shipped core.
That part should stay in full shipped configs as it was till now.

dmlloyd: I think it's harmless to have in there though, and it's a useful example.

sguilhen:
I thought about not including all the "Release x.y.z" /  "Next is a.b.c" commits but I saw that the history of all the other poms in wildfly-core have these commits so I decided to leave them in. I don't think they bring much trouble but I can remove them if that's the preferable way now.

dmlloyd:
They're fine to leave.

bstansberry:
@sguilhen Please send up your full WildFly PR for doing WFLY-5227 on the "full" side. It should *not* include a commit changing the version of WildFly Core, so it will just fail CI of course. But that's ok. Without that PR up and available for use though we're stuck and can't proceed.

sguilhen:
@bstansberry Done https://github.com/wildfly/wildfly/pull/8175

sguilhen:
I'm closing this PR for now. The subsystem has changed in WF due to bug fixes and we need to redo the whole thing to capture the current state of the subsystem.

PR #1099, opened 2015-09-22
WFCORE-937 - --std-out=discard for embedded host controller doesn't work

luck3y:
Starting embeded standalone would overwrite standalone/configuration/logging.properties, quitting and starting after would start with an empty logging.properties.

@bstansberry can you have a look and tell me if you think this is the right thing to do here?

### DIFF
--- a/logging/src/main/java/org/jboss/as/logging/logmanager/ConfigurationPersistence.java
+++ b/logging/src/main/java/org/jboss/as/logging/logmanager/ConfigurationPersistence.java
@@ -377,6 +377,9 @@ public class ConfigurationPersistence implements Configurator, LogContextConfigu
                 loggingConfig = FileResolver.resolvePath(context, "jboss.server.data.dir", PROPERTIES_FILE);
                 break;
             }
+            case EMBEDDED_SERVER: {
+                return;
+            }
             default: {
                 loggingConfig = FileResolver.resolvePath(context, "jboss.server.config.dir", PROPERTIES_FILE);
             }


luck3y:
retest this please

luck3y:
Retest this please

logging/src/main/java/org/jboss/as/logging/logmanager/ConfigurationPersistence.java
```diff
@@ -377,6 +377,9 @@ public void writeConfiguration(final OperationContext context) {
                 loggingConfig = FileResolver.resolvePath(context, "jboss.server.data.dir", PROPERTIES_FILE);
                 break;
             }
+            case EMBEDDED_SERVER: {
+                return;
+            }
```
jamezp: I think this should be added, I don't think it solves the issue though. Testing locally I still see log output. It's possible the `$JBOSS_HOME/domain/configuration/logging.properties` is being found on the class path which is why it's displaying log messages. Logging does seem relevant, but I can see where we don't want it output to the console. I'd have to look further into what `--std-out=discard` actually does.

luck3y: Yeah, this is the first bit, when logging properties is clobbered it
definitely confuses the issue.

luck3y:
Forgot a piece of this...

luck3y:
@jamezp I've moved this change to https://issues.jboss.org/browse/WFCORE-1001 and submitted PR https://github.com/wildfly/wildfly-core/pull/1107

luck3y:
So are you thinking just going with the above then, or back to something like:

            case EMBEDDED_SERVER: {
                return;
            }

as the only one skipping writing?

PR #1101, opened 2015-09-22
[WFLY-5332]: Allow configure max-threads and core-threads independently. Allow idle thread reusage.

panossot:
[WFLY-5332] : https://issues.jboss.org/browse/WFLY-5332

The PR tests will fail till https://github.com/jbossas/jboss-threads/pull/16 is included.

### DIFF
--- a/threads/src/main/java/org/jboss/as/threads/ManagedJBossThreadPoolExecutorReuseIdleThreadsService.java
+++ b/threads/src/main/java/org/jboss/as/threads/ManagedJBossThreadPoolExecutorReuseIdleThreadsService.java
@@ -1,0 +1,150 @@
+/*
+ * JBoss, Home of Professional Open Source.
+ * Copyright 2015, Red Hat, Inc., and individual contributors
+ * as indicated by the @author tags. See the copyright.txt file in the
+ * distribution for a full listing of individual contributors.
+ *
+ * This is free software; you can redistribute it and/or modify it
+ * under the terms of the GNU Lesser General Public License as
+ * published by the Free Software Foundation; either version 2.1 of
+ * the License, or (at your option) any later version.
+ *
+ * This software is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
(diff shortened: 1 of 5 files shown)


ctomc:
parser change is wrong. You need to update 2.0 parser not the legacy one.
There are no parser / model tests for new attribute.
this is just subsystem wise.
@dmlloyd wdyt about the concept? I know you are not fan such thread pools.

panossot:
You are right Tomaz. It seems upstream needs some mere work.

dmlloyd:
First, our thread pools already always reuse idle threads; that certainly isn't a problem.  So I think the pool name is misleading.

However, in general I am highly, highly skeptical of the safety of a new thread pool implementation.  In addition, there is a large amount of locking going on in this pool implementation.  I'm not inclined to approve it as is unless it can be shown to have equal or better performance to the existing pools.

panossot:
Hello David, it should be compared with which thread pool? (JBossThreadPoolExecutor, which was the default, was not reusing idle threads)

dmlloyd:
Are you saying that it was creating a new thread for every task?

panossot:
Till core threads, yes. And since for the default pool we use corethreads=maxthreads, till max threads it creates a new thread for every new task.

dmlloyd:
Oh, yes of course.  The pool must prefill; that's what core threads are.  Max threads are non-prefilled threads, which can be reused in preference to creating new threads.

That said, no existing thread pool has implemented core/max threads in an efficient *and* safe manner.

panossot:
I will try to create some test to prove its efficiency/safety.

dmlloyd:
For a thread pool, tests can prove efficiency but not safety.  To prove safety, because it is concurrent code, you need nothing less than a complete state diagram.

panossot:
Ok. I will try to create a state diagram as well.

bstansberry:
Please submit a design proposal to wildfly-dev. Jason intends to start requiring that for new features, and for this one it sounds particularly useful. We can then review pull requests in comparison to an agreed upon design.

PR #1102, opened 2015-09-22
Possible fix for https://github.com/wildfly/wildfly-core/pull/1086

kabir:
The additional commit is an attempt to fix the mixed domain failures in https://github.com/wildfly/wildfly-core/pull/1086

### DIFF
--- a/controller/src/main/java/org/jboss/as/controller/transform/ResourceTransformationContextImpl.java
+++ b/controller/src/main/java/org/jboss/as/controller/transform/ResourceTransformationContextImpl.java
@@ -405,7 +405,7 @@ class ResourceTransformationContextImpl implements ResourceTransformationContext
         private final ImmutableManagementResourceRegistration registration;
 
         OriginalModel(Resource original, RunningMode mode, ProcessType type, TransformationTarget target, ImmutableManagementResourceRegistration registration) {
-            this.original = original;
+            this.original = original.clone();
             this.mode = mode;
             this.type = type;
             this.target = target;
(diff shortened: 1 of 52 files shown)


controller/src/main/java/org/jboss/as/controller/transform/ResourceTransformationContextImpl.java
```diff
@@ -412,7 +405,7 @@ public TransformersLogger getLogger() {
         private final ImmutableManagementResourceRegistration registration;
 
         OriginalModel(Resource original, RunningMode mode, ProcessType type, TransformationTarget target, ImmutableManagementResourceRegistration registration) {
-            this.original = original;
+            this.original = original.clone();
```
bstansberry: This largely defeats the point of https://github.com/wildfly/wildfly-core/pull/1102/files#diff-89ed7c94b256f549e7a6789336b54404R135 . I'm not saying it's wrong, I'm just a little sad. :(

I think I'll keep the Stage.DOMAIN operation attachment thing though. It at least saves the cost of doing a full OC.readResourceFromRoot each time, and limits the expense to just a Resource.clone().

bstansberry:
retest this please

I want more runs to see if the OperationCancellationTestCase issue is gone.

kabir:
Retest this please

bstansberry:
Thanks. I've pulled this commit into my branch. I saw another failure of OperationCancellationTestCase in a run last night of my branch, so I'm not confident the problem there is fixed with my last commit. So I'm going to close this and continue with my branch with your commit in it.

PR #1105, opened 2015-09-22
WFCORE-715 Xalan Linkage error : TransformerConfigurationException fo…

ctomc:
…r domain setup

### DIFF
--- a/core-feature-pack/src/main/resources/modules/system/layers/base/org/jboss/as/server/main/module.xml
+++ b/core-feature-pack/src/main/resources/modules/system/layers/base/org/jboss/as/server/main/module.xml
@@ -36,8 +36,11 @@
 
     <dependencies>
         <module name="javax.api"/>
-        <!--this needs to remain here until common-beans is fixed to use passed classloader -->
-        <module name="org.jboss.common-beans" services="export" optional="true"/>
+        <!--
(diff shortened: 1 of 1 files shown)


bstansberry:
ClusteredMessagingTestCase will be ignored as soon as my PR doing so compiles.

bstansberry:
@n1hility I'll leave this one to both in terms of whether it's the correct fix, and whether you want it in WF 10. I added a 3.x label so we can mark stuff we don't want in 10.

ctomc:
It would make sense to get this into EAP7, no mater what the core / WF version is.
currently we have behavior where same application that is using XSLT works in standalone mode but fails in domain, that to me is a blocking bug. The jira has reproducer for this exact scenario.

n1hility:
This isn't the correct fix. Jaxpmodule being removed everywhere is the cause.

n1hility:
I'll fix everything up

ctomc:
@n1hility that would mean that also standalone entry point it wrong as it imports xalan atm.

n1hility:
Actually this specific change was fine since jaxpmodule redirection falls back to the boot module, and the server module is the boot module for domain mode. It is just missing the other default implementations though, which I have added in another PR.

PR #1107, opened 2015-09-22
WFCORE-1001 - embedded server start clobbers logging.properties

luck3y:


bstansberry:
@jamezp please approve

jamezp:
Well I told him it looks good, but now I'm questioning my decision. It does seem that if it's an `embed-server` we would want it persisted. If it's from an `embed-host-controller` we don't. Maybe we just need to return from the default and only write if it's a `DOMAIN_SERVER`, `STANDALONE_SERVER` and maybe `EMBEDDED_SERVER` (I can't recall when the process type is set to this). With the current state even `APPLICATION_CLIENT` and `SELF_CONTAINED` would likely overwrite it.

luck3y:
The issue with embed-server is that if it does persist, it persists a
pretty much empty file. So perhaps there is something wrong with how the
loggers are being configured at startup?

jamezp:
Essentially what needs to happen is if there's a logging subsystem configuration change and the change should be persisted, then it needs to be written to `${jboss.server.config.dir}/logging.properties` for standalone or `${jboss.server.data.dir}/logging.properties` for domain servers; e.g. `server-one` needs to be in `$JBOSS_HOME/domain/configuration/servers/server-one/data/logging.properties`.

luck3y:
Yeah, so with this change it doesn't overwrite logging.properties on
startup and it does persist logging changes made subsequently in the CLI.
Thats how I tested it anyway.

Not at all saying this patch is right, just making sure I mention all the
possibly pertinent facts.

jamezp:
Hmmm... ...it's not being persisted for me, but we definitely need something similar to this. Launching an `embed-server` from CLI completely wipes this file.

jamezp:
After looking at this I think it should be:
```java
        switch (context.getProcessType()) {
            case DOMAIN_SERVER: {
                loggingConfig = FileResolver.resolvePath(context, "jboss.server.data.dir", PROPERTIES_FILE);
                break;
            }
            case STANDALONE_SERVER: {
                loggingConfig = FileResolver.resolvePath(context, "jboss.server.config.dir", PROPERTIES_FILE);
                break;
                
            }
            default: {
                return;
            }
        }
```

The `logging.properties` shouldn't be written, at least for now, in embedded containers because the `LogContext` will not be correct. The context is only updated in the `RUNTIME`.

luck3y:
@jamezp thanks & done!

jamezp:
Looks good to me!

luck3y:
@bstansberry @jamezp So based on the discussion yesterday, what do you think we should do with this?

bstansberry:
@luck3y I'd like to understand why the file ends up empty. Until I understand that, I have no idea if just not persisting is the right answer.

luck3y:
Yep, understood, I'll figure out whats going on.

bstansberry:
I'll add the "Fixme" label to this, as the CI integration will nicely remove it if this PR is updated. :)

jamezp:
For some reason the runtime step seems to be executing to write the file. Since the default on the switch was to write the `$JBOSS_HOME/standalone/configuration/logging.properties` file it gets overwritten with an empty `LogContextConfiguration`. This does fix the empty file getting written, but mainly because I think it's more correct to only overwrite the standalone configuration when it's a standalone server.

I still think there's a different issue in that the runtime stage is executing. I'll debug that further.

luck3y:
@jamezp Yes, the empty LogContext doesn't seem to make any sense to write here. The only alternative  would be to populate it correctly with the same logging.properties. I don't know if doing that would make proper sense for the embedded case though, WDYT?

jamezp:
I think at some stage it could make sense to write the `logging.properties` if a configuration change was made. Currently though nothing would be committed and therefore it would always be empty. So for now I'd like to avoid writing it for embedded servers.

luck3y:
@bstansberry @jperkins Anything else you guys would like to see in the comment?

bstansberry:
@luck3y looks good to me @jamezp please remove the "Ready for merge" I'm about to add if you want any changes.

PR #1108, opened 2015-09-22
WFCORE-932 stop-embedded-server stops the embedded host controller and vice versa

luck3y:
Restrict which server stop commands are available based on the type started.

### DIFF
--- a/cli/src/main/java/org/jboss/as/cli/embedded/StopEmbeddedHostControllerHandler.java
+++ b/cli/src/main/java/org/jboss/as/cli/embedded/StopEmbeddedHostControllerHandler.java
@@ -46,7 +46,7 @@ class StopEmbeddedHostControllerHandler extends CommandHandlerWithHelp {
 
     @Override
     public boolean isAvailable(CommandContext ctx) {
-        return hostControllerReference.get() != null;
+        return (hostControllerReference.get() != null && hostControllerReference.get().isHostController());
     }
 
     @Override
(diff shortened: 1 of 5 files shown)


cli/src/main/java/org/jboss/as/cli/embedded/StopEmbeddedHostControllerHandler.java
```diff
@@ -52,7 +52,7 @@ public boolean isAvailable(CommandContext ctx) {
     @Override
     protected void doHandle(CommandContext ctx) throws CommandLineException {
         EmbeddedServerLaunch hostControllerLaunch = hostControllerReference.get();
-        if (hostControllerLaunch != null) {
+        if (hostControllerLaunch != null && hostControllerLaunch.getHostController() != null) {
```
bstansberry: Shouldn't this be in isAvailable()? That's what will prevent the CLI offering the option.

Having it here too is fine, but people shouldn't see the command if it's not valid.

luck3y: This is calling EmbeddedServerLaunch.getHostController() which in turn looks at EmbeddedServerReference to see if its a running hostcontroller or a standalone server. If this is a standalone server here, then we just ignore the shutdown, as standalone handles it instead.

As I recall it was a lot cleaner to implement it this way from the CLI side, but the issue was that you could use either shutdown command to shutdown either embedded process type.

luck3y:
retest this please

luck3y:
retest this please

bstansberry:
@pferraro the SingletonDeploymentDescriptorTestCase failure shown in one of the CI comments above seems to be failing pretty regularly. Twice more here: https://github.com/wildfly/wildfly-core/pull/1112

PR #1115, opened 2015-09-24
WFCORE-1010 standalone.sh hard codes boot log

marschall:
The standalone.sh file hard codes the org.jboss.boot.log.file property
to server.log. This is inconvenient because it overrides the dynamic
configuration in logging.properties which is set to
${org.jboss.boot.log.file:boot.log}.

 - remove hard coding

Issue: WFCORE-1010
https://issues.jboss.org/browse/WFCORE-1010

### DIFF
--- a/core-feature-pack/src/main/resources/content/bin/jboss-cli.ps1
+++ b/core-feature-pack/src/main/resources/content/bin/jboss-cli.ps1
@@ -114,7 +114,6 @@ echo ""
 
 $PROG_ARGS = @()
 $PROG_ARGS += $JAVA_OPTS
-$PROG_ARGS += "-Dorg.jboss.boot.log.file=$JBOSS_LOG_DIR/boot.log"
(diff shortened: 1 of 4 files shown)


jamezp:
Removing the override of the property will put the `server.log` in the incorrect place on the initial boot. Note that this property is really only used once at the initial boot time. The logging subsystem rewrites the `logging.properties` with an absolute path to the `server.log`.

marschall:
What we have been seeing with WildFly 9.0.1 is that we configure the log file in standalone.xml to something other than `server.log` but the boot log still ends up in `server.log`

jamezp:
If you edit the XML file that will happen as the subsystem overwrites the `logging.properties` file. I'd venture to guess that this change doesn't fix that particular issue. https://issues.jboss.org/browse/WFCORE-157 explains some of the issues with writing expressions to the `logging.properties` file.

bstansberry:
Closing as the JIRA was closed.

PR #1117, opened 2015-09-24
WFCORE-1012 Remove Tomcat modeler log configuration

marschall:
The org.apache.tomcat.util.modeler logger is still configured in
several places. This seems to be a leftover from the days when WildFly
used Tomcat as a servlet implementation.

Issue: WFCORE-1012
https://issues.jboss.org/browse/WFCORE-1012

### DIFF
--- a/logging/src/test/resources/default-subsystem.xml
+++ b/logging/src/test/resources/default-subsystem.xml
@@ -38,9 +38,6 @@
     <logger category="com.arjuna">
         <level name="WARN"/>
     </logger>
-    <logger category="org.apache.tomcat.util.modeler">
-        <level name="WARN"/>
-    </logger>
     <logger category="org.jboss.as.config">
         <level name="DEBUG"/>
     </logger>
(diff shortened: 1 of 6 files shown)


logging/src/test/resources/default-subsystem.xml
```diff
@@ -38,9 +38,6 @@
     <logger category="com.arjuna">
         <level name="WARN"/>
     </logger>
-    <logger category="org.apache.tomcat.util.modeler">
-        <level name="WARN"/>
-    </logger>
```
bstansberry: @jamezp WDYT about this one? The rest of them either affect testsuite configs or only affect servers using a current installation, where jboss-web is not used. By this one will affect the logging system in our standard domain.xml profiles, and those profiles can be adapted for use with legacy slaves that still have jboss-web.

OTOH, those profiles would have to be *adapted* for use; i.e. the user isn't just taking our standard profile and using it. So I think having to notice and deal with these kinds of subtle behavior differences isn't unreasonable, and this patch is ok.

jamezp: Yeah I think it's okay. Like you said the main issue would be with older slaves, but they'd likely need some other changes too.

bstansberry:
Thanks!

PR #1118, opened 2015-09-24
Add missing powershell scripts & fix existing

ctomc:
- WFCORE-745 Standalone powershell script does't support --debug and -secmgr option
- WFCORE-742 Unify linux and PowerShell scripts for jboss-cli in printing environment information
- WFCORE-744 Unify add-user script functionality in windows (powershell, bat) and linux
- WFCORE-737 PowerShell script "vault.ps1" is missing
- WFCORE-1030 PS1 scripts for standalone and domain do not survive :shutdown(restart=true)

### DIFF
--- a/core-feature-pack/src/main/resources/content/bin/add-user.ps1
+++ b/core-feature-pack/src/main/resources/content/bin/add-user.ps1
@@ -1,0 +1,25 @@
+#############################################################################
+#                                                                          ##
+#    WildFly Startup Script for adding users							   ##
+#                                                                          ##
+#############################################################################
+. ".\common.ps1"
+
+$JAVA_OPTS = @()
+
+# Sample JPDA settings for remote socket debugging
+#$JAVA_OPTS+="-agentlib:jdwp=transport=dt_socket,address=8787,server=y,suspend=y"
+# Uncomment to override standalone and domain user location
+#$JAVA_OPTS+="-Djboss.server.config.user.dir=$JBOSS_HOME/standalone/configuration"
+#$JAVA_OPTS+="-Djboss.domain.config.user.dir=$JBOSS_HOME/domain/configuration"
+
+
+$PROG_ARGS = Get-Java-Arguments -entryModule "org.jboss.as.domain-add-user" -serverOpts $ARGS -logFileProperties $SCRIPTS_HOME\add-user.properties
+
+try{
+	pushd $JBOSS_HOME
+	& $JAVA $PROG_ARGS
+}finally{
+	popd
+	Env-Clean-Up
+}
--- a/core-feature-pack/src/main/resources/content/bin/common.ps1
+++ b/core-feature-pack/src/main/resources/content/bin/common.ps1
@@ -1,0 +1,279 @@
+if($PSVersionTable.PSVersion.Major -lt 2) {
+    Write-Warning "This script requires PowerShell 2.0 or better; you have version $($Host.Version)."
+    return
+}
+
+$SCRIPTS_HOME = (Get-ChildItem $MyInvocation.MyCommand.Path).Directory.FullName
+$RESOLVED_JBOSS_HOME = (Get-ChildItem $MyInvocation.MyCommand.Path).Directory.Parent.FullName
+
+
+$global:SECMGR = $false
+$global:DEBUG_MODE=$false
+$global:DEBUG_PORT=8787
+$global:RUN_IN_BACKGROUND=$false
+$global:GC_LOG=$false
+#module opts that are passed to jboss modules
+$global:MODULE_OPTS = @()
+
+# A collection of functions that are used by the other scripts
+
+Function Set-Env {
+  $key = $args[0]
+  $value = $args[1]
+  Set-Content -Path env:$key -Value $value
+}
+
+Function Get-Env {
+  $key = $args[0]
+  if( Test-Path env:$key ) {
+    return (Get-ChildItem env:$key).Value
+  }
+  return $args[1]
+}
+
+Function Get-String {
+  $value = ''
+  foreach($k in $args) {
+    $value += $k
+  }
+  return $value
+}
+
+Function String-To-Array($value) {
+  $res = @()
+  if (!$value){
+  	return $res
+  }
+  $tmpArr = $value.split()
+  
+  foreach ($str in $tmpArr) {
+    if ($str) {
+	  $res += $str
+	}
+  }
+  return $res
+}
+
+Function Display-Environment { 
+$JAVA_OPTS = Get-Java-Opts
+# Display our environment
+Write-Host "================================================================================="
+Write-Host ""
+Write-Host "  JBoss Bootstrap Environment"
+Write-Host ""
+Write-Host "  JBOSS_HOME: $JBOSS_HOME"
+Write-Host ""
+Write-Host "  JAVA: $JAVA"
+Write-Host ""
+Write-Host "  MODULE_OPTS: $MODULE_OPTS"
+Write-Host ""
+Write-Host "  JAVA_OPTS: $JAVA_OPTS"
+Write-Host ""
+Write-Host "================================================================================="
+Write-Host ""
+
+}
+
+#todo: bit funky at the moment, should probably be done via global variable
+Function Get-Java-Opts {
+	if($PRESERVE_JAVA_OPTS -ne 'true') { # if not perserve, then check for enviroment variable and use that
+		if( (Test-Path env:JAVA_OPTS)) {
+			$ops = Get-Env JAVA_OPTS
+			# This is Powershell, so split the incoming string on a space into array
+			return String-To-Array -value $ops
+			Write-Host "JAVA_OPTS already set in environment; overriding default settings with values: $JAVA_OPTS"
+		}
+	}
+	return $JAVA_OPTS
+}
+
+Function Display-Array($array){
+	for ($i=0; $i -lt $array.length; $i++) {
+		$v =  "$i " + $array[$i]
+		Write-Host $v
+	}
+}
+
+
+Function Get-Java-Arguments {
+Param(
+   [Parameter(Mandatory=$true)]
+   [string]$entryModule,
+   [string]$logFileProperties = "$JBOSS_CONFIG_DIR/logging.properties",
+   [string]$logFile = "$JBOSS_LOG_DIR/server.log",
+   [string[]]$serverOpts
+   
+
+) #end param
+  $JAVA_OPTS = Get-Java-Opts #takes care of looking at defind settings and/or using env:JAVA_OPTS
+
+  $PROG_ARGS = @()
+  if ($JAVA_OPTS -ne $null){
+  	$PROG_ARGS += $JAVA_OPTS
+  }  
+  $PROG_ARGS += "-Dorg.jboss.boot.log.file=$logFile"
+  $PROG_ARGS += "-Dlogging.configuration=file:$logFileProperties"
+  $PROG_ARGS += "-Djboss.home.dir=$JBOSS_HOME"
+  $PROG_ARGS += "-jar"
+  $PROG_ARGS += "$JBOSS_HOME\jboss-modules.jar"
+  if ($MODULE_OPTS -ne $null){
+  	$PROG_ARGS += $MODULE_OPTS
+  }
+  $PROG_ARGS += "-mp"
+  $PROG_ARGS += "$JBOSS_MODULEPATH"
+  $PROG_ARGS += $entryModule
+  if ($serverOpts -ne $null){
+  	$PROG_ARGS += $serverOpts
+  }
+  return $PROG_ARGS
+}
+
+Function Process-Script-Parameters {
+Param(
+   [Parameter(Mandatory=$false)]
+   [string[]]$Params   
+
+) #end param
+    $res = @()
+	for($i=0; $i -lt $Params.Count; $i++){
+		$arg = $Params[$i]
+		if ($arg -eq '--debug'){
+			$global:DEBUG_MODE=$true
+			if ($args[$i+1] -match '\d+'){ #port number can only follow --debug
+				$global:DEBUG_PORT = $Params[$i+1]
+				$i++
+				continue
+			}
+		}elseif ($arg -contains '-Djava.security.manager'){
+			Write-Warning "ERROR: The use of -Djava.security.manager has been removed. Please use the -secmgr command line argument or SECMGR=true environment variable."
+			exit
+		}elseif ($arg -eq '-secmgr'){
+			$global:SECMGR = $true
+		}elseif ($arg -eq '--background'){
+			$global:RUN_IN_BACKGROUND = $true
+		}else{
+			$res+=$arg
+		}
+	}
+	return $res
+}
+
+Function Start-WildFly-Process {
+ Param(
+   [Parameter(Mandatory=$true)]
+   [string[]] $programArguments,
+   [boolean] $runInBackground = $false   
+
+) #end param
+
+	if(($JBOSS_PIDFILE -ne '') -and (Test-Path $JBOSS_PIDFILE)) {
+		$processId = gc $JBOSS_PIDFILE
+		if ($processId -ne $null){
+			$proc = Get-Process -Id $processId -ErrorAction SilentlyContinue
+		}
+		if ($proc -ne $null){
+			Write-Warning "Looks like a server process is already running. If it isn't then, remove the $JBOSS_PIDFILE and try again"
+			return
+		}else{
+			Remove-Item $JBOSS_PIDFILE
+		}
+	}
+
+	if($runInBackground) {	
+		$process = Start-Process -FilePath $JAVA -ArgumentList $programArguments -NoNewWindow -RedirectStandardOutput $global:CONSOLE_LOG -WorkingDirectory $JBOSS_HOME -PassThru
+		$processId = $process.Id;
+		echo "Started process in background, process id: $processId" 
+		if ($JBOSS_PIDFILE -ne $null){
+			$processId >> $JBOSS_PIDFILE  	
+		}
+	} else {
+		try{
+			pushd $JBOSS_HOME
+			& $JAVA $programArguments
+		}finally{
+			popd			
+		}
+	}
+	Env-Clean-Up
+}
+
+Function Set-Global-Variables {
+PARAM(
+[Parameter(Mandatory=$true)]
+   [string]$baseDir
+)
+
+	# determine the default base dir, if not set
+	$global:JBOSS_BASE_DIR = $baseDir;
+	
+	# determine the default log dir, if not set
+	$global:JBOSS_LOG_DIR = Get-Env JBOSS_LOG_DIR $JBOSS_BASE_DIR\log
+	
+	# determine the default configuration dir, if not set
+	$global:JBOSS_CONFIG_DIR = Get-Env JBOSS_CONFIG_DIR $JBOSS_BASE_DIR\configuration
+	
+	$global:CONSOLE_LOG = $JBOSS_LOG_DIR + '\console.log'
+}
+
+Function Set-Global-Variables-Standalone {
+	$dir = Get-Env JBOSS_BASE_DIR $JBOSS_HOME\standalone
+	Set-Global-Variables -baseDir $dir
+}
+
+Function Set-Global-Variables-Domain {
+	$dir = Get-Env JBOSS_BASE_DIR $JBOSS_HOME\domain
+	Set-Global-Variables -baseDir $dir	
+}
+
+Function Env-Clean-Up {
+	[Environment]::SetEnvironmentVariable("JBOSS_HOME", $null, "Process")
+}
+
+Function Rotate-GC-Logs {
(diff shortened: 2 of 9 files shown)


marekkopecky:
I check these scripts with PowerShell 4:

* add-user.ps1 script print redundant log info to console
```
PS jboss-eap-7.0\bin> ./add-user.ps1 -u marek1 -p Password1* -s
15:30:08,188 INFO  [org.jboss.modules] (main) JBoss Modules version 1.4.4.Final
PS jboss-eap-7.0\bin>
```


* Standalone script do not process any argument. Example:
 .\standalone.ps1 --help
This command do not print help message, but start application server.


* Domain script do not work:
```
PS jboss-eap-7.0\bin> .\domain.ps1 --help
=================================================================================

  WildFly Bootstrap Environment

  JBOSS_HOME: C:\Users\Administrator\playground\7.0.0.DR11\jboss-eap-7.0

  JAVA: java

  MODULE_OPTS:

  JAVA_OPTS: -Xms64M -Xmx512M -Dsun.rmi.dgc.client.gcInterval=3600000 -Dsun.rmi.dgc.server.gcInterval=3600000 -Djava.ne
t.preferIPv4Stack=true -Dorg.jboss.resolver.warning=true -Djboss.modules.system.pkgs=org.jboss.byteman

=================================================================================

Error: Could not find or load main class Controller]
PS jboss-eap-7.0\bin>
```

* jboss-cli.ps1 script still print environment information, see description of WFCORE-742

ctomc:
@marekkopecky all of above should be now fixed

marekkopecky:
* I still see same issue on add-user.ps1 script and domain.ps1 script. Do you commit all changes?

ctomc:
yeah I have everything commited. 
Maybe it is issue that I use powershell v5, but am not using any features like that.
Will test on older windows as well.

marekkopecky:
@ctomc We should support also Powershell v2 and v3. Powershell v2 is default on W2008R2 and we support W2008R2.

ctomc:
@marekkopecky yes it must work in v2 for sure, that was a bug.
I think i managed to fix it now. Tested on WinServer 2012 with v4 of powershell

core-feature-pack/src/main/resources/content/bin/common.ps1
```diff
@@ -0,0 +1,256 @@
...
+		}
+	}
+	return $res
+}
+
+Function Start-WildFly-Process {
+ Param(
+   [Parameter(Mandatory=$true)]
+   [string[]] $programArguments,
+   [boolean] $runInBackground = $false
```
jamezp: Does this mean `false` is a default or does it currently always set `runInBackground` to `false`?

core-feature-pack/src/main/resources/content/bin/common.ps1
```diff
@@ -0,0 +1,256 @@
...
+		echo "Started process in background, process id: $processId" 
+		if ($JBOSS_PIDFILE -ne $null){
+			$processId >> $JBOSS_PIDFILE  	
+		}
+	} else {
+		try{
+			pushd $JBOSS_HOME
+			& $JAVA $programArguments
+		}finally{
+			pushd $SCRIPTS_HOME
```
jamezp: Should this just be `popd`?

core-feature-pack/src/main/resources/content/bin/add-user.ps1
```diff
@@ -0,0 +1,25 @@
...
+$JAVA_OPTS = @()
+
+# Sample JPDA settings for remote socket debugging
+#$JAVA_OPTS+="-agentlib:jdwp=transport=dt_socket,address=8787,server=y,suspend=y"
+# Uncomment to override standalone and domain user location
+#$JAVA_OPTS+="-Djboss.server.config.user.dir=$JBOSS_HOME/standalone/configuration"
+#$JAVA_OPTS+="-Djboss.domain.config.user.dir=$JBOSS_HOME/domain/configuration"
+
+
+$PROG_ARGS = Get-Java-Arguments -entryModule "org.jboss.as.domain-add-user" -serverOpts $ARGS -logFileProperties $SCRIPTS_HOME\add-user.properties
```
jamezp: Why does this have a `-logFileProperties`? Also excuse any other stupid questions as I just don't know PowerShell :)

bstansberry:
@marekkopecky Thanks for helping with the review of this. Can you post a +1 comment once your concerns are addressed?

core-feature-pack/src/main/resources/content/bin/common.ps1
```diff
@@ -0,0 +1,256 @@
+if($PSVersionTable.PSVersion.Major -lt 2) {
+    Write-Warning "posh-git requires PowerShell 2.0 or better; you have version $($Host.Version)."
```
jamezp: What is `posh-git` and is it relevant to these scripts?

core-feature-pack/src/main/resources/content/bin/common.ps1
```diff
@@ -0,0 +1,256 @@
...
+echo ""
+
+}
+
+Function Get-Java-Arguments {
+Param(
+   [Parameter(Mandatory=$true)]
+   [string]$entryModule,
+   [string]$logFileProperties = "$JBOSS_CONFIG_DIR/logging.properties",
+   [string]$logFile = "$JBOSS_LOG_DIR/server.log",
```
jamezp: This doesn't seem like a good default. `server.log` should only be used for standalone servers.

core-feature-pack/src/main/resources/content/bin/common.ps1
```diff
@@ -0,0 +1,256 @@
...
+Function Process-Script-Parameters {
+Param(
+   [Parameter(Mandatory=$false)]
+   [string[]]$Params   
+
+) #end param
+    $res = @()
+	for($i=0; $i -lt $Params.Count; $i++){
+		$arg = $Params[$i]
+		if ($arg -eq '--debug'){
```
jamezp: This should only be checked for standalone.

core-feature-pack/src/main/resources/content/bin/common.ps1
```diff
@@ -0,0 +1,256 @@
...
+			$global:DEBUG_MODE=$true
+			if ($args[$i+1] -match '\d+'){ #port number can only follow --debug
+				$global:DEBUG_PORT = $Params[$i+1]
+				$i++
+				continue
+			}
+		}elseif ($arg -contains '-Djava.security.manager'){
```
jamezp: This should only be checked for standalone. Also @dmlloyd were we going to support using the system property or just  `-semgr`? I can't recall and it's been quite a while since we discussed it.

core-feature-pack/src/main/resources/content/bin/common.ps1
```diff
@@ -0,0 +1,256 @@
...
+			echo "ERROR: The use of -Djava.security.manager has been removed. Please use the -secmgr command line argument or SECMGR=true environment variable."
+			exit
+		}elseif ($arg -eq '-secmgr'){
```
jamezp: This should only be checked for standalone.

core-feature-pack/src/main/resources/content/bin/common.ps1
```diff
@@ -0,0 +1,256 @@
...
+	# determine the default log dir, if not set
+	$global:JBOSS_LOG_DIR = Get-Env JBOSS_LOG_DIR $JBOSS_BASE_DIR\log
+	
+	# determine the default configuration dir, if not set
+	$global:JBOSS_CONFIG_DIR = Get-Env JBOSS_CONFIG_DIR $JBOSS_BASE_DIR\configuration
+	
+	$global:CONSOLE_LOG = $JBOSS_LOG_DIR + '\console.log'
+}
+
+Function Set-Global-Variables-Standalone {
```
jamezp: Why would this be in the common script?

core-feature-pack/src/main/resources/content/bin/common.ps1
```diff
@@ -0,0 +1,256 @@
...
+	$dir = Get-Env JBOSS_BASE_DIR $JBOSS_HOME\standalone
+	Set-Global-Variables -baseDir $dir
+}
+
+Function Set-Global-Variables-Domain {
```
jamezp: Why would this be in the common script?

core-feature-pack/src/main/resources/content/bin/common.ps1
```diff
@@ -0,0 +1,256 @@
...
+  $JBOSS_HOME=$SANITIZED_JBOSS_HOME
+} else {
+    # get the full path (without any relative bits)
+    $JBOSS_HOME=$RESOLVED_JBOSS_HOME
+}
+
+# Setup the JVM
+if (!(Test-Path env:JAVA)) {
+  if( Test-Path env:JAVA_HOME) {
+    $JAVA = (Get-ChildItem env:JAVA_HOME).Value + "\bin\java"
```
jamezp: Should it test `java.exe` too? I seem to recall when using a `ProcessBuilder` using `java.exe` was required.

core-feature-pack/src/main/resources/content/bin/common.ps1
```diff
@@ -0,0 +1,256 @@
...
+  } else {
+    $JAVA = 'java'
+  }
+}
+
+# determine the default module path, if not set
+$JBOSS_MODULEPATH = Get-Env JBOSS_MODULEPATH $JBOSS_HOME\modules
+
+Set-Global-Variables-Standalone
```
jamezp: Why would this be in the common script?

core-feature-pack/src/main/resources/content/bin/domain.conf.ps1
```diff
@@ -0,0 +1,83 @@
...
+$JAVA_OPTS += '-Dsun.rmi.dgc.client.gcInterval=3600000'
+$JAVA_OPTS += '-Dsun.rmi.dgc.server.gcInterval=3600000'
+$JAVA_OPTS += '-Djava.net.preferIPv4Stack=true'
+
+# Warn when resolving remote XML DTDs or schemas.
+$JAVA_OPTS += '-Dorg.jboss.resolver.warning=true'
+
+# Make Byteman classes visible in all module loaders
+# This is necessary to inject Byteman rules into AS7 deployments
+$JAVA_OPTS += "-Djboss.modules.system.pkgs=$JBOSS_MODULES_SYSTEM_PKGS"
```
jamezp: For all of these, shouldn't we test to ensure the `JAVA_OPTS` haven't already been set? https://github.com/wildfly/wildfly-core/blob/master/core-feature-pack/src/main/resources/content/bin/domain.conf#L44-L49

core-feature-pack/src/main/resources/content/bin/common.ps1
dmlloyd: Only ``-secmgr`` and ``-secmgrmodule`` IIRC.  The logic should be the same here as the .sh scripts in any event.

core-feature-pack/src/main/resources/content/bin/domain.ps1
```diff
@@ -4,5 +4,83 @@
...
+         
+
+$PROG_ARGS = @()
+$PROG_ARGS +='-D[Process Controller]' 
+$PROG_ARGS += $PROCESS_CONTROLLER_JAVA_OPTS
+$PROG_ARGS += "-Dorg.jboss.boot.log.file=$JBOSS_LOG_DIR/process-controller.log"
+$PROG_ARGS += "-Dlogging.configuration=file:$JBOSS_CONFIG_DIR/logging.properties"
+$PROG_ARGS += "-Djboss.home.dir=$JBOSS_HOME"
+$PROG_ARGS += "-jar"
+$PROG_ARGS += "$JBOSS_HOME\jboss-modules.jar"
```
jamezp: It probably doesn't matter, but everywhere else we use the `\` delimiter. Should it be that here too?

core-feature-pack/src/main/resources/content/bin/standalone.conf.ps1
```diff
@@ -62,7 +59,7 @@ $JAVA_OPTS += '-Dorg.jboss.resolver.warning=true'
 
 # Make Byteman classes visible in all module loaders
 # This is necessary to inject Byteman rules into AS7 deployments
-$JAVA_OPTS += '-Djboss.modules.system.pkgs=org.jboss.byteman'
+$JAVA_OPTS += "-Djboss.modules.system.pkgs=$JBOSS_MODULES_SYSTEM_PKGS"
```
jamezp: Again we should test the `JAVA_OPTS` before just setting them.

core-feature-pack/src/main/resources/content/bin/common.ps1
jamezp: Oops, it is the same my bad. It errors if it's set.

core-feature-pack/src/main/resources/content/bin/common.ps1
ctomc: this means that default is false if not provided as parameter to function.
think of it as optional parameter with default value

core-feature-pack/src/main/resources/content/bin/common.ps1
ctomc: yeah it should be, just totally forgot about this option.

core-feature-pack/src/main/resources/content/bin/add-user.ps1
ctomc: method Get-Java-Arguments takes it as optional parameter, to build the all java arguments and return it as array to be that used to start the java executable.
Current version of add user script requires that we use special properties file (similar for jboss-cli script) so some special log messages can be displayed.

If parameter is not specified it defaults to our default logging.properties

core-feature-pack/src/main/resources/content/bin/common.ps1
ctomc: Ups, that was copy paste from some local script is use for git command line completion.
Forgot to update the message.

core-feature-pack/src/main/resources/content/bin/common.ps1
ctomc: this just sets global variable $global:DEBUG to true, if we use it or not is up to script iteself.
standalone.ps1 that variable to enable debugging.
Currently also domain one does, which doesn't hurt but could be removed if not needed.

core-feature-pack/src/main/resources/content/bin/common.ps1
ctomc: it is a function, that is only called when needed, basically just domain.ps1 calls it.

core-feature-pack/src/main/resources/content/bin/common.ps1
ctomc: as it is a function and it is defined in common place, it doesn't make sense to have multiple include files.

core-feature-pack/src/main/resources/content/bin/common.ps1
ctomc: this is function call, and yes it is somewhat counter intuitive, but what is does is by default sets the standalone variables, which are then overridden by call to Set-Global-Variables-Domain in domain.ps1 

by default all scrips behave as standalone ones, and set all variables for standalone server.
domain script than just overrides the stuff that is different for domain.

Basically try to do as much as possible in single place and only override/redefine stuff that is different.

core-feature-pack/src/main/resources/content/bin/domain.ps1
ctomc: that should be unified for sure.

core-feature-pack/src/main/resources/content/bin/standalone.conf.ps1
ctomc: the PRESERVE_JAVA_OPTS  conf option is what controls this.
but looking at code again, it looks it is not behaving 100% as expected. 
I will need to rework this.

marekkopecky:
@bstansberry There is probably some issue with vault.ps1. I need some time for investigation ...

core-feature-pack/src/main/resources/content/bin/common.ps1
ctomc: it is used only for standalone servers. domain.ps1 has completely custom parameter construction.

core-feature-pack/src/main/resources/content/bin/common.ps1
ctomc: just sets the global variable, if the script uses it, it is up to the script itself.
Also i don't see any reason why for example domain shouldn't be started with security manager.

core-feature-pack/src/main/resources/content/bin/common.ps1
ctomc: not really required both work just fine. ProcessBuilder is bit more anal about it

core-feature-pack/src/main/resources/content/bin/domain.conf.ps1
ctomc: yes but in this case $JAVA_OPTS is a script variable not environment variable.
that would be $env:JAVA_OPTS we handle that in main script

ctomc:
@marekkopecky Just pushed updated version, there was some combination of vault code requiring env variable JBOSS_HOME to be exported and ps v2 issue.

This should all be fixed now

core-feature-pack/src/main/resources/content/bin/common.ps1
jamezp: Can the function not be in `standalone.ps1`?

marekkopecky:
@bstansberry +1. I finished some base tests. This merge request seems ok to me.

core-feature-pack/src/main/resources/content/bin/common.ps1
```diff
@@ -0,0 +1,279 @@
...
+  }
+  return $res
+}
+
+Function Display-Environment { 
+$JAVA_OPTS = Get-Java-Opts
+# Display our environment
+Write-Host "================================================================================="
+Write-Host ""
+Write-Host "  WildFly Bootstrap Environment"
```
bstansberry: s/WildFly/JBoss

1) Consistent with .sh
2) Does not require editing every time we create an EAP branch

ctomc: I am tempted to change it everywhere with generic "Application Server" without the name....

core-feature-pack/src/main/resources/content/bin/common.ps1
ctomc: sure it can, but all scripts except domain.ps1 are "standalone" scripts

core-feature-pack/src/main/resources/content/bin/standalone.conf.ps1
ctomc: I've reworked how all this works, so it is not as confusing as it was, and works much better when tested.

core-feature-pack/src/main/resources/content/bin/standalone.sh
```diff
@@ -285,7 +285,7 @@ fi
 # Display our environment
 echo "========================================================================="
 echo ""
-echo "  JBoss Bootstrap Environment"
+echo "  WildFly Bootstrap Environment"
```
jamezp: This needs to be changed back to `JBoss`.

marekkopecky:
@ctomc  "sun.rmi.dgc.client.gcInterval", "org.jboss.resolver.warning" and "sun.rmi.dgc.client.gcInterval":
Why are these properties only in powershell scripts? These properties are not in ".sh" or ".bat" scripts.

```
[bin]$ grep  -Ri sun.rmi.dgc.client.gcInterval
domain.conf.ps1:$JAVA_OPTS += '-Dsun.rmi.dgc.client.gcInterval=3600000'
standalone.conf.ps1:$JAVA_OPTS += '-Dsun.rmi.dgc.client.gcInterval=3600000'
[bin]$ grep  -Ri org.jboss.resolver.warning
domain.conf.ps1:$JAVA_OPTS += '-Dorg.jboss.resolver.warning=true'
standalone.conf.ps1:$JAVA_OPTS += '-Dorg.jboss.resolver.warning=true'
[bin]$ grep  -Ri sun.rmi.dgc.client.gcInterval
domain.conf.ps1:$JAVA_OPTS += '-Dsun.rmi.dgc.client.gcInterval=3600000'
standalone.conf.ps1:$JAVA_OPTS += '-Dsun.rmi.dgc.client.gcInterval=3600000'
[bin]$ 
```

marekkopecky:
@ctomc Do you prepare also "Enabling Garbage Collection Logging by Default" in power shell scripts? This feature is not in wildfly, but this feature is currently in EAP (.bat and .sh scripts), it was added by productization ( https://issues.jboss.org/browse/JBEAP-151 ). This feature is RFE in EAP6 and should not be dropped in EAP7 scripts.

This feature was added to EAP 6: https://issues.jboss.org/browse/EAP6-121

ctomc:
@marekkopecky the GC stuff is being in works, first i want to get what we have in, so we don't miss it for WF10. 
as for rmi properties, you are right, we don't have them anywhere else anymore.
They ware default for long time, see https://github.com/wildfly/wildfly/blob/7.1/build/src/main/resources/bin/standalone.conf#L50 as example
I will remove them from ps1 as well if no other scripts has them.

ctomc:
system props commented out / removed

PR #1119, opened 2015-09-25
[WFCORE-1007] Warnings about missing notification descriptions

jmesnil:
check that the notification is described by the resource registry at
the time emit(Notification) is called and buffer the eventual
warnings. These warnings will be logged only at the end of the operation
execution if it actually successful.

JIRA: https://issues.jboss.org/browse/WFCORE-1007


jmesnil:
I have manually checked that calling the `:migrate` operation on the legacy messaging subsystem no longer generate these spurious `WFLYCTL0357: Notification of type resource-removed is not described for the resource ...` warnings

PR #1123, opened 2015-09-25
[WFCORE-1015] init jboss cli jconsole plugin with the remote mbean co…

bmaxwell:
…nnection that is established, so it can load

jira: https://issues.jboss.org/browse/WFCORE-1015

### DIFF
--- a/cli/src/main/java/org/jboss/as/cli/gui/JConsoleCLIPlugin.java
+++ b/cli/src/main/java/org/jboss/as/cli/gui/JConsoleCLIPlugin.java
@@ -119,6 +119,7 @@ public class JConsoleCLIPlugin extends JConsolePlugin {
         if (mbeanServerConn instanceof RemotingMBeanServerConnection) {
             final CommandContext cmdCtx = CommandContextFactory.getInstance().newCommandContext();
             isConnected = connectUsingRemoting(cmdCtx, (RemotingMBeanServerConnection)mbeanServerConn);
+            init(cmdCtx);
         } else {
             //show dialog
             dialog.start();


cli/src/main/java/org/jboss/as/cli/gui/JConsoleCLIPlugin.java
```diff
@@ -119,6 +119,7 @@ private void connect() throws Exception {
         if (mbeanServerConn instanceof RemotingMBeanServerConnection) {
             final CommandContext cmdCtx = CommandContextFactory.getInstance().newCommandContext();
             isConnected = connectUsingRemoting(cmdCtx, (RemotingMBeanServerConnection)mbeanServerConn);
+            init(cmdCtx);
```
bstansberry: It doesn't seem right to ignore the result of L121 here. The init(cmdCtx) call will end up setting isConnected = true, but if L121 returns 'false' the plugin isn't actually connected.

I haven't looked carefully, but it seems like it should be more like this:

```
if (connectUsingRemoting(cmdCtx, (RemotingMBeanServerConnection)mbeanServerConn)) {
    init(cmdCtx);
} else {
   doSomethingButBrianIsntSureWHAT();
}
```

The doSomethingButBrianIsntSureWHAT is probably display some sort of error dialog. This would mean the user has a RemotingMBeanServerConnection to something but a management channel can't be opened on it. This seems highly unlikely and the right thing to do to recover is unclear, so just failing seems reasonable.

bstansberry:
Closing due to inactivity.

PR #1124, opened 2015-09-26
[WFCORE-998] Register a composite op handler on the StoppedServerReso…

bstansberry:
…urce

This implements the second of the two approaches I mention in my comments on the JIRA.


bstansberry:
retest this please

Failure on Full integration - Linux was unrelated

PR #1125, opened 2015-09-28
[WFCORE-987]:  java.lang.StringIndexOutOfBoundsException when empty string is used in a resource address value.

ehsavoie:
Considering empty string as null.

Jira: https://issues.jboss.org/browse/WFCORE-987


ehsavoie:
retest this please

PR #1126, opened 2015-09-28
Replace batch scripts with powershell ones, now they just proxy to ps…

ctomc:
…1 version of scripts

this is optional PR, that should be merged after the new scripts are validated by qe


core-feature-pack/src/main/resources/content/bin/jboss-cli.ps1
```diff
@@ -16,6 +16,7 @@ Function Set-Env {
   Set-Content -Path env:$key -Value $value
 }
 
+<<<<<<< HEAD
```
Ladicek: Is this some special powershell feature? :-)

ctomc: ups, how did that get there :) there was some merge conflict when i was splitting the PR that had two commits that added/fixed powershell scripts and one that is now essentially this.

ctomc: fixed

jamezp:
My fear here is the PowerShell scripts have not been tested and these have. I'm a bit on the paranoid side though especially this close to final :)

ctomc:
That is why the separate PR, which can be merged in later time, preferably after 10.Final that has new ps1. scripts in, is released.

ctomc:
closing for now, will open when time is right

PR #1136, opened 2015-10-01
WFCORE-1024 JUnit version mismatch

ctomc:


bstansberry:
This is fine, but I think excluding junit from full's wildfly-core dependency makes sense as well, if it's easy. Although I can see how doing that might be a real pain as it can come in via so many modules.

ctomc:
we could override(redefine) any dependency in full. But I think junit should be synced if possible. 
And given that release cycle for junit new releases is really slow, it would really come often into upgrade needs.

PR #1137, opened 2015-10-01
WFLY-4959 add a log message at INFO level that SM is enabled

istudens:
This introduces a new INFO log message that Security Manager is enabled.

https://issues.jboss.org/browse/WFLY-4959


bstansberry:
@dmlloyd In a comment on the JIRA you say this should be logged pretty early in the boot. These spots are fairly early, but a fair amount of stuff in the WildFly code will have already run, particularly the command line arg processing that creates the ServerEnvironment and HostControllerEnvironment objects.

This seems ok to me, without thinking hard about it. The message will get logged in the normal case, and in the odd case that some SM related failure prevents boot getting this far it's pretty likely the user will figure out the SM was enabled without the help of this logging.

dmlloyd:
Sounds reasonable.

PR #1140, opened 2015-10-02
[WFCORE-1026] Upgrade jboss-logmanager from 2.0.0.Final to 2.0.3.Final

jamezp:
Contains only a few minor changes. Two bug fixes and two performance changes. https://github.com/jboss-logging/jboss-logmanager/compare/2.0.0.Final...jamezp:2.0.2.Final

### DIFF
--- a/controller/src/main/java/org/jboss/as/controller/audit/SyslogAuditLogHandler.java
+++ b/controller/src/main/java/org/jboss/as/controller/audit/SyslogAuditLogHandler.java
@@ -75,10 +75,13 @@ public class SyslogAuditLogHandler extends AuditLogHandler {
     private volatile TransportErrorManager errorManager;
     private volatile int reconnectTimeout = -1;
     private volatile long lastErrorTime = -1;
+    // This should be guarded by the config lock in the ManagedAuditLoggerImpl
+    private boolean initialized;
 
     public SyslogAuditLogHandler(String name, String formatterName, int maxFailureCount, PathManagerService pathManager) {
         super(name, formatterName, maxFailureCount);
         this.pathManager = pathManager;
+        initialized = false;
     }
 
     public void setHostName(String hostName) {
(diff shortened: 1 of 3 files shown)


kabir:
Retest this please

jamezp:
retest this please

jamezp:
The failing unit test is a timing issue as far as I can tell. retest this please

jamezp:
retest this please

bstansberry:
@jamezp @kabir I think I've found the cause of the org.jboss.as.jmx.rbac.JmxRbacEnabledTestCase.testMonitorNonSensitiveMBeans failure. It's unrelated to this PR. It's a test fixture problem where the TestModelMBean used by the test is not holding a ref to the "listenerWrapper" that the JVM MBeanServer passes into its addNotificationListener method. That allows that listenerWrapper to be gc'd, resulting in the failure of the removeNotificationListener call on the next line.

The test holding a ref to a listener doesn't matter, as it's a "listenerWrapper" object that matters.

Simple fix; I'll send one up.

bstansberry:
@jamezp I'll leave this one to you. The AuditLogSyslogReconnectTestCase failure may be related; may not. I have no clue, so feel free to merge if you are comfortable and @kabir doesn't object.

jamezp:
Perfect, I *may* have a fix now. It seems to be a race in the test as far as I can tell. I'll run this a few times to ensure we don't get a failure.

jamezp:
Just making a general comment so that I remember the status.

There is an issue with the log manager where the reconnect thread will continue to attempt to reconnect even if the handler is closed. I'm reworking the logmanager a bit more which will lead to a 2.0.3.Final release and an upgrade for that.

There's still an underlying issue with the `SyslogAuditLogHandler`. There's potential for the internal handler to fail to set the custom output stream if the socket can't connect. This causes the default `TcpOutputStream` to be used which may attempt to reconnect if a write makes it through. The `SyslogAuditLogHandler` does it's own reconnect algorithm. Regardless a failure to properly construct the internal handler should cause the handler to close or not be usable.

There's still an underlying issue somewhere though with the internal handler writing to either the wrong socket or the `SimpleLogServer` isn't accepting the correct connection. This is what needs to be debugged to determine what the issue is.

jamezp:
@kabir can you review the changes I made to the `SyslogAuditLogHandler`?

PR #1144, opened 2015-10-03
Increase syslog test timeout

kabir:
THis includes @jamezp https://github.com/wildfly/wildfly-core/pull/1140 and increase the timeout since with 1140 on its own we saw some problems


kabir:
This now passed Linux, but failed on Windows so the answer probably isn't that simple

PR #1149, opened 2015-10-05
Minor api improvment for working with capabilites

ctomc:
- takes capabilites defined on resource into account

### DIFF
--- a/controller/src/main/java/org/jboss/as/controller/RestartParentResourceHandlerBase.java
+++ b/controller/src/main/java/org/jboss/as/controller/RestartParentResourceHandlerBase.java
@@ -204,7 +204,7 @@ public abstract class RestartParentResourceHandlerBase implements OperationStepH
         if (parentModel != null && context.revertResourceRestarted(address, this)) {
             try {
                 removeServices(context, serviceName, invalidatedParentModel);
-                recreateParentService(context, address, parentModel, null);
+                recreateParentService(context, address, parentModel);
             } catch (OperationFailedException e) {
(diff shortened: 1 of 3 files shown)


controller/src/main/java/org/jboss/as/controller/RestartParentResourceHandlerBase.java
```diff
@@ -214,7 +214,7 @@ private void recoverServices(final OperationContext context, final ModelNode inv
     private ModelNode getModel(OperationContext ctx, PathAddress address) {
         try {
             Resource resource = ctx.readResourceFromRoot(address);
-            return Resource.Tools.readModel(resource);
+            return Resource.Tools.readModel(resource, 1);
```
bstansberry: What bug does this fix?

controller/src/main/java/org/jboss/as/controller/RestartParentResourceHandlerBase.java
```diff
@@ -223,7 +223,7 @@ private ModelNode getModel(OperationContext ctx, PathAddress address) {
     private ModelNode getOriginalModel(OperationContext ctx, PathAddress address) {
         try {
             Resource resource = ctx.getOriginalRootResource().navigate(address);
-            return Resource.Tools.readModel(resource);
+            return Resource.Tools.readModel(resource, 1);
```
bstansberry: What bug does this fix?

ctomc: Just performance improvement.

PR #1153, opened 2015-10-06
Add asserts that the alias address is different from the target addre…

kabir:
…ss, to guard against things like WFLY-5290


bstansberry:
:+1 once the needed changes in full are in

bstansberry:
retest this please

PR #1154, opened 2015-10-06
[WFCORE-628] Transform the 'query' operation for use on slaves running < WildFly 9

bstansberry:
Legacy slaves will not understand the new 'query' operation added in WF 9 / Core 1.0. This PR fixes this by adding a transformer for legacy kernel version prior to major version 3 (i.e. WF 8 and earlier). The transformer converts the query op to a read-resource and then uses an OperationResultTransformer to do the filter and select work on the DC side that the normal query op does on the remote side.

In the description of WFCORE-628 I mentioned an issue with this task is that the 'query' operation is global, so there would need to be some notion of inherited transformers, with the transformer for this registered that way. It looks like there is already some notion of an inherited transformer (see OperationTransformerEntry.isInherited()) but the hooks to register this transformer that way are not completely in place. To keep this simple, I decided not to deal with that and just did a simple hack to get the transformer in place. See the diff to TransformationTargetImpl.

https://github.com/wildfly/wildfly/pull/8247 is the test part of this work.


bstansberry:
http://brontes.lab.eng.brq.redhat.com/viewLog.html?buildId=74423&tab=buildResultsDiv&buildTypeId=WF_WildFlyCoreIntegrationExperiments is a custom integration run of this in full along with the tests in https://github.com/wildfly/wildfly/pull/8247

PR #1155, opened 2015-10-07
https://issues.jboss.org/browse/WFCORE-877 Redirection (<, <<) do not…

aloubyansky:
… work in CLI

Thanks,
Alexey

### DIFF
--- a/cli/src/main/java/org/jboss/as/cli/impl/CommandContextConfiguration.java
+++ b/cli/src/main/java/org/jboss/as/cli/impl/CommandContextConfiguration.java
@@ -38,7 +38,7 @@ public class CommandContextConfiguration {
     private final boolean disableLocalAuth;
     private final int connectionTimeout;
     private boolean silent;
-    private boolean errorOnInteract;
+    private Boolean errorOnInteract;
 
     private CommandContextConfiguration(String controller, String username, char[] password, String clientBindAddress, boolean disableLocalAuth, boolean initConsole, int connectionTimeout, InputStream consoleInput, OutputStream consoleOutput) {
         this.controller = controller;
@@ -92,7 +92,7 @@ public class CommandContextConfiguration {
         return silent;
     }
 
-    public boolean isErrorOnInteract() {
+    public Boolean isErrorOnInteract() {
         return errorOnInteract;
     }
 
@@ -108,7 +108,7 @@ public class CommandContextConfiguration {
         private int connectionTimeout = -1;
         private boolean disableLocalAuthUnset = true;
         private boolean silent;
-        private boolean errorOnInteract;
+        private Boolean errorOnInteract;
 
         public Builder() {
         }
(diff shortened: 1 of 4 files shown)


bstansberry:
The Full integration - Linux failure seems possibly related as org.jboss.as.test.integration.security.perimeter.CLISecurityTestCase timed out without completion, leading to a cascade of failures:

http://brontes.lab.eng.brq.redhat.com/viewLog.html?buildId=74295&buildTypeId=WildFlyCore_PullRequest_WildFlyCoreFullIntegration&tab=buildLog#_focus=24069

bstansberry:
The Full Integration - Windows run had the same problem.

aloubyansky:
I'll look into that later tonight.

aloubyansky:
The test used pass but masked a problem. Although, I'm not sure the expectation was wrong.

The test expects 'connect' command to fail when the authentication is required. But the CLI is simply asking for the auth info behind the scene and simply keeps waiting while the test doesn't provide any input. Before this PR the CLI would simply throw a NPE due to a bug and that would be considered by the test as an auth failure.

On one hand, the PR fixes the NPE, on the other hand, I'm not sure the expectation in the test is wrong since the CLI API client hasn't configured any input source for the CLI.

We have --error-on-interact cli script flag now which is off by default. So, the CLI by default launched in non-interactive mode will still ask for the auth info. Which makes sense for the script. But I am not sure it makes sense for an API client. Perhaps, for an API client the default has to be the opposite, i.e. unless the input stream is provided by the client, asking for input should fail.

cli/src/main/java/org/jboss/as/cli/impl/CommandContextConfiguration.java
```diff
@@ -145,6 +145,10 @@ public Builder setClientBindAddress(String clientBindAddress) {
 
         public Builder setConsoleInput(InputStream consoleInput) {
             this.consoleInput = consoleInput;
+            if(consoleInput != null) {
```
bstansberry: Should this only happen if this.errorOnInteract == null? So don't override whatever was previously set up via setErrorOnInteract().

bstansberry:
@aloubyansky I'm not clear about your last comment. What you are saying sounds reasonable, but did you mean that you implemented that in this patch (the errorOnInteract stuff looks like it) or is this some future work you are thinking about?

The test passes now so something happened. :)

aloubyansky:
I updated the PR. Fixed the NPE and implemented the change I described in the previous comment. I.e. now when the CLI is launched using the script in non-interactive mode in case the authentication is required, it'll ask the user for the input (which means error-on-interact=false). But if the client is using the CLI API the error-on-interact will be true as the default. I decided to change it because that will force the client to implement the authentication interaction, otherwise it'll simply hang forever, which will be confusing.
I admit, I didn't discuss it with anybody. If you think it's worth discussion, let's postpone merging.
Sorry for slow responses, I've been on PTO since yesterday.

bstansberry:
NP re: the slow responses (didn't seem slow!) and thanks for dealing with this. This seems fine to me.

bstansberry:
I'm not going to hold this up over my comment re: f this.errorOnInteract == null. If you think that's a valid point, please send up a PR some time.

cli/src/main/java/org/jboss/as/cli/impl/CommandContextConfiguration.java
aloubyansky: Good catch! Thanks.

PR #1157, opened 2015-10-07
[WFCORE-1038] Remove required parameter validation as that is handled server side.

darranl:
This reverts the validation that was added in commit e594336c25290fca12eff08be562b5c1e7b26e17

### DIFF
--- a/cli/src/main/java/org/jboss/as/cli/handlers/OperationRequestHandler.java
+++ b/cli/src/main/java/org/jboss/as/cli/handlers/OperationRequestHandler.java
@@ -173,17 +173,12 @@ public class OperationRequestHandler implements CommandHandler, OperationCommand
             throw new CommandFormatException("Failed to perform " + Util.READ_OPERATION_DESCRIPTION + " to validate the request: result is not available.");
         }
         final ModelNode result = outcome.get(Util.RESULT);
-        final Set<String> definedProps = result.get(Util.REQUEST_PROPERTIES).keys();
+        final Set<String> definedProps = result.hasDefined(Util.REQUEST_PROPERTIES) ? result.get(Util.REQUEST_PROPERTIES).keys() : Collections.emptySet();
         if(definedProps.isEmpty()) {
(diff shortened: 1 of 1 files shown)


cli/src/main/java/org/jboss/as/cli/handlers/OperationRequestHandler.java
```diff
@@ -174,7 +173,8 @@ private ModelNode validateRequest(CommandContext ctx, ModelNode request) throws
             throw new CommandFormatException("Failed to perform " + Util.READ_OPERATION_DESCRIPTION + " to validate the request: result is not available.");
         }
         final ModelNode result = outcome.get(Util.RESULT);
-        if(!result.hasDefined(Util.REQUEST_PROPERTIES)) {
+        final Set<String> definedProps = result.get(Util.REQUEST_PROPERTIES).keys();
```
bstansberry: This will fail if(!result.hasDefined(Util.REQUEST_PROPERTIES)). An undefined node throws IAE if keys() is called.

cli/src/main/java/org/jboss/as/cli/handlers/OperationRequestHandler.java
```diff
@@ -184,18 +184,13 @@ private ModelNode validateRequest(CommandContext ctx, ModelNode request) throws
                     throw new CommandFormatException("Operation '" + operationName + "' does not expect any property.");
                 }
             }
```
bstansberry: BTW, if current master L177 is reverted back to what it was before to fix my comment there, then the block whose beginning isn't shown in this diff but which ends here is pointless. That is, it's pointless in the current code. It just repeats an if test that has already evaluated false at L177.

darranl: I will double check but this commit should have reverted everything back to how it was before

bstansberry: Yeah, it's wrong in master. Sorry, I wasn't clear. I only mentioned it in case you felt like deleting it.

PR #1158, opened 2015-10-07
[WFCORE-997] Override the optional createSocket methods and pass these calls onto the delegate.

darranl:
This enables the use of standard timeout settings for outbound LDAPS connections.


dmlloyd:
+1 fwiw

PR #1161, opened 2015-10-08
WFCORE-1043: Show error message when accessing http management consol…

rarguello:
…e of a slave host controller


bstansberry:
@luck3y can you verify this patch has the effect it says? The patch looks fine to me.

bstansberry:
@rarguello -- we need to ask a one-time housekeeping question; we forgot to do this with the system-d patch:

luck3y:
@bstansberry yep, taking a look.

luck3y:
Looks fine to me, I get the following message accessing the slave:

Welcome to WildFly

Your WildFly Application Server is running.

You may not access the admin console on this node, all access must come via the master host controller.

rarguello:
I agree with the terms of the MIT License for these contributions.
However, doesn't the HttpManagementAddHandler.java header says it is licensed under the LGPL license?

bstansberry:
Thanks, Ricardo.

Yes the header references LGPL We ask this question of contributors because in the past, contributor agreements applicable to WildFly have generally limited outbound licensing of contributions under LGPL (or  GPL) only. We would now like to have the flexibility to consider other outbound open source licensing options for components of WildFly that may be incompatible with or less restrictive than LGPL  licensing.

An example of such a change would be changing the licensing to use the Apache Software License.

PR #1167, opened 2015-10-09
[WFCORE-1047] Don't unnecessarily read the model to rollback deployme…

bstansberry:
…nt ops

Also some general cleanup, renaming of vars to be more clear


bstansberry:
retest this please

failure was unrelated; see https://github.com/wildfly/wildfly-core/pull/1168

PR #1171, opened 2015-10-15
ModelController-client needs to be using JDK8 source level, to prepar…

ctomc:
…e it for elytron.

### DIFF
--- a/controller-client/pom.xml
+++ b/controller-client/pom.xml
@@ -36,13 +36,6 @@
 
     <name>WildFly: Controller Client</name>
 
-    <properties>
-        <!-- maven-compiler-plugin -->
-        <!-- Keep the management client libs on JDK 6 -->
-        <maven.compiler.target>1.6</maven.compiler.target>
-        <maven.compiler.source>1.6</maven.compiler.source>
-    </properties>
-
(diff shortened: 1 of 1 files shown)


bstansberry:
This isn't necessary for core 2.0.0, correct?

ctomc:
this should to be in for EAP7.0, whatever core release will that use.
Otherwise we will have problems with 7.1 moving it to 8 because of elytron needs which would cause another set of problems.

bstansberry:
David tells me the Elytron work will not be negatively impacted by deferring this to 7.1 so I don't want to make this change at this late date based on Elytron needs.

ctomc:
Only thing I am arguing is that for 7.0 we will have requirement for modelcontroller client usage to be JDK6, and in 7.1 it will be bumped to 8. which might break some clients that will be using it when they upgrade.
Argument boils down to, isn't it better to have such change in initial major release so we only break users client code once instead of twice.
I am sure that number of users that might be affected by this will be minimal but it will still exist.

dmlloyd:
It's a client library; if the change is a problem for them, they will continue to use the 7.0 version.

PR #1175, opened 2015-10-20
WFCORE-784 [PATCH] fix 1 bug in bin/standalone.sh

ctomc:


ctomc:
completely unrelated

core-feature-pack/src/main/resources/content/bin/standalone.sh
```diff
@@ -82,7 +82,7 @@ if $cygwin ; then
 fi
 
 # Setup JBOSS_HOME
-RESOLVED_JBOSS_HOME=`cd "$DIRNAME/.."; pwd`
+RESOLVED_JBOSS_HOME=`cd "$DIRNAME/.." >/dev/null; pwd`
```
bstansberry: What does this change do to fix the problem?

Not including any output from 'cd' in the value of RESOLVED_JBOSS_HOME makes sense, but I'm not clear on what CDPATH has to do with this, other than that 'cd' might produce output if CDPATH means it can't change to the directory correctly. But then that's an error condition too.

ctomc: if CDPATH is set, every cd command echoes current directory, so you know that CDPATH is effect.
Which can cause that RESOLVED_JBOSS_HOME can include wrong value.

PR #1182, opened 2015-10-20
[WFCORE-1062]:  Deployments from the domain/data/content folder goes missing after few minutes if the Host controller is started with --backup option.

ehsavoie:
When the HC is in backup mode it needs to reference every content it downloads from the DC otherwise those would get deleted by the GC.
Thus we provide the ContentRepository of the HC in the DeploymentAddHandler with the RemoteFileRepository when we have a backup.

Jira: https://issues.jboss.org/browse/WFCORE-1062

### DIFF
--- a/deployment-repository/src/main/java/org/jboss/as/repository/ContentRepository.java
+++ b/deployment-repository/src/main/java/org/jboss/as/repository/ContentRepository.java
@@ -21,6 +21,10 @@
  */
 package org.jboss.as.repository;
 
+import static java.lang.Long.getLong;
+import static java.lang.System.getSecurityManager;
+import static java.security.AccessController.doPrivileged;
+
 import java.io.BufferedInputStream;
 import java.io.Closeable;
 import java.io.File;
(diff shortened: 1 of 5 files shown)


deployment-repository/src/main/java/org/jboss/as/repository/ContentRepository.java
```diff
@@ -53,7 +58,7 @@
  * @author John Bailey
  */
 public interface ContentRepository {
-
+    String UNSUPPORTED_PROPERTY = "org.wildfly.unsupported.content.repository.obscolescence";
```
bstansberry: I'd rather hard code the string in multiple places than make this part of this interface.

bstansberry: BTW my vague concern here is code like wildfly-swarm or torquebox treating that string as some sort of dependable API. Just putting @Deprecated on it may suffice.

bstansberry:
Nit about the constant aside, this looks fine.

For WildFly Core 3 though I'd like to look at tightening up DeploymentAddHandler a bit. Make it less vulnerable to future bugs. For example I just experimented with running the testsuite with the following commit and it passed:

https://github.com/bstansberry/wildfly-core/commit/6cb1e22516dd5ad218ea94e8b3be9a9ecad2943f

So perhaps we can simplify some of the if checks in this class by knowing that contentRepository will not be null in the master case. Things like that. Not for core 2.0 though; at this point we should make the minimal changes possible, which you've done here.

PR #1185, opened 2015-10-22
[WFCORE-1071] Clean up the username when manually parsed from the properties file to compensate for escaped characters.

darranl:
…

### DIFF
--- a/domain-management/src/main/java/org/jboss/as/domain/management/security/PropertiesFileLoader.java
+++ b/domain-management/src/main/java/org/jboss/as/domain/management/security/PropertiesFileLoader.java
@@ -205,7 +205,7 @@ public class PropertiesFileLoader {
                 } else {
                     Matcher matcher = PROPERTY_PATTERN.matcher(trimmed);
                     if (matcher.matches()) {
-                        final String key = matcher.group(1);
+                        final String key = cleanKey(matcher.group(1));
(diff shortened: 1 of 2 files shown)


domain-management/src/main/java/org/jboss/as/domain/management/security/PropertiesFileLoader.java
```diff
@@ -225,6 +225,41 @@ public synchronized void persistProperties() throws IOException {
...
     }
 
+    protected String cleanKey(final String key) {
+        char[] keyChars = key.toCharArray();
+        char[] cleaned = new char[keyChars.length];
+        int keyPos = 0;
+        int cleanedPos = 0;
+
+        while (keyPos < keyChars.length) {
+            if (keyChars[keyPos++] == '\\') {
```
bstansberry: Should this fail here if keyPos == keyChars.length?

IOW what if the user edits the file manulally and doesn't escape a final '\'?

foo\=group-a

instead of the correct

foo\\=group-a

Currently L236 will fail with an IndexOutOfBoundException instead of an error message explaining the issue.

bstansberry: Never mind.

I was going to ask a similar question about the default block at L250; whether it should fail if the next char after \ isn't \ or =.

But both here and at L250 looking at how java.util.Properties works it's basically the same as what you have here, so it's ok. I don't see a need to fail more nicely than the jdk does.

PR #1189, opened 2015-10-26
WFLY-5571 Intermittent failure in ModelPersistenceTestCase.testSimple…

ctomc:
…Operation

It should fix the problem, in worst case it will give us some better info what is going on.

Also moved test from full to core.


ctomc:
unrelated failure

ctomc:
retest this please

PR #1190, opened 2015-10-26
[WFCORE-1068] Add a utility to throw a standard failure when manageme…

bstansberry:
…nt authorization is denied

### DIFF
--- a/controller/src/main/java/org/jboss/as/controller/access/AuthorizationResult.java
+++ b/controller/src/main/java/org/jboss/as/controller/access/AuthorizationResult.java
@@ -22,6 +22,12 @@
 
 package org.jboss.as.controller.access;
 
+import static org.jboss.as.controller.descriptions.ModelDescriptionConstants.OP;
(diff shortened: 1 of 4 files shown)


controller/src/main/java/org/jboss/as/controller/access/AuthorizationResult.java
```diff
@@ -46,20 +52,75 @@
...
     public ModelNode getExplanation() {
         return explanation;
     }
+
+    /**
+     * Utility method to throw a standard failure if {@link #getDecision()} is
+     * {@link org.jboss.as.controller.access.AuthorizationResult.Decision#DENY}.
+     * <p>
+     * This variant extracts the target address from the {@code address} field in the {@code operation} param
+     * and then calls the {@link #failIfDenied(ModelNode, PathAddress) overloaded variant}.
```
jamezp: Minor note, but the `{@link #failIfDenied(ModelNode, PathAddress) overloaded variant}` should probably be `{@linkplain #failIfDenied(ModelNode, PathAddress) overloaded variant}`.

PR #1195, opened 2015-10-27
WFCORE-601 multiple marker files can cause server boot to fail

stuartwdouglas:


bstansberry:
The fix seems reasonable, although I haven't looked at all the tiny details to see if I spot some reason why the test would fail. Perhaps Emmanuel will see something.

PR #1196, opened 2015-10-27
WFCORE-1052 Ability to specify config when reloading

kabir:
https://issues.jboss.org/browse/WFCORE-1052
https://issues.jboss.org/browse/WFCORE-1078

@bstansberry I have opened this for 2.0.0.CR9, although 1052 is really a new feature, so this is probably too late. If that is the case, I will reschedule for 3.0.0


bstansberry:
It's too late. We're locking stuff down at this point. Even bugs need JBEAP + CDW.

cli/src/main/resources/help/reload.txt
```diff
@@ -76,3 +78,18 @@ ARGUMENTS
                  original configuration version; if null or true the current
                  runtime version of the model is used.
                  If not present, true value is assumed.
+
+ --domain-config  - optional, allowed only in the domain mode. Used to specify
+                 a new file name for the domain config. The same rules as for
+                 the --domain-config flag used on the command line when starting
+                 the server apply.
```
bstansberry: s/server/host

cli/src/main/resources/help/reload.txt
```diff
@@ -76,3 +78,18 @@ ARGUMENTS
...
+
+ --host-config   - optional, allowed only in the domain mode. Used to specify
+                 a new file name for the host config. The same rules as for
+                 the --host-config flag used on the command line when starting
+                 the server apply.
```
bstansberry: s/server/host

cli/src/main/resources/help/reload.txt
```diff
@@ -2,7 +2,7 @@ SYNOPSIS
 
    Standalone mode:
 
-      reload [--admin-only=true|false] [--use-current-server-config=true|false]
+      reload [--admin-only=true|false] [--use-current-server-config=true|false] [--server-config=new_server_config_file_name]
```
bstansberry: I think this needs a line break. These are meant to fit within a certain number of columns per line.

controller/src/main/java/org/jboss/as/controller/RunningModeControl.java
```diff
@@ -32,6 +32,7 @@
     private volatile RunningMode runningMode;
     private volatile boolean reloaded;
     private volatile boolean useCurrentConfig;
+    private String newBootFileName;
```
bstansberry: volatile?

I haven't really looked into whether that's needed; I just see that other stuff is volatile.

testsuite/shared/src/main/java/org/jboss/as/test/shared/ServerReload.java
```diff
@@ -0,0 +1,138 @@
...
+import org.jboss.as.controller.client.ModelControllerClient;
+import org.jboss.as.controller.client.helpers.ClientConstants;
+import org.jboss.dmr.ModelNode;
+import org.junit.Assert;
+import org.xnio.IoUtils;
+
+/**
+ * @author Stuart Douglas
+ */
+public class ServerReload {
```
bstansberry: #1482 also brings this one in, but in a different package; we'll have a race!

PR #1198, opened 2015-10-27
Don't include whole server in shaded wildfly-cli-client

ctomc:


bstansberry:
Incorporated in #1201. Thanks for finding and sorting this!

PR #1200, opened 2015-10-27
WFCORE-993 fix embedded HC crash 

luck3y:
Caused by reload dropping --admin-only

### DIFF
--- a/cli/src/main/java/org/jboss/as/cli/handlers/ReloadHandler.java
+++ b/cli/src/main/java/org/jboss/as/cli/handlers/ReloadHandler.java
@@ -167,7 +167,19 @@ public class ReloadHandler extends BaseOperationCommand {
 
     private void doHandleEmbedded(CommandContext ctx, ModelControllerClient client) throws CommandLineException {
 
+        assert(embeddedServerRef != null);
+        assert(embeddedServerRef.get() != null);
+
         final ModelNode op = this.buildRequestWithoutHeaders(ctx);
+        if (embeddedServerRef.get().isHostController()) {
+            // WFCORE-938
+            // for embedded-hc, we require --admin-only=true to be passed until the EHC supports --admin-only=false
+            final ParsedCommandLine args = ctx.getParsedCommandLine();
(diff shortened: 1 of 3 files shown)


cli/src/main/java/org/jboss/as/cli/handlers/ReloadHandler.java
```diff
@@ -167,7 +167,17 @@ protected void doHandle(CommandContext ctx) throws CommandLineException {
 
     private void doHandleEmbedded(CommandContext ctx, ModelControllerClient client) throws CommandLineException {
 
+        assert(embeddedServerRef != null);
+        assert(embeddedServerRef.get() != null);
+
         final ModelNode op = this.buildRequestWithoutHeaders(ctx);
+        if (embeddedServerRef.get().isHostController()) {
```
bstansberry: WDYT about adding some validation in this block?

```
final ParsedCommandLine args = ctx.getParsedCommandLine();
if (this.adminOnly.isPresent(args) && "FALSE".equalsIgnoreCase(this.adminOnly.getValue(args)) {
 throw someExceptionCauseWeCantDoWhatTheyRequested;
}
```

Bonus is overriding the canAppearNext() method of adminOnly so it returns false if embeddedServerRef.get() != null && embeddedServerRef.get().isHostController(). That will remove that param from tab completion.

luck3y: Sure, good idea I'll add.

bstansberry: The docs on the --admin-only param in reload.txt should explain all this too; param won't be offered, value will always be true.

luck3y:
Note this is also disabling /host=master:reload() due to that being able to force admin-only=false also. I can probably alter that if we really want /host=master:reload() to work in embedded, but it seems to be problematic (especially in .cli scripts, as it returns immediatly before the HC is actually restarted).

luck3y:
Forgot docs, updating now.

luck3y:
Hrm, thats not quite what I wanted ;) Fixing.

PR #1202, opened 2015-10-28
[WFLY-5589] return proper socket address for unnamed socket registrat…

soul2zimate:
…ion / unregistration since super class getLocalSocketAddress() returns null if socket is closed first.

https://issues.jboss.org/browse/WFCORE-1081


bstansberry:
Thanks! Can you change the commit message to reference WFCORE-1081 instead of WFLY-5589. This is a bug in WildFly Core, not full WildFly.

soul2zimate:
sure, rebased jira reference to WFCORE-1081 in commit message

PR #1203, opened 2015-10-28
Upgrade client libraries to require Java 7

jamezp:


ctomc:
https://github.com/wildfly/wildfly-core/pull/1171

bstansberry:
We have to go to JDK 7 because some of the dependencies require 7. But we're making the minimum possible move.

PR #1206, opened 2015-10-28
[WFCORE-1082] Fix attribute name for the main-class element in the ge…

jamezp:
…nerated module.xml file. Also added a schema validation test for the generated XML file.

### DIFF
--- a/cli/pom.xml
+++ b/cli/pom.xml
@@ -72,6 +72,33 @@
                     </execution>
                 </executions>
             </plugin>
+            <plugin>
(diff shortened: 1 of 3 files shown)


jamezp:
@aloubyansky Are you okay with adding this test? I'll just assume the one line fix is okay :)

aloubyansky:
Sure, assuming it passes.

jamezp:
Excellent, thanks. It should be passing now. I had forgotten to close and flush the writer :)

PR #1215, opened 2015-11-03
[WFCORE-1093] TempFileProviderService threads consume high CPU

aogburn:
https://issues.jboss.org/browse/WFCORE-1093

### DIFF
--- a/server/src/main/java/org/jboss/as/server/deployment/module/TempFileProviderService.java
+++ b/server/src/main/java/org/jboss/as/server/deployment/module/TempFileProviderService.java
@@ -24,7 +24,8 @@ package org.jboss.as.server.deployment.module;
 
 import java.io.IOException;
 import java.security.PrivilegedAction;
-import java.util.concurrent.Executors;
+import java.util.concurrent.ScheduledThreadPoolExecutor;
+import java.util.concurrent.TimeUnit;
 
 import org.jboss.as.server.logging.ServerLogger;
 import org.jboss.msc.service.Service;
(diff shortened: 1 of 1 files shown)


bstansberry:
For the EAP 6 PR to be merged, a pull request against https://github.com/jbossas/wildfly-core-eap, the 2.x branch, will be needed. Otherwise EAP 7 will regress.

Normally we would require a JBEAP JIRA as well (since merges to the EAP 7 branches can't happen without) but I suppose the EAP 6 BZ will suffice, once it is acked.

Once we start work on WildFly 11 / WildFly Core 3 then this can go in here (assuming it's ok technically) regardless of the EAP branches. But currently we are not merging fixes unless they are going into the EAP 7 branch as well.

server/src/main/java/org/jboss/as/server/deployment/module/TempFileProviderService.java
```diff
@@ -54,7 +56,9 @@ public JBossThreadFactory run() {
                    return new JBossThreadFactory(new ThreadGroup("TempFileProviderService-temp-threads"), Boolean.TRUE, null, "%G - %t", null, null);
                }
            });
-           PROVIDER = TempFileProvider.create("deployment", Executors.newScheduledThreadPool(0, threadFactory), true);
+           ScheduledThreadPoolExecutor ex = (ScheduledThreadPoolExecutor) Executors.newScheduledThreadPool(0, threadFactory);
```
bstansberry: Please use "new ScheduledThreadPoolExecutor(0, threadFactory)".

aogburn:
Thanks, Brian.  Changed the fix per your suggestion and created an EAP 7 PR:

https://github.com/jbossas/wildfly-core-eap/pull/43

PR #1219, opened 2015-11-05
WFCORE-1112 use logger.debugf instead of manual string concatenation

Gordiychuk:
Fix for ticket https://issues.jboss.org/browse/WFCORE-1112
org.jboss.logging.BasicLogger#debugf allow not replace template if debug level not available


jamezp:
this is okay to test

jamezp:
this is ok to test

bstansberry:
:+1: 

Thanks!

PR #1224, opened 2015-11-06
Upgrade to vfs 3.2.10.Final to address BZ#1223081

ctomc:
https://bugzilla.redhat.com/show_bug.cgi?id=1223081


bstansberry:
@ctomc Can you send in patches to core and core-eap to move to 3.2.10? That definitely has to be done, as EAP 6 is using 3.2.10 and WF 10 / EAP 7 can't regress. The move to 3.2.11 may or may not get approved, so it should not block getting to 3.2.10.

ctomc:
@bstansberry updated this PR to point to 3.2.10 instead of 3.2.11

bstansberry:
retest this please

MinimumPermissionsTestCase.testPropertyPerm did not reload in time but I don't see how this upgrade could be related.

PR #1226, opened 2015-11-06
[WFCORE-1008]: Treat ":shutdown" requests similarly to ":reload" in the server-side operation handler.

ehsavoie:
Shutdown and reload are now treated in the same 'way', sending a prepared response before the services are stopped effectively.

Jira: https://issues.jboss.org/browse/WFCORE-1008

### DIFF
--- a/controller/src/main/java/org/jboss/as/controller/remote/ModelControllerClientOperationHandler.java
+++ b/controller/src/main/java/org/jboss/as/controller/remote/ModelControllerClientOperationHandler.java
@@ -33,7 +33,9 @@ import static org.jboss.as.controller.descriptions.ModelDescriptionConstants.OP;
 import static org.jboss.as.controller.descriptions.ModelDescriptionConstants.OPERATION_HEADERS;
 import static org.jboss.as.controller.descriptions.ModelDescriptionConstants.OP_ADDR;
 import static org.jboss.as.controller.descriptions.ModelDescriptionConstants.OUTCOME;
+import static org.jboss.as.controller.descriptions.ModelDescriptionConstants.RELOAD;
 import static org.jboss.as.controller.descriptions.ModelDescriptionConstants.RESULT;
+import static org.jboss.as.controller.descriptions.ModelDescriptionConstants.SHUTDOWN;
 import static org.jboss.as.controller.descriptions.ModelDescriptionConstants.SUCCESS;
 import static org.jboss.as.controller.descriptions.ModelDescriptionConstants.SYNC_REMOVED_FOR_READD;
 import static org.jboss.as.controller.descriptions.ModelDescriptionConstants.USER;
@@ -45,6 +47,9 @@ import java.io.IOException;
 import java.security.PrivilegedActionException;
 import java.security.PrivilegedExceptionAction;
 import java.util.Collections;
+import java.util.Arrays;
+import java.util.HashSet;
+import java.util.Set;
 import java.util.concurrent.CountDownLatch;
 import java.util.concurrent.Executor;
 import java.util.concurrent.ExecutorService;
@@ -79,6 +84,7 @@ import org.jboss.dmr.ModelNode;
  */
 public class ModelControllerClientOperationHandler implements ManagementRequestHandlerFactory {
 
+    private static final Set<String> PREPARED_RESPONSE_OPERATIONS = new HashSet<>(Arrays.asList(RELOAD, SHUTDOWN));
     private final ModelController controller;
 
     private final ManagementChannelAssociation channelAssociation;
@@ -128,7 +134,6 @@ public class ModelControllerClientOperationHandler implements ManagementRequestH
     }
 
     class ExecuteRequestHandler implements ManagementRequestHandler<ModelNode, Void> {
-
         @Override
         public void handleRequest(final DataInput input, final ActiveOperation.ResultHandler<ModelNode> resultHandler,
                                   final ManagementRequestContext<Void> context) throws IOException {
@@ -239,7 +244,7 @@ public class ModelControllerClientOperationHandler implements ManagementRequestH
             final String op = operation.get(OP).asString();
             final int size = address.size();
             if (size == 0) {
-                if ("reload".equals(op)) {
+                if (PREPARED_RESPONSE_OPERATIONS.contains(op)) {
                     return true;
                 } else if (COMPOSITE.equals(op)) {
                     // TODO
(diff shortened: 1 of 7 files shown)


controller/src/main/java/org/jboss/as/controller/remote/ModelControllerClientOperationHandler.java
```diff
@@ -78,6 +83,7 @@
  */
 public class ModelControllerClientOperationHandler implements ManagementRequestHandlerFactory {
 
+    private static final Set<String> PREPARED_RESPONSE_OPERATIONS = Stream.of(RELOAD, SHUTDOWN).collect(Collectors.toSet());
```
jamezp: I'm being a bit nitpicky, but I think `Arrays.asList(RELOAD, SHUTDOWN)` would be better here. The call stack to create this simple collection is huge with a stream.

testsuite/manualmode/src/test/java/org/wildfly/core/test/standalone/mgmt/PreparedResponseTestCase.java
```diff
@@ -0,0 +1,152 @@
...
+    public static void undeploy(ManagementClient managementClient) throws ServerDeploymentHelper.ServerDeploymentException {
+        ServerDeploymentHelper helper = new ServerDeploymentHelper(managementClient.getControllerClient());
+        helper.undeploy(SLOW_STOP_JAR);
+    }
+
+    @Test
+    public void reloadServer() throws Exception {
+        ManagementClient mgmtClient = getManagementClient();
+        long timeout = SHUTDOWN_WAITING_TIME + System.currentTimeMillis();
+        mgmtClient.executeForResult(Operations.createOperation(RELOAD, new ModelNode().setEmptyList()));
```
jamezp: In this case I don't think it matters, but I think in general as we write new tests with operations we shouldn't use the constants. It's mainly a regression check to ensure the operation names don't change in the constant value.... ...though I should probably bring this up on wildfly-dev and not this PR :)

testsuite/manualmode/src/test/java/org/wildfly/core/test/standalone/mgmt/PreparedResponseTestCase.java
```diff
@@ -0,0 +1,152 @@
...
+        while (System.currentTimeMillis() < timeout) {
+            Thread.sleep(FREQUENCY);
+            try {
+                controller.getClient().isServerInRunningState();
+            } catch (RuntimeException ex) {
+                ex.printStackTrace();
+                break;
```
jamezp: Do we want to print the message here or should it just fail?

testsuite/manualmode/src/test/java/org/wildfly/core/test/standalone/mgmt/PreparedResponseTestCase.java
```diff
@@ -0,0 +1,152 @@
...
+            }
+        }
+        Assert.assertFalse(System.currentTimeMillis() < timeout);
+        Thread.sleep(2 * FREQUENCY);
```
jamezp: This should use the `org.jboss.as.test.shared.TimeoutUtil.adjust()` to calculate the sleep time.

testsuite/manualmode/src/test/java/org/wildfly/core/test/standalone/mgmt/PreparedResponseTestCase.java
```diff
@@ -0,0 +1,152 @@
...
+
+    static ManagementClient getManagementClient() {
+        ModelControllerClient client = null;
+        try {
+            client = ModelControllerClient.Factory.create("http-remoting", InetAddress.getByName(TestSuiteEnvironment.getServerAddress()),
+                    MANAGEMENT_HTTP_PORT, new org.wildfly.core.testrunner.Authentication.CallbackHandler());
+        } catch (UnknownHostException e) {
+            throw new RuntimeException(e);
+        }
+        return new ManagementClient(client, TestSuiteEnvironment.getServerAddress(), MANAGEMENT_HTTP_PORT, "http-remoting");
```
jamezp: This should just be
```java
return new ManagementClient(TestSuiteEnvironment.getModelControllerClient(), TestSuiteEnvironment.getServerAddress(), TestSuiteEnvironment.getServerPort(), "http-remoting");
```
It probably shouldn't be using `org.jboss.as.test.integration.management.util.CustomCLIExecutor.MANAGEMENT_HTTP_PORT` at all since this isn't really a CLI test.

testsuite/manualmode/src/test/java/org/wildfly/core/test/standalone/mgmt/PreparedResponseTestCase.java
```diff
@@ -0,0 +1,152 @@
...
+    public void shutdownServer() throws Exception {
+        try (ManagementClient managementClient = getManagementClient()) {
+            long timeout = SHUTDOWN_WAITING_TIME + System.currentTimeMillis();
+            managementClient.executeForResult(Operations.createOperation(SHUTDOWN, new ModelNode().setEmptyList()));
+            while (System.currentTimeMillis() < timeout) {
+                Thread.sleep(FREQUENCY);
+                try {
+                    controller.getClient().isServerInRunningState();
+                } catch (RuntimeException ex) {
+                    ex.printStackTrace();
```
jamezp: Do we need this stack trace to be printed or should it just fail?

testsuite/shared/src/main/java/org/wildfly/test/shutdown/SlowStopServiceActivator.java
```diff
@@ -0,0 +1,64 @@
...
+import org.jboss.msc.service.ServiceActivator;
+import org.jboss.msc.service.ServiceActivatorContext;
+import org.jboss.msc.service.ServiceName;
+import org.jboss.msc.service.ServiceRegistryException;
+
+/**
+ * @author <a href="mailto:ehugonne@redhat.com">Emmanuel Hugonnet</a>  (c) 2015 Red Hat, inc.
+ */
+public class SlowStopServiceActivator implements ServiceActivator {
+    public static final long SHUTDOWN_WAITING_TIME = 3000L;
```
jamezp: This should use the `org.jboss.as.test.shared.TimeoutUtil.adjust()`.

testsuite/manualmode/src/test/java/org/wildfly/core/test/standalone/mgmt/PreparedResponseTestCase.java
ehsavoie: Yes as I didn't know about this

bstansberry:
My comments aside, I think this is fine in general. As for tests in full, if they ain't broke, don't fix 'em. ;)

bstansberry:
@ehsavoie Looks like this one needs a couple trivial tweaks and a rebase, so can you get that done and ping me or James to get it merged?

bstansberry:
retest this please

ehsavoie:
Retest this please

PR #1227, opened 2015-11-06
Improve test runner to allow few more things when running in manual s…

ctomc:
…erver mode

- allows us to not use arquillian in manual tests in full.
- adds support for deploy/undeploy operations on container.
- adds reload() method on server

### DIFF
--- a/testsuite/manualmode/src/test/java/org/jboss/as/test/manualmode/management/cli/ReloadOpsTestCase.java
+++ b/testsuite/manualmode/src/test/java/org/jboss/as/test/manualmode/management/cli/ReloadOpsTestCase.java
@@ -21,39 +21,28 @@
  */
 package org.jboss.as.test.manualmode.management.cli;
 
+import java.util.Map;
+import javax.inject.Inject;
+
+import org.jboss.as.cli.CommandContext;
+import org.jboss.as.controller.descriptions.ModelDescriptionConstants;
+import org.jboss.as.test.integration.management.base.AbstractCliTestBase;
+import org.jboss.as.test.integration.management.util.CLIOpResult;
+import org.jboss.as.test.integration.management.util.CLITestUtil;
+import org.junit.AfterClass;
+import org.junit.BeforeClass;
+import org.junit.Test;
+import org.junit.runner.RunWith;
+import org.wildfly.core.testrunner.ServerControl;
+import org.wildfly.core.testrunner.ServerController;
+import org.wildfly.core.testrunner.WildflyTestRunner;
+
 import static org.hamcrest.CoreMatchers.is;
-import static org.jboss.as.controller.descriptions.ModelDescriptionConstants.NAME;
-import static org.jboss.as.controller.descriptions.ModelDescriptionConstants.OP;
-import static org.jboss.as.controller.descriptions.ModelDescriptionConstants.OP_ADDR;
-import static org.jboss.as.controller.descriptions.ModelDescriptionConstants.READ_ATTRIBUTE_OPERATION;
-import static org.jboss.as.controller.descriptions.ModelDescriptionConstants.RESULT;
 import static org.junit.Assert.assertEquals;
 import static org.junit.Assert.assertNotNull;
 import static org.junit.Assert.assertNull;
 import static org.junit.Assert.assertThat;
 import static org.junit.Assert.assertTrue;
-import static org.junit.Assert.fail;
-
-import java.io.IOException;
-import java.util.Map;
-import javax.inject.Inject;
-
-import org.jboss.as.cli.CommandContext;
-import org.jboss.as.controller.client.ModelControllerClient;
-import org.jboss.as.controller.descriptions.ModelDescriptionConstants;
-import org.jboss.as.test.integration.management.base.AbstractCliTestBase;
-import org.jboss.as.test.integration.management.util.CLIOpResult;
-import org.jboss.as.test.integration.management.util.CLITestUtil;
-import org.jboss.dmr.ModelNode;
-import org.junit.AfterClass;
-import org.junit.BeforeClass;
-import org.junit.Test;
-import org.junit.runner.RunWith;
-import org.wildfly.core.testrunner.ManagementClient;
-import org.wildfly.core.testrunner.ServerControl;
-import org.wildfly.core.testrunner.ServerController;
-import org.wildfly.core.testrunner.WildflyTestRunner;
-import org.xnio.IoUtils;
 
 /**
  * @author <a href="mailto:ehugonne@redhat.com">Emmanuel Hugonnet</a>  (c) 2013 Red Hat, inc.
@@ -62,7 +51,7 @@ import org.xnio.IoUtils;
 @ServerControl(manual = true)
 public class ReloadOpsTestCase extends AbstractCliTestBase {
 
-    private static final long TIMEOUT = 10_000L;
+    private static final int TIMEOUT = 10_000;
 
     @Inject
     private static ServerController container;
@@ -81,7 +70,6 @@ public class ReloadOpsTestCase extends AbstractCliTestBase {
 
     @Test
     public void testWriteAttribvuteWithReload() throws Exception {
-        ManagementClient managementClient = container.getClient();
         cli.sendLine("/subsystem=logging:read-attribute(name=add-logging-api-dependencies)");
         CLIOpResult result = cli.readAllAsOpResult();
         assertTrue(result.isIsOutcomeSuccess());
@@ -91,7 +79,7 @@ public class ReloadOpsTestCase extends AbstractCliTestBase {
         result = cli.readAllAsOpResult();
         assertTrue(result.isIsOutcomeSuccess());
         checkResponseHeadersForProcessState(result);
-        reloadServer(TIMEOUT);
+        executeReload();
         cli.sendLine("/subsystem=logging:read-attribute(name=add-logging-api-dependencies)");
         result = cli.readAllAsOpResult();
         assertTrue(result.isIsOutcomeSuccess());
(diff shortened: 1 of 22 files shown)


jamezp:
I haven't looked this over yet, but the
> allows us to not use arquillian in manual tests in full.
Full is the only way Arquillian gets tested against the running container. If we stop using ARQ for full tests, then parts of ARQ might not get tested on the current container.

ctomc:
@jamezp that is only for manual mode tests where we are actually abusing arquillian for what we are doing and as result it is not as optimal as it should be.
All other testsuites still use and will continue to use ARQ

jamezp:
How are we abusing it? AFAIK you should be able to manually stop/start the container with ARQ and I think that's all we're doing in the manual mode tests right?

ctomc:
yes we are doing it *also* in manual mode tests but not exclusively.
Abusing as in we are testing stuff with it that was not meant to be used in this way.
For example not using any deployment at all just some dummy one so arquillian is happy.
But it is not really needed when we are testing for example something in the model and if it properly applied to runtime, which are lots test in manual mode.

In any case, this doesn't require us to use one or another, but just enables us to be able to use one that is more appropriate for certain scenarios.
for example what is possible see https://github.com/ctomc/wildfly/commit/2af95a1d7578390011b04081f794c95e4ce66808
which already uses new runner

testsuite/manualmode/src/test/java/org/jboss/as/test/manualmode/management/cli/ReloadOpsTestCase.java
```diff
@@ -91,7 +77,7 @@ public void testWriteAttribvuteWithReload() throws Exception {
         result = cli.readAllAsOpResult();
         assertTrue(result.isIsOutcomeSuccess());
         checkResponseHeadersForProcessState(result);
-        reloadServer(TIMEOUT);
+        container.reload(TIMEOUT);
```
bstansberry: Since this is a CLI test, it seems better to let the CLI do the reload.

testsuite/manualmode/src/test/java/org/jboss/as/test/manualmode/management/cli/ReloadOpsTestCase.java
```diff
@@ -108,55 +94,13 @@ public void testWriteAttribvuteWithReload() throws Exception {
         assertThat(value, is("false"));
     }
 
-    private void reloadServer(long timeout) throws Exception {
-        executeReload();
-        waitForLiveServerToReload(timeout);
```
bstansberry: This waitForLiveServerToReload stuff shouldn't be needed. The executeReload method uses the high level CLI reload command, and that should already be waiting.

testsuite/test-runner/src/main/java/org/wildfly/core/testrunner/Server.java
```diff
@@ -221,6 +256,66 @@ private void safeCloseClient() {
...
 
     /**
+     * Reload current server
+     * This method makes sure client on server is still operational after reload is done.
+     * @param adminOnly tells server to boot in admin only mode or normal
+     * @param timeout time in miliseconds to wait for server to come back after reload
+     */
+    public void reload(boolean adminOnly,int timeout){
+        executeReload(adminOnly);
+        waitForLiveServerToReload(timeout); //30 seconds
```
bstansberry: Use org.jboss.as.test.integration.management.util.ServerReload, which has been ported from full to core.

testsuite/test-runner/src/main/java/org/wildfly/core/testrunner/Server.java
```diff
@@ -221,6 +256,66 @@ private void safeCloseClient() {
...
+            safeCloseClient();//close existing client
+        }
+    }
+
+    private void recreateClient(){
+        safeCloseClient();
+        client = createClient();
+    }
+
+    protected void waitForLiveServerToReload(int timeout) {
```
bstansberry: Use org.jboss.as.test.integration.management.util.ServerReload, which has been ported from full to core.

testsuite/test-runner/src/main/java/org/wildfly/core/testrunner/Server.java
ctomc: It wasn't there at time of initial writing of this code :-) 
But in any case this wouldn't work as this is test runner that has no dependencies to shared testsuite. If anything ServerReload in testsuite should depend on test runner's code.

Also most of tests in core should only use test runner and its classes for reload capabilities, instead of relying on yet another utility class for this.

testsuite/test-runner/src/main/java/org/wildfly/core/testrunner/Server.java
ctomc: This two methods are bit different in its implementation, code in ServerReload#waitForLiveServerToReload is leaking ModelControllerClient(s) as they are not properly closed on recreating new client. I had quite some issues with this code(started from what is in ServerReload) when I was testing/looking into reload issue we had few months back. 
In short I trust this implementation better.

testsuite/manualmode/src/test/java/org/jboss/as/test/manualmode/management/cli/ReloadOpsTestCase.java
ctomc: I removed this as it was unifying the code, but yeah it does make sense to use the CLI for this.

testsuite/manualmode/src/test/java/org/jboss/as/test/manualmode/management/cli/ReloadOpsTestCase.java
ctomc: There (w)are lots of similar problems all around our tests, not sure when this was initially introduces, but I agree it shouldn't be needed.

testsuite/manualmode/src/test/java/org/jboss/as/test/manualmode/management/cli/ReloadOpsTestCase.java
ctomc: Just remembered why I did this, the original code for cli reload does
final CommandContext ctx = CLITestUtil.getCommandContext();
which creates new network connection to http-remoting port. and than executes reload.

The replacement code uses already open client and reuses that. 
So in short it speeds up the test.

testsuite/standalone/src/test/java/org/jboss/as/test/integration/management/cli/GenericCommandTestCase.java
```diff
@@ -29,16 +29,13 @@
...
-import org.junit.runner.RunWith;
 import org.wildfly.core.testrunner.ServerSetup;
-import org.wildfly.core.testrunner.WildflyTestRunner;
 
 /**
  * Test generic command features of CLI.
  * @author Dominik Pospisil <dpospisi@redhat.com>
  */
 @ServerSetup(ServerReload.SetupTask.class)
-@RunWith(WildflyTestRunner.class)
```
bstansberry: Did the wrong line get removed here? In GlobalOpsTestCase you removed the ServerReloadSetupTask and added a CLI command call, but here you left the setup task, which would be redundant.

ctomc: yes, you are right, good eye.

bstansberry:
@ctomc This has a checkstyle problem:


[18:16:11][org.wildfly.core:wildfly-core-testsuite-standalone] /opt/buildAgent/work/1780c164c3686efd/core/testsuite/standalone/src/test/java/org/jboss/as/test/integration/management/cli/GenericCommandTestCase.java:28:8: Unused import - org.jboss.as.test.integration.management.util.ServerReload.
[18:16:11][org.wildfly.core:wildfly-core-testsuite-standalone] /opt/buildAgent/work/1780c164c3686efd/core/testsuite/standalone/src/test/java/org/jboss/as/test/integration/management/cli/GenericCommandTestCase.java:33:8: Unused import - org.wildfly.core.testrunner.ServerSetup.
[18:16:11][org.wildfly.core:wildfly-core-testsuite-standalone] Audit done.
[18:16:11][org.wildfly.core:wildfly-core-testsuite-standalone] Failed to execute goal org.apache.maven.plugins:maven-checkstyle-plugin:2.17:checkstyle (check-style) on project wildfly-core-testsuite-standalone: An error has occurred in Checkstyle report generation.

ctomc:
@bstansberry sorry about that, it is fixed now.

bstansberry:
Testing this on master-ignore, BasicOperationsUnitTestCase failed:

https://ci.wildfly.org/viewLog.html?buildId=14679&buildTypeId=WildFlyCore_MasterIgnoreLinux&tab=buildLog

Failed because the client the test uses was closed. 

The log shows the failed test ran right after ReloadWithConfigTestCase, which is updated in this PR.

ctomc:
this is strange, I cannot reproduce this locally on all JDKs I have installed, also PR job passed. weird.

bstansberry:
Since it's doesn't repeatedly fail, look carefully at concurrency related to accessing the client (i.e. callers always get the current one.)

ctomc:
@bstansberry it should be fixed now. it is a interesting case of when we are using

```
@Inject
private static ManagementClient managementClient;
```
in case that filed is static, it gets injected when surefire/junit is creating the test which can happen at start of test run of many test, so when it is static, it gets injected at beginning of test run, but if one of the tests in whole run does .reload() it invalidates the currently injected client which is no longer working.
Just changing field to non-static makes it get injected just before the tests are run, instead of on creation of test. Which makes the problem goes away as the injected client is valid for current running state.

Just heads up if we see similar problem elsewhere.

luck3y:
I ran into this using:
$ mvn test -Dtest=ReloadWithConfigTestCase,DescribeOperationUnitTestCase

On a clean checkout of master, the tests executed in the order: DescribeOperationUnitTestCase, ReloadWithConfigTestCase and passed.

In my own branch, they would always execute in the order ReloadWithConfigTestCase, DescribeOperationUnitTestCase and fail. (I didn't change the tests or order at all.).

ctomc:
@luck3y you are right, problem is the same as before but for different test.
I've changed the @inject logic that does injection that should make sure always correct client is injected and this problem shouldn't occur even for static fields anymore

ctomc:
fixed version of PR https://github.com/wildfly/wildfly-core/pull/1528

PR #1229, opened 2015-11-09
[WFCORE-851] - Remoting Subsystem RemoteOutboundConnectionService is …caching ConnectionURI causing issues when DNS changes

spolti:
Remoting Subsystem RemoteOutboundConnectionService is caching ConnectionURI causing issues when DNS changes

Downstream: https://issues.jboss.org/browse/JBEAP-1960
https://github.com/jbossas/wildfly-core-eap/pull/61

### DIFF
--- a/network/src/main/java/org/jboss/as/network/OutboundSocketBinding.java
+++ b/network/src/main/java/org/jboss/as/network/OutboundSocketBinding.java
@@ -163,8 +163,7 @@ public class OutboundSocketBinding {
         if (this.resolvedDestinationAddress != null) {
             return this.resolvedDestinationAddress;
         }
-        this.resolvedDestinationAddress = InetAddress.getByName(this.unresolvedDestinationAddress);
-        return this.resolvedDestinationAddress;
+        return InetAddress.getByName(this.unresolvedDestinationAddress);
     }
 
     /**
--- a/remoting/subsystem/src/main/java/org/jboss/as/remoting/RemoteOutboundConnectionService.java
+++ b/remoting/subsystem/src/main/java/org/jboss/as/remoting/RemoteOutboundConnectionService.java
@@ -65,7 +65,6 @@ public class RemoteOutboundConnectionService extends AbstractOutboundConnectionS
 
     private final String username;
     private final String protocol;
-    private URI connectionURI;
 
     public RemoteOutboundConnectionService(final String connectionName, final OptionMap connectionCreationOptions, final String username, final String protocol) {
         super(connectionName, connectionCreationOptions);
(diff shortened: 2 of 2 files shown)


bmaxwell:
OutboundSocketBinding has a constructor that takes in and sets the resolvedDestinationAddress , so we need to take that into account.  It looks like the constructor is used for setting up Local Receivers.  It looks like we either always resolve again or if there a reason we should not try to resolve when the resolvedDestinationAddress is passed in, then need a boolean to have it returned the resolved instead of resolving again.

```
    112     public OutboundSocketBinding(final String name, final SocketBindingManager socketBindingManager,
    113                                  final InetAddress destinationAddress, final int destinationPort,
    114                                  final NetworkInterfaceBinding sourceNetworkInterface, final Integer sourcePort,
    115                                  final boolean fixedSourcePort) {
    116         this(name, socketBindingManager, destinationAddress.getHostAddress(), destinationPort, sourceNetworkInterface, sourcePort, fixedSourcePort);
    117         this.resolvedDestinationAddress = destinationAddress;
```

network/src/main/java/org/jboss/as/network/OutboundSocketBinding.java
```diff
@@ -160,9 +160,6 @@ public String getUnresolvedDestinationAddress() {
      * @throws UnknownHostException If the destination address cannot be resolved
      */
     public synchronized InetAddress getResolvedDestinationAddress() throws UnknownHostException {
-        if (this.resolvedDestinationAddress != null) {
-            return this.resolvedDestinationAddress;
-        }
```
bstansberry: This should be kept so we can use the resolvedDestinationAddress passed in to the constructor.

network/src/main/java/org/jboss/as/network/OutboundSocketBinding.java
```diff
@@ -160,9 +160,6 @@ public String getUnresolvedDestinationAddress() {
...
         this.resolvedDestinationAddress = InetAddress.getByName(this.unresolvedDestinationAddress);
         return this.resolvedDestinationAddress;
```
bstansberry: These two lines should just be:

return InetAddress.getByName(this.unresolvedDestinationAddress);

So don't store the value if it wasn't provided to the constructor; re-resolve each time.

The resolvedDestinationAddress field should also be made final as it can no longer be set outside the constructor.

This and my comment above should address the problem Brad Maxwell mentioned in his comment.

remoting/subsystem/src/main/java/org/jboss/as/remoting/RemoteOutboundConnectionService.java
```diff
@@ -145,15 +145,12 @@ public String getProtocol() {
...
-            return this.connectionURI;
-        }
+        /* WFCORE-851 - do not cache connectionURI else reconnect will fail if DNS changes */
         final OutboundSocketBinding destinationOutboundSocket = this.destinationOutboundSocketBindingInjectedValue.getValue();
         final InetAddress destinationAddress = destinationOutboundSocket.getResolvedDestinationAddress();
         final int port = destinationOutboundSocket.getDestinationPort();
 
-        this.connectionURI = new URI(protocol + "://" + NetworkUtils.formatPossibleIpv6Address(destinationAddress.getHostAddress()) + ":" + port);
-        return this.connectionURI;
+        return new URI(protocol + "://" + NetworkUtils.formatPossibleIpv6Address(destinationAddress.getHostAddress()) + ":" + port);
```
bstansberry: The connectionURI field should also be removed altogether, as it is no longer used.

bstansberry:
:+1

bstansberry:
retest this please

bstansberry:
I removed Ready for Merge only because this needs a downstream PR.

bmaxwell:
downstream: https://github.com/jbossas/wildfly-core-eap/pull/61

bmaxwell:
PRs synced up
https://github.com/wildfly/wildfly-core/pull/1229
https://github.com/jbossas/wildfly-core-eap/pull/61
https://github.com/jbossas/jboss-eap/pull/2618

PR #1230, opened 2015-11-09
[WFCORE-1107]:  DirectoryNotEmptyException on undeploy of ear.

ehsavoie:
Checking that the directory is empty before trying to delete it.

Jira: https://issues.jboss.org/browse/WFCORE-1107
EAP: https://issues.jboss.org/browse/JBEAP-1803

### DIFF
--- a/deployment-repository/src/main/java/org/jboss/as/repository/ContentRepository.java
+++ b/deployment-repository/src/main/java/org/jboss/as/repository/ContentRepository.java
@@ -41,6 +41,7 @@ import java.util.HashMap;
 import java.util.HashSet;
 import java.util.Map;
 import java.util.Set;
+import java.util.stream.Stream;
 
 import org.jboss.as.repository.logging.DeploymentRepositoryLogger;
 import org.jboss.msc.service.Service;
@@ -277,8 +278,7 @@ public interface ContentRepository {
             }
 
             private Path getDeploymentContentFile(byte[] deploymentHash, boolean validate) {
-                final Path hashDir = getDeploymentHashDir(deploymentHash, validate);
-                return hashDir.resolve(CONTENT);
+                return getDeploymentHashDir(deploymentHash, validate).resolve(CONTENT);
             }
 
             protected Path getDeploymentHashDir(final byte[] deploymentHash, final boolean validate) {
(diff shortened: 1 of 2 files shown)


deployment-repository/src/main/java/org/jboss/as/repository/ContentRepository.java
```diff
@@ -365,18 +366,23 @@ public void removeContent(ContentReference reference) {
                 Path file = getDeploymentContentFile(reference.getHash(), true);
                 try {
                     Files.deleteIfExists(file);
+                    assert !Files.exists(file);
```
jamezp: This seems weird. I don't think this would ever happen. Also `Files.notExists()` is better IMO :)

ehsavoie: leftover from testing as the folder was recreated by another method call in tests

dmlloyd: Just a note for posterity.  Well, a couple notes.

First, exists and notExists are not strictly opposites.  One determines if the file definitely exists, and one determines if the file definitely does not exist.  If the existence of a file cannot be determined, notExists and exists will both return false.

Second, this is inherently racy as the file could theoretically be recreated between the delete and the check.  Yes, it's unlikely, but unlikely things have a way of becoming likely in production environments.

deployment-repository/src/main/java/org/jboss/as/repository/ContentRepository.java
```diff
@@ -376,7 +375,9 @@ public void removeContent(ContentReference reference) {
                 }
                 Path grandParent = parent.getParent();
                 try {
-                    Files.deleteIfExists(grandParent);
+                    if (!Files.list(grandParent).findAny().isPresent()) {
+                        Files.deleteIfExists(grandParent);
+                    }
                 } catch (IOException ex) {
```
bstansberry: Files.list(Path) notes that the returned stream is a DirectoryStream that should be handled using try-with-resources. So perhaps:

```
try (Stream s = Files.list(grandparent)) {
    if (!s.findAny().isPresent()) {
        Files.deleteIfExists(grandParent);     
    }
}
```

PR #1207, opened 2015-10-29
[WFCORE-1049] Populate the MRR before registering it so outsiders onl…

bstansberry:
…y see the populated state

PR #1231, opened 2015-11-10
[WFCORE-1049] Use locking to protect MRR until it is completely initialized

bstansberry:
This is a replacement for #1207. The 1207 approach avoided any locking but needed a fair bit of complexity to deal with the situation where while the new but unregistered MRR was being populated via calls to the ResourceDefinition, the ResourceDefinition in turn called into the MRR to get data. The approach worked, but was complex and might not work in some obscure cases.

This approach uses locks instead to prevent other threads reading a new MRR before it is fully populated.


ctomc:
retest this please

n1hility:
retest this please

PR #1233, opened 2015-11-10
Few minor scripts improvments

ctomc:
WFCORE-1085 Linux and powershell standalone script should be unified
WFCORE-995 Add possibility to enable debug mode by setting env variable
WFCORE-1084 Standalone powershell script does't support SECMGR=true env variable


bstansberry:
Please separate out WFCORE-995 into a different PR (same in the EAP PR) for getting acks for an enhancement doesn't block merging the 2 already-approved bug fixes.

One issue per PR is going to make everyone's life easier.

ctomc:
well two commits are co-dependent as one builds on top of another.
Will just take the long route to get this fixed all together

PR #1234, opened 2015-11-10
[WFCORE-1074]: Intermittent failure in DeploymentScannerUnitTestCase.

ehsavoie:
Previous commit timeout is too short to have a meaningfull test: moving it back to 30 seconds.


bstansberry:
@ehsavoie Please create an EAP variant of this. No need for a JBEAP. Thanks.

PR #1235, opened 2015-11-10
WFCORE-1120 Launcher shouldn't "-Djava.net.preferIPv4Stack=true" if s…

ctomc:
…pecifed otherwise


bstansberry:
@jamezp please approve

I think the same basic issue applies to all of the DEFALT_VM_ARGUMENTS. However, since we have a JBEAP for this specific thing, if you guys agree that this is a more general problem, I think we'd be better off filing a separate JIRA for that and leaving this targeted at the existing issue.

jamezp:
While it shouldn't matter if the property is there twice, the later wins IIRC, it does seem odd to have two system properties with different values. This PR however will only fix this single property. As Brian said all the `DEFAULT_VM_ARGUMENTS` would have the same issue. I'm closing this in favor of https://github.com/wildfly/wildfly-core/pull/1238.

ctomc:
It is not just odd, there is no rule to know which property value will be used at runtime, in most cases it works fine as "right" one wins but not in all cases. This is why we didn't notice this earlier.

PR #1236, opened 2015-11-10
Do some housekeeping on @Ignored tests in core

bstansberry:
Drop some tests, fix some, improve the description on why some are ignored.


bstansberry:
retest this please

PR #1237, opened 2015-11-10
[WFCORE-1119] Avoid NPE if bind fails during managed binding constructor

bstansberry:
Related EAP JIRA is https://issues.jboss.org/browse/JBEAP-1870 (acked)


bstansberry:
I'm closing this because #1239 incorporates it, both issues in 1239 are EAP Acked, and I want both fixes in.

PR #1239, opened 2015-11-11
[WFCORE-1127] Improve handling of unregistering unnamed socket bindings.

bstansberry:
Also includes the commit for WFCORE-1119 from #1237 as that commit would conflict with this one.

PR #1242, opened 2015-11-11
[WFCORE-1029] Stop using InetSocketAddress as map key in UnnamedBindingRegistry

bstansberry:
This builds on #1239 but adds an additional commit for https://issues.jboss.org/browse/WFCORE-1029.

I'm using separate PRs for these because the different related issues have, IMO, different degrees of acceptability for inclusion in EAP 7. I believe including the WFCORE-1029 fix is the correct ultimate solution, but having added bake time makes sense to me.

### DIFF
--- a/network/src/main/java/org/jboss/as/network/SocketBindingManagerImpl.java
+++ b/network/src/main/java/org/jboss/as/network/SocketBindingManagerImpl.java
@@ -150,7 +150,7 @@ public abstract class SocketBindingManagerImpl implements SocketBindingManager {
         return unnamedRegistry;
     }
 
-    class ManagedSocketFactoryImpl extends ManagedSocketFactory {
+    private class ManagedSocketFactoryImpl extends ManagedSocketFactory {
 
         @Override
         public Socket createSocket() {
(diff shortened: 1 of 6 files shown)


network/src/main/java/org/jboss/as/network/SocketBindingManagerImpl.java
```diff
@@ -490,36 +550,30 @@ public void unregisterBinding(String name) {
         }
     }
 
-    static final class UnnamedRegistryImpl implements UnnamedBindingRegistry {
-        private final Map<InetSocketAddress, ManagedBinding> bindings = new ConcurrentHashMap<InetSocketAddress, ManagedBinding>();
+    private static final class UnnamedRegistryImpl implements UnnamedBindingRegistry {
+        private final Map<WrapperBinding, Void> bindings = new ConcurrentHashMap<>();
```
jamezp: Maybe a Set would be better, but we either can't use a `null` value or we need to change the map type. Per the [ConcurrentHashMap JavaDoc](https://docs.oracle.com/javase/8/docs/api/java/util/concurrent/ConcurrentHashMap.html)

> Like Hashtable but unlike HashMap, this class does not allow null to be used as a key or value.

jamezp:
retest this please

jamezp:
retest this please

PR #1245, opened 2015-11-13
[WFCORE-1115]: Intermittent failures of CompositeOperationTestCase & CoreResourceManagementTestCase.

ehsavoie:
Checking that the slave servers have restarted after reload in IgnoredResourcesProfileCloneTestCase.

Jira: https://issues.jboss.org/browse/WFCORE-1115
https://issues.jboss.org/browse/WFCORE-1089

### DIFF
--- a/testsuite/domain/src/test/java/org/jboss/as/test/integration/domain/suites/IgnoredResourcesProfileCloneTestCase.java
+++ b/testsuite/domain/src/test/java/org/jboss/as/test/integration/domain/suites/IgnoredResourcesProfileCloneTestCase.java
@@ -39,7 +39,6 @@ import static org.jboss.as.controller.descriptions.ModelDescriptionConstants.SUB
 import static org.jboss.as.controller.descriptions.ModelDescriptionConstants.TO_PROFILE;
 import static org.junit.Assert.assertEquals;
 
-import java.io.File;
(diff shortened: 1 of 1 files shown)


testsuite/domain/src/test/java/org/jboss/as/test/integration/domain/suites/IgnoredResourcesProfileCloneTestCase.java
```diff
@@ -92,14 +86,11 @@ public static void beforeClass() throws Exception {
...
         slaveLifecycleUtil = testSupport.getDomainSlaveLifecycleUtil();
-
-        File masterDir = new File(testSupport.getDomainMasterConfiguration().getDomainDirectory());
-        domainCfg = new File(masterDir, "configuration"
-                + File.separator + "testing-domain-standard.xml");
     }
 
     @AfterClass
     public static void afterClass() throws Exception {
+        slaveLifecycleUtil.awaitServers(System.currentTimeMillis());
```
bstansberry: This isn't reliable because it only waits for servers configured with auto-start=true. It's basically written for use during a normal HC boot.

I had to work around this problem in HcExtensionAndSubsystemManagementTestCase. Kind of a pain.

PR #1246, opened 2015-11-16
[WFCORE-1077] Update the regular expression to correctly split out the username and password.

darranl:


darranl:
Just spotted a problem.

PR #1250, opened 2015-11-16
Run subsystem schema tests by default

kabir:
https://issues.jboss.org/browse/WFCORE-1128

Full will fail until https://github.com/wildfly/wildfly/pull/8401 and https://github.com/wildfly/wildfly/pull/8396 are merged

### DIFF
--- a/subsystem-test/framework/src/main/java/org/jboss/as/subsystem/test/AbstractSubsystemBaseTest.java
+++ b/subsystem-test/framework/src/main/java/org/jboss/as/subsystem/test/AbstractSubsystemBaseTest.java
@@ -22,14 +22,20 @@
 
 package org.jboss.as.subsystem.test;
 
+import java.io.File;
 import java.io.IOException;
+import java.nio.file.Files;
+import java.nio.file.Path;
+import java.nio.file.Paths;
 import java.util.Collections;
 import java.util.Comparator;
 import java.util.List;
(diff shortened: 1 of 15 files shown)


subsystem-test/framework/src/main/java/org/jboss/as/subsystem/test/AbstractSubsystemBaseTest.java
```diff
@@ -63,32 +69,92 @@ public AbstractSubsystemBaseTest(final String mainSubsystemName, final Extension
...
+            private Finder(File dir, Pattern pattern, boolean patternIncludesSubsystem) {
+                this.dir = dir;
+                this.pattern = pattern;
+                this.patternIncludesSubsystem = patternIncludesSubsystem;
+            }
+
+            File findHighestVersion(){
+                ModelVersion max = null;
+                File maxFile = null;
+                for (File file : dir.listFiles((d, f) -> pattern.matcher(f).matches())) {
```
ctomc: Probably better version would be to use Files.list(dir)....

bstansberry:
Looks good. Thanks.

I've put the 3.x label on this as I don't thinks it's needed for WF 10.

Last week there were some discussions around this general topic on the EAP side. If those lead to a desire to get this in the EAP branch for 7.0, then we can do a JBEAP and port this and the full PR.

PR #1251, opened 2015-11-17
WFCORE-1136 Subsystem parser tests fail when woodstox parser is on classpath

ctomc:
probably breaks also with other parsers.

### DIFF
--- a/subsystem-test/framework/src/main/java/org/jboss/as/subsystem/test/TestParser.java
+++ b/subsystem-test/framework/src/main/java/org/jboss/as/subsystem/test/TestParser.java
@@ -71,6 +71,8 @@ public final class TestParser implements  ModelTestParser {
                 if (subsystemWriter != null) {
                     subsystemWriter.writeContent(writer, new SubsystemMarshallingContext(subsystem, writer));
                 }
+            }else{
+                writer.writeEmptyElement(Element.SUBSYSTEM.getLocalName());
             }
         }catch (Throwable t){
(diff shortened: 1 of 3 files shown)


subsystem-test/framework/src/main/java/org/jboss/as/subsystem/test/TestParser.java
```diff
@@ -77,7 +78,11 @@ public void writeContent(XMLExtendedStreamWriter writer, ModelMarshallingContext
         } finally {
             writer.setDefaultNamespace(defaultNamespace);
         }
+        if (empty){
+            writer.writeEmptyElement(Element.SUBSYSTEM.getLocalName());
+        }
```
bstansberry: Why not in the try block?

ctomc: as it complicates the code because of the defaultNamespace can be null.

ctomc: but yeah both options would work

ctomc: I've changed it to be in try block, as namespace is not relevant in this scenario at all.

ctomc:
retest this please

n1hility:
Retest this please

bstansberry:
I'm putting 3.x on this. If you want it in EAP, please clone the JIRA to JBEAP.

ctomc:
@bstansberry JBEAP-2098 got acked.

PR #1254, opened 2015-11-19
WFCORE-1056 try to prevent --empty-config from causing issues 

luck3y:
with embed-server if the configuration already exists and --remove-existing was not provided.

### DIFF
--- a/cli/src/main/java/org/jboss/as/cli/embedded/EmbedServerHandler.java
+++ b/cli/src/main/java/org/jboss/as/cli/embedded/EmbedServerHandler.java
@@ -183,6 +183,15 @@ class EmbedServerHandler extends CommandHandlerWithHelp {
                 cmdsList.add("--server-config=" + xml);
             }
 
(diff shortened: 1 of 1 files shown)


luck3y:
retest this please

luck3y:
retest this please

luck3y:
retest this please

bstansberry:
@luck3y This handles the --empty-config problem nicely; better than waiting for the server start to barf. But there are other cases where the InstanceAlreadyExistsException logging will still happen. For example, try corrupting standalone.xml and starting, then fixing and starting again.

We need to ensure the MSC ServiceContainer gets shut down. I think something like https://github.com/wildfly/wildfly-core/compare/master...bstansberry:stop-msc will do the trick.

I expect embed-host-controller has a similar problem.

luck3y:
@bstansberry +1 I'll check against embedded HC and send PR

bstansberry:
retest this please

PR #1256, opened 2015-11-19
WFCORE-1056 prevent --empty-config from causing issues with embed-ser…

luck3y:
…ver if the configurtion already exists and --remove-existing was not provided

Fixed the source branch name (was erronously WFCORE-950, should be WFCORE-1056)

PR #1258, opened 2015-11-19
Additional fixes for WFCORE-1056 for Standalone and HC embedded

luck3y:


luck3y:
These are in addition to #1256

PR #1261, opened 2015-11-20
WFCORE-680 Eliminating dependency on JBoss VFS

ropalka:
This PR removes Wildfly dependency on JBoss VFS API.

First half of commits is focused on minimalizing JBoss VFS usage.
Second half replaces JBoss VFS with new alternative.

Design document for the change is here: https://mojo.redhat.com/docs/DOC-1062182

This PR modifies vast majority of available structure processors.
It also rewrites Deployment Overlays feature completely.
Many subsystems were touched, so review should be careful.

All tests available in both WF and WF Core code base are passing.

Thanks in advance for your review.


ropalka:
It seems TeamCity tries to build WF on top of WF-Core created from this PR. This won't work until I'll send PR for WF.

embedded/src/main/java/org/wildfly/core/embedded/logging/EmbeddedLogger.java
```diff
@@ -51,86 +47,6 @@
...
-
-    /**
-     */
-    @Message(id = 9, value = "One or more exclusion values must be specified")
-    IllegalArgumentException exclusionValuesRequired();
-
-    //@Message(id = 10, value = "WARNING: Failed to load the specified logmodule %s")
-    //String failedToLoadLogModule(ModuleIdentifier moduleId);
-
-    /**
```
bstansberry: Please don't delete log methods. Just comment them out so the reason for gaps in the list of ids is clearly visible.

self-contained/src/main/java/org/jboss/as/selfcontained/ContentProvider.java
```diff
@@ -22,16 +22,18 @@
...
  *
  * @author Bob McWhirter
+ * @author <a href="mailto:ropalka@redhat.com">Richard Opalka</a>
  */
 public interface ContentProvider {
 
     ServiceName NAME = ServiceName.JBOSS.append( "self-contained", "content-provider" );
 
-    VirtualFile getContent(int index);
+    File getContent(int index);
```
bstansberry: I'm not sure if this is/isn't a breaking API change. This comment is mostly a note to Ricahrd and reviewers to check that out.

bstansberry:
A general design comment on this is I'd like us to consider not including this new 'loader' package in the server module. We really need to break that module up between stuff specific to the kernel and stuff needed by subsystems. We'll never have a proper API for ISVs until we do that. The 'loader' package looks to be a new facility we make available to subsystems so I'd like to make sure we don't make the problem with 'server' worse.

bstansberry:
@ropalka is there a design document for this? I think a discussion thread on wildfly-dev makes sense as this is big change and needs visiblity. Plus questions about will probably work better in a threaded form vs comments.

cli/src/main/java/org/jboss/as/cli/handlers/DeploymentHandler.java
```diff
@@ -117,8 +131,60 @@ protected ModelNode buildReadDeploymentResourceRequest(String name) {
...
+                // create directory structure
+                newFile = new File(targetDir + File.separator + fileName);
+                new File(newFile.getParent()).mkdirs();
+                // extract zip
+                if (!entry.isDirectory()) {
+                    // extract zip entry to the disk
+                    FileOutputStream fos = null;
+                    try {
+                        fos = new FileOutputStream(newFile);
+                        copy(zis, fos);
```
ctomc: Files.copy() would work fine there.
Also using Path api for methods like this would be a great help when it comes to error reporting.

cli/src/main/java/org/jboss/as/cli/handlers/DeploymentHandler.java
```diff
@@ -117,8 +131,60 @@ protected ModelNode buildReadDeploymentResourceRequest(String name) {
...
+                    } finally {
+                        safeClose(fos);
+                    }
+                }
+            }
+        } finally {
+            safeClose(zis);
```
ctomc: this can be eliminated if you use try with resources
something like
try (final ZipInputStream zis = new ZipInputStream(new FileInputStream(cliArchive))){
...
}

cli/src/main/java/org/jboss/as/cli/handlers/DeploymentHandler.java
```diff
@@ -117,8 +131,60 @@ protected ModelNode buildReadDeploymentResourceRequest(String name) {
         return request;
     }
 
-    protected MountHandle extractArchive(File archive, TempFileProvider tempFileProvider) throws IOException {
-        return ((MountHandle)VFS.mountZipExpanded(archive, VFS.getChild("cli"), tempFileProvider));
+    static File extractArchive(final File cliArchive) throws IOException {
+        final ZipInputStream zis = new ZipInputStream(new FileInputStream(cliArchive));
```
ctomc: this will produce file handle leak.
moving this to try-with-resources block would eliminate the leak as well.

cli/src/main/java/org/jboss/as/cli/handlers/DeploymentHandler.java
```diff
@@ -117,8 +131,60 @@ protected ModelNode buildReadDeploymentResourceRequest(String name) {
...
+    private static void copy(final InputStream is, final OutputStream os) throws IOException {
+        final byte[] buffer = new byte[1024];
+        int len;
+        while ((len = is.read(buffer)) > 0) {
+            os.write(buffer, 0, len);
+        }
+    }
+
+    private static void safeClose(final Closeable closeable) {
+        if (closeable != null) try { closeable.close(); } catch (final Throwable ignored) {}
```
ctomc: duplicate code with in Utils package

cli/src/main/java/org/jboss/as/cli/handlers/Utils.java
```diff
@@ -0,0 +1,48 @@
...
+ *
+ * @author <A href="mailto:ropalka@redhat.com">Richard Opalka</A>
+ */
+public final class Utils {
+
+    private Utils() {
+        // forbidden instantiation
+    }
+
+    public static void safeClose(final Closeable closeable) {
```
ctomc: this could be probably eliminated by using try with resources in code that now uses it.

management-client-content/src/main/java/org/jboss/as/management/client/content/ManagedDMRContentTypeResource.java
```diff
@@ -282,13 +283,13 @@ ManagedContent getManagedContent(final String name) {
...
-        VirtualFile vf = contentRepository.getContent(initialHash);
-        if (vf == null) {
+        File f = contentRepository.getContent(initialHash);
+        if (f == null) {
             throw ManagedDMRContentLogger.ROOT_LOGGER.noContentFoundWithHash(HashUtil.bytesToHexString(initialHash));
         }
         InputStream is = null;
         try {
-            is = vf.openStream();
+            is = new FileInputStream(f);
```
ctomc: should be moved to try() block

stuartwdouglas:
As Brian says a change of this magnitude really needs a design discussion on wildfly-dev. In particular it needs to cover:

1) What are the actual issues that we are facing with VFS, and why they are so severe that a complete re-write is required.

2) The design of the replacement

3) Compatibility implications for 3rd party subsystems. As I understand it they will all need to be rewritten to use this.

4) The compatibility implications for 3rd party frameworks have special handling for VFS URL's, after this change there is a chance these will no longer work.

stuartwdouglas:
Why are you commiting binary files (archives and class files) into the repo? I think they should be built at run time using Arquillian, the makes it obvious exactly what these archives contain, and keeps things consistent with the rest of the test suite.

ropalka:
Full integration was tested here: http://brontes.lab.eng.brq.redhat.com/viewType.html?buildTypeId=WF_WildFlyCoreIntegrationExperiments

bstansberry:
Since this won't happen any time soon I'm closing this PR to keep the queue clean.

PR #1264, opened 2015-11-23
WFCORE-1143 Cancel button in JConsole do nothing

ctomc:
Downstream: https://issues.jboss.org/browse/JBEAP-1801


bstansberry:
If we get a qa_ack on the downstream then we can get rid of the 3.x label.

PR #1266, opened 2015-11-23
WFCORE-950 - allow changing of --jboss-home in non-modular

luck3y:
The should help if a user enters an incorrect --jboss-home or restarts with a different --jboss-home.

### DIFF
--- a/embedded/src/main/java/org/wildfly/core/embedded/EmbeddedHostControllerFactory.java
+++ b/embedded/src/main/java/org/wildfly/core/embedded/EmbeddedHostControllerFactory.java
@@ -144,7 +144,7 @@ public class EmbeddedHostControllerFactory {
             props.put(HostControllerEnvironment.DOMAIN_BASE_DIR, tempRoot.getAbsolutePath());
             props.put(HostControllerEnvironment.DOMAIN_CONFIG_DIR, configDir.getAbsolutePath());
             props.put(HostControllerEnvironment.DOMAIN_DATA_DIR, dataDir.getAbsolutePath());
-        }  catch (IOException e) {
+        } catch (IOException e) {
             throw EmbeddedLogger.ROOT_LOGGER.cannotSetupEmbeddedServer(e);
         }
 
@@ -220,8 +220,7 @@ public class EmbeddedHostControllerFactory {
                 throw ServerLogger.ROOT_LOGGER.errorCopyingFile(srcFile.getAbsolutePath(), destFile.getAbsolutePath(), e);
             }
         }
-
-}
+    }
 
     private static class HostControllerImpl implements HostController {
 
@@ -443,9 +442,7 @@ public class EmbeddedHostControllerFactory {
 
         private static HostControllerEnvironment createHostControllerEnvironment(File jbossHome, String[] cmdargs, String hostName, long startTime) {
             try {
-                // for SecurityActions.getSystemProperty("jboss.home.dir") in InstallationManagerService
-                System.setProperty("jboss.home.dir", jbossHome.getAbsolutePath());
-                WildFlySecurityManager.setPropertyPrivileged("jboss.home.dir", jbossHome.getAbsolutePath());
+                WildFlySecurityManager.setPropertyPrivileged(HostControllerEnvironment.HOME_DIR, jbossHome.getAbsolutePath());
 
                 Map<String, String> props = new HashMap<String, String>();
                 props.put(HostControllerEnvironment.HOME_DIR, jbossHome.getAbsolutePath());
--- a/embedded/src/main/java/org/wildfly/core/embedded/EmbeddedServerFactory.java
+++ b/embedded/src/main/java/org/wildfly/core/embedded/EmbeddedServerFactory.java
@@ -71,6 +71,28 @@ public class EmbeddedServerFactory {
     private static final String SYSPROP_KEY_JBOSS_MODULES_DIR = "jboss.modules.dir";
     private static final String SYSPROP_VALUE_JBOSS_LOGMANAGER = "org.jboss.logmanager.LogManager";
 
+    private static final String SYSPROP_KEY_JBOSS_SERVER_BASE_DIR = "jboss.server.base.dir";
+    private static final String SYSPROP_KEY_JBOSS_SERVER_CONFIG_DIR = "jboss.server.config.dir";
+    private static final String SYSPROP_KEY_JBOSS_SERVER_DEPLOY_DIR = "jboss.server.deploy.dir";
+    private static final String SYSPROP_KEY_JBOSS_SERVER_TEMP_DIR = "jboss.server.temp.dir";
+    private static final String SYSPROP_KEY_JBOSS_SERVER_LOG_DIR = "jboss.server.log.dir";
+    private static final String SYSPROP_KEY_JBOSS_SERVER_DATA_DIR = "jboss.server.data.dir";
+
+    private static final String SYSPROP_KEY_JBOSS_DOMAIN_BASE_DIR = "jboss.domain.base.dir";
+    private static final String SYSPROP_KEY_JBOSS_DOMAIN_CONFIG_DIR = "jboss.domain.config.dir";
+    private static final String SYSPROP_KEY_JBOSS_DOMAIN_DEPLOY_DIR = "jboss.domain.deploy.dir";
+    private static final String SYSPROP_KEY_JBOSS_DOMAIN_TEMP_DIR = "jboss.domain.temp.dir";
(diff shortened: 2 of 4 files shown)


luck3y:
There appears to be an issue with running the full testsuite that I'm investigating as part of this. Running each suite separately is fine, but running everything together appears to result in failures in the domain suite after completion of manualmode.

luck3y:
retest this please

embedded/src/main/java/org/wildfly/core/embedded/EmbeddedServerFactory.java
```diff
@@ -293,8 +296,14 @@ private static ModuleLoader setupModuleLoader(String modulePath, String... syste
             }
             WildFlySecurityManager.setPropertyPrivileged("jboss.modules.system.pkgs", packages.toString());
 
-            // Get the module loader
-            return Module.getBootModuleLoader();
+            // setup the module class so we can reload it and reset the static properties later if we need to
+            URL[] urls = new URL[] {new File(modulePath).toURI().toURL(), new File(modulePath + "/../jboss-modules.jar").toURI().toURL()};
```
bstansberry: The callers of this method should know where the root is. The EmbeddedServerFactory.create methods take separate params for jbossHomePath and modulePath, so the relationship between modulePath and jboss-modules.jar may not be fixed like this. But the relationship between jbossHomePath and jboss-modules.jar should be.

embedded/src/main/java/org/wildfly/core/embedded/EmbeddedServerFactory.java
```diff
@@ -293,8 +296,14 @@ private static ModuleLoader setupModuleLoader(String modulePath, String... syste
...
+            ClassLoader classloader = new URLClassLoader(urls, ClassLoader.getSystemClassLoader());
+
+            Class clazz = classloader.loadClass("org.jboss.modules.Module");
+            return (ModuleLoader)clazz.getMethod("getBootModuleLoader").invoke(clazz);
```
bstansberry: Nice.  You mentioned reflection and I envisioned something much more painful than this. :)

luck3y: Hackity hack ;)

bstansberry: I don't get how it works though. :) If the Module class is visible in the system classloader, URLClassLoader should be returning that as it's a parent first CL. And if its not visible and Module is loaded from urls[0], then how can the result of getBootModuleLoader be cast to the correct ModuleLoader?

luck3y:
I need to do some final testing to verify, but I think the class loading resets the boot module loader. It behaves like that in the CLI, but I do need to do a final verification.

luck3y:
@bstansberry So re-testing this I'm not seeing the same behavior as I had previously, which is odd. I'm taking another look at what's going on.

bstansberry:
I put the fixme label on this so it doesn't accidentally get merged.

n1hility:
retest this please (don't worry not trying to merge yet)

n1hility:
retest this please

luck3y:
This has been updated after additional debugging & testing to:

- verify that the modules directory exists before we try and start loading modues
- correct properties still pointing to the previous jboss-home when an embedded server is started in a different directory.

A couple of issues remain, I've mail wildfly-dev to ask about a couple of things:

- allowing modules to be re-initialized by some method, and whether this would be safe or not
- after an embedded server has been started, any subsequent servers continue to log to the same server.log.

If someone could review the properties reset, I'd appreciate it, as it wasn't completely clear to me that resetting them like this was entirely correct w.r.t. directory names that may have been configured differently elsewhere.

embedded/src/main/java/org/wildfly/core/embedded/EmbeddedHostControllerFactory.java
```diff
@@ -444,8 +426,8 @@ private void exit() {
         private static HostControllerEnvironment createHostControllerEnvironment(File jbossHome, String[] cmdargs, String hostName, long startTime) {
             try {
                 // for SecurityActions.getSystemProperty("jboss.home.dir") in InstallationManagerService
-                System.setProperty("jboss.home.dir", jbossHome.getAbsolutePath());
-                WildFlySecurityManager.setPropertyPrivileged("jboss.home.dir", jbossHome.getAbsolutePath());
+                System.setProperty(HostControllerEnvironment.HOME_DIR, jbossHome.getAbsolutePath());
+                WildFlySecurityManager.setPropertyPrivileged(HostControllerEnvironment.HOME_DIR, jbossHome.getAbsolutePath());
```
jamezp: Not fully reviewing this, but just caught this in passing. We should likely remove the `System.setProperty()` version if we need to do it privileged.

luck3y: @jamezp Thanks, done.

embedded/src/main/java/org/wildfly/core/embedded/EmbeddedHostControllerFactory.java
```diff
@@ -447,6 +446,8 @@ private static HostControllerEnvironment createHostControllerEnvironment(File jb
                 System.setProperty("jboss.home.dir", jbossHome.getAbsolutePath());
                 WildFlySecurityManager.setPropertyPrivileged("jboss.home.dir", jbossHome.getAbsolutePath());
 
+                WildFlySecurityManager.setPropertyPrivileged(HostControllerEnvironment.HOME_DIR, jbossHome.getAbsolutePath());
+
```
bstansberry: L446 and 447 should be removed as 449 does the same thing.

embedded/src/main/java/org/wildfly/core/embedded/EmbeddedServerFactory.java
```diff
@@ -338,4 +340,44 @@ private static void setupLoggingSystem(ModuleLoader moduleLoader) {
...
+                throw new RuntimeException("Unknown embedded server type: " + embeddedServerType);
+        }
+
+        WildFlySecurityManager.setPropertyPrivileged(SYSPROP_KEY_JBOSS_HOME_DIR, jbossHomeDir);
+        WildFlySecurityManager.setPropertyPrivileged(SYSPROP_KEY_JBOSS_BASE_DIR, jbossBaseDir);
+        WildFlySecurityManager.setPropertyPrivileged(SYSPROP_KEY_JBOSS_CONFIG_DIR, jbossBaseDir + File.separator + "configuration");
+        WildFlySecurityManager.setPropertyPrivileged(SYSPROP_KEY_JBOSS_DEPLOY_DIR, jbossBaseDir + File.separator + "data" + File.separator +"content");
+        WildFlySecurityManager.setPropertyPrivileged(SYSPROP_KEY_JBOSS_TEMP_DIR, jbossBaseDir + File.separator + "data" + File.separator + "tmp");
+        WildFlySecurityManager.setPropertyPrivileged(SYSPROP_KEY_JBOSS_LOG_DIR, jbossBaseDir + File.separator + "log");
+        WildFlySecurityManager.setPropertyPrivileged(SYSPROP_KEY_JBOSS_DATA_DIR, jbossBaseDir + File.separator + "data");
```
bstansberry: The jboss.server.xxx.dir properties should not be being set if the embedded process is a Host Controller. The HC equivalents are jboss.domain.xxx.dir.

embedded/src/main/java/org/wildfly/core/embedded/EmbeddedServerFactory.java
```diff
@@ -338,4 +340,44 @@ private static void setupLoggingSystem(ModuleLoader moduleLoader) {
...
+        Object serverImpl;
+        try {
+            Properties sysprops = WildFlySecurityManager.getSystemPropertiesPrivileged();
+            Map<String, String> sysenv = WildFlySecurityManager.getSystemEnvironmentPrivileged();
+            String[] args = cmdargs != null ? cmdargs : new String[0];
+            serverImpl = createServerMethod.invoke(null, jbossHomeDir, moduleLoader, sysprops, sysenv, args);
+        } catch (final InvocationTargetException ite) {
+            throw EmbeddedLogger.ROOT_LOGGER.cannotCreateStandaloneServer(ite.getCause(), createServerMethod);
+        } catch (final IllegalAccessException iae) {
+            throw EmbeddedLogger.ROOT_LOGGER.cannotCreateStandaloneServer(iae, createServerMethod);
```
bstansberry: This is a pre-existing issue, but might as well correct it now. These error messages are not correct in the HC case.

luck3y:
@bstansberry Thanks, I'll address these and resend right away.

PR #1267, opened 2015-11-24
[WFCORE-892] Set timeout parameter for cli client

mchoma:
Default value (5s) is not enough sometimes, especially on java IBM8. Parameter --timeout has to be same as in with waiting mechanism in CustomCLIExecutor.

JIRA: https://issues.jboss.org/browse/WFCORE-892


ctomc:
this is ok to test

PR #1268, opened 2015-11-24
https://issues.jboss.org/browse/WFCORE-877 fixes handling of the auth…

aloubyansky:
… info from the input stream

Thanks,
Alexey


aloubyansky:
The tests should be fixed now. There is one, imo, minor difference when authentication (the prompt for username and/or password) is interrupted with Ctrl-C. The difference is the stacktrace of InterruptedException is displayed on the screen. I am still going to clarify this with Stale and what can be done to hide it later tonight.

PR #1272, opened 2015-11-27
[WFCORE-1166] Make composite operation transformer transform path

kabir:
addresses the same as TransformersImpl.transformOperation()

https://issues.jboss.org/browse/WFCORE-1166

Once tests pass, I will create the downstream PR

Downstream JIRA: https://issues.jboss.org/browse/JBEAP-2029


bstansberry:
:+1:

PR #1273, opened 2015-11-30
 WFCORE-1163 Proposal

heiko-braun:
This is just for reviewing puposes.

### DIFF
--- a/server/src/main/java/org/jboss/as/server/parsing/ExtensionHandler.java
+++ b/server/src/main/java/org/jboss/as/server/parsing/ExtensionHandler.java
@@ -1,0 +1,47 @@
+/*
+ * JBoss, Home of Professional Open Source
+ * Copyright 2015, Red Hat, Inc., and individual contributors as indicated
+ * by the @authors tag.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.jboss.as.server.parsing;
+
+import java.util.List;
+import java.util.Set;
+
+import javax.xml.stream.XMLStreamException;
+
+import org.jboss.as.controller.parsing.Namespace;
+import org.jboss.as.controller.parsing.ProfileParsingCompletionHandler;
+import org.jboss.dmr.ModelNode;
+import org.jboss.staxmapper.XMLExtendedStreamReader;
+import org.jboss.staxmapper.XMLExtendedStreamWriter;
+
+/**
+ * An {@link ExtensionHandler} is a component of the {@link StandaloneXml} parser
+ * and allows for customisation of the routines to read and write XML extension elements.
+ * This includes the discovery of extension modules and the registration subsystem parsers that are associated with it.
+ * The regular {@link StandaloneXml} parser leverages a default implementation.
+ *
+ * @author Heiko Braun
+ * @since 27/11/15
+ */
+public interface ExtensionHandler {
+    void parseExtensions(XMLExtendedStreamReader reader, ModelNode address, Namespace namespace, List<ModelNode> list) throws XMLStreamException;
+
+    Set<ProfileParsingCompletionHandler> getProfileParsingCompletionHandlers();
+
+    void writeExtensions(XMLExtendedStreamWriter writer, ModelNode modelNode) throws XMLStreamException;
+}
--- a/server/src/main/java/org/jboss/as/server/parsing/StandaloneXml.java
+++ b/server/src/main/java/org/jboss/as/server/parsing/StandaloneXml.java
@@ -25,6 +25,7 @@ package org.jboss.as.server.parsing;
 import static org.jboss.as.controller.parsing.Namespace.CURRENT;
 
 import java.util.List;
+import java.util.Set;
 import java.util.concurrent.ExecutorService;
 
 import javax.xml.stream.XMLStreamException;
@@ -32,6 +33,7 @@ import javax.xml.stream.XMLStreamException;
 import org.jboss.as.controller.extension.ExtensionRegistry;
 import org.jboss.as.controller.parsing.ExtensionXml;
 import org.jboss.as.controller.parsing.Namespace;
+import org.jboss.as.controller.parsing.ProfileParsingCompletionHandler;
 import org.jboss.as.controller.persistence.ModelMarshallingContext;
 import org.jboss.dmr.ModelNode;
 import org.jboss.modules.ModuleLoader;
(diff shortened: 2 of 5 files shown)


heiko-braun:
@bstansberry It's slightly different as we discussed. But these would be the changes needed to re-use the top level parser in swarm. The relevant changes are:

- Encupsulated extension parsing into ExtensionHandler (swarm provides a cusrtom handler)
- Introduce parsing option to signal the profile parse that it can ignore unxpected subsystem elements

darranl:
For Swarm from this implementation I assume it is safe to say that this ability is something that can only be used with the most recent major version of the schema?  From the perspective of Swarm I wonder if it would be an idea to provide something that removes support for major versions 1, 2, and 3.

Related to my changes this does not look a huge task to merge - the only question though - my changes add StandaloneXml_5 - should swarm support 4 and 5 or just 5 if this is only being merged into WildFly 11?

server/src/main/java/org/jboss/as/server/parsing/StandaloneXml.java
```diff
@@ -46,14 +48,26 @@
  */
 public class StandaloneXml extends CommonXml {
```
bstansberry: I'd like this class to be final now, and CommonXml.parseDeployments should be made protected. That will lock down the API to just the constructors and the XMLElementReader<List<ModelNode>> and XMLElementWriter<ModelMarshallingContext> interfaces.

server/src/main/java/org/jboss/as/server/parsing/StandaloneXml.java
```diff
@@ -46,14 +48,26 @@
...
 
-    private final ExtensionXml extensionXml;
-    private final ExtensionRegistry extensionRegistry;
+    private final StandaloneXml_4.ParsingOptions parsingOptions;
```
bstansberry: These ParsingOptions shouldn't be in StandaloneXml_4. That class is not available for outside use.

server/src/main/java/org/jboss/as/server/parsing/StandaloneXml.java
```diff
@@ -46,14 +48,26 @@
...
+    private ModuleLoader loader;
+    private ExecutorService executorService;
+    private ExtensionRegistry extensionRegistry;
```
bstansberry: StandaloneXml_Legacy's constructor should take ExtensionHandler and then these fields are not needed.

server/src/main/java/org/jboss/as/server/parsing/StandaloneXml.java
```diff
@@ -46,14 +48,26 @@
...
+    private ExtensionHandler extensionHandler;
```
bstansberry: final

server/src/main/java/org/jboss/as/server/parsing/StandaloneXml.java
```diff
@@ -63,10 +77,14 @@ public void readElement(final XMLExtendedStreamReader reader, final List<ModelNo
             case 1:
             case 2:
             case 3:
-                new StandaloneXml_Legacy(extensionXml, extensionRegistry, readerNS).readElement(reader, operationList);
+                new StandaloneXml_Legacy(
+                        new ExtensionXml(loader, executorService, extensionRegistry),
+                        extensionRegistry,
+                        readerNS)
```
bstansberry: Parsing previous files from < WF 10 isn't allowed?

server/src/main/java/org/jboss/as/server/parsing/ExtensionHandler.java
```diff
@@ -0,0 +1,23 @@
...
+import java.util.Set;
+
+/**
+ * @author Heiko Braun
+ * @since 27/11/15
+ */
+public interface ExtensionHandler {
+    void parseExtensions(XMLExtendedStreamReader reader, ModelNode address, Namespace namespace, List<ModelNode> list) throws XMLStreamException;
+
+    Set<ProfileParsingCompletionHandler> getParsingCompletionHandler();
```
bstansberry: Use getProfileParsingCompletionHandlers() to be consistent with ExtensionRegistry. Plus you're missing an 's'. ;)

server/src/main/java/org/jboss/as/server/parsing/ExtensionHandler.java
```diff
@@ -0,0 +1,23 @@
...
+
+import javax.xml.stream.XMLStreamException;
+import java.util.List;
+import java.util.Set;
+
+/**
+ * @author Heiko Braun
+ * @since 27/11/15
+ */
+public interface ExtensionHandler {
```
bstansberry: This will need clear javadoc since it's becoming an external API. ExtensionRegistry and ExtensionXml were internal.

server/src/main/java/org/jboss/as/server/parsing/ExtensionHandler.java
```diff
@@ -0,0 +1,23 @@
+package org.jboss.as.server.parsing;
```
bstansberry: Please add the ASL v2 header.

server/src/main/java/org/jboss/as/server/parsing/ExtensionHandler.java
```diff
@@ -0,0 +1,23 @@
...
+
+import org.jboss.as.controller.parsing.Namespace;
+import org.jboss.as.controller.parsing.ProfileParsingCompletionHandler;
+import org.jboss.dmr.ModelNode;
+import org.jboss.staxmapper.XMLExtendedStreamReader;
+import org.jboss.staxmapper.XMLExtendedStreamWriter;
+
+import javax.xml.stream.XMLStreamException;
+import java.util.List;
+import java.util.Set;
```
bstansberry: WildFly codestyle for import ordering is java, then javax, then org

server/src/main/java/org/jboss/as/server/parsing/StandaloneXml_4.java
```diff
@@ -111,19 +110,22 @@
...
+    private final ParsingOptions parsingOptions;
     private AuditLogXml auditLogDelegate;
-
-    private final ExtensionXml extensionXml;
-    private final ExtensionRegistry extensionRegistry;
     private final Namespace namespace;
+    private ExtensionHandler extensionHandler;
+
+    public interface ParsingOptions{
+        boolean strict();
```
bstansberry: I've been making minor comments; this one is the only serious concern I have. This is very vague contract.

server/src/main/java/org/jboss/as/server/parsing/ExtensionHandler.java
heiko-braun: I guess you mean GPL, like the other files?

server/src/main/java/org/jboss/as/server/parsing/StandaloneXml.java
heiko-braun: In swarm, we don't use it. But I agree: We better make it consistent

server/src/main/java/org/jboss/as/server/parsing/StandaloneXml_4.java
heiko-braun: What would you suggest? Better terms? Any other improvement that comes to your mind?

server/src/main/java/org/jboss/as/server/parsing/ExtensionHandler.java
heiko-braun: OK, @darranl clarified it for me.

heiko-braun:
@bstansberry I've change it according to your comments. What about the ParsingOptions contract? Any suggestions how to improve it?

heiko-braun:
Regarding the roadmap: @bstansberry you've scheduled this for 3.x. Does this correspond to WF 11?

bstansberry:
Yes, 3.x is WildFly 11. Basically that label indicates the PR is for "the next feature release" as opposed to what we're currently merging, which is only bug fixes.

Re the ParsingOptions, off the top of my head I see two options:

1) Call the method something more specific like ignoreSubsystemFailures(). Then in the future if you want more exceptional behavior add more methods.

2) Rename it as lenient() and javadoc it explaining what it means and noting that in the future it may also trigger other forms of leniency. So if somehow we end up making this API available to other teams than just swarm at least they have a warning.

To me 1) sounds better than 2) as it can evolve in a more flexible way. It kind of depends on where you see this going. I don't feel really strongly either way though; my main concern is that there is a documented contract.

I think the name should imply leniency not strictness as being lenient is the non-standard behavior that needs to be explained.

heiko-braun:
@bstansberry ok, that makes sense. I prefer 1) as well.

bstansberry:
@heiko-braun Another option if you like it is to make ParsingOptions an enum and pass them in via varargs. Then we can just add more values without requiring anyone to recompile an interface implementation.

Any of these is fine though; I'm kind of overthinking this and brainstorming while on a boring call. ;)

heiko-braun:
@bstansberry I think this reads much better:

```
if(StandaloneXml.ParsingOption.IGNORE_SUBSYSTEM_FAILURES.isSet(this.parsingOptions)) {
[...]
}
```

bstansberry:
Ok, sounds good.

bstansberry:
@heiko-braun This looks fine now. All I'd ask is that you squash this down to a single commit with a message that references the JIRA and then it can get the Ready for Merge label and be merged once we open up master to future work.

How helpful would it be to have this in WildFly 10.x? There are some process hassles to go through to do that, but I'm not opposed to it technically. For the standard non-swarm use case this amounts to a minor refactor.

heiko-braun:
@bstansberry I am going to squash it and update the PR. With regard to WF 10: This would be really helpful. And since you don't see any problmes with this technically (me neither) then I would appreciate if we could include it.

heiko-braun:
Squashed everything into on commit

darranl:
I think 10 would be preferable, I am going to need to rework some of this for 11 anyway so if we get this merged I can then merge the latest into our branches again.

bstansberry:
Superceded by #1318 which was just this with a trivial merge conflict fixed. Thanks!

PR #1274, opened 2015-11-30
[WFCORE-313]: DefaultOperationDescriptionProvider uses incorrect call  to create reply parameter description.

ehsavoie:
Adding support for reply parameter description using a different text from the parameter.

Jira: https://issues.jboss.org/browse/WFCORE-313

https://github.com/wildfly/wildfly/pull/8440 is needed to have integration builds green.

http://brontes.lab.eng.brq.redhat.com/viewLog.html?buildId=80520&tab=buildResultsDiv&buildTypeId=WildFlyCore_PullRequest_WildFlyCoreFullIntegration
http://brontes.lab.eng.brq.redhat.com/viewLog.html?buildId=80523&tab=buildResultsDiv&buildTypeId=WildFlyCore_PullRequest_WildFlyCoreFullIntegration

### DIFF
--- a/controller/src/main/java/org/jboss/as/controller/ListAttributeDefinition.java
+++ b/controller/src/main/java/org/jboss/as/controller/ListAttributeDefinition.java
@@ -175,6 +175,15 @@ public abstract class ListAttributeDefinition extends AttributeDefinition {
         return result;
     }
 
+    @Override
+    public ModelNode addOperationReplyDescription(final ModelNode resourceDescription, final String operationName,
+                                                      final ResourceDescriptionResolver resolver,
+                                                      final Locale locale, final ResourceBundle bundle) {
+        final ModelNode result = super.addOperationReplyDescription(resourceDescription, operationName, resolver, locale, bundle);
+        addOperationReplyValueTypeDescription(result, operationName, resolver, locale, bundle);
+        return result;
+    }
+
     protected abstract void addValueTypeDescription(final ModelNode node, final ResourceBundle bundle);
 
     protected abstract void addAttributeValueTypeDescription(final ModelNode node, final ResourceDescriptionResolver resolver,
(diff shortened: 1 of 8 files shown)


controller/src/main/java/org/jboss/as/controller/ListAttributeDefinition.java
```diff
@@ -184,6 +193,10 @@ protected abstract void addOperationParameterValueTypeDescription(final ModelNod
                                                                       final ResourceDescriptionResolver resolver,
                                                                       final Locale locale, final ResourceBundle bundle);
 
+    protected abstract void addOperationReplyValueTypeDescription(final ModelNode node, final String operationName,
+                                                                      final ResourceDescriptionResolver resolver,
+                                                                      final Locale locale, final ResourceBundle bundle);
+
```
bstansberry: This and the new method in MapAttributeDefinition concern me a bit since they will break subclasses. This won't go in until core 3.0, and that's a new major, so technically that's ok. But there's such a high chance that this will end up in a minor version of some downstream software, e.g. a WildFly 10.1 or an EAP 7.1, and then we'll have hassles if we break external extensions (e.g. Teiid, JDV).

WDYT about having these by default call addOperationParameterValueTypeDescription ? Perhaps with a TODO pointing at a JIRA to get rid of that eventually (i.e. for EAP 8.) At a glance it looks like the impls you added here and in the full wildfly PR end up executing the same code as addOperationParameterValueTypeDescription.

ehsavoie: In the full WildFly PR I call *getOperationReplyValueTypeDescription()* instead of *getOperationParameterValueTypeDescription()* . While getOperationReplyValueTypeDescription() falls back to getOperationParameterValueTypeDescription() if no specific reply description is found this is a difference.

bstansberry:
retest this please

PR #1276, opened 2015-11-30
[WFCORE-1171] Register the capability in the MRR; drop from the OSHs

bstansberry:


bstansberry:
I added the commit from #1277 as its absence was causing tests to fail on this one.

PR #1279, opened 2015-11-30
[WFCORE-1161] Provide an MRR-driven CapabilityReferenceRecorder

bstansberry:
I also enabled a test that I noticed was disabled when I was playing with testing this.

Note that this PR doesn't change any of the usage of setCapabilityReference in core. That's because this method of determining the dependent is more expensive, so I didn't want to change working code to something more expensive.


bstansberry:
retest this please

PR #1287, opened 2015-12-07
[WFCORE-1232] adjust patching testsuite after WFCORE-838

jmartisk:
Test merged patches (introduced by https://issues.jboss.org/browse/WFCORE-838) in addition to the old patch-bundles. Also made CLIWrapper implement AutoCloseable to shorten some pieces of code.

https://issues.jboss.org/browse/WFCORE-1232

### DIFF
--- a/testsuite/patching/src/test/java/org/jboss/as/test/patching/MergedPatchesTestCase.java
+++ b/testsuite/patching/src/test/java/org/jboss/as/test/patching/MergedPatchesTestCase.java
@@ -1,0 +1,379 @@
+package org.jboss.as.test.patching;
+
+import static org.jboss.as.patching.IoUtils.mkdir;
+import static org.jboss.as.patching.IoUtils.newFile;
+import static org.jboss.as.test.patching.PatchingTestUtil.AS_DISTRIBUTION;
+import static org.jboss.as.test.patching.PatchingTestUtil.AS_VERSION;
+import static org.jboss.as.test.patching.PatchingTestUtil.FILE_SEPARATOR;
+import static org.jboss.as.test.patching.PatchingTestUtil.MODULES_PATH;
+import static org.jboss.as.test.patching.PatchingTestUtil.PRODUCT;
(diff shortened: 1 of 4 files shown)


testsuite/patching/src/test/java/org/jboss/as/test/patching/MergedPatchesTestCase.java
```diff
@@ -30,87 +16,75 @@
...
+import static org.jboss.as.patching.IoUtils.mkdir;
+import static org.jboss.as.patching.IoUtils.newFile;
+import static org.jboss.as.test.patching.PatchingTestUtil.AS_DISTRIBUTION;
+import static org.jboss.as.test.patching.PatchingTestUtil.AS_VERSION;
+import static org.jboss.as.test.patching.PatchingTestUtil.FILE_SEPARATOR;
+import static org.jboss.as.test.patching.PatchingTestUtil.MODULES_PATH;
+import static org.jboss.as.test.patching.PatchingTestUtil.PRODUCT;
+import static org.jboss.as.test.patching.PatchingTestUtil.createPatchXMLFile;
+import static org.jboss.as.test.patching.PatchingTestUtil.createZippedPatchFile;
+import static org.jboss.as.test.patching.PatchingTestUtil.randomString;
```
bstansberry: The previous import ordering was correct per WildFly standards. Please adjust your IDE to either not reorder or to use the standard order. Thanks.

bstansberry:
@aloubyansky Please approve when you get time. The CLIWrapper part seems fine, but I don't know what you think about the patching test part.

aloubyansky:
The new tests look valid to me. Perhaps, I'm missing something here but the question I have is why a re-write instead of a separate test for the new feature? We are still going to support bundles, so what's the reason to reduce the test coverage?

jmartisk:
Okay then, I will copy this into a new test case and leave the old one there.

jmartisk:
retest this please

jmartisk:
I created both jiras, reverted the old test case and fixed imports.

PR #1289, opened 2015-12-07
[WFCORE-1191][WFCORE-1189] More permgen tweaks

bstansberry:
Deprecate the permgen attributes in the management API and handle them consistently when launching servers.

There's still an open task (WFCORE-1190) to deprecate them in the xsd but I want to wait for a new xsd version for that, so that can come in 3.x.

Downstream: https://issues.jboss.org/browse/JBEAP-2229


bstansberry:
retest this please

PR #1290, opened 2015-12-08
[WFCORE-1159]: Add a runtime metric which shows filesystem usage and availability for known server locations.

ehsavoie:
Adding metrics to all paths and to server environment.

Jira: https://issues.jboss.org/browse/WFCORE-1159


controller/src/main/java/org/jboss/as/controller/services/path/PathSizeMetricHandler.java
```diff
@@ -0,0 +1,201 @@
...
+ * modify it under the terms of the GNU Lesser General Public
+ * License as published by the Free Software Foundation; either
+ * version 2.1 of the License, or (at your option) any later version.
+ *
+ * This library is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the GNU
+ * Lesser General Public License for more details.
+ *
+ * You should have received a copy of the GNU Lesser General Public
```
ctomc: This is minor, but all new files added to wildfly-core should be added under Apache 2 license...

controller/src/main/java/org/jboss/as/controller/services/path/PathResourceDefinition.java
```diff
@@ -134,6 +133,7 @@ public void registerAttributes(ManagementResourceRegistration resourceRegistrati
         SimpleAttributeDefinition pathAttr = specified ? PATH_SPECIFIED : PATH_NAMED;
         resourceRegistration.registerReadWriteAttribute(pathAttr, null, new PathWriteAttributeHandler(pathManager, pathAttr, services));
         resourceRegistration.registerReadOnlyAttribute(READ_ONLY, null);
+        PathSizeMetricHandler.registerMetric(resourceRegistration, pathManager, PATH, RELATIVE_TO);
```
bstansberry: We need to narrow down which variants of of PathResourceDefinition we register this, as it's not meaningful in some. For example, it's not for meaningful for domain.xml path resources (NamedPathResourceDefinition) because there is no such thing as a domain-wide FS location. It's questionable for host.xml ones too; e.g. a path that is relative-to jboss.server.base.dir means nothing in the HC process; it's only when it gets converted to a server process that it means something.

bstansberry: I see you partially address my concern in PathSizeMetricHandler.registerMetric by skipping server-config.

bstansberry: I wonder if this should be a custom operation instead of a metric. The file walking can be expensive if a path points to the root of a large tree, so calculating it when people/tools do common things like :read-resource(include-runtime=true) can have unintended effects.

ehsavoie: I've removed all named path. But for host path I'm not sure since those could be "the" same across all servers of a given host.

controller/src/main/java/org/jboss/as/controller/services/path/PathInfoHandler.java
```diff
@@ -0,0 +1,428 @@
...
+    private static final String LAST_MODIFIED = "last-modified";
+    private static final String CREATION_TIME = "creation-time";
+    private static final String OPERATION_NAME = "path-info";
+    private static final String FILESYSTEM = "filesystem";
+
+    private static final AttributeDefinition UNIT_ATTRIBUTE = SimpleAttributeDefinitionBuilder
+                        .create(UNIT, ModelType.STRING, true)
+                        .setAllowedValues(BYTES.getName(), KILOBYTES.getName(), MEGABYTES.getName(), GIGABYTES.getName(),
+                                TERABYTES.getName(), PETABYTES.getName(), BITS.getName(), KILOBITS.getName(),
+                                MEGABITS.getName(), GIGABITS.getName(), TERABITS.getName(), PETABITS.getName())
```
bstansberry: You can use EnumValidator for this, which via a hack also results in allowed-values being set. That validator is also forgiving about case sensitivity. If you do that though when sizeUnit = MeasurementUnit.valueOf(unitModelNode.asString()) you have to convert the string to upper case.

PR #1291, opened 2015-12-08
[WFCORE-1193] Override all methods of Parameters so that the child type can be returned.

darranl:
The first alternative option is do nothing, side effect of this is any resource that extends PersistentResourceDefinition will need to keep casting the Parameters back.

The other extreme is to re-design the Parameters hierarchy - if we do that however we need to do it before WildFly Core 2.0.0.Final is tagged as any subsystem developed externally is going to be sensitive to changes here.


ctomc:
:+1:

bstansberry:
Looks fine. Too late for "the re-design the Parameters hierarchy" option as 2.0.0.Final was released long ago.

darranl:
Sorry you're right - mixing it up with the WildFly release.

PR #1292, opened 2015-12-08
WFCORE-1194 Name for wildcard resources should be configurable in parser

ctomc:


controller/src/main/java/org/jboss/as/controller/PersistentResourceXMLDescription.java
```diff
@@ -518,6 +522,11 @@ public PersistentResourceXMLBuilder setMarshallDefaultValues(boolean marshallDef
             return this;
         }
 
+        public PersistentResourceXMLBuilder setNameAttributeName(String nameAttributeName) {
```
bstansberry: This one could use some javadoc as it's not so obvious what it's about. :)

bstansberry:
Javadoc bit aside, looks fine.

ctomc:
javadoc was added.

darranl:
:warning: It has been a while since we talked about this one so just adding a note to say this is the one we need to be sure the SHA of the commit does not change as it is included in a shared topic branch.

ctomc:
I do not plan to change SHA of this commit. 
This should be merged quite soon, given it got all acks for inclusion in EAP.

PR #1293, opened 2015-12-09
Updated remoting Port to work with Wildfly 10.0.0.CR4

mdorenkamp:
I wasn't able to shutdown a Wildfly running as Windows Service created using the Service-Package from contrib directory. The reason was using the old remoting port which has been changed from 9990 to 9999.


ctomc:
this is ok to test

mdorenkamp:
obsolete due to wrong configuration in standalone.xml

PR #1294, opened 2015-12-09
[WFCORE-1198] fix CLI does not resolve multiple properties if one pro…

soul2zimate:
…perty is undefined

https://issues.jboss.org/browse/WFCORE-1198 please see comment in JIRA, thanks

### DIFF
--- a/cli/src/main/java/org/jboss/as/cli/parsing/ExpressionBaseState.java
+++ b/cli/src/main/java/org/jboss/as/cli/parsing/ExpressionBaseState.java
@@ -67,9 +67,10 @@ public class ExpressionBaseState extends DefaultParsingState {
             @Override
             public void handle(ParsingContext ctx)
                     throws CommandFormatException {
+                final int originalLength = ctx.getInput().length();
                 ctx.resolveExpression(resolveSystemProperties, exceptionIfNotResolved);
                 final char resolvedCh = ctx.getCharacter();
-                if(resolvedCh == '$') {
+                if(resolvedCh == '$' && originalLength == ctx.getInput().length()) {
                     getDefaultHandler().handle(ctx);
                 } else {
                     getHandler(resolvedCh).handle(ctx);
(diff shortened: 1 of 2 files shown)


soul2zimate:
retest this please

bstansberry:
@aloubyansky please approve and if ok, ack https://issues.jboss.org/browse/JBEAP-2379

@soul2zimate a wildfly-core-eap PR is needed as well.

aloubyansky:
Thanks for looking into this! I am not sure an empty value should result in an error, though. It can legitimately be defined as you did in your use-case, why shouldn't we support them?

aloubyansky:
Thanks! Functionally I prefer the updated version. Although, it's a bit overcomplicated, imo. I'd propose this:

--- a/cli/src/main/java/org/jboss/as/cli/parsing/ExpressionBaseState.java
+++ b/cli/src/main/java/org/jboss/as/cli/parsing/ExpressionBaseState.java
@@ -67,9 +67,10 @@ public class ExpressionBaseState extends DefaultParsingState {
             @Override
             public void handle(ParsingContext ctx)
                     throws CommandFormatException {
+                final int originalLength = ctx.getInput().length();
                 ctx.resolveExpression(resolveSystemProperties, exceptionIfNotResolved);
                 final char resolvedCh = ctx.getCharacter();
-                if(resolvedCh == '$') {
+                if(resolvedCh == '$' && originalLength == ctx.getInput().length()) { // in case the property couldn't be resolved
                     getDefaultHandler().handle(ctx);
                 } else {

Does it make sense to you? Would be also good to have a test case for it.

soul2zimate:
Yes, it's correct and more straightforward, I have updated the change and add two simple tests with an empty property to cover it. thanks

aloubyansky:
Yes, this looks good to me. Sorry for the late reaction again. And thanks a lot for your efforts!

PR #1297, opened 2015-12-09
[WFCORE-1199] CLI Lists in non-interactive mode are erroneously split…

ryanemerson:
… into multiple commands

JIRA: https://issues.jboss.org/browse/WFCORE-1199

### DIFF
--- a/cli/src/main/java/org/jboss/as/cli/handlers/DeploymentOverlayHandler.java
+++ b/cli/src/main/java/org/jboss/as/cli/handlers/DeploymentOverlayHandler.java
@@ -41,6 +41,7 @@ import org.jboss.as.cli.CommandLineException;
 import org.jboss.as.cli.Util;
 import org.jboss.as.cli.accesscontrol.AccessRequirement;
 import org.jboss.as.cli.accesscontrol.AccessRequirementBuilder;
+import org.jboss.as.cli.impl.ArgumentWithListValue;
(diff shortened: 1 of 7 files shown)


jamezp:
@aloubyansky Could you please give your thoughts on this?

bstansberry:
@aloubyansky if this is ok, please hit the Devel Approve button (aka dev_ack) on the JBEAP as well.

aloubyansky:
Sorry for the delay.
While I am more or less fine with the syntax of grouping the list (the parser actually supports it) I don't like the fact that every argument will have to check whether the value is enclosed or not.

This kind of check could possibly be built into ArgumentWithValue.getValue(...), so it automatically covers all the values. A subclass, e.g. ArgumentWithListValue, would probably be even better.

There is actually a workaround, you can simply escape the comma. It's not as convenient as wrapping the whole list but it's there, just in case.

ryanemerson:
@aloubyansky thanks for the feedback and sorry for the slow reply (PTO).  I will rework this to use a subclass ASAP.

bstansberry:
retest this please

bstansberry:
retest this please

@aloubyansky is this ok now? (I'm assuming the test failures are unrelated; we are seeing failures in those suites elsewhere.)

aloubyansky:
Yes, looks good, thanks!

PR #1302, opened 2015-12-11
[WFCORE-1209] validate operation request in BatchRunHandler to fix invalid attribute ignoring in batch

soul2zimate:
https://issues.jboss.org/browse/WFCORE-1209 "CLI did not throw any error message and just ignored the invalid attribute while using in batch mode.If we use same CLI command in non-batch mode it throws exception."

Move validateRequest() and replaceFilePathsWithBytes() into Util class to allow BatchRunHandler validate operation request in BatchRunHandler as OperationRequestHandler does.


bstansberry:
@aloubyansky please approve

aloubyansky:
Looks good to me, thanks!

PR #1305, opened 2015-12-11
Improve handling of critical errors in domain operation handling

bstansberry:
Two commits related to better handling domain operation errors:

1) WFCORE-1134 which improves how the management op execution handles java.lang.Error. The most meaningful bit is TransactionalProtocolOperationHandler now attempts to communicate the Error to the remote caller (i.e. the HC that made the call.)

This may not prove all that helpful in practice, particularly in OOME cases, as the attempt to notify the remote caller may itself fail due to the OOME condition. But it is better behavior for Errors that do not prevent further work, as the test case in the commit shows.

2) WFCORE-378. This is the more significant change. It introduces timeouts into various blocking calls used in domain op execution, preventing a non-responsive process from locking up the domain indefinitely.

### DIFF
--- a/controller/src/main/java/org/jboss/as/controller/BlockingTimeout.java
+++ b/controller/src/main/java/org/jboss/as/controller/BlockingTimeout.java
@@ -22,80 +22,88 @@
 
 package org.jboss.as.controller;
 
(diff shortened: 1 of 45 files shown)


bstansberry:
retest this please

PR #1306, opened 2015-12-11
[bz-1196343] LDAP authentication/authorization fails due to making unneeded role lookup requests

dehort:
upstream: Wildfly: https://issues.jboss.org/browse/WFCORE-1202
wildfly/wildfly-core#1306
downstream: EAP 7: https://issues.jboss.org/browse/JBEAP-2320
way-downstream: https://bugzilla.redhat.com/show_bug.cgi?id=1196343
https://github.com/jbossas/jboss-eap/pull/2654

### DIFF
--- a/domain-management/src/main/java/org/jboss/as/domain/management/security/LdapGroupSearcherFactory.java
+++ b/domain-management/src/main/java/org/jboss/as/domain/management/security/LdapGroupSearcherFactory.java
@@ -30,7 +30,6 @@ import java.util.ArrayList;
 import java.util.Arrays;
 import java.util.HashSet;
 import java.util.Set;
-import java.util.StringTokenizer;
 
 import javax.naming.NamingEnumeration;
 import javax.naming.NamingException;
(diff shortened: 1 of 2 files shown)


dehort:
Retest

bstansberry:
@darranl please approve. If you agree, please ack https://issues.jboss.org/browse/JBEAP-2320 as well just to keep the paperwork straight. (The BZ is already acked, which is sufficient to merge this if the change itself is ok.)

darranl:
I think I would prefer a config change in core and just use the system property for compatibility with those migrating from older versions - having said that the PR for the next major schema version is currently held up waiting for WildFly 11.

Also I think we need a bit more detail in the Jira issue.

Other than that - looks good.

darranl:
I think both Jira issues need additional information adding.

Also I think the title should be more accurate, the lookup call is only unneeded if the role can be extracted from the DN.

bstansberry:
+1 re: there needing to be a proper config item in the management schema. @dehort or @darranl please file a WFCORE for that. It's correct though that this PR doesn't involve that as it an forward port of a fix to EAP 6.4.

@dehort please improve the WFCORE-1022 content.

darranl:
Once merged I will follow up on https://issues.jboss.org/browse/WFCORE-1228

PR #1308, opened 2015-12-12
[WFCORE-1213] Don't send a suspend op in the kill/destroy cases

bstansberry:


host-controller/src/main/java/org/jboss/as/host/controller/ManagedServer.java
```diff
@@ -858,7 +858,7 @@ public boolean execute(ManagedServer server) throws Exception {
             try {
                 //graceful shutdown
                 //this just suspends the server, it does not actually shut it down
-                if (operationID != -1) {
+                if (permit != -1) {
```
bstansberry: @stuartwdouglas I think using operationID instead of permit here was inadvertent.

PR #1313, opened 2015-12-15
[WFCORE-991]: Starting server with config that doesn't end with .xml fails

ehsavoie:
Fixing how the filename is computed.

Jira: https://issues.jboss.org/browse/WFCORE-991


ehsavoie:
retest this please

PR #1304, opened 2015-12-11
[DO NOT MERGE] Jboss modules 1.5

ctomc:
testing to see if we dont have any regressions

PR #1314, opened 2015-12-15
[WFCORE-1219] Upgrade to JBoss Modules 1.5.0.CR1

dmlloyd:


bstansberry:
@ctomc how do the other changes in #1304 relate to this? Are they necessary to let this upgrade work, or is it more a case of them doing things in a better way now that JBoss Modules makes that possible?

bstansberry:
@ctomc lol; i guess ^^^ answers my question!

bstansberry:
I'm going to close this one as the added commit in #1315 fixes the problem.

PR #1315, opened 2015-12-15
Pull request 1304

bstansberry:
Takes PR #1304 and swaps out the modules snapshot pom change with the real upgrade commit from #1314


bstansberry:
I tweaked the 2nd commit to remove the currently commented out use of JMX stuff, plus whitespace changes. That makes the bit needed to let module 1.5.0 to work easier to see. The JMX stuff can come later.

ctomc:
retest this please

PR #1317, opened 2015-12-16
[WFCORE-1235] : Multiple triggers of each activator service when there are multiple modules with different service activator files in an ear.

panossot:
[WFCORE-1235] : https://issues.jboss.org/browse/WFCORE-1235

### DIFF
--- a/server/src/main/java/org/jboss/as/server/deployment/service/ServiceActivatorProcessor.java
+++ b/server/src/main/java/org/jboss/as/server/deployment/service/ServiceActivatorProcessor.java
@@ -19,9 +19,11 @@
  * Software Foundation, Inc., 51 Franklin St, Fifth Floor, Boston, MA
  * 02110-1301 USA, or see the FSF site: http://www.fsf.org.
  */
-
 package org.jboss.as.server.deployment.service;
 
+import java.util.ArrayList;
+import java.util.List;
+import org.jboss.as.server.deployment.AttachmentList;
 import org.jboss.as.server.deployment.Attachments;
 import org.jboss.as.server.deployment.DeploymentPhaseContext;
 import org.jboss.as.server.deployment.DeploymentUnit;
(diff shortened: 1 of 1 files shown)


panossot:
Retest this please.

bstansberry:
retest this please

Failure was unrelated.

dpospisil:
Could someone verify this patch? Thanks.

panossot:
retest this please

Failure was unrelated.

panossot:
retest this please

Failure was unrelated.

panossot:
Failure is unrelated

msochure:
Retest this please.

panossot:
retest this please

Failure was unrelated.

stuartwdouglas:
I do not think this PR fixes the underlying problem. If two deployments have visibility to each others services directories then it is still possible for the ServiceActivator to be run twice.

panossot:
@stuartwdouglas Could you review again?

panossot:
Retest this please.

When run locally tests pass.

panossot:
SingletonServiceTestCase fails. I will have to rethink it.

panossot:
@stuartwdouglas Could you review again? (Solution without VFS)

panossot:
Retest this please.

Local tests pass.

panossot:
Retest this please.

Irrelevant failures.

panossot:
Retest this please.

Irrelevant failures.

bstansberry:
@dmlloyd I'd like to get this one approved or rejected.

Here's my read on what it is doing:

1) ServiceActivatorProcessor, in Phase.INSTALL, is checking for a SUB_DEPLOYMENTS attachment, and if present, is accessing the subdeployment's SERVICES attachment and adding the names of the impl class of ServiceActivator impls found there. So the result is a list of subdeployment ServiceActivator impls that were recording in Phase.PARSE by ServiceLoaderProcessor.

2) Later it loops over all the ServiceActivator impls loadable via ServiceLoader from the current deployment's module.

a) It loops through the name of ServiceActivator impls recorded in the current deployment's SERVICES attachment

 i) if the SA impl in outer loop 2) isn't included in the list 2a), it won't get activated. I don't know if this is a feature, a bug or just superfluous. IOW I don't know what it means that the impl is found by ServiceLoader but wasn't listed in the SERVICES attachment. My guess is it's a feature; we don't want to activate SA impls just because they are visible via ServiceLoader; we only want them if ServiceLoaderProcessor decided they were relevant.

b) else if the name of the SA impl is included in the list of subdeployment SA impls created in 1) above, the SA is not activated. Otherwise it is.

The effect of 2b) is that if the same SA impl is declared in both the parent level and the subdeployment level, it doesn't get activated in the parent level. Here I don't know if this behavior is a feature or necessary to get standard behavior to work. IOW is it required for some reason that the user has a META-INF/services file for the SA impl in both the parent and the subdeployment? If it isn't required, do we want to support being "helpful" here, in the way 2b) is? (I vote no.)

Apologies if some of this has been discussed in past email threads or issue comments. But they haven't resulted in this PR being approved or rejected, and it's time to do one or the other.

dmlloyd:
Your assessment in 2a) seems correct to me.  The problem stems from subdeployments which all "see" one another, which in turn cases them to have the same list of activators, which in turn causes each one to be started one time for each subdeployment.

For your 2b) question, if a user has an SA in both parent and subdeployment, then it doesn't really make a lot of sense no matter what we do.  Activating the service on the subdeployment but not the parent/other deployments seems like OK behavior to me, but you could easily argue the opposite.  I wouldn't call it a "feature" in any event.

bstansberry:
@dmlloyd re 2b), yes, "feature" is too strong a word. Basically I just mean being nice and working around something that is actually an incorrect deployment. It's friendly to do but similar to a full on feature it generates an obligation to continue to do that, which limits future flexibility.

I don't feel strongly about this though.

bstansberry:
Meh, the alternative is to detect this and throw a special exception, which is added work. Ignoring it and waiting for an DuplicateServiceException with no explanation why there's a dup is poor. So, I'll merge it.

PR #1319, opened 2015-12-17
[WFCORE-1028]: Poor handling of invalid roles.

ehsavoie:
Getting the authorization error passed to the handler so it can get back cleanly.

Jira: https://issues.jboss.org/browse/WFCORE-1028

### DIFF
--- a/controller/src/main/java/org/jboss/as/controller/operations/global/GlobalOperationHandlers.java
+++ b/controller/src/main/java/org/jboss/as/controller/operations/global/GlobalOperationHandlers.java
@@ -63,6 +63,7 @@ import org.jboss.as.controller.UnauthorizedException;
 import org.jboss.as.controller.access.Action;
 import org.jboss.as.controller.access.AuthorizationResult;
 import org.jboss.as.controller.access.ResourceNotAddressableException;
+import org.jboss.as.controller.access.rbac.UnknowRoleException;
 import org.jboss.as.controller.logging.ControllerLogger;
 import org.jboss.as.controller.operations.common.Util;
 import org.jboss.as.controller.registry.AliasEntry;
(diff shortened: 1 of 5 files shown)


controller/src/main/java/org/jboss/as/controller/operations/global/GlobalOperationHandlers.java
```diff
@@ -278,6 +244,58 @@ public void handleResult(OperationContext.ResultAction resultAction, OperationCo
...
+            public MultiTargetResultHandler(ModelNode fakeOperationResponse, FilteredData localFilteredData, ModelNode result) {
+                this.localFilteredData = localFilteredData;
+                this.result = result;
+                this.fakeOperationResponse = fakeOperationResponse;
+            }
+
+            @Override
+            public void handleResult(OperationContext.ResultAction resultAction, OperationContext context, ModelNode operation) {
+                if(fakeOperationResponse != null && fakeOperationResponse.hasDefined(FAILURE_DESCRIPTION)) {
+                    context.setRollbackOnly();
```
bstansberry: Is this necessary? I don't think it should be needed, since whatever step populated fakeOperationResponse has already reported a problem. So I think it's out-of-scope for this result handler to be manipulating the OC's resultAction.

controller/src/main/java/org/jboss/as/controller/operations/global/GlobalOperationHandlers.java
```diff
@@ -778,10 +796,12 @@ protected void executeSingleTargetChild(PathAddress base, PathElement currentEle
 
         @Override
         protected boolean authorize(OperationContext context, PathAddress base, ModelNode operation) {
-//            if (base.size() > 0) {
-                //We should always be able to read the root resource
+            try {
                 context.readResource(base, false);
-//            }
+            } catch(IllegalArgumentException ex) {
```
bstansberry: This is kind of overkill but I wonder if it would be better if ControllerLogger.unknownRole() threw some sort of custom RuntimeException instead of IAE. Documented in the RoleMapper.canRunAs javadoc.

Catching general purpose exceptions that aren't documented smells funny. Granted, OperationContext.readResource wouldn't document the exception either, so really the only benefit of a custom exception is to avoid catching unrelated problems.

WDYT?

ehsavoie: I agree, it's cleaner this way

PR #1324, opened 2015-12-21
WFCORE-1229 - Reloading the embed host controller logs ERROR messages

luck3y:
This stops the logging of the host controller not being registered during polling by ReloadHandler.ensureServerRebootComplete(). I'm not sure if this is actually correct in all cases, @bstansberry thoughts? Is this too broad here or are their other things that could fail to report errors during boot because of this?


bstansberry:
For now it's probably safe enough, as the https://issues.jboss.org/browse/WFCORE-184 "fix" ensures that requests from remote management clients get a NoSuchResourceException until boot is complete, so they never even get to DefaultPrepareStepHandler. It's only in-VM clients that have this problem. Besides the embedded clients, the only in-VM one I can think of is the deployment scanner.

Longer term changing AbstractOperationContext to only log these kinds of boot failures if that OC instance is actually handling a boot op is probably the way to go though, as this could definitely break things if other in-VM clients start to appear. For example, a feature like https://issues.jboss.org/browse/PRODMGT-1288 would likely involve handing during boot an in-vm client to a user-registered listener, which *could* then make requests that would fail but not get an exception.

luck3y:
Aha! That makes a lot more sense now! Thanks!

PR #1329, opened 2015-12-22
Execute the server reload op that the test creates

bstansberry:
Just noticed this test isn't doing what it says because it never executes this op.


bstansberry:
retest this please

ctomc:
retest this please

bstansberry:
retest this please

PR #1332, opened 2015-12-23
[WFCORE-761] Not possible to overlay non existing file in WAR

gaol:
Jira: https://issues.jboss.org/browse/WFCORE-761

This is part of the fix on the overlay issues, the second part of fix and the test cases PR is at: https://github.com/wildfly/wildfly/pull/8528

### DIFF
--- a/server/src/main/java/org/jboss/as/server/deployment/DeploymentOverlayDeploymentUnitProcessor.java
+++ b/server/src/main/java/org/jboss/as/server/deployment/DeploymentOverlayDeploymentUnitProcessor.java
@@ -40,10 +40,10 @@ import java.util.Map;
 import java.util.Set;
 
 import org.jboss.as.repository.ContentRepository;
-import org.jboss.as.server.deploymentoverlay.DeploymentOverlayIndex;
-import org.jboss.as.server.logging.ServerLogger;
 import org.jboss.as.server.deployment.module.ResourceRoot;
 import org.jboss.as.server.deployment.module.TempFileProviderService;
+import org.jboss.as.server.deploymentoverlay.DeploymentOverlayIndex;
+import org.jboss.as.server.logging.ServerLogger;
 import org.jboss.vfs.VFS;
 import org.jboss.vfs.VirtualFile;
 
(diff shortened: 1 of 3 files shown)


server/src/main/java/org/jboss/as/server/deployment/DeploymentOverlayDeploymentUnitProcessor.java
```diff
@@ -65,6 +66,8 @@
 
     private final ContentRepository contentRepository;
 
+    private static final String WAR_EXTENSION = ".war";
```
stuartwdouglas: We should not be relying on WAR_EXTENSION. Instead I think this should just use org.jboss.as.server.deployment.MountExplodedMarker to check if it is a deployment that is mounted exploded.

gaol: Only in the case of overlay for resources like `war/some.file` within an ear deployment, the method `isExplodedSubUnitOverlay(DeploymentUnit deploymentUnit, VirtualFile mountPoint, String path)` is called(), the `deploymentUnit` in the parameters is the top level `ear` deployment, the  `org.jboss.as.server.deployment.MountExplodedMarker.isMountExploded(DeploymentUnit deploymentUnit)` will return false eventually, so this does not work.

After more research, I have 2 options:

* I see that the `Attachments.INDEX_RESOURCE_ROOT` will be attached to the war resource root by [EarStructureProcessor](https://github.com/wildfly/wildfly/blob/master/ee/src/main/java/org/jboss/as/ee/structure/EarStructureProcessor.java#L279), so maybe that can be used to check whether some resources in war of ear are overlayed, like:

<pre>
&& resourceRoot.hasAttachment(Attachments.INDEX_RESOURCE_ROOT)){...}
</pre>

>  This option is more tricky, but only changes in wildfly-core can solve the problem.

* Change the `MountExplodedMarker` to be able to set exploded flag for `ResourceRoot` too, and in `EarStructureProcessor`, puts exploded flag to the war resource root, then check the exploded flag in `DeferredDeploymentOverlayDeploymentUnitProcessor`

> This option needs both changes in wildfly-core and wildfly to solve the problem.

If one of above option is OK, which one do you prefer? Thank you.

gaol:
@stuartwdouglas If I adopt option 2 in above comment, the `MountExplodedMarker` will be changed to attach the Attachment to more than one place, which will be kind of confusing(as you said).

So I removed the 'WAR_EXTENSION' by checking the `Attachments.INDEX_RESOURCE_ROOT` in the sub-ResourceRoot. Would you please take a look at this again? Thank you. :-)

gaol:
retest this please

gaol:
retest this please

gaol:
retest this please

gaol:
@stuartwdouglas I removed the 'WAR_EXTENSION' by checking the Attachments.INDEX_RESOURCE_ROOT in the sub-ResourceRoot. Would you please take a look whether it is OK? Thank you. :-)

gaol:
Updated with an explicit marker `SubExplodedDeploymentMarker` to mark the sub ResourceRoot which will be mounted exploded.

Another PR: https://github.com/wildfly/wildfly/pull/8528 will be updated too to use the new marker for war inside of ear.

stuartwdouglas:
I think this looks ok

PR #1333, opened 2016-01-04
[WFCORE-1261]: Changing the max-history attribute value for configuration-changes fails.

ehsavoie:
Moving the update of the model to the MODEL phase instead of the RUNTIME one.

Jira: https://issues.jboss.org/browse/WFCORE-1261


bstansberry:
:+1: 

Can you clone this to a JBEAP and backport the fix to the 2.x branch? Thanks.

PR #1334, opened 2016-01-04
WFCORE-1259 embedded HC running as slave fix

luck3y:


host-controller/src/main/java/org/jboss/as/domain/controller/operations/SyncModelOperationHandlerWrapper.java
```diff
@@ -127,6 +128,12 @@ public void execute(OperationContext context, ModelNode operation) throws Operat
 
         // Validate references on the new model
         final Resource domain = context.readResourceFromRoot(PathAddress.EMPTY_ADDRESS);
+
+        // for HCs running in admin-only=true, we don't try to update master
+        if (context.getRunningMode() == RunningMode.ADMIN_ONLY) {
+            return;
+        }
```
bstansberry: If the HC is in admin-only via domain.sh, it actually will connect with the master if there is one running:

https://github.com/wildfly/wildfly-core/blob/master/host-controller/src/main/java/org/jboss/as/host/controller/DomainModelControllerService.java#L629

So we shouldn't always disable this syncing; just don't fail below in admin-only if the MasterDomainControllerClient isn't available.

luck3y: Thanks, updated in new commit: https://github.com/luck3y/wildfly-core/commit/275a91bcef68fa685867b414422c7de4b35d2ab7

bstansberry:
:+1:

luck3y:
retest this please

PR #1339, opened 2016-01-05
[WFLY-5928] Avoid NPE

bstansberry:
GlobalTransformerRegistry.registerTransformer will pass a 'null' into the getOrCreate method that creates a OperationTransformerRegistry, so having this var be 'null' is possible via a normal code path. Why this doesn't regularly fail is a mystery though. Perhaps a different OperationTransformerRegistry instance with the same address/version is created via a different code path in some cases, and there is some timing effect involved. The transformer registration is all single threaded AFAIK though.

### DIFF
--- a/controller/src/main/java/org/jboss/as/controller/registry/OperationTransformerRegistry.java
+++ b/controller/src/main/java/org/jboss/as/controller/registry/OperationTransformerRegistry.java
@@ -80,7 +80,7 @@ public class OperationTransformerRegistry {
 
             @Override
             public ResourceTransformer getResourceTransformer() {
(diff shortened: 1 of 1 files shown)


bstansberry:
retest this please

bstansberry:
I've removed the Hold label and will let @kabir decide if this is an appropriate change.

kabir:
Added 'hold', I don't think this is a proper fix as per my comment on https://issues.jboss.org/browse/WFCORE-1270

bstansberry:
I'm closing this. @kabir you know what was in here in case you want to resurrect it. :)

PR #1341, opened 2016-01-06
WFCORE-1271  Description of worker attributes of IO subsystem contain…

ctomc:
…s escaped char sequences


ctomc:
retest this please

PR #1343, opened 2016-01-06
[WFCORE-1272] Clean up the deployment resource on deployment unit service stop

bstansberry:
Bit of a convenience in case DUPs don't undo in undeploy what the do in deploy and then the deployment unit service gets bounced due to some MSC dependency bouncing, e.g. WFLY-4908.

@scottmarlow not critical but I'm curious if this fixes PR the WFLY-4908 issue without the need for your patch. Your patch is still correct though as a DUP should undo in undeploy what it does in deploy; this PR is just a 2nd line of defense.

### DIFF
--- a/server/src/main/java/org/jboss/as/server/deployment/AbstractDeploymentUnitService.java
+++ b/server/src/main/java/org/jboss/as/server/deployment/AbstractDeploymentUnitService.java
@@ -25,7 +25,12 @@ package org.jboss.as.server.deployment;
 import java.util.HashSet;
 import java.util.Set;
 
+import org.jboss.as.controller.capability.CapabilityServiceSupport;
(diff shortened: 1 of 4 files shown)


bstansberry:
Reviewers: the 2nd commit in this PR isn't really needed to fix the problem; it's a refactor to make the code more comfortable for my preferences; e.g. I prefer to avoid overriding a superclass' stop method. It breaks the AbstractDeploymentUnitService's API though by changing a protected constructor. No one outside this package should be subclassing this class though.

But if people object to the 2nd commit I can remove it and still fix the basic problem.

bstansberry:
retest this please

jamezp:
retest this please

PR #1346, opened 2016-01-07
[WFCORE-1274] CLI resolve-parameter-values set to true does not recognize vaulted strings

gaol:
Jira: https://issues.jboss.org/browse/WFCORE-1274

The format of security vault: `${VAULT::xxx::yyy::zzz}` is like system property: `${propName:defaultValue}`. I prefer to let user send escaped string: `$\{VAULT::xxx::yyy::zzz\}`, instead of let CLI distinguish the input for this special case.

### DIFF
--- a/cli/src/main/java/org/jboss/as/cli/util/CLIExpressionResolver.java
+++ b/cli/src/main/java/org/jboss/as/cli/util/CLIExpressionResolver.java
@@ -105,6 +105,9 @@ public class CLIExpressionResolver {
                 }
             } else if(c == '{') {
                 if(state == DOLLAR) {
+                    if (i - 2 >= 0 && input.charAt(i - 2) == '\\') {
+                        return input;
+                    }
                     state = INITIAL;
                     final String[] inputRef = new String[]{input};
                     if(i - 2 == resolveProperty2(inputRef, i - 2, exceptionIfNotResolved)) {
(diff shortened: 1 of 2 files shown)


cli/src/main/java/org/jboss/as/cli/util/CLIExpressionResolver.java
```diff
@@ -48,6 +49,9 @@
     /** Environment variable prefix */
     private static final String ENV_PREFIX = "env.";
 
+    /** Security Vault pattern */
+    private static final Pattern VAULT_PATTERN = Pattern.compile("VAULT::.*::.*::.*");
```
gaol: This pattern comes from: https://github.com/wildfly/wildfly-core/blob/master/server/src/main/java/org/jboss/as/server/services/security/RuntimeVaultReader.java#L53

bstansberry:
@aloubyansky please approve

aloubyansky:
Not ok. First of all this change makes a decision to return too early. If the line in the added test was e.g. "the value is $\{VAULT::text::password::1\} ${os.name}" then ${os.name} wouldn't be resolved.

Second, there is actually a way to escape expressions like that by adding the extra '$' in front. (I copied this from some other expression parser David wrote for DMR or something). That's not very obvious but I remember it was a better idea than messing with backslash escaping in a case like this one.

gaol:
@aloubyansky <pre>there is actually a way to escape expressions like that by adding the extra '$' in front"</pre> Since there is an existing way to escape the expression, this should `NOT` be a problem then, I tested that user can use `$${VAULT::text::password::1}` to escape the expression, which gets the expected: `${VAULT::text::password::1}` in standalone.xml

gaol:
Close it as it should not be a problem.

PR #1351, opened 2016-01-08
[WFCORE-1273]: Can not add property after change of class in custom log formatter.

ehsavoie:
Checking the effective used class before updating the runtime configuration from the model.

Jira: https://issues.jboss.org/browse/WFCORE-1273


ehsavoie:
retest this please

ehsavoie:
retest this please

jamezp:
This won't quite work. The {{logging.properties}} will be persisted with the old values. So during the add this would need to be checked and the formatter replaced. It's convoluted, but here's how it's handled for custom-handler, https://github.com/jamezp/wildfly-core/blob/d0d40addd907f0088a26366049ce1546a5d189ba/logging/src/main/java/org/jboss/as/logging/HandlerOperations.java#L203-L259. 

This also needs to be done for both the {{class}} and {{module}} attributes. I can take this on if you don't have the time.

PR #1352, opened 2016-01-10
[WFCORE-1285] Handle LDAP role DN's that contain special characters better

dehort:


bstansberry:
This looks fine but can you add a test case?

PR #1355, opened 2016-01-12
WFCORE-1314 Propertly handle java 9 version string

ctomc:
Fix the code to properly address new JDK 9 versioning as defined in http://openjdk.java.net/jeps/223


ctomc:
@jamezp can you take a look at this, as it does touch your launcher code.

host-controller/src/test/java/org/jboss/as/host/controller/model/jvm/JvmOptionsBuilderUnitTestCase.java
```diff
@@ -44,13 +44,17 @@
...
             String vmVersionStr = WildFlySecurityManager.getPropertyPrivileged("java.specification.version", null);
             Matcher matcher = Pattern.compile("^1\\.(\\d+)$").matcher(vmVersionStr); //match 1.<number>
-            if (matcher.find()) {
-                vmVersion = Integer.valueOf(matcher.group(1));
+            if (vmVersionStr.contains(".")) {
+                if (matcher.find()) {
+                    vmVersion = Integer.valueOf(matcher.group(1));
+                } else {
+                    throw new RuntimeException("Unknown version of jvm " + vmVersionStr);
+                }
```
dmlloyd: You have a regex here already, why do this with half regex and half string banging?

Just use a regex of "^(?:1\\.)?(\\d+)$" instead.

launcher/src/main/java/org/wildfly/core/launcher/JavaVersionComparator.java
```diff
@@ -59,40 +57,20 @@ public static int compareVersion(final String version1, final String version2) {
...
-                        result.add(Integer.valueOf(s));
-                    } catch (NumberFormatException e) {
-                        throw new IllegalArgumentException(String.format("Version part %s is not a valid integer", s), e);
-                    }
-                }
-            } else {
-                result.add(0);
-            }
+    private int getMajorVersion(String versionString) {
+        final String[] vs = versionString.split("\\.");
```
dmlloyd: Now we're doing two splits here, to acquire literally 2 to 3 characters.  A regex is probably best here too.

ctomc: I just kept it as it was in previous version of code. 
Makes sense to simplify it as much as possible.

ctomc:
@dmlloyd updated.

dmlloyd:
Thanks.

host-controller/src/test/java/org/jboss/as/host/controller/model/jvm/JvmOptionsBuilderUnitTestCase.java
```diff
@@ -43,14 +43,14 @@
...
                 vmVersion = Integer.valueOf(matcher.group(1));
             } else {
-                throw new RuntimeException("Unknown version of jvm "+vmVersionStr);
+                throw new RuntimeException("Unknown version of jvm " + vmVersionStr);
             }
         } catch (Exception e) {
-            vmVersion = 7;
+            vmVersion = 8;
         }
         JVM_MAJOR_VERSION = vmVersion;
```
jamezp: This should just all be removed since we require Java 8 anyway.

jamezp:
The launcher change isn't really needed. The original version works as is without the changes. That said if we're going to change I'd rather just have it removed. We require Java 8 now and if a user wants to use a new version of the launcher to run on older container they can add the JVM arguments manually.

jamezp:
FWIW this also relates to https://issues.jboss.org/browse/WFCORE-1189

host-controller/src/test/java/org/jboss/as/host/controller/model/jvm/JvmOptionsBuilderUnitTestCase.java
ctomc: makes sense

ctomc:
@jamezp I've changed the JvmOptionsBuilderUnitTestCase to not have version check anymore.
Also removed JavaVersionComparator from launcher as you suggested.

ctomc:
@jamezp this got acked for merging to downstream, so we can remove 3.x tag

PR #1357, opened 2016-01-12
WFCORE-1292 SocketBindingManager.createMulticast/DatagramSocket(...) variants needed for unbound sockets

pferraro:
https://issues.jboss.org/browse/WFCORE-1292

See jira for details.


dmlloyd:
Another option could be to have it accept a ```Consumer<XxxSocket>``` which can be used to configure the socket before binding it.

pferraro:
@dmlloyd That would be interesting - but that won't help me with my current issue, since I'm bound by an existing SPI provided by JGroups.

pferraro:
Unrelated failures. This PR only adds methods, none of which are yet utilized - and thus could not possibly cause regressions.

PR #1358, opened 2016-01-13
https://issues.jboss.org/browse/WFCORE-1296 Rejecting the SSL certifi…

aloubyansky:
…cate while connecting via CLI block indefinately

Thanks,
Alexey


n1hility:
@aloubyansky can you submit an EAP core PR?

Thanks!

PR #1360, opened 2016-01-14
WFLY-5994 - all mgmt remote endpoint services should use same xnio worker

ctomc:
also WFCORE-1299 IO Subsystem improperly shuts down the worker


server/src/main/java/org/jboss/as/server/mgmt/ManagementWorkerService.java
```diff
@@ -0,0 +1,85 @@
...
+
+/**
+ * @author Tomaz Cerar (c) 2016 Red Hat Inc.
+ */
+public class ManagementWorkerService implements Service<XnioWorker> {
+    public static final ServiceName SERVICE_NAME = ServiceName.JBOSS.append("serverManagement", "controller", "management", "worker");
+
+
+    private final OptionMap options;
+    private volatile XnioWorker worker;
```
dmlloyd: This does not need to be volatile.  The start/stop create the publishing edges necessary to ensure that this field is visible to dependent services.

server/src/main/java/org/jboss/as/server/mgmt/ManagementWorkerService.java
```diff
@@ -0,0 +1,85 @@
...
+        try {
+            worker = xnio.createWorker(options);
+        } catch (IOException e) {
+            throw new StartException("Could not create worker!", e);
+        }
+    }
+
+    @Override
+    public void stop(StopContext stopContext) {
+        worker.shutdown();
```
dmlloyd: This is wrong, instead you should do this:

```
    private final AtomicReference<StopContext> stopRef = new AtomicReference<>();
    // ...
        public void start(final StartContext context) throws StartException {
        final Xnio xnio = Xnio.getInstance();
        try {
            xnioWorker = xnio.createWorker(null, options, () -> stopDone());
        } catch (IOException e) {
            throw new StartException(e);
        }
    }

    public void stop(final StopContext context) {
        stopRef.set(context);
        context.asynchronous();
        xnioWorker.shutdown();
        xnioWorker = null;
    }

    private void stopDone() {
        final StopContext stopContext = stopRef.get();
        assert stopContext != null;
        stopContext.complete();
    }
    // ...
```

Otherwise the service will stop before the worker is stopped.

dmlloyd: Actually you could just use a volatile field for the StopContext, it doesn't need to be AtomicReference.

PR #1361, opened 2016-01-14
[WFCORE-1297] deploy --url command creates invalid deployment

dpospisil:
JIRA: https://issues.jboss.org/browse/WFCORE-1297


aloubyansky:
This looks good to me. As to your question (from the email) about whether to leave only the last part of the URL, i.e. the file name, let's also @stuartwdouglas for an opinion.

dpospisil:
I think that the argument Martin commented in the JIRA makes sense. We use only last part of the file name so we should do the same with URLs. If @aloubyansky @stuartwdouglas are fine with that I update the PR accordingly.

aloubyansky:
I'm fine with that. Although, that'll be a backward incompatible change.

dpospisil:
OK, I have change it to use it only last part which I believe is better and you are saying you are fine with. Incompatibility (changing the deployment name) seems to be required anyway.

PR #1362, opened 2016-01-15
WFCORE-1302 Rename wildfly-launcher.jar to launcher.jar

ctomc:


jamezp:
Just to throw in my $0.02. I don't see the point in this. It's a rename for the sake of a rename.

PR #1363, opened 2016-01-15
[WFCORE-1135] Enable custom SSL configuration for the connection from the app server to it's host controller.

darranl:
Also includes [WFCORE-1283] to add a new version of the management model / schema.


darranl:
The WildFly side of this PR can be found here https://github.com/wildfly/wildfly/pull/8578

darranl:
So far test failures look to be expected until both PRs are merged in parallel.

darranl:
Before this is merged I have one more commit to add.

darranl:
Retest this please

darranl:
This last failure looks like TeamCity is using an stale copy of the branch - the line calling assertEquals is now line 184 in the test and not 170.

bstansberry:
Added 3.x per yesterday's discussion.

PR #1364, opened 2016-01-18
[WFCORE-1281] NPE in LDAP principal to group loading

darranl:
Check that the simple name attribute can be loaded for the group and if it is required, not loaded and missing group skipping disabled - log an error.


bstansberry:
Please ping Jason and I if you want to lobby for this in WF 10.

PR #1365, opened 2016-01-18
[WFCORE-1304] fix StringIndexOutOfBoundsException when clean an obsolete system added content

soul2zimate:
https://issues.jboss.org/browse/WFCORE-1304

Mac OS can add a hidden file .DS_Store to deployment content repository. This causes an unexpected StringIndexOutOfBoundsException when it calls HashUtil.hexStringToByteArray() to clean obsolete content.

It should exam each char inside hexString is among 0 to f for such system added file, in this case ".DS.Store".

As reported, It's also reproducible in Linux by creating a file under the content dir, but I think it's incorrect behavior to manually create any customer file in this directory.


bstansberry:
I put the 3.x label on this because the BZ isn't acked yet and we're not merging unacked stuff for WildFly 10, which should be done very very soon. Once this master branch opens up, this looks good. Thanks!

PR #1369, opened 2016-01-19
[WFCORE-1246]: System allows enabling two applications with same runtime-name

ehsavoie:
Checking the unicity of the runtime name when deploying a deployment.

Jira: https://issues.jboss.org/browse/WFCORE-1246


ehsavoie:
retest this please

PR #1373, opened 2016-01-20
modified script to allow special characters in username and Password

mdorenkamp:
The original service.bat stripped out special characters like '!' in passwords. This was due to the delayed expansion of variables which is only necessary to determine JBOSS_HOME.


ctomc:
:+1:

PR #1377, opened 2016-01-21
WFCORE-1319 - Defaults for Metasize and Java 8

luck3y:
Initial change for standalone, possibly another change to follow for jvm defaults (currently 256m).

### DIFF
--- a/core-feature-pack/src/main/resources/configuration/host/host-master.xml
+++ b/core-feature-pack/src/main/resources/configuration/host/host-master.xml
@@ -74,7 +74,7 @@
             <heap size="64m" max-size="256m"/>
             <jvm-options>
                 <option value="-server"/>
-                <option value="-XX:MetaspaceSize=256m"/>
+                <option value="-XX:MetaspaceSize=96m"/>
                 <option value="-XX:MaxMetaspaceSize=256m"/>
             </jvm-options>
         </jvm>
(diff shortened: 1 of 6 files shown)


jamezp:
Do these need to be added to domain scripts too? They probably should be added to the launcher, https://github.com/wildfly/wildfly-core/blob/master/launcher/src/main/java/org/wildfly/core/launcher/AbstractCommandBuilder.java#L46

jamezp:
Ha, ignore that first question :) It's all in the timing.

luck3y:
Thanks James, I'll take a look at the launcher.

luck3y:
@jamezp Since we're currently just setting MaxPermSize if it's supported, I think it may be fine to just leave MetaspaceSize out, and have the user set it as a runtime property if they need it specifically for performance, since not specifying a max or a min will allow it to settle on the needed amount dynamically. What do you think?

luck3y:
retest this please

bstansberry:
As we said in chat, any launcher change can come separately, as I want to get the script stuff into .Final

jamezp:
@luck3y It's no real big deal to me. I just like to keep it consistent with whatever the scripts do is all. It's what wildfly-arquillian and the wildfly-maven-plugin use to launch WildFly.

PR #1379, opened 2016-01-22
WFCORE-1321 io.undertow.core module should be private

ctomc:
jira: https://issues.jboss.org/browse/WFCORE-1321


ctomc:
this should remain public

PR #1382, opened 2016-01-25
[JBEAP-2830] Disable console logging for standalone-full-ha.xml

treblereel:
JIRA: https://issues.jboss.org/browse/JBEAP-2830
EAP7 PR: https://github.com/jbossas/jboss-eap7/pull/244


bstansberry:
this is ok to test

bstansberry:
This is EAP-specific behavior, so we don't want it in upstream. Thanks.

PR #1385, opened 2016-01-25
[WFCORE-1066]: Parameter -Djboss.domain.log.dir does not have effect on HP-UX when directory does not exist.

ehsavoie:
Checking if the directory exists and avoiding verifying it if not.

Jira: https://issues.jboss.org/browse/WFCORE-1066

### DIFF
--- a/core-feature-pack/src/main/resources/content/bin/domain.sh
+++ b/core-feature-pack/src/main/resources/content/bin/domain.sh
@@ -193,7 +193,12 @@ if $darwin || $other ; then
              JBOSS_BASE_DIR=`cd ${p#*=} ; pwd -P`
              ;;
         -Djboss.domain.log.dir=*)
-             JBOSS_LOG_DIR=`cd ${p#*=} ; pwd -P`
+             if [ -d "${p#*=}" ]; then
+                JBOSS_LOG_DIR=`cd ${p#*=} ; pwd -P`
+             else
+                #since the specified directory doesn't exist we don't validate it
+                JBOSS_LOG_DIR=${p#*=}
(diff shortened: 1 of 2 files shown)


bstansberry:
Is equivalent logic needed in the Windows scripts?

Obviously not to correct an HP-UX issue, but just in the general sense of maintaining consistency across scripts.

jamezp:
Windows does not appear to validate the directory in the `.bat` scripts and I don't think it does in the `.ps1` scripts either. I think this should be okay. We also don't validate the directory exists for Solaris too. This does mean the user must use `${HOME}` rather than `~`. Validation of the directory will happen when an attempt is made to create the file/directory in the log manager.

bstansberry:
Thanks @jamezp. My question wasn't meant to advocate for adding significant new logic to the Windows scripts, and it sounds like that would be needed. I was more concerned with just not letting small differences creep in, and it sounds like that's not the case here.

ctomc:
we can easily add validation to ps1 scripts but for .bat it would be quite a hassle :(
lets have this in mind for WF11

jamezp:
@ctomc I don't think we really want the validation. The scripts should do as little as possible IMO. The issue this PR solves is in HP-UX the `cd ${p#*=} ; pwd -P` will put the wrong string into the environment variable, e.g. 

> bash cd: /some/dir: No such file or directory

So this just changes it to parse the possible directory which will be validated later, in this case by the logmanager.

ehsavoie:
Does it fail on windows if the directory doesn't exist ? This is an issue on HP that doesn't happen on other unices

PR #1389, opened 2016-01-26
[WFCORE-1308]: Description of suspend and shutdown ops does not include a 'unit'

ehsavoie:
Adding a unit in the operation description and removing the unit from the description text to avoid duplication.

Jira: https://issues.jboss.org/browse/WFCORE-1308


bstansberry:
retest this please

PR #1392, opened 2016-01-27
[WFCORE-1333] validate path and relative-to attributes value.

soul2zimate:
https://issues.jboss.org/browse/WFCORE-1333
validate attribute path and relative-to path, checking directory exists and it is an writable directory.

### DIFF
--- a/deployment-scanner/src/main/java/org/jboss/as/server/deployment/scanner/WritePathAttributeHandler.java
+++ b/deployment-scanner/src/main/java/org/jboss/as/server/deployment/scanner/WritePathAttributeHandler.java
@@ -1,0 +1,88 @@
+/*
+ * JBoss, Home of Professional Open Source.
+ * Copyright 2016, Red Hat, Inc., and individual contributors
+ * as indicated by the @author tags. See the copyright.txt file in the
+ * distribution for a full listing of individual contributors.
+ *
+ * This is free software; you can redistribute it and/or modify it
+ * under the terms of the GNU Lesser General Public License as
+ * published by the Free Software Foundation; either version 2.1 of
+ * the License, or (at your option) any later version.
+ *
+ * This software is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the GNU
+ * Lesser General Public License for more details.
+ *
+ * You should have received a copy of the GNU Lesser General Public
+ * License along with this software; if not, write to the Free
+ * Software Foundation, Inc., 51 Franklin St, Fifth Floor, Boston, MA
+ * 02110-1301 USA, or see the FSF site: http://www.fsf.org.
+ */
+
+package org.jboss.as.server.deployment.scanner;
+
+import static org.jboss.as.server.deployment.scanner.logging.DeploymentScannerLogger.ROOT_LOGGER;
+
+import java.nio.file.Files;
+import java.nio.file.Path;
+import java.nio.file.Paths;
+
+import org.jboss.as.controller.OperationContext;
+import org.jboss.as.controller.OperationFailedException;
+import org.jboss.as.controller.OperationStepHandler;
+import org.jboss.as.controller.PathAddress;
+import org.jboss.as.controller.ReloadRequiredWriteAttributeHandler;
+import org.jboss.as.controller.OperationContext.Stage;
+import org.jboss.as.controller.services.path.PathManager;
+import org.jboss.dmr.ModelNode;
+
+/**
+ * @author wangc
+ *
+ */
+public class WritePathAttributeHandler extends ReloadRequiredWriteAttributeHandler {
+
+    private final PathManager pathManager;
+
+    public WritePathAttributeHandler(PathManager pathManager) {
+        super(DeploymentScannerDefinition.PATH);
+        this.pathManager = pathManager;
+    }
+
+    @Override
+    protected boolean applyUpdateToRuntime(OperationContext context, ModelNode operation, String attributeName,
+            ModelNode resolvedValue, ModelNode currentValue, HandbackHolder<Void> voidHandback)
+                    throws OperationFailedException {
+        boolean reloadRequired = super.applyUpdateToRuntime(context, operation, attributeName, resolvedValue, currentValue,
+                voidHandback);
+        if (reloadRequired) {
+            // The value actually changed so we need to add step to validate the path
+            context.addStep(new OperationStepHandler() {
+                @Override
+                public void execute(OperationContext context, ModelNode operation) throws OperationFailedException {
+                    final ModelNode model = context.readResource(PathAddress.EMPTY_ADDRESS).getModel();
+                    final ModelNode relativeTo = DeploymentScannerDefinition.RELATIVE_TO.resolveModelAttribute(context, model);
+                    Path fullPath;
+                    if (relativeTo.isDefined()) {
+                        final String fullPathName = pathManager.resolveRelativePathEntry(resolvedValue.asString(),
+                                relativeTo.asString());
+                        fullPath = Paths.get(fullPathName);
+                    } else {
+                        fullPath = Paths.get(resolvedValue.asString());
+                    }
+
+                    if (!Files.exists(fullPath)) {
+                        ROOT_LOGGER.directoryIsNonexistent(fullPath.toString());
+                    } else if (!Files.isDirectory(fullPath)) {
+                        ROOT_LOGGER.isNotADirectory(fullPath.toString());
+                    } else if (!Files.isReadable(fullPath)) {
+                        ROOT_LOGGER.directoryIsNotWritable(fullPath.toString());
+
+                    }
+                }
+            }, Stage.RUNTIME);
+        }
+        return reloadRequired;
+    }
+}
--- a/deployment-scanner/src/main/java/org/jboss/as/server/deployment/scanner/WriteRelativeToAttributeHandler.java
+++ b/deployment-scanner/src/main/java/org/jboss/as/server/deployment/scanner/WriteRelativeToAttributeHandler.java
@@ -1,0 +1,60 @@
+/*
+ * JBoss, Home of Professional Open Source.
+ * Copyright 2016, Red Hat, Inc., and individual contributors
+ * as indicated by the @author tags. See the copyright.txt file in the
+ * distribution for a full listing of individual contributors.
+ *
+ * This is free software; you can redistribute it and/or modify it
+ * under the terms of the GNU Lesser General Public License as
+ * published by the Free Software Foundation; either version 2.1 of
+ * the License, or (at your option) any later version.
+ *
+ * This software is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the GNU
+ * Lesser General Public License for more details.
+ *
+ * You should have received a copy of the GNU Lesser General Public
+ * License along with this software; if not, write to the Free
+ * Software Foundation, Inc., 51 Franklin St, Fifth Floor, Boston, MA
+ * 02110-1301 USA, or see the FSF site: http://www.fsf.org.
+ */
+
+package org.jboss.as.server.deployment.scanner;
+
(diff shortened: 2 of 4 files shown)


deployment-scanner/src/main/java/org/jboss/as/server/deployment/scanner/logging/DeploymentScannerLogger.java
```diff
@@ -407,4 +407,16 @@
...
+    String directoryIsNonexistent(String path);
+
+    @Message(id = 38, value = "%s is not a directory")
+    String isNotADirectory(String path);
+
+    @Message(id = 39, value = "%s is not writable")
+    String directoryIsNotWritable(String path);
+
+    @Message(id = 40, value = "Could not find a path called '%s'")
+    String pathEntryNotFound(String pathName);
```
bstansberry: Messages like these that are only used to create an exception should all the exception type (i.e. OperationFailedException) as the return type for the method, not String. Then in the places that use the methods, just

throw DeploymentScannerLogger.ROOT_LOGGER.pathEntryNotFound(pathName);

deployment-scanner/src/main/java/org/jboss/as/server/deployment/scanner/WriteRelativeToAttributeHandler.java
```diff
@@ -0,0 +1,87 @@
...
+                    }
+                    if (!relativePath.isDirectory()) {
+                        throw new OperationFailedException(
+                                DeploymentScannerLogger.ROOT_LOGGER.isNotADirectory(relativePath.getAbsolutePath()));
+                    }
+                    if (!relativePath.canWrite()) {
+                        throw new OperationFailedException(
+                                DeploymentScannerLogger.ROOT_LOGGER.directoryIsNotWritable(relativePath.getAbsolutePath()));
+                    }
+                }
```
bstansberry: This block starting at L69 doesn't belong in this class. It's not the task of the deployment-scanner subsystem to validate these, beyond ensuring the attribute value is a reference is to a known path. Validation of the full path in WritePathAttributeHandler covers the deployment-scanner subsystem's needs.

deployment-scanner/src/main/java/org/jboss/as/server/deployment/scanner/WritePathAttributeHandler.java
```diff
@@ -0,0 +1,83 @@
...
+            if (!relativePath.isDirectory()) {
+                throw new OperationFailedException(
+                        DeploymentScannerLogger.ROOT_LOGGER.isNotADirectory(relativePath.getAbsolutePath()));
+            }
+            if (!relativePath.canWrite()) {
+                throw new OperationFailedException(
+                        DeploymentScannerLogger.ROOT_LOGGER.directoryIsNotWritable(relativePath.getAbsolutePath()));
+            }
+        }
+    }
```
bstansberry: This would need to be done via adding a step. This ensures that if this write-attribute is part of a composite operation (CLI batch or console update) then the validation does not occur until the other steps in the composite have had a chance to execute, e.g. a step that changes the relative-to attribute as well.

bstansberry:
Thanks for this.

Beyond the comments I already made I need to check whether the timing of doing this validation in validateUpdatedModel is correct, as doing it there means the validation is performing a runtime check in Stage.MODEL.

deployment-scanner/src/main/java/org/jboss/as/server/deployment/scanner/WritePathAttributeHandler.java
```diff
@@ -0,0 +1,83 @@
...
+
+    @Override
+    protected void validateUpdatedModel(final OperationContext context, final Resource resource)
+            throws OperationFailedException {
+
+        final ModelNode model = resource.getModel();
+        final ModelNode path = model.get(CommonAttributes.PATH);
+
+        final ServiceController<PathManager> controller = (ServiceController<PathManager>) context.getServiceRegistry(false)
```
bstansberry: DeploymentScannerDefinition is given a ref to the PathManager in its constructor, so it should pass it into this class via its constructor instead of using a service lookup.

deployment-scanner/src/main/java/org/jboss/as/server/deployment/scanner/WritePathAttributeHandler.java
```diff
@@ -0,0 +1,83 @@
...
+                .getService(PathManagerService.SERVICE_NAME);
```
bstansberry: I've checked, and it's more robust to do this in Stage.Runtime. So, more like
```
@Override
protected abstract boolean applyUpdateToRuntime(OperationContext context, ModelNode operation, String attributeName,
                                                    ModelNode resolvedValue, ModelNode currentValue, HandbackHolder<T> handbackHolder)
            throws OperationFailedException {

    boolean reloadRequired = super.applyUpdateToRuntime(....);
    if (reloadRequired) {
        // The value actually changed so we need to add step to validate the path
        content.addStep(new OperationStepHandler() {
           

    @Override
        .... the validation logic
    }
       }, Stage.RUNTIME);
    }
    return reloadRequired

}
```

deployment-scanner/src/main/java/org/jboss/as/server/deployment/scanner/logging/DeploymentScannerLogger.java
```diff
@@ -407,4 +407,16 @@
      */
     @Message(id = 36, value = "Deployment model operation failed. %s")
     RuntimeException deployModelOperationFailed(String msg);
+
+    @Message(id = 37, value = "%s does not exist")
+    String directoryIsNonexistent(String path);
```
bstansberry: A message for this already exists in this class.

deployment-scanner/src/main/java/org/jboss/as/server/deployment/scanner/logging/DeploymentScannerLogger.java
```diff
@@ -407,4 +407,16 @@
...
+
+    @Message(id = 38, value = "%s is not a directory")
+    String isNotADirectory(String path);
```
bstansberry: A message for this already exists in this class.

deployment-scanner/src/main/java/org/jboss/as/server/deployment/scanner/logging/DeploymentScannerLogger.java
```diff
@@ -407,4 +407,16 @@
...
+
+    @Message(id = 39, value = "%s is not writable")
+    String directoryIsNotWritable(String path);
```
bstansberry: A message for this already exists in this class.

deployment-scanner/src/main/java/org/jboss/as/server/deployment/scanner/WritePathAttributeHandler.java
```diff
@@ -0,0 +1,83 @@
...
+        if (controller != null) {
+            File relativePath;
+            if (model.hasDefined(CommonAttributes.RELATIVE_TO)) {
```
bstansberry: Use DeploymentScannerDefinition.RELATIVE_TO.resolveModelAttribute(context, model)

deployment-scanner/src/main/java/org/jboss/as/server/deployment/scanner/WritePathAttributeHandler.java
```diff
@@ -0,0 +1,83 @@
...
+    protected void validateUpdatedModel(final OperationContext context, final Resource resource)
+            throws OperationFailedException {
+
+        final ModelNode model = resource.getModel();
+        final ModelNode path = model.get(CommonAttributes.PATH);
+
+        final ServiceController<PathManager> controller = (ServiceController<PathManager>) context.getServiceRegistry(false)
+                .getService(PathManagerService.SERVICE_NAME);
+        if (controller != null) {
+            File relativePath;
```
bstansberry: Can you call this var something like 'fullPath"? The name relativePath is confusing because it's not obvious whether it's the partial path or the whole thing.

dmlloyd:
Please see my comment at https://issues.jboss.org/browse/WFCORE-1333?focusedCommentId=13154721&page=com.atlassian.jira.plugin.system.issuetabpanels:comment-tabpanel#comment-13154721 before considering this change.

soul2zimate:
Thanks, I have rebased this. WritePathAttributeHandler now shows WARN messages and WriteRelativeToAttributeHandler throws OperationFailedException for different invalid values.

deployment-scanner/src/main/java/org/jboss/as/server/deployment/scanner/WritePathAttributeHandler.java
```diff
@@ -0,0 +1,85 @@
...
+        super(DeploymentScannerDefinition.PATH);
+        this.pathManager = pathManager;
+    }
+
+    @Override
+    protected boolean applyUpdateToRuntime(OperationContext context, ModelNode operation, String attributeName,
+            ModelNode resolvedValue, ModelNode currentValue, HandbackHolder<Void> voidHandback)
+                    throws OperationFailedException {
+        boolean reloadRequired = super.applyUpdateToRuntime(context, operation, attributeName, resolvedValue, currentValue,
+                voidHandback);
```
jamezp: This doesn't seem needed an the return value should just be `true`.

jamezp: Never mind, I see what's happening now.

deployment-scanner/src/main/java/org/jboss/as/server/deployment/scanner/WritePathAttributeHandler.java
```diff
@@ -0,0 +1,85 @@
...
+        if (reloadRequired) {
+            final ModelNode model = context.readResource(PathAddress.EMPTY_ADDRESS).getModel();
+            final ModelNode relativeTo = DeploymentScannerDefinition.RELATIVE_TO.resolveModelAttribute(context, model);
+            // The value actually changed so we need to add step to validate the path
+            context.addStep(new OperationStepHandler() {
+                @Override
+                public void execute(OperationContext context, ModelNode operation) throws OperationFailedException {
+                    File fullPath;
```
jamezp: In this specific case I think we should be using `java.nio.file.Path`. The `fullPath.canWrite()` is not reliable per the JavaDoc while `java.nio.file.Files.isReadable()` tends to be more reliable.

> Tests whether the application can read the file denoted by this abstract pathname. On some platforms it may be possible to start the Java virtual machine with special privileges that allow it to read files that are marked as unreadable. Consequently this method may return true even though the file does not have read permissions.

deployment-scanner/src/main/java/org/jboss/as/server/deployment/scanner/WritePathAttributeHandler.java
```diff
@@ -0,0 +1,85 @@
...
+
+    @Override
+    protected boolean applyUpdateToRuntime(OperationContext context, ModelNode operation, String attributeName,
+            ModelNode resolvedValue, ModelNode currentValue, HandbackHolder<Void> voidHandback)
+                    throws OperationFailedException {
+        boolean reloadRequired = super.applyUpdateToRuntime(context, operation, attributeName, resolvedValue, currentValue,
+                voidHandback);
+        if (reloadRequired) {
+            final ModelNode model = context.readResource(PathAddress.EMPTY_ADDRESS).getModel();
+            final ModelNode relativeTo = DeploymentScannerDefinition.RELATIVE_TO.resolveModelAttribute(context, model);
```
jamezp: These need to be moved down into the new step.

deployment-scanner/src/main/java/org/jboss/as/server/deployment/scanner/logging/DeploymentScannerLogger.java
```diff
@@ -407,4 +408,19 @@
...
+
+    @LogMessage(level = WARN)
+    @Message(id = 38, value = "%s is not a directory")
+    void isNotADirectory(String path);
+
+    @LogMessage(level = WARN)
+    @Message(id = 39, value = "%s is not writable")
+    void directoryIsNotWritable(String path);
+
+    @Message(id = 40, value = "Could not find a path called '%s'")
```
jamezp: This should probably be something more like; `Could not find relative-to path entry for %s`.

soul2zimate:
I have rebased this based on recent comments

PR #1393, opened 2016-01-27
[WFCORE-1306] SlaveReconnectTestCase fails on slow systems

dpospisil:
JIRA: https://issues.jboss.org/browse/WFCORE-1306

### DIFF
--- a/testsuite/domain/src/test/java/org/jboss/as/test/integration/domain/slavereconnect/SlaveReconnectTestCase.java
+++ b/testsuite/domain/src/test/java/org/jboss/as/test/integration/domain/slavereconnect/SlaveReconnectTestCase.java
@@ -40,6 +40,9 @@ import java.util.List;
 
 import org.jboss.as.controller.PathAddress;
 import org.jboss.as.controller.client.helpers.domain.DomainClient;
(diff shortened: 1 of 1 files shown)


kabir:
Retest this please

testsuite/domain/src/test/java/org/jboss/as/test/integration/domain/slavereconnect/SlaveReconnectTestCase.java
```diff
@@ -148,6 +151,14 @@ private void testReconnect(ReconnectTestScenario[] scenarios) throws Exception {
                 slaveReconnected = checkSlaveReconnected(masterClient);
             } while (!slaveReconnected && System.currentTimeMillis() < end);
 
+            //Wait for master servers to come up
+            end = System.currentTimeMillis() + 20 * ADJUSTED_SECOND;
```
bstansberry: 20 sounds too low.

If the test is going to fail, waiting a long time for it to happen is no big deal. But a false failure because of a short timeout is bad.

dpospisil:
Timeout increased to adjusted 60s.

PR #1366, opened 2016-01-19
[WFCORE-872] Add version 5 of the management schema and model.

darranl:
:exclamation: Commits from this branch are used in numerous other topic branches so please do not cherry-pick or rebase. :exclamation:

PR #1399, opened 2016-02-02
[WFCORE-1334] Allow disabling audit log rotation

TomasHofman:
https://issues.jboss.org/browse/WFCORE-1334

### DIFF
--- a/controller/src/main/java/org/jboss/as/controller/audit/FileAuditLogHandler.java
+++ b/controller/src/main/java/org/jboss/as/controller/audit/FileAuditLogHandler.java
@@ -40,14 +40,18 @@ public class FileAuditLogHandler extends AbstractFileAuditLogHandler {
     //Methods on this class will only ever be called from one thread (see class javadoc) so although it looks shared here it is not
     private static final SimpleDateFormat OLD_FILE_FORMATTER = new SimpleDateFormat("yyyy-MM-dd_HHmmss");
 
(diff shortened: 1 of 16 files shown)


TomasHofman:
Investigating the failures.

TomasHofman:
retest this please

controller/src/main/java/org/jboss/as/controller/audit/FileAuditLogHandler.java
```diff
@@ -40,14 +40,18 @@
     //Methods on this class will only ever be called from one thread (see class javadoc) so although it looks shared here it is not
     private static final SimpleDateFormat OLD_FILE_FORMATTER = new SimpleDateFormat("yyyy-MM-dd_HHmmss");
 
-    public FileAuditLogHandler(String name, String formatterName, int maxFailureCount, PathManagerService pathManager, String path, String relativeTo) {
+    private boolean rotateAtStartup;
```
bstansberry: This should be final.

bstansberry:
Other than the minor thing about the field being final, this looks good. But we'll need to wait for https://github.com/wildfly/wildfly-core/pull/1366 to be merged and then tweak this to reflect that this is a change to version 5 of the xsd, not 4.

bstansberry:
#1366 is merged, so this can be updated to use the correct xsd version.

TomasHofman:
I updated the PR to use version 5.

PR #1402, opened 2016-02-03
https://issues.jboss.org/browse/WFCORE-1359 Exiting the CLI after dis…

aloubyansky:
…abling and clearing history will clear content of .jboss-cli-history

The help content describes outdated behavior, probably back from jline days.

Thanks,
Alexey


rhusar:
BTW to keep git history clean we only include the ID in the commit message, not the whole url.

PR #1405, opened 2016-02-03
[WFLY-6119] Fix service.bat execution when JBOSS_HOME contains spaces

ilpianista:
To test, try to execute `service.bat install` when wildfly is under Program Files (x86)


bstansberry:
this is ok to test

PR #1408, opened 2016-02-04
[WFCORE-857]: A read-resource on a subsystem resource of a missing deployment returns a successful completion

ehsavoie:
For a global operation check that if the address doesn't resolve in the middle of it we fail.
Jira: https://issues.jboss.org/browse/WFCORE-857

### DIFF
--- a/controller/src/main/java/org/jboss/as/controller/operations/global/GlobalOperationHandlers.java
+++ b/controller/src/main/java/org/jboss/as/controller/operations/global/GlobalOperationHandlers.java
@@ -778,6 +778,7 @@ public class GlobalOperationHandlers {
             }
         }
 
+        @Override
         protected void executeSingleTargetChild(PathAddress base, PathElement currentElement, PathAddress newRemaining, OperationContext context, boolean ignoreMissing) {
             final PathAddress next = base.append(currentElement);
             // Either require the child or a remote target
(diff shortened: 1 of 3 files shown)


bstansberry:
retest this please

bstansberry:
Failure looks related.

controller/src/main/java/org/jboss/as/controller/operations/global/GlobalOperationHandlers.java
```diff
@@ -772,6 +773,10 @@ protected void executeSingleTargetChild(PathAddress base, PathElement currentEle
             if (resource.hasChild(currentElement) || (nr != null && nr.isRemote())) {
                 safeExecute(next, newRemaining, context, nr, ignoreMissing);
             }
+            //if we are on the wrong host no need to do anything
+            if(!resource.hasChild(currentElement) && !HOST.equals(currentElement.getKey()) && (newRemaining != null && newRemaining.size() > 0)) {
+               throw new Resource.NoSuchResourceException(currentElement);
+            }
```
bstansberry: 1) Shouldn't this be an 'else if'?

2) I don't get the '&& !HOST.equals(currentElement.getKey())' bit. Why not throw an exception? Is this about someone connecting the a slave HC and doing /host=master/...:read-resource? So asking about a real host but one that the slave can't know about? If so I think the exception is fine.

3) Re: '&& (newRemaining != null && newRemaining.size() > 0)' -- looking at the only caller of this method, neither of those conditions will be true, so let's just javadoc that fact in the executeSingleTargetChild javadoc and drop this check.

ehsavoie: I can't reproduce the issue any more. If I remember correctly this was happening during address resolution in multitarget calls

ehsavoie:
retest this please

PR #1409, opened 2016-02-04
[WFCORE-1360] CLI Options missed in WFCORE-1199

ryanemerson:
Now covers --properties option in ASModuleHandler and the --server-groups options in DeployHandler and UndeployHandler.

JIRA: https://issues.jboss.org/browse/WFCORE-1360


bstansberry:
@aloubyansky is this ok?

@aloubyansky @n1hility is this a blocker for EAP 7.0? The change at https://github.com/wildfly/wildfly-core/pull/1409/files#diff-3fcc86ac2bc7cf21e425c408a25bc379R83 implies this is somewhat an API change, as scripts will need to use the bracket syntax. So bringing it in post 7.0.0.GA is a problem.

aloubyansky:
Yes, this is ok. It's a bug fix for non-interactive mode when the commands are specified as the script command line arguments.

PR #1410, opened 2016-02-04
[WFCORE-971]: Autocomplete not working for ha-policy names

ehsavoie:
Adding option to have the allowed names of singletons for read-children-names operation.

Jira: https://issues.jboss.org/browse/WFCORE-971

### DIFF
--- a/controller/src/main/resources/org/jboss/as/controller/descriptions/common/LocalDescriptions.properties
+++ b/controller/src/main/resources/org/jboss/as/controller/descriptions/common/LocalDescriptions.properties
@@ -430,6 +430,7 @@ global.read-attribute-group.resolve-expressions=Resolves expressions to current 
 global.read-attribute-group-names=Gets the names of all the attribute groups under the selected resource
 global.read-children-names=Gets the names of all children under the selected resource with the given type
 global.read-children-names.child-type=The name of the node under which to get the children names
+global.read-children-names.include-singletons=If 'true' include the allowed values for any singleton registration, even if no resource currently exists with that name.
 global.read-children-names.reply=The children names
 global.read-children-types=Gets the type names of all the children under the selected resource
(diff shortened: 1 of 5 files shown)


bstansberry:
AIUI, this is adding the names of non-existent children to the output of read-children-names?

I don't think that's the right solution as it's mixing concepts (MRR vs Resource).

Perhaps we should add another field to the read-resource-description output in addition to the "description" and "model-description" fields discussed at https://docs.jboss.org/author/display/WFLY10/Admin+Guide#AdminGuide-DescriptionofParent%2FChildRelationships

Specifically:

allowed – List – a list of legal values for the names of children of this type. If not specified, valid names are not restricted.

bstansberry:
I've changed my mind about this. I think your approach is ok. It's a bit impure, but getting all the data in one call is a lot simpler for the CLI, and this tab completion is probably the main use of read-children-names.

controller/src/main/resources/org/jboss/as/controller/descriptions/common/LocalDescriptions.properties
```diff
@@ -430,6 +430,7 @@ global.read-attribute-group.resolve-expressions=Resolves expressions to current
 global.read-attribute-group-names=Gets the names of all the attribute groups under the selected resource
 global.read-children-names=Gets the names of all children under the selected resource with the given type
 global.read-children-names.child-type=The name of the node under which to get the children names
+global.read-children-names.include-singletons=If 'true' include the allowed values for any singleton registration.
```
bstansberry: If 'true' include the allowed values for any singleton registrations, even if no resource currently exists with thet name.

bstansberry:
I've added the Fix Me label for the minor LocalDecriptions.properties tweak I mentioned. Otherwise looks good.

PR #1411, opened 2016-02-05
WFCORE-1361 Update the command synopsis for jboss-cli

aloubyansky:
Thanks,
Alexey

### DIFF
--- a/cli/src/main/resources/help/connect.txt
+++ b/cli/src/main/resources/help/connect.txt
@@ -1,6 +1,6 @@
 SYNOPSIS
 
-   connect [host][:port]
+   connect [controller_alias | [protocol://][host][:port]]
 
 DESCRIPTION
 
@@ -34,7 +34,8 @@ DESCRIPTION
   - an IPv4 address, e.g. 127.0.0.1
   - an IPv6 address, e.g. [::1]
 
-
+  Controller addresses can be aliased and stored in the jboss-cli.xml config
+  file. Aliases can later be used as arguments for the connect command.
 
 
 ARGUMENTS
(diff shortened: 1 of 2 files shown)


cli/src/main/resources/help/connect.txt
```diff
@@ -34,7 +34,8 @@ DESCRIPTION
   - an IPv4 address, e.g. 127.0.0.1
   - an IPv6 address, e.g. [::1]
 
-
+  Controller addresses can be aliased and stored in jboss-cli.xml config file.
+  Aliases can later be used as arguments for connect command.
```
bstansberry: "for the connect command"

cli/src/main/resources/help/help.txt
```diff
@@ -14,10 +16,13 @@ Usage:
  --version       - prints the version info of the JBoss AS release, JVM and
                    system environment.
 
- --controller    - the default controller host and port to connect to when
-                   --connect option (described below) is specified or when the
-                   connect command is issued w/o the arguments. The default
-                   controller host is localhost and the port is 9990.
+ --controller    - the default controller (including protocol, host and port)
+                   to connect to when --connect option (described below) is
```
bstansberry: when the --connect option

cli/src/main/resources/help/help.txt
```diff
@@ -77,7 +82,7 @@ Usage:
                    to return. The value provided must be a positive integer.
                    Defaults to 5000 milliseconds when not provided.
 
-  --bind         - specifies to which address the CLI is going to be bound to.
+ --bind          - specifies to which address the CLI is going to be bound to.
```
bstansberry: is going to be bound.

No final "to" is needed as the clause begins with "to which".

cli/src/main/resources/help/connect.txt
bstansberry: Also "store in the jboss-cli.xml config file"

aloubyansky:
Thanks for corrections!

PR #1412, opened 2016-02-05
WFCORE-1362 Add --connection-properties to datasource add command

aloubyansky:
Thanks,
Alexey


aloubyansky:
I am looking into the failure.

bstansberry:
retest this please

bstansberry:
Looks good. I'd like it if the server side datasource add handler could just accept the properties as a param and do this server side, but I don't know when we'll get around to that, so this is good.

Really, I'd like what I describe ^^^ to be something the kernel can do automagically; i.e. look for add op params that look like child resources and convert those into added steps to add the child(ren). Part of doing that would mean the CLI's operation validation logic would need to understand that too and not reject the op because of the "extra" params. Or perhaps the "add" op description generator can automagically add child types as valid params.

Anyway, just FYI about some thoughts.

aloubyansky:
We discussed the issue for the XA properties back in the days. I also argued it had to be done on the server side. If the operation will be able to accept more parameters, it has to be clearly enough reflected in the description, so the operation request validation remains straightforward.

PR #1415, opened 2016-02-11
WFCORE-794 IPv6ScopeIdMatchUnitTestCase fails with link-local subinterfaces

pkremens:
https://issues.jboss.org/browse/WFCORE-794 - IPv6ScopeIdMatchUnitTestCase fails if some IPv6 subinterface exists

InetAddress.getByName(subInterface.getName()) returns null for virtual interfaces making the IPv6ScopeIdMatchUnitTestCase fail on servers with link-local IPv6 addressed sub-interfaces


bstansberry:
retest this please

PR #1416, opened 2016-02-12
[WFLY-5332]: Allow configure max-threads and core-threads independently. Allow idle thread reusage.

panossot:
[WFLY-5332] : https://issues.jboss.org/browse/WFLY-5332

The PR will fail till https://github.com/jbossas/jboss-threads/pull/19 is included.

### DIFF
--- a/threads/src/main/java/org/jboss/as/threads/ThreadsParser.java
+++ b/threads/src/main/java/org/jboss/as/threads/ThreadsParser.java
@@ -512,6 +512,11 @@ public final class ThreadsParser implements XMLStreamConstants, XMLElementReader
                     foundMaxThreads = true;
                     break;
                 }
+                case CORE_THREADS: {
+                    String scaledCount = parseCount(reader, threadsNamespace);
+                    PoolAttributeDefinitions.CORE_THREADS.parseAndSetParameter(scaledCount, op, reader);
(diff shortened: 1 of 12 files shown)


threads/src/main/java/org/jboss/as/threads/ThreadsParser.java
```diff
@@ -512,6 +513,12 @@ public String parseUnboundedQueueThreadPool(final XMLExtendedStreamReader reader
                     foundMaxThreads = true;
                     break;
                 }
+                case CORE_THREADS: {
```
ctomc: introducing new attribute to xsd schema requires new version of xsd

ctomc: never mind, it looks like it is bug in old version of parser.
2.0 parser handles is correctly as is.

panossot: I will check it.

panossot: New version of xsd is added.

bstansberry:
Closing to clean up the queue, as this won't compile. Please reopen if the library change this depends on is available.

PR #1421, opened 2016-02-16
WFCORE-1380 WFCORE-1177 WFCORE-1379 fix most common service.bat issues

ctomc:
- service.bat points user to wrong directory
- service.bat fails when wildfly folder path contains spaces
- Service can't be installed if the path contains a closed bracket
- few other problems community found

no need for downstream PR / JIRA as we use different scripts there.

### DIFF
--- a/core-feature-pack/src/main/resources/content/docs/contrib/scripts/service/service.bat
+++ b/core-feature-pack/src/main/resources/content/docs/contrib/scripts/service/service.bat
@@ -1,60 +1,102 @@
 @echo off
-setlocal EnableExtensions EnableDelayedExpansion
+
 REM -------------------------------------------------------------------------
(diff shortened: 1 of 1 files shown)


core-feature-pack/src/main/resources/content/docs/contrib/scripts/service/service.bat
```diff
@@ -1,60 +1,106 @@
 @echo off
-setlocal EnableExtensions EnableDelayedExpansion
+
+REM JBoss, the OpenSource webOS
+REM
+REM Distributable under LGPL license.
+REM See terms of license at gnu.org.
 REM -------------------------------------------------------------------------
```
bstansberry: Are these comments needed? They are kind of odd.

ctomc:
rebased

PR #1423, opened 2016-02-16
[WFCORE-1378] Forward port version 1.8 of the management schema.

darranl:
:warning: As this is a forward port it should only be merged once included in a previous product release.


bstansberry:
This also needs updates to org.jboss.as.domain.controller.transformers.DomainTransformers so slaves running the new version get transformed models.

bstansberry:
:+1:  -- good to go once 1.8 is ready downstream.

PR #1429, opened 2016-02-18
[WFCORE-683]: Wrong exception handling in ModuleLoadingResourceDefinition and use of jboss-module instead of reflection to get resource paths 

yersan:
Jira issue https://issues.jboss.org/browse/WFCORE-683

### DIFF
--- a/server/src/main/java/org/jboss/as/server/controller/resources/ModuleInfoHandler.java
+++ b/server/src/main/java/org/jboss/as/server/controller/resources/ModuleInfoHandler.java
@@ -128,7 +128,7 @@ public class ModuleInfoHandler implements OperationStepHandler {
     }
 
 
-    private ModuleLoaderMXBean getMxBean(ModuleLoader loader) throws ReflectiveOperationException {
+    protected ModuleLoaderMXBean getMxBean(ModuleLoader loader) throws ReflectiveOperationException {
         Field mxBeanField = ModuleLoader.class.getDeclaredField("mxBean");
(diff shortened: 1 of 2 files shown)


ctomc:
this is ok to test

kabir:
@yersan We need a bit of paperwork first:

yersan:
Hello @kabir 

Yes, I agree with the terms of the MIT License.

server/src/main/java/org/jboss/as/server/controller/resources/ModuleInfoHandler.java
```diff
@@ -128,7 +128,7 @@ public void execute(OperationContext context, ModelNode operation) throws Operat
     }
 
 
-    private ModuleLoaderMXBean getMxBean(ModuleLoader loader) throws ReflectiveOperationException {
+    protected ModuleLoaderMXBean getMxBean(ModuleLoader loader) throws ReflectiveOperationException {
```
bstansberry: Can you make this package protected? Thanks.

bstansberry:
Besides the minor comment above, looks good. Can you squash this down to 2 commits? The first one and then squash the rest into a single commit.

yersan:
Yes, of course, I'm going to change them in just two commits

yersan:
It seems that something was wrong running the tests after reduce my commits in just two. @bstansberry Although you pointed me to do a squash commit, I did a reset in my local branch and commit my last three commits in just one. I will check it because the same test doesn't pass in my local computer

PR #1430, opened 2016-02-18
WFCORE-1394 remove permgen defaults when started with no jvm configured

luck3y:
Note, we should probably add support for MetaSpacesize here too so we can get the current value from /host=foo/jvm, I experimented with this earlier, looks easy enough, I'll open as a separate Jira.


luck3y:
Retest this please

PR #1433, opened 2016-02-23
[WFCORE-1399]: Failure when resolving an expression with the default value containing a '$' at the end.

ehsavoie:
Checking if we are at the end of an expression.


ehsavoie:
retest this please

240ch:
Would be nice if this request was merged. I'm facing the same bug described in https://issues.jboss.org/browse/WFCORE-1399

PR #1435, opened 2016-02-23
[WFCORE-1352] Ensure the log context for embedded servers is cleaned …

jamezp:
…up and log messages end up in the correct log files.

A global embedded `LogContext` is used and reconfigured for each embedded server launched. This allows static loggers to log messages to the correct configuration.


jamezp:
retest this please

bstansberry:
I put the 3.x label on this as I expect the non-blocker JBEAP will get pushed out of EAP 7.0.0. If it does not, feel free to change the label to 2.x.

PR #1440, opened 2016-02-26
WFCORE-1413 Web console doesn't accept passwords with non-ascii chara…

spyrkob:
…cters

JIRA: https://issues.jboss.org/browse/WFCORE-1413

### DIFF
--- a/domain-http/interface/src/main/java/org/jboss/as/domain/http/server/ManagementHttpServer.java
+++ b/domain-http/interface/src/main/java/org/jboss/as/domain/http/server/ManagementHttpServer.java
@@ -34,9 +34,13 @@ import javax.net.ssl.SSLContext;
 
 import java.io.IOException;
 import java.net.InetSocketAddress;
+import java.nio.charset.Charset;
+import java.nio.charset.StandardCharsets;
 import java.util.ArrayList;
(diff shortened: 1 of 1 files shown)


bstansberry:
this is ok to test

spyrkob:
retest this please

domain-http/interface/src/main/java/org/jboss/as/domain/http/server/ManagementHttpServer.java
```diff
@@ -113,6 +118,7 @@
     private volatile AcceptingChannel<SslConnection> secureServer;
     private final SSLContext sslContext;
     private final SslClientAuthMode sslClientAuthMode;
+    private static Map<Pattern, Charset> userAgentCharsetMap = generateCharsetMap();
```
bstansberry: This needs the 'final' keyword.

Also, if you don't mind, I'd prefer that it's declared above the instance fields. That's a minor codestyle nit.

bstansberry:
Please squash the 2 commits when you rebase. Thanks!

bstansberry:
Closing in favor of #1478 (which is just a rebase of this.)

PR #1441, opened 2016-02-29
[WFCORE-1397] Fixed XnioWorker service installation in HttpManagementAddHandler for standalone mode

yersan:
This commit resolves a problem found with wildfly-core in standalone mode if http-upgrade-enabled configuration is false.

Jira issue link:
https://issues.jboss.org/browse/WFCORE-1397

### DIFF
--- a/server/src/main/java/org/jboss/as/server/operations/NativeManagementAddHandler.java
+++ b/server/src/main/java/org/jboss/as/server/operations/NativeManagementAddHandler.java
@@ -75,6 +75,8 @@ public class NativeManagementAddHandler extends AbstractAddStepHandler {
 
         final ServiceName endpointName = ManagementRemotingServices.MANAGEMENT_ENDPOINT;
         final String hostName = WildFlySecurityManager.getPropertyPrivileged(ServerEnvironment.NODE_NAME, null);
+
+        NativeManagementServices.installManagementWorkerService(serviceTarget, context.getServiceRegistry(false));
         NativeManagementServices.installRemotingServicesIfNotInstalled(serviceTarget, hostName, context.getServiceRegistry(false));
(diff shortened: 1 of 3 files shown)


server/src/main/java/org/jboss/as/server/operations/NativeManagementAddHandler.java
```diff
@@ -75,6 +75,8 @@ protected void performRuntime(OperationContext context, ModelNode operation, Mod
 
         final ServiceName endpointName = ManagementRemotingServices.MANAGEMENT_ENDPOINT;
         final String hostName = WildFlySecurityManager.getPropertyPrivileged(ServerEnvironment.NODE_NAME, null);
+
+        NativeManagementServices.installManagementWorkerService(serviceTarget, context.getServiceRegistry(false));
```
bstansberry: I don't think this one is needed as NativeManagementServices.installRemotingServices calls ManagementWorkerService.installService.

yersan: I followed the same piece of code used in the controller version: org.jboss.as.host.controller.operations.HttpManagementAddHandler#installHttpManagementServices. 

In the controller version, before install the xnio mgmt worker first it checks if the service is not already registered in the ServiceRegistry. I used the same logic here in the server version. That was the reason of line 79, I call NativeManagementServices.installManagementWorkerService method there as the controller version does.

So, if I understand you well, you suggest to remove the call in line 79 and call directly to ManagementWorkerService.installService, am I right?

bstansberry: Oh, I see, you changed NativeManagementServices.installRemotingServices to no longer call ManagementWorkerService.installService. I didn't notice the removal before.

PR #1443, opened 2016-03-01
WFCORE-1418 Reloading host-controller via http-api puts the HC into u…

ctomc:
…nresponsive state

jira: https://issues.jboss.org/browse/WFCORE-1418
downstream: https://github.com/jbossas/wildfly-core-eap/pull/193

### DIFF
--- a/domain-http/interface/src/main/java/org/jboss/as/domain/http/server/ManagementHttpServer.java
+++ b/domain-http/interface/src/main/java/org/jboss/as/domain/http/server/ManagementHttpServer.java
@@ -64,6 +64,7 @@ import io.undertow.server.handlers.cache.DirectBufferCache;
 import io.undertow.server.handlers.error.SimpleErrorPageHandler;
 import io.undertow.server.protocol.http.HttpOpenListener;
 import java.util.Collection;
+import java.util.concurrent.Executor;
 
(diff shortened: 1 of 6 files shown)


domain-http/interface/src/main/java/org/jboss/as/domain/http/server/ManagementHttpServer.java
```diff
@@ -245,7 +236,10 @@ private static void setupOpenListener(HttpOpenListener listener, ModelController
         }
 
         ManagementRootConsoleRedirectHandler rootConsoleRedirectHandler = new ManagementRootConsoleRedirectHandler(consoleHandler);
-        DomainApiCheckHandler domainApiHandler = new DomainApiCheckHandler(modelController, controlledProcessStateService, allowedOrigins);
+        HttpHandler domainApiHandler = InExecutorHandler.wrap(
+                managementExecutor,
+                new DomainApiCheckHandler(modelController, controlledProcessStateService, allowedOrigins)
+        );
```
bstansberry: This is a nit; ignore it unless you need to touch this PR for some other reason:

This block would be better starting at L249, right before the HttpHandler readinessHandler... line. That groups the rootConsoleRedirectHandler logic together and then the code starts on the domainApiHandler.

PR #1444, opened 2016-03-01
[WFCORE-1419] Do not set the recursive attribute on the read-resource operation if not explicitly defined.

jamezp:
Removes the `recursive` attribute being defined on a `read-resource` operation if the `createReadResourceOperation(ModelNode address)` is used.


bstansberry:
retest this please

The change looks fine.

PR #1445, opened 2016-03-02
[WFCORE-1420] Add a timeout param to shutdown command

bstansberry:


bstansberry:
I did some label fiddling on this as it's now being proposed for 2.x. I removed "Ready for Merge" as now EAP-ack is wanted, but Kabir's code review ok on it is still valid, so it's ready for merge once the acks are in.

PR #1452, opened 2016-03-07
[WFCORE-515] Unable to create create backup copies of configuration file if main file folder is unwritable by the WildFly process.

yersan:
Server is unable to start if it is started with --read-only-domain-config or --read-only-server-config using a folder unwritable by the Wildfly process. This commit resolves this issue using the configuration dir to create the configuration file backup

Jira issue:
https://issues.jboss.org/browse/WFCORE-515


bstansberry:
Looks good; thanks!

PR #1453, opened 2016-03-08
WFCORE-915 allow concurrent registration of slave hosts

luck3y:
Initial version for review of concurrent host registration using a shared mode lock vs the existing exclusive ModelController lock. Initial (not very scientific) testing shows this to be about 3x as fast as sequential registration. Some cleanup is probably still needed, and I'm still testing some failure scenarios.

I did attempt a version of this using RentrantReadWriteLock, but since that really wanted to manage its own thread lock owners, it turned out to be simpler to use the shared / exclusive modes in AbstractQueuedSynchronizer.

Thoughts & comments appreciated :)

### DIFF
--- a/controller/src/main/java/org/jboss/as/controller/PersistentResourceXMLDescription.java
+++ b/controller/src/main/java/org/jboss/as/controller/PersistentResourceXMLDescription.java
@@ -370,7 +370,7 @@ public final class PersistentResourceXMLDescription {
         //we sort attributes to make sure that attributes that marshall to elements are last
         for (AttributeDefinition ad : attributes.stream().sorted((o1, o2) -> o1.getParser().isParseAsElement() ? 1 : -1).collect(Collectors.toList())) {
             AttributeMarshaller marshaller = attributeMarshallers.getOrDefault(ad.getName(), ad.getAttributeMarshaller());
-            if (marshaller.isMarshallable(ad, model, false)) {
+            if (marshaller.isMarshallable(ad, model, marshallDefaultValues)) {
                 if (!started && group != null) {
                     if (possibleElements) {
                         writer.writeStartElement(group);


controller/src/main/java/org/jboss/as/controller/ModelControllerImpl.java
```diff
@@ -809,6 +809,16 @@ void acquireLock(Integer permit, final boolean interruptibly) throws Interrupted
...
+        if (interruptibly) {
+            //noinspection LockAcquiredButNotSafelyReleased
+            controllerLock.lockSharedInterruptibly(permit);
+        } else {
+            //noinspection LockAcquiredButNotSafelyReleased
+            controllerLock.lockShared(permit);
+        }
+    }
+
     boolean acquireLock(Integer permit, final boolean interruptibly, long timeout) throws InterruptedException {
```
bstansberry: This should be acquireWriteLock now, to be consistent.

host-controller/src/main/java/org/jboss/as/host/controller/mgmt/HostControllerRegistrationHandler.java
```diff
@@ -204,6 +205,22 @@ public HostControllerRegistrationHandler(ManagementChannelHandler handler, Domai
          */
         ModelNode executeReadOnly(ModelNode operation, Resource model, OperationStepHandler handler, ModelController.OperationTransactionControl control);
 
+        /**
+         * Registers a slave host controller. A non-exclusive read lock is held for the duration of this registration
+         * process, blocking writes, but allowing the initiation of additional read locks.
```
bstansberry: This should have a comment about how if the call succeeds, completeRegistration *must* be called.

host-controller/src/main/java/org/jboss/as/host/controller/mgmt/HostControllerRegistrationHandler.java
```diff
@@ -204,6 +205,22 @@ public HostControllerRegistrationHandler(ManagementChannelHandler handler, Domai
...
+         * @param operationId - the operation ID for this registration
+         * @param hostInfo - the registering host's HostInfo
+         * @throws OperationFailedException - if the lock acquisition fails
+         */
+        void beginRegistration(final Integer operationId, final HostInfo hostInfo) throws OperationFailedException;
```
bstansberry: Should these methods be renamed and the javadoc changed? For example, beginRegistration doesn't "Register a slave host controller". It just acquires a lock.

A generic name like beginRegistration is ok if you want flexibility to do more stuff in the future. In that case the javadoc just needs a tweak so it doesn't sound like it does more that it really does.

luck3y: Yeah, good point. I had more code in here at one point, but doing so involves exposing a lot of stuff owned by HCRH at the moment, like RegistrationContext etc.

I'm thinking of just calling them acquireReadlock() / releaseReadlock() now.

host-controller/src/main/java/org/jboss/as/host/controller/mgmt/HostControllerRegistrationHandler.java
```diff
@@ -285,9 +307,9 @@ protected HostRegistrationStepHandler(final TransformerRegistry transformerRegis
 
         @Override
         public void execute(final OperationContext context, final ModelNode operation) throws OperationFailedException {
-            // First lock the domain controller
-            context.acquireControllerLock();
-            // Check with the controller lock held
+            // acquires the shared-mode read lock.
+            operationExecutor.beginRegistration(registrationContext.activeOperation.getOperationId(), registrationContext.hostInfo);
+
```
bstansberry: After this gets called we have to be sure that completeRegistration gets invoked no matter what happens in the chain of requests.

Actually, why can't it be released in this method via

1) A ResultHandler passed to context.completeStep() and
2) a finally block that calls it if a failure occurs before the context.completeStep() call

That would have the same locking/unlocking semantics as the existing context.acquireControllerLock() (except of course it's the shared lock). The way the OC works with locks is if a step acquires the lock, on the way out after any ResultHandler for that step runs the lock is released by the OC.

luck3y: Good, idea, I'll change to this.

controller/src/main/java/org/jboss/as/controller/ModelControllerLock.java
```diff
@@ -102,6 +144,28 @@ protected synchronized boolean tryAcquire(int permit) {
         }
 
         @Override
+        protected synchronized int tryAcquireShared(int permit) {
```
n1hility: The class was this way before, but its sort of strange to see an AQS with synced methods. If the access to the "owner" which IIUC is what "permitHolder" represents is done properly, then the sync is unnecessary. If there is some special case where an atomic update of both is required AbstractQueuedLongSynchronizer using a bit mask might be a better solution. This is mostly a quibble though, and could be addressed in a subsequent update.

host-controller/src/main/java/org/jboss/as/host/controller/mgmt/HostControllerRegistrationHandler.java
```diff
@@ -332,9 +354,15 @@ public void execute(final OperationContext context, final ModelNode operation) t
...
-            context.addStep(READ_DOMAIN_MODEL.getOperation(), handler, OperationContext.Stage.MODEL);
+
+            final Transformers.ResourceIgnoredTransformationRegistry ignoredTransformationRegistry;
+            final Resource resource = context.readResourceFromRoot(PathAddress.EMPTY_ADDRESS);
+            final ReadMasterDomainModelUtil.RequiredConfigurationHolder rc = hostInfo.populateRequiredConfigurationHolder(resource, domainController.getExtensionRegistry());
+            ignoredTransformationRegistry = ReadMasterDomainModelUtil.createHostIgnoredRegistry(hostInfo, rc);
+
+            final ReadMasterDomainModelUtil readUtil = ReadMasterDomainModelUtil.readMasterDomainResourcesForInitialConnect(transformers,
+                    transformationInputs, ignoredTransformationRegistry, transformationInputs.getRootResource());
+            context.getResult().set(readUtil.getDescribedResources());
```
bstansberry: My guess is that you pulled this logic in here because ReadMasterDomainModelHandler calls context.acquireControllerLock(). As does ReadDomainModelHandler.

We'll need to factor this somehow so this logic only exists in one place, as it's really obscure and we'll never remember to fix any bugs in > 1 spot. Perhaps just a boolean flag to ReadMasterDomainModelHandler and ReadDomainModelHandler to turn off the context.acquireControllerLock()?

host-controller/src/main/java/org/jboss/as/host/controller/mgmt/HostControllerRegistrationHandler.java
```diff
@@ -204,6 +205,22 @@ public HostControllerRegistrationHandler(ManagementChannelHandler handler, Domai
          */
         ModelNode executeReadOnly(ModelNode operation, Resource model, OperationStepHandler handler, ModelController.OperationTransactionControl control);
 
+        /**
+         * Registers a slave host controller. A non-exclusive read lock is held for the duration of this registration
+         * process, blocking writes, but allowing the initiation of additional read locks.
+         * @param operationId - the operation ID for this registration
```
bstansberry: Cannot be {@code null}

host-controller/src/main/java/org/jboss/as/host/controller/mgmt/HostControllerRegistrationHandler.java
```diff
@@ -204,6 +205,22 @@ public HostControllerRegistrationHandler(ManagementChannelHandler handler, Domai
...
+         * @param hostInfo - the registering host's HostInfo
```
bstansberry: Cannot be {@code null}

host-controller/src/main/java/org/jboss/as/host/controller/mgmt/HostControllerRegistrationHandler.java
luck3y: Yeah, can definitely do that, those methods were public, so I wasn't so sure how easily changed they were.

controller/src/main/java/org/jboss/as/controller/ModelControllerLock.java
luck3y: Oh, thanks for pointing that out, I'll fix as part of this.

host-controller/src/main/java/org/jboss/as/host/controller/mgmt/HostControllerRegistrationHandler.java
```diff
@@ -204,6 +205,22 @@ public HostControllerRegistrationHandler(ManagementChannelHandler handler, Domai
...
+         * @throws OperationFailedException - if the lock acquisition fails
```
bstansberry: OperationFailedException is an "Exception indicating an operation has failed due to a client mistake (e.g. an operation with invalid parameters was invoked.) Should not be used to report server failures." The failures here indicate some sort of server problem (e.g. an operation had no id, or a thread was interrupted).

I don't want to use OFE for any other purpose because lots of devs, particularly subsystem authors, tend to misuse it. So I don't want examples of that in the kernel code.

Perhaps these can just throw IAE (for invalid/null params) and InterruptedException.

host-controller/src/main/java/org/jboss/as/host/controller/mgmt/HostControllerRegistrationHandler.java
```diff
@@ -204,6 +205,22 @@ public HostControllerRegistrationHandler(ManagementChannelHandler handler, Domai
...
+         */
+        void beginRegistration(final Integer operationId, final HostInfo hostInfo) throws OperationFailedException;
+
+        /**
+         * Complete a host registration by unlocking the shared mode lock.
+         * @param operationId - the operationId for this registration
+         * @param hostInfo - the registering host's HostInfo
+         * @throws OperationFailedException - if something goes wrong with releasing the lock
```
bstansberry: Same comment re: OFE. This one can just throw IAE.

controller/src/main/java/org/jboss/as/controller/AbstractControllerService.java
```diff
@@ -543,6 +543,28 @@ protected OperationStepHandler createExtraValidationStepHandler() {
...
 
+    /*
+     * This attempts to obtain a shared mode ModelControllerLock, which will allow
+     * multiple acquisition for concurrent reads while blocking acquisition of the exclusive
+     * lock. Acquisition of the the exclusive lock will block read locks, until released.
+     */
+    protected void acquireReadLock(int operationId) throws OperationFailedException {
+        try {
+            // block until a read lock is available
+            controller.acquireReadLock(operationId, false);
```
bstansberry: This should be interruptible (i.e. 'true').

host-controller/src/main/java/org/jboss/as/host/controller/mgmt/HostControllerRegistrationHandler.java
bstansberry: Re: exception handling, if beginRegistration throws InterruptedException, the following is equivalent logic to what happens when context.acquireControllerLock() is called.

```
try {
    operationExecutor.beginRegistration(registrationContext.activeOperation.getOperationId(), registrationContext.hostInfo);
} catch (InterruptedException e) {
    Thread.currentThread().interrupt();
    throw ControllerLogger.ROOT_LOGGER.operationCancelledAsynchronously();
}
```

If an InterruptedException occurs in an OperationStepHandler, Thread.currentThread().interrupt() should be called so when OSH.execute() returns the OperationContext is aware of the request to stop execution and its logic for dealing with that gets a chance to run. So the handling of InterruptedException can happen here, or in the deeper layers, but whoever catches it needs to call Thread.currentThread().interrupt().

host-controller/src/main/java/org/jboss/as/host/controller/mgmt/HostControllerRegistrationHandler.java
bstansberry: Anything in the host-controller module can be changed at will. That module is internal to the kernel and no one else has any business depending on it.

Even within modules like that though it's always good though to reduce visibility of things if they are broader than needed. Lots of stuff has to be public though for use in other packages.

Of the "kernel" (i.e. non-Extension, and non-test-related) modules in core, only these expose API usable outside the kernel:

cli
controller
controller-client
embedded
launcher
network
self-contained
server

The controller and server modules need to be broken up to separate out externally usable stuff from internals. Big task. The module.xml for some others, like cli and controller-client, could do with some tweaking to limit what they export.

host-controller/src/main/java/org/jboss/as/host/controller/DomainModelControllerService.java
```diff
@@ -1268,6 +1267,24 @@ public ModelNode executeReadOnly(ModelNode operation, OperationStepHandler handl
         public ModelNode executeReadOnly(ModelNode operation, Resource model, OperationStepHandler handler, OperationTransactionControl control) {
             return executeReadOnlyOperation(operation, model, control, handler);
         }
+
+        @Override
+        public void acquireReadlock(final Integer operationID) throws IllegalArgumentException, InterruptedException {
+            if (operationID == null) {
+                throw new IllegalArgumentException("operationID may not be null.");
+            }
```
bstansberry: Needs i18n, or instead of the IAE an "assert operationID != null;" is probably ok. I use assert a lot for checks of internal contracts.

i18n is easy though:

```
throw HostControllerLogger.DOMAIN_LOGGER.nullVar("operationID");
```

Same comment applies to releaseReadLock.

controller/src/main/java/org/jboss/as/controller/ModelControllerLock.java
```diff
@@ -23,26 +23,36 @@
...
 
-    boolean lock(Integer permit, long timeout, TimeUnit unit) throws InterruptedException {
+    void lockShared(final Integer permit) {
+        if (permit == null) {
+            throw new IllegalArgumentException();
+        }
+        sync.acquireShared(permit);
+    }
+
+    boolean lock(final Integer permit, final long timeout, final TimeUnit unit) throws InterruptedException {
```
bstansberry: This doesn't need to declare "throws InterruptedException" since it doesn't throw that.

This is a preexisting problem but the new "lockShared" has it too now, so might as well clean up the signature of both of them.

controller/src/main/java/org/jboss/as/controller/ModelControllerLock.java
```diff
@@ -52,42 +62,73 @@ boolean lock(Integer permit, long timeout, TimeUnit unit) throws InterruptedExce
         return result;
     }
 
-    void lockInterruptibly(Integer permit) throws InterruptedException {
+    boolean lockShared(final Integer permit, final long timeout, final TimeUnit unit) throws InterruptedException {
```
bstansberry: See comment re the "lock" method signature.

controller/src/main/java/org/jboss/as/controller/ModelControllerLock.java
luck3y: Ah, missed that, thanks!

controller/src/main/java/org/jboss/as/controller/ModelControllerLock.java
```diff
@@ -52,42 +62,73 @@ boolean lock(Integer permit, long timeout, TimeUnit unit) throws InterruptedExce
...
 
         @Override
-        protected synchronized boolean tryAcquire(int permit) {
+        protected boolean tryAcquire(final int permit) {
             if (compareAndSetState(0, 1)) {
-                permitHolder.set(permit);
+                permitHolder = permit;
                 return true;
-            } else if (permitHolder.get().equals(permit)) {
+            } else if (permitHolder != null && permitHolder.equals(permit)) {
```
bstansberry: The permitHolder.equals(permit) call can NPE if tryRelease has been called concurrently and permitHolder was set to null after the null check.

The existing code would have had this problem too, but the methods being synchronized prevented that.

At the cost of an object allocation, this is safe:

```
} else if (Integer.valueOf(permit).equals(permitHolder)) {
```

controller/src/main/java/org/jboss/as/controller/ModelControllerLock.java
```diff
@@ -52,42 +62,73 @@ boolean lock(Integer permit, long timeout, TimeUnit unit) throws InterruptedExce
...
+
+    boolean detectDeadlockAndGetLock(final int permit) {
         return sync.tryAcquire(permit);
     }
 
     private class Sync extends AbstractQueuedSynchronizer {
         private static final long serialVersionUID = 1L;
 
-        private final AtomicReference<Object> permitHolder = new AtomicReference<>(null);
+        private Integer permitHolder = null;
```
bstansberry: This needs to be volatile for tryAcquireShared to see the current value.

controller/src/main/java/org/jboss/as/controller/ModelControllerLock.java
```diff
@@ -102,12 +143,31 @@ protected synchronized boolean tryAcquire(int permit) {
...
-        protected synchronized boolean tryRelease(int permit) {
-            final Object value = permitHolder.get();
-            if (value == null) {
+        protected int tryAcquireShared(final int permit) {
+            if (permitHolder != null) {
+                return -1;
+            }
+            if (compareAndSetState(0, 1)) {
+                return 1;
+            } else {
```
bstansberry: If this thread is racing with one calling tryAcquire, at this point we don't the status of 'permit'. Perhaps the other thread has set state to 1, but hasn't set permit yet. Or perhaps by now it has set permit. We just know it hadn't set state to 1 yet when our thread got to L150.

Deferring the if (permitHolder != null) check until we get to this else block avoids that, but now we're racing with a thread calling tryRelease. Perhaps checking both places will help; I haven't thought that through yet. :)

I think these issues are why the methods were synchronized.

controller/src/main/java/org/jboss/as/controller/ModelControllerLock.java
luck3y: This one is done similarly to the owning thread in AbstractOwnableSynchronizer, I think theres a comment there saying why it doesn't need to be volatile, I'll find it.

controller/src/main/java/org/jboss/as/controller/ModelControllerLock.java
luck3y: Yeah, this has been bugging me too, I've been looking at this code quite a bit today. Internally I think the read / write could synchronize on permitholder, or a bitmask in the internal state perhaps. That would at least allow just the single getValue() and CAS in those methods, which is how it seems to be intended to work.

luck3y: I'm thinking that the options here are:

(1) Go back to using synchronized, and fix this as a seperate issue
(2) Synchronize on permitHolder, which would basically end up being the same as #1, and I think I'd rather have permitHolder just be an int anyway, rather than an Integer etc.
(3) Subclass AQS, and provide methods to getValue() / setValue(long). We could bitmask the int to store the permit in the upper bits, allowing just the use of a single CAS.
(4) Change operationID to use 28(or some amount) bits instead of the current 32 and mask the count in the int state variable in AQS.

luck3y: I think the best option right now is to put this back to how it was, and open a seperate issue to fix it :)

luck3y:
I've reverted the ModelController sync changes, and I'll open a seperate jira issue.

luck3y:
retest this please

luck3y:
retest this please

controller/src/main/java/org/jboss/as/controller/ModelControllerLock.java
bstansberry: Agreed.

These shouldn't be highly contended (or even highly used) methods, so it's ok IMO to use a brute-force synchronized until we are comfortable that something better is solid.

bstansberry: Re: "(3) Subclass AQS, and provide methods to getValue() / setValue(long). We could bitmask the int to store the permit in the upper bits, allowing just the use of a single CAS."

I think this is what Jason was talking about in his comment mentioning AbstractQueuedLongSynchronizer.

luck3y:
retest this please

controller/src/main/java/org/jboss/as/controller/ModelControllerLock.java
luck3y: Sorry I missed this with all the other spam from that PR. I realized this
after I'd already found AQLS, then re-read Jason's comment after I'd done
it ;)

I ended up closing this PR, new one is:
https://github.com/wildfly/wildfly-core/pull/1457

PR #1455, opened 2016-03-08
[WFCORE-1430]: Deployment scanner logs ERROR when server is shutting down.

ehsavoie:
When server is shutting down we are setting the scanEnabled to false and ignore errors on ScanContext creation since the scan doesn't make any sense.


bstansberry:
Looks good.

This isn't related to this PR but I noticed it when reviewing it, so something to look into some day: the ScanResult class has two fields 'requireUndeploy' and 'tasks' that are unused. So if they aren't going to be used someday, the whole ScanResult class really amounts to the 'scheduleRescan' boolean.

PR #1459, opened 2016-03-15
[WFCORE-1435]:  Users with low privileges can see logged operations they shouldn't be able to see

ehsavoie:
Checking read-access to the operation address parent

Jira: https://issues.jboss.org/browse/WFCORE-1435


bstansberry:
Assuming tests pass, :+1:

PR #1462, opened 2016-03-16
WFCORE-1425 embedded slave started with host-slave.xml has incorrect hostname

luck3y:
This also removes the embedded specific environment setup in favor of the existing determineEnvironment / HostControllerEnvironment.

### DIFF
--- a/embedded/src/main/java/org/wildfly/core/embedded/EmbeddedHostControllerFactory.java
+++ b/embedded/src/main/java/org/wildfly/core/embedded/EmbeddedHostControllerFactory.java
@@ -26,17 +26,15 @@ import java.beans.PropertyChangeEvent;
 import java.beans.PropertyChangeListener;
 import java.io.File;
 import java.io.IOException;
-import java.lang.management.ManagementFactory;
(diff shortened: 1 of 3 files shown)


embedded/src/main/java/org/wildfly/core/embedded/EmbeddedHostControllerFactory.java
```diff
@@ -401,95 +399,28 @@ private void exit() {
...
-                        }
-                    }
-                }
-            } catch (Exception e) {
-                EmbeddedLogger.ROOT_LOGGER.cannotSetupEmbeddedServer(e);
-            }
-            return hostSystemProperties;
-        }
-
         private static HostControllerEnvironment createHostControllerEnvironment(File jbossHome, String[] cmdargs, String hostName, long startTime) {
```
kabir: hostName seems to be unused now

bstansberry: Once you drop that param, this looks ready to go. I like how it's simpler. :)

PR #1463, opened 2016-03-17
[WFLY-4882] Ensure the ObjectList attribute is initialized with an em…

sguilhen:
…pty list when its enclosing element is present but empty. 

This is being done so that in the security manager subsystem we are able to differentiate between an empty and an absent maximum permission set.

Link to the upstream Jira:
https://issues.jboss.org/browse/WFLY-4882

### DIFF
--- a/controller/src/main/java/org/jboss/as/controller/AttributeParser.java
+++ b/controller/src/main/java/org/jboss/as/controller/AttributeParser.java
@@ -194,6 +194,10 @@ public abstract class AttributeParser {
                 }
                 ParseUtils.requireNoContent(reader);
             }
+            // the object list element was specified but it is empty, so initialize it with an empty list.
+            if (!operation.get(attribute.getName()).isDefined()) {
+                operation.get(attribute.getName()).setEmptyList();
+            }
         }
     };
 


controller/src/main/java/org/jboss/as/controller/AttributeParser.java
```diff
@@ -194,6 +194,10 @@ public void parseElement(AttributeDefinition attribute, XMLExtendedStreamReader
                 }
                 ParseUtils.requireNoContent(reader);
             }
+            // the object list element was specified but it is empty, so initialize it with an empty list.
+            if (!operation.get(attribute.getName()).isDefined()) {
```
ctomc: This shouldn't be needed, as when you are resolving attribute with AD#resolveModelAttribute() it should take care of that.
Setting empty values to attributes in general is wrong and redundant.

Looking at how resolving of this attribute is done in https://github.com/wildfly/wildfly/blob/master/security-manager/src/main/java/org/wildfly/extension/security/manager/SecurityManagerSubsystemAdd.java#L113
It is clearly wrong.

sguilhen:
You are right, Tomaz. I should be handling this via AttributeDefinition.

PR #1467, opened 2016-03-21
[WFCORE-1444]: Fix schema as native-interface shouldn't be mandatory for host-management-interfaces

ehsavoie:
Adding minOccurs="0" for native-interface.

Jira: https://issues.jboss.org/browse/WFCORE-1444


bstansberry:
Isn't the 4.0 xsd the same? AFAIK it's the same as 4.1 except for addition of some FIPs stuff.

PR #1470, opened 2016-03-23
WFCORE-1448 AttributeParser for ObjectListAttributeDefinitions doesn'…

ctomc:
…t differentiate between empty and absent object lists

- also fixes WFCORE-1440 Expression resolver breaks return value for empty lists
- fixes rare scenario in which wrapped properties attribute without group would fail if same element also had child resources
- make sure that same thing applys also to maps
- add convinence method ObjectListAttributeDefinition.unwarp

downstream jiras JBEAP-3924 JBEAP-408

### DIFF
--- a/controller/src/main/java/org/jboss/as/controller/AttributeMarshallers.java
+++ b/controller/src/main/java/org/jboss/as/controller/AttributeMarshallers.java
@@ -55,12 +55,15 @@ public interface AttributeMarshallers {
 
         @Override
         public void marshallAsElement(AttributeDefinition attribute, ModelNode resourceModel, boolean marshallDefault, XMLStreamWriter writer) throws XMLStreamException {
-            if (!resourceModel.hasDefined(attribute.getName())) {
+            String wrapper =  wrapperElement == null ? attribute.getName() : wrapperElement;
+            resourceModel = resourceModel.get(attribute.getName());
+            if (!resourceModel.isDefined()){//this is scenario to handle WFCORE-1448 cases
+                writer.writeEmptyElement(wrapper);
                 return;
             }
-            resourceModel = resourceModel.get(attribute.getName());
+
             if (wrapElement) {
-                writer.writeStartElement(wrapperElement == null ? attribute.getName() : wrapperElement);
+                writer.writeStartElement(wrapper);
             }
             for (ModelNode property : resourceModel.asList()) {
                 writer.writeEmptyElement(elementName);
--- a/controller/src/main/java/org/jboss/as/controller/ObjectListAttributeDefinition.java
+++ b/controller/src/main/java/org/jboss/as/controller/ObjectListAttributeDefinition.java
@@ -134,11 +134,11 @@ public class ObjectListAttributeDefinition extends ListAttributeDefinition {
         if (superResult.getType() != ModelType.LIST) {
             return superResult;
         }
-
         // Resolve each element.
         // Don't mess with the original value
         ModelNode clone = superResult == value ? value.clone() : superResult;
         ModelNode result = new ModelNode();
+        result.setEmptyList();
         for (ModelNode element : clone.asList()) {
             result.add(valueType.resolveValue(resolver, element));
         }
@@ -228,6 +228,7 @@ public class ObjectListAttributeDefinition extends ListAttributeDefinition {
     ObjectTypeAttributeDefinition getValueType() {
         return valueType;
     }
+
 
     public static final class Builder extends ListAttributeDefinition.Builder<Builder, ObjectListAttributeDefinition> {
         private final ObjectTypeAttributeDefinition valueType;
--- a/controller/src/main/java/org/jboss/as/controller/PersistentResourceXMLDescription.java
+++ b/controller/src/main/java/org/jboss/as/controller/PersistentResourceXMLDescription.java
@@ -1,16 +1,5 @@
 package org.jboss.as.controller;
 
-import org.jboss.as.controller.descriptions.ModelDescriptionConstants;
-import org.jboss.as.controller.logging.ControllerLogger;
-import org.jboss.as.controller.operations.common.Util;
-import org.jboss.as.controller.parsing.Element;
-import org.jboss.as.controller.parsing.ParseUtils;
-import org.jboss.dmr.ModelNode;
-import org.jboss.staxmapper.XMLExtendedStreamReader;
-import org.jboss.staxmapper.XMLExtendedStreamWriter;
-
-import javax.xml.stream.XMLStreamConstants;
-import javax.xml.stream.XMLStreamException;
 import java.util.ArrayList;
 import java.util.Collection;
 import java.util.Collections;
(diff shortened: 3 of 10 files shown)


controller/src/main/java/org/jboss/as/controller/AttributeMarshallers.java
```diff
@@ -55,12 +55,18 @@ public boolean isMarshallable(AttributeDefinition attribute, ModelNode resourceM
 
         @Override
         public void marshallAsElement(AttributeDefinition attribute, ModelNode resourceModel, boolean marshallDefault, XMLStreamWriter writer) throws XMLStreamException {
-            if (!resourceModel.hasDefined(attribute.getName())) {
+            if (!resourceModel.has(attribute.getName())) {
```
bstansberry: For a properly implemented resource, this will always return 'true' because AbstractAddHandler.populateModel() will iterate over all attributes, for each adding a child (defined or undefined) to the resource's model.

I believe the behavior should be:

if (!resourceModel.hasDefined(attribute.getName())) -> return without marshalling anything for the attribute
List<ModelNode> list = resource.get(attribute.getName()).asList();
if (list.isEmpty()) {
   if (wrapElement) -> marshal an empty wrapper
   else -> return without marshalling anything (see below)
}
else -> [marshal a start element] / iterate over list / [marshal end element].

A problem here is "if (list.isEmpty() && !wrapElement)". This indicates a subtle programming error, where the developer doesn't provide an xml schema that prevents ambiguity between the 'null' and 'empty collection' cases. I don't think we should try to deal with this at this stage of the release process though, other than perhaps adding a TODO and filing a JIRA. A fix would likely involve throwing an ISE or AssertionError in at least some cases where ^^^ is true, and that's too risky right now.

controller/src/main/java/org/jboss/as/controller/ObjectListAttributeDefinition.java
```diff
@@ -131,14 +134,16 @@ public ModelNode resolveValue(ExpressionResolver resolver, ModelNode value) thro
...
         if (superResult.getType() != ModelType.LIST) {
             return superResult;
+        }else{
+            result.setEmptyList();
         }
 
         // Resolve each element.
         // Don't mess with the original value
         ModelNode clone = superResult == value ? value.clone() : superResult;
-        ModelNode result = new ModelNode();
```
bstansberry: Nit:

Can the changes at L137-141 be removed and then here we have:

ModelNode clone = superResult == value ? value.clone() : superResult;
ModelNode result = new ModelNode();
result.setEmptyList();

Just seems a bit easier to understand with 'result' not being introduced in the middle of the logic for checking it's a LIST

controller/src/main/java/org/jboss/as/controller/ObjectListAttributeDefinition.java
```diff
@@ -229,6 +234,19 @@ ObjectTypeAttributeDefinition getValueType() {
         return valueType;
     }
 
+
+    public List<ModelNode> unwrap(final ExpressionResolver context, final ModelNode model) throws OperationFailedException {
+        ModelNode value = resolveModelAttribute(context, model);
+        if (!model.isDefined()) {
+            return Collections.emptyList();
```
bstansberry: Why not null? I think a method like this should support null, empty and non-empty. If you don't want the caller to have to deal with null, add an overloaded variant.

This method needs clear javadoc to establish its contract.

controller/src/main/java/org/jboss/as/controller/ObjectListAttributeDefinition.java
```diff
@@ -229,6 +234,19 @@ ObjectTypeAttributeDefinition getValueType() {
         return valueType;
     }
 
+
+    public List<ModelNode> unwrap(final ExpressionResolver context, final ModelNode model) throws OperationFailedException {
```
bstansberry: Nit: 'resolver' is a better param name.

controller/src/main/java/org/jboss/as/controller/PersistentResourceXMLDescription.java
```diff
@@ -174,6 +174,7 @@ private void parseGroup(XMLExtendedStreamReader reader, ModelNode op, boolean wi
             if (reader.nextTag() == START_ELEMENT && attributeElements.containsKey(reader.getLocalName())) {
                 AttributeDefinition ad = attributeElements.get(reader.getLocalName());
                 ad.getParser().parseElement(ad, reader, op);
+                ParseUtils.requireNoContent(reader);
```
bstansberry: Why this? Does it fix a problem detected by your new tests? 

What happens if the attribute group includes > 1 element attribute? AIUI parseAttributes will only parse xml attributes (because attributeGroups.isEmpty() is false; see L201 below), so parsing of element attributes is either done here or when this method returns and the loop in parseAttributeGroups continues. So if this fails if there is > 1 element attribute in a group, there doesn't seem to be way to have that kind of structure.

controller/src/main/java/org/jboss/as/controller/AttributeMarshallers.java
ctomc: The initial if (!resourceModel.has(attribute.getName())) {
was there as part of the not using setEmptyObject() and is at this point completely irrelevant and I have removed it.

controller/src/main/java/org/jboss/as/controller/ObjectListAttributeDefinition.java
ctomc: fixed

controller/src/main/java/org/jboss/as/controller/ObjectListAttributeDefinition.java
ctomc: this is a good point and a tricky at the same time.
good point become we are properly introducing the third state and as such we should have null.
tricky because of StringListAD and PropertiesAD already return emptyList/Map for undefined for same #unwrap method.
If anything we should have it unified across the board.
At this point in time, I would rather omit this method and return to it after we have eap7 out the door.
created https://issues.jboss.org/browse/WFCORE-1451 for unifying them, for now I will remove this method from this PR.

controller/src/main/java/org/jboss/as/controller/PersistentResourceXMLDescription.java
ctomc: Yes it fixes a problem that new tests found. It was broken in scenario where you had resource that had wrapped properties attribute and than some children. It would fail to read children properly.
Inital bug was using wrong variable for wrapping attribute name when checking for "complete read" in https://github.com/wildfly/wildfly-core/pull/1470/files#diff-429895dde8feaf784ee6dc586357c5bbL101 it worked fine if said attribute was part of attribute group. Most other changes like https://github.com/wildfly/wildfly-core/pull/1470/files#diff-43cfbbd8554a2dc5cbf83398bbb56f83L208 in parser ware just to fix the original bug.
Test that test this is mail-parser.xml https://github.com/wildfly/wildfly-core/pull/1470/files#diff-5f866925622285f598bd12e5bc40920dR21

So in short, when we had wrapped properties/map attribute inside a resource that also has children it would break. 
But if you had same scenario but have said attribute as part of attribute group it would work fine.

controller/src/main/java/org/jboss/as/controller/AttributeMarshallers.java
ctomc: Another reason why this method (and similarly the List marshaller one) look bit funky is becouse of bugs in various jaxb parsers.

the whole
```
if (!resourceModel.isDefined()){//this is scenario to handle WFCORE-1448 cases
   writer.writeEmptyElement(wrapper);
   return;
}
```
is only needed as some parsers complain if you do 
```
writer.writeStartElement(wrapper);
writer.writeEndElement();
```
even if semantically the do exactly the same thing.

ctomc:
retest this please

bstansberry:
I put the Hold label on this as I'm looking into making some tweaks.

bstansberry:
Replaced by #1477.

PR #1472, opened 2016-03-23
[WFCORE-1450] Use a copy constructor instead of a subclass

bstansberry:
Also includes a commit converting some classes to nested classes of the main class in the file.


bstansberry:
retest this please

ctomc:
@bstansberry while you are at it, I think we should @deprecate this class all together.
It is mostly used by tests and WS subsystem, which can be easily converted to PersistentRD
wdyt?

bstansberry:
I kind of like the class though. :) I'll think about it.

bstansberry:
This is used in 4 subsystems in full, so I'm not going to deprecate it here.

PR #1473, opened 2016-03-23
[WFCORE-701] Apply changes to get a consistence between the HC's view state and server's own view state when the server process is stopped

yersan:
The status reported by the HC's view of server management lifecycle is different from the status reported by the server's itself. This commit tries to avoid that discrepancy if the server was stopped by an action, which wasn't initialized by the HC. In that case instead of report FAILED from the HC's view, now STOPPED is reported if the server was successful stopped.

Jira issue:
https://issues.jboss.org/browse/WFCORE-701


luck3y:
retest this please

bstansberry:
Sorry it took a while to look at this. These transitions are tricky so it took a while to find a block of time to look carefully.

This looks good; thanks!

PR #1475, opened 2016-03-29
[WFCORE-1106] Don't execute Stage.RUNTIME steps when capabilities associated with the address are reload-required

bstansberry:
See https://issues.jboss.org/browse/WFCORE-1106 description for the conceptual details.

This requires merge of https://github.com/wildfly/wildfly/pull/8806 before CI will pass.

I can squash the seven commits before merging; right now they are separate commits to perhaps help reviewers understand the change.

On WFCORE-1106 I commented about how perhaps we should reject "reload" ops when the server is in restart-required. I haven't done anything about that here.

http://brontes.lab.eng.brq.redhat.com/viewLog.html?buildId=95091&tab=buildResultsDiv&buildTypeId=WF_WildFlyCoreIntegrationExperiments shows the successful result of a CI run of this PR plus the required downstream PR.


bstansberry:
Replaced by #1566 which doesn't have so much CI noise to scroll through.

PR #1477, opened 2016-03-30
[WFCORE-1448] Parsing and marshalling for map and list attributes don't differentiate between empty and absent object lists

bstansberry:
These are some changes to #1470 that I want. He's AFK right now so I've put up a separate PR.


ctomc:
:+1:

PR #1480, opened 2016-03-31
[WFCORE-1458] remove extra realm name block check which changes add-user.sh scripts layout after adding a new user.

soul2zimate:
https://issues.jboss.org/browse/WFCORE-1458

Method UserPropertiesFileLoader.addLineContent was added in https://github.com/wildfly/wildfly/pull/4878
This method attempts to remove realm name and rewrite no matter if -r or --realm presents while adding a new user with add-user.sh. This causes problem in jira above. However, you can not define any new realm name if it's not first in properties file due to following exception from DomainManagementLogger.userRealmNotMatchDiscovered

> AddUserFailedException: WFLYDM0065: The user supplied realm name 'AnotherManagementRealm' does not match the realm name discovered from the property file(s) 'ManagementRealm'.

### DIFF
--- a/domain-management/src/main/java/org/jboss/as/domain/management/security/UserPropertiesFileLoader.java
+++ b/domain-management/src/main/java/org/jboss/as/domain/management/security/UserPropertiesFileLoader.java
@@ -155,36 +155,6 @@ public class UserPropertiesFileLoader extends PropertiesFileLoader {
         }
     }
 
-    /**
-     * Remove the realm name block.
-     *
-     * @see PropertiesFileLoader#addLineContent(java.io.BufferedReader, java.util.List, String)
-     */
-    @Override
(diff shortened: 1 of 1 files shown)


darranl:
I am not completely following the comment above, if a realm name is specified in the configuration file it must never be possible to override this with the add-user utility.  

The only time the add user utility should be supporting providing a realm name is if no realm name is already discovered in the file.

soul2zimate:
Thanks, I was not aware that we can provide a realm name if no realm name is already discovered in the file.
With this change, the behavior is like if there is a realm name 

`"#$REALM_NAME=ManagementRealm$ This line is used by the add-user utility to identify the realm name already used in this file."`

in application-users.properties or mgmt-users.properties, adding a new user won't remove realm name and add it back to the end of file.

If there is no realm name provided, it adds realm name at the end of file.

PR #1482, opened 2016-04-01
[WFCORE-1460] Ensure tests don't leave a running server in reload-required state

bstansberry:
These changes are part of #1475 since they are required to make the main code change there pass CI, but I decided to submit them separately as well. These are useful on their own and I figure #1475 is a big chunk to review, while these should be more digestible. If they get merged, then #1475 will become more digestible too.

The commits do 3 things:

1) Change some tests such that they don't trigger reload-required.
2) Add a SetupTask that triggers reload in tearDown, and then use it in tests that really need to trigger reload required
3) Change the test runner such that it fails in non-manual control tests that leave the server in reload-required.

### DIFF
--- a/controller/src/main/java/org/jboss/as/controller/access/constraint/ApplicationTypeConstraint.java
+++ b/controller/src/main/java/org/jboss/as/controller/access/constraint/ApplicationTypeConstraint.java
@@ -122,17 +122,31 @@ public class ApplicationTypeConstraint extends AllowAllowNotConstraint {
             return false;
         }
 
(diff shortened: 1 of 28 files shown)


kabir:
Retest this pleae

kabir:
Retest this please

bstansberry:
Looks like there's another test issue to fix.

bstansberry:
@kabir I added a couple commits. The "Track the correct SensitivityClassification/ApplicationTypeConfig instance" one should fix the test failure.

PR #1483, opened 2016-04-04
cli, ssl config issue with vault in wildfly-core context

jfdenise:
This is a possible fix for https://issues.jboss.org/browse/WFCORE-1462

### DIFF
--- a/testsuite/standalone/src/test/java/org/jboss/as/test/integration/management/cli/JBossCliXmlConfigTestCase.java
+++ b/testsuite/standalone/src/test/java/org/jboss/as/test/integration/management/cli/JBossCliXmlConfigTestCase.java
@@ -1,0 +1,119 @@
+/*
+ * JBoss, Home of Professional Open Source.
+ * Copyright 2016, Red Hat, Inc., and individual contributors
+ * as indicated by the @author tags. See the copyright.txt file in the
+ * distribution for a full listing of individual contributors.
+ *
+ * This is free software; you can redistribute it and/or modify it
+ * under the terms of the GNU Lesser General Public License as
+ * published by the Free Software Foundation; either version 2.1 of
+ * the License, or (at your option) any later version.
+ *
+ * This software is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the GNU
+ * Lesser General Public License for more details.
+ *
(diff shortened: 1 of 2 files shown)


aloubyansky:
Looks good to me. Thanks,  Jean-Francois!

bstansberry:
this is ok to test

testsuite/standalone/src/test/java/org/jboss/as/test/integration/management/cli/JBossCliXmlConfigTestCase.java
```diff
@@ -0,0 +1,108 @@
...
+        final CliCommandBuilder commandBuilder = CliCommandBuilder.of(jbossDist);
+        commandBuilder.addJavaOptions("-Djboss.cli.config=" + f.toPath());
+
+        commandBuilder.addCliArgument("--command=help");
+        Process cliProc = null;
+        try {
+            cliProc = Launcher.of(commandBuilder)
+                    .setRedirectErrorStream(true)
+                    .launch();
+            cliProc.waitFor(2000, TimeUnit.MILLISECONDS);
```
bstansberry: We have a util that we run all testuite timeouts through to make it easier to make them longer in odd environments. So just as a code standard thing this should be:

cliProc.waitFor(TimeoutUtil.adjust(2000), TimeUnit.MILLISECONDS);

org.jboss.as.test.shared.TimeoutUtil

Also, is 2000ms the appropriate value here? I think the test failure that was reported on the CI run is due to that amount being insufficient.

If the goal of the test isn't to check for something happening quickly, then use a real long timeout (like 30 secs). If the test fails but takes a long time to do so, that's ok; that should be a rare situation. OTOH intermittent CI failures due to short timeouts being exceeded are a real pain.

jfdenise:
I have enlarged and adjusted the timeout.

bstansberry:
Don't worry about the PassivationTestCase failure; that's an annoying intermittent problem. 

It looks like it's still failing on windows though, despite the increased timeout.

jfdenise:
I reproduced the issue on a Windows lab machine. Help command dumps enough content that it makes the Process to hang if out and err streams are not read.

aloubyansky:
Then perhaps use 'version' command instead?

aloubyansky:
retest this please!

kabir:
I think it looks good now, but can you please squash the commits into one?

bstansberry:
When you squash, please also change the commit message to start with [WFCORE-1462]. Thanks!

jfdenise:
Ok, commits squashed

PR #1484, opened 2016-04-04
WFCORE-43 runtime-configuration-state introduction and mapping to server-state / host-state for compatibility.

luck3y:
This is a somewhat trimmed down version of the previous, being more careful to not use the new attribute anywhere we might interact directly with an older version that doesn't have it (example jboss-cli.sh & old embedded impls).

Tested against wf-head, wf-10-final and eap-6.4

Updated to not change the ControlledProcessState public enum, and just map the new attribute in ProcessStateAttributeHandler.


luck3y:
retest this please

luck3y:
retest this please

host-controller/src/main/resources/org/jboss/as/host/controller/descriptions/LocalDescriptions.properties
```diff
@@ -155,6 +156,7 @@ server-config.ssl.loopback.truststore-password=The password to open the trust st
 server=The managed server instance.
 server.launch-type=The manner in which the server process was launched. Either "DOMAIN" for a domain mode server launched by a Host Controller, "STANDALONE" for a standalone server launched from the command line, or "EMBEDDED" for a standalone server launched as an embedded part of an application running in the same virtual machine.
 server.server-state=The current state of the server controller
+server.runtime-configuration-state=The current state of the server controller; either STARTING, RUNNING, RESTART_REQUIRED, RELOAD_REQUIRED or STOPPING.
```
kabir: Should this be the same states as for the host variety?

testsuite/domain/src/test/java/org/jboss/as/test/integration/domain/slavereconnect/DeploymentOverlayScenario.java
```diff
@@ -196,11 +196,17 @@ void testAfterReconnect(DomainClient masterClient, DomainClient slaveClient) thr
...
+                DomainTestUtils.executeForResult(
+                        Util.getReadAttributeOperation(SLAVE_ADDR.append(UnaffectedScenario.SERVER), "runtime-configuration-state"), slaveClient).asString());
+        Assert.assertEquals(/* "running" */ "reload-required",
                 DomainTestUtils.executeForResult(
                         Util.getReadAttributeOperation(SLAVE_ADDR.append(UnaffectedScenario.SERVER), "server-state"), slaveClient).asString());
         Assert.assertEquals("reload-required",
                 DomainTestUtils.executeForResult(
+                        Util.getReadAttributeOperation(SLAVE_ADDR.append(SERVER, "server-affected"), "runtime-configuration-state"), slaveClient).asString());
+        Assert.assertEquals("reload-required",
+                DomainTestUtils.executeForResult(
```
kabir: We can probably get rid of this and the other 'server-state' checks for the slavereconnect scenarios, and just go with the new runtime-configuration-state ones?

testsuite/domain/src/test/java/org/jboss/as/test/integration/domain/slavereconnect/SlaveReconnectTestCase.java
```diff
@@ -198,9 +201,12 @@ private boolean checkHostServersStarted(DomainClient masterClient, String host)
                 String server = entry.asString();
                 op = Util.createEmptyOperation(READ_ATTRIBUTE_OPERATION, PathAddress.pathAddress(HOST, host).append(SERVER, server));
                 op.get(NAME).set("server-state");
-                DomainTestUtils.executeForResult(op, masterClient); // success is enaught
+                ModelNode state = DomainTestUtils.executeForResult(op, masterClient);
+                if (SUCCESS.equals(state.get(OUTCOME).asString())) {
+                    return "running".equals(state.get(RESULT).asString());
```
kabir: I think the new attribute/state name instead?

luck3y: I'd changed a bunch of these in the previous, and was getting a lot of failures with the wildfly mixed  domain tests, I'll look at these again, maybe they're safe.

kabir: These particular tests would not be doing mixed domain stuff

controller/src/main/java/org/jboss/as/controller/ControlledProcessState.java
```diff
@@ -38,10 +38,18 @@
...
+         * @see {@link org.jboss.as.controller.operations.common.ProcessStateAttributeHandler}
+         * for how the new values are mapped to the old attributes for compatibility.
          * The process is started, is running normally and has a runtime state consistent with its persistent configuration.
          */
         RUNNING("running"),
         /**
+         * replaces #RUNNING
+         * The process is started, is operating normally and has a runtime state consistent with its persistent configuration.
+         */
+        OK("ok"),
```
bstansberry: This is a problem. ControlledProcessStateService exposes this enum, and it may be used by subsystem to check for state transitions. ControlledProcessStateService allows for registration of property change listeners for that purpose. That stuff is marked deprecated, but I think we can only break it in an EAP major.

I know I've pointed out that API a few times to folks who are looking for some way to track the server lifecycle. They may break if a new OK value is introduced.

Sorry; I should have mentioned this before. :( I was thinking in terms of the external API, not the internals.

luck3y: Ok, thats good to know. What about removing that from the enum and just mapping runtime-configuration-state in the attribute handler, but leaving everything else the same? That would allow us to introduce the new attribute, but not break anything until the next major.

bstansberry: That's fine.

PR #1485, opened 2016-04-05
Fix HSM-based SSL

ramsperger:
Using a KeyStore provider that does not allow or returns empty from getEncoded() on private keys causes a KeyStoreException at SSL startup. This is common in HSM-backed key operations.

When creating a reduced KeyStore, use the same provider as the source KeyStore to
minimize incompatibilities.

While it's still not guaranteed to work with all KeyStore providers, switching `KeyStore.getInstance("JKS")` to `KeyStore.getInstance(provider)` fixes the issue for SafeNet "Luna" and SunPKCS11 "PKCS11" KeyStore implementations while not breaking the "PKCS12" and "JKS" cases.

### DIFF
--- a/domain-management/src/main/java/org/jboss/as/domain/management/security/FileKeystore.java
+++ b/domain-management/src/main/java/org/jboss/as/domain/management/security/FileKeystore.java
@@ -123,7 +123,7 @@ final class FileKeystore {
             if (alias == null) {
                 this.setKeyStore(loadedKeystore);
             } else {
-                KeyStore newKeystore = KeyStore.getInstance("JKS");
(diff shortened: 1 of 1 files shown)


kabir:
Thanks for the pull request, a few questions:
Is there a Jira issue for this? (if not we can create one)
@darranl does this clash/become redundant with the Elytron integration work?

Finally, before we can accept any third-party contributions we have some paper work to do.

ramsperger:
JIRA issue is https://issues.jboss.org/browse/WFCORE-1466

I'm happy with the MIT license.

darranl:
This code does become redundant but I approve it's inclusion for now.

PR #1486, opened 2016-04-05
[WFCORE-1464] ensure LogContexts for sub deployments are unregistered

aogburn:
https://issues.jboss.org/browse/WFCORE-1464

### DIFF
--- a/logging/src/main/java/org/jboss/as/logging/deployments/AbstractLoggingDeploymentProcessor.java
+++ b/logging/src/main/java/org/jboss/as/logging/deployments/AbstractLoggingDeploymentProcessor.java
@@ -72,7 +72,7 @@ abstract class AbstractLoggingDeploymentProcessor implements DeploymentUnitProce
     @Override
     public final void undeploy(final DeploymentUnit context) {
         // OSGi bundles deployments may not have a module attached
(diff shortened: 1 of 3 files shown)


logging/src/main/java/org/jboss/as/logging/deployments/AbstractLoggingDeploymentProcessor.java
```diff
@@ -72,18 +72,21 @@ public final void deploy(final DeploymentPhaseContext phaseContext) throws Deplo
...
         // OSGi bundles deployments may not have a module attached
-        if (hasRegisteredLogContext(context) && context.hasAttachment(Attachments.MODULE)) {
+        if (context.hasAttachment(Attachments.MODULE)) {
             // don't process sub-deployments as they are processed by processing methods
             final ResourceRoot root = context.getAttachment(Attachments.DEPLOYMENT_ROOT);
             if (SubDeploymentMarker.isSubDeployment(root)) return;
             // Remove any log context selector references
-            final Module module = context.getAttachment(Attachments.MODULE);
-            unregisterLogContext(context, module);
+            if (hasRegisteredLogContext(context)) {
```
jamezp: I think we should just move the `hasRegisteredLogContext()` to the `unregisterLogContext()`.

PR #1487, opened 2016-04-05
[WFCORE-1464] ensure LogContexts for sub deployments are unregistered

aogburn:
https://issues.jboss.org/browse/WFCORE-1464


bstansberry:
Is this the same change as the one James already marked Ready for Merge? I had tested and was about to merge that one when I saw it was closed. Looked like you closed it because the branch got messed up or something. If this is the same I'll just merge it.

aogburn:
Yes, this is the same.

PR #1489, opened 2016-04-06
WFCORE-1467 This commit avoids call two times a suspend operation when suspend-servers is used with a timeout parameter different than 0

yersan:
In a domain mode :suspend-servers command with a timeout parameter different than 0 executes two suspend operations for each server in the domain. It first executes a suspend with 0ms and later a suspend with the timeout specified.

Jira issue:
https://issues.jboss.org/browse/WFCORE-1467

### DIFF
--- a/host-controller/src/main/java/org/jboss/as/domain/controller/operations/DomainServerLifecycleHandlers.java
+++ b/host-controller/src/main/java/org/jboss/as/domain/controller/operations/DomainServerLifecycleHandlers.java
@@ -346,16 +346,19 @@ public class DomainServerLifecycleHandlers {
                     context.getServiceRegistry(true);
                     Map<String, ProcessInfo> processes = serverInventory.determineRunningProcesses(true);
                     final Set<String> serversInGroup = getServersForGroup(model, group);
-                    final Set<String> waitForServers = new HashSet<String>();
+                    final Set<String> waitForServers = new HashSet<>();
                     for (String serverName : processes.keySet()) {
                         final String serverModelName = serverInventory.getProcessServerName(serverName);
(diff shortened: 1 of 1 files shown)


bstansberry:
retest this please

Failure was due to unrelated issue WFCORE-1270.

bstansberry:
I think a deeper change is needed. ServerInventoryImpl.awaitServerSuspend is using the timeout param as the overall timeout for the set of servers. It then loops and uses the *remaining* time as the time param to ManagedServer.awaitSuspended. If the overall timeout is used up, for some servers awaitSuspended won't even be invoked.

Also, ManagedServer.awaitSuspended is taking its timeout param, which is in ms, and sending it to the server in the management op's timeout param, which is in seconds. This is why the logging on the server is "10000000ms" when it should be "10000ms".

I think the way this needs to work is:

ManagedServer provides a method with this signature:

BlockingQueueOperationListener<TransactionalProtocolClient.Operation> suspend(int timeoutInSeconds)

It just sends the op to the server and returns the BlockingQueueOperationListener. At this point the requests have been sent to each server, but responses have not been sent. 

ServerInventoryImpl.awaitServerSuspend loops through the servers calling that and caches the results in a list. From this point it starts tracking elapsed time and iterates through the cached BlockingQueueOperationListeners, calling retrievePreparedOperation(final long timeout, final TimeUnit timeUnit) with whatever time is left as the 'timeout' param. It also does the commit etc of the prepared operation, similar to what ManagedServer.suspend and awaitSuspended do.

Also, org.jboss.as.host.controller.operations.ServerSuspendHandler has the same problem that you're fixing in SuspendServersLifecycleHandler.

kabir:
I am closing this one, since I am attempting what you suggested in #1490

PR #1490, opened 2016-04-07
WFCORE 1467 Avoids calling suspend twice when a timeout is used with suspend.

kabir:
Attempt at following Brian's suggestions on https://github.com/wildfly/wildfly-core/pull/1489.

I wasn't familiar with BlockingQueueOperationListener, but hopefully if something needs tweaking this will help.

https://issues.jboss.org/browse/WFCORE-1467

### DIFF
--- a/host-controller/src/main/java/org/jboss/as/host/controller/ServerInventoryImpl.java
+++ b/host-controller/src/main/java/org/jboss/as/host/controller/ServerInventoryImpl.java
@@ -32,6 +32,7 @@ import java.security.SecureRandom;
 import java.util.Base64;
 import java.util.Collection;
 import java.util.HashMap;
+import java.util.HashSet;
 import java.util.LinkedList;
 import java.util.List;
 import java.util.Locale;
@@ -41,6 +42,7 @@ import java.util.Set;
 import java.util.concurrent.ConcurrentHashMap;
 import java.util.concurrent.ConcurrentMap;
 import java.util.concurrent.CountDownLatch;
+import java.util.concurrent.ExecutionException;
 import java.util.concurrent.TimeUnit;
 
 import javax.security.auth.callback.Callback;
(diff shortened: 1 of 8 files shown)


host-controller/src/main/java/org/jboss/as/host/controller/ServerInventoryImpl.java
```diff
@@ -399,26 +402,39 @@ public void resumeServer(String serverName) {
...
+                } catch (IOException ignore) {
                 }
             }
         }
-        long time = end - System.currentTimeMillis();
-        return time > 0;
+
+        for (BlockingQueueOperationListener<?> listener : listeners) {
+            try {
+                final TransactionalProtocolClient.PreparedOperation<?> prepared = listener.retrievePreparedOperation();
```
bstansberry: The following shouldn't prevent merging this; it's meant for consideration in future work.

I'm concerned that both here and the get() at line 428 are blocking calls with no timeout. 

This retrievePreparedOperation call *could* perhaps use the timeoutInSeconds param to somehow set a limit. (There is an overloaded retrievePreparedOperation() variant that takes a timeout.) That would need care though, as the "timeoutInSeconds" the user provides is the time the server is allowed to take to suspend, not the time it can take to suspend *plus* get a prepared response back to the HC. That "plus" is an unknown amount of time.

So, maybe for limiting the blocking time at L423 it's better to use the greater of timeoutInSeconds + a fudge factor, or the general blocking timeout used internally for safety by the OperationContext. (By default that's 300 seconds). IIRC that internal timeout isn't exposed, though.

Ideally the get() call at L428 would somehow tie in with blocking timeout stuff that we use internally in the OperationContext to prevent blocking forever. By the time the call gets to L428, the server side op is prepared, which means the server is suspended. So any delay in getting a response to get() means there was some problem finishing the server side op or transmitting/receiving the final response. So dealing with that is more of a case of being robust about inter-process operation handling, and not about "suspend" itself. So the general OperationContext blocking timeout is more relevant, as opposed to any user provided param like timeoutInSeconds.

The other question is what to do if these timeouts trip. Generally it would be to trigger cancellation of the remote op and then fail the management request.

Domain coordination handlers like DomainSlaveHandler and DomainRolloutStepHandler have a lot of logic along these lines.

n1hility:
retest this please

PR #1491, opened 2016-04-07
[WFCORE-890] Split http-upgrade for the HTTP Managenent interface into it's own element.

darranl:

This change is a minor update to the HTTP management interface configuration in preparation for subsequent Elytron changes that will be coming in the future.  Also unfortunately I implemented some changes in the wrong order - following on from this PR I also have another PR to refactor the management interfaces so they can share as much common code as possible as at the moment there is a lot of duplication.

The reason for this change is that Elytron provides authentication policies for SASL based authentication and for HTTP based authentication - once HTTP Upgrade occurs we switch to Remoting to handle the authentication which means we are switching to SASL so this new child element is to logically group the configuration specific to HTTP Upgrade.


server/src/main/resources/schema/wildfly-config_5_0.xsd
```diff
@@ -2023,7 +2023,31 @@
             </xs:annotation>
         </xs:attribute>
     </xs:complexType>
-    
+
+    <xs:complexType name="http-management-interfaceType">
+        <xs:complexContent>
+            <xs:extension base="management-interfaceType">
+                <xs:sequence>
+                    <xs:element name="http-upgrade" minOccurs="0">
```
ehsavoie: maybe we should also set the maxOccurs to 1 ?

bstansberry:
Looks good but needs a rebase. Ping me on hipchat once it's rebased and I'll try and get it merged immediately.

bstansberry:
I just pushed 3 PRs to master-ignore and plan to merge them once the testsuite finishes (say 4:30 EDT) so any rebase should probably wait for that.

bstansberry:
Those 3 PRs are merged.

PR #1494, opened 2016-04-12
WFCORE-1469 - Support match-comparison operation for if-command

thomasdarimont:
Introduced regex based match comparison-operation for if-commands.
The match operation is denoted by the symbol `~=`.
The left operand is a string to match against the right operand
which is a regex string.

This enables a simple way to do feature flag detection
via partial regex matches.

E.g. given the system property "features" which contains a space
separated list of enabled "features".
```
-Dfeatures="activemq jgroups"
```

One could conditionally configure wildfly components via CLI
in one generic CLI script, e.g.:
```
if (features ~= ".*activemq.*") of /:resolve-expression(expression=${features})
echo configuring activemq
end-if

if (features ~= ".*jgroups.*") of /:resolve-expression(expression=${features})
echo configuring jgroups
end-if

if (features ~= ".*postgres.*") of /:resolve-expression(expression=${features})
echo configuring postgres
end-if
```

### DIFF
--- a/testsuite/standalone/src/test/java/org/jboss/as/test/integration/management/cli/ifelse/BasicIfElseTestCase.java
+++ b/testsuite/standalone/src/test/java/org/jboss/as/test/integration/management/cli/ifelse/BasicIfElseTestCase.java
@@ -51,6 +51,24 @@ public class BasicIfElseTestCase extends CLISystemPropertyTestBase {
             cliOut.reset();
         }
     }
+
+    @Test
+    public void testIfMatchComparison() throws Exception {
+
+        final CommandContext ctx = CLITestUtil.getCommandContext(cliOut);
+        try {
+            ctx.connectController();
+            ctx.handle(this.getAddPropertyReq("match-test-values", "\"AAA BBB\""));
+            assertEquals("true", runIfWithMatchComparison("match-test-values", "AAA", ctx));
+            assertEquals("true", runIfWithMatchComparison("match-test-values", "BBB", ctx));
+            assertEquals("false", runIfWithMatchComparison("match-test-values", "CCC", ctx));
+        } finally {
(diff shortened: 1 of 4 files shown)


aloubyansky:
Looks good to me. Thanks! Could you please also add a simple test to the testsuite too? Something like org.jboss.as.test.integration.management.cli.ifelse.BasicIfElseTestCase? Just have it there in action. Thanks again!

thomasdarimont:
I ammended the commit by adding the mentioned test case.

testsuite/standalone/src/test/java/org/jboss/as/test/integration/management/cli/ifelse/BasicIfElseTestCase.java
```diff
@@ -51,14 +51,48 @@ public void testMain() throws Exception {
...
+            ctx.terminateSession();
+            cliOut.reset();
+        }
+    }
+
     protected String runIf(CommandContext ctx) throws Exception {
         ctx.handle("if result.value==\"true\" of " + this.getReadPropertyReq());
-        ctx.handle(this.getWritePropertyReq("\"false\""));
-        ctx.handle("else");
         ctx.handle(this.getWritePropertyReq("\"true\""));
```
aloubyansky: What is the reason behind this change?

thomasdarimont:
Updated the PR again since I had edits in an unrelated test case.

aloubyansky:
Thanks a lot! This commit looks good to me.

kabir:
Retest this please

kabir:
Thanks for the patch, before it can be merged (assuming you are not a Red Hat employee), we need to do a bit of paperwork.

thomasdarimont:
Hello @kabir 

yes, I agree to these terms.

Cheers,
Thomas

PR #1495, opened 2016-04-13
[WFCORE-1456] Allow CLI class to support offline mode

jfdenise:
This fix allows to use the class in offline mode and connected mode. Furthermore these changes cover the issue reported in WFCORE-1417 ( CLI was not usable after a failing connect and error was wrongly reported).
CLI class has been refactored in order to reduce code replication and remove some formatting issues.

The existing API and expected behaviour has been preserved:
CLI cli = CLI.newInstance();
cli.connect(....);

The new behaviour:
CLI cli = CLI.newInstance();
cli.cmd("embed-server");
cli.connect(...);
cli.disconnect();
cli.cmd("embed-server"); || cli.connect(...);
cli.disconnect();
...

### DIFF
--- a/testsuite/standalone/src/test/java/org/jboss/as/test/integration/management/cli/CLITestCase.java
+++ b/testsuite/standalone/src/test/java/org/jboss/as/test/integration/management/cli/CLITestCase.java
@@ -1,0 +1,145 @@
+/*
+ * JBoss, Home of Professional Open Source.
+ * Copyright 2016, Red Hat Middleware LLC, and individual contributors
+ * as indicated by the @author tags. See the copyright.txt file in the
+ * distribution for a full listing of individual contributors.
+ *
+ * This is free software; you can redistribute it and/or modify it
+ * under the terms of the GNU Lesser General Public License as
+ * published by the Free Software Foundation; either version 2.1 of
+ * the License, or (at your option) any later version.
+ *
+ * This software is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the GNU
+ * Lesser General Public License for more details.
+ *
+ * You should have received a copy of the GNU Lesser General Public
+ * License along with this software; if not, write to the Free
+ * Software Foundation, Inc., 51 Franklin St, Fifth Floor, Boston, MA
+ * 02110-1301 USA, or see the FSF site: http://www.fsf.org.
+ *
+ */
+package org.jboss.as.test.integration.management.cli;
+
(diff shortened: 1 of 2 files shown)


aloubyansky:
Looks good. Thanks Jeff!

testsuite/standalone/src/test/java/org/jboss/as/test/integration/management/cli/CLITestCase.java
```diff
@@ -0,0 +1,145 @@
...
+        executeCommand(cli, "embed-server --std-out=echo");
+
+        if (cli.getCommandContext().getControllerPort() > 0) {
+            Assert.fail("Invalid port for embedded "
+                    + cli.getCommandContext().getControllerPort());
+        }
+
+        // Enable management
+        executeCommand(cli, "reload --admin-only=false");
+
```
luck3y: @jfdenise Sorry for coming in late on this. I think you may need to add a method to poll for the reload to be actually completed here, as it seems to be racing against itself. If I run in IDEA, the test always seems to pass, but running with mvn on the command line, it always seems to fail:

$ testsuite/standalone/
$ mvn test -Dtest=org.jboss.as.test.integration.management.cli.CLITestCase

What appears to be happening is that its trying to connect before the reload has completed.

If you take a look in ReloadHandler at the ensureServerRebootCompleted method, you'll see how that polls. We've recently added a new attribute runtime-configuration-state that should have the value "ok" when the reload is complete, which you can probably use here.

Some quick code if it helps: https://gist.github.com/luck3y/d563c3820c3e22c116327f4c5737bd58.

On a side note, I've often wondered about the immediate return of reload, it would seem more friendly if there was a way to have it block (perhaps as an optional parameter) until it was actually complete.

aloubyansky: We are aware of the issue and @jfdenise is going to follow up with a fix. Apparently, instead of the reload command, what gets executed is :reload operation, so the logic for making sure the controller is up in the reload handler is not run.
Sorry about that. And thanks for the tips!

luck3y: Great, thanks @aloubyansky and @jfdenise! Also, that makes a lot of sense now :)

jfdenise: Yes, this test helped discover the CLI class behaviour, I am working on a patch to make CLI class to behave properly.

jfdenise: The bug that is tracking the CLI command/operation mix: https://issues.jboss.org/browse/WFCORE-1471

PR #1501, opened 2016-04-15
[WFCORE-1456] [WFCORE-1471] CLI class to handle offline mode and commands

jfdenise:
[WFCORE-1456]
Approach is to make this change compliant with existing CLI class behaviour.
The class can now handle offline mode commands.

 [WFCORE-1471]
The test added for WFCORE-1456 has shown that the CLI class was mixing operations and commands. Furthermore Batch and workflow features were not supported.
Although all could be routed to CommandContext.handle, this fix does its best to allow CLI.cmd to return a response when doable.

### DIFF
--- a/cli/src/main/java/org/jboss/as/cli/CommandContext.java
+++ b/cli/src/main/java/org/jboss/as/cli/CommandContext.java
@@ -1,6 +1,6 @@
 /*
  * JBoss, Home of Professional Open Source.
- * Copyright 2015, Red Hat, Inc., and individual contributors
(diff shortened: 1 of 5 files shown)


jfdenise:
retest this please!

luck3y:
The passivation test case s a flaky test, will probably be fine if you retest.

jfdenise:
Ok, thanks, I just squashed commits, tests are starting.

aloubyansky:
The changes look good to me to merge. Thanks!

PR #1504, opened 2016-04-17
[WFCORE-1473] Improve inter-process handling for suspend servers operation using a blocking timeout

yersan:
This issue has been created from @bstansberry comments on https://github.com/wildfly/wildfly-core/pull/1490. 
It adds a blocking timout for the calls done by the HC when :suspend-servers operation is executed.

Jira issue is:
https://issues.jboss.org/browse/WFCORE-1473

### DIFF
--- a/host-controller/src/main/java/org/jboss/as/host/controller/DomainModelControllerService.java
+++ b/host-controller/src/main/java/org/jboss/as/host/controller/DomainModelControllerService.java
@@ -1179,8 +1179,8 @@ public class DomainModelControllerService extends AbstractControllerService impl
         }
 
         @Override
-        public boolean awaitServerSuspend(Set<String> waitForServers, int timeout) {
-            return getServerInventory().awaitServerSuspend(waitForServers, timeout);
+        public List<ModelNode> awaitServerSuspend(Set<String> waitForServers, int timeout, BlockingTimeout blockingTimeout) {
+            return getServerInventory().awaitServerSuspend(waitForServers, timeout, blockingTimeout);
         }
     }
 
@@ -1512,8 +1512,8 @@ public class DomainModelControllerService extends AbstractControllerService impl
             }
 
             @Override
-            public boolean awaitServerSuspend(Set<String> waitForServers, int timeout) {
-                return false;
+            public List<ModelNode> awaitServerSuspend(Set<String> waitForServers, int timeout, BlockingTimeout blockingTimeout) {
+                return Collections.emptyList();
             }
         };
         future.setInventory(inventory);
--- a/host-controller/src/main/java/org/jboss/as/host/controller/ManagedServer.java
+++ b/host-controller/src/main/java/org/jboss/as/host/controller/ManagedServer.java
@@ -46,6 +46,7 @@ import org.jboss.as.controller.ProxyOperationAddressTranslator;
 import org.jboss.as.controller.TransformingProxyController;
 import org.jboss.as.controller.client.OperationAttachments;
 import org.jboss.as.controller.client.OperationMessageHandler;
+import org.jboss.as.controller.client.OperationResponse;
 import org.jboss.as.controller.client.helpers.domain.ServerStatus;
 import org.jboss.as.controller.remote.BlockingQueueOperationListener;
 import org.jboss.as.controller.remote.TransactionalProtocolClient;
@@ -64,6 +65,7 @@ import org.jboss.marshalling.Marshalling;
 import org.jboss.marshalling.MarshallingConfiguration;
 import org.jboss.marshalling.SimpleClassResolver;
 import org.jboss.msc.service.ServiceActivator;
+import org.jboss.threads.AsyncFuture;
 
 /**
  * Represents a managed server.
@@ -125,6 +127,8 @@ class ManagedServer {
     private volatile int operationID = CurrentOperationIdHolder.getCurrentOperationID();
     private volatile ManagedServerBootConfiguration bootConfiguration;
 
+    private final PathAddress address;
+
     ManagedServer(final String hostControllerName, final String serverName, final String authKey,
                   final ProcessControllerClient processControllerClient, final URI managementURI,
                   final TransformationTarget transformationTarget) {
@@ -144,7 +148,7 @@ class ManagedServer {
 
         // Setup the proxy controller
         final PathElement serverPath = PathElement.pathElement(RUNNING_SERVER, serverName);
-        final PathAddress address = PathAddress.EMPTY_ADDRESS.append(PathElement.pathElement(HOST, hostControllerName), serverPath);
+        address = PathAddress.EMPTY_ADDRESS.append(PathElement.pathElement(HOST, hostControllerName), serverPath);
         this.protocolClient = new ManagedServerProxy(this);
         this.proxyController = TransformingProxyController.Factory.create(protocolClient,
                 Transformers.Factory.create(transformationTarget), address, ProxyOperationAddressTranslator.SERVER);
@@ -718,15 +722,13 @@ class ManagedServer {
         return true;
     }
 
-    BlockingQueueOperationListener<?> suspend(int timeoutInSeconds) throws IOException {
+    AsyncFuture<OperationResponse> suspend(int timeoutInSeconds, final BlockingQueueOperationListener<TransactionalProtocolClient.Operation> listener) throws IOException {
         final ModelNode operation = new ModelNode();
         operation.get(OP).set(SUSPEND);
         operation.get(OP_ADDR).setEmptyList();
         operation.get(TIMEOUT).set(timeoutInSeconds);
 
-        BlockingQueueOperationListener<TransactionalProtocolClient.Operation> listener = new BlockingQueueOperationListener<>();
-        protocolClient.execute(listener, operation, OperationMessageHandler.DISCARD, OperationAttachments.EMPTY);
-        return listener;
+        return protocolClient.execute(listener, operation, OperationMessageHandler.DISCARD, OperationAttachments.EMPTY);
     }
 
     static enum InternalState {
@@ -938,6 +940,10 @@ class ManagedServer {
             }
         }
         return result;
+    }
+
+    PathAddress getAddress(){
+        return address;
     }
 
 }
--- a/host-controller/src/main/java/org/jboss/as/host/controller/ServerInventory.java
+++ b/host-controller/src/main/java/org/jboss/as/host/controller/ServerInventory.java
@@ -23,11 +23,13 @@
 package org.jboss.as.host.controller;
 
 import java.util.Collection;
+import java.util.List;
 import java.util.Map;
 import java.util.Set;
 
 import javax.security.auth.callback.CallbackHandler;
 
+import org.jboss.as.controller.BlockingTimeout;
 import org.jboss.as.controller.ProxyController;
 import org.jboss.as.controller.client.helpers.domain.ServerStatus;
 import org.jboss.as.process.ProcessInfo;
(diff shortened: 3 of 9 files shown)


host-controller/src/main/java/org/jboss/as/host/controller/ServerInventoryImpl.java
```diff
@@ -401,37 +405,63 @@ public void resumeServer(String serverName) {
...
 
-        for (Map.Entry<String, BlockingQueueOperationListener<?>> listenerEntry : listeners.entrySet()) {
+        for (Map.Entry<String, OperationData> operationDataEntry : operationDataMap.entrySet()) {
+            final OperationData operationData = operationDataEntry.getValue();
             try {
                 final TransactionalProtocolClient.PreparedOperation<?> prepared =
-                        listenerEntry.getValue().retrievePreparedOperation();
+                        operationData.listener.retrievePreparedOperation(operationData.blockingTimeout, TimeUnit.MILLISECONDS);
+                if ( prepared == null ){
+                    throw new TimeoutException(String.format(
```
ehsavoie: You should use your  HostControllerLogger.timedOutAwaitingPreparedResponse

host-controller/src/main/java/org/jboss/as/host/controller/ManagedServer.java
```diff
@@ -940,4 +942,8 @@ public boolean execute(ManagedServer server) throws Exception {
         return result;
     }
 
+    protected PathAddress getAddress(){
```
bstansberry: I think this can be package protected.

host-controller/src/main/java/org/jboss/as/host/controller/ServerInventory.java
```diff
@@ -320,6 +321,6 @@
      * @param timeoutInSeconds The maximum amount of time to wait in seconds, with -1 meaning indefinitly
      * @return <code>true</code> if all the servers suspended in time
      */
-    boolean awaitServerSuspend(Set<String> waitForServers, int timeoutInSeconds);
+    boolean awaitServerSuspend(Set<String> waitForServers, int timeoutInSeconds, BlockingTimeout blockingTimeout);
```
bstansberry: The new param needs javadoc.

host-controller/src/main/java/org/jboss/as/host/controller/ServerInventoryImpl.java
bstansberry: Yes, we can't have hardcoded exception messages or log messages that are INFO or above.

host-controller/src/main/java/org/jboss/as/host/controller/logging/HostControllerLogger.java
```diff
@@ -1265,4 +1266,12 @@ void noDomainControllerConfigurationProvidedForAdminOnly(String policyAttribute,
     @LogMessage(level = Level.ERROR)
     @Message( id = 172, value = "Failed executing the suspend operation for server: %s")
     void suspendExecutionFailed(@Cause IOException cause, String serverName);
+
+    @LogMessage(level = Level.INFO)
+    @Message(id = 173, value = "Timed out after %d ms awaiting server prepared response(s) for server: %s")
+    void timedOutAwaitingPreparedResponse(@Cause TimeoutException e, int blockingTimeout, String serverName);
```
bstansberry: I suggest changing the name of this method and the one below and the method text to refer to "suspend" rather than "prepared".

When we use "prepared" in messages end users may see we only mean the initial response the server sends, which you read in ServerInventoryImpl L444. But you can get timeouts or interrupted exceptions both there and when waiting for the final response in L454. So in messages that may be used due to L454, we shouldn't refer to "prepared".

host-controller/src/main/java/org/jboss/as/host/controller/logging/HostControllerLogger.java
```diff
@@ -1265,4 +1266,12 @@ void noDomainControllerConfigurationProvidedForAdminOnly(String policyAttribute,
     @LogMessage(level = Level.ERROR)
     @Message( id = 172, value = "Failed executing the suspend operation for server: %s")
     void suspendExecutionFailed(@Cause IOException cause, String serverName);
+
+    @LogMessage(level = Level.INFO)
```
bstansberry: This is an ERROR.

host-controller/src/main/java/org/jboss/as/host/controller/logging/HostControllerLogger.java
```diff
@@ -1265,4 +1266,12 @@ void noDomainControllerConfigurationProvidedForAdminOnly(String policyAttribute,
...
+    @Message(id = 173, value = "Timed out after %d ms awaiting server prepared response(s) for server: %s")
+    void timedOutAwaitingPreparedResponse(@Cause TimeoutException e, int blockingTimeout, String serverName);
+
+    @LogMessage(level = Level.INFO)
```
bstansberry: This is an ERROR.

host-controller/src/main/java/org/jboss/as/host/controller/ServerInventoryImpl.java
yersan: Hi, yes, but this exception manually throw is caught in the catch block in the same try{}catch{}, there I have logged it as @ehsavoie commented. The only question/detail here is with the current code I cannot distinguish between a timeout throwed by get in line 454 and a timeout due to line 444. Note that operationData.listener.retrievePreparedOperation doesn't throw an exception, if a timout exist, it just returning null, I throw here a timeout exception to use the same pice of code if the exception is throwed by the get in line 454.

But of course, I can change it if you think that currently it's not a correct way to do it.

host-controller/src/main/java/org/jboss/as/host/controller/ManagedServer.java
yersan: Sure, thanks! it can be changed, I will do it and I will pass the tests on local.

host-controller/src/main/java/org/jboss/as/host/controller/logging/HostControllerLogger.java
bstansberry: I think. ;) If you think it should be INFO, don't hesitate to tell me why. Same with the other one.

yersan: Yes, for it makes sense an INFO, I think :), I can argue two reasons:
1) Only the changes in this issue are using these log traces and if those exceptions are threw the returning operation is still a success (there are no other message for the user, he will see success as a result of :suspend-servers if a timeout occurrs) and maybe the suspend in one server was correct. For this reason I think that maybe it shouldn't be an error.
2) I followed the same traces in org.jboss.as.domain.controller.logging.DomainControllerLogger, there same traces are logged as INFO instead of ERROR.

Well, maybe should they be an error if the suspend operation failed for all existing servers? and even more, shouldn't the returning operation be an error instead of a success in that case?
Just let me know how do you see it.

host-controller/src/main/java/org/jboss/as/host/controller/ServerInventory.java
yersan: Javadoc added

host-controller/src/main/java/org/jboss/as/host/controller/logging/HostControllerLogger.java
yersan: Method signatures changed

host-controller/src/main/java/org/jboss/as/host/controller/ManagedServer.java
yersan: Changed to package protected

host-controller/src/main/java/org/jboss/as/host/controller/ServerInventoryImpl.java
kabir: You could
     if (prepared == null) {
           operationData.future.asyncCancel(true);
           HostControllerLogger.ROOT_LOGGER.timedOutAwaitingSuspendResponse(e, operationData.blockingTimeout, operationDataEntry.getKey());
           return false;
     }

Or simply use an empty string for the TimeoutException with a comment saying it will be caught below

bstansberry: The catch block L461 will log the TimeoutException, including its message, so if it has a message it has to come from HostControllerLogger so we can have proper i18n. If Kabir's first suggestion doesn't seem good, his second one of a comment and no message in the TimeoutException is fine.

Even if you dropped the @Cause param from the logger methods we'd still want to get rid of the hardcoded exception message. Otherwise at some point someone is sure to come along and decide to "fix" it.

host-controller/src/main/java/org/jboss/as/host/controller/logging/HostControllerLogger.java
bstansberry: I agree that if any servers failed to suspend that details of that should be provided to the caller (i.e. the OperationStepHandlers that call this method) and those OSHs should use that data to compose a failure message and pass it to OperationContext.getFailureDescription(). The user should see the failure.

host-controller/src/main/java/org/jboss/as/host/controller/ServerInventoryImpl.java
yersan: Ok, now I got it, of course i18n is the reason!, I will change this part.

yersan: Done, I have removed the hardcode text using the suggestion from @kabir.

host-controller/src/main/java/org/jboss/as/host/controller/logging/HostControllerLogger.java
yersan: These changes has been implemented. I'm not sure if there were implemented correctly, I'm waiting for feedback.

bstansberry: Thanks. The github UI is hiding this discussion because the original line I commented on has been changed. :)

kabir:
@yersan can you rebase?

yersan:
Hi @kabir , done !

yersan:
It looks like the errors are not related with the changes in this issue.
retest this please

host-controller/src/main/java/org/jboss/as/host/controller/DomainModelControllerService.java
```diff
@@ -1512,8 +1512,8 @@ public void resumeServer(String serverName) {
             }
 
             @Override
-            public boolean awaitServerSuspend(Set<String> waitForServers, int timeout) {
-                return false;
+            public List<ModelNode> awaitServerSuspend(Set<String> waitForServers, int timeout, BlockingTimeout blockingTimeout) {
+                return Collections.EMPTY_LIST;
```
bstansberry: This should be Collections.emptyList() to avoid the compiler warn.

host-controller/src/main/java/org/jboss/as/host/controller/ServerInventoryImpl.java
```diff
@@ -720,4 +765,36 @@ public void handle(Callback[] callbacks) throws IOException, UnsupportedCallback
...
+    private ModelNode getSuspendListenerFailedResponse(String serverName) {
+        String msg = HostControllerLogger.ROOT_LOGGER.suspendListenerFailedMsg(serverName);
+        return getFailureResponse(msg);
+    }
+
+    private ModelNode getFailureResponse(String msg){
+        final ModelNode response = new ModelNode();
+        response.get(OUTCOME).set(FAILED);
+        response.get(FAILURE_DESCRIPTION).set(msg);
+        return response;
```
bstansberry: The outcome and failure-description bits are not part of what this logic should do; the OperationContext handles that stuff. This could just be 

return new ModelNode(msg);

which can of course be inlined in the calling methods.

bstansberry:
Sorry for the slow response! Small changes I noted aside, this looks good.

yersan:
Sure, no problem! I have applied your suggestions.

bstansberry:
Thanks!

PR #1506, opened 2016-04-18
[WFLY-6529] Using regex for ports in XML schemas

rnetuka:
https://issues.jboss.org/browse/WFLY-6529


dmlloyd:
As a matter of policy, we have kept the variable resolution syntax out of the schema.  Unless we have agreed that the schema must now explicitly support this (something I disagree with), this change should not be merged.

dmlloyd:
Rejecting this PR.  Please see JIRA comments for more info.

PR #1508, opened 2016-04-19
JDK9 Support

ropalka:


core-feature-pack/src/main/resources/modules/system/layers/base/org/jboss/as/cli/main/module.xml
```diff
@@ -35,6 +35,25 @@
     </resources>
 
     <dependencies>
+        <!-- JAXP default dependencies. DO NOT REMOVE!!!
+                   These are loaded by the JAXP redirect facility in jboss-modules when this module is used
+                   as a boot module (this is the case for server processes in domain mode). The purpose is
```
bstansberry: Can you provide more details about how this module is used as a boot module for server processes in domain mode? It sounds strange to me that this module is relevant to a server process, except perhaps if the server is embedded.

pom.xml
```diff
@@ -138,9 +138,13 @@
         <version.org.zanata.plugin>3.1.1</version.org.zanata.plugin>
         <version.xml.plugin>1.0</version.xml.plugin>
 
+        <!-- Modularized JDK support (various workarounds) - activated via profile -->
+        <modular.jdk.args/>
+        <modular.jdk.props/>
+
         <!-- Surefire args -->
         <surefire.jpda.args/>
-        <surefire.system.args>-ea ${surefire.jpda.args}</surefire.system.args>
```
bstansberry: Is dropping -ea intentional?

host-controller/src/main/java/org/jboss/as/host/controller/Main.java
```diff
@@ -564,24 +564,18 @@ private static URL makeURL(String urlspec) throws MalformedURLException {
             for (String arg : runtime.getInputArguments()) {
                 if (arg != null && arg.length() > 2 && arg.startsWith("-D")) {
                     arg = arg.substring(2);
-                    String[] split = arg.split("=");
-                    if (!hostSystemProperties.containsKey(split[0])) {
```
bstansberry: This check has been dropped.

platform-mbean/src/main/java/org/jboss/as/platform/mbean/RuntimeMXBeanAttributeHandler.java
```diff
@@ -116,7 +116,11 @@ static void storeResult(final String name, final ModelNode store) {
         } else if (PlatformMBeanConstants.BOOT_CLASS_PATH_SUPPORTED.equals(name)) {
             store.set(ManagementFactory.getRuntimeMXBean().isBootClassPathSupported());
         } else if (PlatformMBeanConstants.BOOT_CLASS_PATH.equals(name)) {
-            store.set(ManagementFactory.getRuntimeMXBean().getBootClassPath());
+            if (!ManagementFactory.getRuntimeMXBean().isBootClassPathSupported()) {
+                store.set(""); // JDK9 and above does not have boot class path
+            } else {
+                store.set(ManagementFactory.getRuntimeMXBean().getBootClassPath());
+            }
```
bstansberry: How are the semantics here different from JDK 8, where this call could throw UnsupportedOperationException?

The way these resources handle these cases are if read-attribute is invoked reading an attribute like this, the UnsupportedOperationException is caught and an OperationFailedException is thrown. If read-resource is invoked, we catch the UnsupportedOperationException and return undefined as the attribute's value in the overall response. We fail on read-attribute because it's a direct request where we behave consistently with what the RuntimeMXBean does with a direct request. There is no RuntimeMXBean equivalent to read-resource, so we do what's logical when reading numerous attributes and aggregating.  

I don't want to deviate from that pattern for particular attributes unless it's absolutely necessary.

pom.xml
ropalka: It's not. It can be rolled back if necessary.

bstansberry:
Please move WFLY-4236 and WFLY-6359 to WFCORE and update the commit messages. Thanks!

pom.xml
bstansberry: Please do.

host-controller/src/main/java/org/jboss/as/host/controller/Main.java
ropalka: With this refactoring the semantics of this method didn't change (it is backward compatible).
What this code does it it fixes the handling of special -XaddExport parameters,
which have the form jdk9.module.name1=package1,jdk9.module.name2=package2.export.

core-feature-pack/src/main/resources/modules/system/layers/base/org/jboss/as/cli/main/module.xml
ropalka: I picked this change from Tomaz Cerar repository.
It's copy paste from org.jboss.as.server.main module AFAICS.
The purpose of this commit is CLI should always use OUR specified JAXP providers
and don't rely on those provided by JDK. I can update the comment accordingly.

host-controller/src/main/java/org/jboss/as/host/controller/Main.java
ropalka: Aha, you're right. I already see it Brian. I'll return it back.

platform-mbean/src/main/java/org/jboss/as/platform/mbean/RuntimeMXBeanAttributeHandler.java
ropalka: AFAIR there was one test failing because of this. I'll take a look if we could fix it somehow in that test instead.

core-feature-pack/src/main/resources/modules/system/layers/base/org/jboss/as/cli/main/module.xml
ctomc: comment was copy-pasted from server modules, but in any case it is still valid, problem that this change address is booting embedded server. CLI itself works embedded server boot doesn't.

platform-mbean/src/main/java/org/jboss/as/platform/mbean/RuntimeMXBeanAttributeHandler.java
ctomc: without this change we have mbean tests failures, I think it was the test that tries to read whole model via jmx.
But yes it should have thrown UnsupportedOperationException and we would catch it but it doesn't. 
I am still convinced this is bug in JDK itself and as such am not really happy getting this "fix/hack" into master just yet.

pom.xml
```diff
@@ -1465,6 +1469,20 @@
...
+          Descr: various workarounds activation for modularized JDK
+        -->
+        <profile>
+            <id>modularizedJdk</id>
+            <activation>
+                <jdk>9</jdk>
+            </activation>
+            <properties>
+                <modular.jdk.args>-XaddExports:java.security.sasl/com.sun.security.sasl.digest=ALL-UNNAMED,java.security.sasl/com.sun.security.sasl=ALL-UNNAMED</modular.jdk.args>
+                <modular.jdk.props> -Dsun.util.logging.disableCallerCheck=true</modular.jdk.props>
```
ctomc: we should have jira for this, so it is properly addressed in jboss logging when time comes.

ropalka: I agree, I'll create one.

core-feature-pack/src/main/resources/modules/system/layers/base/org/jboss/as/cli/main/module.xml
ropalka: So I can leave that comment there as it is or it needs clarification?

pom.xml
jamezp: There's likely nothing the JBoss Log Manager can do. I'll have to review the JDK code again (it's recently changed quite a bit for jigsaw), but if a platform logger is used on the class with a `null` class loader it loads the JUL log manager. This happens before the system property is set IIRC.

There's a parent log manager JIRA for JDK 9 things BTW https://issues.jboss.org/browse/LOGMGR-130

platform-mbean/src/main/java/org/jboss/as/platform/mbean/RuntimeMXBeanAttributeHandler.java
ctomc: this was my "hack" at the time to properly fix it https://github.com/ctomc/wildfly-core/commit/d5f31409304254c94a58f1d10c0f02852719f425 but this looks like problem in JDK as we don't get UnsupportedOperationException itself but only has that in message.

core-feature-pack/src/main/resources/modules/system/layers/base/org/jboss/as/cli/main/module.xml
bstansberry: Please clarify it as it's quite confusing now.

bstansberry:
Re: the alternative fix, I don't want to deal with this attribute as an exception case, and we don't want our management API to fluctuate that way. We either expose an attribute or we don't.

I discussed our general behavior around these UOE-throwing methods with David and we think we should change it. So I opened https://issues.jboss.org/browse/WFCORE-1492.  But that's a big task; for this PR let's go with Tomaz's workaround for that test.

remoting/tests/pom.xml
```diff
@@ -51,6 +51,22 @@
...
+        </dependency>
+        <dependency>
+            <groupId>xerces</groupId>
+            <artifactId>xercesImpl</artifactId>
+            <scope>test</scope>
+        </dependency>
+        <dependency>
+            <groupId>xalan</groupId>
+            <artifactId>xalan</artifactId>
+            <version>2.7.1.jbossorg-2</version>
```
ctomc: this is nitpicking but I this should be managed by parent pom, so you can remove the version here

ropalka: Just for tracking this is for tests only and we agreed with Tomaz it is not necessary to manage it from parent.

bstansberry:
The 3 test failures shown at http://brontes.lab.eng.brq.redhat.com/viewLog.html?buildId=97158&buildTypeId=WildFlyCore_PullRequestWindows are recurring

ropalka:
Unfortunately I don't have access to Windows machine to reproduce it.
I asked Tomaz Cerar to test it and provide PR to fix it.

ctomc:
failure on windows is actual problem in patching code, it is fixed in https://github.com/wildfly/wildfly-core/pull/1525 PR with is this pr with extra commit that fixes windows failures.
Origin of the problem is non setting of encoding when writing xml file to disk. This can result on platform dependent encoding that in some cases xml parser doesn't understand. 
I think this problem started to pop up more frequently on public ci as it uses newer JDKu build that has some changes on how it determines default charset for files.

bstansberry:
Closing in favor of #1525 which is this plus a fix for the testsute.

PR #1509, opened 2016-04-20
[WFCORE-1484] Emit management notifications for beginning/completing …

bstansberry:
…MSC modifications


bstansberry:
Any reviewers are welcome, but I'd like a +1 from @stuartwdouglas before merging, as I wrote this to meet his use case. So might as well make sure it works for that. ;)

stuartwdouglas:
Sorry, missed the notification. This looks good to me.

PR #1510, opened 2016-04-20
[WFCORE-924] Refactor management interfaces so common code can be shared across standalone and domain mode.

darranl:

Originally when the management interfaces were first implemented their configuration only contained information about what interfaces and ports to use, as this problem was solved differently in standalone mode and in a host controller the two approaches were truly different.  As time has passed more and more common code has been added which has to be added for both modes.

This pull requests gives the management interfaces common base code shared across both modes, it also better defines how common configuration is obtained as we had also been abusing an object that is only supposed to provide connection information and not all the configuration details.

This pull request does not change the model at all.


bstansberry:
retest this please

bstansberry:
This is ok, but at some point in 3.x this new org.jboss.as.controller.management package is going to move to a different module and perhaps be renamed. I'm going to split up the controller module so that all that's left is API and support classes for subsystem developers. The actual impl stuff, plus the smattering of "shared" code between the server and the HC will get moved, mostly to a new module.

PR #1513, opened 2016-04-20
[WFCORE-1486] PersistanceResourceTestCase fails on some windows machines

pkremens:
PersistanceResourceTestCase could fail on the servers with multiple acl entries for file owner. Weaken the assertion of the test to pass in case that at least one principlal.equals(owner) acl entry contain WRITE_OWNER permission.

JIRA: 
https://issues.jboss.org/browse/WFCORE-1486 - PersistanceResourceTestCase fails on some windows machines

### DIFF
--- a/controller/src/test/java/org/jboss/as/controller/persistence/PersistanceResourceTestCase.java
+++ b/controller/src/test/java/org/jboss/as/controller/persistence/PersistanceResourceTestCase.java
@@ -49,10 +49,12 @@ import java.nio.file.attribute.PosixFilePermission;
 import java.nio.file.attribute.UserPrincipal;
 import java.util.ArrayList;
 import java.util.Arrays;
+import java.util.Collection;
 import java.util.Collections;
 import java.util.HashSet;
 import java.util.List;
 import java.util.Set;
+import java.util.stream.Collectors;
 
 import org.jboss.as.controller.PathAddress;
 import org.jboss.dmr.ModelNode;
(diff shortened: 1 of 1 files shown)


pkremens:
retest this please

ctomc:
@pkremens this is wrong fix. As now this test fails on my Windows 10 machine.
Main problem is that assumption you build upon that owner needs to have "special" owner permissions is wrong.
as owner can be in administrators / users or any other group that has WRITE permissions so test shouldn't test for this, at least not in current situation.

pkremens:
@ctomc thanks for alerting me. I will look into it.

ctomc:
@pkremens 
I've locally commented this out
```
//commented out as this is wrong assumption, user can be an owner and doesn't need to have any special owner permissions depending in what groups he belongs in.
//assertThat(ownerPermissions.toString(), ownerPermissions, hasItem(AclEntryPermission.WRITE_OWNER));
```
I can send PR with this change if you want. In my case ownerPermissions is empty list.
I am on windows 10 default install, no special fondling with permissions and users.

pkremens:
@ctomc I manage to get access to Windows10 instance from Central CI. I cannot find an easy way to get a super set of user permissions. I thinking of either comment the assertion out (as you suggested), or test the ownerPermissions for emptiness for assertion:

     if (!ownerPermissions.isEmpty()) {
         assertThat(ownerPermissions.toString(), ownerPermissions, hasItem(AclEntryPermission.WRITE_OWNER));
     }

I'll try this on other servers in MWQE lab where this was causing us trouble.

ctomc:
@pkremens sounds good.

PR #1515, opened 2016-04-22
[WFCORE-1498] Log different messages in different shutdown scenarios

bstansberry:


kabir:
Retest this please

host-controller/src/main/java/org/jboss/as/host/controller/logging/HostControllerLogger.java
```diff
@@ -104,7 +106,7 @@
      * @param port          the port number.
      * @param securePort    the secure port number.
      */
-    @LogMessage(level = Level.INFO)
+    @LogMessage(level = INFO)
```
bstansberry: Oops; this kind of cosmetic change wasn't intentional. I can reverse it if wanted.

bstansberry:
retest this please

 failure is due to WFCORE-1501

PR #1516, opened 2016-04-22
[WFCORE-1497] Pattern matching scoped roles

bstansberry:
Just for CI; needs some tests, although it works with manual testing.

### DIFF
--- a/controller/src/main/java/org/jboss/as/controller/access/constraint/PatternScopedConstraint.java
+++ b/controller/src/main/java/org/jboss/as/controller/access/constraint/PatternScopedConstraint.java
@@ -1,0 +1,217 @@
+/*
+Copyright 2016 Red Hat, Inc.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+  http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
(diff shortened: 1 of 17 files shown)


bstansberry:
retest this please

controller/src/main/java/org/jboss/as/controller/access/constraint/PatternScopedConstraint.java
```diff
@@ -0,0 +1,215 @@
...
+                assert !psc.user : "illegal comparison";
+                if (!readOnly) {
+                    if (!global) {
+                        if (actionEffect == Action.ActionEffect.WRITE_RUNTIME || actionEffect == Action.ActionEffect.WRITE_CONFIG) {
+                            Set<Pattern> ourPatterns = patternsHolder.patterns;
+                            String theTarget = psc.target;
+                            //  Ok as long as one of our patterns match
+                            boolean anyMatch = anyMatch(ourPatterns, theTarget);
+                            if (!anyMatch) ControllerLogger.ACCESS_LOGGER.tracef("pattern constraint violated " +
+                                            "for action %s due to no match between target %s and allowed patterns %s",
```
Ladicek: Could that possibly leak information, if the patterns target non-addressable resources? That should be unlikely, but the possibility occured to me immediately :-)

bstansberry: Good point; I'll change that.

bstansberry:
I'm closing this for now as I've gotten the feedback I was hoping for and I don't want this polluting the PR queue until I get a chance to incorporate that feedback. Plus add some tests. ;)

PR #1517, opened 2016-04-22
[WFCORE-1499] Parse the module name for custom handlers so slots can …

jamezp:
…be used.

### DIFF
--- a/logging/src/main/java/org/jboss/as/logging/HandlerOperations.java
+++ b/logging/src/main/java/org/jboss/as/logging/HandlerOperations.java
@@ -287,7 +287,7 @@ final class HandlerOperations {
             if (moduleName != null) {
                 // Check if this is a log4j appender
                 final ModuleLoader moduleLoader = ModuleLoader.forClass(HandlerOperations.class);
-                final ModuleIdentifier id = ModuleIdentifier.create(moduleName);
+                final ModuleIdentifier id = ModuleIdentifier.fromString(moduleName);
                 try {
                     final Class<?> actualClass = Class.forName(className, false, moduleLoader.loadModule(id).getClassLoader());
                     if (Appender.class.isAssignableFrom(actualClass)) {


jamezp:
retest this please

bstansberry:
retest this please

ppalaga:
@jamezp looking at how to backport this to EAP 6.4, I am finding out that in addition to the present fix, a logmanager 2.0.x release and upgrade is needed - in particular, this logmanager commit https://github.com/jboss-logging/jboss-logmanager/commit/01f5ad3c8b53249b87cf5262955be4ef6d415952 Note that logmanager 1.5.6.Final delivers an analogous commit already.

@jamezp I'd prepare a test case for WFCORE-1499 unless you have one somewhere already?

ppalaga:
@jamezp I prepared a test [1]. Please feel free to attach it to your future PR that will upgrade the logmanager.

[1] https://github.com/ppalaga/wildfly-core/commit/aeb4d8536fda8881fd1d7293752bf6fcf8991a35 in https://github.com/ppalaga/wildfly-core/tree/WFCORE-1499-test - that is on top of 2.x branch

jamezp:
@ppalaga Thanks! AFAIK the test goes I think I'd rather leave it out of WildFly for 2 reasons. 1) Slots in modules are going away. 2) I prefer tests like these in the project (jboss-logmanager in this case) rather than the consumer of the project. Mainly because you won't know it's broken until after a release.

ppalaga:
> 1) Slots in modules are going away.

Not in EAP 6 and 7 ;)

> Mainly because you won't know it's broken until after a release.

Or until you start using srcdeps ;)

ctomc:
>> 1) Slots in modules are going away.
> Not in EAP 6 and 7 ;)
Well we are limiting it usage for new stuff already in 7 as we want to be able to run EAP7 on JDK9 some day in the future, so why shut ourselfs in the leg if not necessary :)

PR #1519, opened 2016-04-25
[WFCORE-1472] Boolean value could become optional

jfdenise:
- CLI parsing and completion changes to handle implicit values.
- Introduce CommandContext.Scope to formalise context storage duration.
- New unit tests

### DIFF
--- a/cli/src/main/java/org/jboss/as/cli/batch/impl/DefaultBatch.java
+++ b/cli/src/main/java/org/jboss/as/cli/batch/impl/DefaultBatch.java
@@ -23,6 +23,9 @@ package org.jboss.as.cli.batch.impl;
 
 import java.util.ArrayList;
 import java.util.List;
+import org.jboss.as.cli.CommandContext;
+import org.jboss.as.cli.CommandContext.Scope;
+import org.jboss.as.cli.CommandFormatException;
 
 import org.jboss.as.cli.Util;
 import org.jboss.as.cli.batch.Batch;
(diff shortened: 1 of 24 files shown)


jfdenise:
retest this please!

jfdenise:
retest this please!

cli/src/main/java/org/jboss/as/cli/batch/impl/DefaultBatch.java
```diff
@@ -103,8 +106,25 @@ public ModelNode toRequest() {
...
+        for (BatchedCommand cmd : commands) {
+            CommandContext ctx = cmd.getCommandContext();
+            ModelNode request = cmd.getRequest();
+            if (ctx.getConfig().isValidateOperationRequests()) {
+                try {
+                    ctx.set(Scope.REQUEST, Util.DESCRIPTION_RESPONSE,
+                            cmd.getDescriptionResponse());
+                    ModelNode opDescOutcome = Util.validateRequest(ctx, request);
+                    if (opDescOutcome != null) { // operation has params that might need to be replaced
+                        Util.replaceFilePathsWithBytes(request, opDescOutcome);
```
aloubyansky: File paths substitution appears to happen only if the validation is on. It should still happen when the validation is off.

aloubyansky: It's actually how it used to be before, so it's not a regression. I just spotted that.

jfdenise: The description is only retrieved when validation is enabled, can we really move it outside validation? It would mean the the would make the description retrieval in all cases.

jfdenise:
retest this please!

cli/src/main/java/org/jboss/as/cli/batch/impl/DefaultBatch.java
aloubyansky: In this case, we have to do it. Because if the validation is disabled the substitution of file paths won't happen and the request will fail.
But as I wrote before, it's a separate issue and not a regression introduced by these changes.

aloubyansky:
This is good to merge. Excellent work, Jean-Francois!

PR #1520, opened 2016-04-25
[WFCORE-1505] Add services for local and remote outbound socket bindi…

kabir:
…ngs for the HC model

https://issues.jboss.org/browse/WFCORE-1505


bstansberry:
@kabir I believe the remove and write-attribute handling needs to be addressed as well.

kabir:
@bstansberry Good catch. I have added it to the write attribute handler as well now

PR #1521, opened 2016-04-25
[WFCORE-1501] Log ERROR, don't fail, if timestamped history can't be …

bstansberry:
…created

### DIFF
--- a/controller/src/main/java/org/jboss/as/controller/persistence/ConfigurationFile.java
+++ b/controller/src/main/java/org/jboss/as/controller/persistence/ConfigurationFile.java
@@ -653,7 +653,7 @@ public class ConfigurationFile {
             if (currentHistoryFiles != null && currentHistoryFiles.length > 0) {
                 String backupName = getTimeStamp(date);
                 File old = new File(historyRoot, backupName);
-                if (!new File(currentHistory.getAbsolutePath()).renameTo(old)) {
+                if (!forcedMove(currentHistory.toPath(), old.toPath())) {
                     if (old.exists()) {
                         // AS7-5801. Unit tests sometimes fail on File.renameTo due to only having 100 ms
                         // precision on the timestamps we use for dir names on some systems. So, if that happens,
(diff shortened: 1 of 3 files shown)


controller/src/main/java/org/jboss/as/controller/persistence/FilePersistenceUtils.java
```diff
@@ -77,9 +77,13 @@ static void copyFile(final File file, final File backup) throws IOException {
...
-    static void rename(File file, File to) throws IOException {
-        if (!file.renameTo(to) && file.exists()) {
-            copyFile(file, to);
+    static boolean rename(Path from, Path to) {
+        try {
+            Files.move(from, to);
+            return true;
+        } catch (IOException e) {
+            ControllerLogger.ROOT_LOGGER.cannotRename(e, from, to);
+            return false;
```
bstansberry: Note: this method wasn't used at all in current code so there's no need to worry about any behavior change from this part of the PR. It's just whether the logic is correct.

luck3y: It seems a little strange to have the method be called rename(), but it actually perform a forced move, maybe a boolean param to indicate you really want to force move if the file already exists would be safer?

bstansberry: Fair point; I just took an existing unused method and changed it's internals. No reason not to give it a new name as well.

I definitely really want a forced move. If someone has a different use case they can add a param. But, there's no reason I can't move this method to the only class that uses it which removes any question about whether its API is correct.

luck3y:
+1 looks good.

controller/src/main/java/org/jboss/as/controller/persistence/ConfigurationFile.java
```diff
@@ -763,6 +763,16 @@ private File mkdir(final File dir) {
         return dir;
     }
 
+    private static boolean forcedMove(Path from, Path to) {
+        try {
+            Files.move(from, to);
```
ctomc: If you call method forceMove maybe it should use StandardCopyOption.REPLACE_EXISTING

bstansberry: OK, will do. This is only used in cases where 'from' and 'to' are dirs, and with REPLACE_EXISTING Files.move will fail if 'to' exists and is non-empty. But without REPLACE_EXISTING it will fail if 'to' exists, regardless of whether it's empty, so adding the option may help.

luck3y:
retest this please

PR #1524, opened 2016-04-26
[WFCORE-1508]

stalep:
make sure the prompt is always displayed
if the console is started from other methods than CommandContextImpl.interact
we need to make sure that we redisplay the prompt when interact is eventually
called.


aloubyansky:
This is good to merge. Thanks, Stale!

aloubyansky:
Stale, sorry, the issue is WFCORE-1508. WFCORE-1315 is a previous double prompt which has already been resolved.

stalep:
ah, ok. ill rebase it :)

PR #1528, opened 2016-04-28
Improve test runner to allow few more things when running in manual server mode

ctomc:
fixed version of #1227 

- allows us to not use arquillian in manual tests in full.
- adds support for deploy/undeploy operations on container.
- adds reload() method on server

### DIFF
--- a/testsuite/manualmode/src/test/java/org/jboss/as/test/manualmode/deployment/DeploymentScannerShutdownTestCase.java
+++ b/testsuite/manualmode/src/test/java/org/jboss/as/test/manualmode/deployment/DeploymentScannerShutdownTestCase.java
@@ -21,20 +21,12 @@
  */
 package org.jboss.as.test.manualmode.deployment;
 
-import static org.jboss.as.controller.ControlledProcessState.State.RUNNING;
 import static org.jboss.as.controller.descriptions.ModelDescriptionConstants.ADD;
 import static org.jboss.as.controller.descriptions.ModelDescriptionConstants.EXTENSION;
-import static org.jboss.as.controller.descriptions.ModelDescriptionConstants.NAME;
 import static org.jboss.as.controller.descriptions.ModelDescriptionConstants.OP;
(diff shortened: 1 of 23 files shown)


ctomc:
retest this please

testsuite/manualmode/src/test/java/org/jboss/as/test/manualmode/deployment/DeploymentScannerShutdownTestCase.java
```diff
@@ -120,22 +110,32 @@ public static void addDeploymentScanner() throws Exception {
...
+        try {
+            removeDeploymentScanner(scannerName);
+        } catch (Exception e) {
+
+        }
+        try {
+            removeDeploymentScannerExtension();
+        } catch (Exception e) {
+
+        }
```
kabir: If one of those exceptions happen, it could be worth storing in a variable and rethrowing after container.stop()?

ctomc: not really, they are here just so that container.stop() is called for sure.
otherwise we would get inconsistent state, they also log all the errors.

testsuite/shared/src/main/java/org/jboss/as/test/integration/security/common/CoreUtils.java
```diff
@@ -178,7 +178,7 @@ public static void applyUpdates(final List<ModelNode> updates, final ModelContro
 
     public static void applyUpdate(ModelNode update, final ModelControllerClient client) throws Exception {
         ModelNode result = client.execute(new OperationBuilder(update).build());
-        if (LOGGER.isInfoEnabled()) {
+        if (LOGGER.isDebugEnabled()) {
```
kabir: Nitpick. why?

ctomc: less logging by default === less IO, less logging, --> faster execution --> faster CI execution.
You can still enable it if you want it.

PR #1529, opened 2016-04-28
[WFCORE-307]: When Wildfly fails with JBAS015810: failed to resolve interface management, it should terminate.

ehsavoie:
Changing the way a server boots by moving all management and socket related operations to a first batch that will stop the boot process if there is an error.
The other operations subsystems mostly) are in a second batch that won't stop the server from booting.
Fixing some tests according to this new paradigm.
Adding a hack in the ModelControllerImpl for Core Model Tests as we don't have services ready in those tests and they would be breaking with this change.

Jira: https://issues.jboss.org/browse/WFCORE-307

### DIFF
--- a/controller/src/main/java/org/jboss/as/controller/ModelControllerImpl.java
+++ b/controller/src/main/java/org/jboss/as/controller/ModelControllerImpl.java
@@ -100,6 +100,7 @@ import org.jboss.msc.service.ServiceRegistry;
 import org.jboss.msc.service.ServiceTarget;
 import org.jboss.threads.AsyncFuture;
 import org.jboss.threads.AsyncFutureTask;
+import org.wildfly.security.manager.WildFlySecurityManager;
 import org.wildfly.security.manager.action.GetAccessControlContextAction;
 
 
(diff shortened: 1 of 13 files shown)


ehsavoie:
retest this please

controller/src/main/java/org/jboss/as/controller/ModelControllerImpl.java
```diff
@@ -448,11 +449,19 @@ boolean boot(final List<ModelNode> bootList, final OperationMessageHandler handl
...
                 ? EnumSet.of(AbstractOperationContext.ContextFlag.ROLLBACK_ON_FAIL)
                 : EnumSet.noneOf(OperationContextImpl.ContextFlag.class);
+        //TODO Gigantic HACK to disable the runtime part of this for the core model testing.
+        //The core model testing currently uses RunningMode.ADMIN_ONLY, but in the real world
+        //the http interface needs to be enabled even when that happens.
+        //I don't want to wire up all the services unless I can avoid it, so for now the tests set this system property
+        //@see https://issues.jboss.org/browse/WFCORE-1509
+        EnumSet<OperationContextImpl.ContextFlag> firstBatchFlags = WildFlySecurityManager.getPropertyPrivileged("wildfly.test.boot.disable.rollback", null) == null
+                ? EnumSet.of(AbstractOperationContext.ContextFlag.ROLLBACK_ON_FAIL)
+                : EnumSet.noneOf(OperationContextImpl.ContextFlag.class);
```
bstansberry: I haven't thought this through at all, but perhaps using ProcessType.EMBEDDED_SERVER for these tests is the way to go. That plus RunningMode.ADMIN_ONLY should result in not starting the management interface services.

bstansberry: BTW, if all else is ok on this, ^^^ is thoughts for a future change, not required for this PR.

controller/src/main/java/org/jboss/as/controller/ModelControllerImpl.java
```diff
@@ -619,8 +637,8 @@ private BootOperations organizeBootOperations(List<ModelNode> bootList, final in
                 }
             }
         }
-
-
+        Collections.sort(managementOps, (ParsedBootOp o1, ParsedBootOp o2) -> o1.address.toString().compareTo(o2.address.toString()));
```
bstansberry: Why?

controller/src/main/java/org/jboss/as/controller/ModelControllerImpl.java
```diff
@@ -619,8 +637,8 @@ private BootOperations organizeBootOperations(List<ModelNode> bootList, final in
...
+        initialOps.addAll(managementOps);
```
bstansberry: Why do the management ops have to come later?

bstansberry: I'm guessing the answer has something to do with audit logging. :)

testsuite/domain/src/test/java/org/jboss/as/test/integration/domain/suites/ServerRestartRequiredTestCase.java
```diff
@@ -144,29 +134,6 @@ public void stopServers() throws Exception {
...
-        socketBindingAdd.get(PORT).set(new ValueExpression("${my.custom.socket}"));
-
-        // Don't rollback server failures, rather mark them as restart-required
-        socketBindingAdd.get(OPERATION_HEADERS).get(ROLLOUT_PLAN)
-                .get(IN_SERIES).add().get(CONCURRENT_GROUPS).get("reload-test-group").get(MAX_FAILURE_PERCENTAGE).set(100);
-
-        executeOperation(socketBindingAdd, client);
-
-        checkRestartOneAndTwo(client);
-    }
```
bstansberry: This seems like a useful test, but needs different checks when the servers are started.

controller/src/main/java/org/jboss/as/controller/audit/ManagedAuditLoggerImpl.java
```diff
@@ -528,6 +528,35 @@ public void updateInMemoryHandlerMaxHistory(String name, int maxHistory) {
         }
     }
 
+    public void copy(ManagedAuditLoggerImpl other) throws IOException {
```
bstansberry: If the process is booting and the audit logger is not configured for boot logging, I think this should just not copy.

bstansberry: Never mind; I missed that you handled this on the other side.

ehsavoie: Yes, maybe using two references instead of switching would make it feel less hacky and more straightforward.

ehsavoie:
retest this please

PR #1530, opened 2016-04-28
WFCORE-1510 Fix code to not use default platform dependant encoding

ctomc:


cli/src/main/java/org/jboss/as/cli/gui/metacommand/ScriptAction.java
```diff
@@ -85,22 +85,11 @@ protected void runScript(File script) {
...
-                lines.add(line);
-                line = reader.readLine();
-            }
-        } catch (Throwable e) {
-            throw new IllegalStateException("Failed to process file " + file.getAbsolutePath(), e);
-        } finally {
-            StreamUtils.safeClose(reader);
+            return Files.readAllLines(file.toPath(), StandardCharsets.UTF_8);
+        } catch (IOException e1) {
+            return Collections.emptyList();
```
bstansberry: This is an behavior change.

bstansberry:
There are a lot of cases here where XXXUtils.safeClose() or equivalent code has been replaced with a simple try-with-resources. But a simple try-with-resources is not equivalent, as the close() call at the end of try-with-resources can throw IOException[1], and the simple handling is not catching and discarding it.

To be equivalent you'd need something like this:

```
boolean ok = false;
try (InputStream is = getInputStream()) {
    doStuffWithStream(is);
    ok = true;
} catch (IOException e) {
    if (!ok) {
        throw e;
    }
}
```

Changes like this that touch a lot of classes need to stick closely to their goal as it is too error prone to analyze a large number of small semantic changes to see if they are all ok (i.e. if not suppressing the exception on close() is ok.)

[1] See Suppressed Exceptions on https://docs.oracle.com/javase/tutorial/essential/exceptions/tryResourceClose.html

cli/src/main/java/org/jboss/as/cli/gui/metacommand/ScriptAction.java
ctomc: Not really. Original code does 
`List<String> lines = new ArrayList<String>();`
on top and than returns that it if noting was read.
new code does the same with just different API

```
try {
    return Files.readAllLines(file.toPath(), StandardCharsets.UTF_8);
} catch (IOException e1) {
    return Collections.emptyList();
}
```

ctomc:
@bstansberry good point on suppressed exceptions, didn't about know that about try-with-resources.
I did go trough code and there are only handful of cases where code needs to be fixed to properly account for that, but for majority change is fine (mostly in the test code).
Will update PR once I've change the code to account for suppressed exceptions.

cli/src/main/java/org/jboss/as/cli/gui/metacommand/ScriptAction.java
bstansberry: The old code handled any exception by throwing ISE; the new code suppresses IOException.

ctomc: fixed

ctomc:
@bstansberry @jamezp any love for this?

bstansberry:
I'd love it a lot more if it did the bare minimum and avoided changes not directly related to specifying the character encoding, e.g. no changing try/finally to try-with-resources unless the try/finally was clearly broken. There are just a lot of changes in here and too many times these kinds of changes have introduced bugs.

ctomc:
@bstansberry Thing is that in most cases changes like that ware required as API that was used did "support" a way to specify encoding / charset to read / write with.

cli/src/main/java/org/jboss/as/cli/handlers/CommandHandlerWithHelp.java
```diff
@@ -91,8 +91,8 @@ public void handle(CommandContext ctx) throws CommandLineException {
     protected void printHelp(CommandContext ctx) throws CommandLineException {
         InputStream helpInput = WildFlySecurityManager.getClassLoaderPrivileged(CommandHandlerWithHelp.class).getResourceAsStream(filename);
         if(helpInput != null) {
-            BufferedReader reader = new BufferedReader(new InputStreamReader(helpInput));
-            try {
+
+            try (final BufferedReader reader = new BufferedReader(new InputStreamReader(helpInput, StandardCharsets.UTF_8));){
```
bstansberry: This is an example of what I am talking about. This change to try with resources is unrelated to the change to pass the charset to new InputStreamReader.

If we merge a PR with a lot of these kinds of changes I can almost guarantee that there will be bugs introduced and the hassle from them will be greater than whatever trouble we get from not passing in the charset. And I don't want to go through dozens of unrelated changes like this and try to analyze if they are safe.

ctomc: I've removed all unnecessary try-with-resource changes.
They are only left in few places where the whole open "resource/use it/close it" was used wrongly from beginning. In most cases that is where reader/writer is created outside try block and closed in finally. Only those ware kept with change to try-with-resources, all others ware changed to minimal change needed.

cli/src/main/java/org/jboss/as/cli/handlers/ArchiveHandler.java
```diff
@@ -104,8 +105,7 @@ public ModelNode buildRequestWithoutHeaders(CommandContext ctx) throws CommandFo
             }
             ctx.printLine("Processing script '" + script + "'.");
 
-            try {
-                BufferedReader reader = new BufferedReader(new FileReader(scriptFile));
+            try (BufferedReader reader = Files.newBufferedReader(scriptFile.toPath(), StandardCharsets.UTF_8)){
```
ctomc: Here is changed to try-with-resources as reader is not closed at all.

PR #1531, opened 2016-04-29
[WFCORE-1517] add the 1.8 XML namespace

Ladicek:


bstansberry:
I think there must be more missing than this. If the enum isn't there, then the parsing logic that gets switched to must not be there either.

Ladicek:
Aha. I've only tried the default configuration, so I couldn't find any other problem. In that case, please reject this PR -- I'm definitely not capable of doing a proper backport :-)

bstansberry:
Ok, I'll get this sorted. Thanks for the report!

bstansberry:
Turns out the parsing code was ported, and this change is all that was missing. The missing enum value wasn't noticed because the parsing for the new stuff was handled by a "default" case in the switch.

PR #1533, opened 2016-04-29
WFCORE-362 - fix jvm management model for server and groups

luck3y:


luck3y:
This seems like the simplest approach, with minimal changes. I tried several approaches generically validating in ValidateModelStepHandler, but without extensive changes, this doesn't seem feasible at the moment.

I think we'd need a way to indicate a child key was unique, rather than the default (key, value) from the ResourceRegistration, so like a boolean validateByKey() or something, or we'd need to change the default implementations in ResourceDefinition / ImmutableManagementResourceRegistration to somehow be aware of the two different semantic meanings.

luck3y:
This is a legit failure I'm checking that test.

host-controller/src/main/java/org/jboss/as/host/controller/model/jvm/JVMAddHandler.java
```diff
@@ -43,12 +52,48 @@
...
+        super.populateModel(context, operation, resource);
+        final ModelNode model = resource.getModel();
+        final PathAddress jvm = PathAddress.pathAddress(operation.require(ModelDescriptionConstants.OP_ADDR));
+
+        context.addStep(new OperationStepHandler() {
+            public void execute(final OperationContext context, final ModelNode operation) throws OperationFailedException {
+                Resource root;
+                try {
+                    root = context.readResourceFromRoot(context.getCurrentAddress().getParent(), false);
+                } catch (Resource.NoSuchResourceException e) {
```
luck3y: This occurs in the domain ignore test case, where the slave is ignoring server-group="group2". When the group is added via changing the server group, the jvm child is added, but the parent isn't available at to RR. Is there a better way to do this?

It seems to work and test out fine, but it seems a bit clumsy.

luck3y:
retest this please

luck3y:
Hmm, isn't the fixme supposed to drop off after test pass?

luck3y:
retest this please

luck3y:
retest this please

PR #1535, opened 2016-04-30
[WFCORE-1503] Improve performance of the dynamic log files on the logging subsystem resource.

jamezp:
If several file handlers are defined with several associated files traversing the log directory can be slow. This keeps walking the file tree to a minimum to improve performance.

### DIFF
--- a/logging/src/main/java/org/jboss/as/logging/LoggingResource.java
+++ b/logging/src/main/java/org/jboss/as/logging/LoggingResource.java
@@ -39,7 +39,7 @@ import java.util.LinkedHashSet;
 import java.util.List;
 import java.util.Set;
 import java.util.TreeSet;
-import java.util.stream.Collectors;
+import java.util.concurrent.atomic.AtomicBoolean;
 
 import org.jboss.as.controller.PathAddress;
 import org.jboss.as.controller.PathElement;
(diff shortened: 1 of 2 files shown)


logging/src/main/java/org/jboss/as/logging/LoggingResource.java
```diff
@@ -290,7 +319,10 @@ public FileVisitResult visitFile(final Path file, final BasicFileAttributes attr
                     // This should always exist, but better to be safe
                     if (handlerModel.hasDefined(CommonAttributes.FILE.getName())) {
                         final ModelNode fileModel = handlerModel.get(CommonAttributes.FILE.getName());
-                        if (fileModel.hasDefined(PathResourceDefinition.PATH.getName())) {
+                        // Only allow from the jboss.server.log.dir
+                        if (fileModel.has(PathResourceDefinition.RELATIVE_TO.getName())
```
bstansberry: hasDefined would be better. A proper resource model will return true for "has" for all attributes as the add handler will populate everything either with a normal value or undefined.

logging/src/main/java/org/jboss/as/logging/LoggingResource.java
```diff
@@ -126,7 +125,19 @@ public Resource requireChild(final PathElement element) {
...
+                final Collection<String> validFileNames = findValidFileNames(getFileHandlersModel());
+                // If there's at least one child, then we have children
+                for (String name : validFileNames) {
+                    final Path file = dir.resolve(name);
+                    if (Files.exists(file) && Files.isReadable(file)) {
+                        return true;
+                    }
+                }
+            }
+            return false;
```
bstansberry: Will this result in no longer reporting children if the base file has been removed but a rotated file is still present? And if so is that ok?

jamezp: Correct. If the `*-file-handler` has been removed the rotated files will no longer be visible. I think it's the correct behavior as we have a rule that only defined handlers can be read, viewed, downloaded.

bstansberry: I don't mean the handler has been removed, I mean the base file. So if the handler that writes the server log is still there, the user has for some reason deleted server.log but 2016-05-03-1234353-server.log is still present, this will return false.

jamezp: Yeah that's a good point. I guess I should care, but I'm trying to decide if I do :) Well more weighing the performance cost vs consequences. We'd have to walk the file tree which seems to take a big hit on Windowst when there are a significant number files.

bstansberry: I'm not particularly arguing for it or against it. It's definitely an edge case that someone would delete the base file but then still expect to be able to use the management API to read the other ones.

Perhaps you can optimize the normal case (which you have here) and also optimize the "no files in the dir at all" case, and then have the slow approach for the hopefully unusual corner case. Or something like that.

jamezp:
retest this please

logging/src/main/java/org/jboss/as/logging/LoggingResource.java
jamezp: Yes that's what I did. I first check if file existence and if not found, then walk the tree.

PR #1537, opened 2016-05-02
Strip carriage returns from content files to ensure the SHA-1 matches.

jamezp:
Not really an ideal fix, but if files are checkout out on Windows replacing the '\n` with `\r\n` the tests will fail as the hash won't match. This is replaces the input stream with one that strips out all the '\r` bytes.

It's possible you're not supposed to use the line separator replacement when setting up git. If that's the case this can just be rejected.


bstansberry:
Since you've done it, might as well keep it. :)

PR #1538, opened 2016-05-03
[WFCORE-1520] fixed AttributeName parser used at completion time

jfdenise:
Fixed parser to handle non attribute compliant character when parsing attribute names. Added unit test.


kabir:
@aloubyansky one for you :)

aloubyansky:
Yes, that's good. Thanks!

PR #1539, opened 2016-05-03
[WFCORE-1513] Register metrics depends on the server type

jmesnil:
[WFCORE-1513] Register metrics depends on the server type

* register metrics in the ConcreteResourceRegistration only when they are
  meaningful (for the given ProcessType/RunningMode).
* add the AttibuteAccess.Flag.FORCE_REGISTRATION to override this
  behaviour and always force the registration of the metric. This is
  meant for metrics that are *always* relevant (eg MBean metrics that
  are computed by the JVM)
* add this flag to MBean metrics, service-config's status and
   socket-binding metrics
* add the ManagementResourceRegistration.Factory#forEnvironment  API to
   create a ManagementResourceRegistration aware of the process type and
   running mode.
* replace deprecated ManagementResourceRegistration.Factory calls
  (without process type/running mode) by the new ones in the tests

JIRA: https://issues.jboss.org/browse/WFCORE-1513

### DIFF
--- a/controller/src/main/java/org/jboss/as/controller/AbstractAttributeDefinitionBuilder.java
+++ b/controller/src/main/java/org/jboss/as/controller/AbstractAttributeDefinitionBuilder.java
@@ -414,6 +414,16 @@ public abstract class AbstractAttributeDefinitionBuilder<BUILDER extends Abstrac
     }
 
     /**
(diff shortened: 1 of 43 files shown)


jmesnil:
I've preserved the old  `MRR.Factory` API that creates MRR which always register the metrics.
Maybe with core 3.0, we could remove them as there are not called anymore and ensures the same behaviour from the MMR.

PR #1540, opened 2016-05-03
Use @FunctionalInterface where appropriate

ctomc:


controller/src/main/java/org/jboss/as/controller/operations/global/GlobalOperationHandlers.java
```diff
@@ -296,6 +296,7 @@ public void handleResult(OperationContext.ResultAction resultAction, OperationCo
         }
     }
 
+    @FunctionalInterface
     interface FilterPredicate {
```
dmlloyd: This nonpublic interface could probably be replaced with Predicate<ModelNode>.

core-model-test/framework/src/main/java/org/jboss/as/core/model/test/ModelHarmonizer.java
```diff
@@ -34,6 +34,7 @@
  *
  * @author Brian Stansberry (c) 2012 Red Hat Inc.
  */
+@FunctionalInterface
```
dmlloyd: Adding this in test classes is probably pointless.

controller/src/main/java/org/jboss/as/controller/operations/global/GlobalOperationHandlers.java
ctomc: good point. I will take a look at it

core-model-test/framework/src/main/java/org/jboss/as/core/model/test/ModelHarmonizer.java
ctomc: not really, it is where it is most useful, for shorting test code.

kabir:
This looks fine to me. I'll let Brian have the final say though

bstansberry:
This is OK, but I absolutely positively 100% without question do not want to see PRs putting a bunch of lambda expressions in core.

I hate lambda expressions. They simplify a low order one-time-only task, typing, and complicate a high order potentially repeatable task, understanding what a piece of code does. I don't consider the change made to QueryOperationHandler to be an improvement.

ctomc:
QueryOperationHandler is just an example how predicates can be utilized, it is not directly related to adding @FunctionalInterface but just something @dmlloyd pointed out we could use.

But said all that, I agree that marking certain interfaces with functional interface doesn't give a green light that we should refactor APIs just so they utilize latest lambda features.

I see this more as improvement for long term new development, but at most as a way to shorten code used in tests.

bstansberry:
Sounds good. I don't mind lambdas in tests at all.

PR #1541, opened 2016-05-03
[WFCORE-1513] Register metrics depends on the server type

jmesnil:
* register metrics in the ConcreteResourceRegistration only when they are
  meaningful (for the given ProcessType/RunningMode).
* add the AttibuteAccess.Flag.FORCE_REGISTRATION to override this
  behaviour and always force the registration of the metric. This is
  meant for metrics that are *always* relevant (eg MBean metrics that
  are computed by the JVM)
* add this flag to MBean metrics, service-config's status and
  socket-binding metrics
* add the ManagementResourceRegistration.Factory#forEnvironment  API to
  create a ManagementResourceRegistration aware of the process type and
  running mode.s type/running mode) by the new ones in the tests
* replace deprecated ManagementResourceRegistration.Factory calls
  (without process type/running mode) by the new ones in the tests

JIRA: https://issues.jboss.org/browse/WFCORE-1513

### DIFF
--- a/controller/src/main/java/org/jboss/as/controller/AbstractAttributeDefinitionBuilder.java
+++ b/controller/src/main/java/org/jboss/as/controller/AbstractAttributeDefinitionBuilder.java
@@ -414,6 +414,14 @@ public abstract class AbstractAttributeDefinitionBuilder<BUILDER extends Abstrac
     }
 
     /**
+     * Adds the {@link AttributeAccess.Flag#FORCE_REGISTRATION} flag.
+     *
+     * @return a builder that can be used to continue building the attribute definition
+     */
+    public BUILDER forceRegistration() {
+        return addFlag(AttributeAccess.Flag.FORCE_REGISTRATION);
+    }
+    /**
      * Adds the {@link AttributeAccess.Flag#RESTART_ALL_SERVICES} flag and removes any conflicting flag.
      *
      * @return a builder that can be used to continue building the attribute definition
--- a/controller/src/main/java/org/jboss/as/controller/operations/common/Util.java
+++ b/controller/src/main/java/org/jboss/as/controller/operations/common/Util.java
@@ -25,6 +25,7 @@ import static org.jboss.as.controller.descriptions.ModelDescriptionConstants.OP;
 import static org.jboss.as.controller.descriptions.ModelDescriptionConstants.OP_ADDR;
 import static org.jboss.as.controller.descriptions.ModelDescriptionConstants.OUTCOME;
 import static org.jboss.as.controller.descriptions.ModelDescriptionConstants.READ_ATTRIBUTE_OPERATION;
+import static org.jboss.as.controller.descriptions.ModelDescriptionConstants.READ_RESOURCE_DESCRIPTION_OPERATION;
 import static org.jboss.as.controller.descriptions.ModelDescriptionConstants.REMOVE;
 import static org.jboss.as.controller.descriptions.ModelDescriptionConstants.UNDEFINE_ATTRIBUTE_OPERATION;
 import static org.jboss.as.controller.descriptions.ModelDescriptionConstants.VALUE;
@@ -133,6 +134,11 @@ public class Util {
     public static ModelNode getReadAttributeOperation(final PathAddress address, String attributeName) {
         ModelNode op = createEmptyOperation(READ_ATTRIBUTE_OPERATION, address);
         op.get(NAME).set(attributeName);
+        return op;
+    }
+
+    public static ModelNode getReadResourceDescriptionOperation(final PathAddress address) {
+        ModelNode op = createEmptyOperation(READ_RESOURCE_DESCRIPTION_OPERATION, address);
         return op;
     }
 
--- a/controller/src/main/java/org/jboss/as/controller/registry/ConcreteResourceRegistration.java
+++ b/controller/src/main/java/org/jboss/as/controller/registry/ConcreteResourceRegistration.java
@@ -44,6 +44,7 @@ import org.jboss.as.controller.OperationDefinition;
 import org.jboss.as.controller.OperationStepHandler;
 import org.jboss.as.controller.PathAddress;
 import org.jboss.as.controller.PathElement;
+import org.jboss.as.controller.ProcessType;
 import org.jboss.as.controller.ProxyController;
 import org.jboss.as.controller.ResourceDefinition;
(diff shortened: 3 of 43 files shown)


controller/src/main/java/org/jboss/as/controller/registry/ManagementResourceRegistration.java
```diff
@@ -310,7 +311,31 @@
...
+         * to determine whether resource metrics are registered or not.
+         *
+         * If the {@code processType} id {@code null}, metrics are <em>always</em> registered.
+         *
+         * @param processType can be {@code null}
+
+         * @return a Factory which creates ManagementResourceRegistration that
+         * dynamically determine whether resource metrics are actually registered
+         */
+        public static Factory forServer(ProcessType processType) {
```
kabir: This is a bit of a nitpick, but personally, I don't like the name 'forServer'. When I saw the first usage, I got curious if there was a forDomain() method as well.

controller/src/main/java/org/jboss/as/controller/operations/common/Util.java
```diff
@@ -136,6 +137,12 @@ public static ModelNode getReadAttributeOperation(final PathAddress address, Str
         return op;
     }
 
+    public static ModelNode getReadResourceDescriptionOperation(final PathAddress address, String attributeName) {
+        ModelNode op = createEmptyOperation(READ_RESOURCE_DESCRIPTION_OPERATION, address);
+        op.get(NAME).set(attributeName);
```
kabir: This is not an attribute on r-r-d

controller/src/main/java/org/jboss/as/controller/registry/ConcreteResourceRegistration.java
```diff
@@ -79,14 +80,19 @@
 
     private final Lock readLock;
     private final Lock writeLock;
+    /**
+     * Can be {@code null}. In that case, the MMR will always register metrics.
+     */
+    final ProcessType processType;
```
kabir: Is that for unit testing or something else?

controller/src/main/java/org/jboss/as/controller/AbstractAttributeDefinitionBuilder.java
```diff
@@ -414,6 +414,16 @@ public BUILDER setStorageRuntime() {
     }
 
     /**
+     * Adds the {@link AttributeAccess.Flag#STORAGE_RUNTIME} flag and removes any conflicting flag.
+     *
+     * @return a builder that can be used to continue building the attribute definition
+     */
+    public BUILDER setStorageRuntime(boolean forceRegistration) {
+        setStorageRuntime();
+        return addFlag(AttributeAccess.Flag.FORCE_REGISTRATION);
```
kabir: Should this not happen only if forceRegistration is true?

controller/src/main/java/org/jboss/as/controller/registry/ManagementResourceRegistration.java
```diff
@@ -438,15 +468,23 @@ public static ManagementResourceRegistration create(final ResourceDefinition res
...
          */
+        @Deprecated
         public static ManagementResourceRegistration create(final ResourceDefinition resourceDefinition,
                                                             AccessConstraintUtilizationRegistry constraintUtilizationRegistry,
                                                             CapabilityRegistry registry) {
+            return DEFAULT.createRegistration(resourceDefinition, constraintUtilizationRegistry, registry);
+        }
+
+
+        public ManagementResourceRegistration createRegistration(final ResourceDefinition resourceDefinition,
```
kabir: javadoc

controller/src/main/java/org/jboss/as/controller/AbstractAttributeDefinitionBuilder.java
```diff
@@ -414,6 +414,16 @@ public BUILDER setStorageRuntime() {
     }
 
     /**
+     * Adds the {@link AttributeAccess.Flag#STORAGE_RUNTIME} flag and removes any conflicting flag.
+     *
```
bstansberry: The param needs javadoc.

controller/src/main/java/org/jboss/as/controller/AbstractAttributeDefinitionBuilder.java
```diff
@@ -414,6 +414,16 @@ public BUILDER setStorageRuntime() {
...
+     * @return a builder that can be used to continue building the attribute definition
+     */
+    public BUILDER setStorageRuntime(boolean forceRegistration) {
```
kabir: Actually, I am afraid I don't really like this parameter either. The reason is that .setStorageRuntime(false) will make it look like it turns it off. Can we just keep the original method and have an additional .forceRegistration() method or something like that instead?

controller/src/main/java/org/jboss/as/controller/AbstractAttributeDefinitionBuilder.java
kabir: See my next comment, I think we should remove this parameter

controller/src/main/java/org/jboss/as/controller/registry/ManagementResourceRegistration.java
jmesnil: I've renamed it to `forProcessType()`

controller/src/main/java/org/jboss/as/controller/operations/common/Util.java
jmesnil: fixed

controller/src/main/java/org/jboss/as/controller/registry/ManagementResourceRegistration.java
jmesnil: commented

controller/src/main/java/org/jboss/as/controller/AbstractAttributeDefinitionBuilder.java
jmesnil: documented

controller/src/main/java/org/jboss/as/controller/AbstractAttributeDefinitionBuilder.java
jmesnil: I removed this `setStorageRuntime(boolean)` method and added a `forceRegistration()`

jmesnil:
I've updated the PR based on the code review

PR #1545, opened 2016-05-04
WFCORE-1525 ageout-history operations reports failure if it was unabl…

spyrkob:
…e to remove some history files

https://issues.jboss.org/browse/WFCORE-1525

If ageout-history fails to delete some history files (eg. due to wrong permissions), it returns with 
```
{
  "outcome" => "failed",
  "failure-description" => "WFLYCTL0158: Operation handler failed: java.lang.IllegalStateException: WFLYPAT0048: Unable to     remove some of the history files.",
  "rolled-back" => true
} 
```
In addition offending files are logged at warning level in the patching logger:
```
[33m22:12:47,080 WARN  [org.jboss.as.patching] (management-handler-thread - 3) WFLYPAT0001: Cannot delete file /Users/spyrkob/workspaces/set/wildfly-core/testsuite/patching/target/wildfly-core/.installation/patches/cp0/configuration
[33m22:12:47,081 WARN  [org.jboss.as.patching] (management-handler-thread - 3) WFLYPAT0001: Cannot delete file /Users/spyrkob/workspaces/set/wildfly-core/testsuite/patching/target/wildfly-core/.installation/patches/cp0/timestamp
```

### DIFF
--- a/patching/src/main/java/org/jboss/as/patching/logging/PatchLogger.java
+++ b/patching/src/main/java/org/jboss/as/patching/logging/PatchLogger.java
@@ -238,4 +238,11 @@ public interface PatchLogger extends BasicLogger {
 
     @Message(id = 47, value = "Cannot copy files from %s to %s: %s")
     IOException cannotCopyFiles(String from, String to, String reason, @Cause Throwable cause);
+
+    @LogMessage(level = ERROR)
+    @Message(id = 48, value = "Error when restoring file[%s] - %s")
+    void deleteRollbackError(String path, String message);
+
+    @Message(id = 49, value = "Some backup files were not removed.")
+    IOException failedToDeleteBackup();
 }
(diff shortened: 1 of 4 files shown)


spyrkob:
retest this please

patching/src/main/java/org/jboss/as/patching/logging/PatchLogger.java
```diff
@@ -238,4 +238,7 @@
 
     @Message(id = 47, value = "Cannot copy files from %s to %s: %s")
     IOException cannotCopyFiles(String from, String to, String reason, @Cause Throwable cause);
+
+    @Message(id = 48, value = "Unable to remove some of the history files.")
+    String failedToRemoveHistoryFiles();
```
bstansberry: This can return IllegalStateException, and then the callers can just do this:

throw PatchLogger.ROOT_LOGGER.failedToRemoveHistoryFiles();

bstansberry:
@aloubyansky The nit I mentioned above aside, this looks ok to me. Does it look ok to you?

aloubyansky:
We chatted with @spyrkob and he is going to try to improve the PR by also implementing a proper undo for the files that have been deleted before throwing an error to the caller.

spyrkob:
@aloubyansky could you take a look at the updated version? It's now trying to delete both the layer and history files together rolling back both in case of an error.

aloubyansky:
This is a great contribution. Thanks, Bartosz!

PR #1546, opened 2016-05-05
Fix corner case where startup task doesn't use provided client but static injected one

ctomc:
… static injected one


jamezp:
Just adding my comment here for other reviewers.

I think we need to attack this issue from the `Server`. Something like https://github.com/wildfly/wildfly-core/commit/faafb49d660649e12a5cfb6cc8503310f9750d5a would resolve this issue with no changes to the test runner.

ctomc:
will open new PR with changes proposed by @jamezp

PR #1549, opened 2016-05-05
Improve test runner to be more resiliant

ctomc:
- use junit api instead of pure reflection for injection
- make ManagmentClient final in server
- use client delegate to reduce injection and "stale" clients

### DIFF
--- a/testsuite/test-runner/src/main/java/org/wildfly/core/testrunner/Server.java
+++ b/testsuite/test-runner/src/main/java/org/wildfly/core/testrunner/Server.java
@@ -1,5 +1,13 @@
 package org.wildfly.core.testrunner;
 
+import static org.jboss.as.controller.client.helpers.ClientConstants.NAME;
+import static org.jboss.as.controller.client.helpers.ClientConstants.OP;
+import static org.jboss.as.controller.client.helpers.ClientConstants.OP_ADDR;
+import static org.jboss.as.controller.client.helpers.ClientConstants.READ_ATTRIBUTE_OPERATION;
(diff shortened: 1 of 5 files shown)


testsuite/test-runner/src/main/java/org/wildfly/core/testrunner/Server.java
```diff
@@ -131,8 +133,6 @@ protected void start() {
             shutdownThread = ProcessHelper.addShutdownHook(proc);
 
 
-            client = createClient();
```
jamezp: This should be replaced by {{ServerClientProvider.INSTANCE.setClient(createModelControllerClient())}}. Maybe ensure any previous client was closed as well. If a {{Server.stop()}} is executed followed by a {{Server.start()}} a new client needs to be created.

testsuite/test-runner/src/main/java/org/wildfly/core/testrunner/Server.java
```diff
@@ -59,7 +61,7 @@
     private Thread shutdownThread;
 
     private volatile Process process;
-    private volatile ManagementClient client;
+    private final ManagementClient client = createClient();
```
jamezp: I almost wonder if this should just be `new ManagementClient(new DelegatingModelControllerClient(ServerClientProvider.INSTANCE), managementAddress, managementPort, managementProtocol)` so initially no `ModelControllerClient` is created. Essentially only leaving the creation of the MCC to `start()` and `reload()` operations.

ctomc:
@jamezp fixed as you suggested, I also removed the createNewClient() method from the ServerController as is not really needed. 
We can add it back if such need ever arises.

PR #1552, opened 2016-05-07
[WFCORE-1535] Avoid a potential blocking forever for ManagedServer suspend() and resume()

yersan:
Basically this PR continues previous changes done in WFCORE-1473 applying them to other suspend and resume operations in ManagedServer.

Jira issue is: https://issues.jboss.org/browse/WFCORE-1535

Implementation note:
It looks like that ServerInventory#suspendServer(String serverName) does 99% the same as ServerInventory#awaitServerSuspend(Set<String> waitForServers, int timeoutInSeconds, BlockingTimeout blockingTimeout) using a timeoutInSeconds = 0. So in this issue there is a proposal to simplifying them.

The small difference is the following. Internally ServerInventory#suspendServer(String serverName) is executing this operation using the ManagedServer#suspend():

```
final ModelNode operation = new ModelNode();
operation.get(OP).set(SUSPEND);
operation.get(OP_ADDR).setEmptyList();
```

and ServerInventory#awaitServerSuspend(Set<String> waitForServers, int timeoutInSeconds, BlockingTimeout blockingTimeout) is executing this operation using the ManagedServer#suspend(timeoutInSeconds):

```
final ModelNode operation = new ModelNode();
operation.get(OP).set(SUSPEND);
operation.get(OP_ADDR).setEmptyList();
operation.get(TIMEOUT).set(timeoutInSeconds);
```

If there is no TIMEOUT attribute in the operation sent to ServerSuspendHandler, it will use 0 for the suspend timeout. 
With the changes in this issue when :suspend-servers (without any explicity timeout) is executed, now the TIMEOUT attribute is sent with the operation with 0 as value simplifying the code (I think) in ServerInventoryImpl. I cannot see any problems with this difference behaviour. Is it correct for you too?

### DIFF
--- a/host-controller/src/main/java/org/jboss/as/domain/controller/operations/DomainServerLifecycleHandlers.java
+++ b/host-controller/src/main/java/org/jboss/as/domain/controller/operations/DomainServerLifecycleHandlers.java
@@ -340,7 +340,7 @@ public class DomainServerLifecycleHandlers {
             context.readResource(PathAddress.EMPTY_ADDRESS, false);
             final ModelNode model = Resource.Tools.readModel(context.readResourceFromRoot(PathAddress.EMPTY_ADDRESS, true));
             final String group = getServerGroupName(operation);
-            final int suspendTimeout = TIMEOUT.resolveModelAttribute(context, operation).asInt();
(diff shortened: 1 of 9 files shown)


bstansberry:
Excellent! Thanks! +1 re: how you simplified this.

PR #1553, opened 2016-05-09
WFCORE-1405 Be able to register notification handlers without a subsystem

kabir:
https://issues.jboss.org/browse/WFCORE-1405


controller/src/main/java/org/jboss/as/controller/notification/NotificationRegistry.java
```diff
@@ -0,0 +1,41 @@
...
+package org.jboss.as.controller.notification;
+
+import org.jboss.as.controller.PathAddress;
+
+/**
+ * Interface exposed to the users to register notifications via the {@link NotificationRegistrar}
+ *
+ * @author Kabir Khan
+ */
+public interface NotificationRegistry {
```
bstansberry: ModelController's 

NotificationHandlerRegistration getNotificationRegistry()

is really poorly named. :(

The method name is reasonable but the return type is kind of yuck.

We could perhaps make NotificationHandlerRegistration a subinterface of NotificationRegistry since they share the one method.

This comment isn't really about your PR here; it's more me not liking something in what we have already and fishing for thoughts about how to make it better.  :)

bstansberry:
WDYT about having the properties just be an attribute instead of child resources? With the API we now have for manipulating map attributes I don't see the benefit of child resources.

kabir:
Retest this please

domain-management/src/main/java/org/jboss/as/domain/management/notification/registrar/NotificationRegistrarResourceDefinition.java
```diff
@@ -0,0 +1,189 @@
...
+        final Map<String, String> properties;
+        if (!propsNode.isDefined()) {
+            properties = Collections.emptyMap();
+        } else {
+            properties = new HashMap<>();
+            for (String key : propsNode.keys()) {
+                String value = propsNode.hasDefined(key) ? propsNode.get(key).asString() : null;
+                properties.put(key, value);
+            }
+        }
```
bstansberry: Once I fix https://issues.jboss.org/browse/WFCORE-1557 in a minute you can use Properties.unwrap(context, model) to replace L109-119.

kabir:
Retest this please

kabir:
Retest this please

controller/src/main/java/org/jboss/as/controller/notification/NotificationRegistrar.java
```diff
@@ -0,0 +1,43 @@
...
+ */
+public interface NotificationRegistrar {
+
+    /**
+     * Registers notification handlers. The notification handlers will be automatically removed when this registrar
+     * is removed.
+     *
+     * @param context the context
+     */
+    void registerNotificationListeners(NotificationRegistrarContext context);
```
jmesnil: for consistency, it should be renamed to registerNotificationHandlers()

bstansberry:
I put Hold on this while we do some rethink to make sure this is the right overall direction.

bstansberry:
Since @jmesnil is taking this over I'm closing this to clean out the queue.

PR #1554, opened 2016-05-09
WFCORE-1540 Expose capability registry via mgmt api

ctomc:


ctomc:
@hpehl this is what we discussed, just bit more cleaned up code

server/src/main/java/org/jboss/as/server/controller/resources/CapabilityRegistryResourceDefinition.java
```diff
@@ -48,58 +53,143 @@
...
+    private static final StringListAttributeDefinition POSSIBLE_REGISTRATION_POINTS = new StringListAttributeDefinition.Builder("registration-points")
+                .build();
+
+    private static final SimpleAttributeDefinition ADDRESS = SimpleAttributeDefinitionBuilder.create("address", ModelType.STRING, false)
+                .build();
+    private static final SimpleAttributeDefinition ATTRIBUTE = SimpleAttributeDefinitionBuilder.create("attribute", ModelType.STRING, true)
             .build();
 
-    private static final SimpleAttributeDefinition DYNAMIC = SimpleAttributeDefinitionBuilder.create("dynamic", ModelType.BOOLEAN, false)
+    private static final ObjectTypeAttributeDefinition REGISTRATION_POINT = new ObjectTypeAttributeDefinition.Builder("registration-point", ADDRESS, ATTRIBUTE )
```
bstansberry: The registration point for a capability should only be the address, as we determined (on a wildfly-dev thread IIRC) that only a resource can provide a capability, not just an attribute. That's why OperationContext.registerCapability(RuntimeCapability, String) is deprecated and not used in core or full.

The registration point for a requirement can involve an attribute (e.g. a model ref) but what you're displaying here is capabilty registration points, not requirements.

ctomc: fixed

PR #1557, opened 2016-05-11
[WFCORE-1546] whitespaces in the middle of value probably should not …

soul2zimate:
…be ignored.

https://issues.jboss.org/browse/WFCORE-1546
see more discussion on the JBEAP issue inside


soul2zimate:
failing PassivationTestCase and LdapExtPasswordCachingTestCase seem irrelevant.

bstansberry:
@aloubyansky please approve

aloubyansky:
I know it's one line fix but I'd like to see a simple testcase for the issue, e.g. in OperationParsingTestCase in the CLI module. And while we are here, please, do the same PropertyState class. Thanks!

soul2zimate:
I have added a simple test to cover case there is whitespaces in the middle of property value.

About PropertyState, there is an exceptional case that "rollout" command takes whitespace as propSeparator in RolloutPlanState

I noticed that `putHandler(propSeparator, GlobalCharacterHandlers.LEAVE_STATE_HANDLER); `has been recently added to [PropertyState](https://github.com/wildfly/wildfly-core/blob/ffdb782f7823bf8364f294d7d5cf66ebbe119e53/cli/src/main/java/org/jboss/as/cli/parsing/operation/PropertyState.java#L68) in order to accept property without a value

Thus, the setIgnoreWhitespaces(true); actually avoid to be handled by GlobalCharacterHandlers.LEAVE_STATE_HANDLER in such case.

aloubyansky:
Thanks @soul2zimate! This is good to merge.

PR #1560, opened 2016-05-13
[WFCORE-1550]: wildfly-init-redhat.sh script does not work on RHEL 5 OS

vpakan:
JIRA: https://issues.jboss.org/browse/WFCORE-1550


bstansberry:
this is ok to test

PR #1562, opened 2016-05-13
Read the runtime attributes in Stage.RUNTIME which is when they would…

bstansberry:
… be read in a real case

This relates to ongoing work on WFCORE-1481, just trying to make the test better aligned with real world use cases.


bstansberry:
I'm closing this because I don't think this test is actually trying to simulate a runtime-only attribute. It's just validating that a normal Storage.CONFIGURATION attribute with a handler that updates the runtime works properly, all the way through to updating the runtime.

PR #1564, opened 2016-05-16
[master] WFCORE-1554 Building an attribute based on another attribute does not copy CapabilityReferenceRecorder

pferraro:
https://issues.jboss.org/browse/WFCORE-1554


ctomc:
> This was probably an oversight, since a protected accessor method for a protected field is pointless.

CapabilityRecorder shouldn't be exposed, as it is current implementation detail. Not a public API subsystems should build against.
You can set your own custom CapabilityRecorder impl in builder, but I don't think we should expose it on *AttributeDefinition.

pferraro:
@ctomc I already use a custom CapabilityReferenceRecorder implementation - but in this case, I want to also override the circumstances when it gets called, specifically, the code in AttributeDefinition.addCapabilityRequirements(...) and removeCapabilityRequirements(...).  Sounds like I'll need a custom AttributeDefinition to do this then.

PR #1568, opened 2016-05-19
[WFCORE-1536]:  NPE thrown during application redeployment, slaves taken offline

ehsavoie:
Protecting loop against File.list() that may return a null array.

Jira: https://issues.jboss.org/browse/WFCORE-1536


ehsavoie:
retest this please

PR #1575, opened 2016-05-24
WFCORE-1130 Add support for an ObjectMapAttributeDefinition

ctomc:


ctomc:
@darranl here you go.

controller/src/main/java/org/jboss/as/controller/ObjectMapAttributeDefinition.java
```diff
@@ -0,0 +1,170 @@
...
+            super(name);
+            this.valueType = valueType;
+            setElementValidator(valueType.getValidator());
+            setAttributeParser(AttributeParsers.OBJECT_MAP_WRAPPED);
+            setAttributeMarshaller(AttributeMarshaller.OBJECT_MAP_MARSHALLER);
+        }
+
+        public static ObjectListAttributeDefinition.Builder of(final String name, final ObjectTypeAttributeDefinition valueType) {
+            return new ObjectListAttributeDefinition.Builder(name, valueType);
+        }
```
bstansberry: This should return ObjectMapAttributeDefinition.Builder, not ObjectList...

bstansberry:
@darranl IIRC you were the driver for this enhancement. Does this look like what you wanted?

Some test links that give you a sense:

https://github.com/wildfly/wildfly-core/pull/1575/files#diff-dadbae2ed567e9d636a3418915dc447dR243
https://github.com/wildfly/wildfly-core/pull/1575/files#diff-bd0359fe391dd52f43fa6e39fc89727dR4

The minor thing I noted above, this seems ok to me.

darranl:
+1 that does look like what we were looking for.

controller/src/main/java/org/jboss/as/controller/ObjectMapAttributeDefinition.java
ctomc: ups :) fixed.

ctomc:
@bstansberry @darranl do you guys maybe want this also for 2.x branch. If so i can prepare PR.

darranl:
All my work is on 3 only so not for me thank you.

bstansberry:
@ctomc Let's stick to 3.x.

PR #1577, opened 2016-05-24
[master] WFCORE-1568 AttributeDefinition should defer handling of undefined values to CapabilityReferenceRecorder

pferraro:
https://issues.jboss.org/browse/WFCORE-1568

N.B. This commit will cause the core+full integration tests to fail until https://github.com/wildfly/wildfly/pull/8914 is merged.

### DIFF
--- a/controller/src/main/java/org/jboss/as/controller/AttributeDefinition.java
+++ b/controller/src/main/java/org/jboss/as/controller/AttributeDefinition.java
@@ -889,13 +889,10 @@ public abstract class AttributeDefinition {
      */
     public void addCapabilityRequirements(OperationContext context, ModelNode attributeValue) {
         if (referenceRecorder != null) {
-            if (!attributeValue.isDefined()) {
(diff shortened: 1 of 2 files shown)


bstansberry:
Please update the CapabilityReferenceRecorder javadoc to specify the param can be null and that it is permissible to do nothing in that case.

I should have specified that kind of thing when I wrote the API, although in this case it seems not doing so is a benefit!

@darranl will this break any elytron implementations?

darranl:
I don't think it should break anything for us - any places where we must resolve a capability the attribute will not support undefined anyway.

pferraro:
@bstansberry Will do.  Just to say, the only external impact of this change are custom CapabilityReferenceRecorder implementations (hence the failure in the wildfly integration run), which is handled in https://github.com/wildfly/wildfly/pull/8914

pferraro:
Requisite downstream have been merged.

bstansberry:
retest this please

PR #1585, opened 2016-05-27
[WFCORE-1565] Upgrade JBoss Remoting to 4.0.21.Final

dmlloyd:

PR #1586, opened 2016-05-27
[2.x] [WFCORE-1565] Upgrade JBoss Remoting to 4.0.21.Final

dmlloyd:


bstansberry:
Note to self: #1585 is the master equivalent.

PR #1589, opened 2016-05-29
WFCORE-1427: Added timeout attribute to HC reload operation allowing the graceful shutdown timeout when the servers are restarted using this reload operation

yersan:
With these changes the user is able to configure a timeout value in the HC reload for the shutdown operation triggered over the managed servers.
The shutdown operation is triggered only if restart-servers is true.

Jira issue is:
https://issues.jboss.org/browse/WFCORE-1427


host-controller/src/main/java/org/jboss/as/domain/controller/operations/HostProcessReloadHandler.java
```diff
@@ -76,10 +76,13 @@
             .setAlternatives(ModelDescriptionConstants.USE_CURRENT_DOMAIN_CONFIG)
             .build();
 
+    protected static final AttributeDefinition TIMEOUT = new SimpleAttributeDefinitionBuilder(ModelDescriptionConstants.TIMEOUT, ModelType.INT, true)
+            .setDefaultValue(new ModelNode(0)).build();
```
yersan: I think for this case having a 0 by default for the timeout could not be great idea because after this PR, the user is not able to configure different suspend values for a reload operation. That means if we use 0 as default, doing something like this (suppose we are in a domain mode with on HC and two managed servers):
`/host=master/server-config=server-one:suspend(timeout=60)`
`/host=master/server-config=server-one:suspend(timeout=120)`
`/host=master:reload`
Last reload will automatically overwrite the previous suspend configured timeout with 0. 

The user can use now:
`/host=master:reload(timeout=120)`
But he is not longer able to do the reload with different suspend timeouts configured for each managed server

kabir: Does the suspend operation block? I don't remember off the top of my head.

If it does, then won't the two server-one suspends will finish before the master reload. If not, I see your point :)

yersan: > Does the suspend operation block?

Thanks for the tip!, yes, although the HC reload operation response is received immediately, the HC is restarted when all managed servers have been stopped, with the changes in this issue, the suspend operation blocks and later stops the managed servers. So, yes, you're right, independently of the previous suspend configurations done for managed servers, reload will always wait until all servers are stopped. That means (I think) it makes no sense configure different suspends for managed servers if you are going to use a reload HC. If that reload without timeout should overwrite any previous suspend with 0 timeout, is still a minor doubt for me.

Additionally, suspend operation used in this PR should be using BlockingQueueOperationListener as you did in #1490 ,that something still pending to do here, I will change it.

I have other doubts about if suspend operation is working fine anyway, I will write those thoughts in the Jira issue instead here.

yersan:
Now it seems that suspend operation is working as expected for this issue.

yersan:
Failed test on Linux platform shouldn't be related with this changes.
Retest this please

yersan:
Retest this please

host-controller/src/main/java/org/jboss/as/domain/controller/operations/HostProcessReloadHandler.java
bstansberry: I need to triple check but I don't believe the suspend op blocks waiting for the server to suspend. Actually, looking at the ServerSuspendHandler code I suspect it's quite broken. Hope I'm wrong though.

Looking how ServerSuspendHandler is written though I think the intent was to have some sort of blocking behavior, but I don't think that's what really happens.

bstansberry: FYI https://issues.jboss.org/browse/WFCORE-1616 is the ServerSuspendHandler problem.

yersan: Sorry, in fact it was an oversight from my side don't comment it here. My initial assumption was wrong. The suspend operation doesn't block for a server shutdown, it was my mistake in my previous comment. For that reason I applied some additional changes which are in commit 8322b20 to allow the new timeout in the reload operation working fine with its suspend operation.

Relevant lines about why the server reloads are not blocking due to a previous suspend executed in other operation are:
GracefulShutdownService L82 --> awaitSuspend method needs the shuttingDown flag as true
DomainServerMain L171 --> at this point the shutdown waits until all activities has been finished or until a time out.

That means if you execute these operation with the code in master branch:
`/host=master/server-config=server-two:suspend(timeout=120)`
`/host=master:reload`
The reload won't block

bstansberry: WFCORE-1616 has been fixed; see #1630. So now a call that hits ServerSuspendHandler will not complete until the suspend is done.

yersan:
Added reload operation with timeout in standalone mode

bstansberry:
This all looks good. I would like @jfdenise to have a look at it though, in terms of the CLI side. A double check of the impl but more in terms of the user experience. He has been working a general task of having client-side timeouts for all CLI operations. That's a completely different thing than this, but I'd like his input about how to make sure that users don't get confused about the different things. Perhaps calling the CLI param --suspend-timeout or something similar would help. Or maybe he won't think it's an issue.

The only thing that makes me a tiny bit nervous about this is the possibility of the HostRunningModeControl ending up keeping the timeout around beyond the intended period, i.e. something prevents it getting set back to zero. And then something else causes ServerInventoryService.stop() to run and it reads the old value. Possibly it could be implemented as a one-time-read (e.g. store it in an AtomicInteger and the getter returns timeout.getAndSet(0);) But that might be fragile too if code needs to read it twice for a single request.

Great job on this!

jfdenise:
Having the option named --shutdown-servers-timeout would make it possibly clearer that it is only used when restarting servers in domain mode.
In term of implementation I am seeing 4 things that should be addressed:
- The option shouldn't be exposed by completion if we are not in domain mode. 
- Some checks (and associated exceptions) should be added when the option is used in standalone mode. This would help users to properly setup the command.
- The timeout should only be added to the request in domain mode. Right?
- Shouldn't you check that --timeout is used in conjonction with --restart-servers? Throwing an exception if that is not the case would help users properly use the command.

yersan:
Hello @jfdenise, thank you for your feedback and name suggestion for the timeout attribute,

Initially it was required to add the timeout operation for domain mode only, but later it was requested to add it to standalone mode too. I think it makes sense in standalone mode, if I'm not wrong the reload operation stops and restarts the root standalone service even loading new configuration based in changes applied to standalone.xml so it affects to availability to handle external requests. A timeout in this mode allows the graceful shutdown of deployed applications in the server.

Adding the timeout in standalone mode was not commented in the Jira issue, maybe it was somewhat confusing, let me add some comments to your previous inputs:
> The option shouldn't be exposed by completion if we are not in domain mode. 

Currently the completion works in standalone because the timeout attribute is available in domain and standalone mode, that's the reason.

> Some checks (and associated exceptions) should be added when the option is used in standalone mode. This would help users to properly setup the command.

Could you explain me a little more this point? What do you mean with Some checks (and associated exceptions)?

> The timeout should only be added to the request in domain mode. Right?

Currently it is added to standalone, reason commented above.

> Shouldn't you check that --timeout is used in conjunction with --restart-servers? Throwing an exception if that is not the case would help users properly use the command.

Yes, I think I could add this check for domain mode.

bstansberry:
Note that this timeout isn't really a "shutdown servers timeout". It's a timeout for *suspending* the server(s) prior to a subsequent action. If the target of the request is a standalone server, the subsequent action is reloading the server. If the target of the request is a Host Controller, the subsequent action on each server is a shutdown. But if the target is an HC, the timeout is only relevant if the --restart-servers param is not set or is set to true. Otherwise the servers are not impacted by the HC reload.

jfdenise:
I started by reading the help changes. This made me think that the timeout was only used in domain mode when --restart-servers was present. You should fix the synopsis.

Then I would vote for --suspend-timeout. 

Perhaps that the description provided by @bstansberry could be partly reused to document the option.

@yersan "Some checks (and associated exceptions)". Checking that timeout option was not present in standalone mode and throwing an exception if present. But that is not needed anymore.
Thanks.

bstansberry:
@jfdenise Agreed, the help text can be clearer. And it wasn't my intent when I wrote my last comment, but re-reading it I do think a lot of it can be used in the help text. Thanks for looking at this, it's been very helpful.

jfdenise:
yes, in particular the synopsis that makes it clear that it is domain only.

yersan:
Let me add some additional details found working with this issue:

1. It is still pending to add suspend-timeout at host controller's server view reload operation. That means, add suspend-timeout to:
`[domain@localhost:9990 /] /host=master/server-config=server-one:reload()`

2. It is still pending to add suspend-timeout to :reload-servers operation:
`[domain@localhost:9990 /] :reload-servers`

3. Maybe it could be useful add suspend-timeout to restart operations too, that means, add suspend-timeout to:
`[domain@localhost:9990 /] :restart-servers`
`[domain@localhost:9990 /] /host=master/server-config=server-one:restart`

4. This is just a minor note regarding general operation and attribute names, I'm wondering if you think these attributes should be renamed from timeout to suspend-timeout too, although maintain a backward compatibility is important, I would expect to have consistences for operation names and attributes
`[domain@localhost:9990 /] shutdown --host=master --timeout=10`
`[standalone@localhost:9990 /] shutdown --timeout=10`
`[domain@localhost:9990 /] :stop-servers(timeout=10)`
`[domain@localhost:9990 /] host=master/server-config=server-one:stop(timeout=10)`
Maybe even renaming stop by shutdown, please consider it just as a simple note.

I think 1 and 2 are still pending for this issue, waiting for feedback to know if you want address them together with this issue or you prefer open a new one.

bstansberry:
@yersan I think 1-3 are fine to include with this issue. 1 and 2 for sure; they are all aspects of reload. I think expanding the scope to cover the restart op is ok too, although I don't mind using a separate issue.

Re: #4 that's a valid point. But we can't rename the existing attributes or operations as that would be an incompatible API change. At most we could introduce new operations/attributes. I don't want to do that as part of this work. I don't think I'd want to deal with "shutdown" vs "stop" at all, as having two ops that do the same thing both with intuitive names would be kind of confusing too. Maybe at some point we should reconsider the name of the 'timeout' param for shutdown though as its meaning is unclear. It's not as unclear as 'timeout' would be in a reload or restart case though. A reload can take a long time for reasons other than waiting to suspend, e.g. deployments can take a long time to deploy.

bstansberry:
Re: 3) perhaps the issue is just that the operation description doesn't include the 'timeout' param. DomainServerLifecycleHandlers.RestartServersLifecycleHandler actually processes a 'timeout' param.

It looks like DomainServerLifecycleHandlers.ReloadServersLifecycleHandler was also meant to support timeout but doesn't at all. The LocalDescriptions.properties file includes this entry which unless I'm way wrong isn't actually used anywhere:

```
domain.reload-servers.timeout=The graceful shutdown timeout. If this is zero then a graceful shutdown will not be attempted, if this is -1 then the server will wait for a graceful shutdown indefinitely.
```

For that one even though there's a LocalDescriptions.properties entry for 'timeout' we should use 'suspend-timeout' and change the entry. It's never been possible to use 'timeout' so there's no backward compatibility concern.

bstansberry:
FYI, since I filed this JIRA and you started working on it, we've introduced a new process where for significant features we start with a formal requirements analysis document. That's good in general, and for stuff that is going to end up in EAP where we want the Red Hat Quality Engineering team to verify the feature, it's really helpful. Since this is a pretty substantial change and QE verification is a good idea I decided to "go back in time" and create an requirements doc:  https://developer.jboss.org/docs/DOC-55724

You've thought harder about this than anyone, so if you see something wrong there or feel something should be added, please feel free to edit the document.

host-controller/src/main/java/org/jboss/as/host/controller/ManagedServer.java
```diff
@@ -814,39 +844,68 @@ public boolean execute(ManagedServer server) throws Exception {
...
+        private final int gracefulTimeout;
+        private final BlockingTimeout blockingTimeout;
 
-        private ServerStopTask(int permit, int timeout) {
-            this.permit = permit;
-            this.timeout = timeout;
+        private ServerStopTask(int gracefulTimeout) {
+            this.gracefulTimeout = gracefulTimeout;
+            this.blockingTimeout = null;
+        }
```
yersan: An alternative here instead of null is make BlockingTimeoutImpl public and just use BlockingTimeoutImpl(null)

host-controller/src/main/java/org/jboss/as/host/controller/ManagedServer.java
```diff
@@ -700,6 +711,25 @@ private static InternalState nextState(final InternalState state, final Internal
...
+
+        return protocolClient.execute(listener, operation, OperationMessageHandler.DISCARD, OperationAttachments.EMPTY);
+    }
+
+    private AsyncFuture<OperationResponse> reload(int timeoutInSeconds, final BlockingQueueOperationListener<TransactionalProtocolClient.Operation> listener, int operationID) throws IOException {
+        final ModelNode operation = new ModelNode();
+        operation.get(OP).set(RELOAD);
+        operation.get(OP_ADDR).setEmptyList();
+        operation.get(SUSPEND_TIMEOUT).set(timeoutInSeconds);
+        operation.get("operation-id").set(operationID);
```
yersan: I'm not really sure yet if "operation-id" has an utility here.

host-controller/src/main/java/org/jboss/as/host/controller/ManagedServer.java
```diff
@@ -872,26 +931,41 @@ public boolean execute(ManagedServer server) throws Exception {
...
-                final TransactionalProtocolClient.PreparedOperation<?> prepared = TransactionalProtocolHandlers.executeBlocking(operation, protocolClient);
+                future = reload(gracefulTimeout, listener, permit);
+                int blockingTimeoutValue = blockingTimeout.getProxyBlockingTimeout(server.getAddress(), server.getProxyController());
+                final TransactionalProtocolClient.PreparedOperation<?> prepared = listener.retrievePreparedOperation(blockingTimeoutValue, TimeUnit.MILLISECONDS);
+
                 if (prepared.isFailed()) {
                     return false;
                 }
+
                 prepared.commit(); // Just commit and discard the result
```
yersan: Does it make sense just commit and discard the result?

host-controller/src/main/java/org/jboss/as/host/controller/ManagedServer.java
yersan: The use of this "operation-id" has been clear up by @bstansberry in https://issues.jboss.org/browse/WFCORE-1791. The "operation-id" here makes sense and it is working fine. It should not be removed from reload op.

yersan:
Closing for now, finally, we are going to use a different approach. More details here: https://github.com/wildfly/wildfly-proposals/pull/54

PR #1590, opened 2016-05-30
[WFCORE-1474] CLI version command should not use JBoss AS in its outp…

aloubyansky:
…ut field names

This commit removes "JBoss AS" from the headers of the displayed info table.

This may potentially break scripts or applications analyzing the output and looking for specific fields.

Alexey

### DIFF
--- a/cli/src/main/java/org/jboss/as/cli/handlers/VersionHandler.java
+++ b/cli/src/main/java/org/jboss/as/cli/handlers/VersionHandler.java
@@ -67,7 +67,7 @@ public class VersionHandler implements CommandHandler {
         final StringBuilder buf = new StringBuilder();
         buf.append("JBoss Admin Command-line Interface\n");
         buf.append("JBOSS_HOME: ").append(WildFlySecurityManager.getEnvPropertyPrivileged("JBOSS_HOME", null)).append('\n');
-        buf.append("JBoss AS release: ");
+        buf.append("Release: ");
         final ModelControllerClient client = ctx.getModelControllerClient();
         if(client == null) {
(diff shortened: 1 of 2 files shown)


kabir:
I'll defer this to Brian :) The main worry my side is this making its way into EAP 7.1, i.e. a breaking change in a minor.

bstansberry:
It will definitely end up in EAP 7.x if it's merged. There is no "EAP feature branch" of core; this is it.

bstansberry:
@rsvoboda WDYT about this in 7.x? I'm on the fence. It's not critical, but the change is better going forward and it's a pain to defer to EAP 8.

cli/src/main/java/org/jboss/as/cli/handlers/VersionHandler.java
```diff
@@ -113,6 +112,7 @@ public void handle(CommandContext ctx) throws CommandFormatException {
                 throw new CommandFormatException("Failed to get the AS release info: " + e.getLocalizedMessage());
             }
         }
+        buf.append("JBOSS_HOME: ").append(WildFlySecurityManager.getEnvPropertyPrivileged("JBOSS_HOME", null)).append('\n');
```
bstansberry: Should this move back up so we don't introduce an ordering change?

aloubyansky: Moved it back up and rebased.

rsvoboda:
@bstansberry I think this is ok

If someone mysteriously depends on output of version command they should use :read-attribute(name=release-version) or :read-attribute(name=product-version)

PR #1592, opened 2016-06-01
Systemd example

BenjaminLefoul:
Original examples were better WFLY-4570 and /usr/share is actually packaged on fedora (wildfly 8: https://apps.fedoraproject.org/packages/wildfly/contents/ ) and used in wildfly 10 in COPR: https://github.com/goldmann/wildfly-copr

Regardless, these are just examples.

Additionally, it can be useful to activate the httpd.socket if it exists.

### DIFF
--- a/core-feature-pack/src/main/resources/content/docs/contrib/scripts/init.d/wildfly-init-debian.sh
+++ b/core-feature-pack/src/main/resources/content/docs/contrib/scripts/init.d/wildfly-init-debian.sh
@@ -59,7 +59,7 @@ fi
 
 # Location of wildfly
 if [ -z "$JBOSS_HOME" ]; then
-	JBOSS_HOME="/opt/wildfly"
+	JBOSS_HOME="/usr/share/wildfly"
 fi
 export JBOSS_HOME
 
--- a/core-feature-pack/src/main/resources/content/docs/contrib/scripts/systemd/README
+++ b/core-feature-pack/src/main/resources/content/docs/contrib/scripts/systemd/README
@@ -3,21 +3,21 @@
 == Create a wildfly user
 
     # groupadd -r wildfly
-    # useradd -r -g wildfly -d /opt/wildfly -s /sbin/nologin wildfly
+    # useradd -r -g wildfly -d /usr/share/wildfly -s /sbin/nologin wildfly
 
 == Install WildFly
 
-    # tar xvzf wildfly-10.0.0.Final.tar.gz -C /opt
-    # ln -s /opt/wildfly-10.0.0.Final /opt/wildfly
-    # chown -R wildfly:wildfly /opt/wildfly
+    # tar xvzf wildfly-10.0.0.Final.tar.gz -C /var/tmp
+    # ln -s /var/tmp/wildfly-10.0.0.Final /usr/share/wildfly
+    # chown -R wildfly:wildfly /usr/share/wildfly
 
 == Configure systemd
 
     # mkdir /etc/wildfly
     # cp wildfly.conf /etc/wildfly/
     # cp wildfly.service /etc/systemd/system/
-    # cp launch.sh /opt/wildfly/bin/
-    # chmod +x /opt/wildfly/bin/launch.sh
+    # cp launch.sh /usr/share/wildfly/bin/
+    # chmod +x /usr/share/wildfly/bin/launch.sh
 
 == Start and enable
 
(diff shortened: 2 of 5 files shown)


ctomc:
this is ok to test

core-feature-pack/src/main/resources/content/docs/contrib/scripts/systemd/README
```diff
@@ -3,21 +3,21 @@
...
     # groupadd -r wildfly
-    # useradd -r -g wildfly -d /opt/wildfly -s /sbin/nologin wildfly
+    # useradd -r -g wildfly -d /usr/share/wildfly -s /sbin/nologin wildfly
 
 == Install WildFly
 
-    # tar xvzf wildfly-10.0.0.Final.tar.gz -C /opt
-    # ln -s /opt/wildfly-10.0.0.Final /opt/wildfly
-    # chown -R wildfly:wildfly /opt/wildfly
+    # tar xvzf wildfly-10.0.0.Final.tar.gz -C /var/tmp
```
rarguello: This is wrong, the softlink can't point to /var/tmp

core-feature-pack/src/main/resources/content/docs/contrib/scripts/systemd/wildfly.service
```diff
@@ -5,12 +5,15 @@ Before=httpd.service
 
 [Service]
 Environment=LAUNCH_JBOSS_IN_BACKGROUND=1
+EnvironmentFile=-/etc/wildfly/wildfly.d/*
```
rarguello: Why?

core-feature-pack/src/main/resources/content/docs/contrib/scripts/systemd/wildfly.service
```diff
@@ -5,12 +5,15 @@ Before=httpd.service
...
 EnvironmentFile=-/etc/wildfly/wildfly.conf
+Sockets=httpd.socket
```
rarguello: 8080 is not an httpd.socket

BenjaminLefoul: But 80 is (at least in Fedora: https://pkgs.fedoraproject.org/cgit/rpms/httpd.git/ )

The main reason why 8080 is used by default on wildfly and elsewhere is because of the kernel limitation on port numbers. Using systemd, this is no longer a problem.

core-feature-pack/src/main/resources/content/docs/contrib/scripts/systemd/wildfly.service
BenjaminLefoul: I have seen it used. Maybe not so necessary indeed.

core-feature-pack/src/main/resources/content/docs/contrib/scripts/init.d/wildfly-init-debian.sh
```diff
@@ -59,7 +59,7 @@ fi
 
 # Location of wildfly
 if [ -z "$JBOSS_HOME" ]; then
-	JBOSS_HOME="/opt/wildfly"
+	JBOSS_HOME="/usr/share/wildfly"
```
jorsol: This is not a native package, if /usr/share was approved in Fedora it should be changed in Fedora it self.
http://www.tldp.org/LDP/Linux-Filesystem-Hierarchy/html/opt.html

The scripts here are not just "examples", they are fully working startup scripts, I don't see the case to change the path just because was approved in Fedora, the scripts apply to both debian and red hat based distros.

core-feature-pack/src/main/resources/content/docs/contrib/scripts/systemd/wildfly.service
ctomc: so you can bind to port under 1024 with non root user now?

BenjaminLefoul: Well you can have systemd spawn them, which essentially solves this historical problem we are stuck with on non-custom kernels. I think this is what apache does as packaged in Fedora. See this discussion from last year in the kernel mailing list: https://lkml.org/lkml/2015/12/14/395

jmesnil:
@BenjaminLefoul  As this pull request has been held up for a long time, I am closing it to get the pull request queue under control again. Since this is very old now, it will be a very long time until this one will be addressed. Also, I don't have much time to spend on this to evaluate on a case by case basis.

If it is a new feature, please follow the new process for features. Ping @wildfly/mergers-core  if you are unsure what this means, and we can help. Please do that before opening the pull request though.

Sorry for any inconvenience!

PR #1594, opened 2016-06-01
WFCORE-1572 Fixed parsing of argument value with a whitespace

spyrkob:
JIRA: https://issues.jboss.org/browse/WFCORE-1572

### DIFF
--- a/cli/src/test/java/org/jboss/as/cli/parsing/test/CommandTestCase.java
+++ b/cli/src/test/java/org/jboss/as/cli/parsing/test/CommandTestCase.java
@@ -264,6 +264,36 @@ public class CommandTestCase {
         assertEquals("\"a \\\"b\\\"\"", cmd.getPropertyValue("--arg"));
     }
 
+    @Test
+    public void testCurlyBrackets() throws Exception {
+        DefaultCallbackHandler cmd = parse("cmd --arg={a ");
+        assertEquals("cmd", cmd.getOperationName());
+        assertEquals("--arg", cmd.getLastParsedPropertyName());
+        assertEquals("{a ", cmd.getPropertyValue("--arg"));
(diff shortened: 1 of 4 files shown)


kabir:
@aloubyansky please approve

cli/src/test/java/org/jboss/as/cli/parsing/test/CommandTestCase.java
```diff
@@ -264,6 +264,22 @@ public void testQuotesInQuotes() throws Exception {
         assertEquals("\"a \\\"b\\\"\"", cmd.getPropertyValue("--arg"));
     }
 
+    @Test
+    public void testCurlyBrackets() throws Exception {
+        DefaultCallbackHandler cmd = parse("cmd --arg={a ");
+        assertEquals("cmd", cmd.getOperationName());
+        assertEquals("--arg", cmd.getLastParsedPropertyName());
+        assertEquals("{a", cmd.getPropertyValue("--arg"));
```
aloubyansky: I'm wondering whether the expected value should include the space too, i.e '{a ', since the value was not complete.

spyrkob:
@aloubyansky yes it should - I just checked and it seems to be working. Wildfly needs to be started in domain mode for the rollout header to work, but that wouldn't suggest '--' either... maybe it was using older version of core?

cli/src/test/java/org/jboss/as/cli/parsing/test/CommandTestCase.java
spyrkob: Good point, thanks. I didn't think about it. I'll fix that asap

aloubyansky:
This is good to merge. Thanks, Bartosz!

PR #1596, opened 2016-06-07
WFCORE-1576 allow the server to auto generate self signed certificates

stuartwdouglas:


stuartwdouglas:
Retest this please

controller/src/main/java/org/jboss/as/controller/parsing/Namespace.java
```diff
@@ -42,53 +42,57 @@
     XML_SCHEMA_INSTANCE("http://www.w3.org/2001/XMLSchema-instance"),
 
     // domain versions, in numerical order
-    DOMAIN_1_0(1, "urn:jboss:domain:1.0"),
+    DOMAIN_1_0(1, 0, "urn:jboss:domain:1.0"),
```
darranl: Can we please avoid reformatting all of these - it takes a lot of effort porting schema changes between different branches - if we want to fix formatting I think it should be upstream.

core-feature-pack/src/main/resources/configuration/domain/template.xml
```diff
@@ -24,7 +24,7 @@
   ~ */
   -->
 
-<domain xmlns="urn:jboss:domain:4.1">
+<domain xmlns="urn:jboss:domain:4.2">
```
darranl: I think submitting upstream first could make more sense, a lot of these changes will not have equivalents upstream.

domain-management/src/main/java/org/jboss/as/domain/management/security/FileKeyManagerService.java
```diff
@@ -21,26 +21,49 @@
  */
 package org.jboss.as.domain.management.security;
 
-import java.security.KeyStore;
```
darranl: The import order looks back to front.

domain-management/src/main/java/org/jboss/as/domain/management/security/FileKeyManagerService.java
```diff
@@ -117,45 +141,117 @@ public void setAlias(String alias) {
...
     }
 
-    @Override
-    protected KeyStore loadKeyStore() {
-        return keyStore.getKeyStore();
+    X509Certificate generateCertificate(KeyPair pair) throws Exception {
+        PrivateKey privkey = pair.getPrivate();
+        X509CertInfo info = new X509CertInfo();
+        Date from = new Date();
+        Date to = new Date(from.getTime() + (1000L * 60L * 60L * 24L * 365L * 10L));
```
darranl: I am still tempted to think this should be shorter rather than longer, i.e. it is an out of the box solution not a long term solution.

domain-management/src/main/java/org/jboss/as/domain/management/security/FileKeyManagerService.java
```diff
@@ -117,45 +141,117 @@ public void setAlias(String alias) {
...
+        }
     }
 
-    @Override
-    public void stop(StopContext context) {
-        super.stop(context);
-        keyStore = null;
+    private void generateFileKeyStore(Path path) {
+        try {
+            DomainManagementLogger.SECURITY_LOGGER.keystoreWillBeCreated(path.toString());
```
darranl: I think in addition to logging that this is being created it would be beneficial to log information about the self signed certificate - at least this way the users have a chance to verify the certificate is the one generated by the server.

darranl:
Also as I said by e-mail I think the forking of the schema and the editing of the schema should be split into two commits - that way we can see the changes to the schema specific for this PR.

darranl:
Also please get rid of ManagementXML_4_2.java - we don't add a new fork for every minor, we just edit ManagementXML_4.

Also as this is only a new optional attribute we don't need any special version handling in there - if a user does add it to a version 4.0 or 4.1 configuration it is not going to break anything.

domain-management/src/main/java/org/jboss/as/domain/management/parsing/ManagementXml.java
```diff
@@ -47,7 +47,11 @@ public static ManagementXml newInstance(Namespace namespace, ManagementXmlDelega
             case 3:
                 return new ManagementXml_Legacy(namespace, delegate);
             default:
-                return new ManagementXml_4(namespace, delegate);
+                if(namespace.getMinorVersion() < 2) {
+                    return new ManagementXml_4(namespace, delegate);
+                } else {
+                    return new ManagementXml_4_2(namespace, delegate);
```
darranl: As I said in the main comments this should all be handled in ManagementXml_4 - a minor version doesn't need a complete clone.

domain-management/src/main/java/org/jboss/as/domain/management/security/FileKeyManagerService.java
```diff
@@ -21,26 +21,49 @@
...
 import org.jboss.as.controller.services.path.PathManager.Callback;
 import org.jboss.as.controller.services.path.PathManager.Event;
 import org.jboss.as.controller.services.path.PathManager.PathEventContext;
+import org.jboss.as.domain.management.logging.DomainManagementLogger;
 import org.jboss.msc.inject.Injector;
-import org.jboss.msc.service.StartContext;
 import org.jboss.msc.service.StartException;
 import org.jboss.msc.service.StopContext;
 import org.jboss.msc.value.InjectedValue;
+import sun.security.x509.AlgorithmId;
```
darranl: The problem we are going to have here is the IBM JDK, although to start with we could detect if running on the IBM JDK and make this feature unavailable.

domain-management/src/main/java/org/jboss/as/domain/management/security/FileKeyManagerService.java
```diff
@@ -117,45 +141,117 @@ public void setAlias(String alias) {
...
+        info.set(X509CertInfo.SERIAL_NUMBER, new CertificateSerialNumber(sn));
+        info.set(X509CertInfo.SUBJECT, owner);
+        info.set(X509CertInfo.ISSUER, owner);
+        info.set(X509CertInfo.KEY, new CertificateX509Key(pair.getPublic()));
+        info.set(X509CertInfo.VERSION, new CertificateVersion(CertificateVersion.V3));
+        AlgorithmId algo = new AlgorithmId(AlgorithmId.sha256WithRSAEncryption_oid);
+        info.set(X509CertInfo.ALGORITHM_ID, new CertificateAlgorithmId(algo));
+
+        // Sign the cert to identify the algorithm that's used.
+        X509CertImpl cert = new X509CertImpl(info);
```
dmlloyd: It'd be nice to use the Elytron DER-encoding API instead of (possibly JDK-specific) internal classes to build up the certificate data.  Maybe we should introduce a CertificateBuilder in fact...

darranl: Yes we are planning to do that - it is very tightly related to the CSR handling as well.

domain-management/src/main/java/org/jboss/as/domain/management/security/KeystoreAttributes.java
```diff
@@ -71,6 +71,10 @@
             .setFlags(AttributeAccess.Flag.RESTART_RESOURCE_SERVICES)
             .build();
 
+    public static final SimpleAttributeDefinition GENERATE_SELF_SIGNED_CERTIFICATE_HOST = new SimpleAttributeDefinitionBuilder(org.jboss.as.controller.descriptions.ModelDescriptionConstants.GENERATE_SELF_SIGNED_CERTIFICATE_HOST, ModelType.STRING, true)
+            .setAllowExpression(true)
+            .setAllowNull(true)
```
bstansberry: The others in here have .setFlags(AttributeAccess.Flag.RESTART_RESOURCE_SERVICES) and this one should too as it has the same handling.

Overall we're not good about adding this but since this particular class happens to be good about it.... :)

bstansberry:
retest this please

We'll see if my CI config changes make the StandardConfigsXMLValidationUnitTestCase failures go away.

+1 to just using ManagementXml_4 and thus avoiding the need to change Namespace.java.

stuartwdouglas:
Master branch version is at: https://github.com/wildfly/wildfly-core/pull/1606

core-feature-pack/src/main/resources/configuration/domain/template.xml
stuartwdouglas: I have created the upstream PR

domain-management/src/main/java/org/jboss/as/domain/management/security/FileKeyManagerService.java
stuartwdouglas: I disagree. IMHO this should only be used for development purposes, and I think it is highly unlikely that developers will get a real security certificate for local use.

Even if this is used outside of development having a longer expiration means that users will see less of the certificate error message. If the certificate needs to be constantly regenerated it would train users to just ignore the SSL error.

domain-management/src/main/java/org/jboss/as/domain/management/security/FileKeyManagerService.java
stuartwdouglas: Once Elytron is in I will update this.

domain-management/src/main/java/org/jboss/as/domain/management/security/KeystoreAttributes.java
stuartwdouglas: Done

stuartwdouglas:
Retest this please

domain-management/src/main/java/org/jboss/as/domain/management/security/FileKeyManagerService.java
darranl: Maybe not so short they need to constantly keep re-generating it but 10 years is such a long time to put this into production and forget about it.

One thing that is 100% guaranteed is any features in this server that are for development only will go into production.

domain-management/src/main/java/org/jboss/as/domain/management/security/FileKeyManagerService.java
darranl: To get the Elytron implementation in we will probably need to create a new branch for Elytron 1.1 that can have the changes from David and bump Elytron development to 1.2.

stuartwdouglas:
Retest this please

stuartwdouglas:
Retest this please

domain-management/src/main/java/org/jboss/as/domain/management/security/X509CertificateBuilder.java
```diff
@@ -0,0 +1,1194 @@
+/*
+ * JBoss, Home of Professional Open Source.
+ * Copyright 2015, Red Hat, Inc., and individual contributors
+ * as indicated by the @author tags. See the copyright.txt file in the
+ * distribution for a full listing of individual contributors.
+ *
+ * This is free software; you can redistribute it and/or modify it
+ * under the terms of the GNU Lesser General Public License as
```
darranl: Why has the licence been changed?  This is Apache licensed code.

bstansberry:
I removed mgmt-review label as this is good to go; fixme label is for the license header change in X509CertificateBuilder. Other than that this is good to go.

bstansberry:
Superceded by #1619 (which just adds a pom dep)

PR #1597, opened 2016-06-07
[WFCORE-1578] Better check of names of existing resources when adding…

soul2zimate:
… '{local|remote-destination-outbound-socket-binding'
https://issues.jboss.org/browse/WFCORE-1578

### DIFF
--- a/server/src/main/java/org/jboss/as/server/services/net/LocalDestinationOutboundSocketBindingAddHandler.java
+++ b/server/src/main/java/org/jboss/as/server/services/net/LocalDestinationOutboundSocketBindingAddHandler.java
@@ -28,11 +28,20 @@ import static org.jboss.as.server.services.net.LocalDestinationOutboundSocketBin
 import static org.jboss.as.server.services.net.OutboundSocketBindingResourceDefinition.OUTBOUND_SOCKET_BINDING_CAPABILITY;
 
 import java.net.UnknownHostException;
+import java.util.ArrayList;
+import java.util.List;
+import java.util.Set;
 
 import org.jboss.as.controller.AbstractAddStepHandler;
 import org.jboss.as.controller.OperationContext;
 import org.jboss.as.controller.OperationFailedException;
+import org.jboss.as.controller.OperationStepHandler;
 import org.jboss.as.controller.PathAddress;
+import org.jboss.as.controller.OperationContext.Stage;
+import org.jboss.as.controller.descriptions.ModelDescriptionConstants;
+import org.jboss.as.controller.logging.ControllerLogger;
+import org.jboss.as.controller.parsing.Element;
+import org.jboss.as.controller.registry.Resource;
 import org.jboss.as.network.NetworkInterfaceBinding;
 import org.jboss.as.network.OutboundSocketBinding;
 import org.jboss.as.network.SocketBinding;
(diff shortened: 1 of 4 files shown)


soul2zimate:
retest this please

server/src/main/java/org/jboss/as/server/services/net/LocalDestinationOutboundSocketBindingAddHandler.java
```diff
@@ -71,6 +79,30 @@ protected void performRuntime(final OperationContext context, final ModelNode op
...
+                                        Element.SOCKET_BINDING.getLocalName(), Element.OUTBOUND_SOCKET_BINDING.getLocalName(),
+                                        name, Element.SOCKET_BINDING_GROUP.getLocalName(), currentSocketBindingGroupName);
+
+                        }
+                        break;
+                    }
+                }
+            }
+
+        }, Stage.RUNTIME);
```
bstansberry: This is model validation and hence needs to be in Stage.MODEL. So it needs to be moved out of performRuntime().

server/src/main/java/org/jboss/as/server/services/net/LocalDestinationOutboundSocketBindingAddHandler.java
```diff
@@ -71,6 +79,30 @@ protected void performRuntime(final OperationContext context, final ModelNode op
...
         final String outboundSocketName = address.getLastElement().getValue();
+
+        context.addStep(new OperationStepHandler() {
+            @Override
+            public void execute(OperationContext context, ModelNode operation) throws OperationFailedException {
+                final String currentSocketBindingGroupName = address.getParent().getLastElement().getValue();
+                final Set<ResourceEntry> children = context.readResourceFromRoot(PathAddress.EMPTY_ADDRESS)
+                        .getChildren(ModelDescriptionConstants.SOCKET_BINDING_GROUP);
+                for (Resource.ResourceEntry re : children) {
+                    if (re.getName().equalsIgnoreCase(currentSocketBindingGroupName)) {
```
bstansberry: 1) of 2)

Do you know if this is an issue with a managed domain config? There may be some validation for this problem already. 

If not, then we need to deal with the fact that socket-binding-group A can include B and C, and B can include D. Names need to be unique across the entire included set. I believe we check for this, although that check may need to be enhanced to deal with the fact that the unique names need to across socket-binding AND {remote|local}-destination-outbound-socket-binding.

Ideally the existing managed domain checks will handle this in the domain case, and then all this logic can be guarded by a check of context.getProcessType(), confirming that 1) processType.isServer() and 2) processType != ProcessType.DOMAIN_SERVER.

bstansberry: 2) of 2)

Assuming the analysis of 1) shows this can be limited to non-domain servers, then L87-90 seems overly complex.

Resource re = context.readResourceFromRoot(PathAddress.pathAddress(SOCKET_BINDING_GROUP, currentSocketBindingGroupName)) should give you the only resource that's relevant.

server/src/main/java/org/jboss/as/server/services/net/LocalDestinationOutboundSocketBindingAddHandler.java
```diff
@@ -71,6 +79,30 @@ protected void performRuntime(final OperationContext context, final ModelNode op
...
+                        for (Resource.ResourceEntry entry : re.getChildren(ModelDescriptionConstants.SOCKET_BINDING)) {
```
bstansberry: You need to check REMOTE_DESTINATION_OUTBOUND_SOCKET_BINDING as well.

Instead of iterating over the children, all re.getChildrenNames(SOCKET_BINDING) and then do a simple contains check.

server/src/main/java/org/jboss/as/server/services/net/LocalDestinationOutboundSocketBindingAddHandler.java
bstansberry: Make that:
```
Resource re = context.readResourceFromRoot(PathAddress.pathAddress(SOCKET_BINDING_GROUP, currentSocketBindingGroupName), false);
```

The 'false' param reduces the amount of work the server does and still provides the necessary data.

bstansberry:
Does there need to be a fix for the handling of socket-binding itself as well? This fixes the outbound-s-b handlers.

soul2zimate:
Hi Brian, I have rebased this as per comment above. Now the validation is moved to Stage.MODEL. And it also checks socket-binding names in host.xml. SocketBindingAddHandler has the same check.

About the domain configuration, An error due to "Duplicate resource" can be seen if I use same socket-binding name or same {remote|local}-destination-outbound-socket-binding name across socket binding groups.

But, I don't see any operation failure across socket-binding-group when I use same names between socket-binding and {remote|local}-destination-outbound-socket-binding. For instance, following configuration is accepted without an error in server start.

```
        <socket-binding-group name="full-sockets" default-interface="public">
 		... ...
            <socket-binding name="txn-status-manager" port="4713"/>
            <socket-binding name="myBinding"/>
            <outbound-socket-binding name="mail-smtp">
                <remote-destination host="localhost" port="25"/>
            </outbound-socket-binding>
        </socket-binding-group>
        	...	...
        <socket-binding-group name="my-group" default-interface="public" includes="full-sockets">
            <socket-binding name="mail-smtp"/>
            <outbound-socket-binding name="myBinding">
                <remote-destination host="localhost" port="4714"/>
            </outbound-socket-binding>
        </socket-binding-group>
    </socket-binding-groups>
		... ...
    <server-groups>
        <server-group name="main-server-group" profile="full">
            <jvm name="default">
                <heap size="64m" max-size="512m"/>
            </jvm>
            <socket-binding-group ref="my-group"/>
        </server-group>
		... ...
```


This does not look right to me even the server did not complain anything as in standalone mode.

soul2zimate:
new related failure, will check it

soul2zimate:
retest this please

soul2zimate:
```
steps.add(Util.getWriteAttributeOperation(getSlaveServerConfigAddress(SERVER1), GROUP, GROUP2));
steps.add(Util.getWriteAttributeOperation(getSlaveServerConfigAddress(SERVER1), GROUP, "testgroup"));
```
AutoIgnoredResourcesDomainTestCase is passed after removing first line, it turns out it's not correct to add "GROUP2" as it should test "testgroup"

server/src/main/java/org/jboss/as/server/services/net/LocalDestinationOutboundSocketBindingAddHandler.java
```diff
@@ -67,6 +74,51 @@ protected boolean requiresRuntime(OperationContext context) {
...
+                }else{
+                    resource = context.readResourceFromRoot(PathAddress.pathAddress(ModelDescriptionConstants.SOCKET_BINDING_GROUP, socketBindingGroupName), false);
+                }
+
+                Set<String> socketBindingNames = resource.getChildrenNames(ModelDescriptionConstants.SOCKET_BINDING);
+                for (String name : socketBindingNames) {
+                    if (name.equalsIgnoreCase(outboundSocketBindingName)) {
+                        throwEx(name, socketBindingGroupName);
+                    }
+                }
```
bstansberry: I don't think there's any requirement to be unique ignoring case, and all of these can just be a test of "if (existingnames.contains(newname)". If somehow there's some code that fails in the case of equalsIgnoreCase that code is likely buggy. These strings are used in resource addresses, MSC service names, and capabilty names and none of those require uniqueness ignoring case.

testsuite/domain/src/test/java/org/jboss/as/test/integration/domain/autoignore/AutoIgnoredResourcesDomainTestCase.java
```diff
@@ -493,7 +493,6 @@ public void test51_testCompositeOperation() throws Exception {
 
         steps.add(addGroupOp);
         steps.add(Util.getWriteAttributeOperation(getSlaveServerConfigAddress(SERVER1), SOCKET_BINDING_GROUP, SOCKETS3));
-        steps.add(Util.getWriteAttributeOperation(getSlaveServerConfigAddress(SERVER1), GROUP, GROUP2));
```
bstansberry: I don't think we want to change this test, as it is testing some very hard use cases and the failure is likely telling us there is something wrong with this fix.

I suspect that what it's telling us is that the validation step you've added needs to handle the fact that the socket binding group may no longer exist by the time that step executes. It handles it by returning cleanly; if the group is no longer present, then the child cannot be invalid.

bstansberry:
Re this:

{code}

        <socket-binding-group name="full-sockets" default-interface="public">
        ... ...
            <socket-binding name="txn-status-manager" port="4713"/>
            <socket-binding name="myBinding"/>
            <outbound-socket-binding name="mail-smtp">
                <remote-destination host="localhost" port="25"/>
            </outbound-socket-binding>
        </socket-binding-group>
            ... ...
        <socket-binding-group name="my-group" default-interface="public" includes="full-sockets">
            <socket-binding name="mail-smtp"/>
            <outbound-socket-binding name="myBinding">
                <remote-destination host="localhost" port="4714"/>
            </outbound-socket-binding>
        </socket-binding-group>
    </socket-binding-groups>
{code}

If the process type isn't a server (i.e. it's an HC processing the domain.xml change) then your validation step needs to start with the socket-binding group the s-b references and check it, and then look for any defined 'includes' attribute on that group and check those, including checking any groups included by those and so on. A given group should only be checked once though.

soul2zimate:
This has been rebased based on review comments. thanks.

soul2zimate:
retest this please

bstansberry:
Thanks for this; it sure wasn't simple!

PR #1599, opened 2016-06-07
[WFCORE-1581] Convert LIST of PROPERTY to OBJECT for attributes of type OBJECT

bstansberry:
It wouldn't surprise me if this requires changes in full, as I had to fix some stuff in core to avoid failures due to incorrectly formatted params being compared to the resulting model. That's what the last 3 commits are.

Update: I was right. PR to fix test failures is https://github.com/wildfly/wildfly/pull/8957 . CI job to test this in combo with that PR is https://ci.wildfly.org/viewLog.html?buildId=18356&tab=buildResultsDiv&buildTypeId=WF_WildFlyCoreIntegrationExperiments


kabir:
Retest this please

PR #1608, opened 2016-06-13
WFCORE-1374 - make code work with jdk9b112 and newer

ctomc:
- upgrade jboss-modules
- WFCORE-1290 upgrade jboss-marshalling

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -105,10 +105,10 @@
         <version.org.jboss.logging.jboss-logging-tools>2.0.1.Final</version.org.jboss.logging.jboss-logging-tools>
         <version.org.jboss.logmanager.jboss-logmanager>2.0.4.Final</version.org.jboss.logmanager.jboss-logmanager>
         <version.org.jboss.logmanager.log4j-jboss-logmanager>1.1.2.Final</version.org.jboss.logmanager.log4j-jboss-logmanager>
-        <version.org.jboss.marshalling.jboss-marshalling>1.4.10.Final</version.org.jboss.marshalling.jboss-marshalling>
+        <version.org.jboss.marshalling.jboss-marshalling>1.4.11.Final</version.org.jboss.marshalling.jboss-marshalling>
         <!-- this is test only dependancy -->
         <version.org.jboss.metadata.jboss-metadata-common>10.0.0.Final</version.org.jboss.metadata.jboss-metadata-common>
-        <version.org.jboss.modules.jboss-modules>1.5.1.Final</version.org.jboss.modules.jboss-modules>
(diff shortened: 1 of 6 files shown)


pom.xml
```diff
@@ -105,10 +105,10 @@
         <version.org.jboss.logging.jboss-logging-tools>2.0.1.Final</version.org.jboss.logging.jboss-logging-tools>
         <version.org.jboss.logmanager.jboss-logmanager>2.0.3.Final</version.org.jboss.logmanager.jboss-logmanager>
         <version.org.jboss.logmanager.log4j-jboss-logmanager>1.1.2.Final</version.org.jboss.logmanager.log4j-jboss-logmanager>
-        <version.org.jboss.marshalling.jboss-marshalling>1.4.10.Final</version.org.jboss.marshalling.jboss-marshalling>
+        <version.org.jboss.marshalling.jboss-marshalling>1.4.11.Final</version.org.jboss.marshalling.jboss-marshalling>
```
rhusar: Component upgrade Jira https://issues.jboss.org/browse/WFCORE-1290

testsuite/domain/pom.xml
```diff
@@ -177,7 +177,6 @@
                     <argLine>${jvm.args.ip.client} ${jvm.args.timeouts} ${surefire.system.args}</argLine>
 
                     <systemPropertyVariables>
-                        <jboss.options>-ea ${surefire.system.args}</jboss.options>
```
bstansberry: Why?

How is -ea getting set on the PC and HC process without this?

What other values were there in ${surefire.system.args} and how are they getting set on the PC and HC process without this?

testsuite/domain/pom.xml
```diff
@@ -240,7 +239,6 @@
                             <argLine>${jvm.args.ip.client} ${jvm.args.timeouts} ${surefire.system.args}</argLine>
 
                             <systemPropertyVariables>
-                                <jboss.options>${surefire.system.args}</jboss.options>
```
bstansberry: Same question as above.

testsuite/domain/pom.xml
ctomc: This is somewhat an oversight, only -ea should have been removed as it is by default now present in surefire.system.args already. parameter itself should stay.

ctomc: fixed

testsuite/domain/pom.xml
ctomc: fixed

testsuite/domain/pom.xml
ctomc: Also fixed on #1609

bstansberry: testsuite/pom.xml overrides the setting from pom.xml and does not include -ea.

PR #1609, opened 2016-06-13
[2.x] WFCORE-1374 - backport jdk9 related changes from master

ctomc:
- Update compiler config for JDK9 b118+
- WFCORE-1374 - make code work with jdk9b115 and newer
- upgrade jboss-modules to 1.5.2.Final
- upgrade jboss-marshalling to 1.4.11.Final
- [WFCORE-1490] JDK9 doesn't provide sun.boot.class.path attribute
- [WFCORE-1491] Don't rely on default (JDK) JAXP providers - specify explicit ones
- [WFCORE-1493] Fix proper -Djdk.launcher.addexports.%d JVM property values propagation
- [WFCORE-1431] Define and propagate ${modular.jdk.args} property where necessary
- Properly create writers with charset and cleanup resources

### DIFF
--- a/testsuite/domain/pom.xml
+++ b/testsuite/domain/pom.xml
@@ -51,7 +51,6 @@
         <xslt.scripts.dir>${basedir}/../src/test/xslt</xslt.scripts.dir>
         <enforcer.skip>true</enforcer.skip>
         <jbossas.ts.dir>${basedir}/..</jbossas.ts.dir>
-
     </properties>
 
     <!--
@@ -178,7 +177,7 @@
                     <argLine>${jvm.args.ip.client} ${jvm.args.timeouts}</argLine>
 
                     <systemPropertyVariables>
-                        <jboss.options>-ea ${surefire.system.args}</jboss.options>
+                        <jboss.options>${surefire.system.args}</jboss.options>
                         <java.util.logging.manager>org.jboss.logmanager.LogManager</java.util.logging.manager>
                         <jboss.home>${project.basedir}/target/wildfly-core</jboss.home>
                         <module.path>${jboss.dist}/modules</module.path>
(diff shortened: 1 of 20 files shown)


testsuite/shared/src/main/java/org/jboss/as/test/integration/management/util/CustomCLIExecutor.java
```diff
@@ -195,37 +203,6 @@ public static String execute(File cliConfigFile, String operation, String contro
...
-            try {
-                Thread.sleep(100);
-            } catch (InterruptedException e) {
-            }
-            now = System.currentTimeMillis();
-        } while (now - start < timeout);
-
-        fail("Server did not reload in the imparted time.");
-    }
-
```
bstansberry: Is this related to this work?

ctomc: This is unused method, it was in original commit on master. Just took whole commit...
I can remove it if you want.

bstansberry: Please restore it. On master it was removed as part of the "Improve test runner to allow few more things when running in manual server mode" commit.

testsuite/domain/pom.xml
```diff
@@ -178,7 +177,7 @@
                     <argLine>${jvm.args.ip.client} ${jvm.args.timeouts}</argLine>
 
                     <systemPropertyVariables>
-                        <jboss.options>-ea ${surefire.system.args}</jboss.options>
+                        <jboss.options>${surefire.system.args}</jboss.options>
```
bstansberry: testsuite/pom.xml overrides the surefire.system.args setting from pom.xml and drops the -ea, so without this the PC and HC processes will not have it.

bstansberry:
Superceded by #1618

PR #1613, opened 2016-06-15
[WFCORE-1481] list-add operation doesn't work on login-modules attribute

TomasHofman:
https://issues.jboss.org/browse/WFCORE-1481

### DIFF
--- a/controller/src/main/java/org/jboss/as/controller/AbstractWriteAttributeHandler.java
+++ b/controller/src/main/java/org/jboss/as/controller/AbstractWriteAttributeHandler.java
@@ -96,7 +96,7 @@ public abstract class AbstractWriteAttributeHandler<T> implements OperationStepH
 
         if (requiresRuntime(context)) {
             final ModelNode updatedValue = newValue;
-            context.addStep(new OperationStepHandler() {
+            context.addStep(operation, new OperationStepHandler() {
                 @Override
                 public void execute(OperationContext context, ModelNode operation) throws OperationFailedException {
                     final ModelNode resolvedValue = attributeDefinition != null ? attributeDefinition.resolveModelAttribute(context, submodel) : updatedValue.resolve();
--- a/controller/src/main/java/org/jboss/as/controller/operations/global/AbstractCollectionHandler.java
+++ b/controller/src/main/java/org/jboss/as/controller/operations/global/AbstractCollectionHandler.java
@@ -19,6 +19,7 @@
 package org.jboss.as.controller.operations.global;
 
 import static org.jboss.as.controller.descriptions.ModelDescriptionConstants.OP_ADDR;
+import static org.jboss.as.controller.descriptions.ModelDescriptionConstants.RESULT;
 import static org.jboss.as.controller.operations.global.EnhancedSyntaxSupport.containsEnhancedSyntax;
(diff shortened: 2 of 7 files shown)


controller/src/main/java/org/jboss/as/controller/operations/global/ReadAttributeHandler.java
```diff
@@ -183,8 +183,26 @@ private void doExecuteInternal(OperationContext context, ModelNode operation) th
                 }
             }
             if (useEnhancedSyntax) {
-                ModelNode resolved = EnhancedSyntaxSupport.resolveEnhancedSyntax(attributeExpression.substring(attributeName.length()), context.getResult());
-                context.getResult().set(resolved);
+                // remove attribute name from expression string ("attribute-name.rest" => "rest")
+                int prefixLength = attributeName.length();
+                if (attributeExpression.length() > prefixLength && attributeExpression.charAt(prefixLength) == '.') {
```
ctomc: this looks bit redundant, as attributeExpression will always be longer as name attribute itself.
also if is of same size which would mean it is not expression at all, it will not come to this code block at all.

TomasHofman: Fixed, thanks!

controller/src/main/java/org/jboss/as/controller/operations/global/AbstractCollectionHandler.java
```diff
@@ -81,28 +80,60 @@ public void execute(OperationContext context, ModelNode operation) throws Operat
             attributeName = extractAttributeName(attributeName);
         }
         final AttributeAccess attributeAccess = context.getResourceRegistration().getAttributeAccess(PathAddress.EMPTY_ADDRESS, attributeName);
-        if (attributeAccess == null) {
-            throw new OperationFailedException(ControllerLogger.ROOT_LOGGER.unknownAttribute(attributeName));
-        } else if (requiredReadWriteAccess && attributeAccess.getAccessType() != AttributeAccess.AccessType.READ_WRITE) {
-            throw new OperationFailedException(ControllerLogger.ROOT_LOGGER.attributeNotWritable(attributeName));
-        }
```
bstansberry: Note to self: I believe this is just pruning because ReadAttributeHandler or WriteAttributeHandler will have already done these checks before this handler ever gets called.

controller/src/main/java/org/jboss/as/controller/AbstractWriteAttributeHandler.java
```diff
@@ -96,7 +96,7 @@ public void execute(OperationContext context, ModelNode operation) throws Operat
 
         if (requiresRuntime(context)) {
             final ModelNode updatedValue = newValue;
-            context.addStep(new OperationStepHandler() {
+            context.addStep(operation, new OperationStepHandler() {
```
bstansberry: Why this? I don't believe there is an change in behavior from this.

bstansberry: Apparently there is. Which is odd.

bstansberry:
Superceded by #1694

PR #1615, opened 2016-06-16
[master] Enhancement WFCORE-1597

jaikiran:
The commit here implements the enhancement noted in https://issues.jboss.org/browse/WFCORE-1597 to be a bit lenient while checking if the configured truststore's keystore provider type is JKS, to avoid running into hard to debug SSL issues as noted in the forum thread here https://developer.jboss.org/message/958142#958142

### DIFF
--- a/domain-management/src/main/java/org/jboss/as/domain/management/security/SecurityRealmAddHandler.java
+++ b/domain-management/src/main/java/org/jboss/as/domain/management/security/SecurityRealmAddHandler.java
@@ -557,8 +557,8 @@ public class SecurityRealmAddHandler implements OperationStepHandler {
 
         ServiceName keyManagerServiceName = null;
 
-        String provider = KeystoreAttributes.KEYSTORE_PROVIDER.resolveModelAttribute(context, ssl).asString();
-        if (ssl.hasDefined(KEYSTORE_PATH) || (JKS.equals(provider) == false)) {
+        final String provider = KeystoreAttributes.KEYSTORE_PROVIDER.resolveModelAttribute(context, ssl).asString();
+        if (ssl.hasDefined(KEYSTORE_PATH) || !JKS.equalsIgnoreCase(provider)) {
             keyManagerServiceName = AbstractKeyManagerService.ServiceUtil.createServiceName(SecurityRealm.ServiceUtil.createServiceName(realmName));
             addKeyManagerService(context, ssl, keyManagerServiceName, serviceTarget);
         }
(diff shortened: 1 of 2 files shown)


domain-management/src/main/java/org/jboss/as/domain/management/security/SecurityRealmAddHandler.java
```diff
@@ -674,18 +675,17 @@ private void addTrustManagerService(OperationContext context, ModelNode ssl, Ser
...
+            final ProviderTrustManagerService trustManagerService = new ProviderTrustManagerService(provider, keystorePassword);
             serviceBuilder = serviceTarget.addService(serviceName, trustManagerService);
         } else {
-            String path = KeystoreAttributes.KEYSTORE_PATH.resolveModelAttribute(context, ssl).asString();
+            final String path = KeystoreAttributes.KEYSTORE_PATH.resolveModelAttribute(context, ssl).asString();
             ModelNode relativeToNode = KeystoreAttributes.KEYSTORE_RELATIVE_TO.resolveModelAttribute(context, ssl);
             String relativeTo = relativeToNode.isDefined() ? relativeToNode.asString() : null;
 
-            FileTrustManagerService trustManagerService = new FileTrustManagerService(provider, path, relativeTo, keystorePassword);
+            final FileTrustManagerService trustManagerService = new FileTrustManagerService(JKS.toLowerCase(Locale.ENGLISH), path, relativeTo, keystorePassword);
```
mchoma: Isnt with this change other then "JKS" truststore forbidden? Wouldn't be provider.toLowerCase better?

mchoma: Are you sure with IBM java? I found https://www.ibm.com/support/knowledgecenter/SSYKE2_8.0.0/com.ibm.java.security.component.80.doc/security-component/keytoolDocs/supportedkeystoretypes.html , where uppercase is documented

ctomc:
@mchoma we can run PR also with IBM jdk if you think we should

jaikiran:
Hi @ctomc, I think it would be good to run it against IBM JDK if possible. That will help in verifying if this is going to introduce any kind of regression.

ctomc:
here you go 
https://ci.wildfly.org/viewLog.html?buildId=18955&tab=buildResultsDiv&buildTypeId=WildFlyCore_PullRequest_WildFlyCoreFullIntegration
and 
https://ci.wildfly.org/viewLog.html?buildId=18955&tab=buildResultsDiv&buildTypeId=WildFlyCore_PullRequest_WildFlyCoreFullIntegration
both running with IBM JDK 8. You can see status of this job also on the PR statuses.

jaikiran:
Thanks Tomaz.

I thought a bit more about this last night and decided to do a minor change to the commit in this and the other PR so that I no longer force the lower case conversion. This way the user can specify the right (case) value depending on the JDK and we just pass it along literally and yet at the same time while checking if it's JKS, we do a case insensitive check. 

This PR and the other one https://github.com/wildfly/wildfly-core/pull/1616 are now ready for review.

PR #1620, opened 2016-06-18
WFCORE-1604 Fail-fast start up init.d scripts on error / WFCORE-1615 Make wildfly-init-redhat.sh consistent with wildfly-init-debian.sh

jorsol:
Now that WildFly creates an startup-marker we can detect an error before timeout of the init.d startup scripts.

I also change tabs for spaces, but the real change is in the start() section on:
grep -i 'error:' "$JBOSS_MARKERFILE"

### DIFF
--- a/core-feature-pack/src/main/resources/content/docs/contrib/scripts/init.d/wildfly-init-debian.sh
+++ b/core-feature-pack/src/main/resources/content/docs/contrib/scripts/init.d/wildfly-init-debian.sh
@@ -13,10 +13,10 @@
 # Default-Start:        2 3 4 5
 # Default-Stop:         0 1 6
 # Short-Description:    WildFly Application Server
-# Description:          Provide WildFly startup/shutdown script
+# Description:          WildFly startup/shutdown script
 ### END INIT INFO
 
-NAME=wildfly
+NAME=$(readlink -f ${0} | xargs basename)
 DESC="WildFly Application Server"
 DEFAULT="/etc/default/$NAME"
 
@@ -32,6 +32,7 @@ if [ -r /etc/default/locale ]; then
 	export LANG
 fi
 
+# Source function library.
 . /lib/lsb/init-functions
 
 if [ -r /etc/default/rcS ]; then
@@ -59,7 +60,7 @@ fi
 
 # Location of wildfly
 if [ -z "$JBOSS_HOME" ]; then
-	JBOSS_HOME="/opt/wildfly"
+	JBOSS_HOME="/opt/$NAME"
 fi
 export JBOSS_HOME
 
@@ -142,7 +143,7 @@ fi
 export JBOSS_CONSOLE_LOG
 
 # Location to set the pid file
-JBOSS_PIDFILE="/var/run/$NAME/$NAME.pid"
+JBOSS_PIDFILE="/var/run/wildfly/$NAME.pid"
 export JBOSS_PIDFILE
 
 # Launch wildfly in background
@@ -178,7 +179,6 @@ case "$1" in
 		fi
 
 		count=0
-		launched=0
 		until [ $count -gt $STARTUP_WAIT ]
 		do
 			sleep 1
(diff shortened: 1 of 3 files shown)


bstansberry:
Thank you, but please remove the whitespace changes so we can readily evaluate the functional change.

jorsol:
Done Brian, removed the withespace change, the failed test are not related to this commit.

Retest this please.

core-feature-pack/src/main/resources/content/docs/contrib/scripts/init.d/wildfly-init-debian.sh
```diff
@@ -16,7 +16,7 @@
 # Description:          Provide WildFly startup/shutdown script
 ### END INIT INFO
 
-NAME=wildfly
+NAME=$(basename ${0})
```
bstansberry: Note to self: I don't think this is directly part of WFCORE-1604 but it makes this script consistent with wildfly-init-redhat.sh which seems like a good thing.

core-feature-pack/src/main/resources/content/docs/contrib/scripts/init.d/wildfly-init-debian.sh
```diff
@@ -59,7 +59,7 @@ fi
 
 # Location of wildfly
 if [ -z "$JBOSS_HOME" ]; then
-	JBOSS_HOME="/opt/wildfly"
+	JBOSS_HOME="/opt/$NAME"
```
bstansberry: Note to self: I don't think this is directly part of WFCORE-1604 but it makes this script consistent with wildfly-init-redhat.sh which seems like a good thing.

core-feature-pack/src/main/resources/content/docs/contrib/scripts/init.d/wildfly-init-debian.sh
```diff
@@ -142,7 +142,7 @@ fi
 export JBOSS_CONSOLE_LOG
 
 # Location to set the pid file
-JBOSS_PIDFILE="/var/run/$NAME/$NAME.pid"
+JBOSS_PIDFILE="/var/run/$NAME/wildfly.pid"
```
bstansberry: Note to self: I don't think this is directly part of WFCORE-1604 but it makes this script consistent with wildfly-init-redhat.sh which seems like a good thing.

core-feature-pack/src/main/resources/content/docs/contrib/scripts/init.d/wildfly-init-debian.sh
```diff
@@ -187,22 +187,23 @@ case "$1" in
...
-		if check_status; then
-			log_end_msg 0
-		else
-			log_end_msg 1
-		fi
 
 		if [ $launched -eq 0 ]; then
-			log_warning_msg "$DESC hasn't started within the timeout allowed"
-			log_warning_msg "please review file \"$JBOSS_CONSOLE_LOG\" to see the status of the service"
+			log_warning_msg "$DESC started with errors, please see server log for details."
```
bstansberry: Here there is a change in behavior, in that if it doesn't start in time, but there is no error, we log this "started with errors" message instead of a message discussing the timeout. Which is confusing if it really was just a timeout.

OTOH this is what wildfly-init-redhat.sh is doing.

Nicer though would be to discriminate between the two cases in both scripts.

@ctomc WDYT?

jorsol: That's a great idea @bstansberry, I can make the change to discriminate between the two cases, hold the merge.

BTW, I could make 2 jira issues and pull request if you like, one to make wildfly-init-redhat.sh consistent with wildfly-init-debian.sh, and this one.

bstansberry: If you want to create a separate JIRA, that would be nice. I don't see a need for a separate PR, and even separate commits for the two aspects isn't critical if it's a hassle. A single commit referencing two JIRAs in its message works fine too.

jorsol:
Note: The wildfly-init-debian.sh was tested on Ubuntu 14.04 (pre-systemd) and 16.04 (systemd). The wildfly-init-redhat.sh was superficially tested on Fedora 24 (systemd), it works, but the correct tests needs to be made in pre-systemd versions.

If someone can fully test on RHEL or any other pre-systemd would appreciate.

bstansberry:
@jorsol Thanks for this! I need to do a bit of housekeeping before merging:

jorsol:
@bstansberry, hold this merge, I'm making improvements to the redhat script.

I'm testing it in CentOS (RHEL) 6.8, CentOS (RHEL) 7.2 and Fedora 24 Server. The redhat script will have some features found in debian script, and a little housekeeping too.

I'm also making them more compatible with systemd.

bstansberry:
Ok, Jorge; sounds good.

jorsol:
Done Brian, it can be reviewed now and merged if approved.
I agree with the MIT License.

PR #1626, opened 2016-06-21
[WFCORE-1609] Allow static definition of the product slot. Have SelfContainedContai…

bstansberry:
…ner use "main"

### DIFF
--- a/core-model-test/framework/src/main/java/org/jboss/as/core/model/test/TestModelControllerService.java
+++ b/core-model-test/framework/src/main/java/org/jboss/as/core/model/test/TestModelControllerService.java
@@ -263,7 +263,7 @@ class TestModelControllerService extends ModelTestModelControllerService {
             RunningMode initialRunningMode = runningModeControl.getRunningMode();
             boolean backupDomainFiles = false;
             boolean useCachedDc = false;
-            ProductConfig productConfig = new ProductConfig(null, "",  props);
+            ProductConfig productConfig = ProductConfig.fromFilesystemSlot(null, "",  props);
             return new HostControllerEnvironment(props, isRestart, modulePath, processControllerAddress, processControllerPort,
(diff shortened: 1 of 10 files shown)


gastaldi:
LGTM :+1:

bstansberry:
@gastaldi if you can confirm this actually works right for Swarm, I can backport it to 2.2.0 for use in WildFly 10.1.

gastaldi:
I think it's ok, wdyt @kenfinnigan @bobmcwhirter @heiko-braun ? This is needed for https://issues.jboss.org/browse/SWARM-526 as ProductConfig relies on having a product.conf file in the jboss.home dir.

heiko-braun:
What goes into that module? @gastaldi  Can you briefly outline how it's going to work?

gastaldi:
@heiko-braun: Basically a META-INF/MANIFEST.MF file containing the product name and version entries. If you have a wildfly installation you'll find this module under the org.jboss.as.product:wildfly-full module path

gastaldi:
What is the best way to test it with Swarm? Do I need to build WildFly entirely and change the swarm dependency to effectively test that?

heiko-braun:
I think it's more then that. Since this is change to WF core, you need to build that, then build WF with the updated core version and finally build Swarm with the updated WF version.

gastaldi:
Yes, that's what I meant. Downloading now the internet for that test, thanks! :)

gastaldi:
I can confirm that this works as expected. Here is the output: 

    [org.jboss.as] (Controller Boot Thread) WFLYSRV0025: WildFly Swarm 1.0.0.CR2 (WildFly Core 2.0.10.Final) started in 1089ms - Started 63 of 75 services (14 services are lazy, passive or on-demand)

gastaldi:
I cherry-picked your commit and added to wildfly-core 2.0.10.Final to test

heiko-braun:
pretty cool. regrading @bstansberry 's question of backporting: if you we could get it into 10.1 that would be awesome.

PR #1628, opened 2016-06-22
Fix for WFCORE-1599

jfdenise:
This is a large rewrite of the ValueTypeCompleter class in order to replace the usage of stack and list with an object model of the parsed instances relationships.

The ValueCompleter used to support a syntax that looked invalid. Inside a List, one could type the properties of a complex object without the '{' and '}' characters. I really didn't see the point and Alexey confirmed that it was not a supported syntax. As soon as we have more than an item in the list, this is broken. I have removed this completion behavior.

In addition, support for List of complex types has been added.

### DIFF
--- a/cli/src/main/java/org/jboss/as/cli/impl/ValueTypeCompleter.java
+++ b/cli/src/main/java/org/jboss/as/cli/impl/ValueTypeCompleter.java
@@ -19,19 +19,20 @@
  * Software Foundation, Inc., 51 Franklin St, Fifth Floor, Boston, MA
  * 02110-1301 USA, or see the FSF site: http://www.fsf.org.
  */
-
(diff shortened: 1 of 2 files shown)


aloubyansky:
Put it on hold, please. We are investigating one issue.

jfdenise:
The completion for type annotated as File System path has been added with the associated unit test.

aloubyansky:
This is good to merge. Thanks, Jeff!

jfdenise:
retest this please!

bstansberry:
@ehsavoie Any idea about this DeploymentScannerUnitTestCase failure? I'm seeing it pop up from time to time.

aloubyansky:
From the CLI perspective, this is good to merge. Thanks, Jeff!

PR #1630, opened 2016-06-22
WFCORE-1616 ServerSuspendHandler tries to do step completion and roll…

stuartwdouglas:
…back handling in a different thread

### DIFF
--- a/server/src/main/java/org/jboss/as/server/operations/ServerSuspendHandler.java
+++ b/server/src/main/java/org/jboss/as/server/operations/ServerSuspendHandler.java
@@ -23,6 +23,9 @@
 package org.jboss.as.server.operations;
 
 
+import java.util.concurrent.CountDownLatch;
+import java.util.concurrent.atomic.AtomicBoolean;
+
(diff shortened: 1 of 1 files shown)


server/src/main/java/org/jboss/as/server/operations/ServerSuspendHandler.java
```diff
@@ -99,24 +106,36 @@ public void suspendStarted() {
...
                     }
                 };
                 suspendController.addListener(operationListener);
                 suspendController.suspend(timeout > 0 ?  timeout * 1000 : timeout);
+                if(timeout != 0) {
+                    try {
+                        latch.await();
+                    } catch (InterruptedException e) {
+                        Thread.currentThread().interrupt();
+                        throw new OperationFailedException(e);
```
bstansberry: You don't need/want to throw an OFE. When this method returns, the OperationContext will detect the interrupt status on the thread and fall into its cancellation handling. The OFE will just introduce noise.

PR #1632, opened 2016-06-23
[WFCORE-1617] fix tab completion for cli variables and properties.

soul2zimate:
https://issues.jboss.org/browse/WFCORE-1617

DefaultCallbackHandler sets "lastChunkIndex" during every necessary validation (validatedNodeType, validatedNodeName, validatedOperationName, etc) which is based on resolved value. This gives an incorrect offset value in case of variable or property substitution is applied.

### DIFF
--- a/cli/src/main/java/org/jboss/as/cli/operation/impl/DefaultCallbackHandler.java
+++ b/cli/src/main/java/org/jboss/as/cli/operation/impl/DefaultCallbackHandler.java
@@ -625,7 +625,7 @@ public class DefaultCallbackHandler extends ValidatingCallbackHandler implements
 
     @Override
     public int getLastChunkIndex() {
-        return lastChunkIndex;
+        return lastChunkIndex - substitutedLine.length() + originalLine.length();
     }
 
     @Override
--- a/cli/src/test/java/org/jboss/as/cli/parsing/test/VariablesTestCase.java
+++ b/cli/src/test/java/org/jboss/as/cli/parsing/test/VariablesTestCase.java
@@ -67,6 +67,7 @@ public class VariablesTestCase {
         ctx.setVariable(NODE_NAME_VAR_NAME, NODE_NAME_VAR_VALUE);
         ctx.setVariable(OP_VAR_NAME, OP_VAR_VALUE);
         ctx.setVariable(OP_PROP_VAR_NAME, OP_PROP_VAR_VALUE);
+        ctx.setVariable("myvar", "/subsystem=logging");
     }
 
     @After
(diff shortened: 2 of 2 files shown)


kabir:
@aloubyansky @jfdenise Please approve

cli/src/main/java/org/jboss/as/cli/operation/impl/DefaultCallbackHandler.java
```diff
@@ -625,7 +625,13 @@ public int getLastSeparatorIndex() {
...
     @Override
     public int getLastChunkIndex() {
-        return lastChunkIndex;
+        int originalLineLength = originalLine.length();
+        int substitutedLineLength = substitutedLine.length();
+        if (originalLineLength == substitutedLineLength)
+            return lastChunkIndex;
+        else {
+            return lastChunkIndex - (substitutedLineLength - originalLineLength);
+        }
```
aloubyansky: Would you mind changing it to a single line? I.e.
return lastChunkIndex - substitutedLine.length + originalLine.length();

aloubyansky:
Thanks for the contribution, @soul2zimate! I added a comment to the change. And please add a testcase to org.jboss.as.cli.parsing.test.VariablesTestCase. Thanks!

soul2zimate:
thanks,  I have rebased to add a test for system property use in tab completion

aloubyansky:
Have you tried to run the test w/o the fix? It will still pass. Can you confirm that?
I wanted to see something like this

@@ -67,6 +67,7 @@ public class VariablesTestCase {
         ctx.setVariable(NODE_NAME_VAR_NAME, NODE_NAME_VAR_VALUE);
         ctx.setVariable(OP_VAR_NAME, OP_VAR_VALUE);
         ctx.setVariable(OP_PROP_VAR_NAME, OP_PROP_VAR_VALUE);
+        ctx.setVariable("myvar", "/subsystem=logging");
     }
 
     @After
@@ -75,6 +76,7 @@ public class VariablesTestCase {
         ctx.setVariable(NODE_NAME_VAR_NAME, null);
         ctx.setVariable(OP_VAR_NAME, null);
         ctx.setVariable(OP_PROP_VAR_NAME, null);
+        ctx.setVariable("myvar", null);
     }
 
     @Test
@@ -220,6 +222,13 @@ public class VariablesTestCase {
         assertEquals("v=\\$" + OP_VAR_NAME, props.get(0));
     }
 
+    @Test
+    public void testOperationCompletion() throws Exception {
+        final ParsedCommandLine parsed = parse("$myvar:re");
+        assertEquals("re", parsed.getOperationName());
+        assertEquals(7, parsed.getLastChunkIndex());
+    }

Thanks!

soul2zimate:
Hmm, you're right. it still passed...
I have removed the invalid test and add the one you proposed. thanks for the help.

aloubyansky:
This is good to merge. Thanks!

PR #1636, opened 2016-06-27
Fix for WFCORE-1614

jfdenise:
In addition to allow for echoing the command in non interactive mode, the CLI XSD config file has been upgraded to release 3.1. Although the latest XSD was 3.0, the shipped xml config file was still 2.0. I have upgraded the shipped XML file to release 3.1.
The test that was conflicting with WFCORE-1590 has been fixed.

### DIFF
--- a/cli/src/main/java/org/jboss/as/cli/CliConfig.java
+++ b/cli/src/main/java/org/jboss/as/cli/CliConfig.java
@@ -177,4 +177,12 @@ public interface CliConfig {
      *          otherwise - false
      */
     boolean isAccessControl();
+
+    /**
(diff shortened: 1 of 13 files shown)


aloubyansky:
Looks OK to me. Thanks!

jfdenise:
Hold on, one test is now failing locally, investigating.

jfdenise:
The added test was changing the state of the server and was not properly removing some properties, making other tests to fail latter in the run. I am surprised to not have seen it previously. I have fixed this problem.

jfdenise:
retest this please!

PR #1637, opened 2016-06-27
[WFCORE-1579] Perform check when changing 'path' attribute of deploym…

TomasHofman:
…ent-scanner instance

https://issues.jboss.org/browse/WFCORE-1579
https://issues.jboss.org/browse/JBEAP-4874

Changes to the deployment scanner so that if deployment dir is not accessible:
* bootTimeScan() is skipped and error is logged (before the server failed to start)
* establishDeployedContentList() is not performed during startScanner(), instead it will be performed during first scan when deployment dir is accessible

### DIFF
--- a/deployment-scanner/src/main/java/org/jboss/as/server/deployment/scanner/FileSystemDeploymentService.java
+++ b/deployment-scanner/src/main/java/org/jboss/as/server/deployment/scanner/FileSystemDeploymentService.java
@@ -127,6 +127,7 @@ class FileSystemDeploymentService implements DeploymentScanner, NotificationHand
     private long scanInterval = 0;
     private volatile boolean scanEnabled = false;
     private volatile boolean firstScan = true;
+    private volatile boolean deployedContentEstablished = false;
     private ScheduledFuture<?> scanTask;
     private ScheduledFuture<?> rescanIncompleteTask;
     private ScheduledFuture<?> rescanUndeployTask;
@@ -158,6 +159,9 @@ class FileSystemDeploymentService implements DeploymentScanner, NotificationHand
     private final String relativePath;
     private final PropertyChangeListener propertyChangeListener;
     private Future<?> undeployScanTask;
+
+    private volatile boolean deploymentDirAccessible = true;
+    private volatile boolean lastScanSuccessful = true;
 
     @Override
     public void handleNotification(Notification notification) {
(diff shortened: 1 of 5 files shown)


TomasHofman:
Failing tests in the Wildfly testsuite need to be modified to reflect changes in this PR.

deployment-scanner/src/main/java/org/jboss/as/server/deployment/scanner/FileSystemDeploymentService.java
```diff
@@ -780,6 +804,44 @@ private boolean handleCompositeResult(ModelNode resultNode, StringBuilder failur
...
+        else if (!directory.canRead()) {
+            if (deploymentDirAccessible) {
+                deploymentDirAccessible = false;
+                ROOT_LOGGER.directoryIsNotReadable(deploymentDir.getAbsolutePath());
+            }
+        }
+        else if (!directory.canWrite()) {
+            if (deploymentDirAccessible) {
+                deploymentDirAccessible = false;
+                ROOT_LOGGER.directoryIsNotWritable(deploymentDir.getAbsolutePath());
```
bstansberry: There needs to be a parameter or something to disable the logging from this method if the information has already been logged. Otherwise it looks like the same logging will happen 3 times at boot (bootTimeScan, startScanner, scan) and then periodically thereafter every time the scanner runs.

bstansberry: Never mind; I'm not paying attention; that's what deploymentDirAccessible does!

deployment-scanner/src/main/java/org/jboss/as/server/deployment/scanner/FileSystemDeploymentService.java
```diff
@@ -567,6 +576,21 @@ private boolean scan(boolean oneOffScan, final DeploymentOperations deploymentOp
...
+                if (lastScanSuccessful) {
+                    lastScanSuccessful = false;
+                    ROOT_LOGGER.scanFailed(deploymentDir.getAbsolutePath());
+                }
+                return scheduleRescan;
+            }
+            // if deployed content list was not established during scanner start (due to inaccessible deployment dir),
+            // do it now
+            if (!deployedContentEstablished) {
+                establishDeployedContentList(deploymentDir, deploymentOperations);
```
bstansberry: @ehsavoie Can you review this PR please?

This bit here is the only concern I have. Not that it's necessarily wrong. Just that this is the one change where I think careful review is needed as it could introduce subtle bugs. And, if it's ok to do this here, then it seems unnecessary to do an establishDeployedContentList in startScanner() and L399-406 can just be dropped.

bstansberry:
Thanks for this, Tomas!

bstansberry:
retest this please

bstansberry:
https://github.com/wildfly/wildfly/pull/9071 is a PR to remove that test in full; #1701 is to add it to core. So this PR should be updated to update the test that will be in core.

TomasHofman:
I updated FileSystemDeploymentService according to Brian's suggestion (left out establishDeployedContentList() call from FileSystemDeploymentService#startScanner()) - it required few more changes in FileSystemDeploymentServiceUnitTestCase - and fixed the test case moved from full.

TomasHofman:
retest this please

TomasHofman:
retest this please

PR #1640, opened 2016-06-28
[WFCORE-1618] Refactoring management

bstansberry:
These are a bunch of refactoring commits that are intended to make the logic behind the management handling classes easier to understand. They are not meant to produce any change in behavior.

These do produce incompatible API changes in the 'protocol' module. But, 1) that is an internal module and 2) core 3.0 is a major release. Still, if anyone can think of any negative impacts from such a change let me know.

I created these as part of doing analysis to produce https://developer.jboss.org/docs/DOC-55624. Doing that I wanted to remove some of the conceptual complexity of these classes.


bstansberry:
retest this please

PR #1641, opened 2016-06-28
[WFCORE-1623] BoundedQueueThreadPoolResourceDefinition does not able …

gaol:
…to add additional check on the threads executor

Jira: https://issues.jboss.org/browse/WFCORE-1623

### DIFF
--- a/threads/src/main/java/org/jboss/as/threads/BoundedQueueThreadPoolResourceDefinition.java
+++ b/threads/src/main/java/org/jboss/as/threads/BoundedQueueThreadPoolResourceDefinition.java
@@ -75,7 +75,11 @@ public class BoundedQueueThreadPoolResourceDefinition extends PersistentResource
         return new BoundedQueueThreadPoolResourceDefinition(blocking, registerRuntimeOnly, type, poolNameBase, resolverPrefix, addHandler, removeHandler);
     }
 
-    private BoundedQueueThreadPoolResourceDefinition(boolean blocking, boolean registerRuntimeOnly,
+    /**
+     * @deprecated This class is not designed for subclassing and having this constructor be accessible is a specific workaround for WFCORE-1623 that may be reverted at any time
+     */
+    @Deprecated
+    protected BoundedQueueThreadPoolResourceDefinition(boolean blocking, boolean registerRuntimeOnly,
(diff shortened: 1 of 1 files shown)


kabir:
What is the check mentioned in the Jira that the jca subsystem needs to do?

gaol:
So that [JcaWorkManagerDefinition](https://github.com/wildfly/wildfly/blob/master/connector/src/main/java/org/jboss/as/connector/subsystems/jca/JcaWorkManagerDefinition.java#L74) in JCA subsystem can instance the `BoundedQueueThreadPoolResourceDefinition` by specifying a customized `BoundedQueueThreadPoolAdd`  to check whether there are any existing `long-running-threads` or `short-running-threads` defined by the WorkManager. Only one `long-running-threads` or `short-running-threads` sub-resource can be added.

threads/src/main/java/org/jboss/as/threads/BoundedQueueThreadPoolResourceDefinition.java
```diff
@@ -75,7 +75,7 @@ public static BoundedQueueThreadPoolResourceDefinition create(String type, Threa
         return new BoundedQueueThreadPoolResourceDefinition(blocking, registerRuntimeOnly, type, poolNameBase, resolverPrefix, addHandler, removeHandler);
     }
 
-    private BoundedQueueThreadPoolResourceDefinition(boolean blocking, boolean registerRuntimeOnly,
+    protected BoundedQueueThreadPoolResourceDefinition(boolean blocking, boolean registerRuntimeOnly,
```
bstansberry: Please add

```
/**
 * @deprecated This class is not designed for subclassing and having this constructor be accessible is a specific workaround for WFCORE-1623 that may be reverted at any time
 */
@Deprecated
```

This whole module needs a bit of a rethink, but there are too many higher priorities right now. Until that happens I think a workaround like this is ok for closely related code like the JCA subsystem, but I don't want other people using it.

A longer term possibility is to have the constructor (private again) take a Parameters object, and then have public static create methods to generate that Parameters object. Plus a create method that takes the Parameters object and calls the constructor.

But I don't know if that really gains much; a review of the whole library is needed. So no point pursuing that right now.

gaol:
Added `@Deprecated` to the constructor.

PR #1642, opened 2016-06-29
Fix for WFCORE-1626

jfdenise:
Introduce --module-paths and associated doc and testing.

### DIFF
--- a/cli/src/main/resources/help/module.txt
+++ b/cli/src/main/resources/help/module.txt
@@ -2,8 +2,8 @@ SYNOPSIS
 
   To add a module locally
 
-    module add --name=module_name [--slot=slot_name]
-               --resources=list_of_jars [--resource-delimiter=delimiter]
+    module add [--module-root-dir=modules_root_directory_path] --name=module_name 
+               [--slot=slot_name] --resources=list_of_jars [--resource-delimiter=delimiter]
                (--module-xml=file_path |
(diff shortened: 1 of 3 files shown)


cli/src/main/resources/help/module.txt
```diff
@@ -39,7 +39,11 @@ DESCRIPTION
 
 ARGUMENTS
 
- --name          - (required) the name of the module to be added or removed.
+ --modules-path - (optional) The absolute file path to the server's modules
```
bstansberry: @dmlloyd I thought about recommending "--module-path" (no 's') for this, to be consistent with JBoss Modules' term, but I now think it's better that it is different. This isn't a $PATH-like path which is what the JBoss Modules concept is. This is a single filesystem path where the newly added module should be placed. WDYT?

Or, perhaps --modules-dir to be even further from the JBoss Modules term.

dmlloyd: Maybe --module-root-dir as the plural is also a bit confusing in this context?

jfdenise: I saw the recommandation but thought that, being the root dir of the modules of a server, modules-path would be simpler for users to get what this option is. JBoss default location is "modules", and this option points to such "root location".
As David is suggesting, --module-root-dir is even more obvious.

PR #1643, opened 2016-06-29
[WFCORE-1627] CLI tab comple should reflect current position of prompt. Enable DeployCompletionTestCase.

soul2zimate:
…rompt. Enable DeployCompletionTestCase.

https://issues.jboss.org/browse/WFCORE-1627 
1. fix a tab completion issue when cursor moves.
2. DeployCompletionTest is never running, rename it to match *TestCase pattern. Also enable an ignored test.

### DIFF
--- a/cli/src/main/java/org/jboss/as/cli/CommandCompleter.java
+++ b/cli/src/main/java/org/jboss/as/cli/CommandCompleter.java
@@ -83,7 +83,15 @@ public class CommandCompleter implements CommandLineCompleter {
 
         final DefaultCallbackHandler parsedCmd = (DefaultCallbackHandler) ctx.getParsedCommandLine();
         try {
-            parsedCmd.parse(ctx.getCurrentNodePath(), buffer, false, ctx);
+            // WFCORE-1627 parse current position if cursor moves
+            if (ctx.getArgumentsString() == null && buffer.length() > cursor) {
+                parsedCmd.parse(ctx.getCurrentNodePath(), buffer.substring(0, cursor), false, ctx);
+            } else if (ctx.getArgumentsString() != null && buffer.length() - ctx.getArgumentsString().length() > cursor) {
+                // multiple lines
+                parsedCmd.parse(ctx.getCurrentNodePath(), buffer.substring(0, ctx.getArgumentsString().length() + cursor), false, ctx);
+            } else {
+                parsedCmd.parse(ctx.getCurrentNodePath(), buffer, false, ctx);
+            }
         } catch(UnresolvedVariableException e) {
             final String variable = e.getExpression();
             if(buffer.endsWith(variable)) {
--- a/cli/src/main/java/org/jboss/as/cli/impl/Console.java
+++ b/cli/src/main/java/org/jboss/as/cli/impl/Console.java
@@ -144,8 +144,9 @@ public interface Console {
                                     co.getBuffer(), co.getCursor(), candidates);
                             co.setOffset(offset);
                             co.setCompletionCandidates(candidates);
-                            if (co.getCompletionCandidates().size() == 1 &&
-                                    co.getCompletionCandidates().get(0).getCharacters().startsWith(co.getBuffer()))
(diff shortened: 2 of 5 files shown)


soul2zimate:
retest this please

bstansberry:
@aloubyansky or @jfdenise Can one of you approve this please?

cli/src/main/java/org/jboss/as/cli/CommandCompleter.java
```diff
@@ -83,7 +83,15 @@ protected int doComplete(CommandContext ctx, String buffer, int cursor, List<Str
 
         final DefaultCallbackHandler parsedCmd = (DefaultCallbackHandler) ctx.getParsedCommandLine();
         try {
-            parsedCmd.parse(ctx.getCurrentNodePath(), buffer, false, ctx);
+            // WFCORE-1627 parse current position if cursor moves
+            if (ctx.getArgumentsString() == null && buffer.length() > cursor) {
+                parsedCmd.parse(ctx.getCurrentNodePath(), buffer.substring(0, cursor), false, ctx);
+            } else if (ctx.getArgumentsString() != null && buffer.length() - ctx.getArgumentsString().length() > cursor) {
```
jfdenise: Could you add a test case for multiline completion?

cli/src/main/java/org/jboss/as/cli/impl/Console.java
```diff
@@ -144,8 +144,9 @@ public void complete(CompleteOperation co) {
                                     co.getBuffer(), co.getCursor(), candidates);
                             co.setOffset(offset);
                             co.setCompletionCandidates(candidates);
-                            if (co.getCompletionCandidates().size() == 1 &&
-                                    co.getCompletionCandidates().get(0).getCharacters().startsWith(co.getBuffer()))
+                            String buffer = cmdCtx.getArgumentsString() == null ? co.getBuffer() : ctx.getArgumentsString() + co.getBuffer();
```
jfdenise: Could you add a test case to cover this case, it doesn't seem an obvious one.

cli/src/test/java/org/jboss/as/cli/completion/operation/test/OperationNameCompletionTestCase.java
```diff
@@ -44,6 +46,8 @@
 
     private MockCommandContext ctx;
     private OperationRequestCompleter completer;
+    private CommandRegistry cmdRegistry;
+    private CommandCompleter cmdCompleter;
```
jfdenise: I am not sure that this test should be extended with a test for CommandCompleter. You should possibly create a new CommandCompleterTestCase with some tests to cover specifically the CommandCompleter for:
- operations
-commands
and multilines.
Thank-you.

soul2zimate:
Thanks, I have added more tests to cover multilines.
For the second commit on Console.java. I have tested inside my terminal for scenario

> [standalone@embedded /] /subsystem=logging/logger=\
> > comTAB_HERE

It's a bit tricky that these codes are adding an anonymous Completion object to a private aesh.console.Console instance which does not and should not expose itself.

jfdenise:
Thank-you for adding the tests, this will help maintain the CLI code. 
I don't really follow you for the second commit. I have tested 
[standalone@embedded /] /subsystem=logging/logger=\
comTAB_HERE

with wildly-core master branch and completion already works properly. It would be nice if you could develop a bit on this corner case. 

Thanks again for your contribution!
JF

soul2zimate:
For the second commit, the difference is about not appending separator after completion. 

> [standalone@embedded /] /subsystem=logging/logger=\
> comTAB_HERE

It previously added an extra white space after completion as:

> [standalone@embedded /] /subsystem=logging/logger=\    
> > com.arjuna :read-resource()

But there is no real harm happened as operation ignores white space in between. I can remove this change if this is not considered as an issue. Thanks.

jfdenise:
I think that you can keep the fix, just add a simple comment. 
The fix is now good for me. Thanks!
JF

PR #1644, opened 2016-06-29
[2.x]Properly handle :undefine-attribute in domain sync-model

ctomc:
for master it is part of #1608


bstansberry:
@ctomc can you file a JIRA for this and resolve it against 2.2.0.CRwhatever and 3.0.0.Alphawhatever please? And link the PRs. This seems like a bug that people are going to hit in the real world so it should be findable in JIRA.

ctomc:
@bstansberry done https://issues.jboss.org/browse/WFCORE-1630

PR #1645, opened 2016-06-29
WFCORE-1629: Fixed potential NPE

gastaldi:
modulePath is null when executed inside WildFly Swarm, therefore causing a NPE in this method


gastaldi:
This patch just performs a null check in modulePath before attempting to discover the console versions

bstansberry:
this is ok to test

gastaldi:
Can we make this fix available in the next WildFly 10 release?

bstansberry:
@gastaldi see #1646

gastaldi:
You are awesome, thank you!

PR #1651, opened 2016-06-30
WFCORE-1561 restore truncated result address

luck3y:
This restores the ProxyAddressTranslated address for a result by providing a method in ProxyAddressTranslater to merge the original op address, and the result. Thoughts / comments appreciated, as always.

### DIFF
--- a/controller/src/main/java/org/jboss/as/controller/ProxyOperationAddressTranslator.java
+++ b/controller/src/main/java/org/jboss/as/controller/ProxyOperationAddressTranslator.java
@@ -22,6 +22,7 @@
 package org.jboss.as.controller;
 
 import org.jboss.as.controller.descriptions.ModelDescriptionConstants;
+import org.jboss.dmr.ModelNode;
 
 /**
  * Used by {@link ProxyController} implementations to translate addresses to the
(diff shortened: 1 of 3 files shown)


controller/src/main/java/org/jboss/as/controller/ProxyOperationAddressTranslator.java
```diff
@@ -48,6 +55,31 @@ public PathAddress translateAddress(PathAddress address) {
                 return translated;
             }
 
+            public PathAddress restoreAddress(PathAddress opAddress, PathAddress translatedAddress) {
+                // original is /host=foo/server=bar/subsystem=*
+                // translated is /subsystem=*
+                // we want the result to be translated back to /host=foo/server=bar/subsystem=quux
```
kabir: This comment stumped me a bit. In my head an address with wildcards should expand into a list. After reading the Jira it makes sense though.

luck3y: I agree this isn't explicit enough, I'll expand on the comment to make it clearer.

luck3y:
Yeah, actually that's a case I was a bit concerned about, where the same path key could occur > 1 times,    I'll add another test and fix the code.

luck3y:
retest this please

luck3y:
@bstansberry I pushed, and the fixme didn't drop. Anyway, how does this look? Test added and I retested against full and I get the expected full addresses for everything I can think of to test.

bstansberry:
@luck3y Looks good. Our github integration seems to be broken; besides the fixme it doesn't appear to have rerun CI either. I'll do that manually as part of merge processing.

PR #1654, opened 2016-07-04
[WFCORE-1632] Server processing request isn't stopped immediately but…

gaol:
… waits for request processing to finish

Jira: https://issues.jboss.org/browse/WFCORE-1632

If we should not be waiting for threads to complete before shutting down, I think we should call the `XnioWorker.shutdownNow()` instead of `XnioWorker.shutdown()`, the `XnioWorker.shutdownNow()` will interrupt the execution of the running threads.

### DIFF
--- a/io/subsystem/src/main/java/org/wildfly/extension/io/WorkerService.java
+++ b/io/subsystem/src/main/java/org/wildfly/extension/io/WorkerService.java
@@ -61,7 +61,7 @@ public class WorkerService implements Service<XnioWorker> {
     public void stop(StopContext context) {
         this.stopContext = context;
         context.asynchronous();
-        worker.shutdown();
+        worker.shutdownNow();
         worker = null;
     }
 


kabir:
@stuartwdouglas wdyt?

bstansberry:
We need to be really careful about this, as this effects more than just Undertow. It would be nice if it's good though. :)

@gaol This halts execution of a request that's actively being processed? For example a test servlet that does a long Thread.sleep() in doGet()? I'm asking because shutdownNow is documented as:

```
     * Immediately terminate the worker.  Any outstanding tasks are collected and returned in a list.  Upon return
     * worker shutdown will have commenced but not necessarily completed; however the worker will only complete its
     * current tasks instead of completing all tasks.
```

That last sentence implies tasks being executed will not be interrupted.

dmlloyd:
By the time we have shut down the worker, everything else should be completely done with it.  If they aren't, then we need to fix our dependency management.

Also I am generally opposed to a context-free shutdownNow() like this one as it breaks the otherwise intact contract that every task that is accepted by an executor is also run by it.

Overall I think this commit is probably an XY problem.  We should look at the original problem in more detail; I don't think we should accept this commit.

bstansberry:
@dmlloyd The dependency management can be fine, in the sense that the services that use the worker all properly depend on the worker service and have been stopped. But unless those services somehow track the threads that are handling requests and interrupt them in stop(), the worker can still have threads executing work. That work will continue until done or it hits some logic that needs some no-longer-available resource and fails.

(I'm not arguing for shutdownNow() here.)

Reverting WFCORE-1299 is an option; that will allow a process to exit since our shutdown hook that awaits MSC termination will no longer be held up by the WorkerService not stopping. But in a reload situation there would still be a zombie worker executing a task while the WorkerService happily goes DOWN and then eventually a new set of services get started.

BTW, the XnioWorker shutdownNow javadoc looks wrong as it implies currently executing tasks will complete, but the TaskPool threads are actually interrupted. @gaol never mind my question above to you; I see in the XnioWorker code that tasks are interrupted.

stuartwdouglas:
I don't think this approach is great, unless the threads are also marked as daemon threads the server will still take time to shut down. 
If the threads are being interupped an alternate approach would be to call shutdownNow() and then wait of the list of futures that is returned. This will provide a faster shutdown (assuming the threads respond to the interupt) while also making sure all threads have finished.

gaol:
The original problem is that user wants to shutdown the server quickly while the user has a time-consuming servlet processing the request(`Thread.sleep(25s)`).

IMO, the thread must be interrupted to get the problem fixed.

I also feel a little violent to use `shutdownNow()` in WorkerService for all. :-)

Maybe one more parameter can be added to the shutdown|reload operation like: `interrupt=true|false` (default to false to keep the same as before), so that end user decides whether to shut down the server by interrupting all executing tasks? If it is the case, we only need to care about how the `WorkerService.stop()` knows about the parameter.

Another possible approach might be a configuration in Undertow subsystem to decide whether the request processing should be interrupted on `ListenerService.stop()` in Undertow subsystem. (I am totally guessing here, just in case it might be an option. :-))

dmlloyd:
Yeah I think interrupting running tasks is not a bad idea, but I also think that it might be expensive to do it, depending on how Undertow is or is not tracking them.  In any event I don't think any variant using shutdownNow() is workable because you'd still have to run the resultant tasks somehow; it becomes harder to do so when you have no threads left.

bstansberry:
@ctomc What was the driver for WFCORE-1299? Was there reported incorrect behavior or was it more that the WorkerService was stopping without truly being stopped?

bstansberry:
If we go for an approach where dependent services track and interrupt request threads then we'll need to apply that to other places than Undertow.

gaol:
I don't know what to do next, :-) Should we start a thread in wildfly-dev for this?

bstansberry:
Last commit on https://github.com/wildfly/wildfly-core/compare/master...bstansberry:worker-svc-shutdown is an experiment with using shutdownNow and then offloading the unprocessed tasks to the general management executor.

gaol:
@bstansberry Thank you. I tried your branch in my local environment by re-building latest WildFly, the long-running-servlet can be shutdown immediately.

The log with DEBUG on after `:shutdown(timeout=1)` is called can be found here: https://gist.github.com/gaol/feebbd48d37f347060c32a31c4523519

It seems there are no tasks left when the `List<Runnable> tasks = worker.shutdownNow()` is invoked in [WorkerService.java#90](https://github.com/wildfly/wildfly-core/compare/master...bstansberry:worker-svc-shutdown#diff-61bbb9c9485d5c844d1daf62207dd7e9R90)

bstansberry:
@gaol Thanks! How does that compare to the logging from the current code in master? I'm curious whether those DEBUG logs are related to introducing shutdownNow() or whether they happen already, just from the existing shutdown() call.

FWIW I did a couple CI runs against that branch:

https://ci.wildfly.org/viewLog.html?buildId=49100&tab=buildResultsDiv&buildTypeId=WF_WildFlyCoreIntegrationExperiments

https://ci.wildfly.org/viewLog.html?buildId=49102&tab=buildResultsDiv&buildTypeId=WF_WildFlyCoreIntegrationExperimentsWindows

The latter timed out. I haven't investigated so I don't know if that's meaningful, beyond that it wasn't due to a hang in the final test. We get build timeouts intermittently. But, this change affects shutdown and reload so if it is meaningfully increasing time for those, it can result in an overall increase in CI time, which would be a sign of a problem.

gaol:
@bstansberry The DEBUG log in current code in master can be found here: https://gist.github.com/gaol/b38189b8493df5b5c6b779c510ba64f7

From that, we can see the same `DEBUG` message from `io.undertow` with `XNIO007007` and from `io.undertow.request.io` with `UT005013`. So I think the `DEBUG` messages are **not** related to `shutdownNow()` call. :)

bstansberry:
@gaol @dmlloyd @stuartwdouglas I've rebased https://github.com/wildfly/wildfly-core/compare/master...bstansberry:worker-svc-shutdown and here are two new experimental runs with it:

https://ci.wildfly.org/viewLog.html?buildId=56664&tab=buildResultsDiv&buildTypeId=WF_WildFlyCoreIntegrationExperiments

https://ci.wildfly.org/viewLog.html?buildId=56666&tab=buildResultsDiv&buildTypeId=WF_WildFlyCoreIntegrationExperimentsWindows

bstansberry:
Another set of experiments runs:

https://ci.wildfly.org/viewLog.html?buildId=58512&tab=buildResultsDiv&buildTypeId=WF_WildFlyCoreIntegrationExperiments

https://ci.wildfly.org/viewLog.html?buildId=58514&tab=buildResultsDiv&buildTypeId=WF_WildFlyCoreIntegrationExperimentsWindows

bstansberry:
See #2557

bstansberry:
Closing in favor of #2557.

PR #1669, opened 2016-07-11
[WFCORE-307]: When Wildfly fails with JBAS015810: failed to resolve interface management, it should terminate.

ehsavoie:
Checking that one of the services for Management Interface is available at the end of the operation on boot.
Rollbacking if this is not the case.

Jira: https://issues.jboss.org/browse/WFCORE-307

### DIFF
--- a/controller/src/main/java/org/jboss/as/controller/management/ManagementInterfaceAddStepHandler.java
+++ b/controller/src/main/java/org/jboss/as/controller/management/ManagementInterfaceAddStepHandler.java
@@ -1,0 +1,115 @@
+/*
+ * Copyright 2016 JBoss by Red Hat.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.jboss.as.controller.management;
+
+
+import java.util.ArrayList;
+import java.util.List;
+import org.jboss.as.controller.AbstractAddStepHandler;
+import org.jboss.as.controller.AttributeDefinition;
+import org.jboss.as.controller.OperationContext;
+import org.jboss.as.controller.OperationFailedException;
+import org.jboss.as.controller.OperationStepHandler;
+import org.jboss.as.controller.capability.RuntimeCapability;
+import org.jboss.dmr.ModelNode;
+import org.jboss.msc.service.ServiceController;
+import org.jboss.msc.service.ServiceController.State;
+import org.jboss.msc.service.ServiceName;
+import org.jboss.msc.service.ServiceNotFoundException;
+import org.jboss.msc.service.ServiceRegistry;
+import org.wildfly.security.manager.WildFlySecurityManager;
+
+/**
+ * The base add handler for management interfaces.
+ * @author Emmanuel Hugonnet (c) 2016 Red Hat, inc.
+ */
+public abstract class ManagementInterfaceAddStepHandler extends AbstractAddStepHandler {
+
+    public static final OperationContext.AttachmentKey<Boolean> MANAGEMENT_INTERFACE_KEY = OperationContext.AttachmentKey.create(Boolean.class);
+
+    protected ManagementInterfaceAddStepHandler(final RuntimeCapability capability, final AttributeDefinition[] attributeDefinitions) {
+        super(capability, attributeDefinitions);
+    }
+
+    protected ManagementInterfaceAddStepHandler() {
+        super();
+    }
+
+   @Override
+    protected boolean requiresRuntime(OperationContext context) {
+        //TODO Gigantic HACK to disable the runtime part of this for the core model testing.
+        //The core model testing currently uses RunningMode.ADMIN_ONLY, but in the real world
+        //the http interface needs to be enabled even when that happens.
+        //I don't want to wire up all the services unless I can avoid it, so for now the tests set this system property
+        return WildFlySecurityManager.getPropertyPrivileged("jboss.as.test.disable.runtime", null) == null;
+    }
+
+    protected void addVerifyInstallationStep(OperationContext context, List<ServiceName> requiredServices) {
+        if(context.isBooting()) {
+            context.addStep(new LenientVerifyInstallationStep(requiredServices), OperationContext.Stage.VERIFY);
+        }
+    }
+
+    protected String asStringIfDefined(OperationContext context, AttributeDefinition attribute, ModelNode model) throws OperationFailedException {
+        ModelNode attributeValue = attribute.resolveModelAttribute(context, model);
+        return attributeValue.isDefined() ? attributeValue.asString() : null;
+    }
+
+    private static class LenientVerifyInstallationStep implements OperationStepHandler {
+        private final List<ServiceName> requiredServices;
+
+        private LenientVerifyInstallationStep(List<ServiceName> requiredServices) {
+            assert requiredServices != null;
+            this.requiredServices = requiredServices;
+        }
+
+        @Override
+        public void execute(OperationContext context, ModelNode operation) throws OperationFailedException {
+            List<ServiceName> failures = new ArrayList<>();
+            ServiceRegistry registry = context.getServiceRegistry(false);
+            for (ServiceName serviceName : requiredServices) {
+                try {
+                    ServiceController<?> controller = registry.getService(serviceName);
+                    if (controller == null || State.UP != controller.getState()) {
+                        failures.add(serviceName);
+                    }
+                } catch (ServiceNotFoundException ex) {
+                    failures.add(serviceName);
+                }
+            }
+            if (!failures.isEmpty()) {
+                Boolean attachment = context.getAttachment(MANAGEMENT_INTERFACE_KEY);
+                if (attachment == null || !context.getAttachment(MANAGEMENT_INTERFACE_KEY)) {
+                    context.attach(MANAGEMENT_INTERFACE_KEY, false);
+                    context.addStep(new VerifyInstallationStep(), OperationContext.Stage.VERIFY);
+                }
+            } else {
+                context.attach(MANAGEMENT_INTERFACE_KEY, true);
+            }
+        }
+    }
+
+    private static class VerifyInstallationStep implements OperationStepHandler {
+
+        @Override
+        public void execute(OperationContext context, ModelNode operation) throws OperationFailedException {
+            Boolean attachment = context.getAttachment(MANAGEMENT_INTERFACE_KEY);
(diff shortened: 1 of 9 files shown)


host-controller/src/main/java/org/jboss/as/host/controller/operations/NativeManagementAddHandler.java
```diff
@@ -81,6 +85,7 @@ protected void installServices(OperationContext context, NativeInterfaceCommonPo
         NativeManagementServices.installManagementWorkerService(serviceTarget, context.getServiceRegistry(false));
         ManagementRemotingServices.installDomainConnectorServices(serviceTarget, ManagementRemotingServices.MANAGEMENT_ENDPOINT,
                 nativeManagementInterfaceBinding, hostControllerInfo.getNativeManagementPort(), securityRealm, options);
+        return Arrays.asList(ManagementRemotingServices.MANAGEMENT_ENDPOINT, UndertowHttpManagementService.SERVICE_NAME);
```
bstansberry: I don't believe anything related to ManagementRemotingServices is required unless http-upgrade is turned on.

server/src/main/java/org/jboss/as/server/operations/HttpManagementAddHandler.java
```diff
@@ -185,6 +183,10 @@ protected void installServices(OperationContext context, HttpInterfaceCommonPoli
             RemotingHttpUpgradeService.installServices(serviceTarget, ManagementRemotingServices.HTTP_CONNECTOR, httpConnectorName,
                     ManagementRemotingServices.MANAGEMENT_ENDPOINT, commonPolicy.getConnectorOptions());
         }
+        if (socketBindingServiceName != null) {
+            return Arrays.asList(socketBindingServiceName, UndertowHttpManagementService.SERVICE_NAME);
```
bstansberry: UndertowHttpManagementService depends on socketBindingServiceName so only returning UndertowHttpManagementService is sufficient.

server/src/main/java/org/jboss/as/server/operations/HttpManagementAddHandler.java
```diff
@@ -185,6 +183,10 @@ protected void installServices(OperationContext context, HttpInterfaceCommonPoli
...
+        }
+        return Arrays.asList(ManagementRemotingServices.MANAGEMENT_ENDPOINT, UndertowHttpManagementService.SERVICE_NAME);
```
bstansberry: I don't believe anything related to ManagementRemotingServices is required unless http-upgrade is turned on.

On both this and the host controller variant I believe the service that best reflects the requirement for HTTP Upgrade isn't the endpoint service. Rather it's the one installed by RemotingHttpUpgradeService.installServices, named "UPGRADE_SERVICE_NAME.append(remotingConnectorName)". That service depends on the endpoint.

host-controller/src/main/java/org/jboss/as/host/controller/operations/NativeManagementAddHandler.java
bstansberry: Sorry; I put the comment ^^^ in the wrong place; should be no the http add handler.

host-controller/src/main/java/org/jboss/as/host/controller/operations/HttpManagementAddHandler.java
```diff
@@ -173,10 +176,7 @@ protected void installServices(OperationContext context, HttpInterfaceCommonPoli
             RemotingHttpUpgradeService.installServices(serviceTarget, ManagementRemotingServices.HTTP_CONNECTOR, httpConnectorName,
                     ManagementRemotingServices.MANAGEMENT_ENDPOINT, commonPolicy.getConnectorOptions());
         }
-    }
-
-    private ConsoleMode consoleMode(boolean consoleEnabled, boolean adminOnly) {
-        return consoleEnabled ? adminOnly ?  ConsoleMode.ADMIN_ONLY : ConsoleMode.CONSOLE : ConsoleMode.NO_CONSOLE;
+        return Arrays.asList(ManagementRemotingServices.MANAGEMENT_ENDPOINT, UndertowHttpManagementService.SERVICE_NAME);
```
bstansberry: I don't believe anything related to ManagementRemotingServices is required unless http-upgrade is turned on.

host-controller/src/main/java/org/jboss/as/host/controller/operations/NativeManagementAddHandler.java
bstansberry: The question here is I don't think UndertowHttpManagementService.SERVICE_NAME is related to native management.

For native I believe you want the service added by ManagementRemotingServices.installDomainConnectorServices, which is RemotingServices.serverServiceName(connectorName). That's the service that opens a socket for native requests. It depends on the various others like the endpoint so it won't be UP unless the others are.

server/src/main/java/org/jboss/as/server/operations/NativeManagementAddHandler.java
```diff
@@ -88,6 +92,7 @@ protected void installServices(OperationContext context, NativeInterfaceCommonPo
         ManagementRemotingServices.installConnectorServicesForSocketBinding(serviceTarget, endpointName,
                     ManagementRemotingServices.MANAGEMENT_CONNECTOR,
                     socketBindingServiceName, commonPolicy.getConnectorOptions());
+        return Arrays.asList(ManagementRemotingServices.MANAGEMENT_ENDPOINT, UndertowHttpManagementService.SERVICE_NAME);
```
bstansberry: Same comment as applies for the host controller variant applies here.

controller/src/main/java/org/jboss/as/controller/management/ManagementInterfaceAddStepHandler.java
```diff
@@ -0,0 +1,97 @@
...
+                    if (controller == null || State.UP != controller.getState()) {
+                        failures.add(serviceName);
+                    }
+                } catch (ServiceNotFoundException ex) {
+                    failures.add(serviceName);
+                }
+            }
+            if (failures.containsAll(requiredServices)) {
+                context.setRollbackOnly();
+                ModelNode requiredServicesNode = context.getFailureDescription().get(ROOT_LOGGER.missingRequiredServices());
```
bstansberry: You don't need to deal with the failure description. The existing verification steps will already output the correct  failure description. All this verification needs to do is force the rollback with the context.setRollbackOnly().

controller/src/main/java/org/jboss/as/controller/management/ManagementInterfaceAddStepHandler.java
```diff
@@ -0,0 +1,97 @@
...
+                try {
+                    ServiceController<?> controller = registry.getService(serviceName);
+                    if (controller == null || State.UP != controller.getState()) {
+                        failures.add(serviceName);
+                    }
+                } catch (ServiceNotFoundException ex) {
+                    failures.add(serviceName);
+                }
+            }
+            if (failures.containsAll(requiredServices)) {
```
bstansberry: Ah, now I get where ManagementRemotingServices.MANAGEMENT_ENDPOINT was coming from in the HTTP cases. I think you're going for an "either-or" thing where you check all possible services and if any are ok, it's ok.

ManagementRemotingServices.MANAGEMENT_ENDPOINT doesn't suffice though, as the endpoint being up doesn't mean the socket to receive native is up (for native) or the service to do HTTP Upgrade is up (for http.)

I think the required services should be those known to be required by a given handler subclass and no others.  The subclasses should not need to know the details of other subclasses, as that will be fragile.

To be lenient and not fail so long as 1 subclass succeeds, without requiring the various subclasses to know each others' details:

1) ManagementInterfaceAddStepHandler can define a constant OperationContext.AttachmentKey, named for example REMOTE_MANAGEMENT.
2) VerifyInstallationStep checks all the required services, which are only the ones relevant to the given subclass. 
a) If all are ok, it attaches REMOTE_MANAGEMENT to the context.
b) If any are not ok, it doesn't call context.setRollbackOnly() immediately. Instead it adds another VERIFY step. That step checks for the present of context attachment REMOTE_MANAGEMENT. If present, some other handler succeeded so all is well and it just returns. If not present, then none have succeeded and it calls context.setRollbackOnly().

server/src/main/java/org/jboss/as/server/operations/HttpManagementAddHandler.java
```diff
@@ -184,7 +183,15 @@ protected void installServices(OperationContext context, HttpInterfaceCommonPoli
 
             RemotingHttpUpgradeService.installServices(serviceTarget, ManagementRemotingServices.HTTP_CONNECTOR, httpConnectorName,
                     ManagementRemotingServices.MANAGEMENT_ENDPOINT, commonPolicy.getConnectorOptions());
+            if (socketBindingServiceName != null) {
+                return Arrays.asList(socketBindingServiceName, UndertowHttpManagementService.SERVICE_NAME, HTTP_UPGRADE_REGISTRY.append(httpConnectorName));
+            }
+            return Arrays.asList(UndertowHttpManagementService.SERVICE_NAME, HTTP_UPGRADE_REGISTRY.append(httpConnectorName));
+        }
+        if (socketBindingServiceName != null) {
```
bstansberry: This check and the one at L186 shouldn't be needed as UndertowHttpManagementService.SERVICE_NAME will depend on the socket binding service if this var isn't null.

server/src/main/java/org/jboss/as/server/operations/HttpManagementAddHandler.java
```diff
@@ -184,7 +183,15 @@ protected void installServices(OperationContext context, HttpInterfaceCommonPoli
...
+            return Arrays.asList(socketBindingServiceName, UndertowHttpManagementService.SERVICE_NAME);
         }
+        return Arrays.asList(UndertowHttpManagementService.SERVICE_NAME);
```
bstansberry: This can be Collections.singletonList() (IDEA complains about this.)

host-controller/src/main/java/org/jboss/as/host/controller/operations/HttpManagementAddHandler.java
```diff
@@ -172,11 +176,9 @@ protected void installServices(OperationContext context, HttpInterfaceCommonPoli
 
             RemotingHttpUpgradeService.installServices(serviceTarget, ManagementRemotingServices.HTTP_CONNECTOR, httpConnectorName,
                     ManagementRemotingServices.MANAGEMENT_ENDPOINT, commonPolicy.getConnectorOptions());
+            return Arrays.asList(UndertowHttpManagementService.SERVICE_NAME, HTTP_UPGRADE_REGISTRY.append(httpConnectorName));
         }
-    }
-
-    private ConsoleMode consoleMode(boolean consoleEnabled, boolean adminOnly) {
-        return consoleEnabled ? adminOnly ?  ConsoleMode.ADMIN_ONLY : ConsoleMode.CONSOLE : ConsoleMode.NO_CONSOLE;
+        return Arrays.asList(UndertowHttpManagementService.SERVICE_NAME);
```
bstansberry: This can be Collections.singletonList() (IDEA complains about this.)

bstansberry:
A standalone server can also use NativeRemotingManagementAddHandler so this should cover that too.

bstansberry:
retest this please

bstansberry:
retest this please

PR #1675, opened 2016-07-13
[WFCORE-429]: Incremental redeployment (single file update) over management API

ehsavoie:
HashUtil for hashing exploded deployments.
PathUtil for recursive deleting/copying.
add-content, remove-content, read-conten, browse-content, explode in standalone and domain.
Support for empty directories and empty files.
Support for sub content explode.
Preparing CLI integration for streams.
Adding lock + local temp copy to avoid conflicts between read-content and remove deployment / content.
Adding "empty" parameter to avoid missing content in deployment.
Adding "overwrite" parameter to add-content to avoid creating an update-content operation.
Adding transformers for retro compatibility taking unmanaged deployments into account.

Jira: https://issues.jboss.org/browse/WFCORE-429
Analysis document: https://developer.jboss.org/docs/DOC-55497

There is a branch for some more integration tests on WildFly
https://github.com/ehsavoie/wildfly/tree/WFCORE-429
Integration with the update in full: https://ci.wildfly.org/viewLog.html?buildId=21461&tab=buildResultsDiv&buildTypeId=WF_WildFlyCoreIntegrationExperiments


deployment-repository/src/main/java/org/jboss/as/repository/ContentRepository.java
```diff
@@ -72,7 +58,11 @@
      * Currently 5 minutes.
      */
     long OBSOLETE_CONTENT_TIMEOUT = getSecurityManager() == null ? getLong(Factory.UNSUPPORTED_PROPERTY, 300000L) : doPrivileged((PrivilegedAction<Long>) () -> getLong(Factory.UNSUPPORTED_PROPERTY, 300000L));
-
+    /**
+     * Timeout duiration to get a lock on a deployment for removing it or reading some content.
+     * Currently 1 minutes.
+     */
+    long LOCK_TIMEOUT = 60000L;
```
bstansberry: This seems long, unless there's reason to expect that something preventing the lock would go away. We want to avoid failing due to a fairly long full gc or something but beyond that if it's not going to unlock the op shouldn't hang around waiting.

deployment-repository/src/main/java/org/jboss/as/repository/RemoteFileRequestAndHandler.java
```diff
@@ -172,21 +169,28 @@ private void writeResponse(final File localPath, final FlushableDataOutput outpu
...
     private List<File> getChildFiles(final File base) {
-        final List<File> childFiles = new ArrayList<File>();
+        final List<File> childFiles = new ArrayList<>();
         getChildFiles(base, childFiles);
         return childFiles;
     }
 
     private void getChildFiles(final File base, final List<File> childFiles) {
         for (File child : base.listFiles()) {
+            childFiles.add(child);
```
bstansberry: This seems odd, since L181 and L183 do the same.

host-controller/src/main/java/org/jboss/as/domain/controller/operations/deployment/DeploymentAddHandler.java
```diff
@@ -120,6 +128,15 @@ public void execute(OperationContext context, ModelNode operation) throws Operat
                 // Ensure the local repo has the files
                 fileRepository.getDeploymentFiles(ModelContentReference.fromModelAddress(address, hash));
             }
+        } else if (!contentItemNode.isDefined()) {
+            // Store and transform operation
+            hash = DeploymentUploadUtil.storeEmptyContentAndTransformOperation(context, correctedOperation, contentRepository);
+            contentItemNode = new ModelNode();
```
bstansberry: contentItemNode is already assigned to a new ModelNode.

server/src/main/java/org/jboss/as/server/deployment/ExplodedDeploymentAddContentHandler.java
```diff
@@ -0,0 +1,181 @@
...
+                        ExecutorService executor = (ExecutorService) context.getServiceRegistry(false).getRequiredService(JBOSS_SERVER_EXECUTOR).getValue();
+                        CountDownLatch latch = copy(executor, relativePaths, managementName, newHash);
+                        if (latch != null) {
+                            try {
+                                if (!latch.await(60, TimeUnit.SECONDS)) {
+                                    throw ServerLogger.ROOT_LOGGER.couldNotCopyFiles();
+                                }
+                            } catch (InterruptedException e) {
+                                Thread.currentThread().interrupt();
+                                throw createFailureException(e.toString());
```
bstansberry: Don't throw an exception, just return. You restored the interrupted status on the thread, so the OperationContext will detect that and go into its operation cancellation logic. Throwing an exception just adds noise to that.

server/src/main/java/org/jboss/as/server/deployment/ExplodedDeploymentRemoveContentHandler.java
```diff
@@ -0,0 +1,163 @@
...
+                        ExecutorService executor = (ExecutorService) context.getServiceRegistry(false).getRequiredService(JBOSS_SERVER_EXECUTOR).getValue();
+                        CountDownLatch latch = delete(executor, paths, managementName);
+                        if (latch != null) {
+                            try {
+                                if (!latch.await(60, TimeUnit.SECONDS)) {
+                                    throw ServerLogger.DEPLOYMENT_LOGGER.couldNotDeleteFiles();
+                                }
+                            } catch (InterruptedException ex) {
+                                Thread.currentThread().interrupt();
+                                throw createFailureException(ex.toString());
```
bstansberry: Don't throw an exception; just return; same reason as with the add handler.

server/src/main/java/org/jboss/as/server/deployment/RootDeploymentUnitService.java
```diff
@@ -65,12 +66,33 @@ public RootDeploymentUnitService(final String name, final String managementName,
                                      final ImmutableManagementResourceRegistration registration, final ManagementResourceRegistration mutableRegistration,
                                      final Resource resource, final CapabilityServiceSupport capabilityServiceSupport,
                                      final AbstractVaultReader vaultReader, DeploymentOverlayIndex deploymentOverlays) {
+        this(name, managementName, parent, registration, mutableRegistration, resource, capabilityServiceSupport, vaultReader, deploymentOverlays, false);
```
bstansberry: Unless this constructor is used outside of core you don't need to keep the old variant.

If it's used outside of core something is odd. :)

server/src/main/resources/org/jboss/as/server/controller/descriptions/LocalDescriptions.properties
```diff
@@ -223,9 +223,22 @@ deployment.enabled=Boolean indicating whether the deployment content is currentl
 deployment.persistent=Boolean indicating whether the existence of the deployment should be recorded in the persistent server configuration. Only relevant to a standalone mode server. Default is 'true'. A deployment managed by a deployment scanner would have this set to 'false' to ensure the deployment is only deployed at server start if the scanner again detects the deployment.
 deployment.owner=Address of a resource that has indicated it controls the lifecycle of this deployment (e.g. a deployment scanner resource.) Will be undefined if no resource claimed ownership when the deployment was added.
 deployment.add=Adds previously uploaded content to the list of content available for use. Does not actually deploy the content unless the 'enabled' parameter is 'true'.
+deployment.add-content=Add contents to an existing deployment.
+deployment.add-content=Add contents to an existing deployment.
```
bstansberry: duplicate

server/src/main/resources/org/jboss/as/server/controller/descriptions/LocalDescriptions.properties
```diff
@@ -223,9 +223,22 @@ deployment.enabled=Boolean indicating whether the deployment content is currentl
...
+deployment.add-content.overwrite=Iondicates if added content should overwrite existing content.
```
bstansberry: s/Iondicates/Indicates

server/src/main/resources/org/jboss/as/server/controller/descriptions/LocalDescriptions.properties
```diff
@@ -223,9 +223,22 @@ deployment.enabled=Boolean indicating whether the deployment content is currentl
...
+deployment.add-content.content=List of pieces of content to be added to the deployment.
+deployment.browse-content=List the pieces of content of a deployment.
+deployment.browse-content.path=The relative path from a deployment to list the contents of.
+deployment.browse-content.archive=To return only the archive file relative paths.
```
bstansberry: The meaning of this one isn't very clear.

server/src/main/resources/org/jboss/as/server/controller/descriptions/LocalDescriptions.properties
```diff
@@ -223,9 +223,22 @@ deployment.enabled=Boolean indicating whether the deployment content is currentl
...
+deployment.browse-content.depth=The depth to browse.
 deployment.deploy=Deploy the specified deployment content into the runtime, optionally replacing existing content.
 deployment.deploy.name=The name of the new content.
 deployment.deploy.to-replace=The name of the content that is to be replaced.
+deployment.empty=Indicates that the deployment to be added is empty - so without any content.
+deployment.read-content=Read the content of an existing deployment.
+deployment.read-content.path=The relative path of the content to be read from an existing deployment.
+deployment.remove-content=Remove contents from an existing deployment.
+deployment.remove-content.path=List of paths of content to be removed from the deployment.
```
bstansberry: Should this param be 'paths' since it can be many? For add-content, the name of param 'content' doesn't imply only 1 but 'path' does.

bstansberry:
Please update the requirements doc to mention the requirement to enhance the controller-client module helper classes. That's a significant added piece that needs to be mentioned so QE/docs etc are aware of it.

When you do, please add a brief comment noting that you made the change. The version history of the doc will record the change but the comment will help people tracking it be aware that there was one.

bstansberry:
Does the new "empty" param fit better as a field in the "content" param instead of being a top-level parameter to the add op? It is similar to "bytes", "url" etc in that it's an instruction to the server as to how to access the content.

host-controller/src/main/java/org/jboss/as/domain/controller/logging/DomainControllerLogger.java
```diff
@@ -768,4 +768,43 @@
...
+    @Message(id = 92, value = "Cannot remove content from an unmanaged deployment")
+    OperationFailedException cannotRemoveContentFromUnmanagedDeployment();
+
+    @Message(id = 93, value = "Cannot remove content from an unexploded deployment")
+    OperationFailedException cannotRemoveContentFromUnexplodedDeployment();
+
+    @Message(id = 94, value = "Cannot read content from a deployment in a self-contained server")
+    OperationFailedException cannotReadContentFromSelfContainedServer();
+
+    @Message(id = 95, value = "Cannot remove content from an unmanaged deployment")
```
claudio4j: s/remove/read

host-controller/src/main/java/org/jboss/as/domain/controller/logging/DomainControllerLogger.java
```diff
@@ -768,4 +768,43 @@
...
+    OperationFailedException cannotReadContentFromUnmanagedDeployment();
+
+    @Message(id = 96, value = "Cannot remove content from an unexploded deployment")
```
claudio4j: s/remove/read

bstansberry:
@ehsavoie I'm feeling more strongly now that the 'empty' setting should be a field in the existing 'content' param/attribute and not a new top level param.

The ModelControllerMBeanTestCase failure shows that adding it as a new top level param results in a breaking change for JMX clients. I thought for a while that making it as a field in 'content' would also have a breaking effect, so it would be a breaking change either way. But that is not the case. The way ModelControllerMBeanTestCase was implemented fooled me into thinking the requirements for creating the 'content' param in JMX was that all fields must be used, including any new 'empty' field. But that's not true. See #1745 where I change to test to use simpler JMX code, something more like what real JMX-based end users would likely do. The user only needs to set the fields that are relevant to them, and code that was unaware of 'empty' would continue to work unmodified.

WDYT?

PR #1680, opened 2016-07-15
WFCORE-1657 Confusing tab completion for adding a module dependencies

spyrkob:
JIRA: https://issues.jboss.org/browse/WFCORE-1657

### DIFF
--- a/cli/src/main/java/org/jboss/as/cli/handlers/ModuleNameTabCompleter.java
+++ b/cli/src/main/java/org/jboss/as/cli/handlers/ModuleNameTabCompleter.java
@@ -1,0 +1,243 @@
+/*
+ * JBoss, Home of Professional Open Source
+ * Copyright 2016, JBoss Inc., and individual contributors as indicated
+ * by the @authors tag.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.jboss.as.cli.handlers;
+
+import org.jboss.as.cli.EscapeSelector;
+import org.jboss.as.cli.Util;
+
+import java.io.File;
+import java.util.ArrayList;
+import java.util.Arrays;
+import java.util.Collection;
+import java.util.List;
+import java.util.Set;
+import java.util.TreeSet;
+
+/**
+ * Generates suggestions for module names. Each suggestion generates only next part of the name (ie. up to the name separator).
+ *
+ * Assumes the module repository used has standard layered repository layout. Matching suggestions are found:
+ * <uL>
+ *  <li>under the repository root (excluding 'system' directory)
+ *  <li>under the system/layers/{layer name}
+ *  <li>under the system/add-ons/{add-on name}
+ * </uL>
+ * Modules changed, removed or added via patches are not included in the suggestions.
+ * The modules are not validated - invalid or disabled modules and empty directories are included in the suggestions.
+ *
+ * @author Bartosz Spyrko-Smietanko
+ */
+public class ModuleNameTabCompleter {
+
+    private static final EscapeSelector ESCAPE_SELECTOR = ch -> ch == '\\' || ch == ' ' || ch == '"';
+    private static final String MODULE_NAME_SEPARATOR = ".";
+    public static final String LAYERS_DIR = "system/layers";
+    public static final String ADDONS_DIR = "system/add-ons";
+
+    private final File modulesRoot;
+    private final File layersDir;
+    private final File addonsDir;
+    private final boolean includeSystemModules;
+    private final boolean excludeNonModuleFolders;
+
+    private ModuleNameTabCompleter(Builder builder) {
+        modulesRoot = builder.modulesRoot.getAbsoluteFile();
+        layersDir = new File(modulesRoot, LAYERS_DIR);
+        addonsDir = new File(modulesRoot, ADDONS_DIR);
+
+        this.excludeNonModuleFolders = builder.excludeNonModuleFolders;
+        this.includeSystemModules = builder.includeSystemModules;
+    }
+
+
+    public List<String> complete(String buffer) {
+        final String userEntry = buffer == null ? "" : buffer;
+        final Set<String> suggestions = new TreeSet<>(); // TreeSet deals with duplication and ordering
+
+        List<File> moduleTrees = findInitialModuleDirectories();
+
+        moduleTrees.forEach(f -> findSuggestion(f, f.getName(), userEntry, suggestions));
+
+        return new ArrayList<>(suggestions);
+    }
+
+    private List<File> findInitialModuleDirectories() {
+        List<File> moduleTrees = new ArrayList<>();
+
+        moduleTrees.addAll(Arrays.asList(modulesRoot.listFiles(this::isNotSystemFolder)));
+
+        if (includeSystemModules && layersDir.exists()) {
+            for (File layer : layersDir.listFiles(File::isDirectory)) {
+                moduleTrees.addAll(Arrays.asList(layer.listFiles(this::isNotPatchFolder)));
+            }
+        }
+
+        if (includeSystemModules && addonsDir.exists()) {
+            for (File addon : addonsDir.listFiles(File::isDirectory)) {
+                moduleTrees.addAll(Arrays.asList(addon.listFiles(this::isNotPatchFolder)));
+            }
+        }
+
+        return moduleTrees;
+    }
+
+    private void findSuggestion(File currentDirectory, String suggestion, String userEntry, Collection<String> candidates) {
+        if (!matchesUserEntry(currentDirectory, userEntry) || (excludeNonModuleFolders && isSlotDirectory(currentDirectory))) {
+            return;
+        }
+
+        if (tail(userEntry).isEmpty() && !requestsSubmodules(userEntry)) {
+            final String fullModuleName = Util.escapeString(suggestion, ESCAPE_SELECTOR);
+            final String partialModuleName = Util.escapeString(suggestion + MODULE_NAME_SEPARATOR, ESCAPE_SELECTOR);
+
+            if (excludeNonModuleFolders) {
+                final boolean isExactMatch = currentDirectory.getName().equals(userEntry);
+                final boolean hasNestedModules = hasNestedModules(currentDirectory);
+                final boolean isCompleteModule = isCompleteModule(currentDirectory);
+
+
+                /*
+                The suggestion should have a trailing separator ('.') if it's a part of longer module name.
+                If the suggested name is both a full module name and a part of longer name (ie. nested modules), suggest
+                the name without separator - unless user input is a complete name in which case suggest both options.
+                 */
+                if (isCompleteModule && hasNestedModules && isExactMatch) {
+                    candidates.add(fullModuleName);
+                    candidates.add(partialModuleName);
+                } else if (isCompleteModule) {
+                    candidates.add(fullModuleName);
+                } else if (hasNestedModules) {
+                    candidates.add(partialModuleName);
+                }
+            } else {
+                final boolean hasChildren = currentDirectory.listFiles(File::isDirectory).length > 0;
+                final boolean isExactMatch = currentDirectory.getName().equals(userEntry);
+
+                if (hasChildren && isExactMatch) {
+                    candidates.add(partialModuleName);
+                }
+                candidates.add(fullModuleName);
+            }
+        } else {
+            for (File file : currentDirectory.listFiles(File::isDirectory)) {
+                findSuggestion(file, suggestion + MODULE_NAME_SEPARATOR + file.getName(), tail(userEntry), candidates);
+            }
+        }
+    }
+
+    private boolean matchesUserEntry(File currentDirectory, String userEntry) {
+        if (!userEntry.endsWith(MODULE_NAME_SEPARATOR) && tail(userEntry).isEmpty()) {
+            return currentDirectory.getName().startsWith(head(userEntry));
+        } else {
+            return currentDirectory.getName().equals(head(userEntry));
+        }
+    }
+
+    private boolean isCompleteModule(File file) {
+        return file.listFiles(f -> f.isDirectory() && isSlotDirectory(f)).length > 0;
+    }
+
+
+    private boolean hasNestedModules(File file) {
+        final File[] nonSlotChildren = file.listFiles(f -> f.isDirectory() && !isSlotDirectory(f));
+        for (File potentialModule : nonSlotChildren) {
+            if (subModuleExists(potentialModule)) {
+                return true;
+            }
+        }
+        return false;
+    }
+
+    // depth- first search for any module - just to check that the suggestion has any chance of delivering correct result
+    private boolean subModuleExists(File dir) {
+        if (isSlotDirectory(dir)) {
+            return true;
+        } else {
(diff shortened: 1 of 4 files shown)


spyrkob:
retest this please

bstansberry:
@aloubyansky or @jfdenise please approve

jfdenise:
Hi, the implementation and tests are very good. 
One thing that I would change is the content exposed by the completer. The list of all module names is presented upfront. For example, module add --name=<TAB> should only expose the roots: jdk, ibm, io, org, ... Doing so we would take benefit of navigating a tree. Less information would be exposed to the user. This should be more scalable specially for systems where you have 100s of modules.
JF

aloubyansky:
Thanks Bartosz! I am with Jeff on this point though.

spyrkob:
I made the change to list the roots instead of full names. One exception is when there is a nested module eg "jboss.as.patching" and "jboss.as.patching.cli" will generate:
jboss.as.patching
jboss.as.patching.

Let me know if the change looks good, I'll squash the commits and upload.

jfdenise:
Hi, thank you for the update, the completion behaviour is just fine! I am wandering if retrieving all modules then doing the filtering could have a performance impact. Not sure you can do something there, you are now retrieving the name from the module.xml so you need to read them all. If this is becoming an issue, we could cache the module names and clear the cache when a module command is executed.

spyrkob:
@jfdenise I've got a version that only descends a tree if it matches user's query here: https://github.com/spyrkob/wildfly-core/commit/0da0bd1efa08adc8804c233826336b5e0ae5b598 . That should improve the performance in case user provides some input.

If the performance is an issue I can try to get rid of reading module.xml files - I went this route because it takes care of module slots and avoids suggesting directories that are not modules.

jfdenise:
Hi, I was thinking at filtering based on the user input, the downside is that it should make you hardcode the knowledge of the layers file system structure. But that is perhaps fine if this is not changing often. 

We should possibly check with @dmlloyd if there is a way to discover the layer structure (or if it is not supposed to change).

How is the slot structured on file system? This is the last directory, right? If this is all the time the last directory (main, xxx), then scanning the file system should be enough to retrieve the module names. And you could do that on the fly (based on user input as you are suggesting).

spyrkob:
> How is the slot structured on file system? This is the last directory, right? If this is all the time the last directory (main, xxx), then scanning the file system should be enough to retrieve the module names. And you could do that on the fly (based on user input as you are suggesting).

That's the problem - as far as I can tell the slot can be a nested directory - eg. for this directory structure:
```
com
└── module
    └── foo
        └── bar
            └── module.xml
```
both
```
<module xmlns="urn:jboss:module:1.3" name="com.module" slot="foo/bar">
</module>
```
and
```
<module xmlns="urn:jboss:module:1.3" name="com.module.foo" slot="bar">
</module>
```
seems to be valid. That might be not intended use case, but it seems possible...

jfdenise:
Looking at the modules descriptor schema, the slotType shouldn't contain any "/" character.
https://github.com/jboss-modules/jboss-modules/blob/master/src/main/resources/schema/module-1_6.xsd
Where did you find a slot containing "/"? Inside a test? This is perhaps a wrong test.

spyrkob:
I only run into that issue when doing some testing - sorry I didn't think to check in the module descriptor.

Updated the completer to use the filesystem approach.

jfdenise:
Hi,
this is definitely a more efficient approach. Thank-you for the update. We will need to see how we can get rid-off hard coded layers root dir.
Some notes:
- Look at the slot completer in the ASModuleHandler class, I think that it needs to be updated to have it reconstruct the file system path to the slots. No unit tests caught it, perhaps adding a unit test would be a nice addition.
- is findInitialModelDirectories the right method name? requestsSubmodules could be capitalised.
- ModuleNameTabCompleter class could be a package class in handlers.module dir.

bstansberry:
Something to consider:

These commands eventually need to become proper remote commands and stop relying on local filesystem access.

When then happens, how is the functionality going to be provided? We shouldn't add behavior that we may not be able to continue to provide in the future.

In practice, doing this remotely may be easier, as the remote process will definitely be modular (unlike the possibly the non-modular CLI) and can integrate with JBoss Modules to determine the names of available modules, e.g. with ModuleLoader.iterateModules(...). No problems with interpreting system/layers, system/add-ons etc. So perhaps this isn't a problem, it's just something to think about.

cli/src/main/java/org/jboss/as/cli/handlers/ModuleNameTabCompleter.java
```diff
@@ -0,0 +1,153 @@
...
+    private List<File> findInitialModelDirectories() {
+        List<File> moduleTrees = new ArrayList<>();
+
+        moduleTrees.addAll(Arrays.asList(modulesRoot.listFiles(f -> isNotLayersFolder(f))));
+
+        if (layersDir.exists()) {
+            for (File layer : layersDir.listFiles(File::isDirectory)) {
+                moduleTrees.addAll(Arrays.asList(layer.listFiles(f -> isNotLayersFolder(f))));
+            }
+        }
```
bstansberry: This needs updating.

The valid locations that may be the first level of a module directory tree are:

1) Child dirs of the module root that are not named 'system'. That's line 67 above.
2) Child dirs of any dir whose parent dir is <modulesDir>/system/layers. So the 'isNotLayersFolder' check on L71 is not correct as <modulesDir>/system/layers/system would be a valid location if someone produced a layer named 'system'.
3) Child dirs of any dir whose parent dir is <modulesDir>/system/addons.

bstansberry:
My last line note doesn't consider the impact of patches, which adds further complication. See "Module Organization on the Filesystem" on https://developer.jboss.org/docs/DOC-47927 for details.

I'm not really a fan of hard coding this all into the CLI. Since this likely has been broken since EAP 6.1 in 2013, until we make this command a proper remote command I think we should consider simply excluding 'system' and anything below it from the tab completion. That preserves the part of tab completion that has worked since EAP 6.1 while also fixing the problem of introducing incorrect module names.

spyrkob:
I spoke with @dmlloyd about it - he also suggested using the ModuleLoader.iterateModules instead of hardcoding the filesystem usage in CLI. I opened a JBoss Modules JIRA ([MODULES-262](https://issues.jboss.org/browse/MODULES-262)) to implement module enumeration on LocalModuleFinder.

I think performance might be a concern with this approach though. When implementing the tree browsing specifically for the suggestion generation we can do some optimization (eg. stopping at the first directory level rather than getting full module name) - that might not be the case with a more generic API.

Until that's resolved should I update this PR to @bstansberry's suggestion from previous comment (ie. ignore any modules in 'system')?

bstansberry:
I assume the ModuleLoader.iterateModules implementation will be efficient enough that perf is not a significant concern at that level. The issue then becomes taking that full set of output and presenting it efficiently to the remote client. Since the remote client will be asking for this data via management operations against a new part of the management API we'd add, that API can include whatever params are needed to let the client instruct  the server to restrict output.

Alternatively the server can send the full set of data and the CLI can cache it in the CommandContext for the life of the request. That may prove just as efficient overall since there would only be one round trip.

bstansberry:
FYI: https://issues.jboss.org/browse/WFCORE-1608 would be the JIRA for a remote management API to list modules.

bstansberry:
@jfdenise Bartosz's question above is for you: "Until that's resolved should I update this PR to @bstansberry's suggestion from previous comment (ie. ignore any modules in 'system')?"

jfdenise:
Hi, yes, I missed this one. 
I am not sure @bstansberry wants to ignore any modules in 'system'. 
@bstansberry seems to list the pieces of paths that should be removed from the module name.
In bracket, the pieces that should be removed from the module name:
 <modules root dir>/<a module not named system>/
 <modules root dir>[/system/addons/<addonName>]/<a module>
 <modules root dir>[/system/layers/<layerName>]/<another module>

bstansberry:
@jfdenise 
I did propose just ignoring 'system'. The details of what you described removing from the paths in your last comment are correct for a non-patched server, but once the server is patched it gets more complex. See "Module Organization on the Filesystem" on https://developer.jboss.org/docs/DOC-47927 . It's that complexity that starts to seem like too much to hard code in the CLI.

We could do this, removing the bracketed parts of the paths:

{modules root dir}/{a module not named system}/
{modules root dir}[/system/addons/{addonName}]/{a module not named patches}
{modules root dir}[/system/layers/{layerName}]/{another module not named patches}

I guess that's not so complex. But, that would not allow completion to find a module *added* in a patch. I believe it would also not detect the fact that a module had been disabled in a patch, and would still suggest those. Neither of those are really that big of a deal; as neither are common and after all this is just tab completion we are talking about. The user really shouldn't be counting on tab completion to tell him what the valid modules are to depend on. That should be a known thing when whatever module is being added was designed. Tab completion here is just an aid in typing.

So, I'm ok either way; doing it as described ^^^ or just keeping it simple, ignoring 'system' and just solving the problem of tab completion presenting incorrect paths. Whatever you and Alexey prefer.

jfdenise:
@bstansberry, skipping the system directory would hide the whole list of distribution modules from completion. Isn't it too radical? I must admit that I am not very familiar with the use cases of this command.

bstansberry:
@jfdenise I wouldn't say 'radical'; I'd say 'hopelessly conservative". ;-) It's more like the bare minimum correction. It's only an option because right now the tab completion cannot correctly expose any of those modules anyway since it prefixes them with irrelevant parts of the path. So we fix the bug of 'wrong suggestions' but don't address the problem of 'correct suggestions.'

TBH when I first suggested it I hadn't considered the option of ignoring the "patches" subdirectories. I was thinking more about the CLI needing to actually look into those patches subdirectories for modules. And doing that IMO is just too much. But ignoring "patches" doesn't sound too bad.

jfdenise:
@bstansberry , thank-you. I understand your opinion. 

+1 on doing a bit of path parsing to extract the bracketed part.

aloubyansky:
Right, it's too much to take into account patching overlays here, although w/o those the tab-completion won't be accurate. Not only specifying dependencies but also for general module add/remove. It would have to consult the patching state to suggest the right names and locations. On the other hand, a user should not really mess with patching overlays directly (even using this CLI command) but only through the patching interface. The same probably for the modules under the system directory.

So ignoring the system dir would be a good guard against misuse, I guess, but pretty sure people will get upset about it.
I guess, document that you are not supposed to mess with the system modules, let the tab-completion still show them but also document that the completion is not comprehensive in this implementation.

bstansberry:
Good point @aloubyansky about add/remove. The completion requirements for the --name parameter are different than for --dependencies. This command only supports adding and removing user supplied modules. It is not meant to work for removing system modules. See the impl of ASModuleHandler.removeModule(...) to see what I mean.[1] And we shouldn't try to change that at this point.

So if we are tab completing the --name param for the 'remove' subcommand the 'system' folder should be ignored.

For the 'add' subcommand, all it's doing is helping to type the name of a new module by suggesting already used bits of names. I don't have a strong opinion one way or the other as to whether suggesting bits from modules in 'system' is helpful there.

@jfdenise I'm putting a fixme label on this so I can stop monitoring it until it's ready. Please @ mention me when it is, or if there is further stuff to discuss.
 
[1] A user could perhaps hack the command to remove a system module via something like:

module remove --name=system.layers.base.org.wildfly.controller --slot=main

But tab completion shouldn't suggest that. I don't consider no longer suggesting that to be a regression.

spyrkob:
The completer works in 3 different ways now depending on context:
1) on `module add --name=<TAB>` it will suggest folders in the modules root (ie. ignoring system folder, but listing any other folder even if it is not a module name
2) on `module remove --name=<TAB>` it will suggest modules in the modules root ignoring system modules. It will not suggest directory trees that have no modules.xml
3) on `module add --name=foo.bar --dependecies=<TAB>` it will suggest any module from modules root, layers or add-ons

For case 2 & 3 I'm descending a directory tree depth-first to verify that it's a valid suggestion (has an actual module). Apart from some edge cases that should have better performance than finding all modules first.

Since the module descriptors are not parsed, disabled or invalid modules will still be suggested.

jfdenise:
Hi, thank you for the updated fixe. 
I think that proposing the existing system module names for add --name, is a good help. You should perhaps re-enable it?

The completion doesn't seem to work (and the added unit test didn't catch it):
module add --name=cscds --dependencies=<TAB>
ibm.  io.  javax.  org.  sun.  
module add --name=cscds --dependencies=o<TAB>
Nothing is proposed, org is not inlined.

One note on unit tests. In order to activate the whole feature, you should look at also adding tests in the testsuite/standalone directory. Look at the CLIWrapper, this is a good way to emulate user inputs. Such a test would have caught the regression.

Thank-you.
JF

spyrkob:
@jfdenise sorry, I must have failed to rebuild cli before manual tests...

I enabled system directory for add operation and added a test to the integration suite. 
Thanks

jfdenise:
Hi,
I have tested a bit your last changes. I am wandering if ModuleNameTabCompleter can properly deal with both existingModulesOnly and includeSystemModules. The "only" part makes me think that it is disjunctive with systemModules. But I guess that this is only a naming thing.

The issue I am observing with add:
--name=<TAB> ==> all modules (my 3 modules and system presented). That is fine.
--name=ibm.jdk.<TAB> ==> main is proposed although it is a slot directory.

with dependencies:
--dependencies=<TAB> ==> some modules are missing (I have created 3 modules mymodule, mymodule2 and mymodule3, only mymodule2 is listed along with the system modules.
--dependencies=sun.scripting<TAB> ==> space character is added at the end.
--dependencies=sun.scripting,sun.jdk<TAB> ==> Nothing happens, no space character added.

Otherwise, the --dependencies completer works just fine!

spyrkob:
> I am wandering if ModuleNameTabCompleter can properly deal with both existingModulesOnly and includeSystemModules. The "only" part makes me think that it is disjunctive with systemModules. But I guess that this is only a naming thing.

existingModulesOnly flag was meant to mean "do not suggest folders if they're not a part of a module name (ie. don't have a module.xml somewhere down the tree)". Maybe "excludeNonModuleFolders" would be better description.

> --name=ibm.jdk. ==> main is proposed although it is a slot directory.

I wasn't sure how to deal the slot dirs in case of --name suggestions because the slot directory could be used as a part of a new, nested module name. Though I guess that's not a common use case and we shouldn't encourage it...
>--dependencies= ==> some modules are missing (I have created 3 modules mymodule, mymodule2 and mymodule3, only mymodule2 is listed along with the system modules.

Do all of those have a slot with module.xml descriptor? --dependencies will only suggest trees that has at least one valid module
>--dependencies=sun.scripting ==> space character is added at the end.
>--dependencies=sun.scripting,sun.jdk ==> Nothing happens, no space character added.

I was looking into that issue in the morning - the same thing happens for example on deployment-overlay's --contents completion. It seems to be caused by a condition in [OperationRequestCompleter](https://github.com/wildfly/wildfly-core/blob/bc4988960e9bd245318321a06ae3d001ec661bc5/cli/src/main/java/org/jboss/as/cli/operation/OperationRequestCompleter.java#L233) - if the completer returns valueResult > 0, only a part of the chunk should be matched against the suggestions (in this case chunk = 'sun.scripting,sun.jdk'; suggestion='sun.jdk'; valueResult=13).
I was going to open a new issue for this since it affects other completers as well

jfdenise:
- +1 on excludeNonModuleFolders 
- Good point for dependencies, no, only the displayed one has a module.xml. So that is correct.
- Just log a new issue, and don't talk it there.

Once you are done, just squash the commits, I will do a final review and it will be good to merge!
Thank-you!
JF

jfdenise:
Thank you for the squash. My testing caught a strange case:
module add --name=i.<TAB>
Nothing should be proposed although everything starting by i is proposed.

spyrkob:
Ups I didn't think to test that case... fixed now, sorry.

jfdenise:
Hi, thank you for the fix. Works fine! 
Minor point, there is 1 or 2 places that still contain the '.' character, you could replace it with the SEPARATOR you introduced. 
Other than that this is good to merge. As soon as you make the change, I will ask Brian to merge it.
Thanks again.
JF

spyrkob:
Updated the PR with the '.' change. Thanks!

bstansberry:
:-)

PR #1682, opened 2016-07-19
[WFCORE-1683] Update XNIO to 3.4.0.Beta3

dmlloyd:


bstansberry:
We can produce another core 2.2.0CR with a beta in it but for .Final we'll need a .Final of this.

PR #1683, opened 2016-07-19
WFCORE-823 Introduce new way for registering transformers

ctomc:
An idea of new way of registering transformers via service loader.

### DIFF
--- a/controller/src/main/java/org/jboss/as/controller/extension/ExtensionRegistry.java
+++ b/controller/src/main/java/org/jboss/as/controller/extension/ExtensionRegistry.java
@@ -464,6 +464,7 @@ public class ExtensionRegistry {
         }
     }
 
+    @SuppressWarnings("deprecation")
     private class ExtensionContextImpl implements ExtensionContext, ExtensionContextSupplement {
 
         private final ExtensionInfo extension;
@@ -532,8 +533,12 @@ public class ExtensionRegistry {
             if (deprecated){
                 ControllerLogger.DEPRECATED_LOGGER.extensionDeprecated(name);
             }
-            return new SubsystemRegistrationImpl(name, version,
-                    profileRegistration, deploymentsRegistration, extensionRegistryType, extension.extensionModuleName, processType);
+            SubsystemRegistrationImpl result =  new SubsystemRegistrationImpl(name, version,
+                                profileRegistration, deploymentsRegistration, extensionRegistryType, extension.extensionModuleName, processType);
+            if (registerTransformers){
+                transformerRegistry.loadAndRegisterTransformers(name, version, extension.extensionModuleName);
+            }
+            return result;
         }
 
         @Override
@@ -638,6 +643,11 @@ public class ExtensionRegistry {
             return version != null ? version.getMicro() : null;
         }
 
+        @Override
+        public ModelVersion getManagementInterfaceVersion() {
+            return version;
+        }
+
         private void setVersion(ModelVersion version) {
             this.version = version;
         }
@@ -648,15 +658,6 @@ public class ExtensionRegistry {
 
         private void setDeprecated(boolean deprecated) {
             this.deprecated = deprecated;
-        }
-
-
-        private void setHostCapable() {
-            hostCapable = true;
-        }
-
-        public boolean isHostCapable() {
-            return hostCapable;
         }
     }
 
--- a/controller/src/main/java/org/jboss/as/controller/extension/SubsystemInformation.java
+++ b/controller/src/main/java/org/jboss/as/controller/extension/SubsystemInformation.java
@@ -24,6 +24,8 @@ package org.jboss.as.controller.extension;
 
 import java.util.List;
 
+import org.jboss.as.controller.ModelVersion;
+
 /**
 * @author <a href="mailto:tomaz.cerar@redhat.com">Tomaz Cerar</a>
 */
@@ -56,4 +58,19 @@ public interface SubsystemInformation {
      * @return the micro interface version, or {@code null} if the subsystem does not have a versioned interface
      */
     Integer getManagementInterfaceMicroVersion();
+
+
+    /**
+     * Gets full version of the subsystem
+     * @return full {@link ModelVersion}
+     */
+    default ModelVersion getManagementInterfaceVersion(){
+        return ModelVersion.create(getManagementInterfaceMajorVersion() != null ? getManagementInterfaceMajorVersion() : 0,
+                getManagementInterfaceMinorVersion() != null ? getManagementInterfaceMinorVersion() : 0,
+                getManagementInterfaceMicroVersion() != null ? getManagementInterfaceMicroVersion() : 0
+        );
+    }
+
+
+    boolean isDeprecated();
 }
--- a/controller/src/main/java/org/jboss/as/controller/transform/SubsystemTransformerRegistration.java
+++ b/controller/src/main/java/org/jboss/as/controller/transform/SubsystemTransformerRegistration.java
@@ -1,0 +1,69 @@
+/*
+ * JBoss, Home of Professional Open Source
+ * Copyright 2016, Red Hat, Inc., and individual contributors as indicated
+ * by the @authors tag.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.jboss.as.controller.transform;
+
+import org.jboss.as.controller.ModelVersion;
+import org.jboss.as.controller.ModelVersionRange;
+
+/**
+ * Subsystem transformers registration API
+ *
+ * Methods on this interface provide main transformers registration points for all {@link ExtensionTransformerRegistration} implementations.
+ *
+ * @author Tomaz Cerar (c) 2016 Red Hat Inc.
+ */
+public interface SubsystemTransformerRegistration {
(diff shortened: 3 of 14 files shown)


controller/src/main/java/org/jboss/as/controller/extension/ExtensionRegistry.java
```diff
@@ -532,8 +533,12 @@ public SubsystemRegistration registerSubsystem(String name, ModelVersion version
             if (deprecated){
                 ControllerLogger.DEPRECATED_LOGGER.extensionDeprecated(name);
             }
-            return new SubsystemRegistrationImpl(name, version,
-                    profileRegistration, deploymentsRegistration, extensionRegistryType, extension.extensionModuleName, processType);
+            SubsystemRegistrationImpl result =  new SubsystemRegistrationImpl(name, version,
+                                profileRegistration, deploymentsRegistration, extensionRegistryType, extension.extensionModuleName, processType);
+            if (registerTransformers){
+                transformerRegistry.loadAndRegisterTransformers(name, version, extension.extensionModuleName);
```
kabir: What about connector which contains several subsystems?

controller/src/main/java/org/jboss/as/controller/transform/description/TransformationDescription.java
```diff
@@ -174,5 +189,27 @@ public static TransformersSubRegistration register(TransformationDescription des
...
+        /**
+         * Register a transformation description as a sub-resource at a given {@linkplain SubsystemRegistration}.
+         *
+         * @param description the subsystem transformation description
+         * @param registration the subsystem registrations
+         * @param range the model version range the transformation applies to
+         * @return the create sub registration
+         */
+        public static TransformersSubRegistration register(TransformationDescription description, SubsystemTransformerRegistration registration, ModelVersionRange range) {
+            final TransformersSubRegistration subRegistration = registration.registerModelTransformers(range, description.getResourceTransformer(),
```
kabir: Perhaps some wrapper around registration to avoid code duplication with the other register() method?

controller/src/main/java/org/jboss/as/controller/extension/ExtensionRegistry.java
ctomc: hmm, good point. Will look into this.

controller/src/main/java/org/jboss/as/controller/transform/description/TransformationDescription.java
ctomc: Will take a look what could be done.

controller/src/main/java/org/jboss/as/controller/extension/ExtensionRegistry.java
ctomc: Fixes, there is not extra check to register only transformers that are for same subsystem as subsystem we are registering.

controller/src/main/java/org/jboss/as/controller/transform/description/TransformationDescription.java
ctomc: I fixed that, just diff is after this line so it is not seen.

controller/src/main/java/org/jboss/as/controller/transform/SubsystemTransformerRegistration.java
```diff
@@ -0,0 +1,65 @@
+/*
+ * JBoss, Home of Professional Open Source
+ * Copyright 2015, Red Hat, Inc., and individual contributors as indicated
```
darranl: Your work is so last year ;-)

darranl:
We may have a little bit of work to do to merge this with our branch assuming it is merged to master but TBH I don't see it as a major issue from a management perspective our changes are just a couple of additional attributes on existing resources to reference new capabilities.

controller/src/main/java/org/jboss/as/controller/transform/SubsystemTransformerRegistration.java
ctomc: Damn, yeah initial prototype was done than, as I wanted to squeeze it in EAP7 :-)

ctomc:
@bstansberry could we merge this?

subsystem-test/framework/src/main/java/org/jboss/as/subsystem/test/AbstractKernelServicesImpl.java
```diff
@@ -140,7 +140,7 @@ public ModelNode readTransformedModel(ModelVersion modelVersion) {
         return readTransformedModel(modelVersion, true);
     }
 
-    ExtensionRegistry getExtensionRegistry() {
+    protected ExtensionRegistry getExtensionRegistry() {
```
bstansberry: This method is still only called from within the package.

controller/src/main/java/org/jboss/as/controller/extension/SubsystemInformation.java
```diff
@@ -56,4 +58,25 @@
...
+        return ModelVersion.create(getManagementInterfaceMajorVersion() != null ? getManagementInterfaceMajorVersion() : 0,
+                getManagementInterfaceMinorVersion() != null ? getManagementInterfaceMinorVersion() : 0,
+                getManagementInterfaceMicroVersion() != null ? getManagementInterfaceMicroVersion() : 0
+        );
+    }
+
+
+    default boolean isDeprecated(){
+        return false;
+    }
```
bstansberry: This is an unrelated change and no code is using this method. If we're going to make an unrelated change around this we should drop the getter/setter from SubsystemInformationImpl as no one uses the property.

controller/src/main/java/org/jboss/as/controller/extension/SubsystemInformation.java
```diff
@@ -56,4 +58,25 @@
...
+
+    default boolean isHostCapable() {
+                return false;
+            }
```
bstansberry: This is an unrelated change and no code is using this method. If we're going to make an unrelated change around this we should drop the getter/setter from SubsystemInformationImpl as no one uses the property.

controller/src/main/java/org/jboss/as/controller/transform/SubsystemTransformerRegistration.java
```diff
@@ -0,0 +1,65 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.jboss.as.controller.transform;
+
+import org.jboss.as.controller.ModelVersion;
+import org.jboss.as.controller.ModelVersionRange;
+
+/**
```
bstansberry: Needs class javadoc to explain what this interface is.

controller/src/main/java/org/jboss/as/controller/transform/TransformerRegistration.java
```diff
@@ -0,0 +1,38 @@
...
+
+/**
+ * @author Tomaz Cerar (c) 2016 Red Hat Inc.
+ */
+public interface TransformerRegistration {
+
+    /** Return name of subsystem this transformer registration is for.
+     * @return non null name of the subsystem
+     */
+    String getSubsystemName();
```
bstansberry: This forces multiple implementations of the ServiceLoader for an extension with multiple subsystems. An alternative is to expose the subsystem name via a getter on SubsystemTransformerRegistration and trust the impl to use that getter and do the right thing.

I don't feel strongly about this though. We only have 1 extension with multiple subsystems anyway. :)

controller/src/main/java/org/jboss/as/controller/transform/TransformerRegistration.java
```diff
@@ -0,0 +1,38 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.jboss.as.controller.transform;
+
+/**
+ * @author Tomaz Cerar (c) 2016 Red Hat Inc.
+ */
+public interface TransformerRegistration {
```
bstansberry: Perhaps ExtensionTransformerRegistration?

The very general name bothers me, as most of our class names in the transformer area are kind of confusing. I can never figure out what is for what. And this interface is clearly limited to use by extensions.

ctomc: That name would definitely be better. Will change it.

controller/src/main/java/org/jboss/as/controller/extension/SubsystemInformation.java
ctomc: There was a reason I added that, but I cannot remember what it was. 
If and when I remember again I can send separate PR to do that.

controller/src/main/java/org/jboss/as/controller/extension/SubsystemInformation.java
ctomc: This one is tricky, as we have setters as we have "two" phase subsystem initializaton.
to remove setters we would need to set everything in constructor. 
But as getSubsystemInfo() is called by initialize() and initializeParsers() 
problem is that only initialize() has full data needed to construct it without setters.
In most cases parsers are initialized first which create empty SubystemInfoImpl and than when initialize() is called it gets that info and fills it with all extra data.
To properly fix this we would need to do some bigger refactoring which is out of scope of this PR. 
I can look into it afterwards but lets keep it out of this PR for now.

bstansberry: The getter is never used. The field in the implementation is also never read outside the getter. So the setter is pointless.

I don't want a pointless method added to an interface.

controller/src/main/java/org/jboss/as/controller/extension/SubsystemInformation.java
bstansberry: After you drop this from this PR, if you remember the reason for it you can restore this method along with the use of it.

controller/src/main/java/org/jboss/as/controller/transform/SubsystemTransformerRegistration.java
ctomc: added

bstansberry:
This was merged with an added commit on #1752. I don't know why github didn't mark this one as merged too like it normally does.

PR #1685, opened 2016-07-20
WFCORE-316 Slave host does not register to DC when starting with --cached-dc and DC is available

luck3y:


process-controller/src/main/java/org/jboss/as/process/logging/ProcessLogger.java
```diff
@@ -289,15 +289,15 @@
      *
      * @return the message.
      */
-    @Message(id = Message.NONE, value = "Keep a copy of the persistent domain configuration even if this host is not the Domain Controller")
+    @Message(id = Message.NONE, value = "Keep a complete copy of the persistent domain configuration even if this host is not the Domain Controller. This parameter will override any ignore-unused-configuration setting in host.xml and attempt to store the complete domain configuration.")
```
bstansberry: Is this the semantic? I think --backup controls the default behavior when ignore-unused-configuration is not set, but if it is set, the setting applies.

This allows a slave to keep a copy of the relevant parts of the domain wide config locally for its own use when the DC is down. It doesn't need that copy to have parts that are irrelevant to its servers.

luck3y: Ah, good catch, I'll double check.

luck3y: Yep, you're correct, --backup only sets it if there is no value set, otherwise whatever is set is used.

Do you think we should warn if this is set like this? (I think we should). It seems like something a user could potentially do and not realize, then not be able to promote the backup to DC.

bstansberry: When we do the HA DC thing, then promoting manually a non-DCC to master will be an edge case and the main use for --backup would be for a slave.

If you'd like to warn though, that's ok. We could always remove that when we get the HA DC thing going.

bstansberry: Let's do that separately though as this is good to go other than a wording tweak here.

luck3y:
@bstansberry Tweaked, and I'll submit the warn separately.

PR #1706, opened 2016-07-27
Fix for WFCORE-1712

jfdenise:
Update the test to use the right IP address on the client and the server side.


bstansberry:
Please use WFCORE JIRAs in commit messages.  We clone JBEAPs to WFCORE, although it would be much better if the creator of the JBEAP did that for us. JBEAP does not align with WildFly Core versions so using it we can't track what was fixed when. It's only real purpose when it comes to WildFly Core minor releases is for QE to track their verification work on the EAP side.

PR #1710, opened 2016-07-27
WFCORE-1592, fix persisting removed entries in PropertiesFileLoader

luck3y:


domain-management/src/test/java/org/jboss/as/domain/management/security/PropertiesFileLoaderTestCase.java
```diff
@@ -0,0 +1,180 @@
...
+            Properties props = loader.getProperties();
+            props.put("NEW", "VALUE");
+            loader.persistProperties();
+            loader.stop(null);
+
+            // reload the file and make sure everything is there
+            PropertiesFileLoader loader2 = new PropertiesFileLoader(tmpFile.getAbsolutePath(), null);
+            loader2.start(null);
+            Properties props2 = loader2.getProperties();
+            verifyProperties(props2, props2.size()-1, "NEW", "VALUE");
```
bstansberry: This call doesn't validate that NEW is present with a value of VALUE, since that's not an entry in the  "properties" array. You could pass null for both params and get the same effect.

Kind of a nit; it just confused me a bit.

luck3y: That's not a nit, that's me being an idiot! I refactored that into a shared method at the end to remove some duplicate code ;)

I'll send up a fix.

bstansberry:
@luck3y looks good. Can you squash this one and then up a variant against 2.x? This sounds like a serious enough one to justify getting it sorted in 10.1.

luck3y:
Sure, will send that shortly.

PR #1711, opened 2016-07-28
WFCORE-1665 Improve usability on default value for gc logging

ctomc:
Mostly improve usabilty for enabling/disabling GC_LOG so you can also configure it in .conf/.conf.ps1/.conf.bat config file.

With this change changing default for gc logging is one simpler, so downstream project doesn't need to do lots of changes to whole script.

### DIFF
--- a/core-feature-pack/src/main/resources/content/bin/common.ps1
+++ b/core-feature-pack/src/main/resources/content/bin/common.ps1
@@ -34,7 +34,7 @@ $global:SECMGR = Get-Env-Boolean SECMGR $false
 $global:DEBUG_MODE=Get-Env DEBUG $false
 $global:DEBUG_PORT=Get-Env DEBUG_PORT 8787
 $global:RUN_IN_BACKGROUND=$false
-$global:GC_LOG=Get-Env GC_LOG $false
+$GC_LOG=Get-Env GC_LOG
 #module opts that are passed to jboss modules
 $global:MODULE_OPTS = @()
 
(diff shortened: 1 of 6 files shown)


core-feature-pack/src/main/resources/content/bin/common.ps1
```diff
@@ -34,7 +34,7 @@ $global:SECMGR = Get-Env-Boolean SECMGR $false
 $global:DEBUG_MODE=Get-Env DEBUG $false
 $global:DEBUG_PORT=Get-Env DEBUG_PORT 8787
 $global:RUN_IN_BACKGROUND=$false
-$global:GC_LOG=Get-Env GC_LOG $false
+$GC_LOG=Get-Env GC_LOG $false
```
ctomc: @psakar for downstream, only change needed would be to replace this line with
`$GC_LOG=Get-Env GC_LOG $false`

core-feature-pack/src/main/resources/content/bin/standalone.bat
```diff
@@ -12,21 +12,16 @@ rem         standalone.bat --debug 9797
 
 rem By default debug mode is disable.
 set DEBUG_MODE=false
-set DEBUG_PORT_VAR=8787
-set GC_LOG_VAR=false
+set DEBUG_PORT=8787
+set GC_LOG=false
```
ctomc: @psakar for downstream, `set GC_LOG=false`

PR #1717, opened 2016-08-01
Fix for WFCORE-1680

jfdenise:
Cursor is now used to allow for completion in the middle of a line.  Some completers were passing 0 as a cursor value, this has now to be replaced by the buffer size.
New unit tests added. CliCompletionTestCase (that was not run in the TestSuite) has been also updated. It has been added to the TestSuite.


bstansberry:
@aloubyansky please approve

aloubyansky:
Looks good, thanks.

PR #1721, opened 2016-08-02
Deploy all

jfdenise:
Deploying all inactive deployments. This is an attempt to run all tests on all platforms with latest Aesh 0.66.9 branch.


bstansberry:
I put a fixme on this as it uses a snapshot. Hopefully when the PR is updated the label will automatically be removed.

PR #1727, opened 2016-08-03
[WFCORE-1692] Upgrade to JBoss Modules 1.6.0.Beta1

darranl:

### DIFF
--- a/controller/src/main/java/org/jboss/as/controller/ModelControllerImpl.java
+++ b/controller/src/main/java/org/jboss/as/controller/ModelControllerImpl.java
@@ -100,7 +100,6 @@ import org.jboss.msc.service.ServiceRegistry;
 import org.jboss.msc.service.ServiceTarget;
 import org.jboss.threads.AsyncFuture;
 import org.jboss.threads.AsyncFutureTask;
-import org.wildfly.security.manager.WildFlySecurityManager;
 import org.wildfly.security.manager.action.GetAccessControlContextAction;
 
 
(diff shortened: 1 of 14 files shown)


bstansberry:
I put the fixme label on this because the "Full integration - Windows" failure set is recurring.

darranl:
That is going to need a bit more debugging, I thought we would already have the fix in our topic branch but I don't think David would have run the testsuite on Windows so looks like something new ;-(

bstansberry:
Perhaps something to do with this?

```
   <resources>
    	<resource-root path="."/>
        <!-- Insert resources here -->
    </resources>
```

The first failing test creates a module with that in its module.xml then drops archives in the module. So maybe the path="." is problematic on Windows?

Just a guess.

bstansberry:
@ctomc once you tweak the transformer API PR ;) can you have a look at this? Unless @darranl is already looking. There aren't that many changes in 1.6.0.Beta1 and one relates to the filesystem loader. It wouldn't surprise me if it breaks <resource-root path="."/>. I'm not certain <resource-root path="."/> is why the test fails but it sure smells right. We don't have much use of that construct in our testsuite and it seems like the kind of thing where an OS-specific bug could pop up.

```
$ git grep "path=\"\.\""
feature-pack/src/main/resources/modules/system/layers/base/org/jboss/genericjms/main/module.xml:       <resource-root path="."/>
testsuite/integration/basic/src/test/java/org/jboss/as/test/integration/jca/moduledeployment/module-jar.xml:      <resource-root path="."/>
testsuite/integration/basic/src/test/java/org/jboss/as/test/integration/jca/moduledeployment/module.xml:          <resource-root path="."/>
testsuite/integration/basic/src/test/java/org/jboss/as/test/integration/jca/moduledeployment/module1-jar.xml:     <resource-root path="."/>
testsuite/integration/basic/src/test/java/org/jboss/as/test/integration/jca/moduledeployment/module1.xml:         <resource-root path="."/>
```

Same command finds nothing in core.

bstansberry:
Never mind Tomaz, I forgot you're off duty!

darranl:
@dmlloyd FYI I haven't had a chance to dig into this one but I tried to split out the modules update from our incubator branches so it could get in there earlier but it quickly ran into a failure on Windows.

dmlloyd:
Looking into this.  Whatever it is, it's definitely not obvious.  It might be related to the new implementation of FileResourceLoader; that's the only consequential change I can see that might impact this somehow.  @bsideup might have some insight as he originally implemented the new resource loader.  In the meantime I'll continue to examine the logs and hope something pops out.

bsideup:
@dmlloyd I'll take a look as well

bsideup:
@bstansberry any chance to localize it?

bstansberry:
@bsideup I don't understand.

dmlloyd:
The resource path is actually assembled in the ModuleXmlParser.  The code looks like this:
```
                final File file;
                final File loaderFile = new File(loaderPath);
                if (loaderFile.isAbsolute()) {
                    file = loaderFile;
                } else {
                    file = new File(rootPath, loaderPath);
                }
                if (file.isDirectory()) {
                    return ResourceLoaders.createFileResourceLoader(loaderName, file);
                } else {
                    final JarFile jarFile = new JarFile(file, true);
                    return ResourceLoaders.createJarResourceLoader(loaderName, jarFile);
                }
```

The first branch is not taken in the "." case because the path is not absolute (hopefully).  Thus we get, in effect: ``file = new File(rootPath, ".");``.  The next test, ``if (file.isDirectory()) {`` should be true, resulting in the file resource loader being constructed.  The root ``File`` is translated into a ``Path`` in the constructor.  From that point everything should work per normal.  The only reason I can think of that this might fail is that for some reason, one of the above assumptions does not hold up on Windows.

dmlloyd:
I've resolved the problem.  I'll tag a Beta2 and send a PR to the branch of this PR.

dmlloyd:
The windows build failed due to OOME.

dmlloyd:
Retest this please

dmlloyd:
Retest this please

dmlloyd:
Yay

PR #1731, opened 2016-08-10
[WFCORE-1686] distinguish abort reason in order to log message properly.

soul2zimate:
please see https://issues.jboss.org/browse/WFCORE-1686
As suggested in Jira above, this differentiates between "real" abort and "harmless" abort. In case of  parameter -v and --help, it does not show any fatal message. For other incomplete server environment cases, it adds the message into log.

### DIFF
--- a/host-controller/src/main/java/org/jboss/as/host/controller/HostControllerEnvironmentWrapper.java
+++ b/host-controller/src/main/java/org/jboss/as/host/controller/HostControllerEnvironmentWrapper.java
@@ -1,0 +1,53 @@
+/*
+ * Copyright 2016 Red Hat, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *   http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.jboss.as.host.controller;
+
+/**
+ * @author wangc
+ *
+ */
+public class HostControllerEnvironmentWrapper {
+
+    enum HostControllerEnvironmentStatus {
+        NORMAL, // expected abort
+        ERROR // problematic abort
+    }
+
+    private HostControllerEnvironment hostControllerEnvironment;
+    private HostControllerEnvironmentStatus hostControllerEnvironmentStatus;
+
+    private HostControllerEnvironmentWrapper(HostControllerEnvironment hostControllerEnvironment, HostControllerEnvironmentStatus hostControllerEnvironmentStatus) {
+        this.hostControllerEnvironment = hostControllerEnvironment;
+        this.hostControllerEnvironmentStatus = hostControllerEnvironmentStatus;
+    }
+
+    protected HostControllerEnvironmentWrapper(HostControllerEnvironment hostControllerEnvironment) {
+        this(hostControllerEnvironment, null);
+    }
+
+    protected HostControllerEnvironmentWrapper(HostControllerEnvironmentStatus hostControllerEnvironmentStatus) {
+        this(null, hostControllerEnvironmentStatus);
+    }
+
+    public HostControllerEnvironment getHostControllerEnvironment() {
+        return hostControllerEnvironment;
+    }
+
+    public HostControllerEnvironmentStatus getHostControllerEnvironmentStatus() {
+        return hostControllerEnvironmentStatus;
+    }
+}
--- a/host-controller/src/main/java/org/jboss/as/host/controller/Main.java
+++ b/host-controller/src/main/java/org/jboss/as/host/controller/Main.java
@@ -48,6 +48,7 @@ import org.jboss.as.process.ExitCodes;
 import org.jboss.as.process.ProcessController;
 import org.jboss.as.process.protocol.StreamUtils;
 import org.jboss.as.process.stdin.Base64InputStream;
+import org.jboss.as.server.SystemExiter;
 import org.jboss.as.version.ProductConfig;
 import org.jboss.logging.MDC;
 import org.jboss.logmanager.Level;
@@ -130,14 +131,18 @@ public final class Main {
         try {
             // TODO make this settable via an embedding process
             final long startTime = Module.getStartTime();
-            final HostControllerEnvironment config = determineEnvironment(args, startTime);
-            if (config == null) {
+            final HostControllerEnvironmentWrapper hostControllerEnvironmentWrapper = determineEnvironment(args, startTime);
+            if (hostControllerEnvironmentWrapper.getHostControllerEnvironment() == null) {
                 usage(); // In case there was an error determining the environment print the usage
-                abort();
+                if (hostControllerEnvironmentWrapper.getHostControllerEnvironmentStatus() == HostControllerEnvironmentWrapper.HostControllerEnvironmentStatus.ERROR) {
+                    abort();
+                } else {
+                    SystemExiter.safeAbort();
+                }
                 return null;
             } else {
                 try {
-                    final HostControllerBootstrap hc = new HostControllerBootstrap(config, authCode);
+                    final HostControllerBootstrap hc = new HostControllerBootstrap(hostControllerEnvironmentWrapper.getHostControllerEnvironment(), authCode);
                     hc.bootstrap();
                     return hc;
                 } catch(Throwable t) {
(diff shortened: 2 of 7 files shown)


host-controller/src/main/java/org/jboss/as/host/controller/HostControllerEnvironmentWrapper.java
```diff
@@ -0,0 +1,63 @@
...
+
+package org.jboss.as.host.controller;
+
+import java.io.Serializable;
+
+/**
+ * @author wangc
+ *
+ */
+public class HostControllerEnvironmentWrapper implements Serializable {
```
bstansberry: This shouldn't need to be Serializable.

host-controller/src/main/java/org/jboss/as/host/controller/HostControllerEnvironmentWrapper.java
```diff
@@ -0,0 +1,63 @@
...
+
+    private HostControllerEnvironment hostControllerEnvironment;
+    private HostControllerEnvironmentStatus hostControllerEnvironmentStatus;
+
+    private HostControllerEnvironmentWrapper(HostControllerEnvironment hostControllerEnvironment, HostControllerEnvironmentStatus hostControllerEnvironmentStatus) {
+        this.hostControllerEnvironment = hostControllerEnvironment;
+        this.hostControllerEnvironmentStatus = hostControllerEnvironmentStatus;
+    }
+
+    public HostControllerEnvironmentWrapper(HostControllerEnvironment hostControllerEnvironment) {
```
bstansberry: Constructor can be package-protected.

host-controller/src/main/java/org/jboss/as/host/controller/HostControllerEnvironmentWrapper.java
```diff
@@ -0,0 +1,63 @@
...
+        this(hostControllerEnvironment, null);
+    }
+
+    public HostControllerEnvironmentWrapper(HostControllerEnvironmentStatus hostControllerEnvironmentStatus) {
```
bstansberry: Constructor can be package-protected.

server/src/main/java/org/jboss/as/server/ServerEnvironmentWrapper.java
```diff
@@ -0,0 +1,63 @@
...
+
+    private ServerEnvironment serverEnvironment;
+    private ServerEnvironmentStatus serverEnvironmentStatus;
+
+    private ServerEnvironmentWrapper(ServerEnvironment serverEnvironment, ServerEnvironmentStatus serverEnvironmentStatus) {
+        this.serverEnvironment = serverEnvironment;
+        this.serverEnvironmentStatus = serverEnvironmentStatus;
+    }
+
+    public ServerEnvironmentWrapper(ServerEnvironment serverEnvironment) {
```
bstansberry: Constructor can be package-protected.

server/src/main/java/org/jboss/as/server/ServerEnvironmentWrapper.java
```diff
@@ -0,0 +1,63 @@
...
+        this(serverEnvironment, null);
+    }
+
+    public ServerEnvironmentWrapper(ServerEnvironmentStatus serverEnvironmentStatus) {
```
bstansberry: Constructor can be package-protected.

server/src/main/java/org/jboss/as/server/ServerEnvironmentWrapper.java
```diff
@@ -0,0 +1,63 @@
...
+
+package org.jboss.as.server;
+
+import java.io.Serializable;
+
+/**
+ * @author wangc
+ *
+ */
+public class ServerEnvironmentWrapper implements Serializable {
```
bstansberry: This shouldn't need to be Serializable.

server/src/main/java/org/jboss/as/server/ServerEnvironmentWrapper.java
```diff
@@ -0,0 +1,63 @@
...
+ * This software is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the GNU
+ * Lesser General Public License for more details.
+ *
+ * You should have received a copy of the GNU Lesser General Public
+ * License along with this software; if not, write to the Free
+ * Software Foundation, Inc., 51 Franklin St, Fifth Floor, Boston, MA
+ * 02110-1301 USA, or see the FSF site: http://www.fsf.org.
+ */
```
bstansberry: We use the ASL copyright header for all new code.

host-controller/src/main/java/org/jboss/as/host/controller/HostControllerEnvironmentWrapper.java
```diff
@@ -0,0 +1,63 @@
...
+ * This software is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the GNU
+ * Lesser General Public License for more details.
+ *
+ * You should have received a copy of the GNU Lesser General Public
+ * License along with this software; if not, write to the Free
+ * Software Foundation, Inc., 51 Franklin St, Fifth Floor, Boston, MA
+ * 02110-1301 USA, or see the FSF site: http://www.fsf.org.
+ */
```
bstansberry: We use the ASL copyright header for all new code.

host-controller/src/main/java/org/jboss/as/host/controller/Main.java
```diff
@@ -130,14 +130,16 @@ private HostControllerBootstrap boot(String[] args, final String authCode) {
...
             final long startTime = Module.getStartTime();
-            final HostControllerEnvironment config = determineEnvironment(args, startTime);
-            if (config == null) {
+            final HostControllerEnvironmentWrapper hostControllerEnvironmentWrapper = determineEnvironment(args, startTime);
+            if (hostControllerEnvironmentWrapper.getHostControllerEnvironment() == null) {
                 usage(); // In case there was an error determining the environment print the usage
-                abort();
+                if (hostControllerEnvironmentWrapper.getHostControllerEnvironmentStatus() == HostControllerEnvironmentWrapper.HostControllerEnvironmentStatus.ERROR) {
+                    abort();
+                }
```
bstansberry: We need an 'else' clause here where we still end up calling SystemExiter.logAndExit(), just with an ExitLogger that's a no-op. Might as well create one as a constant in the SystemExiter class.

server/src/main/java/org/jboss/as/server/Main.java
```diff
@@ -91,14 +91,16 @@ public static void main(String[] args) {
...
-            ServerEnvironment serverEnvironment = determineEnvironment(args, WildFlySecurityManager.getSystemPropertiesPrivileged(),
+            ServerEnvironmentWrapper serverEnvironmentWrapper = determineEnvironment(args, WildFlySecurityManager.getSystemPropertiesPrivileged(),
                     WildFlySecurityManager.getSystemEnvironmentPrivileged(), ServerEnvironment.LaunchType.STANDALONE,
                     Module.getStartTime());
-            if (serverEnvironment == null) {
-                abort(null);
+            if (serverEnvironmentWrapper.getServerEnvironment() == null) {
+                if (serverEnvironmentWrapper.getServerEnvironmentStatus() == ServerEnvironmentWrapper.ServerEnvironmentStatus.ERROR) {
+                    abort(null);
+                }
```
bstansberry: We need an 'else' clause here where we still end up calling SystemExiter.logAndExit(), just with an ExitLogger that's a no-op.

bstansberry:
Thanks for this!

soul2zimate:
Thanks for review. I use the same template from [wildfly-templates.xml](https://github.com/wildfly/wildfly-core/blob/master/ide-configs/eclipse/wildfly-templates.xml) If the new updated one is correct, I will send a PR to update the provided old template.

bstansberry:
I'd appreciate a fix to the template; it's out of date. Thanks!

host-controller/src/main/java/org/jboss/as/host/controller/HostControllerEnvironmentWrapper.java
```diff
@@ -0,0 +1,53 @@
...
+    private HostControllerEnvironmentWrapper(HostControllerEnvironment hostControllerEnvironment, HostControllerEnvironmentStatus hostControllerEnvironmentStatus) {
+        this.hostControllerEnvironment = hostControllerEnvironment;
+        this.hostControllerEnvironmentStatus = hostControllerEnvironmentStatus;
+    }
+
+    protected HostControllerEnvironmentWrapper(HostControllerEnvironment hostControllerEnvironment) {
+        this(hostControllerEnvironment, null);
+    }
+
+    protected HostControllerEnvironmentWrapper(HostControllerEnvironmentStatus hostControllerEnvironmentStatus) {
```
bstansberry: These constructors can be package-protected. There is no subclassing allowed. The classes can even be final.

Sorry to be a pain about this but I kind of have a mania about visibility. It's easier to maintain stuff when visibility is restricted as much as possible; no one then has to guess whether broader visibility was intentional.

bstansberry:
I merged this because the nit about package-protected vs protected isn't worth not getting this in Alpha5. Thanks!

PR #1737, opened 2016-08-11
better Controlling Access and ASL copyright

soul2zimate:
follow up  https://github.com/wildfly/wildfly-core/pull/1731


bstansberry:
:)

PR #1740, opened 2016-08-15
update License in README

soul2zimate:
follow up [wildfly-templates.xml](https://github.com/wildfly/wildfly-core/blob/master/ide-configs/eclipse/wildfly-templates.xml) update License to ASL.

### DIFF
--- a/README.md
+++ b/README.md
@@ -62,5 +62,5 @@ Using Eclipse
 
 License
 -------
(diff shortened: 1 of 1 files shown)


bstansberry:
We haven't officially relicensed the project. We are just using ASL for all new code.

soul2zimate:
close this for now

PR #1743, opened 2016-08-16
[WFCORE-1429] Allow the log manager check to skipped

jamezp:
The log manager may be wrapped which shouldn't stop the subsystem from working as long as it's a JUL based log manager.

This is a fix for https://issues.jboss.org/browse/WFCORE-1429. I've tried to think of the consequences of wrapping the log manager. The only possible issues I could come up with MDC and NDC. However the jboss-logging binding provider does seem to be the JBoss Log Manager provider. This should work as should the logging subsystem configuration. Since the configuration doesn't directly use the `org.jboss.logmanager.LogManager` directly. 

One cause for concern would be boot strap logging might not be configured correctly. This is likely okay though as care was taken to get around using the correct log manager and allowing WildFly to boot with the logging subsystem and not check the log manager.

### DIFF
--- a/logging/src/main/java/org/jboss/as/logging/logging/LoggingLogger.java
+++ b/logging/src/main/java/org/jboss/as/logging/logging/LoggingLogger.java
@@ -926,4 +926,15 @@ public interface LoggingLogger extends BasicLogger {
     @LogMessage(level = ERROR)
     @Message(id = 88, value = "Could not determine %s had any children resources.")
     void errorDeterminingChildrenExist(@Cause Throwable cause, String childType);
+
+    /**
+     * Logs a warning message indicating the log manager does not appear to the {@link org.jboss.logmanager.LogManager}.
+     *
+     * @param logManagerName the log manager system property value
+     */
+    @LogMessage(level = WARN)
+    @Message(id = 89, value = "The log manager check was skipped and the log manager system property, " +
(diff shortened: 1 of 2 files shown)


logging/src/main/java/org/jboss/as/logging/logging/LoggingLogger.java
```diff
@@ -926,4 +926,15 @@
     @LogMessage(level = ERROR)
     @Message(id = 88, value = "Could not determine %s had any children resources.")
     void errorDeterminingChildrenExist(@Cause Throwable cause, String childType);
+
+    /**
+     * Logs a warning message indicating the log manager does not appear to the {@link org.jboss.logmanager.LogManager}.
+     *
+     * @param logManagerName the log manager system property value
+     */
+    @LogMessage(level = WARN)
```
bstansberry: Is this a WARN, not an INFO? The user set the property, which is unlikely to happen by mistake or without prior research, so they are presumably somewhat knowledgable.

In the past I had the mentality of "assume a WARN or ERROR log message means some poor sysadmin might get paged." I know that we aren't so restrictive everywhere though.

bstansberry: I've changed my mind; I like WARN. Ping me if you are curious why. ;)

PR #1744, opened 2016-08-17
[WFCORE-1714] command completion in case no UnresolvedVariableExcepti…

soul2zimate:
…on is caught.
please see https://issues.jboss.org/browse/WFCORE-1714

### DIFF
--- a/cli/src/main/java/org/jboss/as/cli/CommandCompleter.java
+++ b/cli/src/main/java/org/jboss/as/cli/CommandCompleter.java
@@ -24,6 +24,7 @@ package org.jboss.as.cli;
 
 import java.util.Collections;
 import java.util.List;
+import java.util.stream.Collectors;
 
 import org.jboss.as.cli.impl.CommandCandidatesProvider;
 import org.jboss.as.cli.operation.OperationCandidatesProvider;
@@ -82,6 +83,7 @@ public class CommandCompleter implements CommandLineCompleter {
         }
 
         final DefaultCallbackHandler parsedCmd = (DefaultCallbackHandler) ctx.getParsedCommandLine();
+        boolean unresolvedVariableException = false;
         try {
             // WFCORE-1627 parse current position if cursor moves
             if (ctx.getArgumentsString() == null && buffer.length() > cursor) {
@@ -93,6 +95,7 @@ public class CommandCompleter implements CommandLineCompleter {
                 parsedCmd.parse(ctx.getCurrentNodePath(), buffer, false, ctx);
             }
         } catch(UnresolvedVariableException e) {
+            unresolvedVariableException = true;
             final String variable = e.getExpression();
             if(buffer.endsWith(variable)) {
                 for(String var : ctx.getVariables()) {
@@ -139,6 +142,25 @@ public class CommandCompleter implements CommandLineCompleter {
                 candidates.remove(notIndex);
             }
         }
+
+        // https://issues.jboss.org/browse/WFCORE-1714
+        if (!unresolvedVariableException) {
+            int index = buffer.lastIndexOf('$');
+            if (index != -1 && buffer.length() != index + 1) {
+                String variable = buffer.substring(index + 1);
+                if (buffer.endsWith(variable)) {
+                    List<String> candidateVariables = ctx.getVariables().stream()
+                            .filter(e -> e.startsWith(variable) && !e.equals(variable)).collect(Collectors.toList());
+                    if (!candidateVariables.isEmpty()) {
+                        candidates.addAll(candidateVariables);
+                        Collections.sort(candidates);
+                    }
+                } else {
+                    // failed to resolve a variable in the middle of the line
+                }
+            }
+        }
+
         // if there is nothing else to suggest, check whether it could be a start of a variable
         if(candidates.isEmpty() && buffer.charAt(buffer.length() - 1) == '$' && !ctx.getVariables().isEmpty()) {
             candidates.addAll(ctx.getVariables());
--- a/testsuite/standalone/src/test/java/org/jboss/as/test/integration/management/cli/VariablesCompletionTestCase.java
+++ b/testsuite/standalone/src/test/java/org/jboss/as/test/integration/management/cli/VariablesCompletionTestCase.java
@@ -1,0 +1,122 @@
+/*
+ * Copyright 2016 Red Hat, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *   http://www.apache.org/licenses/LICENSE-2.0
+ *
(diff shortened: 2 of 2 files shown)


bstansberry:
@jfdenise please approve

jfdenise:
Hi,
thank you for your fix!
I have done a quick experiment with it and it appears that the completion is no more working as expected. This does show that we need new tests for this fix and variable completion in general ;-)
It would be nice if you could take the opportunity to cover the variables completion in this PR.

For example:
set foo=/subsystem=logging
$foo<TAB> ==> nothing happens, used to propose :, /

soul2zimate:
retest this please

soul2zimate:
rebased to add more tests on variable completion.
testCompleteFirstVariableCompletion covers the reported scenario.

cli/src/main/java/org/jboss/as/cli/CommandCompleter.java
```diff
@@ -124,6 +127,31 @@ protected int doComplete(CommandContext ctx, String buffer, int cursor, List<Str
...
+                if (buffer.endsWith(variable)) {
+                    List<String> candidateVariables = new ArrayList<>();
+                    for (String var : ctx.getVariables()) {
+                        if (var.startsWith(variable)) {
+                            candidateVariables.add(var);
+                        }
+                    }
+                    if (candidateVariables.size() > 1) {
+                        candidates.addAll(candidateVariables); // more than one candidate variables
+                        Collections.sort(candidates);
```
jfdenise: I think that you shouldn't add candidates prior to properly complete the request. This is why you added the special case in OperationRequestCompleter.
You should add candidates after having called the completer.

cli/src/main/java/org/jboss/as/cli/operation/OperationRequestCompleter.java
```diff
@@ -487,6 +487,14 @@ protected int complete(CommandContext ctx, ParsedCommandLine parsedCmd, Operatio
                 candidates.set(0, Util.escapeString(candidate, ESCAPE_SELECTOR) + '=');
             }
         } else {
+            if (address.endsOnType()) {
```
jfdenise: This code is induced by the var being in candidates when completing. You shouldn't need this change. This is the kind of change that, when we read the code, we don't understand why it is there. Difficult to maintain when a futur bug will be chased.

testsuite/standalone/src/test/java/org/jboss/as/test/integration/management/cli/VariablesCompletionTestCase.java
```diff
@@ -0,0 +1,146 @@
...
+        ctx.handle("set " + FOO_NAME + "=" + FOO_VALUE);
+        ctx.handle("set " + FOOBAR_NAME + "=" + FOOBAR_VALUE);
+        cliOut.reset();
+
+        ctx.handle("echo $" + FOO_NAME);
+        assertTrue(cliOut.toString().contains(FOO_VALUE));
+        ctx.handle("echo $" + FOOBAR_NAME);
+        assertTrue(cliOut.toString().contains(FOOBAR_VALUE));
+
+        ctx.handle("connect"); // This is necessary to get : and /
```
jfdenise: You should use the class CLIWrapper and call the constructor CLIWrapper(true). It will do all setup for you ;-)

cli/src/main/java/org/jboss/as/cli/CommandCompleter.java
soul2zimate: Yes, I have moved this down now, And tests are using proposed CLIWrapper

cli/src/main/java/org/jboss/as/cli/CommandCompleter.java
```diff
@@ -139,6 +142,31 @@ protected int doComplete(CommandContext ctx, String buffer, int cursor, List<Str
                 candidates.remove(notIndex);
             }
         }
+
+        // https://issues.jboss.org/browse/WFCORE-1714
+        if (unresolvedVariableException == false) {
```
jfdenise: ! unresolvedVariableException could be used.

cli/src/main/java/org/jboss/as/cli/CommandCompleter.java
```diff
@@ -139,6 +142,31 @@ protected int doComplete(CommandContext ctx, String buffer, int cursor, List<Str
...
+            int index = buffer.lastIndexOf('$');
+            if (index != -1) {
+                String variable = buffer.substring(index + 1);
```
jfdenise: Shouldn't you check that index is not the last char of the buffer?

cli/src/main/java/org/jboss/as/cli/CommandCompleter.java
```diff
@@ -139,6 +142,31 @@ protected int doComplete(CommandContext ctx, String buffer, int cursor, List<Str
...
+                if (buffer.endsWith(variable)) {
+                    List<String> candidateVariables = new ArrayList<>();
+                    for (String var : ctx.getVariables()) {
+                        if (var.startsWith(variable)) {
+                            candidateVariables.add(var);
+                        }
+                    }
+                    if (candidateVariables.size() > 1) {
```
jfdenise: You could filter out the existing variable first, then simply check if the list of variables isEmpty? It would seem more natural to me.

cli/src/main/java/org/jboss/as/cli/CommandCompleter.java
```diff
@@ -139,6 +142,31 @@ protected int doComplete(CommandContext ctx, String buffer, int cursor, List<Str
...
+                        candidates.addAll(candidateVariables); // more than one candidate variables
+                        Collections.sort(candidates);
+                        if (!candidateVariables.stream().anyMatch(e -> e.equals(variable)))
+                            return buffer.length() - variable.length();
+                    }
+                } else {
+                    // failed to resolve a variable in the middle of the line
```
jfdenise: Just a suggestion, perhaps you could revert if and else to avoid a negative if test.

jfdenise:
Hi @soul2zimate, thank you for updating the fix. This looks much better now. I have made some small remarks on the code.

cli/src/main/java/org/jboss/as/cli/CommandCompleter.java
soul2zimate: I have updated to use the stream filter here, but make it greater than 1 inside the if. 
If there is only one candidate matches without seeing unresolvedVariableException here, it means variable matches whole variable name, and it does not need to add it in the candidate list.

soul2zimate:
retest this please

bstansberry:
@jfdenise please @ mention me when this is good to go. I'll put a hold label on it for now to stop me checking on it until then.

jfdenise:
Hi @soul2zimate the code looks good to me. Thanks!
@bstansberry the PR is good to merge.
JF

PR #1745, opened 2016-08-17
Demonstrate a more realistic simple way of creating a JMX deployment …

bstansberry:
…invocation

This is a not particularly important test change; it's more the result of my proving to myself that the "content" attribute in a deployment operation can be built up in JMX in a relatively simple manner, one that does not require specifying all of the fields in the management model's 'content' param. Simply specifying the relevant fields (just 'hash' in this case) is enough to make a successful invocation.

This all relates to my question of whether a new 'empty' configuration related to exploded managed deployments should be set via a new param in the operation or whether it should be a new field in the existing 'content' param. This test change demonstrates that if it were a new field, existing JMX code could simply ignore it and continue to work without modification. OTOH adding a new param means existing JMX code breaks.


bstansberry:
retest this please

PR #1748, opened 2016-08-18
WFCORE-1723 - OperationRequestCompleter checks if CommandArgument's V…

spyrkob:
…alueCompleter can be applied before choosing it

JIRA: https://issues.jboss.org/browse/WFCORE-1723

### DIFF
--- a/cli/src/main/java/org/jboss/as/cli/impl/ArgumentWithoutValue.java
+++ b/cli/src/main/java/org/jboss/as/cli/impl/ArgumentWithoutValue.java
@@ -206,7 +206,12 @@ public class ArgumentWithoutValue implements CommandArgument {
                 if(index == -1) {
                     return false;
                 }
-                return !(index == 0 && values.size() == 1);
+                if(index == 0 && values.size() == 1) {
+                    // isValueComplete check is used by overriding classes (like ArgumentWithValue)
+                    return !isValueComplete(args);
+                } else {
+                    return true;
+                }
             }
 
             if(propertyNames.size() != 1) {
@@ -230,8 +235,8 @@ public class ArgumentWithoutValue implements CommandArgument {
             return fullName.startsWith(args.getLastParsedPropertyName()) || (shortName != null && shortName.startsWith(args.getLastParsedPropertyName()));
         }
 
-        if (isPresent(args)) {
-            // An argument without value has no value
+        if (isPresent(args) && isValueComplete(args)) {
+            // isValueComplete check is used by overriding classes (like ArgumentWithValue)
             return false;
         }
 
--- a/cli/src/main/java/org/jboss/as/cli/operation/OperationRequestCompleter.java
+++ b/cli/src/main/java/org/jboss/as/cli/operation/OperationRequestCompleter.java
@@ -167,10 +167,12 @@ public class OperationRequestCompleter implements CommandLineCompleter {
                     boolean needNeg = false;
                     for (CommandArgument arg : allArgs) {
                         if (arg.canAppearNext(ctx)) {
-                            if (arg.getIndex() >= 0) {
-                                final CommandLineCompleter valCompl = arg.getValueCompleter();
-                                if (valCompl != null) {
-                                    valCompl.complete(ctx, "", 0, candidates);
+                            if (arg.getIndex() >= 0 ) {
+                                if (arg.getIndex() == parsedCmd.getOtherProperties().size() || arg.getIndex() == Integer.MAX_VALUE) {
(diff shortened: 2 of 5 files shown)


spyrkob:
@jfdenise could you please take a look at this? I came across this issue when working on https://issues.jboss.org/browse/WFCORE-1840

The actual fix was the change in ArgumentWithoutValue#canAppearNext and OperationRequestCompleter#getValueCompleter to handle a case when an indexed value is not completed, but I also had to make some changes to OperationRequestCompleter to preserve current indexed argument behaviour in other cases.

jfdenise:
Hi, 
you fix makes full sense. I would need some time to review this change. It is impacting some critical code in the CLI completion.  We could very easily introduce regressions.
JF

jfdenise:
My general feeling is that the fix seems very complex. I don't like too much math ;-).
The problem you are trying to fix seems to be:
- An Handler defines more than 1 Completer at index i. It means that there is some discriminents that are already present in what the user has typed (for example, the right action).
- When the decision is to be made of which completer should be called, then only 1 should be found and used.
If you agree with this, then I have the feeling that we could imagine a simpler fix (Simpler I mean that doesn't spread-out in multiple classes, and doesn't impact existing tests).

Look at OperationRequestCompleter.getValueCompleter. It does return the first found. Perhaps you could:
1) store all the args that would match in a list.
2) At the end of the loop, if you have more than one match, then call canAppearNext in a loop on the matching args. The first one that canAppearNext, just return it.
3) If you don't have multiple arg, return the one at index 0.
3) If match is empty, return the maxIndex (as done today).

I did a quick hack locally and it seems to work. I added a completer to patch-id (so 2 completer at index 0). When rollback is used this new completer is called. When apply is used, the FileSystem one is used. I Ran the tests, and all is green.

You could perhaps experiment with this approach? What do you think?

JF

spyrkob:
I think I started with similar approach, but run into an issue when trying to complete an unfinished value for an indexed argument, eg:
* `patch rollback <TAB>` would suggest correct completer, but
* `path rollback a<TAB>` would revert to the path completer (from patch apply command)

Fixing that led to all that nasty complexity :(

Another option I was thinking about was introducing a new base class (something like IndexedArgument) and adding special treatment for it in OperationRequestCompleter. This way arguments that need it could be moved over to the new base class without disturbing others

jfdenise:
Hi,
I didn't push the implementation as far as you did. We definitely needs more design here. This piece of code starts to be difficult to maintain and evolve. I would make things more explicit in the treatment of this case. We see it today as a special case but, if we evolve the current CLI implementation and syntax, this would become a very common case : command <action> <arguments and/or options>. So should be properly implemented.
As you suggest, designing a new class seems the right approach. It could be a first step to simplify the whole OperationRequestCompleter.complete method.

bstansberry:
I'm going to put the Hold label on this as it seems there's an ongoing discussion and I don't want to keep looking at the PR before it's ready. @jfdenise please ping me when this is good to go, or @spyrkob if this is unlikely to progress relatively soon please close it until ready and you and Jeff coordinate on it using your personal github repos.

jfdenise:
Having 2 unnamed argument at the same index, one being active according to the value of the previous argument is not supported. To support it, I think that it would impact deeply the current Argument/parsing/completion logic. @aloubyansky could comment on that.

I would prefer that we address it in the CLI.next. This is natively supported by the Aesh Command API. A Command can have sub commands (the action). Sub commands can then have an unnamed argument. It would address this patch use case.

aloubyansky:
We have to support it actually, i.e. to be able to differentiate between two arguments assigned to the same index but having mutually exclusive activating conditions. The current code relies on that, for example patch command mentioned in the issue. It has to remain supported. What we may say is not supported is the completion part if it's too difficult to implement and maintain.
I haven't looked into the very details of the patch but it does look complicated (which was probably dictated by the limitations of the current design). So, I will leave it to Jeff to decide whether to accept it, since he'll be the one to maintain the code.
Still, it's a good effort, @spyrkob, thanks for investigating the issue.

jfdenise:
Thanks @aloubyansky,
the first attempt has impacted a piece of code that we know can easily cause regression.
If this is a critical requirement for the customer, we need to address it properly. If that is not, I would delay it to the CLI.next.

spyrkob:
@jfdenise FYI I've been playing with an [alternative solution](https://github.com/wildfly/wildfly-core/compare/master...spyrkob:cli-indexed-argument). This limits the changes to existing functionality, but still operates on OperationRequestCompleter.

I'm gonna close this PR for now and reopen it if needed.

PR #1750, opened 2016-08-23
Ignore intermittently failing test

bstansberry:
@jfdenise @schvarc FYI. WFCORE-1729 is to fix the test or fix the bug that causes the failure.


bstansberry:
retest this please

PR #1751, opened 2016-08-23
WFCORE-316: --cached-dc additional fixes

luck3y:
- keep the remote file repository so it's available when the DC becomes available again
 - parse domain.cached-remote.xml to a temporary file until boot is successful to prevent any boot errors leaving it in a clobbered state
 - if any remote content is unavailable from an offline DC during boot, and not present in the local repository, abort the boot.

### DIFF
--- a/host-controller/src/main/java/org/jboss/as/domain/controller/resources/DomainRootDefinition.java
+++ b/host-controller/src/main/java/org/jboss/as/domain/controller/resources/DomainRootDefinition.java
@@ -334,7 +334,7 @@ public class DomainRootDefinition extends SimpleResourceDefinition {
         //TODO socket-binding-group currently lives in controller and the child RDs live in domain so they currently need passing in from here
         resourceRegistration.registerSubModel(SocketBindingGroupResourceDefinition.INSTANCE);
 
-        //TODO perhaps all these desriptions and the validator log messages should be moved into management-client-content?
+        //TODO perhaps all these descriptions and the validator log messages should be moved into management-client-content?
         resourceRegistration.registerSubModel(
                 new ManagedDMRContentTypeResourceDefinition(contentRepo, ROLLOUT_PLAN,
                 PathElement.pathElement(MANAGEMENT_CLIENT_CONTENT, ROLLOUT_PLANS), new RolloutPlanValidator(), DomainResolver.getResolver(ROLLOUT_PLANS), DomainResolver.getResolver(ROLLOUT_PLAN)));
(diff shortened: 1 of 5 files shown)


luck3y:
retest this please

luck3y:
retest this please

luck3y:
retest this please

host-controller/src/main/java/org/jboss/as/host/controller/HostControllerConfigurationPersister.java
```diff
@@ -82,8 +82,7 @@ public void initializeDomainConfigurationPersister(boolean slave) {
         ConfigurationFile domainConfigurationFile = null;
         if (slave) {
             // in either case of --backup and/or --cached-dc, we persist with the same persister
-            if (environment.isBackupDomainFiles() || environment.isUseCachedDc()) {
-                domainConfigurationFile = getBackupDomainConfigurationFile();
```
bstansberry: Why is this removed? AFAICT the only effect of this is null now gets passed to this below:

```
        // Store this back to environment so mgmt api that exposes it can still work
        environment.setDomainConfigurationFile(domainConfigurationFile);
```

Looking at the impact of that being null, it would only be to disable output from HostEnvironmentResourceDefinition.HostEnvironmentReadHandler. The other read of that value in HostProcessReloadHandler won't happen on a slave, so what we pass in here doesn't matter.

bstansberry:
@luck3y small question above aside, looks good. Ping @ctomc about the compiler failures on windows; we're seeing those on a lot of CI runs. I assume there's some environmental problem on the teamcity servers.

luck3y:
@bstansberry Hmm, that's odd I thought I'd checked what that was used for. Thanks for the catch :)

bstansberry:
@luck3y Please double check my analysis of the code for that point. I could have missed something.

luck3y:
@bstansberry, Removing that wasn't intentional on my part, you're right about:

`// Store this back to environment so mgmt api that exposes it can still work
        environment.setDomainConfigurationFile(domainConfigurationFile);`

Though. I'll retest this to be sure it still works the same.

luck3y:
@bstansberry Can you drop a fixme on this, and I'll ping you or push after I've restested?

luck3y:
retest this please

luck3y:
@bstansberry Windows CI should be fixed, and I've retested that change, and it's working fine, so when the tests pass, should be gtg.

luck3y:
retest this please

luck3y:
Retest this please

luck3y:
@bstansberry this should be all set now

PR #1753, opened 2016-08-25
WFCORE-1732 permissions of the parent should be applied to the subdep…

istudens:
…loyment too

Supersedes https://github.com/wildfly/wildfly/pull/8842.

### DIFF
--- a/server/src/main/java/org/jboss/as/server/deployment/module/ModuleSpecProcessor.java
+++ b/server/src/main/java/org/jboss/as/server/deployment/module/ModuleSpecProcessor.java
@@ -94,7 +94,7 @@ public class ModuleSpecProcessor implements DeploymentUnitProcessor {
     private void deployModuleSpec(final DeploymentPhaseContext phaseContext) throws DeploymentUnitProcessingException {
 
         final DeploymentUnit deploymentUnit = phaseContext.getDeploymentUnit();
-        final DeploymentUnit topLevelDeployment = deploymentUnit.getParent() == null ? deploymentUnit : deploymentUnit.getParent();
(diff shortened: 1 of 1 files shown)


istudens:
@stuartwdouglas @dmlloyd could you review this please? thanks

istudens:
This fix seems to be too general. I was afraid it would cause regressions. But at least VirtualFilePermissions might be cloned from the parent to subdeployments this way, wdyt?

istudens:
I mean something like 
https://github.com/istudens/wildfly-core/commit/d925cf45ce785eb1382bd0df833ce3b29a78c101#diff-3f1bdca0ad46a27814c89dfb03e146b1R257

dmlloyd:
See comments on downstream PR - do not merge yet.

dmlloyd:
I talked this over with Petr.  It's a +1 from me.

dmlloyd:
Assuming the tests pass. :)

istudens:
David, do you have any idea how to fix this in a different way? My proposed fix fixes this issue but introduces regressions because it is most likely not a good idea to clone perms from ear to war automatically. I am not sure if at least VirtualFilePermission might be cloned this way as I asked above, see the commit linked above. What do you think about it?
Otherwise any web deployment of an ear would need an explicit permissions to get to the lib packaged within the ear.

dmlloyd:
Rather than cloning permissions, we should be just setting the required permissions on each module separately.  It should be known what additional VFSPermission instances are required by just browsing the structure of the deployment.

istudens:
This has been refactored by @dmlloyd comments and it is finally passing the tests on TeamCity.

PR #1754, opened 2016-08-25
WFCORE-1733 SynchServerOperationHandler op and attribute restart-required checking

luck3y:


luck3y:
First go at this one. There are a couple of things I notice, that I'm not sure are correct.
- adding a system property changes the command line, so this overrides reload-required to restart-required. This is probably fine for anything else on the command line that isn't a system property though.
- for ops / attributes that are RESTART_NONE, should we still default to reload-required? This isn't really clear from the comments in AttributeAccess etc.
- My reading of the comments on flags suggests that only RESTART_JVM should trigger restart-required explicity, is that right?

luck3y:
Actually ignore the above wrt to sysprops, I just confused myself regarding them appearing in the management model vs actually taking effect in the jvm ;)

host-controller/src/main/java/org/jboss/as/domain/controller/operations/SyncServerStateOperationHandler.java
```diff
@@ -79,42 +88,55 @@ public void execute(OperationContext context, ModelNode operation) throws Operat
...
+                            opName = ServerProcessStateHandler.REQUIRE_RESTART_OPERATION;
+                            break;
+                        case RELOAD_REQUIRED:
+                            opName = ServerProcessStateHandler.REQUIRE_RELOAD_OPERATION;
+                            break;
+                        default:
+                            throw new RuntimeException("Unknown SyncServerResultAction: " + entry.getValue());
+                    }
+
+                    if (opName != null) {
```
bstansberry: opName can't be null

host-controller/src/main/java/org/jboss/as/domain/controller/operations/SyncServerStateOperationHandler.java
```diff
@@ -126,30 +148,51 @@ public void handleResult(OperationContext.ResultAction resultAction, OperationCo
...
-            PathAddress addr = PathAddress.pathAddress(operation.get(OP_ADDR));
+            final PathAddress addr = PathAddress.pathAddress(operation.get(OP_ADDR));
             contentDownloader.checkContent(operation, addr);
-
-            Map<Set<ServerIdentity>, ModelNode> serverMap = resolver.getServerOperations(context, operation, addr);
+            final ImmutableManagementResourceRegistration registration = context.getResourceRegistration();
+            final Map<Set<ServerIdentity>, ModelNode> serverMap = resolver.getServerOperations(context, operation, addr);
             for (Map.Entry<Set<ServerIdentity>, ModelNode> entry : serverMap.entrySet()) {
-                ModelNode op = entry.getValue();
-                boolean restart = op.get(OP).asString().equals(ServerProcessStateHandler.REQUIRE_RESTART_OPERATION);
```
bstansberry: This check still needs to be there. Then, if restart == true, the SyncServerResultAction must RESTART_REQUIRED and the stuff you are adding here should be skipped. Your new stuff is a second line of defense for cases where ServerOperationResolver doesn't have hard coded logic that knows the domain/host level op always requires server restart.

host-controller/src/main/java/org/jboss/as/domain/controller/operations/SyncServerStateOperationHandler.java
```diff
@@ -126,30 +148,51 @@ public void handleResult(OperationContext.ResultAction resultAction, OperationCo
...
+                } else { // all other ops
+                    final Set<OperationEntry.Flag> flags = registration.getOperationFlags(addr, opName);
+                    if (flags.contains(OperationEntry.Flag.RESTART_JVM)) {
+                        action = SyncServerResultAction.RESTART_REQUIRED;
                     }
                 }
+
+                for (final ServerIdentity id : entry.getKey()) {
+                    final String serverName = id.getServerName();
+                    serverStateChanges.put(serverName, serverStateChanges.get(serverName) == null ? SyncServerResultAction.RELOAD_REQUIRED : action);
```
bstansberry: You can't store RELOAD_REQUIRED if RESTART_REQUIRED is already stored under the server name key.

host-controller/src/main/java/org/jboss/as/domain/controller/operations/SyncServerStateOperationHandler.java
```diff
@@ -126,30 +148,51 @@ public void handleResult(OperationContext.ResultAction resultAction, OperationCo
...
             for (Map.Entry<Set<ServerIdentity>, ModelNode> entry : serverMap.entrySet()) {
-                ModelNode op = entry.getValue();
-                boolean restart = op.get(OP).asString().equals(ServerProcessStateHandler.REQUIRE_RESTART_OPERATION);
-                for (ServerIdentity id : entry.getKey()) {
-                    String serverName = id.getServerName();
-                    Boolean existing = serverStateChanges.get(serverName);
-                    if (existing == null || (!existing && restart)) {
-                        serverStateChanges.put(serverName, restart);
+                final ModelNode op = entry.getValue();
+                final String opName = op.get(OP).asString();
```
bstansberry: I don't think you can safely use this op for the registration.getAttributeAccess/registration.getOperationFlags check. It's the op that's expected to be executed on the server, which may not be the op that is registered here on the HC. Above you are using the address from the op that is being executed on the server. And rightly so as you have no idea if the server-op address is registered on the HC. For any /subsytem=* address without the preceding /profile=* level it won't be.

This is kind of a nit, since in practice it's likely going to just work, particularly since the "boolean restart = op.get(OP).asString().equals(ServerProcessStateHandler.REQUIRE_RESTART_OPERATION);" check that was at L139 will handle a lot of the odd cases. But it's the kind of thing that can might break some day in a hard to understand way.

Some possibilities:

1) Limit this second level check to addresses in the /profile=* tree where addr.size() > 1 (i.e. not the profile root address itself). For those address, ServerOperationResolver just takes the domain level op and strips the /profile=* part off the address, i.e. there won't be a change of the op name.

With the existing L139 "boolean restart = ..." check restored, this approach then just leaves a hole for domain/host ops outside the /profile=* tree where ServerOperationResolver isn't already returning ServerProcessStateHandler.REQUIRE_RESTART_OPERATION. Those cases really shouldn't exist. There are two reasons a reload is insufficient and restart is required:

1) The op changes something about how the HC would launch the server. The ServerOperationResolver logic handles these by returning  ServerProcessStateHandler.REQUIRE_RESTART_OPERATION.
2) Some lib we integrate stores data in a static field and that can't be changed without restart. We try really hard not to do this with kernel resources though, and if we do the ServerOperationResolver logic can handle it as well. So this case is really just a problem with extensions where the kernel has no idea what the extension will do.

2) If we decide 1) is wrong or too hacky, perhaps we can check for cases where the op name is different for 'op' versus 'operation'. In that case we know something unusual has happened and the lookup against the HC's registration is not a valid algorithm. So then just default to SyncServerResultAction.RESTART_REQUIRED.

host-controller/src/main/java/org/jboss/as/domain/controller/operations/SyncServerStateOperationHandler.java
bstansberry: Further to this, ServerOperationResolver can also return ServerProcessStateHandler.REQUIRE_RELOAD_OPERATION. In this case I *believe* it is fine to accept that result, use SyncServerResultAction.RELOAD_REQUIRED, and skip your new stuff. This needs a bit of confirmation, but I think ServerOperationResolver would not return ServerProcessStateHandler.REQUIRE_RELOAD_OPERATION for an op where actually a restart is needed. It's all custom logic meant for these kind of cases and when it does something specific like this we can just trust it.

host-controller/src/main/java/org/jboss/as/domain/controller/operations/SyncServerStateOperationHandler.java
luck3y: Yah, sorry about those, I'd been going back and forth trying to avoid really verbose checking, will make that clear & fix.

host-controller/src/main/java/org/jboss/as/domain/controller/operations/SyncServerStateOperationHandler.java
luck3y: Ah, this makes a lot more sense now. I couldn't ever see the ServerProcessStateHandler.REQUIRE_RESTART_OPERATION arriving, so I had suspected that it was just wrong.

That makes a lot more sense now!

host-controller/src/main/java/org/jboss/as/domain/controller/operations/SyncServerStateOperationHandler.java
luck3y: I'm fixing and trying #1 above, will see how hacky it turns out.

bstansberry: @luck3y I think 2) is fine if 1) is uncomfortable at all. For any op in profile=*/subystem=*, ServerOperationResolver is *not* going to change the operation name, so the 2) type check isn't going to affect anything for those ops. For other resources, the odds of ServerOperationResolver  changing the name but not returning REQUIRE_RESTART_OPERATION or REQUIRE_RELOAD_OPERATION are really low, and if it happens we'd just require restart which is hardly the end of the world if reload was sufficient.

luck3y:
@bstansberry I've reworked this to be more in line with your comments. Feedback welcome. It still feels a bit clunky, although this does handle the jts toggle correctly (/profile=full/subsystem=transactions:write-attribute(name=jts, value=false)

One additional concern I have is an odd one, but probably unrelated to this change (I'm still looking).

When I've been testing this I've basically been doing:

(a) Start DC
(b) Start slave with --cached-dc
(c) Stop slave.
(d) Make domain config change on DC in CLI.
(e) Stop DC
(f) Start slave (which starts with previous cached domain config from --cached-dc)
(g) Once slave is up, start DC and watch reconnection.

The odd part is that from a fresh build, there are a huge number of adds coming on in the first incidence of (g). Subsequent passes, it doesn't occur.

So I think there may be something triggering additional domain config elements to be sent over, but only on a slave reconnection.

Diff of domain.cached-remote.xml before and after the first occurance of (g) above.

https://paste.fedoraproject.org/423667/73286569/

bstansberry:
@luck3y What does the host.xml look like? My guess is this has something to do with --ignore-unused-configuration.  That diff shows the slave had "empty" profiles for full and full-ha (which itself is odd) and then at the end it pulled down the details of full-ha.

What was the change in d) ?

Side issue -- that diff shows the full-ha subsystems are getting added in reverse order from how they are on DC. Which is probably just a cosmetic thing but something to check out in case it's more and a minor thing to correct if it's just cosmetic.

luck3y:
@bstansberry Yah, ignore-unused-configuration was my initial guess too, I'm looking for somewhere that isn't being checked on reconnect.

host.xml default for ignore-unused--configuration, so I think this is probably a safe guess that something like this is going on.

luck3y:
Retest this please

PR #1757, opened 2016-08-29
[WFCORE-1741]:  read-content operation does not return the uuid of thstream as the operation result

ehsavoie:
Adding the stream uuid in the reponse node.

Jira: https://issues.jboss.org/browse/WFCORE-1741

### DIFF
--- a/host-controller/src/main/java/org/jboss/as/domain/controller/operations/deployment/ExplodedDeploymentReadContentHandler.java
+++ b/host-controller/src/main/java/org/jboss/as/domain/controller/operations/deployment/ExplodedDeploymentReadContentHandler.java
@@ -15,6 +15,7 @@
  */
 package org.jboss.as.domain.controller.operations.deployment;
 
+import static org.jboss.as.controller.descriptions.ModelDescriptionConstants.UUID;
 import static org.jboss.as.server.controller.resources.DeploymentAttributes.CONTENT_HASH;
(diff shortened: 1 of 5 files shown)


ehsavoie:
retest this please

bstansberry:
retest this please

bstansberry:
DEPLOYMENT_READ_CONTENT_DEFINITION needs to declare the return type for the operation.

A key question is whether we want to return a complex object like you are doing here; i.e. a ModelType.OBJECT with a single field UUID of type STRING. Alternative is to simply return ModelType.STRING, which is simpler for a client to consume.

Advantage of what you have done here though is it leaves a way forward for adding other fields in the future without needing to make an incompatible API change by changing the response type from STRING to OBJECT. @jfdenise any thoughts on this?

(Just an FYI, IIRC @jamezp went through some similar questions when doing the log file read feature. He ended up creating a custom dynamically created Resource for each log file. That's harder to do here though as unlike log files the deployment content files aren't already just sitting there waiting to be modeled as resources. I'm not advocating doing resources for these; this is just FYI on the history of a somewhat similar feature.)

jfdenise:
With respect to the CLI, the structure of the response is supported in both cases. In general, because we have the browse-content that will contain all the file metadata, I am not sure the complex type is worth it. read-content is really retrieving the content, so an uuid should be enough.

bstansberry:
I put a fixme on this as the operation description needs to be updated regardless.

I think just returning the uuid is sufficient; browse-content provides the other data that someone might want.

PR #1759, opened 2016-08-29
[WFCORE-1746]:  Update the browse-content operation to return more information about the content.

ehsavoie:
Returning a list of complex types instead of a list of Strings.

Jira: https://issues.jboss.org/browse/WFCORE-1746


ehsavoie:
retest this please

bstansberry:
retest this please

controller/src/main/java/org/jboss/as/controller/descriptions/ModelDescriptionConstants.java
```diff
@@ -184,6 +184,7 @@
     public static final String FAILURE_DESCRIPTION = "failure-description";
     public static final String FILE = "file";
     public static final String FILE_HANDLER = "file-handler";
+    public static final String FILE_LENGTH = "file-length";
```
bstansberry: LogFileResourceDefinition uses file-size; we should be consistent. Using file-length is closer to the java api but since we've already gone with file-size lets stick with it.

controller/src/main/java/org/jboss/as/controller/descriptions/ModelDescriptionConstants.java
```diff
@@ -192,6 +193,7 @@
     public static final String FILTERED_OPERATIONS = "filtered-operations";
     public static final String FIXED_PORT = "fixed-port";
     public static final String FIXED_SOURCE_PORT = "fixed-source-port";
+    public static final String FOLDER = "folder";
```
bstansberry: This should be directory, not folder

http://superuser.com/questions/169457/directory-vs-folder

bstansberry: Never mind; reading my own link more carefully, 'folder' is better!

bstansberry: Hmm, continuing my debate with myself, I think directory is better. That comment I linked discusses zip file 'folders' but it seems to be in the context of Windows, not as a universal notion.

The zip spec refers to a directory https://pkware.cachefly.net/webdocs/casestudies/APPNOTE.TXT section 4.3.8

Java's ZipEntry class also calls it a directory "isDirectory()"

server/src/main/java/org/jboss/as/server/controller/resources/DeploymentAttributes.java
```diff
@@ -346,10 +355,12 @@ public void marshallAsAttribute(AttributeDefinition attribute, ModelNode resourc
             .build();
     public static final OperationDefinition DEPLOYMENT_READ_CONTENT_DEFINITION = new SimpleOperationDefinitionBuilder(ModelDescriptionConstants.READ_CONTENT, DEPLOYMENT_RESOLVER)
             .setParameters(DEPLOYMENT_CONTENT_PATH)
+            .setReplyParameters(new SimpleAttributeDefinitionBuilder(UUID, ModelType.STRING, false).build())
```
bstansberry: This one doesn't belong in this PR but is missing from the other one. :)

ehsavoie:
retest this please

PR #1762, opened 2016-08-29
[WFCORE-1737] Properly handle non-existent registrations in wildcard read-resource-description ops

bstansberry:
Also includes 2 commits beefing up some testing of wildcard read-resource ops


bstansberry:
retest this please

PR #1763, opened 2016-08-30
Fix for WFCORE-1726. Support for attachments in Response.

jfdenise:
A new attachment command to download/display attachments. Evolved a batch command with an optional ResponseHandler that is called for each step once the composite reply has been received.

### DIFF
--- a/cli/src/main/java/org/jboss/as/cli/handlers/AttachmentHandler.java
+++ b/cli/src/main/java/org/jboss/as/cli/handlers/AttachmentHandler.java
@@ -1,0 +1,404 @@
+/*
+ * JBoss, Home of Professional Open Source.
+ * Copyright 2016, Red Hat, Inc., and individual contributors
+ * as indicated by the @author tags. See the copyright.txt file in the
+ * distribution for a full listing of individual contributors.
+ *
+ * This is free software; you can redistribute it and/or modify it
+ * under the terms of the GNU Lesser General Public License as
+ * published by the Free Software Foundation; either version 2.1 of
+ * the License, or (at your option) any later version.
+ *
+ * This software is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the GNU
+ * Lesser General Public License for more details.
+ *
+ * You should have received a copy of the GNU Lesser General Public
+ * License along with this software; if not, write to the Free
+ * Software Foundation, Inc., 51 Franklin St, Fifth Floor, Boston, MA
+ * 02110-1301 USA, or see the FSF site: http://www.fsf.org.
+ */
+package org.jboss.as.cli.handlers;
+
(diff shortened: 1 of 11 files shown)


bstansberry:
retest this please

bstansberry:
@aloubyansky please approve

aloubyansky:
Looks good. I'd think that in non-batch mode the target file should be overridden. But perhaps there are good reasons to use the index instead.

jfdenise:
@aloubyansky, in non batch mode, in order to preserve all the received streams, I implemented the file indexing. With file overriding we could confuse users that would expect multiple saved file.

aloubyansky:
That makes sense. What I meant though was when you execute the same operation consequently, it does not override the file specified in the argument.
This is not an objection. That would be just my first expectation.

jfdenise:
I see, calling multiple time the same operation. 
I was on the safe side, not alter any user data. If a user file with the same name already exists, I would not override it. Ideally, in a UI context I would ask the user for override or copy.

We could had the option --override to the command (used in conjunction with save) in order to override the file explicitly.

aloubyansky:
It all goes down to user expectation. One thing is to safe as you meant it. Another thing is to have the output at the location specified. I think it could be more important in case of automation/scripts when the output is then consumed and processed.

jfdenise:
I agree, scripts would expect deterministic paths for files. Your experience in this matter makes me feel that we need the override option. Do you think that it would answer the scenarios you are seeing? If yes, then I would introduce the option. We would be safe by default, and override would be explicit. 
Perhaps that you would expect the opposite ;-)?

aloubyansky:
The override option would surely cover the issue I was thinking about. I don't have anything against the safe default. That's actually a great feature and pretty clean implementation. Thanks, Jeff!

jfdenise:
Thanks @aloubyansky for your great feedbacks.

bstansberry:
@aloubyansky please @ mention me when this looks good to you. I'll put a fixme label on it now to stop me checking it. The EAP7 is through analysis. :)

cli/src/main/java/org/jboss/as/cli/handlers/AttachmentHandler.java
```diff
@@ -0,0 +1,399 @@
...
+                            name = name.substring(0, indexed);
+                        } catch (NumberFormatException ex) {
+                            // XXX OK, not a number.
+                        }
+                    } else {
+                        index += 1;
+                    }
+                    targetFile = new File(name + "(" + index + ")");
+                }
+            }
```
aloubyansky: If override is true and the response contains multiple streams, shouldn't the index be incremented here?

jfdenise: Yes, well spotted. PR updated.

aloubyansky:
@bstansberry this is good to merge, thanks!

jfdenise:
Documenting the command, I realise that override should be replaced by overwrite, override doesn't apply well to file. I am making the change and re-sync the PR.

bstansberry:
retest this please

PR #1768, opened 2016-09-01
WFCORE-1757 Openssl support

stuartwdouglas:


domain-management/src/main/java/org/jboss/as/domain/management/security/SecurityRealmResourceDefinition.java
```diff
@@ -48,6 +50,17 @@
  */
 public class SecurityRealmResourceDefinition extends SimpleResourceDefinition {
 
+    static {
+        //register the Openssl Provider, if possible
+        //not really sure if this is the best place for it
```
darranl: I would suggest not including this within this commit.

We still need to rework it further but the Elytron subsystem contains resource(s) to locate and optionally register security providers - we can also make use of discovered providers without actually registering them globally.

darranl:
Apart from my comment re provider registration had a chat with Stuart and this is good to get included ASAP, the incoming Elytron changes also bring SSL configuration so this can start to be used very quickly for the Undertow subsystem.

domain-management/src/main/java/org/jboss/as/domain/management/security/SecurityRealmResourceDefinition.java
stuartwdouglas: Without this OpenSSL is not usable, so there is not much point in merging it. IMHO it is better to just merge this and then it can be removed why Elytron is ready for it.

PR #1771, opened 2016-09-02
[WFCORE-1759] secret value in non-interactive mode

hradecek:
JIRA: https://issues.jboss.org/browse/WFCORE-1759

### DIFF
--- a/domain-management/src/main/java/org/jboss/as/domain/management/security/adduser/AddUser.java
+++ b/domain-management/src/main/java/org/jboss/as/domain/management/security/adduser/AddUser.java
@@ -98,6 +98,9 @@ public class AddUser {
         }
         stateValues.setHowInteractive(howInteractive);
 
+        final boolean shouldDisplaySecret = Boolean.valueOf(argsCliProps.getProperty(CommandLineArgument.DISPLAY_SECRET.key()));
+        stateValues.setDisplaySecret(shouldDisplaySecret);
+
         // Username should not be null or empty.
         if (user == null || user.isEmpty()) {
             nextState = new ErrorState(theConsole, DomainManagementLogger.ROOT_LOGGER.noUsernameExiting(), null, stateValues);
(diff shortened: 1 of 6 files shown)


domain-management/src/main/java/org/jboss/as/domain/management/security/adduser/AddUser.java
```diff
@@ -405,6 +413,12 @@ public String instructions() {
             public String instructions() {
                 return DomainManagementLogger.ROOT_LOGGER.argHelp();
             }
+        },
+        DISPLAY_SECRET("-sv", "--secret") {
```
darranl: Where does the 'v' come from?

darranl:
The original issue was in the wrong project, can you please update the commit and PR to reference WFCORE-1759.

domain-management/src/main/java/org/jboss/as/domain/management/security/adduser/AddUser.java
hradecek: It comes from a "**s**ecret **v**alue", as also _--help_ output would suggest you.
See [DomainManagementLogger.java#L1447](https://github.com/hradecek/wildfly-core/blob/ff3cda09e37c52bd79b635b795e85a35fbcc7095/domain-management/src/main/java/org/jboss/as/domain/management/logging/DomainManagementLogger.java#L1447) for help string.

darranl: IMO I think this may be better to be something along the lines of --display-secret and -ds this way the abbreviated form would have a closer relationship with the full form - also there will be no ambiguity of anyone thinks the argument means to set the secret value.

bstansberry:
this is ok to test

domain-management/src/main/java/org/jboss/as/domain/management/security/adduser/AddUser.java
hradecek: Indeed, that sounds more accurate. I've made the changes.

darranl:
@bstansberry if we get a clean run I think this is good to merge.

PR #1772, opened 2016-09-02
[WFCORE-952] Use standard null checks

dmlloyd:


dmlloyd:
https://issues.jboss.org/browse/WFCORE-952

dmlloyd:
It's retesting, not sure why it's showing failure still.  Whatever

bstansberry:
Looks like modules are missing dependencies on org.wildfly.common.

dmlloyd:
hmm it definitely *was* but it should be fixed in my updated commits, and it seemed to be working locally.  I'll check it again.

dmlloyd:
Should be OK now.  Sorry about that!

dmlloyd:
Guess I didn't run with -DallTests.

host-controller/src/main/java/org/jboss/as/domain/controller/resources/DomainRootDefinition.java
```diff
@@ -361,7 +361,7 @@ public void initialize(ManagementResourceRegistration resourceRegistration) {
         @Override
         public void validateParameter(String parameterName, ModelNode plan) throws OperationFailedException {
             if(plan == null) {
-                throw new OperationFailedException(DomainControllerLogger.ROOT_LOGGER.nullVar("plan").getLocalizedMessage());
+                throw new OperationFailedException(DomainControllerLogger.ROOT_LOGGER.nullVar("plan"));
```
bstansberry: This should have been an assert in the first place -- https://issues.jboss.org/browse/WFCORE-1769. I'll fix that after I merge this.

bstansberry:
If you plan to squash let me know; otherwise I'll merge this as is in my next batch.

PR #1778, opened 2016-09-06
Fix for WFCORE-1767. Exception on ValueType completion

jfdenise:
Fixed the problem highlighted by the jgroup protocol complex type. A property named type was part of the set of valid properties. It revealed incorrect piece of code.

- Found an issue during testing with List of simple types.
- Added new unit tests for jgroups and list of simple types.


aloubyansky:
This change looks good to merge, thanks!

PR #1779, opened 2016-09-06
[WFCORE-1768] Update the _private exclusions for the Elytron module definition.

darranl:

https://issues.jboss.org/browse/WFCORE-1768
https://issues.jboss.org/browse/JBEAP-5877

### DIFF
--- a/core-feature-pack/src/main/resources/modules/system/layers/base/org/wildfly/security/elytron/main/module.xml
+++ b/core-feature-pack/src/main/resources/modules/system/layers/base/org/wildfly/security/elytron/main/module.xml
@@ -26,7 +26,9 @@
 
     <exports>
         <exclude path="org/wildfly/security/_private"/>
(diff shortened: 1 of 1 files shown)


core-feature-pack/src/main/resources/modules/system/layers/base/org/wildfly/security/elytron/main/module.xml
```diff
@@ -26,7 +26,9 @@
 
     <exports>
         <exclude path="org/wildfly/security/_private"/>
-        <exclude path="org/wildfly/wildfly/security/manager/_private"/>
+        <exclude path="org/wildfly/security/manager/_private"/>
+        <exclude path="org/wildfly/security/sasl.digest/_private"/>
```
rhusar: That seems wrong, looke like it should be sasl.digest -> sasl/digest

PR #1781, opened 2016-09-07
[WFCORE-1556] Correct the handling of required vs nillable

bstansberry:
https://issues.jboss.org/browse/WFCORE-1556
Also https://issues.jboss.org/browse/WFCORE-1761

### DIFF
--- a/controller/src/main/java/org/jboss/as/controller/AbstractAttributeDefinitionBuilder.java
+++ b/controller/src/main/java/org/jboss/as/controller/AbstractAttributeDefinitionBuilder.java
@@ -90,12 +90,12 @@ public abstract class AbstractAttributeDefinitionBuilder<BUILDER extends Abstrac
      * Creates a builder for an attribute with the give name and type and nullability setting.
      * @param attributeName the {@link AttributeDefinition#getName() name} of the attribute. Cannot be {@code null}
      * @param type the {@link AttributeDefinition#getType() type} of the attribute. Cannot be {@code null}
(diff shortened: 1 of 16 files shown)


server/src/main/java/org/jboss/as/server/mgmt/NativeManagementResourceDefinition.java
```diff
@@ -56,7 +56,6 @@
     public static final SimpleAttributeDefinition SOCKET_BINDING = new SimpleAttributeDefinitionBuilder(ModelDescriptionConstants.SOCKET_BINDING, ModelType.STRING, false)
             .setXmlName(Attribute.NATIVE.getLocalName())
             .setValidator(new StringLengthValidator(1, Integer.MAX_VALUE, true, false))
-            .setAlternatives(ModelDescriptionConstants.INTERFACE)
```
ctomc: we should probably remove all references to INTERFACE / PORT by this time,
It would also simplify code that handles is quite a bit.

bstansberry: I don't see any uses of those on the server side. They are there on the host side, where they are the standard settings.

An HC process can use socket binding configs now, so we can look into adding that to the host management-interface resources. We'd need to deprecate and support the old style though, probably until EAP 8.

bstansberry:
@jmesnil Does this look ok to you?

PR #1784, opened 2016-09-08
Fix for WFCORE-1771, exception in :composite completion.

jfdenise:
Fixed the exception in the case where the List is a List of OBJECT (no specified value type).
Fixed the completion of "[<TAB>" in this case. It was not proposing "{".
Added unit test.


aloubyansky:
Good to merge, thanks!

PR #1788, opened 2016-09-10
[WFCORE-1781] Don't install DC op handlers for a slave using cached-dc

bstansberry:
https://issues.jboss.org/browse/WFCORE-1781

### DIFF
--- a/core-model-test/framework/src/main/java/org/jboss/as/core/model/test/TestModelControllerService.java
+++ b/core-model-test/framework/src/main/java/org/jboss/as/core/model/test/TestModelControllerService.java
@@ -414,11 +414,6 @@ class TestModelControllerService extends ModelTestModelControllerService {
             }
 
             @Override
-            public RunningMode getCurrentRunningMode() {
-                return null;
-            }
-
-            @Override
             public ExpressionResolver getExpressionResolver() {
                 return null;
             }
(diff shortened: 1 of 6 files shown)


luck3y:
Aha, good catch! +1

bstansberry:
It's not quite right though; problem is broader.

1) We shouldn't install this for an admin-only DC either.
2) However, HostControllerRegistrationHandler.InitiateRegistrationHandler performs a useful service of throwing particular exceptions when the HC is not master or is admin-only. Data from those are meant to propagate to the slave and result in more informative connection failure message. But that logic is only getting invoked sometimes, and with this fix and a fix for 1) above it will happen even less.

For 2) I'm thinking about continuing with this approach here but separately registering something on slaves or admin-only masters to do the rejection currently done in  HostControllerRegistrationHandler.InitiateRegistrationHandler.

bstansberry:
[~luck3y] I think this is ok now. Crossing fingers on CI

bstansberry:
I'm going to close this for now. It came out of the HA DC work, although it's valid even without it. But the main thinking about it was in the context of that work, and that's on hold until the next release. For 3.0 I'd rather not introduce a change like this that's not terribly urgent. Will reopen when we start on the next release.

PR #1789, opened 2016-09-12
[WFCORE-1783] CLI GUI throws exception when editing boolean attribute

hradecek:
JIRA: [WFCORE-1783](https://issues.jboss.org/browse/WFLY-6739)
Use default value when boolean attribute is an expression.
Remove duplicated if condition.

### DIFF
--- a/cli/src/main/java/org/jboss/as/cli/gui/OperationDialog.java
+++ b/cli/src/main/java/org/jboss/as/cli/gui/OperationDialog.java
@@ -317,23 +317,23 @@ public class OperationDialog extends JDialog {
             this.value = value;
             this.type = props.get("type").asType();
 
-            if (props.get("description").isDefined()) {
+            if (props.hasDefined("description")) {
                 this.description = props.get("description").asString();
             }
 
-            if (props.get("required").isDefined()) {
+            if (props.hasDefined("required")) {
                 this.isRequired = props.get("required").asBoolean();
             }
 
-            if (props.get("nillable").isDefined()) {
+            if (props.hasDefined("nillable")) {
                 this.nillable = props.get("nillable").asBoolean();
             }
 
-            if (props.get("expressions-allowed").isDefined()) {
+            if (props.hasDefined("expressions-allowed")) {
                 this.expressionsAllowed = props.get("expressions-allowed").asBoolean();
             }
 
-            if (props.get("default").isDefined()) {
+            if (props.hasDefined("default")) {
                 this.defaultValue = props.get("default");
             }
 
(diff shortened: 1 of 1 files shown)


darranl:
The WFLY issue in Jira needs to be moved to the WFCORE project, then the commit message and PR needs to be updated to reference the new issue.

hradecek:
@darranl - Moved and updated.

cli/src/main/java/org/jboss/as/cli/gui/OperationDialog.java
```diff
@@ -328,6 +329,10 @@ public RequestProp(String name, ModelNode props, ModelNode value) {
                 this.nillable = props.get("nillable").asBoolean();
             }
 
+            if (props.get("expressions-allowed").isDefined()) {
```
bstansberry: Use

```
props.hasDefined("expressions-allowed")
```

The ModeNode "get" method actually mutates the model node by adding an undefined child under that key. This makes a fluent API like node.get("enabled").set(true) possible. But it also means we need to use "hasDefined" for this kind of check to avoid unintended mutation of the node. In this case it probably doesn't matter, but we enforce this as a code style kind of thing so people always see the correct usage and thus use it when it really does matter.

cli/src/main/java/org/jboss/as/cli/gui/OperationDialog.java
```diff
@@ -328,6 +329,10 @@ public RequestProp(String name, ModelNode props, ModelNode value) {
...
+                this.expressionsAllowed = props.get("expressions-allowed").asBoolean();
+            }
+
             if (props.get("default").isDefined()) {
```
bstansberry: LOL, and of course the wrong style is right here. ;)

Oh, well, even so you might as well make your new code proper.

cli/src/main/java/org/jboss/as/cli/gui/OperationDialog.java
```diff
@@ -424,10 +429,6 @@ private void setInputComponentValue() {
 
             if (!valueToSet.isDefined()) return;
 
-            if (valueComponent instanceof JCheckBox) {
-                ((JCheckBox)this.valueComponent).setSelected(valueToSet.asBoolean());
-            }
```
bstansberry: Should this still be deleted?

hradecek: Answer lies [8 lines](https://github.com/wildfly/wildfly-core/pull/1789/files#diff-87f9841e29fc742b456448d0beca613bL435) below :).

bstansberry: :)

PR #1790, opened 2016-09-12
Fix for WFCORE-1784. Attachment to use the right target directory in …

jfdenise:
…all cases.
Fixed the case and added unit test.


aloubyansky:
This is good to merge, thanks!

PR #1792, opened 2016-09-13
Fix for WFCORE-1787. Exception when traversing complex Objects

jfdenise:
Fixed the case where the Value-type advertise an OBJECT type, with Value-Type being a String although the actual content is an Object.
Added unit test.


aloubyansky:
Looks good to merge, thanks!

PR #1797, opened 2016-09-14
WFCORE-1786: Allow for custom ContainerPersisterFactory in SelfContainedContainer

heiko-braun:


heiko-braun:
@bstansberry Assuming this change is acceptable for you, how would we best proceed to bring that into swarm? Backporting to 2.x? Waiting for 11.0? What's the schedule for the later?

heiko-braun:
Relates to https://issues.jboss.org/browse/SWARM-612

bstansberry:
@heiko-braun I don't forsee WildFly Core 3.0.0.Final being released before mid-November, perhaps later. WildFly 11 will even more likely be later than that.

We can do a WildFly Core 2.2.1.Final pretty much any time. It will never go into a WildFly release unless we need to do another 10.x release for some reason, but I have no problem doing the core release.

bstansberry:
If you want the 2.2.1 route, please ping me, and add 2.2.1 as a Fix Version to WFCORE-1786.

heiko-braun:
thanks @bstansberry . I'll discuss that with @bobmcwhirter and @kenfinnigan . We have to see if a separate wf-core release can be used. IMO that would be the best way to move forward.

heiko-braun:
As a side note: there are other, swarm related wf-core changes in the pipeline as well. most of it around the self contained server

kenfinnigan:
Currently we consume WF core via its feature pack, which is brought in via depMan with:

```
      <dependency>
        <groupId>org.wildfly</groupId>
        <artifactId>wildfly-parent</artifactId>
        <version>${version.wildfly}</version>
        <type>pom</type>
        <scope>import</scope>
      </dependency>
```

In theory we should be able to add the WF core feature pack, with a different version, into depMan to override.

heiko-braun:
@kenfinnigan Yes, I think separate deps for WF core and WF would be helpful in preparation to simplify maintenance. As @bstansberry said, it's usually easier to release core then WF.

PR #1801, opened 2016-09-15
[WFCORE-1793]: add-content operation fails to overwrite existing content with overwrite=true set when passing content by file path

ehsavoie:
Updating the operation description with was wrong and adding a new validator.

Jira: https://issues.jboss.org/browse/WFCORE-1793


bstansberry:
How hard would it be to get the whitespace change out of this? The DeploymentAttributes class has a lot of stuff and having git blame pointing to this commit for lots of it isn't great.

PR #1809, opened 2016-09-16
WFCORE-1772 /core-service=capability-registry:get-provider-points onl…

ctomc:
…y returning a single provider point.


bstansberry:
#1810 ignores the testPathInfo failure. The mixed domain failure at a glance looks like a FS space issue on the CI server.

PR #1812, opened 2016-09-19
Change port used by JMX tests from 12345 to prevent collisions

honza889:
JMX tests was using port 12345, which caused unexplainable "random" JMX tests failing - tests failed when HipChat was running.

Changed to (hopefuly) less used port 27258 and extracted to constant.


honza889:
Hmm.. closed becouse it needs remoting change yet, but now cannot reopen...

PR #1813, opened 2016-09-19
Change port used by JMX/remoting tests to prevent collisions

honza889:
JMX tests was using port 12345, which caused unexplainable "random" JMX tests failing - tests failed when HipChat was running.

Changed to (hopefuly) less used port 27258 and extracted to constant.


darranl:
@honza889 FYI for changes to the WildFly Core project you should generally have a WFCORE issue in Jira so you can reference it in the commit and pull request.

Admins, can we please get Honza added to the list of approved committers so his PRs can be run automatically?

bstansberry:
this is ok to test

PR #1814, opened 2016-09-19
WFCORE-1797: Move temp file handling to SelfContainerServer

heiko-braun:
The temp files created in `LaunchType.SELF_CONTAINED` are now managed by the `SelfContainedContainer`.  Alongside I've added a `SelfContainedContainer#stop()` to explicitly manage the teardown lifecycle.

### DIFF
--- a/server/src/main/java/org/jboss/as/server/ServerEnvironment.java
+++ b/server/src/main/java/org/jboss/as/server/ServerEnvironment.java
@@ -365,31 +365,11 @@ public class ServerEnvironment extends ProcessEnvironment implements Serializabl
             serverLogDir = new File(WildFlySecurityManager.getPropertyPrivileged("user.dir", "."));
 
 
-            try {
-                File tmpDir = File.createTempFile( "wildfly-self-contained", ".d" );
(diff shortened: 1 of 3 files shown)


server/src/main/java/org/jboss/as/server/ServerEnvironment.java
```diff
@@ -365,31 +365,11 @@ public ServerEnvironment(final String hostControllerName, final Properties props
...
-                    }
-                }
-                tmpDir.mkdirs();
-                tmpDir.deleteOnExit();
-                serverTempDir = tmpDir;
-            } catch (IOException e) {
-                throw new RuntimeException(e);
+            String serverDirProp = props.getProperty(SERVER_TEMP_DIR);
+            if(null==serverDirProp) {
+                throw new IllegalStateException("LaunchType.SELF_CONTAINED requires the property "+SERVER_TEMP_DIR+ " to be set");
```
bstansberry: This needs i18n.

heiko-braun:
Added the i18n logging as requested

PR #1816, opened 2016-09-20
[WFCORE-1547] Iniital CredentialStore Integration

darranl:

This is initial commit that introduces:

- CredentialReference class unifying access to credentials defined through CredentialStore or holding simple secret value. It defines credential reference attribute that other subsystems can use to reference external credentials of various types.
- CredentialStoreClient class used to talk to CredentialStore to obtain secret value (credential) from it.
- CredentialStoreURIParser class that parses configuration or reference URIs for easier parameter marshaling

These changes are tracked under the following Jira issue: -
  https://issues.jboss.org/browse/WFCORE-1547

This Jira issue can be closed once the PR is merged.

If this PR is accepted is there also any chance of an early tag of WildFly Core, we have subsystem changes to follow up with but first require a tagged release we can reference.

### DIFF
--- a/controller/src/main/java/org/jboss/as/controller/logging/ControllerLogger.java
+++ b/controller/src/main/java/org/jboss/as/controller/logging/ControllerLogger.java
@@ -3449,33 +3449,33 @@ public interface ControllerLogger extends BasicLogger {
 
     // ---- Credential Store related messages -----------
 
-    @Message(id = 450, value = "Credential store \"%s\" doesn't support credential type \"%s\" ")
+    @Message(id = 423, value = "Credential store \"%s\" doesn't support credential type \"%s\" ")
     UnsupportedCredentialTypeException unsupportedCredentialType(String credentialStore, String credentialType);
 
-    @Message(id = 451, value = "Supposed Credential Store URI has no scheme or different from '" + CredentialStoreURIParser.CR_STORE_SCHEME + "://' ('%s')")
+    @Message(id = 424, value = "Supposed Credential Store URI has no scheme or different from '" + CredentialStoreURIParser.CR_STORE_SCHEME + "://' ('%s')")
     IllegalArgumentException credentialStoreURIWrongScheme(String uri);
 
-    @Message(id = 452, value = "Credential Store URI has to be absolute '%s'")
+    @Message(id = 425, value = "Credential Store URI has to be absolute '%s'")
     IllegalArgumentException credentialStoreURIisNotAbsolute(String uri);
 
-    @Message(id = 453, value = "More than one fragment defined for Credential Store URI")
+    @Message(id = 426, value = "More than one fragment defined for Credential Store URI")
     String moreThanOneFragmentDefined();
 
-    @Message(id = 454, value = "Credential Store name has to be defined '%s'")
+    @Message(id = 427, value = "Credential Store name has to be defined '%s'")
     IllegalArgumentException credentialStoreHasNoName(String uri);
 
-    @Message(id = 455, value = "Attribute name is defined, but empty '%s'")
+    @Message(id = 428, value = "Attribute name is defined, but empty '%s'")
     IllegalArgumentException credentialStoreURIAttributeEmpty(String uri);
 
-    @Message(id = 456, value = "Opening quote has to be the first character in parameter value '%s'")
+    @Message(id = 429, value = "Opening quote has to be the first character in parameter value '%s'")
     IllegalArgumentException credentialStoreURIParameterOpeningQuote(String uri);
 
-    @Message(id = 457, value = "Closing quote has to be the last character of parameter value '%s'")
+    @Message(id = 430, value = "Closing quote has to be the last character of parameter value '%s'")
     IllegalArgumentException credentialStoreURIParameterClosingQuote(String uri);
 
-    @Message(id = 458, value = "Unexpected end of parameter part of '%s'")
+    @Message(id = 431, value = "Unexpected end of parameter part of '%s'")
     IllegalArgumentException credentialStoreURIParameterUnexpectedEnd(String uri);
 
-    @Message(id = 459, value = "Parameter name expected, but is missing '%s'")
+    @Message(id = 432, value = "Parameter name expected, but is missing '%s'")
     IllegalArgumentException credentialStoreURIParameterNameExpected(String uri);
 }
--- a/controller/src/main/java/org/jboss/as/controller/logging/ControllerLogger.java
+++ b/controller/src/main/java/org/jboss/as/controller/logging/ControllerLogger.java
@@ -62,6 +62,7 @@ import org.jboss.as.controller.persistence.ConfigurationPersistenceException;
 import org.jboss.as.controller.registry.AttributeAccess;
 import org.jboss.as.controller.registry.OperationEntry;
 import org.jboss.as.controller.registry.Resource;
+import org.jboss.as.controller.security.CredentialStoreURIParser;
 import org.jboss.as.protocol.mgmt.RequestProcessingException;
 import org.jboss.dmr.ModelNode;
 import org.jboss.dmr.ModelType;
@@ -78,6 +79,7 @@ import org.jboss.modules.ModuleLoadException;
 import org.jboss.modules.ModuleNotFoundException;
 import org.jboss.msc.service.ServiceName;
 import org.jboss.msc.service.StartException;
+import org.wildfly.security.credential.store.UnsupportedCredentialTypeException;
 
 /**
  * @author <a href="mailto:jperkins@redhat.com">James R. Perkins</a>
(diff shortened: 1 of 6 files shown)


controller/src/main/java/org/jboss/as/controller/security/CredentialStoreClient.java
```diff
@@ -0,0 +1,190 @@
...
+    }
+
+    /**
+     * Retrieve associated {@link CredentialStore}
+     * This method should perform check whether the {@link CredentialStore} can be returned to caller.
+     *
+     * @return associated {@link CredentialStore}
+     */
+    public CredentialStore getCredentialStore() {
+        // TODO: check caller if possible
```
bstansberry: Is there a JIRA to take care of this?

Also, is it necessary to have this method public? Perhaps other code outside this PR uses it but the only use I see here is in this package to create a CredentialStoreClient from another CredentialStoreClient. So that can be done without public visibility.

controller/src/main/java/org/jboss/as/controller/logging/ControllerLogger.java
```diff
@@ -3444,4 +3446,36 @@ XMLStreamException discoveryOptionsMustBeDeclared(String adminOnlyCmd, String po
 
     @Message(id = 422, value = "Could not load module '%s' for transformers")
     RuntimeException couldNotLoadModuleForTransformers(String name, @Cause ModuleLoadException e);
+
```
bstansberry: Can you add a comment that a bunch of ids were deliberately skipped? That will make it less mysterious a few years from now.

controller/src/main/java/org/jboss/as/controller/security/CredentialReference.java
```diff
@@ -0,0 +1,417 @@
...
+     * on this {@code Object} will result in an
+     * {@code IllegalStateException} being thrown.
+     * <p>
+     * <p>
+     * The default implementation throws {@code DestroyFailedException}.
+     *
+     * @throws DestroyFailedException if the destroy operation fails. <p>
+     * @throws SecurityException      if the caller does not have permission
+     *                                to destroy this {@code Object}.
+     */
```
bstansberry: This javadoc is confusing, as it's not how the method actually works.

controller/src/main/java/org/jboss/as/controller/security/CredentialReference.java
```diff
@@ -0,0 +1,417 @@
...
+        if (secret != null) {
+            this.secret = secret.clone();
+        } else {
+            this.secret = null;
+        }
+    }
+
+    /**
+     * Get the credential store name part of this reference.
+     * @return credential store name
```
bstansberry: adding "May return {@code null}" is good.

controller/src/main/java/org/jboss/as/controller/security/CredentialReference.java
```diff
@@ -0,0 +1,417 @@
...
+     */
+    public String getCredentialStoreName() {
+        return credentialStoreName;
+    }
+
+    /**
+     * Get the credential alias which denotes credential stored inside named credential store.
+     * @return alias of the referenced credential
```
bstansberry: adding "May return {@code null}" is good.

controller/src/main/java/org/jboss/as/controller/security/CredentialReference.java
```diff
@@ -0,0 +1,417 @@
...
+     */
+    public String getAlias() {
+        return alias;
+    }
+
+    /**
+     * Get credential type which narrows selection of the credential stored under the alias in the credential store.
+     * @return credential type (class name of desired credential type)
```
bstansberry: adding "May return {@code null}" is good.

controller/src/main/java/org/jboss/as/controller/security/CredentialReference.java
```diff
@@ -0,0 +1,417 @@
...
+     */
+    public String getCredentialType() {
+        return credentialType;
+    }
+
+    /**
+     * Get the secret stored as clear text in this reference.
+     * @return secret value as clear text
```
bstansberry: adding "May return {@code null}" is good.

controller/src/main/java/org/jboss/as/controller/security/CredentialReference.java
```diff
@@ -0,0 +1,417 @@
...
+
+    /**
+     * Determine if this {@code Object} has been destroyed.
+     * <p>
+     * <p>
+     * The default implementation returns false.
+     *
+     * @return true if this {@code Object} has been destroyed,
+     * false otherwise.
+     */
```
bstansberry: Incorrect javadoc. 

There's no need to copy javadoc from the interface.

controller/src/main/java/org/jboss/as/controller/security/CredentialReference.java
```diff
@@ -0,0 +1,417 @@
...
+        injectedCredentialStoreClient.setValue(() -> updatedCredentialStoreClient);
+    }
+
+    /**
+     * Marshall the value from {@code credentialReferenceModelNode} as an xml element into the given {@code writer}.
+     * @param credentialReferenceModelNode he model, a non-null node of {@link org.jboss.dmr.ModelType#OBJECT}.
+     * @param writer stream writer to use for writing the attribute
+     * @throws XMLStreamException if thrown by {@code writer}
+     */
+    public static void marshallAsElement(ModelNode credentialReferenceModelNode, XMLExtendedStreamWriter writer) throws XMLStreamException {
```
bstansberry: I'm being lazy and not researching here, so apologies if this is silly.

ObjectTypeAttributeDefinition.Builder exposes setAttributeMarshaller and setAttributeParser. If the expected uses of this method and readCredentialReference can be covered by using that to set up credentialReferenceAttributeDefinition, and then the callers just use getAttributeDefinition(), then these two methods don't need to be public.

If possible I'd much prefer to not have public utility code for parsing and marshalling proliferating around.

controller/src/main/java/org/jboss/as/controller/security/CredentialReference.java
```diff
@@ -0,0 +1,417 @@
...
+     * Parse {@link CredentialReference} from {@code XMLStreamReader} and set addOperation.
+     * @param addOperation to set
+     * @param reader stream reader to parse
+     * @throws XMLStreamException if thrown by {@code reader}
+     */
+    public static void readCredentialReference(ModelNode addOperation, XMLExtendedStreamReader reader) throws XMLStreamException {
+        AttributeParser.OBJECT_PARSER.parseElement(CredentialReference.getAttributeDefinition(), reader, addOperation);
+    }
+
+    static class ClearTextCredentialStore extends CredentialStore {
```
bstansberry: This can be private.

controller/src/main/java/org/jboss/as/controller/security/CredentialReference.java
```diff
@@ -0,0 +1,417 @@
...
+         * Each must be of different credential type.
+         *
+         * @param credentialAlias to store the credential to the store
+         * @param credential      instance of {@link Credential} to store
+         * @throws CredentialStoreException           when the credential cannot be stored
+         * @throws UnsupportedCredentialTypeException when the credentialType is not supported
+         */
+        @Override
+        public <C extends Credential> void store(String credentialAlias, C credential) throws CredentialStoreException, UnsupportedCredentialTypeException {
+            throw new RuntimeException("method not implemented");
```
bstansberry: Should this be one of the 2 documented exception types?

I only care if this class isn't private.

controller/src/main/java/org/jboss/as/controller/security/CredentialReference.java
```diff
@@ -0,0 +1,417 @@
...
+         * Remove the credentialType with from given alias from the store.
+         *
+         * @param credentialAlias alias to remove
+         * @param credentialType  - credential type to be removed from under the credentialAlias from the store
+         * @throws CredentialStoreException           - if credentialAlias credentialType combination doesn't exist or credentialAlias cannot be removed
+         * @throws UnsupportedCredentialTypeException when the credentialType is not supported
+         */
+        @Override
+        public <C extends Credential> void remove(String credentialAlias, Class<C> credentialType) throws CredentialStoreException, UnsupportedCredentialTypeException {
+            throw new RuntimeException("method not implemented");
```
bstansberry: Should this be one of the 2 documented exception types?

I only care if this class isn't private.

controller/src/main/java/org/jboss/as/controller/security/CredentialStoreClient.java
```diff
@@ -0,0 +1,190 @@
...
+        this(credentialStore, name, alias, toClass(className, credentialStore.getClass().getClassLoader()));
+    }
+
+    private static Class<? extends Credential> toClass(final String className, final ClassLoader classLoader) throws ClassNotFoundException {
+        return (Class<Credential>) Class.forName(className, true, classLoader);
+    }
+
+    /**
+     * Get the secret in form of clear text.
+     * @return secret as clear text
```
bstansberry: May return {@code null}

controller/src/main/java/org/jboss/as/controller/security/CredentialStoreClient.java
```diff
@@ -0,0 +1,190 @@
...
+import org.wildfly.common.Assert;
+import org.wildfly.security.credential.Credential;
+import org.wildfly.security.credential.PasswordCredential;
+import org.wildfly.security.credential.store.CredentialStore;
+import org.wildfly.security.credential.store.CredentialStoreException;
+import org.wildfly.security.credential.store.UnsupportedCredentialTypeException;
+import org.wildfly.security.password.interfaces.ClearPassword;
+
+/**
+ * Client class used to talk to {@link CredentialStore} to obtain secret value (credential) from it.
```
bstansberry: This class should be documented as not being thread safe of should be made thread safe.

controller/src/main/java/org/jboss/as/controller/security/CredentialReference.java
```diff
@@ -0,0 +1,417 @@
...
+import org.wildfly.security.password.interfaces.ClearPassword;
+import org.wildfly.security.password.spec.ClearPasswordSpec;
+
+/**
+ * Class unifying access to credentials defined through {@link org.wildfly.security.credential.store.CredentialStore}
+ * or holding simply {@code char[]} as a secret.
+ *
+ * It defines credential reference attribute that other subsystems can use to reference external credentials of various
+ * types.
+ *
```
bstansberry: This class should be documented as not being thread safe or should be made thread safe.

bstansberry:
Lots of nitpicky comments which are mostly due to this being public API in controller module, so I'm at my most picky. My only substantial point is whether the marshall/parse util methods can be hidden away.

controller/src/main/java/org/jboss/as/controller/security/CredentialStoreClient.java
pskopek: Yes, this method has to be public. I have created JIRA to track caller check. I prefer to do it in separate PR.
Here is the JIRA: https://issues.jboss.org/browse/WFCORE-1811

pskopek:
I have resolved all comments you have made (except the first one, see my partial comment). 
Update to this PR is coming shortly.

darranl:
Retest this please

bstansberry:
CredentialReference still isn't thread safe or documented as not being thread safe, but I'm not going to delay merging this over that. Thanks for all the tweaks. :)

PR #1817, opened 2016-09-20
WFCORE-1773 Fix domain backup copy persistence when using --backup or --cached-dc.

luck3y:
https://issues.jboss.org/browse/WFCORE-1773


luck3y:
@ehsavoie FYI, this was my issue with model sync, it affected persisting only. The rest of model sync is fine, so I doubt this has any effect on your standalone sync change.

PR #1819, opened 2016-09-21
[WFCORE-1812]: Makes ModuleOpsCompletionTestCase more resilient.

ehsavoie:
Cleaning modules directory before and after the test.

Jira: https://issues.jboss.org/browse/WFCORE-1812


bstansberry:
This PR will also need to revert #1829

PR #1821, opened 2016-09-22
Fix for WFCORE-1813. Add command-timeout element to CLI schema

jfdenise:


cli/src/main/resources/schema/wildfly-cli_3_1.xsd
```diff
@@ -140,6 +140,14 @@
                         </xs:documentation>
                     </xs:annotation>
                 </xs:element>
+
+                <xs:element name="command-timeout" type="xs:int" default="0" minOccurs="0">
+                    <xs:annotation>
+                        <xs:documentation>
+                            The maxium amount of seconds the CLI will wait for a command to complete.
```
bstansberry: s/maxium/maximum

Also, what is the behavior of a value of 0? I believe it means "no limit".

PR #1822, opened 2016-09-22
WFCORE-1814 Move to woodstox 5.0.3

ctomc:
This will come handy in future wrt JDK9

Changelist vs 4.4.x we ware using before 

5.0.3 (23-Aug-2016)

- 13: `BasicStreamReader.getElementText()` behavior doesn't match Java documentation
- 15: `BasicStreamReader` should use `WstxUnexpectedCharException` instead of WstxParsingException
- 16: Incorrect validation error(s) only using Stax2 writer
- 21: 500 characters limit when calling XMLStreamReader#getText() after CDATA event

5.0.2 (10-Dec-2015)

- 10: Stax 4.0.0 is not compatible with Woodstox 5.0.1

5.0.1 (02-Apr-2015)
- 3: Initial 5.0.0 release did not contain class files (d'oh!)
- 4: Fix validation of text for an XMLStreamWriter

5.0.0 
- Attempted initial release under new home on github with new GAV

### DIFF
--- a/cli/pom.xml
+++ b/cli/pom.xml
@@ -201,8 +201,8 @@
 
         <!-- Test Dependencies -->
         <dependency>
-            <groupId>org.codehaus.woodstox</groupId>
-            <artifactId>woodstox-core-asl</artifactId>
+            <groupId>com.fasterxml.woodstox</groupId>
+            <artifactId>woodstox-core</artifactId>
             <scope>test</scope>
         </dependency>
         <dependency>
--- a/core-feature-pack/pom.xml
+++ b/core-feature-pack/pom.xml
@@ -55,8 +55,8 @@
         </dependency>
 
         <dependency>
-            <groupId>org.codehaus.woodstox</groupId>
-            <artifactId>woodstox-core-asl</artifactId>
+            <groupId>com.fasterxml.woodstox</groupId>
+            <artifactId>woodstox-core</artifactId>
         </dependency>
 
         <dependency>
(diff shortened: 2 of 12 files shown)


jamezp:
We should validate that CXF works with this new version too. It will likely be fine but the following worried me a bit.

> 10: Stax 4.0.0 is not compatible with Woodstox 5.0.1

bstansberry:
@asoldano Please approve this change, as it seems CXF depends on woodstox.

asoldano:
@bstansberry Apache CXF is still relying on Woodstox 4.4.1, so I need some time to try if it still works with 5 or not.

ctomc:
@jamezp 5.0.3 fixes the stax 4.0.0 compatiblity.
But that is not issue any any way as we don't use stax api 4.0.0
Also, this is https://github.com/FasterXML/stax2-api stax api 4.0 not the java one. 4.0 is just refactored 3.1 see https://github.com/FasterXML/stax2-api/blob/master/release-notes/VERSION

In short woodstox 4.4 we used before and 5.0.3 are the pretty much same, with just new GAV and source repo (codehaus --> github) with few bugs fixed.
@asoldano So compatibility is not an issue in any way.

asoldano:
@bstansberry @ctomc I've tried running jbossws-cxf testsuite on top of woodstox-upgrade branches (wfly-core and wfly) and it's passing. So I'd say we can move to 5 on wfly/wfly-core master.

PR #1825, opened 2016-09-23
Fix for WFCORE-1774. Suggest capabilities and CLI capability completion.

jfdenise:
This change incorporates cli, controller and server changes.
Something to mention is that I am using the Java FileSystem glob pattern matcher to match the capabilities registration points with the possible registration points. Looking at the /xxx/* kind of syntax, it is compliant with the Filesystem matcher (and avoid to create a special Java Pattern for these paths).

A controller unit test has been added to cover the main combinations of capability resolution according to the existing scopes.
This test is quite long and it could be reduced to cover fewer nominal cases. Let me know if you are OK to keep it or would like it to be reduced even removed if it goes too far in the knowledge of scopes.

### DIFF
--- a/controller/src/main/java/org/jboss/as/controller/CapabilityRegistry.java
+++ b/controller/src/main/java/org/jboss/as/controller/CapabilityRegistry.java
@@ -1,6 +1,6 @@
 /*
  * JBoss, Home of Professional Open Source.
- * Copyright 2014, Red Hat, Inc., and individual contributors
+ * Copyright 2016, Red Hat, Inc., and individual contributors
  * as indicated by the @author tags. See the copyright.txt file in the
  * distribution for a full listing of individual contributors.
  *
@@ -22,6 +22,9 @@
 
 package org.jboss.as.controller;
 
+import java.nio.file.FileSystems;
+import java.nio.file.PathMatcher;
+import java.nio.file.Paths;
 import static org.jboss.as.controller.descriptions.ModelDescriptionConstants.HOST;
 import static org.jboss.as.controller.descriptions.ModelDescriptionConstants.SUBSYSTEM;
 
@@ -70,7 +73,7 @@ public final class CapabilityRegistry implements ImmutableCapabilityRegistry, Po
     private final boolean forServer;
     private final Set<CapabilityScope> knownContexts;
     private final ResolutionContextImpl resolutionContext = new ResolutionContextImpl();
-    private final Map<CapabilityId, CapabilityRegistration> possibleCapabilities = new ConcurrentHashMap<>();
+    private final Map<CapabilityId, CapabilityRegistration<?>> possibleCapabilities = new ConcurrentHashMap<>();
     private final Set<CapabilityId> reloadCapabilities = new HashSet<>();
     private final Set<CapabilityId> restartCapabilities = new HashSet<>();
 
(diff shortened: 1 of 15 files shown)


controller/src/main/java/org/jboss/as/controller/CapabilityRegistry.java
```diff
@@ -905,6 +909,45 @@ private SatisfactoryCapability findSatisfactoryCapability(String capabilityName,
         return null;
     }
 
+    @Override
```
ctomc: Could you please add some basic javadoc.

controller/src/main/java/org/jboss/as/controller/CapabilityRegistry.java
```diff
@@ -905,6 +909,45 @@ private SatisfactoryCapability findSatisfactoryCapability(String capabilityName,
...
+                || dependentScope == null) {
+            return Collections.emptySet();
+        }
+
+        // Retrieve all the provider points that matching capabilities
+        // must be compliant with. Each ProviderPoint is converted onto a
+        // PathMatcher that is used to filter-in/out the capabilities.
+
+        //For possible capabilities it is always global.
+        CapabilityId id = new CapabilityId(referencedCapability,
```
ctomc: So idea here is to again look trough possible capabilities registry. Not to have completion for actual capabilities already registered.
"Problem" with :get-possible-provider-points operation is that returns possible points that is why there are wildcards in the addresses.

controller/src/main/java/org/jboss/as/controller/CapabilityRegistry.java
```diff
@@ -905,6 +909,45 @@ private SatisfactoryCapability findSatisfactoryCapability(String capabilityName,
...
+        return getCapabilities().stream().
+                // Keep capability that matches at least one of the registration point
+                filter((registration) -> {
+                    return registration.getRegistrationPoints().stream().
+                            map((rp) -> Paths.get(rp.getAddress().toPathStyleString())).
+                            filter((path) -> matchers.stream().anyMatch((matcher) -> matcher.matches(path))).
+                            count() > 0;
+                }).
+                // Keep capability that can be reached from the provided scope
+                filter((registration) -> hasCapability(registration.getCapabilityName(), dependentScope)).
```
ctomc: ah, never mind, I see what is going on.

controller/src/main/java/org/jboss/as/controller/CapabilityRegistry.java
```diff
@@ -905,6 +909,45 @@ private SatisfactoryCapability findSatisfactoryCapability(String capabilityName,
         return null;
     }
 
+    @Override
+    public Set<String> suggestCapabilities(String referencedCapability,
```
ctomc: I would just rename this method to something else, suggest might be fine for CLI.
but for generic use it should be something like getProviderPoints() vs getPossibleProviderPoints() we already have.
or even getRegistrationPoints()

bstansberry:
I'm quite happy with the test. It's fine for a test in the controller module to have detailed knowledge of how the scopes work.

jfdenise:
@ctomc, thank you for your review. 
I renamed the operation getCapabilityNames, we are returning the dynamic part of the capability names the scope can have access to.

PR #1830, opened 2016-09-26
JBEAP-2240 upgrade MSC to 1.2.7.Final

istudens:
https://issues.jboss.org/browse/JBEAP-2240


bstansberry:
Please file a WFCORE and use it in the commit message.

This isn't EAP.

PR #1832, opened 2016-09-27
[WFCORE-1827] Add core-management subsystem

jmesnil:
Add the org.wildfly.extension.core-management extensions that provides
a core-management subsystem.
The subsystem is currently empty and not installed in the default
standalone and domain configurations.

JIRA: https://issues.jboss.org/browse/WFCORE-1827

### DIFF
--- a/core-feature-pack/src/main/resources/modules/system/layers/base/org/wildfly/extension/core-management/main/module.xml
+++ b/core-feature-pack/src/main/resources/modules/system/layers/base/org/wildfly/extension/core-management/main/module.xml
@@ -1,0 +1,44 @@
+<?xml version="1.0" encoding="UTF-8"?>
+
+<!--
+  ~ /*
+  ~ * JBoss, Home of Professional Open Source.
+  ~ * Copyright 2016, Red Hat, Inc., and individual contributors
+  ~ * as indicated by the @author tags. See the copyright.txt file in the
+  ~ * distribution for a full listing of individual contributors.
+  ~ *
+  ~ * This is free software; you can redistribute it and/or modify it
+  ~ * under the terms of the GNU Lesser General Public License as
+  ~ * published by the Free Software Foundation; either version 2.1 of
+  ~ * the License, or (at your option) any later version.
+  ~ *
+  ~ * This software is distributed in the hope that it will be useful,
+  ~ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+  ~ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the GNU
+  ~ * Lesser General Public License for more details.
+  ~ *
+  ~ * You should have received a copy of the GNU Lesser General Public
+  ~ * License along with this software; if not, write to the Free
+  ~ * Software Foundation, Inc., 51 Franklin St, Fifth Floor, Boston, MA
+  ~ * 02110-1301 USA, or see the FSF site: http://www.fsf.org.
+  ~ */
+  -->
+
+<module xmlns="urn:jboss:module:1.5" name="org.wildfly.extension.core-management">
+    <properties>
+        <property name="jboss.api" value="private"/>
+    </properties>
+
+    <resources>
+        <artifact name="${org.wildfly.core:wildfly-core-management}"/>
+    </resources>
+
+    <dependencies>
+        <module name="javax.api"/>
+        <module name="org.jboss.as.controller"/>
+        <module name="org.jboss.logging"/>
+        <module name="org.jboss.modules"/>
+        <module name="org.jboss.msc"/>
+        <module name="org.jboss.staxmapper"/>
+    </dependencies>
+</module>
--- a/core-management/src/main/java/org/wildfly/extension/core/management/CoreManagementSubsystemParser_1_0.java
+++ b/core-management/src/main/java/org/wildfly/extension/core/management/CoreManagementSubsystemParser_1_0.java
@@ -1,0 +1,58 @@
+/*
+ * JBoss, Home of Professional Open Source.
+ * Copyright 2016, Red Hat, Inc., and individual contributors
+ * as indicated by the @author tags. See the copyright.txt file in the
+ * distribution for a full listing of individual contributors.
+ *
+ * This is free software; you can redistribute it and/or modify it
+ * under the terms of the GNU Lesser General Public License as
+ * published by the Free Software Foundation; either version 2.1 of
+ * the License, or (at your option) any later version.
+ *
+ * This software is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the GNU
+ * Lesser General Public License for more details.
+ *
+ * You should have received a copy of the GNU Lesser General Public
+ * License along with this software; if not, write to the Free
+ * Software Foundation, Inc., 51 Franklin St, Fifth Floor, Boston, MA
+ * 02110-1301 USA, or see the FSF site: http://www.fsf.org.
+ */
+
+package org.wildfly.extension.core.management;
+
+import static org.jboss.as.controller.PersistentResourceXMLDescription.builder;
+
+import org.jboss.as.controller.PersistentResourceXMLDescription;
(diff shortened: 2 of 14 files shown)


core-management/src/main/java/org/wildfly/extension/core/management/CoreManagementSubsystemParser_1_0.java
```diff
@@ -0,0 +1,77 @@
...
+ * This ensures that if the resource definitions change in later version (e.g. a new attribute is added),
+ * this will have no impact on parsing this specific version of the subsystem.
+ *
+ * @author <a href="http://jmesnil.net/">Jeff Mesnil</a> (c) 2016 Red Hat inc.
+ */
+public class CoreManagementSubsystemParser_1_0 implements XMLStreamConstants, XMLElementReader<List<ModelNode>>, XMLElementWriter<SubsystemMarshallingContext> {
+
+    static final String NAMESPACE = "urn:jboss:domain:core-management:1.0";
+
+    protected static final CoreManagementSubsystemParser_1_0 INSTANCE = new CoreManagementSubsystemParser_1_0();
```
darranl: Brian knows the details as well but recent experience has showed us that where we build up a resource model and parser entirely from static instances we loose a very powerful aspect of the management model where we can also maintain runtime state independently of services.

For other subsystems this maybe is more of a hypothetical discussion and they can rely completely on services but anything related to management is very :chicken: and :egg:y - as an example for some security related services we have needed to start up relatively unsecured and as the management model is built up we provide new implementation of specific SPIs.

With static references to resources there are still opportunities to maintain some state but the recent case where this broke down was a server reload as the instances were not recreated so state was maintained between server reloads.

jmesnil: I see your point and I agree that references to the management model might not be static if we want to maintain some flexibility.
However, I don't think that it applies to the parser, right? We provide a XSD that should validate the XML representation of the management model. I see this mapping as being static (the XSD won't change)

darranl: +1 Having a static parser is completely fine as that is supposed to be independent - the problem however is that is requires references to actual resource definition instances so the static parser causes something else to be static where we may want more flexibility.

core-feature-pack/src/main/resources/modules/system/layers/base/org/wildfly/extension/core-management/main/module.xml
```diff
@@ -0,0 +1,44 @@
...
+  ~ * Lesser General Public License for more details.
+  ~ *
+  ~ * You should have received a copy of the GNU Lesser General Public
+  ~ * License along with this software; if not, write to the Free
+  ~ * Software Foundation, Inc., 51 Franklin St, Fifth Floor, Boston, MA
+  ~ * 02110-1301 USA, or see the FSF site: http://www.fsf.org.
+  ~ */
+  -->
+
+<module xmlns="urn:jboss:module:1.3" name="org.wildfly.extension.core-management">
```
ctomc: 1.5 please

core-management/src/main/java/org/wildfly/extension/core/management/CoreManagementSubsystemParser_1_0.java
```diff
@@ -0,0 +1,77 @@
...
+/**
+ * Parser and Marshaller for core-management's {@link #NAMESPACE}.
+ *
+ * <em>All resources and attributes must be listed explicitly and not through any collections.</em>
+ * This ensures that if the resource definitions change in later version (e.g. a new attribute is added),
+ * this will have no impact on parsing this specific version of the subsystem.
+ *
+ * @author <a href="http://jmesnil.net/">Jeff Mesnil</a> (c) 2016 Red Hat inc.
+ */
+public class CoreManagementSubsystemParser_1_0 implements XMLStreamConstants, XMLElementReader<List<ModelNode>>, XMLElementWriter<SubsystemMarshallingContext> {
```
ctomc: extends PersistentResourceXMLParser instead of all implements stuff

core-management/src/main/java/org/wildfly/extension/core/management/CoreManagementSubsystemParser_1_0.java
```diff
@@ -0,0 +1,77 @@
...
+    static {
+        xmlDescription = builder(CoreManagementRootResourceDefinition.INSTANCE)
+                .build();
+    }
+
+    private CoreManagementSubsystemParser_1_0() {
+    }
+
+    @Override
+    public void writeContent(XMLExtendedStreamWriter writer, SubsystemMarshallingContext context) throws XMLStreamException {
```
ctomc: when you extend PersistentResourceXMLParser this is already done for you. 
see IOSubsystemParser_1_1 for example

core-management/src/test/resources/logging.properties
```diff
@@ -0,0 +1,53 @@
...
+# Lesser General Public License for more details.
+#
+# You should have received a copy of the GNU Lesser General Public
+# License along with this software; if not, write to the Free
+# Software Foundation, Inc., 51 Franklin St, Fifth Floor, Boston, MA
+# 02110-1301 USA, or see the FSF site: http://www.fsf.org.
+#
+
+# Additional logger names to configure (root logger is always configured)
+loggers=javax.security.sasl,org.xnio.ssl,org.xnio,com.google.inject.internal
```
ctomc: do we really need logging.properties here? all other subsystems are fine?

core-management/src/main/java/org/wildfly/extension/core/management/CoreManagementSubsystemParser_1_0.java
jmesnil: you are right. https://issues.jboss.org/browse/WFCORE-1831 will address this (so we no longer require static reference to ResourceDefinition to build the parser)

ctomc: Jira https://issues.jboss.org/browse/WFCORE-1831
on top of this we need to fix also parser registration so we wont have all parsers loaded in memory even if they are not needed

core-management/src/main/java/org/wildfly/extension/core/management/CoreManagementSubsystemParser_1_0.java
jmesnil: fixed

core-management/src/main/java/org/wildfly/extension/core/management/CoreManagementSubsystemParser_1_0.java
jmesnil: fixed

core-management/src/test/resources/logging.properties
jmesnil: fxied. it was not needed (a bad leftover from a copy/paste from another subsystem)

bstansberry:
@jmesnil Is your goal here to have a shared branch for coordinating later changes done by multiple people? Or just to get immediate feedback?

This all looks fine, but I'm reluctant to merge stuff into master that doesn't do anything useful until later work is done.

core-management/src/main/java/org/wildfly/extension/core/management/CoreManagementSubsystemParser_1_0.java
jmesnil: @darranl that's fixed. I've removed the static instance of ResourceDefinition.

jmesnil:
@bstansberry I wanted to get a commit/PR ready to be merged that Darran could use to add any management attribute related to Elytron attribute.

This PR can be put on hold as it as been reviewed and all feedback from Darran and Tomaz has been addressed. The actual commit may be merged as part of Elytron's integration or my work for WFCORE-1157, whichever comes first.

kabir:
Adding hold until someone needs it

bstansberry:
Let's talk about this on today's bluejeans.

darranl:
I will review if it makes sense for our resources to go in within a new subsystem at this point.

If we do end up with multiple groups building on this we will need to ensure the SHA is frozen and never rebased.

bstansberry:
Superceded by #1932 which picks up this commit.

PR #1833, opened 2016-09-27
[WFCORE-1789]: PreparedResponseTestCase from wf-core-eap fails intermittently.

ehsavoie:
Using the blocker extension instead of some deployment for a more prof tested code.

Jira: https://issues.jboss.org/browse/WFCORE-1789


ehsavoie:
retest this please

testsuite/shared/src/main/java/org/jboss/as/test/integration/management/extension/blocker/BlockerExtension.java
```diff
@@ -301,17 +308,23 @@ public void start(final StartContext context) throws StartException {
...
         }
 
         @Override
         public void stop(final StopContext context) {
             if (!blockStart) {
-                // Not yet used
-                throw new UnsupportedOperationException();
+                try {
+                    synchronized (waitObject) {
+                        log.info("BlockService blocking in start");
```
bstansberry: s/start/stop

testsuite/shared/src/main/java/org/jboss/as/test/integration/management/extension/blocker/BlockerExtension.java
```diff
@@ -212,6 +209,16 @@ public void handleResult(OperationContext.ResultAction resultAction, OperationCo
                         }, OperationContext.Stage.RUNTIME);
                         break;
                     }
+                    case SERVICE_STOP: {
```
bstansberry: Please add a big comment here explaining that this usage is not part of the main purpose of BlockerExtension and instead is to satisfy a different PreparedResponseTestCase. Otherwise it will be quite confusing later on if we ever try and get the original intent of SERVICE_STOP to work.

testsuite/manualmode/src/test/java/org/wildfly/core/test/standalone/mgmt/PreparedResponseTestCase.java
```diff
@@ -20,104 +20,100 @@
...
+                Thread.sleep(FREQUENCY);
+                try {
+                    controller.getClient().isServerInRunningState();
+                } catch (RuntimeException ex) {
+                    break;
+                }
             }
+            Assert.assertFalse(System.currentTimeMillis() < timeout);
+            Thread.sleep(2 * FREQUENCY);
+            Assert.assertTrue(managementClient.isServerInRunningState());
```
bstansberry: This seems too fragile. It wouldn't be an odd thing if this took > 1 second. I think it's valid to loop for quite a long time (say 20 secs) intermittently testing for reload completion. The RuntimeException at L103 tells you the reload has progressed, so AIUI it's not relevant to this test how long it takes from there to complete.

testsuite/manualmode/src/test/java/org/wildfly/core/test/standalone/mgmt/PreparedResponseTestCase.java
```diff
@@ -20,104 +20,100 @@
...
  */
 @RunWith(WildflyTestRunner.class)
 @ServerControl(manual = true)
-@Ignore("WFCORE-1789")
 public class PreparedResponseTestCase {
 
     public static Logger LOGGER = Logger.getLogger(PreparedResponseTestCase.class);
-
-    public static final String SLOW_STOP_JAR = "slow-stop.jar";
     private static final long FREQUENCY = TimeoutUtil.adjust(500);
```
bstansberry: Why not 50? If the polling this controls detects what it wants quickly, all the better.

testsuite/manualmode/src/test/java/org/wildfly/core/test/standalone/mgmt/PreparedResponseTestCase.java
```diff
@@ -20,104 +20,108 @@
...
-        long timeout = SHUTDOWN_WAITING_TIME + System.currentTimeMillis();
-        mgmtClient.executeForResult(Operations.createOperation(RELOAD, new ModelNode().setEmptyList()));
-        while (System.currentTimeMillis() < timeout) {
-            Thread.sleep(FREQUENCY);
-            try {
-                controller.getClient().isServerInRunningState();
-            } catch (RuntimeException ex) {
-                break;
+        try (ManagementClient managementClient = getManagementClient()) {
+            block(managementClient);
```
ctomc: why not just managementClient.reload();

testsuite/manualmode/src/test/java/org/wildfly/core/test/standalone/mgmt/PreparedResponseTestCase.java
```diff
@@ -20,104 +20,108 @@
...
-        Thread.sleep(2 * FREQUENCY);
-        Assert.assertTrue(mgmtClient.isServerInRunningState());
     }
 
     @Test
     public void shutdownServer() throws Exception {
         try (ManagementClient managementClient = getManagementClient()) {
+            block(managementClient);
             long timeout = SHUTDOWN_WAITING_TIME + System.currentTimeMillis();
             managementClient.executeForResult(Operations.createOperation(SHUTDOWN, new ModelNode().setEmptyList()));
```
ctomc: controller.stop();

PR #1835, opened 2016-09-27
Fix for WFCORE-1826. CLI value-type completion fixes.

jfdenise:
Took the opportunity of this change to fixe various issues:
- Nested lists
- Completed objects (propose })
- Completed values (propose ,)
- Resist to some user invalid syntaxes
- Add new unit tests

Side effect is to fix WFCORE-1825


bstansberry:
@aloubyansky please approve. Looked good to me, 10% because I understood the details and 90% because there are big fast tests!

aloubyansky:
Yes, looks great. Thanks!

PR #1836, opened 2016-09-27
WFCORE-1815 It's not possible to re-add extension

ctomc:
jira: https://issues.jboss.org/browse/WFCORE-1815


controller/src/main/java/org/jboss/as/controller/extension/ExtensionRegistry.java
```diff
@@ -341,6 +341,7 @@ public void removeExtension(Resource rootResource, String moduleName, Management
                     profileReg.unregisterSubModel(PathElement.pathElement(ModelDescriptionConstants.SUBSYSTEM, subsystem));
                     if (deploymentsReg != null) {
                         deploymentsReg.unregisterSubModel(PathElement.pathElement(ModelDescriptionConstants.SUBSYSTEM, subsystem));
+                        deploymentsReg.unregisterSubModel(PathElement.pathElement(ModelDescriptionConstants.SUBDEPLOYMENT, subsystem));
```
bstansberry: Interesting. The MRR used for deployments is meant to automatically do this kind of this. See DeploymentManagementResourceRegistration.unregisterSubModel().

Ah, ok, as part of my commit 84450d69ceb552d5c4e33635032dc998c21bc606 I broke this. DeploymentManagementResourceRegistration is now just a special-purpose wrapper used by ExtensionContextImpl, so its logic has no effect after the extension is initialized.

👍

PR #1837, opened 2016-09-27
fix AttachmentTestCase leaving a deployment around causing BasicOperationsUnitTestCase#testPathInfo to measure size of the content dir > 0b

luck3y:
@jfdenise, fyi, I think this only shows up if the test order is randomly different.

### DIFF
--- a/testsuite/standalone/src/test/java/org/wildfly/core/test/standalone/mgmt/BasicOperationsUnitTestCase.java
+++ b/testsuite/standalone/src/test/java/org/wildfly/core/test/standalone/mgmt/BasicOperationsUnitTestCase.java
@@ -188,7 +188,7 @@ public class BasicOperationsUnitTestCase {
         result = managementClient.getControllerClient().execute(operation);
         assertTrue(result.toJSONString(true), Operations.isSuccessfulOutcome(result));
         assertTrue(result.toJSONString(true), result.hasDefined(RESULT));
-        assertTrue(result.toJSONString(true), Operations.readResult(result).get("content-dir").get("used-space").asDouble() == 0.0D);
+        assertTrue(result.toJSONString(true), Operations.readResult(result).get("content-dir").get("used-space").asDouble() >= 0.0D);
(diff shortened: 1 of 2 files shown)


bstansberry:
This kind of cleanup is good, so the PR is fine with me if ok with Jeff. But BasicOperationsUnitTestCase failing any time there is data in the content dir seems overly fragile.

luck3y:
We could clean that up in @Before in BOUTC maybe, as long as those aren't sharing configs (Which I bet they aren't.)

Yah, I'll add that too quick, might save someone some time later.

jfdenise:
Thank you for the cleanup, test change is fine.
JF

testsuite/standalone/src/test/java/org/wildfly/core/test/standalone/mgmt/BasicOperationsUnitTestCase.java
```diff
@@ -112,6 +115,38 @@
...
+
+            for (File f : file.listFiles()) {
+                if (f.isFile()) {
+                    f.delete();
+                } else if (f.isDirectory()) {
+                    FileUtils.deleteDirectory(f);
+                }
+            }
+        }
+    }
```
bstansberry: This is likely just as fragile. If a test has left content in this dir it probably has left config referencing it in the model and once this runs any subsequent reload will blow up.

I don't see a significant benefit to this test asserting that this dir has a size of zero. At least not any benefit that justifies complexity.

luck3y: Ah, OK, I'd interpreted this as something the test really wanted to verify. I'll remove that and my last change and squash.

luck3y:
retest this please

luck3y:
retest this please

luck3y:
retest this please

bstansberry:
@luck3y needs a rebase

luck3y:
@bstansberry Should be good to go now.

bstansberry:
retest this please

PR #1841, opened 2016-09-29
WFCORE-1829 Allow server to start in suspended mode

stuartwdouglas:
https://issues.jboss.org/browse/WFCORE-1829


stuartwdouglas:
Retest this please!

rhatlapa:
I've taken a look at the changes and tried it out. I can see there are `:reload` operations containing now `suspend` option. Still I am missing this option for CLI `reload` operation, I believe the suspend option should be available for the CLI `reload` operation.

rhatlapa:
@stuartwdouglas When taking a look, I've noticed this behavior: When having server started with `--suspend` and then call `:reload(suspend=false)`. The server stays suspended after reload instead of being then fully started.

stuartwdouglas:
I have pushed some changes so that if a server was started suspended it will not always be suspended on reload

kabir:
Included by https://github.com/wildfly/wildfly-core/pull/1984

PR #1842, opened 2016-09-30
Fix for WFCORE-1843. module command to export dependencies.

jfdenise:
New option for the module command and associated unit tests.

### DIFF
--- a/cli/src/main/java/org/jboss/as/cli/handlers/module/ASModuleHandler.java
+++ b/cli/src/main/java/org/jboss/as/cli/handlers/module/ASModuleHandler.java
@@ -1,6 +1,6 @@
 /*
  * JBoss, Home of Professional Open Source.
- * Copyright 2015, Red Hat, Inc., and individual contributors
+ * Copyright 2016, Red Hat, Inc., and individual contributors
  * as indicated by the @author tags. See the copyright.txt file in the
  * distribution for a full listing of individual contributors.
  *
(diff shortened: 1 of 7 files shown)


cli/src/main/resources/help/module.txt
```diff
@@ -85,6 +86,14 @@ ARGUMENTS
                    is generated, i.e. when the --module-xml argument isn't
                    specified.
 
+ --export-dependencies  - (used with add, optional) a comma-separated list of module
+                   names the module being added depends on and are exported.
```
bstansberry: I would say

"names the module being added depends on and exports."

cli/src/main/java/org/jboss/as/cli/handlers/module/ASModuleHandler.java
```diff
@@ -396,13 +416,30 @@ protected void addModule(CommandContext ctx, final ParsedCommandLine parsedCmd)
...
-        if(config != null) {
+        if (config != null) {
+            Set<String> modules = new HashSet<>();
             final String dependenciesStr = dependencies.getValue(parsedCmd);
             if(dependenciesStr != null) {
                 final String[] depsArr = dependenciesStr.split(",+");
                 for(String dep : depsArr) {
                     // TODO validate dependencies
                     config.addDependency(new ModuleDependency(dep));
+                    modules.add(dep);
```
bstansberry: Should this have the same kind of duplication check you do at L436 below?

bstansberry:
Couple minor points aside, this looks good. But I'll delay merging it until QE is done with their analysis step. Please ping me when that is done.

cli/src/main/java/org/jboss/as/cli/handlers/module/ASModuleHandler.java
jfdenise: I didn't wanted to change the existing behavior. We could have customer scripts that wrongly add dependencies twice.

jfdenise:
@bstansberry, QE is done.
JF

PR #1849, opened 2016-10-05
service.bat - fix quotes around domain BASE path

grantwohl:
This was causing an extra pair of double quotes to appear in the call to wildfly-service.exe, breaking the installation for domain mode.


ctomc:
this is ok to test

bstansberry:
@ctomc please approve

PR #1850, opened 2016-10-06
[WFCORE-296]: Switch URI scheme from remoting:// http-remoting:// https-remoting:// to remote:// remote+http:// and remote+https://

ehsavoie:
Changing also the URL format for JMX and the default protocol is schemas.

Jira: https://issues.jboss.org/browse/WFCORE-296


testsuite/domain/src/test/java/org/jboss/as/test/integration/respawn/RespawnHttpTestCase.java
```diff
@@ -73,7 +73,8 @@
 import static org.jboss.as.controller.descriptions.ModelDescriptionConstants.SHUTDOWN;
 import static org.jboss.as.controller.descriptions.ModelDescriptionConstants.SUCCESS;
 
-import org.jboss.as.remoting.Protocol;
+import org.junit.Ignore;
```
bstansberry: Why do we need to ignore these tests?

ehsavoie: this is a leftover, please ignore ;)

PR #1853, opened 2016-10-07
Various JDK9 fixes

ropalka:


bstansberry:
@jamezp WDYT about the last commit? One question I had is whether ignoring the invalid option is the correct behavior, as opposed to failing.

@ropalka I believe the last commit should be 2 commits. The launcher module stuff seems generally ok, i.e. it's behavior that may not be changed after this is merged. But the ManagedServerBootCmdFactory change needs to be associated with a blocker JIRA to get it done in the long-term supportable fashion. We can't ship a CR with that in it, unless it turns out that's the permanent solution.

ropalka:
I'd like to have it in one commit, please, because it's just a temporary solution @bstansberry .
See https://issues.jboss.org/browse/WFCORE-1859 issue description for more information.

host-controller/src/main/java/org/jboss/as/host/controller/DomainModelControllerService.java
```diff
@@ -1235,14 +1235,12 @@ private static AbstractVaultReader loadVaultReaderService() {
...
+        try {
+            while (it.hasNext()) {
                 return it.next();
-            } catch (ServiceConfigurationError e) {
-                if (sce == null) {
-                    sce = e;
-                }
             }
+        } catch (ServiceConfigurationError e) {
+            sce = e;
```
jamezp: Was there a reason to change this? It looks like previously it would only log the first error and now it will log the last error.

launcher/src/main/java/org/wildfly/core/launcher/CliCommandBuilder.java
```diff
@@ -376,8 +376,17 @@ public CliCommandBuilder addJavaOption(final String jvmArg) {
...
         if (javaOpts != null) {
-            for (String javaOpt : javaOpts) {
-                addJavaOption(javaOpt);
+            for (int i = 0; i < javaOpts.length; i++) {
+                if (Arguments.isModuleSystemArgument(javaOpts[i])) {
+                    // handle java module system options
+                    if (i < javaOpts.length - 1) {
+                        this.javaOpts.add(Arguments.JavaModuleSystemArgument.of(javaOpts[i], javaOpts[++i]));
+                    } else {
+                        // ignore corrupted module system argument
```
jamezp: I don't think we should ignore these. We should just add the argument to the command and allow it to fail when executed.

launcher/src/main/java/org/wildfly/core/launcher/CliCommandBuilder.java
```diff
@@ -392,8 +401,18 @@ public CliCommandBuilder addJavaOptions(final String... javaOpts) {
...
-            for (String javaOpt : javaOpts) {
-                addJavaOption(javaOpt);
+            final Iterator<String> i = javaOpts.iterator();
+            for (String javaOpt = i.next(); i.hasNext(); javaOpt = i.next()) {
+                if (Arguments.isModuleSystemArgument(javaOpt)) {
+                    // handle java module system options
+                    if (i.hasNext()) {
+                        this.javaOpts.add(Arguments.JavaModuleSystemArgument.of(javaOpt, i.next()));
+                    } else {
+                        // ignore corrupted module system argument
```
jamezp: Same as above, we should just add the arguments and allow the command to fail when executed.

launcher/src/main/java/org/wildfly/core/launcher/DomainCommandBuilder.java
```diff
@@ -522,8 +523,17 @@ public DomainCommandBuilder addHostControllerJavaOption(final String arg) {
...
         if (args != null) {
-            for (String arg : args) {
-                addHostControllerJavaOption(arg);
+            for (int i = 0; i < args.length; i++) {
+                if (Arguments.isModuleSystemArgument(args[i])) {
+                    // handle java module system options
+                    if (i < args.length - 1) {
+                        this.hostControllerJavaOpts.add(Arguments.JavaModuleSystemArgument.of(args[i], args[++i]));
+                    } else {
+                        // ignore corrupted module system argument
```
jamezp: Same as above, we should just add the arguments and allow the command to fail when executed.

launcher/src/main/java/org/wildfly/core/launcher/DomainCommandBuilder.java
```diff
@@ -538,8 +548,18 @@ public DomainCommandBuilder addHostControllerJavaOptions(final String... args) {
...
-            for (String arg : args) {
-                addHostControllerJavaOption(arg);
+            final Iterator<String> i = args.iterator();
+            for (String arg = i.next(); i.hasNext(); arg = i.next()) {
+                if (Arguments.isModuleSystemArgument(arg)) {
+                    // handle java module system options
+                    if (i.hasNext()) {
+                        this.hostControllerJavaOpts.add(Arguments.JavaModuleSystemArgument.of(arg, i.next()));
+                    } else {
+                        // ignore corrupted module system argument
```
jamezp: Same as above, we should just add the arguments and allow the command to fail when executed.

launcher/src/main/java/org/wildfly/core/launcher/DomainCommandBuilder.java
```diff
@@ -610,8 +630,17 @@ public DomainCommandBuilder addProcessControllerJavaOption(final String arg) {
...
         if (args != null) {
-            for (String arg : args) {
-                addProcessControllerJavaOption(arg);
+            for (int i = 0; i < args.length; i++) {
+                if (Arguments.isModuleSystemArgument(args[i])) {
+                    // handle java module system options
+                    if (i < args.length - 1) {
+                        this.processControllerJavaOpts.add(Arguments.JavaModuleSystemArgument.of(args[i], args[++i]));
+                    } else {
+                        // ignore corrupted module system argument
```
jamezp: Same as above, we should just add the arguments and allow the command to fail when executed.

launcher/src/main/java/org/wildfly/core/launcher/DomainCommandBuilder.java
```diff
@@ -626,8 +655,18 @@ public DomainCommandBuilder addProcessControllerJavaOptions(final String... args
...
-            for (String arg : args) {
-                addProcessControllerJavaOption(arg);
+            final Iterator<String> i = args.iterator();
+            for (String arg = i.next(); i.hasNext(); arg = i.next()) {
+                if (Arguments.isModuleSystemArgument(arg)) {
+                    // handle java module system options
+                    if (i.hasNext()) {
+                        this.processControllerJavaOpts.add(Arguments.JavaModuleSystemArgument.of(arg, i.next()));
+                    } else {
+                        // ignore corrupted module system argument
```
jamezp: Same as above, we should just add the arguments and allow the command to fail when executed.

launcher/src/main/java/org/wildfly/core/launcher/StandaloneCommandBuilder.java
```diff
@@ -137,8 +138,17 @@ public StandaloneCommandBuilder addJavaOption(final String jvmArg) {
...
         if (javaOpts != null) {
-            for (String javaOpt : javaOpts) {
-                addJavaOption(javaOpt);
+            for (int i = 0; i < javaOpts.length; i++) {
+                if (Arguments.isModuleSystemArgument(javaOpts[i])) {
+                    // handle java module system options
+                    if (i < javaOpts.length - 1) {
+                        this.javaOpts.add(Arguments.JavaModuleSystemArgument.of(javaOpts[i], javaOpts[++i]));
+                    } else {
+                        // ignore corrupted module system argument
```
jamezp: Same as above, we should just add the arguments and allow the command to fail when executed.

launcher/src/main/java/org/wildfly/core/launcher/StandaloneCommandBuilder.java
```diff
@@ -153,8 +163,18 @@ public StandaloneCommandBuilder addJavaOptions(final String... javaOpts) {
...
-            for (String javaOpt : javaOpts) {
-                addJavaOption(javaOpt);
+            final Iterator<String> i = javaOpts.iterator();
+            for (String javaOpt = i.next(); i.hasNext(); javaOpt = i.next()) {
+                if (Arguments.isModuleSystemArgument(javaOpt)) {
+                    // handle java module system options
+                    if (i.hasNext()) {
+                        this.javaOpts.add(Arguments.JavaModuleSystemArgument.of(javaOpt, i.next()));
+                    } else {
+                        // ignore corrupted module system argument
```
jamezp: Same as above, we should just add the arguments and allow the command to fail when executed.

server/src/main/java/org/jboss/as/server/ApplicationServerService.java
```diff
@@ -227,14 +227,12 @@ private static AbstractVaultReader loadVaultReaderService() {
...
+        try {
+            while (it.hasNext()) {
                 return it.next();
-            } catch (ServiceConfigurationError e) {
-                if (sce == null) {
-                    sce = e;
-                }
             }
+        } catch (ServiceConfigurationError e) {
+            sce = e;
```
jamezp: Same question here as to if this should be done. This will log the last error instead of the first now.

host-controller/src/main/java/org/jboss/as/host/controller/DomainModelControllerService.java
ropalka: It still logs the first error it detects

jamezp: Oops yes it does. Sorry about that. I missed that the try/catch moved.

host-controller/src/main/java/org/jboss/as/host/controller/ManagedServerBootCmdFactory.java
```diff
@@ -239,6 +242,8 @@ public HostControllerEnvironment getHostControllerEnvironment() {
 
         command.add("-D[" + ManagedServer.getServerProcessName(serverName) + "]");
 
+        propagateJavaModuleParams(command);
```
bstansberry: Please start a wildfly-dev list about how we are going to ultimately deal with this.

I don't mind putting in a breaking change like this as a workaround while we actually implement the final solution, but I don't think it's good to break things when we have no idea what the correct solution will be. That's just asking for a future crisis.

We want the common case where the servers are on the same jre as the host to be easy, but we also need to support servers running on a different jre.

ropalka: Sorry for delay to reply to your comment.
I'll exclude these 'unsafe' changes from my PR @bstansberry .
Until JDK9 is not in final state it is hard to say what final solution will be.
According to latest discussions on jigsaw-dev mailing list it seems 
this change will not be necessary at the end. But things are still evolving there ...

pom.xml
```diff
@@ -95,7 +95,7 @@
         <version.org.codehaus.woodstox.woodstox-core>5.0.3</version.org.codehaus.woodstox.woodstox-core>
         <version.org.fusesource.jansi>1.11</version.org.fusesource.jansi>
         <version.org.jboss.aesh>0.66.11</version.org.jboss.aesh>
-        <version.org.jboss.byteman>3.0.3</version.org.jboss.byteman>
+        <version.org.jboss.byteman>4.0.0-ALPHA</version.org.jboss.byteman>
```
ropalka: I have no idea but I guess once JDK9 is final? Try ask Andrew Dinn.
If you see this commit problematic I can exclude it from change set too.
There's already Byteman BETA0 available, but it works just on JDK9 based on jake branch.
JDK9 mainline isn't supported yet. ALPHA works on both JDK9 versions.

bstansberry: @ropalka WildFly Core 3.0.0.CR1 is probably going to be out early next year and can't have dependencies that are not .Final (or at least .CR with a clear schedule to be .Final in time for core's .Final)

I think this is going to be a general issue with this JDK 9 work. Our schedule requires releases before JDK 9 so any components that are making changes that target 9 may not be ready in time.

ropalka:
@bstansberry Yes, I see it as a problem ATM too (wrong timing). There are 2 options we have I'd say:
 * either I will maintain our WildFly-Core and WildFly JDK9 branches without sending PRs till January
 * or we will create separate JDK9 branches for both WildFly-Core and WildFly github repos
I don't care which way we'll handle it. We just need to inform others where these branches reside.

pom.xml
ctomc: I don't think byteman is problematic as it is only test dependency and as such should be fine to use in our build / testsuite.
Non tests dependencies are different story.

pom.xml
```diff
@@ -1320,8 +1321,21 @@
 
             <dependency>
                 <groupId>org.mockito</groupId>
-                <artifactId>mockito-all</artifactId>
+                <artifactId>mockito-core</artifactId>
```
ctomc: re-add mockito-all just here, as wildfly full needs it and  build fails if not defined here.
Keep both defined here for now, but add comment on why the --all is still here.

ropalka:
@bstansberry I reworked this PR to include only non conflicting fixes everything else will be kept in my JDK9 topic branch

PR #1856, opened 2016-10-11
[WFCORE-1863] Make CredentialReference thread safe

pskopek:
making CredentialReference thread safe as per comment in https://github.com/wildfly/wildfly-core/pull/1816
https://issues.jboss.org/browse/WFCORE-1863

### DIFF
--- a/controller/src/main/java/org/jboss/as/controller/security/CredentialReference.java
+++ b/controller/src/main/java/org/jboss/as/controller/security/CredentialReference.java
@@ -84,12 +84,12 @@ public final class CredentialReference implements Destroyable {
     static final SimpleAttributeDefinition credentialTypeAttribute;
     static final SimpleAttributeDefinition clearTextAttribute;
 
-    static ObjectTypeAttributeDefinition credentialReferenceAttributeDefinition;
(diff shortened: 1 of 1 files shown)


controller/src/main/java/org/jboss/as/controller/security/CredentialReference.java
```diff
@@ -164,7 +163,6 @@ public void destroy() {
             for (int i = 0; i < secret.length; i++) {
                 secret[i] = 0;
             }
-            secret = null;
```
bstansberry: This breaks isDestroyed().

pskopek: Thanks for spotting it @bstansberry !

pskopek:
New changes pushed.

dmlloyd:
Actually simply changing the "secret" field to be volatile should be sufficient to make this class thread-safe.  All the other stuff just adds bulk and complexity that is unwarranted AFAICT.

bstansberry:
@pskopek any thoughts re: dmlloyd's comment?

pskopek:
I have made the changes dmlloyd suggested.

dmlloyd:
Looks fine to me now.

PR #1858, opened 2016-10-13
[WFCORE-1864] Trim dependcies names in module add command

spyrkob:
JIRA: https://issues.jboss.org/browse/WFCORE-1864

### DIFF
--- a/testsuite/standalone/src/test/java/org/jboss/as/test/integration/management/cli/ModuleTestCase.java
+++ b/testsuite/standalone/src/test/java/org/jboss/as/test/integration/management/cli/ModuleTestCase.java
@@ -186,6 +186,20 @@ public class ModuleTestCase extends AbstractCliTestBase {
         testModuleAndSlotExist(false, slot, getModulePath());
     }
 
+    @Test
+    public void addModuleWithWhitespacesInDependencies() throws Exception {
+        // should trim the surrounding whitespaces
+        try {
+            cli.sendLine("module add --name=" + MODULE_NAME
+                    + " --resources=" + jarFile.getAbsolutePath()
+                    + " --dependencies=[ org.jboss.logging ]"
+                    + " --export-dependencies=[ org.jboss.logmanager ]");
+            testModuleFiles("main", getModulePath(), false, false, false, true, false);
+        } finally {
+            testRemove("main", false);
+        }
+    }
+
     private void testAddRemove(String slotName, boolean addModuleRootDir, boolean addResources,
                                boolean addAbsoluteResources, boolean addPropeties,
                                boolean addDependencies, boolean addMainClass) throws Exception {
--- a/cli/src/main/java/org/jboss/as/cli/handlers/module/ASModuleHandler.java
+++ b/cli/src/main/java/org/jboss/as/cli/handlers/module/ASModuleHandler.java
@@ -423,8 +423,9 @@ public class ASModuleHandler extends CommandHandlerWithHelp {
                 final String[] depsArr = dependenciesStr.split(",+");
                 for(String dep : depsArr) {
                     // TODO validate dependencies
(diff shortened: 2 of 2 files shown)


bstansberry:
@jfdenise please approve

jfdenise:
That is great. One thing to mention is that the PR (https://github.com/wildfly/wildfly-core/pull/1842) that should be merged soon introduce the same issue for exported dependencies.
@spyrkob, would you be ready to update your PR with a fix for exported dependencies too once this PR is merged?
Thanks.
JF

spyrkob:
Updated the PR to handle export-dependencies. @jfdenise could you verify please?

testsuite/standalone/src/test/java/org/jboss/as/test/integration/management/cli/ModuleTestCase.java
```diff
@@ -186,6 +186,20 @@ public void addRemoveModuleMetaInf() throws Exception {
         testModuleAndSlotExist(false, slot, getModulePath());
     }
 
+    @Test
+    public void addModuleWithWhitespacesInDependencies() throws Exception {
+        // should trim the surrounding whitespaces
+        try {
+            cli.sendLine("module add --name=" + MODULE_NAME
+                    + " --resources=" + jarFile.getAbsolutePath()
+                    + " --dependencies=[ org.jboss.logging ]"
```
jfdenise: using the '[' and ']' characters for dependencies seems not right. Same for export-dependencies.

spyrkob: I've seen this used in https://issues.jboss.org/browse/JBEAP-6282. I guess it's useful to visually separate the list of deps from rest of command for easier reading.

I thought the [] brackets are used for value lists (as in ArgumentWithListValue) in which case the use seems justified?

jfdenise: The list of modules is ',' separated, no need for '['. At least I never used it.

bstansberry: The current module.txt says:

```
NOTE: In non-interactive mode this list must be surrounded by square brackets, e.g. [dep1,dep2].
```

If that's no longer true, great! But we need to be sure that the [] syntax still works since it was documented.

jfdenise: I just checked, '[' and ']' are supported and needed to escape ',' when command is passed to CLI process:
jboss-cli.sh "module add --name=mod --resources=my.jar --dependencies=[ibm.jdk,javax.org]"

If not used, the ',' is interpreted as a command separator. So the '[' and ']' are only there to escape ','.

I don't think that we need them in this particular case. We could keep them but this is a bit strange to escape ',' in a context where we don't have to.

bstansberry: I commented above but don't have any particular opinion about using the brackets in this test, so no argument from me about removing them. It's good to have coverage somewhere that the [xxx,yyy] syntax parses correctly, but that very well may be covered elsewhere and even if it isn't a test specifically aimed at that syntax seems better.

jfdenise: @spyrkob seems that you can do what you want ;-). I would remove them, but this is a minor point. Feel free to keep them or not. 
If you keep them, you should possibly add a comment that say that they are not required in this case.

jfdenise: Ok, thanks. That is good to merge.

PR #1861, opened 2016-10-17
[WFCORE-1870][WFCORE-1871] Better handling of reload-required with ops that modify the DUP chain

bstansberry:


ehsavoie:
retest this please

PR #1864, opened 2016-10-18
WFCORE-1841 lazy subsystem parser support

ctomc:
- introduces parser providers (via staxmapper 1.3)
- migrates core subsystems to use it
- should reduce memory usage down a bit

### DIFF
--- a/controller/src/main/java/org/jboss/as/controller/PersistentResourceXMLDescription.java
+++ b/controller/src/main/java/org/jboss/as/controller/PersistentResourceXMLDescription.java
@@ -35,14 +35,14 @@ import org.jboss.staxmapper.XMLExtendedStreamWriter;
 public final class PersistentResourceXMLDescription {
 
     protected final PathElement pathElement;
-    protected final String xmlElementName;
-    protected final String xmlWrapperElement;
-    protected final LinkedHashMap<String, LinkedHashMap<String, AttributeDefinition>> attributesByGroup;
+    private final String xmlElementName;
+    private final String xmlWrapperElement;
+    private final LinkedHashMap<String, LinkedHashMap<String, AttributeDefinition>> attributesByGroup;
     protected final List<PersistentResourceXMLDescription> children;
-    protected final Map<String, AttributeDefinition> attributeElements = new HashMap<>();
-    protected final boolean useValueAsElementName;
-    protected final boolean noAddOperation;
-    protected final AdditionalOperationsGenerator additionalOperationsGenerator;
+    private final Map<String, AttributeDefinition> attributeElements = new HashMap<>();
+    private final boolean useValueAsElementName;
+    private final boolean noAddOperation;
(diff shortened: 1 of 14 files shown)


controller/src/main/java/org/jboss/as/controller/PersistentResourceXMLDescription.java
```diff
@@ -462,7 +463,8 @@ public static PersistentResourceXMLBuilder builder(final PathElement pathElement
         protected boolean noAddOperation;
         protected AdditionalOperationsGenerator additionalOperationsGenerator;
         protected final LinkedList<AttributeDefinition> attributeList = new LinkedList<>();
-        protected final List<PersistentResourceXMLBuilder> children = new ArrayList<>();
+        final List<PersistentResourceXMLBuilder> childrenBuilder = new ArrayList<>();
+        final List<PersistentResourceXMLDescription> children = new ArrayList<>();
```
bstansberry: This changes the visibility of 'children'.

It's too bad this class isn't final and all of these aren't just private. There are no subclasses in core or full.

controller/src/main/java/org/jboss/as/controller/PersistentResourceXMLDescription.java
```diff
@@ -462,7 +463,8 @@ public static PersistentResourceXMLBuilder builder(final PathElement pathElement
         protected boolean noAddOperation;
         protected AdditionalOperationsGenerator additionalOperationsGenerator;
         protected final LinkedList<AttributeDefinition> attributeList = new LinkedList<>();
-        protected final List<PersistentResourceXMLBuilder> children = new ArrayList<>();
+        final List<PersistentResourceXMLBuilder> childrenBuilder = new ArrayList<>();
```
bstansberry: Nit: 'childrenBuilders' is a better name as this is a collection.

controller/src/main/java/org/jboss/as/controller/parsing/ExtensionParsingContext.java
```diff
@@ -66,6 +67,9 @@
      */
     void setSubsystemXmlMapping(String subsystemName, String namespaceUri, XMLElementReader<List<ModelNode>> reader);
 
+
+    void setSubsystemXmlMapping(String subsystemName, String namespaceUri, Supplier<XMLElementReader<List<ModelNode>>> supplier);
```
bstansberry: Needs verbose javadoc.

ctomc: It was written, but lost in refactoring somewhere. 
fixed.

controller/src/main/java/org/jboss/as/controller/PersistentResourceXMLDescription.java
ctomc: fixed

bstansberry:
retest this please

PR #1865, opened 2016-10-19
WFCORE-773 ArrayIndexOutOfBoundsException occurs when a queue-length is 0 in th…

jurakp:
Set range validation to avoid ArrayIndexOfBoundsException.

https://issues.jboss.org/browse/WFCORE-773


ctomc:
this is ok to test

PR #1866, opened 2016-10-19
[WFCORE-1884] Discovery subsystem

dmlloyd:


dmlloyd:
I can't say for sure but I think the windows failures might be environmental.  Unless I messed up the configuration assembly thing, there should be no way I could have caused any of that stuff.

luck3y:
Retest this please

dmlloyd:
Retest this please

discovery/src/main/java/org/wildfly/extension/discovery/AggregateProviderAddHandler.java
```diff
@@ -0,0 +1,91 @@
...
+
+    static void modifyRegistrationModel(OperationContext context, ModelNode op) throws OperationFailedException {
+        WriteAttributeHandler.INSTANCE.execute(context, op);
+        context.addStep(op, AggregateProviderAddHandler::modifyRegistration, OperationContext.Stage.RUNTIME);
+    }
+
+    static void modifyRegistration(OperationContext context, ModelNode op) {
+        final MutableDiscoveryProvider mutableDiscoveryProvider = context.getCapabilityRuntimeAPI(DiscoveryExtension.DISCOVERY_PROVIDER_CAPABILITY, context.getCurrentAddressValue(), MutableDiscoveryProvider.class);
+        if (op.hasDefined(DiscoveryExtension.PROVIDERS)) {
+            final List<ModelNode> list = op.require(DiscoveryExtension.PROVIDERS).asList();
```
ctomc: List<String> list = AggregateProviderDefinition.PROVIDER_NAMES.unwrap(context, op);
or at least  AggregateProviderDefinition.PROVIDER_NAMES.resolveModelAttribute(context, op);

discovery/src/main/java/org/wildfly/extension/discovery/AggregateProviderDefinition.java
```diff
@@ -0,0 +1,63 @@
...
+import org.jboss.as.controller.StringListAttributeDefinition;
+import org.jboss.as.controller.registry.ManagementResourceRegistration;
+import org.jboss.as.controller.registry.OperationEntry;
+
+/**
+ * @author <a href="mailto:david.lloyd@redhat.com">David M. Lloyd</a>
+ */
+final class AggregateProviderDefinition extends SimpleResourceDefinition {
+
+    static final AttributeDefinition PROVIDER_NAMES = new StringListAttributeDefinition.Builder(DiscoveryExtension.PROVIDERS)
```
ctomc: static final StringListAttributeDefinition PROVIDER_NAMES =
this way you could use unwrap method.

discovery/src/main/java/org/wildfly/extension/discovery/AggregateProviderRemoveHandler.java
```diff
@@ -0,0 +1,37 @@
...
+ * You should have received a copy of the GNU Lesser General Public
+ * License along with this software; if not, write to the Free
+ * Software Foundation, Inc., 51 Franklin St, Fifth Floor, Boston, MA
+ * 02110-1301 USA, or see the FSF site: http://www.fsf.org.
+ */
+package org.wildfly.extension.discovery;
+
+import org.jboss.as.controller.AbstractRemoveStepHandler;
+
+final class AggregateProviderRemoveHandler extends AbstractRemoveStepHandler {
```
ctomc: why not just use ReloadRequiredRemoveStepHandler instead of having a no-op one

discovery/src/main/java/org/wildfly/extension/discovery/DiscoverySubsystemRemove.java
```diff
@@ -0,0 +1,37 @@
...
+
+package org.wildfly.extension.discovery;
+
+import org.jboss.as.controller.AbstractRemoveStepHandler;
+import org.jboss.as.controller.OperationStepHandler;
+
+/**
+ * @author <a href="mailto:david.lloyd@redhat.com">David M. Lloyd</a>
+ */
+final class DiscoverySubsystemRemove extends AbstractRemoveStepHandler {
```
ctomc: why no-op one?
noting to cleanup? And it is intended?

discovery/src/main/java/org/wildfly/extension/discovery/StaticProviderAddHandler.java
```diff
@@ -0,0 +1,117 @@
...
+    }
+
+    static void modifyRegistrationModel(OperationContext context, ModelNode op) throws OperationFailedException {
+        WriteAttributeHandler.INSTANCE.execute(context, op);
+        context.addStep(op, StaticProviderAddHandler::modifyRegistration, OperationContext.Stage.RUNTIME);
+    }
+
+    static void modifyRegistration(OperationContext context, ModelNode op) {
+        final MutableDiscoveryProvider mutableDiscoveryProvider = context.getCapabilityRuntimeAPI(DiscoveryExtension.DISCOVERY_PROVIDER_CAPABILITY, context.getCurrentAddressValue(), MutableDiscoveryProvider.class);
+        System.out.println("Operation is " + op);
```
ctomc: System.out :)

discovery/src/main/java/org/wildfly/extension/discovery/StaticProviderAddHandler.java
```diff
@@ -0,0 +1,117 @@
...
+        if (op.hasDefined(DiscoveryExtension.SERVICES)) {
+            final List<ModelNode> list = op.require(DiscoveryExtension.SERVICES).asList();
```
ctomc: at least AD.resolveModelAttribute should be used

discovery/src/main/java/org/wildfly/extension/discovery/AggregateProviderRemoveHandler.java
dmlloyd: Reload isn't required to remove these guys.

discovery/src/main/java/org/wildfly/extension/discovery/StaticProviderAddHandler.java
dmlloyd: Oops!

discovery/src/main/java/org/wildfly/extension/discovery/AggregateProviderAddHandler.java
dmlloyd: Fixed... I think!

discovery/src/main/java/org/wildfly/extension/discovery/StaticProviderAddHandler.java
dmlloyd: Fixed!

dmlloyd:
Is there something else I have to do to add to the main feature pack template?

discovery/src/main/java/org/wildfly/extension/discovery/AggregateProviderAddHandler.java
```diff
@@ -0,0 +1,90 @@
...
+    }
+
+    static void modifyRegistrationModel(OperationContext context, ModelNode op) throws OperationFailedException {
+        WriteAttributeHandler.INSTANCE.execute(context, op);
+        context.addStep(op, AggregateProviderAddHandler::modifyRegistration, OperationContext.Stage.RUNTIME);
+    }
+
+    static void modifyRegistration(OperationContext context, ModelNode op) throws OperationFailedException {
+        final MutableDiscoveryProvider mutableDiscoveryProvider = context.getCapabilityRuntimeAPI(DiscoveryExtension.DISCOVERY_PROVIDER_CAPABILITY, context.getCurrentAddressValue(), MutableDiscoveryProvider.class);
+        if (op.hasDefined(DiscoveryExtension.PROVIDERS)) {
```
bstansberry: This executes in Stage.RUNTIME so generally you don't want to use param values from 'op' here. By now the param values will have been stored in the model. So get the model and use that where you're using 'op' here:

ModelNode model = context.readResource(PathAddress.EMPTY_ADDRESS).getModel();

This ensures that if people do weird composite operations, wit multiple steps modifying the model, when your RUNTIME steps runs it is working with the end result of that.

Ah, actually you have to use the model here regardless, since in a write-attribute, the 'op' would be

{
 "operation=>"write-attribute",
 "address"=>[.....],
 "name"=>"providers",
 "value"=>["a","b","c"]
}

Doing op.get("providers") on that is not going to give you ["a","b","c"].

discovery/src/main/java/org/wildfly/extension/discovery/AggregateProviderDefinition.java
```diff
@@ -0,0 +1,61 @@
...
+ */
+final class AggregateProviderDefinition extends SimpleResourceDefinition {
+
+    static final StringListAttributeDefinition PROVIDER_NAMES = new StringListAttributeDefinition.Builder(DiscoveryExtension.PROVIDERS)
+        .setCapabilityReference(DiscoveryExtension.DISCOVERY_PROVIDER_CAPABILITY)
+        .build();
+
+    AggregateProviderDefinition() {
+        super(new Parameters(PathElement.pathElement(AGGREGATE_PROVIDER), DiscoveryExtension.getResourceDescriptionResolver(AGGREGATE_PROVIDER))
+            .setAddHandler(AggregateProviderAddHandler.getInstance())
```
bstansberry: This needs a remove step handler.

discovery/src/main/java/org/wildfly/extension/discovery/AggregateProviderDefinition.java
```diff
@@ -0,0 +1,61 @@
...
+            .setAddRestartLevel(OperationEntry.Flag.RESTART_RESOURCE_SERVICES)
```
bstansberry: It seems like this is RESTART_NONE; i.e. you call add whenever and it just works. (If so you don't need to set this.)

discovery/src/main/java/org/wildfly/extension/discovery/DiscoverySubsystemDefinition.java
```diff
@@ -0,0 +1,50 @@
...
+
+/**
+ * @author <a href="mailto:david.lloyd@redhat.com">David M. Lloyd</a>
+ */
+final class DiscoverySubsystemDefinition extends SimpleResourceDefinition {
+    private static final ResourceDefinition INSTANCE = new DiscoverySubsystemDefinition();
+
+    private DiscoverySubsystemDefinition() {
+        super(new Parameters(DiscoveryExtension.SUBSYSTEM_PATH, DiscoveryExtension.getResourceDescriptionResolver())
+            .setAddHandler(DiscoverySubsystemAddHandler.getInstance())
```
bstansberry: Needs a remove handler

discovery/src/main/java/org/wildfly/extension/discovery/StaticProviderAddHandler.java
```diff
@@ -0,0 +1,116 @@
...
+                .build().fromBaseCapability(context.getCurrentAddressValue()));
+        super.execute(context, operation);
+    }
+
+    protected void recordCapabilitiesAndRequirements(final OperationContext context, final ModelNode operation, final Resource resource) throws OperationFailedException {
+        // no operation
+    }
+
+    protected void performRuntime(final OperationContext context, final ModelNode operation, final Resource resource) throws OperationFailedException {
+        modifyRegistrationModel(context, operation);
```
bstansberry: WriteAttributeHandler.INSTANCE is a Stage.MODEL handler and will fail if called from performRuntime, which executes in Stage.RUNTIME. Also, WriteAttributeHandler won't understand the 'operation' ModelNode passed in from performRuntime as an 'add' operation has a different structure of parameters from a 'write-attribute'.

discovery/src/main/java/org/wildfly/extension/discovery/StaticProviderAddHandler.java
```diff
@@ -0,0 +1,116 @@
...
+        context.addStep(op, StaticProviderAddHandler::modifyRegistration, OperationContext.Stage.RUNTIME);
+    }
+
+    static void modifyRegistration(OperationContext context, ModelNode op) throws OperationFailedException {
+        final MutableDiscoveryProvider mutableDiscoveryProvider = context.getCapabilityRuntimeAPI(DiscoveryExtension.DISCOVERY_PROVIDER_CAPABILITY, context.getCurrentAddressValue(), MutableDiscoveryProvider.class);
+        if (op.hasDefined(DiscoveryExtension.SERVICES)) {
+            final List<ModelNode> list = StaticProviderDefinition.SERVICES.resolveModelAttribute(context, op).asList();
+            final List<ServiceURL> services = new ArrayList<>();
+            for (ModelNode servicesNode : list) {
+                final ModelNode abstractTypeNode = servicesNode.get(DiscoveryExtension.ABSTRACT_TYPE);
```
bstansberry: Best practice for all of these is to use AttributeDefinition.resolveModelAttribute(). It may or may not add any value versus a simple modelNode.get(...) but it's more resilient if things change in the future. (The resolveModelAttribute ensures expressions are resolved and validated if supported and that any default value for the attribute is returned if the attribute is undefined.)

discovery/src/main/java/org/wildfly/extension/discovery/StaticProviderAddHandler.java
```diff
@@ -0,0 +1,116 @@
...
+    }
+
+    static void modifyRegistrationModel(OperationContext context, ModelNode op) throws OperationFailedException {
+        WriteAttributeHandler.INSTANCE.execute(context, op);
+        context.addStep(op, StaticProviderAddHandler::modifyRegistration, OperationContext.Stage.RUNTIME);
+    }
+
+    static void modifyRegistration(OperationContext context, ModelNode op) throws OperationFailedException {
+        final MutableDiscoveryProvider mutableDiscoveryProvider = context.getCapabilityRuntimeAPI(DiscoveryExtension.DISCOVERY_PROVIDER_CAPABILITY, context.getCurrentAddressValue(), MutableDiscoveryProvider.class);
+        if (op.hasDefined(DiscoveryExtension.SERVICES)) {
```
bstansberry: Same comment re: 'model' vs 'op' that I had in AggregateProviderAddHandler applies here.

discovery/src/main/java/org/wildfly/extension/discovery/StaticProviderDefinition.java
```diff
@@ -0,0 +1,82 @@
...
+        ATTRIBUTES
+    ).build();
+
+    static final ObjectListAttributeDefinition SERVICES = new ObjectListAttributeDefinition.Builder(DiscoveryExtension.SERVICES, SERVICE).build();
+
+    private static final ResourceDefinition INSTANCE = new StaticProviderDefinition();
+
+    StaticProviderDefinition() {
+        super(new Parameters(PathElement.pathElement(STATIC_PROVIDER), DiscoveryExtension.getResourceDescriptionResolver(STATIC_PROVIDER))
+            .setAddHandler(StaticProviderAddHandler.getInstance())
```
bstansberry: Needs a remove handler.

discovery/src/main/java/org/wildfly/extension/discovery/StaticProviderDefinition.java
```diff
@@ -0,0 +1,82 @@
...
+            .setAddRestartLevel(OperationEntry.Flag.RESTART_RESOURCE_SERVICES)
```
bstansberry: It looks like this can be RESTART_NONE in which case you don't need to set it.

discovery/src/main/java/org/wildfly/extension/discovery/StaticProviderDefinition.java
```diff
@@ -0,0 +1,82 @@
...
+    static final SimpleAttributeDefinition ABSTRACT_TYPE_AUTHORITY = new SimpleAttributeDefinition(DiscoveryExtension.ABSTRACT_TYPE_AUTHORITY, ModelType.STRING, true);
+    static final SimpleAttributeDefinition URI = new SimpleAttributeDefinition(DiscoveryExtension.URI, ModelType.STRING, false);
+    static final SimpleAttributeDefinition URI_SCHEME_AUTHORITY = new SimpleAttributeDefinition(DiscoveryExtension.URI_SCHEME_AUTHORITY, ModelType.STRING, true);
+
+    static final SimpleAttributeDefinition NAME = new SimpleAttributeDefinition(DiscoveryExtension.NAME, ModelType.STRING, false);
+    static final SimpleAttributeDefinition VALUE = new SimpleAttributeDefinition(DiscoveryExtension.VALUE, ModelType.STRING, true);
+
+    static final ObjectTypeAttributeDefinition ATTRIBUTE = new ObjectTypeAttributeDefinition.Builder(DiscoveryExtension.ATTRIBUTE, NAME, VALUE).build();
+
+    static final ObjectListAttributeDefinition ATTRIBUTES = new ObjectListAttributeDefinition.Builder(DiscoveryExtension.ATTRIBUTES, ATTRIBUTE).build();
```
bstansberry: Most places we call these kinds of things "properties". Calling them 'attributes' is ok if people are going to learn that term in other contexts or if it's just important somehow, but otherwise being like other subsystems is good too.

discovery/src/main/java/org/wildfly/extension/discovery/StaticProviderDefinition.java
```diff
@@ -0,0 +1,82 @@
...
+ */
+final class StaticProviderDefinition extends SimpleResourceDefinition {
+
+    static final SimpleAttributeDefinition ABSTRACT_TYPE = new SimpleAttributeDefinition(DiscoveryExtension.ABSTRACT_TYPE, ModelType.STRING, true);
+    static final SimpleAttributeDefinition ABSTRACT_TYPE_AUTHORITY = new SimpleAttributeDefinition(DiscoveryExtension.ABSTRACT_TYPE_AUTHORITY, ModelType.STRING, true);
+    static final SimpleAttributeDefinition URI = new SimpleAttributeDefinition(DiscoveryExtension.URI, ModelType.STRING, false);
+    static final SimpleAttributeDefinition URI_SCHEME_AUTHORITY = new SimpleAttributeDefinition(DiscoveryExtension.URI_SCHEME_AUTHORITY, ModelType.STRING, true);
+
+    static final SimpleAttributeDefinition NAME = new SimpleAttributeDefinition(DiscoveryExtension.NAME, ModelType.STRING, false);
+    static final SimpleAttributeDefinition VALUE = new SimpleAttributeDefinition(DiscoveryExtension.VALUE, ModelType.STRING, true);
```
bstansberry: Our general rule is we support expressions on all simple attributes except model references and ones that encode class and module names (the class/module one has no strong reason but we just do it.)

core-feature-pack/src/main/resources/configuration/standalone/subsystems.xml
```diff
@@ -5,5 +5,6 @@
       <subsystem>logging.xml</subsystem>
       <subsystem>jmx.xml</subsystem>
       <subsystem>request-controller.xml</subsystem>
+      <subsystem>discovery.xml</subsystem>
```
dmlloyd: OK will do

discovery/src/main/java/org/wildfly/extension/discovery/AggregateProviderDefinition.java
dmlloyd: Actually I don't believe it does (rather, the default should suffice).  There's no action to perform, no services to remove, etc.

discovery/src/main/java/org/wildfly/extension/discovery/AggregateProviderDefinition.java
dmlloyd: OK, I wasn't sure about this.

discovery/src/main/java/org/wildfly/extension/discovery/DiscoverySubsystemDefinition.java
dmlloyd: See above (should be OK)

discovery/src/main/java/org/wildfly/extension/discovery/StaticProviderAddHandler.java
dmlloyd: OK will do

discovery/src/main/java/org/wildfly/extension/discovery/StaticProviderDefinition.java
dmlloyd: Hmm okay I think that'd be fine, I'll ponder it more to make sure though.

discovery/src/main/java/org/wildfly/extension/discovery/StaticProviderDefinition.java
dmlloyd: The proper name in the related RFCs is "attribute" so in this case I'd rather stick with that if possible.

discovery/src/main/java/org/wildfly/extension/discovery/StaticProviderDefinition.java
dmlloyd: See above.

discovery/src/main/java/org/wildfly/extension/discovery/AggregateProviderAddHandler.java
dmlloyd: Hmm I need to think about this when the kids aren't around...

discovery/src/main/java/org/wildfly/extension/discovery/AggregateProviderAddHandler.java
```diff
@@ -0,0 +1,90 @@
...
+                .build().fromBaseCapability(context.getCurrentAddressValue()));
+        super.execute(context, operation);
+    }
+
+    protected void recordCapabilitiesAndRequirements(final OperationContext context, final ModelNode operation, final Resource resource) throws OperationFailedException {
+        final AttributeDefinition ad = AggregateProviderDefinition.PROVIDER_NAMES;
+        ad.addCapabilityRequirements(context, resource.getModel().get(ad.getName()));
+    }
+
+    static void modifyRegistrationModel(OperationContext context, ModelNode op) throws OperationFailedException {
```
dmlloyd: No, but I put it here so it's near the runtime method, just for clarity.

discovery/src/main/java/org/wildfly/extension/discovery/StaticProviderAddHandler.java
dmlloyd: Hmm I think there shouldn't actually be a performRuntime here at all.

discovery/src/main/java/org/wildfly/extension/discovery/AggregateProviderAddHandler.java
dmlloyd: OK this will be fixed in the next iteration.

discovery/src/main/java/org/wildfly/extension/discovery/StaticProviderAddHandler.java
dmlloyd: OK

discovery/src/main/java/org/wildfly/extension/discovery/AggregateProviderDefinition.java
bstansberry: There is no default remove handler.

bstansberry: ModelOnlyRemoveStepHandler does a resource remove with no runtime activity.

discovery/src/main/java/org/wildfly/extension/discovery/StaticProviderDefinition.java
dmlloyd: Added a trivial remove handler.

dmlloyd:
This should be almost ready, but I want to wait for #1883 first to get the XML output that I want.

dmlloyd:
Note that at this point in time, all checks are passing. bbe0150

bstansberry:
Superceded by #1923 (which is merged.)

PR #1870, opened 2016-10-24
[WCFORE-1845] During boot skip validating/verifying runtime resources

jamezp:
During boot runtime resources should be ignored for validation. This also expands the resource to allow for a shallow copy of the resource. This is useful when child values may not be important, but the child key is important.


bstansberry:
@jamezp those failures concern me a bit; e.g. did we add a race or something?

jamezp:
@bstansberry I'm not sure. I just ran it several times locally and didn't have an issue at all. It's definitely a bit weird though. retest this please

jamezp:
Placing a hold on this until the Windows failure can be fixed.

PR #1871, opened 2016-10-24
WFCORE-1836 Rejection-style transformer tests assume each operation generated by parser has distinct PathAddress

rhusar:
Based on @kabir 's work which needed bit more tweaking. 

https://issues.jboss.org/browse/WFCORE-1836


bstansberry:
@kabir I'll leave this one to you. Looks ok to me, as AFAICT it doesn't do anything except add a new checkValue variant that I presume will be overridden in some subsystems in full.

PR #1873, opened 2016-10-25
[WFCORE-1885] Avoid NPE in boot processing if AbstractControllerServi…

bstansberry:
…ce is stopped while boot thread is executing. Also, in that situation interrupt the boot thread to cancel any ongoing boot ops

https://issues.jboss.org/browse/WFCORE-1855


bstansberry:
I'm going to close this because the branch and commit use the wrong JIRA #. I'll resubmit with that corrected.

PR #1877, opened 2016-10-26
WFCORE-1893 Wildfly build tools 1.1.7.Final

stuartwdouglas:
https://issues.jboss.org/browse/WFCORE-1893


jmesnil:
corresponding PR for wildfly: https://github.com/wildfly/wildfly/pull/9294

stuartwdouglas:
Note that these do not have to be merged at the same time, it does not matter if the tool versions are out of sync (which they are at the moment).

PR #1879, opened 2016-10-26
[WFCORE-1855] Avoid NPE in boot processing if AbstractControllerServi…

bstansberry:
…ce is stopped while boot thread is executing. Also, in that situation interrupt the boot thread to cancel any ongoing boot ops


bstansberry:
Closing for now; we can resurrect if there is urgency for this.

PR #1881, opened 2016-10-26
return abort instead of exiting on a registration exception

luck3y:
this allows the ServiceContainer to run a normal shutdown. If the Exiter is called here, the SC appears to wait forever as the service remains in STARTING.


luck3y:
My error, this should have been for the WFCORE-338 branch, not master. Closing.

PR #1883, opened 2016-10-27
WFCORE-1834 General parser improvements around complex attributes

ctomc:
- parser can now handle "unlimited" complex attribute nesting
 - introduced UNWRAPPED_OBJECT_LIST parser
 - fixes for overriding parsers with attribute groups
 - few general bug fixes

mostly done to support elytron subsystem parsing, see https://github.com/wildfly-security/elytron-subsystem/pull/277 for details

### DIFF
--- a/controller/src/main/java/org/jboss/as/controller/AttributeParsers.java
+++ b/controller/src/main/java/org/jboss/as/controller/AttributeParsers.java
@@ -79,7 +79,7 @@ public interface AttributeParsers {
 
         @Override
         public void parseElement(AttributeDefinition attribute, XMLExtendedStreamReader reader, ModelNode operation) throws XMLStreamException {
-            String wrapper = wrapperElement == null ? attribute.getName() : wrapperElement;
+            String wrapper = getXmlName(attribute);
             assert attribute instanceof MapAttributeDefinition;
             MapAttributeDefinition mapAttribute = (MapAttributeDefinition) attribute;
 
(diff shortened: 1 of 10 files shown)


controller/src/main/java/org/jboss/as/controller/AttributeParsers.java
```diff
@@ -200,6 +198,7 @@ public void parseElement(AttributeDefinition attribute, XMLExtendedStreamReader
             if (attribute.getXmlName().equals(reader.getLocalName())) {
                 ObjectTypeAttributeDefinition objectType = ((ObjectTypeAttributeDefinition) attribute);
                 ModelNode op = operation.get(attribute.getName());
+                op.setEmptyObject();
```
bstansberry: This seems strange.

bstansberry: I guess it's not so strange; this is a parser, so it expects to be starting with a clean operation node.

Was this showing up as a bug somehow or did you add this to be thorough?

controller/src/main/java/org/jboss/as/controller/SimpleMapAttributeDefinition.java
```diff
@@ -125,9 +127,9 @@ public SimpleMapAttributeDefinition build() {
             if (elementValidator == null) {
                 elementValidator = new ModelTypeValidator(valueType, allowNull, allowExpression);
             }
-            if (attributeMarshaller == null) {
+            /*if (attributeMarshaller == null) {
                 attributeMarshaller = new MapAttributeMarshaller();
-            }
+            }*/
```
bstansberry: This looks like a breaking change. There are a number of uses of the other Builder constructors besides the one you change at L108, so if any of those are used attributeMarshaller will be null.

bstansberry: I'm not sure how the change at L108 ties in to the rest of this PR either. It looks like the idea here is SimpleMapAttributeDefinition.MapAttributeMarshaller is meant to no longer be used (since the change here comments out the only use) with replacement by AttributeMarshallers.PropertiesAttributeMarshaller and a parser change too.

bstansberry:
retest this please

PR #1884, opened 2016-10-27
Fix for WFCORE-1901. Don't swallow exception in try/catch/finally

jfdenise:
I am proposing to use the "suppressed Exception" feature of the Exception class.
When the Finally exception is thrown, it is suppressing the try or catch exception that could have been previously thrown.
When an Exception is printed to the console, the suppressed exceptions are also printed.
Added unit tests to cover this feature.

### DIFF
--- a/cli/src/main/java/org/jboss/as/cli/Util.java
+++ b/cli/src/main/java/org/jboss/as/cli/Util.java
@@ -1187,6 +1187,17 @@ public class Util {
             }
             t1 = t1.getCause();
         }
+
+        // Add the suppressed exceptions attached to the passed exception.
+        // suppressed exceptions can contain valuable information to help trace
+        // exception path.
+        for (Throwable suppressed : t.getSuppressed()) {
+            if (suppressed.getLocalizedMessage() != null) {
+                buf.append("\n").append(suppressed.getLocalizedMessage());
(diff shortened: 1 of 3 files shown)


bstansberry:
@aloubyansky please approve

aloubyansky:
Looks good. Thanks!

cli/src/main/java/org/jboss/as/cli/Util.java
```diff
@@ -1187,6 +1187,10 @@ public static String getMessagesFromThrowable(Throwable t){
             }
             t1 = t1.getCause();
         }
+
+        for (Throwable suppressed : t.getSuppressed()) {
+            buf.append("\n").append(suppressed.getLocalizedMessage());
```
bstansberry: Should this have some sort of explanatory text to indicate that these are suppressed exceptions? We don't do that above for the getCause() handling, but now with 2 possible types of extra information does it get more confusing?

Should this be recursive:?

 buf.append("\n").append(getMessagesFromThrowable(suppressed));

If not should it have the if (t1.getLocalizedMessage() != null) { ... } else {...} handling like we do above? getLocalizedMessage() will be null if no details were provided to the exception constructor.

PR #1885, opened 2016-10-27
Move CLI to use aesh-readline.

jfdenise:
This is a a preliminary PR based on ash-readline 1.0-SNAPSHOT.
- faster CLI, startup, execution and shutdown, no more aesh threads.
- no more active CLI polling, aesh-readline is blocking.
- simpler handling of prompting (name, password, certificate).
- simpler handling of CTRL-C

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -97,7 +97,7 @@
         <version.org.codehaus.woodstox.woodstox-core>5.0.3</version.org.codehaus.woodstox.woodstox-core>
         <version.org.fusesource.jansi>1.16</version.org.fusesource.jansi>
         <version.org.glassfish.javax.json>1.0.4</version.org.glassfish.javax.json>
-        <version.org.jboss.aesh>0.66.18</version.org.jboss.aesh>
+        <version.org.aesh-readline>1.0-CR17-SNAPSHOT</version.org.aesh-readline>
         <version.org.jboss.byteman>3.0.6</version.org.jboss.byteman>
         <version.org.jboss.classfilewriter>1.2.1.Final</version.org.jboss.classfilewriter>
         <version.org.jboss.invocation>1.5.0.Beta4</version.org.jboss.invocation>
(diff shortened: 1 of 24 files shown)


cli/src/main/java/org/jboss/as/cli/handlers/QuitHandler.java
```diff
@@ -41,5 +41,6 @@ public QuitHandler(String command) {
     @Override
     protected void doHandle(CommandContext ctx) {
         ctx.terminateSession();
+        System.exit(0);
```
bstansberry: Calling System.exit needs to be carefully controlled or it will break embedded usage. "Quitting" an embedded CLI may not mean the user wants the embedding app to exit. On the server side we do this by using a SystemExiter interface, and all "exit" calls use that. We then swap in different impls of the interface depending on how the server is being used. The standard impl that is used when we know the server was started from the command line calls System.exit. Others do not.

jfdenise:
@bstansberry, actually this call to System.exit was not needed. I have removed it. When the aesh console is closed, it exists its waiting loop. The Main thread continues and the CLI exists.

bstansberry:
I put the Hold label on this to stop me looking at it while it awaits the move away from requiring the -SNAPSHOT. When it's ready please have Alexey have a look and then ping me.

jfdenise:
retest this please!

pom.xml
```diff
@@ -94,7 +94,7 @@
         <version.org.codehaus.woodstox.stax2-api>3.1.4</version.org.codehaus.woodstox.stax2-api>
         <version.org.codehaus.woodstox.woodstox-core>5.0.3</version.org.codehaus.woodstox.woodstox-core>
         <version.org.fusesource.jansi>1.11</version.org.fusesource.jansi>
-        <version.org.jboss.aesh>0.66.11</version.org.jboss.aesh>
+        <version.org.jboss.aesh-readline>1.0-SNAPSHOT</version.org.jboss.aesh-readline>
```
aloubyansky: I think we can't depend on a snapshot in the core.

jfdenise: Yes, the plan is to move to 1.0 prior to merge.

marekkopecky:
@jfdenise Are there jiras for this PR?

jfdenise:
@marek, not yet, will create one.

jfdenise:
@marekkopecky, new JIRA created: https://issues.jboss.org/browse/WFCORE-2021

marekkopecky:
@jfdenise You mean https://issues.jboss.org/browse/WFCORE-2038, right? :)

jfdenise:
@marekkopecky , yes, I edited the comment.

jfdenise:
retest this please!

jfdenise:
retest this please!

jfdenise:
retest this please!

PR #1887, opened 2016-10-31
WFCORE-1908 Metric attribute is not writable.

soul2zimate:
https://issues.jboss.org/browse/WFCORE-1908

METRIC is read-only attribute https://github.com/wildfly/wildfly-core/blob/master/controller/src/main/java/org/jboss/as/controller/registry/AttributeAccess.java#L44

### DIFF
--- a/cli/src/main/java/org/jboss/as/cli/impl/AttributeNamePathCompleter.java
+++ b/cli/src/main/java/org/jboss/as/cli/impl/AttributeNamePathCompleter.java
@@ -225,7 +225,11 @@ public class AttributeNamePathCompleter implements CommandLineCompleter {
         }
 
         private boolean isWritable(ModelNode attrDescr) {
-            return !attrDescr.has(Util.ACCESS_TYPE) || !Util.READ_ONLY.equals(attrDescr.get(Util.ACCESS_TYPE).asString());
+            return !isReadOnly(attrDescr);
+        }
+
+        private boolean isReadOnly(ModelNode attrDescr) {
+            return attrDescr.has(Util.ACCESS_TYPE) && (Util.READ_ONLY.equals(attrDescr.get(Util.ACCESS_TYPE).asString()) || Util.METRIC.equals(attrDescr.get(Util.ACCESS_TYPE).asString()));
         }
 
         public int getCandidateIndex() {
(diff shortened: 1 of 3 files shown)


jmesnil:
@soul2zimate could you open a corresponding PR issue for EAP7 please?

jmesnil:
@jfdenise could you approve it please?

soul2zimate:
I think this master branch is for EAP 7.1.0 GA.

jmesnil:
@soul2zimate you're right. sorry about that

cli/src/main/java/org/jboss/as/cli/impl/AttributeNamePathCompleter.java
```diff
@@ -225,7 +225,7 @@ protected AttributeNamePathCallbackHandler parse(String line) throws CommandForm
         }
 
         private boolean isWritable(ModelNode attrDescr) {
-            return !attrDescr.has(Util.ACCESS_TYPE) || !Util.READ_ONLY.equals(attrDescr.get(Util.ACCESS_TYPE).asString());
+            return !attrDescr.has(Util.ACCESS_TYPE) || !(Util.READ_ONLY.equals(attrDescr.get(Util.ACCESS_TYPE).asString()) || Util.METRIC.equals(attrDescr.get(Util.ACCESS_TYPE).asString()));
```
jfdenise: I think that this condition starts to be a bit complex. It could be rewritten in a more explicit way:
private boolean isWritable(ModelNode attrDescr) {
  return !isReadOnly(attrDescr);
}
 private boolean isReadOnly(ModelNode attrDescr) {
   return attrDescr.has(Util.ACCESS_TYPE) && 
              (Util.READ_ONLY.equals(attrDescr.get(Util.ACCESS_TYPE).asString()) ||
                Util.METRIC.equals(attrDescr.get(Util.ACCESS_TYPE).asString()));
}

jfdenise:
Thank you for this fix. If that is not too complex to test, could we have a simple unit test to cover it?
JF.

soul2zimate:
Thanks for review, a simple test is added to cover this issue.

jfdenise:
Hi @soul2zimate, thank you very much for the changes. That is good to merge.
Thanks.
JF

PR #1894, opened 2016-11-03
[WFCORE-1928] OperationFailedException for when there is no value par…

kabir:
…ameter in :validate-operation

https://issues.jboss.org/browse/WFCORE-1928
https://issues.jboss.org/browse/JBEAP-6846


bstansberry:
I'm closing this one as it will conflict with #1893. I think it should a ParameterValidator too; see #1897.

PR #1898, opened 2016-11-04
WFCORE-1930 PersistentResourceDefinition registers a ReloadRequiredWr…

stuartwdouglas:
…iteHandler however it does not enfore that the correct flag has been set

This will require https://github.com/wildfly/wildfly/pull/9335 to be merged, as it now checks that the appropriate flag is set before registering the attribute


ctomc:
This needs a different fix.
as reload required write handler should be registered only you have that flag set.
in other cases something different should be done. 
What is that different thing is bit more problematic

ctomc:
This needs a different fix.
as reload required write handler should be registered only you have that flag set.
in other cases something different should be done. 
What is that different thing is bit more problematic

ctomc:
This needs a different fix.
as reload required write handler should be registered only you have that flag set.
in other cases something different should be done. 
What is that different thing is bit more problematic

stuartwdouglas:
@ctomc This PR is not ideal, however this is the best we can do ATM without doing proper write handlers for each attribute

bstansberry:
retest this please

stuartwdouglas:
Looks like I forgot to update a separate PR that added some new attributes. I have filed https://github.com/wildfly/wildfly/pull/9341

bstansberry:
retest this please

luck3y:
retest this please

bstansberry:
retest this please

luck3y:
retest this please

luck3y:
Note, I've opened WFCORE-2090 for the Embedded test issue in build 5015 (https://ci.wildfly.org/viewLog.html?buildId=36570&buildTypeId=WildFlyCore_PullRequest)

josejulio:
To use [`RESTART_RESOURCE_SERVICES`](https://github.com/wildfly/wildfly-core/blob/master/controller/src/main/java/org/jboss/as/controller/registry/AttributeAccess.java#L128) on an attribute, would I need to write a custom write handler?

PR #1913, opened 2016-11-07
WFCORE-1374 - use posix syntax for --add-exports that works with launcher

ctomc:
This way launcher doesn't choke on parameters and most of testsuite runs.


ctomc:
JDK9 CI run https://ci.wildfly.org/viewLog.html?buildId=31796&tab=buildResultsDiv&buildTypeId=WildFlyCore_MasterLinuxJdk9

PR #1919, opened 2016-11-08
[WFCORE-1959]: Deploying an empty managed exploded deployment to server group in domain fails.

ehsavoie:
When routing to managed servers, remove the EMPTY content if the hash is correctly filled.

Jira: https://issues.jboss.org/browse/WFCORE-1959
JBEAP: https://issues.jboss.org/browse/JBEAP-6977


bstansberry:
Looks fine assuming it passes.

PR #1924, opened 2016-11-09
[WFCORE-1966] Move util to create test xml parser into core from full

kabir:
https://issues.jboss.org/browse/WFCORE-1966

### DIFF
--- a/testsuite/shared/src/main/java/org/wildfly/test/mixed/domain/TestParserUtils.java
+++ b/testsuite/shared/src/main/java/org/wildfly/test/mixed/domain/TestParserUtils.java
@@ -1,0 +1,137 @@
+/*
+ * JBoss, Home of Professional Open Source
+ * Copyright 2016, Red Hat, Inc., and individual contributors as indicated
+ * by the @authors tag.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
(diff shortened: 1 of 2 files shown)


kabir:
Retest this please

testsuite/shared/src/main/java/org/wildfly/test/mixed/domain/TestParserUtils.java
```diff
@@ -0,0 +1,103 @@
...
+            this.processType = processType;
+            return this;
+        }
+
+        public Builder setRunningModeControl(RunningModeControl runningModeControl) {
+            this.runningModeControl = runningModeControl;
+            return this;
+        }
+
+        public Builder setHostControllerInfoAccessor(RuntimeHostControllerInfoAccessor hostControllerInfoAccessor) {
```
bstansberry: This is another class that shouldn't be used in full.

I think this could be a simple setMasterHostController(Boolean master) and then build creates the RuntimeHostControllerInfoAccessor (with null meaning use the SERVER constant.) Or something like that. ;)

testsuite/shared/src/main/java/org/wildfly/test/mixed/domain/TestParserUtils.java
```diff
@@ -0,0 +1,103 @@
...
+            this.namespace = namespace;
+            return this;
+        }
+
+        public Builder setProcessType(ProcessType processType) {
+            this.processType = processType;
+            return this;
+        }
+
+        public Builder setRunningModeControl(RunningModeControl runningModeControl) {
```
bstansberry: This class should not be used in full

kabir:
@bstansberry should be ok now. I am not sure if I should somehow flag that changes have been made

PR #1929, opened 2016-11-09
WFCORE-1971 Incorrect address completion when node starts with quote

TomasHofman:
https://issues.jboss.org/browse/WFCORE-1971
https://issues.jboss.org/browse/JBEAP-6872

### DIFF
--- a/cli/src/main/java/org/jboss/as/cli/parsing/ParserUtil.java
+++ b/cli/src/main/java/org/jboss/as/cli/parsing/ParserUtil.java
@@ -238,10 +238,13 @@ public class ParserUtil {
                         }
                     } else {
                         final String value = buffer.toString().trim();
-                        if (ch == '=') {
+                        // If ctx.isEndOfContent() is true and the last character is '=' or ':', it must be escaped and
+                        // it won't be considered a type/name separator (if it wasn't escaped the NodeState would've
+                        // been left already).
+                        if (ch == '=' && !ctx.isEndOfContent()) {
                             handler.nodeType(bufferStartIndex, value);
                             handler.nodeTypeNameSeparator(ctx.getLocation());
-                        } else if (ch == ':') {
+                        } else if (ch == ':' && !ctx.isEndOfContent()) {
                             handler.nodeName(bufferStartIndex, value);
                         } else {
                             if (".".equals(value)) {
--- a/cli/src/main/java/org/jboss/as/cli/operation/OperationRequestCompleter.java
+++ b/cli/src/main/java/org/jboss/as/cli/operation/OperationRequestCompleter.java
@@ -35,6 +35,9 @@ import org.jboss.as.cli.EscapeSelector;
 import org.jboss.as.cli.Util;
 import org.jboss.as.cli.impl.ArgumentWithoutValue;
 import org.jboss.as.cli.operation.impl.DefaultCallbackHandler;
+import org.jboss.as.cli.operation.impl.SegmentParsingInitialState;
+import org.jboss.as.cli.parsing.StateParser;
+import org.jboss.logging.Logger;
 
 
 /**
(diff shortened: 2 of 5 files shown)


TomasHofman:
Legit failure, investigating.

TomasHofman:
Retest this please.

jfdenise:
Thank you for this fix and all the unit tests you have added! 
I made a comment on a change of behaviour that has been introduced. Having a more solid support for escaped/not escaped chars and partially quoted values would help.
JF

TomasHofman:
@jfdenise thanks for your comment. I can work on this more next week, I will try to take this further.

bstansberry:
@jfdenise Please ping me when this one is ready to go and I'll remove the 'hold' label.

TomasHofman:
I pushed updated version - not complete yet.

I'm reposting @jfdenise's comment on change in NodeNameFormattingTestCase.java, as it doesn't show after pushing new changes. The change was:
```
 -        assertEquals(Arrays.asList("equals\\=equals"), fetchCandidates("/root=\"equals=\"")); 
 +        assertEquals(Arrays.asList("\"equals\\=equals\""), fetchCandidates("/root=\"equals=\"")); 
```

>I think that this case is still valid and should be kept. It means that we can have parts of the name (or type) that are quoted.
/root="equals=" should give /root="equals="equals
>
>I made a simpler test (with a property named fafafafa, no character to escape).
/system-property="fafa" ==> should give /system-property="fafa"fafa but gives /system-property="fafa"afa" (notice the missing f and '"' at the end).

I see, I reverted this change in behaviour. The closed quotes like in your example are working now, even combination of closed and open quotes:

/subsystem="data" => /subsystem="data"sources
/subsystem="data"" => /subsystem="data""sources"
/subsystem="data""sour" => /subsystem="data""sour"ces

>I think that we have another issue when the characters should be escaped but are not because located inside a quoted string. For example property named gogo=gaga=gugu can be used as is once quoted, if not quoted, one need to escape the '=': gogo=gaga=gugu

Specifically the = character is causing some exceptions which I didn't look into yet. With other characters like : or / it's working now.

>/system-property="gogo=gag" ==> the candidates are internally escaped (line 501) although they are not in the user input. This create an invalid offset.
>
>Inside a quoted String, it doesn't seem that the candidate should be escaped except if the candidates contains '"' character.
A complex case would be: "gogo=gaga=to"to". Only subpart would need to be escaped,

Fixed, now I'm only escaping " and \ characters when inside quotes. I think the \ character should be always escaped, otherwise it wouldn't be clear if it's there to escape next char or simply as a character itself.

I also think a user should be allowed to enter both ways - escaped and unescaped - inside quotes. I'm trying to keep it working both ways.

Also fixed a case when there were leading spaces, which didn't work properly previously:

/subsystem= data => /subsystem= datasources

TomasHofman:
@jfdenise I think I'm done - I tested all kinds of quoting and escaping I could think of. The case you mentioned with "gogo=gaga=to\"to" is working too.

bstansberry:
@jfdenise is this one ok now?

jfdenise:
@bstansberry , it should be. I need to take some time to properly review it.

cli/src/main/java/org/jboss/as/cli/parsing/ParserUtil.java
```diff
@@ -238,7 +238,7 @@ public void leavingState(ParsingContext ctx) throws CommandFormatException {
                         }
                     } else {
                         final String value = buffer.toString().trim();
-                        if (ch == '=') {
+                        if (ch == '=' && !ctx.isEndOfContent()) {
```
jfdenise: Perhaps you could add a comment saying something like:
"if '=' is escaped, it will be not considered as a type/name separator. There is an exception when it is the last parsed character (e.g.: "test=foo\="). In this case, if we are at the end of the content, we don't consider it a separator. If we are not at the end of content, we consider it a separator."

jfdenise:
@TomasHofman , excellent enhancement. Very well designed. Thank-you for your contribution!
I simply made a trivial comment on a possible piece of explanation that could be added. Other than that, that is good to merge.
JF

cli/src/main/java/org/jboss/as/cli/parsing/ParserUtil.java
```diff
@@ -238,10 +238,13 @@ public void leavingState(ParsingContext ctx) throws CommandFormatException {
...
                         final String value = buffer.toString().trim();
-                        if (ch == '=') {
+                        // If ctx.isEndOfContent() is true and the last character is '=' or ':', it must be escaped and
+                        // it won't be considered a type/name separator (if it wasn't escaped the NodeState would've
+                        // been left already).
+                        if (ch == '=' && !ctx.isEndOfContent()) {
                             handler.nodeType(bufferStartIndex, value);
                             handler.nodeTypeNameSeparator(ctx.getLocation());
-                        } else if (ch == ':') {
+                        } else if (ch == ':' && !ctx.isEndOfContent()) {
```
TomasHofman: @jfdenise thanks - glad I could help. I added the comment about ```ctx.isEndOfContent()``` and additionally handled also the ':' character in the same way. ':' was accidentally working, but basically the same case as '='.

jfdenise:
@bstansberry , this one is good to merge. Thanks.
JF

PR #1931, opened 2016-11-10
[WFCORE-1731] Where we bind an existing client to the CommandContext create a ConnectionInfoBean to track the time.

darranl:
This eliminates a NullPointerException calling connection-info when running an embedded server.

### DIFF
--- a/cli/src/main/java/org/jboss/as/cli/impl/CommandContextImpl.java
+++ b/cli/src/main/java/org/jboss/as/cli/impl/CommandContextImpl.java
@@ -1186,7 +1186,9 @@ class CommandContextImpl implements CommandContext, ModelControllerClientFactory
 
     @Override
     public void bindClient(ModelControllerClient newClient) {
-        initNewClient(newClient, null, null);
+        ConnectionInfoBean conInfo = new ConnectionInfoBean();
+        conInfo.setLoggedSince(new Date());
+        initNewClient(newClient, null, conInfo);
     }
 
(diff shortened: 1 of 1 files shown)


bstansberry:
@jfdenise Please approve.

This PR is fine IMO; I'm asking you to have a look more to point out some things in this ConnectionInfo interface. Things not really relevant to this PR. The "disableLocalAuth" property is never read, and the "username" property is only read in a test case, where it is set on a MockConnectionInfo by a MockCommandContext -- i.e. the test is checking that the mocks work.

So those properties look like leftover cruft.

jfdenise:
Approved.

The ConnectionInfo has been a replacement for a bunch of properties that the CommandContext was exposing (I guess as a private contract). Seems to me an implementation detail of the way the ConnectionInfoHandler is implemented more than a complete connection state API. But it is now a public interface with interesting piece of information.

The ConnectionInfo should have been updated when the whoami operation has been used to retrieve the name/roles. For some reason, the username used to log is no more in use. The whoami returned value is used instead. Can we have a mismatch between these 2? One reason I see is that $local is retrieved from whoami and is not hardcoded in CLI. But a generic "Local User" could be good as well.

ConnectionInfo doesn't contain the roles that the command display. It could be an interesting piece of information exposed from this interface. At least we would have a command and an interface aligned (lazily retrieve the roles from the ConnectionInfo implementation).

I would be ready to remove the "disableLocalAuth" property, it is redundant with the userName. A userName == null means !disableLocalAuth. Removing the userName is more problematic. I would keep it, although not used, it could be used by third party code.

PR #1938, opened 2016-11-11
WFCORE-1677 Make AbstractSubsystemBaseTest#testSchemaOfSubsystemTempl…

rhusar:
…ates opt-in as opposed to flooding reports with @Ignored tests

This will save 18 ignores in wf-core and 57 unnecessary ignores in wf (needs a subsequent PR to wf which can go asynchronously to this one).

https://issues.jboss.org/browse/WFCORE-1677

### DIFF
--- a/subsystem-test/framework/src/main/java/org/jboss/as/subsystem/test/AbstractSubsystemBaseTest.java
+++ b/subsystem-test/framework/src/main/java/org/jboss/as/subsystem/test/AbstractSubsystemBaseTest.java
@@ -130,13 +130,16 @@ public abstract class AbstractSubsystemBaseTest extends AbstractSubsystemTest {
         SchemaValidator.validateXML(getSubsystemXml(), schemaPath, getResolvedProperties());
     }
 
-    @Test
-    public void testSchemaOfSubsystemTemplates() throws Exception {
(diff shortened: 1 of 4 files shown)


subsystem-test/framework/src/main/java/org/jboss/as/subsystem/test/AbstractSubsystemBaseTest.java
```diff
@@ -130,8 +130,7 @@ public void testSchema() throws Exception {
         SchemaValidator.validateXML(getSubsystemXml(), schemaPath, getResolvedProperties());
     }
 
-    @Test
-    public void testSchemaOfSubsystemTemplates() throws Exception {
+    protected void testSchemaOfSubsystemTemplates() throws Exception {
         String schemaPath = getSubsystemXsdPath();
         Assume.assumeTrue("getSubsystemXsdPath() has been overridden to disable the validation of the subsystem templates",
```
kabir: I would change this to an assert

rhusar: Ah, of course. Done.

PR #1939, opened 2016-11-11
[WFCORE-1975]: EmbeddedProcessFactory does not accept module pathes w…

stoty:
…ith multiple directories

This PR fixes the check that tries to interpret the path as a directory. With this patch only the first element of the path is checked.

It also removes the setting of the `jboss.modules.dir` system property that is deprecated, and has no effect whatsoever.

### DIFF
--- a/embedded/src/main/java/org/wildfly/core/embedded/EmbeddedProcessFactory.java
+++ b/embedded/src/main/java/org/wildfly/core/embedded/EmbeddedProcessFactory.java
@@ -260,15 +260,16 @@ public class EmbeddedProcessFactory {
 
         assert modulePath != null : "modulePath not null";
 
-        // verify the supplied modules path exists, and if it does not, stop and allow the user to correct.
+        // verify the the first element of the supplied modules path exists, and if it does not, stop and allow the user to correct.
         // Once modules are initialized and loaded we can't change Module.BOOT_MODULE_LOADER (yet).
 
-        File moduleDir = new File(modulePath);
+        File moduleDir = new File(trimPathToModulesDir(modulePath));
         if (!moduleDir.exists() || !moduleDir.isDirectory()) {
-            throw new RuntimeException("The specified module directory " + modulePath + " is invalid or does not exist.");
+            throw new RuntimeException("The first directory of the specified module path " + modulePath + " is invalid or does not exist.");
         }
 
-        WildFlySecurityManager.setPropertyPrivileged(SYSPROP_KEY_JBOSS_MODULES_DIR, trimPathToModulesDir(modulePath));
+        // deprecated property
+        WildFlySecurityManager.setPropertyPrivileged(SYSPROP_KEY_JBOSS_MODULES_DIR, moduleDir.getPath());
 
         final String classPath = WildFlySecurityManager.getPropertyPrivileged(SYSPROP_KEY_CLASS_PATH, null);
         try {


embedded/src/main/java/org/wildfly/core/embedded/EmbeddedProcessFactory.java
```diff
@@ -260,15 +259,14 @@ private static ModuleLoader setupModuleLoader(String modulePath, String... syste
...
+        // verify the the first element of the supplied modules path exists, and if it does not, stop and allow the user to correct.
         // Once modules are initialized and loaded we can't change Module.BOOT_MODULE_LOADER (yet).
 
-        File moduleDir = new File(modulePath);
+        File moduleDir = new File(trimPathToModulesDir(modulePath));
         if (!moduleDir.exists() || !moduleDir.isDirectory()) {
             throw new RuntimeException("The specified module directory " + modulePath + " is invalid or does not exist.");
         }
 
-        WildFlySecurityManager.setPropertyPrivileged(SYSPROP_KEY_JBOSS_MODULES_DIR, trimPathToModulesDir(modulePath));
```
luck3y: I think this still might need to be set. In org.jboss.as.server.ServerEnvironment its still referenced, and if present used to set a deprecated variable.

I haven't tested that though, thoughts?

luck3y:
Thanks for this change, I had a minor comment on the PR.

embedded/src/main/java/org/wildfly/core/embedded/EmbeddedProcessFactory.java
stoty: As far as I was able to follow, that system property is indeed set and read at some points of the code, but apart from checking that the referenced directory exists, nothing is affected by it.

It is referenced in org.jboss.as.server.ServerEnvironment, but all it does is check if it is set to a real directory, and sets it to a default if it is not set, then stores it back to the property.

It is also referenced in a bunch of other locations, that read, write, verify, and massage the configuration model, but has no actual effect on classloading (or anything).

The PR passes the Embedded test cases and my use case. 
AFAIK wildfly Arquillian smoke tests also use this internally to start/sop the WF copies.

That said, it is still possible that I have missed something, and keeping it it doesn't affect the fix for the original bug (only the readability of the code)

stoty:
I realize that mixing bug fixes and refactoring is not always a good idea, so if you do not have the bandwidth to check the effects, I will revert the environment property removal.

embedded/src/main/java/org/wildfly/core/embedded/EmbeddedProcessFactory.java
luck3y: ServerEnvironment.getModulesDir() is marked deprecated, so until we remove this, we should probably keep this set (just to cover the case where someone is using that method for some reason).

Would you mind updating this to keep the property set? Sorry for the extra work, and thanks a lot for the patch!

stoty:
I have amended the PR. It reverts the Property setting, fixes the error message, and removes the double calling of the trimPathToModulesDir() method.

luck3y:
Retest this please

luck3y:
Retest this please

luck3y:
retest this please

bstansberry:
@luck3y some day we should look into doing one or the other of these:

1) Drop this setting of jboss.modules.dir. That prop is just cruft; see javadoc of MODULES_DIR in org.jboss.as.server.ServerEnvironment
2) or consider changing trimPathToModulesDir to return the last element in the path, not the first. I'm not sure what our documented best practices are around multiple elements in the module path are (if they exist) but my gut instinct is they would say that user-provided modules would take precedence over system ones. In which case $JBOSS_HOME/modules would come last in the path.

The fact that we parse out the first element here doesn't affect the real functioning of the system. It just affects what the server reports to users of the management API as being the "modules dir" via the deprecated modules-dir attribute of the /core-service=server-environment resource.

None of this is relevant or in scope to this PR/JIRA; this comment is just because looking at the PR is what sparked the thought. :)

Thanks @stoty!

PR #1944, opened 2016-11-15
WFCORE-1980, add an option to ignore unresolved expression to resolve…

soul2zimate:
… as more as possible for argument value expressions.

https://issues.jboss.org/browse/WFCORE-1980

Hi @jfdenise, Could you review this? I found out the problem as I comment in Jira, but I'm not sure if this is the best way to fix it.

### DIFF
--- a/cli/src/main/java/org/jboss/as/cli/ArgumentValueConverter.java
+++ b/cli/src/main/java/org/jboss/as/cli/ArgumentValueConverter.java
@@ -47,7 +47,7 @@ public interface ArgumentValueConverter {
                 return new ModelNode();
             }
             if(ctx.isResolveParameterValues()) {
-                value = CLIExpressionResolver.resolveOrOriginal(value);
+                value = CLIExpressionResolver.resolveLax(value);
             }
             try {
                 return ModelNode.fromString(value);
@@ -66,7 +66,7 @@ public interface ArgumentValueConverter {
                 return new ModelNode();
             }
             if(ctx.isResolveParameterValues()) {
-                value = CLIExpressionResolver.resolveOrOriginal(value);
+                value = CLIExpressionResolver.resolveLax(value);
             }
             ModelNode toSet = null;
             try {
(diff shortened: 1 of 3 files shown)


jfdenise:
Hi @soul2zimate, I checked the 6.4 source code and Property resolution was permissive (unknown permission was kept as is). So your approach is good to me.

I would not change the resolveOrOriginal method, I would keep it. I would introduce a new method named "resolveLax" that would simply call resolve(value, false).

Thank you, one more that you fixed!
JF

jfdenise:
Hi, thank you for the changes. I think that resolveLax shouldn't throw an exception. You should catch in resolveLax.
Other than that that is good to merge.
JF

soul2zimate:
That's safe one to just satisfy the compiler, it will actually not throw any exception as the parameter _exceptionIfNotResolved_ is false.  The catch clause in resolveLax will do nothing for same reason. 
Otherwise, it needs to catch the UnresolvedExpressionException in resolve() instead of dealing with it in resolveOrOriginal() and resolveLax() for something like: 

```
+        String original = input;
         int i = 0;
         while(i < input.length()) {
             final char c = input.charAt(i++);
@@ -106,10 +113,14 @@ public class CLIExpressionResolver {
             } else if(c == '{') {
                 if(state == DOLLAR) {
                     state = INITIAL;
-                    final String[] inputRef = new String[]{input};
-                    if(i - 2 == resolveProperty2(inputRef, i - 2, exceptionIfNotResolved)) {
-                        input = inputRef[0];
-                        i -= 2;
+                    final String[] inputRef = new String[] { input };
+                    try {
+                        if (i - 2 == resolveProperty2(inputRef, i - 2, exceptionIfNotResolved)) {
+                            input = inputRef[0];
+                            i -= 2;
+                        }
+                    } catch (UnresolvedExpressionException e) {
+                        return original;
```

jfdenise:
It is really a matter of contract that the resolveLax method exposes. If it can't technically throw exception, then it shouldn't declare an Exception and add a catch that does nothing. Just add a comment in the resolveLax catch block saying // XXX OK, never thrown in this case.

soul2zimate:
OK,  it's rebased.

soul2zimate:
Oh, I thought "// XXX OK" means a leave out for some explanation words. 
OK, now it's added.

jfdenise:
Approved. Thanks again!
JF

PR #1947, opened 2016-11-15
[WFCORE-1954] Enabling previously disabled SSL tests.

darranl:

Various changes as required to enable most of the previously disabled SSL tests.

https://issues.jboss.org/browse/WFCORE-1954
https://issues.jboss.org/browse/WFCORE-1990
https://issues.jboss.org/browse/WFCORE-1991
https://issues.jboss.org/browse/WFCORE-2030
https://issues.jboss.org/browse/WFCORE-2031

Do not resolve WFCORE-1978 - that is mentioned here as one affected test needs additional follow up for domain mode.


darranl:
Retest this please

testsuite/manualmode/src/test/java/org/wildfly/core/test/standalone/mgmt/HTTPSConnectionWithCLITestCase.java
```diff
@@ -75,7 +74,6 @@
 
 @RunWith(WildflyTestRunner.class)
 @ServerControl(manual = true)
-@Ignore("[WFCORE-1954] Wrap CLIENT_CERT behaviour of legacy realms with Elytron components.")
```
bstansberry: Testing this on master-ignore with a number of other changes, this test failed on windows:

https://ci.wildfly.org/viewLog.html?buildId=33100&tab=buildResultsDiv&buildTypeId=WildFlyCore_MasterIgnorWindows

The log didn't show anything. The test does a number of config mods to set up HTTPS on the management interface and then adds a native interface so the CLI can connect and do a reload. The log shows a message that the server is listening on 9999 but then when the CLI tries to connect to 9999 it fails.

darranl:
I will try and reproduce on a local machine.

bstansberry:
Please @ ping me when you think this is good to go and I'll remove the fixme label.

darranl:
10 Runs of this test on a local windows machine and all successful.

I do have some native management issues to investigate so I will try adding them to this PR and see if that clears the issue.

darranl:
Failures in Full are to be expected as the following PRs need to be merged first: -

https://github.com/wildfly/wildfly-core/pull/1980
https://github.com/wildfly/wildfly/pull/9416

darranl:
Still struggling to find out how the changes made could be impacting the tests here, of the two tests failing only one was ignored the other seems to be running fine on other test runs.  

I have seen locally some issues with the CLI not shutting down properly using the native management interface but that is on Linux where we don't have a test failure.  Maybe not related but worth keeping in mind.

However I have realised the native management interface is not correctly establishing an SSLSession so need to get that corrected first.

darranl:
Retest this please

PR #1950, opened 2016-11-15
[WFCORE-1789]: Ignoring intermittent failing test

ehsavoie:
Jira: https://issues.jboss.org/browse/WFCORE-1789


kabir:
I see that the most frequent failures are when executing the reload operation and the shutdown operation.

There is a org.jboss.as.test.integration.management.util.ServerReload util used to reload which I think should be used for the first case. @ctomc any thoughts on the shutdown part?

(I made this comment a little while ago on the wrong PR)

kabir:
@bstansberry I think the IOException during shutdown should probably be allowed since ServerReload allows that for reload. After all the socket is getting closed down

PR #1951, opened 2016-11-15
[WFCORE-1899]: Sensitive data can be exposed at the server level in domain mode.

ehsavoie:
Propagate the sensistive-classification and the application-classification to the managed servers.

Jira: https://issues.jboss.org/browse/WFCORE-1899
JBEAP: https://issues.jboss.org/browse/JBEAP-6281


kabir:
I see that the most frequent failures are when executing the reload operation and the shutdown operation.

There is a org.jboss.as.test.integration.management.util.ServerReload util used to reload which I think should be used for the first case. @ctomc any thoughts on the shutdown part?

bstansberry:
Something has gone wrong here.

ehsavoie:
The issue is due to Elytron requiring rbac tests to be ignored for now.

PR #1956, opened 2016-11-16
[WFCORE-1823] port JDK method to util class

baranowb:
https://issues.jboss.org/browse/WFCORE-1823


bstansberry:
We'll need to just drop the ensureContrast call as this is not ASL or similarly reusable code. The ensureContrast isn't doing much anyway.

bstansberry:
Superceded by #2007 which is just a rebase of this. Sorry I didn't notice this was updated earlier. I'll give up on the fixme label as the github hook meant to clear it when the PR changes isn't reliable.

baranowb:
Weird, Ive rebased it and those commits were gone. Now they are back.

PR #1959, opened 2016-11-16
WFCORE-1510 Fix code to not use default platform dependant encoding

ctomc:


bstansberry:
@jamezp Please ok the change to the 2 logging module classes. They change the default behavior of the read-log-file op's ENCODING param (i.e. if unspecified) from "whatever the system default is" to UTF-8. Which seems ok to a naive guy like me.

@ctomc writing the above made me realize that with this change the default value of that param's AD should be set. But be careful, that AD is now used in other places too and those shouldn't be affected.

Other than that, this looks ok to me. Note I didn't really pay attention to the test classes!

logging/src/main/java/org/jboss/as/logging/LogFileResourceDefinition.java
```diff
@@ -257,7 +258,7 @@ public void execute(final OperationContext context, final ModelNode operation) t
             }
             try (
                     final InputStream in = (tail ? new LifoFileInputStream(file) : Files.newInputStream(file.toPath()));
-                    final InputStreamReader isr = (encoding == null ? new InputStreamReader(in) : new InputStreamReader(in, encoding));
+                    final InputStreamReader isr = (encoding == null ? new InputStreamReader(in, StandardCharsets.UTF_8) : new InputStreamReader(in, encoding));
```
jamezp: I think we should stick with the default here and not use UTF-8. The encoding on the file handler does not default to UTF-8 but the system default. I think here we should stick with the system default unless explicitly defined. I could see a UTF-8 default possibly being problematic on IBM bases systems.

logging/src/main/java/org/jboss/as/logging/LoggingResourceDefinition.java
```diff
@@ -300,7 +301,7 @@ public void execute(final OperationContext context, final ModelNode operation) t
                     in = new FileInputStream(file);
                 }
                 if (encoding == null) {
-                    reader = new BufferedReader(new InputStreamReader(in));
+                    reader = new BufferedReader(new InputStreamReader(in, StandardCharsets.UTF_8));
```
jamezp: Same note as above, this should use the system default.

logging/src/main/java/org/jboss/as/logging/logmanager/ConfigurationPersistence.java
```diff
@@ -59,7 +60,7 @@
     private static final Object LOCK = new Object();
     private static final String PROPERTIES_FILE = "logging.properties";
     private static final byte[] NOTE_MESSAGE = String.format("# Note this file has been generated and will be overwritten if a%n" +
-            "# logging subsystem has been defined in the XML configuration.%n%n").getBytes();
+            "# logging subsystem has been defined in the XML configuration.%n%n").getBytes(StandardCharsets.UTF_8);
```
jamezp: This looks good, thanks!

domain-management/src/main/java/org/jboss/as/domain/management/security/SecretIdentityService.java
```diff
@@ -66,7 +67,7 @@ public void start(StartContext startContext) throws StartException {
         final char[] thePassword;
         if (base64) {
             byte[] value = Base64.getDecoder().decode(password);
-            String tempPassword = new String(value);
+            String tempPassword = new String(value, StandardCharsets.UTF_8);
```
jamezp: Probably not a big deal, but the `Base64.getDecoder().decode(string)` uses ISO_8859_1 encoding by default. UTF-8 should work fine, I just wasn't sure if we wanted to be consistent with what the decoder uses.

domain-management/src/main/java/org/jboss/as/domain/management/security/adduser/DisplaySecret.java
```diff
@@ -43,7 +45,7 @@ public DisplaySecret(ConsoleWrapper theConsole, final StateValues stateValues) {
     }
 
     public State execute() {
-        String pwdBase64 = Base64.getEncoder().encodeToString(stateValues.getPassword().getBytes());
+        String pwdBase64 = Base64.getEncoder().encodeToString(stateValues.getPassword().getBytes(StandardCharsets.UTF_8));
```
jamezp: Same as the decode this by default uses ISO_8859_1. Again though I don't think it's a big deal.

ctomc: This one is not the same, as this value is fed to base64 encoder not being decoded from it.

ctomc:
Updated PR to reflect latest comments

bstansberry:
An IBM run (core only, no full): https://ci.wildfly.org/viewLog.html?buildId=33491&tab=buildResultsDiv&buildTypeId=WildFlyCore_MasterLinuxIbmJdk

jfdenise:
The CLI changes are fine.
Thanks.
JF

jamezp:
Looks good to me. Thanks @ctomc.

bstansberry:
I chatted with James. It looks ok and it passes the IBM CI. We (i.e. the WildFly devs) are not really set up to do a lot of manual experimentation on different systems. So our consensus is bake time is the best way to check for odd effects on different systems.

PR #1963, opened 2016-11-17
https://issues.jboss.org/browse/WFCORE-2008 Unable to use path in inp…

aloubyansky:
…ut-stream-index argument in /core-service=patching:patch


bstansberry:
Looks good.

@aloubyansky FYI, see the discussion about this kind of handling on https://issues.jboss.org/browse/JBEAP-7044.

PR #1967, opened 2016-11-18
Fix for WFCORE-2005. Better filesystem path completion. Reworked file…

jfdenise:
Took the opportunity of this fix to rework and fix the File completers. 
- Escaping and quotes handling during completion.
- '~' support.
- Fixed path translation for attachments.
- Small improvements.

Added unit tests.

### DIFF
--- a/cli/src/main/java/org/jboss/as/cli/handlers/FilenameTabCompleter.java
+++ b/cli/src/main/java/org/jboss/as/cli/handlers/FilenameTabCompleter.java
@@ -23,11 +23,17 @@ package org.jboss.as.cli.handlers;
 
 import java.io.File;
 import java.io.IOException;
+import java.util.Collections;
 import java.util.List;
 import java.util.Objects;
 
 import org.jboss.as.cli.CommandContext;
+import org.jboss.as.cli.CommandFormatException;
 import org.jboss.as.cli.CommandLineCompleter;
+import org.jboss.as.cli.Util;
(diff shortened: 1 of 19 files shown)


cli/src/main/java/org/jboss/as/cli/handlers/FilenameTabCompleter.java
```diff
@@ -60,45 +65,62 @@ protected int getCandidates(String buffer, List<String> candidates) {
...
-        final String translated;
+        String translated;
         // special character: ~ maps to the user's home directory
         if (path.startsWith("~" + File.separator)) {
             translated = System.getProperty("user.home") + path.substring(1);
-        } else if (path.startsWith("~")) {
-            translated = new File(System.getProperty("user.home"))
-                    .getParentFile().getAbsolutePath();
-        } else if (!startsWithRoot(path)) {
+        } else if (path.startsWith("~") && !Util.isWindows()) {// Makes sense only on Unix.
```
jfdenise: I noticed that Powershell recognises the "~\" shortcut. With the CLI. on all platforms, one can use this shortcut.

The else case seems fine, if Windows and the path starts with ~xxx, then this path will be returned as is, completion will not propose anything because the path is invalid on this platform. If the path doesn't start with ~, check for Absolute, if not, complete it, finally returns the original path.

bstansberry:
@aloubyansky please approve

aloubyansky:
Looks good, thanks!

PR #1968, opened 2016-11-18
[WFCORE-2009] Correct attribute validation to reflect that SSL configuration can now come from either an ssl-context or a security-realm.

darranl:
https://issues.jboss.org/browse/WFCORE-2009


darranl:
Retest this please

PR #1969, opened 2016-11-19
[WFCORE-2017] Don't enable the access=identity service if running in embedded admin only mode.

darranl:
https://issues.jboss.org/browse/WFCORE-2017

### DIFF
--- a/domain-management/src/main/java/org/jboss/as/domain/management/access/AccessIdentityResourceDefinition.java
+++ b/domain-management/src/main/java/org/jboss/as/domain/management/access/AccessIdentityResourceDefinition.java
@@ -34,9 +34,11 @@ import org.jboss.as.controller.AttributeDefinition;
 import org.jboss.as.controller.OperationContext;
 import org.jboss.as.controller.OperationFailedException;
 import org.jboss.as.controller.PathElement;
+import org.jboss.as.controller.ProcessType;
 import org.jboss.as.controller.ReloadRequiredRemoveStepHandler;
 import org.jboss.as.controller.ReloadRequiredWriteAttributeHandler;
 import org.jboss.as.controller.ResourceDefinition;
+import org.jboss.as.controller.RunningMode;
 import org.jboss.as.controller.SimpleAttributeDefinition;
 import org.jboss.as.controller.SimpleAttributeDefinitionBuilder;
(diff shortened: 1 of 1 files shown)


bstansberry:
Why is this service never appropriate in an embedded server?

bstansberry:
Never mind; I misread the change.

domain-management/src/main/java/org/jboss/as/domain/management/access/AccessIdentityResourceDefinition.java
```diff
@@ -170,7 +172,7 @@ protected void performRuntime(OperationContext context, ModelNode operation, Mod
 
         @Override
         protected boolean requiresRuntime(OperationContext context) {
-            return true;
+            return context.getProcessType() != ProcessType.EMBEDDED_SERVER || context.getRunningMode() != RunningMode.ADMIN_ONLY;
```
bstansberry: This needs to account for ProcessType.EMBEDDED_HOST_CONTROLLER as well.

I'm not sure on this but perhaps ProcessType could expose and isEmbedded() property to help with keeping this particular kind of discrimination consistent. My only hesitation is whether that will always have a single value for all process types. For example, for SELF_CONTAINED it's now false but maybe someday there would be a flavor of that kind of thing where it is ambiguous.

bstansberry: I'll make ^^^ part of a formal "Request changes" review.

darranl: The HTTP management interfaces don't make that check - I have tried to make sure this matches those so when they come up this comes up.

darranl: This change matches the check for the management interfaces, I have tried to match the interface as it is important if they come up this comes up.

darranl: Never mind - I see it in the other add handler.

darranl: Never mind - I see it in the other add handler.

PR #1971, opened 2016-11-21
WFCORE-2021 informative error message if it can not convert result to…

soul2zimate:
… list.
https://issues.jboss.org/browse/WFCORE-2021

### DIFF
--- a/cli/src/main/java/org/jboss/as/cli/handlers/LsHandler.java
+++ b/cli/src/main/java/org/jboss/as/cli/handlers/LsHandler.java
@@ -571,6 +571,9 @@ public class LsHandler extends BaseOperationCommand {
             while (iterator.hasNext()) {
                 OperationRequestAddress.Node node = iterator.next();
                 if (node.getName() != null) {
+                    if (node.getName().equals("*")) {
+                        throw new CommandFormatException("* is not supported in node path argument");
+                    }
                     addressNode.add(node.getType(), node.getName());
                 } else if (iterator.hasNext()) {
                     throw new OperationFormatException(


bstansberry:
@jfdenise can confirm but I think the problem here is that the ls command is not designed to handle wildcard addresses (i.e. those with * in the address). Which I think is fine. But I think the solution then should be to detect such addresses and reject them before submission to the server, with a targeted error message.

The fact that the subsystem=* example in the JIRA fails with a somewhat meaningful error message from the server side is really just luck. It fails because it is trying to read resource description data for a resource type (subsystem=*) that has no registration. There is no wildcard resource registration for "subsystem" while there is for "socket-binding-group."

soul2zimate:
Yes, I agree with you. I will see how to distinguish such addresses

jfdenise:
ls handler is not able to deal with response being a list of resources (as @soul2zimate detailed in the bug description). It could be enhanced but rejecting the syntax seems a good fix easy to change in case we want to support it.

soul2zimate:
I have rebased this to reject any address contains * in ls handler before sending to server

jfdenise:
Hi, I would possibly replace the message by "* not supported in node path argument"

soul2zimate:
Thanks, the message is updated as suggested.

soul2zimate:
retest this please

cli/src/main/java/org/jboss/as/cli/handlers/LsHandler.java
```diff
@@ -571,6 +571,9 @@ private ModelNode getAddressNode(CommandContext ctx, OperationRequestAddress add
             while (iterator.hasNext()) {
                 OperationRequestAddress.Node node = iterator.next();
                 if (node.getName() != null) {
+                    if (node.getName().equals("*")) {
+                        throw new CommandFormatException("* is not supported in node path argument");
```
bstansberry: @jfdenise Do we use the 'node' much in CLI docs, help, etc?

I think I've seen it before but it's a bit odd to use that term  in messages unless it's the typical way navigation is described.

jfdenise: @bstansberry , yes node path is used specially in the ls --help.

PR #1975, opened 2016-11-21
WFCORE-2026 source-port is wrongly offsetted when configured to 0, i.e. ephemeral port; results in "Permission denied (Bind failed)"

rhusar:
Jira
https://issues.jboss.org/browse/WFCORE-2026

Causes 
https://issues.jboss.org/browse/WFLY-6915

### DIFF
--- a/network/src/main/java/org/jboss/as/network/OutboundSocketBinding.java
+++ b/network/src/main/java/org/jboss/as/network/OutboundSocketBinding.java
@@ -200,7 +200,7 @@ public class OutboundSocketBinding {
 
     /**
      * Returns the source address of this outbound socket binding if one is configured. If no explicit source address
-     * is specified for this binding, then this method returns {@code null}. Use {@link org.jboss.as.network.OutboundSocketBinding#getSourceAddress()}
+     * is specified for this binding, then this method returns {@code null}. Use {@link #getSourceAddress()}
      * instead to obtain the default interface of the socket binding manager if none is specified for this binding.
      *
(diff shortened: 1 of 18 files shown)


network/src/main/java/org/jboss/as/network/OutboundSocketBinding.java
```diff
@@ -215,21 +215,21 @@ public InetAddress getOptionalSourceAddress() {
      * this outbound socket binding has a port offset. To get the absolute source port, use the {@link #getAbsoluteSourcePort()}
      * method.
      *
-     * @return the source port for this outbound socket binding
+     * @return the source port for this outbound socket binding not accounting for port offset/fixation; {@code null} if an ephemeral port should be used
      */
     public Integer getSourcePort() {
-        return this.sourcePort;
+        return sourcePort == 0 ? null : this.sourcePort;
```
ahaeber: Looks like a null check on sourcePort is missing here because sourcePort is Integer and therefore can be null. If sourcePort is null then the comparison "== 0" will cause problems when trying to debox sourcePort to int.

Just tried the following similar code and it crashed with NPE:
`Integer sourcePort = null;`
`int x = (sourcePort == 0 ? null : sourcePort);`

bstansberry: Plus let's not mix the code style; use 'this.sourceport' or just 'sourceport' consistently in the statement.

But the NPE is more important. :D

PR #1978, opened 2016-11-22
Better text; l10n team rightly found the current text hard to understand

bstansberry:


jamezp:
The tests haven't run, but this should be okay.

PR #1981, opened 2016-11-24
[WFCORE-1157] listener API for process state changes

ehsavoie:
* add API to let user be notified of process state changes by adding a
class to the resource process-state-listeners in the core-management subsystem.
* add by default the core-management subsystems to the standalone and
  domain subsystem templates.
* split the core-management module in 2 parts : client and subsystem.
* one service per listener
* add a specific JMX notifier that listens on RuntimeConfigurationState and RunningState changes : RunningStateJmx.
* basic documentation.

JIRA: https://issues.jboss.org/browse/WFCORE-1157
EAP: https://issues.jboss.org/browse/JBEAP-7201


core-management/core-management-client/src/main/java/org/wildfly/extension/core/management/client/ProcessStateListener.java
```diff
@@ -0,0 +1,51 @@
...
+ * @author <a href="http://jmesnil.net/">Jeff Mesnil</a> (c) 2016 Red Hat inc.
+ */
+public interface ProcessStateListener {
+
+    /**
+     * Initialize the listener.
+     * This should not throw an exception.
+     * @param properties the properties to initialize the listener.
+     */
+    default void init(Map<String, String> properties) {}
```
kabir: **If** we do this, we should make sure that we wrap this in a class (ProcessStateListenerInitEvent or something like that) so that we can extend it with more information in future

core-management/core-management-subsystem/src/main/java/org/wildfly/extension/core/management/ProcessStateListenerResourceDefinition.java
```diff
@@ -0,0 +1,154 @@
...
+                    context.getProcessType(),
+                    context.getRunningMode(), context.getCurrentAddress().getLastElement().getValue(),
+                    listener,
+                    properties,
+                    timeout);
+        }
+
+        @Override
+        protected boolean requiresRuntime(OperationContext context) {
+            return super.requiresRuntime(context) || context.getRunningMode() == RunningMode.ADMIN_ONLY;
```
kabir: Probably just return true
* if processType.isServer()
* Otherwise, check that the address begins with the 'host' and 'subsystem' keys

ehsavoie:
retest this please

PR #1983, opened 2016-11-25
WFCORE-2045 print warning when argument '--redeploy-affected' is not …

soul2zimate:
…specified in deployment-overlay actions

It tries to print a warning message about missing --redeploy-affected in cases:

1. Action add, Creating an Overlay and linking it to deployments in one command when optional --deployments is specified
2. Action link, Linking an existing Overlay to deployments when required --deployment is specified
3. Action upload, Uploading content to an existing overlay when required --content is specified.
4. Action remove, Unlinking deployments when --deployments is specified, removing content when --content is specified, and removing the overlay altogether with its content and links if none of above is specified.

https://issues.jboss.org/browse/WFCORE-2045

### DIFF
--- a/cli/src/main/java/org/jboss/as/cli/handlers/DeploymentOverlayHandler.java
+++ b/cli/src/main/java/org/jboss/as/cli/handlers/DeploymentOverlayHandler.java
@@ -76,6 +76,8 @@ public class DeploymentOverlayHandler extends BatchModeCommandHandler {//Command
     private static final byte REDEPLOY_ONLY_AFFECTED = 1;
     private static final byte REDEPLOY_ALL = 2;
 
+    private static final String WARN_MSG = "WARNING: redeployment is required on deployment ";
+
     private final ArgumentWithoutValue l;
     private final ArgumentWithValue action;
     private final ArgumentWithValue name;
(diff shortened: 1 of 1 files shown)


bstansberry:
@jfdenise Do you agree with this enhancement?

Does this get printed in both interactive and non-interactive sessions?

jfdenise:
@bstansberry , yes I like it. This seems to be valuable for users to identify possible erroneous commands. In non interactive it should make sense too, you get a trace that something seems strange in the list of executed commands.

soul2zimate:
Thanks, this is rebased.

jfdenise:
Hi, thank you, that is much better! Perhaps one last tiny point, introduce a static to fully avoid message duplication:
private static final String WARN_MSG = "WARNING: redeployment is required on deployment ";

soul2zimate:
yep, now the message part is static.

cli/src/main/java/org/jboss/as/cli/handlers/DeploymentOverlayHandler.java
```diff
@@ -1153,11 +1196,19 @@ protected void addAddRedeployLinksSteps(CommandContext ctx, ModelNode steps,
                         }
                     }
                 }
+            } else {
+                warn = true;
             }
         }
+        if (warn) {
+            String warningMsg = serverGroup == null
+                    ? "WARNING: redeployment is required on deployment " + Arrays.toString(links)
```
jfdenise: You should use the static part there.

soul2zimate:
my apologies, totally forgot it in the "add" action.

jfdenise:
Ok, thank you!

PR #1988, opened 2016-11-28
Fix for WFCORE-2049. Prompting for username/password breaks redirection.

jfdenise:
Calling continuous too early breaks redirection. Due to some change in the way Auth callbacks are received, the console is put in an invalid state.
I have tried to simplify the general logic, this does fixes the problem and doesn't seem to introduce regression. Only manual testing required there.


bstansberry:
@aloubyansky please approve

aloubyansky:
Yes, looks good, we've discussed it with jfdenise.

PR #1991, opened 2016-11-28
Fix for WFCORE-2054. keep-content if not all groups are listed

jfdenise:
My understanding is that, in domain mode, user can provide a subset of serverGroup at undeploy time.
If --keep-content is not set, then the artefact will be not undeployed because still deployed in other serverGroup. 
The change is as follow:
- If keep-content has not been set, we check that all server groups have been listed. If that is not the case, then we keep the content.

Unit test added.

### DIFF
--- a/cli/src/main/resources/help/undeploy.txt
+++ b/cli/src/main/resources/help/undeploy.txt
@@ -29,12 +29,14 @@ ARGUMENTS
                     is required in domain mode. This argument is not applicable
                     in the standalone mode.
 
- --keep-content   - by default undeploy, besides disabling the deployment, also
-                    removes its content from the repository. The presence of
-                    --keep-content will only disable the deployment w/o removing
-                    its content from the repository. This argument can be used
-                    in both standalone and domain modes.
-						
+ --keep-content   - Only disable the deployment w/o removing its content from the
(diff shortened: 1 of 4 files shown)


cli/src/main/resources/help/undeploy.txt
```diff
@@ -30,10 +30,13 @@ ARGUMENTS
...
-                    removes its content from the repository. The presence of
-                    --keep-content will only disable the deployment w/o removing
-                    its content from the repository. This argument can be used
-                    in both standalone and domain modes.
+                    removes its content from the repository. There is an exception
+                    in domain mode if not all server-groups that reference the
+                    deployment are listed. In this case, the content is kept in
+                    the repository. The presence of --keep-content will only
+                    disable the deployment w/o removing its content from the
+                    repository. This argument can be used in both standalone and
```
bstansberry: I think this would be better if this "The --presence of --keep-content..." sentence was dropped and the first line of the help was "only disable the deployment without removing its content from repository. By default...". Currently the text provides quite a bit of background before getting to the key point of what the param does, so that point gets a bit lost.

jfdenise:
@bstansberry , thanks. Yes better phrasing.
JF

bstansberry:
retest this please

PR #1994, opened 2016-11-29
[WFCORE-1842] Add a use-identity-roles attribute to enable the roles from the identity to be used directly.

darranl:
https://issues.jboss.org/browse/WFCORE-1842


darranl:
Retest this please

PR #2001, opened 2016-12-02
WFCORE-2011 Inconsistency of formatter and named-formatter in console logging handler

tedwon:
https://issues.jboss.org/browse/WFCORE-2011

### DIFF
--- a/controller/src/main/java/org/jboss/as/controller/operations/global/ReadAttributeHandler.java
+++ b/controller/src/main/java/org/jboss/as/controller/operations/global/ReadAttributeHandler.java
@@ -281,9 +281,19 @@ public class ReadAttributeHandler extends GlobalOperationHandlers.AbstractMultiT
             final ModelNode result = subModel.get(attribute.getName());
             context.getResult().set(result);
         } else if (defaults && attribute.getDefaultValue() != null) {
-            // No defined value in the model. See if we should reply with a default from the metadata,
-            // reply with undefined, or fail because it's a non-existent attribute name
-            context.getResult().set(attribute.getDefaultValue());
+            if(!attribute.hasAlternative(subModel)) {
+                // No defined value in the model. See if we should reply with a default from the metadata,
+                // reply with undefined, or fail because it's a non-existent attribute name
+                context.getResult().set(attribute.getDefaultValue());
+            } else {
+                final ModelNode undefined = new ModelNode();
+                for(final String alternative : attribute.getAlternatives()) {
+                    if(!undefined.equals(subModel.get(alternative))) {
+                        context.getResult().set(undefined);
+                        break;
+                    }
+                }
+            }
         } else {
             // model had no defined value, but we treat its existence in the model or the metadata
             // as proof that it's a legit attribute name
--- a/model-test/src/main/java/org/jboss/as/model/test/ModelTestUtils.java
+++ b/model-test/src/main/java/org/jboss/as/model/test/ModelTestUtils.java
@@ -683,4 +683,26 @@ public class ModelTestUtils {
         return result;
     }
 
+    public static ModelNode removeDefaultFormatter(final ModelNode model) {
+        for (String key : model.keys()) {
+            final ModelNode subModel = model.get(key);
+            if (subModel.getType() == ModelType.OBJECT) {
+                for (String subKey : subModel.keys()) {
+                    final ModelNode childModel = subModel.get(subKey);
+                    if (childModel.getType() == ModelType.OBJECT) {
+                        final Set<String> keys = childModel.keys();
+                        if(keys.contains("named-formatter")) {
+                            final ModelNode undefined = new ModelNode();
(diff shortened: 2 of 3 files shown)


controller/src/main/java/org/jboss/as/controller/operations/global/ReadAttributeHandler.java
```diff
@@ -284,6 +284,16 @@ static void resolveAttribute(OperationContext context, AttributeDefinition attri
...
+            // Reply with undefined when the attribute has declared alternatives and one of those alternatives is defined.
+            if (attribute.getAlternatives() != null) {
+                final ModelNode undefined = new ModelNode();
+                for(String alternative : attribute.getAlternatives()) {
+                    if(!undefined.equals(subModel.get(alternative))) {
+                        context.getResult().set(undefined);
+                        break;
+                    }
+                }
+            }
```
bstansberry: This checking for alternatives should be a callout to a separate method that returns a boolean, with the callout added to the 'else if' check at L283. So if there's a default and no alternative, set the result to the default, else drop to the 'else' case.

bstansberry:
The test failures are going to take some thought. What is happening is the test is comparing output from the current kernel with output from a previous version, and now they are different because of this change.

The test fixture has mechanisms to deal with this kind of thing (i.e. to configure the test to ignore discrepancies that are ok); that's probably what needs to be used here. That's the 'legacyModelFixer' param passed to the AbstractSubsystemTest.checkSubsystemModelTransformation method you see in the stack trace of the failed tests.

It's a pain to get a clean CI run though, as the failed tests are in WildFly full. So we'd have to alter full to get the tests to pass before merging this patch.

controller/src/main/java/org/jboss/as/controller/operations/global/ReadAttributeHandler.java
tedwon: I fixed the code with using hasAlternative method. I found and applied hasAlternative method of AttributeDefinition object to check alternatives as a callout to a separate method.  Please review the change. :-)

tedwon:
Hello @bstansberry ,
I created a new WildFly PR https://github.com/wildfly/wildfly/pull/9745 to get the tests to pass in WildFly full.
I really appreciate for your lots of hint. :D

To pass tests, the WildFly PR's wildfly-core dependency should base on this PR's branch (https://github.com/tedwon/wildfly-core/tree/WFCORE-2011).
I ran tests of both WildFly and WildFly-Core successfully in my local with the changes. However, I'm wondering how to make(or configure) CI run with applying two tightly related PRs to run tests successfully. 

Would you please review them ?
Thanks in advance.

controller/src/main/java/org/jboss/as/controller/operations/global/ReadAttributeHandler.java
```diff
@@ -281,9 +281,19 @@ static void resolveAttribute(OperationContext context, AttributeDefinition attri
             final ModelNode result = subModel.get(attribute.getName());
             context.getResult().set(result);
         } else if (defaults && attribute.getDefaultValue() != null) {
-            // No defined value in the model. See if we should reply with a default from the metadata,
-            // reply with undefined, or fail because it's a non-existent attribute name
-            context.getResult().set(attribute.getDefaultValue());
+            if(!attribute.hasAlternative(subModel)) {
```
JiriOndrusek: I was backporting this change to EAP 7.0 and I've found that there is a flag, that attribute is allowed to be null. I think that even in case that there is an alternative, but field cannot be null, we can not set it to null. I backported it with this change - https://github.com/jbossas/wildfly-core-eap/pull/378 (only difference is on this line) - may I ask for opinion?

jamezp: I don't think we want to back port this at this point. It's not in WildFly Core currently and I'm not sure it will make it into EAP 7.1. @bstansberry can make the final call, but it's too late to change this behavior IMO. It's been like this for some time now.

jamezp: This is an enhancement not a bug BTW.

JiriOndrusek: Yes you are right, it is an enhancement. I started backporting this issue, because on 7.2 it is reported as bug - https://issues.jboss.org/browse/JBEAP-7392 and I missed that on WF it is an enhancement. In that case backporting doesn't make sense.

jmesnil:
@tedwon As this pull request has been held up for a long time, I am closing it to get the pull request queue under control again. Since this is very old now, it will be a very long time until this one will be addressed. Also, I don't have much time to spend on this to evaluate on a case by case basis.

If it is a new feature, please follow the new process for features. Ping @wildfly/mergers-core if you are unsure what this means, and we can help. Please do that before opening the pull request though.

Sorry for any inconvenience!

PR #2003, opened 2016-12-02
[WFCORE-2080] Instantiating the WildFlyElytronProvider is an expensive operation so only do it once.

darranl:

https://issues.jboss.org/browse/WFCORE-2080

### DIFF
--- a/domain-management/src/main/java/org/jboss/as/domain/management/security/SecurityRealmService.java
+++ b/domain-management/src/main/java/org/jboss/as/domain/management/security/SecurityRealmService.java
@@ -97,8 +97,6 @@ import org.wildfly.security.sasl.util.SortedMechanismSaslServerFactory;
  */
 public class SecurityRealmService implements Service<SecurityRealm>, SecurityRealm {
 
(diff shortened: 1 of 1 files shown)


domain-management/src/main/java/org/jboss/as/domain/management/security/SecurityRealmService.java
```diff
@@ -97,6 +97,8 @@
  */
 public class SecurityRealmService implements Service<SecurityRealm>, SecurityRealm {
 
+    private static final Provider ELYTRON_PROVIDER = new WildFlyElytronProvider();
```
darranl: Thinking about this I can probably get away with moving this into the service lifecycle, the reported problem was it was being created per request.

ctomc: As long as it is not created in both providers directly it is fine.
I was testing locally by just having 
`Provider[] providers = new Provider[]{new WildFlyElytronProvider()}`
in service code and than returning that in both providers.

PR #2005, opened 2016-12-05
WFCORE-2083  Reduce number of object allocations and memory footprint

ctomc:
WFCORE-2083 reduce memory footprint and number of allocations
WFCORE-2083 Only register transformers when needed
WFCORE-2083 Use serviceTarget from deployment phase context
WFCORE-2083 Builder inner class should be static

### DIFF
--- a/controller/src/main/java/org/jboss/as/controller/AbstractAttributeDefinitionBuilder.java
+++ b/controller/src/main/java/org/jboss/as/controller/AbstractAttributeDefinitionBuilder.java
@@ -76,6 +76,8 @@ public abstract class AbstractAttributeDefinitionBuilder<BUILDER extends Abstrac
     protected Map<String, ModelNode> arbitraryDescriptors = null;
     private ModelNode undefinedMetricValue;
 
+    private static AccessConstraintDefinition[] ZERO_CONSTRAINTS = new AccessConstraintDefinition[0];
+
     /**
      * Creates a builder for an attribute with the give name and type. Equivalent to
      * {@link #AbstractAttributeDefinitionBuilder(String, org.jboss.dmr.ModelType, boolean) AbstractAttributeDefinitionBuilder(attributeName, type, false}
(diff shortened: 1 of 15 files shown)


controller/src/main/java/org/jboss/as/controller/AbstractAttributeDefinitionBuilder.java
```diff
@@ -907,6 +909,9 @@ public String getAttributeGroup() {
         if (toCopy == null) {
             return null;
         }
+        if (toCopy.length == 0){
+            return ZERO_CONSTRAINTS;
```
bstansberry: In the end this value goes into AttributeDefinition.wrapConstraints. If you tweak that to return Collections.emptyList() for a zero-sized array, that will save a bit more.

server/src/main/java/org/jboss/as/server/deployment/integration/Seam2Processor.java
```diff
@@ -78,19 +78,19 @@
     public static final ModuleIdentifier EXT_CONTENT_MODULE = ModuleIdentifier.create("org.jboss.integration.ext-content");
     public static final ModuleIdentifier VFS_MODULE = ModuleIdentifier.create("org.jboss.vfs");
 
-    private ServiceTarget serviceTarget; // service target from the service that added this dup
     private ResourceRoot seamIntResourceRoot;
 
-    public Seam2Processor(ServiceTarget serviceTarget) {
```
bstansberry: This change isn't valid. I checked with Stuart and the JaxrsSpringProcessor one in full isn't either and needs to be reverted.

The service installed here is not scoped to the deployment, so using the deployment's ServiceTarget is not valid.

Perhaps this and the JaxrsSpringProcessor thing could be done differently, but I think we can make the OperationContextImpl classes work better. An attempt at this is at https://github.com/wildfly/wildfly-core/pull/2008

bstansberry: I hit the wrong button; now ^^^ is part of a github review.

PR #2006, opened 2016-12-05
WFCORE-2084 all module.xml should be using 1.5 version of xsd schema

ctomc:
jira: https://issues.jboss.org/browse/WFCORE-2084


bstansberry:
@ehsavoie please have a look at the Linux test failure; it's not due to this PR. :)

ehsavoie:
It's org.wildfly.core.test.standalone.mgmt.RemoveManagementInterfaceTestCase.testRemoveManagementInterface that breaks, my test is run after this failure which leaves the server in an unstartable state :)

PR #2008, opened 2016-12-06
[WFCORE-2096] Decouple OperationContext-specific ServiceTarget/Builder from the con…

bstansberry:
…text when op execution is done

https://issues.jboss.org/browse/WFCORE-2096

Goal is to prevent refs to the ServiceTarget provided by OperationContext holding refs to all the state associated with the op after the op is done, since it is valid for users of the OC to hold a ref to the target after the op is done.

Makes the ServiceTarget and ServiceBuilder impl classes created by OperationContextImpl static so they hold no OperationContext.this ref. Provides a temporary ref to the OCI logic to their constructors, and clears that ref when op execution is done, turning the objects after that point into simple wrappers around the normal MSC classes.


ctomc:
retest this please

PR #2009, opened 2016-12-07
Credential store tool scripts

pskopek:
https://issues.jboss.org/browse/WFCORE-2079
https://issues.jboss.org/browse/EAP7-539

This PR adds new dependency/module to wf-core. It is apache commons-cli which is already part of WildFly.
It also creates new module with main-class to be used by scripts.

### DIFF
--- a/core-feature-pack/pom.xml
+++ b/core-feature-pack/pom.xml
@@ -126,6 +126,11 @@
         </dependency>
 
         <dependency>
+            <groupId>commons-cli</groupId>
+            <artifactId>commons-cli</artifactId>
+        </dependency>
+
+        <dependency>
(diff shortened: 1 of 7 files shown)


dmlloyd:
Also I think it's important that these command scripts live in the Elytron repo proper so all Elytron users can utilize them.

bstansberry:
I'm closing this as it needs discussion and we need to clean out the queue.

Besides the points David raises I'm not sure we'd want to add another dependency to core for such a purpose. Core is meant to be as small as possible. In an analogous situation, the org.jboss.as.vault-tool module is not part of core. (The vault scripts are, which is perhaps a bug, although not one I feel any great urgency to fix.)

PR #2010, opened 2016-12-07
[WFLY-6936] RebindTestCase fails with security manager

iweiss:
Issue: https://issues.jboss.org/browse/WFLY-6936

### DIFF
--- a/protocol/src/main/java/org/jboss/as/protocol/ProtocolConnectionUtils.java
+++ b/protocol/src/main/java/org/jboss/as/protocol/ProtocolConnectionUtils.java
@@ -48,10 +48,14 @@ import javax.security.auth.callback.NameCallback;
 import javax.security.auth.callback.UnsupportedCallbackException;
 
 import java.io.IOException;
+import java.lang.reflect.UndeclaredThrowableException;
 import java.net.InetAddress;
 import java.net.InetSocketAddress;
 import java.net.NetworkInterface;
 import java.net.URI;
+import java.security.AccessController;
+import java.security.PrivilegedActionException;
+import java.security.PrivilegedExceptionAction;
 import java.util.ArrayList;
 import java.util.Enumeration;
 import java.util.List;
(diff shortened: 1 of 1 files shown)


protocol/src/main/java/org/jboss/as/protocol/ProtocolConnectionUtils.java
```diff
@@ -163,11 +167,36 @@ public static Connection connectSync(final ProtocolConnectionConfiguration confi
         final AuthenticationContext context = AuthenticationContext.empty().with(MatchRule.ALL, mergedConfiguration);
 
         if (clientBindAddress == null) {
-            return endpoint.connect(uri, options, context);
+            try {
+                return AccessController.doPrivileged((PrivilegedExceptionAction<IoFuture<Connection>>) () ->
+                        endpoint.connect(uri, options, context));
+            } catch (PrivilegedActionException pae) {
+                ProtocolLogger.CONNECTION_LOGGER.connectionPrivilegedAccessError(uri, pae);
```
dmlloyd: We don't normally both log *and* rethrow exceptions as that adds undue noise when the exception bubbles up again.

The PAE itself does not add useful information so it's better to just discard it completely and only throw the cause.

luck3y:
this is ok to test

luck3y:
retest this please

bstansberry:
@dmlloyd Is making this privileged valid? This is driven by https://issues.jboss.org/browse/WFLY-6936 where we have a deployment creating a ModelControllerClient inside the server. This is not something that screams "should be privileged" to me.

@iweiss Can't that test be fixed simply by making it @RunAsClient, making BindingLookupBean @Remote and having the test use it remotely? That's not a test that gains much by running in-container.

Hmm, I guess testRebindingObjectFactory would need some tweaking to use an ObjectFactory that produced something serializable. I haven't tried this but I bet the deployment could package an ObjectFactory that produced a string. The 'module' attribute would then be deployment.RebindTestCase.jar (see https://developer.jboss.org/thread/169894).

dmlloyd:
You need to have the "connect" ``EndpointPermission`` in order to create a Remoting connection.  That  said, if you're doing something privileged on behalf of the user, the user had better also be authorized, so either the permission check should be passed through (i.e. no doPrivileged) or else another permission check should be introduced.

bstansberry:
I don't want to add a new permission, e.g. to the remote ModelControllerClient impl. It's primarily a client-side class and that means people running a client in a security manager enabled process would need to alter their existing policy to account for a new permission. (Note that there is already a permission needed to access the ModelController in-vm and create an in-vm client from it; but that's a different thing from a client that connects over the network.)

I think the test should be fixed; either to @RunAsClient or to package the necessary perms in the deployment.

iweiss:
@dmlloyd @bstansberry Thank you for the comments, I will re-evaluate this PR

PR #2011, opened 2016-12-08
Deprecate old setSubsystemXmlMapping(..) method to drive users to a Supplier-based version instead

rhusar:
I think we should use deprecation to drive users into using a more efficient @ctomc 's version of registering XML mappings to readers. Given that this can be easily overlooked and also easily fixed in many cases with just adding "() ->". Is there any reason not to?


ctomc:
It wasn't deprecated as both are still valid scenarios.

bstansberry:
@ctomc What's the scenario the new way can't cover? If that's detailed here (as a form of documentation) then I'll close this.

Apologies if I should know and have just forgotten... and am too lazy to look. ;)

ctomc:
@bstansberry It is fine to merge it, I initially just didn't want to mark it as deprecated as the "old" way is still valid and we already have enough deprecation warnings as it is...

PR #2012, opened 2016-12-08
[WFCORE-2094] Domain server notifies HC upon controller instability

bstansberry:
This commit introduces a notification mechanism via which a listener for "controller instability" notifications can be created and wired into the ModelController.

In a domain server, such a listener is registered. In this case it's an additional function of HostControllerClient, which responds to notifications by asynchronously sending a message back to the managing HC.

Currently the managing HC simply logs a WARN when receiving such a message. In the future it could also do something else with the data, e.g. expose it over the management API for use by the admin console.

Notifications are currently sent in two cases:

1) An OperationContext is interrupted or times out waiting for MSC stability during operation finalization. This means MSC was not stable at the time of finalization and the state of the system is no longer knowable.
2) An Error is thrown during operation execution. Again in such a situation the state of the system is no longer knowable.

In the latter case, there is no guarantee that the listener will be able to send the message to the HC. For example if the Error is an OOME sending a message may not be possible. So this is not a foolproof means of ensuring the HC is aware of server instability. Possible future work to cover that kind of scenario could be to have the HC send some sort of health check request to the server when it detects that an operation has timed out.

This also doesn't send a notification if the server fails to shut down properly, e.g. from problems with the MSC ServiceContainer not stopping. That would be detected by the bootstrap shutdown hook, which has no relationship to this logic.

### DIFF
--- a/controller/src/main/java/org/jboss/as/controller/OperationContextImpl.java
+++ b/controller/src/main/java/org/jboss/as/controller/OperationContextImpl.java
@@ -1180,13 +1180,18 @@ final class OperationContextImpl extends AbstractOperationContext {
                 try {
                     modelController.awaitContainerStability(timeout, TimeUnit.MILLISECONDS, true);
                 }  catch (InterruptedException e) {
+                    // Cancelled in some way
                     interrupted = true;
-                    MGMT_OP_LOGGER.interruptedWaitingStability();
+                    // Cancellation prevented us ensuring that MSC will be stable when this operation completes.
+                    // We can no longer have any sense of MSC state or how the model relates to the runtime and
+                    // we need to start from a fresh service container. Notify the controller of this.
+                    modelController.containerCannotStabilize();
+                    MGMT_OP_LOGGER.interruptedWaitingStability(activeStep.operationId.name, activeStep.operationId.address);
                 } catch (TimeoutException te) {
                     // If we can't attain stability on the way out after rollback ops have run,
                     // we can no longer have any sense of MSC state or how the model relates to the runtime and
-                    // we need to start from a fresh service container.
-                    processState.setRestartRequired(); // don't use our restartRequired() method as this is not reversible in rollback
+                    // we need to start from a fresh service container. Notify the controller of this.
+                    modelController.containerCannotStabilize();
                     // Just log; this doesn't change the result of the op. And if we're not stable here
                     // it's almost certain we never stabilized during execution or we are rolling back and destabilized there.
                     // Either one means there is already a failure message associated with this op.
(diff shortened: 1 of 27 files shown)


bstansberry:
retest this please

bstansberry:
retest this please

controller/src/main/java/org/jboss/as/controller/OperationContextImpl.java
```diff
@@ -1180,13 +1180,18 @@ void releaseStepLocks(AbstractOperationContext.Step step) {
                 try {
                     modelController.awaitContainerStability(timeout, TimeUnit.MILLISECONDS, true);
                 }  catch (InterruptedException e) {
+                    // Cancelled in some way
                     interrupted = true;
-                    MGMT_OP_LOGGER.interruptedWaitingStability();
+                    // Cancellation prevented us ensuring that MSC will be stable when this operation completes.
+                    // We can no longer have any sense of MSC state or how the model relates to the runtime and
+                    // we need to start from a fresh service container. Notify the controller of this.
+                    modelController.containerCannotStabilize();
```
kabir: Should there be some kind of differentiation between this interrupted case, and the timeout case below?

host-controller/src/main/java/org/jboss/as/host/controller/ServerInventoryImpl.java
```diff
@@ -689,6 +689,20 @@ public void serverProcessStarted(final String serverProcessName) {
     }
 
     @Override
+    public void serverUnstable(final String serverProcessName) {
+        final String serverName = ManagedServer.getServerName(serverProcessName);
+        final ManagedServer server = servers.get(serverName);
+        boolean change = true;
```
kabir: Nitpick - s/change/unstable?

bstansberry: The variable reflects whether the 'unstable' report represents a 'change' from previous status and hence needs to be reported up the chain.

This is all about doing this part of the reporting once, and then not repeating each time.

controller/src/main/java/org/jboss/as/controller/OperationContextImpl.java
bstansberry: Originally I only did it for timeout. But what I saw is that in actual hang situations (where the timeout would trip), the HC's end up cancelling the op, resulting in this code path being the one that is followed. If we don't report here, then the report never happens. And it is true that if the op is allowed to complete before MSC stability is obtained, we no longer know the true state of the system.

I think the chances of 'false reporting' here are quite low. If MSC is unstable, rolling back probably won't stabilize it. (Instability is likely due to service start() or stop() not completing, and changes made by rollback, e.g. removing a service, don't repair that.) Perhaps in another PR I could tweak some of the timeout values to reduce the case where the HC cancels before the server times out here. And/or, I could add something where if another ops finds that MSC is stable again it triggers a new notification that "all is well".

bstansberry:
Kabir acked this in chat, so....

PR #2015, opened 2016-12-09
WFCORE-1374 Update build to work on jdk9 b148 or later

ctomc:


pom.xml
```diff
@@ -1605,41 +1606,49 @@
...
             <activation>
                 <jdk>9</jdk>
             </activation>
             <properties>
-                <!-- [WFCORE-1431] remove SASL workaround -->
-                <modular.jdk.args>--add-exports=java.security.sasl/com.sun.security.sasl.digest=ALL-UNNAMED
-                    --add-exports=java.security.sasl/com.sun.security.sasl=ALL-UNNAMED
-                    --add-exports=jdk.unsupported/sun.reflect=ALL-UNNAMED
-                    --add-exports=jdk.unsupported/sun.misc=ALL-UNNAMED
+                <modular.jdk.args>
```
ropalka: This change looks good but we need to keep:
--add-exports=java.base/sun.nio.ch=ALL-UNNAMED

ctomc: re-added

pom.xml
```diff
@@ -1605,41 +1606,50 @@
...
+                    --add-opens=java.base/java.util=ALL-UNNAMED
+                    --add-opens=java.base/java.util.concurrent=ALL-UNNAMED
+                    --add-opens=java.base/jdk.internal.reflect=ALL-UNNAMED
+                    --add-opens=java.management/javax.management=ALL-UNNAMED
+                    --add-opens=java.management/javax.management.openmbean=ALL-UNNAMED
                 </modular.jdk.args>
                 <!-- [WFCORE-1494] remove JUL workaround -->
-                <modular.jdk.props>-Dsun.util.logging.disableCallerCheck=true</modular.jdk.props>
+                <modular.jdk.props>-Dsun.util.logging.disableCallerCheck=true -Dsun.reflect.debugModuleAccessChecks=true</modular.jdk.props>
+                <version.org.jboss.byteman>4.0.0-ALPHA</version.org.jboss.byteman>
```
bstansberry: IMO this is ok because this profile only provides a mechanism to trigger an experimental build, and is not part of the standard build of WildFly Core or of any project based on core. So there is no need for any such project to consider Byteman 4 in its own build requirements. For example, there is no need to productize byteman 4.

PR #2016, opened 2016-12-09
[WFCORE-1890]: Replacing a scanner deployment throught the web console will crash the next server reboot.

ehsavoie:
When replacing a scanner deployment using the (web) console won't create the undeployed marker file in the scanner.
Using notifications to make sure of that.

Jira: https://issues.jboss.org/browse/WFCORE-1890
JBEAP: https://issues.jboss.org/browse/JBEAP-7992

### DIFF
--- a/testsuite/manualmode/src/test/java/org/jboss/as/test/manualmode/deployment/DeploymentScannerUnitTestCase.java
+++ b/testsuite/manualmode/src/test/java/org/jboss/as/test/manualmode/deployment/DeploymentScannerUnitTestCase.java
@@ -21,7 +21,10 @@ import static org.jboss.as.controller.descriptions.ModelDescriptionConstants.ADD
 import static org.jboss.as.controller.descriptions.ModelDescriptionConstants.CONTENT;
 import static org.jboss.as.controller.descriptions.ModelDescriptionConstants.DEPLOY;
 import static org.jboss.as.controller.descriptions.ModelDescriptionConstants.DEPLOYMENT;
+import static org.jboss.as.controller.descriptions.ModelDescriptionConstants.ENABLED;
+import static org.jboss.as.controller.descriptions.ModelDescriptionConstants.FULL_REPLACE_DEPLOYMENT;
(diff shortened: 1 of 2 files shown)


ehsavoie:
retest this please

ehsavoie:
retest this please

bstansberry:
There was a scanner test case failure when I put this on master-ignore:

https://ci.wildfly.org/viewLog.html?buildId=42457&tab=buildResultsDiv&buildTypeId=WildFlyCore_MasterIgnorWindows

I notice that FileSystemDeploymentService handleNotification drops the notification if it can't acquire the scan lock. So if the scanner happens to be running when an op runs, expected behavior might not occur.

testsuite/manualmode/src/test/java/org/jboss/as/test/manualmode/deployment/DeploymentScannerUnitTestCase.java
```diff
@@ -222,6 +225,47 @@ public void testFailedDeploymentWithPersistentDeployment() throws Exception {
...
+                    Path deployedMarker = deployDir.resolve(JAR_ONE + ".deployed");
+                    waitFor(String.format("Missing .deployed marker for %s", JAR_ONE),
+                            () -> Files.exists(deployedMarker));
+                    Assert.assertTrue(String.format("%s should be deployed", JAR_ONE), exists(persistentDeploymentAddress));
+                    //Replace deployment
+                    Archive<?> validDeployment = createDeploymentArchive();
+                    replaceWithPersistent(JAR_ONE, validDeployment);
+                    Assert.assertTrue(String.format("%s should be deployed", JAR_ONE), exists(persistentDeploymentAddress));
+                    Assert.assertTrue(String.format("Missing .undeployed marker for %s", JAR_ONE),
+                            Files.exists(deployDir.resolve(JAR_ONE + ".undeployed")));
```
bstansberry: We discussed this in chat; I'm just putting a comment here in case others review the PR. The management notification to the scanner is async to the management op that emits it so the test can't assume it's received and handled immediately after the management op completes. There needs to be some sort allowance for a possible delay.

ehsavoie:
retest this please

PR #2019, opened 2016-12-10
WFCORE-2105 Raise log level of deprecations to WARN

marschall:
We believe the log level of log level of deprecations should be WARN
instead of INFO. We were using a deprecated attribute and almost didn't
see the message in the see of INFO messages.

This commit changes the level of the following messages to WARN

 * ControllerLogger#attributeDeprecated
 * ControllerLogger#operationDeprecated

https://issues.jboss.org/browse/WFCORE-2105


marschall:
> java.net.BindException: Address already in use: bind

Is that really caused by this PR?

ctomc:
@marschall no, there are some intermittent testsuite issues following recent upgrade to remoting 5 and elytron. Fixes are being worked on.

bstansberry:
The use of INFO was deliberate, as some users may have systems in place that trigger disruptive consequences in reaction to a WARN. A deprecation is not something that requires immediate attention.

marschall:
@bstansberry fair enough, should I close the PR and JIRA issue?

bstansberry:
Yes, please.

PR #2020, opened 2016-12-10
WFCORE-2106 Allow expressions in relative-to of keystore

marschall:
All attributes on the keystore element including the path element
support expressions with the exception of the relative-to element.
This seem inconsistent and forces us to have a more complicated than
necessary expression in the path element instead of being able to split
it between the path and relative-to element.

https://issues.jboss.org/browse/WFCORE-2106


darranl:
Relative to is a service dependency which means we do not support expressions for this attribute.

marschall:
In that case

PR #2040, opened 2016-12-20
[WFCORE-1606]: Server reload is needed for modified security-realm even if {allow-resource-service-restart=true} is used

ehsavoie:
Updating metadata and handlers to require for server reload as restarting the SecurityRealm is not sufficient.

Jira: https://issues.jboss.org/browse/WFCORE-1606
JBEAP: https://issues.jboss.org/browse/JBEAP-5073
Full 'fix':  https://github.com/wildfly/wildfly/pull/10097 
Experimental build https://ci.wildfly.org/viewType.html?buildTypeId=WF_WildFlyCoreIntegrationExperiments&branch_WF=%2Fpre-WFCORE-1606%3B%2Fehsavoie%3A%2FWFCORE-1606%3B%2Fehsavoie

### DIFF
--- a/domain-management/src/main/java/org/jboss/as/domain/management/security/SecurityRealmChildAddHandler.java
+++ b/domain-management/src/main/java/org/jboss/as/domain/management/security/SecurityRealmChildAddHandler.java
@@ -31,6 +31,7 @@ import org.jboss.as.controller.PathAddress;
 import org.jboss.as.controller.capability.RuntimeCapability;
 import org.jboss.as.controller.registry.Resource;
 import org.jboss.dmr.ModelNode;
+import org.jboss.msc.service.ServiceController;
 
 /**
  * Add handler for a child resource of a management security realm.
@@ -53,6 +54,11 @@ public class SecurityRealmChildAddHandler extends SecurityRealmParentRestartHand
         this.validateAuthentication = validateAuthentication;
         this.validateAuthorization = validateAuthorization;
         this.attributeDefinitions = attributeDefinitions == null ? new AttributeDefinition[0] : attributeDefinitions;
+    }
+
+    @Override
+    protected boolean isResourceServiceRestartAllowed(OperationContext context, ServiceController<?> service) {
+        return false;
     }
 
     @Override
--- a/domain-management/src/main/java/org/jboss/as/domain/management/security/SecurityRealmChildRemoveHandler.java
+++ b/domain-management/src/main/java/org/jboss/as/domain/management/security/SecurityRealmChildRemoveHandler.java
@@ -32,6 +32,7 @@ import org.jboss.as.controller.registry.AttributeAccess;
 import org.jboss.as.controller.registry.ImmutableManagementResourceRegistration;
 import org.jboss.as.controller.registry.Resource;
 import org.jboss.dmr.ModelNode;
+import org.jboss.msc.service.ServiceController;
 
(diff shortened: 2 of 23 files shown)


bstansberry:
retest this please

bstansberry:
Three times when I've tried to put this on master-ignore the test against full failed with the log showing issues related to adding stuff to a security realm:

https://ci.wildfly.org/viewLog.html?buildId=42937&tab=buildResultsDiv&buildTypeId=WF_WildFlyCoreMasterIgnoreIntegration

So I think there's a problem here.

ehsavoie:
The issue is that we don't reload the server after adding the security realm, relying on the ALLOW_RESOURCE_SERVICE_RESTART. Reloading the server in WSTrustTestCaseSecuritySetupTask would fix the issue.

ctomc:
retest this please

domain-management/src/main/java/org/jboss/as/domain/management/security/SecurityRealmChildAddHandler.java
```diff
@@ -56,6 +57,11 @@ public SecurityRealmChildAddHandler( boolean validateAuthentication, boolean val
     }
 
     @Override
+    protected boolean isResourceServiceRestartAllowed(OperationContext context, ServiceController<?> service) {
+        return false;
+    }
```
bstansberry: @darranl WDYT about this:

I don't think this should override the SecurityRealmParentRestartHandler impl. Instead, lets change ManagementUtil.isSecurityRealmReloadRequired so it ignores context.isResourceServiceRestartAllowed and returns true if the service is present and UP. Then in SecurityRealmAddHandler L219 we change it to

```
realmBuilder.setInitialMode(Mode.ON_DEMAND);
```

PRO: 
1) People can still build up a realm without needing a composite or a reload, since until it is demanded, child add ops will reinstall the service. So this may avoid breaking some scripts.
2) We don't start undemanded services.

CON: The realm service isn't actively started so we don't aggressively detect problems with it. They are discovered later when the admin adds something that demands it.

domain-management/src/main/java/org/jboss/as/domain/management/security/SecurityRealmChildRemoveHandler.java
```diff
@@ -52,6 +53,11 @@ public SecurityRealmChildRemoveHandler(boolean validateAuthentication, RuntimeCa
     }
 
     @Override
+    protected boolean isResourceServiceRestartAllowed(OperationContext context, ServiceController<?> service) {
+        return false;
+    }
```
bstansberry: Same comment as for SecurityRealmChildAddHandler.

domain-management/src/main/java/org/jboss/as/domain/management/security/SecurityRealmChildAddHandler.java
ehsavoie: @darranl confirmed what I found experimenting: there are some services that are not declaring a dependency on the Security realm and are expecting it to be up.

bstansberry: Ok, then -1 to my own idea.

bstansberry:
retest this please

ehsavoie:
retest this please

PR #2044, opened 2017-01-02
Fix for WFCORE-2142. Fix unknown overlay error messages.

jfdenise:
Check the overlay name prior to make a request. In batch mode the name is only checked for actions that are not added to the batch (list-links, list-content).
The side effect is that upload action can now be added to a batch if the overlay doesn't exist (overlay could be added in the same batch, in a previous step).
Unit tests added.


bstansberry:
@aloubyansky please approve

aloubyansky:
@bstansberry looks good!

PR #2049, opened 2017-01-05
[WFCORE-846] Some tests throws "Permission check failed" while run with security manager

iweiss:
Issue: https://issues.jboss.org/browse/WFCORE-846

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -80,6 +80,7 @@
             For example: <version.org.jboss.hal.release-stream>
          -->
 
+        <version.com.io7m.xom>1.2.10</version.com.io7m.xom>
         <version.commons-io>2.5</version.commons-io>
         <version.commons-lang>2.6</version.commons-lang>
         <version.io.undertow>1.4.8.Final</version.io.undertow>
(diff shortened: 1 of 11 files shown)


iweiss:
retest this please

testsuite/shared/pom.xml
```diff
@@ -165,6 +165,21 @@
             <artifactId>byteman</artifactId>
             <scope>provided</scope>
         </dependency>
+        <dependency>
+            <groupId>com.io7m.xom</groupId>
+            <artifactId>xom</artifactId>
+            <version>1.2.10</version>
```
bstansberry: Use dependency-management in the root pom to specify versions, driven by a new version.xxxx property.

bstansberry:
CI failures are related.

pom.xml
```diff
@@ -143,6 +143,7 @@
         <!-- Non-default maven plugin versions and configuration -->
         <version.org.zanata.plugin>3.9.1</version.org.zanata.plugin>
         <version.xml.plugin>1.0.1</version.xml.plugin>
+        <version.com.io7m.xom>1.2.10</version.com.io7m.xom>
```
bstansberry: These are kept in alphabetical order.

bstansberry:
I suspect these failures are due to the same sort of thing you already fixed in a couple test classes so the fix would be to apply the same thing.

iweiss:
@bstansberry That's correct. The deployment test issues are fixed on another PR, IIRC.

iweiss:
retest this please

PR #2056, opened 2017-01-08
[WFCORE-2126] Update Undertow to latest version (1.4.11)

langfr:


langfr:
retest this please

bstansberry:
@stuartwdouglas is this ok in a community micro release (2.2.1.Final)?

There are no EAP implications.

bstansberry:
retest this please

I'm concerned the CI failures are related as both the Windows and Linux jobs showed the same failures and a quick glance at one failure showed it was somewhat HTTP-related; i.e. a response was a 403 instead of a 200.

jamezp:
retest this please

famod:
```IdentityLoginModuleTestCase``` will fail as soon as Undertow is updated to 1.4.1 or above.
In fact, there are a couple of security related fixes in the [1.4.1 changelog](https://issues.jboss.org/secure/ConfigureReport.jspa?versions=12331199&sections=all&style=none&selectedProjectId=12312521&reportKey=org.jboss.labs.jira.plugin.release-notes-report-plugin%3Areleasenotes&atl_token=AQZJ-FV3A-N91S-UDEU%7C0c93c673989baad3cffff3a89cea032b92f6df44%7Clin&Next=Next).

I guess the test needs to be adjusted. I'll see what I can find out.
Maybe @langfr can also have a look...

langfr:
Strange. IdentityLoginModuleTestCase is not part of WFCORE but part of WFLY.
How to reproduce and test this locally?

This commit https://github.com/wildfly/wildfly/commit/7cd6d9fda7cae808685f7664b6655eab3cd27522 on master looks related.
I'll prepare a PR to bring this change to 10.x branch too.

famod:
My assumption was correct and luckily the test has already been fixed in master of ```wildfly``` which builds on top of ```wildfly-core``` 3.x and in turn uses a later Undertow version: https://github.com/wildfly/wildfly/commit/7cd6d9fda7cae808685f7664b6655eab3cd27522
Quote from https://issues.jboss.org/browse/WFLY-7063:
> This test assumes that the security path of '/' will be a wildcard match, while in reality it should only match the root url. This test only passed due to UNDERTOW-805

So all there is to do is to backport this commit to 10.x branch of ```wildfly``` which can be done right away (without waiting for this PR to be integrated).

famod:
@langfr
I tested it this way:
`call mvn clean install -DskipTests -T2 -f wildfly-core\pom.xml && call mvn clean install -DallTests -DskipTests -T2 -f wildfly\pom.xml && call mvn clean install -f wildfly\testsuite\integration\basic\pom.xml -Dtest=IdentityLoginModuleTestCase`
You can also split it into three different calls.
I run it from a directory which has ```wildfly-core``` 2.x and ```wildfly``` 10.x checked out next to each other (in separate subfolders of course).

Please let me know in case you need any futher help regarding the backport PR. I am new to hacking on WildFly but I will try my best.

bstansberry:
retest this please

bstansberry:
(Note: I've deleted a lot of CI failure comments above as they are too noisy and hide the discussion by humans. Much of the discussion above and the 'retest' comments are about failures like you see at https://ci.wildfly.org/viewLog.html?buildId=48349&buildTypeId=WildFlyCore_PullRequest2x_WildFlyCoreFullIntegration)

PR #2057, opened 2017-01-08
[WFCORE-1620 2.x] Update slf4j to 1.7.22

langfr:


bstansberry:
@jamezp is this ok in a community micro release (2.2.1.Final)?

There are no EAP implications.

PR #2062, opened 2017-01-10
Add a test of deploying interdependent deployments

bstansberry:
People are deploying interrelated sets of deployments sometimes in large graphs, so let's test that some.


bstansberry:
I removed the Ready for Merge as until the related fixes come in this test may fail intermittently.

bstansberry:
Superceded by #2099

PR #2064, opened 2017-01-10
[WFCORE-2171] Avoid limitless nesting of Collections.UnmodifiableSet …

bstansberry:
…as the resource tree is cloned


bstansberry:
Stuart acked this same commit on #2062

PR #2073, opened 2017-01-11
[WFCORE-2179]: When in SUSPENDED state we shouldn't suspend on stopping

ehsavoie:
If in state SUSPENDED when suspend is called, just return.
Jira: https://issues.jboss.org/browse/WFCORE-2179

Please note that this will impact the RunningState listeners (as some transition won't happen anymore)


bstansberry:
"Please note that this will impact the RunningState listeners (as some transition won't happen anymore)" -- that's the goal, right?

ehsavoie:
Yes forgot the word "tests" in the note. The tests will fail after merging this one.

bstansberry:
Superceded by #2110

PR #2081, opened 2017-01-12
[WFCORE-2189]: JmxControlledStateNotificationsTestCase fails on IBM JDK (domain module).

ehsavoie:
Base path is not resolved in the same way as with the Oracle JDK.

Jira: https://issues.jboss.org/browse/WFCORE-2189

IBM build: https://ci.wildfly.org/viewLog.html?buildId=40845


bstansberry:
Next time we chat, please explain this one to me as I'm curious. :) I kind of get why the change for IBM is needed, as the utility code that launches the domain sets the working directory of the PC to target/domains/XXX/master and presumably that carries through to the servers. I don't understand why the same isn't the case for other JVMs. Is it because that working directory *doesn't* carry through to the servers on the other JVMs?

PR #2084, opened 2017-01-12
WFCORE-2192 Change deployment restart on missing dependency to happen…

stuartwdouglas:
… in the phase stop()

Previously this happened on next start, however this could lead to patalogical situations
if optional dependencies were in use where deployments would continually restart. After
this change when a deployment has missing dependencies it will immediatly attempt to
restart, and will pause at the point that it requires the missing dependencies

### DIFF
--- a/server/src/main/java/org/jboss/as/server/deployment/DeploymentUnitPhaseService.java
+++ b/server/src/main/java/org/jboss/as/server/deployment/DeploymentUnitPhaseService.java
@@ -27,7 +27,6 @@ import java.util.LinkedList;
 import java.util.List;
 import java.util.ListIterator;
 import java.util.Set;
-import java.util.concurrent.atomic.AtomicBoolean;
 
 import org.jboss.as.server.logging.ServerLogger;
 import org.jboss.msc.service.AbstractServiceListener;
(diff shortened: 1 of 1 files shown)


server/src/main/java/org/jboss/as/server/deployment/DeploymentUnitPhaseService.java
```diff
@@ -208,19 +171,34 @@ public void transition(final ServiceController<?> controller, final ServiceContr
...
+                public void transition(final ServiceController<?> controller, final ServiceController.Transition transition) {
+                    if(transition.getAfter().equals(ServiceController.Substate.DOWN)) {
+                        //its possible an undeploy happened in the meantime
+                        //so we use compareAndSetMode to make sure the mode is still NEVER and not REMOVE
+                        controller.compareAndSetMode(Mode.NEVER, Mode.ACTIVE);
+                        controller.removeListener(this);
+                    }
+                }
+            });
+            topController.setMode(Mode.NEVER);
```
bstansberry: I think this needs to be a CAS as well or again it can race with an undeploy op. It perhaps could be moved to the if test.

ppalaga:
@stuartwdouglas based on Brian's comment, I assume there are more changes to be expected here. If so, could you please submit them, so that I can backport to 6.4? The downstream [BZ1412833](https://bugzilla.redhat.com/show_bug.cgi?id=1412833) was recently scheduled for 6.4.CP14.

ppalaga:
This PR seems to have been superseded by #2095

PR #2087, opened 2017-01-14
[WFCORE-2182] Don't treat vault 'lookup misses' as a SecurityException

bstansberry:
https://issues.jboss.org/browse/WFCORE-2182
https://issues.jboss.org/browse/WFCORE-2198
https://issues.jboss.org/browse/WFCORE-2199

Clean up handling of the vault to better distinguish types of problems. Distinguish requests for non-existent vault data from technical problems with the vault. For vault reads by the management operation handers, convert failures due to non-existing vault data to OperationFailedException as these are client input data problems.

VaultReader will still throw a runtime exception for non-existent vault data, so other uses of VaultReader (i.e. VaultPropertyResolver used in deployment descriptor parsing in full) will still fail in the way they did in the past (which hasn't been reported as a problem), so this is a compatible change for that usage. The RuntimeException is no longer SecurityException, but the particular flavor of runtime exception was never relied on or part of a contract.

In a perfect world requests for non-existent vault data might not be handled via an exception, but rather by VaultReader returning null. But that would lead to problems with the deployment descriptor expression resolution in full, as returning null from VaultPropertyResolver would result in the next property resolver running, which would likely treat the first ':' in the vault expression as the separator between a system property name and a default value, returning everything after that ':'. Which is wrong. So to keep it simple I stuck with the approach of throwing an exception.

Also includes WFCORE-2199, a further improvement in this area for cases where vault expressions are used when PicketBox isn't even present, and WFCORE-2198 which is necessary to get proper testing of the vault functionality in core.


bstansberry:
retest this please

bstansberry:
retest this please

bstansberry:
Failure is unrelated; one of the intermittent ones that has been turning up since the first big elytron/remoting 5 integration.

server/src/test/java/org/jboss/as/server/services/security/VaultHandler.java
```diff
@@ -0,0 +1,377 @@
...
+        }
+
+        this.encodedVaultFileDirectory = vaultDirectory.getAbsolutePath();
+
+        if (salt == null) {
+            String tmp = Long.toHexString(System.currentTimeMillis())
+                    + Long.toHexString(System.currentTimeMillis())
+                    + Long.toHexString(System.currentTimeMillis())
+                    + Long.toHexString(System.currentTimeMillis());
+            this.salt = tmp.substring(0, 8);
```
jamezp: I'm not sure the concatenation is needed if we're just using the first 8 characters of the string. The first `Long.toHexString(System.currentTimeMillis())` will always generate a hex string longer than 8 characters.

jamezp: Though since it's a test it probably doesn't really matter.

server/src/test/java/org/jboss/as/server/services/security/VaultSession.java
```diff
@@ -0,0 +1,348 @@
...
+     * @param attributeName
+     */
+    private void attributeCreatedDisplay(String vaultBlock, String attributeName) {
+        System.out.println(String.format("Secured attribute value has been stored in Vault.\n" +
+            "Please make note of the following:\n" +
+            "********************************************\n" +
+            "Vault Block:%s\n" + "Attribute Name:%s\n" +
+            "Configuration should be done as follows:\n" +
+            "%s\n" +
+            "********************************************", vaultBlock, attributeName, securedAttributeConfigurationString(vaultBlock, attributeName)));
```
jamezp: You could remove the `String.format` and just use `System.out.printf`. The `\n` should probably be replaced with `%n` to be platform independent.

server/src/test/java/org/jboss/as/server/services/security/VaultHandler.java
bstansberry: This class and VaultSession are just copies of the same code in testsuite/shared.

PR #2090, opened 2017-01-16
Fix for WFCORE-2186. Handle bytes type in completion and execution.

jfdenise:
Evolved parser and completer to support bytes syntax. Added unit test.


bstansberry:
@aloubyansky please review

aloubyansky:
Looks good to me, thanks!

PR #2091, opened 2017-01-16
WFCORE-2188 WFCORE-2159 WFCORE-2187 WFCORE-1622 WFCORE-2195 IO subsystem fixes

ctomc:
- WFCORE-2187 Description of IO worker task-max-threads attribute is inaccurate
- WFCORE-1622 Worker section in IO subsystem does not allow expressions for io-threads and task-max-threads attributes
- WFCORE-2159 worker=default:read-resource(include-runtime) wrong values in domain mode
- WFCORE-2188 IO subsystem requires reload when increasing size of io-threads or task-max-threads
- WFCORE-2195 XSD of IO subsystem should contain description of all resources and attributes
- fix how undefined metrics
-  WFCORE-2269 Add transformers for IO subsystem

### DIFF
--- a/io/subsystem/src/main/java/org/wildfly/extension/io/IOSubsystemTransformers.java
+++ b/io/subsystem/src/main/java/org/wildfly/extension/io/IOSubsystemTransformers.java
@@ -1,0 +1,68 @@
+/*
+ * JBoss, Home of Professional Open Source
+ * Copyright 2017, Red Hat, Inc., and individual contributors as indicated
+ * by the @authors tag.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.wildfly.extension.io;
+
+import org.jboss.as.controller.ModelVersion;
+import org.jboss.as.controller.transform.ExtensionTransformerRegistration;
+import org.jboss.as.controller.transform.SubsystemTransformerRegistration;
+import org.jboss.as.controller.transform.description.AttributeConverter;
+import org.jboss.as.controller.transform.description.ChainedTransformationDescriptionBuilder;
+import org.jboss.as.controller.transform.description.RejectAttributeChecker;
+import org.jboss.as.controller.transform.description.ResourceTransformationDescriptionBuilder;
+import org.jboss.as.controller.transform.description.TransformationDescriptionBuilder;
+
+/**
+ * @author Tomaz Cerar (c) 2017 Red Hat Inc.
+ */
+public class IOSubsystemTransformers implements ExtensionTransformerRegistration {
+    static final ModelVersion VERSION_2_0 = ModelVersion.create(2, 0);
+
+
+    @Override
+    public String getSubsystemName() {
+        return IOExtension.SUBSYSTEM_NAME;
+    }
+
+    @Override
+    public void registerTransformers(SubsystemTransformerRegistration registration) {
+        ChainedTransformationDescriptionBuilder chainedBuilder = TransformationDescriptionBuilder.Factory.createChainedSubystemInstance(registration.getCurrentSubsystemVersion());
+
+        // Current 3.0.0 to 2.0.0, aka EAP 7.0.0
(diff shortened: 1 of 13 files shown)


io/subsystem/src/main/java/org/wildfly/extension/io/WorkerResourceDefinition.java
```diff
@@ -157,20 +194,52 @@ public void registerChildren(ManagementResourceRegistration resourceRegistration
...
+        @Override
+        protected boolean applyUpdateToRuntime(OperationContext context, ModelNode operation, String attributeName, ModelNode value, ModelNode currentValue, HandbackHolder handbackHolder) throws OperationFailedException {
+            XnioWorker worker = getXnioWorker(context);
+            if (worker == null) { //worker can be null if it is not started yet, it can happen when there are no dependencies to it.
+                return true;
+            }
+            try {
+                return setValue(worker, value);
+            } catch (IOException e) {
+                throw new OperationFailedException(e);
```
bstansberry: I think this should be some sort of RuntimeException, unless the IOE means the user made a mistake. OFE is for user errors, not internal server faults.

io/subsystem/src/main/java/org/wildfly/extension/io/WorkerResourceDefinition.java
```diff
@@ -157,20 +194,52 @@ public void registerChildren(ManagementResourceRegistration resourceRegistration
...
+            }
+        }
+
+        @Override
+        protected void revertUpdateToRuntime(OperationContext context, ModelNode operation, String attributeName, ModelNode valueToRestore, ModelNode valueToRevert, Object handback) throws OperationFailedException {
+
```
bstansberry: Why no impl?

io/subsystem/src/main/java/org/wildfly/extension/io/WorkerResourceDefinition.java
```diff
@@ -77,7 +77,7 @@
             .build();
     static final OptionAttributeDefinition WORKER_TASK_KEEPALIVE = new OptionAttributeDefinition.Builder(Constants.WORKER_TASK_KEEPALIVE, Options.WORKER_TASK_KEEPALIVE)
             .setFlags(AttributeAccess.Flag.RESTART_ALL_SERVICES)
-            .setDefaultValue(new ModelNode(60))
+            .setDefaultValue(new ModelNode(60_000))
```
bstansberry: Transfomer. :(

io/subsystem/src/main/java/org/wildfly/extension/io/WorkerResourceDefinition.java
ctomc: In most cases this would throw error if value of attribute is not correct or not proper in current context. So in a way it is a user mistake.

io/subsystem/src/main/java/org/wildfly/extension/io/WorkerResourceDefinition.java
ctomc: Tricky one. As 60 was mistake in the unit. 
created https://issues.jboss.org/browse/WFCORE-2269 to add transformers

io/subsystem/src/main/java/org/wildfly/extension/io/WorkerResourceDefinition.java
ctomc: fixed

io/subsystem/src/main/java/org/wildfly/extension/io/WorkerResourceDefinition.java
bstansberry: This needs to be fixed as part of this PR.

It's not tricky, it's a standard change in a default value. If the user submits undefined we send 60000 to the slave, not undefined. So the legacy servers behave in a manner consistent with the current ones.

ctomc: added transformers. They ware also needed for expressions that ware added as part of this commit.

io/subsystem/src/main/java/org/wildfly/extension/io/IOSubsystemTransformers.java
```diff
@@ -0,0 +1,68 @@
...
+        // Current 3.0.0 to 2.0.0, aka EAP 7.0.0
+        buildTransformers_2_0(chainedBuilder.createBuilder(registration.getCurrentSubsystemVersion(), VERSION_2_0));
+
+        chainedBuilder.buildAndRegister(registration, new ModelVersion[]{VERSION_2_0});
+    }
+
+    private void buildTransformers_2_0(ResourceTransformationDescriptionBuilder builder) {
+        builder.addChildResource(WorkerResourceDefinition.INSTANCE.getPathElement()).getAttributeBuilder()
+                .setValueConverter(
+                        new AttributeConverter.DefaultValueAttributeConverter(WorkerResourceDefinition.WORKER_TASK_KEEPALIVE),
```
bstansberry: Nice; I hadn't noticed this one or had forgotten about it.

PR #2093, opened 2017-01-17
WFCORE-2201 fix server side exception when remove read-only /path=<x>

soul2zimate:
please see https://issues.jboss.org/browse/WFCORE-2201


bstansberry:
Thanks! I commented on the JIRA but I see now you'd fixed it hours before I commented. :)

PR #2095, opened 2017-01-17
WFCORE-2192 Change deployment restart on missing dependency to happen…

stuartwdouglas:
… in the phase stop()

For now this only takes effect if the system property 'org.jboss.msc.directionalExecutor' is set

Previously this happened on next start, however this could lead to patalogical situations
if optional dependencies were in use where deployments would continually restart. After
this change when a deployment has missing dependencies it will immediatly attempt to
restart, and will pause at the point that it requires the missing dependencies.

### DIFF
--- a/server/src/main/java/org/jboss/as/server/deployment/DeploymentUnitPhaseService.java
+++ b/server/src/main/java/org/jboss/as/server/deployment/DeploymentUnitPhaseService.java
@@ -67,6 +67,24 @@ final class DeploymentUnitPhaseService<T> implements Service<T> {
      */
     private final AtomicBoolean runOnce = new AtomicBoolean();
 
+    /**
+     * If this is true then when a deployment goes down due to a dependency restart it will immediately attempt redeployment,
+     * otherwise it will wait till the dependency is available before doing the restart.
+     *
+     * At present this behaviour has a high chance of hitting MSC race conditions, that are only prevented at this stage
+     * by using the new directional executor. As this executor is only enabled via the <code>org.jboss.msc.directionalExecutor</code>
+     * system property this new restart behaviour is also only enabled if this system property is set.
+     *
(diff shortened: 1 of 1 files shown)


bstansberry:
@ppalaga @gaol This change needs to be paired with an MSC upgrade. So it should not be merged into EAP branches without the related upgrade.

ppalaga:
@bstansberry you probably mean the upgrade to MSC 1.2.7.SP1 tracked by https://issues.jboss.org/browse/WFCORE-2208 that incorporates just https://issues.jboss.org/browse/MSC-159, right? 

MSC-159 does not seem to have a PR yet. Who is working on it?

bstansberry:
@ppalaga MSC-159 is resolved; 1.2.7.SP1 is now released, as is a 1.1.7.SP1 suitable for EAP 6.4.x. I just sent up a PR for the WildFly Core upgrade to 1.2.7.SP1.

I'm also going to rework https://github.com/wildfly/wildfly-core/pull/2062 a bit to move the test to testsuite/manualmode as having these fixes take effect requires setting a system property as server launch. So a manual launch.

ppalaga:
Thanks for the info, @bstansberry

bstansberry:
Superceded by #2099 which is merged.

@ppalaga FYI https://github.com/jbossas/jboss-eap/commit/a5054ce17142981cfdf31cd19cd742b7c459704c is a variant of the 3rd commit on #2099, but ported to 6.4.x. But I don't have time to do a backport of the 4th commit, which moves the test to the manualmode testsuite and tweaks things to ensure the server is started with the org.jboss.msc.directionalExecutor system property. The way manualmode works in 6.4.x is different so it's not just a cherry-pick backport.

I don't think backporting the test is really required; it's far from a 100% reproducer of the problem anyway.

PR #2096, opened 2017-01-18
credential-reference changes with regards to mail subsystem integration

pskopek:
https://issues.jboss.org/browse/WFCORE-2210
https://issues.jboss.org/browse/WFCORE-2211
Fixed issues of credential-reference marshalling using standard marshaller.
Removed Destroyable interface implementation as it is not needed any more.

### DIFF
--- a/controller/src/main/java/org/jboss/as/controller/security/CredentialReference.java
+++ b/controller/src/main/java/org/jboss/as/controller/security/CredentialReference.java
@@ -17,10 +17,10 @@
  */
 package org.jboss.as.controller.security;
 
-import javax.xml.stream.XMLStreamException;
-import javax.xml.stream.XMLStreamWriter;
+import java.io.IOException;
+import java.security.spec.AlgorithmParameterSpec;
+import java.util.StringTokenizer;
 
-import org.jboss.as.controller.AttributeDefinition;
 import org.jboss.as.controller.AttributeMarshaller;
 import org.jboss.as.controller.AttributeParser;
 import org.jboss.as.controller.ObjectTypeAttributeDefinition;
@@ -35,7 +35,6 @@ import org.jboss.dmr.ModelType;
 import org.jboss.msc.service.ServiceBuilder;
 import org.jboss.msc.service.ServiceName;
 import org.jboss.msc.value.InjectedValue;
-import org.jboss.staxmapper.XMLExtendedStreamReader;
 import org.wildfly.common.function.ExceptionSupplier;
 import org.wildfly.security.auth.SupportLevel;
 import org.wildfly.security.credential.Credential;
@@ -45,10 +44,6 @@ import org.wildfly.security.credential.source.CredentialSource;
 import org.wildfly.security.credential.source.CredentialStoreCredentialSource;
 import org.wildfly.security.credential.store.CredentialStore;
 import org.wildfly.security.password.interfaces.ClearPassword;
-
-import java.io.IOException;
-import java.security.spec.AlgorithmParameterSpec;
-import java.util.StringTokenizer;
 
 /**
  * Utility class holding attribute definitions for credential-reference attribute in the model.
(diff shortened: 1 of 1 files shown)


bstansberry:
Please clone JBEAPs into WFCORE before working them.

controller/src/main/java/org/jboss/as/controller/security/CredentialReference.java
```diff
@@ -169,12 +169,21 @@ private static AttributeMarshaller credentialReferenceAttributeMarshaller() {
...
+                    crNode = credentialReferenceModelNode.get(attribute.getName());
                 } else {
-                    credentialStoreAttribute.marshallAsAttribute(credentialReferenceModelNode, writer);
-                    credentialAliasAttribute.marshallAsAttribute(credentialReferenceModelNode, writer);
-                    credentialTypeAttribute.marshallAsAttribute(credentialReferenceModelNode, writer);
+                    crNode = credentialReferenceModelNode;
+                }
+                if (!crNode.isDefined()) {
+                    throw new XMLStreamException(ControllerLogger.ROOT_LOGGER.invalid(attribute.getName()));
+                }
```
bstansberry: If this is possible something has been allowed previously that should have been detected and rejected. We shouldn't be doing model validation during xml marshaling.

If this shouldn't be possible and this is the equivalent of an assert, perhaps use an assert.

controller/src/main/java/org/jboss/as/controller/security/CredentialReference.java
```diff
@@ -169,12 +168,21 @@ private static AttributeMarshaller credentialReferenceAttributeMarshaller() {
...
-                if (credentialReferenceModelNode.hasDefined(clearTextAttribute.getName())) {
-                    clearTextAttribute.marshallAsAttribute(credentialReferenceModelNode, writer);
+                ModelNode crNode = null;
+                if (credentialReferenceModelNode.hasDefined(attribute.getName())) {
+                    crNode = credentialReferenceModelNode.get(attribute.getName());
                 } else {
-                    credentialStoreAttribute.marshallAsAttribute(credentialReferenceModelNode, writer);
-                    credentialAliasAttribute.marshallAsAttribute(credentialReferenceModelNode, writer);
-                    credentialTypeAttribute.marshallAsAttribute(credentialReferenceModelNode, writer);
+                    crNode = credentialReferenceModelNode;
```
bstansberry: The existence of this 'else' block and the name of this 'credentialReferenceModelNode' param tells me this has gone wrong. The ModelNode param is meant to be a ModelNode that is a parent, with this attribute as a child. Generally that means it is a ModelNode that represents the Resource, with children for all attributes in that resource. The param is not meant to be the attribute value itself. If it ever is I fear something has gone wrong.

Have a look at DefaultAttributeMarshaller for the standard kind of implementation. A check of isMarshallable() with no xml writing at all if that returns false, and then the attribute data is extracted from the model as you do here in L173.

bstansberry: Note I have assigned a JIRA to @ctomc to polish up the javadoc on AttributeMarshaller. The marshallAsElement javadoc is missing, although marshalAsAttribute gives a good indication of the expected behavior.

pskopek:
@bstansberry WFCORE JIRAs created and referenced
and marshaller and parser set to standard ones.
Thanks for spotting it!

controller/src/main/java/org/jboss/as/controller/security/CredentialReference.java
```diff
@@ -135,72 +129,30 @@ public static ObjectTypeAttributeDefinition getAttributeDefinition() {
...
-                return namedNode.asString();
-            }
-            return null;
+    public static String credentialReferencePartAsStringIfDefined(ModelNode credentialReferenceValue, String name) throws OperationFailedException {
+        assert !credentialReferenceValue.isDefined() : credentialReferenceValue;
+        ModelNode result = credentialReferenceValue.get(name);
+        if (result.isDefined()) {
+            return result.asString();
         }
         return null;
```
bstansberry: Something seems odd here.

L148 asserts that credentialReferenceValue is undefined. So the test at L150 will always test false and the method will return null.

On L148 did you not intend for the '!' to be there?

bstansberry: Yeah, reading the usage I'm almost sure the '!' was by mistake. :)

pskopek: yes, wrong copy/paste. Fixed.

PR #2097, opened 2017-01-18
Fix for WFCORE-2205. NPE in completion with invalid command

jfdenise:
Swallow exceptions thrown during completion. Avoid to complete some obvious invalid structures. Added unit test.

### DIFF
--- a/cli/src/main/java/org/jboss/as/cli/impl/ValueTypeCompleter.java
+++ b/cli/src/main/java/org/jboss/as/cli/impl/ValueTypeCompleter.java
@@ -48,6 +48,8 @@ import org.jboss.as.cli.parsing.WordCharacterHandler;
 import org.jboss.as.cli.parsing.arguments.ArgumentValueState;
 import org.jboss.dmr.ModelNode;
 import org.jboss.dmr.ModelType;
+import org.jboss.logging.Logger;
+import org.jboss.logging.Logger.Level;
 
 /**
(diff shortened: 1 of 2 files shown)


cli/src/main/java/org/jboss/as/cli/impl/ValueTypeCompleter.java
```diff
@@ -375,15 +379,24 @@ public int complete(CommandContext ctx, String buffer, int cursor, List<String>
...
-        final Collection<String> foundCandidates = handler.getCandidates(propDescr);
-        if(foundCandidates.isEmpty()) {
+        try {
+            final Collection<String> foundCandidates = handler.getCandidates(propDescr);
+            if (foundCandidates.isEmpty()) {
+                return -1;
+            }
+            candidates.addAll(foundCandidates);
+            return handler.getCompletionIndex();
+        } catch (Throwable th) {
```
bstansberry: Should this be RuntimeException? You can't really recover from Error and if someone changes something such that the try block throws a non-Runtime Exception then they probably should have to think about whether this catch logic is what they want.

Not a big deal, it's just that I've fixed (or introduced myself!!) a fair number of bugs that pop up later due to old code using catch (Exception) or catch (Throwable).

jfdenise:
@bstansberry , thank-you to pinpoint this, updated.

PR #2103, opened 2017-01-19
WFCORE-2202 multiple dynamic capability name parts

ctomc:
- adds support for capability multiple dynamic parts

### DIFF
--- a/controller/src/main/java/org/jboss/as/controller/AbstractAttributeDefinitionBuilder.java
+++ b/controller/src/main/java/org/jboss/as/controller/AbstractAttributeDefinitionBuilder.java
@@ -697,8 +697,8 @@ public abstract class AbstractAttributeDefinitionBuilder<BUILDER extends Abstrac
      *                             represented by the attribute's value
      * @param dependentCapability  the capability that depends on {@code referencedCapability}
      * @return the builder
-     * @see SimpleAttributeDefinition#addCapabilityRequirements(OperationContext, ModelNode)
-     * @see SimpleAttributeDefinition#removeCapabilityRequirements(OperationContext, ModelNode)
+     * @see AttributeDefinition#addCapabilityRequirements(OperationContext, org.jboss.as.controller.registry.Resource, ModelNode)
+     * @see AttributeDefinition#removeCapabilityRequirements(OperationContext, org.jboss.as.controller.registry.Resource, ModelNode)
      */
     public BUILDER setCapabilityReference(String referencedCapability, RuntimeCapability<?> dependentCapability) {
         return setCapabilityReference(referencedCapability, dependentCapability.getName(), dependentCapability.isDynamicallyNamed());
@@ -714,17 +714,67 @@ public abstract class AbstractAttributeDefinitionBuilder<BUILDER extends Abstrac
      * constructed using the parameters passed to this method.
      * <p>
      * <strong>NOTE:</strong> This method of recording capability references is only suitable for use in attributes
-     * only used in resources that themselves expose a single capability. When the capability requirement
-     * is registered, the dependent capability will be that capability.
+     * only used in resources that themselves expose a single capability.
+     * If your resource exposes more than single you should use {@link #setCapabilityReference(RuntimeCapability, String, AttributeDefinition...)} variant
+     * When the capability requirement is registered, the dependent capability will be that capability.
      *
      * @param referencedCapability the name of the dynamic capability the dynamic portion of whose name is
      *                             represented by the attribute's value
      * @return the builder
-     * @see SimpleAttributeDefinition#addCapabilityRequirements(OperationContext, ModelNode)
-     * @see SimpleAttributeDefinition#removeCapabilityRequirements(OperationContext, ModelNode)
+     * @see AttributeDefinition#addCapabilityRequirements(OperationContext, org.jboss.as.controller.registry.Resource, ModelNode)
+     * @see AttributeDefinition#removeCapabilityRequirements(OperationContext, org.jboss.as.controller.registry.Resource, ModelNode)
      */
     public BUILDER setCapabilityReference(String referencedCapability) {
         referenceRecorder = new CapabilityReferenceRecorder.ContextDependencyRecorder(referencedCapability);
+        return (BUILDER) this;
+    }
+
+    /**
+     * Records that this attribute's value represents a reference to an instance of a
+     * {@link org.jboss.as.controller.capability.RuntimeCapability#isDynamicallyNamed() dynamic capability}.
+     * <p>
+     * This method is a convenience method equivalent to calling * {@link #setCapabilityReference(CapabilityReferenceRecorder)}
+     * passing in a {@link CapabilityReferenceRecorder.CompositeAttributeDependencyRecorder}
+     * constructed using the parameters passed to this method.
+     * <p>
+     * <strong>NOTE:</strong> This method of recording capability references is only suitable for use in attributes
+     * only used in resources that themselves expose a single capability.
+     * If your resource exposes more than single capability, you should use {@link #setCapabilityReference(RuntimeCapability, String, AttributeDefinition...)}
+     * When the capability requirement
+     * is registered, the dependent capability will be that capability.
+     *
+     * @param referencedCapability the name of the dynamic capability the dynamic portion of whose name is
+     *                             represented by the attribute's value
+     *  @param dependantAttributes attribute from same resource that will be used to derive multiple dynamic parts for the dependant capability
+     * @return the builder
+     * @see AttributeDefinition#addCapabilityRequirements(OperationContext, org.jboss.as.controller.registry.Resource, ModelNode)
+     * @see AttributeDefinition#removeCapabilityRequirements(OperationContext, org.jboss.as.controller.registry.Resource, ModelNode)
+     */
+    public BUILDER setCapabilityReference(String referencedCapability, AttributeDefinition ... dependantAttributes) {
+        referenceRecorder = new CapabilityReferenceRecorder.CompositeAttributeDependencyRecorder(referencedCapability, dependantAttributes);
+        return (BUILDER) this;
+    }
+
+    /**
+     * Records that this attribute's value represents a reference to an instance of a
+     * {@link org.jboss.as.controller.capability.RuntimeCapability#isDynamicallyNamed() dynamic capability}.
+     * <p>
+     * This method is a convenience method equivalent to calling {@link #setCapabilityReference(CapabilityReferenceRecorder)}
+     * passing in a {@link CapabilityReferenceRecorder.CompositeAttributeDependencyRecorder}
+     * constructed using the parameters passed to this method.
+     * <p>
+     * When the capability requirement is registered, the dependent capability will be that capability.
+     *
+     * @param capability requirement capability
+     * @param referencedCapability the name of the dynamic capability the dynamic portion of whose name is
+     *                             represented by the attribute's value
+     * @param dependantAttributes attributes on resource which will be used for registering capability reference, can be multiple.
+     * @return the builder
+     * @see AttributeDefinition#addCapabilityRequirements(OperationContext, org.jboss.as.controller.registry.Resource, ModelNode)
+     * @see AttributeDefinition#removeCapabilityRequirements(OperationContext, org.jboss.as.controller.registry.Resource, ModelNode)
+     */
+    public BUILDER setCapabilityReference(RuntimeCapability capability, String referencedCapability, AttributeDefinition ... dependantAttributes) {
+        referenceRecorder = new CapabilityReferenceRecorder.CompositeAttributeDependencyRecorder(capability, referencedCapability, dependantAttributes);
         return (BUILDER) this;
     }
 
@@ -744,8 +794,8 @@ public abstract class AbstractAttributeDefinitionBuilder<BUILDER extends Abstrac
      *                             portion of which comes from the name of the resource with which
      *                             the attribute is associated
      * @return the builder
-     * @see SimpleAttributeDefinition#addCapabilityRequirements(OperationContext, ModelNode)
-     * @see SimpleAttributeDefinition#removeCapabilityRequirements(OperationContext, ModelNode)
+     * @see AttributeDefinition#addCapabilityRequirements(OperationContext, org.jboss.as.controller.registry.Resource, ModelNode)
+     * @see AttributeDefinition#removeCapabilityRequirements(OperationContext, org.jboss.as.controller.registry.Resource, ModelNode)
      */
     public BUILDER setCapabilityReference(String referencedCapability, String dependentCapability, boolean dynamicDependent) {
         referenceRecorder = new CapabilityReferenceRecorder.DefaultCapabilityReferenceRecorder(referencedCapability, dependentCapability, dynamicDependent);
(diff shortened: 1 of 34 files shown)


controller/src/main/java/org/jboss/as/controller/AbstractAttributeDefinitionBuilder.java
```diff
@@ -734,6 +734,35 @@ public BUILDER setCapabilityReference(String referencedCapability) {
...
+     * @return the builder
+     * @see SimpleAttributeDefinition#addCapabilityRequirements(OperationContext, ModelNode)
+     * @see SimpleAttributeDefinition#removeCapabilityRequirements(OperationContext, ModelNode)
+     */
+    public BUILDER setCapabilityReference(String referencedCapability, SimpleAttributeDefinition ... dependantAttributes) {
+        referenceRecorder = new CapabilityReferenceRecorder.CompositeAttributeDependencyRecorder(referencedCapability, dependantAttributes);
+        return (BUILDER) this;
+    }
+
+    public BUILDER setCapabilityReference(RuntimeCapability capability, String referencedCapability, SimpleAttributeDefinition ... dependantAttributes) {
```
bstansberry: javadoc

controller/src/main/java/org/jboss/as/controller/CapabilityReferenceRecorder.java
```diff
@@ -237,4 +245,104 @@ public boolean isDynamicDependent() {
...
+
+    /**
+     * {@link ResourceNameCompositeDependencyRecorder} that determines the dependent capability
+     * from the {@link OperationContext} and resource name.
+     * This assumes that the {@link OperationContext#getResourceRegistration() resource registration associated with currently executing step}
+     * will expose a {@link ImmutableManagementResourceRegistration#getCapabilities() capability set} including
+     * one and only one capability. <strong>This recorder cannot be used with attributes associated with resources
+     * that do not meet this requirement.</strong>
+     */
+    public class ResourceNameCompositeDependencyRecorder extends CompositeAttributeDependencyRecorder {
```
bstansberry: 'public' is unnecessary

I'm saving myself from IDEAs future whinging. ;)

controller/src/main/java/org/jboss/as/controller/CapabilitiesServiceTarget.java
```diff
@@ -0,0 +1,18 @@
...
+/**
+ * @author Tomaz Cerar (c) 2017 Red Hat Inc.
+ */
+public interface CapabilitiesServiceTarget extends ServiceTarget {
+
+    <T> CapabilitiesServiceBuilder<T> addService(final ServiceName name, final Service<T> service) throws IllegalArgumentException;
+
+    <T> CapabilitiesServiceBuilder<T> addCapability(final RuntimeCapability<?> capability, final Service<T> service) throws IllegalArgumentException;
+
+
```
bstansberry: full javadoc

controller/src/main/java/org/jboss/as/controller/CapabilitiesServiceBuilder.java
```diff
@@ -0,0 +1,21 @@
...
+
+    <I> CapabilitiesServiceBuilder<T> addCapabilityRequirement(String capabilityName, Class<I> type, Injector<I> target, String ... referenceNames);
+
+    <I> CapabilitiesServiceBuilder<T> addCapabilityRequirement(String capabilityName, Class<I> type, Injector<I> target);
+
+    <I> CapabilitiesServiceBuilder<T> addDependency(ServiceName dependency, Class<I> type, Injector<I> target);
+
+    @Override
+    CapabilitiesServiceBuilder<T> setInitialMode(ServiceController.Mode mode);
+}
```
bstansberry: full javadoc

controller/src/main/java/org/jboss/as/controller/capability/AbstractCapability.java
```diff
@@ -45,6 +48,7 @@
     // specify the dynamic part of the requirement name, and a
     // runtime-only requirement by definition cannot be specified
     // via configuration
+    protected final Function<PathAddress,String[]> dynamicNameMapper;
```
bstansberry: package protected

controller/src/main/java/org/jboss/as/controller/capability/RuntimeCapability.java
```diff
@@ -391,9 +462,7 @@ private Builder(String baseName, boolean dynamic, T runtimeAPI) {
          * @param requirements the capability names
          * @return the builder
          *
-         * @deprecated nothing is currently done with this data and unless a true use is implemented this method may be removed
          */
-        @Deprecated
```
bstansberry: IIRC the use case we were thinking about for this went away, so it should remain deprecated

controller/src/main/java/org/jboss/as/controller/capability/RuntimeCapability.java
```diff
@@ -449,9 +518,7 @@ private Builder(String baseName, boolean dynamic, T runtimeAPI) {
...
```
bstansberry: same as above

bstansberry:
There are tons of test failures, so something needs fixing. ;) But this seems ok; my comments are all minor stuff or polishing.

controller/src/main/java/org/jboss/as/controller/CapabilityReferenceRecorder.java
ctomc: This one is actually needed as this recorder is not used in builders.
Maybe it would make sense to add new method to attribute builder that would use this class as well.

controller/src/main/java/org/jboss/as/controller/capability/RuntimeCapability.java
ctomc: Use case is still here, it is now actually in use. 
See https://github.com/wildfly/wildfly-core/pull/2103/files#diff-340f1dd448e96a0a6a57e4028fb436faR243
We talked about deferring the actual validation to validateCapabilites() in CapabilityRegistry

Recorder now uses this information to know if requirement should be added or not.

controller/src/main/java/org/jboss/as/controller/CapabilityReferenceRecorder.java
```diff
@@ -186,34 +232,42 @@ public final void addCapabilityRequirements(OperationContext context, String att
...
+        protected void processCapabilityRequirement(OperationContext context, Resource resource, String attributeName, boolean remove, String... attributeValues) {
+            RuntimeCapability cap = getDependentCapability(context);
+            String dependentName = getDependentName(cap, context);
             for (String attributeValue : attributeValues) {
-                String requirementName = RuntimeCapability.buildDynamicCapabilityName(baseRequirementName, attributeValue);
-                if (remove) {
-                    context.deregisterCapabilityRequirement(requirementName, dependentName);
-                } else {
-                    context.registerAdditionalCapabilityRequirement(requirementName, dependentName, attributeName);
+                if (attributeValue != null || !cap.getDynamicOptionalRequirements().contains(baseRequirementName)) {
```
bstansberry: How can null ever be a useful value here?

If the AD allows undefined, since no one ever populates RuntimeCap's dynamic optional requirements, this will just result in registering a strange requirement.

ctomc: Currently there are few uses of registering optional dynamic requirements.
It was just that this info was nowhere to be used.

By definition all requirements are mandatory, so if you have an optional requirement which is in some cases a design requirement, you need to express this in some way.

With this change, what happens is that if attribute is null aka is not configured and requirement is not specified as "optional" requirement as such must be added to the registry.

The requirement name becomes bit bogus, as it is basename + "null" but none the less it tells the subsystem author something is wrong.

Going forward whole validation for such scenarios should be improved, but that belongs into validation of capabilities not gathering the requirements.

controller/src/main/java/org/jboss/as/controller/AttributeDefinition.java
```diff
@@ -1007,37 +1008,57 @@ public ModelNode getNoTextDescription(boolean forOperation) {
...
+    }
+    /**
+     * Based on the given attribute value, add capability requirements. If this definition
+     * is for an attribute whose value is or contains a reference to the name of some capability,
+     * this method should record the addition of a requirement for the capability.
+     * <p>
+     * This is a no-op in this base class. Subclasses that support attribute types that can represent
+     * capability references should override this method.
+     *  @param context the operation context
+     * @param resource
```
bstansberry: Needs a description

controller/src/main/java/org/jboss/as/controller/AttributeDefinition.java
```diff
@@ -1007,37 +1008,57 @@ public ModelNode getNoTextDescription(boolean forOperation) {
...
             // We can't process expressions
             if (attributeValue.getType() != ModelType.EXPRESSION) {
                 ModelNode value = attributeValue.isDefined() ? attributeValue : (defaultValue != null) ? defaultValue : new ModelNode();
-                referenceRecorder.addCapabilityRequirements(context, name, value.isDefined() ? value.asString() : null);
+                referenceRecorder.addCapabilityRequirements(context, resource, name, value.isDefined() ? value.asString() : null);
             }
         }
     }
 
+    @Deprecated
```
bstansberry: Needs the javadoc restored.

controller/src/main/java/org/jboss/as/controller/AttributeDefinition.java
```diff
@@ -1007,37 +1008,57 @@ public ModelNode getNoTextDescription(boolean forOperation) {
...
      * Based on the given attribute value, remove capability requirements. If this definition
      * is for an attribute whose value is or contains a reference to the name of some capability,
      * this method should record the removal of a requirement for the capability.
      * <p>
      * This is a no-op in this base class. Subclasses that support attribute types that can represent
      * capability references should override this method.
-     *
-     * @param context the operation context
+     *  @param context the operation context
+     * @param resource
```
bstansberry: Needs a description.

controller/src/main/java/org/jboss/as/controller/CapabilityServiceBuilder.java
```diff
@@ -0,0 +1,55 @@
+package org.jboss.as.controller;
```
bstansberry: Needs the copyright header.

controller/src/main/java/org/jboss/as/controller/CapabilityServiceTarget.java
```diff
@@ -0,0 +1,17 @@
+package org.jboss.as.controller;
```
bstansberry: Needs the copyright header.

controller/src/main/java/org/jboss/as/controller/AttributeDefinition.java
bstansberry: The description should state that null is legal.

controller/src/main/java/org/jboss/as/controller/AttributeDefinition.java
bstansberry: The description should state that null is legal.

controller/src/main/java/org/jboss/as/controller/OperationContextImpl.java
```diff
@@ -2018,15 +2028,17 @@ private CapabilityScope createCapabilityContext(Step step) {
      * Service target that delegates to another target for most calls, but during execution of the
      * management op that created it does management specific integration and limits the available API.
      */
-    private static class ContextServiceTarget implements ServiceTarget {
+    private static class ContextServiceTarget implements CapabilityServiceTarget {
 
         private final ServiceTarget delegate;
         private final Set<ContextServiceBuilder> builders = new HashSet<>();
         private volatile ContextServiceBuilderSupplier builderSupplier;
+        private PathAddress targetAddress;
```
bstansberry: final

controller/src/main/java/org/jboss/as/controller/OperationContextImpl.java
```diff
@@ -2695,4 +2616,47 @@ public ServiceName getCapabilityServiceName(String capabilityBaseName, String dy
         }
     }
 
+
+    private static class CapabilityServiceBuilderImpl<T> extends DelegatingServiceBuilder<T> implements CapabilityServiceBuilder<T> {
+        private PathAddress targetAddress;
```
bstansberry: final

controller/src/main/java/org/jboss/as/controller/services/DelegatingServiceBuilder.java
```diff
@@ -0,0 +1,201 @@
+package org.jboss.as.controller.services;
```
bstansberry: Copyright header

controller/src/main/java/org/jboss/as/controller/services/DelegatingServiceTarget.java
```diff
@@ -0,0 +1,137 @@
+package org.jboss.as.controller.services;
```
bstansberry: Copyright header.

controller/src/main/java/org/jboss/as/controller/services/DelegatingServiceTarget.java
```diff
@@ -0,0 +1,137 @@
...
+
+/**
+ * An "insulated" view of a service target which prevents access to other public methods on the delegate target object.
+ *
+ * this class is copy of {@link org.jboss.msc.service.DelegatingServiceBuilder} with exception of not being final
+ *
+ * @author <a href="mailto:david.lloyd@redhat.com">David M. Lloyd</a>
+ * @author <a href="mailto:ropalka@redhat.com">Richard Opalka</a>
+ */
+public class DelegatingServiceTarget implements ServiceTarget {
```
bstansberry: package protected

controller/src/main/java/org/jboss/as/controller/services/DelegatingServiceBuilder.java
```diff
@@ -0,0 +1,201 @@
...
+/**
+ * A service builder which delegates to another service builder.
+ * this class is copy of {@link org.jboss.msc.service.DelegatingServiceBuilder} with exception of not being final
+ *
+ * @param <T> the service type
+ *
+ * @author <a href="mailto:david.lloyd@redhat.com">David M. Lloyd</a>
+ * @author <a href="mailto:ropalka@redhat.com">Richard Opalka</a>
+ */
+public class DelegatingServiceBuilder<T> implements ServiceBuilder<T> {
```
bstansberry: package protected

controller/src/main/java/org/jboss/as/controller/OperationContext.java
```diff
@@ -482,6 +482,16 @@
...
 
     /**
+     * Get the service target.  If the step is not a runtime operation handler step, an exception will be thrown.  The
+     * returned service target is limited such that only the service add methods are supported.  If a service added
+     * to this target was removed by a prior operation step, the install will wait until the removal completes.
+     *
+     * @return the service target
+     * @throws UnsupportedOperationException if the calling step is not a runtime operation step
+     */
+    CapabilityServiceTarget getCapabilityServiceTarget() throws UnsupportedOperationException;
```
bstansberry: Is a new method needed here vs just changing the return type of the existing method?

controller/src/main/java/org/jboss/as/controller/CapabilityReferenceRecorder.java
bstansberry: Let's drop this and we can come back to it later. This PR is mostly ready to go and I'd like to finish it off today for a possible core release tonight/tomorrow.

I have code in full that I know would be affected by this once I tweaked it to no longer force use of DefaultCapabilityReferenceRecorder, which I was planning to do once the null check here was in. So, if I would have to work around this, others might too and it's too late in the cycle and we're too busy to add further work like that.

Also, DefaultCapabilityReferenceRecorder would need the same change if we decide to do this later.

controller/src/main/java/org/jboss/as/controller/OperationContext.java
ctomc: Initially it was just returning different type, but that breaks binary compatibility for libs that were compiled against previous version of wildfly-controller and used this method.

It showed obvious with elytron-subsystem in wildfly full. I would prefer it to just have different return type but backward compatibility is important.

controller/src/main/java/org/jboss/as/controller/CapabilityServiceTarget.java
ctomc: fixed, not sure why GH is not showing it as resolved

ctomc:
@bstansberry Updated PR to address all your comments.

controller/src/main/java/org/jboss/as/controller/OperationContext.java
bstansberry: OK, np. Now I'm curious why it breaks. :)

PR #2066, opened 2017-01-10
[WFCORE-2169]: ProcessStateListener receives first running state transition from SUSPENDED to NORMAL when start-mode=suspend

ehsavoie:
Fixing transitions when starting.
Removing unused code.
Fixing intermittent failure on the remove interface test that occurred on initial PR. (https://github.com/wildfly/wildfly-core/pull/2021)
Jira: https://issues.jboss.org/browse/WFCORE-2169
JBEAP: https://issues.jboss.org/browse/JBEAP-8110

PR #2047, opened 2017-01-03
[WFCORE-2145]:  Confusing transitions of ControlledProcessState.

ehsavoie:
Wrong transition recorded when starting in suspended mode.

Jira: https://issues.jboss.org/browse/WFCORE-2145
JBEAP: https://issues.jboss.org/browse/JBEAP-8017

PR #2021, opened 2016-12-12
Removing unused code, and renaming some utilities.

ehsavoie:
Leftover code from WFCORE-1157 that is not required anymore.

PR #2110, opened 2017-01-23
Massive merge : a single PR to merge  3 Prs into one

ehsavoie:
This PR is a merge of 
 * [WFCORE-2179]: When in SUSPENDED state we shouldn't suspend on stopping : https://github.com/wildfly/wildfly-core/pull/2073
If in state SUSPENDED when suspend is called, just return.
Jira: https://issues.jboss.org/browse/WFCORE-2179

 * [WFCORE-2145]: Confusing transitions of ControlledProcessState. : https://github.com/wildfly/wildfly-core/pull/2047
Wrong transition recorded when starting in suspended mode.
Jira: https://issues.jboss.org/browse/WFCORE-2145
JBEAP: https://issues.jboss.org/browse/JBEAP-8017

 * [WFCORE-2169]: ProcessStateListener receives first running state transition from SUSPENDED to NORMAL when start-mode=suspend : https://github.com/wildfly/wildfly-core/pull/2066
Fixing transitions when starting.
Removing unused code.
Fixing intermittent failure on the remove interface test that occurred on initial PR. (#2021)
Jira: https://issues.jboss.org/browse/WFCORE-2169
JBEAP: https://issues.jboss.org/browse/JBEAP-8110


testsuite/manualmode/src/test/java/org/wildfly/core/test/standalone/mgmt/RemoveManagementInterfaceTestCase.java
```diff
@@ -184,4 +192,59 @@ static ModelControllerClient getHttpModelControllerClient() {
...
+            } catch (IOException e) {
+            }
+            try {
+                Thread.sleep(100);
+            } catch (InterruptedException e) {
+            }
+
+        }
+        fail("Live Server did not reload in the imparted time.");
+    }
```
bstansberry: This util stuff looks like what's already in the ServerReload class in testsuite/shared.

ehsavoie: It is different in that the client is http-remoting in the reload op and native remote when checking for reload.
I guess we are missing a executeReloadAndWaitForCompletion with a protocol parameter in the ServerReload

core-management/core-management-subsystem/src/main/java/org/wildfly/extension/core/management/ProcessStateListenerService.java
```diff
@@ -72,14 +74,14 @@
     private final PropertyChangeListener propertyChangeListener;
     private final OperationListener operationListener;
     private final ProcessStateListener listener;
-    private final ProcessStateListenerInitParameters parameters;
+    private ProcessStateListenerInitParameters parameters;
```
bstansberry: This can still be final

bstansberry:
Superceded by #2118

PR #2106, opened 2017-01-20
Pull request 2102 update

bstansberry:
Update #2102 with a commit to ensure its new test cleans up after itself

PR #2111, opened 2017-01-23
Revert "Pull request 2102 update"

bstansberry:
Reverts wildfly/wildfly-core#2106

We are seeing CI failures related to CommandContextImpl dealing with RedirectException during reload handling, which is close enough to this work that we're assuming there is a relationship.


bstansberry:
retest this please

PR #2113, opened 2017-01-24
Fix for WFCORE-2216. Handle redirect in reload and shutdown

jfdenise:
Second attempt to fix this issue. We were not too far in the previous fix but we were wrongly exiting the reload waiting logic too early.

For reload, when redirect occurs, attempt to reconnect the client. If failure occurs, ignore it (could be caused by exotic Redirect sent during server startup) and carry-on the reload loop. If reconnect succeeds, exit the reload loop.

For shutdown, if IOException occurs, CLIModelControllerClient waits the end of the waiting time then propagate the RedirectException (if any). In turn the shutdown handler catchs the RedirectException and attempt to reconnect. If it fails, it disconnects the client.

Unit test added (the one from reverted commit).

### DIFF
--- a/cli/src/main/java/org/jboss/as/cli/impl/CLIModelControllerClient.java
+++ b/cli/src/main/java/org/jboss/as/cli/impl/CLIModelControllerClient.java
@@ -58,6 +58,7 @@ import org.jboss.remoting3.Connection;
 import org.jboss.remoting3.Endpoint;
 import org.jboss.threads.JBossThreadFactory;
 import org.xnio.OptionMap;
+import org.xnio.http.RedirectException;
 
 /**
  * @author Alexey Loubyansky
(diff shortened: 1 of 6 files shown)


cli/src/main/java/org/jboss/as/cli/impl/CLIModelControllerClient.java
```diff
@@ -297,8 +301,22 @@ public void ensureConnected(long timeoutMillis) throws CommandLineException {
...
-                    throw new CommandLineException("Failed to establish connection in " + (System.currentTimeMillis() - start)
-                            + "ms", ioe);
+                    if (timeoutOccured) {
+                        throw new CommandLineException("Failed to establish connection in " + (System.currentTimeMillis() - start)
+                                + "ms", ioe);
+                    } else {
+                        timeoutOccured = true;
+                    }
+                }
+                // Only propagate RedirectException at the end of timeout.
```
bstansberry: It's unfortunate that this means it can take a long time to connect, if the CliConfig connectTimeout is high. (IIRC there was a request to make that 30 secs by default to avoid some spurious failures, although maybe that was a different timeout.)

I don't have a good solution though; the approach used by CommandContextImpl to analyze the redirect exception isn't possible here as the requested protocol isn't known.

Bit of a corner case anyway; this http/https redirect in combination with "shutdown --restart=true" + a long timeout config.

jfdenise: @bstansberry, if this becomes an issue, we can change this. I first implemented a logic based on the protocol. We can retrieve it from the channelConfig and compare wth the protocol received in the redirect exception.
I although thought that it was a corner case. It would have been problematic for reload.

PR #2115, opened 2017-01-24
[WFCORE-2229]: Create serverside operations to redeploy deployments affected by an overlay.

ehsavoie:
Adding support for a new redeploy-affected method on deployment-overlays.
Adding a flag to redeploy affected when removing an overlay link.

Jira: https://issues.jboss.org/browse/WFCORE-2229
This is blocking https://issues.jboss.org/browse/WFCORE-1752

### DIFF
--- a/host-controller/src/main/java/org/jboss/as/domain/controller/logging/DomainControllerLogger.java
+++ b/host-controller/src/main/java/org/jboss/as/domain/controller/logging/DomainControllerLogger.java
@@ -802,4 +802,7 @@ public interface DomainControllerLogger extends BasicLogger {
 
     @Message(id = 97, value = "Cannot explode a subdeployment of an unexploded deployment")
     OperationFailedException cannotExplodeSubDeploymentOfUnexplodedDeployment();
+
+    @Message(id = 98, value = "You are trying to redeploy %s which are deployments not affected by the overlay")
+    OperationFailedException redeployingUnaffactedDeployments(Collection<String> deployments);
 }
--- a/host-controller/src/main/java/org/jboss/as/domain/controller/operations/DomainDeploymentOverlayRedeployLinksHandler.java
+++ b/host-controller/src/main/java/org/jboss/as/domain/controller/operations/DomainDeploymentOverlayRedeployLinksHandler.java
@@ -1,0 +1,90 @@
+/*
+ * Copyright 2017 JBoss by Red Hat.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.jboss.as.domain.controller.operations;
+
+import java.util.HashSet;
+import java.util.Set;
+import org.jboss.as.controller.OperationContext;
+import org.jboss.as.controller.OperationDefinition;
+import org.jboss.as.controller.OperationFailedException;
+import org.jboss.as.controller.OperationStepHandler;
+import org.jboss.as.controller.SimpleOperationDefinitionBuilder;
+import org.jboss.as.controller.StringListAttributeDefinition;
+import org.jboss.as.controller.client.helpers.Operations;
+import org.jboss.as.controller.descriptions.ModelDescriptionConstants;
+import static org.jboss.as.controller.descriptions.ModelDescriptionConstants.DEPLOYMENT;
+import static org.jboss.as.controller.descriptions.ModelDescriptionConstants.DEPLOYMENTS;
+import static org.jboss.as.controller.descriptions.ModelDescriptionConstants.DEPLOYMENT_OVERLAY_LINK_REMOVAL;
+import static org.jboss.as.controller.descriptions.ModelDescriptionConstants.REDEPLOY_LINKS;
+import org.jboss.as.controller.descriptions.common.ControllerResolver;
+import org.jboss.as.domain.controller.logging.DomainControllerLogger;
+import org.jboss.as.server.deploymentoverlay.AffectedDeploymentOverlay;
+import org.jboss.dmr.ModelNode;
+
+/**
+ * Handler to redeploy deployments linked to an overlay in domain mode.
+ * @author Emmanuel Hugonnet (c) 2017 Red Hat, inc.
+ */
+public class DomainDeploymentOverlayRedeployLinksHandler implements OperationStepHandler {
+
+    public static final StringListAttributeDefinition RUNTIME_NAMES_DEFINITION =
+            new StringListAttributeDefinition.Builder(DEPLOYMENTS)
+                            .setRequired(false)
+                            .build();
+
+    public static final OperationDefinition REDEPLOY_LINKS_DEFINITION = new SimpleOperationDefinitionBuilder(
+            REDEPLOY_LINKS, ControllerResolver.getResolver(ModelDescriptionConstants.DEPLOYMENT_OVERLAY))
+            .addParameter(RUNTIME_NAMES_DEFINITION)
+            .build();
+
+    private final boolean domainRoot;
+
+    public DomainDeploymentOverlayRedeployLinksHandler(boolean domainRoot) {
+        this.domainRoot = domainRoot;
+    }
+
+    @Override
+    public void execute(OperationContext context, ModelNode operation) throws OperationFailedException {
+        Set<String> runtimeNames = domainRoot ? AffectedDeploymentOverlay.listAllLinks(context, context.getCurrentAddressValue()) : AffectedDeploymentOverlay.listLinks(context, context.getCurrentAddress());
+        if(operation.hasDefined(RUNTIME_NAMES_DEFINITION.getName())) {
+            Set<String> requiredRuntimeNames = new HashSet<>(RUNTIME_NAMES_DEFINITION.unwrap(context, operation));
+            Set<String> unaffectedDeployments = AffectedDeploymentOverlay.checkRequiredNames(requiredRuntimeNames, runtimeNames);
+            if(!unaffectedDeployments.isEmpty() && !isRedeployAfterRemoval(operation)) {
+                throw DomainControllerLogger.ROOT_LOGGER.redeployingUnaffactedDeployments(unaffectedDeployments);
+            }
+            runtimeNames = requiredRuntimeNames;
+        }
+        ModelNode removeOperation = null;
+        if(isRedeployAfterRemoval(operation)) {
+            removeOperation = Operations.createRemoveOperation(context.getCurrentAddress().append(DEPLOYMENT, runtimeNames.iterator().next()).toModelNode());
+        }
+        if (domainRoot) {
+            AffectedDeploymentOverlay.redeployLinksAndTransformOperationForDomain(context, runtimeNames, removeOperation);
+        } else {
+            AffectedDeploymentOverlay.redeployLinksAndTransformOperation(context, removeOperation, context.getCurrentAddress().getParent(), runtimeNames);
+        }
+    }
+
+    /**
+     * @param operation the current operation.
+     * @return true if this is a redeploy after the removal of a link.
+     * @see org.jboss.as.server.deploymentoverlay.DeploymentOverlayDeploymentRemoveHandler
+     */
+    private boolean isRedeployAfterRemoval(ModelNode operation) {
+        return operation.hasDefined(DEPLOYMENT_OVERLAY_LINK_REMOVAL) &&
+                operation.get(DEPLOYMENT_OVERLAY_LINK_REMOVAL).asBoolean();
+    }
+}
--- a/host-controller/src/main/java/org/jboss/as/domain/controller/operations/coordination/ServerOperationResolver.java
+++ b/host-controller/src/main/java/org/jboss/as/domain/controller/operations/coordination/ServerOperationResolver.java
@@ -25,36 +25,25 @@
  */
 package org.jboss.as.domain.controller.operations.coordination;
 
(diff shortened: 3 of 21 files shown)


jfdenise:
@bstansberry, is this PR planed to be merged? It would allow me to submit the client side.
Thanks.

bstansberry:
@jfdenise After I get a chance to review it. My task list is clearing out a bit so that should happen this week.

jfdenise:
@bstansberry , thank-you!

server/src/main/java/org/jboss/as/server/deploymentoverlay/DeploymentOverlayRedeployLinksHandler.java
```diff
@@ -0,0 +1,61 @@
...
+            .addParameter(RUNTIME_NAMES_DEFINITION)
+            .build();
+
+    @Override
+    public void execute(OperationContext context, ModelNode operation) throws OperationFailedException {
+        Set<String> runtimeNames = AffectedDeploymentOverlay.listLinks(context, context.getCurrentAddress());
+        if(operation.hasDefined(RUNTIME_NAMES_DEFINITION.getName())) {
+            Set<String> requiredRuntimeNames = new HashSet<>(RUNTIME_NAMES_DEFINITION.unwrap(context, operation));
+            if(!runtimeNames.containsAll(requiredRuntimeNames)) {
+                throw new OperationFailedException("WTF are you trying to do !!!!!");
```
bstansberry: :)

server/src/main/java/org/jboss/as/server/deploymentoverlay/AffectedDeploymentOverlay.java
```diff
@@ -0,0 +1,212 @@
...
+        Operations.CompositeOperationBuilder opBuilder = Operations.CompositeOperationBuilder.create();
+        deploymentNames.forEach((deploymentName) -> opBuilder.addStep(addRedeployStep(context.getCurrentAddress().getParent().append(DEPLOYMENT, deploymentName))));
+        // Add the domain op transformer
+        List<DomainOperationTransformer> transformers = context.getAttachment(OperationAttachments.SLAVE_SERVER_OPERATION_TRANSFORMERS);
+        if (transformers == null) {
+            context.attach(OperationAttachments.SLAVE_SERVER_OPERATION_TRANSFORMERS, transformers = new ArrayList<>());
+        }
+        final ModelNode slave = opBuilder.build().getOperation();
+        transformers.add(new OverlayOperationTransformer(slave));
+    }
```
bstansberry: Unfortunately, DomainOperationTransformer is an extremely poorly named thing used for a quite specialized thing -- replacing content params without a "hash" field with ones with a "hash" field. It wasn't original meant as a "transformer" in the sense that's needed here, i.e. something that can translate an op that a legacy slave in a mixed domain doesn't understand into one it does understand, or produce a user friendly error message if that's not possible.

It looks like you've repurposed DomainOperationTransformer to reuse it for mixed domain. I suspect it will work too! But it makes me really nervous adding a whole new pattern to how we do these things. People won't remember it and are therefore likely to break it.

ehsavoie: it works also even without a legacy slave: this was a nice side effect ;)

server/src/main/java/org/jboss/as/server/deploymentoverlay/AffectedDeploymentOverlay.java
```diff
@@ -0,0 +1,212 @@
...
+     * @return
+     */
+    public static Set<String> listDeploymentNames(OperationContext context, PathAddress deploymentRootAddress, Set<Pattern> patterns) {
+        Set<String> deploymentNames = new HashSet<>();
+        Resource deploymentRootResource = context.readResourceFromRoot(deploymentRootAddress);
+        if (deploymentRootResource.hasChildren(DEPLOYMENT)) {
+            for (Resource.ResourceEntry deploymentResource : deploymentRootResource.getChildren(DEPLOYMENT)) {
+                if (isAcceptableDeployment(deploymentResource.getModel(), patterns)) {
+                    deploymentNames.add(deploymentResource.getName());
+                } else if (deploymentResource.hasChildren(SUBDEPLOYMENT)) {
```
bstansberry: Does the L132 test always work correctly on an HC? There is no SUBDEPLOYMENT child on the HC.

host-controller/src/main/java/org/jboss/as/domain/controller/operations/coordination/ServerOperationResolver.java
```diff
@@ -499,7 +503,7 @@ private static ModelNode getDomainModel(OperationContext context) {
         return ret;
     }
 
-    private Map<Set<ServerIdentity>, ModelNode> getServerGroupOperations(ModelNode operation, PathAddress address,
+    private Map<Set<ServerIdentity>, ModelNode> getServerGroupOperations(OperationContext context, ModelNode operation, PathAddress address,
```
bstansberry: This param is never used.

server/src/main/java/org/jboss/as/server/deploymentoverlay/AffectedDeploymentOverlay.java
bstansberry: @ehsavoie This makes me very nervous. But, it's very clever! So I'll go for it. But, two conditions:

1) There needs to be a PR in full adding coverage of both the new op and the remove+param case to the mixed domain testsuite. This would be necessary even if you did something less clever though.

2) There have to be really good comments added in quite a few places here. 3 months from now everyone but you will look at this and have no clue. 3 years from now you'll have no clue either. ;) It's not doing standard stuff so people won't be able to reason from the typical ways transformers work and domain rollout works, so it needs to be explained in the code. I'll add some comments now pointing out places where I'd like to see explanatory comments.

host-controller/src/main/java/org/jboss/as/domain/controller/operations/DomainDeploymentOverlayRedeployLinksHandler.java
```diff
@@ -0,0 +1,70 @@
...
+    @Override
+    public void execute(OperationContext context, ModelNode operation) throws OperationFailedException {
+        Set<String> runtimeNames = domainRoot ? AffectedDeploymentOverlay.listAllLinks(context, context.getCurrentAddressValue()) : AffectedDeploymentOverlay.listLinks(context, context.getCurrentAddress());
+        if(operation.hasDefined(RUNTIME_NAMES_DEFINITION.getName())) {
+            Set<String> requiredRuntimeNames = new HashSet<>(RUNTIME_NAMES_DEFINITION.unwrap(context, operation));
+            if(!runtimeNames.containsAll(requiredRuntimeNames)) {
+                throw new OperationFailedException("WTF are you trying to do !!!!!");
+            }
+            runtimeNames = requiredRuntimeNames;
+        }
```
bstansberry: At this point we need a big fat comment explaining what the L64-68 stuff is about to do. How it's going to set things up such that when DomainSlaveHandler and ServerOperationResolver run they convert the op executing here into ops all HCs understand. So the slaves, whether current version or older, see the old ops.

server/src/main/java/org/jboss/as/server/deploymentoverlay/DeploymentOverlayDeploymentRemoveHandler.java
```diff
@@ -0,0 +1,96 @@
...
+            = new SimpleOperationDefinitionBuilder(REMOVE, ControllerResolver.getResolver(DEPLOYMENT_OVERLAY + '.' + DEPLOYMENT))
+                    .addParameter(REDEPLOY_AFFECTED_DEFINITION)
+                    .build();
+
+    public static final DeploymentOverlayDeploymentRemoveHandler INSTANCE = new DeploymentOverlayDeploymentRemoveHandler();
+
+    @Override
+    protected void performRemove(OperationContext context, final ModelNode operation, final ModelNode model) throws OperationFailedException {
+        if (REDEPLOY_AFFECTED_DEFINITION.resolveModelAttribute(context, operation).asBoolean()) {
+            if (SERVER_GROUP.equals(context.getCurrentAddress().getElement(0).getKey())) {
```
bstansberry: At this point we need a big fat comment explaining what the L59 is about to do. How it's going to set things up such that when DomainSlaveHandler and ServerOperationResolver run they convert the op executing here into ops all HCs understand. So the slaves, whether current version or older, see the old ops.

server/src/main/java/org/jboss/as/server/deploymentoverlay/DeploymentOverlayDeploymentRemoveHandler.java
```diff
@@ -0,0 +1,96 @@
...
+                AffectedDeploymentOverlay.redeployLinksAndTransformOperationForServerGroup(context, context.getCurrentAddress().getParent().getParent(), Collections.singleton(context.getCurrentAddressValue()));
+            } else {
```
bstansberry: Here we need a comment about how this is going to add a step to do the redeploy.

server/src/main/java/org/jboss/as/server/deploymentoverlay/AffectedDeploymentOverlay.java
```diff
@@ -0,0 +1,212 @@
...
+import org.jboss.as.controller.PathAddress;
+import org.jboss.as.controller.PathElement;
+import org.jboss.as.controller.client.helpers.Operations;
+import org.jboss.as.controller.operations.DomainOperationTransformer;
+import org.jboss.as.controller.operations.OperationAttachments;
+import org.jboss.as.controller.registry.Resource;
+import org.jboss.dmr.ModelNode;
+
+/**
+ *
```
bstansberry: This class needs lots of comments explaining what it does. DomainOperationTransformer and OperationAttachments.SLAVE_SERVER_OPERATION_TRANSFORMERS are not commonly used or well understood so please explain how with comments what the effect of the code will be.

Also, if any of these methods can have lower visibility, please do that.

host-controller/src/main/java/org/jboss/as/domain/controller/operations/coordination/ServerOperationResolver.java
```diff
@@ -538,13 +542,28 @@ private static ModelNode getDomainModel(OperationContext context) {
...
                 Set<ServerIdentity> servers = getServersForGroup(groupName, host, localHostName, serverProxies);
-                ModelNode serverOp = operation.clone();
-                PathAddress serverAddress = address.subAddress(1);
-                serverOp.get(OP_ADDR).set(serverAddress.toModelNode());
-                result = Collections.singletonMap(servers, serverOp);
+                if (address.getLastElement().getKey().equals(DEPLOYMENT)) {
+                    PathAddress serverAddress = address.subAddress(1);
+                    serverOp.get(OP_ADDR).set(serverAddress.toModelNode());
+                    result = Collections.singletonMap(servers, serverOp);
+                } else if (COMPOSITE.equals(operation.get(OP).asString())) { //This should be a redeploy affected transformed operation
```
bstansberry: This needs more of a comment explaining how this would even happen as it's not something people will expect. Probably point people to AffectedDeploymentOverlay.

server/src/main/java/org/jboss/as/server/deploymentoverlay/AffectedDeploymentOverlay.java
ehsavoie: The operation can't be run directly on a HC

bstansberry: Doh!

ehsavoie:
retest this please

ehsavoie:
Mixed domain tests + updated tests in WildFly full: https://github.com/ehsavoie/wildfly/tree/WFCORE-2229

bstansberry:
retest this please

bstansberry:
retest this please

host-controller/src/main/java/org/jboss/as/domain/controller/logging/DomainControllerLogger.java
```diff
@@ -802,4 +802,7 @@
 
     @Message(id = 97, value = "Cannot explode a subdeployment of an unexploded deployment")
     OperationFailedException cannotExplodeSubDeploymentOfUnexplodedDeployment();
+
+    @Message(id = 98, value = "You are trying to redeploy rutimes not affected by the overlay")
```
bstansberry: typo

s/rutimes/runtimes

The method should also be redeployingUnaffectedRuntimes (typo in 'affected')

bstansberry: Actually better is ""You are trying to redeploy deployments not affected by the overlay", as "runtimes" has no clear meaning. Do you have the list of invalid names? If so, or if it's easy to get it, it's better to include the list in the message.

host-controller/src/main/java/org/jboss/as/domain/controller/operations/DomainDeploymentOverlayRedeployLinksHandler.java
```diff
@@ -0,0 +1,84 @@
...
+        if (domainRoot) {
+            AffectedDeploymentOverlay.redeployLinksAndTransformOperationForDomain(context, runtimeNames, removeOperation);
+        } else {
+            AffectedDeploymentOverlay.redeployLinksAndTransformOperation(context, removeOperation, context.getCurrentAddress().getParent(), runtimeNames);
+        }
+    }
+
+    private boolean isRedeployAfterRemoval(ModelNode operation) {
+        return operation.hasDefined(REMOVE) && operation.get(REMOVE).asBoolean();
```
bstansberry: I don't think this is right. AIUI "remove" is never a param to an operation model node, it is the value of its "operation" child node.

ehsavoie: I set it in https://github.com/wildfly/wildfly-core/pull/2115/files/112f59a8dda4d52f86fd32a1f3635d6c6cb6a462#diff-9e579e339e84bbfe7b4638d2f6e40394R68
This is not a 'real' parameter. Maybe I should rename it

PR #2117, opened 2017-01-24
[WFCORE-2223] Drop pointless validation of jboss.modules.dir

bstansberry:
https://issues.jboss.org/browse/WFCORE-2223

The only use of this system property is to drive the value of a deprecated attribute in the core-service=[server|host]-environment resource. There is no good reason to validate it and doing so breaks some embedded use cases.


bstansberry:
Jeff did basically the same PR so I'll take that as a code review. :)

PR #2123, opened 2017-01-26
[WFCORE-1987] - add warning capability to core.

baranowb:
WFCORE: https://issues.jboss.org/browse/WFCORE-1987


bstansberry:
@baranowb This looks ok but it needs integration tests in both testsuite/standalone and testsuite/domain. Particularly domain as we need assurance that the output is as expected and won't regress.

baranowb:
@bstansberry : https://github.com/wildfly/wildfly/pull/9646 is this enough? There is also test to cover https://issues.jboss.org/browse/WFLY-8053

bstansberry:
@baranowb I made a comment on the other about asserting the output of the server-group part of the response in the domain test. But other than that, yes, that's what I was looking for.

Those tests should be in this PR though. And it looks like there's a response structure problem of some sort in the domain response, if the test passes. The test asserts response-headers is a child of result.

PR #2124, opened 2017-01-26
Fix for WFCORE-2223, removed check of jboss.modules.dir value.

jfdenise:
Trivial removal of jboss.modules.dir value check. This fix allows to start multiple embedded server in CLI  from a JBOSS home not containing modules directory (JBOSS_MODULEPATH being set).
No unit tests added. Manual testing.

### DIFF
--- a/server/src/main/java/org/jboss/as/server/ServerEnvironment.java
+++ b/server/src/main/java/org/jboss/as/server/ServerEnvironment.java
@@ -406,8 +406,6 @@ public class ServerEnvironment extends ProcessEnvironment implements Serializabl
             File tmp = getFileFromProperty(MODULES_DIR, props);
             if (tmp == null) {
                 tmp = new File(homeDir, "modules");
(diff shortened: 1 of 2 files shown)


luck3y:
@jfdenise When you say multiple embedded servers, are they all started in the same CLI session?

jfdenise:
@luck3y , I should have been more precise. In sequence. Start an embedded server, stop it, then start a new one. At any time, a single embedded server in CLI.

bstansberry:
This is a duplicate of #2117 (but missing some related bits).

We avoid deleting methods from the xxxLogger files; instead we comment them out. This helps document the reason for gaps in the sequence and perhaps reduces the risk that people will fill in the gaps with new, incompatible messages. Git annotate won't help you understand why something was removed.

jfdenise:
@bstansberry , thank you for the "comment" instead of remove.  Your change also cover HostController. I will close this one.

PR #2128, opened 2017-01-27
[WFCORE-2203]: Outcome from /host=x/server-config=y:remove() operation in managed domain with multiple hosts lists out all the steps in result

ehsavoie:
The synchronization model operation after an operation  between the HC and the DC shouldn't use the current operation's response.

Jira: https://issues.jboss.org/browse/WFCORE-2203


bstansberry:
I removed the "Ready for Merge" label because Emmanuel wants to follow up on some concerns he had about what happens in case of sync failure; i.e. whether this change will result in the failure being hidden from the users.

bstansberry:
When testing this on master-ignore the new test failed:

https://ci.wildfly.org/viewLog.html?buildId=47369&tab=buildResultsDiv&buildTypeId=WildFlyCore_MasterIgnoreLinux

ehsavoie:
removing the organization attribute in the domain.xml that caused the issue.

bstansberry:
retest this please

PR #2053, opened 2017-01-06
 [WFCORE-2089]: Infinite wildfly boot on StartException in service start. Also WFCORE-2238

ehsavoie:
Asynchronous start of IdentityService so that a failure will break the boot.

Jira: https://issues.jboss.org/browse/WFCORE-2089
JBEAP: https://issues.jboss.org/browse/JBEAP-7653

PR #2134, opened 2017-01-30
[WFCORE-2238] Remove support for inflow-security-domains on the access=identity resource.

darranl:

Resolves https://issues.jboss.org/browse/WFCORE-2238


bstansberry:
Added a hold label as the plan is @ehsavoie will add this changes as a commit in #2053 which otherwise would need to be rebased.

bstansberry:
We'll go with #2053

PR #2135, opened 2017-01-30
[WFCORE-2244] Add a dependencies phase for Elytron.

darranl:

Resolves https://issues.jboss.org/browse/WFCORE-2244

### DIFF
--- a/server/src/main/java/org/jboss/as/server/deployment/Phase.java
+++ b/server/src/main/java/org/jboss/as/server/deployment/Phase.java
@@ -380,6 +380,7 @@ public enum Phase {
     public static final int DEPENDENCIES_SEAM                           = 0x0A01;
     public static final int DEPENDENCIES_WS                             = 0x0C00;
     public static final int DEPENDENCIES_SECURITY                       = 0x0C50;
+    public static final int DEPENDENCIES_ELYTRON                        = 0x0C51;
     public static final int DEPENDENCIES_JAXRS                          = 0x0D00;
     public static final int DEPENDENCIES_JAXRS_SPRING                   = 0x0D80;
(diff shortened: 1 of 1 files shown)


server/src/main/java/org/jboss/as/server/deployment/Phase.java
```diff
@@ -380,6 +380,7 @@ public Phase next() {
     public static final int DEPENDENCIES_SEAM                           = 0x0A01;
     public static final int DEPENDENCIES_WS                             = 0x0C00;
     public static final int DEPENDENCIES_SECURITY                       = 0x0C50;
+    public static final int DEPENDENCIES_ELYTRON                        = 0x0C51;
```
kabir: @dmlloyd Should there be a bigger gap here? I know you had some opinions on this in the past. On the other hand, this is pretty identical to security, so perhaps no gap is ok in this case.

darranl:
I did think about a bigger gap but TBH this is a replacement for security.

server/src/main/java/org/jboss/as/server/deployment/Phase.java
bstansberry: Perhaps you are remembering David chuckling at peoples' habit of using 0x0A50 as the "halfway" value between 0x0A00 and 0x0B00, instead of 0x0A80. :)

bstansberry: To follow up on this for the record, since the new one is a replacement for the one before it in the sequence, there is no use case for inserting others in between so a gap is not needed.

And if some day that proves wrong it's always possible to change the sequence number.

PR #2138, opened 2017-01-31
[WFCORE-2252] Reject 'required' attribute definitions that also have …

bstansberry:
…a default

https://issues.jboss.org/browse/WFCORE-2252

Currently just an experiment.


bstansberry:
Experiment over.

PR #2147, opened 2017-02-03
[WFCORE-2270] Only allow a single registration of a runtime (not 'possible') capabi…

bstansberry:
…lity with the same name and scope

This will require updates in full to clean out some subsystem test stuff that does duplicate registration, plus to fix a clustering issue where some items are duplicated.


bstansberry:
Superceded by #2155

PR #2155, opened 2017-02-07
[WFCORE-2270] Part 2 -- Change the default behavior to only allow a single registration of a runtime (not 'possible') capability with the same name and scope

bstansberry:
https://issues.jboss.org/browse/WFCORE-2270

This is a follow up to #2154 with a second commit changing default behavior.

The expectation is #2154  would be merged, then a release performed, then uses in full that need to use the new API in #2154 would do so, and then this would be merged.

This will fail CI until the full changes are made.

https://github.com/wildfly/wildfly/pull/9645 has the full changes. https://ci.wildfly.org/viewLog.html?buildId=46237&tab=buildResultsDiv&buildTypeId=WF_WildFlyCoreIntegrationExperiments is a custom job testing this in combination with that PR.

### DIFF
--- a/controller/src/main/java/org/jboss/as/controller/capability/RuntimeCapability.java
+++ b/controller/src/main/java/org/jboss/as/controller/capability/RuntimeCapability.java
@@ -70,9 +70,7 @@ public class RuntimeCapability<T> extends AbstractCapability  {
     }
 
     // Default value for allowMultipleRegistrations.
-    // I use this constant because I intend to shortly change the
-    // default and don't want to risk mistake by changing many places
-    private static final boolean ALLOW_MULTIPLE = true;
+    private static final boolean ALLOW_MULTIPLE = false;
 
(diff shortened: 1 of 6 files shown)


bstansberry:
retest this please

jmesnil:
The error is coming from the messaging-activemq subsystem.
The name of the ActiveMQ server is missing in the capability name to distinguish between 2 broadcast-groups with the same names belonging to 2 different ActiveMQ servers (`default` and `backup-server`).

@bstansberry if you modify the name of the 2nd broadcast-group (at https://github.com/wildfly/wildfly/blob/master/messaging-activemq/src/main/resources/subsystem-templates/messaging-activemq-colocated.xml#L95) to bg-group2, this PR should pass.

I have created https://issues.jboss.org/browse/WFLY-8174 to fix the underlying issue in the messaging-activemq subsystem

bstansberry:
retest this please

PR #2156, opened 2017-02-07
Move to XNIO 3.5

dmlloyd:


kabir:
Seems included in https://github.com/wildfly/wildfly-core/pull/2159

PR #2160, opened 2017-02-08
[WFCORE-2278] use PicketBox compatible masked passwords

pskopek:
https://issues.jboss.org/browse/WFCORE-2278
upstream issue: https://issues.jboss.org/browse/JBEAP-8544

### DIFF
--- a/controller/src/main/java/org/jboss/as/controller/logging/ControllerLogger.java
+++ b/controller/src/main/java/org/jboss/as/controller/logging/ControllerLogger.java
@@ -3448,8 +3448,8 @@ public interface ControllerLogger extends BasicLogger {
 
     // ---- Credential Store related messages -----------
 
-    @Message(id = 423, value = "Name of the credential store has to be specified in this credential-reference")
-    IllegalStateException nameOfCredentialStoreHasToBeSpecified();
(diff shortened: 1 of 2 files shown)


controller/src/main/java/org/jboss/as/controller/logging/ControllerLogger.java
```diff
@@ -3448,8 +3448,8 @@ XMLStreamException discoveryOptionsMustBeDeclared(String adminOnlyCmd, String po
 
     // ---- Credential Store related messages -----------
 
-    @Message(id = 423, value = "Name of the credential store has to be specified in this credential-reference")
-    IllegalStateException nameOfCredentialStoreHasToBeSpecified();
+    @Message(id = 423, value = "Masked password command has the wrong format.\nUsage: MASK-<encoded secret>;<salt>;<iteration count> where <salt>=UTF-8 characters, <iteration count>=reasonable sized positive integer")
```
bstansberry: @jamezp is the "\n" here the correct syntax?

Granted, I'm sure we have lots of other places we use it.

jamezp: No we should use `%n` so that it's platform independent.

bstansberry: I'll clean this up separately.

PR #2163, opened 2017-02-08
WFCORE-2265/WFCORE-2279 Fixes for override models

pferraro:
https://issues.jboss.org/browse/WFCORE-2279
https://issues.jboss.org/browse/WFCORE-2265

### DIFF
--- a/controller/src/main/java/org/jboss/as/controller/operations/global/GlobalOperationHandlers.java
+++ b/controller/src/main/java/org/jboss/as/controller/operations/global/GlobalOperationHandlers.java
@@ -52,6 +52,7 @@ import java.util.Map;
 import java.util.Set;
 import java.util.concurrent.atomic.AtomicBoolean;
 import java.util.function.Predicate;
+import java.util.stream.Collectors;
 
 import org.jboss.as.controller.ModelVersion;
 import org.jboss.as.controller.OperationContext;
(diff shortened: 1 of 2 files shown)


controller/src/main/java/org/jboss/as/controller/operations/global/GlobalOperationHandlers.java
```diff
@@ -1004,65 +1004,47 @@ public void execute(OperationContext context, ModelNode operation) throws Operat
...
-                assert !childAddr.equals(target) : "Alias was not translated";
-                PathAddress targetParent = target.subAddress(0, target.size() - 1);
-                Resource parentResource = context.readResourceFromRoot(targetParent, false);
-                if (parentResource != null) {
-                    PathElement targetElement = target.getLastElement();
-                    if (targetElement.isWildcard()) {
-                        set.addAll(parentResource.getChildrenNames(targetElement.getKey()));
-                    } else if (parentResource.hasChild(targetElement)) {
-                        set.add(element.getValue());
+                ImmutableManagementResourceRegistration childRegistration = registry.getSubModel(PathAddress.pathAddress(path));
```
bstansberry: I think the null check that was at L1015 before may be necessary. Unlike the Resource tree, we don't do a copy-on-write/publish-on-commit deal when an op changes the MRR tree (yet). So a read op that calls this can have an MRR disappear underneath it.

bstansberry:
Other than the null check thing, this looks good; much simpler.

controller/src/main/java/org/jboss/as/controller/operations/global/GlobalOperationHandlers.java
pferraro: OK - will add a check for null child registration.

pferraro: Done.

PR #2164, opened 2017-02-09
[WFCORE-1529] Tweak management thread pools to decrease thread churn

bstansberry:
https://issues.jboss.org/browse/WFCORE-1529

1) Bump managment-handler-thread pool thread timeout from 1 to 10 mins to retain threads during active CLI sessions (see https://issues.jboss.org/browse/WFCORE-2282 for proposed future work related to this.)
2) ServerService pool on a standalone server keeps 1 core thread
3) On a domain server, it keeps 3, as comms with the HC involve more async tasks
4) ServerService core pool size can be set via system property
5) HostControllerService pool keeps 1 core thread

With these settings light CLI activity doesn't produce pool churn. The management-handler-thread pool will still churn if activity is infrequent, with gaps longer than 10 mins.

Repeated complex tasks like deploy ops will still produce churn in ServerService thread pool.

In both cases where churn still happens, the alternative is never releasing threads and wasting resources, which seems worse than a bit of churn.


bstansberry:
retest this please

PR #2165, opened 2017-02-09
[WFCORE-1789] PreparedResponseTestCase from wf-core-eap fails intermittently

mjurc:
Issue: https://issues.jboss.org/browse/WFCORE-1789

* Adding an expected exception after server reload to reloadServer test.

### DIFF
--- a/testsuite/manualmode/src/test/java/org/wildfly/core/test/standalone/mgmt/PreparedResponseTestCase.java
+++ b/testsuite/manualmode/src/test/java/org/wildfly/core/test/standalone/mgmt/PreparedResponseTestCase.java
@@ -21,7 +21,6 @@
 package org.wildfly.core.test.standalone.mgmt;
 
 import static org.jboss.as.controller.descriptions.ModelDescriptionConstants.EXTENSION;
-import static org.jboss.as.controller.descriptions.ModelDescriptionConstants.RELOAD;
 import static org.jboss.as.controller.descriptions.ModelDescriptionConstants.SHUTDOWN;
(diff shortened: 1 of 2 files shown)


testsuite/manualmode/src/test/java/org/wildfly/core/test/standalone/mgmt/PreparedResponseTestCase.java
```diff
@@ -98,7 +100,16 @@ public void reloadServer() throws Exception {
...
+            try {
+                ModelNode result = managementClient.getControllerClient().execute(
+                        Operations.createOperation(RELOAD, new ModelNode().setEmptyList()));
+                Assert.assertEquals("success", result.get(OUTCOME).asString());
+            } catch (IOException e) {
+                final Throwable cause = e.getCause();
+                if (!(cause instanceof ExecutionException) && !(cause instanceof CancellationException)) {
+                    throw new RuntimeException(e);
+                } // else ignore, this might happen if the channel gets closed before we got the response
+            }
```
bstansberry: Thanks for this. We have a util class, ServerReload, that should be used for this kind of thing. It looks like moving this test to it was missed. The block here looks the same as ServerReload.executeReload(managementClient.getControllerClient(), false) so please use that instead.

I thought it might be, but the ServerReload.executeReloadAndWaitForCompletion call is not appropriate here as this test is asserting some intermediate state while awaiting completion.

mjurc:
@bstansberry : I have committed the required changes.

bstansberry:
@mjurc Thanks! I hadn't noticed that method wasn't public. Good; it's better making that public.

PR #2166, opened 2017-02-09
[WFCORE-1673] Integration tests for controller configuration+aliases

mschvarc:
JIRA: https://issues.jboss.org/browse/WFCORE-1673
JBEAP: https://issues.jboss.org/browse/JBEAP-5432

Added integration tests for server aliases defined in jboss-cli.xml

### DIFF
--- a/testsuite/standalone/src/test/java/org/jboss/as/test/integration/management/cli/CliControllerConfigTestCase.java
+++ b/testsuite/standalone/src/test/java/org/jboss/as/test/integration/management/cli/CliControllerConfigTestCase.java
@@ -1,0 +1,420 @@
+/*
+ * JBoss, Home of Professional Open Source.
+ * Copyright 2017 Red Hat, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *   http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.jboss.as.test.integration.management.cli;
+
+import org.jboss.as.remoting.Protocol;
+import org.jboss.as.test.shared.TestSuiteEnvironment;
+import org.junit.After;
+import org.junit.Before;
+import org.junit.Test;
+import org.junit.runner.RunWith;
+import org.wildfly.core.testrunner.WildflyTestRunner;
+
+import java.io.BufferedWriter;
+import java.io.File;
+import java.io.FileOutputStream;
+import java.io.IOException;
+import java.io.OutputStreamWriter;
+
+import static org.jboss.as.remoting.Protocol.HTTPS_REMOTING;
+import static org.jboss.as.remoting.Protocol.HTTP_REMOTING;
+import static org.jboss.as.remoting.Protocol.REMOTE;
+import static org.junit.Assert.assertTrue;
+import static org.junit.Assert.fail;
+
+/**
+ * Tests for jboss-cli.xml controller aliases
+ *
+ * @author Martin Schvarcbacher
+ */
+@RunWith(WildflyTestRunner.class)
+public class CliControllerConfigTestCase {
+
+    private static final Protocol TEST_RUNNER_PROTOCOL = HTTP_REMOTING;
+    private static final String CONTROLLER_ALIAS_NAME = "Test_Suite_Server1_Name";
+    private static final int INVALID_PORT = TestSuiteEnvironment.getServerPort() - 1;
+    private static final String DISCONNECTED_PROMPT = "[disconnected /]";
+    private static final String CONNECTED_PROMPT = "@";
+    private static final String READ_SERVER_STATE = ":read-attribute(name=server-state)";
+    private File tempJbossConfigFile;
+
+    @Before
+    public void prepareFiles() {
+        try {
+            tempJbossConfigFile = File.createTempFile("tmp-jboss-cli", ".xml");
+        } catch (IOException e) {
+            fail(e.getLocalizedMessage());
+        }
+    }
+
+    @After
+    public void deleteFiles() {
+        assertTrue("failed to delete test config file", tempJbossConfigFile.delete());
+    }
+
+    /**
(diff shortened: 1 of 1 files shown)


bstansberry:
@jfdenise please approve

testsuite/standalone/src/test/java/org/jboss/as/test/integration/management/cli/CliControllerConfigTestCase.java
```diff
@@ -0,0 +1,438 @@
...
+        if (connectToAlias) {
+            cli.addCliArgument("--controller=" + CONTROLLER_ALIAS_NAME);
+        }
+        return cli;
+    }
+
+    private void assertDisconnected(String output) {
+        assertTrue("connected when disconnected is expected", output.contains("Failed to connect to the controller"));
+        assertTrue("connected when disconnected is expected", output.contains("The connection failed"));
+        assertFalse("connected when disconnected is expected", output.contains("standalone@"));
```
jfdenise: Is this check really required? Did you observe a prompt displayed with a failed connection?

testsuite/standalone/src/test/java/org/jboss/as/test/integration/management/cli/CliControllerConfigTestCase.java
```diff
@@ -0,0 +1,438 @@
...
+    }
+
+    private void assertConnected(String output) {
+        String expectedIP = "@" + TestSuiteEnvironment.getServerAddress() + ":" + TestSuiteEnvironment.getServerPort();
```
jfdenise: We have some timing issue where the prompt would be not displayed when you ask for it. I would advise you to use the CLIProcessWrapper.pushLineAndWaitForResults(cmd, prompt). 
Instead of having: --command=:read-attribute(name=server-state)
You could call boolean ret = pushLineAndWaitForResults(":read-attribute(name=server-state)", null);
null prompt means the default connected prompt. ret will be true if the prompt is found. So should be safer to use that instead of assertConnected.

mschvarc: I've removed the assertConnected / assertDisconnected methods and replaced them with pushLineAndWaitForResults and assert on its returned state. 
When testing invalid controller settings, pushLineAndWaitForResults(":read-attribute(name=server-state)", null); returns true even on disconnected server. For disconnected state now I’m using: "[disconnected /]" and "@" for connected state. Would this be an acceptable solution? 
cli.pushLineAndWaitForResults("connect", null); succeeds in all cases,  cli.pushLineAndWaitForResults("connect", "[disconnected /]") only when invalid controller is specified.

testsuite/standalone/src/test/java/org/jboss/as/test/integration/management/cli/CliControllerConfigTestCase.java
```diff
@@ -0,0 +1,441 @@
...
+                .setDefaultControllerProtocol(HTTPS_REMOTING)
+                .setAliasControllerPort(TestSuiteEnvironment.getServerPort())
+                .setAliasControllerProtocol(TEST_RUNNER_PROTOCOL);
+        jbossCliConfig.writeJbossCliConfig(tempJbossConfigFile);
+
+        CliProcessWrapper cli = getTestCliProcessWrapper(true);
+        try {
+            cli.executeInteractive();
+            cli.pushLineAndWaitForResults("connect");
+            boolean returnState = cli.pushLineAndWaitForResults(READ_SERVER_STATE, CONNECTED_PROMPT);
```
mschvarc: I'm using CONNECTED_PROMPT instead of null; for example, setting TestSuiteEnvironment.getServerPort() + 100 with pushLineAndWaitForResults(..,null) passes the test when it should fail.

mschvarc:
Hello @jfdenise , I've made the changes you have requested. For your comment on assertConnected, I've had to make a slight change (please see my response to your comment). 
Thank you.

testsuite/standalone/src/test/java/org/jboss/as/test/integration/management/cli/CliControllerConfigTestCase.java
```diff
@@ -0,0 +1,441 @@
...
+    @Test
+    public void testImplicitSettings() {
+        JbossCliConfig jbossCliConfig = new JbossCliConfig(); //empty config
+        jbossCliConfig.writeJbossCliConfig(tempJbossConfigFile);
+        CliProcessWrapper cli = getTestCliProcessWrapper(false);
+        try {
+            cli.executeInteractive();
+            cli.pushLineAndWaitForResults("connect");
+            boolean returnState = cli.pushLineAndWaitForResults(READ_SERVER_STATE, CONNECTED_PROMPT);
+            assertTrue(returnState);
```
jfdenise: This test case is failing on both windows and linux. Could you replace assertTrue(returnState); by assertTrue("Failing output:" +cli.getOutput(), returnState); It should help understand what is going on.
Thanks.

jfdenise:
Hi @mschvarc , I am fine with your changes. One of the test is failing, I suggested adding some debug output.
JF

testsuite/standalone/src/test/java/org/jboss/as/test/integration/management/cli/CliControllerConfigTestCase.java
mschvarc: The following output is from our Jenkins node:
```
The controller is not available at localhost:9990: java.net.ConnectException: WFLYPRT0053: Could not connect to remote+http://localhost:9990. The connection failed: WFLYPRT0053: Could not connect to remote+http://localhost:9990. The connection failed: Connection refused
[disconnected /] :read-attribute(name=server-state)

```
I cannot reproduce this error locally. The CLI should be able to connect to the controller (remote+http://localhost:9990) using default values, no?

jfdenise: Connection refused means that the server is not running. Any chance you can access to the test output (to see what happened to the server)?

jfdenise: I think that the issue is related to the CLI having the "localhost" hard coded as default. 
In case your machine is not properly configured, it could fail (as we can see it). Tests are generally using TestSuiteEnvironment.getServerAddress().
I suggest that you remove this test case, it doesn't test controller configuration/alias. Right?

mschvarc: Thank you for your help, I've now removed the test case.

mschvarc:
Retest this please

jfdenise:
@bstansberry, @mschvarc I am ok wit the test now. Thanks.

PR #2168, opened 2017-02-09
WFCORE-2139 domain-organization attribute causes error when starting domain

luck3y:
Also, aliases this value to the /host= read, which was always unset.

### DIFF
--- a/host-controller/src/main/java/org/jboss/as/host/controller/model/host/HostResourceDefinition.java
+++ b/host-controller/src/main/java/org/jboss/as/host/controller/model/host/HostResourceDefinition.java
@@ -22,6 +22,7 @@
 
 package org.jboss.as.host.controller.model.host;
 
+import static org.jboss.as.controller.descriptions.ModelDescriptionConstants.DOMAIN_ORGANIZATION;
 import static org.jboss.as.controller.descriptions.ModelDescriptionConstants.HOST;
 
 import org.jboss.as.controller.BootErrorCollector;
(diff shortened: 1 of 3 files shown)


host-controller/src/main/java/org/jboss/as/host/controller/model/host/HostResourceDefinition.java
```diff
@@ -300,7 +304,16 @@ protected boolean requiresRuntime(OperationContext context) {
...
         });
         hostRegistration.registerReadWriteAttribute(ORGANIZATION_IDENTIFIER, null, new ModelOnlyWriteAttributeHandler(ORGANIZATION_IDENTIFIER));
-        hostRegistration.registerReadOnlyAttribute(DOMAIN_ORGANIZATION_IDENTIFIER, null);
+        // provide the domain-organization, this was defined here, but never had any handlers or storage defined.
+        hostRegistration.registerReadOnlyAttribute(DOMAIN_ORGANIZATION_IDENTIFIER, new ReadAttributeHandler() {
+                    @Override
+                    public void execute(OperationContext context, ModelNode operation) throws OperationFailedException {
+                        ModelNode rootModel = context.readResourceFromRoot(PathAddress.EMPTY_ADDRESS,false).getModel();
+                        final String domainOrganization = (rootModel.hasDefined(DOMAIN_ORGANIZATION)) ? rootModel.get(DOMAIN_ORGANIZATION).asString() : "undefined";
+                        context.getResult().set(domainOrganization);
```
bstansberry: This should just be 

context.getResult(rootModel.get(DOMAIN_ORGANIZATION));

We don't use the String "undefined" to mean an undefined node.

luck3y: Thanks, fixed.

luck3y:
retest this please

luck3y:
retest this please

PR #2169, opened 2017-02-09
Use %n instead of \n in XXXLogger messages and in LocalDescriptions.p…

bstansberry:
…roperties

@jamezp please approve. I did a search for \n in the Logger and .properties files and this is what I found.

### DIFF
--- a/controller/src/main/java/org/jboss/as/controller/logging/ControllerLogger.java
+++ b/controller/src/main/java/org/jboss/as/controller/logging/ControllerLogger.java
@@ -3449,7 +3449,7 @@ public interface ControllerLogger extends BasicLogger {
 
     // ---- Credential Store related messages -----------
 
-    @Message(id = 423, value = "Masked password command has the wrong format.\nUsage: MASK-<encoded secret>;<salt>;<iteration count> where <salt>=UTF-8 characters, <iteration count>=reasonable sized positive integer")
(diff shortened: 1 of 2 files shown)


threads/src/main/resources/org/jboss/as/threads/LocalDescriptions.properties
```diff
@@ -14,7 +14,7 @@ thread-factory=A thread factory (implementing java.util.concurrent.ThreadFactory
 thread-factory.add=Adds a thread factory
 thread-factory.name=The name of the created thread factory.
 thread-factory.group-name=Specifies the name of a  thread group to create for this thread factory.
-thread-factory.thread-name-pattern=The template used to create names for threads.  The following patterns may be used:\n\t%% - emit a percent sign\n\t%t - emit the per-factory thread sequence number\n\t%g - emit the global thread sequence number\n\t%f - emit the factory sequence number\n\t%i - emit the thread ID.
+thread-factory.thread-name-pattern=The template used to create names for threads.  The following patterns may be used:%n\t%% - emit a percent sign%n\t%t - emit the per-factory thread sequence number%n\t%g - emit the global thread sequence number%n\t%f - emit the factory sequence number%n\t%i - emit the thread ID.
```
jamezp: This one may not work. IIRC the strings from these resource bundles are not run through a `java.util.Formatter` like `String.format()`.

bstansberry: Thanks; I'll remove that.

PR #2172, opened 2017-02-10
[WFCORE-2025] Correct the priority when selecting the SSLContext to use with the CLI.

darranl:

The SSLContext priority order becomes: -
 1. The SSLContext from jboss-cli.xml
 2. An SSLContext from wildfly-config.xml
 3. The default SSLContext that can prompt for trust decisions.

The following issue can be resolved with this PR: -
  https://issues.jboss.org/browse/WFCORE-2025

### DIFF
--- a/cli/src/main/java/org/jboss/as/cli/impl/CLIModelControllerClient.java
+++ b/cli/src/main/java/org/jboss/as/cli/impl/CLIModelControllerClient.java
@@ -27,6 +27,7 @@ import static java.security.AccessController.doPrivileged;
 import java.io.IOException;
 import java.net.URI;
 import java.net.URISyntaxException;
+import java.security.GeneralSecurityException;
 import java.security.PrivilegedAction;
 import java.util.Map;
 import java.util.concurrent.BlockingQueue;
@@ -57,6 +58,9 @@ import org.jboss.remoting3.CloseHandler;
 import org.jboss.remoting3.Connection;
 import org.jboss.remoting3.Endpoint;
 import org.jboss.threads.JBossThreadFactory;
+import org.wildfly.security.auth.client.AuthenticationContext;
+import org.wildfly.security.auth.client.AuthenticationContextConfigurationClient;
+import org.wildfly.security.auth.client.MatchRule;
 import org.xnio.OptionMap;
 import org.xnio.http.RedirectException;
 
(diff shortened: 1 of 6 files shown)


cli/src/main/java/org/jboss/as/cli/impl/CLIModelControllerClient.java
```diff
@@ -146,6 +150,23 @@ public void close() throws IOException {
             throw new IOException("Failed to create URI" , e);
         }
 
+        if (sslContext != null && fallbackSslContext == false) {
```
jfdenise: perhaps use !fallbackSslContext

cli/src/main/java/org/jboss/as/cli/impl/CommandContextImpl.java
```diff
@@ -252,6 +252,8 @@
     private final char[] password;
     /** flag to disable the local authentication mechanism */
     private final boolean disableLocalAuth;
+    /** Flag to indicate that the SSLContext was the defailt context and not configured */
```
jfdenise: defailt ==> default

cli/src/main/java/org/jboss/as/cli/impl/CLIModelControllerClient.java
darranl: Never been a fan of '!' especially after wasted time on support cases.

cli/src/main/java/org/jboss/as/cli/impl/CommandContextImpl.java
darranl: Lets leave that and pick it up next time we are here.

protocol/src/main/java/org/jboss/as/protocol/ProtocolConnectionConfiguration.java
```diff
@@ -103,12 +102,13 @@ public void setEndpoint(Endpoint endpoint) {
...
+    @Deprecated
     public OptionMap getOptionMap() {
-        return optionMap;
+        return OptionMap.EMPTY;
     }
 
+    @Deprecated
     public void setOptionMap(OptionMap optionMap) {
-        this.optionMap = optionMap;
     }
```
bstansberry: @darranl this change to how optionMap is handled looks to me to be a breaking change. RemoteDomainConnectionService.start is using this to configure heartbeats and a read timeout.

PR #2182, opened 2017-02-12
WFCORE-2066  Kill and destroy operations on the server-group

yersan:
This issue adds Kill and Destroy operations at server-group level. It adds some refactored code in ManagedServer#ServerStopTask and removes an unused operation-id from it.

Jira issue is:
https://issues.jboss.org/browse/WFCORE-2066

### DIFF
--- a/controller/src/main/java/org/jboss/as/controller/descriptions/ModelDescriptionConstants.java
+++ b/controller/src/main/java/org/jboss/as/controller/descriptions/ModelDescriptionConstants.java
@@ -326,6 +326,7 @@ public class ModelDescriptionConstants {
     public static final String ORGANIZATION = "organization";
     public static final String OPERATION_DATE = "operation-date";
     public static final String OPERATION_HEADERS = "operation-headers";
+    public static final String OPERATION_ID = "operation-id";
     public static final String OPERATION_NAME = "operation-name";
     public static final String OPERATIONS = "operations";
     public static final String OPTIONS = "options";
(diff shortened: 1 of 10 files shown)


yersan:
retest this please

bstansberry:
Re: the test failures, we've added notifications of lifecycle changes and tests that assert the notifications are being emitted in the patterns we expect. So perhaps there is a problem with this PR or perhaps the tests are assuming something invalid. I haven't investigated.

The PR seems ok but will need to have transformers added, as slave HCs running previous versions will not be able to handle these new operations. The ServerGroupTransformers class is handling registering those. It will need to be a bit different from what is there though, as what is there is for transforming to slaves running kernel management API version 1.8 and earlier (i.e. <= EAP 6.4) while this will need to be for version 4.1 and earlier.

yersan:
All test pass without any issues in my local computer. Let's try again.

retest this please

yersan:
retest this please

yersan:
retest this please

bstansberry:
The transformer stuff looks ok. This will need to be held of WF Core 4 though, at which time there will be a new version (i.e. target would be VERSION_5_0 instead of VERSION_4_1). Might as well let this wait until we open up for work on 4 and then we can set the exact version.

@kabir I could use some re-education on these transformer chains. We have a version 4.2 now (WF 10.1) that we should transform to. Does simply replacing any chained build aiming at VERSION_4_1 with VERSION_4_2 automatically handle 4.1 as well?

yersan:
Branch rebased to resolve conflicts with master

yersan:
- Transformers have been updated accoring to current version
- Added some tests to check them for kill/destroy servers

PR #2184, opened 2017-02-14
[WFCORE-2294] Improve timeouts in WF-CORE TS

marekkopecky:
Improve timeouts in WF-CORE TS
* SuspendResumeTestCase.java should use TimeoutUtil for timeout
* Parametrize timeouts in TestRunner 
* CliProcessWrapper.java should use TimeoutUtil for timeout

This stabilizes TS on slower machines.

### DIFF
--- a/testsuite/shared/src/main/java/org/jboss/as/test/integration/management/cli/CliProcessWrapper.java
+++ b/testsuite/shared/src/main/java/org/jboss/as/test/integration/management/cli/CliProcessWrapper.java
@@ -22,6 +22,8 @@
 
 package org.jboss.as.test.integration.management.cli;
 
+import org.jboss.as.test.shared.TimeoutUtil;
+
 import java.io.BufferedWriter;
 import java.io.IOException;
 import java.io.InputStream;
@@ -203,7 +205,7 @@ public class CliProcessWrapper extends CliProcessBuilder {
         }
     }
 
-    private int resultTimeout = 10000;
+    private int resultTimeout = TimeoutUtil.adjust(20000);
     private int resultInterval = 100;
 
     private boolean waitForPrompt(String prompt) {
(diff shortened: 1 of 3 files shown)


testsuite/shared/src/main/java/org/jboss/as/test/integration/management/cli/CliProcessWrapper.java
```diff
@@ -203,7 +205,7 @@ private void pushToInput(String line) {
         }
     }
 
-    private int resultTimeout = 10000;
+    private int resultTimeout = TimeoutUtil.adjust(30000);
```
bstansberry: @jfdenise Any concerns with this? Best practice is to use TimeoutUtil.adjust. I don't know if a 30 second base value means allowing incorrect behavior a test should catch. My impression is not, but I wanted to check with you. Taking a longer time to detect a failure is fine; that's a good tradeoff for avoiding false failures due to things like long gc pauses of CI slave VMs being unscheduled under load on the machine.

jfdenise:
@bstansberry , 30secs to wait for the prompt seems to me a bit extreme. Do we have tests that are now passing with 30secs and were failing with 10secs?

bstansberry:
@jfdenise I assume so, but @marekkopecky would need to confirm.

marekkopecky:
@bstansberry, @jfdenise: 10s is not enough on some slow IPv6 machines. So I increased this timeout to 30s. But if you want, 20s may be compromise and I can increase it later, if I see timeout issue even with this commit. WDYT?

jfdenise:
@marekkopecky , of for 20secs.

marekkopecky:
Ok, thank you. Timeout in CliProcessWrapper.java is changed to 20s

bstansberry:
@marekkopecky is the factor used in TimeoutUtil.adjust used on these machines?

I'm asking more out of curiosity, or perhaps concern that we are using inadequate hardware and then having to devote expensive human resources to chasing intermittent failures. I have no objection to the change here to 20s and will merge it.

A -Dts.timeout.factor=100 means now this delay could be up to 40s. If we need that long a buffer for things to happen, something seems wrong.

PR #2194, opened 2017-02-20
[WFCORE-2312] Don't assume itemCount is always at least 2

bstansberry:
https://issues.jboss.org/browse/WFCORE-2312


kabir:
Looks good - where is the approve button?

PR #2201, opened 2017-02-23
[WFCORE-2321]: ProcessStateListenerTestCase from wf-core fails intermittently

ehsavoie:
The future is canceled in case of error.

Jira: https://issues.jboss.org/browse/WFCORE-2321
JBEAP: https://issues.jboss.org/browse/JBEAP-8595


ehsavoie:
retest this please

PR #2205, opened 2017-02-26
[WFCORE-1507][WFCORE-2327] Clean up in VM ModelControllerClient handling

bstansberry:
https://issues.jboss.org/browse/WFCORE-1507
https://issues.jboss.org/browse/WFCORE-2327

[WFCORE-1507] Expose in-vm ModelControllerClient creation via a capability
Also addresses [WFCORE-2327] by creating a standardized way of establishing a client that acts as SuperUser.

The real driver here was WFCORE-2327 but it seemed obvious the right way to fix that was by doing what was needed for WFCORE-1507.

I also added a commit that cleans up some unrelated stuff in one of the test classes touched by the main WFCORE-2327 work

### DIFF
--- a/controller-client/src/main/java/org/jboss/as/controller/client/ModelControllerClient.java
+++ b/controller-client/src/main/java/org/jboss/as/controller/client/ModelControllerClient.java
@@ -51,7 +51,9 @@ public interface ModelControllerClient extends Closeable {
      * @return the result of the operation
      * @throws IOException if an I/O error occurs while executing the operation
      */
-    ModelNode execute(ModelNode operation) throws IOException;
+    default ModelNode execute(ModelNode operation) throws IOException{
+        return execute(Operation.Factory.create(operation), OperationMessageHandler.DISCARD);
+    }
 
     /**
      * Execute an operation synchronously.
@@ -63,7 +65,9 @@ public interface ModelControllerClient extends Closeable {
      * @return the result of the operation
      * @throws IOException if an I/O error occurs while executing the operation
      */
-    ModelNode execute(Operation operation) throws IOException;
+    default ModelNode execute(Operation operation) throws IOException{
+        return execute(operation, OperationMessageHandler.DISCARD);
+    }
 
     /**
      * Execute an operation synchronously, optionally receiving progress reports.
@@ -73,7 +77,9 @@ public interface ModelControllerClient extends Closeable {
      * @return the result of the operation
      * @throws IOException if an I/O error occurs while executing the operation
      */
-    ModelNode execute(ModelNode operation, OperationMessageHandler messageHandler) throws IOException;
+    default ModelNode execute(ModelNode operation, OperationMessageHandler messageHandler) throws IOException {
+        return execute(Operation.Factory.create(operation), messageHandler);
+    }
 
     /**
      * Execute an operation synchronously, optionally receiving progress reports.
@@ -86,7 +92,11 @@ public interface ModelControllerClient extends Closeable {
      * @return the result of the operation
      * @throws IOException if an I/O error occurs while executing the operation
      */
-    ModelNode execute(Operation operation, OperationMessageHandler messageHandler) throws IOException;
+    default ModelNode execute(Operation operation, OperationMessageHandler messageHandler) throws IOException {
+        try (OperationResponse or = executeOperation(operation, messageHandler)) {
+            return or.getResponseNode();
+        }
+    }
 
     /**
      * Execute an operation synchronously, optionally receiving progress reports, with the response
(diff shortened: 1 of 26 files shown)


controller-client/src/main/java/org/jboss/as/controller/client/ModelControllerClient.java
```diff
@@ -86,7 +92,12 @@
      * @return the result of the operation
      * @throws IOException if an I/O error occurs while executing the operation
      */
-    ModelNode execute(Operation operation, OperationMessageHandler messageHandler) throws IOException;
+    default ModelNode execute(Operation operation, OperationMessageHandler messageHandler) throws IOException {
+        OperationResponse or = executeOperation(operation, messageHandler);
+        ModelNode result = or.getResponseNode();
+        or.close();
```
kabir: Should this close happen in a finally block?

controller/src/main/java/org/jboss/as/controller/LocalModelControllerClient.java
```diff
@@ -0,0 +1,66 @@
...
+ * Used for clients that operate in the same VM as the target {@link ModelController} and hence
+ * are not subject to IO failures associated with remote calls.
+ *
+ * @author Brian Stansberry
+ */
+public interface LocalModelControllerClient extends ModelControllerClient {
+
+    @Override
+    default ModelNode execute(ModelNode operation) {
+        return execute(Operation.Factory.create(operation), OperationMessageHandler.DISCARD);
```
kabir: Aren't these doing the same as the default methods in the super interface? Perhaps that is ok, I just wanted to point that out.

bstansberry: The default methods in the superinterface throw IOException so I need an implementation that does not.

controller-client/src/main/java/org/jboss/as/controller/client/ModelControllerClient.java
bstansberry: Yeah, that would be better.

testsuite/manualmode/src/test/java/org/jboss/as/test/manualmode/management/cli/CLIEmbedServerTestCase.java
```diff
@@ -795,9 +798,9 @@ private void testPathDir(final String propName, final String value) throws IOExc
             for(String prop : SERVER_PROPS) {
                 if (prop.equals(propName)) {
                     assertPath(propName, ROOT + File.separator + newStandalone + File.separator + value);
-                } else {
-                    // just make sure the unchanged property has the default basedir
-                    //assertTrue(WildFlySecurityManager.getPropertyPrivileged(prop, "").indexOf(currBaseDir) != -1);
+//                } else {
```
luck3y: The intent here is to positively assert that an unchanged property hasn't been changed inadvertently, so has the default base.dir. Was this causing a failure somewhere?

bstansberry: The assert was already commented out. I just commented out the whole else block instead of just that line to stop my IDE warning me about an empty else block.

If the assert should be uncommented, I can stick in a TODO to get that done. I don't want to uncomment it as part of this though.

bstansberry: It seems to pass locally uncommented so I'll modify this to uncomment. If it fails CI then I'll comment again with a TODO.

bstansberry:
I just updated this to address the CapabilityRegistryTestCase.testClear failure that happened once. AbstractControllerService.start needs to publish the capability registry after modifying it, not waiting for the boot thread to do it in finishBoot. This ensures there are no races between the test checking the registry and the boot thread completing.

testsuite/manualmode/src/test/java/org/jboss/as/test/manualmode/management/cli/CLIEmbedServerTestCase.java
luck3y: Ok, if it causes issues I'll take a look.

host-controller/src/main/java/org/jboss/as/host/controller/DomainModelControllerService.java
```diff
@@ -515,6 +519,10 @@ public JBossThreadFactory run() {
         });
         pingScheduler = Executors.newScheduledThreadPool(PINGER_POOL_SIZE, pingerThreadFactory);
 
+        ContentCleanerService.addServiceOnHostController(context.getChildTarget(), DomainModelControllerService.SERVICE_NAME,
+                CLIENT_FACTORY_CAPABILITY.getCapabilityServiceName(),
```
ehsavoie: Why not use the constant CLIENT_FACTORY_SERVICE_NAME that you have defined ?

bstansberry: Sure, I'll change that. I added that constant after I wrote this part.

PR #2206, opened 2017-02-26
[WFCORE-2320] Execute internal domain management calls as privileged calls with full RBAC perms

bstansberry:
https://issues.jboss.org/browse/WFCORE-2320

Also reenables the large number of testsuite/domain tests that had been @Ignored with ref to WFCORE-1958. Many of these related to rbac, hence our not knowing RBAC was broken in domain mode. Some others are not related to RBAC, but since they pass I see no reason to have them disabled.

Also tweaks some stuff re: WFCORE-2139 that we would have noticed if the disabled tests had been running.

### DIFF
--- a/host-controller/src/main/java/org/jboss/as/host/controller/parsing/DomainXml_5.java
+++ b/host-controller/src/main/java/org/jboss/as/host/controller/parsing/DomainXml_5.java
@@ -306,6 +306,7 @@ class DomainXml_5 extends CommonXml implements ManagementXmlDelegate {
     }
 
     protected void readDomainElementAttributes(XMLExtendedStreamReader reader, ModelNode address, List<ModelNode> list) throws XMLStreamException {
+        boolean hasDomainOrg = false;
         final int count = reader.getAttributeCount();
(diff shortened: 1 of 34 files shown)


bstansberry:
Failures in Windows run are unrelated.

host-controller/src/main/java/org/jboss/as/host/controller/parsing/DomainXml_5.java
```diff
@@ -336,6 +336,7 @@ protected void readDomainElementAttributes(XMLExtendedStreamReader reader, Model
                         list.add(op);
                         break;
                     }
+                    case  ORGANIZATION: // not in the xsd but let's be forgiving to ease migration
```
luck3y: Not that I think that this will happen a lot, but if a customer did have both organization and domain-organization here, they'd just end up with whatever is defined last.

bstansberry: You've shamed me out of my laziness. ;) A couple lines will address this so I'll add them.

PR #2207, opened 2017-02-27
[WFCORE-2330] JBoss-CLI "deploy -l" always returns exit code 1

msfm:
https://issues.jboss.org/browse/WFCORE-2330

DeploymentHandler.listDeployments() should not always throw CommandFormatException.

### DIFF
--- a/cli/src/main/java/org/jboss/as/cli/handlers/DeploymentHandler.java
+++ b/cli/src/main/java/org/jboss/as/cli/handlers/DeploymentHandler.java
@@ -82,7 +82,7 @@ public abstract class DeploymentHandler extends BatchModeCommandHandler {
                 table.nextRow();
             }
         }
-        throw new CommandFormatException(table.toString());
+        ctx.printLine(table.toString());
     }
 
     protected ModelNode getDeploymentDescriptions(CommandContext ctx, List<String> names) throws CommandFormatException {


cli/src/main/java/org/jboss/as/cli/handlers/DeploymentHandler.java
```diff
@@ -64,25 +64,29 @@ protected void listDeployments(CommandContext ctx, boolean l) throws CommandForm
...
-                    table.addCell(Util.STATUS, result.get(Util.STATUS).asString());
+                if(!table.isAtLastRow()) {
+                    table.nextRow();
                 }
             }
-            if(!table.isAtLastRow()) {
-                table.nextRow();
-            }
+            ctx.printLine(table.toString());
+        } catch (Exception e) {
```
jfdenise: Do you need to catch an exception there?

msfm: I think this is not required in normal scenario. However, I did not have absolute confidence that I can totally remove throwing the CommandFormatException from there. That's why I added such `try ~ catch (Exception e) {...}` block to throw the CommandFormatException for safety's sake.

Is it safe to simply replace `throw new CommandFormatException(table.toString())` with `ctx.printLine(table.toString())` (without adding such `try-catch` block)?

jfdenise: In the existing code, the CommandFormatException is not thrown because of an Exception being thrown. It is thrown at the end of the block.
If a RuntimeException or CommandFormatException is thrown, it will be thrown, no need to re-throw it. So yes, replacing the throw with printLine is safe.

msfm:
Thanks @jfdenise , I've just updated this PR with your review comment.

jfdenise:
Ok, thanks @msfm ! I approve this change.
JF

PR #2208, opened 2017-02-27
WFCORE-2331 - log warning a reconnected slave if reload-required or restart-required triggered

luck3y:


host-controller/src/main/java/org/jboss/as/host/controller/RemoteDomainConnectionService.java
```diff
@@ -597,6 +600,18 @@ private boolean applyRemoteDomainModel(final List<ModelNode> bootOperations, fin
 
             final String outcome = result.get(OUTCOME).asString();
             final boolean success = SUCCESS.equals(outcome);
+
+            // check if anything we synced triggered reload-required or restart-required.
+            // if they did we log a warning on the synced slave.
+            if (result.has(RESPONSE_HEADERS)) {
+                final ModelNode headers = result.get(RESPONSE_HEADERS);
+                if (headers.has(OPERATION_REQUIRES_RELOAD) && headers.get(OPERATION_REQUIRES_RELOAD).asBoolean()) {
```
kabir: It is generally better to use .hasDefined() since then you know there is actually a value. .has() just tells you there is a key there, which might not be defined. It is probably not important to this case though, since this is part of the DC contract

luck3y: @kabir Good catch ;)

PR #2211, opened 2017-02-27
[WFCORE-2334] It is possible to set buffer pool buffer-size attribute…

treblereel:
… to negative value
JIRA: https://issues.jboss.org/browse/WFCORE-2334


bstansberry:
@ctomc please approve

PR #2212, opened 2017-02-27
[WFCORE-2335] It is possible to set buffer pool buffers-per-slice att…

treblereel:
…ribute to negative value
JIRA: https://issues.jboss.org/browse/WFCORE-2335


bstansberry:
@ctomc please approve

PR #2213, opened 2017-02-28
[WFCORE-1507] Expose NotificationHandler registration via a capability

bstansberry:
Finish up on https://issues.jboss.org/browse/WFCORE-1507 by exposing a NotificationHandlerRegistry

Now also includes the https://issues.jboss.org/browse/WFCORE-2340 work previously in #2258

### DIFF
--- a/controller/src/main/java/org/jboss/as/controller/AbstractControllerService.java
+++ b/controller/src/main/java/org/jboss/as/controller/AbstractControllerService.java
@@ -137,6 +137,15 @@ public abstract class AbstractControllerService implements Service<ModelControll
             RuntimeCapability.Builder.of("org.wildfly.management.path-manager", PathManager.class)
                     .build();
 
+    /**
+     * Capability users of the controller use to perform asynchronous management tasks.
+     * This capability isn't necessarily directly related to this class but we declare it
+     * here as it's as good a place as any at this time.
+     */
+    public static final RuntimeCapability<Void> EXECUTOR_CAPABILITY =
+            RuntimeCapability.Builder.of("org.wildfly.management.executor", ExecutorService.class)
+                    .build();
+
     private static final OperationDefinition INIT_CONTROLLER_OP = new SimpleOperationDefinitionBuilder("boottime-controller-initializer-step", null)
         .setPrivateEntry()
         .build();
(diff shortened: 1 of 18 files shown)


bstansberry:
I put the hold label on as this is somewhat an experiment right now. The last commit I coded up last night when the internet was out so I'm not sure it's right. But probably is.

controller/src/main/java/org/jboss/as/controller/registry/NotificationHandlerRegistration.java
```diff
@@ -44,38 +43,7 @@
...
-     * Unregister the given NotificationHandler to stop receiving notifications emitted by the resource at the given source address.
-     *
-     * The source, handler and filter must match the values that were used during registration to be effectively unregistered.
-     *
-     * @param source the path address of the resource that emit notifications.
-     * @param handler the notification handler
-     * @param filter the notification filter
-     */
-    void unregisterNotificationHandler(PathAddress source, NotificationHandler handler, NotificationFilter filter);
+public interface NotificationHandlerRegistration extends NotificationHandlerRegistry {
```
bstansberry: Note: this interface doesn't need to be a subinterface; it could just expose the same methods. The users of NotificationHandlerRegistration are distinct from the users of NotificationHandlerRegistry. I did it this way because the implementation was simpler.

The NotificationHandlerRegistration users are internal to the kernel so we have freedom to change things in the future; e.g. to separate the interfaces.

controller/src/main/java/org/jboss/as/controller/AbstractControllerService.java
```diff
@@ -116,10 +117,16 @@ private static int getBootStackSize() {
...
     }
 
+    /** Capability in-vm users of the controller use to create clients */
     protected static final RuntimeCapability<Void> CLIENT_FACTORY_CAPABILITY =
             RuntimeCapability.Builder.of("org.wildfly.managment.model-controller-client-factory", ModelControllerClientFactory.class)
             .build();
 
+    /** Capability in-vm users of the controller use to register notification handlers */
+    protected static final RuntimeCapability<Void> NOTIFICATION_REGISTRY_CAPABILITY =
+            RuntimeCapability.Builder.of("org.wildfly.managment.notification-handler-registry", NotificationHandlerRegistry.class)
```
jmesnil: This capability would have to be documented in https://github.com/wildfly/wildfly-capabilities.

bstansberry: Yep; thanks. I have a few where I need to do that!

bstansberry:
retest this please

controller/src/main/java/org/jboss/as/controller/AbstractControllerService.java
bstansberry: @jmesnil here's the PR for the ones I've been doing this month:

https://github.com/bstansberry/wildfly-capabilities/pull/3

PR #2214, opened 2017-02-28
[WFCORE-2341] ldap role should ignore PartialResultException when

ppalaga:
referrals=ignore

https://issues.jboss.org/browse/WFCORE-2341

Related issues:
WF https://issues.jboss.org/browse/WFLY-8231
EAP 7.1 https://issues.jboss.org/browse/JBEAP-9188
EAP 7.0 https://issues.jboss.org/browse/JBEAP-9189
EAP 6.4 https://bugzilla.redhat.com/show_bug.cgi?id=1418685

### DIFF
--- a/domain-management/src/main/java/org/jboss/as/domain/management/security/LdapGroupSearcherFactory.java
+++ b/domain-management/src/main/java/org/jboss/as/domain/management/security/LdapGroupSearcherFactory.java
@@ -21,6 +21,7 @@
  */
 package org.jboss.as.domain.management.security;
 
+import static org.jboss.as.domain.management.ModelDescriptionConstants.IGNORE;
 import static org.jboss.as.domain.management.logging.DomainManagementLogger.SECURITY_LOGGER;
 
 import java.io.IOException;
@@ -30,10 +31,13 @@ import java.nio.charset.StandardCharsets;
 import java.util.ArrayList;
 import java.util.Arrays;
 import java.util.HashSet;
+import java.util.Hashtable;
 import java.util.Set;
 
 import javax.naming.NamingEnumeration;
 import javax.naming.NamingException;
+import javax.naming.PartialResultException;
+import javax.naming.Context;
 import javax.naming.NameNotFoundException;
 import javax.naming.directory.Attribute;
 import javax.naming.directory.Attributes;
@@ -336,6 +340,15 @@ public class LdapGroupSearcherFactory {
                         } catch (NameNotFoundException e) {
                             SECURITY_LOGGER.tracef("Failed to query roleNameAttrName: %s", e.getMessage());
                             if (skipMissingGroups == false) {
+                                throw e;
+                            }
+                        } catch (PartialResultException e) {
+                            Hashtable<?, ?> env = null;
+                            if (dirContext != null && (env = dirContext.getEnvironment()) != null && IGNORE.equals(env.get(Context.REFERRAL))) {
+                                /* We can ignore the PartialResultException if we asked for ignoring the referrals
+                                 * - see WFCORE-2341 */
+                                SECURITY_LOGGER.tracef(e, "Ignoring a failed referral query roleNameAttrName");
+                            } else {
                                 throw e;
                             }
                         } catch (LdapReferralException e) {


ppalaga:
@darranl could you please review? 

I admit the fix is a bit hacky in https://github.com/wildfly/wildfly-core/pull/2214/commits/833d3fa03254baf8f439735c047edde937509b1c#diff-0739c4f6354cd5777a9e6fe87b5695d2R343 - I tried to be defensive there, trying to select only the ignorable instances of PartialResultException there as I suppose that it may get thrown also at other circumstances. Unfortunately I found no better way to filter then by checking the exception message. Do you happen to have a better idea?

Although I was able to reproduce the issue with a manually configured ApacheDS and WF, I also tried to write a test in domain management https://github.com/wildfly/wildfly-core/tree/master/domain-management/src/test/java/org/jboss/as/domain/management/security/realms . However, I was not able to come to a state when the test would show the same failure as in the manual test. Here is what I have in a separate branch: https://github.com/ppalaga/wildfly-core/commit/6ed384ae52451c6cdfe56fc49dcd8aef3a60a611#diff-6e7a96f111b881e7aab9b4f871812a61R95 whatever I do in the test, the `PrincipalToGroupSearcher` returns the `ReferralGroupSeven` although it is a referral. Do you happen to have any idea what I am doing wrong?

bstansberry:
Is this an ApacheDS message?

ppalaga:
> Is this an ApacheDS message?

Thanks @bstansberry , you have asked the right question! Yes, the part I check in 833d3fa comes from ApacheDS, and my code is completely useless when it comes to other LDAP servers. Let me find a better solution.

ppalaga:
I checked the OpenJDK sources of com.sun.jndi.ldap. The PartialResultException seems to be thrown only due to failed referral resolution. Based on that, I think that we can ignore all PartialResultExceptions that we get with the LDAP client configured with referrals=ignore - that's what I do in aec97bc660fd43076f218d33d5ef1918a3bd3855

darranl:
I just need to clarify that for this kind of issue I do need the engineer writing the fix to fully understand the problem themselves and to be confident in their fix before it is merged.

The LDAP interactions including backwards compatibility can be very complex and some PRs can end up taking as much effort if not more to review than it would have taken to implement a fix,

darranl:
Can we please keep this PR on hold for now, I am going to have higher priorities over the next couple of weeks before I can really understand the changes.

bstansberry:
@darranl When you get a chance to review it ping me and I'll remove the hold

domain-management/src/main/java/org/jboss/as/domain/management/security/LdapGroupSearcherFactory.java
```diff
@@ -338,6 +342,15 @@ private PrincipalToGroupSearcher(final String groupAttribute, final String group
                             if (skipMissingGroups == false) {
                                 throw e;
                             }
+                        } catch (PartialResultException e) {
+                            Hashtable<?, ?> env = null;
+                            if (dirContext != null && (env = dirContext.getEnvironment()) != null && IGNORE.equals(env.get(Context.REFERRAL))) {
+                                /* We can ignore the PartialResultException if we asked for ignoring the referrals
+                                 * - see WFCORE-2341 */
+                                SECURITY_LOGGER.tracef("Ignoring a failed referral query roleNameAttrName: %s", e.getMessage());
```
darranl: I would suggest here the trace logging should actually log the complete exception, not sure if this is one of those cases but in many cases you can find a message has been logged when the underlying cause had the useful information.

darranl:
@bstansberry I think we can go ahead an remove the HOLD label on this one.

I would recommend in the trace logging we log the exception and not just the message so any cause can be logged as well but after that go ahead and merge.

domain-management/src/main/java/org/jboss/as/domain/management/security/LdapGroupSearcherFactory.java
ppalaga: Accepted in e89669175ecca0e20df0645dc8c5d9d139e47ce0

ppalaga:
e89669175ecca0e20df0645dc8c5d9d139e47ce0 replaces the previous iteration. @darranl 's proposal to log the exception accepted.

bstansberry:
retest this please

This is currently on master-ignore too. The retest here is just in case master-ignore shows problems, since those problems may be unrelated.

PR #2218, opened 2017-02-28
Reenable last remaining test disabled with WFCORE-1958 comment

bstansberry:
Seems to pass now, so I'm enabling it.


bstansberry:
retest this please

PR #2219, opened 2017-03-01
WFCORE-2317: add validation for nested attributes

michpetrov:
Issue: [WFCORE-2317](https://issues.jboss.org/browse/WFCORE-2317)

### DIFF
--- a/controller/src/main/java/org/jboss/as/controller/ValidateModelStepHandler.java
+++ b/controller/src/main/java/org/jboss/as/controller/ValidateModelStepHandler.java
@@ -169,10 +169,16 @@ class ValidateModelStepHandler implements OperationStepHandler {
     }
 
     private void validateNestedAttributes(ModelNode subModel, AttributeDefinition attr, String absoluteParentName) throws OperationFailedException {
+        if (!subModel.isDefined()) {
+            return;
+        }
         AttributeDefinition[] subAttrs = ((ObjectTypeAttributeDefinition) attr).getValueTypes();
         for (AttributeDefinition subAttr : subAttrs) {
             String subAttributeName = subAttr.getName();
             String absoluteName = absoluteParentName + "." + subAttributeName;
+            if (!subModel.hasDefined(subAttributeName) && isRequired(subAttr, subModel)) {
+                throw new OperationFailedException(ControllerLogger.ROOT_LOGGER.required(subAttributeName));
+            }
             if (!subModel.hasDefined(subAttributeName)) {
                 continue;
             }
--- a/controller/src/main/java/org/jboss/as/controller/ValidateModelStepHandler.java
+++ b/controller/src/main/java/org/jboss/as/controller/ValidateModelStepHandler.java
@@ -117,6 +117,9 @@ class ValidateModelStepHandler implements OperationStepHandler {
                 }
                 throw new OperationFailedException(ControllerLogger.ROOT_LOGGER.invalidAttributeCombo(attributeName, sb));
             }
+
+            handleObjectAttributes(model, attr, attributeName);
+
         }
         context.completeStep(OperationContext.RollbackHandler.NOOP_ROLLBACK_HANDLER);
     }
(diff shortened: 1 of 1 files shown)


controller/src/main/java/org/jboss/as/controller/ValidateModelStepHandler.java
```diff
@@ -147,6 +150,62 @@ public void execute(OperationContext context, ModelNode operation) throws Operat
...
+            }
+        }
+    }
+
+    private void validateNestedAttributes(ModelNode subModel, AttributeDefinition attr, String absoluteParentName) throws OperationFailedException {
+        AttributeDefinition[] subAttrs = ((ObjectTypeAttributeDefinition) attr).getValueTypes();
+        for (AttributeDefinition subAttr : subAttrs) {
+            String subAttributeName = subAttr.getName();
+            String absoluteName = absoluteParentName + "." + subAttributeName;
+            if (!subModel.hasDefined(subAttributeName)) {
```
ctomc: this is wrong.
if (!subModel.hasDefined(absoluteParentName, subAttributeName)) {
would work

controller/src/main/java/org/jboss/as/controller/ValidateModelStepHandler.java
```diff
@@ -147,6 +150,62 @@ public void execute(OperationContext context, ModelNode operation) throws Operat
...
+                                sb.append(", ");
+                            }
+                            sb.append(absoluteParentName + "." + alt);
+                        }
+                    }
+                }
+                throw new OperationFailedException(ControllerLogger.ROOT_LOGGER.invalidAttributeCombo(absoluteName, sb));
+            }
+
+            handleObjectAttributes(subModel, subAttr, absoluteName);
```
ctomc: absoluteName is wrong at this point. it cannot be string with "." in it.

controller/src/main/java/org/jboss/as/controller/ValidateModelStepHandler.java
```diff
@@ -147,6 +150,62 @@ public void execute(OperationContext context, ModelNode operation) throws Operat
...
+        } else if (attr instanceof ObjectMapAttributeDefinition) {
+            AttributeDefinition valueType = ((ObjectMapAttributeDefinition) attr).getValueType();
+            ModelNode map = model.get(attr.getName());
+            for (String key : map.keys()) {
+                validateNestedAttributes(map.get(key), valueType, absoluteParentName + "." + key);
+            }
+        }
+    }
+
+    private void validateNestedAttributes(ModelNode subModel, AttributeDefinition attr, String absoluteParentName) throws OperationFailedException {
```
ctomc: You could have here two params for name.
String absoluteParentName,  //readable string form, only for displaying in validation message
String[] fullName // actual name segments, that can be used in resolving and lookup

controller/src/main/java/org/jboss/as/controller/ValidateModelStepHandler.java
```diff
@@ -147,6 +150,62 @@ public void execute(OperationContext context, ModelNode operation) throws Operat
...
+            validateNestedAttributes(model.get(attr.getName()), attr, absoluteParentName);
+        } else if (attr instanceof ObjectListAttributeDefinition) {
+            AttributeDefinition valueType = ((ObjectListAttributeDefinition) attr).getValueType();
+            ModelNode list = model.get(attr.getName());
+            for (int i = 0; i < list.asInt(); i++) {
+                validateNestedAttributes(list.get(i), valueType, absoluteParentName + "[" + i + "]");
+            }
+        } else if (attr instanceof ObjectMapAttributeDefinition) {
+            AttributeDefinition valueType = ((ObjectMapAttributeDefinition) attr).getValueType();
+            ModelNode map = model.get(attr.getName());
```
ctomc: why not use 
Map<String, ModelNode>map = ((ObjectMapAttributeDefinition) attr).unwrap(...

controller/src/main/java/org/jboss/as/controller/ValidateModelStepHandler.java
michpetrov: That seems more trouble than it's worth, given that lists would require special handling.

controller/src/main/java/org/jboss/as/controller/ValidateModelStepHandler.java
michpetrov: The subModel is (more or less) equal to model.get(parentName), it doesn't contain absoluteParentName.

controller/src/main/java/org/jboss/as/controller/ValidateModelStepHandler.java
michpetrov: I can change it but it seems insufficient, we're nested in a nested object at this point, I'd expect showing `object.nestedObject.list[0].attr` is clearer than just `list[0].attr`.

controller/src/main/java/org/jboss/as/controller/ValidateModelStepHandler.java
michpetrov: I'm not seeing any unwrap method for ObjectMapAttributeDefinition, also I just need to get to the nested ModelNodes, I don't think it matters what I retrieve them from.

michpetrov:
@ctomc Any further comments? I'd like to see this resolved.

controller/src/main/java/org/jboss/as/controller/ValidateModelStepHandler.java
```diff
@@ -147,6 +150,62 @@ public void execute(OperationContext context, ModelNode operation) throws Operat
...
+        }
+    }
+
+    private void validateNestedAttributes(ModelNode subModel, AttributeDefinition attr, String absoluteParentName) throws OperationFailedException {
+        AttributeDefinition[] subAttrs = ((ObjectTypeAttributeDefinition) attr).getValueTypes();
+        for (AttributeDefinition subAttr : subAttrs) {
+            String subAttributeName = subAttr.getName();
+            String absoluteName = absoluteParentName + "." + subAttributeName;
+            if (!subModel.hasDefined(subAttributeName)) {
+                continue;
```
bstansberry: A check of isRequired(subAttr, subModel) is needed before continuing, throwing OFE if it fails.

controller/src/main/java/org/jboss/as/controller/ValidateModelStepHandler.java
```diff
@@ -147,6 +150,62 @@ public void execute(OperationContext context, ModelNode operation) throws Operat
...
     }
 
+    private void handleObjectAttributes(ModelNode model, AttributeDefinition attr, String absoluteParentName) throws OperationFailedException {
+        if (attr instanceof ObjectTypeAttributeDefinition) {
+            validateNestedAttributes(model.get(attr.getName()), attr, absoluteParentName);
+        } else if (attr instanceof ObjectListAttributeDefinition) {
+            AttributeDefinition valueType = ((ObjectListAttributeDefinition) attr).getValueType();
+            ModelNode list = model.get(attr.getName());
+            for (int i = 0; i < list.asInt(); i++) {
+                validateNestedAttributes(list.get(i), valueType, absoluteParentName + "[" + i + "]");
```
bstansberry: It's possible that list.get(i) is undefined, in which case the validateNestedAttributes call should not be made.

This is unlikely but might as well be robust.

controller/src/main/java/org/jboss/as/controller/ValidateModelStepHandler.java
```diff
@@ -147,6 +150,62 @@ public void execute(OperationContext context, ModelNode operation) throws Operat
...
+            }
+        } else if (attr instanceof ObjectMapAttributeDefinition) {
+            AttributeDefinition valueType = ((ObjectMapAttributeDefinition) attr).getValueType();
+            ModelNode map = model.get(attr.getName());
+            for (String key : map.keys()) {
+                validateNestedAttributes(map.get(key), valueType, absoluteParentName + "." + key);
```
bstansberry: It's possible that map.get(key) is undefined, in which case validateNestedAttributes should not be called.

Again unlikely, but might as well be robust. (Note that in both the list case and this map case if the undefined element was illegal it would have been rejected before this OSH ever gets called.)

controller/src/main/java/org/jboss/as/controller/ValidateModelStepHandler.java
bstansberry: The use of absoluteName seems correct to me. As this recurses it's building a '.' and '[n]' based path to potentially incorrect fields. The string is only used in the OFE messages where it seems to fit ok.

controller/src/main/java/org/jboss/as/controller/ValidateModelStepHandler.java
michpetrov: Couldn't it then be undefined in the simple object as well? I added the check at the start of validateNestedAttributes, if it's ok now I will rebase the PR

bstansberry: The simple case is handled at L84:
```
            if (!has) {
                continue;
            }
```

bstansberry:
@michpetrov Looks good; please squash. Please @ ping me here or even better in HipChat when it's updated and I'll merge it.

bstansberry:
this is ok to test

bstansberry:
It seems CI never tested this! Once it does it's going to fail, not because it's wrong but because it's exposing bugs. This answers a minor mystery that came up as we were thinking about WFCORE-2674 -- why what you did here wasn't resulting in a failure given how content attributes in deployment resources are configured. :)

bstansberry:
Please don't worry about the test failure once it happens; please just squash. My WFCORE-2674 branch is based on your branch here; I'll update it once you squash, complete 2674 and fix the deployment resource failure, and send up a PR with your commits and mine to clear out the whole set.

bstansberry:
I squashed these and included them in #2339. Thanks a lot; this is a big improvement, and sorry this took so long.

PR #2220, opened 2017-03-01
Fix for WFCORE-2347. Intermittent failure of some CLI tests.

jfdenise:
The test that is failing in the bug report is FileArgumentTestCase.java but FileWithPropertiesTestCase.java could be affected by the same problem. Both tests have been refactored similarly.
I suspect that using commonly used CliProcessWrapper to interact with the CLI would make these intermitent failures to vanish. This would have to be validated over time in QA context.
This fix is an attempt.

### DIFF
--- a/testsuite/standalone/src/test/java/org/jboss/as/test/integration/management/cli/FileArgumentTestCase.java
+++ b/testsuite/standalone/src/test/java/org/jboss/as/test/integration/management/cli/FileArgumentTestCase.java
@@ -27,22 +27,14 @@ import static org.junit.Assert.fail;
 
 import java.io.File;
 import java.io.IOException;
-import java.io.InputStream;
 import java.io.Writer;
 import java.nio.charset.StandardCharsets;
 import java.nio.file.Files;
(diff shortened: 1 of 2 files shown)


bstansberry:
I approved this but have a question:

Should CliCommandBuilder be fixed in some way? Not as an alternative solution for this JIRA or as a priority; just should we file a JIRA to fix something? IOW is CliCommandBuilder launching the CLI incorrectly, or is this just a matter of using CliProcessWrapper as it makes it possible to better handle other things besides launch.

jfdenise:
CliCommandBuilder is a fine class. In this case we had tests that were reimplementing the logic present in CLIProcessWrapper (CLIProcessWrapper uses CliCommandBuilder).
 
I noticed the class CustomCLIExecutor that does something similar to CLIProcessWrapper but add some logic to it (in the context of triaging JBEAP-4013). We could end by aligning all tests to use CLIProcessWrapper, but I am pretty we would have to make changes to CLIProcessWrapper to handle fully what tests are doing with CustomCLIExecutor.

PR #2221, opened 2017-03-02
WFCORE-2348 Fix issue with old initscripts when starting WF as a service

rstancel:
https://issues.jboss.org/browse/WFCORE-2348


bstansberry:
this is ok to test

PR #2226, opened 2017-03-06
[WFCORE-2337]: Removal of overlay in a server group fails inside a composite.

ehsavoie:
Fixing the way the resource is removed.

Jira: https://issues.jboss.org/browse/WFCORE-2337

### DIFF
--- a/host-controller/src/main/java/org/jboss/as/domain/controller/operations/DomainDeploymentOverlayRedeployLinksHandler.java
+++ b/host-controller/src/main/java/org/jboss/as/domain/controller/operations/DomainDeploymentOverlayRedeployLinksHandler.java
@@ -21,6 +21,7 @@ import org.jboss.as.controller.OperationContext;
 import org.jboss.as.controller.OperationDefinition;
 import org.jboss.as.controller.OperationFailedException;
 import org.jboss.as.controller.OperationStepHandler;
+import org.jboss.as.controller.PathAddress;
 import org.jboss.as.controller.SimpleOperationDefinitionBuilder;
 import org.jboss.as.controller.StringListAttributeDefinition;
 import org.jboss.as.controller.client.helpers.Operations;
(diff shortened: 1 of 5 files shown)


jfdenise:
@ehsavoie with respect to CLI, all new tests are passing. Thanks!

controller/src/main/java/org/jboss/as/controller/CompositeOperationHandler.java
```diff
@@ -113,6 +113,7 @@ public void handleRollback(OperationContext context, ModelNode operation) {
                     ModelNode stepResponse = responseMap.get(stepName);
                     if (stepResponse.hasDefined(FAILURE_DESCRIPTION)) {
                         failureMsg.get(ControllerLogger.ROOT_LOGGER.compositeOperationFailed(), ControllerLogger.ROOT_LOGGER.operation(stepName)).set(stepResponse.get(FAILURE_DESCRIPTION));
+                        failureMsg.get(ControllerLogger.ROOT_LOGGER.compositeOperationFailed(), "Operation: " + ControllerLogger.ROOT_LOGGER.operation(stepName)).set(operationMap.get(stepName));
```
bstansberry: Is this change necessary to resolve WFCORE-2337? If not I'd prefer to deal with it separately. We'd need i18n of "Operation:", and I'm not sure whether the output here is right.

bstansberry: Not right; this drops the failure details from the top level failure description and replaces them with the full details of the step operation, including params, which I don't really want to output. The format of the address (dmr syntax) isn't great either. If we output operation details it needs to be in addition to the failure info we already report.

ehsavoie: Yes, sorry this was a leftover when tracking which remove op was failing

PR #2232, opened 2017-03-08
WFCORE-1307 rollback on failure to persist changes to config files

luck3y:
This turned out a little on the hacky side, but unless we prepare() the persistence when we're still executing Stage.Model, the changes will get applied & committed on domain servers, but rolled back on the hostcontroller(s).

This can probably be cleaned up a little, for example the OSH for the prepare could be moved somewhere else, but I couldn't find anything doing much similar to this.

### DIFF
--- a/controller/src/main/java/org/jboss/as/controller/AbstractOperationContext.java
+++ b/controller/src/main/java/org/jboss/as/controller/AbstractOperationContext.java
@@ -124,6 +124,7 @@ abstract class AbstractOperationContext implements OperationContext {
     static final ThreadLocal<Thread> controllingThread = new ThreadLocal<Thread>();
 
     static final String INTERNAL_MODEL_VALIDATION_NAME = "internal-model-validation";
+    static final String INTERNAL_MODEL_PREPARE_PERSISTENCE = "internal-model-prepare-persistence";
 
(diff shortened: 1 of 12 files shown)


luck3y:
retest this please

luck3y:
retest this please

luck3y:
retest this please

bstansberry:
I put a 'hold' label on this as Ken's investigating why he's seeing the server's persist before the DC does, thus preventing clean rollback on the servers if there is a persistence problem on the DC.

luck3y:
I've reworked a version of this that doesn't use the OSH for the persistence prepare, the issue with commit on servers is that the default transactionControl used (ModelControllerClientOperationHandler:202) always commits when its called, and also that DomainRolloutHandler.finalizeOp doesn't check the current operationcontext for isRollbackOnly().

I think we still need to do the OSH though, as more testing with --backup / --cached-dc cases shows, if prepare fails on the master but works on the slave, by the time executeDoneStage() is running on the DC, the slaves may have already persisted the changes to their own cached domain.xml.

So I don't see a way to avoid preparing peristence over the whole domain in the OSH then calling commit in executeDoneState().

I'll rework this PR to incorporate these changes.

bstansberry:
I'm cleaning out the queue a bit. Please open this again when you're ready.

PR #2236, opened 2017-03-09
[WFCORE-2511]: overlay, redeploy-links should ignore unknown deployments.

ehsavoie:
It was first thought that if the deployment didn't exist we should fail. It appears we were wrong so now we just ignore those deployments.

Jira: https://issues.jboss.org/browse/WFCORE-2511


bstansberry:
@jfdenise Does this work for you? You're my QE guy here. ;)

jfdenise:
@bstansberry , yes tests are passing.

PR #2237, opened 2017-03-11
[WFCORE-1337] Expose paths and the PathManager as capabilities

bstansberry:
https://issues.jboss.org/browse/WFCORE-1337


rhusar:
Does this require PR to https://github.com/wildfly/wildfly-capabilities as well?

PR #2242, opened 2017-03-14
Add mixed domain host excludes for WF 10.0 and 10.1

bstansberry:


bstansberry:
retest this please

bstansberry:
I removed the 'hold' label as I see no negative effects in the testsuite.

ctomc:
Why is this needed? mixed-domain tests only check EAP releases not intermittent WildFly ones....

luck3y:
This will be needed for just using core & old slaves, when we figure out how to add elytron to the core feature pack.

PR #2244, opened 2017-03-14
Change shebang to work with debian-based distributions

jairojunior:
In RedHat-based distributions sh is a symbolic link to `bash`, but in Debian-based distributions it points to `dash`.

I have been maintaining a Puppet Module [1] to manage Wildfly and this change was enough to make it work with debian-based distributions with systemd.

[1] https://github.com/biemond/biemond-wildfly/


bstansberry:
this is ok to test

bstansberry:
Thanks!

PR #2247, opened 2017-03-14
Fix for WFCORE-1534. Generic command name issue

jfdenise:
Fix generic command generated help to expose the command name.


jfdenise:
retest this please!

PR #2251, opened 2017-03-15
[WFCORE-316] adding testcase to check functionality of --cached-dc parameter

ochaloup:
Adding testcase for EAP 7.1 RFE EAP7-496

https://issues.jboss.org/browse/EAP7-496
https://issues.jboss.org/browse/WFCORE-316
https://issues.jboss.org/browse/JBEAP-8605

### DIFF
--- a/testsuite/domain/src/test/java/org/jboss/as/test/integration/domain/cacheddc/CachedDcDomainTestCase.java
+++ b/testsuite/domain/src/test/java/org/jboss/as/test/integration/domain/cacheddc/CachedDcDomainTestCase.java
@@ -1,0 +1,514 @@
+/*
+* JBoss, Home of Professional Open Source
+* Copyright 2017, Red Hat, Inc., and individual contributors as indicated
+* by the @authors tag.
+*
+* Licensed under the Apache License, Version 2.0 (the "License");
+* you may not use this file except in compliance with the License.
+* You may obtain a copy of the License at
+*
+* http://www.apache.org/licenses/LICENSE-2.0
+*
+* Unless required by applicable law or agreed to in writing, software
+* distributed under the License is distributed on an "AS IS" BASIS,
+* WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+* See the License for the specific language governing permissions and
+* limitations under the License.
+*/
+package org.jboss.as.test.integration.domain.cacheddc;
+
+import static org.jboss.as.controller.client.helpers.ClientConstants.OUTCOME;
+import static org.jboss.as.controller.client.helpers.ClientConstants.SUCCESS;
+import static org.jboss.as.controller.descriptions.ModelDescriptionConstants.HOST;
+import static org.jboss.as.controller.descriptions.ModelDescriptionConstants.LOGGER;
+import static org.jboss.as.controller.descriptions.ModelDescriptionConstants.PROFILE;
+import static org.jboss.as.controller.descriptions.ModelDescriptionConstants.SERVER;
+import static org.jboss.as.controller.descriptions.ModelDescriptionConstants.SUBSYSTEM;
+import java.io.File;
+import java.io.IOException;
+import java.net.InetAddress;
+import java.net.UnknownHostException;
+import java.nio.charset.Charset;
+import java.util.concurrent.TimeUnit;
+import java.util.function.Supplier;
+import org.jboss.as.controller.client.helpers.Operations;
+import org.jboss.as.controller.client.helpers.domain.DomainClient;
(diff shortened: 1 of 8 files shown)


testsuite/domain/src/test/java/org/jboss/as/test/integration/domain/cacheddc/CachedDcDomainTestCase.java
```diff
@@ -0,0 +1,525 @@
...
+ * This software is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the GNU
+ * Lesser General Public License for more details.
+ *
+ * You should have received a copy of the GNU Lesser General Public
+ * License along with this software; if not, write to the Free
+ * Software Foundation, Inc., 51 Franklin St, Fifth Floor, Boston, MA
+ * 02110-1301 USA, or see the FSF site: http://www.fsf.org.
+ */
```
bstansberry: This needs to be the ASL copyright header.

testsuite/domain/src/test/java/org/jboss/as/test/integration/domain/cacheddc/CachedDcDomainTestCase.java
```diff
@@ -0,0 +1,525 @@
...
+import org.junit.Test;
+
+/**
+ * Testing <code>--cached-dc</code> configuration option.<br>
+ * Tests shows that <code>domain.cached-remote.xml</code> is created on slave HC
+ * and that slave HC is capable to start when that file exists even DC is down.
+ *
+ * @author <a href="ochaloup@redhat.com">Ondra Chaloupka</a>
+ */
+public class CachedDcDomainTestCase {
```
bstansberry: How long does this test take to run? I fear it's a really long time as it looks like it starts a domain 9 times, and I could be wrong but I think in one test it waits 30 secs for an-expected-to-not-start HC to definitely not start.

This needs to be optimized to run as fast as possible. The master shouldn't need a host.xml that has servers; the slave shouldn't need more than enough servers to do the validation of server state (ideally just one.)

If the master can be left running and only restarted when necessary that will help too. I'm not sure how well DomainTestSupport supports that, given the slave config changes from test to test, but if it's doable it will be good.

bstansberry:
@luck3y Can you have a look at the new test?

bstansberry:
BTW, thanks for the fluent API idea! And for the tests.

testsuite/domain/src/test/java/org/jboss/as/test/integration/domain/cacheddc/CachedDcDomainTestCase.java
ochaloup: Yes, that's right. Thanks for your points. I will try to tune it base on your recommendation.

To particular points
Yeap, it took some time. From testing on my computer I can say the new CahedDcTestCase takes twice more time than AutoIgnoredResourcesDomainTestCase which I was playing with. The second one about 100 sec and the new one about 200 sec.

True, my wrong about the 30 sec for an-expected-to-not-start. I was doing some refactorings just before the pushing the PR and I didn't realized that. That one is the easiest to be fixed. I need to take a look on the handling of starting just master/server and not all that are configured in `DomainTestSupport`. Maybe I would create a new TestSupport class. I need to check it in more details.

ochaloup: @bstansberry I made it as simple as possible when not changing whole the concept of the test. I based it on changing host*xml configuration to start only one server each time. I shrink the running time to half that means slightly more than 100 sec on my laptop. Would that be sufficient for you?

If it's not ok I'm thinking about removing some testcases. I think that all presented scenarios have value to be tested but...
...I think I'm not capable to rewrite testcases in reasonable amount of time to run faster.

bstansberry: It's 135 secs on my laptop. That's not great but I don't mind taking time to cover things, as long as we're making good effort to be efficient, which you've done. So it's ok, pending any feeback from @luck3y re the details of the tests.

The pom change I mentioned still needs to happen though.

ochaloup: ouch, I overlooked the point about `pom.xml`. fixed now, I hope.

and thank you!

PR #2255, opened 2017-03-15
[WFCORE-2536]: JmxSensitiveTestCase fails with security manager in WF core.

ehsavoie:
Fixes also WFCORE-2537: ModelControllerMBeanTestCase fails with security manager in WF core

Jira: https://issues.jboss.org/browse/WFCORE-2536
Jira: https://issues.jboss.org/browse/WFCORE-2537

### DIFF
--- a/testsuite/shared/src/main/java/org/wildfly/test/jmx/ServiceActivatorDeploymentUtil.java
+++ b/testsuite/shared/src/main/java/org/wildfly/test/jmx/ServiceActivatorDeploymentUtil.java
@@ -38,6 +38,7 @@ import org.wildfly.test.jmx.staticmodule.JMXControlledStateNotificationListenerE
 import org.wildfly.test.jmx.staticmodule.JMXNotificationsService;
 
 import javax.management.MBeanPermission;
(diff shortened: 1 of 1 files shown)


testsuite/shared/src/main/java/org/wildfly/test/jmx/ServiceActivatorDeploymentUtil.java
```diff
@@ -60,6 +61,11 @@ public static void createServiceActivatorDeployment(File destination, String obj
         sb.append('=');
         sb.append(objectName);
         sb.append("\n");
+        archive.addAsManifestResource(PermissionUtils.createPermissionsXmlAsset(
+                new MBeanPermission("org.wildfly.test.jmx.Dynamic#-[" + objectName + "]", "registerMBean"),
+                new MBeanPermission("org.wildfly.test.jmx.Dynamic#-[" + objectName + "]", "unregisterMBean"),
```
bstansberry: I think these should use mbeanClass.getName() instead of hard coding Dynamic.

PR #2258, opened 2017-03-15
[WFCORE-2340] Expose the management executor as a capability

bstansberry:
https://issues.jboss.org/browse/WFCORE-2340

Besides capabilities being a better way to access services, this makes it easier to write subsystems usable on both the HC and a server by providing a unified way to find the executor service.

#2213 up to now has included this but I'm separating it to simplify reviews.

Hint hint. ;)


jamezp:
The failed test seems to be problematic on Windows. I don't think this PR had anything to do with it.

bstansberry:
I'm going to pull this commit over into the #2213 PR, as I expect it would have merge conflicts. Both have been reviewed/approved so now it's easier to deal with them together.

PR #2276, opened 2017-03-27
[WFCORE-2532] Add permissions for logging tests when running with a s…

jamezp:
…ecurity manager. Added workaround for LOGMGR-149.

https://issues.jboss.org/browse/WFCORE-2532

There is a workaround for https://issues.jboss.org/browse/LOGMGR-149 which is a legitimate failure. I can do a new release of the log manager if we feel it's important. Regardless I will get it fixed in the log manager.

### DIFF
--- a/testsuite/manualmode/src/test/java/org/jboss/as/test/manualmode/logging/AbstractLoggingTestCase.java
+++ b/testsuite/manualmode/src/test/java/org/jboss/as/test/manualmode/logging/AbstractLoggingTestCase.java
@@ -33,6 +33,7 @@ import java.nio.charset.StandardCharsets;
 import java.nio.file.Files;
 import java.nio.file.Path;
 import java.nio.file.Paths;
+import java.security.Permission;
 import java.util.Collections;
 import java.util.Map;
 import javax.inject.Inject;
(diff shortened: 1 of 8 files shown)


jamezp:
This [run](https://ci.wildfly.org/viewLog.html?buildId=52491&tab=buildResultsDiv&buildTypeId=WildFlyCore_MasterIgnoreLinux) has the security manager enabled. It will have other failures, but the logging tests should be passing.

testsuite/manualmode/src/test/java/org/jboss/as/test/manualmode/logging/AbstractLoggingTestCase.java
```diff
@@ -114,7 +126,18 @@ static JavaArchive createDeployment(final Class<? extends ServiceActivator> serv
...
+
+    public static JavaArchive addPermissions(final JavaArchive archive, final Permission... additionalPermissions) {
+        final Permission[] permissions;
+        if (additionalPermissions == null || additionalPermissions.length == 0) {
+            permissions = DEFAULT_PERMISSIONS;
+        } else {
+            permissions = Arrays.copyOf(DEFAULT_PERMISSIONS, DEFAULT_PERMISSIONS.length + additionalPermissions.length);
+            System.arraycopy(additionalPermissions, 0, permissions, DEFAULT_PERMISSIONS.length, additionalPermissions.length);
+        }
+        return archive.addAsManifestResource(PermissionUtils.createPermissionsXmlAsset(permissions), "permissions.xml");
```
bstansberry: The DEFAULT_PERMISSIONS are all for what UndertowServiceActivator does, right? If so, how about moving this to UndertowServiceActivator? ModuleResourceRootPathsTestCase can then call it too, plus another tests we add that use the UndertowServiceActivator facility.

jamezp: Yeah that's a good point. I'll have a look at that.

jamezp:
I've updated this to include a fix for #2283. I can split this into 3 commits if it's better to have the default permissions part separate or squash these because the second won't really work without the first. I just wasn't sure what the best option would be so I left it as two for now.

jamezp:
Also the [new test run](https://ci.wildfly.org/viewLog.html?buildId=52645&) with the security manager enabled.

PR #2277, opened 2017-03-28
WFCORE-2562 don't fail boot if we can't remove old configuration history

luck3y:


luck3y:
FYI, updated this to never throw, since its only used to remove old configuration backups and shouldn't prevent the server from starting.

PR #2279, opened 2017-03-28
WFCORE-2586 fix intermittent failures of JmxControlledStateNotificati…

jmartisk:
…ons*TestCase

https://issues.jboss.org/browse/WFCORE-2586

### DIFF
--- a/testsuite/shared/src/main/java/org/wildfly/test/jmx/ControlledStateNotificationListener.java
+++ b/testsuite/shared/src/main/java/org/wildfly/test/jmx/ControlledStateNotificationListener.java
@@ -17,6 +17,7 @@ package org.wildfly.test.jmx;
 
 import java.io.BufferedWriter;
 import java.io.IOException;
+import java.nio.channels.ClosedByInterruptException;
 import java.nio.charset.StandardCharsets;
 import java.nio.file.Files;
 import java.nio.file.Path;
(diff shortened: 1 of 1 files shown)


testsuite/shared/src/main/java/org/wildfly/test/jmx/ControlledStateNotificationListener.java
```diff
@@ -82,6 +83,20 @@ private void writeNotification(AttributeChangeNotification notification, Path pa
             in.write(String.format("%s %s %s %s", notification.getType(), notification.getSequenceNumber(), notification.getSource().toString(), notification.getMessage()));
             in.newLine();
             in.flush();
+        } catch (ClosedByInterruptException cbie) {
+            /* This happens sometimes during the "reload" operation, the BufferedWriter won't write anymore
+               because our thread was interrupted by the server.
+               So we clear the interruption flag and try writing again and hope that it will work this time.
+            */
+            Thread.interrupted();
```
bstansberry: If this returns true you should interrupt the thread again in a finally block. The server wanted the thread interrupted so that should be respected.

jmartisk: Fixed, thanks for pointing that out.

jmartisk:
retest this please

bstansberry:
Thanks!

PR #2281, opened 2017-03-28
[WFCORE-2588] Deprecating optional dependencies in DDs and manifests

ropalka:
…nifests


bstansberry:
This one better have a JBEAP and acks. I know that customers use this so removing it and deciding to log warnings about that requires higher visibility than just a PR review.

ropalka:
I created JBEAP-9962. Do I need to reference it in git commit message?

dmlloyd:
I don't think we ever all agreed to deprecate optional deployment module dependencies.  I still believe it is possible to change the mechanism of the dependency to simply query for the module at the correct point in time, and if it's missing, skip the dependency.

ropalka:
Ah right! Now I remember saying you @dmlloyd this should be worked around in DU chain somehow.
Thanks for reminding it. I'm closing this PR and associated JIRAs.

bstansberry:
@ropalka You don't need to reference JBEAP JIRAs in WildFly Core commit messages (including wildfly-core-eap). Nor in upstream WildFly. It's nice to do (but AFAIK not required) in the EAP 7.x branch as there the JBEAP is the true tracker for the change to that code base.

PR #2282, opened 2017-03-28
[WFCORE-2589][WFCORE-2590] Deprecating org.jboss.as.server.Services

ropalka:
... method


server/src/main/java/org/jboss/as/server/Services.java
```diff
@@ -73,8 +73,14 @@ private Services() {
 
     public static final ServiceName JBOSS_PRODUCT_CONFIG_SERVICE = JBOSS_AS.append("product-config");
 
+    @Deprecated
```
bstansberry: Please add javadoc with @deprecated stating that this will be removed in the next WildFly Core major or minor release.

PR #2283, opened 2017-03-28
[WFCORE-2535]: ModuleResourceRootPathsTestCase fails with security manager in WF core.

ehsavoie:
Adding required permissions.

Jira: https://issues.jboss.org/browse/WFCORE-2535
JBEAP: https://issues.jboss.org/browse/JBEAP-9553


bstansberry:
Please see my comment on #2276. If that PR or this one does that, then the other can use it.

jamezp:
I'm closing this only because I added the fix to #2276. I didn't seem to need the `MBeanPermission` so that has been left off.

PR #2286, opened 2017-03-28
[WFCORE-2536 & WFCORE-2537] Ensure the correct name is used for the p…

jamezp:
…ermissions.

https://issues.jboss.org/browse/WFCORE-2536
https://issues.jboss.org/browse/WFCORE-2537

Follow up for #2255


ehsavoie:
Maybe we should use the correct constructor of MBeanPermission instead ?

bstansberry:
@ehsavoie I didn't see your comment and merged this as it passed CI with a bunch of other ones overnight. Please send a PR if you want something different. Thanks!

PR #2289, opened 2017-03-30
WFCORE-2600 set connection timeout to 20 secs when creating new comma…

istudens:
…nd context

https://issues.jboss.org/browse/WFCORE-2600

### DIFF
--- a/testsuite/shared/src/main/java/org/jboss/as/test/integration/management/util/CLITestUtil.java
+++ b/testsuite/shared/src/main/java/org/jboss/as/test/integration/management/util/CLITestUtil.java
@@ -32,6 +32,7 @@ import java.net.URISyntaxException;
 import org.jboss.as.cli.CliInitializationException;
 import org.jboss.as.cli.CommandContext;
 import org.jboss.as.cli.CommandContextFactory;
+import org.jboss.as.cli.impl.CommandContextConfiguration;
 import org.jboss.as.test.integration.domain.management.util.DomainTestSupport;
 import org.jboss.as.test.integration.domain.management.util.WildFlyManagedConfiguration;
 import org.jboss.as.test.shared.TestSuiteEnvironment;
@@ -65,11 +66,11 @@ public class CLITestUtil {
     public static CommandContext getCommandContext(WildFlyManagedConfiguration config) throws CliInitializationException {
         System.setProperty("aesh.terminal","org.jboss.aesh.terminal.TestTerminal");
         setJBossCliConfig();
-            return CommandContextFactory.getInstance().newCommandContext(
-                    constructUri(config.getHostControllerManagementProtocol(),
-                            config.getHostControllerManagementAddress(),
-                            config.getHostControllerManagementPort()),  isRemote ? username : null, isRemote ? password.toCharArray() : null);
-        }
+        return CommandContextFactory.getInstance().newCommandContext(
+                constructUri(config.getHostControllerManagementProtocol(),
+                        config.getHostControllerManagementAddress(),
+                        config.getHostControllerManagementPort()),  isRemote ? username : null, isRemote ? password.toCharArray() : null);
+    }
 
     public static CommandContext getCommandContext(String address, int port, InputStream in, OutputStream out)
             throws CliInitializationException {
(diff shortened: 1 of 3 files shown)


istudens:

testsuite/shared/src/main/java/org/jboss/as/test/integration/management/util/CLITestUtil.java
```diff
@@ -76,7 +78,16 @@ public static CommandContext getCommandContext(String address, int port, InputSt
...
-        return CommandContextFactory.getInstance().newCommandContext(address + ":" + port, isRemote ? username : null, isRemote ? password.toCharArray() : null, in, out);
+        return CommandContextFactory.getInstance().newCommandContext(new CommandContextConfiguration.Builder()
+                .setController(address + ":" + port)
+                .setUsername(isRemote ? username : null)
+                .setPassword(isRemote ? password.toCharArray() : null)
+                .setConsoleInput(in)
+                .setConsoleOutput(out)
+                .setDisableLocalAuth(false)
+                .setInitConsole(false)
+                .setConnectionTimeout(TimeoutUtil.adjust(20 * 1000))
```
jfdenise: I am wandering it it wouldn't be better to add a new getCommandContext operation with a timeout parameter instead of hardcoding 20secs for all. You would then call is from your test.

istudens: There should not be any hardcoded timeouts in the testsuite. Every timeout should be adjusted by TimeoutUtil. That's also the reason why JBEAP-8244 is failing. You are, however, right that I don't need to increase the connectionTimeout for other tests, I will rework it a bit.

istudens:
I've reworked this so that the only testcase touched by this change will be ```ObjectStoreTypeTestCase```. 
Just a note that there are too many ```getCommandContext``` methods in ```CLITestUtil```, it would be better to have a Builder for it.

testsuite/shared/src/main/java/org/jboss/as/test/integration/management/util/CLIWrapper.java
```diff
@@ -93,12 +94,26 @@ public CLIWrapper(boolean connect, String cliAddress) throws CliInitializationEx
      * @param consoleInput input stream to use for sending to the CLI, or {@code null} if the standard input stream should be used
      */
     public CLIWrapper(boolean connect, String cliAddress, InputStream consoleInput) throws CliInitializationException {
+        this(connect, cliAddress, consoleInput, CONNECTION_TIMEOUT);
```
jfdenise: Do you want to supersede the CLI defaults? You could not provide any connectionTimeout if a negative connectionTimeout is provided.

istudens: Thanks for pointing that out. I will change it to ```-1```.

testsuite/shared/src/main/java/org/jboss/as/test/integration/management/util/CLITestUtil.java
```diff
@@ -79,6 +81,23 @@ public static CommandContext getCommandContext(String address, int port, InputSt
...
+        setJBossCliConfig();
+        return CommandContextFactory.getInstance().newCommandContext(new CommandContextConfiguration.Builder()
+                .setController(address + ":" + port)
+                .setUsername(isRemote ? username : null)
+                .setPassword(isRemote ? password.toCharArray() : null)
+                .setConsoleInput(in)
+                .setConsoleOutput(out)
+                .setDisableLocalAuth(false)
+                .setInitConsole(false)
+                .setConnectionTimeout(TimeoutUtil.adjust(connectionTimeout))
```
jfdenise: TimeoutUtil.adjust will throw an exception is the passed value is negative, no?

istudens: omg, I didn't concentrate on this, hopefully it is done now, thanks a lot

istudens:
retest this please

jfdenise:
Hi @istudens , thank-you for the changes. Are you going to update the test with a 20seconds timeout? This must be passed as ms so 20000.
JF

istudens:
PR for the test can be found here https://github.com/wildfly/wildfly/pull/9878

jfdenise:
Thank-you. I approve. Thanks!
JF

PR #2301, opened 2017-03-31
WFCORE-2385: remove extraneous methods

michpetrov:
Issue: [WFCORE-2385](https://issues.jboss.org/browse/WFCORE-2385)

----

As far as I understand these methods were put in place to assure the alias property is case-insensitive, this is now wholly taken care of by CredentialStoreResource and as such these methods are unnecessary. As a result I think [this line](https://github.com/wildfly/wildfly-core/blob/master/elytron/src/main/java/org/wildfly/extension/elytron/CredentialStoreAliasDefinition.java#L196) might not be reachable but I left it there for now.


bstansberry:
@michpetrov I believe your change is correct, and if the case-insensitive attribute on the credential store were set to true and the user tried to use lower case, they'd have problems. @ehsavoie please have a look to confirm.

@michpetrov Right now, changes to this subsystem are first being submitted as PRs against the 'ladybird' branch at https://github.com/wildfly-security-incubator/wildfly-core and then get brought in hear as part of a large change set that Darran Lofthouse submits each week. Unless Darran has told you otherwise, please submit your PR there.

PR #2312, opened 2017-04-04
[WFCORE-2623] Deprecating domain management methods ....

ropalka:
…onal dependencies configurations

### DIFF
--- a/domain-management/src/main/java/org/jboss/as/domain/management/CallbackHandlerFactory.java
+++ b/domain-management/src/main/java/org/jboss/as/domain/management/CallbackHandlerFactory.java
@@ -48,12 +48,17 @@ public interface CallbackHandlerFactory {
         private ServiceUtil() {
         }
 
+        @Deprecated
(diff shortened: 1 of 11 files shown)


darranl:
@kabir can you please let me know if this one will be merged?  I will probably get it into ladybird quite quickly if merged as we have other fixes in these classes.

ropalka:
These deprecated methods are used in WildFly so we need to do this change in three steps @bstansberry :
 * first deprecate and provide alternative in Core
 * second switch wildfly to non deprecated methods (once core is updated there)
 * third (and finally) eliminate the deprecated methods
This way we don't break WildFly builds.

darranl:
I would not be surprised of JDG use them as well - deprecation seems safer all round.

bstansberry:
Sounds fine. I see that the static method in SecurityRealm is used. We should replace that use with a capability, but that's a separate thing from what Richard described above.

PR #2314, opened 2017-04-05
[WFCORE-2628]:  DeploymentPlan doesn't have an operation to explode content.

ehsavoie:
Adding an operation to be able to explode sub-content as only the root deployment could be exploded.

Jira: https://issues.jboss.org/browse/WFCORE-2628
JBEAP: https://issues.jboss.org/browse/JBEAP-10161


ehsavoie:
retest this please

PR #2322, opened 2017-04-07
WFLY-8540 Former singleton provider node leaving cluster after re-election causes another re-election

pferraro:
https://issues.jboss.org/browse/WFLY-8540

Currently, deployment unit phase services do not support restarts.  Rather than decorating a specific phase service as a singleton service, we decorate a service that installs the initial phase.  Once a deployment is detected to be a singleton service, the requisite DeploymentUnitPhaseBuilder is attached to the deployment unit, and the deployment unit service is restarted.  We retain the reference to the DeploymentUnitPhaseBuilder and build the initial phase within a service decorated with that builder.  This service can be restarted without issue - and singleton deployments work as expected.

Integration CI runs are here:
https://ci.wildfly.org/viewLog.html?buildId=54928
https://ci.wildfly.org/viewLog.html?buildId=55066

### DIFF
--- a/server/src/main/java/org/jboss/as/server/deployment/AbstractDeploymentUnitService.java
+++ b/server/src/main/java/org/jboss/as/server/deployment/AbstractDeploymentUnitService.java
@@ -33,7 +33,6 @@ import org.jboss.as.server.logging.ServerLogger;
 import org.jboss.as.server.services.security.AbstractVaultReader;
 import org.jboss.msc.inject.Injector;
 import org.jboss.msc.service.Service;
-import org.jboss.msc.service.ServiceBuilder;
 import org.jboss.msc.service.ServiceController;
 import org.jboss.msc.service.ServiceName;
 import org.jboss.msc.service.ServiceRegistry;
@@ -43,6 +42,7 @@ import org.jboss.msc.service.StartContext;
 import org.jboss.msc.service.StartException;
 import org.jboss.msc.service.StopContext;
 import org.jboss.msc.value.InjectedValue;
+import org.wildfly.common.function.ExceptionConsumer;
 
 /**
  * Abstract service responsible for managing the life-cycle of a {@link DeploymentUnit}.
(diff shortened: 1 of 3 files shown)


server/src/main/java/org/jboss/as/server/services/FunctionalVoidService.java
```diff
@@ -0,0 +1,69 @@
...
+import org.jboss.msc.service.StartContext;
+import org.jboss.msc.service.StartException;
+import org.jboss.msc.service.StopContext;
+import org.wildfly.common.function.ExceptionConsumer;
+
+/**
+ * Generic {@link Service} that provides no value, and delegates its start/stop to consumers.
+ * @author Paul Ferraro
+ */
+public class FunctionalVoidService implements Service<Void> {
```
bstansberry: Do you intend for this to be used outside this module?

I ask because I'd like to break this module in two, roughly most things in the "deployment" package(s) in one module, usable by extensions, and then the rest in another, not usable by extensions. To be able to support end-user extension writing in EAP we're going to need to do that kind of separation. (Same in the contrller module.) So I want to avoid classes used by extensions getting spread around.

pferraro: This class isn't referenced from anywhere outside this module - so in the short term, I would say to keep it with the deployment packages.  In the long term, if there is a better place to put generic service implementations, maybe we can move it there.

pferraro:
The full-integration failures are to be expected until the wildfly PR is merged.

kabir:
@pferraro Can you open some experimental jobs to prove that the core and full PRs together will pass? Without that the chances of getting these merged are pretty low :) These are the two jobs:
https://ci.wildfly.org/viewType.html?buildTypeId=WF_WildFlyCoreIntegrationExperiments
https://ci.wildfly.org/viewType.html?buildTypeId=WF_WildFlyCoreIntegrationExperimentsWindows

Please include links to the jobs once you have started them

pferraro:
@kabir FYI:
https://ci.wildfly.org/viewLog.html?buildId=54928
https://ci.wildfly.org/viewLog.html?buildId=54929

kabir:
Unfortunately the windows one needs a rerun, it ran out of disk space before complete (one of the banes of the CI at the moment)

server/src/main/java/org/jboss/as/server/services/FunctionalVoidService.java
bstansberry: Ok, so please move it to the deployment package. :)

pferraro:
@kabir New windows build: https://ci.wildfly.org/viewLog.html?buildId=55064

PR #2324, opened 2017-04-08
[WFCORE-2633] Allow user to declare that undefined attributes with de…

bstansberry:
…fault values are not considered sensitive in add operations

The original approach I outlined in the WFCORE-2633 description doesn't work. The read-resource-description access control data for an attribute with a "defined but non-sensitive" value can't be correctly determined as the answer depends on the values in a specific resource, while r-r-d is often used for the wildcard address. So, tools like HAL would not be able to get an accurate r-r-d answer for use in tailoring the UI.

Instead I went for a much simpler approach to the WildFly issue that WFCORE-2633 was trying to address. Now the kernel allows an attribute with a default value in its definition to be configured such that the existence of the default value is ignored during the add op, so long as the attribute is not specifically configured (i.e. it's value is undefined). This is simply a small change to existing behavior, utilizing the already present AttributeDefinition.isNullSignificant() setting.  Things work as before, so clients will see no differences, but now the existing behavior can be applied to more attributes. For attributes where AttributeDefinition.isNullSignificant() == false, the attribute can be set with an undefined value by a non-sensitive user in an add op, but can't be edited by such a user, regardless of new or current value, and can't be defined in the add op.

### DIFF
--- a/controller/src/main/java/org/jboss/as/controller/AbstractAttributeDefinitionBuilder.java
+++ b/controller/src/main/java/org/jboss/as/controller/AbstractAttributeDefinitionBuilder.java
@@ -615,7 +615,8 @@ public abstract class AbstractAttributeDefinitionBuilder<BUILDER extends Abstrac
      * the default value is whether the attribute {@link AttributeDefinition#isRequired()} () is not required} and
      * has a {@link AttributeDefinition#getDefaultValue() default value}.
      *
-     * @param nullSignificant {@code true} if an undefined value is significant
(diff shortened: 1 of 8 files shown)


bstansberry:
Custom CI runs with the WFLY-8521 and WFCORE-2633 fixes in combo:

https://ci.wildfly.org/viewLog.html?buildId=54626&tab=buildResultsDiv&buildTypeId=WF_WildFlyCoreIntegrationExperiments

https://ci.wildfly.org/viewLog.html?buildId=54628&tab=buildResultsDiv&buildTypeId=WF_WildFlyCoreIntegrationExperimentsWindows

PR #2325, opened 2017-04-10
update classfilewriter to 1.2.1.Final

tisoft:
This fixes the following error, that occurs if you have default methods in resteasy interfaces:

java.lang.VerifyError: Expecting a stackmap frame at branch target 13

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -98,7 +98,7 @@
         <version.org.fusesource.jansi>1.11</version.org.fusesource.jansi>
         <version.org.jboss.aesh>0.66.15</version.org.jboss.aesh>
         <version.org.jboss.byteman>3.0.6</version.org.jboss.byteman>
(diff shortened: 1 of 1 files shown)


bstansberry:
this is ok to test

bstansberry:
@dmlloyd @mkouba this is going to need an ok from you guys as the uses of this lib that I see are in weld, invocation and ejb3.

mkouba:
@bstansberry This should be ok. Weld compiles against classfilewriter 1.1.x but is compatible with (and even leverages some features of) 1.2.x. We don't run the full test suite against 1.2.x though. @tremes @manovotn guys could you verify the test suite (including TCK) passes with this update?

manovotn:
@mkouba @bstansberry Weld looks OK -> tested Weld (2.4) with WildFly (master) with this PR applies and found no problems

bstansberry:
retest this please

bstansberry:
Thanks!

PR #2330, opened 2017-04-11
WFCORE-2620 - Add ability to read computed runtime values of IO subsystem buffer-pool attributes

rpelisse:
Issue: [JBEAP-6984](https://issues.jboss.org/browse/JBEAP-6984)
Upstream Issue: [WFCORE-2620](https://issues.jboss.org/browse/WFCORE-2620)

### DIFF
--- a/io/subsystem/src/main/java/org/wildfly/extension/io/BufferPoolResourceDefinition.java
+++ b/io/subsystem/src/main/java/org/wildfly/extension/io/BufferPoolResourceDefinition.java
@@ -85,15 +85,18 @@ class BufferPoolResourceDefinition extends PersistentResourceDefinition {
             .setFlags(AttributeAccess.Flag.RESTART_ALL_SERVICES)
             .setAllowExpression(true)
             .setValidator(new IntRangeValidator(1, true, true))
+            .setDefaultValue(new ModelNode(defaultBufferSize))
             .build();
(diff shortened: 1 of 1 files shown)


bstansberry:
Failures are unrelated.

bstansberry:
Using the attribute definition default value for this is not accurate, as there is no global default that applies across a managed domain.

This approach could be done if there are separate AttributeDefinition instances for the attributes on a server process vs those for the domain wide resources. That would mean separate BufferPoolResourceDefinition and IORootDefinition instances for the two cases. IOExtension can determine what environment it is executing in and use the correct one.

Also, if these static fields are used this way the other uses of them in BufferPoolAdd should be removed. For example

```
final ModelNode directModel = DIRECT_BUFFERS.resolveModelAttribute(context, model);
...
final boolean direct = directModel.isDefined() ? directModel.asBoolean() : defaultDirectBuffers;
```

becomes

```
final boolean direct = DIRECT_BUFFERS.resolveModelAttribute(context, model).asBoolean();
```

rpelisse:
@ctomc thanks for the tip ! I'll rework the PR in a similar fashion then.

@bstansberry thanks for confirming the failures are unrelated, I was not sure. Granted, I overlooked the domain scenario in my naive proposal. Let me rework that and come up with a better solution (or at least a better start of a solution).

bstansberry:
Closing this for now to clean out the queue, as there will be substantial changes coming. Please open a new PR when ready. Thanks!

PR #2331, opened 2017-04-11
[WFCORE-2647] Add an option to always send the client SSL certificate to

ppalaga:
LDAP server

https://issues.jboss.org/browse/WFCORE-2647

### DIFF
--- a/domain-management/src/main/java/org/jboss/as/domain/management/connections/ldap/LdapConnectionAddHandler.java
+++ b/domain-management/src/main/java/org/jboss/as/domain/management/connections/ldap/LdapConnectionAddHandler.java
@@ -23,6 +23,7 @@
 package org.jboss.as.domain.management.connections.ldap;
 
 import static org.jboss.as.controller.descriptions.ModelDescriptionConstants.OP_ADDR;
+import static org.jboss.as.domain.management.connections.ldap.LdapConnectionResourceDefinition.ALWAYS_SEND_CLIENT_CERT;
 import static org.jboss.as.domain.management.connections.ldap.LdapConnectionResourceDefinition.HANDLES_REFERRALS_FOR;
 import static org.jboss.as.domain.management.connections.ldap.LdapConnectionResourceDefinition.INITIAL_CONTEXT_FACTORY;
 import static org.jboss.as.domain.management.connections.ldap.LdapConnectionResourceDefinition.REFERRALS;
@@ -51,6 +52,7 @@ import org.jboss.dmr.ModelNode;
 import org.jboss.msc.service.ServiceBuilder;
 import org.jboss.msc.service.ServiceController;
 import org.jboss.msc.service.ServiceTarget;
+import org.wildfly.security.manager.WildFlySecurityManager;
 
 /**
  * Handler for adding ldap management connections.
@@ -125,7 +127,17 @@ public class LdapConnectionAddHandler extends AbstractAddStepHandler {
             handlesReferralsForSet = Collections.emptySet();
         }
 
-        return connectionManagerService.setConfiguration(initialContextFactory, url, searchDn, searchCredential, referralHandling, handlesReferralsForSet);
+        final boolean alwaysSendClientCert;
+        if (model.hasDefined(ALWAYS_SEND_CLIENT_CERT.getName())) {
+            /* If always-send-client-cert is defined in the management model, take it */
+            alwaysSendClientCert = ALWAYS_SEND_CLIENT_CERT.resolveModelAttribute(context, model).asBoolean();
+        } else {
+            /* Otherwise, use the system property if set, or false as default */
+            alwaysSendClientCert = Boolean.valueOf(WildFlySecurityManager
+                    .getPropertyPrivileged("jboss.as.management.outbound.ldap.alwaysSendClientCert", "false"));
+        }
+
+        return connectionManagerService.setConfiguration(initialContextFactory, url, searchDn, searchCredential, referralHandling, handlesReferralsForSet, alwaysSendClientCert);
     }
 
 }
--- a/controller/src/main/java/org/jboss/as/controller/parsing/Attribute.java
+++ b/controller/src/main/java/org/jboss/as/controller/parsing/Attribute.java
@@ -46,6 +46,7 @@ public enum Attribute {
     ALLOW_EMPTY_PASSWORDS("allow-empty-passwords"),
     ALLOWED_ORIGINS("allowed-origins"),
     ALLOWED_USERS("allowed-users"),
+    ALWAYS_SEND_CLIENT_CERT("always-send-client-cert"),
     APP_NAME("app-name"),
     APPLICATION("application"),
     ARCHIVE("archive"),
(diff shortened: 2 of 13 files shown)


ppalaga:
@darranl could you please throw a quick look esp. if I got the sysprop vs. management model priorities right? https://github.com/wildfly/wildfly-core/pull/2331/commits/813dd533323bd3ddd7c869d8067e6962dea4e4ec#diff-d70cf63b97e558172f194715f77d39b5R131

I can say that the code is doing all it is supposed to do in manual tests. I also checked that the auth fails with an incorrect password.

Now, I am going to add a transformer and look how to test this.

domain-management/src/main/java/org/jboss/as/domain/management/connections/ldap/LdapConnectionAddHandler.java
```diff
@@ -125,7 +127,17 @@ static Config updateRuntime(final OperationContext context, final ModelNode mode
             handlesReferralsForSet = Collections.emptySet();
         }
 
-        return connectionManagerService.setConfiguration(initialContextFactory, url, searchDn, searchCredential, referralHandling, handlesReferralsForSet);
+        final boolean alwaysSendClientCert;
+        if (model.has(ALWAYS_SEND_CLIENT_CERT.getName()) && model.get(ALWAYS_SEND_CLIENT_CERT.getName()).isDefined()) {
```
bstansberry: if (model.hasDefined(ALWAYS_SEND_CLIENT_CERT.getName())) {

ppalaga: Thanks, accepted in 73dcdcb2d2b25a7695041a30179de91cc42018e1

bstansberry:
@ppalaga I just noticed your sentence about a transformer. You don't need a transformer for this. These resources are not part of the domain wide configuration, they are part of the host-specific configuration (i.e. persisted in host.xml.) So the DC is not responsible for managing this stuff for any HC other than itself and therefore it doesn't need to know how to transform it for slaves running previous versions.

ppalaga:
@bstansberry thanks for the info about not having to add a transformer. Do you happen to have any tip where should I add a test for this?

bstansberry:
The handling of the attribute in the kernel management layer (parsing, marshalling etc) could be tested via org.jboss.as.core.model.test.security.HostLdapTestCase. That doesn't test the interaction with the LDAP server though.

In the full WildFly testsuite there is org.jboss.as.test.manualmode.security.OutboundLdapConnectionTestCase which seems relevant. I haven't looked at that one closely but I suspect the only reason it is in full WF instead of core is because for convenience it's using a servlet as the mechanism for having an authenticated request.

domain-management/src/main/java/org/jboss/as/domain/management/connections/ldap/LdapConnectionAddHandler.java
```diff
@@ -125,7 +127,17 @@ static Config updateRuntime(final OperationContext context, final ModelNode mode
...
 
-        return connectionManagerService.setConfiguration(initialContextFactory, url, searchDn, searchCredential, referralHandling, handlesReferralsForSet);
+        final boolean alwaysSendClientCert;
+        if (model.hasDefined(ALWAYS_SEND_CLIENT_CERT.getName())) {
+            /* If always-send-client-cert is defined in the management model, take it */
+            alwaysSendClientCert = ALWAYS_SEND_CLIENT_CERT.resolveModelAttribute(context, model).asBoolean();
+        } else {
+            /* Otherwise, use the system property if set, or false as default */
+            alwaysSendClientCert = Boolean.valueOf(WildFlySecurityManager
+                    .getPropertyPrivileged("jboss.as.management.outbound.ldap.alwaysSendClientCert", "false"));
```
ctomc: why introducing new system property? 
You have defined configuration property, as such I don't see any good reason to have this extra property here.

Also, if new property is really *needed* it should be added as default value to the attribute and resolve that instead of all this checking here.

domain-management/src/main/java/org/jboss/as/domain/management/connections/ldap/LdapConnectionAddHandler.java
```diff
@@ -125,7 +127,17 @@ static Config updateRuntime(final OperationContext context, final ModelNode mode
             handlesReferralsForSet = Collections.emptySet();
         }
 
-        return connectionManagerService.setConfiguration(initialContextFactory, url, searchDn, searchCredential, referralHandling, handlesReferralsForSet);
+        final boolean alwaysSendClientCert;
+        if (model.hasDefined(ALWAYS_SEND_CLIENT_CERT.getName())) {
+            /* If always-send-client-cert is defined in the management model, take it */
+            alwaysSendClientCert = ALWAYS_SEND_CLIENT_CERT.resolveModelAttribute(context, model).asBoolean();
```
ctomc: given you have default value defined on attribute all this should be simplified to
`boolean alwaysSendClientCert = ALWAYS_SEND_CLIENT_CERT.resolveModelAttribute(context, model).asBoolean();`

and that is it.

ppalaga:
> why introducing new system property? You have defined configuration property, as such I don't see any good reason to have this extra property here.

The sysprop part is there to make this fix backportable to older versions and it is approved by Darran.

> Also, if new property is really needed it should be added as default value to the attribute and resolve that instead of all this checking here.

Yes, thanks for the hint, you seem to be right that pulling in the sysprop via attribute default would be equivalent with the present proposal's long and explicit code. Anyway, I took the long and explicit way to make it easier to the reviewers to check whether I got the precedence if model-attrib vs sysprop right. Once I get this approved I am ready to simplify the expression in favor of @ctomc 's proposal. 

> Also this should at least need some simple test that would show that at least parser works.

7a4dc8447a36167bf6270907905df14bb2f9972b aims at that, inspired by the advice of @bstansberry to look at `HostLdapTestCase`.

bstansberry:
-1 to making the default value dynamic (i.e. configurable via system property.) Our management API should be stable and system property values aren't even stable across all processes in a domain. The system property is just a work around that can be documented in the text description of the attribute. If anyone sets the property but not the attribute and then is confused about the value of the attribute then they'll get an explanation.

ppalaga:
> -1 to making the default value dynamic

Accepted.

ppalaga:
2bce571 replaces the previous iteration. It just adds some changes required by an integration test that I just submitted to WildFly: https://github.com/wildfly/wildfly/pull/9961

darranl:
Overall I approve but will let any model specific discussions continue.

Also +1 on Brian's system property comment - the system property is required here to provide compatibility with previous releases but should not be tightly tied into the management model.  If the new attribute is set we can assume they have migrated to using the model, if not we can check the property.

ppalaga:
> Overall I approve but will let any model specific discussions continue.

Thanks for the review @darranl. 

@bstansberry, @ctomc, @kabir, does anybody of you want to add any more comments? I'd need to know if I can consider this approved so that I can start backporting to older versions.

bstansberry:
retest this please

PR #2332, opened 2017-04-11
[WFCORE-2665]: RemotingModelControllerClient should be able to be used with security manager

ehsavoie:
Adding privileged blocks to allow arquillian tests to call model operations.
Protecting those privileged blocks in RemotingModelControllerClient with a specific permission.

Jira: https://issues.jboss.org/browse/WFCORE-2665

### DIFF
--- a/controller-client/src/main/java/org/jboss/as/controller/client/ModelControllerClientPermission.java
+++ b/controller-client/src/main/java/org/jboss/as/controller/client/ModelControllerClientPermission.java
@@ -1,0 +1,69 @@
+/*
+ * Copyright 2017 JBoss by Red Hat.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.jboss.as.controller.client;
+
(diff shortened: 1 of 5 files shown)


controller-client/src/main/java/org/jboss/as/controller/client/RemotingModelControllerClientPermission.java
```diff
@@ -0,0 +1,60 @@
...
+ * <p>
+ * The target name is the name of the permission. The following table lists all the possible {@link org.jboss.as.controller.client.RemotingModelControllerClientPermission} target names,
+ * and for each provides a description of what the permission allows.
+ * <ul>
+ * <li>performRemoteCall: permission to perform a remote call on the server model controller</li>
+ * </ul>
+ * </p>
+ * @author Emmanuel Hugonnet (c) 2017 Red Hat, inc.
+ */
+public class RemotingModelControllerClientPermission extends BasicPermission {
```
bstansberry: s/Remoting/Remote

"Remoting" is an implemenation detail; ie part the name of the lib we use to allow remote management.

bstansberry:
It feels odd to do a "RemoteModelControllerClientPermission.performRemoteCall" perm check in order to instantiate the client. *Perhaps* the perm class should just be ModelControllerClientPermission and then the perm check is on AbstractModelControllerClient.performRemoteCall.

That broadens the scope of who gets perm checks applied (e.g. the CLI will now as will the JConsoleCliPlugin). And in the CliModelControllerClient case at least, the remoting Endpoint is created outside the client, so this perm check would not remove the need to grant the remoting perms for the endpoint creation.

Plus CliModelControllerClient is creating a channel the same way RemotingModelControllerClient does, but there's no doPrivileged there.

Really, this needs a design post and discussion on wildfly-dev or someplace. I don't want to do something partial just to make a testsuite issue go away and then end up finding what we did doesn't allow us to have a nice overall solution later.

bstansberry:
Closing as there is ongoing work; please reopen when ready.

PR #2336, opened 2017-04-12
[WFCORE-2670] Deal with undefined values for the final part of enhanced syntax expressions

bstansberry:
https://issues.jboss.org/browse/WFCORE-2670

### DIFF
--- a/controller/src/main/java/org/jboss/as/controller/operations/global/EnhancedSyntaxSupport.java
+++ b/controller/src/main/java/org/jboss/as/controller/operations/global/EnhancedSyntaxSupport.java
@@ -21,6 +21,9 @@ package org.jboss.as.controller.operations.global;
 import java.util.regex.Matcher;
 import java.util.regex.Pattern;
 
+import org.jboss.as.controller.AttributeDefinition;
+import org.jboss.as.controller.ObjectListAttributeDefinition;
+import org.jboss.as.controller.ObjectTypeAttributeDefinition;
 import org.jboss.as.controller.OperationFailedException;
 import org.jboss.as.controller.PathAddress;
 import org.jboss.as.controller.logging.ControllerLogger;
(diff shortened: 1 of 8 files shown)


bstansberry:
@ctomc I've been experimenting with EnhancedSyntaxTestCase here and in the 2.1.x branch to see what may have worked before vs what works now in terms of automagically adding missing bits on the path the final point in the expression.; That is, to check for regressions since 2.1 that we need to repair.

I don't find any, and this PR enhances what we had in 2.1.x by allowing list-add(name=object.list,value=foo) when 'object' is defined but 'list' is not. It does not allow it if 'object' is undefined. See the TODO comment in the test case addition in the PR. I think it's fine to defer supporting that to a later release.

controller/src/main/java/org/jboss/as/controller/operations/global/EnhancedSyntaxSupport.java
```diff
@@ -51,16 +54,24 @@ static String extractAttributeName(String expression) {
         }
     }
 
-    static ModelNode resolveEnhancedSyntax(final String attributeExpression, final ModelNode model) throws OperationFailedException {
+    static ModelNode resolveEnhancedSyntax(final String attributeExpression, final ModelNode model, AttributeDefinition attributeDefinition) throws OperationFailedException {
         ModelNode result = model;
-        for (String part : attributeExpression.split("\\.")) {
+        String[] parts = attributeExpression.split("\\.");
+        for (int i = 0; i < parts.length; i++) {
```
luck3y: Possibly need a null check here? I'm not sure if attributeExpression could get here null.

bstansberry: It would not be null. I added a couple asserts though because that's not obvious.

luck3y:
I have no idea why it won't show it as me approving! Sorry for the spam though & LGTM!

controller/src/test/java/org/jboss/as/controller/operation/global/EnhancedSyntaxTestCase.java
```diff
@@ -342,6 +356,68 @@ public void testComplexAttributes() throws OperationFailedException {
...
+        item = new ModelNode();
+        item.get(ATTR_1.getName()).set("newvalue");
+        item.get(ATTR_2.getName()).set(true);
+        mapValue = item.get(MAP_ATTRIBUTE.getName());
+        mapValue.get("key1").set("value1");
+        mapValue.get("key2").set("value2");
+        op = createOperation("list-add", TEST_ADDRESS);
+        op.get("name").set(COMPLEX_ATTRIBUTE2.getName() + "." + OBJECT_LIST.getName());
+        op.get(VALUE).set(item);
+        executeCheckForFailure(op);  // TODO perhaps this could be made to work and this line would be replaced by the following
```
ctomc: why should this fail?
write operations of which list-add is one off, should be able to write whole structure.
That is one of the useful features of this syntax

bstansberry: I added this test basically to cover the current behavior, so if it changes we'd notice and could be sure that the change was intentional/valid. I didn't have time to pursue actually making this work though.

PR #2338, opened 2017-04-14
[WFCORE-2675]: Simplify and fix ContentReference management

ehsavoie:
Using RollBackHandler to manage rollback and put adding the ContentReference in the execution step.
Fixing some rollback missing issues.

Jira: https://issues.jboss.org/browse/WFCORE-2675

### DIFF
--- a/host-controller/src/main/java/org/jboss/as/domain/controller/operations/deployment/DeploymentExplodeHandler.java
+++ b/host-controller/src/main/java/org/jboss/as/domain/controller/operations/deployment/DeploymentExplodeHandler.java
@@ -115,6 +115,7 @@ public class DeploymentExplodeHandler implements OperationStepHandler {
             newHash = DeploymentUploadUtil.synchronizeSlaveHostController(operation, address, fileRepository, contentRepository, backup, oldHash);
         }
         if (newHash != null) {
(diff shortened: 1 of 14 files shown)


host-controller/src/main/java/org/jboss/as/domain/controller/operations/deployment/DeploymentExplodeHandler.java
```diff
@@ -115,6 +115,7 @@ public void execute(OperationContext context, ModelNode operation) throws Operat
             newHash = DeploymentUploadUtil.synchronizeSlaveHostController(operation, address, fileRepository, contentRepository, backup, oldHash);
         }
         if (newHash != null) {
+            contentRepository.addContentReference(ModelContentReference.fromModelAddress(context.getCurrentAddress(), newHash));
```
bstansberry: To be picky, there is no ref to this content until a few lines later.

Also, the ResultHandler is not added unless contentRepository != null, which means the add/remove may not be consistent. I think actually the contentRepo is never null. If that's true it would be good to clean the null checking up and simplify this code.

ehsavoie: np :)

bstansberry:
I'm closing this to empty out the queue since your are doing other stuff related to this. Please file a new PR when all is ready.

PR #2339, opened 2017-04-17
ValidateModelStepHandler fixes

bstansberry:
1) https://issues.jboss.org/browse/WFCORE-2317 fix from #2219 
2) https://issues.jboss.org/browse/WFCORE-2674 -- deal with requires + alternatives properly
3) https://issues.jboss.org/browse/WFCORE-2676 -- fix the deployment metadata

### DIFF
--- a/controller/src/main/java/org/jboss/as/controller/ValidateModelStepHandler.java
+++ b/controller/src/main/java/org/jboss/as/controller/ValidateModelStepHandler.java
@@ -117,6 +117,9 @@ class ValidateModelStepHandler implements OperationStepHandler {
                 }
                 throw new OperationFailedException(ControllerLogger.ROOT_LOGGER.invalidAttributeCombo(attributeName, sb));
             }
+
+            handleObjectAttributes(model, attr, attributeName);
+
         }
         context.completeStep(OperationContext.RollbackHandler.NOOP_ROLLBACK_HANDLER);
     }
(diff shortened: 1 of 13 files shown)


controller/src/main/java/org/jboss/as/controller/ValidateModelStepHandler.java
```diff
@@ -147,9 +183,81 @@ public void execute(OperationContext context, ModelNode operation) throws Operat
...
+            String absoluteName = absoluteParentName + "." + subAttributeName;
+            if (!subModel.hasDefined(subAttributeName) && isRequired(subAttr, subModel)) {
+                throw new OperationFailedException(ControllerLogger.MGMT_OP_LOGGER.required(subAttributeName));
+            }
+            if (!subModel.hasDefined(subAttributeName)) {
+                continue;
+            }
+            String[] requires = subAttr.getRequires();
+            if (requires != null) {
+                for (final String required : requires) {
```
ctomc: what happens if requires is defined with dot notation, aka having more than one level of nesting?

bstansberry: That's not supported.

bstansberry:
@ctomc -- good call. I added some tests (which found a bug in how the nested validation of requires+alternatives was working.)

PR #2346, opened 2017-04-19
WFCORE-2652 WARNING on boot of server in case where JBOSS_HOME is not…

ctomc:
… properly set on Windows

### DIFF
--- a/server/src/main/java/org/jboss/as/server/logging/ServerLogger.java
+++ b/server/src/main/java/org/jboss/as/server/logging/ServerLogger.java
@@ -1273,6 +1273,10 @@ public interface ServerLogger extends BasicLogger {
     @Message(id = 265, value = "Invalid value '%s' for system property '%s' -- value must be a non-negative integer")
     void invalidPoolCoreSize(String val, String configSysProp);
 
+    @LogMessage(level = WARN)
+    @Message(id = 266, value = "Server home is set to '%s', but server real home is '%s' - unpredictable results may occur.")
+    void serverHomeMismatch(Path passed, Path real);
+
     ////////////////////////////////////////////////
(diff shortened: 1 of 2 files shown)


ctomc:
@bstansberry Other probably even better thing to do would be to force use of toRealPath() as part of boot.
instead of just logging warning, but that might be a bit of behavior change, wdyt, should we change this?

ctomc:
@rsvoboda changed the PR so it corrects the server home if wrongly passed.

server/src/main/java/org/jboss/as/server/logging/ServerLogger.java
```diff
@@ -1273,6 +1273,10 @@ OperationFailedException illegalCombinationOfHttpManagementInterfaceConfiguratio
     @Message(id = 265, value = "Invalid value '%s' for system property '%s' -- value must be a non-negative integer")
     void invalidPoolCoreSize(String val, String configSysProp);
 
+    @LogMessage(level = WARN)
+    @Message(id = 266, value = "Server home is set to '%s', but server real home is '%s', This might cause problems.")
```
luck3y: This error is a bit vague, would it be of any value to mention an example of what it might cause?

bstansberry:
I asked @luck3y to look at this from the perspective of the offline CLI, as manipulation of these 'location' system properties is an important issue there.

luck3y:
This seems fine to me from the embedded side, other than the comment on the warning. We have a slightly different version in the startup scripts already:

"WARNING JBOSS_HOME may be pointing to a different installation - unpredictable results may occur."

ctomc:
@luck3y I've updated the message.

PR #2352, opened 2017-04-20
[WFCORE-2704] PersistentResourceXMLDescription should by default mars…

bstansberry:
…hal configured attribute values that match the default

We'll see what CI says.

### DIFF
--- a/controller/src/main/java/org/jboss/as/controller/PersistentResourceXMLDescription.java
+++ b/controller/src/main/java/org/jboss/as/controller/PersistentResourceXMLDescription.java
@@ -492,7 +492,7 @@ public final class PersistentResourceXMLDescription {
         private final LinkedHashMap<String, AttributeMarshaller> attributeMarshallers = new LinkedHashMap<>();
         private boolean useElementsForGroups = true;
         private String forcedName;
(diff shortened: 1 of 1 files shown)


ctomc:
I think we could remove the offending test all together now days.
It was there to test WildFly 8.0.0 (undertow subsystem 1.0.0) to WildFly 8.1 migration of default shipped config, as our initial 8.0.0 one was wrong

ctomc:
Looking bit further, with this change, subsystem templates we ship are no longer match the ones that server marshalls on first change. All templates should be updated before merging this.
Or just move to CLI based provisioning

bstansberry:
Thanks @ctomc both for the test removal suggestion and for noting the templates need tweaking. +1 for getting all that sorted. This isn't an urgent thing; I just saw a comment in a JIRA that led to this.

The template thing is a bit strange. Setting this to true should only mean we marshal a defined value even if it's the same as the default. So if our we marshall different from what we parsed, where is that defined value coming from? With the persistent resource stuff it shouldn't be from something custom in a parser. But no worries; investigating the diffs will reveal what's up easily enough.

bstansberry:
Closing until I can get around to dealing with the impacts.

PR #2353, opened 2017-04-20
[WFCORE-2705] Block when closing the RemotingModelControllerClient un…

jamezp:
…til it's resources have been closed.

https://issues.jboss.org/browse/WFCORE-2705

### DIFF
--- a/controller-client/src/main/java/org/jboss/as/controller/client/impl/RemotingModelControllerClient.java
+++ b/controller-client/src/main/java/org/jboss/as/controller/client/impl/RemotingModelControllerClient.java
@@ -23,6 +23,7 @@
 package org.jboss.as.controller.client.impl;
 
 import java.io.IOException;
+import java.io.InterruptedIOException;
 import java.util.concurrent.TimeUnit;
 
 import org.jboss.as.controller.client.ModelControllerClient;
@@ -88,12 +89,13 @@ public class RemotingModelControllerClient extends AbstractModelControllerClient
                 strategy = null;
             }
             // Then the endpoint
+            final Endpoint endpoint = this.endpoint;
             if (endpoint != null) {
+                this.endpoint = null;
                 try {
                     endpoint.closeAsync();
                 } catch (UnsupportedOperationException ignored) {
                 }
-                endpoint = null;
             }
             // Cancel all still active operations
             channelAssociation.shutdownNow();
@@ -103,6 +105,15 @@ public class RemotingModelControllerClient extends AbstractModelControllerClient
                 Thread.currentThread().interrupt();
             } finally {
                 StreamUtils.safeClose(clientConfiguration);
+            }
+            // Per WFCORE-1573 remoting endpoints should be closed asynchronously, however consumers of this client
+            // likely need to wait until the endpoints are fully shutdown.
+            if (endpoint != null) try {
+                endpoint.awaitClosed();
+            } catch (InterruptedException e) {
+                final InterruptedIOException cause = new InterruptedIOException(e.getLocalizedMessage());
+                cause.initCause(e);
+                throw cause;
             }
         }
     }


jamezp:
My one concern with with the to [block](https://github.com/wildfly/wildfly-core/pull/2353/files#diff-4e4bad7e8ceb8ef3c22af7cd50f6c35bR230) while closing the {{ManagementClientChannelStrategy}} will [block the close](https://github.com/jamezp/wildfly-core/blob/146ea680ab8e6fdd00fe61101cbfc32e05ea5d3e/controller-client/src/main/java/org/jboss/as/controller/client/impl/RemotingModelControllerClient.java#L87) of the {{RemotingModelControllerClient}} until the {{ProtocolConnectionManager}} has been fully closed. This may not be a huge issue, but something to note as it was changed to an {{asyncClosed()}} because of some hangs in the test suite.

bstansberry:
retest this please

jamezp:
retest this please

bstansberry:
retest this please

protocol/src/main/java/org/jboss/as/protocol/mgmt/ManagementClientChannelStrategy.java
```diff
@@ -88,13 +88,53 @@ public static ManagementClientChannelStrategy create(final ProtocolConnectionCon
...
+     * @return the management client channel strategy
+     */
+    public static ManagementClientChannelStrategy create(final ProtocolConnectionConfiguration baseConfig,
+                                                         final ManagementMessageHandler handler,
+                                                         final CallbackHandler cbHandler,
+                                                         final Map<String, String> saslOptions,
+                                                         final SSLContext sslContext,
+                                                         final CloseHandler<Channel> closeHandler,
+                                                         final boolean blockOnClose) {
+        return create(createConfiguration(baseConfig, saslOptions, cbHandler, sslContext), ManagementChannelReceiver.createDelegating(handler), closeHandler, blockOnClose);
```
bstansberry: Even though it is public this is an internal class so I'd rather just add a param to the existing methods than create new ones.

This management comms code is like a fat man struggling to lose weight, so no twinkies!

bstansberry: Another reason for that is I want all uses of the old method to be explicitly changed to the new one so there is no mystery in the code history about whether they are blockOnClose == false on purpose.

bstansberry:
This comment is all just for discussion, not a request that you change anything.

I'm not sure about the blockOnClose in ProtocolConnectionManager. I don't really understand the problems that led to the connection and endpoint close calls being made async, so it's hard for me to have an opinion. My impression from looking at the close logic in EndpointImpl is it would wait for connections to be closed anyway, so I'm not sure blocking waiting for the connection to close is needed. But I'm confident I don't understand this stuff well enough.

But, CLIModelControllerClient is inconsistent now, as it isn't using blockOnClose=true. Granted it functions differently from RemotingModelControllerClient because it keeps a single endpoint for the life of the VM, closing it in a shutdown hook. But, if blockOnClose=true before proceeding to closing the endpoint is is necessary for RemotingModelControllerClient it seems like it would be for the CLI too. The sequence of events the lead to the ultimate close of the endpoint may be different but in the end in the VM shutdown case you have connection close followed by endpoint close.

I'm curious why closeAsync() followed by awaitXXX is different/better than simply closing synchronously.

controller-client/src/main/java/org/jboss/as/controller/client/impl/RemotingModelControllerClient.java
```diff
@@ -104,6 +105,9 @@ public void close() throws IOException {
             } finally {
                 StreamUtils.safeClose(clientConfiguration);
             }
+            // Per WFCORE-1573 remoting endpoints should be closed asynchronously, however consumers of this client
+            // likely need to wait until the endpoints are fully shutdown.
+            if (endpoint != null) endpoint.awaitClosedUninterruptibly();
```
bstansberry: Should this just be endpoint.awaitClosed() with a catch for InterruptedException and conversion to InterruptedIOException, a la EnpdointImpl.close()?

Not being able to interrupt concerns me.

protocol/src/main/java/org/jboss/as/protocol/ProtocolConnectionManager.java
```diff
@@ -122,6 +122,16 @@ public void shutdown() {
     }
 
     /**
+     * Waits until the connection is closed.
+     */
+    public void awaitConnectionClosed() {
+        final Connection connection = this.connection;
+        if (connection != null) {
+            connection.awaitClosedUninterruptibly();
```
bstansberry: Same comment as for https://github.com/wildfly/wildfly-core/pull/2353/files#diff-8a7f63d9e35642093e986ee782c1b369R110

jamezp:
I'm a bit skeptical of the blocking close on the `ManagementClientChannelStrategy` myself. I was tempted to leave that the way it was, but had concerns the issue may be seen there as well. However I could not reproduce the issue with keeping the `ManagementClientChannelStrategy` close asynchronous. Maybe we should just revert that part and only block on the endpoint closing. That is where the original issue was coming from and seemed to fix the issue.

jamezp:
The commit message for the async close on the `ProtocolConnectionManager` says:
> Close the connection asynchronously to avoid a hang in ProtocolConnectionManager.shutdown()

I'm really starting think maybe I should revert the changes there. Waiting for the endpoint to close in `RemotingModelControllerClient` seems like enough.

jamezp:
I went ahead and reverted everything with the exception of the blocking on the endpoint. I ran the `wildfly:deploy` goal several times and did not see the issue reported in https://issues.jboss.org/browse/WFMP-77. This just feels like a safer approach to me as this stuff can be rather fragile.

jamezp:
retest this please

jamezp:
retest this please

jamezp:
retest this please

jamezp:
retest this please

PR #2358, opened 2017-04-23
Networking management improvments

bstansberry:
Main thing here is exposing the SBM as a capability which then allows us to use capability presence checking to replace an optional MSC dependency:

https://issues.jboss.org/browse/WFCORE-2707
https://issues.jboss.org/browse/WFCORE-2664

Doing that though made it clear there was a general need for some improvements in the related management code:

https://issues.jboss.org/browse/WFCORE-2713
https://issues.jboss.org/browse/WFCORE-2714


bstansberry:
@jamezp I expect at some point I'll create some constants.

PR #2360, opened 2017-04-25
WFCORE-2720 Adding keystore with generate-self-signed-certificate-hos…

stuartwdouglas:
…t and without key-password specified fails upon first request


domain-management/src/main/java/org/jboss/as/domain/management/security/SSLServerIdentityAddHandler.java
```diff
@@ -0,0 +1,57 @@
...
+    protected void updateModel(OperationContext context, ModelNode operation) throws OperationFailedException {
+
+
+        ModelNode selfSignedCert = KeystoreAttributes.GENERATE_SELF_SIGNED_CERTIFICATE_HOST.resolveModelAttribute(context, operation);
+        if(selfSignedCert.isDefined()) {
+            ModelNode password = KeystoreAttributes.KEY_PASSWORD.resolveModelAttribute(context, operation);
+            if(!password.isDefined()) {
+                throw DomainManagementLogger.ROOT_LOGGER.noKeyPasswordForSelfSignedCert();
+            }
+        }
```
bstansberry: Can't this just be "setRequires(ModelDescriptionConstants.KEY_PASSWORD)" on the GENERATE_SELF_SIGNED_CERTIFICATE_HOST definition builder?

stuartwdouglas: I did not realize that was possible, I will change it

stuartwdouglas:
Retest this please

PR #2361, opened 2017-04-25
[WFCORE-2717] try-catch block doesn't work if catch block doesn't contains anything

iweiss:
Issue: https://issues.jboss.org/browse/WFCORE-2717

### DIFF
--- a/cli/src/main/java/org/jboss/as/cli/handlers/trycatch/TryCatchFinallyControlFlow.java
+++ b/cli/src/main/java/org/jboss/as/cli/handlers/trycatch/TryCatchFinallyControlFlow.java
@@ -51,7 +51,7 @@ class TryCatchFinallyControlFlow implements CommandLineRedirection {
     }
 
     private Registration registration;
-    private List<String> tryList;
+    private final List<String> tryList = new ArrayList<>();
     private List<String> catchList;
     private List<String> finallyList;
     private int state;
@@ -99,6 +99,7 @@ class TryCatchFinallyControlFlow implements CommandLineRedirection {
     void moveToCatch() throws CommandLineException {
         switch(state) {
             case IN_TRY:
+                catchList = new ArrayList<>();
                 state = IN_CATCH;
                 break;
             case IN_CATCH:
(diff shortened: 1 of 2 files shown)


cli/src/main/java/org/jboss/as/cli/handlers/trycatch/TryCatchFinallyControlFlow.java
```diff
@@ -57,6 +57,7 @@ static TryCatchFinallyControlFlow get(CommandContext ctx) {
     private int state;
 
     TryCatchFinallyControlFlow(CommandContext ctx) {
+        tryList = new ArrayList<>();
```
jfdenise: You could make tryList final and perhaps inline initialisation in declaration.

jfdenise: doing so, any check for null in the class (I think there are some) should be removed.

cli/src/main/java/org/jboss/as/cli/handlers/trycatch/TryCatchFinallyControlFlow.java
```diff
@@ -97,6 +98,9 @@ boolean isInFinally() {
     }
 
     void moveToCatch() throws CommandLineException {
+        if(catchList == null) {
```
jfdenise: You could move the list initialisation in the switch (IN_TRY). Doing so, no need to check for null and you will only initialise the list when needed.

cli/src/main/java/org/jboss/as/cli/handlers/trycatch/TryCatchFinallyControlFlow.java
```diff
@@ -111,6 +115,9 @@ void moveToCatch() throws CommandLineException {
     }
 
     void moveToFinally() throws CommandLineException {
+        if(finallyList == null) {
```
jfdenise: You could move the list initialisation in the switch (IN_TRY and IN_CATCH). Doing so, no need to check for null and you will only initialise the list when needed.
Seems that you could get rid-off a call to break and merge the 2 cases.

jfdenise:
retest this please!

jfdenise:
retest this please!

PR #2362, opened 2017-04-26
WFCORE-2723 Add ability to remove providers in the Elytron config

stuartwdouglas:


darranl:
Lets get this sent to ladbird please so we can review and merge there - this is too risky a change to go in independently.

https://github.com/wildfly-security-incubator/wildfly-core

PR #2369, opened 2017-04-28
WFCORE-2735 - Missing desription of IO worker's server attributes

rpelisse:
Issue: [JBEAP-8447](https://issues.jboss.org/browse/JBEAP-8447)
Upstream Issue: [WFCORE-2735](https://issues.jboss.org/browse/WFCORE-2735)


io/subsystem/src/main/resources/org/wildfly/extension/io/LocalDescriptions.properties
```diff
@@ -24,6 +24,9 @@ io.worker.queue-size=An estimate of the number of tasks in the worker queue.
 io.worker.io-thread-count=I/O thread count
 io.worker.server=Runtime resource that describes listening servers utilizing this worker
 io.worker.outbound-bind-address=Defines bind addresses to use when connecting to specified destinations
+io.worker.server.connection-count=Estimate of the current connection count
+io.worker.server.connection-limit-low-water-mark=If the connection count hits this number, no new connections will be accepted until the count drops below the low-water mark.
+io.worker.server.connection-limit-high-water-mark=If the connection count has previously hit the high water mark, once it drops back down below this count, connections will be accepted again.
```
bstansberry: The text for these two -water-mark attributes is switched.

rpelisse:
@bstansberry thanks for spotting that, it's fix now. Let me know if there is anything looking bad here (and thanks again for the review)

PR #2308, opened 2017-04-03
WFCORE-2617 Credential store flush failure creates inconsistency...

TomasHofman:
... between the the version on the filesystem and in the memory

https://issues.jboss.org/browse/WFCORE-2617
https://issues.jboss.org/browse/JBEAP-9989

PR #2371, opened 2017-04-28
[WFCORE-2736] Upgrade WildFly Elytron to 1.1.0.Beta39 and backport latest integration changes.

darranl:
The following Jiras can be resolved with this PR: -

https://issues.jboss.org/browse/WFCORE-2478
https://issues.jboss.org/browse/WFCORE-2646
https://issues.jboss.org/browse/WFCORE-2718
https://issues.jboss.org/browse/WFCORE-2716
https://issues.jboss.org/browse/WFCORE-2736
https://issues.jboss.org/browse/WFCORE-2725

### DIFF
--- a/elytron/src/main/java/org/wildfly/extension/elytron/CredentialStoreService.java
+++ b/elytron/src/main/java/org/wildfly/extension/elytron/CredentialStoreService.java
@@ -114,6 +114,9 @@ class CredentialStoreService implements Service<CredentialStore> {
             credentialStoreAttributes.put(CS_LOCATION_ATTRIBUTE, loc.toAbsolutePath().toString());
             credentialStore = getCredentialStoreInstance();
             credentialStore.initialize(credentialStoreAttributes, resolveCredentialStoreProtectionParameter(), otherProviders.getOptionalValue());
+            if ( credentialStoreAttributes.get(ElytronDescriptionConstants.CREATE).equals("true") && !loc.toFile().exists() ){
+                credentialStore.flush();
+            }
         } catch (Exception e) {
             throw ElytronSubsystemMessages.ROOT_LOGGER.unableToStartService(e);
         }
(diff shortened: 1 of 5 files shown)


elytron/src/main/java/org/wildfly/extension/elytron/CredentialStoreService.java
```diff
@@ -114,6 +114,9 @@ public void start(StartContext startContext) throws StartException {
             credentialStoreAttributes.put(CS_LOCATION_ATTRIBUTE, loc.toAbsolutePath().toString());
             credentialStore = getCredentialStoreInstance();
             credentialStore.initialize(credentialStoreAttributes, resolveCredentialStoreProtectionParameter(), otherProviders.getOptionalValue());
+            if ( credentialStoreAttributes.get(ElytronDescriptionConstants.CREATE).equals("true") && !loc.toFile().exists() ){
+                credentialStore.flush();
```
bstansberry: [@darranl] [@yersan] I recall merging a recent change that had some catch logic related to flush, IIRC to ensure that the corresponding model change was reverted. Is that needed here?

darranl: @pskopek could you please have a look at the comment from Brian?  If something else is needed if we can have a PR against ladybird I will merge it into this PR.

bstansberry: I hadn't merged it yet but I did see it. :)

https://github.com/wildfly/wildfly-core/pull/2308/files

bstansberry: @pskopek Never mind. Looking at #2308 I see that it's about keeping the FS store and the in-memory one in sync. And that's not an issue here as this call is not storing anything to the in-memory store.

PR #2387, opened 2017-05-05
[WFCORE-2747 & WFCORE-2769]

jamezp:
The first commit adds a new `ContextualModelControllerClient` which allows a context to be used when executing operations. It also adds a dependency to `org.wildfly.common:wildfly-common` and `org.wildfly.elytron:wildfly-elytron`. Both of these were already transitive dependencies from the `org.wildfly.protocol:wildfly-protocol` dependency. It also adds an option to the client builder to allow a `wildfly-config.xml` to be defined for authentication.

The second commit enables the usage of authentication options for the `ModelControllerClient` so that an Elytron authentication configuration can be used to authentication the management client.

The next two could likely be squashed. I just left them separate for now.

### DIFF
--- a/controller-client/pom.xml
+++ b/controller-client/pom.xml
@@ -71,6 +71,14 @@
             <groupId>org.jboss.threads</groupId>
             <artifactId>jboss-threads</artifactId>
         </dependency>
+        <dependency>
+            <groupId>org.wildfly.common</groupId>
+            <artifactId>wildfly-common</artifactId>
+        </dependency>
+        <dependency>
+            <groupId>org.wildfly.security</groupId>
+            <artifactId>wildfly-elytron</artifactId>
+        </dependency>
 
         <!-- Test Dependencies -->
         <dependency>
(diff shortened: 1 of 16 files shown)


jamezp:
These failures are legitimate.

bstansberry:
@jamezp What's the use case for the CustomCliExecutor stuff? It seems fine; I'm mostly curious, plus want to check if it's something @luck3y needs, etc.

The test case in the last commit can probably be removed eventually since I hope the overall WFCORE-2746 would make it redundant. But it's fine now; it proves that the stuff in this PR works.

jamezp:
@bstansberry The `CustomCliExecutor` stuff was just to execute offline CLI commands in a new process. The issue was that surefire put Elytron on the `java.class.path` and when looking up security providers it would use the library found on the class path and give a `ClassCastException` since the management operations themselves used the Elytron library on the module class path.

To be honest I'm somewhat surprised we hadn't seen this before. I'm not sure this is the best solution for it, but it was the easiest solution.

bstansberry:
@jamezp @luck3y That classpath problem tells me it's important that we get good elytron coverage in CLIEmbedServerTestCase and CLIEmbedHostControllerTestCase.

If an embedding process can't use the offline CLI but instead needs to spawn a separate process, that's a problem.

luck3y:
Understood, I'll take a look at adding some elytron tests in those.

jamezp:
@luck3y It's pretty easy to duplicate. I can write a quick example test that exposes the issue if you want.

luck3y:
@jamezp Sure, that'd be helpful. I've been using embedded doing test setup stuff, but I haven't seen it happen yet.

bstansberry:
Thanks, guys!

PR #2390, opened 2017-05-07
[WFCORE-2771] Console should log bound ports

jasontedor:
When specifying that Wildfly should bind the management console to an ephemeral port (by specifying the port number as 0), Wildfly logs that the management console is not enabled. In fact, this is not correct as can be verified with lsof (or any other tool that enables seeing the sockets that Wildfly has bound). The networking layer determined an actual port number when binding the socket. Instead, Wildfly should log the actual port that it is bound to. This commit causes this to be the case.

Relates https://issues.jboss.org/browse/WFCORE-2771

### DIFF
--- a/domain-http/interface/src/main/java/org/jboss/as/domain/http/server/ManagementHttpServer.java
+++ b/domain-http/interface/src/main/java/org/jboss/as/domain/http/server/ManagementHttpServer.java
@@ -26,6 +26,7 @@ import static org.xnio.Options.SSL_CLIENT_AUTH_MODE;
 
 import java.io.IOException;
 import java.net.InetSocketAddress;
+import java.net.SocketAddress;
 import java.nio.charset.Charset;
 import java.nio.charset.StandardCharsets;
 import java.security.cert.X509Certificate;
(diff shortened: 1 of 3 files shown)


luck3y:
This is OK to test

jasontedor:
I should add, this relates to wildfly/wildfly#9943 which did the same for the Undertow components.

luck3y:
@jasontedor Yep, sorry for the spam. Your first PR to each project needs to be whitelisted, then they auto run in CI after that.

luck3y:
this is ok to test

jasontedor:
> Yep, sorry for the spam.

No apology necessary, thanks for looking so quickly after opening!

luck3y:
retest this please

jasontedor:
I looked at the CI failure here:

```
java.lang.AssertionError: Deployment two shouldn't exist at 2017-05-07 05:24:45,151
    at org.junit.Assert.fail(Assert.java:88)
    at org.junit.Assert.assertTrue(Assert.java:41)
    at org.junit.Assert.assertFalse(Assert.java:64)
    at org.jboss.as.test.manualmode.deployment.DeploymentScannerUnitTestCase.testStartup(DeploymentScannerUnitTestCase.java:159)
```

I believe that this is unrelated to my change. Please let me know if that is wrong.

jasontedor:
Thanks @kabir.

PR #2394, opened 2017-05-08
WFCORE-1374 JDK9u166+ updates

ctomc:
- fix elytron integration module to have same surefire settings as others
- fix elytron test to properly bootstrap to management only
- update jboss-parent to 23
- update jboss-logging-tools to latest version that works with all "versions" of @Generated annotation
- update maven compiler config after logging-tools updated as we don't need to fork anymore
- update jmockit test dependency (only used by elytron) to one that works also with jdk9

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -26,7 +26,7 @@
     <parent>
       <groupId>org.jboss</groupId>
       <artifactId>jboss-parent</artifactId>
-      <version>22</version>
+      <version>23</version>
     </parent>
 
     <groupId>org.wildfly.core</groupId>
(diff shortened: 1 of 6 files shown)


ctomc:
JDK9 job run off this branch https://ci.wildfly.org/viewLog.html?buildId=57823&buildTypeId=WildFlyCore_MasterLinuxJdk9&tab=buildResultsDiv

darranl:
With the clean CI runs the Elytron changes look fine to me.

ctomc:
@bstansberry removed the domain related commit, as all other changes are also useful without it.
Will prepare separate PR for addressing domain

pom.xml
```diff
@@ -104,7 +104,7 @@
         <version.org.jboss.jboss-dmr>1.4.0.Final</version.org.jboss.jboss-dmr>
         <version.org.jboss.jboss-vfs>3.2.12.Final</version.org.jboss.jboss-vfs>
         <version.org.jboss.logging.jboss-logging>3.3.1.Final</version.org.jboss.logging.jboss-logging>
-        <version.org.jboss.logging.jboss-logging-tools>2.0.1.Final</version.org.jboss.logging.jboss-logging-tools>
+        <version.org.jboss.logging.jboss-logging-tools>2.1.0.Alpha2</version.org.jboss.logging.jboss-logging-tools>
```
bstansberry: @jamezp is going to look into this, whether the needed stuff can be in the 2.0.x stream or when the 2.1.x stream can get to .Final. It's too late to be bringing in alphas unless there is a clear plan to get them to .Final.

PR #2395, opened 2017-05-08
[WFCORE-2770] Move org.jboss.as.test.integration.security.perimeter.*…

jamezp:
… from WildFly to WildFly Core.

Tests were copied from WildFly full, however I did make additional changes to clean the tests up a bit. There is also change to enable the `CLISecurityTestCase`. The assumption was changed from assuming some text exists to assuming the CLI client is not connected.

https://issues.jboss.org/browse/WFCORE-2770


jamezp:
Failures seem legit on Windows, but odd as well. I'll have to setup a Windows VM to test these out.

bstansberry:
retest this please

PR #2396, opened 2017-05-08
Remove WildFly Swarm's self-contained bits.

bobmcwhirter:
Because WildFly Swarm has a different release candence
than WildFly Core, by moving all of this code out of
WildFly Core, we gain more control of our own destiny.


bobmcwhirter:
retest this please

bstansberry:
I'm ok with this if @n1hility is.

The downside to doing this is if we change stuff this uses we can no longer easily adapt this code to it. Instead we'd either have to do some compatibility thing or you'd have to update your code to use before pulling in our latest.

Here's the contracts we'd have to keep compatible to avoid issues:

ExtensibleConfigurationPersister
ContentRepository
Bootstrap / Bootstrap.Factory / Bootstrap.Configuration
ConfigurationPersisterFactory
ProductConfig.fromKnownSlot
ServerEnvironment constructor
various black magic related to logging and vfs loading in SelfContainedContainer.start 

I don't see a big problem in the above. ContentRepository is a bit worrisome, but interface method defaulting reduces the risk.

bobmcwhirter:
We understand the no guarantee on innards and accept those risks. They seem lower than the risk of our bugs being in your codebase and incompatible release cadences.

PR #2399, opened 2017-05-09
[WFCORE-2483]: Adding support for credential reference.

ehsavoie:
Registering and unregistering capabilities in SecurityRealmChildAddHandler
Adding a capability for Management security realms
Using Credential Reference as an alternative to passwords

Jira: https://issues.jboss.org/browse/WFCORE-2483
JBEAP: https://issues.jboss.org/browse/JBEAP-9321
Full PR to fix compilation issue: https://github.com/wildfly/wildfly/pull/10033

Experimental Build: https://ci.wildfly.org/viewLog.html?buildId=58122 and https://ci.wildfly.org/viewQueued.html?itemId=58126
Experimental Build on Windows: https://ci.wildfly.org/viewQueued.html?itemId=58129 and 
https://ci.wildfly.org/viewQueued.html?itemId=58145


kabir:
Retest this please

controller/src/main/java/org/jboss/as/controller/management/Capabilities.java
```diff
@@ -40,4 +40,6 @@
 
     public static final String SSL_CONTEXT_CAPABILITY = "org.wildfly.security.ssl-context";
 
+    public static final String MANAGEMENT_SECURITY_REALM_CAPABILITY = "org.wildfly.core.management.security.realm";
```
bstansberry: I think we'll want to change this name but as long as no one is going to use this outside of core before the core release next week, I don't care for now.

We also need wildfly-capabilities entries for all the elyron capabilities so it is easy to see how this name fits in with the others.

ehsavoie: All the usages in full go through SecurityRealm.ServiceUtil.createServiceName so we can easily rename

darranl: I have a fork of the capabilities repo with all of the Elytron capabilities defined, just needs an audit to double check recent changes have been applied and then it can be merged to the main repo.

https://github.com/wildfly-security-incubator/wildfly-capabilities

domain-management/src/main/java/org/jboss/as/domain/management/parsing/ManagementXml_5.java
```diff
@@ -2046,22 +2075,33 @@ private void writeServerIdentities(XMLExtendedStreamWriter writer, ModelNode rea
...
+                    writer.writeEmptyElement(Element.KEYSTORE.getLocalName());
+                }
                 KeystoreAttributes.KEYSTORE_PROVIDER.marshallAsAttribute(ssl, writer);
                 KeystoreAttributes.KEYSTORE_PATH.marshallAsAttribute(ssl, writer);
                 KeystoreAttributes.KEYSTORE_RELATIVE_TO.marshallAsAttribute(ssl, writer);
                 KeystoreAttributes.KEYSTORE_PASSWORD.marshallAsAttribute(ssl, writer);
                 KeystoreAttributes.ALIAS.marshallAsAttribute(ssl, writer);
                 KeystoreAttributes.KEY_PASSWORD.marshallAsAttribute(ssl, writer);
                 KeystoreAttributes.GENERATE_SELF_SIGNED_CERTIFICATE_HOST.marshallAsAttribute(ssl, writer);
+                KeystoreAttributes.KEYSTORE_PASSWORD_CREDENTIAL_REFERENCE.marshallAsElement(ssl, writer);
```
bstansberry: We can fix this later but this ordering of attributes isn't ideal. Probably this ref attributes should be next to the equivalent old ones, or at least not separated by GENERATE_SELF_SIGNED_CERTIFICATE_HOST.

kabir:
@ehsavoie I've by no means looked very carefully, but if possible some of the tests should try to use other attributes than just clear-text. e.g. in full I see things like 
```
connector/src/test/resources/org/jboss/as/connector/subsystems/datasources/datasources-elytron-enabled_5_0.xml:        <credential-reference store="test-store" alias="test-alias" type="org.wildfly.Foo" />

```

domain-management/src/main/java/org/jboss/as/domain/management/security/UserDomainCallbackHandler.java
```diff
@@ -196,11 +210,35 @@ public void handle(Callback[] callbacks) throws IOException, UnsupportedCallback
                     SECURITY_LOGGER.tracef("User '%s' not found.", userName);
                     throw new UserNotFoundException(userName);
                 }
-                String password = user.require(PASSWORD).asString();
-                ((PasswordCallback) current).setPassword(password.toCharArray());
+                char[] password = new char[0];
+                if (user.hasDefined(PASSWORD)) {
+                    password = user.require(PASSWORD).asString().toCharArray();
+                }
+                SECURITY_LOGGER.tracef("User '%s' has the password '%'.",user, resolvePassword(userCredentialSourceSupplier, password));
```
bstansberry: This seems like a very bad idea. :)

darranl: Bad for two reasons.
1 - resolvePassword will be called regardless of if trace logging is enabled.
2 - It will lead to a bug report when clear passwords are found in logs ;-)

domain-management/src/main/java/org/jboss/as/domain/management/parsing/ManagementXml_5.java
ehsavoie: I'm marshalling as element while the other attributes are marshalled as attribute. Changing the order will not respect the xml.

PR #2400, opened 2017-05-09
WFCORE-2780 - Improve buffer-pool model description

rpelisse:
Issue: [JBEAP-9140](https://issues.jboss.org/browse/JBEAP-9140)
Upstream Issue: [WFCORE-2780](https://issues.jboss.org/browse/WFCORE-2780)

(no downstream required)


rpelisse:
The failing test above looks unrelated. On top of this, the changes are only an update in a resource bundle, can't really see how this could cause the EJB test to somehow fails.

rpelisse:
retest this please

rpelisse:
Retest this please

kabir:
Retest this please

kabir:
@rpelisse I just did a core release/upgrade with a lot of simultaneous stuff needing merge in full, so it could be related to that

io/subsystem/src/main/resources/org/wildfly/extension/io/LocalDescriptions.properties
```diff
@@ -14,9 +14,9 @@ io.worker.task-keepalive=Specify the number of milliseconds to keep non-core tas
 io.buffer-pool=Defines buffer pool
 io.buffer-pool.add=Adds new buffer pool
 io.buffer-pool.remove=Removes buffer pool
-io.buffer-pool.buffers-per-slice=How many buffers per slice, if not set optimal value is calculated based on available RAM resources in your system.
-io.buffer-pool.buffer-size=The size of each buffer slice in bytes, if not set optimal value is calculated based on available RAM resources in your system.
-io.buffer-pool.direct-buffers=Does the buffer pool use direct buffers, some platforms don't support direct buffers
+io.buffer-pool.buffers-per-slice=Maximum number of buffers
+io.buffer-pool.buffer-size=Size of the all buffer at initialization
```
rpelisse: @kabir yes, it is a bit confusing indeed. So when I look at the code, I saw that the NIO susbsys allocated one large buffer, at initialization, that one can then either use fully or sliced. The buffer-size value is obviously this value.

However, it can be somewhat confusing with the NIO documentation which also has concept of slice and buffer. This is what I try (and fail) to describe accurately here. Would this be a better description:

"Total size of the buffer handled by the subsystem" ?

kabir: @rpelisse I think that sounds good

rpelisse:
@kabir I've updated the description of 'io.buffer-pool.buffer-size' as we discussed,  let me know if you think we can improve the description more.

io/subsystem/src/main/resources/org/wildfly/extension/io/LocalDescriptions.properties
```diff
@@ -14,9 +14,9 @@ io.worker.task-keepalive=Specify the number of milliseconds to keep non-core tas
 io.buffer-pool=Defines buffer pool
 io.buffer-pool.add=Adds new buffer pool
 io.buffer-pool.remove=Removes buffer pool
-io.buffer-pool.buffers-per-slice=How many buffers per slice, if not set optimal value is calculated based on available RAM resources in your system.
-io.buffer-pool.buffer-size=The size of each buffer slice in bytes, if not set optimal value is calculated based on available RAM resources in your system.
-io.buffer-pool.direct-buffers=Does the buffer pool use direct buffers, some platforms don't support direct buffers
+io.buffer-pool.buffers-per-slice=Maximum number of buffers
+io.buffer-pool.buffer-size=Total size of the buffer handled by the subsystem
```
bstansberry: This isn't the total size of the buffer. The existing comment is valid.

The basic problem here is "buffers-per-slice" would be better named "slices-per-buffer" and "buffer-size" would be better named "slice-size". Or, ignore the 'slice' concept and instead describe and refer to the 'region' concept, using 'buffers-per-region' and 'buffer-size'. The "buffers-per-slice" is particularly problematic as it is actually backwards in terms of what the attribute does.

Basically there is a larger backing byte buffer which is the unit of memory allocation from the JVM. Then there are smaller chunks of that that are provided to users of this resource. The larger backing buffer can be called 'buffer' and the smaller chunks are 'slices' or the larger backing buffer is a 'region' and the smaller chunks are 'buffers'.

The size of the larger backing buffer is buffers-per-slice * buffer-size.

We can't drop the existing attributes and if new better ones are added it's necessary to add a transformer to ensure that older version hosts in a managed domain are not broken. Probably better for the next release as we are about to only accept blockers.

rpelisse:
retest this please

rpelisse:
@bstansberry Your last comment has been removed somehow, so here it is for completion sake:

`This isn't the total size of the buffer. The existing comment is valid.

The basic problem here is "buffers-per-slice" would be better named "slices-per-buffer" and "buffer-size" would be better named "slice-size". Or, ignore the 'slice' concept and instead describe and refer to the 'region' concept, using 'buffers-per-region' and 'buffer-size'. The "buffers-per-slice" is particularly problematic as it is actually backwards in terms of what the attribute does.

Basically there is a larger backing byte buffer which is the unit of memory allocation from the JVM. Then there are smaller chunks of that that are provided to users of this resource. The larger backing buffer can be called 'buffer' and the smaller chunks are 'slices' or the larger backing buffer is a 'region' and the smaller chunks are 'buffers'.

The size of the larger backing buffer is buffers-per-slice * buffer-size.

We can't drop the existing attributes and if new better ones are added it's necessary to add a transformer to ensure that older version hosts in a managed domain are not broken. Probably better for the next release as we are about to only accept blockers.`

... so I feel the time has come to resume work on this one :) - if I understood your comment right, it seems the easiest way to improve the situation is just to rename the following attributes:
- "buffers-per-slice" would be better named "slices-per-buffer" and 
- "buffer-size" would be better named "slice-size"

...And we have to provide a transformer for that. I plan to do so following the `[available documentation on Transformer](https://docs.jboss.org/author/display/WFLY10/Domain+Mode+Subsystem+Transformers) (unless you have a better starting point) ?

bstansberry:
@dmlloyd fyi, in case you think I'm all wrong in my comment Romaine quoted above.

@rpelisse We can't do a simple rename. A simple rename is a breaking API change, since the old attributes immediately disappear, breaking scripts etc. The only way you can do a rename is to add new attributes, with a transformer, and deprecate the old ones. Then eventually you can remove the old ones. So a slow motion rename. ;)  

The transformer code would be a new method in IOSubsystemTransformers. I believe the transformation would simply be what's discussed at https://docs.jboss.org/author/display/WFLY10/Domain+Mode+Subsystem+Transformers#DomainModeSubsystemTransformers-Renamingattributes.  (Even though in the previous paragraph I describe this as not a simple rename, as far as the domain mode transformer is concerned, it looks like one. Any use of the new attribute name gets translated to the old one before being sent on to the legacy slave.)

dmlloyd:
I agree with your assessment @bstansberry.

rpelisse:
@bstansberry yes, slow motion remove is my plan :) - and thanks for pointing out the right part of the doc, this should help a lot! 

@dmlloyd thanks for confirming the assessment is correct. I'll start working (again) on this asap.

PR #2401, opened 2017-05-09
WFCORE-2768: Avoid multiple threads add concurrently credential store ﻿aliases

yersan:
Jira issues:
https://issues.jboss.org/browse/WFCORE-2768
https://issues.jboss.org/browse/JBEAP-10749


bstansberry:
I'm closing this one to help tidy the PR queue as this feature is being rewritten in wildfly-security-incubator and as we discussed in email there's also an issue with reads, so the store itself needs to be thread safe. So a new PR can be opened when all that is sorted out.

PR #2406, opened 2017-05-10
Fix for WFCORE-2782. CLI, advertise character encoding issues

jfdenise:
Revert UTF-8 constraints added when reading files from CLI. Redirected output is also not constrained by UTF-8 encoding. Encoding exception message is now clearer.


jfdenise:
Some discussions are on going. This PR shouldn't be merged yet.

jfdenise:
@ctomc , thank-you for your feedback. I agree that we could do better refactoring and offer a properly designed feature to better support encoding. For example, enforce UTF-8 usage and add a configuration item to relax the constraint. I would prefer to stick with existing behaviour in 7.1 and tackle it for 7.2, it would impact QA, documentation.

PR #2408, opened 2017-05-10
[WFCORE-2723] Ability to remove globally installed security providers.

darranl:
https://issues.jboss.org/browse/WFCORE-2723
https://issues.jboss.org/browse/JBEAP-10483


elytron/src/main/java/org/wildfly/extension/elytron/ElytronSubsystemParser.java
```diff
@@ -118,7 +118,9 @@ public void readElement(XMLExtendedStreamReader reader, List<ModelNode> operatio
                         ElytronDefinition.FINAL_PROVIDERS.parseAndSetParameter(value, subsystemAdd, reader);
                         break;
                     case DISALLOWED_PROVIDERS:
-                        ElytronDefinition.DISALLOWED_PROVIDERS.parseAndSetParameter(value, subsystemAdd, reader);
+                        for (String disallowed : reader.getListAttributeValue(i)) {
+                            ElytronDefinition.DISALLOWED_PROVIDERS.parseAndAddParameterElement(disallowed, subsystemAdd, reader);
+                        }
```
bstansberry: @ctomc This tells me we have a bug in the AttributeDefinition hierarchy. AD holds a ref to an AttributeParser but it isn't consistently used for parseAndSetParameter. ListAttributeDefinition doesn't use it; it hard codes splitting on ',' which is why Darran had to make this change.

I'll file a JIRA.

bstansberry: https://issues.jboss.org/browse/WFCORE-2792

PR #2410, opened 2017-05-10
WFCORE-2741: Add missing vault module attribute to the model boot updates sent by a HC

yersan:
In a managed domain a custom vault implementation doesn't work because the module name is not sent to the managed servers by the HC. This issue resolve it.

Jira issues:
https://issues.jboss.org/browse/WFCORE-2741
https://issues.jboss.org/browse/JBEAP-10646


bstansberry:
@ehsavoie we're seeing ProcessStateListenerTestCase failures again. :(

PR #2414, opened 2017-05-10
[WFCORE-2786] - Improve the composition operation output

spolti:


bstansberry:
Sorry; but no.

Perhaps I'd consider adding a field to the per-step output to repeat the value of the step's "operation" value. But the step-1 step-2 is not going to change. This is utilized in complex ways and having to account for different scenarios is too much effort and risk.

The "name" parameter in an operation is also not a reserved word with a fix meaning; it just happens to be a param in some operations. Since it's not a reserved word we can't treat it like it is.

spolti:
@bstansberry, thanks for your explanation.
I understood the reason behind the step-n.

I'll look on how add a new field to per-step ( I also suggested it on the jira).
But by using the operation value sometimes will give us operations like, for example, read/write-attribute, it but does not give much details as well, in case of 
> {"operation": "read-attribute", "name": "server-state"}

Will return something like this:
`"step-1": {
            "operation-name": "read-attribute",
            "outcome": "success",
            "result": "running"
        },`

What do you suggest in this case?

bstansberry:
@spolti Perhaps

```
"step-operation" => {
  "operation" => "read-attribute",
  "address" => [("subsystem","xyz",("foo","bar")],
  "name" => "baz"
}
```

with the 'name' bit only shown in the op has such a param. That's more verbose than simply reporting "name" => "baz" and still somewhat treats "name" in a magical way. But it's less magical.

jmesnil:
@spolti  As this pull request has been held up for a long time, I am closing it to get the pull request queue under control again. Since this is very old now, it will be a very long time until this one will be addressed. Also, I don't have much time to spend on this to evaluate on a case by case basis.

If it is a new feature, please follow the new process for features. Ping @wildfly/mergers-core  if you are unsure what this means, and we can help. Please do that before opening the pull request though.

Sorry for any inconvenience!

PR #2416, opened 2017-05-10
ElytronSubsystemMessages#credentialStoreIssueEncountered is never thr…

rhusar:
…own in CredentialStoreResource

@pskopek Please confirm these were meant to be messages, since the exception is never thrown in https://github.com/wildfly/wildfly-core/blob/master/elytron/src/main/java/org/wildfly/extension/elytron/CredentialStoreResource.java#L90 and other 3 places.


darranl:
These messages should only ever be used by the subsystem so if nothing in the subsystem is using them should be safe to remove.

PR #2422, opened 2017-05-12
WFCORE-2801 Upgrade jansi to 1.16

ctomc:


jamezp:
@jfdenise Please approve. The [jansi version aesh](https://github.com/aeshell/aesh/blob/0.66.15/pom.xml#L55-L59) 0.66.15 is using is 1.11 which matches the current version in WildFly Core.

jfdenise:
I have asked to make aesh to support jansi 1.16. I will comment there when aesh 0.66.16 has been released. You will then be able to update this PR with the right aesh release in sync with jansi 1.16.
Thank-you.
JF

jfdenise:
You can update your PR with a dependency on aesh 0.66.16

ctomc:
@jfdenise @jamezp updated PR so that it also updates aesh to 0.66.16

PR #2425, opened 2017-05-16
WFCORE-2756: Added periodic-rotating and size-rotating audit logs for Elytron audit

yersan:
Now there are two different rotating file logs in Elytron:

- periodic-rotating-file-audit-log: Which rotates based on a suffix in DateTimeFormatter format. Similar behaviour than /core-service=management/access=audit/periodic-rotating-file-handler. The suffix is mandatory here.
- size-rotating-file-audit-log: which rotates in size and can use a suffix in DateTimeFormatter format appended to each file rotation. Similar behaviour than /subsystem=logging/size-rotating-file-handler. The suffix is optional here.

Jira issues:
https://issues.jboss.org/browse/WFCORE-2756
https://issues.jboss.org/browse/JBEAP-10700

Depends on: https://github.com/wildfly-security/wildfly-elytron/pull/845

### DIFF
--- a/elytron/src/main/java/org/wildfly/extension/elytron/AuditResourceDefinitions.java
+++ b/elytron/src/main/java/org/wildfly/extension/elytron/AuditResourceDefinitions.java
@@ -21,7 +21,8 @@ import static org.wildfly.extension.elytron.Capabilities.SECURITY_EVENT_LISTENER
 import static org.wildfly.extension.elytron.Capabilities.SECURITY_EVENT_LISTENER_RUNTIME_CAPABILITY;
 import static org.wildfly.extension.elytron.Capabilities.SSL_CONTEXT_CAPABILITY;
 import static org.wildfly.extension.elytron.ElytronDescriptionConstants.FILE_AUDIT_LOG;
-import static org.wildfly.extension.elytron.ElytronDescriptionConstants.ROTATING_FILE_AUDIT_LOG;
+import static org.wildfly.extension.elytron.ElytronDescriptionConstants.PERIODIC_ROTATING_FILE_AUDIT_LOG;
+import static org.wildfly.extension.elytron.ElytronDescriptionConstants.SIZE_ROTATING_FILE_AUDIT_LOG;
 import static org.wildfly.extension.elytron.ElytronDescriptionConstants.SYSLOG_AUDIT_LOG;
 import static org.wildfly.extension.elytron.ElytronExtension.asStringIfDefined;
 import static org.wildfly.extension.elytron.FileAttributeDefinitions.RELATIVE_TO;
@@ -33,7 +34,8 @@ import java.io.File;
 import java.io.IOException;
 import java.net.InetAddress;
 import java.net.UnknownHostException;
-import java.text.SimpleDateFormat;
+import java.time.ZoneId;
+import java.time.format.DateTimeFormatter;
 import java.util.function.Function;
 import java.util.function.Supplier;
 
@@ -57,13 +59,16 @@ import org.jboss.msc.value.InjectedValue;
 import org.wildfly.extension.elytron.FileAttributeDefinitions.PathResolver;
 import org.wildfly.extension.elytron.TrivialService.ValueSupplier;
 import org.wildfly.extension.elytron.capabilities._private.SecurityEventListener;
+import org.wildfly.extension.elytron.validators.SizeValidator;
+import org.wildfly.extension.elytron.validators.SuffixValidator;
 import org.wildfly.security.audit.AuditEndpoint;
 import org.wildfly.security.audit.AuditLogger;
 import org.wildfly.security.audit.EventPriority;
 import org.wildfly.security.audit.FileAuditEndpoint;
 import org.wildfly.security.audit.JsonSecurityEventFormatter;
-import org.wildfly.security.audit.RotatingFileAuditEndpoint;
+import org.wildfly.security.audit.PeriodicRotatingFileAuditEndpoint;
 import org.wildfly.security.audit.SimpleSecurityEventFormatter;
+import org.wildfly.security.audit.SizeRotatingFileAuditEndpoint;
 import org.wildfly.security.audit.SyslogAuditEndpoint;
 import org.wildfly.security.auth.server.event.SecurityEventVisitor;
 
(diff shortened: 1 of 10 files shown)


elytron/src/main/java/org/wildfly/extension/elytron/AuditResourceDefinitions.java
```diff
@@ -240,14 +242,16 @@ static ResourceDefinition getRotatingFileAuditLogResourceDefinition() {
...
+                        FileAuditEndpoint.Builder builder = RotatingFileAuditEndpoint.builder()
                                 .setMaxBackupIndex(maxBackupIndex)
                                 .setRotateOnBoot(rotateOnBoot)
                                 .setRotateSize(rotateSize)
-                                .setSuffix(suffix)
                                 .setLocation(resolvedPath.toPath())
-                                .setSyncOnAccept(synv)
-                                .build();
+                                .setSyncOnAccept(synv);
+                        if ( !suffix.equals(UNDEFINED.asString()) ){
```
jamezp: This should be `!suffix.isDefined()`.

elytron/src/main/java/org/wildfly/extension/elytron/AuditResourceDefinitions.java
```diff
@@ -153,6 +153,8 @@
 
     static final String DATE_FORMAT = "yyyy-MM-dd HH:mm:ss";
 
+    private static final ModelNode UNDEFINED = new ModelNode();
```
jamezp: This should not be required and should be removed. See the comment below.

yersan:
@jamezp PR changed according with your review.

elytron/src/main/java/org/wildfly/extension/elytron/AuditResourceDefinitions.java
```diff
@@ -240,14 +240,16 @@ static ResourceDefinition getRotatingFileAuditLogResourceDefinition() {
                     final SecurityEventVisitor<?, String> formatter = Format.JSON == format ? JsonSecurityEventFormatter.builder().setDateFormatSupplier(bind(SimpleDateFormat::new, DATE_FORMAT)).build() : SimpleSecurityEventFormatter.builder().setDateFormatSupplier(bind(SimpleDateFormat::new, DATE_FORMAT)).build();
                     AuditEndpoint endpoint;
                     try {
-                        endpoint = RotatingFileAuditEndpoint.builder()
+                        FileAuditEndpoint.Builder builder = RotatingFileAuditEndpoint.builder()
```
jamezp: Sorry, I missed this one yesterday. Is there a reason the type isn't `RotationgFileAuditEndpoint.Builder`? It would save the casting needed if the `suffix` is defined.

yersan: Not really, the cast was necesary because the builder chain returns a FileAuditEndpoint, but since the order in the builder chain order doesn't matter, we can change it to avoid the cast.
I'm going to do it

yersan: @jamezp I have removed the cast

bstansberry:
Further to my last comment, rotate-size also allows rotation without a suffix.

A further issue though is 'suffix' appears to be doing double-duty. It is used to configure the suffix appended to file names when rotation happens, which is intuitive enough. But it also seems to be used to drive periodic file rotation, which IMHO is not at all obvious. At a minimum the xsd and management api attribute description need to document that behavior.

The logging subsystem uses 'suffix' for two things in the same way, so if that's the norm, that's ok, but the descriptions still need to be clear. I haven't looked closely though, but in the logging subsystem case the usage seems to be in resources called "periodic-xxx" so the periodic nature of the rotation could be assumed. Here it's for a general purpose "rotating-file-audit-log".

honza889:
Hi @yersan, I think you will have to prepare PR for wildfly-elytron too, to ensure that unfilled suffix will not cause NPE in RotatingFileAuditEndpoint.

jamezp:
I'm just putting a hold label on this as I know @yersan is working on some changes.

jamezp:
I haven't reviewed the changes here yet as this PR will require a new version of Elytron to be integrated once the Elytron changes are merged.

darranl:
@jamezp Can this one please be reviewed?  The Elytron change is ready to go in but as it does go in I would like to know the WFCORE change is also ready to go in.

elytron/src/main/java/org/wildfly/extension/elytron/AuditResourceDefinitions.java
```diff
@@ -207,8 +219,65 @@ static ResourceDefinition getFileAuditLogResourceDefinition() {
...
+                    File resolvedPath = pathResolver.resolve();
+
+                    final SecurityEventVisitor<?, String> formatter = Format.JSON == format ? JsonSecurityEventFormatter.builder().setDateFormatSupplier(bind(SimpleDateFormat::new, DATE_FORMAT)).build() : SimpleSecurityEventFormatter.builder().setDateFormatSupplier(bind(SimpleDateFormat::new, DATE_FORMAT)).build();
+                    AuditEndpoint endpoint;
+                    try {
+                        FileAuditEndpoint.Builder builder =  PeriodicRotatingFileAuditEndpoint.builder()
+                                .setSuffix(suffix)
+                                .setLocation(resolvedPath.toPath())
+                                .setSyncOnAccept(synv)
+                                .setDateFormatSupplier(bind(SimpleDateFormat::new, DATE_FORMAT));
```
jamezp: See my comment on https://github.com/wildfly-security/wildfly-elytron/pull/845#issuecomment-306260856. We may want this to use a `java.time.format.DateTimeFormatter`. We're using JSR 310 to format the dates so we should probably use the formatter from the `java.time` API.

elytron/src/main/java/org/wildfly/extension/elytron/AuditResourceDefinitions.java
```diff
@@ -244,14 +313,18 @@ static ResourceDefinition getRotatingFileAuditLogResourceDefinition() {
...
-                                .setSuffix(suffix)
-                                .setLocation(resolvedPath.toPath())
+                                .setRotateSize(rotateSize);
+                        if ( suffix.isDefined() ){
+                            builder.setSuffix(suffix.asString());
+                        }
+                        builder.setLocation(resolvedPath.toPath())
                                 .setSyncOnAccept(synv)
-                                .build();
+                                .setDateFormatSupplier(bind(SimpleDateFormat::new, DATE_FORMAT));
```
jamezp: Same comment about the `java.time` API.

jamezp:
@darranl Good call and done. There is a minor change I'd like to personally see.

darranl:
This PR does not add the synchronized attribute so I don't think it should be included in the review of this PR.

However the reason the name 'synchronized' was chosen was because this is not just about ensuring we automatically flush to disk, the handling of the security event is held until this has happened - in the future at the expense of reliability of the audited events we could even use this attribute to control if we allow an in memory backlog.

@hpehl / @claudio4j FYI this is a management model change for you to be aware of.

darranl:
Changes pulled into ladybird to be handled in a coordinated PR.

PR #2428, opened 2017-05-17
[WFCORE-2632][JBEAP-10169] remove service "org.wildfly.security.secur…

soul2zimate:
…ity-domain.ManagementDomain.initial" before its service recreation.

https://issues.jboss.org/browse/WFCORE-2632

Override applyUpdateToRuntime method to remove both services org.wildfly.security.security-domain.ManagementDomain and org.wildfly.security.security-domain.ManagementDomain.initial before recreating both of them.

more notes on https://issues.jboss.org/browse/JBEAP-10169


soul2zimate:
Thanks, Its' true. I have rebased it.

PR #2430, opened 2017-05-17
Do not attempt to start the server in the constructor. JUnit attempts…

jamezp:
… to create all tests before it runs any test. The failure to start the server is captured by JUnit, but all other tests are created before JUnit fails. Moving the startup out of the constructor allows the tests to fail fast if the startup of the server fails.

### DIFF
--- a/testsuite/test-runner/src/main/java/org/wildfly/core/testrunner/WildflyTestRunner.java
+++ b/testsuite/test-runner/src/main/java/org/wildfly/core/testrunner/WildflyTestRunner.java
@@ -8,6 +8,7 @@ import java.security.Security;
 import java.util.Collections;
 import java.util.LinkedList;
 import java.util.List;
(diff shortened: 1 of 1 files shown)


testsuite/test-runner/src/main/java/org/wildfly/core/testrunner/WildflyTestRunner.java
```diff
@@ -94,7 +94,7 @@ public void testRunFinished(Result result) throws Exception {
                 }
             }
         });
-        startServerIfRequired();
+        configureTestRun();
         if (!serverSetupTasks.isEmpty() && !automaticServerControl) {
```
bstansberry: Is this handled correctly now?

bstansberry: Never mind; I just need to not be a lazy ass and look.

PR #2432, opened 2017-05-18
[WFCORE-2828]: org.wildfly.core.test.standalone.mgmt.events.ProcessStateListenerTestCase from wf-core fails intermittently

ehsavoie:
Executing stopping of the server out of the management-thread for a reload.
Preventing the ProcessStateListenerService from stopping while it is notifying its listener.

Cleaning the code :
 * making the test code more usable
 * using capability name for dependencies

Jira: https://issues.jboss.org/browse/WFCORE-2828
JBEAP: https://issues.jboss.org/browse/JBEAP-10998

### DIFF
--- a/core-management/core-management-subsystem/src/main/java/org/wildfly/extension/core/management/ProcessStateListenerResourceDefinition.java
+++ b/core-management/core-management-subsystem/src/main/java/org/wildfly/extension/core/management/ProcessStateListenerResourceDefinition.java
@@ -23,9 +23,11 @@ package org.wildfly.extension.core.management;
 
 
 
+import static org.jboss.as.controller.AbstractControllerService.EXECUTOR_CAPABILITY;
 import static org.jboss.as.controller.descriptions.ModelDescriptionConstants.MODULE;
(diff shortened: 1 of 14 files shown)


core-management/core-management-subsystem/src/main/java/org/wildfly/extension/core/management/ProcessStateListenerResourceDefinition.java
```diff
@@ -57,7 +58,9 @@
     private static final String CLASS = "class";
     private static final String PROCESS_STATE_LISTENER_CAPABILITY_NAME = "org.wildfly.extension.core-management.process-state";
     static final RuntimeCapability<Void> PROCESS_STATE_LISTENER_CAPABILITY =
-            RuntimeCapability.Builder.of(PROCESS_STATE_LISTENER_CAPABILITY_NAME, true, ProcessStateListenerService.class).build();
+            RuntimeCapability.Builder.of(PROCESS_STATE_LISTENER_CAPABILITY_NAME, true, ProcessStateListenerService.class)
```
bstansberry: I believe this should be Void.class

It's the value type of the Service, not the implementation class of the Service.

PR #2439, opened 2017-05-19
Revert the WFCORE-2801 jansi upgrade as there are problems productizi…

bstansberry:
…ng it

Just an experiment for now.


bstansberry:
Experiment succeeded; closing as likely if we have to revert, aesh would too and there'd be a related aesh version change.

PR #2444, opened 2017-05-19
WFCORE-2768: Read aliases operation should use read only credential store service

yersan:
Although the issue is talking about concurrency problems adding aliases, those problems have been fixed with the work done in https://issues.jboss.org/browse/WFCORE-2777. Together with that issue, credential store service is being correctly retrieved using a lock at the operation level. 

The changes in this commit allow the read_aliases operation to work without acquiring the lock retrieving the credential store service because it is a read-only operation.

Jira issues:
https://issues.jboss.org/browse/WFCORE-2768
https://issues.jboss.org/browse/JBEAP-10749


yersan:
Retest this please

PR #2448, opened 2017-05-22
[WFCORE-2851] Stabilise CLIExecutor

mchoma:
1. Use standard Process.waitFor() instead of custom polling mechanism
2. Use standard CLI option --error-on-interact instead of custom
answering "Accept Certificate?" prompt
3. Always log CLI output

### DIFF
--- a/testsuite/shared/src/main/java/org/jboss/as/test/integration/management/util/CustomCLIExecutor.java
+++ b/testsuite/shared/src/main/java/org/jboss/as/test/integration/management/util/CustomCLIExecutor.java
@@ -38,7 +38,10 @@ import java.util.Arrays;
 import java.util.Collections;
 import java.util.List;
 import java.util.Map;
+import java.util.concurrent.TimeUnit;
 import java.util.regex.Pattern;
+
+import org.junit.Assert;
 
 import org.jboss.as.test.shared.TestSuiteEnvironment;
 import org.jboss.as.test.shared.TimeoutUtil;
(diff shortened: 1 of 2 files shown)


testsuite/shared/src/main/java/org/jboss/as/test/integration/management/util/CustomCLIExecutor.java
```diff
@@ -255,34 +243,16 @@ private static String executeProcess(final CliCommandBuilder commandBuilder, fin
...
-                }
             }
-            if (runningTime >= CLI_PROC_TIMEOUT) {
-                cliProc.destroy();
-                wait = false;
-                LOGGER.info("A timeout has occurred while invoking CLI command.");
-            }
-        } while (wait);
+        } catch (InterruptedException e1) {
+            new RuntimeException("Interupted while waiting", e1);
```
darranl: Is this missing a throw?

bstansberry:
@jfdenise Please approve

testsuite/shared/src/main/java/org/jboss/as/test/integration/management/util/CustomCLIExecutor.java
```diff
@@ -255,34 +243,16 @@ private static String executeProcess(final CliCommandBuilder commandBuilder, fin
...
         int exitCode = Integer.MIN_VALUE;
-        do {
-            try {
-                Thread.sleep(STATUS_CHECK_INTERVAL);
-            } catch (InterruptedException e) {
-            }
-            runningTime += STATUS_CHECK_INTERVAL;
-            try {
+        try {
+            boolean finished = cliProc.waitFor(CLI_PROC_TIMEOUT, TimeUnit.MILLISECONDS);
```
jfdenise: In the previous implementation, in case of timeout we had the trace LOGGER.info("A timeout has occurred while invoking CLI command."); If finished is false, you should resurrect this trace.

testsuite/shared/src/main/java/org/jboss/as/test/integration/management/util/CustomCLIExecutor.java
```diff
@@ -255,34 +243,16 @@ private static String executeProcess(final CliCommandBuilder commandBuilder, fin
...
-            if (runningTime >= CLI_PROC_TIMEOUT) {
-                cliProc.destroy();
-                wait = false;
-                LOGGER.info("A timeout has occurred while invoking CLI command.");
-            }
-        } while (wait);
+        } catch (InterruptedException e1) {
+            new RuntimeException("Interupted while waiting", e1);
+        } finally {
+            cliProc.destroy();
```
jamezp: To be consistent with https://github.com/wildfly/wildfly-core/pull/2451 we should probably make this `cliProc.destroyForcibly()`.

testsuite/shared/src/main/java/org/jboss/as/test/integration/management/util/CustomCLIExecutor.java
jamezp: Even better it should probably use `Assert.fail()` since it's in the test suite.

PR #2456, opened 2017-05-23
WFCORE-2722: Add OTP support for identity set-password operation using filesystem-realm

yersan:
Added the possibility to set OTP configurations for an identity using the file system realm

Jira Issues:
https://issues.jboss.org/browse/WFCORE-2722
https://issues.jboss.org/browse/JBEAP-10532

### DIFF
--- a/elytron/src/main/java/org/wildfly/extension/elytron/ModifiableRealmDecorator.java
+++ b/elytron/src/main/java/org/wildfly/extension/elytron/ModifiableRealmDecorator.java
@@ -53,15 +53,18 @@ import org.wildfly.security.password.PasswordFactory;
 import org.wildfly.security.password.interfaces.BCryptPassword;
 import org.wildfly.security.password.interfaces.ClearPassword;
 import org.wildfly.security.password.interfaces.DigestPassword;
+import org.wildfly.security.password.interfaces.OneTimePassword;
 import org.wildfly.security.password.interfaces.SaltedSimpleDigestPassword;
 import org.wildfly.security.password.interfaces.SimpleDigestPassword;
 import org.wildfly.security.password.spec.ClearPasswordSpec;
 import org.wildfly.security.password.spec.DigestPasswordAlgorithmSpec;
 import org.wildfly.security.password.spec.EncryptablePasswordSpec;
 import org.wildfly.security.password.spec.IteratedSaltedPasswordAlgorithmSpec;
+import org.wildfly.security.password.spec.OneTimePasswordSpec;
 import org.wildfly.security.password.spec.PasswordSpec;
 import org.wildfly.security.password.spec.SaltedPasswordAlgorithmSpec;
 
+import java.nio.charset.StandardCharsets;
 import java.security.NoSuchAlgorithmException;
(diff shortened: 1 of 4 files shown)


elytron/src/main/java/org/wildfly/extension/elytron/ModifiableRealmDecorator.java
```diff
@@ -503,31 +537,42 @@ protected void executeRuntimeStep(final OperationContext context, final ModelNod
...
             final String algorithm;
 
             if (passwordType.equals(ElytronDescriptionConstants.BCRYPT)) {
+                password = SetPasswordHandler.Bcrypt.PASSWORD.resolveModelAttribute(parentContext, passwordNode).asString();
                 byte[] salt = SetPasswordHandler.Bcrypt.SALT.resolveModelAttribute(parentContext, passwordNode).asBytes();
                 int iterationCount = SetPasswordHandler.Bcrypt.ITERATION_COUNT.resolveModelAttribute(parentContext, passwordNode).asInt();
                 passwordSpec = new EncryptablePasswordSpec(password.toCharArray(), new IteratedSaltedPasswordAlgorithmSpec(iterationCount, salt));
                 algorithm = SetPasswordHandler.Bcrypt.ALGORITHM.resolveModelAttribute(parentContext, passwordNode).asString();
             } else if (passwordType.equals(ElytronDescriptionConstants.CLEAR)) {
+                password = SetPasswordHandler.Bcrypt.PASSWORD.resolveModelAttribute(parentContext, passwordNode).asString();
```
yersan: You're right Brian; this has been a complete oversight on my part. Right now should be moved to SetPasswordHandler level because OTP has no password attribute. We can compose them extending from a base class but maybe just for a single attribute heredity, maybe it doesn't make sense, WDYT? Anyway there is a slight difference between password definitions on each attribute, Bcrypt, Clear and SimpleDigest have a minimum size defined, SaltedSimpleDigest and Digest don't. Which has no so much sense to me, but I will clear it up.

yersan: About appending -from, it was verified with @honza889. -from are derived values used in LDAP, OTP will use direct values so attributes without -from are fine here.
About the minimum size, all of the password types aren't optional, we can remove the minimum size in Bcrypt, Clear and SimpleDigest

bstansberry: From the comments above, I'm not clear, Yeray, on whether the various PASSWORD attributes can all have the same settings or not. If they are all the same I think it's fine to declare the constant at the SetPasswordHandler level even though OTP doesn't use it. It's just a reasonable place for the constant. But if you don't like that it's fine to just have duplicate PASSWORD attributes for Bcrypt, Clear etc. Just use the related one in each of the if/else if blocks here so it's not confusing.

yersan:
@bstansberry commit was updated with changes requested, I think now cover your points, param names were discussed on Elytron room.

bstansberry:
retest this please

PR #2465, opened 2017-05-24
[WFCORE-2236]: Reload required while enabling HTTPS in domain mode.

ehsavoie:
Propagating the security realm operations to the running servers on a host.

Jiira: https://issues.jboss.org/browse/WFCORE-2236
JBEAP: https://issues.jboss.org/browse/JBEAP-5963

### DIFF
--- a/host-controller/src/main/java/org/jboss/as/domain/controller/operations/coordination/ServerOperationResolver.java
+++ b/host-controller/src/main/java/org/jboss/as/domain/controller/operations/coordination/ServerOperationResolver.java
@@ -77,6 +77,7 @@ import static org.jboss.as.controller.descriptions.ModelDescriptionConstants.MAN
 import static org.jboss.as.controller.descriptions.ModelDescriptionConstants.REDEPLOY;
 import static org.jboss.as.controller.descriptions.ModelDescriptionConstants.REMOVE;
 import static org.jboss.as.controller.descriptions.ModelDescriptionConstants.REPLACE_DEPLOYMENT;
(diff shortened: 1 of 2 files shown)


bstansberry:
retest this please

host-controller/src/main/java/org/jboss/as/domain/controller/operations/coordination/ServerOperationResolver.java
```diff
@@ -1003,56 +1004,49 @@ private ModelNode getSystemPropertyValue(ModelNode root, String propName) {
...
-                        } else {
-                            PathElement myElement = element;
-                            if (myElement.getKey().equals(SERVER_LOGGER)) {
-                                //server-logger=audit-log gets sent to the servers as logger=>audit-log
-                                myElement = PathElement.pathElement(LOGGER, element.getValue());
+        if (MANAGEMENT.equals(address.getElement(0).getValue()) && address.size() >= 2) {
+            ModelNode op = operation.clone();
+            switch (address.getElement(1).getKey()) {
+                case ACCESS:
+                case AUDIT:
```
bstansberry: "audit" is not a valid key here.

bstansberry: Right now "audit" is the only value for key "access" in this location, but to protect against future changes you should put all this in a if (AUDIT.equals(address.getElement(1).getValue()) block.

PR #2466, opened 2017-05-24
[WFCORE-2867] Credential store reload operation doesn't have effect on dependent resources

pskopek:
https://issues.jboss.org/browse/WFCORE-2867
https://issues.jboss.org/browse/JBEAP-10650

### DIFF
--- a/controller/src/main/java/org/jboss/as/controller/security/CredentialReference.java
+++ b/controller/src/main/java/org/jboss/as/controller/security/CredentialReference.java
@@ -36,9 +36,11 @@ import org.jboss.as.controller.capability.RuntimeCapability;
 import org.jboss.as.controller.logging.ControllerLogger;
 import org.jboss.dmr.ModelNode;
 import org.jboss.dmr.ModelType;
+import org.jboss.msc.service.Service;
 import org.jboss.msc.service.ServiceBuilder;
+import org.jboss.msc.service.ServiceController;
 import org.jboss.msc.service.ServiceName;
-import org.jboss.msc.value.InjectedValue;
+import org.jboss.msc.service.ServiceRegistry;
 import org.wildfly.common.function.ExceptionSupplier;
 import org.wildfly.security.auth.SupportLevel;
 import org.wildfly.security.credential.Credential;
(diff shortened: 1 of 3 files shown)


pskopek:
retest this please

controller/src/main/java/org/jboss/as/controller/security/CredentialReference.java
```diff
@@ -311,12 +312,16 @@ public static String credentialReferencePartAsStringIfDefined(ModelNode credenti
...
-        final InjectedValue<CredentialStore> credentialStoreInjectedValue = new InjectedValue<>();
+        final Service<CredentialStore> credentialStoreService;
         if (credentialAlias != null) {
             // use credential store service
             String credentialStoreCapabilityName = RuntimeCapability.buildDynamicCapabilityName(CREDENTIAL_STORE_CAPABILITY, credentialStoreName);
             ServiceName credentialStoreServiceName = context.getCapabilityServiceName(credentialStoreCapabilityName, CredentialStore.class);
-            serviceBuilder.addDependency(ServiceBuilder.DependencyType.REQUIRED, credentialStoreServiceName, CredentialStore.class, credentialStoreInjectedValue);
+            serviceBuilder.addDependency(ServiceBuilder.DependencyType.REQUIRED, credentialStoreServiceName);
+            ServiceController<?> controller = context.getServiceRegistry(false).getService(credentialStoreServiceName);
+            credentialStoreService = controller != null ? (Service<CredentialStore>)controller.getService() : null;
```
bstansberry: This is a race. There is no guarantee that when L321 executes the credential store service will be installed in MSC. Subsystems execute their operations in parallel during boot so subsystem A can in no way rely on subystem B having executed its ops first. And even without parallel boot there its totally valid for users to place the elytron subsystem last in their config.

elytron/src/main/java/org/wildfly/extension/elytron/CredentialStoreService.java
```diff
@@ -112,8 +113,12 @@ public void start(StartContext startContext) throws StartException {
...
         try {
             credentialStoreAttributes.put(CS_LOCATION_ATTRIBUTE, loc.toAbsolutePath().toString());
-            credentialStore = getCredentialStoreInstance();
-            credentialStore.initialize(credentialStoreAttributes, resolveCredentialStoreProtectionParameter(), otherProviders.getOptionalValue());
+            CredentialStore cs = getCredentialStoreInstance();
+            cs.initialize(credentialStoreAttributes, resolveCredentialStoreProtectionParameter(), otherProviders.getOptionalValue());
+            if ( credentialStoreAttributes.get(ElytronDescriptionConstants.CREATE).equals("true") && !loc.toFile().exists() ){
+                cs.flush();
+            }
+            credentialStore.set(cs);
```
bstansberry: In stop() should there be a credentialStore.set(null)?

pskopek: Yes @bstansberry , you are right. Pushing update.

PR #2467, opened 2017-05-24
[WFCORE-2792]: AttributeDefinition.parser is not consistently used in subclasses.

ehsavoie:
* fixing parsing
* fixing typo

Jira: https://issues.jboss.org/browse/WFCORE-2792

### DIFF
--- a/elytron/src/main/java/org/wildfly/extension/elytron/ElytronDefinition.java
+++ b/elytron/src/main/java/org/wildfly/extension/elytron/ElytronDefinition.java
@@ -36,6 +36,8 @@ import java.security.Provider;
 import java.util.List;
 
 import org.jboss.as.controller.AbstractBoottimeAddStepHandler;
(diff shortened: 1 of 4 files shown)


elytron/src/main/java/org/wildfly/extension/elytron/ElytronDefinition.java
```diff
@@ -106,6 +108,8 @@
 
     static final StringListAttributeDefinition DISALLOWED_PROVIDERS = new StringListAttributeDefinition.Builder(ElytronDescriptionConstants.DISALLOWED_PROVIDERS)
             .setRequired(false)
+            .setAttributeParser(AttributeParser.COMMA_DELIMITED_STRING_LIST)
```
ctomc: this is wrong COMMA_DELIMITED was wrong default in the past.
it should be blank space, same as defined by xs:list.
So new attributes shouldn't use comma delimited list attributes.

PR #2475, opened 2017-05-25
WFCORE-2784 Host controller should handle jdk9 specific jvm parameters

ctomc:
+ update domain config for jdk9


ctomc:
jdk9b170 run https://ci.wildfly.org/viewLog.html?buildId=61350&tab=buildResultsDiv&buildTypeId=WildFlyCore_MasterLinuxJdk9

PR #2476, opened 2017-05-26
[WFCORE-2877] users loading for add-user script fix

honza889:
https://issues.jboss.org/browse/WFCORE-2877

### DIFF
--- a/domain-management/src/main/java/org/jboss/as/domain/management/security/UserPropertiesFileLoader.java
+++ b/domain-management/src/main/java/org/jboss/as/domain/management/security/UserPropertiesFileLoader.java
@@ -31,6 +31,8 @@ import java.util.ArrayList;
 import java.util.List;
 import java.util.regex.Matcher;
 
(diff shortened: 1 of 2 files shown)


bstansberry:
@darranl please approve

domain-management/src/main/java/org/jboss/as/domain/management/security/UserPropertiesFileLoader.java
```diff
@@ -101,7 +101,7 @@ protected void load() throws IOException {
         try {
             br = Files.newBufferedReader(propertiesFile.toPath(), StandardCharsets.UTF_8);
             String currentLine;
-            while (realmName == null && (currentLine = br.readLine()) != null) {
+            while ((currentLine = br.readLine()) != null) {
```
bstansberry: With this it's now possible that more than one realm comment will be parsed. That should probably be handled, either with an error or by ignoring subsequent realm comments.

honza889: Updated to throw IOException.

PR #2477, opened 2017-05-26
Fix for WFCORE-2878, CLI complete empty complex object

jfdenise:
Handle special case of object being complete and lastStateChar being a separator.
Added unit tests.


bstansberry:
@jfdenise I'm not sure if the failure is related, although I doubt it, given WFCORE-2679. But please confirm.

jfdenise:
@bstansberry , that is another symptom of WFCORE-2679, unrelated to this change.

bstansberry:
Thanks!

PR #2478, opened 2017-05-26
WFCORE-2836 - better version of verifying the slave has actually deregistered before continuing tests

luck3y:
I've used a JUnit @Rule to run this multiple times in succession both locally and in CI. Seems better now.

### DIFF
--- a/testsuite/domain/src/test/java/org/jboss/as/test/integration/domain/cacheddc/CachedDcDomainTestCase.java
+++ b/testsuite/domain/src/test/java/org/jboss/as/test/integration/domain/cacheddc/CachedDcDomainTestCase.java
@@ -19,12 +19,21 @@ package org.jboss.as.test.integration.domain.cacheddc;
 
 import static org.jboss.as.controller.client.helpers.ClientConstants.OUTCOME;
 import static org.jboss.as.controller.client.helpers.ClientConstants.SUCCESS;
+import static org.jboss.as.controller.descriptions.ModelDescriptionConstants.FAILED;
+import static org.jboss.as.controller.descriptions.ModelDescriptionConstants.FAILURE_DESCRIPTION;
 import static org.jboss.as.controller.descriptions.ModelDescriptionConstants.HOST;
+import static org.jboss.as.controller.descriptions.ModelDescriptionConstants.HOST_STATE;
 import static org.jboss.as.controller.descriptions.ModelDescriptionConstants.LOGGER;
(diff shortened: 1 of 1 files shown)


testsuite/shared/src/main/java/org/jboss/as/test/integration/domain/management/util/DomainLifecycleUtil.java
```diff
@@ -575,6 +585,36 @@ public boolean isHostControllerStarted() {
...
+     */
+    public boolean isHostPresentInModel(final String hostname, final ControlledProcessState.State requiredState) throws Exception {
+        final ModelNode operation = new ModelNode();
+        operation.get(OP).set(READ_ATTRIBUTE_OPERATION);
+        operation.get(OP_ADDR).add(HOST, hostname);
+        operation.get(NAME).set(HOST_STATE);
+
+        ModelNode result;
+        try {
+            result = executeForResult(operation);
```
bstansberry: Don't use executeForResult here. It just returns the value of {"result"=> ???} not the full response. As is this method will never return true. You're analyzing the full response here, so just do getDomainClient().execute(operation) and analyze that.

Hopefully with that the exception handling can be simpler; i.e. no catching Throwable, throwing Exception, converting Exception into InterruptedException.

luck3y: Ah, yes. Thanks!

luck3y:
@bstansberry I've changed this as suggested. I think I may be being a bit too paranoid with the throwing RuntimeException() at the end rather than just returning false, but I was concerned this could potentially hide other failures. WDYT?

I'll set up a CI repeated CI run separately and see if I can get a failure.

bstansberry:
The RuntimeException is probably ok. If it proves to be a problem it's easy enough to change.

luck3y:
This just failed on windows, with: 

java.lang.RuntimeException: {
    "outcome" : "failed",
    "result" : null,
    "failure-description" : "Writes closed",
    "rolled-back" : true
}

So I've updated the check to return false on resource not found, but true on any other failure state, with the assumption that the host is present, but returned an error etc.

luck3y:
retest this please

luck3y:
I think I'm going to take that method out of DomainLifecycleUtils and just keep it in the CachedDc test.

luck3y:
retest this please

luck3y:
retest this please

PR #2479, opened 2017-05-29
WFCORE-2483 Extend HTTPSManagementInterfaceTestCase tests with usage of CR

mchoma:
Controversy of this PR is reload is needed during each test method. This is because reload is needed to change security realm on management interface.

Reload optimalization can be performed, if it would be possible to define order of test methods. But as this is not possible I have to ensure correct security realm is configured before each test method.

Locally this test takes 1.5 sec

### DIFF
--- a/testsuite/manualmode/src/test/java/org/wildfly/core/test/standalone/mgmt/HTTPSManagementInterfaceTestCase.java
+++ b/testsuite/manualmode/src/test/java/org/wildfly/core/test/standalone/mgmt/HTTPSManagementInterfaceTestCase.java
@@ -20,6 +20,10 @@ package org.wildfly.core.test.standalone.mgmt;
 import static org.hamcrest.CoreMatchers.containsString;
 import static org.hamcrest.CoreMatchers.not;
 import static org.jboss.as.controller.descriptions.ModelDescriptionConstants.NAME;
+import static org.jboss.as.controller.descriptions.ModelDescriptionConstants.OUTCOME;
+import static org.jboss.as.controller.descriptions.ModelDescriptionConstants.READ_ATTRIBUTE_OPERATION;
+import static org.jboss.as.controller.descriptions.ModelDescriptionConstants.RESULT;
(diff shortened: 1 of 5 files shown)


testsuite/manualmode/src/test/java/org/wildfly/core/test/standalone/mgmt/HTTPSManagementInterfaceTestCase.java
```diff
@@ -265,6 +317,19 @@ private static void serverSetup(ManagementClient managementClient) throws Except
...
+    private static void changeSecurityRealmForHttpMngmntInterface(String securityRealmName) throws Exception {
+        //ModelControllerClient client = controller.getClient().getControllerClient();
+        try (ModelControllerClient client = getNativeModelControllerClient()) {
+            ModelNode operation = createOpNode("core-service=management/management-interface=http-interface",
+                    ModelDescriptionConstants.WRITE_ATTRIBUTE_OPERATION);
+            operation.get(NAME).set("security-realm");
+            operation.get(VALUE).set(securityRealmName);
+            CoreUtils.applyUpdate(operation, client);
+
+            reloadServer();
```
bstansberry: There's no need for the write-attribute or reload if a read-attribute shows the attribute already has the desired value.

mchoma: I have added that optimalization and I get locally to 1.2 sec

testsuite/shared/src/main/java/org/jboss/as/test/integration/security/common/config/realm/CredentialReference.java
```diff
@@ -0,0 +1,98 @@
+package org.jboss.as.test.integration.security.common.config.realm;
```
jamezp: This is missing a copyright header.

bstansberry:
retest this please

PR #2481, opened 2017-05-29
[WFCORE-2881] SASL mechanism tests

kwart:
* https://issues.jboss.org/browse/WFCORE-2881
* https://issues.jboss.org/browse/JBEAP-11194

Elytron SASL mechanism tests (EAP7-530) which use `ModelControllerClient` to test Elytron+Remoting behavior.

The tests are added to the new testsuite module `elytron` which always have the Elytron configured as the default security provider. Most of the logic (test and configuration) is located in the `AbstractMgmtSaslTestBase` class.


luck3y:
@kwart These tests only run with -Delytron enabled right? Do any of them have legacy behavior that should be tested if -Delytron isn't used?

Otherwise, seems fine to me.

kwart:
These tests runs always with Elytron enabled (therefor they are in the new module).

luck3y:
@kwart I think for the shared elytron stuff, you probably should move to the elytron testsuite. Is this actually shared with other things?

kwart:
If the configuration utils are in the shared stuff, they can be reused in other core TS modules and in full. Part of them already exist in the `testsuite/shared` in wildfly. I've picked just the ones really used by the SASL tests in this PR.

Nevertheless, If I move these utils into elytron module in core, then it could be simpler to update/improve/extend the utils implementations in full wildfly, where more modules use them. But it also adds code duplications (i.e. more places to deal with during changing/fixing).

Maybe the best option would be following:

1) Leave only the following classes in core's `testsuite/shared`:
* `ConfigurableElement` - interface for providing configuration changes (has `create`, `remove` and `getName` methods), and `AbstractConfigurableElement` which just forces the child classes to provide an element name.
* `AbstractConfigSetupTask` - abstract parent for ServerSetupTasks working with arrays of `ConfigurableElement`s. Children implementations are for WildFly Test Runner (in core) and Arquillian (in full)
* `TestRunnerConfigSetupTask` - WildFly Test Runner implementation of setup task working with `ConfigurableElement`s
* `ModelNodeUtil` - static methods for simpler work with `ModelNode`s

2) Move all the `ConfigurableElement` implementations into the new `elytron` package.

3) (after the core upgrade in full) - Remove classes listed in 1) from full wildfly. Put the new ConfigurableElement implementations also into `testsuite/shared` in full.  Change the `AbstractElytronSetupTask` (already in full) just to extend the `AbstractConfigSetupTask` from this PR.

@luck3y WDYT?

luck3y:
I think we're OK with duplication, but would rather avoid having dependencies on core's shared testsuite stuff in full, it ends up leading to commits needing to be orchestrated at the same time etc.

kwart:
In that case, I'll move everything to the new `elytron` module.

jamezp:
While I understand the desire to have utilities to generate commands or operations for tests it makes it very difficult to understand what ops are actually being executed. If you need to configure a server to duplicate an issue it's hard to figure out what the configuration is. Just a personal opinion from experience :)

testsuite/elytron/src/test/java/org/wildfly/test/integration/elytron/sasl/mgmt/AbstractMgmtSaslTestBase.java
```diff
@@ -0,0 +1,346 @@
...
+
+    protected static final String USERNAME = "guest";
+
+    // set-password algorithm names, used also as user names for testing given algorithm
+    protected static final String DIGEST_ALGORITHM_MD5 = "digest-md5";
+    protected static final String DIGEST_ALGORITHM_SHA = "digest-sha";
+    protected static final String DIGEST_ALGORITHM_SHA256 = "digest-sha-256";
+    protected static final String DIGEST_ALGORITHM_SHA512 = "digest-sha-512";
+
+    protected static final int CONNECTION_TIMEOUT_IN_MS = 6 * 1000;
```
jamezp: We should probably use the `TimeoutUtil.adjust()` here.

testsuite/elytron/src/test/java/org/wildfly/test/integration/elytron/sasl/mgmt/AbstractMgmtSaslTestBase.java
```diff
@@ -0,0 +1,346 @@
...
+                                .withRealmMapper("JWT").build())
+                        .addMechanismConfiguration(
+                                MechanismConfiguration.builder().withMechanismName(saslMechanismName)
+                                        .addMechanismRealmConfiguration(
+                                                MechanismRealmConfiguration.builder().withRealmName(NAME).build())
+                                        .build())
+                        .build());
+
+        elements.add(SimpleSocketBinding.builder().withName(NAME).withPort(PORT_NATIVE).build());
+        elements.add(SimpleMgmtNativeInterface.builder().withSocketBinding(NAME).withSaslAuthenticationFactory(NAME).build());
```
jamezp: This configuration section, to me, shows my point about making the configuration confusing. From reading the test I've got no idea what configuration we end up with here.

testsuite/shared/src/main/java/org/jboss/as/test/integration/security/common/CoreUtils.java
```diff
@@ -631,4 +631,19 @@ public static String encodeQueryParam(final String paramName, final String param
...
+     *
+     * @param prefix folder name prefix
+     * @return created folder
+     */
+    public static File createTemporaryFolder(String prefix) throws IOException {
+        File file = File.createTempFile(prefix, "", null);
+        LOGGER.debugv("Creating temporary folder {0}", file);
+        file.delete();
+        file.mkdir();
+        return file;
```
jamezp: You can achieve the same thing here with `Files.createTempDirectory()` then there is no need to do all the extra deletes and creations.

testsuite/elytron/src/test/java/org/wildfly/test/security/common/AbstractConfigSetupTask.java
```diff
@@ -0,0 +1,93 @@
...
+        configurableElements = getConfigurableElements();
+
+        if (configurableElements == null || configurableElements.length == 0) {
+            LOGGER.warn("Empty Elytron configuration.");
+            return;
+        }
+
+        try (CLIWrapper cli = new CLIWrapper(true)) {
+            for (final ConfigurableElement configurableElement : configurableElements) {
+                LOGGER.infov("Adding element {0} ({1})", configurableElement.getName(),
```
jamezp: Maybe this should be `debug` so we don't use up too much space on CI with logs. Not a huge deal though.

testsuite/elytron/src/test/java/org/wildfly/test/security/common/AbstractConfigSetupTask.java
```diff
@@ -0,0 +1,93 @@
...
+            LOGGER.warn("Empty Elytron configuration.");
+            return;
+        }
+
+        try (CLIWrapper cli = new CLIWrapper(true)) {
+            final ListIterator<ConfigurableElement> reverseConfigIt = Arrays.asList(configurableElements)
+                    .listIterator(configurableElements.length);
+            while (reverseConfigIt.hasPrevious()) {
+                final ConfigurableElement configurableElement = reverseConfigIt.previous();
+                LOGGER.infov("Removing element {0} ({1})", configurableElement.getName(),
```
jamezp: Same note as above. Again though I don't think it's a huge deal.

testsuite/elytron/src/test/java/org/wildfly/test/security/common/elytron/CliFragment.java
```diff
@@ -0,0 +1,39 @@
...
+ * @author Josef Cacek
+ */
+public interface CliFragment {
+
+    /**
+     * Generates part of CLI string which uses configuration for this fragment.
+     *
+     * @return part of CLI command
+     */
+    String asString();
```
jamezp: Why not just use `toString()`?

testsuite/elytron/src/test/java/org/wildfly/test/security/common/elytron/ConfigurableElement.java
```diff
@@ -0,0 +1,71 @@
...
+     */
+    String getName();
+
+    /**
+     * Creates this element in domain model and also creates other resources if needed (e.g. external files)
+     *
+     * @param cli connected {@link CLIWrapper} instance
+     */
+    default void create(CLIWrapper cli) throws Exception {
+        throw new IllegalStateException("The create() method was not properly implemented");
```
jamezp: Having this be a `default` method to me indicates it should work with the default, however it looks required that a subtype would have to implement this.

testsuite/elytron/src/test/java/org/wildfly/test/security/common/elytron/ConfigurableElement.java
```diff
@@ -0,0 +1,71 @@
...
+     */
+    default void create(ModelControllerClient client, CLIWrapper cli) throws Exception {
+        create(cli);
+    }
+
+    /**
+     * Reverts the changes introdued by {@code create(...)} method(s).
+     */
+    default void remove(CLIWrapper cli) throws Exception {
+        throw new IllegalStateException("The remove() method was not properly implemented");
```
jamezp: Same note as above.

testsuite/elytron/src/test/java/org/wildfly/test/security/common/elytron/Path.java
```diff
@@ -0,0 +1,102 @@
...
+
+import static org.apache.commons.lang3.StringUtils.isNotBlank;
+import static org.jboss.as.test.shared.CliUtils.escapePath;
+
+/**
+ * Helper class for adding "path" and "relative-to" attributes into CLI commands.
+ *
+ * @author Josef Cacek
+ */
+public class Path implements CliFragment {
```
jamezp: I think we should rename this to something like `CliPath` only for the fact that some tests that end up using this *may* also want to use a `java.nio.file.Path`.

testsuite/elytron/src/test/java/org/wildfly/test/security/common/elytron/KeyManager.java
```diff
@@ -0,0 +1,32 @@
...
+package org.wildfly.test.security.common.elytron;
+
+/**
+ * Interface representing Elytron key-manager.
+ *
+ * @author Josef Cacek
+ */
+public interface KeyManager extends ConfigurableElement {
+
+}
```
jamezp: Any reason for this since it doesn't add any behavior?

testsuite/elytron/src/test/java/org/wildfly/test/security/common/elytron/KeyStore.java
```diff
@@ -0,0 +1,32 @@
...
+package org.wildfly.test.security.common.elytron;
+
+/**
+ * Interface representing Elytron key-stores.
+ *
+ * @author Josef Cacek
+ */
+public interface KeyStore extends ConfigurableElement {
+
+}
```
jamezp: Any reason for this since it doesn't add any behavior?

testsuite/elytron/src/test/java/org/wildfly/test/security/common/elytron/RoleMapper.java
```diff
@@ -0,0 +1,32 @@
...
+package org.wildfly.test.security.common.elytron;
+
+/**
+ * Marking interface for configuration classes for Elytron elements providing role-mapper capability.
+ *
+ * @author Josef Cacek
+ */
+public interface RoleMapper extends ConfigurableElement {
+
+}
```
jamezp: Any reason for this since it doesn't add any behavior?

testsuite/elytron/src/test/java/org/wildfly/test/security/common/elytron/SaslAuthenticationFactory.java
```diff
@@ -0,0 +1,32 @@
...
+package org.wildfly.test.security.common.elytron;
+
+/**
+ * Interface representing Elytron sasl-authentication-factory capability.
+ *
+ * @author Josef Cacek
+ */
+public interface SaslAuthenticationFactory extends ConfigurableElement {
+
+}
```
jamezp: Any reason for this since it doesn't add any behavior?

testsuite/elytron/src/test/java/org/wildfly/test/security/common/elytron/SaslServerFactory.java
```diff
@@ -0,0 +1,32 @@
...
+package org.wildfly.test.security.common.elytron;
+
+/**
+ * Interface representing Elytron sasl-server-factory capability.
+ *
+ * @author Josef Cacek
+ */
+public interface SaslServerFactory extends ConfigurableElement {
+
+}
```
jamezp: Any reason for this since it doesn't add any behavior?

testsuite/elytron/src/test/java/org/wildfly/test/security/common/elytron/SecurityRealm.java
```diff
@@ -0,0 +1,32 @@
...
+package org.wildfly.test.security.common.elytron;
+
+/**
+ * Interface representing Elytron Security realm and resources related to it.
+ *
+ * @author Josef Cacek
+ */
+public interface SecurityRealm extends ConfigurableElement {
+
+}
```
jamezp: Any reason for this since it doesn't add any behavior?

testsuite/elytron/src/test/java/org/wildfly/test/security/common/elytron/ServerSslContext.java
```diff
@@ -0,0 +1,32 @@
...
+package org.wildfly.test.security.common.elytron;
+
+/**
+ * Interface representing Elytron server-ssl-context.
+ *
+ * @author Josef Cacek
+ */
+public interface ServerSslContext extends ConfigurableElement {
+
+}
```
jamezp: Any reason for this since it doesn't add any behavior?

testsuite/elytron/src/test/java/org/wildfly/test/security/common/elytron/TrustManager.java
```diff
@@ -0,0 +1,32 @@
...
+package org.wildfly.test.security.common.elytron;
+
+/**
+ * Interface representing Elytron trust-manager.
+ *
+ * @author Josef Cacek
+ */
+public interface TrustManager extends ConfigurableElement {
+
+}
```
jamezp: Any reason for this since it doesn't add any behavior?

kwart:
@jamezp Thanks for the review. I'm going to make the requested fixes.

testsuite/elytron/src/test/java/org/jboss/as/test/shared/CliUtils.java
```diff
@@ -0,0 +1,53 @@
...
+
+    /**
+     * Escapes given path String for CLI.
+     *
+     * @param path path string to escape (must be not-<code>null</code>)
+     * @return escaped path
+     */
+    public static String escapePath(String path) {
+        return Objects.requireNonNull(path, "Path to escape can't be null.").replace("\\", "\\\\");
+    }
```
kwart: It's used in `Path.asString()` method which constructs CLI command fragment.

testsuite/elytron/src/test/java/org/jboss/as/test/shared/CliUtils.java
```diff
@@ -0,0 +1,53 @@
...
+
+    /**
+     * Returns escaped absolute path of given File instance.
+     *
+     * @param file instance to get the path from (must be not-<code>null</code>)
+     * @return escaped absolute path
+     */
+    public static String asAbsolutePath(File file) {
+        return escapePath(Objects.requireNonNull(file, "File can't be null.").getAbsolutePath());
+    }
```
kwart: I will remove this one. I was just copying the class from full wildfly.

testsuite/elytron/src/test/java/org/wildfly/test/integration/elytron/sasl/mgmt/AbstractMgmtSaslTestBase.java
kwart: IMO, the class names are self-descriptive enough:
* SimpleSocketBinding - `/socket-binding-group=standard-sockets/socket-binding=[NAME]`
* SimpleMgmtNativeInterface - `/core-service=management/management-interface=native-interface`

testsuite/elytron/src/test/java/org/wildfly/test/security/common/AbstractConfigSetupTask.java
kwart: I would keep it on INFO level. When sth. changes in the model or some test doesn't do a proper clean-up and the configuration task stops to work, then it's easier to find the exact place which failed.

testsuite/elytron/src/test/java/org/wildfly/test/security/common/elytron/CliFragment.java
kwart: The `asString` makes it IMO more obvious that it's not used just for printing into log. It's not just a textual representation of the data, it also has to have a correct form to be used in CLI command.

testsuite/elytron/src/test/java/org/wildfly/test/security/common/elytron/ConfigurableElement.java
kwart: At least one `create` method and at least one `remove` method has to be implemented for the ConfigurableElements.

Originally (in full) there were just the methods with single argument (`CLIWrapper`) and no default. When we needed to add the `ModelControllerClient` to have possibility to handle the config in safer way, we have choosen this approach with default implementations and throwing exception if the override implementation is missing.

As we don't need to keep the utils in sync with full now, I will try to fix it here.

testsuite/elytron/src/test/java/org/wildfly/test/security/common/elytron/KeyManager.java
kwart: They were meant as marker interfaces representing elytron capabilities. Nevertheless we still don't have usage for them, so I'll remove all of them.

testsuite/elytron/src/test/java/org/wildfly/test/security/common/elytron/Path.java
kwart: It makes sense, I will change it.

kwart:
The fixes were implemented. The `@Ignore` annotations were removed as the issues are already resolved by Elytron version in wildfly-core. The only `Assume` remains for `PLAIN` mechanism and information about it was included in [JBEAP-11147|https://issues.jboss.org/browse/JBEAP-11147] description.

testsuite/elytron/server-provisioning.xml
```diff
@@ -0,0 +1,27 @@
+<!--
```
luck3y: This license bit should be ASL, (or just removed, its just an xml file?)

testsuite/elytron/src/test/java/org/jboss/as/test/shared/CliUtils.java
jamezp: Ah gotcha sorry. It's not a file system path it's a CLI path.

testsuite/elytron/server-provisioning.xml
kwart: I'll remove it. I've used just a copy from the `standalone` module.

kabir:
Retest this please

kwart:
Retest this please.

testsuite/elytron/src/test/java/org/wildfly/test/security/common/elytron/SimpleServerSslContext.java
```diff
@@ -0,0 +1,141 @@
...
+        this.protocols = builder.protocols;
+        this.needClientAuth = builder.needClientAuth;
+        this.authenticationOptional = builder.authenticationOptional;
+    }
+
+    @Override
+    public void create(ModelControllerClient client, CLIWrapper cli) throws Exception {
+        // /subsystem=elytron/server-ssl-context=twoWaySSC:add(key-manager=twoWayKM,protocols=["TLSv1.2"],
+        // trust-manager=twoWayTM,security-domain=test,need-client-auth=true)
+        StringBuilder sb = new StringBuilder("/subsystem=elytron/server-ssl-context=").append(name).append(":add(");
```
ctomc: for all this concatenation of paths, you should use PathAddress class which would really make core more readable and less error prone.

testsuite/elytron/src/test/resources/logging.properties
```diff
@@ -0,0 +1,31 @@
+# Additional loggers to configure (the root logger is always configured)
```
ctomc: why is this file needed?

testsuite/elytron/src/test/java/org/wildfly/test/security/common/elytron/SimpleServerSslContext.java
kwart: I've reused [the implementation from WildFly]( https://github.com/wildfly/wildfly/blob/master/testsuite/shared/src/main/java/org/wildfly/test/security/common/elytron/SimpleServerSslContext.java) here. If it doesn't fit into core, I can refactor it.

testsuite/elytron/src/test/resources/logging.properties
kwart: It allows to control what's included in the log on the client side. E.g. when debugging SASL mechanisms, I want to see sth like:

```
2017-06-19 16:47:45,584 TRACE [org.wildfly.security] (main) getAuthenticationConfiguration uri=remote://127.0.0.1:10567, protocolDefaultPort=-1, abstractType=null, abstractTypeAuthority=null, MatchRule=[], AuthenticationConfiguration=[AuthenticationConfiguration:principal=digest-md5,set-host=127.0.0.1,set-port=10567,credentials-present,providers-supplier=org.wildfly.security.util.ProviderUtil$1@53f0a4cb,sasl-mechanism-selector=DIGEST-MD5,mechanism-properties={wildfly.sasl.local-user.quiet-auth=true}]
2017-06-19 16:47:45,592 TRACE [org.wildfly.security] (main) getAuthenticationConfiguration uri=remote://127.0.0.1:10567, protocolDefaultPort=-1, abstractType=null, abstractTypeAuthority=null, MatchRule=[], AuthenticationConfiguration=[AuthenticationConfiguration:principal=digest-sha,set-host=127.0.0.1,set-port=10567,credentials-present,providers-supplier=org.wildfly.security.util.ProviderUtil$1@8c11eee,sasl-mechanism-selector=DIGEST-SHA,mechanism-properties={wildfly.sasl.local-user.quiet-auth=true}]
2017-06-19 16:47:45,592 TRACE [org.wildfly.security] (main) getAuthenticationConfiguration uri=remote://127.0.0.1:10567, protocolDefaultPort=-1, abstractType=null, abstractTypeAuthority=null, MatchRule=[], AuthenticationConfiguration=[AuthenticationConfiguration:principal=digest-sha,set-host=127.0.0.1,set-port=10567,credentials-present,providers-supplier=org.wildfly.security.util.ProviderUtil$1@8c11eee,sasl-mechanism-selector=DIGEST-SHA,mechanism-properties={wildfly.sasl.local-user.quiet-auth=true}]
2017-06-19 16:47:45,599 TRACE [org.wildfly.security] (main) getAuthenticationConfiguration uri=remote://127.0.0.1:10567, protocolDefaultPort=-1, abstractType=null, abstractTypeAuthority=null, MatchRule=[], AuthenticationConfiguration=[AuthenticationConfiguration:principal=digest-sha-256,set-host=127.0.0.1,set-port=10567,credentials-present,providers-supplier=org.wildfly.security.util.ProviderUtil$1@333dd51e,sasl-mechanism-selector=DIGEST-SHA-256,mechanism-properties={wildfly.sasl.local-user.quiet-auth=true}]
```

So I just need to change level from `INFO` to `TRACE` in the `logging.properties`.

testsuite/elytron/src/test/resources/logging.properties
```diff
@@ -0,0 +1,31 @@
...
+logger.handlers=FILE,CONSOLE
+
+logger.org.wildfly.security.level=INFO
+logger.org.wildfly.security.useParentHandlers=true
+
+logger.org.jboss.remoting3.level=INFO
+logger.org.jboss.remoting3.useParentHandlers=true
+
+handler.FILE=org.jboss.logmanager.handlers.PeriodicRotatingFileHandler
+handler.FILE.level=ALL
```
ctomc: never in default config.
default level for testsuite could only be OFF not ALL.

testsuite/elytron/src/test/resources/logging.properties
```diff
@@ -0,0 +1,31 @@
+# Additional loggers to configure (the root logger is always configured)
+loggers=org.wildfly.security,org.jboss.remoting3
+
+logger.level=INFO
+logger.handlers=FILE,CONSOLE
```
ctomc: if anything only FILE

kwart: IMO, if not both, then at least `CONSOLE`, the redirection to file is then controlled by surefire plugin.

testsuite/elytron/src/test/java/org/wildfly/test/security/common/elytron/SimpleServerSslContext.java
jamezp: I'm personally okay with not using a `PathAddress` as it's _not_ a public API. I do realize this is just a test though.

testsuite/elytron/src/test/java/org/wildfly/test/integration/elytron/sasl/mgmt/AbstractMgmtSaslTestBase.java
```diff
@@ -0,0 +1,361 @@
...
+            assertThat("SaslException was expected as the second cause when SASL authentication fails", cause.getCause(),
+                    is(instanceOf(SaslException.class)));
+        }
+    }
+
+    /**
+     * Executes :whoami operation and returns the result ModelNode.
+     */
+    protected static ModelNode executeWhoAmI() throws IOException {
+        ModelControllerClient client = ModelControllerClient.Factory
```
jamezp: This should be in a try-with-resources block.

kwart: OK, will do.

PR #2483, opened 2017-05-30
WFCORE-2792 AttributeDefinition.parser is not consistently used in su…

ctomc:
…bclasses

- update parser / marshaller api
- cleanup some deprecated usages

### DIFF
--- a/controller/src/main/java/org/jboss/as/controller/AttributeDefinition.java
+++ b/controller/src/main/java/org/jboss/as/controller/AttributeDefinition.java
@@ -130,7 +130,7 @@ public abstract class AttributeDefinition {
     private AttributeDefinition(String name, String xmlName, final ModelNode defaultValue, final ModelType type,
                                 final boolean allowNull, final boolean allowExpression, final MeasurementUnit measurementUnit,
                                 final ParameterCorrector valueCorrector, final ParameterValidator validator, final boolean validateNull,
-                                final String[] alternatives, final String[] requires, AttributeMarshaller attributeMarshaller,
+                                final String[] alternatives, final String[] requires, AttributeMarshaller marshaller,
                                 boolean resourceOnly, DeprecationData deprecationData, final List<AccessConstraintDefinition> accessConstraints,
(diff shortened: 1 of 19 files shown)


controller/src/main/java/org/jboss/as/controller/AttributeDefinition.java
```diff
@@ -81,7 +81,7 @@
     private final ParameterCorrector valueCorrector;
     private final ParameterValidator validator;
     private final EnumSet<AttributeAccess.Flag> flags;
-    protected final AttributeMarshaller attributeMarshaller;
+    protected final AttributeMarshaller marshaller;
```
bstansberry: This is a breaking change, as the field is visible.

Let's 

private final AttributeMarshaller marshaller;
/** @deprecated use {@link #getMarshaller()} */
@Deprecated
protected final AttributeMarshaller attributeMarshaller;

and in c'tor

this.attributeMarshaller = marshaller;

protected fields are annoying because it's too easy to break them.

ctomc: good point, will revert this, but will also look into removing bunch of deprecated constructors in AttributeDefinition and its inheritance chain, as we moved to builder api long time ago and I think they should be ripe to be removed.

controller/src/main/java/org/jboss/as/controller/AttributeDefinition.java
```diff
@@ -109,7 +109,7 @@ protected AttributeDefinition(AbstractAttributeDefinitionBuilder<?, ?> toCopy) {
                 toCopy.getAttributeGroup(), toCopy.referenceRecorder, toCopy.getAllowedValues(), toCopy.getArbitraryDescriptors(),
                 toCopy.getUndefinedMetricValue(), wrapFlags(toCopy.getFlags()));
     }
-
+
```
bstansberry: Fails checkstyle:


[10:37:32][Step 2/3] /store/work/tc-work/4c983eca15c0017f/controller/src/main/java/org/jboss/as/controller/AttributeDefinition.java:112: Line has trailing spaces.

PR #2487, opened 2017-05-31
[WFCORE-2876]: Runtime-failure-causes-rollback does not seem to have effect when configured in model

ehsavoie:
The runtime-failure-causes-rollback wasn't properly propagated to the boot scanner.
If runtime-failure-causes-rollback is set to 'false' then failed deployments added at boot shouldn't be undeployed.

Jira: https://issues.jboss.org/browse/WFCORE-2876

### DIFF
--- a/deployment-scanner/src/main/java/org/jboss/as/server/deployment/scanner/FileSystemDeploymentService.java
+++ b/deployment-scanner/src/main/java/org/jboss/as/server/deployment/scanner/FileSystemDeploymentService.java
@@ -767,7 +767,9 @@ class FileSystemDeploymentService implements DeploymentScanner, NotificationHand
 
         @Override
         public void run() {
-            forcedUndeployScan();
+            if(rollbackOnRuntimeFailure) {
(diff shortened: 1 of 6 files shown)


deployment-scanner/src/main/java/org/jboss/as/server/deployment/scanner/FileSystemDeploymentService.java
```diff
@@ -274,7 +274,7 @@ public void run() {
         assert scheduledExecutor != null;
         assert deploymentDir != null;
 
-        this.resourceAddress = resourceAddress.toModelNode();
+       this.resourceAddress = resourceAddress.toModelNode();
```
bstansberry: The tests are failing so you'll need to update this branch to fix that, so might as well correct this whitespace. :)

ehsavoie:
retest this please

ehsavoie:
retest this please

ehsavoie:
retest this please

PR #2497, opened 2017-06-02
[WFCORE-2907] Ensure the KeyManager is unwrapped if we detect the JVM is running with FIPS enabled.

darranl:
https://issues.jboss.org/browse/WFCORE-2907
https://issues.jboss.org/browse/JBEAP-11259

### DIFF
--- a/elytron/src/main/java/org/wildfly/extension/elytron/SSLDefinitions.java
+++ b/elytron/src/main/java/org/wildfly/extension/elytron/SSLDefinitions.java
@@ -40,6 +40,8 @@ import static org.wildfly.extension.elytron._private.ElytronSubsystemMessages.RO
 import java.io.File;
 import java.io.FileInputStream;
 import java.io.FileNotFoundException;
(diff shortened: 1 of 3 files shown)


elytron/src/main/java/org/wildfly/extension/elytron/SSLDefinitions.java
```diff
@@ -991,4 +1003,28 @@ protected void executeRuntimeStep(OperationContext context, ModelNode operation)
         protected abstract ServiceUtil<SSLContext> getSSLContextServiceUtil();
     }
 
+    private static BooleanSupplier getFipsSupplier() {
+        try {
+            final Class<?> providerClazz = SSLDefinitions.class.getClassLoader().loadClass("com.sun.net.ssl.internal.ssl.Provider");
```
mchoma: I can't find such class in IBM distribution. So I assume that will throw ClassNotFoundException.

PR #2499, opened 2017-06-02
WFCORE-2903: Remove case-sensistive attribute from Elytron security realm configurations

yersan:
Jira issues:
https://issues.jboss.org/browse/WFCORE-2903
https://issues.jboss.org/browse/JBEAP-10255

It depends on https://github.com/wildfly/wildfly/pull/10187


darranl:
I would suggest rather than requiring coordinated PRs we try the following: -

1. See if we can remove the use of the attribute from WildFly first before this PR is merged.

2. Temporarily disable the test in WildFly and enable it again after core is tagged and WildFly using it then the test can be enabled and fixed.

yersan:
@darranl  I confirm you we can remove the case sensitivity configuration from Elytron test suite on wildfly first and disable temporarily the AggregateRealmTestCase. Do you want to proceed with the changes?

yersan:
This commit has to be coordinated with a wildfly dependent commit. There are no pending changes from my side.

bstansberry:
Superceded by #2536

PR #2509, opened 2017-06-05
[WFCORE-2893] Wrong schema version in several xsd files

apintea:
https://issues.jboss.org/browse/WFCORE-2893


bstansberry:
this is ok to test

PR #2511, opened 2017-06-05
WFCORE-2882 Review & fix xml persistence of elytron subsystem, Part 1

ctomc:
- move security-property sub resource to security-properties properties attribute
 - unify "properties" attribute usage and persistence
 - move parsers to parser definition
 - consistently use attribute groups
 - consistently use complex parsers / marshallers
 - adjust xsd to reflect common model rules (attribute groups, complex attribute, simple attributes)
 - cleanup


elytron/src/test/resources/org/wildfly/extension/elytron/tls-test.xml
```diff
@@ -23,21 +23,25 @@
     </credential-stores>
     <tls>
         <key-stores>
-            <key-store name="FireflyKeystore" type="JKS">
-                <file path="firefly.keystore" relative-to="jboss.server.config.dir"/>
+            <key-store name="FireflyKeystore" >
                 <credential-reference store="test" alias="master-password-alias"/>
+                <implementation type="JKS" />
```
darranl: Why have we needed to move the type to a new element?

ctomc: To follow the model configuration.
you have attribute group "implementation" that has this attributes on it.
It is present in model and as such grouped in admin console, so it makes sense to have it grouped also in xml.

on same resource there are other 3 attributes that are in group "file" which is already used also in xml. So for consistency sake all attribute groups are now persisted the same.
We could change parser config to not use elements for groups but in that case file attributes would follow.

kabir:
Retest this please

kabir:
Retest this please

elytron/src/test/resources/org/wildfly/extension/elytron/tls-test.xml
darranl: I think we need to be careful if anything could end up with defaults defined for the attributes (we may be Ok with the ones you identify so far) - if attributes have defaults defined I don't think it makes sense for them to be represented in a child element.

ctomc: And it is not just type attribute, that come here, you have
- type
- provider-name
- providers
that are all part of same attribute group

darranl: I think none of those have defaults defined so in this case we can get away with them moving to their own element and we can still document the behaviour in the schema in a meaningful way, although type should probably be mandatory if the implementation element is present, I don't think it makes sense to have implementation without type.

ctomc: type is mandatory when implementation element is present.

also it could make sense to have type default to "JKS" as I think in most scenarios this is what users will configure anyway.

darranl: No don't add any default there - I believe we have agreed to go for the JDK default if not specified.  If we set a default then it does not really make sense to split out into it's own element.

elytron/src/main/resources/org/wildfly/extension/elytron/LocalDescriptions.properties
```diff
@@ -12,7 +12,7 @@ elytron.disallowed-providers=A list of providers that are not allowed, and will
 # Security Properties #
 #######################
 
-elytron.security-property=A definition of a security property to be set.
+elytron.security-properties=A definition of a security property to be set.
```
bstansberry: This should be reworded now, since it's now multiple properties.

elytron/src/main/resources/org/wildfly/extension/elytron/LocalDescriptions.properties
```diff
@@ -12,7 +12,7 @@ elytron.disallowed-providers=A list of providers that are not allowed, and will
...
 # Operations
 elytron.security-property.add=Add operation for adding a new security property.
 elytron.security-property.remove=Remove operation for removing a security property.
```
bstansberry: Are these two relevant any more?

elytron/src/test/resources/org/wildfly/extension/elytron/tls-test.xml
bstansberry: I'm a bit confused by this discussion. KeyStoreDefinition.TYPE is a required attribute. So I'm not sure how defaults or "go for the JDK default if not specified" come into it, unless the concern is because there's other work that will make that attribute not required.

FWIW, the persistence rules are we persist an attribute if the user explicitly configured it, whether the value matches the default value or not. But we don't write a default value if the user-provided value is undefined. The presence or absence of an attribute group *shouldn't* affect that.

I haven't looked at the details of the marshaller code but my expectation is that if an attribute group has no defined attributes, then no element for that group would be written.

bstansberry:
Has this been discussed with @claudio4j ? The security-property change will affect the console.

I haven't thought about it hard, but I think I'm fine leaving those as resources, unless @darranl wants the change. Are they analogous to system-property resources, where each is an independent entity? That's a different thing from a set of properties that are an inherent part of the config of some parent.

bstansberry:
What is part 2?

ctomc:
@bstansberry part 2 (and maybe 3) is to fix & cleanup even more parser code in elytron subsystem, this PR makes a good dent but it still not there.
Optimal end solution would be that we would get rid of all custom code and only have "defined" parser and that is it, but we can only do that after all "sub" parsers are migrated.

elytron/src/test/resources/org/wildfly/extension/elytron/tls-test.xml
darranl: What I don't like is having a default value defined for an attribute and then that attribute being moved to an optional element.

If the element does not exist in the document does that default apply or not?  If it makes it easier to avoid this I would rather just loose the attribute group.

darranl:
@bstansberry Yes security properties would have been modelled to be analogous to system properties.

I am happy for any changes to be applied that bring consistency with the rest of the model but this does raise the question of why treat these differently to system properties.

claudio4j:
Hi @ctomc does the change Brian says above, is to remove security-property resource and add it as an attribute of the elytron subsystem ? If yes, them I am fine, it is an easy modification on web console side.

ctomc:
@claudio4j yes, that is the model change part of this part. 
Tnx for approving of the change.

bstansberry:
retest this please

PR #2513, opened 2017-06-05
WFCORE-2923, Credential reference integration with

mchoma:
client-certificate-store test


ehsavoie:
Duplicates https://github.com/wildfly/wildfly-core/pull/2514 without using the credential store for format that are not clear text

PR #2517, opened 2017-06-07
WFCORE-2446: Use required attribute for Elytron key-store element in xsd file

yersan:
There is an inconsistency between the xsd file and model definition for key-store element in key-manager and trust-manager 

Jira issues:
https://issues.jboss.org/browse/WFCORE-2446
https://issues.jboss.org/browse/JBEAP-7401


yersan:
Retest this please

PR #2518, opened 2017-06-07
Fix for WFCORE-2926. Disable error-on-interract for CLI.connect.

jfdenise:
CLI class connect method should set ignoreOnInteract to false.
In addition I added a connect method to allow to provide a CommandContextConfiguration. This class is in an impl package but has already been exposed as a public API in the CommandContextFactory. I wanted to move it to public API, but being already exposed we can't change its location.


bmaxwell:
retest this please!!

PR #2520, opened 2017-06-07
WFCORE-2933 add support for empty host.xml & domain.xml when starting…

luck3y:
… an embedded-HC

(This should probably have a 4.0 label attached.)

(--empty-domain-config, --remove-existing-domain-config, --empty-host-config,
--remove-existing-host-config)

@aloubyansky FYI, this still has an issue I'll push an update for with using reload after configuring, the persisted changes still aren't loading for some reason. (Doing stop-embedded-host-controller and a start loads correctly, however).

### DIFF
--- a/host-controller/src/main/java/org/jboss/as/host/controller/DomainModelControllerService.java
+++ b/host-controller/src/main/java/org/jboss/as/host/controller/DomainModelControllerService.java
@@ -648,41 +648,18 @@ public class DomainModelControllerService extends AbstractControllerService impl
             ConfigurationFile.InteractionPolicy configPolicy = environment.getHostConfigurationFile().getInteractionPolicy();
             if (!environment.getRunningModeControl().isReloaded() && (configPolicy == ConfigurationFile.InteractionPolicy.NEW || configPolicy == ConfigurationFile.InteractionPolicy.DISCARD)) {
 
+                String hostControllerName = environment.getHostControllerName();
                 // minimum needed to boot an embedded HC from an empty host.xml
-                ModelNode baseOp;
-
-                baseOp = new ModelNode();
+                ModelNode baseOp = new ModelNode();
                 baseOp.get("operation").set("register-host-model");
-                baseOp.get("name").set("master");
+                baseOp.get("name").set(hostControllerName);
                 baseOp.get("address").addEmptyList();
-                hostBootOps.add(baseOp);
-
-                baseOp = new ModelNode();
-                baseOp.get("operation").set("write-attribute");
-                baseOp.get("name").set("name");
-                baseOp.get("value").set("master");
-                baseOp.get("address").set(PathAddress.pathAddress("host", "master").toModelNode());
-                hostBootOps.add(baseOp);
-
-                // mgmt extension
-                baseOp = new ModelNode();
-                baseOp.get("operation").set("add");
-                baseOp.get("extension").set("name");
-                baseOp.get("address").set(PathAddress.pathAddress("host", "master").append(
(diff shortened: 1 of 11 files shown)


luck3y:
retest this please

luck3y:
@aloubyansky This should be good to go if you want to give it a try. The reload stuff tests OK now.

One warning. Booting with an empty config and changing the HC's name (/host=master:write-attribute(name=name, value=newName) followed by a reload will cause the cli to timeout and the HC will need to be started again, as it's trying to connect to the old named HC. I think I can figure out a fix, but it will take a little time.

aloubyansky:
That is good enough for me right now. Thanks! Just need it merged.

luck3y:
retest this please

luck3y:
retest this please

luck3y:
retest this please

luck3y:
retest this please

luck3y:
retest this please

luck3y:
retest this please

luck3y:
retest this please

luck3y:
@bstansberry @aloubyansky This just needs review now, and it should be ready for merge into 4.x.

host-controller/src/main/java/org/jboss/as/host/controller/DomainModelControllerService.java
```diff
@@ -640,11 +641,53 @@ protected void boot(final BootContext context) throws ConfigurationPersistenceEx
...
+                baseOp.get("address").set(PathAddress.pathAddress("host", "master").append(
+                        PathAddress.pathAddress("extension", "org.wildfly.extension.core-management")).toModelNode());
+                hostBootOps.add(baseOp);
+
+                // mgmt subsystem
+                baseOp = new ModelNode();
+                baseOp.get("operation").set("add");
+                baseOp.get("address").set(PathAddress.pathAddress("host", "master").
+                        append(PathAddress.pathAddress("subsystem", "core-management")).toModelNode());
+                hostBootOps.add(baseOp);
```
bstansberry: Why is this subsystem and extension needed?

luck3y: This is going back a while, but as I recall not having it present would result in the CLI being unable to connect or something similar. It's probably worth a retest at this point and see if it's still behaving in the same way.

If it doesn't work with this present, would you consider that a bug? The specific use case here is booting with empty configs, host name set to 'master' and assumed to be domain controller.

luck3y:
@bstansberry This has all of the changes we discussed, and still appears to work ok (and the tests pass).

@aloubyansky Brian and I discussed the initial boot ops being used here, and decided to remove the management extension and subsystem as not needed, The host-controller name (in /host=) is no long defaulted to 'master', but is set to the computed hostControllerName from HostControllerEnvironment.

Let me know if any of these changes cause you headaches and we can discuss additional changes.

Other, than that, this should be ready to go.

luck3y:
Note, failures are legitimate, I missed something when tests were run locally, fix coming.

luck3y:
@aloubyansky as discussed this adds a --host-controller-name= parameter to the start-embedded-server-command, which will override the name obtained from HostControllerEnvironment.

In the case you don't want to use this, but read the environment one, something like the below should work (You probably know this already, but just in case).

[domain@embedded /] :read-attribute(name=local-host-name)
{
    "outcome" => "success",
    "result" => "master"
}

luck3y:
I've added a restriction for only allowing the HC name to be set when using --empty-host-config

I think this should be all set to go now.

luck3y:
Closing this one in favor of: https://github.com/wildfly/wildfly-core/pull/2938

PR #2521, opened 2017-06-08
WFCORE-2929: Added missing default values in management model for Elytron ldap-key-store

yersan:
Default values were already being correctly used in LdapKeyStore. The only advantage I can see setting default values on LdapKeyStoreDefinition is LdapKeyStoreService will end up with those defaults instead of nulls, and any class using the service will get the "real" defaults, so I created this PR instead of solving the issue directly. 

Jira issues:
https://issues.jboss.org/browse/WFCORE-2929
https://issues.jboss.org/browse/JBEAP-7487

### DIFF
--- a/elytron/src/main/java/org/wildfly/extension/elytron/LdapKeyStoreDefinition.java
+++ b/elytron/src/main/java/org/wildfly/extension/elytron/LdapKeyStoreDefinition.java
@@ -95,6 +95,7 @@ final class LdapKeyStoreDefinition extends SimpleResourceDefinition {
             .setXmlName("recursive")
             .setAllowExpression(true)
             .setRestartAllServices()
+            .setDefaultValue(new ModelNode(true))
             .build();
 
     static final SimpleAttributeDefinition SEARCH_TIME_LIMIT = new SimpleAttributeDefinitionBuilder(ElytronDescriptionConstants.SEARCH_TIME_LIMIT, ModelType.INT, true)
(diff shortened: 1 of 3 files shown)


bstansberry:
retest this please

@darranl please approve

darranl:
@bstansberry / @kabir approved

bstansberry:
@yersan This needs a rebase now. Ping me when it's rebased and I'll merge it once CI passes since it's been approved.

yersan:
Hi @bstansberry , the branch has just been rebased

darranl:
I don't thinks this one is affected but we need to double check the schema following the previous model changes - as we have default values we need to ensure they are not on an optional element unless they all require another attribute on that element.

yersan:
@darranl shouldn't optional values be the values that are likely to be configured with defaults? Probably I'm misunderstanding what defaults mean in this context, for me was, if an element doesn't have a value, then it gets the default.

The problem here is the test expects the same xml after marshalling the configuration, but since we have defatult values at model level, default values are not written to config file, then the test will fail comparing the configuration files if you include default values in your configuration.

in general, is that behaviour correct? I guess yes, so maybe we have just to adap the failed ldap.xml file removing defaults values from the configuration or, maybe, the test should be adapted to work in those cases.

darranl:
Looking at the latest version of the schema this does appear to still be Ok.

yersan:
The I have chosen the simplest way (I hope is the correct in this case), I have removed the default values from ldap.xml test file to avoid a test failure comparing the configuration after marshalling.

yersan:
Retest this please

yersan:
retest this please

bstansberry:
@yersan if you had to remove default values from the config file to prevent failures, we have a bug that must be fixed.

Looks like https://github.com/wildfly/wildfly-core/pull/2352 needs to be resurrected.

ctomc:
fix for that could also be
https://github.com/wildfly/wildfly-core/blob/master/elytron/src/main/java/org/wildfly/extension/elytron/RealmParser.java#L88

` .setMarshallDefaultValues(true|false)`

yersan:
@bstansberry @ctomc Thanks that was the trick here, since #2352 is still dead, I used `.setMarshallDefaultValues(true)` at TLS ldap-key-store parser level. It worked smoothly.

PR #2527, opened 2017-06-08
[WFCORE-2938] Ensure the SSLContext from an injected legacy security realm is used for a remote outbound connection regardless of whether or not a username is specified

fjuma:
https://issues.jboss.org/browse/WFCORE-2938
https://issues.jboss.org/browse/JBEAP-10860


fjuma:
Retest this please

PR #2534, opened 2017-06-12
Fix tests that rely on English messages.

jamezp:
The changes where the message bundle interface was used may not be needed if the same locale is used for tests and the server boot. However testing with `-Djvm.args.other=-Duser.language=de` only boots the server with the locale and the test itself will still run under the users default locale.

This is a test run with translated messages added https://ci.wildfly.org/viewLog.html?buildId=63595&.

### DIFF
--- a/testsuite/standalone/src/test/java/org/jboss/as/test/integration/management/cli/BasicOpsTestCase.java
+++ b/testsuite/standalone/src/test/java/org/jboss/as/test/integration/management/cli/BasicOpsTestCase.java
@@ -72,26 +72,24 @@ public class BasicOpsTestCase {
     public void testReadAttribute() throws Exception {
         CLIWrapper cli = new CLIWrapper(true);
         {
-            cli.sendLine("read-attribute namespaces");
-            String output = cli.readOutput();
-            assertFalse(output.contains("Unknown attribute 'namespaces'"));
+            assertTrue("Failed to read attribute: " + cli.readOutput(),
+                    cli.sendLine("read-attribute namespaces", true));
         }
         {
-            cli.sendLine("read-attribute --node=subsystem=request-controller namespaces", true);
+            assertFalse("Expected command to fail on invalid node",
+                    cli.sendLine("read-attribute --node=subsystem=request-controller namespaces", true));
             String output = cli.readOutput();
-            assertTrue(output.contains("Unknown attribute 'namespaces'"));
-        }
-
-        {
-            cli.sendLine("read-attribute active-requests --node=subsystem=request-controller");
-            String output = cli.readOutput();
-            assertFalse(output.contains("Unknown attribute 'active-requests'"));
+            assertTrue(output.matches("WFLYCTL0201.*namespaces.*"));
         }
 
         {
-            cli.sendLine("read-attribute --node=subsystem=request-controller active-requests");
-            String output = cli.readOutput();
-            assertFalse(output.contains("Unknown attribute 'active-requests'"));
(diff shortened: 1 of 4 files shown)


jamezp:
At least some of the failures are legitimate. It looks like in the German translation the `'` was changed to `"`. The test should likely just ignore quotations.

bstansberry:
@jfdenise please approve the BasicOpsTestCase part.

testsuite/standalone/src/test/java/org/jboss/as/test/integration/management/cli/BasicOpsTestCase.java
```diff
@@ -72,26 +72,24 @@ public void testLs() throws Exception {
     public void testReadAttribute() throws Exception {
         CLIWrapper cli = new CLIWrapper(true);
         {
-            cli.sendLine("read-attribute namespaces");
-            String output = cli.readOutput();
-            assertFalse(output.contains("Unknown attribute 'namespaces'"));
+            assertTrue("Failed to read attribute: " + cli.readOutput(),
+                    cli.sendLine("read-attribute namespaces", false));
```
jfdenise: The way I see the test, it seems that true should be passed to sendLine method call in order to ignore error and have the return value be true if no error, false is error occurred.

testsuite/standalone/src/test/java/org/jboss/as/test/integration/management/cli/BasicOpsTestCase.java
```diff
@@ -72,26 +72,24 @@ public void testLs() throws Exception {
...
-            assertTrue(output.contains("Unknown attribute 'namespaces'"));
+            assertTrue(output.matches("WFLYCTL0201.*namespaces.*"));
         }
 
         {
-            cli.sendLine("read-attribute active-requests --node=subsystem=request-controller");
-            String output = cli.readOutput();
-            assertFalse(output.contains("Unknown attribute 'active-requests'"));
+            assertTrue("Failed to read attribute: " + cli.readOutput(),
+                    cli.sendLine("read-attribute active-requests --node=subsystem=request-controller", false));
```
jfdenise: The way I see the test, it seems that true should be passed to sendLine method call in order to ignore error and have the return value be true if no error, false is error occurred.

testsuite/standalone/src/test/java/org/jboss/as/test/integration/management/cli/BasicOpsTestCase.java
```diff
@@ -72,26 +72,24 @@ public void testLs() throws Exception {
...
         }
 
         {
-            cli.sendLine("read-attribute --node=subsystem=request-controller active-requests");
-            String output = cli.readOutput();
-            assertFalse(output.contains("Unknown attribute 'active-requests'"));
+            assertTrue("Failed to read attribute: " + cli.readOutput(),
+                    cli.sendLine("read-attribute --node=subsystem=request-controller active-requests", false));
```
jfdenise: The way I see the test, it seems that true should be passed to sendLine method call in order to ignore error and have the return value be true if no error, false is error occurred.

testsuite/standalone/src/test/java/org/jboss/as/test/integration/management/cli/BasicOpsTestCase.java
jamezp: Yes good catch. I'll fix that.

PR #2535, opened 2017-06-13
[WFCORE-2942] Adding credential-reference with non-exist alias to aut…

gaol:
…hentication-configuration through CLI results to NPE

Jira: https://issues.jboss.org/browse/WFCORE-2942

### DIFF
--- a/elytron/src/main/java/org/wildfly/extension/elytron/AuthenticationClientDefinitions.java
+++ b/elytron/src/main/java/org/wildfly/extension/elytron/AuthenticationClientDefinitions.java
@@ -317,7 +317,15 @@ class AuthenticationClientDefinitions {
                         try {
                             CredentialSource cs = sourceSupplier.get();
                             if (cs != null) {
-                                return c.usePassword(cs.getCredential(PasswordCredential.class).getPassword());
+                                PasswordCredential passCredential = cs.getCredential(PasswordCredential.class);
+                                String alias = credentialReference.hasDefined(CredentialReference.ALIAS) ? credentialReference.get(CredentialReference.ALIAS).asString() : null;
+                                if (passCredential == null) {
+                                    if (alias != null && alias.length() > 0) {
+                                        throw ROOT_LOGGER.credentialDoesNotExist(alias, PasswordCredential.class.getName());
+                                    }
+                                    throw ROOT_LOGGER.credentialCannotBeResolved();
+                                }
+                                return c.usePassword(passCredential.getPassword());
                             } else {
                                 throw ROOT_LOGGER.credentialCannotBeResolved();
                             }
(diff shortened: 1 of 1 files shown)


elytron/src/main/java/org/wildfly/extension/elytron/AuthenticationClientDefinitions.java
```diff
@@ -317,7 +318,12 @@ static ResourceDefinition getAuthenticationClientDefinition() {
                         try {
                             CredentialSource cs = sourceSupplier.get();
                             if (cs != null) {
-                                return c.usePassword(cs.getCredential(PasswordCredential.class).getPassword());
+                                PasswordCredential passCredential = cs.getCredential(PasswordCredential.class);
+                                String alias = credentialReference.hasDefined(CredentialReference.ALIAS) ? credentialReference.get(CredentialReference.ALIAS).asString() : null;
+                                if (passCredential == null && alias != null && alias.length() > 0) {
+                                    throw ROOT_LOGGER.credentialDoesNotExist(alias, PasswordCredential.class.getName());
+                                }
+                                return c.usePassword(passCredential.getPassword());
```
bstansberry: This can still NPE if alias == null, although perhaps it's not really possible for passCredential to be null if alias is null. (I haven't looked to see.)

elytron/src/main/java/org/wildfly/extension/elytron/AuthenticationClientDefinitions.java
```diff
@@ -328,7 +334,16 @@ static ResourceDefinition getAuthenticationClientDefinition() {
...
                 final Function<AuthenticationConfiguration, AuthenticationConfiguration> finalConfiguration = configuration;
-                return () -> finalConfiguration.apply(null);
+                return () -> {
+                    try {
+                        return finalConfiguration.apply(null);
+                    } catch (IllegalStateException e) {
+                        if (e.getCause() != null) {
+                            throw new StartException(e.getCause().getMessage());
+                        }
+                        throw new StartException(e.getMessage());
```
bstansberry: I think both of these throws would be better if they included a cause so the stack trace is more useful.

elytron/src/main/java/org/wildfly/extension/elytron/AuthenticationClientDefinitions.java
gaol: The [alias is required](https://github.com/wildfly/wildfly-core/blob/6f65f331de79123468cf69d5a67a68d92a591feb/controller/src/main/java/org/jboss/as/controller/security/CredentialReference.java#L107) in this situation, but for thoroughly checking, I agree it will be better to check the case as well. :)

elytron/src/main/java/org/wildfly/extension/elytron/AuthenticationClientDefinitions.java
gaol: Yes, it is.

I decided to discard the cause because both CLI and server log will print the error message twice, like:
<pre>
{
    "outcome" => "failed",
    "failure-description" => {"WFLYCTL0080: Failed services" => {"org.wildfly.security.authentication-configuration.cfg2" => "WFLYELY00004: Unable to start the service.
    Caused by: org.jboss.as.controller.OperationFailedException: WFLYELY00920: Credential alias \"d\" of credential type \"org.wildfly.security.credential.PasswordCredential\" does not exist in the store [ \"WFLYELY00920: Credential alias \\\"d\\\" of credential type \\\"org.wildfly.security.credential.PasswordCredential\\\" does not exist in the store\" ]"}},
    "rolled-back" => true
}
</pre>

but it should be another issue if you agree, I will add the cause back for the accurate stack trace.

gaol:
Updated according to the comments. :)

PR #2538, opened 2017-06-13
[WFCORE-2946]:  Errors publishing a large deployment to a remote server.

ehsavoie:
Modifying the API so that we may use a java.nio.file.Path instead of a InputStream thus removing the need to read the full input to get its length.

Jira: https://issues.jboss.org/browse/WFCORE-2946
JBEAP: https://issues.jboss.org/browse/JBEAP-11513

### DIFF
--- a/controller-client/src/main/java/org/jboss/as/controller/client/OperationBuilder.java
+++ b/controller-client/src/main/java/org/jboss/as/controller/client/OperationBuilder.java
@@ -21,19 +21,21 @@
  */
 package org.jboss.as.controller.client;
 
-import org.jboss.as.controller.client.impl.InputStreamEntry;
-import org.jboss.as.protocol.StreamUtils;
+import java.io.BufferedInputStream;
+import java.io.DataOutput;
 import org.jboss.dmr.ModelNode;
 import org.wildfly.common.Assert;
 
-import java.io.DataOutput;
 import java.io.File;
-import java.io.FileInputStream;
 import java.io.FilterInputStream;
 import java.io.IOException;
 import java.io.InputStream;
+import java.nio.file.Files;
+import java.nio.file.Path;
 import java.util.ArrayList;
 import java.util.List;
+import org.jboss.as.controller.client.impl.InputStreamEntry;
+import org.jboss.as.protocol.StreamUtils;
 
 /**
  * Builder for a {@link Operation}.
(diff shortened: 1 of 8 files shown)


controller-client/src/main/java/org/jboss/as/controller/client/OperationBuilder.java
```diff
@@ -146,25 +169,25 @@ public static OperationBuilder create(final ModelNode operation, final boolean a
...
 
         @Override
         public void copyStream(final DataOutput output) throws IOException {
-            final FileInputStream is = new FileInputStream(file);
-            try {
-                StreamUtils.copyStream(is, output);
-                is.close();
-            } finally {
-                StreamUtils.safeClose(is);
+            try (InputStream in = new BufferedInputStream(Files.newInputStream(file))) {
```
bstansberry: A try-with-resource is not the same as a try/finally with safeClose. The latter will not throw an exception on close; the former may.

controller-client/src/main/java/org/jboss/as/controller/client/helpers/domain/impl/DeploymentPlanBuilderImpl.java
```diff
@@ -85,11 +87,8 @@ public AddDeploymentPlanBuilder add(URL url) throws IOException, DuplicateDeploy
 
     @Override
     public AddDeploymentPlanBuilder add(String name, File file) throws IOException, DuplicateDeploymentNameException {
-        final InputStream is = new FileStreamEntry(file);
-        try {
+        try (final InputStream is = new FileStreamEntry(file)) {
```
bstansberry: Same here; the try-with-resources semantic is different.

controller-client/src/main/java/org/jboss/as/controller/client/helpers/domain/impl/DeploymentPlanBuilderImpl.java
```diff
@@ -181,11 +180,8 @@ public RemoveDeploymentPlanBuilder replace(URL url) throws IOException {
 
     @Override
     public RemoveDeploymentPlanBuilder replace(String name, File file) throws IOException {
-        final InputStream is = new FileStreamEntry(file);
-        try {
+        try (final InputStream is = new FileStreamEntry(file)) {
```
bstansberry: Same here.

controller-client/src/main/java/org/jboss/as/controller/client/OperationBuilder.java
bstansberry: Actually this one is ok, since L165 was closing the IS and that can throw an exception.

PR #2541, opened 2017-06-14
[WFCORE-2483] add test where server-identity/secret is obtained from credential store

hsvabek:
tests where server-identity/secret has set credential reference.

Jira: https://issues.jboss.org/browse/WFCORE-2483
JBEAP: https://issues.jboss.org/browse/JBEAP-9321


darranl:
Can we please add a reference to the WFCORE issue in the commit message.

bstansberry:
There's a SlaveHostControllerElytronAuthenticationTestCase. Shouldn't this be there?

Also, I haven't look closely but I believe the test needs to revert any config changes or it won't work reliably with the other test in the class which I guess isn't expecting the config to have been modified. (Test method order is not reliable.)

hsvabek:
Hi @bstansberry,
It is located in right test case. There is tested integration legacy core management with credential store only. And because host-slave-elytron.xml is used only for this testcase so any configuration revert is not necessary.
Could you please confirm these?

I will change commit message as wrote Darran.

PR #2545, opened 2017-06-14
Fix for WFCORE-2956. CLI deploy archive stuck when command timeout set

jfdenise:
Make executor to use a cachedThreadPool. Added unit tests.

### DIFF
--- a/cli/src/main/java/org/jboss/as/cli/impl/CommandExecutor.java
+++ b/cli/src/main/java/org/jboss/as/cli/impl/CommandExecutor.java
@@ -633,7 +633,7 @@ public class CommandExecutor {
     }
 
     private final CommandContext ctx;
-    private final ExecutorService executorService = Executors.newSingleThreadExecutor();
+    private final ExecutorService executorService = Executors.newCachedThreadPool();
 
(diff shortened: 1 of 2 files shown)


cli/src/main/java/org/jboss/as/cli/impl/CommandExecutor.java
```diff
@@ -633,7 +633,7 @@ public ModelNode execute(Operation op, String description) throws CommandLineExc
     }
 
     private final CommandContext ctx;
-    private final ExecutorService executorService = Executors.newSingleThreadExecutor();
+    private final ExecutorService executorService = Executors.newCachedThreadPool();
```
bstansberry: This isn't necessary for this PR -- just a thought...

Any pools we create should have a ThreadFactory so the threads get a meaningful name.

This isn't such a big deal with a standalone CLI, but given embed-server makes embedding the CLI more useful, it's a good practice.

jfdenise: Thank-you Brian, yes I have generalised that for 4.0 CLI. I missed this one.

PR #2546, opened 2017-06-14
WFCORE-2945 - Code is using ProcessType.HOST_CONTROLLER to identify HC, ignoring EMBEDDED_HOST_CONTROLLER

luck3y:
Also resolves:
WFLY-8923
JBEAP-11489

### DIFF
--- a/jmx/src/main/java/org/jboss/as/jmx/RemotingConnectorAdd.java
+++ b/jmx/src/main/java/org/jboss/as/jmx/RemotingConnectorAdd.java
@@ -28,6 +28,7 @@ import org.jboss.as.controller.AbstractAddStepHandler;
 import org.jboss.as.controller.OperationContext;
 import org.jboss.as.controller.OperationFailedException;
 import org.jboss.as.controller.PathAddress;
+import org.jboss.as.controller.ProcessType;
 import org.jboss.as.controller.descriptions.ModelDescriptionConstants;
 import org.jboss.as.controller.registry.Resource;
 import org.jboss.as.remoting.management.ManagementRemotingServices;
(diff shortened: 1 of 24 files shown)


jmx/src/main/java/org/jboss/as/jmx/RemotingConnectorAdd.java
```diff
@@ -47,6 +48,11 @@ private RemotingConnectorAdd() {
     }
 
     @Override
+    protected boolean requiresRuntime(OperationContext context) {
```
luck3y: This is the only oddity in here. Without this, embedded boot consistently fails with: 

19:39:12,063 ERROR [org.jboss.as.controller.management-operation] (Controller Boot Thread) WFLYCTL0013: Operation ("add") failed - address: ([
    ("host" => "master"),
    ("subsystem" => "jmx"),
    ("remoting-connector" => "jmx")
]) - failure description: {
    "WFLYCTL0412: Required services that are not installed:" => [
        "org.wildfly.management.jmx",
        "jboss.remoting.endpoint.management"
    ],
    "WFLYCTL0180: Services with missing/unavailable dependencies" => ["jboss.jmx.remoting-connector-ref is missing [org.wildfly.management.jmx, jboss.remoting.endpoint.management]"]
}

bstansberry: This is good; it's a fix for another (unrecorded) bug.

It's analogous to:

https://github.com/wildfly/wildfly-core/blob/master/host-controller/src/main/java/org/jboss/as/host/controller/operations/HttpManagementAddHandler.java#L102

https://github.com/wildfly/wildfly-core/blob/master/host-controller/src/main/java/org/jboss/as/host/controller/operations/NativeManagementAddHandler.java#L69

For embedded HC we didn't want it visible on the network, so those two handlers added that check. But we missed that the JMX subsystem can now run on the HC process too. Its connector should have the same behavior.

If we ever allow embedded HC to not be admin-only then we'll need to look into this handler the same way we do the other two.

bstansberry: @luck3y Once this is in can you launch an embedded HC and also an admin-only embedded server and check what sockets they open? Should be zero, but it's worth checking whether we let other things slip through like this one did.

luck3y:
@bstansberry Yep, good idea.

luck3y:
@bstansberry FYI, using both embed-server and embed-host-controller, netstat -anp shows only the stream in use by jboss-cli.sh. So this looks fine.

bstansberry:
That's good.

PR #2549, opened 2017-06-15
WFCORE-2959 Export a low MALLOC_ARENA_MAX value in standalone.conf an…

ctomc:
…d domain.conf

### DIFF
--- a/core-feature-pack/src/main/resources/configuration/host/host-master.xml
+++ b/core-feature-pack/src/main/resources/configuration/host/host-master.xml
@@ -76,6 +76,9 @@
                 <option value="-XX:MetaspaceSize=96m"/>
                 <option value="-XX:MaxMetaspaceSize=256m"/>
             </jvm-options>
+            <environment-variables>
+                <variable name="MALLOC_ARENA_MAX" value="1"/>
+            </environment-variables>
         </jvm>
     </jvms>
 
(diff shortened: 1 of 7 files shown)


core-feature-pack/src/main/resources/configuration/domain/template.xml
```diff
@@ -77,6 +77,9 @@
         <server-group name="main-server-group" profile="default">
             <jvm name="default">
                 <heap size="64m" max-size="512m"/>
+                <environment-variables>
+                    <variable name="MALLOC_ARENA_MAX" value="1"/>
+                </environment-variables>
```
bstansberry: This should go in the host.xml, host-master.xml and host-slave.xml templates rather than here. Here it only applies to servers in this group. The expectation is the jvm setting in server-group is just to override some settings from host.xml. Having one in our standard domain.xml is really just an illustration.

Leaving it here too is ok as a second line of defense if you want.

Note that this change doesn't satisfy any requirement in the JBEAP as the templates in full-feature-pack are the ones that are relevant to the EAP dist and the full WildFly dist. So changes in full are needed too.

n1hility:
@ctomc you planning on doing an update to this one?

ctomc:
@n1hility updated & fixed this PR

rpelisse:
Issue: https://issues.jboss.org/browse/JBEAP-11565
Upstream Issue: https://issues.jboss.org/browse/WFCORE-2959

No downstream PR required.

PR #2550, opened 2017-06-15
[WFCORE-2964] Use the default-server-logging.properties file on initi…

jamezp:
…al boot of domain servers. Ignore the logging.configuration system property which should always be set to the incorrect file by the domain startup scripts.

This could be considered a slight change in behavior if a user has modified the startup scripts, e.g. `domain\.(sh|bat|ps1)`. However on Windows there is an issue when comparing the system property, `logging.configuration`, and the resolved `%JBOSS_CONFIG_DIR%\logging.properties`. The system property is always set to `/logging.properties` and the one created internally uses `\logging.propeties`, with the correct file separator.

It seems okay to ignore the system property as it would be a very special use case and ONLY used on the initial boot assuming the logging subsystem is present.

### DIFF
--- a/host-controller/src/main/java/org/jboss/as/host/controller/ManagedServerBootCmdFactory.java
+++ b/host-controller/src/main/java/org/jboss/as/host/controller/ManagedServerBootCmdFactory.java
@@ -35,6 +35,7 @@ import static org.jboss.as.controller.descriptions.ModelDescriptionConstants.VAL
 import static org.jboss.as.host.controller.logging.HostControllerLogger.ROOT_LOGGER;
 
 import java.io.File;
+import java.net.MalformedURLException;
 import java.util.ArrayList;
 import java.util.Arrays;
 import java.util.HashMap;
@@ -64,7 +65,6 @@ import org.jboss.as.server.controller.resources.SystemPropertyResourceDefinition
 import org.jboss.as.server.mgmt.domain.HostControllerConnectionService.SSLContextSupplier;
 import org.jboss.dmr.ModelNode;
 import org.jboss.dmr.Property;
-import org.wildfly.security.manager.WildFlySecurityManager;
 
 /**
  * Combines the relevant parts of the domain-level and host-level models to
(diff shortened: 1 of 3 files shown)


host-controller/src/main/java/org/jboss/as/host/controller/ManagedServerBootCmdFactory.java
```diff
@@ -300,20 +299,17 @@ public HostControllerEnvironment getHostControllerEnvironment() {
...
+            // If the default file exists, it should but we should be safe in case it was deleted by a user
+            if (defaultConfigFile.exists()) {
                 path = "file:" + defaultConfigFile.getAbsolutePath();
-            } else if (systemPropConfigFile != null) {
-                path = systemPropConfigFile;
             } else {
-                // Default to the domain/configuration/logging.properties if nothing else found
+                // Default to the domain/configuration/logging.properties if the default wasn't found to get initial
+                // boot logging
                 path = domainConfigFile;
```
bstansberry: What if this doesn't exist because the user specified something else and deleted the standard one? Will the server boot?

jamezp: Yes though it will print an error to `System.err`: `Unable to read the logging configuration from '' (e.getMessage())`. We could check before adding the `logging.configuration` system property.

Or maybe we just log an error on the HC though I guess it wouldn't have logging either since there was no config file.

jamezp:
retest this please

host-controller/src/main/java/org/jboss/as/host/controller/ManagedServerBootCmdFactory.java
bstansberry: If it doesn't exist why not check the system property and use that if set?

jamezp: No reason other than for some reason I didn't think about it :). Once I'm done with my current change set I'll modify the commit.

jamezp: This should be handled better now. It will attempt to log an error if no configuration file could be found which may or may not work depending on how the HC logging is configured.

It also uses a file to convert the property to a real URL rather than attempting to create one with a simple `file:` prefix.

jamezp:
Failure is legitimate.

PR #2552, opened 2017-06-16
[WFCORE-2368] Set default value of filter-name in Elytron ldap-realm …

danilAREFY:
…in management model
https://issues.jboss.org/projects/WFCORE/issues/WFCORE-2368


bstansberry:
this is ok to test

bstansberry:
@darranl please approve

kabir:
@danilAREFY Are you looking into these failures?

yersan:
@bstansberry , @kabir 

Since @danilAREFY seems is not working on this issue, in my opinion, can be safely closed. It should be replaced by https://github.com/wildfly/wildfly-core/pull/2642. The issue here is: the default value is dependent on another attribute in the realm. These attributes are not good candidates to get a constant default values using set-default, so we have added the default in the attribute description.

bstansberry:
Closing per the comment about about #2642.

PR #2557, opened 2017-06-20
[WFCORE-1632] Interrupt Xnio worker tasks when the worker is stopped

bstansberry:
https://issues.jboss.org/browse/WFCORE-1632

@gaol @dmlloyd @stuartwdouglas 

This is a continuation of the discussion of #1654, converting my experimental branch that was being looked at there into a PR.

This implements the approach discussed in the other PR of doing a shutdownNow on the xnio worker. It addresses, at least somewhat, the concern that tasks already assigned to the worker should be run by passing them off to the management executor to run.

In the end though the management executor needs to ensure tasks are finished too, so the stop logic for it also eventually does a shutdownNow, and it just discards any returned tasks. So, it's possible some tasks won't run.

The shutdown stuff is a bit kludgy in that a simple shutdown() is performed, and then a wait up to magic number 100ms will happen before the stop impl moves on the shutdownNow. So in flight requests are given a bit of a grace period to complete before their thread is interrupted. If no requests are in flight, e.g. if a graceful shutdown has happened and all active requests completed, then the wait is basically instantaneous. Perhaps the magic number could be a bit less.

I added a second commit today to what was in my previous experiment @gaol looked at in the comments on the other PR. I added a 0ms timeout suspend to the server reload handling. This is the right thing to do regardless IMHO. (In core 4.0 we intend to make the timeout configurable.) I wanted to add it here because triggering a suspend and then immediately stopping will at least result in new requests that come in perhaps getting the suspend response rather than getting into the normal logic and then getting a 500 when the thread is interrupted.

### DIFF
--- a/host-controller/src/main/java/org/jboss/as/host/controller/HostControllerService.java
+++ b/host-controller/src/main/java/org/jboss/as/host/controller/HostControllerService.java
@@ -45,6 +45,7 @@ import java.util.concurrent.TimeUnit;
 import org.jboss.as.controller.CapabilityRegistry;
 import org.jboss.as.controller.ControlledProcessState;
 import org.jboss.as.controller.ProcessType;
+import org.jboss.as.host.controller.logging.HostControllerLogger;
 import org.jboss.as.remoting.HttpListenerRegistryService;
 import org.jboss.as.remoting.management.ManagementRemotingServices;
 import org.jboss.as.server.BootstrapListener;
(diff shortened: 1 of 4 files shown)


dmlloyd:
Still non-ideal; a thing is by definition not properly shut down if there are still tasks running.  But, I accept that the analysis to solve this in the "proper" way is probably impossible to achieve at this point.  At least by running the tasks in another executor, they have some chance of finishing, but I still worry that there can be inconsistencies due to services which are expected to have been stopped but are still executing tasks which may be assumed to be completed by dependents.

bstansberry:
@dmlloyd I agree this isn't ideal. Services that depend on the worker would need to track their tasks and cancel them before stopping. That's doable but has a perf impact, and is less self-contained.

I think that's kind of an improvement on this rather than a replacement though. In the end when the worker service stops it should handle any active tasks the worker has for some reason. Ideally there would be none and this just becomes a second line of defense.

bstansberry:
I'm going to add the hold label since besides the failures in full above, locally I'm seeing testsuite/manualmode failures. I was impatient and pushed yesterday 1/2 through testsuite/manualmode.

bstansberry:
I'm going to repush without the suspend-on-reload commit to see the effect on the testsuite.

bstansberry:
retest this please

bstansberry:
retest this please

jmesnil:
@bstansberry what is the status of this PR and #1654?  They have been open for a long time for experimentation.
Are you still planning to merge them?

bstansberry:
@jmesnil If you agree I think this should be merged. It's not experimental; it's just been a subject of discussion and thus fell off the radar screen. But the discussion is not really saying what this PR does is wrong; it's more about how it's a second line of defense that wouldn't be necessary if services were tracking their own tasks and stopping their tasks themselves as part of service stop. I don't see that "stop tasks yourself" approach happening though, because of the perf impact and because I don't see anyone who would be willing to or have time to implement it.

If we merge it it should be done pretty early in a release cycle, e.g. in the next week or so if wanted in 16.

PR #2560, opened 2017-06-20
[WFCORE-2978]: Verify that WFCORE-2923 fix is valid.

ehsavoie:
* Adding tests.
* Adding capability dependencies to ensure that the credential store is defined.
* Adding a dummy service to associate to the credential store.

Jira: https://issues.jboss.org/browse/WFCORE-2978
JBEAP: https://issues.jboss.org/browse/JBEAP-11387


domain-management/src/main/java/org/jboss/as/domain/management/audit/SyslogAuditLogHandlerResourceDefinition.java
```diff
@@ -277,6 +281,35 @@ private static String resolveAppName(OperationContext context, ModelNode appName
...
+        }
+
+        @Override
+        protected boolean requiresRuntime(OperationContext context) {
+            return true;
+        }
+
+        @Override
+        protected void performRuntime(OperationContext context, ModelNode operation, ModelNode model) throws OperationFailedException {
+            String name = Util.getNameFromAddress(operation.require(OP_ADDR));
```
bstansberry: No need to change this but this should be String name = context.getCurrentAddressValue();

PR #2561, opened 2017-06-20
[WFCORE-2704] PersistentResourceXMLDescription should by default mars…

bstansberry:
…hal configured attribute values that match the default

Reopening of #2352.

Requires https://github.com/wildfly/wildfly/pull/10219 and https://github.com/jbossas/jboss-eap7/pull/2035 to remove a no longer relevant test that explicitly requires the incorrect behavior WFCORE-2704 is fixing.


bstansberry:
@ctomc I did a build of full with this, started a DC, added a system prop and then removed it, and then compared domain.xml to the original one we ship. There are no differences related to persistence of default values. There are differences in the elytron subsystem but those look to be because our template has some elements in a different order from how we persist. Primarily "provider-sasl-server-factory" does not follow "provider-loader" when we persist. So, in terms of negative effects of this on how we persist our standard configs, I see none.

kabir:
Retest this please

PR #2564, opened 2017-06-21
[WFCORE-2880] The remove-alias operation should fail when attempting to remove a non-existing alias from a credential store

fjuma:
https://issues.jboss.org/browse/WFCORE-2880
https://issues.jboss.org/browse/JBEAP-11189


fjuma:
Retest this please

PR #2566, opened 2017-06-21
[WFCORE-2359]: Failure message when an op requires a non-existent capability doesn't list registration points for dynamic caps

ehsavoie:
if no possible registration point exists, looks for dynamic capabilities that might be a candidate and returns those registration points.

Jira: https://issues.jboss.org/browse/WFCORE-2359


ehsavoie:
retest this please

PR #2567, opened 2017-06-21
[WFCORE-2958] Only remove default named handlers (those created via t…

jamezp:
…he formatter attribute) if a named-formatter attribute is defined. Also ensure that only a formatter or named-formatter attribute can be defined.

https://issues.jboss.org/browse/WFCORE-2958

Note the removal of allowing a `formatter` or `named-formatter` to overwrite each other on an undefine/write operation. Other subsystems fail the operation if alternatives are both defined. There is a fix in here which will now fail operations that have both attributes defined.

It's possible this should be in a separate commit, but with the changes to how the attributes are handled it seemed okay. to include it.

Note too I'm also using the `ContollerLogger` only to get the same messages as an `add` op would or other ops if multiple attributes are defined that are alternatives for each other.

### DIFF
--- a/logging/src/main/java/org/jboss/as/logging/HandlerOperations.java
+++ b/logging/src/main/java/org/jboss/as/logging/HandlerOperations.java
@@ -425,21 +425,6 @@ final class HandlerOperations {
                 // Undefine the filter-spec
                 model.getModel().get(CommonAttributes.FILTER_SPEC.getName()).set(filterSpecValue);
             }
-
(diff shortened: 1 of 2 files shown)


jamezp:
Hmm... ...kind of weird that two log related tests failed however these are audit log tests which this should touch at all. retest this please.

jamezp:
retest this please

jamezp:
retest this please

kabir:
Retest this please

jamezp:
retest this please

PR #2577, opened 2017-06-23
WFLY-6974 improve transformers testing framework + remoting transformers

ctomc:


remoting/subsystem/src/main/java/org/jboss/as/remoting/RemotingTransformers.java
```diff
@@ -90,19 +93,56 @@ private void buildTransformers_2_1(ResourceTransformationDescriptionBuilder buil
...
+        Different 'alternatives' for parameter 'protocol' of operation 'add'. Current: ["authentication-context"]; legacy: undefined
+        Different 'alternatives' for parameter 'security-realm' of operation 'add'. Current: ["authentication-context"]; legacy: undefined
+        Different 'alternatives' for parameter 'username' of operation 'add'. Current: ["authentication-context"]; legacy: undefined
+         */
+
+        builder.addChildResource(RemotingEndpointResource.ENDPOINT_PATH)
+                .getAttributeBuilder().setValueConverter(new AttributeConverter.DefaultAttributeConverter() {
+            @Override
+            protected void convertAttribute(PathAddress address, String attributeName, ModelNode attributeValue, TransformationContext context) {
+                if ("remote".equals(attributeValue.asString())) {
```
kabir: I have spoken to @dmlloyd and 'remote' and 'remoting' are two separate implementations, so we should not replace this behind the scenes.

Instead, I would 
a) reject this if it is an expression 
b) introduce 'remote' with the converter if the attribute is **undefined**

If 'remote' is not a valid value on 6.2 - 7.0 then we need to do something else :)

bstansberry: Why reject an expression? The attribute allowed expressions in ModelVersion 3

But +1 re undefined. A change in default version is described here:

https://docs.jboss.org/author/display/WFLY/Domain+Mode+Subsystem+Transformers#DomainModeSubsystemTransformers-Attributehasadifferentdefaultvalue

The transformer converts undefined to the new default value so the legacy slave uses the same value as new slaves.

kabir: My rejecting expressions is based on if this attribute can have a value that is not allowed on the legacy slave. If that is not the case, you're correct and we should not reject

bstansberry: I don't think we should reject expressions whenever an attribute allows a new value. In these kinds of situations, the transformers aren't adding a whole lot of value:

1) If the legacy server is joined to the domain, they provide a hopefully better error message than what the legacy server would provide. Nice, but the op fails regardless.
2) If the legacy server is not connected to the domain and people are foolish enough to put their configs into production without testing them first, the transformer is a guard against a bad config that will fail later when the legacy server joins.

I don't think either of those is a big enough benefit that we should reject things that were previously legal and probably are fine just because we can't prove in advance that the user hasn't screwed up.

kabir: ok, so introducing remote if the attribute is undefined (and remote works on 6.2.0+) that should be fine

ctomc: so given that 6.2 also supports "remote" we don't need to transform this attribute at all.
I will update transformer to do so.

ctomc:
@kabir PR updated

kabir:
It seems I have permissions to push to your branch now, so I have updated it :)

kabir:
The problem is the chained converter to 7.0 adds a value for the sasl-protocol attribute, but then this attribute is rejected in previous versions. The converter can add an attachment to the transformation context if it adds the attribute, and if that is present later in the chain the attribute can be discarded.

However, it seems we need to know what the actual value was on older slaves

PR #2583, opened 2017-06-27
[WFCORE-3015] / [WFCORE-3016] Upgrade WildFly Elytron to 1.1.0.Beta55 and Expose principal-decoder as principal-transformer

darranl:
https://issues.jboss.org/browse/WFCORE-3015
https://issues.jboss.org/browse/WFCORE-3016

https://issues.jboss.org/browse/JBEAP-11801
https://issues.jboss.org/browse/JBEAP-11084

### DIFF
--- a/elytron/src/main/java/org/wildfly/extension/elytron/PrincipalDecoderDefinitions.java
+++ b/elytron/src/main/java/org/wildfly/extension/elytron/PrincipalDecoderDefinitions.java
@@ -175,7 +175,7 @@ class PrincipalDecoderDefinitions {
                 final List<String> requiredOids = REQUIRED_OIDS.unwrap(context, model);
                 requiredOids.addAll(REQUIRED_ATTRIBUTES.unwrap(context, model).stream().map(name -> OidsUtil.attributeNameToOid(OidsUtil.Category.RDN, name)).collect(Collectors.toList()));
 
-                return () -> new X500AttributePrincipalDecoder(oid, joiner, startSegment, maximumSegments, reverse, requiredOids.toArray(new String[requiredOids.size()]));
+                return () -> new X500AttributePrincipalDecoder(oid, joiner, startSegment, maximumSegments, reverse, false, requiredOids.toArray(new String[requiredOids.size()]));
             }
 
         };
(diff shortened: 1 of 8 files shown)


elytron/src/main/java/org/wildfly/extension/elytron/PrincipalDecoderDefinitions.java
```diff
@@ -247,30 +258,48 @@ public void registerAttributes(ManagementResourceRegistration resourceRegistrati
...
+        static {
+            Set<RuntimeCapability> capabilities = new HashSet<>(2);
+            capabilities.add(PRINCIPAL_DECODER_RUNTIME_CAPABILITY);
+            capabilities.add(PRINCIPAL_TRANSFORMER_RUNTIME_CAPABILITY);
+            CAPABILITIES = capabilities;
+        }
 
         private PrincipalDecoderAddHandler(AttributeDefinition ... attributes) {
-            super(PRINCIPAL_DECODER_RUNTIME_CAPABILITY, attributes);
+            super(CAPABILITIES, attributes);
```
bstansberry: This is ok but we should rework BaseAddHandler to get rid of the need to pass capabilities to the add handler constructor. AFAICT the only reason BaseAddHandler exists is for this:

```
runtimeCapabilities.forEach(r -> context.registerAdditionalCapabilityRequirement(ELYTRON_CAPABILITY, r.isDynamicallyNamed() ? r.getDynamicName(pathValue) : r.getName(), null));
```

That would not be necessary if the various capabilities simply did

```
static final RuntimeCapability<Void> PRINCIPAL_DECODER_RUNTIME_CAPABILITY =  RuntimeCapability
        .Builder.of(PRINCIPAL_DECODER_CAPABILITY, true, PrincipalDecoder.class)
           .addRequirements(ELYTRON_CAPABILITY)
        .build();
```

darranl: PrincipalDecoderDefinitions is being removed entirely after the next tag, this is just an incremental change to minimise impact to the console.  Debugging another issue we realise the principal decoders are an obsolete concept.

bstansberry: OK, but I'll file an issue anyway. My comment is applicable across the subsystem.

darranl: Yeah went through a few iterations, TBH 99% can be converted to TrivialResourceDefinition and TrivialAddHander anything that looks like this is likely from early in the development.

darranl: BTW there may be a WFCORE issue out there for more conversion.

PR #2591, opened 2017-06-28
[WFCORE-3024] - SecurityDomain is lost when setting up remote outbound connections

pedroigor:
### References

* https://issues.jboss.org/browse/WFCORE-3024
* https://issues.jboss.org/browse/JBEAP-11377

### DIFF
--- a/remoting/subsystem/src/main/java/org/jboss/as/remoting/RemoteOutboundConnectionService.java
+++ b/remoting/subsystem/src/main/java/org/jboss/as/remoting/RemoteOutboundConnectionService.java
@@ -25,6 +25,7 @@ package org.jboss.as.remoting;
 import java.net.URI;
 import java.net.URISyntaxException;
 import java.security.GeneralSecurityException;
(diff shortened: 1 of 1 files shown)


bstansberry:
It looked good to me before I noticed the failing tests!

dmlloyd:
The change seems OK but there are test failures.

dmlloyd:
Probably needs to be coordinated with https://github.com/wildfly/wildfly/pull/10256

pedroigor:
+1. I think that may be the cause.

pedroigor:
FYI, I found the issue. The failing tests are related with scenarios where client configuration in deployment is supposed to be discarded in favor of server configuration. I need to check with QE if the test is really correct or if is considering some wrong behavior fixed by this issue.

pedroigor:
After investigating the failing tests, I've found that my previous changes were not handling properly some scenarios with legacy configuration. The issues were related with using a default authentication context when a connector is not configured explicitly with one.

PR #2602, opened 2017-07-07
[WFCORE-3050] Upgrade jboss-logmanager from 2.0.6.Final to 2.0.7.Final.

jamezp:
https://issues.jboss.org/browse/WFCORE-3050

This only contains two fixes. The fix for [LOGMGR-163](https://issues.jboss.org/browse/LOGMGR-163) may be needed as if weekly log rotation is used log messages will be lost.


n1hility:
retest this please

PR #2611, opened 2017-07-12
[WFCORE-3062] Add access constraints information (if available) to th…

sguilhen:
…e value types description.


bstansberry:
I have my doubts about this, as I believe RBAC enforcement is on attributes, not on fields therein.

IOW I think putting the constraints on the fields is wrong.

bstansberry:
I'm closing this; see my comment on the JIRA. Adding an RBAC constraint to a field in a top level attribute is meaningless, so reporting it would just make the problem worse.

PR #2615, opened 2017-07-17
[WFCORE-3066] Elytron subsystem - wrong description of ldap-realm.otp-credential-mapper.hash-from attribute

yersan:
Jira issues:
https://issues.jboss.org/browse/WFCORE-3066
https://issues.jboss.org/browse/JBEAP-12137


yersan:
Retest this please

yersan:
> The LocalDescriptions texts are just weird.

 Reading my own descriptions again, I can only say I completely agree.

@bstansberry Not only OTP attribute description have the same "problem" with the verb 'map', but also other attribute descriptions, see for example x509CredentialMapperType in xsd. 

I am more inclined to think that map is correct here since it is being broadly used for other LDAP attributes, and reading them again, I understand them as an assignement between the value in the attribute and the value to be used in LDAP server, but I'm not a native English speaker so my opinion shouldn't be taken into account at all.

Before updating this PR, I will wait until see other opinions. If we have to use "to use as the" instead of "map", I will update the other descriptions according.

bstansberry:
Using "map" is ok with me too. I don't want to make too big deal about that when I know people have lots of far more critical stuff to focus on.

yersan:
@bstansberry PR updated. I also found some additional minor typo errors, I suppose we can merge all together in the same PR. They are really minor corrections.

yersan:
Retest this please

yersan:
Retest this please

PR #2618, opened 2017-07-18
[WFCORE-3070] Found multiple secret keys sharing same CKA_LABEL

pskopek:
Synchronization issue when using FIPS enabled NSS DB as PKCS11 keystore.
https://issues.jboss.org/browse/WFCORE-3070
https://issues.jboss.org/browse/JBEAP-11693

### DIFF
--- a/server/src/main/java/org/jboss/as/server/mgmt/domain/HostControllerConnectionService.java
+++ b/server/src/main/java/org/jboss/as/server/mgmt/domain/HostControllerConnectionService.java
@@ -34,7 +34,6 @@ import java.security.KeyStore;
 import java.security.KeyStoreException;
 import java.security.NoSuchAlgorithmException;
 import java.security.cert.CertificateException;
-import java.security.cert.X509Certificate;
 import java.util.concurrent.ExecutorService;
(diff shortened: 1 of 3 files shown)


server/src/main/java/org/jboss/as/server/mgmt/domain/HostControllerConnectionService.java
```diff
@@ -206,23 +204,15 @@ private static SSLContext getAcceptingSSLContext() {
          * to be established.
          */
         try {
+            TrustManagerFactory trustManagerFactory = TrustManagerFactory.getInstance("SunX509");
```
mchoma: I wonder, will that work on IBM java??

pskopek:
@mchoma I was getting exception that in FIPS mode only standard trust manager can be used.

mchoma:
@pskopek , so probably you was hitting WFCORE-1135? Suficient solution to this was introduction of possibility to configure server fips compliant ssl context.  

Havent't had that custom trust manager implementation its place? I mean that empty check* methods? 

@darranl, can you have a look?

pskopek:
@mchoma the problematic commit has been reverted

PR #2620, opened 2017-07-18
Fix for WFCORE-3072. Fix CapabilityRegistry.suggest-capabilities on Windows

jfdenise:
Replace ':' character with _ when converting into Path. Modified unit test to make some values to contain ':' character.


bstansberry:
This seems like an ok close-to-the-deadline workaround but we should not resolve the JIRA based on this.

java.nio.files.Paths is being misused in this code. Capability names are not filesystem paths.

bstansberry:
I opened WFCORE-3074 to get rid of use of Paths.

PR #2621, opened 2017-07-18
WFCORE-2968 - Servers in a domain won't boot if local auth is disabled on the host controller

luck3y:
This adds a new internal realm for servers to auth to the HC using the authKey generated when they are started. Local auth is also disabled on the client for the connection back to the HC.

@darranl @bstansberry 

This got a bit involved and a bit hacky, but it is working for authenticating servers with local auth removed. It still needs a little more testing, but I'm sure that people will have feedback on some of the more "adventurous* bits.

In particular, it would be great if anyone had suggestions on anyu of the following:

- The DMCS linking of the serverinventory callback to the new DomainManagedServerCallbackHandler is super hacky and using a service name lookup etc, but I don't see a common way of doing this in the code. Most of the existing realms are static or being loaded from either property files that already have a fixed location, or something in the domain model. I thought about exposing the server keys in the domain model, but that seemed excessive for this use case.

 - Installing DomainManagedServerCallbackHandler
(https://github.com/wildfly/wildfly-core/compare/master...luck3y:WFCORE-2968?expand=1#diff-73aeedc480f0fb4506e791c2aeff6a00R144) 
Is there a better spot to install this realmcallbackhandler from? It only needs to be installed once, so its checking to see if it's already there in a hacky way.

- RealmName restrictions.
 I need to verify if this was an existing problem or not, but with DomainManagedServerCallbackHandler installed I can auth to the management console using the server credentials. This should be restricted to the internal realm name (https://github.com/wildfly/wildfly-core/compare/master...luck3y:WFCORE-2968?expand=1#diff-73aeedc480f0fb4506e791c2aeff6a00R83), but I always seem to get ManagementRealm.

### DIFF
--- a/domain-management/src/main/java/org/jboss/as/domain/management/security/SecurityRealmAddHandler.java
+++ b/domain-management/src/main/java/org/jboss/as/domain/management/security/SecurityRealmAddHandler.java
@@ -160,6 +160,11 @@ public class SecurityRealmAddHandler extends AbstractAddStepHandler {
             addPlugInLoaderService(realmName, plugIns, serviceTarget);
         }
         InjectedSetValue<CallbackHandlerService> injectorSet = securityRealmService.getCallbackHandlerService();
+
+        // install the managed server realm callback handle for domain server authkeys
+        DomainManagedServerCallbackHandler dmc = new DomainManagedServerCallbackHandler();
+        dmc.installServerSecurityRealm(context, securityRealmService, serviceTarget, realmBuilder, injectorSet.injector());
+
         if (authentication != null) {
             // Authentication can have a truststore defined at the same time as a username/password based mechanism.
             //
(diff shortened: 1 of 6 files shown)


luck3y:
Closing this, will reopen shortly after some additional cleanup..

domain-management/src/main/java/org/jboss/as/domain/management/security/SecurityRealmAddHandler.java
```diff
@@ -160,6 +160,11 @@ protected void installServices(final OperationContext context, final String real
             addPlugInLoaderService(realmName, plugIns, serviceTarget);
         }
         InjectedSetValue<CallbackHandlerService> injectorSet = securityRealmService.getCallbackHandlerService();
+
+        // install the managed server realm callback handle for domain server authkeys
+        DomainManagedServerCallbackHandler dmc = new DomainManagedServerCallbackHandler();
+        dmc.installServerSecurityRealm(context, securityRealmService, serviceTarget, realmBuilder, injectorSet.injector());
```
luck3y: Ok, cool thanks. I've made additional changes, and cleaned it up quite a bit. I have a couple of things left to debug, and I'll reopen.

PR #2622, opened 2017-07-19
[WFCORE-3067] add-alias for credential store might fail when executed…

gaol:
… in the same batch operation with credential store creation

WildFly Core Jira: https://issues.jboss.org/browse/WFCORE-3067
EAP 7.1 Jira: https://issues.jboss.org/browse/JBEAP-11183

In batch mode, the CredentialStoreService may not be `UP` yet when the following up operations like: `add-alias` execute. Give it a chance to check the state after `awaitValue(500, TimeUnit.MILLISECONDS)` should be able to fix it.


gaol:
retest this please

gaol:
retest this please

bstansberry:
This is still racy and can fail, but it should work fine almost always. The only way I see to do better is to an entire new OperationContext.Stage to operation execution, and that's massive overkill for this.

PR #2626, opened 2017-07-20
WFCORE-3509: fix capability name trimming

michpetrov:
Issue: [WFCORE-3059](https://issues.jboss.org/browse/WFCORE-3059)


michpetrov:
Maybe I'm missing something but if simply renaming the default-host wasn't an option for 2690 then the only option is to disallow dots in names. Or have capabilities report both their full and short names but that is imho needlessly complicated.

bstansberry:
We can't break the resource naming rules we've had since AS7 in order to suit a value completion feature in our clients, one that wasn't available in previous releases. We'll have to come back to this in WF Core 4 and see if we can find a way to have the RuntimeCapability itself provide a way to sort this out. Or something. ;)

It's too late in the WF Core 3 release cycle to take that on for 3. We'll be down to fixing nothing but blockers within a week, and we need to focus elsewhere this week.

bstansberry:
I'm going to close this one to help reduce the size of the PR queue, since we know any fix will need to be quite different.

PR #2627, opened 2017-07-20
[WFCORE-3078]: Intermittent failure in AuditLogBootingSyslogTest shows that WFCORE-2923 doesn't cover every case

ehsavoie:
Changing the AuditLogger in the ResultHandler of AuditLogLoggerAddHandler instead of as a RUNTIME step.
Clearing the queue when writting log entries after the ManagedAuditLogger change status to LOGGING.

Jira: https://issues.jboss.org/browse/WFCORE-3078
JBEAP: https://issues.jboss.org/browse/JBEAP-11387

### DIFF
--- a/controller/src/main/java/org/jboss/as/controller/audit/ManagedAuditLoggerImpl.java
+++ b/controller/src/main/java/org/jboss/as/controller/audit/ManagedAuditLoggerImpl.java
@@ -226,6 +226,7 @@ public class ManagedAuditLoggerImpl implements ManagedAuditLogger, ManagedAuditL
                     } catch (Exception e) {
                         handleLoggingException(e);
                     }
+                    queuedItems.clear();
                 }
             } else if (newStatus == Status.DISABLED){
                 queuedItems.clear();
--- a/domain-management/src/main/java/org/jboss/as/domain/management/audit/AuditLogLoggerResourceDefinition.java
+++ b/domain-management/src/main/java/org/jboss/as/domain/management/audit/AuditLogLoggerResourceDefinition.java
@@ -145,6 +145,7 @@ public class AuditLogLoggerResourceDefinition extends SimpleResourceDefinition {
 
             if (auditLoggerProvider != null) {
                 context.addStep(new OperationStepHandler() {
+                    @Override
                     public void execute(final OperationContext context, final ModelNode operation) throws OperationFailedException {
                         final boolean wasReadOnly = auditLoggerProvider.isLogReadOnly();
                         final AuditLogger.Status oldStatus = auditLoggerProvider.getLoggerStatus();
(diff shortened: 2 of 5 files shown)


domain-management/src/main/java/org/jboss/as/domain/management/audit/AuditLogLoggerResourceDefinition.java
```diff
@@ -153,25 +154,11 @@ public void execute(final OperationContext context, final ModelNode operation) t
...
-                                });
-                            }
-                        }, OperationContext.Stage.RUNTIME);
-
-
-                        context.completeStep(new OperationContext.RollbackHandler() {
-                            @Override
-                            public void handleRollback(OperationContext context, ModelNode operation) {
+                            } else {
+                                auditLoggerProvider.setLoggerStatus(oldStatus);
```
bstansberry: You shouldn't need this since if this else block is entered this OSH is not calling setLoggerStatus anywhere else.

testsuite/manualmode/src/test/java/org/jboss/as/test/manualmode/auditlog/AuditLogBootingSyslogTest.java
```diff
@@ -115,8 +115,7 @@ public void testSyslog() throws Exception {
         final BlockingQueue<SyslogServerEventIF> queue = BlockedSyslogServerEventHandler.getQueue();
         queue.clear();
         container.start();
-        long endTime = System.currentTimeMillis() + TimeoutUtil.adjust(5000);
-        waitForExpectedQueueSize(18, queue);
+        waitForExpectedQueueSize(1, queue);
```
bstansberry: Why this change? We shouldn't be changing the number of logged items.

ehsavoie: This is due to the fact that there was a bug in boot : we didn't clear the queue when writing log entries on status change. The 18 was due to leftovers from the previous operations (we didn't understand that at that  time). Now this is coherent with the AuditLogBootingLogTestCase.

ehsavoie:
retest this please

bstansberry:
retest this please

controller/src/main/java/org/jboss/as/controller/audit/ManagedAuditLoggerImpl.java
```diff
@@ -226,6 +226,7 @@ public void setLoggerStatus(final Status newStatus) {
                     } catch (Exception e) {
                         handleLoggingException(e);
                     }
+                    queuedItems.clear();
```
rsvoboda: `queuedItems.clear()` inside `for (AuditLogItem record : queuedItems)`
so only one AuditLogItem is written ?

ehsavoie: Correct :( this is a bug

mchoma: tracked by WFCORE-3109

PR #2628, opened 2017-07-20
[WFCORE-3080] fix t o PS1 script: proper append of script name to the first parameter in the list

pskopek:
https://issues.jboss.org/browse/WFCORE-3080
https://issues.jboss.org/browse/JBEAP-11376

### DIFF
--- a/core-feature-pack/src/main/resources/content/bin/elytron-tool.ps1
+++ b/core-feature-pack/src/main/resources/content/bin/elytron-tool.ps1
@@ -9,8 +9,11 @@ $scripts = (Get-ChildItem $MyInvocation.MyCommand.Path).Directory.FullName;
 $SCRIPT_NAME = $MyInvocation.MyCommand | select -ExpandProperty Name
 $SCRIPT_NAME = "{" + $SCRIPT_NAME + "}"
 
-$ELYTRON_TOOL_OPTS = Process-Script-Parameters -Params $ARGS
-$ELYTRON_TOOL_OPTS = $SCRIPT_NAME + $ELYTRON_TOOL_OPTS 
+$ELYTRON_TOOL_OPTS=@()
+if ($ARGS.Count -gt 0){
+  $ELYTRON_TOOL_OPTS+=$SCRIPT_NAME + $ARGS[0]
+  $ELYTRON_TOOL_OPTS+=$ARGS[1..$ARGS.Count]
+}
 
 $JAVA_OPTS = @()
 


bstansberry:
@ctomc please approve

core-feature-pack/src/main/resources/content/bin/elytron-tool.ps1
```diff
@@ -9,8 +9,11 @@ $scripts = (Get-ChildItem $MyInvocation.MyCommand.Path).Directory.FullName;
 $SCRIPT_NAME = $MyInvocation.MyCommand | select -ExpandProperty Name
 $SCRIPT_NAME = "{" + $SCRIPT_NAME + "}"
 
-$ELYTRON_TOOL_OPTS = Process-Script-Parameters -Params $ARGS
-$ELYTRON_TOOL_OPTS = $SCRIPT_NAME + $ELYTRON_TOOL_OPTS 
+$ELYTRON_TOOL_OPTS=@()
+if ($ARGS.Count -gt 0){
+  $ELYTRON_TOOL_OPTS+=$SCRIPT_NAME + $ARGS[0]
+  $ELYTRON_TOOL_OPTS+=$ARGS[1..$ARGS.Count]
```
ctomc: This looks odd, If I am reading this code right what you are doing is:
if script has any parameters, than you set ELYTRON_TOOL_OPTS to "script name" + all parameters passed to script.
which makes no sense at all.

I think your original problem comes from fact that $ELYTRON_TOOL_OPTS is not declared as array but on the fly which changes the semantics of the concatenation
for examle

`$ELYTRON_TOOL_OPTS = $SCRIPT_NAME + $ELYTRON_TOOL_OPTS `
try to add together string $SCRIPT_NAME and array, which is done by first converting both params to strings and than concatenating them which results in string instead of array that you want.

proper fix should be something like this:
```
$ELYTRON_TOOL_OPTS  = @($SCRIPT_NAME)
$ELYTRON_TOOL_OPTS += Process-Script-Parameters -Params $ARGS
```

pskopek: @ctomc actually the first parameter should contain "{script_name}arg0" and rest of parameters should be the same as they are passed to the script. It works like this across scripts (.bat, .sh) and java code will cope with it.
The reason why is that we can have couple of scripts (.sh, .bat, .sh1) to start the tool and also java -jar as the last possibility.

PR #2630, opened 2017-07-21
[WFCORE-3083] Removing a resource should result in its cap being trac…

bstansberry:
…ked as reload-required

https://issues.jboss.org/browse/WFCORE-3083


bstansberry:
I added the 'hold' label because I intend to add some tests to this. But I wanted to see what effect this had on the testsuite (i.e. any tests that inadvertently rely on the old broken behavior) so I submitted it early to get the CI run.

PR #2426, opened 2017-05-16
[WFCORE-389[WFCORE-2858] Support runtime-only ops on profile resources

bstansberry:
Previously we didn't support runtime-only ops on profile resources, although at least for ops not marked as read-only, they actually work fine, and a couple have gotten in undetected. This PR is about formalizing support for them, mostly by cleaning up some ease of use things and adding some safeguards, plus it adds support for read-only behavior (WFCORE-2858).

https://issues.jboss.org/browse/WFCORE-389
https://issues.jboss.org/browse/WFCORE-2858
https://issues.jboss.org/browse/WFCORE-2815
https://issues.jboss.org/browse/WFCORE-2816
https://issues.jboss.org/browse/WFCORE-2823
https://issues.jboss.org/browse/WFCORE-2849
https://issues.jboss.org/browse/WFCORE-2870
https://issues.jboss.org/browse/WFCORE-2873

This DOES NOT resolve WFCORE-2850; it has a commit that is part one of a two phase change, with the 2nd phase following related changes in full.

PR #2636, opened 2017-07-21
[WFCORE-2850] Part II. Eliminate the temp workaround now that classes…

bstansberry:
… in full no longer require it

https://issues.jboss.org/browse/WFCORE-2850

This completes work started in https://github.com/wildfly/wildfly-core/pull/2426


bstansberry:
I'm withdrawing this. I thought I'd updated the places in full that needed updating to account for this, but I must have dreamed that. The test failures show 2 places that needed update but looking at all the uses of AbstractRuntimeOnlyHandler in full, there are likely others that need it as well and the number of them make me question whether disabling the op in a runtime-only server is the right thing to do.

PR #2640, opened 2017-07-24
[WFCORE-3093] Unclear failure description when removing elytron polic…

xstefank:
…y referenced as the default-policy

https://issues.jboss.org/browse/WFCORE-3093

### DIFF
--- a/elytron/src/main/java/org/wildfly/extension/elytron/PolicyDefinitions.java
+++ b/elytron/src/main/java/org/wildfly/extension/elytron/PolicyDefinitions.java
@@ -29,10 +29,12 @@ import java.security.Policy;
 import java.security.Principal;
 import java.security.PrivilegedAction;
 import java.security.acl.Group;
+import java.util.ArrayList;
 import java.util.Collections;
 import java.util.Enumeration;
 import java.util.HashMap;
 import java.util.HashSet;
+import java.util.List;
 import java.util.Map;
 import java.util.Set;
 import java.util.function.Supplier;
@@ -75,6 +77,7 @@ import org.jboss.security.SecurityConstants;
 import org.jboss.security.auth.callback.CallbackHandlerPolicyContextHandler;
 import org.jboss.security.jacc.SubjectPolicyContextHandler;
 import org.wildfly.common.Assert;
+import org.wildfly.extension.elytron._private.ElytronSubsystemMessages;
 import org.wildfly.security.auth.principal.NamePrincipal;
 import org.wildfly.security.auth.server.SecurityDomain;
 import org.wildfly.security.auth.server.SecurityIdentity;
@@ -97,6 +100,9 @@ import org.wildfly.security.manager.WildFlySecurityManager;
 class PolicyDefinitions {
 
     private static final String DEFAULT_POLICY_NAME = "policy";
+    private static final SimpleAttributeDefinition NAME_ATTRIBUTE_DEFINITION = new SimpleAttributeDefinitionBuilder(ElytronDescriptionConstants.NAME, ModelType.STRING, false)
+            .setMinSize(1)
+            .build();
 
     // providers
     static final SimpleAttributeDefinition DEFAULT_POLICY = new SimpleAttributeDefinitionBuilder(ElytronDescriptionConstants.DEFAULT_POLICY, ModelType.STRING, true)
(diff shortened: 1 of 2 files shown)


elytron/src/main/java/org/wildfly/extension/elytron/PolicyDefinitions.java
```diff
@@ -229,14 +230,22 @@ private Policy getPolicy() {
...
                         ModelNode model = resource.getModel();
                         String defaultPolicy = context.getCurrentAddress().getLastElement().getValue();
 
                         if (model.hasDefined(ElytronDescriptionConstants.DEFAULT_POLICY)) {
                             defaultPolicy = ElytronExtension.asStringIfDefined(context, DEFAULT_POLICY, model);
                         }
 
+                        if (attributeName.equals(JACC_POLICY) || attributeName.equals(CUSTOM_POLICY)) {
+                            if (checkPolicyProviderRemoved(context, defaultPolicy, newValue, oldValue)) {
+                                throw new OperationFailedException("Cannot remove the default policy provider [ " + defaultPolicy + " ]");
```
bstansberry: This needs to be i18n

bstansberry: There's a similar problem at L177 of the existing code too. And L271, L335 and L453.

elytron/src/main/java/org/wildfly/extension/elytron/PolicyDefinitions.java
```diff
@@ -248,6 +257,24 @@ protected void validateUpdatedModel(OperationContext context, Resource resource)
 
     }
 
+    private static boolean checkPolicyProviderRemoved(OperationContext context, String defaultPolicy, ModelNode newValue, ModelNode oldValue) throws OperationFailedException {
+        List<String> newPolicies = getPoliciesNames(context, newValue);
+        List<String> oldPolicies = getPoliciesNames(context, oldValue);
+        return oldPolicies.contains(defaultPolicy) && !newPolicies.contains(defaultPolicy);
```
bstansberry: Nit: there's no point computing newPolicies if !oldPolicies.contains(defaultPolicy).

elytron/src/main/java/org/wildfly/extension/elytron/PolicyDefinitions.java
```diff
@@ -229,14 +230,22 @@ private Policy getPolicy() {
             public void registerAttributes(ManagementResourceRegistration resourceRegistration) {
                 OperationStepHandler write = new ReloadRequiredWriteAttributeHandler(attributes) {
                     @Override
-                    protected void validateUpdatedModel(OperationContext context, Resource resource) throws OperationFailedException {
+                    protected void finishModelStage(OperationContext context, ModelNode operation, String attributeName, ModelNode newValue, ModelNode oldValue, Resource resource) throws OperationFailedException {
+                        super.finishModelStage(context, operation, attributeName, newValue, oldValue, resource);
```
bstansberry: You can drop this since the base class does nothing.

elytron/src/main/java/org/wildfly/extension/elytron/PolicyDefinitions.java
```diff
@@ -229,14 +230,22 @@ private Policy getPolicy() {
...
                             defaultPolicy = ElytronExtension.asStringIfDefined(context, DEFAULT_POLICY, model);
                         }
 
+                        if (attributeName.equals(JACC_POLICY) || attributeName.equals(CUSTOM_POLICY)) {
+                            if (checkPolicyProviderRemoved(context, defaultPolicy, newValue, oldValue)) {
+                                throw new OperationFailedException("Cannot remove the default policy provider [ " + defaultPolicy + " ]");
+                            }
+                        }
+
                         getPolicyProvider(context, model, defaultPolicy, null);
```
bstansberry: Everything in this finishModelStage method needs to be in an added step, i.e. context.addStep((ctx, op) -> {...}, Stage.MODEL);

That was true before as well; this PR just brought it to my attention.

Running the validation in a separate step allows other ops in a composite to "fix" whatever you'd detect as invalid before the validation is run.

xstefank:
@bstansberry no problem, thanks for the review!

PR #2644, opened 2017-07-24
WFCORE-3097 add support for on-demand creation of xml writers

ctomc:
add option to register xml writers on-demand, this reduces creation of xml writers during boot.
In common scenarios when user runs application server in production for longer period also allows that xml writers are never created or after creation (usage) they are able to be GC-ed.

With this change also registration of writer suppliers is now more concurrency friendly and does no longer block in synchronized block.

### DIFF
--- a/controller/src/main/java/org/jboss/as/controller/persistence/AbstractConfigurationPersister.java
+++ b/controller/src/main/java/org/jboss/as/controller/persistence/AbstractConfigurationPersister.java
@@ -22,10 +22,12 @@
 
 package org.jboss.as.controller.persistence;
 
-import java.io.OutputStream;
-import java.util.HashMap;
-import java.util.Map;
+import static org.jboss.as.controller.logging.ControllerLogger.ROOT_LOGGER;
 
+import java.io.OutputStream;
+import java.util.Map;
+import java.util.concurrent.ConcurrentHashMap;
+import java.util.function.Supplier;
 import javax.xml.stream.XMLOutputFactory;
 import javax.xml.stream.XMLStreamWriter;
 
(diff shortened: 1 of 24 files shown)


controller/src/main/java/org/jboss/as/controller/persistence/AbstractConfigurationPersister.java
```diff
@@ -86,13 +89,12 @@ public ModelNode getModelNode() {
 
                     @Override
                     public XMLElementWriter<SubsystemMarshallingContext> getSubsystemWriter(String extensionName) {
-                        synchronized (subsystemWriters) {
-                            return subsystemWriters.get(extensionName);
-                        }
+                        return subsystemWriters.computeIfAbsent(extensionName, name -> subsystemWriterSuppliers.getOrDefault(name, () -> null).get());//lazy create writer, but only once per config serialization
```
bstansberry: subsystemWriters should now be a final local var in marshalAsXml. Then there is no need for the instance field any more and no need for the clear() call. If it's an instance var the "clear" call becomes a possible concurrency issue some day, e.g. concurrent CLI requests to read-config-as-xml.

```
final Map<String, XMLElementWriter<SubsystemMarshallingContext>> subsystemWriters = new HashMap<>();
ModelMarshallingContext extensibleModel = new ModelMarshallingContext() {
```

ctomc: fixed

bstansberry:
retest this please

@ctomc I wonder how much impact this will have on testsuite execution time, since we execute so many write ops in the testsuite.

ctomc:
@bstansberry I don't think it will effect testsuite in any noticeable manner. 
When it comes to writers I did find some allocation hot spots but are more in building / preparing model that is to be passed to writer.
But even that is barely noticeable for some very large subsystem models (like messaging-activemq)

I think any extra time spent in lazy creation of writers could be easily offset by removing just one "reload" from testsuite, as that is much more invasive op.

PR #2645, opened 2017-07-25
Extend Management Client API in TestSuite

mmadzin:
Extend API for getting node's IP address and allows to specify node address and port for management connection.

WFCORE-3098: https://issues.jboss.org/browse/WFCORE-3098

### DIFF
--- a/testsuite/shared/src/main/java/org/jboss/as/test/shared/TestSuiteEnvironment.java
+++ b/testsuite/shared/src/main/java/org/jboss/as/test/shared/TestSuiteEnvironment.java
@@ -33,9 +33,24 @@ public class TestSuiteEnvironment {
      * @return the client
      */
     public static ModelControllerClient getModelControllerClient(final URI authConfigUri) {
+        return getModelControllerClient(authConfigUri, getServerAddress(), getServerPort());
+    }
+
+    /**
+     * Creates a client based on the {@linkplain #getServerAddress() server address} and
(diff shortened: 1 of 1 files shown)


testsuite/shared/src/main/java/org/jboss/as/test/shared/TestSuiteEnvironment.java
```diff
@@ -33,9 +33,24 @@ public static ModelControllerClient getModelControllerClient() {
...
+     * {@linkplain #getServerPort() port}. If the {@code authConfigUri} is not {@code null} the it will be used to
+     * authenticate the client.
+     *
+     * @param authConfigUri the path too the authentication configuration or {@code null}
+     * @param address of the server
+     * @param port the management port of the server
+     *
+     * @return the client
+     */
+    public static ModelControllerClient getModelControllerClient(final URI authConfigUri, String address, int port) {
```
ctomc: It is fine to add new method with this arguments, but old one should still be there as this is public static method, there might be uses outside of just wildfly-core, so removing a method without deprecating it first is dangerous.

mmadzin: Thanks for review, I agree with your comment however I would like to point that the method with original set of parameters is still available (line 35).

bstansberry:
I hadn't noticed this before merging, but in the future please reference the JIRA in the commit message and if practical in the title of the pull request.

PR #2651, opened 2017-07-25
[WFCORE-3104] Missing license files from core-feature-pack-licenses.xml

psakar:
issue: https://issues.jboss.org/browse/WFCORE-3104


bstansberry:
this is ok to test

PR #2652, opened 2017-07-25
[WFCORE-3103] Support cleanup of embedded process' ModuleLoader if th…

bstansberry:
…e ModuleLoader supports it

See also MODULES-301

https://issues.jboss.org/browse/WFCORE-3103

This doesn't really do anything without a JBoss Modules release with MODULES-301 in it but I'm putting it up in case people want to have a look and to see what CI says.

### DIFF
--- a/cli/src/main/java/org/jboss/as/cli/Util.java
+++ b/cli/src/main/java/org/jboss/as/cli/Util.java
@@ -1484,4 +1484,16 @@ public class Util {
                 && mn.has(ATTACHED_STREAMS)
                 && mn.get(ATTACHED_STREAMS).asBoolean();
     }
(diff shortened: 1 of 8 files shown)


bstansberry:
Closing; this doesn't work. The testsuite doesn't pass when a jboss-modules snapshot with the MODULES-301 fix in it is used.

Not sure why, but one thing that's obvious is that EmbeddedProcessFactory.setupModuleLoader using Module.getBootModuleLoader() is not the way to go if we're going to close the ModuleLoader. :)

luck3y:
Interesting. I'd be happy to take a look at this at some point and see what's going on.

PR #2677, opened 2017-07-31
WFCORE-3125 Intermittent failures in ServerAuthenticationTestCase.testDisableLocalAuthAndStartServers

luck3y:
This uses waitUntilState on the master HC to verify the slave server has actually started. Also includes some cleanup in the test.


testsuite/domain/src/test/java/org/jboss/as/test/integration/domain/ServerAuthenticationTestCase.java
```diff
@@ -131,6 +116,7 @@ public void testDisableLocalAuthAndStartServers() throws Exception {
         // now restart the slave HC, but don't restart the servers and verify they've reconnected afterwards
         reloadHostController(master.getDomainClient(), "slave", false);
         slave.awaitHostController(System.currentTimeMillis());
+        waitUntilState(master.getDomainClient(), "slave", "main-three", "STARTED");
         verifyServerStarted(master.getDomainClient(), "slave", "main-three");
```
bstansberry: This shouldn't be necessary since AIUI waitUntilState will fail if the server doesn't get to the right state.

Can you change waitUntilState so it's not waiting a full 1 second between tries? That seems like a long time. Maybe 100ms? 250ms?

testsuite/domain/src/test/java/org/jboss/as/test/integration/domain/ServerAuthenticationTestCase.java
```diff
@@ -141,81 +127,47 @@ private void addUser(final File domainConfigDir, final String realm, final Strin
...
     }
 
     private static void removeLocalAuth(final DomainClient client, final String host) throws Exception {
         // remove local auth, requires restart
-        final ModelNode op = new ModelNode();
-        op.get(OP).set(REMOVE);
-        op.get(OP_ADDR).set(getLocalAuthPath(host));
+        final ModelNode op = Util.createEmptyOperation(REMOVE, getLocalAuthPath(host));
         final ModelNode result = client.execute(op);
         assertEquals(result.toJSONString(true), SUCCESS, result.get(OUTCOME).asString());
```
bstansberry: Why are all these 'toJSONString()"? The JSON format isn't our standard format so this becomes an oddity when people look at failures.

bstansberry:
At this point is it possible to make this part of DomainTestSuite? That would require restoring local auth in an @After.

testsuite/domain/src/test/java/org/jboss/as/test/integration/domain/ServerAuthenticationTestCase.java
luck3y: Yeah, I thought about removing this, but I was feeling a bit conservative and figured it'd be a fast check anyway, so why not. You're right though, its not needed.

testsuite/domain/src/test/java/org/jboss/as/test/integration/domain/ServerAuthenticationTestCase.java
luck3y: Yeah, I thought about removing this, but I was feeling a bit conservative and figured it'd be a fast check anyway, so why not. You're right though, its not needed.

testsuite/domain/src/test/java/org/jboss/as/test/integration/domain/ServerAuthenticationTestCase.java
luck3y: Oh, I thought that was normal for reporting failures where you want to see the details. I'll fix.

luck3y:
I'll see if I can get the local auth restore working reliably, I think I should be able to.

testsuite/domain/src/test/java/org/jboss/as/test/integration/domain/ServerAuthenticationTestCase.java
luck3y: 100ms sounds like a good idea to me.

luck3y:
@bstansberry I think this should be more reliable now. Most of the issues were occurring with slave reloads after local auth was removed and operations getting rolled back due to being executed too early or too late, using a combination of checking for the slave being registered in the domain model and waitUntilState() on the slave servers should make this reliable.

luck3y:
@bstansberry this is good to go now

luck3y:
actually, @bstansberry  never mind, I must have had another push behind that one, so I'll let this one clear first.

luck3y:
retest this please

luck3y:
Swabra cleanup thing is unrelated. Going to claim this as a pass.

PR #2680, opened 2017-08-01
[WFCORE-3133] - [Migration operation] [Web to Undertow] truststore - ke…

JiriOndrusek:
…ystore-password does it really needs to be mandatory?

WFCORE: https://issues.jboss.org/browse/WFCORE-3133

(part of solution for: https://github.com/wildfly/wildfly/pull/10381)

### DIFF
--- a/domain-management/src/main/java/org/jboss/as/domain/management/parsing/ManagementXml_5.java
+++ b/domain-management/src/main/java/org/jboss/as/domain/management/parsing/ManagementXml_5.java
@@ -751,12 +751,6 @@ class ManagementXml_5 extends ManagementXml {
                 }
             }
         }
-        /*
-         * The only mandatory attribute now is the KEYSTORE_PASSWORD.
-         */
-        if (keystorePasswordSet == false) {
-            throw missingRequired(reader, EnumSet.of(Attribute.KEYSTORE_PASSWORD));
-        }
     }
 
     private void parseKerberosIdentity(final XMLExtendedStreamReader reader,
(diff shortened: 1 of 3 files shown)


bstansberry:
Please correct the commit message and the PR title to reflect the WFCORE issue. Thanks.

bstansberry:
I'm not clear on what this has to do with Web to Undertow migration.

bstansberry:
@JiriOndrusek Any update on this one?

JiriOndrusek:
@bstansberry @darranl I'm sorry for the delay (I've missed notifications), I'll look into it.

JiriOndrusek:
@darranl May I ask for clarification of you change request? 
I have tried migration without trusstore password and it works (with the wildfly from master with changes from this PR and linked one) -> password for trusstore is optional.
For the keystore it behaves in the same way, although I haven't found a way of creation keystore with key without password. I was able simulate it with keystore with entry only (in that case there was a problem that there are no keys in store) - but as a prove that it can be defined without password it is enough.

bstansberry:
AIUI the issues are as follows:

1) The xsd for the version that is current when this change is made needs to be updated to reflect the fact the password is not required.

2) The xsd version this applies to is not wildfly_config_5_0.xsd, as 5 had shipped and we don't change it. It would have been 6, which is the version for WF Core 4 / WF 12. But this likely isn't going to make WF Core 4. So it will be wildfly_config_7_0.xsd.

It's ok that even though the  5 and 6 xsds say the attribute isn't optional that the parser for those versions (ManagementXml_5) is lenient and doesn't require it. IOW we shouldn't need a ManagementXml_7 that doesn't require the attribute while ManagementXml_5 would continue to.

JiriOndrusek:
@bstansberry 

I've checked definition of password, and it is optional - https://github.com/wildfly/wildfly-core/blob/master/server/src/main/resources/schema/wildfly-config_6_0.xsd#L652
(I've validate it also as xml against xsd)

So I don't think there is need to ass anything.

bstansberry:
retest this please

@jmesnil Assuming no relevant CI issues this is good to go.

PR #2681, opened 2017-08-02
WFCORE-3132 Creation of RedirectHandler results in Undertow predicate…

stuartwdouglas:
… parser being initialised on server boot


stuartwdouglas:
The stops around 40 classes from being initialised at boot time

PR #2683, opened 2017-08-02
WFCORE-3135 security-manager subsystem has wrong xsd schema

ctomc:
jira: https://issues.jboss.org/browse/WFCORE-3135
https://issues.jboss.org/browse/JBEAP-12532


bstansberry:
This will need an acked Blocker JBEAP. This kind of mistake is a valid Blocker.

ctomc:
created https://issues.jboss.org/browse/JBEAP-12532 for that

security-manager/src/main/resources/schema/wildfly-security-manager_3_0.xsd
```diff
@@ -1,113 +0,0 @@
...
-        <xs:sequence>
-            <xs:element name="deployment-permissions" minOccurs="0" maxOccurs="1" type="deploymentPermissionsConfigType"/>
-        </xs:sequence>
-    </xs:complexType>
-
-    <xs:complexType name="deploymentPermissionsConfigType">
-        <xs:annotation>
-            <xs:documentation>
-                <![CDATA[
-                The deploymentPermissionConfigType specifies the elements and attributes that can be used to configure
```
rsvoboda: pure removal of wildfly-security-manager_3_0.xsd is not the right thing to me.

`<xs:complexType name="deploymentPermissionsConfigType">` and `<xs:complexType name="permissionType">` has better `<xs:documentation>` in XSD version 3.0 comparing to 1.0

Thus I would suggest to also enhance XSD version 1.0 with the diff as part of this PR

bstansberry: +1

ctomc: update PR to use docs from 3.0 schema in 1.0

PR #2686, opened 2017-08-03
[WFCORE-3137] Server stops when creating new elytron policy with wrong default-policy value

ivassile:
wildfly-elytron-integration: https://issues.jboss.org/browse/WFCORE-3137
JBEAP: https://issues.jboss.org/browse/JBEAP-12498

### DIFF
--- a/elytron/src/main/java/org/wildfly/extension/elytron/PolicyDefinitions.java
+++ b/elytron/src/main/java/org/wildfly/extension/elytron/PolicyDefinitions.java
@@ -99,14 +99,12 @@ import org.wildfly.security.manager.WildFlySecurityManager;
  */
 class PolicyDefinitions {
 
-    private static final String DEFAULT_POLICY_NAME = "policy";
     private static final SimpleAttributeDefinition NAME_ATTRIBUTE_DEFINITION = new SimpleAttributeDefinitionBuilder(ElytronDescriptionConstants.NAME, ModelType.STRING, false)
             .setMinSize(1)
             .build();
 
     // providers
     static final SimpleAttributeDefinition DEFAULT_POLICY = new SimpleAttributeDefinitionBuilder(ElytronDescriptionConstants.DEFAULT_POLICY, ModelType.STRING, true)
-            .setDefaultValue(new ModelNode(DEFAULT_POLICY_NAME))
             .setAllowExpression(false)
             .setMinSize(1)
             .setFlags(AttributeAccess.Flag.RESTART_RESOURCE_SERVICES)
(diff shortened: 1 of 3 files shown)


elytron/src/main/java/org/wildfly/extension/elytron/PolicyDefinitions.java
```diff
@@ -146,6 +146,9 @@ static ResourceDefinition getPolicy() {
             @Override
             protected void performRuntime(OperationContext context, ModelNode operation, ModelNode model) throws OperationFailedException {
                 String defaultPolicy = context.getCurrentAddress().getLastElement().getValue();
+                if (model.hasDefined(ElytronDescriptionConstants.DEFAULT_POLICY)) {
+                    defaultPolicy = ElytronExtension.asStringIfDefined(context, DEFAULT_POLICY, model);
+                }
```
bstansberry: The following would be cleaner code, but won't work, because there's another problem that needs to be addressed:

```
String defaultPolicy = ElytronExtension.asStringIfDefined(context, DEFAULT_POLICY, model);
if (defaultPolicy == null) {
    defaultPolicy = context.getCurrentAddressValue();;
}
```

It won't work though. That's because DEFAULT_POLICY has a default value assigned. See L109.

The default value handling must be fixed.

If an AttributeDefinition has a default value configured, then that value must be what is used if the user doesn't define the attribute. But here it's not used at all. Here if the user doesn't define the attribute, the name of the resource is used.

Please check with the elytron team and then do one or the other of the following:

1) Remove the setDefaultValue at L109 and update LocalDescriptions.properties so the description of the default-policy attribute explains that if the attribute is undefined the name of the resource will be used. Then the idiom I used above is ok.

If that's not what the elytron team wants and DEFAULT_POLICY really should have a default value of 'policy', then, 

2)  This stuff about defaultPolicy = context.getCurrentAddress().getLastElement().getValue() should be removed from here and from the write handler (L240 in current code) and should be replaced by a simple

```
String defaultPolicy = DEFAULT_POLICY.resolveModelAttribute(context, model).asString();
```

honza889: From the code I am pretty sure default value should be removed.
(btw, it works, because ElytronExtension.asStringIfDefined, as opposite to DEFAULT_POLICY.resolveModelAttribute, does not resolve default value and returns undefined, default value is ignored)

ivassile:
@bstansberry @honza889 Thank you for the review. I've submitted the requested changes.

bstansberry:
retest this please

PR #2690, opened 2017-08-03
[WFCORE-3114] Wildfly 10.1.0 add-user command

Dvikasred:
Upstream fix for Bug https://issues.jboss.org/browse/WFCORE-3114

### DIFF
--- a/cli/src/main/resources/help/command.txt
+++ b/cli/src/main/resources/help/command.txt
@@ -29,7 +29,7 @@ DESCRIPTION
   Here is an example of invoking operation 'flush-all-connection-in-pool' on
   datasource with jndi-name 'myds'.
 
-  data-source flush-all-connection-in-pool --jndi-name=myds
+  data-source flush-all-connection-in-pool --jndi-name=myds 
 
   where flush-all-connection-in-pool is an operation name exposed by data-source
   type and jndi-name is the property name that identifies the specific
@@ -46,7 +46,13 @@ DESCRIPTION
 
   data-source --jndi-name=my-new-ds --min-pool-size=11 --max-pool-size=22
 
-  Tab-completion will help completing operation and property names.
+  
+ARGUMENTS
+  
+  Arguments should be passed with the commands, in CLI mode command we are having arguments(--help, --help --commands and etc)
+
+  For example, command /subsystem=datasources/data-source.
+  Tab-completion will help completing operation, property names and arguments.
 
   Generic type commands support --help option. The content of the help will be
   generated from the description of the operations and properties provided by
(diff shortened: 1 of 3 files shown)


bstansberry:
The first of these commits is unrelated; please remove it.

Dvikasred:
Thanks for the update,
Do I need to remove  "[WFCORE-3136] Added Generic type CLI commands into docs (CLI guide)" in this pull request ?
OR 
Both the changes I have made is not related ?

bstansberry:
@Dvikasred Please remove the WFCORE-3136 commit.

Dvikasred:
Hi Sure I will do so 
Thanks  a lot for the help

Dvikasred:
Hi,

Thanks for the help. 
As per the request I have revert the "[WFCORE-3136] Added Generic type CLI commands into docs (CLI guide)" Commit. 
Please let me know if you need any support from my side. 

Thanks and Regards,
Vikas Dadhich

darranl:
If I am reading the diff correctly is the proposal to just completely ignore any configured JBOSS_HOME?

Dvikasred:
Yes it is because, While we are running the script.
It is taking the default JBOSS_HOME what set at OS level. 

Once we unset it in script. I will not call the same.

Dvikasred:
It's just like initializing variable with no value. 
I feel which will not create any problem with the existing script(Which is working fine).  
Regarding the warning part. 
In this scenarios we will not face this condition.
Please feel free to share the idea with me. We may find the best one :)
Thanks again for such valuable comments and feedback.

darranl:
I think the opposite question is: -

"What purpose is the code within the script serving now that checks the JBOSS_HOME locations match and emits a warning if the JBOSS_HOME variable is always cleared?"

Dvikasred:
Thanks for the update,
We can either uncomment  and keep it as note or leave it like this. 
This will help others to understand the changes.
Also, If we get some new approach we can use/refer this.

darranl:
My point is, if you take add-user.bat as an example.

Line 21 we discover our local JBOSS_HOME.
Lines 24,25,26 we then check if we have JBOSS_HOME set or use our locally discovered one instead.
Lines 32 to 38 we are comparing the discovered location with the one from the variable and outputting a message if they differ.

If we are going to take the decision to not support using the JBOSS_HOME environment variable then all of this code will need to be cleaned up in the script, otherwise all of this code is redundant and serves no ongoing purpose.

Dvikasred:
I understand this and it was not removed the reason I have mentioned in above comments.
Since this is redundant I will do so. 
Please allow me to go through the test cases in Linux as well as Windows environment. 
thanks again for the feedback.

darranl:
If you are going to have a look at some more testing would it be possible to check why the comparison was not working in the first place?  Maybe that could be double checked to get the error correctly reported.

Dvikasred:
Sure I will check and update the pull request.

Dvikasred:
I have finished testing with Linux.
For windows still it is going on. 
Once everything is done I will commit the changes

darranl:
I suggest we close this one, if the tested changes are available a new PR can be raised.

PR #2692, opened 2017-08-04
[WFCORE-3146] Prevent wasted classloading time and metaspace use for …

bstansberry:
…management parsers

https://issues.jboss.org/browse/WFCORE-3146

Related to JBEAP-9326https://issues.jboss.org/browse/JBEAP-9326 (saves classloading time) and JBEAP-6731https://issues.jboss.org/browse/JBEAP-6731 (memory saving).

### DIFF
--- a/domain-management/src/main/java/org/jboss/as/domain/management/parsing/AuditLogXml.java
+++ b/domain-management/src/main/java/org/jboss/as/domain/management/parsing/AuditLogXml.java
@@ -32,8 +32,8 @@ import org.jboss.staxmapper.XMLExtendedStreamWriter;
  *
  * @author <a href="mailto:ehugonne@redhat.com">Emmanuel Hugonnet</a> (c) 2015 Red Hat, inc.
  */
-public abstract class AuditLogXml {
-    public static AuditLogXml newInstance(Namespace namespace, boolean host) {
+public interface AuditLogXml {
+    static AuditLogXml newInstance(Namespace namespace, boolean host) {
         switch (namespace.getMajorVersion()) {
             case 1:
             case 2:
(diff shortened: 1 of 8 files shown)


domain-management/src/main/java/org/jboss/as/domain/management/parsing/AuditLogXml.java
```diff
@@ -46,17 +46,16 @@ public static AuditLogXml newInstance(Namespace namespace, boolean host) {
...
     }
 
-    AuditLogXml() {
-    }
-
     // TODO - The problem is that it is version dependent but it would be nicer to not have processing instructions and find a
     // better way to decide if a native interface is required.
 
-    public abstract void parseAuditLog(final XMLExtendedStreamReader reader, final ModelNode address, final Namespace expectedNs, final List<ModelNode> list) throws XMLStreamException;
+    void parseAuditLog(final XMLExtendedStreamReader reader, final ModelNode address, final Namespace expectedNs, final List<ModelNode> list) throws XMLStreamException;
```
jamezp: I don't recall for sure, but this might fail the next checkstyle upgrade because the parameters use the `final` qualifier on an interface method.

domain-management/src/main/java/org/jboss/as/domain/management/parsing/ManagementXml.java
```diff
@@ -53,19 +53,18 @@ public static ManagementXml newInstance(Namespace namespace, ManagementXmlDelega
...
     }
 
-    ManagementXml() {
-    }
-
     // TODO - The problem is that it is version dependent but it would be nicer to not have processing instructions and find a
     // better way to decide if a native interface is required.
 
-    public abstract void parseManagement(final XMLExtendedStreamReader reader, final ModelNode address,
+    void parseManagement(final XMLExtendedStreamReader reader, final ModelNode address,
```
jamezp: Same note as above about the `final` qualifier on interface methods.

domain-management/src/main/java/org/jboss/as/domain/management/parsing/AuditLogXml_4.java
```diff
@@ -639,175 +633,4 @@ private void parseAuditLogHandlersReference(final XMLExtendedStreamReader reader
             }
         }
     }
-
-    @Override
-    public void writeAuditLog(XMLExtendedStreamWriter writer, ModelNode auditLog) throws XMLStreamException {
```
jamezp: Should this be removed? It looks like the `StandaloneXml_5` and `HostXml_5` may invoke this and get a `UnsupportedOperationException`.

domain-management/src/main/java/org/jboss/as/domain/management/parsing/AuditLogXml.java
bstansberry: Thanks; I'll clean that up.

domain-management/src/main/java/org/jboss/as/domain/management/parsing/AuditLogXml_4.java
bstansberry: If StandaloneXml_5 or HostXml_5 have a reference to AuditLogXml_4 or AuditLogXml_Legacy, that's a bug and a failure is good. They are constructed with a ref to a Namespace that when passed to AuditLogXml.getInstance(ns) will not result in these impls being returned.

jamezp: Ah okay. That makes sense.

PR #2694, opened 2017-08-04
[WFCORE-3150] Workaround by moving first use of kernel classes outsid…

bstansberry:
…e of the ParallelBootTask

https://issues.jboss.org/browse/WFCORE-3150

Related to https://issues.jboss.org/browse/JBEAP-9326 and https://issues.jboss.org/browse/JBEAP-6731

### DIFF
--- a/controller/src/main/java/org/jboss/as/controller/ParallelBootOperationContext.java
+++ b/controller/src/main/java/org/jboss/as/controller/ParallelBootOperationContext.java
@@ -57,24 +57,27 @@ class ParallelBootOperationContext extends AbstractOperationContext {
 
     private final OperationContextImpl primaryContext;
     private final List<ParsedBootOp> runtimeOps;
-
+    private final Thread controllingThread;
     private Step lockStep;
     private final int operationId;
     private final ModelControllerImpl controller;
 
     ParallelBootOperationContext(final ModelController.OperationTransactionControl transactionControl,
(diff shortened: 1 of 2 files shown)


bstansberry:
FYI for reviewers re: the only thing I wasn't 100% confident in when I submitted this (just maybe 98% confident):

[1:59 PM] Brian Stansberry: @here I'd like a confirmation on something....
[1:59 PM] Brian Stansberry: https://github.com/wildfly/wildfly-core/pull/2694/files#diff-8c34f6a37487f247decdb28409a74575R133
[2:00 PM] Brian Stansberry: previously boot code was creating a SecurityDomain a la the code indicated there once per "ParallelBootTask" in boot
[2:00 PM] Brian Stansberry: ParallelBootTask is a Runnable that executes mgmt ops for a given subsystem during boot, with the various subsystems executing concurrently
[2:01 PM] Brian Stansberry: the PR ^^^ changes stuff and one thing is now that SecurityDomain is created just once for the overall boot, and then all the tasks started right after get a ref to it
[2:02 PM] Brian Stansberry: I see no reason to create > 1 domain
[2:35 PM] Darran Lofthouse: @BrianStansberry I think the domain you are talking about was just temporary until everything else had bootstrapped (we needed some form of identity) so creating a single one should also be fine
[2:36 PM] Brian Stansberry: @DarranLofthouse thanks

PR #2695, opened 2017-08-07
[WFCORE-3151] TrustManager - non-lazy CRL reload

hkalina:
Fixes of:
- errors occurred during reload not visible as operation result
- reload flag is shared across individual TrustManagers (first used TrustManager is reloaded instead of the chosen one)

https://issues.jboss.org/browse/WFCORE-3151
https://issues.jboss.org/browse/JBEAP-11225

### DIFF
--- a/elytron/src/main/java/org/wildfly/extension/elytron/SSLDefinitions.java
+++ b/elytron/src/main/java/org/wildfly/extension/elytron/SSLDefinitions.java
@@ -493,6 +493,10 @@ class SSLDefinitions {
 
     }
 
+    private abstract static class ReloadableX509ExtendedTrustManager extends X509ExtendedTrustManager {
+        abstract void reload();
+    }
+
     static ResourceDefinition getTrustManagerDefinition() {
 
         final SimpleAttributeDefinition providersDefinition = new SimpleAttributeDefinitionBuilder(PROVIDERS)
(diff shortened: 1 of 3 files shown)


bstansberry:
This isn't a blocker and the related issue state is pretty clear this isn't targeted at WF Core 3, so I'm putting the 4.x label on it. Please make some noise if that's wrong.

elytron/src/main/java/org/wildfly/extension/elytron/SSLDefinitions.java
```diff
@@ -710,18 +707,22 @@ public void registerOperations(ManagementResourceRegistration resourceRegistrati
...
+                            protected void executeRuntimeStep(OperationContext context, ModelNode operation) throws OperationFailedException {
+                                ServiceName serviceName = TRUST_MANAGER_RUNTIME_CAPABILITY.fromBaseCapability(context.getCurrentAddressValue()).getCapabilityServiceName();
+                                ServiceController<TrustManager> serviceContainer = getRequiredService(context.getServiceRegistry(true), serviceName, TrustManager.class);
+                                State serviceState;
+                                if ((serviceState = serviceContainer.getState()) != State.UP) {
+                                    throw ROOT_LOGGER.requiredServiceNotUp(serviceName, serviceState);
+                                }
+                                TrustManager trustManager = serviceContainer.getValue();
+                                if (! (trustManager instanceof ReloadableX509ExtendedTrustManager)) {
+                                    throw ROOT_LOGGER.unableToReloadCRLNotReloadable();
```
bstansberry: I think this should throw OFE, not RuntimeException. A RuntimeException means some sort of server fault, and it will lead to an error in the logs. The user trying to perform an operation on a resource not configured to allow that operation is a user mistake, and that's an OFE.

My assumption here is it's valid for trustManager to be something other than ReloadableX509ExtendedTrustManager; i.e. that this check isn't doing something like an "assert trustManager instanceof ReloadableX509ExtendedTrustManager;" would. If it's like an assert then don't throw OFE.

PR #2705, opened 2017-08-09
[WFCORE-3176] Upgrade jboss-logging-tools to version 2.1.0.Final

jamezp:
https://issues.jboss.org/browse/WFCORE-3176


bstansberry:
@jamezp what's the JBEAP? none is linked to the WFCORE

jamezp:
@bstansberry There currently isn't one though I could create it. The only reason I didn't was because this is a build-only tool. It's not shipped with the runtime, however now that I think about WildFly and EAP would inherit this version from Core.

bstansberry:
@jamezp EAP productization has to rebuild everything from source, including stuff that's just used for builds. Maven plugins, everything. At this stage they want JIRAs for anything that's going to change, which is reasonable enough IMHO.

I also don't want to have to think about these PRs and decide whether it's ok to merge it because of this or that reason. I'm lazy and want folks to get their things approved. :)

PR #2706, opened 2017-08-09
WFCORE-3175 The BootLoggingPatchingStateTestCase fails if the message…

aloubyansky:
… assertion is flipped


bstansberry:
@aloubyansky I just merged https://github.com/wildfly/wildfly-core/pull/2704 which disabled some tests in order to work around this. That PR had other necessary stuff too, so I wanted it in. That disabling needs to be undone now, plus it produces merge conflicts with this.

jamezp:
@bstansberry I just spoke with @aloubyansky and I'm going to fix it. I should be able to force push to his branch which should update the PR.

jamezp:
retest this please

jamezp:
FWIW I tested this fix with both English and German and it appeared to work perfectly.

PR #2707, opened 2017-08-10
[WFCORE-3168] warning when setting or removing secret value for alias

pskopek:
https://issues.jboss.org/browse/WFCORE-3168
https://issues.jboss.org/browse/JBEAP-12642

### DIFF
--- a/elytron/src/main/java/org/wildfly/extension/elytron/_private/ElytronSubsystemMessages.java
+++ b/elytron/src/main/java/org/wildfly/extension/elytron/_private/ElytronSubsystemMessages.java
@@ -343,10 +343,10 @@ public interface ElytronSubsystemMessages extends BasicLogger {
     @Message(id = 920, value = "Credential alias \"%s\" of credential type \"%s\" does not exist in the store")
     OperationFailedException credentialDoesNotExist(String alias, String credentialType);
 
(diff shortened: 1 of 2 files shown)


elytron/src/main/java/org/wildfly/extension/elytron/_private/ElytronSubsystemMessages.java
```diff
@@ -343,6 +343,12 @@
     @Message(id = 920, value = "Credential alias \"%s\" of credential type \"%s\" does not exist in the store")
     OperationFailedException credentialDoesNotExist(String alias, String credentialType);
 
+    @Message(id = Message.NONE, value = "Reload dependant services which might already have cached the secret value")
```
bstansberry: s/dependant/dependent

elytron/src/main/java/org/wildfly/extension/elytron/_private/ElytronSubsystemMessages.java
```diff
@@ -343,6 +343,12 @@
...
+    String reloadDependantServices();
+
+    @Message(id = Message.NONE, value = "Update dependant resources as alias \"%s\" does not exist anymore")
```
bstansberry: s/dependant/dependent

bstansberry:
I'll just fix that here.

PR #2709, opened 2017-08-10
[WFCORE-3179] Elytron - Kerberos SASL tests

kwart:
Upstream: https://issues.jboss.org/browse/WFCORE-3179
Downstream: https://issues.jboss.org/browse/JBEAP-11775

Tests for Elytron Kerberos SASL mechanisms - `GSSAPI`, `GS2-KRB5`, `GS2-KRB5-PLUS`.

Replaces:
https://github.com/jbossas/jboss-eap7/pull/2058
https://github.com/wildfly/wildfly/pull/10243

### DIFF
--- a/testsuite/elytron/pom.xml
+++ b/testsuite/elytron/pom.xml
@@ -20,9 +20,22 @@
         <jbossas.ts.dir>${jbossas.ts.integ.dir}/..</jbossas.ts.dir>
         <jbossas.project.dir>${jbossas.ts.dir}</jbossas.project.dir>
         <wildfly.home>${project.basedir}/target/wildfly-core</wildfly.home>
+        <version.javax.enterprise>1.2</version.javax.enterprise>
 
         <ts.elytron.cli>enable-elytron.cli</ts.elytron.cli>
     </properties>
+
+    <dependencyManagement>
+        <dependencies>
+            <dependency>
+                <groupId>org.apache.directory.server</groupId>
+                <artifactId>apacheds-parent</artifactId>
+                <version>${version.org.apache.ds}</version>
+                <scope>import</scope>
+                <type>pom</type>
+            </dependency>
+        </dependencies>
+    </dependencyManagement>
 
     <dependencies>
         <dependency>
@@ -74,6 +87,28 @@
         <dependency>
             <groupId>org.wildfly.core</groupId>
             <artifactId>wildfly-jmx</artifactId>
+        </dependency>
+
+        <dependency>
+            <groupId>org.apache.directory.api</groupId>
+            <artifactId>api-ldap-codec-standalone</artifactId>
+            <scope>test</scope>
+        </dependency>
+        <dependency>
+            <groupId>org.apache.directory.server</groupId>
+            <artifactId>apacheds-interceptor-kerberos</artifactId>
+            <scope>test</scope>
+        </dependency>
+        <dependency>
+            <groupId>org.apache.directory.server</groupId>
+            <artifactId>apacheds-server-annotations</artifactId>
+            <scope>test</scope>
+            <exclusions>
+                <exclusion>
+                    <groupId>org.slf4j</groupId>
+                    <artifactId>*</artifactId>
+                </exclusion>
+            </exclusions>
         </dependency>
     </dependencies>
 
--- a/testsuite/elytron/src/test/java/org/wildfly/test/integration/elytron/sasl/mgmt/AbstractKerberosMgmtSaslTestBase.java
+++ b/testsuite/elytron/src/test/java/org/wildfly/test/integration/elytron/sasl/mgmt/AbstractKerberosMgmtSaslTestBase.java
@@ -1,0 +1,538 @@
+/*
+ * Copyright 2017 Red Hat, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *   http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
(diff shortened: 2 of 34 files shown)


bstansberry:
@kwart This is a lot of code, so please find someone to review it.

@psakar Please note the new (test) dependencies. I don't know if these are new to WildFly/EAP overall.

kwart:
@darranl Could you please review this? This PR comes from the decision to have the coverage in core -
 closer to Elytron. 

This was discussed on HipChat a month ago.

```
Kabir Khan	@DarranLofthouse Do you agree that @JosefCacek 's https://github.com/jbossas/jboss-eap7/pull/2058 should go in full?	1:01 PM
Darran Lofthouse	Not really if new tests recently developed, some modules in core already test with ApacheDS	1:03 PM
Darran Lofthouse	And being SASL the Remoting subsystem is also already available in Core	1:03 PM
Josef Cacek	@DarranLofthouse Can you point me to some kerberos test in the core please?	1:07 PM
Josef Cacek	maybe I've missed when I was searching for it.	1:07 PM
Darran Lofthouse	I think domain management already pulled a lot in, so far at least for an LDAP server	1:08 PM
Josef Cacek	but moving Kerberos tests would mean non-trivial additional work if the infrastructure is not yet ready.	1:10 PM
Darran Lofthouse	But continuing to add more and more tests to WildFly is increasing the technical debt of test cases that are going to require clean up - they also increase the chance that three way PRs need coordination and in general make it much more complicated for engineers to verify Elytron changes are causing regressions without either waiting for slow CI or running multiple builds locally generally taking a local environment out of action whilst waiting for builds to complete.
```

@psakar The dependencies should be the same as the ones used in the WildFly full testsuite.

testsuite/elytron/src/test/java/org/wildfly/test/security/common/AbstractSystemPropertiesServerSetupTask.java
```diff
@@ -0,0 +1,176 @@
...
+    }
+
+    // Protected methods -----------------------------------------------------
+
+    /**
+     * Returns configuration of the login modules.
+     *
+     * @return
+     */
+    protected abstract SystemProperty[] getSystemProperties();
```
stuartwdouglas: Why do you have both an interface and an implementation class for something that can be represented as Map<String, String> ?

testsuite/elytron/pom.xml
```diff
@@ -20,10 +20,39 @@
...
+            <dependency>
+                <groupId>org.apache.directory.server</groupId>
+                <artifactId>apacheds-parent</artifactId>
+                <version>${version.org.apache.ds}</version>
+                <scope>import</scope>
+                <type>pom</type>
+            </dependency>
+
+            <dependency>
+                <groupId>javax.enterprise</groupId>
```
stuartwdouglas: Why is this in the dependency management section? It should already be managed by the parent pom

testsuite/elytron/src/test/java/org/wildfly/test/security/common/AbstractSystemPropertiesServerSetupTask.java
kwart: It's a copy from WildFly full. Maybe I was thinking about wider usage in the ancient times I've implemented it. I'll change it to simple Map here.

testsuite/elytron/pom.xml
kwart: IMO this shouldn't be in the parent POM. It's used just to workaround the ApacheDS limitations by providing customizable "annotations" (`extends AnnotationLiteral<T> implements T`).

stuartwdouglas: It is already in the parent pom, as CDI is part of the server

kwart: It's only in full. Should we move the dependency to core?

kwart: I've resolved this by removing the dependency and providing just the AnnnotationLiteral implementation.

fjuma:
@kwart The test cases in ``AbstractKerberosMgmtSaslTestBase`` look ok.

testsuite/elytron/src/test/java/org/wildfly/test/integration/elytron/sasl/mgmt/AbstractKerberosMgmtSaslTestBase.java
```diff
@@ -0,0 +1,538 @@
...
+            assertKerberosSaslMechFails("GSSAPI", "hnelson", "secret", false);
+            assertKerberosSaslMechFails("GS2-KRB5-PLUS", "hnelson", "secret", false);
+        }
+    }
+
+    /**
+     * Test GS2-KRB5 SASL mechanism configured on the server-side for management interface with using SSL.
+     */
+    @Test
+    @Ignore("ELY-1330")
```
fjuma: This can be unignored now since Elytron 1.1.0.CR6 has been merged to Core.

testsuite/elytron/src/test/java/org/wildfly/test/integration/elytron/sasl/mgmt/AbstractKerberosMgmtSaslTestBase.java
```diff
@@ -0,0 +1,538 @@
...
+            assertKerberosSaslMechFails("GS2-KRB5", "hnelson", "secret", false);
+            assertKerberosSaslMechFails("GSSAPI", "hnelson", "secret", false);
+        }
+    }
+
+    /**
+     * Test GS2-KRB5-PLUS SASL mechanism configured on the server-side for management interface with using SSL.
+     */
+    @Test
+    @Ignore("ELY-1328")
```
fjuma: Same here.

bstansberry:
Replaced by #2728 which has been merged.

PR #2713, opened 2017-08-14
[WFCORE-3183] CallbackKinds restriction for for CLI handler

honza889:
https://issues.jboss.org/browse/WFCORE-3183
https://issues.jboss.org/browse/JBEAP-12696

### DIFF
--- a/protocol/src/main/java/org/jboss/as/protocol/ProtocolConnectionUtils.java
+++ b/protocol/src/main/java/org/jboss/as/protocol/ProtocolConnectionUtils.java
@@ -31,6 +31,7 @@ import java.net.NetworkInterface;
 import java.net.URI;
 import java.security.GeneralSecurityException;
 import java.util.Collections;
+import java.util.EnumSet;
 import java.util.Enumeration;
 import java.util.HashMap;
 import java.util.Map;
@@ -46,6 +47,7 @@ import org.jboss.remoting3.Endpoint;
 import org.wildfly.security.auth.client.AuthenticationConfiguration;
 import org.wildfly.security.auth.client.AuthenticationContext;
 import org.wildfly.security.auth.client.AuthenticationContextConfigurationClient;
+import org.wildfly.security.auth.client.CallbackKind;
 import org.wildfly.security.auth.client.MatchRule;
 import org.wildfly.security.sasl.SaslMechanismSelector;
 import org.wildfly.security.sasl.localuser.LocalUserClient;
@@ -142,6 +144,12 @@ public class ProtocolConnectionUtils {
         return connectSync(config);
     }
 
+    private static final EnumSet<CallbackKind> DEFAULT_CALLBACK_KINDS = EnumSet.of(
+            CallbackKind.PRINCIPAL,
+            CallbackKind.CREDENTIAL,
+            CallbackKind.REALM
+    );
+
     private static IoFuture<Connection> connect(final CallbackHandler handler, final ProtocolConnectionConfiguration configuration) throws IOException {
         configuration.validate();
         final Endpoint endpoint = configuration.getEndpoint();
@@ -150,7 +158,9 @@ public class ProtocolConnectionUtils {
 
         AuthenticationContext captured = AuthenticationContext.captureCurrent();
         AuthenticationConfiguration mergedConfiguration = AUTH_CONFIGURATION_CLIENT.getAuthenticationConfiguration(uri, captured);
-        if (handler != null) mergedConfiguration = mergedConfiguration.useCallbackHandler(handler);
+        if (handler != null) {
+            mergedConfiguration = mergedConfiguration.useCallbackHandler(handler, DEFAULT_CALLBACK_KINDS);
+        }
 
         Map<String, String> saslOptions = configuration.getSaslOptions();
         mergedConfiguration = configureSaslMechanisms(saslOptions, isLocal(uri), mergedConfiguration);


bstansberry:
@jfdenise Please approve.

The change to CommandContextImpl means ModelControllerClientFactory.CUSTOM is no longer used. Already MCCF.DEFAULT is not used. So then neither of the 2 impls of that interface are used. That has a bit of code smell to it. Perhaps that's fine for the current release and can just be cleaned up in the next release, but I think you'd know better if it's a sign of something that needs to be corrected.

honza889:
It is possible callbackKinds will need to be added for other too - probably should add directly into the interface - will rework yet...

honza889:
@jamezp yes, if they implement custom userCallbackHandler but does not handle SSL callback, they will need another fix yet
otherwise (unused DEFAULT and CUSTOM) fixed as requested

jamezp:
@honza889 Is there an example of the custom callback? Both Arquillian and the wildfly-maven-plugin will likely need. As will JBoss Tools, the Netbeans plugin, Hawkular, and probably others.

honza889:
@jamezp as example can be used its using in CommandContextImpl (with AuthenticationCallbackHandler):
https://github.com/wildfly/wildfly-core/pull/2713/files#diff-c929deeb25b1e0886f6c256907ea7c44R1150

cli/src/main/java/org/jboss/as/cli/impl/CommandContextImpl.java
```diff
@@ -1787,6 +1788,15 @@ public String getVariable(String name) {
         return variables == null ? Collections.<String>emptySet() : variables.keySet();
     }
 
+    /** {@link CallbackKind}s supported by {@link AuthenticationCallbackHandler} */
+    private static final Set<CallbackKind> SUPPORTED_CALLBACK_KINDS = new HashSet<>();
```
jfdenise: EnumSet.of could be used there. 
How can we know that we are listing all the supported callbacks and not missing some? A comment that would detail the mapping between kind and callback instances could help.

cli/src/main/java/org/jboss/as/cli/impl/ModelControllerClientFactory.java
```diff
@@ -50,39 +51,19 @@
         void handleClose();
     }
 
-    ModelControllerClient getClient(ControllerAddress address, CallbackHandler handler,
+    ModelControllerClient getClient(ControllerAddress address, CallbackHandler handler, Set<CallbackKind> callbackKinds,
             boolean disableLocalAuth, SecurityFactory<SSLContext> sslContextFactory, boolean fallbackSslContext, int connectionTimeout,
             ConnectionCloseHandler closeHandler, ProtocolTimeoutHandler timeoutHandler, String clientBindAddress) throws IOException;
 
-    ModelControllerClientFactory DEFAULT = new ModelControllerClientFactory() {
```
jfdenise: I agree with the removal of this static instance. Its added value is very low and the signature would become even more confusing with the introduction of CallbackKind. 
Even if one is relying on it, it would have to recompile and re-emplementing it directly is straight forward.
BTW, ModelControllerClientConfiguration.Builder requires a change to support CallbackKind.

cli/src/main/java/org/jboss/as/cli/impl/CommandContextImpl.java
jfdenise: @honza889 , thank-you for the mapping, it seems that CHOICE is not needed, from javadoc: (not including {@link RealmChoiceCallback}). REALM kind should cover both.

honza889:
Updated: added callbackKinds to other builders, to allow to specify it everywhere where is CallbackHandler passed

cli/src/main/java/org/jboss/as/cli/impl/CommandContextImpl.java
honza889: fixed

cli/src/main/java/org/jboss/as/cli/impl/ModelControllerClientFactory.java
jfdenise: @honza889, now that you have added support for the Builder, you should possibly restore the DEFAULT instance and update the call to the Builder with the Kinds. We could create a separate JIRA to handle dead code.

bstansberry:
Now we've got obscure elytron concepts creeping into general purpose public API like controller-client. Who can convince me that this is not a workaround-of-the-day and we're not going to need some other obscure concept added next month?

jamezp:
This is my concern as well. We can't expect tooling; Arquillian, wildfly-maven-plugin, Hawular, etc. to have this much knowledge of Elytron IMO. These tools just want an MCC that can talk to a server. Having to keep making changes on the consumers of the API will be problematic.

cli/src/main/java/org/jboss/as/cli/impl/ModelControllerClientFactory.java
honza889: Builder update reverted in the end - keeping DEFAULT removed

honza889:
controller-client module api changes was removed - is it ok now?

honza889:
By hipchat discussion this will be closed and repaced by fix https://github.com/wildfly-security/wildfly-elytron/pull/951

honza889:
> What I was really saying is where any of the CLI and ModelControllerClients are already used and already have a CallbackHandler passed in we should always assume that CBH is just compatible with this small subset of kinds - so really quite a lot of your original PR could be retained but we don't need to expose the callback kinds on the APIs
So reopening and closing the replacement ;)

honza889:
As requested, simplified to just set default directly in ProtocolConnectionUtils, so no api changes in the end...

bstansberry:
[10:00 AM] Brian Stansberry: DarranLofthouse HonzaKalina https://github.com/wildfly/wildfly-core/pull/2713/files has gotten broader, but with the troublesome public API part gone. now it looks like it would affect the behavior not only of external clients, but also of internal connections within a domain. probably fine, but I'm going to need to doublecheck
[10:25 AM] Darran Lofthouse: BrianStansberry historically all existing callback handlers that already exist could only ever have supported those three callback kinds
[10:30 AM] Brian Stansberry: DarranLofthouse ok, good. Does that PR look good to you?
[10:37 AM] Darran Lofthouse: PR looks good,

PR #2718, opened 2017-08-16
[WFCORE-3190] CLI autocompletion handles incorrectly spaces in front …

xstefank:
…of attribute name

WFCORE: https://issues.jboss.org/browse/WFCORE-3190
JBEAP: https://issues.jboss.org/browse/JBEAP-12484

### DIFF
--- a/cli/src/main/java/org/jboss/as/cli/CommandCompleter.java
+++ b/cli/src/main/java/org/jboss/as/cli/CommandCompleter.java
@@ -65,6 +65,7 @@ public class CommandCompleter implements CommandLineCompleter {
         if(result < 0) {
             return result;
         }
+
         return result - offset;
     }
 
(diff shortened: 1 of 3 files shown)


cli/src/main/java/org/jboss/as/cli/CommandCompleter.java
```diff
@@ -65,7 +65,8 @@ public int complete(CommandContext ctx, String buffer, int cursor, List<String>
         if(result < 0) {
             return result;
         }
-        return result - offset;
```
jfdenise: @xstefank, this would work but I am wandering if you could isolate the fix to the failing piece (value completion). If you look at the OperationRequestCompleter, line 300: https://github.com/wildfly/wildfly-core/blob/master/cli/src/main/java/org/jboss/as/cli/operation/OperationRequestCompleter.java#L300
The chunk that has been completed doesn't contain the ending whitespaces. I think that it is the issue root cause. 
Computing the length diff between the original line (the substring of the buffer starting at result index) and the chunk should capture this case.

jfdenise:
retest this please!

xstefank:
retest this please

xstefank:
retest this please

bstansberry:
@jfdenise looks ok now?

bstansberry:
@xstefank Sorry, Martin; needs a rebase. Please ping when the rebase passes CI and I'll merge it.

xstefank:
@bstansberry should be good to go now

PR #2724, opened 2017-08-17
[WFCORE-3192]:Remarshaling the default configuration xml files shouldn't change the files

ehsavoie:
* Reordering elements so that they appear in the same order after a marshaling.
* Removing the creation of empty lines so that configuration files are not updated with just new empty lines.

Jira: https://issues.jboss.org/browse/WFCORE-3192
JBEAP: https://issues.jboss.org/browse/JBEAP-12768

### DIFF
--- a/server/src/main/java/org/jboss/as/server/parsing/StandaloneXml_5.java
+++ b/server/src/main/java/org/jboss/as/server/parsing/StandaloneXml_5.java
@@ -674,40 +674,31 @@ class StandaloneXml_5 extends CommonXml implements ManagementXmlDelegate {
         writeNamespaces(writer, modelNode);
         writeSchemaLocation(writer, modelNode);
 
-        WriteUtils.writeNewLine(writer);
-
         if (modelNode.hasDefined(EXTENSION)) {
             extensionHandler.writeExtensions(writer, modelNode.get(EXTENSION));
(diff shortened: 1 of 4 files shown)


server/src/main/java/org/jboss/as/server/parsing/StandaloneXml_5.java
```diff
@@ -843,27 +832,37 @@ public boolean parseAccessControl(XMLExtendedStreamReader reader, ModelNode addr
...
-                ModelNode provider = AccessAuthorizationResourceDefinition.PERMISSION_COMBINATION_POLICY.parse(value, reader);
-                ModelNode op = Util.getWriteAttributeOperation(accAuthzAddr,
-                        AccessAuthorizationResourceDefinition.PERMISSION_COMBINATION_POLICY.getName(), provider);
-
-                operationsList.add(op);
-            } else {
+            if (null == attribute) {
                 throw unexpectedAttribute(reader, i);
             }
+            switch (attribute) {
```
bstansberry: An enum switch results in an inner class that has to be loaded. If we ever had time we'd probably try to get rid of all of the ones we have in the parsers.

ehsavoie: it is a string switch ;)

PR #2727, opened 2017-08-18
[WFCORE-2441] Inconsistency between DMR and XSD representation of Elytron simple-permission-mapper

ivassile:
wildfly-elytron-integration: https://issues.jboss.org/browse/WFCORE-2441
JBEAP: https://issues.jboss.org/browse/JBEAP-7515


bstansberry:
What does the parser do if there are no permission child elements?

I understand from the issue that the management API will not object if presented DMR with no permissions, but I'm checking whether the parser will object. If it does we would have to correct that.

PR #2728, opened 2017-08-18
Update Pull request #2709 to unignore tests

bstansberry:
Follow up on comments from @fjuma on #2709 so I can get it in today if the original PR isn't updated before then.


fjuma:
Just discovered a subtle test issue that is causing testGs2Krb5OverSsl() to fail. This fixes it:

https://github.com/bstansberry/wildfly-core/pull/9

PR #2729, opened 2017-08-18
Fix for WFCORE-3199 - Make sure the security permissions.xml DUPs are registered before the deployment chain is frozen

jaikiran:
The commit here fixes the issue reported in https://issues.jboss.org/browse/WFLY-6657

The issue is a bit involved (which I explain below) but the gist of the issue is that the processing of `permissions.xml` and its associated security semantics, within a deployment, are completed skipped and ignored, because the deployment unit processor(s) responsible for setting up these security constraints never get triggered. That's because they get added by the `SecurityManagerSubsystemAdd` too late in the game after the `RUNTIME` step of `DeployerChainAddHandler` has already formalized deployment unit processors that are enrolled with it and has created a "final view" of those processors. 

The reason why this shows up only when the `org.jboss.server.bootstrap.maxThreads` system property is set, is the involved part. The way the boot process is setup, in the absence of this system property, a `parallel boot step handler` gets involved which in its implementation organizes the boot operations such that the subsystem specific operations are parallely executed _first_ and  then the non-subsystem related operations, like the `add-deployers-chain` (which is the one which deals with forming the deployment unit processors chain) are executed in the _end_. So, it just so happens that due to this implementation, the `SecurityManagerSubsystemAdd` (through its subsystem specific management operation steps) does indeed add the DUPs before the deployers chain is formed.

Now, when the `org.jboss.server.bootstrap.maxThreads` system property is set, this `parallel boot` for subsystems is skipped. Even then, the boot operations are arranged such that the `add-deployers-chain` ends up being at the end of the list of `RUNTIME` steps that run. In fact, the `DeployerChainAddHandler` does its best to make sure it ends up being the last `RUNTIME` step as it notes in its comment[1]:

>>> // Our real work needs to run after all RUNTIME steps that add DUPs have run. ServerService adds
            // this operation at the end of the boot op list, so our MODEL stage step is last, and
            // this RUNTIME step we are about to add should therefore be last as well *at the time
            // we register it*. However, other RUNTIME steps can themselves add new RUNTIME steps that
            // will then come after this one. So we do the same -- add a RUNTIME step that adds another
            // RUNTIME step that does the real work.
            // Theoretically this kind of "predecessor runtime step adds another runtime step, so we have to
            // add one to come later" business could go on forever. *But any operation that does that with
            // a DUP-add step is just broken and should just find another way.*

but as noted in the last sentence of that comment, if there are operations that create a RUNTIME step which then creates another RUNTIME step to add a DUP, then this whole finely balanced logic,  goes for a toss. The `SecurityManagerSubsystemAdd` is such a operation which, IMO, is faulty since it creates a `RUNTIME` step in its `performBoottime` and then that `RUNTIME` step goes ahead and creates another `RUNTIME` step which finally adds the DUPs.

The commit in this PR, removes the creation of the first `RUNTIME` step in this `SecurityManagerSubsystemAdd`, because from what I see, there's no need for it since all it does is read the complete model, relative to the subsystem, which should already be available in this `performBoottime` `RUNTIME` step. Of course, it's been a while since I have fiddled with the WildFly management operations, so I might have misunderstood or missed certain details of why this additional `RUNTIME` step was being created/needed.

The commit also contains a test case which reproduces the issue and verifies the proposed fix.

[1] https://github.com/wildfly/wildfly-core/blob/master/server/src/main/java/org/jboss/as/server/DeployerChainAddHandler.java#L96

P.S: On an unrelated note, shouldn't these `permissions.xml` related DUPs be part of the WildFly itself instead of wildfly-core, since that `permissions.xml` file within a deployment, is a EE functionality.

### DIFF
--- a/security-manager/src/main/java/org/wildfly/extension/security/manager/SecurityManagerSubsystemAdd.java
+++ b/security-manager/src/main/java/org/wildfly/extension/security/manager/SecurityManagerSubsystemAdd.java
@@ -42,7 +42,6 @@ import java.util.List;
 import org.jboss.as.controller.AbstractBoottimeAddStepHandler;
 import org.jboss.as.controller.OperationContext;
 import org.jboss.as.controller.OperationFailedException;
-import org.jboss.as.controller.OperationStepHandler;
 import org.jboss.as.controller.PathAddress;
 import org.jboss.as.controller.registry.Resource;
 import org.jboss.as.server.AbstractDeploymentChainStep;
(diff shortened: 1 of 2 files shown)


bstansberry:
Thanks. I agree there is no reason for the added step. All the needed model state is complete when performBoottime runs. And the child resources do nothing in Stage.RUNTIME during boot, so they are not impacted by when the parent resource does its RUNTIME work.

We decided that security manager configuration isn't purely an EE concern.

bstansberry:
BTW, that installProcessors method isn't used anywhere else and can just be inlined now. It shouldn't have been protected before.

jaikiran:
You're right - I've now updated this PR to inline that method and also change the commit log to refer to the new WFCORE JIRA id instead of the WFLY one.

bstansberry:
Thanks!

PR #2730, opened 2017-08-19
Boot optimizations

bstansberry:
This PR includes a couple different types of optimizations related to boot perf:

1) Small improvements to some object interning logic that I'd added to cut down on the WF memory footprint. (2nd and 3rd commits.)

2) A lot of management code (classloading and initialization of constants like AttributeDefinition FOO or XXXResourceDefinition.INSTANCE) is happening when xml parsers are invoked, which is a single-threaded part of boot meaning it all happens in series on one CPU core. All but one of the other commits here aim at pushing this work into the calls to Extension.initializeParsers, which happen concurrently for each extension.

There's also a trivial cleanup commit.

### DIFF
--- a/controller/src/main/java/org/jboss/as/controller/capability/RuntimeCapability.java
+++ b/controller/src/main/java/org/jboss/as/controller/capability/RuntimeCapability.java
@@ -94,7 +94,7 @@ public class RuntimeCapability<T> extends AbstractCapability  {
     private static final ServiceName ORG_WILDFLY = ServiceName.parse("org.wildfly");
 
     private final Class<?> serviceValueType;
-    private final ServiceName serviceName;
+    private volatile ServiceName serviceName;
     private final T runtimeAPI;
     private final boolean allowMultipleRegistrations;
 
@@ -112,7 +112,6 @@ public class RuntimeCapability<T> extends AbstractCapability  {
         super(name, false, requirements, optionalRequirements, null, null, null, null);
         this.runtimeAPI = runtimeAPI;
         this.serviceValueType = null;
-        this.serviceName = null;
         this.allowMultipleRegistrations = ALLOW_MULTIPLE;
     }
 
@@ -143,7 +142,6 @@ public class RuntimeCapability<T> extends AbstractCapability  {
         super(name, false, new HashSet<>(Arrays.asList(requirements)), null, null, null, null, null);
         this.runtimeAPI = runtimeAPI;
         this.serviceValueType = null;
-        this.serviceName = null;
         this.allowMultipleRegistrations = ALLOW_MULTIPLE;
     }
 
@@ -155,7 +153,6 @@ public class RuntimeCapability<T> extends AbstractCapability  {
                 builder.runtimeOnlyRequirements, builder.dynamicRequirements, builder.dynamicOptionalRequirements, builder.dynamicNameMapper);
         this.runtimeAPI = builder.runtimeAPI;
         this.serviceValueType = builder.serviceValueType;
-        this.serviceName = this.serviceValueType == null ? null : ServiceNameFactory.parseServiceName(builder.baseName);
         this.allowMultipleRegistrations = builder.allowMultipleRegistrations;
     }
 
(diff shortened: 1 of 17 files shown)


controller/src/main/java/org/jboss/as/controller/capability/RuntimeCapability.java
```diff
@@ -329,14 +325,22 @@ public boolean isAllowMultipleRegistrations() {
...
                 getRuntimeOnlyRequirements(), getDynamicRequirements(),
                 getDynamicOptionalRequirements(), allowMultipleRegistrations,
                 dynamicNameMapper, dynamicElement);
 
     }
 
+    private ServiceName getServiceName() {
+        ServiceName result = serviceName;
+        if (result == null && serviceValueType != null) {
+            result = this.serviceName = ServiceNameFactory.parseServiceName(getName());
```
jamezp: Does this need to be thread safe? I just noticed it was volatile is why I ask. I guess though the worst case scenario is we end up with the same `ServiceName`.

controller/src/main/java/org/jboss/as/controller/services/path/PathInfoHandler.java
```diff
@@ -100,6 +100,20 @@
                         .setDefaultValue(new ModelNode(MeasurementUnit.BYTES.getName()))
                         .build();
 
+    private static DateTimeFormatter getDateFormat() {
+        if (DATE_FORMAT == null) {
+            DATE_FORMAT = new DateTimeFormatterBuilder().appendInstant().appendZoneId().toFormatter(Locale.ENGLISH);
```
jamezp: Not really relevant for this PR, but we should probably use something like `DateTimeFormatter.ISO_OFFSET_DATE_TIME` so it's a standard format that could be parsed.

Current format:
```
2017-09-18T22:48:24.367ZAmerica/Los_Angeles
```
`DateTimeFormatter.ISO_OFFSET_DATE_TIME`
```
2017-09-18T15:48:24.367-07:00
```

controller/src/main/java/org/jboss/as/controller/services/path/PathInfoHandler.java
```diff
@@ -100,6 +100,20 @@
...
+        }
+        return DATE_FORMAT;
+    }
+
+    private static ZoneId getZoneId() {
+        if (ZONE_ID == null) {
+            ZONE_ID = ZoneId.of(Calendar.getInstance().getTimeZone().getID());
```
jamezp: Can we just use `ZoneId.systemDefault()` here?

controller/src/main/java/org/jboss/as/controller/services/path/PathInfoHandler.java
jamezp: I guess another option is `DateTimeFormatter.ISO_ZONED_DATE_TIME` which would be
```
2017-09-18T15:56:21.875-07:00[America/Los_Angeles]
```

bstansberry: +1. I opened https://issues.jboss.org/browse/WFCORE-3304

controller/src/main/java/org/jboss/as/controller/capability/RuntimeCapability.java
bstansberry: Exactly. It's volatile so the thread will see the latest at L337 and hopefully avoid falling into the if block, but if there's a race and two threads both see null at L337, it just means we do the work twice.

I could wrap L339 in a synchronized block and check this.serviceName again but I don't really expect that kind of race so I kept it simple for the expected non-race case.

controller/src/main/java/org/jboss/as/controller/services/path/PathInfoHandler.java
bstansberry: I'll add a commit to change it to that.

jamezp:
Failures don't look related, retest this please

PR #2735, opened 2017-08-21
WFCORE-2147 & WFCORE-1295 - A proposal to add expression expansion support for deployment descriptors without using JBoss Metadata project constructs

jaikiran:
**Related JIRAs:**

https://issues.jboss.org/browse/WFCORE-2147
https://issues.jboss.org/browse/WFCORE-1295

**Background:**

WildFly (not WildFly Core), through its EE subsystem supports system property expansion (a.k.a expression expansion) in spec descriptors as well as JBoss/WildFly specific descriptors. For example, the content of the form `${foobar:blah}` is processed to look for the `foobar` system property to be set and in the absence of it, the value is computed as `blah`. In the presence of the system property, the value is computed to be whatever the system property value is.

The way this support is implemented by the EE subsystem involves 2 things:

- A toggle/switch, controlled through a EE subsystem resource attribute to decide whether the spec descriptors need to be processed with this support. A separate resource attribute to decide whether the JBoss/WildFly descriptors need to be processed with this support.

- If the spec or JBoss descriptors are enabled for expression expansion support, a `PropertyResolver`/`PropertyReplacer` is setup accordingly by the EE subsystem and this construct is passed around to various parsers that handle the spec/JBoss descriptors' parsing. The whole interface, implementation of the `PropertyResolver`/`PropertyReplacer` resides in the JBoss Metadata project. Many parsers (not all) resides in that project too

**The problem:**

With the split of WildFly Core and WildFly projects, there's now a case where certain WildFly specific descriptors and in fact even (EE) spec descriptor in one specific case (`permissions.xml`) can no longer make use of this support, since the support, which involves the toggle/switch management resource attribtue plus the JBoss metadata interfaces/implementation lies within the EE subsystem which is outside of WildFly core.

As noted in this JIRA https://issues.jboss.org/browse/WFCORE-1295 by Brian, the other aspect here is that if this support needs to be brought into WildFly core, then it can't be backed by JBoss Metadata project constructs (for reasons discussed elsewhere).

**Potential solution:**

Brian, in https://issues.jboss.org/browse/WFCORE-1295, notes that one way to solve this is to provide a hook in WildFly core which allows WildFly (and other consumers of WildFly Core) to somehow enable/disable the expression expansion in such descriptors. This PR, introduces such a hook, through the introduction of a couple of `org.jboss.as.server.deployment.Attachments` - specifically:

```
    /**
     * A function which will be used to expand expressions within spec descriptors
     */
    public static final AttachmentKey<BiConsumer<ResolveContext<RuntimeException>, StringBuilder>> SPEC_DESCRIPTOR_EXPR_EXPAND_FUNCTION = AttachmentKey.create(BiConsumer.class);

    /**
     * A function which will be used to expand expressions within JBoss/WildFly (vendor specific) descriptors
     */
    public static final AttachmentKey<BiConsumer<ResolveContext<RuntimeException>, StringBuilder>> WFLY_DESCRIPTOR_EXPR_EXPAND_FUNCTION = AttachmentKey.create(BiConsumer.class);

```
The 2 important aspects here are:

- Well-known attachment keys, which can be used to set an expression expansion function. One for spec descriptors and one for JBoss/WildFly descriptors. The absence of the respective attachment keys implies no expression expansion support in the corresponding descriptors.

- Using the `Expression` API from `wildfly-common` project as a formal contract, instead of JBoss Metadata project, for such expression expansion

With this change, one part of the problem would be solved - i.e. allowing a hook in WildFly Core, which doesn't rely on JBoss Metadata, to enable/disable expression expansion. This PR (intentionally) doesn't try to solve the other part of the problem, which is, who controls the toggle/switch to enable/disable the support. For now, it's the EE subsystem which does that and for the scope of the JIRAs noted above, IMO, it makes sense to leave that control to the EE subsystem for now until we decide whether that needs to be moved out to a more "core" level.

If the change proposed in this PR gets accepted, then all we need to do in WildFly is a small addition to an existing deployment unit processor, so that it constructs this expression expansion function that is backed by the currently integrated JBoss Metadata property resolvers/replacers. That change will look like this https://github.com/jaikiran/wildfly/commit/c6f5a2f06c886b163db33597664c1089431d4fdf

This PR also uses this proposed support in supporting property expansion in `permissions.xml`/`jboss-permissions.xml` files and as such should resolve https://issues.jboss.org/browse/WFCORE-2147. If this is accepted, subsequent PRs can be issued to add support for some other WildFly specific descriptor parsers like the one for `jboss-deployment-structure.xml`.

### DIFF
--- a/security-manager/src/main/java/org/wildfly/extension/security/manager/deployment/PermissionsParser.java
+++ b/security-manager/src/main/java/org/wildfly/extension/security/manager/deployment/PermissionsParser.java
@@ -29,6 +29,7 @@ import java.util.Iterator;
 import java.util.List;
 import java.util.Map;
 import java.util.Set;
+import java.util.function.BiConsumer;
 
 import javax.xml.stream.XMLStreamConstants;
 import javax.xml.stream.XMLStreamException;
@@ -38,6 +39,8 @@ import org.jboss.modules.ModuleIdentifier;
 import org.jboss.modules.ModuleLoader;
 import org.jboss.modules.security.ModularPermissionFactory;
 import org.jboss.modules.security.PermissionFactory;
+import org.wildfly.common.expression.Expression;
+import org.wildfly.common.expression.ResolveContext;
 import org.wildfly.extension.security.manager.logging.SecurityManagerLogger;
 
 /**
@@ -48,8 +51,39 @@ import org.wildfly.extension.security.manager.logging.SecurityManagerLogger;
  */
 public class PermissionsParser {
 
+    /**
+     * This method is equivalent to calling {@link #parse(XMLStreamReader, ModuleLoader, ModuleIdentifier, BiConsumer)} with {@code null}
+     * as the last argument.
+     *
+     * @param reader     The {@link XMLStreamReader} for the XML file
+     * @param loader     A {@link ModuleLoader} which will be used to load the configured permission class and create a {@link PermissionFactory}
+     *                   out of it
+     * @param identifier A {@link ModuleIdentifier} which identifies the module, which contains the permission class to load
+     * @return A list of {@link PermissionFactory}s constructed out of the parsed content
+     * @throws XMLStreamException
+     * @deprecated Use {@link #parse(XMLStreamReader, ModuleLoader, ModuleIdentifier, BiConsumer)} instead
+     */
+    @Deprecated
     public static List<PermissionFactory> parse(final XMLStreamReader reader, final ModuleLoader loader, final ModuleIdentifier identifier)
             throws XMLStreamException {
+        return parse(reader, loader, identifier, null);
+    }
+
+    /**
+     * Parses the contents of a permissions.xml/jboss-permissions.xml file
+     *
+     * @param reader             The {@link XMLStreamReader} for the XML file
+     * @param loader             A {@link ModuleLoader} which will be used to load the configured permission class and create a {@link PermissionFactory}
+     *                           out of it
+     * @param identifier         A {@link ModuleIdentifier} which identifies the module, which contains the permission class to load
+     * @param exprExpandFunction A function which will be used, if provided, to expand any expressions (of the form of {@code ${foobar}})
+     *                           in the content being parsed. This function can be null, in which case the content is processed literally.
+     * @return A list of {@link PermissionFactory}s constructed out of the parsed content
+     * @throws XMLStreamException
+     */
+    public static List<PermissionFactory> parse(final XMLStreamReader reader, final ModuleLoader loader,
+                                                final ModuleIdentifier identifier,
+                                                final BiConsumer<ResolveContext<RuntimeException>, StringBuilder> exprExpandFunction) throws XMLStreamException {
 
         reader.require(XMLStreamConstants.START_DOCUMENT, null, null);
 
@@ -59,7 +93,7 @@ public class PermissionsParser {
                     Element element = Element.forName(reader.getLocalName());
                     switch (element) {
                         case PERMISSIONS: {
-                            return parsePermissions(reader, loader, identifier);
+                            return parsePermissions(reader, loader, identifier, exprExpandFunction);
                         }
                         default: {
                             throw unexpectedElement(reader);
@@ -74,7 +108,7 @@ public class PermissionsParser {
         throw unexpectedEndOfDocument(reader);
     }
 
-    private static List<PermissionFactory> parsePermissions(final XMLStreamReader reader, final ModuleLoader loader, final ModuleIdentifier identifier)
+    private static List<PermissionFactory> parsePermissions(final XMLStreamReader reader, final ModuleLoader loader, final ModuleIdentifier identifier, final BiConsumer<ResolveContext<RuntimeException>, StringBuilder> exprExpandFunction)
             throws XMLStreamException {
 
         List<PermissionFactory> factories = new ArrayList<PermissionFactory>();
@@ -115,7 +149,7 @@ public class PermissionsParser {
                     Element element = Element.forName(reader.getLocalName());
                     switch (element) {
                         case PERMISSION: {
-                            PermissionFactory factory = parsePermission(reader, loader, identifier);
+                            PermissionFactory factory = parsePermission(reader, loader, identifier, exprExpandFunction);
                             factories.add(factory);
                             break;
                         }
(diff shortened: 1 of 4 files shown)


jaikiran:
Failure's unrelated to the changes:

>>> [08:43:51][Step 2/3] org.jboss.as.controller.test.TransactionalProtocolClientTestCase
[08:43:51][org.jboss.as.controller.test.TransactionalProtocolClientTestCase] testSequentialGroup
[08:43:51][org.jboss.as.controller.test.TransactionalProtocolClientTestCase] testCancelBeforePrepared
[08:43:51][org.jboss.as.controller.test.TransactionalProtocolClientTestCase] testSimpleRequest
[08:43:51][org.jboss.as.controller.test.TransactionalProtocolClientTestCase] testClosePrepared
[08:43:51][org.jboss.as.controller.test.TransactionalProtocolClientTestCase] testConcurrentGroup
[08:43:51][testConcurrentGroup] java.net.BindException: Address already in use: bind
[08:43:51]
[testConcurrentGroup] java.net.BindException: Address already in use: bind
	at sun.nio.ch.Net.bind0(Native Method)
	at sun.nio.ch.Net.bind(Net.java:433)
	at sun.nio.ch.Net.bind(Net.java:425)
	at sun.nio.ch.ServerSocketChannelImpl.bind(ServerSocketChannelImpl.java:223)
	at sun.nio.ch.ServerSocketAdaptor.bind(ServerSocketAdaptor.java:74)
	at sun.nio.ch.ServerSocketAdaptor.bind(ServerSocketAdaptor.java:67)
	at org.xnio.nio.NioXnioWorker.createTcpConnectionServer(NioXnioWorker.java:181)
	at org.xnio.XnioWorker.createStreamConnectionServer(XnioWorker.java:282)
	at org.jboss.remoting3.remote.RemoteConnectionProvider$ProviderInterface.createServer(RemoteConnectionProvider.java:372)
	at org.jboss.as.controller.support.ChannelServer.create(ChannelServer.java:92)
	at org.jboss.as.controller.test.TransactionalProtocolClientTestCase.startChannelServer(TransactionalProtocolClientTestCase.java:108)
[00:44:00][Step 2/3] Failed tests detected
[08:43:51][org.jboss.as.controller.test.TransactionalProtocolClientTestCase] testConcurrentGroup
[08:43:51][testConcurrentGroup] java.lang.NullPointerException:
[08:43:51]
[testConcurrentGroup] java.lang.NullPointerException: null
	at org.jboss.as.controller.test.TransactionalProtocolClientTestCase.stopChannels(TransactionalProtocolClientTestCase.java:142)

jaikiran:
Just catching up and noticed that WildFly 11 CR1 is released. I was hoping to get this fix/change into that release, but I guess it's now too late for the 11.0 release to solve this https://developer.jboss.org/message/964059#964059.

I can work on changes that might be required to have that support in a subsequent release.

jaikiran:
retest this please

bstansberry:
Hi @jaikiran. Sorry; I just noticed this. Yes, it is too late for WF 11. (It was too late quite a while ago, so my not noticing this until now didn't affect that.)

I'll have a look. One concern is whether the wildfly-common expression resolution is semantically equivalent to what we support in management model expression resolution and in the JBoss Metadata resolution. It will have to be. It was independently developed so that is not a given. IIRC the intent was that it would be flexible enough to support what we do, plus other uses patterns that other projects may want, so I think this is really a matter of proving it does what we want.

jaikiran:
>> One concern is whether the wildfly-common expression resolution is semantically equivalent to what we support in management model expression resolution and in the JBoss Metadata resolution. It will have to be. It was independently developed so that is not a given. IIRC the intent was that it would be flexible enough to support what we do, plus other uses patterns that other projects may want

Hi @bstansberry. From what I see of the `Expression` API in wildfly-common project, it doesn't do the expression expansion on its own and instead lets the callers of the API(s) provide expansion functions to expand the expression in whatever way it sees fit. Having said that, I do understand what you mean by having to provide that it will work for other cases too, because the expression _compilation_ (the part where it parses a string and decides what expressions are part of the string) is an internal detail of that API which means that it has to understand the management model expression syntax to parse it correctly. 

Looking at the commit history of that API @dmlloyd might know more about the current state of that API.

jaikiran:
retest this please

security-manager/src/main/java/org/wildfly/extension/security/manager/deployment/PermissionsParser.java
```diff
@@ -48,8 +51,39 @@
...
+    }
+
+    /**
+     * Parses the contents of a permissions.xml/jboss-permissions.xml file
+     *
+     * @param reader             The {@link XMLStreamReader} for the XML file
+     * @param loader             A {@link ModuleLoader} which will be used to load the configured permission class and create a {@link PermissionFactory}
+     *                           out of it
+     * @param identifier         A {@link ModuleIdentifier} which identifies the module, which contains the permission class to load
+     * @param exprExpandFunction A function which will be used, if provided, to expand any expressions (of the form of {@code ${foobar}})
```
dmlloyd: Hi @jaikiran, what was the reason for introducing a special parameter for this (particularly if we want to standardize on wildfly-common expression parsing)?

jaikiran: Hi @dmlloyd, the sole reason (from what I remember) was to allow the existing implementation in WildFly EE (which is backed by JBoss Metadata) to be plugged into this new scheme of things. Here's what I did in the WildFly side https://github.com/jaikiran/wildfly/commit/c6f5a2f06c886b163db33597664c1089431d4fdf. I didn't want to change the semantics of the property expansion, that's already existing in WildFly EE descriptors. It's been a while since I submitted this PR, so might be mis-remembering other details, but I'm pretty sure that was the main reason.

jaikiran:
The PR requires a rebase, I'll sort it out sometime later today.

security-manager/src/main/java/org/wildfly/extension/security/manager/deployment/PermissionsParser.java
dmlloyd: OK thanks.

jaikiran:
More of a FYI - I have rebased this PR against latest master to resolve the merge conflicts.

jaikiran:
Closing this. It's been a while since I looked into this and don't even know if this is still relevant or there are better ways to do this.

PR #2736, opened 2017-08-22
[WFCORE-3201] Add JaCoCo profile

marekkopecky:
Add JaCoCo profile to wf-core TS

PR adds new profile, current default behaviour is not changed.

JBEAP jira: https://issues.jboss.org/browse/JBEAP-12795
WFCORE jira: https://issues.jboss.org/browse/WFCORE-3201
WF-CORE 3.0.x PR: https://github.com/jbossas/wildfly-core-eap/pull/406

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -150,6 +150,7 @@
         <version.org.wildfly.security.elytron-web.undertow-server>1.0.0.Final</version.org.wildfly.security.elytron-web.undertow-server>
         <version.xml-resolver>1.2</version.xml-resolver> <!-- Apache xml-resolver -->
         <!-- Non-default maven plugin versions and configuration -->
+        <version.jacoco.plugin>0.8.0</version.jacoco.plugin>
         <version.org.zanata.plugin>3.9.1</version.org.zanata.plugin>
         <version.xml.plugin>1.0.1</version.xml.plugin>
 
@@ -159,7 +160,7 @@
 
         <!-- Surefire args -->
         <surefire.jpda.args/>
-        <surefire.system.args>${modular.jdk.args} ${modular.jdk.props} -ea -Duser.region=US -Duser.language=en -Duser.timezone=America/Chicago -XX:MaxMetaspaceSize=256m ${surefire.jpda.args}</surefire.system.args>
+        <surefire.system.args>${modular.jdk.args} ${modular.jdk.props} -ea -Duser.region=US -Duser.language=en -Duser.timezone=America/Chicago -XX:MaxMetaspaceSize=256m ${surefire.jpda.args} ${surefire.jacoco.args}</surefire.system.args>
 
         <!-- Checkstyle configuration -->
         <linkXRef>false</linkXRef>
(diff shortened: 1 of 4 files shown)


pom.xml
```diff
@@ -1771,5 +1772,75 @@
...
+                                <groupId>org.jacoco</groupId>
+                                <artifactId>org.jacoco.ant</artifactId>
+                                <version>${version.jacoco.plugin}</version>
+                            </dependency>
+                        </dependencies>
+                    </plugin>
+                </plugins>
+            </build>
+        </profile>
+        <profile> <!-- surefire.jacoco.args property needs to be set correctly, if coverage profile is not enabled -->
```
bstansberry: Why is a separate profile needed for this, instead of what we do with surefire.jpda.args at L160?

marekkopecky: Hi Brian.

surefire.jpda.args property is set in jpda profile directly in <properties>
https://github.com/wildfly/wildfly-core/blob/master/pom.xml#L1730

surefire.jacoco.args property is set by JaCoCo plugin:
https://github.com/marekkopecky/wildfly-core/blob/WFCORE-3201/pom.xml#L1804

Although echo from maven-antrun-plugin is able to print this property correctly, JaCoCo plugin is unfortunately unable to set surefire.jacoco.args to the rest of maven with <surefire.jacoco.args/> defined like surefire.jpda.args is defined on L160. The consequence is that surefire JVM and EAP JVM is started without jacoco property and coverage is not stored to exec file.

See this experiment: https://paste.fedoraproject.org/paste/6zSiMYxcu5Y9nj73TYdSEw/raw

pom.xml
```diff
@@ -1771,5 +1772,75 @@
...
+                            <execution>
+                                <id>ts.jacoco.debug</id>
+                                <phase>post-integration-test</phase>
+                                <goals><goal>run</goal></goals>
+                                <inherited>false</inherited>
+                                <configuration>
+                                    <target>
+                                        <echo>Jacoco jvm args: ${surefire.jacoco.args}</echo>
+                                    </target>
+                                </configuration>
```
ctomc: please remove this, for debugging properties and profiles, use maven help plugin
help:effective-settings should give you what you want there.
adding ant-run-plugin just for that that is just bad practice.

pom.xml
```diff
@@ -1771,5 +1772,75 @@
                 </plugins>
             </build>
         </profile>
+
+
+
+        <!-- JaCoCo test coverage. Will set ${surefire.jacoco.args} to be used in Arquillian config and as surefire jvm argument. -->
```
ctomc: There are no arquillian tests in WildFly core
And I don't know if jacoco will work with our WildFly test runner.

Question is should we even have this enabled for testsuite at all, maybe having it for unit tests would be enough

pom.xml
marekkopecky: Done

pom.xml
marekkopecky: There are no arquillian tests in WildFly core, you are right, I update this comment.

Yes, JaCoCo is basically java agent, it works correctly also with WildFly Test Runner.

If we want to know test coverage of this testsuite, we need to enable JaCoCo in this testsuite. If we enable JaCoCo only in unit tests, we don't have test coverage data for testsuite. Testsuite adds more test coverage then unit tests.

Anyway, JaCoCo is disabled by default in this PR, it needs to be activated by -Dcoverage.

bstansberry:
@ctomc is this ok now?

pom.xml
```diff
@@ -149,6 +149,7 @@
         <version.org.wildfly.security.elytron-web.undertow-server>1.0.0.Final</version.org.wildfly.security.elytron-web.undertow-server>
         <version.xml-resolver>1.2</version.xml-resolver> <!-- Apache xml-resolver -->
         <!-- Non-default maven plugin versions and configuration -->
+        <version.jacoco.plugin>0.7.9</version.jacoco.plugin>
```
ctomc: 0.8.0 is out

pom.xml
```diff
@@ -1771,5 +1772,49 @@
...
+                                    <propertyName>surefire.jacoco.args</propertyName>
+                                </configuration>
+                            </execution>
+                        </executions>
+                    </plugin>
+                </plugins>
+            </build>
+        </profile>
+        <profile> <!-- surefire.jacoco.args property needs to be set correctly, if coverage profile is not enabled -->
+            <id>ts.jacoco.profile.not.enabled</id>
```
ctomc: you can probably remove this profile completely.
just add <surefire.jacoco.args> to top of pom.xml, similarly as we do add  <modular.jdk.args />

marekkopecky: If surefire.jacoco.args is defined on the top of pom.xml, jacoco-maven-plugin is not able to replace its value. I check it with <surefire.jacoco.args /> and with <surefire.jacoco.args ></surefire.jacoco.args >

Append property of jacoco-maven-plugin ( https://github.com/wildfly/wildfly-core/pull/2736/files#diff-600376dffeb79835ede4a0b285078036R1795 )is not related with property content. Append=true means, that jacoco adds data to exec, but doesn't replace exec file, see http://www.eclemma.org/jacoco/trunk/doc/prepare-agent-mojo.html

marekkopecky:
retest this please

marekkopecky:
retest this please

PR #2738, opened 2017-08-23
WFCORE-1374 fix security provider loading on jdk9

ctomc:
Pretty much same as https://github.com/jboss-modules/jboss-modules/commit/58cee5d7267310cbaee082cd66e0f58925461cd2


ctomc:
Just to add, with this patch, our security manager testsuite module, passes completely on JDK9

PR #2744, opened 2017-08-25
[WFCORE-3064] Suffix in Elytron periodic-rotating-file-audit-log cannot include 's' or 'S' in text

ivassile:
wildfly-elytron-integration: https://issues.jboss.org/browse/WFCORE-3064
JBEAP: https://issues.jboss.org/browse/JBEAP-12106

### DIFF
--- a/elytron/src/main/java/org/wildfly/extension/elytron/AuditResourceDefinitions.java
+++ b/elytron/src/main/java/org/wildfly/extension/elytron/AuditResourceDefinitions.java
@@ -510,10 +510,21 @@ class AuditResourceDefinitions {
             if (value.isDefined()) {
                 final String suffix = value.asString();
                 try {
-                    if (denySeconds && (suffix.contains("s") || suffix.contains("S"))) {
-                        throw ElytronSubsystemMessages.ROOT_LOGGER.suffixContainsMillis(suffix);
-                    }
                     DateTimeFormatter.ofPattern(suffix);
+                    if (denySeconds) {
+                        for (int i = 0; i < suffix.length(); i++) {
+                            char c = suffix.charAt(i);
+                            if (c == '\'') {
+                                c = suffix.charAt(++i);
+                                while (c != '\'') {
+                                    c = suffix.charAt(++i);
+                                }
+                            }
+                            if (c == 's' || c == 'S') {
+                                throw ElytronSubsystemMessages.ROOT_LOGGER.suffixContainsMillis(suffix);
+                            }
+                        }
+                    }
                 } catch (IllegalArgumentException e) {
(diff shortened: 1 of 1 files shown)


bstansberry:
@jamezp Please approve.

Also, I see this elsewhere so if its wrong we should fix it everywhere:

```
$ git grep "contains(\"s\")"
domain-management/src/main/java/org/jboss/as/domain/management/audit/validators/SuffixValidator.java:                if (denySeconds && (suffix.contains("s") || suffix.contains("S"))) {
elytron/src/main/java/org/wildfly/extension/elytron/AuditResourceDefinitions.java:                    if (denySeconds && (suffix.contains("s") || suffix.contains("S"))) {
logging/src/main/java/org/jboss/as/logging/validators/SuffixValidator.java:                if (denySeconds && (suffix.contains("s") || suffix.contains("S"))) {
```

elytron/src/main/java/org/wildfly/extension/elytron/AuditResourceDefinitions.java
```diff
@@ -510,7 +510,9 @@ public void validateParameter(String parameterName, ModelNode value) throws Oper
             if (value.isDefined()) {
                 final String suffix = value.asString();
                 try {
-                    if (denySeconds && (suffix.contains("s") || suffix.contains("S"))) {
+                    //remove text quoted using single quotes
+                    String suffixToCheck = suffix.replaceAll("'([^']*?)'", "");
+                    if (denySeconds && (suffixToCheck.contains("s") || suffixToCheck.contains("S"))) {
```
jamezp: Maybe not a huge deal as this is likely not invoked all that often but something like the following would likely perform a bit better.
```java
for (int i = 0; i < suffix.length(); i++) {
    char c = suffix.charAt(i);
    if (c == '\'') {
        c = suffix.charAt(++i);
        while (c != '\'') {
            c = suffix.charAt(++i);
        }
    }
    if (c == 's' || c == 'S') {
        throw ElytronSubsystemMessages.ROOT_LOGGER.suffixContainsMillis(suffix);
    }
}
```

Either way the `denySeconds` should probably be checked before we use regex on the string.

This is going to have to change in the other validators, the logging subsystem and management audit logging, as well. I'm okay with fixing the other two in a separate PR or making the issue more generic and fixing them all in the same PR.

ivassile: @jamezp Thanks for the review. I'll make the changes, create new issues and fix the other validators in separate PR.

ivassile:
@jamezp Small change: First validate the suffix and then check for 's' in order to prevent checking invalid formats like "m'suffix" (one single quote) which in this case will throw StringIndexOutOfBoundsException.

PR #2755, opened 2017-09-01
[WFCORE-3219] unignore AuditLogPeriodicRotatingFileHandlerTestCase.testRotate

kwart:
* https://issues.jboss.org/browse/WFCORE-3219
* https://issues.jboss.org/browse/JBEAP-12942


bstansberry:
How can we be smarter here?

darranl:
@bstansberry Are you looking for some rework on the test?

jmesnil:
How can the test be faster to run?

The test takes more than 1 minute to run because it is sleeping for 61 seconds.
Could we change the rotating period to be smaller (e.g. at the second, with `.yyyy-MM-dd-hh-mm-ss`?

wolfc:
Maybe put the test in a separate maven profile, so it only runs on CI machines?

kwart:
As this will require more work, I'm closing this PR.

PR #2757, opened 2017-09-01
[WFCORE-3218] Upgrade jboss-logmanager from 2.0.7.Final to 2.1.0.Alpha4.

jamezp:
https://issues.jboss.org/browse/WFCORE-3218

I don't think the custom byteman settings are required any longer. They passed for me locally without them.  This was done to fix some kind of issue with running on IPv6. Locally this seemed to work so it may not be required any longer.


jamezp:
To reference the original JIRA that added the byteman agent config stuff https://issues.jboss.org/browse/WFCORE-935. There are other places this could likely be removed as well, though it likely doesn't hurt having it either.

jamezp:
This test is failing for me locally as well. There were some changes in the way TCP reconnects which may be the cause. I'll place a hold on this for now.

PR #2762, opened 2017-09-05
[WFCORE-3239] CLIWrapper does not allow for use in multinode tests

bmaxwell:
https://issues.jboss.org/browse/WFCORE-3239


bstansberry:
@jfdenise Please approve. This is fairly urgent.

I don't like how the new method at https://github.com/wildfly/wildfly-core/pull/2762/files#diff-29e0b52627e394ccbfffa4de316f5944R146 results in different connect logic from what's in the current code at L111. That is, why doesn't the current L111 method call into the new method?  A good argument for not changing the L111 method in 3.x/3.0.x is to not not change any existing behavior at this late date, just add new for this specific test. But for master we can make it work the way we want it to be for the long term.

bstansberry:
OK, let's clean that part up later.

PR #2767, opened 2017-09-06
WFCORE-3226: Legacy slaves can't understand new enum version value

luck3y:


host-controller/src/main/java/org/jboss/as/domain/controller/operations/DomainServerLifecycleHandlers.java
```diff
@@ -439,6 +444,20 @@ public static void registerSuspendedStartTransformers(ResourceTransformationDesc
...
+    //         "address" => [("host-exclude" => "EAP70")],
+    //          "host-release" => "EAP7.0",
+    //  ... }
+    // TODO: WFCORE-3252 make this filtering depend on the slaves version and only send that and prior known values?
+    public static void registerHostReleaseForPriorReleaseEnumValuesTransformers(ResourceTransformationDescriptionBuilder builder) {
+        builder.addOperationTransformationOverride(ADD)
+                .setDiscard(new DiscardAttributeChecker.DiscardAttributeValueChecker(new ModelNode(HOST_RELEASE)), EAP70)
+                .addRejectCheck(RejectAttributeChecker.DEFINED, EAP70)
+                .end();
+    }
```
bstansberry: This is unused and I don't think it's needed.

bstansberry:
Once ready this will need a 3.x branch variant as well as this will end up in WF 11.

bstansberry:
As part of this work, please send up PRs to full to change the mixed domain test to use the enum.

host-controller/src/main/java/org/jboss/as/domain/controller/operations/DomainServerLifecycleHandlers.java
luck3y: Hmm, so you mean just leave it with the .addRejectCheck, but drop the .setDiscard? Ah, yes, that would make sense, since it would never be understood if it were present.

luck3y:
@bstansberry 3.x is here:

https://github.com/jbossas/wildfly-core-eap/pull/403

I'll update it with changes after I've pushed them here.

host-controller/src/main/java/org/jboss/as/domain/controller/operations/DomainServerLifecycleHandlers.java
bstansberry: Sorry; I missed where this is used in the other class; I thought the whole method was cruft and was saying it should be removed.

luck3y: Never mind, its just a missing registration for 4.1 on the existing host-exclude, thanks for correcting me :)

bstansberry:
@luck3y LGTM. Please ping me when the PR to change DomainHostExcludes700TestCase is up and we'll run a custom job to prove the combination passes.

PR #2768, opened 2017-09-06
[WFCORE-3253] Upgrade the wildfly-maven-plugin to version 1.2.0.Final

jamezp:
https://issues.jboss.org/browse/WFCORE-3253


testsuite/pom.xml
```diff
@@ -78,7 +78,7 @@
 
     <properties>
 
-	<version.org.wildfly.plugin>1.2.0.Alpha4</version.org.wildfly.plugin>
+	<version.org.wildfly.plugin>1.2.0.Final</version.org.wildfly.plugin>
```
bstansberry: Should this be in the root pom? If it had been we may not have had this issue.

If you think so but only want to change it in this PR for master/4.x, that's ok with me too.

jamezp: I guess that always has been the policy, so maybe we should move it. I'll go ahead and move it on this PR only.

PR #2771, opened 2017-09-06
Fix for WFCORE-3074, replaced Path matching

jfdenise:
Replaced Filesystem based matching with PathAddress/PathElement based matching.
Added unit test for path matching. Added unit test to cover the case that was failing with Filesystem based matching (with a possible provider address's value containing a ":" character).

### DIFF
--- a/controller/src/main/java/org/jboss/as/controller/PathAddress.java
+++ b/controller/src/main/java/org/jboss/as/controller/PathAddress.java
@@ -451,6 +451,49 @@ public class PathAddress implements Iterable<PathElement> {
         return toString('/');
     }
 
+    /**
+     * Check if this path matches the address path.
+     * An address matches this address if its path elements match or are valid
+     * multi targets for this path elements. Addresses that are equal are matching.
+     *
+     * @param address The path to check against this path. If null, this method
+     * returns false.
+     * @return true if the provided path matches, false otherwise.
+     */
+    public boolean matches(PathAddress address) {
+        if (address == null) {
+            return false;
+        }
+        if (equals(address)) {
+            return true;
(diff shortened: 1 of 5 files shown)


controller/src/main/java/org/jboss/as/controller/PathElement.java
```diff
@@ -177,6 +177,15 @@ public boolean matches(Property property) {
     }
 
     /**
+     * Determine whether the given element matches this element.
```
bstansberry: This needs javadoc to describe the matching algorithm, i.e. that it differs from equals in that wildcards are involved.

A couple subtleties worth noting:

1) This only differs from equals(pe) if this.isWildCard() but not if only pe.isWildCard().
2) PathElement also has the isMultiTarget() concept, which is goes beyond isWildCard() and isn't factored into this matches() method. Perhaps it should be though, since this is a general purpose public method in a very widely used class.

The way that isMultiTarget() goes beyond isWildCard() isn't relevant to the CapabilityRegistry use case here, as no such addresses will be stored in the registry.

controller/src/main/java/org/jboss/as/controller/PathAddress.java
```diff
@@ -451,6 +451,33 @@ public String toPathStyleString() {
         return toString('/');
     }
 
+    /**
+     * Check if this path matches the address path.
```
bstansberry: This could also use a bit of javadoc mentioning how it differs from equals by checking if the elements match rather than if they are equal.

jfdenise:
take into account multiTargets when matching, updated javadoc, removed stream usage.

PR #2773, opened 2017-09-06
Work around the WFCORE-2235 problem to avoid intermittent failures fo…

bstansberry:
…r the known issue

Related to but not a fix for https://issues.jboss.org/browse/WFCORE-2235

Example failure I'm trying to avoid: https://ci.wildfly.org/viewLog.html?buildId=72854&buildTypeId=WildFlyCore_PullRequestWindows&tab=buildResultsDiv


ropalka:
Avoiding optional dependencies everywhere is right thing to do.

bstansberry:
@ropalka This isn't an optional MSC dep, it's an optional module dep, and that needs to work.

IIIRC it's implemented via an optional MSC dep (or was at one point) but that's our problem, not the user's problem.

PR #2777, opened 2017-09-08
[WFCORE-3259] Fail if running the transformed boot ops on the legacy …

bstansberry:
…slave doesn't succeed


bstansberry:
I added the downstream-update-required label as before we can merge this we'll have to fix the tests in full that that CI shows as being broken.

PR #2779, opened 2017-09-08
WFCORE-3261 Unignore NotificationSupportImplTestCase.doManyNotificati…

jmartisk:
…onOrderingWithExecutor

https://issues.jboss.org/browse/WFCORE-3261


bstansberry:
@jmesnil Do you want this unignored? Does running it add value to the testsuite proportionate to the time spent? If not we should just comment out the @Test since it seems we're on a drive to eliminate use of @Ignore. I looks a bit like experimental/debugging type stuff.

jmartisk:
I reduced the number of iterations to make this run within 2 seconds. But if we decide that this test doesn't add enough value, I'm not opposed to removing it instead.

jmesnil:
@jmartisk let's remove it then. It adds no value to the unit test we already have. thanks!

jmartisk:
I updated the PR, the test is now removed.

PR #2780, opened 2017-09-08
[WFCORE-2154, JBEAP-4904] Enhance a CLI logging

xstefank:
JBEAP: https://issues.jboss.org/browse/JBEAP-4904
upstreams: https://issues.jboss.org/browse/WFCORE-2154, https://issues.jboss.org/browse/WFCORE-1583

including also
JBEAP: https://issues.jboss.org/browse/JBEAP-12970
upstream: https://issues.jboss.org/browse/WFCORE-3262

### DIFF
--- a/cli/pom.xml
+++ b/cli/pom.xml
@@ -144,6 +144,22 @@
             <groupId>org.jboss.logging</groupId>
             <artifactId>jboss-logging</artifactId>
         </dependency>
+        <dependency>
+            <groupId>org.jboss.logging</groupId>
+            <artifactId>jboss-logging-annotations</artifactId>
+            <!-- This is a compile-time dependency of this project, but is not needed at compile or runtime by other
+                  projects that depend on this project.-->
+            <scope>provided</scope>
+            <optional>true</optional>
+        </dependency>
+        <dependency>
+            <groupId>org.jboss.logging</groupId>
+            <artifactId>jboss-logging-processor</artifactId>
+            <!-- This is a compile-time dependency of this project, but is not needed at compile or runtime by other
+                  projects that depend on this project.-->
+            <scope>provided</scope>
+            <optional>true</optional>
+        </dependency>
         <!-- Used for the complete client JAR -->
         <dependency>
             <groupId>org.jboss.logmanager</groupId>
--- a/cli/src/main/java/org/jboss/as/cli/accesscontrol/BaseAccessRequirement.java
+++ b/cli/src/main/java/org/jboss/as/cli/accesscontrol/BaseAccessRequirement.java
@@ -25,7 +25,7 @@ package org.jboss.as.cli.accesscontrol;
 import org.jboss.as.cli.CliEvent;
 import org.jboss.as.cli.CliEventListener;
 import org.jboss.as.cli.CommandContext;
-import org.jboss.logging.Logger;
+import org.jboss.as.cli.logger.CliLogger;
 
 
 /**
@@ -34,17 +34,16 @@ import org.jboss.logging.Logger;
  */
 public abstract class BaseAccessRequirement implements AccessRequirement, CliEventListener {
 
-    protected Logger log = Logger.getLogger(getClass());
-    protected final boolean traceEnabled = log.isTraceEnabled();
-
     private Boolean satisfied;
 
     @Override
     public boolean isSatisfied(CommandContext ctx) {
         if(satisfied == null) {
             satisfied = checkAccess(ctx);
-            if(traceEnabled) {
-                log.trace(toString() + " " + satisfied);
+            if(satisfied) {
+                CliLogger.ROOT_LOGGER.tracef("Requirement %s is satisfied", this);
+            } else {
+                CliLogger.ROOT_LOGGER.tracef("Requirement %s is not satisfied", this);
             }
         }
         return satisfied;
--- a/cli/src/main/java/org/jboss/as/cli/handlers/OperationRequestHandler.java
+++ b/cli/src/main/java/org/jboss/as/cli/handlers/OperationRequestHandler.java
@@ -37,6 +37,7 @@ import org.jboss.as.cli.CommandLineException;
 import org.jboss.as.cli.OperationCommand;
 import org.jboss.as.cli.RequestWithAttachments;
 import org.jboss.as.cli.Util;
+import org.jboss.as.cli.logger.CliLogger;
 import org.jboss.as.cli.operation.impl.DefaultCallbackHandler;
 import org.jboss.as.controller.client.ModelControllerClient;
 import org.jboss.as.controller.client.Operation;
@@ -87,6 +88,7 @@ public class OperationRequestHandler implements CommandHandler, OperationCommand
         }
 
         try {
+            CliLogger.ROOT_LOGGER.debugf("Executing operation request:%n%s", op.getOperation());
             final ModelNode result = ctx.execute(op, "Operation request");
             if (Util.isSuccess(result)) {
                 ctx.printLine(result.toString());
--- a/cli/src/main/java/org/jboss/as/cli/impl/CliConfigImpl.java
+++ b/cli/src/main/java/org/jboss/as/cli/impl/CliConfigImpl.java
@@ -42,9 +42,9 @@ import org.jboss.as.cli.CliInitializationException;
 import org.jboss.as.cli.CommandContext;
 import org.jboss.as.cli.ControllerAddress;
 import org.jboss.as.cli.SSLConfig;
+import org.jboss.as.cli.logger.CliLogger;
 import org.jboss.as.cli.util.CLIExpressionResolver;
 import org.jboss.as.protocol.StreamUtils;
-import org.jboss.logging.Logger;
 import org.jboss.staxmapper.XMLElementReader;
 import org.jboss.staxmapper.XMLExtendedStreamReader;
 import org.jboss.staxmapper.XMLMapper;
@@ -84,7 +84,6 @@ class CliConfigImpl implements CliConfig {
     private static final String ECHO_COMMAND = "echo-command";
     private static final String COMMAND_TIMEOUT = "command-timeout";
 
-    private static final Logger log = Logger.getLogger(CliConfig.class);
 
     static CliConfig load(final CommandContext ctx) throws CliInitializationException {
         return load(ctx, null);
@@ -585,9 +584,7 @@ class CliConfigImpl implements CliConfig {
                         config.silent = resolveBoolean(reader.getElementText());
                     } else if(localName.equals(ACCESS_CONTROL)) {
                         config.accessControl = resolveBoolean(reader.getElementText());
-                        if(log.isTraceEnabled()) {
-                            log.trace(ACCESS_CONTROL + " is " + config.accessControl);
-                        }
+                        CliLogger.ROOT_LOGGER.tracef("Access-control is %s", config.accessControl);
                     } else {
                         throw new XMLStreamException("Unexpected element: " + localName);
                     }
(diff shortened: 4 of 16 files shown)


xstefank:
@jfdenise can you review this PR please?

cli/src/main/java/org/jboss/as/cli/impl/ReadlineConsole.java
```diff
@@ -140,10 +137,7 @@
         CLITerminalConnection(Terminal terminal) {
             super(terminal);
             interceptor = (int[] ints) -> {
-                if (LOG.isLoggable(Level.FINER)) {
-                    LOG.log(Level.FINER, "Writing {0}",
-                            Parser.stripAwayAnsiCodes(Parser.fromCodePoints(ints)));
-                }
+                CliLogger.ROOT_LOGGER.writingCodePoints(Parser.stripAwayAnsiCodes(Parser.fromCodePoints(ints)));
```
jfdenise: Perhaps "writingToTerminal" instead of writingCodePoints

cli/pom.xml
```diff
@@ -144,6 +144,18 @@
             <groupId>org.jboss.logging</groupId>
             <artifactId>jboss-logging</artifactId>
         </dependency>
+        <dependency>
+            <groupId>org.jboss.logging</groupId>
+            <artifactId>jboss-logging-annotations</artifactId>
```
jamezp: This should be marked with:
```
<scope>provided</scope>
<optional>true</optional>
```

cli/src/main/java/org/jboss/as/cli/accesscontrol/BaseAccessRequirement.java
```diff
@@ -34,17 +34,16 @@
...
     public boolean isSatisfied(CommandContext ctx) {
         if(satisfied == null) {
             satisfied = checkAccess(ctx);
-            if(traceEnabled) {
-                log.trace(toString() + " " + satisfied);
+            if(satisfied) {
+                CliLogger.ROOT_LOGGER.accessRequirementSatisfied(toString());
+            } else {
+                CliLogger.ROOT_LOGGER.accessRequirementNotSatisfied(toString());
             }
```
jamezp: This should probably be
```java
if (CliLogger.ROOT_LOGGER.isTraceEnabled()) {
    if(satisfied) {
        CliLogger.ROOT_LOGGER.accessRequirementSatisfied(toString());
    } else {
        CliLogger.ROOT_LOGGER.accessRequirementNotSatisfied(toString());
    }
}
```

The `toString()` may be expensive.

cli/src/main/java/org/jboss/as/cli/handlers/OperationRequestHandler.java
```diff
@@ -87,6 +88,7 @@ public void handle(CommandContext ctx) throws CommandLineException {
         }
 
         try {
+            CliLogger.ROOT_LOGGER.executingOperationRequest(op.getOperation().toString());
```
jamezp: This is okay, but debug logging isn't required in a logger interface. Also {{toString()}} shouldn't be required. If debug is enabled the logger will essentially us `String.valueOf()` for the `op.getOperation()` value.

cli/src/main/java/org/jboss/as/cli/accesscontrol/BaseAccessRequirement.java
jamezp: I guess the other option would be to use.
```java
if(satisfied) {
    CliLogger.ROOT_LOGGER.accessRequirementSatisfied(this);
} else {
    CliLogger.ROOT_LOGGER.accessRequirementNotSatisfied(this);
}
```
You'd just need to change the `accessRequirementNotSatisfied()` method to accept a `BaseAccessRequirement` or just `Object`. Either should be fine though.

cli/src/main/java/org/jboss/as/cli/impl/CommandContextImpl.java
```diff
@@ -1076,9 +1065,7 @@ public void connectController(String controller) throws CommandLineException {
         do {
             try {
                 CallbackHandler cbh = new AuthenticationCallbackHandler(username, password);
-                if (log.isDebugEnabled()) {
-                    log.debug("connecting to " + connectionAddress.getHost() + ':' + connectionAddress.getPort() + " as " + username);
-                }
+                CliLogger.ROOT_LOGGER.connectingToHostAs(getAddressString(connectionAddress), username);
```
jamezp: This really could just be;
```java
if (CliLogger.ROOT_LOGGER.isDebugEnabled()) {
    CliLogger.ROOT_LOGGER.debugf("Connecting to %s:%d as %s", connectionAddress.getHost(), connectionAddress.getPort()));
}
```

cli/src/main/java/org/jboss/as/cli/impl/ReadlineConsole.java
```diff
@@ -140,10 +137,7 @@
         CLITerminalConnection(Terminal terminal) {
             super(terminal);
             interceptor = (int[] ints) -> {
-                if (LOG.isLoggable(Level.FINER)) {
-                    LOG.log(Level.FINER, "Writing {0}",
-                            Parser.stripAwayAnsiCodes(Parser.fromCodePoints(ints)));
-                }
+                CliLogger.ROOT_LOGGER.writingToTerminal(Parser.stripAwayAnsiCodes(Parser.fromCodePoints(ints)));
```
jamezp: This should check whether or not debug is enabled.

cli/src/main/java/org/jboss/as/cli/impl/ReadlineConsole.java
```diff
@@ -471,19 +462,15 @@ private CLITerminalConnection newConnection() throws IOException {
...
     }
 
     @Override
     public void addCompleter(CommandLineCompleter completer) {
         completions.add((Completion) (CompleteOperation co) -> {
             List<String> candidates = new ArrayList<>();
-            if (LOG.isLoggable(Level.FINER)) {
-                LOG.log(Level.FINER, "Completing {0}", co.getBuffer());
-            }
+            CliLogger.ROOT_LOGGER.completingBuffer(co.getBuffer());
```
jamezp: This should probably check if trace is enabled. I'm guessing `co.getBuffer()` may be expensive.

cli/src/main/java/org/jboss/as/cli/logger/CliLogger.java
```diff
@@ -0,0 +1,235 @@
+package org.jboss.as.cli.logger;
```
jamezp: This needs a copyright header.

cli/src/main/java/org/jboss/as/cli/logger/CliLogger.java
```diff
@@ -0,0 +1,235 @@
...
+import org.jboss.logging.annotations.Cause;
+import org.jboss.logging.annotations.LogMessage;
+import org.jboss.logging.annotations.Message;
+import org.jboss.logging.annotations.MessageLogger;
+
+import java.util.Collection;
+import java.util.List;
+
+@MessageLogger(projectCode = "WFLYCLI")
+public interface CliLogger {
```
jamezp: This should extend the `BasicLogger` so that you can use debug and trace logging from the `CliLogger`.

jamezp: Two other general comments. I don't like the `BASE + x`. I'd also prefer using the default `PRINTF` format over the `MESSAGE_FORMAT`. It's much more robust.

jamezp: Also trace and debug logging don't need to be in here. Those are not required to be internationalized.

cli/src/main/java/org/jboss/as/cli/logger/CliLogger.java
```diff
@@ -0,0 +1,235 @@
...
+    @LogMessage(level = Logger.Level.DEBUG)
+    @Message(id = BASE + 51, value = "Error when parsing last chunk of operation request", format = Message.Format.MESSAGE_FORMAT)
+    void errorParsingLastChunk(@Cause Exception cause);
+
+    @LogMessage(level = Logger.Level.TRACE)
+    @Message(id = BASE + 52, value = "Error getting capability names", format = Message.Format.MESSAGE_FORMAT)
+    void errorGettingCapabilityNames(@Cause Exception cause);
+
+    @LogMessage(level = Logger.Level.DEBUG)
+    @Message(id = BASE + 53, value = "Executing operation request:\n{0}", format = Message.Format.MESSAGE_FORMAT)
```
jamezp: This should be changed to:
```java
@Message(id = 153, value = "Executing operation request:%n%s")
```

jamezp:
Just something to consider. This changes all the logging categories. Currently if you just want to see logging from specific packages it would work. With this change if you enable trace logging on `org.jboss.as.cli.logger` you'd get trace logging for all packages. Not a big deal as we do this in WildFly in general, however just something to consider.

cli/src/main/java/org/jboss/as/cli/logger/CliLogger.java
xstefank: @jamezp thank you for the review. Would you prefer me to move DEBUG/TRACE messages out of CliLogger?

jamezp: It doesn't matter too much to me, but the rest of WildFly Core and WildFly loggers do not include DEBUG and TRACE messages in the interface. If you extend the `BasicLogger` interface you can just use `CliLogger.ROOT_LOGGER.debugf("Some debug message with %s", parameter)`.

cli/src/main/java/org/jboss/as/cli/logger/CliLogger.java
jamezp: To clarify what I mean here is we shouldn't be using `\n` as on Windows this won't print as a new line. Using the `MessageFormat.PRINTF` and `%n` will make it platform independent.

xstefank: @jamezp Thanks for the comments, I will update the PR

bstansberry:
I'm not a big fan of bringing in the overhead of a CLILogger. We do not intend to localize the CLI and the benefit of message codes is pretty low.

jfdenise:
I like the fact that all messages are in a single location, we can track redundant messages and homogenise them. I think that is it cleaner than it used to be. If the added dependency is an issue, then that need to be revised.

cli/src/main/java/org/jboss/as/cli/impl/CliConfigImpl.java
```diff
@@ -585,9 +584,7 @@ public void readCLIElement_2_0(XMLExtendedStreamReader reader, Namespace expecte
                         config.silent = resolveBoolean(reader.getElementText());
                     } else if(localName.equals(ACCESS_CONTROL)) {
                         config.accessControl = resolveBoolean(reader.getElementText());
-                        if(log.isTraceEnabled()) {
-                            log.trace(ACCESS_CONTROL + " is " + config.accessControl);
-                        }
+                        CliLogger.ROOT_LOGGER.tracef("Access-control is {0}", config.accessControl);
```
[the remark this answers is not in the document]
xstefank: sorry, I don't know how I missed this

jamezp:
The logging interfaces do add some overhead. I am approving this only based on the analysis of the logging interface and the log messages.

xstefank:
Closing per discussion, I will create a different PR for https://issues.jboss.org/browse/JBEAP-4904 with different approach

PR #2782, opened 2017-09-08
[3.x] WFCORE-3267 - Authorization identity forwarding not exposed to configuration

n1hility:
https://issues.jboss.org/browse/WFCORE-3267


ctomc:
@bstansberry @n1hility this is wrong!
We don't need 
- model bump
- xsd schema bump
- transformers

None of this is needed as there is no .GA / Final version of server that shipped with 1.0 version of schema / model.
This just adds extra clutter without any need for it.

PR #2791, opened 2017-09-11
WFCORE-3267 Update security domain and forwarding mode to support exp…

n1hility:
…ressions


elytron/src/main/java/org/wildfly/extension/elytron/AuthenticationClientDefinitions.java
```diff
@@ -120,12 +120,14 @@
             .build();
 
     static final SimpleAttributeDefinition SECURITY_DOMAIN = new SimpleAttributeDefinitionBuilder(ElytronDescriptionConstants.SECURITY_DOMAIN, ModelType.STRING, true)
+            .setAllowExpression(true)
```
bstansberry: This should not be an expression. Model references should not be expressions as they can't be resolved in Stage.MODEL 100% of the time.

PR #2798, opened 2017-09-12
Revert "Work around the WFCORE-2235 problem to avoid intermittent failures fo…"

bstansberry:
Reverts wildfly/wildfly-core#2773

The test still intermittently fails despite this change, so it wasn't helping with that and we should restore the original version.

I'm tempted to @ Ignore the entire test but it covers an important scenario. Perhaps the issue isn't the "optional" bit but rather the chained dependencies. That's still an important scenario but less so than the test overall and maybe disabling that bit will avoid the intermittent issue. 

@ropalka FYI


ropalka:
I just successfully executed this test 100 000 times
with my latest MSC changes. There was no failure :)

ropalka:
@bstansberry ^^^

bstansberry:
@ropalka :)

bstansberry:
@ropalka Was that with this revert in place or with the current master that has the "optional" bit commented out?

ropalka:
It was with revert in place @bstansberry . I successfully fixed broken MSC optional dependencies implementation.

bstansberry:
Thanks! In that case I'll merge this.

PR #2739, opened 2017-08-23
[WFCORE-3204] Upgrade WildFly Elytron to 1.2.0.Beta3

darranl:
https://issues.jboss.org/browse/WFCORE-3204

PR #2799, opened 2017-09-12
[WFCORE-3279] WildFly Elytron Component Upgrades

darranl:
Upstream PR https://github.com/wildfly/wildfly-core/pull/2739

https://issues.jboss.org/browse/WFCORE-3279
https://issues.jboss.org/browse/WFCORE-3280
https://issues.jboss.org/browse/WFCORE-3281


bstansberry:
@darranl please update #2739 with the other two components.

darranl:
@bstansberry The upstream PR doesn't require component upgrades yet, they don't actually contain any fixes yet.

The WildFly Elytron Tool component upgrade here is just to ensure the tool is built using the same Elytron version as is used by the application server.  The Elytron Web component upgrade is just to ensure when the EAP 7.1 CP process begins the starting point contains an update to a test case, there is no bug fix in there.

PR #2803, opened 2017-09-13
Component updates

dmlloyd:


bstansberry:
@dmlloyd The wildfly-core-eap/3.0.x PR doesn't have the modules update.

This is admittedly a pain but these branches are meant to be kept in sync except wildfly-core-eap/3.0.x has "eapification" changes like putting -redhat builds into the pom. We want the 3.0.2.Final we produce here and the 3.0.2.Final-redhat we produce there to be the same code except for the eapification.

Once we are done produce community core 3.0.x releases for WF 11 Final I'll retire this branch.

bstansberry:
@dmlloyd to clarify -- this branch should only have changes we can put in 3.0.x.

dmlloyd:
> @dmlloyd The wildfly-core-eap/3.0.x PR doesn't have the modules update.
> 
> This is admittedly a pain but these branches are meant to be kept in sync except wildfly-core-eap/3.0.x has "eapification" changes like putting -redhat builds into the pom. We want the 3.0.2.Final we produce here and the 3.0.2.Final-redhat we produce there to be the same code except for the eapification.

There is no EAP JIRA for the modules update, but Swarm will need it.  So this puts me in a bit of a bind...

bstansberry:
retest this please

bstansberry:
retest this please

bstansberry:
retest this please

PR #2804, opened 2017-09-13
[WFCORE-3257] add-identity filesystem-realm operation in Elytron subsystem might fail in batch-mode

yersan:
The add-identity operation requires `ServiceController<ModifiableSecurityRealm>` service in UP state. In some circumstances, this service controller could require more time to start, generating failures in batch-mode if add-identity is called after adding the file-system-realm.  This patch gives a chance to get the service UP waiting until 500 milliseconds.

Issue: https://issues.jboss.org/browse/WFCORE-3257
JBEAP issue: https://issues.jboss.org/browse/JBEAP-13038


yersan:
restest this please

PR #2805, opened 2017-09-13
[WFCORE-3286] - It is possible to add multiple http-connectors to the…

JiriOndrusek:
… same undertow listener

Issue: https://issues.jboss.org/browse/WFCORE-3286

### DIFF
--- a/remoting/subsystem/src/main/java/org/jboss/as/remoting/HttpConnectorAdd.java
+++ b/remoting/subsystem/src/main/java/org/jboss/as/remoting/HttpConnectorAdd.java
@@ -28,7 +28,9 @@ import static org.jboss.as.remoting.CommonAttributes.SECURITY_REALM;
 import org.jboss.as.controller.AbstractAddStepHandler;
 import org.jboss.as.controller.OperationContext;
 import org.jboss.as.controller.OperationFailedException;
+import org.jboss.as.controller.OperationStepHandler;
 import org.jboss.as.controller.PathAddress;
+import org.jboss.as.controller.logging.ControllerLogger;
 import org.jboss.as.controller.registry.Resource;
 import org.jboss.dmr.ModelNode;
 import org.jboss.msc.service.ServiceName;
@@ -48,6 +50,13 @@ public class HttpConnectorAdd extends AbstractAddStepHandler {
     private HttpConnectorAdd() {
         super(HttpConnectorResource.CONNECTOR_REF, HttpConnectorResource.AUTHENTICATION_PROVIDER, HttpConnectorResource.SECURITY_REALM,
                 HttpConnectorResource.SASL_AUTHENTICATION_FACTORY, ConnectorCommon.SASL_PROTOCOL, ConnectorCommon.SERVER_NAME);
+    }
+
+    @Override
+    protected void populateModel(OperationContext context, ModelNode operation, Resource resource) throws OperationFailedException {
+        super.populateModel(context, operation, resource);
+
+        context.addStep(operation, HttpConnectorValidationStep.INSTANCE, OperationContext.Stage.MODEL);
     }
 
     @Override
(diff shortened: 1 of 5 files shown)


remoting/subsystem/src/main/java/org/jboss/as/remoting/HttpConnectorAdd.java
```diff
@@ -54,6 +55,9 @@ private HttpConnectorAdd() {
     protected void performRuntime(OperationContext context, ModelNode operation, Resource resource) throws OperationFailedException {
         final PathAddress address = PathAddress.pathAddress(operation.get(OP_ADDR));
         final String connectorName = address.getLastElement().getValue();
+
+        validateDuplicatedResources(context, address, connectorName, resource);
```
ctomc: any validation should be done in stage MODEL not runtime!

remoting/tests/pom.xml
```diff
@@ -82,5 +82,10 @@
             <type>pom</type>
             <scope>test</scope>
         </dependency>
+        <dependency>
+            <groupId>org.mockito</groupId>
+            <artifactId>mockito-all</artifactId>
```
ctomc: mockito-core not -all

JiriOndrusek:
@ctomc Thank you for your response. Both problems are fixed.

remoting/subsystem/src/main/java/org/jboss/as/remoting/HttpConnectorAdd.java
```diff
@@ -51,6 +52,14 @@ private HttpConnectorAdd() {
     }
 
     @Override
+    protected void populateModel(OperationContext context, ModelNode operation, Resource resource) throws OperationFailedException {
+        super.populateModel(context, operation, resource);
+        final PathAddress address = PathAddress.pathAddress(operation.get(OP_ADDR));
+        final String connectorName = address.getLastElement().getValue();
```
ctomc: shortcut for this is, context.getCurrentAddressValue();

remoting/subsystem/src/main/java/org/jboss/as/remoting/HttpConnectorAdd.java
```diff
@@ -72,4 +81,16 @@ void launchServices(OperationContext context, String connectorName, ModelNode mo
         final String saslAuthenticationFactory = saslAuthenticationFactoryModel.isDefined() ? saslAuthenticationFactoryModel.asString() : null;
         RemotingHttpUpgradeService.installServices(context, connectorName, connectorRef, RemotingServices.SUBSYSTEM_ENDPOINT, optionMap, securityRealm, saslAuthenticationFactory);
     }
+
+    void validateDuplicatedResources(OperationContext context, PathAddress address, String connectorName, Resource resource) throws OperationFailedException {
+        Resource parent = context.readResourceFromRoot(address.getParent(), true);
```
ctomc: This will have a big performance impact.
I think without recursively reading it, it should still work fine for this use case.

remoting/subsystem/src/main/java/org/jboss/as/remoting/HttpConnectorAdd.java
JiriOndrusek: @ctomc thank you, I agree, I'll make this change in the next commit.

JiriOndrusek:
@ctomc Hi, both problems were fixed.
First one (using shortcut) was without problems.
Second one (do not use recursive read was complicated). I had to add non-recursive read for all children (because in other case I couldn't read connector-ref property) I think that N+1 non-recursive reads (where N is number of children) is in performance perspective better then 1 recursive read. 
I'd like to ask you for review of these changes.

remoting/tests/pom.xml
```diff
@@ -82,5 +82,10 @@
             <type>pom</type>
             <scope>test</scope>
         </dependency>
+        <dependency>
+            <groupId>org.mockito</groupId>
+            <artifactId>mockito-core</artifactId>
+            <scope>test</scope>
+        </dependency>
```
bstansberry: -1 to mocks like this.

The subsystem test framework allows you to create a config you want and execute management ops against it. There should be no need for mocks.

remoting/subsystem/src/main/java/org/jboss/as/remoting/HttpConnectorAdd.java
```diff
@@ -51,6 +52,14 @@ private HttpConnectorAdd() {
     }
 
     @Override
+    protected void populateModel(OperationContext context, ModelNode operation, Resource resource) throws OperationFailedException {
+        super.populateModel(context, operation, resource);
+        final PathAddress address = context.getCurrentAddress();
+        final String connectorName = address.getLastElement().getValue();
+        validateDuplicatedResources(context, address, connectorName, resource);
```
bstansberry: This needs to be wrapped in an added step so it executes after other ops in the same composite, since those other ops might correct the condition the validator will reject.

remoting/subsystem/src/main/java/org/jboss/as/remoting/HttpConnectorAdd.java
```diff
@@ -72,4 +81,18 @@ void launchServices(OperationContext context, String connectorName, ModelNode mo
         final String saslAuthenticationFactory = saslAuthenticationFactoryModel.isDefined() ? saslAuthenticationFactoryModel.asString() : null;
         RemotingHttpUpgradeService.installServices(context, connectorName, connectorRef, RemotingServices.SUBSYSTEM_ENDPOINT, optionMap, securityRealm, saslAuthenticationFactory);
     }
+
+    void validateDuplicatedResources(OperationContext context, PathAddress address, String connectorName, Resource resource) throws OperationFailedException {
+        PathAddress parentAddress = address.getParent();
+        Resource parent = context.readResourceFromRoot(parentAddress, false);
+        ModelNode resourceRef = resource.getModel().get(CommonAttributes.CONNECTOR_REF);
+        boolean sameUndertowListener = parent.getChildren(CommonAttributes.HTTP_CONNECTOR).stream()
```
bstansberry: Please just use a for loop. 3 lambdas = over 2K in code space vs a couple of if statements.

JiriOndrusek:
@bstansberry thank you for your response. I've fixed all 3 comments.

PR #2806, opened 2017-09-13
Fix for WFCORE-3289, variable substitution completion issue

jfdenise:
Handling variable substitution:
- fixed operation completion in commands (echo-dmr, if, edit-batch-line, attachment)
- support for multiple variables in a path

Added new completion unit tests.

### DIFF
--- a/cli/src/main/java/org/jboss/as/cli/parsing/ParserUtil.java
+++ b/cli/src/main/java/org/jboss/as/cli/parsing/ParserUtil.java
@@ -26,6 +26,7 @@ import org.jboss.as.cli.CommandContext;
 import org.jboss.as.cli.CommandFormatException;
 import org.jboss.as.cli.CommandLineFormat;
 import org.jboss.as.cli.operation.CommandLineParser;
+import org.jboss.as.cli.parsing.StateParser.SubstitutedLine;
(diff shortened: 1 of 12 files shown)


cli/src/main/java/org/jboss/as/cli/parsing/ParserUtil.java
```diff
@@ -74,20 +75,36 @@ public static String parse(String commandLine, final CommandLineParser.CallbackH
...
-        return StateParser.parse(commandLine, callbackHandler, InitialState.INSTANCE, strict, ctx);
+        return StateParser.parseLine(commandLine, callbackHandler, InitialState.INSTANCE, strict, ctx);
     }
 
     /**
      * Returns the string which was actually parsed with all the substitutions performed
      */
     public static String parseOperationRequest(String commandLine, final CommandLineParser.CallbackHandler handler) throws CommandFormatException {
-        if(commandLine == null) {
+        return parseOperationRequestLine(commandLine, handler).getSubstitued();
```
bstansberry: This can NPE if commandLine == null.

bstansberry:
Other than the possible NPE, looks good.

jfdenise:
@bstansberry , thank-you. Fixed.

PR #2809, opened 2017-09-14
[WFCORE-3290] Transformer shouldn't reject remoting endpoint resource…

bstansberry:
… if no attributes are configured

https://issues.jboss.org/browse/WFCORE-3290


bstansberry:
@luck3y FYI

luck3y:
Thanks for looking at that @bstansberry. I think it would have taken me quite some time to come to the same conclusion. :)

PR #2810, opened 2017-09-14
[WFCORE-3290] Transformer shouldn't reject remoting endpoint resource…

bstansberry:
… if no attributes are configured

https://issues.jboss.org/browse/WFCORE-3290


bstansberry:
Not currently scheduled.

PR #2811, opened 2017-09-14
Fix for WFCORE-2283. Completion for alternatives,requires and required

jfdenise:


cli/src/main/java/org/jboss/as/cli/operation/impl/DefaultOperationCandidatesProvider.java
```diff
@@ -210,6 +212,193 @@ public CommandLineCompleter getCompleter() {
...
+                    if (!invalidProperties.contains(prop.getName())) {
+                        Set<String> requires = requiresMap.get(prop.getName());
+                        if (requires != null) {
+                            for (String req : requires) {
+                                // A require could be hidden because it is in alternatives with a present property.
+                                if (!invalidProperties.contains(req)) {
+                                    Set<String> requiring = requiredBy.get(req);
+                                    if (requiring == null) {
+                                        requiring = new HashSet<>();
+                                        requiredBy.put(req, requiring);
```
bstansberry: Please ignore this if I'm wrong; i.e. it's a lot of code and there's no need to waste your time explaining how I'm wrong. ;) But I don't see any reads of the Set<String> values stored in requiredBy. I just see at L355 "return requiredBy.get(prop.getName()) != null;". So it seems that requiredBy could just be a Set<String> and not a map.

cli/src/main/java/org/jboss/as/cli/operation/impl/DefaultOperationCandidatesProvider.java
```diff
@@ -210,6 +212,193 @@ public CommandLineCompleter getCompleter() {
...
+        private final Set<String> presentProperties;
+        private final String radical;
+        private final boolean all;
+        private final boolean multiple;
+        private final Set<String> invalidProperties = new HashSet<>();
+        private final List<Property> propList;
+        private final Map<String, Set<String>> requiredBy = new HashMap<>();
+        public PropertyVisibility(boolean all, List<Property> propList,
+                Set<String> presentProperties, String radical) {
+            this.all = all;
```
bstansberry: I don't see this set to 'true' anywhere. Is that planned for the future?

server/src/main/java/org/jboss/as/server/controller/resources/DeploymentAttributes.java
```diff
@@ -198,12 +198,14 @@ public void marshallAsAttribute(AttributeDefinition attribute, ModelNode resourc
                     ModelDescriptionConstants.INPUT_STREAM_INDEX, ModelDescriptionConstants.HASH, ModelDescriptionConstants.BYTES,
                     ModelDescriptionConstants.URL, ModelDescriptionConstants.EMPTY)
                     .setRequires(ModelDescriptionConstants.PATH)
+                    .setRequired(false)
```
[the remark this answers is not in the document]
bstansberry: That PR is merged.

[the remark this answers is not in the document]
bstansberry: It turns out the issues you're fixing here and those Chao Wang was addressing are different. And it seems there's no conflict. What you've done in this class looks fine.

cli/src/main/java/org/jboss/as/cli/operation/impl/DefaultOperationCandidatesProvider.java
jfdenise: Good point, it was planed at some point but removed, I removed it.

cli/src/main/java/org/jboss/as/cli/operation/impl/DefaultOperationCandidatesProvider.java
jfdenise: Yes, no need for a map. I think that I used it to help debug. Replaced.

testsuite/standalone/src/test/java/org/jboss/as/test/integration/management/cli/CliCompletionTestCase.java
```diff
@@ -89,4 +91,120 @@ public void test() throws Exception {
...
+
+            {
+                String cmd = op + "[{empty=true,archive=true";
+                final List<String> candidates = new ArrayList<>();
+                ctx.getDefaultCommandCompleter().complete(ctx,
+                        cmd, cmd.length(), candidates);
+                assertEquals(Arrays.asList("}"), candidates);
+            }
+
+            {
```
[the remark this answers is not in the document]
jfdenise: Ok, done.

bstansberry:
@jfdenise After merging #2847 and #2903 this one needs a rebase.

jfdenise:
@bstansberry , unit test rebased.

PR #2815, opened 2017-09-14
[WFCORE-3292] Use correct attributes for remoting configuration in th…

jamezp:
…e wildfly-config.xml for tests.

https://issues.jboss.org/browse/WFCORE-3292

This likely doesn't create any real issues with it being incorrect. However it should probably use the correct attributes.


jamezp:
@dmlloyd Does this look correct?

PR #2819, opened 2017-09-15
Fix for WFCORE-426, CLI, automatic long output pause

jfdenise:
Automatic paging of output longer than terminal height. Only applies in interactive mode when CLI runs in a terminal.
Evolves dependency to aesh-readline 1.1. Manual testing only.


stalep:
This looks fine from what i can see.

PR #2820, opened 2017-09-15
[JBEAP-12826, WFCORE-3203] WildFlyTestRunner throws NullPointerExcept…

xstefank:
…ion during stop intermittently

JBEAP: https://issues.jboss.org/browse/JBEAP-12826
WFCORE: https://issues.jboss.org/browse/WFCORE-3203

### DIFF
--- a/testsuite/test-runner/src/main/java/org/wildfly/core/testrunner/Server.java
+++ b/testsuite/test-runner/src/main/java/org/wildfly/core/testrunner/Server.java
@@ -19,6 +19,7 @@ import java.nio.file.Path;
 import java.nio.file.Paths;
 import java.util.concurrent.CancellationException;
 import java.util.concurrent.ExecutionException;
+import java.util.concurrent.TimeUnit;
 import java.util.concurrent.atomic.AtomicReference;
 import java.util.regex.Pattern;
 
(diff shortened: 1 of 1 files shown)


testsuite/test-runner/src/main/java/org/wildfly/core/testrunner/Server.java
```diff
@@ -224,11 +207,13 @@ protected void stop() {
                 }
 
                 if (process != null) {
```
bstansberry: I think this can never return false now, so can be removed.

testsuite/test-runner/src/main/java/org/wildfly/core/testrunner/Server.java
```diff
@@ -224,11 +207,13 @@ protected void stop() {
...
-                    process.waitFor();
-                    process = null;
+                    if (process.waitFor(stopTimeout, TimeUnit.SECONDS)) {
+                        process = null;
+                    } else {
+                        // The process hasn't shutdown within the specified timeout. Terminate forcibly.
+                        process.destroy();
```
bstansberry: The old logic amounted to

process.waitFor(timeout); // our own equivalent
if (!stopped) {
  process.destroy();
  process.waitFor();
}

Now it's

process.waitFor(timeout); // our own equivalent
if (!stopped) {
  process.destroy();
}

I don't think the second process.waitFor() should be dropped.

bstansberry: Once the process.waitFor() is restored the process field must be nulled in all cases.

xstefank: @bstansberry Sorry, I don't understand what you mean by the second process.waitFor(). Current implementation starts a new thread that does the timeout loop. Original thread is blocked in process.waitFor(). If the timeout runs out process.destroy() is called which causes process.waitFor() in the original thread to return. Am I missing something?

bstansberry:
My previous comment / your response is hidden a bit now, so I'll reply here.

@xstefank The current code in master, L227 the thread that invoked stop() will not proceed until process.waitFor() returns. It doesn't care whether destroy() was called or not -- it waits until process.waitFor() returns.

The new code assumes that a call to process.destroy() means the process is gone when that call returns. But that in my experience that is not a correct assumption -- a call to destroy() does not mean the process is gone when that call returns. The destroyForcibly() method has explicit javadoc saying that's not the case and I'm pretty sure it was never the case with plain old destroy() either.

xstefank:
@bstansberry Thank you for the explanation! I really appreciate that I learned something new :) hope that now it is alright

PR #2822, opened 2017-09-15
WFCORE-3293 - Transformers logging duplicate messages on DC when managing mixed domain

luck3y:


luck3y:
@bstansberry The reason this initially looked like it was multiple invocations was due to the fact that the first time it was called and generated the error, it was placed into the TransformerLogger messageQueue, but never flushed. Subsequent transformed operations (not resources) would then re-spam this error into the logs (for example reading the domain model after registration etc).

I'm fairly certain this addresses the issue completely.

PR #2823, opened 2017-09-15
[WFCORE-3230] Domain management audit logging: Fix SuffixValidator to allow 's' or 'S' symbols in text quoted using single quotes in suffix.

ivassile:
wildfly-core: https://issues.jboss.org/browse/WFCORE-3230
JBEAP: https://issues.jboss.org/browse/JBEAP-12951


ivassile:
Added unit tests. I first validate the suffix on L54 and then check for 's' in order to prevent checking invalid formats like "m'suffix" (one single quote) which in this case will throw StringIndexOutOfBoundsException.

PR #2826, opened 2017-09-18
[WFCORE-3123] ManagementXml should distinguish different management e…

soul2zimate:
…lements parsing and marshaling based on configuration file type.
https://issues.jboss.org/browse/WFCORE-3123

Just add an enum to skip some management elements parsing and marshaling for domain.xml


bstansberry:
Thanks! Can this just be a boolean instead of a new type? We're on a big memory efficiency drive.

soul2zimate:
sure, it's been rebased.

bstansberry:
Thanks!

PR #2827, opened 2017-09-18
Fix for WFCORE-3225. CLI exception, keep calling stack

jfdenise:
Throw a new CommandlineException to wrap the ExecutionException. Cleanup the exception prior to print the error message.
Fixed unit test that strictly relied on content of the thrown exception.


jfdenise:
I need to update this PR, it is not yet ready for merge.

jfdenise:
PR updated with a more adequate fix.

PR #2829, opened 2017-09-18
[WFCORE-3303] Bump the logging subsystem schema and model versions.

jamezp:
https://issues.jboss.org/browse/WFCORE-3303

### DIFF
--- a/logging/src/main/java/org/jboss/as/logging/LoggingExtension.java
+++ b/logging/src/main/java/org/jboss/as/logging/LoggingExtension.java
@@ -30,7 +30,6 @@ import java.util.Comparator;
 import java.util.List;
 import java.util.Locale;
 import java.util.ResourceBundle;
-import java.util.function.Supplier;
 
 import org.jboss.as.controller.Extension;
 import org.jboss.as.controller.ExtensionContext;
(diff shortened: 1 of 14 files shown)


logging/src/main/java/org/jboss/as/logging/LoggingExtension.java
```diff
@@ -238,6 +238,7 @@ public void initializeParsers(final ExtensionParsingContext context) {
         setParser(context, Namespace.LOGGING_1_5, LoggingSubsystemParser_1_5::new);
         setParser(context, Namespace.LOGGING_2_0, LoggingSubsystemParser_2_0::new);
         setParser(context, Namespace.LOGGING_3_0, LoggingSubsystemParser_3_0::new);
+        setParser(context, Namespace.LOGGING_4_0, LoggingSubsystemParser_4_0::new);
```
bstansberry: These should all be converted back to "new LoggingSubsystemParser_?_0()".

The class gets loaded either way, so using the Supplier variant doesn't save that metaspace. The actual parser instances are stateless in the case of this subsystem, so they have minimal heap cost. OTOH the lambdas end up adding 1360 of class data each.

So getting rid of these method references will probably save about 10K in memory.

BTW, if parsers have a lot of state, e.g. a PersistentResourceXmlDescription used in many subsystems, then I recommend using "new FooParser_3_0()" for the latest one but FooParser_2_0::new for the older ones. There's a boot perf benefit of loading one PersistentResourceXmlDescription in initializeParsers -- see also #2730

PR #2830, opened 2017-09-18
Work around WFCORE-3302 problem by disposing of endpoint along with the stream server

bstansberry:
https://issues.jboss.org/browse/WFCORE-3302


bstansberry:
Doesn't help.

PR #2831, opened 2017-09-18
Work around WFCORE-3302 problem by disposing of endpoint along with th…

bstansberry:
https://issues.jboss.org/browse/WFCORE-3302

master: https://github.com/wildfly/wildfly-core/pull/2830


bstansberry:
Doesn't help.

PR #2833, opened 2017-09-19
Controller Boot Thread should have its TCCL set to the same as the TC…

bstansberry:
…CL of the thread calling AbstractControllerService.start()

MSC before calling start() sets the TCCL to the classloader of the service impl, so this means in production the TCCL will be the CL of the server or host-controller modules.

HOLD: this is just an experiment. I saw code that looks wrong so I'm seeing what the simple fix does.


bstansberry:
Closing; the thread inherits its tccl from the parent.

PR #2836, opened 2017-09-20
[WFCORE-3306] add-user.sh shouldn't throw AddUserFailedException if '%' character is used in user name.

apintea:
https://issues.jboss.org/browse/WFCORE-3306

### DIFF
--- a/domain-management/src/main/java/org/jboss/as/domain/management/security/adduser/PromptNewUserState.java
+++ b/domain-management/src/main/java/org/jboss/as/domain/management/security/adduser/PromptNewUserState.java
@@ -52,6 +52,11 @@ public class PromptNewUserState implements State {
             theConsole.printf(usernamePrompt);
             String temp = theConsole.readLine(" : ");
             if (temp != null && temp.length() > 0) {
+                // If the username contains '%', insert another '%' to avoid the Formatter Exception (WFCORE-3306)
+                if (temp.contains("%")) {
+                    // Insert '%' before every "%", to format the String as the initial typed username
+                    temp = temp.replaceAll("%", "%%");
+                }
                 existingUsername = temp;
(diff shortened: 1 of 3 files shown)


domain-management/src/main/java/org/jboss/as/domain/management/security/adduser/PromptNewUserState.java
```diff
@@ -52,6 +52,11 @@ public State execute() {
             theConsole.printf(usernamePrompt);
             String temp = theConsole.readLine(" : ");
             if (temp != null && temp.length() > 0) {
+                // If the username contains '%', insert another '%' to avoid the Formatter Exception (JBEAP-6637)
```
darranl: This comment should be referencing a WFCORE issue not a JBEAP issue.

darranl:
Can you please squash those two commits into one so all references to JBEAP can be dropped from the history entirely.

apintea:
I have squashed the commits.

"the username with '%' in the name should be verified as being compatible with the Digest authentication mechanism" - I need to write additional tests for the Digests? (DigestMd5MgmtSaslTestCase, DigestSha256MgmtSaslTestCase and DigestSha512MgmtSaslTestCase)

bstansberry:
@darranl ok now?

darranl:
Latest changes look good, can we please at least manually verify Digest authentication accepts these usernames both using the CLI and over HTTP.

bstansberry:
@apintea Please note the request from @darranl above. Someone please comment here when this has been done.

darranl:
@jmesnil I think this is another one we can close, to progress would need some testing that using '%' in the username works correctly - we will likely be revisiting user management fairly soon anyway.

PR #2852, opened 2017-10-02
WFCORE-2882 move elytron subsystem parser to new API

ctomc:
- introduce mix-in parsers (deprecated until we decide we really want to support it)
- introduce decorator child concept (deprecated until we decide we really want to support it)

Move all elytron parsers to definition based parser instead of manual impl.


testsuite/elytron/src/test/java/org/wildfly/test/integration/elytron/sasl/mgmt/ScramPlusMgmtSaslTestCase.java
```diff
@@ -403,6 +403,10 @@ private static KeyStore loadKeyStore(final File ksFile) throws Exception {
                     .addMechanismConfiguration(plusMechsConfig.withMechanismName("SCRAM-SHA-256-PLUS").build())
                     .addMechanismConfiguration(plusMechsConfig.withMechanismName("SCRAM-SHA-384-PLUS").build())
                     .addMechanismConfiguration(plusMechsConfig.withMechanismName("SCRAM-SHA-512-PLUS").build())
+                    .addMechanismConfiguration(plusMechsConfig.withMechanismName("SCRAM-SHA-1").build())
+                    .addMechanismConfiguration(plusMechsConfig.withMechanismName("SCRAM-SHA-256").build())
+                    .addMechanismConfiguration(plusMechsConfig.withMechanismName("SCRAM-SHA-384").build())
+                    .addMechanismConfiguration(plusMechsConfig.withMechanismName("SCRAM-SHA-512").build())
                     .addMechanismConfiguration(MechanismConfiguration.builder().build()).build());
```
fjuma: The last line here (i.e., ``.addMechanismConfiguration(MechanismConfiguration.builder().build())``) actually ends up adding "undefined" to the mechanism configuration list, which likely wasn't intentional. This line should probably just be removed.

elytron/src/main/java/org/wildfly/extension/elytron/MapperParser.java
```diff
@@ -17,1986 +17,216 @@
...
-    private boolean writeSimplePermissionMappers(boolean started, ModelNode subsystem, XMLExtendedStreamWriter writer) throws XMLStreamException {
-        if (subsystem.hasDefined(SIMPLE_PERMISSION_MAPPER)) {
-            startMappers(started, writer);
-            ModelNode permissionMappers = subsystem.require(SIMPLE_PERMISSION_MAPPER);
-            for (String name : permissionMappers.keys()) {
-                ModelNode permissionMapper = permissionMappers.require(name);
-                writer.writeStartElement(SIMPLE_PERMISSION_MAPPER);
-                writer.writeAttribute(NAME, name);
-                PermissionMapperDefinitions.MAPPING_MODE.marshallAsAttribute(permissionMapper, false, writer);
-                if (permissionMap
```
darranl: One small comment on this one, legacy becomes relative after a while - to make it clear wouldn't t be better to say version1Format or something like that?

darranl: Yeah I see this is just to represent the different simple permission mapper, I think in this case we would be better to just take a version is a parameter so the correct version can be returned based on namespace.

ctomc: yeah, naming variables is always the hardest :) will update it.

ctomc: fixed

controller/src/main/java/org/jboss/as/controller/PersistentResourceXMLDescription.java
```diff
@@ -477,6 +565,19 @@ public static PersistentResourceXMLBuilder builder(final PathElement pathElement
         return new PersistentResourceXMLBuilder(pathElement, namespaceURI);
     }
 
+    /**
+     * Creates builder for passed path element
+     *
+     * @param elementName name of xml element that is used as decorator
+     * @return PersistentResourceXMLBuilder
+     * @deprecated Not decided yet if this should become official API!
```
bstansberry: We can't have stuff like this if we are going to be rapidly releasing. It's also not used in the PR. Please remove this method.

ctomc: This one is actually used quite heavily in this PR. 
But looking forward I would love to remove need for it by refactoring xml model of elytron subsystem.
I've updated deprecated comment for time being.
The other deprecated method #addChild(final String xmlElementName, ResourceParser parser, ResourceMarshaller marshaller)
I've commented out for now.

controller/src/main/java/org/jboss/as/controller/PersistentResourceXMLDescription.java
```diff
@@ -242,7 +289,7 @@ private String parseAttributes(final XMLExtendedStreamReader reader, ModelNode o
         return res;
     }
 
-    private void parseChildren(final XMLExtendedStreamReader reader, PathAddress parentAddress, List<ModelNode> list, ModelNode op) throws XMLStreamException {
+    public void parseChildren(final XMLExtendedStreamReader reader, PathAddress parentAddress, List<ModelNode> list, ModelNode op) throws XMLStreamException {
```
bstansberry: Why public?

bstansberry:
Just an FYI, sorting out https://issues.jboss.org/browse/JBEAP-13445 / WFCORE-3041 is going to impact this some. I'll probably have to do a 1.2 xsd (which 2.0 would match, at least for the policyType) and then the 1.0/1.1 parser would use ADs that aren't used except for parsing. I'll try to do this in such a way that it's not to painful here.

controller/src/main/java/org/jboss/as/controller/PersistentResourceXMLDescription.java
ctomc: good catch, it was used in some experiments, but forgot to revert the change. Fixed

ctomc:
@bstansberry without this change, be careful adding new xsd version, as there are no parser tests beyond "latest" xsd version. It will only break later on in some domain / manual tests.

bstansberry:
It should be fairly straightforward to rebase this onto #2887 once I merge it in the morning. The main WFCORE-3041 changes were to PolicyDefinitions (where you just have a 1 line TODO) and to 'PolicyParsers', which you delete. In the last commit in #2887 I revert the changes to  'PolicyParsers' (so you won't get conflicts) and instead add a new temp 'PolicyParsersTemp' class that holds a couple constants with PersistentResourceXmlDescriptions for the two varieties of policy parsing we've supported. Those two PersistentResourceXmlDescriptions can be moved to your new ElytronSubsystemParserX_X classes and 'PolicyParsersTemp'  dropped. A new ElytronSubsystemParser1_2 will be needed.

PR #2867, opened 2017-10-04
[WFCORE-3327] Convert remoting subsystem configuration=endpoint resource into just an alias exposing attributes on its parent

bstansberry:
https://issues.jboss.org/browse/WFCORE-3327

Also includes required fix to 

https://issues.jboss.org/browse/WFCORE-3332

The /subsystem=remoting resource always installs an endpoint. It has a configuration=endpoint child resource that lets the user configure the endpoint, but even if not configured, the endpoint is installed, and with a lot of default values used. If the user doesn't add the configuration=endpoint child, the parent adds it anyway. There are a lot of problems with this:

1) The child resource does not provide any services but is only used to configure services provided by its parent. This is a fairly common pattern, but it's suboptimal.

2) The child resource is kind of a "default value" for its parent, providing config data even if not added by the user, in a similar way to how an attribute can have a default value if the user doesn't define it. But in the attribute case, the resource description data clearly describes this. But there is no such description mechanism for this "default value resource" concept.

3) The endpoint is capability, but the child configures a requirement. This relationship is completely obscured due to this "default value" resource usage. That's a problem for the new provisioning tooling.

This PR corrects this by 

1) Moving all the endpoint config attributes to the parent resource.
2) Adding necessary mixed domain transformers
3) Maintaining compatibility converting the configuration=endpoint child resource into a kind of proxy to the relevant attributes exposed by the parent.

There's a fair bit of side complexity, due to the fact that:

a) the 'worker' attribute is a capability reference, but for mixed domain compatibity reasons it can't register a requirement when in a /profile tree.
b) there are some 'worker-thread' attributes on the parent that only exist for mixed domain support of EAP 6.4.

The handling of these points has been adjusted to better fit with the new code.

### DIFF
--- a/remoting/subsystem/src/main/java/org/jboss/as/remoting/RemotingEndpointResource.java
+++ b/remoting/subsystem/src/main/java/org/jboss/as/remoting/RemotingEndpointResource.java
@@ -21,120 +21,75 @@
 */
 package org.jboss.as.remoting;
 
-import static org.jboss.as.remoting.RemotingSubsystemRootResource.IO_WORKER_CAPABILITY;
-import static org.jboss.as.remoting.RemotingSubsystemRootResource.REMOTING_ENDPOINT_CAPABILITY;
+import static org.jboss.as.controller.descriptions.ModelDescriptionConstants.OP;
+import static org.jboss.as.controller.descriptions.ModelDescriptionConstants.OP_ADDR;
 
-import java.util.Arrays;
-import java.util.Collection;
 import java.util.Collections;
-import java.util.LinkedHashSet;
-import java.util.logging.Level;
-
+import java.util.LinkedHashMap;
+import java.util.Map;
 
 import org.jboss.as.controller.AttributeDefinition;
+import org.jboss.as.controller.ModelVersion;
 import org.jboss.as.controller.OperationContext;
 import org.jboss.as.controller.OperationFailedException;
+import org.jboss.as.controller.OperationStepHandler;
+import org.jboss.as.controller.PathAddress;
 import org.jboss.as.controller.PathElement;
-import org.jboss.as.controller.PersistentResourceDefinition;
-import org.jboss.as.controller.ReloadRequiredRemoveStepHandler;
-import org.jboss.as.controller.ReloadRequiredWriteAttributeHandler;
-import org.jboss.as.controller.SimpleAttributeDefinition;
-import org.jboss.as.controller.SimpleAttributeDefinitionBuilder;
-import org.jboss.as.controller.operations.validation.StringLengthValidator;
-import org.jboss.as.controller.registry.AttributeAccess;
+import org.jboss.as.controller.SimpleResourceDefinition;
 import org.jboss.as.controller.registry.ManagementResourceRegistration;
 import org.jboss.as.controller.registry.OperationEntry;
-import org.jboss.as.controller.registry.Resource;
-import org.jboss.as.remoting.logging.RemotingLogger;
 import org.jboss.dmr.ModelNode;
-import org.jboss.dmr.ModelType;
-import org.jboss.remoting3.RemotingOptions;
-import org.wildfly.extension.io.OptionAttributeDefinition;
 
 /**
  * @author <a href="mailto:tomaz.cerar@redhat.com">Tomaz Cerar</a> (c) 2013 Red Hat Inc.
  */
-public class RemotingEndpointResource extends PersistentResourceDefinition {
-    protected static final SimpleAttributeDefinition WORKER = new SimpleAttributeDefinitionBuilder(CommonAttributes.WORKER, ModelType.STRING)
-            .setRequired(false)
-            .setFlags(AttributeAccess.Flag.RESTART_ALL_SERVICES)
-            .setValidator(new StringLengthValidator(1))
-            .setDefaultValue(new ModelNode("default"))
-            .setCapabilityReference(IO_WORKER_CAPABILITY, REMOTING_ENDPOINT_CAPABILITY)
-            .build();
+class RemotingEndpointResource extends SimpleResourceDefinition {
 
 
-    public static final OptionAttributeDefinition SEND_BUFFER_SIZE = OptionAttributeDefinition.builder("send-buffer-size", RemotingOptions.SEND_BUFFER_SIZE).setDefaultValue(new ModelNode(RemotingOptions.DEFAULT_SEND_BUFFER_SIZE)).setAllowExpression(true).setFlags(AttributeAccess.Flag.RESTART_ALL_SERVICES).build();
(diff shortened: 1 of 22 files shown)


remoting/subsystem/src/main/java/org/jboss/as/remoting/RemotingSubsystemXMLPersister.java
```diff
@@ -133,7 +148,19 @@ private void writeWorkerThreadPoolIfAttributesSet(final XMLExtendedStreamWriter
     }
 
     private void writeEndpointIfAttributesSet(final XMLExtendedStreamWriter writer, final ModelNode model) throws XMLStreamException {
-        RemotingSubsystem30Parser.ENDPOINT_PARSER.persist(writer, model);
+        boolean defined = false;
+        for (String adName : RemotingEndpointResource.ATTRIBUTES.keySet()) {
```
ctomc: Why complicate this much?

Given this is new version of model, lets just introduce new version of XSD as well.
and have clean parser / persister for current version of the subsystem.

bstansberry: Because I don't want to do that now. It means a complete conversion and I don't want us to spend our time on that right now.

bstansberry: Also I don't think the xsd is wrong. The endpoint element is a valid wrapper for a set of elements that are a natural attribute group. I wouldn't want to stick all the attributes on the root xml element. 

The thing I don't want to spend time on now is converting the whole subsystem to PersistentResourceXmlDescription.

remoting/subsystem/src/main/java/org/jboss/as/remoting/RemotingTransformers.java
```diff
@@ -42,6 +51,9 @@
     private static final ModelVersion VERSION_1_3 = ModelVersion.create(1, 3); //eap 6.3 & eap 6.2
     private static final ModelVersion VERSION_1_4 = ModelVersion.create(1, 4); //eap 6.4
     private static final ModelVersion VERSION_3_0 = ModelVersion.create(3, 0);
+    private static final ModelVersion VERSION_4_0 = ModelVersion.create(4, 0); // eap 7.1
+
+    private static AttributeDefinition[] endpointAttrArray = RemotingEndpointResource.ATTRIBUTES.values().toArray(new AttributeDefinition[RemotingEndpointResource.ATTRIBUTES.values().size()]);
```
jamezp: Not a big deal, but this could probably be final.

remoting/subsystem/src/main/resources/org/jboss/as/remoting/LocalDescriptions.properties
```diff
@@ -13,12 +13,33 @@ remoting.worker-task-max-threads=The maximum number of threads for the remoting
 remoting.worker-task-max-threads.deprecated=Use IO subsystem worker configuration
 remoting.worker-write-threads=The number of write threads to create for the remoting worker.
 remoting.worker-write-threads.deprecated=Use IO subsystem worker configuration
+remoting.authorize-id=The SASL authorization ID.  Used as authentication user name to use if no authentication {@code CallbackHandler} is specified\
```
jamezp: The `{@code..}` should probably be removed.

remoting/subsystem/src/main/resources/org/jboss/as/remoting/LocalDescriptions.properties
```diff
@@ -13,12 +13,33 @@ remoting.worker-task-max-threads=The maximum number of threads for the remoting
...
+  and the selected SASL mechanism demands a user name.
+remoting.auth-realm=The authentication realm to use if no authentication {@code CallbackHandler} is specified.
```
jamezp: The `{@code..}` should probably be removed.

remoting/subsystem/src/main/resources/org/jboss/as/remoting/LocalDescriptions.properties
```diff
@@ -13,12 +13,33 @@ remoting.worker-task-max-threads=The maximum number of threads for the remoting
...
 
 remoting.connector=The remoting connectors.
 remoting.configuration=Configuration options
 endpoint=Endpoint configuration
 endpoint.add=Adds endpoint
 endpoint.remove=Removes endpoint configuration
+endpoint.deprecated=The child resource for configuring the remoting endpoint is deprecated. Use the attributes on the parent resource to configure the remoting endpoint.
 endpoint.authorize-id=The SASL authorization ID.  Used as authentication user name to use if no authentication {@code CallbackHandler} is specified\
   and the selected SASL mechanism demands a user name.
 endpoint.auth-realm=The authentication realm to use if no authentication {@code CallbackHandler} is specified.
```
jamezp: Looks like there are other places not associated with this change the use `{@code..}` as well.

remoting/subsystem/src/main/java/org/jboss/as/remoting/RemotingEndpointResource.java
```diff
@@ -21,120 +21,75 @@
...
 
-    public static final java.util.List<OptionAttributeDefinition> OPTIONS = Arrays.asList(SEND_BUFFER_SIZE, RECEIVE_BUFFER_SIZE, BUFFER_REGION_SIZE, TRANSMIT_WINDOW_SIZE, RECEIVE_WINDOW_SIZE,
-            MAX_OUTBOUND_CHANNELS, MAX_INBOUND_CHANNELS, AUTHORIZE_ID, AUTH_REALM, AUTHENTICATION_RETRIES, MAX_OUTBOUND_MESSAGES,
-            MAX_INBOUND_MESSAGES, HEARTBEAT_INTERVAL, MAX_INBOUND_MESSAGE_SIZE, MAX_OUTBOUND_MESSAGE_SIZE, SERVER_NAME, SASL_PROTOCOL);
-
     static {
-        ATTRIBUTES = new LinkedHashSet<AttributeDefinition>(Collections.singletonList(WORKER));
-        ATTRIBUTES.addAll(OPTIONS);
+        Map<String, AttributeDefinition> attrs = new LinkedHashMap<>();
+        assert RemotingSubsystemRootResource.WORKER.getName().equals(RemotingSubsystemRootResource.WORKER.getXmlName());
```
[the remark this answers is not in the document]
bstansberry: No. TBH I don't recall exactly why I was concerned about this when I put this in, but it's a guard that no subsystem dev changes the attribute name to be different from the xml name. It's meant to blow up the testsuite if a dev did that.

bstansberry:
@jamezp Thanks. Somehow I missed your comments for weeks! I corrected all your points; thanks for noticing the {@code CallbackHandler} thing. I fixed them all.

PR #2868, opened 2017-10-05
[WFCORE-3333] Expose number of busy threads in the task worker thread pool

msfm:
https://issues.jboss.org/browse/WFCORE-3333


bstansberry:
Thanks for this but I'm going to close this one as we have no intention of doing further feature releases from the 3.x branch. If we change our minds (unlikely) then we can reconsider. We'd need to figure out what the new management API version to use would be then.

PR #2870, opened 2017-10-05
[WFCORE-3311] Initialize the log manager early for the CLIEmbededServ…

jamezp:
…erTestCase to ensure early initialization doesn't break replacing stdout.

https://issues.jboss.org/browse/WFCORE-3311

In the `org.jboss.logmanager:jboss-logmanager:2.1.0.Alpha4` `System.out` and `System.err` were captured too and used in the `ConsoleHandler`. This caused issues with embedded server logging output even if `--stdout=discard` was set. The test passed because the `StdioContext` was initialized before the log manager. Initializing the log manager before the `StdioContext` allows the log manager to capture `System.out` and `System.err` before it's been replaced.


jamezp:
For completeness here's a CI job run with `org.jboss.logmanager:jboss-logmanager:2.1.0.Alpha4` that shows the test failing https://ci.wildfly.org/viewLog.html?buildId=76413&tab=buildResultsDiv&buildTypeId=WildFlyCore_MasterIgnoreLinux.

bstansberry:
Outstanding!

PR #2872, opened 2017-10-05
[WFCORE-2951] Added a new json-formatter to the logging subsystem.

jamezp:
https://issues.jboss.org/browse/WFCORE-2951

Analysis Document: https://developer.jboss.org/docs/DOC-56153

### DIFF
--- a/logging/src/main/java/org/jboss/as/logging/formatters/StructuredFormatterResourceDefinition.java
+++ b/logging/src/main/java/org/jboss/as/logging/formatters/StructuredFormatterResourceDefinition.java
@@ -1,0 +1,463 @@
+/*
+ * JBoss, Home of Professional Open Source.
+ *
+ * Copyright 2017 Red Hat, Inc., and individual contributors
+ * as indicated by the @author tags.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.jboss.as.logging.formatters;
+
+import static org.jboss.as.logging.Logging.createOperationFailure;
+
+import java.util.Iterator;
+import java.util.List;
+
+import org.jboss.as.controller.AttributeDefinition;
+import org.jboss.as.controller.AttributeMarshallers;
+import org.jboss.as.controller.ObjectTypeAttributeDefinition;
+import org.jboss.as.controller.OperationContext;
+import org.jboss.as.controller.OperationFailedException;
+import org.jboss.as.controller.OperationStepHandler;
+import org.jboss.as.controller.PathAddress;
+import org.jboss.as.controller.PathElement;
+import org.jboss.as.controller.SimpleAttributeDefinition;
+import org.jboss.as.controller.SimpleAttributeDefinitionBuilder;
+import org.jboss.as.controller.SimpleMapAttributeDefinition;
+import org.jboss.as.controller.operations.validation.StringAllowedValuesValidator;
+import org.jboss.as.controller.registry.AttributeAccess;
+import org.jboss.as.controller.registry.ManagementResourceRegistration;
+import org.jboss.as.controller.transform.description.ResourceTransformationDescriptionBuilder;
+import org.jboss.as.logging.ElementAttributeMarshaller;
+import org.jboss.as.logging.KnownModelVersion;
+import org.jboss.as.logging.LoggingExtension;
+import org.jboss.as.logging.LoggingOperations;
+import org.jboss.as.logging.PropertyAttributeDefinition;
+import org.jboss.as.logging.TransformerResourceDefinition;
+import org.jboss.as.logging.logging.LoggingLogger;
+import org.jboss.as.logging.resolvers.ModelNodeResolver;
+import org.jboss.dmr.ModelNode;
+import org.jboss.dmr.ModelType;
+import org.jboss.dmr.Property;
+import org.jboss.logmanager.PropertyValues;
+import org.jboss.logmanager.config.FormatterConfiguration;
+import org.jboss.logmanager.config.LogContextConfiguration;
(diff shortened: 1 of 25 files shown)


logging/src/main/java/org/jboss/as/logging/formatters/StructuredFormatterResourceDefinition.java
```diff
@@ -0,0 +1,462 @@
...
+                }
+            }
+            return false;
+        }
+    };
+
+    /**
+     * A step handler to remove
+     */
+    private static final OperationStepHandler REMOVE = new LoggingOperations.LoggingRemoveOperationStepHandler() {
```
jmesnil: Why isn't LoggingOperations.LoggingRemoveOperationStepHandler inheriting from AbstractRemoveStepHandler so that you get most of the clean up done by the super class?

It was my understanding that AbstractRemoveStepHandler and AbstractAddStepHandler where the "correct" base classes to define :add and :remove attributes and make sure that the operations were behaving as expected (wrt to cleaning up resources, rolling back ops, etc.)

logging/src/main/java/org/jboss/as/logging/formatters/StructuredFormatterResourceDefinition.java
```diff
@@ -0,0 +1,462 @@
...
+        }
+
+        @Override
+        public void updateModel(final ModelNode operation, final ModelNode model) throws OperationFailedException {
+            for (AttributeDefinition attribute : ATTRIBUTES) {
+                attribute.validateAndSet(operation, model);
+            }
+        }
+
+        private static boolean changeProperty(final FormatterConfiguration configuration, final String name, final String value) {
```
jmesnil: I'm not sure about this method. Given its name, I'd assume it is changing the configuration but it's only checking if the current value is the same. 
Shouldn't it be renamed to something like `isSamePropertyValue()`?

logging/src/main/resources/schema/jboss-as-logging_4_0.xsd
```diff
@@ -565,6 +639,80 @@
...
+                    <xs:enumeration value="formatted">
+                        <xs:annotation>
+                            <xs:documentation>
+                                The cause, if present, will be a string representation of the stack trace in a
+                                stackTrace property. The property value is a string created by
+                                Throwable.printStackTrace().
+                            </xs:documentation>
+                        </xs:annotation>
+                    </xs:enumeration>
+                    <xs:enumeration value="detailed-and-formatter">
```
jmesnil: typo: this should be `detailed-and-formatted`

logging/src/test/resources/logging.xml
```diff
@@ -151,6 +151,17 @@
         <pattern-formatter pattern="%d{HH:mm:ss,SSS} %-5p [%c] (%t) %s%e%n" color-map="info:cyan,warn:yellow,error:red"/>
     </formatter>
 
+    <formatter name="JSON">
+        <json-formatter date-format="yyyy-MM-dd'T'HH:mm:ssSSS" pretty-print="true" print-details="true" zone-id="GMT">
+            <exception-output-type value="detailed-and-formatter"/>
```
jmesnil: typo: should be `detailed-and-formatted`

jmesnil: why is the test passing? The EXCEPTION_OUTPUT_TYPE attribute uses `detailed-and-formatted` in its allowed values, it should have rejected the `detailed-and-formatter` value

logging/src/main/java/org/jboss/as/logging/formatters/StructuredFormatterResourceDefinition.java
```diff
@@ -0,0 +1,462 @@
...
+public abstract class StructuredFormatterResourceDefinition extends TransformerResourceDefinition {
+
+    public static final PropertyAttributeDefinition DATE_FORMAT = PropertyAttributeDefinition.Builder.of("date-format", ModelType.STRING, true)
+            .setAllowExpression(true)
+            .setPropertyName("dateFormat")
+            .build();
+
+    public static final PropertyAttributeDefinition EXCEPTION_OUTPUT_TYPE = PropertyAttributeDefinition.Builder.of("exception-output-type", ModelType.STRING, true)
+            .setAllowExpression(true)
+            .setAllowedValues("detailed", "formatted", "detailed-and-formatted")
```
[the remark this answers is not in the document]
jamezp: We definitely could. It just thought using lower case with dashes vs `DETAILED`, `FORMATTED` and `DETAILED_AND_FORMATTED` fit the model better. I've got no real strong opinion either way though.

logging/src/main/java/org/jboss/as/logging/formatters/StructuredFormatterResourceDefinition.java
jamezp: At some point this is something I would like to fix. However currently logging has some special cases that need to be handled. The main one is when a change is made the `logging.properties` file needs to be rewritten. Hopefully soon we can remove the `logging.properties` which will trigger a rewrite of a lot of these :)

logging/src/test/resources/logging.xml
jamezp: That's a good question. I can see the error being thrown because of the typo, but the test controller seems to be booting fine. I'll have to dig into this.

jamezp: It looks like this produced a runtime failure and the test controller doesn't boot with `rollbackOnRuntimeFailure` set to `true`. So the error was hidden.

logging/src/main/java/org/jboss/as/logging/formatters/StructuredFormatterResourceDefinition.java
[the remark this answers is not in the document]
jmesnil: sorry, I should have been more explicit.
Having lower case with dashes is the correct way to show these values.
I was thinkg of creating a 2nd (internal) enum that would match the ExceptionOutputType enum with these strings. This way, the attribute definition can delegate its validator and model resolve to this 2nd enum

jmesnil:
retest this please

logging/src/main/java/org/jboss/as/logging/formatters/StructuredFormatterResourceDefinition.java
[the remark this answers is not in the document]
jamezp: Ah got it. I found the `StringAllowedValuesValidator` and just used that as it seems to not require the allowed values to be set and works like the enum validator.

jamezp:
One more time, retest this please

jamezp:
retest this please

jamezp:
I'm closing this as there is a downgrade for the log manager to be final in #3069

PR #2843, opened 2017-09-25
Fix for WFCORE-2088, aesh 1.0 and patching rewrite

jfdenise:
Integration of aesh 1.0 CommandRuntime. Rewrote patch command to allow for patch ID completion and rollback activation only if some patches can be rollback.

PR #2878, opened 2017-10-09
WFCORE-3223, CLI, unified deployment command

jfdenise:
Introduce a new command "deployment" with a set of 11 actions to replace deploy/undeploy and deployment-info. These 3 commands are still active but hidden from completion.


jamezp:
Should #2843 and #2851 be closed as it seems to be included here?

jfdenise:
@jamezp , it is time for cleanup, I will close this one and the older ones.

PR #2880, opened 2017-10-10
WFCORE-3352 Use thin provisioned server everywhere in testsuite

ctomc:


bstansberry:
@ctomc Please send a mail to wildfly-dev about this change. It needs to be publicized and discussed before it can go in so people have a chance to comment on how it might impact their workflows.

For me it means this very frequently used thing no longer works:

$ dist/target/wildfly-core-4.0.0.Alpha1-SNAPSHOT/bin/standalone.sh 

Shifting from dist to build works, but this is the kind of thing that needs to be discussed.

pom.xml
```diff
@@ -95,9 +95,10 @@
         <version.org.apache.ds>2.0.0-M15</version.org.apache.ds> <!-- TODO Elytron - Bump to M17 -->
         <version.org.apache.httpcomponents.httpclient>4.5.2</version.org.apache.httpcomponents.httpclient>
         <version.org.apache.httpcomponents.httpcore>4.4.4</version.org.apache.httpcomponents.httpcore>
-        <version.org.apache.maven>3.3.9</version.org.apache.maven>
+        <version.org.apache.maven>3.5.0</version.org.apache.maven>
```
jamezp: Have we tested with maven 3.3.1 as that is the current minimum requirement? I just want to make sure we don't have any weird clashes.

dist/pom.xml
```diff
@@ -55,64 +55,56 @@
             </exclusions>
```
jamezp: Just a minor note. I noticed the `dist` module is `<packaging>jar</packaging>`. If we change it to `pom` it would stop the JAR from being created.

testsuite/shared/src/main/java/org/jboss/as/test/integration/management/cli/CliProcessWrapper.java
```diff
@@ -276,17 +276,21 @@ private boolean waitForClose() throws IOException {
     }
 
     private boolean outputHasPrompt(String prompt){
-        if(cliOutputBuffer.toString().contains("\n")) {
+        String buffer = cliOutputBuffer.toString();
+        if (buffer.startsWith("Exception")){
```
jamezp: Is there a reason for this change?

pom.xml
psakar: @ctomc Can you please provide more info why this change is needed ?

ctomc: @jamezp @psakar this version of maven here has nothing to do with requirements for building with maven 3.5 it is only for dependency management of [maven-resolver](https://maven.apache.org/resolver/)  (ex eclipse aether) which we use in subsystem legacy test framework to be able to load legacy version of extensions when testing transformers.

testsuite/shared/src/main/java/org/jboss/as/test/integration/management/cli/CliProcessWrapper.java
ctomc: yeah, it is here to aid with exception handling. 
With this change you will get proper exception reported in test result if exception occurs inside CLI execution.
Without this, it is swallowing the originating exception that happens inside CLI command.

dist/pom.xml
ctomc: good point.

pom.xml
jamezp: Right yeah. I was more worried about ending up with a mismatch on the class path for tests is all. Those tests use some custom class loading, plus surefire itself has it's own class loading issues at times. I have a feeling it will work, but I just think we should test it.

ctomc: It does work, it is tested, if testsuite passes that is a huge test.

CI uses maven 3.3.9, locally I am using maven 3.5 and it works.

And yeah, this stuff ends up in custom classloader completely.

testsuite/shared/src/main/java/org/jboss/as/test/integration/management/cli/CliProcessWrapper.java
bstansberry: @jfdenise I'm going to merge this PR, but if you have any objection to this change, please let Tomaz know.

bstansberry:
This broke the build once I bumped the poms to 4.0.0.Alpha2-SNAPSHOT. :(

[ERROR] Failed to execute goal on project wildfly-core-testsuite-standalone: Could not resolve dependencies for project org.wildfly.core:wildfly-core-testsuite-standalone:jar:4.0.0.Alpha2-SNAPSHOT: Failure to find org.wildfly.core:wildfly-core-build:zip:4.0.0.Alpha2-SNAPSHOT in https://repository.jboss.org/nexus/content/groups/public-jboss/ was cached in the local repository, resolution will not be reattempted until the update interval of jboss-public-repository-group has elapsed or updates are forced -> [Help 1]

I guess it didn't fail before because local maven repos had a copy of a previous 4.0.0.Alpha1-SNAPSHOT zip.

bstansberry:
I'm reverting this as I need to do a 4.0.0.Alpha2 now since Alpha1 might be broken, and I don't want spend my Sunday sorting this.

PR #2882, opened 2017-10-11
[WFCORE-3041] Correct handling of the elytron subsystem policy resource

bstansberry:
https://issues.jboss.org/browse/WFCORE-3041

There is no version of this for master yet; that's a task for tomorrow.

Eliminates the ability to specify multiple java.security.Policy configurations in the policy=* resource, with a 'default-policy' attribute, or the resource name if default-policy is undefined, indicating which policy is effective. This is bug prone and unintuitive, and in the end only one policy takes effect, so we should just follow the KISS principle and require configuration of just one policy -- either a jacc-policy or a custom-policy.

So, the jacc-policy and custom-policy attributes are no longer lists; rather they are each a single (complex) item, and are alternatives to each other.

The default-policy management attribute is now ignored. The user must pick between jacc-policy and custom-policy.  (In the 1.0 and 1.1 xsd version parsers the default-policy xml attribute is still used, as it provides the name of the resource. Those xsds have been corrected to reflect the reality that the parser has required and still requires that attribute.)

The jacc-policy and custom-policy complex attributes no longer have a 'name' field as it's not needed now that we're no longer trying to match possibly multiple policies to 'default-policy'.

The policy=* resource is an effective singleton and as such could have a static name, but I don't change that. That would be too disruptive at this stage of WF 11 work, and allowing the user to set the name isn't harmful.

I introduce a new 1.2 xsd which changes the xml representation to match the management model.

a) default-policy attribute is replaced by 'name' as that's the only remaining function of the attribute.
b) either 1 jacc-policy or 1 custom-policy is allowed, rather than 0..n of each.
c) the 'name' attribute in jacc-policy and custom-policy is dropped.

This subsystem has never been used in a WildFly .Final release, so API compatibility is not a critical concern. However, since it has been in a WildFly Core .Final I made some efforts to prevent breaking people who may have learned the old API.

1) If multiple policies are configured the 1.0 and 1.1 parsers look for the one that matches the 'default-policy' attribute and only include that in the management op. Other policies are discarded, but a WARN is logged.

2) The 'default-policy' management attribute is retained but any value for it in an op is discarded. It's deprecated, so an INFO is logged if an op includes it. The parsers don't include it.

3) The jacc-policy and custom-policy attributes definitions attempt to correct a LIST provided by the user (since LIST used to be the correct attribute type.) If the LIST has a single element, that element is extracted and is used. Otherwise, the attribute will be rejected. This rejection is the main incompatibility, but I think users configuring multiple policies would be unlikely, since doing so would have served no purpose.

4) A mixed domain transformer was added. Just because. ;)

I haven't tried HAL with this, but an important effect of 1-3 above is they should allow HAL to work unchanged. I haven't looked at how HAL treats this resource, and it's possible HALwill allow users to attempt to configure things in a way that will fail (e.g. multiple policies) but if they do normal things it should work.

### DIFF
--- a/elytron/pom.xml
+++ b/elytron/pom.xml
@@ -202,7 +202,11 @@
                             <includes>
                                 <include>*.xml</include>
                             </includes>
-                            <systemId>src/main/resources/schema/wildfly-elytron_1_1.xsd</systemId>
+                            <excludes>
+                                <exclude>custom-policies.xml</exclude>
+                                <exclude>jacc-with-providers.xml</exclude>
+                            </excludes>
+                            <systemId>src/main/resources/schema/wildfly-elytron_1_2.xsd</systemId>
                         </validationSet>
                         <validationSet>
                             <dir>src/main/resources/subsystem-templates</dir>
--- a/elytron/src/main/java/org/wildfly/extension/elytron/ElytronExtension.java
+++ b/elytron/src/main/java/org/wildfly/extension/elytron/ElytronExtension.java
@@ -58,7 +58,8 @@ public class ElytronExtension implements Extension {
      */
     public static final String NAMESPACE_1_0 = "urn:wildfly:elytron:1.0";
     public static final String NAMESPACE_1_1 = "urn:wildfly:elytron:1.1";
-    public static final String CURRENT_NAMESPACE = NAMESPACE_1_1;
+    public static final String NAMESPACE_1_2 = "urn:wildfly:elytron:1.2";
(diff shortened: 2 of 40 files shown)


hpehl:
HAL (and HAL.next) breaks with the PR applied. We implemented a custom behaviour to work around [WFCORE-3041](https://issues.jboss.org/browse/WFCORE-3041). 

With this PR in place we need to undo our workaround.

PR #2885, opened 2017-10-12
[WFCORE-3355] Check for a failed file visit and if the file is readab…

jamezp:
…le throw an exception, if the file is not readable just allow the process to continue.

https://issues.jboss.org/browse/WFCORE-3355


jamezp:
retest this please

PR #2886, opened 2017-10-13
[WFCORE-3041] Correct handling of the elytron subsystem policy resource

bstansberry:
https://issues.jboss.org/browse/WFCORE-3041

Forward ported variant of #2882

PR #2887, opened 2017-10-13
WFCORE-3041 helper for WFCORE-2882

bstansberry:
Goes beyond #2886 to add a commit dropping the custom policy element parsing we used in WildFly Core 3.0.6 for the 1.1 and 1.0 xsds and instead use PersistentResourceXmlDescription.
    
This is meant to make it easier to merge the WFCORE-2882 work (see #2852 ) on top of the WFCORE-3041 work. To this end the policy parsing logic has been moved to a temporary new class 'PolicyParserTemp' and the legacy 'PolicyParser' class which is no longer used, has been restored to its state prior to any WFCORE-3041 changes. (This avoid difficult to untangle merge conflicts with the WFCORE-2882 work.) The PolicyParserTemp temp class shows the proper configuration of the PersistentResourceXmlDescriptions that are needed to parse the policy resource, both for 1.0/1.1 xsds and for 1.2/2.0 xsds. WFCORE-2882 should use these descriptions.
    
Moving to PersistentResourceXmlDescription loses some custom 'forgiveness' logic we had in 3.0.6 as a compatibility aid. Forgiveness meaning the parser discarding extra jacc-policy or custom-policy elements that weren't used in the runtime anyway. I think this is ok; any need for such forgiveness was always an extreme edge case (since adding unused config would have been pointless), and by the time WildFly Core 4 is released any such xml should long since have been converted. The 1.1 and 1.0 xsds were never the base version used in a final WildFly or EAP release.

### DIFF
--- a/elytron/pom.xml
+++ b/elytron/pom.xml
@@ -203,6 +203,7 @@
                                 <include>*.xml</include>
                             </includes>
                             <excludes>
+                                <exclude>*-1.*.xml</exclude>
                                 <exclude>custom-policies.xml</exclude>
                                 <exclude>jacc-with-providers.xml</exclude>
                             </excludes>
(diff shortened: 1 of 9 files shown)


bstansberry:
retest this please

PR #2888, opened 2017-10-13
[WFCORE-3338] Add a readResource call on StoppedServerResource attrib…

soul2zimate:
…utes to check server existence

https://issues.jboss.org/browse/WFCORE-3338
Read resource to check domain server existence and to throw NoSuchResourceException for non-existent server from OSH. Since it reads server-config=xx, stopped server (server-three) can still return correct result as long as it presents in configuration file.

### DIFF
--- a/host-controller/src/main/java/org/jboss/as/host/controller/resources/StoppedServerResource.java
+++ b/host-controller/src/main/java/org/jboss/as/host/controller/resources/StoppedServerResource.java
@@ -22,7 +22,12 @@
 
 package org.jboss.as.host.controller.resources;
 
+import static org.jboss.as.controller.descriptions.ModelDescriptionConstants.HOST;
+import static org.jboss.as.controller.descriptions.ModelDescriptionConstants.SERVER_CONFIG;
+
 import org.jboss.as.controller.CompositeOperationHandler;
+import org.jboss.as.controller.OperationContext;
+import org.jboss.as.controller.PathAddress;
 import org.jboss.as.controller.PathElement;
 import org.jboss.as.controller.SimpleResourceDefinition;
 import org.jboss.as.controller.client.helpers.ClientConstants;
@@ -31,7 +36,7 @@ import org.jboss.as.controller.registry.ManagementResourceRegistration;
 import org.jboss.as.host.controller.descriptions.HostResolver;
 import org.jboss.as.server.ServerEnvironment;
 import org.jboss.as.server.controller.resources.ServerRootResourceDefinition;
-import org.jboss.as.server.operations.LaunchTypeHandler;
+import org.jboss.dmr.ModelNode;
 
 /**
  * {@code ResourceDescription} describing a stopped server instance.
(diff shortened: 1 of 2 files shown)


host-controller/src/main/java/org/jboss/as/host/controller/resources/StoppedServerResource.java
```diff
@@ -51,11 +59,30 @@ public void registerAttributes(ManagementResourceRegistration resourceRegistrati
...
+                (context, operation) -> {
+                    readResourceServerConfig(context, operation);
+                    context.getResult().set(ClientConstants.CONTROLLER_PROCESS_STATE_STOPPED);
+                    }
+                );
+    }
+
+    // https://issues.jboss.org/browse/WFCORE-3338 read server-config to test name of the server in operation is present
+    private void readResourceServerConfig(OperationContext context, ModelNode operation) {
+        final PathAddress address = PathAddress.pathAddress(operation.require(OP_ADDR));
```
bstansberry: This should use context.getCurrentAddress().

That's a minor thing, but I want to eliminate handlers parsing of the address from the op from the code base.

host-controller/src/main/java/org/jboss/as/host/controller/resources/StoppedServerResource.java
```diff
@@ -67,4 +94,19 @@ public void registerOperations(final ManagementResourceRegistration resourceRegi
         resourceRegistration.registerOperationHandler(CompositeOperationHandler.INTERNAL_DEFINITION, CompositeOperationHandler.INSTANCE);
     }
 
+    class LaunchTypeHandler implements OperationStepHandler {
```
bstansberry: This can be private and static. Static removes a ref to StoppedServerResource.this from memory.

host-controller/src/main/java/org/jboss/as/host/controller/resources/StoppedServerResource.java
```diff
@@ -67,4 +94,19 @@ public void registerOperations(final ManagementResourceRegistration resourceRegi
...
+        // this is for operation in domain model and embed-host-controller mode
+        // org.jboss.as.server.operations.LaunchTypeHandler takes care of standalone mode and embed-server mode.
+        private final ServerEnvironment.LaunchType launchType;
```
bstansberry: This field is not needed as it's always the same.

host-controller/src/main/java/org/jboss/as/host/controller/resources/StoppedServerResource.java
soul2zimate: fixed.

host-controller/src/main/java/org/jboss/as/host/controller/resources/StoppedServerResource.java
soul2zimate: Thanks, since this does not need to be parameterized launch type, I removed class LaunchTypeHandler, and put it into lambda as other OSH.

soul2zimate:
This is rebased as per requested changes. Thanks.

PR #2890, opened 2017-10-16
[JBEAP-13502] : Not able to separate application(EAR) logging with the use of logging profile

panossot:
[JBEAP-13502] : https://issues.jboss.org/browse/JBEAP-13502
Upstream : https://github.com/wildfly/wildfly-core/pull/2881


bstansberry:
We don't intend to maintain the 3.x branch beyond what's needed for WildFly 11.0.0.Final, and this issue is not something we are tracking for a fix there. So I'm closing this. The 3.x branch is not an upstream of any fixes meant for EAP 7.1 CP releases.

Also upstream commit messages should not only reference JBEAP JIRAs. They should reference the relevant WFCORE.

PR #2891, opened 2017-10-17
[WFCORE-3361] GSSCredential obtaining fail caching for fast fallback

honza889:
Depends on https://github.com/wildfly-security/wildfly-elytron/pull/1013

https://issues.jboss.org/browse/WFCORE-3361
https://issues.jboss.org/browse/JBEAP-11055


bstansberry:
I'm dropping the "rebase this" label since making requested changes is going to mean the rebase anyway.

honza889:
@darranl which transformer do you mean in comment above?

darranl:
As you have added a new attribute to version 2 of the management model of this subsystem a transformer is also going to need to be added to reject this attribute if this is used on a later domain controller managing earlier domain controllers as the earlier ones will not understand the attribute.

kabir:
@honza889 https://docs.jboss.org/author/display/WFLY/Domain+Mode+Subsystem+Transformers

honza889:
Should be ok now.

PR #2895, opened 2017-10-19
[WFCORE-3367] Elytron doc: note that provider attributes affects AS globally

honza889:
Note: change of XSD is meaningless, documentation-improvement-only
https://issues.jboss.org/browse/WFCORE-3367
https://issues.jboss.org/browse/JBEAP-11256

### DIFF
--- a/elytron/src/main/resources/org/wildfly/extension/elytron/LocalDescriptions.properties
+++ b/elytron/src/main/resources/org/wildfly/extension/elytron/LocalDescriptions.properties
@@ -4,9 +4,9 @@ elytron.add=Operation adds the Elytron subsystem
 elytron.remove=Operation removes the Elytron subsystem
 # Attributes
 elytron.default-authentication-context=The default authentication context to be associated with all deployments.
-elytron.initial-providers=Reference to the Providers that should be registered ahead of all existing Providers.
-elytron.final-providers=Reference to the Providers that should be registered after all existing Providers.
-elytron.disallowed-providers=A list of providers that are not allowed, and will be removed from the providers list.
+elytron.initial-providers=Reference to the Providers that should be registered ahead of all existing Providers. When set, affects a whole server globally, regardless it is referenced from other subsystems.
+elytron.final-providers=Reference to the Providers that should be registered after all existing Providers. When set, affects a whole server globally, regardless it is referenced from other subsystems.
+elytron.disallowed-providers=A list of providers that are not allowed, and will be removed from the providers list. When set, affects a whole server globally, regardless it is referenced from other subsystems.
 
 #######################
 # Security Properties #
--- a/elytron/src/main/resources/schema/wildfly-elytron_1_0.xsd
+++ b/elytron/src/main/resources/schema/wildfly-elytron_1_0.xsd
@@ -56,6 +56,7 @@
             <xs:annotation>
                 <xs:documentation>
                     Reference to a capability providing a Provider[] which will be registered globally ahead of all existing Provider registrations.
+                    When set, affects a whole server globally, regardless it is referenced from other subsystems.
                 </xs:documentation>
             </xs:annotation>
         </xs:attribute>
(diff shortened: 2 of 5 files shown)


honza889:
Changed to "server" as it is really more used.

elytron/src/main/resources/org/wildfly/extension/elytron/LocalDescriptions.properties
```diff
@@ -4,9 +4,9 @@ elytron.add=Operation adds the Elytron subsystem
 elytron.remove=Operation removes the Elytron subsystem
 # Attributes
 elytron.default-authentication-context=The default authentication context to be associated with all deployments.
-elytron.initial-providers=Reference to the Providers that should be registered ahead of all existing Providers.
-elytron.final-providers=Reference to the Providers that should be registered after all existing Providers.
-elytron.disallowed-providers=A list of providers that are not allowed, and will be removed from the providers list.
+elytron.initial-providers=Reference to the Providers that should be registered ahead of all existing Providers. When set, affects a whole server globally, regardless it is referenced from other subsystems.
```
jamezp: Sorry I missed this earlier, but I wonder if it should read something more like:
```
When set this globally overrides all other $name.
```
For example `When set this globally overrides all other initial-providers.`

@darranl WDYT? I don't really know much about these settings.

honza889: This is not about overriding any other configuration options (in standalone.xml meaning) - this mean setting of this affect whole server regardless elytron subsystem is "enabled" by referencing http-authentication-factory from undertow for example...

honza889: As opposite to elytron resources like http-authentication-factory is, which are ignored when they are not referenced from other subsystem (udertow, ejb3...), this configuration option affect server immediately.

jamezp: Hmm... ...okay. What is meant by the elytron subsystem "enabled"? Is there a way for the subsystem to exist, but not have it enabled?

honza889: When wildfly documentation talks about elytron "enabling", it mean we start to using it - by setting references to elytron resources in undertow subsystem, ejb subsystem and so on. The elytron subsystem is present for the whole time, but when its resources are not referenced, it is not used by wildfly, so we usually say it is disabled. There is only one exception - attributes mentioned in this PR - which is reason for this warning.

darranl:
The enabled / not enabled discussion is a bad practice we have got into.

WildFly Elytron is always enabled and always used within the application server.  

The only thing that changes is if the Elytron subsystem is present and if the resources defined within it are actually used.

jamezp:
@darranl WRT your comment, do you think this clarification would even be needed then? 

When I change an attribute in a subsystem, or really anywhere, I expect that to make a behavioral change. IMO the additional wording shouldn't be required and almost makes it a bit more confusing.

darranl:
I think in this case it does make sense, as WildFly Elytron is neither enabled or disabled the configuration here will always have an impact.

jamezp:
Okay. I guess to me I would expect the behavior since they are on the root subsystem. A good example is the `deployment-scanner` subsystem. All attributes are essentially global. Or the `request-controller` subsystem.

The main thing I'm concerned about is creating confusion around which attributes are global and which attributes are not.

That said I think we need to remove the "subsystem" reference. I don't really have a better wording for it, but maybe something like `When defined this globally affects the server`.

darranl:
It probably is sufficient to remove the mention of other subsystems although I think the current proposed change is also fine.

bstansberry:
If you decide to go forward with this, please change "regardless it..." to "regardless of whether it...".

PR #2900, opened 2017-10-20
[WFCORE-3347] add ability to remove alternatives and remove unnecessary ones in DeploymentAttributes.

soul2zimate:
please see issue https://issues.jboss.org/browse/WFCORE-3347

### DIFF
--- a/controller/src/main/java/org/jboss/as/controller/AbstractAttributeDefinitionBuilder.java
+++ b/controller/src/main/java/org/jboss/as/controller/AbstractAttributeDefinitionBuilder.java
@@ -334,6 +334,51 @@ public abstract class AbstractAttributeDefinitionBuilder<BUILDER extends Abstrac
     }
 
     /**
+     * Removes {@link AttributeDefinition#getAlternatives() names of alternative attributes} from the set of those that should not
+     * be defined if this attribute is defined.
+     * @param alternatives the attribute names
+     * @return a builder that can be used to continue building the attribute definition
+     */
+    public final BUILDER removeAlternatives(String... alternatives) {
+        if (this.alternatives == null) {
+            return (BUILDER) this;
+        } else {
+            for (String alternative : alternatives) {
+                if (isAlternativePresent(alternative)) {
+                    int length = this.alternatives.length;
+                    String[] newAlternatives = new String[length - 1];
+                    int k = 0;
+                    for (String alt : this.alternatives) {
+                        if (alt != alternative) {
+                            newAlternatives[k] = alt;
+                            k++;
+                        }
+                    }
+                    this.alternatives = newAlternatives;
+                }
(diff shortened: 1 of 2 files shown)


controller/src/main/java/org/jboss/as/controller/AbstractAttributeDefinitionBuilder.java
```diff
@@ -334,6 +334,51 @@ public BUILDER addAlternatives(String... alternatives) {
...
+        }
+        return (BUILDER) this;
+    }
+
+    /**
+     * Checks if an alternative has been recorded in {@link AttributeDefinition#getAlternatives() names of alternative attributes}
+     * @param alternative the alternative
+     * @return a builder that can be used to continue building the attribute definition
+     */
+    protected boolean isAlternativePresent(final String alternative) {
```
bstansberry: This should be private.  The similar isFlagPresent method should have been private too, but since it was released as protected I don't want to reduce the visibility. But this one is new so let's lock it down.

controller/src/main/java/org/jboss/as/controller/AbstractAttributeDefinitionBuilder.java
```diff
@@ -334,6 +334,51 @@ public BUILDER addAlternatives(String... alternatives) {
     }
 
     /**
+     * Removes {@link AttributeDefinition#getAlternatives() names of alternative attributes} from the set of those that should not
+     * be defined if this attribute is defined.
+     * @param alternatives the attribute names
+     * @return a builder that can be used to continue building the attribute definition
+     */
+    public BUILDER removeAlternatives(String... alternatives) {
```
bstansberry: Please make this final. A lot of methods in this class should have been final but aren't, but since this is a new one let's lock it down.

soul2zimate:
my bad, I copied from existing ones.
I have updated these two modifiers as review comments above. thanks.

soul2zimate:
retest this please

controller/src/main/java/org/jboss/as/controller/AbstractAttributeDefinitionBuilder.java
```diff
@@ -334,6 +334,51 @@ public BUILDER addAlternatives(String... alternatives) {
...
+        if (this.alternatives == null) {
+            return (BUILDER) this;
+        } else {
+            for (String alternative : alternatives) {
+                if (isAlternativePresent(alternative)) {
+                    int length = this.alternatives.length;
+                    String[] newAlternatives = new String[length - 1];
+                    int k = 0;
+                    for (String alt : this.alternatives) {
+                        if (alt != alternative) {
```
rsvoboda: Shouldn't be `! .equals()` instead of `!=` ?

PR #2901, opened 2017-10-20
[WFCORE-3116] Add support for digest-sha-384 for Elytron set-password operation

yersan:
This issue depends on https://github.com/wildfly-security/wildfly-elytron/pull/1021 and has to be merged with that one. 
Tests will fail if it is merged with the incorrect Elytron version.

Issue: https://issues.jboss.org/browse/WFCORE-3116
JBEAP issue: https://issues.jboss.org/browse/JBEAP-12316

### DIFF
--- a/elytron/src/main/java/org/wildfly/extension/elytron/ModifiableRealmDecorator.java
+++ b/elytron/src/main/java/org/wildfly/extension/elytron/ModifiableRealmDecorator.java
@@ -418,6 +418,7 @@ class ModifiableRealmDecorator extends DelegatingResourceDefinition {
                             DigestPassword.ALGORITHM_DIGEST_MD5,
                             DigestPassword.ALGORITHM_DIGEST_SHA,
                             DigestPassword.ALGORITHM_DIGEST_SHA_256,
+                            DigestPassword.ALGORITHM_DIGEST_SHA_384,
                             DigestPassword.ALGORITHM_DIGEST_SHA_512
                     ))
                     .build();
--- a/testsuite/elytron/src/test/java/org/wildfly/test/integration/elytron/sasl/mgmt/DigestSha384MgmtSaslTestCase.java
+++ b/testsuite/elytron/src/test/java/org/wildfly/test/integration/elytron/sasl/mgmt/DigestSha384MgmtSaslTestCase.java
@@ -1,0 +1,69 @@
+/*
+ * Copyright 2017 Red Hat, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *   http://www.apache.org/licenses/LICENSE-2.0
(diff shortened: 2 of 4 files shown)


testsuite/elytron/src/test/java/org/wildfly/test/integration/elytron/sasl/mgmt/DigestSha384MgmtSaslTestCase.java
```diff
@@ -0,0 +1,53 @@
+package org.wildfly.test.integration.elytron.sasl.mgmt;
```
jamezp: This needs a copyright header.

elytron/src/main/java/org/wildfly/extension/elytron/ModifiableRealmDecorator.java
```diff
@@ -418,6 +418,7 @@ protected void executeRuntimeStep(final OperationContext context, final ModelNod
                             DigestPassword.ALGORITHM_DIGEST_MD5,
                             DigestPassword.ALGORITHM_DIGEST_SHA,
                             DigestPassword.ALGORITHM_DIGEST_SHA_256,
+                            DigestPassword.ALGORITHM_DIGEST_SHA_384,
```
bstansberry: This requires an addition to ElytronSubsystemTransformers, tranforming to ELYTRON_1_2_0.

testsuite/elytron/src/test/java/org/wildfly/test/integration/elytron/sasl/mgmt/DigestSha384MgmtSaslTestCase.java
yersan: @jamezp Copyright header was added to this test

elytron/src/main/java/org/wildfly/extension/elytron/ModifiableRealmDecorator.java
yersan: @bstansberry As we have commented by email, I have verified that set-password operation is not registered under /profile=XXXX/ path because it's a runtime op, so it looks like unnecessary to register a transform here against to 1.2.0, Are you agree?

bstansberry: @yersan Yes, I agree. This AttributeDefinition is only used by SetPasswordHandler, and ModifiableRealmDecorator.registerOperations only registers that hander on a server or an HC subsystem, not on a domain-wide /profile=* resource where a transformer is required.

Sorry for steering you in the wrong direction!

bstansberry:
retest this please

bstansberry:
retest this please

bstansberry:
@soul2zimate https://ci.wildfly.org/viewLog.html?buildId=80848&buildTypeId=WildFlyCore_PullRequestWindows is another intermittent failure.

PR #2902, opened 2017-10-23
[WFLY-3377] thread-factory resource creation

jmesnil:
* do not use the values of group-name and thread-name-pattern if the
  attibutes are not defined in the model.

JIRA: https://issues.jboss.org/browse/WFCORE-3377

### DIFF
--- a/threads/src/main/java/org/jboss/as/threads/ThreadFactoryAdd.java
+++ b/threads/src/main/java/org/jboss/as/threads/ThreadFactoryAdd.java
@@ -55,10 +55,12 @@ public class ThreadFactoryAdd extends AbstractAddStepHandler {
     protected void performRuntime(final OperationContext context, final ModelNode operation, final ModelNode model) throws OperationFailedException {
 
         ModelNode priorityModelNode = PoolAttributeDefinitions.PRIORITY.resolveModelAttribute(context, model);
+        ModelNode groupNameModelNode = PoolAttributeDefinitions.GROUP_NAME.resolveModelAttribute(context, model);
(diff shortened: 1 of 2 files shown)


threads/src/main/java/org/jboss/as/threads/ThreadFactoryAdd.java
```diff
@@ -55,10 +55,12 @@ private ThreadFactoryAdd() {
     protected void performRuntime(final OperationContext context, final ModelNode operation, final ModelNode model) throws OperationFailedException {
 
         ModelNode priorityModelNode = PoolAttributeDefinitions.PRIORITY.resolveModelAttribute(context, model);
+        ModelNode groupNameModelNode = PoolAttributeDefinitions.GROUP_NAME.resolveModelAttribute(context, model);
+        ModelNode threadNamePatternModelNode = PoolAttributeDefinitions.THREAD_NAME_PATTERN.resolveModelAttribute(context, model);
 
-        final String threadNamePattern = PoolAttributeDefinitions.THREAD_NAME_PATTERN.resolveModelAttribute(context, model).asString();
+        final String threadNamePattern = threadNamePatternModelNode.isDefined() ? threadNamePatternModelNode.asString() : null;
```
bstansberry: This can all be simplified to 

```
final String threadNamePattern = 
        PoolAttributeDefinitions.THREAD_NAME_PATTERN.resolveModelAttribute(context, model).asStringOrNull();
final String groupName =
        PoolAttributeDefinitions.GROUP_NAME.resolveModelAttribute(context, model).asStringOrNull();
```

I added asXxxOrNull() methods to ModelNode some time this year.

bstansberry: This can be dealt with some other time. The method already uses the old pattern for the other attribute anyway.

PR #2903, opened 2017-10-23
Fix for WFCORE-3363, CLI to output JSON

jfdenise:
- New option, XML element and unit test.

### DIFF
--- a/cli/src/main/java/org/jboss/as/cli/handlers/OperationRequestHandler.java
+++ b/cli/src/main/java/org/jboss/as/cli/handlers/OperationRequestHandler.java
@@ -89,7 +89,7 @@ public class OperationRequestHandler implements CommandHandler, OperationCommand
         try {
             final ModelNode result = ctx.execute(op, "Operation request");
             if (Util.isSuccess(result)) {
-                ctx.printLine(result.toString());
+                ctx.printDMR(result);
             } else {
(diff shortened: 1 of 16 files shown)


cli/src/main/resources/help/help.txt
```diff
@@ -97,6 +98,9 @@ Usage:
  --command-timeout  - The maximum amount of seconds the CLI will wait for a 
                       command to complete.
 
+ --output-json      - makes the CLI to display operation responses using JSON syntax.
```
bstansberry: s/makes/causes

bstansberry:
Minor language thing in the help text aside, looks good!

cli/src/main/java/org/jboss/as/cli/handlers/OperationRequestHandler.java
```diff
@@ -89,7 +89,7 @@ public void handle(CommandContext ctx) throws CommandLineException {
         try {
             final ModelNode result = ctx.execute(op, "Operation request");
             if (Util.isSuccess(result)) {
-                ctx.printLine(result.toString());
+                ctx.printDMR(result);
             } else {
                 throw new CommandLineException(result.toString());
```
bstansberry: I commented on https://developer.jboss.org/wiki/CLI-OutputJSON re: exception handling (i.e. I think this is right.)

PR #2917, opened 2017-11-01
[WFCORE-3393] For ManagementClientChannelStrategy.Establishing incorp…

bstansberry:
…orate the ProtocolConnectionConfiguration connectionTimeout in the deadline for opening the Channel once the connection is opened.

https://issues.jboss.org/browse/WFCORE-3393

@jfdenise Does this solve the problem?


jfdenise:
@bstansberry , thank-you, yes that solves it, channel open will timeout after the expected amount of time.

PR #2918, opened 2017-11-02
[WFCORE-3395] Elytron policy is not validated for write-attribute ope…

TomasHofman:
…ration

https://issues.jboss.org/browse/WFCORE-3395
EAP Issue: https://issues.jboss.org/browse/JBEAP-13585


bstansberry:
I don't want to do this in principle, and this isn't a correct solution.

TomasHofman:
Yes, I wasn't sure about this. Thanks for feedback.

PR #2924, opened 2017-11-06
[WFCORE-3404] Elytron audit non-synchronized by default

honza889:
Elytron logging has performance impact when set to be synchronized.
Changing default to use non-synchronized audit logging.
Keeping audit logging disabled in default configuration. (Even through it should not be necessary for non-sychronized audit logging.)

https://issues.jboss.org/browse/WFCORE-3404
https://issues.jboss.org/browse/JBEAP-10639

### DIFF
--- a/elytron/src/main/resources/subsystem-templates/elytron.xml
+++ b/elytron/src/main/resources/subsystem-templates/elytron.xml
@@ -85,7 +85,7 @@
     <supplement name="domain">
         <replacement placeholder="AUDIT_DEFINITIONS">
             <audit-logging>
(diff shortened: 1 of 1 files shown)


kwart:
@olukas Could you check it, please? Thanks. :)

bstansberry:
@darranl @honza889 Can I close this altogether then? If the issue needs major rework I'd rather close it until that is done.

darranl:
I think for now as we are at the start of a new major release and not in the final stages of one we should close for now and revisit the underlying performance issues first - once we understand those in more detail and have optimised where possible we can then come up with a new recommendation for our default config.

PR #2927, opened 2017-11-07
WFCORE-3409 embed-server doesn't honor "module.path" system property

ctomc:


testsuite/shared/src/main/java/org/jboss/as/test/shared/ModelParserUtils.java
```diff
@@ -45,7 +45,9 @@
 /**
  *
  * @author <a href="mailto:ehugonne@redhat.com">Emmanuel Hugonnet</a> (c) 2014 Red Hat, inc.
+ * @deprecated this belongs to full together with single user of it.
  */
+@Deprecated
```
jmesnil: is that related to the module.path sys prop?

bstansberry:
Hold this; there is other work on going on related to how embedded works with JBoss Modules.

ctomc:
@bstansberry can we just merge this in mean time, as this will helps tremendously with testsuite optimization (no need to copy over modules jar in manualmode)
In any case, this fix/functionality should be present also in reworked version of embedded no matter how it is done

PR #2930, opened 2017-11-10
[WFCORE-3415] Faster resource tree size calculation for getMBeanCount()

bstansberry:
The biggest thing this does is eliminate RBAC checks as a factor in determining whether a resource appears in the count. Which means, in the case of non-addressable resources the total number of mbeans a user could "see", for example via MBeanServerConnection.queryNames, could be less than the number returned by a simultaneous call to getMBeanCount().

The other big thing this does is avoiding doing two complete counts for the jboss.as and jboss.as.expr JMX domains. There should be no difference in counts between those so if both are installed it just counts one and doubles the result.

I have a variant of this that retains the old getMBeanCount logic plus adds a simpler optimization of what's already there, and then adds this, and then times all 3, logging the times and the mbean counts. Against current wildfly master, standalone-full-ha.xml, it reports, for three tests:

```
19:50:42,684 INFO  [org.jboss.as.jmx] (pool-3-thread-1) Elapsed getMBeanCount time using the original RootResourceIterator approach was 752808315, resulting in a count of 805
19:50:42,699 INFO  [org.jboss.as.jmx] (pool-3-thread-1) Elapsed getMBeanCount time using the non-RBAC RootResourceIterator approach was 15326911, resulting in a count of 805
19:50:42,707 INFO  [org.jboss.as.jmx] (pool-3-thread-1) Elapsed getMBeanCount time using the Resource.getMBeanCount() approach was 7532190, resulting in a count of 805
```

```
19:50:57,515 INFO  [org.jboss.as.jmx] (pool-3-thread-1) Elapsed getMBeanCount time using the original RootResourceIterator approach was 303192886, resulting in a count of 805
19:50:57,525 INFO  [org.jboss.as.jmx] (pool-3-thread-1) Elapsed getMBeanCount time using the non-RBAC RootResourceIterator approach was 10050281, resulting in a count of 805
19:50:57,528 INFO  [org.jboss.as.jmx] (pool-3-thread-1) Elapsed getMBeanCount time using the Resource.getMBeanCount() approach was 2476300, resulting in a count of 805
```

```
19:51:04,204 INFO  [org.jboss.as.jmx] (pool-3-thread-1) Elapsed getMBeanCount time using the original RootResourceIterator approach was 231361182, resulting in a count of 805
19:51:04,214 INFO  [org.jboss.as.jmx] (pool-3-thread-1) Elapsed getMBeanCount time using the non-RBAC RootResourceIterator approach was 10143555, resulting in a count of 805
19:51:04,217 INFO  [org.jboss.as.jmx] (pool-3-thread-1) Elapsed getMBeanCount time using the Resource.getMBeanCount() approach was 2076253, resulting in a count of 805
```

The call times are roughly 100 times less, with the first call using the existing method taking 753ms! See https://issues.jboss.org/browse/WFLY-9408 for an example of a user-reported problem due to this. The approach used in this PR was also about 7ms faster than the alternative optimization approach.

If I map the user I'm connecting as to the Monitor RBAC role and then turn on RBAC, you can see the mbean count that results from not using RBAC to trim mbeans the user won't be able to address from the count:

```
19:57:54,278 INFO  [org.jboss.as.jmx] (pool-3-thread-1) Elapsed getMBeanCount time using the original RootResourceIterator approach was 663849859, resulting in a count of 403
19:57:54,295 INFO  [org.jboss.as.jmx] (pool-3-thread-1) Elapsed getMBeanCount time using the non-RBAC RootResourceIterator approach was 17223565, resulting in a count of 805
19:57:54,306 INFO  [org.jboss.as.jmx] (pool-3-thread-1) Elapsed getMBeanCount time using the Resource.getMBeanCount() approach was 9907518, resulting in a count of 805
```

Looking at the JMX specification (https://docs.oracle.com/javase/8/docs/technotes/guides/jmx/JMX_1_4_specification.pdf) I see nothing directly on the topic of the effect of any security checks on the result of getMBeanCount(). Section 12.1.2.4 "Permission Checking for Queries" clearly states that permission checks should remove mbeans that should not be visible according to the permission scheme from the query result, so definitely queries returning a subset of the entire mbean universe is contemplated. Now, that section is about security manager permissions, not our RBAC scheme, but it's the most on topic thing I see.

### DIFF
--- a/controller/src/main/java/org/jboss/as/controller/registry/AbstractModelResource.java
+++ b/controller/src/main/java/org/jboss/as/controller/registry/AbstractModelResource.java
@@ -23,6 +23,7 @@
 package org.jboss.as.controller.registry;
 
 import java.util.ArrayList;
+import java.util.Collection;
 import java.util.Collections;
 import java.util.HashSet;
 import java.util.LinkedHashMap;
@@ -220,6 +221,23 @@ public abstract class AbstractModelResource extends ResourceProvider.ResourcePro
         return orderedChildTypes;
     }
 
+    @Override
+    public int getTreeSize() {
+        int result = 1; // ourself
+        ResourceProvider[] providers; // don't use a Set as we don't want to lose duplicates registered under different keys!
+        synchronized (children) {
+            if (children.isEmpty()){
+                return result;
+            }
+            Collection<ResourceProvider> source = children.values();
+            providers = source.toArray(new ResourceProvider[source.size()]);
+        }
+        for (ResourceProvider provider : providers) {
+            result += provider.getTreeSize();
+        }
+        return result;
+    }
+
     protected void registerResourceProvider(final String type, final ResourceProvider provider) {
         synchronized (children) {
             if (children.containsKey(type)) {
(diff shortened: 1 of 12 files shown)


bstansberry:
@fjuma Another org.jboss.as.test.integration.jsf.doctype.DoctypeDeclTestCase.testDoctypeDeclAllowed failure. Can you ping me about this to discuss details?

PR #2931, opened 2017-11-10
Fix the test.level system property used to override the default root …

jamezp:
…logger level for tests.

I can file a JIRA if needed, but this is just a simple fix to allow the `-Dtest.level` system property to work for tests.

### DIFF
--- a/controller/pom.xml
+++ b/controller/pom.xml
@@ -36,10 +36,6 @@
 
     <name>WildFly: Controller Core</name>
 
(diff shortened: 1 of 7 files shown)


bstansberry:
@fjuma I've seen a number of failures of org.jboss.as.test.integration.jsf.doctype.DoctypeDeclTestCase.testDoctypeDeclAllowed on various CI runs. Can you check it out please?

PR #2933, opened 2017-11-13
Fix for WFCORE-3405 and WFCORE-3406 on find-non-progressing-operation

soul2zimate:
issue: https://issues.jboss.org/browse/WFCORE-3405

1. Set runtime-only and read-only for find-non-progressing-operation and add some tests with domain server address as recommended in WFCORE-3405
2. Acquire exclusive lock for standalone server in BlockerExtension, which I think is the reason why find-non-progressing-operation does not work for standalone in manual test with BlockerExtension. Please see my examples in https://issues.jboss.org/browse/WFCORE-3406


soul2zimate:
I withdraw the context.readResourceForUpdate() call in BlockerExtension for standalone server as it breaks some manual test in standalone mode

PR #2934, opened 2017-11-13
WFCORE-3416 User redirected with HTTP 301 instead of 302 in admin-onl…

TomasHofman:
…y mode

https://issues.jboss.org/browse/WFCORE-3416
https://issues.jboss.org/browse/JBEAP-10363

### DIFF
--- a/domain-http/interface/src/main/java/org/jboss/as/domain/http/server/ManagementRootConsoleRedirectHandler.java
+++ b/domain-http/interface/src/main/java/org/jboss/as/domain/http/server/ManagementRootConsoleRedirectHandler.java
@@ -21,11 +21,11 @@
  */
 package org.jboss.as.domain.http.server;
 
-import static org.jboss.as.domain.http.server.Common.*;
+import static org.jboss.as.domain.http.server.Common.METHOD_NOT_ALLOWED_HANDLER;
+import static org.jboss.as.domain.http.server.Common.NOT_FOUND;
 
 import io.undertow.server.HttpHandler;
 import io.undertow.server.HttpServerExchange;
-import io.undertow.util.Headers;
 import io.undertow.util.HttpString;
 
 /**
@@ -47,15 +47,8 @@ class ManagementRootConsoleRedirectHandler implements HttpHandler {
             return;
         }
 
-        String requestUrl = exchange.getRequestURL();
         if (consoleHandler != null && "/".equals(exchange.getRequestPath())) {
-            if (requestUrl.endsWith("/")) {
-                requestUrl = requestUrl.substring(0, requestUrl.length() - 1);
-            }
-            StringBuilder redirect = new StringBuilder(requestUrl);
-            redirect.append(consoleHandler.getContext()).append(consoleHandler.getDefaultPath());
-            exchange.getResponseHeaders().add(Headers.LOCATION, redirect.toString());
-            MOVED_PERMANENTLY.handleRequest(exchange);
+            consoleHandler.getHandler().handleRequest(exchange);
             return;
         }
         NOT_FOUND.handleRequest(exchange);
--- a/testsuite/standalone/src/test/java/org/jboss/as/test/integration/management/console/WebConsoleRedirectionTestCase.java
+++ b/testsuite/standalone/src/test/java/org/jboss/as/test/integration/management/console/WebConsoleRedirectionTestCase.java
@@ -1,0 +1,76 @@
+/*
+ * JBoss, Home of Professional Open Source
+ * Copyright 2017, Red Hat, Inc., and individual contributors as indicated
+ * by the @authors tag.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.jboss.as.test.integration.management.console;
+
+import static org.junit.Assert.assertEquals;
+import static org.junit.Assert.assertNotNull;
+
+import java.net.HttpURLConnection;
+import java.net.URL;
+
+import javax.inject.Inject;
+
+import org.apache.http.client.methods.HttpGet;
+import org.jboss.as.test.integration.management.util.ServerReload;
+import org.jboss.as.test.shared.TestSuiteEnvironment;
+import org.junit.Test;
+import org.junit.runner.RunWith;
+import org.wildfly.core.testrunner.ManagementClient;
+import org.wildfly.core.testrunner.WildflyTestRunner;
+
+/**
+ * Since console is not loaded in core dist, in both modes the server is supposed to return 404.
+ *
+ * @author Tomas Hofman (thofman@redhat.com)
+ */
+@RunWith(WildflyTestRunner.class)
+public class WebConsoleRedirectionTestCase {
+
+    @SuppressWarnings("unused")
+    @Inject
+    private static ManagementClient managementClient;
+
+    @Test
+    public void testRedirectionInAdminMode() throws Exception {
+        ServerReload.executeReloadAndWaitForCompletion(managementClient.getControllerClient(), true);
+        try {
+            final HttpURLConnection connection = getConnection();
+            assertEquals(HttpURLConnection.HTTP_NOT_FOUND, connection.getResponseCode());
+        } finally {
+            ServerReload.executeReloadAndWaitForCompletion(managementClient.getControllerClient(), false);
+        }
+    }
+
+    @Test
+    public void testRedirectionInNormalMode() throws Exception {
+        final HttpURLConnection connection = getConnection();
+        assertEquals(HttpURLConnection.HTTP_NOT_FOUND, connection.getResponseCode());
+    }
+
+    private HttpURLConnection getConnection() throws Exception {
+        final URL url = new URL("http://" + TestSuiteEnvironment.getServerAddress() + ":" + TestSuiteEnvironment.getServerPort() + "/");
+        final HttpURLConnection connection = (HttpURLConnection) url.openConnection();
+        assertNotNull(connection);
+        connection.setInstanceFollowRedirects(false);
+        connection.setRequestMethod(HttpGet.METHOD_NAME);
+        connection.connect();
+        return connection;
+    }
+
+}
--- a/domain-http/interface/src/main/java/org/jboss/as/domain/http/server/ConsoleMode.java
+++ b/domain-http/interface/src/main/java/org/jboss/as/domain/http/server/ConsoleMode.java
@@ -193,6 +193,10 @@ public enum ConsoleMode {
         }
 
         static ResourceHandlerDefinition createNoConsoleForAdminMode(String slot) throws ModuleLoadException {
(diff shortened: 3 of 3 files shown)


domain-http/interface/src/main/java/org/jboss/as/domain/http/server/ManagementRootConsoleRedirectHandler.java
```diff
@@ -47,15 +46,8 @@ public void handleRequest(HttpServerExchange exchange) throws Exception {
...
-        String requestUrl = exchange.getRequestURL();
         if (consoleHandler != null && "/".equals(exchange.getRequestPath())) {
-            if (requestUrl.endsWith("/")) {
-                requestUrl = requestUrl.substring(0, requestUrl.length() - 1);
-            }
-            StringBuilder redirect = new StringBuilder(requestUrl);
-            redirect.append(consoleHandler.getContext()).append(consoleHandler.getDefaultPath());
-            exchange.getResponseHeaders().add(Headers.LOCATION, redirect.toString());
-            MOVED_PERMANENTLY.handleRequest(exchange);
+            consoleHandler.getHandler().handleRequest(exchange);
```
bstansberry: What happens with ConsoleMode.SLAVE_HC?

testsuite/manualmode/src/test/java/org/jboss/as/test/manualmode/adminonly/WebConsoleRedirectionTestCase.java
```diff
@@ -0,0 +1,76 @@
...
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the GNU
+ * Lesser General Public License for more details.
+ *
+ * You should have received a copy of the GNU Lesser General Public
+ * License along with this software; if not, write to the Free
+ * Software Foundation, Inc., 51 Franklin St, Fifth Floor, Boston, MA
+ * 02110-1301 USA, or see the FSF site: http://www.fsf.org.
+ *
+ */
```
bstansberry: Please use the ASL header in all new files in WildFly Core.

testsuite/manualmode/src/test/java/org/jboss/as/test/manualmode/adminonly/WebConsoleRedirectionTestCase.java
```diff
@@ -0,0 +1,76 @@
...
+
+package org.jboss.as.test.manualmode.adminonly;
```
bstansberry: This should not need a manualmode test, which is more costly in terms of time vs one that uses an existing server and simply reloads it into / out of admin-only.

testsuite/manualmode/src/test/java/org/jboss/as/test/manualmode/adminonly/WebConsoleRedirectionTestCase.java
```diff
@@ -0,0 +1,76 @@
...
+    protected static ServerController controller;
+
+    @Test
+    public void testRedirectionNoConsoleInAdminMode() throws Exception {
+        controller.startInAdminMode();
+
+        final HttpURLConnection connection = getConnection();
+        connection.setRequestMethod(HttpGet.METHOD_NAME);
+        connection.connect();
+        assertEquals(HttpURLConnection.HTTP_MOVED_TEMP, connection.getResponseCode());
```
bstansberry: The fact this passes makes me suspect something is wrong. Without the change to ManagementRootConsoleRedirectHandler what response code is returned? A 404 or a 301?

If it's a 301 there's a separate issue that we'd need to sort out. There is no console possible in a simple WildFly Core server. The console module isn't even present. A redirect to a page that talks about admin-only is not valid.

If it's a 404, then we're good. The test of admin-only would be in WildFly full which is where we ship a console.

domain-http/interface/src/main/java/org/jboss/as/domain/http/server/ManagementRootConsoleRedirectHandler.java
TomasHofman: That behaves the same way as ADMIN_ONLY, only with different URL.

To make sure, I tested this way:
1) Added http management interface into host-slave.xml:
```
<http-interface security-realm="ManagementRealm">                   
    <http-upgrade enabled="true"/>                                  
    <socket interface="management" port="${jboss.management.http.port:9990}"/>                                                                     
</http-interface>
```
2) $ ./domain.sh --host-config=host-slave.xml ...
3) $ curl -s -D - http://123.0.0.1:9990/ 
HTTP/1.1 302 Found
Connection: keep-alive
Location: /consoleerror/noConsoleForSlaveDcError.html

testsuite/manualmode/src/test/java/org/jboss/as/test/manualmode/adminonly/WebConsoleRedirectionTestCase.java
TomasHofman: Will fix.

testsuite/manualmode/src/test/java/org/jboss/as/test/manualmode/adminonly/WebConsoleRedirectionTestCase.java
TomasHofman: It returns 301 before the change. I tried starting the container in normal mode instead of admin-only, and that returns 404 - no console. I will look into why it gives different results.

So the test case would have to go into wildfly testsuite. Probably best if I create a separate WFLY issue then.

Thanks for feedback!

testsuite/manualmode/src/test/java/org/jboss/as/test/manualmode/adminonly/WebConsoleRedirectionTestCase.java
TomasHofman: Will fix.

testsuite/manualmode/src/test/java/org/jboss/as/test/manualmode/adminonly/WebConsoleRedirectionTestCase.java
bstansberry: Thanks. My guess is the logic that decides what ConsoleHandler to use is checking things in the wrong order.

TomasHofman: Okey, so first the BaseHttpInterfaceAddStepHandler sets consoleEnabled attribute to true [1], because that's the default value for the attribute (console-enabled is undefined in the model).

After that, HttpManagementAddHandler selects either ConsoleMode.CONSOLE or ConsoleMode.ADMIN_ONLY [2], based on whether we are in admin-only or not.

If ADMIN_ONLY is selected, it will simply be used to create redirect handler with 302 (which however is not used in current code, hence this PR).

But, if ConsoleMode.CONSOLE was selected, handler creation will fail because of "ModuleNotFoundException: WFLYDMHTTP0007: No console module available with module name org.jboss.as.console.main" [3]. The console handler therefore stays null, which leads to 404 being returned [4] eventually.

Can we set console-enabled to false in default standalone.xml configuration file? I think it's generated in core-feature-pack module. Would that affect wildfly full somehow?

[1] https://github.com/wildfly/wildfly-core/blob/master/controller/src/main/java/org/jboss/as/controller/management/BaseHttpInterfaceAddStepHandler.java#L75
[2] https://github.com/wildfly/wildfly-core/blob/master/server/src/main/java/org/jboss/as/server/operations/HttpManagementAddHandler.java#L204
[3] https://github.com/wildfly/wildfly-core/blob/master/domain-http/interface/src/main/java/org/jboss/as/domain/http/server/ManagementHttpServer.java#L350
[4] https://github.com/wildfly/wildfly-core/blob/master/domain-http/interface/src/main/java/org/jboss/as/domain/http/server/ManagementRootConsoleRedirectHandler.java#L61

TomasHofman:
Test case: https://github.com/wildfly/wildfly/pull/10657

darranl:
Retest this please

core-feature-pack/src/main/resources/configuration/standalone/template.xml
```diff
@@ -46,7 +46,7 @@
             </logger>
         </audit-log>
         <management-interfaces>
-            <http-interface security-realm="ManagementRealm">
+            <http-interface security-realm="ManagementRealm" console-enabled="false">
```
bstansberry: This is probably a good thing to do in general, but I'm a bit concerned changing this could affect someone in unexpected ways. (It doesn't affect full WildFly which has its own templates.) If we did this we'd want to change the host.xml and host-master.xml templates too.

The last commit at https://github.com/wildfly/wildfly-core/compare/master...bstansberry:pullRequest2934 results in admin-only behaving the same as normal when no console module is present.

domain-http/interface/src/main/java/org/jboss/as/domain/http/server/ManagementRootConsoleRedirectHandler.java
bstansberry: Sorry, I misread this and thought it was saying it was giving a 301, which is arguably correct since a slave does not provide a console in normal operation.

But, thinking about it harder I think the 302 is better. A slave is not a slave forever; it can be reconfigured manually and reloaded as a master, and once we do HA DC some slaves will be capable of dynamic switching from slave to master.

core-feature-pack/src/main/resources/configuration/standalone/template.xml
darranl: I agree, I think it is better if the fix is in the code rather than changing a template - an end user can always tweak the config effectively breaking the fix, if the fix is in the code it is always present.

darranl:
Just tested this locally, this does clear up the issue I describe in the Jira where after enabling SSL as the browser was previously passed a permanent redirect it uses a cached version of the console triggering cross origin issues when the console loads.

domain-http/interface/src/main/java/org/jboss/as/domain/http/server/ManagementRootConsoleRedirectHandler.java
darranl: Moving the status from 301 to 302 only impacts this initial redirect, once the browser receives the redirect the remainder of the caching for things like the console remain in place so the browser only requests those if required.  

Initially the server state was reasonably static in this area so the 301 worked well but with ongoing enhancements we have a variety of situations where this is no longer the case so having the root level redirect happen every time a user attempts to use the console is a good thing to ensure they are using it in relation to the current state of the server.

TomasHofman:
I pushed changes:
- the template change removed,
- added Brian's commit causing `DisabledConsoleHandler` to return 404, if console is not present,
- added test case verifying that in both modes the handler returns 404.

testsuite/standalone/src/test/java/org/jboss/as/test/integration/management/console/WebConsoleRedirectionTestCase.java
```diff
@@ -0,0 +1,76 @@
...
+import org.jboss.as.test.integration.management.util.ServerReload;
+import org.jboss.as.test.shared.TestSuiteEnvironment;
+import org.junit.Test;
+import org.junit.runner.RunWith;
+import org.wildfly.core.testrunner.ManagementClient;
+import org.wildfly.core.testrunner.WildflyTestRunner;
+
+/**
+ * Since console is not loaded in core dist, in both modes the server is supposed to return 404.
+ *
```
bstansberry: The PR is failing checkstyle due to trailing spaces on this line.

Otherwise. looks good.

TomasHofman: Didn't notice the failure, fixed.

PR #2937, opened 2017-11-14
[WFCORE-3406] explicitly get the exclusive lock for standalone server in BlockerExtension

soul2zimate:
https://issues.jboss.org/browse/WFCORE-3406
It explicitly gets the exclusive lock for standalone server by calling acquireControllerLock(). JavaDoc on Method states:
> The lock is automatically released when the execute method of the handler that invoked this method returns.

I have manually added new BlockerExtension to test with standalone server as "steps to reproduce" in issue WFCORE-3406. It returns back correct locked time and operation.


bstansberry:
Thanks!

PR #2938, opened 2017-11-14
WFCORE-2933 - /host=foo:add() and embedded hostcontroller support for starting with empty configs

luck3y:


luck3y:
Closing, will reopen shortly after looking at test failures.

PR #2939, opened 2017-11-14
WFCORE-2933 - /host=foo:add() and embedded hostcontroller support for starting with empty configs

luck3y:


luck3y:
@bstansberry I think this should be in better shape to review now. I think I got all of the things changed you'd mentioned in comments yesterday, but I ran into a couple of other testing related issues.

luck3y:
retest this please

PR #2941, opened 2017-11-15
[WFCORE-3412] elytron-tool.* scripts use SEP variable

ivassile:
wildfly-core: https://issues.jboss.org/browse/WFCORE-3412
JBEAP: https://issues.jboss.org/browse/JBEAP-13758


darranl:
@pskopek Can you have a look at this one?  What does the SEP variable control?

As we have released WildFly 11 with this variable supported I am wondering if changing the name could be considered an API change if anyone is making use of it.

pskopek:
@darranl I don't that change of this variable could be considered API change. Users are not supposed to use it. They should use literal separator for respective OS.

darranl:
Thanks @pskopek

PR #2942, opened 2017-11-15
[WFCORE-3417] - JmxControlledStateNotificationsTestCase fails with se…

JiriOndrusek:
…curity manager with IBM JDK in WF core

Issue: https://issues.jboss.org/browse/WFCORE-3417

### DIFF
--- a/testsuite/shared/src/main/java/org/wildfly/test/jmx/ServiceActivatorDeploymentUtil.java
+++ b/testsuite/shared/src/main/java/org/wildfly/test/jmx/ServiceActivatorDeploymentUtil.java
@@ -24,7 +24,14 @@ package org.wildfly.test.jmx;
 import java.io.File;
 import java.io.FilePermission;
 import java.io.IOException;
+import java.security.Permission;
+import java.util.ArrayList;
+import java.util.Arrays;
+import java.util.Collections;
+import java.util.List;
 import java.util.PropertyPermission;
+import java.util.Set;
+
 import org.jboss.as.test.shared.PermissionUtils;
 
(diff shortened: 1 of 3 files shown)


testsuite/shared/src/main/java/org/wildfly/test/jmx/ServiceActivatorDeploymentUtil.java
```diff
@@ -100,7 +111,15 @@ public static void createServiceActivatorListenerDeployment(File destination, St
                 getMBeanPermission(RunningStateJmx.class, targetName, "addNotificationListener"),
                 getMBeanPermission(RunningStateJmx.class, targetName, "removeNotificationListener"),
                 new PropertyPermission("user.dir", "read")
-        ), "permissions.xml");
+        ));
+        if(IS_IBM) {
+            //fix of WFCORE-3417
+            permissions.add(new FilePermission(destination.getAbsolutePath()
+                    .replace("archives", "domains/JmxControlledStateNotificationsTestCase/master")
```
jmesnil: ServiceActivatorDeploymentUtil is an generic utility class, it should not mention something specific to JmxControlledStateNotificationsTestCase.
Instead, you could improve ServiceActivatorDeploymentUtil by adding a `createServiceActivatorListenerDeployment` helper method that would take an additional `Set<Permission>` so that tests that require additional permissions will be able to add them.

JiriOndrusek: It makes sense. I'll update PR to  reflect it. Thank you.

PR #2947, opened 2017-11-16
[WFCORE-3424] Refactoring - use stability monitors instead of listeners

ropalka:


server/src/main/java/org/jboss/as/server/BootstrapImpl.java
```diff
@@ -116,52 +115,14 @@ public BootstrapImpl() {
...
-                        break;
-                    }
-                }
-            }
-        });
+        tracker.addService(Services.JBOSS_AS, applicationServerService).install();
+        final FutureServiceContainer future = new FutureServiceContainer(container);
+        try {
+            container.awaitStability();
+            future.done();
```
bstansberry: I don't think this provides the same behavior. We're done when Services.JBOSS_SERVER_CONTROLLER is up, not when the ServiceContainer is stable. ServerService.start() (aka Services.JBOSS_SERVER_CONTROLLER ) spawns the boot thread which begins asychronously doing work that eventually installs services. If MSC stabilizes before that thread starts installing, then this will be close to being done when Services.JBOSS_SERVER_CONTROLLER is up. But if it doesn't, then it won't stabilize until the work done by the boot thread reaches a stable point.

ropalka:
You are right @bstansberry . I excluded BootstrapImpl changes from this changeset.

bstansberry:
retest this please

bstansberry:
Full integration - Windows fail was a known intermittent failure pattern, but since it results in so many failures I kicked off a retest to make sure nothing got hidden in the noise.

ropalka:
Changes present here cannot have impact on our test suite.

PR #2950, opened 2017-11-21
Revert "[WFCORE-3405] Test for find-non-progressing-operation call fr…

soul2zimate:
…om domain server"

This reverts commit f98c8bf964ffa022313d0759448935f6c89591c5.

Hi @bstansberry , I have locally ran the OperationCancellationTestCase 100 times in a loop from a simple script. But I can not reproduce any failure in almost 500 runs. Since it makes too much noise on CI results (it does show some frequency on recent CI results), this reverts the new tests added in previous PR. Let's keep the operation accessible change made in https://github.com/wildfly/wildfly-core/pull/2933. 

I need to tweak a bit more about the test locally and see where the intermittent failures come from. Once it's identified, I will make a new PR to add them back.


bstansberry:
Thanks!

PR #2952, opened 2017-11-22
Minor code cleanup

ropalka:


ropalka:
What do you think about the latch removal commit @stuartwdouglas ?

PR #2963, opened 2017-11-30
[WFCORE-3441] Undeprecate

bstansberry:
https://issues.jboss.org/browse/WFCORE-3441

### DIFF
--- a/controller/src/main/java/org/jboss/as/controller/SubsystemRegistration.java
+++ b/controller/src/main/java/org/jboss/as/controller/SubsystemRegistration.java
@@ -71,9 +71,7 @@ public interface SubsystemRegistration {
      * the subsystem's configuration to XML.
      *
      * @param writer the writer
-     * @deprecated use the {@link #registerXMLElementWriter(Supplier)} variant
      */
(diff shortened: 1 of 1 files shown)


bstansberry:
@jmesnil FYI this is in response to a good comment you made on https://github.com/wildfly/wildfly/pull/10641

bstansberry:
@ctomc @rhusar @jmesnil Please see my comment on WFLY-9597.  The most memory efficient way to deal with PersistentResourceXmlParser is to use a Supplier for legacy parsers and share the instance for the current version parser and the writer. It seems strange to me to deprecate the most efficient method.

The metaspace used for one of these method references is 1360 bytes, not 4 bytes. Ping me in chat if you'd like details.

Even if people don't share an instance for the current parser and the writer, a PersistentResourceXmlParser itself is only an expensive object if it retains the PersistentResourceXMLDescription in an instance field. Doing that is a kind of worst practice.

jmesnil:
@rhusar @ctomc after reading Brian comment, does it change your disapproval of the PR?

PR #2965, opened 2017-12-01
[WFCORE-3440] CLI incorrectly parses byte literals

TomasHofman:
https://issues.jboss.org/browse/WFCORE-3440

### DIFF
--- a/cli/src/main/java/org/jboss/as/cli/parsing/arguments/ArgumentValueCallbackHandler.java
+++ b/cli/src/main/java/org/jboss/as/cli/parsing/arguments/ArgumentValueCallbackHandler.java
@@ -215,7 +215,14 @@ public class ArgumentValueCallbackHandler implements ParsingStateCallbackHandler
             String b = buf.toString().trim();
             Byte val;
             if (b.startsWith("0x")) {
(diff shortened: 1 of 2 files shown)


cli/src/main/java/org/jboss/as/cli/parsing/arguments/ArgumentValueCallbackHandler.java
```diff
@@ -215,7 +215,13 @@ private Byte buildByte() {
             String b = buf.toString().trim();
             Byte val;
             if (b.startsWith("0x")) {
-                val = Byte.parseByte(b.substring(2), 16);
+                // when reading a hexadecimal value, treat it as unsigned byte (ie numbers from 0x0 to 0xff) rather than
+                // standard java byte range -0x80 to 0x7f, because that is how byte values are printed by CLI
```
jfdenise: A minor point, instead of "are printed by CLI", you could write "are specified by the DMR text syntax".

TomasHofman: Sounds much better, fixed :).

bstansberry:
Nice!

PR #2968, opened 2017-12-04
WFCORE-3445 Update jandex to 2.0.4.Final

ctomc:
includes https://issues.jboss.org/browse/JANDEX-41 fix for running on JDK9


bstansberry:
Thanks. For sure this is needed, as it came into 3.x via #2905. So missing it here is a regression.

PR #2970, opened 2017-12-05
[WFCORE-887] Log an "expressions deprecated" warning if an expression…

bstansberry:
… is used for an attribute that is either a capability reference or has been specifically flagged to produce the warn.

https://issues.jboss.org/browse/WFCORE-887


bstansberry:
I removed the Hold label on this as I've done an analysis of what attributes this would affect. See my comment on WFCORE-887.

PR #2971, opened 2017-12-05
Fix for WFCORE-3442, reload with changed authentication configuration

jfdenise:
Makes the reload handler to attempt a reconnect if authentication failed. This will give a change to provide new credentials.
Added unit test.

### DIFF
--- a/cli/src/main/java/org/jboss/as/cli/handlers/ReloadHandler.java
+++ b/cli/src/main/java/org/jboss/as/cli/handlers/ReloadHandler.java
@@ -26,6 +26,7 @@ import java.io.IOException;
 import java.util.Arrays;
 import java.util.Collection;
 import java.util.concurrent.atomic.AtomicReference;
+import javax.security.sasl.SaslException;
 
 import org.jboss.as.cli.CommandContext;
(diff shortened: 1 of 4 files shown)


bstansberry:
11.5 more seconds in another manual mode test. :(

To be fair, dealing with this one in the main test suite would be problematic, as there is the risk that the server is left all screwed up if it fails, with later tests unable to connect.

Hmm, could this set up a native-interface and use that instead of altering the http-interface?

An alternative is to set up a native-interface and use that for setup/cleanup so you know it will be available for cleanup even if the http-interface is left in a broken state due to a failure. But that sounds like too much work.

jfdenise:
With respect to putting the server into an invalid state if something goes wrong, in any case the standalone.xml will be put back to its original state. 
@BeforeClass first thing is to copy the original standalone.xml. @AfterClass (called in all cases) first thing is to replace the updated standalone.xml file with a clean one copied in @BeforeClass.

With respect to the time spent in this test, I am ready to move it where it would make more sense.

jfdenise:
The test has been moved to standalone and rewritten to do explicit cleanup of the server state. Time is now 2secs.

PR #2972, opened 2017-12-05
[WFCORE-3448] Handle the situation where the ssl enabled-protocol and…

dehort:
… enabled-cipher-suites are set as system properties that contain multiple values


bstansberry:
I put the Hold label on this because while this solution might be fine for a downstream backport branch, I'm not sure it's how I want this handled in master. This is a basically a one-off workaround for 1 attribute.

bstansberry:
#2973 is a generally applicable fix.

bstansberry:
Thanks, @dehort . The other variant is merged but we'll use this one downstream.

PR #2973, opened 2017-12-06
[WFCORE-3448] Have AttributeDefinition impls for simple lists deal wi…

bstansberry:
…th expressions that resolve to list-parseable strings

https://issues.jboss.org/browse/WFCORE-3448

### DIFF
--- a/controller/src/main/java/org/jboss/as/controller/PrimitiveListAttributeDefinition.java
+++ b/controller/src/main/java/org/jboss/as/controller/PrimitiveListAttributeDefinition.java
@@ -26,6 +26,8 @@ import java.util.List;
 import java.util.Locale;
 import java.util.ResourceBundle;
 
+import javax.xml.stream.XMLStreamException;
+
 import org.jboss.as.controller.descriptions.ModelDescriptionConstants;
 import org.jboss.as.controller.descriptions.ResourceDescriptionResolver;
 import org.jboss.as.controller.operations.validation.ModelTypeValidator;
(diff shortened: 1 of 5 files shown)


controller/src/main/java/org/jboss/as/controller/PrimitiveListAttributeDefinition.java
```diff
@@ -102,6 +104,38 @@ public void removeCapabilityRequirements(OperationContext context, Resource reso
...
+    ModelNode parseResolvedValue(ModelNode original, ModelNode resolved) {
+        return parseSingleElementToList(this, original, resolved);
+    }
+
+    static ModelNode parseSingleElementToList(AttributeDefinition ad, ModelNode original, ModelNode resolved) {
+        ModelNode result = resolved;
+        if (original.isDefined()
+                && !resolved.equals(original)
+                && resolved.getType() == ModelType.LIST
+                && resolved.asInt() == 1) {
```
ctomc: is this correct? this would only pass if value would be "1" while having type of LIST which is kind a hard.

did you maybe mean resolve.asList().size() == 1?

controller/src/main/java/org/jboss/as/controller/PrimitiveListAttributeDefinition.java
```diff
@@ -102,6 +104,38 @@ public void removeCapabilityRequirements(OperationContext context, Resource reso
...
+            // So that implies we had an expression as the element, which is what we would have gotten
+            // if the expression string was passed by an xml parser to parseAndSetParameter. See if the
+            // resolved form of that expression in turn parses to a list and if it does, used the parsed list.
+            ModelNode element = resolved.get(0);
+            if (element.getType() == ModelType.STRING) {
+                ModelNode holder = new ModelNode();
+                try {
+                    ad.getParser().parseAndSetParameter(ad, element.asString(), holder, null);
+                    ModelNode parsed = holder.get(ad.getName());
+                    if (parsed.getType() == ModelType.LIST && parsed.asInt() > 1) {
```
ctomc: in similar way here

controller/src/main/java/org/jboss/as/controller/PrimitiveListAttributeDefinition.java
bstansberry: asInt() for ModelType.LIST returns the list size. 

https://github.com/jbossas/jboss-dmr/blob/master/src/main/java/org/jboss/dmr/ListModelValue.java#L104

For OBJECT it returns the number of keys.

PR #2991, opened 2017-12-12
[WFCORE-3455] Reduce lambda usage (controller)

dmlloyd:


controller/src/main/java/org/jboss/as/controller/AttributeParsers.java
```diff
@@ -239,13 +239,21 @@ public void parseElement(AttributeDefinition attribute, XMLExtendedStreamReader
...
-            Map<String, AttributeDefinition> attributes = Arrays.stream(valueTypes)
-                    .collect(Collectors.toMap(AttributeDefinition::getXmlName, Function.identity()));
-
-            Map<String, AttributeDefinition> attributeElements = Arrays.stream(valueTypes)
-                                   .filter(attributeDefinition -> attributeDefinition.getParser().isParseAsElement())
-                    .collect(Collectors.toMap(a -> a.getParser().getXmlName(a) , Function.identity()));
+            Map<String, AttributeDefinition> attributes = new HashMap<>();
+            for (AttributeDefinition valueType : valueTypes) {
+                if (attributes.put(valueType.getXmlName(), valueType) != null) {
+                    throw new IllegalStateException("Duplicate key");
```
bstansberry: This technically should be i18n, the overhead of which probably defeats the purpose of this PR. But really the duplicate check should just be removed. It was a side effect of Collectors.toMap and doing that kind of check is not the responsibility of this method. So we can save more code. ;)

controller/src/main/java/org/jboss/as/controller/AttributeParsers.java
```diff
@@ -239,13 +239,21 @@ public void parseElement(AttributeDefinition attribute, XMLExtendedStreamReader
...
+                }
+            }
 
+            Map<String, AttributeDefinition> attributeElements = new HashMap<>();
+            for (AttributeDefinition attributeDefinition : valueTypes) {
+                if (attributeDefinition.getParser().isParseAsElement()) {
+                    if (attributeElements.put(attributeDefinition.getParser().getXmlName(attributeDefinition), attributeDefinition) != null) {
+                        throw new IllegalStateException("Duplicate key");
```
bstansberry: Same comment as L245

dmlloyd:
Fixed.  Also fixed a little unnecessary formatting change.

PR #2992, opened 2017-12-12
[WFCORE-3455] Reduce lambda usage (cli)

dmlloyd:


bstansberry:
@jfdenise Please approve. This is part of a metaspace saving effort. Not so critical in the CLI but part of the point is to avoid having code that uses lambdas instead of simple loops so people won't see that in our code and adapt it server-side code.

jfdenise:
Approved.

PR #2993, opened 2017-12-12
[WFCORE-3454] Ensure the dist archives contain the correct content. R…

jamezp:
…enamed the wildfly-vault-test-feature-pack maven module to wildfly-core-vault-test-feature-pack.

https://issues.jboss.org/browse/WFCORE-3454

One thing to note here is I did rename the `wildfly-vault-test-feature-pack` maven module to `wildfly-core-vault-test-feature-pack`. This was due to the fact that the default `server.output.dir.prefix` property was set to `wildfly` and for core should likely be `wildfly-core`. This feature pack seems to only be used in WildFly Core. If we feel this shouldn't change I can remove that change. I didn't see any usage in WildFly.

### DIFF
--- a/dist/assembly.xml
+++ b/dist/assembly.xml
@@ -28,6 +28,7 @@
     <id>distro</id>
     <formats>
        <format>zip</format>
+        <format>tar.gz</format>
     </formats>
 
     <includeBaseDirectory>false</includeBaseDirectory>
@@ -37,7 +38,7 @@
             <directory>target</directory>
             <outputDirectory/>
             <includes>
-                <include>${server.output.dir.prefix}-core-${project.version}/**</include>
+                <include>${server.output.dir.prefix}-${wildfly.core.release.version}/**</include>
             </includes>
         </fileSet>
     </fileSets>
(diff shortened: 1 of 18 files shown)


bstansberry:
The rename of wildfly-vault-test-feature-pack is fine. That FP is just there to make picketbox available so we can test vault. WildFly has picketbox the normal way.

jamezp:
That's what it looked like which is why I thought renaming it wasn't a big deal. To be honest this fix really could be a one line fix in the `assembly.xml`, but I figured we may as well make it match what we do in full as close as we can.

dist/assembly.xml
```diff
@@ -37,7 +38,7 @@
             <directory>target</directory>
             <outputDirectory/>
             <includes>
-                <include>${server.output.dir.prefix}-core-${project.version}/**</include>
+                <include>${server.output.dir.prefix}-${wildfly.core.release.version}/**</include>
```
bstansberry: Note for historical research: this bit is the key change in the PR; before it was using the wrong value, causing the bug.

PR #2996, opened 2017-12-15
[WFCORE3470] Configure service display name

dleborgne:
Added an option to configure the service display name during service install (WFLY-9619)

### DIFF
--- a/core-feature-pack/src/main/resources/content/docs/contrib/scripts/service/service.bat
+++ b/core-feature-pack/src/main/resources/content/docs/contrib/scripts/service/service.bat
@@ -113,6 +113,7 @@ echo(
 
 rem defaults
 set SHORTNAME=Wildfly
+rem NO quotes around the display name here !
 set DISPLAYNAME=WildFly
 rem NO quotes around the description here !
 set DESCRIPTION=WildFly Application Server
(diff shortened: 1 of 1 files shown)


dleborgne:
I think the patch is ready for review

ctomc:
this is ok to test

core-feature-pack/src/main/resources/content/docs/contrib/scripts/service/service.bat
```diff
@@ -147,10 +148,11 @@ echo Usage:
 echo(
 echo   service install ^<options^>  , where the options are:
 echo(
-echo     /startup                  : Set the service to auto start
+echo     /startup                    : Set the service to auto start
```
dmlloyd: The spacing was changed here

core-feature-pack/src/main/resources/content/docs/contrib/scripts/service/service.bat
```diff
@@ -195,13 +204,14 @@ echo(
...
 echo                                 default: %LOGLEVEL%
+echo(
 echo     /logpath ^<path^>           : Path of the log
 echo                                 default depends on domain or standalone mode
 echo                                 /base applies when /logpath is not set.
 echo                                   %JBOSS_HOME%\domain\log
 echo                                   %JBOSS_HOME%\standalone\log
 echo(
-echo     /debug                    : run the service install in debug mode
+echo     /debug                      : run the service install in debug mode
```
dmlloyd: Spacing also changed here

dmlloyd:

dleborgne:
Yes, I agree that my contributions are made under the terms of the MIT License.

bstansberry:
I moved the JIRA for this to WFCORE as the Fix Version tracking will be against a WildFly Core release. I'm going to change the title of this PR to use the new JIRA # WFCORE-3470 so the git log will include it in the history (to help find this PR), but there's no need to change the commit message. If any goes to WFLY-9619 JIRA will redirect them to WFCORE-3470.

PR #3002, opened 2017-12-20
WFCORE-3469 add legacy controller for EAP 7.1

ctomc:


bstansberry:
I sense a transformer test pruning PR for full is coming. ;)

bstansberry:
@ctomc I've forgotten the specific reason I had to restore that superclass, but I know I wasn't happy to do it. Is there any chance something like your "possible workaround" would let us remove it again?

This is not a real important question, so please don't waste your time on it if the answer isn't obvious.

ctomc:
@bstansberry I think i got a working solution now that works for both cases.
And with this it looks like we can even remove that classes from master.
I just need to do some more testing before I remove them.

elytron/src/test/java/org/wildfly/extension/elytron/SubsystemTransformerTestCase.java
```diff
@@ -63,14 +61,14 @@ public void testRejectingTransformersWFCore300() throws Exception {
...
-        ModelTestUtils.checkFailedTransformedBootOperations(mainServices, elytronVersion, ops, new FailedOperationTransformationConfig()
+        /*ModelTestUtils.checkFailedTransformedBootOperations(mainServices, elytronVersion, ops, new FailedOperationTransformationConfig()
                 .addFailedAttribute(subsystemAddress.append(PathElement.pathElement(ElytronDescriptionConstants.SIMPLE_PERMISSION_MAPPER)),
                         new MatchAllConfig()
                 )
                 .addFailedAttribute(subsystemAddress.append(PathElement.pathElement(ElytronDescriptionConstants.AUTHENTICATION_CONFIGURATION)),
                         new FailedOperationTransformationConfig.NewAttributesConfig(AuthenticationClientDefinitions.FORWARDING_MODE)
                 )
-        );
+        );*/
```
bstansberry: Please restore this test and add a new one for EAP 7.1.  We may be altering code in this part of the model and I want this test to alert us in case we break things in a way we didn't expect. I don't care all that much long term about transforming to WF Core 3.0.0 but I'm not in a hurry to drop this once until the rework is done.

ctomc: That is actually it. Test was updated to properly test transformation to 7.1.
But as there are no model changes in elytron subsystem since 7.1 and master, reject check cannot reject anything (yet)

ctomc: I've pushed another commit that adds 7.1 transformers testing as well as cleaned up transformers registration for 7.1.
But as currently master == 7.1 model wise, there are no transformers and tests catch noting.

PR #3006, opened 2017-12-20
[WFCORE-3475] Tighten down field visibility in AttributeDefinition cl…

bstansberry:
…asses

https://issues.jboss.org/browse/WFCORE-3475


bstansberry:
The core CI failures are the same in both Windows and Linux, but I don't see how this PR is relevant and I see them occurring elsewhere too. I'll rebase this and we'll see if they continue.

jamezp:
retest this please

controller/src/main/java/org/jboss/as/controller/SimpleListAttributeDefinition.java
```diff
@@ -167,12 +167,12 @@ public Builder setWrapXmlList(boolean wrap) {
         }
 
         public SimpleListAttributeDefinition build() {
-            if (attributeMarshaller == null) {
-                attributeMarshaller = AttributeMarshallers.getSimpleListMarshaller(wrapXmlList);
+            if (getAttributeMarshaller() == null) {
+                setAttributeMarshaller(AttributeMarshallers.getSimpleListMarshaller(wrapXmlList));
             }
             //todo add parser for SimpleListAttributeDefinition, for now no one is using it yet.
```
jmesnil: Is that TODO still relevant?
If that's the case it should point to a WFCORE to track it

ctomc: It is no longer relevant, we should remove this comment

PR #3009, opened 2017-12-21
[WFCORE-3282] Duplicate record is written in property file when group…

xstefank:
… name set to empty

https://issues.jboss.org/browse/WFCORE-3282

performance testing (still using possessive quantifiers) - https://github.com/xstefank/wildfly-core/tree/WFCORE-3282-performance


xstefank:
@bstansberry can you take a look at this please?

PR #3010, opened 2017-12-21
[WFCORE-3458] External CS, PKCS11 can't be configured with externalPath

ivassile:
elytron-subsystem: https://issues.jboss.org/browse/WFCORE-3458
JBEAP: https://issues.jboss.org/browse/JBEAP-13978

Removing the requirement "if location is not set in CLI, default
credential-store name is used as location"

When location is not set for most common filebased keystore types (JKS,
JCEKS and PKCS12), Elytron will throw exception.

Elytron PR: https://github.com/wildfly-security/wildfly-elytron/pull/1051

### DIFF
--- a/elytron/src/main/java/org/wildfly/extension/elytron/CredentialStoreResourceDefinition.java
+++ b/elytron/src/main/java/org/wildfly/extension/elytron/CredentialStoreResourceDefinition.java
@@ -32,8 +32,11 @@ import java.security.NoSuchAlgorithmException;
 import java.security.Provider;
 import java.security.spec.InvalidKeySpecException;
 import java.util.ArrayList;
+import java.util.Arrays;
+import java.util.Collections;
 import java.util.HashMap;
 import java.util.List;
+import java.util.Locale;
 import java.util.Map;
 import java.util.concurrent.TimeUnit;
 import java.util.concurrent.TimeoutException;
@@ -76,6 +79,7 @@ import org.wildfly.security.credential.PasswordCredential;
 import org.wildfly.security.credential.store.CredentialStore;
 import org.wildfly.security.credential.store.CredentialStoreException;
 import org.wildfly.security.credential.store.UnsupportedCredentialTypeException;
+import org.wildfly.security.credential.store.impl.KeyStoreCredentialStore;
 import org.wildfly.security.password.PasswordFactory;
 import org.wildfly.security.password.interfaces.ClearPassword;
 import org.wildfly.security.password.spec.ClearPasswordSpec;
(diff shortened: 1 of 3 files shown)


mchoma:
My understaning is with this proper fix we don't need fix introduced by https://github.com/wildfly-security/wildfly-elytron/pull/1051 . Or am I missing something?

ivassile:
@mchoma Correct. I'll submit new PR which will revert change [1], but we need another change in Elytron: When useExternalStorage=true and externalPath is null we should not default externalPath to location, but throw an exception. 

[1] wildfly-security/wildfly-elytron#1051

mchoma:
+1 for "When useExternalStorage=true and externalPath is null we should not default externalPath to location, but throw an exception."
@pskopek , wdyt?

ivassile:
@mchoma @pskopek Implemented the above requirement [1]. The issue should be fixed in Elytron Subsystem and Elytron.

[1] wildfly-security/wildfly-elytron#1051

mchoma:
Can we get merged this as Elytron part of fix is already in wf-core?
wildfly-security/wildfly-elytron#1051

bstansberry:
@darranl Please have an SME review this.

elytron/src/main/java/org/wildfly/extension/elytron/CredentialStoreResourceDefinition.java
```diff
@@ -291,6 +294,16 @@ protected void performRuntime(OperationContext context, ModelNode operation, Res
             String relativeTo = asStringIfDefined(context, RELATIVE_TO, model);
             ServiceTarget serviceTarget = context.getServiceTarget();
 
+            if (type == null || type.equals(KeyStoreCredentialStore.KEY_STORE_CREDENTIAL_STORE)) {
+                implementationAttributes.putIfAbsent(CS_KEY_STORE_TYPE_ATTRIBUTE, "JCEKS");
+            }
+
+            List<String> filebasedKeystoreTypes = Arrays.asList("JKS", "JCEKS", "PKCS12");
+            String implAttrKeyStoreType = implementationAttributes.get(CS_KEY_STORE_TYPE_ATTRIBUTE);
+            if (location == null && implAttrKeyStoreType != null && filebasedKeystoreTypes.stream().anyMatch(implAttrKeyStoreType::equalsIgnoreCase)) {
```
darranl: We probably should avoid a strem / lambda here - we should just be able to convert the value to upper case and compare against the collection.

elytron/src/main/java/org/wildfly/extension/elytron/CredentialStoreResourceDefinition.java
```diff
@@ -291,6 +294,16 @@ protected void performRuntime(OperationContext context, ModelNode operation, Res
             String relativeTo = asStringIfDefined(context, RELATIVE_TO, model);
             ServiceTarget serviceTarget = context.getServiceTarget();
 
+            if (type == null || type.equals(KeyStoreCredentialStore.KEY_STORE_CREDENTIAL_STORE)) {
+                implementationAttributes.putIfAbsent(CS_KEY_STORE_TYPE_ATTRIBUTE, "JCEKS");
+            }
+
+            List<String> filebasedKeystoreTypes = Arrays.asList("JKS", "JCEKS", "PKCS12");
```
darranl: As this is a constant list it can probably be created once and wrapped in a Collections.unModifiableList() as a static member variable.

darranl:
Other than my comment about the lambda these changes look good.

darranl:
@wildfly/mergers-core Can someone please add a rebase label to this one as well as it contains a conflict.

ivassile:
@darranl Implemented requested changes and did rebase.

mchoma:
Can we get merged this into as Elytron part of fix is already in wf-core? But without this change fix as a whole is incomplete.

wildfly-security/wildfly-elytron#1051

mchoma:
@wildfly/mergers-core we missed this to be in WF12?

jmesnil:
@mchoma We have not merged this PR as @darranl has requested changes with the GitHub review UI.

Looking at his comments, he approved it in a comment but when we review and merge PRs, we will not look at comments for those that are requesting changes from GitHub UI.

jmesnil:
retest this please

darranl:
Sounds like the scenario in the review process could use a Feature Request to GitHub - after further commits are pushed to a branch it would be better if the UI indicated a subsequent update has been applied since last review.

PR #3013, opened 2018-01-01
[WFCORE-3483] Deprecate PathAddress.navigate and remove, replace controller catches of NoSuchElementException with Resource.NoSuchResourceException

bstansberry:
https://issues.jboss.org/browse/WFCORE-3483

Minor thing to help make it easier to move to a cleaner exception model in future management code.

### DIFF
--- a/controller/src/main/java/org/jboss/as/controller/PathAddress.java
+++ b/controller/src/main/java/org/jboss/as/controller/PathAddress.java
@@ -314,7 +314,14 @@ public class PathAddress implements Iterable<PathElement> {
      * @param create {@code true} to create the last part of the node if it does not exist
      * @return the submodel
      * @throws NoSuchElementException if the model contains no such element
+     *
+     * @deprecated manipulating a deep DMR node tree via PathAddress is no longer how the management layer works
+     *             internally, so this method has become legacy cruft. Management operation handlers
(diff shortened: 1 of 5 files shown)


controller/src/main/java/org/jboss/as/controller/PathAddress.java
```diff
@@ -314,7 +314,11 @@ public PathAddress append(String key) {
      * @param create {@code true} to create the last part of the node if it does not exist
      * @return the submodel
      * @throws NoSuchElementException if the model contains no such element
+     *
+     * @deprecated manipulating a deep DMR node tree via PathAddress is no longer how the management layer works
```
jmesnil: Should the deprecated message mention the correct alternative?

jmesnil:
retest this please

controller/src/main/java/org/jboss/as/controller/PathAddress.java
bstansberry: @jmesnil I updated this with a pointer to Resource and OperationContext.[read|remove]Resource. TBH I don't think these have been used in many years as they haven't done anything useful since we introduced Resource for AS 7.0 or 7.1.

jmesnil:
retest this please

PR #3014, opened 2018-01-03
Evaluate to the current year in code templates.

soul2zimate:
it's 2018 now.


bstansberry:
What I've read on this subject from Red Hat legal is that if we even have dates like these in our headers we should not be changing them like this.

darranl:
This is the template used by Eclipse for new classes added using Eclipse - not a template to retrospectively edit other years.

soul2zimate:
Should I update to fixed value 2018 instead of property value?

darranl:
@soul2zimate I think your current proposed change is correct.

bstansberry:
Thanks, Darran!

PR #3015, opened 2018-01-03
[WFCORE-1383] Startup success detection mechanism fails if JBOSS_HOME…

gaol:
… and jboss.server.base.dir are set differently.

Jira: https://issues.jboss.org/browse/WFCORE-1383

Proposed fix for this issue:

The marker file is created at: $JBOSS_BASE_DIR/tmp/startup-marker instead of under $JBOSS_HOME when the actual JVM process starts, and there is no simple way to know what the JBOSS_BASE_DIR is before the standalone.sh|domain.sh starts.

Introducing the JBOSS_BASE_DIR variable into the startup scripts make it simple to implement, and it works.


gaol:
retest this please

PR #3019, opened 2018-01-04
Fix for WFCORE-3490, better message for non iterable operations

jfdenise:
Catch asList exception and throw better message. Added unit test.


cli/src/main/java/org/jboss/as/cli/handlers/loop/ForControlFlow.java
```diff
@@ -78,7 +78,11 @@ static ForControlFlow get(CommandContext ctx) {
             throw new CommandLineException("iterable request failed, no result");
         }
         ModelNode mn = targetValue.get(Util.RESULT);
-        result = mn.asList();
+        try {
+            result = mn.asList();
+        } catch (Exception ex) {
+            throw new CommandLineException("for cannot be used with operations producing non iterable result");
```
bstansberry: s/producing non iterable/that produce a non-iterable/g

jfdenise:
retest this please!

PR #3021, opened 2018-01-07
[WFCORE-973][WFCORE-3493] Don't store a minValue or maxValue in OpenM…

bstansberry:
…BeanParameterInfo if the value is not numeric

https://issues.jboss.org/browse/WFCORE-3493


bstansberry:
@jmesnil Thanks; good catch. I updated this to uncomment that.

PR #3024, opened 2018-01-08
Fix for WFCORE-3494, CLI, infinite loop when accepting temporarily SSL certificate

jfdenise:
Add certificate with unique ID. Manual testing.

### DIFF
--- a/cli/src/main/java/org/jboss/as/cli/impl/CommandContextImpl.java
+++ b/cli/src/main/java/org/jboss/as/cli/impl/CommandContextImpl.java
@@ -50,6 +50,7 @@ import java.util.Map;
 import java.util.Objects;
 import java.util.ServiceLoader;
 import java.util.Set;
+import java.util.UUID;
 import java.util.concurrent.Callable;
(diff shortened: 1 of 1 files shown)


cli/src/main/java/org/jboss/as/cli/impl/CommandContextImpl.java
```diff
@@ -1978,7 +1979,7 @@ private synchronized X509TrustManager getDelegate() {
                         theTrustStore.load(null);
                     }
                     for (X509Certificate current : temporarilyTrusted) {
-                        theTrustStore.setCertificateEntry(current.getSubjectX500Principal().getName(), current);
+                        theTrustStore.setCertificateEntry(UUID.randomUUID().toString(), current);
```
bstansberry: Are these ever cleaned up?

jfdenise: They are not persisted, they are just added to the key-store instance but not stored. 
So they are present until a new delegate is created.

PR #3027, opened 2018-01-09
[WFCORE-3500] Upgrade JSON-P from 1.0.4 to 1.1.2. This is a requireme…

jamezp:
…nt for Java EE 8.

https://issues.jboss.org/browse/WFCORE-3500

Previously the implementation packed the API. The 1.1 implementation now correctly does not package the API. This is why there is now a hard dependency on the API.

### DIFF
--- a/core-feature-pack/pom.xml
+++ b/core-feature-pack/pom.xml
@@ -44,6 +44,10 @@
 
     <dependencies>
         <dependency>
+            <groupId>javax.json</groupId>
+            <artifactId>javax.json-api</artifactId>
+        </dependency>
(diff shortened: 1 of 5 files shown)


bstansberry:
I added the 'hold' label as this needs an EE7 compatibility strategy.

jamezp:
This could be merged after #10834 or simply closed. WildFly is now overriding the version of JSON-P provided by WildFly Core.

jamezp:
Closing this for now. We'll need to upgrade at a later date, but WIldFly is currently overriding this module.

PR #3028, opened 2018-01-10
WFCORE-3501 Allow System properties to be set at parse time before ex…

stuartwdouglas:
…tensions are loaded in standalone mode

### DIFF
--- a/controller/src/main/java/org/jboss/as/controller/parsing/DeferredExtensionContext.java
+++ b/controller/src/main/java/org/jboss/as/controller/parsing/DeferredExtensionContext.java
@@ -1,0 +1,146 @@
+/*
+ * Copyright 2018 Red Hat, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *   http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.jboss.as.controller.parsing;
+
+import java.util.ArrayList;
+import java.util.LinkedHashMap;
+import java.util.List;
+import java.util.Map;
+import java.util.concurrent.Callable;
+import java.util.concurrent.ExecutionException;
+import java.util.concurrent.ExecutorService;
+import java.util.concurrent.Future;
+
+import javax.xml.stream.XMLStreamException;
+
+import org.jboss.as.controller.Extension;
+import org.jboss.as.controller.extension.ExtensionRegistry;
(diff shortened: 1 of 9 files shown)


controller/src/main/java/org/jboss/as/controller/parsing/DeferredExtensionContext.java
```diff
@@ -0,0 +1,149 @@
+/*
+ * JBoss, Home of Professional Open Source
+ * Copyright 2018 Red Hat Inc. and/or its affiliates and other contributors
+ * as indicated by the @authors tag. All rights reserved.
+ * See the copyright.txt in the distribution for a
+ * full listing of individual contributors.
+ *
+ * This copyrighted material is made available to anyone wishing to use,
+ * modify, copy, or redistribute it subject to the terms and conditions
+ * of the GNU Lesser General Public License, v. 2.1.
```
ctomc: this should be APL not LGPL.

dmlloyd:
Maybe it would be simpler if _all_ properties were set at parse time, _except_ for vault expressions which are set later in the second pass?

server/src/main/java/org/jboss/as/server/controller/resources/SystemPropertyResourceDefinition.java
```diff
@@ -69,27 +70,35 @@
...
 
+    public static final SimpleAttributeDefinition PARSE_TIME = SimpleAttributeDefinitionBuilder.create(ModelDescriptionConstants.PARSE_TIME, ModelType.BOOLEAN, true)
+            .setValidator(new ModelTypeValidator(ModelType.BOOLEAN, true))
+            .setDefaultValue(new ModelNode(false))
+            .setAllowExpression(false)
+            .build();
+
     static final AttributeDefinition[] ALL_ATTRIBUTES = new AttributeDefinition[] {VALUE, BOOT_TIME};
-    static final AttributeDefinition[] SERVER_ATTRIBUTES = new AttributeDefinition[] {VALUE};
+    static final AttributeDefinition[] SERVER_ATTRIBUTES = new AttributeDefinition[] {VALUE, PARSE_TIME};
```
bstansberry: A reason to drop this attribute and always try to set as David suggested is it's really only valid for standalone servers, not domain servers. For domain it's cruft.

bstansberry:
Can ExtensionXml allow a null DeferredExtensionContext? If it's null the parsing method creates a local one and then calls load() on it before returning.

That saves all the passing around of an irrelevant object for the DomainXml, HostXml and older StandaloneXml variants, plus we don't have to coordinate getting appclient in full WF changed at the same time we merge this.

stuartwdouglas:
Retest this please

server/src/main/resources/schema/wildfly-config_6_0.xsd
```diff
@@ -130,7 +130,7 @@
         <xs:complexType>
             <xs:sequence>
                 <xs:element name="extensions" type="extensionsType" minOccurs="0" maxOccurs="1"/>
-                <xs:element name="system-properties" type="properties" minOccurs="0" maxOccurs="1"/>
+                <xs:element name="system-properties" type="properties-with-parse-time" minOccurs="0" maxOccurs="1"/>
```
bstansberry: The changes to this file should be dropped.

PR #3029, opened 2018-01-10
[WFCORE-3502] Fix debug options in test-runner

marekkopecky:
Upstream jira: https://issues.jboss.org/browse/WFCORE-3502

---

Test-runner doesn't handle debug options correctly.

If jdpa profile is used, surefire jvm uses this JVM attribute: -agentlib:jdwp=transport=dt_socket,address=${as.debug.port},server=y,suspend=y.

The same attribute is passed to started server. But server debugging should be set by wildfly.debug and wildfly.debug.port java properties. So if user wants to debug surefire with 8787 port and server with 8786 port, user uses jdpa profile, wildfly.debug property and wildfly.debug.port property and test fails:

mvn test -DfailIfNoTests=false -DtestLogToFile=false -Dtest=CdTestCase -Djpda -Dwildfly.debug=true -Dwildfly.debug.port=8786

```
ERROR: Cannot load this JVM TI agent twice, check your java command line for duplicate jdwp options.
--
Error occurred during initialization of VM
agent library failed to init: jdwp
Running null
```

### DIFF
--- a/testsuite/test-runner/src/main/java/org/wildfly/core/testrunner/Server.java
+++ b/testsuite/test-runner/src/main/java/org/wildfly/core/testrunner/Server.java
@@ -22,6 +22,8 @@ import java.util.concurrent.ExecutionException;
 import java.util.concurrent.TimeUnit;
 import java.util.concurrent.atomic.AtomicReference;
 import java.util.regex.Pattern;
+import java.util.stream.Collectors;
+import java.util.stream.Stream;
 
 import org.jboss.as.controller.client.ModelControllerClient;
 import org.jboss.as.controller.client.ModelControllerClientConfiguration;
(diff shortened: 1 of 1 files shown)


marekkopecky:
retest this please

marekkopecky:
Test errors are irrelevant with PR. @ctomc please review

testsuite/test-runner/src/main/java/org/wildfly/core/testrunner/Server.java
```diff
@@ -131,7 +133,7 @@ protected void start(PrintStream out) {
 
             commandBuilder.setJavaHome(legacyJavaHome == null ? javaHome : legacyJavaHome);
             if (jvmArgs != null) {
-                commandBuilder.setJavaOptions(jvmArgs.split("\\s+"));
+                commandBuilder.setJavaOptions(Stream.of(jvmArgs.split("\\s+")).filter(s -> !s.contains("agentlib:jdwp")).collect(Collectors.toList()));
```
bstansberry: Please use a for loop.

This is only a test fixture but I don't want Stream API in our code if it's not required. It uses a lot more metaspace than a simple for loop and our code should send a consistent message that we strive to be optimal about metaspace.

@ctomc @jamezp please review in terms of the functional change.

jamezp:
I suppose since it's just the test server it's not a bit deal, but I don't really like filtering out command line parameters like this. It could likely be worked around in the POM's themselves. The main issue is the test server uses the `jvm.args` system property which the `surefire.system.args` are added to.

I'd personally rather see this fixed in the POM, but I don't have a strong opinion either way.

marekkopecky:
Thank you for the review. I will try to fix this issue by pom approach in other PR, I close this one ...

PR #3030, opened 2018-01-11
[WFCORE-3073] Invoke the SuspendController if the BootstrapImpl shutd…

bstansberry:
…own hook runs when the server is in ControlledProcessState.State.RUNNING

https://issues.jboss.org/browse/WFCORE-3073

https://github.com/wildfly/wildfly/pull/10786 is the corresponding doc update.

Update: We've gotten the ok from Jan Stourac that this is good to go from the testing POV.

### DIFF
--- a/server/src/main/java/org/jboss/as/server/BootstrapImpl.java
+++ b/server/src/main/java/org/jboss/as/server/BootstrapImpl.java
@@ -33,6 +33,7 @@ import org.jboss.as.controller.ControlledProcessState;
 import org.jboss.as.controller.ControlledProcessStateService;
 import org.jboss.as.server.jmx.RunningStateJmx;
 import org.jboss.as.server.logging.ServerLogger;
+import org.jboss.as.server.suspend.OperationListener;
 import org.jboss.as.server.suspend.SuspendController;
 import org.jboss.modules.Module;
 import org.jboss.modules.ModuleIdentifier;
@@ -55,6 +56,8 @@ import org.wildfly.security.manager.WildFlySecurityManager;
  * @author <a href="mailto:david.lloyd@redhat.com">David M. Lloyd</a>
  */
 final class BootstrapImpl implements Bootstrap {
+
+    private static final String SIGTERM_SUSPEND_TIMEOUT_PROP = "org.wildfly.sigterm.suspend.timeout";
 
     private static final int MAX_THREADS = ServerEnvironment.getBootstrapMaxThreads();
     private final ShutdownHook shutdownHook;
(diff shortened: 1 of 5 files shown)


bstansberry:
I put a hold label on this pending completion of QE review of https://developer.jboss.org/docs/DOC-56032 but I think it's good to go otherwise, assuming CI agrees.

bstansberry:
Apparently CI disagrees! These tests pass locally. Oh, well.

I'm going to rebase this but I don't expect that to help.

bstansberry:
retest this please

testsuite/manualmode/src/test/java/org/jboss/as/test/manualmode/suspend/SuspendOnSoftKillTestCase.java
```diff
@@ -0,0 +1,414 @@
...
+/**
+ * WFCORE-3073. Tests that a softly killed process shuts down gracefully.
+ *
+ * @author Brian Stansberry
+ */
+@RunWith(WildflyTestRunner.class)
+@ServerControl(manual = true)
+public class SuspendOnSoftKillTestCase {
+
+    private static final int TIMEOUT = 20000;
```
jmesnil: Is this value correct?
I suppose you mean 20000ms (20 seconds) but the code in `BootstrapImpl` will compute it as 20000 seconds

server/src/main/java/org/jboss/as/server/BootstrapImpl.java
```diff
@@ -274,5 +282,82 @@ public void handleTermination(Info info) {
...
+            if (sc != null && !sc.isShutdownComplete()) {
+                final ServiceController serviceController = sc.getService(SuspendController.SERVICE_NAME);
+                if (serviceController != null && serviceController.getState() == ServiceController.State.UP) {
+                    result = (SuspendController) serviceController.getValue();
+                }
+            }
+            return result;
+        }
+
+        private static long getSuspendTimeout() {
```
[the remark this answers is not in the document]
jmesnil: sorry, I did not pay attention to the commit message. It is documented in https://github.com/wildfly/wildfly/pull/10786/files#diff-6b64f216bcc12e34890b915ae0e10dbdR198

testsuite/manualmode/src/test/java/org/jboss/as/test/manualmode/suspend/SuspendOnSoftKillTestCase.java
bstansberry: @jmesnil This value isn't passed to the server for use by BootstrapImpl; it's just the # of ms used in waitForServerShutdown() when the test is waiting for the server to completely stop after the test of suspend behavior is done.

I'll clean this up a bit; get rid of the constant so it's less confusing, use TimeoutUtil more consistently etc.

testsuite/manualmode/src/test/java/org/jboss/as/test/manualmode/suspend/SuspendOnSoftKillTestCase.java
```diff
@@ -0,0 +1,414 @@
...
+
+            // Send a request that will block
+            Future<Object> result = executorService.submit(new Callable<Object>() {
+                @Override
+                public Object call() throws Exception {
+                    return HttpRequest.get(address, 60, TimeUnit.SECONDS);
+                }
+            });
+
+            Thread.sleep(1000); //nasty, but we need to make sure the HTTP request has started
```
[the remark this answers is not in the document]
bstansberry: Yes.

bstansberry:
@jmesnil I updated this. I figure the latest failures on full are not relevant as we're not using sigterm there.

server/src/main/java/org/jboss/as/server/BootstrapImpl.java
```diff
@@ -274,5 +282,82 @@ public void handleTermination(Info info) {
...
+            long result = 0;
+            String timeoutString = System.getProperty(SIGTERM_SUSPEND_TIMEOUT_PROP);
+            if (timeoutString != null && timeoutString.length() > 0) {
+                try {
+                    int max = Integer.decode(timeoutString);
+                    result = max > 0 ? max * 1000 : max;
```
rsvoboda: Should we care about overflow?  Cast either max or 1000 to type long as `result` is long

PR #3033, opened 2018-01-12
Fix for WFCORE-3507, use arrow Down key to scroll

jfdenise:
Arrow down key to scroll one line. Updated analysis document, manual testing.


jfdenise:
@bstansberry, we already discussed this change with QE, we agreed that it could be added. We will see if we have other reactions.

bstansberry:
@jfdenise Great. Next time I do some merging I'll include this. In the meantime if you hear any concerns, please leave a comment here.

Anyone else doing merging, this is good to go.

PR #3035, opened 2018-01-15
[WFCORE-3457] Make 'TrivialResourceDefinition' final and use a Builder instead of extending each time.

darranl:
Also update the XML reader and writer to be supplied using Lambda expressions, this avoids those classes being loaded when not used i.e. subsystem extension installed but no subsystem in config or domain mode servers.

### DIFF
--- a/elytron/src/main/java/org/wildfly/extension/elytron/ElytronExtension.java
+++ b/elytron/src/main/java/org/wildfly/extension/elytron/ElytronExtension.java
@@ -96,10 +96,10 @@ public class ElytronExtension implements Extension {
 
     @Override
     public void initializeParsers(ExtensionParsingContext context) {
-        context.setSubsystemXmlMapping(SUBSYSTEM_NAME, NAMESPACE_1_0, ElytronSubsystemParser1_0::new);
-        context.setSubsystemXmlMapping(SUBSYSTEM_NAME, NAMESPACE_1_1, ElytronSubsystemParser1_1::new);
-        context.setSubsystemXmlMapping(SUBSYSTEM_NAME, NAMESPACE_1_2, ElytronSubsystemParser1_2::new);
-        context.setSubsystemXmlMapping(SUBSYSTEM_NAME, NAMESPACE_2_0, ElytronSubsystemParser2_0::new);
+        context.setSubsystemXmlMapping(SUBSYSTEM_NAME, NAMESPACE_1_0, () -> new ElytronSubsystemParser1_0());
+        context.setSubsystemXmlMapping(SUBSYSTEM_NAME, NAMESPACE_1_1, () -> new ElytronSubsystemParser1_1());
+        context.setSubsystemXmlMapping(SUBSYSTEM_NAME, NAMESPACE_1_2, () -> new ElytronSubsystemParser1_2());
+        context.setSubsystemXmlMapping(SUBSYSTEM_NAME, NAMESPACE_2_0, () -> new ElytronSubsystemParser2_0());
     }
 
     @Override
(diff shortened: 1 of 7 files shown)


elytron/src/main/java/org/wildfly/extension/elytron/ElytronExtension.java
```diff
@@ -106,10 +106,10 @@ static boolean isServerOrHostController(ImmutableManagementResourceRegistration
...
     @Override
     public void initializeParsers(ExtensionParsingContext context) {
-        context.setSubsystemXmlMapping(SUBSYSTEM_NAME, NAMESPACE_1_0, ElytronSubsystemParser1_0::new);
-        context.setSubsystemXmlMapping(SUBSYSTEM_NAME, NAMESPACE_1_1, ElytronSubsystemParser1_1::new);
-        context.setSubsystemXmlMapping(SUBSYSTEM_NAME, NAMESPACE_1_2, ElytronSubsystemParser1_2::new);
-        context.setSubsystemXmlMapping(SUBSYSTEM_NAME, NAMESPACE_2_0, ElytronSubsystemParser2_0::new);
+        context.setSubsystemXmlMapping(SUBSYSTEM_NAME, NAMESPACE_1_0, () -> new ElytronSubsystemParser1_0());
+        context.setSubsystemXmlMapping(SUBSYSTEM_NAME, NAMESPACE_1_1, () -> new ElytronSubsystemParser1_1());
+        context.setSubsystemXmlMapping(SUBSYSTEM_NAME, NAMESPACE_1_2, () -> new ElytronSubsystemParser1_2());
+        context.setSubsystemXmlMapping(SUBSYSTEM_NAME, NAMESPACE_2_0, () -> new ElytronSubsystemParser2_0());
```
bstansberry: For the current one please pass in the constructed instance, not a supplier. It helps optimize parallel boot in black magic ways.

darranl: Ok, I will revert the others back to method references as well.

bstansberry: @darranl That's not what I meant. :) I meant just the one at L112 should be

```
context.setSubsystemXmlMapping(SUBSYSTEM_NAME, NAMESPACE_2_0, new ElytronSubsystemParser2_0());
```

We don't want to defer construction of the current parser, as it results in deferring a lot of classloading and static management API initialization out of the concurrent extension initialization part of boot and into the single threaded parsing part. There's black magic to avoid that but it doesn't work if the variant that takes a lambda/method ref/other form of supplier is used.  Just the current parser is enough to get the benefit.

darranl: I need to double check something else my own figures don't quite show yet but once you instantiate the most recent one the chain causes all the previous ones to load regardless.  

As the classes are now all loaded I believe the method references start to come out as having the lower metaspace overhead. 

A couple of things that do spring to mind when looking at this area of subsystems: -
 # I think it would help us greatly if we could make a small change to the APIs to skip these registrations entirely in processes that will not use them.
 # I believe you have parallel boot issues that may negate this but registering a factory with a list of supported namespaces seems more optimal compared to one Supplier per namespace.  Maybe parallel boot could look after it's own optimisation and request the most recent one at the correct stage.

darranl: LOL GitHub went all shouty that wasn't me ;-)

bstansberry: We'll instantiate the most recent one regardless, assuming the subsystem is present and uses the latest schema. So if the classes are a chain, then yeah, they'll be loaded.

We could skip the initializeParsers call in a domain server pretty easily. It's more problematic on an HC, as whether a given HC will need to parse depends on a number of factors. I doubt the HC one is worth it as our big metaspace concern is servers.

For parallel boot to request the most recent one, it needs to identify most recent and right now that means lexical comparison of namespaces and hoping that maps to "recent".

That might actually work ok as a fallback.

bstansberry: #3049 attempts to load the "latest".

I also filed https://issues.jboss.org/browse/WFCORE-3523 re: domain servers, but decided not to pursue it.

bstansberry:
retest this please

PR #3036, opened 2018-01-16
WFCORE-3510 Build tools 1.2.6.Final

stuartwdouglas:


stuartwdouglas:
retest this please

dmlloyd:
Is that the right JIRA #?  It references something about boot context.

dmlloyd:
But it's not clear what was fixed here from the JIRA # at any rate...

bstansberry:
@dmlloyd https://github.com/wildfly/wildfly-build-tools/compare/1.2.5.Final...1.2.6.Final

stuartwdouglas:
The main one is the support for hard coded resources, as it allows us to create feature packs with different versions of the same jar for EE8.

PR #3038, opened 2018-01-16
WFLY-9680 Fixing bug in handling default methods in EJB interfaces.

dbildhauer:
Fixing WFLY-9680. Prior to this commit, if a method defined on an remote interface A is missing in a EJB class B implementing the interface A but a default method is provided the declaring interface, the method will get resolved successfully. If there is a interface C extending A and providing the default implementation and B implements C, the call fails with a message "Invocation cannot proceed (end of interceptor chain has been hit)". This is fixed by this commit.

### DIFF
--- a/server/src/main/java/org/jboss/as/server/deployment/reflect/ClassReflectionIndex.java
+++ b/server/src/main/java/org/jboss/as/server/deployment/reflect/ClassReflectionIndex.java
@@ -64,6 +64,9 @@ public final class ClassReflectionIndex {
      */
     private volatile Set<Method> classMethods;
 
+    /* Map of all methods defined by this class and its superclasses (including default methods) */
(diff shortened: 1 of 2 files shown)


bstansberry:
this is ok to test

stuartwdouglas:
Thanks for the PR, but I think a more correct solution would be https://github.com/wildfly/wildfly-core/pull/3040 . The existing findMethod implementation should return the method if it is defaulted or not, so I don't think it is necessary to add a new method.

dbildhauer:
Hi,

I just introduced a new method because I didn't want to break some existing behaviour at other places using the method. But I agree, more correct would be a adaption to the existing findMethod method.

stuartwdouglas:
I think that it is better to change the existing method, as anything that uses it should also find defaulted methods.

PR #3039, opened 2018-01-16
Work around WFCORE-3302 / REM3-303 by sleeping

bstansberry:
Really ugly but timing on ci.wildfly.org seems to have changed recently such that these failures are pretty common again, plus they've always been common on brontes and there's a Critical JBEAP to make that go away. The failures are common enough that I can't really argue with the Critical designation.

Workaround for https://issues.jboss.org/browse/WFCORE-3302.  Doesn't really resolve it, although it's ok to resolve the issue since we have REM3-303 for the basic problem.


bstansberry:
retest this please

PR #3040, opened 2018-01-17
WFCORE-3512 ClassReflectionIndexUtil does not return default methods …

stuartwdouglas:
…from findMethod


bstansberry:
How can this be tested?

stuartwdouglas:
I have added a test in Wildfly. In theory all this reflection stuff could probably be moved to the EE module, as it is not really used directly by core, however that could break other products if they are relying on it.

PR #3046, opened 2018-01-18
[WFCORE-3521]: Display dynamic capabilities with a pattern.

ehsavoie:
Jira: https://issues.jboss.org/browse/WFCORE-3521

This is helpful for provisioning

### DIFF
--- a/controller/src/main/java/org/jboss/as/controller/AttributeDefinition.java
+++ b/controller/src/main/java/org/jboss/as/controller/AttributeDefinition.java
@@ -982,8 +982,15 @@ public abstract class AttributeDefinition {
                 result.get(ModelDescriptionConstants.REQUIRES).add(required);
             }
         }
+
         if (getReferenceRecorder() != null) {
+            String[] capabilityPatternElements = getReferenceRecorder().getRequirementPatternSegments(name);
             result.get(ModelDescriptionConstants.CAPABILITY_REFERENCE).set(getReferenceRecorder().getBaseRequirementName());
+            if(capabilityPatternElements.length > 0 && (capabilityPatternElements.length > 1 || !name.equals(capabilityPatternElements[0]))) {
+                for(String patternElement : capabilityPatternElements) {
+                    result.get(ModelDescriptionConstants.CAPABILITY_REFERENCE_PATTERN_ELEMENTS).add(patternElement);
+                }
+            }
         }
 
         if (validator instanceof MinMaxValidator) {
--- a/controller/src/main/java/org/jboss/as/controller/CapabilityReferenceRecorder.java
+++ b/controller/src/main/java/org/jboss/as/controller/CapabilityReferenceRecorder.java
@@ -22,6 +22,8 @@
 
 package org.jboss.as.controller;
 
+import java.util.ArrayList;
+import java.util.List;
 import java.util.Set;
 
 import org.jboss.as.controller.capability.RuntimeCapability;
@@ -35,7 +37,6 @@ import org.jboss.dmr.ModelNode;
  * @author Brian Stansberry (c) 2015 Red Hat Inc.
  */
 public interface CapabilityReferenceRecorder {
-
     /**
      * Registers capability requirement information to the given context.
      * @param context         the context
@@ -74,6 +75,20 @@ public interface CapabilityReferenceRecorder {
     @Deprecated
     default boolean isDynamicDependent() {
         throw new UnsupportedOperationException();
+    }
+
+    /**
+     * Returns the elements to be added to the baseRequirementName to build the capability name pattern.
+     * It will return an array  of the form `segment[.segment]` where each segment represents either the name of one of the
+     * resource's attributes or one of the keys in the resource's address. In the actual name the attribute name
+     * or address key will be replaced by the value associated with that attribute or key.
+     * @return the elements to be added to the baseRequirementName to build the capability name pattern.
+     */
+    default String[] getRequirementPatternSegments(String name) {
+        if (name != null && !name.isEmpty()) {
+            return new String[]{name};
+        }
+        return null;
     }
 
     /**
@@ -149,7 +164,6 @@ public interface CapabilityReferenceRecorder {
         public String getBaseRequirementName() {
             return baseRequirementName;
         }
-
     }
 
     /**
@@ -214,10 +228,10 @@ public interface CapabilityReferenceRecorder {
         String getDependentName(RuntimeCapability cap, OperationContext context) {
             if (cap.isDynamicallyNamed()) {
                 return cap.fromBaseCapability(context.getCurrentAddress()).getName();
-            } else {
-                return cap.getName();
             }
+            return cap.getName();
         }
+
         protected String getRequirementName(OperationContext context, Resource resource, String attributeValue){
             return RuntimeCapability.buildDynamicCapabilityName(baseRequirementName, attributeValue);
         }
@@ -245,7 +259,6 @@ public interface CapabilityReferenceRecorder {
      * that do not meet this requirement.</strong>
      */
     class CompositeAttributeDependencyRecorder extends ContextDependencyRecorder {
-
         private AttributeDefinition[] attributes;
         private RuntimeCapability capability;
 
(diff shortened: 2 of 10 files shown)


ehsavoie:
retest this please

ehsavoie:
retest this pease

controller/src/main/java/org/jboss/as/controller/CapabilityReferenceRecorder.java
```diff
@@ -215,8 +229,8 @@ String getDependentName(RuntimeCapability cap, OperationContext context) {
             if (cap.isDynamicallyNamed()) {
                 return cap.fromBaseCapability(context.getCurrentAddress()).getName();
             } else {
-                return cap.getName();
-            }
+            return cap.getName();
+        }
```
bstansberry: This whitespace change looks incorrect.

controller/src/main/java/org/jboss/as/controller/descriptions/DefaultResourceDescriptionProvider.java
```diff
@@ -167,4 +172,17 @@ private ModelNode addDeprecatedInfo(final ModelNode model) {
...
+        while (iter.hasNext()) {
+            PathElement elt = iter.next();
+            if (elt.isWildcard()) {
+                aliasAddress = aliasAddress.append(PathElement.pathElement(elt.getKey(), "$" + elt.getKey()));
+            } else {
+                aliasAddress = aliasAddress.append(elt);
+            }
+        }
+        return aliasAddress;
+    }
```
bstansberry: Please document this. It's quite clever. :) But the cleverness makes it confusing.

controller/src/main/java/org/jboss/as/controller/CapabilityReferenceRecorder.java
```diff
@@ -76,6 +76,21 @@ default boolean isDynamicDependent() {
...
+     * dynamicElements the dynmaic elements of the requirement capability name.
+     * @return requirement name of the capability this reference depends on
+     */
+    default String getRequirementName(String... dynamicElements) {
+        if (dynamicElements != null && dynamicElements.length > 0) {
+            String[] dynamicParts = new String[dynamicElements.length];
+            for (int i = 0; i < dynamicElements.length; i++) {
+                dynamicParts[i] = "$" + dynamicElements[i];
+            }
+            return RuntimeCapability.buildDynamicCapabilityName(getBaseRequirementName(), dynamicParts);
```
bstansberry: It's better to not repeat the static part. We've gotten complaints before about the size of our metadata.

bstansberry:
https://github.com/wildfly/wildfly/compare/master...bstansberry:cap-req-doc is a draft doc with possible names for this metadata.

Also, we should avoid outputting this in the standard case, again to cut the size of our descriptions.

Perhaps the capability ref recorder should return List<String> instead of String and then the consumer deals with the $ part. That way if you pass in 'foo' and it replies with [`foo`] it can efficiently detect that and skip the output.

controller/src/main/java/org/jboss/as/controller/CapabilityReferenceRecorder.java
```diff
@@ -76,6 +76,21 @@ default boolean isDynamicDependent() {
         throw new UnsupportedOperationException();
     }
 
+    /**
+     * dynamicElements the dynmaic elements of the requirement capability name.
+     * @return requirement name of the capability this reference depends on
+     */
+    default String getRequirementName(String... dynamicElements) {
```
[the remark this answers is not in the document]
ehsavoie: i wasn't sure at first, but this is not required

controller/src/main/java/org/jboss/as/controller/CapabilityReferenceRecorder.java
ehsavoie: ok

controller/src/main/java/org/jboss/as/controller/descriptions/DefaultResourceDescriptionProvider.java
```diff
@@ -103,6 +105,9 @@ public ModelNode getModelDescription(Locale locale) {
             for (Capability capability: capabilities) {
                 ModelNode cap = result.get(ModelDescriptionConstants.CAPABILITIES).add();
                 cap.get(NAME).set(capability.getName());
+                if(capability.isDynamicallyNamed()) {
+                    cap.get(DYNAMIC_NAME).set(capability.getDynamicName(createAliasPathAddress(pa)));
+                }
```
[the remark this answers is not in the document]
ehsavoie: ok

bstansberry:
retest this please

ehsavoie:
retest this please

ehsavoie:
retest this please

ehsavoie:
retest this please

ehsavoie:
retest this please

ehsavoie:
retest this please

controller/src/main/java/org/jboss/as/controller/CapabilityReferenceRecorder.java
```diff
@@ -76,6 +77,23 @@ default boolean isDynamicDependent() {
         throw new UnsupportedOperationException();
     }
 
+    /**
+     * Returns the elements to be added to the baseRequirementName to build the capability name pattern.
+     * The pattern is of the form `$segment[.$segment]` where each segment represents either the name of one of the
```
bstansberry: Not: The form is now `segment[.segment]`. The RRD output is now a LIST of STRING with the $ char no longer relevant.

controller/src/main/java/org/jboss/as/controller/AttributeDefinition.java
```diff
@@ -982,8 +982,15 @@ public ModelNode getNoTextDescription(boolean forOperation) {
                 result.get(ModelDescriptionConstants.REQUIRES).add(required);
             }
         }
+
         if (getReferenceRecorder() != null) {
+            String[] capabilityPatternElements = getReferenceRecorder().getRequirementPatternSegments(name);
             result.get(ModelDescriptionConstants.CAPABILITY_REFERENCE).set(getReferenceRecorder().getBaseRequirementName());
+            if(capabilityPatternElements.length > 0) {
```
bstansberry:
```
if(capabilityPatternElements.length > 0 && (capabilityPatternElements.length > 1 || !name.equals(capabilityPatternElements[0])) { 
```

IOW don't output this for the default case.

If not outputting this makes the feature-spec generator too difficult let me know.

controller/src/main/java/org/jboss/as/controller/descriptions/DefaultResourceDescriptionProvider.java
```diff
@@ -99,11 +104,18 @@ public ModelNode getModelDescription(Locale locale) {
...
         Set<? extends Capability> capabilities = registration.getCapabilities();
-        if (capabilities!=null&&!capabilities.isEmpty()){
+        if (capabilities !=null && !capabilities.isEmpty()){
             for (Capability capability: capabilities) {
                 ModelNode cap = result.get(ModelDescriptionConstants.CAPABILITIES).add();
                 cap.get(NAME).set(capability.getName());
                 cap.get(DYNAMIC).set(capability.isDynamicallyNamed());
+                if(capability.isDynamicallyNamed()) {
+                    String dynamicPart = capability.getDynamicName(createAliasPathAddress(pa)).substring(capability.getName().length());
+                    StringTokenizer tokenizer = new StringTokenizer(dynamicPart, ".$", false);
```
bstansberry: If you use String.split you can get an array and again test for the case of a 1 element array whose element is the same as pa.getLastElement().getKey(). Don't output more description data for the default case.

ehsavoie:
retest this please

controller/src/main/java/org/jboss/as/controller/descriptions/DefaultResourceDescriptionProvider.java
ehsavoie: Looks like it is not breaking my usage

PR #3047, opened 2018-01-19
WFCORE-3525 Expose Xnio buffer pools as an Undertow ByteBufferPool ca…

stuartwdouglas:
…pability for compatibility

This needs upstream https://github.com/wildfly/wildfly/pull/10768 to be merged first, otherwise the removal of the default buffer pool will break the build.

### DIFF
--- a/io/subsystem/src/main/java/org/wildfly/extension/io/BufferPoolResourceDefinition.java
+++ b/io/subsystem/src/main/java/org/wildfly/extension/io/BufferPoolResourceDefinition.java
@@ -31,6 +31,7 @@ import java.util.List;
 
 import org.jboss.as.controller.AbstractAddStepHandler;
 import org.jboss.as.controller.AttributeDefinition;
+import org.jboss.as.controller.ModelVersion;
 import org.jboss.as.controller.OperationContext;
 import org.jboss.as.controller.OperationFailedException;
 import org.jboss.as.controller.PathAddress;
@@ -44,8 +45,16 @@ import org.jboss.as.controller.registry.AttributeAccess;
 import org.jboss.as.controller.registry.ManagementResourceRegistration;
 import org.jboss.dmr.ModelNode;
 import org.jboss.dmr.ModelType;
+import org.jboss.msc.service.Service;
 import org.jboss.msc.service.ServiceController;
+import org.jboss.msc.service.StartContext;
+import org.jboss.msc.service.StartException;
+import org.jboss.msc.service.StopContext;
+import org.jboss.msc.value.InjectedValue;
 import org.xnio.Pool;
+
+import io.undertow.connector.ByteBufferPool;
+import io.undertow.server.XnioByteBufferPool;
 
 /**
  * @author <a href="mailto:tomaz.cerar@redhat.com">Tomaz Cerar</a> (c) 2013 Red Hat Inc.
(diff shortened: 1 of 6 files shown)


io/subsystem/src/main/java/org/wildfly/extension/io/IOExtension.java
```diff
@@ -69,7 +69,7 @@ public void initializeParsers(ExtensionParsingContext context) {
 
     @Override
     public void initialize(ExtensionContext context) {
-        final SubsystemRegistration subsystem = context.registerSubsystem(SUBSYSTEM_NAME, ModelVersion.create(4));
+        final SubsystemRegistration subsystem = context.registerSubsystem(SUBSYSTEM_NAME, ModelVersion.create(4, 1));
```
stuartwdouglas: Do I need to up the model version because I have deprecated a resource? I am assuming yes, although I am not sure if it should be major minor or micro.

bstansberry: @stuartwdouglas It was already bumped for WF Core 4 as part of https://issues.jboss.org/browse/WFCORE-3333

io/subsystem/src/main/java/org/wildfly/extension/io/BufferPoolResourceDefinition.java
```diff
@@ -151,6 +165,35 @@ protected void performRuntime(OperationContext context, ModelNode operation, Mod
                     .setInitialMode(ServiceController.Mode.ACTIVE)
                     .install();
 
+            ByteBufferPoolService poolService = new ByteBufferPoolService();
+
+            context.getCapabilityServiceTarget().addCapability(IO_BYTE_BUFFER_POOL_RUNTIME_CAPABILITY, poolService)
+                    .addCapabilityRequirement(IO_POOL_RUNTIME_CAPABILITY.getDynamicName(address), Pool.class, poolService.bufferPool)
+                    .setInitialMode(ServiceController.Mode.ACTIVE)
```
bstansberry: Should this be ON_DEMAND? Same question for the existing one.

io/subsystem/src/main/java/org/wildfly/extension/io/IOExtension.java
ctomc: In this case it should stay at 4.0.
unless it doesn't get merged for wildfly-core 4.0, in that case it needs to be 5.0 (and few other PRs as well)

stuartwdouglas:
Turns out this is actually needed by the upstream bit first. I have left the buffer pool in the IO subsystem for now, and will remove it in a later PR once the downstream PR is merged.

io/subsystem/src/main/java/org/wildfly/extension/io/BufferPoolResourceDefinition.java
```diff
@@ -126,6 +139,7 @@ private BufferPoolResourceDefinition() {
     @Override
     public void registerCapabilities(ManagementResourceRegistration resourceRegistration) {
         resourceRegistration.registerCapability(IO_POOL_RUNTIME_CAPABILITY);
+        resourceRegistration.registerCapability(IO_BYTE_BUFFER_POOL_RUNTIME_CAPABILITY);
```
jmesnil: this does not prevent to merge this PR but this capability needs to be documented in https://github.com/wildfly/wildfly-capabilities

jmesnil:
@bstansberry fiy, I've merged this PR as stuart as modified it according to your requests

PR #3048, opened 2018-01-19
[WFCORE-3522] Deployment module content comes before dependencies if isLocalLast is false

dmlloyd:


dmlloyd:
Discussion:

> [8:57 AM] David Lloyd: the problem that arises is that even if "isLocalLast" is false, "dependencies" and "userDependencies" still come before local classes
[8:57 AM] David Lloyd: the reason for doing this is presumably to prevent user JARs from overwriting dependency JARs (classic parent-first fallacy)
[8:59 AM] David Lloyd: I think that if "isLocalLast" is false then we should put the createLocalDependencySpec() *first* before everything else
[9:00 AM] David Lloyd: the potential problem is user deployments that will break if they had included a wrong dependency
[9:01 AM] David Lloyd: but I'm tempted to take the Linus Torvalds approach: just fix it, and if anyone really wants to fix the wrong dependency problem, they will be motivated to do it then :)
[9:01 AM] David Lloyd: TBH we've needed a better solution to that for years now

ropalka:
Looks good to me.

dmlloyd:
It looks like there are numerous tests which use shrinkwrap and which duplicate classes all over the place.

dmlloyd:
Looks like this solution is too simplistic and doesn't account for compatibility with existing cases.

PR #3049, opened 2018-01-19
[WFCORE-3524] If only parser suppliers are registered, initialize the…

bstansberry:
… one with the 'latest' namespace

https://issues.jboss.org/browse/WFCORE-3524

See discussion on #3035

### DIFF
--- a/controller/src/main/java/org/jboss/as/controller/extension/ExtensionRegistry.java
+++ b/controller/src/main/java/org/jboss/as/controller/extension/ExtensionRegistry.java
@@ -40,6 +40,7 @@ import java.util.function.Supplier;
 import javax.xml.namespace.QName;
 
 import org.jboss.as.controller.AttributeDefinition;
+import org.jboss.as.controller.Extension;
 import org.jboss.as.controller.ExtensionContext;
 import org.jboss.as.controller.ModelVersion;
 import org.jboss.as.controller.ModelVersionRange;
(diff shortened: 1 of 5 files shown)


controller/src/main/java/org/jboss/as/controller/extension/ExtensionRegistry.java
```diff
@@ -463,6 +477,12 @@ public void setSubsystemXmlMapping(String subsystemName, String namespaceUri, Su
                 extension.getSubsystemInfo(subsystemName).addParsingNamespace(namespaceUri);
                 if (extension.xmlMapper != null) {
                     extension.xmlMapper.registerRootElement(new QName(namespaceUri, SUBSYSTEM), supplier);
+                    if (!hasNonSupplierParser) {
+                        if (latestNamespace == null || latestNamespace.compareTo(namespaceUri) < 0) {
```
jmesnil: You are using lexicographical order to determine the latest namespace.
However we should in a future API, make the order of XML (and management model version) more explicit or let the subsystem developer specifies among the list of supported XML namespace which one is the latest.

dmlloyd: Lexicographical order will fail on a transition from e.g. ``xxx:1.9`` to ``xxx:1.10``.

Ideally we'd map XML namespaces to model versions (``org.jboss.as.controller.ModelVersion``?  Or perhaps ``org.jboss.modules.Version`` or something like it...).  Then earlier-later comparison will generally work correctly.

bstansberry: @dmlloyd @jmesnil This is definitely a hack just until we write a new extension API.  I don't believe we have any versions close to xxx:y.9, so the hack should be ok for a while. Also, the cost if this hack is wrong is just that we waste some resources; it doesn't produce incorrect behavior.

https://github.com/bstansberry/wildfly-core/blob/new-mgmt-api/management-api/src/main/java/org/wildfly/management/api/SchemaVersion.java is meant to capture the relationship between a management API version and the related xml namespaces, and the branching relationship between those tuples. Of course the details can all change but it's definitely the idea that it be explicit.

jmesnil: I'm fine with having this optimization in place for our current API (with caveats that a proper schema version will be introduced for our supported extension API)

PR #3050, opened 2018-01-19
[WFCORE-3527] OSGi cleanup

ropalka:


bstansberry:
@stuartwdouglas wdyt about this one?  Ripping out some undeprecated public API in Attachments and DeploymentUtil is a tiny red flag, but in reality I figure no one is using this stuff.  If it wouldn't work anymore anyway (possible if it's never called) then there isn't even a tiny red flag.

ropalka:
@bstansberry @stuartwdouglas I just sent PR to wildfly-camel to address binary compatibility problem

bstansberry:
In that case I'm going to revert this.

bstansberry:
@ropalka Since it's 3 commits it's too much work for me to revert and I guess since the branch is deleted the normal button on this page to revert is gone. Please send a revert PR that eliminates the breakage.

ropalka:
Here is it @bstansberry https://github.com/wildfly/wildfly-core/pull/3060

PR #2480, opened 2017-05-29
[WFCORE-2879][JBEAP-11181] key-password-credential-reference fix

honza889:
https://issues.jboss.org/browse/WFCORE-2879
https://issues.jboss.org/browse/JBEAP-11181

PR #3056, opened 2018-01-22
[WFCORE-2902] elytron cleanup: replacing as*IfDefined util by as*OrNull

hkalina:
Refactoring: removed unnecessary elytron utils, which was deprecated as suggested in
https://github.com/wildfly/wildfly-core/pull/2480#discussion_r119153768

https://issues.jboss.org/browse/WFCORE-2902


bstansberry:
this is ok to test

darranl:
@bstansberry I can review changes like this if you want but TBH if it is general management API clean up I don't think we need to hold for a security specific review unless you want to.

bstansberry:
@darranl I put the 'security' label on there just because it relates to the subsystem. It doesn't have any particular meaning to me that security expertise is required, although I can only use it that way if you prefer.

In general the first choice for reviews of a change to a subsystem are the people associated with that subsystem. This one looks fine though.

darranl:
Either way is Ok with me @bstansberry - I just know realistically if it comes to correct management API utilisation you know more ;-)

PR #3060, opened 2018-01-23
[WFCORE-3527] Removing OSGi from WildFly-Core

ropalka:


bstansberry:
@aloubyansky @jamezp please approve

Alexey the main thing I wonder here is whether in the updated patching tests some of the use of 'osgi' is important to the tests, not because the test is concerned with anything about osgi but because the test is evaluating certain structures, and 'osgi' just happened to be a name given to something that could just as well have been called 'foo'.

jamezp:
It looks like in logging we just removed the check to see if there was a module attachment. Do we really need to remove that? I do recall it only being there for OSGi, but it seems like a harmless check to leave in there.

ropalka:
Attachments.MODULE is always attached to DeploymentUnit @jamezp - there's no need for check.
In the past it wasn't always true because of OSGi integration.

kabir:
@ropalka Needs a rebase

ropalka:
Rebased @kabir

jmesnil:
@aloubyansky could you review it please?

ropalka:
Updated @wildfly/mergers-core

ropalka:
How about this PR @jmesnil . What else needs to be accomplished here?

jmesnil:
@ropalka PR looks good to me. I'm just waiting for @aloubyansky approval wrt to patching before we can merge it.

PR #3062, opened 2018-01-24
[WFCORE-3540] hardcode terminal width and height to avoid malformed t…

soul2zimate:
…able view in non interactive mode.
https://issues.jboss.org/browse/WFCORE-3540

### DIFF
--- a/cli/src/main/java/org/jboss/as/cli/impl/CommandContextImpl.java
+++ b/cli/src/main/java/org/jboss/as/cli/impl/CommandContextImpl.java
@@ -1849,8 +1849,8 @@ public class CommandContextImpl implements CommandContext, ModelControllerClient
 
     @Override
     public int getTerminalWidth() {
-        if( !INTERACT ){
-            return 0;
+        if (!INTERACT) {
+            return 80;
         }
 
         if(console == null) {
@@ -1866,8 +1866,8 @@ public class CommandContextImpl implements CommandContext, ModelControllerClient
 
     @Override
     public int getTerminalHeight() {
-        if( !INTERACT ){
-            return 0;
+        if (!INTERACT) {
+            return 24; // WFCORE-3540 24 has no special meaning except that it is a value greater than 0
         }
 
         if(console == null) {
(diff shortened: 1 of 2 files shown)


cli/src/main/java/org/jboss/as/cli/impl/CommandContextImpl.java
```diff
@@ -1650,7 +1650,7 @@ public int getTerminalWidth() {
     @Override
     public int getTerminalHeight() {
         if( !INTERACT ){
-            return 0;
+            return 40;
```
jfdenise: Strangely, in the same method when an error occurs, 24 is returned. This value seems random. That is strange to have 24 returned in one case and 40 in another.  I looked at GIT history but nothing special. You should perhaps try to replace 24 by 40, I can't see how this could have an effect.

soul2zimate:
will do.
I saw also two failures from AttachmentTestCase related to this change. I am looking into it.

soul2zimate:
Last run,  AttachmentTestCase failed due to MissingFormatWidthException from Formatter. Because SimpleTable constructor allows to create object without headers, that makes an empty columnLengths array. It was previously filled with value.length() + 1 as it allows to have terminal width 0 in tests.
Once the hardcoded values replace zero, it bypass that step and column width remains 0.

soul2zimate:
retest this please

soul2zimate:
retest this please

soul2zimate:
@jfdenise  how do you feel about the change in SimpleTable.java ?

cli/src/main/java/org/jboss/as/cli/impl/CommandContextImpl.java
```diff
@@ -1650,7 +1650,7 @@ public int getTerminalWidth() {
     @Override
     public int getTerminalHeight() {
         if( !INTERACT ){
-            return 0;
+            return 24;
```
jfdenise: Perhaps add a comment saying that 24 has no special meaning except that it is a value greater than 0.

soul2zimate:
test failure is unrelated.

PR #2715, opened 2017-08-16
[WFCORE-3187] add-user.sh tool prints unexpected characters if arrow keys are pushed during interactive mode

MMarus:
JBEAP: https://issues.jboss.org/browse/JBEAP-6640
WFCORE: https://issues.jboss.org/browse/WFCORE-3187

Please DON'T merge before https://github.com/wildfly/wildfly-core/pull/2533 is merged, since this pull request requires changes introduced in https://github.com/wildfly/wildfly-core/pull/2533 .

PR #3064, opened 2018-01-25
[WFCORE-3187] add-user.sh tool prints unexpected characters if arrow …

MMarus:
…keys are pushed during interactive mode

Moved https://github.com/wildfly/wildfly-core/pull/2715 to a new rebased PR for easier review.

JBEAP: https://issues.jboss.org/browse/JBEAP-6640
WFCORE: https://issues.jboss.org/browse/WFCORE-3187

### DIFF
--- a/domain-management/pom.xml
+++ b/domain-management/pom.xml
@@ -72,6 +72,10 @@
 
     <dependencies>
         <dependency>
+            <groupId>org.aesh</groupId>
+            <artifactId>aesh-readline</artifactId>
+        </dependency>
+        <dependency>
             <groupId>org.apache.directory.api</groupId>
             <artifactId>api-ldap-codec-standalone</artifactId>
             <scope>test</scope>
--- a/domain-management/src/main/java/org/jboss/as/domain/management/security/adduser/JavaConsole.java
+++ b/domain-management/src/main/java/org/jboss/as/domain/management/security/adduser/JavaConsole.java
@@ -28,7 +28,10 @@ import static org.jboss.as.domain.management.logging.DomainManagementLogger.ROOT
 
 import java.io.Console;
 import java.io.IOError;
+import java.io.IOException;
 import java.util.IllegalFormatException;
+import org.aesh.readline.Prompt;
+import org.aesh.readline.tty.terminal.TerminalConnection;
 
 /**
  * Describe the purpose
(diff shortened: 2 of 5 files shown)


domain-management/src/main/java/org/jboss/as/domain/management/security/adduser/JavaConsole.java
```diff
@@ -24,10 +24,14 @@
 
 package org.jboss.as.domain.management.security.adduser;
 
+import org.aesh.readline.Prompt;
```
darranl: These imports are in the wrong location, they should be after the static imports but before the java.* imports.

domain-management/src/main/java/org/jboss/as/domain/management/security/adduser/ReadLineHandler.java
```diff
@@ -0,0 +1,72 @@
...
+import org.aesh.readline.Readline;
+import org.aesh.readline.ReadlineBuilder;
+import org.aesh.terminal.Connection;
+import org.aesh.terminal.tty.Signal;
+
+/**
+ * Class to read line from terminal using Aesh-readline library.
+ * Created by Marek Marusic <mmarusic@redhat.com> on 8/7/17.
+ */
+public class ReadLineHandler implements Consumer<Connection> {
```
darranl: Is this only used by classes in this package?  If so please reduce the visibility.

domain-management/src/test/java/org/jboss/as/domain/management/security/adduser/AeshJavaConsoleTest.java
```diff
@@ -0,0 +1,134 @@
...
+import org.aesh.readline.tty.terminal.TerminalConnection;
+import org.junit.Test;
+
+import java.io.ByteArrayOutputStream;
+import java.io.IOException;
+import java.io.PipedInputStream;
+import java.io.PipedOutputStream;
+import java.nio.charset.Charset;
+
+import static org.junit.Assert.*;
```
darranl: The static imports need to be the first set of imports,

core-feature-pack/src/main/resources/modules/system/layers/base/org/jboss/as/domain-add-user/main/module.xml
```diff
@@ -34,6 +34,7 @@
     </resources>
 
     <dependencies>
+        <module name="org.aesh"/>
```
darranl: Was the org.aesh import really needed in the end on this one?  This module is predominantly a dependency on org.jboss.as.domain-management with a main class specified.

domain-management/pom.xml
```diff
@@ -71,6 +71,10 @@
     </dependencyManagement>
 
     <dependencies>
+        <dependency>
+            <groupId>org.aesh</groupId>
```
darranl: I will let @bstansberry also comment on this dependency.

IMO I am not overly concerned as the add-user utility is destined to be deprecated and either re-written or replaced hopefully soon.

bstansberry:
I don't want aesh as a dependency of the server runtime domain-management module or the domain-management maven module. If it has to be used we need to split this tool into a separate module. Really that should be done anyway; people wanting a slim server shouldn't have to carry a client tool.

We need to narrow our server dependency graph, not expand it.

darranl:
Ok so this is adding a dependency on aesh which the CLI already has a dependency on, I wonder if we should talk to @jfdenise about co locating with the CLI.

For add-user 2.0 we have some other options to look at anyway but I would like to see if we can have a closer relationship with the CLI as we have some options to tie into management opts as well (if we want to take those options) 

Or another option is we move it over to wildfly-elytron-tool if we really do want a stand alone utility which also has it's own merits.

jfdenise:
@darranl , in WF12 we are introducing a new 'security' group of commands in CLI. User management would make full sense there. CLI offers some nice features (prompting, connections, ...).
One issue I see is the fact that generic CLI script doesn't have 'user management' related options such as -u and/or -p (that I really like). This would have to be handled by wrapper scripts calling into CLI script I guess.

core-feature-pack/src/main/resources/modules/system/layers/base/org/jboss/as/domain-add-user/main/module.xml
MMarus: Yes, You are right. I am sorry I didn't noticed this. I fixed it along with the other suggestions from your comments. Thank You.

ctomc:
this is ok to test

darranl:
@bstansberry I think we can close this one out, within the next couple of WildFly releases we expect major changes this area likely involving collaboration with different tools.

PR #3066, opened 2018-01-25
[WFCORE-3541] ScramDigestPassword into set-password operation

hkalina:
Elytron security:
Adding scram password support into set-password operation of modifiable security realms.

https://issues.jboss.org/browse/WFCORE-3541


hkalina:
@darranl can you approve this?

elytron/src/main/java/org/wildfly/extension/elytron/ModifiableRealmDecorator.java
```diff
@@ -414,6 +415,30 @@ protected void executeRuntimeStep(final OperationContext context, final ModelNod
                     .build();
         }
 
+        static class ScramDigest {
+            static final SimpleAttributeDefinition ALGORITHM = new SimpleAttributeDefinitionBuilder(ElytronDescriptionConstants.ALGORITHM, ModelType.STRING)
+                    .setRequired(false)
+                    .setDefaultValue(new ModelNode(ScramDigestPassword.ALGORITHM_SCRAM_SHA_1))
```
fjuma: It would be good to use a stronger hash function by default. It looks like SHA-512 is being used for the defaults for other password types.

hkalina:
@fjuma updated

PR #3067, opened 2018-01-25
WFCORE-3546 Update MoreTestCase to hide failures caused by WFCORE-3545

dudaerich:


jfdenise:
I approve this change.

PR #3068, opened 2018-01-25
Fix for WFCORE-3548, CLI, Command executor should create daemon thread

jfdenise:
Make the created threads to be daemon. Manually tested with a standalone client application calling into CLI class:
CLI cli = CLI.newInstance();
cli.cmd("version");
==> process can exit.

This doesn't mean that terminate() (or disconnect if connected) shouldn't be called, that is still the way to go. It seems cleaner to have this thread a daemon.

This behaviour has been introduced in 7.1, older CLI class was not able to execute command if not connected (so no way, for example, to handle embedded nor version command without connecting).


bstansberry:
retest this please

bstansberry:
retest this please

PR #3069, opened 2018-01-25
[WFCORE-3549] Downgrade jboss-logmanager to a final version, 2.0.9.Final

jamezp:
https://issues.jboss.org/browse/WFCORE-3549

We should be keeping components as final. An alpha version of the log manager was added so the [`json-formatter`](https://github.com/wildfly/wildfly-core/pull/2872) could be added. However this should only be done once the log manager itself is final.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -113,7 +113,7 @@
         <version.org.jboss.logging.jboss-logging>3.3.1.Final</version.org.jboss.logging.jboss-logging>
         <version.org.jboss.logging.jboss-logging-tools>2.1.0.Final</version.org.jboss.logging.jboss-logging-tools>
         <version.org.jboss.logmanager.commons-logging-jboss-logmanager>1.0.3.Final</version.org.jboss.logmanager.commons-logging-jboss-logmanager>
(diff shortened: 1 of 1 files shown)


bstansberry:
I don't know why TeamCity doesn't report the exception cause; the ModuleLoadException should have one.

From the first line in the stack trace I believe it's failing at ModuleXmlParser L976:

```
resourceLoader = MavenArtifactUtil.createMavenArtifactLoader(mavenResolver, name);
```

My guess is it's something funky in how the internal maven stuff in jboss-modules is resolving things. Obviously the artifact is available in the CI process or build would have failed.

jamezp:
Yeah that's what didn't make sense to me. If the artifact was missing all the tests should be failing. It really just doesn't make sense to me.

jamezp:
retest this please

bstansberry:
retest this please

I'm real tempted to merge this for today's core release if keeps on passing now. I fear the gods will punish me though.

jamezp:
I just reran the two failed jobs. Fingers crossed they will pass this time.

PR #3073, opened 2018-01-30
Fix WFCORE-3565: Check task queue for more tasks

cboehme:
Updates the runQueuedTask() method to always check whether there are any
tasks in the task queue. This prevents tasks from getting stuck in the
task queue forever if they were placed in there while the container was
at the max request limit.

Additionally, the changes to the runQueuedTask() method in this commit
ensure that forced tasks do not get lost anymore if the container is
paused and has reached the request limit.

https://issues.jboss.org/browse/WFCORE-3565


luck3y:
This is ok to test.

bstansberry:
@stuartwdouglas Does this look ok?

@cboehme Thanks for the report and the patch! A bit of paperwork: I don't believe we've asked you this before, and before we can merge a PR from someone the first time we'll have to, so I'll ask now:

cboehme:
Thanks for the quick feedback! Of course, I do agree to the terms of the MIT license.

PR #3077, opened 2018-02-01
[WFCORE-3555] Pipe with echo commad gives incorrect result

xstefank:
WFCORE jira: https://issues.jboss.org/browse/WFCORE-3555


xstefank:
@jfdenise thanks, that is a valid point, I was aiming for the reusability with other commands, I will fix this pr

PR #3079, opened 2018-02-05
[WFCORE-3553] Improve message from CLI when writing into the director…

xstefank:
…y without permission

WFCORE: https://issues.jboss.org/browse/WFCORE-3553

### DIFF
--- a/cli/src/main/java/org/jboss/as/cli/impl/CommandContextImpl.java
+++ b/cli/src/main/java/org/jboss/as/cli/impl/CommandContextImpl.java
@@ -29,6 +29,7 @@ import java.io.InputStream;
 import java.io.PrintStream;
 import java.net.URI;
 import java.net.URISyntaxException;
+import java.nio.file.AccessDeniedException;
 import java.security.AccessController;
 import java.security.GeneralSecurityException;
 import java.security.KeyStore;
@@ -1638,6 +1639,10 @@ public class CommandContextImpl implements CommandContext, ModelControllerClient
                             invContext.getConfiguration().getOutputRedirection().close();
                         } catch (IOException ex) {
                             // Message must contain the Exception and the localized message.
+                            if (ex instanceof AccessDeniedException) {
+                                String message = ex.getMessage();
+                                throw new CommandLineException((message != null ? message : line) + " (Access denied)");
+                            }
                             throw new CommandLineException(ex.toString());
                         }
                     }
--- a/testsuite/standalone/src/test/java/org/jboss/as/test/integration/management/cli/FilePermissionTestCase.java
+++ b/testsuite/standalone/src/test/java/org/jboss/as/test/integration/management/cli/FilePermissionTestCase.java
@@ -17,16 +17,18 @@ package org.jboss.as.test.integration.management.cli;
 
 import org.hamcrest.CoreMatchers;
 import org.jboss.as.cli.CommandContext;
+import org.jboss.as.cli.CommandLineException;
 import org.jboss.as.cli.Util;
 import org.jboss.as.test.integration.management.util.CLITestUtil;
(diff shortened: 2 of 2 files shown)


cli/src/main/java/org/jboss/as/cli/impl/CommandContextImpl.java
```diff
@@ -1634,6 +1635,9 @@ private void handleCommand(ParsedCommandLine parsed) throws CommandFormatExcepti
                             invContext.getConfiguration().getOutputRedirection().close();
                         } catch (IOException ex) {
                             // Message must contain the Exception and the localized message.
+                            if (ex instanceof AccessDeniedException) {
+                                throw new CommandLineException(ex.getLocalizedMessage() + " (Permission denied)");
+                            }
                             throw new CommandLineException(ex.toString());
```
hkalina: Would not be attaching cause (at least for the else part) appropriate?

xstefank: @hkalina Thanks for taking a look. If part is requested simplification of the message this issue is about, the else part has not been changed, if you think that the cause is required, it should probably be covered by the different issue

jfdenise: Attaching the cause has an impact on the content displayed to the user. We need to find the right balance of exposed content to a user (as opposed to a programmatic call that would require all full stacks).

xstefank:
retest this please

cli/src/main/java/org/jboss/as/cli/impl/CommandContextImpl.java
```diff
@@ -1634,6 +1635,9 @@ private void handleCommand(ParsedCommandLine parsed) throws CommandFormatExcepti
                             invContext.getConfiguration().getOutputRedirection().close();
                         } catch (IOException ex) {
                             // Message must contain the Exception and the localized message.
+                            if (ex instanceof AccessDeniedException) {
+                                throw new CommandLineException(ex.getLocalizedMessage() + " (Permission denied)");
```
jfdenise: you could add a check for null localized message.

jfdenise: When looking at the AcessDenied exception, I can see some interesting methods (getFile, getMessage, ...) Can we use them to build a user friendly message?

xstefank: I tried to get all the information from the exception, it does not look like we can improve: 
```
localized message - /dir/without/permission
message - /dir/without/permission
file - /dir/without/permission
other file - null
cause - null
toString - java.nio.file.AccessDeniedException: /dir/without/permission
reason - null
```

jfdenise: @xstefank , thank-you to have look into this possibility, the safest message seems to me:
"Access denied to " + file
What do you think?

xstefank: @jfdenise the file may be null, I was thinking that `(message != null ? message : line) + (Permission denied)` would be sufficient in the rare case that exception did not included the file name

but definitely I can include your suggestion

jfdenise: Yes, if we have a null file, then message should be the fallback. I would prefer "Access denied" instead of "Permission denied" this seems more in line with the reported problem.

xstefank: @jfdenise definitely, thanks for the review

xstefank:
Test failures looks unrelated

PR #3084, opened 2018-02-07
[WFCORE-3576] help message and debug log message for deployment runti…

soul2zimate:
…me-name without a file extension.
https://issues.jboss.org/browse/WFCORE-3576

adds help message and warning log message if there is a --runtime-name value (or copied from provided --name value) without a file extension in deployment operations.

### DIFF
--- a/server/src/main/java/org/jboss/as/server/deployment/DeploymentHandlerUtils.java
+++ b/server/src/main/java/org/jboss/as/server/deployment/DeploymentHandlerUtils.java
@@ -33,6 +33,7 @@ import java.io.InputStream;
 import java.net.MalformedURLException;
 import java.net.URL;
 import java.util.List;
+import java.util.regex.Pattern;
 import java.util.zip.ZipEntry;
 import java.util.zip.ZipOutputStream;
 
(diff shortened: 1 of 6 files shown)


soul2zimate:
retest this please

server/src/main/java/org/jboss/as/server/deployment/DeploymentHandlerUtils.java
```diff
@@ -47,6 +48,8 @@
  */
 public abstract class DeploymentHandlerUtils {
 
+    public static final Pattern ARCHIVE_PATTERN = Pattern.compile("^.*\\.(?:(?:[SsWwJjEeRr][Aa][Rr])|(?:[Ww][Aa][Bb])|(?:[Ee][Ss][Aa]))$");
```
jmesnil: Out of curiosity, could you describe the pattern?
It seems to matches more than ear, jar, war, sar, rar. What are the last 2 groups for?

bstansberry:
I do not agree with this. It is not a requirement of WildFly Core, nor of WildFly, that deployments end in ear war jar sar etc, or really any pattern whatsoever. An extension isn't even required, as is shown by the point mentioned in the JIRA about arquillian-service. This is an architectural violation in that it has the kernel imposing requirements that an Extension, by design, is free to violate.

soul2zimate:
thanks, I'm closing this and moving back to Jira for further action.

PR #3086, opened 2018-02-08
WFCORE-3584 Stability monitor should be exposed to DUP's

stuartwdouglas:
There are some use cases (e.g. shared java:global bindings) that result in services being installed outside of a deployments ServiceTarget. This means that are not added to the stability monitor, which results in intermittent test failures due to a false stability being achieved before all services have come up.
At present it is not possible to access this stability monitor, so there is no way to fix this issue.


bstansberry:
@dmlloyd Please approve. This seems ok to me. The concern I have is people not cleaning up properly, but that's an issue for any code that uses this attachment, not for the question of whether to provide the attachment at all.

Any future version of our DUP API I think we'd need to solve the equivalent problem in some way, so I don't see this as creating some new category of behavior that we wouldn't want later.

PR #3088, opened 2018-02-08
WFCORE-3409 embed-server doesn't honor "module.path" system property

ctomc:
This helps tremendously with testsuite optimization (no need to copy over modules jar in manualmode)


ctomc:
@luck3y  can you approve this

rsvoboda:
@wildfly/mergers-core can you look at this ?

marekkopecky:
Whouldn't be better to use the same module path, as parent CLI process uses? This path is stored in JBOSS_MODULEPATH env variable:
https://github.com/wildfly/wildfly-core/blob/master/core-feature-pack/src/main/resources/content/bin/jboss-cli.sh#L42

cc: @jfdenise

ctomc:
@marekkopecky as long as it is consistent. Currently embed-server does not honor this, but embed-host-controller does.

PR #3090, opened 2018-02-08
WFCORE-3588 Intermittent failures in LoggingSubsystemRollbackTestCase

stuartwdouglas:
This change makes every byteman scanner use a different port. If a test is slow to release
a port for some reason it will not interfere with subsequent tests.

### DIFF
--- a/logging/pom.xml
+++ b/logging/pom.xml
@@ -51,13 +51,19 @@
                     </execution>
                 </executions>
             </plugin>
+
             <plugin>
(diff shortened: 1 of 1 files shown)


jamezp:
This looks good and makes sense to me.

bstansberry:
retest this please

bstansberry:
@stuartwdouglas Can you update the PR description as this is no longer doing what you said there? Now it's using the surefire+byteman config patterns used in other modules that use byteman. Which is fine; they seem to work reliably.

PR #3091, opened 2018-02-08
WFCORE-3586 - issues with adding jmx to embedded HC started with empty config

luck3y:
https://issues.jboss.org/browse/WFCORE-3586

### DIFF
--- a/controller/src/main/java/org/jboss/as/controller/ModelControllerImpl.java
+++ b/controller/src/main/java/org/jboss/as/controller/ModelControllerImpl.java
@@ -898,7 +898,12 @@ class ModelControllerImpl implements ModelController {
 
     PathAddress getModelControllerResourceAddress(ManagementModel managementModel) {
         if (modelControllerResourceAddress == null) {
-            String hostName = managementModel.getRootResource().getChildrenNames(HOST).iterator().next();
+            Set<String> hosts = managementModel.getRootResource().getChildrenNames(HOST);
+            // if we don't actually have a host name yet (e.g. --empty-host-config) we return null for now.
+            if (hosts.size() == 0) {
+                return null;
+            }
+            String hostName = hosts.iterator().next();
             modelControllerResourceAddress = PathAddress.pathAddress(HOST, hostName).append(MODEL_CONTROLLER_ADDRESS);
         }
         return modelControllerResourceAddress;
--- a/host-controller/src/main/java/org/jboss/as/host/controller/DomainModelControllerService.java
+++ b/host-controller/src/main/java/org/jboss/as/host/controller/DomainModelControllerService.java
@@ -635,7 +635,6 @@ public class DomainModelControllerService extends AbstractControllerService impl
 
         final ServiceTarget serviceTarget = context.getServiceTarget();
         boolean ok = false;
-        boolean isEmptyHost = false;
         boolean reachedServers = false;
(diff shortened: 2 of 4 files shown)


luck3y:
@bstansberry WDYT? It's hacky.

@aloubyansky This does allow your test script (https://s3.amazonaws.com/uploads.hipchat.com/126035/1291022/gcLrGyH8oV4w1Wv/host-slave.xml)
to execute.

controller/src/main/java/org/jboss/as/controller/ModelControllerImpl.java
```diff
@@ -898,6 +898,11 @@ void addFailureDescription(ModelNode operation, ModelNode failure) {
 
     PathAddress getModelControllerResourceAddress(ManagementModel managementModel) {
         if (modelControllerResourceAddress == null) {
+            Set<String> hosts = managementModel.getRootResource().getChildrenNames(HOST);
+            // if we don't actually have a host name yet (e.g. --empty-host-config) we return null for now.
+            if (hosts.size() == 0) {
+                return null;
+            }
```
bstansberry: Is this directly related to the JMX problem, or just something you noticed as you were working this issue? If the later please file a separate JIRA and split the commit. I can certainly see why this would need to be fixed, so no problem with the change.

controller/src/main/java/org/jboss/as/controller/ModelControllerImpl.java
```diff
@@ -898,6 +898,11 @@ void addFailureDescription(ModelNode operation, ModelNode failure) {
...
             String hostName = managementModel.getRootResource().getChildrenNames(HOST).iterator().next();
```
bstansberry: Please reuse the 'hosts' var you have now.

host-controller/src/main/java/org/jboss/as/host/controller/DomainModelControllerService.java
```diff
@@ -649,14 +649,15 @@ protected void boot(final BootContext context) throws ConfigurationPersistenceEx
             if (hostBootOps.isEmpty()) { // booting with empty config
                 isEmptyHost = true;
                 ok = bootEmptyConfig();
-                return;
```
bstansberry: This is kind of just a code style thing, but is it correct to have bootEmptyConfig take the context as a param and create and boot with the registerModelControllerServiceInitializationBootStep(context) step, instead of booting with an empty list? Doing that and continuing to return here is cleaner looking code vs multiple checks of isEmptyHost.  (Actually if we continue to return here the 'isEmptyHost' var should be removed as it's not used in the current code.)

bstansberry:
@luck3y I had some comments but TBH this is less hacky than I imagined a fix could be. I figured you'd have to find a way to pass a null hostName to PatchIntegrationFactory with hacks to get that to be null, but the way you did it avoided all that. Really, I'm not sure the PatchIntegrationFactory change qualifies as a hack at all. You made that method check for desired but not required state and if it's not present it reacts correctly.

bstansberry:
Nice. :)

luck3y:
retest this please

luck3y:
retest this please

jmesnil:
The ManagementOnlyModeTestCase failure looks intermittend and not related to the PR.

PR #3095, opened 2018-02-11
Revert "WFCORE-3584 Stability monitor should be exposed to DUP's"

stuartwdouglas:
Reverts wildfly/wildfly-core#3086

Turns out that this is the wrong stability monitor


stuartwdouglas:
retest this please

PR #3098, opened 2018-02-11
WFCORE-3594 Reload is racey with regard to reading the process state

stuartwdouglas:
Presently when issuing a reload operation the process state is updated asyncrounously,
which means that it is possible for a subsequent read of the process-state to return
a state of 'running' even though the server is actually in the process of reloading.

This can cause intermittent failures in any tests that perform a reload, as the general
pattern followed by these tests is to do a reload and then check the process state in a
loop until it is back to 'running'.

### DIFF
--- a/controller/src/main/java/org/jboss/as/controller/operations/common/ProcessReloadHandler.java
+++ b/controller/src/main/java/org/jboss/as/controller/operations/common/ProcessReloadHandler.java
@@ -83,12 +83,12 @@ public abstract class ProcessReloadHandler<T extends RunningModeControl> impleme
                     @Override
                     public void handleResult(OperationContext.ResultAction resultAction, OperationContext context, ModelNode operation) {
                         if(resultAction == OperationContext.ResultAction.KEEP) {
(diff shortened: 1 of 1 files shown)


bstansberry:
retest this please

It's still racy, this just makes it less so. It's ok though, as it's an improvement. 

It's still racy because the reload op response is sent to the client before this code gets called; that's necessary to ensure the response goes out before the reload shuts down the connection.

To make this non-racy we'd have to make available some kind of hook such that this ResultHandler can force the response to go out after it sets this state, rather that having the response go out when the op commits, with no control by ProcessReloadHandler.

Graceful reload will add another twist to this as the delay between op commit and the beginning of stopping can be arbitrarily long.

bstansberry:
A hacky way to provide a hook would be to add an Attachment API to org.jboss.as.controller.ModelController.OperationTransaction, with the kernel passing any attachment through to the OperationContext. That would be pretty nasty though. It's all kernel stuff though; OperationTransaction etc are not part of the Extension API.

PR #3099, opened 2018-02-12
WFCORE-3595 SuspendOnSoftKillTestCase can fail intermittently

stuartwdouglas:
The issue is that TestUndertowService is designed for testing suspend/resume rather than an actual graceful shutdown. Because it attempts to perform a dispatch() and queue a task rather than just immediately setting a 503 response it is possible that this request will fail with a 500 response if the XNIO executor is in the process of shutting down.
It would be better if this test checked for a 503 before the skip-graceful request is made.

### DIFF
--- a/testsuite/manualmode/src/test/java/org/jboss/as/test/manualmode/suspend/SuspendOnSoftKillTestCase.java
+++ b/testsuite/manualmode/src/test/java/org/jboss/as/test/manualmode/suspend/SuspendOnSoftKillTestCase.java
@@ -198,6 +198,15 @@ public class SuspendOnSoftKillTestCase {
             }
             Assert.assertEquals("SUSPENDING", suspendState);
 
+            final HttpURLConnection conn = (HttpURLConnection) new URL(address).openConnection();
+            try {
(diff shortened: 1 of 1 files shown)


testsuite/manualmode/src/test/java/org/jboss/as/test/manualmode/suspend/SuspendOnSoftKillTestCase.java
```diff
@@ -210,18 +219,6 @@ public Object call() throws Exception {
                 // ignore; it's fine if the server's down
             }
 
-            final HttpURLConnection conn = (HttpURLConnection) new URL(address).openConnection();
```
jmesnil: Should we keep this call and just check it is a 5xx response (as it can be either 500 or 503 aiui)?

stuartwdouglas: It can actually be 500, 503 or a complete failure (IOException), so I don't know if it i really proves much.

PR #3100, opened 2018-02-13
[WFCORE-3600] Add a new org.wildfly.javax.json module to temporally a…

jamezp:
…llow core modules to depend on it until we update to JSON-P 1.1 in WildFly.

https://issues.jboss.org/browse/WFCORE-3600

This introduced a new temporary module in core for core components to rely on. This is required so the wrong module is not loaded if the `ee8.preview.mode` property is set to `true` in WildFly. Note this module *will* go away eventually.


jamezp:
The test failure seems to be known and #3101 should help resolve the failure.

PR #3104, opened 2018-02-14
[WFCORE-3606]: Covariant issue when compiling with java 9.

ehsavoie:
Casting to fix the issue.

Jira: https://issues.jboss.org/browse/WFCORE-3606


dmlloyd:
I don’t think this is the right fix. What we probably ought to be doing is adding a profile which is activated on JDK9 or higher which sets the release property of the compiler plugin to 8, else we will hit this problem again.

ctomc:
probably better if we just override maven-compiler-plugin for deployment-scanner
and set <release>8</release> in it.
we cannot do that globally yet.

PR #3107, opened 2018-02-14
WFCORE-3608 There is no way to define core pool size for io subsystem

ctomc:


LittleJohnII:
They tell us we should participate more on the PRs under the new process, so yeah, I went through the PR and FWIW, this looks good to me :).

io/subsystem/src/main/java/org/wildfly/extension/io/IOExtension.java
```diff
@@ -65,14 +65,15 @@ public void initializeParsers(ExtensionParsingContext context) {
         context.setSubsystemXmlMapping(SUBSYSTEM_NAME, Namespace.IO_1_0.getUriString(), IOSubsystemParser_1_0::new);
         context.setSubsystemXmlMapping(SUBSYSTEM_NAME, Namespace.IO_1_1.getUriString(), IOSubsystemParser_1_1::new);
         context.setSubsystemXmlMapping(SUBSYSTEM_NAME, Namespace.IO_2_0.getUriString(), IOSubsystemParser_2_0::new);
+        context.setSubsystemXmlMapping(SUBSYSTEM_NAME, Namespace.IO_3_0.getUriString(), IOSubsystemParser_3_0::new);
```
jmesnil: we decided to instantiate the latest parser (and use the same instance to register the XMLElementWriter)

io/subsystem/src/main/java/org/wildfly/extension/io/IOSubsystemTransformers.java
```diff
@@ -44,15 +45,27 @@ public String getSubsystemName() {
     public void registerTransformers(SubsystemTransformerRegistration registration) {
         ChainedTransformationDescriptionBuilder chainedBuilder = TransformationDescriptionBuilder.Factory.createChainedSubystemInstance(registration.getCurrentSubsystemVersion());
 
-        // Current 3.0.0 to 2.0.0, aka EAP 7.0.0
-        buildTransformers_2_0(chainedBuilder.createBuilder(registration.getCurrentSubsystemVersion(), VERSION_2_0));
+        // Current 4.0.0 to 3.0.0, aka EAP 7.1.0
```
jmesnil: I don't understand the mapping between versions.

Current is 3.0.0 (that you add in this PR), there is no 4.0.0, right?

io/subsystem/src/main/java/org/wildfly/extension/io/IOSubsystemTransformers.java
```diff
@@ -32,7 +32,8 @@
  * @author Tomaz Cerar (c) 2017 Red Hat Inc.
  */
 public class IOSubsystemTransformers implements ExtensionTransformerRegistration {
-    static final ModelVersion VERSION_2_0 = ModelVersion.create(2, 0);
+    private static final ModelVersion EAP_7_0 = ModelVersion.create(2, 0);
```
jmesnil: Does that mean that the 2.0 version is the one in EAP 7.0?

io/tests/src/test/java/org/wildfly/extension/io/IOSubsystemTransformerTestCase.java
```diff
@@ -53,7 +52,7 @@ public IOSubsystemTransformerTestCase() {
 
     @Override
     protected String getSubsystemXml() throws IOException {
-        return readResource("io-2.0-transformer.xml");
+        return readResource("eap-7.1.0-transformer.xml");
```
jmesnil: why did you rename the XML files using EAP versions instead of the subsystem version?

io/subsystem/src/main/resources/org/wildfly/extension/io/LocalDescriptions.properties
```diff
@@ -4,6 +4,7 @@ io.remove=Operation Removes subsystem
 io.worker=Defines workers
 io.worker.add=Adds new worker
 io.worker.remove=Removes worker
+io.worker.task-core-threads=Number of task threads with which worker starts with.
```
rsvoboda: Text should be same as in wildfly-io_3_0.xsd - 
https://github.com/wildfly/wildfly-core/pull/3107/files#diff-4f40ca798a3d232a7d9cc3fb027ebdd5R78

```
                        Specify the starting number of threads for the worker task thread pool.
                        If not set, it defaults to 2.
```

io/tests/src/test/resources/org/wildfly/extension/io/eap-7.0.0-transformer.xml
```diff
@@ -16,9 +16,9 @@
   ~ limitations under the License.
   -->
 
-<subsystem xmlns="urn:jboss:domain:io:2.0">
-    <worker name="default" task-keepalive="100" stack-size="5000"/>
-    <worker name="second-worker" io-threads="5" stack-size="300" task-keepalive="100" task-max-threads="200"/>
+<subsystem xmlns="urn:jboss:domain:io:3.0">
+    <worker name="default" task-keepalive="100" stack-size="5000" />
+    <worker name="second-worker" io-threads="5" stack-size="300" task-keepalive="100" task-max-threads="400"/>
```
rsvoboda: Why is task-max-threads moved to 400 ?

io/subsystem/src/main/java/org/wildfly/extension/io/WorkerResourceDefinition.java
```diff
@@ -74,6 +74,12 @@
     static final RuntimeCapability<Void> IO_WORKER_RUNTIME_CAPABILITY =
             RuntimeCapability.Builder.of(IOServices.IO_WORKER_CAPABILITY_NAME, true, XnioWorker.class).build();
 
+
+    static final OptionAttributeDefinition WORKER_TASK_CORE_THREADS = new OptionAttributeDefinition.Builder(Constants.WORKER_TASK_CORE_THREADS, Options.WORKER_TASK_CORE_THREADS)
+            .setAllowExpression(true)
+            .setDefaultValue(new ModelNode(2))
```
rsvoboda: Can you please share some reasoning why is default value set to 2?

ctomc: we always had (this attribute was already present in 1.0 version of subsystem) that as 2, as a starting thread pool size sounds reasonable. 
We also discussed that with David and agreed that is a reasonable default.

dmlloyd:
@ctomc did you want to wrap this up, or shall I take it over?

rsvoboda:
Tomaz is on PTO, I think he will be off for 2 more weeks.

dmlloyd:
Superceded by #3230.

PR #3109, opened 2018-02-14
[WFCORE-3597] Allow reload handler to signal when to send the early r…

bstansberry:
…esponse to the client rather than always sending it when the op is prepared

https://issues.jboss.org/browse/WFCORE-3597

Our remote request handlers for external end-user requests send the response to reload and shutdown early, at op preparation time instead of at completion time. This is needed so the response can go out before the services providing the connection to the client are closed. But preparation time is a bit too early. This commit allows the remote request handlers to attach an object to the OperationContext which the OSHs can use to signal when it's ok to send the response.

Currently used just for 'reload', which is the one where the too-early response is causing problems in the testsuite. Adding for shutdown will take a bit more thought.

The change looks fairly big to domain-http but that's because I factored out some common code between DomainApiHandler and DomainApiGenericOperationHandler.

This is built on top of the commit in #3121 as testing of this work resulted in the WFCORE-3624 issue appearing frequently. I've seen the WFCORE-3624 issue in other CI runs though, so that problem is not a result of this change.

Currently this is open just to get CI feedback. I'd want a lot of runs of this before merging.

### DIFF
--- a/controller/src/main/java/org/jboss/as/controller/AbstractControllerService.java
+++ b/controller/src/main/java/org/jboss/as/controller/AbstractControllerService.java
@@ -60,6 +60,7 @@ import org.jboss.msc.service.Service;
 import org.jboss.msc.service.ServiceContainer;
 import org.jboss.msc.service.ServiceController;
 import org.jboss.msc.service.ServiceTarget;
+import org.jboss.msc.service.StabilityMonitor;
 import org.jboss.msc.service.StartContext;
 import org.jboss.msc.service.StartException;
 import org.jboss.msc.service.StopContext;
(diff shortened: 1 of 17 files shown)


bstansberry:
retest this please

bstansberry:
retest this please

bstansberry:
retest this please

bstansberry:
retest this please

bstansberry:
retest this please

bstansberry:
retest this please

bstansberry:
retest this please

bstansberry:
retest this please

bstansberry:
retest this please

bstansberry:
retest this please

bstansberry:
retest this please

bstansberry:
retest this please

bstansberry:
retest this please

bstansberry:
retest this please

bstansberry:
retest this please

PR #3111, opened 2018-02-15
[WFCORE-3604] Attribute required-attributes of Elytron x500-attribute-principal-decoder cannot be added to configuration

yersan:
REQUIRED_OIDS attribute unwrap returns an immutable array throwing an UnsupportedOperationException when new elements are added. This patch fixes this situation.

Jira issue: https://issues.jboss.org/browse/WFCORE-3604


yersan:
@darranl I updated the issue following your suggestion. Test case it is still pending, I'm checking how to add it.

yersan:
@darranl Added this specific case to SubsystemParsingTestCase

PR #3119, opened 2018-02-19
[WFCORE-3622] Fix the SuspendOnSoftKillTestCase cleanup work

bstansberry:
https://issues.jboss.org/browse/WFCORE-3622

Test was calling serverController.start() on a killed controller, assuming that would start the server again to allow a subsequent undeploy op. But the serverController didn't know it was killed, so the 'start()' call did nothing and the undeploy wasn't getting executed.

Use https://github.com/wildfly/wildfly-core/pull/3119/files?w=1 to look at this or the change looks much bigger due to removal of a try/catch/finally surrounding the main part of the test.


bstansberry:
#3115 fixes the test failure.

PR #3120, opened 2018-02-20
[WFCORE-3623] Remove ModuleIndexService

bstansberry:
https://issues.jboss.org/browse/WFCORE-3623

I see no uses of this in core or full and the class is unmodified since 2011.


bstansberry:
Apparently it IS used.

stuartwdouglas:
It should still be removed, the useless dependency in core just needs to be removed first

stuartwdouglas:
https://github.com/wildfly/wildfly/pull/10908/files

The constant Services.JBOSS_MODULES_INDEX_SERVICE can also be removed

PR #3124, opened 2018-02-20
[WFCORE-3579] LocalClient should establish an AccessAuditContext for …

bstansberry:
…all user calls (but only for user calls)

https://issues.jboss.org/browse/WFCORE-3579

This updates the client used by in-vm callers, including the offline CLI, to establish an AccessAuditContext when the client is for end user use (as opposed to internal functions like the deployment-scanner and the content repository cleaner.) A check for an AccessAuditContext is the basis of the WFCORE-184 fix that prevents end users being able to invoke ops while boot is in progress. Preventing that is the fix for the specific problem reported in the JIRA. 

Mergers: Please talk with me before merging.


bstansberry:
@darranl WDYT?

bstansberry:
retest this please

PR #3125, opened 2018-02-20
[WFCORE-3629] Don't ignore an empty host.xml in the non-embedded case

bstansberry:
https://issues.jboss.org/browse/WFCORE-3629


bstansberry:
@luck3y WDYT?

PR #3128, opened 2018-02-21
[WFCORE-3633]: Read only mode is not properly supported for domain.

ehsavoie:
Updating the ConfigurationFile.InteractionPolicy for host / domain if the args are --read-only-host-config and --read-only-domain-config.

Jira: https://issues.jboss.org/browse/WFCORE-3633


bstansberry:
@ehsavoie Have you manually verified this works, i.e. that the specified files are not updated but that if you make a change to domain and host config (say some system props) and then reload, those props remain in the config? But they are lost following a restart.

I see we have no regression tests of this, afaict neither for these domain ones nor for read-only-server-config. :(

ehsavoie:
Well, in my case i had an OSH that was registered only in read-only so i could check this. i'll write some test for this

bstansberry:
@ehsavoie If you've tested this manually and adding regression tests proves a hassle, please go ahead and re-open this and the tests can come later. I made this a release blocker and want to be sure the fix is merged this week.

PR #3129, opened 2018-02-21
[WFCORE-3587] Remove unused dependencies

dmlloyd:


core-feature-pack/src/main/resources/modules/system/layers/base/org/jboss/as/controller/main/module.xml
```diff
@@ -38,7 +38,6 @@
         <module name="org.jboss.as.controller-client" export="true"/>
         <module name="org.jboss.as.core-security"/>
         <module name="org.jboss.as.protocol"/>
-        <module name="org.jboss.as.version"/>
```
bstansberry: org.jboss.as.controller.ModelVersion imports org.jboss.as.version.Version.  It's to bring in some constant ints though, so the compiler eliminates the link.

I guess it's fine to have this kind of discrepancy between the compile-time and runtime dependencies. It makes me a bit nervous though.

dmlloyd: Hmm I wonder if we should go to measures to prevent it from being folded into a constant.

dmlloyd:
Retest this please

core-feature-pack/src/main/resources/modules/system/layers/base/org/jboss/as/controller/main/module.xml
jmesnil: @dmlloyd are you planning to reintroduce this `org.jboss.as.version` or are you ok to merge as is?

dmlloyd: I'm okay with it as is, but I can reintroduce it if we want to keep it.

bstansberry: OK with me too.

PR #3131, opened 2018-02-22
[WFCORE-3633]: Read only mode is not properly supported for domain.

ehsavoie:
Updating the ConfigurationFile.InteractionPolicy for host / domain if the args are --read-only-host-config and --read-only-domain-config.

Jira: https://issues.jboss.org/browse/WFCORE-3633

### DIFF
--- a/testsuite/domain/src/test/java/org/jboss/as/test/integration/domain/ReadOnlyModeTestCase.java
+++ b/testsuite/domain/src/test/java/org/jboss/as/test/integration/domain/ReadOnlyModeTestCase.java
@@ -1,0 +1,171 @@
+/*
+ * Copyright 2018 JBoss by Red Hat.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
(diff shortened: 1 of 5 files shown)


testsuite/domain/src/test/java/org/jboss/as/test/integration/domain/ReadOnlyModeTestCase.java
```diff
@@ -0,0 +1,171 @@
...
+        domainManager.getDomainSlaveLifecycleUtil().startAsync();
+        try (final DomainClient clientMaster = getDomainClient(domainConfig.getMasterConfiguration())) {
+            waitForHostControllerBeingStarted(TIMEOUT_S, clientMaster);
+            Assert.assertTrue(Operations.getFailureDescription(clientMaster.execute(Operations.createReadAttributeOperation(domainAddress, "value"))).asString().contains("WFLYCTL0216"));
+            Assert.assertTrue(Operations.getFailureDescription(clientMaster.execute(Operations.createReadAttributeOperation(masterAddress, "value"))).asString().contains("WFLYCTL0216"));
+            Assert.assertTrue(Operations.readResult(clientMaster.execute(Operations.createReadAttributeOperation(slaveAddress, "value"))).asBoolean());
+        }
+    }
+
+
```
bstansberry: Can you file a JIRA to make everything from here on down go away? This seems like something the DomainLifecycleUtil should handle. We've got a few bits and pieces like this scattered around the domain testsuite and it's the kind of thing that ends up causing trouble later.

PR #3137, opened 2018-02-23
WFCORE-3635 Introduce version-matrix bom

ctomc:
Move dependency version management under separate bom, so it is easier to be consumed in full or any other use case.
Keep parent clean of any deps beyond build configuration itself.

change how we consume ApacheDS deps in tests by using their shaded "all" jar instead of importing a osgi bom parent.

Remove need for test runtime creating of org.picketbox module, as it is always added as part of vault feature pack to testsuite modules anyways.


testsuite/shared/src/main/resources/picketbox-module.xml
```diff
@@ -18,7 +18,7 @@
 
 <module xmlns="urn:jboss:module:1.5" name="org.picketbox">
     <resources>
-        <artifact name="org.picketbox:picketbox:${version.org.picketbox}"/>
+        <artifact name="org.picketbox:picketbox:5.0.2.Final"/>
```
psakar: can be used property as before ?

psakar:
Where are all properties gone ?

ctomc:
All properties are part of new version-matrix/pom.xml

name we could probably change to component-matrix.

psakar:
@rsvoboda for WF/EL13. I would look at it once WF/EL13 is out, it would mean to reconfigure some of automation for builds / release

rsvoboda:
@mjurc @msve fyi

What's the main motivation for explicit component version BOM? 
Is it because of Quickstars or to have nice way (maven GAV) to consume WildFly dependencies ?

testsuite/shared/src/main/resources/picketbox-module.xml
rsvoboda: +1 for property

psakar:
@ctomc Based on the experience with those compatibility related dependencies, I would say that the dependency management should be per feature pack, not per project

testsuite/shared/src/main/resources/picketbox-module.xml
ctomc: fixed.

ctomc:
@rsvoboda quickstarts & easier way of consuming dependencies by 3rd party users.
@psakar that is noble idea, but it would require to refactoring much more than just on bom, but whole concept of how we build the server.
Core itself is a single feature-pack in sense. WildFly full is a different beast
But even this only holds true with current concept of feature packs, with new provisioning tool even that wont be true anymore.

PR #3140, opened 2018-02-24
[WFCORE-3646] Improve reload handling in WildFly Core test runner

bstansberry:
1) Don't close client immediately after reload; it's redundant as it's closed again when checking for running status and doing it so quickly risks interrupting the reload in progress
2) Delay 100 ms before the first status check. It takes at least that long to do a reload anyway. This aligns behavior with the ServerReload util.

https://issues.jboss.org/browse/WFCORE-3646


bstansberry:
retest this please

PR #3141, opened 2018-02-26
WFCORE-3648 Options --disabled/--enabled/--replace are missing in CLI code completion

rhusar:
Jira
https://issues.jboss.org/browse/WFCORE-3648

The problem is that activators are establishing expected attributes to be present and passing empty String value (`DependOptionActivator.ARGUMENT_NAME`).

Also, cleans up unnecessary modifiers.

### DIFF
--- a/cli/src/main/java/org/jboss/as/cli/impl/aesh/cmd/deployment/security/OptionActivators.java
+++ b/cli/src/main/java/org/jboss/as/cli/impl/aesh/cmd/deployment/security/OptionActivators.java
@@ -32,7 +32,7 @@ import org.wildfly.core.cli.command.aesh.activator.StandaloneOptionActivator;
  */
 public interface OptionActivators {
 
-    public static class UrlActivator implements OptionActivator {
+    class UrlActivator implements OptionActivator {
 
         @Override
         public boolean isActivated(ParsedCommand pc) {
@@ -42,7 +42,7 @@ public interface OptionActivators {
         }
     }
 
-    public static class NameActivator implements OptionActivator {
+    class NameActivator implements OptionActivator {
 
         @Override
         public boolean isActivated(ParsedCommand pc) {
@@ -53,7 +53,7 @@ public interface OptionActivators {
 
     }
 
-    public static class UndeployNameActivator implements OptionActivator {
+    class UndeployNameActivator implements OptionActivator {
 
         @Override
         public boolean isActivated(ParsedCommand pc) {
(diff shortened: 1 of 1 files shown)


rhusar:
cc @jfdenise

jfdenise:
@rhusar, this is done in purpose, enabled and disabled are not proposed in domain mode.

rhusar:
@jfdenise Please look again, you seem to be confusing the lines with these which are intact:

https://github.com/wildfly/wildfly-core/pull/3141/files#diff-4951a4ca5479737edcea023bef26b67fR147
https://github.com/wildfly/wildfly-core/pull/3141/files#diff-4951a4ca5479737edcea023bef26b67fR176

After the fix, these still do not appear in domain mode as expected.

jfdenise:
I see, so the behavior is expected, in standalone, --enabled and --disabled depend on the Argument (the file path) to be present to be proposed.

rhusar:
So if I understand you correctly, you do not want to present the options until the file path is specified?
I don't see how that makes sense. The file path is required (org.jboss.as.cli.impl.aesh.cmd.deployment.DeployFileCommand#file) so there will always be a file specified for the command to be valid. This there is no reason not to propose those options "dependending on file path to be present". Options are typically proposed right after typing the command, then followed by argument.

jfdenise:
When pressing <TAB> the file path completer will be called first so you will be asked for a file (displaying the lists of file present in the current directory), once you selected a file, the options will be proposed. So you will see --enabled --disabled at the right time.

rhusar:
Unfortunately you are making a wrong assumption that users will always want to type "deployment deploy-file FILE_PATH --disabled" in that exact order. The other way around is perfectly fine "deployment deploy-file --disabled FILE_PATH". The file path completer does not kick in after the typing the first hyphen-minus character is input, e.g. 
```
[standalone@localhost:9990 /] deployment deploy-file --
--headers=  --name=  --runtime-name=  --unmanaged
```
which is correct, but missing all the possible options. This PR fixes that.

jfdenise:
@rhusar , good point! Approved.
JF

rhusar:
@jfdenise Great, thanks. Note that I stumbled upon this when trying to look into https://issues.jboss.org/browse/WFCORE-3649

jfdenise:
@rhusar, I just looked into WFCORE-3649, yes, tha tis caused by the legacy deploy command calling into the new deployment command. The name of the option needs to be passed down, --replace is the new name of --force.

PR #3142, opened 2018-02-26
[WFCORE-3562] Deployment disable-all doesn't correct function at domain

xstefank:
JIRA: https://issues.jboss.org/browse/WFCORE-3562

### DIFF
--- a/cli/src/main/java/org/jboss/as/cli/impl/aesh/cmd/deployment/AbstractUndeployCommand.java
+++ b/cli/src/main/java/org/jboss/as/cli/impl/aesh/cmd/deployment/AbstractUndeployCommand.java
@@ -149,6 +149,7 @@ public abstract class AbstractUndeployCommand extends CommandWithPermissions
         DefaultOperationRequestBuilder builder;
 
         final List<String> deploymentNames;
+        boolean all = false;
         if (name.indexOf('*') < 0) {
             deploymentNames = Collections.singletonList(name);
         } else {
@@ -156,6 +157,7 @@ public abstract class AbstractUndeployCommand extends CommandWithPermissions
             if (deploymentNames.isEmpty()) {
                 throw new CommandFormatException("No deployment matched wildcard expression " + name);
             }
+            all = true;
         }
 
         for (String deploymentName : deploymentNames) {
@@ -165,6 +167,9 @@ public abstract class AbstractUndeployCommand extends CommandWithPermissions
                 if (allRelevantServerGroups) {
                     if (keepContent) {
                         serverGroups = Util.getAllEnabledServerGroups(deploymentName, client);
+                        if (all && serverGroups.isEmpty()) {
+                            continue;
+                        }
                     } else {
                         try {
                             serverGroups = Util.getServerGroupsReferencingDeployment(deploymentName, client);
@@ -179,10 +184,8 @@ public abstract class AbstractUndeployCommand extends CommandWithPermissions
                     serverGroups = Arrays.asList(serverGroupsStr.split(","));
                 }
 
-                if (serverGroups.isEmpty()) {
-                    if (keepContent) {
-                        throw new OperationFormatException("None of the server groups is specified or references specified deployment.");
-                    }
+                if (serverGroups.isEmpty() && keepContent) {
+                    throw new OperationFormatException("None of the server groups is specified or references deployment " + deploymentName);
                 } else {
                     // If !keepContent, check that all server groups have been listed by user.
                     if (!keepContent) {
@@ -195,6 +198,9 @@ public abstract class AbstractUndeployCommand extends CommandWithPermissions
                         }
                     }
                     for (String group : serverGroups) {
+                        if (all && !Util.isDeploymentPresent(deploymentName, client, group)) {
+                            continue;
+                        }
                         ModelNode groupStep = Util.configureDeploymentOperation(Util.UNDEPLOY, deploymentName, group);
                         steps.add(groupStep);
                         if (!keepContent) {
@@ -217,6 +223,10 @@ public abstract class AbstractUndeployCommand extends CommandWithPermissions
                 builder.setOperationName(Util.REMOVE);
                 builder.addNode(Util.DEPLOYMENT, deploymentName);
                 steps.add(builder.buildRequest());
+            }
+        } else {
+            if (ctx.isDomainMode() && allRelevantServerGroups && all && !steps.isDefined()) {
+                throw new OperationFormatException("No enabled deployment found in any server-groups.");
             }
         }
         if (headers != null) {
--- a/testsuite/domain/src/test/java/org/jboss/as/test/integration/domain/management/cli/DeployAllDomainTestCase.java
+++ b/testsuite/domain/src/test/java/org/jboss/as/test/integration/domain/management/cli/DeployAllDomainTestCase.java
@@ -36,6 +36,7 @@ import org.jboss.as.test.integration.management.util.CLITestUtil;
 import org.junit.After;
 import org.junit.AfterClass;
 
+import static org.hamcrest.MatcherAssert.assertThat;
 import static org.hamcrest.CoreMatchers.allOf;
 import static org.hamcrest.CoreMatchers.containsString;
 import static org.jboss.as.test.deployment.DeploymentArchiveUtils.createCliArchive;
@@ -49,7 +50,6 @@ import static org.jboss.as.test.deployment.DeploymentInfoUtils.checkEmpty;
 import static org.jboss.as.test.deployment.DeploymentInfoUtils.deploymentInfo;
 import static org.jboss.as.test.deployment.DeploymentInfoUtils.deploymentList;
 import static org.jboss.as.test.deployment.DeploymentInfoUtils.legacyDeploymentInfo;
-import static org.junit.Assert.assertThat;
 import static org.junit.Assert.fail;
 
 import org.junit.Before;
@@ -117,6 +117,7 @@ public class DeployAllDomainTestCase extends AbstractCliTestBase {
     public void beforeTest() throws Exception {
         ctx = CLITestUtil.getCommandContext(testSupport);
         ctx.connectController();
+        ctx.handleSafe("deployment undeploy * --all-relevant-server-groups");
     }
 
     @After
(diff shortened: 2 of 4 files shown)


xstefank:
getAllEnabledServerGroups() is called only in https://github.com/wildfly/wildfly-core/blob/master/cli/src/main/java/org/jboss/as/cli/impl/aesh/cmd/deployment/AbstractUndeployCommand.java#L167

the problem from the issue is that the first deployment is ADDED in the main-sg and NOT ADDED in other-sg. When disable-all is called it will check for all sg in which this deployment is ENABLED, which in this case is none and fails with exception not allowing other deployments to be processed, that is why I think that getAllEnabledServerGroups is rather scanning for server groups referencing the deployment.

I see that I missed that there is already such method present. The use case here is that disable of already disabled deployment takes no action, but specifying all-relevant-server groups returns no group if the deployment is not enabled in any server group and throws the exception 

I will update this PR if the approach is ok? @jfdenise

jfdenise:
Thank-you, it seems that the problem is, when disable-all is called with --all-relevant-server-groups, we shouldn't stop and throw when we have one deployment that is not enabled in any server-groups. Obviously if we are explicitly disabling a deployment from all-relevant-server-groups we should throw if no server-groups is found.
So I would:
1) Check that we are in "all" case, the name has this information
2) If we are in "all" mode, if all-relevant-server-groups has been passed, if keep-content, then if Util.getAllEnabledServerGroups(deploymentName, client); returns an empty list, I would ignore the deployment name and continue the loop.
At the end of the iteration,
If no server-groups are containing deployments retrieved from the "all" pattern, then we should possibly throw an exception to be inline with what happens when we are disabling a given deployment from all-relevant-server-groups.

xstefank:
@jfdenise updated according to the requirements

xstefank:
retest this please

cli/src/main/java/org/jboss/as/cli/impl/aesh/cmd/deployment/AbstractUndeployCommand.java
```diff
@@ -112,7 +112,7 @@ public CommandResult execute(CommandContext ctx)
         return null;
     }
 
-    static void undeployName(CommandContext ctx, String name,
```
jfdenise: it should stay static.

cli/src/main/java/org/jboss/as/cli/impl/aesh/cmd/deployment/AbstractUndeployCommand.java
```diff
@@ -137,7 +137,7 @@ public ModelNode buildRequest(CommandContext context)
                 serverGroups, keepContent(), headers);
     }
 
-    private static ModelNode buildRequest(CommandContext ctx, String name,
```
jfdenise: it should stay static

cli/src/main/java/org/jboss/as/cli/impl/aesh/cmd/deployment/AbstractUndeployCommand.java
```diff
@@ -179,10 +179,8 @@ private static ModelNode buildRequest(CommandContext ctx, String name,
                     serverGroups = Arrays.asList(serverGroupsStr.split(","));
                 }
 
-                if (serverGroups.isEmpty()) {
-                    if (keepContent) {
-                        throw new OperationFormatException("None of the server groups is specified or references specified deployment.");
-                    }
+                if (serverGroups.isEmpty() && keepContent && !isDisableAll()) {
```
jfdenise: I would move this check in the if allRelevantServerGroups is true block.

cli/src/main/java/org/jboss/as/cli/impl/aesh/cmd/deployment/AbstractUndeployCommand.java
```diff
@@ -225,4 +227,8 @@ private static ModelNode buildRequest(CommandContext ctx, String name,
         }
         return composite;
     }
+
+    private boolean isDisableAll() {
```
jfdenise: by checking the name argument == "*" you have the information that the command is disable "all".

cli/src/main/java/org/jboss/as/cli/impl/aesh/cmd/deployment/AbstractUndeployCommand.java
xstefank: I don't think this is possible, as this will be used also in case when the user won't specify any group

jfdenise: I would not remove the if(keepContent). I sketched a diff that I will send you directly, better than multiple comments ;-).

xstefank:
Waiting for test inclusion

xstefank:
@jfdenise can you review/approve this once more please? thank you

xstefank:
retest this please

marekkopecky:
This PR can be approved&merged after tests will be added. Anyway, it can be reviewed now, and I'm ok with current commit.

xstefank:
@jfdenise I've extended this PR to handle also case when no --all-relevant-server-groups is specified. e.g. `deployment disable-all --server-groups=main-server-group`

xstefank:
@bstansberry test included, if approved this is good to go

bstansberry:
@jfdenise From the comments above it looks like this changed since you approved it. Still ok?

testsuite/domain/src/test/java/org/jboss/as/test/integration/domain/management/cli/DeployAllDomainTestCase.java
```diff
@@ -447,4 +448,225 @@ public void testLegacyEnableSingleAppDeployment() throws CommandLineException, I
         checkExist(result, cliTestApp2War.getName(), ADDED, ctx);
         checkExist(result, cliTestAppEar.getName(), ENABLED, ctx);
     }
+
+    /**
+     * Disabling all deployed application deployments between all server groups.
```
jfdenise: "between all" ==> "present in all server groups"

testsuite/domain/src/test/java/org/jboss/as/test/integration/domain/management/cli/DeployAllDomainTestCase.java
```diff
@@ -447,4 +448,225 @@ public void testLegacyEnableSingleAppDeployment() throws CommandLineException, I
...
+     * Separate test for check status of issue WFCORE-3562 by Aesh command.
```
jfdenise: "Separate test for check status of issue WFCORE-3562 by Aesh command." Is it really required?

testsuite/domain/src/test/java/org/jboss/as/test/integration/domain/management/cli/DeployAllDomainTestCase.java
```diff
@@ -447,4 +448,225 @@ public void testLegacyEnableSingleAppDeployment() throws CommandLineException, I
...
+     * Concentration is for disabling all deployments with all group by Aesh commands.
```
jfdenise: Is this sentence really required?

testsuite/domain/src/test/java/org/jboss/as/test/integration/domain/management/cli/DeployAllDomainTestCase.java
```diff
@@ -447,4 +448,225 @@ public void testLegacyEnableSingleAppDeployment() throws CommandLineException, I
...
+    @Test
+    public void testDisableAllAppDeploymentAllGroup() throws CommandLineException, IOException {
+        // Step 1) Deploy applications deployments to defined server groups
+        ctx.handle("deployment deploy-file --server-groups=" + sgOne + ' ' + cliTestApp1War.getAbsolutePath());
+        ctx.handle("deployment deploy-file --server-groups=" + sgOne + ' ' + cliTestAnotherWar.getAbsolutePath());
+        ctx.handle("deployment deploy-file --server-groups=" + sgTwo + ' ' + cliTestApp2War.getAbsolutePath());
+        ctx.handle("deployment deploy-file --server-groups=" + sgTwo + ',' + sgOne + ' ' + cliTestAppEar.getAbsolutePath());
+
+
+        // Step 2) Disabling two selected applications deployments
```
jfdenise: "applications deployments" ==> "deployments". It applies to some other locations in this test.

testsuite/domain/src/test/java/org/jboss/as/test/integration/domain/management/cli/DeployAllDomainTestCase.java
```diff
@@ -447,4 +448,225 @@ public void testLegacyEnableSingleAppDeployment() throws CommandLineException, I
...
+        ctx.handle("deployment disable --server-groups=" + sgOne + ' ' + cliTestApp1War.getName());
+        ctx.handle("deployment disable --server-groups=" + sgTwo + ' ' + cliTestApp2War.getName());
+
+        // Step 3) Disable all deployed applications deployments in all server groups
+        ctx.handle("deployment disable-all --all-relevant-server-groups");
+
+        // Step 4) Check if all applications deployments is disabled in all server groups
```
jfdenise: "is disabled" ==> "are disabled"

testsuite/domain/src/test/java/org/jboss/as/test/integration/domain/management/cli/DeployAllDomainTestCase.java
```diff
@@ -447,4 +448,225 @@ public void testLegacyEnableSingleAppDeployment() throws CommandLineException, I
...
+
+        result = deploymentInfo(cli, sgTwo);
+        checkExist(result, cliTestApp1War.getName(), NOT_ADDED, ctx);
+        checkExist(result, cliTestAnotherWar.getName(), NOT_ADDED, ctx);
+        checkExist(result, cliTestApp2War.getName(), ADDED, ctx);
+        checkExist(result, cliTestAppEar.getName(), ADDED, ctx);
+    }
+
+    /**
+     * Disabling all deployed application deployments by only one server group from two server groups.
```
jfdenise: "Disabling all deployed application deployments by only one server group from two server groups." Seems that your test is doing: "Disabling all deployments of one server group that are also present in a second server group."

Some comments I made in previous test also apply to the content of this test.

testsuite/domain/src/test/java/org/jboss/as/test/integration/domain/management/cli/DeployAllDomainTestCase.java
```diff
@@ -447,4 +448,225 @@ public void testLegacyEnableSingleAppDeployment() throws CommandLineException, I
...
+        result = deploymentInfo(cli, sgTwo);
+        checkExist(result, cliTestApp1War.getName(), NOT_ADDED, ctx);
+        checkExist(result, cliTestAnotherWar.getName(), NOT_ADDED, ctx);
+        checkExist(result, cliTestApp2War.getName(), ADDED, ctx);
+        checkExist(result, cliTestAppEar.getName(), ADDED, ctx);
+    }
+
+    /**
+     * Enabling one application deployment.
+     * Separate test for check status of issue WFCORE-3562 by Aesh command.
```
jfdenise: Are these 2 sentences needed?

testsuite/domain/src/test/java/org/jboss/as/test/integration/domain/management/cli/DeployAllDomainTestCase.java
vmarekrh: Done

testsuite/domain/src/test/java/org/jboss/as/test/integration/domain/management/cli/DeployAllDomainTestCase.java
vmarekrh: Removed

testsuite/domain/src/test/java/org/jboss/as/test/integration/domain/management/cli/DeployAllDomainTestCase.java
vmarekrh: No, this sentence isn't required. Removed

testsuite/domain/src/test/java/org/jboss/as/test/integration/domain/management/cli/DeployAllDomainTestCase.java
vmarekrh: Is just the word 'deployments'? Ok, replaced.

testsuite/domain/src/test/java/org/jboss/as/test/integration/domain/management/cli/DeployAllDomainTestCase.java
vmarekrh: Done

testsuite/domain/src/test/java/org/jboss/as/test/integration/domain/management/cli/DeployAllDomainTestCase.java
vmarekrh: Thank you, your explanation in a comment are better than mine. Replaced

testsuite/domain/src/test/java/org/jboss/as/test/integration/domain/management/cli/DeployAllDomainTestCase.java
vmarekrh: No, it isn't needed. Removed

xstefank:
retest this please

PR #3145, opened 2018-02-26
Fix for WFCORE-3647, CLI, Cygwin issue when paging output

jfdenise:
- Guarantee that TerminalConnection thread is put in wait state synchronously. This is good in general (although not fully required, no side effect observed when putting the state asynchronously) except on Cygwin platform on which interacting with the terminal (eg: calling stty -a) while the connection is reading on the terminal gets you a blocking sub process outputstream (that has been reproduced outside any CLI/aesh context).

- Found a left over unrelated to the original issue, set back original attributes.

That is tested by MoreTestCase run on all platforms (Cygwin is used by QE).


jmesnil:
The job for Full Integration - Linux passed at https://ci.wildfly.org/viewLog.html?buildId=92182&tab=buildResultsDiv&buildTypeId=WildFlyCore_PullRequest_WildFlyCoreFullIntegration

jmesnil:
retest this please

PR #3146, opened 2018-02-27
[WFCORE-3656] Replace commons-logging implementation.

jamezp:
https://issues.jboss.org/browse/WFCORE-3656

Analysis document: https://github.com/wildfly/wildfly-proposals/pull/29

The first commit simply adds tests for each supported logging API to test against logging profiles. The initial change of the Apache Commons Logging implementation to a new which delegates to JBoss Log Manager was initiated by commons-logging not working with logging profiles.

The next commit replaces the `org.jboss.logmanager:commons-logging-jboss-logmanager` implementation with `org.jboss.logging:commons-logging-jboss-logging` which delegates to JBoss Logging. There was an issue using the embedded API with Artemis if the JBoss Log Manager was not the default log manager. This prompted a change to delegate commons-logging to JBoss Logging instead which allows for common log managers to be used.

There are two minor differences in the implementations. With the new implementation a new logger is created each time a `LogFactory.getLog()` is invoked. In most cases this should be okay as loggers should generally be static or not created all that often.

The second difference is the `LogFactory` attributes are effectively static and will be shared across all log contexts. This is likely not a huge issue as these attributes are probably not used all that often.

### DIFF
--- a/core-feature-pack/pom.xml
+++ b/core-feature-pack/pom.xml
@@ -152,8 +152,8 @@
         </dependency>
 
         <dependency>
-            <groupId>org.jboss.logmanager</groupId>
-            <artifactId>commons-logging-jboss-logmanager</artifactId>
+            <groupId>org.jboss.logging</groupId>
+            <artifactId>commons-logging-jboss-logging</artifactId>
         </dependency>
 
         <dependency>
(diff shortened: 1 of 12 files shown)


jamezp:
Oops these failures look like they're real issues. I must have messed something up when I attempted to rebase on master.

jamezp:
https://github.com/wildfly/wildfly/pull/10947 is required to make the full builds work.

jamezp:
retest this please

PR #3147, opened 2018-02-27
[WFCORE-3579] The embed-host-controller command should check for boot…

bstansberry:
… completion even in the empty config case.

Checking the host-state is not necessary any more as WFCORE-3579 means the read of the local host name will be rejected until the HC is booted.

https://issues.jboss.org/browse/WFCORE-3579

See also the comments on https://issues.jboss.org/browse/WFCORE-3655

@luck3y @jmesnil FYI / approval.


bstansberry:
Closing; it produced hangs in TS. Too late in the 4.x cycle for something that does that.

PR #3155, opened 2018-03-06
[WFCORE-3411] CLI operation completion issues.

MMarus:
WFCORE: https://issues.jboss.org/browse/WFCORE-3411

I have encountered another completion issue which I described in first comment of the WFCORE-3411. Following lines are related to the new mentioned issue https://github.com/wildfly/wildfly-core/compare/master...MMarus:WFCORE-3411-CLI_operation_completion_issues?expand=1#diff-b5ac7c9723bb46958c770f161f4759c8R432 . I can move this into separate WFCORE issue and PR if needed.

### DIFF
--- a/cli/src/main/java/org/jboss/as/cli/operation/OperationRequestCompleter.java
+++ b/cli/src/main/java/org/jboss/as/cli/operation/OperationRequestCompleter.java
@@ -396,7 +396,7 @@ public class OperationRequestCompleter implements CommandLineCompleter {
                 if (argFullName.equals(argName)) {
                     if (!arg.isValueRequired()) {
                         candidates.add(Util.FALSE);
-                        return parsedCmd.getLastChunkIndex();
+                        return buffer.length();
                     }
                 }
             }
@@ -577,11 +577,14 @@ public class OperationRequestCompleter implements CommandLineCompleter {
                 // enough.
                 if (!parsedCmd.isLastPropertyNegated()) {
                     candidates.add("=" + Util.FALSE);
+                    result = recalculateResult(parsedCmd, candidates, chunk, result);
                 }
                 if (format != null && format.getPropertyListEnd() != null && format.getPropertyListEnd().length() > 0) {
                     candidates.add(format.getPropertyListEnd());
+                    result = recalculateResult(parsedCmd, candidates, chunk, result);
                     if (!allPropertiesPresent) {
                         candidates.add(format.getPropertySeparator());
+                        result = recalculateResult(parsedCmd, candidates, chunk, result);
                     }
                 }
             }
@@ -592,10 +595,20 @@ public class OperationRequestCompleter implements CommandLineCompleter {
                 final CommandLineFormat format = parsedCmd.getFormat();
                 if (format != null && format.getPropertyListEnd() != null && format.getPropertyListEnd().length() > 0) {
                     candidates.add(format.getPropertyListEnd());
+                    result = recalculateResult(parsedCmd, candidates, chunk, result);
                 }
             }
         } else {
             Collections.sort(candidates);
+        }
+        return result;
+    }
+
+    private int recalculateResult(ParsedCommandLine parsedCmd, List<String> candidates, String chunk, int result) {
+        if (candidates.size() == 1 && chunk != null) {
+            // Move the offset to the end of the line, since the name of the last property is completely specified
+            // and there are no other candidates
+            result = parsedCmd.getLastChunkIndex() + chunk.length();
         }
         return result;
     }
--- a/cli/src/main/java/org/jboss/as/cli/operation/impl/DefaultOperationCandidatesProvider.java
+++ b/cli/src/main/java/org/jboss/as/cli/operation/impl/DefaultOperationCandidatesProvider.java
@@ -409,8 +409,6 @@ public class DefaultOperationCandidatesProvider implements OperationCandidatesPr
             throw new IllegalStateException("Failed to build operation", e1);
         }
 
-        final Map<String,CommandLineCompleterFactory> globalOpProps = globalOpPropCompleters.get(operationName);
-
         List<CommandArgument> result;
         try {
             ModelNode outcome = client.execute(request);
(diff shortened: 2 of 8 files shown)


cli/src/main/java/org/jboss/as/cli/operation/OperationRequestCompleter.java
```diff
@@ -592,6 +601,9 @@ private int completeArgumentValueAndPropertyNames(CommandContext ctx, ParsedComm
                 final CommandLineFormat format = parsedCmd.getFormat();
                 if (format != null && format.getPropertyListEnd() != null && format.getPropertyListEnd().length() > 0) {
                     candidates.add(format.getPropertyListEnd());
+                    if (candidates.size() == 1) {
+                        result = parsedCmd.getLastChunkIndex() + chunk.length();
+                    }
```
xstefank: this can probably be refactored to a separate method

cli/src/main/java/org/jboss/as/cli/operation/impl/DefaultOperationCandidatesProvider.java
```diff
@@ -428,8 +428,22 @@ private boolean canAppearNext(Property prop) {
...
+                    if(ctx.getParsedCommandLine().isLastPropertyNegated()) {
+                        for (Property prop : propList) {
+                            if(radical.equals(prop.getName())){
+                                radical = null;
+                                break;
+                            }
+                        }
+                    }
+                } else {
+                    radical = null;
```
xstefank: not necessary

cli/src/main/java/org/jboss/as/cli/operation/OperationRequestCompleter.java
```diff
@@ -592,6 +601,9 @@ private int completeArgumentValueAndPropertyNames(CommandContext ctx, ParsedComm
                 final CommandLineFormat format = parsedCmd.getFormat();
                 if (format != null && format.getPropertyListEnd() != null && format.getPropertyListEnd().length() > 0) {
                     candidates.add(format.getPropertyListEnd());
+                    if (candidates.size() == 1) {
```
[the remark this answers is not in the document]
MMarus: Yes, You are right. I have already extracted this part of code, which repeats 4 times in the code into a separate method. Should I change this part to "result = parsedCmd.getLastChunkIndex() + chunk.length();" or just leave the method call as it is now in the https://github.com/wildfly/wildfly-core/pull/3155/commits/11911af7eea6dfd89c8cb2f843524a8c75230282 commit ?

cli/src/main/java/org/jboss/as/cli/operation/OperationRequestCompleter.java
```diff
@@ -592,6 +601,9 @@ private int completeArgumentValueAndPropertyNames(CommandContext ctx, ParsedComm
...
+                        result = parsedCmd.getLastChunkIndex() + chunk.length();
```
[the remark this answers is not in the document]
MMarus: It should be ensured by line https://github.com/wildfly/wildfly-core/commit/11911af7eea6dfd89c8cb2f843524a8c75230282#diff-f91220412cef158ff6623a2dccc627a4R462 
where the lastArg is only populated in case the FullName of the arg is not null (.equals would fail with NPE if it was called on null) and is equal to the chunk and the line
https://github.com/wildfly/wildfly-core/commit/11911af7eea6dfd89c8cb2f843524a8c75230282#diff-f91220412cef158ff6623a2dccc627a4R554
where it is checked that lastArg is not null.

However, the lastArg and chunk are mutable, therefore in case there would be committed a code which assigns a null to either of them, this indirect "check" would fail.
Do You think it would be better to add a null check of the chunk in the new method recalculateResult() commited in 11911af ?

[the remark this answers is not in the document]
jfdenise: It would be nice to add a check and a comment.

cli/src/main/java/org/jboss/as/cli/operation/OperationRequestCompleter.java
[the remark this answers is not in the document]
jfdenise: you can keep the method.

cli/src/main/java/org/jboss/as/cli/operation/impl/DefaultOperationCandidatesProvider.java
```diff
@@ -428,8 +428,20 @@ private boolean canAppearNext(Property prop) {
                 final List<Property> propList = reqProps.asPropertyList();
                 result = new ArrayList<CommandArgument>(propList.size());
 
-                String radical = ctx.getParsedCommandLine().getLastParsedPropertyValue() == null
-                        ? ctx.getParsedCommandLine().getLastParsedPropertyName() : null;
+                String radical = null;
+                if(ctx.getParsedCommandLine().getLastParsedPropertyValue() == null) {
+                    radical = ctx.getParsedCommandLine().getLastParsedPropertyName();
+                    //Check if the property is completely specified and is negated
```
jfdenise: Why do you need to set the radical to null in this case? Would you have an example?

MMarus: The methond canAppearNext() [1] returns false when the property is already in the input line or when the name does not start with radical. That means if the operation e.g. :reload-servers has 2 possible properties "blocking" and "start-mode" then in case the "!blocking" is specified the radical would be set to "blocking" and the  canAppearNext() would return false for every one command, since the "blocking" is already  in the input and the other properties don't start with "blocking" string.

In the old code the radical was set to null for example in this case ":reload-servers(blocking=false<TAB>" because the last property blocking has the value false and getLastParsedPropertyValue() returned "false". 
However, In case  ":reload-servers(!blocking<TAB>" which has similar meaning as the first one, the radical was set to "blocking". This was caused because the getLastParsedPropertyValue() returned null which I think is right, since there is no explicit specified value like in the first case the "=false". Therefore when the radical was set to null in the first case I think we could do the same for the case when the property is negated and its name is completely specified e.g. "!blocking".

[1] https://github.com/MMarus/wildfly-core/blob/11911af7eea6dfd89c8cb2f843524a8c75230282/cli/src/main/java/org/jboss/as/cli/operation/impl/DefaultOperationCandidatesProvider.java#L378

cli/src/main/java/org/jboss/as/cli/operation/OperationRequestCompleter.java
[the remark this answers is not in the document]
MMarus: Thank you very much, I have added them.

cli/src/main/java/org/jboss/as/cli/operation/impl/DefaultOperationCandidatesProvider.java
jfdenise: Thank-you for taking the time to detail the case. You are right, you perhaps have to take into account a corner case, let's imagine we have blocking and blocking-param options.
reload:(!blocking<TAB> ==> should propose "," and blocking-param. This is something that you should be able to reproduce in your new cli test.
BTW, when making changes to completion it is good to extend the test https://github.com/wildfly/wildfly-core/blob/master/testsuite/domain/src/test/java/org/jboss/as/test/integration/domain/management/cli/CliCompletionTestCase.java

jmesnil:
retest this please

cli/src/main/java/org/jboss/as/cli/operation/impl/DefaultOperationCandidatesProvider.java
MMarus: I just looked into the test *PropertiesCompletionTestCase* again and the *DefaultOperationCandidatesProvider* is not tested there, since there is a *MockOperationCandidatesProvider* used instead. I think it's because the method *getProperties()* first creates and executes request in order to get list of properties which wouldn't be easy to mock. However, in order to create the resulting list of  *CommandArgument* only the *Property* list is needed (line 429-524 [1]) which could be very easily moved to separate method and it would allow to test this functionality.

Do You think I can move that part into separate method or is it too big change ?
Thank You for answer.

[1] https://github.com/MMarus/wildfly-core/blob/80f7c778602d3be0b05d0cc89d20a7efd4c89c0b/cli/src/main/java/org/jboss/as/cli/operation/impl/DefaultOperationCandidatesProvider.java#L429

jfdenise: @MMarus , +1 to increase unit-testability of completion!

MMarus:
I have moved (lines 429-524 [1]) to separate method and added the tests mentioned in the comments. 

 For the corner case,when we have blocking and blocking-param options I have created 2 separate tests to test the relevant logic in relevant classes (*PropertiesCompletionTestCase* and *DefaultOperationCandidatesProviderTestCase*).

[1] https://github.com/MMarus/wildfly-core/blob/80f7c778602d3be0b05d0cc89d20a7efd4c89c0b/cli/src/main/java/org/jboss/as/cli/operation/impl/DefaultOperationCandidatesProvider.java#L429

MMarus:
retest this please

jmesnil:
@jfdenise are you ok with the changes you requested?

PR #3158, opened 2018-03-07
[WFCORE-3670] load correct additional module with alias name.

soul2zimate:
https://issues.jboss.org/browse/WFCORE-3670 
Reproducer in https://issues.jboss.org/browse/JBEAP-14310 shows that a sub-deployment with dependency referenced to module alias name can't be properly loaded.

As spec from jboss-deployment-structure.xml parsed values are unmodifiable, this fix just replace module alias name with found real module identifier, and make sure it will later load that correct module as dependency.

WFLY test case https://github.com/soul2zimate/wildfly/commit/438f85a71316838fe2a8815d772e7886456e3a23

### DIFF
--- a/server/src/main/java/org/jboss/as/server/deployment/module/descriptor/DeploymentStructureDescriptorParser.java
+++ b/server/src/main/java/org/jboss/as/server/deployment/module/descriptor/DeploymentStructureDescriptorParser.java
@@ -30,12 +30,10 @@ import java.io.InputStream;
 import java.util.ArrayList;
 import java.util.Collection;
 import java.util.HashMap;
-import java.util.HashSet;
 import java.util.List;
 import java.util.ListIterator;
 import java.util.Map;
 import java.util.Map.Entry;
-import java.util.Set;
 
 import javax.xml.namespace.QName;
 import javax.xml.stream.XMLInputFactory;
@@ -57,6 +55,7 @@ import org.jboss.as.server.deployment.SubDeploymentMarker;
 import org.jboss.as.server.deployment.annotation.ResourceRootIndexer;
 import org.jboss.as.server.deployment.jbossallxml.JBossAllXmlParserRegisteringProcessor;
 import org.jboss.as.server.deployment.module.AdditionalModuleSpecification;
+import org.jboss.as.server.deployment.module.ModuleDependency;
 import org.jboss.as.server.deployment.module.ModuleRootMarker;
 import org.jboss.as.server.deployment.module.ModuleSpecification;
 import org.jboss.as.server.deployment.module.ResourceRoot;
@@ -131,9 +130,9 @@ public class DeploymentStructureDescriptorParser implements DeploymentUnitProces
             //if the parent has already attached parsed data for this sub deployment we need to process it
             if (deploymentRoot.hasAttachment(SUB_DEPLOYMENT_STRUCTURE)) {
                 final ModuleSpecification subModuleSpec = deploymentUnit.getAttachment(Attachments.MODULE_SPECIFICATION);
-                Set<ModuleIdentifier> additionalModules = new HashSet<>();
+                Map<ModuleIdentifier, AdditionalModuleSpecification> additionalModules = new HashMap<>();
                 for(AdditionalModuleSpecification i : deploymentUnit.getParent().getAttachmentList(Attachments.ADDITIONAL_MODULES)) {
-                    additionalModules.add(i.getModuleIdentifier());
+                    additionalModules.put(i.getModuleIdentifier(), i);
                 }
                 handleDeployment(phaseContext, deploymentUnit, subModuleSpec, deploymentRoot.getAttachment(SUB_DEPLOYMENT_STRUCTURE), additionalModules);
             }
@@ -166,7 +165,7 @@ public class DeploymentStructureDescriptorParser implements DeploymentUnitProces
                 result = parse(deploymentFile.getPhysicalFile(), deploymentUnit, moduleLoader);
             }
             // handle additional modules
-            Set<ModuleIdentifier> additionalModuleSet = new HashSet<>();
+            Map<ModuleIdentifier, AdditionalModuleSpecification> additionalModules = new HashMap<>();
             for (final ModuleStructureSpec additionalModule : result.getAdditionalModules()) {
                 for (final ModuleIdentifier identifier : additionalModule.getAnnotationModules()) {
                     //additional modules don't support annotation imports
@@ -182,10 +181,10 @@ public class DeploymentStructureDescriptorParser implements DeploymentUnitProces
                         itr.remove();
                     }
                 }
-                additionalModuleSet.add(additionalModule.getModuleIdentifier());
                 final AdditionalModuleSpecification additional = new AdditionalModuleSpecification(additionalModule.getModuleIdentifier(), additionalModuleResourceRoots);
                 additional.addAliases(additionalModule.getAliases());
                 additional.addSystemDependencies(additionalModule.getModuleDependencies());
+                additionalModules.put(additional.getModuleIdentifier(), additional);
                 deploymentUnit.addToAttachmentList(Attachments.ADDITIONAL_MODULES, additional);
                 for (final ResourceRoot root : additionalModuleResourceRoots) {
                     ResourceRootIndexer.indexResourceRoot(root);
@@ -204,7 +203,7 @@ public class DeploymentStructureDescriptorParser implements DeploymentUnitProces
             // handle the the root deployment
             final ModuleStructureSpec rootDeploymentSpecification = result.getRootDeploymentSpecification();
             if (rootDeploymentSpecification != null) {
-                handleDeployment(phaseContext, deploymentUnit, moduleSpec, rootDeploymentSpecification, additionalModuleSet);
+                handleDeployment(phaseContext, deploymentUnit, moduleSpec, rootDeploymentSpecification, additionalModules);
             }
             // handle sub deployments
             final Map<String, ResourceRoot> subDeploymentMap = new HashMap<String, ResourceRoot>();
@@ -238,9 +237,30 @@ public class DeploymentStructureDescriptorParser implements DeploymentUnitProces
         }
     }
 
-    private void handleDeployment(final DeploymentPhaseContext phaseContext, final DeploymentUnit deploymentUnit, final ModuleSpecification moduleSpec, final ModuleStructureSpec rootDeploymentSpecification, Set<ModuleIdentifier> additionalModules) throws DeploymentUnitProcessingException {
+    private void handleDeployment(final DeploymentPhaseContext phaseContext, final DeploymentUnit deploymentUnit, final ModuleSpecification moduleSpec, final ModuleStructureSpec rootDeploymentSpecification, final Map<ModuleIdentifier, AdditionalModuleSpecification> additionalModules) throws DeploymentUnitProcessingException {
         final Map<VirtualFile, ResourceRoot> resourceRoots = resourceRoots(deploymentUnit);
-        moduleSpec.addUserDependencies(rootDeploymentSpecification.getModuleDependencies());
+        // handle unmodifiable module dependencies with alias
+        List<ModuleDependency> moduleDependencies = rootDeploymentSpecification.getModuleDependencies();
+        List<ModuleDependency> aliasDependencies = new ArrayList<>();
+
+        // Pre-index additionalmodules
+        Map<ModuleIdentifier, AdditionalModuleSpecification> index = new HashMap<>();
+
+        // Only iterate additionalModules once, instead of once per dependency
+        for (AdditionalModuleSpecification module : additionalModules.values()) {
+            for (ModuleIdentifier alias : module.getAliases()) {
+                index.put(alias, module);
+            }
+        }
+        // No more nested loop
+        for (ModuleDependency dependency : moduleDependencies) {
+            ModuleIdentifier identifier = dependency.getIdentifier();
+            if (index.containsKey(identifier)) {
+                aliasDependencies.add(new ModuleDependency(dependency.getModuleLoader(), index.get(identifier).getModuleIdentifier(), dependency.isOptional(), dependency.isExport(), dependency.isImportServices(), dependency.isUserSpecified()));
+            }
+        }
+
+        moduleSpec.addUserDependencies(moduleDependencies);
         moduleSpec.addExclusions(rootDeploymentSpecification.getExclusions());
         moduleSpec.addAliases(rootDeploymentSpecification.getAliases());
         moduleSpec.addModuleSystemDependencies(rootDeploymentSpecification.getSystemDependencies());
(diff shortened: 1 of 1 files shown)


soul2zimate:
investigate test failures.

soul2zimate:
ReverseProxyTestCase, RootContextEarUnitTestCase and RootContextWarUnitTestCase These failures appear also in other PR.

server/src/main/java/org/jboss/as/server/deployment/module/descriptor/DeploymentStructureDescriptorParser.java
```diff
@@ -131,9 +133,9 @@ public void deploy(final DeploymentPhaseContext phaseContext) throws DeploymentU
             //if the parent has already attached parsed data for this sub deployment we need to process it
             if (deploymentRoot.hasAttachment(SUB_DEPLOYMENT_STRUCTURE)) {
                 final ModuleSpecification subModuleSpec = deploymentUnit.getAttachment(Attachments.MODULE_SPECIFICATION);
-                Set<ModuleIdentifier> additionalModules = new HashSet<>();
+                Set<AdditionalModuleSpecification> additionalModules = new HashSet<>();
```
dmlloyd: ``AdditionalModuleSpecification`` does not have equals/hashcode so this is not semantically equivalent to the code that came before.  Can you explain if that is OK and if so, why?

server/src/main/java/org/jboss/as/server/deployment/module/descriptor/DeploymentStructureDescriptorParser.java
```diff
@@ -238,9 +240,25 @@ public void deploy(final DeploymentPhaseContext phaseContext) throws DeploymentU
...
-    private void handleDeployment(final DeploymentPhaseContext phaseContext, final DeploymentUnit deploymentUnit, final ModuleSpecification moduleSpec, final ModuleStructureSpec rootDeploymentSpecification, Set<ModuleIdentifier> additionalModules) throws DeploymentUnitProcessingException {
+    private void handleDeployment(final DeploymentPhaseContext phaseContext, final DeploymentUnit deploymentUnit, final ModuleSpecification moduleSpec, final ModuleStructureSpec rootDeploymentSpecification, final Set<AdditionalModuleSpecification> additionalModules) throws DeploymentUnitProcessingException {
         final Map<VirtualFile, ResourceRoot> resourceRoots = resourceRoots(deploymentUnit);
-        moduleSpec.addUserDependencies(rootDeploymentSpecification.getModuleDependencies());
+        // handle unmodifiable module dependencies with alias
+        List<ModuleDependency> moduleDependencies = rootDeploymentSpecification.getModuleDependencies();
+        List<ModuleDependency> copy = new ArrayList<>();
+        List<ModuleDependency> aliasDependencies = new ArrayList<>();
+        copy.addAll(moduleDependencies);
+        Iterator<ModuleDependency> it = copy.iterator();
```
dmlloyd: Minor comment, but this could be a bit more readable as a for-each.

server/src/main/java/org/jboss/as/server/deployment/module/descriptor/DeploymentStructureDescriptorParser.java
```diff
@@ -278,6 +303,17 @@ private void handleDeployment(final DeploymentPhaseContext phaseContext, final D
         }
     }
 
+    private boolean moduleIncluded(final Set<AdditionalModuleSpecification> additionalModules, final ModuleIdentifier dependency) {
+        Iterator<AdditionalModuleSpecification> it = additionalModules.iterator();
+        while (it.hasNext()) {
+            ModuleIdentifier identifier = it.next().getModuleIdentifier();
+            if (identifier.equals(dependency)) {
```
dmlloyd: The fact we're iterating this makes me wonder: why is it a set?

server/src/main/java/org/jboss/as/server/deployment/module/descriptor/DeploymentStructureDescriptorParser.java
```diff
@@ -238,9 +240,25 @@ public void deploy(final DeploymentPhaseContext phaseContext) throws DeploymentU
...
+            ModuleDependency dependency = it.next();
+            ModuleIdentifier identifier = dependency.getIdentifier();
+            for (AdditionalModuleSpecification module : additionalModules) {
+                if (module.getAliases().contains(identifier)) {
+                    aliasDependencies.add(new ModuleDependency(dependency.getModuleLoader(), module.getModuleIdentifier(), dependency.isOptional(), dependency.isExport(), dependency.isImportServices(), dependency.isUserSpecified()));
+                }
+            }
+        }
+        copy.addAll(aliasDependencies);
+        moduleSpec.addUserDependencies(copy);
```
dmlloyd: This all seems a bit overly complicated; what is this code trying to accomplish?

server/src/main/java/org/jboss/as/server/deployment/module/descriptor/DeploymentStructureDescriptorParser.java
soul2zimate: By changing `Set<ModuleIdentifier>` to `Set<AdditionalModuleSpecification>`? I think it's fine. https://github.com/wildfly/wildfly-core/pull/2786/files#diff-a84c1903d76fdbbf52a51bfdf6c2fe4f changed `handleDeployment`() method signature to add Set<ModuleIdentifier> additionalModules which is only used in

` if(dependency.getName().startsWith(ServiceModuleLoader.MODULE_PREFIX) && !additionalModules.contains(dependency)) {`

to check if `ModuleIdentifier` dependency is included in `additionalModules`. I update to `AdditionalModuleSpecification` as I need to use the alias list to compare. Method `moduleIncluded()` covers the same check on ModuleIdentifier as before

server/src/main/java/org/jboss/as/server/deployment/module/descriptor/DeploymentStructureDescriptorParser.java
soul2zimate: changed to use for-each.

server/src/main/java/org/jboss/as/server/deployment/module/descriptor/DeploymentStructureDescriptorParser.java
soul2zimate: Iterator is not necessary, I changed to for-each.
This only adds real module dependency to moduleSpec if matched alias is found.

server/src/main/java/org/jboss/as/server/deployment/module/descriptor/DeploymentStructureDescriptorParser.java
soul2zimate: same as replaced Set<ModuleIdentifier> which contains no duplicate elements (module identifier).

jmesnil:
retest this please

soul2zimate:
retest this please

soul2zimate:
retest this please

server/src/main/java/org/jboss/as/server/deployment/module/descriptor/DeploymentStructureDescriptorParser.java
dmlloyd: OK, but, this wasn't an opinion question. :)  What makes this OK?  Is it a guarantee that every AdditionalModuleSpecification will have a unique identifier, for example?

dmlloyd: It seems like it might want to be a list instead of a set.

server/src/main/java/org/jboss/as/server/deployment/module/descriptor/DeploymentStructureDescriptorParser.java
```diff
@@ -238,9 +239,20 @@ public void deploy(final DeploymentPhaseContext phaseContext) throws DeploymentU
...
+    private void handleDeployment(final DeploymentPhaseContext phaseContext, final DeploymentUnit deploymentUnit, final ModuleSpecification moduleSpec, final ModuleStructureSpec rootDeploymentSpecification, final Set<AdditionalModuleSpecification> additionalModules) throws DeploymentUnitProcessingException {
         final Map<VirtualFile, ResourceRoot> resourceRoots = resourceRoots(deploymentUnit);
-        moduleSpec.addUserDependencies(rootDeploymentSpecification.getModuleDependencies());
+        // handle unmodifiable module dependencies with alias
+        List<ModuleDependency> moduleDependencies = rootDeploymentSpecification.getModuleDependencies();
+        List<ModuleDependency> aliasDependencies = new ArrayList<>();
+        for (ModuleDependency dependency : moduleDependencies) {
+            ModuleIdentifier identifier = dependency.getIdentifier();
+            for (AdditionalModuleSpecification module : additionalModules) {
+                if (module.getAliases().contains(identifier)) {
```
dmlloyd: Is module.getAliases() as Set as well?  If not, this contains operation could get somewhat expensive in some cases.

server/src/main/java/org/jboss/as/server/deployment/module/descriptor/DeploymentStructureDescriptorParser.java
soul2zimate: I have updated Set to List.
There is no guarantee at that point right after xml parsing. User can write same identifier for two modules in config file, but if there is either duplicate module-identifier or duplicate module alias e.g. https://gist.github.com/soul2zimate/b2d9bf06f68137c8271db7aaea4c8054, it fails at service installation due to [DuplicateServiceException](https://gist.github.com/soul2zimate/fbcf0cda9b3c4fbd140aaee1e622fbca) 
It is incorrect to contain a duplicate module identifier, Should it fail earlier here?

soul2zimate:
retest this please

bmaxwell:
retest this please!

soul2zimate:
retest this please

darranl:
@dmlloyd would it be possible for you to please have another look at this one?  I think your comments have made everyone nervous of approving.

server/src/main/java/org/jboss/as/server/deployment/module/descriptor/DeploymentStructureDescriptorParser.java
soul2zimate: A Map<ModuleIdentifier, AdditionalModuleSpecification> doesn't help to eliminate it. It doesn't compare with map key, it needs to compare module dependency's identifier with each map value additional module's aliases something like:
```
List<ModuleDependency> moduleDependencies = rootDeploymentSpecification.getModuleDependencies();
        List<ModuleDependency> aliasDependencies = new ArrayList<>();
        for (ModuleDependency dependency : moduleDependencies) {
            ModuleIdentifier identifier = dependency.getIdentifier();
            for (AdditionalModuleSpecification module : additionalModules.values()) {
                if (module.getAliases().contains(identifier)) {
                    aliasDependencies.add(new ModuleDependency(dependency.getModuleLoader(), module.getModuleIdentifier(), dependency.isOptional(), dependency.isExport(), dependency.isImportServices(), dependency.isUserSpecified()));
                }
            }
        }
```

soul2zimate: Hi @dmlloyd IIUC, we need to compare module dependency identifier with additional module aliases. A map with ModuleIdentifier as key does not eliminate the linear searches.

dmlloyd: You're iterating all of `additionalModules.values()` for every single value in `moduleDependencies` looking for the given `dependency`, and you're iterating every alias looking for `identifier`.  To reduce a triply-nested loop to a doubly-nested loop, you could write it like this:

```java
        List<ModuleDependency> moduleDependencies = rootDeploymentSpecification.getModuleDependencies();
        List<ModuleDependency> aliasDependencies = new ArrayList<>();
        // Pre-index additionalmodules
        Map<ModuleIdentifier, ModuleDependency> index = new HashMap<>();
        // Only iterate additionalModules once, instead of once per dependency
        for (AdditionalModuleSpecification module : additionalModules.values()) {
            for (ModuleIdentifier alias : module.getAliases()) {
                index.put(alias, module);
            }
        }
        // No more nested loop
        for (ModuleDependency dependency : moduleDependencies) {
            ModuleIdentifier identifier = dependency.getIdentifier();
            if (index.containsKey(identifier)) {
                aliasDependencies.add(new ModuleDependency(dependency.getModuleLoader(), module.getModuleIdentifier(), dependency.isOptional(), dependency.isExport(), dependency.isImportServices(), dependency.isUserSpecified()));
            }
        }
```

The `index` could probably even be a simple `Set` because it doesn't look as if the corresponding value is even used by the existing code (you might want to double-check that for correctness because it looks a little wrong to me).

xstefank:
@soul2zimate the CI failures looks strange. Maybe try to force push once more?

server/src/main/java/org/jboss/as/server/deployment/module/descriptor/DeploymentStructureDescriptorParser.java
soul2zimate: Thanks for the review and kind explanation
The corresponding values in map are needed in order to obtain real ModuleIdentifier so we can add it into aliasDependencies like:

```
            if (index.containsKey(identifier)) {
                aliasDependencies.add(new ModuleDependency(dependency.getModuleLoader(), index.get(identifier).getModuleIdentifier(), dependency.isOptional(), dependency.isExport(), dependency.isImportServices(), dependency.isUserSpecified()));
            }
```

I pushed the change again, please review at your convenience. Once this is approved,  I will squash commits together. Thanks.

dmlloyd: Looks OK to me.

xstefank:
@dmlloyd CI passed, can you take a look again, please?

dmlloyd:
My original review was from over a year ago; I finished it off, but subsequently, someone else should be handling the review of this.

soul2zimate:
Thanks @dmlloyd 
Hi @wildfly-review  can someone give a review on this please ?

xstefank:
not sure who wildfly-review is :) @jmesnil @bstansberry @jamezp can any of you take a look, please?

kabir:
@soul2zimate @xstefank the correct alias is @wildfly/mergers-core (-full for wildfly/wildfly)

xstefank:
@wildfly/mergers-core can someone take a look, please?

server/src/main/java/org/jboss/as/server/deployment/module/descriptor/DeploymentStructureDescriptorParser.java
```diff
@@ -263,12 +283,19 @@ private void handleDeployment(final DeploymentPhaseContext phaseContext, final D
...
+            ModuleIdentifier identifier = dependency;
+            for (AdditionalModuleSpecification module : additionalModules.values()) {
+                if (module.getAliases().contains(dependency)) {
+                    identifier = module.getModuleIdentifier();
+                    break;
+                }
+            }
+            deploymentUnit.addToAttachmentList(Attachments.ADDITIONAL_ANNOTATION_INDEXES, identifier);
+            // additional modules will not be created till much later, a dep on them would fail
+            if (identifier.getName().startsWith(ServiceModuleLoader.MODULE_PREFIX) && !moduleIncluded(additionalModules.values(), identifier)) {
```
jmesnil: do you need to iterate on `additionalModules.values()` to check if it it includes the` identifier`?

Could you just check if `additionalModules.keys()` contains `identifier` instead?

server/src/main/java/org/jboss/as/server/deployment/module/descriptor/DeploymentStructureDescriptorParser.java
```diff
@@ -166,7 +165,7 @@ public void deploy(final DeploymentPhaseContext phaseContext) throws DeploymentU
                 result = parse(deploymentFile.getPhysicalFile(), deploymentUnit, moduleLoader);
             }
             // handle additional modules
-            Set<ModuleIdentifier> additionalModuleSet = new HashSet<>();
+            Map<ModuleIdentifier, AdditionalModuleSpecification> additionalModuls = new HashMap<>();
```
jmesnil: typo, should be `additionalModules`

soul2zimate: typo fixed.

server/src/main/java/org/jboss/as/server/deployment/module/descriptor/DeploymentStructureDescriptorParser.java
soul2zimate: Yes, the whole private moduleIncluded method can be eliminated.

jmesnil:
@soul2zimate thanks!

PR #3161, opened 2018-03-07
Fix for WFCORE-3671, fixed cd command path validation

jfdenise:
- Fixed handler to not send a request to validate an empty address or an address containing only a type.
- Added unit test.


jmesnil:
retest this please

jmesnil:
retest this please

PR #3135, opened 2018-02-23
[WFCORE-3566] Different results of disabling commands already disable…

xstefank:
…d deployment

JIRA: https://issues.jboss.org/browse/WFCORE-3566

PR #3166, opened 2018-03-14
[WFCORE-3566] Different results of disabling commands already disable…

xstefank:
…d deployment

JIRA: https://issues.jboss.org/browse/WFCORE-3566
Superseding https://github.com/wildfly/wildfly-core/pull/3135

### DIFF
--- a/cli/src/main/java/org/jboss/as/cli/impl/aesh/cmd/deployment/AbstractUndeployCommand.java
+++ b/cli/src/main/java/org/jboss/as/cli/impl/aesh/cmd/deployment/AbstractUndeployCommand.java
@@ -158,6 +158,9 @@ public abstract class AbstractUndeployCommand extends CommandWithPermissions
             }
         }
 
+        // WFCORE-3566 - adding the UNDEPLOY step for each found deployment is required as this
+        // request is included in the composite operation which would fail with empty steps otherwise,
+        // if the deployment does not exist, the composite will report a proper error
         for (String deploymentName : deploymentNames) {
 
             final List<String> serverGroups;
@@ -203,7 +206,7 @@ public abstract class AbstractUndeployCommand extends CommandWithPermissions
                         }
                     }
                 }
-            } else if (Util.isDeployedAndEnabledInStandalone(deploymentName, client)) {
+            } else {
                 builder = new DefaultOperationRequestBuilder();
                 builder.setOperationName(Util.UNDEPLOY);
                 builder.addNode(Util.DEPLOYMENT, deploymentName);
--- a/testsuite/standalone/src/test/java/org/jboss/as/test/integration/management/cli/DeployTestCase.java
+++ b/testsuite/standalone/src/test/java/org/jboss/as/test/integration/management/cli/DeployTestCase.java
@@ -28,6 +28,7 @@ import java.util.List;
 import org.jboss.as.cli.CommandContext;
 import org.jboss.as.cli.CommandContextFactory;
 
+import static org.hamcrest.CoreMatchers.allOf;
 import static org.hamcrest.MatcherAssert.assertThat;
 import static org.hamcrest.CoreMatchers.is;
 import static org.hamcrest.CoreMatchers.containsString;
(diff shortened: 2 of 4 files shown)


cli/src/main/java/org/jboss/as/cli/impl/aesh/cmd/deployment/AbstractUndeployCommand.java
```diff
@@ -118,10 +118,13 @@ static void undeployName(CommandContext ctx, String name,
         try {
             ModelNode request = buildRequest(ctx,
                     name, allServerGroups, serverGroups, keepContent, headers);
-            final ModelNode result = ctx.
-                    getModelControllerClient().execute(request);
-            if (!Util.isSuccess(result)) {
-                throw new CommandException(Util.getFailureDescription(result));
+
+            if (request.get(Util.STEPS).isDefined()) {
```
bstansberry: This should be:

```
if (request.hasDefined(Util.STEPS)) {
```

A call to ModelNode.get is actually a mutating call, as it creates the child if it isn't present. You don't want to mutate just to check existence.

I'll leave it to @jfdenise to do the official review. I do think though that there should be some comment to explain why that check is relevant, i.e. why is it guaranteed that only with the presence of 'steps' that 'request' is useful. It seems a bit fragile, since in general an op without a 'steps' param is valid.

xstefank: Even if the request is not used anymore, there is no problem in changing this to hasDefined

bstansberry:
@jfdenise please approve

xstefank:
@bstansberry this is still waiting for the test inclusion, the tester should be available next week, please wait with the merge

vmarekrh:
This patch fixes standalone variant and doesn't break the domain variant. Good.
But it affects error message if you disabling the wrong name of application deployments, an error isn't thrown.

xstefank:
@vmarekrh thank you. I updated the PR with the standalone handling of the disable in the same way as the domain does: it adds the UNDEPLOY step even if there is no need for it. The method `Util.isDeployedAndEnabledInStandalone` cannot be removed as it is public and GH search gives some cases of its use.

vmarekrh:
@xstefank Now it looks good. Now fixes standalone variant, doesn't break the domain variant and error message are thrown if you disabling the wrong name of application deployments.

cli/src/main/java/org/jboss/as/cli/impl/aesh/cmd/deployment/AbstractUndeployCommand.java
```diff
@@ -118,6 +118,7 @@ static void undeployName(CommandContext ctx, String name,
         try {
             ModelNode request = buildRequest(ctx,
                     name, allServerGroups, serverGroups, keepContent, headers);
+
```
jfdenise: You can remove this new line.

jfdenise: @bstansberry we are changing the behavior but in a manner that should be aligned with what a user expect.
Today, in standalone mode a check is done on the client to see if the deployment exists and is enabled. If the deployment doesn't exist or is already disabled, then an empty composite is sent and we got a generic "empty composite error" that is meaningless. This PR removes this client check and send the request in all cases.
To me that seems to be fine.

It would help to have a comment explaining what is the domain and new standalone logic there:
- In domain mode, when disabling a deployment from a server-group, the operation is sent to the server whatever the state of the deployment (no check is made). If the deployment is already disabled, there doesn't seem to have any side effect to disable an already disabled deployment. If the deployment doesn't exist, then the server will report a proper error.

- Similarly, in standalone mode, an operation to disable the deployment is sent whatever the fact that the deployment exists and/or is enabled. If the deployment is already disabled, there doesn't seem to have any side effect to disable an already disabled deployment. If the deployment doesn't exist, then the server will report a proper error.

testsuite/standalone/src/test/java/org/jboss/as/test/integration/management/cli/DeployTestCase.java
```diff
@@ -503,7 +504,7 @@ public void testDeployWithWrongCli() throws Exception {
         final DeploymentInfoResult before = deploymentInfo(cli);
 
         // Try deploy cli archive with wrong path
-        tempCliTestAppWar = createCliArchive("ls "+ wrongArgument +" sgsfgfd ghf d");
+        tempCliTestAppWar = createCliArchive("ls " + wrongArgument + " sgsfgfd ghf d");
```
jfdenise: you should revert this line.

testsuite/standalone/src/test/java/org/jboss/as/test/integration/management/cli/DeployTestCase.java
```diff
@@ -571,4 +572,142 @@ public void testEnableWrongDeployments() throws Exception {
...
+    /**
+     * Testing disabling of non-deployed application deployments.
+     * Verify if status of application deployments hasn't change.
+     * Verify error message.
+     *
+     * @throws Exception
+     * @see <a href="https://issues.jboss.org/browse/WFCORE-3566">WFCORE-3566</a>
+     */
+    @Test
+    public void testDisableWrongDeploymentWithServerGroups() throws Exception {
```
jfdenise: ServerGroups are out of scope. Could you rename the methods. Apply to all of them I think.

vmarekrh: I mixed two classes. Thank you.

testsuite/standalone/src/test/java/org/jboss/as/test/integration/management/cli/DeployTestCase.java
vmarekrh: Ok

xstefank:
retest this please

xstefank:
@bstansberry this should be good to go

PR #3170, opened 2018-03-16
[WFCORE-3587] Remove maven deps corresponding to the unneeded module deps

bstansberry:
This follows up #3129 with a commit that removes unneeded maven deps that correspond to the unneeded module deps.

This also helps validate the #3129 change by proving that the removed runtime deps aren't need at compile time either.

### DIFF
--- a/core-management/core-management-client/pom.xml
+++ b/core-management/core-management-client/pom.xml
@@ -39,7 +39,7 @@
     <dependencies>
         <dependency>
             <groupId>org.wildfly.core</groupId>
-            <artifactId>wildfly-controller</artifactId>
+            <artifactId>wildfly-controller-client</artifactId>
         </dependency>
     </dependencies>
 </project>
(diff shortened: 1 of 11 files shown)


jmesnil:
The PR has compilation issues:

```
[02:03:46][Step 2/3] [ERROR] Failed to execute goal org.apache.maven.plugins:maven-compiler-plugin:3.7.0:compile (default-compile) on project wildfly-core-management-client: Compilation failure: Compilation failure: 
[02:03:46][Step 2/3] [ERROR] /store/work/tc-work/734840e5f962fb3f/core-management/core-management-client/src/main/java/org/wildfly/extension/core/management/client/Process.java:[25,53] package org.jboss.as.controller.client.helpers does not exist
```

It is likely caused by removing the dep to wildfly-controller in core-management-client/pom.xml

jmesnil:
The failures in this PR confirms that #3129 does not take into account java constants that are inlined during compilation.

bstansberry:
@jmesnil Should be fixed. It actually had a compile-time dep on controller-client, not on controller, so the pom was wrong. It just worked because of transitive dependencies.

PR #3178, opened 2018-03-20
[WFCORE-3700] Upgrade to SLF4J 1.7.22.jbossorg-1

kabir:
https://issues.jboss.org/browse/WFCORE-3700

Bring this into sync with what is brought in by full


kabir:
See comment on https://issues.jboss.org/browse/JBEAP-14386 (@jamezp will add soon), for why we don't want this

PR #3180, opened 2018-03-21
[WFCORE-3683]: Enable a resource to define requirements without binding them to an attribute.

ehsavoie:
Passing the resource registration address to build the requirement capability pattern.

JIRA: https://issues.jboss.org/browse/WFCORE-3683

### DIFF
--- a/controller/src/main/java/org/jboss/as/controller/CapabilityReferenceRecorder.java
+++ b/controller/src/main/java/org/jboss/as/controller/CapabilityReferenceRecorder.java
@@ -19,12 +19,12 @@
  * Software Foundation, Inc., 51 Franklin St, Fifth Floor, Boston, MA
  * 02110-1301 USA, or see the FSF site: http://www.fsf.org.
  */
-
 package org.jboss.as.controller;
 
 import java.util.ArrayList;
 import java.util.List;
 import java.util.Set;
+import java.util.function.Function;
 
 import org.jboss.as.controller.capability.RuntimeCapability;
 import org.jboss.as.controller.registry.ImmutableManagementResourceRegistration;
@@ -37,20 +37,23 @@ import org.jboss.dmr.ModelNode;
  * @author Brian Stansberry (c) 2015 Red Hat Inc.
  */
 public interface CapabilityReferenceRecorder {
+
     /**
      * Registers capability requirement information to the given context.
-     * @param context         the context
-     * @param resource        the resource on which requirements are gathered
-     * @param attributeName   the name of the attribute
+     *
+     * @param context the context
+     * @param resource the resource on which requirements are gathered
+     * @param attributeName the name of the attribute
      * @param attributeValues the values of the attribute, which may contain null
      */
     void addCapabilityRequirements(OperationContext context, Resource resource, String attributeName, String... attributeValues);
 
     /**
      * Deregisters capability requirement information from the given context.
-     * @param context         the context
-     * @param resource        the resource on which requirements are gathered
-     * @param attributeName   the name of the attribute
+     *
+     * @param context the context
+     * @param resource the resource on which requirements are gathered
+     * @param attributeName the name of the attribute
      * @param attributeValues the values of the attribute, which may contain null
      */
     void removeCapabilityRequirements(OperationContext context, Resource resource, String attributeName, String... attributeValues);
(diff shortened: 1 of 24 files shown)


jmesnil:
@ehsavoie both Linux and Linux JDK9 jobs failed with a checkstyle error in wildfly-controller. 

Could you have a look please? (that's weird, as the Windows job did not fail...)

pferraro:
After playing around with this a bit with clustering capabilities, I think we should really create a separate interface for resource capability references.
 + Attribute capability references are always treated separately from resource capability references
 + The CapabilityReferenceRecorder API is clearly intended for attributes with a set of values
   * addCapabilityRequirements/removeCapabilityRequirements send null for attributeName, and an empty attribute value array
   * getRequirementPatternSegments is called with a null "name" (i.e. the attribute value).
 + Associating a ResourceCapabilityReferenceRecorder to an attribute, or a DefaultCapabilityReferenceRecorder with a resource would only fail at runtime via assertion failures.

controller/src/main/java/org/jboss/as/controller/CapabilityReferenceRecorder.java
```diff
@@ -321,4 +330,97 @@ protected String getRequirementName(OperationContext context, Resource resource,
...
+
+        public ResourceCapabilityReferenceRecorder(Function<PathAddress, String[]> dynamicDependentNameMapper, String baseDependentName, Function<PathAddress, String[]> dynamicRequirementNameMapper, String baseRequirementName) {
+            this.dynamicDependentNameMapper = dynamicDependentNameMapper;
+            this.dynamicRequirementNameMapper = dynamicRequirementNameMapper;
+            this.baseRequirementName = baseRequirementName;
+            this.baseDependentName = baseDependentName;
+        }
+
+        public ResourceCapabilityReferenceRecorder(Function<PathAddress, String[]> dynamicDependentNameMapper, String baseDependentName, String baseRequirementName) {
+            this(dynamicDependentNameMapper, baseDependentName, null, baseRequirementName);
```
pferraro: Why would anyone use this constructor (i.e. a null dynamicRequirementNameMapper)?
If a capability requires a statically named requirement, it would define this requirement via RuntimeCapability.addRequirement(...).  There would be no reason to create a separate capability reference recorder.

controller/src/main/java/org/jboss/as/controller/CapabilityReferenceRecorder.java
```diff
@@ -321,4 +330,97 @@ protected String getRequirementName(OperationContext context, Resource resource,
...
+            return baseRequirementName;
+        }
+
+        @Override
+        public boolean isDynamicDependent() {
+            return dynamicDependentNameMapper != null;
+        }
+
+        @Override
+        public String[] getRequirementPatternSegments(String dynamicElement, PathAddress registrationAddress) {
```
pferraro: We should be able to just return: dynamicRequirementNameMapper.apply(registrationAddress)
This method is always invoked with a null dynamicElement.
Additionally, dynamicRequirementNameMapper should never be null (see my comment above).

ehsavoie: We can't because the address might be aliased

controller/src/main/java/org/jboss/as/controller/CapabilityReferenceRecorder.java
ehsavoie: I agree

controller/src/main/java/org/jboss/as/controller/CapabilityReferenceRecorder.java
pferraro: I'm a little confused.  If getRequirementPatternSegments(...) is called with a null dynamicElement (which is how it is called when building the resource description), this method just returns the value of dynamicRequirementNameMapper.apply(registrationAddress);
I don't see any special handling for aliased addresses here.

ehsavoie: I've updated the code in the meanwhile

jmesnil:
As discussed during the management call, this PR is blocking progress on provisioning work.
The API can be further improved but its current state allows to move forward with provisioning. The important thing is to have a stable `read-feature` model representation as it is consumed by the provisioing tool.

I'll merge it as it is and prepare a Core release so that we can merge https://github.com/wildfly/wildfly/pull/11113 too.

@ehsavoie Could you prepare a 2nd PR related to this to flag as `@Deprecated` any new experimental API so that we have a way to know in the subsystems that calling this API may be subject to change?

PR #3182, opened 2018-03-22
[WFCORE-3703] Move all SLF4J dependencies to Core

jmesnil:
* move the org.slf4j.ext module (and its ch.qos.cal10n dependency) to
  core-feature-pack
* add dependency to jcl-over-slf4j. It is not directly required for
  WildFly but this allows to have all SLF4J dependencies in Core and
  have a single authoritative version for them.

JIRA: https://issues.jboss.org/browse/WFCORE-3703

### DIFF
--- a/core-feature-pack/pom.xml
+++ b/core-feature-pack/pom.xml
@@ -59,6 +59,11 @@
         </dependency>
 
         <dependency>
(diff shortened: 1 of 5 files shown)


jmesnil:
@jamezp I have updated the slf4j version to 1.7.22.jbossorg-1

PR #3183, opened 2018-03-22
Bug fix: WFCORE-3522

yelhouti:
try to fix dependency order, local first


dmlloyd:
This is OK to test

dmlloyd:
Though it's safe to say, it's unlikely to succeed :)

bstansberry:
this is ok to test

jmesnil:
The test failures are related to the PR:

```
Caused by: java.util.ServiceConfigurationError: org.jboss.msc.service.ServiceActivator: Provider org.jboss.as.test.integration.management.cli.modules.ModulesServiceActivator not a subtype"}}
```

```
java.lang.ClassCastException: org.jboss.as.connector.subsystems.datasources.WildFlyDataSource cannot be cast to org.jboss.as.connector.subsystems.datasources.WildFlyDataSource
```

yelhouti:
the fix is much more complex, the problem is still there though

PR #3186, opened 2018-03-23
[WFCORE-3404] Elytron audit performance - separated sync and flushing

hkalina:
Depends on https://github.com/wildfly-security/wildfly-elytron/pull/1117

https://issues.jboss.org/browse/WFCORE-3404
https://issues.jboss.org/browse/JBEAP-10639


jmesnil:
Patch does not build:

```
Compilation failure
/store/work/tc-work/aa7b981af3f49e3c/elytron/src/main/java/org/wildfly/extension/elytron/AuditResourceDefinitions.java:[231,37] cannot find symbol
  symbol:   method setFlushOnAccept(boolean)
  location: class org.wildfly.security.audit.FileAuditEndpoint.Builder
```

hkalina:
@jmesnil as mentioned in description, depends on https://github.com/wildfly-security/wildfly-elytron/pull/1117 (sooner was required from me to have created both PRs, for subsystem and for wildfly-elytron, before both of them could be merged, so I suppose it is still needed)

jmesnil:
@hkalina ok, could you then ping the mergers when the PR for Elytron upgrade is also opened so we look at both at the same time?

thanks

hkalina:
ok, closing this until elytron upgrade pr will be available

hkalina:
reopening as this PR was considered requirement for elytron PR merging

darranl:
There may still be a small compatibility gap, in legacy config if an administrator had set SYNCHRONIZED=false at the time that was the equivalent to SYNCHRONIZED=false, FLUSHED=false

After this change is applied that config becomes SYNCHRONIZED=false, FLUSHED=true - maybe we could handle that in the parsing but it is still an issue for operations.  

Changing the default value of FLUSHED doesn't help as the problem just inverts when SYNCHORINZED=true.

I think this PR is missing transformers as well.  Transformers could be interesting as well as depending on the permutations of values for these two attributes we may be able to transform to the older model.

An alternative would be deprecate SYNCHRONIZED and add two alternative attributes instead of just the one - if SYNCHRONIZED is defined it affects both options as it did previously, if not we use the other two.

@bstansberry What do you think, this is quite a small incompatibility - am I suggesting we go further than we really need to?

hkalina:
Fixed: flushed takes synchronized value as default from now, added transformers.
Still depends on related wildfly-elytron change to be merged, tagged and upgraded.

jmesnil:
@hkalina this PR can be applied after #3228, is that correct?

jmesnil:
retest this please

darranl:
@kabir could you have a look at this one please in relation to the RFE process.

I understand Martin's point but I just want to be sure we have the cut off between bug and RFE correctly defined - if we go with this policy I would expect no PRs across WildFly Core or WildFly that modify any aspect of the model go in without following the RFE process.

kabir:
I would say it is historically a bug, but the way it is fixed it becomes a feature (as it introduces more config options, which needs more test coverage). I realise you'd probably hoped to get this into 13 as a bug, but on the other hand the amount of paperwork needed should be quite small so it can come in early for 14

darranl:
If that is the policy I am happy for us to go that way, just want to be sure we are not accidentally getting security onto a stricter review process ;-)

Can we please make sure all mergers know that changes that affect the model must not be merged without following the RFE process?

jmesnil:
this PR requires to be rebased

hkalina:
@jmesnil rebased

kabir:
@darranl @hkalina I don't see any RFE related work on this one. Will comment more on WFCORE-3404

elytron/src/main/java/org/wildfly/extension/elytron/AuditResourceDefinitions.java
```diff
@@ -87,6 +87,11 @@
             .setRestartAllServices()
             .build();
 
+    static final SimpleAttributeDefinition FLUSHED = new SimpleAttributeDefinitionBuilder(ElytronDescriptionConstants.FLUSHED, ModelType.BOOLEAN, true)
```
jamezp: I'd suggest `autoflush` to be consistent with what the logging subsystem is using.

hkalina:
rebased

jamezp:
@hkalina Since it looks like since this is a new feature we need community documentation for this.

jmesnil:
@hkalina Feature freeze for WildFly Core is tomorrow (August 10th). It seems there are still missing requirements before this PR can be merged. Could you please check that the pre-merge requirement parallel tasks are up to date and report any update?

darranl:
@jmesnil we have discussed this with QE and agreed to defer to WildFly 15.

As this affects the management model maybe we will be better closing and resubmitted after porting to the next version of the model.

PR #3190, opened 2018-03-29
[WFCORE-3513] Add unit test

adamecp:
Link to JIRA [https://issues.jboss.org/browse/WFCORE-3513](https://issues.jboss.org/browse/WFCORE-3513)

Description: In case a file not named as yyyyMMdd-HHmmssSSSname.xml is present in the snapshot folder, the server start abort with a StringIndexOutOfBoundsException.

Original pull request with manual mode integration test [https://github.com/wildfly/wildfly/pull/11035](https://github.com/wildfly/wildfly/pull/11035)

### DIFF
--- a/controller/src/test/java/org/jboss/as/controller/persistence/ConfigurationFileUnitTestCase.java
+++ b/controller/src/test/java/org/jboss/as/controller/persistence/ConfigurationFileUnitTestCase.java
@@ -1,0 +1,136 @@
+/*
+Copyright 2018 Red Hat, Inc.
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+  http://www.apache.org/licenses/LICENSE-2.0
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+ */
+
+package org.jboss.as.controller.persistence;
+
+import org.junit.After;
+import org.junit.Assert;
+import org.junit.Before;
+import org.junit.Test;
+
+import java.io.File;
+import java.io.IOException;
+import java.nio.charset.StandardCharsets;
+import java.nio.file.Files;
+
+/**
+ *
+ * @author <a href="padamec@redhat.com">Petr Adamec</a>
+ */
+public class ConfigurationFileUnitTestCase {
+    private static final String BASE_CONFIGURATION_NAME = "standalone.xml";
+    private static final String HISTORY_CONFIGURATION_NAME = "20180101-000000000standalone.xml";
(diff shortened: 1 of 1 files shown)


controller/src/test/java/org/jboss/as/controller/persistence/ConfigurationFileUnitTestCase.java
```diff
@@ -0,0 +1,123 @@
...
+     * For more information visit <a href="issues.jboss.org/browse/WFCORE-3513">https://issues.jboss.org/browse/WFCORE-3513</a>
+     */
+    @Test
+    public void testConstructorWithCorrectFileAtSnapshot() {
+        try{
+            createFile(configurationDir, BASE_CONFIGURATION_NAME, "");
+            createFile(snapshotsDirectory, BASE_CONFIGURATION_NAME, "");
+            createFile(snapshotsDirectory, HISTORY_CONFIGURATION_NAME, "");
+            new ConfigurationFile(configurationDir, BASE_CONFIGURATION_NAME, BASE_CONFIGURATION_NAME, true);
+        } catch (StringIndexOutOfBoundsException e){
```
bstansberry: Can this just catch Exception?  That makes it a more broadly useful test.

adamecp: @bstansberry I think, with StringIndexOutOfBoundsException, it will be better to see where is problem. If StringIndexOutOfBoundsException will be catch, test will be failed with message _Could not create an instance of ConfigurationFile ..._ . And we will know that problem is with configuration file.  When another thing happends, test will be failed too. And we will can see where is problem from exception.
But if you wanna it, I can use only general exception.

bstansberry: Right; good point. Ok, this is fine.

controller/src/test/java/org/jboss/as/controller/persistence/ConfigurationFileUnitTestCase.java
```diff
@@ -0,0 +1,123 @@
+package org.jboss.as.controller.persistence;
```
bstansberry: Please add the standard ASL copyright header. Thanks.

adamecp:
@bstansberry  Thank you for review. I have added ASL copyright header.

controller/src/test/java/org/jboss/as/controller/persistence/ConfigurationFileUnitTestCase.java
jmesnil: @bstansberry did you mean the LGPL header?

bstansberry: @jmesnil @adamecp I meant the ASL header. All new files in WildFly Core should use the ASL header, not the LGPL one.

adamecp:
@jmesnil @bstansberry  I have updated header.

PR #3191, opened 2018-03-29
WIP: [WFCORE-3568] Allow CLI output scrolling

MMarus:
WFCORE: https://issues.jboss.org/browse/WFCORE-3568

### DIFF
--- a/cli/src/main/java/org/jboss/as/cli/impl/ReadlineConsole.java
+++ b/cli/src/main/java/org/jboss/as/cli/impl/ReadlineConsole.java
@@ -560,6 +560,9 @@ public class ReadlineConsole {
             int max = connection.size().getHeight();
             int currentLines = 0;
             int allLines = 0;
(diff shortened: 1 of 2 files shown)


MMarus:
Retest this please.

cli/src/main/java/org/jboss/as/cli/impl/ReadlineConsole.java
```diff
@@ -609,6 +639,19 @@ private void printCollectedOutput() {
                     connection.write(l + Config.getLineSeparator());
                 }
             }
+            connection.write(ANSI.MAIN_BUFFER);
```
jfdenise: This should also be called in a finally block. In case something goes wrong we should guarantee that the console is put in the nominal state.

MMarus:
Closing this PR since it is going to be included in PR for https://issues.jboss.org/browse/WFCORE-3743 .

PR #3192, opened 2018-03-29
[WFCORE-3721] Upgrade build tools to 1.2.10.Final

dmlloyd:
https://issues.jboss.org/browse/WFCORE-3721


dmlloyd:
The duplication of the version property is covered by https://issues.jboss.org/browse/WFCORE-3722

PR #3198, opened 2018-04-06
[WFCORE-3731] Minor code cleanup in CommandCommandHandler, Connection…

rsvoboda:
…InfoHandler, RealmIdentityManager

CommandCommandHandler - removal of unnecessary if, checked with Jean-Francois Denise
ConnectionInfoHandler - lastChainCertificate is X509Certificate[], instanceof check is not necessary
RealmIdentityManager - credential is always PasswordCredential, only null check is relevant

JIRA: https://issues.jboss.org/browse/WFCORE-3731


bstansberry:
This is failing checkstyle checks in the cli module. I didn't look where but my guess is it's from

https://github.com/wildfly/wildfly-core/pull/3198/files#diff-482969cdb3512e73324a990131103cadL108

My guess is 'Certificate' is now an unused import.

Once that's corrected this looks fine.

rsvoboda:
Thanks Brian, that was the problem. I should stop using `-Denforcer.skip=true -Dcheckstyle.skip=true` to gain few seconds.

Failed tests on Windows is org.jboss.as.test.integration.ejb.stateful.passivation.PassivationTestCase testPassivationFailure
Doesn't see to be PR changes related - changes in cli and domain-http module

jmesnil:

PR #3201, opened 2018-04-09
[WFCORE-3684] Upgrade to JBoss Modules 1.8.0.Final, including [WFCORE-3705] support for JDK modules

stuartwdouglas:
https://issues.jboss.org/browse/WFCORE-3684
https://issues.jboss.org/browse/WFCORE-3705
https://issues.jboss.org/browse/WFCORE-3733
wildfly/wildfly-proposals#39


stuartwdouglas:
Retest this please

psakar:
@stuartwdouglas How will be artifacts version determined ? I can not find any information on related JIRAs 
@rsvoboda FYI

stuartwdouglas:
@psakar that is a question for @dmlloyd, all I did was fix the ee8 preview stuff

dmlloyd:
@bstansberry my plan was to follow up with dependency updates.  I wanted to separate that because it should be clear that it isn't a requirement to make this work.

That said, I think deprecation warnings are unlikely because I believe all the currently impacted modules are implicit dependencies (I will double-check this though).

dmlloyd:
Here's how the deprecations are mapped:

``javax.sql.api`` -> ``java.sql``
``javax.api`` -> ``java.se`` (or better yet, break out the independent deps)
``javax.xml.*.api`` -> ``java.xml``

bstansberry:
@dmlloyd Ah, ModuleLoadService doesn't log the WARN except for explicitly user-specified dependencies, so that addresses my concern. For explicitly user-specified ones, the WARN is of course the right thing to do. 

Sorry for the noise.

rsvoboda:
JBoss Modules 1.8 are feature/enhancement rich, maybe we should have some summary for this release and send it to the wildfly-dev list

https://issues.jboss.org/issues/?jql=project%20%3D%20MODULES%20AND%20fixVersion%20%3D%201.8.0.Final%20ORDER%20BY%20issuetype%20DESC

dmlloyd:
Done: http://lists.jboss.org/pipermail/wildfly-dev/2018-April/006467.html

PR #3202, opened 2018-04-09
[WFCORE-3205 & WFCORE-3698] Fix embedded server property setting and update the logging works

jamezp:
# JIRA's
https://issues.jboss.org/browse/WFCORE-3205
https://issues.jboss.org/browse/WFCORE-3698

# General Comments
These commits are admittedly a bit out of order. If needed I can organize them a bit better or squash the last 3 into one commit.

# First commit (WFCORE-3698)
This simply adds a new `Context` API which the `EmbeddedManagedProcess` uses as a setup/tear down type of solution. It sets up and removes system properties that the container will set honoring user defined system properties.

# Second Commit
This commit adds a new embedded test suite.

# Third and Forth Commits
These commits could likely be squashed. These third commit just adds a new `LoggerContext` which is used to configure a `ModuleLogger` that delegates to JBoss Logging

The forth commit adds a new `Configuration` API to the embedded API. It also allows the logging subsystem to ignore the log manager check if the subsystem is present.

### DIFF
--- a/cli/src/main/java/org/jboss/as/cli/embedded/EmbedServerHandler.java
+++ b/cli/src/main/java/org/jboss/as/cli/embedded/EmbedServerHandler.java
@@ -49,6 +49,7 @@ import org.jboss.logmanager.LogContext;
 import org.jboss.modules.ModuleLoader;
 import org.jboss.stdio.NullOutputStream;
 import org.jboss.stdio.StdioContext;
+import org.wildfly.core.embedded.Configuration;
 import org.wildfly.core.embedded.EmbeddedManagedProcess;
 import org.wildfly.core.embedded.EmbeddedProcessFactory;
 import org.wildfly.core.embedded.EmbeddedProcessStartException;
@@ -224,13 +225,20 @@ class EmbedServerHandler extends CommandHandlerWithHelp {
 
             String[] cmds = cmdsList.toArray(new String[cmdsList.size()]);
 
-            EmbeddedManagedProcess server;
+            final Configuration.Builder configBuilder;
             if (this.jbossHome == null) {
-                // Modular environment
-                server = EmbeddedProcessFactory.createStandaloneServer(ModuleLoader.forClass(getClass()), jbossHome, cmds);
+                // Modular environment, note that the jbossHome here is resolved from the JBOSS_HOME environment
+                // variable and should never be null according to the getJBossHome() method.
+                configBuilder = Configuration.Builder.of(jbossHome)
+                        .setModuleLoader(ModuleLoader.forClass(getClass()))
+                        .setCommandArguments(cmds);
             } else {
-                server = EmbeddedProcessFactory.createStandaloneServer(jbossHome.getAbsolutePath(), null, null, cmds);
+                configBuilder = Configuration.Builder.of(jbossHome.getAbsoluteFile())
+                        .setCommandArguments(cmds);
             }
+            // Disables the logging subsystem from registering an embedded log context if the subsystem is present
+            WildFlySecurityManager.setPropertyPrivileged("org.wildfly.logging.embedded", "false");
+            final EmbeddedManagedProcess server = EmbeddedProcessFactory.createStandaloneServer(configBuilder.build());
             server.start();
             serverReference.set(new EmbeddedProcessLaunch(server, restorer, false));
             ModelControllerClient mcc = new ThreadContextsModelControllerClient(server.getModelControllerClient(), contextSelector);
--- a/embedded/src/main/java/org/wildfly/core/embedded/ChainedContext.java
+++ b/embedded/src/main/java/org/wildfly/core/embedded/ChainedContext.java
@@ -1,0 +1,60 @@
+/*
+ * JBoss, Home of Professional Open Source.
+ *
+ * Copyright 2018 Red Hat, Inc., and individual contributors
+ * as indicated by the @author tags.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.wildfly.core.embedded;
+
+import java.util.ArrayList;
+import java.util.List;
+
+import org.wildfly.core.embedded.logging.EmbeddedLogger;
+
+/**
+ * @author <a href="mailto:jperkins@redhat.com">James R. Perkins</a>
+ */
+class ChainedContext implements Context {
+
+    private final List<Context> contexts;
+
+    ChainedContext() {
+        contexts = new ArrayList<>();
+    }
+
+    void add(final Context context) {
+        contexts.add(context);
+    }
+
+    @Override
+    public void activate() {
+        for (Context context : contexts) {
+            context.activate();
+        }
+    }
+
+    @Override
+    public void restore() {
+        for (int i = (contexts.size() - 1); i >= 0; i--) {
+            final Context context = contexts.get(i);
+            try {
+                context.restore();
+            } catch (Exception e) {
+                EmbeddedLogger.ROOT_LOGGER.failedToRestoreContext(e, context);
+            }
+        }
+    }
+}
--- a/embedded/src/main/java/org/wildfly/core/embedded/Configuration.java
+++ b/embedded/src/main/java/org/wildfly/core/embedded/Configuration.java
@@ -1,0 +1,410 @@
+/*
+ * JBoss, Home of Professional Open Source.
+ *
+ * Copyright 2018 Red Hat, Inc., and individual contributors
+ * as indicated by the @author tags.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.wildfly.core.embedded;
+
+import java.io.File;
+import java.nio.file.Files;
+import java.nio.file.Path;
+import java.nio.file.Paths;
+import java.util.ArrayList;
+import java.util.Arrays;
+import java.util.List;
+import java.util.concurrent.atomic.AtomicBoolean;
+
+import org.jboss.modules.Module;
+import org.jboss.modules.ModuleLoader;
+import org.wildfly.core.embedded.logging.EmbeddedLogger;
+
+/**
+ * Represents a configuration for the embedded server.
+ *
+ * @author <a href="mailto:jperkins@redhat.com">James R. Perkins</a>
+ */
+public interface Configuration {
+
+    /**
+     * Hints which logger provider should be used.
+     */
+    enum LoggerHint {
+        DEFAULT(null, null),
+        // Don't set the system package for the JBoss Log Manager. This causes some loggers to be set on the log context
+        // from the logging subsystem as opposed to the log context created by the user.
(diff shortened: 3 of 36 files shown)


jmesnil:
Dev analysis is at https://github.com/wildfly/wildfly-proposals/pull/28

embedded/src/main/java/org/wildfly/core/embedded/LoggerContext.java
```diff
@@ -0,0 +1,202 @@
...
+ * A context for the logging environment.
+ *
+ * @author <a href="mailto:jperkins@redhat.com">James R. Perkins</a>
+ */
+class LoggerContext implements Context {
+    private static final String MODULE_ID_LOGGING = "org.jboss.logging";
+
+    private final ModuleLoader moduleLoader;
+
+    private volatile ModuleLogger current;
```
jmesnil: just a note to myself: `current` is the ModuleLogger *before activation* (using the name `current` is a bit misleading when you are in the `restore` method).

embedded/src/main/java/org/wildfly/core/embedded/LoggerContext.java
```diff
@@ -0,0 +1,202 @@
...
+            // Reset TCCL
+            WildFlySecurityManager.setCurrentContextClassLoaderPrivileged(tccl);
+        }
+    }
+
+    @Override
+    public void restore() {
+        final ModuleLogger current = this.current;
+        if (current == null) {
+            Module.setModuleLogger(NoopModuleLogger.getInstance());
```
jmesnil: Should you set the module logger after calling WildFlySecurityManager.setCurrentContextClassLoaderPrivileged(logModuleClassLoader)? (as it is done in `activate`)?

embedded/src/main/java/org/wildfly/core/embedded/Configuration.java
```diff
@@ -0,0 +1,421 @@
...
+        private ModuleLoader moduleLoader;
+        private String modulePath;
+        private boolean embeddedLoggingHint;
+
+        /**
+         * Creates a new builder for the configuration.
+         *
+         * @param jbossHome the servers home directory
+         */
+        public Builder(final File jbossHome) {
```
jmesnil: should the Builder constructors be private as they seem to be hidden behind the `of` methods?

cli/src/main/java/org/jboss/as/cli/embedded/EmbedServerHandler.java
```diff
@@ -224,13 +225,19 @@ protected void doHandle(CommandContext ctx) throws CommandLineException {
 
             String[] cmds = cmdsList.toArray(new String[cmdsList.size()]);
 
-            EmbeddedManagedProcess server;
+            final Configuration.Builder configBuilder;
             if (this.jbossHome == null) {
                 // Modular environment
-                server = EmbeddedProcessFactory.createStandaloneServer(ModuleLoader.forClass(getClass()), jbossHome, cmds);
+                configBuilder = Configuration.Builder.of(jbossHome)
```
jmesnil: this piece of code is a bit confusing.
`this.jbossHome` which is a `ArgumentWithValue` can be `null` but `jbossHome` which is `File` can not be `null`  (according to the `getJBossHome` method).

The presence of a `this.jbossHome` means we are in a modular environment and then use the `jbossHome` value that comes from the env property `JBOSS_HOME`.
Else we are not in an environment module and uses the absolute path from the `--jboss-home` argument.

embedded/src/main/java/org/wildfly/core/embedded/Configuration.java
```diff
@@ -0,0 +1,421 @@
...
+            embeddedLoggingHint = true;
+        }
+
+        /**
+         * Creates a new builder for the configuration.
+         *
+         * @param jbossHome the servers home directory
+         */
+        public static Builder of(final File jbossHome) {
+            return new Builder(jbossHome);
```
jmesnil: From reading the code, it seems that `jbossHome` must never be `null`, right?

embedded/src/main/java/org/wildfly/core/embedded/Configuration.java
```diff
@@ -0,0 +1,421 @@
...
+
+        /**
+         * Creates a new immutable configuration.
+         *
+         * @return the configuration
+         */
+        public Configuration build() {
+            final Path jbossHome = this.jbossHome;
+            final LoggerHint loggerHint = this.loggerHint == null ? LoggerHint.DEFAULT : this.loggerHint;
+            final boolean embeddedLoggingHint = this.embeddedLoggingHint;
```
jmesnil: when is this `embeddedLoggingHint` used to build the `Configuration`?

jmesnil: ah, nm, the field itself is used in `configureLogging()` :)

embedded/src/main/java/org/wildfly/core/embedded/Configuration.java
```diff
@@ -0,0 +1,421 @@
...
+                }
+
+                @Override
+                public ModuleLoader getModuleLoader() {
+                    return moduleLoader;
+                }
+
+                @Override
+                public String[] getCommandArguments() {
+                    return Arrays.copyOf(cmdArgs, cmdArgs.length);
```
jmesnil: Why are you copying the `cmdArgs` array instead of passing it directly?

jmesnil:
@jamezp that looks good to me except for a doubt on the order of ChainedContext restoration.

embedded/src/main/java/org/wildfly/core/embedded/LoggerContext.java
jamezp: I'll change the name as that does make sense. I'll just name it `toRestore` or something similar.

embedded/src/main/java/org/wildfly/core/embedded/ChainedContext.java
```diff
@@ -0,0 +1,59 @@
...
+    @Override
+    public void activate() {
+        for (Context context : contexts) {
+            context.activate();
+        }
+    }
+
+    @Override
+    public void restore() {
+        for (Context context : contexts) {
```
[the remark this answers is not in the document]
jamezp: Good catch. This should likely be done in reverse order.

cli/src/main/java/org/jboss/as/cli/embedded/EmbedServerHandler.java
jamezp: Correct and I do agree it's a bit confusing. I'll add a comment as to how it works. I just copied what the previous version used for it's home directory.

embedded/src/main/java/org/wildfly/core/embedded/Configuration.java
jamezp: Only to ensure the `Configuration` is immutable.

jamezp:
@jmesnil All the changes have been complete. Thanks for the review!

embedded/src/main/java/org/wildfly/core/embedded/Configuration.java
jmesnil: I don't think that's necessary. `cmdArgs` is already an array that is created in `build()` by calling `this.cmdArgs.toArray(new String[0]);`.

jmesnil:
This feature is missing community documentation (maybe in http://docs.wildfly.org/Admin_Guide.html#Logging)  before it can be merged

jamezp:
@jmesnil This was more of a general change with the embedded API than an actual logging change. I'd probably consider the logging changes more of a bug fix than a feature TBH.

That said the documentation might take a while. I noticed there is no documentation at all for the embedded API. Most of this fix is for Galleon provisioning. I'll try to get to the documentation next week though.

jamezp:
@rsvoboda Could we get someone from QE to have a look and sign off on this?

rsvoboda:
I will try to find the right person for this

embedded/src/main/java/org/wildfly/core/embedded/SystemPropertyContext.java
```diff
@@ -0,0 +1,164 @@
...
+import java.nio.file.Paths;
+import java.util.HashMap;
+import java.util.HashSet;
+import java.util.Iterator;
+import java.util.Map;
+import java.util.Set;
+
+import org.jboss.as.server.ServerEnvironment;
+import org.wildfly.common.Assert;
+import org.wildfly.security.manager.WildFlySecurityManager;
```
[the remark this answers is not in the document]
jamezp: Almost forgot to address the wildfly-common dependency. That was not already a dependency on the embedded API. I'll remove the references to that so we can make it provided.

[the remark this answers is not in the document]
dmlloyd: I ran jdeps on a recent build of the embedded module and it did require wildfly-common even before this change.  It may have previously come in transitively but I think it's a legitimate dependency either way.

nziakova:
@jamezp I'm the QE contact for the logging part (WFCORE-3205). However, this PR includes additional work. Would it be possible to split this PR and create a new requirement for the embedded context, too?

rsvoboda:
logging related part contact is @nziakova 
non-logging related part contact is @mjurc 

Both are looking into it

jamezp:
@nziakova Normally I wouldn't combine PR's like this, but they are related somewhat. It would make it more difficult to separate them because one would have to be merged before the other one would work. If you look at the last two commits those should be the logging only commits.

Thanks @rsvoboda!

jamezp:
I've now added some embedded API documentation https://github.com/wildfly/wildfly-core/pull/3202

embedded/src/main/java/org/wildfly/core/embedded/Configuration.java
```diff
@@ -0,0 +1,426 @@
...
+         * @return this builder
+         */
+        public Builder setLoggerHint(final LoggerHint loggerHint) {
+            this.loggerHint = loggerHint;
+            return this;
+        }
+
+        /**
+         * If the logging subsystem is present and the value is set to {@code false}, the embedded log context will not
+         * be used.
```
bstansberry: What is 'the embedded log context'?

embedded/src/main/java/org/wildfly/core/embedded/Configuration.java
```diff
@@ -0,0 +1,426 @@
...
+         *
+         * @return this builder
+         */
+        public Builder setCommandArguments(final String... args) {
+            cmdArgs.clear();
+            return addCommandArguments(args);
+        }
+
+        /**
+         * Sets the hint for the logger provider.
```
bstansberry: What is a logger provider?

jamezp:
@bstansberry I've attempted to clarify the JavaDoc's and changed that one method name. Hopefully it's more clear now.

embedded/src/main/java/org/wildfly/core/embedded/Configuration.java
```diff
@@ -0,0 +1,432 @@
...
+         * @return this builder
+         */
+        public Builder setLoggerHint(final LoggerHint loggerHint) {
+            this.loggerHint = loggerHint;
+            return this;
+        }
+
+        /**
+         * Sets a hint for the logging subsystem to indicate whether or not subsystem should configure logging on a
+         * special log context or not. A value of {@code true} indicates the special log context should be used. A
```
bstansberry: What's the special log context?

I think this and setLoggerHint need to explain pretty clearly to non-logging-experts what's going to happen. This javadoc is going to be the primary way users and future code maintainers understand what this does.

embedded/src/main/java/org/wildfly/core/embedded/Configuration.java
```diff
@@ -0,0 +1,432 @@
...
+         * value of {@code false} indicates no special log context should be used. The default is {@code true}.
+         *
+         * @param embeddedLoggingHint {@code false} if the logging subsystem should not use a special log context for
+         *                            it's configuration
+         *
+         * @return this builder
+         */
+        public Builder setEmbeddedLogContextHint(final boolean embeddedLoggingHint) {
```
bstansberry: Is this basically "setUseEmbeddedLogContext(boolean)?  What's the difference between a hint and directly stating what to do?

jamezp: Essentially this was done for offline CLI. Maybe I should just remove this method and explicitly set the `org.wildfly.logging.embedded` system property in offline CLI.

jamezp:
@bstansberry I just removed that method in question. It was really a shortcut for setting a system property which would really requiring working knowledge of how log contexts work.

bstansberry:
So now the behavior is by default there is a special log context (I'm still curious precisely what it does) and if appropriate the CLI can use a system prop to turn on the traditional logging subsystem. That seems fine, just treat the system prop kind of like an API (i.e. don't break it) and then if there's demand, a Configuration method can always be added later.

PR #3207, opened 2018-04-11
[WFCORE-3705] Cleanup: move to undeprecated modules

dmlloyd:


dmlloyd:
Failures seem related.  I can't reproduce them locally yet though; still working on that.

dmlloyd:
Ah, it only fails on Java 8.  OK, now to fix...

dmlloyd:
All fixed.

rsvoboda:
`<module name="jdk.unsupported"/>` on several places brings some uncomfortable feelings.

http://openjdk.java.net/jeps/260 is not saying how long the module will be available. @dmlloyd are you aware of future plans for It? I hope we would be safe with it for Java 11. After that some drops could happen I guess.

dmlloyd:
The ``jdk.unsupported`` module will be around until the ``Unsafe`` API has been replaced by a new, as-yet undetermined API for serialization.

dmlloyd:
@jmesnil done; I've split the commit.

PR #3209, opened 2018-04-12
[WFCORE-3724] Avoid reload operation waiting for servers until they are stopped

yersan:
Jira issue: https://issues.jboss.org/browse/WFCORE-3724
Analysis doc: https://github.com/wildfly/wildfly-proposals/pull/54

This PR allows a Host Controller reloading without waiting for the servers to complete its restart. The change will allow the user continue using the CLI even if there are still servers stopping (by a previous operation) or restarting due to the reload itself.

Although to have servers in a Domain Controller is not the recommended configuration, the fact of being able to continue using the CLI in a reload sounds reasonable and it will also let in the future to do a Host Controller reload with a graceful shutdown without blocking the CLI or even the whole domain.

Observed changes:
This PR makes the reload operation faster for HC/DC with managed servers since the operation does not have to wait for the servers before returning.

That means we can be retrieving host-state attribute as RUNNING when 1) the servers are still STOPPING or 2) when another slave has not been yet registered in the domain.

The case 1) is completely new, maybe we could change the reload operation to be blocking/un-blocking based on a user decision in order to maintain the previous behaviour.

The case 2) is new for HC/DC with servers. The failure was detected in ReadOnlyModeTestCase, where a master DC is reloaded, and as soon as it is running, a value in a slave is read. It fails because we are trying to read a value in a slave HC when it has not been registered yet in the domain. The same issue can be reproduced with the master branch if we run the test using a master DC without servers. Waiting until getting the servers started it is enough in order to allow the test pass with this PR.

Before continue adding some tests, I would like to get some early feedback. I am not sure if the issue should be considered isolated or should be part of https://issues.jboss.org/browse/WFCORE-1427


yersan:
The error in JDK9 looks relevant to this PR, reviewing

yersan:
Some rework done.

PR divided into two commits, the first one moves the stop operation at the end of HostControllerService boot sequence; it will allow waiting for servers after the HC has been almost reloaded, the user will receive `"WFLYCTL0379: System boot is in process; execution of remote management operations is not currently available"` if tries to execute additional operations while servers are being stopped.

The second commit adds the use of a `blocking` attributed for reload operation to change the default behaviour. The user can decide whether he wants to wait for the reload or wants to restart the servers in the background and continue using the CLI.

yersan:
Test failure in 'Linux (Pull Request) - merge' is unrelated

yersan:
Rebased to resolve conflicts

host-controller/src/main/java/org/jboss/as/host/controller/operations/StartServersHandler.java
```diff
@@ -95,27 +96,47 @@ public void execute(OperationContext context, ModelNode operation) throws Operat
                 final Resource resource =  context.readResource(PathAddress.EMPTY_ADDRESS);
                 final ModelNode hostModel = Resource.Tools.readModel(resource);
                 if(hostModel.hasDefined(SERVER_CONFIG)) {
+
                     final ModelNode servers = hostModel.get(SERVER_CONFIG).clone();
+                    final boolean restartRequired = runningModeControl.getRestartMode() == RestartMode.SERVERS;
+                    final boolean blockUntilStopped = hostControllerEnvironment.isRestart() || !runningModeControl.isReloadByHandler() || (runningModeControl.isReloadByHandler() && runningModeControl.isBlockUntilStopped());
```
bstansberry: Is there any difference between these two cases?

hostControllerEnvironment.isRestart() || !runningModeControl.isReloadByHandler()

In other words if this op is invoked during a boot where hostControllerEnvironment.isRestart(), won't !runningModeControl.isReloadByHandler() also be true? AIUI isReloadByHandler() is true for any reload and is false for any non-reload boot.

Also, HostControllerEnvironment.isRestart is final, so for the life of an HC process started that way isRestart() will never return 'false' and therefore blockUntilStopped will always be 'true'. Which doesn't seem right.

host-controller/src/main/java/org/jboss/as/host/controller/operations/StartServersHandler.java
```diff
@@ -95,27 +96,47 @@ public void execute(OperationContext context, ModelNode operation) throws Operat
...
+                //Reset runningModeControl values to default for the values that cannot be reset at ProcessReloadHandler.ReloadContext.doReloado
+                runningModeControl.setRestartMode(RestartMode.SERVERS);
+                runningModeControl.setBlockUntilStopped(true);
                 context.completeStep(OperationContext.RollbackHandler.NOOP_ROLLBACK_HANDLER);
             }
         }, OperationContext.Stage.RUNTIME);
     }
 
-    private void cleanStartServers(final ModelNode servers, final ModelNode domainModel, OperationContext context) throws OperationFailedException {
+    private void reconnectAndWaitStoppingServers(final ModelNode servers, final ModelNode domainModel, boolean blockUntilStopped){
```
bstansberry: I believe the 'blockUntilStopped' param is always 'true' and therefore doesn't need to be a param.

host-controller/src/main/java/org/jboss/as/host/controller/ServerInventory.java
```diff
@@ -171,14 +175,32 @@
...
      * @param stopping whether the process is currently stopping
+     * @param blockUntilStopped wait until the server is stopped
+     * @param requestRestart whether the server will be restarted in background once it is stopped
+     * @param restartBlocking whether the server will be restarted in background in a thread waiting until the server is started, otherwise the thread
+     *                        will wait until the server is starting.
+     *
+     * @return Returns {@code true} if the server was selected to be restarted, otherwise it returns false.
      */
-    void reconnectServer(String serverName, ModelNode domainModel, String authKey, boolean running, boolean stopping);
+    boolean reconnectServer(String serverName, ModelNode domainModel, String authKey, boolean running, boolean stopping, boolean blockUntilStopped, boolean requestRestart, boolean restartBlocking);
```
bstansberry: I don't see any code using this return value.

host-controller/src/main/java/org/jboss/as/host/controller/operations/StartServersHandler.java
yersan: I didn't realize that HostControllerEnvironment.isRestart was final. The intention was to wait for servers until stopped always if we were restarting the HC process, but being a final variable, it will affect always independent of isReloadByHandler or isBlockUntilStopped.

@bstansberry Changed to use only a single condition: runningModeControl.isBlockUntilStopped(), it only gets false if the user has used a reload(blocking=false), otherwise, we wait.

host-controller/src/main/java/org/jboss/as/host/controller/operations/StartServersHandler.java
yersan: Now `blockUntilStopped` is always `runningModeControl.isBlockUntilStopped()`, by default always true, false if the user wants, so it makes more sense now,  @bstansberry do you agree?

host-controller/src/main/java/org/jboss/as/host/controller/ServerInventory.java
yersan: You're right, initially, I thought it could be useful to the caller to know if the server was elected to be restarted or not, and indeed I was using it in a previous version, but later it was not longer necessary. Removed.

yersan:
Rebased to resolve conflicts

host-controller/src/main/java/org/jboss/as/host/controller/operations/StartServersHandler.java
```diff
@@ -95,27 +96,47 @@ public void execute(OperationContext context, ModelNode operation) throws Operat
...
+
+    private void cleanStartServers(final ModelNode servers, final ModelNode domainModel, OperationContext context, boolean restartRequired) throws OperationFailedException {
         Map<String, ProcessInfo> processInfos = serverInventory.determineRunningProcesses();
         for(final Property serverProp : servers.asPropertyList()) {
             String serverName = serverProp.getName();
             if (ServerConfigResourceDefinition.AUTO_START.resolveModelAttribute(context, serverProp.getValue()).asBoolean(true)) {
                 ProcessInfo info = processInfos.get(serverInventory.getServerProcessName(serverName));
                 if ( info != null ){
-                    serverInventory.reconnectServer(serverName, domainModel, info.getAuthKey(), info.isRunning(), info.isStopping());
+                    serverInventory.reconnectServer(serverName, domainModel, info.getAuthKey(), info.isRunning(), info.isStopping(), false, restartRequired, START_BLOCKING);
```
yersan: I found an issue here, we can get duplicate registration error because we are only reconnecting here servers which have the auto-start set. For example, if we are reloading with server-three started (auto-start=false), we will get a duplicate registration executing `:start-servers` after reloading, because server-three was not reconnected in the HC reload, I will fix it.

yersan:
Closing for now, finally, we are going to use a different approach. More details here: https://github.com/wildfly/wildfly-proposals/pull/54

PR #3216, opened 2018-04-17
WFCORE-3751 DefaultCapabilityReferenceRecorder generates wrong dependent name if corresponding RuntimeCapability uses a dynamic name mapper

pferraro:
https://issues.jboss.org/browse/WFCORE-3751

### DIFF
--- a/controller/src/main/java/org/jboss/as/controller/CapabilityReferenceRecorder.java
+++ b/controller/src/main/java/org/jboss/as/controller/CapabilityReferenceRecorder.java
@@ -27,6 +27,7 @@ import java.util.Set;
 import java.util.function.Function;
 
 import org.jboss.as.controller.capability.RuntimeCapability;
+import org.jboss.as.controller.logging.ControllerLogger;
 import org.jboss.as.controller.registry.ImmutableManagementResourceRegistration;
 import org.jboss.as.controller.registry.Resource;
 import org.jboss.dmr.ModelNode;
@@ -103,72 +104,33 @@ public interface CapabilityReferenceRecorder {
      * Derives the dependent capability name from the {@code baseDependentName} provided to the constructor, and, if the
      * dependent name is dynamic, from the address of the resource currently being processed.
      */
-    class DefaultCapabilityReferenceRecorder implements CapabilityReferenceRecorder {
+    class DefaultCapabilityReferenceRecorder extends ContextDependencyRecorder {
 
-        private final String baseRequirementName;
         private final String baseDependentName;
-        private final boolean dynamicDependent;
 
-        DefaultCapabilityReferenceRecorder(String baseRequirementName, String baseDependentName, boolean dynamicDependent) {
-            this.baseRequirementName = baseRequirementName;
+        DefaultCapabilityReferenceRecorder(String baseRequirementName, String baseDependentName) {
+            super(baseRequirementName);
             this.baseDependentName = baseDependentName;
-            this.dynamicDependent = dynamicDependent;
+        }
+
+        @Deprecated DefaultCapabilityReferenceRecorder(String baseRequirementName, String baseDependentName, boolean dynamicDependent) {
+            this(baseRequirementName, baseDependentName);
         }
 
         @Override
-        public final void addCapabilityRequirements(OperationContext context, Resource resource, String attributeName, String... attributeValues) {
-            processCapabilityRequirement(context, attributeName, false, attributeValues);
-        }
-
-        @Override
-        public final void removeCapabilityRequirements(OperationContext context, Resource resource, String attributeName, String... attributeValues) {
-            processCapabilityRequirement(context, attributeName, true, attributeValues);
-        }
-
-        private void processCapabilityRequirement(OperationContext context, String attributeName, boolean remove, String... attributeValues) {
-            String dependentName;
-            if (dynamicDependent) {
-                dependentName = RuntimeCapability.buildDynamicCapabilityName(baseDependentName, getDynamicDependentName(context.getCurrentAddress()));
-            } else {
-                dependentName = baseDependentName;
-            }
-
-            for (String attributeValue : attributeValues) {
-                // This implementation does not handle null attribute values
-                if (attributeValue != null) {
-                    String requirementName = RuntimeCapability.buildDynamicCapabilityName(baseRequirementName, attributeValue);
-                    if (remove) {
-                        context.deregisterCapabilityRequirement(requirementName, dependentName);
-                    } else {
-                        context.registerAdditionalCapabilityRequirement(requirementName, dependentName, attributeName);
-                    }
+        protected RuntimeCapability<?> getDependentCapability(OperationContext context) {
+            ImmutableManagementResourceRegistration mrr = context.getResourceRegistration();
+            for (RuntimeCapability<?> capability : mrr.getCapabilities()) {
+                if (this.baseDependentName.equals(capability.getName())) {
+                    return capability;
                 }
             }
-        }
-
-        /**
-         * Determines the dynamic portion of the dependent capability's name. Only invoked if {@code dynamicDependent}
-         * is set to {@code true} in the constructor.
-         * <p>
-         * This base implementation returns the value of the last element in {@code currentAddress}. Subclasses that
-         * wish to extract the relevant name from some other element in the address may override this.
-         * </p>
-         *
-         * @param currentAddress the address of the resource currently being processed. Will not be {@code null}
-         * @return the dynamic portion of the dependenty capability name. Cannot be {@code null}
-         */
-        String getDynamicDependentName(PathAddress currentAddress) {
-            return currentAddress.getLastElement().getValue();
+            throw ControllerLogger.ROOT_LOGGER.unknownCapability(this.baseDependentName);
         }
 
         @Override
         public String getBaseDependentName() {
-            return baseDependentName;
-        }
-
-        @Override
-        public String getBaseRequirementName() {
-            return baseRequirementName;
+            return this.baseDependentName;
         }
     }
 
@@ -182,7 +144,7 @@ public interface CapabilityReferenceRecorder {
      */
     class ContextDependencyRecorder implements CapabilityReferenceRecorder {
 
-        final String baseRequirementName;
+        private final String baseRequirementName;
 
         ContextDependencyRecorder(String baseRequirementName) {
             this.baseRequirementName = baseRequirementName;
@@ -209,7 +171,7 @@ public interface CapabilityReferenceRecorder {
         }
 
         void processCapabilityRequirement(OperationContext context, Resource resource, String attributeName, boolean remove, String... attributeValues) {
-            RuntimeCapability cap = getDependentCapability(context);
+            RuntimeCapability<?> cap = getDependentCapability(context);
             String dependentName = getDependentName(cap, context);
             for (String attributeValue : attributeValues) {
                 //if (attributeValue != null || !cap.getDynamicOptionalRequirements().contains(baseRequirementName)) { //once we figure out what to do with optional requirements
@@ -224,14 +186,14 @@ public interface CapabilityReferenceRecorder {
             }
         }
 
-        protected RuntimeCapability getDependentCapability(OperationContext context) {
+        protected RuntimeCapability<?> getDependentCapability(OperationContext context) {
             ImmutableManagementResourceRegistration mrr = context.getResourceRegistration();
             Set<RuntimeCapability> capabilities = mrr.getCapabilities();
             assert capabilities != null && capabilities.size() == 1;
             return capabilities.iterator().next();
         }
 
-        String getDependentName(RuntimeCapability cap, OperationContext context) {
+        String getDependentName(RuntimeCapability<?> cap, OperationContext context) {
             if (cap.isDynamicallyNamed()) {
                 return cap.fromBaseCapability(context.getCurrentAddress()).getName();
             }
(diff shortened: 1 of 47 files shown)


controller/src/main/java/org/jboss/as/controller/CapabilityReferenceRecorder.java
```diff
@@ -156,8 +165,9 @@ private void processCapabilityRequirement(OperationContext context, String attri
          *
          * @param currentAddress the address of the resource currently being processed. Will not be {@code null}
          * @return the dynamic portion of the dependenty capability name. Cannot be {@code null}
+         * @deprecated No longer used.
          */
-        String getDynamicDependentName(PathAddress currentAddress) {
+        @Deprecated String getDynamicDependentName(PathAddress currentAddress) {
```
jmesnil: this is a (package) method. If it is no longer used, we can delete it instead of deprecating it.

controller/src/main/java/org/jboss/as/controller/CapabilityReferenceRecorder.java
```diff
@@ -126,12 +129,8 @@ public final void removeCapabilityRequirements(OperationContext context, Resourc
...
 
         private void processCapabilityRequirement(OperationContext context, String attributeName, boolean remove, String... attributeValues) {
-            String dependentName;
-            if (dynamicDependent) {
-                dependentName = RuntimeCapability.buildDynamicCapabilityName(baseDependentName, getDynamicDependentName(context.getCurrentAddress()));
-            } else {
-                dependentName = baseDependentName;
-            }
+            RuntimeCapability<?> dependentCapability = this.getDependentCapability(context);
+            String dependentName = (dependentCapability.isDynamicallyNamed() ? dependentCapability.fromBaseCapability(context.getCurrentAddress()) : dependentCapability).getName();
```
jmesnil: The main change is here. 

Before, we were checking if the dependent capability was dynamic (according to the author) to be able to build its name.
Now, we fetch the dependent `RuntimeCapability` according to the current OperationContext MRR and build its name according to the MRR capabilities.

It is similar to the ContextDependencyRecorder implementation (although its `getDependentCapability` impl differs a bit as it checks there is a single registered capability).

My concern is the warning in `ContextDependencyRecorder` javadoc:

> CapabilityReferenceRecorder that determines the dependent capability from the OperationContext. This assumes that the resource registration associated with currently executing step will expose a capability set including one and only one capability. __This recorder cannot be used with attributes associated with resources that do not meet this requirement.__

Your change to the DefaultCapabilityReferenceRecorder generalises the requirement of the ContextDependencyRecorder as the MRR can contain several capabilities and you only use the one corresponding to the `baseDependentName`.

Is my understanding correct?
Is there still a sense to have separate DefaultCapabilityReferenceRecorder and ContextDependencyRecorder as they now both require a Context MRR to work.
We could have a special case for attribute-based capability but that's a special case of your general DefaultCapabilityReferenceRecorder now, right?

controller/src/main/java/org/jboss/as/controller/CapabilityReferenceRecorder.java
```diff
@@ -146,6 +145,16 @@ private void processCapabilityRequirement(OperationContext context, String attri
...
         }
 
+        private RuntimeCapability<?> getDependentCapability(OperationContext context) {
+            ImmutableManagementResourceRegistration mrr = context.getResourceRegistration();
+            for (RuntimeCapability<?> capability : mrr.getCapabilities()) {
+                if (this.baseDependentName.equals(capability.getName())) {
+                    return capability;
+                }
+            }
+            throw ControllerLogger.ROOT_LOGGER.unknownCapability(this.baseDependentName);
```
jmesnil: Should we pass the OperationContext address to the exception so that the users can identify which resource is causing the issue?

domain-management/src/main/java/org/jboss/as/domain/management/access/AccessIdentityResourceDefinition.java
```diff
@@ -73,7 +73,7 @@
     public static final SimpleAttributeDefinition SECURITY_DOMAIN = new SimpleAttributeDefinitionBuilder(ModelDescriptionConstants.SECURITY_DOMAIN, ModelType.STRING, false)
             .setMinSize(1)
             .setFlags(AttributeAccess.Flag.RESTART_RESOURCE_SERVICES)
-            .setCapabilityReference(SECURITY_DOMAIN_CAPABILITY, MANAGEMENT_IDENTITY_CAPABILITY, false)
+            .setCapabilityReference(SECURITY_DOMAIN_CAPABILITY, MANAGEMENT_IDENTITY_CAPABILITY)
```
jmesnil: juste a note that I much prefer to let the CapabilityReference determines where the dependent capability is dynamic or not according to the registered RuntimeCapability instead to have to specify it myself when I set the capability reference :)

domain-management/src/main/java/org/jboss/as/domain/management/access/AccessIdentityResourceDefinition.java
```diff
@@ -82,9 +82,10 @@
     private AccessIdentityResourceDefinition(AbstractAddStepHandler add) {
         super(new Parameters(PATH_ELEMENT, DomainManagementResolver.getResolver("core.identity"))
                 .setAddHandler(add)
-                .setRemoveHandler(new ReloadRequiredRemoveStepHandler(MANAGEMENT_IDENTITY_RUNTIME_CAPABILITY))
+                .setRemoveHandler(ReloadRequiredRemoveStepHandler.INSTANCE)
                 .setAddRestartLevel(OperationEntry.Flag.RESTART_NONE)
                 .setRemoveRestartLevel(OperationEntry.Flag.RESTART_RESOURCE_SERVICES)
+                .setCapabilities(MANAGEMENT_IDENTITY_RUNTIME_CAPABILITY)
```
jmesnil: is your change equivalent?

Capability added to the ReloadRequiredRemoveStepHandler will be handled in org.jboss.as.controller.AbstractRemoveStepHandler#recordCapabilitiesAndRequirements that does a whole lot more that when you set it on the resource definition where it is handled in org.jboss.as.controller.SimpleResourceDefinition#registerCapabilities.

I agree with your change in principle as I think the capabilities belongs to the resource and not to its add/remove operation. However, are we missing some process from org.jboss.as.controller.AbstractRemoveStepHandler#recordCapabilitiesAndRequirements in doing so?

host-controller/pom.xml
```diff
@@ -120,6 +120,10 @@
            <artifactId>junit</artifactId>
            <scope>test</scope>
         </dependency>
+        <dependency>
+            <groupId>org.mockito</groupId>
+            <artifactId>mockito-core</artifactId>
```
jmesnil: should it be in `<scope>test</scope>` ?

host-controller/src/test/java/org/jboss/as/domain/controller/operations/ReloadRequiredServerTestCase.java
```diff
@@ -518,6 +522,10 @@ public void execute(ModelNode operation, OperationMessageHandler handler, ProxyO
         private OperationStepHandler nextStep;
         protected MockOperationContext(final Resource root, final boolean booting, final PathAddress operationAddress) {
             super(root, booting, operationAddress);
+            Set<RuntimeCapability> capabilities = new HashSet<>();
```
jmesnil: That goes back to my concern in the `DefaultCapabilityReferenceRecorder` implementation that now requires to get info from the MRR.

I am not sure if that's a constraint that will always be there. (it used to be absent in this test fixture for example).

jmesnil:
@pferraro After finishing my review, your PR looks right except for the added requirement to always get RuntimeCapability info from the MRR in the DefaultCapabilityReferenceRecorder implementation.

@ehsavoie @bstansberry Do you have an use case where this would *not* be the case?

The tests were such a case but I don't know if that was by design or just because it was not needed to get info from the MRR.

controller/src/main/java/org/jboss/as/controller/CapabilityReferenceRecorder.java
pferraro: That's true.  I'll remove it.

controller/src/main/java/org/jboss/as/controller/CapabilityReferenceRecorder.java
pferraro: Your understanding is correct.  In retrospect, DefaultCapabilityReferenceRecorder should probably extend ContextDependencyRecorder, and just override its logic for generating the dependent name (i.e. from a specific baseDependentName).

> Is there still a sense to have separate DefaultCapabilityReferenceRecorder and ContextDependencyRecorder

A valid question.  Personally, I didn't even know that ContextDependencyRecorder existed until this week :) - so I don't know how many people are using it.
ContextDependencyRecorder requires fewer parameters, which is nice, but it can't be used if the resource defines multiple capabilities.
I suspect it is worth having both - but we should at least consolidate the implementations.

Another logical implementation is one that accepts the RuntimeCapability itself (this is how the CapabilityReferenceRecorder in wildfly-clustering-common works), which simplifies the getDependentCapability(...) even further (and without an exception case), but that is probably out of scope for this fix.

> We could have a special case for attribute-based capability

What do you mean by attribute-based capability?

host-controller/pom.xml
pferraro: Oops - yes, absolutely.

domain-management/src/main/java/org/jboss/as/domain/management/access/AccessIdentityResourceDefinition.java
pferraro: > is your change equivalent?

Yes - the resulting behavior is exactly the same.
See: https://github.com/wildfly/wildfly-core/blob/master/controller/src/main/java/org/jboss/as/controller/AbstractRemoveStepHandler.java#L185

host-controller/src/test/java/org/jboss/as/domain/controller/operations/ReloadRequiredServerTestCase.java
pferraro: Supplying capabilities to the add/remove operation handler is redundant.

controller/src/main/java/org/jboss/as/controller/CapabilityReferenceRecorder.java
pferraro: We could - but unless the same capability is defined by multiple resources (which is rare), the problematic resource is usually obvious.

controller/src/main/java/org/jboss/as/controller/CapabilityReferenceRecorder.java
pferraro: Ah - maybe you mean CompositeAttributeDependencyRecorder?

Yeah - to be more universal, I would expect this to be a special case of DefaultCapabilityReferenceRecorder, not ContextDependencyRecorder.

domain-management/src/main/java/org/jboss/as/domain/management/access/AccessIdentityResourceDefinition.java
jmesnil: ah, right.

So my question is not specific to your PR but more generally why have we 2 different ways to specify capabilities (either in add/remove handler or in the resource definition)?
In any case, your PR is fine wrt that issue

kabir:
@jmesnil since you've already been involved I'll leave the approval of this one to you :)

domain-management/src/main/java/org/jboss/as/domain/management/access/AccessIdentityResourceDefinition.java
bstansberry: @jmesnil That's just a matter of code history; in the initial cap/req code there was no MRR.getCapabilities(); the relevant caps were stored in the 'capabilities' field in the add/remove handlers.

bstansberry:
@jmesnil The main way the cap would not be in the MRR is if people relied on the old code that passed the relevant caps to the Abstract[Add|Remove]StepHandler constructor and didn't record the cap with the MRR.

That would largely be a bug, as not recording with the MRR means the possible capability info is not available. I don't think it's done in WF code but it might be done in downstream projects that haven't kept up with best practices.

I say "largely a bug" because we've contemplated having a capability only registered if an attribute is set to a particular value. For example jts=true in subsystem=transactions, when we get around to doing caps in that subsystem. But:

1) I think using an attribute that way is wrong. Registering an actual capability should simply be due to adding the resource. The tx subsystem should use a child resource for jts. (Note this rule means the new management-api javadoc at https://github.com/bstansberry/wildfly-core/blob/new-mgmt-api/management-api/src/main/java/org/wildfly/management/api/capability/CapabilityRegistrar.java#L29 needs to be changed, as there's no such thing as an "optional capability".)

2) Even if we say using an attribute that way is ok it doesn't mean the MRR shouldn't mention the cap and thus expose it as a possible capability. It just means the add handler would need to specially handle registering the actual capability.

@pferraro if we do this we need to deprecate all the add/remove handler constructors in the controller module that accept a RuntimeCapability param. Their presence obscures that the caps to register really should come from the MRR.

Finally, I think this will avoid breaking any downstream code that isn't properly registering the cap with the MRR:

https://github.com/wildfly/wildfly-core/commit/d98a56377f9e807b00ab2d8d4ad56fc41d9b3baa

ehsavoie:
AbstractBoottimeAddStepHandler, ReloadRequiredAddStepHandler, ReloadRequiredRemoveStepHandler , ServiceRemoveStepHandler  constructors with capabilities should be deprecated too.
Otherwise +1

pferraro:
@jmesnil @ehsavoie Suggested changes incorporated.

jmesnil:
@pferraro I'm looking at merging this PR. 
It seems you have not cherry-picked (or squashed) https://github.com/wildfly/wildfly-core/commit/d98a56377f9e807b00ab2d8d4ad56fc41d9b3baa as brian proposed. Is that correct?

pferraro:
@jmesnil I've cherry-picked it into this PR.

PR #3218, opened 2018-04-19
[WFCORE-3757] Fix RBAC test suite

jmesnil:
Do not use @Suite anymore to run the LDAP variants of the
PropertiesRoleMappingTestCase test as the 3 executions were producing
the same XML report.

The actual tests have been moved to AbstractPropertiesRoleMappingTestCase.

* PropertiesRoleMappingTestCase adds a fixture to start/stop WildFly
   container before the tests
* LdapRoleMappingG2UTestCase and LdapRoleMappingU2GTestCase add a
  fixture to start/stop the LDAP server and WildFly container before the
  tests

Maven correctly reports that there are failures in the LDAP tests but
since there is a single XML report, any information from other execution
is lost.
Our CI was also only reporting only 1 execution for the
PropertiesRoleMappingTestCase test.

Now the LDAP tests are run as normal tests (no longer in a @Suite) and
reports their execution in their respective XML files.

JIRA: https://issues.jboss.org/browse/WFCORE-3757


jmesnil:
CI jobs for this PR should report 2 new tests for LdapRoleMappingG2UTestCase and LdapRoleMappingU2GTestCase.

PR #3219, opened 2018-04-19
[WFCORE-3500] Upgrade javax.json from 1.0.4 to 1.1.2.

jamezp:
https://issues.jboss.org/browse/WFCORE-3500

Java EE 8 requires `javax.json:javax.json-api` upgrade to 1.1. This was already done in WildFly https://github.com/wildfly/wildfly/pull/10872. This just upgrades core to the new version as before we move to full Java EE 8 compatibility this will be required. There should not be backwards compatibility issues and this will allow core to be tested with the newer version.

There is also a possible issue where the wrong version of the temporary `org.wildfly.javax.json` module could be overridden in Galleon from the version in full.

### DIFF
--- a/component-matrix/pom.xml
+++ b/component-matrix/pom.xml
@@ -61,6 +61,7 @@
         <version.commons-lang>2.6</version.commons-lang>
         <version.io.undertow>2.0.4.Final</version.io.undertow>
         <version.javax.inject.javax.inject>1</version.javax.inject.javax.inject>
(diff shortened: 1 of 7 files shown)


jamezp:
@jmesnil I wasn't going to fully remove the 1.0.4 reference until we are fully Java EE 8 compliant. So for WildFly full we still need the 1.0.4 version. We could change the version in the client, but I don't think it's required.

PR #3227, opened 2018-04-23
Adding test case for ReadFeatureDescriptionHandler

TomasHofman:
https://issues.jboss.org/browse/WFCORE-3742

### DIFF
--- a/controller/src/main/java/org/jboss/as/controller/operations/global/ReadFeatureDescriptionHandler.java
+++ b/controller/src/main/java/org/jboss/as/controller/operations/global/ReadFeatureDescriptionHandler.java
@@ -21,76 +21,78 @@
  */
 package org.jboss.as.controller.operations.global;
 
-import java.util.ArrayList;
 import static org.jboss.as.controller.descriptions.ModelDescriptionConstants.ADDRESS;
+import static org.jboss.as.controller.descriptions.ModelDescriptionConstants.ADDR_PARAMS;
+import static org.jboss.as.controller.descriptions.ModelDescriptionConstants.ADDR_PARAMS_MAPPING;
 import static org.jboss.as.controller.descriptions.ModelDescriptionConstants.ANNOTATION;
 import static org.jboss.as.controller.descriptions.ModelDescriptionConstants.ATTRIBUTES;
 import static org.jboss.as.controller.descriptions.ModelDescriptionConstants.CAPABILITY_REFERENCE;
+import static org.jboss.as.controller.descriptions.ModelDescriptionConstants.CAPABILITY_REFERENCE_PATTERN_ELEMENTS;
 import static org.jboss.as.controller.descriptions.ModelDescriptionConstants.CHILDREN;
+import static org.jboss.as.controller.descriptions.ModelDescriptionConstants.COMPLEX_ATTRIBUTE;
 import static org.jboss.as.controller.descriptions.ModelDescriptionConstants.DEFAULT;
 import static org.jboss.as.controller.descriptions.ModelDescriptionConstants.EXCEPTIONS;
 import static org.jboss.as.controller.descriptions.ModelDescriptionConstants.EXTENSION;
 import static org.jboss.as.controller.descriptions.ModelDescriptionConstants.FAILURE_DESCRIPTION;
 import static org.jboss.as.controller.descriptions.ModelDescriptionConstants.FEATURE;
+import static org.jboss.as.controller.descriptions.ModelDescriptionConstants.FEATURE_ID;
 import static org.jboss.as.controller.descriptions.ModelDescriptionConstants.FEATURE_REFERENCE;
 import static org.jboss.as.controller.descriptions.ModelDescriptionConstants.HOST;
 import static org.jboss.as.controller.descriptions.ModelDescriptionConstants.INCLUDE;
 import static org.jboss.as.controller.descriptions.ModelDescriptionConstants.NAME;
 import static org.jboss.as.controller.descriptions.ModelDescriptionConstants.NILLABLE;
 import static org.jboss.as.controller.descriptions.ModelDescriptionConstants.OP;
+import static org.jboss.as.controller.descriptions.ModelDescriptionConstants.OP_PARAMS;
+import static org.jboss.as.controller.descriptions.ModelDescriptionConstants.OP_PARAMS_MAPPING;
+import static org.jboss.as.controller.descriptions.ModelDescriptionConstants.OPTIONAL;
 import static org.jboss.as.controller.descriptions.ModelDescriptionConstants.OP_ADDR;
+import static org.jboss.as.controller.descriptions.ModelDescriptionConstants.PACKAGE;
+import static org.jboss.as.controller.descriptions.ModelDescriptionConstants.PACKAGES;
+import static org.jboss.as.controller.descriptions.ModelDescriptionConstants.PARAMS;
 import static org.jboss.as.controller.descriptions.ModelDescriptionConstants.PROFILE;
+import static org.jboss.as.controller.descriptions.ModelDescriptionConstants.PROVIDES;
 import static org.jboss.as.controller.descriptions.ModelDescriptionConstants.READ;
+import static org.jboss.as.controller.descriptions.ModelDescriptionConstants.READ_FEATURE_DESCRIPTION_OPERATION;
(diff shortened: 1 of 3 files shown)


controller/src/test/java/org/jboss/as/controller/test/ReadFeatureDescriptionTestCase.java
```diff
@@ -0,0 +1,623 @@
+package org.jboss.as.controller.test;
```
bstansberry: This should have the ASL copyright header.

controller/src/test/java/org/jboss/as/controller/test/ReadFeatureDescriptionTestCase.java
```diff
@@ -0,0 +1,623 @@
...
+import static org.jboss.as.controller.descriptions.ModelDescriptionConstants.READ_FEATURE_DESCRIPTION_OPERATION;
+import static org.jboss.as.controller.descriptions.ModelDescriptionConstants.RECURSIVE;
+import static org.jboss.as.controller.descriptions.ModelDescriptionConstants.REQUIRES;
+import static org.jboss.as.controller.descriptions.ModelDescriptionConstants.SUBSYSTEM;
+import static org.jboss.as.controller.operations.global.ReadFeatureDescriptionHandler.ADDRESS_PARAMETERS;
+import static org.jboss.as.controller.operations.global.ReadFeatureDescriptionHandler.FEATURE_ID;
+import static org.jboss.as.controller.operations.global.ReadFeatureDescriptionHandler.OPERATION_PARAMETERS;
+import static org.jboss.as.controller.operations.global.ReadFeatureDescriptionHandler.OPERATION_PARAMETERS_MAPPING;
+import static org.jboss.as.controller.operations.global.ReadFeatureDescriptionHandler.PARAMETERS;
+import static org.jboss.as.controller.operations.global.ReadFeatureDescriptionHandler.REFERENCES;
```
bstansberry: Please use the standard WildFly import ordering:

import static

java
javax

others (including org.jboss etc)

If it's not that way it will just end up getting reordered some time when someone edits the file. :)

controller/src/main/java/org/jboss/as/controller/operations/global/ReadFeatureDescriptionHandler.java
```diff
@@ -122,16 +122,16 @@
...
+    public static final String $PROFILE_PREFIX = "$profile.";
+    public static final String ADDRESS_PARAMETERS = "addr-params";
+    public static final String ADDRESS_PARAMETERS_MAPPING = "addr-params-mapping";
+    public static final String DOMAIN_EXTENSION = "domain.extension";
+    public static final String FEATURE_ID = "feature-id";
+    public static final String HOST_EXTENSION = "host.extension";
+    public static final String OPERATION_PARAMETERS = "op-params";
+    public static final String OPERATION_PARAMETERS_MAPPING = "op-params-mapping";
+    public static final String PARAMETERS = "params";
+    public static final String REFERENCES = "refs";
```
jmesnil: All these constants should remain private to the ReadFeatureDescriptionHandler.java class.
If some needs to be referenced from the test (e.g. `ADDRESS_PARAMETERS`), they should be moved to `org.jboss.as.controller.descriptions.ModelDescriptionConstants`.

bstansberry:
Nice. :)

I added some tests in https://github.com/wildfly/wildfly-core/pull/3226 because I wanted to be sure my code cleanup in the other commit in that PR didn't break something. But I think most of what's in those tests can be removed, as testing in the controller module like you do here is much better.

For integration testing it's probably ok to just do a recursive read of the full tree. Basically 1) test a real remote client, which in-vm tests in controller module don't do, and 2) read the whole tree as a kind of smoke test.  Actually doing that in full WildFly is probably a good idea too, as most of the features are in full not core.

If you already have stuff going to do that kind of integration test then I can just drop the commit in my PR.

@ehsavoie You should probably review this PR as you best understand what the details of the output are meant to be.

controller/src/test/java/org/jboss/as/controller/test/ReadFeatureDescriptionTestCase.java
```diff
@@ -0,0 +1,623 @@
...
+        Assert.assertEquals(serializeAddress(address), feature.require(NAME).asString());
+
+        // check params
+        Map<String, ModelNode> params = extractParamsToMap(feature);
+        Assert.assertEquals(2, params.size());
+        assertFeatureIdParam(SUBSYSTEM, TEST_SUBSYSTEM, params); // path parameter
+        assertParamWithDefault(EXTENSION, TEST_EXTENSION, params); // referencing extension
+
+        // check provides
+        ModelNode provides = feature.require("provides");
```
jmesnil: likewise if some constants are missing that are used in the model of the `read-feature` operation, they should be introduced in `org.jboss.as.controller.descriptions.ModelDescriptionConstants` and used in org.jboss.as.controller.operations.global.ReadFeatureDescriptionHandler

controller/src/test/java/org/jboss/as/controller/test/ReadFeatureDescriptionTestCase.java
```diff
@@ -0,0 +1,623 @@
...
+
+    private ModelNode readFeature(PathAddress address) throws OperationFailedException {
+        return readFeature(address, false);
+    }
+
+    private ModelNode readFeature(PathAddress address, boolean recursive) throws OperationFailedException {
+        ModelNode operation = createOperation(READ_FEATURE_DESCRIPTION_OPERATION, address);
+        operation.get(RECURSIVE).set(recursive);
+        ModelNode result = executeForResult(operation);
+        System.out.println(result);
```
bstansberry: Please remove or comment this out. We want to limit disk IO as it adds a lot of burden to our CI servers where in the end disk is a shared resource amongst many VMs.

jmesnil:
@TomasHofman thanks, the tests look good :)

We just need to avoid making public constants for things that are not meant to be used outside of the class.
Our typical usage is to put constants that could be used by management clients in ModelDescriptionConstants.

controller/src/main/java/org/jboss/as/controller/operations/global/ReadFeatureDescriptionHandler.java
```diff
@@ -122,16 +122,16 @@
...
-    private static final String ADDRESS_PARAMETERS = "addr-params";
-    private static final String ADDRESS_PARAMETERS_MAPPING = "addr-params-mapping";
-    private static final String DOMAIN_EXTENSION = "domain.extension";
-    private static final String FEATURE_ID = "feature-id";
-    private static final String HOST_EXTENSION = "host.extension";
-    private static final String OPERATION_PARAMETERS = "op-params";
-    private static final String OPERATION_PARAMETERS_MAPPING = "op-params-mapping";
-    private static final String PARAMETERS = "params";
-    private static final String REFERENCES = "refs";
+    public static final String $PROFILE_PREFIX = "$profile.";
```
jmesnil: That's not related to the PR per se but I am curious 

@aloubyansky @ehsavoie Starting the field name with `$` is legal in Java but it's the first time I see it. Is there a good reason to use this name instead of plain `PROFILE_PREFIX`?

TomasHofman: OK, I'm gonna fix that too :).

TomasHofman:
@bstansberry @jmesnil thank you for reviews! All issues should be fixed.

I moved couple of constants into ModelDescriptionConstants. I kept the original constant names, which however in some cases aren't verbatim reflections of the strings they represent, e.g.

```
public static final String REFERENCES = "refs"; 
public static final String PARAMETERS = "params"; 
public static final String ADDRESS_PARAMETERS = "addr-params"; 
```

I wonder if I should change the constant names to verbatim transcriptions of the strings to avoid future conflicts. Some are general enough that someone might need to add constants also for "references" or "parameters" in future...

TomasHofman:
@bstansberry I can definitely create the integration test in full. Will ping you when it's done.

aloubyansky:
Or maybe simply call them what they are, i.e. REFS, PARAMS and ADDR_PARAMS.

TomasHofman:
@aloubyansky yes, that's what I meant. I will change that once again then. :)

PR #3229, opened 2018-04-26
[WFCORE-3765] Explicitly list all dependencies, organizing them by ho…

bstansberry:
…w they are used. Use 'provided' scope for dependencies that are not required to be visible to the embedding app's classloader at runtime.

Also, drop use of WildFlySecurityManager to avoid the requirement for elytron on the embedding app's classpath.

Also, drop use of wildfly-common Assert, even though it's only required on the modular classloader side, just to cut the pom footprint and help avoid it leaking to the embedded app side.

https://issues.jboss.org/browse/WFCORE-3765

This will conflict with #3202. I planned to wait on this and fix conflicts when #3202 is merged. But if there are going to be delays on #3202, which is a Feature, then this can go first.

### DIFF
--- a/embedded/src/main/java/org/wildfly/core/embedded/EmbeddedHostControllerFactory.java
+++ b/embedded/src/main/java/org/wildfly/core/embedded/EmbeddedHostControllerFactory.java
@@ -60,9 +60,7 @@ import org.jboss.msc.service.ServiceContainer;
 import org.jboss.msc.service.ServiceController;
 import org.jboss.msc.value.Value;
 import org.jboss.stdio.StdioContext;
-import org.wildfly.common.Assert;
 import org.wildfly.core.embedded.logging.EmbeddedLogger;
-import org.wildfly.security.manager.WildFlySecurityManager;
 
 /**
  * This is the host controller counterpart to EmbeddedServerFactory which lives behind a module class loader.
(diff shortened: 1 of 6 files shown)


embedded/src/main/java/org/wildfly/core/embedded/EmbeddedHostControllerFactory.java
```diff
@@ -98,11 +96,11 @@ private EmbeddedHostControllerFactory() {
...
-        Assert.checkNotNullParam("jbossHomeDir", jbossHomeDir);
-        Assert.checkNotNullParam("moduleLoader", moduleLoader);
-        Assert.checkNotNullParam("systemProps", systemProps);
-        Assert.checkNotNullParam("systemEnv", systemEnv);
-        Assert.checkNotNullParam("cmdargs", cmdargs);
+        assert jbossHomeDir != null : "jbossHomeDir is null";
+        assert moduleLoader != null : "moduleLoader is null";
+        assert systemProps != null : "systemProps is null";
+        assert systemEnv != null : "systemEnv is null";
+        assert cmdargs != null : "cmdargs is null";
```
jamezp: These methods are public but generally shouldn't be used by consumers of the API. Though if they are these assertions will only work if the user enables assertions. Since this, and the `EmbeddedStandaloneFactory`, are only used internally by the `ManagedProcessFactory` this might not matter.

bstansberry:
@jamezp re the 'assert' stuff I'd originally thought to switch those to nullVar once your PR restoring that method was in. But I realized waiting is silly; my restoring it and using it here shouldn't cause a conflict. So I did.

jamezp:
@bstansberry No problem and makes sense. I can rebase my PR on top of this once it's merged. This one is more important IMO.

PR #3230, opened 2018-04-26
[WFCORE-3608] Add core pool size attribute to IO subsystem

dmlloyd:
https://issues.jboss.org/browse/WFCORE-3608

### DIFF
--- a/component-matrix/pom.xml
+++ b/component-matrix/pom.xml
@@ -115,7 +115,7 @@
         <version.org.wildfly.build-tools>1.2.10.Final</version.org.wildfly.build-tools>
         <version.org.wildfly.checkstyle-config>1.0.5.Final</version.org.wildfly.checkstyle-config>
         <version.org.wildfly.client.config>1.0.0.Final</version.org.wildfly.client.config>
-        <version.org.wildfly.common>1.3.1.Final</version.org.wildfly.common>
+        <version.org.wildfly.common>1.3.2.Final</version.org.wildfly.common>
         <version.org.wildfly.discovery>1.0.0.Final</version.org.wildfly.discovery>
         <version.org.wildfly.legacy.test>3.0.0.Final</version.org.wildfly.legacy.test>
         <version.org.wildfly.openssl>1.0.6.Final</version.org.wildfly.openssl>


dmlloyd:
Supercedes #3107

dmlloyd:
I think I got some extra changes in here by mistake.  Checking...

io/subsystem/src/main/java/org/wildfly/extension/io/IOExtension.java
```diff
@@ -65,14 +65,15 @@ public void initializeParsers(ExtensionParsingContext context) {
         context.setSubsystemXmlMapping(SUBSYSTEM_NAME, Namespace.IO_1_0.getUriString(), IOSubsystemParser_1_0::new);
         context.setSubsystemXmlMapping(SUBSYSTEM_NAME, Namespace.IO_1_1.getUriString(), IOSubsystemParser_1_1::new);
         context.setSubsystemXmlMapping(SUBSYSTEM_NAME, Namespace.IO_2_0.getUriString(), IOSubsystemParser_2_0::new);
+        context.setSubsystemXmlMapping(SUBSYSTEM_NAME, Namespace.IO_3_0.getUriString(), new IOSubsystemParser_3_0());
```
dmlloyd: Fixed @jmesnil's comment about directly instantiating the latest parser.

io/subsystem/src/main/java/org/wildfly/extension/io/IOSubsystemTransformers.java
```diff
@@ -44,10 +47,19 @@ public String getSubsystemName() {
     public void registerTransformers(SubsystemTransformerRegistration registration) {
         ChainedTransformationDescriptionBuilder chainedBuilder = TransformationDescriptionBuilder.Factory.createChainedSubystemInstance(registration.getCurrentSubsystemVersion());
 
-        // Current 3.0.0 to 2.0.0, aka EAP 7.0.0
-        buildTransformers_2_0(chainedBuilder.createBuilder(registration.getCurrentSubsystemVersion(), VERSION_2_0));
+        // Current 3.0.0 to 2.0.0, aka EAP 7.1.0
```
dmlloyd: To @jmesnil's comment on #3107 - I think I've untangled the knot here but I'm not 100% sure.  A transformers expert should check this out to be sure.

io/tests/src/test/java/org/wildfly/extension/io/IOSubsystemTransformerTestCase.java
```diff
@@ -53,7 +52,7 @@ public IOSubsystemTransformerTestCase() {
 
     @Override
     protected String getSubsystemXml() throws IOException {
-        return readResource("io-2.0-transformer.xml");
+        return readResource("io-3.0-transformer.xml");
```
dmlloyd: I kept the original name scheme of the XML resources.

io/subsystem/src/main/resources/org/wildfly/extension/io/LocalDescriptions.properties
```diff
@@ -4,6 +4,8 @@ io.remove=Operation Removes subsystem
 io.worker=Defines workers
 io.worker.add=Adds new worker
 io.worker.remove=Removes worker
+io.worker.task-core-threads=Specify the starting number of threads for the worker task thread pool.\
+  If not set, it defaults to 2.
```
dmlloyd: Fixed in accordance with @rsvoboda's comment on https://github.com/wildfly/wildfly-core/pull/3107#discussion_r172507513.

dmlloyd:
OK should be all fixed now, pending clean CI

dmlloyd:
Looks like the transformer fixed were not correct. Will look into it tomorrow.

io/subsystem/src/main/java/org/wildfly/extension/io/IOSubsystemTransformers.java
```diff
@@ -44,10 +47,19 @@ public String getSubsystemName() {
...
+        buildTransformers_3_0(chainedBuilder.createBuilder(VERSION_3_0, VERSION_2_0));
+        // 2.0.0 to 1.0.0, aka EAP 7.0.0
+        buildTransformers_2_0(chainedBuilder.createBuilder(VERSION_2_0, VERSION_1_0));
 
-        chainedBuilder.buildAndRegister(registration, new ModelVersion[]{VERSION_2_0});
+        chainedBuilder.buildAndRegister(registration, new ModelVersion[]{ VERSION_3_0, VERSION_2_0 });
```
jmesnil: That does not seem correct, the version chain goes from the oldest to the newest ("current version excluded").
It should be:

```
chainedBuilder.buildAndRegister(registration, new ModelVersion[]{ VERSION_2_0, VERSION_3_0 });
```

io/tests/src/test/java/org/wildfly/extension/io/IOSubsystemTransformerTestCase.java
```diff
@@ -68,11 +67,14 @@ public void testTransformerEAP710() throws Exception {
     }
 
     private KernelServices buildKernelServices(ModelTestControllerVersion controllerVersion, ModelVersion version, String... mavenResourceURLs) throws Exception {
-        return this.buildKernelServices(this.getSubsystemXml(), controllerVersion, version, mavenResourceURLs);
+        String xmlName = String.format("eap-%s-transformer.xml", controllerVersion.getRealVersionName());
```
jmesnil: The test fails because there is no `eap-7.0.0-transformer.xml` or `eap-7.1.0-transformer.xml` files. 

I now understand why @ctomc renamed the `io-${subsystem-version}-transformer.xml` to `eap-${eap-version}-transformer.xml` in his PR :)

io/subsystem/src/main/java/org/wildfly/extension/io/IOSubsystemTransformers.java
jmesnil: I think the error in `IOSubsystemTransformerTestCase.testRejectingTransformersEAP_7_1_0` is caused by the wrong version chain order.

dmlloyd: This didn't seem to fix the problem... I thought the transformers were meant to go from newer to older though (as per the comments)?  Maybe I need to swap the order in createBuilder() as well?

io/tests/src/test/java/org/wildfly/extension/io/IOSubsystemTransformerTestCase.java
dmlloyd: I fixed this using a switch to set up the correct IO version.

dmlloyd:
So... what would you guys say if I were restore the transformer tests exactly as they were with the original PR and to @Ignore them, with a new issue to resolve the problem which can be worked by someone who has a might higher knowledge-of-transformers:list-of-backed-up-work ratio than I do?

rsvoboda:
We shouldn't merge stuff which can have backward compatibility issues.

@jmesnil, any idea who could help with transformers stuff ?

kabir:
@dmlloyd @rsvoboda I'll take a look at the transformers

io/subsystem/src/main/resources/org/wildfly/extension/io/LocalDescriptions.properties
kabir: @rsvoboda @dmlloyd I disagree - the default-value in the r-r-d output already contains this information

dmlloyd:
Holding until transformer tests are added.

dmlloyd:
Should be all fixed now.

dmlloyd:
@jmesnil is this OK now?

io/tests/src/test/java/org/wildfly/extension/io/IOSubsystemTestCase.java
```diff
@@ -108,26 +104,5 @@ protected RunningMode getRunningMode() {
...
     }
 
-    protected static final OptionAttributeDefinition ENABLED_PROTOCOLS = OptionAttributeDefinition.builder("enabled-protocols", Options.SSL_ENABLED_PROTOCOLS)
-            .setRequired(false)
-            .setFlags(AttributeAccess.Flag.RESTART_ALL_SERVICES)
-            .setAllowExpression(true)
-            .build();
-
-    @Test
-    public void testSequence() throws Exception {
```
jmesnil: why did you remove that test?

jmesnil:
@dmlloyd the PR looks good except for the removed test `testSequence`. It seems unrelated to the new attribute, why did you remove it?

io/tests/src/test/java/org/wildfly/extension/io/IOSubsystemTestCase.java
dmlloyd: I think this must have been removed as a part of the original PR.  I could try adding it back.

dmlloyd:
@jmesnil OK all passed with the missing test restored.

jmesnil:
@dmlloyd it seems you pushed the wrong commit to this PR. it now contains the commit to upgrade  wildfly common 1.3.2.Final

PR #3231, opened 2018-04-27
[WFCORE-3769] Use Galleon for the WildFly Core build

aloubyansky:
This change replaces the current build tools used to produce the build and dist directories containing distributions as well the wildfly-core distributions built for the testsuite.
The original feature-packs are still built and installed under the same group and artifact Ids.

The legacy build (based on the original feature-packs) is built in module build-legacy by activating profile legacyBuild with -DlegacyBuild.

The legacy dist (based on the original feature-packs) is built in module build-dist by activating profile legacyRelease with -DlegacyRelease.


aloubyansky:
The error is related to the use of the old maven version in the CI "Current M2_HOME = /opt/buildAgent/tools/maven3_3". The current maven version used to build WildFly is 3.5.2.

core-galleon-pack/src/main/resources/content/README.txt
```diff
@@ -0,0 +1,83 @@
+ _    _  _  _      _ ______  _           __   _____
+| |  | |(_)| |    | ||  ___|| |         /  | |  _  |
+| |  | | _ | |  __| || |_   | | _   _   `| | | |/' |
+| |/\| || || | / _` ||  _|  | || | | |   | | |  /| |
+\  /\  /| || || (_| || |    | || |_| |  _| |_\ |_/ /
+ \/  \/ |_||_| \__,_|\_|    |_| \__, |  \___/ \___/
```
kabir: Perhaps we should remove the 10 part here

bstansberry: If we do we must remove it in the same file in core-feature-pack as well, as a goal of this PR is to produce a build where the galleon build output is the same as the legacy build.

If Alexey wants to do that, ok, but in general I don't think we should make fixing existing problems a requirement of this PR.

bstansberry:
The StandardConfigsXMLValidationUnitTestCase failures are interesting. They fail because the configs in docs/examples/configs are not there. I don't get why that would happen, since those are created in full and full would still be using the legacy build.

aloubyansky:
Full is built using Galleon. The examples configs are not in the docs/examples/configs they need to move there. I'm gonna do that.

aloubyansky:
Ah, you meant full in this integration. Well, we have those failures in the full galleon branch too.

bstansberry:
For the record, the test results for the jobs I'll link in this comment previously were mysteriously appearing in the CI result block below. Those are what led to my comment about StandardConfigsXmlValidationUnitTestCase. But these jobs are not standard CI jobs and are not actually tests of this PR, so the discussion of them is not relevant.

https://ci.wildfly.org/viewLog.html?buildId=100306&tab=buildResultsDiv&buildTypeId=WildFlyCore_WildFlyCoreFullGalleonIntegrationLinux

https://ci.wildfly.org/viewLog.html?buildId=100305&tab=buildResultsDiv&buildTypeId=WildFlyCore_WildFlyCoreFullGalleonIntegrationWindows

@luck3y FYI

bstansberry:
@aloubyansky when doing a diff of the build and build-legacy output I get a huge diff on the Windows stuff in the bin dir. I think this is a line ending thing, as these files have CRLF in build-legacy but not in build:

add-user.bat
domain.bat
domain.conf.bat
elytron-tool.bat
jboss-cli.bat
standalone.bat
standalone.conf.bat
vault.bat

Looking at the source in core-feature-pack and core-galleon-pack in neither case do the files have CRLF, so there must be a conversion step that's been dropped.

aloubyansky:
Thanks, will make sure the line-ending tasks are executed.

bstansberry:
https://issues.jboss.org/browse/WFCORE-3771 will eliminate a source of diff I see between build and build-legacy.

luck3y:
@bstansberry When you say ' not standard CI jobs,' I think these would run as part of the normal WF Core - Full Integration CI builds, which is what I based those CI jobs on. If we want a different config I can switch them over easily enough tonight.

bstansberry:
https://issues.jboss.org/browse/WFCORE-3772 will eliminate another source of diff.

bstansberry:
@luck3y I'm not sure what those jobs are, but any results from them are not part of the current "All checks have passed"  output for this PR. Before they were.  I'm not super concerned either way.

bstansberry:
https://issues.jboss.org/browse/WFCORE-3773 will eliminate another source of diff.

core-galleon-pack/src/main/resources/feature_groups/host-interfaces.xml
```diff
@@ -0,0 +1,15 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<feature-group-spec name="host-interfaces" xmlns="urn:jboss:galleon:feature-group:1.0">
+    <feature spec="host.interface">
+        <param name="interface" value="public"/>
+        <param name="inet-address" value="${jboss.bind.address:127.0.0.1}"/>
+    </feature>
```
bstansberry: I see some differences interface element ordering between the galleon output and legacy. Perhaps this is why.

The traditional order is management, public, private, unsecure.  Not all four are used in all configs of course; I don't think private is used at all in core.

This isn't a huge problem but OTOH if it's as simple as moving this element ^^^ below the 'management' one then that will eliminate a diff to legacy.

core-galleon-pack/src/main/resources/feature_groups/host-profile.xml
```diff
@@ -0,0 +1,23 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<feature-group-spec name="host-profile" xmlns="urn:jboss:galleon:feature-group:1.0">
+    <feature-group name="jmx"/>
```
bstansberry: Putting this after elytron is the traditional order.

core-galleon-pack/src/main/resources/feature_groups/host-interfaces.xml
bstansberry: I tried it and confirmed that moving this done eliminates the diff.

bstansberry:
This dir is left behind and should be removed:

build/target/wildfly-core-5.0.0.Alpha5-SNAPSHOT/standalone/tmp/vfs

luck3y:
OK, so here's what is causing this. We've added VCS triggers to these build, and the full - integration builds are picking up on them as well as the core ones. (I'm guessing you only want core here, eh?)

luck3y:
OK, so easy fix is to just disable the full triggers, would that work short-term?

luck3y:
Full integration triggers disabled for now. Let me know if we should re-enable.

bstansberry:
My only hard requirement is that any other core PR than this one, the CI result reports must show the traditional tests against current wildfly/wildfly:master. If they also show tests against aloubyansky/wildfly:galleon and Alexey wants that, that's ok, although the report should have a different name in results box.

For this one, whatever Alexey wants is fine and there's no urgency.

luck3y:
Oh, I see now, I think. Both of those upstreams were pointed to aloubyansky/wildfly[-core] branches.

core-galleon-pack/src/main/resources/feature_groups/host-interfaces.xml
bstansberry: However, changing core-galleon-pack/src/main/resources/feature_groups/interfaces.xml to put public after management does not result in public coming after management in standalone.xml.

So, I'd rather have consistent interface order between the galleon-produced configs than consistency with the legacy configs in some. If we can get both, great, if not the way you have it now where public comes first is fine.

luck3y:
@aloubyansky @bstansberry How do you want this testing to work? It's currently trying to test -core branch, then -full branch built with that -core. Should it be using upstream -full with this -core?

core-galleon-pack/src/main/resources/feature_groups/access-control.xml
```diff
@@ -0,0 +1,16 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<feature-group-spec name="access-control" xmlns="urn:jboss:galleon:feature-group:1.0">
+    <feature spec="core-service.management">
+        <feature spec="core-service.management.access.authorization">
+            <param name="provider" value="simple"/>
+            <feature spec="core-service.management.access.authorization.role-mapping">
+                <param name="role-mapping" value="SuperUser"/>
+                <feature spec="core-service.management.access.authorization.role-mapping.include">
+                    <param name="include" value="user-\$local"/>
+                    <param name="name" value="&quot;\$local&quot;"/>
```
bstansberry: This escaping produces a different resource address than the current config does, one that I can't navigate to with the CLI. The following works and seems more correct too:
```
                    <param name="include" value="user-$local"/>
                    <param name="name" value="$local"/>
```

bstansberry:
https://issues.jboss.org/browse/WFCORE-3775 will eliminate another diff.

bstansberry:
I pushed up a branch that has three corrections to problems I saw in this. 

https://github.com/wildfly/wildfly-core/compare/master...bstansberry:pullRequest3231+

The first two commits eliminate some license file diffs I saw by syncing files with the legacy variants. The 2nd commit may be a bigger change than was necessary to eliminate the diffs, but my assumptions was the xsl shouldn't be significantly different from the one used by legacy.

Please feel free to cherry-pick and squash in if you agree with those changes.

I also filed #3233 to correct the various problems with the legacy configs that I mentioned above.  That PR would be valid even if galleon didn't even exist.

aloubyansky:
About the xsl used for the license, my idea was to use the same stylesheet for core, servlet and full dists. So I took the one from the full feature-pack and adjusted it slightly.
IMO, license generation should be re-worked completely. I.e. modules (their packages) should indicate what licenses they require. When the tool determines the package set to install, it will collect the actual license requirements and generate the corresponding summary.

core-galleon-pack/src/main/resources/feature_groups/host-interfaces.xml
aloubyansky: For standalone it's because the deps kick in and make the public added first. But if you re-order the groups in standalone group by moving the interfaces before socket-binding-group, will preserve order of the interfaces. I'm gonna correct this to be in sync with the legacy. Thanks.

core-galleon-pack/src/main/resources/feature_groups/host-profile.xml
aloubyansky: ok, thanks!

bstansberry:
@aloubyansky Ok, sounds like the second commit in my branch may not be the way to go. I believe the first one is necessary though; it just brings in a very recent change to the legacy file and without it I see substantive diffs in the output.

Beyond the diffs from the missing 1st commit, there are quite a few others, often just simple stuff like extra spaces between words. I went the simple route and just made the xsl match, but there may be smaller tweaks that will make the diffs go away. I'll investigate.

I'm certain though that we need to get the license-related output very closely aligned as part of getting this merged. Otherwise it will be a hassle.  If the xsl from full, slightly tweaked to correct stuff like extra spaces, can produce the correct output then that means the legacy xsl can be changed to match and another form of diff is gone.

Longer term, rethinking the overall generation process sounds good.

bstansberry:
The last commit on https://github.com/wildfly/wildfly-core/compare/master...bstansberry:pullRequest3231+2 (i.e. https://github.com/wildfly/wildfly-core/commit/4053142f5ab8bcea023579699a2498de7094aab0) eliminates nearly all of the xsl related diff. It does it by bringing some stuff needed for core but not for full into the galleon xsl, and then tweaks the legacy xsl to add the galleon stuff from full. The latter is just to eliminate unnecessary diff in the xsl itself, which is also part of the output.

Note also that in the galleon PR for full you'll want to pick up for galleon this fix that I just submitted for legacy:

https://github.com/wildfly/wildfly/pull/11162

That branch also includes the fixes to legacy for which I filed a separate PR. When I build the whole set the diff looks good. I see:

1) Galleon build has a .galleon dir, which is expected.

2) The CRLF diff in the bat files mentioned previously.

3) Minor diffs in the .xsl files due to intentional differences in how the params are passed. We could make this go away too by tweaking legacy, but lets not do that as part of this. These files are altered a bit as part of "EAP-ization" to create a wildfly-core-eap branch, and when I adapt the set of commits we use to do that to reflect Galleon will be a good time to think about the params. I'll do that once this is merged.

4) The jmx extension element appears in a different spot in the order. This appears to be a flaw in the legacy plugin. Galleon's order is correct, and the input to the legacy plugin is also correct. It's the legacy plugin output that is wrong.

5) Various other config elements are in different order, primarily the location of the discovery subsystem or details in the elytron subsystem. Looking at those it seems that they are natural outcomes from the way the galleon feature[-group]s are organized. That is, there is a group that includes "basic" features and then there are "other" features, with the "other" features applied separately from the basic ones only if they are relevant to a config. This separation results in slight order differences and is just a natural byproduct of galleon's greater flexibility.

aloubyansky:
BTW, I can remove the galleon dir as well. This dir is useful to preserve the state with the goal to perform provisioning operations afterwords and/or for the user to check what is actually installed in galleon terms.

aloubyansky:
@luck3y in terms of the CI for the full integration, I think I would prefer to see the integration with the upstream master instead of my branch, given that I already have the CI for my branches setup. Thanks!

luck3y:
No, problem, I'll switch it over.

luck3y:
@aloubyansky I've set up the full integratiuon job to build with your core and upstream Wildfly.

I think the only issue now is there is no CI runnign your WIldfly branch, only your Wildfly-core branches.

Do you want something set up for the Wildfly branch?

bstansberry:
The docs/contrib/scripts/service/service.bat also needs the CRLF change.

Also, all the galleon .bat files now have an empty line at the top and no newline at the end.

```
diff -r dist/target/wildfly-core-5.0.0.Alpha5-SNAPSHOT/bin/add-user.bat dist-legacy/target/wildfly-core-5.0.0.Alpha5-SNAPSHOT/bin/add-user.bat
1d0
< 
74c73
< if "x%NOPAUSE%" == "x" pause
\ No newline at end of file
```

If that's really easy to make go away, great; otherwise perhaps I can eliminate it in the legacy files or we just ignore it. It's fairly easy to understand in the diff so ignoring it is a valid option. My basic concern about this kind of thing is we'll likely be doing these diffs fairly often until we get rid of the legacy f-ps. So there's value in getting rid of 'diff noise' in that it prevents future 'non-expert' people asking about that noise.  (Once we move to the galleon build and it is canonical, the diffs are still useful as they then can somewhat prove the correctness of the legacy f-ps.)

There's also still a license diff:

```
diff -r dist/target/wildfly-core-5.0.0.Alpha5-SNAPSHOT/docs/licenses/core-feature-pack-licenses.html dist-legacy/target/wildfly-core-5.0.0.Alpha5-SNAPSHOT/docs/licenses/core-feature-pack-licenses.html
36c36
< <td>javax.json</td><td>javax.json-api</td><td>1.1.2</td><td><a href="https://oss.oracle.com/licenses/CDDL+GPL-1.1">Dual license consisting of the CDDL v1.1 and GPL v2</a>
---
> <td>javax.json</td><td>javax.json-api</td><td>1.1.2</td><td><a href="https://javaee.github.io/glassfish/LICENSE">Common Development and Distribution License 1.1</a>
38c38,42
< </td><td><a href="dual license consisting of the cddl v1.1 and gpl v2.txt">dual license consisting of the cddl v1.1 and gpl v2.txt</a>
---
> <a href="http://www.gnu.org/licenses/old-licenses/gpl-2.0-standalone.html">GNU General Public License v2.0 only</a>
> <br>
> </td><td><a href="common development and distribution license 1.1.txt">common development and distribution license 1.1.txt</a>
> <br>
> <a href="gnu general public license v2.0 only.html">gnu general public license v2.0 only.html</a>
diff -r dist/target/wildfly-core-5.0.0.Alpha5-SNAPSHOT/docs/licenses/core-feature-pack-licenses.xml dist-legacy/target/wildfly-core-5.0.0.Alpha5-SNAPSHOT/docs/licenses/core-feature-pack-licenses.xml
46,47c46,52
<           <name>Dual license consisting of the CDDL v1.1 and GPL v2</name>
<           <url>https://oss.oracle.com/licenses/CDDL+GPL-1.1</url>
---
>           <name>Common Development and Distribution License 1.1</name>
>           <url>https://javaee.github.io/glassfish/LICENSE</url>
>           <distribution>repo</distribution>
>         </license>
>         <license>
>           <name>GNU General Public License v2.0 only</name>
>           <url>http://www.gnu.org/licenses/old-licenses/gpl-2.0-standalone.html</url>
```

bstansberry:
The xsl diff is because the 1st commit from my pullRequest3231+ branch didn't get brought over:

https://github.com/bstansberry/wildfly-core/commit/7eff10ae36f250a70515fcaf5ce496e741753610

luck3y:
@aloubyansky Yeah, sorry, I got confused between those jobs and this one running on -core. As to why some different showed up here, I don't know, as AFAIK, nothing has changed in these jobs.

core-galleon-pack/wildfly-feature-pack-build.xml
```diff
@@ -0,0 +1,126 @@
...
+            <prop name="--domain-config" value="domain.xml"/>
+            <prop name="--empty-host-config" value=""/>
+            <prop name="--remove-existing-host-config" value=""/>
+        </props>
+    </config>
+
+    <config name="standalone.xml" model="standalone">
+        <!-- config name is the resulting xml file name which can be changed by setting the property below
+        <props>
+            <prop name="- -server-config" value="standalone-custom.xml"/>
```
bstansberry: Should that space in --server-config be there given this is commented out?

core-galleon-pack/wildfly-feature-pack-build.xml
```diff
@@ -0,0 +1,126 @@
...
+        </props>
+        -->
+        <feature-group name="standalone"/>
+    </config>
+
+    <config name="domain.xml" model="domain">
+        <!-- config name is the resulting xml file name which can be changed by setting the property below
+        <props>
+            <prop name="- -domain-config" value="domain-custom.xml"/>
```
bstansberry: Should that space in --domain-config be there given this is commented out?

core-galleon-pack/wildfly-feature-pack-build.xml
```diff
@@ -0,0 +1,126 @@
...
+        </props>
+        -->
+        <feature-group name="domain"/>
+    </config>
+
+    <config name="host.xml" model="host">
+        <!-- config name is the resulting xml file name which can be changed by setting the property below
+        <props>
+            <prop name="- -host-config" value="host-custom.xml"/>
```
bstansberry: Should that space in --host-config be there given this is commented out?

aloubyansky: Unfortunately, yes, -- is not allowed in the comments.

luck3y:
Hmm, we have a bug open on this, it looks like a pull player thing. 1 second, please. :)

luck3y:
retest this please

bstansberry:
The 7212 "Full integration" build failure with the long trace above looks unrelated; it was a clustering tests not completing:

```
[11:18:38][Step 4/5] [INFO] Running org.jboss.as.test.clustering.cluster.singleton.SingletonPolicyServiceTestCase
[12:12:34][Step 4/5] [INFO] 
[12:12:34][Step 4/5] [INFO] Results:
[12:12:34][Step 4/5] [INFO] 
[12:12:34][Step 4/5] [WARNING] Tests run: 15, Failures: 0, Errors: 0, Skipped: 1
[12:12:34][Step 4/5] [INFO] 
[12:12:34][Step 4/5] [ERROR] There was a timeout or other error in the fork
```

core-galleon-pack/wildfly-feature-pack-build.xml
```diff
@@ -0,0 +1,126 @@
...
+            <prop name="config.branch-is-batch" value="true"/>
+            <prop name="--empty-domain-config" value=""/>
+            <prop name="--remove-existing-domain-config" value=""/>
+            <prop name="--empty-host-config" value=""/>
+            <prop name="--remove-existing-host-config" value=""/>
+        </props>
+        <packages>
+            <package name="product.conf" optional="true"/>
+            <package name="misc.domain"/>
+            <package name="org.jboss.as.domain-management.main"/>
```
bstansberry: This package must be present in any server. Despite its name It is not specific to the 'domain' config model. Just like the JIRA component formally known as 'Domain Management' was not specific to domain mode. :)

core-galleon-pack/wildfly-feature-pack-build.xml
```diff
@@ -0,0 +1,126 @@
...
+        <packages>
+            <package name="product.conf" optional="true"/>
+            <package name="misc.standalone"/>
+            <package name="org.jboss.as.standalone.main"/>
+
+            <!--
+             This one is here to avoid
+             ERROR [org.jboss.as.domain.http.api.undertow] (MSC service thread 1-4) WFLYDMHTTP0004: Unable to load error context for slot main, disabling error context.
+              -->
+            <package name="org.jboss.as.domain-http-error-context.main"/>
```
bstansberry: This has a bit of code smell (as does the same element below in the 'domain' config.) It should probably be a package required by a feature, probably core-service.management.management-interface.http-interface

bstansberry: I guess doing this correctly becomes a requirement of the feature-spec generation, i.e. the ability to associate required packages with a resource such that the feature-spec generation can include those.

core-galleon-pack/wildfly-feature-pack-build.xml
aloubyansky: Oh, sure, added to standalone as well, thanks!

core-galleon-pack/src/main/resources/feature_groups/audit.xml
```diff
@@ -0,0 +1,22 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<feature-group-spec name="audit" xmlns="urn:jboss:galleon:feature-group:1.0">
```
bstansberry: This one perhaps should be name management-audit as there's other audit functionality in the server.

@darranl WDYT?

core-galleon-pack/src/main/resources/feature_groups/core-management.xml
```diff
@@ -0,0 +1,4 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<feature-group-spec name="core-management" xmlns="urn:jboss:galleon:feature-group:1.0">
```
bstansberry: This feature group seems odd. It follows the pattern of a feature group per subsystem, which in other cases includes config details for the feature.  My concern with this one is "core management" is a much bigger concept than just this one subsystem.

General question: what is the benefit of feature groups like this one, discovery.xml and request-controller.xml?

core-galleon-pack/src/main/resources/feature_groups/default-sockets.xml
```diff
@@ -0,0 +1,13 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<feature-group-spec name="default-sockets" xmlns="urn:jboss:galleon:feature-group:1.0">
```
bstansberry: This one doesn't seem right.  These sockets are associated with uses of core-service.management.management-interface.http-interface. Shouldn't these be tied more directly to the uses of those?

I see they don't belong with host-management-interfaces.xml as that's not how host.xml works. They do relate to management-interfaces.xml which is only used for standalone.

If you want these kept in a separate feature-group, e.g. for reuse in other contexts, then I think the name of the group should change. These are management-sockets.

core-galleon-pack/src/main/resources/feature_groups/domain-default-interfaces.xml
```diff
@@ -0,0 +1,8 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<feature-group-spec name="domain-default-interfaces" xmlns="urn:jboss:galleon:feature-group:1.0">
```
bstansberry: I don't see any use of this one.

core-galleon-pack/src/main/resources/feature_groups/domain-profile.xml
```diff
@@ -0,0 +1,19 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<feature-group-spec name="domain-profile" xmlns="urn:jboss:galleon:feature-group:1.0">
```
bstansberry: I think the legacy domain profile is wrong. It's missing the discovery and security-manager subsystems and I don't see any good reason for that.

That probably doesn't change the structure of your feature groups much though.  Your basic-profile seems like a reasonable base, with discovery and security-manager kind of extra. And then elytron seems to have different detail settings between standalone and domain.

core-galleon-pack/src/main/resources/feature_groups/host-server-configs.xml
```diff
@@ -0,0 +1,19 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<feature-group-spec name="host-server-configs" xmlns="urn:jboss:galleon:feature-group:1.0">
```
bstansberry: I don't think this group should exist, or perhaps if it is important to reuse these settings elsewhere it should be called host-example-server-configs. Declaring these three server configs has always just been a kind of example and their specific settings are not meaningful beyond that.

The host.xml file we ship is like that in general.

core-galleon-pack/src/main/resources/feature_groups/audit.xml
aloubyansky: Renaming to management-audit.

core-galleon-pack/src/main/resources/feature_groups/core-management.xml
aloubyansky: The idea was consistency of the building blocks (for subsystems at least) used to compose the configs. But I guess it didn't really help at the end. Ok, I'm gonna remove them and we find it useful, we we'll add the group back later.

core-galleon-pack/src/main/resources/feature_groups/default-sockets.xml
aloubyansky: Renaming.

core-galleon-pack/src/main/resources/feature_groups/domain-default-interfaces.xml
aloubyansky: Thanks, removing.

core-galleon-pack/wildfly-feature-pack-build.xml
```diff
@@ -0,0 +1,126 @@
...
+        <props>
+            <prop name="config.branch-is-batch" value="true"/>
+            <prop name="--admin-only" value=""/>
+            <prop name="--internal-empty-config" value=""/>
+            <prop name="--internal-remove-config" value=""/>
+        </props>
+        <packages>
+            <package name="product.conf" optional="true"/>
+            <package name="misc.standalone"/>
+            <package name="org.jboss.as.standalone.main"/>
```
dmlloyd: We discussed this earlier in chat, but I really believe that the default module slot (main) should not be included in package names.

core-galleon-pack/src/main/resources/feature_groups/host-server-configs.xml
aloubyansky: If you don't mind, I'd still keep it. For the purpose of having a dedicated building block that represents the server configs. I get your point and find it reasonable. But I don't feel good about renaming to an "example" block. Especially with how I see the provisioning groups evolving.

bstansberry: How would a user use the building block? How much manipulation would they need in order to go from the building block to a useful config?

A building block makes sense, but I'm doubtful about whether the content here is a good one.

core-galleon-pack/src/main/resources/feature_groups/domain-profile.xml
bstansberry: https://issues.jboss.org/browse/WFCORE-3784

I'll send up a PR fixing that for legacy and then paste links here to a branch that incorporates that plus a tweak to this file to make it align.

core-galleon-pack/wildfly-feature-pack-build.xml
bstansberry: https://issues.jboss.org/browse/WFCORE-3785

core-galleon-pack/src/main/resources/feature_groups/domain-profile.xml
bstansberry: The last two commits on https://github.com/wildfly/wildfly-core/compare/master...bstansberry:WFCORE-3769_3784 correct that bug and update the galleon domain-profile.xml to match.

#3240 is an independent PR that just fixes legacy.

aloubyansky: The fix for domain-profile.xml has been cherry-picked into this branch. Thanks!

rsvoboda:
Looking at build times of Full integration - Windows job

2h:13m - https://ci.wildfly.org/viewLog.html?buildId=101145&buildTypeId=WildFlyCore_PullRequest_WildFlyCoreFullIntegrationWindows   

1h:52m - https://ci.wildfly.org/viewLog.html?buildId=101153&tab=buildResultsDiv&buildTypeId=WildFlyCore_PullRequest_WildFlyCoreFullIntegrationWindows for PR#3232

Both jobs were running on the same slave, both were running the same number of tests.
Difference is just 1 day between these 2 runs.

@aloubyansky have you checked time difference for build+test run between your fork vs. master ?

aloubyansky:
@rsvoboda i haven't actually.

rsvoboda:
@dosoudil @psakar did you guys look into build times for old vs. new way ?

I will try to return to this tmr, but can't promise :(

luck3y:
retest this please

aloubyansky:
I am ready to open a PR for the full but it will require the release of the core with this PR merged. I would of course appreciate the feedback for it though like for this one. Do you think it makes sense to open the one for full anyway? Thanks.

rsvoboda:
**MASTER**
mvn clean install -fae
 * [INFO] Total time: 12:29 min

mvn clean install -DskipTests -Denforcer.skip=true -Dcheckstyle.skip=true
 * [INFO] Total time: 01:03 min
 * [INFO] Total time: 59.235 s
 * [INFO] Total time: 58.001 s

**GALLEON**
mvn clean install -fae
 * [INFO] Total time: 12:54 min

mvn clean install -DskipTests -Denforcer.skip=true -Dcheckstyle.skip=true
 * [INFO] Total time: 01:14 min
 * [INFO] Total time: 01:13 min
 * [INFO] Total time: 01:11 min

**mvn -version**
```
Apache Maven 3.5.3 (3383c37e1f9e9b3bc3df5050c29c8aff9f295297; 2018-02-24T20:49:05+01:00)
Maven home: /usr/local/Cellar/maven/3.5.3/libexec
Java version: 9.0.4, vendor: Oracle Corporation
Java home: /Library/Java/JavaVirtualMachines/jdk-9.0.4.jdk/Contents/Home
Default locale: en_CZ, platform encoding: UTF-8
OS name: "mac os x", version: "10.13.4", arch: "x86_64", family: "mac"
```

bstansberry:
retest this please

bstansberry:
@luck3y the 'retest this please' above doesn't seem to have done anything.

bstansberry:
Here's the diff between dist/target/wildfly-... and dist-legacy/target/wildfly-... after I rebase this branch onto master and build:

https://paste.fedoraproject.org/paste/lkzo13JAmIpYQPdopvlTfA

Some newline stuff, the different order of extensions subsystem=discovery I've mentioned before (not a problem) and some different order of elements within the elytron system due to galleon using different feature groups for management vs application resources (which is valid.)

So the diff looks good.

Once this is merged I think we should tweak the legacy configs to eliminate the subsystem=discovery elytron details order differences. We want future diffs to show regressions and not hide them amongst 'known ok' diffs, so once we accept galleon output anything in legacy that can be made to align should be made to align.

bstansberry:
@aloubyansky Can you squash this down and put [WFCORE-3769] in the commit message? I don't mind my commits getting squashed.

aloubyansky:
I am going to fix the line endings in the script files, since the diff is introduced by the new plugins for no reason. The configs are written by the server, so the legacy ones will need to be adjusted to match.

bstansberry:
retest this please

bstansberry:
retest this please

luck3y:
retest this please

bstansberry:
With this branch rebased on master plus my commit from #3253 the diff is down to: https://paste.fedoraproject.org/paste/68WrxPqkzGN0SZcTTyX46Q

BTW I'm just reporting status, not pushing to eliminate everything.  I think where it is now is plenty good enough and Alexey has another small tweak in progress.

luck3y:
retest this please

bstansberry:
retest this please

bstansberry:
retest this please

aloubyansky:
Rebased and squashed. Thanks!

PR #3236, opened 2018-05-01
[WFCORE-3762] Actually create the response warning; the previous fix …

bstansberry:
…would never actually do that.

Make the level WARNING not INFO as by default INFO is not sent to the user, making this pointless.

Also, avoid repeated response warnings from different steps for the same end user op.

Second shot at https://issues.jboss.org/browse/WFCORE-3762

### DIFF
--- a/controller/src/main/java/org/jboss/as/controller/AbstractOperationContext.java
+++ b/controller/src/main/java/org/jboss/as/controller/AbstractOperationContext.java
@@ -84,7 +84,6 @@ import org.jboss.as.controller.audit.AuditLogger;
 import org.jboss.as.controller.capability.registry.RuntimeCapabilityRegistry;
 import org.jboss.as.controller.client.MessageSeverity;
 import org.jboss.as.controller.client.OperationResponse;
-import org.jboss.as.controller.descriptions.ModelDescriptionConstants;
 import org.jboss.as.controller.logging.ControllerLogger;
 import org.jboss.as.controller.notification.Notification;
 import org.jboss.as.controller.notification.NotificationSupport;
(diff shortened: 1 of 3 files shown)


controller/src/main/java/org/jboss/as/controller/AbstractOperationContext.java
```diff
@@ -404,6 +402,24 @@ private void createWarning(final Level level, final ModelNode warning){
...
+
+        final ModelNode warnings = getResponseHeaders().get(WARNINGS);
+        boolean unique = true;
+        if (warnings.isDefined()) {
+            // Don't repeat a warning. This is basically a secondard safeguard
+            // against different steps for the same op ending up reporting the
+            // same warning. List iteration is not efficient but > 1 warning
+            // in an op is an edge case
+            for (ModelNode existing : warnings.asList()) {
+                if (existing.equals(warning)) {
```
jmesnil: Is the comparison correct?

* `existing` is an item of `getResponseHeaders().get(WARNINGS)` that contains ModelNode of the `warningEntry` type (with `warning` and `level` properties).
* the `warning` field is a ModelNode that is stored inside the `warningEntry` property.

It seems the call should be:
```
if (existing.get(WARNING).equals(warning)) { // ignore the level warning for the comparison
  ...
}
```

bstansberry: I want to compare the entire item, including the operation and address. That's what makes it a for sure duplicate. The same warn message for different ops would be unlikely but not a dup.

This is definitely just a 2nd line of defense which is why I'm cautious. The main line of defense is https://github.com/wildfly/wildfly-core/pull/3236/files#diff-9258709e9ad2bac62681701ed93df39fR1051 -- that's what prevents steps added by a step that already produced a warning recording another one. That handles the standard case like Stage.MODEL adding Stage.RUNTIME, and also recursive calls like read-feature-description where the added child may have a different address.

I could make the avoiding dups from Stage.MODEL adding Stage.RUNTIME even stronger by just not doing the deprecation check out of Stage.MODEL. All runtime-only ops go through Stage.MODEL anyway, so the check there should suffice.

I'm going to put a Hold label on this as I want to add tests. I originally wrote this along with a PR adding the deprecation stuff to read-feature-description and then that commit had a lot of tests added for that op. But I've had second thoughts about adding this to r-f-d, so now the tests need to be moved.

bstansberry: Sorry; I misunderstood your point. My code was wrong; the test should be

```
if (existing.equals(warningEntry)) 
```

PR #3244, opened 2018-05-02
[WFCORE-3765] Remove org.jboss.msc from the modular system class path

jamezp:
https://issues.jboss.org/browse/WFCORE-3765

Follows up on #3229 to declare remove `org.jboss.msc` from the modular system class path.

### DIFF
--- a/embedded/src/main/java/org/wildfly/core/embedded/EmbeddedProcessFactory.java
+++ b/embedded/src/main/java/org/wildfly/core/embedded/EmbeddedProcessFactory.java
@@ -298,7 +298,7 @@ public class EmbeddedProcessFactory {
             clearPropertyPrivileged(SYSPROP_KEY_CLASS_PATH);
             SecurityActions.setPropertyPrivileged(SYSPROP_KEY_MODULE_PATH, modulePath);
 
-            StringBuilder packages = new StringBuilder("org.jboss.modules,org.jboss.msc,org.jboss.dmr,org.jboss.threads,org.jboss.as.controller.client");
+            StringBuilder packages = new StringBuilder("org.jboss.modules,org.jboss.dmr,org.jboss.threads,org.jboss.as.controller.client");
             if (systemPackages != null) {
                 for (String packageName : systemPackages) {
                     packages.append(",");


jamezp:
@dmlloyd Correct. It's required in for the `ManagedProcessFactory.createXxx()` to work.

jamezp:
@bstansberry Yes that's where it's at. It's trying to load MSC from the modular system class loader.
```
Exception in thread "main" java.lang.IllegalStateException: WFLYEMB0019: Cannot create standalone server using factory: public static org.wildfly.core.embedded.StandaloneServer org.wildfly.core.embedded.EmbeddedStandaloneServerFactory.create(java.io.File,org.jboss.modules.ModuleLoader,java.util.Properties,java.util.Map,java.lang.String[])
	at org.wildfly.core.embedded.EmbeddedProcessFactory.createManagedProcess(EmbeddedProcessFactory.java:301)
	at org.wildfly.core.embedded.EmbeddedProcessFactory.createStandaloneServer(EmbeddedProcessFactory.java:183)
	at org.wildfly.reproducer.Main.main(Main.java:57)
Caused by: java.lang.NoClassDefFoundError: org/jboss/msc/value/Value
	at org.wildfly.core.embedded.EmbeddedStandaloneServerFactory.create(EmbeddedStandaloneServerFactory.java:101)
	at sun.reflect.NativeMethodAccessorImpl.invoke0(Native Method)
	at sun.reflect.NativeMethodAccessorImpl.invoke(NativeMethodAccessorImpl.java:62)
	at sun.reflect.DelegatingMethodAccessorImpl.invoke(DelegatingMethodAccessorImpl.java:43)
	at java.lang.reflect.Method.invoke(Method.java:498)
	at org.wildfly.core.embedded.EmbeddedProcessFactory.createManagedProcess(EmbeddedProcessFactory.java:296)
	... 2 more
Caused by: java.lang.ClassNotFoundException: org.jboss.msc.value.Value
	at java.net.URLClassLoader.findClass(URLClassLoader.java:381)
	at java.lang.ClassLoader.loadClass(ClassLoader.java:424)
	at sun.misc.Launcher$AppClassLoader.loadClass(Launcher.java:349)
	at java.lang.ClassLoader.loadClass(ClassLoader.java:357)
	at org.jboss.modules.JDKSpecific.getSystemClass(JDKSpecific.java:180)
	at org.jboss.modules.ConcurrentClassLoader.performLoadClass(ConcurrentClassLoader.java:395)
	at org.jboss.modules.ConcurrentClassLoader.loadClass(ConcurrentClassLoader.java:116)
	... 8 more
```

bstansberry:
@jamezp It's cruft left over from the pre-WildFly Core embedded, where StandaloneServer exposed MSC:

https://github.com/jbossas/jboss-eap/blob/6.x/embedded/src/main/java/org/jboss/as/embedded/StandaloneServer.java#L64

That got removed along the way.

Unless @dmlloyd can think of a way that the MSC lib used by the appserver also being visible to the embedding app would be useable (e.g. some static methods in MSC that would let the embedding app hack into the appserver ServiceContainer), I think we should just remove org.jboss.msc from the system package list. And even if there is a way to hack like that we probably should cut it off.

jamezp:
@bstansberry I think that makes sense. I'll place a hold label on this until we make a decision the.

dmlloyd:
I'm OK with removing it.

jamezp:
The dependency change has been removed and `org.jboss.msc` package has been removed from the modular system packages.

PR #3248, opened 2018-05-03
[WFCORE-3795]: ReadFeatureDescription doesn't return ObjectAttributeDefinition capabilities

ehsavoie:
Adding required capabilities in features for Object attributes.

Jira: https://issues.jboss.org/browse/WFCORE-3795

### DIFF
--- a/controller/src/main/java/org/jboss/as/controller/operations/global/ReadFeatureDescriptionHandler.java
+++ b/controller/src/main/java/org/jboss/as/controller/operations/global/ReadFeatureDescriptionHandler.java
@@ -394,7 +394,9 @@ public class ReadFeatureDescriptionHandler extends GlobalOperationHandlers.Abstr
             String capabilityName = cap.getName();
             if (cap.isDynamicallyNamed()) {
                 PathAddress aliasAddress = createAliasPathAddress(registration, pa);
-                capabilityName = cap.getDynamicName(aliasAddress);
+                if (aliasAddress.size() > 0) {
+                    capabilityName = cap.getDynamicName(aliasAddress);
+                }
             }
             if (isProfile) {
                 capabilityName = PROFILE_PREFIX + capabilityName;
@@ -408,7 +410,7 @@ public class ReadFeatureDescriptionHandler extends GlobalOperationHandlers.Abstr
                 provide.add(cap);
             }
         }
-        processComplexAttributes(feature, registration);
+        processComplexAttributes(feature, isProfile, registration);
         addReferences(feature, registration);
         addRequiredCapabilities(feature, registration, requestProperties, capabilityScope, isProfile, capabilities,
                 featureParamMappings);
@@ -423,7 +425,7 @@ public class ReadFeatureDescriptionHandler extends GlobalOperationHandlers.Abstr
         return result;
     }
 
-    private void processComplexAttributes(final ModelNode parentFeature, final ImmutableManagementResourceRegistration registration) {
+    private void processComplexAttributes(final ModelNode parentFeature, final boolean isProfile, final ImmutableManagementResourceRegistration registration) {
         for (AttributeAccess attAccess : registration.getAttributes(PathAddress.EMPTY_ADDRESS).values()) {
             if(attAccess.getStorageType() != CONFIGURATION || attAccess.getAccessType() != AttributeAccess.AccessType.READ_WRITE) {
                 continue;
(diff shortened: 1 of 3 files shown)


controller/src/main/java/org/jboss/as/controller/operations/global/ReadFeatureDescriptionHandler.java
```diff
@@ -392,22 +392,24 @@ private ModelNode describeFeature(final Locale locale, final ImmutableManagement
             String capabilityName = cap.getName();
             if (cap.isDynamicallyNamed()) {
                 PathAddress aliasAddress = createAliasPathAddress(registration, pa);
-                capabilityName = cap.getDynamicName(aliasAddress);
+                if (aliasAddress.size() > 0) {
+                    capabilityName = cap.getDynamicName(createAliasPathAddress(registration, pa));
```
aloubyansky: Why not re-use aliasAddress like it used to be?

controller/src/main/java/org/jboss/as/controller/operations/global/ReadFeatureDescriptionHandler.java
```diff
@@ -497,6 +500,31 @@ private void processObjectAttribute(final ModelNode parentFeature, final Immutab
...
+                    capabilityName = baseName + ".$" + paramName;
+                }
+                capability.get(OPTIONAL).set(attrDescription.hasDefined(NILLABLE) && attrDescription.get(NILLABLE).asBoolean());
+                if (isProfile) {
+                    capabilityName = PROFILE_PREFIX + capabilityName;
+                }
+                capability.get(NAME).set(capabilityName);
+                required.add(capability);
+            }
+        }
```
bstansberry: Why doesn't this block do everything that is done in the block starting at https://github.com/bstansberry/wildfly-core/blob/master/controller/src/main/java/org/jboss/as/controller/operations/global/ReadFeatureDescriptionHandler.java#L736?  Except the 'filteredOut' bit at https://github.com/bstansberry/wildfly-core/blob/master/controller/src/main/java/org/jboss/as/controller/operations/global/ReadFeatureDescriptionHandler.java#L750. That 'filteredOut' bit BTW smells funny. AFAICT if filteredOut==true the entire for loop starting at L735 can be skipped as no non-local state is changed in the for loop.

controller/src/main/java/org/jboss/as/controller/operations/global/ReadFeatureDescriptionHandler.java
ehsavoie: Because an empty address creates a NPE in AbstractCapability

aloubyansky: I don't see how it's different.

ehsavoie: line 396 should reuse aliasAddress

jmesnil:
retest this please

controller/src/test/java/org/jboss/as/controller/test/ReadFeatureDescriptionTestCase.java
```diff
@@ -229,6 +230,26 @@ public void testStorageResource() throws OperationFailedException {
...
+        annotation = complexObjectAttr.require(ANNOTATION);
+        Assert.assertEquals("write-attribute", annotation.require(NAME).asString());
+        Assert.assertEquals("complex-object-attr", annotation.require(COMPLEX_ATTRIBUTE).asString());
+        Assert.assertArrayEquals(new String[] {SUBSYSTEM, RESOURCE}, annotation.require(ADDR_PARAMS).asString().split(","));
+        assertSortedArrayEquals(new String[] {"optional-attr", "dynamic-attr"}, annotation.require(OP_PARAMS).asString().split(","));
+        Assert.assertEquals("subsystem.testsubsystem.resource.main-resource", listAttr.require(REFS).asList().get(0).require(FEATURE).asString());
+        // requires
+        ModelNode attrRequires = complexObjectAttr.require(REQUIRES);
+        Assert.assertEquals(2, attrRequires.asList().size());
+        Assert.assertEquals(ROOT_CAPABILITY_NAME, attrRequires.asList().get(0).require(NAME).asString());
```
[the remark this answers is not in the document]
ehsavoie: no ordering of caps isn't significant

PR #3257, opened 2018-05-06
[WFCORE-3804] On the domain server root resource, register OperationD…

bstansberry:
…efinitions for the server lifecycle ops that match what the HC provides on its server proxy resource

Consolidates the various logic related to creating the definitions for the domain server lifecycle ops.

Expose the 'start' method on the /host=x/server=y resource when the server is not started.

https://issues.jboss.org/browse/WFCORE-3804


controller/src/main/java/org/jboss/as/controller/operations/validation/EnumValidator.java
```diff
@@ -136,13 +136,16 @@ public EnumValidator(final Class<E> enumType,  final boolean nullable, final boo
...
+     * @param allowed  the enum values that are allowed. If {@code null} or zero length this is interpreted as meaning all values
      * @param <E>      the type of the enum.
      *
      * @return a new validator.
      */
     @SafeVarargs
     public static <E extends Enum<E>> EnumValidator<E> create(final Class<E> enumType, final E... allowed) {
+        if (allowed == null || allowed.length == 0) {
+            return create(enumType, EnumSet.allOf(enumType));
+        }
```
bstansberry: Note that this is a behavior change in this method, as before providing a null or empty array would result in a validator that would reject all values.  That is non-sensical behavior and I checked wf-core and wf code for any uses that passed null or empty and found none, so I made this change.

host-controller/src/main/java/org/jboss/as/host/controller/resources/ServerConfigResourceDefinition.java
```diff
@@ -219,20 +231,34 @@ public void registerCapabilities(ManagementResourceRegistration resourceRegistra
...
-        resourceRegistration.registerOperationHandler(ServerReloadHandler.DEFINITION, reloadHandler);
-        resourceRegistration.registerOperationHandler(ServerSuspendHandler.DEFINITION, new ServerSuspendHandler(serverInventory));
-        resourceRegistration.registerOperationHandler(ServerResumeHandler.DEFINITION, new ServerResumeHandler(serverInventory));
-        ServerProcessHandlers.ServerDestroyHandler destroyHandler = new ServerProcessHandlers.ServerDestroyHandler(serverInventory);
-        resourceRegistration.registerOperationHandler(ServerProcessHandlers.DESTROY_OPERATION, destroyHandler);
-        ServerProcessHandlers.ServerKillHandler killHandler = new ServerProcessHandlers.ServerKillHandler(serverInventory);
-        resourceRegistration.registerOperationHandler(ServerProcessHandlers.KILL_OPERATION, killHandler);
+        // Lifecycle ops are deprecated on the server-config resource but not on server
+        boolean serverConfig = resourceRegistration.getPathAddress().getLastElement().getKey().equals(SERVER_CONFIG);
+        ModelVersion deprecatedVersion = serverConfig ? ModelVersion.create(9) : null;
```
bstansberry: Note: if this isn't merged for WF 15, the # will need to be increased.

bstansberry:
@yersan Does this look ok to you?

bstansberry:
retest this please

host-controller/src/main/java/org/jboss/as/host/controller/resources/ServerConfigResourceDefinition.java
yersan: This is really a tiny point, I'm not sure if it should be considerated an issue for this PR, but due to this change together with https://github.com/wildfly/wildfly-core/pull/3257/files#diff-d5d53a445abb957562382b7f5df50d9eL51, https://github.com/wildfly/wildfly-core/pull/3257/files#diff-992bfe231fcd944a7d05053a896995eeL91 and https://github.com/wildfly/wildfly-core/pull/3257/files#diff-f2492d9ce98c7f8c0b2c524dd19c61f1L103, we are changing the deprecated since version of `reload`, `suspend` and `resume` operations under server-config from 7.0.0 to 9.0.0, which is not striclty true. I cannot see any big consequence if we are not going to remove those operations soon though, so I let you to your criteria. The other changes LGTM.

bstansberry: @yersan Thanks; I'll clean that up.

bstansberry: @yersan Fixed.

PR #3258, opened 2018-05-07
[WFCORE-3806] Eliminating dead code.

ropalka:
https://issues.jboss.org/browse/WFCORE-3806

Prior MSC 1.2.8 stability monitors were reporting problems if
some transitive dependency wasn't available. This has changed
since MSC 1.2.8 and above. There is no notion of transitive
dependencies in JBoss MSC anymore. If stability monitor detects
a problem it is always immediate dependency of monitored controller.


bstansberry:
Is the current code broken? If not I would prefer to defer this to right at the beginning of WF 14. If this introduces subtle problems I'd rather have a full cycle to detect and fix them vs having blockers popping up at the end.

ropalka:
It isn't broken @bstansberry - it's just dead code. Feel free to postpone it to WF 14.

ropalka:
I guess WFCORE version 5.0.0.Beta1-SNAPSHOT is targeting WF 14, right @bstansberry ?

bstansberry:
@ropalka No, WF Core 5 is for WF 13.

PR #3259, opened 2018-05-07
[WFCORE-323] Allow multistep op handling to defer resolution of the O…

bstansberry:
…perationEntry until it's time to execute the step

https://issues.jboss.org/browse/WFCORE-323

I bumped the priority on this as it can be useful to tools like galleon by removing the need to segregate extension add ops associated with a feature from the related subsystem ops.

The basic point here is to allow an extension=foo:add and a subsystem=foo:add op to both be in the same composite, without the latter failing because initially the MRR for the subsystem doesn't exist.

The way this works is via the util that registers op steps for composites tracking. If a step has added or removed an extension, for subsequent subsystem ops instead of the util requiring resolution of the handler for the op and using that for the step, it can use a handler that when it executes will try at that point to resolve the handler. This means the resolution is deferred until after the extension add/remove has executed.

### DIFF
--- a/host-controller/src/main/java/org/jboss/as/domain/controller/operations/coordination/OperationRouting.java
+++ b/host-controller/src/main/java/org/jboss/as/domain/controller/operations/coordination/OperationRouting.java
@@ -29,6 +29,9 @@ import static org.jboss.as.controller.descriptions.ModelDescriptionConstants.OP_
 import static org.jboss.as.controller.descriptions.ModelDescriptionConstants.PROFILE;
 import static org.jboss.as.controller.descriptions.ModelDescriptionConstants.SERVER;
 import static org.jboss.as.controller.descriptions.ModelDescriptionConstants.STEPS;
+import static org.jboss.as.controller.descriptions.ModelDescriptionConstants.SUBSYSTEM;
+import static org.jboss.as.controller.operations.global.GlobalOperationHandlers.STD_READ_OPS;
+import static org.jboss.as.controller.operations.global.GlobalOperationHandlers.STD_WRITE_OPS;
 
 import java.util.Collections;
 import java.util.HashSet;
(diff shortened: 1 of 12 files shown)


bstansberry:
@jmesnil  This should be good to go now. If this is WF 17, then Galleon can use it in WF 18.

host-controller/src/main/java/org/jboss/as/domain/controller/operations/coordination/OperationRouting.java
```diff
@@ -97,6 +123,23 @@ private static OperationRouting determineRouting(final ModelNode operation, fina
         return result;
     }
 
+    private static PathAddress findSubsystemRootAddress(PathAddress address) {
+        PathAddress result = null;
+        int size = address.size();
+        if (size > 1) {
+            int subsystemKey = Integer.MAX_VALUE;
+            String firstKey = address.getElement(0).getKey();
+            if (HOST.equals(firstKey) || PROFILE.equals(firstKey)) {
```
jmesnil: what about a subsystem operation beneath a DEPLOYMENT?

bstansberry: @jmesnil This logic deals with the steps provided by the user, so if your concern is a step added by another step, then this logic doesn't apply to those.

Also, this logic executes on an HC, and /deployment=foo/subsystem=bar registration would  be on the server. So the user-provided address would be /host=x/server=y/deployment=foo/subsystem=bar. And that would result in a 'true' at https://github.com/wildfly/wildfly-core/blob/master/host-controller/src/main/java/org/jboss/as/domain/controller/operations/coordination/OperationRouting.java#L87 so long as the HC knows about host 'x' (if it's the DC) and server 'y' (if it's host 'x').

If L87 was false then we have an unknown remote process and we want to fail. With my change here it will drop into the 'else if (compositeStep)' block. The logic there won't match  /host=x/server=y/deployment=foo/subsystem=bar which is what we want.

A flaw I realized while writing this is it will match /host=invalid_typo/subsystem=jmx. The request will still fail later anyway so I don't think that's a big deal. But if it's easy enough to correct I'll correct it.

PR #3262, opened 2018-05-08
[WFCORE-2589] Eliminating last remaining usages of MSC optional dependencies in WF-CORE.

ropalka:


host-controller/src/main/java/org/jboss/as/host/controller/DomainModelControllerService.java
```diff
@@ -833,10 +832,12 @@ public void execute(OperationContext context, ModelNode operation) throws Operat
                         .install();
 
                 // demand http mgmt services
-                serviceTarget.addService(ServiceName.JBOSS.append("http-mgmt-startup"), Service.NULL)
-                        .addDependency(ServiceBuilder.DependencyType.OPTIONAL, UndertowHttpManagementService.SERVICE_NAME)
-                        .setInitialMode(ServiceController.Mode.ACTIVE)
-                        .install();
+                if (capabilityRegistry.hasCapability(UndertowHttpManagementService.EXTENSIBLE_HTTP_MANAGEMENT_CAPABILITY.getName(), CapabilityScope.GLOBAL)) {
```
ropalka: Is it proper usage of Capabilities API @bstansberry ? I'm not familiar with capabilities API so I can't say if it is correct.

ropalka:
Thank you @darranl for reviewing security bits of this PR.

host-controller/src/main/java/org/jboss/as/host/controller/DomainModelControllerService.java
bstansberry: Yes, this is ok.

PR #3263, opened 2018-05-08
[WFCORE-3809] Refactoring - use Delegating MSC classes everywhere.

ropalka:
https://issues.jboss.org/browse/WFCORE-3809

### DIFF
--- a/server/src/main/java/org/jboss/as/server/deployment/service/SecuredServiceRegistry.java
+++ b/server/src/main/java/org/jboss/as/server/deployment/service/SecuredServiceRegistry.java
@@ -1,54 +1,35 @@
 package org.jboss.as.server.deployment.service;
 
 import org.jboss.as.server.security.ServerPermission;
-import org.jboss.msc.service.ServiceController;
-import org.jboss.msc.service.ServiceName;
-import org.jboss.msc.service.ServiceNotFoundException;
+import org.jboss.msc.service.DelegatingServiceRegistry;
 import org.jboss.msc.service.ServiceRegistry;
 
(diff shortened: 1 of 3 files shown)


server/src/main/java/org/jboss/as/server/deployment/service/SecuredServiceRegistry.java
```diff
@@ -1,54 +1,35 @@
...
+public class SecuredServiceRegistry extends DelegatingServiceRegistry {
 
     /**
-     * A {@link org.jboss.as.server.security.ServerPermission} needed to use {@link org.jboss.as.server.deployment.service.SecuredServiceRegistry}, i.e. invoke its methods. The name of the permission is "{@code useServiceRegistry}."
+     * A {@link org.jboss.as.server.security.ServerPermission} needed to use
+     * {@link org.jboss.as.server.deployment.service.SecuredServiceRegistry}, i.e. invoke its methods.
+     * The name of the permission is "{@code useServiceRegistry}."
      */
-    public static final Permission PERMISSION = ServerPermission.USE_SERVICE_REGISTRY;
+    private static final Permission PERMISSION = ServerPermission.USE_SERVICE_REGISTRY;
```
bstansberry: Making this field private is fine, except it will suppress the javadoc about the use of the permission. Probably no one will read that, but still I think the class javadoc should describe the required permission. The class javadoc is the better place to describe this anyway; the only reason the class exists is to do this perm check.

ropalka:
Fixed @bstansberry , please review.

PR #3265, opened 2018-05-08
[WFCORE-3810] Use the right name for the galleon pack zip

bstansberry:
https://issues.jboss.org/browse/WFCORE-3810


aloubyansky:
Approved, thanks!

PR #3266, opened 2018-05-08
[WFCORE-3811] Upgrade to WildFly Galleon Plugins 1.0.0.Alpha20

aloubyansky:
This change eliminates the need for the commit in WFCORE-3808.
Also removes properties from the configs that are matching the defaults, so not effective.


bstansberry:
I confirmed that this fixes the WFCORE-3808 problem.

PR #3269, opened 2018-05-09
[WFCORE-3812] Eliminating usages of deprecated SC.getSubstate() method.

ropalka:
https://issues.jboss.org/browse/WFCORE-3812

### DIFF
--- a/server/src/main/java/org/jboss/as/server/deployment/DeploymentStatusHandler.java
+++ b/server/src/main/java/org/jboss/as/server/deployment/DeploymentStatusHandler.java
@@ -47,15 +47,14 @@ public class DeploymentStatusHandler implements OperationStepHandler {
         final String runtimeName = RUNTIME_NAME.resolveModelAttribute(context, deployment).asString();
         context.addStep(new OperationStepHandler() {
             @Override
-            public void execute(final OperationContext context, final ModelNode operation) throws OperationFailedException {
+            public void execute(final OperationContext context, final ModelNode operation) {
                 final ModelNode result = context.getResult();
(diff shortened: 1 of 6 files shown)


server/src/main/java/org/jboss/as/server/deployment/DeploymentStatusHandler.java
```diff
@@ -47,15 +47,14 @@ public void execute(OperationContext context, ModelNode operation) throws Operat
...
+            public void execute(final OperationContext context, final ModelNode operation) {
                 final ModelNode result = context.getResult();
                 if (!isEnabled) {
                     result.set(AbstractDeploymentUnitService.DeploymentStatus.STOPPED.toString());
                 } else {
                     final ServiceController<?> controller = context.getServiceRegistry(false).getService(Services.deploymentUnitName(runtimeName));
                     if (controller != null) {
-                        if (controller.getSubstate() == ServiceController.Substate.WONT_START &&
-                                controller.getState() == ServiceController.State.DOWN) {
+                        if (controller.getState() == ServiceController.State.DOWN) {
```
bstansberry: I don't think this is equivalent.  There are other variants of DOWN than WONT_START and if those are true the existing code drops into a call to AbstractDeploymentUnitService.getStatus. And that checks a StabilityMonitor for both failed and problems and if either are found the result is DeploymentStatus.FAILED, not DeploymentStatus.STOPPED.  In other words, DOWN due to missing dependencies -> DeploymentStatus.FAILED.

ropalka:
Thanks for spoting it out @bstansberry - you are right of course - fixed.

PR #3271, opened 2018-05-09
[WFCORE-3814] Upgrade to Galleon and plugins 1.0.0.CR1

aloubyansky:
https://issues.jboss.org/browse/WFCORE-3814

### DIFF
--- a/core-galleon-pack/src/main/resources/packages/org.wildfly.openssl/pm/wildfly/tasks.xml
+++ b/core-galleon-pack/src/main/resources/packages/org.wildfly.openssl/pm/wildfly/tasks.xml
@@ -1,13 +1,13 @@
 <?xml version="1.0" ?>
 
 <tasks xmlns="urn:wildfly:wildfly-feature-pack-tasks:2.0">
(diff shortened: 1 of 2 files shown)


psakar:
@bstansberry  @aloubyansky both issues (A hang when using maven proxy B failure because of artifact not found without proxy) are present even after applying this patch

aloubyansky:
@psakar i'll verify again today.

psakar:
@aloubyansky hint - If I remove scope test for artifact which is not found (eg. https://github.com/wildfly/wildfly-core/blob/5.0.0.Alpha7/logging/pom.xml#L195), it is found

aloubyansky:
retest this please

bstansberry:
I'm not sure from the discussion above whether the proxy issue is corrected or not, but in any case this passes and the non-component-upgrade part is fine, so I'll merge it. If there's further change needed that can come in the next upgrade.

bstansberry:
The GAL or WFGP projects in JIRA can be used for issues in Galleon or the WildFly plugins to it.

PR #3273, opened 2018-05-10
[WFCORE-3816] ExtensionIndexService throws NPE while reading extensions on the IBM jdk

pkremens:
https://issues.jboss.org/browse/WFCORE-3816 ExtensionIndexService throws NPE while reading extensions on the IBM jdk   
   
IBM jdk contains 
```
ibm-java-8.0-5.11/jre/lib/ext/traceformat.jar: java.lang.NullPointerException
ibm-java-8.0-5.11/jre/lib/ext/dtfj.jar: java.lang.NullPointerException
ibm-java-8.0-5.11/jre/lib/ext/dtfjview.jar: java.lang.NullPointerException
``` 
where `manifest.getMainAttributes();` throws NPE.
   
causes ->
https://issues.jboss.org/browse/WFLY-10349 WeldBundledLibraryDeployment Test cases fail on IBM jdk

### DIFF
--- a/server/src/main/java/org/jboss/as/server/moduleservice/ExtensionIndexService.java
+++ b/server/src/main/java/org/jboss/as/server/moduleservice/ExtensionIndexService.java
@@ -91,6 +91,10 @@ public final class ExtensionIndexService implements Service<ExtensionIndex>, Ext
                         final JarFile jarFile = new JarFile(jar);
                         try {
                             final Manifest manifest = jarFile.getManifest();
+                            if (manifest == null) {
+                                // jar lacks manifest file
+                                continue;
+                            }
                             final Attributes mainAttributes = manifest.getMainAttributes();
                             final String extensionName = mainAttributes.getValue(Attributes.Name.EXTENSION_NAME);
                             if (extensionName == null) {


stuartwdouglas:
What does the stack trace look like? In general you should only catch NPE as a last resort.

manovotn:
The problem here is that some JARs from IBM JDK lack manifest files (which is bad practice but seems valid).
Code here [presumes](https://github.com/wildfly/wildfly-core/pull/3273/files#diff-0b4f3990c55380e1e17394af376ee553R93) that there is always one (no null check on `final Manifest manifest = jarFile.getManifest();`).

Stacktrace can be seen in [WFLY-10349](https://issues.jboss.org/browse/WFLY-10349).

pkremens:
```
$ for JAR in `find /usr/lib/java/ibm-java-8.0-5.11/jre/lib/ext/ -name "*.jar" ` ; do unzip -l $JAR | grep MANIFEST &> /dev/null || echo $JAR ; done
/usr/lib/java/ibm-java-8.0-5.11/jre/lib/ext/traceformat.jar
/usr/lib/java/ibm-java-8.0-5.11/jre/lib/ext/dtfj.jar
/usr/lib/java/ibm-java-8.0-5.11/jre/lib/ext/dtfjview.jar
```
The code now check for null instead.

PR #3281, opened 2018-05-11
[WFCORE-3722] delete duplicate version properties.

soul2zimate:
https://issues.jboss.org/browse/WFCORE-3722 delete duplicate version properties.


bstansberry:
@soul2zimate Can you expand on this a bit and get the rest on the plugin version props in that section moved to be properly grouped in alphabetically with all the rest? Perhaps wait a day or two in case people disagree.

@dosoudil @psakar @kabir FYI, in case you disagree with organizing them all together. I see some vague benefit in separating out plugins from normal dependencies, but OTOH experience is teaching me that any organizing principle other than the simplest possible is just going to fall apart over time, as it has here.

jamezp:
I agree that plugin versions should be in a separate section. There is already a commented with `<!-- Non-default maven plugin versions and configuration -->`.

bstansberry:
@jamezp You said you agree, but I asked Chao Wang to move those version props out of that section. :)

jamezp:
@bstansberry Whoops I read that wrong for sure. I don't have a strong opinion either way as long as it's consistent.

bstansberry:
I think I'll go ahead and merge this as there's no reason to leave the duplicates while we discuss organization.  By doing that I don't mean to imply I'm not interested in any input.

I know Kabir is looking at generating the former component-matrix/pom ,the content of which probably (???) is not meant to include plugins. It was very vague wondering about ease of maintenance of that kind of thing that led to my question.

kabir:
@bstansberry Correct, at least the component-matrix we removed did not really contain plugins.

bstansberry:
Gosh, I totally misread this PR anyway. It fixed the duplication by removing the plugins from the alphabetical list.

PR #3294, opened 2018-05-13
[WFCORE-3842] Simplify generation of core-feature-pack-licenses.xml

bstansberry:
https://issues.jboss.org/browse/WFCORE-3842

Also cleans up a whitespace bug in the root pom.

### DIFF
--- a/core-feature-pack/pom.xml
+++ b/core-feature-pack/pom.xml
@@ -518,7 +518,7 @@
             <plugin>
                 <groupId>org.apache.maven.plugins</groupId>
                 <artifactId>maven-resources-plugin</artifactId>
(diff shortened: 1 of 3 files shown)


aloubyansky:
I see the change as an improvement (in a sense that it's cleaner) even if it's not to workaround the order of execution.
I am not sure though whether this issue will be fixed w/o taking the same approach in core-galleon-pack.

bstansberry:
I updated it to make the same change in core-galleon-pack. For whatever reason it didn't have the problem which is why I didn't change it initially. But it's cleaner and more robust to have it work the same as it does in core-galleon-pack..

PR #3300, opened 2018-05-14
[WFCORE-3842] Use an actually existing input file for insert-versions…

bstansberry:
… in core-galleon-pack

https://issues.jboss.org/browse/WFCORE-3842

My initial fix I screwed up the core-galleon-pack.


aloubyansky:
Verified, it fixes the issue.

PR #3309, opened 2018-05-16
BLOCKER - [WFCORE-3859] Add transformer to transform from version  2.0.0 to 1.2.0 of the Elytron subsystem model.

darranl:
https://issues.jboss.org/browse/WFCORE-3859

### DIFF
--- a/elytron/src/test/resources/org/wildfly/extension/elytron/elytron-transformers-1.2-reject.xml
+++ b/elytron/src/test/resources/org/wildfly/extension/elytron/elytron-transformers-1.2-reject.xml
@@ -1,45 +1,5 @@
 <subsystem xmlns="urn:wildfly:elytron:3.0">
-    <authentication-client>
-        <authentication-configuration name="base" anonymous="true" />
-        <authentication-configuration name="forward-authz"
-                                      extends="base"
-                                      security-domain="TestDomain"
-                                      forwarding-mode="authorization">
-            <credential-reference clear-text="supersecret!"/>
-        </authentication-configuration>
-    </authentication-client>
-    <security-domains>
-        <security-domain name="TestDomain" default-realm="identity">
-            <realm name="identity" />
-        </security-domain>
-    </security-domains>
-    <security-realms>
-        <identity-realm name="identity" identity="identity" />
-    </security-realms>
-    <mappers>
-        <simple-permission-mapper name="SimplePermissionMapperLegacy" mapping-mode="and">
-            <permission-mapping match-all="true">
-                <permission class-name="a.b.MyPermission" />
-            </permission-mapping>
-        </simple-permission-mapper>
-        <simple-permission-mapper name="SimplePermissionMapper" mapping-mode="and">
-            <permission-mapping match-all="true">
-                <permission-set name="my-permission" />
-            </permission-mapping>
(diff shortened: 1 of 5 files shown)


jmesnil:
Is there a transformer test covering this change?
I didn't see a elytron-transformers-2.0.xml or elytron-transformers-2.0-reject.xml in the test resources

darranl:
Just looking to clean something up in the testsuite - we have an elytron-transformers-1.2.xml and the reject pair as that is what we are testing we are transforming to - but there seem to some mixed code in the test case that doesn't actually test anything that I am looking into.

darranl:
Retest this please

elytron/src/test/resources/org/wildfly/extension/elytron/elytron-transformers-1.2-reject.xml
```diff
@@ -1,45 +1,5 @@
 <subsystem xmlns="urn:wildfly:elytron:3.0">
```
jmesnil: shouldn't the XML namespace be `urn:wildfly:elytron:1.2`?

elytron/src/test/resources/org/wildfly/extension/elytron/elytron-transformers-1.2.xml
```diff
@@ -16,6 +16,9 @@
     <security-realms>
         <identity-realm name="identity" identity="identity" />
     </security-realms>
+    <credential-security-factories>
```
jmesnil: same remark as above on the xml namespace

elytron/src/test/resources/org/wildfly/extension/elytron/elytron-transformers-1.2-reject.xml
darranl: No as I already said above - the version in this name is the version we are transforming to.

elytron/src/test/resources/org/wildfly/extension/elytron/elytron-transformers-1.2.xml
darranl: No as I already said above - the version in this name is the version we are transforming to.

darranl:
The naming convention was set in this PR https://github.com/wildfly/wildfly-core/pull/3002 and approved by @bstansberry - at the time the elytron-transformers-1.2.xml was using namespace 2.0.

I would prefer we test transformations to the versions of WildFly Core we release but a decision was made to test only to the versions used in EAP.

bstansberry:
I don't care much about file names, but transformers are testing transformation to API versions not xsd versions, so if a file is used for such a test naming it after the target API version makes sense to me.

Re; "test only to the versions used in EAP", I think that's about API versions 1.0 and 1.1?  Those were in Core 3.0.0 and 3.0.2 respectively, with 3.0.6 introducing version 1.2.  Core 3.0.8 was what was used in WildFly 11.0.0.Final.  So not testing 1.0 and 1.1 doesn't mean we only care about EAP, it just means we are ignoring core versions that were never used in a WildFly Final.

In a perfect world we'd test transforming to 1.0 and 1.1 too. I wrote the tests that #3002 removed! And they passed!  But IIRC there was some hassle related to them or something, or maybe I just got tired of fighting over it.

PR #3311, opened 2018-05-17
[WFLY-10394] Workaround wrong TCCL issue in naming subsystem by restoring lost (otherwise unused) dependency

dmlloyd:
https://issues.jboss.org/browse/WFLY-10394


dmlloyd:
This does not resolve WFLY-10394 (which is why I didn't move it to WFCORE) but it does work around it for now.

dmlloyd:
Failure is unrelated (disk full)

PR #3316, opened 2018-05-18
[WFCORE-3864] Switch to using WildFly Common byte and code point iterators

fjuma:
https://issues.jboss.org/browse/WFCORE-3864

### DIFF
--- a/core-feature-pack/src/main/resources/modules/system/layers/base/org/jboss/as/cli/main/module.xml
+++ b/core-feature-pack/src/main/resources/modules/system/layers/base/org/jboss/as/cli/main/module.xml
@@ -74,6 +74,7 @@
         <module name="org.jboss.threads"/>
         <module name="org.jboss.vfs"/>
         <module name="org.picketbox" optional="true"/>
+        <module name="org.wildfly.common"/>
         <module name="org.wildfly.embedded"/>
     </dependencies>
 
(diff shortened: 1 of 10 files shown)


core-feature-pack/src/main/resources/modules/system/layers/base/org/jboss/as/cli/main/module.xml
```diff
@@ -74,6 +74,7 @@
         <module name="org.jboss.threads"/>
         <module name="org.jboss.vfs"/>
         <module name="org.picketbox" optional="true"/>
+        <module name="org.wildfly.common"/>
```
bstansberry: Can you add this in thecli/pom.xml as well please?  I don't see any previous compile-time uses of this in the cli module codebase itself.  The org.wildfly.common stuff ends up in jboss-cli-client.jar anyway, probably because of a transitive dependency via controller-client. But I think it's better if we're explicit in the poms and not rely on transitive deps.

fjuma: Done.

nekdozjam:
@fjuma I think you missed one at org.wildfly.extension.elytron.AdvancedModifiableKeyStoreDecorator line 601, there is still security ByteStringBuilder instead of common one

PR #3317, opened 2018-05-18
[WFCORE-3862] Ensure the modulePath set by the user is used if set.

jamezp:
https://issues.jboss.org/browse/WFCORE-3862


jamezp:
The only change was to edit made was to edit the commit message as there was a spelling mistake.

PR #3318, opened 2018-05-21
[WFCORE-3866] CLI does not need an MSC dependency to function as the …

bstansberry:
…embedding app for an embedded server

https://issues.jboss.org/browse/WFCORE-3866

### DIFF
--- a/cli/pom.xml
+++ b/cli/pom.xml
@@ -148,11 +148,6 @@
                 </exclusion>
             </exclusions>
         </dependency>
-        <!-- added here as it is dep of embedded and needs to be included in shaded jar-->
-        <dependency>
-            <groupId>org.jboss.msc</groupId>
-            <artifactId>jboss-msc</artifactId>
-        </dependency>
         <dependency>
             <groupId>org.wildfly.security</groupId>
             <artifactId>wildfly-elytron</artifactId>
--- a/cli/src/main/java/org/jboss/as/cli/embedded/EmbeddedControllerHandlerRegistrar.java
+++ b/cli/src/main/java/org/jboss/as/cli/embedded/EmbeddedControllerHandlerRegistrar.java
@@ -51,7 +51,7 @@ public class EmbeddedControllerHandlerRegistrar {
             // nothing, as the prop has already been read. Does no harm either.
             //
             // Why these packages?
-            // 1) msc, dmr, controller.client are part of the StandaloneServer API
+            // 1) dmr, controller.client are part of the StandaloneServer API
             // 2) org.jboss.threads is unfortunately part of the ModelControllerClient API
             // 3) org.jboss.modules -- dunno, seems vaguely logical; the AS7 embedded code included this so I do too
             // 4) org.jboss.logging and org.jboss.logmanager -- logmanager.LogManager is installed as
@@ -59,7 +59,7 @@ public class EmbeddedControllerHandlerRegistrar {
             //    heavily. If the CLI-side LogContext class and the modular side LogContext class are different,
             //    nothing works properly. Since logmanager is a system pkg, the org.jboss.logging API stuff must be too
             WildFlySecurityManager.setPropertyPrivileged("jboss.modules.system.pkgs",
-                    "org.jboss.modules,org.jboss.msc,org.jboss.dmr,org.jboss.threads," +
+                    "org.jboss.modules,org.jboss.dmr,org.jboss.threads," +
                     "org.jboss.as.controller.client,org.jboss.logging,org.jboss.logmanager");
 
             String classname = "org.jboss.modules.Module";


bstansberry:
retest this please

bstansberry:
This fails:

```
$ java -jar jboss-cli-client.jar 
WARN: can't find jboss-cli.xml. Using default configuration values.
You are disconnected at the moment. Type 'connect' to connect to the server or 'help' for the list of supported commands.
[disconnected /] embed-server --jboss-home=~/dev/wildfly/wildfly/build/target/wildfly-14.0.0.Beta1-SNAPSHOT
Cannot start embedded server: WFLYEMB0019: Cannot create standalone server using factory: public static org.wildfly.core.embedded.StandaloneServer org.wildfly.core.embedded.EmbeddedStandaloneServerFactory.create(java.io.File,org.jboss.modules.ModuleLoader,java.util.Properties,java.util.Map,java.lang.String[]): org/jboss/msc/value/Value: org.jboss.msc.value.Value
[disconnected /] 
```

bstansberry:
The 2nd commit here fixes the failure:

```
$ java -jar build/target/wildfly-14.0.0.Beta1-SNAPSHOT/bin/client/jboss-cli-client.jar 
WARN: can't find jboss-cli.xml. Using default configuration values.
You are disconnected at the moment. Type 'connect' to connect to the server or 'help' for the list of supported commands.
[disconnected /] embed-server --jboss-home=~/dev/wildfly/wildfly/build/target/wildfly-14.0.0.Beta1-SNAPSHOT
Warning! The CLI is running in a non-modular environment and cannot load commands from management extensions.
[standalone@embedded /] ls
core-service                               subsystem                                  namespaces=[]                              release-version=6.0.0.Alpha2-SNAPSHOT      
deployment                                 system-property                            organization=undefined                     running-mode=ADMIN_ONLY                    
deployment-overlay                         launch-type=EMBEDDED                       process-type=Server                        runtime-configuration-state=ok             
extension                                  management-major-version=8                 product-name=WildFly Full                  schema-locations=[]                        
interface                                  management-micro-version=0                 product-version=14.0.0.Beta1-SNAPSHOT      server-state=running                       
path                                       management-minor-version=0                 profile-name=undefined                     suspend-state=RUNNING                      
socket-binding-group                       name=taozi                                 release-codename=                          uuid=65444a8c-f3da-4fb3-bcbb-8d1786e1093d  
```

It can be squashed; I'm leaving it separate for now to help highlight what I mention below.

That commit is fine for solving this particular issue. @jfdenise @jamezp -- fyi: that logic in the CLI code itself isn't the ideal way to be handling things. The comments before that block explains why it's done that way though; the system property must be set before the Module class is loaded. If someone has a better idea that would be great.

It seems I/we never wrote any tests of embedded using jboss-cli-client.jar, otherwise the failure in my previous comment would have been detected.  I kind of understand why; such a test would be a hassle since the CLI would be a separate process and a problem with the embedded server would not cause a failed exit code from the process. I suppose a test could manipulate stdio or something.

jfdenise:
@bstansberry , I am wandering if this logic is actually needed. The CLI doesn't seem designed to run without loadable JBoss Modules (from modular or not modular).  For example ExtensionsLoader has dependency on ModuleClassLoader and is instantiated from CommandContextImpl that is a key class of CLI.
So if JBOSS Module is all the time there, do we need hasModules? If we don't need this check, we can then check that 'cl' classloader (or one of its parents) is instanceof ModuleClassLoader. If that is the case, then we are in Modular context, otherwise modular = false.
So we could avoid to load Module class in a non modular context.
If I am missing something and we must check that JBOSS Module is loadable, we could share the set of required system packages with embedded Configuration that could expose it in order to avoid duplication of this knowledge. BTW, I can see that we have more system packages in CLI than in the embedded configuration. If they are required and we don't load anymore the Module class in non modular env, then we would need to provide them when creating the embedded Configuration.

jfdenise:
@bstansberry , adding a note, if we want to keep the hasModules, we could call Class clazz = Class.forName("org.jboss.modules.Module", false, EmbeddedControllerHandlerRegistrar.class.getClassLoader()); in order to not call static initializers of Module. This should work.

bstansberry:
@jamezp  Unless @jfdenise objects I think this is good to go, and the rest of this discussion is kind of a very good tangent any action on which would be separate work.

@jfdenise Thanks for the research!

So AIUI, CLI has a compile time dep on JBoss Modules in order to support embedded and ExtensionLoader. The ExtensionLoader also brings in a runtime dep since there's no logic to avoid trying to load the ModuleLoader class even if ExtensionLoader can't work because the CLI is running non-modular.

In theory we could get rid of the hard runtime dep and make it optional. But there's no point doing that. :) In the end we are going to want embedded to work with jboss-cli-client.jar so we are going to include the modules classes in the jar.

So, ^^^ was a long way of saying that yes, it seems like 'hasModules' is not needed. :)

Your Class.forName approach should work fine too if you want to be conservative and not drop hasModules. Either is fine with me. 

A general thing to be cautious about is ensuring the Module class isn't loaded indirectly before the system package prop is set.

Re: what the system packages are, let's use the embedded module pom.xml as the authoritative source of such information.  The entries in the dependencies section have been organized into groups:

```
<dependencies>
        <!--
             All dependencies must be strictly controlled so it is clear what is needed
             by the embedding app. This means for nearly all dependencies we exclude all
             transitive deps, forcing them to be explicitly declared.

             Please keep dependencies organized by how they are used.
        -->

        <!-- Compile time dependencies that must be available at runtime
             via the embedding app's classloader -->
       ...

        <!-- Runtime only dependencies that must be on the embedding app's classpath -->
       ...

        <!-- Compile time dependencies that at runtime are loaded
             by the embedded process' modular classloader from its
             module path. These do not need to be visible to the
             embedding app's classloader and are thus treated as 'provided' -->
       ...

        <!-- Compile time only dependencies not needed at runtime -->
       ...

        <!-- Test dependencies -->
       ...

    </dependencies>
```

So the first two categories define what must be system packages. The discrepancy you mention perhaps is the second category?

bstansberry:
I'll go ahead and squash the commits now though.

jamezp:
@bstansberry +1 I think it's GTG.

jfdenise:
@bstansberry , @jamezp the 2 system packages that the CLI adds are "org.jboss.logging,org.jboss.logmanager", from the comment it is clear that they are required in the CLI context, otherwise logging would not work properly.
I have logged WFCORE-3935 in order to make the changes once you merged this fix.

PR #3329, opened 2018-05-28
[WFCORE-3831] added check and report of execution error

rsearls:


rsearls:
I am withdrawing this PR.   Someone from the wildfly CLI team should address this jira.  They know best the script and underlaying code.

PR #3336, opened 2018-05-31
[WFCORE-3868] Notify the suspend listener when timeout is 0

yersan:
If a ManagerExecutor is executing a task which is re-scheduling a new instance of itself just before exiting then the domain cannot be stopped when the timeout of stop operation is 0.
The problem is the listener is not notified when the user uses 0 as a timeout.

Since EE subsystem is not available in wildfly-core, I will add a test case in wildfly once this patch is merged and available from wildfly.

Jira issue: https://issues.jboss.org/browse/WFCORE-3868


jmesnil:
This PR can merged as it is. I've opened https://issues.jboss.org/browse/WFCORE-3905 to track Brian's comment for improvement

PR #3338, opened 2018-05-31
[WFCORE-3897] set explicitly domain model to available before domain ops boot up.

soul2zimate:
https://issues.jboss.org/browse/WFCORE-3897
Try to resolve required server group existence for configured server.

### DIFF
--- a/host-controller/src/main/java/org/jboss/as/host/controller/DomainModelControllerService.java
+++ b/host-controller/src/main/java/org/jboss/as/host/controller/DomainModelControllerService.java
@@ -240,11 +240,11 @@ public class DomainModelControllerService extends AbstractControllerService impl
     private final DomainDelegatingResourceDefinition rootResourceDefinition;
     private final CapabilityRegistry capabilityRegistry;
     private final DomainHostExcludeRegistry domainHostExcludeRegistry;
-    private final AtomicBoolean domainModelComplete = new AtomicBoolean(false);
+    private final AtomicBoolean domainConfigAvailable = new AtomicBoolean(false);
     private final PartialModelIndicator partialModelIndicator = new PartialModelIndicator() {
         @Override
         public boolean isModelPartial() {
-            return !domainModelComplete.get();
+            return !domainConfigAvailable.get();
         }
     };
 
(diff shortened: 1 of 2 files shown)


host-controller/src/main/java/org/jboss/as/host/controller/operations/StartServersHandler.java
```diff
@@ -112,6 +118,11 @@ private void cleanStartServers(final ModelNode servers, final ModelNode domainMo
         Map<String, ProcessInfo> processInfos = serverInventory.determineRunningProcesses();
         for(final Property serverProp : servers.asPropertyList()) {
             String serverName = serverProp.getName();
+
+            // WFCORE-3897 resolve required server group early no matter if server is auto-started
+            String serverGroupName = serverProp.getValue().get(GROUP).asString();
+            resolveExpressions(domainModel.require(SERVER_GROUP).require(serverGroupName), expressionResolver, true);
```
jmesnil: @soul2zimate I don't understand your patch.

Do you need to resolve the expressions of the `server-groups`?
It seems to me, you only need to check for the presence of a `server-group` resource named `serverGroupName`.

Why are you also resolving its expressions?

soul2zimate: You're right, it does not need to be expression resolved to check its value. Please hold this, there is a comment in WFCORE-3897 about detecting the missing server-group by capability validation once domain model is available(such validation is currently ignored as explanation at https://github.com/wildfly/wildfly-core/blob/5.0.0.Final/controller/src/main/java/org/jboss/as/controller/CapabilityRegistry.java#L860), I am investigating on it.

host-controller/src/main/java/org/jboss/as/host/controller/DomainModelControllerService.java
```diff
@@ -778,6 +778,8 @@ public ManagementResourceRegistration getRootResourceRegistrationForUpdate(Opera
                     List<ModelNode> domainBootOps = domainPersister.load();
 
                     HostControllerLogger.ROOT_LOGGER.debug("Invoking domain.xml ops");
+                    // https://issues.jboss.org/browse/WFCORE-3897
+                    domainModelComplete.set(true);
```
bstansberry: Can you change the name of this field, and the related params and field in RemoteDomainConnectionService, to 'domainConfigAvailable'?  The old 'domainModelComplete' name made sense because of the buggy code I wrote, where it wasn't true until the domain.xml ops had executed and the 'model' (i.e. the internal resource tree) was "complete".  But that's not really what it's used for. It's used to signal to the ModelController that the domain 'config' (i.e. either resource tree or just the ops needed to create it) is 'available' so it can go ahead and validate.

soul2zimate:
I have renamed them as per request, the ConcurrentFineWebFailoverTestCase failure is unrelated.

PR #3208, opened 2018-04-12
WFCORE-3736 WFCORE-3737 Remove sparc support from OpenSSL native bindings and upgrade to 1.0.6.Final

stuartwdouglas:

PR #3341, opened 2018-06-04
[WFCORE-3901] Remove wildfly-openssl-solaris-sparcv9 lincense

yersan:
This PR follows up https://github.com/wildfly/wildfly-core/pull/3208

From the point of view of distribution created by the feature pack, the wildfly-openssl-solaris-sparcv9 is not included in the final license, but since the library was removed, it looks like it makes sense remove the license as well.

Jira issue: https://issues.jboss.org/browse/WFCORE-3901

### DIFF
--- a/core-feature-pack/src/license/core-feature-pack-licenses.xml
+++ b/core-feature-pack/src/license/core-feature-pack-licenses.xml
@@ -1056,17 +1056,6 @@
     </dependency>
     <dependency>
       <groupId>org.wildfly.openssl</groupId>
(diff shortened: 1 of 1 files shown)


jstourac:
@yersan AFAIK removal of the solaris machine was done just as a temporary solution due to lack of building machines available. Solaris support should be put back to EAP7.2, see https://issues.jboss.org/browse/JBEAP-14596. We should assure this bit is not forgotten when moving back this support.

PR #3342, opened 2018-06-04
[WFCORE-3902] Support Maven staged repositories

jmesnil:
* add staging repository to MavenUtil

If the jboss.staging.repository.group System property is present, adds
JBoss staging repository to the remote repositories used to download
artifacts from MavenUtil.

JIRA: https://issues.jboss.org/browse/WFCORE-3902


jmesnil:
@kabir this is related to https://github.com/wildfly/wildfly/pull/11311 so that WildFly can use MavenUtil to download staged components.
The staging repo is added only when the sysprop is set by the `staged-releases` profile.

PR #3343, opened 2018-06-05
[WFCORE-3806] Eliminating dead code.

ropalka:
Prior MSC 1.2.8 stability monitors were reporting problems if
some transitive dependency wasn't available. This has changed
since MSC 1.2.8 and above. There is no notion of transitive
dependencies in JBoss MSC anymore. If stability monitor detects
a problem it is always immediate dependency of monitored controller.


ropalka:
Here is it @bstansberry (addressing WF 14 now)

PR #3344, opened 2018-06-06
[WFCORE-3907] Rename wildfly-component-matrix-builder to

ppalaga:
wildfly-core-component-matrix-builder to be able to import it into
Eclipse

https://issues.jboss.org/browse/WFCORE-3907

@jmesnil could you please review?

### DIFF
--- a/component-matrix-builder/pom.xml
+++ b/component-matrix-builder/pom.xml
@@ -30,7 +30,7 @@
     </parent>
 
     <groupId>org.wildfly.core</groupId>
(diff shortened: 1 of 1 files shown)


jmesnil:
@kabir is the change in the artifact name impacting your plugin?

kabir:
@jmesnil It should not, nothing should depend on the builder and the resulting artifact name is configured in https://github.com/wildfly/wildfly-core/pull/3344/files#diff-7bb8c0b1e1e088871585c4eaae3cdd26R58

PR #3345, opened 2018-06-06
[WFCORE-3908] When adding property to remoting check first if it is valid

rstancel:
Jira: https://issues.jboss.org/browse/WFCORE-3908

### DIFF
--- a/remoting/subsystem/src/main/java/org/jboss/as/remoting/ConnectorChildResource.java
+++ b/remoting/subsystem/src/main/java/org/jboss/as/remoting/ConnectorChildResource.java
@@ -18,6 +18,8 @@
 
 package org.jboss.as.remoting;
 
+import java.util.List;
+
 import org.jboss.as.controller.AttributeDefinition;
 import org.jboss.as.controller.OperationContext;
 import org.jboss.as.controller.OperationFailedException;
@@ -28,9 +30,12 @@ import org.jboss.as.controller.RestartParentResourceAddHandler;
 import org.jboss.as.controller.RestartParentResourceRemoveHandler;
 import org.jboss.as.controller.RestartParentWriteAttributeHandler;
 import org.jboss.as.controller.SimpleResourceDefinition;
+import org.jboss.as.controller.descriptions.ModelDescriptionConstants;
 import org.jboss.as.controller.descriptions.ResourceDescriptionResolver;
 import org.jboss.dmr.ModelNode;
+import org.jboss.dmr.Property;
 import org.jboss.msc.service.ServiceName;
+import org.wildfly.security.manager.WildFlySecurityManager;
 
 /**
  * @author Tomaz Cerar (c) 2015 Red Hat Inc.
@@ -63,7 +68,25 @@ abstract class ConnectorChildResource extends SimpleResourceDefinition {
 
         protected void populateModel(ModelNode operation, ModelNode model) throws OperationFailedException {
             for (AttributeDefinition ad : attributes) {
+                validateProperty(operation);
                 ad.validateAndSet(operation, model);
+            }
+        }
+
+        /**
+         * Perform remoting specific validation.
+         * @param operation
+         */
+        private void validateProperty(ModelNode operation) throws OperationFailedException {
+            ModelNode addressModelNode = operation.get(ModelDescriptionConstants.ADDRESS);
+            final ClassLoader loader = WildFlySecurityManager.getClassLoaderPrivileged(ConnectorChildResource.class);
+            if (addressModelNode.isDefined()) {
+                List<Property> propertyListModelNode = addressModelNode.asPropertyList();
+                for (Property property : propertyListModelNode) {
+                    if (property.getName().equals(CommonAttributes.PROPERTY)) {
+                        ConnectorUtils.getAndValidateOption(loader, property.getValue().asString());
+                    }
+                }
             }
         }
 
@@ -110,5 +133,4 @@ abstract class ConnectorChildResource extends SimpleResourceDefinition {
             return ConnectorChildResource.getParentServiceName(parentAddress);
         }
     }
-
 }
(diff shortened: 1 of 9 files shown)


remoting/subsystem/src/main/java/org/jboss/as/remoting/ConnectorChildResource.java
```diff
@@ -63,10 +66,27 @@ static ServiceName getParentServiceName(PathAddress parentAddress) {
...
             }
         }
 
+        /**
+         * Perform remoting specific validation.
+         * @param operation
+         */
+        private void validateProperty(ModelNode operation) {
+            ModelNode addressModelNode = operation.get(CommonAttributes.ADDRESS);
+            if (!addressModelNode.getType().toString().equals("u")) {
```
jmesnil: This condition is hard to understand.

Do you simply mean `if (addressModelNode.isDefined())` ?

remoting/subsystem/src/main/java/org/jboss/as/remoting/ConnectorUtils.java
```diff
@@ -156,16 +156,20 @@ private static void addSasl(OperationContext context, ModelNode sasl, OptionMap.
...
-                name = "org.xnio.Options." + name;
-            }
-            final Option option = Option.fromString(name, loader);
+            final Option option = getAndValidateOption(property.getName());
             String value = PropertyResource.VALUE.resolveModelAttribute(context, property.getValue()).asString();
             builder.set(option, option.parseValue(value, loader));
         }
     }
 
+    static Option<?> getAndValidateOption(String name) {
```
jmesnil: method should be private and should pass the `ClassLoader` in parameter instead of calling `WildFlySecurityManager.getClassLoaderPrivileged`every time.

remoting/tests/src/test/java/org/jboss/as/remoting/RemotingCLITestCase.java
```diff
@@ -0,0 +1,22 @@
+package org.jboss.as.remoting;
+
+import org.jboss.as.cli.scriptsupport.CLI;
+import org.junit.Assert;
+import org.junit.Ignore;
+import org.junit.Test;
+
+@Ignore // This should be run with started EAP
```
jmesnil: What is the point of adding this test if it is ignored?

I suppose this test should be moved to the basic integration test suite.

remoting/subsystem/src/main/java/org/jboss/as/remoting/ConnectorChildResource.java
rstancel: @jmesnil You are right. I have fixed it.

remoting/subsystem/src/main/java/org/jboss/as/remoting/ConnectorUtils.java
rstancel: @jmesnil I have added passing ClassLoader as a parameter but it meant to add it also in ConnectorChildResource. And that's the other thing. Because of ConnectorChildResource it cannot be private

remoting/tests/src/test/java/org/jboss/as/remoting/RemotingCLITestCase.java
rstancel: @jmesnil You mean here: https://github.com/wildfly/wildfly-core/tree/master/testsuite/standalone or https://github.com/wildfly/wildfly/tree/master/testsuite/integration ?

rstancel: Fixed

rstancel:
@jmesnil Jeff, I have fixed your comments. Can you please review now?

remoting/subsystem/src/main/java/org/jboss/as/remoting/CommonAttributes.java
```diff
@@ -27,6 +27,7 @@
  */
 interface CommonAttributes {
 
+    String ADDRESS = "address";
```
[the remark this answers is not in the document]
rstancel: @jmesnil I have changed the ADDRESS attribute

Regarding write-attribute test. It makes no sense. I am checking the name of the property, not the value. So when updating something, the names always renames the same. Right now, if there was some bad name you will not start the server (issue I am fixing) and only possibility how to put it there is add :)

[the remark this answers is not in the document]
jmesnil: @rstancel you are right, I got confused.

remoting/subsystem/src/main/java/org/jboss/as/remoting/ConnectorUtils.java
```diff
@@ -156,16 +156,19 @@ private static void addSasl(OperationContext context, ModelNode sasl, OptionMap.
...
             String value = PropertyResource.VALUE.resolveModelAttribute(context, property.getValue()).asString();
             builder.set(option, option.parseValue(value, loader));
         }
     }
 
+    static Option<?> getAndValidateOption(ClassLoader loader, String name) {
+        if (!name.contains(".")) {
+            name = "org.xnio.Options." + name;
+        }
+        return Option.fromString(name, loader);
```
bstansberry: This should catch any IAE and throw an OperationFailedException. An IAE will propagate up and be handled as a server fault, and this is not a server fault. OFE is for user mistakes.

The existing code (i.e. L163) of course had this same problem. :)

dmlloyd:
No, options have been deprecated but not removed IIRC.

remoting/subsystem/src/main/java/org/jboss/as/remoting/ConnectorUtils.java
rstancel: @bstansberry Fixed

PR #3346, opened 2018-06-06
Fix for WFCORE-3910, CLI remove dependency on jdk.jconsole

jfdenise:
Fix for WFCORE-3910: https://issues.jboss.org/browse/WFCORE-3910
Has been introduced due to jdeps advertising jdk.jconsole as a dependency. But this dependency is only present in a jconsole context, outside jboss modules.


jmesnil:
retest this please

PR #3351, opened 2018-06-08
[WFCORE-3919] missing algorithms into elytron jdbc scram-mapper

hkalina:
https://issues.jboss.org/browse/WFCORE-3919


hkalina:
Reworked to extend schema 4.0 and added transfomers.

marschall:
How does the versioning of schemas work? What I mean by this is why is the version bumped to 4.0 and not to 3.1 or 3.0.1 when just two enum values are added?

hkalina:
@marschall there are another changes in queue based on the same schema bump commit - this is only part of the change

marschall:
@hkalina fair enough

darranl:
@marschall overall the version bumps are linked to knowing the application server will be getting a major version bump before this is included in a .Final release - although we don't necessarily have maintenance releases of WildFly 13 scheduled by not blocking the minor versions we have those available to use.

Additionally as @hkalina says we also have a number of engineers working in parallel on configuration changes but as this one is a relatively small change to the model it is the first through the queue.

hkalina:
Hi, thanks, however the addition of the 2 values in the scramMapperType IS in the commit for WFCORE-3919 (in different commit then XSD upgrade) ;)
https://github.com/wildfly/wildfly-core/pull/3351/commits/a8ddd1444cb67781bfdda2feb58039fdc9b4c789#diff-1bf7bcebfe55b84436318182d452701fL1293

jmesnil:
@hkalina you are right, I missed it in GitHub UI and i diffed the XSD from the whole PR :)

PR #3354, opened 2018-06-11
[WFCORE-3337] Add capabilities to the logging subsystem

jamezp:
https://issues.jboss.org/browse/WFCORE-3337 

Capability PR: https://github.com/wildfly/wildfly-capabilities/pull/26

The first commit is a simple change to allow string module identifiers to be registered for deployments.

The next commit refactors the logging subsystem. The bulk of the change is moving things into better packages. There are also some function changes for using the better construction parameters for the resource definitions as well as using the correct abstract OSH's for add and remove operations. 

The last commit adds the capabilities to the subsystem. This could be squashed with the previous commit, but it seemed easier to review this way.


bstansberry:
This is a big PR to review in detail but I've skimmed through it and it looks fine to me. And you know your way around the management code so you know what you're about with the big refactoring commit.

I vote keeping these as separate commits.  Squashing them will really bury the key bits from the 3rd commit.

darranl:
Another +1 to avoid squashing if the commits make sense to break down the changes but maybe their commit messages should also reference the Jira?

logging/src/main/java/org/jboss/as/logging/capabilities/LoggingProfileCapabilityRecorder.java
```diff
@@ -0,0 +1,113 @@
...
+import org.jboss.as.controller.PathElement;
+import org.jboss.as.controller.registry.Resource;
+import org.jboss.as.logging.CommonAttributes;
+
+/**
+ * Allows the name of the logging profile to be appended to the capability name.
+ *
+ * @author <a href="mailto:jperkins@redhat.com">James R. Perkins</a>
+ */
+class LoggingProfileCapabilityRecorder implements CapabilityReferenceRecorder {
```
jmesnil: @jamezp maybe the logging code could benefit from https://github.com/wildfly/wildfly-core/blob/master/controller/src/main/java/org/jboss/as/controller/CapabilityReferenceRecorder.java#L325 with the right mapper?

jamezp: Unfortunately I don't think it will work. The dynamic name is always based on the resource address so the value of the requirement gets the name from the address rather than the attribute.

PR #3356, opened 2018-06-12
[WFCORE-3923] Provide a profile to run enforcer rule for downstream p…

gaol:
…rojects - wildfly-core

Added a profile enforce-product to check redhat dependencies

Jira: https://issues.jboss.org/browse/WFCORE-3923

### DIFF
--- a/core-feature-pack/pom.xml
+++ b/core-feature-pack/pom.xml
@@ -686,6 +686,44 @@
             </build>
         </profile>
         <profile>
+            <id>enforce-product</id>
+            <activation>
+                <property>
+                    <name>enforce-product</name>
+                </property>
+            </activation>
+            <build>
+                <plugins>
+                    <plugin>
+                        <groupId>org.apache.maven.plugins</groupId>
+                        <artifactId>maven-enforcer-plugin</artifactId>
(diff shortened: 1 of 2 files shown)


gaol:
retest this please

gaol:
@jmesnil Could you please re-run the failed CI jobs and to see if it is OK to merge?
BTW, the similar PR for WildFly https://github.com/wildfly/wildfly/pull/11342 has been merged.

jmesnil:
retest this please

jmesnil:
@gaol Could the `enforce-product` profile be declared in the root `pom.xml` instead of duplicating it in the `core-feature-pack`  and `core-galleon-pack` modules?

gaol:
@jmesnil it is intended to be only included in `core-feature-pack` and `core-galleon-pack`, so that we can only run enforcer rule for these 2 distributed packs in EAP CP streams.

In downstream, all artifacts in these 2 distributed packs are required built from prod team(*-redhat-X). But we still have other community version dependencies, like some test scope dependencies, build tools related dependencies etc. so that we cannot have the profile in the root `pom.xml` which will enforce all modules.

PR #3359, opened 2018-06-13
[WFCORE-3666] Provide Elytron alternative to RoleMappingLoginModule

nekdozjam:
https://issues.jboss.org/browse/WFCORE-3666

requires ELY-1591
https://issues.jboss.org/browse/ELY-1591

### DIFF
--- a/elytron/src/main/java/org/wildfly/extension/elytron/ElytronSubsystemTransformers.java
+++ b/elytron/src/main/java/org/wildfly/extension/elytron/ElytronSubsystemTransformers.java
@@ -88,6 +88,7 @@ public final class ElytronSubsystemTransformers implements ExtensionTransformerR
                         ))
                 ), PRINCIPAL_QUERIES)
                 .end();
+        builder.rejectChildResource(PathElement.pathElement(ElytronDescriptionConstants.MAPPED_ROLE_MAPPER));
     }
 
     private static void from3(ChainedTransformationDescriptionBuilder chainedBuilder) {
--- a/elytron/src/main/java/org/wildfly/extension/elytron/MapperParser.java
+++ b/elytron/src/main/java/org/wildfly/extension/elytron/MapperParser.java
@@ -30,6 +30,7 @@ import static org.wildfly.extension.elytron.ElytronDescriptionConstants.PRINCIPA
 import static org.wildfly.extension.elytron.ElytronDescriptionConstants.PRINCIPAL_TRANSFORMER;
 import static org.wildfly.extension.elytron.ElytronDescriptionConstants.REALM_MAPPING;
 import static org.wildfly.extension.elytron.ElytronDescriptionConstants.ROLE_MAPPER;
+import static org.wildfly.extension.elytron.ElytronDescriptionConstants.ROLE_MAPPING;
 import static org.wildfly.extension.elytron.ElytronDescriptionConstants.TO;
 
 import javax.xml.stream.XMLStreamException;
@@ -44,6 +45,8 @@ import org.jboss.as.controller.PersistentResourceXMLDescription;
 import org.jboss.as.controller.parsing.ParseUtils;
 import org.jboss.dmr.ModelNode;
 import org.jboss.staxmapper.XMLExtendedStreamReader;
+
+import java.util.Iterator;
 
 /**
  * XML handling for the <mappers /> element.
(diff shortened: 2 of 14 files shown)


jmesnil:
this is ok to test

elytron/src/main/resources/schema/wildfly-elytron_4_0.xsd
```diff
@@ -2245,6 +2245,7 @@
             <xs:element name="constant-role-mapper" type="constantRoleMapperType" minOccurs="0" maxOccurs="unbounded" />
             <xs:element name="custom-role-mapper" type="customRoleMapperType" minOccurs="0" maxOccurs="unbounded" />
             <xs:element name="logical-role-mapper" type="logicalRoleMapperType" minOccurs="0" maxOccurs="unbounded" />
+            <xs:element name="mapped-role-mapper" type="mappedRoleMapperType" minOccurs="0" maxOccurs="unbounded" />
```
hkalina: Please, when adding new elements, add example of their usage into `elytron-subsystem-4.0.xml`

elytron/src/main/java/org/wildfly/extension/elytron/MapperParser.java
```diff
@@ -167,6 +170,37 @@ public void marshallSingleElement(AttributeDefinition attribute, ModelNode prope
             .addAttribute(RoleDecoderDefinitions.ATTRIBUTE)
             .build();
 
+    private PersistentResourceXMLDescription mappedRoleMapperParser = PersistentResourceXMLDescription.builder(RoleMapperDefinitions.getMappedRoleMapperDefinition().getPathElement())
+            .addAttribute(RoleMapperDefinitions.KEEP_MAPPED)
+            .addAttribute(RoleMapperDefinitions.KEEP_NON_MAPPED)
+            .addAttribute(RoleMapperDefinitions.ROLE_MAPPING_MAP, new AttributeParsers.MapParser(null, ROLE_MAPPING, false) {
```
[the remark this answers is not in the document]
nekdozjam: But MAPPED_REGEX_REALM_MAPPER uses PropertiesAttributeDefinition, which is a STRING to STRING map, internally it has no from/to attributes, it is an OBJECT where attribute name represents delegate role and value represents mapped role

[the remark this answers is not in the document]
hkalina: Sorry, I have mislooked - this is consistent. OK then.

darranl:
I am suspecting there may be a bug affecting the model representation, looking at the resource definition it is defined as a SimpleMapAttributeDefinition

elytron/src/main/java/org/wildfly/extension/elytron/RoleMapperDefinitions.java
```diff
@@ -94,6 +100,24 @@
         .setRestartAllServices()
         .build();
 
+    static final SimpleMapAttributeDefinition ROLE_MAPPING_MAP = new SimpleMapAttributeDefinition.Builder(ElytronDescriptionConstants.ROLE_MAP, ModelType.LIST, false)
+            .setMinSize(1)
+            .setAllowExpression(true)
+            .setRestartAllServices()
+            .build();
+
+    static final SimpleAttributeDefinition KEEP_MAPPED = new SimpleAttributeDefinitionBuilder(ElytronDescriptionConstants.KEEP_MAPPED, ModelType.BOOLEAN, true)
```
darranl: Are both booleans needed?  By having two boolean values we end up with four permutations, do they all definitely make sense?

hkalina: KEEP_MAPPED can be omitted, as we can define the map as:
`originalRole -> newRole,originalRole`
Both options can be replaced by using `logical-role-mapper` - I would suggest removing both of them.

darranl: I think however keeping one does make sense, relying on a logical role mapper is actually quite a bit more configuration compared to a single boolean. - Just want to be sure four permutations make sense but as you say we can actually eliminate one of these quite easy by mapping back to the same role.

mchoma: I personally like idea of full control of logic in mapper itself. I think all of this four permutation can be useful in some situations.

At least legacy solution provide 2 permutation of them https://access.redhat.com/documentation/en-us/red_hat_jboss_enterprise_application_platform/7.0/html-single/login_module_reference/#rolemapping_login_module

In situation when we face complaints about Elytron beeing difficult to configure I wouldn't complicate things with another resource (logical-role-mapper) if it is not really necessary.

elytron/src/main/java/org/wildfly/extension/elytron/ElytronSubsystemTransformers.java
```diff
@@ -88,6 +88,7 @@ private static void from4(ChainedTransformationDescriptionBuilder chainedBuilder
                         ))
                 ), PRINCIPAL_QUERIES)
                 .end();
+        builder.rejectChildResource(PathElement.pathElement(ElytronDescriptionConstants.MAPPED_ROLE_MAPPER));
```
hkalina: please cover this by transformers tests too (check https://hkalina.github.io/2018/05/15/transformers/ )

darranl:
I am going to debug and see what is happening regarding the model.

darranl:
I have been back and double checked, the representation in the model does seem to be correct, in DMR an OBJECT is effectively a Map.

I don't see that as long as we are using a SimpleMapAttributeDefinition we will have opportunities to name things 'to' and 'from' as we don't seem to have that level of control in the model definition, I think the best we can do is ensure the resource has a sufficient description.

We could look at some form List of a custom type where each entry in the list has from and to values but then I think we make it harder for tooling to automatically generate anything - also we end up loosing the exclusive key checking you get by using an Object.

darranl:
Overall I think it is just transformers that Honza mentions that need to be checked for this PR.

bstansberry:
@darranl @mchoma I haven't been able to look at this at all, beyond the discussion in the comments, so apologies if this reply is useless.

An attribute whose type is ModelType.OBJECT but which has a fixed set of DMR object keys is not a "Map".  It would use ObjectTypeAttributeDefinition.

If the issue is that there isn't a fixed set of keys, but there are a couple "well known" keys, "from" and "to", then it is a Map and there's no mechanism for describing those well known keys outside of the text description of the attribute. And that of course is not useful for things like tab completion.

hkalina:
Looks ok - transformers are correctly test-covered.

darranl:
@wildfly-core/mergers Can we please get the rebase this and missing reqs labels removed, the proposal is here: -

https://github.com/wildfly/wildfly-proposals/pull/70

darranl:
@wildfly/mergers-core Can we please get the rebase this and missing reqs labels removed, the proposal is here: -

https://github.com/wildfly/wildfly-proposals/pull/70

darranl:
@jmesnil @bstansberry @kabir @jamezp The group ping doesn't seem to be working for me ^^^

kabir:
@darranl Most of the pre-check requirements are still missing on the EAP7 issue, so I am leaving the missing-reqs label

mchoma:
Can it be retested on CI?

kabir:
@jmesnil I've removed the missing reqs label as this all appears to be done now

PR #3372, opened 2018-06-20
WFCORE-3939 Add PersistentResourceXMLParser support for logical grouping of child resources

pferraro:
https://issues.jboss.org/browse/WFCORE-3939

Enables usage of PersistentResourceXMLParser in subsystems where child resources are logically grouped (e.g. clustering subsystems).

### DIFF
--- a/controller/src/main/java/org/jboss/as/controller/PersistentResourceXMLDescription.java
+++ b/controller/src/main/java/org/jboss/as/controller/PersistentResourceXMLDescription.java
@@ -36,7 +36,7 @@ import org.jboss.staxmapper.XMLExtendedStreamWriter;
  */
 public final class PersistentResourceXMLDescription implements ResourceParser, ResourceMarshaller {
 
-    protected final PathElement pathElement;
+    private final PathElement pathElement;
     private final String xmlElementName;
     private final String xmlWrapperElement;
     private final LinkedHashMap<String, LinkedHashMap<String, AttributeDefinition>> attributesByGroup;
@@ -125,13 +125,7 @@ public final class PersistentResourceXMLDescription implements ResourceParser, R
         return this.pathElement;
     }
 
-    /**
-     * Parse xml from provided <code>reader</code> and add resulting operations to passed list
-     * @param reader xml reader to parse from
-     * @param parentAddress address of the parent, used as base for all child elements
-     * @param list list of operations where result will be put to.
-     * @throws XMLStreamException if any error occurs while parsing
-     */
+    @Override
     public void parse(final XMLExtendedStreamReader reader, PathAddress parentAddress, List<ModelNode> list) throws XMLStreamException {
         if (decoratorElement != null) {
             parseDecorator(reader, parentAddress, list);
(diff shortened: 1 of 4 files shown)


pferraro:
@jmesnil Please review.

pferraro:
@jmesnil Any comments?

pferraro:
@jmesnil Any comments?

pferraro:
@jmesnil Any chance you can review this soon?  This is a dependency of several WF pull requests.

jmesnil:
This PR and the associated WFCORE-3939 looks problematic as it subverts the semantic of the management metadata to drive the XML representation.
The desired output XML format must not require to provide invalid metadata semantic.

Looking at the provided test, the `subsystem` resource has a `default-channel` attribute.
This will naturally be represented in XML by:

```
<subsystem default-channel="foo">
   <channel name="foo>
      ...
   </channel>
<subsystem>
```

However this PR wants to force an alternative XML representation. In this only purpose, the test changes the xml name of the `default-channel` to `default` __and its `attributeGroup` to `channels`__. 

This is not correct.

 Attribute group is meant to group logically related attributes of a resource (e.g. `cluster-user` and `cluster-password` can both belong to a `cluster` attributeGroup).
However in this case, `default-channel` is the _only_ attribute in the `channels` attributeGroup. This provides no additional information and is a misuse of the management model semantic.

I understand that rejecting this PR means that the clustering extensions as they stand can not use the standard XML parser.
However the key advantage of having extensions use the standard XML parser is to provide consistent configuration representation across extensions.
With that PR, we would introduce 2 ways to represent "default among child resources":
* the standard one (where the `default-xxx` attribute is on the XML element of the parent resource)
* yours where the `default` attribute is on the XML wrapper of the child resource.

I don't see a good reason to introduce 2 different representations when the 1st one is fine and makes the mental exercise to map between the XML elements and the management resource straightforward.

jmesnil:
> PersistentResourceXMLDescription is not generic management metadata.

I agree. The problematic part of this issue is about `AttributeDefinion.attributeGroup`. With this PR, you would use the `attributeGroup` of an attribute for the sole purpose of making the XML looks "good". That's not correct.

PersistentResourceXMLDescription already has too many ways to tweak the XML output. They are in the API now and we must maintain them. 
We might even deprecate some to improve the consistency of our configurations but that's low priority.
In any case, we need to be much more conservative to add such new API that make the code more complex and the configuration less consistent.

Every time a subsystem developer decides to tweak the XML output so that it looks good (and believe me, I did that too many times for the messaging subsystems), we make it more complex for our users to understand our configuration as a whole. We have 1 way to represent child groups in clustering subsystems, another in undertow and yet another in elytron... For our users, that's unnecessary noise that makes it even more complex to properly configure WildFly as a whole.

We might all have different opinion on the best XML representation but
_configuration consistency must come first_ as it's a win-win for developers (that now longer have to maintain XML parser in their extensions) and users (that don't have to figure out how each extensions represents the same things whether it is properties, child groups, etc.)

pferraro:
Reviewing the initial concept for attribute groups: http://lists.jboss.org/pipermail/wildfly-dev/2014-December/003414.html
... clearly, the ability to persist all attributes of an attribute group within a single XML element unit (i.e. to make the XML look good) was a core design goal.

I just want to make sure you are rejecting WFCORE-3939 based on a rejection of the general concept (i.e. that you think child resources that are associated with a specific attribute group should be persisted as sibling elements rather than as nested elements) and not just based on the example where the attribute group contains only 1 attribute.

jmesnil:
> 2) XML parsing/marshalling

> Modify PersistentResourceXMLDescription such that attributes in an 
> attribute group get persisted in their own child element, whose name is 
> the name of the group.

This is still valid. I'm rejecting this PR that would allow to use the same XML element for both:

1. representing an attribute group
2. wraps other child resources

The example illustrated how this new enhancement would be abused to make the XML look good at the detriment of incorrect metadata information that would cause issues for other clients of the management model (e.g. HAL).

PR #3373, opened 2018-06-22
[WFCORE-3930] Possible file permission issues in .properties files

aloubyansky:
https://issues.jboss.org/browse/WFCORE-3930


jmesnil:
@aloubyansky Could you please confirm that we do not need to apply the same kind of fix in WildFly's `galleon-pack` & `servlet-galleon-pack`?

aloubyansky:
@jmesnil yes, this config is inherited from core, there is no corresponding file in wildfly codebase.

PR #3375, opened 2018-06-25
[WFCORE-3920] Update test suite to use certificate generation utilities

justinmcook:
Jira: https://issues.jboss.org/browse/WFCORE-3920


elytron/src/test/java/org/wildfly/extension/elytron/KeyStoresTestCase.java
```diff
@@ -96,8 +114,197 @@ private ModelNode assertFailed(ModelNode response) {
...
+        File testFile = new File(workingDir, "test.keystore");
+
+        KeyStore fireflyKeyStore = createFireflyKeyStore();
+        KeyStore testKeyStore = createTestKeyStore(rootSelfSignedX509CertificateAndSigningKey, sallySelfSignedX509CertificateAndSigningKey);
+
+        createTemporaryKeyStoreFile(fireflyKeyStore, fireflyFile);
+        createTemporaryKeyStoreFile(testKeyStore, testFile);
+
+        // Cert Files
+        File testSingleCertReplyFile = new File(workingDir, "test-single-cert-reply.cert");
```
fjuma: To avoid some duplication, you could create static variables for the file declarations and then reference them in this method and in ```removeTestFiles```.

elytron/src/test/java/org/wildfly/extension/elytron/KeyStoresTestCase.java
```diff
@@ -96,8 +114,197 @@ private ModelNode assertFailed(ModelNode response) {
...
+        createTemporaryCertFile(testUntrustedCert, testUntrustedFile);
+
+        createTemporaryCertFile(testUntrustedCertChainReplyCert, testUntrustedCertChainReplyFile);
+        createTemporaryCertFile(anotherRootSelfSignedX509CertificateAndSigningKey.getSelfSignedCertificate(), testUntrustedCertChainReplyFile);
+    }
+
+    private static void removeTestFiles() {
+        File workingDir = new File(WORKING_DIRECTORY_LOCATION);
+        if (workingDir.exists() == false) {
+            workingDir.mkdirs();
```
fjuma: No need to create the directory if it doesn't exist. The ```file.exists()``` check below ensures that the file will only be deleted if it actually exists.

elytron/src/test/java/org/wildfly/extension/elytron/TlsTestCase.java
```diff
@@ -60,15 +86,193 @@
...
+    }
+
+    private static void setUpKeyStores() throws Exception {
+        File workingDir = new File(WORKING_DIRECTORY_LOCATION);
+        if (workingDir.exists() == false) {
+            workingDir.mkdirs();
+        }
+
+        SelfSignedX509CertificateAndSigningKey issuerSelfSignedX509CertificateAndSigningKey = createIssuer();
+        File trustFile = new File(workingDir, "ca.truststore");
```
fjuma: Same thing here with the File declarations. Static variables could be used and referenced in this method and in ```deleteKeyStoreFiles()```.

testsuite/shared/src/main/java/org/jboss/as/test/integration/auditlog/AuditLogToTLSSyslogSetup.java
```diff
@@ -105,6 +198,7 @@ protected SyslogServerConfigIF getSyslogConfig() {
     public void setup(ManagementClient managementClient) throws Exception {
         FileUtils.deleteDirectory(WORK_DIR);
         WORK_DIR.mkdirs();
+        beforeTest();
```
fjuma: Are the three ```createTestResource()``` calls below still needed here? Looks like these are being created from ```beforeTest()``` now.

elytron/src/test/java/org/wildfly/extension/elytron/KeyStoresTestCase.java
justinmcook: Fixed

elytron/src/test/java/org/wildfly/extension/elytron/KeyStoresTestCase.java
justinmcook: Fixed

elytron/src/test/java/org/wildfly/extension/elytron/TlsTestCase.java
justinmcook: Fixed

testsuite/shared/src/main/java/org/jboss/as/test/integration/auditlog/AuditLogToTLSSyslogSetup.java
justinmcook: They are still needed because they copy the created files to audit-workdir, which is best done using those calls because of where the directory is located

elytron/src/test/java/org/wildfly/extension/elytron/LdapTestCase.java
```diff
@@ -62,13 +79,72 @@
...
+
+    private static void loadResources() throws Exception {
+        SCARAB_CERTIFICATE = createScarabCertificate();
+        rewriteLdapDataLdif(SCARAB_CERTIFICATE);
+    }
 
     @BeforeClass
-    public static void startLdapService() {
+    public static void startLdapService() throws Exception {
+        loadResources();
```
fjuma: Looks like this test class is missing code to delete the test files created by ```TestEnvironment.startLdapService()```.

justinmcook: Fixed

PR #3376, opened 2018-06-25
WFCORE-3944 NPE in SyncServerStateOperationHandler when group is not …

TomasHofman:
…present on host

The SyncServerStateOperationHandler throws a NPE when the synchronizing a deployment that is deployed on a server group which doesn't has any server on current host.

https://issues.jboss.org/browse/WFCORE-3944
https://issues.jboss.org/browse/JBEAP-14817


bstansberry:
retest this please

PR #3379, opened 2018-07-05
[WFCORE-3945] Remove the application-http-authentication http-authentication-factory from the default configuration.

darranl:
This pull request is 2 of 3 and can be merged after https://github.com/wildfly/wildfly/pull/11400

https://issues.jboss.org/browse/WFCORE-3945

This pull request removes the application-http-authentication http-authentication-factory from the default configuration, it also contains updates to the JBoss CLI to use the new attribute added under https://github.com/wildfly/wildfly/pull/11400 and to no longer depend on the removed default configuration.


darranl:
Test failures are expected as the first pull request is yet to be merged.

PR #3386, opened 2018-07-06
[WFCORE-2953] Add a socket-handler resource to the logging subsystem.

jamezp:
Associated JIRA's:
https://issues.jboss.org/browse/WFCORE-2953 - `socket-handler`
https://issues.jboss.org/browse/WFCORE-3927 - logging subsystem schema bump
https://issues.jboss.org/browse/WFCORE-3946 - test fix

Design Document: https://github.com/wildfly/wildfly-proposals/pull/103

Community documentation is coming.

This adds a `socket-handler` resource to the logging subsystem. An `/socket-binding=*/*-destination-outbound-socket-binding` is required the connections. This allows a standard socket binding resource to be used rather than require the `socket-binding` resource to handle it's own connections.

### DIFF
--- a/logging/src/main/java/org/jboss/as/logging/handlers/SocketHandlerResourceDefinition.java
+++ b/logging/src/main/java/org/jboss/as/logging/handlers/SocketHandlerResourceDefinition.java
@@ -1,0 +1,472 @@
+/*
+ * JBoss, Home of Professional Open Source.
+ *
+ * Copyright 2018 Red Hat, Inc., and individual contributors
+ * as indicated by the @author tags.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.jboss.as.logging.handlers;
+
+import static org.jboss.as.logging.CommonAttributes.AUTOFLUSH;
+import static org.jboss.as.logging.CommonAttributes.ENABLED;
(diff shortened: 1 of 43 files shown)


jamezp:
Documentation PR https://github.com/wildfly/wildfly/pull/11406

jamezp:
This needs a way to configure an SSL context. I mistakenly thought it was configurable a different way.

logging/src/main/java/org/jboss/as/logging/handlers/SocketHandlerResourceDefinition.java
```diff
@@ -0,0 +1,472 @@
...
+            .build();
+
+    public static final SimpleAttributeDefinition OUTBOUND_SOCKET_BINDING_REF = SimpleAttributeDefinitionBuilder.create("outbound-socket-binding-ref", ModelType.STRING, false)
+            .addAccessConstraint(SensitiveTargetAccessConstraintDefinition.SOCKET_BINDING_REF)
+            .setAllowExpression(true)
+            .setCapabilityReference(Capabilities.OUTBOUND_SOCKET_BINDING_CAPABILITY)
+            .addFlag(AttributeAccess.Flag.RESTART_RESOURCE_SERVICES)
+            .build();
+
+    public static final SimpleAttributeDefinition PROTOCOL = SimpleAttributeDefinitionBuilder.create("protocol", ModelType.STRING, true)
```
jamezp: Just a minor note for reviewers. The management audit logging for the `syslog-handler` there is a child resource of `protocol` with the values; `tcp`. `tls` and `upd`. For the Elytron `syslog-audit-log` there is an attribute called `transport` with the values `TCP`, `SSL_TCP` and `UDP`.

I went with `protocol` for the attribute name as the `SocketHandler` in the log manager has an enum named `Protocol` with the values of `TCP`, `SSL_TCP` and `UDP`. However maybe `transport` would be better to keep the names consistent. I can change this if it's agreed it should be changed.

jmesnil:
@jamezp Feature freeze for WildFly Core is tomorrow (August 10th). It seems there are still missing requirements before this PR can be merged (in particular Test Dev are still in progress). Could you please check that the pre-merge requirement parallel tasks are up to date and report any update?

jamezp:
@jmesnil this now meets all the requirements.

PR #3391, opened 2018-07-16
[WFCORE-3963] Fix of WFCORE-3826 breaks plain authentic…

JiriOndrusek:
…ation for ejbs using legacy configuration

Issue: https://issues.jboss.org/browse/WFCORE-3963

Test: https://github.com/wildfly/wildfly/pull/11432

### DIFF
--- a/domain-management/src/main/java/org/jboss/as/domain/management/security/SecurityRealmService.java
+++ b/domain-management/src/main/java/org/jboss/as/domain/management/security/SecurityRealmService.java
@@ -590,7 +590,7 @@ public class SecurityRealmService implements Service<SecurityRealm>, SecurityRea
 
             for(Iterator<String> iter = requestedMechanisms.iterator();iter.hasNext();) {
                 AuthMechanism authMechanism = toAuthMechanism("SASL", iter.next());
-                if(authMechanism != null) {
+                if(authMechanism != null && (registeredServices.containsKey(authMechanism) || (authMechanism == AuthMechanism.PLAIN && registeredServices.containsKey(AuthMechanism.DIGEST)))) {
                    iter.remove();
                 }
             }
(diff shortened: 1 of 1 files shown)


xstefank:
@JiriOndrusek [WFCORE-3963] is mentioned twice in commit and PR title messages

JiriOndrusek:
@xstefank Bad ctrl+c, I'll fix it, thanks

kabir:
@darranl @fjuma please approve

domain-management/src/main/java/org/jboss/as/domain/management/security/SecurityRealmService.java
```diff
@@ -590,7 +590,7 @@ public SaslAuthenticationFactory getSaslAuthenticationFactory(final String[] mec
 
             for(Iterator<String> iter = requestedMechanisms.iterator();iter.hasNext();) {
                 AuthMechanism authMechanism = toAuthMechanism("SASL", iter.next());
-                if(authMechanism != null) {
+                if(authMechanism != null && (registeredServices.containsKey(authMechanism) || registeredServices.containsKey(AuthMechanism.DIGEST))) {
```
fjuma: @JiriOndrusek Why is there a special case for DIGEST here and below? Could you explain the reason for these changes a bit? Thanks!

JiriOndrusek: Now I understand it correctly, (I hope):
- fix should allow anonymous - no problem
- fix should allow only these mechanisms, which are currently enabled by elytron, with exception for PLAIN which could use DIGEST (this substitution works, I've tested it)

@darranl  Is my understanding correct?

JiriOndrusek:
@fjuma I'll change fix according to result of discussion:
- fix should allow only these mechanisms, which are currently enabled by elytron, with exception for PLAIN which could use DIGEST

JiriOndrusek:
@fjuma changes are commited

PR #3397, opened 2018-07-22
[WFCORE-3971] Referral mode should be an enumeration

marschall:
https://issues.jboss.org/browse/WFCORE-3971

### DIFF
--- a/elytron/src/main/java/org/wildfly/extension/elytron/DirContextDefinition.java
+++ b/elytron/src/main/java/org/wildfly/extension/elytron/DirContextDefinition.java
@@ -42,7 +42,7 @@ import org.jboss.as.controller.SimpleAttributeDefinition;
 import org.jboss.as.controller.SimpleAttributeDefinitionBuilder;
 import org.jboss.as.controller.SimpleResourceDefinition;
 import org.jboss.as.controller.capability.RuntimeCapability;
-import org.jboss.as.controller.operations.validation.EnumValidator;
+import org.jboss.as.controller.operations.validation.StringAllowedValuesValidator;
 import org.jboss.as.controller.registry.ManagementResourceRegistration;
 import org.jboss.as.controller.registry.OperationEntry;
 import org.jboss.as.controller.security.CredentialReference;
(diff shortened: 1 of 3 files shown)


marschall:
@jmesnil did I break all those tests? I can't login to the CI server?

jamezp:
There should be a "Log in as guest" link that will allow you to see it. But yes it breaks the build. It looks like previously some configurations use `THROW` which isn't allowed with the new constraints to lowercase only.

marschall:
There was an additional issue that model serialization would generate ALL-CAPS values.

jamezp:
@marschall In that case we might not be able to do this or we're going to have to allow for both upper/lower case. Otherwise we break compatibility.

darranl:
Yeah updating the schema is Ok as that is a new schema.

We can update the code now to always write using lowercase, that will be fine as the write always uses the latest schema.

But we must continue to support older configuration which may use upper or lower case.

jamezp:
+1 to what @darranl says. It should just be the parser that needs to be updated.

marschall:
@darranl I'm not that familiar with the model and the parsers, how do I change it so that lower case values are written?
Also I don't understand why both `setAllowedValues` and `.setValidator(EnumValidator)` are called.

Currently a `.toUpperCase()` is already performed so it looks as if the values are already case insensitive
https://github.com/wildfly/wildfly-core/blob/master/elytron/src/main/java/org/wildfly/extension/elytron/DirContextDefinition.java#L194

kabir:
@marschall I am just jumping in real quick without really remembering how this works, or having looked much at the failure. But in case it helps take a look at https://github.com/wildfly/wildfly/blob/master/ejb3/src/main/java/org/jboss/as/ejb3/subsystem/StrictMaxPoolResourceDefinition.java#L75 which accepts lower case values https://github.com/wildfly/wildfly/blob/master/ejb3/src/main/resources/subsystem-templates/ejb3.xml#L16. The magic seems to be in the DeriveSize enum

bstansberry:
I think changing ReferralMode.toString() to return getValue() will solve this with no need to change the management code.

jmesnil:
closing this PR as it is superseded by #3523

PR #3401, opened 2018-07-24
[WFCORE-3974] Adjust AuditLogBootingSyslogTest test case to test number of operations within the audit log events instead of counting how many events they are chopped into.

darranl:
:warning: Do not Rebase this PR, the commits in this PR are contained in further topic branches. :warning: 

This test is particularly sensitive to changes in the default configuration, by counting the operations it should be more stable.

https://issues.jboss.org/browse/WFCORE-3974


jamezp:
@darranl I don't think it would have to do with this PR, but there are the same tests failing on both full runs.

PR #3403, opened 2018-07-27
[WFLY-10091] Check Java version before setting up GC logging

rstancel:
Jira: ~https://issues.jboss.org/browse/WFLY-10091~ 
https://issues.jboss.org/browse/WFCORE-3996

### DIFF
--- a/core-feature-pack/src/main/resources/content/bin/common.ps1
+++ b/core-feature-pack/src/main/resources/content/bin/common.ps1
@@ -70,6 +70,16 @@ if (-Not $javaOpts){
 	$javaOpts = Get-Java-Opts
 }
 
+Function Check-For-Java9+ {
+	& $JAVA --add-modules java.se -version >$null 2>&1
+
+	if ($LastExitCode -eq 0){
+		$JAVA_9_PLUS = $true
+	}
+
+	return $JAVA_9_PLUS
+}
+
 # Display our environment
 Write-Host "================================================================================="
 Write-Host ""
@@ -141,14 +151,22 @@ Param(
 		if (-not(Test-Path $JBOSS_LOG_DIR)) {
 			$dir = New-Item $JBOSS_LOG_DIR -type directory -ErrorAction SilentlyContinue
 		}
-        $PROG_ARGS += "-verbose:gc"
-        $PROG_ARGS += "-XX:+PrintGCDetails"
-        $PROG_ARGS += "-XX:+PrintGCDateStamps"
-        $PROG_ARGS += "-XX:+UseGCLogFileRotation"
-        $PROG_ARGS += "-XX:NumberOfGCLogFiles=5"
-        $PROG_ARGS += "-XX:GCLogFileSize=3M"
-        $PROG_ARGS += "-XX:-TraceClassUnloading"
-        $PROG_ARGS += "-Xloggc:$JBOSS_LOG_DIR\gc.log"
+
+		$JAVA_9_PLUS = Check-For-Java9+
+
+		$PROG_ARGS += "-verbose:gc"
+		if ($JAVA_9_PLUS -eq $true)
+		{
+			$PROG_ARGS += "-Xlog:gc*:file=%JBOSS_LOG_DIR%\gc.log:time,uptimemillis:filecount=5,filesize=3M"
+		} else {
+			$PROG_ARGS += "-XX:+PrintGCDetails"
+			$PROG_ARGS += "-XX:+PrintGCDateStamps"
+			$PROG_ARGS += "-XX:+UseGCLogFileRotation"
+			$PROG_ARGS += "-XX:NumberOfGCLogFiles=5"
+			$PROG_ARGS += "-XX:GCLogFileSize=3M"
+			$PROG_ARGS += "-XX:-TraceClassUnloading"
+			$PROG_ARGS += "-Xloggc:$JBOSS_LOG_DIR\gc.log"
+		}
     }
   }
   $global:FINAL_JAVA_OPTS = $PROG_ARGS
@@ -303,7 +321,16 @@ Function Rotate-GC-Logs {
 
 Function Check-For-GC-Log {
 	if (GC_LOG){
-		$args = (,'-verbose:gc',"-Xloggc:$JBOSS_LOG_DIR/gc.log","-XX:+PrintGCDetails","-XX:+PrintGCDateStamps","-XX:+UseGCLogFileRotation","-XX:NumberOfGCLogFiles=5","-XX:GCLogFileSize=3M","-XX:-TraceClassUnloading",'-version')
+		$JAVA_9_PLUS = Check-For-Java9+
+
+		if ($JAVA_9_PLUS -eq $true)
+		{
+			$args = (,'-verbose:gc',"-Xlog:gc*:file=%JBOSS_LOG_DIR%\gc.log:time,uptimemillis:filecount=5,filesize=3M",'-version')
+		}
+		else
+		{
+			$args = (,'-verbose:gc',"-Xloggc:$JBOSS_LOG_DIR/gc.log","-XX:+PrintGCDetails","-XX:+PrintGCDateStamps","-XX:+UseGCLogFileRotation","-XX:NumberOfGCLogFiles=5","-XX:GCLogFileSize=3M","-XX:-TraceClassUnloading",'-version')
+		}
 		$OutputVariable = (&$JAVA $args )  | Out-String
 	}
 }
--- a/core-feature-pack/src/main/resources/content/bin/standalone.bat
+++ b/core-feature-pack/src/main/resources/content/bin/standalone.bat
@@ -15,6 +15,7 @@ set DEBUG_MODE=false
 set DEBUG_PORT_VAR=8787
 rem Set to all parameters by default
 set "SERVER_OPTS=%*"
+set JAVA_9_PLUS=
 
 
 if NOT "x%DEBUG%" == "x" (
@@ -142,6 +143,8 @@ if "x%JAVA_HOME%" == "x" (
   )
 )
 
+"%JAVA%" --add-modules java.se --version >nul 2>&1 || set JAVA_9_PLUS=no
+
 if not "%PRESERVE_JAVA_OPTS%" == "true" (
   rem Add -client to the JVM options, if supported (32 bit VM), and not overriden
   echo "%JAVA_OPTS%" | findstr /I \-server > nul
@@ -233,14 +236,26 @@ if not "%PRESERVE_JAVA_OPTS%" == "true" (
         move /y "%JBOSS_LOG_DIR%\gc.log.4" "%JBOSS_LOG_DIR%\backupgc.log.4" > nul 2>&1
         move /y "%JBOSS_LOG_DIR%\gc.log.*.current" "%JBOSS_LOG_DIR%\backupgc.log.current" > nul 2>&1
 
-        "%JAVA%" -verbose:gc -XX:+PrintGCDetails -XX:+PrintGCDateStamps -XX:+UseGCLogFileRotation -XX:NumberOfGCLogFiles=5 -XX:GCLogFileSize=3M -Xloggc:"%JBOSS_LOG_DIR%\gc.log" -XX:-TraceClassUnloading -version > nul 2>&1
+        setlocal EnableDelayedExpansion
+        if "x!JAVA_9_PLUS!" == "x" (
+            "%JAVA%" -verbose:gc -Xlog:gc*:file="%JBOSS_LOG_DIR%\gc.log":time,uptimemillis:filecount=5,filesize=3M -version > nul 2>&1
+        ) else (
+            "%JAVA%" -verbose:gc -XX:+PrintGCDetails -XX:+PrintGCDateStamps -XX:+UseGCLogFileRotation -XX:NumberOfGCLogFiles=5 -XX:GCLogFileSize=3M -Xloggc:"%JBOSS_LOG_DIR%\gc.log" -XX:-TraceClassUnloading -version > nul 2>&1
+        )
+
         if not errorlevel == 1 (
           if not exist "%JBOSS_LOG_DIR" > nul 2>&1 (
             mkdir "%JBOSS_LOG_DIR%"
           )
-		set JAVA_OPTS=%JAVA_OPTS% -verbose:gc -XX:+PrintGCDetails -XX:+PrintGCDateStamps -XX:+UseGCLogFileRotation -XX:NumberOfGCLogFiles=5 -Xloggc:"%JBOSS_LOG_DIR%\gc.log" -XX:GCLogFileSize=3M -XX:-TraceClassUnloading
+            if "x!JAVA_9_PLUS!" == "x" (
+                set JAVA_OPTS=%JAVA_OPTS% -verbose:gc -Xlog:gc*:file="%JBOSS_LOG_DIR%\gc.log":time,uptimemillis:filecount=5,filesize=3M
+            } else {
+                set JAVA_OPTS=%JAVA_OPTS% -verbose:gc -XX:+PrintGCDetails -XX:+PrintGCDateStamps -XX:+UseGCLogFileRotation -XX:NumberOfGCLogFiles=5 -Xloggc:"%JBOSS_LOG_DIR%\gc.log" -XX:GCLogFileSize=3M -XX:-TraceClassUnloading
+            )
         )
        )
+
+       setlocal DisableDelayedExpansion
     )
 )
 
--- a/core-feature-pack/src/main/resources/content/bin/standalone.sh
+++ b/core-feature-pack/src/main/resources/content/bin/standalone.sh
@@ -131,6 +131,8 @@ if [ "x$JAVA" = "x" ]; then
     fi
 fi
 
+$JAVA --add-modules java.se -version > /dev/null || JAVA_9_PLUS=no
+
 if $linux; then
     # consolidate the server and command line opts
     CONSOLIDATED_OPTS="$JAVA_OPTS $SERVER_OPTS"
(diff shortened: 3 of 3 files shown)


marekkopecky:
I don't see any regression on JDK8.

============================================

Once I start standalone with JDK10 with default settings and `export GC_LOG="true"`, standalone prints to console logs this:
```
[0.007s][info][gc] Using G1
....
[1.089s][info][gc] GC(1) Pause Young (G1 Evacuation Pause) 34M->8M(64M) 7.700ms
[1.199s][info][gc] GC(2) Pause Young (G1 Evacuation Pause) 36M->11M(64M) 5.351ms
[1.249s][info][gc] GC(3) Pause Young (G1 Evacuation Pause) 41M->14M(64M) 2.678ms
[1.288s][info][gc] GC(4) Pause Young (GCLocker Initiated GC) 42M->16M(64M) 4.192ms
[1.346s][info][gc] GC(5) Pause Young (G1 Evacuation Pause) 44M->18M(64M) 3.945ms
[1.545s][info][gc] GC(6) Pause Young (G1 Evacuation Pause) 46M->19M(64M) 5.240ms
[1.717s][info][gc] GC(7) Pause Young (G1 Evacuation Pause) 46M->21M(77M) 5.348ms
[1.753s][info][gc] GC(8) Pause Young (G1 Evacuation Pause) 54M->23M(77M) 5.560ms
```
this is change of behaviour against JDK8, do we really want this?

===========================================

I test this with JDK10.
* I start standalone and stop standalone
* JDK10 creates JBOSS_HOME/standalone/log/gc.log (with detailed information) and JBOSS_HOME/standalone/log/gc.log.0 (with brief summary of gc logging)
* I start standalone and stop standalone again
* JDK10 moves original gc.log to gc.log.1
* JDK10 keep original gc.log.0
* JDK10 creates new gc.log (with detailed information) and gc.log.2 (with brief summary of gc logging)
* I start standalone and stop standalone again
* JDK10 keep original gc.log.0, gc.log.1, gc.log.2
* JDK10 moves original gc.log to gc.log.3
* JDK10 creates new gc.log (with detailed information) and gc.log.4 (with brief summary of gc logging)

I'm not sure, whether this behaviour is intuitive ...  can this behaviour be changed by some JDK settings?

Anyway, current [mv commands](https://github.com/wildfly/wildfly-core/blob/master/core-feature-pack/src/main/resources/content/bin/standalone.sh#L268-L273) may be revisited for JDK10, maybe removed.

marekkopecky:
https://github.com/chewiebug/GCViewer (build from develop branch) is unable to parse gc.log, gc.log.4 or both files together (from previous example). This is regression of behaviour against JDK8 gc logs, maybe caused by missing support on GCViewer side or by wrong WF-Core settings

rsvoboda:
@bmaxwell Seems that GC logging needs redesign, it's unusable in production on Java 9+ even with this PR
@msve @kabir fyi

marekkopecky:
Current rotation behaviour:
* JDK10 creates gc.log (detailed information), and gc.log.1 (brief information).
* Once gc.log needs to be larged that 3MB, current gc.log is archived to gc.log.2 and gc.log contains latest gc information.
* Once there are more gc.log.1,2,3 and 4, gc.log.1 doesn't contains brief information anymore, but it contains detailed backuped gc information. In my PoV, this is not user-friendly too.

JDK10 gc logs are much more verbose against jdk8 gc logs (regression of behaviour against JDK8 too).

rstancel:
GCViewer currently support SOME JDK9+ loggging options. So not all of them.

ropalka:
Looks good @jmesnil & @kabir .

bstansberry:
Can this JIRA be moved to WFCORE please and any commit messages and PR titles use the WFCORE JIRA? Thanks.

I realize this PR may be reworked.

rsvoboda:
Moved - https://issues.jboss.org/browse/WFCORE-3996

core-feature-pack/src/main/resources/content/bin/common.ps1
```diff
@@ -141,14 +151,22 @@ Param(
...
-        $PROG_ARGS += "-XX:GCLogFileSize=3M"
-        $PROG_ARGS += "-XX:-TraceClassUnloading"
-        $PROG_ARGS += "-Xloggc:$JBOSS_LOG_DIR\gc.log"
+
+		$JAVA_9_PLUS = Check-For-Java9+
+
+		$PROG_ARGS += "-verbose:gc"
+		if ($JAVA_9_PLUS -eq $true)
+		{
+			$PROG_ARGS += "-Xlog:gc*:file=%JBOSS_LOG_DIR%\gc.log:time,uptimemillis:filecount=5,filesize=3M"
```
kabir: Bumping the filesize to a larger value here (perhaps also exposed as an environment variable with a default) should probably get rid of @marekkopecky 's concern about the rotation?

kabir:
@marekkopecky I am not sure exactly what you are asking from @rstancel? I am just jumping in and am by no means a JDK 9 expert but I am good at googling ;) I believe you have two problems:

1) Too aggressive log rotation
Will my suggestion at https://github.com/wildfly/wildfly-core/pull/3403#pullrequestreview-144518989 help?

2) Log file format
A quick search comes up with https://dzone.com/articles/disruptive-changes-to-gc-logging-in-java-9 which says 'So, if you are using any custom scripts/tools to analyze GC logs, those scripts/tools will need to be enhanced to support this new format.'

Do we support GCViewer? What is the significance of not being able to use that? The latest release of GCViewer seems to be over a year old with little activity leading up to the release https://github.com/chewiebug/GCViewer/commits/1.35 - I believe this date was before JDK 9 was even released. There is more activity on the https://github.com/chewiebug/GCViewer/commits/develop branch, but I don't know if that is JDK 9 related or not.

CC @ropalka @jmesnil

rsvoboda:
pasting email content from Marek 2 days ago, you are in that thread guys :)

Sure,
top issues, based on priority/importance in QE PoV
 1) JDK10 with settings from PR3403 [1] prints all GC logs to console, see first part of my comment [2]. This is regression against JDK8. This IO traffic can slowdown performance and can confuse console output.
 2) weird naming of gc log files, can this be changed by GC settings? See the second part of [2]
 3) mv commands needs to be revisited/fixed, see the third part of [2]
 4) weird rotation, see first part of [5], related with 3.
 5) JDK10 gc logs are much more verbose against jdk8 gc logs, can this be fixed by some JDK10 gc settings? See the second part of [5].
 6) wrong gc-viewer behaviour, see [4]

I agree with Rostislav, wrong gc-viewer behaviour has the loweset priority in PR3403 issues [1] based on previous list

Marek
[1] https://github.com/wildfly/wildfly-core/pull/3403
[2] https://github.com/wildfly/wildfly-core/pull/3403#issuecomment-409109169
[3] https://github.com/wildfly/wildfly-core/blob/master/core-feature-pack/src/main/resources/content/bin/standalone.sh#L268-L273
[4] https://github.com/wildfly/wildfly-core/pull/3403#issuecomment-409111061
[5] https://github.com/wildfly/wildfly-core/pull/3403#issuecomment-409122973

core-feature-pack/src/main/resources/content/bin/common.ps1
marekkopecky: JDK8 use log rotation, I believe that log rotation is useful feature that can be useful also on JDK11. I believe that unlimited size could cause problems to WF users ... anyway, it would be regression of behaviour as well ...

rsvoboda:
@jamezp "logging" is your area, can you please look into this ? Especially log rotation point, also GC log items printed to console. A bit weird are those 2 log files, one with detailed information and one with brief summary.

core-feature-pack/src/main/resources/content/bin/common.ps1
```diff
@@ -70,6 +70,16 @@ if (-Not $javaOpts){
...
 
+Function Check-For-Java9+ {
+	& $JAVA --add-modules java.se -version >$null 2>&1
+
+	if ($LastExitCode -eq 0){
+		$JAVA_9_PLUS = $true
+	}
+
+	return $JAVA_9_PLUS
+}
```
jamezp: This shouldn't be tab delimited :)

jamezp:
I noticed in a few places we're executing `java $some-args -version`. What does everyone thing of an approach more like this https://github.com/wildfly/wildfly-core/compare/master...jamezp:WFCORE-3996.

The idea is to add a simple entry point to the launcher.jar that will return just the required GC and optional `--add-modules java.se` if required. I'm not too sure what the performance repercussions would be, but my guess is it wouldn't be all that bad.

I've test on Fedora and Windows with Java 9 only. On Fedora it worked on Java 8, 9, 10, 11 and a build of 12. I didn't test the ps1 scripts as I didn't change those yet.

Please note this was just a quick work up. We could use the same thing in other scripts to help with unifying things like `-secmgr` as well.

jamezp:
I should note the output when GC is enabled to the console is still a bit goofy, but we can take a deeper look at that if this approach seems okay.
```
jperkins@localhost ~/projects/jboss/wildfly/wildfly-core/build (WFCORE-3996) $ ./target/wildfly-core-6.0.0.Alpha6-SNAPSHOT/bin/standalone.sh 
=========================================================================

  JBoss Bootstrap Environment

  JBOSS_HOME: /home/jperkins/projects/jboss/wildfly/wildfly-core/build/target/wildfly-core-6.0.0.Alpha6-SNAPSHOT

  JAVA: /usr/lib/jvm/java-10-openjdk-10.0.2.13-1.fc28.x86_64/bin/java

  JAVA_OPTS:  -server -verbose:gc -Xlog:gc:file=/home/jperkins/projects/jboss/wildfly/wildfly-core/build/./gc.log:time,uptimemillis:filecount=5,filesize=3M -Xms64m -Xmx512m -XX:MetaspaceSize=96M -XX:MaxMetaspaceSize=256m -Djava.net.preferIPv4Stack=true -Djboss.modules.system.pkgs=com.yourkit,org.jboss.byteman -Djava.awt.headless=true

=========================================================================

[0.003s][info][gc] Using G1
11:45:10,926 INFO  [org.jboss.modules] (main) JBoss Modules version 1.8.5.Final
[0.524s][info][gc] GC(0) Pause Young (G1 Evacuation Pause) 24M->5M(64M) 5.054ms
11:45:11,212 INFO  [org.jboss.msc] (main) JBoss MSC version 1.4.2.Final
11:45:11,223 INFO  [org.jboss.threads] (main) JBoss Threads version 2.3.2.Final
11:45:11,336 INFO  [org.jboss.as] (MSC service thread 1-1) WFLYSRV0049: WildFly Core 6.0.0.Alpha6-SNAPSHOT starting
[0.744s][info][gc] GC(1) Pause Young (G1 Evacuation Pause) 15M->6M(64M) 4.099ms
[1.080s][info][gc] GC(2) Pause Young (G1 Evacuation Pause) 28M->8M(64M) 4.719ms
11:45:11,778 INFO  [org.wildfly.security] (ServerService Thread Pool -- 6) ELY00001: WildFly Elytron version 1.5.1.Final
11:45:11,900 INFO  [org.jboss.as.controller.management-deprecated] (Controller Boot Thread) WFLYCTL0028: Attribute 'security-realm' in the resource at address '/core-service=management/management-interface=http-interface' is deprecated, and may be removed in a future version. See the attribute description in the output of the read-resource-description operation to learn more about the deprecation.
11:45:11,943 INFO  [org.jboss.as.server] (Controller Boot Thread) WFLYSRV0039: Creating http management service using socket-binding (management-http)
11:45:11,969 INFO  [org.xnio] (MSC service thread 1-1) XNIO version 3.6.4.Final
11:45:11,977 INFO  [org.xnio.nio] (MSC service thread 1-1) XNIO NIO Implementation Version 3.6.4.Final
11:45:12,028 INFO  [org.jboss.remoting] (MSC service thread 1-7) JBoss Remoting version 5.0.8.Final
[1.466s][info][gc] GC(3) Pause Young (G1 Evacuation Pause) 34M->11M(64M) 11.121ms
11:45:12,084 INFO  [org.jboss.as.patching] (MSC service thread 1-4) WFLYPAT0050: WildFly cumulative patch ID is: base, one-off patches include: none
11:45:12,103 WARN  [org.jboss.as.domain.management.security] (MSC service thread 1-6) WFLYDM0111: Keystore /home/jperkins/projects/jboss/wildfly/wildfly-core/build/target/wildfly-core-6.0.0.Alpha6-SNAPSHOT/standalone/configuration/application.keystore not found, it will be auto generated on first use with a self signed certificate for host localhost
11:45:12,160 WARN  [org.jboss.as.domain.http.api.undertow] (MSC service thread 1-2) WFLYDMHTTP0003: Unable to load console module for slot main, disabling console
11:45:12,232 INFO  [org.jboss.as.server] (Controller Boot Thread) WFLYSRV0212: Resuming server
11:45:12,234 INFO  [org.jboss.as] (Controller Boot Thread) WFLYSRV0060: Http management interface listening on http://127.0.0.1:9990/management
11:45:12,234 INFO  [org.jboss.as] (Controller Boot Thread) WFLYSRV0051: Admin console listening on http://127.0.0.1:9990
11:45:12,235 INFO  [org.jboss.as] (Controller Boot Thread) WFLYSRV0025: WildFly Core 6.0.0.Alpha6-SNAPSHOT started in 1558ms - Started 85 of 88 services (23 services are lazy, passive or on-demand)
```

rsvoboda:
I think the main concern is WHAT to add and not really HOW to add.

Launcher idea was discussed on the mailing list, https://github.com/wildfly/wildfly-core/compare/master...jamezp:WFCORE-3996 shows a bit different approach where the launcher acts as a helper. I have no strong feelings about the helper approach, but we first need to know WHAT should be added.

jamezp:
That is a good point @rsvoboda. The what is definitely important. I'm honestly not sure what the original requirements were, but the [article](https://dzone.com/articles/disruptive-changes-to-gc-logging-in-java-9) @kabir posted does seem to have a mapping between the settings so maybe we can use that to help a bit.

The _only_ real thing I like about the launcher is it keeps the logic in one place so don't have to keep touching a minimum of 6 scripts if we need to change anything.

jamezp:
Another possible option I've thought of is we could move the settings, not the rotation logic just the VM arguments, into the standalone.conf. Then it would just need to be uncommented based on the Java version they're using.

jamezp:
I think this is what we want for the `JAVA_OPTS` on Java 9+
```
-Xlog:gc*:file=$JBOSS_LOG_DIR/gc.log:time,uptimemillis:filecount=5,filesize=3M
```

We need to remove the `-verbose:gc` which will stop the stderr output. We could also do something like:
```
-Xlog:gc*=$GC_LOG_LEVEL:file=$JBOSS_LOG_DIR/gc.log:time,uptimemillis:filecount=5,filesize=3M
```

which would default `GC_LOG_LEVEL` to `info` so if we need `debug` or `trace` it could be overridden without requiring a script change.

core-feature-pack/src/main/resources/content/bin/standalone.bat
```diff
@@ -233,14 +236,26 @@ if not "%PRESERVE_JAVA_OPTS%" == "true" (
...
-		set JAVA_OPTS=%JAVA_OPTS% -verbose:gc -XX:+PrintGCDetails -XX:+PrintGCDateStamps -XX:+UseGCLogFileRotation -XX:NumberOfGCLogFiles=5 -Xloggc:"%JBOSS_LOG_DIR%\gc.log" -XX:GCLogFileSize=3M -XX:-TraceClassUnloading
+            if "x!JAVA_9_PLUS!" == "x" (
+                set JAVA_OPTS=%JAVA_OPTS% -verbose:gc -Xlog:gc*:file="%JBOSS_LOG_DIR%\gc.log":time,uptimemillis:filecount=5,filesize=3M
+            } else {
+                set JAVA_OPTS=%JAVA_OPTS% -verbose:gc -XX:+PrintGCDetails -XX:+PrintGCDateStamps -XX:+UseGCLogFileRotation -XX:NumberOfGCLogFiles=5 -Xloggc:"%JBOSS_LOG_DIR%\gc.log" -XX:GCLogFileSize=3M -XX:-TraceClassUnloading
+            )
         )
        )
+
+       setlocal DisableDelayedExpansion
```
jamezp: This didn't work for me on Windows 7. I had to change it to something like
```
        if "x%JAVA_9_PLUS%" == "x" (
            set JAVA_OPTS=%JAVA_OPTS% -Xlog:gc*:file=\"%JBOSS_LOG_DIR%\gc.log\":time,uptimemillis:filecount=5,filesize=3M
        ) else (
            "%JAVA%" -verbose:gc -XX:+PrintGCDetails -XX:+PrintGCDateStamps -XX:+UseGCLogFileRotation -XX:NumberOfGCLogFiles=5 -XX:GCLogFileSize=3M -Xloggc:"%JBOSS_LOG_DIR%\gc.log" -XX:-TraceClassUnloading -version > nul 2>&1
            if not errorlevel == 1 (
                set JAVA_OPTS=%JAVA_OPTS% -verbose:gc -XX:+PrintGCDetails -XX:+PrintGCDateStamps -XX:+UseGCLogFileRotation -XX:NumberOfGCLogFiles=5 -Xloggc:"%JBOSS_LOG_DIR%\gc.log" -XX:GCLogFileSize=3M -XX:-TraceClassUnloading
            )
        )
       )
```

Note I'm not using delayed expansion and the quotes need to be escaped `\"`. Also there's no need to check that `-Xlog` is supported by the JVM.

rsvoboda:
* `-Xlog:gc*=$GC_LOG_LEVEL` is new feature, needs to be tracked in EAP7 + ranked

 * VM arguments in standalone.conf ... this file defines GC_LOG="true"/"false", this is the interface for the users, implementation is done in standalone.* scripts and not sure it should be exposed in .conf files

core-feature-pack/src/main/resources/content/bin/standalone.bat
rsvoboda: On EAP side we care only about Windows Server, but understand the concern for community users and developers. Still, Windows 7 is pretty old :)

jamezp: It's the only version of Windows I have to test on in a VM :) I do agree it's rather old at this point though. It was the escaping of the quote that didn't work for me.

jamezp:
We definitely don't need the `$GC_LOG_LEVEL` part. It was just more of an idea in case we were missing something in the logs since they've changed. Something to avoid users having to change the scripts :)

My main concern is the scripts are already complicated and detecting the Java version in the scripts is making them more complicated. We're going to end up having to do the same for all the domain scripts too if we need to add the `--add-modules=java.se` argument. Granted that isn't part of this PR however this PR is setting precedence and it's something we need to consider IMO.

jamezp:
Where are we at with this? I know the tabs need to be changed into spaces. The *.sh scripts need to redirect stdout/stderr so they don't show up on the console and I think the batch scripts the quoting needs to be fixed.

core-feature-pack/src/main/resources/content/bin/common.ps1
rstancel: @jamezp may I ask why? I see in the whole script a lot of places delimited with tab. When I take for example next function Get-Java-Opts it is delimited in the exactly same way

jamezp: Sorry yeah. I didn't realize the script was tab delimited itself. We must not have checkstyle checking these files. No change there required.

rstancel:
I have taken latest updates from James and Richard and opened new PR. I have added fixes also from your comments. New PR can be found here: https://github.com/wildfly/wildfly-core/pull/3496

@jamezp @marekkopecky @rsvoboda FYI

PR #3407, opened 2018-07-31
[WFCORE-3519] Add JASPI module to core-feature-pack.

darranl:
:warning: Do not Rebase this PR, the commits in this PR are contained in further topic branches. :warning: 

This is just a small sub-step as the WildFly Elytron subsystem will depend on this module in a similar way it already does for JACC, as this is a move from WildFly to WildFly Core it is better to split it out and get it included on it's own.

https://issues.jboss.org/browse/WFCORE-3519

### DIFF
--- a/core-feature-pack/pom.xml
+++ b/core-feature-pack/pom.xml
@@ -121,6 +121,11 @@
         </dependency>
 
         <dependency>
+            <groupId>org.jboss.spec.javax.security.auth.message</groupId>
+            <artifactId>jboss-jaspi-api_1.1_spec</artifactId>
+        </dependency>
+
+        <dependency>
             <groupId>org.jboss.classfilewriter</groupId>
             <artifactId>jboss-classfilewriter</artifactId>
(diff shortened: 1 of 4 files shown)


core-feature-pack/pom.xml
```diff
@@ -120,6 +120,11 @@
             </exclusions>
         </dependency>
 
+        <dependency>
+            <groupId>org.jboss.spec.javax.security.auth.message</groupId>
+            <artifactId>jboss-jaspi-api_1.1_spec</artifactId>
+        </dependency>
+
```
yersan: I understand this dependency should be added to core-galleon-pack/pom.xml as well. Galleon won't be able to include it in the module if it's not in that pom.xml file.

darranl: Added, it was in a later commit on the topic branch but does need to be included in this PR.

darranl:
Retest this please.

PR #3420, opened 2018-08-07
Fixing  testsuite issues on latest JDK11

ropalka:
This pull request addresses two issues:
 * https://issues.jboss.org/browse/WFCORE-4000 Domain tests fail on JDK11
 * https://issues.jboss.org/browse/WFCORE-4001 Allow illegal access in domain tests

### DIFF
--- a/host-controller/src/main/java/org/jboss/as/host/controller/model/jvm/JvmOptionsBuilderFactory.java
+++ b/host-controller/src/main/java/org/jboss/as/host/controller/model/jvm/JvmOptionsBuilderFactory.java
@@ -177,7 +177,7 @@ public class JvmOptionsBuilderFactory {
 
     private static class OracleJvmOptionsBuilder extends JvmOptionsBuilder {
         private static final int JVM_MAJOR_VERSION = JvmElement.getJVMMajorVersion();
-        private static final List<String> ALLOWED_JDK9_PARAMS = Arrays.asList("--permit-illegal-access", "--add-exports", "--add-opens", "--add-modules", "--add-reads", "--illegal-access");
+        private static final List<String> ALLOWED_MODULAR_JDK_PARAMS = Arrays.asList("--add-exports", "--add-opens", "--add-modules", "--add-reads", "--illegal-access");
 
         private OracleJvmOptionsBuilder(JvmType type) {
             super(type);
@@ -185,9 +185,9 @@ public class JvmOptionsBuilderFactory {
 
         @Override
         boolean checkAdditionalJvmOption(String option) {
-            if (JVM_MAJOR_VERSION >= 9 ) {//when java 11 comes, make sure this params are still fine.
+            if (JVM_MAJOR_VERSION >= 9 ) {
                 return true; //all params are fine
-            }else if (ALLOWED_JDK9_PARAMS.contains(option.contains("=") ? option.substring(0, option.indexOf("=")) : option)){ //drop jdk9 specific params on jdk 8
+            } else if (ALLOWED_MODULAR_JDK_PARAMS.contains(option.contains("=") ? option.substring(0, option.indexOf("=")) : option)){ //drop jdk9 specific params on jdk 8
                 return false;
             }
             return true;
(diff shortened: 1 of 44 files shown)


rsvoboda:
Shouldn't be following `<jvm-options>` handled in JvmOptionsBuilderFactory ?
```
                <option value="--illegal-access=permit"/>
                <option value="--add-modules=java.se"/>
```

 * will we change default configs to include these options ?
 * will EAP 7.1 configs work out of the box on both Java 8 and Java 11 ?

rsvoboda:
This PR is also https://issues.jboss.org/browse/WFCORE-3962 related

ropalka:
In my opinion @rsvoboda JvmOptionsBuilderFactory shouldn't automagically configure --illegal-access & --add-modules parameters. Everything should be explicit.
To answer your two questions:
Q: Will we change default configs to include these options ? YES
Q: Will EAP 7.1 configs work out of the box on both Java 8 and Java 11 ? YES (modular JDK options are ignored on non modular JDKs. That's what JvmOptionsBuilderFactory.checkAdditionalJvmOption() method does.

rsvoboda:
Thanks for the answers.
The second question is really about `EAP 7.1` where those jvm options are not present ... so the config won't be probably boot-able on JDK 11. We have requirement that customers can bring their older configs (e.g. from 7.1) and use them with the latest version.

>Q: Will we change default configs to include these options ? YES

Is there WFCORE jira for this ?

ropalka:
Shouldn't config migration tool be responsible for such forward config conversions @rsvoboda ?
CC-ing @bstansberry and @n1hility who may suggest how to deal with such usecases / requirements.

rsvoboda:
NO, migration tool is for EAP 6.4 to EAP 7.x migration

Config compatibility / transformers are the way to go across minor releases of EAP

bstansberry:
I disagree that we have a hard requirement that users can bring their WF 11/12/13 + JDK 8 config and apps to WF 14 and JDK 11 without modification.  JDK 11 is a far bigger change than any previous JDK change where we guaranteed that.

But, that statement is kind of a get out of jail free card for cases where it's too hard, convoluted etc to make things work.  I suspect this is such a case:

If these things are true, I don't see why we can't automagically set these options:

1) We know the JVM of the server being launched will not fail if the options are set.   The HC's JVM does not count!  The server can use a different one.  If we don't know for sure an automagic setting won't break things, we can't use one. 

2) We can reliable detect any conflict between explicit user settings and our automagic ones so we can defer to the explicit ones.  This seems doable.

3) We know the options are going to be valid for the life of the minor.  IOW we're not going to do something in some micro that makes them wrong and then we have to remember to fix the automagic.  This one I think is really unlikely.

I haven't looked hard but I think #1 is the key point.  IIRC in the past the JVM options magic we did was stripping out old options that we knew no supported JVM for the release would accept, e.g. permgen stuff. We weren't trying to insert new options that an old but still supported JVM would have to accept.

bstansberry:
I looked more and to do anything automagic we would have to:

1) Analyze the jvm element to see if it's set to use the same JVM as the HC. If yes then we could make decisions based on the HC's knowledge of its own VM.

2) If not, either there's no automagic or we'd have to launch the specified vm in order to learn about it (e.g. does it barf if --illegal-access etc are set) and then configure the server settings based on that knowledge.

host-controller/src/main/java/org/jboss/as/host/controller/model/jvm/JvmOptionsBuilderFactory.java
```diff
@@ -177,17 +177,17 @@ boolean checkAdditionalJvmOption(String option){
...
 
         @Override
         boolean checkAdditionalJvmOption(String option) {
-            if (JVM_MAJOR_VERSION >= 9 ) {//when java 11 comes, make sure this params are still fine.
+            if (JVM_MAJOR_VERSION >= 9 ) {
                 return true; //all params are fine
-            }else if (ALLOWED_JDK9_PARAMS.contains(option.contains("=") ? option.substring(0, option.indexOf("=")) : option)){ //drop jdk9 specific params on jdk 8
+            } else if (ALLOWED_MODULAR_JDK_PARAMS.contains(option.contains("=") ? option.substring(0, option.indexOf("=")) : option)){ //drop jdk9 specific params on jdk 8
                 return false;
             }
```
bstansberry: Thinking about it more, none of this from L188 is correct. Wasn't correct before either. It is using the JVM version of the HC to decide what's legal for the server, and in this case that is not a safe algorithm.

A JDK 11 HC can launch a JDK 8 server.

This kind of logic is only ok in cases where either we know the server is running the same VM version as the HC.  Otherwise we have to know what the VM the server is running is.

PR #3424, opened 2018-08-09
WFCORE-3873 Support for SNI for server side SSLContext

stuartwdouglas:
https://issues.jboss.org/browse/WFCORE-3873
https://issues.jboss.org/browse/EAP7-726

### DIFF
--- a/elytron/src/main/resources/org/wildfly/extension/elytron/LocalDescriptions.properties
+++ b/elytron/src/main/resources/org/wildfly/extension/elytron/LocalDescriptions.properties
@@ -1340,6 +1340,16 @@ elytron.certificate-authority-account.credential-reference.store=The name of the
 elytron.certificate-authority-account.credential-reference.alias=The alias which denotes stored secret or credential in the store.
 elytron.certificate-authority-account.credential-reference.type=The type of credential this reference is denoting.
 elytron.certificate-authority-account.credential-reference.clear-text=Secret specified using clear text. Check credential store way of supplying credential/secrets to services.
+elytron.server-ssl-sni-context=A server side SNI Aware SSLContext that selects between an underlying context based on the provided SNI name
+elytron.server-ssl-sni-context.add=Adds a SNI context
+elytron.server-ssl-sni-context.remove=Removes a SNI context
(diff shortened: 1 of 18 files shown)


elytron/src/main/resources/org/wildfly/extension/elytron/LocalDescriptions.properties
```diff
@@ -1340,6 +1340,16 @@ elytron.certificate-authority-account.credential-reference.store=The name of the
 elytron.certificate-authority-account.credential-reference.alias=The alias which denotes stored secret or credential in the store.
 elytron.certificate-authority-account.credential-reference.type=The type of credential this reference is denoting.
 elytron.certificate-authority-account.credential-reference.clear-text=Secret specified using clear text. Check credential store way of supplying credential/secrets to services.
+elytron.server-ssl-sni-context=A server side SNI Aware SSLContext that selects between an underlying context based on the provided SNI name
+elytron.server-ssl-sni-context.add=Adds a SNI context
+elytron.server-ssl-sni-context.remove=Removes a SNI context
+elytron.server-ssl-sni-context.default-ssl-context=The context to use if no SNI information is present, or if it does not match any mappings
+elytron.server-ssl-sni-context.sni-mapping=A mapping between a server name an an SSContext
+elytron.sni-mapping=A mapping between a server name an an SSContext
```
jstourac: the properties `elytron.sni-mapping` and `elytron.server-ssl-sni-context.sni-mapping` contain two typos each:
```
an -> and
SSContext -> SSLContext
```

jstourac:
@stuartwdouglas, @jmesnil, I can see that wildfly-core tag has been done without this merged. I guess that mean this will be postponed to CD15?

darranl:
If this is going to go to 15 I have a couple of comments to consider.

Within the Elytron subsystem at the outset we took a decision to avoid the use of child resources, there was extensive discussions but one of the main motivations was around where the CLI is used for configuration adding a parent followed by children commonly ends up with a prematurely ready parent with the server needing to be reloaded to consider the child resources properly.

Secondly we had discussed the SSLContext would be within the wildfly-elyton project, if we can get a pull request over there I can tag whenever, we are producing tags quite fast at the moment.

jmesnil:
closing it as it is superseded by #3424

PR #3426, opened 2018-08-09
[WFCORE-3995] forbid invalid Sensitivity Constraint attribute read-wr…

soul2zimate:
…ite combinations.

issues: https://issues.jboss.org/browse/WFCORE-3995 and https://issues.jboss.org/browse/WFCORE-4061

### DIFF
--- a/domain-management/src/main/java/org/jboss/as/domain/management/access/ApplicationClassificationConfigResourceDefinition.java
+++ b/domain-management/src/main/java/org/jboss/as/domain/management/access/ApplicationClassificationConfigResourceDefinition.java
@@ -99,14 +99,11 @@ public class ApplicationClassificationConfigResourceDefinition extends SimpleRes
             final String attribute = operation.require(NAME).asString();
             final ApplicationTypeConfigResource resource = (ApplicationTypeConfigResource)context.readResource(PathAddress.EMPTY_ADDRESS);
             final ApplicationTypeConfig applicationType = resource.applicationType;
-            Boolean result;
+            Boolean result = null;
             if (attribute.equals(DEFAULT_APPLICATION.getName())) {
                 result = applicationType.isDefaultApplication();
             } else if (attribute.equals(CONFIGURED_APPLICATION.getName())) {
                 result = applicationType.getConfiguredApplication();
-            } else {
-                //TODO i18n
-                throw new IllegalStateException();
             }
 
             context.getResult();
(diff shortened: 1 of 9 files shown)


bstansberry:
There's a couple problems in this area:

1) SensitivityClassificationWriteAttributeHandler does not have any rollback handling.  That's not a problem in this PR; it's just this PR made me realize it.  I suspect other handlers like this may have the same issue.  I filed https://issues.jboss.org/browse/WFCORE-4061

2) Cross-attribute validation like what's done here must be done in an added step, as that allows composite ops to work without undue concern about order of steps in the composite. Without that this might break existing scripts.  (It would be ok to change the AbstractSensitivity and then add the validation step, counting on the rollback to correct the AbstractSensitivity if the validation fails. But WFCORE-4061 would have to be corrected for that to be a correct approach.)

soul2zimate:
I have rebased to

- add rollback handling first for WFCORE-4061
- remove some redundant code because attribute name should have been validated before execution.
- validate attributes for WFCORE-3995

soul2zimate:
Hi @bstansberry can this be merged ?

soul2zimate:
retest this please

domain-management/src/main/java/org/jboss/as/domain/management/access/ApplicationClassificationConfigResourceDefinition.java
```diff
@@ -121,17 +119,29 @@ public void execute(OperationContext context, ModelNode operation) throws Operat
 
         @Override
         public void execute(OperationContext context, ModelNode operation) throws OperationFailedException {
-            final String attribute = operation.require(NAME).asString();
+            PathAddress address = PathAddress.pathAddress(operation.get(OP_ADDR));
```
bstansberry: If you need the address of the current op, use 'context.getCurrentAddress()'.

I don't think you need it though; see my next comment.

domain-management/src/main/java/org/jboss/as/domain/management/access/ApplicationClassificationConfigResourceDefinition.java
```diff
@@ -121,17 +119,29 @@ public void execute(OperationContext context, ModelNode operation) throws Operat
...
+            ModelNode modelNode = context.readResourceFromRoot(address).getModel();
+            // record model values for rollback handler
+            ModelNode configuredApplication = modelNode.get(ModelDescriptionConstants.CONFIGURED_APPLICATION);
```
bstansberry: These lines can come after L128 and the first line becomes

ModelNode modelNode = resource.getModel();

The Resource from context.readResourceForUpdate includes the data that you need to restore.

soul2zimate:
Thanks, I have updated this one.

PR #3430, opened 2018-08-09
[WFCORE-3991] read-config-as-features management operation

aloubyansky:
https://issues.jboss.org/browse/WFCORE-3991

This is a read-only and hidden operation which reads the management model and represents it in terms of provisioning features. This operation is used by Galleon provisioning plugins to generate configuration diffs.

### DIFF
--- a/controller/src/main/java/org/jboss/as/controller/operations/PathAddressFilter.java
+++ b/controller/src/main/java/org/jboss/as/controller/operations/PathAddressFilter.java
@@ -20,7 +20,7 @@
  * 02110-1301 USA, or see the FSF site: http://www.fsf.org.
  */
 
-package org.jboss.as.domain.controller.operations;
+package org.jboss.as.controller.operations;
 
 import java.util.HashMap;
 import java.util.Iterator;
@@ -33,17 +33,17 @@ import org.jboss.as.controller.PathElement;
 /**
  * @author Emanuel Muckenhuber
  */
-class PathAddressFilter {
+public final class PathAddressFilter {
 
-    static final OperationContext.AttachmentKey<PathAddressFilter> KEY = OperationContext.AttachmentKey.create(PathAddressFilter.class);
+    public static final OperationContext.AttachmentKey<PathAddressFilter> KEY = OperationContext.AttachmentKey.create(PathAddressFilter.class);
 
     private final boolean accept;
     private final Node node = new Node(null);
-    protected PathAddressFilter(boolean accept) {
+    public PathAddressFilter(boolean accept) {
         this.accept = accept;
     }
 
-    boolean accepts(PathAddress address) {
+    public boolean accepts(PathAddress address) {
         final Iterator<PathElement> i = address.iterator();
         Node node = this.node;
         while (i.hasNext()) {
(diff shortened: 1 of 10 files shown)


controller/src/main/java/org/jboss/as/controller/operations/PathAddressFilter.java
```diff
@@ -33,17 +33,17 @@
 /**
  * @author Emanuel Muckenhuber
  */
-class PathAddressFilter {
+public class PathAddressFilter {
```
bstansberry: If this has to be public it must be final now. (Unless it already has subclasses.)

controller/src/main/java/org/jboss/as/controller/operations/global/ReadConfigAsFeaturesOperationHandler.java
```diff
@@ -0,0 +1,382 @@
...
+/**
+ *
+ * @author Alexey Loubyansky
+ */
+public class ReadConfigAsFeaturesOperationHandler implements OperationStepHandler {
+
+    private static final String ID = "id";
+
+    public static final SimpleOperationDefinition DEFINITION = new SimpleOperationDefinitionBuilder(READ_CONFIG_AS_FEATURES_OPERATION, ControllerResolver.getResolver(SUBSYSTEM))
+            .addAccessConstraint(SensitiveTargetAccessConstraintDefinition.READ_WHOLE_CONFIG)
```
bstansberry: Nice; you already did this!

controller/src/main/java/org/jboss/as/controller/operations/global/ReadConfigAsFeaturesOperationHandler.java
```diff
@@ -0,0 +1,382 @@
...
+
+    private void describeChildren(final Resource resource, final ImmutableManagementResourceRegistration registration,
+            final PathAddress address, OperationContext context,
+            final AtomicReference<ModelNode> failureRef, final ModelNode results, ModelNode operation, Boolean nest) {
+        resource.getChildTypes().forEach((childType) -> {
+            resource.getChildren(childType).stream()
+                    .filter(entry -> {
+                        final ImmutableManagementResourceRegistration childRegistration = registration.getSubModel(PathAddress.EMPTY_ADDRESS.append(entry.getPathElement()));
+                        if(childRegistration == null) {
+                            ControllerLogger.ROOT_LOGGER.warnf("Couldn't find a registration for %s at %s for resource %s at %s", entry.getPathElement().toString(), registration.getPathAddress().toCLIStyleString(), resource, address.toCLIStyleString());
```
bstansberry: At WARN this requires a method in ControllerLogger for i18n. But this is so unlikely that I think debug is fine. If this were to happen it's almost certain the child would be runtime and thus not relevant anyway. A resource that's an actual feature wouldn't work without a registration.

controller/src/main/java/org/jboss/as/controller/operations/global/ReadConfigAsFeaturesOperationHandler.java
```diff
@@ -0,0 +1,382 @@
...
+                            return false;
+                        }
+                        return !childRegistration.isRuntimeOnly() && !childRegistration.isRemote() && !childRegistration.isAlias();
+                    })
+                    .filter(entry ->  addressFilter == null || addressFilter.accepts(address.append(entry.getPathElement())))
+                    .forEach((entry) -> describeChildResource(entry, registration, address, context, failureRef, results, operation, nest));
```
bstansberry: Our general coding standard is to not use streams to replace for loops as the lambdas add a lot to our memory footprint compared to normal loop content.

PR #3434, opened 2018-08-09
[WFCORE-3386] Add JASPI Configuration to the Elytron Subsystem

darranl:
https://issues.jboss.org/browse/WFCORE-3386

This PR will also be dependent on component upgrades including the following fixes: -

https://github.com/wildfly-security/wildfly-elytron/pull/1170
https://github.com/wildfly-security/elytron-web/pull/137

Other than the component upgrades the only planned change on this branch is to add some exclusivity checks on the layer / application-context attributes.


core-galleon-pack/pom.xml
```diff
@@ -128,6 +128,11 @@
             </exclusions>
         </dependency>
 
+        <dependency>
+            <groupId>org.jboss.spec.javax.security.auth.message</groupId>
+            <artifactId>jboss-jaspi-api_1.1_spec</artifactId>
```
kabir: @darranl This also needs adding to the core-feature-pack/pom.xml

elytron/src/main/resources/schema/wildfly-elytron_4_0.xsd
```diff
@@ -5178,4 +5179,118 @@
...
+        <xs:sequence>
+            <xs:element name="options" minOccurs="0">
+                <xs:annotation>
+                    <xs:documentation>
+                        Configuration options to be passed into the ServerAuthModule during initialisation.
+                    </xs:documentation>
+                </xs:annotation>
+                <xs:complexType>
+                    <xs:sequence>
+                        <xs:element name="property" type="propertyType" maxOccurs="unbounded" />
```
kabir: @darranl In the model this looks like it is called 'options', and since you are using PersistentResourceXMLDescription I think that is what the parser expects?

kabir:
@darranl Looks ok once you rebase and add it to the galleon pack

core-galleon-pack/pom.xml
darranl: Hi @kabir it is already there in the preparatory work https://github.com/wildfly/wildfly-core/blob/master/core-feature-pack/pom.xml#L123

darranl: The core-galleon-pack one is also there, once I synchronise with upstream this should drop from this PR https://github.com/wildfly/wildfly-core/blob/master/core-galleon-pack/pom.xml#L115

OndrejKotek:
@kabir @jmesnil @darranl What is expected time frame? Do we plan to force this to WF 14, or will it be posponed? Any special treatment expected? Note that the RFE is rather base/requirement for the EAP7-993 which is [postponed to WF 15](https://issues.jboss.org/browse/EAP7-993?focusedCommentId=13618628&page=com.atlassian.jira.plugin.system.issuetabpanels:comment-tabpanel#comment-13618628).

Could you comment on [EAP7-649](https://issues.jboss.org/browse/EAP7-649?focusedCommentId=13618652&page=com.atlassian.jira.plugin.system.issuetabpanels:comment-tabpanel#comment-13618652) please?

darranl:
Looks like GitHub is messing up some diffs - this is still going to require a Wildfly Elytron component upgrade but is now based on version 5 of the WildFly Elytron subsystem model and schema.

jmesnil:
Requirements are met to be able to review (and eventually merge) this PR.

darranl:
@jmesnil FYI we have clean CI runs on this one now after all merge conflicts were addressed.

elytron/src/main/java/org/wildfly/extension/elytron/ElytronSubsystemParser5_0.java
```diff
@@ -28,11 +37,42 @@
  */
 public class ElytronSubsystemParser5_0 extends ElytronSubsystemParser4_0 {
 
+    final PersistentResourceXMLDescription jaspiConfigurationParser = PersistentResourceXMLDescription.builder(PathElement.pathElement(JASPI_CONFIGURATION))
+            .setXmlWrapperElement(JASPI)
+            .addAttributes(JaspiDefinition.ATTRIBUTES)
```
jmesnil: This is not holding the PR but we should discourage referencing attributes collections when defining a XML parser.

If another attribute is added to the `JaspiDefinition.ATTRIBUTES` in a later version, it would be picked up by the XML 5.0 when it must not.

Most other Elytron's `PersistentResourceXMLDescription` explicitly enumerates the attribute instead of relying on their `ATTRIBUTES` collection.

elytron/src/main/java/org/wildfly/extension/elytron/JaspiDefinition.java
```diff
@@ -0,0 +1,210 @@
...
+            .build();
+
+    static final PropertiesAttributeDefinition OPTIONS = new PropertiesAttributeDefinition.Builder(ElytronDescriptionConstants.OPTIONS, true)
+            .setAllowExpression(true)
+            .setRestartAllServices()
+            .build();
+
+    static final ObjectTypeAttributeDefinition SERVER_AUTH_MODULE = new ObjectTypeAttributeDefinition.Builder(ElytronDescriptionConstants.SERVER_AUTH_MODULE, CLASS_NAME, MODULE, FLAG, OPTIONS)
+            .setFlags(AttributeAccess.Flag.RESTART_RESOURCE_SERVICES)
+            .setXmlName(ElytronDescriptionConstants.SERVER_AUTH_MODULE)
```
jmesnil: not needed

elytron/src/main/java/org/wildfly/extension/elytron/JaspiDefinition.java
```diff
@@ -0,0 +1,210 @@
...
+            .build();
+
+    static final  ObjectListAttributeDefinition SERVER_AUTH_MODULES = new ObjectListAttributeDefinition.Builder(ElytronDescriptionConstants.SERVER_AUTH_MODULES, SERVER_AUTH_MODULE)
+            .setRequired(false)
+            .setRestartAllServices()
+            .setXmlName(ElytronDescriptionConstants.SERVER_AUTH_MODULES)
```
jmesnil: not needed

elytron/src/main/java/org/wildfly/extension/elytron/_private/ElytronSubsystemMessages.java
```diff
@@ -545,4 +545,7 @@
             " information associated with this existing account, use %s. To change the key associated with this existing account, use %s.")
     OperationFailedException certificateAuthorityAccountAlreadyExists(String updateAccount, String changeAccountKey);
 
+    @Message(id = 1057, value = "Failed to create ServerAuthModule [%s]")
+    RuntimeException failedToCreateServerAuthModule(String className, @Cause Exception cause);
```
jmesnil: should you output the module that is expecting to provide the class too?

elytron/src/main/java/org/wildfly/extension/elytron/ElytronSubsystemParser5_0.java
darranl: I think what tends to happen is many start out referencing the Attributes array at the outset - then once we add more attributes we split both versions into new and old.

darranl:
Retest this please

darranl:
@jmesnil the amended message is in including the name of the module.

I will have a look at that setXmlName calls once in - we ran into so many issues with the ObjectType attributes in the past we may have needed them but possibly we are carrying them over unnecessarily now https://issues.jboss.org/browse/WFCORE-4146.

PR #3437, opened 2018-08-10
[WFCORE-3991] Test coverage for read-config-as-features

LittleJohnII:
https://issues.jboss.org/browse/WFCORE-3991

This is additional test coverage for the `read-config-as-features` operation. Covers both standalone and domain mode. The difference between these tests and the tests in #3438 is that these make changes on real subsystems, then expect the operation output to change accordingly. The changes made were chosen fairly arbitrarily, but the idea is that each change is done on a different part of the model.

Depends on #3438  

Cc @aloubyansky


LittleJohnII:
I updated the PR with `@author` tags and copyright. The tests will blow up again, because this expects some fields from #3430 to be in place.

jmesnil:
@LittleJohnII after I merged #3438 and this PR, one of the test failed: https://ci.wildfly.org/viewLog.html?buildId=116533&tab=buildResultsDiv&buildTypeId=WildFlyCore_MasterIgnorWindows.

I am reverting this PR to avoid this failure. Could you please update the  test to fix it?

PR #3453, opened 2018-08-14
[WFCORE-4035]operation-requires-reload message is missing

cyron:
when modify values of task-max-threads and task-keepalive,
operation-requires-reload message is not outputed.
these two attributes is processed in XnioWorker#setOption, and have values return,
so it should always return true.

```
[standalone@localhost:9990 /] /subsystem=io/worker=default:write-attribute(name=task-keepalive, value=10000)
{"outcome" => "success"}
[standalone@localhost:9990 /] /subsystem=io/worker=default:write-attribute(name=task-max-threads, value=10000)
{"outcome" => "success"}
[standalone@localhost:9990 /] /subsystem=io/worker=default:write-attribute(name=io-threads, value=10000)
{
    "outcome" => "success",
    "response-headers" => {
        "operation-requires-reload" => true,
        "process-state" => "reload-required"
    }
}
[standalone@localhost:9990 /] /subsystem=io/worker=default:write-attribute(name=task-core-threads, value=10000)
{
    "outcome" => "success",
    "response-headers" => {
        "operation-requires-reload" => true,
        "process-state" => "reload-required"
    }
}
```

### DIFF
--- a/io/subsystem/src/main/java/org/wildfly/extension/io/WorkerResourceDefinition.java
+++ b/io/subsystem/src/main/java/org/wildfly/extension/io/WorkerResourceDefinition.java
@@ -166,7 +166,8 @@ class WorkerResourceDefinition extends PersistentResourceDefinition {
                 new WorkerWriteAttributeHandler(WORKER_TASK_MAX_THREADS){
                     @Override
                     boolean setValue(XnioWorker worker, ModelNode value) throws IOException {
-                        return worker.setOption(Options.WORKER_TASK_MAX_THREADS, value.asInt()) == null;
+                        worker.setOption(Options.WORKER_TASK_MAX_THREADS, value.asInt());
+                        return true;
                     }
                 });
         resourceRegistration.registerReadWriteAttribute(WORKER_TASK_CORE_THREADS,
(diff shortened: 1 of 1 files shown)


bstansberry:
this is ok to test

bstansberry:
Thanks @cyron! Can you file a JIRA for this in https://issues.jboss.org/projects/WFCORE and update the commit message to reference it?

bstansberry:
Looking at the current code I'm not sure any of these should be triggering reload-required. Looking at XnioWorker.setOption they all seem to be passing the value into the executor, implying that it takes immediate effect. Perhaps the Xnio behavior has changed.

It's not correct to both have something take immediate effect and report reload-required as if no change was made to the runtime.

io/subsystem/src/main/java/org/wildfly/extension/io/WorkerResourceDefinition.java
```diff
@@ -166,7 +166,10 @@ public void registerAttributes(ManagementResourceRegistration resourceRegistrati
                 new WorkerWriteAttributeHandler(WORKER_TASK_MAX_THREADS){
                     @Override
                     boolean setValue(XnioWorker worker, ModelNode value) throws IOException {
-                        return worker.setOption(Options.WORKER_TASK_MAX_THREADS, value.asInt()) == null;
+                        if (value.isDefined()) {
```
bstansberry: Thanks for this. Even if we decide that these various WorkerWriteAttributeHandler subclasses should return false even if worker.setOption returns non-null, if !value.isDefined() the setOption shouldn't be called and the method should return true.

cyron:
Hi, Brian, I just filed 2 issues and update my commit messages.
should I create a separated pull request for each commit?

https://issues.jboss.org/browse/WFCORE-4035
https://issues.jboss.org/browse/WFCORE-4036

jamezp:
The test failures are related to this change.

jmesnil:
@cyron this PR has not been updated for a while and there were failures related to the changes. I'm closing it but do not hesitate to reopen it if you fix the related failures.

PR #3454, opened 2018-08-14
[WFCORE-4031] Test coverage for read-config-as-features(nested=false)

LittleJohnII:
https://issues.jboss.org/browse/WFCORE-4031

PR #3455, opened 2018-08-14
[WFCORE-4030] Upgrade to Galleon and WildFly Galleon Plugins 2.0.0.Beta1

aloubyansky:
https://issues.jboss.org/browse/WFCORE-4030

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -101,10 +101,10 @@
 
         <!-- Non-default maven plugin versions and configuration -->
         <version.jacoco.plugin>0.8.0</version.jacoco.plugin>
-        <version.org.jboss.galleon>2.0.0.Alpha4</version.org.jboss.galleon>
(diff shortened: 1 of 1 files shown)


aloubyansky:
Hopefully the failures are going to be fixed by https://github.com/wildfly/wildfly-core/pull/3454

LittleJohnII:
The first one from https://ci.wildfly.org/viewLog.html?buildId=117134&buildTypeId=WildFlyCore_PullRequest should be. The second failure looks like something else - I'll have to look into that separately.

But even then, Galleon update could not have caused these. The failures are most definitely test issues. I suggest we don't hold this PR because of them.

aloubyansky:
Just for the info, the equivalent version upgrade has passed in  WFLY https://github.com/wildfly/wildfly/pull/11520

PR #3465, opened 2018-08-20
Wfcore 4045 health error codes

kabir:
https://issues.jboss.org/browse/WFCORE-4045 which is needed to resolve https://issues.jboss.org/browse/WFLY-10861.

If the first three 'housekeeping' commits are too many, they can be squashed

### DIFF
--- a/domain-http/interface/src/main/java/org/jboss/as/domain/http/server/ErrorContextHandler.java
+++ b/domain-http/interface/src/main/java/org/jboss/as/domain/http/server/ErrorContextHandler.java
@@ -21,8 +21,6 @@
  */
 package org.jboss.as.domain.http.server;
 
-import java.util.Locale;
-
 import io.undertow.attribute.ExchangeAttributes;
 import io.undertow.predicate.Predicates;
 import io.undertow.server.HttpHandler;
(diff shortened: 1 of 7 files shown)


domain-http/interface/src/main/java/org/jboss/as/domain/http/server/OperatingSystemDetector.java
```diff
@@ -0,0 +1,53 @@
...
+
+    private OperatingSystemDetector(boolean windows) {
+        this.windows = windows;
+    }
+
+    public boolean isWindows() {
+        return windows;
+    }
+
+    public boolean isLinux() {
```
jamezp: Not really a big deal, but it doesn't look like this is used so it could be removed.

kabir: It isn't, I just liked the 'symmetry' :)

domain-http/interface/src/main/java/org/jboss/as/domain/http/server/security/ServerErrorReadinessHandler.java
```diff
@@ -0,0 +1,53 @@
...
+import org.jboss.as.domain.http.server.OperatingSystemDetector;
+import org.jboss.as.domain.http.server.logging.HttpServerLogger;
+
+import io.undertow.server.HttpHandler;
+import io.undertow.server.HttpServerExchange;
+
+/**
+ * A filter to return 500 if the realm is not ready to handle authentication requests.
+ *
+ * @author <a href="mailto:darran.lofthouse@jboss.com">Darran Lofthouse</a>
```
rsvoboda: Darran or Kabir ?

domain-http/interface/src/main/java/org/jboss/as/domain/http/server/ErrorContextHandler.java
```diff
@@ -21,7 +21,13 @@
  */
 package org.jboss.as.domain.http.server;
 
-import java.util.Locale;
+import static io.undertow.predicate.Predicates.not;
```
rsvoboda: unnecessary imports reorder

PR #3474, opened 2018-08-23
[WFCORE-3479]: Deploying an empty deployment and adding content later fails in a composite

ehsavoie:
* Using the original domain to be sure that the deployment is there before using exploded deployment operation.

Jira: https://issues.jboss.org/browse/WFCORE-3479

### DIFF
--- a/host-controller/src/main/java/org/jboss/as/domain/controller/operations/coordination/ServerOperationResolver.java
+++ b/host-controller/src/main/java/org/jboss/as/domain/controller/operations/coordination/ServerOperationResolver.java
@@ -111,6 +111,7 @@ public class ServerOperationResolver {
 
     public static final AttachmentKey<Set<ModelNode>> DONT_PROPAGATE_TO_SERVERS_ATTACHMENT = AttachmentKey.create(Set.class);
     private static final AttachmentKey<ModelNode> DOMAIN_MODEL_ATTACHMENT = AttachmentKey.create(ModelNode.class);
+    private static final AttachmentKey<ModelNode> ORIGINAL_DOMAIN_MODEL_ATTACHMENT = AttachmentKey.create(ModelNode.class);
 
     private enum DomainKey {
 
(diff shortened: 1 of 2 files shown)


testsuite/domain/src/test/java/org/jboss/as/test/integration/domain/suites/ExplodedDeploymentTestCase.java
```diff
@@ -101,11 +100,11 @@
 
 /**
  *
- * @author <a href="mailto:ehugonne@redhat.com">Emmanuel Hugonnet</a>  (c) 2015 Red Hat, inc.
+ * @author <a href="mailto:ehugonne@redhat.com">Emmanuel Hugonnet</a> (c) 2015 Red Hat, inc.
  */
 public class ExplodedDeploymentTestCase {
 
-    private static final int TIMEOUT = TimeoutUtil.adjust(20000);
+    private static final int TIMEOUT = TimeoutUtil.adjust(2000000);
```
bstansberry: Is this just a leftover from dev?

testsuite/domain/src/test/resources/host-configs/host-master.xml
```diff
@@ -127,6 +127,13 @@
 	             <environment-variables>
 	                 <variable name="DOMAIN_TEST_SERVER" value="server"/>
 	             </environment-variables>
+                     <jvm-options>
+                <option value="-Xdebug"/>
+                <option value="-Xnoagent"/>
+                <option value="-Djava.compiler=NONE"/>
+                <!--<option value="-agentlib:jdwp=transport=dt_socket,address=8787,server=y,suspend=y"/>-->
+
+            </jvm-options>
```
bstansberry: This looks like dev leftovers.

bstansberry:
I would prefer we hold this until WF 15. AIUI it's not a blocker bugfix and it changes the way all domain mode rollouts work by passing the original domain model instead of the current one into the server operation resolution. That may have side effects that we need back time to detect.

kabir:
Retest this please

host-controller/src/main/java/org/jboss/as/domain/controller/operations/coordination/ServerOperationResolver.java
```diff
@@ -257,7 +258,7 @@ private static ModelNode cleanOpForDontPropagate(ModelNode op) {
                     return Collections.singletonMap(allServers, operation);
                 }
                 case DEPLOYMENT: {
-                    return getServerExplodedDeploymentOperations(operation, address, domain, host);
+                    return getServerExplodedDeploymentOperations(getOriginalDomainModel(context), operation, address, host);
```
jmesnil: just a note to myself, `getOriginalDomainModel(context)` replaces the `domain` parameter (order of params has changed)

PR #3475, opened 2018-08-23
[WFCORE-4026] Add version 5.0.0 of the Elytron subsystem model and schema.

darranl:
This is for WildFly Core 7

Presently it is not possible to set up transformer tests to transform to WildFly 14 so initially the new tests are set to transform to EAP_7_1_0 but this can be easily switched later rather than delaying adding the tests at all.

https://issues.jboss.org/browse/WFCORE-4026


elytron/src/test/java/org/wildfly/extension/elytron/SubsystemTransformerTestCase.java
```diff
@@ -46,22 +46,58 @@
...
 
     @Test
     public void testTransformerEAP710() throws Exception {
+        testTransformation(EAP_7_1_0, getSubsystemXml("elytron-transformers-1.2.xml"));
+    }
+
+    @Test
+    public void testTransformerEAP720() throws Exception {
+        // TODO Once the controller version EAP_7_2_0 is available this should switch to use it.
         testTransformation(EAP_7_1_0);
```
kabir: We now have a ModelTestControllerVersion supporting this

kabir:
Retest this please

darranl:
I have resubmitted as a new PR as GitHub seems to have lost the ability to display an accurate diff!

https://github.com/wildfly/wildfly-core/pull/3526

PR #3477, opened 2018-08-23
[WFCORE-4023] Upgrade to xerces 2.12.0.SP02

kabir:
https://issues.jboss.org/browse/WFCORE-4023


kabir:
Adding hold so we can see how the TCK runs before merging this

PR #3483, opened 2018-08-24
[WFCORE-4065] Correct output diffs between galleon build and legacy build

bstansberry:
https://issues.jboss.org/browse/WFCORE-4065

I assume moving the formatters earlier in the logging f-g spec corrects the problem because otherwise the order of the refs to the formatters controls the order of their appearance.

### DIFF
--- a/core-feature-pack/pom.xml
+++ b/core-feature-pack/pom.xml
@@ -273,6 +273,21 @@
         </dependency>
 
         <dependency>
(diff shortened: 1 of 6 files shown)


aloubyansky:
The deps are not new, btw. The *-gen artifacts were resolved by the provisioning maven plugin at runtime. I just added them to pom.xml to avoid "hidden" deps.

PR #3485, opened 2018-08-28
[WFCORE-4066] Avoid overriding galleon pack properties when core-feature-pack-licenses.html is created

yersan:
Do not override the feature pack properties for this task so we can get the correct wildfly-core version in the core-feature-pack-licenses.html file when the server is provisioned in wildfly-full.

Jira issue: https://issues.jboss.org/browse/WFCORE-4066

cc: @aloubyansky


aloubyansky:
Thanks, Yeray. Could you please do the same for servlet-galleon-pack and galleon-pack in wildfly?

yersan:
@aloubyansky yes, makes sense, done. See the referenced PR.

PR #3494, opened 2018-08-30
[WFCORE-4079] Ensure tests pass with internationalized loggers and messages

jamezp:
https://issues.jboss.org/browse/WFCORE-4079

These could be squashed to a single commit. I've only left them separate as each commit belongs to a specific test/test suite that fails.


kabir:
Retest this please

PR #3496, opened 2018-08-31
[WFCORE-3996] Update GC logging to work on Java 9+

rstancel:
Jira: https://issues.jboss.org/browse/WFCORE-3996


core-feature-pack/src/main/resources/content/bin/standalone.sh
```diff
@@ -262,6 +262,9 @@ if [ "$PRESERVE_JAVA_OPTS" != "true" ]; then
         fi
     fi
 
+    # Set Java 9+ variable which will be used later
+    setJava9_plus
+
     if [ "$GC_LOG" = "true" ]; then
         # Enable rotating GC logs if the JVM supports it and GC logs are not already enabled
         NO_GC_LOG_ROTATE=`echo $JAVA_OPTS | $GREP "\-verbose:gc"`
```
marekkopecky: `verbose:gc` should not be used for gc-log detection, because `verbose:gc` is not used in JDK11+ ... right?

core-feature-pack/src/main/resources/content/bin/common.bat
```diff
@@ -2,6 +2,10 @@
 call %*
 goto :eof
 
+:setJava9_plus
+    "%JAVA%" --add-modules=java.se -version >nul 2>&1 && (set JAVA_9_PLUS=true) || (set JAVA_9_PLUS=false)
+goto :eof
+
 :setDefaultModularJvmOptions
   "%JAVA%" --add-modules=java.se -version >nul 2>&1 && (set MODULAR_JDK=true) || (set MODULAR_JDK=false)
```
marekkopecky: Code duplication with setJava9_plus

Btw, what is the reason for renaming MODULAR_JDK var to JAVA_9_PLUS var?

marekkopecky:
@jamezp Can you please do some dev precheck of this PR?

core-feature-pack/src/main/resources/content/bin/common.bat
jamezp: I agree we probably should just use the `MODULE_JDK` variable. My guess however is that the unified GC logging was added in Java 9 which is why the variable name is `JAVA_9_PLUS`.

core-feature-pack/src/main/resources/content/bin/common.ps1
```diff
@@ -70,6 +70,14 @@ if (-Not $javaOpts){
 	$javaOpts = Get-Java-Opts
 }
 
+Function Check-For-Java9+ {
```
jamezp: I don't know much about PowerShell but it seems a bit weird to have a `+` in a function name.

core-feature-pack/src/main/resources/content/bin/common.sh
```diff
@@ -1,8 +1,11 @@
 #!/bin/sh -x
 
+setJava9_plus() {
+  $JAVA --add-modules=java.se -version > /dev/null 2>&1 && JAVA_9_PLUS=true || JAVA_9_PLUS=false
```
jamezp: We could probably just use `MODULAR_JDK` here as well and maybe change the function name.

core-feature-pack/src/main/resources/content/bin/common.sh
```diff
@@ -1,8 +1,11 @@
...
+}
+
 setDefaultModularJvmOptions() {
-  $JAVA --add-modules=java.se -version > /dev/null 2>&1 && MODULAR_JDK=true || MODULAR_JDK=false
-  if [ "$MODULAR_JDK" = "true" ]; then
+  if [ "$JAVA_9_PLUS" = "true" ]; then
```
jamezp: The `setJava9_plus` function, or whatever it ends up being called, should be invoked in this function to ensure the variable is set.

core-feature-pack/src/main/resources/content/bin/standalone.sh
jamezp: As I understand it this is true that we don't need or want the `verbose:gc`.

core-feature-pack/src/main/resources/content/bin/standalone.bat
```diff
@@ -232,14 +237,25 @@ if not "%PRESERVE_JAVA_OPTS%" == "true" (
...
 
-        "%JAVA%" -verbose:gc -XX:+PrintGCDetails -XX:+PrintGCDateStamps -XX:+UseGCLogFileRotation -XX:NumberOfGCLogFiles=5 -XX:GCLogFileSize=3M -Xloggc:"%JBOSS_LOG_DIR%\gc.log" -XX:-TraceClassUnloading -version > nul 2>&1
-        if not errorlevel == 1 (
-          if not exist "%JBOSS_LOG_DIR" > nul 2>&1 (
-            mkdir "%JBOSS_LOG_DIR%"
-          )
-		set JAVA_OPTS=%JAVA_OPTS% -verbose:gc -XX:+PrintGCDetails -XX:+PrintGCDateStamps -XX:+UseGCLogFileRotation -XX:NumberOfGCLogFiles=5 -Xloggc:"%JBOSS_LOG_DIR%\gc.log" -XX:GCLogFileSize=3M -XX:-TraceClassUnloading
+            setlocal EnableDelayedExpansion
+            if "!JAVA_9_PLUS!" == "true" (
+                "%JAVA%" -Xlog:gc*:file="%JBOSS_LOG_DIR%\gc.log":time,uptimemillis:filecount=5,filesize=3M -version > nul 2>&1
```
jamezp: I'm not sure this test is really needed for Java 9. Since GC logging is now unified I believe all compliant JVM's are required to support the unified GC log settings.

core-feature-pack/src/main/resources/content/bin/common.bat
rstancel: @marekkopecky this is not renaming :) JAVA 9+ information is necessary for logging. While logging there can be added some JAVA_OPTS options which are then taken as input for modular JDK. So that's the reason why I had to split this function in two parts.

The duplicate code is of course wrong. It should be done in the same way as in common.sh. I will fix it.

core-feature-pack/src/main/resources/content/bin/common.sh
rstancel: @jamezp see my comment above about splitting this function into two parts

core-feature-pack/src/main/resources/content/bin/common.bat
marekkopecky: Line 10 uses MODULAR_JDK, line 6 JAVA_9_PLUS. Both means the same (both are created by the same command). We should have one "conventional" name for both. Currently, WF uses MODULAR_JDK ... I understand the idea for logging (unified GC logging was addid in JDK9, so you name it `JAVA_9_PLUS`, but anyway, we should be consistent and use just one variable name.

rstancel:
@marekkopecky @jamezp I have updated the code. I have removed Java 9+ naming of functions and variables and I have changed log checking to Xloggc. I have made changes only to sh scripts. Once approved I will propagate changes also to other scripts (bat/ps1)

core-feature-pack/src/main/resources/content/bin/standalone.bat
```diff
@@ -232,14 +237,25 @@ if not "%PRESERVE_JAVA_OPTS%" == "true" (
...
+            ) else (
+                "%JAVA%" -XX:+PrintGCDetails -XX:+PrintGCDateStamps -XX:+UseGCLogFileRotation -XX:NumberOfGCLogFiles=5 -XX:GCLogFileSize=3M -Xloggc:"%JBOSS_LOG_DIR%\gc.log" -XX:-TraceClassUnloading -version > nul 2>&1
+            )
+            if not errorlevel == 1 (
+                if not exist "%JBOSS_LOG_DIR" > nul 2>&1 (
```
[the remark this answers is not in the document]
dmlloyd: Also why not just create the directory ahead of time always instead of probing for errors after failure?  The JVM could fail for many other reasons.

rstancel:
Updated bat files

core-feature-pack/src/main/resources/content/bin/standalone.bat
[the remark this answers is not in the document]
rstancel: ad 1st comment - fixed

marekkopecky:
Still doesn't work on linux with JDK11:
1. prepare wf base on wf-core from this PR
2. uncomment line with definition of GC_LOG in standalone.conf.
3. start standalone
4. gc logs are not created -> **bug**
5. start standalone again
6. gc logs are created

core-feature-pack/src/main/resources/content/bin/standalone.sh
```diff
@@ -273,7 +276,13 @@ if [ "$PRESERVE_JAVA_OPTS" != "true" ]; then
...
             mv -f "$JBOSS_LOG_DIR"/gc.log.*.current "$JBOSS_LOG_DIR/backupgc.log.current" >/dev/null 2>&1
-            "$JAVA" $JVM_OPTVERSION -verbose:gc -Xloggc:"$JBOSS_LOG_DIR/gc.log" -XX:+PrintGCDetails -XX:+PrintGCDateStamps -XX:+UseGCLogFileRotation -XX:NumberOfGCLogFiles=5 -XX:GCLogFileSize=3M -XX:-TraceClassUnloading -version >/dev/null 2>&1 && mkdir -p $JBOSS_LOG_DIR && PREPEND_JAVA_OPTS="$PREPEND_JAVA_OPTS -verbose:gc -Xloggc:\"$JBOSS_LOG_DIR/gc.log\" -XX:+PrintGCDetails -XX:+PrintGCDateStamps -XX:+UseGCLogFileRotation -XX:NumberOfGCLogFiles=5 -XX:GCLogFileSize=3M -XX:-TraceClassUnloading"
+            if [ "$MODULAR_JDK" = "true" ]; then
+                TMP_PARAM="-Xlog:gc*:file=$JBOSS_LOG_DIR/gc.log:time,uptimemillis:filecount=5,filesize=3M"
+            else
+                TMP_PARAM="-Xloggc:$JBOSS_LOG_DIR/gc.log -XX:+PrintGCDetails -XX:+PrintGCDateStamps -XX:+UseGCLogFileRotation -XX:NumberOfGCLogFiles=5 -XX:GCLogFileSize=3M -XX:-TraceClassUnloading"
+            fi
+
+            "$JAVA" $JVM_OPTVERSION $TMP_PARAM -version >/dev/null 2>&1 && mkdir -p $JBOSS_LOG_DIR && PREPEND_JAVA_OPTS="$PREPEND_JAVA_OPTS $TMP_PARAM"
```
dmlloyd: I don't think this extra execution is even necessary.  Once we establish that the JVM is modular, we should just go head and create the log directory (maybe we should always create the log directory regardless, in fact).

rstancel: @dmlloyd I agree. Moreover, it was causing one more issue in this script so I have changed it.

rstancel:
@marekkopecky @rsvoboda 
 - log rotation via mv command is working now
 - there is fixed issue with gc log creation after first start (was issue with first mkdir...)
 - .sh and .bat scripts should be fully working now correctly (tested on JDK 8 and 10 on both Fedora and Win10)

core-feature-pack/src/main/resources/content/bin/common.sh
```diff
@@ -1,7 +1,10 @@
 #!/bin/sh -x
 
-setDefaultModularJvmOptions() {
+setModularJvm() {
```
marekkopecky: Method name is setModularJvm, but this method sets MODULAR_JDK variable. It could be a little bit confusing ...

core-feature-pack/src/main/resources/content/bin/standalone.sh
```diff
@@ -262,18 +262,28 @@ if [ "$PRESERVE_JAVA_OPTS" != "true" ]; then
...
+        mkdir -p $JBOSS_LOG_DIR
+        NO_GC_LOG_ROTATE=`echo $JAVA_OPTS | $GREP "\-Xlog\:\?gc"`
         if [ "x$NO_GC_LOG_ROTATE" = "x" ]; then
             # backup prior gc logs
+            mv -f "$JBOSS_LOG_DIR/gc.log" "$JBOSS_LOG_DIR/backupgc.log" >/dev/null 2>&1
             mv -f "$JBOSS_LOG_DIR/gc.log.0" "$JBOSS_LOG_DIR/backupgc.log.0" >/dev/null 2>&1
             mv -f "$JBOSS_LOG_DIR/gc.log.1" "$JBOSS_LOG_DIR/backupgc.log.1" >/dev/null 2>&1
             mv -f "$JBOSS_LOG_DIR/gc.log.2" "$JBOSS_LOG_DIR/backupgc.log.2" >/dev/null 2>&1
             mv -f "$JBOSS_LOG_DIR/gc.log.3" "$JBOSS_LOG_DIR/backupgc.log.3" >/dev/null 2>&1
-            mv -f "$JBOSS_LOG_DIR/gc.log.4" "$JBOSS_LOG_DIR/backupgc.log.4" >/dev/null 2>&1
```
marekkopecky: Why we need remove this line? I start WF with JDK8, invoke some rest end-point that consumes a lot of memory and:
```
[mkopecky@dhcp-10-40-4-197 wildfly]$ ll standalone/log/
total 4360
-rw-rw-r--. 1 mkopecky mkopecky       0 Sep 20 09:44 audit.log
-rw-rw-r--. 1 mkopecky mkopecky 1048832 Sep 20 09:54 gc.log.0
-rw-rw-r--. 1 mkopecky mkopecky 1048826 Sep 20 09:56 gc.log.1
-rw-rw-r--. 1 mkopecky mkopecky 1048821 Sep 20 09:58 gc.log.2
-rw-rw-r--. 1 mkopecky mkopecky  235106 Sep 20 09:58 gc.log.3.current
-rw-rw-r--. 1 mkopecky mkopecky 1048832 Sep 20 09:53 gc.log.4
-rw-rw-r--. 1 mkopecky mkopecky   13813 Sep 20 09:44 server.log
[mkopecky@dhcp-10-40-4-197 wildfly]$
```

rstancel: @marekkopecky copy-paste issue. Line is back

core-feature-pack/src/main/resources/content/bin/common.sh
rstancel: @marekkopecky You are right. Name changed

core-feature-pack/src/main/resources/content/bin/common.sh
```diff
@@ -1,7 +1,10 @@
 #!/bin/sh -x
 
-setDefaultModularJvmOptions() {
+setModularJdk() {
   $JAVA --add-modules=java.se -version > /dev/null 2>&1 && MODULAR_JDK=true || MODULAR_JDK=false
+}
+
+setDefaultModularJvmOptions() {
```
marekkopecky: I'm trying to prepare some automation tests for JDK11 && gc-logging but it seems that you need to call `setModularJdk` at the beginning of the `setDefaultModularJvmOptions`, right? For example, `jboss-cli.sh` doesn't work without this change:
```
[mkopecky@dhcp-10-40-4-197 bin]$ ./jboss-cli.sh 
Exception in thread "main" org.jboss.modules.ModuleLoadError: java.se
...
```

marekkopecky:
> .sh and .bat scripts should be fully working now correctly (tested on JDK 8 and 10 on both Fedora and Win10)

I check bat scripts with JDK11 (because JDK10 is not so relevant now) and it seems that size of gc log files are 9kB, can you please double check bat scripts? Maybe some missing quotes, or something like that?

rstancel:
> > .sh and .bat scripts should be fully working now correctly (tested on JDK 8 and 10 on both Fedora and Win10)
> 
> I check bat scripts with JDK11 (because JDK10 is not so relevant now) and it seems that size of gc log files are 9kB, can you please double check bat scripts? Maybe some missing quotes, or something like that?

Fixed

core-feature-pack/src/main/resources/content/bin/common.sh
rstancel: @marekkopecky Marek, the issue is in jboss-cli.sh not in standalone.sh. I have started EAP with JDK11. When I started CLI with JDK11 it really failed but when I am starting CLI with JDK10 (or lower) it is started successfully.

core-feature-pack/src/main/resources/content/bin/common.bat
```diff
@@ -2,8 +2,11 @@
 call %*
 goto :eof
 
+:setModularJdk
+    "%JAVA%" --add-modules=java.se -version >nul 2>&1 && (set MODULAR_JDK=true) || (set MODULAR_JDK=false)
+goto :eof
+
 :setDefaultModularJvmOptions
-  "%JAVA%" --add-modules=java.se -version >nul 2>&1 && (set MODULAR_JDK=true) || (set MODULAR_JDK=false)
```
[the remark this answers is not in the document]
rstancel: See previous comment.

jmesnil:
tracked by https://issues.jboss.org/browse/JBEAP-14476

core-feature-pack/src/main/resources/content/bin/common.sh
marekkopecky: CLI is just an example as I mentioned in [my previous comment](https://github.com/wildfly/wildfly-core/pull/3496#discussion_r219422721). Same behaviour are in all other scripts (domain, add-user, etc.). CLI works correctly on wf-core master. CLI doesn't work with your PR, so your PR causes regression. 

This regression in your PR blocks automatic testing of your PR.

rsvoboda: Java 10 is dead now, Java 11 is the thing to use and to focus on.
Java 11 is LTS release and will stay with us for long time.

rstancel: @marekkopecky Fixed. You can try now.

jamezp: I just ran the script tests in WildFly Core on Java 11 and they passed on Linux. I've not tested Windows as I don't have Java 11 installed on my VM. I'll try to see if I can install it and test there as well.

core-feature-pack/src/main/resources/content/bin/standalone.bat
```diff
@@ -219,32 +219,44 @@ if "x%JBOSS_CONFIG_DIR%" == "x" (
...
-		set JAVA_OPTS=%JAVA_OPTS% -verbose:gc -XX:+PrintGCDetails -XX:+PrintGCDateStamps -XX:+UseGCLogFileRotation -XX:NumberOfGCLogFiles=5 -Xloggc:"%JBOSS_LOG_DIR%\gc.log" -XX:GCLogFileSize=3M -XX:-TraceClassUnloading
+            setlocal EnableDelayedExpansion
+            if "!MODULAR_JDK!" == "true" (
+                set TMP_PARAM=-Xlog:gc*:file="\"%JBOSS_LOG_DIR%\gc.log\"":time,uptimemillis:filecount=5,filesize=3M
+            ) else (
+                set TMP_PARAM=-Xloggc:"%JBOSS_LOG_DIR%\gc.log" -XX:+PrintGCDetails -XX:+PrintGCDateStamps -XX:+UseGCLogFileRotation -XX:NumberOfGCLogFiles=5 -XX:GCLogFileSize=3M -XX:-TraceClassUnloading
+            )
+            "%JAVA%" !TMP_PARAM! -version > nul 2>&1
+            if not errorlevel == 1 (
+               set JAVA_OPTS="%JAVA_OPTS%" !TMP_PARAM!
```
jamezp: This should be `set "JAVA_OPTS=!JAVA_OPTS! !TMP_PARAM!"`

core-feature-pack/src/main/resources/content/bin/standalone.bat
```diff
@@ -219,32 +219,44 @@ if "x%JBOSS_CONFIG_DIR%" == "x" (
...
+            )
         )
-       )
+        setlocal DisableDelayedExpansion
     )
 
     rem set default modular jvm parameters
     setlocal EnableDelayedExpansion
-    call "!DIRNAME!common.bat" :setDefaultModularJvmOptions "!JAVA_OPTS!"
+    call "!DIRNAME!common.bat" :setDefaultModularJvmOptions !JAVA_OPTS!
```
jamezp: Is there a reason the quotes were removed? Might be okay, just checking.

core-feature-pack/src/main/resources/content/bin/standalone.sh
```diff
@@ -262,18 +262,29 @@ if [ "$PRESERVE_JAVA_OPTS" != "true" ]; then
...
     fi
 
+    # Set flag if JVM is modular
+    setModularJdk
+
     if [ "$GC_LOG" = "true" ]; then
         # Enable rotating GC logs if the JVM supports it and GC logs are not already enabled
-        NO_GC_LOG_ROTATE=`echo $JAVA_OPTS | $GREP "\-verbose:gc"`
+        mkdir -p $JBOSS_LOG_DIR
+        NO_GC_LOG_ROTATE=`echo $JAVA_OPTS | $GREP "\-Xlog\:\?gc"`
```
jamezp: This would need to check for both `-Xlog:gc` and `-Xloggc`.

core-feature-pack/src/main/resources/content/bin/common.ps1
```diff
@@ -167,19 +178,22 @@ Param(
   $PROG_ARGS += "-Djboss.server.config.dir=$global:JBOSS_CONFIG_DIR"
 
   if ($GC_LOG -eq $true){
-    if ($PROG_ARGS -notcontains "-verbose:gc"){
+    $dir = New-Item $JBOSS_LOG_DIR -type directory -ErrorAction SilentlyContinue
```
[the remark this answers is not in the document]
rstancel: @jamezp This is creating new directory. And in case it already exists it is ignoring the error message

core-feature-pack/src/main/resources/content/bin/common.ps1
```diff
@@ -167,19 +178,22 @@ Param(
...
+    if ($PROG_ARGS -notmatch "-Xlog:?gc"){
```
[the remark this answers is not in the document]
rstancel: @jamezp It is a search character. It is evaluated as any other standard character. You can find special characters here: https://ss64.com/ps/syntax-regex.html

core-feature-pack/src/main/resources/content/bin/common.ps1
```diff
@@ -333,12 +348,22 @@ Function Rotate-GC-Logs {
...
+#            $args = (,"-Xlog:gc*:file=%JBOSS_LOG_DIR%\gc.log:time,uptimemillis:filecount=5,filesize=3M",'-version')
+#        }
+#        else
+#        {
+#            $args = (,"-Xloggc:$JBOSS_LOG_DIR/gc.log","-XX:+PrintGCDetails","-XX:+PrintGCDateStamps","-XX:+UseGCLogFileRotation","-XX:NumberOfGCLogFiles=5","-XX:GCLogFileSize=3M","-XX:-TraceClassUnloading",'-version')
+#        }
+#
+#		$OutputVariable = (&$JAVA $args )  | Out-String
+#	}
+#}
```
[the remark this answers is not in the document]
rstancel: @jamezp I was waiting for QE response but I do agree. This seems to be a function which is not used anywhere

core-feature-pack/src/main/resources/content/bin/common.ps1
```diff
@@ -101,17 +101,27 @@ Function Get-Java-Opts {
...
 Param(
-   [string[]]$opts
+   [string[]]$opts,
+   [bool]$modularJDK
 
 ) #end param
     if($PRESERVE_JAVA_OPTS -eq 'true') {
         return $null
     }
+    SetModularJDK
```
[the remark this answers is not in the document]
rstancel: @jamezp You are right. In ps1 it is redundant.

core-feature-pack/src/main/resources/content/bin/standalone.bat
rstancel: @jamezp Why? JAVA_OPTS are not modified before this call. The reason for changing it to !JAVA_OPTS! would be the update of this parameter (as is used later on line 259

core-feature-pack/src/main/resources/content/bin/standalone.bat
rstancel: @jamezp It is not working with quotes. (I cannot explain why)

core-feature-pack/src/main/resources/content/bin/standalone.sh
rstancel: @jamezp it should be checking both of them
$ echo "-Xlog:gc" | grep "\-Xlog\:\?gc"
-Xlog:gc
$ echo "-Xloggc" | grep "\-Xlog\:\?gc"
-Xloggc

core-feature-pack/src/main/resources/content/bin/common.ps1
```diff
@@ -101,17 +101,27 @@ Function Get-Java-Opts {
 	return $JAVA_OPTS
 }
 
+Function SetModularJDK {
```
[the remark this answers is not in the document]
jamezp: That does make sense and I'm good with either keeping the name or changing it.

core-feature-pack/src/main/resources/content/bin/standalone.bat
jamezp: My explanation would be because it's a Windows batch script :) That's fine though, i was mostly just curious.

core-feature-pack/src/main/resources/content/bin/standalone.sh
jamezp: Yup sorry I misread the regex.

core-feature-pack/src/main/resources/content/bin/standalone.bat
jamezp: Ah okay. I honestly don't really understand the delayed expansion thing I just usually try to use the same variable character. If it works it works though :)

core-feature-pack/src/main/resources/content/bin/common.ps1
```diff
@@ -101,17 +101,26 @@ Function Get-Java-Opts {
 	return $JAVA_OPTS
 }
```
marekkopecky: "standalone.bat + jdk11 + gc enabled" creates:
* standalone\log\gc.log
* standalone\log\gc.log.0

"standalone.ps1 + jdk11 + gc enabled" creates:
* standalone\log\gc.log

What causes this different behavior?

rsvoboda:
uff, based on comments it seems just .sh works fine

Not .ps1 or .bat ...

core-feature-pack/src/main/resources/content/bin/common.ps1
jamezp: I kind of thought it was a weird JVM issue. It noticed it was some times created. The PowerShell script was the only odd one out.

core-feature-pack/src/main/resources/content/bin/common.ps1
```diff
@@ -167,19 +177,22 @@ Param(
...
-        $PROG_ARGS += "-XX:NumberOfGCLogFiles=5"
-        $PROG_ARGS += "-XX:GCLogFileSize=3M"
-        $PROG_ARGS += "-XX:-TraceClassUnloading"
-        $PROG_ARGS += "-Xloggc:$JBOSS_LOG_DIR\gc.log"
+
+        if ($MODULAR_JDK -eq $true)
+        {
+            $PROG_ARGS += "-Xlog:gc*:file=`\`"$JBOSS_LOG_DIR\gc.log`\`":time,uptimemillis:filecount=5,filesize=3M"
+        } else {
+            $PROG_ARGS += "-XX:+PrintGCDetails"
```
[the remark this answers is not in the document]
jamezp: That shouldn't be needed I don't think, but it I guess for consistency we should probably add it back.

core-feature-pack/src/main/resources/content/bin/standalone.bat
```diff
@@ -219,32 +219,44 @@ if "x%JBOSS_CONFIG_DIR%" == "x" (
...
         move /y "%JBOSS_LOG_DIR%\gc.log.4" "%JBOSS_LOG_DIR%\backupgc.log.4" > nul 2>&1
         move /y "%JBOSS_LOG_DIR%\gc.log.*.current" "%JBOSS_LOG_DIR%\backupgc.log.current" > nul 2>&1
 
-        "%JAVA%" -verbose:gc -XX:+PrintGCDetails -XX:+PrintGCDateStamps -XX:+UseGCLogFileRotation -XX:NumberOfGCLogFiles=5 -XX:GCLogFileSize=3M -Xloggc:"%JBOSS_LOG_DIR%\gc.log" -XX:-TraceClassUnloading -version > nul 2>&1
-        if not errorlevel == 1 (
-          if not exist "%JBOSS_LOG_DIR" > nul 2>&1 (
-            mkdir "%JBOSS_LOG_DIR%"
-          )
-		set JAVA_OPTS=%JAVA_OPTS% -verbose:gc -XX:+PrintGCDetails -XX:+PrintGCDateStamps -XX:+UseGCLogFileRotation -XX:NumberOfGCLogFiles=5 -Xloggc:"%JBOSS_LOG_DIR%\gc.log" -XX:GCLogFileSize=3M -XX:-TraceClassUnloading
+            setlocal EnableDelayedExpansion
```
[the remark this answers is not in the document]
jamezp: Looks like a tab vs spaces issues :) We have a checkstyle rule to ban tabs, but it looks like scripts aren't processed so I'm personally not too worried about this part.

core-feature-pack/src/main/resources/content/bin/standalone.bat
```diff
@@ -219,32 +219,44 @@ if "x%JBOSS_CONFIG_DIR%" == "x" (
   set  "JBOSS_CONFIG_DIR=%JBOSS_BASE_DIR%\configuration"
 )
```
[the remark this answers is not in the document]
jamezp: I checked this again and I don't see these same issues on Windows 7. Here is what I saw https://gist.github.com/jamezp/4da0ef3ab3a80e4d4943b67941bc4ea0.

core-feature-pack/src/main/resources/content/bin/common.ps1
marekkopecky: sh and bat files creates two log files, so I assume that this (two gc log files) is correct behaviour.

PowerShell script should start WF correctly. PowerShell script should start WF in the same way as bat and sh script.

So this is a bug and this needs to be fixed.

core-feature-pack/src/main/resources/content/bin/common.ps1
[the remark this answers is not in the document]
marekkopecky: I check the also ps1 and sh scritps. `-verbose:gc` was removed also from sh and bat scripts in JDK8 usage. This PR shouldn't do any changes in JDK8 behavior.

I did deeper JDK8 param check.

----

Sh script:
Originally, there was `PREPEND_JAVA_OPTS="$PREPEND_JAVA_OPTS -verbose:gc -Xloggc:\"$JBOSS_LOG_DIR/gc.log\"`

This PR chage this to `TMP_PARAM="-Xloggc:$JBOSS_LOG_DIR/gc.log`

@rstancel What is the reason for the removing of quotation for `-Xloggc` param?

core-feature-pack/src/main/resources/content/bin/standalone.bat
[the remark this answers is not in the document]
rstancel: I agree. All scripts have mixed tab intendation

rstancel:
Issues raised after merge are fixed here: https://github.com/wildfly/wildfly-core/pull/3538

core-feature-pack/src/main/resources/content/bin/standalone.bat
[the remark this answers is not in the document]
marekkopecky: @jamezp I double check this and I confirm my [original comment](https://github.com/wildfly/wildfly-core/pull/3496#discussion_r220894015).

```
VM Flags:
... -XX:MaxHeapSize=2122317824 ...
```

I see this issue on both JDK8 and JDK11, and on both W2012 and W2016. 

Can you please double check? How do you start standalone? Do you use `set "GC_LOG=true"`?

marekkopecky:
> Issues raised after merge are fixed here: #3538

Not sure about issues raised after merge, but issues raised before merge are not fixed by #3538  :

* #3538 fixes just one of two regressions caused by this PR.
   * https://github.com/wildfly/wildfly-core/pull/3496#discussion_r220894015 is not fixed
* And #3538 doesn't fix another new bugs introduced in this PR.
   * https://github.com/wildfly/wildfly-core/pull/3496#discussion_r220904553 is not fixed

core-feature-pack/src/main/resources/content/bin/common.ps1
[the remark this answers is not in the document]
rstancel: @marekkopecky the whole logic was rewritten so also quotes were updated to match this logic. Is there any issue with it?

core-feature-pack/src/main/resources/content/bin/standalone.bat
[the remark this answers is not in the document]
jamezp: @marekkopecky I changed the `standalone.conf.(bat|ps1)` to set the environment variable. Then I just executed `standalone.(bat|ps1)`. This was on both JDK 8 and JDK 11. I checked with `jinfo` and in VisualVM and it seemed to work for me. I didn't get the double quotes like you got around the `JAVA_OPTS` which seems weird. Unfortunately I only have access to Windows 7 which is quite old I realize.

core-feature-pack/src/main/resources/content/bin/common.ps1
jamezp: The only difference I was an extra quote. On the 3rd boot of the `standalone.ps1` the `gc.log.0` shows up for some reason.

```-Xlog:gc*:file="\"C:\Users\jperkins\projects\wildfly-core\build\target\wildfly-core-7.0.0.Alpha2-SNAPSHOT\standalone\log\gc.log\"":time,uptimemillis:filecount=5,filesize=3M
```
```
-Xlog:gc*:file=\"C:\Users\jperkins\projects\wildfly-core\build\target\wildfly-core-7.0.0.Alpha2-SNAPSHOT\standalone\log\gc.log\":time,uptimemillis:filecount=5,filesize=3M
```

The extra quote is in the `.bat` startup though. I'm honestly not sure the issue is.

core-feature-pack/src/main/resources/content/bin/common.ps1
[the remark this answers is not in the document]
marekkopecky: @rstancel I don't see any issue, but AFAIK, WF users can use our scripts in creative way, so we shouldn't change anything, that works correctly.

> the whole logic was rewritten so also quotes were updated to match this logic.

@rstancel Can you please share more details? If I add escaped quotes to `TMP_PARAM` and `eval` to test line 287, scripts seems to work correctly ...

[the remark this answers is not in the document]
rstancel: @marekkopecky from my point of view there were two places where we called exactly same larger sets of arguments. With bringing JDK_MODULE logic it was increased to two such sets. And with consideration the fact that these are not simple calls it is quite possibly that someone (we, WF users...) will do the mistake and will not copy these parameters correctly in both places. So my point here was to make this logic a little simpler and more robust to avoid possible issues.

I see your point not to change things which are working correctly but it is also our responsibility to bring to code the stability and innovation.

core-feature-pack/src/main/resources/content/bin/standalone.bat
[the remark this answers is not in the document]
marekkopecky: @jamezp I check it also on W2008R2 and I see the issue also on W2008. But I'm not sure about W2008r2/W2008 <-> W7 relationship TBH, I guess that W2008 is based on W7.

Anyway, we need to fix this issue.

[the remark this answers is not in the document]
rsvoboda: https://en.wikipedia.org/wiki/Windows_Server_2008_R2 says:
`Microsoft introduced Windows Server 2008 R2 at the 2008 Professional Developers Conference as the server variant of Windows 7. `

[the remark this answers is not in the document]
jamezp: The problem is I'm not sure how to fix this as I only have access to Windows 7. It would be constant guess work and pushing the testing to someone else.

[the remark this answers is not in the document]
jamezp: I was reminded I can ssh into CI and test there so never mind :)

rstancel:
@marekkopecky @jamezp guys, Jeff is going to create new tag of WF core. He asked to open new Jira ticket for issues found after the tag.

marekkopecky:
WFCORE issues, assigned to @jamezp, created yesterday (based on @jmesnil comment on https://github.com/wildfly/wildfly-core/pull/3538):
* https://issues.jboss.org/browse/WFCORE-4147
* https://issues.jboss.org/browse/WFCORE-4148

PR #3503, opened 2018-09-04
License fixes

kabir:
Adding the hold label while waiting for @psakar 's feedback. If the "generate project sources zip delivery bit" commit is needed, some investigation is needed into how to parameterise the assemblyRootFolder path and to check it works

These were pulled into the downstream product branch but need to be part of upstream for future reference.


kabir:
@psakar is the "generate project sources zip delivery bit" commit necessary? If it is we need to adjust the assembly root folder to use properties. If not I can remove it.

psakar:
@kabir it is necessary - it generates release bit

PR #3506, opened 2018-09-05
[WFCORE-4085] Temporarily use WF14 as the transformation target for EAP

kabir:
7.2.0. Once EAP 7.2.0 is released there is a follow up task to switch it
to use EAP 7.2.0 proper in WFCORE-4089

https://issues.jboss.org/browse/WFCORE-4085

### DIFF
--- a/model-test/src/main/java/org/jboss/as/model/test/ModelTestControllerVersion.java
+++ b/model-test/src/main/java/org/jboss/as/model/test/ModelTestControllerVersion.java
@@ -43,7 +43,13 @@ public enum ModelTestControllerVersion {
     EAP_6_4_0 ("7.5.0.Final-redhat-21", true, "7.5.0", "6.4.0"), //EAP 6.4 is the earliest version we support for transformers
     EAP_6_4_7 ("7.5.7.Final-redhat-3", true, "7.5.0", "6.4.7"), //this one is special as it has model change in micro release
     EAP_7_0_0 ("7.0.0.GA-redhat-2", true, "10.0.0", "2.1.0.Final", "7.0.0"),
-    EAP_7_1_0 ("7.1.0.GA-redhat-11", true, "11.0.0", "3.0.10.Final", "7.1.0")
+    EAP_7_1_0 ("7.1.0.GA-redhat-11", true, "11.0.0", "3.0.10.Final", "7.1.0"),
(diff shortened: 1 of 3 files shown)


model-test/src/main/java/org/jboss/as/model/test/ModelTestControllerVersion.java
```diff
@@ -43,7 +43,13 @@
     EAP_6_4_0 ("7.5.0.Final-redhat-21", true, "7.5.0", "6.4.0"), //EAP 6.4 is the earliest version we support for transformers
     EAP_6_4_7 ("7.5.7.Final-redhat-3", true, "7.5.0", "6.4.7"), //this one is special as it has model change in micro release
     EAP_7_0_0 ("7.0.0.GA-redhat-2", true, "10.0.0", "2.1.0.Final", "7.0.0"),
-    EAP_7_1_0 ("7.1.0.GA-redhat-11", true, "11.0.0", "3.0.10.Final", "7.1.0")
+    EAP_7_1_0 ("7.1.0.GA-redhat-11", true, "11.0.0", "3.0.10.Final", "7.1.0"),
+
+    // TODO https://issues.jboss.org/browse/WFCORE-4089 Get rid of EAP_7_2_0_TEMP and enable EAP_7_2_0, once
+    // 7.2.0.GA-redhat-xx version is known
+    EAP_7_2_0_TEMP("11.0.0.Final", false, "14.0.0", "6.0.1.Final", "wf14"),
```
darranl: Wouldn't it be better to just call this 'EAP_7_2_0' - that way subsystems that reference it will not need to be updated when this gets renamed / a new one added.

kabir: @darranl I prefer a replace, for the reasons given in https://issues.jboss.org/browse/WFCORE-4085

darranl: Ok, I suppose the Strings in this enum will still be updatable so as we have later tags we can still have CI runs against those tags.

If we have any further non Final tags will probably still be worth updating the referenced models so we don't end up in a position where we tag WF 15 and find it is incompatible with a later EAP tag.

PR #3510, opened 2018-09-07
[WFCORE-4098] Update SimpleMapAttributeDefinition to record capabilities based on the values.

darranl:
https://issues.jboss.org/browse/WFCORE-4098

### DIFF
--- a/controller/src/main/java/org/jboss/as/controller/SimpleMapAttributeDefinition.java
+++ b/controller/src/main/java/org/jboss/as/controller/SimpleMapAttributeDefinition.java
@@ -26,10 +26,12 @@ package org.jboss.as.controller;
 
 import java.util.Locale;
 import java.util.ResourceBundle;
+import java.util.Set;
 
 import org.jboss.as.controller.descriptions.ModelDescriptionConstants;
 import org.jboss.as.controller.descriptions.ResourceDescriptionResolver;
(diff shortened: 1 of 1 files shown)


controller/src/main/java/org/jboss/as/controller/SimpleMapAttributeDefinition.java
```diff
@@ -65,6 +67,37 @@ protected void addOperationParameterValueTypeDescription(ModelNode node, String
...
+    @Override
+    public void removeCapabilityRequirements(OperationContext context, Resource resource, ModelNode attributeValue) {
+        handleCapabilityRequirements(context, resource, attributeValue, true);
+    }
+
+    private void handleCapabilityRequirements(OperationContext context, Resource resource, ModelNode attributeValue, boolean remove) {
+        CapabilityReferenceRecorder refRecorder = getReferenceRecorder();
+        if (refRecorder != null && attributeValue.isDefined()) {
+            Set<String> KeyList = attributeValue.keys();
+            String[] attributeValues = new String[KeyList.size()];
```
jmesnil: Is that an issue if the `attributeValues` array is not fully filled when `add|removeCapabilityRequirements` is called?

darranl: I think I borrowed the general pattern from the StringList resource so should be similar.

PR #3513, opened 2018-09-10
WFCORE-3873 Support for SNI for server side SSLContext

darranl:
https://issues.jboss.org/browse/WFCORE-3873
This pull request follows up on the following PR and ports the changes on top of version 5 of the WildFly Elytron subsystem's model and schema: -

https://github.com/wildfly/wildfly-core/pull/3424

Additionally the WildFly Elytron subsystem follows a flat model hierarchy without parents depending on their children so I have flattened the model.

I did also have a general chat with Brian if we should be using alternative capability names to avoid circular referencing and in the end decided we didn't need to go to that level of enforcement in the model - we also have other resources that could potentially be configured with impossible cycles so this resource is not unique.

Finally the original discussion that we had regarding this feature was that we would provide a general implementation within WildFly Elytron which is the reason it is exposed by the Elytron subsystem so I have ported the implementation over with the following two PRs: -

https://github.com/wildfly-security/wildfly-elytron/pull/1188
https://github.com/wildfly-security/elytron-web/pull/139

Documentation PR: -

https://github.com/wildfly/wildfly/pull/11659

### DIFF
--- a/core-feature-pack/src/main/resources/modules/system/layers/base/io/undertow/core/main/module.xml
+++ b/core-feature-pack/src/main/resources/modules/system/layers/base/io/undertow/core/main/module.xml
@@ -34,6 +34,7 @@
         <module name="org.jboss.xnio.nio" services="import"/>
         <module name="org.jboss.logging"/>
         <module name="org.wildfly.openssl" optional="true"/>
+        <module name="org.wildfly.security.elytron-web.undertow-server" services="import" />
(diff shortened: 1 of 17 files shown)


jmesnil:
@darranl should #3424 be closed as this one seems to supersedes it?

darranl:
+1 this takes #3424 and ports it on top of the next version of the model.

PR #3517, opened 2018-09-11
[WFCORE-4105] Add tests for add-user, domain, elytron-tool, jboss-cli…

jamezp:
… and standalone scripts.

https://issues.jboss.org/browse/WFCORE-4105

There are some ignored tests in this PR for shells the scripts do not seem to work in. If there is no intention of the scripts to work with these shells we can remove the tests all together. Otherwise we should consider attempting to fixing the scripts.

I also discovered there PowerShell is available for other platforms, https://github.com/PowerShell/PowerShell#get-powershell. Our scripts do not work however as we hard-code the path separator (`\`) which didn't seem to work on Linux. We could attempt to fix that so the `*.ps1` scripts could be tested on other platforms. However these tests may be enough since they'll run on the Windows PR runs.

### DIFF
--- a/testsuite/pom.xml
+++ b/testsuite/pom.xml
@@ -72,6 +72,7 @@
         <module>vault-test-feature-pack</module>
         <module>elytron</module>
         <module>embedded</module>
+        <module>scripts</module>
     </modules>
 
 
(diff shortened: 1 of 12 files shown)


jmesnil:
@jamezp Could you update the testsuite/scripts/pom.xml to use the 7.0.0.Alpha2-SNAPSHOT version?

jamezp:
I removed the label as I've updated the version to 7.0.0.Alpha2-SNAPSHOT.

PR #3521, opened 2018-09-13
[WFCORE-4113] Add a commented line to standalone.conf to show how to use a custom java.security file to override all the Java security properties

fjuma:
https://issues.jboss.org/browse/WFCORE-4113

### DIFF
--- a/core-feature-pack/src/main/resources/content/bin/standalone.conf
+++ b/core-feature-pack/src/main/resources/content/bin/standalone.conf
@@ -81,3 +81,6 @@ fi
 
 # Uncomment this out to control garbage collection logging
 # GC_LOG="true"
+
+# Uncomment and edit to use a custom java.security file to override all the Java security properties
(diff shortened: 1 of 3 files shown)


core-feature-pack/src/main/resources/content/bin/standalone.conf
```diff
@@ -81,3 +81,6 @@ fi
 
 # Uncomment this out to control garbage collection logging
 # GC_LOG="true"
+
+# Uncomment and edit to use a custom java.security file to override all the Java security properties
+#JAVA_OPTS="$JAVA_OPTS -Djava.security.properties==/path/to/custom/java.security"
```
jamezp: Is the double equal meant to be there?

fjuma: Yes, the double equals are needed to make sure that the java.security file is completely overridden.

PR #3525, opened 2018-09-20
[WFCORE-4121] Add modular Java options to the host controller argumen…

jamezp:
…ts if required.

https://issues.jboss.org/browse/WFCORE-4121


launcher/src/test/java/org/wildfly/core/launcher/CommandBuilderTest.java
```diff
@@ -90,8 +88,13 @@ public void testStandaloneBuilder() {
...
 
-        Assert.assertTrue("Missing --add-exports option", commands.contains("--add-exports=jdk.unsupported/sun.reflect=ALL-UNNAMED"));
-
+        // If we're using Java 9+ ensure the modular JDK options were added
+        if (Environment.isModularJavaHome(JAVA_HOME)) {
+            assertArgumentExists(commands, "--add-exports=java.base/sun.nio.ch=ALL-UNNAMED");
+            assertArgumentExists(commands, "--add-exports=jdk.unsupported/sun.reflect=ALL-UNNAMED");
+            assertArgumentExists(commands, "--add-exports=jdk.unsupported/sun.misc=ALL-UNNAMED");
+            assertArgumentExists(commands, "--add-modules=java.se");
+        }
```
marekkopecky: We can also add `else` block here and assert that arguments doesn't exist in non-modular JDK, right?

launcher/src/test/java/org/wildfly/core/launcher/CommandBuilderTest.java
```diff
@@ -138,12 +141,51 @@ public void testDomainBuilder() {
...
         Assert.assertTrue("Missing -secmgr option", commands.contains("-secmgr"));
 
+        // If we're using Java 9+ ensure the modular JDK options were added
+        if (Environment.isModularJavaHome(JAVA_HOME)) {
+            assertArgumentExists(commands, "--add-exports=java.base/sun.nio.ch=ALL-UNNAMED", 2);
+            assertArgumentExists(commands, "--add-exports=jdk.unsupported/sun.reflect=ALL-UNNAMED", 2);
+            assertArgumentExists(commands, "--add-exports=jdk.unsupported/sun.misc=ALL-UNNAMED", 2);
+            assertArgumentExists(commands, "--add-modules=java.se", 2);
+        }
+
```
marekkopecky: Same here ^^^

launcher/src/test/java/org/wildfly/core/launcher/CommandBuilderTest.java
jamezp: Yeah that's probably a good idea. I'll update the PR, thanks!

ropalka:
Looks good to me.

PR #3529, opened 2018-09-26
[WFCORE-4129] WFLYSRV0266: Server home is set to... info msg in domai…

gaol:
…n for RPM installation

Jira: https://issues.jboss.org/browse/WFCORE-4129

There is warning message in script already, it makes no sense to print such warning message in server log anymore. And it has side effect in case that there are symbolic links between distributed files.

### DIFF
--- a/server/src/main/java/org/jboss/as/server/ServerEnvironment.java
+++ b/server/src/main/java/org/jboss/as/server/ServerEnvironment.java
@@ -438,15 +438,6 @@ public class ServerEnvironment extends ProcessEnvironment implements Serializabl
                     throw ServerLogger.ROOT_LOGGER.couldNotCreateServerBaseDirectory(tmp);
                 }
             }
-            try {
-                Path resolved = tmp.toPath().toRealPath();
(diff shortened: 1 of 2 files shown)


jmesnil:
@gaol Which script are you mentioning?

Having the info in the server log ensures that we notify the user of inconsistencies in the server environment regardless of the way the server was started.

gaol:
The original fix is for the warning in Windows system, [the script in common.ps1](https://github.com/wildfly/wildfly-core/blob/master/core-feature-pack/src/main/resources/content/bin/common.ps1#L354) has the check to display the warning message. The same [check is on standalone.sh](https://github.com/wildfly/wildfly-core/blob/master/core-feature-pack/src/main/resources/content/bin/standalone.sh#L100) too.

The original fix has the side effect in case of symbolic links in the distribution. RPM has even more complicated symbolic links.

I think the script can tell the warning message also, so I propose to revert the old changes. :)

gaol:
@jamezp Right, I agree with that. :)

I opened another PR: https://github.com/wildfly/wildfly-core/pull/3596 to suppress this one

PR #3531, opened 2018-09-26
[WFCORE-4135] Added new startup property: --disable-configuration-history

rruizher:
--disable-configuration-history
to do not use backup logic in 'standalone_xml_history'. Useful for using inside immutable containers. Applied by default when Standalone server configuration is in read only mode.
https://issues.jboss.org/browse/WFCORE-4135


darranl:
Reading some of the issue comments it sounds like this one still has development to be completed, as this PR is not complete would it be better to close it and resubmit once development as completed?

rruizher:
True. It is not finished.

PR #3533, opened 2018-09-27
Fix for WFCORE-4096, CLI replace paging support with aesh one

jfdenise:
- Remove support for paging and rely on aesh PagingSupport.


rsvoboda:
@mnovak1 can you please look into this ?

mnovak1:
@jfdenise I've run CLI test against jfdenise:WFCORE-4096 and all tests passed (including Windows where I've applied the workaround with emulateAlternateBuffer()).

jfdenise:
@jmesnil , that is good to merge.

PR #3544, opened 2018-10-05
[WFCORE-4152] HC cannot connect to DC after lost connect if models are out of sync and RBAC is enabled

yersan:
Some operations cannot be applied to the affected servers when the HC reconnects when RBAC is used. The problem is the principal used for the reconnection is anonymous and that user does not have elevated privileges to reconnect, so In-VM style call should be performed to address it and allow the reconnection.

Jira issue: https://issues.jboss.org/browse/WFCORE-4152

### DIFF
--- a/controller/src/main/java/org/jboss/as/controller/remote/TransactionalProtocolOperationHandler.java
+++ b/controller/src/main/java/org/jboss/as/controller/remote/TransactionalProtocolOperationHandler.java
@@ -22,6 +22,7 @@
 
 package org.jboss.as.controller.remote;
 
+import static java.security.AccessController.doPrivileged;
 import static org.jboss.as.controller.descriptions.ModelDescriptionConstants.CANCELLED;
 import static org.jboss.as.controller.descriptions.ModelDescriptionConstants.FAILED;
 import static org.jboss.as.controller.descriptions.ModelDescriptionConstants.FAILURE_DESCRIPTION;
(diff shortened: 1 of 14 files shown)


luck3y:
This, and all related jobs passed here: https://ci.wildfly.org/project.html?projectId=WildFlyCore_PullRequest&branch_WildFlyCore_PullRequest=pull%2F3544

I've corrected the security-manager aggregator and reporting and run it manually for this job, so it should add the Linux - security manager result when it finishes.

luck3y:
You can ignore 'Master - Linux - Security Manager (Pull Request) - merge' this job was renamed to match the others a little better: 'Linux - Security Manager (Pull Request)'

yersan:
Retest this please.

controller/src/main/java/org/jboss/as/controller/remote/TransactionalProtocolOperationHandler.java
```diff
@@ -155,15 +160,27 @@ public Void run() {
...
-                    AccessController.doPrivileged(new PrivilegedAction<Void>() {
-
-                        @Override
-                        public Void run() {
-                            AccessAuditContext.doAs(securityIdentity != null , securityIdentity, remoteAddress, action);
-                            return null;
-                        }
-                    });
-
+                    if (executableRequest.inVmCall && securityIdentity == null) {
```
jmesnil: Could that logic be wrapped up inside the Supplier anonymous class's `get` method?

yersan: Yes, done

yersan:
Hi @jmesnil , commits were squashed and the Supplier function refactorized. Test passed.

PR #3545, opened 2018-10-05
[WFCORE-4155] Upgrade to Galleon and WildFly Galleon Plugins 2.0.1.Final

aloubyansky:
https://issues.jboss.org/browse/WFCORE-4155

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -101,10 +101,10 @@
 
         <!-- Non-default maven plugin versions and configuration -->
         <version.jacoco.plugin>0.8.0</version.jacoco.plugin>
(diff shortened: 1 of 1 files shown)


aloubyansky:
The integration tests fail due to the Galleon and WFGP version conflict between WildFly Core and WildFly Full. There are two ways to proceed:
1) merge the corresponding PR for WildFly https://github.com/wildfly/wildfly/pull/11719;
2) include WildFly provisioning plugin into Servlet feature-pack and not inherit it from Core.

Either way it requires a change in WildFly Full build though.

PR #3547, opened 2018-10-08
[WFCORE-4156] Add missing KernelAPIVersion for WildFly 14

jmesnil:
* Add missing VERSION_7 and VERSION_8 to KernelAPIVersion so that domain transformers are
  properly setup for WildFly 13 & 14.
* Use KernelAPIVersion in KnownRelease enum instead of ints so that
  KernelAPIVersion is the only authoritative sources of Kernel versions.
* Fix inconsistency in KernelAPIVersion that was listing VERSION_4_1 for
  WildFly 10.1 while the correct version was VERSION_4_2 (info in
  KnownRelease was correct).

JIRA: https://issues.jboss.org/browse/WFCORE-4156

### DIFF
--- a/host-controller/src/main/java/org/jboss/as/domain/controller/transformers/KernelAPIVersion.java
+++ b/host-controller/src/main/java/org/jboss/as/domain/controller/transformers/KernelAPIVersion.java
@@ -28,8 +28,12 @@ import org.jboss.as.version.Version;
  *
  * @author Brian Stansberry
  */
-enum KernelAPIVersion {
+public enum KernelAPIVersion {
 
+    // EAP 6.2.0
+    VERSION_1_5(1, 5, 0),
+    // EAP 6.3.0
+    VERSION_1_6(1, 6, 0),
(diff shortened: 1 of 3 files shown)


host-controller/src/main/java/org/jboss/as/domain/controller/transformers/KernelAPIVersion.java
```diff
@@ -28,8 +28,12 @@
  *
  * @author Brian Stansberry
  */
-enum KernelAPIVersion {
+public enum KernelAPIVersion {
 
+    // EAP 6.2.0
+    VERSION_1_5(1, 5, 0),
+    // EAP 6.3.0
+    VERSION_1_6(1, 6, 0),
```
bstansberry: This is fine but someday as part of an overall cleanup of code related to EAP < 6.4 in mixed domain we can clean this out too.

Now the EAP domain.xml actually includes host-exclude resources for 6.2 and 6.3, but we don't actually support that.

PR #3549, opened 2018-10-09
Fix for WFCORE-4159, remove invalid dep on javafx

jfdenise:


testsuite/standalone/src/test/java/org/jboss/as/test/integration/management/cli/LongOutputTestCase.java
```diff
@@ -1,6 +1,5 @@
 package org.jboss.as.test.integration.management.cli;
 
-import com.sun.javafx.util.Utils;
```
jmesnil: do you know if that the only use of JavaFX in the test suite?
Could we also clean up the JavaFX dependency in the pom?

jfdenise: That is an invalid import, really caused by "automatic import resolution of IDE" and overlook at it. Removing from pom, I can't say. Didn't find other occurrences in tests.

PR #3550, opened 2018-10-09
[WFCORE-4127] Disable CLI paging with an option

MMarus:
WFCORE issue: https://issues.jboss.org/browse/WFCORE-4127
EAP issue: https://issues.jboss.org/browse/EAP7-1113
Analysis document: https://github.com/wildfly/wildfly-proposals/pull/131

Disable CLI paging with an option through the jboss-cli.xml or a command-line argument.


core-feature-pack/src/main/resources/content/bin/jboss-cli.xml
```diff
@@ -68,4 +68,11 @@
         <prompt-color>blue</prompt-color>
     </color-output>
 
+    <!-- Uncomment to disable the output paging.
+    By default the output of commands is not printed at once, instead you can scroll through it and search
```
jfdenise: I would say "By default, if the OS supports it, .... int it."

testsuite/standalone/src/test/java/org/jboss/as/test/integration/management/cli/LongOutputTestCase.java
```diff
@@ -297,4 +312,22 @@ private static void waitFor(Callable<Boolean> condition, long timeout) throws Ex
...
     }
 
+
+    /**
+     * Check that the whole output was written at once when the output paging is disabled
+     */
+    @Test
+    public void testDisabledOutputPaging() throws Exception {
+        setupConsole(false);
+        consoleWriter.println("/:read-resource");
```
jfdenise: You could perhaps use the elytron resource to be homogenous with the other testcase.

testsuite/standalone/src/test/java/org/jboss/as/test/integration/management/cli/CliConfigUtils.java
```diff
@@ -0,0 +1,101 @@
...
+import javax.xml.stream.XMLStreamWriter;
+import java.io.File;
+import java.io.IOException;
+import java.io.Writer;
+import java.nio.charset.StandardCharsets;
+import java.nio.file.Files;
+
+import static org.junit.Assert.fail;
+
+public class CliConfigUtils {
```
jfdenise: I am not sure we really need a new class there. Do you plan to refactor other tests to benefit from this refactoring?

MMarus: I think only CliConfigTestCase.java and LongOutputTestCase.java would benefit from this, since only those 2 would use it. I proposed this change to avoid the duplicate code of the createConfigFile() and writeColorConfig() methods in those 2 tests. Could we leave those methods in the CliConfigTestCase and increase their visibility so they can be reused in the LongOutputTestCase ?

PR #3555, opened 2018-10-15
[WFCORE-4167] Upgrade to WildFly Galleon Plugins 2.0.2.Final and prop…

aloubyansky:
…erly configure plugin dependencies

https://issues.jboss.org/browse/WFCORE-4167


aloubyansky:
@wolfc fyi, this is the feature-spec-gen maven dependency-related change in WFCORE

PR #3542, opened 2018-10-03
[WFCORE-4147] Fixed issue with double quotation in .bat script

rstancel:
Jira: https://issues.jboss.org/browse/WFCORE-4147

PR #3556, opened 2018-10-15
[WFCORE-4147] Delete the gc.log after the -version check to see if the GC settings are valid.

jamezp:
https://issues.jboss.org/browse/WFCORE-4147

This replaces #3542


jamezp:
@marekkopecky Could please review this?

PR #3564, opened 2018-10-18
WFCORE-4175 Undertow 2.0.14.Final

stuartwdouglas:


jamezp:
@dosoudil This is going to require a prod re-build of undertow.
@jmesnil Likely won't be the only upgrade required for core, but we will need a core tag probably next Tuesday the 23.Oct.

PR #3568, opened 2018-10-24
[WFCORE-4179] Update MBeanRegistrationService to support new MSC API

ropalka:
https://issues.jboss.org/browse/WFCORE-4179


bstansberry:
This entire class should be removed from core and moved to the sar subsystem, which is the only place it is used. The jmx module is not part of the kernel api and is therefore not something extensions should use (i.e. I don't care about breaking any uses of this class outside of WildFly and WildFly Core.)

If you or @jmesnil feel more cautious about just removing it, then copying to sar and just deprecating the whole thing seems fine too.

ropalka:
@bstansberry follow up here https://github.com/wildfly/wildfly/pull/11792

PR #3572, opened 2018-10-27
[WFCORE-4183][WFCORE-4184]][WFCORE-4186] Fix problem of incorrectly including runtime-only attri…

bstansberry:
…butes in read-resource by dropping support for invalid-for-years-now inclusion of attributes without a ManagementResourceRegistration entry.

https://issues.jboss.org/browse/WFCORE-4183

Also fixes TestUtils.createNillableAttribute, which wasn't actually creating a nillable attribute, it was creating a runtime-only one.

In addition, this cleans up additional problems discovered during tests of WFCORE-4183, wherein transformer tests would fail because the test's read of a transformed domain model would include some runtime-only values, while the WFCORE-4183 fix meant the read-resource result it was comparing to would not. So, the transformed domain model read needs to be fixed:

https://issues.jboss.org/browse/WFCORE-4185
https://issues.jboss.org/browse/WFCORE-4186

Further, the comparison utility itself had a bug, which fixing all of the above brought to the surface:

https://issues.jboss.org/browse/WFCORE-4184

### DIFF
--- a/core-model-test/tests/src/test/java/org/jboss/as/core/model/test/access/AccessAuthorizationTransformationTestCase.java
+++ b/core-model-test/tests/src/test/java/org/jboss/as/core/model/test/access/AccessAuthorizationTransformationTestCase.java
@@ -71,7 +71,8 @@ public class AccessAuthorizationTransformationTestCase extends AbstractCoreModel
         KernelServices legacyServices = mainServices.getLegacyServices(modelVersion);
         Assert.assertTrue(legacyServices.isSuccessfulBoot());
 
-        checkCoreModelTransformation(mainServices, modelVersion);
+        checkCoreModelTransformation(mainServices, modelVersion, new RbacModelFixer(modelVersion), null);
+
         mainServices.shutdown();
     }
 }
--- a/controller/src/main/java/org/jboss/as/controller/operations/global/ReadResourceHandler.java
+++ b/controller/src/main/java/org/jboss/as/controller/operations/global/ReadResourceHandler.java
@@ -23,8 +23,6 @@
 package org.jboss.as.controller.operations.global;
 
 import static org.jboss.as.controller.descriptions.ModelDescriptionConstants.ACCESS_CONTROL;
-import static org.jboss.as.controller.descriptions.ModelDescriptionConstants.ATTRIBUTES;
-import static org.jboss.as.controller.descriptions.ModelDescriptionConstants.DEFAULT;
 import static org.jboss.as.controller.descriptions.ModelDescriptionConstants.FAILURE_DESCRIPTION;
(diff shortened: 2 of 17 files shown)


core-model-test/tests/src/test/java/org/jboss/as/core/model/test/access/AccessAuthorizationTransformationTestCase.java
```diff
@@ -71,7 +73,17 @@ public void testAllowNonRBAC() throws Exception {
         KernelServices legacyServices = mainServices.getLegacyServices(modelVersion);
         Assert.assertTrue(legacyServices.isSuccessfulBoot());
 
-        checkCoreModelTransformation(mainServices, modelVersion);
+        ModelFixer fixer = new RbacModelFixer(modelVersion) {
+
+            @Override
+            public ModelNode fixModel(ModelNode modelNode) {
+                System.out.println("Running for " + modelVersion);
```
jmesnil: should be removed

jmesnil:
the failures looks related to the PR

bstansberry:
@jmesnil I added some logging to this to figure out why the mixed-domain test failed and it showed me why. Based on that I've rejected WFCORE-4185 -- details in the comment there.

So if this PR survives that part will be stripped. WFCORE-4184 makes sense regardless, as does WFCORE-4186. I already filed a separate PR for 4184 and will do the same at some point for 4186 if this one isn't worth pursuing.

WFCORE-4813 is conceptually correct. The only question is whether clients (e.g. HAL) may be counting on the current buggy behavior of including some runtime-only attributes in the read-resource data, thus making the benefit of cleaning that up not worth the effort. As part of this PR I had to correct a couple tests that were doing that.

I probably won't come back to this for some days.

bstansberry:
@jmesnil I went ahead and cleaned this up.  It appears there's really not too much to be concerned about re: any negative impacts on clients from WFCORE-4183. We'd just need to confirm CLI and HAL aren't negatively impacted by no longer getting these in a default read-resourcere:

- The 'content' attribute in the rollout plan resource, source of the testsuite fixes in the last commit here.
- The WFCORE-4186 attributes
- The 'expose-all-logs' attribute in /subsystem=transactions/log-store=log-store (the one that led to all this.)

jmesnil:
@claudio4j Are you ok with that change (given Brian's comment above 
https://github.com/wildfly/wildfly-core/pull/3572#issuecomment-434483854)?

claudio4j:
> @claudio4j Are you ok with that change (given Brian's comment above

@jmesnil no problem with this change, thanks.

bstansberry:
Thanks @claudio4j, and @jmesnil for checking into it!

PR #3575, opened 2018-10-30
[WFCORE-4189] Remove failIfNoTests=true from surefire plugin for rbac…

mnovak1:
… module


mnovak1:
This allows to run one test like:
mvn clean install -DfailIfNoTests=false -Dtest=WebConsoleRedirectionTestCase -DallTests

without having fail on Role Based Access Control module.

PR #3576, opened 2018-10-30
[WFCORE-4190] Distinguish undefined metric value

jmesnil:
add include-undefined-metric parameter for the read-attribute (and
read-resource) operation so they return undefined result for metrics
that are not able to compute their values in their read handlers.

By default, the nclude-undefined-metric parameter is false (to preserve
existing behaviour).

JIRA: https://issues.jboss.org/browse/WFCORE-4190


jmesnil:
retest this please

controller/src/main/resources/org/jboss/as/controller/descriptions/common/LocalDescriptions.properties
```diff
@@ -474,6 +474,7 @@ subsystem.read-config-as-features.nested=Whether child features should be nested
 global.read-attribute=Gets the value of an attribute for the selected resource
 global.read-attribute.name=The name of the attribute to get the value for under the selected resource
 global.read-attribute.include-defaults=Boolean to enable/disable default reading. In case it is set to false only attribute set by user are returned ignoring undefined.
+global.read-attribute.include-undefined-metric=Whether to include undefined metric. If the underlying metric value can not be computed, this flag ensures that the value will remain undefined (without being replaced by a possible 'undefined metric valur' from the attribute definition.
```
bstansberry: Perhaps this should be "include-undefined-metric-values".  That's long but this setting doesn't really control whether the metric is included at all, it's just about what the value will be.  That distinction isn't so meaningful for read-attribute but it is for read-resource, which has 'include-runtime' which turns off including the attribute at all, and also 'include-defaults' which does not turn on/off including the attribute, but only affects what the value is. Because both kinds of behavior exist its more important to be clear about what will happen.

jmesnil: I've amended the commit to use `include-undefined-metric-values` as you suggested.

jmesnil:
retest this please

PR #3578, opened 2018-11-01
[WFCORE-4192] Server-server EJB invocation fails if no security is de…

JiriOndrusek:
…fined on target bean due to faulty permission check

Issue: https://issues.jboss.org/browse/WFCORE-4192


jmesnil:
@dmlloyd do you see an issue in sharing this method between domain-management and remoting codebase?

JiriOndrusek:
@jmesnil I have discussed this solution with @darranl , and it seems that anonymous impl certainly needs revisiting, but this fix should work for now.

dmlloyd:
Yeah I think it's probably okay.

PR #3582, opened 2018-11-02
[WFCORE-4196] SNICombinedWithALPNTestCase fails with IBM JDK

jstourac:
https://issues.jboss.org/browse/WFCORE-4196

I've left this PR with two separate commits as I think it makes sense in this case. Although, I can squash them together if required.


darranl:
Looks good to me, will see what the test results show but accessing the system properties have the potential to be an issue with the security manager,

jstourac:
@darranl, thank you for review. I've just tried locally executing just this one test with security manager and it passed:
```
mvn install -Dtest=SNICombinedWithALPNTestCase -Dsecurity.manager -DfailIfNoTests=false
```
do you think it is a sufficient check?

Also, regarding to the CI runs - I don't think those two failed tests are related to changes in this PR.

PR #3590, opened 2018-11-08
[WFCORE-4204] Add Phase priorities for MicroProfile extensions

jmesnil:
JIRA: https://issues.jboss.org/browse/WFCORE-4204


jmesnil:
@jamezp that one is a leftover from WildFly 14 where the MicroProfile extensions did not define their priorities.
I'm not sure that the rules are to add new priorities (it's been a long time since I added a new one).
Could you check it please?

PR #3593, opened 2018-11-12
[WFCORE-4212] Upgrade to Galleon and WildFly Galleon Plugins 3.0.0.CR1

aloubyansky:
https://issues.jboss.org/browse/WFCORE-4212

Key changes here are:
* generate-feature-specs goal merged into build-feature-pack;
* default config definitions moved from wildfly-feature-pack-build.xml to their own files (and from the generated feature-pack.xml to their own files).

### DIFF
--- a/core-galleon-pack/pom.xml
+++ b/core-galleon-pack/pom.xml
@@ -645,17 +645,6 @@
                 <artifactId>wildfly-galleon-maven-plugin</artifactId>
                 <executions>
                     <execution>
-                        <id>feature-spec-build</id>
-                        <goals>
-                            <goal>generate-feature-specs</goal>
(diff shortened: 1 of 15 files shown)


aloubyansky:
The integration tests failed due to a new XML attribute added here https://github.com/aloubyansky/wildfly-core/blob/galleon3/core-galleon-pack/src/main/resources/packages/cleanup.standalone.data.dir/pm/wildfly/tasks.xml#L5 which is not recognized in the version used in wildfly master.
Here is the corresponding WFLY PR, which when merged should fix the failed builds here https://github.com/wildfly/wildfly/pull/11839

jmesnil:
retest this please

PR #3594, opened 2018-11-13
[WFCORE-4225] Add io.jaegertracing.Configuration logger to log at a W…

jamezp:
…ARN level.

https://issues.jboss.org/browse/WFCORE-4225

### DIFF
--- a/core-galleon-pack/src/main/resources/feature_groups/logging.xml
+++ b/core-galleon-pack/src/main/resources/feature_groups/logging.xml
@@ -27,6 +27,10 @@
         <param name="level" value="WARN"/>
       </feature>
       <feature spec="subsystem.logging.logger">
+        <param name="logger" value="io.jaegertracing.Configuration"/>
+        <param name="level" value="WARN"/>
+      </feature>
+      <feature spec="subsystem.logging.logger">
         <param name="logger" value="org.jboss.as.config"/>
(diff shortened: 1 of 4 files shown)


bstansberry:
This seems ok as a workaround but for the future I wonder if this is the best approach to the xml files given how galleon works.  IOW the stuff in full that brings in features should also be able to bring in related logging subsystem features.

That doesn't help with logging.properties, but I don't know how critical having this kind of thing in logging.properties is.

In any case, comment ^^^ is food for thought; for now this LGTM.

jamezp:
That is a good point. The `logging.properties` would be overwritten anyway and in this specific case it would be overwritten before deployments are processed which is where the message is logged. I still have hope for one day removing the `logging.properties` anyway :)

core-galleon-pack/src/main/resources/feature_groups/logging.xml
```diff
@@ -26,6 +26,10 @@
         <param name="logger" value="com.arjuna"/>
         <param name="level" value="WARN"/>
       </feature>
+      <feature spec="subsystem.logging.logger">
```
jpkrohling: I guess that's what I was missing :)

jamezp: I definitely did at first as well :)

PR #3595, opened 2018-11-14
[WFCORE-4223] IllegalArgumentException when add a server-ssl-sni-context with no host-context-map

nekdozjam:
https://issues.jboss.org/browse/WFCORE-4223


fjuma:
retest this please

PR #3596, opened 2018-11-15
[WFCORE-4129] WFLYSRV0266: Server home is set to... info msg in domain for RPM installation

gaol:
Jira: https://issues.jboss.org/browse/WFCORE-4129

This reduces the side effect in Linux where the symbolic link is used

### DIFF
--- a/server/src/main/java/org/jboss/as/server/ServerEnvironment.java
+++ b/server/src/main/java/org/jboss/as/server/ServerEnvironment.java
@@ -438,15 +438,6 @@ public class ServerEnvironment extends ProcessEnvironment implements Serializabl
                     throw ServerLogger.ROOT_LOGGER.couldNotCreateServerBaseDirectory(tmp);
                 }
             }
-            try {
-                Path resolved = tmp.toPath().toRealPath();
-                if (!tmp.toPath().equals(resolved)){ //WFCORE-2652
-                    ServerLogger.ROOT_LOGGER.serverHomeMismatch(tmp.toPath(), resolved);
-                    tmp = resolved.toFile();
-                }
-            } catch (IOException e) {
-                //
-            }
             serverBaseDir = tmp;
 
             tmp = getFileFromProperty(SERVER_CONFIG_DIR, props);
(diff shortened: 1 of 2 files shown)


server/src/main/java/org/jboss/as/server/ServerEnvironment.java
```diff
@@ -439,9 +440,16 @@ public ServerEnvironment(final String hostControllerName, final Properties props
...
                 if (!tmp.toPath().equals(resolved)){ //WFCORE-2652
-                    ServerLogger.ROOT_LOGGER.serverHomeMismatch(tmp.toPath(), resolved);
+                    Path resolvedHome = resolved.getParent();
+                    if (!standalone) {
+                        int level = 0;
+                        while(level++ < 2 && resolvedHome != null) {
+                            resolvedHome = resolvedHome.getParent();
+                        }
+                    }
+                    ServerLogger.ROOT_LOGGER.serverHomeMismatch(homeDir.toPath(), resolvedHome);
```
bstansberry: If the idea is the user reading this might be confused by the extra levels when the real problem is higher up, I'd rather not add code for helping with that. Also, this results in a different meaning for server home vs the standalone case, so that's also confusing.

gaol: Sorry for the late reply, I was on Chinese new year holidays and PTOs, :)

Yeah, I agree that it is also confusing somehow.

Now [the original enhancement issue](https://issues.jboss.org/browse/JBEAP-10261) for the helping(warning in case of c:\ vs C:\ in Windows) has been rejected, because capital letter in Windows should be supported. So I think the original code for the helping should be removed as well, which means to revert the PR: https://github.com/wildfly/wildfly-core/pull/2346, what do you think?

bstansberry: +1 to removing this.

Please don't just revert the original commit though, because the logger method should remain in the file. It should be commented out though if it is no longer used.  Leaving it but commented helps document the message ID sequence.

gaol:
@bstansberry I have updated the PR to remove the original codes that caused the side effect, the 2 CI errors are non-relevant. :)

PR #3597, opened 2018-11-15
[WFCORE-4228] Add methods to the management API to get Capability Runtime API as an Optional object

yersan:
- Required changes in wildfly core for WFLY-11186
- Add additional method signature to allow get an optional capability API

Jira issue: https://issues.jboss.org/browse/WFCORE-4228

### DIFF
--- a/controller/src/main/java/org/jboss/as/controller/capability/CapabilityServiceSupport.java
+++ b/controller/src/main/java/org/jboss/as/controller/capability/CapabilityServiceSupport.java
@@ -22,6 +22,8 @@
 
 package org.jboss.as.controller.capability;
 
+import java.util.Optional;
+
 import org.jboss.msc.service.ServiceName;
 
 /**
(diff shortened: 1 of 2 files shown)


yersan:
Retest this please

bstansberry:
If @jmesnil is ok with this API addition, can you update the commit message, PR title, PR and JIRA description to match the current scope of the change?

controller/src/main/java/org/jboss/as/controller/capability/CapabilityServiceSupport.java
```diff
@@ -86,6 +88,21 @@ public NoSuchCapabilityException(String message) {
      */
     <T> T getCapabilityRuntimeAPI(String capabilityBaseName, String dynamicPart, Class<T> apiType) throws NoSuchCapabilityException;
```
bstansberry: For consistency, this variant should have an Optional version too.

yersan: I added the missing variant using optional, it is still pending for a review from Jeff.

PR #3604, opened 2018-11-22
[WFCORE-4224] -secmgr and SECMGR=true env property doesn't work for d…

tmiyargi:
https://issues.jboss.org/browse/WFCORE-4224
-secmgr and SECMGR=true env property doesn't work for domain.ps1 script

### DIFF
--- a/core-feature-pack/src/main/resources/content/bin/domain.ps1
+++ b/core-feature-pack/src/main/resources/content/bin/domain.ps1
@@ -16,6 +16,12 @@ $DOMAIN_CONF = $scripts +'\domain.conf.ps1'
 $DOMAIN_CONF = Get-Env RUN_CONF $DOMAIN_CONF
 . $DOMAIN_CONF
 
+Write-Debug "sec mgr: $SECMGR"
+
+if ($SECMGR) {
+    $MODULE_OPTS +="-secmgr";
+}
+
 Set-Global-Variables-Domain
 
 # consolidate the host-controller and command line opts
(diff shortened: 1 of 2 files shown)


marekkopecky:
This PR allows secmgr on host controller and on servers. This PR doesn't allow secmgr on Process controller. So this is partial fix only.

core-feature-pack/src/main/resources/content/bin/domain.conf.ps1
```diff
@@ -56,7 +56,7 @@ if (-Not $JAVA_OPTS) {
 }
 
 # Uncomment this to run with a security manager enabled
-# $SECMGR=$true
+# $global:SECMGR=$true
```
marekkopecky: Is "global:" prefix necessary here? Other variables from this file don't need "global:" prefix ...

jamezp: I'm not sure whether or not it needs to be global, but the `standalone.ps1` checks the `$global:SECMGR` parameter. So either we need to make it global or change the `standalone.ps1` to not use a global environment variable.

The following snippet is from the `standalone.ps1` script.
```
Write-Debug "sec mgr: $global:SECMGR"

if ($global:SECMGR) {
    $MODULE_OPTS +="-secmgr";
}
```

tmiyargi: It is set in the global scope in the common.ps1 so it should not be necessary (it is not, I've tested) to access to it with the global prefix there, I will change the standalone as well

marekkopecky: @tmiyargi +1, https://github.com/wildfly/wildfly-core/blob/master/core-feature-pack/src/main/resources/content/bin/common.ps1#L33

PR #3608, opened 2018-11-28
[WFCORE-2663] CLI, Help match open/close characters

iweiss:
Issue: https://issues.jboss.org/browse/WFCORE-2663

DO NOT MERGE: https://github.com/aeshell/aesh-readline/pull/50

### DIFF
--- a/cli/src/main/java/org/jboss/as/cli/Util.java
+++ b/cli/src/main/java/org/jboss/as/cli/Util.java
@@ -353,6 +353,14 @@ public class Util {
         return WildFlySecurityManager.getPropertyPrivileged("os.name", null).toLowerCase(Locale.ENGLISH).indexOf("windows") >= 0;
     }
 
+    public static boolean isSolaris() {
+        String osName = WildFlySecurityManager.getPropertyPrivileged("os.name", null);
+        if (osName == null) {
+            return false;
+        }
+        return (osName.indexOf("Solaris") > -1) || (osName.indexOf("solaris") > -1) || (osName.indexOf("SunOS") > -1);
+    }
+
     public static boolean isSuccess(ModelNode operationResponse) {
         if (operationResponse != null) {
             return operationResponse.hasDefined(Util.OUTCOME) && operationResponse.get(Util.OUTCOME).asString().equals(Util.SUCCESS);
--- a/cli/src/main/java/org/jboss/as/cli/impl/ReadlineConsole.java
+++ b/cli/src/main/java/org/jboss/as/cli/impl/ReadlineConsole.java
@@ -24,15 +24,19 @@ import java.io.IOException;
 import java.io.InputStream;
 import java.io.OutputStream;
 import java.util.ArrayList;
+import java.util.Arrays;
 import java.util.Collection;
 import java.util.EnumMap;
+import java.util.HashMap;
 import java.util.List;
+import java.util.Map;
 import java.util.Optional;
 import java.util.concurrent.CountDownLatch;
 import java.util.concurrent.ExecutorService;
 import java.util.concurrent.Executors;
 import java.util.function.Consumer;
 import java.util.function.Function;
+
 import org.aesh.readline.PagingSupport;
 import org.aesh.readline.Prompt;
 import org.aesh.readline.Readline;
@@ -42,7 +46,10 @@ import org.aesh.readline.alias.AliasCompletion;
 import org.aesh.readline.alias.AliasManager;
 import org.aesh.readline.alias.AliasPreProcessor;
 import org.aesh.readline.completion.Completion;
+import org.aesh.readline.cursor.CursorListener;
+import org.aesh.readline.cursor.Line;
 import org.aesh.readline.history.FileHistory;
+import org.aesh.readline.terminal.formatting.Color;
 import org.aesh.terminal.Terminal;
 import org.aesh.terminal.Attributes;
 import org.aesh.terminal.utils.ANSI;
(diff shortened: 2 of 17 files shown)


cli/src/main/java/org/jboss/as/cli/impl/ReadlineConsole.java
```diff
@@ -862,7 +1053,7 @@ private void loop() {
                             }
                         }
                     });
-                }, completions, preProcessors, readlineHistory, null, READLINE_FLAGS);
+                }, completions, preProcessors, readlineHistory, cursorListener, READLINE_FLAGS);
```
jfdenise: If char highlighting is disabled, the cursorListener could be null, aesh would then not call into the listener.

iweiss: OK. I thought a null matcher would be a better way to implement but I see your point

rsvoboda:
https://github.com/aeshell/aesh-readline/releases/tag/1.14 is available

rsvoboda:
Ah, already there https://github.com/wildfly/wildfly-core/commit/cfb3370ab861423e4d9c826aa3bc64b0448c5bbc :)

iweiss:
The failed tests on Windows have nothing to do with this PR.

mnovak1:
@iweiss Did you make any tests for this PR? Could you share tested patterns, please? Thanks

iweiss:
@mnovak1 I haven't. The test plan is here, https://github.com/wildfly/wildfly-proposals/pull/82.

rsvoboda:
> Could you share tested patterns, please? 

I read this as not related to tests itself. But if you can share CLI commands you were experimenting with when developing this feature. Some commands full of brackets etc.

iweiss:
Ah, sorry about the misunderstanding. I did all tests using the following pattern:

1. Type `{ [ ( ] } )`
1. Move the cursor to each one of the characters
1. Notice the highlight on the corresponding open/close bracket

iweiss:
@mnovak1 To be completely honest, I don't think it there would be an easy way. I think it would require using æsh's low level functions to manipulate the cursor and command line. Makes sense?

iweiss:
@tterem I was thinking about introducing something similar to that, but decided not too thinking highlight would be enough. What you are suggesting, I think, would require us to do syntax highlighting  because it would require the CLI to be aware of the command being typed to actively highlight the proper bracket.

One use case is `/subsystem=elytron:read-resource(recursive=true`. This is a valid command, so syntax checking the command would fail in this case.

tterem:
@iweiss Well, I was thinking only about changes involving brackets, not other syntax. For example when you have ( ... { ... ) somewhere in the command, it would detect that '{' wasn't closed before ')' and therefore it wouldn't be highlighting '(' and ')' with green colour. But I agree that not all use cases would be solved by this.

rsvoboda:
> when you have ( ... { ... ) somewhere in the command, it would detect that '{' wasn't closed before ')' 

TBH I would like to see this check, @jfdenise what about you?

jfdenise:
@rsvoboda , I agree that your use case is common, it would help validate the whole command. '{' should be colored as red in this case.

iweiss:
I will get back to the drawing board on this one. Stay tuned :)

iweiss:
@jfdenise @rsvoboda @tterem Something like this, https://asciinema.org/a/w9F72hKOEWzGDxTbcTTRBpiHc?

jfdenise:
Nice! I am thinking that colouring in red up front the open character seems quite intrusive. red means wrong, but the operation is not yet wrong.

rsvoboda:
time 0:15 ... `( { [ } ]` ... is that valid ?  shouldn't it be `( { [ ] }` ?
Or CLI will just highlight brackets with odd counts ?

iweiss:
The CLI is highlighting all unclosed brackets, so it highlights `(`. It's not about correctness in this case.

tterem:
I think it looks nice, but as mentioned before, goal should be to highlight  `{ [ } ]` red because even though count of brackets is right, their order is wrong. I think only `{ [ ] }` should be green. Do you plan to implement this as well?

iweiss:
@tterem The CLI already has a "highlight matches" functionality. What about this behaviour, https://asciinema.org/a/bF9q2GQNIvJMQbA1Iny2EOT5h?

tterem:
@iweiss I am not sure I understand. 00:13 - what would happen with red `[` if you wrote `]` to the end of the command? Would it stay red or turned green?

iweiss:
@tterem The scan for unclosed brackets happen when the cursor hits a 'closing bracket character' (`)`, `]`, `}`). So after typing `]` `[` stays red until the cursor reaches a closing bracket

jfdenise:
At 13, I am not sure about the character staying red. When the cursor is on ')' 'The invalid '[' should be coloured in red (as you do it), but when the cursor moves out of ')', it should be back to the default colour. The user is warned that something is wrong, so can go and fixes it. Colouring would be happening only when the cursor is on a "special character".

iweiss:
@tterem @jfdenise @rsvoboda How does this look for you, https://asciinema.org/a/uJYDiEMFqfNaluBZgvdz4v9ya?

jfdenise:
@iweiss ,+1.

tterem:
@iweiss Nice. I think it's good.

iweiss:
Cool. PR updated for review.

tterem:
@iweiss Your code doesn't compile now. I've also checked that locally.

iweiss:
> @iweiss Your code doesn't compile now. I've also checked that locally.

It requires a new release of æsh-readline with https://github.com/aeshell/aesh-readline/pull/48

tterem:
@iweiss Ok, I built it with that but I experienced this weird behaviour:

https://asciinema.org/a/H2oXrU1N34TqWJ1c36QulFQhW

Problem occurs  when moving cursor through unclosed open characters. Can you please have a look at it?

It also happened to me that I typed 

`( [  ]` 

and when I moved cursor to '[', the other ']' just vanished. Then I tried it few more times but it wouldn't happen again.

Also, last week I have prepared some tests for scenarios where all open characters have its close counterpart, but they are broken now. They check ansi code sequence describing cursor movement and highlighting, but although on screen it looks the same, ansi sequences has changed a little after your last update. So, for now, I will do only manual testing until I am sure that we have reached final version of code.

iweiss:
@tterem Weird. I'll take a look and thanks for taking a look at this.

iweiss:
@tterem I think I got it fixed. Would you mind testing it again please?

tterem:
@iweiss It's better now but it still happens when there are two or more unclosed characters:

https://asciinema.org/a/YIPlgXE8nE2edJIQ76OXYHCId

iweiss:
Quick update. Apparently the line doesn't change and those extra characters popping in seem to be some terminal issue. Need to investigate more.

iweiss:
@tterem I'm back at it and after rebasing this onto the latest master I'm not seeing the issues we were seeing before. Do you mind taking a look at this again please?

tterem:
@iweiss I looked at it but I still see it. When I type `( {{{{ )` and then go through it back and forth, I see { popping up on the right. Can you please rebase PR and confirm that you don't see that behaviour now?

tterem:
@iweiss After last rebase, it works fine. I am going to update my tests.

cli/src/main/java/org/jboss/as/cli/impl/ReadlineConsole.java
```diff
@@ -88,6 +98,225 @@
...
+            if (brackets.size() == 1) {
+                AbstractMap.SimpleEntry<Character, Integer> solo = brackets.get(0);
+                if (endStartSeparators.containsKey(solo.getKey()) || startEndSeparators.containsKey(solo.getKey())) {
+                    colorizeOpenPair(line, solo.getValue());
+                    return;
+                }
+            }
+
+            Set<AbstractMap.SimpleEntry<Character, Integer>> closedBrackets = new HashSet<>();
+            Set<AbstractMap.SimpleEntry<Character, Integer>> openBrackets = new HashSet<>();
```
tterem: @iweiss Would you mind using a collection that preserves some reasonable ordering (e.g. LinkedHashSet) for closedBrackets and openBrackets? I am writing tests and only way to test this is to move the cursor through the command and then check ansi sequence that describes cursor movement and color changes. However, this sequence is dependent on HashSet which ordering is not guaranteed. I believe this change wouldn't affect functionality, but would simplify testing and tests maintenance.

iweiss: Not a problem @tterem. PR updated.

iweiss:
PR updated with test cherry-picked from https://github.com/tterem/wildfly-core/commit/db8518509214bef5fafcd1bca261fec0b0998a89.

marekkopecky:
@iweiss: @tterem is OOO this week. Anyway, it seems that Windows tests doesn't work on TeamCity CI. Can you please add https://github.com/marekkopecky/wildfly-core/commit/d7dd34d57b98c625888e28abafe306dc6f3a8037 to this PR?

iweiss:
@marekkopecky thanks for the heads up. PR updated.

cli/src/main/java/org/jboss/as/cli/Util.java
```diff
@@ -346,6 +346,14 @@ public static boolean isWindows() {
         return WildFlySecurityManager.getPropertyPrivileged("os.name", null).toLowerCase(Locale.ENGLISH).indexOf("windows") >= 0;
     }
 
+    public static boolean isSolaris() {
+        String osName = System.getProperty("os.name");
```
jamezp: We should use the `WildFlySecurityManager.getPropertyPrivileged` as it looks like that's what `isWindows()` and others do.

tterem:
@jfdenise Can you please review this PR?
(cc @jamezp)

jfdenise:
@tterem @iweiss , I did a quick test, ran into 2 issues:
/subsystem=keycloak/realm=master:read-resource[(()] ==> we should see '(' in red
/subsystem=keycloak/realm=master:read-resource(\
foo=bar) ==> on multiple lines, move cursor back to closing ')', you got 
in thread "CLI Terminal Connection (uninterruptable)" java.lang.IndexOutOfBoundsException
	at org.aesh.readline.Buffer.get(Buffer.java:92)
	at org.aesh.readline.cursor.Line.getCharacterAtPosition(Line.java:336)
	at org.jboss.as.cli.impl.ReadlineConsole$CharacterMatcher.scan(ReadlineConsole.java:190)

@iweiss The second case implies a fix in aesh-readline and an upgrade.

iweiss:
@jfdenise Second issue is fixed after the fix on æsh-readline. I'm working on the first one.

jfdenise:
@iweiss , thank-you. I should upgrade aesh quite soon.

iweiss:
@tterem @jfdenise I created a separate branch with a new implementation, https://github.com/iweiss/wildfly-core/tree/WFCORE_2663_v2, and the necessary modifications for the new æsh-readline packages, but some of the CLI tests are failing and I don't quite understand why. For example:
```
[ERROR]   BracketsHighlightingTestCase.testBasic:138 expected()> but was:<()                                                                                                       
'()' is not a valid operation name.                                                      
[standalone@127.0.0.1:9990 /] > 
```

@tterem could you take a look at this please?

tterem:
Tests don't pass because cursor/colour operations changed so ansi sequences are different now even in cases when they produce same result. As for error message, terminal doesn't display it correctly (you can see whole ansi sequences for example in IntelliJ IDEA). I will update tests once I am sure that highlighting works correctly.

However, I tested the branch manually and I noticed an issue. Suppose we have '[(()()]'. When we move cursor on ']', first '(' is highlighted with red. Then after moving cursor to ')', '(' is still red, but it shouldn't be. According to [analysis document](https://github.com/wildfly/wildfly-proposals/pull/82), 'If there are unmatched brackets between the matched pair, then these brackets will be highlighted in red'. There is no other mention about highlighting with red.

@jfdenise In private chat, @iweiss mentioned that you were discussing scanning whole line for unclosed brackets, not only between the matched pair. So I want to clarify if you agree with the current proposal or if you want to expand it?

jfdenise:
@tterem , we will comply with the analysis document, no change expected there.

tterem:
@jfdenise Ok, thanks for clarification. @iweiss In that case, your implementation in https://github.com/iweiss/wildfly-core/tree/WFCORE_2663_v2 is not correct.

iweiss:
@tterem https://github.com/iweiss/wildfly-core/tree/WFCORE_2663_v2  has been updated and I think fixes all the issues we had.

tterem:
@iweiss I checked updated branch and mentioned issues are indeed fixed but now command from [testWellFormedExpression()](https://github.com/wildfly/wildfly-core/pull/3608/files#diff-1396f08c332c6ce44b427a4445a3be77R317) doesn't work as expected. When I move cursor to last bracket ')', there are two brackets ('{' and '}') highlighted with red. It seems it's happening when there is bracket pair nested inside another pair of same bracket type. You can see it also on this example: `({{}})`.

iweiss:
@tterem fixed.

tterem:
Still not working as exptected. Try e.g. `[()()]` or `/abc:add({a=[], b=[], c={[],()}})`.

tterem:
@iweiss Seems like something went wrong with your branch. Can you please squash commits together so the tests are triggered again?

iweiss:
@tterem the tests fail because the new code does things in a different way, changing the tests

tterem:
@iweiss I have found following inconsistency:
When I have ((), then third character is correctly matched with the second, as we agreed on analysis PR. However, if I surround them with another bracket pair like this: [ ( ( ) ] and move cursor to ], then middle ( is highlighted with red _[_ ( **(** ) _]_ . I think the first ( shloud be red because the other two belong together: _[_ **(** ( ) _]_.

iweiss:
@tterem Right, that's due to what I clarified on the proposal as the context. When you move the cursor to `]`, then the context is whatever is inside `]` and it's closing bracket, `[`. Then marking the middle `(` is inline with that.

When you leave the `[ ]` context, then the implementation will be triggered again on a new context, `( ( )`. In that new context, the first `(` is the open bracket.

marekkopecky:
> @tterem the tests fail because the new code does things in a different way, changing what the tests expect. 

@iweiss : CI run was even not started because of some git issues, see [logs](https://ci.wildfly.org/viewLog.html?buildId=162664&buildTypeId=WildFlyCore_PullRequest_WildFlyCoreFullIntegration&tab=buildLog). This is not related with testcase related with this RFE.

iweiss:
@marekkopecky Now that's something I've never seen. I'll try to sort this out today.

iweiss:
@marekkopecky All green.

iweiss:
@marekkopecky @tterem I updated the PR with the latest version of the code and squashed it to 1 commit. You can now see the failed tests I was referring to.

tterem:
> @tterem **Right, that's due to what I clarified on the proposal as the context.** When you move the cursor to `]`, then the context is whatever is inside `]` and it's closing bracket, `[`. Then marking the middle `(` is inline with that.
> 
> When you leave the `[ ]` context, then the implementation will be triggered again on a new context, `( ( )`. In that new context, the first `(` is the open bracket.

Hi, I am not sure what do you mean. Where on the proposal did you clarify this? 
I believe ( ( ) is working correctly but when outer [ ] are added, the behaviour should be consistent. If two brackets are matched together, then they belong together - it shouldn't depend on the position of the cursor.

iweiss:
> > @tterem **Right, that's due to what I clarified on the proposal as the context.** When you move the cursor to `]`, then the context is whatever is inside `]` and it's closing bracket, `[`. Then marking the middle `(` is inline with that.
> > When you leave the `[ ]` context, then the implementation will be triggered again on a new context, `( ( )`. In that new context, the first `(` is the open bracket.
> 
> Hi, I am not sure what do you mean. Where on the proposal did you clarify this?

```
In a context where we have a nesting of complex objects as operation parameters, it is difficult to match `(`, `{` and `[` with their closing counterparts. In particular at the end of the command, it is needed to quickly identify missing closing characters.

We are proposing here to highlight the matching pairs when the cursor is on the closing or opening character and, within this context, highlight any open pairs, helping users quickly identify potential issue
```

tterem:
@iweiss Ok, but where did you clarify this particular behaviour? Why should change of 'context' mean that different bracket is marked as wrong?

iweiss:
I finally got time to get back to this, sorry for the delay.

@tterem the idea here is that, once inside the outer `[ ]`, we are in a new "context". So, in the context of `[ ]`, the middle `(` is illegal, but once you step inside then the second `(` is being correctly closed by `)` and the first `(` is just hanging there.

Think that when you are inside `[ ]` you are in "another line", if that makes sense.

tterem:
@iweiss I understand that it's different context but why is that middle bracket red in [ ( ( ) ] ? It's closed by ) right next to it and the first ( is unclosed again. Yes, context is different but if bracket is marked incorrect in one context then it should be incorrect also in second context (and in all other contexts).

iweiss:
@tterem I see your point and that's why I am thinking in contexts. Each context marks what is wrong.

The only way I can think of, at the moment, to implement something like what you are suggesting is by using the æsh syntax interpreter so we can parse the line and know exactly what's wrong, but @jfdenise and I thought that might not be possible based on how the CLI was developed.

I would need to think about it some more but both implementations I did tried and failed to achieve that.

PR #3610, opened 2018-11-29
[WFCORE-4242] allow empty values for Elytron security identity attrib…

Skyllarr:
…utes

JIRA: https://issues.jboss.org/browse/WFCORE-4242

Security identity in Elytron can have attributes associated with it. Right now, when adding attribute through CLI, values must have a minimum length of 1 character. However, it makes sense to allow empty value for an attribute (e.g., for the case where some information is optional so we may not have all the values)


fjuma:
@Skyllarr Looks good. Perhaps a simple test case could be added to [IdentityOperationsTestCase](https://github.com/wildfly/wildfly-core/blob/master/elytron/src/test/java/org/wildfly/extension/elytron/IdentityOperationsTestCase.java) to verify the fix.

Skyllarr:
@fjuma I added a test case.

> @Skyllarr Looks good. Perhaps a simple test case could be added to [IdentityOperationsTestCase](https://github.com/wildfly/wildfly-core/blob/master/elytron/src/test/java/org/wildfly/extension/elytron/IdentityOperationsTestCase.java) to verify the fix.

PR #3611, opened 2018-11-30
WFCORE-4243 test case for [WFCORE-4149 Uploading content in SSL doesn…

TomasHofman:
…'t work]

https://issues.jboss.org/browse/WFCORE-4243

### DIFF
--- a/testsuite/manualmode/src/test/java/org/wildfly/core/test/standalone/mgmt/ManagementMultipartRequestTestCase.java
+++ b/testsuite/manualmode/src/test/java/org/wildfly/core/test/standalone/mgmt/ManagementMultipartRequestTestCase.java
@@ -1,0 +1,391 @@
+package org.wildfly.core.test.standalone.mgmt;
+
+import static org.jboss.as.controller.descriptions.ModelDescriptionConstants.ADD;
+import static org.jboss.as.controller.descriptions.ModelDescriptionConstants.AUTHENTICATION;
+import static org.jboss.as.controller.descriptions.ModelDescriptionConstants.CORE_SERVICE;
+import static org.jboss.as.controller.descriptions.ModelDescriptionConstants.DEPLOYMENT;
+import static org.jboss.as.controller.descriptions.ModelDescriptionConstants.HTTP_INTERFACE;
+import static org.jboss.as.controller.descriptions.ModelDescriptionConstants.INPUT_STREAM_INDEX;
+import static org.jboss.as.controller.descriptions.ModelDescriptionConstants.KEYSTORE_PASSWORD;
+import static org.jboss.as.controller.descriptions.ModelDescriptionConstants.KEYSTORE_PATH;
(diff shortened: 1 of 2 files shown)


TomasHofman:
Retest this please.

testsuite/manualmode/src/test/java/org/wildfly/core/test/standalone/mgmt/ManagementMultipartRequestTestCase.java
```diff
@@ -0,0 +1,410 @@
...
+
+    private static void deleteManagementUsers() throws IOException {
+        Files.write(MGMT_USERS_PROPERTY_PATH, "\n".getBytes(), StandardOpenOption.TRUNCATE_EXISTING);
+    }
+
+    private static String getServerAddress() throws UnknownHostException {
+        return InetAddress.getByName(TestSuiteEnvironment.getServerAddress()).getHostAddress();
+    }
+
+    private static void executeReloadAndWaitForCompletion(ModelControllerClient client, ModelControllerClient afterReloadClient) {
```
jmesnil: why did you not use the methods from org.jboss.as.test.shared.ServerReload to wait for the live server to reload?

testsuite/manualmode/src/test/java/org/wildfly/core/test/standalone/mgmt/ManagementMultipartRequestTestCase.java
```diff
@@ -0,0 +1,410 @@
...
+    }
+
+    private static WebArchive createDeployment() {
+        // include whatever classes, but need to reach certain minimal size
+        return ShrinkWrap.create(WebArchive.class, DEPLOYMENT_NAME)
+                .addPackage(ShrinkWrap.class.getPackage());
+    }
+
+    private static void keyMaterialSetup() throws Exception {
+        FileUtils.deleteDirectory(WORK_DIR);
```
jmesnil: the WORK_DIR should also be cleaned up when the test is tear down

testsuite/manualmode/src/test/java/org/wildfly/core/test/standalone/mgmt/ManagementMultipartRequestTestCase.java
TomasHofman: I did not use it because I needed the waiting logic after reload to use different interface and protocol (remoting over native). When looking again I found that there's a method for this too in ServerReload which I originally overlooked, so I'm switching to that. Thanks.

testsuite/manualmode/src/test/java/org/wildfly/core/test/standalone/mgmt/ManagementMultipartRequestTestCase.java
TomasHofman: Done.

TomasHofman:
There's been a suggestion to move this test to undertow. I'm gonna look into that, closing this for now.

PR #3616, opened 2018-12-05
[WFCORE-4239] WARN if system-property is already set and is being overridden

soul2zimate:
https://issues.jboss.org/browse/WFCORE-4239

### DIFF
--- a/server/src/main/java/org/jboss/as/server/parsing/SystemPropertiesXml.java
+++ b/server/src/main/java/org/jboss/as/server/parsing/SystemPropertiesXml.java
@@ -36,11 +36,13 @@ import static org.jboss.as.controller.parsing.ParseUtils.unexpectedElement;
 
 import java.util.Collections;
 import java.util.List;
+import java.util.Properties;
 
 import javax.xml.stream.XMLStreamException;
 
 import org.jboss.as.controller.ExpressionResolver;
 import org.jboss.as.controller.OperationFailedException;
+import org.jboss.as.controller.logging.ControllerLogger;
 import org.jboss.as.controller.operations.common.Util;
 import org.jboss.as.controller.parsing.Element;
 import org.jboss.as.controller.parsing.Namespace;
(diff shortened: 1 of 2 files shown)


server/src/main/java/org/jboss/as/server/parsing/SystemPropertiesXml.java
```diff
@@ -130,6 +133,18 @@ void parseSystemProperties(final XMLExtendedStreamReader reader, final ModelNode
             if (!setName) {
                 throw ParseUtils.missingRequired(reader, Collections.singleton(NAME));
             }
+
+            Properties properties = WildFlySecurityManager.getSystemPropertiesPrivileged();
```
jmesnil: you could read the system properties outside the `while` block

server/src/main/java/org/jboss/as/server/parsing/SystemPropertiesXml.java
```diff
@@ -130,6 +133,18 @@ void parseSystemProperties(final XMLExtendedStreamReader reader, final ModelNode
...
+            try {
+                String newPropertyValue = SystemPropertyResourceDefinition.VALUE.resolveValue(ExpressionResolver.SIMPLE, op.get(VALUE)).asString();
+                String oldPropertyValue = properties.getProperty(name);
+                if (oldPropertyValue != null && !oldPropertyValue.equals(newPropertyValue)) {
+                    ControllerLogger.ROOT_LOGGER.systemPropertyAlreadyExist(name, oldPropertyValue, newPropertyValue);
+                }
+            } catch (OperationFailedException e) {
+                ServerLogger.AS_ROOT_LOGGER.tracef(e, "Failed to set property %s at parse time, it will be set later in the boot process", name);
```
jmesnil: is the message correct? You are not setting the property here, you just check its resolved value against the `System.getProperties()`.

soul2zimate: Thanks. This message is copied from trace message a few lines below, I will adjust it.

server/src/main/java/org/jboss/as/server/parsing/SystemPropertiesXml.java
soul2zimate: I have moved it outside.

PR #3622, opened 2018-12-08
[WFLY-11488] Deprecating o.j.a.s.Services.addServerExecutorDependency() method.

ropalka:
https://issues.jboss.org/browse/WFLY-11488

### DIFF
--- a/server/src/main/java/org/jboss/as/server/deployment/ManagedExplodedContentServitor.java
+++ b/server/src/main/java/org/jboss/as/server/deployment/ManagedExplodedContentServitor.java
@@ -16,7 +16,7 @@ limitations under the License.
 
 package org.jboss.as.server.deployment;
 
-import static org.jboss.as.server.Services.addServerExecutorDependency;
+import static org.jboss.as.server.Services.requireServerExecutor;
 
 import java.io.IOException;
 import java.nio.file.Files;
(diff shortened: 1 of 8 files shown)


server/src/main/java/org/jboss/as/server/deployment/ManagedExplodedContentServitor.java
```diff
@@ -47,32 +49,44 @@
  * content repo to the jboss.server.data.dir.
  *
  * @author Brian Stansberry
+ * @author <a href="mailto:ropalka@redhat.com">Richard Opalka</a>
  */
-class ManagedExplodedContentServitor implements Service<VirtualFile> {
+class ManagedExplodedContentServitor implements Service {
```
bstansberry: This service value is used outside of this file.

I'm not a fan of getting rid of service values, at least not for non-void services. I do not see the cost+risk/reward ratio as a good one.

ropalka: Value was not removed @bstansberry but replaced with new mechanism.
"Consumer<VirtualFile> virtualFileConsumer" is now responsible for providing the value to service consumers.
IOW line 134 replaces deleted line 159.

bstansberry: Ah,ok, thanks. TBH I don't know if I ever really properly understood that before.

PR #3623, opened 2018-12-11
[WFCORE-4207]: Add capabilities in the Thread subsystem resource definitions.

ehsavoie:
* Adding capabilities and registering services with the new names.

JIRA: https://issues.jboss.org/browse/WFCORE-4207

### DIFF
--- a/threads/src/main/java/org/jboss/as/threads/BoundedQueueThreadPoolResourceDefinition.java
+++ b/threads/src/main/java/org/jboss/as/threads/BoundedQueueThreadPoolResourceDefinition.java
@@ -22,6 +22,9 @@
 
 package org.jboss.as.threads;
 
+import static org.jboss.as.threads.CommonAttributes.BLOCKING_BOUNDED_QUEUE_THREAD_POOL;
+import static org.jboss.as.threads.CommonAttributes.BOUNDED_QUEUE_THREAD_POOL;
+
 import org.jboss.as.controller.AttributeDefinition;
 import org.jboss.as.controller.OperationStepHandler;
 import org.jboss.as.controller.PathElement;
@@ -32,6 +35,8 @@ import org.jboss.msc.service.ServiceName;
 
 import java.util.Arrays;
 import java.util.Collection;
+import org.jboss.as.controller.SimpleResourceDefinition;
+import org.jboss.as.controller.capability.RuntimeCapability;
 
 /**
  * {@link org.jboss.as.controller.ResourceDefinition} for a bounded queue thread pool resource.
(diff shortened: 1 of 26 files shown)


threads/src/main/java/org/jboss/as/threads/BoundedQueueThreadPoolResourceDefinition.java
```diff
@@ -32,13 +35,16 @@
...
+import org.jboss.as.controller.SimpleResourceDefinition;
+import org.jboss.as.controller.capability.RuntimeCapability;
 
 /**
  * {@link org.jboss.as.controller.ResourceDefinition} for a bounded queue thread pool resource.
  *
  * @author Brian Stansberry (c) 2011 Red Hat Inc.
  */
 public class BoundedQueueThreadPoolResourceDefinition extends PersistentResourceDefinition {
+    @Deprecated
```
jmesnil: what's the replacement for this blocking thread pool?

threads/src/main/java/org/jboss/as/threads/BoundedQueueThreadPoolResourceDefinition.java
```diff
@@ -48,46 +54,76 @@
...
+        final BoundedQueueThreadPoolAdd addHandler = new BoundedQueueThreadPoolAdd(blocking, threadFactoryResolver,
+                handoffExecutorResolver, poolNameBase, capability);
         final OperationStepHandler removeHandler = new BoundedQueueThreadPoolRemove(addHandler);
-        return new BoundedQueueThreadPoolResourceDefinition(blocking, registerRuntimeOnly, type, poolNameBase, resolverPrefix, addHandler, removeHandler);
+        return new BoundedQueueThreadPoolResourceDefinition(blocking, registerRuntimeOnly, path, capability,
+                poolNameBase, resolverPrefix, addHandler, removeHandler);
     }
-
     /**
+     * @param blocking
```
jmesnil: I'm nitpicking but if there is no description associated to the javadoc parameters, I see no point to add them

threads/src/main/java/org/jboss/as/threads/ScheduledThreadPoolMetricsHandler.java
```diff
@@ -44,7 +45,11 @@
             PoolAttributeDefinitions.QUEUE_SIZE);
 
     public ScheduledThreadPoolMetricsHandler(final ServiceName serviceNameBase) {
-        super(METRICS, serviceNameBase);
+        this(null, serviceNameBase);
+    }
+
+    public ScheduledThreadPoolMetricsHandler(RuntimeCapability capability, final ServiceName serviceNameBase) {
+        super(METRICS, capability, null);
```
jmesnil: is it correct? shouldn't the 3rd parameter be `serviceNameBase`?

threads/src/main/java/org/jboss/as/threads/ThreadFactoryAdd.java
```diff
@@ -45,10 +46,11 @@
     static final AttributeDefinition[] RW_ATTRIBUTES = new AttributeDefinition[] {
         PoolAttributeDefinitions.GROUP_NAME, PoolAttributeDefinitions.THREAD_NAME_PATTERN, PoolAttributeDefinitions.PRIORITY};
 
-    static final ThreadFactoryAdd INSTANCE = new ThreadFactoryAdd();
+    private final RuntimeCapability cap;
```
jmesnil: it could be worth mentioning that this `cap` can be `null`

threads/src/main/java/org/jboss/as/threads/ThreadFactoryResourceDefinition.java
```diff
@@ -46,13 +49,19 @@ public ThreadFactoryResourceDefinition() {
     }
 
     public ThreadFactoryResourceDefinition(String type) {
-        super(PathElement.pathElement(type),
+        this(PathElement.pathElement(type), RuntimeCapability.Builder.of("org.wildfly.threads." + type , true, ThreadFactory.class).build());
```
jmesnil: should `type` be an open-ended string or rather an enum that list all the supported types.

ehsavoie: There was no previous constraint on type, so an open ended string is acceptable.

PR #3636, opened 2019-01-08
[WFCORE-4266] Support MR JAR resource loading from VFSResourceLoader

dmlloyd:
https://issues.jboss.org/browse/WFCORE-4266

### DIFF
--- a/server/src/main/java/org/jboss/as/server/deployment/module/VFSResourceLoader.java
+++ b/server/src/main/java/org/jboss/as/server/deployment/module/VFSResourceLoader.java
@@ -45,7 +45,10 @@ import java.util.HashMap;
 import java.util.Iterator;
 import java.util.List;
 import java.util.Map;
+import java.util.jar.Attributes;
 import java.util.jar.Manifest;
+import java.util.regex.Matcher;
+import java.util.regex.Pattern;
 
 import org.jboss.as.server.logging.ServerLogger;
 import org.jboss.modules.AbstractResourceLoader;
@@ -64,6 +67,7 @@ import org.jboss.vfs.VirtualFilePermission;
 import org.jboss.vfs.VisitorAttributes;
 import org.jboss.vfs.util.FilterVirtualFileVisitor;
 import org.wildfly.security.manager.WildFlySecurityManager;
+import org.wildfly.security.manager.action.ReadPropertyAction;
 
 /**
  * Resource loader capable of loading resources from VFS archives.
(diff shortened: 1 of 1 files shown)


dmlloyd:
This class does not have unit tests.  We might want to have a test plan for this class before merging.

server/src/main/java/org/jboss/as/server/deployment/module/VFSResourceLoader.java
```diff
@@ -73,6 +76,18 @@
  */
 public class VFSResourceLoader extends AbstractResourceLoader implements IterableResourceLoader {
 
+    private static final String MR_PREFIX = "META-INF/releases/";
```
jaikiran: This appears to be a typo. As per the javadoc of JarFile[1], the path is ```META-INF/versions/```

[1] https://docs.oracle.com/javase/9/docs/api/java/util/jar/JarFile.html

dmlloyd: That's what happens while you think about one part of the code while typing in another part :)  Fixed.

dmlloyd:
I've also pushed up an update where it will only seek out versioned entries if the `Multi-Release` JAR attribute is present.  Note that the spec says that its value doesn't matter.

dmlloyd:
Pushed up an NPE fix. https://github.com/wildfly/wildfly-core/compare/f1ddea9..1cfb7a0

dmlloyd:
I do believe that CI is lying about what version it ran with.

dmlloyd:
Retest this please

bstansberry:
retest this please

server/src/main/java/org/jboss/as/server/deployment/module/VFSResourceLoader.java
```diff
@@ -73,10 +76,23 @@
  */
 public class VFSResourceLoader extends AbstractResourceLoader implements IterableResourceLoader {
 
+    private static final String MR_PREFIX = "META-INF/versions/";
+    private static final int RELEASE;
+
+    static {
+        final Matcher matcher = Pattern.compile("^(?:1\\.)?([0-9]+)").matcher(doPrivileged(new ReadPropertyAction("java.version")));
```
hd42: This RegEx does not match 11.0.1

server/src/main/java/org/jboss/as/server/deployment/module/VFSResourceLoader.java
```diff
@@ -123,18 +139,41 @@ public Manifest run() throws IOException {
                 throw new UndeclaredThrowableException(e);
             }
         }
+        // "A multi-release jar file is a jar file that contains a manifest with a main attribute named "Multi-Release", [...]"
+        multiRelease = manifest != null && manifest.getMainAttributes().containsKey("Multi-Release");
```
hd42: The Keys in the Attributes Map ar not Strings but Attirbutes#Names, so this should probably be containsKey(new Attributes.Name("Multi-Release"))

server/src/main/java/org/jboss/as/server/deployment/module/VFSResourceLoader.java
```diff
@@ -123,18 +141,41 @@ public Manifest run() throws IOException {
                 throw new UndeclaredThrowableException(e);
             }
         }
+        // "A multi-release jar file is a jar file that contains a manifest with a main attribute named "Multi-Release", [...]"
+        multiRelease = manifest != null && manifest.getMainAttributes().containsKey(MULTI_RELEASE_NAME);
```
jaikiran: Hi @dmlloyd, apparently the value of that attribute must be `true` for the multi-release jar to be considered enabled https://mail.openjdk.java.net/pipermail/core-libs-dev/2019-January/058131.html

dmlloyd: Thanks for the note.  Could you please open a JIRA (it can have "minor" priority)?  And a PR if you want to.  Thanks.

jaikiran: Done - https://issues.jboss.org/browse/WFCORE-4298

dmlloyd: Thanks Jaikiran!

PR #3640, opened 2019-01-11
Fix for WFCORE-4273, Galleon layers

jfdenise:
Fix for [WFCORE-4273](https://issues.jboss.org/browse/WFCORE-4273) 
Fix for [WFCORE-4286](https://issues.jboss.org/browse/WFCORE-4286)

- Introduce core layers.
- ResourceDefintion to expose required/passive/optional packages.
- Server and some subsystems updated to advertise injected modules.
- Upgrade to galleon 3.0.1.CR1 and galleon-plugins to 3.0.1.CR2


controller/src/main/java/org/jboss/as/controller/SimpleResourceDefinition.java
```diff
@@ -102,6 +104,7 @@ public SimpleResourceDefinition(final PathElement pathElement, final Description
         this.minOccurs = 0;
         this.maxOccurs = Integer.MAX_VALUE;
         this.feature = true;
+        this.additionalPackages = null;
```
jmesnil: maybe initialize it as an empty array so that the `null`check is not necessary?

core-galleon-pack/src/main/resources/configs/standalone/standalone.xml/config.xml
```diff
@@ -1,5 +1,21 @@
 <?xml version="1.0" ?>
 
 <config xmlns="urn:jboss:galleon:config:1.0" name="standalone.xml" model="standalone">
-    <feature-group name="standalone"/>
+    <layers>
```
jmesnil: does order matter?

core-galleon-pack/src/main/resources/feature_groups/elytron-standalone.xml
```diff
@@ -0,0 +1,20 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<feature-group-spec name="elytron-standalone" xmlns="urn:jboss:galleon:feature-group:1.0">
```
jmesnil: for consistency sake, we should either name the feature group `standalone-xxx` or `xxx-standalone` (e.g. in the modified file just above, we have `standalone-security-realms`)

core-galleon-pack/src/main/resources/feature_groups/elytron-standalone.xml
```diff
@@ -0,0 +1,20 @@
...
+
+    <feature spec="subsystem.elytron">
+        <param name="final-providers" value="combined-providers"/>
+        <param name="disallowed-providers" value="[OracleUcrypto]"/>
+        <feature-group name="elytron-app">
+            <include feature-id="subsystem.elytron.security-domain:security-domain=ApplicationDomain">
+                <param name="realms" value="[{realm=ApplicationRealm,role-decoder=groups-to-roles},{realm=local}]"/>
+                <!--
```
jmesnil: why is it commented? This looks related to domain mode and could be removed, right?

core-galleon-pack/src/main/resources/feature_groups/management-legacy-security.xml
```diff
@@ -0,0 +1,8 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<feature-group-spec name="management-legacy-security" xmlns="urn:jboss:galleon:feature-group:1.0">
```
jmesnil: I'm not sure to understand the name `management-legacy-security`. The feature is about adding the management HTTP interface.

controller/src/main/java/org/jboss/as/controller/ResourceBuilder.java
```diff
@@ -82,6 +83,8 @@
 
     ResourceBuilder addCapabilities(Capability... capability);
 
+    ResourceBuilder addAdditionalPackages(RuntimePackageDependency... packages);
```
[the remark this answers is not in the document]
jfdenise: Thanks for javadoc, will add it. 
Yes, it could be renamed to addAdditionalRuntimePackageDependencies. They are added in addition to the required one defined in module.xml. addAdditional could be rename to: addExtra Seems less redundant than addAdd... and shorter.

controller/src/main/java/org/jboss/as/controller/SimpleResourceDefinition.java
jfdenise: BTW, this field can be initialised from Parameters.additionalPackages  that can be null too. 
Not calling into API if the field is null seems fine to me (and common approach in this class).

core-galleon-pack/src/main/resources/configs/standalone/standalone.xml/config.xml
jfdenise: It has no real impact on the generated configuration. The order of layers is the order galleon will process them. Due to dependency (if any), galleon can change this order when generating the config. In addition wildfly persists configuration in a defined order, @bstansberry and yourself know more on that than I do.

core-galleon-pack/src/main/resources/feature_groups/elytron-standalone.xml
jfdenise: good point, standalone-elytron would be a better name.

core-galleon-pack/src/main/resources/feature_groups/elytron-standalone.xml
jfdenise: It was commented in the original feature-group: standalone-profile from which I copied the content. That is an alternative syntax to set the attribute. We should keep it.

core-galleon-pack/src/main/resources/feature_groups/management-legacy-security.xml
jfdenise: The http interface is added by the management-interface feature-group. This one is only setting the security-realm on the interface.

jfdenise:
> Looking at the example, it seems there is no namespace (or conventions) for layer names.
> @jfdenise do you envision this could be a problem when we add them to WildFly?

I don't see conflict in names in layered servlet and full cloud-profile branch.

> Should we have conventions like a `core.` prefix for WildFly Core layers or something similar?

I am not sure, layers are identified piece of technology and the FP that define them is perhaps no that relevant. A layer is not sealed, a FP (that depends on an FP that defines layer A) could extend A by re-using the same name (A) and update the layer content. I am doing that in full to evolve ee and undertow layers with packages only present in full.

controller/src/main/java/org/jboss/as/controller/ResourceBuilder.java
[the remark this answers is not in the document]
jfdenise: Extra is too vague actually, I will keep addAdditionalRuntimePackageDependencies

bstansberry:
We should do a build with this in it and the -DlegacyBuild prop set and diff the build/target/wildfly... folder vs build-legacy/target/wildfly... (The same can be done with -Drelease and -Dlegacy-release to compare dist output, but build is a good first step.)

But I *think* we never did that for WF 15 (@yersan might know) so it's possible we have some diffs before this patch. So it's probably worthwhile checking master first.

Note that there are always some diffs, so if you find some check with @yersan or myself before wasting time. We kind of know the normal ones that we ignore (some whitespace stuff etc.)

Moving to layers is a lot like the move to using galleon internally for the build. There's a large element of it just being a different way of composing our dists so we can look for deltas vs the 'legacy' way (which thankfully is still around) as a way of checking for problems.

One of the things we'll see is whether element ordering changes and if it does we can decide what to do. This is relevant to one of the previous line review comments from @jmesnil.

I know I did some things to help avoid inconsistent ordering issues, i.e. some top level resource types are persisted in alphabetical order. I think extension and socket-binding; I don't recall what else. I know I considered subsystem as well but I don't think we ever did that.

jfdenise:
@bstansberry , the diff I am observing is that public interface is now located before the management interface. In the legacy-build that is the opposite. This PR introduces that (base-server layer being provisioned first).
WF15 has the same order as legacy (I noticed a difference between legacy and WF15, there is no new line at the end of the standalone.xml file). This PR doesn't change that, no new line.

I suppose that this change to interface order is OK.

I am observing that, whatever the order of management operations sent to create the configuration, the subsystem order is: Logging (if any) 1st, then subsystem ordered alphabetically.

bstansberry:
The interface order has been something that has tended to jump around throughout the move to galleon; it's one of those things that we've ignored as trying to get the legacy and galleon to match wasn't practical. Perhaps we can fix the order of those. In any case that order difference doesn't bother me.

Re: subsystem order, seems I made it fixed in #3260. :)

yersan:
> But I think we never did that for WF 15 (@yersan might know) so it's possible we have some diffs before this patch. So it's probably worthwhile checking master first.
I remember doing one some time ago (I think it was because the legacy build failed), but nothing done specifically to get a complete maintenance in WF15. I've just done one and there are minor differences in legacy configuration files (missing metrics-smallryesubsystem in some legacy configuration, missing some default configurations). I'll create the jiras to address them.

controller/src/main/java/org/jboss/as/controller/ResourceDefinition.java
```diff
@@ -89,6 +89,63 @@ default void registerCapabilities(final ManagementResourceRegistration resourceR
         // no op
     }
 
+    /**
+     * Register "additional" Galleon packages that must be installed in order
+     * for this Resource to function. NB: the packages need to be visible from the feature pack
+     * that contains the ResourceDeifinition. It can't be any package from any feature pack.
```
kabir: s/ResourceDeifinition/ResourceDefinition

jfdenise:
@jmesnil , the failing test on Windows, that seems to occur quite frequently, doesn't show-up in my local run on Windows.
If @bstansberry, @kabir and yourself are done with review, I will squash the commits tomorrow morning.

core-galleon-pack/src/main/resources/layers/standalone/elytron/layer-spec.xml
```diff
@@ -0,0 +1,17 @@
+<?xml version="1.0" ?>
+<layer-spec xmlns="urn:jboss:galleon:layer-spec:1.0" name="elytron">   
+    <dependencies>
+        <layer name="base-server"/> 
+    </dependencies>
+    <!-- elytron layer comes with access-control -->
+    <feature-group name="access-control"/>
```
bstansberry: This doesn't sound right. There shouldn't be anything special about elytron that adds a need for management RBAC configuration.

core-galleon-pack/src/main/resources/layers/standalone/management/layer-spec.xml
```diff
@@ -0,0 +1,14 @@
+<?xml version="1.0" ?>
+<layer-spec xmlns="urn:jboss:galleon:layer-spec:1.0" name="management">  
+    <dependencies>
+        <layer name="base-server"/> 
+    </dependencies>
+    <feature-group name="management-interface"/>
+    <feature-group name="management-sockets-group"/>
+    <!-- management layer comes with security -->
+    <feature-group name="access-control"/>
```
bstansberry: This use of access-control seems a bit odd too. RBAC is about authorizing an authenticated user, so if the user isn't authenticated it doesn't serve much purpose.

controller/src/main/java/org/jboss/as/controller/ResourceDefinition.java
```diff
@@ -89,6 +89,63 @@ default void registerCapabilities(final ManagementResourceRegistration resourceR
...
+     *          The root package for the process type; i.e. the package that provides
+     *          the main module whose name is passed to JBoss Modules when the process
+     *          is launched.
+     *      </li>
+     *      <li>
+     *          The package that installs the module that provides the extension in
+     *          which the resource is defined.
+     *      </li>
+     *      <li>
+     *          Any package that is non-optionally directly required
```
bstansberry: directly or transitively

Perhaps this is wrong in the text I wrote that this was based upon. :)

controller/src/main/java/org/jboss/as/controller/registry/RuntimePackageDependency.java
```diff
@@ -0,0 +1,118 @@
...
+/**
+ * A runtime package dependency expresses a dependency to a galleon package. A
+ * RuntimePackageDependency models 3 different types of dependencies. 'required'
+ * (needed to operate), optional (not required to operate) and passive (required
+ * to operate only if all the dependencies of this dependency's package are
+ * present).
+ *
+ * @author jdenise@redhat.com
+ */
+public class RuntimePackageDependency {
```
bstansberry: final

core-galleon-pack/src/main/resources/layers/standalone/elytron/layer-spec.xml
jfdenise: There is no RBAC layer, RBAC seems to make sense when security is configured. That is why elytron brings it in.

controller/src/main/java/org/jboss/as/controller/registry/RuntimePackageDependency.java
jfdenise: Thank-you. Updated.

controller/src/main/java/org/jboss/as/controller/ResourceDefinition.java
jfdenise: No you are right. Fixed.

core-galleon-pack/src/main/resources/layers/standalone/management/layer-spec.xml
jfdenise: It is a way to bring the feature-in without defining a layer. Seemed to make sense to colocate with management.

jmesnil:
@jfdenise I think code can be merged but could you do some paperwork before please?
Could you squash all the commits together and also create a Component upgrade issue for Galleon in WFCORE (so we can track when components are upgraded).

Please ping me when it's done and I'll merge the PR.

PR #3652, opened 2019-01-23
Fix deprecation warnings resulting from the improper deprecation of AbstractCapability.

pferraro:
AbstractCapability was deprecated over a year ago. However, because RuntimeCapability inherits most of its Capability method implementations from it, users of RuntimeCapability encounter lots of inappropriate deprecation warnings when referencing many of the non-deprecated methods from the Capability interface.  This PR fixes that.


pferraro:
Submitted in haste - fixing.

PR #3653, opened 2019-01-23
WFCORE-4376 Fix deprecation warnings resulting from the improper deprecation of AbstractCapability.

pferraro:
https://issues.jboss.org/browse/WFCORE-4376

AbstractCapability was deprecated a little over a year ago. However, because RuntimeCapability inherits most of its Capability method implementations from it, users of RuntimeCapability encounter lots of inappropriate deprecation warnings when referencing many of the non-deprecated methods from the Capability interface. This PR fixes that and any compilation errors resulting from the invocation of these methods from a raw-typed RuntimeCapability.


jmesnil:
@pferraro patch looks good to me. Could you please open a WFCORE issue so we can track this change?

pferraro:
@jmesnil No problem.  Will fix today.

PR #3661, opened 2019-02-06
[WFCORE-4251] Provide ability to list module dependencies for a deployment and subdeployment

yersan:
A feature that enables a command to list the module dependencies of a specific deployment and subdeployments. 

It is included only a test case for non ear deployments. Ear deployment processors are in wildfly-full, so a new PR will be added by @zhantemirov to cover this section of the test plan.

Jira issue: https://issues.jboss.org/browse/WFCORE-4251
Analysis doc: https://github.com/wildfly/wildfly-proposals/pull/159
Community doc: https://github.com/wildfly/wildfly/pull/12041

### DIFF
--- a/server/src/main/java/org/jboss/as/server/controller/resources/DeploymentAttributes.java
+++ b/server/src/main/java/org/jboss/as/server/controller/resources/DeploymentAttributes.java
@@ -361,6 +361,15 @@ public class DeploymentAttributes {
     @SuppressWarnings("unchecked")
     public static final Map<String, AttributeDefinition> ALL_CONTENT_ATTRIBUTES = createAttributeMap(MANAGED_CONTENT_ATTRIBUTES, UNMANAGED_CONTENT_ATTRIBUTES);
 
+    public static SimpleAttributeDefinition VERBOSE = SimpleAttributeDefinitionBuilder.create(ModelDescriptionConstants.VERBOSE, ModelType.BOOLEAN, true)
+            .setDefaultValue(new ModelNode(false))
+            .build();
+
+    public static final OperationDefinition LIST_MODULES = SimpleOperationDefinitionBuilder.of(ModelDescriptionConstants.LIST_MODULES, DEPLOYMENT_RESOLVER)
+            .addParameter(VERBOSE)
+            .withFlags(Flag.READ_ONLY)
+            .build();
+
     public static final OperationDefinition DEPLOY_DEFINITION = SimpleOperationDefinitionBuilder.of(ModelDescriptionConstants.DEPLOY, DEPLOYMENT_RESOLVER).build();
     public static final OperationDefinition UNDEPLOY_DEFINITION = SimpleOperationDefinitionBuilder.of(ModelDescriptionConstants.UNDEPLOY, DEPLOYMENT_RESOLVER).build();
     public static final OperationDefinition REDEPLOY_DEFINITION = SimpleOperationDefinitionBuilder.of(ModelDescriptionConstants.REDEPLOY, DEPLOYMENT_RESOLVER).build();
--- a/server/src/main/java/org/jboss/as/server/controller/resources/ServerSubDeploymentResourceDefinition.java
+++ b/server/src/main/java/org/jboss/as/server/controller/resources/ServerSubDeploymentResourceDefinition.java
@@ -1,0 +1,47 @@
+/*
+ * Copyright 2018 Red Hat, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *   http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.jboss.as.server.controller.resources;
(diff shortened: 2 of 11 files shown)


bstansberry:
@jmesnil the analysis, test plan and precheck for this are done, and the community docs PR is basically fine (minor polish). So this is good to go from the non-PR-review POV.

server/src/main/java/org/jboss/as/server/controller/resources/ServerSubDeploymentResourceDefinition.java
```diff
@@ -0,0 +1,47 @@
+/*
+ * Copyright 2018 Red Hat, Inc.
```
jmesnil: @yersan fiy you should update your header template to 2019 (this PR can keep them at 2018, that's ok)

yersan: Ok, yes, you're right, thank!

server/src/main/java/org/jboss/as/server/controller/resources/DeploymentAttributes.java
```diff
@@ -361,6 +361,15 @@ public void marshallAsAttribute(AttributeDefinition attribute, ModelNode resourc
     @SuppressWarnings("unchecked")
     public static final Map<String, AttributeDefinition> ALL_CONTENT_ATTRIBUTES = createAttributeMap(MANAGED_CONTENT_ATTRIBUTES, UNMANAGED_CONTENT_ATTRIBUTES);
 
+    public static SimpleAttributeDefinition VERBOSE = SimpleAttributeDefinitionBuilder.create(ModelDescriptionConstants.VERBOSE, ModelType.BOOLEAN, true)
+            .setDefaultValue(new ModelNode(false))
+            .build();
+
+    public static final OperationDefinition LIST_MODULES = SimpleOperationDefinitionBuilder.of(ModelDescriptionConstants.LIST_MODULES, DEPLOYMENT_RESOLVER)
+            .addParameter(VERBOSE)
+            .withFlags(Flag.READ_ONLY)
```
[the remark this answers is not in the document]
yersan: Sadly I did not get this point, what does "reply is describable" mean in this context? Should we not use Flag.READ_ONLY for the operations?

testsuite/standalone/src/test/java/org/wildfly/core/test/standalone/mgmt/api/core/DeploymentModulesListTestCase.java
```diff
@@ -0,0 +1,279 @@
...
+    private static final String EXAMPLE_MODULE_TO_EXCLUDE = "ibm.jdk";
+    private static final String EXAMPLE_USER_MODULE = "jdk.net";
+    private static final String JAR_DEPLOYMENT_NAME_SUFFIX = "-module-test.jar";
+    private static final String JAR_DEPLOYMENT_NAME = EXAMPLE_USER_MODULE + JAR_DEPLOYMENT_NAME_SUFFIX;
+
+    @Inject
+    private static ManagementClient managementClient;
+    private ModelControllerClient client;
+
+    @Before
```
jmesnil: your code is running fine but I don't understand why you can't simply use `@Deployment` to create and deploy your archive?

zhantemirov: @jmesnil Hi Jeff, that was my first written WildFly deployment related test case. I wasn't aware of such annotation existence. I will have to use @Deployment in my future test cases.

yersan: reason: we are using WildflyTestRunner for the test, so `@Deployment` is not available. We don't have arquillian libraries in core (I didn't know that), but it makes sense to use arquillian in wildfly-full since it is the standard way to do it. I'll comment the PR there.

server/src/main/java/org/jboss/as/server/deployment/module/ModuleSpecProcessor.java
```diff
@@ -282,14 +282,9 @@ private ServiceName createModuleService(final DeploymentPhaseContext phaseContex
...
 
-        final List<ModuleDependency> allDependencies = new ArrayList<ModuleDependency>();
-        allDependencies.addAll(dependencies);
-        allDependencies.addAll(localDependencies);
-        allDependencies.addAll(userDependencies);
-
         ModuleResolvePhaseService.installService(phaseContext.getServiceTarget(), moduleDefinition);
 
-        return ModuleLoadService.install(phaseContext.getServiceTarget(), moduleIdentifier, allDependencies);
+        return ModuleLoadService.install(phaseContext.getServiceTarget(), moduleIdentifier, dependencies, localDependencies, userDependencies);
```
rsvoboda: You are passing `dependencies, localDependencies, userDependencies);` 

The method signature is `install(... ,final List<ModuleDependency> userDependencies, final List<ModuleDependency> localDependencies) {`

So localDependencies and userDependencies should switch places.

Tracked in https://issues.jboss.org/browse/WFCORE-4327

yersan: Good catch @rsvoboda , yes, you're right, thanks!

yersan: The fact that the test pass is because the error is corrected later in https://github.com/wildfly/wildfly-core/pull/3661/files/c51e74ff348d0b8febd74068b5ee53ba5117529a#diff-a5f0937871539867ca86d662b4e2fd09R150 where the list are swhitched again

PR #3668, opened 2019-02-12
[WFCORE-4314] - enchance elytron key store commands

baranowb:

- read-alias
-- dumps HEXed key, signature etc. Usually its not required? Add 'verbose' parameter
-- requires 'alias' parameter, maybe should be 'name' to align with other server resources?
- read-aliases - add 'recursive' and 'verbose'(check above) to allow read of more than names?
- add default 'relative-to' to avoid punching "jboss.server.config.dir"


baranowb:
Hmm default dir change seems to break something.

PR #3675, opened 2019-02-15
WFCORE-4314 - improve elytron keystore commands

baranowb:

* read-alias
  -- dumps HEXed key, signature etc. Usually its not required? Add 'verbose' parameter
  -- requires 'alias' parameter, maybe should be 'name' to align with other server resources?
* read-aliases - add 'recursive' and 'verbose'(check above) to allow read of more than names?
* add default 'relative-to' to avoid punching "jboss.server.config.dir"


darranl:
I am questioning if this should follow the RFE process, there are issues with this in relation to changing the default behaviour which likely should be explained in analysis - additionally documentation may be needed in relation to these operations.

darranl:
@jamezp Do you think this should be handled as a feature?

jamezp:
Yes I would agree this should be an RFE. It looks like there are defaults changing, new attributes and attributes changing names. I think we need further explanation via an analysis doc as @darranl suggested.

baranowb:
Thought I closed this one. Ive talked with Farah about it before my PTO and we decided that change in param names might be confusing so it has been dropped.

PR #3677, opened 2019-02-16
[WFCORE-4334] Upgrade to Galleon and WFGP 3.0.1.Final

aloubyansky:
https://issues.jboss.org/browse/WFCORE-4334

The `offline=true` added to `provision` goal is not changing anything for our builds, since all the artifacts needed to produce the dist have already been built and/or installed in the local repo by the Maven build process. This option simply makes sure that if one day Galleon needs to resolve something from a remote repo, it will fail.

### DIFF
--- a/build/pom.xml
+++ b/build/pom.xml
@@ -62,6 +62,7 @@
                             <install-dir>${project.build.directory}/${project.build.finalName}</install-dir>
                             <record-state>false</record-state>
                             <log-time>${galleon.log.time}</log-time>
+                            <offline>true</offline>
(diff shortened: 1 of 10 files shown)


aloubyansky:
Sorry, I realized I broke the integration tests at the point where WildFly feature-packs are being created. I did pay attention to backward compatibility of the provisioning aspect and left out the feature-pack creation. Sorry about that. Could we merge the equivalent upgrade into WildFly first https://github.com/wildfly/wildfly/pull/12080?

PR #3678, opened 2019-02-19
[WFCORE-4332] Add the named-formatter attribute to the syslog-handler.

jamezp:
JIRA for schema bump: https://issues.jboss.org/browse/WFCORE-4337

JIRA for `named-formatter` addition: https://issues.jboss.org/browse/WFCORE-4332

### DIFF
--- a/logging/src/main/java/org/jboss/as/logging/handlers/SyslogHandlerResourceDefinition.java
+++ b/logging/src/main/java/org/jboss/as/logging/handlers/SyslogHandlerResourceDefinition.java
@@ -33,9 +33,13 @@ import org.jboss.as.controller.AttributeDefinition;
 import org.jboss.as.controller.DefaultAttributeMarshaller;
 import org.jboss.as.controller.OperationContext;
 import org.jboss.as.controller.PathElement;
+import org.jboss.as.controller.SimpleAttributeDefinition;
+import org.jboss.as.controller.SimpleAttributeDefinitionBuilder;
 import org.jboss.as.controller.operations.validation.EnumValidator;
 import org.jboss.as.controller.operations.validation.IntRangeValidator;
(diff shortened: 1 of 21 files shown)


jamezp:
retest this please

jamezp:
retest this please

logging/src/main/java/org/jboss/as/logging/handlers/SyslogHandlerResourceDefinition.java
```diff
@@ -118,6 +120,11 @@ public void marshallAsElement(final AttributeDefinition attribute, final ModelNo
             .setValidator(EnumValidator.create(SyslogType.class, EnumSet.allOf(SyslogType.class)))
             .build();
 
+    public static final SimpleAttributeDefinition NAMED_FORMATTER = SimpleAttributeDefinitionBuilder.create(AbstractHandlerDefinition.NAMED_FORMATTER)
+            .setAlternatives()
```
bstansberry: Something missing?

bstansberry: Probably not; looks like cruft.

bstansberry:
Minor point about the cruft aside this looks good.

logging/src/main/java/org/jboss/as/logging/handlers/SyslogHandlerResourceDefinition.java
jamezp: This was to remove the alternatives as the `syslog-handler` doesn't and won't have a `formatter` attribute. All the other resources that have a `named-formatter` attribute have an alternative set of `formatter`. It probably makes sense to document why this is done though.

PR #3679, opened 2019-02-19
[WFCORE-4335] null file is created in bin directory when starting sta…

tmiyargi:
…ndalone.bat

Issue: https://issues.jboss.org/browse/WFCORE-4335


bstansberry:
this is ok to test

PR #3681, opened 2019-02-19
Fix for WFCORE-4339, The layers testsuite should use thin servers

jfdenise:
- Layers based servers provisioned as thin-server.
- Convey maven local repo to sub-process.


jfdenise:
@jmesnil , I made multiple CI runs to experiment with Wildfly full and didn't run into problems. This fix seems good.

PR #3683, opened 2019-02-19
[WFCORE-4309] Value validator for 'host-context-map' attribute of 'se…

Skyllarr:
…rver-ssl-sni-context' resource

JIRA: https://issues.jboss.org/browse/WFCORE-4309

### DIFF
--- a/elytron/src/main/java/org/wildfly/extension/elytron/_private/ElytronSubsystemMessages.java
+++ b/elytron/src/main/java/org/wildfly/extension/elytron/_private/ElytronSubsystemMessages.java
@@ -557,4 +557,6 @@ public interface ElytronSubsystemMessages extends BasicLogger {
     @Message(id = 1060, value = "Fileless KeyStore needs to have a defined type.")
     OperationFailedException filelessKeyStoreMissingType();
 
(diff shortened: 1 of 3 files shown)


elytron/src/main/java/org/wildfly/extension/elytron/_private/ElytronSubsystemMessages.java
```diff
@@ -557,4 +557,6 @@
     @Message(id = 1060, value = "Fileless KeyStore needs to have a defined type.")
     OperationFailedException filelessKeyStoreMissingType();
 
+    @Message(id = 1061, value = "Value of host context map contains invalid hostname.")
```
bstansberry: This message should include the invalid value.

Skyllarr: @bstansberry Thank you. Fixed.

Skyllarr:
@darranl I have incorporated the changes we talked about. I added precompiled pattern and removed the toLowerCase method (was slower) and added the licence header to the test file.

darranl:
@bstansberry do you have any more review on this one?  Just trying to gauge if this can come in as a bug or if additional validation could suggest RFE?

PR #3684, opened 2019-02-20
[WFCORE-4330] Handle model overrides better in the transformer testing

kabir:
https://issues.jboss.org/browse/WFCORE-4330


kabir:
Any test failures not in the subsystem/core-model tests are not caused by this PR

PR #3685, opened 2019-02-20
[WFCORE-4341] Upgrade Undertow version

fl4via:
Jira: https://issues.jboss.org/browse/WFCORE-4341

PR #3687, opened 2019-02-21
[WFCORE-4346] Allow tests to configure LegacyKernelServicesInitialize…

bstansberry:
…r to not load certain resources from the parent CL

https://issues.jboss.org/browse/WFCORE-4346

This is a longer term solution to the test problem hit by #3685. https://github.com/wildfly/wildfly/pull/12098 was a temporary workaround.

@fl4via and @stuartwdouglas FYI.


fl4via:
Thanks, @bstansberry ! I created https://issues.jboss.org/browse/WFLY-11759 to use the new exclusion feature in the tests and then revert the workaround in https://github.com/wildfly/wildfly/pull/12098/commits/8c4127aa1cf1e8a50e4cbd05b6259f93540b04c4

bstansberry:
I added tests for this, so it's ready to go now (assuming they pass on CI as they do locally.)

wolfc:
This PR violates build from source principles by containing child.jar and parent.jar.

PR #3696, opened 2019-02-27
[WFCORE-4352] Avoid hangs from https://bugs.openjdk.java.net/browse/J…

bstansberry:
…DK-8219658

https://issues.jboss.org/browse/WFCORE-4352

### DIFF
--- a/testsuite/manualmode/src/test/java/org/jboss/as/test/manualmode/auditlog/AuditLogBootingSyslogTest.java
+++ b/testsuite/manualmode/src/test/java/org/jboss/as/test/manualmode/auditlog/AuditLogBootingSyslogTest.java
@@ -35,7 +35,9 @@ import org.jboss.as.test.syslogserver.BlockedSyslogServerEventHandler;
 import org.jboss.dmr.ModelNode;
 import org.junit.After;
 import org.junit.Assert;
+import org.junit.Assume;
 import org.junit.Before;
+import org.junit.BeforeClass;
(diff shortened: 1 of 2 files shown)


testsuite/manualmode/src/test/java/org/jboss/as/test/manualmode/auditlog/AuditLogBootingSyslogTest.java
```diff
@@ -60,6 +62,12 @@
     private static final AuditLogToTLSElytronSyslogSetup SYSLOG_SETUP = new AuditLogToTLSElytronSyslogSetup();
 
 
+    @BeforeClass
+    public static void noJDK12() {
+        Assume.assumeFalse("Avoiding JDK 12 due to https://bugs.openjdk.java.net/browse/JDK-8219658", "12".equals(System.getProperty("java.version")));
```
jaikiran: Hi @bstansberry :)
Given that this property can have values like `12-ea`, `12.0.1` and such[1], maybe use `java.specification.version` instead for the String equality check?

[1] https://openjdk.java.net/jeps/223

P.S: (Belated) congratulations on becoming the WildFly project lead :)

bstansberry: @jaikiran Hello, and thanks! Both for the suggestion and the congrats!

PR #3698, opened 2019-03-05
[WFCORE-4359] Ensure the CLI embedded stdio and log contexts are used for CLI loggers.

jamezp:
### [WFCORE-4358](https://issues.jboss.org/browse/WFCORE-4358l)

This commit just upgrades the log manager with a single commit to expose the [`LogContext.getLogContextSelector()`](https://issues.jboss.org/browse/LOGMGR-241) which is required for the second commit.

### [WFCORE-4358](https://issues.jboss.org/browse/WFCORE-4359)
This commit fixes the issue where the incorrect log context could be used when creating CLI loggers and the incorrect `StdioContext` being used when CLI writes log messages.

Instead of just wrapping the current `LogContext` this wraps the CLI `ThreadLocalContextSelector` instead to ensure the correct contexts are used. I tested both the `embed-server` and `embed-host-controller` with the `jboss-cli.sh` script and the `jboss-cli-client.jar`. Log messages from CLI will now use the `jboss-cli-logging.properties` configuration (assuming that's used) and log messages from the server us the correct contexts from the server itself.

@jfdenise Could you please have a look and approve this?

### DIFF
--- a/cli/src/main/java/org/jboss/as/cli/embedded/ThreadLocalContextSelector.java
+++ b/cli/src/main/java/org/jboss/as/cli/embedded/ThreadLocalContextSelector.java
@@ -26,10 +26,14 @@ import org.jboss.logmanager.LogContext;
 import org.jboss.logmanager.LogContextSelector;
 import org.jboss.stdio.StdioContext;
 import org.jboss.stdio.StdioContextSelector;
+import org.wildfly.security.manager.WildFlySecurityManager;
 
 /**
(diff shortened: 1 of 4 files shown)


jamezp:
Note the `ModuleInfoTestCase.testModuleInfo` consistently fail when a log manager upgrade is done. It's due to the dependency not being found in the repository cache. It can be manually done or the tests can be run the next day after the cache is replicated.

PR #3700, opened 2019-03-07
[WFCORE-4355] Introduce a simple event logging API.

jamezp:
JIRA: https://issues.jboss.org/browse/WFCORE-4355
Proposal: https://github.com/wildfly/wildfly-proposals/pull/181

Introduces a simple event logging API into WildFly Core. This will be able to be used for things like access logging, security audit logging, EJB access logging and management audit logging. The purpose of this API for now is to simply have a unified way of formatting a JSON message and writing to `stdout`.

### DIFF
--- a/core-feature-pack/pom.xml
+++ b/core-feature-pack/pom.xml
@@ -537,6 +537,11 @@
         </dependency>
 
         <dependency>
+            <groupId>org.wildfly.core</groupId>
+            <artifactId>wildfly-event-logger</artifactId>
+        </dependency>
+
+        <dependency>
             <groupId>org.wildfly.openssl</groupId>
             <artifactId>wildfly-openssl-java</artifactId>
         </dependency>
--- a/core-feature-pack/src/license/core-feature-pack-licenses.xml
+++ b/core-feature-pack/src/license/core-feature-pack-licenses.xml
@@ -816,6 +816,17 @@
     </dependency>
     <dependency>
       <groupId>org.wildfly.core</groupId>
+      <artifactId>wildfly-event-logger</artifactId>
+      <licenses>
+        <license>
+          <name>Apache License 2.0</name>
+          <url>https://www.apache.org/licenses/LICENSE-2.0</url>
+          <distribution>repo</distribution>
+        </license>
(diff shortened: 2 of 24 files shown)


jamezp:
I've added the Feature label, however this is an internal only feature and not subject to the full feature process. Though I did create a proposal for discussion purposes.

jamezp:
https://github.com/wildfly/wildfly/pull/12205 should fix the test issues with the full integration tests.

jmesnil:
The org.wildfly.core:wildfly-event-logger should be specified in core-feature-pack/src/license/core-feature-pack-licenses.xml

jamezp:
@jmesnil done. I only added the ASL license as it's all new code. If the LGPL license needs to be there too please just let me know.

jamezp:
@jmesnil Possibly `source` is a bad name. It is currently more like a logging category so more like the `kind` or `type`. For example with the access logging I'm working the `source` is currently "web-access". So essentially the `source` is what is used to distinguish different log types.

I'm definitely open to different names as well. I originally had `eventId`, but that seemed like something that implied it should be unique. Maybe `type` is better though. I attempted to find a common type of name among the various log aggregators with little luck other than `source` seemed to be used quite frequently :)

jamezp:
retest this please

jmesnil:
@jamezp Naming is hard and I can see arguments both for and against `source`, `kind` and `type`.

The important thing is the actual use of this field.
AIUI, if I want to aggregate all access logs, I need to have a common json field I can specify and `source` is likely this field.
I think the doc should provide an example of good practices.
The doc should be updated to provide this use case by using a sensible `source` value (e.g. `access-log`) that can be aggregated instead of a generic non-descriptive one like `my-event`.

jamezp:
@jmesnil I've updated the README doc. Hopefully it makes it more clear now and if not please let me know. I'll be happy to add more content or edit content if needed.

jmesnil:
re the `deps-changed` flag, the only change is a new internal dependency to the `org.wildfly.core:wildfly-event-logger` maven module.

PR #3701, opened 2019-03-11
Explicitly excluse commons-lang3

kabir:
https://issues.jboss.org/browse/WFCORE-4366


kabir:
Adding hold so we don't merge until we decide what to do with the similar commit in https://github.com/jbossas/wildfly-core-eap/pull/656

kabir:
This should now be fine

PR #3703, opened 2019-03-13
[WFCORE-4370] Add the maven.repo.local argument to the surefire setti…

jamezp:
…ngs so JBoss Modules can see the correct repository.

https://issues.jboss.org/browse/WFCORE-4370


bstansberry:
@jfdenise FYI. I recall there was something related to this property that was relevant to galleon layer tests. I believe getting its value into the right process or something like that. So just FYI in case this change is relevant to you.

jamezp:
It makes me wonder if we should just set it in the main testsuite configuration or even the main pom.

bstansberry:
@jamezp The galleon layers thing I was referring to was quite specialized and I believe involved changing the java source code to pass the value through to a process launched by a test class.

PR #3705, opened 2019-03-15
[WFCORE-4374] security-manager permissions are loaded using java.se m…

spyrkob:
…odule

Issue: https://issues.jboss.org/browse/WFCORE-4374

Adding dependency on java.se module to security-manager would allow most of standard permissions to be defined in security-manager permissions configuration without the need to specify module option


spyrkob:
@jmesnil could you take a look please

PR #3707, opened 2019-03-20
[WFCORE-4379] / [WFCORE-4378] / [WFCORE-4259] WildFly Common and Elytron Upgrades

darranl:
And updates related to the modularisation of WildFly Elytron.

https://issues.jboss.org/browse/WFCORE-4259
https://issues.jboss.org/browse/WFCORE-4378
https://issues.jboss.org/browse/WFCORE-4379

Also clean up WildFly Elytron Tool dependency: -

https://issues.jboss.org/browse/WFCORE-4380


pom.xml
```diff
@@ -211,9 +211,8 @@
         <version.org.wildfly.openssl.wildfly-openssl-solaris-x86_64>${version.org.wildfly.openssl.natives}</version.org.wildfly.openssl.wildfly-openssl-solaris-x86_64>
         <version.org.wildfly.openssl.wildfly-openssl-windows-i386>${version.org.wildfly.openssl.natives}</version.org.wildfly.openssl.wildfly-openssl-windows-i386>
         <version.org.wildfly.openssl.wildfly-openssl-windows-x86_64>${version.org.wildfly.openssl.natives}</version.org.wildfly.openssl.wildfly-openssl-windows-x86_64>
-        <version.org.wildfly.security.elytron>1.8.0.Final</version.org.wildfly.security.elytron>
+        <version.org.wildfly.security.elytron>1.9.0.CR2</version.org.wildfly.security.elytron>
         <version.org.wildfly.security.elytron-web>1.4.0.Final</version.org.wildfly.security.elytron-web>
-        <version.org.wildfly.security.elytron.tool>1.6.0.Final</version.org.wildfly.security.elytron.tool>
```
jmesnil: does that mean that Elytron tool has been "put back" with other Elytron modules and will now be updated with the rest of Elytron components?

pom.xml
```diff
@@ -211,9 +211,8 @@
         <version.org.wildfly.openssl.wildfly-openssl-solaris-x86_64>${version.org.wildfly.openssl.natives}</version.org.wildfly.openssl.wildfly-openssl-solaris-x86_64>
         <version.org.wildfly.openssl.wildfly-openssl-windows-i386>${version.org.wildfly.openssl.natives}</version.org.wildfly.openssl.wildfly-openssl-windows-i386>
         <version.org.wildfly.openssl.wildfly-openssl-windows-x86_64>${version.org.wildfly.openssl.natives}</version.org.wildfly.openssl.wildfly-openssl-windows-x86_64>
-        <version.org.wildfly.security.elytron>1.8.0.Final</version.org.wildfly.security.elytron>
+        <version.org.wildfly.security.elytron>1.9.0.CR2</version.org.wildfly.security.elytron>
         <version.org.wildfly.security.elytron-web>1.4.0.Final</version.org.wildfly.security.elytron-web>
```
jmesnil: Elytron Web continue to get its own release cycle, right?

pom.xml
darranl: Yes Elytron Tool is now tagged when we tag Elytron so it is no longer required that we tag it separately so it will always come in via the same component upgrade.

pom.xml
darranl: Once this is in I will create a new tag of Elytron Web that depends on the modularised version - however yes Elytron web will continue to be independent. 

The reason we keep that one independent is it depends on one or more HTTP servers so we don't want to tie all those dependencies to the main Elytron project.

PR #3712, opened 2019-03-24
[WFCORE-4385] DomainTestSupport and DomainLifecycleUtil implement AutoCloseable

bstansberry:
Stopping DomainTestSupport causes it to close its controlled DomainLifecycleUtil objects
Stopping DomainLifecyleUtil closes its shared DomainTestClient.
Deprecate DomainTestSupport.stop() in favor of close() as close has a clearer semantic
Also a lot of cleanup to get rid of IDE warns for these classes.

https://issues.jboss.org/browse/WFCORE-4385

Needs #3709 and #3711 first as those correct invalid uses of DomainTestSupport that the main commits in this PR disallow.  The commits in those PRs are in this one.

### DIFF
--- a/testsuite/domain/src/test/java/org/jboss/as/test/integration/domain/AbstractConfigurationChangesTestCase.java
+++ b/testsuite/domain/src/test/java/org/jboss/as/test/integration/domain/AbstractConfigurationChangesTestCase.java
@@ -89,7 +89,7 @@ public abstract class AbstractConfigurationChangesTestCase {
 
     @AfterClass
     public static void tearDownDomain() throws Exception {
-        testSupport.stop();
+        testSupport.close();
         domainMasterLifecycleUtil = null;
         domainSlaveLifecycleUtil = null;
         testSupport = null;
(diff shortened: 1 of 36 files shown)


testsuite/domain/src/test/java/org/jboss/as/test/integration/domain/AbstractConfigurationChangesTestCase.java
```diff
@@ -89,7 +89,7 @@ public static void setupDomain() throws Exception {
 
     @AfterClass
     public static void tearDownDomain() throws Exception {
-        testSupport.stop();
```
jmesnil: Is there still usage of the `stop()` method after your refactoring?
Should the method visibility be reduced (to protected/package/private)?

bstansberry: It's still used in WildFly full.

jmesnil:
@bstansberry I merged #3709 and #3711 before this one.
Could you please drop their commits from this PR to avoid the conflict, please?

bstansberry:
@jmesnil Done. Thanks!  All of this because one test had too low of a timeout and failed about 1.5% of the time. But the timeout was used in smelly code so down the rabbit hole I went.

PR #3713, opened 2019-03-27
WFCORE-3891: ensure unique names for default formatters

michpetrov:
Issue: [WFCORE-3891](https://issues.jboss.org/browse/WFCORE-3891)

### DIFF
--- a/logging/src/main/java/org/jboss/as/logging/formatters/PatternFormatterResourceDefinition.java
+++ b/logging/src/main/java/org/jboss/as/logging/formatters/PatternFormatterResourceDefinition.java
@@ -57,6 +57,8 @@ public class PatternFormatterResourceDefinition extends TransformerResourceDefin
 
     public static final String NAME = "pattern-formatter";
 
+    public static final String DEFAULT_FORMATTER_SUFFIX = "-wfcore-pattern-formatter";
+
     // Pattern formatter options
     public static final PropertyAttributeDefinition COLOR_MAP = PropertyAttributeDefinition.Builder.of("color-map", ModelType.STRING)
             .setAllowExpression(true)
@@ -111,6 +113,9 @@ public class PatternFormatterResourceDefinition extends TransformerResourceDefin
         @Override
         public void performRuntime(final OperationContext context, final ModelNode operation, final ModelNode model, final LogContextConfiguration logContextConfiguration) throws OperationFailedException {
             final String name = context.getCurrentAddressValue();
+            if (name.endsWith(DEFAULT_FORMATTER_SUFFIX)) {
+                throw LoggingLogger.ROOT_LOGGER.illegalFormatterName();
+            }
             FormatterConfiguration configuration = logContextConfiguration.getFormatterConfiguration(name);
             if (configuration == null) {
                 LoggingLogger.ROOT_LOGGER.tracef("Adding formatter '%s' at '%s'", name, context.getCurrentAddress());
--- a/logging/src/main/java/org/jboss/as/logging/formatters/StructuredFormatterResourceDefinition.java
+++ b/logging/src/main/java/org/jboss/as/logging/formatters/StructuredFormatterResourceDefinition.java
@@ -369,6 +369,9 @@ public abstract class StructuredFormatterResourceDefinition extends TransformerR
             }
 
             final String name = context.getCurrentAddressValue();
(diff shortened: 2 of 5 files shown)


logging/src/main/java/org/jboss/as/logging/formatters/PatternFormatterResourceDefinition.java
```diff
@@ -111,6 +113,9 @@ public boolean isMarshallable(final AttributeDefinition attribute, final ModelNo
         @Override
         public void performRuntime(final OperationContext context, final ModelNode operation, final ModelNode model, final LogContextConfiguration logContextConfiguration) throws OperationFailedException {
             final String name = context.getCurrentAddressValue();
+            if (name.endsWith(DEFAULT_FORMATTER_SUFFIX)) {
+                throw new IllegalArgumentException("Formatter name cannot end with '" + DEFAULT_FORMATTER_SUFFIX + "'");
```
jamezp: This needs to be internationalized and I think an `OperationFailedException` would be a better one to throw.

logging/src/main/java/org/jboss/as/logging/formatters/StructuredFormatterResourceDefinition.java
```diff
@@ -369,6 +369,9 @@ public void performRuntime(final OperationContext context, final ModelNode opera
             }
 
             final String name = context.getCurrentAddressValue();
+            if (name.endsWith(PatternFormatterResourceDefinition.DEFAULT_FORMATTER_SUFFIX)) {
+                throw new IllegalArgumentException("Formatter name cannot end with '" + PatternFormatterResourceDefinition.DEFAULT_FORMATTER_SUFFIX + "'");
```
jamezp: This needs to be internationalized and I think an `OperationFailedException` would be a better one to throw.

logging/src/main/java/org/jboss/as/logging/handlers/HandlerOperations.java
```diff
@@ -650,22 +651,19 @@ private static void handleProperty(final AttributeDefinition attribute, final Op
             if (valueNode.isDefined()) {
                 final String resolvedValue = valueNode.asString();
                 configuration.setFormatterName(resolvedValue);
-                // If the formatter was previously defined by the formatter attribute, remove the formatter
-                if (logContextConfiguration.getFormatterNames().contains(handlerName)) {
-                    logContextConfiguration.removeFormatterConfiguration(handlerName);
-                }
```
jamezp: We still need this to remove the `handlerName + PatternFormatterResourceDefinition.DEFAULT_FORMATTER_SUFFIX`.

logging/src/main/java/org/jboss/as/logging/formatters/PatternFormatterResourceDefinition.java
michpetrov: How do I do that? Should I turn this into a new message in LoggingLogger?

logging/src/main/java/org/jboss/as/logging/handlers/HandlerOperations.java
michpetrov: That is taken care of [here](https://github.com/wildfly/wildfly-core/blob/8839ff61c041f930b4128d1d7351d779a8981159/logging/src/main/java/org/jboss/as/logging/handlers/HandlerOperations.java#L632), isn't it? The test is passing.

logging/src/main/java/org/jboss/as/logging/formatters/PatternFormatterResourceDefinition.java
jamezp: Yes. You add a method to the `LoggingLogger` that returns an `OperationFailedException`.

logging/src/main/java/org/jboss/as/logging/handlers/HandlerOperations.java
jamezp: Only if the `formatter` attribute is processed. So for example if you were to add a handler then write a `named-formatter` attribute, that remove would not be hit.

For example:
```
/subsystem=logging/console-handler=my-console:add
/subsystem=logging/console-handler=my-console:write-attribute(name=named-formatter, value=COLOR-PATTERN)
```

PR #3718, opened 2019-03-27
[WFCORE-4357] Introducing DeploymentPhaseContext.requires() method

ropalka:
https://issues.jboss.org/browse/WFCORE-4357

### DIFF
--- a/server/src/main/java/org/jboss/as/server/deployment/DelegatingSupplier.java
+++ b/server/src/main/java/org/jboss/as/server/deployment/DelegatingSupplier.java
@@ -1,0 +1,57 @@
+/*
+ * JBoss, Home of Professional Open Source.
+ * Copyright 2019, Red Hat, Inc., and individual contributors
+ * as indicated by the @author tags. See the copyright.txt file in the
+ * distribution for a full listing of individual contributors.
+ *
+ * This is free software; you can redistribute it and/or modify it
+ * under the terms of the GNU Lesser General Public License as
+ * published by the Free Software Foundation; either version 2.1 of
+ * the License, or (at your option) any later version.
+ *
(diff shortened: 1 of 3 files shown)


ropalka:
@bstansberry @jmesnil please review guys.

jmesnil:
@ropalka Could you please add a test for this new API please?

jstourac:
Just curious - doesn't it make sense to migrate also (some?) of the services to use this new API? So it is used in the wild also. The https://issues.jboss.org/browse/WFLY-11603 is a good candidate as this whole issue was brought to light by that one... WDYT?

ropalka:
Hello @jstourac 
That was my intention to start migrating existing code to the new API.
Unfortunately the method I deprecated, see

https://github.com/wildfly/wildfly-core/pull/3718/commits/ef65831d6d5470bf3b8e34670a7a53b8855888ca#diff-53b437bb1e5c27312b769ffbac3a9a3cR102

isn't used at WildFly-Core code base but just in WildFly code base.
The new method we introduced can be used in WildFly once it is in WildFly Core and this component ugpraded in WildFly.

@jmesnil are you really sure test for this new simple method is necessary?
It is just migration step to new MSC API and I will migrate all existing code to this new method
once this method is available in WildFly.

jmesnil:
@ropalka given the importance of the DeploymentPhaseContext API, that would not hurt to get a test to (1) make sure the new API is behaving as expected (2) illustrate how it is meant to use.

However, if there are no such deployment test in core, we will have to rely on the WildFly PR to make sure it is working.

ropalka:
Thank you @jmesnil - there's really no test in WildFly Core test suite
that would test e.g. via mocks deployment processors processing.
I hoped I could find such test that could be easily reused to add our test scenario.

PR #3719, opened 2019-03-27
[WFCORE-4392] Migrating CapabilityServiceTarget to new MSC API

ropalka:
https://issues.jboss.org/browse/WFCORE-4392


ropalka:
Please review @bstansberry it is related to your comment https://github.com/wildfly/wildfly/pull/12120#issuecomment-472263079

bstansberry:
@jmesnil fyi. What this is doing is what I was thinking when I filed the JIRA.

@ropalka Can you add a bit of method javadoc to CapabilityServiceTarget? Also its methods declaring they throw IAE is odd. The reason for that is RuntimeCapability.getCapabilityServiceName will throw that, but it's better to explain that in javadoc via @throws IAE the way RuntimeCapability does, rather than having the IAE in the method signature.

PR #3722, opened 2019-04-03
WFCORE-4399 ResourceEntry implementations to not implement equals/hashCode()

pferraro:
https://issues.jboss.org/browse/WFCORE-4399

### DIFF
--- a/controller/src/main/java/org/jboss/as/controller/extension/ExtensionResource.java
+++ b/controller/src/main/java/org/jboss/as/controller/extension/ExtensionResource.java
@@ -248,4 +248,15 @@ public class ExtensionResource implements Resource.ResourceEntry {
     public Set<String> getOrderedChildTypes() {
         return Collections.emptySet();
     }
+
+    @Override
+    public int hashCode() {
+        return this.moduleName.hashCode();
+    }
+
+    @Override
+    public boolean equals(Object object) {
+        if (!(object instanceof ExtensionResource)) return false;
+        return this.moduleName.equals(((ExtensionResource) object).moduleName);
+    }
 }
--- a/controller/src/main/java/org/jboss/as/controller/registry/AbstractModelResource.java
+++ b/controller/src/main/java/org/jboss/as/controller/registry/AbstractModelResource.java
@@ -450,6 +450,17 @@ public abstract class AbstractModelResource extends ResourceProvider.ResourcePro
         public Resource clone() {
            return delegate.clone();
         }
+
+        @Override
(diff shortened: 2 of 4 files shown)


controller/src/main/java/org/jboss/as/controller/extension/ExtensionResource.java
```diff
@@ -248,4 +248,15 @@ public PlaceholderResourceEntry clone() {
     public Set<String> getOrderedChildTypes() {
         return Collections.emptySet();
     }
+
+    @Override
+    public int hashCode() {
+        return this.moduleName.hashCode();
```
jmesnil: why are you not taking `extensionRegistry` into account?

controller/src/main/java/org/jboss/as/controller/registry/AbstractModelResource.java
```diff
@@ -450,6 +450,17 @@ public boolean isProxy() {
         public Resource clone() {
            return delegate.clone();
         }
+
+        @Override
+        public int hashCode() {
+            return this.getPathElement().hashCode();
```
jmesnil: Shouldn't the hashCode take into account the delegate's hashCode()?

It seems to me that 2 instances of `DelegateResource` with the same path element but different `delegate` should not be considered identical.

controller/src/main/java/org/jboss/as/controller/extension/ExtensionResource.java
pferraro: Because the path element of an extension is sufficient to identify uniqueness.

controller/src/main/java/org/jboss/as/controller/registry/AbstractModelResource.java
pferraro: Don't confuse the hashCode/equals implementation of a ResourceEntry with that of a Resource.  A ResourceEntry is a Resource located a specific path.  If I wanted to test equality of a Resource, I would use Resource.equals().  If I want to store a number of ResourceEntry objects within a hash-based collection, then the path element of the ResourceEntry is what identifies it.

{quote}It seems to me that 2 instances of DelegateResource with the same path element but different delegate should not be considered identical.{quote}

Please refer back to the premise of the jira.  If I have a hash-based Set containing the ResourceEntry's of a given child type, and I want to replace one of the entries with a different Resource implementation (typically by decorating the one that exists for a given path), I cannot do this unless the PathElement of the ResourceEntry is what identifies it in the map.

PR #3724, opened 2019-04-04
[WFCORE-3542] / [WFCORE-3832] JDBC Security Realm Enhancements

darranl:
https://issues.jboss.org/browse/WFCORE-3542
https://issues.jboss.org/browse/WFCORE-3832

### DIFF
--- a/elytron/src/main/java/org/wildfly/extension/elytron/JdbcRealmDefinition.java
+++ b/elytron/src/main/java/org/wildfly/extension/elytron/JdbcRealmDefinition.java
@@ -20,8 +20,11 @@ package org.wildfly.extension.elytron;
 
 import static org.wildfly.extension.elytron.Capabilities.SECURITY_REALM_RUNTIME_CAPABILITY;
 import static org.wildfly.extension.elytron.ElytronDefinition.commonDependencies;
+import static org.wildfly.extension.elytron.ElytronDescriptionConstants.BASE64;
 import static org.wildfly.extension.elytron.ElytronDescriptionConstants.BCRYPT_MAPPER;
 import static org.wildfly.extension.elytron.ElytronDescriptionConstants.CLEAR_PASSWORD_MAPPER;
+import static org.wildfly.extension.elytron.ElytronDescriptionConstants.HEX;
+import static org.wildfly.extension.elytron.ElytronDescriptionConstants.MODULAR_CRYPT_MAPPER;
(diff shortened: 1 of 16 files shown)


darranl:
Documentation and test cases are submitted under https://github.com/wildfly/wildfly/pull/12206

elytron/src/main/java/org/wildfly/extension/elytron/JdbcRealmDefinition.java
```diff
@@ -165,34 +156,51 @@ public PasswordKeyMapper toPasswordKeyMapper(OperationContext context, ModelNode
...
-        @Override
-        public ObjectTypeAttributeDefinition getObjectDefinition() {
-            return OBJECT_DEFINITION;
-        }
-
-        @Override
-        public SimpleAttributeDefinition[] getAttributes() {
-            return new SimpleAttributeDefinition[] { PASSWORD, SALT, ITERATION_COUNT};
-        }
+        static final ObjectTypeAttributeDefinition OBJECT_DEFINITION_7_0 = new ObjectTypeAttributeDefinition.Builder(
```
jmesnil: Why can't you add the new attributes on `OBJECT_DEFINITION`?

This feels that the "updated" object definition is called OBJECT_DEFINITION_7_0 while the previous one is named `OBJECT_DEFINITION` (and is `@Deprecated`).
Until now we have always had a single object definition in the subsystem and restrict any "versioning" consideration  to transformer and parser code.
I am wondering why did you update the model this way for this feature?

darranl: In the end I needed to nest this quite deeply due to the hierachy in how the persistent resource definition is pulled together.

I did consider not going for versioned XML parsing but then it would mean configuration valid against one server would be invalid against another even though the declared schema version may be common.

darranl: @jmesnil do you have any further comment on this one?  

I can swap the names over if that is the main convert but as this PR is now coming up to being four weeks old would be good to get it in.

jmesnil: @darranl I asked out of curiosity as I wonder how you will update the code next time you add/modify a parameter. Would you create an `OBJECT_DEFINITION_X`?

My personal preference would be to keep `OBJECT_DEFINITION` for the current definition and renamed the deprecated one to `OBJECT_DEFINITION_6_0`.
However this internal refactoring does not impact the feature from this PR so I'm ok with merging it as it is.

PR #3732, opened 2019-04-08
[WFCORE-4409] Avoid hangs from https://bugs.openjdk.java.net/browse/JDK-8219658

ropalka:
https://issues.jboss.org/browse/WFCORE-4409


ropalka:
Good point @darranl  - fixed.

PR #3735, opened 2019-04-09
[WFCORE-4412] Eliminating CapabilityServiceTarget.addCapability() deprecated method usages

ropalka:
https://issues.jboss.org/browse/WFCORE-4412

### DIFF
--- a/elytron/src/main/java/org/wildfly/extension/elytron/TrivialAddHandler.java
+++ b/elytron/src/main/java/org/wildfly/extension/elytron/TrivialAddHandler.java
@@ -58,12 +58,14 @@ abstract class TrivialAddHandler<T> extends BaseAddHandler {
         this.initialMode = checkNotNullParam("initialMode", initialMode);
     }
 
+    @SuppressWarnings("unchecked")
     @Override
     protected final void performRuntime(OperationContext context, ModelNode operation, Resource resource)
             throws OperationFailedException {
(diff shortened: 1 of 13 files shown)


ropalka:
The commit was excluded as you suggested @darranl

ropalka:
Retest this please

ropalka:
Retest this please

darranl:
@ropalka please don't use the test phrase unless you see a lot of errors as you kick off loads of jobs on CI that already passed.  Instead click on the details link and you can click run at the top right of the results screen to kick off the specific job again.

ropalka:
Thanks for the hint @darranl - will do it since now on.

elytron/src/main/java/org/wildfly/extension/elytron/TrivialAddHandler.java
```diff
@@ -58,12 +58,14 @@
         this.initialMode = checkNotNullParam("initialMode", initialMode);
```
jmesnil: @darranl @ropalka Is it ok to keep this elytron change or should it be handled by another Elytron PR?

darranl: @jmesnil this current change looks Ok to me, the issue was previously @ropalka was removing an 'unused' method that coincidentally a second PR was about to start using.

PR #3725, opened 2019-04-05
Fix for WFCORE-4402, Upgrade CLI to use aesh 2.3

jfdenise:
Fix for: https://issues.jboss.org/browse/WFCORE-4402 and https://issues.jboss.org/browse/WFCORE-4408

- Aesh release contains jansi upgrade needed to fix dll copy on Windows.
- jansi upgrade to the one expected by aesh.

PR #3736, opened 2019-04-10
[WFCORE-4413] Fixing javax.api and javax.sql.api modules BC issues

ropalka:
https://issues.jboss.org/browse/WFCORE-4413

### DIFF
--- a/core-feature-pack/src/main/resources/modules/system/layers/base/javax/api/main/module.xml
+++ b/core-feature-pack/src/main/resources/modules/system/layers/base/javax/api/main/module.xml
@@ -29,5 +29,6 @@
         <module name="java.se" export="true"/>
         <module name="java.xml" export="true"/>
         <module name="java.xml.crypto" export="true"/>
+        <module name="jdk.xml.dom" export="true"/>
     </dependencies>
 </module>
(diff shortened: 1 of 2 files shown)


ropalka:
I analyzed modules before and after the changes and I identified just two backward compatibility issues.
Who's volunteering to review this PR @bstansberry @jmesnil @darranl @kabir @jamezp ?

darranl:
Looks like the failures and hang are the same as in https://github.com/wildfly/wildfly-core/pull/3725 so maybe not related to that component upgrade after all.

darranl:
Aren't some of these modules supposed to be automatically generated in jboss-modules now or are these two not a part of that?

ropalka:
There was implemented an ability to define dependencies on **Jigsaw** aka **JPMS (Java Platform Module System) modules**  in module.xml (JBoss module descriptors). If there's a module referenced in module.xml that has no module.xml definition it is considered to be JPMS module and JBoss Modules implementation delegates to PlatformModuleLoader for not-found modules.
Also note that JPMS system is emulated on top of JDK8. 

To answer your question neither **javax.api** nor **javax.sql.api** represent a JPMS module.
That is also the reason why these modules have been deprecated.
It is necessary to include missing module dependencies there explicitly.

PR #3737, opened 2019-04-10
WFCORE-3298: do not try to modify protected nodes

michpetrov:
Issue: [WFCORE-3298](https://issues.jboss.org/browse/WFCORE-3298)


controller/src/main/java/org/jboss/as/controller/operations/validation/EnumValidator.java
```diff
@@ -220,7 +220,7 @@ public EnumValidator(final Class<E> enumType,  final boolean nullable, final boo
     public void validateParameter(String parameterName, ModelNode value) throws OperationFailedException {
         super.validateParameter(parameterName, value);
         ModelType type = value.getType();
-        if (type == ModelType.STRING ) {
+        if (type == ModelType.STRING && !value.isProtected()) {
```
bstansberry: This if (!value.isProtected()) can only guard the 'value.set(enumValue.toString());' call. Having it here disables the validation. The value being protected doesn't mean it doesn't need validation.

bstansberry:
retest this please

PR #3749, opened 2019-04-17
[WFCORE-4433] Update wildfly-discovery-client to 1.2.0.Final

tadamski:
https://issues.jboss.org/browse/WFCORE-4433


jmesnil:
@tadamski could you please copy or link to the release notes so that we can see what the component upgrade is about?

PR #3750, opened 2019-04-17
[WFCORE-3947] Support SSL Certificate revocation using OCSP

nekdozjam:
https://issues.jboss.org/browse/WFCORE-3947


nekdozjam:
Requires [ELY-1617]https://github.com/wildfly-security/wildfly-elytron/pull/1269

kabir:
@nekdozjam it does not compile

darranl:
This one is a 'pipelined' RFE so will depend on the chain of PRs including a PR to the Elytron project.

elytron/src/main/java/org/wildfly/extension/elytron/SSLDefinitions.java
```diff
@@ -282,9 +283,8 @@
             .setAllowExpression(false)
             .build();
 
-    private static final SimpleAttributeDefinition MAXIMUM_CERT_PATH = new SimpleAttributeDefinitionBuilder(ElytronDescriptionConstants.MAXIMUM_CERT_PATH, ModelType.INT, true)
+    static final SimpleAttributeDefinition MAXIMUM_CERT_PATH = new SimpleAttributeDefinitionBuilder(ElytronDescriptionConstants.MAXIMUM_CERT_PATH, ModelType.INT, true)
             .setAllowExpression(true)
-            .setDefaultValue(new ModelNode(5))
```
jstourac: Why this change?

elytron/src/test/resources/org/wildfly/extension/elytron/tls.xml
```diff
@@ -41,9 +41,16 @@
             <trust-manager name="trust-with-crl" algorithm="SunX509" key-store="jks_store">
                 <certificate-revocation-list path="crl.pem" relative-to="jboss.server.config.dir" maximum-cert-path="2"/>
             </trust-manager>
-            <trust-manager name="trust-with-crl-dp" algorithm="SunX509" key-store="jks_store">
+            <trust-manager name="trust-with-crl-dp" algorithm="PKIX" key-store="ElytronCaTruststore">
```
jstourac: Just curious - why this change?

elytron/src/main/java/org/wildfly/extension/elytron/SSLDefinitions.java
nekdozjam: this is required to retain backward compatibility, as the older configuration has max-cert-path defined inside CRL and now it should be in trust-manager. If this still had default value, there would be no way of knowing if user defined it or if it is the default. (Now you can have max-cert-path either in CRL or trust-manager, but not both)

elytron/src/main/java/org/wildfly/extension/elytron/SSLDefinitions.java
```diff
@@ -294,6 +294,32 @@
...
+
+    static final SimpleAttributeDefinition ONLY_LEAF_CERT = new SimpleAttributeDefinitionBuilder(ElytronDescriptionConstants.ONLY_LEAF_CERT, ModelType.BOOLEAN, true)
+            .setDefaultValue(new ModelNode(false))
+            .setRequired(false)
+            .setRestartAllServices()
+            .build();
+
+    static final ObjectTypeAttributeDefinition OCSP = new ObjectTypeAttributeDefinition.Builder(ElytronDescriptionConstants.OCSP, RESPONDER, PREFER_CRLS)
+            .setRequired(false)
+            .setRestartAllServices()
```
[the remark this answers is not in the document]
nekdozjam: I am a bit confused, as this comment is on OCSP and the XSD references prefer-crls, so I assume you are talking about that. The default is in the code that resolves the attribute value, but I guess it would be better in the attribute definition as you suggested

elytron/src/test/resources/org/wildfly/extension/elytron/tls.xml
nekdozjam: This looks like a mistake, will fix

elytron/src/main/java/org/wildfly/extension/elytron/SSLDefinitions.java
jstourac: Okay, I see now. There is a bit pity that with this change, user does not know what does it mean when it is not configured at all - does it mean unlimited? If this default value is really not possible to set in the code, we should update attribute description at least, though.

Another approach might be to have these attributes defined as separate - e.g.: MAXIMUM_CERT_PATH and MAXIMUM_CERT_PATH_CRL? With this change, we could also mark MAXIMUM_CERT_PATH_CRL attribute definition as deprecated in the CLI which is something we should do anyway I think. Not sure how much additional complexity this approach brings in.

elytron/src/main/java/org/wildfly/extension/elytron/SSLDefinitions.java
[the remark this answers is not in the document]
jstourac: Oh, sorry, my bad - I commented on wrong line. Anyway you decoded that correctly - yes, I was talking about `prefer-crls` attribute. The thing is that specifying default in attribute definition itself will be propagated to the attribute description which is visible to user when using a CLI. As such default value is clearer to user (it not only hidden in the code).

elytron/src/main/java/org/wildfly/extension/elytron/SSLDefinitions.java
```diff
@@ -601,19 +631,68 @@ static ResourceDefinition getTrustManagerDefinition() {
...
+                }
+                if (ocspNode.isDefined()) {
+                    builder.setResponderURI(responderUri);
+                    if (!crlNode.isDefined()) {
+                        builder.setPreferCrls(false);
+                        builder.setNoFallback(true);
+                    } else {
+                        builder.setPreferCrls(preferCrls);
                     }
                 }
```
jstourac: I don't see any `setSoftFail()` method call - has it been missed?

elytron/src/main/java/org/wildfly/extension/elytron/SSLDefinitions.java
nekdozjam: It is set in code to 5, but not in the model, it is later when the service starts and both MAXIMUM_CERT_PATH for TM and CRL are undefined. But I will split them to make the CRL one deprecated

elytron/src/main/java/org/wildfly/extension/elytron/SSLDefinitions.java
```diff
@@ -612,18 +667,79 @@ static ResourceDefinition getTrustManagerDefinition() {
                 };
             }
 
-            private ValueSupplier<TrustManager> createX509CRLExtendedTrustManager(ServiceBuilder<TrustManager> serviceBuilder, OperationContext context, String algorithm, String providerName, InjectedValue<Provider[]> providersInjector, InjectedValue<KeyStore> keyStoreInjector, ModelNode crlNode, String aliasFilter) throws OperationFailedException {
-                String crlPath = PATH.resolveModelAttribute(context, crlNode).asStringOrNull();
-                String crlRelativeTo = RELATIVE_TO.resolveModelAttribute(context, crlNode).asStringOrNull();
-                int certPath = MAXIMUM_CERT_PATH.resolveModelAttribute(context, crlNode).asInt();
+            private ValueSupplier<TrustManager> createX509RevocationExtendedTrustManager(ServiceBuilder<TrustManager> serviceBuilder, OperationContext context, String algorithm, String providerName, InjectedValue<Provider[]> providersInjector, InjectedValue<KeyStore> keyStoreInjector, ModelNode crlNode, ModelNode ocspNode, Integer maxCertPath, String aliasFilter) throws OperationFailedException {
```
jstourac: Just a nitpick, optional - based on the latest change of `X509RevocationExtendedTrustManager` -> `X509RevocationTrustManager`, name of this method could also be updated.

darranl:
Sorry @nekdozjam but you are also the first with a transformer conflict.

elytron/src/main/java/org/wildfly/extension/elytron/SSLDefinitions.java
nekdozjam: Fixed

nekdozjam:
MAXIMUM_CERT_PATH transformer test added

elytron/src/main/java/org/wildfly/extension/elytron/SSLDefinitions.java
nekdozjam: done

elytron/src/main/java/org/wildfly/extension/elytron/SSLDefinitions.java
```diff
@@ -282,14 +284,49 @@
...
             .setValidator(new IntRangeValidator(1))
             .setRestartAllServices()
             .build();
 
-    static final ObjectTypeAttributeDefinition CERTIFICATE_REVOCATION_LIST = new ObjectTypeAttributeDefinition.Builder(ElytronDescriptionConstants.CERTIFICATE_REVOCATION_LIST, PATH, RELATIVE_TO, MAXIMUM_CERT_PATH)
+    @Deprecated
+    static final SimpleAttributeDefinition MAXIMUM_CERT_PATH_CRL = new SimpleAttributeDefinitionBuilder(ElytronDescriptionConstants.MAXIMUM_CERT_PATH, ModelType.INT, true)
+            .setAllowExpression(true)
+            .setValidator(new IntRangeValidator(1))
+            .setDeprecated(ModelVersion.create(7))
```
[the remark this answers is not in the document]
nekdozjam: you are reading trust-manager maximum-cert-path attribute, which is the new one. The deprecated is part of certificate-revocation-list attribute of trust manager. The version should be correct, it is elytron subsystem model version

PR #3751, opened 2019-04-18
Fix for WFCORE-4406 and Upgrade to Galleon 4.0.0.alpha

jfdenise:
F- ix for https://issues.jboss.org/browse/WFCORE-4406
- Upgrade to galleon 4.0 alpha. Final upgrade is tracked by (https://issues.jboss.org/browse/WFCORE-4422)
- Produces FP with Galleon 4.0, generates and deploy wildfly-core offliner. That is a required step to address: https://issues.jboss.org/browse/WFLY-11989
- Make dependencies provided, it is required to not download all dependencies when building User feature-packs from wildfly feature-packs.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -113,10 +113,10 @@
 
         <!-- Non-default maven plugin versions and configuration -->
         <version.jacoco.plugin>0.8.0</version.jacoco.plugin>
-        <version.org.jboss.galleon>3.0.1.Final</version.org.jboss.galleon>
+        <version.org.jboss.galleon>4.0.0.Alpha1</version.org.jboss.galleon>
         <version.org.wildfly.build-tools>1.2.10.Final</version.org.wildfly.build-tools>
         <version.org.wildfly.checkstyle-config>1.0.5.Final</version.org.wildfly.checkstyle-config>
-        <version.org.wildfly.galleon-plugins>3.0.1.Final</version.org.wildfly.galleon-plugins>
+        <version.org.wildfly.galleon-plugins>4.0.0.Alpha4</version.org.wildfly.galleon-plugins>
         <!-- wildfly plugin-->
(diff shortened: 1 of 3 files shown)


kabir:
@psakar @bstansberry FYI this is not labelled with deps-changed as it does not go into the feature packs

pom.xml
```diff
@@ -113,10 +113,10 @@
 
         <!-- Non-default maven plugin versions and configuration -->
         <version.jacoco.plugin>0.8.0</version.jacoco.plugin>
-        <version.org.jboss.galleon>3.0.1.Final</version.org.jboss.galleon>
+        <version.org.jboss.galleon>4.0.0.Alpha1</version.org.jboss.galleon>
         <version.org.wildfly.build-tools>1.2.10.Final</version.org.wildfly.build-tools>
         <version.org.wildfly.checkstyle-config>1.0.5.Final</version.org.wildfly.checkstyle-config>
-        <version.org.wildfly.galleon-plugins>3.0.1.Final</version.org.wildfly.galleon-plugins>
+        <version.org.wildfly.galleon-plugins>4.0.0.Alpha4</version.org.wildfly.galleon-plugins>
```
jmesnil: Was it a coincidence that both org.jboss.galleon & org.wildfly.galleon-plugins were at 3.0.1.Final before?
Is there any relation between the 2 versions?

jfdenise: @jmesnil , the difference in version is due to some issues found in galleon-plugins. The 2 galleon deps will be 4.0.Final

jfdenise:
@jmesnil , I think that it should be ok, @aloubyansky we have 2 weeks before the final release, is it fine with you?

aloubyansky:
Yes, we'll have the final release by that date.

PR #3753, opened 2019-04-23
Revert "[WFCORE-3185] Run parallel boot tasks in coarser grained chunks"

jmesnil:
Reverts wildfly/wildfly-core#2935 as it is causing issues with some Jenkins jobs


jamezp:
retest this please

PR #3755, opened 2019-04-25
[WFCORE-4442] Migrating UndertowHttpManagementService & HttpShutdownService to new MSC API

ropalka:
https://issues.jboss.org/browse/WFCORE-4442

### DIFF
--- a/server/src/main/java/org/jboss/as/server/mgmt/UndertowHttpManagementService.java
+++ b/server/src/main/java/org/jboss/as/server/mgmt/UndertowHttpManagementService.java
@@ -33,6 +33,8 @@ import java.util.Collection;
 import java.util.List;
 import java.util.Set;
 import java.util.concurrent.Executor;
+import java.util.function.Consumer;
+import java.util.function.Supplier;
 
 import javax.net.ssl.SSLContext;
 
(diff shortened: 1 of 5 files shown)


server/src/main/java/org/jboss/as/server/mgmt/UndertowHttpManagementService.java
```diff
@@ -237,17 +266,17 @@ public synchronized void start(StartContext context) throws StartException {
...
-        final NetworkInterfaceBinding interfaceBinding = interfaceBindingValue.getOptionalValue();
-        final NetworkInterfaceBinding secureInterfaceBinding = secureInterfaceBindingValue.getOptionalValue();
+        final SocketBinding basicBinding = socketBindingSupplier != null ? socketBindingSupplier.get() : null;
+        final SocketBinding secureBinding = secureSocketBindingSupplier != null ? secureSocketBindingSupplier.get() : null;
+        final NetworkInterfaceBinding interfaceBinding = interfaceBindingSupplier != null ? interfaceBindingSupplier.get() : null;
+        final NetworkInterfaceBinding secureInterfaceBinding = secureInterfaceBindingSupplier != null ? secureInterfaceBindingSupplier.get() : null;
         if (interfaceBinding != null) {
             useUnmanagedBindings = true;
-            final int port = portValue.getOptionalValue();
+            final int port = this.port != null ? this.port : -1;
```
bstansberry: I think this null check can be changed to an assert this.port != null.  If an interfaceBinding is provided a port must be provided, and the intended user if it, org.jboss.as.host.controller.operations.HttpManagementAddHandler always provides a port.

server/src/main/java/org/jboss/as/server/mgmt/UndertowHttpManagementService.java
```diff
@@ -237,17 +266,17 @@ public synchronized void start(StartContext context) throws StartException {
...
             if (port > 0) {
                 bindAddress = new InetSocketAddress(interfaceBinding.getAddress(), port);
             }
-            final int securePort = securePortValue.getOptionalValue();
+            final int securePort = this.securePort != null ? this.securePort : -1;
```
bstansberry: Same comment as above re 'port'.

server/src/main/java/org/jboss/as/server/mgmt/UndertowHttpManagementService.java
ropalka: Updated accordingly @bstansberry

bstansberry:
@jmesnil This LGTM.

PR #3756, opened 2019-04-25
[WFCORE-4446] Migrating AbsolutePathService & RelativePathService to new MSC API

ropalka:
https://issues.jboss.org/browse/WFCORE-4446

### DIFF
--- a/controller/src/main/java/org/jboss/as/controller/services/path/AbsolutePathService.java
+++ b/controller/src/main/java/org/jboss/as/controller/services/path/AbsolutePathService.java
@@ -25,6 +25,7 @@ package org.jboss.as.controller.services.path;
 import static org.jboss.as.controller.descriptions.ModelDescriptionConstants.PATH;
 
 import java.io.File;
+import java.util.function.Consumer;
 
 import org.jboss.dmr.ModelNode;
 import org.jboss.msc.service.ServiceBuilder;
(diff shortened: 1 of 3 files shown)


controller/src/main/java/org/jboss/as/controller/services/path/AbsolutePathService.java
```diff
@@ -22,50 +22,47 @@
...
-        AbsolutePathService service = new AbsolutePathService(abstractPath);
-        ServiceBuilder<String> builder = serviceTarget.addService(sname, service);
+    public static ServiceController<?> addService(final ServiceName name, final String abstractPath, final ServiceTarget target) {
+        final ServiceBuilder<?> builder = target.addService(name);
+        final Consumer<String> pathConsumer = builder.provides(name);
+        builder.setInstance(new AbsolutePathService(abstractPath, pathConsumer));
         return builder.install();
     }
 
-    public static String convertPath(String abstractPath) {
```
jmesnil: Is this public static method considered as API for this class?

controller/src/main/java/org/jboss/as/controller/services/path/AbsolutePathService.java
```diff
@@ -22,50 +22,47 @@
...
+    static String convertPath(final String abstractPath) {
         Assert.checkNotNullParam("abstractPath", abstractPath);
         Assert.checkNotEmptyParam("abstractPath", abstractPath);
         // Use File.getAbsolutePath() to make relative paths absolute
         File f = new File(abstractPath);
         return f.getAbsolutePath();
     }
 
-    public static void addService(final ServiceName name, final ModelNode element, final ServiceTarget serviceTarget) {
```
jmesnil: same question as above?

controller/src/main/java/org/jboss/as/controller/services/path/PathManagerService.java
```diff
@@ -265,18 +265,12 @@ final void removePathEntry(final String pathName, boolean check) throws Operatio
      * @param pathName the name of the relevant path. Cannot be {@code null}
      * @param path  the value of the path within the model. This is either an absolute path or
      *              the relative portion of the path. Cannot be {@code null}
-     * @param possiblyAbsolute {@code true} if the path may be absolute and a check should be performed before installing
-     *                         a service variant that depends on the service associated with {@code relativeTo}
      * @param relativeTo the name of the path this path is relative to. If {@code null} this is an absolute path
      * @return the service controller for the {@code Service<String>}
      */
     final ServiceController<?> addRelativePathService(final ServiceTarget serviceTarget, final String pathName, final String path,
-                                                      final boolean possiblyAbsolute, final String relativeTo) {
```
jmesnil: This seems like a API breaking change.
The removal of the `possiblyAbsolute` changes the behaviour of the method. Without it, it is no longer possible for callers to get an AbsolutePathService from this method.

ropalka:
All fixed @jmesnil

PR #3761, opened 2019-04-30
[WFCORE-4457] Ensure that any tests that set the JVM wide default SSLContext restore it to the default default.

darranl:
https://issues.jboss.org/browse/WFCORE-4457

### DIFF
--- a/elytron/src/main/java/org/wildfly/extension/elytron/ElytronDefinition.java
+++ b/elytron/src/main/java/org/wildfly/extension/elytron/ElytronDefinition.java
@@ -108,6 +108,13 @@ import org.wildfly.security.manager.action.ReadPropertyAction;
  */
 class ElytronDefinition extends SimpleResourceDefinition {
 
+    /**
+     * System property which if set to {@code true} will cause the JVM wide default {@link SSLContext} to be restored when the subsystem shuts down.
+     *
+     * This property is only for use by test cases.
+     */
+    static final String RESTORE_DEFAULT_SSL_CONTEXT = ElytronDefinition.class.getPackage().getName() + ".restore-default-ssl-context";
+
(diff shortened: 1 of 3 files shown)


elytron/src/main/java/org/wildfly/extension/elytron/ElytronDefinition.java
```diff
@@ -108,6 +108,11 @@
  */
 class ElytronDefinition extends SimpleResourceDefinition {
 
+    /**
+     * System property which if set to {@code true} will cause the JVM wide default {@link SSLContext} to be restored when the subsystem shuts down.
+     */
+    public static final String RESTORE_DEFAULT_SSL_CONTEXT = ElytronDefinition.class.getPackage().getName() + ".restore-default-ssl-context";
```
bstansberry: Can this be package protected and the javadoc note that it's for internal test use only?

The package protected is just me being anal since the whole class is.

darranl:
@bstansberry I have made the changes you requested.

elytron/src/main/java/org/wildfly/extension/elytron/ElytronDefinition.java
```diff
@@ -108,6 +108,13 @@
  */
 class ElytronDefinition extends SimpleResourceDefinition {
 
+    /**
+     * System property which if set to {@code true} will cause the JVM wide default {@link SSLContext} to be restored when the subsystem shuts down.
+     *
+     * This property is only for use by test cases.
+     */
+    static final String RESTORE_DEFAULT_SSL_CONTEXT = ElytronDefinition.class.getPackage().getName() + ".restore-default-ssl-context";
```
jamezp: More of a personal preference than anything, but I'd rather just see the whole string hard-coded. If for whatever reason this class gets refactored it may unknowingly break things.

darranl:
Just kicked off the Security Manager job again, I don't think these changes should affect it but it did add some doPrivileged calls so is possible

PR #3770, opened 2019-05-08
[WFCORE-4460] jboss-cli.sh doesn't return json when the output command is 'failed'

iweiss:
Issue: https://issues.jboss.org/browse/WFCORE-4460


cli/pom.xml
```diff
@@ -206,6 +206,11 @@
             <artifactId>junit</artifactId>
             <scope>test</scope>
         </dependency>
+        <dependency>
+            <groupId>org.glassfish</groupId>
+            <artifactId>javax.json</artifactId>
+            <scope>test</scope>
+        </dependency>
```
jamezp: What's this dep needed for? It looks like DMR is used to output JSON.

iweiss: This dependency is used on the test to parse the output of the DMR as a correctly formed JSON.

jamezp: Sorry I missed that part.

PR #3773, opened 2019-05-10
WFCORE-2252: reject attribute configuration that is both required and…

michpetrov:
… has a default

### DIFF
--- a/controller/src/main/java/org/jboss/as/controller/AttributeDefinition.java
+++ b/controller/src/main/java/org/jboss/as/controller/AttributeDefinition.java
@@ -134,6 +134,10 @@ public abstract class AttributeDefinition {
         if (defaultValue != null && defaultValue.isDefined()) {
             this.defaultValue = defaultValue;
             this.defaultValue.protect();
+            if (this.required) {
+                // WFCORE-2252. Incorrect combination from the code author
+                throw new IllegalArgumentException("Attribute " + name + " uses a combination of a default value and a required flag. At most one of those should be set.");
+            }
         } else {
             this.defaultValue = null;
         }


darranl:
If this change is applied any subsystems that are added to a server distribution at a later point would subsequently break - it could be safer to log a warning encouraging a bug report to be raised.

io/subsystem/src/main/java/org/wildfly/extension/io/OutboundBindAddressResourceDefinition.java
```diff
@@ -49,8 +49,7 @@
         .setRestartAllServices()
         .build();
 
-    static final SimpleAttributeDefinition BIND_ADDRESS = new SimpleAttributeDefinitionBuilder("bind-address", ModelType.STRING)
-        .setRequired(true)
+    static final SimpleAttributeDefinition BIND_ADDRESS = new SimpleAttributeDefinitionBuilder("bind-address", ModelType.STRING, true)
```
darranl: According to the XML schema this attribute is required: -

        <xs:attribute name="bind-address" type="xs:string" use="required">
            <annotation xmlns="http://www.w3.org/2001/XMLSchema">
                <documentation>
                    The bind address to use if the destination address matches.
                </documentation>
            </annotation>
        </xs:attribute>

By making this optional could mean the attribute is not written leaving the configuration with an error if validated against the schema.

darranl:
I don't think there should be an automatic assumption that having a default defined in the model does mean an attribute is optional, it could be possible that the definition of a default is the invalid part.

For any subsystem changes we will then need to ask the leads for each subsystem to double check the changes that relate to their subsystem.

darranl:
One related point, if following on from work from another engineer it is also generally a good practice to keep their commit in tact and then add your own commit so the changes are correctly attributed.

There is no requirement for a 1:1 relationship between commits and PRs provided they are suitable units of work.

io/subsystem/src/main/java/org/wildfly/extension/io/OutboundBindAddressResourceDefinition.java
jamezp: IMO that's up to the marsharaller and that is okay to make it optional in the model since there is a default value.

jamezp:
The problem is if you don't make it optional the default value is never used as during an add operation the attribute will be required to be defined. This makes the default value only useful for documentation purposes which doesn't make much sense to me.

darranl:
@jamezp +1 on this being something that needs fixing but there are two possible fixes: -

1. Make all attributes with a default optional.
2. Make remove all default values from optional attributes.

IMO that decision should be checked on a subsystem by subsystem basis.

io/subsystem/src/main/java/org/wildfly/extension/io/OutboundBindAddressResourceDefinition.java
darranl: @jamezp possibly but has anyone checked that the marshaller will always marshal the default here?

jamezp: I've not checked in this specific case no. IIRC the default is `true` though. I could be wrong and we'd definitely need to check that out first.

jamezp:
I do agree this should involved the subsystem author for sure @darranl. I also agree on both points too with a bias of 1.

darranl:
This PR needs a rebase but I would still recommend splitting it into PRs for distinct subsystems so each can have the appropriate review.

controller/src/main/java/org/jboss/as/controller/AttributeDefinition.java
```diff
@@ -134,6 +134,9 @@ private AttributeDefinition(String name, String xmlName, final ModelNode default
         if (defaultValue != null && defaultValue.isDefined()) {
             this.defaultValue = defaultValue;
             this.defaultValue.protect();
+            if (this.required) {
+                throw new IllegalArgumentException(); // WFCORE-2252. Incorrect combination from the code author
```
darranl: This error should be updated to include a message, without a message a developer will always be required to identify the line number the exception was thrown from and check the code to identify the cause.

darranl:
@michpetrov Are you able to continue with the changes discussed so far in review or close this one for now and re-raise when ready?

michpetrov:
@darranl I'll check the current situation but I think I can continue; I'll keep this PR for the change in controller and split the rest off

PR #3778, opened 2019-05-16
[WFCORE-4400] Validation of 'host-context-map' attribute of 'server-s…

Skyllarr:
…sl-sni-context' should allow regular expressions

Jira: [WFCORE-4400](https://issues.jboss.org/browse/WFCORE-4400)

### DIFF
--- a/elytron/src/main/java/org/wildfly/extension/elytron/SSLDefinitions.java
+++ b/elytron/src/main/java/org/wildfly/extension/elytron/SSLDefinitions.java
@@ -62,6 +62,7 @@ import java.util.concurrent.atomic.AtomicBoolean;
 import java.util.concurrent.atomic.AtomicReference;
 import java.util.function.BooleanSupplier;
 import java.util.regex.Pattern;
+import java.util.regex.PatternSyntaxException;
 
 import javax.net.ssl.KeyManager;
 import javax.net.ssl.KeyManagerFactory;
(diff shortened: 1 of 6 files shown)


Skyllarr:
@darranl This pattern does not have any false negatives, as I only restricted first character, last character and multiple dots next to each other.  Hopefully the regular expression is readable with the comments next to it. 

Some notes just in case:
Since the input can be regular expression, if the user means dot as a delimeter (and not 'any character'), it is necessary to escape it with backslash. Double backslash is necessary when using CLI, because CLI removes one. This validation will not allow anything besides dot to be escaped, because it would not make sense for the hostnames (since hostnames can contain only letters, digits 0-9, hyphens and dots).

If this regular expression is too long, maybe it would make sense to remove the validation that was added by [WFCORE-4309](https://issues.jboss.org/browse/WFCORE-4309) completely. But it depends on priorities, as this way at least some feedback is provided to the user. If we decide to keep the validation, I can add some tests for this. Thanks!

darranl:
@Skyllarr Just to double check, this change now opens the validation back up so previously acceptable and valid values will be accepted?

Skyllarr:
@darranl Yes, I believe all valid values will be accepted after this change. 

just a note,  but I also changed value in tests from `*.server` to `.*\.server`. The old value was never valid with current implementation, because regex cannot start with asterisk. If value like this was passed to `server-ssl-sni-context` it would result in exception

```
WFLYCTL0080: Failed services" => {"org.wildfly.security.ssl-context.sniSSC" => "Failed to start service
    Caused by: java.util.regex.PatternSyntaxException: Dangling meta character '*' near index 0
```

PR #3782, opened 2019-05-22
[WFCORE-944] Allow more file-based truststores than just JKS

stahlbock:
With this PR, the decision between ProviderTrustManagerService or FileTrustManagerService is handled the same way it is done for ProviderKeyManagerService/FileKeyManagerService, allowing for more types than only JKS.


darranl:
Superseded by https://github.com/wildfly/wildfly-core/pull/3841

PR #3784, opened 2019-05-23
[WFCORE-4487] Migrate domain-management maven module to new MSC API

ropalka:
https://issues.jboss.org/browse/WFCORE-4487

### DIFF
--- a/domain-management/src/main/java/org/jboss/as/domain/management/security/SecurityRealmService.java
+++ b/domain-management/src/main/java/org/jboss/as/domain/management/security/SecurityRealmService.java
@@ -47,6 +47,8 @@ import java.util.Map.Entry;
 import java.util.function.Function;
 import java.util.Set;
 import java.util.TreeSet;
+import java.util.function.Consumer;
+import java.util.function.Supplier;
 import java.util.stream.Collectors;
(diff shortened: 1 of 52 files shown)


darranl:
@jmesnil / @bstansberry Do you have any thoughts on this one?  Looks reasonable to me, I am thinking get this in now so we have a chance to react to any regressions rather than at the end of WF18.

domain-management/src/main/java/org/jboss/as/domain/management/security/SecurityRealmService.java
```diff
@@ -117,7 +118,9 @@
  * requiring any of the capabilities provided by the realm.
  *
  * @author <a href="mailto:darran.lofthouse@jboss.com">Darran Lofthouse</a>
+ * @author <a href="mailto:ropalka@redhat.com">Richard Opalka</a>
  */
+// TODO: refactor org.jboss.msc.service.Service to org.jboss.msc.Service when WildFly codebase is fixed
```
bstansberry: We can't change anything in this one that would break a subsystem until WF 23. External subsystems can depend on this class and we can't break them.

PR #3793, opened 2019-05-27
use default charset for cli --properties file

andrew--i:
jboss-cli can read only ISO 8859-1 properties files via --properties arg
this patch fix it

### DIFF
--- a/cli/src/main/java/org/jboss/as/cli/impl/CliLauncher.java
+++ b/cli/src/main/java/org/jboss/as/cli/impl/CliLauncher.java
@@ -28,6 +28,7 @@ import java.io.FileInputStream;
 import java.io.FileNotFoundException;
 import java.io.FileReader;
 import java.nio.charset.Charset;
+import java.nio.charset.StandardCharsets;
 import java.util.ArrayList;
 import java.util.Collections;
 import java.util.List;
@@ -67,6 +68,7 @@ public class CliLauncher {
             boolean connect = false;
             boolean version = false;
             int connectionTimeout = -1;
+            Charset propertiesEncoding = StandardCharsets.UTF_8;
 
             final CommandContextConfiguration.Builder ctxBuilder = new CommandContextConfiguration.Builder();
             ctxBuilder.setErrorOnInteract(errorOnInteract);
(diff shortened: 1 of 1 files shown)


kabir:
@wildfly/prod Test mention

bstansberry:
This is ok to test.

@jfdenise Can you have a look at this?

@andrew--i If Jeff has no problem with the idea, can you file a JIRA in https://issues.jboss.org/browse/WFCORE and reference its number in the commit message? Thanks.

cli/src/main/java/org/jboss/as/cli/impl/CliLauncher.java
```diff
@@ -21,11 +21,8 @@
  */
 package org.jboss.as.cli.impl;
 
-import java.io.BufferedReader;
-import java.io.File;
-import java.io.FileInputStream;
-import java.io.FileNotFoundException;
-import java.io.FileReader;
+import java.io.*;
```
bstansberry: Please restore the non-wildcard imports as that's the WildFly codestyle.

andrew--i: done

andrew--i:
> This is ok to test.
> 
> @jfdenise Can you have a look at this?
> 
> @andrew--i If Jeff has no problem with the idea, can you file a JIRA in https://issues.jboss.org/browse/WFCORE and reference its number in the commit message? Thanks.

done, WFCORE-4505

jfdenise:
@andrew--i , good catch. I am wandering if we shouldn't use UTF-8 instead of the default that is bound to the platform. Long term we should decode all files in CLI using UTF-8 as the default and offer a way to set the encoding.

jamezp:
+1 for using `StandardCharsets.UTF_8`

andrew--i:
> @andrew--i , good catch. I am wandering if we shouldn't use UTF-8 instead of the default that is bound to the platform. Long term we should decode all files in CLI using UTF-8 as the default and offer a way to set the encoding.

I made changes

jfdenise:
@andrew--i , I think that we shouldn't introduce a new property with your change. It requires a proper RFE (doc, help, ...) and should be applied to various places. Could you just fix the properties file? Thank-you.

andrew--i:
> @andrew--i , I think that we shouldn't introduce a new property with your change. It requires a proper RFE (doc, help, ...) and should be applied to various places. Could you just fix the properties file? Thank-you.

done

jmesnil:
@andrew--i thanks

PR #3797, opened 2019-05-31
[WFCORE-4489] Wrap setting the JAVA_OPTS in quotes for the batch script.

jamezp:
https://issues.jboss.org/browse/WFCORE-4489

The first commit is the actual changes needed. 

The second commit is not required and will add time to the overall test run. For me this seems to add about 2 minutes which doesn't seem significant, however it's not required and we have concerns I can remove the commit. It essentially runs each test twice, 4 times in the `StandaloneScriptTestCase` with GC logging enabled, in 3 different directories. The hope is to catch issues with scripts and spaced directories sooner rather than later.


bstansberry:
+1 to tests.

PR #3798, opened 2019-05-31
[WFCORE-4440] Changes made via management operations in static-discovery are not reflec…

tmiyargi:
…ted in host.xml

Issue: https://issues.jboss.org/browse/WFCORE-4440

### DIFF
--- a/host-controller/src/main/java/org/jboss/as/host/controller/operations/StaticDiscoveryWriteAttributeHandler.java
+++ b/host-controller/src/main/java/org/jboss/as/host/controller/operations/StaticDiscoveryWriteAttributeHandler.java
@@ -1,0 +1,89 @@
+/*
+ * JBoss, Home of Professional Open Source.
+ * Copyright 2019, Red Hat, Inc., and individual contributors
+ * as indicated by the @author tags. See the copyright.txt file in the
+ * distribution for a full listing of individual contributors.
+ *
+ * This is free software; you can redistribute it and/or modify it
+ * under the terms of the GNU Lesser General Public License as
+ * published by the Free Software Foundation; either version 2.1 of
+ * the License, or (at your option) any later version.
+ *
+ * This software is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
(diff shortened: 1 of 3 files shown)


host-controller/src/main/java/org/jboss/as/host/controller/operations/StaticDiscoveryWriteAttributeHandler.java
```diff
@@ -0,0 +1,96 @@
...
+    }
+
+    @Override
+    protected void finishModelStage(OperationContext context, ModelNode operation, String attributeName, ModelNode newValue,
+            ModelNode oldValue, Resource model) throws OperationFailedException {
+        super.finishModelStage(context, operation, attributeName, newValue, oldValue, model);
+
+        final PathAddress operationAddress = PathAddress.pathAddress(operation.get(OP_ADDR));
+        final PathAddress discoveryOptionsAddress = operationAddress.subAddress(0, operationAddress.size() - 1);
+        final ModelNode discoveryOptions = Resource.Tools.readModel(context.readResourceFromRoot(discoveryOptionsAddress));
```
bstansberry: I believe this can just be:

final ModelNode discoveryOptions = context.readResourceFromRoot(discoveryOptionsAddress, false).getModel();

The 'false' param is important because it saves reading children of the resource, which is unnecessary.

host-controller/src/main/java/org/jboss/as/host/controller/operations/StaticDiscoveryWriteAttributeHandler.java
```diff
@@ -0,0 +1,96 @@
...
+
+        // Get the current list of discovery options and set new value for attribute
+        final ModelNode currentList = discoveryOptions.get(ModelDescriptionConstants.OPTIONS);
```
bstansberry: If you get rid of the Resource.Tools.readModel per my comment above you should clone this though; otherwise you are manipulating the underlying model. Resource.Tools.readModel clones the whole thing internally.

final ModelNode currentList = discoveryOptions.get(ModelDescriptionConstants.OPTIONS).clone();

PR #3801, opened 2019-06-03
WFCORE-4502 Unify line-endings of bat scripts (regression against WF15)

rhusar:
Jira
https://issues.jboss.org/browse/WFCORE-4502

WF PR
https://github.com/wildfly/wildfly/pull/12336

### DIFF
--- a/.gitattributes
+++ b/.gitattributes
@@ -1,0 +1,2 @@
+*.bat    text eol=crlf
+*.ps1    text eol=crlf


jamezp:
One thing to note is if want the current files to all be CRLF we're going to have to manually change the files first. Just some kind of change to trigger git to change the line endings.

rhusar:
@jamezp The assumption is that the release build script performs a clean checkout every time thus this ensures files are CRLF on checkout.

It does not matter what they are in the source tree. But yes, that can be done too, but its not necessary.

jamezp:
What do you mean by "clean checkout"? I know for sure when I do a release it's just based on master. I definitely don't re-checkout the project.

rhusar:
@jamezp The maven release plugin does this for you automatically in `release:perform` goal. It downloads the tag you have just tagged from SCM in the `release:prepare` goal. Thus, when ti does the checkout, it will checkout those files according to `.gitattributes` file, which will set CRLF line endings regardless of the OS you are running for the release.

*well, unless somebody sets use `localCheckout` to true :-) https://maven.apache.org/plugins-archives/maven-release-plugin-2.3.2/perform-mojo.html#localCheckout

jamezp:
Ah well as far as I know none of us that do releases use the release plugin :). I know for sure I don't. I don't think we should block this PR based on this BTW. I do think it's a good thing to have configured.

rhusar:
@jamezp That's a bit surprising to me. What do you use to release then?

jamezp:
Just manually commit it. This is basically the release process https://developer.jboss.org/wiki/WildFlyCoreReleaseProcess.

rhusar:
@jamezp Ouch. In that case I would change the files in the source tree as well... I will update tomorrow.

rhusar:
@jamezp I have updated the PR to change the files that are supposed to use CRLF line endings in the source tree as well as `.gitattributes`. These need to be separate commits (since if they were a single commit, git looks at the attributes and automatically fixes line endings thus won't commit the files into the source tree). 

Note that this is not only about scripts that we package but also about scripts that are not-packaged, e.g. mvn wrapper for windows. Currently this is extremely messy, for instance if you look at `mvnw.cmd`:
```
[rhusar@ribera wildfly-core]$ file mvnw.cmd
mvnw.cmd: DOS batch file text, ASCII text, with CRLF, LF line terminators
```
it mixes both CRLF and LF line terminators. This should resolve all these problems.

darranl:
@jamezp I think you had the most comments on this one - are your concerns now resolved?  If so I will bundle this with the PRs I am about to push through CI to merge.

jamezp:
@darranl I'm fairly indifferent about it, but I don't think it hurts anything. It won't fix the issue for anyone that has the project already checked out without doing something like `find feature-pack/src -name "*.bat" | xargs -I {} git reset-file "{}"`. However for any new checkouts or changes to the file that require a local update it should fix those.

PR #3810, opened 2019-06-11
[WFCORE-4516] The expression for secret value under server-identities in test-security-realm does not work 

iingawal:
https://issues.jboss.org/browse/WFCORE-4516


iingawal:
I am closing this PR , I will raise new PR with some added changes.

PR #3812, opened 2019-06-11
Await stability of added services before completing creation of Abstr…

bstansberry:
…actKernelServicesImpl

HOLD; just to see what happens


jamezp:
Replaced by #3813

PR #3814, opened 2019-06-12
[WFCORE-4516] The expression for secret value under server-identities in test-security-realm does not work

iingawal:
Issue - https://issues.jboss.org/browse/WFCORE-4516

### DIFF
--- a/domain-management/src/main/java/org/jboss/as/domain/management/security/SecurityRealmAddHandler.java
+++ b/domain-management/src/main/java/org/jboss/as/domain/management/security/SecurityRealmAddHandler.java
@@ -62,6 +62,8 @@ import java.util.List;
 import java.util.Map;
 import java.util.Set;
 import java.util.function.Consumer;
+import java.util.regex.Matcher;
+import java.util.regex.Pattern;
 
 import javax.net.ssl.SSLContext;
 import javax.net.ssl.TrustManager;
@@ -779,7 +781,15 @@ public class SecurityRealmAddHandler extends AbstractAddStepHandler {
         ServiceName secretServiceName = SecretIdentityService.ServiceUtil.createServiceName(realmName);
 
         ModelNode resolvedValueNode = SecretServerIdentityResourceDefinition.VALUE.resolveModelAttribute(context, secret);
-        boolean base64 = secret.get(SecretServerIdentityResourceDefinition.VALUE.getName()).getType() != ModelType.EXPRESSION;
+
+        boolean base64=true;
+        if(secret.get(SecretServerIdentityResourceDefinition.VALUE.getName()).getType() == ModelType.EXPRESSION) {
+           String pattern = "^([A-Za-z0-9+/]{4})*([A-Za-z0-9+/]{4}|[A-Za-z0-9+/]{3}=|[A-Za-z0-9+/]{2}==)$";
+           Pattern p = Pattern.compile(pattern);
+           Matcher m = p.matcher(resolvedValueNode.asString());
+           base64 =(m.find())?true:false;
+        }
+
         SecretIdentityService sis;
         if (secret.hasDefined(CredentialReference.CREDENTIAL_REFERENCE)) {
             sis = new SecretIdentityService(resolvedValueNode.asString(), false);
--- a/domain-management/src/main/java/org/jboss/as/domain/management/security/SecurityRealmAddHandler.java
+++ b/domain-management/src/main/java/org/jboss/as/domain/management/security/SecurityRealmAddHandler.java
@@ -783,6 +783,11 @@ public class SecurityRealmAddHandler extends AbstractAddStepHandler {
         ModelNode resolvedValueNode = SecretServerIdentityResourceDefinition.VALUE.resolveModelAttribute(context, secret);
 
         boolean base64=true;
+
+        /*
+          It will only go through this "if" block if value of secret is expression and then it is confirming whether the string "resolvedValueNode.asString()" is base64 or not with the help of pattern mentioned below .As per base64 , the count of characters should be multiple of 4 and the characters should be from this list - (A-Z ,a-z ,0-9,+,/) .In a set of 4 characters if only 3 characters are from that list then the 4th cahracter should be "=" , or if only 2 character are from that list then 3rd and 4th shold be "=" . 
+        */
+
         if(secret.get(SecretServerIdentityResourceDefinition.VALUE.getName()).getType() == ModelType.EXPRESSION) {
            String pattern = "^([A-Za-z0-9+/]{4})*([A-Za-z0-9+/]{4}|[A-Za-z0-9+/]{3}=|[A-Za-z0-9+/]{2}==)$";
            Pattern p = Pattern.compile(pattern);


bstansberry:
This is ok to test.

domain-management/src/main/java/org/jboss/as/domain/management/security/SecurityRealmAddHandler.java
```diff
@@ -779,7 +778,7 @@ private void addSecretService(OperationContext context, ModelNode secret, String
         ServiceName secretServiceName = SecretIdentityService.ServiceUtil.createServiceName(realmName);
 
         ModelNode resolvedValueNode = SecretServerIdentityResourceDefinition.VALUE.resolveModelAttribute(context, secret);
-        boolean base64 = secret.get(SecretServerIdentityResourceDefinition.VALUE.getName()).getType() != ModelType.EXPRESSION;
+        boolean base64 = true;
```
bstansberry: This seems like a change that breaks compatibility.  AIUI the issue is that the attribute description, and perhaps some other docs, state that the value of the attribute is Base64 encoded, but the logic assumes that the value from a resolved expression would be plain text. Which is a reasonable expectation, as encoding a password in Base64 brings no real security value, whereas putting it in plain text in a vault and resolving it from a vault would be standard vault usage.

Regardless of what the attribute description and the docs say, this behavior has been in the code since at least 2013. I don't think we can just change it to require a Base64 encoded value. No one using an expression would have Base64 encoded the resolved value as it wouldn't have worked.

*Perhaps* we could make base64 as Boolean, not a boolean and pass 'null' if it was an expression. Then SecretIdentityService *tries* to Base64 decode, and just uses the plain text if that fails.  But even then you could have a corner case breaking change where a existing plain text value just happened to be valid Base64 text.

iingawal: If the value of expression is not base64 then anyways it will throw error like below at the starting of JBoss and deployment will be failed . 

----------------------------------------
2019-06-18 16:50:54,966 ERROR [org.jboss.msc.service.fail] (MSC service thread 1-6) MSC000001: Failed to start service org.wildfly.core.management.security.realm.ejb-security-realm.secret: org.jboss.msc.service.StartException in service org.wildfly.core.management.security.realm.ejb-security-realm.secret: Failed to start service 


2019-06-18 16:50:57,296 ERROR [org.jboss.as.controller.management-operation] (Controller Boot Thread) WFLYCTL0013: Operation ("deploy") failed - address: ([("deployment" => "TestEJb_RemoteOutbound.war")]) - failure description: {
.
.
2019-06-18 16:50:57,299 ERROR [org.jboss.as.controller.management-operation] (Controller Boot Thread) WFLYCTL0013: Operation ("add") failed - address: ([
    ("core-service" => "management"),
    ("security-realm" => "ejb-security-realm")
----------------------------------------


However , I updated the pull request

domain-management/src/main/java/org/jboss/as/domain/management/security/SecurityRealmAddHandler.java
```diff
@@ -779,7 +781,15 @@ private void addSecretService(OperationContext context, ModelNode secret, String
         ServiceName secretServiceName = SecretIdentityService.ServiceUtil.createServiceName(realmName);
 
         ModelNode resolvedValueNode = SecretServerIdentityResourceDefinition.VALUE.resolveModelAttribute(context, secret);
-        boolean base64 = secret.get(SecretServerIdentityResourceDefinition.VALUE.getName()).getType() != ModelType.EXPRESSION;
+
+        boolean base64=true;
+        if(secret.get(SecretServerIdentityResourceDefinition.VALUE.getName()).getType() == ModelType.EXPRESSION) {
+           String pattern = "^([A-Za-z0-9+/]{4})*([A-Za-z0-9+/]{4}|[A-Za-z0-9+/]{3}=|[A-Za-z0-9+/]{2}==)$";
```
darranl: I think the remainder of the discussion needs to continue but for any place where we add a regular expression can we please add a comment block documenting how the pattern is constructed and if relevant why.

Retrospectively working on a regular expression without the context that led to it's development can be a very painful process.

domain-management/src/main/java/org/jboss/as/domain/management/security/SecurityRealmAddHandler.java
darranl: Is there no more error logged.

If more if available it is often helpful to just paste the complete error, often the exception reported contains a simple error message however the detail is really within the nested causes.

bstansberry: @iingawal I don't understand how org.jboss.as.domain.management.security.SecretIdentityService.start(SecretIdentityService.java:77 is reached if the of the SecretServerIdentityResourceDefinition.VALUE attribute was an expression.  With the change in this PR it will fail, but I don't see how it would have before.

bstansberry: Perhaps my last comment is out of date though as I see now you've added code to check for Base64.

darranl:
I think one final point we are still stuck on is how these values are coming back as Base64 at all - after resolving a vault expression they should be clear text only.

iingawal:
> I think one final point we are still stuck on is how these values are coming back as Base64 at all - after resolving a vault expression they should be clear text only.

- The string "ZWpidXNlcjEyMyE=" is a base64 encoded , and we saved it in expression "password"  , so after resolving the expression i.e system property "password" (I believe it is not a vault expression) , we get value "ZWpidXNlcjEyMyE=" which is base64 encoded .
`
<system-properties>
<property name="password" value="ZWpidXNlcjEyMyE="/>
</system-properties>

<security-realm name="test-security-realm">
<server-identities>
<!secret value="ZWpidXNlcjEyMyE="/>
<secret value="${password}"/>
</server-identities>
</security-realm>
`

darranl:
Reviewing this further we do not feel this is a change that can be accepted into WildFly Core.

The component being discussed here is one of the original components of JBoss AS 7 - this component was written so that if the password was contained in the configuration it would be encoded using Base64 to prevent casual observation of the password however if it is resolved using an expression it is expected that value is clear text.

Making a change like this has the potential for causing problems for existing installations already out there, should their password once resolved using an expression appear to look like it is Base64 encoded these new changes will attempt to convert from Base64.  The only option they would then have would be to change their existing password.  As this password is used for process to process authentication a randomly generated value which looks like Base64 is a real possibility.

Overall changes in this component would need to be handled in an RFE, however this component is deprecated and due to be released in the future so RFEs in this area are not a priority.

darranl:
I am going to close this one for now, realistically as I said before a change like this is risking bringing in backwards compatibility issues.

Maybe come and discuss this further in Zulip but TBH I don't really see alternative options other than adding more configuration but as this resource is deprecated that is not really advisable.

It was intended from the outset that resolved expressions are clear test as it was only intended to be used with vault exposed Strings.

PR #3815, opened 2019-06-12
 [WFCORE-4172] Add support for TLS 1.3

fjuma:
https://issues.jboss.org/browse/WFCORE-4172
https://issues.jboss.org/browse/WFCORE-4391
https://issues.jboss.org/browse/EAP7-1022

This also contains the Elytron subsystem version bump included in #3777

To pass CI, this will require an Elytron upgrade but this is ready for the subsystem changes to be reviewed.

Analysis document: https://github.com/wildfly/wildfly-proposals/pull/148
WildFly PR with documentation and test updates: https://github.com/wildfly/wildfly/pull/12361
Depends on https://github.com/wildfly-security/wildfly-elytron/pull/1295
Also depends on https://issues.jboss.org/browse/WFWIP-160

### DIFF
--- a/elytron/src/main/java/org/wildfly/extension/elytron/SSLDefinitions.java
+++ b/elytron/src/main/java/org/wildfly/extension/elytron/SSLDefinitions.java
@@ -213,6 +213,14 @@ class SSLDefinitions {
             .setDefaultValue(new ModelNode("DEFAULT"))
             .build();
 
+    static final SimpleAttributeDefinition CIPHER_SUITE_NAMES = new SimpleAttributeDefinitionBuilder(ElytronDescriptionConstants.CIPHER_SUITE_NAMES, ModelType.STRING, true)
+            .setAllowExpression(true)
+            .setMinSize(1)
+            .setRestartAllServices()
+            .setValidator(new CipherSuiteNamesValidator())
+            .setDefaultValue(new ModelNode(CipherSuiteSelector.OPENSSL_DEFAULT_CIPHER_SUITE_NAMES))
+            .build();
+
     private static final String[] ALLOWED_PROTOCOLS = { "SSLv2", "SSLv3", "TLSv1", "TLSv1.1", "TLSv1.2", "TLSv1.3" };
 
     static final StringListAttributeDefinition PROTOCOLS = new StringListAttributeDefinition.Builder(ElytronDescriptionConstants.PROTOCOLS)
@@ -443,6 +451,25 @@ class SSLDefinitions {
                     } catch (PatternSyntaxException exception) {
                         throw ROOT_LOGGER.invalidHostContextMapValue(hostname);
                     }
+                }
+            }
+        }
+    }
+
+    static class CipherSuiteNamesValidator extends ModelTypeValidator {
+
+        CipherSuiteNamesValidator() {
+            super(ModelType.STRING, true, true, false);
+        }
+
+        @Override
+        public void validateParameter(String parameterName, ModelNode value) throws OperationFailedException {
+            super.validateParameter(parameterName, value);
+            if (value.isDefined()) {
+                try {
+                    CipherSuiteSelector.fromNamesString(value.asString());
+                } catch (IllegalArgumentException e) {
+                    throw ROOT_LOGGER.invalidCipherSuiteNames(e, e.getLocalizedMessage());
                 }
             }
         }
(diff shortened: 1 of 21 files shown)


jmesnil:
@fjuma I have merged #3777 (electron subsystem bump) and #3821 (elytron component upgrades), could you please remove the subsystem bump commits from this PR please?

darranl:
@jmesnil This is a GitHub bug not an issue with this PR, I will have a look at the current PRs.

elytron/src/main/java/org/wildfly/extension/elytron/SSLDefinitions.java
```diff
@@ -210,6 +210,14 @@
             .setDefaultValue(new ModelNode("DEFAULT"))
             .build();
 
+    static final SimpleAttributeDefinition CIPHER_SUITE_NAMES = new SimpleAttributeDefinitionBuilder(ElytronDescriptionConstants.CIPHER_SUITE_NAMES, ModelType.STRING, true)
+            .setAllowExpression(true)
+            .setMinSize(1)
+            .setRestartAllServices()
+            .setValidator(new CipherSuiteNamesValidator())
+            .setDefaultValue(new ModelNode("TLS_AES_256_GCM_SHA384:TLS_CHACHA20_POLY1305_SHA256:TLS_AES_128_GCM_SHA256"))
```
jmesnil: Could this default value be a constant (afaict, it's the same value than in ServerSSLContext above)?

jmesnil: is there any relation between `CIPHER_SUITE_NAMES` and `CIPHER_SUITE_FILTER`?

In their description, you wrote that:

```
elytron.server-ssl-context.cipher-suite-filter=The filter to apply to specify the enabled cipher suites for TLSv1.2 and below.
elytron.server-ssl-context.cipher-suite-names=The filter to apply to specify the enabled cipher suites for TLSv1.3.
```

I suppose that means that I can set both attributes and depending on the TLS version, one or the other would be used. Is that correct?

fjuma: > Could this default value be a constant (afaict, it's the same value than in ServerSSLContext above)?

I've updated both ServerSSLContext and SSLDefinitions to make use of a constant.

fjuma: > is there any relation between `CIPHER_SUITE_NAMES` and `CIPHER_SUITE_FILTER`?
> 
> In their description, you wrote that:
> 
> ```
> elytron.server-ssl-context.cipher-suite-filter=The filter to apply to specify the enabled cipher suites for TLSv1.2 and below.
> elytron.server-ssl-context.cipher-suite-names=The filter to apply to specify the enabled cipher suites for TLSv1.3.
> ```
> 
> I suppose that means that I can set both attributes and depending on the TLS version, one or the other would be used. Is that correct?

Yes, that's correct, both attributes can be set. If TLS 1.3 is supported, then the cipher suite will be selected using `CIPHER_SUITE_NAMES`. If TLS 1.3 is not supported, then the cipher suite will be selected using `CIPHER_SUITE_FILTER`. (There were many changes made to the way cipher suites work starting in TLS 1.3. As a result, pre-TLS 1.3 cipher suites cannot be used with TLS 1.3 and TLS 1.3 cipher suites cannot be used with TLS 1.2 or earlier.)

bstansberry:
retest this please

pom.xml
```diff
@@ -219,7 +223,7 @@
         <version.org.wildfly.openssl.wildfly-openssl-solaris-x86_64>${version.org.wildfly.openssl.natives}</version.org.wildfly.openssl.wildfly-openssl-solaris-x86_64>
         <version.org.wildfly.openssl.wildfly-openssl-windows-i386>${version.org.wildfly.openssl.natives}</version.org.wildfly.openssl.wildfly-openssl-windows-i386>
         <version.org.wildfly.openssl.wildfly-openssl-windows-x86_64>${version.org.wildfly.openssl.natives}</version.org.wildfly.openssl.wildfly-openssl-windows-x86_64>
-        <version.org.wildfly.security.elytron>1.11.0.CR1</version.org.wildfly.security.elytron>
+        <version.org.wildfly.security.elytron>1.11.0.CR2-SNAPSHOT</version.org.wildfly.security.elytron>
```
jstourac: This change is causing CI tests to fail as artifact cannot be found.

fjuma: Sorry about that, that snuck in by mistake. CI will still fail though as this PR depends on the corresponding Elytron PR.

jmesnil:
@fjuma this PR needs to be rebased on top of recent Elytron changes

fjuma:
@jmesnil Just realized that this was probably the one you were referring to on yesterday's call, was thinking of a different one :) I've rebased this but the Elytron PR that this depends on has not been merged yet so CI will still fail here. We can't merge the Elytron PR since this RFE has not been marked as pre-checked due to a Java related [performance regression](https://issues.redhat.com/browse/WFWIP-160) that occurs in Undertow with TLS 1.3 and JDK 11. To try to move this RFE forward, one of the things I'm going to look at is possibly having TLS 1.3 disabled by default and allow it to be enabled for those who want to try it out.

darranl:
Retest this please

jmesnil:
@fjuma thanks

PR #3816, opened 2019-06-12
[WFCORE-4521] PropertyFileFinder should not validate perms for non-ex…

bstansberry:
…istent files as they are ignored

https://issues.jboss.org/browse/WFCORE-4521


bstansberry:
@fjuma or @darranl please review

PR #3817, opened 2019-06-12
[WFCORE-4521] PropertyFileFinder should not validate perms for non-ex…

bstansberry:
…istent files as they are ignored

https://issues.jboss.org/browse/WFCORE-4521

Replaces #3816


darranl:
In general the code changes seem fine but the test has broken on Windows

bstansberry:
Clearing between tests the system properties the tests set to control the finder behavior seems to fix the problem. I don't know why that was only an issue on Windows (test order I suppose, although the order difference isn't obvious) but anyway the cleanup is the right thing to do regardless. It looks like the PropertyFileFinder was finding system property jboss.server.config.user.dir set to a leftover value and therefore was looking for the property file there, and therefore not finding the file the current test wanted used.

PR #3819, opened 2019-06-17
[WFCORE-4519] Fix content references management when a deployment is replaced

yersan:
Specifically, it addresses two incorrect behavior 

1. When a deployment is replaced using --force, which translates into `:full-replace-deployment` operation, we were not tracking correctly the references to the replaced content. The problem here is the content could be deleted by the cleanup task.
2. Server group replacement operation was using an incorrect path creating the content reference, instead of /server-group=server-group-name it should be /server-group=server-group-name/deployment=deployment-name. The use of that address is incorrect; if a server group deployment is later removed, then the HC content repository won't be deleted until the HC is reloaded. 
 
Additionally, added debug trace to follow up the current references of the local content repository, that could help debugging issues related. 

Jira issue: https://issues.jboss.org/browse/WFCORE-4519

### DIFF
--- a/host-controller/src/main/java/org/jboss/as/domain/controller/operations/deployment/ServerGroupDeploymentReplaceHandler.java
+++ b/host-controller/src/main/java/org/jboss/as/domain/controller/operations/deployment/ServerGroupDeploymentReplaceHandler.java
@@ -79,7 +79,7 @@ public class ServerGroupDeploymentReplaceHandler implements OperationStepHandler
 
         final PathElement deploymentPath = PathElement.pathElement(DEPLOYMENT, name);
         final PathElement replacePath = PathElement.pathElement(DEPLOYMENT, toReplace);
-        final PathAddress address = PathAddress.pathAddress(operation.get(OP_ADDR));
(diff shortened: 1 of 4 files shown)


host-controller/src/main/java/org/jboss/as/domain/controller/operations/deployment/ServerGroupDeploymentReplaceHandler.java
```diff
@@ -92,7 +92,8 @@ public void execute(OperationContext context, ModelNode operation) throws Operat
         final List<ContentReference> locallyAddedReferences = new LinkedList<ContentReference>();
         for (ModelNode content : deployment.require(CONTENT).asList()) {
             if ((content.hasDefined(HASH))) {
-                ContentReference reference = ModelContentReference.fromModelAddress(address, content.require(HASH).asBytes());
+                final PathAddress referenceAddress = address.append(deploymentPath);
```
jmesnil: the `referenceAddress` can be instantiated once outside of the `for` loop as it does not use anything from the content iteration.

yersan: Fixed, thanks @jmesnil

PR #3826, opened 2019-06-18
[WFCORE-4498] Only obtain the Provider[] once and re-use it.

darranl:
This is just a small optimisation to reuse a result we are guaranteed to want to use at least once more, if not many times more.

https://issues.jboss.org/browse/WFCORE-4498


darranl:
@fjuma Could you give this one a quick review please?

PR #3831, opened 2019-06-20
[WFCORE-4501] Provide XML schema for credential-reference

fjuma:
https://issues.jboss.org/browse/WFCORE-4501


darranl:
It may be worth adding a readme describing how we manage versioning of a shared resource like this - e.g. would adding a new version of this schema place any requirement on updating all existing subsystems?

fjuma:
Good point. Added a readme.

darranl:
@jmesnil I am happy to merge this one but did you want to check it over as you raised the original Jira issue?

PR #3835, opened 2019-06-21
WFCORE-4535,  WFCORE-4542, WFCORE-4541, WFCORE-4613 Overhaul discovery subsystem

pferraro:
https://issues.jboss.org/browse/WFCORE-4535
https://issues.jboss.org/browse/WFCORE-4541
https://issues.jboss.org/browse/WFCORE-4542
https://issues.jboss.org/browse/WFCORE-4613

In the process of investigating solutions for https://issues.jboss.org/browse/WFLY-12222 involving the discovery subsystem, I discovered that this subsystem never really worked correctly.
The most significant issues include:
* Inconsistent capability definitions (service value vs runtime API) between resource registration and those registered at runtime (WFCORE-4535)
  + Consequently, capability constraints were not correctly enforced for providers referenced by aggregate-provider
* Add operation handlers did not populate the discovery provider implementations they created
* "services" write attribute handler for "static-provider" throws StackOverflowErrors (WFCORE--4541).

I've re-written the add operation handlers and capability logic to address the issues above.
Subsystem capabilities are documented here: https://github.com/wildfly/wildfly-capabilities/pull/37


discovery/src/main/java/org/wildfly/extension/discovery/AggregateDiscoveryProviderService.java
```diff
@@ -0,0 +1,62 @@
...
+import org.jboss.msc.service.StartException;
+import org.jboss.msc.service.StopContext;
+import org.wildfly.discovery.impl.AggregateDiscoveryProvider;
+import org.wildfly.discovery.spi.DiscoveryProvider;
+
+/**
+ * Service that provides an aggregate discovery provider.
+ * @author Paul Ferraro
+ */
+public class AggregateDiscoveryProviderService implements Service {
```
bstansberry: Can this be package protected?

bstansberry:
Why the last commit? Spelling out the dependencies instead of just inheriting everything transitively is kind of nice IMHO.

This seems good; I want to look again at the main commit re the change in write-attribute behavior; now just sets reload-required.  Probably fine if the custom runtime API thing was not really usable.

pferraro:
> Why the last commit? Spelling out the dependencies instead of just inheriting everything transitively is kind of nice IMHO.
Unless this module was excluding all transitive dependencies from wildfly-controller (it only excludes transitive dependencies from org.wildfly.core), I don't see much point to spelling out the dependencies, if they are already part of the dependency hierarchy.  Also, I don't see any other subsystem modules that use wildcard exclusions, so I was just trying to be consistent.

If you prefer I leave them, I'll comply. :)

> I want to look again at the main commit re the change in write-attribute behavior; now just sets reload-required.

We can also support the allow-resource-service-restart operation header.

discovery/src/main/java/org/wildfly/extension/discovery/AggregateDiscoveryProviderService.java
pferraro: Yes.

pferraro:
@bstansberry Any other comments?

rhusar:
FWIW I have been working with this branch for clustering subsystems integration work and haven't found any major problems besides a minor tangential issue https://issues.jboss.org/browse/WFCORE-4613 about missing validation (not sure if originated in this branch).

pferraro:
@rhusar The missing validation was a pre-existing problem.  I've added a commit that addresses ￼
WFCORE-4613 to this PR.

rhusar:
retest this please

PR #3843, opened 2019-06-26
[WFCORE-4543] Don't assume it's safe to use the system log context. I…

jamezp:
…n the case of a CLI embedded server this could configure logging on the incorrect LogContext.

https://issues.jboss.org/browse/WFCORE-4543

Note I do think we need to write a test for this, but it's going to take some thought on how to handle it. It seems we're going to need to use a separate process in a manual mode test. I've filed https://issues.jboss.org/browse/WFCORE-4550 to create the test.

### DIFF
--- a/logging/src/main/java/org/jboss/as/logging/logmanager/WildFlyLogContextSelectorImpl.java
+++ b/logging/src/main/java/org/jboss/as/logging/logmanager/WildFlyLogContextSelectorImpl.java
@@ -34,6 +34,7 @@ import org.jboss.logmanager.ThreadLocalLogContextSelector;
 */
 class WildFlyLogContextSelectorImpl implements WildFlyLogContextSelector {
 
+    private final LogContextSelector defaultLogContextSelector;
     private final ClassLoaderLogContextSelector contextSelector;
 
     private final ThreadLocalLogContextSelector threadLocalContextSelector;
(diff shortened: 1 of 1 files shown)


logging/src/main/java/org/jboss/as/logging/logmanager/WildFlyLogContextSelectorImpl.java
```diff
@@ -51,10 +52,16 @@ public LogContext getLogContext() {
 
     WildFlyLogContextSelectorImpl(final LogContextSelector defaultLogContextSelector) {
         // There is not a way to reset the LogContextSelector after a reload. If the current selector is already a
-        // WildFlyLogContextSelectorImpl we should use the system default selector. This avoids possibly wrapping the
+        // WildFlyLogContextSelectorImpl we should use the previous default selector. This avoids possibly wrapping the
         // same log context several times. It should also work with the embedded CLI selector as the commands handle
         // setting and resetting the contexts.
-        final LogContextSelector dft = (defaultLogContextSelector instanceof WildFlyLogContextSelectorImpl ? LogContext.DEFAULT_LOG_CONTEXT_SELECTOR : defaultLogContextSelector);
+        final LogContextSelector dft;
+        if (defaultLogContextSelector instanceof WildFlyLogContextSelector) {
```
bstansberry: Nit: you check for instanceof the interface but then cast to the concrete impl.

jamezp: Whoops yeah. I had a getter at one pint then decided it was a bad idea and forgot to change the case. I'll fix that.

PR #3844, opened 2019-06-26
[WFCORE-4550] Create a test to validate the console data from a CLI embedded server reload

jamezp:
https://issues.jboss.org/browse/WFCORE-4550

Note this can be combined with #3843 or will need to be created as a PR and merged after.

@jfdenise I've moved and renamed the `CliEmbeddedServerTestCase` from the standalone test suite to the manualmode test suite as I don't see the reason for the server to be started for these tests as they start an embedded server. Does that seem reasonable to you?


jmesnil:
@jamezp thanks!

PR #3846, opened 2019-06-27
Minor license fixes

jamezp:
No JIRA required


jamezp:
@darranl These are commits from a downstream prod branch to sync the repositories. To be completely honest I just don't want more paperwork, but if it's required we can do it.

jmesnil:
@jamezp @darranl I'm ok with letting these changes go without a JIRA issue as they deal with cleanup of licenses (removing some, adding others, sorting them) and do no touch codebase.

PR #3847, opened 2019-06-27
Minor license fixes

jamezp:
No JIRA required. 
Upstream PR: #3846


jmesnil:
@jamezp @darranl I'm ok with letting these changes go without a JIRA issue as they deal with cleanup of licenses (removing some, adding others, sorting them) and do no touch codebase.

PR #3850, opened 2019-06-29
[WFCORE-4553] The IBM JVM requires the log manager to be set on the a…

jamezp:
…rgLine when forking surefire tests. The log manager is initialized before the surefire system properties are set.

https://issues.jboss.org/browse/WFCORE-4553
https://issues.jboss.org/browse/WFCORE-4554

The fix for WFCORE-4554 was added to this PR as WFCORE-4553 blocks these tests from being enabled.


testsuite/embedded/src/test/java/org/wildfly/core/test/embedded/host/controller/JBossLogManagerHostControllerTestCase.java
```diff
@@ -32,11 +35,16 @@
 
     @Test
     public void testLogManager() throws Exception {
-        // The IBM JDK creates a logger early on doesn't allow the JBoss Log Manager to be used. For now we will ignore
-        // the test when the IBM JDK is being used.
-        Assume.assumeFalse(isIbmJdk());
-        System.setProperty("test.log.file", "test-standalone-jbl.log");
-        System.setProperty("logging.configuration", getClass().getResource("/logging.properties").toExternalForm());
+        System.setProperty("test.log.file", "test-hc-jbl.log");
```
bstansberry: You changed this value which is probably why the test is failing. I think L49 needs to be changed to match.

jamezp: Correct. It passed for me locally because the order the tests ran. I'm going to make an additional change to delete the files as well so this won't happen again.

PR #3851, opened 2019-07-02
[WFCORE-4557] Do not acquire the write lock to read http-upgrade-enabled attribute

yersan:
The same handler is used to write and read, in case of reads, the model was being retrieved for an update, which acquired the model controller lock unnecessarily.

This patch changes the read of the attribute to use `readResource` instead of `readResourceForUpdate`

Jira issue: https://issues.jboss.org/browse/WFCORE-4557


jmesnil:
@yersan thanks!

PR #3852, opened 2019-07-03
[WFCORE-4558] Correctly compare null values against configured values…

jamezp:
… when determining if a handler should be reconfigured.

https://issues.jboss.org/browse/WFCORE-4558


jmesnil:
@jamezp thanks!

PR #3854, opened 2019-07-05
WFCORE-4560 Allow management interface to be able to use a https socket binding

jaikiran:
The commit here fixes the issue reported in https://issues.jboss.org/browse/WFCORE-4560. 
This fix however is more of a workaround to what seems like a (reproducible) issue in JBoss MSC, where the service registry returns null when it is queried by a "provided" name. This results in the `ManagementInterfaceAddStepHandler` failing in its `verify` step here https://github.com/wildfly/wildfly-core/blob/master/controller/src/main/java/org/jboss/as/controller/management/ManagementInterfaceAddStepHandler.java#L85 due to the "missing" service.

I don't have an easy way to add a test for this, but my manual testing with this change, against a patched WildFly instance, for the use case reported in that thread, worked fine.


jaikiran:
Hi @ropalka,  FYI ^

jaikiran:
retest this please

jmesnil:
@jaikiran thanks!

PR #3855, opened 2019-07-09
[WFCORE-4561] JMX audit log does not show operation parameters

iweiss:
Issue: https://issues.jboss.org/browse/WFCORE-4561

### DIFF
--- a/controller/src/main/java/org/jboss/as/controller/audit/InMemoryAuditLogHander.java
+++ b/controller/src/main/java/org/jboss/as/controller/audit/InMemoryAuditLogHander.java
@@ -154,8 +154,14 @@ public final class InMemoryAuditLogHander extends AuditLogHandler {
             }
             entry.get(METHOD_PARAMETERS);
             for (Object param : item.getMethodParams()) {
-                //TODO handle arrays better
-                entry.get(METHOD_PARAMETERS).add(param == null ? UNDEFINED : new ModelNode(param.toString()));
+                if (param != null && param.getClass().isArray()) {
+                    Object[] arrayParams = (Object[]) param;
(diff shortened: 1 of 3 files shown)


controller/src/main/java/org/jboss/as/controller/audit/InMemoryAuditLogHander.java
```diff
@@ -154,8 +154,14 @@ String formatAuditLogItem(AuditLogItem.JmxAccessAuditLogItem item) {
...
-                //TODO handle arrays better
-                entry.get(METHOD_PARAMETERS).add(param == null ? UNDEFINED : new ModelNode(param.toString()));
+                if (param != null && param.getClass().isArray()) {
+                    Object[] arrayParams = (Object[]) param;
+                    for (Object arrayParam : arrayParams) {
+                        entry.get(METHOD_PARAMETERS).add(arrayParam == null ? UNDEFINED : new ModelNode(arrayParam.toString()));
+                    }
+                } else {
+                    entry.get(METHOD_PARAMETERS).add(param == null ? UNDEFINED : new ModelNode(param.toString()));
+                }
```
yersan: I've been unable to configure JMX audit logger using an in-memory-handler, so at first glance, it looks like this TYPE_JMX management on this handler is dead-code, @iweiss do you agree?

The changes LGTM, only that small doubt

iweiss: I too was unable to do it, but since the code is there and could be used, I chose to sync up both methods just in case.

jmesnil:
@iweiss thanks!

PR #3856, opened 2019-07-09
[WFCORE-4394] Enhanced Audit Logging for Elytron - RFC Support and Configuring Reconnects

justinmcook:
https://issues.jboss.org/browse/WFCORE-4394
https://issues.jboss.org/browse/EAP7-699

Analysis Document: https://github.com/wildfly/wildfly-proposals/pull/184
WildFly Documentation PR: https://github.com/wildfly/wildfly/pull/12422
Depends on https://github.com/wildfly-security/wildfly-elytron/pull/1300

### DIFF
--- a/elytron/pom.xml
+++ b/elytron/pom.xml
@@ -35,6 +35,11 @@
             <groupId>org.wildfly.core</groupId>
             <artifactId>wildfly-controller</artifactId>
         </dependency>
+
+        <dependency>
+            <groupId>org.wildfly.core</groupId>
+            <artifactId>wildfly-controller-client</artifactId>
+        </dependency>
         <dependency>
             <groupId>org.wildfly.core</groupId>
             <artifactId>wildfly-server</artifactId>
@@ -87,6 +92,12 @@
             <artifactId>jboss-logging-processor</artifactId>
             <scope>provided</scope>
             <optional>true</optional>
+        </dependency>
+
+        <dependency>
+            <groupId>org.jboss.logmanager</groupId>
+            <artifactId>jboss-logmanager</artifactId>
+            <scope>provided</scope>
         </dependency>
 
         <dependency>
@@ -149,6 +160,12 @@
         <dependency>
             <groupId>org.jmockit</groupId>
             <artifactId>jmockit</artifactId>
+            <scope>test</scope>
+        </dependency>
+
+        <dependency>
+            <groupId>commons-lang</groupId>
+            <artifactId>commons-lang</artifactId>
             <scope>test</scope>
         </dependency>
 
--- a/elytron/src/main/java/org/wildfly/extension/elytron/ElytronSubsystemTransformers.java
+++ b/elytron/src/main/java/org/wildfly/extension/elytron/ElytronSubsystemTransformers.java
@@ -110,6 +110,13 @@ public final class ElytronSubsystemTransformers implements ExtensionTransformerR
         builder.rejectChildResource(PathElement.pathElement(ElytronDescriptionConstants.CUSTOM_EVIDENCE_DECODER));
         builder.rejectChildResource(PathElement.pathElement(ElytronDescriptionConstants.AGGREGATE_EVIDENCE_DECODER));
         builder.rejectChildResource(PathElement.pathElement(ElytronDescriptionConstants.CERTIFICATE_AUTHORITY));
+        builder.addChildResource(PathElement.pathElement(ElytronDescriptionConstants.SYSLOG_AUDIT_LOG))
+                .getAttributeBuilder()
+                .addRejectCheck(RejectAttributeChecker.DEFINED, ElytronDescriptionConstants.SYSLOG_FORMAT)
+                .setDiscard(new DiscardAttributeChecker.DiscardAttributeValueChecker(new ModelNode("RFC5424")), ElytronDescriptionConstants.SYSLOG_FORMAT)
+                .addRejectCheck(RejectAttributeChecker.DEFINED, ElytronDescriptionConstants.RECONNECT_ATTEMPTS)
+                .setDiscard(new DiscardAttributeChecker.DiscardAttributeValueChecker(AuditResourceDefinitions.RECONNECT_ATTEMPTS.getDefaultValue()), ElytronDescriptionConstants.RECONNECT_ATTEMPTS)
+                .end();
         builder.addChildResource(PathElement.pathElement(ElytronDescriptionConstants.CERTIFICATE_AUTHORITY_ACCOUNT))
                 .getAttributeBuilder()
                 .addRejectCheck(new RejectAttributeChecker.DefaultRejectAttributeChecker() {
--- a/elytron/src/main/resources/org/wildfly/extension/elytron/LocalDescriptions.properties
+++ b/elytron/src/main/resources/org/wildfly/extension/elytron/LocalDescriptions.properties
@@ -168,6 +168,8 @@ elytron.syslog-audit-log.transport=The transport to use to connect to the syslog
 elytron.syslog-audit-log.format=The format to use to record the audit event.
 elytron.syslog-audit-log.host-name=The host name to embed withing all events sent to the remote syslog server.
 elytron.syslog-audit-log.ssl-context=The SSLContext to use to connect to the syslog server when SSL_TCP transport is used.
+elytron.syslog-audit-log.syslog-format=The RFC format to be used for describing the audit event.
+elytron.syslog-audit-log.reconnect-attempts=The maximum amount of failed reconnect attempts that should be made for sending messages to a syslog server before the endpoint is closed.
 
 ################################
 # HTTP and SASL Authentication #
--- a/elytron/src/main/resources/schema/wildfly-elytron_8_0.xsd
+++ b/elytron/src/main/resources/schema/wildfly-elytron_8_0.xsd
@@ -753,6 +753,20 @@
                         </xs:documentation>
                     </xs:annotation>
                 </xs:attribute>
+                <xs:attribute name="syslog-format" type="xs:string">
+                    <xs:annotation>
+                        <xs:documentation>
+                            The RFC format to be used for formatting the log entry, default value of RFC5424.
+                        </xs:documentation>
+                    </xs:annotation>
+                </xs:attribute>
+                <xs:attribute name="reconnect-attempts" type="xs:int">
+                    <xs:annotation>
+                        <xs:documentation>
+                            The maximum amount of failed reconnect attempts that should be made for sending messages to a syslog server before the endpoint is closed, default value of 0 (no reconnect attempts).
+                        </xs:documentation>
+                    </xs:annotation>
+                </xs:attribute>
             </xs:extension>
         </xs:complexContent>
     </xs:complexType>
--- a/elytron/src/test/java/org/wildfly/extension/elytron/SubsystemTransformerTestCase.java
+++ b/elytron/src/test/java/org/wildfly/extension/elytron/SubsystemTransformerTestCase.java
@@ -121,6 +121,9 @@ public class SubsystemTransformerTestCase extends AbstractSubsystemBaseTest {
                 .addFailedAttribute(SUBSYSTEM_ADDRESS.append(PathElement.pathElement(ElytronDescriptionConstants.SIZE_ROTATING_FILE_AUDIT_LOG, "audit6")),
                         new FailedOperationTransformationConfig.NewAttributesConfig(AuditResourceDefinitions.AUTOFLUSH)
                 )
+                .addFailedAttribute(SUBSYSTEM_ADDRESS.append(PathElement.pathElement(ElytronDescriptionConstants.SYSLOG_AUDIT_LOG, "audit7")),
+                        new FailedOperationTransformationConfig.NewAttributesConfig(AuditResourceDefinitions.SYSLOG_FORMAT, AuditResourceDefinitions.RECONNECT_ATTEMPTS)
+                )
                 .addFailedAttribute(SUBSYSTEM_ADDRESS.append(PathElement.pathElement(ElytronDescriptionConstants.SERVER_SSL_SNI_CONTEXT)),
                         FailedOperationTransformationConfig.REJECTED_RESOURCE)
                 .addFailedAttribute(SUBSYSTEM_ADDRESS.append(PathElement.pathElement(JDBC_REALM, "JdbcBcryptHashHex")), REJECTED_RESOURCE)
--- a/elytron/src/test/java/org/wildfly/extension/elytron/SyslogAuditLogTestCase.java
+++ b/elytron/src/test/java/org/wildfly/extension/elytron/SyslogAuditLogTestCase.java
@@ -33,6 +33,7 @@ import org.junit.Assert;
 import org.junit.Assume;
 import org.junit.Before;
 import org.junit.BeforeClass;
+import org.junit.Ignore;
 import org.junit.Test;
 
 /**
@@ -165,6 +166,7 @@ public class SyslogAuditLogTestCase extends AbstractSubsystemTest {
      * Tests that the server does not attempt to resend the message
      */
     @Test
+    @Ignore("WFCORE-4631 - need to possibly move this or invoke a SecurityDomain authentication attempt which will invoke a log message")
     public void testZeroReconnectAttemptsBadHost() {
         // Windows Server can have an issue echoing back an error, causing the test to fail by not seeing a failure sending
         Assume.assumeFalse("Test does not run on Windows Server", SystemUtils.OS_NAME.contains("Windows Server"));
@@ -178,6 +180,7 @@ public class SyslogAuditLogTestCase extends AbstractSubsystemTest {
      * Tests that the server reattempts to send the message {@link SyslogAuditLogTestCase#RECONNECT_NUMBER} times
      */
     @Test
+    @Ignore("WFCORE-4631 - need to possibly move this or invoke a SecurityDomain authentication attempt which will invoke a log message")
     public void testNumberedReconnectAttemptsBadHost() {
         // Windows Server can have an issue echoing back an error, causing the test to fail by not seeing a failure sending
         Assume.assumeFalse("Test does not run on Windows Server", SystemUtils.OS_NAME.contains("Windows Server"));
--- a/elytron/src/test/java/org/wildfly/extension/elytron/SyslogAuditLogTestCase.java
+++ b/elytron/src/test/java/org/wildfly/extension/elytron/SyslogAuditLogTestCase.java
@@ -1,0 +1,334 @@
+/*
+Copyright 2017 Red Hat, Inc.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+  http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+ */
+package org.wildfly.extension.elytron;
+
+import static org.jboss.as.controller.descriptions.ModelDescriptionConstants.CANCELLED;
+import static org.jboss.as.controller.descriptions.ModelDescriptionConstants.FAILED;
+import static org.jboss.as.controller.descriptions.ModelDescriptionConstants.FAILURE_DESCRIPTION;
+import static org.jboss.as.controller.descriptions.ModelDescriptionConstants.OUTCOME;
+import static org.jboss.as.controller.descriptions.ModelDescriptionConstants.SUCCESS;
+import static org.junit.Assert.assertTrue;
+
+import org.apache.commons.lang3.SystemUtils;
+import org.jboss.as.controller.client.helpers.ClientConstants;
+import org.jboss.as.subsystem.test.AbstractSubsystemTest;
+import org.jboss.as.subsystem.test.KernelServices;
+import org.jboss.dmr.ModelNode;
+import org.junit.After;
+import org.junit.AfterClass;
+import org.junit.Assert;
+import org.junit.Assume;
+import org.junit.Before;
+import org.junit.BeforeClass;
+import org.junit.Test;
+
+/**
+ * Tests of syslog audit logging attributes
+ *
+ * @author <a href="mailto:jucook@redhat.com">Justin Cook</a>
+ */
+public class SyslogAuditLogTestCase extends AbstractSubsystemTest {
+
+    private static final int UDP_PORT = 10838;
+    private final int RECONNECT_NUMBER = 10;
+    private final int BAD_RECONNECT_NUMBER = -2;
+    private final int RECONNECT_TIMEOUT = 1;
+    private final String HOST_NAME = "localhost";
+    private final String SERVER_ADDRESS = "127.0.0.1";
+    private final String BAD_SERVER_ADDRESS = "0.0.0.1";
+    private final String UDP_TRANSPORT = "UDP";
+    private final String BASE_SYSLOG_MESSAGE = "Elytron audit logging enabled with RFC format: ";
+    private final String RFC3164_STRING = "RFC3164";
+    private final String RFC3164_SYSLOG_MESSAGE = BASE_SYSLOG_MESSAGE + RFC3164_STRING;
+    private final String RFC5424_STRING = "RFC5424";
+    private final String RFC5424_SYSLOG_MESSAGE = BASE_SYSLOG_MESSAGE + RFC5424_STRING;
+    private final String BAD_RFC_STRING = "RFC1";
+    private String failString = "";
+
+    private KernelServices services = null;
+    private static SimpleSyslogServer udpServer = null;
+    private ModelNode udpOperation = null;
+    private ModelNode badHostUdpOperation = null;
+
+    public SyslogAuditLogTestCase() {
+        super(ElytronExtension.SUBSYSTEM_NAME, new ElytronExtension());
+    }
+
+    /**
+     * Creates the kernel services and sets up the base operations before each test
+     */
+    @Before
+    public void init() throws Exception {
+        services = super.createKernelServicesBuilder(new TestEnvironment()).setSubsystemXmlResource("syslog-audit-logging.xml").build();
+        if (!services.isSuccessfulBoot()) {
+            Assert.fail(services.getBootError().toString());
+        }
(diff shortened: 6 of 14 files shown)


elytron/src/main/resources/org/wildfly/extension/elytron/LocalDescriptions.properties
```diff
@@ -168,6 +168,8 @@ elytron.syslog-audit-log.transport=The transport to use to connect to the syslog
 elytron.syslog-audit-log.format=The format to use to record the audit event.
 elytron.syslog-audit-log.host-name=The host name to embed withing all events sent to the remote syslog server.
 elytron.syslog-audit-log.ssl-context=The SSLContext to use to connect to the syslog server when SSL_TCP transport is used.
+elytron.syslog-audit-log.syslog-format=The RFC format to be used for describing the audit event.
```
fjuma: It might be good to indicate the default value here.

elytron/src/main/resources/org/wildfly/extension/elytron/LocalDescriptions.properties
```diff
@@ -168,6 +168,8 @@ elytron.syslog-audit-log.transport=The transport to use to connect to the syslog
...
+elytron.syslog-audit-log.reconnect-attempts=The maximum number of reconnect attempts that should be made for sending the message with UDP.
```
fjuma: Same here.

elytron/src/main/resources/schema/wildfly-elytron_7_0.xsd
```diff
@@ -748,6 +748,20 @@
                         </xs:documentation>
                     </xs:annotation>
                 </xs:attribute>
+                <xs:attribute name="syslog-format" type="xs:string">
+                    <xs:annotation>
+                        <xs:documentation>
+                            The RFC format to be used for formatting the log entry.
```
fjuma: Could indicate the default value here.

elytron/src/main/resources/schema/wildfly-elytron_7_0.xsd
```diff
@@ -748,6 +748,20 @@
...
+                        </xs:documentation>
+                    </xs:annotation>
+                </xs:attribute>
+                <xs:attribute name="reconnect-attempts" type="xs:int">
+                    <xs:annotation>
+                        <xs:documentation>
+                            The maximum number of reconnect attempts that should be made.
```
fjuma: Same here.

elytron/src/main/resources/schema/wildfly-elytron_7_0.xsd
```diff
@@ -748,6 +748,20 @@
                         </xs:documentation>
                     </xs:annotation>
                 </xs:attribute>
+                <xs:attribute name="syslog-format" type="xs:string">
```
fjuma: Since the schema has been bumped, these changes should be in ```wildfly-elytron_8_0.xsd```.

elytron/src/main/resources/org/wildfly/extension/elytron/LocalDescriptions.properties
justinmcook: Fixed

elytron/src/main/resources/org/wildfly/extension/elytron/LocalDescriptions.properties
justinmcook: Fixed

elytron/src/main/resources/schema/wildfly-elytron_7_0.xsd
justinmcook: Fixed

elytron/src/main/resources/schema/wildfly-elytron_7_0.xsd
justinmcook: Fixed

elytron/src/main/resources/schema/wildfly-elytron_7_0.xsd
justinmcook: Fixed

justinmcook:
Added the transformers in ```org.wildfly.extension.elytron.ElytronSubsystemTransformers``` and ```SubsystemTransformerTestCase```

elytron/src/main/resources/org/wildfly/extension/elytron/LocalDescriptions.properties
```diff
@@ -168,6 +168,8 @@ elytron.syslog-audit-log.transport=The transport to use to connect to the syslog
 elytron.syslog-audit-log.format=The format to use to record the audit event.
 elytron.syslog-audit-log.host-name=The host name to embed withing all events sent to the remote syslog server.
 elytron.syslog-audit-log.ssl-context=The SSLContext to use to connect to the syslog server when SSL_TCP transport is used.
+elytron.syslog-audit-log.syslog-format=The RFC format to be used for describing the audit event, default value of RFC5424.
+elytron.syslog-audit-log.reconnect-attempts=The maximum number of reconnect attempts that should be made for sending the message with UDP, default value of 0 (no reconnect attempts).
```
fjuma: May need to double check this default. The comments in the analysis document mention that the default should be -1 but the description in the analysis document currently mentions 0.

elytron/src/main/resources/schema/wildfly-elytron_8_0.xsd
```diff
@@ -753,6 +753,20 @@
...
+                    <xs:annotation>
+                        <xs:documentation>
+                            The RFC format to be used for formatting the log entry, default value of RFC5424.
+                        </xs:documentation>
+                    </xs:annotation>
+                </xs:attribute>
+                <xs:attribute name="reconnect-attempts" type="xs:int">
+                    <xs:annotation>
+                        <xs:documentation>
+                            The maximum number of reconnect attempts that should be made, default value of 0 (no reconnect attempts).
```
fjuma: Same here.

elytron/src/test/java/org/wildfly/extension/elytron/SubsystemTransformerTestCase.java
```diff
@@ -121,6 +121,12 @@ public void testRejectingTransformersEAP720() throws Exception {
                 .addFailedAttribute(SUBSYSTEM_ADDRESS.append(PathElement.pathElement(ElytronDescriptionConstants.SIZE_ROTATING_FILE_AUDIT_LOG, "audit6")),
                         new FailedOperationTransformationConfig.NewAttributesConfig(AuditResourceDefinitions.AUTOFLUSH)
                 )
+                .addFailedAttribute(SUBSYSTEM_ADDRESS.append(PathElement.pathElement(ElytronDescriptionConstants.SYSLOG_AUDIT_LOG, "audit7")),
+                        new FailedOperationTransformationConfig.NewAttributesConfig(AuditResourceDefinitions.RECONNECT_ATTEMPTS)
+                )
+                .addFailedAttribute(SUBSYSTEM_ADDRESS.append(PathElement.pathElement(ElytronDescriptionConstants.SYSLOG_AUDIT_LOG, "audit8")),
+                        new FailedOperationTransformationConfig.NewAttributesConfig(AuditResourceDefinitions.SYSLOG_FORMAT)
+                )
```
fjuma: XML elements corresponding to "audit7" and "audit8" should also be added in elytron-transformers-4.0-reject.xml

elytron/src/main/java/org/wildfly/extension/elytron/ElytronSubsystemTransformers.java
```diff
@@ -110,6 +110,8 @@ private static void from8(ChainedTransformationDescriptionBuilder chainedBuilder
         builder.rejectChildResource(PathElement.pathElement(ElytronDescriptionConstants.CUSTOM_EVIDENCE_DECODER));
         builder.rejectChildResource(PathElement.pathElement(ElytronDescriptionConstants.AGGREGATE_EVIDENCE_DECODER));
         builder.rejectChildResource(PathElement.pathElement(ElytronDescriptionConstants.CERTIFICATE_AUTHORITY));
+        builder.rejectChildResource(PathElement.pathElement(ElytronDescriptionConstants.RECONNECT_ATTEMPTS));
+        builder.rejectChildResource(PathElement.pathElement(ElytronDescriptionConstants.SYSLOG_FORMAT));
```
fjuma: Since two new attributes (as opposed to resources) were added, you'll want to use:

```builder.addChildResource(...).getAttributeBuilder().addRejectCheck(...)```

**Comment by jamezp** [comment]
This looks like it's got some compile issues.

File: elytron/pom.xml [hunk 10]
```diff
@@ -89,6 +94,12 @@
             <optional>true</optional>
         </dependency>
 
+        <dependency>
+            <groupId>org.jboss.logmanager</groupId>
+            <artifactId>jboss-logmanager</artifactId>
+            <scope>provided</scope>
+        </dependency>
```
jamezp: Is any part, exception the enum, of the log manager used here?

elytron/src/test/java/org/wildfly/extension/elytron/SyslogAuditLogTestCase.java
```diff
@@ -0,0 +1,327 @@
...
+    }
+
+    /**
+     * Tests that the syslog configuration operation fails if an invalid RFC format is used
+     */
+    @Test
+    public void testWrongSyslogFormat() {
+        udpOperation.get(ElytronDescriptionConstants.SYSLOG_FORMAT).set(BAD_RFC_STRING);
+        ModelNode response = services.executeOperation(udpOperation);
+        assertCorrectError(response, "No enum constant org.jboss.logmanager.handlers.SyslogHandler.SyslogType.RFC1");
```
jamezp: We shouldn't use the actual message here. We should use the message ID or a pattern of the message ID and the "RFC1" portion. The reason is this test could break if once the translations make it in.

elytron/src/test/java/org/wildfly/extension/elytron/SyslogAuditLogTestCase.java
```diff
@@ -0,0 +1,327 @@
...
+    }
+
+    /**
+     * Tests that the syslog configuration operation fails if an invalid icmp-error-reconnect-attempts number is used
+     */
+    @Test
+    public void testBadIntegerReconnectAttempts() {
+        udpOperation.get(ElytronDescriptionConstants.ICMP_ERROR_RECONNECT_ATTEMPTS).set(BAD_RECONNECT_NUMBER);
+        ModelNode response = services.executeOperation(udpOperation);
+        assertCorrectError(response, BAD_RECONNECT_NUMBER + " is an invalid value for parameter icmp-error-reconnect-attempts. A minimum value of -1 is required");
```
jamezp: Same as above. We should use the message ID or assume a failure is enough.

elytron/src/test/java/org/wildfly/extension/elytron/SyslogAuditLogTestCase.java
```diff
@@ -0,0 +1,327 @@
...
+    /**
+     * Tests that the server does not attempt to resend the message
+     */
+    @Test
+    public void testZeroReconnectAttemptsBadHost() {
+        // Windows Server can have an issue echoing back an error, causing the test to fail by not seeing a failure sending
+        Assume.assumeTrue("Test does not run on Windows Server", SystemUtils.OS_NAME.contains("Windows Server"));
+        badHostUdpOperation.get(ElytronDescriptionConstants.ICMP_ERROR_RECONNECT_ATTEMPTS).set(0);
+        ModelNode response = services.executeOperation(badHostUdpOperation);
+        assertCorrectError(response, "The maximum icmp-error-reconnect-attempts value of 0 was reached");
```
jamezp: Same as above.

elytron/src/test/java/org/wildfly/extension/elytron/SyslogAuditLogTestCase.java
```diff
@@ -0,0 +1,327 @@
...
+        assertFailed(response);
+    }
+
+    /**
+     * Tests that the server does not attempt to resend the message
+     */
+    @Test
+    public void testZeroReconnectAttemptsBadHost() {
+        // Windows Server can have an issue echoing back an error, causing the test to fail by not seeing a failure sending
+        Assume.assumeTrue("Test does not run on Windows Server", SystemUtils.OS_NAME.contains("Windows Server"));
```
jamezp: Does it seem to work on Windows 10?

elytron/src/test/java/org/wildfly/extension/elytron/SyslogAuditLogTestCase.java
```diff
@@ -0,0 +1,327 @@
...
+    /**
+     * Tests that the server reattempts to send the message {@link SyslogAuditLogTestCase#RECONNECT_NUMBER} times
+     */
+    @Test
+    public void testNumberedReconnectAttemptsBadHost() {
+        // Windows Server can have an issue echoing back an error, causing the test to fail by not seeing a failure sending
+        Assume.assumeTrue("Test does not run on Windows Server", !SystemUtils.OS_NAME.contains("Windows Server"));
+        badHostUdpOperation.get(ElytronDescriptionConstants.ICMP_ERROR_RECONNECT_ATTEMPTS).set(RECONNECT_NUMBER);
+        ModelNode response = services.executeOperation(badHostUdpOperation);
+        assertCorrectError(response, "The maximum icmp-error-reconnect-attempts value of 10 was reached");
```
jamezp: Same as the others.

elytron/src/test/java/org/wildfly/extension/elytron/SyslogAuditLogTestCase.java
```diff
@@ -0,0 +1,327 @@
...
+        assertFailed(response);
+    }
+
+    /**
+     * Tests that the server reattempts to send the message {@link SyslogAuditLogTestCase#RECONNECT_NUMBER} times
+     */
+    @Test
+    public void testNumberedReconnectAttemptsBadHost() {
+        // Windows Server can have an issue echoing back an error, causing the test to fail by not seeing a failure sending
+        Assume.assumeTrue("Test does not run on Windows Server", !SystemUtils.OS_NAME.contains("Windows Server"));
```
jamezp: This one seems to be the opposite of above, but the same comments. I'm guessing this should be:
```
Assume.assumeFalse("Test does not run on Windows Server", SystemUtils.OS_NAME.contains("Windows Server"));
```

elytron/src/test/java/org/wildfly/extension/elytron/SyslogAuditLogTestCase.java
```diff
@@ -0,0 +1,327 @@
...
+        assertFailed(response);
+    }
+
+    /**
+     * Tests that the server continuously attempts to resend the message for {@link SyslogAuditLogTestCase#RECONNECT_TIMEOUT} milliseconds
+     */
+    @Test
+    public void testInfiniteReconnectAttemptsBadHost() throws Exception {
+        // Windows Server can have an issue echoing back an error, causing the test to fail by not seeing a failure sending
+        Assume.assumeTrue("Test does not run on Windows Server", !SystemUtils.OS_NAME.contains("Windows Server"));
```
jamezp: Same as above assume.

elytron/src/test/java/org/wildfly/extension/elytron/SyslogAuditLogTestCase.java
```diff
@@ -0,0 +1,327 @@
...
+
+    /**
+     * Calls the syslog operation creator with a given address using UDP
+     */
+    private ModelNode createUdpSyslogOperation(String serverAddress) {
+        return createSyslogOperation(UDP_PORT, UDP_TRANSPORT, serverAddress);
+    }
+
+    private void assertCorrectError(ModelNode response, String msg) {
+        System.out.println(response.get(FAILURE_DESCRIPTION).asString());
```
jamezp: This should probably be removed.

elytron/pom.xml
justinmcook: Only the SyslogType enum is used, should I just copy the enum over and remove the dependency?

elytron/src/test/java/org/wildfly/extension/elytron/SyslogAuditLogTestCase.java
justinmcook: Updated

elytron/src/test/java/org/wildfly/extension/elytron/SyslogAuditLogTestCase.java
justinmcook: Updated

elytron/src/test/java/org/wildfly/extension/elytron/SyslogAuditLogTestCase.java
justinmcook: Updated

elytron/src/test/java/org/wildfly/extension/elytron/SyslogAuditLogTestCase.java
justinmcook: It ran successfully on my Windows 10 VM, it seems to only be Windows Server that has the issue but I was unable to figure out what was causing it

elytron/src/test/java/org/wildfly/extension/elytron/SyslogAuditLogTestCase.java
justinmcook: Updated

elytron/src/test/java/org/wildfly/extension/elytron/SyslogAuditLogTestCase.java
justinmcook: Updated

elytron/src/test/java/org/wildfly/extension/elytron/SyslogAuditLogTestCase.java
justinmcook: Updated

elytron/src/test/java/org/wildfly/extension/elytron/SyslogAuditLogTestCase.java
justinmcook: Left it in by accident, removed it now

justinmcook:
> This looks like it's got some compile issues.
@jamezp The compile issue is due to requiring an Elytron upgrade

elytron/pom.xml
jamezp: I'll say we should consider it. However since this doesn't rely on the logmanager being the actual log manager we're probably okay.

jamezp:
I placed a hold on this as it requires https://github.com/wildfly-security/wildfly-elytron/pull/1300 and a Elytron upgrade.

darranl:
@jamezp I don't think we really need a hold label on this one, I think the missing reqs label is sufficient.

Also thanks for the detailed review but in general once a hold label ends up on a PR that really does end reviews happening on that PR, these PRs are raised for code review to be performed.

jamezp:
That makes sense @darranl. I should have looked at the reqs I just thought hold because we're waiting for a component upgrade is all.

elytron/src/test/java/org/wildfly/extension/elytron/SyslogAuditLogTestCase.java
```diff
@@ -0,0 +1,334 @@
...
+        if (! response.get(OUTCOME).asString().equals(FAILED)) {
+            Assert.fail(response.toJSONString(false));
+        }
+        return response;
+    }
+
+    /**
+     * Verifies the operation cancelled before completion
+     */
+    private ModelNode assertCancelled(ModelNode response) {
```
honza-kasik: Isn't the method name misleading? I would assume that it throws some kind of exception based on its name.  Wouldn't also this approach mute any previous responses which are not cancelled and leave only the last one which would then be in `failString`? Does it make sense to continue in test if any of requests hasn't been cancelled? (I'm referencing https://github.com/wildfly/wildfly-core/pull/3856/files#diff-db84c0d398e11d3e59f5b01d0367d032R202)

justinmcook: The method name is based on what the ModelNode response string contains ("cancelled"), with the naming being to keep it in line with the other two response checks (assertFailed and assertSuccess). Only one response would be created, since it is in the executeOperation that the child thread waits (while it attempts to send the initial connection message). If the response seen in assertCancelled is not "cancelled" (meaning that there was not infinite reconnect attempts since otherwise an exception would be thrown) then failString will be set, causing the test to fail due to failString being non empty after the threads are joined. Otherwise, failString will get set due to an exception being thrown during the threads execution which will also cause the test to fail after the threads are joined.

honza-kasik: Thank you. When I was talking about misleading name I had the "assert" part in mind.

Yes, `failString` is set, but if there was more than one error (result with status set to anything else than "cancelled"), we can see only last error.

justinmcook: I'm not sure that I see the case where it gets set, due to the only cases that it gets set being (with only one of the following scenarios being possible in any given test run):
1. Once if there is an unexpected exception that gets thrown during the infinite attempts while executeOperation is running (where the infinite attempts happens inside of)
2. After the response gets set, which will occur only one time after the parent thread interrupts the child thread (the expected outcome of failString being kept empty due to assertCancelled seeing the status as cancelled and so skipping over setting the value) or if the message gets successfully sent (which should never happen regardless due to the invalid IP)/if the operation fails due to some other issue, in either case assertCancelled will be called once and will set the failString value with the bad outcome and then the child thread will end before the interruption
In either of the above cases, the child thread will no longer be executing once failString gets set. What's the situation you're thinking of where failString will get set multiple times?

honza-kasik: Ah, my bad. I was reading the code wrong. Thank you.

elytron/src/main/java/org/wildfly/extension/elytron/ElytronSubsystemTransformers.java
```diff
@@ -110,6 +110,13 @@ private static void from8(ChainedTransformationDescriptionBuilder chainedBuilder
         builder.rejectChildResource(PathElement.pathElement(ElytronDescriptionConstants.CUSTOM_EVIDENCE_DECODER));
         builder.rejectChildResource(PathElement.pathElement(ElytronDescriptionConstants.AGGREGATE_EVIDENCE_DECODER));
         builder.rejectChildResource(PathElement.pathElement(ElytronDescriptionConstants.CERTIFICATE_AUTHORITY));
+        builder.addChildResource(PathElement.pathElement(ElytronDescriptionConstants.SYSLOG_AUDIT_LOG))
+                .getAttributeBuilder()
+                .addRejectCheck(RejectAttributeChecker.DEFINED, ElytronDescriptionConstants.SYSLOG_FORMAT)
+                .setDiscard(DiscardAttributeChecker.UNDEFINED, ElytronDescriptionConstants.SYSLOG_FORMAT)
```
kabir: Maybe `DiscardAttributeValueChecker(new ModelNode("RFC5424"))`?

elytron/src/main/java/org/wildfly/extension/elytron/ElytronSubsystemTransformers.java
```diff
@@ -110,6 +110,13 @@ private static void from8(ChainedTransformationDescriptionBuilder chainedBuilder
...
+                .addRejectCheck(RejectAttributeChecker.DEFINED, ElytronDescriptionConstants.RECONNECT_ATTEMPTS)
+                .setDiscard(DiscardAttributeChecker.UNDEFINED, ElytronDescriptionConstants.RECONNECT_ATTEMPTS)
```
kabir: Might need to take the implied value in the old server into account

justinmcook:
@kabir Updated it to use the ```DiscardAttributeValueChecker``` for the syslog-format, left the reconnect-attempts as is because in the past it functioned as just not checking if there was an error. Removed the default values from LocalDescription.properties

elytron/src/test/java/org/wildfly/extension/elytron/SyslogAuditLogTestCase.java
```diff
@@ -0,0 +1,334 @@
...
+        ModelNode response = services.executeOperation(badHostUdpOperation);
+        assertCorrectError(response, new String[] {"ELY12001", Integer.toString(RECONNECT_NUMBER)});
+        assertFailed(response);
+    }
+
+    /**
+     * Tests that the server continuously attempts to resend the message for {@link SyslogAuditLogTestCase#RECONNECT_TIMEOUT} milliseconds
+     */
+    @Test
+    public void testInfiniteReconnectAttemptsBadHost() throws Exception {
```
jamezp: I'm not sure I understand this test. It looks like it only tests that the operation fails.

justinmcook: It's to test that it doesn't stop resending the message before the interrupt from the parent thread happens, since if the it either thinks it successfully sent the message or stops sending it without the parent thread interrupting then the ModelNode response will not be cancelled and the test will fail.

justinmcook:
@kabir @jamezp I've updated the PR with the requested changes, I'll be away on Monday and Tuesday so won't be able to respond to any additional feedback after today until Wednesday.

elytron/src/main/java/org/wildfly/extension/elytron/ElytronSubsystemTransformers.java
jamezp: @justinmcook I think what Kabir means here is that if the old behavior was similar the same as {{0}} or {{-1}} we should transform the value to that value if undefined.

jamezp:
I placed a hold on this because there seems to be issues with the https://github.com/wildfly/wildfly-core/pull/3856/files#diff-db84c0d398e11d3e59f5b01d0367d032R194 test. The `ValueSupplier.dispose()` doesn't seem to be getting invoked.

elytron/src/main/resources/org/wildfly/extension/elytron/LocalDescriptions.properties
justinmcook: I originally updated it to reference RFC5424 but removed it based on https://github.com/wildfly/wildfly-core/pull/3856#pullrequestreview-276419748 . I could add the section detailing the possible values?

elytron/src/main/java/org/wildfly/extension/elytron/ElytronSubsystemTransformers.java
justinmcook: Updated it to change the undefined value to 0

bstansberry:
retest this please

PR #3857, opened 2019-07-10
[WFCORE-4564] ${maven.repository.url} in repository url is not evaluated when org.jboss:jboss-parent is missing

gaol:
Issue: https://issues.jboss.org/browse/WFCORE-4564


jamezp:
I'm closing this as we do want these properties for proxies. @gaol I'm putting an email together now explaining what we want to do here.

PR #3858, opened 2019-07-13
[WFCORE-4565] Use ModelNode.TRUE, ModelNode.FALSE, ModelNode.ZERO and ModelNode.ZERO_LONG

bstansberry:
https://issues.jboss.org/browse/WFCORE-4565

Each commit is the result of a separate find and replace across the code base. Perhaps best to review commit by commit as any anomaly would stand out more if the whole change is following the same pattern.


jmesnil:
@bstansberry thanks!

PR #3860, opened 2019-07-14
[WFCORE-3255] Object...AttributeDefinition should be sure to call nes…

bstansberry:
…ted AD correctors

https://issues.jboss.org/browse/WFCORE-3255


jmesnil:
@bstansberry thanks!

PR #3863, opened 2019-07-16
WFCORE-4287: do not display default value if alternative is set

michpetrov:
Issue: [WFCORE-4287](https://issues.jboss.org/browse/WFCORE-4287)

preliminary PR, I want to see of any more tests fail

Btw how is it going to be resolved when this PR fails because of WildFly tests, which cannot be resolved until Core upgrade?

### DIFF
--- a/controller/src/main/java/org/jboss/as/controller/operations/global/ReadAttributeHandler.java
+++ b/controller/src/main/java/org/jboss/as/controller/operations/global/ReadAttributeHandler.java
@@ -284,7 +284,7 @@ public class ReadAttributeHandler extends GlobalOperationHandlers.AbstractMultiT
         } else if (subModel.hasDefined(attribute.getName())) {
             final ModelNode result = subModel.get(attribute.getName());
             context.getResult().set(result);
-        } else if (defaults && attribute.getDefaultValue() != null) {
+        } else if (defaults && attribute.getDefaultValue() != null && !attribute.hasAlternative(subModel)) {
             // No defined value in the model. See if we should reply with a default from the metadata,
             // reply with undefined, or fail because it's a non-existent attribute name
             context.getResult().set(attribute.getDefaultValue());
--- a/subsystem-test/framework/src/main/java/org/jboss/as/subsystem/test/SubsystemTestDelegate.java
+++ b/subsystem-test/framework/src/main/java/org/jboss/as/subsystem/test/SubsystemTestDelegate.java
@@ -21,9 +21,16 @@
 */
 package org.jboss.as.subsystem.test;
 
+import static org.jboss.as.controller.descriptions.ModelDescriptionConstants.ALTERNATIVES;
+import static org.jboss.as.controller.descriptions.ModelDescriptionConstants.ATTRIBUTES;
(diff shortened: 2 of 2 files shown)


jamezp:
Those full failures are what I expected. We may want to have a look at the model comparison and see if we can fix these issues here without a `ModelFixer`. I'll have a look at that today with this PR and see if that works.

michpetrov:
The model comparison can be set to ignore defaults and that fixes things. Otherwise I've solved it with the ModelFixers - https://github.com/michpetrov/wildfly/commit/a9ee2f0683a6dc97a7a5566949ecf3e9ae9d2ff6

jamezp:
Something like this gets us close I think https://github.com/wildfly/wildfly-core/commit/4b3f24f3a05d37bad158fbf8720aa089c517fb42. It worked for logging and messaging. Transactions didn't work, but that can be looked at if we want to use a global approach like this.

michpetrov:
@jamezp there are several reasons for the transaction tests failing, fortunately the rest of the build is ok

* your code wasn't correctly processing root-level attributes, I think I've fixed it - https://github.com/michpetrov/wildfly-core/commit/db8bfdb08820d7dafd3be17ae5ecf2685184f387

* there is another comparison between the models being made in `SubsystemTestDelegate.LegacyKernelServiceInitializerImpl#bootCurrentVersionWithLegacyBootOperations` which is where the tests are failing, also fixed above; not sure why the models are being tested twice like that

* the attributes all have default values so `legacyModel.hasDefined(alternateAttributeName)` may first encounter the default and delete the actual set value, I think this one can be fixed

* lastly `use-journal-store` and `use-hornetq-store` are alternatives but one is also an alias for the other so they have to have the same value; I don't know if this violates any restrictions placed by the alternatives but if ti doesn't I don't think we can get this information out of the model description

jamezp:
So I'm good with either approach TBH. If just fixing the failing tests themselves is enough I'm okay with that. I'm guessing the upstream PR can be filed as well without requiring a core upgrade. Which ever approach you take is okay.

The main benefit I see to your first approach, just fixing the failing tests, is it only touches the failing tests and not the newer tests.

michpetrov:
@jamezp sorry this took a while, since I'm using your commit do you want to change the commit message?

jamezp:
No problem @michpetrov. Feel free to squash it if you want. We also might want to validate https://github.com/wildfly/wildfly-core/pull/3863/files#diff-818a73e27c9ada5c4f27af373f62de8cR1181. I never really looked much at it as this was just a quick and dirty test commit.

darranl:
@michpetrov what is the status of this one to get it passing?

michpetrov:
@darranl Sorry for the delay.

There's kind of a deadlock between Core and WildFly, currently this PR fails because the changes introduced here make WildFly tests fail
 * wildfly/wildfly#12515 and wildfly/wildfly#12496 need to be merged to fix the WildFly tests however merging them is going to break the tests in other ways until this PR is merged;

I don't know what's the procedure here

darranl:
@jmesnil do you have any thoughts on this one?  The changes probably need a bit more review but maybe just before the next Core tag we will need a custom CI run and coordinate both PRs being merged at the same time?

PR #3865, opened 2019-07-18
[WFCORE-4283] Move the starting of HC servers after cleaning the HC booting flag

yersan:
This patch reduces the time where the HC boot flag is set, so read operations on the HC can success. That helps HAL to be able to display the HC slave nodes and its servers when the servers are still starting after botting up the HC.

There are additional tweaks to do in HAL to allow the console work with domains with non-patched HC servers. @hpehl Maybe you could link this PR with the one required for HAL.

Jira issue: https://issues.jboss.org/browse/WFCORE-4283


jmesnil:
@yersan thanks!

hpehl:
Related HAL PR: https://github.com/hal/console/pull/331

PR #3867, opened 2019-07-18
[WFCORE-4570] quote path element if it contains = or /

kabir:
https://issues.jboss.org/browse/WFCORE-4570


jmesnil:
@kabir thanks!

PR #3870, opened 2019-07-19
[WFCORE-4569] Ensure the AuthenticationContext is also registered as the JVM wide default.

darranl:
We currently automatically associate this with the class loader of every deployment but we don't register it globally meaning any internal services that also need the default such as XA recovery don't have access to it.

https://issues.jboss.org/browse/WFCORE-4569


darranl:
This fix relates to a long standing issue that has been investigated across different teams, @fl4via is also going to take a look at verifying if this does address the larger issue.

darranl:
Thank you very much @fl4via this is good news.

jmesnil:
@darranl thanks!

PR #3871, opened 2019-07-19
[WFCORE-4574] Adding code to shut down ScheduledThreadPoolExecutor cr…

ashley-abdelsayed98:
…eated in HttpAuthenticationFactory

https://issues.jboss.org/browse/WFCORE-4574

This depends on the changes in wildfly-security/wildfly-elytron#1304

To pass CI, this will require an Elytron upgrade but this is ready for the subsystem changes to be reviewed.

### DIFF
--- a/elytron/src/main/java/org/wildfly/extension/elytron/HttpAuthenticationFactoryService.java
+++ b/elytron/src/main/java/org/wildfly/extension/elytron/HttpAuthenticationFactoryService.java
@@ -1,0 +1,63 @@
+/*
+ * JBoss, Home of Professional Open Source.
+ * Copyright 2019 Red Hat, Inc., and individual contributors
+ * as indicated by the @author tags.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.wildfly.extension.elytron;
+
+import static org.wildfly.common.Assert.checkNotNullParam;
+
+import java.util.function.Supplier;
+
+import org.jboss.msc.service.Service;
+import org.jboss.msc.service.StartContext;
+import org.jboss.msc.service.StartException;
+import org.jboss.msc.service.StopContext;
+import org.wildfly.security.auth.server.HttpAuthenticationFactory;
+
+/**
+ * A very trivial {@link Service} implementation where creation of the value type can easily be wrapped using a {@link Supplier}
+ * and the scheduled executor in HttpAuthenticationFactory is shut down.
+ *
+ * @author <a href="mailto:aabdelsa@gmail.com">Ashley Abdel-Sayed</a>
+ */
+
+class HttpAuthenticationFactoryService implements Service<HttpAuthenticationFactory> {
+
+    final Supplier<HttpAuthenticationFactory> httpAuthenticationFactorySupplier;
+    private volatile HttpAuthenticationFactory httpAuthenticationFactory;
+
+    HttpAuthenticationFactoryService(Supplier<HttpAuthenticationFactory> httpAuthenticationFactorySupplier) {
+        this.httpAuthenticationFactorySupplier = checkNotNullParam("httpAuthenticationFactorySupplier", httpAuthenticationFactorySupplier);
+    }
+
+    @Override
+    public void start(StartContext context) throws StartException {
+        httpAuthenticationFactory = httpAuthenticationFactorySupplier.get();
+    }
+
+    @Override
+    public void stop(StopContext context) {
+        httpAuthenticationFactory.shutdownAuthenticationMechanismFactory();
+        httpAuthenticationFactory = null;
+    }
+
+    @Override
+    public HttpAuthenticationFactory getValue() throws IllegalStateException, IllegalArgumentException {
+        return httpAuthenticationFactory;
+    }
+
+}
--- a/domain-management/src/main/java/org/jboss/as/domain/management/security/SecurityRealmService.java
+++ b/domain-management/src/main/java/org/jboss/as/domain/management/security/SecurityRealmService.java
@@ -379,6 +379,7 @@ public class SecurityRealmService implements Service<SecurityRealm>, SecurityRea
         }
         registeredServices.clear();
         saslAuthenticationFactory = null;
+        httpAuthenticationFactory.shutdownAuthenticationMechanismFactory();
         httpAuthenticationFactory = null;
     }
 
--- a/elytron/src/main/java/org/wildfly/extension/elytron/AuthenticationFactoryDefinitions.java
+++ b/elytron/src/main/java/org/wildfly/extension/elytron/AuthenticationFactoryDefinitions.java
@@ -29,6 +29,7 @@ import static org.wildfly.extension.elytron.Capabilities.SASL_AUTHENTICATION_FAC
 import static org.wildfly.extension.elytron.Capabilities.SASL_SERVER_FACTORY_CAPABILITY;
 import static org.wildfly.extension.elytron.Capabilities.SECURITY_DOMAIN_CAPABILITY;
 import static org.wildfly.extension.elytron.Capabilities.SECURITY_FACTORY_CREDENTIAL_CAPABILITY;
+import static org.wildfly.extension.elytron.ElytronDefinition.commonDependencies;
 import static org.wildfly.extension.elytron.ElytronExtension.getRequiredService;
 
 import org.wildfly.security.http.HttpConstants;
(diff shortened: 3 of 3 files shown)


domain-http/interface/src/main/java/org/jboss/as/domain/http/server/ManagementHttpServer.java
```diff
@@ -180,6 +180,7 @@ public void start() {
     public void stop() {
         IoUtils.safeClose(normalServer);
         IoUtils.safeClose(secureServer);
+        httpAuthenticationFactory.shutdownAuthenticationMechanismFactoryExecutory();
```
bstansberry: This method name has a typo (final 'y').

ashley-abdelsayed98:
retest this please

elytron/src/main/java/org/wildfly/extension/elytron/HttpAuthenticationFactoryAddHandler.java
```diff
@@ -0,0 +1,97 @@
...
+ * A {@link OperationStepHandler} for adding a {@link HttpAuthenticationFactoryService} for an HttpAuthenticationFactory.
+ *
+ * @author <a href="mailto:aabdelsa@redhat.com">Ashley Abdel-Sayed</a>
+ */
+
+abstract class HttpAuthenticationFactoryAddHandler<T> extends BaseAddHandler {
+
+    private final RuntimeCapability<?> runtimeCapability;
+    private final ServiceController.Mode initialMode;
+    private final ServiceController.Mode embeddedInitialMode;
```
bstansberry: You don't need these 3 fields. They only ever have a single value, so whatever uses them can just use that value.

elytron/src/main/java/org/wildfly/extension/elytron/HttpAuthenticationFactoryAddHandler.java
```diff
@@ -0,0 +1,97 @@
...
+        return true;
+    }
+
+    protected boolean dependOnProviderRegistration() {
+        return true;
+    }
+
+    protected void installedForResource(ServiceController<T> serviceController, Resource resource) {}
+
+    protected abstract Callable<HttpAuthenticationFactory> getHttpAuthenticationFactorySupplier(ServiceBuilder<T> serviceBuilder, OperationContext context, ModelNode model) throws OperationFailedException;
```
bstansberry: This method (and the class) should not be abstract and the implementation logic starting at https://github.com/wildfly/wildfly-core/pull/3871/files#diff-f20215bf36b61ecbb06008151d2f257cR379 should be moved into this class.

elytron/src/main/java/org/wildfly/extension/elytron/HttpAuthenticationFactoryAddHandler.java
```diff
@@ -0,0 +1,97 @@
...
+        httpAuthenticationFactoryService.setHttpAuthenticationFactorySupplier(getHttpAuthenticationFactorySupplier(serviceBuilder, context, resource.getModel()));
+
+        installedForResource(commonDependencies(serviceBuilder, dependOnProperties(), dependOnProviderRegistration())
+                .setInitialMode(context.getProcessType() == ProcessType.EMBEDDED_SERVER && context.getRunningMode() == RunningMode.ADMIN_ONLY ? embeddedInitialMode : initialMode)
+                .install(), resource);
+    }
+
+    protected boolean dependOnProperties() {
+        return true;
+    }
```
bstansberry: This method should be removed and the call to it replaced with a static 'true'.

elytron/src/main/java/org/wildfly/extension/elytron/HttpAuthenticationFactoryAddHandler.java
```diff
@@ -0,0 +1,97 @@
...
+
+    protected boolean dependOnProviderRegistration() {
+        return true;
+    }
```
bstansberry: This method should be removed and the call to it replaced with a static 'true'.

elytron/src/main/java/org/wildfly/extension/elytron/HttpAuthenticationFactoryAddHandler.java
```diff
@@ -0,0 +1,97 @@
...
+
+    protected void installedForResource(ServiceController<T> serviceController, Resource resource) {}
```
bstansberry: This method should be removed and the call to it removed, since it is a no-op method and there won't be any subclasses of this overriding it to do something.

elytron/src/main/java/org/wildfly/extension/elytron/HttpAuthenticationFactoryService.java
```diff
@@ -0,0 +1,75 @@
...
+/**
+ * A very trivial {@link Service} implementation where creation of the value type can easily be wrapped using a {@link Supplier}
+ * and the scheduled executor in HttpAuthenticationFactory is shut down.
+ *
+ * @author <a href="mailto:aabdelsa@gmail.com">Ashley Abdel-Sayed</a>
+ */
+
+class HttpAuthenticationFactoryService implements Service {
+
+    private Callable<HttpAuthenticationFactory> httpAuthenticationFactorySupplier;
```
bstansberry: This should be 'final' with the supplier passed into the constructor and the setter method removed. The no-arg constructor should be removed as well.

elytron/src/main/java/org/wildfly/extension/elytron/HttpAuthenticationFactoryService.java
```diff
@@ -0,0 +1,75 @@
...
+    }
+
+    void setHttpAuthenticationFactorySupplier(Callable<HttpAuthenticationFactory> httpAuthenticationFactorySupplier) {
+        this.httpAuthenticationFactorySupplier = checkNotNullParam("httpAuthenticationFactorySupplier", httpAuthenticationFactorySupplier);
+    }
+
+    @Override
+    public void start(StartContext context) throws StartException {
+        try {
+            httpAuthenticationFactory = checkNotNullParam("httpAuthenticationFactorySupplier", httpAuthenticationFactorySupplier).call();
```
bstansberry: Once 'httpAuthenticationFactorySupplier' is final, you know you've validated it in the constructor and you don't have to do it here. And that means the try/catch can be removed.

elytron/src/main/java/org/wildfly/extension/elytron/HttpAuthenticationFactoryAddHandler.java
```diff
@@ -0,0 +1,97 @@
...
+
+abstract class HttpAuthenticationFactoryAddHandler<T> extends BaseAddHandler {
+
+    private final RuntimeCapability<?> runtimeCapability;
+    private final ServiceController.Mode initialMode;
+    private final ServiceController.Mode embeddedInitialMode;
+
+    HttpAuthenticationFactoryAddHandler(Class<T> serviceType, AttributeDefinition[] attributes, RuntimeCapability<?> runtimeCapability) {
+        this(serviceType, ServiceController.Mode.ACTIVE, attributes, runtimeCapability);
+    }
```
bstansberry: The only constructor needed is whatever is invoked by AuthenticationFactoryDefinitions.

elytron/src/main/java/org/wildfly/extension/elytron/HttpAuthenticationFactoryAddHandler.java
```diff
@@ -0,0 +1,97 @@
...
+
+    HttpAuthenticationFactoryAddHandler(Class<T> serviceType, ServiceController.Mode initialMode, AttributeDefinition[] attributes, RuntimeCapability<?> runtimeCapability) {
+        this(serviceType, initialMode, initialMode, attributes, runtimeCapability);
+    }
```
bstansberry: The only constructor needed is whatever is invoked by AuthenticationFactoryDefinitions.

elytron/src/main/java/org/wildfly/extension/elytron/HttpAuthenticationFactoryAddHandler.java
bstansberry: An alternative to moving all that logic is to get rid of this class, and change

https://github.com/wildfly/wildfly-core/pull/3871/files#diff-f20215bf36b61ecbb06008151d2f257cR375

to

```
AbstractAddStepHandler add = new BaseAddHandler(HTTP_AUTHENTICATION_FACTORY_RUNTIME_CAPABILITY, attributes) {

            @Override
    protected final void performRuntime(OperationContext context, ModelNode operation, Resource resource)
            throws OperationFailedException {
                   // the performRuntime logic you have here in HttpAuthenticationFactoryAddHandler
     }

     private Callable<HttpAuthenticationFactory> getHttpAuthenticationFactorySupplier(
```

That might be cleanest as it leaves all that code that's in AuthenticationFactoryDefinitions where it is.

ashley-abdelsayed98:
@bstansberry thanks for the feedback! I've updated the PR and removed the HttpAuthenticationFactoryAddHandler class

elytron/src/main/java/org/wildfly/extension/elytron/HttpAuthenticationFactoryService.java
ashley-abdelsayed98: I've made httpAuthenticationFactorySupplier but the try/catch is needed to catch errors thrown by the call()

ashley-abdelsayed98:
retest this please

ashley-abdelsayed98:
@bstansberry Thanks Brian! I've updated the PR based on your comments.

PR #3872, opened 2019-07-19
[WFCORE-4575] remove non existent managed dependency wildfly-self-con…

emmartins:
…tained
Issue: https://issues.jboss.org/browse/WFCORE-4575


jmesnil:
@emmartins thanks!

PR #3873, opened 2019-07-19
[WFCORE-4576] move wildfly-core-vault-test-galleon-pack dep managemen…

emmartins:
…t to testsuite
Issue: https://issues.jboss.org/browse/WFCORE-4576


jmesnil:
@emmartins thanks!

PR #3875, opened 2019-07-26
Add underscore as allowed character

theit:
This small fix for wildfly-core adds support for underscores in usernames.


darranl:
For a change such as this verification would need to be performed to confirm that usernames with an underscore work correctly with both the DIGEST_MD5 SASL mechanism and the HTTP DIGEST mechanism.

The add-user utility is primarily for use with the default configuration with makes use of the Digest mechanisms.

darranl:
I am going to close this one for now as there is no activity and the legacy components are planned to be removed in future releases.  If you would like us to reconsider please resubmit and with the required WFCORE Jira issues referenced in the commit messages.  Also this still needs verification that it does not break authentication.

PR #3879, opened 2019-07-30
WFCORE-4586 Don't fail on wrong copied hashes if override options pro…

soul2zimate:
…vided.
issue: https://issues.jboss.org/browse/WFCORE-4586


jmesnil:
@soul2zimate thanks!

PR #3881, opened 2019-07-31
[WFCORE-4587] Clarify description of 'maximum-cert-path'

jstourac:
Trivial extension of 'maximum-cert-path' attribute of trust-manager to give
message when this value actually takes any effect - only if OCSP or CRL are
configured.


jmesnil:
@jstourac thanks!

PR #3882, opened 2019-08-01
[WFCORE-4525] Disable CLI embedded server tests that do not work with…

jamezp:
… the IBM J9 JVM.

This does not fix the issue, but does disable the tests if an IBM JVM is used.

https://issues.jboss.org/browse/WFCORE-4525


jmesnil:
@jamezp Should I leave WFCORE-4525 open until the tests are fixed? Or should I rename the issue "disable ...." and resolve it?

jamezp:
@jmesnil My thought was to leave it open for now. It seems like it should be solvable, but maybe it's not really worth the effort. I tried a few different things that didn't work and I guess it is just a test environment issue. If @ropalka is okay with it being resolved I'm okay with just ignoring the tests.

ropalka:
If tests will be disabled I am OK with it @jamezp and @jmesnil . The issue can be renamed, resolved and closed. What I care about is to have 100% success on IBM JDK and I don't care if tests are disabled. I agree with @jamezp it is just test environment issue, see my JIRA comment.

PR #3883, opened 2019-08-01
[WFCORE-4551] Configure JDBC Realm services to LAZY initial mode when…

ashley-abdelsayed98:
… in admin only mode

https://issues.jboss.org/browse/WFCORE-4551

### DIFF
--- a/elytron/src/main/java/org/wildfly/extension/elytron/JdbcRealmDefinition.java
+++ b/elytron/src/main/java/org/wildfly/extension/elytron/JdbcRealmDefinition.java
@@ -48,6 +48,7 @@ import org.jboss.as.controller.OperationFailedException;
 import org.jboss.as.controller.OperationStepHandler;
 import org.jboss.as.controller.PathElement;
 import org.jboss.as.controller.ResourceDefinition;
+import org.jboss.as.controller.RunningMode;
 import org.jboss.as.controller.SimpleAttributeDefinition;
 import org.jboss.as.controller.SimpleAttributeDefinitionBuilder;
 import org.jboss.as.controller.SimpleResourceDefinition;
@@ -630,7 +631,7 @@ class JdbcRealmDefinition extends SimpleResourceDefinition {
             }
 
             commonDependencies(serviceBuilder)
-                    .setInitialMode(ServiceController.Mode.ACTIVE)
+                    .setInitialMode(context.getRunningMode() == RunningMode.ADMIN_ONLY ? ServiceController.Mode.LAZY : ServiceController.Mode.ACTIVE)
                     .install();
         }
 


elytron/src/main/java/org/wildfly/extension/elytron/JdbcRealmDefinition.java
```diff
@@ -630,7 +631,7 @@ public void uninject() {
             }
 
             commonDependencies(serviceBuilder)
-                    .setInitialMode(ServiceController.Mode.ACTIVE)
+                    .setInitialMode(!context.isBooting() && context.getRunningMode() == RunningMode.ADMIN_ONLY ? ServiceController.Mode.LAZY : ServiceController.Mode.ACTIVE)
```
yersan: Hi @ashley-abdelsayed98 , have you considered to have the service LAZY as well when the embedded server starts? 

Datasources services are not UP when the embedded server starts, and if we  keep `!context.isBooting()`, we will end up with different states on the jdbc-realm service after adding a jdbc-realm and starting the embedded server again. Take a look at this sequence:
```
[disconnected /] embed-server --server-config=standalone.xml
[standalone@embedded /] /subsystem=elytron/jdbc-realm=MyRealm:add(principal-query=[{sql="SELECT * FROM Users WHERE username = ?", data-source=ExampleDS}])
{"outcome" => "success"}

[standalone@embedded /] /core-service=service-container:dump-services() | grep ExampleDS
Service \"org.wildfly.security.security-realm.MyRealm\" (class org.wildfly.extension.elytron.TrivialService) mode LAZY state DOWN (WAITING) (parent: jboss.a
s.server-controller) (dependencies: elytron.security-properties, elytron.provider-registration, org.wildfly.data-source.ExampleDS) (has unavailable dependen
cy)

[standalone@embedded /] stop-embedded-server
[disconnected /] embed-server --server-config=standalone.xml
[standalone@embedded /] /core-service=service-container:dump-services() | grep ExampleDS
Service \"org.wildfly.security.security-realm.MyRealm\" (class org.wildfly.extension.elytron.TrivialService) mode ACTIVE state DOWN (PROBLEM) (parent: jboss
.as.server-controller) (dependencies: elytron.security-properties, org.wildfly.data-source.ExampleDS, elytron.provider-registration) (has unavailable depend
ency)
```

I'm not sure if that could be a problem, but consistency getting the same status for the jdbc-realm sounds better, isn't it?

ashley-abdelsayed98: I see what you're saying, it would make sense for the state to be consistent.  I've updated the PR, thanks!

jmesnil:
@ashley-abdelsayed98 thanks!

PR #3890, opened 2019-08-09
[WFCORE-4375] Provide ability to easily apply certain JBoss module libraries to all deployments running in a server

yersan:
Wildfly core changes to allow the uses of an external folder as a module that can be added as a dependency for a deployment.

Basically, it uses an exiting service that creates a module from an external jar file expanding the functionality to scan a directory and create a module spec using all the Jar files found in that directory and subdirectories.

This feature should be still on hold since it is being blocked by https://issues.jboss.org/browse/WFCORE-4597, however, after discussing it, I open the issue to get some early feedback about the feature implementation.

Jira issue: https://issues.jboss.org/browse/WFCORE-4375

### DIFF
--- a/server/src/main/java/org/jboss/as/server/moduleservice/ExternalModuleService.java
+++ b/server/src/main/java/org/jboss/as/server/moduleservice/ExternalModuleService.java
@@ -62,15 +62,20 @@ public class ExternalModuleService implements Service<ExternalModuleService> {
      * @return true if valid, false otherwise
      */
     public boolean isValid(String externalModule) {
-        return new File(externalModule).exists();
+        File f = new File(externalModule);
+        return f.exists() && !f.isDirectory();
     }
 
-    public ModuleIdentifier addExternalModule(String externalModule, ServiceRegistry serviceRegistry, ServiceTarget serviceTarget) {
-        ModuleIdentifier identifier = ModuleIdentifier.create(EXTERNAL_MODULE_PREFIX + externalModule);
+    public ModuleIdentifier addExternalModule(String moduleName, ServiceRegistry serviceRegistry, ServiceTarget serviceTarget) {
+        return addExternalModule(moduleName, serviceRegistry, serviceTarget, moduleName);
+    }
+
+    public ModuleIdentifier addExternalModule(String moduleName, ServiceRegistry serviceRegistry, ServiceTarget serviceTarget, String path) {
+        ModuleIdentifier identifier = ModuleIdentifier.create(EXTERNAL_MODULE_PREFIX + moduleName);
         ServiceName serviceName = ServiceModuleLoader.moduleSpecServiceName(identifier);
         ServiceController<?> controller = serviceRegistry.getService(serviceName);
         if (controller == null) {
-            ExternalModuleSpecService service = new ExternalModuleSpecService(identifier, new File(externalModule));
+            ExternalModuleSpecService service = new ExternalModuleSpecService(identifier, new File(path));
             serviceTarget.addService(serviceName)
                     .setInstance(service)
                     .setInitialMode(Mode.ON_DEMAND)
--- a/server/src/main/java/org/jboss/as/server/moduleservice/ExternalModuleSpecService.java
+++ b/server/src/main/java/org/jboss/as/server/moduleservice/ExternalModuleSpecService.java
@@ -21,7 +21,21 @@
  */
 package org.jboss.as.server.moduleservice;
 
+import java.io.File;
+import java.io.IOException;
+import java.nio.file.Files;
+import java.nio.file.Path;
+import java.util.ArrayList;
+import java.util.Arrays;
+import java.util.Collections;
+import java.util.List;
+import java.util.TreeSet;
+import java.util.jar.JarFile;
+import java.util.stream.Stream;
+
 import org.jboss.as.server.deployment.module.ModuleDependency;
+import org.jboss.as.server.logging.ServerLogger;
+import org.jboss.logging.Logger;
 import org.jboss.modules.DependencySpec;
 import org.jboss.modules.ModuleDependencySpecBuilder;
 import org.jboss.modules.ModuleIdentifier;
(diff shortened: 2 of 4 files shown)


server/src/main/java/org/jboss/as/server/moduleservice/ExternalModuleSpecService.java
```diff
@@ -32,42 +46,80 @@
...
+                        currentName = jar.toString();
+                        JarFile jarFile = new JarFile(jar.toFile());
+                        this.jarFiles.add(jarFile);
+                        addResourceRoot(specBuilder, jarFile);
+                        log.debugf("Added %s jar file as resource root for %s module identifier", jar.toString(), moduleIdentifier.getName());
+                    }
+                }
+            }
         } catch (IOException e) {
+            throw ServerLogger.ROOT_LOGGER.errorOpeningZipFile(currentName);
```
bstansberry: Shouldn't this pass in the underlying IOException so it can be logged?

server/src/main/java/org/jboss/as/server/moduleservice/ExternalModuleSpecService.java
```diff
@@ -78,13 +130,25 @@ public synchronized void start(StartContext context) throws StartException {
...
     }
 
     @Override
     public synchronized void stop(StopContext context) {
-        VFSUtils.safeClose(jarFile);
-        jarFile = null;
+        for (JarFile jarFile : jarFiles) {
+            log.debugf("Closing %s jar file which was added as resource root for %s module identifier", jarFile.toString(), moduleIdentifier.getName());
+            VFSUtils.safeClose(jarFile);
+            log.debugf("Closed %s jar file which was added as resource root for %s module identifier", jarFile.toString(), moduleIdentifier.getName());
```
bstansberry: These debug statements don't add much, as jarFile.toString() is just getClass().getName() + "@" + Integer.toHexString(hashCode())

At most one per file should be sufficient.

bstansberry:
I had a couple minor comments but in general this seems ok.

server/src/main/java/org/jboss/as/server/moduleservice/ExternalModuleService.java
```diff
@@ -63,15 +63,20 @@
      * @return true if valid, false otherwise
      */
     public boolean isValid(String externalModule) {
-        return new File(externalModule).exists();
+        File f = new File(externalModule);
+        return f.exists() && !f.isDirectory();
```
bstansberry: AIUI this has the effect of preventing this new functionality being used to add directories via Class-Path entries in  MANIFEST.MF.

Which is fine; users doing that is not part of the scope of the RFE that is driving this.

If that is true then WFCORE-4375 is not a Feature Request; it is an Enhancement. There is no end user accessible feature here, just some internal changes to make things possible. The point matters because a Feature Request has different requirements from an Enhancement.

bstansberry:
This is not a requirement for this work, but if not done as part of it please file a JIRA and get it done for WF 19:

1) There should be a capability for this (probably registered by ServerService, a la https://github.com/wildfly/wildfly-core/pull/3889/files#diff-3570f28aaafabd8d9ad0212b763e6557)

2) The service type exposed by that capability should not be ExternalModuleService, but rather some new interface that ExternalModuleService implements, a la what I did with ControlledProcessStateService in https://github.com/wildfly/wildfly-core/pull/3889. Basic point is the capability should not expose start/stop/getValue.

server/src/main/java/org/jboss/as/server/moduleservice/ExternalModuleService.java
yersan: Correct, the intention here was to protect the uses of directory absolute paths defined in manifest Class-Path attribute.

Before this change, if a user configures a directory in Class-Path attribute, then an exception was thrown. Now if a directory is used, we warn and ignore. I think that's valid change, I cannot think in something relying on this exception to work, so warning+ignore sounds enough. 

Based on the above, @bstansberry can we still convert it as an enhancement?

server/src/main/java/org/jboss/as/server/moduleservice/ExternalModuleSpecService.java
yersan: Ups, sorry, some scrap left from the development

server/src/main/java/org/jboss/as/server/moduleservice/ExternalModuleSpecService.java
yersan: Correct, adding the exception to the error trace will log the original exception stacktrace too

server/src/main/java/org/jboss/as/server/moduleservice/ExternalModuleService.java
bstansberry: @yersan I think that's fine. I removed the feature label.

bstansberry:
@yersan If the two trivial remaining comment items here and a fix for WFCORE-4597 could be done by tomorrow that would be great as I'd like to get another core beta done with those in it.

yersan:
@bstansberry I'm still dealing with a minor change that probably will affect to this PR code. The idea is to move the creation of the global module from STRUCTURE to the end of DEPENDENCIES phase. That would add the new module that represents the global directory as the last dependency of the current deployment.

That would protect all the dependencies added by the server and will minimize the interferences coming from the jars included in the global directory. The classes added by the implicit server dependencies will be loaded before the classes included in a jar file found in the global directory, which are managed by the user. That could avoid problems like linkage errors.

A fast solution is to manipulate the order here: https://github.com/wildfly/wildfly-core/blob/master/server/src/main/java/org/jboss/as/server/deployment/module/ModuleSpecProcessor.java#L317 and put the global modules at the end, but it looks a bit weak.

Let me know if you prefer the simplest fast first and I push it together with https://issues.jboss.org/browse/WFCORE-4597 today

bstansberry:
@yersan I don't truly understand the issue but my basic advice is try to make this as much as possible like the handling of the ee subsystem 'global-modules' attribute, which has been in use for many years. The difference is just that the module is not static, it is dynamically generated.

Don't rush for the core release; I just wanted it included if it was done.

yersan:
Well, it seems it was not a good idea at all, I picked up WFCORE-4612 and WFCORE-4597 as first two commits with the hope of do not see changes once those two are merged. We will see how this looks like after merging them.

PR #3891, opened 2019-08-10
[WFCORE-4598] Migrating remoting maven module to new MSC API

ropalka:
https://issues.jboss.org/browse/WFCORE-4598

### DIFF
--- a/remoting/subsystem/src/main/java/org/jboss/as/remoting/HttpListenerRegistryService.java
+++ b/remoting/subsystem/src/main/java/org/jboss/as/remoting/HttpListenerRegistryService.java
@@ -2,11 +2,13 @@ package org.jboss.as.remoting;
 
 import io.undertow.server.ListenerRegistry;
 import org.jboss.msc.service.Service;
+import org.jboss.msc.service.ServiceBuilder;
 import org.jboss.msc.service.ServiceName;
 import org.jboss.msc.service.ServiceTarget;
 import org.jboss.msc.service.StartContext;
-import org.jboss.msc.service.StartException;
 import org.jboss.msc.service.StopContext;
+
+import java.util.function.Consumer;
 
 /**
  * Service that maintains a registry of all Undertow listeners, and the services that are registered on them.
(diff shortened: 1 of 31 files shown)


ropalka:
The hold label can be removed from this PR @bstansberry 
as https://github.com/wildfly/wildfly-core/pull/3922 solves 
JBoss MSC lock contention issue you were afraid of.

ropalka:
There is no more a reason to hold on this PR @bstansberry as JBoss MSC contention issues have been solved.

bstansberry:
I removed the hold because we are on to WF 19.

This needs very careful review. One thing to be careful of is any form of new public API in the controller module. I'm not saying that can't happen but in a perfect world we'd be removing things not adding.

ropalka:
I introduced ModelControllerOperationHandlerFactory in controller subsystem because I
wanted to make Consumer & Supplier fields final in AbstractModelControllerOperationHandlerFactoryService. Another possibility can be to make these Consumer and Supplier fields volatile and just introduce new AbstractModelControllerOperationHandlerFactoryService constructor. But I think having fields final is better solution @bstansberry .

remoting/subsystem/src/main/java/org/jboss/as/remoting/HttpListenerRegistryService.java
```diff
@@ -2,44 +2,66 @@
...
  *
  * @author Stuart Douglas
+ * @author <a href="mailto:ropalka@redhat.com">Richard Opalka</a>
  */
 public class HttpListenerRegistryService implements Service<ListenerRegistry> {
 
     @Deprecated
     public static final ServiceName SERVICE_NAME = ServiceName.JBOSS.append("http", "listener", "registry");
 
+    private final Consumer<ListenerRegistry> serviceConsumer;
```
jmesnil: Could using an `Optional<Consumer<ListenerRegistry>>` be an alternative?

remoting/subsystem/src/main/java/org/jboss/as/remoting/LocalOutboundConnectionWriteHandler.java
```diff
@@ -57,12 +56,11 @@ protected void revertUpdateToRuntime(OperationContext context, ModelNode operati
         if (handback != null && !handback.booleanValue()) {
             final ModelNode restored = Resource.Tools.readModel(context.readResource(PathAddress.EMPTY_ADDRESS));
             restored.get(attributeName).set(valueToRestore);
-            applyModelToRuntime(context, operation, restored);
+            applyModelToRuntime(context, operation);
```
jmesnil: this change is confusing.
At first glance, I thought that the change  would not take into account the `restored` model but it turns out that the parameters end up not being used in the `LocalOutboundConnectionService` constructor.
Is that correct?

remoting/subsystem/src/main/java/org/jboss/as/remoting/RemoteOutboundConnectionService.java
```diff
@@ -125,7 +119,7 @@ public void start(final StartContext context) throws StartException {
             URI finalUri = uri;
             authenticationConfiguration = () -> AUTH_CONFIGURATION_CLIENT.getAuthenticationConfiguration(finalUri, injectedContext);
         } else {
-            final SecurityRealm securityRealm = securityRealmInjectedValue.getOptionalValue();
+            final SecurityRealm securityRealm = securityRealmSupplier != null ? securityRealmSupplier.get() : null;
```
jmesnil: likewise, would some `Optional<Supplier<SecurityRealm>>` be more self-explanatory?

server/src/main/java/org/jboss/as/server/operations/NativeRemotingManagementAddHandler.java
```diff
@@ -70,7 +78,16 @@ public void performRuntime(OperationContext context, ModelNode operation, ModelN
         ManagementChannelRegistryService.addService(serviceTarget, endpointName);
         ManagementRemotingServices.installManagementChannelServices(serviceTarget,
                 endpointName,
-                new ModelControllerClientOperationHandlerFactoryService(),
+                new ModelControllerOperationHandlerFactory() {
```
jmesnil: That'd change the API but why do we need to instantiate the service when `ManagementRemotingServices.installManagementChannelServices` is called? 
Could we instead instantiate the service *inside* `ManagementRemotingServices.installManagementChannelServices`?

remoting/subsystem/src/main/java/org/jboss/as/remoting/HttpListenerRegistryService.java
ropalka: Yes, Optional is an alternative. But it depends on developer's preference. I'm not big fan of Optional.

remoting/subsystem/src/main/java/org/jboss/as/remoting/LocalOutboundConnectionWriteHandler.java
ropalka: Correct, I eliminated many unused fields in second commit of this PR.

remoting/subsystem/src/main/java/org/jboss/as/remoting/RemoteOutboundConnectionService.java
ropalka: Depends on developer's taste. But I have to agree with you it would be more self-explanatory.

server/src/main/java/org/jboss/as/server/operations/NativeRemotingManagementAddHandler.java
ropalka: See my comment above from 8-th of October @jmesnil .

jmesnil: After reading more carefully the code, I see now why you did it that way. thanks

jmesnil:
@ropalka I'm fine with merging your PR as it stands.
Do you want to change it wrt to `Optional` or can it go as it is now?

ropalka:
If you think Optional should be used then please create separate JIRA for that @jmesnil .
I am not going to change it in this PR. Thanks for reviewing this big PR.

jmesnil:
@ropalka thanks

PR #3892, opened 2019-08-12
WFCORE-4599 Handle wildfly-config.xml better for authentication

stuartwdouglas:
This was previously cached in a static, which does not work for
multiple deployments. This uses an empty authentication context
for the default (if it was not explicitly set), then explicitly
sets per-deployment ClassLoader defaults if the relevant file is
present in the deployment.

### DIFF
--- a/elytron/src/main/java/org/wildfly/extension/elytron/AuthenticationContextAssociationProcessor.java
+++ b/elytron/src/main/java/org/wildfly/extension/elytron/AuthenticationContextAssociationProcessor.java
@@ -19,12 +19,19 @@ package org.wildfly.extension.elytron;
 
 import static org.jboss.as.server.deployment.Attachments.MODULE;
 import static org.wildfly.extension.elytron.ElytronExtension.AUTHENTICATION_CONTEXT_KEY;
+import static org.wildfly.extension.elytron.SecurityActions.doPrivileged;
+
+import java.security.PrivilegedAction;
+import java.util.function.Supplier;
 
 import org.jboss.as.server.deployment.DeploymentPhaseContext;
 import org.jboss.as.server.deployment.DeploymentUnit;
 import org.jboss.as.server.deployment.DeploymentUnitProcessingException;
 import org.jboss.as.server.deployment.DeploymentUnitProcessor;
+import org.jboss.modules.ModuleClassLoader;
 import org.wildfly.security.auth.client.AuthenticationContext;
+import org.wildfly.security.auth.client.ElytronXmlParser;
+import org.wildfly.security.auth.client.InvalidAuthenticationConfigurationException;
 
 /**
  * A {@link DeploymentUnitProcessor} to associate a previously obtained {@link AuthenticationContext} with the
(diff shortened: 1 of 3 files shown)


darranl:
As I understand it this is a change in response to a regression caused by https://github.com/wildfly/wildfly-core/pull/3870 - it seems like quite a large change in behaviour to cover a regression.

I haven't seen which error is being discussed but from https://github.com/wildfly/wildfly-core/pull/3870 a simple change there could have been to set the global default back to 'null' so discovery can happen again.

Not necessarily against this change going in but I had already raised this Jira https://issues.jboss.org/browse/WFCORE-4573 to revisit the handling of wildfly-config.xml in deployments - this descriptor contains the configuration for multiple components.  The reason I had raised it as a feature request is one area to consider is the question of precedence - should the global default have priority over the deployment's configuration?  Should instead the two be merged?  If we merge what order should we use?  Once we using the deployments configuration in a DUP should other managed resources be available such as security domains / credential stores?

Going to add a couple of code comments but other than that no objection.

elytron/src/main/java/org/wildfly/extension/elytron/AuthenticationContextAssociationProcessor.java
```diff
@@ -17,14 +17,24 @@
  */
 package org.wildfly.extension.elytron;
 
+import static java.security.AccessController.doPrivileged;
```
darranl: This import should really be org.wildfly.extension.elytron.SecurityActions.doPrivileged that way a WildFlySecurityManager.isChecking() call will be made to check if we really do need the AccessController.doPrivileged().

elytron/src/main/java/org/wildfly/extension/elytron/AuthenticationContextAssociationProcessor.java
```diff
@@ -34,12 +44,57 @@
  */
 class AuthenticationContextAssociationProcessor implements DeploymentUnitProcessor {
 
+    private static final String[] FILES = {"wildfly-config.xml", "META-INF/wildfly-config.xml"};
```
darranl: If a DUP is now handling the parsing of the configuration should we allow it to be located in WEB-INF?

stuartwdouglas:
On Tue, 13 Aug 2019 at 21:17, Darran Lofthouse <notifications@github.com>
wrote:

> As I understand it this is a change in response to a regression caused by
> #3870 <https://github.com/wildfly/wildfly-core/pull/3870> - it seems like
> quite a large change in behaviour to cover a regression.
>

It's not really. If there is only a single deployment in the server then
this behaviour should be exactly the same. If there are multiple
deployments this will stop potential leakage of the wildfly-config.xml from
other deployments.


> I haven't seen which error is being discussed but from #3870
> <https://github.com/wildfly/wildfly-core/pull/3870> a simple change there
> could have been to set the global default back to 'null' so discovery can
> happen again.
>

A global default that is discovered once by the TCCL is 100% wrong for an
application server. Each deployment can have its own TCCL and its own
wildfly-config.xml, you can't just pick one at random every time like we
currently do.


> Not necessarily against this change going in but I had already raised this
> Jira https://issues.jboss.org/browse/WFCORE-4573 to revisit the handling
> of wildfly-config.xml in deployments - this descriptor contains the
> configuration for multiple components. The reason I had raised it as a
> feature request is one area to consider is the question of precedence -
> should the global default have priority over the deployment's
> configuration? Should instead the two be merged? If we merge what order
> should we use? Once we using the deployments configuration in a DUP should
> other managed resources be available such as security domains / credential
> stores?
>
Going to add a couple of code comments but other than that no objection.

elytron/src/main/java/org/wildfly/extension/elytron/AuthenticationContextAssociationProcessor.java
stuartwdouglas: This is only every run once, so it does not make any real difference.

elytron/src/main/java/org/wildfly/extension/elytron/AuthenticationContextAssociationProcessor.java
stuartwdouglas: If we want 100% backward compat then WEB-INF/classes and WEB-INF/classes/META-INF should probably also be present.

elytron/src/main/java/org/wildfly/extension/elytron/AuthenticationContextAssociationProcessor.java
darranl: Once per first request to each deployment whilst parallel requests are waiting for the lock.

elytron/src/main/java/org/wildfly/extension/elytron/AuthenticationContextAssociationProcessor.java
darranl: +1

darranl:
Overall the only scenario I see this breaking is if a user has configured the server with the 'wildfly.config.url' system property and has accidentally included a wildfly-config.xml in their deployment as now the deployment gets higher priority than the system property.

darranl:
Just thinking if a check could be made to also check if the system property has been set but that would only work without the line on boot setting the empty AuthenticationContext as well.

jmesnil:
@stuartwdouglas thanks

PR #3903, opened 2019-08-16
[WFCORE-4610] TestModule should not write modules in any *build/targe…

bstansberry:
…t or *dist/target dir

https://issues.jboss.org/browse/WFCORE-4610

### DIFF
--- a/testsuite/shared/src/main/java/org/jboss/as/test/module/util/TestModule.java
+++ b/testsuite/shared/src/main/java/org/jboss/as/test/module/util/TestModule.java
@@ -52,6 +52,8 @@ import org.jboss.shrinkwrap.api.spec.JavaArchive;
 public class TestModule {
 
     private static final Logger log  = Logger.getLogger(TestModule.class);
+    private static final String ILLEGAL_BUILD_DIR = "build" + File.separator + "target" + File.separator;
+    private static final String ILLEGAL_DIST_DIR = "dist" + File.separator + "target" + File.separator;
 
     private final String moduleName;
     private final File moduleXml;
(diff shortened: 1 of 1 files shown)


bstansberry:
retest this please

testsuite/shared/src/main/java/org/jboss/as/test/module/util/TestModule.java
```diff
@@ -293,6 +318,25 @@ private File getModulesDirectory() {
         return moduleDir;
     }
 
+    private static boolean isImmutableModulePath(String path) {
+        return path.contains(ILLEGAL_BUILD_DIR) || path.contains(ILLEGAL_DIST_DIR);
+    }
+
+    private static boolean isUnderCurrentProjectBuildDir(File moduleRoot) {
+        String buildDir = System.getProperty("project.build.directory");
```
yersan: CI errors are related; It looks like this will return always `null`, since `project.build.directory` is not passed by surefire maven plugin via system.properties. An alternative could be to use `Paths.get(System.getProperty("basedir"), "target").toString()`

${basedir} should not be null since it is always passed by surefire retrieved from the pom ${basedir} (although it is deprecated)

If so, I would bet this function will never return false

yersan: Ups, I mean, never will return false due to the `if (buildDir != null)` check

bstansberry: Thanks; I added that in; we'll see if it fixes it.

It's strange that the most recent run passed on Windows.

PR #3904, opened 2019-08-19
[WFCORE-4314] - verbose and recursive parameter in read-alias(es)

baranowb:
Improve read commands for elytron keystore.

### DIFF
--- a/elytron/src/main/java/org/wildfly/extension/elytron/ModifiableKeyStoreDecorator.java
+++ b/elytron/src/main/java/org/wildfly/extension/elytron/ModifiableKeyStoreDecorator.java
@@ -85,8 +85,20 @@ class ModifiableKeyStoreDecorator extends DelegatingResourceDefinition {
 
     static class ReadAliasesHandler extends ElytronRuntimeOnlyHandler {
 
+        static final SimpleAttributeDefinition RECURSIVE = new SimpleAttributeDefinitionBuilder(ElytronDescriptionConstants.RECURSIVE, ModelType.BOOLEAN, true)
+                .setAllowExpression(false)
+                .setDefaultValue(ModelNode.FALSE)
+                .build();
+
+        static final SimpleAttributeDefinition VERBOSE = new SimpleAttributeDefinitionBuilder(ElytronDescriptionConstants.VERBOSE, ModelType.BOOLEAN, true)
+                .setAllowExpression(false)
+                .setDefaultValue(ModelNode.TRUE)
+                .setRequires(RECURSIVE.getName())
+                .build();
+
         static void register(ManagementResourceRegistration resourceRegistration, ResourceDescriptionResolver descriptionResolver) {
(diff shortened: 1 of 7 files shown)


bstansberry:
A version bump is required if one hasn't already been done for the subsystem during dev of the same WF Core .Final release.

bstansberry:
And a transformer is required as the op can be executed on the profile resource.

elytron/src/main/java/org/wildfly/extension/elytron/ModifiableKeyStoreDecorator.java
```diff
@@ -85,8 +85,20 @@ public void registerOperations(ManagementResourceRegistration resourceRegistrati
 
     static class ReadAliasesHandler extends ElytronRuntimeOnlyHandler {
 
+        static final SimpleAttributeDefinition RECURSIVE = new SimpleAttributeDefinitionBuilder(ElytronDescriptionConstants.RECURSIVE, ModelType.BOOLEAN, true)
+                .setAllowExpression(false)
+                .setDefaultValue(new ModelNode(false))
```
bstansberry: ModelNode.FALSE

elytron/src/main/java/org/wildfly/extension/elytron/ModifiableKeyStoreDecorator.java
```diff
@@ -85,8 +85,20 @@ public void registerOperations(ManagementResourceRegistration resourceRegistrati
...
+                .build();
+
+        static final SimpleAttributeDefinition VERBOSE = new SimpleAttributeDefinitionBuilder(ElytronDescriptionConstants.VERBOSE, ModelType.BOOLEAN, true)
+                .setAllowExpression(false)
+                .setDefaultValue(new ModelNode(true))
```
bstansberry: ModelNode.TRUE

elytron/src/main/java/org/wildfly/extension/elytron/ModifiableKeyStoreDecorator.java
```diff
@@ -115,9 +135,14 @@ protected void executeRuntimeStep(OperationContext context, ModelNode operation)
                 .setAllowExpression(false)
                 .build();
 
+        static final SimpleAttributeDefinition VERBOSE = new SimpleAttributeDefinitionBuilder(ElytronDescriptionConstants.VERBOSE, ModelType.BOOLEAN, true)
+                .setAllowExpression(false)
+                .setDefaultValue(new ModelNode(true))
```
bstansberry: ModelNode.TRUE

darranl:
Before asking an engineer for a model bump can we please double check ourselves if one has already been performed (in this case it already has) - those handling reviews tend to be closer to this codebase so can quickly check.

I would also add although very light this really should be handled as an RFE, this is a change in behaviour to public API - most of the steps should be a formality but it needs to the usual plan, test, and docs.

baranowb:
Darn it, missed Darrans comment. So bump to 8 is enough?

baranowb:
This depends on https://github.com/wildfly/wildfly-core/pull/4018

PR #3906, opened 2019-08-19
[WFLY-4612] ExternalModuleDependencySpecService should make dependenc…

bstansberry:
…ies on 'javaee.api' optional

https://issues.jboss.org/browse/WFCORE-4612

This allows app dependencies added via Class-Path entries in MANIFEST.MF or from files in $JBOSS_HOME/lib/ext or the java.ext.dirs dirs to work even with no javaee.api module provisioned. Of course the dependency can't need any of the javaee.api artifacts.

@yersan @jfdenise FYI.


bstansberry:
Closing in favor of #3907

PR #3907, opened 2019-08-19
[WFCORE-4612] ExternalModuleDependencySpecService should make depende...

bstansberry:
…ncies on 'javaee.api' optional

https://issues.jboss.org/browse/WFCORE-4612

This allows app dependencies added via Class-Path entries in MANIFEST.MF or from files in $JBOSS_HOME/lib/ext or the java.ext.dirs dirs to work even with no javaee.api module provisioned. Of course the dependency can't need any of the javaee.api artifacts.

@yersan @jfdenise FYI.

Replaces #3906 which had an incorrect branch name.

I added a second commit here that goes beyond making the javaee.api dep optional and instead adds individual optional deps on the various EE API modules. This allows things to depend on whatever is provisioned without requiring irrelevant things to be provisioned.

### DIFF
--- a/server/src/main/java/org/jboss/as/server/moduleservice/ExternalModuleSpecService.java
+++ b/server/src/main/java/org/jboss/as/server/moduleservice/ExternalModuleSpecService.java
@@ -23,6 +23,7 @@ package org.jboss.as.server.moduleservice;
 
 import org.jboss.as.server.deployment.module.ModuleDependency;
 import org.jboss.modules.DependencySpec;
+import org.jboss.modules.ModuleDependencySpecBuilder;
 import org.jboss.modules.ModuleIdentifier;
 import org.jboss.modules.ModuleSpec;
 import org.jboss.modules.ResourceLoaderSpec;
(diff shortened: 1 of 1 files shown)


jamezp:
I'm sure this has been considered, but I just didn't see any comments about it. This feels a bit odd being in core as any new dependency in full will require a change in core and I assume a core upgrade before the new dependency in full can be merged.

bstansberry:
@jamezp Yes, the dual maintenance is my main concern with this.

An alternative I'd considered was to add a new 'wildflyee.api' module to core with these in it as optional deps. And then javaee.api in full would just become an alias to wildflyee.api, with a comment explaining where to go to add more specs. So only 1 place to maintain.  But because Galleon uses passive+ provisioning for WF (https://docs.wildfly.org/galleon/#_feature_pack_original_effective_package_set) the effect is Galleon would provision all the 'optional' packages, defeating the slimming goal (somewhat.)

Hmm, perhaps I could make 'wildflyee.api' a dynamic module. So Galleon is unaware.

yersan:
@bstansberry Are you planning to squash this before merging? It is going to introduce conflict with one of my PR, so I was planning to update my PR onto these changes resolving the conflicts, but there are two commits, just asking

bstansberry:
@yersan I rebased; thanks for the reminder!

PR #3908, opened 2019-08-20
[WFCORE-4336] Add the ability for a user to add custom filters for loggers and log handlers

jamezp:
JIRA: https://issues.jboss.org/browse/WFCORE-4336
Model Upgrade JIRA: https://issues.jboss.org/browse/WFCORE-4591
Feature JIRA: https://issues.jboss.org/browse/EAP7-1199

For testing with the subsystem tests I added some modules to the subsystem test resources. I'm not sure whether or not this was a good idea as these will require maintenance. However there is a potential to catch issues with unit tests which seems like a decent trade-off for now.

For the integration test I also add a module. I didn't see, or I'm unaware of, a standard way of doing this. If I missed that please let me know and I'll fix it.

I also went with using two attributes `constructor-properties` and `properties` for configuring the filter. I had considered `properties` with a boolean to indicate whether or not this was a constructor property, however this seemed less user friendly as the stand key/value pair properties are already widely used and likely understood.

### DIFF
--- a/logging/src/main/java/org/jboss/as/logging/filters/FilterResourceDefinition.java
+++ b/logging/src/main/java/org/jboss/as/logging/filters/FilterResourceDefinition.java
@@ -1,0 +1,250 @@
+/*
+ * JBoss, Home of Professional Open Source.
+ *
+ * Copyright 2019 Red Hat, Inc., and individual contributors
+ * as indicated by the @author tags.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.jboss.as.logging.filters;
+
+import static org.jboss.as.logging.CommonAttributes.CLASS;
+import static org.jboss.as.logging.CommonAttributes.MODULE;
+import static org.jboss.as.logging.CommonAttributes.PROPERTIES;
+
+import java.util.Arrays;
+import java.util.Collection;
+import java.util.LinkedHashMap;
+import java.util.List;
+import java.util.Map;
+
+import org.jboss.as.controller.AttributeDefinition;
+import org.jboss.as.controller.OperationContext;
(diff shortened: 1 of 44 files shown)


logging/src/main/java/org/jboss/as/logging/filters/FilterResourceDefinition.java
```diff
@@ -0,0 +1,250 @@
...
+            super.populateModel(context, operation, resource);
+        }
+
+        @Override
+        public void performRuntime(final OperationContext context, final ModelNode operation, final ModelNode model, final LogContextConfiguration logContextConfiguration) throws OperationFailedException {
+            final String name = context.getCurrentAddressValue();
+            FilterConfiguration configuration = logContextConfiguration.getFilterConfiguration(name);
+            final String className = CLASS.resolveModelAttribute(context, model).asString();
+            final ModelNode moduleNameNode = MODULE.resolveModelAttribute(context, model);
+            final String moduleName = moduleNameNode.isDefined() ? moduleNameNode.asString() : null;
```
bstansberry: No need to change this but in the future you can do

final String moduleName = MODULE.resolveModelAttribute(context, model).asStringOrNull();

logging/src/main/java/org/jboss/as/logging/filters/FilterResourceDefinition.java
```diff
@@ -0,0 +1,250 @@
...
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.jboss.as.logging.filters;
+
+import static org.jboss.as.logging.CommonAttributes.CLASS;
+import static org.jboss.as.logging.CommonAttributes.MODULE;
```
bstansberry: @jamezp I reviewing this I noticed the definitions of these is incorrect, at least for this use case and the other ones I could see. So I filed https://issues.jboss.org/projects/WFCORE/issues/WFCORE-4631. Not necessary to correct in this PR as it is a pre-existing issue.

logging/src/main/java/org/jboss/as/logging/filters/FilterResourceDefinition.java
jamezp: I saw that on a PR the other day and was quite surprised I didn't know about it :)

logging/src/main/java/org/jboss/as/logging/filters/FilterResourceDefinition.java
jamezp: That seems correct. I'll get a fix for that probably today.

PR #3914, opened 2019-08-22
[WFCORE-4617] jboss-cli.sh does not error on invalid options

spyrkob:
JIRA: https://issues.jboss.org/browse/WFCORE-4617

### DIFF
--- a/cli/src/main/java/org/jboss/as/cli/impl/CliLauncher.java
+++ b/cli/src/main/java/org/jboss/as/cli/impl/CliLauncher.java
@@ -245,7 +245,12 @@ public class CliLauncher {
                         WildFlySecurityManager.setPropertyPrivileged(prop, "true");
                         systemPropertyKeys.add(prop);
                     }
-                } else if(!arg.equals("-XX:")) {// skip system properties and jvm options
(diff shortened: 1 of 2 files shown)


darranl:
Do keep in mind with a change like this - all scripts that already exist that pass in an unrecognised option will now fail even if that option could be ignored.

spyrkob:
@darranl as far as I can tell, currently if an option is not recognised, it will be interpreted as a command and the script will fail when trying to execute this command. The difference with the change should be earlier failure (before connection attempt) and new error message.

darranl:
@spyrkob I understand the purpose of this change, but that doesn't eliminate the possibility that previously "working" calls to this script may now start failing - these calls could be coming from existing scripts such as cron jobs etc...

jfdenise:
@darranl , we are replacing an error message with another one. We don't have "working" script that would suddenly start to fail.

darranl:
@jfdenise Ok if we are happy this is replacing one error state with another error state - sounds good and we shouldn't have issues with user scripts.

bstansberry:
@jfdenise If you are ok with this please approve via github approval.

PR #3922, opened 2019-08-27
[WFCORE-4629] Upgrading JBoss MSC to 1.4.9.Final

ropalka:
Incorporates:
 * [MSC-247] Allow also provided not just aliased services to be accessible
 * [MSC-248] Fix Suplier.get() lock contention
 * [MSC-250] StabilityMonitor problematic and failed services are improperly reported


ropalka:
This PR addresses https://issues.jboss.org/browse/WFLY-12372 @bstansberry

PR #3937, opened 2019-08-30
[WFCORE-4640] Add a DUP to the Elytron subsystem that adds the 'javax.security.auth.message.api' and 'javax.security.jacc.api' modules to the deployment

fjuma:
https://issues.jboss.org/browse/WFCORE-4640


bstansberry:
This allows testsuite/integration/elytron in full to pass when my https://issues.jboss.org/browse/WFLY-12464 work is applied. :) Thanks!

PR #3943, opened 2019-09-05
[WFCORE-4630] Switch from javax.inject:javax.inject to jakarta.inject:jakarta.inject-api

yersan:
Switch from javax.inject:javax.inject to jakarta.inject:jakarta.inject-api

Jira issue: https://issues.jboss.org/browse/WFCORE-4630


jamezp:
I only placed on a hold on this as I may include in a JSON-P PR I've got coming up as it will be a multi-stage commit between core and WildFly.

jamezp:
Replaced by #3946

PR #3950, opened 2019-09-10
WFCORE-4656 fix ConfigurationPersistenceException due to incorrect em…

soul2zimate:
…pty ALLOWED_ORIGINS list parsing

issue: https://issues.jboss.org/browse/WFCORE-4656

First add then clear ALLOWED_ORIGINS list cause ConfigurationPersistenceException at next server reload, because the parser doesn't handle the empty list properly.

### DIFF
--- a/host-controller/src/main/java/org/jboss/as/host/controller/parsing/HostXml_10.java
+++ b/host-controller/src/main/java/org/jboss/as/host/controller/parsing/HostXml_10.java
@@ -423,9 +423,7 @@ final class HostXml_10 extends CommonXml implements ManagementXmlDelegate {
                         break;
                     }
                     case ALLOWED_ORIGINS: {
-                        for (String origin : reader.getListAttributeValue(i)) {
-                            HttpManagementResourceDefinition.ALLOWED_ORIGINS.parseAndAddParameterElement(origin, addOp, reader);
-                        }
+                        HttpManagementResourceDefinition.ALLOWED_ORIGINS.getParser().parseAndSetParameter(HttpManagementResourceDefinition.ALLOWED_ORIGINS, value, addOp, reader);
                         break;
                     }
                     default:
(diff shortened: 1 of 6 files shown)


soul2zimate:
I have updated this one to include test.

host-controller/src/main/java/org/jboss/as/host/controller/parsing/HostXml_10.java
```diff
@@ -423,9 +423,7 @@ private void parseHttpManagementInterfaceAttributes(XMLExtendedStreamReader read
                         break;
                     }
                     case ALLOWED_ORIGINS: {
-                        for (String origin : reader.getListAttributeValue(i)) {
-                            HttpManagementResourceDefinition.ALLOWED_ORIGINS.parseAndAddParameterElement(origin, addOp, reader);
-                        }
+                        HttpManagementResourceDefinition.ALLOWED_ORIGINS.getParser().parseAndSetParameter(HttpManagementResourceDefinition.ALLOWED_ORIGINS, value, addOp, reader);
```
yersan: Just out of curiosity, should not we fix this also in the parsers for older model versions before 10?, I don't know if there could be possibilities of using older namespaces with newer server code. Just a doubt.

soul2zimate: me neither, I'm not aware such case. but if that's required, I can apply the same fix to old versions as well.

bstansberry: Thanks @soul2zimate. If you want to send up a separate PR for older parsers that's fine but it's not required.  The steps to reproduce in the JIRA would result in the config being persisted, which would be with the latest namespace. So for an older namespace to be involved the "" value would have to be in the original xml, which is a corner case.

PR #3952, opened 2019-09-11
[WFCORE-4659] reverting change of not defining the static auth context

ochaloup:
https://issues.jboss.org/browse/WFCORE-4659

This is reversion of part of the WFCORE-4599 where global static auth
context was disabled to be set to empty and it was expected
every component will manage the context on its own.
That makes troubles for transaction recovery as it has no such context
and it depended on the the global one workaround.
For the time being this change is to get the workaround to work again
but the proper fix should follow.

/CC @stuartwdouglas @darranl


jmesnil:
@ochaloup thanks

PR #3956, opened 2019-09-12
[WFCORE-4664] Upgrade wildfly-build-tools to 1.2.12.Final

bstansberry:
https://issues.jboss.org/browse/WFCORE-4664


jmesnil:
@bstansberry thanks

PR #3958, opened 2019-09-18
[WFCORE-4631] Remove the reconnect tests which check if the operation…

jamezp:
…s fail due to a UDP syslog connection failure. These tests rely on timing and only log a message once therefore they cannot be unit tests.

One thing to note is on my original @Ignore I ignored a wrong test and somehow got lucky the other did not fail. This test has been re-enabled and the test that should have been ignored has been removed.

https://issues.jboss.org/browse/WFCORE-4631


jmesnil:
@jamezp thanks

PR #3961, opened 2019-09-24
[WFCORE-4679] Update the mock-server-netty dep; specify the logback dep

bstansberry:
https://issues.jboss.org/browse/WFCORE-4679

If this mock-server-netty upgrade is a problem, then an alternative approach to getting rid of the annoying vulnerability warns is to fix the logback version to 1.2.3 in the root pom dependencyManagement.


jmesnil:
@bstansberry thanks

PR #3962, opened 2019-09-26
[WFCORE-4678] Use a more recent version of guava

bstansberry:
https://issues.jboss.org/browse/WFCORE-4678


jmesnil:
@bstansberry thanks

PR #3965, opened 2019-09-30
[WFCORE-4686] Add a profile to add the log manager to the boot class path and add set the log manager for Eclipse OpenJ9.

jamezp:
https://issues.jboss.org/browse/WFCORE-4686

Also ignore the standalone script tests that check for GC logging. This could be a separate PR if desired.


jamezp:
OpenJ9 builds is https://ci.wildfly.org/viewLog.html?buildId=169985&

jamezp:
I'm gonna close this in favor of a new PR coming up which contains more fixes for Eclipse OpenJ9.

PR #3963, opened 2019-09-27
[WFCORE-4681] Add MP OpenAPI constants; bump Jandex to current release

MikeEdgar:
Fixes WFCORE-4681, in support of WFLY-12313

PR #3972, opened 2019-10-04
WFCORE-4702 Add deployment phases for MicroProfile Fault Tolerance

rhusar:
https://issues.jboss.org/browse/WFCORE-4702

### DIFF
--- a/server/src/main/java/org/jboss/as/server/deployment/Phase.java
+++ b/server/src/main/java/org/jboss/as/server/deployment/Phase.java
@@ -447,6 +447,7 @@ public enum Phase {
     public static final int DEPENDENCIES_MICROPROFILE_HEALTH            = 0x1870;
     public static final int DEPENDENCIES_MICROPROFILE_OPENTRACING       = 0x1880;
     public static final int DEPENDENCIES_MICROPROFILE_OPENAPI           = 0x1890;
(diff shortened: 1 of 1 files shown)


rsvoboda:
@msve seems MP FT is in move

server/src/main/java/org/jboss/as/server/deployment/Phase.java
```diff
@@ -446,6 +446,7 @@ public Phase next() {
     public static final int DEPENDENCIES_MICROPROFILE_METRICS           = 0x1860;
     public static final int DEPENDENCIES_MICROPROFILE_HEALTH            = 0x1870;
     public static final int DEPENDENCIES_MICROPROFILE_OPENTRACING       = 0x1880;
+    public static final int DEPENDENCIES_MICROPROFILE_FAULT_TOLERANCE   = 0x1890;
```
jmesnil: @rhusar could you bump the constants please? I'm going to merge #3963 that takes these values.

rhusar: Done; would have needed a rebase anyway.

rhusar:
@jmesnil This has been updated as you requested and rebased so this is ready (I am realizing we probably need to proactively ping here but I assumed there is a notification send by github).

rhusar:
I see there is now an option to ask for re-review, so that is probably what was needed to get proper notifications.

jmesnil:
@rhusar thanks

PR #3973, opened 2019-10-07
[WFCORE-4680] Redact passwords in the sun.java.command logging

ivassile:
https://issues.jboss.org/browse/WFCORE-4680

### DIFF
--- a/host-controller/src/main/java/org/jboss/as/host/controller/HostControllerService.java
+++ b/host-controller/src/main/java/org/jboss/as/host/controller/HostControllerService.java
@@ -41,6 +41,8 @@ import java.util.concurrent.SynchronousQueue;
 import java.util.concurrent.ThreadFactory;
 import java.util.concurrent.ThreadPoolExecutor;
 import java.util.concurrent.TimeUnit;
+import java.util.regex.Matcher;
+import java.util.regex.Pattern;
 
 import org.jboss.as.controller.CapabilityRegistry;
 import org.jboss.as.controller.ControlledProcessState;
@@ -127,7 +129,19 @@ public class HostControllerService implements Service<AsyncFuture<ServiceContain
             final StringBuilder b = new StringBuilder(8192);
             b.append("Configured system properties:");
             for (String property : new TreeSet<String>(properties.stringPropertyNames())) {
-                String propVal = property.toLowerCase(Locale.getDefault()).contains("password") ? "<redacted>" : properties.getProperty(property, "<undefined>");
+                String propVal = property.toLowerCase(Locale.ROOT).contains("password") ? "<redacted>" : properties.getProperty(property, "<undefined>");
+                if (property.toLowerCase(Locale.ROOT).equals("sun.java.command") && !propVal.isEmpty()) {
+                    Pattern pattern = Pattern.compile("(-D(?:[^ ])+=)((?:[^ ])+)");
+                    Matcher matcher = pattern.matcher(propVal);
+                    StringBuffer sb = new StringBuffer(propVal.length());
+                    while (matcher.find()) {
+                        if (matcher.group(1).contains("password")) {
+                            matcher.appendReplacement(sb, Matcher.quoteReplacement(matcher.group(1) + "<redacted>"));
+                        }
+                    }
+                    matcher.appendTail(sb);
+                    propVal = sb.toString();
+                }
                 b.append("\n\t").append(property).append(" = ").append(propVal);
             }
             ServerLogger.CONFIG_LOGGER.debug(b);
@@ -137,7 +151,7 @@ public class HostControllerService implements Service<AsyncFuture<ServiceContain
                 final Map<String, String> env = System.getenv();
                 b.append("Configured system environment:");
                 for (String key : new TreeSet<String>(env.keySet())) {
-                    String envVal = key.toLowerCase(Locale.getDefault()).contains("password") ? "<redacted>" : env.get(key);
+                    String envVal = key.toLowerCase(Locale.ROOT).contains("password") ? "<redacted>" : env.get(key);
                     b.append("\n\t").append(key).append(" = ").append(envVal);
                 }
                 ServerLogger.CONFIG_LOGGER.trace(b);
--- a/server/src/main/java/org/jboss/as/server/ApplicationServerService.java
+++ b/server/src/main/java/org/jboss/as/server/ApplicationServerService.java
@@ -32,6 +32,8 @@ import java.util.Properties;
 import java.util.ServiceConfigurationError;
 import java.util.ServiceLoader;
 import java.util.TreeSet;
+import java.util.regex.Matcher;
+import java.util.regex.Pattern;
 
 import org.jboss.as.controller.ControlledProcessState;
 import org.jboss.as.controller.RunningModeControl;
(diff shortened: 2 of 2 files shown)


server/src/main/java/org/jboss/as/server/ApplicationServerService.java
```diff
@@ -109,6 +111,20 @@ public synchronized void start(final StartContext context) throws StartException
             b.append(ServerLogger.ROOT_LOGGER.configuredSystemPropertiesLabel());
             for (String property : new TreeSet<String>(properties.stringPropertyNames())) {
                 String propVal = property.toLowerCase(Locale.getDefault()).contains("password") ? "<redacted>" : properties.getProperty(property, "<undefined>");
+                if (property.toLowerCase(Locale.getDefault()).equals("sun.java.command") && !propVal.isEmpty()) {
```
jamezp: I wonder if 1 we should care about the property and 2 if we should have a more generic way of checking for other keywords where the password might not want to be printed.

ivassile: @jamezp 
>I wonder if 1 we should care about the property
"sun.java.command" exposes all system properties and makes password masking on individual properties obsolete.    
>2 if we should have a more generic way of checking for other keywords where the password might not want to be printed.
How to determine that a property is a password type?

jamezp: @ivassile It seems like bad practice to me to pass a password via the command line. That said my thought was to maybe have some kind of map or set of possible system properties which could be filtered. I feel like we've done this elsewhere, but I haven't had a chance to look.

This is just more of an idea though and I could be way off basis.

bstansberry: That fix fulfilled the request from the Red Hat support folks who wanted it. I deliberately kept it simple as the whole problem can be avoided by turning the level of the log category to INFO.

I see WFCORE-4680 as a bug fix correcting a hole in my original work. Configuration options to control this would be a separate RFE, IMO.

jamezp: This makes sense to me then. I just didn't know if we had random other properties we wanted replaced. I hadn't realized this was a follow up of the original.

host-controller/src/main/java/org/jboss/as/host/controller/HostControllerService.java
```diff
@@ -128,6 +130,20 @@ public void start(StartContext context) throws StartException {
...
+                if (property.toLowerCase(Locale.getDefault()).equals("sun.java.command") && !propVal.isEmpty()) {
+                    Pattern pattern = Pattern.compile("(-D(?:[^ ])+=)((?:[^ ])+)");
+                    Matcher matcher = pattern.matcher(propVal);
+                    StringBuffer sb = new StringBuffer(propVal.length());
+                    while (matcher.find()) {
+                        String option = matcher.group(2);
+                        if (matcher.group(1).contains("password")) {
+                            option = "<redacted>";
+                        }
+                        matcher.appendReplacement(sb, Matcher.quoteReplacement(matcher.group(1) + option));
```
jamezp: I think we should just make this:
```
if (matcher.group(1).contains("password")) {
   matcher.appendReplacement(sb, Matcher.quoteReplacement(matcher.group(1) + "<redacted>"));
}
```
There's not a need to add the overhead of replacing all values.

host-controller/src/main/java/org/jboss/as/host/controller/HostControllerService.java
```diff
@@ -128,6 +130,20 @@ public void start(StartContext context) throws StartException {
             b.append("Configured system properties:");
             for (String property : new TreeSet<String>(properties.stringPropertyNames())) {
                 String propVal = property.toLowerCase(Locale.getDefault()).contains("password") ? "<redacted>" : properties.getProperty(property, "<undefined>");
+                if (property.toLowerCase(Locale.getDefault()).equals("sun.java.command") && !propVal.isEmpty()) {
+                    Pattern pattern = Pattern.compile("(-D(?:[^ ])+=)((?:[^ ])+)");
+                    Matcher matcher = pattern.matcher(propVal);
+                    StringBuffer sb = new StringBuffer(propVal.length());
```
jamezp: Slight nitpick, this should probably be a `StringBuilder`. We shouldn't need a thread safe object here.

server/src/main/java/org/jboss/as/server/ApplicationServerService.java
```diff
@@ -109,6 +111,20 @@ public synchronized void start(final StartContext context) throws StartException
...
+                    Pattern pattern = Pattern.compile("(-D(?:[^ ])+=)((?:[^ ])+)");
+                    Matcher matcher = pattern.matcher(propVal);
+                    StringBuffer sb = new StringBuffer(propVal.length());
+                    while (matcher.find()) {
+                        String option = matcher.group(2);
+                        if (matcher.group(1).contains("password")) {
+                            option = "<redacted>";
+                        }
+                        matcher.appendReplacement(sb, Matcher.quoteReplacement(matcher.group(1) + option));
+                    }
```
jamezp: Sames comments as above for this one too.

host-controller/src/main/java/org/jboss/as/host/controller/HostControllerService.java
ivassile: appendReplacement(StringBuffer sb, String replacement) implementation is only for StringBuffer [1]. StringBuilder is available in Java 11 [2].
[1] https://docs.oracle.com/javase/8/docs/api/java/util/regex/Matcher.html
[2] https://docs.oracle.com/en/java/javase/11/docs/api/java.base/java/util/regex/Matcher.html

host-controller/src/main/java/org/jboss/as/host/controller/HostControllerService.java
ivassile: Corrected it. Thanks.

host-controller/src/main/java/org/jboss/as/host/controller/HostControllerService.java
jamezp: Oops, my mistake. I was definitely looking at Java 11. Not a big deal at all :)

host-controller/src/main/java/org/jboss/as/host/controller/HostControllerService.java
```diff
@@ -128,6 +130,18 @@ public void start(StartContext context) throws StartException {
             b.append("Configured system properties:");
             for (String property : new TreeSet<String>(properties.stringPropertyNames())) {
                 String propVal = property.toLowerCase(Locale.getDefault()).contains("password") ? "<redacted>" : properties.getProperty(property, "<undefined>");
+                if (property.toLowerCase(Locale.getDefault()).equals("sun.java.command") && !propVal.isEmpty()) {
```
jamezp: Sorry I missed this previously. These should probably be `Locale.ROOT`.

ivassile: Also corrected the other two places.

jmesnil:
@ivassile thanks

PR #3975, opened 2019-10-07
WFCORE-4656 port fix parsers for old namespaces

soul2zimate:
issue: https://issues.jboss.org/browse/WFCORE-4656

Follow up discussion in https://github.com/wildfly/wildfly-core/pull/3950


jmesnil:
@soul2zimate thanks

PR #3981, opened 2019-10-11
[WFCORE-4683] User with Monitor role can stop the servers in domain mode

gaol:
Issue: https://issues.jboss.org/browse/WFCORE-4683


jmesnil:
@gaol thanks

PR #3983, opened 2019-10-15
[WFCORE-4710] Don't arbitrarily turn unescaped mbean names into quote…

bstansberry:
…d values

https://issues.jboss.org/browse/WFCORE-4710


jmesnil:
@bstansberry thanks

PR #3984, opened 2019-10-15
[WFCORE-4338] Race condition on PID file allows for termination of ar…

gaol:
…bitrary processes by local users

Issue: https://issues.jboss.org/browse/WFCORE-4338


jmesnil:
@gaol thanks

PR #3986, opened 2019-10-15
[WFCORE-4714]  Upgrade galleon-plugins to 4.0.4.Final

bstansberry:
https://issues.jboss.org/browse/WFCORE-4714


jmesnil:
@bstansberry thanks

PR #3992, opened 2019-10-18
[WFCORE-4720] Clarify that SetupActions.dependencies() cannot return …

bstansberry:
…null; also some other javadoc tweaks

https://issues.jboss.org/browse/WFCORE-4720

Note that with https://github.com/wildfly/wildfly/pull/12727 the only impl in WildFly that returned null will no longer do so. There are a couple callers of this method in WildFly that assume it will not return null. So I figure the best thing to do is enhance the contract to require the best behavior.


jmesnil:
@bstansberry thanks

PR #4002, opened 2019-11-04
[WFCORE-2147] Enable property replacements for permissions and jboss-permissions files

yersan:
Enables the property replacements in permissions.xml and jboss-permissions.xml by using a Function passed by wildfly. This function will contain the resolved. We use a function there since all the java classes and code for the resolver resides in wildfly instead of core. 

An alternative approach to avoid pass a function could have been to move all the required dependencies from wildfly to wildfly core, however, the approach of using a function carries fewer modifications and it is enough. 

This is an RFE, tests will be added in wildfly.

Jira issue: https://issues.jboss.org/browse/WFCORE-2147


yersan:
I've just force-pushed the PR just to grab all the master changes since this branch was created, testsuite should pass, no changes added.

darranl:
@yersan Do you have the link to the WildFly pull request that goes with this?

yersan:
Hi @darranl , yes, here is: https://github.com/wildfly/wildfly/pull/12768

darranl:
Retest this please

PR #4007, opened 2019-11-08
[WFCORE-4745] Use the correct version properties in the module.xml fi…

jamezp:
…le and define the maven.repo.local system property in the surefire system properties

https://issues.jboss.org/browse/WFCORE-4745


jmesnil:
@jamezp thanks

PR #4011, opened 2019-11-14
[WFCORE-4733] Rename a Host Controller does not work properly

yersan:
LocalHostControllerInfoImpl stores information about the environment of the local host controller. When a  Host Controller is reloaded, the current host controller name is kept to allow the embedded host controller to work with fictitious names that are not finally persisted in the host.xml. See https://issues.jboss.org/browse/WFCORE-2933 for more information

The problem is the LocalHostControllerInfoImpl#overrideLocalHostName flag is never cleared when a new name is read from the host.xml. That means, when we rename a host controller and later we reload it, the old name is still taken. This patch fixes this situation.

Jira issue: https://issues.jboss.org/browse/WFCORE-4733


jmesnil:
@yersan thanks

PR #4013, opened 2019-11-16
[WFCORE-1110] Add support for custom headers to the HTTP management endpoints.

darranl:
Test cases and community documentation are in progress but this PR contains the implementation ready to get some CI and feedback in case anything needs to change.

https://issues.jboss.org/browse/WFCORE-1110

Proposal pull request https://github.com/wildfly/wildfly-proposals/pull/263


darranl:
Can we get these into WFWIP issues and we can start to work through them?

tterem:
@darranl I've created WFWIP issues.

darranl:
Perfect, thank you @tterem I will work through them one by one in the morning.

jmesnil:
@darranl the Version11StabilityTestCase did not pass on Windows (while it does on Linux).

domain-http/interface/src/main/java/org/jboss/as/domain/http/server/StaticHeadersHandler.java
```diff
@@ -0,0 +1,79 @@
...
+        private final HttpString headerName;
+        private final String value;
+
+        HeaderConstant(final String headerName, final String value) {
+            this.headerName = new HttpString(headerName);
+            this.value = value;
+        }
+
+        void apply(HttpServerExchange exchange) {
+            exchange.getResponseHeaders().add(headerName, value);
```
jmesnil: @darranl According to the dev analysis, [Configured headers will not override any headers added by the corresponding endpoint](https://github.com/wildfly/wildfly-proposals/pull/263/files#diff-7135813aa89d8b31636212899b87ba26R63)

We should check here that the response headers does not already contain the header.

darranl: I will be rewording that analysis section a little further after some debugging one of the WIPs today but in general that kind of check is not likely possible.

I have to set the headers as the request is on the way in as by the time it is on it's way out a handler could have already initiated sending the request to the client meaning I have no opportunity on the way out.

So for that sentence what I was really trying to say was we can set a header on the way in but as we don't know which headers an endpoint may also set we can not override that behaviour.

darranl:
@jmesnil The stability test may end up getting itself deleted unless I see an easy fix for Windows, it was supposed to be an early warning but if it ends up causing more problems than it resolves.

darranl:
@jmesnil so it looks like on the Windows job the number of bytes read is greater than on Linux, I suspect the checkout has changed the line endings making it a different file.  I think I could either alter the parsing to eliminate the line endings or skip it on Windows.

jmesnil:
@darranl thanks

PR #4017, opened 2019-11-19
[WFCORE-4750] Using regex for role in Elytron

Skyllarr:
Jira: [WFCORE-4750](https://issues.jboss.org/browse/WFCORE-4750)

Requires [ELY-1909](https://issues.jboss.org/browse/ELY-1909)

### DIFF
--- a/elytron/src/main/java/org/wildfly/extension/elytron/ElytronDescriptionConstants.java
+++ b/elytron/src/main/java/org/wildfly/extension/elytron/ElytronDescriptionConstants.java
@@ -431,6 +431,7 @@ interface ElytronDescriptionConstants {
     String REFERRAL_MODE = "referral-mode";
     String REGISTER_JASPI_FACTORY = "register-jaspi-factory";
     String REGEX_PRINCIPAL_TRANSFORMER = "regex-principal-transformer";
+    String REGEX_ROLE_MAPPER = "regex-role-mapper";
     String REGEX_VALIDATING_PRINCIPAL_TRANSFORMER = "regex-validating-principal-transformer";
     String RELATIVE_TO = "relative-to";
     String REMOVE_ALIAS = "remove-alias";
--- a/elytron/src/main/java/org/wildfly/extension/elytron/RoleMapperDefinitions.java
+++ b/elytron/src/main/java/org/wildfly/extension/elytron/RoleMapperDefinitions.java
@@ -58,6 +58,7 @@ import org.jboss.msc.service.ServiceTarget;
 import org.jboss.msc.value.InjectedValue;
 import org.wildfly.extension.elytron.TrivialService.ValueSupplier;
 import org.wildfly.security.authz.MappedRoleMapper;
+import org.wildfly.security.authz.RegexRoleMapper;
 import org.wildfly.security.authz.RoleMapper;
 import org.wildfly.security.authz.Roles;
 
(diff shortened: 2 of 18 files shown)


elytron/src/main/java/org/wildfly/extension/elytron/RoleMapperDefinitions.java
```diff
@@ -80,6 +85,19 @@
         .setRestartAllServices()
         .build();
 
+    static final SimpleAttributeDefinition REGEX = new SimpleAttributeDefinitionBuilder(ElytronDescriptionConstants.REGEX, ModelType.STRING, false)
```
fjuma: I think you can use `SimpleAttributeDefinitionBuilder(RegexAttributeDefinitions.PATTERN)` here instead and then remove the `set` calls below.

elytron/src/test/java/org/wildfly/extension/elytron/RegexRoleMapperTestCase.java
```diff
@@ -0,0 +1,148 @@
+package org.wildfly.extension.elytron;
```
fjuma: Missing copyright header.

elytron/src/test/java/org/wildfly/extension/elytron/RegexRoleMapperTestCase.java
```diff
@@ -0,0 +1,148 @@
...
+        operation.get(ElytronDescriptionConstants.REPLACEMENT).set("$1");
+        ModelNode response = services.executeOperation(operation);
+        // operation will fail because regex is not valid (starts with asterisk)
+        if (! response.get(OUTCOME).asString().equals(FAILED)) {
+            Assert.fail(response.toJSONString(false));
+        }
+    }
+
+    @Before
+    public void init() throws Exception {
```
fjuma: Is this method actually needed?

elytron/src/main/java/org/wildfly/extension/elytron/ElytronDescriptionConstants.java
```diff
@@ -427,7 +427,9 @@
     String REFERENCE = "reference";
     String REFERRAL_MODE = "referral-mode";
     String REGISTER_JASPI_FACTORY = "register-jaspi-factory";
+    String REGEX = "regex";
```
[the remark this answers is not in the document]
Skyllarr: Fixed. Thanks

elytron/src/main/java/org/wildfly/extension/elytron/RoleMapperDefinitions.java
Skyllarr: Fixed. Thank you

elytron/src/test/java/org/wildfly/extension/elytron/RegexRoleMapperTestCase.java
Skyllarr: Fixed.

elytron/src/test/java/org/wildfly/extension/elytron/RegexRoleMapperTestCase.java
Skyllarr: Fixed.

fjuma:
Thanks, @Skyllarr. This looks good, it just has some conflicts that need to be resolved.

elytron/src/test/java/org/wildfly/extension/elytron/RoleMappersTestCase.java
```diff
@@ -156,4 +167,216 @@ public void testKeepBothMappedRoleMapper() throws Exception {
...
+        context.setAuthenticationName("user5");
+        Assert.assertTrue(context.exists());
+        Assert.assertTrue(context.authorize());
+        context.succeed();
+        SecurityIdentity identity = context.getAuthorizedIdentity();
+        Assert.assertEquals("user5", identity.getPrincipal().getName());
+
+        Roles roles = identity.getRoles();
+        Assert.assertTrue(roles.contains("admin"));
+        Assert.assertTrue(roles.contains("admin"));
```
fabiobrz, line 365: Maybe duplicated line here? From test domain definition I ca see that `user5` should have the following roles, resulting from `AggregateRoleMapper1` processing:
* `1-user-1`: match, translated into `admin`
* `joe`: no match, not kept
* `2-guest-2`: match, translated into `guest`

Hence looks like result should be `{admin, guest}`.
WDYT?

Skyllarr: fixed

PR #4020, opened 2019-11-20
[WFCORE-4596] Do not acquire the write lock reading core-service patching resource

yersan:
The read operation should not acquire the write lock, I haven't found a useful use case where this could be necessary. This patch prevents that HAL could get stuck clicking on Patching when there is already a write operation in progress.

Jira issue: https://issues.jboss.org/browse/WFCORE-4596

### DIFF
--- a/patching/src/main/java/org/jboss/as/patching/management/PatchStreamResourceOperationStepHandler.java
+++ b/patching/src/main/java/org/jboss/as/patching/management/PatchStreamResourceOperationStepHandler.java
@@ -40,6 +40,7 @@ import org.jboss.msc.service.ServiceController;
  * @author Alexey Loubyansky
  */
 abstract class PatchStreamResourceOperationStepHandler implements OperationStepHandler {
+    boolean acquireWriteLock = true;
 
     @Override
     public void execute(OperationContext context, ModelNode operation) throws OperationFailedException {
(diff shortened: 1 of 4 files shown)


patching/src/main/java/org/jboss/as/patching/management/PatchStreamResourceOperationStepHandler.java
```diff
@@ -40,6 +40,7 @@
  * @author Alexey Loubyansky
  */
 abstract class PatchStreamResourceOperationStepHandler implements OperationStepHandler {
+    protected boolean acquireWriteLock = true;
```
bstansberry: Nit: It looks like this can be package protected.

I hate 'protected' due to too many cases where I couldn't improve some piece of code because some external code might have been accessing a protected member. That's not possible here but it's kind of a code standard thing for me not to use it if it's not required.

bstansberry:
Minor point about 'protected' aside, LGTM.

patching/src/main/java/org/jboss/as/patching/management/PatchStreamResourceOperationStepHandler.java
yersan: Nice tip, I've just removed the `protected` modifier.

jmesnil:
@yersan thanks

PR #4025, opened 2019-11-25
[WFCORE-4759] Upgrade to Galleon and WFGP 4.2.1.Final

aloubyansky:
Upgrade to the latest Galleon and WildFly Galleon Plugins that at the moment are at 4.2.1.Final.

https://issues.jboss.org/browse/WFCORE-4759


jmesnil:
@aloubyansky thanks

PR #4030, opened 2019-12-01
[WFCORE-4766] Report error if triggering rollback due to missing services.

darranl:
Addition of an additional error message: -

https://issues.jboss.org/browse/WFCORE-4766

I haven't added loads of detail to the message as in general this is a situation that should not occur but as discovered recently if it does occur we don't log anything to provide a starting point.  With this message we can see the starting point more clearly and at least know where to attach the debugger.

### DIFF
--- a/controller/src/main/java/org/jboss/as/controller/logging/ControllerLogger.java
+++ b/controller/src/main/java/org/jboss/as/controller/logging/ControllerLogger.java
@@ -3562,4 +3562,8 @@ public interface ControllerLogger extends BasicLogger {
     @Message(id = 458, value = "Disallowed HTTP Header name '%s'")
     OperationFailedException disallowedHeaderName(String value);
 
+    @LogMessage(level = ERROR)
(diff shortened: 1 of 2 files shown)


controller/src/main/java/org/jboss/as/controller/logging/ControllerLogger.java
```diff
@@ -3562,4 +3562,8 @@ OperationFailedRuntimeException capabilityAlreadyRegisteredInContext(String capa
     @Message(id = 458, value = "Disallowed HTTP Header name '%s'")
     OperationFailedException disallowedHeaderName(String value);
 
+    @LogMessage(level = ERROR)
+    @Message(id = 459, value = "Triggering roll back due to missing services.")
```
jmesnil: I think it should be `missing *management* services` to be more meaningful

jmesnil:
@darranl thanks

PR #4031, opened 2019-12-02
[WFCORE-4768] WFLYIO001: Worker 'default' has auto-configured to 24 core threads should be IO threads

rmartinc:
Jira issue: https://issues.jboss.org/browse/WFCORE-4768

It's a silly change about the messages for the auto-configured values in the IO subsystem. Current messages seems to be confusing for users. We are using now a more similar similar words to the ones present in the CLI option (`io-threads` => `XX IO threads` and `task-max-threads` => `XX max task threads`). The main problem seems to be the core instead IO. I asked Brad to decide the final message, but if you prefer something different just let me know.


jmesnil:
@rmartinc thanks

PR #4034, opened 2019-12-07
[WFCORE-3948] Revert the CopyOnWriteArrayList approach in favor of Sy…

bstansberry:
…nchronizedList semantics and synchronized iteration where needed

The first attempt at this backed AttachmentList with a CopyOnWriteArrayList but that meant not supporting modifications from the iterators. It turns out that was a breaking change.

Iteration while concurrent writes are going on is not a common case, so go with SynchronizedList semantics and require uses that need thread safe iteration to synchronize on the list.

https://issues.redhat.com/browse/WFCORE-3948


jmesnil:
@bstansberry thanks

PR #4038, opened 2019-12-13
[WFCORE-4774] Avoid circular recursion searching for capability status

yersan:
We could have capabilities with circular requirements, e.g. an attribute that adds a capability reference from a parent capability to a child capability when the child capability already requires its parent.

In those cases, when the server is in restart-required and the capability to check is not the one that was recorded as restart-required, a stack overflow exception is thrown.

Jira issue: https://issues.redhat.com/browse/WFCORE-4774

### DIFF
--- a/controller/src/main/java/org/jboss/as/controller/CapabilityRegistry.java
+++ b/controller/src/main/java/org/jboss/as/controller/CapabilityRegistry.java
@@ -352,14 +352,14 @@ public final class CapabilityRegistry implements ImmutableCapabilityRegistry, Po
             if (size == 0) {
                 result = Collections.emptyMap();
             } else {
-                Set<CapabilityId> examined = new HashSet<>();
+                Set<CapabilityId> visited = new HashSet<>();
                 if (size == 1) {
                     CapabilityId id = ids.iterator().next();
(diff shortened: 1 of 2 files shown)


controller/src/main/java/org/jboss/as/controller/CapabilityRegistry.java
```diff
@@ -385,11 +385,12 @@ private RuntimeStatus getCapabilityStatus(CapabilityId id, Set<CapabilityId> exa
             if (reloadCapabilities.contains(id)) {
                 return RuntimeStatus.RELOAD_REQUIRED;
             }
-            examined.add(id);
         } // else defer reload-required check until after we search requirements for restart-required
 
+        visited.add(id);
```
yersan: This is the only relevant change, move it outside of the previous if to make it count when we have capabilities in the `restartCapabilities`. I'm not sure what was the original intention, looks like it was expected to have it in the `restartCapabilities` array, but that's not always true. I renamed it to `visited` instead of `examined,` since examined could mean already checked its status, which is not correct at this point. At the end of this method, the capability is checked (examined) to see if it is in the `reloadCapabilities.` I found visited more adequate.

jmesnil:
@yersan thanks

PR #4039, opened 2019-12-13
Fix for WFCORE-4776, NPE in EmbeddedServer for version/help arguments

jfdenise:
Issue: https://issues.redhat.com/browse/WFCORE-4776

Avoid NPE by returning in cases where we don't have a server env and status is normal.

### DIFF
--- a/embedded/src/main/java/org/wildfly/core/embedded/EmbeddedStandaloneServerFactory.java
+++ b/embedded/src/main/java/org/wildfly/core/embedded/EmbeddedStandaloneServerFactory.java
@@ -267,7 +267,10 @@ public class EmbeddedStandaloneServerFactory {
 
                     // Determine the ServerEnvironment
                     ServerEnvironment serverEnvironment = Main.determineEnvironment(cmdargs, systemProps, systemEnv, ServerEnvironment.LaunchType.EMBEDDED, startTime).getServerEnvironment();
-
+                    if (serverEnvironment == null) {
+                        // Nothing to do
+                        return;
+                    }
                     bootstrap = Bootstrap.Factory.newInstance();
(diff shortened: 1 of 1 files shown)


embedded/src/main/java/org/wildfly/core/embedded/EmbeddedStandaloneServerFactory.java
```diff
@@ -266,8 +267,12 @@ public void exit(int status) {
                     });
 
                     // Determine the ServerEnvironment
-                    ServerEnvironment serverEnvironment = Main.determineEnvironment(cmdargs, systemProps, systemEnv, ServerEnvironment.LaunchType.EMBEDDED, startTime).getServerEnvironment();
-
+                    ServerEnvironmentWrapper wrapper = Main.determineEnvironment(cmdargs, systemProps, systemEnv, ServerEnvironment.LaunchType.EMBEDDED, startTime);
+                    ServerEnvironment serverEnvironment = wrapper.getServerEnvironment();
+                    if (serverEnvironment == null && ServerEnvironmentWrapper.ServerEnvironmentStatus.NORMAL.equals(wrapper.getServerEnvironmentStatus())) {
+                        // Nothing to do
+                        return;
```
jmesnil: what would happen in the wrapper's environment status is `ERROR`?

jfdenise: @jmesnil , I overlooked the Error case. We would get an NPE in case of ERROR. There is no need to check for the status actually. A null env is enough to return.

jmesnil:
@jfdenise thanks

PR #4040, opened 2019-12-16
[WFCORE-4700] Adding MicroProfile JWT

darranl:
https://issues.redhat.com/browse/WFCORE-4700

The changes to WildFly Core to add support for MicroProfile JWT are relatively small.

This adds a dependency on the elytron-jwt module, however this is not used or installed as a module until WildFly.

Additional phases are added for DUPs which will be added to WildFly.

Finally an internal utility is added to define a virtual security domain to be dynamically associated with a deployment - when working with MicroProfile JWT this means configuration can be 100% bundled with the deployment without relying on central security configuration.

This PR depends on (and will fail CI) without https://github.com/wildfly-security/wildfly-elytron/pull/1358


darranl:
Retest this please

darranl:
I have removed the hold label as the Elytron component upgrade is in, once this passes CI it should be ready to go in.

PR #4056, opened 2020-01-10
[WFLY-11566] Add EJB hibernate validator phase

tadamski:
https://issues.redhat.com/browse/WFLY-11566

The added phase will be used to provide weld with EJB proxy normalizer as in:
https://github.com/tadamski/wildfly/tree/WFLY-11566


jmesnil:
@tadamski thanks

PR #4058, opened 2020-01-17
[WFCORE-4764] Enable web console just before starting the managed server when the domain controller is booting up

yersan:
Basically it adds a service that can be used to check the web console availability and callers can use it to enable it before the DC has transitioned to running.

I temporarily open it as a draft since it will require review HAL with this in.

Jira issue: https://issues.redhat.com/browse/WFCORE-4764

### DIFF
--- a/core-feature-pack/src/main/resources/modules/system/layers/base/org/jboss/as/domain-http-interface/main/module.xml
+++ b/core-feature-pack/src/main/resources/modules/system/layers/base/org/jboss/as/domain-http-interface/main/module.xml
@@ -50,6 +50,8 @@
         <module name="org.jboss.xnio"/>
         <module name="org.wildfly.common" />
         <module name="org.wildfly.security.elytron-web.undertow-server" />
+        <module name="org.jboss.msc"/>
+        <module name="javax.api"/>
     </dependencies>
 
 
--- a/domain-http/interface/src/main/java/org/jboss/as/domain/http/server/ConsoleAvailability.java
+++ b/domain-http/interface/src/main/java/org/jboss/as/domain/http/server/ConsoleAvailability.java
@@ -1,0 +1,54 @@
+/*
+ * Copyright 2020 Red Hat, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *   http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.jboss.as.domain.http.server;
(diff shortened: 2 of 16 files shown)


darranl:
FYI Drafts don't trigger CI if that was your intent, we can instead add a hold label if you want to prevent it from being merged.

What would be the merge order with HAL, can this go into Core without breaking things whilst waiting for HAL in WildFly?  Or can HAL be updated first in WildFly to be followed by this change in Core?

yersan:
> FYI Drafts don't trigger CI if that was your intent, we can instead add a hold label if you want to prevent it from being merged.

I was trying to manually trigger the Pull request aggregator to verify this, but still unable to do it. 

It can go into Core without breaking things in Core or WildFly but I saw one issue in HAL with this change, I wanted to get a review of this but do not merge it until getting a patch for HAL. 

In this scenario, can this be reviewed with a hold label meanwhile HAL is fully tested or it is better just close this an manage it independently of the review and CI jobs? (and create the PR when all is done)

yersan:
retest this please

yersan:
This is ready to tests, related HAL issues have been identified and fixed; @jmesnil Could you remove the 'hold' label to allow this patch to be reviewed? thanks!

yersan:
restest this please

yersan:
retest this please

yersan:
@bstansberry @jmesnil Could any of you take a look and review this one? it is blocking downstream merging, thanks!

controller/src/main/java/org/jboss/as/controller/ConsoleAvailabilityService.java
```diff
@@ -0,0 +1,122 @@
...
+ * <p>
+ *     By default the console is not available, this service listen to changes in the process controller state.
+ *     The web console is available as soon as the process controller transitions to running. This service allows
+ *     to make the console available even before than the process controller is running by using {@link #setAvailable()}
+ *     method.
+ * </p>
+ *
+ * @author @author <a href="mailto:yborgess@redhat.com">Yeray Borges</a>
+ */
+public class ConsoleAvailabilityService implements Service, ConsoleAvailability, PropertyChangeListener {
```
bstansberry: final

bstansberry:
This seems ok other than I dislike adding more public classes to the controller module, particularly for something quite specific.

Can those go in domain-http?  The org.jboss.as.server and org.jboss.as.host-controller modules both depend on it and the server maven module already has an explicit compile time dependency.

yersan:
@bstansberry This is not complete yet, I still have the `ConsoleAvailabity` interface at `org.jboss.as.controller` level, I'm evaluating if we can remove it from there, indeed if we can add it only for host-controller leaving standalone without it.

I'm not sure if the second option is possible and if it makes sense at all. Although currently this class is not necessary/used in standalone, it could be useful to have standalone prepared for its use. I'm not sure yet is if it is better remove it completely from standalone. I'll let you know when you can review this again.

yersan:
@bstansberry When you have a chance, you can review it again, thanks!

I moved all to domian-http. I kept the capability enabled for standalone, since both, standalone and domain, are using common classes to configure the management server.

yersan:
I've just squashed the commits.

jmesnil:
@yersan thanks

PR #4061, opened 2020-01-21
[WFCORE-4803] EJB Client authentication does not work using SASL DIGEST-MD5 and EXTERNAL mechanisms in Legacy security

rmartinc:
Issue: https://issues.redhat.com/browse/WFCORE-4803
Test Issue: https://issues.redhat.com/browse/WFLY-12999
Test PR: https://github.com/wildfly/wildfly/pull/12936

Although there is a test in WFCORE the `SaslLegacyMechanismConfigurationTestCase` in WFLY is also modified to include a DIGEST-MD5 test.

### DIFF
--- a/domain-management/src/main/java/org/jboss/as/domain/management/security/SecurityRealmService.java
+++ b/domain-management/src/main/java/org/jboss/as/domain/management/security/SecurityRealmService.java
@@ -690,7 +690,10 @@ public class SecurityRealmService implements Service<SecurityRealm>, SecurityRea
                     String preferredMechanism;
                     if(authMechanism == AuthMechanism.PLAIN && !registeredServices.containsKey(AuthMechanism.PLAIN) && registeredServices.containsKey(AuthMechanism.DIGEST)) {
                         preferredMechanism = AuthMechanism.DIGEST.name();
+                    } else if (authMechanism != null) {
+                        preferredMechanism = authMechanism.name();
                     } else {
+                        // mechanismName is ANONYMOUS
                         preferredMechanism = mechanismName;
                     }
 
--- a/domain-management/src/test/java/org/jboss/as/domain/management/security/SecurityRealmServiceUtilTestCase.java
+++ b/domain-management/src/test/java/org/jboss/as/domain/management/security/SecurityRealmServiceUtilTestCase.java
@@ -16,16 +16,25 @@ limitations under the License.
 
 package org.jboss.as.domain.management.security;
 
+import java.io.File;
+import java.io.FileOutputStream;
 import java.io.IOException;
+import java.net.URI;
 import java.nio.file.FileVisitResult;
 import java.nio.file.Files;
 import java.nio.file.Path;
 import java.nio.file.SimpleFileVisitor;
 import java.nio.file.attribute.BasicFileAttributes;
+import java.security.AccessController;
+import java.security.Security;
+import java.util.Collections;
 import java.util.HashSet;
+import java.util.Properties;
 import java.util.concurrent.TimeUnit;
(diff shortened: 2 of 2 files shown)


domain-management/src/main/java/org/jboss/as/domain/management/security/SecurityRealmService.java
```diff
@@ -690,6 +690,8 @@ public SaslAuthenticationFactory getSaslAuthenticationFactory(final String[] mec
                     String preferredMechanism;
                     if(authMechanism == AuthMechanism.PLAIN && !registeredServices.containsKey(AuthMechanism.PLAIN) && registeredServices.containsKey(AuthMechanism.DIGEST)) {
                         preferredMechanism = AuthMechanism.DIGEST.name();
+                    } else if (authMechanism != null) {
+                        preferredMechanism = authMechanism.name();
                     } else {
                         preferredMechanism = mechanismName;
```
darranl: Just looking at the code, I think this line should likely have always been the line you are adding i.e. I am not sure if using 'mechanismName` was valid as that is a SASL mechanism name which would not match a realm name which is what this String is being used for. 

Overall I think if authMechanism is null maybe this mechanismName should be skipped and the loop continues.

rmartinc: I thought initially the same, moreover when I saw that the previous lines throw an exception if there is an unknown mechanism ([here](https://github.com/wildfly/wildfly-core/blob/8cfd8e002157641249f8fc986c736dfca3636213/domain-management/src/main/java/org/jboss/as/domain/management/security/SecurityRealmService.java#L650)). But that else is necessary because of the ANONYMOUS. Anonymous is not managed in the `toAuthMechanism` method and it's necessary. When I deleted that line (that can only be anonymous because of the previous check) the [test](https://github.com/wildfly/wildfly/blob/master/testsuite/integration/basic/src/test/java/org/jboss/as/test/integration/ejb/security/SaslLegacyMechanismConfigurationTestCase.java) associated in wildfly failed. So, that line is not completely bad, it was necessary for anonymous, but the conversion is necessary for the rest (digest for example). I can add a comment or similar if you want.
There is a second PR for wildfly with the modified test if you want to check it too: https://github.com/wildfly/wildfly/pull/12936

darranl: Ok that makes sense, in that case maybe we should have an else if for anonymous and then continue the for loop for any remaining mechanisms?  

There are other possible SASL mechanisms that would never match so if somehow they were activated at authentication time we likely do want them to fail quickly.

rmartinc: If there is an unknown mechanism [the method throws an exception before](https://github.com/wildfly/wildfly-core/blob/8cfd8e002157641249f8fc986c736dfca3636213/domain-management/src/main/java/org/jboss/as/domain/management/security/SecurityRealmService.java#L650). I mean, the last else is only for anonymous (the var mechanismName cannot be any other thing at that time). Maybe a comment saying all this would be useful.

darranl: +1 If you are happy the other cases are covered just a quick comment would likely be helpful.

rmartinc: Done! I have added a little comment saying that is anonymous. Just pushed.

jmesnil:
@rmartinc thanks

PR #4064, opened 2020-01-22
[WFCORE-4809] Do not acquire the Dc write lock for composite operations targeting to a single host

yersan:
Currently, when a composite operation that is targetted to a single host is being routed via two-phase operation handler, the write lock is acquired in the Domain Controller, even if we also return the routing as a non-multiphase.

In order to facilitate the tasks of the web console reading resources under this scenario, this patch removes the lock acquisition for these operations.

My understanding:
The write lock is acquired to prevent changes in the topology. Under the scenario for this patch, an operation targeting a single slave host, only a change removing the slave could affect to the results if we are not acquiring the write lock. Removing a slave could be as a cause of:

1. User action, the user has requested a host shutdown or reload.
2. The host process crashed and was removed from the list of available servers.

My argument to allow this change is, in both cases, if there is a read operation in the middle and a host is removed, it could be expected an error.

Notice that write operations should not be affected by this case since a write operation could have acquired the write lock on the slave, and then a manual host reloading done by the user will be blocked in the slave until the write ends. The shutdown doesn't acquire the lock, so it is always executed independently of this patch.

Jira issue: https://issues.redhat.com/browse/WFCORE-4809


yersan:
I'm sorry, this is not ready to test yet, I've just realized the test case is not valid, reviewing it

yersan:
retest this please

jmesnil:
@yersan thanks

PR #4069, opened 2020-01-27
WFCORE-4819 Add OperationDefinition generated by default SimpleResourceDefinition logic never contains parameters

pferraro:
https://issues.redhat.com/browse/WFCORE-4819

PR #4071, opened 2020-01-27
WFCORE-4814 Consolidate common instances of DiscardAttributeChecker that discard an attribute set to the default value

pferraro:
https://issues.redhat.com/browse/WFCORE-4814

Requires https://github.com/wildfly/wildfly-core/pull/4069


jmesnil:
@pferraro thanks

PR #4072, opened 2020-01-29
[WFCORE-4749] Bump domain management version to 11.0.0

yersan:
- Add KnownRelease.EAP73 as known release
- Add missing host-release types to wildlfy-config_11_0.xsd

Jira issue: https://issues.redhat.com/browse/WFCORE-4749

### DIFF
--- a/server/src/main/resources/schema/wildfly-config_11_0.xsd
+++ b/server/src/main/resources/schema/wildfly-config_11_0.xsd
@@ -4978,6 +4978,33 @@
                             </xs:documentation>
                         </xs:annotation>
                     </xs:enumeration>
+                    <xs:enumeration value="WildFly16.0">
+                        <xs:annotation>
+                            <xs:documentation>
+                                <![CDATA[
+                                   A shorthand identifier for kernel management API version 10.0.x
+                                ]]>
+                            </xs:documentation>
+                        </xs:annotation>
+                    </xs:enumeration>
+                    <xs:enumeration value="WildFly17.0">
+                        <xs:annotation>
+                            <xs:documentation>
+                                <![CDATA[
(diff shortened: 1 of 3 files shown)


server/src/main/resources/schema/wildfly-config_11_0.xsd
```diff
@@ -4978,6 +4978,33 @@
                             </xs:documentation>
                         </xs:annotation>
                     </xs:enumeration>
+                    <xs:enumeration value="WildFly16.0">
```
jmesnil: could you also add `EAP7.3` please?

darranl:
@jmesnil it is correct that the test is failing, this schema should not be modified except for cosmetic changes as it is already included in a release, if this is making changes to the allowed values in an enumeration then version 12 of the schema is required.

server/src/main/resources/schema/wildfly-config_11_0.xsd
darranl: It doesn't make sense for EAP 7.3 (which included this schema version) to have an exclude for EAP 7.3.

yersan:
@darranl The schema is for Beta4, is it considered as released?

darranl:
@yersan Yes this is release in EAP 7.3 so must be maintained in a compatible way now.

server/src/main/resources/schema/wildfly-config_11_0.xsd
yersan: EAP 7.3 comes with wildfly-core 10.x, this is for WF19

darranl: EAP 7.3 includes version 11 of the schema.

yersan:
Closing since we are going to use a different approach

PR #4073, opened 2020-01-29
[WFCORE-4822] Add version 12 of the management schema

yersan:
- Bump domain management version to 12.0.0
- Add KnownRelease.EAP73 as known release

Jira issue: https://issues.redhat.com/browse/WFCORE-4822


yersan:
I manually canceled the previous job because I forgot one detail. I'm not sure why the current job list only shows two jobs running, anyway all the jobs are in progress now and they can be checked here:

https://ci.wildfly.org/viewLog.html?buildId=184253&buildTypeId=WFCorePRAggregator

yersan:
Yes @jmesnil , thanks for confirming, I saw the previous core upgrades you included the XSD bump. I'm waiting for such an upgrade to create a PR with the new host exclusions on WildFly.

PR #4076, opened 2020-01-31
[WFCORE-4829] Ensure the pattern-formatter is always marshalled to th…

jamezp:
…e XML configuration correctly. The pattern-formatter element should always be written as should the pattern attribute on the pattern-formatter attribute.

https://issues.redhat.com/browse/WFCORE-4829


jmesnil:
@jamezp thanks

PR #4078, opened 2020-02-03
[WFCORE-4830] Do not acquire the write lock when starting servers to …

yersan:
…allow hosts being registered in parallel when the Domain Controller is starting

Jira issue: https://issues.redhat.com/browse/WFCORE-4830


yersan:
Errors are unrelated, full integration (master) branch requires the schema bump we did on core.

jmesnil:
@yersan thanks

PR #4080, opened 2020-02-04
[WFCORE-4833] Allow TrivialAddHandler and AuthenticationFactories sta…

yersan:
…rt services with a different initial mode for admin-only

This is basically a follow up on #3861 allowing the configuration of a different service initial mode not only for the embedded server process type. This patch allows the configuration of a legacy security domain integration in the standard admin-only mode as well.

Jira issue: https://issues.redhat.com/browse/WFCORE-4833

cc: @darranl

### DIFF
--- a/elytron/src/main/java/org/wildfly/extension/elytron/AuthenticationFactoryDefinitions.java
+++ b/elytron/src/main/java/org/wildfly/extension/elytron/AuthenticationFactoryDefinitions.java
@@ -59,7 +59,6 @@ import org.jboss.as.controller.ObjectListAttributeDefinition;
 import org.jboss.as.controller.ObjectTypeAttributeDefinition;
 import org.jboss.as.controller.OperationContext;
 import org.jboss.as.controller.OperationFailedException;
-import org.jboss.as.controller.ProcessType;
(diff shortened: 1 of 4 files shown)


yersan:
Errors are related, the testsuite is unable to connect when the server is started in admin-only mode. This needs more investigation, @darranl Could you add an 'onhold' label?

yersan:
Hi @darranl , I' more confident with the changes on this issue now, you can remove the 'hold' label, from my side this is ready to test, thanks!

jmesnil:
@darranl could you please review it?

darranl:
I have added an approval so nothing to hold up the merge of this one however I think this general problem could use some kind of follow up.

In a number of places we now have these complex permutations to decide the initial mode required for a service, I wonder (which is why it outside the scope of this issue) if we should look into a more generic solution.

Something along the lines of a capability describing where it's service must be installed or may be installed, maybe combined with the initial mode - we could then handle the cross referencing of the capabilities for the current mode.  I think something like that could also make it easier to have less services even installed in admin mode as the installation would now also be triggered by a check.

PR #4083, opened 2020-02-12
Fix for WFCORE-4844, CommandLineArgumentUsage custom usage

jfdenise:
Fix for: https://issues.redhat.com/browse/WFCORE-4844
Add a new method to CommandLineArgumentUsage for custom usage.


jmesnil:
@jfdenise thanks

PR #4084, opened 2020-02-12
[WFCORE-4841] Fix the capabilities for the logging subsystem. 

jamezp:
[WFCORE-4841] Fix the capabilities for the logging subsystem. Removed the previously unneeded recorder for filters. Moved the tests to the manual mode test suite as a result of WFCORE-4846.
    
 https://issues.redhat.com/browse/WFCORE-4841

The first commit adds the ability to run tests like the `RunsWith(Parameterized.class)`. This can definitely be removed if we don't want this. I can just change the tests.

The second commit adds a `ModuleBuilder` for creating test modules. It's somewhat similar to the `TestModule`, however it didn't quite work right for my use case and a refactor seemed a bit much. Again this can be removed if we don't want it.


jamezp:
The error seems legitimate and I'll have to dig in to find where in the configuration it seems to be wrong.

jamezp:
This requires https://github.com/wildfly/wildfly/pull/12991 to be merged first.

jamezp:
retest this please

jmesnil:
@jamezp thanks

PR #4086, opened 2020-02-18
Fix for WFCORE-4854, Layers for application and management realms

jfdenise:
Issue: https://issues.redhat.com/browse/WFCORE-4854
Provide more flexibility, realms can be excluded from core-security-realms layer.


jmesnil:
@jfdenise thanks

PR #4087, opened 2020-02-26
[WFCORE-4862] Return ServerEnvironmentStatus.ERROR in case of incorrect git server start argument values

yersan:
This is just to avoid NPE traces printing the usage if git configuration values are incorrect.

Jira issue: https://issues.redhat.com/browse/WFCORE-4862


jmesnil:
@yersan thanks

PR #4088, opened 2020-03-02
[WFCORE-4824] Upgrade JBoss Modules from 1.9.2.Final to 1.10.0.Final

ropalka:
https://issues.redhat.com/browse/WFCORE-4824

This update brings in the following fixes and enhancements:

* [MODULES-377] Getting 'IllegalArgumentException: moduleLoader is null' when iterating modules and module.xml contains a permissions markup
* [MODULES-378] Resource in modules.xml that references a directory link ignored when loading a resource bundle with ModuleClassLoader
* [MODULES-393] Make jboss-modules a Java agent and allow other agents to be set as module arguments
* [MODULES-397] Eliminate deprecated -addindex , -modify , -logmodule command line options
* [MODULES-398] Deprecate -jaxpmodule command line option


jmesnil:
@ropalka thanks

PR #4089, opened 2020-03-03
[WFCORE-4621] Create wildflyee.api static module to serve the Jakarta…

yersan:
… EE api dependencies to the external modules

Jira issue: https://issues.redhat.com/browse/WFCORE-4621

- Full integration job tests require this one merged to pass: https://github.com/wildfly/wildfly/pull/13070
- It will follow up with a PR on WildFly full to create a module alias for javaee.api to wildflyee.api


jmesnil:
@yersan The PR for https://github.com/wildfly/wildfly/pull/13070 was done agains wildfly `master` branch which now targets WildFly 20.
That means that this PR for wildfly-core will not make it into WildFly 19. Is that intended?

yersan:
Hello @jmesnil , yes that's intentional

jmesnil:
@yersan thanks

PR #4092, opened 2020-03-05
[WFCORE-4653] Upgrade picketbox to 5.0.3.Final-redhat-00005

bstansberry:
https://issues.redhat.com/browse/WFCORE-4653

I don't think this is urgent. The lib is available in MRRC so I filed this.


jmesnil:
@bstansberry thanks

PR #4095, opened 2020-03-13
[WFCORE-4873]: Upgrade JGit to 5.7.0.202003090808-r

ehsavoie:
* Adding a new layer to be able to 'remove' git support.
     * Adding bouncycastle dependencies and splitting the modules.
     * Removing jcraft jsch and jzlib dependencies.
     * Upgrading jgit and its dependency javaewah.
     * Disabling commit signing.

    Jira: https://issues.redhat.com/browse/WFCORE-4873


ehsavoie:
This requires some changes in full: https://github.com/ehsavoie/wildfly/tree/WFCORE-4873

ehsavoie:
For 'full' integration https://ci.wildfly.org/buildConfiguration/WildFlyCore_PullRequest_WildFlyCoreFullIntegration/192480?buildTab=overview

jmesnil:
@ehsavoie could you rebase and resolve conflicts please?

bstansberry:
My impression is the way this is set up if a user provisions by providing a list of layers, e.g. https://github.com/wildfly/wildfly/blob/master/testsuite/integration/microprofile/pom.xml#L213 then they will no longer have git-history support. They would have to specifically add it, as opposed to having it always be there and then they exclude it if they don't want it. That would be a breaking change.

@jfdenise please correct me if I'm wrong, but this PR adds the new git-history layer to the standalone.xml config's config.xml, but that just makes it a default layer used if the provisioning isn't specifically listing the desired layers.

I don't know if this works but if 'base-server' has an optional dep on git-history, then it gets provisioned in all cases unless the user specifically wants to exclude it.

jfdenise:
@bstansberry , yes, adding it to config.xml allows to provision a slimmed default config with jgit history .
Adding it to base-layer would work (although it creates a circular dependency), I tried it, it reveals a bug in Galleon CLI (not present in galleon core provisioning), So that could be an approach to have it provisioned in all cases (although it brings 10MB).

bstansberry:
@jfdenise When I suggested the base-server layer having an option dep on git-history I didn't mean a circular dep. I meant flipping the relationship so git-history didn't depend on base-server layer, but rather base-server optional depends on it.

Of the 10MB that adds, I assume about 4 is already there. BouncyCastle adds 6. That's added anyway in any server with jaxrs, but still it's hard to swallow!

jfdenise:
@bstansberry , ok, the git-history layer will not be bootable, it doesn't add config element. I think it is fine it is not supposed to be used alone.

darranl:
Have removed the rebase this label as the conflict has been fixed.

I think this needs checking @bstansberry and @jfdenise concluded the discussion regarding layers.

One point I would like to undertake at some point is review how components actually depend on BouncyCastle - ideally for us there would be an SPI allowing us if we want to use the BouncyCastle implementation or swap in something else.

server/src/main/java/org/jboss/as/server/controller/git/GitRepositoryConfiguration.java
```diff
@@ -36,14 +36,16 @@
     private final String branch;
     private final URI authenticationConfig;
     private final Set<String> ignored;
+    private boolean sign = false;
```
bstansberry, line 39: @ehsavoie This should be final.

core-feature-pack/common/pom.xml
```diff
@@ -212,6 +212,21 @@
             <artifactId>slf4j-api</artifactId>
         </dependency>
 
+        <dependency>
```
jmesnil, line 215: Should these new dependencies be listed in `core-feature-pack/common/src/license/core-feature-pack-common-licenses.xml`?

core-feature-pack/common/src/main/resources/modules/system/layers/base/org/bouncycastle/bcpg/main/module.xml
```diff
@@ -1,7 +1,8 @@
 <?xml version="1.0" encoding="UTF-8"?>
+
 <!--
   ~ JBoss, Home of Professional Open Source.
-  ~ Copyright 2011, Red Hat, Inc., and individual contributors
+  ~ Copyright 2010, Red Hat, Inc., and individual contributors
```
jmesnil, line 5: Why did you change the copyright year? If that's for a new file, should it be `2020`?

core-feature-pack/common/src/main/resources/modules/system/layers/base/org/bouncycastle/bcpkix/main/module.xml
```diff
@@ -1,7 +1,8 @@
 <?xml version="1.0" encoding="UTF-8"?>
+
 <!--
   ~ JBoss, Home of Professional Open Source.
-  ~ Copyright 2011, Red Hat, Inc., and individual contributors
+  ~ Copyright 2010, Red Hat, Inc., and individual contributors
```
jmesnil, line 5: same question as above

core-feature-pack/common/src/main/resources/modules/system/layers/base/org/bouncycastle/bcprov/main/module.xml
```diff
@@ -0,0 +1,41 @@
+<?xml version="1.0" encoding="UTF-8"?>
+
+<!--
+  ~ JBoss, Home of Professional Open Source.
+  ~ Copyright 2010, Red Hat, Inc., and individual contributors
```
jmesnil, line 5: ok so I'm pretty sure it must be `2020` :)

pom.xml
```diff
@@ -143,11 +143,11 @@
             For example: <version.org.jboss.hal.release-stream>
          -->
         <version.com.fasterxml.jackson.core.jackson-databind>2.10.4</version.com.fasterxml.jackson.core.jackson-databind>
-        <version.com.googlecode.javaewah>1.1.6</version.com.googlecode.javaewah>
+        <version.com.googlecode.javaewah>1.1.7</version.com.googlecode.javaewah>
         <version.com.google.code.findbugs>3.0.2</version.com.google.code.findbugs>
         <version.com.google.guava>25.0-jre</version.com.google.guava>
         <version.com.io7m.xom>1.2.10</version.com.io7m.xom>
-        <version.com.jcraft.jsch>0.1.54</version.com.jcraft.jsch>
+        <version.com.jcraft.jsch>0.1.55</version.com.jcraft.jsch>
```
jmesnil, line 150: are 2 these libs still used anywhere else in core?

ehsavoie:
I have updated the comments about the dependency changes due to the upgrade.

jmesnil:
@ehsavoie could you please update the JIRA and commit message to explicit mention the version upgrades?
It makes our lives much easier if we don't have to look at the diffs to figure out that javaewah was upgraded to 1.1.7 with that PR... thanks

PR #4097, opened 2020-03-17
Fix for WFCORE-4877, Move testsuite/layers in ts.layers profile

jfdenise:
Issue: https://issues.redhat.com/browse/WFCORE-4877

Layers tests no more executed with -DallTests, only run with -Dts.layers.


jmesnil:
@jfdenise thanks

PR #4100, opened 2020-03-18
Remove patch-config xsds, they are not used here

kabir:
https://issues.redhat.com/browse/WFCORE-4879

https://github.com/jbossas/patch-gen/pull/25


jmesnil:
@kabir thanks

PR #4102, opened 2020-03-19
[WFCORE-4874] Turn pre-computed annotation index loading error into a…

yersan:
… warning and print the related exception

Jira issue: https://issues.redhat.com/browse/WFCORE-4874

### DIFF
--- a/server/src/main/java/org/jboss/as/server/deployment/annotation/ResourceRootIndexer.java
+++ b/server/src/main/java/org/jboss/as/server/deployment/annotation/ResourceRootIndexer.java
@@ -62,7 +62,7 @@ public class ResourceRootIndexer {
                 ServerLogger.DEPLOYMENT_LOGGER.tracef("Found and read index at: %s", indexFile);
                 return;
             } catch (Exception e) {
-                ServerLogger.DEPLOYMENT_LOGGER.cannotLoadAnnotationIndex(indexFile.getPathName());
+                ServerLogger.DEPLOYMENT_LOGGER.cannotLoadAnnotationIndex(indexFile.getPathName(), e.toString());
             }
         }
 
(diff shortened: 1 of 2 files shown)


server/src/main/java/org/jboss/as/server/deployment/annotation/ResourceRootIndexer.java
```diff
@@ -62,7 +62,7 @@ public static void indexResourceRoot(final ResourceRoot resourceRoot) throws Dep
                 ServerLogger.DEPLOYMENT_LOGGER.tracef("Found and read index at: %s", indexFile);
                 return;
             } catch (Exception e) {
-                ServerLogger.DEPLOYMENT_LOGGER.cannotLoadAnnotationIndex(indexFile.getPathName());
+                ServerLogger.DEPLOYMENT_LOGGER.cannotLoadAnnotationIndex(indexFile.getPathName(), e.toString());
```
jmesnil: Should it be `e.getMessage()`?

yersan: The e.toString() is more complete. We could have the exception without any message. The to show the exception message if it exist plus the exception class name.

server/src/main/java/org/jboss/as/server/logging/ServerLogger.java
```diff
@@ -140,9 +140,9 @@
      * Message for when a pre-computed annotation index cannot be loaded
      * @param index name of the annotation index
      */
-    @LogMessage(level = ERROR)
-    @Message(id = 2, value = "Could not read provided index: %s")
-    void cannotLoadAnnotationIndex(String index);
+    @LogMessage(level = WARN)
+    @Message(id = 2, value = "Loading failed for the annotation index \"%s\" with the following failure message: %s")
```
jmesnil: thanks for the explanation. In that case, this string should be more generic like `Loading failed for the annotation index \"%s\" with the following exception: %s`

jmesnil:
@yersan thanks

PR #4104, opened 2020-03-19
[WFCORE-4135] Support for read only server configuration directory

yersan:
This is an RFE. WildFly proposal can be found here: https://github.com/wildfly/wildfly-proposals/pull/304

Jira issue: https://issues.redhat.com/browse/WFCORE-4135


yersan:
Is superseded by https://github.com/wildfly/wildfly-core/pull/4385

PR #4105, opened 2020-03-20
WFCORE-4882 correct minOccurts value for host element in XSD

soul2zimate:
Issue: https://issues.redhat.com/browse/WFCORE-4882 
Broken 'domain.xml' after migration of <host-scoped-roles> leading to 'Boostrap error' in HAL. 

A broken domain configuration
```
           <host-scoped-roles>
                <role name="custom-role-name" base-role="Administrator">
                </role>
            </host-scoped-roles>
```

This broken configuration doesn't bother the old HAL, so user only see the issue from HAL after the migration.

The correct one should be:

```
            <host-scoped-roles>
                <role name="dbaish" base-role="Administrator">
                    <host name="master"/>
                </role>
            </host-scoped-roles>
```
https://github.com/wildfly/wildfly-core/blob/11.0.0.Final/server/src/main/resources/schema/wildfly-config_9_0.xsd#L4567 show minOccurs="1"

~~Attribute `host` list should be required, thus a server boot error can be logged during migration~~

This PR only corrects the minOccruts value in XSD for the host element.

### DIFF
--- a/server/src/main/resources/schema/wildfly-config_13_0.xsd
+++ b/server/src/main/resources/schema/wildfly-config_13_0.xsd
@@ -4610,7 +4610,7 @@
         <xs:complexContent>
             <xs:extension base="base-scoped-roleType">
                 <xs:choice>
-                    <xs:element name="host" type="namedType" minOccurs="1" maxOccurs="unbounded">
+                    <xs:element name="host" type="namedType" minOccurs="0" maxOccurs="unbounded">
                         <xs:annotation>
                             <xs:documentation>
                                 <![CDATA[


jmesnil:
This does not look correct. The code that uses the `HOSTS` attribute expect it to be undefined (as you can see in https://github.com/wildfly/wildfly-core/blob/master/domain-management/src/main/java/org/jboss/as/domain/management/access/HostScopedRoleAdd.java#L69 or https://github.com/wildfly/wildfly-core/blob/master/domain-management/src/main/java/org/jboss/as/domain/management/access/HostScopedRoleRemove.java#L71).

The parsing of the XML elements also stipulates that it is not required: https://github.com/wildfly/wildfly-core/blob/master/domain-management/src/main/java/org/jboss/as/domain/management/parsing/AccessControlXml.java#L180 (`requireChildren` is false).

It seems to me that we should fix the XSD so that `minOccurs` is set to `0`

soul2zimate:
OK, In that case, if the empty list is allowed, HAL is broken (since 3.0.0.Final) with such configuration.

soul2zimate:
@jmesnil I have rebased this to just correct these XSD, and opened https://issues.redhat.com/browse/HAL-1677 to track the console issue when host is empty list (undefined)

soul2zimate:
Yes, sure. I have updated this title and the description in the WFCORE issue.

server/src/main/resources/schema/jboss-as-config_1_5.xsd
```diff
@@ -3353,7 +3353,7 @@
         <xs:complexContent>
             <xs:extension base="base-scoped-roleType">
                 <xs:choice>
-                    <xs:element name="host" type="namedType" minOccurs="1" maxOccurs="unbounded">
```
jmesnil, line 3356: We can not change any schema that has already been released.
That means that the only change we can make is in the new wildfly-config_13_0.xsd that I added last week that will be released with WildFly Core 13 (for WildFly 20)

soul2zimate: pushed a new change to only update that value in new wildfly-config_13_0.xsd

darranl:
If the changes to schemas are to correct bugs where the schemas were incorrect I don't believe it is an issue to correct the older versions, however the changes requested have been made so I will include this PR as it is in a batch I am putting on CI.

PR #4112, opened 2020-03-23
[WFCORE-4860] If no deployments are defined with their own log contex…

jamezp:
…t use the default log context selector. This will avoid the call stack being walked when determining the log context to use.

https://issues.redhat.com/browse/WFCORE-4860


jamezp:
The upstream tests seem to have some issues. I need to sort those out before this can be merged.

jmesnil:
@jamezp thanks

PR #4119, opened 2020-03-27
[WFCORE-4893] Expose the process state from the embedded server and controller

yersan:
Expose the current process state to the client and instead of polling for the process state by using management operations, use the exposed value to return when the process state is not starting.

Jira issue: https://issues.redhat.com/browse/WFCORE-4893


jmesnil:
@yersan thanks

PR #4120, opened 2020-03-27
[WFCORE-4860] Small follow up to original commit for a slight perform…

jamezp:
…ance increase.

https://issues.redhat.com/browse/WFCORE-4860

A small follow up to #4112. @jmesnil if you'd prefer I file a new JIRA let me know. This seemed small enough to not require a new one.


jmesnil:
@jamezp thanks

PR #4122, opened 2020-03-31
Exclude commons-codec dep from httpclient.  WildFly Core doesn't ship…

bstansberry:
… commons-codec, but full WildFly does. I am seeing security scanner warns about bugs in the version transitively depended on by httpclient, which is older than the unaffected version full WildFly uses. I don't want to put commons-codec under WildFly Core's dependency management, as that just makes it a hassle when full WildFly wants to upgrade.  So let's see if just excluding it works.

Note that core has some test fixtures that use httpclient and those uses may need the transitive commons-code dep, in which case this commit will fail CI and we'll need a different solution.

Note: If this passes CI I'll file a JIRA and update the commit message and PR title.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -797,6 +797,10 @@
                         <groupId>commons-logging</groupId>
                         <artifactId>commons-logging</artifactId>
                     </exclusion>
+                    <exclusion>
+                        <groupId>commons-codec</groupId>
+                        <artifactId>commons-codec</artifactId>
+                    </exclusion>
                 </exclusions>
             </dependency>
             <dependency>


bstansberry:
Nope; doesn't work; some tests involving BASIC auth use commons-codec Base64.

darranl:
@bstansberry Feel free to raise an issue for us on the last point, should be fairly easy to move - we have our own Base64 handling in Elytron we can switch to.

bstansberry:
@darranl Would that be a WFCORE or an Elytron issue?

darranl:
@bstansberry from your comment it sounds like the issue is tests in WildFly Core using it for Base64 handling so should just be a WildFly Core issue needed, I can have a look at the test results here and see what is needed.

pom.xml
```diff
@@ -797,6 +797,10 @@
                         <groupId>commons-logging</groupId>
                         <artifactId>commons-logging</artifactId>
                     </exclusion>
+                    <exclusion>
```
darranl, line 800: Could we please add a comment with the exclusion before the final PR?

In the future this will look like someone was just keeping the dependencies tidy so could easily be deleted if new tests need it.

darranl:
Retest this please

darranl:
@bstansberry my PR doesn't contain your changes, if you want I can quickly take your commit and convert to a real PR so we can get it in.

darranl:
This PR has been replaced by https://github.com/wildfly/wildfly-core/pull/4133

PR #4123, opened 2020-04-01
[WFCORE-4748] Allow the MODULE_OPTS to be set and if an -javaagent: is found in the options automatically add jboss-modules as an agent to the VM.

jamezp:
**NOTE:** This is not yet complete, but I wanted to get a CI run to test the different environments.

The first commit just upgrades the wildfly-maven-plugin and changes the property name is it was rather confusing with the `version.org.wildfly.plugins` property.

The second commit exposes the `MODULE_OPTS` environment variable and adds `jboss-modules.jar` as an agent if there is an agent defined in the options. A documentation PR will be required upstream before this can be merged.

https://issues.redhat.com/browse/WFCORE-4748
https://issues.redhat.com/browse/WFCORE-4897

### DIFF
--- a/core-feature-pack/src/main/resources/content/bin/standalone.ps1
+++ b/core-feature-pack/src/main/resources/content/bin/standalone.ps1
@@ -17,7 +17,9 @@ $STANDALONE_CONF_FILE = Get-Env RUN_CONF $STANDALONE_CONF_FILE
 Write-Debug "debug is: $global:DEBUG_MODE"
 Write-Debug "debug port: $global:DEBUG_PORT"
 Write-Debug "sec mgr: $SECMGR"
-
+if ($MODULE_OPTS -contains "-javaagent:") {
+    $JAVA_OPTS += "-javaagent:`\`"$JBOSS_HOME\jboss-modules.jar`\`""
+}
 if ($SECMGR) {
     $MODULE_OPTS +="-secmgr";
 }
--- a/launcher/src/main/java/org/wildfly/core/launcher/StandaloneCommandBuilder.java
+++ b/launcher/src/main/java/org/wildfly/core/launcher/StandaloneCommandBuilder.java
@@ -26,6 +26,7 @@ import static org.wildfly.core.launcher.logger.LauncherMessages.MESSAGES;
 
 import java.nio.file.Path;
 import java.util.ArrayList;
+import java.util.Collection;
 import java.util.LinkedHashMap;
 import java.util.List;
 import java.util.Map;
@@ -39,7 +40,7 @@ import org.wildfly.core.launcher.Arguments.Argument;
  *
  * @author <a href="mailto:jperkins@redhat.com">James R. Perkins</a>
  */
-@SuppressWarnings("unused")
+@SuppressWarnings({"unused", "MagicNumber", "UnusedReturnValue"})
 public class StandaloneCommandBuilder extends AbstractCommandBuilder<StandaloneCommandBuilder> implements CommandBuilder {
 
     // JPDA remote socket debugging
@@ -55,6 +56,8 @@ public class StandaloneCommandBuilder extends AbstractCommandBuilder<StandaloneC
     private String modulesLocklessArg;
     private String modulesMetricsArg;
     private final Map<String, String> securityProperties;
+    private boolean addModuleAgent;
+    private final Collection<String> moduleOpts;
 
     /**
      * Creates a new command builder for a standalone instance.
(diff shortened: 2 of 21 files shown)


core-feature-pack/src/main/resources/content/bin/standalone.ps1
```diff
@@ -17,9 +17,11 @@ $STANDALONE_CONF_FILE = Get-Env RUN_CONF $STANDALONE_CONF_FILE
 Write-Debug "debug is: $global:DEBUG_MODE"
 Write-Debug "debug port: $global:DEBUG_PORT"
 Write-Debug "sec mgr: $SECMGR"
-
+if ($MODULE_OPTS -contains "-javaagent:") {
+    $JAVA_OPTS += "-javaagent:`\`"$JBOSS_HOME\jboss-modules.jar`\`""
+}
 if ($SECMGR) {
-    $MODULE_OPTS +="-secmgr";
+    $MODULE_OPTS +=" -secmgr";
```
jamezp, line 24: This needs to be reverted.

launcher/src/main/java/org/wildfly/core/launcher/StandaloneCommandBuilder.java
```diff
@@ -39,7 +40,7 @@
  *
  * @author <a href="mailto:jperkins@redhat.com">James R. Perkins</a>
  */
-@SuppressWarnings("unused")
+@SuppressWarnings("ALL")
```
jamezp, line 43: This should not have changed to `ALL`.

launcher/src/main/java/org/wildfly/core/launcher/StandaloneCommandBuilder.java
```diff
@@ -102,29 +124,33 @@ public static StandaloneCommandBuilder of(final String wildflyHome) {
...
-                    }
-                    break;
-                case SECURITY_MANAGER_PROP:
-                    setUseSecurityManager(true);
-                    break;
-                default:
-                    javaOpts.add(argument);
-                    break;
+            if (jvmArg.startsWith("-javaagent:")) {
+                agents.add(jvmArg.substring(11));
```
jamezp, line 128: Maybe we shouldn't remove the `-javaagent:` here and add it in the `addAgent()` method instead.

testsuite/scripts/pom.xml
```diff
@@ -80,6 +109,34 @@
...
+                    <jboss-home>${wildfly.home}</jboss-home>
+                </configuration>
+                <executions>
+                    <execution>
+                        <id>configure-standalone</id>
+                        <phase>process-test-classes</phase>
+                        <goals>
+                            <goal>start</goal>
+                            <goal>execute-commands</goal>
+                            <goal>shutdown</goal>
```
jamezp, line 127: Ideally this should be done in an embedded server. However the one issue is the `logging.properties` needs to be updated. The main issue with running `start` first is if the `execute-commands` fails it leaves a server running.

testsuite/scripts/src/test/java/org/wildfly/launcher/test/LauncherTestCase.java
```diff
@@ -0,0 +1,223 @@
...
+
+        Process process = null;
+        try {
+            process = Launcher.of(builder)
+                    .setRedirectErrorStream(true)
+                    .redirectOutput(STDOUT_PATH)
+                    .launch();
+            waitForStandalone(process);
+            final List<JsonObject> lines = readLogFileFromModel("json.log");
+            Assert.assertEquals("Expected 1 line found " + lines.size(), 1, lines.size());
```
jamezp, line 86: This would fail if this test is not run first. We should see if set `append` to `false` and get this working in the CLI commands.

testsuite/scripts/src/test/java/org/wildfly/scripts/test/StandaloneScriptTestCase.java
```diff
@@ -94,6 +94,7 @@ public StandaloneScriptTestCase() {
         final Collection<Object> result = new ArrayList<>(2);
         result.add(Collections.emptyMap());
         result.add(Collections.singletonMap("GC_LOG", "true"));
+        result.add(Collections.singletonMap("MODULE_OPTS", "-javaagent:logging-agent-tests.jar"));
```
jamezp, line 97: When this is set we should check the message is logged. Also maybe we should test this with GC_LOG set too.

PR #4125, opened 2020-04-01
WFCORE-4895, Bootable jar runtime

jfdenise:
Issue: https://issues.redhat.com/browse/WFCORE-4895

* Added new bootable-jar parent module with boot and runtime sub modules.
* Added new artifacts to galleon-feature-pack and legacy feature-pack
* Added API to launch bootable jar in Launcher API
* Evolved test-runner to support bootable jar
* New bootable test profile, -Dts.bootable, rely on wildfly-jar-maven-plugin 2.0.0.Alpha1
* Experimental run of the bootable profile (green): https://ci.wildfly.org/buildConfiguration/WildFlyCore_Experiments/196010
* Upgraded vaut-test-feature-pack build to not need local resolution (useless with latest Galleon and not supported by bootable jar). Removed associated resolve-locals elements in pom.xml.


jfdenise:
@jmesnil , when upgrading full to core release containing this change, you will need to update full LayersTest with: https://github.com/wildfly/wildfly/commit/bebdbd1b349ae4501519cf99d8fc5f6cd6c70283
This will fix: https://ci.wildfly.org/viewLog.html?buildId=196057&buildTypeId=WildFlyCore_PullRequest_GalleonWildFlyCoreFullIntegrationLinuxJdk8

jmesnil:
@jfdenise could we merge  wildfly/wildfly@bebdbd1 beforehand?

jfdenise:
@jmesnil , yes we could.

bootable-jar/runtime/src/main/java/org/wildfly/core/jar/runtime/BootableJar.java
```diff
@@ -0,0 +1,371 @@
...
+    private final ModuleLoader loader;
+
+    private BootableJar(Path jbossHome, Arguments arguments, ModuleLoader loader, long unzipTime) throws Exception {
+        this.jbossHome = jbossHome;
+        this.arguments = arguments;
+        this.loader = loader;
+        startServerArgs.addAll(arguments.getServerArguments());
+        startServerArgs.add(CommandLineConstants.READ_ONLY_SERVER_CONFIG + "=" + STANDALONE_CONFIG);
+
+        configureLogger();
```
jamezp, line 112: This is just a minor note. It might be useful to add a comment here that logging needs to be configured before other components have a chance to initialize a logger.

bootable-jar/runtime/src/main/java/org/wildfly/core/jar/runtime/BootableJar.java
```diff
@@ -0,0 +1,371 @@
...
+                    Files.copy(file, targetFile);
+                }
+            } catch (IOException ex) {
+                throw new RuntimeException(ex);
+            }
+        });
+    }
+
+    private void configureLogger() throws IOException {
+        System.setProperty(LOG_MANAGER_PROP, LOG_MANAGER_CLASS);
```
jamezp, line 202: If we do this we mandate that users must use the JBoss Log Manager. Maybe this is okay, but maybe it's something we should look at.

bootable-jar/runtime/src/main/java/org/wildfly/core/jar/runtime/BootableJar.java
```diff
@@ -0,0 +1,371 @@
...
+        }
+        try {
+            Files.walkFileTree(root, new SimpleFileVisitor<Path>() {
+                @Override
+                public FileVisitResult visitFile(Path file, BasicFileAttributes attrs)
+                        throws IOException {
+                    try {
+                        Files.delete(file);
+                    } catch (IOException ex) {
+                    }
```
jamezp, line 273: Is there a reason we're swallowing the errors here?

bootable-jar/runtime/src/main/java/org/wildfly/core/jar/runtime/BootableJar.java
```diff
@@ -0,0 +1,371 @@
...
+                @Override
+                public FileVisitResult postVisitDirectory(Path dir, IOException e)
+                        throws IOException {
+                    if (e != null) {
+                        // directory iteration failed
+                        throw e;
+                    }
+                    try {
+                        Files.delete(dir);
+                    } catch (IOException ex) {
```
jamezp, line 286: Is there a reason we're swallowing the errors here?

bootable-jar/runtime/src/main/java/org/wildfly/core/jar/runtime/BootableJar.java
```diff
@@ -0,0 +1,371 @@
...
+                    }
+                    return FileVisitResult.CONTINUE;
+                }
+
+                @Override
+                public FileVisitResult postVisitDirectory(Path dir, IOException e)
+                        throws IOException {
+                    if (e != null) {
+                        // directory iteration failed
+                        throw e;
```
jamezp, line 282: This seems odd as we swallow the error below at line 291.

bootable-jar/runtime/src/main/java/org/wildfly/core/jar/runtime/BootableJar.java
```diff
@@ -0,0 +1,371 @@
...
+        }
+        BootableJar bootableJar = new BootableJar(jbossHome, arguments, moduleLoader, unzipTime);
+        bootableJar.run();
+    }
+
+    static void setTccl(final ClassLoader cl) {
+        if (System.getSecurityManager() == null) {
+            Thread.currentThread().setContextClassLoader(cl);
+        } else {
+            AccessController.doPrivileged(new PrivilegedAction<Object>() {
```
jamezp, line 362: Is there a reason we're doing this privileged? I ask because if we're running this under a security manager the other places where we set system properties are gonna be an issue as well.

core-feature-pack/pom.xml
```diff
@@ -786,6 +786,16 @@
...
+        
+        <dependency>
+            <groupId>org.wildfly.core</groupId>
+            <artifactId>wildfly-jar-runtime</artifactId>
+        </dependency>
+        
+        <dependency>
+            <groupId>org.wildfly.core</groupId>
+            <artifactId>wildfly-jar-boot</artifactId>
+        </dependency>
```
jamezp, line 798: We need to add these to the license `core-feature-pack-licenses.xml` with the appropriate license.

bootable-jar/runtime/src/main/java/org/wildfly/core/jar/runtime/BootableJar.java
jfdenise: Ok, added a comment.

bootable-jar/runtime/src/main/java/org/wildfly/core/jar/runtime/BootableJar.java
jfdenise: That is un-needed, I removed the block.

core-feature-pack/pom.xml
jfdenise: Thank-you, added.

bootable-jar/runtime/src/main/java/org/wildfly/core/jar/runtime/BootableJar.java
jfdenise: @jamezp , I was under the assumption that the jboss logmanager was the only usable implementation with the Logging subsystem. You are suggesting that user could set the system property and not rely on the logging subsystem? I could check if the system property exists and do nothing if that is the case.

bootable-jar/runtime/src/main/java/org/wildfly/core/jar/runtime/BootableJar.java
jfdenise: We want to clean as much as we can even if a failure occurs (eg: Windows and opened jar deletion). I should add a log.

bootable-jar/runtime/src/main/java/org/wildfly/core/jar/runtime/BootableJar.java
jfdenise: Same, continue to delete although we have errors. I will add a log.

bootable-jar/runtime/src/main/java/org/wildfly/core/jar/runtime/BootableJar.java
jfdenise: Yep, need to remove it.

bootable-jar/runtime/src/main/java/org/wildfly/core/jar/runtime/BootableJar.java
jamezp: I'm redacting this concern. The logging subsystem does require the JBoss Log Manager, my thought was if the subsystem was removed maybe the user would want to use a different log manager. However the servers main entry point does do some log manager initialization as well so removing a requirement like that is out of scope for this. Sorry about the confusion there.

core-feature-pack/src/main/resources/modules/system/layers/base/org/wildfly/bootable-jar/main/module.xml
```diff
@@ -0,0 +1,53 @@
...
+        <artifact name="${org.wildfly.core:wildfly-jar-runtime}"/>
+    </resources>
+
+    <exports>
+        <exclude path="org/wildfly/core/jar/runtime/_private"/>
+    </exports>
+
+    <dependencies>
+        <module name="java.xml"/>
+        <module name="java.desktop"/>
```
yersan, line 40: Why do we need java.desktop? I cannot find the reason why this dependency is required

bootable-jar/runtime/src/main/java/org/wildfly/core/jar/runtime/Server.java
```diff
@@ -0,0 +1,297 @@
...
+        this.systemEnv = systemEnv;
+        this.moduleLoader = moduleLoader;
+        this.shutdownHandler = shutdownHandler;
+
+        processStateListener = new PropertyChangeListener() {
+            @Override
+            public void propertyChange(PropertyChangeEvent evt) {
+                if ("currentState".equals(evt.getPropertyName())) {
+                    ControlledProcessState.State newState = (ControlledProcessState.State) evt.getNewValue();
+                    establishModelControllerClient(newState, false);
```
yersan, line 94: The boolean argument should be yes instead of false. Doing so the internal variable that stores the currentProcessState gets updated upon an nitification.

core-feature-pack/src/main/resources/modules/system/layers/base/org/wildfly/bootable-jar/main/module.xml
```diff
@@ -0,0 +1,53 @@
...
+        <module name="org.wildfly.common"/>
```
yersan, line 41: Why do we need org.wildfly.common? I cannot find the reason why this dependency is required

core-feature-pack/src/main/resources/modules/system/layers/base/org/wildfly/bootable-jar/main/module.xml
jfdenise: We need it for dependencies on java.beans.PropertyChangeListener

core-feature-pack/src/main/resources/modules/system/layers/base/org/wildfly/bootable-jar/main/module.xml
jfdenise: Good catch, was a left over from the time we were using Embedded. I removed it.

bootable-jar/runtime/src/main/java/org/wildfly/core/jar/runtime/BootableJar.java
```diff
@@ -0,0 +1,358 @@
...
+            // Give max 10 seconds for the server to stop before to delete jbossHome.
+            ModelNode mn = new ModelNode();
+            mn.get(ADDRESS);
+            mn.get(OP).set(READ_ATTRIBUTE_OPERATION);
+            mn.get(NAME).set(SERVER_STATE);
+            for (int i = 0; i < 10; i++) {
+                try {
+                    ModelControllerClient client = server.getModelControllerClient();
+                    if (client != null) {
+                        ModelNode ret = client.execute(mn);
```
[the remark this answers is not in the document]
jfdenise: Good point, I removed DMR request and exposed the process state.

bootable-jar/runtime/src/main/java/org/wildfly/core/jar/runtime/Server.java
jfdenise: Ok.

jfdenise:
@jmesnil , I have opened PR for the failing test in full: https://github.com/wildfly/wildfly/pull/13240

core-feature-pack/src/main/resources/modules/system/layers/base/org/wildfly/bootable-jar/main/module.xml
yersan: ok, thanks!

bootable-jar/runtime/src/main/java/org/wildfly/core/jar/runtime/Server.java
yersan: Just for the record, the requested changes were rejected due to https://issues.redhat.com/browse/WFCORE-4922

bootable-jar/runtime/src/main/java/org/wildfly/core/jar/runtime/BootableJar.java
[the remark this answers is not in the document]
yersan: Just for the record, the requested changes were rejected due to https://issues.redhat.com/browse/WFCORE-4922

darranl:
@jfdenise Looks like we are getting closer but there is a merge conflict in the pom.

Also the test failure on Windows JDK 11 is not something I have seen before, was just about to kick off that job again but as you need to push again that will trigger CI but just in case you want to check if that one is relevent.

jmesnil:
@jfdenise thanks

PR #4128, opened 2020-04-02
[WFCORE-4904] Remove setting Metaspace

moulalis:
Issue JIRA : https://issues.redhat.com/browse/WFCORE-4904

### DIFF
--- a/core-feature-pack/src/main/resources/content/bin/standalone.conf
+++ b/core-feature-pack/src/main/resources/content/bin/standalone.conf
@@ -50,7 +50,7 @@ fi
 # Specify options to pass to the Java VM.
 #
 if [ "x$JAVA_OPTS" = "x" ]; then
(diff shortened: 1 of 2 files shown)


bstansberry:
We need a clear case for removing the -XX:MetaspaceSize=96M setting. The discussion I've seen around this is focused on the -XX:MaxMetaspaceSize setting.

The -XX:MetaspaceSize=96M setting was added because we observed that relying on the default value resulted in a full gc in the middle of server boot as the jvm adjusted to the necessary amount of memory.  That had a small impact on boot speed.

darranl:
Based on Brian's comments I have added the hold label as this sounds like it needs to be a part of a larger investigation.

moulalis:
> Based on Brian's comments I have added the hold label as this sounds like it needs to be a part of a larger investigation.

@darranl Thanks for adding Hold label on it, even I was looking for the same. Investigating more on it.

moulalis:
Working on it.

PR #4136, opened 2020-04-08
[WFCORE-4909] Correct GAV versions in ModelTestControllerVersion

rachmatowicz:
This PR corrects the GAV versions for Wildfly Core in the helper class ModelTestControllerVersion.
For details, see:    https://issues.redhat.com/browse/WFCORE-4909

### DIFF
--- a/model-test/src/main/java/org/jboss/as/model/test/ModelTestControllerVersion.java
+++ b/model-test/src/main/java/org/jboss/as/model/test/ModelTestControllerVersion.java
@@ -42,11 +42,11 @@ public enum ModelTestControllerVersion {
     EAP_6_3_0 ("7.4.0.Final-redhat-19", true, null, "6.3.0"),
     EAP_6_4_0 ("7.5.0.Final-redhat-21", true, "7.5.0", "6.4.0"), //EAP 6.4 is the earliest version we support for transformers
     EAP_6_4_7 ("7.5.7.Final-redhat-3", true, "7.5.0", "6.4.7"), //this one is special as it has model change in micro release
-    EAP_7_0_0 ("7.0.0.GA-redhat-2", true, "10.0.0", "2.1.0.Final", "7.0.0"),
-    EAP_7_1_0 ("7.1.0.GA-redhat-11", true, "11.0.0", "3.0.10.Final", "7.1.0"),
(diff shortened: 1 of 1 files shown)


bstansberry:
Hi @rachmatowicz -- PRs for WildFly Core should generally be against master

rachmatowicz:
Hi Brian, i'll create a PR for master. The reason I chose 11.0.x was that I need this for wildfly master which uses 11.0.1-SNAPSHOT at present.

bstansberry:
@rachmatowicz @jmesnil  I'm not sure I agree with these changes.

We are testing the WildFly management APIs. The fact we use EAP release names to identify management API versions is a convenience for us humans who won't recall how they map to various releases. If EAP has introduced changes that invalidate the test against community, then it has a bug as EAP does not control these API, WildFly does.

I'm not enthused about introducing the need to pull in -redhat artifacts into WildFly code.  We unfortunately have a few cases where we need to do that because some community projects have been archived and we have no other source for CVE fixes than consuming -redhat artifacts. But the ultimate solution there is to stop using those projects.

bstansberry:
I guess we already have -redhat in these classes so I suppose it's ok. I'm not thrilled about this spreading around though.

rachmatowicz:
@bstansberry If I understand correctly what you are saying, rather than make the current changes, wouldn't a better solution be to reference the maven artifact for the Wildfly version corresponding to the EAP version in question, rather than the productized versions of Wildfly?
For example, this:
EAP_7_1_0 ("11.0.0.Final", true, "11.0.0", "3.0.10.Final", "7.1.0")
instead of this:
EAP_7_1_0 ("7.1.0.GA-redhat-11", true, "11.0.0", "3.0.10.Final", "7.1.0")

However, i'm scratching my head, as I just looked at the Wildfly 11.0.0.Final sources and it uses 3.0.8.Final of Wildfly Core and not 3.0.10.Final as used by the productized version. On the other hand, for EAP 7.2, Wildfly 14.0.0.Final does use Wildfly Core 6.0.11.Final.

bstansberry:
@rachmatowicz Sorry; never mind my comments. It seems we have -redhat versions all over the place in these tests so there's nothing wrong with what you're doing. We went down this path long ago -- and I was probably fine with it at the time ;-) -- so there's no reason not to be consistent.

jmesnil:
@rachmatowicz thanks

PR #4141, opened 2020-04-09
Add GitHub Action for dependency checker

kabir:
@jmesnil @luck3y @bstansberry  Here is a PR to introduce the dependency tree checker using GitHub Actions.

The action lives in https://github.com/wildfly/dep-tree-diff-action and calls through to the same https://github.com/wildfly/dep-tree-diff as we use on TeamCity.

I'll be interested how the overall status of the job works out now that we have a mix of GH Actions and TeamCity jobs reporting back.


kabir:
Hmm, seems I need to turn on something. I will play over the weekend

kabir:
Superseded by https://github.com/wildfly/wildfly-core/pull/4143

PR #4142, opened 2020-04-09
[WFCORE-4915] Tighten up the dependency tree of subsystem-test

bstansberry:
https://issues.redhat.com/browse/WFCORE-4915


bstansberry:
It wouldn't shock me if this introduced problems in the tests against full due to modules there inappropriately getting transitive deps via their subsystem-test dep. But probably not; the core modules themselves didn't have that kind of problem.

jmesnil:
@bstansberry thanks

PR #4144, opened 2020-04-10
Dep tree analyser gh actions

kabir:
Supersedes #4141

PR to introduce the dependency tree checker using GitHub Actions.

The action lives in https://github.com/wildfly/dep-tree-diff-action and calls through to the same https://github.com/wildfly/dep-tree-diff as we use on TeamCity.

I'll be interested how the overall status of the job works out now that we have a mix of GH Actions and TeamCity jobs reporting back.


jmesnil:
@kabir thanks!

PR #4156, opened 2020-04-14
[WFCORE-4897] Upgrade the wildfly-maven-plugin to 2.0.2.Final and ren…

jamezp:
…ame the property as it's too closely named to another plugin version property.

https://issues.redhat.com/browse/WFCORE-4897

This is not really required, but I had done it as part of another commit and felt I may as well submit it. If we don't want to do it we can just close it.


jmesnil:
@jamezp thanks

PR #4158, opened 2020-04-15
[WFLY-4908] WriteAttributeHandler should better validate attribute na…

bstansberry:
…me expressions

https://issues.redhat.com/browse/WFCORE-4908


jmesnil:
@bstansberry thanks

PR #4163, opened 2020-04-15
WFCORE-4924: add plain-text option to security command

michpetrov:
Issue: [WFCORE-4924](https://issues.redhat.com/browse/WFCORE-4924)

### DIFF
--- a/cli/src/main/java/org/jboss/as/cli/impl/aesh/cmd/security/auth/AbstractEnableAuthenticationCommand.java
+++ b/cli/src/main/java/org/jboss/as/cli/impl/aesh/cmd/security/auth/AbstractEnableAuthenticationCommand.java
@@ -15,6 +15,24 @@ limitations under the License.
  */
 package org.jboss.as.cli.impl.aesh.cmd.security.auth;
 
+import static org.jboss.as.cli.impl.aesh.cmd.security.SecurityCommand.OPT_EXPOSED_REALM;
+import static org.jboss.as.cli.impl.aesh.cmd.security.SecurityCommand.OPT_FILE_SYSTEM_REALM_NAME;
+import static org.jboss.as.cli.impl.aesh.cmd.security.SecurityCommand.OPT_GROUP_PROPERTIES_FILE;
+import static org.jboss.as.cli.impl.aesh.cmd.security.SecurityCommand.OPT_KEY_STORE_NAME;
+import static org.jboss.as.cli.impl.aesh.cmd.security.SecurityCommand.OPT_KEY_STORE_REALM_NAME;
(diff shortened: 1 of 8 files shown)


cli/src/main/java/org/jboss/as/cli/impl/aesh/cmd/security/auth/AbstractEnableAuthenticationCommand.java
```diff
@@ -53,16 +44,11 @@
...
 import org.jboss.as.cli.impl.aesh.cmd.security.model.ExistingPropertiesRealmConfiguration;
-import static org.jboss.as.cli.impl.aesh.cmd.security.SecurityCommand.OPT_FILE_SYSTEM_REALM_NAME;
-import static org.jboss.as.cli.impl.aesh.cmd.security.SecurityCommand.OPT_PROPERTIES_REALM_NAME;
-import static org.jboss.as.cli.impl.aesh.cmd.security.SecurityCommand.OPT_EXPOSED_REALM;
-import static org.jboss.as.cli.impl.aesh.cmd.security.SecurityCommand.OPT_KEY_STORE_NAME;
-import static org.jboss.as.cli.impl.aesh.cmd.security.SecurityCommand.OPT_KEY_STORE_REALM_NAME;
-import static org.jboss.as.cli.impl.aesh.cmd.security.SecurityCommand.OPT_ROLES;
 import org.jboss.as.cli.impl.aesh.cmd.security.model.ExistingKeyStoreConfiguration;
 
+import static org.jboss.as.cli.impl.aesh.cmd.security.SecurityCommand.*;
```
darranl, line 50: Please move these back to the top and expand back so they are fully qualified.

michpetrov: Is this ok? The imports in the original are kinda all over the place.

darranl: They do look a bit disorganised to begin with but all the static imports should be first in the import list so I think they should still move up.

michpetrov: done

jfdenise:
@darranl , looks good to me. yes there is a community doc but is generic enough tonot have to update it: https://github.com/wildfly/wildfly/blob/master/docs/src/main/asciidoc/_admin-guide/CLI_Recipes.adoc#security-commands
We are suggesting to use the help to discover the options one can use.

darranl:
Thanks you @jfdenise will see if I can include this in a second round of merges later today.

PR #4167, opened 2020-04-17
[WFCORE-4922] Fix possible hang starting the embedded host and server

yersan:
Rollbacks the changes done in https://github.com/wildfly/wildfly-core/pull/4119 for the embedded server and fix the hang reading the process state for the embedded host controller

Jira issue: https://issues.redhat.com/browse/WFCORE-4922

### DIFF
--- a/embedded/src/main/java/org/wildfly/core/embedded/EmbeddedStandaloneServerFactory.java
+++ b/embedded/src/main/java/org/wildfly/core/embedded/EmbeddedStandaloneServerFactory.java
@@ -215,7 +215,7 @@ public class EmbeddedStandaloneServerFactory {
         private final ModuleLoader moduleLoader;
         private final ClassLoader embeddedModuleCL;
         private ServiceContainer serviceContainer;
-        private volatile ControlledProcessState.State currentProcessState;
+        private ControlledProcessState.State currentProcessState;
         private ModelControllerClient modelControllerClient;
         private ExecutorService executorService;
         private ProcessStateNotifier processStateNotifier;
@@ -232,7 +232,7 @@ public class EmbeddedStandaloneServerFactory {
                 public void propertyChange(PropertyChangeEvent evt) {
                     if ("currentState".equals(evt.getPropertyName())) {
                         ControlledProcessState.State newState = (ControlledProcessState.State) evt.getNewValue();
-                        establishModelControllerClient(newState, true);
+                        establishModelControllerClient(newState, false);
                     }
                 }
             };
@@ -341,10 +341,9 @@ public class EmbeddedStandaloneServerFactory {
 
         @Override
         public String getProcessState() {
-            if (currentProcessState == null) {
-                return null;
-            }
-            return currentProcessState.toString();
+            // The access to the internal process state is unsupported for the standalone embedded server, you can use a
+            // management operation to query the current server process state instead.
+            throw new UnsupportedOperationException();
         }
 
         private void exit() {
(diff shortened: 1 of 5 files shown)


yersan:
The latest hang looks different, have you seen it before?

bstansberry:
@yersan tl;dr; don't worry about it

I don't recall seeing that exact variant of hitting the overall job timeout before (where maven was done but it ran out of time in the final cleanup work just after that) but Windows jobs failing because the overall job just took too long is common.

cli/src/main/java/org/jboss/as/cli/embedded/EmbedHostControllerHandler.java
```diff
@@ -276,7 +276,7 @@ protected void doHandle(CommandContext ctx) throws CommandLineException {
                 do {
                     status = hostController.getProcessState();
 
-                    if (status == null || "starting".equals(status)) {
```
bstansberry, line 279: I believe status == null is a likely scenario while the host is starting. HostControllerImpl.currentProcessState starts as null and won't change to something else until it receives a notification of a change via ProcessStateNotifier.

embedded/src/main/java/org/wildfly/core/embedded/EmbeddedStandaloneServerFactory.java
```diff
@@ -341,10 +341,7 @@ public void stop() {
 
         @Override
         public String getProcessState() {
-            if (currentProcessState == null) {
-                return null;
-            }
-            return currentProcessState.toString();
+            throw new UnsupportedOperationException("The access to the internal process state is unsupported for this standalone embedded server");
```
bstansberry, line 344: If you put a message here it really should be internationalized.

A plain no-message UOE with a code comment should be enough. I think it would be a developer messing around who would hit this, not a user. They can analyze the stack trace find the code and read the comment.

cli/src/main/java/org/jboss/as/cli/embedded/EmbedHostControllerHandler.java
yersan: Yes, I change it because I think we are never going to get a `null` as returned value here if https://github.com/wildfly/wildfly-core/blob/c1c217a784283e1f9413bc08da5bce16dd5061f9/cli/src/main/java/org/jboss/as/cli/embedded/EmbedHostControllerHandler.java#L266 is executed without throwing any exceptions.

The thread which updates the `HostControllerImpl.currentProcessState` via ProcessStateNotifier listener propertyChange() is the "Controller Boot Thread", which is the same thread that calls the done() for the future container. That means https://github.com/wildfly/wildfly-core/blob/c1c217a784283e1f9413bc08da5bce16dd5061f9/embedded/src/main/java/org/wildfly/core/embedded/EmbeddedHostControllerFactory.java#L279 is always returned with the HostControllerImpl.currentProcessState as "starting" and never with null.

This is pretty hidden; it is better to keep the null check and avoid these subtleties to make the code easier to follow. I changed it. Thanks!

embedded/src/main/java/org/wildfly/core/embedded/EmbeddedStandaloneServerFactory.java
yersan: Understood, thanks!, yes that was the intention, I throw the UOE for a developer point of view only. Fixed.

cli/src/main/java/org/jboss/as/cli/embedded/EmbedHostControllerHandler.java
bstansberry: Thanks; that makes sense. When I wrote that comment I thought it was odd no test hit the race I was concerned about. Now I know why. :)

jmesnil:
@yersan thanks!

PR #4170, opened 2020-04-20
[WFCORE-4935] if server is already in suspend mode, don't register  any listener, just shut it down.

soul2zimate:
Issue: https://issues.redhat.com/browse/WFCORE-4935

If server is already in suspend mode, don't register any listener, just shut it down. Otherwise, a suspend mode started server won't be shutdown before resuming.


jmesnil:
@soul2zimate thanks

PR #4173, opened 2020-04-21
[WFCORE-4932] Reading of identity from security domain causes NPE

Skyllarr:
https://issues.redhat.com/browse/WFCORE-4932


testsuite/elytron/src/test/java/org/wildfly/test/security/common/elytron/SecurityDomainTest.java
```diff
@@ -0,0 +1,45 @@
...
+        cli.sendLine("/subsystem=elytron/security-domain=mySD:remove");
+        cli.sendLine("/subsystem=elytron/filesystem-realm=myFsRealm:remove");
+        cli.sendLine("reload");
+    }
+
+    @Test
+    public void testValidHostContextMapValue() {
+        boolean success = cli.sendLine("/subsystem=elytron/security-domain=mySD:read-identity(name=myIdentity)", true);
+        Assert.assertFalse(success);
+        Assert.assertThat("Wrong error message", cli.readOutput(), containsString("Identity with name [myIdentity] not authorized."));
```
Skyllarr, line 43: @darranl @fjuma After fix for WFCORE-4407 security domain service was not started before attempting to read the identity which resulted in NPE. 

This fix removes NPE but it is still not possible to read the identity and resulting message is "Identity with name [myIdentity] not authorized." This is also invalid response as I should be able to read the identity I just added, right? If so, I will create separate issue for this.

Skyllarr: Just a clarification that the response `Identity with name [myIdentity] not authorized.` is unrelated to this NPE bug. This NPE bug is happening since wf-core 9.0.0.Beta7 and before that the response was also `Identity with name [myIdentity] not authorized.`

Skyllarr: update: when I add `default-permission-mapper` to the security domain I can read the identity. I have updated the PR

testsuite/elytron/src/test/java/org/wildfly/test/security/common/elytron/SecurityDomainTest.java
```diff
@@ -0,0 +1,45 @@
+package org.wildfly.test.security.common.elytron;
```
darranl, line 1: Just missing a copyright header on this one.

darranl:
@Skyllarr other than the missing header are you happy this one is ready to go?

Skyllarr:
@darranl Yes, I added missing header.

PR #4126, opened 2020-04-02
[WFCORE-4901] TCCL not set to application classloader when MBean Noti…

moulalis:
JIRA issue: https://issues.redhat.com/browse/WFCORE-4901

PR #4175, opened 2020-04-21
[WFCORE-4901] TCCL not set to application classloader when MBean Noti…

moulalis:
JIRA issue: https://issues.redhat.com/browse/WFCORE-4901

Old PR : https://github.com/wildfly/wildfly-core/pull/4126

### DIFF
--- a/jmx/src/test/java/org/jboss/as/jmx/PluggableMBeanServerTestCase.java
+++ b/jmx/src/test/java/org/jboss/as/jmx/PluggableMBeanServerTestCase.java
@@ -25,10 +25,13 @@ import javax.management.InstanceNotFoundException;
 import javax.management.IntrospectionException;
 import javax.management.MBeanRegistration;
 import javax.management.MBeanServer;
+import javax.management.Notification;
+import javax.management.NotificationBroadcasterSupport;
+import javax.management.NotificationFilterSupport;
+import javax.management.NotificationListener;
 import javax.management.ObjectName;
 import javax.management.ReflectionException;
 import javax.management.RuntimeOperationsException;
-
 import org.jboss.as.controller.OperationContext;
 import org.jboss.as.controller.OperationFailedException;
 import org.jboss.as.controller.OperationStepHandler;
@@ -46,7 +49,11 @@ import org.jboss.msc.service.ServiceController;
 import org.junit.Assert;
 import org.junit.Before;
 import org.junit.Test;
+import org.wildfly.security.manager.WildFlySecurityManager;
 
+import java.net.URL;
+import java.net.URLClassLoader;
+import java.util.concurrent.atomic.AtomicBoolean;
 /**
  * @author <a href="kabir.khan@jboss.com">Kabir Khan</a>
  */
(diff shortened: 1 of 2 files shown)


moulalis:
Hi @bstansberry ,
This is new PR for https://github.com/wildfly/wildfly-core/pull/4126 which I've closed.
Please review it.
Thanks.

jmx/src/test/java/org/jboss/as/jmx/PluggableMBeanServerTestCase.java
```diff
@@ -105,6 +112,33 @@ public void testReservedDomainMBeanRegistrationFails() throws Exception {
         reservedDomainTest("jboss.as.expr:bean=test-null", NAME);
     }
 
+    @Test
+    public void testTccl() throws Exception {
```
bstansberry, line 116: Minor: can you call this "testNotificationTccl"?  Or something like that.

jmx/src/test/java/org/jboss/as/jmx/PluggableMBeanServerTestCase.java
```diff
@@ -165,6 +199,10 @@ public void preDeregister() throws Exception {
         public void postDeregister() {
         }
 
+        @Override
+        public void proceed() {
+            sendNotification(new Notification("testtccl", 1, 1));
```
bstansberry, line 204: If you change the return type of this method to ClassLoader, then you can add this:

```
return WildFlySecurityManager.getCurrentContextClassLoaderPrivileged();
```

And then the test can validate that it returns newOutTccl. That both serves as a sanity check that newOutTccl was in place before the notification handling, and confirms that it got restored.

moulalis: @bstansberry  I've Renamed it to testNotificationTccl. Initailly I did the same as returntype ClassLoader, But I thought this is not invoking notification was just a test,  so I wrote it inside notification, In that way it actually invoked notification and test the same. Is that okay to have this or still you want me to change?

jmx/src/test/java/org/jboss/as/jmx/PluggableMBeanServerTestCase.java
moulalis: @bstansberry  I've changed it to testNotificationTccl

jmx/src/test/java/org/jboss/as/jmx/PluggableMBeanServerTestCase.java
moulalis: @bstansberry  Initailly I did the same as returntype ClassLoader, But I thought this is not invoking notification was just a test, so I wrote it inside notification, In that way it actually invoked notification and test the same. Is that okay to have this or still you want me to change?

bstansberry: What your NotificationListener is doing is fine -- it checks that the expected TCCL is in place when it is invoked. I was suggesting an additional check that the expected TCCL (newOutTccl) is in place after the notification call is done.

But this is ok as is.

jmesnil:
@moulalis thanks

PR #4132, opened 2020-04-06
[WFCORE-4905] Provide common capability for Remoting connectors.

rachmatowicz:
This PR allows Remoting subsystem connector elements to be referenced by a common capability reference.

For details, see: https://issues.redhat.com/browse/WFCORE-4905

PR #4180, opened 2020-04-23
Revert "[WFCORE-4905] Provide common capability for Remoting connectors."

darranl:
Reverts wildfly/wildfly-core#4132

Sorry did not see this was against 11.0.x - the original PR should have been against master.


pferraro:
My bad.  I did not notice either when I approved it.

darranl:
I should have checked the top more closely, looked more at the dep-tree stuff - the good news is I know it does pass on master as I pushed it through an aggregate CI there.

PR #4182, opened 2020-04-24
WFCORE-4041: supply default values for nested attributes

michpetrov:
Issue: [WFCORE-4041](https://issues.redhat.com/browse/WFCORE-4041)


michpetrov:
I don't think the failing test is my fault, the PR deals with `read-attribute` and the test fails because of `write-attribute`.

PR #4183, opened 2020-04-27
WFCORE-2252: remove default value

michpetrov:

Blocks #3773

The `empty` attribute is set as required but also has a default value, this seems like an oversight but if it's supposed to be required we can remove the default value instead.


server/src/main/java/org/jboss/as/server/controller/resources/DeploymentAttributes.java
```diff
@@ -159,6 +159,7 @@ public void marshallAsAttribute(AttributeDefinition attribute, ModelNode resourc
                     ModelDescriptionConstants.HASH, ModelDescriptionConstants.INPUT_STREAM_INDEX,
                     ModelDescriptionConstants.BYTES, ModelDescriptionConstants.URL,
                     ModelDescriptionConstants.PATH, ModelDescriptionConstants.RELATIVE_TO)
+            .setRequired(false)
```
bstansberry, line 162: This is required, as it is one of a set of alternatives one of which must be configured. So it's the default that's 'wrong'.

Changing that is a kernel API change so it needs to be done in a release that has a kernel API version bump.

michpetrov:
@bstansberry I've changed both PRs to remove the default instead.

PR #4184, opened 2020-04-27
WFCORE-2252: remove default value

michpetrov:

Blocks #3773

The `bind-address` attribute is set as required but also has a default value, I set it as optional but if it's supposed to be required we can remove the default value instead.


jmesnil:
@michpetrov why did you open 2 PRs (#4183 & 4184) for the same issue?

io/subsystem/src/main/java/org/wildfly/extension/io/OutboundBindAddressResourceDefinition.java
```diff
@@ -49,8 +49,7 @@
         .setRestartAllServices()
         .build();
 
-    static final SimpleAttributeDefinition BIND_ADDRESS = new SimpleAttributeDefinitionBuilder("bind-address", ModelType.STRING)
-        .setRequired(true)
+    static final SimpleAttributeDefinition BIND_ADDRESS = new SimpleAttributeDefinitionBuilder("bind-address", ModelType.STRING, true)
```
bstansberry, line 52: I think the problem here is the default value.  A default value of numeric zero makes no sense for a string attribute whose value is meant to either be a resolvable hostname or some IP address notation.

michpetrov:
@jmesnil I was to told to split the original PR so the changes can be decided on a per-subsystem basis.

PR #4187, opened 2020-04-29
[WFCORE-4950] Regression: Legacy Ldap Realm securing EJB with JDK8 not working

rmartinc:
Issue: https://issues.redhat.com/browse/WFCORE-4950
Test PR in wildfly: https://github.com/wildfly/wildfly/pull/13249

Fix for WFCORE-4950. The final problem is a ClassNotFoundException of the class "com.sun.jndi.ldap.LdapCtxFactory" from module "org.wildfly.extension.io". So I added "sun.jdk" to that module. If you see something better to do just let me know. 

Tested with ibm jdk -8 too (just in case).

Tests included in issue WFLY-13358, I'll attach the PR for wildfly asap.

### DIFF
--- a/domain-management/src/main/java/org/jboss/as/domain/management/connections/ldap/LdapConnectionManagerService.java
+++ b/domain-management/src/main/java/org/jboss/as/domain/management/connections/ldap/LdapConnectionManagerService.java
@@ -251,11 +251,10 @@ public class LdapConnectionManagerService implements Service<LdapConnectionManag
     }
 
     private DirContext getConnection(final Hashtable<String, String> properties, final SSLContext sslContext) throws NamingException {
-        ClassLoader old = WildFlySecurityManager.getCurrentContextClassLoaderPrivileged();
+        ClassLoader old = WildFlySecurityManager.setCurrentContextClassLoaderPrivileged(LdapConnectionManagerService.class);
         try {
             if (sslContext != null) {
                 ThreadLocalSSLSocketFactory.setSSLSocketFactory(sslContext.getSocketFactory());
-                WildFlySecurityManager.setCurrentContextClassLoaderPrivileged(ThreadLocalSSLSocketFactory.class);
                 properties.put("java.naming.ldap.factory.socket", ThreadLocalSSLSocketFactory.class.getName());
             }
             if (SECURITY_LOGGER.isTraceEnabled()) {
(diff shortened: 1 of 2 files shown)


rmartinc:
I changed my mind. I think that the proper thing to add is "java.naming". See my comments in [WFCORE-4950](https://issues.redhat.com/browse/WFCORE-4950). Let me know if I'm missing something.

rmartinc:
@darranl Can you review this or assign someone suited to do it? (It seems we need to move this forward). Mainly I'm managing two ideas:

1. Adding `java.naming` dependency to the module as it is used in case of jdk-11. The PR is adding this change.
2. Change the TCCL always as it is done now if a SSLContext is passed.

See the JIRA for more details. If you prefer the second idea, or a different one, just let me know. The PR for testing is https://github.com/wildfly/wildfly/pull/13249.

Thanks!

darranl:
@rmartinc are you happy with the changes you have made so far adequately address the issue?

rmartinc:
@darranl The change fixes the issue for sure, indeed it does exactly the same that jdk-11 is doing to make it work. But I have to say that I'm not totally happy with it. I doubted a lot between the two options. Mainly because the `java.naming` dependency was removed previously.

I chose this option because in jdk-11 the `java.naming` package is included, so better do the same in jdk-8. Doing something different in jdk-8 makes no sense to me.

bstansberry:
@rmartinc @darranl 

What's the argument *against* always setting the TCCL?

The jdk.security.auth module having visibility to java.naming with JDK 11 is not the same as org.wildfly.extension.io having visibility.  The o.w.e.io ClassLoader is the default classloader in effect when various code does deserialization. The fewer classes there are that are visible to it when it does that the better.

If o.w.e.io didn't need jdk.security.auth because some other module's CL was the TCCL when access was needed, that would be good too.

darranl:
I suspect there is no problem always setting the TCCL, we are going back a bit now but I believe it was set so solve the specific problem of getting the SSL Configuration in the first place.

rmartinc:
I checked both options and this specific issue was resolved by both of them. I needed to implement one and I just chose the wrong option. ;-)

OK then, I'll change the code to use the TCCL solution, run the tests again and update the PR. Thanks a lot guys!

rmartinc:
I have implemented the TCCL option. I'm going to change the test in wildfly too in order to use ldaps instead of plain ldap. This way the test is the most complicated situation. Regards!

rmartinc:
@darranl How do you see the PR now?

PR #4192, opened 2020-05-05
WFCORE-4957 Reduce heap footprint of single attribute write-attribute handlers

pferraro:
https://issues.redhat.com/browse/WFCORE-4957


jmesnil:
@pferraro thanks

PR #4197, opened 2020-05-12
[WFCORE-4895] Fix bootable jar maven module versions

yersan:
Versions got state after merging. This throws an error building wildfly-core.

Jira issue: https://issues.redhat.com/browse/WFCORE-4895

FYI @jmesnil


jmesnil:
@yersan thanks Yeray. I'm merging it now (without waiting for CI) so that wildfly-core can be properly built.

PR #4199, opened 2020-05-14
WFCORE-4967 fix: RequestController drops queued tasks randomly during server start.

khroolick:
solves: https://issues.redhat.com/browse/WFCORE-4967

findForcedTask method implementation could be improved but synchronization is fine from my point of view when server is paused anyway.

### DIFF
--- a/request-controller/src/main/java/org/wildfly/extension/requestcontroller/RequestController.java
+++ b/request-controller/src/main/java/org/wildfly/extension/requestcontroller/RequestController.java
@@ -396,7 +396,7 @@ public class RequestController implements Service<RequestController>, ServerActi
         }
     }
 
-    private QueuedTask findForcedTask() {
+    private synchronized QueuedTask findForcedTask() {
         QueuedTask task = null;
         List<QueuedTask> storage = new ArrayList<>();
         while (task == null && !taskQueue.isEmpty()) {
--- a/request-controller/src/main/java/org/wildfly/extension/requestcontroller/RequestController.java
+++ b/request-controller/src/main/java/org/wildfly/extension/requestcontroller/RequestController.java
@@ -396,21 +396,21 @@ public class RequestController implements Service<RequestController>, ServerActi
         }
     }
 
-    private synchronized QueuedTask findForcedTask() {
-        QueuedTask task = null;
+    private QueuedTask findForcedTask() {
+        QueuedTask _ret = null;
+        QueuedTask task;
         List<QueuedTask> storage = new ArrayList<>();
-        while (task == null && !taskQueue.isEmpty()) {
-            QueuedTask tmp = taskQueue.poll();
-            if(tmp.forceRun) {
-                task = tmp;
+        while ((task = taskQueue.poll()) != null) {
+            if (task.forceRun) {
+                _ret = task;
             } else {
-                storage.add(tmp);
+                storage.add(task);
             }
         }
-        //this screws the order somewhat, but the container is suspending anyway, and the order
-        //was never guarenteed. if we push them back onto the front we will need to just go through them again
+        // this screws the order somewhat, but the container is suspending anyway, and the order
+        // was never guarenteed. if we push them back onto the front we will need to just go through them again
         taskQueue.addAll(storage);
-        return task;
+        return _ret;
     }
 
     private static final class ControlPointIdentifier {
--- a/request-controller/src/main/java/org/wildfly/extension/requestcontroller/RequestController.java
+++ b/request-controller/src/main/java/org/wildfly/extension/requestcontroller/RequestController.java
@@ -397,12 +397,12 @@ public class RequestController implements Service<RequestController>, ServerActi
     }
 
     private QueuedTask findForcedTask() {
-        QueuedTask _ret = null;
+        QueuedTask forcedTask = null;
         QueuedTask task;
         List<QueuedTask> storage = new ArrayList<>();
-        while ((task = taskQueue.poll()) != null) {
+        while (forcedTask == null && (task = taskQueue.poll()) != null) {
             if (task.forceRun) {
-                _ret = task;
+                forcedTask = task;
             } else {
                 storage.add(task);
             }
@@ -410,7 +410,7 @@ public class RequestController implements Service<RequestController>, ServerActi
         // this screws the order somewhat, but the container is suspending anyway, and the order
         // was never guarenteed. if we push them back onto the front we will need to just go through them again
         taskQueue.addAll(storage);
-        return _ret;
+        return forcedTask;
     }
 
     private static final class ControlPointIdentifier {
--- a/request-controller/src/test/java/org/wildfly/extension/requestcontroller/WFCORE4967_TestCase.java
+++ b/request-controller/src/test/java/org/wildfly/extension/requestcontroller/WFCORE4967_TestCase.java
@@ -22,12 +22,23 @@ public class WFCORE4967_TestCase {
         });
 
         CountDownLatch latch = new CountDownLatch(THREADS_QTY);
-        createSynchronisedThreads(latch, () -> {
+        List<Thread> threads = createSynchronisedThreads(latch, () -> {
             rc.requestComplete();
-        }).forEach(Thread::start);
+        });
+        threads.forEach(Thread::start);
+        // wait until all above threads ready to fire rc.requestComplete() together with bellow rc.resume()
         latch.await();
-
         rc.resume();
+
+        for (Thread t : threads) {
+            t.join();
+        }
+
+        // simulate just enough requests after server is resumed to drain potential outstanding tasks from taskQueue
+        for (int requestNo = 0; requestNo < TASKS_QTY; requestNo++) {
+            rc.requestComplete();
+        }
+
         assertEquals(TASKS_QTY, executedTaskCount.intValue());
     }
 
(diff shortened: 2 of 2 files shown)


bstansberry:
Hi @khroolick 

Thanks for the patch!

request-controller/src/test/java/org/wildfly/extension/requestcontroller/WFCORE4967_TestCase.java
```diff
@@ -0,0 +1,63 @@
...
+            executedTaskCount.incrementAndGet();
+        });
+
+        CountDownLatch latch = new CountDownLatch(THREADS_QTY);
+        createSynchronisedThreads(latch, () -> {
+            rc.requestComplete();
+        }).forEach(Thread::start);
+        latch.await();
+
+        rc.resume();
```
bstansberry, line 30: So AIUI the basic problem is that one or more of the threads that called rc.requestComplete() when paused == true  is executing findForcedTask and has drained some tasks from taskQueue into the 'storage' list. And at the same time rc.resume is called and it therefore finds the queue empty and returns.  When the thread in findForcedTask puts its 'storage' tasks back in the queue there is nothing to execute them.

Is that a correct summary?

khroolick:
> We require that all contributions be made under the terms of the MIT License, http://www.opensource.org/licenses/mit-license.php. Do you agree to the these terms?

Yes, I agree to the these MIT licensing terms.

request-controller/src/test/java/org/wildfly/extension/requestcontroller/WFCORE4967_TestCase.java
khroolick: Correct but not full in-depth. This tmp 'storage' list is problematic in multithreading context.
It causes problems before `rc.resume()` is called as well. 
We can get 
```
java.lang.NullPointerException
	at org.wildfly.extension.requestcontroller.RequestController$QueuedTask.access$1(RequestController.java:457)
	at org.wildfly.extension.requestcontroller.RequestController.findForcedTask(RequestController.java:404)
	at org.wildfly.extension.requestcontroller.RequestController.runQueuedTask(RequestController.java:386)
	at org.wildfly.extension.requestcontroller.RequestController.requestComplete(RequestController.java:238)
```
as an example

bstansberry: @khroolick.  Yes, I saw the NPE potential and was hoping preventing that would be sufficient to fix the bug without adding the synchronized.  But your test quickly demonstrated that wasn't the case.

The synchronized makes me nervous because of possible interaction between the current synchronization on the RequestController object, which involve the server management layer manipulating the RC, and this new one which involves users of the RC queuing or completing tasks. For example the 'suspended' method holds the lock while the 'requestCountListener.done()' callback does its thing, and meanwhile threads trying to queue/complete tasks are blocking.

TBH that concern might be excessive.

I thought about instead synchronizing on the taskQueue where necessary but I don't know if that is better. With that you've got two separate locks being taken in some cases, which can have its own problems.  I didn't think hard about it though. WDYT?

khroolick: I had similar thinking, but I was not afraid of these operations as in our case we do not use them often. I didn't think too hard as well and just went with “simplest” fix I could apply.

In fact, if there is a fear of unnecessary blocking, we can improve the implementation of the `findForcedTask` method and avoid synchronization. It looks good after static analysis.

- If a thread has drained some tasks from taskQueue into the 'storage' list when `resume ()` has been called, these tasks will return to `taskQueue` and will be started under `requestComplete()` invocations. 
The only risk I see: Very unusual coincidence in which there will be too few calls requestComplete after server is resumed and timers schedule will be delayed until server will receive some traffic.

This is very very unlikely if not possible at all, as if we are already in this scenario of high concurrent access to `taskQueue` via` findForcedTask`.

request-controller/src/main/java/org/wildfly/extension/requestcontroller/RequestController.java
```diff
@@ -396,7 +396,7 @@ private boolean runQueuedTask(boolean hasPermit) {
         }
     }
 
-    private QueuedTask findForcedTask() {
+    private synchronized QueuedTask findForcedTask() {
```
khroolick, line 399:
```suggestion
    private QueuedTask findForcedTask() {
        QueuedTask _ret = null;
        QueuedTask task;
        List<QueuedTask> storage = new ArrayList<>();
        while ((task = taskQueue.poll()) != null) {
            if(task.forceRun) {
                _ret = task;
            } else {
                storage.add(task);
            }
        }
        //this screws the order somewhat, but the container is suspending anyway, and the order
        //was never guarenteed. if we push them back onto the front we will need to just go through them again
        taskQueue.addAll(storage);
        return _ret;
    }```

File: request-controller/src/test/java/org/wildfly/extension/requestcontroller/WFCORE4967_TestCase.java (hunk 1 above)
Comment (khroolick) [reply request-controller/src/test/java/org/wildfly/extension/requestcontroller/WFCORE4967_TestCase.java:31]: - I've extended unit test to simulate some additional requests after `rc.resume()` is called. 
- Tested above change with both 
 o [minimal example project](https://github.com/guestlinelabs/wildfly-timer-issue)
 o our complex EAR application

LGTM, please review this change again @bstansberry

**Comment by khroolick** [comment]
FYI: I've created a patch and deployed this fix to our production servers already.
that fixed our issue.

File: request-controller/src/main/java/org/wildfly/extension/requestcontroller/RequestController.java [hunk 3]
```diff
@@ -397,20 +397,20 @@ private boolean runQueuedTask(boolean hasPermit) {
     }
 
     private QueuedTask findForcedTask() {
-        QueuedTask task = null;
+        QueuedTask _ret = null;
```
jmesnil, line 400: I'll let @bstansberry review your changes as he already reviewed your PR

However could you please rename this variable? WildFly code convention do not use `_` to prefix variables? thanks

khroolick: done. 
One additional thing I noticed when I saw this code once again after a while @bstansberry :
There is a potential problem but as I'm not sure what forced task really is I'm not sure is it intended or just another bug: 
This code: 
```java
        while ((task = taskQueue.poll()) != null) {
            if (task.forceRun) {
                forcedTask = task;
            } else {
                storage.add(task);
            }
        }
```
will go trough all tasks in `taskQueue` so there is potential problem of dropping forced tasks if there is more than one. Only last forced task will be returned to a caller, previous ones are not added to a `storage` collection so they are just lost.

bstansberry: Apologies @khroolick, I got distracted. This should fix that:

while (forcedTask == null && (task = taskQueue.poll()) != null) {

With that the code looks much as it did before though, with different variable names, the NPE fixed and pretty redundant taskQueue.isEmpty() call removed. If that resolves the issue you were seeing though, it's fine with me.

request-controller/src/test/java/org/wildfly/extension/requestcontroller/WFCORE4967_TestCase.java
bstansberry: I'm unresolving this just so the discussion stays visible in the github ui as it may be useful in the future.

request-controller/src/main/java/org/wildfly/extension/requestcontroller/RequestController.java
khroolick: Same on my side, it takes longer than it should :-)
I fully agree with you @bstansberry, applied suggested change and I'm happy to merge it in the current shape.

khroolick:
Is there anything I can do to help merging this pull request in?
It fixes quite serious BUG leading to EJB Scheduled tasks being not run at all.

khroolick:
> There has been no activity on this PR for 30 days. It will be auto-closed after 90 days.
May we merge this in pretty please? 
the change is safe since the code is almost untouched with different variable names and the NPE fixed.

The above change was applied to our prod environment and fixed all issues.

bstansberry:
@jmesnil note to self (and you in case you remind me) -- I want to be sure we clear this one in the next week.

request-controller/src/test/java/org/wildfly/extension/requestcontroller/WFCORE4967_TestCase.java
```diff
@@ -0,0 +1,74 @@
...
+        for (int requestNo = 0; requestNo < TASKS_QTY; requestNo++) {
+            rc.requestComplete();
+        }
+
+        assertEquals(TASKS_QTY, executedTaskCount.intValue());
+    }
+
+    private RequestController suspendedRCWithQueuedTasks(int i, Runnable whenExecuted) {
+        RequestController _ret = new RequestController(false);
+        _ret.suspended(() -> {
```
jmesnil, line 47: could you please rename all `_ret` variables ? that's the only remaining request before I can merge this PR.
thanks

jmesnil (changes requested):

request-controller/src/test/java/org/wildfly/extension/requestcontroller/WFCORE4967_TestCase.java
khroolick: Sure, done.

bstansberry (approved):

khroolick:
Are flaky tests the only reason why this is not merged in yet? 

Could you help me merge this in before Christmas @bstansberry? 
It would be so great!

bstansberry:
@khroolick I'm so sorry! I thought this was merged in October so I stopped looking for it.

PR #4206, opened 2020-05-15
[WFCORE-4750] Using regex for role in Elytron

darranl:
This PR combines two PRs to get the feature into CI along with the component upgrade.

Combination of the following PRs: -
  https://github.com/wildfly/wildfly-core/pull/4205
  https://github.com/wildfly/wildfly-core/pull/4017


bstansberry:
@fjuma Is this ok? You had a comment on #4017 basically approving it but if you see this an officially approve it would be good.

PR #4211, opened 2020-05-19
[WFCORE-4973] Increasing LongOutputTestCase timeout from 10 to 20 seconds

ropalka:
https://issues.redhat.com/browse/WFCORE-4973

### DIFF
--- a/testsuite/standalone/src/test/java/org/jboss/as/test/integration/management/cli/LongOutputTestCase.java
+++ b/testsuite/standalone/src/test/java/org/jboss/as/test/integration/management/cli/LongOutputTestCase.java
@@ -10,6 +10,7 @@ import org.jboss.as.cli.impl.CommandContextConfiguration;
 import org.jboss.as.cli.impl.ReadlineConsole;
 import org.jboss.as.test.integration.management.util.CLITestUtil;
 import org.jboss.as.test.shared.TestSuiteEnvironment;
+import org.jboss.as.test.shared.TimeoutUtil;
 import org.jboss.logging.Logger;
 import org.junit.After;
 import org.junit.Assert;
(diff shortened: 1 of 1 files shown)


testsuite/standalone/src/test/java/org/jboss/as/test/integration/management/cli/LongOutputTestCase.java
```diff
@@ -389,7 +389,7 @@ public void testEnableOutputPagingViaXml() throws Exception {
     private void testDisabledOutputPaging() throws Exception {
         consoleWriter.println("/subsystem=elytron:read-resource-description(recursive=true)");
         Assert.assertFalse(consoleWriter.checkError());
-        String window = queue.poll(10, TimeUnit.SECONDS);
+        String window = queue.poll(100, TimeUnit.SECONDS);
```
jmesnil, line 392: 100 seconds is 1 order of magnitude more. It seems a lot.

Would doubling to 20s be enough?
Or could we instead use org.jboss.as.test.shared.TimeoutUtil#adjust to adjust the time of this test?

ropalka: This problem pops up if I'm running whole WildFly test suite only (~ 2.5 hours).
If I run the test standalone, it always passes for me. 
You are right @jmesnil 100 seconds is much more higher but
test usually don't spend so much time in it. It is just safer upper bound.
If you still prefer 20 seconds instead, please let me know - I will change it and will test it locally.

jmesnil: @ropalka Let's double it and use org.jboss.as.test.shared.TimeoutUtil#adjust to adjust it automatically when it is run in our CI, thanks

ropalka:
I doubled the timeout value + used TimeoutUtil.adjust() method as you suggested @jmesnil

ropalka:
Fixed @darranl - thanks for spotting it out.

darranl:
CI errors appear unrelated, adding to ready to merge batch.

PR #4212, opened 2020-05-20
[WFCORE-4974] Bump the org.wildfly.core.testrunner.Server start/stop …

bstansberry:
…timeouts

https://issues.redhat.com/browse/WFCORE-4974

### DIFF
--- a/testsuite/test-runner/src/main/java/org/wildfly/core/testrunner/Server.java
+++ b/testsuite/test-runner/src/main/java/org/wildfly/core/testrunner/Server.java
@@ -17,6 +17,8 @@ import java.net.URI;
 import java.nio.file.Files;
 import java.nio.file.Path;
 import java.nio.file.Paths;
+import java.security.AccessController;
+import java.security.PrivilegedAction;
 import java.util.Objects;
 import java.util.concurrent.CancellationException;
 import java.util.concurrent.ExecutionException;
@@ -37,6 +39,7 @@ import org.wildfly.core.launcher.CommandBuilder;
 import org.wildfly.core.launcher.Launcher;
 import org.wildfly.core.launcher.ProcessHelper;
 import org.wildfly.core.launcher.StandaloneCommandBuilder;
+import org.wildfly.security.manager.WildFlySecurityManager;
 
 /**
  * encapsulation of a server process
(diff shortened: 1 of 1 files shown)


bstansberry:
@jamezp @jmesnil I pulled these new numbers out of thin air. If there are equivalent defaults in the ARQ containers, aligning would make sense just as a 'one less variation to keep in mind' kind of thing. I think a min of 60s for a start timeout is the right number though, not less.

bstansberry:
I killed CI on this because we need the resources elsewhere today. We can trigger a 'retest' later.

testsuite/test-runner/src/main/java/org/wildfly/core/testrunner/Server.java
```diff
@@ -66,8 +66,8 @@
     private final String managementProtocol = System.getProperty("management.protocol", "remote+http");
 
     // timeouts
-    private final int startupTimeout = Integer.getInteger("server.startup.timeout", 30);
-    private final int stopTimeout = Integer.getInteger("server.stop.timeout", 10);
+    private final int startupTimeout = Integer.getInteger("server.startup.timeout", 60);
```
jmesnil, line 69: I'm fine with doubling these values.

I also wonder whether we could introduce a mechanism similar to `org.jboss.as.test.shared.TimeoutUtil` to influence the timeouts with the `ts.timeout.factor` system property?

bstansberry: Makes sense. We'd need to port TimeoutUtil though as I don't think we want to add the dep on testsuite-shared.  That's no big deal though.

bstansberry:
retest this please

testsuite/test-runner/src/main/java/org/wildfly/core/testrunner/Server.java
bstansberry: I updated the commit with the relevant code added directly to the Server class.

jamezp: I'm pretty sure the `TimeoutUtil` is already in core right? https://github.com/wildfly/wildfly-core/blob/master/testsuite/shared/src/main/java/org/jboss/as/test/shared/TimeoutUtil.java

jmesnil: That's correct, it's in core but for classes shared by the test suite. afaict, the testrunner does not depend on these shared classes.

PR #4213, opened 2020-05-21
[WFCORE-4955] Create a wildfly-network OutboundConnection interface f…

yersan:
…rom AbstractOutboundConnectionService and expose via capabilities

Exposes a new interface `OutboundConnection` on `org.jboss.as.network` via capabilities to helping out to make `org.jboss.as.remoting` optional on `org.jboss.as.ejb3`. It deprecates the existing service name in favor of the one generated from the capability.

Jira issue: https://issues.redhat.com/browse/WFCORE-4955

### DIFF
--- a/network/src/main/java/org/jboss/as/network/OutboundConnection.java
+++ b/network/src/main/java/org/jboss/as/network/OutboundConnection.java
@@ -1,0 +1,54 @@
+/*
+ * Copyright 2020 Red Hat, Inc.
+ *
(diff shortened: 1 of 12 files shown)


network/src/main/java/org/jboss/as/network/OutboundConnection.java
```diff
@@ -0,0 +1,38 @@
+package org.jboss.as.network;
```
jmesnil, line 1: The file header is missing

PR #4215, opened 2020-05-23
[WFCORE-4982] Convert DeploymentUnitPhaseService to use a for loop instead of a stream to capture known subsystem names.

darranl:
https://issues.redhat.com/browse/WFCORE-4982

### DIFF
--- a/server/src/main/java/org/jboss/as/server/deployment/DeploymentUnitPhaseService.java
+++ b/server/src/main/java/org/jboss/as/server/deployment/DeploymentUnitPhaseService.java
@@ -29,7 +29,6 @@ import java.util.List;
 import java.util.ListIterator;
 import java.util.Set;
 import java.util.concurrent.atomic.AtomicBoolean;
-import java.util.stream.Collectors;
 
 import org.jboss.as.server.deployment.module.ModuleSpecification;
 import org.jboss.as.server.logging.ServerLogger;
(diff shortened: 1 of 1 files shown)


server/src/main/java/org/jboss/as/server/deployment/DeploymentUnitPhaseService.java
```diff
@@ -141,10 +140,17 @@ public void handleEvent(final ServiceController<?> controller, final LifecycleEv
...
         }
 
-        Set<String> subSystemNames = list.stream().map(e -> e.getSubsystemName()).collect(Collectors.toSet());
-        Set<String> registeredSubSystems = (phase == Phase.STRUCTURE) ? new HashSet<>() : deploymentUnit.getAttachment(Attachments.REGISTERED_SUBSYSTEMS);
-        registeredSubSystems.addAll(subSystemNames);
-        deploymentUnit.putAttachment(Attachments.REGISTERED_SUBSYSTEMS, registeredSubSystems);
+        final Set<String> registeredSubSystems;
+        if (phase == Phase.STRUCTURE) {
+            registeredSubSystems = new HashSet<>();
+            deploymentUnit.putAttachment(Attachments.REGISTERED_SUBSYSTEMS, registeredSubSystems);
```
jmesnil, line 146: The code is correct but I find the new sequence hard to read: we instantiate the `registeredSubSystems` and put it as an attachment. And then, *later* we fill registeredSubSystems with the names of the subsystems...

````
Set<String> registeredSubSystems = (phase == Phase.STRUCTURE) ? new HashSet<>() : deploymentUnit.getAttachment(Attachments.REGISTERED_SUBSYSTEMS);
for (...) {
}
deploymentUnit.putAttachment(Attachments.REGISTERED_SUBSYSTEMS, registeredSubSystems);
````

darranl: That was to remove a redundant put which I believe will mean the underlying collection of attachments will need to compare the keys of the attachments present for every phase except for Structure for each deployment being deployed.


Alternatively the if would need to be repeated just for the purpose of separating out the putAttachment to a later point.

PR #4217, opened 2020-05-26
Fix for WFCORE-4984, Upgrade tests to use wildfly-jar-maven-plugin Alpha3

jfdenise:
Issue: https://issues.redhat.com/browse/WFCORE-4984

Impact on tests is a change to configure extra content to package in jar


jmesnil:
@jfdenise thanks

PR #4228, opened 2020-06-09
[WFCORE-4995] Bump management API version to 14.0.0

jmesnil:
and `urn:jboss:domain:14.0` for the XSD

JIRA: https://issues.redhat.com/browse/WFCORE-4995

### DIFF
--- a/controller/src/main/java/org/jboss/as/controller/parsing/Namespace.java
+++ b/controller/src/main/java/org/jboss/as/controller/parsing/Namespace.java
@@ -99,12 +99,15 @@ public enum Namespace {
     DOMAIN_12_0(12, "urn:jboss:domain:12.0"),
 
     // WF 20
-    DOMAIN_13_0(13, "urn:jboss:domain:13.0");
+    DOMAIN_13_0(13, "urn:jboss:domain:13.0"),
+
+    // WF 21
+    DOMAIN_14_0(14, "urn:jboss:domain:14.0");
 
     /**
      * The current namespace version.
      */
-    public static final Namespace CURRENT = DOMAIN_13_0;
+    public static final Namespace CURRENT = DOMAIN_14_0;
 
     public static final Namespace[] ALL_NAMESPACES = domainValues();
 
--- a/core-feature-pack/src/main/resources/configuration/domain/template.xml
+++ b/core-feature-pack/src/main/resources/configuration/domain/template.xml
@@ -24,7 +24,7 @@
   ~ */
   -->
 
-<domain xmlns="urn:jboss:domain:13.0">
+<domain xmlns="urn:jboss:domain:14.0">
 
     <extensions>
         <?EXTENSIONS?>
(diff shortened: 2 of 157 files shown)


host-controller/src/main/java/org/jboss/as/host/controller/parsing/DomainXml_13.java
```diff
@@ -694,106 +602,6 @@ private void initializeRolloutPlans(ModelNode address, List<ModelNode> list) {
         list.add(addOp);
     }
```
yersan, line 604: `writeAccessControl` can also be removed since this file will be used for read only.
https://github.com/wildfly/wildfly-core/pull/4228/files#diff-3a186e12b2e248de03d22e9991e1d13cR686

host-controller/src/main/java/org/jboss/as/host/controller/parsing/HostXml_13.java
```diff
@@ -1512,174 +1399,6 @@ private void parseSocketBindingGroup(final XMLExtendedStreamReader reader, final
...
-        if (!profileNode.hasDefined(SUBSYSTEM)) {
-            return;
-        }
-
-        writer.writeStartElement(Element.PROFILE.getLocalName());
-        writeSubsystems(profileNode, writer, context);
-        writer.writeEndElement();
-    }
-
     /*
```
yersan, line 1402: I would also expect the following methods removed from this file:
`writeNativeManagementProtocol` https://github.com/wildfly/wildfly-core/pull/4228/files#diff-ae4aaafa96590692ce977218d56e3326R1448

`writeHttpManagementProtocol` https://github.com/wildfly/wildfly-core/pull/4228/files#diff-ae4aaafa96590692ce977218d56e3326R1467

`writeAuditLog` https://github.com/wildfly/wildfly-core/pull/4228/files#diff-ae4aaafa96590692ce977218d56e3326R1514

server/src/main/java/org/jboss/as/server/parsing/StandaloneXml_13.java
```diff
@@ -717,120 +701,6 @@ private void setServerName(final ModelNode address, final List<ModelNode> operat
         }
     }
 
-    void writeContent(final XMLExtendedStreamWriter writer, final ModelMarshallingContext context)
```
yersan, line 720: There are still write methods in this file, however it is supposed to be a read-only parser, see all the `write*` method names.

host-controller/src/main/java/org/jboss/as/host/controller/parsing/HostXml_13.java
jmesnil: fixed

server/src/main/java/org/jboss/as/server/parsing/StandaloneXml_13.java
jmesnil: fixed

jmesnil:
@yersan thanks for the review, I've fixed the commit and updated the PR with it.

PR #4230, opened 2020-06-10
[WFCORE-5002] Upgrade the minimum Maven version to 3.6.0 and default …

jamezp:
…the mvnw version to 3.6.3. Remove the unneeded config file.

https://issues.redhat.com/browse/WFCORE-5002


jmesnil:
@jamezp thanks

PR #4231, opened 2020-06-15
[WFCORE-5006] Improve readability of domain host controller server logs in tests

rachmatowicz:
This PR does the following:
- adds host name information to HostController start and started log messages to allow differentiating between multiple HostController instances on the same host, which often arises in domain test cases. 

More details: https://issues.redhat.com/browse/WFCORE-5006


jmesnil:
@rachmatowicz thanks

PR #4234, opened 2020-06-16
[WFCORE-4993] Organize feature pack code to facilitate separate EE 8 and EE 9 variants

bstansberry:
https://issues.redhat.com/browse/WFCORE-4993
Also https://issues.redhat.com/browse/WFCORE-5008 which is needed to allow license generation from the multiple license configuration files the result from the refactoring described below.

This is essentially a code reorganization:

1) Convert the 'core-feature-pack' maven module to a parent module containing submodules related to both the legacy core fp and the galleon one. The top level core-galleon-pack maven module is eliminated.

2) Under 'core-feature-pack' create a 'common' submodule that holds content that is expected to be usable in the legacy fp, the current EE 8 galleon FP, and a future EE 9 galleon FP. This includes the various module.xml files, a license information file for the artifacts referenced in those module.xml files, other shared content, and the dependency set declaration in the pom for all the artifacts referenced in the module.xml files.

3) Under 'core-feature-pack' create a 'ee-8-api' submodule that holds content that is expected to be usable in the legacy fp and the current EE 8 galleon FP but not by a future EE 9 galleon FP. This includes the various module.xml files, a license information file for the artifacts referenced in those module.xml files,, and the dependency set declaration in the pom for all the artifacts referenced in the module.xml files.

4) Under 'core-feature-pack' create a 'galleon-common' submodule that holds content that is expected to be usable in the current EE 8 galleon FP and a future EE 9 galleon FP. It is not meant to include galleon content that would not be used in both FP variants, e.g. deprecated config items that won't appear in the EE 9 variant without a conscious decision to bring them in.

5) Under 'core-feature-pack' create a 'galleon-feature-pack' submodule that replaces the previous top level 'core-galleon-pack' module.  It copies in content from the 'common', 'ee-8-api' and 'galleon-common' sibling maven modules discussed above, and also depends on the 'common' and 'ee-8-api' poms, as a source of dependency information. It also has some content of its own that is unique to the EE 8 galleon FP.

6) Under 'core-feature-pack' create a 'legacy-feature-pack' submodule that replaces the function of the previous top level 'core-feature-pack' module.  It copies in content from the 'common' and 'ee-8-api' sibling maven modules discussed above, and also depends on the 'common' and 'ee-8-api' poms, as a source of dependency information. It also has some content of its own that is unique to the legacy FP.

A goal here is to remove the current requirement to declare detailed dependency info in the poms for both the legacy FP and the galleon FP.  @jfdenise we need to confirm this doesn't introduce problems for external FP authors.

The 'galleon-common' content needs further investigation to see if more things there are not truly common between EE 8 and EE 9 feature packs. I'd like that to be future work.

In the future the Galleon FPs in full WF may be able to directly depend on artifacts produced by the submodules here as source inputs, eliminating the need for feature pack dependencies. See GAL-319.


core-feature-pack/galleon-feature-pack/pom.xml
```diff
@@ -0,0 +1,372 @@
...
+
+        <!-- module and copy artifact dependencies -->
+
+        <!-- Bring in all the common deps in one go -->
+        <dependency>
+            <groupId>${project.groupId}</groupId>
+            <artifactId>wildfly-core-feature-pack-common</artifactId>
+            <type>pom</type>
+            <scope>provided</scope>
+            <!-- Exclude the ee-8-api dependencies -->
```
jmesnil, line 73: I'm not sure to follow.  `wildfly-core-feature-pack-common` "holds content that is expected to be usable in the legacy fp, the current EE 8 galleon FP, and a future EE 9 galleon FP". It depends on ee-8-api though but we exclude them in this feature pack to bring them back with wildfly-core-feature-pack-ee-8-api.

The comments seems to imply that this is for all ee-8-api dependencies, but you only exclude the JSON API.

core-feature-pack/legacy-feature-pack/pom.xml
```diff
@@ -0,0 +1,408 @@
...
+         xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/xsd/maven-4.0.0.xsd">
+    <modelVersion>4.0.0</modelVersion>
+
+    <parent>
+        <groupId>org.wildfly.core</groupId>
+        <artifactId>wildfly-core-feature-pack-parent</artifactId>
+        <version>13.0.0.Beta1-SNAPSHOT</version>
+    </parent>
+
+    <artifactId>wildfly-core-feature-pack</artifactId>
```
jmesnil, line 36: should we rename the artifactId to `wildfly-core-legacy-feature-pack` and accept that starting from now a feature pack means Galleon unless specified otherwise?

core-feature-pack/legacy-feature-pack/pom.xml
```diff
@@ -0,0 +1,408 @@
...
+    <dependencies>
+
+        <!-- module and copy artifact dependencies -->
+
+        <!-- Bring in all the common deps in one go -->
+        <dependency>
+            <groupId>${project.groupId}</groupId>
+            <artifactId>wildfly-core-feature-pack-common</artifactId>
+            <type>pom</type>
+        </dependency>
```
jmesnil, line 58: related to my comment above, you don't exclude `Jakarta.json:jakarta.json-api` as you did in the Galleon feature pack. Is it intended?

core-feature-pack/legacy-feature-pack/pom.xml
jmesnil: (although it is not required, I like to have Maven GAV matching the names of the directories as much as possible)

core-feature-pack/galleon-feature-pack/pom.xml
bstansberry: I only excluded json because that was the one that led to a ban-transitive-dependencies failure.  Excluding the others too would make sense just to avoid confusion or having to do it later.

It does make me wonder why there was no error reported for any of the others.

core-feature-pack/legacy-feature-pack/pom.xml
bstansberry: We'd need to discuss with the keycloak folks to see if it breaks them.

core-feature-pack/galleon-feature-pack/pom.xml
bstansberry: Hmm, I thought you were asking about the exclusion in the 'common' module pom. TBH I forget exactly why this was needed; I'll investigate.

bstansberry: I removed it and building with -rf core-feature-pack worked fine so it seems this wasn't necessary.

In the 'common' pom.xml I went ahead and excluded all the EE API artifacts.

core-feature-pack/legacy-feature-pack/pom.xml
bstansberry: It looks like it was not needed in the Galleon one so I removed it.

jfdenise:
@bstansberry , I don't see immediate issue with the changes in core. That is transparent to the user. The "provided" scope works as expected, the dependencies are not retrieved when building an external FP that depends on core. An issue will arise when Full FPs don't depend on core anymore. Users who have defined a wildfly-feature-pack-build.xml or wildfly-user-feature-pack-build.xml that contains a dependency on wildfly-core would have to be updated to no more express the dependency. The creation of such feature-pack will fail if no dependency on wildfly-core galleon-pack is present in external FP pom file (this is the way to do it, transitive deps are automatically resolved thanks to maven dependencies and not required to be in the pom file).

bstansberry:
@jfdenise I'm not clear on your point re full FPs. Are you describing a general issue re dropping the wildfly-core FP from the full FPs? If so, yes, understood, that's in incompatible change. Or are you describing something specific to how I've structured the dependencies in this PR?

jfdenise:
@bstansberry , it was a comment on the general issue.

core-feature-pack/legacy-feature-pack/pom.xml
jmesnil: never mind, that's a silly suggestion. This artifact is consumed by other project than wildFly and it is not correct to change its GAV just for the sake of changing it.

jmesnil:
While reviewing your PR I noticed you've added a `galleon-common` dir under https://github.com/bstansberry/wildfly-core/tree/WFCORE-4993/core-feature-pack which contains code but is not a maven module.

Is it an omission or am I missing how it is used?

bstansberry:
Re galleon-common it has no pom because no other artifact currently needs it's pom (unlike 'common' and 'ee-8-api' which are dependencies of the f-p modules.  It also produces no artifact; it's just a collection of source that gets copied via the maven-resource-plugin.

We chatted earlier today about the possible benefit of having 'common', 'ee-8-api' and 'galleon-common' producing an artifact so a Galleon f-p in another code base, using a Galleon WF plugin with GAL-319 support, could consume that artifact and have direct access to the code and no need to depend on the wildfly-core feature pack.  But to do this in WF full we'd need to first deal with the issues described in https://lists.jboss.org/pipermail/wildfly-dev/2020-June/007342.html. So I don't think we need to block on that.

jmesnil:
ah you're right, you mentioned it earlier...
For references, it's used by https://github.com/bstansberry/wildfly-core/blob/1adc06c724ef75242928694478bd894f3dbce376/core-feature-pack/galleon-feature-pack/pom.xml#L171

bstansberry:
@jmesnil I've rebased this.

@dosoudil @rgdoliveira FYI.

core-feature-pack/legacy-feature-pack/pom.xml
bstansberry: If we can eventually reach a point where the WildFly galleon FPs no longer depend on the wildfly-core FP, then the GAV of the wildfly-core FP becomes much less importance as most (maybe all) of its use outside core's own build will be gone.

jmesnil:
@bstansberry thanks

PR #4240, opened 2020-06-21
[WFCORE-5016] Have the feature pack 'source modules' produce zip arti…

bstansberry:
…facts

https://issues.redhat.com/browse/WFCORE-5016


jmesnil:
@bstansberry thanks

PR #3841, opened 2019-06-25
[WFCORE-944] Allow more file-based truststores than just JKS

darranl:
https://issues.jboss.org/browse/WFCORE-944

This PR follows up from https://github.com/wildfly/wildfly-core/pull/3782 to resolve a merge conflict.

PR #4242, opened 2020-06-22
[WFCORE-944] truststore path is ignored if provider is not JKS

rmartinc:
Issue: https://issues.redhat.com/browse/WFCORE-944

I just used @darranl 's fix (https://github.com/wildfly/wildfly-core/pull/3841) but a manual test that used PKCS12 was modified to include this use-case. Indeed the test is better now. Previously the provider was not specified in the configuration (server-identity/ssl or authentication/truststore) and that seems to work in openjdk but fails in other implementations (IBM for example). So we are killing two birds with one stone.


darranl:
It may be in the existing code, I can't quite see how the test activates a trust store with a different type?

rmartinc:
@darranl Yes, you are right, I'm re-using a previous PKCS12 test. Check this previous PR https://github.com/wildfly/wildfly-core/pull/4196 that adds it. This one just complements that previous one (adding the provider to everything and to the client part too).

jmesnil:
@rmartinc thanks

PR #4245, opened 2020-06-22
[WFCORE-4994] Add the ability to adjust the case of a user name using an Elytron principal transformer

SoniaZaldana:
Jira: https://issues.redhat.com/browse/WFCORE-4994

To pass CI, this will require an Elytron upgrade.

Analysis Document: https://github.com/wildfly/wildfly-proposals/pull/315
Requires: https://issues.redhat.com/browse/ELY-1987
Elytron PR with tests: https://github.com/wildfly-security/wildfly-elytron/pull/1418
Wildfly PR with tests and documentation updates: https://github.com/wildfly/wildfly/pull/13379

### DIFF
--- a/elytron/src/test/java/org/wildfly/extension/elytron/DomainTestCase.java
+++ b/elytron/src/test/java/org/wildfly/extension/elytron/DomainTestCase.java
@@ -109,6 +109,43 @@ public class DomainTestCase extends AbstractSubsystemTest {
         TestEnvironment.activateService(services, Capabilities.SECURITY_DOMAIN_RUNTIME_CAPABILITY, "SubjectAltNameEvidenceDecoderDomain");
         TestEnvironment.activateService(services, Capabilities.SECURITY_DOMAIN_RUNTIME_CAPABILITY, "SubjectEvidenceDecoderDomain");
         TestEnvironment.activateService(services, Capabilities.SECURITY_DOMAIN_RUNTIME_CAPABILITY, "SourceAddressRoleDecoderDomain");
+        TestEnvironment.activateService(services, Capabilities.SECURITY_DOMAIN_RUNTIME_CAPABILITY, "LowerCasePrincipalTransformerDomain");
+        TestEnvironment.activateService(services, Capabilities.SECURITY_DOMAIN_RUNTIME_CAPABILITY, "UpperCasePrincipalTransformerDomain");
+
+    }
+
+    @Test
+    public void testCasePrincipalTransformerLowerCase() throws Exception {
+        init();
+        ServiceName serviceName = Capabilities.SECURITY_DOMAIN_RUNTIME_CAPABILITY.getCapabilityServiceName("LowerCasePrincipalTransformerDomain");
+        SecurityDomain domain = (SecurityDomain) services.getContainer().getService(serviceName).getValue();
+        Assert.assertNotNull(domain);
+
+        ServerAuthenticationContext context = domain.createNewAuthenticationContext();
+        context.setAuthenticationName("USER1"); // the realm contains "user1"
+        Assert.assertTrue(context.exists());
+        context.authorize();
+        context.succeed();
+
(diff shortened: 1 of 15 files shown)


elytron/src/test/java/org/wildfly/extension/elytron/DomainTestCase.java
```diff
@@ -109,6 +109,43 @@ private void init() throws Exception {
...
+
+    @Test
+    public void testCasePrincipalTransformerUpperCase() throws Exception {
+        init();
+        ServiceName serviceName = Capabilities.SECURITY_DOMAIN_RUNTIME_CAPABILITY.getCapabilityServiceName("UpperCasePrincipalTransformerDomain");
+        SecurityDomain domain = (SecurityDomain) services.getContainer().getService(serviceName).getValue();
+        Assert.assertNotNull(domain);
+
+        ServerAuthenticationContext context = domain.createNewAuthenticationContext();
+        context.setAuthenticationName("user8");
```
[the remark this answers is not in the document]
SoniaZaldana: fixed!

fjuma:
@SoniaZaldana This PR has some conflicts that need to be resolved whenever you get a chance.

SoniaZaldana:
@fjuma I resolved the conflicts.

darranl:
This PR is currently being processed via the RFE process so it still active.

elytron/src/main/java/org/wildfly/extension/elytron/ElytronSubsystemParser11_0.java
```diff
@@ -33,11 +33,6 @@ String getNameSpace() {
         return ElytronExtension.NAMESPACE_11_0;
     }
 
-    @Override
-    protected PersistentResourceXMLDescription getMapperParser() {
-        return new MapperParser().getParser();
```
fjuma: Before, you had updated ElytronSubsystemParser10_0.java (i.e., the previous parser), to update the return value to `return new MapperParser(MapperParser.Version.VERSION_6_0).getParser();`. That change should be removed from that file and added here instead (since now ElytronSubsystemParser11_0.java is the previous parser).

SoniaZaldana: In that case, as ElytronSubsystemParser10_0 is also supposed to return version 6, I leave the return value as  `new MapperParser(MapperParser.Version.VERSION_6_0).getParser()` there too, right? Previously, that parser used to return the current version.

fjuma: Ah, just realized that the newly introduced `MapperParser.Version.VERSION_6_0` corresponds to `ElytronSubsystemParser10_0` and is also needed for `ElytronSubsystemParser11_0`. This means it is correct to specify this version in `ElytronSubsystemParser10_0` and to remove the `getMapperParser()` method from `ElytronSubsystemParser11_0`.

To help make things clearer though, I think it would be good to rename `MapperParser.CURRENT` so that it references the actual version instead. That way, the next time we add a new version in `MapperParser`, we won't need to worry about updating an older version of the parser from `CURRENT` to the previous version.

SoniaZaldana: So just verify I understood correctly, you're suggesting we rename `MapperParser.CURRENT` to say `MapperParser.VERSION_7_0` and then in the ElytronSubsystemParser12_0 class I instantiate the mapper parser with version 7, as opposed to leaving an empty constructor? That way in the future, we won't have to update ElytronSubsystemParser12_0.

fjuma: Yes, exactly.

fjuma (approved):

darranl:
@SoniaZaldana FYI this one has a merge conflict now that will need resolving.

darranl (approved):

PR #4246, opened 2020-06-23
[WCORE-5014] Remove the PropertySorter and specially handle the file parameter. 

jamezp:
- https://issues.redhat.com/browse/WFCORE-5021
- https://issues.redhat.com/browse/WFCORE-5014

First commit includes the JBoss Log Manager upgrade with a bug fix this fix exposed:
- https://issues.redhat.com/browse/LOGMGR-279

The second commit fixes the issue described in [WFCORE-5014](https://issues.redhat.com/browse/WFCORE-5014). During start if an operation timeout occurs log messages may be lost. This removes the `PropertySorter` which caused some configurations options to be set more than once or configure options that should not be configured. In this case it resulted in files not being appended to.


gaol:
@jamezp is the upgrade of `jboss-logmanager` to `2.1.17.Final` required to solve WFCORE-5014 ? I am asking because I need to know if we need to upgrade `jboss-logmanager` as well in downstream, thanks .

jamezp:
> @jamezp is the upgrade of `jboss-logmanager` to `2.1.17.Final` required to solve WFCORE-5014 ? I am asking because I need to know if we need to upgrade `jboss-logmanager` as well in downstream, thanks .

@gaol I would suggest upgrading. IIRC the plan is to upgrade downstream to 2.1.16.Final and 2.1.17.Final only adds a single fix. There is a workaround for the exposed bug, but I really think the log manager bug fix should be included.

gaol:
@jamezp The downstream PR failed the CI with the exception like:
```
org.junit.ComparisonFailure: expected:<Test[ first ]message> but was:<Test[_first_]message>
	at org.junit.Assert.assertEquals(Assert.java:115)
	at org.junit.Assert.assertEquals(Assert.java:144)
	at org.jboss.as.test.integration.logging.handlers.SocketHandlerTestCase.testOutboundSocketBindingChange(SocketHandlerTestCase.java:273)
```
, it looks that the test `SocketHandlerTestCase#testWithFilter` affects `SocketHandlerTestCase#testOutboundSocketBindingChange`, but I don't know why it does not affect upstream.

The downstream fix can pass the tests after the changes like: https://github.com/jamezp/wildfly-core/compare/WFCORE-5014...gaol:fix-filter-test-in-logging, would you please check to see if there is anything we need to do in upstream? and do you think it is OK to change the downstream test this way to pass the CI? Thanks :)

jamezp:
Just a note on this I should have added previously. A test for this is fairly complicated to write given how the failure needs to occur to expose the bug. I've tested this manually several times and this fix definitely works. If I can think of a way to test this I'll follow up with a separate PR.

jmesnil:
@jamezp thanks, I've merged the PR as it is

PR #4248, opened 2020-06-26
[WFCORE-482] Add log4j2 support for WildFly

boris-unckel:

* Added dependency to log4j2 API
* Added dependency to log4j2-jboss-logmanager
* Modified logging subsystem to provide log4j2 API
* Added integration test

### DIFF
--- a/core-feature-pack/common/pom.xml
+++ b/core-feature-pack/common/pom.xml
@@ -148,6 +148,16 @@
         </dependency>
 
         <dependency>
+            <groupId>org.jboss.logmanager</groupId>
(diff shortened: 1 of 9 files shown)


jamezp:
Hi @boris-unckel. Unfortunately for something like this we have a process we have to adhere to. I truly appreciate this work though. I'm going to work on pushing this process faster. I did update my branch yesterday, https://github.com/jamezp/wildfly-core/tree/WFCORE-482. Hopefully we can get this pushed through sooner rather than later :)

boris-unckel:
You're WFCORE-482 looks great, thanks for providing that.

PR #4250, opened 2020-06-29
[WFCORE-5010] Startup error messages caused by expression where expressions are not allowed are confusing

lvydra:
https://issues.redhat.com/browse/WFCORE-5010

### DIFF
--- a/controller/src/main/java/org/jboss/as/controller/OperationContextImpl.java
+++ b/controller/src/main/java/org/jboss/as/controller/OperationContextImpl.java
@@ -63,7 +63,6 @@ import java.util.concurrent.TimeUnit;
 import java.util.concurrent.TimeoutException;
 import java.util.function.Consumer;
 import java.util.function.Supplier;
-import java.util.regex.Pattern;
 
 import org.jboss.as.controller._private.OperationFailedRuntimeException;
 import org.jboss.as.controller.access.Action;
@@ -144,10 +143,6 @@ final class OperationContextImpl extends AbstractOperationContext {
     private static final Set<Action.ActionEffect> WRITE_CONFIG = EnumSet.of(Action.ActionEffect.WRITE_CONFIG);
     private static final Set<Action.ActionEffect> WRITE_RUNTIME = EnumSet.of(Action.ActionEffect.WRITE_RUNTIME);
     private static final Set<Action.ActionEffect> ALL_READ_WRITE = EnumSet.of(Action.ActionEffect.READ_CONFIG, Action.ActionEffect.READ_RUNTIME, Action.ActionEffect.WRITE_CONFIG, Action.ActionEffect.WRITE_RUNTIME);
-
-    /** Pattern that can be used to identify capabilities name strings that include expression syntax
-     * started with one or more of any character, followed by '.${' sequence, then one or more of any character ended with '}' */
-    private static final Pattern UNSUPPORTED_EXPRESSION_PATTERN = Pattern.compile(".+\\.\\$\\{.+\\}");
 
     private final ModelControllerImpl modelController;
     private final OperationMessageHandler messageHandler;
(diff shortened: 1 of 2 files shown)


controller/src/main/java/org/jboss/as/controller/OperationContextImpl.java
```diff
@@ -347,32 +349,38 @@ private boolean validateCapabilities() {
...
-                    String formattedCapability = ignoreContext
-                            ? ControllerLogger.ROOT_LOGGER.formattedCapabilityName(id.getName())
-                            : ControllerLogger.ROOT_LOGGER.formattedCapabilityId(id.getName(), id.getScope().getName());
-                    Set<PathAddress> possiblePoints = managementModel.getCapabilityRegistry().getPossibleProviderPoints(id);
-                    if (msg != null) {
-                        msg = appendPossibleProviderPoints(msg, formattedCapability, possiblePoints);
-                    }
-                    if (bootMsg != null) {
-                        bootMsg = appendPossibleProviderPoints(bootMsg, formattedCapability, possiblePoints);
+                    if (id.getName().matches(UNSUPPORTED_EXPRESSION_PATTERN)) {
```
darranl, line 355: What kind of performance overhead will this cause?  This pattern is going to have to be recompiled for every ID encountered.

controller/src/main/java/org/jboss/as/controller/OperationContextImpl.java
```diff
@@ -144,6 +144,8 @@
     private static final Set<Action.ActionEffect> WRITE_RUNTIME = EnumSet.of(Action.ActionEffect.WRITE_RUNTIME);
     private static final Set<Action.ActionEffect> ALL_READ_WRITE = EnumSet.of(Action.ActionEffect.READ_CONFIG, Action.ActionEffect.READ_RUNTIME, Action.ActionEffect.WRITE_CONFIG, Action.ActionEffect.WRITE_RUNTIME);
 
+    private static final String UNSUPPORTED_EXPRESSION_PATTERN = ".+\\.\\$\\{.+\\}";
```
darranl, line 147: For any regular expression IMO it is better to add a comment describing how it is structured so if anyone needs to work on it or debug it in the future they can understand the original intent of the regular expression.

lvydra:
Hi @darranl, thank you for your review. How about this?

controller/src/main/java/org/jboss/as/controller/OperationContextImpl.java
```diff
@@ -144,6 +145,10 @@
     private static final Set<Action.ActionEffect> WRITE_RUNTIME = EnumSet.of(Action.ActionEffect.WRITE_RUNTIME);
     private static final Set<Action.ActionEffect> ALL_READ_WRITE = EnumSet.of(Action.ActionEffect.READ_CONFIG, Action.ActionEffect.READ_RUNTIME, Action.ActionEffect.WRITE_CONFIG, Action.ActionEffect.WRITE_RUNTIME);
 
+    /** Pattern that can be used to identify capabilities name strings that include expression syntax
+     * started with one or more of any character, followed by '.${' sequence, then one or more of any character ended with '}' */
+    private static final Pattern UNSUPPORTED_EXPRESSION_PATTERN = Pattern.compile(".+\\.\\$\\{.+\\}");
```
bstansberry, line 150: ExpressionResolver.EXPRESSION_PATTERN exists and should be used.

controller/src/main/java/org/jboss/as/controller/logging/ControllerLogger.java
```diff
@@ -3623,4 +3623,7 @@ OperationFailedRuntimeException capabilityAlreadyRegisteredInContext(String capa
 
     @Message(id = 475, value = "Value for attribute '%s' is invalid.")
     OperationFailedException invalidAttributeValue(String attributeName);
+
+    @Message(id = 476, value = "Unsupported usage of expression:")
```
bstansberry, line 3627: I think this needs a bit more info to explain things, as now it's going to output a capability name, which the user may not understand. The name will include an expression that perhaps they will recognize.

Perhaps

"Required capabilities are not available. This likely is due to the use of an expression string in a configuration attribute that does not support expressions:"

Hmm, but thinking further... with what is in the current code we have, for example:

```
Required capabilities are not available:
     some.capability.name; Possible registration points for this capability:
          blahblahblah
     some.other.capability.name;  There are no known registration points which can provide this capability.
```

So shouldn't the result here be:

```
Required capabilities are not available:
     some.capability.name; Possible registration points for this capability:
          blahblahblah
     some.other.capability.name; There are no known registration points which can provide this capability.
     another.capability.${name}; This unresolvable capability likely is due to the use of an expression string in a configuration attribute that does not support expressions.
```

The appendPossibleProviderPoints method could handle that. If there are no possible provider points, report that, as it means something is odd beyond the use of the expression. Otherwise if there's an expression use the "does not support expressions" message. Otherwise report the provider points.

darranl:
As we have an approval from Brian going to mark this as ready to merge and see if we can include in 13.x

PR #4252, opened 2020-07-03
[WFCORE-5028] Don't make the CallbackHandler provided by the CLI preferred as it is only intended as a fallback if authentication can not be performed using the available configuration.

darranl:
https://issues.redhat.com/browse/WFCORE-5028


jmesnil:
PR for WildFly Core 12.0.x is #4253

PR #4237, opened 2020-06-17
[WFCORE-5011] Remove wildfly-jar-boot license

rgdoliveira:
artifact wildfly-jar-boot jar is not present in WildFly distribution

JIRA: https://issues.redhat.com/browse/WFCORE-5011

PR #4254, opened 2020-07-03
[WFCORE-5011] Remove wildfly-jar-boot license

jmesnil:
Artifact wildfly-jar-boot jar is not present in WildFly distribution

JIRA: https://issues.redhat.com/browse/WFCORE-5011
PR for master is #4237


kabir:
We should probably tame the dependency analyser so it doesn't go bananas on other branches than master :) @luck3y Or do the TeamCity parameters contain the name of the target branch?

luck3y:
@kabir The TC jobs have the name of the branch yeah. But if we want to not run this on branches other than master thats would be fine as well. (When the branches are created the PR jobs are generally copied as a whole.)

PR #4256, opened 2020-07-03
WFCORE-5031 Remove the audit-log deprecation

RanabirChakraborty:
Issue: https://issues.redhat.com/browse/WFCORE-5031
JBEAP Issue: https://issues.redhat.com/browse/JBEAP-19792


bstansberry:
@jmesnil For WildFly the key issue here is whether we truly intend to replace these some other resource, e.g. in a host-controller-capable subsystem.  If know we won't we should remove the deprecation. If we're not sure we should pick a time by which we must decide, e.g. by the WF 22 code freeze.

PR #4258, opened 2020-07-07
[WFCORE-5015] Create 'core-specific' variants of Galleon feature grou…

bstansberry:
…ps and packages that are 'overridden' in full WildFly

https://issues.redhat.com/browse/WFCORE-5015


jmesnil:
@bstansberry thanks

PR #4259, opened 2020-07-07
WFCORE-5029 Add propery read handler for MANAGED attirbute at server …

soul2zimate:
…group deployment level

Issue: https://issues.redhat.com/browse/WFCORE-5029
WFCORE-5029 Add propery read handler for MANAGED attribute at server group deployment level

### DIFF
--- a/host-controller/src/main/java/org/jboss/as/domain/controller/resources/DomainDeploymentResourceDefinition.java
+++ b/host-controller/src/main/java/org/jboss/as/domain/controller/resources/DomainDeploymentResourceDefinition.java
@@ -21,17 +21,24 @@
 */
 package org.jboss.as.domain.controller.resources;
 
+import static org.jboss.as.controller.descriptions.ModelDescriptionConstants.DEPLOYMENT;
+import static org.jboss.as.server.controller.resources.DeploymentAttributes.CONTENT_RESOURCE_ALL;
+import static org.jboss.as.server.controller.resources.DeploymentAttributes.isUnmanagedContent;
+
+import org.jboss.as.controller.OperationContext;
 import org.jboss.as.controller.OperationDefinition;
 import org.jboss.as.controller.OperationStepHandler;
+import org.jboss.as.controller.PathAddress;
+import org.jboss.as.controller.PathElement;
 import org.jboss.as.controller.registry.ManagementResourceRegistration;
 import org.jboss.as.controller.registry.OperationEntry.Flag;
 import org.jboss.as.domain.controller.operations.deployment.DeploymentAddHandler;
 import org.jboss.as.domain.controller.operations.deployment.DeploymentExplodeHandler;
 import org.jboss.as.domain.controller.operations.deployment.DeploymentRemoveHandler;
 import org.jboss.as.domain.controller.operations.deployment.ExplodedDeploymentAddContentHandler;
+import org.jboss.as.domain.controller.operations.deployment.ExplodedDeploymentRemoveContentHandler;
 import org.jboss.as.domain.controller.operations.deployment.ManagedDeploymentBrowseContentHandler;
 import org.jboss.as.domain.controller.operations.deployment.ManagedDeploymentReadContentHandler;
-import org.jboss.as.domain.controller.operations.deployment.ExplodedDeploymentRemoveContentHandler;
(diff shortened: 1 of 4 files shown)


host-controller/src/main/java/org/jboss/as/domain/controller/resources/DomainDeploymentResourceDefinition.java
```diff
@@ -125,4 +131,26 @@ protected void registerAddOperation(ManagementResourceRegistration registration,
...
+                 OperationStepHandler explodedDeploymentReadContentHandler,
+                 OperationStepHandler explodedDeploymentBrowseContentHandler) {
+             super(parent, addDefinition, addHandler, removeHandler, explodeDeploymentHandler,
+                     explodedDeploymentAddContentHandler, explodedDeploymentRemoveContentHandler,
+                     explodedDeploymentReadContentHandler, explodedDeploymentBrowseContentHandler);
+         }
+
+         @Override
+         public void extractedManaged(OperationContext context, ModelNode operation) {
+              String name = PathAddress.pathAddress(operation.get(ModelDescriptionConstants.OP_ADDR)).getLastElement().getValue();
```
ehsavoie, line 148: Shouldn't you use context.getCurrentAddressValue() instead ?

host-controller/src/main/java/org/jboss/as/domain/controller/resources/DomainDeploymentResourceDefinition.java
```diff
@@ -125,4 +131,26 @@ protected void registerAddOperation(ManagementResourceRegistration registration,
         registration.registerOperationHandler(addDefinition, handler);
     }
 
+    static class ServerGroupDomainDeploymentResourceDefinition extends DomainDeploymentResourceDefinition {
```
ehsavoie, line 134: Should be private

darranl: I don't think being private here should make much difference, we had this discussion with David a number of years back on Elytron.  I believe even if you set this to private the resulting class still needs to be given package level visibility so it can be accessed by the class which contains it.

Most importantly nothing here is public so nothing will leak as API used by other packages.

darranl:
This was about to be merged but removing "ready-for-merge" as changes have been requested.

host-controller/src/main/java/org/jboss/as/domain/controller/resources/DomainDeploymentResourceDefinition.java
soul2zimate: changed to private

host-controller/src/main/java/org/jboss/as/domain/controller/resources/DomainDeploymentResourceDefinition.java
soul2zimate: switched `context.getCurrentAddressValue()` instead

soul2zimate:
@ehsavoie  I have made some changes based on last review, can you check this again ?

testsuite/domain/src/test/java/org/jboss/as/test/integration/domain/management/cli/ServerGroupDeploymentManagedAttributeTestCase.java
```diff
@@ -0,0 +1,110 @@
+/*
+ * JBoss, Home of Professional Open Source.
+ * Copyright 2020, Red Hat, Inc., and individual contributors
+ * as indicated by the @author tags. See the copyright.txt file in the
+ * distribution for a full listing of individual contributors.
+ *
+ * This is free software; you can redistribute it and/or modify it
+ * under the terms of the GNU Lesser General Public License as
```
darranl, line 8: Unless this is being copied from somewhere else anything new to WildFly Core should really be contributed using an Apache license.

soul2zimate: license updated.

darranl:
I am going to include this in the next batch, the requested changes were made.

PR #4260, opened 2020-07-08
WFCORE-5037 - Handle maps of complex objects in JMX Facade

kabir:
https://issues.redhat.com/browse/WFCORE-5037

### DIFF
--- a/jmx/src/main/java/org/jboss/as/jmx/model/MBeanInfoFactory.java
+++ b/jmx/src/main/java/org/jboss/as/jmx/model/MBeanInfoFactory.java
@@ -75,7 +75,9 @@ import javax.management.openmbean.OpenMBeanParameterInfoSupport;
 import javax.management.openmbean.OpenType;
 import javax.management.openmbean.SimpleType;
 
+import org.jboss.as.controller.AttributeDefinition;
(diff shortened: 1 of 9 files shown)


darranl:
@jmesnil I think you were talking about this one last week, were there still issues you thought should be addressed?

controller/src/main/java/org/jboss/as/controller/AttributeDefinition.java
```diff
@@ -1295,6 +1295,14 @@ public ModelNode getUndefinedMetricValue() {
         return undefinedMetricValue;
     }
 
+    /**
+     * @deprecated internal use only
+     */
+    @Deprecated
+    public boolean isNestedMap() {
```
jmesnil, line 1302: As we discussed during the meeting, I don't think that method belong to the `AttributeDefinition` API. You mentioned that we can achieve the same by doing an `instanceof` check for the `ObjectMapAttributeDefinition` in https://github.com/wildfly/wildfly-core/pull/4260/files#diff-738bea9b01b7d11ea0c7968b4390eeccR299.

kabir:
@jmesnil I have removed that deprecated method and am using instanceof now

jmesnil:
@kabir thanks

PR #4263, opened 2020-07-13
[WFCORE-5047] Automatically add the -Djboss.modules.os-name property if it has not already been specified when running on RHEL 6, RHEL 7, or RHEL 8 to ensure the correct WildFly OpenSSL native library will be used

fjuma:
https://issues.redhat.com/browse/WFCORE-5047


jamezp:
I'm placing a hold on this because it won't work for servers in a domain. I think we need to discuss this though.

fjuma:
Closing this one. Will fix this directly in the WildFly OpenSSL code base instead of via the scripts.

PR #4268, opened 2020-07-15
Fix for WFCORE-5054, Allow to load bootable JAR provider during boot

jfdenise:
Issue: https://issues.redhat.com/browse/WFCORE-5054


jmesnil:
@jfdenise thanks

PR #4280, opened 2020-07-27
[WFCORE-5064] Incorrect use of KeyManagerFactory.getDefaultAlgorithm instead of TrustManagerFactory

rmartinc:
Issue: https://issues.redhat.com/browse/WFCORE-5064

Just changing to use the correct default trust-manager algorithm defined at JDK level (right now it's using the KeyManagerFactory default, the default value `SunX509` also exists for TrustManagerFactory and it's valid). I'm not adding any tests because the final https availability for the realm is tested for example in HTTPSManagementInterfaceTestCase. We can add a test similar to the steps I added to reproduce it in the JIRA if you really think it is needed.


jmesnil:
@rmartinc thanks

PR #4281, opened 2020-07-27
[WFCORE-5065] Add an optionally implemented, ServiceLoader loadable i…

bstansberry:
…nterface to do deployment transformation

https://issues.redhat.com/browse/WFCORE-5065

This has no effect unless a dist provides a module with an impl. The idea being that an EE 9 variant of full WildFly would.

### DIFF
--- a/server/src/main/java/org/jboss/as/server/deployment/transformation/DeploymentTransformer.java
+++ b/server/src/main/java/org/jboss/as/server/deployment/transformation/DeploymentTransformer.java
@@ -1,0 +1,55 @@
+/*
+ * JBoss, Home of Professional Open Source.
+ * Copyright 2020 Red Hat, Inc., and individual contributors
+ * as indicated by the @author tags.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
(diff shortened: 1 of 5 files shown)


server/src/main/java/org/jboss/as/server/deployment/transformation/DeploymentTransformer.java
```diff
@@ -0,0 +1,53 @@
...
+
+    /**
+     * Perform transformation of filesystem deployment content.
+     * @param src path of the deployment content to be transformed
+     * @param target path of either the file to which the transformed content should be written or to the directory
+     *               in which it should be written. If the latter that name of the written file will be the same as
+     *               the name of the {@code src} file.
+     * @throws IOException if a problem occurs reading or writing the content
+     */
+    void transform(Path src, Path target) throws IOException ;
```
jmesnil, line 52: Should we mention anything for in-place replacement (like it is used above in `deploymentTransformer.transform(deploymentFile.toPath(), deploymentFile.toPath())`?

jmesnil:
@bstansberry I'm fine with merging the PR as it is. My only question is about in-place replacement of files. Should we add a comment in the java doc to let the implementors know that src and target could be the same and they're the ones that have to take that into consideration.

server/src/main/java/org/jboss/as/server/deployment/transformation/DeploymentTransformer.java
bstansberry: @jmesnil I just updated the javadoc to clarify. Thanks!

jmesnil:
@bstansberry thanks

PR #4282, opened 2020-07-28
[WFCORE-5069] Automatically add the -Djboss.modules.os-name property if it has not already been specified when running on RHEL 6, RHEL 7, or RHEL 8 to ensure the correct WildFly OpenSSL native library will be used

fjuma:
https://issues.redhat.com/browse/WFCORE-5069

### DIFF
--- a/launcher/src/main/java/org/wildfly/core/launcher/AbstractCommandBuilder.java
+++ b/launcher/src/main/java/org/wildfly/core/launcher/AbstractCommandBuilder.java
@@ -29,6 +29,9 @@ import java.util.ArrayList;
 import java.util.Collection;
 import java.util.Collections;
 import java.util.List;
+import java.util.Locale;
+import java.util.regex.Matcher;
+import java.util.regex.Pattern;
 
 import org.wildfly.core.launcher.Arguments.Argument;
 import org.wildfly.core.launcher.logger.LauncherMessages;
@@ -44,6 +47,8 @@ abstract class AbstractCommandBuilder<T extends AbstractCommandBuilder<T>> imple
     static final String SECURITY_MANAGER_PROP = "java.security.manager";
     static final String[] DEFAULT_VM_ARGUMENTS;
     static final Collection<String> DEFAULT_MODULAR_VM_ARGUMENTS;
+    private static final Pattern RHEL_PATTERN = Pattern.compile(".*\\.(el[678])\\..*");
+    protected static final String JBOSS_MODULES_OS_NAME = "jboss.modules.os-name";
 
     static {
         // Default JVM parameters for all versions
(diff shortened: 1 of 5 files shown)


launcher/src/main/java/org/wildfly/core/launcher/AbstractCommandBuilder.java
```diff
@@ -587,6 +606,35 @@ protected Path normalizePath(final Path parent, final String path) {
         return serverArgs.asList();
     }
 
+    /**
+     * Returns the value of the {@code jboss.modules.os-name} property.
+     *
+     * @return the value of the {@code jboss.modules.os-name} property
+     */
+    public String getJBossModulesOSName() {
```
jmesnil, line 614: I don't understand this patch.
It seems we are duplicating the logic to determine the `jbossModulesOSName` in both the scripts and here. Am I missing a difference?

Wouldn't it be enough to have only this `getJBossModulesOSName`method that will automatically set the field if the `os.name` correspond to RHEL 6-8?

jstourac: Also again, I am a bit worried that there will be a problem with WildFly distro executed on the RHELX platform - If I didn't miss anything then with this modification we fix this for EAP, although user of WildFly will now have to set `jboss.modules.os-name` with proper value in such case, right?

fjuma: > I don't understand this patch.
> It seems we are duplicating the logic to determine the `jbossModulesOSName` in both the scripts and here. Am I missing a difference?
> 
> Wouldn't it be enough to have only this `getJBossModulesOSName`method that will automatically set the field if the `os.name` correspond to RHEL 6-8?

Initially, I only modified standalone.sh and domain.sh. @jamezp had mentioned that we'd need launcher updates as well. @jamezp Did I misunderstand what you meant?

fjuma:
Just FYI, after more discussions on this approach, am going to investigate fixing the underlying issue in JBoss Modules directly instead of here so will likely close this PR.

launcher/src/main/java/org/wildfly/core/launcher/AbstractCommandBuilder.java
jamezp: @fjuma I just meant we should probably do what we're doing in the bash script here. I don't think we need to expose it, just add it as a JAVA_OPT, e.g. `addJavaOption("-Djboss.modules.os-name=el8)`.

jamezp:
The main question I have is does the host executing `domain.sh` *always* launch other hosts on the same OS? For example can the host that executes `domain.sh` run on RHEL 8, but has another host controller on RHEL 7 it launches?

fjuma:
I've created https://issues.redhat.com/browse/MODULES-400 to track fixing the underlying issue directly in JBoss Modules instead so am going to close this PR.

PR #4283, opened 2020-07-29
[WFCORE-3818] Improved console color capability detection

lvydra:
https://issues.redhat.com/browse/WFCORE-3818

### DIFF
--- a/core-feature-pack/common/src/main/resources/content/bin/jboss-cli.sh
+++ b/core-feature-pack/common/src/main/resources/content/bin/jboss-cli.sh
@@ -5,6 +5,8 @@ GREP="grep"
 
 . "$DIRNAME/common.sh"
 
+args=("$@")
+
 # OS specific support (must be 'true' or 'false').
 cygwin=false;
 darwin=false;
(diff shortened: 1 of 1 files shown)


jmesnil:
@jfdenise could you please review?

core-feature-pack/common/src/main/resources/content/bin/jboss-cli.sh
```diff
@@ -74,6 +76,37 @@ else
     JAVA_OPTS="$JAVA_OPTS -Djboss.modules.system.pkgs=com.sun.java.swing"
 fi
 
+# Determine whether ascii coloring could be used
```
jmesnil, line 79: @lvydra how did you determine that algorithm? Is is straight from David's description in the JIRA issue?

core-feature-pack/common/src/main/resources/content/bin/jboss-cli.sh
```diff
@@ -82,8 +115,8 @@ JAVA_OPTS="$JAVA_OPTS -Dcom.ibm.jsse2.overrideDefaultTLS=true"
 
 LOG_CONF=`echo $JAVA_OPTS | grep "logging.configuration"`
 if [ "x$LOG_CONF" = "x" ]; then
-    exec "$JAVA" $JAVA_OPTS -Dlogging.configuration=file:"$JBOSS_HOME"/bin/jboss-cli-logging.properties -jar "$JBOSS_HOME"/jboss-modules.jar -mp "${JBOSS_MODULEPATH}" org.jboss.as.cli "$@"
+    exec "$JAVA" $JAVA_OPTS -Dlogging.configuration=file:"$JBOSS_HOME"/bin/jboss-cli-logging.properties -jar "$JBOSS_HOME"/jboss-modules.jar -mp "${JBOSS_MODULEPATH}" org.jboss.as.cli "${args[@]}"
```
jmesnil, line 118: could you explain why color detection requires that change in passing arguments to the java command?

lvydra: In case that ascii coloring is not supported, '--no-color-output', which is later used in CliLauncher to disable CLI output and prompt colors, is added to arguments.

core-feature-pack/common/src/main/resources/content/bin/jboss-cli.sh
lvydra: Hi @jmesnil  No, this is quite different approach. I have done some research, and this one seems to be a little bit more suitable in this context.

jfdenise:
@lvydra, thank-you. I see this change as an optimization of an existing feature that could be at risk. We have the --no-colour-output that a user can explicitly set if what he observes is not correct. We have so much OS on which this CLI script is running, that I would be concerned that this change breaks an exotic execution context.

lvydra:
Hi @jfdenise , thank you for your comment. OK, I understand, so I suppose I should close this PR and subsequently WFCORE-3818 issue, with this justification?

jfdenise:
@lvydra , the issue you are mentioning is more general than CLI, not sure you should close it.

PR #4286, opened 2020-07-30
[WFCORE-5071] Extra spaces in DeploymentTransformer.java produces error at build

rmartinc:
Issue: https://issues.redhat.com/browse/WFCORE-5071

@wildfly/mergers-full Issue with WFCORE-5065 make the build fail because of the check-style. Just removing that line.


jmesnil:
@rmartinc thanks!

PR #4290, opened 2020-07-31
[WFCORE-5073] ConsoleMode should not return true from hasConsole if t…

bstansberry:
…he module is not present

https://issues.redhat.com/browse/WFCORE-5073


jmesnil:
@bstansberry thanks

PR #4291, opened 2020-07-31
[WFCORE-5076] The "remoting-http-upgrade-service" should be ACTIVE.

darranl:
https://issues.redhat.com/browse/WFCORE-5076

This is the simplest fix for the issue, a missing dependencies error is now reported:

```
17:02:54,850 ERROR [org.jboss.as.controller.management-operation] (Controller Boot Thread) WFLYCTL0013: Operation ("add") failed - address: ([
    ("subsystem" => "remoting"),
    ("http-connector" => "http-remoting-connector")
]) - failure description: {
    "WFLYCTL0412: Required services that are not installed:" => ["org.wildfly.core.management.security.realm.OtherRealm"],
    "WFLYCTL0180: Services with missing/unavailable dependencies" => ["jboss.remoting.remoting-http-upgrade-service.http-remoting-connector is missing [org.wildfly.core.management.security.realm.OtherRealm]"]
}
```

This service does not have it's own dependants so I don't really see a justification for a resource defined in the remoting subsystem to come up if it can, if not wait for a client to connect with an obscure error message.

We could handle this with capabilities if we wanted as an alternative we have just been trying to avoid spending time converting legacy security realms to capabilities as they are deprecated but if that is seen as the more correct fix that may be required.

_Note: Once legacy security is removed the Elytron sasl-authentication-factory attribute is capability based so does not suffer the same issue here._

This PR also contains the following fixes to adjust the remoting layer(s):

https://issues.redhat.com/browse/WFCORE-5078
https://issues.redhat.com/browse/WFCORE-5082

The end result of the two layers can be seen in the following gist, in each case the gist was populated with files from a server provisioned with "web-server" only then either "remoting" or "legacy-remoting" were added so the effect of the change can be seen:

https://gist.github.com/darranl/5503feaf0ab547fdde5622bd31de8dad/revisions

### DIFF
--- a/core-feature-pack/galleon-feature-pack/src/main/resources/layers/standalone/legacy-remoting/layer-spec.xml
+++ b/core-feature-pack/galleon-feature-pack/src/main/resources/layers/standalone/legacy-remoting/layer-spec.xml
@@ -1,0 +1,9 @@
+<?xml version="1.0" ?>
+<layer-spec xmlns="urn:jboss:galleon:layer-spec:1.0" name="legacy-remoting">
+    <dependencies>
+        <layer name="io"/>
+        <layer name="application-security-realm"/>
+    </dependencies>
+
+    <feature-group name="remoting"/>
+</layer-spec>
--- a/core-feature-pack/galleon-common/src/main/resources/layers/standalone/remoting/layer-spec.xml
+++ b/core-feature-pack/galleon-common/src/main/resources/layers/standalone/remoting/layer-spec.xml
@@ -2,6 +2,14 @@
 <layer-spec xmlns="urn:jboss:galleon:layer-spec:1.0" name="remoting">
     <dependencies>
         <layer name="io"/>
+        <layer name="elytron"/>
     </dependencies>
     <feature-group name="remoting"/>
+
+    <feature spec="subsystem.remoting.http-connector">
+        <param name="http-connector" value="http-remoting-connector"/>
+        <param name="connector-ref" value="default"/>
+        <unset param="security-realm"/>
+        <param name="sasl-authentication-factory" value="application-sasl-authentication"/>
+    </feature>
 </layer-spec>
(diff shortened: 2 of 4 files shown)


jmesnil:
The [failure in the Galleon Linux - JDK8](https://ci.wildfly.org/viewLog.html?buildId=215522&buildTypeId=WildFlyCore_PullRequest_GalleonLinuxJdk8) job looks related to the PR:

```
"WFLYCTL0412: Required services that are not installed:" => ["jboss.http-upgrade-registry.default"],
"WFLYCTL0180: Services with missing/unavailable dependencies" => ["jboss.remoting.remoting-http-upgrade-service.http-remoting-connector is missing [jboss.http-upgrade-registry.default]"]
}
&amp#27;[0m&amp#27;[0m02:11:18,409 INFO  [org.jboss.as.controller] (Controller Boot Thread) WFLYCTL0183: Service status report
WFLYCTL0184:    New missing/unsatisfied dependencies:
service jboss.http-upgrade-registry.default (missing) dependents: [service jboss.remoting.remoting-http-upgrade-service.http-remoting-connector]
```

darranl:
That is good, that is the broken remoting layer I mentioned yesterday - I will add the commit to fix that to this PR.  I am hoping with a confirmed test failure we can handle that as a bug report now.

darranl:
@yersan FYI

darranl:
The remoting layer(s) should possibly be moved over to WildFly, the configuration in these layers is not usable without Undertow also being provisioned but we can't declare that dependency in WildFly Core as Undertow does not exist yet.

I have added the following to test in WildFly:
  https://github.com/wildfly/wildfly/pull/13463

Although the layer may need to move I think the subsystem can likely remain where it is - as a subsystem it can be configured in alternative ways which don't depend on Undertow.

core-feature-pack/galleon-feature-pack/src/main/resources/layers/standalone/legacy-remoting/layer-spec.xml
```diff
@@ -0,0 +1,8 @@
+<?xml version="1.0" ?>
+<layer-spec xmlns="urn:jboss:galleon:layer-spec:1.0" name="legacy-remoting">
+    <dependencies>
+        <layer name="application-security-realm"/>
```
yersan, line 4: Hi @darranl , I don't get this point, why is `io` Galleon layer is no longer required on this variant or the remoting feature-group?

darranl: That is a good point, that probably should be in there.

darranl:
@yersan Just pushed again with that update.

This time this diff on top of base-server probably gives a more accurate comparison https://gist.github.com/darranl/870199223d3f063687a97177aa0db745/revisions

bstansberry:
If we're going to add a legacy-remoting layer there should be an execution added in testsuite/layers to provision it.

darranl:
@bstansberry if we want to be completely correct with this one we probably should move the remoting layers to WildFly so they can define the dependency on Undertow but not sure if we want to go that far.

The subsystem itself I don't think needs to move.

darranl:
I have just kicked off the failed jobs again, the failures seemed CI related not PR related.

darranl:
The one still red is in the queue waiting for a turn

darranl:
Just investigating if the full job timeouts are caused by this change.

darranl:
A local allTests run just completed locally for Linux.

PR #4292, opened 2020-08-05
[WFCORE-5083] check invalid attribute in ObjectTypeValidator.

soul2zimate:
Issue: https://issues.redhat.com/browse/WFCORE-5083

Adding a global-module with an invalid attribute doesn't give an error. The input ObjectTypeAttribute parameter should have been validated in ObjectTypeValidator.

### DIFF
--- a/controller/src/main/java/org/jboss/as/controller/logging/ControllerLogger.java
+++ b/controller/src/main/java/org/jboss/as/controller/logging/ControllerLogger.java
@@ -3646,5 +3646,7 @@ public interface ControllerLogger extends BasicLogger {
     @Message(id = 481, value = "The runtime dependency package '%s' is already registered at location '%s'")
     void runtimePackageDependencyAlreadyRegistered(String pckg, String location);
 
-
+    @LogMessage(level = WARN)
+    @Message(id = 482, value = "Unknown attribute '%s' encountered. Valid attribute names are: '%s'")
+    void unknownAttribute(StringBuilder unexpectedAttr, StringBuilder possibleAttributes);
 }
--- a/controller/src/main/java/org/jboss/as/controller/operations/validation/ObjectTypeValidator.java
+++ b/controller/src/main/java/org/jboss/as/controller/operations/validation/ObjectTypeValidator.java
@@ -22,11 +22,16 @@
 
 package org.jboss.as.controller.operations.validation;
 
+import java.util.Collections;
 import java.util.HashMap;
+import java.util.HashSet;
+import java.util.Iterator;
 import java.util.Map;
+import java.util.Set;
 
 import org.jboss.as.controller.AttributeDefinition;
 import org.jboss.as.controller.OperationFailedException;
+import org.jboss.as.controller.logging.ControllerLogger;
 import org.jboss.dmr.ModelNode;
 import org.jboss.dmr.ModelType;
 
@@ -51,6 +56,8 @@ public class ObjectTypeValidator extends ModelTypeValidator {
     public void validateParameter(final String parameterName, final ModelNode value) throws OperationFailedException {
         super.validateParameter(parameterName, value);
         if (value.isDefined()) {
+            // unlike ObjectModelValue maintains a map inside, PropertyModelValue is irrelevant as it only contains one property.
+            Set<String> keys = value.getType() == ModelType.OBJECT ? new HashSet<>(value.keys()): Collections.emptySet();
             for (AttributeDefinition ad : allowedValues.values()) {
                 String key = ad.getName();
                 // Don't modify the value by calls to get(), because that's best in general.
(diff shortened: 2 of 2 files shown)


controller/src/main/java/org/jboss/as/controller/logging/ControllerLogger.java
```diff
@@ -3623,4 +3623,7 @@ OperationFailedRuntimeException capabilityAlreadyRegisteredInContext(String capa
 
     @Message(id = 475, value = "Value for attribute '%s' is invalid.")
     OperationFailedException invalidAttributeValue(String attributeName);
+
+    @Message(id = 476, value = "Invalid attribute(s) '%s' .")
+    OperationFailedException invalidAttribute(ModelNode node);
```
jmesnil, line 3628: I'm not sure that the message is meaningful enough... What makes the attribute invalid? It would be more informational to put the keys that are still present in the node...

soul2zimate: Thanks, I have added more information into the message and made it more clear.

bstansberry:
We need to think carefully about this as the server is generally meant to be forgiving of unknown data.  We don't, for example, reject on the server side unknown parameters in an 'add' op.

All the stuff we've done with domain mode transformers over the years helps remove one of the key initial needs for being forgiving. But there are other use cases; e.g. CLI scripts that get run against multiple server versions.

jmesnil:
Brian is raising a good point. Until now, the validation is only meant to validate what the server is expecting. This PR now also verifies what is unexpected and rejects the operation if any unexpected attribute is found.
However having additional ModelNode properties does not have any impact on the runtime. And this additional properties are only transient and are not persistent in the XML so they are quite harmless

I'm fine with logging a warning in that case to let the user know that something unexpected was passed to the operation.

If we want to achieve further constraints it should be done in a consistent way (e.g. the `:add` operation must also verified it) and make sure that we have a proper understanding on the implications for all our use cases (including CLI scripts, domain mode, etc.)

soul2zimate:
You're right, it's harmless. In this example, people tries to use unknown parameter for adding global modules.

If you're all agree with the warning approach, I can update this to log a warning rather than fail it with OperationFailedException

bstansberry:
If we warn it should not be a server.log warning. Client side mistakes should not end up in the server log.

Unfortunately the ParameterValidator API doesn't provide a hook for sending a warning back to the client via a response header (i.e. OperationContext.addResponseWarning

I'm going to make a couple concrete comments in case this moves forward in its current form.

controller/src/main/java/org/jboss/as/controller/logging/ControllerLogger.java
```diff
@@ -3623,4 +3623,8 @@ OperationFailedRuntimeException capabilityAlreadyRegisteredInContext(String capa
 
     @Message(id = 475, value = "Value for attribute '%s' is invalid.")
     OperationFailedException invalidAttributeValue(String attributeName);
+
+    @LogMessage(level = WARN)
+    @Message(id = 476, value = "Unknown attribute '%s' encoutered. Valid attribute names are: '%s'")
```
bstansberry, line 3628: s/encoutered/encountered

controller/src/main/java/org/jboss/as/controller/operations/validation/ObjectTypeValidator.java
```diff
@@ -51,14 +53,33 @@ public ObjectTypeValidator(final boolean nullable, final AttributeDefinition...
     public void validateParameter(final String parameterName, final ModelNode value) throws OperationFailedException {
         super.validateParameter(parameterName, value);
         if (value.isDefined()) {
+            ModelNode clonedValue = null;
```
bstansberry, line 56: This can be a Set<String>.

controller/src/main/java/org/jboss/as/controller/operations/validation/ObjectTypeValidator.java
```diff
@@ -51,14 +53,33 @@ public ObjectTypeValidator(final boolean nullable, final AttributeDefinition...
...
+            StringBuilder allowed = new StringBuilder();
+            // unlike ObjectModelValue maintains a map inside, PropertyModelValue is irrelevant as it only contains one property.
+            if (value.getType() == ModelType.OBJECT) {
+                clonedValue = value.clone();
```
bstansberry, line 60: then this is

clonedValue = value.keys();

controller/src/main/java/org/jboss/as/controller/operations/validation/ObjectTypeValidator.java
```diff
@@ -51,14 +53,33 @@ public ObjectTypeValidator(final boolean nullable, final AttributeDefinition...
     public void validateParameter(final String parameterName, final ModelNode value) throws OperationFailedException {
         super.validateParameter(parameterName, value);
         if (value.isDefined()) {
+            ModelNode clonedValue = null;
+            StringBuilder allowed = new StringBuilder();
```
bstansberry, line 57: I'd prefer this isn't declared until the https://github.com/wildfly/wildfly-core/pull/4292/files#diff-377f68b7791de46e877e8bc15afc390eR75 failure handling block is entered. Then in that block loop through the allowedValues.keys().

IOW the failure case will be an edge case so for the normal case we should do the minimal work related to it, even if that means a bit more work in the failure case.

darranl:
@soul2zimate FYI this one currently has a merge conflict.

soul2zimate:
PR rebased

darranl:
Thanks @soul2zimate are you happy the other comments are handled from Brian and Jeff to request another review from them?

bstansberry:
My only concern is this logs client mistakes in the server log instead of sending the warn to the client. I don't see how to avoid that though, at least not without a far bigger change.

darranl:
Sorry @soul2zimate but looks like another merge conflict in this PR.

PR #4300, opened 2020-08-12
[WFLY-10879] Adding new EE deployment phase constant

spyrkob:
Issue: https://issues.redhat.com/browse/WFLY-10879 

This adds a new phase constant required by wildfly/wildfly#11711.

Is it OK to use the WFLY issue for this or should I create a separate WFCORE ticket?

### DIFF
--- a/server/src/main/java/org/jboss/as/server/deployment/Phase.java
+++ b/server/src/main/java/org/jboss/as/server/deployment/Phase.java
@@ -613,6 +613,7 @@ public enum Phase {
     public static final int POST_MODULE_UNDERTOW_SERVLET_CONTAINER_DEPENDENCY       = 0x2F1A;
 
     public static final int POST_MODULE_EE_CONCURRENT_CONTEXT           = 0x3000;
+    public static final int POST_MODULE_EE_STARTUP_COUNTDOWN            = 0x3080;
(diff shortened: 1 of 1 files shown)


server/src/main/java/org/jboss/as/server/deployment/Phase.java
```diff
@@ -613,6 +613,7 @@ public Phase next() {
     public static final int POST_MODULE_UNDERTOW_SERVLET_CONTAINER_DEPENDENCY       = 0x2F1A;
 
     public static final int POST_MODULE_EE_CONCURRENT_CONTEXT           = 0x3000;
+    public static final int POST_MODULE_EE_STARTUP_COUNTDOWN            = 0x3001;
```
bstansberry, line 616: @spyrkob Does this need to immediately follow POST_MODULE_EE_CONCURRENT_CONTEXT or just be after and before POST_MODULE_BATCH_ENVIRONMENT?

If the latter, 0x3080 is better.

spyrkob: Updated to 0x3080. 

Timing-wise it just need to happen after POST_MODULE_EE_CONCURRENT_CONTEXT and before INSTALL phase

PR #4302, opened 2020-08-13
[WFCORE-5089] FailedOperationTransformationConfig: correct all attrib…

tadamski:
https://issues.redhat.com/browse/WFCORE-5089

### DIFF
--- a/model-test/src/main/java/org/jboss/as/model/test/FailedOperationTransformationConfig.java
+++ b/model-test/src/main/java/org/jboss/as/model/test/FailedOperationTransformationConfig.java
@@ -425,6 +425,7 @@ public class FailedOperationTransformationConfig {
         @Override
         public ModelNode correctOperation(ModelNode operation) {
             ModelNode op = operation.clone();
+            boolean corrected = false;
             for (String attr : attributes) {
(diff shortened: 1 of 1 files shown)


model-test/src/main/java/org/jboss/as/model/test/FailedOperationTransformationConfig.java
```diff
@@ -437,10 +437,9 @@ public ModelNode correctOperation(ModelNode operation) {
                     } else {
                         op.get(attr).set(complexChildConfig.correctOperation(operation.get(attr)));
                     }
-                    return op;
                 }
             }
-            return operation;
+            return op;
```
bstansberry, line 442: It might be safer to track whether 'op' has been modified (ie the 'if (checkValue...' block was entered) and return 'operation' if not.

The op.get call at L429 may modify 'op' without really changing its semantic content and then there is other code that checks if the return val from this method equals what was passed in. Having a different result from that check when there was no actual change might mess something up.

tadamski: fixed

PR #4307, opened 2020-08-21
[WFCORE-5100] Add missing distribution repo in core-feature-pack-common-licenses.xml license file

rgdoliveira:
Issue: https://issues.redhat.com/browse/WFCORE-5100


jamezp:
This doesn't really need to run through CI, but I want to add Roberto to the whilelist so this is ok to test

PR #4310, opened 2020-08-30
[WFCORE-5107 Provide an equivalent to RemotingConnectorInfo via org.j…

bstansberry:
…boss.as.network module; expose it via "org.wildfly.remoting.connector" capability

https://issues.redhat.com/browse/WFCORE-5107


jmesnil (approved):
@bstansberry Could you please also update https://github.com/wildfly/wildfly-capabilities/blob/master/org/wildfly/remoting/connector/capability.adoc to add information about the `ProtocolSocketBinding` interface?

PR #4311, opened 2020-08-31
[WFCORE-5101] [Community] SSLContext to support delegation to alternate instances based on peer information.

Skyllarr:
https://issues.redhat.com/browse/WFCORE-5101

Requires: https://issues.redhat.com/browse/ELY-1996 which has PR here: https://github.com/wildfly-security/wildfly-elytron/pull/1382


elytron/src/main/java/org/wildfly/extension/elytron/AuthenticationClientDefinitions.java
```diff
@@ -505,6 +507,18 @@ static ResourceDefinition getAuthenticationContextDefinition() {
...
+            @Override
+            protected void performRuntime(OperationContext context, ModelNode operation, ModelNode model) throws OperationFailedException {
+                commonRequirements(installService(context, model)).setInitialMode(ServiceController.Mode.ACTIVE).install();
+            }
+
+            ServiceBuilder<AuthenticationContext> installService(OperationContext context, ModelNode model)  throws OperationFailedException {
+                ServiceTarget serviceTarget = context.getServiceTarget();
+                ServiceBuilder<AuthenticationContext> serviceBuilder =
+                        (ServiceBuilder<AuthenticationContext>)context.getCapabilityServiceTarget().addCapability(AUTHENTICATION_CONTEXT_RUNTIME_CAPABILITY);
+                TrivialService<AuthenticationContext> certificateAuthorityTrivialService = new TrivialService<AuthenticationContext>(getValueSupplier(serviceBuilder, context, model));
```
fjuma, line 519: Very minor but looks like `certificateAuthorityTrivialService` should be renamed.

Skyllarr: Fixed. Thanks!

jmesnil:
this PR is out of sync with the main branch.
I'm closing it. Please reopen it when it is up to date and can be built.

darranl:
Converted to draft whilst awaiting pre-check

darranl:
@Skyllarr FYI looks like we have a merge conflict here.

Also in general this PR is still active we are just awaiting approval to proceed with the merge,

elytron/src/main/resources/org/wildfly/extension/elytron/LocalDescriptions.properties
```diff
@@ -1384,6 +1384,15 @@ elytron.client-ssl-context.ssl-session.peer-certificates.version=The certificate
 # Operations
 elytron.client-ssl-context.ssl-session.invalidate=Invalidate the SSLSession (Note: This does not terminate current connections, only prevents future connections from joining or resuming this session).
 
+elytron.dynamic-client-ssl-context=An Dynamic SSLContext for use on the client side of a connection. It allows to choose ssl context based on peer's host and port information
```
fjuma, line 1387: s/An Dynamic/A dynamic

elytron/src/main/resources/org/wildfly/extension/elytron/LocalDescriptions.properties
```diff
@@ -1384,6 +1384,15 @@ elytron.client-ssl-context.ssl-session.peer-certificates.version=The certificate
...
+# operations
+elytron.dynamic-client-ssl-context.add=Add the Dynamic Client SSLContext definition.
+elytron.dynamic-client-ssl-context.remove=Remove the Dynamic Client SSLContext definition.
+#Attributes
+elytron.dynamic-client-ssl-context.authentication-context=The authentication context that will be queried for ssl context based on peer information.
+elytron.dynamic-client-ssl-context.active-session-count=The count of current active sessions.
```
fjuma, line 1393: Since `hasActiveSessions` always returns `false`, is `active-session-count` always going to be 0? If so, it might be worth pointing that out here since something similar has been pointed out for `ssl-session` below.

elytron/src/main/resources/org/wildfly/extension/elytron/LocalDescriptions.properties
```diff
@@ -1384,6 +1384,15 @@ elytron.client-ssl-context.ssl-session.peer-certificates.version=The certificate
...
+elytron.dynamic-client-ssl-context.ssl-session=The ssl session of dynamic ssl context is always undefined.
```
fjuma, line 1394: s/ssl/SSL

elytron/src/main/resources/schema/wildfly-elytron_14_0.xsd
```diff
@@ -6104,6 +6105,39 @@
...
+        </xs:annotation>
+        <xs:sequence>
+            <xs:element name="dynamic-client-ssl-context" type="dynamicClientSSLContextType" maxOccurs="unbounded"/>
+        </xs:sequence>
+    </xs:complexType>
+
+    <xs:complexType name="dynamicClientSSLContextType">
+        <xs:annotation>
+            <xs:documentation>
+                Definitions of a single client side dynamic SSL Context. This context chooses ssl context based on peer's host and port information.
```
fjuma, line 6122: s/ssl/SSL

elytron/src/main/resources/schema/wildfly-elytron_14_0.xsd
```diff
@@ -6104,6 +6105,39 @@
...
+            <xs:annotation>
+                <xs:documentation>
+                    The unique name of this client side dynamic SSL Context.
+                </xs:documentation>
+            </xs:annotation>
+        </xs:attribute>
+        <xs:attribute name="authentication-context" type="xs:string" use="required">
+            <xs:annotation>
+                <xs:documentation>
+                    The authentication context that will be used to query for rules when deciding which ssl context to use when connecting to a peer.
```
fjuma, line 6135: s/ssl/SSL

elytron/src/main/java/org/wildfly/extension/elytron/SSLDefinitions.java
```diff
@@ -156,6 +164,12 @@
             .setRestartAllServices()
             .build();
 
+    static final SimpleAttributeDefinition AUTHENTICATION_CONTEXT_ATTRIBUTE = new SimpleAttributeDefinitionBuilder(ElytronDescriptionConstants.AUTHENTICATION_CONTEXT, ModelType.STRING, true)
```
fjuma, line 167: Since `authentication-context` is a required attribute, the last parameter here should be `false` (since this attribute isn't optional).

elytron/src/test/java/org/wildfly/extension/elytron/SubsystemTransformerTestCase.java
```diff
@@ -208,6 +210,10 @@ public void testRejectingTransformersEAP730() throws Exception {
                     new FailedOperationTransformationConfig.NewAttributesConfig(SSLDefinitions.CIPHER_SUITE_NAMES))
             .addFailedAttribute(SUBSYSTEM_ADDRESS.append(PathElement.pathElement(ElytronDescriptionConstants.EXPRESSION, ElytronDescriptionConstants.ENCRYPTION)),
                     FailedOperationTransformationConfig.REJECTED_RESOURCE)
+            .addFailedAttribute(SUBSYSTEM_ADDRESS.append(PathElement.pathElement(ElytronDescriptionConstants.REGEX_ROLE_MAPPER, "RegexRoleMapper")),
+                    FailedOperationTransformationConfig.REJECTED_RESOURCE)
+            .addFailedAttribute(SUBSYSTEM_ADDRESS.append(PathElement.pathElement(DISTRIBUTED_REALM, "DistributedRealm")), REJECTED_RESOURCE)
+            .addFailedAttribute(SUBSYSTEM_ADDRESS.append(PathElement.pathElement(FAILOVER_REALM, "FailoverRealm")), REJECTED_RESOURCE)
```
fjuma, line 216: Just a minor comment but these shouldn't be here.

elytron/src/main/java/org/wildfly/extension/elytron/TlsParser.java
```diff
@@ -332,5 +339,6 @@ public void marshallSingleElement(AttributeDefinition attribute, ModelNode mappi
             .addChild(certificateAuthorityParser)
             .addChild(certificateAuthorityAccountParser)
             .addChild(serverSslSniContextParser)
+            .addChild(dynamicClientSslContextParser) // new
```
fjuma, line 342: A new `tlsParser` variable needs to be added and then `dynamicClientSslContextParser` should be added to the new variable instead of here.

fjuma (changes requested):

elytron/src/main/resources/org/wildfly/extension/elytron/LocalDescriptions.properties
Skyllarr: @fjuma active-session-count is determined by calling of the methods sslContext.getServerSessionContext() and sslContext.getClientSessionContext()  if serverOrHostController is true. 

But since these methods are throwing an exception for dynamic SSL context , I changed serverOrHostController to be always false and removed active session count from LocalDescriptions.properties.

elytron/src/main/resources/org/wildfly/extension/elytron/LocalDescriptions.properties
Skyllarr: fixed

elytron/src/main/resources/org/wildfly/extension/elytron/LocalDescriptions.properties
Skyllarr: fixed

elytron/src/main/resources/schema/wildfly-elytron_14_0.xsd
Skyllarr: fixed

elytron/src/main/resources/schema/wildfly-elytron_14_0.xsd
Skyllarr: fixed

elytron/src/main/java/org/wildfly/extension/elytron/SSLDefinitions.java
Skyllarr: fixed

elytron/src/main/java/org/wildfly/extension/elytron/TlsParser.java
Skyllarr: fixed

darranl:
This PR is still active, we are just waiting for the process to progress.

Skyllarr:
This PR is still active, we are just waiting for the process to progress.

Skyllarr:
This PR is still active, we are just waiting for the process to progress.

Skyllarr:
This PR is still active, we are just waiting for the process to progress.

Skyllarr:
This PR is still active, we are just waiting for the process to progress.

fjuma:
This PR is still active.

Skyllarr:
This PR is still active.

Skyllarr:
Just FYI, this is an RFE and this PR is still active.

Skyllarr:
Just FYI, this is an RFE and this PR is still active.

Skyllarr:
This is an RFE and this PR is still active.

Skyllarr:
This is an RFE and this PR is still active.

Skyllarr:
This is an RFE and the PR is still active

Skyllarr:
This is an RFE and the PR is still active

Skyllarr:
This is an RFE and the PR is still active

Skyllarr:
@pferraro I have parametrized elytron subsystem namespaces based on your comment [here](https://github.com/wildfly/wildfly/pull/17018#discussion_r1503298345) and added a new feature with the COMMUNITY stability level. Please review. Thank you very much!

elytron/src/main/java/org/wildfly/extension/elytron/ElytronSubsystemSchema.java
```diff
@@ -0,0 +1,102 @@
...
+        this.namespace = SubsystemSchema.createSubsystemURN(ElytronExtension.SUBSYSTEM_NAME, stability, new IntVersion(major));
+    }
+
+    @Override
+    public VersionedNamespace<IntVersion, ElytronSubsystemSchema> getNamespace() {
+        return this.namespace;
+    }
+
+    @Override
+    public PersistentResourceXMLDescription getXMLDescription() {
```
pferraro, line 73: This doesn't look right.  Why does every version use the same PersistentResourceXMLDescription?

elytron/src/main/java/org/wildfly/extension/elytron/ElytronSubsystemParserCommunity18_0.java
```diff
@@ -0,0 +1,20 @@
+/*
+ * Copyright The WildFly Authors
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+package org.wildfly.extension.elytron;
+
+import org.jboss.as.controller.PersistentResourceXMLDescription;
+
+public class ElytronSubsystemParserCommunity18_0 extends ElytronSubsystemParser18_0 {
```
pferraro, line 10: I strongly discourage the use of polymorphism to model versioning.

elytron/src/main/java/org/wildfly/extension/elytron/SSLDefinitions.java
```diff
@@ -1511,6 +1532,56 @@ protected void installedForResource(ServiceController<SSLContext> serviceControl
...
+                ServiceController<?> dynamicSSLContextService = serviceRegistry.getRequiredService(dynamicSSLContextName);
+                State dynamicSSLContextServiceState = dynamicSSLContextService.getState();
+                if (!dynamicSSLContextServiceState.equals(State.UP)) {
+                    dynamicSSLContextService.setMode(ServiceController.Mode.ACTIVE);
+                }
+                try {
+                    return new DynamicSSLContext(new DynamicSSLContextImpl(getAuthenticationContextService(serviceRegistry, authenticationContextName).getValue()));
+                } catch (GeneralSecurityException | DynamicSSLContextException e) {
+                    throw new StartException(e);
+                }
```
pferraro, line 1573: This manner of obtaining the value provided by a service is not thread safe.

elytron/src/main/java/org/wildfly/extension/elytron/ElytronExtension.java
```diff
@@ -145,6 +147,7 @@ public void initializeParsers(ExtensionParsingContext context) {
         context.setSubsystemXmlMapping(SUBSYSTEM_NAME, NAMESPACE_16_0, () -> new ElytronSubsystemParser16_0());
         context.setSubsystemXmlMapping(SUBSYSTEM_NAME, NAMESPACE_17_0, () -> new ElytronSubsystemParser17_0());
         context.setSubsystemXmlMapping(SUBSYSTEM_NAME, NAMESPACE_18_0, () -> new ElytronSubsystemParser18_0());
+        context.setSubsystemXmlMapping(SUBSYSTEM_NAME, NAMESPACE_COMMUNITY_18_0, () -> new ElytronSubsystemParserCommunity18_0());
```
pferraro, line 150: If your ElytronSubsystemSchema enumeration implements getXMLDescription() correctly, then schema registration for all versions becomes a one-liner, e.g.

```java
context.setSubsystemXMLMappings(SUBSYSTEM_NAME, EnumSet.allOf(ElytronSubsystemSchema.class));
```

elytron/src/main/java/org/wildfly/extension/elytron/TrivialResourceDefinition.java
```diff
@@ -173,6 +182,11 @@ Builder setRuntimeCapabilities(RuntimeCapability<?> ... runtimeCapabilities) {
             return this;
         }
 
+        Builder setStability(Stability stability) {
+            this.stability = stability;
+            return this;
+        }
+
```
pferraro, line 189: This seems redundant.  The super implementation already defines a stability field.

elytron/src/main/java/org/wildfly/extension/elytron/TrivialResourceDefinition.java
```diff
@@ -37,10 +38,17 @@ final class TrivialResourceDefinition extends SimpleResourceDefinition {
...
-            RuntimeCapability<?>[] runtimeCapabilities) {
+                                      Map<AttributeDefinition, OperationStepHandler> readOnlyAttributes, Map<OperationDefinition, OperationStepHandler> operations, List<ResourceDefinition> children,
+                                      RuntimeCapability<?>[] runtimeCapabilities) {
+
+        this(pathKey, resourceDescriptionResolver, add, remove, attributes, readOnlyAttributes, operations, children, runtimeCapabilities, Stability.DEFAULT);
+    }
+
+    private TrivialResourceDefinition(String pathKey, ResourceDescriptionResolver resourceDescriptionResolver, AbstractAddStepHandler add, AbstractRemoveStepHandler remove, AttributeDefinition[] attributes,
+                                      Map<AttributeDefinition, OperationStepHandler> readOnlyAttributes, Map<OperationDefinition, OperationStepHandler> operations, List<ResourceDefinition> children,
+                                      RuntimeCapability<?>[] runtimeCapabilities, Stability stability) {
```
pferraro, line 49: The above constructor is private, and the changes below mean that it will never be called.  Why not just change it rather than create a new one?

elytron/src/main/java/org/wildfly/extension/elytron/TrivialResourceDefinition.java
```diff
@@ -37,10 +38,17 @@ final class TrivialResourceDefinition extends SimpleResourceDefinition {
...
         super(new Parameters(PathElement.pathElement(pathKey),
-                resourceDescriptionResolver)
+                resourceDescriptionResolver, stability)
```
pferraro, line 51: I'd prefer that you use the existing super constructor rather than create a redundant one.

elytron/src/main/java/org/wildfly/extension/elytron/ElytronExtension.java
Skyllarr: @pferraro Thank you very much for quick review! It all sounds good! I am working on all these changes now

elytron/src/main/java/org/wildfly/extension/elytron/SSLDefinitions.java
```diff
@@ -1511,6 +1532,56 @@ protected void installedForResource(ServiceController<SSLContext> serviceControl
...
 
+    private static final AttributeDefinition[] DYNAMIC_SSL_CONTEXT_ATTRIBUTES = new AttributeDefinition[]{AUTHENTICATION_CONTEXT_ATTRIBUTE};
+
+    static ResourceDefinition getDynamicClientSSLContextDefinition(boolean serverOrHostController) {
+        return createSSLContextDefinition(ElytronDescriptionConstants.DYNAMIC_CLIENT_SSL_CONTEXT, false, new DynamicSSLContextAddHandler(), DYNAMIC_SSL_CONTEXT_ATTRIBUTES, serverOrHostController, Stability.COMMUNITY);
+    }
+
+    private static class DynamicSSLContextAddHandler extends BaseAddHandler {
+        private DynamicSSLContextAddHandler() {
+            super(SSL_CONTEXT_RUNTIME_CAPABILITY, DYNAMIC_SSL_CONTEXT_ATTRIBUTES);
```
pferraro, line 1543: FYI, this BaseAddHandler super constructor (all of them, actually) relies on a deprecated (for removal) AbstractAddStepHandler constructor.

elytron/src/main/java/org/wildfly/extension/elytron/SSLDefinitions.java
```diff
@@ -1511,6 +1532,56 @@ protected void installedForResource(ServiceController<SSLContext> serviceControl
...
+        }
+
+        @Override
+        protected void performRuntime(OperationContext context, ModelNode operation, ModelNode model) throws OperationFailedException {
+            ServiceBuilder<SSLContext> serviceBuilder = installService(context, model);
+            ServiceName authenticationContextServiceName =
+                    AUTHENTICATION_CONTEXT_RUNTIME_CAPABILITY.getCapabilityServiceName(AUTHENTICATION_CONTEXT_ATTRIBUTE.resolveModelAttribute(context, model).asString());
+            serviceBuilder.requires(authenticationContextServiceName);
+            commonRequirements(serviceBuilder).setInitialMode(ServiceController.Mode.ACTIVE).install();
```
pferraro, line 1552: Why must this service start ACTIVEly?

elytron/src/main/java/org/wildfly/extension/elytron/ElytronSubsystemSchema.java
Skyllarr: @pferraro Thanks! I fixed this and used a new ElytronPersistentResourceXMLDescriptionFactory class here

elytron/src/main/java/org/wildfly/extension/elytron/ElytronSubsystemParserCommunity18_0.java
Skyllarr: @pferraro I removed the polymorphism from versioning,  Thank you!

elytron/src/main/java/org/wildfly/extension/elytron/ElytronExtension.java
Skyllarr: @pferraro Thanks! I now use `context.setSubsystemXmlMappings(SUBSYSTEM_NAME, EnumSet.allOf(ElytronSubsystemSchema.class));` as you suggested

elytron/src/main/java/org/wildfly/extension/elytron/TrivialResourceDefinition.java
Skyllarr: @pferraro you're right, I changed the existing private constructor instead of creating new one

elytron/src/main/java/org/wildfly/extension/elytron/TrivialResourceDefinition.java
Skyllarr: @pferraro I removed this change and Parameters constructors now stay the same. Thanks!

elytron/src/main/java/org/wildfly/extension/elytron/SSLDefinitions.java
Skyllarr: @pferraro True, I have changed this to be `ON_DEMAND` instead

elytron/src/main/java/org/wildfly/extension/elytron/SSLDefinitions.java
Skyllarr: @pferraro Good point, since we use it on many places and it is not related to this issue only, I have created jira https://issues.redhat.com/browse/WFCORE-6736 to address this

elytron/src/main/java/org/wildfly/extension/elytron/SSLDefinitions.java
Skyllarr: @pferraro Can you please elaborate on what is not thread safe in this method? I think we obtain a value from a service like this on other places also and I am not sure what can cause issues. What should be a correct way to do this?

elytron/src/main/java/org/wildfly/extension/elytron/SSLDefinitions.java
pferraro: Perfect, thanks.

elytron/src/main/java/org/wildfly/extension/elytron/SSLDefinitions.java
pferraro: > Can you please elaborate on what is not thread safe in this method? 

* Testing that the state of the ServiceController is UP does not guarantee that the service is still UP when you call Service.getValue().
* ServiceController.setMode(...) is not a blocking operation, i.e. there are no guarantees that the service will be UP when this call returns.
* Service.getValue() was deprecated several years ago to discourage use for this reason.

> I think we obtain a value from a service like this on other places

If that is the case, those other places are problematic too.

Skyllarr: @pferraro Thank you that makes sense. I have changed the code now and it is no longer thread unsafe. 

I have addressed all feedback now, so this can be reviewed. Thank you!

Skyllarr:
@pferraro@fjuma @darranl Just btw, I addressed all feedback on this PR so it can be reviewed again when you'll have time. Just btw, this also has tests in wildfly repo https://github.com/wildfly/wildfly/pull/13687 . I've run full testsuites for wildfly-core and wildfly and those are passing. Thanks!

elytron/src/main/java/org/wildfly/extension/elytron/JaspiConfigurationParser.java
```diff
@@ -0,0 +1,20 @@
...
+
+package org.wildfly.extension.elytron;
+
+import org.jboss.as.controller.PathElement;
+import org.jboss.as.controller.PersistentResourceXMLDescription;
+
+import static org.wildfly.extension.elytron.ElytronDescriptionConstants.JASPI;
+import static org.wildfly.extension.elytron.ElytronDescriptionConstants.JASPI_CONFIGURATION;
+
+public class JaspiConfigurationParser {
```
darranl, line 14: Does this need to be public?

elytron/src/main/java/org/wildfly/extension/elytron/ElytronDefinition.java
```diff
@@ -294,6 +294,10 @@ public void registerChildren(ManagementResourceRegistration resourceRegistration
         resourceRegistration.registerSubModel(SSLDefinitions.getServerSNISSLContextDefinition());
         resourceRegistration.registerSubModel(new CertificateAuthorityDefinition());
         resourceRegistration.registerSubModel(new CertificateAuthorityAccountDefinition());
+        ResourceDefinition dynamicClientSSLContextDefinition = SSLDefinitions.getDynamicClientSSLContextDefinition(false);
+        if (resourceRegistration.enables(dynamicClientSSLContextDefinition)) {
+            resourceRegistration.registerSubModel(dynamicClientSSLContextDefinition);
+        }
```
pferraro, line 300: This is redundant.  The MRR already filters registration based on stability (do you don't have to).

pferraro (changes requested):

elytron/src/main/java/org/wildfly/extension/elytron/TrivialResourceDefinition.java
```diff
@@ -36,9 +37,11 @@ final class TrivialResourceDefinition extends SimpleResourceDefinition {
     private final Map<AttributeDefinition, OperationStepHandler> readOnlyAttributes;
     private final List<ResourceDefinition> children;
 
+    private final Stability stability;
```
pferraro, line 40: The SimpleResourceDefinition parent class already has a stability field.

pferraro (changes requested):

elytron/src/main/java/org/wildfly/extension/elytron/AuthenticationClientDefinitions.java
```diff
@@ -491,6 +494,18 @@ protected ValueSupplier<AuthenticationContext> getValueSupplier(ServiceBuilder<A
                 return () -> finalContext.apply(parentSupplier.get());
             }
 
+            @Override
+            protected void performRuntime(OperationContext context, ModelNode operation, ModelNode model) throws OperationFailedException {
+                commonRequirements(installService(context, model)).setInitialMode(ServiceController.Mode.ACTIVE).install();
```
pferraro, line 499: Does this service need to start eagerly?

elytron/src/main/java/org/wildfly/extension/elytron/AuthenticationClientDefinitions.java
```diff
@@ -491,6 +494,18 @@ protected ValueSupplier<AuthenticationContext> getValueSupplier(ServiceBuilder<A
...
+            }
+
+            ServiceBuilder<AuthenticationContext> installService(OperationContext context, ModelNode model)  throws OperationFailedException {
+                ServiceTarget serviceTarget = context.getServiceTarget();
+                ServiceBuilder<AuthenticationContext> serviceBuilder =
+                        (ServiceBuilder<AuthenticationContext>)context.getCapabilityServiceTarget().addCapability(AUTHENTICATION_CONTEXT_RUNTIME_CAPABILITY);
```
pferraro, line 505: Is this cast necessary?

elytron/src/main/java/org/wildfly/extension/elytron/TlsParser.java
```diff
@@ -349,4 +355,20 @@ public void marshallSingleElement(AttributeDefinition attribute, ModelNode mappi
...
+            )
+            .addChild(keyManagerParser_12_0)
+            .addChild(trustManagerParser_14_0)
+            .addChild(serverSslContextParser_9_0)
+            .addChild(clientSslContextParser_9_0)
+            .addChild(certificateAuthorityParser)
+            .addChild(certificateAuthorityAccountParser)
+            .addChild(serverSslSniContextParser)
+            .addChild(dynamicClientSslContextParser) // new
+            .build();
```
pferraro, line 373: Why not parameterize your PersistentResourceXMLDescription builder rather than duplicate this?

elytron/src/main/java/org/wildfly/extension/elytron/ElytronPersistentResourceXMLDescriptionFactory.java
```diff
@@ -0,0 +1,230 @@
...
+import static org.wildfly.extension.elytron.ElytronDescriptionConstants.DIR_CONTEXTS;
+import static org.wildfly.extension.elytron.ElytronDescriptionConstants.ENCRYPTION;
+import static org.wildfly.extension.elytron.ElytronDescriptionConstants.EXPRESSION;
+import static org.wildfly.extension.elytron.ElytronDescriptionConstants.EXPRESSION_RESOLVER;
+import static org.wildfly.extension.elytron.ElytronDescriptionConstants.PERMISSION_SETS;
+import static org.wildfly.extension.elytron.ElytronDescriptionConstants.SECURITY_PROPERTY;
+import static org.wildfly.extension.elytron.PermissionMapperDefinitions.PERMISSIONS;
+
+public enum ElytronPersistentResourceXMLDescriptionFactory implements Function<ElytronSubsystemSchema, PersistentResourceXMLDescription> {
+    INSTANCE;
```
fjuma, line 25: Should this be `static final ElytronPersistentResourceXMLDescriptionFactory INSTANCE = new ElytronPersistentResourceXMLDescriptionFactory();`?

elytron/src/main/java/org/wildfly/extension/elytron/ElytronPersistentResourceXMLDescriptionFactory.java
```diff
@@ -0,0 +1,230 @@
...
+        }
+
+        if (schema.since(ElytronSubsystemSchema.VERSION_5_0)) {
+            builder.addAttribute(ElytronDefinition.REGISTER_JASPI_FACTORY)
+                    .addAttribute(ElytronDefinition.DEFAULT_SSL_CONTEXT);
+        }
+
+        addAuthenticationClientParser(builder, schema);
+
+        if (schema.since(ElytronSubsystemSchema.VERSION_1_0)) {
```
fjuma, line 45: Just from a readability perspective, might be good to get this into an `addProviderParser` method like you have for the other cases so that in the future, any additional changes could be made in that method.

elytron/src/main/java/org/wildfly/extension/elytron/ElytronPersistentResourceXMLDescriptionFactory.java
```diff
@@ -0,0 +1,230 @@
...
+import static org.jboss.as.controller.PersistentResourceXMLDescription.builder;
+import static org.wildfly.extension.elytron.ElytronDescriptionConstants.DIR_CONTEXTS;
+import static org.wildfly.extension.elytron.ElytronDescriptionConstants.ENCRYPTION;
+import static org.wildfly.extension.elytron.ElytronDescriptionConstants.EXPRESSION;
+import static org.wildfly.extension.elytron.ElytronDescriptionConstants.EXPRESSION_RESOLVER;
+import static org.wildfly.extension.elytron.ElytronDescriptionConstants.PERMISSION_SETS;
+import static org.wildfly.extension.elytron.ElytronDescriptionConstants.SECURITY_PROPERTY;
+import static org.wildfly.extension.elytron.PermissionMapperDefinitions.PERMISSIONS;
+
+public enum ElytronPersistentResourceXMLDescriptionFactory implements Function<ElytronSubsystemSchema, PersistentResourceXMLDescription> {
```
pferraro, line 24: I would suggest moving this logic (and all other PersistentResourceXMLDescription logic) to your ElytronSubsystemSchema.getXMLDescription() method directly.  That way you can build your PersistentResourceXMLDescription dynamically without having to pass around the schema.

elytron/src/main/java/org/wildfly/extension/elytron/ElytronPersistentResourceXMLDescriptionFactory.java
```diff
@@ -0,0 +1,230 @@
...
+        if (schema.since(ElytronSubsystemSchema.VERSION_13_0)) {
+            builder.addChild(credentialStoreParser.getCredentialStoresParser_13().build());
+        } else if (schema.since(ElytronSubsystemSchema.VERSION_1_0)) {
+            builder.addChild(credentialStoreParser.getCredentialStoresParser().build());
+        }
+    }
+
+    private static void addTlsParser(ElytronSubsystemSchema schema, PersistentResourceXMLDescription.PersistentResourceXMLBuilder builder) {
+        TlsParser tlsParser = new TlsParser();
+        if (schema.since(ElytronSubsystemSchema.VERSION_18_0_COMMUNITY)) {
```
pferraro, line 110: N.B. ElytronSubsystemSchema.since(...) only compares version numbers.  If a resource needs to be conditionally added based on a specific min version and stability, you would use something like:

```java
if (this.since(VERSION_X_Y) && this.enables(ResourceRegistration.FOO)) {
   // This community-level resource was added at VERSION_X_Y
   builder.addChild(ResourceRegistration.FOO);
}
```

Otherwise, the parser will attempt to persist this resource in the next version regardless of stability level.

pferraro, line 110: N.B. ElytronSubsystemSchema.since(...) only compares version numbers.  If a resource needs to be conditionally added based on a specific min version and stability, you would use something like:

```java
if (this.since(VERSION_X_Y) && this.enables(ResourceRegistration.FOO)) {
   // This community-level resource was added at VERSION_X_Y
   builder.addChild(ResourceRegistration.FOO);
}
```

Otherwise, the parser will attempt to persist this resource in the next version regardless of stability level.

elytron/src/main/java/org/wildfly/extension/elytron/ElytronPersistentResourceXMLDescriptionFactory.java
```diff
@@ -0,0 +1,230 @@
...
+            builder.addChild(tlsParser.tlsParser_9_0);
+        } else if (schema.since(ElytronSubsystemSchema.VERSION_8_0)) {
+            builder.addChild(tlsParser.tlsParser_8_0);
+        } else if (schema.since(ElytronSubsystemSchema.VERSION_5_0)) {
+            builder.addChild(tlsParser.tlsParser_5_0);
+        } else if (schema.since(ElytronSubsystemSchema.VERSION_4_0)) {
+            builder.addChild(tlsParser.tlsParser_4_0);
+        } else if (schema.since(ElytronSubsystemSchema.VERSION_1_0)) {
+            builder.addChild(tlsParser.tlsParser_1_0);
+        }
```
pferraro, line 126: These static PersistenceResourceXMLDescription references kind of defeat the purpose of paramaterizing your parser.  Would it not be better to build this PersistenceResourceXMLDescription in one place?  That way you can easily tell, at a glance, what changed when.

elytron/src/main/java/org/wildfly/extension/elytron/ElytronPersistentResourceXMLDescriptionFactory.java
```diff
@@ -0,0 +1,228 @@
...
+                    .addAttribute(ElytronDefinition.DEFAULT_SSL_CONTEXT);
+        }
+
+        addAuthenticationClientParser(builder, schema);
+
+        if (schema.since(ElytronSubsystemSchema.VERSION_1_0)) {
+            builder.addChild(new ProviderParser().parser);
+        }
+
+        addAuditLoginParser(builder, schema);
```
fjuma, line 49: s/addAuditLoginParser/addAuditLoggingParser

elytron/src/main/java/org/wildfly/extension/elytron/ElytronPersistentResourceXMLDescriptionFactory.java
```diff
@@ -0,0 +1,228 @@
...
+        }
+
+        addTlsParser(schema, builder);
+        addCredentialStoreParser(schema, builder);
+
+        if (schema.since(ElytronSubsystemSchema.VERSION_13_0)) {
+            builder.addChild(getExpressionResolverParser());
+        }
+
+        if (schema.since(ElytronSubsystemSchema.VERSION_3_0)) {
```
fjuma, line 78: Should this be schema.since(ElytronSubsystemSchema.VERSION_1_0)?

elytron/src/main/java/org/wildfly/extension/elytron/ElytronPersistentResourceXMLDescriptionFactory.java
```diff
@@ -0,0 +1,228 @@
...
+        }
+
+        return builder.build();
+    }
+
+    private static void addPolicyParser(ElytronSubsystemSchema schema, PersistentResourceXMLDescription.PersistentResourceXMLBuilder builder) {
+        PolicyParser policyParser = new PolicyParser();
+        if (schema.since(ElytronSubsystemSchema.VERSION_1_2)) {
+            builder.addChild(policyParser.parser_1_2);
+        } else if (schema.since(ElytronSubsystemSchema.VERSION_1_1)) {
```
fjuma, line 94: Should this be schema.since(ElytronSubsystemSchema.VERSION_1_0)?

elytron/src/main/java/org/wildfly/extension/elytron/SSLContextResource.java
```diff
@@ -129,6 +130,9 @@ public Resource clone() {
      */
     private boolean hasActiveSessions() {
         final SSLContext sslContext = getSSLContext(sslContextServiceController);
+        if (sslContext instanceof DynamicSSLContext) {
```
fjuma, line 133: What about the case where the module that provides this class doesn't exist?

elytron/src/main/resources/org/wildfly/extension/elytron/LocalDescriptions.properties
```diff
@@ -1439,10 +1439,17 @@ elytron.client-ssl-context.ssl-session.peer-certificates.version=The certificate
 # Operations
 elytron.client-ssl-context.ssl-session.invalidate=Invalidate the SSLSession (Note: This does not terminate current connections, only prevents future connections from joining or resuming this session).
 
+elytron.dynamic-client-ssl-context=A Dynamic SSL context for use on the client side of a connection. It allows to choose ssl context based on peer's host and port information
```
fjuma, line 1442: s/It allows to choose ssl context based on peer's host and port information/It makes it possible to choose the SSL context to use based on the peer's host and port information

elytron/src/main/java/org/wildfly/extension/elytron/SSLContextResource.java
Skyllarr: If so, the best approach is to actually include this dynamic-ssl module in product but document it as not supported module. And the resource itself won't be possible to register in DEFAULT stability. This sounds better than removing these calls in wildfly-to-product chain

elytron/src/test/resources/org/wildfly/extension/elytron/elytron-subsystem-community-18.0.xml
```diff
@@ -0,0 +1,396 @@
...
+                                session-timeout="120" wrap="false" key-manager="serverKey" trust-manager="serverTrust" pre-realm-principal-transformer="a"
+                                post-realm-principal-transformer="b" final-principal-transformer="c" realm-mapper="d" providers="custom-loader" provider-name="first"/>
+            <server-ssl-context name="server2" protocols="TLSv1.2" want-client-auth="true" need-client-auth="true" authentication-optional="true"
+                                use-cipher-suites-order="false" maximum-session-cache-size="10"
+                                session-timeout="120" wrap="false" key-manager="serverKey" trust-manager="serverTrust" pre-realm-principal-transformer="a"
+                                post-realm-principal-transformer="b" final-principal-transformer="c" realm-mapper="d" providers="custom-loader" provider-name="first"/>
+        </server-ssl-contexts>
+        <client-ssl-contexts>
+            <client-ssl-context name="client" protocols="TLSv1.3 TLSv1.2" key-manager="clientKey" trust-manager="serverTrust" providers="custom-loader"
```
fjuma, line 328: Should this file have a dynamic-client-ssl-context configured?

elytron/src/main/java/org/wildfly/extension/elytron/SSLContextResource.java
Skyllarr: Or should I use classloading and ServiceLoader here and in other instances in wildfly-core?

fjuma: @pferraro Do you have any thoughts on this?

elytron/src/main/java/org/wildfly/extension/elytron/ElytronPersistentResourceXMLDescriptionFactory.java
Skyllarr: @pferraro Do you mean not having separate parsers for TLS, credential store etc. but registering everything at one place based on version?

Skyllarr: or  maybe that is not what you meant. What did you mean by building PersistenceResourceXMLDescription in one place? As it is being built in this apply method at one place IIUC

elytron/src/main/java/org/wildfly/extension/elytron/TlsParser.java
Skyllarr: @pferraro Some of these parses do not only add a new child, but also replace some child. For example look at tlsParser_14_0 which uses trustManagerParser_14_0  but the previous tlsParser_12_0 uses trustManagerParser.

Skyllarr: So IMO this is why we did it this way up until now. Since eg. having a previous version's builder in parameter and only adding a new child would not always work.

But I can create a subsequent jira if this should be refactored if you prefer? it does not have to be refactored in this PR

elytron/src/main/java/org/wildfly/extension/elytron/SSLContextResource.java
darranl: Can we make DynamicSSLContext extend another interface in Elytron that will not be excluded, we can then do the instance check based on that.

elytron/src/main/java/org/wildfly/extension/elytron/TlsParser.java
darranl: Another one of our reasons was also to make it easier to remove older versions in the future - by duplicating we know the new version contains everything we need so we can delete the older ones with reduced risk once the time is right.

elytron/src/main/java/org/wildfly/extension/elytron/AuthenticationClientDefinitions.java
```diff
@@ -491,6 +494,17 @@ protected ValueSupplier<AuthenticationContext> getValueSupplier(ServiceBuilder<A
...
 
+            @Override
+            protected void performRuntime(OperationContext context, ModelNode operation, ModelNode model) throws OperationFailedException {
+                commonRequirements(installService(context, model)).setInitialMode(ServiceController.Mode.ON_DEMAND).install();
+            }
+
+            ServiceBuilder<AuthenticationContext> installService(OperationContext context, ModelNode model)  throws OperationFailedException {
+                ServiceTarget serviceTarget = context.getCapabilityServiceTarget();
+                ServiceBuilder<?> serviceBuilder = context.getCapabilityServiceTarget().addCapability(AUTHENTICATION_CONTEXT_RUNTIME_CAPABILITY);
+                TrivialService<AuthenticationContext> authenticationContextTrivialService = new TrivialService<AuthenticationContext>(getValueSupplier((ServiceBuilder<AuthenticationContext>) serviceBuilder, context, model));
```
pferraro, line 505: Not something you need to address here, but at some point, I would encourage someone to update TrivialService such that it does not rely on deprecated MSC API - which is what is requiring a lot of this unfortunate casting.

elytron/src/main/java/org/wildfly/extension/elytron/TlsParser.java
pferraro: > @pferraro Some of these parses do not only add a new child, but also replace some child. For example look at `tlsParser_14_0` which uses `trustManagerParser_14_0` but the previous `tlsParser_12_0` uses `trustManagerParser` instead.

Hmm - I'm confused.  Why would that preclude building the PersistentResourceXMLDescription dynamically based on the schema version?

pferraro: > So IMO this is why we did it this way up until now. Since eg. having a previous version's builder in parameter and only adding a new child would not always work.

Sorry if I was unclear.  The "parameter" to which I was referring is an ElytronSubsystemSchema instance, not the previous version's builder.

pferraro (changes requested):

elytron/src/main/java/org/wildfly/extension/elytron/TlsParser.java
darranl: Before we do anything else to refactor parsing I think we need to understand the different future updates we need to be able to support.

elytron/src/main/java/org/wildfly/extension/elytron/SSLContextResource.java
darranl: Those are only going to be used at Stage.RUNTIME once this feature is enabled AND a resource is configured.

I think we can move those calls into a class with a method that can return the instance and the SSLDefininitions class updated to call that directly.

Skyllarr: @darranl Okay I'll be working on that today and will post an update here

darranl: @Skyllarr I think the part that needs changes in WildFly Elytron is the higher priority to be included in the release from @fjuma on the WildFly (and core) side if the rest of the PR is approved I think we can get away with a blocker Jira issue to follow up.  There are also changed coming to Galleon that we may be able to take advantage of after Beta to automate the provisiononing of this artifact based on stability level.

core-feature-pack/common/src/main/resources/modules/system/layers/base/org/wildfly/security/elytron-base/main/module.xml
```diff
@@ -34,6 +34,7 @@
         <artifact name="${org.wildfly.security:wildfly-elytron-credential-source-impl}"/>
         <artifact name="${org.wildfly.security:wildfly-elytron-credential-store}"/>
         <artifact name="${org.wildfly.security:wildfly-elytron-digest}"/>
+        <artifact name="${org.wildfly.security:wildfly-elytron-dynamic-ssl}"/>
```
darranl, line 37: FYI not a reason to avoid merging this PR but I think we are going to need to move the artifact into it's own module, we can then add an optional dependency on the new module here.

As I mention in my other comment we will look at the capabilities being added with Galleon to not provision this artifact and the new module when provisioning at the default stability level.

elytron/src/main/java/org/wildfly/extension/elytron/AuthenticationClientDefinitions.java
Skyllarr: @pferraro Created https://issues.redhat.com/browse/WFCORE-6752

elytron/src/main/java/org/wildfly/extension/elytron/AuthenticationClientParser.java
```diff
@@ -40,7 +40,7 @@ class AuthenticationClientParser {
             .addAttribute(AuthenticationClientDefinitions.MATCH_RULES, AttributeParser.UNWRAPPED_OBJECT_LIST_PARSER, AttributeMarshaller.UNWRAPPED_OBJECT_LIST_MARSHALLER)
             .build();
 
-    final PersistentResourceXMLDescription parser = decorator(ElytronDescriptionConstants.AUTHENTICATION_CLIENT)
+    final PersistentResourceXMLDescription parser_1_0 = decorator(ElytronDescriptionConstants.AUTHENTICATION_CLIENT)
```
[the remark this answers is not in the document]
Skyllarr: I have changed this back to be `parser`

elytron/src/test/resources/org/wildfly/extension/elytron/elytron-subsystem-community-18.0.xml
Skyllarr: Added

elytron/src/main/resources/org/wildfly/extension/elytron/LocalDescriptions.properties
Skyllarr: Updated

elytron/src/main/java/org/wildfly/extension/elytron/SSLContextResource.java
Skyllarr: @darranl  I have added new commits to both wildfly-elytron and wildfly-core PRs that make use of new interface for this reason. I have used separate commits so that if you will end up preferring a different solution or a different interface you can easily remove it. I did not manage to do a refactoring of SSLDefinitions class though

elytron/src/main/java/org/wildfly/extension/elytron/ElytronPersistentResourceXMLDescriptionFactory.java
Skyllarr: Fixed

elytron/src/main/java/org/wildfly/extension/elytron/ElytronPersistentResourceXMLDescriptionFactory.java
Skyllarr: Fixed

elytron/src/main/java/org/wildfly/extension/elytron/ElytronPersistentResourceXMLDescriptionFactory.java
Skyllarr: Fixed

elytron/src/main/java/org/wildfly/extension/elytron/ElytronPersistentResourceXMLDescriptionFactory.java
Skyllarr: Fixed

Skyllarr: This is now fixed

elytron/src/main/java/org/wildfly/extension/elytron/ElytronPersistentResourceXMLDescriptionFactory.java
Skyllarr: @pferraro  Sounds good, I have moved this logic to ElytronSubsystemSchema.getXMLDescription() method

elytron/src/main/java/org/wildfly/extension/elytron/ElytronPersistentResourceXMLDescriptionFactory.java
Skyllarr: Fixed

elytron/src/main/java/org/wildfly/extension/elytron/ElytronPersistentResourceXMLDescriptionFactory.java
Skyllarr: This enum was removed

elytron/src/main/java/org/wildfly/extension/elytron/AuthenticationClientDefinitions.java
Skyllarr: @pferraro You're right, I have updated it to be ON_DEMAND

elytron/src/main/java/org/wildfly/extension/elytron/TrivialResourceDefinition.java
Skyllarr: Fixed!

elytron/src/main/java/org/wildfly/extension/elytron/ElytronDefinition.java
Skyllarr: Fixed!

elytron/src/main/java/org/wildfly/extension/elytron/JaspiConfigurationParser.java
Skyllarr: Fixed!

Skyllarr:
@darranl @fjuma Thank you very much for reviews! Just an update that I have addressed some of your comments and I commented on those that I've addressed. The remaining comments I was not able to address before my PTO

elytron/src/main/java/org/wildfly/extension/elytron/SSLContextResource.java
Skyllarr: @darranl Maybe we can remove the method from the new interface and leave the interface empty, and only return `false` here instead of `((ActiveSessionsSSLContext) sslContext).hasActiveSessions();`

darranl:
@fjuma it sounds like the main piece missing here is finishing off the skipping provisioning when not at the community stability level - I think we can move that to a Blocker issue WFCORE / ELY to address before WildFly 32 Final

elytron/src/main/java/org/wildfly/extension/elytron/ElytronSubsystemTransformers.java
```diff
@@ -170,6 +170,8 @@ private static void from18(ChainedTransformationDescriptionBuilder chainedBuilde
         .setDiscard(DiscardAttributeChecker.UNDEFINED, EMIT_EVENTS)
         .addRejectCheck(new RejectAttributeChecker.SimpleRejectAttributeChecker(ModelNode.TRUE), IGNORE_UNAVAILABLE_REALMS)
         .addRejectCheck(RejectAttributeChecker.DEFINED, EMIT_EVENTS);
+        builder.rejectChildResource(PathElement.pathElement(ElytronDescriptionConstants.DYNAMIC_CLIENT_SSL_CONTEXTS));
+        builder.rejectChildResource(PathElement.pathElement(ElytronDescriptionConstants.DYNAMIC_CLIENT_SSL_CONTEXT));
```
darranl, line 174: I think we can add a follow up commit to not block this PR but @pferraro from something you said elsewhere, are we missing a subsystem model bump?

I think you said a change to any resource at any stability level should trigger a bump so we probably need this bumping to 19 so the transformer should then be when transforming from19

pferraro: > I think we can add a follow up commit to not block this PR but @pferraro from something you said elsewhere, are we missing a subsystem model bump?

Yes - this requires a model version bump.
Modifying an existing transformer for a version that was already released will cause regressions.

elytron/src/test/java/org/wildfly/extension/elytron/TestEnvironment.java
```diff
@@ -44,6 +45,8 @@ class TestEnvironment extends AdditionalInitialization {
     private static final X500Principal ISSUER_DN = new X500Principal("O=Root Certificate Authority, EMAILADDRESS=elytron@wildfly.org, C=UK, ST=Elytron, CN=Elytron CA");
     private static final X500Principal LOCALHOST_DN = new X500Principal("OU=Elytron, O=Elytron, C=CZ, ST=Elytron, CN=localhost");
 
+    private static Stability stability;
```
pferraro, line 48: Why is this static?

pferraro (changes requested):

elytron/src/test/java/org/wildfly/extension/elytron/TestEnvironment.java
Skyllarr: Oops, fixed

elytron/src/main/java/org/wildfly/extension/elytron/AuthenticationClientDefinitions.java
```diff
@@ -491,6 +494,17 @@ protected ValueSupplier<AuthenticationContext> getValueSupplier(ServiceBuilder<A
...
+                return serviceTarget.addService(AUTHENTICATION_CONTEXT_RUNTIME_CAPABILITY.getCapabilityServiceName(context.getCurrentAddressValue()), authenticationContextTrivialService);
```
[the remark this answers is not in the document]
fjuma: Created an issue, [WFCORE-6757](https://issues.redhat.com/browse/WFCORE-6757), to track this.

fjuma (approved):
Adding my approval on this one since I've created WFCORE issues to track the remaining items. The created issues have also been linked from [WFCORE-5101](https://issues.redhat.com/browse/WFCORE-5101).

PR #4312, opened 2020-09-01
Fix for WFCORE-5110, Bootable JAR doesn't setup the JAXP factories at boot

jfdenise:
Issue: https://issues.redhat.com/browse/WFCORE-5110

Replace JAXP factories with the proper implementations.


jmesnil:
@jfdenise thanks

PR #4315, opened 2020-09-02
Fix for WFCORE-5116, Bootable JAR, JMX not properly initialized at boot.

jfdenise:
Issue: https://issues.redhat.com/browse/WFCORE-5116


jmesnil:
@jfdenise thanks

PR #4316, opened 2020-09-03
Bootable JAR Logging Updates

jamezp:
# The following issues are fixed in this PR:

- https://issues.redhat.com/browse/WFCORE-5115
- https://issues.redhat.com/browse/WFCORE-5109

## [WFCORE-5115](https://issues.redhat.com/browse/WFCORE-5115) 
Use a hard-coded boot-config.properties to load system properties in the bootable JAR entry point. This properties file is generated by the wildfly-jar-maven-plugin. It's used at boot to configure properties required by the log manager during initialization.

## [WFCORE-4895](https://issues.redhat.com/browse/WFCORE-4895)
Minor follow up to the bootable JAR module. Actually throw the exception when a shutdown failure occurs.
    

## [WFCORE-5109](https://issues.redhat.com/browse/WFCORE-5109)
Ensure the configurator is attached to the root logger. This means the logging subsystem will not reconfigure resources that do not need to be reconfigured.

### DIFF
--- a/bootable-jar/runtime/src/main/java/org/wildfly/core/jar/runtime/BootableEnvironment.java
+++ b/bootable-jar/runtime/src/main/java/org/wildfly/core/jar/runtime/BootableEnvironment.java
@@ -1,0 +1,226 @@
+/*
+ * JBoss, Home of Professional Open Source.
+ *
+ * Copyright 2020 Red Hat, Inc., and individual contributors
+ * as indicated by the @author tags.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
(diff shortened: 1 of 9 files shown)


jamezp:
I'm placing a temporary hold as we need to ensure the default `relative-to` directories are available. More than likely only the `jboss.server.log.dir` is used, but we should ensure they're all set as system properties for the log manager to use.

bootable-jar/runtime/src/main/java/org/wildfly/core/jar/runtime/BootableEnvironment.java
```diff
@@ -0,0 +1,226 @@
...
+            logDebug("Setting system property %s to %s", key, path);
+        } else {
+            logDebug("Replacing system property %s with a value of %s. The previous value was %s.", key, path, previousValue);
+        }
+    }
+
+    @SuppressWarnings("UseOfSystemOutOrSystemErr")
+    static void logDebug(final String format, final Object... args) {
+        if (DEBUG.get()) {
+            System.out.printf("[DEBUG] " + format, args);
```
jamezp, line 223: Note this is just an idea. I'm happy to remove it, but felt it may be useful to see some information about the environment when it's booting.

jfdenise:
Approved, @jamezp really awesome changes, thank-you!

jmesnil:
@jamezp thanks

PR #4317, opened 2020-09-04
Fix for WFCORE-5112, Upgrade bootable jar maven plugin to 2.0.0.Beta4

jfdenise:
Issue:https://issues.redhat.com/browse/WFCORE-5112

In addition to the upgrade, the plugin configuration has been simplified, inherit-configs has been removed. inherit-packages is false by default.


jmesnil:
@jfdenise thanks

PR #4320, opened 2020-09-07
[WFCORE-5118] Cannot get resources from ContextClassLoader with Server Lifecycle Events

rmartinc:
Issue: https://issues.redhat.com/browse/WFCORE-5118

I have added just a try/finally to add the TCCL to the one in the module specified. I have used the `WildFlySecurityManager` class to do that which needs to add that dependency in the `module.xml`. If you prefer to add a helper class or something different just let me know (I think that it's better to not repeat the code).


rmartinc:
@jmesnil I was trying to assign the review to you, but it seems it's not possible in this project. So just pinging you instead. :-)

darranl:
IMO the dependency on elytron-private is fine, this module is the common API / SPI for the server so expected extensions will depend on it - this is just the first time it was needed for this extenstion.

The use of WildFlySecurityManager to manage the context class loader looks good to me as well.

jmesnil (approved):

jmesnil:
@rmartinc thanks

PR #4321, opened 2020-09-07
Fix for WFCORE-5119, Bootable JAR - Remove dependency on elytron-private

jfdenise:
Issue: https://issues.redhat.com/browse/WFCORE-5119

Remove useless dependency on elytron-private.

@darranl FYI.


darranl:
@jfdenise Sorry it may only be the services="import" that can be removed - if you do run into issues the APIs may be needed just not the definition of the services.

jfdenise:
@darranl , I can't find reference to elytron in the bootable module pom file. Elytron is not directly used.

darranl:
Ok should be good then, I think normal module loading should commence during boot anyway for the other modules.

jmesnil:
@jfdenise thanks

PR #4322, opened 2020-09-08
[WFCORE-5121] Add ability to ignore excluded extensions to FetchMissi…

spyrkob:
JIRA: https://issues.redhat.com/browse/WFCORE-5121

Attempts to solve the issue by passing the domain exclude info used to generate initial sync to handler doing missing config sync. This requires changing visibility of some methods, I'm not sure if there's a better way. 
@jmesnil could you review please?

### DIFF
--- a/host-controller/src/main/java/org/jboss/as/domain/controller/operations/FetchMissingConfigurationHandler.java
+++ b/host-controller/src/main/java/org/jboss/as/domain/controller/operations/FetchMissingConfigurationHandler.java
@@ -50,10 +50,12 @@ public class FetchMissingConfigurationHandler implements OperationStepHandler {
 
     private final Transformers transformers;
     private final ExtensionRegistry extensionRegistry;
+    private final Set<String> excludedExtensions;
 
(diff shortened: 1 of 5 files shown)


jmesnil:
@yersan could you please review?

yersan (changes requested):
@spyrkob A slightly different approach could be exposing the `HostInfo.domainIgnoredExtensions `and add them as an attachment to the slave host controller channel, then, using the similar approach as you did with the Version, you could get the exclusions on `FetchMissingConfigurationHandler`. That could reduce the coupling and the amount of methods made public. What do you think?

When the slave is connected, the HostInfo gets the domain exclusions for this host. If an exclusion is modified by the user at runtime, then the slave has to be reloaded, but I think that would be fine and expected. Anyway, change an exclusion at runtime does not look like as a common operation.

spyrkob:
@yersan, thanks I like your idea, it makes this much cleaner :)

I made the changes in separate commit for now, let me know if it looks good and I'll squash them.

yersan (approved):
Thank you @spyrkob, yes looks good to me too, ready to squash them

spyrkob:
Done, thanks :)

jmesnil:
@spyrkob thanks

PR #4326, opened 2020-09-11
Fix for WFWIP-355, Bootable JAR - scripts in bin folder of the server installation directory are not executable

jfdenise:
Issue: https://issues.redhat.com/browse/WFWIP-355

### DIFF
--- a/bootable-jar/boot/src/main/java/org/wildfly/core/jar/boot/Main.java
+++ b/bootable-jar/boot/src/main/java/org/wildfly/core/jar/boot/Main.java
@@ -27,11 +27,15 @@ import java.nio.file.Files;
 import java.nio.file.Path;
 import java.nio.file.Paths;
 import java.nio.file.StandardCopyOption;
+import java.nio.file.attribute.PosixFilePermission;
 import java.security.Policy;
 import java.util.ArrayList;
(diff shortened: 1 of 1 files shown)


bootable-jar/boot/src/main/java/org/wildfly/core/jar/boot/Main.java
```diff
@@ -190,13 +208,20 @@ private static void unzip(InputStream wf, Path dir) throws Exception {
                         Files.createDirectories(parent);
                     }
                     Files.copy(zis, newFile, StandardCopyOption.REPLACE_EXISTING);
+                    if (!isWindows && newFile.getFileName().toString().endsWith(".sh")) {
+                        Files.setPosixFilePermissions(newFile, EXECUTE_PERMISSIONS);
```
darranl, line 212: Just a general FYI we may want to record for later, the method setPosixFilePermissions involves a SecurityManager permissions check - HOWEVER this Main class is the entry point to the application meaning it will be the only protection domain on the call stack, all a doPrivileged call does is drop the parent protection domains from the call stack so as this is class is the first this is not nescesary.

darranl (approved):

PR #4333, opened 2020-09-16
[WFCORE-5135] Add a test case to verify that banned modules are not provisioned

yersan:
Jira issue: https://issues.redhat.com/browse/WFCORE-5135

### DIFF
--- a/testsuite/layers/pom.xml
+++ b/testsuite/layers/pom.xml
@@ -766,9 +766,11 @@
                                                 <layer>core-management</layer>
                                                 <layer>core-security-realms</layer>
                                                 <layer>core-server</layer>
+                                                <layer>core-tools</layer>
                                                 <layer>deployment-scanner</layer>
                                                 <layer>discovery</layer>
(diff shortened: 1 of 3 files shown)


bstansberry:
I think some kind of scoping mechanism is needed.  For example, the org.jboss.as.security module is not banned in all contexts, only in most.  We actually have a legacy-security layer that explicitly provides it and it's valid to include that layer in combination with others in an overall config being tested.

yersan:
Yes right Brian, I've just added such possibility. The idea is to be able to configure a banned module and a list of expected installations where that module can be found.

I don't like so much such coupling between the pom.xml declaration where the layer is provisioned and the test itself though, but it should work as it is.

It forces us to always have at least a provision of each new layer in isolation, which sounds very reasonable to me, at least for full, where the provision that installs a server with all the layers together should be ignored due to legacy-security

darranl (approved):
This is needed for testing in WildFly so going to go ahead and merge.

PR #4334, opened 2020-09-17
[WFCORE-5136] NPE in ModelControllerMBeanHelper

spyrkob:
Issue: https://issues.redhat.com/browse/WFCORE-5136

### DIFF
--- a/jmx/src/main/java/org/jboss/as/jmx/model/MBeanInfoFactory.java
+++ b/jmx/src/main/java/org/jboss/as/jmx/model/MBeanInfoFactory.java
@@ -266,8 +266,11 @@ public class MBeanInfoFactory {
         List<OpenMBeanParameterInfo> params = new ArrayList<OpenMBeanParameterInfo>(propertyList.size());
 
         Map<String, AttributeDefinition> attributeDefinitions = new HashMap<>();
-        for (AttributeDefinition attributeDefinition : opDef.getParameters()) {
-            attributeDefinitions.put(attributeDefinition.getName(), attributeDefinition);
+        AttributeDefinition[] attrs = opDef.getParameters();
+        if (attrs != null) {
+            for (AttributeDefinition attributeDefinition : attrs) {
+                attributeDefinitions.put(attributeDefinition.getName(), attributeDefinition);
+            }
         }
 
         for (Property prop : propertyList) {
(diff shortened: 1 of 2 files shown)


jmesnil (approved):

bstansberry:
@spyrkob @jmesnil 

In the original change that introduced this bug this was added as well:

https://github.com/wildfly/wildfly-core/blob/961b89a7abf5eeef84cbd107b56de6a18c72bf40/jmx/src/main/java/org/jboss/as/jmx/model/ModelControllerMBeanHelper.java#L522

It's theoretically possible 'attrs' would be null which would NPE two lines below.  We haven't hit that so maybe it can't actually happen, so doing anything about it can be out of scope for this. But if there is reason to update this patch, guarding that for loop at L524 would be good.

spyrkob:
I'm happy to update the PR - there's a similar call in https://github.com/wildfly/wildfly-core/blob/961b89a7abf5eeef84cbd107b56de6a18c72bf40/jmx/src/main/java/org/jboss/as/jmx/model/MBeanInfoFactory.java#L269 
Might be good to catch that as well?

jmesnil:
Brian is right, we should also guard against it.

Having said that... the NPE occurs because the OperationDefinition for `add-jndi` is using a deprecated constructor of SimpleOperationDefinition:

https://github.com/wildfly/wildfly-core/blob/68501b65f33e3a6c87ac9873333ea7775adc544a/controller/src/main/java/org/jboss/as/controller/SimpleOperationDefinition.java#L78-L80

This constructor is inconsistent as unspecified `parameters` defaults to `new AttributeDefinition[0]` but unspecified `replyParameters` defaults to `null`.

The `SimpleOperationDefinitionBuilder` (which is the recommended way to build operation definitions) defaults both to `new AttributeDefinition[0]`:

https://github.com/wildfly/wildfly-core/blob/68501b65f33e3a6c87ac9873333ea7775adc544a/controller/src/main/java/org/jboss/as/controller/SimpleOperationDefinitionBuilder.java#L44

Independent of this PR, I'll file an issue to update deprecated SimpleOperationDefinition constructors so that they also use `new AttributeDefinition[0]` for unspecified reply parameters.

jmesnil (approved):

jmesnil:
@spyrkob thanks

PR #4340, opened 2020-09-22
Fix for WFCORE-5139, Upgrade Bootable JAR Maven plugin to 2.0.0.Beta8

jfdenise:
Issue: https://issues.redhat.com/browse/WFCORE-5139

Changes since last upgrade: https://github.com/wildfly-extras/wildfly-jar-maven-plugin/compare/2.0.0.Beta4...2.0.0.Beta8

In addition to the upgrade, the plugin version has been removed from parent pom.xml and moved in ts.bootable profile defined in testsuite/pom.xml.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -118,7 +118,7 @@
         <version.org.wildfly.galleon-plugins>4.2.8.Final</version.org.wildfly.galleon-plugins>
         <!-- wildfly plugin-->
         <version.wildfly.plugin>2.0.2.Final</version.wildfly.plugin>
-        <version.org.wildfly.jar.plugin>2.0.0.Beta4</version.org.wildfly.jar.plugin>
+        <version.org.wildfly.jar.plugin>2.0.0.Beta8</version.org.wildfly.jar.plugin>
         <!-- plugins related to wildfly build and tooling -->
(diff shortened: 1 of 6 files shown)


pom.xml
```diff
@@ -118,7 +118,6 @@
         <version.org.wildfly.galleon-plugins>4.2.8.Final</version.org.wildfly.galleon-plugins>
         <!-- wildfly plugin-->
         <version.wildfly.plugin>2.0.2.Final</version.wildfly.plugin>
-        <version.org.wildfly.jar.plugin>2.0.0.Beta4</version.org.wildfly.jar.plugin>
```
jmesnil, line 121: @jfdenise could you please keep this property in `pom.xml` instead of moving it to `testsuite/pom.xml`.

Even though the dependency is only used in the test suite, having the version defined in `pom.xml` make sure we keep a centralized place for all versions of the components we depend on

jmesnil (changes requested):

pom.xml
jfdenise: Thank-you. Done.

jmesnil (approved):

jmesnil:
@jfdenise thanks

PR #4341, opened 2020-09-22
[WFCORE-5095] Add the ability to make use of an automatically generated self-signed certificate with Elytron

fjuma:
https://issues.redhat.com/browse/WFCORE-5095
https://issues.redhat.com/browse/EAP7-1545

This is for WildFly Core 14. This PR also contains the Elytron subsystem bump from https://github.com/wildfly/wildfly-core/pull/4342.

Analysis document: https://github.com/wildfly/wildfly-proposals/pull/332
WildFly PR with documentation and tests: https://github.com/wildfly/wildfly/pull/13596

### DIFF
--- a/core-feature-pack/galleon-common/src/main/resources/feature_groups/core-domain-profile.xml
+++ b/core-feature-pack/galleon-common/src/main/resources/feature_groups/core-domain-profile.xml
@@ -14,6 +14,13 @@
                     <param name="relative-to" value="jboss.domain.config.dir"/>
                 </feature>
             </include>
+            <include feature-id="subsystem.elytron.key-store:key-store=applicationKS">
+                <param name="key-store" value="applicationKS" />
+                <param name="path" value="application.keystore" />
+                <param name="relative-to" value="jboss.domain.config.dir" />
+                <param name="type" value="PKCS12" />
+                <param name="credential-reference" value="{clear-text=password}" />
+            </include>
         </feature-group>
     </feature>
 
--- a/elytron/src/main/java/org/wildfly/extension/elytron/SSLDefinitions.java
+++ b/elytron/src/main/java/org/wildfly/extension/elytron/SSLDefinitions.java
@@ -24,6 +24,7 @@ import static org.jboss.as.controller.security.CredentialReference.rollbackCrede
 import static org.wildfly.extension.elytron.Capabilities.KEY_MANAGER_CAPABILITY;
 import static org.wildfly.extension.elytron.Capabilities.KEY_MANAGER_RUNTIME_CAPABILITY;
 import static org.wildfly.extension.elytron.Capabilities.KEY_STORE_CAPABILITY;
+import static org.wildfly.extension.elytron.Capabilities.KEY_STORE_RUNTIME_CAPABILITY;
 import static org.wildfly.extension.elytron.Capabilities.PRINCIPAL_TRANSFORMER_CAPABILITY;
 import static org.wildfly.extension.elytron.Capabilities.PROVIDERS_CAPABILITY;
 import static org.wildfly.extension.elytron.Capabilities.REALM_MAPPER_CAPABILITY;
@@ -93,6 +94,7 @@ import org.jboss.as.controller.SimpleAttributeDefinitionBuilder;
 import org.jboss.as.controller.SimpleMapAttributeDefinition;
 import org.jboss.as.controller.SimpleOperationDefinitionBuilder;
 import org.jboss.as.controller.StringListAttributeDefinition;
+import org.jboss.as.controller.capability.RuntimeCapability;
 import org.jboss.as.controller.descriptions.ResourceDescriptionResolver;
 import org.jboss.as.controller.descriptions.StandardResourceDescriptionResolver;
 import org.jboss.as.controller.logging.ControllerLogger;
@@ -111,6 +113,7 @@ import org.jboss.msc.service.ServiceBuilder;
 import org.jboss.msc.service.ServiceController;
 import org.jboss.msc.service.ServiceController.State;
 import org.jboss.msc.service.ServiceName;
+import org.jboss.msc.service.ServiceRegistry;
 import org.jboss.msc.service.StartException;
 import org.jboss.msc.value.InjectedValue;
 import org.wildfly.common.function.ExceptionSupplier;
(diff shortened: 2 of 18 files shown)


fjuma:
Note that the failing WildFly test (`org.jboss.as.test.integration.management.cli.SecurityCommandsTestCase#testKeyManagerDifferentAlgorithm`) passes with the changes from my WildFly PR: https://github.com/wildfly/wildfly/pull/13596

darranl:
This RFE is being reviewed so is not stale.

elytron/src/main/java/org/wildfly/extension/elytron/SSLDefinitions.java
```diff
@@ -1017,6 +1027,110 @@ public String chooseEngineServerAlias(String keyType, Principal[] issuers, SSLEn
         }
     }
 
+    private static class LazyDelegatingKeyManager extends DelegatingKeyManager {
```
darranl, line 1030: Not a reason to delay the feature as this is private so exposure is limited but should a class such as this be a utility in the Elytron project?

elytron/src/main/java/org/wildfly/extension/elytron/SSLDefinitions.java
```diff
@@ -1017,6 +1027,110 @@ public String chooseEngineServerAlias(String keyType, Principal[] issuers, SSLEn
...
+            this.generateSelfSignedCertificateHostName = generateSelfSignedCertificateHostName;
+            this.aliasFilter = aliasFilter;
+        }
+
+        private void doInit() {
+            if(! init) {
+                synchronized (this) {
+                    if(! init) {
+                        try {
+                            ((KeyStoreService) keyStoreService).generateAndSaveSelfSignedCertificate(generateSelfSignedCertificateHostName, password);
```
darranl, line 1052: Ah I see it is MSC aware.

elytron/src/main/resources/org/wildfly/extension/elytron/LocalDescriptions.properties
```diff
@@ -1260,6 +1260,7 @@ elytron.key-manager.credential-reference.type=The type of credential this refere
 elytron.key-manager.credential-reference.clear-text=The secret specified using clear text. Check credential store way of supplying credential/secrets to services.
 elytron.key-manager.key-store=Reference to the KeyStore to use to initialise the underlying KeyManagerFactory.
 elytron.key-manager.alias-filter=A filter to apply to the aliases returned from the KeyStore, can either be a comma separated list of aliases to return or one of the following formats ALL:-alias1:-alias2, NONE:+alias1:+alias2
+elytron.key-manager.generate-self-signed-certificate-host=If the file that backs the KeyStore does not exist and this attribute is set, then a self-signed certificate will be generated for the specified host name. This is not intended for production use.
```
darranl, line 1263: How about if the file exists but is empty?

core-feature-pack/galleon-common/src/main/resources/feature_groups/core-domain-profile.xml
```diff
@@ -14,6 +14,13 @@
                     <param name="relative-to" value="jboss.domain.config.dir"/>
                 </feature>
             </include>
+            <include feature-id="subsystem.elytron.key-store:key-store=applicationKS">
+                <param name="key-store" value="applicationKS" />
+                <param name="path" value="application.keystore" />
+                <param name="relative-to" value="jboss.domain.config.dir" />
+                <param name="type" value="JKS" />
```
darranl, line 21: Would now be the time to review the default KeyStore type we choose?

darranl:
I have added a couple of small comments in the actual commits, just a couple of additional points to clarify here as well:

- What is the trigger for the creation of the file?

- What is the trigger for the generation of the certificate?

elytron/src/main/resources/org/wildfly/extension/elytron/LocalDescriptions.properties
fjuma: The legacy equivalent only generates the self-signed certificate if the file that backs the keystore does not exist, i.e., if the file exists but is empty, legacy security doesn't generate the certificate. Currently, the behaviour with Elytron is the same as it was for legacy security. However, we could change the behaviour with Elytron so that we also generate the certificate for the case where the file exists but is empty.

core-feature-pack/galleon-common/src/main/resources/feature_groups/core-domain-profile.xml
fjuma: Good point, we should probably switch to pkcs12.

fjuma:
> I have added a couple of small comments in the actual commits, just a couple of additional points to clarify here as well:
> 
> * What is the trigger for the creation of the file?

The file (and the certificate) will be created on first use, i.e., when an HTTPS client connects to the server, if the file does not exist.

> * What is the trigger for the generation of the certificate?

Currently, the certificate is only generated if the file does not exist (to match the behaviour with legacy security). However, I think we could tweak this so that with Elytron, a file that exists but is empty is also a trigger for the generation of the certificate.

darranl:
Thanks @fjuma so I think we are good with the point in time the file and the certificate are generated, we could always expand that later - I just wasn't sure if we had a gap where the file could be created without the cert and end up preventing the cert generating on the next restart.

So just the PKCS12 default and I think we are good to go.

fjuma:
@darranl Sounds good. I've updated the default keystore type to PKCS12 and I've also updated my corresponding WildFly PR and my analysis document accordingly.

darranl (approved):

darranl:
@fjuma FYI the test failure on this PR looks like a genuine failure, it just came up again on the combined PR.

fjuma:
@darranl The failing WildFly test (`org.jboss.as.test.integration.management.cli.SecurityCommandsTestCase#testKeyManagerDifferentAlgorithm`) makes assertions that involve the number of configured TLS resources in the Elytron subsystem. Because my changes add new default TLS resources, the test fails because the numbers are off. The test passes with my changes from my corresponding WildFly PR:

https://github.com/wildfly/wildfly/pull/13596/files#diff-c47b177a13fdc2507089c3bfa6734735caf656df2e138a6a8f2a539c7dd4605b

darranl:
@fjuma Can anything be done to split the WildFly PR in two and get just that update into WildFly first?  If not the option may be submit a PR to WildFly to ignore the test and then in the PR you reference activate it again.

fjuma:
@darranl I don't think we can get that test update into WildFly first because it will fail on its own without these WildFly Core changes. I'll submit a PR to WildFly to ignore the test for now and then update my existing WildFly PR to un-ignore it.

fjuma:
@darranl I've submitted the following PR to ignore `SecurityCommandsTestCase` (ignoring only the one test case wasn't enough because then the other tests started failing with the same underlying issue):
https://github.com/wildfly/wildfly/pull/13759

I've also updated my existing WildFly PR to un-ignore `SecurityCommandsTestCase`:
https://github.com/wildfly/wildfly/pull/13596

PR #4352, opened 2020-10-05
[WFCORE-5147] Adjusting git integration logging and stability of RemoteSshGitRepositoryTestCase

darranl:
https://issues.redhat.com/browse/WFCORE-5147

The RemoteSshGitRepositoryTestCase test case was quite unstable on a resource constrained environment, the test was dependent on monitoring the server.log for an error message but it appears the git integration may log the error in a separate thread so I believe the process can be terminated before the message is logged.

Attempting to debug the issue I realised the test case was not properly restoring the default server configuration between test as attempts to add additional logging categories were being overridden.

Finally we were not logging useful information ourselves, although log and throw can be a bad practice at the moment the uncaught exception only made it's way to System.err - the problem with this was if you are reliant on the server.log the messages would be missing.

### DIFF
--- a/server/src/main/java/org/jboss/as/server/controller/git/ElytronClientCredentialsProvider.java
+++ b/server/src/main/java/org/jboss/as/server/controller/git/ElytronClientCredentialsProvider.java
@@ -39,12 +39,12 @@ import org.wildfly.security.auth.client.ElytronXmlParser;
  * Currently only login/password is supported.
  * @author Emmanuel Hugonnet (c) 2018 Red Hat, inc.
  */
-public class ElytronClientCredentialsProvider extends CredentialsProvider {
+class ElytronClientCredentialsProvider extends CredentialsProvider {
 
     private static AuthenticationContextConfigurationClient CLIENT = AccessController.doPrivileged(AuthenticationContextConfigurationClient.ACTION);
     private final AuthenticationContext context;
 
-    public ElytronClientCredentialsProvider(URI authenticationConfig) throws ConfigXMLParseException, GeneralSecurityException {
+    ElytronClientCredentialsProvider(URI authenticationConfig) throws ConfigXMLParseException, GeneralSecurityException {
         context =  ElytronXmlParser.parseAuthenticationClientConfiguration(authenticationConfig).create();
     }
 
--- a/server/src/main/java/org/jboss/as/server/controller/git/ElytronClientSshdSessionFactory.java
+++ b/server/src/main/java/org/jboss/as/server/controller/git/ElytronClientSshdSessionFactory.java
@@ -66,7 +66,7 @@ import org.wildfly.security.password.interfaces.ClearPassword;
  *
  * @author <a href="mailto:aabdelsa@redhat.com">Ashley Abdel-Sayed</a>
  */
-public class ElytronClientSshdSessionFactory extends SshdSessionFactory {
+class ElytronClientSshdSessionFactory extends SshdSessionFactory {
 
     private static AuthenticationContextConfigurationClient CLIENT = AccessController.doPrivileged(AuthenticationContextConfigurationClient.ACTION);
     private final AuthenticationContext context;
@@ -77,7 +77,7 @@ public class ElytronClientSshdSessionFactory extends SshdSessionFactory {
     private URI uri;
     private String knownHostsFile;
 
-    public ElytronClientSshdSessionFactory(URI authenticationConfig) throws ConfigXMLParseException, GeneralSecurityException {
+    ElytronClientSshdSessionFactory(URI authenticationConfig) throws ConfigXMLParseException, GeneralSecurityException {
         if (authenticationConfig != null) {
             context = ElytronXmlParser.parseAuthenticationClientConfiguration(authenticationConfig).create();
         } else {
--- a/server/src/main/java/org/jboss/as/server/logging/ServerLogger.java
+++ b/server/src/main/java/org/jboss/as/server/logging/ServerLogger.java
@@ -1353,6 +1353,9 @@ public interface ServerLogger extends BasicLogger {
     @Message(id = 279, value = "Git initialized in %s")
     void gitRespositoryInitialized(String name);
 
+    @Message(id = 280, value = "Unable to initialise the git repository.")
+    IllegalArgumentException unableToInitialiseGitRepository(@Cause Throwable cause);
+
     ////////////////////////////////////////////////
     //Messages without IDs
 
--- a/server/src/main/java/org/jboss/as/server/logging/ServerLogger.java
+++ b/server/src/main/java/org/jboss/as/server/logging/ServerLogger.java
@@ -1345,6 +1345,14 @@ public interface ServerLogger extends BasicLogger {
     @Message(id = 277, value = "Failed to load SSH Credentials %s")
     RuntimeException failedToLoadSSHCredentials(@Cause Throwable cause, String message);
 
+    @LogMessage(level = INFO)
+    @Message(id = 278, value = "The configuration history is managed through Git")
+    void usingGit();
+
+    @LogMessage(level = INFO)
+    @Message(id = 279, value = "Git initialized in %s")
+    void gitRespositoryInitialized(String name);
+
     ////////////////////////////////////////////////
     //Messages without IDs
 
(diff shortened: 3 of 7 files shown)


server/src/main/java/org/jboss/as/server/controller/git/ElytronClientCredentialsProvider.java
```diff
@@ -39,12 +39,12 @@
  * Currently only login/password is supported.
  * @author Emmanuel Hugonnet (c) 2018 Red Hat, inc.
  */
-public class ElytronClientCredentialsProvider extends CredentialsProvider {
+class ElytronClientCredentialsProvider extends CredentialsProvider {
```
jmesnil, line 42: @darranl, Is this class considered API? can the restriction of the visibility impact user code or just server code?

server/src/main/java/org/jboss/as/server/controller/git/ElytronClientSshdSessionFactory.java
```diff
@@ -66,7 +66,7 @@
  *
  * @author <a href="mailto:aabdelsa@redhat.com">Ashley Abdel-Sayed</a>
  */
-public class ElytronClientSshdSessionFactory extends SshdSessionFactory {
+class ElytronClientSshdSessionFactory extends SshdSessionFactory {
```
jmesnil, line 69: idem

server/src/main/java/org/jboss/as/server/logging/ServerLogger.java
```diff
@@ -1345,6 +1345,17 @@
     @Message(id = 277, value = "Failed to load SSH Credentials %s")
     RuntimeException failedToLoadSSHCredentials(@Cause Throwable cause, String message);
 
+    @LogMessage(level = INFO)
+    @Message(id = 278, value = "The configuration history is managed through Git")
```
jmesnil, line 1349: would it make sense to mention the path to the  cloned Git repository in that INFO log?

server/src/main/java/org/jboss/as/server/logging/ServerLogger.java
```diff
@@ -1354,23 +1365,10 @@
     @Message(id = Message.NONE, value = "The attribute '%s' has changed from '%s' to '%s'")
     String jmxAttributeChange(String name, String oldState, String stateString);
 
-    @LogMessage(level = INFO)
-    @Message(id = Message.NONE, value = "The configuration history is managed through Git")
-    void usingGit();
-
     @Message(id = Message.NONE, value = "Repository initialized")
```
jmesnil, line 1368: It's suspicious when I see a `@Message` without an ID. Most of the time, I consider that this should be a DEBUG log and in that case, I don't see which benefit we have to use a @Message for it instead of a plain `debugf()` call.

In that case, it turns out that this is a String used in a Git commit:

```
git.commit().setSign(gitConfig.isSign()).setMessage(ServerLogger.ROOT_LOGGER.repositoryInitialized()).call();
```

I'm not sure what we want to do here... 
Do we need to i18nize this commit message?
Should it be a stable message with a Message ID that can be tracked in a knowledge base?

jmesnil: @darranl that's not directly related to your PR though :)

testsuite/manualmode/src/test/java/org/jboss/as/test/manualmode/management/persistence/AbstractGitRepositoryTestCase.java
```diff
@@ -58,7 +58,7 @@
  */
 public class AbstractGitRepositoryTestCase {
 
-    private final Path jbossServerBaseDir = new File(System.getProperty("jboss.home", System.getenv("JBOSS_HOME"))).toPath().resolve("standalone");
+    private static final Path jbossServerBaseDir = new File(System.getProperty("jboss.home", System.getenv("JBOSS_HOME"))).toPath().resolve("standalone");
```
jmesnil, line 61: if that's a constant, we should rename it to `JBOSS_SERVER_BASE_DIR`

testsuite/manualmode/src/test/java/org/jboss/as/test/manualmode/management/persistence/RemoteSshGitRepositoryTestCase.java
```diff
@@ -171,10 +174,19 @@ public static void setUp() throws Exception {
...
+        PathUtil.copyRecursively(getJbossServerBaseDir().resolve("configuration"), backUpRoot, true);
+
+        RemoteSshGitRepositoryTestCase.backupRoot = backUpRoot;
+    }
+
     @AfterClass
     public static void afterClass() throws IOException {
         Security.removeProvider(CREDENTIAL_STORE_PROVIDER.getName());
         FileUtils.delete(CS_PUBKEY.toFile(), FileUtils.RECURSIVE | FileUtils.RETRY);
+        PathUtil.deleteRecursively(backupRoot);
```
jmesnil, line 189: should we also restore the configuration in the `@AfterClass` to avoid keeping the config from the last test for the rest of the test suite?

jmesnil (changes requested):

server/src/main/java/org/jboss/as/server/controller/git/ElytronClientCredentialsProvider.java
darranl: Server code.

server/src/main/java/org/jboss/as/server/controller/git/ElytronClientSshdSessionFactory.java
darranl: Server code.

server/src/main/java/org/jboss/as/server/logging/ServerLogger.java
darranl: I need to move on from this issue now, the logging I added was to get to the bottom of a test failure - I think other messages will help identify the path if needed.

server/src/main/java/org/jboss/as/server/logging/ServerLogger.java
darranl: IMO these should not have an ID - these messages are used for the commit messages in the git repository so similar to how we don't have message IDs for internationalised CLI messages we don't have message IDs for internationalised commit messages.

testsuite/manualmode/src/test/java/org/jboss/as/test/manualmode/management/persistence/RemoteSshGitRepositoryTestCase.java
darranl: The configuration is automatically restored after each individual test on line 260 at the end of the tearDown method.

The @BeforeClass and @AfterClass methods are focused on taking a backup once at the start of the test run and deleting that backup at the end of the test run with the @After method handling the individual clean up after every test.

darranl:
I think we need to think about how far we want to go with these kinds of review, the majority of this change is related to some opportunities I have seen to improve the logging situation of something that is difficult to debug which could also be a problem if we need to debug this remotely.

The issue here is it is not an are of code I am not likely to directly support myself.  Two weeks after submitting my PR to clean something up where I have seen some opportunities to improve logging I suddenly receive a lot of comments which are largely cosmetic.

Just as I did whilst working on the changes the module definition for the classes with the changed visibility could have been opened to see we already declare all these classes as private API:

```
<module xmlns="urn:jboss:module:1.8" name="org.jboss.as.server">
    <properties>
        <property name="jboss.api" value="private"/>
        <property name="jboss.require-java-version" value="1.8"/>
    </properties>
```

Regarding the name of an existing variable in the test suite that should have been static all along, sure it is correct it should be in upper case but it is also a private test variable.

The discussion about test clean up is worthwhile but it is quick to confirm that the clean up already happens after each test.

One point not picked up was the Temp commit that had crept in so I am quickly addressing that now.

But overall with PRs such as this I think we really need to ask if the code changes have improved the codebase or introduced a problem, ideally all engineers should be able to quickly submit PRs to clean things up especially around missing logging - but if these come back to you two weeks later so you have to context switch back to them it risks the individual engineer asking the question "Is it worth MY time to clean this up?".

_Note:  This is not the first PR like this, this is just the most recent one where I have stopped my planned work to come back to._

kabir:
Retest this please

jmesnil (approved):

PR #4353, opened 2020-10-06
[WFCORE-5151] dir-contexts referral-mode attribute upper/lower case discrepancy

rmartinc:
Issue: https://issues.redhat.com/browse/WFCORE-5151

In [ELY-1637](https://issues.redhat.com/browse/ELY-1637)/[WFCORE-3971](https://issues.redhat.com/browse/WFCORE-3971) the values for the `referral-mode` were changed to be in lower case to match the documentation and XSD. This PR is just to also change the allowed values and the default value (they now keep using `name()` instead `toString()` and are upper case). QA team filed the issue to be consistent. The real value for the attribute is case insensitive since the beginning so I think nothing more is needed (no transformer) and the change is just cosmetic.

### DIFF
--- a/elytron/src/main/java/org/wildfly/extension/elytron/DirContextDefinition.java
+++ b/elytron/src/main/java/org/wildfly/extension/elytron/DirContextDefinition.java
@@ -106,8 +106,8 @@ class DirContextDefinition extends SimpleResourceDefinition {
             .build();
 
     static final SimpleAttributeDefinition REFERRAL_MODE = new SimpleAttributeDefinitionBuilder(ElytronDescriptionConstants.REFERRAL_MODE, ModelType.STRING, true)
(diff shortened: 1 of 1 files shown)


rmartinc:
@bstansberry @darranl Could you take a look to this one once wildfly 21 is out? I think the change is just cosmetic but the more eyes the better.

rmartinc:
@bstansberry @darranl Can you take a look to this PR when you have time? If I need to rebase this just let me know.

bstansberry:
I think the change is fine. I don't know whether the subsystem API version has already been bumped for WF 22 (or for sure will be). If not I'll defer to @darranl or @fjuma as to whether they are willing to change these values without an API version bump.

darranl:
Our subsystem version has now been bumped anyway for WildFly 22.

bstansberry (approved):

jmesnil:
@rmartinc thanks

PR #4356, opened 2020-10-08
 [WFCORE-2687] Addition of getCurrentOperationName() and getCurrentOperationParameter(String name) methods

lvydra:
https://issues.redhat.com/browse/WFCORE-2687

### DIFF
--- a/controller/src/main/java/org/jboss/as/controller/AbstractOperationContext.java
+++ b/controller/src/main/java/org/jboss/as/controller/AbstractOperationContext.java
@@ -541,6 +541,41 @@ abstract class AbstractOperationContext implements OperationContext {
         return pa.getLastElement().getValue();
     }
 
+    @Override
+    public final ModelNode getCurrentOperationName() {
+        assert activeStep != null;
+        ModelNode operation = activeStep.operation;
+
+        assert operation != null;
+        return operation.get(NAME);
+    }
+
+    @Override
+    public final ModelNode getCurrentOperationParameter(final String parameterName) {
+        return getCurrentOperationParameter(parameterName, true);
+    }
+
+    @Override
+    public final ModelNode getCurrentOperationParameter(final String parameterName, boolean nullable) {
+        if (isLegalParameterName(parameterName)) {
+            assert activeStep != null;
+            ModelNode operation = activeStep.operation;
+
+            assert operation != null;
+            if (!operation.has(parameterName) && nullable) {
+                return null;
+            } else {
+                return operation.get(parameterName);
+            }
+        } else {
+            throw new IllegalArgumentException(ControllerLogger.ROOT_LOGGER.invalidParameterName(parameterName));
+        }
+    }
+
+    private boolean isLegalParameterName(final String parameterName) {
+        return !(parameterName.equals(NAME) || parameterName.equals(OP) || parameterName.equals(OP_ADDR) || parameterName.equals(OPERATION_HEADERS));
+    }
+
     /**
      * Notification that all steps in a stage have been executed.
      * <p>This default implementation always returns {@code true}.</p>
--- a/controller/src/main/java/org/jboss/as/controller/AbstractOperationContext.java
+++ b/controller/src/main/java/org/jboss/as/controller/AbstractOperationContext.java
@@ -542,12 +542,12 @@ abstract class AbstractOperationContext implements OperationContext {
     }
 
     @Override
-    public final ModelNode getCurrentOperationName() {
+    public final String getCurrentOperationName() {
         assert activeStep != null;
         ModelNode operation = activeStep.operation;
 
         assert operation != null;
-        return operation.get(NAME);
+        return operation.get(NAME).asString();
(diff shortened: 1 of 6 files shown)


controller/src/main/java/org/jboss/as/controller/AbstractOperationContext.java
```diff
@@ -541,6 +541,41 @@ public final String getCurrentAddressValue() {
...
+            } else {
+                return operation.get(parameterName);
+            }
+        } else {
+            throw new IllegalArgumentException(ControllerLogger.ROOT_LOGGER.invalidParameterName(parameterName));
+        }
+    }
+
+    private boolean isLegalParameterName(final String parameterName) {
+        return !(parameterName.equals(NAME) || parameterName.equals(OP) || parameterName.equals(OP_ADDR) || parameterName.equals(OPERATION_HEADERS));
```
bstansberry, line 576: @lvydra Please remove 'parameterName.equals(NAME) 'from this. My apologies; I should not have listed 'name' as an invalid key in the JIRA description. I've corrected the JIRA description.

controller/src/main/java/org/jboss/as/controller/logging/ControllerLogger.java
```diff
@@ -3629,4 +3629,7 @@ OperationFailedRuntimeException capabilityAlreadyRegisteredInContext(String capa
 
     @Message(id = 476, value = "Value for attribute '%s' is invalid: either '%s' must be specified on its own or '%s' needs to be specified with at least one of '%s' or '%s'")
     OperationFailedException invalidCredentialReferenceValue(String attributeName, String clearTextAttributeName, String credentialStoreAttributeName, String clearTextAttributeName1, String alias);
+
+    @Message(id = 477, value = "Parameter name '%s' is invalid.")
+    OperationFailedException invalidParameterName(String parameterName);
```
bstansberry, line 3634: The return type here should be IllegalArgumentException, as that is what you intend to throw.

controller/src/main/java/org/jboss/as/controller/AbstractOperationContext.java
```diff
@@ -541,6 +541,41 @@ public final String getCurrentAddressValue() {
...
+            ModelNode operation = activeStep.operation;
+
+            assert operation != null;
+            if (!operation.has(parameterName) && nullable) {
+                return null;
+            } else {
+                return operation.get(parameterName);
+            }
+        } else {
+            throw new IllegalArgumentException(ControllerLogger.ROOT_LOGGER.invalidParameterName(parameterName));
```
bstansberry, line 571: ControllerLogger.ROOT_LOGGER.invalidParameterName should return IllegalArgumentException so you can just throw what it returns without wrapping.

controller/src/main/java/org/jboss/as/controller/OperationContext.java
```diff
@@ -418,6 +418,30 @@
...
+     *
+     * @return operation name node
+     */
+    ModelNode getCurrentOperationName();
+
+    /**
+     * Get parameter node by its name
+     *
+     * @param name of desired parameter
+     * @return node for parameter of given name
```
bstansberry, line 432: This should also say 'or {@code null} if the operation has no such parameter'.

controller/src/main/java/org/jboss/as/controller/OperationContext.java
```diff
@@ -418,6 +418,30 @@
      */
     String getCurrentAddressValue();
 
+    /**
+     * Get the node with current operation name
+     *
+     * @return operation name node
+     */
+    ModelNode getCurrentOperationName();
```
bstansberry, line 426: This should return String. The node will always hold a string.

controller/src/main/java/org/jboss/as/controller/OperationContext.java
```diff
@@ -418,6 +418,30 @@
...
+     * @param name of desired parameter
+     * @return node for parameter of given name
+     */
+    ModelNode getCurrentOperationParameter(String name);
+
+    /**
+     * Get parameter node by its name
+     *
+     * @param name of desired parameter
+     * @param nullable whether return value can be null
```
bstansberry, line 440: whether the return value can be null if the operation has no parameter with the given name

controller/src/main/java/org/jboss/as/controller/OperationContext.java
```diff
@@ -418,6 +418,30 @@
...
+     * @return node for parameter of given name
```
bstansberry, line 441: node for the parameter of the given name, or {@code null} if {@code nullable} is {@code true} and the operation has no parameter with the given name

bstansberry (changes requested):

lvydra:
Hi @bstansberry , thank you for your review. I have added requested changes.

bstansberry (approved):

bstansberry:
Thanks, @lvydra !

PR #4369, opened 2020-10-15
[WFCORE-5146] WFLYPRT0057 message handling improvement

khermano:
Issue: https://issues.redhat.com/browse/WFCORE-5146

### DIFF
--- a/protocol/src/main/java/org/jboss/as/protocol/logging/ProtocolLogger.java
+++ b/protocol/src/main/java/org/jboss/as/protocol/logging/ProtocolLogger.java
@@ -22,6 +22,7 @@
 
 package org.jboss.as.protocol.logging;
 
+import static org.jboss.logging.Logger.Level.DEBUG;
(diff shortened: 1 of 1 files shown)


bstansberry:
I don't see how this is an improvement. Cancelling these tasks should be an unusual event and it is helpful to have information about it in the log. It's true that it's possible no cancellation message will get logged if the cancel happens after run() checks if it was cancelled but before it caches the executing thread. But that seems like a very low probability thing and not logging at all doesn't seem better.  Is this causing some kind of problem?

khermano:
As I understand from discussion in JBEAP issue, this ticket was filed with the customer's strong demand. He wanted to get rid of this message, but to change it like that was also a possibility.
Please have a quick look here https://issues.redhat.com/browse/JBEAP-18854.

khermano:
Closing this PR as upstream issue JBEAP-18854 has been closed.

PR #4370, opened 2020-10-19
[WFCORE-5163] Don't reject EE 9 permissions.xml files

bstansberry:
https://issues.redhat.com/browse/WFCORE-5163


jmesnil:
@bstansberry thanks

PR #4373, opened 2020-10-21
[WFCORE-5013] Remove the WildFly OpenSSL linux-i386, windows-i386, and solaris-x86_64 natives

fjuma:
https://issues.redhat.com/browse/WFCORE-5013

Note that a [proposal](https://lists.jboss.org/archives/list/wildfly-dev@lists.jboss.org/thread/7W5B2K4C2QXWBVVSZPHDBXDY3CW4JRA4/) for removing these natives was sent to the WildFly Dev mailing list a few weeks ago. We haven't received any concerns so we are going to proceed with removing them. Anyone running on these platforms who needs these natives will still be able to build and make use of them on their own. I've created a PR against WildFly that adds documentation on how to do this:

https://github.com/wildfly/wildfly/pull/13667

### DIFF
--- a/core-feature-pack/common/pom.xml
+++ b/core-feature-pack/common/pom.xml
@@ -549,11 +549,6 @@
 
         <dependency>
             <groupId>org.wildfly.openssl</groupId>
-            <artifactId>wildfly-openssl-linux-i386</artifactId>
-        </dependency>
-
-        <dependency>
-            <groupId>org.wildfly.openssl</groupId>
             <artifactId>wildfly-openssl-linux-s390x</artifactId>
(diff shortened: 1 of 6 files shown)


darranl (approved):

darranl:
@bstansberry I think this would be worth a quick review from you as well but at the moment producing the builds for these architectures is the slowest part of our release process so our reasoning is anyone who needs this binaries will also by definition have access to an environment they could use to add them back themselves.

bstansberry (approved):

bstansberry:
@darranl ok with me.

PR #4375, opened 2020-10-23
[WFCORE-5167] Use individual ApacheDS dependencies to avoid beinging in additional shaded content.

darranl:
https://issues.redhat.com/browse/WFCORE-5167

### DIFF
--- a/elytron/pom.xml
+++ b/elytron/pom.xml
@@ -183,10 +183,84 @@
         </dependency>
 
         <dependency>
-            <groupId>org.apache.directory.server</groupId>
-            <artifactId>apacheds-all</artifactId>
-            <version>2.0.0-M24</version>  <!--TODO Elytron Remove once parent on AM26+ -->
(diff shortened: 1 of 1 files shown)


jamezp (commented):
@darranl What is the reasoning for directly using the versions in the test? While I'm personally not opposed as they are just test dependencies, we generally always use the parent property.

darranl:
@jamezp The reasoning here is the version was already defined here to avoid a conflict with any parent version.  A Jira issue has been raised to clean up ApacheDS dependencies further but this change is not the time to refactoring the use of ApacheDS across the whole testsuite.

jamezp:
@darranl Okay. What we've done before is add a separate property. @jmesnil do you have a preference?

darranl:
This pom already had the versions inlined and not in a separate property, changes like refactoring into properties can come under the next Jira.

jmesnil:
@darranl thanks

PR #4377, opened 2020-10-23
[WFCORE-5145] Elytron server-ssl-context allowed protocols

SoniaZaldana:
Jira: https://issues.redhat.com/browse/WFCORE-5145

To pass CI, this will require an Elytron upgrade. 

Analysis document: https://github.com/wildfly/wildfly-proposals/pull/338
Requires: https://issues.redhat.com/browse/ELY-2024
Elytron PR with tests: https://github.com/wildfly-security/wildfly-elytron/pull/1458
WildFly PR with tests and documentation updates: https://github.com/wildfly/wildfly/pull/13673


elytron/src/main/java/org/wildfly/extension/elytron/ElytronDescriptionConstants.java
```diff
@@ -532,6 +532,7 @@
     String SSL_CONTEXT = "ssl-context";
     String SSL_CONTEXT_REGISTRATION = "ssl-context-registration";
     String SSL_SESSION = "ssl-session";
+    String SSL_V2HELLO = "SSLv2Hello";
```
fjuma, line 535: Very minor nit, I think `SSL_V2_HELLO` reads a bit better than `SSL_V2HELLO`.

elytron/src/main/java/org/wildfly/extension/elytron/ElytronSubsystemTransformers.java
```diff
@@ -119,7 +119,14 @@ public void registerTransformers(SubsystemTransformerRegistration registration)
 
     private static void from12(ChainedTransformationDescriptionBuilder chainedBuilder) {
         ResourceTransformationDescriptionBuilder builder = chainedBuilder.createBuilder(ELYTRON_12_0_0, ELYTRON_11_0_0);
-
+        builder.addChildResource(PathElement.pathElement(ElytronDescriptionConstants.SERVER_SSL_CONTEXT))
+                .getAttributeBuilder()
+                .addRejectCheck(new RejectAttributeChecker.SimpleAcceptListAttributeChecker(new ModelNode(ElytronDescriptionConstants.SSL_V2HELLO)), ElytronDescriptionConstants.PROTOCOLS)
```
fjuma, line 124: Looks like the `ListRejectAttributeChecker` could be used here instead.

elytron/src/test/java/org/wildfly/extension/elytron/TlsTestCase.java
```diff
@@ -368,6 +368,27 @@ public void testSslServiceAuthTLS13() throws Throwable {
...
+    @Test
+    public void testSslServiceAuthSSLv2Hello() throws Throwable {
+        Assume.assumeFalse("Skipping testSslServiceAuthSSLv2Hello as IBM JDK does not support enabling SSLv2Hello " +
+                "in the client", JdkUtils.isIbmJdk());
+        testCommunication("ServerSslContextSSLv2Hello", "ClientSslContextSSLv2Hello", false, "OU=Elytron,O=Elytron,C=CZ,ST=Elytron,CN=localhost",
+                "OU=Elytron,O=Elytron,C=UK,ST=Elytron,CN=Firefly");
+    }
+
+    @Test
+    public void testSslServiceAuthProtocolMismatchSSLv2Hello() throws Throwable {
```
fjuma, line 380: It would also be good to add a test for the case where SSLv2Hello is only configured on the server side.

fjuma (changes requested):

elytron/src/test/java/org/wildfly/extension/elytron/TlsTestCase.java
SoniaZaldana: Added

controller/src/main/java/org/jboss/as/controller/logging/ControllerLogger.java
```diff
@@ -3629,4 +3629,7 @@ OperationFailedRuntimeException capabilityAlreadyRegisteredInContext(String capa
 
     @Message(id = 476, value = "Value for attribute '%s' is invalid: either '%s' must be specified on its own or '%s' needs to be specified with at least one of '%s' or '%s'")
     OperationFailedException invalidCredentialReferenceValue(String attributeName, String clearTextAttributeName, String credentialStoreAttributeName, String clearTextAttributeName1, String alias);
+
+    @Message(id = 477, value = "The following attributes must not include %s in the current model: %s")
+    String attributesMustNotInclude(ModelNode value, Set<String> names);
```
fjuma, line 3634: This can be removed now since it's unused.

elytron/src/test/java/org/wildfly/extension/elytron/TlsTestCase.java
fjuma: Looks like it would be good to add one more test case that makes use of the default protocols on the server side and the client side to ensure that SSLv2Hello isn't enabled by default.

fjuma: It would also be good to either tweak `testCommunication` or create a new method that could be used instead in order to assert that `serverSocket.getEnabledProtocols()`, `clientSocket.getEnabledProtocols()`, `serverSocket.getSession().getProtocol()`, and `clientSocket.getSession().getProtocol()` return the expected values.

fjuma (changes requested):

elytron/src/test/java/org/wildfly/extension/elytron/TlsTestCase.java
SoniaZaldana: I've modified testCommunication to check the configured protocols.

elytron/src/test/resources/org/wildfly/extension/elytron/tls-ibm.xml
```diff
@@ -111,6 +111,14 @@
                                 key-manager="ServerKeyManager" trust-manager="CaTrustManager" want-client-auth="true" need-client-auth="true"
                                 authentication-optional="false" use-cipher-suites-order="false" providers="openssl"
                                 session-timeout="321" maximum-session-cache-size="123"/>
+            <server-ssl-context name="ServerSslContextSSLv2Hello" protocols="TLSv1 SSLv2Hello"
```
fjuma, line 114: I think double check that the same additions made in `tls-sun.xml` are also made in `tls-ibm.xml` and `tls-oracle13plus.xml`.xml. Notice that in `TlsTestCase#prepare`, we can see that one of these three test files will be used depending on the JDK version. So for any resources that we're adding for new test cases, we need to make sure that we add the same ones to all three test files.

SoniaZaldana: I've updated the files so that the same additions are added across all three files.

fjuma: I think tls-ibm.xml is missing "SeverSslContextSSLv2HelloOpenSsl".

fjuma: Actually, ignore that last comment, I can see that the test that makes use of that SSL context doesn't get run on IBM JDK. :)

fjuma (approved):
Looks good, thanks @SoniaZaldana!

elytron/src/main/java/org/wildfly/extension/elytron/ElytronSubsystemTransformers.java
```diff
@@ -127,6 +127,14 @@ private static void from13(ChainedTransformationDescriptionBuilder chainedBuilde
...
     private static void from12(ChainedTransformationDescriptionBuilder chainedBuilder) {
         ResourceTransformationDescriptionBuilder builder = chainedBuilder.createBuilder(ELYTRON_12_0_0, ELYTRON_11_0_0);
+        builder.addChildResource(PathElement.pathElement(ElytronDescriptionConstants.SERVER_SSL_CONTEXT))
+                .getAttributeBuilder()
+                .addRejectCheck(new RejectAttributeChecker.ListRejectAttributeChecker(new RejectAttributeChecker.SimpleRejectAttributeChecker( new ModelNode(ElytronDescriptionConstants.SSL_V2_HELLO))), ElytronDescriptionConstants.PROTOCOLS)
+                .end();
+        builder.addChildResource(PathElement.pathElement(ElytronDescriptionConstants.CLIENT_SSL_CONTEXT))
+                .getAttributeBuilder()
+                .addRejectCheck(new RejectAttributeChecker.ListRejectAttributeChecker(new RejectAttributeChecker.SimpleRejectAttributeChecker( new ModelNode(ElytronDescriptionConstants.SSL_V2_HELLO))), ElytronDescriptionConstants.PROTOCOLS)
+                .end();
```
fjuma, line 137: This block should be moved to the new `from13` method.

jmesnil:
this PR is out of sync with the main branch.
I'm closing it. Please reopen it when it is up to date and can be built.

If it can not be built (when a component upgrade is missing), please open it as a draft.

darranl:
Converted to draft whilst waiting for approval to include in the next Elytron tag.

darranl:
@SoniaZaldana it looks like this one also has a couple of conflicts to address.

fjuma (approved):

PR #4378, opened 2020-10-26
[WFCORE-3825] Unify -server option in windows scripts (ps1, bat) for domain mode (and its servers) on all JDKs

lvydra:
https://issues.redhat.com/browse/WFCORE-3825

### DIFF
--- a/core-feature-pack/common/src/main/resources/content/bin/common.ps1
+++ b/core-feature-pack/common/src/main/resources/content/bin/common.ps1
@@ -308,6 +308,41 @@ Function Start-WildFly-Process {
 	Env-Clean-Up
 }
 
+# Returns java -version output for JDK defined by java opts
+Function Get-Java-Version ($javaOpts) {
+    if ($javaOpts -match '-d64') {
+        $opt_version = '-d64'
+    } elseif ($javaOpts -match '-d32') {
+        $opt_version = '-d32'
+    }
+
+    $result = & $JAVA $opt_version -version 2>&1
+    if ($LASTEXITCODE -eq 1) {
+        $result = & $JAVA -version 2>&1
+    }
+
+    return $result
+}
+
+# Check if can set option to use HotSpot VM
+Function Set-Java-Server-Option ($javaOpts) {
+    if ( -Not ($javaOpts -match '-server') ) {
+        # Check user requested JDK 'data model'
+        $version = Get-Java-Version ($javaOpts)
+
+        # Check if data model is supported by JDK
+        # Check for SUN(tm) JVM w/ HotSpot support
+        # Check for OpenJDK JVM w/server support
+        if ($LASTEXITCODE -eq 1) {
+            Write-Host $version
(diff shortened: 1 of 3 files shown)


jamezp:
Maybe I'm missing the idea of the JIRA, but I don't see `-server` being added to the `domain.bat` or `domain.ps1`.

lvydra:
Hi @jamezp ,
thank you for your response.
As I get it, the point of issues is to make the behavior of sh, bat, and ps1 scripts consistent. So in this case to add, to the bat and ps1 scripts, functionality, which will add -server option to Java VM Options in the same/similar way as the sh script does at :

https://github.com/wildfly/wildfly-core/blob/4d836539c7b1fd8f9900c26652df046bfca96752/core-feature-pack/common/src/main/resources/content/bin/domain.sh#L108-L138

core-feature-pack/common/src/main/resources/content/bin/common.ps1
```diff
@@ -308,6 +308,41 @@ Function Start-WildFly-Process {
...
+    if ( -Not ($javaOpts -match '-server') ) {
+        # Check user requested JDK 'data model'
+        $version = Get-Java-Version ($javaOpts)
+
+        # Check if data model is supported by JDK
+        # Check for SUN(tm) JVM w/ HotSpot support
+        # Check for OpenJDK JVM w/server support
+        if ($LASTEXITCODE -eq 1) {
+            Write-Host $version
+        } elseif ($version -match 'HotSpot' -or $version -match 'OpenJDK') {
```
jamezp, line 338: It looks like we might want to test for IBM too.
```
# Check for IBM JVM w/server support
if [ "x$HAS_IBM" = "x" ]; then
    HAS_IBM=`"$JAVA" $JVM_OPTVERSION 2>&1 | $GREP -i "IBM J9"`
fi
```

core-feature-pack/common/src/main/resources/content/bin/domain.bat
```diff
@@ -103,14 +103,38 @@ call "!DIRNAME!common.bat" :setDefaultModularJvmOptions "!PROCESS_CONTROLLER_JAV
...
+
+    :CHECK_VERSION
+    "%JAVA%" !JVM_OPTVERSION! 2>&1 | findstr /I /C:"Unrecognized option" > nul
+    if not errorlevel == 1 (
+        set JVM_OPTVERSION=-version
+    )
+
+    rem Add -server to the JVM options, if supported
+    :CHECK_SUPPORTED
+    "%JAVA%" !JVM_OPTVERSION! 2>&1 | findstr /I "hotspot openJDK" > nul
```
jamezp, line 131: This may also need to check for IBM J9.

lvydra:
Hi @jamezp , thank you for your review. I have added IBM J9 check.

jamezp (approved):

bstansberry:
Thanks @lvydra. Sorry; one of us mergers should have merged this long ago.

PR #4382, opened 2020-11-03
[WFCORE-5181] Expression properties with trailing whitespaces are resolved to a trimmed value

jbaesner:
Issue: https://issues.redhat.com/browse/WFCORE-5181

### DIFF
--- a/controller/src/main/java/org/jboss/as/controller/parsing/ParseUtils.java
+++ b/controller/src/main/java/org/jboss/as/controller/parsing/ParseUtils.java
@@ -542,8 +542,12 @@ public final class ParseUtils {
         final String trimmed = value == null ? null : value.trim();
         ModelNode node;
         if (trimmed != null) {
-            if (isExpressionAllowed && isExpression(trimmed)) {
-                node = new ModelNode(new ValueExpression(trimmed));
+            if (isExpressionAllowed && isExpression(value)) {
+                if(attributeType == STRING || attributeType == PROPERTY) {
+                    node = new ModelNode(new ValueExpression(value));
+                } else {
+                    node = new ModelNode(new ValueExpression(trimmed));
+                }
             } else {
                 if(attributeType == STRING || attributeType == PROPERTY) {
                     node = new ModelNode().set(value);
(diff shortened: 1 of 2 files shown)


spyrkob:
@jmesnil could you review this when you have a moment please?

controller/src/main/java/org/jboss/as/controller/parsing/ParseUtils.java
```diff
@@ -565,8 +565,8 @@ public static ModelNode parseAttributeValue(final String value, final boolean is
         final String trimmed = value == null ? null : value.trim();
         ModelNode node;
         if (trimmed != null) {
-            if (isExpressionAllowed && isExpression(trimmed)) {
-                node = new ModelNode(new ValueExpression(trimmed));
+            if (isExpressionAllowed && isExpression(value)) {
+                node = new ModelNode(new ValueExpression(value));
```
bstansberry, line 569: This should do the same test that the non-expression else block does, and use 'trimmed' unless attributeType is STRING or PROPERTY. An expression that's meant to resolve to a string convertible to some other type shouldn't have leading/trailing whitespace and we shouldn't stop being helpful about that.

bstansberry (changes requested):

controller/src/test/java/org/jboss/as/controller/parsing/ParseUtilsTestCase.java
```diff
@@ -0,0 +1,51 @@
...
+ *
+ * @author Joerg Baesner (c) 2020 Red Hat Inc.
+ */
+public class ParseUtilsTestCase {
+
+    @Test
+    public void testExpressionsWithTrailingWhitespace() {
+
+        String propertyValue = "${key} ";
+        ModelType attributeType = ModelType.EXPRESSION;
```
bstansberry, line 41: @jbaesner I believe this should be ModelType.STRING

jbaesner: Yes, exactly, you're right!

controller/src/main/java/org/jboss/as/controller/parsing/ParseUtils.java
jbaesner: Ok, done...

bstansberry (approved):

bstansberry:
Thanks, @jbaesner !

PR #4384, opened 2020-11-04
[WFCORE-482] Add log4j2 support. Added the log4j-api and an implement…

jamezp:
…ation that uses the jboss-logmanager as it's delegating log manager.

### DIFF
--- a/logging/src/main/java/org/jboss/as/logging/LoggingModuleDependency.java
+++ b/logging/src/main/java/org/jboss/as/logging/LoggingModuleDependency.java
@@ -1,0 +1,86 @@
+/*
+ * JBoss, Home of Professional Open Source.
+ *
+ * Copyright 2020 Red Hat, Inc., and individual contributors
+ * as indicated by the @author tags.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.jboss.as.logging;
+
+import org.jboss.as.controller.registry.RuntimePackageDependency;
+
+/**
+ * Describes a module dependency.
+ *
+ * @author <a href="mailto:jperkins@redhat.com">James R. Perkins</a>
+ */
+public enum LoggingModuleDependency {
+    APACHE_COMMONS_LOGGING("org.apache.commons.logging"),
+    LOG4J("org.apache.log4j"),
+    LOG4J2("org.apache.logging.log4j.api", true),
+    JBOSS_LOGGING("org.jboss.logging"),
+    JUL_TO_SLF4J("org.jboss.logging.jul-to-slf4j-stub"),
+    JBOSS_LOG_MANAGER("org.jboss.logmanager"),
+    SLF4J("org.slf4j"),
+    SLF4J_IMPL("org.slf4j.impl"),
+    ;
+
+    private static final RuntimePackageDependency[] RUNTIME_DEPENDENCIES = {
+            RuntimePackageDependency.optional(APACHE_COMMONS_LOGGING.moduleName),
+            RuntimePackageDependency.optional(LOG4J2.moduleName),
+            RuntimePackageDependency.optional(SLF4J.moduleName),
+            RuntimePackageDependency.optional(JUL_TO_SLF4J.moduleName),
+            RuntimePackageDependency.passive(SLF4J_IMPL.moduleName),
+    };
+
+    private final String moduleName;
+    private final boolean importServices;
+
+    LoggingModuleDependency(final String moduleName) {
+        this(moduleName, false);
+    }
+
+    LoggingModuleDependency(final String moduleName, final boolean importServices) {
+        this.moduleName = moduleName;
+        this.importServices = importServices;
+    }
+
+    /**
+     * The module name.
+     *
+     * @return the module name
+     */
+    public String getModuleName() {
+        return moduleName;
+    }
+
+    /**
+     * Indicates whether or not services should be imported.
+     *
+     * @return {@code true} if services from the module should be imported
+     */
+    public boolean isImportServices() {
+        return importServices;
+    }
+
+    /**
+     * An array of the runtime dependencies.
+     *
+     * @return the runtime dependencies
+     */
+    public static RuntimePackageDependency[] getRuntimeDependencies() {
+        return RUNTIME_DEPENDENCIES;
+    }
+}
--- a/logging/src/main/java/org/jboss/as/logging/stdio/LogContextStdioContextSelector.java
+++ b/logging/src/main/java/org/jboss/as/logging/stdio/LogContextStdioContextSelector.java
@@ -22,6 +22,9 @@
 
 package org.jboss.as.logging.stdio;
 
+import java.security.AccessController;
+import java.security.PrivilegedAction;
+
 import org.jboss.as.logging.CommonAttributes;
 import org.jboss.logmanager.Level;
 import org.jboss.logmanager.LogContext;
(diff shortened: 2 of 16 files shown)


jamezp:
I'm not too sure why the `CommandTimeoutHandlerTestCase` is failing. I tested on a local Windows VM and it passed.

logging/src/main/java/org/jboss/as/logging/LoggingModuleDependency.java
```diff
@@ -0,0 +1,88 @@
...
+/**
+ * Describes a module dependency.
+ *
+ * @author <a href="mailto:jperkins@redhat.com">James R. Perkins</a>
+ */
+public enum LoggingModuleDependency {
+    APACHE_COMMONS_LOGGING("org.apache.commons.logging"),
+    LOG4J("org.apache.log4j"),
+    LOG4J2("org.apache.logging.log4j.api"),
+    LOG4J2_IMPL("org.jboss.logmanager.log4j2", true),
```
boris-unckel, line 33: @jamezp I was wondering why you put in the LOG4J2_IMPL org.jboss.logmanager.lo4j2 as dependency. In [JCL](https://github.com/wildfly/wildfly-core/blob/master/core-feature-pack/common/src/main/resources/modules/system/layers/base/org/apache/commons/logging/main/module.xml) it is "This allows the org.jboss.logging.commons.logging module to be private while keeping this module public"
I think the Apache Commons Logging way, to provide the API but not the Impl is quite good. It prevents app developers to depend on that classes.

logging/src/main/java/org/jboss/as/logging/LoggingModuleDependency.java
```diff
@@ -0,0 +1,88 @@
...
+    JBOSS_LOGGING("org.jboss.logging"),
+    JUL_TO_SLF4J("org.jboss.logging.jul-to-slf4j-stub"),
+    JBOSS_LOG_MANAGER("org.jboss.logmanager"),
+    SLF4J("org.slf4j"),
+    SLF4J_IMPL("org.slf4j.impl"),
```
boris-unckel, line 38: @jamezp Same for SLF4_IMPL - I would prefer to keep it private/unreachable.

logging/src/main/java/org/jboss/as/logging/LoggingModuleDependency.java
jamezp: If we don't add it here we have to add a circular dependency in the `org.apache.logging.log4j.api` module so it knows where to find the implementation. I suppose another option would be to create a `org.apache.logging.log4j` module and make the add that as a dependency to the deployment exporting only the API.

boris-unckel: Yes, you're right. I'll change my backports in the same way.

pom.xml
```diff
@@ -190,6 +191,7 @@
         <version.org.jboss.logging.jul-to-slf4j-stub>1.0.1.Final</version.org.jboss.logging.jul-to-slf4j-stub>
         <version.org.jboss.logmanager.jboss-logmanager>2.1.17.Final</version.org.jboss.logmanager.jboss-logmanager>
         <version.org.jboss.logmanager.log4j-jboss-logmanager>1.2.0.Final</version.org.jboss.logmanager.log4j-jboss-logmanager>
+        <version.org.jboss.logmanager.log4j2-jboss-logmanager>1.0.0.Beta3</version.org.jboss.logmanager.log4j2-jboss-logmanager>
```
tommaso-borgato, line 194: @jamezp, @boris-unckel  is it ok to use a "Beta" version for log4j2-jboss-logmanager?

testsuite/standalone/src/test/java/org/jboss/as/test/integration/logging/profiles/Log4j2LoggingProfilesTestCase.java
```diff
@@ -0,0 +1,58 @@
...
+@ServerSetup(AbstractLoggingProfilesTestCase.LoggingProfilesTestCaseSetup.class)
+public class Log4j2LoggingProfilesTestCase extends AbstractLoggingProfilesTestCase {
+
+    public Log4j2LoggingProfilesTestCase() {
+        super(Log4j2ServiceActivator.class, 2);
+    }
+
+    @Override
+    protected void processDeployment(final JavaArchive deployment) {
+        final Permission[] permissions = {
```
tommaso-borgato, line 45: @jamezp , @boris-unckel why are these permissions needed here? I don't see them in akin tests like `Log4jLoggingProfilesTestCase` .... is the customer required to grant these permissions to the deployment?

pom.xml
jamezp: @tommaso-borgato nope, but I've upgraded it to final now :)

testsuite/standalone/src/test/java/org/jboss/as/test/integration/logging/profiles/Log4j2LoggingProfilesTestCase.java
jamezp: @tommaso-borgato I tried to explain the best I could in the comments. These are all because of the way the log4j2 API loads things so we have no control over it. I've made a note to update the future documentation for this. This documentation change will not be part of this RFE because it's going to be a large change and I don't want to hold this RFE up based a large documentation re-do.

jamezp: I just created https://issues.redhat.com/browse/WFLY-14160 to track the documentation changes as a whole.

logging/src/main/java/org/jboss/as/logging/stdio/LogContextStdioContextSelector.java
```diff
@@ -50,16 +53,37 @@ public StdioContext getStdioContext() {
...
         StdioContext stdioContext = root.getAttachment(STDIO_CONTEXT_ATTACHMENT_KEY);
         if (stdioContext == null) {
-            stdioContext = StdioContext.create(
-                    new NullInputStream(),
-                    new LoggingOutputStream(logContext.getLogger("stdout"), Level.INFO),
-                    new LoggingOutputStream(logContext.getLogger("stderr"), Level.ERROR)
-            );
-            final StdioContext appearing = root.attachIfAbsent(STDIO_CONTEXT_ATTACHMENT_KEY, stdioContext);
+            // Create the StdioContext possibly using a privileged action. This is required for scenarios where a
+            // deployment or library may attempt to write to System.out or System.err.
```
bstansberry, line 57: @jamezp As we discussed in chat please follow up later with a more detailed comment as it's not obvious why this is needed and why it's safe.

Partly I may be confused because I'm not sure what in the first sentence the "This is required" in the second sentence refers to: "Create the StdioContext" or "using a privileged action".

testsuite/standalone/src/test/java/org/jboss/as/test/integration/logging/profiles/Log4j2LoggingProfilesTestCase.java
```diff
@@ -0,0 +1,58 @@
...
+                // The getClassLoader permissions is required for the org.apache.logging.log4j.util.ProviderUtil.
+                new RuntimePermission("getClassLoader"),
+                // The FilePermissions is also for the org.apache.logging.log4j.util.ProviderUtil as it needs to read the JAR
+                // for the service loader.
+                new FilePermission("<<ALL FILES>>", "read"),
```
bstansberry, line 50: @jamezp When you work on WFLY-14160 you might want to experiment with something like

```
new FilePermission("${jboss.home.dir}/modules/*", "read"),
```

Expression resolution in a permissions.xml should work, and if it does that's a much better thing to document than suggesting giving ALL FILES.

logging/src/main/java/org/jboss/as/logging/stdio/LogContextStdioContextSelector.java
jamezp: Yes will do. The issue in this case was this blocked me from seeing the other log4j2 security issues because it attempted to write to `System.out` which triggered this error so I lost messages. I should have been more clear in the description.

testsuite/standalone/src/test/java/org/jboss/as/test/integration/logging/profiles/Log4j2LoggingProfilesTestCase.java
jamezp: Ah good to know. I wasn't aware that expressions worked. That will be a better approach than `<<ALL FILES>>`. I really didn't like using that.

bstansberry: Yeray added it earlier this year.

bstansberry (approved):

PR #4389, opened 2020-11-12
[WFCORE-5191] Enforce reuseForks to true in Scripts module of TS

marekkopecky:
Jira with more details: https://issues.redhat.com/browse/WFCORE-5191

This PR enforce reuseForks to true in Scripts module of TS, because reuseForks=false doesn't work on Scripts module of TS

### DIFF
--- a/testsuite/scripts/pom.xml
+++ b/testsuite/scripts/pom.xml
@@ -79,6 +79,7 @@
                     <skip>${skipTests}</skip>
                     <failIfNoTests>false</failIfNoTests>
                     <redirectTestOutputToFile>${testLogToFile}</redirectTestOutputToFile>
+                    <reuseForks>true</reuseForks>
 
                     <systemPropertyVariables>
                         <jboss.home>${wildfly.home}</jboss.home>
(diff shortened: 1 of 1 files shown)


marekkopecky:
@jamezp Can you please review this PR and possibly merge it?

jamezp:
@marekkopecky I'm a bit confused on the issue here. The default is `true` and it's not overridden to `false` in this pom. Is there a way to reproduce the issue?

marekkopecky:
This PR fixes the following use-case:

1. I want to start whole WF-core TS, but I know that time-to-time I see intermittent failures caused by missing cleanup etc. So I started TS with `-DreuseForks=false` 
2. `mvn install  -DallTests -DreuseForks=false`
3. Almost all TS modules (like `standalone` or `manualmode`) works correctly with `-DreuseForks=false` option, so I can really benefit from reuseForks=false settings (I have more stable results, although TS run takes a little more time because each testcase has new thread). But `scripts` module is not ready for `-DreuseForks=false` . I see `WFLYCTL0212: Duplicate resource` exceptions in all testcases from `scripts` module except first testcase, because these test-case (in new threads) wants to initialize management model settings even if it has already been initialized).
5. Now I have two options:
* enforce reuseForks=true in scripts module
* check whether model is already initialized in each testcase (this means that embedded server needs to be started in prepare phase of each tests if `reuseForks=false`, which is quite expensive)
6. I apply this PR (first option from previous step)
7. `mvn install  -DallTests -DreuseForks=false`
8. All test-cases from script module passed.

jamezp (approved):

PR #4392, opened 2020-11-20
[WFCORE-5198] Ignore deployment failures such as duplicate classes

scottmarlow:
Signed-off-by: Scott Marlow <smarlow@redhat.com>
https://issues.redhat.com/browse/WFCORE-5198

### DIFF
--- a/server/src/main/java/org/jboss/as/server/deployment/DeploymentAddHandler.java
+++ b/server/src/main/java/org/jboss/as/server/deployment/DeploymentAddHandler.java
@@ -53,6 +53,7 @@ import org.jboss.as.controller.descriptions.ModelDescriptionConstants;
 import org.jboss.as.controller.registry.Resource;
 import org.jboss.as.protocol.StreamUtils;
 import org.jboss.as.repository.ContentRepository;
+import org.jboss.as.server.controller.resources.DeploymentAttributes;
 import org.jboss.as.server.deployment.transformation.DeploymentTransformer;
 import org.jboss.as.server.logging.ServerLogger;
 import org.jboss.as.server.services.security.AbstractVaultReader;
@@ -217,7 +218,7 @@ public class DeploymentAddHandler implements OperationStepHandler {
         InputStream transformed = null;
         try {
             try {
-                transformed = deploymentTransformer == null ? in : deploymentTransformer.transform(in, name);
+                transformed = deploymentTransformer == null ? in : doTransformation(context, contentItemNode, name, in);
                 hash = contentRepository.addContent(transformed);
             } catch (IOException e) {
                 throw createFailureException(e.toString());
@@ -228,6 +229,25 @@ public class DeploymentAddHandler implements OperationStepHandler {
             StreamUtils.safeClose(transformed);
         }
         return new DeploymentHandlerUtil.ContentItem(hash);
+    }
+
+    private InputStream doTransformation(OperationContext context, ModelNode contentItemNode, String name, InputStream in) throws IOException, OperationFailedException {
+        try {
+            return deploymentTransformer.transform(in, name);
+        } catch (RuntimeException t) {
+            // Check if the InputStream is already attached to the operation request (as per CONTENT_INPUT_STREAM_INDEX check) and ignore that case
+            // as calling getInputStream would of returned the already partially consumed InputStream.
+            // Also verify that the thrown exception is the specific WFCORE-5198 `Error code 3`.
+            if (!contentItemNode.hasDefined(DeploymentAttributes.CONTENT_INPUT_STREAM_INDEX.getName()) &&
+                    t.getCause() != null && t.getCause().getCause() != null &&
+                    t.getCause().getCause() instanceof IOException &&
+                    t.getCause().getCause().getMessage().contains("during transformation. Error code 3")) {
+                ServerLogger.ROOT_LOGGER.tracef(t, "Ignoring transformation error and using original archive %s", name);
+                return getInputStream(context, contentItemNode);
+            } else {
+                throw t;
+            }
+        }
     }
 
     DeploymentHandlerUtil.ContentItem addUnmanaged(ModelNode contentItemNode) throws OperationFailedException {


scottmarlow:
@bstansberry in case of transformation failures that are not caused by IO exceptions (e.g. we catch all others via catch(Throwable t), we will return the hash of the original (untransformed) content instead.  The caller then uses the hash of the original content to locate the original deployment archive.  

In the case of a successful transformation, we return the hash of the (newly transformed) application archive so that the new generated archive is used instead of the original content.  

In order for this change to work, the transformation must not write over the input deployment archive but instead must write to a new output file.  If we ever do change transformation to overwrite the original application, this change will break, as part of the transformation may of occurred leaving the application in a bad (partially transformed) state.  At least this is how it looks to me.

scottmarlow:
I'm not sure of all of the possible exceptions that Throwable may be catching but am thinking we should check specifically for the target failure, instead of all Exceptions.

Also since we are eating the exception, we probably should DEBUG or TRACE log the exception.

bstansberry:
@scottmarlow This will only work with certain deployment operations. It works with the TCK deployment because the TCK uses the deployment scanner, and the scanner uses the 'url' parameter to pass content:

https://github.com/wildfly/wildfly-core/blob/master/deployment-scanner/src/main/java/org/jboss/as/server/deployment/scanner/FileSystemDeploymentService.java#L1468

If you look at getInputStream, it is safe to call it twice for 'url' content (and for the little used 'bytes').  In both of those cases getInputStream itself creates and returns an InputStream. Call it twice and it will just create a 2nd InputStream.

But for the very common 'if (contentItem.hasDefined(DeploymentAttributes.CONTENT_INPUT_STREAM_INDEX.getName()))' case, this will not work. For that case, getInputStream is just finding and returning an input stream attached to the operation request, and that stream will have been at least partially consumed. It's not reusable.

 DeploymentAttributes.CONTENT_INPUT_STREAM_INDEX is what is used when either the CLI or HAL deploy content.

That said, this approach could be a partial workaround, if we just exclude deployments that use DeploymentAttributes.CONTENT_INPUT_STREAM_INDEX from the retry.

bstansberry:
> In order for this change to work, the transformation must not write over the input deployment archive but instead must write to a new output file. If we ever do change transformation to overwrite the original application, this change will break, as part of the transformation may of occurred leaving the application in a bad (partially transformed) state. At least this is how it looks to me.

The transformer must never overwrite the original application, and it doesn't. It is provided an IS and has no idea where the original application is. It reads the IS, creates a tmp file, and returns an IS to read that temp file. The content repository then reads that IS to store the application in $WILDFLY_HOME/standalone/data/content.

server/src/main/java/org/jboss/as/server/deployment/DeploymentAddHandler.java
```diff
@@ -219,8 +219,17 @@ private static String getRuntimeName(String name, ModelNode deployment) {
             try {
                 transformed = deploymentTransformer == null ? in : deploymentTransformer.transform(in, name);
                 hash = contentRepository.addContent(transformed);
-            } catch (IOException e) {
+            } catch (IOException e) { // fail hard on IO error
                 throw createFailureException(e.toString());
+            } catch(Throwable t) {  // for other errors, return original file stream
```
bstansberry, line 224: I assume what we are really trying to catch is a RuntimeException thrown by EclipseTransformer, not a java.lang.Error. So this catch should be limited to RuntimeException.

Beyond a general code standard thing of avoiding suppressing Error, this matters because if we bypass trying again for deployments that use DeploymentAttributes.CONTENT_INPUT_STREAM_INDEX, then for those we are going to have to rethrow the exception.

bstansberry:
@scottmarlow If we get to the point where we agree we want to go with this approach, please move WFLY-14104 to WFCORE and change the commit message and PR title.

server/src/main/java/org/jboss/as/server/deployment/DeploymentAddHandler.java
bstansberry: Also, I think this catch should be for a try block that only includes L220.  We want to react to failures in deploymentTransformer.transform, not to anything that happens with L221's contentRepository.addContent.

Creating a method for the L220 and the try/catch would be cleaner looking.

scottmarlow:
> @scottmarlow This will only work with certain deployment operations. It works with the TCK deployment because the TCK uses the deployment scanner, and the scanner uses the 'url' parameter to pass content:
> 
> https://github.com/wildfly/wildfly-core/blob/master/deployment-scanner/src/main/java/org/jboss/as/server/deployment/scanner/FileSystemDeploymentService.java#L1468
> 
> If you look at getInputStream, it is safe to call it twice for 'url' content (and for the little used 'bytes'). In both of those cases getInputStream itself creates and returns an InputStream. Call it twice and it will just create a 2nd InputStream.
> 
> But for the very common 'if (contentItem.hasDefined(DeploymentAttributes.CONTENT_INPUT_STREAM_INDEX.getName()))' case, this will not work. For that case, getInputStream is just finding and returning an input stream attached to the operation request, and that stream will have been at least partially consumed. It's not reusable.

Thanks, I missed that!

> 
> DeploymentAttributes.CONTENT_INPUT_STREAM_INDEX is what is used when either the CLI or HAL deploy content.
> 
> That said, this approach could be a partial workaround, if we just exclude deployments that use DeploymentAttributes.CONTENT_INPUT_STREAM_INDEX from the retry.

+1, I'll give that a try.

scottmarlow:
@bstansberry thanks, I made the suggested changes.  I'm running all of the servlet TCK tests locally to verify the fix.

server/src/main/java/org/jboss/as/server/deployment/DeploymentAddHandler.java
```diff
@@ -230,6 +231,25 @@ private static String getRuntimeName(String name, ModelNode deployment) {
...
+        try {
+            return deploymentTransformer.transform(in, name);
+        } catch (RuntimeException t) {
+            // Check if the InputStream is already attached to the operation request (as per CONTENT_INPUT_STREAM_INDEX check) and ignore that case
+            // as calling getInputStream would of returned the already partially consumed InputStream.
+            // Also verify that the thrown exception is the specific WFCORE-5198 `Error code 3`.
+            if (!contentItemNode.hasDefined(DeploymentAttributes.CONTENT_INPUT_STREAM_INDEX.getName()) &&
+                    t.getCause() != null && t.getCause().getCause() != null &&
+                    t.getCause().getCause() instanceof IOException &&
+                    t.getCause().getCause().getMessage().contains("during transformation. Error code 3")) {
```
bstansberry, line 244: I think it will be ok to retry with the untransformed deployment no matter what the exception was but it's ok to start off being more restrictive.

bstansberry (approved):

jmesnil:
@scottmarlow thanks

PR #4394, opened 2020-11-20
WFCORE-5126 WARN if user tries to set unrealistic io subsystem stack-…

RanabirChakraborty:
Issue: https://issues.redhat.com/browse/WFCORE-5126

### DIFF
--- a/io/subsystem/src/main/java/org/wildfly/extension/io/logging/IOLogger.java
+++ b/io/subsystem/src/main/java/org/wildfly/extension/io/logging/IOLogger.java
@@ -75,4 +75,8 @@ public interface IOLogger extends BasicLogger {
 
     @Message(id = 7, value = "Unexpected bind address conflict in resource \"%s\" when attempting to establish binding for destination %s to %s: a binding of %s already existed")
     OperationFailedException unexpectedBindAddressConflict(PathAddress currentAddress, CidrAddress cidrAddress, InetSocketAddress bindAddress, InetSocketAddress existing);
+
+    @LogMessage(level = WARN)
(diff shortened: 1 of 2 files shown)


io/subsystem/src/main/java/org/wildfly/extension/io/logging/IOLogger.java
```diff
@@ -75,4 +75,7 @@
 
     @Message(id = 7, value = "Unexpected bind address conflict in resource \"%s\" when attempting to establish binding for destination %s to %s: a binding of %s already existed")
     OperationFailedException unexpectedBindAddressConflict(PathAddress currentAddress, CidrAddress cidrAddress, InetSocketAddress bindAddress, InetSocketAddress existing);
+
+    @Message(id = 8, value = "Please proivde a realistic stack-size value")
```
bstansberry, line 79: s/proivde/provide

Also, please pass in the value they provided and include it in the message. "The stack-size value of %d bytes for IO worker %s is low and may result in problems. A value of at least 150,000 is recommended."

Also this needs the @LogMessage annotation and the method return type should be 'void'. That's how methods that produce log messages are handled. Methods without that are not for log messages; they return objects that include localized strings and then the code that calls the method does something with the object.

bstansberry (changes requested):

io/subsystem/src/main/java/org/wildfly/extension/io/logging/IOLogger.java
RanabirChakraborty: Thank you for the review :) I have changed it accordingly.

gaol:
@bstansberry could you please check again if it is OK? thanks

bstansberry (approved):

jmesnil:
@RanabirChakraborty thanks

PR #4418, opened 2020-12-04
Fix for WFCORE-5206, Remove workaround to access ModuleLoader.installMBeanServer

jfdenise:
Issue: https://issues.redhat.com/browse/WFCORE-5206
Removed reflection calls to access ModuleLoader.installMBeanServer that is now public.

PR #4420, opened 2020-12-04
[WFCORE-5206] Remove workaround for ModuleLoader.installMBeanServer() method invocation

ropalka:
https://issues.redhat.com/browse/WFCORE-5206


ropalka:
Duplicates https://github.com/wildfly/wildfly-core/pull/4418

PR #4421, opened 2020-12-04
[WFCORE-5216] jboss-cli.sh script should use eval instead of exec

istudens:
https://issues.redhat.com/browse/WFCORE-5216

### DIFF
--- a/core-feature-pack/common/src/main/resources/content/bin/jboss-cli.sh
+++ b/core-feature-pack/common/src/main/resources/content/bin/jboss-cli.sh
@@ -80,6 +80,9 @@ JAVA_OPTS="$JAVA_OPTS -Dcom.ibm.jsse2.overrideDefaultTLS=true"
 # Sample JPDA settings for remote socket debugging
 #JAVA_OPTS="$JAVA_OPTS -agentlib:jdwp=transport=dt_socket,address=8787,server=y,suspend=n"
 
+# WFCORE-5216 - evaluate any eventual env variables
+JBOSS_MODULEPATH=$(eval echo \"${JBOSS_MODULEPATH}\")
+
 LOG_CONF=`echo $JAVA_OPTS | grep "logging.configuration"`
 if [ "x$LOG_CONF" = "x" ]; then
     exec "$JAVA" $JAVA_OPTS -Dlogging.configuration=file:"$JBOSS_HOME"/bin/jboss-cli-logging.properties -jar "$JBOSS_HOME"/jboss-modules.jar -mp "${JBOSS_MODULEPATH}" org.jboss.as.cli "$@"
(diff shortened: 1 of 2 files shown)


core-feature-pack/common/src/main/resources/content/bin/jboss-cli.sh
```diff
@@ -82,8 +82,15 @@ JAVA_OPTS="$JAVA_OPTS -Dcom.ibm.jsse2.overrideDefaultTLS=true"
 
 LOG_CONF=`echo $JAVA_OPTS | grep "logging.configuration"`
 if [ "x$LOG_CONF" = "x" ]; then
-    exec "$JAVA" $JAVA_OPTS -Dlogging.configuration=file:"$JBOSS_HOME"/bin/jboss-cli-logging.properties -jar "$JBOSS_HOME"/jboss-modules.jar -mp "${JBOSS_MODULEPATH}" org.jboss.as.cli "$@"
+    eval \"$JAVA\" $JAVA_OPTS \
```
jfdenise, line 85: We are using exec due to the fix for https://issues.redhat.com/browse/JBEAP-8808.
We can't use eval there. At least in the general case.

jfdenise (changes requested):
I don't think that the reported problem should be fixed in the jboss-cli script.sh. I asked for more information in the JIRA issue.

DavyKoravand:
Hi, I would like to add that we also experienced jboss-cli problems due to exec being used after we updated WildFly from v10 to v21. We had issues using JAVA_OPTS parameters that used spaces, even though they were correctly encapsulated by quotes (we tried every escape/quote combination). We currently use a workaround where we make sure that JAVA_OPTS is empty when using the jboss-cli, but it would be nice to see eval being used again.

jfdenise:
@DavyKoravand , could you provide some JAVA_OPTS values that create problems? Thank-you.

jfdenise:
@istudens , instead of using eval to launch the CLI process, we should call the following just before to call exec:
JBOSS_MODULEPATH=$(eval "echo $JBOSS_MODULEPATH")
JAVA_OPTS=$(eval "echo $JAVA_OPTS")

istudens:
@jfdenise ok, I will change the PR accordingly

jfdenise:
@istudens , it happens that the problem reported with JAVA_OPTS will be fixed differently. Could you update your PR by just adding:
JBOSS_MODULEPATH=$(eval "echo $JBOSS_MODULEPATH")
and forget about the JAVA_OPTS.

istudens:
I had to toss the testcase modification out as it wasn't working on Windows platform.

jfdenise:
@istudens , could you make the addition to the test ignored on Windows and kept on other platforms?

istudens:
I've finally get to the correct `eval` syntax which now even works with special characters in JBOSS_MODULEPATH like '(' etc. However, it is still `eval` which means it might be hacked to execute any arbitrary command if JBOSS_MODULEPATH is set by 
an evil user accordingly.

jfdenise (approved):
Thank-you, approved.

jmesnil:
@istudens thanks

PR #4424, opened 2020-12-08
[WFCORE-482] Improve the comment on why the privileged action is requ…

jamezp:
…ired. Also reduce the file permissions to the log4j2-jboss-logmanager instead of ALL FILES.

Follows up on #4384

### DIFF
--- a/testsuite/standalone/src/test/java/org/jboss/as/test/integration/logging/profiles/Log4j2LoggingProfilesTestCase.java
+++ b/testsuite/standalone/src/test/java/org/jboss/as/test/integration/logging/profiles/Log4j2LoggingProfilesTestCase.java
@@ -20,6 +20,7 @@
 package org.jboss.as.test.integration.logging.profiles;
 
 import java.io.FilePermission;
+import java.nio.file.Paths;
 import java.security.Permission;
 import java.util.PropertyPermission;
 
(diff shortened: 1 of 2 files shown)


testsuite/standalone/src/test/java/org/jboss/as/test/integration/logging/profiles/Log4j2LoggingProfilesTestCase.java
```diff
@@ -47,12 +48,27 @@ protected void processDeployment(final JavaArchive deployment) {
...
+                new FilePermission(resolveFilePermissions(), "read"),
                 // Required for the EnvironmentPropertySource System.getenv().
                 new RuntimePermission("getenv.*"),
                 // Required for the SystemPropertiesPropertySource System.getProperties().
                 new PropertyPermission("*", "read,write"),
         };
         addPermissions(deployment, permissions);
     }
+
+    private static String resolveFilePermissions() {
```
bstansberry, line 60: Does this testsuite support using -Djboss.dist to run against a different distribution? If so I don't think this will work as that dist would probably be a fat server.

jamezp: Oh right no it won't. I forgot that was an option. I'll see if there is a way to make both work, otherwise we'll need `<<ALL FILES>>`.

bstansberry: @jamezp yeah, sorry; i led you totally in a wrong direction.

My focus on my comment on this line on the other PR was more about what to document in the docs, not so much this test. This test would be a great way to prove that something you document actually works, but I forgot the permissions.xml expression resolution one use of which you'd be documenting does not work unless the server is full WildFly.

jamezp: @bstansberry Actually it doesn't look like the `jboss.dist` is honored for core. I'm gonna search through the pom's, but I see this in the system properties:
```
jboss.dist = /home/jperkins/projects/wildfly/wildfly-core/dist/target/wildfly-core-14.0.0.Beta4-SNAPSHOT
module.path = /home/jperkins/projects/wildfly/wildfly-core/testsuite/standalone/target/wildfly-core/modules
```

The surefire config sets the jboss.home https://github.com/wildfly/wildfly-core/blob/a8975aa124f914d38af4314fe1b6221d589ccd2a/testsuite/pom.xml#L325. Only the Elytron profile sets the `module.path`, but somehow the test still passed then.

jamezp: Oh the `module.path` is only overridden for the `wildfly-maven-plugin` to execute CLI commands. So it doesn't look like it will ever be overridden for core tests. Is this something separate I should look at fixing?

bstansberry: I have no idea :)

I'd say if we can find something that doesn't do perform the function it should then we can fix it!

PR #4431, opened 2020-12-16
[WFCORE-5232] Use a new process to delete the install directory. JBos…

jamezp:
…s Modules holds locks on the JAR's which results in a failure to delete the install directory on Windows. By using a new process we wait until the main process has died and we delete the files in the new process avoiding the files being locked.

https://issues.redhat.com/browse/WFCORE-5232

@jfdenise This is my proposal to the issue with deleting the Windows install directory. What this will do is for Windows by default it will launch a new process which deletes the directory. For all other operating systems it will just delete the directory in the same process. Note this can be overridden with the `org.wildfly.core.jar.cleanup.newProcess` system property.

There is also a slight change to the extraction of the container from the bootable JAR. It simply checks for the existence of a `wildfly-cleanup-marker` file in the install directory and waits for a settable timeout before it just continues. This is to avoid possible issues where a directory is being deleted, but the restart of the bootable JAR happens too fast.

### DIFF
--- a/bootable-jar/runtime/src/main/java/org/wildfly/core/jar/runtime/BootableJar.java
+++ b/bootable-jar/runtime/src/main/java/org/wildfly/core/jar/runtime/BootableJar.java
@@ -23,14 +23,16 @@ import java.io.InputStream;
 import java.io.InputStreamReader;
 import java.lang.management.ManagementFactory;
 import java.nio.charset.StandardCharsets;
-import java.nio.file.FileVisitResult;
 import java.nio.file.Files;
 import java.nio.file.Path;
-import java.nio.file.SimpleFileVisitor;
-import java.nio.file.attribute.BasicFileAttributes;
+import java.nio.file.StandardOpenOption;
 import java.util.ArrayList;
+import java.util.Collections;
 import java.util.List;
 import java.util.Properties;
+import java.util.concurrent.ExecutorService;
+import java.util.concurrent.Executors;
+import java.util.concurrent.TimeUnit;
 import javax.xml.parsers.DocumentBuilder;
 import javax.xml.parsers.DocumentBuilderFactory;
 import javax.xml.transform.Transformer;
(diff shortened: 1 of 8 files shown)


bootable-jar/runtime/src/main/java/org/wildfly/core/jar/runtime/InstallationCleaner.java
```diff
@@ -0,0 +1,196 @@
...
+        final Path java;
+        if (environment.isWindows()) {
+            java = javaHome.resolve("bin").resolve("java.exe");
+        } else {
+            java = javaHome.resolve("bin").resolve("java");
+        }
+        if (Files.exists(java)) {
+            return java.toString();
+        }
+        return "java";
```
jfdenise, line 176: Should it be java.exe for Windows?

jamezp: I thought about that, but I know on Windows from CMD `java` works. That said `java.home` should never be a missing property and this is a fallback that should never happen. I should probably just verify the full path exists and then fall back though to be safe.

bootable-jar/runtime/src/main/java/org/wildfly/core/jar/runtime/BootableJar.java
```diff
@@ -233,21 +230,15 @@ private LogContext configureLogContext() throws IOException {
...
-
-    private void cleanup() {
-        log.deletingHome(environment.getJBossHome());
-        deleteDir(environment.getJBossHome());
+        server = buildServer(startServerArgs);
 
+        if (Files.notExists(pidFile)) {
+            Files.write(pidFile, Collections.singleton(Long.toString(org.wildfly.common.os.Process.getProcessId())), StandardCharsets.UTF_8, StandardOpenOption.CREATE_NEW);
+        } else {
+            log.pidFileAlreadyExists(pidFile, environment.getJBossHome());
```
jamezp, line 239: Maybe instead of just logging this should throw an exception. What do you think @jfdenise?

jfdenise: I agree, that seems an invalid state. We should start from an empty directory.

jmesnil:
@jamezp thanks!

PR #4435, opened 2021-01-03
[WFCORE-5233] Improve ability for WildFly Core consumers to use its Galleon source without depending on the wildfly-core feature pack

bstansberry:
https://issues.redhat.com/browse/WFCORE-5233


core-feature-pack/galleon-pruned/pom.xml
```diff
@@ -0,0 +1,82 @@
...
+        <groupId>org.wildfly.core</groupId>
+        <artifactId>wildfly-core-feature-pack-parent</artifactId>
+        <!--
+        Maintain separation between the artifact id and the version to help prevent
+        merge conflicts between commits changing the GA and those changing the V.
+        -->
+        <version>14.0.0.Beta5-SNAPSHOT</version>
+    </parent>
+
+    <artifactId>wildfly-core-feature-pack-galleon-legacy</artifactId>
```
jmesnil, line 37: could you add a <name> to better identify this module?

I'm not sure for consistency sake if it should be rather named `wildfly-core-feature-pack-galleon-pruned`.
I interpret "legacy" feature pack as our previous feature pack model (before Galleon).

core-feature-pack/galleon-common/src/main/resources/feature_groups/core-basic-profile.xml
```diff
@@ -1,5 +1,5 @@
 <?xml version="1.0" encoding="UTF-8"?>
-<feature-group-spec name="basic-profile" xmlns="urn:jboss:galleon:feature-group:1.0">
+<feature-group-spec name="core-basic-profile" xmlns="urn:jboss:galleon:feature-group:1.0">
```
jmesnil, line 2: the rule of thumb is that feature-groups which share the same name in WildFly (Full) and WildFly Core should be renamed to `core-xxx` to disambiguate. Is that correct?

core-feature-pack/galleon-pruned/pom.xml
bstansberry: Done

core-feature-pack/galleon-common/src/main/resources/feature_groups/core-basic-profile.xml
bstansberry: Yes. 

A simple way to think of it is all these shared galleon files end up being unzipped or copied into the same target/resources folder, so the same file in WF will overwrite the one from core. So to pick up the content from core either the core file needs a unique name, or all the content from the core file would have to be duplicated in the WF file.

bstansberry:
@jmesnil I updated this to change the GAV of the new module.  I also added an additional commit to move a package out of the shared galleon stuff.

jmesnil (approved):

jmesnil:
@bstansberry thanks

PR #4447, opened 2021-01-07
[WFCORE-5240] Add a ManagementResourceRegistration.registerSubModelIf…

bstansberry:
…Absent method

https://issues.redhat.com/browse/WFCORE-5240


bstansberry:
I stuck the hold label on this as I think the use case needs a bit more of a think.

bstansberry:
I don't think this is a good idea; closing.

PR #4453, opened 2021-01-14
[WFCORE-5243] NullPointerException when invalid <permission> classes specified

rmartinc:
Issue: https://issues.redhat.com/browse/WFCORE-5243

A new deferred factory is used because some permissions cannot be constructed at parsing time (module is not ready at that time, that's why factories are used right now for the deployment). Besides this way a better warn message can be displayed showing what set gave the error (maximum, minimum or deployment).

### DIFF
--- a/security-manager/src/main/java/org/wildfly/extension/security/manager/DeferredPermissionFactory.java
+++ b/security-manager/src/main/java/org/wildfly/extension/security/manager/DeferredPermissionFactory.java
@@ -1,0 +1,98 @@
+/*
+* Copyright 2021 Red Hat, Inc.
+*
+* Licensed under the Apache License, Version 2.0 (the "License");
+* you may not use this file except in compliance with the License.
+* You may obtain a copy of the License at
+*
+*   http://www.apache.org/licenses/LICENSE-2.0
+*
+* Unless required by applicable law or agreed to in writing, software
+* distributed under the License is distributed on an "AS IS" BASIS,
+* WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+* See the License for the specific language governing permissions and
+* limitations under the License.
+*/
+
+package org.wildfly.extension.security.manager;
+
+import java.security.Permission;
+import org.jboss.modules.ModuleLoader;
+import org.jboss.modules.security.LoadedPermissionFactory;
+import org.jboss.modules.security.ModularPermissionFactory;
+import org.jboss.modules.security.PermissionFactory;
+import org.wildfly.extension.security.manager.logging.SecurityManagerLogger;
+import org.wildfly.security.manager.WildFlySecurityManager;
+
+/**
+ * <p>A deferred permission factory that logs a warning if the construction fails.</p>
+ *
+ * @author rmartinc
+ */
+public class DeferredPermissionFactory implements PermissionFactory {
+
+    private final Type type;
+    private final ModuleLoader moduleLoader;
+    private final String moduleName;
+    private final String permissionClass;
+    private final String permissionName;
+    private final String permissionActions;
+
+    private volatile PermissionFactory factory;
+
+    public enum Type {
+        DEPLOYMENT,
+        MAXIMUM_SET,
+        MINIMUM_SET,
+    };
+
+    public DeferredPermissionFactory(Type type, ModuleLoader moduleLoader, String moduleName, String permissionClass, String permissionName, String permissionActions) {
+        if (type == null) {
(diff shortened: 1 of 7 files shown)


jmesnil:
@darranl could you please review?

security-manager/src/main/java/org/wildfly/extension/security/manager/DeferredPermissionFactory.java
```diff
@@ -0,0 +1,98 @@
...
+import org.jboss.modules.security.PermissionFactory;
+import org.wildfly.extension.security.manager.logging.SecurityManagerLogger;
+import org.wildfly.security.manager.WildFlySecurityManager;
+
+/**
+ * <p>A deferred permission factory that logs a warning if the construction fails.</p>
+ *
+ * @author rmartinc
+ */
+public class DeferredPermissionFactory implements PermissionFactory {
```
darranl, line 32: Is this only referenced from the same package?  If so it should not be public but could it be better in the Elytron project?  We would prefer the WildFly Core code to just contain the code required for management,

darranl:
This seems to be a change in behaviour where the original intent appears to be aborting the boot sequence but now we have to rely on log messages.

I agree Stage.MODEL validation is not the time for class loading validation but why not during Stage.RUNTIME?  Is this running into an issue with the deployments ClassLoader?

rmartinc:
> This seems to be a change in behaviour where the original intent appears to be aborting the boot sequence but now we have to rely on log messages.

The problem is that there are two different behaviors right now. With the min-set (standalone.xml) there is a NPE and the boot is aborted (note it's a NPE, not something controlled). It's the same with the deployment permissions.xml file, an invalid permission there aborts the deployment (NPE too). But with the max-set the invalid permissions are just ignored silently (jboss-modules does it).

I tried to unify the behavior and always ignore the invalid permission but logging a warning, the same for max, min or deployment. See [JBEAP-20798](https://issues.redhat.com/browse/JBEAP-20798) where there was a conversation about this.

I can continue throwing an exception in the min and deployment set if it's preferred.

> I agree Stage.MODEL validation is not the time for class loading validation but why not during Stage.RUNTIME? Is this running into an issue with the deployments ClassLoader?

For example the problem is in the deployment parsing. The factories are created and added to the module specification at parsing time [here](https://github.com/wildfly/wildfly-core/blob/master/security-manager/src/main/java/org/wildfly/extension/security/manager/deployment/PermissionsParser.java#L155) but the permission is finally instantiated at POST_MODULE [here](https://github.com/wildfly/wildfly-core/blob/master/security-manager/src/main/java/org/wildfly/extension/security/manager/deployment/PermissionsValidationProcessor.java#L71). The NPE should be controlled at this point but we cannot even create a good message from the factory. Other option is doing changes in jboss-modules to obtain more info about the permission being created.

security-manager/src/main/java/org/wildfly/extension/security/manager/DeferredPermissionFactory.java
rmartinc: Sadly it is used by two packages: org.wildfly.extension.security.manager and org.wildfly.extension.security.manager.deployment. The first one for max and min sets (standalone.xml) and the second one for the deployment set (permissions.xml).

darranl:
It have a complete idea yet but it does sound like different behaviour for minimum and maximum could be justified.  But yes the NPE is definitely a bug, anything that stops the boot should be controlled.

rmartinc:
OK @darranl, if you have a better idea just let me know. The current PR is just allowing the deployment/boot but warning a message in the three sets for any permission that returns null.

rmartinc:
@darranl I have been looking this again and the deferred factory is only needed for the deployment section. So we can do the following:

1. In the minimum and maximum sets (standalone) we continue using the previous {{LoadedPermissionFactory}}. We execute the construction before when we now the permission data to log a message or throw an exception. No performance issue as the permission is cached in the factory.
2. Then the new {{DeferredPermissionFactory}} can be only used for the deployment set (permissions.xml) and it can be not public inside that package.
3. Finally we have to decide if we prefer to just log a message as in the current PR or throw a {{OperationFailedException}} and {{DeploymentUnitProcessingException}} for the minimum and deployment set (not for maximum because that would break current configurations). I preferred the first option but I understand both are valid. Just let me know what you prefer.

What do you think?

darranl (approved):
I have just kicked off CI again as some jobs were red and been a while since the last run but I think we are good to go on this one.

PR #4455, opened 2021-01-15
[WFCORE-4516] Updated the description of 'core.management.security-realm.server-identity.secret.value' attribute.

ivassile:
https://issues.redhat.com/browse/WFCORE-4516

Updated the description of 'core.management.security-realm.server-identity.secret.value' attribute.


darranl (changes requested):
The text for the comment message and title in this PR don't seem to reflect the change being made.

ivassile:
@darranl Added description to the comment message. Should I also update the title or leave the jira title?

bstansberry:
I think @darranl meant "commit message" not "comment message". The commit message and the the title of the PR should reflect what the change did so the git history is understandable later.  The JIRA itself should change as well. Issue type becomes "Bug". Title becomes something like "Documentation says server-identities expressions should resolve to Base64 values, but only resolving to plain text works."

darranl:
Sorry @ivassile @bstansberry is correct I mean commit messages so when we come back to the source in 5 years time we can understand the history better.

ivassile:
Updated the commit message and jira.

bstansberry (approved):
Thanks, @ivassile !

PR #4458, opened 2021-01-22
[WFCORE-4291] Add support for non-graceful startup

jasondlee:
https://issues.redhat.com/browse/WFCORE-4291

Squashed commit of the following:
commit 874016e
Author: Brian Stansberry <brian.stansberry@redhat.com>
Date:   Mon Jul 29 13:50:48 2019 -0500

    [WFCORE-4291] Basic implementation of disabling initial boot-time suspension of a server.

Change name from "graceless" to "graceful-startup" and invert intention
Add support for '--graceful-startup' command line param for server startup
Modify schema to add graceful-startup attribute to server-group for domain mode support
Modify ManagedServerBootCmdFactory to pass arg as needed to managed servers

### DIFF
--- a/host-controller/src/main/java/org/jboss/as/domain/controller/transformers/ServerGroupTransformers.java
+++ b/host-controller/src/main/java/org/jboss/as/domain/controller/transformers/ServerGroupTransformers.java
@@ -47,29 +47,36 @@ class ServerGroupTransformers {
         //////////////////////////////////
         //The EAP/AS 7.x chains
 
-        // module-options was introduced in 13, WildFly 20
-        ResourceTransformationDescriptionBuilder currentTo13 = createBuilderFromCurrent(chainedBuilder, KernelAPIVersion.VERSION_13_0);
-        JvmTransformers.registerTransformers13_AndBelow(currentTo13);
+        // graceful-startup attribute was introduced in Kernel 16 (Wildfly 23)
+        ResourceTransformationDescriptionBuilder currentTo15 = createBuilderFromCurrent(chainedBuilder, KernelAPIVersion.VERSION_15_0)
+                .getAttributeBuilder()
+                .setDiscard(DiscardAttributeChecker.DEFAULT_VALUE, ServerGroupResourceDefinition.GRACEFUL_STARTUP)
+                .addRejectCheck(RejectAttributeChecker.DEFINED, ServerGroupResourceDefinition.GRACEFUL_STARTUP)
+                .end();
 
-        //timeout attribute renamed to suspend-timeout in Version 9.0. Must be renamed for 8.0 and below
+        // module-options was introduced in Kernel API 14 (WildFly Core 13, WildFly 21)
+        ResourceTransformationDescriptionBuilder builder15to13 = createBuilder(chainedBuilder, KernelAPIVersion.VERSION_15_0, KernelAPIVersion.VERSION_13_0);
+        JvmTransformers.registerTransformers13_AndBelow(builder15to13);
+
+        //timeout attribute renamed to suspend-timeout in Version Kernel 9.0 (WildFly Core 7, WildFly 15). Must be renamed for 8.0 and below
         ResourceTransformationDescriptionBuilder builder10to8 = createBuilder(chainedBuilder, KernelAPIVersion.VERSION_10_0, KernelAPIVersion.VERSION_8_0);
         DomainServerLifecycleHandlers.registerTimeoutToSuspendTimeoutRename(builder10to8);
 
-        // kill-servers and destroy-servers are rejected since 5.0 and below
+        // kill-servers and destroy-servers are rejected for Kernel 5.0 (WildFly Core 3, WildFly 11) and below
         ResourceTransformationDescriptionBuilder builder60to50 = createBuilder(chainedBuilder, KernelAPIVersion.VERSION_6_0, KernelAPIVersion.VERSION_5_0);
         DomainServerLifecycleHandlers.registerKillDestroyTransformers(builder60to50);
 
-        // The use of default-interface attribute in socket-binding-group is rejected since 1.8 and below
+        // The use of default-interface attribute in socket-binding-group is rejected for 1.8 and below
         ResourceTransformationDescriptionBuilder builder20to18 = createBuilder(chainedBuilder, KernelAPIVersion.VERSION_2_0, KernelAPIVersion.VERSION_1_8)
                 .getAttributeBuilder()
                 .setDiscard(DiscardAttributeChecker.UNDEFINED, ServerGroupResourceDefinition.SOCKET_BINDING_DEFAULT_INTERFACE)
                 .addRejectCheck(RejectAttributeChecker.DEFINED, ServerGroupResourceDefinition.SOCKET_BINDING_DEFAULT_INTERFACE)
                 .end();
 
-        //suspend and resume servers are rejected since 1.8 and below
+        //suspend and resume servers are rejected for Kernel 1.8 (EAP 6.4.0 CP07) and below
         DomainServerLifecycleHandlers.registerServerLifeCycleOperationsTransformers(builder20to18);
 
-        // The use of launch-command is rejected since 2.1 and below
+        // The use of launch-command is rejected since Kernel 2.1 (WildFly 8.1.0) and below
         ResourceTransformationDescriptionBuilder builder30To21 = createBuilder(chainedBuilder, KernelAPIVersion.VERSION_3_0, KernelAPIVersion.VERSION_2_1);
         JvmTransformers.registerTransformers2_1_AndBelow(builder30To21);
 
--- a/host-controller/src/main/resources/org/jboss/as/domain/controller/resources/LocalDescriptions.properties
+++ b/host-controller/src/main/resources/org/jboss/as/domain/controller/resources/LocalDescriptions.properties
@@ -76,6 +76,7 @@ server-group.add.jvm=The jvm name.
 server-group.remove=Remove an existing new server group.
 server-group.profile=The profile name.
 server-group.jvm=The named jvm.
+server-group.graceful-startup=Start the servers gracefully, queuing or cleanly rejecting incoming requests until the server is fully started.
 server-group.socket-binding-group=The default socket binding group used for servers associated with this group.
 server-group.socket-binding-default-interface=The socket binding group default interface for this server.
 server-group.socket-binding-port-offset=The default offset to be added to the port values given by the socket binding group.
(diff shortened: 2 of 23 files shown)


server/src/main/resources/schema/wildfly-config_16_0.xsd
```diff
@@ -2923,6 +2923,13 @@
                 </xs:documentation>
             </xs:annotation>
         </xs:attribute>
+        <xs:attribute name="graceful-startup" type="xs:boolean" default="true">
+            <xs:annotation>
+                <xs:documentation>
+                    Set to true to have servers in the group start gracefully, queuing incoming requests until the server is fully started.
```
bstansberry, line 2929: Perhaps some words describing more about what 'false' means?

Also, s/queueing/queueing or cleanly rejecting/g

host-controller/src/main/resources/org/jboss/as/domain/controller/resources/LocalDescriptions.properties
```diff
@@ -76,6 +76,7 @@ server-group.add.jvm=The jvm name.
 server-group.remove=Remove an existing new server group.
 server-group.profile=The profile name.
 server-group.jvm=The named jvm.
+server-group.graceful-startup=Start the servers in gracefully, queuing incoming requests until the server is fully started.
```
bstansberry, line 79: s/in gracefully/gracefully

s/queuing incoming/queueing or cleanly rejecting/g

server/src/main/java/org/jboss/as/server/ServerService.java
```diff
@@ -312,9 +312,14 @@ protected void boot(final BootContext context) throws ConfigurationPersistenceEx
             newExtDirs[extDirs.length] = new File(serverEnvironment.getServerBaseDir(), "lib/ext");
             serviceTarget.addService(org.jboss.as.server.deployment.Services.JBOSS_DEPLOYMENT_EXTENSION_INDEX,
                     new ExtensionIndexService(newExtDirs)).setInitialMode(ServiceController.Mode.ON_DEMAND).install();
-            final Boolean suspend = runningModeControl.getSuspend()!= null ? runningModeControl.getSuspend() : serverEnvironment.isStartSuspended();
+            final boolean suspend = runningModeControl.getSuspend()!= null ? runningModeControl.getSuspend() : serverEnvironment.isStartSuspended();
+            final boolean gracefulStartup = serverEnvironment.isStartGracefully();
+            assert !suspend || gracefulStartup;
```
bstansberry, line 317: This was a bad thing for me to have written. It might be ok but only if we've validated the user input elsewhere so this is basically asserting that that elsewhere did what it wanted. And TBH I don't think that 'validation elsewhere' is readily doable; this code is probably the place to validate and decide what to do.

What to do == TBD!

bstansberry: Re the TBD, we've been chatting and I think if 'suspend=true' and 'gracefulStartup=false' then 'suspend' should win, but with logging.

The reason is because gracefulStartup can only be set from the cmd line, while the start-mode can be changed during a reload. So there can be workflows (odd ones, but they exist) where the user wants the long term setting to be graceless but temporarily they want the server suspended. The basic use case for graceless is a multiple deployment scenario where app B needs to invoke on app A during start. So the odd workflows would involve cases where app B isn't deployed yet or has been undeployed, and the admin is doing stuff and reloading.

jmesnil:
@jasondlee Could you please check the issue related to this PR (and link it in the commit message)?

https://issues.redhat.com/browse/WFCORE-2491 does not seem to be the correct issue...

jmesnil:
https://issues.redhat.com/browse/WFCORE-4291 is the right one! ;)

host-controller/src/main/java/org/jboss/as/domain/controller/transformers/ServerGroupTransformers.java
```diff
@@ -47,9 +47,16 @@ static ChainedTransformationDescriptionBuilder buildTransformerChain() {
...
-        ResourceTransformationDescriptionBuilder currentTo13 = createBuilderFromCurrent(chainedBuilder, KernelAPIVersion.VERSION_13_0);
-        JvmTransformers.registerTransformers13_AndBelow(currentTo13);
+        // graceful-startup attribute was introduced in Kernel 16 (Wildfly 23)
+        ResourceTransformationDescriptionBuilder currentTo15 = createBuilderFromCurrent(chainedBuilder, KernelAPIVersion.VERSION_15_0)
+                .getAttributeBuilder()
+                .setDiscard(DiscardAttributeChecker.DEFAULT_VALUE, ServerGroupResourceDefinition.GRACEFUL_STARTUP)
+                .addRejectCheck(RejectAttributeChecker.DEFINED, ServerGroupResourceDefinition.GRACEFUL_STARTUP)
+                .end();
+
+        // module-options was introduced in 14, WildFly 20
```
bstansberry, line 57: This says 14, but the code says 15.

I'm pretty sure that's because the 14 means WildFly Core 14, which used Kernel API version 15. This stuff is confusing and some day someone is going to mess up and we won't notice.

Can you change this (and others in this file) something like:

'blah blah Kernel X (WildFly Core Y, WildFly Z)'

bstansberry (commented):
This LGTM other than a minor point about a code comment.

Squashing to 1 commit is good since AIUI the 2nd is fixing a bug in the 1st.

jasondlee:
> @jasondlee Could you please check the issue related to this PR (and link it in the commit message)?
> 
> https://issues.redhat.com/browse/WFCORE-2491 does not seem to be the correct issue...

Wow. That's been that way forever. Fat fingered that. Got it fixed. :)

host-controller/src/main/java/org/jboss/as/domain/controller/transformers/ServerGroupTransformers.java
```diff
@@ -47,15 +47,22 @@ static ChainedTransformationDescriptionBuilder buildTransformerChain() {
...
-        ResourceTransformationDescriptionBuilder currentTo13 = createBuilderFromCurrent(chainedBuilder, KernelAPIVersion.VERSION_13_0);
-        JvmTransformers.registerTransformers13_AndBelow(currentTo13);
+        // graceful-startup attribute was introduced in Kernel 16 (Wildfly 23)
+        ResourceTransformationDescriptionBuilder currentTo15 = createBuilderFromCurrent(chainedBuilder, KernelAPIVersion.VERSION_15_0)
+                .getAttributeBuilder()
+                .setDiscard(DiscardAttributeChecker.DEFAULT_VALUE, ServerGroupResourceDefinition.GRACEFUL_STARTUP)
+                .addRejectCheck(RejectAttributeChecker.DEFINED, ServerGroupResourceDefinition.GRACEFUL_STARTUP)
+                .end();
+
+        // module-options was introduced in Kernel 13 (WildFly Core 12, WildFly 20)
```
bstansberry, line 57: // module-options was introduced in Kernel API 14 (WildFly Core 13, WildFly 21)

host-controller/src/main/java/org/jboss/as/domain/controller/transformers/ServerGroupTransformers.java
```diff
@@ -47,15 +47,22 @@ static ChainedTransformationDescriptionBuilder buildTransformerChain() {
...
+        ResourceTransformationDescriptionBuilder builder15to13 = createBuilder(chainedBuilder, KernelAPIVersion.VERSION_15_0, KernelAPIVersion.VERSION_13_0);
+        JvmTransformers.registerTransformers13_AndBelow(builder15to13);
 
-        //timeout attribute renamed to suspend-timeout in Version 9.0. Must be renamed for 8.0 and below
+        //timeout attribute renamed to suspend-timeout in Version Kernel 9.0 (WildFly Core xx, WildFly 15). Must be renamed for 8.0 and below
```
bstansberry, line 61: xx == 7

host-controller/src/main/java/org/jboss/as/domain/controller/transformers/ServerGroupTransformers.java
```diff
@@ -47,15 +47,22 @@ static ChainedTransformationDescriptionBuilder buildTransformerChain() {
...
         ResourceTransformationDescriptionBuilder builder10to8 = createBuilder(chainedBuilder, KernelAPIVersion.VERSION_10_0, KernelAPIVersion.VERSION_8_0);
         DomainServerLifecycleHandlers.registerTimeoutToSuspendTimeoutRename(builder10to8);
 
-        // kill-servers and destroy-servers are rejected since 5.0 and below
+        // kill-servers and destroy-servers are rejected since Kernel 5.0 (WildFly Core xx, WildFly 11) and below
```
bstansberry, line 65: s/rejected since/rejected for/g

xx == 3

host-controller/src/main/java/org/jboss/as/domain/controller/transformers/ServerGroupTransformers.java
```diff
@@ -66,10 +73,10 @@ static ChainedTransformationDescriptionBuilder buildTransformerChain() {
                 .addRejectCheck(RejectAttributeChecker.DEFINED, ServerGroupResourceDefinition.SOCKET_BINDING_DEFAULT_INTERFACE)
                 .end();
 
-        //suspend and resume servers are rejected since 1.8 and below
+        //suspend and resume servers are rejected since Kernel 1.8 (EAP 6.4.0 CP07) and below
```
bstansberry, line 76: s/rejected since/rejected for/g

host-controller/src/main/java/org/jboss/as/domain/controller/transformers/ServerGroupTransformers.java
```diff
@@ -66,10 +73,10 @@ static ChainedTransformationDescriptionBuilder buildTransformerChain() {
...
         DomainServerLifecycleHandlers.registerServerLifeCycleOperationsTransformers(builder20to18);
 
-        // The use of launch-command is rejected since 2.1 and below
+        // The use of launch-command is rejected since Kernel 2.1 (WildFly Core xx, WildFly 8.1.0) and below
```
bstansberry, line 79: the WildFly Core xx can be dropped as WF Core did not exist yet (it hadn't been split out from main WF)

bstansberry (changes requested):

host-controller/src/main/java/org/jboss/as/domain/controller/transformers/ServerGroupTransformers.java
jasondlee: Updated

bstansberry (approved):

jmesnil:
Pre-check is done, I'm merging this PR.

@jasondlee Thanks!

PR #4461, opened 2021-01-25
[WFCORE-5254] Fix domain-management module OSHs that are populating t…

bstansberry:
…he runtime with data from the operation instead of the Resource tree

https://issues.redhat.com/browse/WFCORE-5254


yersan (approved):
At host-controller module level, StaticDiscoveryAddHandler and DiscoveryOptionAddHandler seems to have the similar potential bug, I'll file a Jira separate Jira to address them.

yersan:
`StaticDiscoveryAddHandler` and `DiscoveryOptionAddHandler` seem don't fit in this category. They work upon placeholder resources, which doesn't have a model associated.

bstansberry:
@yersan I think your instinct to file a JIRA about those handlers is a good one; at least a minor priority task to look into them. That code is pretty unusual and it's good to make it clean to avoid future bugs introduced by maintainers. At worst the task might result in adding some comments to explain why the code does what it does.

This kind of thing is generally a minor thing; it only matters in edge cases involving composite ops where more than one step alters the relevant part of the model. Then the 'operation' object would not be showing the true configuration.

yersan:
StaticDiscoveryAddHandler and DiscoveryOptionAddHandler details will be tracked on https://issues.redhat.com/browse/WFCORE-5281

jmesnil:
@bstansberry thanks

PR #4464, opened 2021-01-28
WFCORE-5247 IllegalArgumentException and CLI crash with read-attribute in /system-property

parsharma:
Issue: https://issues.redhat.com/browse/WFCORE-5247

### DIFF
--- a/cli/src/main/java/org/jboss/as/cli/handlers/ReadAttributeHandler.java
+++ b/cli/src/main/java/org/jboss/as/cli/handlers/ReadAttributeHandler.java
@@ -116,8 +116,12 @@ public class ReadAttributeHandler extends BaseOperationCommand {
                     op.get(Util.ADDRESS).setEmptyList();
                 } else {
                     final ModelNode addrNode = op.get(Util.ADDRESS);
-                    for(OperationRequestAddress.Node node : address) {
-                        addrNode.add(node.getType(), node.getName());
+                    for (OperationRequestAddress.Node node : address) {
+                        if (node.getName() == null) {
+                            addrNode.add(node.getType(), "*");
+                        } else {
+                            addrNode.add(node.getType(), node.getName());
+                        }
                     }
                 }
 
@@ -219,6 +223,7 @@ public class ReadAttributeHandler extends BaseOperationCommand {
         return req;
     }
 
+    @Override
     protected void handleResponse(CommandContext ctx, ModelNode response, boolean composite) throws CommandFormatException {
         if (!Util.isSuccess(response)) {
             throw new CommandFormatException(Util.getFailureDescription(response));
--- a/testsuite/standalone/src/test/java/org/jboss/as/test/integration/management/cli/CliCompletionTestCase.java
+++ b/testsuite/standalone/src/test/java/org/jboss/as/test/integration/management/cli/CliCompletionTestCase.java
@@ -22,6 +22,28 @@
  */
 package org.jboss.as.test.integration.management.cli;
 
+import static org.hamcrest.MatcherAssert.assertThat;
+import static org.junit.Assert.assertEquals;
+import static org.junit.Assert.assertFalse;
+import static org.junit.Assert.assertTrue;
+
+import org.aesh.complete.AeshCompleteOperation;
+import org.aesh.readline.completion.Completion;
+import org.aesh.readline.terminal.formatting.TerminalString;
+import org.hamcrest.CoreMatchers;
+import org.jboss.as.cli.CommandContext;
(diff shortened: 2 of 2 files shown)


bstansberry:
@jfdenise Please review

cli/src/main/java/org/jboss/as/cli/handlers/ReadAttributeHandler.java
```diff
@@ -116,8 +116,12 @@ public boolean canAppearNext(CommandContext ctx) throws CommandFormatException {
                     op.get(Util.ADDRESS).setEmptyList();
                 } else {
                     final ModelNode addrNode = op.get(Util.ADDRESS);
-                    for(OperationRequestAddress.Node node : address) {
-                        addrNode.add(node.getType(), node.getName());
+                    for (OperationRequestAddress.Node node : address) {
+                        if (!(node.getName() == null)) {
+                            addrNode.add(node.getType(), node.getName());
```
jfdenise, line 121: @parsharma , in this case you are replacing the current address (that is being built) by the empty address, so you loose the context (the various cd commands the user did). In order to keep the context, you could detect that the node.getName() returns null and in this case use the "*" value for the name to express the fact that the node name has not been set.
As a side note, it would be nice if you could add a test. It should be added in this file:
https://github.com/wildfly/wildfly-core/blob/master/testsuite/standalone/src/test/java/org/jboss/as/test/integration/management/cli/CliCompletionTestCase.java

cli/src/main/java/org/jboss/as/cli/handlers/ReadAttributeHandler.java
```diff
@@ -116,8 +116,12 @@ public boolean canAppearNext(CommandContext ctx) throws CommandFormatException {
                     op.get(Util.ADDRESS).setEmptyList();
                 } else {
                     final ModelNode addrNode = op.get(Util.ADDRESS);
-                    for(OperationRequestAddress.Node node : address) {
-                        addrNode.add(node.getType(), node.getName());
+                    for (OperationRequestAddress.Node node : address) {
+                        if (!(node.getName() == null)) {
```
jfdenise, line 120: Thank-you for the change, I would suggest a tiny change to first deal with the null case if (node.getName() == null), it simplify a bit code reading.

testsuite/standalone/src/test/java/org/jboss/as/test/integration/management/cli/CliCompletionTestCase.java
```diff
@@ -118,6 +118,36 @@ public void complexOperationWithObjectAsAttributeTest() {
             assertEquals(candidates.toString(), Arrays.asList(","), candidates);
         }
     }
+    /**
```
jfdenise, line 121: Nice to add some tests. It would be great to cover the fix you are putting in place. 
Something like:
CommandContext ctx = CLITestUtil.getCommandContext(TestSuiteEnvironment.getServerAddress(),          TestSuiteEnvironment.getServerPort(), System.in, System.out);
ctx.connectController();
// cd into a type to validate that completion tha toccurs in this context is correct.
ctx.handle("cd /system-property");
try {
            {
                String cmd = "read-attribute ";
                ctx.getDefaultCommandCompleter().complete(ctx,
                        cmd, cmd.length(), candidates);
                candidates = complete(ctx, cmd, null);
            }
} finally {
  ctx.terminateSession();
}

parsharma: Thanks for the suggestion @jfdenise. Please feel free to let me know if any further changes needed.

testsuite/standalone/src/test/java/org/jboss/as/test/integration/management/cli/CliCompletionTestCase.java
```diff
@@ -56,6 +39,20 @@
...
+import java.util.ArrayList;
+import java.util.Arrays;
+import java.util.Collections;
+import java.util.HashSet;
+import java.util.List;
+import java.util.Set;
+import java.util.stream.Collectors;
+
+import static org.hamcrest.MatcherAssert.assertThat;
+import static org.junit.Assert.*;
```
darranl, line 54: It looks like the order was wrong in the first place on this file, the static imports should be first - also can you please expand these back out instead of switching to a wildcard import?

jfdenise (approved):

bstansberry (approved):
Jean-Francois' approval is enough to get this merged but I'm adding mine because that will cause it to be shown as 'Approved' when looking at the PR queue.

PR #4466, opened 2021-01-29
[WFCORE-5262] Add a timestamp at the beginning of image startup in EAP images

khermano:
Issue: https://issues.redhat.com/browse/WFCORE-5262

related (cloned) issue: https://issues.redhat.com/browse/JBEAP-18005


bstansberry:
Hi @khermano 

I'm sorry, but I don't want this added to our server start, where it will get logged in every context, not just OpenShift. Also I believe the request is that it be added to the start of the OpenShift container, which is before this.

If this is to be done at all it should be in the OpenShift container integration code, for which the CLOUD JIRA project is used.

khermano:
Hi @bstansberry .
Thank you very much for your quick review and I am very sorry for messing up.
I will take a closer look at CLOUD JIRA project and I will close this PR.
But if I could have one question? Should I close WFCORE-5262 with some specific label or should I delete it completely (if it is possible)?

PR #4468, opened 2021-01-29
WFCORE-4868: implement aliases for standalone configuration files

michpetrov:
Issue: [WFCORE-4868](https://issues.redhat.com/browse/WFCORE-4868)

Should this be mentioned somewhere? I'm thinking `--help`, not user documentation.

### DIFF
--- a/controller/src/main/java/org/jboss/as/controller/persistence/ConfigurationFile.java
+++ b/controller/src/main/java/org/jboss/as/controller/persistence/ConfigurationFile.java
@@ -343,8 +343,13 @@ public class ConfigurationFile {
         if (mainName == null) {
             // Try the basic case, where name is the name
             final File directoryFile = new File(configurationDir, name);
+            // Translate alias if possible
+            String aliasedName = translateFileAlias(name);
+            final File aliasedFile = new File(configurationDir, aliasedName);
             if (directoryFile.exists()) {
                 mainName = stripPrefixSuffix(name); // TODO what if the stripped name doesn't exist? And why would there be a file like configuration/standalone.last.xml?
+            } else if (aliasedFile.exists()) {
+                mainName = aliasedName;
             } else if (interactionPolicy.isReadOnly()) {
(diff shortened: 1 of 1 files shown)


controller/src/main/java/org/jboss/as/controller/persistence/ConfigurationFile.java
```diff
@@ -362,6 +367,20 @@ private File determineMainFile(final String rawName, final String name) {
         throw ControllerLogger.ROOT_LOGGER.mainFileNotFound(name != null ? name : rawName, configurationDir);
     }
 
+    private String translateFileAlias(String alias) {
+        final int pos = rawFileName.lastIndexOf('.');
+        final String base = pos != -1 ? rawFileName.substring(0, pos) : rawFileName;
+        switch (alias) {
```
darranl, line 373: Do we really want a list like this in the code, this becomes something else that we need to remember to maintain if the list of files changes.

If the argument does not specify a suffix could we not instead assume standalone-{argument}.xml and check if we have a match?

michpetrov: I mean we could do that as well. The intention is not having to type long config names and e.g. `load-balancer` or 'microprofile` are IMHO still kinda long.

darranl:
@michpetrov This PR has been flagged as a feature so should be following the RFE process, will you be progressing this as an RFE?

michpetrov:
@darranl I am not familiar with the RFE process, what are the requirements?

darranl:
This pull request is still active and awaiting approval to proceed.

rsvoboda:
@michpetrov / @darranl what actions are remaining to get this PR merged?

darranl:
@rsvoboda Continuation of the RFE process up until the point a pre-check has been granted.

michpetrov:
@darranl well I wrote to the WildFly dev list twice and didn't get a reply, what do I do now?

PR #4473, opened 2021-02-04
[WFCORE-2957] read-resource without include-runtime shows runtime-only resources

petrberan:
https://issues.redhat.com/browse/WFCORE-2957


controller/src/main/java/org/jboss/as/controller/operations/global/ReadResourceHandler.java
```diff
@@ -326,8 +327,16 @@ private void doExecuteInternal(OperationContext context, ModelNode operation, bo
...
+                            // In case of runtime resources adds '=> undefined' if there's no include-runtime parameter,
+                            // in read-resource operation, otherwise adds '{"child" => undefined}'
+                            if (childReg.isRuntimeOnly()){
+                                if (queryRuntime){
+                                    childMap.get(child);
+                                }
+                            }
+                            else {
+                                childMap.get(child);
+                            }
```
bstansberry, line 339: This can be reduced to

```
if (queryRuntime || !childReg.isRuntimeOnly()) { 
  childMap.get(child);
}
```

bstansberry: Actually, it needs to be null safe, so...

```
if (queryRuntime || (childReg != null && !childReg.isRuntimeOnly())) { 
  childMap.get(child);
}
```

testsuite/shared/src/main/java/org/jboss/as/test/integration/management/extension/customcontext/testbase/CustomManagementContextTestBase.java
```diff
@@ -22,8 +22,7 @@
 import static org.jboss.as.controller.descriptions.ModelDescriptionConstants.ROLLBACK_ON_RUNTIME_FAILURE;
 import static org.jboss.as.controller.descriptions.ModelDescriptionConstants.SUBSYSTEM;
 import static org.jboss.as.controller.descriptions.ModelDescriptionConstants.SUCCESS;
-import static org.junit.Assert.assertEquals;
-import static org.junit.Assert.assertTrue;
+import static org.junit.Assert.*;
```
bstansberry, line 25: Our code style is not to use wildcard imports.

bstansberry (changes requested):

testsuite/shared/src/main/java/org/jboss/as/test/integration/management/extension/customcontext/testbase/CustomManagementContextTestBase.java
petrberan: Sorry, I added AssertFalse and IDEA decided to group all Assert imports into * import. Will take better care of this next time

controller/src/main/java/org/jboss/as/controller/operations/global/ReadResourceHandler.java
petrberan: Thank you, code is looking better now

bstansberry:
The various test failures (not talking about the 'Full integration' jobs here) are all because in a managed domain 'proxy' resources are being dropped from the response to read-resource operations that don't have the 'include-runtime' parameter. So the test does read-resource and then checks for '/host=a-remote-one' or '/host=x/server=y' and fails because that host or any server is 'runtime' resource.

Arguably that's correct, but it concerns me that this would become a breaking change for a kind of read that's common enough that we have it in several places in the testsuite.

Also, in the case of child type 'host' in some cases (i.e. remote hosts when the read-resource is executed on the DC) the child is a runtime-only resource, and in other cases it is NOT (i.e. the host resource for the host you invoke on, either the DC itself if you connected to a DC, or a non-DC host if you connected directly).  That's an obscure distinction.

This will likely fix the tests; question is whether it's a good idea:

```
if (queryRuntime || (childReg != null && (!childReg.isRuntimeOnly() || childReg.isRemote()))) { 
  childMap.get(child);
}
```

@jmesnil @yersan WDYT?

yersan:
> This will likely fix the tests; question is whether it's a good idea:
> 
> ```
> if (queryRuntime || (childReg != null && (!childReg.isRuntimeOnly() || childReg.isRemote()))) { 
>   childMap.get(child);
> }
> ```

I don't have any strong opinion against it. From my point of view, it can be assumed that proxy resources (current and future) will have this behavior when being read. So I think it could be fine.

I have my doubts about how useful to the users would be to have this exact distinction for these proxy resources when they are using a read operation. I mean whether these proxy runtime resources are shown or hidden based on the include-runtime flag.
The servers at `servers` runtime resource are always shown at `server-config`, so I don't believe it could be relevant if they are not shown.  The obscure distinction in the case of child type 'host' seems worst, but I'm not sure if it really could impact users, maybe internal operations done by the server but have doubts if it could affect the users. Keep them, even if we have the weird effect this Jira is trying to fix by using read-resource() and read-resource(recursive), doesn't sound too bad and will reduce the possibility to introduce a bug in DC / HC communication.

The alternative is to fix the tests, and hopefully, not be introducing any side effects. HAL should be reviewed as well, as far as I know, it uses extensively read operations to read the topology, just wanted to mention it.

darranl:
Do we have a plan to move forward on this one?  It looks like the PR in it's current form is not mergable so additional changes will be requird.

petrberan:
Hi @darranl , I apologize for this one. Recently I had some problems with notifications from github and I completely missed this requests in this PR.

I'll look into it ASAP

petrberan:
Hi @bstansberry and @yersan , sorry for getting this late to the changes in this issue and thank you for your help. 

I pushed the changes you proposed as tests run fine on my device, but there still seems to be few problems with integration jobs. Is there anything I can do either for this issue or regarding integration jobs, or is PR alright now?

yersan:
@petrberan I kicked them off again, they should be fine. One failed with an "Execution timeout", the other was using an incorrect wildfly-core version which is already fixed.

petrberan:
I apologize for this late update, found out that I pushed requested changes into a different branch by mistake.

PR #4475, opened 2021-02-05
[WFCORE-5272] Setting jacc provider to Elytron throws exceptions

spyrkob:
Issue: https://issues.redhat.com/browse/WFCORE-5272

### DIFF
--- a/elytron/src/main/java/org/wildfly/extension/elytron/PolicyDefinitions.java
+++ b/elytron/src/main/java/org/wildfly/extension/elytron/PolicyDefinitions.java
@@ -21,6 +21,7 @@ package org.wildfly.extension.elytron;
 import static org.wildfly.extension.elytron.Capabilities.JACC_POLICY_CAPABILITY;
 import static org.wildfly.extension.elytron.Capabilities.JACC_POLICY_RUNTIME_CAPABILITY;
 import static org.wildfly.extension.elytron.Capabilities.POLICY_RUNTIME_CAPABILITY;
+import static org.wildfly.extension.elytron.ElytronDescriptionConstants.CONFIGURATION_FACTORY;
 import static org.wildfly.extension.elytron.ElytronDescriptionConstants.CUSTOM_POLICY;
 import static org.wildfly.extension.elytron.ElytronDescriptionConstants.JACC_POLICY;
 import static org.wildfly.extension.elytron.ElytronDescriptionConstants.NAME;
@@ -57,6 +58,7 @@ import org.jboss.as.controller.ResourceDefinition;
 import org.jboss.as.controller.SimpleAttributeDefinition;
 import org.jboss.as.controller.SimpleAttributeDefinitionBuilder;
 import org.jboss.as.controller.SimpleResourceDefinition;
+import org.jboss.as.controller.capability.CapabilityServiceSupport;
 import org.jboss.as.controller.registry.ManagementResourceRegistration;
 import org.jboss.as.controller.registry.OperationEntry;
 import org.jboss.as.controller.registry.Resource;
@@ -238,6 +240,15 @@ class PolicyDefinitions {
                 .setAddHandler(add)
                 .setRemoveHandler(new ReloadRequiredRemoveStepHandler() {
                     @Override
+                    protected void performRuntime(OperationContext context, ModelNode operation, ModelNode model) throws OperationFailedException {
+                        super.performRuntime(context, operation, model);
+                        // JACC settings require a restart to take place
+                        if (model.get(JACC_POLICY).isDefined()) {
+                            context.restartRequired();
+                        }
+                    }
+
+                    @Override
                     protected void recordCapabilitiesAndRequirements(OperationContext context, ModelNode operation, Resource resource) throws OperationFailedException {
                         super.recordCapabilitiesAndRequirements(context, operation, resource);
                         context.deregisterCapability(JACC_POLICY_CAPABILITY); // even if it wasn't registered, deregistering is ok
@@ -250,6 +261,18 @@ class PolicyDefinitions {
             @Override
             public void registerAttributes(ManagementResourceRegistration resourceRegistration) {
                 OperationStepHandler write = new ReloadRequiredWriteAttributeHandler(attributes) {
+
+                    @Override
+                    protected boolean applyUpdateToRuntime(OperationContext context, ModelNode operation, String attributeName, ModelNode resolvedValue, ModelNode currentValue, HandbackHolder<Void> voidHandback) throws OperationFailedException {
+                        // the configuration factory will only be applied on restart
+                        for (ModelNode modelNode : resolvedValue.asList()) {
+                            if (modelNode.asProperty().getName().equals(CONFIGURATION_FACTORY)) {
+                                context.restartRequired();
+                            }
+                        }
+                        return super.applyUpdateToRuntime(context, operation, attributeName, resolvedValue, currentValue, voidHandback);
+                    }
+
                     @Override
                     protected void recordCapabilitiesAndRequirements(OperationContext context, AttributeDefinition attributeDefinition, ModelNode newValue, ModelNode oldValue) {
                         super.recordCapabilitiesAndRequirements(context, attributeDefinition, newValue, oldValue);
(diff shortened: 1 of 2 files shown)


elytron/src/main/java/org/wildfly/extension/elytron/PolicyDefinitions.java
```diff
@@ -237,6 +239,15 @@ private Policy getPolicy() {
...
                 .setRemoveHandler(new ReloadRequiredRemoveStepHandler() {
+                    @Override
+                    protected void performRuntime(OperationContext context, ModelNode operation, ModelNode model) throws OperationFailedException {
+                        super.performRuntime(context, operation, model);
+                        // JACC settings require a restart to take place
+                        if (model.get(JACC_POLICY).isDefined()) {
+                            context.restartRequired();
+                        }
+                    }
+
```
bstansberry, line 250: This needs to override recoverServices as well.

I think it would be better if this were just an AbstractRemoveStepHandler, not ReloadRequiredRemoveStepHandler, and then this  explicitly deals with the reload-required case in performRuntime and recoverServices. Calling super.performRuntime(context, operation, model) to deal with the non-restart-required case obscures what is really happening.

elytron/src/main/java/org/wildfly/extension/elytron/PolicyDefinitions.java
```diff
@@ -250,6 +261,18 @@ protected void recordCapabilitiesAndRequirements(OperationContext context, Model
             @Override
             public void registerAttributes(ManagementResourceRegistration resourceRegistration) {
                 OperationStepHandler write = new ReloadRequiredWriteAttributeHandler(attributes) {
+
+                    @Override
+                    protected boolean applyUpdateToRuntime(OperationContext context, ModelNode operation, String attributeName, ModelNode resolvedValue, ModelNode currentValue, HandbackHolder<Void> voidHandback) throws OperationFailedException {
+                        // the configuration factory will only be applied on restart
+                        for (ModelNode modelNode : resolvedValue.asList()) {
+                            if (modelNode.asProperty().getName().equals(CONFIGURATION_FACTORY)) {
```
bstansberry, line 269: I believe this handler is used for DEFAULT_POLICY, JaccPolicyDefinition.POLICY, CustomPolicyDefinition.POLICY. The resolvedValue.asList will break fro DEFAULT_POLICY, and the CONFIGURATION_FACTORY field is not relevant to CustomPolicyDefinition.POLICY.

Can this just check the value of attributeName?

bstansberry: Further to this, I was going to comment on how reverting this restartRequired needs to be handled as well. But I think this entire method override can be removed by just adding a call to setRestartJvm() to the builder that builds JaccPolicyDefinition.POLICY

elytron/src/main/java/org/wildfly/extension/elytron/PolicyDefinitions.java
```diff
@@ -323,9 +349,17 @@ public void accept(Consumer<Policy> policyConsumer) {
...
 
-                        doPrivileged((PrivilegedExceptionAction<PolicyConfigurationFactory>) () -> newPolicyConfigurationFactory(
-                                configurationFactory,
-                                defaultConfigurationFactory ? PolicyDefinitions.class.getClassLoader() : configuredClassLoader));
+                        if (context.isBooting() && legacyJacc) {
+                            throw ElytronSubsystemMessages.ROOT_LOGGER.unableToEnableJaccSupport();
+                        }
+
+                        if (legacyJacc) {
+                            context.restartRequired();
```
bstansberry, line 357: This feels wrong.  Not that context.restartRequired is called somewhere, but that it is called here.

Code that changes the runtime and then calls either reloadRequired or restartRequired is likely buggy code. Those things are logical alternatives.

This (and the exception throwing above) seems like something that should be handled in PolicyDefinitions.getPolicy() in its performRuntime implementation.

I also question whether the restartRequired is needed.  I don't think the 'context.isBooting()' at L352 is quite right.  Either the config is valid or it is not and the op that made it invalid should be rejected. You don't want to tell them to restart, and then they do and the boot fails.  The use case for not failing immediately would be so they could add a second op that removes legacy JACC. But if two ops are required to make a change, and the first leaves things in an invalid state, then the user should combine the two ops in a composite (e.g. a CLI batch). If the user used a composite like they should, when this OperationContext.Stage.RUNTIME handling executes the org.wildfly.legacy-security.jacc would no longer be present.

The one case where we do allow invalid sets of capabilities is during boot in an admin-only server. So that combination could WARN and set restartRequired. Otherwise the context should be setRollbackOnly and an OperationFailedException thrown.  And the 'boot in an admin-only server' case only needs to be considered if the OSH actually does runtime work in admin-only. (Most subsystem OSHs don't, but Elytron is unusual in this regard.)

spyrkob: The problem I was trying to solve with the context.isBooting was the sequence of 
```
/subsystem=security:write-attribute(name=initialize-jacc, value=false)
/subsystem=elytron/policy=jacc:add(jacc-policy={})
```
This leads to error since org.wildfly.legacy-security.jacc is not de-regestired until restart. I guess a better solution would be to de-register the capability in similar way Policy remove handler is doing? Or is OK to fail the 2nd operation in this case until the server is restarted?

elytron/src/main/java/org/wildfly/extension/elytron/PolicyDefinitions.java
spyrkob: I missed that it's applied to other policies, sorry.
As for using setRestartJVM() - that would apply the restartRequired to changes to other attributes as well. I was trying to restrict this to only CONFIGURATION_FACTORY, but if that's not a problem, I'll make this change.

elytron/src/main/java/org/wildfly/extension/elytron/PolicyDefinitions.java
bstansberry: It is fine to fail. A user should use a composite / CLI batch for this kind of change.

darranl: @bstansberry a composite wouldn't work here - a problem with the JACC APIs is they don't provide a way to replace the policy after first initialisation.

A more complex solution would be we register some form of delegate policy where we can subsequently swap in a different instance - or look to contribute changes to the APIs to allow a replacement but there are probably a lot of expectations out there about the stability of the global policy.

spyrkob: @darranl could we  make the elytron policy changes be applied to runtime only during boot? This way we could guarantee the policy is only initialised once.

darranl: I don't believe this is about initialisation once, we want to ensure it is initialised to the correct value and trigger a restart if it is not the correct value.

bstansberry: Got it; sorry I didn't understand the issue properly before and replied hastily. :( 

So AIUI there are two kinds of problems related to the use of capabilities here.

One is that a config change to legacy security can result in the "org.wildfly.legacy-security.jacc" capability being gone from the config, but the Policy is still in place. And then this handler needs to decide what to do. That's the more fundamental problem that I'll focus on below.

The other is that the initialize-jacc attribute's value is what really determines whether that capability should be present. The handling of that is an issue in the legacy security subsystem, but it's better to discuss that there.

Re the more fundamental problem the only idea I have is something we called a 'tombstone' in some old clustering code -- a marker for something that used to be alive but is now gone. So a "org.wildfly.legacy-security.jacc.tombstone" capability. It gets added when "org.wildfly.legacy-security.jacc" is added but never gets removed (except of course by process exit).

Presence of "org.wildfly.legacy-security.jacc" when Stage.RUNTIME executes for this add handler is an error condition (except maybe during admin-only boot.) But presence of "org.wildfly.legacy-security.jacc.tombstone" is not an error condition, it's just a sign that the Elytron JACC cannot be put in place now and restart is required.

Perhaps at some point the general capability system could add some kind of less hacky support for this general concept.

darranl: I think we do get to keep in mind one of the two subsystems registering this is deprecated and destined to be deleted at some point.

spyrkob:
@bstansberry @darranl updated the PR using a tombstone capability introduced by https://github.com/wildfly/wildfly/pull/13962

spyrkob:
rebased to resolve conflict in ElytronSubsystemMessages

spyrkob:
@darranl could you review the changes when you have a chance?

darranl:
I think there is still an outstanding review comment from Brian that needs to be followed up but other than that I think we are Ok.  Although is there a reciprocal issue where legacy security could be enabled after Elytron?

spyrkob:
Thanks @darranl. The corresponding legacy security change is in https://github.com/wildfly/wildfly/pull/13962

As for the outstanding review - do you mean https://github.com/wildfly/wildfly-core/pull/4475#discussion_r571350290 ? I think that should be addressed now - I changed the handler and added recoverServices implementation.

darranl (approved):

darranl:
@bstansberry I have flagged as ready to merge so speak up if your request is still outstanding ;-)

bstansberry (approved):

PR #4482, opened 2021-02-18
[WFCORE-5271] NullPointerException creating new KeyStore without type

lvydra:
https://issues.redhat.com/browse/WFCORE-5271

### DIFF
--- a/elytron/src/main/java/org/wildfly/extension/elytron/KeyStoreService.java
+++ b/elytron/src/main/java/org/wildfly/extension/elytron/KeyStoreService.java
@@ -177,7 +177,10 @@ class KeyStoreService implements ModifiableKeyStoreService {
                     }
                 } else {
                     if (keyStore == null) {
-                        ROOT_LOGGER.keyStoreNotExistsDefaultWillBeUsed();
+                        ROOT_LOGGER.debugf(
+                                "KeyStore: provider = %s  path = %s  resolvedPath = %s  password = %b  aliasFilter = %s does not exist. New keystore with default type will be created.",
+                                provider, path, resolvedPath, password != null, aliasFilter
+                        );
                         keyStore = AtomicLoadKeyStore.newInstance(KeyStore.getDefaultType());
                     }
 
--- a/elytron/src/main/java/org/wildfly/extension/elytron/KeyStoreService.java
+++ b/elytron/src/main/java/org/wildfly/extension/elytron/KeyStoreService.java
@@ -177,11 +177,12 @@ class KeyStoreService implements ModifiableKeyStoreService {
                     }
                 } else {
                     if (keyStore == null) {
+                        String defaultType = KeyStore.getDefaultType();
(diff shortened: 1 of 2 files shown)


elytron/src/main/java/org/wildfly/extension/elytron/KeyStoreService.java
```diff
@@ -176,6 +176,11 @@ public void start(StartContext startContext) throws StartException {
                         keyStore = AtomicLoadKeyStore.atomize(detected);
                     }
                 } else {
+                    if (keyStore == null) {
+                        ROOT_LOGGER.keyStoreNotExistsDefaultWillBeUsed();
```
darranl, line 180: See what the opinion of others is but I think we need to start asking if we really do need one more log message which will be on the console.  We are already receiving bug reports for the verbosity of our current messages so worth checking if any new ones are really justified.

As this is a one time event I think this could safely be logged at DEBUG level, this would mean we don't need the error to be defined in ElytronSubsystemMessages - if would be good for the message to also contain the type that was used.

jmesnil: I agree with Darran.
WARN is too high as this behaviour will not prevent the server to work. 
Besides, the log in itself does not contain enough contextual information to let the user "fix" the problem. Which keystore is impacted? What's "Used default" means: to us a "default keystore" (if such thing exists) or that a new keystore with "default type" will be created? What's the default type in that case?

lvydra: Hi @darranl @jmesnil, thank you for your suggestions.  I have added the requested changes.

elytron/src/main/java/org/wildfly/extension/elytron/KeyStoreService.java
```diff
@@ -176,6 +176,14 @@ public void start(StartContext startContext) throws StartException {
                         keyStore = AtomicLoadKeyStore.atomize(detected);
                     }
                 } else {
+                    if (keyStore == null) {
+                        ROOT_LOGGER.debugf(
```
darranl, line 180: Don't change unless you feel you want to as this PR does look good as it is.

You could immediately before the debugf call add a line like:

String type = KeyStore.getDefaultType();

This would mean you could include the type in the debug message as well.  You could then use this String at the point the KeyStore is created to avoid a second call to getDefaultType.

But this is only a minor consideration, I am happy to approve as is.  Thank you for the changes.

darranl (approved):
Added a minor comment but happy to merge as is.

Thank you.

elytron/src/main/java/org/wildfly/extension/elytron/KeyStoreService.java
lvydra: Hi @darranl, commit added :-) Thanks

jmesnil:
@lvydra thanks

PR #4487, opened 2021-02-21
[WFCORE-5285] Improve existing Objects.requireNonNull with message

boris-unckel:
Fixes https://issues.redhat.com/browse/WFCORE-5285

### DIFF
--- a/bootable-jar/runtime/src/main/java/org/wildfly/core/jar/runtime/Arguments.java
+++ b/bootable-jar/runtime/src/main/java/org/wildfly/core/jar/runtime/Arguments.java
@@ -55,7 +55,8 @@ final class Arguments {
     private Path deployment;
 
     static Arguments parseArguments(final List<String> args, final BootableEnvironment environment) throws Exception {
-        Objects.requireNonNull(args);
+        Objects.requireNonNull(args, "Parameter args cannot be null");
+        Objects.requireNonNull(environment, "Parameter environment cannot be null");
         Arguments arguments = new Arguments(environment);
         arguments.handleArguments(args);
         return arguments;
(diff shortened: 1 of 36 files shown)


boris-unckel:
@jmesnil The test have passed. The timeout failures are not related to the change. Could you please review and merge?

darranl:
This would mean all the error messages added are hard coded to English, we we are looking to add error messages for each of these we may be better moving to the following check instead:

https://github.com/wildfly/wildfly-common/blob/master/src/main/java/org/wildfly/common/Assert.java#L47

boris-unckel:
@darranl 
> This would mean all the error messages added are hard coded to English, we we are looking to add error messages for each of these we may be better moving to the following check instead:
> https://github.com/wildfly/wildfly-common/blob/master/src/main/java/org/wildfly/common/Assert.java#L47

I agree I18N support is a good idea. Is o.w.commons a global available dependency? If not, is it really worth introducing it in that way? Second thing: I saw IllegalArgumentEx vs NullPointerEx on stackoverflow but the discussion was not usecase driven. JavaSE decided for NullPointerEx. At the moment this is a change to existing code with NullPointerEx - do you want that refactored to a IllegalArgumentEx?

boris-unckel:
@darranl I have changed the null check to the Assert.checkNotNullParamWithNullPointerException to keep existing NPE behaviour and have I18N messages. Please review and merge.

boris-unckel:
@bstansberry Could you please have a look at this PR? It catched a stale label and is a minor only. Thanks in advance, happy easter!

jmesnil:
@boris-unckel I think this PR is now superseded by #4552 and its buddies.
I'm closing it so we can focus on the other smaller PRs

PR #4498, opened 2021-02-24
[WFCORE-4910] detect non custom jul and log warning

baranowb:
ISSUE: https://issues.redhat.com/browse/WFCORE-4910

### DIFF
--- a/logging/src/main/java/org/jboss/as/logging/deployments/LoggingDeploymentResourceProcessor.java
+++ b/logging/src/main/java/org/jboss/as/logging/deployments/LoggingDeploymentResourceProcessor.java
@@ -23,7 +23,9 @@
 package org.jboss.as.logging.deployments;
 
 import org.jboss.as.logging.CommonAttributes;
+import org.jboss.as.logging.LoggingModuleDependency;
 import org.jboss.as.logging.deployments.resources.LoggingDeploymentResources;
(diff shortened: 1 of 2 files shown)


jamezp:
I wonder if we should do this differently and add something to the log manager which will log a warning if handlers are updated. This is a highly specific fix where as making a change in the log manager itself would work for any library that attempts to configure logging.

darranl:
@baranowb did you have any thoughts on the feedback from @jamezp ?

baranowb:
@darranl yes. what would be threshold for warning, what kind of information should it provide?
Thing is, logging warning could be perceived as noisy behavior, since its normal operation for this interface. Just because user abuse it, does not necessarily it should rise alarm on every call. One criteria might be that at least one jboss.logging handler must be present or console handler must be present and warn only on its removal. Im fine with that. I will ping james and ask.

jamezp:
My idea is to modify the `LogContext.checkAccess()` to allowing a logging access after the security check. This though will require a change in the log manager itself as well as core to enable/disable access when it edits the logging configuration.

darranl:
@jamezp How should this one progress?  Do you think this approach will be mergable or does this need to be closed to consider the alternative approach you suggest?

baranowb:
@darranl  - Ive talked with james and this PR most likely can be rejected as fix will most likely be an RFE ?

PR #4502, opened 2021-03-01
Fix for WFCORE-5317, Upgrade galleon-plugins to 5.1.0

jfdenise:
Issue: https://issues.redhat.com/browse/WFCORE-5317
diff: https://github.com/wildfly/galleon-plugins/compare/5.0.0.Final...5.1.0.Final


jmesnil:
@jfdenise thanks

PR #4504, opened 2021-03-02
[WFCORE-5324] Support for CLI script executing at boot for bootable JAR

jfdenise:
Issues:
https://issues.redhat.com/browse/WFCORE-5322
https://issues.redhat.com/browse/WFCORE-5323
https://issues.redhat.com/browse/WFCORE-5324

Community doc PR:
https://github.com/wildfly/wildfly/pull/14069

* Introduce --cli-script=[path to cli script]
* Fix output when error occur during CLI script execution
* Fix ability to set server log config directory
* Evolved ts.bootable profile of elytron tests with CLI configuration applied at runtime.

### DIFF
--- a/bootable-jar/runtime/src/main/java/org/wildfly/core/jar/runtime/Arguments.java
+++ b/bootable-jar/runtime/src/main/java/org/wildfly/core/jar/runtime/Arguments.java
@@ -53,6 +53,7 @@ final class Arguments {
     private final List<String> serverArguments = new ArrayList<>();
     private final BootableEnvironment environment;
     private Path deployment;
+    private Path cliScript;
 
     static Arguments parseArguments(final List<String> args, final BootableEnvironment environment) throws Exception {
         Objects.requireNonNull(args);
@@ -101,6 +102,11 @@ final class Arguments {
                 serverArguments.add(a);
             } else if (CommandLineConstants.HELP.equals(a)) {
                 isHelp = true;
+            } else if (a.startsWith("--cli-script")) {
+                cliScript = Paths.get(getValue(a));
+                if (!Files.exists(cliScript) || !Files.isReadable(cliScript)) {
+                    throw new Exception("File doesn't exist or is not readable: " + cliScript);
+                }
             } else {
                 throw BootableJarLogger.ROOT_LOGGER.unknownArgument(a);
             }
@@ -151,6 +157,10 @@ final class Arguments {
      */
     public Path getDeployment() {
         return deployment;
+    }
+
+    public Path getCLIScript() {
+        return cliScript;
     }
 
     private static void addSystemProperty(final String arg, final Map<String, String> properties) {
--- a/bootable-jar/runtime/src/main/java/org/wildfly/core/jar/runtime/BootableJar.java
+++ b/bootable-jar/runtime/src/main/java/org/wildfly/core/jar/runtime/BootableJar.java
@@ -230,6 +230,18 @@ public final class BootableJar implements ShutdownHandler {
     }
 
     public void run() throws Exception {
+        Path script = arguments.getCLIScript();
+        if (script != null) {
+            long id = System.currentTimeMillis();
+            Path markerDir = environment.getTmpDir().resolve(id + "-cli-boot-hook-dir");
+            Path outputFile = environment.getTmpDir().resolve(id + "-cli-boot-hook-output-file.txt");
+            Files.createDirectories(markerDir);
+            startServerArgs.add("--start-mode=admin-only");
+            startServerArgs.add("-Dorg.wildfly.internal.cli.boot.hook.script=" + script.toAbsolutePath().toString());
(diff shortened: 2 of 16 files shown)


bootable-jar/runtime/src/main/java/org/wildfly/core/jar/runtime/Arguments.java
```diff
@@ -101,6 +102,11 @@ private void handleArguments(List<String> args) throws Exception {
                 serverArguments.add(a);
             } else if (CommandLineConstants.HELP.equals(a)) {
                 isHelp = true;
+            } else if (a.startsWith("--cli-script")) {
+                cliScript = Paths.get(getValue(a));
```
jmesnil, line 106: what would happen if the path of the script is relative? is it resolved based on the bootable jar path or some of our well known dirs?

bootable-jar/runtime/src/main/java/org/wildfly/core/jar/runtime/BootableJar.java
```diff
@@ -230,6 +230,17 @@ private LogContext configureLogContext() throws IOException {
     }
 
     public void run() throws Exception {
+        Path script = arguments.getCLIScript();
+        if (script != null) {
+            Path markerDir = environment.getJBossHome().resolve("cli-boot-hook-dir");
+            Path outputFile = environment.getJBossHome().resolve("cli-boot-hook-output-file.txt");
+            Files.createDirectories(markerDir);
```
jmesnil, line 237: being consistent with S2I is more important but we should avoid creating these types of dirs directly in `$JBOSS_HOME`. the `data` directory is a better location for such resources.

bootable-jar/runtime/src/main/java/org/wildfly/core/jar/runtime/_private/BootableJarLogger.java
```diff
@@ -162,4 +162,7 @@
 
     @Message(id = Message.NONE, value = "Display the content of the Galleon configuration used to build this bootable JAR")
     String argDisplayGalleonConfig();
+
+    @Message(id = Message.NONE, value = "Path to a CLI script to execute")
```
jmesnil, line 166: maybe specify that this option is when running the bootable jar (to avoid being confused with CLI scripts that are executed when the jar is built).

cli/src/main/java/org/jboss/as/cli/impl/BootScriptInvoker.java
```diff
@@ -87,11 +88,6 @@ public void runCliScript(ModelControllerClient client, File file) {
         } catch (Exception ex) {
             try {
                 ROOT_LOGGER.errorProcessingScript(file);
-                StringBuilder errBuilder = new StringBuilder();
```
jmesnil, line 90: Is it intentional to get rid of this?

cli/src/main/java/org/jboss/as/cli/impl/CliConfigImpl.java
```diff
@@ -94,6 +94,7 @@
     static CliConfig newBootConfig() throws CliInitializationException {
         CliConfigImpl config = new CliConfigImpl();
         config.validateOperationRequests = false;
+        config.echoCommand = true;
```
jmesnil, line 97: is this change required to support the new `--cli-script`?

bootable-jar/runtime/src/main/java/org/wildfly/core/jar/runtime/Arguments.java
jfdenise: If relative, it is relative to the JVM work dir (from where the JVM has been launched).

bootable-jar/runtime/src/main/java/org/wildfly/core/jar/runtime/BootableJar.java
jfdenise: @jmesnil , you mean <home>/standalone/data? Wondering if <home>/standalone/tmp would be more appropriate?

bootable-jar/runtime/src/main/java/org/wildfly/core/jar/runtime/_private/BootableJarLogger.java
jfdenise: Good point!

cli/src/main/java/org/jboss/as/cli/impl/BootScriptInvoker.java
jfdenise: Yes, the way errors were reported was not nice. The command is now printed to the output and output displayed when an error occurs. Much easier to match the faulty line and the error. The output is not shown if the CLI script execution was successful.

cli/src/main/java/org/jboss/as/cli/impl/CliConfigImpl.java
jfdenise: No, not required. The error reporting is now better.

jmesnil (approved):

jmesnil:
@jfdenise thanks!

PR #4509, opened 2021-03-03
[WFCORE-4302] Validation of 'host-context-map' attribute of 'server-ssl-sni-context' should not reject ^ character

Skyllarr:
https://issues.redhat.com/browse/WFCORE-4302


Skyllarr:
@darranl I fixed this and added a simple test as well.

darranl (approved):

darranl:
@bstansberry are you likely to want a model bump and transformer for this one?

bstansberry:
@darranl @Skyllarr I guess the argument for a transformer would be to detect values that were previously rejected and have the transformer reject them?

Both doing nothing (so the resource/op fails on the legacy host) or a reject transformer mean a failure, but the reject transformer is perhaps an easier to understand one.  That's a pretty minor benefit; not worth a version bump + transformer IMO.

A middle ground (that I'm not asking for) is to file a reminder JIRA and if you end up with a model bump for WF 24 anyway, add the transformer. If you don't, then 'Won't Do' resolve it. My impression is a lot of the pain is the model bump.

darranl:
@Skyllarr I think to be complete we should just add the transformer, the model bump PR is here https://github.com/wildfly/wildfly-core/pull/4530

Skyllarr:
> @Skyllarr I think to be complete we should just add the transformer, the model bump PR is here #4530

@darranl  Thanks. I added the transformer.

darranl:
This PR is still active, we just need at least one feature for the model bump.

PR #4516, opened 2021-03-04
[WFCORE-5328] Excluding ExpressionResolutionTestCase on IBM JDK8 until it is fixed

ropalka:
https://issues.redhat.com/browse/WFCORE-5328


darranl (changes requested):
There is no follow up Jira issue identifying the underlying problem here - TBH unless the underlying problem is understood a test failure on another JVM could be the sign of a Blocker.

ropalka:
I didn't plan to create follow up JIRA @darranl . My intentention was to create a JIRA to have this issue tracked in our issue tracker.
I changed both WFCORE-5328 title and instructions about how to reproduce this issue to be aligned with my original intention.

darranl:
I think this is going to need investigating now, if there is a failure on the IBM JDK there is a risk this is a sign of something that could affect the other JDKs.

This new feature is integrated very tightly into the boot process of the servers so if there is any instability I want to be sure we know what it is.

darranl:
Superseded by https://github.com/wildfly/wildfly-core/pull/4520

PR #4519, opened 2021-03-05
WFCORE-5041 The public elytron module exports private API

parsharma:
Issue: https://issues.redhat.com/browse/WFCORE-5041


darranl:
The failure here is related to the code change, I think it will need debugging further to get to the bottom of what is happening but it could be a sign that we have been relying on the additional exports the attribute was triggering.

darranl:
@parsharma I think the test results here have highlighted this will be a much bigger task.

I think WildFly 24 may be too soon for this change but if you want to continue ping me and we can use a custom CI job rather than the parallel jobs here to identify issues to work through.

I will close this one for now and then we can reopen once ready,

PR #4522, opened 2021-03-06
[WFCORE-5331] Upgrading JBoss Marshalling to 2.0.11.Final

ropalka:
https://issues.redhat.com/browse/WFCORE-5331

Fixes:
 * [JBMAR-234] Fix JDK11 insufficient privileges issue
 * [JBMAR-235] Reinitialize array in IdentityIntMap and IdentityIntSet when array size becomes large


darranl:
Looking at the diff in JBoss Marshalling there seems to be a mixture with some calls using doPrivileged without checking is a security manager is installed - was there a reason to skip the security manager check?

ropalka:
Thanks for the review @darranl - good catch! It was an oversight on my side.
The following PR addresses it: https://github.com/jboss-remoting/jboss-marshalling/pull/99

ropalka:
This should be now ready to go upstream (after clarification).

jmesnil:
@ropalka Would you mind creating a PR against 15.x so that we incorporate this upgrade in WildFly Core 15.0.1.Final please?

PR #4523, opened 2021-03-11
[WFCORE-5334]: Deleting Configuration Data when Git connection is fai…

ehsavoie:
…led.

* Moving files into an "attic" folder.
* Restoring files in case of git failure.
* Deleting attic files afterwards.

Jira: https://issues.redhat.com/browse/WFCORE-5334

### DIFF
--- a/server/src/main/java/org/jboss/as/server/controller/git/GitRepository.java
+++ b/server/src/main/java/org/jboss/as/server/controller/git/GitRepository.java
@@ -26,11 +26,9 @@ import static org.eclipse.jgit.lib.Constants.R_REMOTES;
 import java.io.Closeable;
 import java.io.File;
 import java.io.IOException;
-import java.nio.file.FileVisitResult;
-import java.nio.file.FileVisitor;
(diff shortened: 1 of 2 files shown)


jmesnil:
I'm not sure to understand the issue.

If the git pull fails, what has been modified in the local repository?

ehsavoie:
To be able to pull you need to target an empty directory. So at the first pull WildFly deletes some directories. If the pull in question fails then those directories/ files are 'lost' and this blocks the server from starting.

jmesnil:
@ehsavoie so do you need to copy the existing directory or move it?

ehsavoie:
I move the files to a temp place (the 'attic') to be able to restore if the git operation fails

jmesnil (approved):

jmesnil:
@ehsavoie Windows failures are related to the patch

jmesnil:
@ehsavoie I've added a `fixme` label to hold this PR until the Windows failures are fixed

PR #4528, opened 2021-03-16
WFCORE-5263 Update all projects with a dependency on wildfly-elytron to use individual modules

parsharma:
Issue: https://issues.redhat.com/browse/WFCORE-5263

### DIFF
--- a/testsuite/pom.xml
+++ b/testsuite/pom.xml
@@ -212,6 +212,14 @@
             <artifactId>wildfly-core-testsuite-shared</artifactId>
             <scope>test</scope>
         </dependency>
+        <dependency>
+            <groupId>org.wildfly.security</groupId>
+            <artifactId>wildfly-elytron-auth</artifactId>
+        </dependency>
+        <dependency>
+            <groupId>org.wildfly.security</groupId>
+            <artifactId>wildfly-elytron-auth-server</artifactId>
+        </dependency>
     </dependencies>
 
     <build>
--- a/cli/pom.xml
+++ b/cli/pom.xml
@@ -150,8 +150,38 @@
         </dependency>
         <dependency>
             <groupId>org.wildfly.security</groupId>
-            <artifactId>wildfly-elytron</artifactId>
+            <artifactId>wildfly-elytron-auth</artifactId>
         </dependency>
+
+        <dependency>
+            <groupId>org.wildfly.security</groupId>
+            <artifactId>wildfly-elytron-base</artifactId>
+        </dependency>
+
+        <dependency>
+            <groupId>org.wildfly.security</groupId>
+            <artifactId>wildfly-elytron-credential</artifactId>
+        </dependency>
+
+        <dependency>
+            <groupId>org.wildfly.security</groupId>
+            <artifactId>wildfly-elytron-password-impl</artifactId>
+        </dependency>
+        <dependency>
+            <groupId>org.wildfly.security</groupId>
+            <artifactId>wildfly-elytron-security-manager</artifactId>
+        </dependency>
+
+        <dependency>
+            <groupId>org.wildfly.security</groupId>
+            <artifactId>wildfly-elytron-security-manager-action</artifactId>
+        </dependency>
+
+        <dependency>
+            <groupId>org.wildfly.security</groupId>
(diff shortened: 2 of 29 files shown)


parsharma:
Still work is going on. I have raised this PR before the whole changes to test if anything is breaking because of it.

jmesnil:
@parsharma I'll put this PR on hold to avoid merging until it is ready.

Please add a comment when you need a review.

parsharma:
@jmesnil the work is done. PR is ready for review. Thanks

jmesnil:
@fjuma or @darranl could you please review?

testsuite/pom.xml
```diff
@@ -212,6 +212,18 @@
...
+            <groupId>org.wildfly.security</groupId>
+            <artifactId>wildfly-elytron-auth</artifactId>
+        </dependency>
+        <dependency>
+            <groupId>org.wildfly.security</groupId>
+            <artifactId>wildfly-elytron-auth-server</artifactId>
+        </dependency>
+        <dependency>
+            <groupId>org.wildfly.security</groupId>
+            <artifactId>wildfly-elytron</artifactId>
```
fjuma, line 225: This should be removed.

fjuma (changes requested):
Thanks for working on this @parsharma! It looks like there are a few more places where the dependency on wildfly-elytron should be removed:

* pom.xml
* core-feature-pack/common/pom.xml
* elytron/pom.xml

parsharma:
@fjuma I have not removed elytron dependency from parent/pom.xml because of transitive dependencies.

fjuma:
Thanks for the update @parsharma! To be able to remove the `org.wildfly.security:wildfly-elytron` dependency from the main `pom.xml`, I think we will also need to add an exclusion to the `org.jboss.remoting:jboss-remoting` and the `org.jboss.remotingjmx:remoting-jmx` dependencies in the main `pom.xml` and then work through the compilation errors that occur. From a quick look at the initial compilation failures that will result, it appears that the `domain-management/pom.xml` is missing dependencies on individual Elytron modules.

fjuma:
Thanks very much for the updates @parsharma! The WildFly Core testsuite is now passing - that's great! The WildFly full testsuite is now failing because the `org.wildfly.security:wildfly-elytron` dependency is no longer available. To get that to pass, the next step would be to work on [WFLY-14377](https://issues.redhat.com/browse/WFLY-14377) which involves updating WildFly to depend on the individual Elytron modules instead of `wildfly-elytron`.
 
We have a couple options on how to proceed:

Option 1: Work on and then merge the changes for WFLY-14377 first in order to get CI to pass for this PR.
Option 2: Now that we know all the missing dependencies on individual Elytron modules have been added where needed in this PR, we could temporarily add back the `org.wildfly.security:wildfly-elytronwildfly-elytron` dependency to the main pom.xml here to get CI passing for this PR in order to merge it. We could then work on the changes for WFLY-14377 and then finally remove the `wildfly-elytron` dependency from the main pom.xml for WildFly Core after that.

@darranl Which option do you prefer?

darranl:
Thanks for the ping @fjuma 

Something I am going to be starting some discussions on separately is how we can cut down on code going stale in long running PRs - so on that basis I think option 2, get this merged and then we come back for the final clean up.  As this PR has touched a number of poms there is a very serious risk other PRs will introduce conflicts if this sits too long.  Git is strong in many places, concurrent modifications to XML files is not one of those strengths!

We have a lot of other clean up work that is likely to be iterative anyway so this will not be the first.

Thank you for all the contributions @parsharma this is really good to have WildFly Core tackled.

fjuma:
+1

I've created [WFCORE-5374](https://issues.redhat.com/browse/WFCORE-5374) to track removing the `wildfly-elytron` dependency from the main pom.xml.

darranl:
@parsharma are we ready to take the hold label off this one?  I assume once CI passes we are ready for final review and merge.

fjuma (changes requested):
Just have one small comment. The rest looks good to me. Thanks @parsharma!

fjuma (approved):
Thanks @parsharma!

PR #4532, opened 2021-03-17
[WFCORE-4214] VaultPasswordsInCLITestCase fails with Elytron profile

petrberan:
https://issues.redhat.com/browse/WFCORE-4214


petrberan:
As discussed with Darran Lofthouse, this test case should not run with -Delytron profile as it is vault related. However, running this test with -Delytron profile causes failures in tests following this one in manualmode module, so a check of whether Elytron is used.

In case of Elytron use, this test case is skipped, allowing following tests to be run properly and not to fail.

darranl (approved):

jmesnil:
@petrberan thanks

PR #4533, opened 2021-03-18
[WFCORE-2925] Heap settings separation in standalone and domain conf scripts

lvydra:
https://issues.redhat.com/browse/WFCORE-2925

### DIFF
--- a/core-feature-pack/common/src/main/resources/content/bin/domain.conf
+++ b/core-feature-pack/common/src/main/resources/content/bin/domain.conf
@@ -45,8 +45,11 @@ fi
 #
 # Specify options to pass to the Java VM.
 #
+if [ "x$JBOSS_JAVA_SIZING" = "x" ]; then
+   JBOSS_JAVA_SIZING="-Xms64m -Xmx512m -XX:MaxMetaspaceSize=256m"
+fi
 if [ "x$JAVA_OPTS" = "x" ]; then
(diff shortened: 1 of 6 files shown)


darranl:
@lvydra have you had any discussions with anyone about taking on this issue, and around the naming of the variable to use?

lvydra:
Hi @darranl , I have discussed taking on this issue, but not the naming. I'm open to suggestions/discussion.

darranl:
Thanks, I was asking because Brian says he only gave the names 20s thought but each of the variables does either have size in it's description or name so I think we are good.

Everyone has now had over a month to come up with alternatives.

darranl (approved):

darranl:
@bstansberry I have added an approval of this one as the suggested changes have been implemented and the name feels correct - but do you think we need some documentation updates as well?

bstansberry:
@luck3y @jfdenise Any concerns about this? This idea (which I forgot I had!) predates us doing cloud stuff, bootable jar etc. 

Partly I'm asking as an FYI in case we need to adjust anything elsewhere

luck3y:
This looks fine to me, we'd just have to set JBOSS_JAVA_SIZING to some not empty value in images to meet the other requirement (I don't have the ticket # handy) for not specifying any default values for max metaspace. So +1 from me.

bstansberry:
@darranl I don't see any docs of JAVA_OPTS, so it seems we can live without. ;)

@luck3y We should get rid of the max metaspace in wf core as well. :)

bstansberry:
Thanks @lvydra !

PR #4547, opened 2021-03-30
Fix for WFCORE-5362, Upgrade galleon-plugins to 5.1.2

jfdenise:
Issue: https://issues.redhat.com/browse/WFCORE-5362
Diff: https://github.com/wildfly/galleon-plugins/compare/5.1.1.Final...5.1.2.Final


jmesnil:
@jfdenise thanks

PR #4551, opened 2021-04-07
[WFCORE-5285] Utilize o.w.common.Assert for Null-Checks (controller)

boris-unckel:

Original PR was too big to verify. It has been splitted up, here: controller


jmesnil (approved):

boris-unckel:
@jmesnil Sorry for causing a commit after approval. I have noticed that array checks have to be done with checkNotNullArrayParam, avoiding String concatenation in regular cases. It's fixed in OperationContextImpl.

PR #4555, opened 2021-04-07
[WFCORE-5285] Utilize o.w.common.Assert for Null-Checks (security)

boris-unckel:

Original PR was too big to verify. It has been splitted up, here: security

### DIFF
--- a/core-security/implementation/pom.xml
+++ b/core-security/implementation/pom.xml
@@ -60,6 +60,11 @@
         </dependency>
 
         <dependency>
+            <groupId>org.wildfly.common</groupId>
+            <artifactId>wildfly-common</artifactId>
+        </dependency>
+
+        <dependency>
             <groupId>junit</groupId>
             <artifactId>junit</artifactId>
             <scope>test</scope>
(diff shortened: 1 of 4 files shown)


core-security/implementation/pom.xml
```diff
@@ -59,6 +59,11 @@
             <optional>true</optional>
         </dependency>
 
+        <dependency>
+            <groupId>org.wildfly.common</groupId>
```
jmesnil, line 63: @darranl do you see any issue with adding this dependency?

darranl: I think the dependency is find but it looks like the module does not have the dependency so it is odd this is not resulting in an error.

darranl: Actually there are a number of test failures, I would expect this module to also need the dependency adding:

https://github.com/wildfly/wildfly-core/blob/master/core-feature-pack/common/src/main/resources/modules/system/layers/base/org/jboss/as/core-security/main/module.xml

jmesnil (changes requested):
could you please add the dependency to the `org.jboss.as.core-security` module?

boris-unckel:
@jmesnil all tests are successful, except "Full integration - Windows" where the CI failed. Please check for approval.

jmesnil (approved):

PR #4561, opened 2021-04-08
[WFCORE-5334]: Deleting Configuration Data when Git connection is failed.

ehsavoie:
* Moving files into an "attic" folder.
* Restoring files in case of git failure.
* Deleting attic files afterwards.

Jira: https://issues.redhat.com/browse/WFCORE-5334

### DIFF
--- a/testsuite/manualmode/src/test/java/org/jboss/as/test/manualmode/management/persistence/RemoteSshGitRepositoryTestCase.java
+++ b/testsuite/manualmode/src/test/java/org/jboss/as/test/manualmode/management/persistence/RemoteSshGitRepositoryTestCase.java
@@ -13,7 +13,6 @@
  * See the License for the specific language governing permissions and
  * limitations under the License.
  */
-
 package org.jboss.as.test.manualmode.management.persistence;
 
 import static java.nio.charset.StandardCharsets.US_ASCII;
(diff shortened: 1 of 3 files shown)


testsuite/manualmode/src/test/java/org/jboss/as/test/manualmode/management/persistence/RemoteSshGitRepositoryTestCase.java
```diff
@@ -220,14 +222,14 @@ public void prepareTest() throws Exception {
         hostPrivateKey.flush();
         hostKeyPair.writePublicKey(publicHostKey, "");
 
-        sshServer = new SSHServer(EC_USER, Paths.get(SSH_DIR +'/' + EC_PUBKEY),
+        sshServer = new SSHServer(EC_USER, Paths.get(SSH_DIR + '/' + EC_PUBKEY),
```
jmesnil, line 225: @ehsavoie the tests pass so it might be a silly question but should we use proper File constant for the file separator to make sure it works on Unix and Windows?

ehsavoie: Paths.get(SSH_DIR).resolve(EC_PUBKEY) would be even better i think

jmesnil (approved):

jmesnil:
@ehsavoie thanks

PR #4562, opened 2021-04-09
WFCORE-4162 Incorrect OperationContext.getCapabilityServiceName javadoc

parsharma:
Issue: https://issues.redhat.com/browse/WFCORE-4162

### DIFF
--- a/controller/src/main/java/org/jboss/as/controller/OperationContext.java
+++ b/controller/src/main/java/org/jboss/as/controller/OperationContext.java
@@ -1037,7 +1037,7 @@ public interface OperationContext extends ExpressionResolver {
     /**
      * Gets the name of a service associated with a given capability, if there is one.
      * @param capabilityName the name of the capability. Cannot be {@code null}
-     * @param serviceType class of the java type that exposes by the service. Cannot be {@code null}
+     * @param serviceType class of the java type that exposes by the service. Can be null.
      * @return the name of the service. Will not return {@code null}
      *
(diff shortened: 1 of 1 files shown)


darranl (changes requested):
If the javadoc in this interface is being changed we should also look at the related API https://github.com/wildfly/wildfly-core/blob/master/controller/src/main/java/org/jboss/as/controller/capability/registry/ImmutableCapabilityRegistry.java which also specifies it must not be null.

As part of this work have you double checked every implementation of the APIs is null safe?

parsharma:
Hello @darranl,
Here https://github.com/wildfly/wildfly-core/blob/master/controller/src/main/java/org/jboss/as/controller/capability/registry/ImmutableCapabilityRegistry.java#L78 it is mentioned it may be {code null} if the user is unconcerned about the validation. And I have also checked API is null safe. For example, here https://github.com/wildfly/wildfly-core/blob/master/controller/src/main/java/org/jboss/as/controller/capability/RuntimeCapability.java#L141 in case user doesn't want the validation and they will pass the serviceType {code null} then it will handle here. Please let me know if my understanding is wrong.
Thanks

darranl:
I think this should still look at the second interface I linked which has the same restrictions defined, if that implementations of that interface allow null we should update the javadoc there as well.

The reason I am suggesting this is we are relaxing the documented restrictions on an API but not applying the same relaxation to the API likely to be called so anyone implementing the second one could inadvertently require it to not be null as it is documented that way.

darranl (approved):

darranl:
Sorry disregard my comment, the second interface doesn't need a change.

PR #4565, opened 2021-04-09
[WFCORE-5285] Utilize o.w.common.Assert for Null-Checks (protocol)

boris-unckel:

Original PR was too big to verify. It has been splitted up, here: protocol

### DIFF
--- a/protocol/pom.xml
+++ b/protocol/pom.xml
@@ -70,6 +70,10 @@
             <artifactId>jboss-remoting</artifactId>
         </dependency>
         <dependency>
+            <groupId>org.wildfly.common</groupId>
+            <artifactId>wildfly-common</artifactId>
+        </dependency>
+        <dependency>
             <groupId>org.wildfly.security</groupId>
             <artifactId>wildfly-elytron</artifactId>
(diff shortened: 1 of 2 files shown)


protocol/pom.xml
```diff
@@ -69,6 +69,10 @@
             <groupId>org.jboss.remoting</groupId>
             <artifactId>jboss-remoting</artifactId>
         </dependency>
+        <dependency>
```
jmesnil, line 72: I've checked and `org.jboss.as.protocol` module already had a dependency on `org.wildfly.common`: https://github.com/wildfly/wildfly-core/blob/150c652e5b2e916751afada7cddaf0367765961d/core-feature-pack/common/src/main/resources/modules/system/layers/base/org/jboss/as/protocol/main/module.xml#L40

jmesnil (approved):

jmesnil:
@boris-unckel thanks for all the PRs, I've merged them all now that they were approved.

boris-unckel:
@jmesnil thanks for review, support and your patience :-)

PR #4569, opened 2021-04-14
[WFCORE-5373] Fix missing resource handling

jasondlee:
https://issues.redhat.com/browse/WFCORE-5370

If the resource to be removed doesn't exist, the operation should fail


jmesnil (approved):

jmesnil:
@jasondlee thanks

PR #4570, opened 2021-04-19
[WFCORE-5367] Sync model handler wrapper does not take into account a…

yersan:
…dditional steps added by the wrapped handler

Jira issue: https://issues.redhat.com/browse/WFCORE-5367

SyncModelOperationHandlerWrapper wraps a handler and when the wrapped handler is done, it verifies whether it is necessary to sync the domain model between the DC and the slave. When the wrapped handler adds additional steps that modify the model, the verifications are done without taking into account any immediate step added by the wrapped handler.

That's a problem with the remove handler, which adds recursive immediate steps to delete the child resources before deleting the requested resource.

### DIFF
--- a/controller/src/main/java/org/jboss/as/controller/operations/common/Util.java
+++ b/controller/src/main/java/org/jboss/as/controller/operations/common/Util.java
@@ -26,6 +26,7 @@ import static org.jboss.as.controller.descriptions.ModelDescriptionConstants.OP_
 import static org.jboss.as.controller.descriptions.ModelDescriptionConstants.OUTCOME;
 import static org.jboss.as.controller.descriptions.ModelDescriptionConstants.READ_ATTRIBUTE_OPERATION;
 import static org.jboss.as.controller.descriptions.ModelDescriptionConstants.READ_RESOURCE_DESCRIPTION_OPERATION;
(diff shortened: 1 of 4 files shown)


bstansberry (commented):
@yersan Can the test be included in DomainTestSuite? The setup/teardown looks standard.

Fix looks good.

yersan:
Thanks for the review @bstansberry , yes, probably it can be included there. I was avoiding it because I'm deleting all the servers in the test and I didn't want to hardcode server configurations in the test itself. However, those are the standard configurations for all the tests and they are unlikely to change. I'll move it to the testsuite.

In a more general idea, maybe we could add to DomainTestSuite the ability to save and recover snaphosts, so any test included in the suite can use that facility.

Let me know if you have any concerns, if not, I'll move the test to the test suite and explore the snaphost ability for the DomainTestSuite

bstansberry:
@yersan i just did a retest; if that looks good I'd like to just merge this, and if you want to do anything further re tests that can come separately.

yersan:
Thank  @bstansberry for merging it. I forgot to notify you or request your review again after moving the test to the testsuite. I'm fine adding the take snapshots capabilities to just this test. The test suite itself didn't look like a correct place.

PR #4571, opened 2021-04-22
WFCORE-5305 Response warnings don't propagate up in read-resource response assembly

parsharma:
Issue: https://issues.redhat.com/browse/WFCORE-5305

### DIFF
--- a/controller/src/main/java/org/jboss/as/controller/operations/global/ReadResourceHandler.java
+++ b/controller/src/main/java/org/jboss/as/controller/operations/global/ReadResourceHandler.java
@@ -26,9 +26,13 @@ import static org.jboss.as.controller.descriptions.ModelDescriptionConstants.ACC
 import static org.jboss.as.controller.descriptions.ModelDescriptionConstants.FAILURE_DESCRIPTION;
 import static org.jboss.as.controller.descriptions.ModelDescriptionConstants.OP;
 import static org.jboss.as.controller.descriptions.ModelDescriptionConstants.OP_ADDR;
-import static org.jboss.as.controller.descriptions.ModelDescriptionConstants.READ_ATTRIBUTE_OPERATION;
 import static org.jboss.as.controller.descriptions.ModelDescriptionConstants.READ_RESOURCE_OPERATION;
 import static org.jboss.as.controller.descriptions.ModelDescriptionConstants.RESULT;
+import static org.jboss.as.controller.descriptions.ModelDescriptionConstants.READ_ATTRIBUTE_OPERATION;
+import static org.jboss.as.controller.descriptions.ModelDescriptionConstants.RESPONSE_HEADERS;
+import static org.jboss.as.controller.descriptions.ModelDescriptionConstants.WARNING;
+import static org.jboss.as.controller.descriptions.ModelDescriptionConstants.WARNINGS;
+
 import static org.jboss.as.controller.operations.global.GlobalOperationAttributes.INCLUDE_ALIASES;
 import static org.jboss.as.controller.operations.global.GlobalOperationAttributes.INCLUDE_DEFAULTS;
 import static org.jboss.as.controller.operations.global.GlobalOperationAttributes.INCLUDE_RUNTIME;
@@ -45,6 +49,7 @@ import java.util.LinkedHashMap;
 import java.util.Map;
 import java.util.Set;
 import java.util.TreeMap;
+import java.util.logging.Level;
 
 import org.jboss.as.controller.AttributeDefinition;
 import org.jboss.as.controller.OperationContext;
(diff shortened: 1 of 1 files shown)


parsharma:
Zulip Chat:
Parul Sharma: I was thinking if we create another issue for ReadAttribute because it needs some more analysis and this issue is different https://issues.redhat.com/browse/WFCORE-5305 from ReadAttribute.
What do you suggest?

Brian Stansberry: 7:18 PM
I agree to use another issue for ReadAttributeHandler

controller/src/main/java/org/jboss/as/controller/operations/global/ReadResourceHandler.java
```diff
@@ -506,6 +508,8 @@ public void execute(OperationContext context, ModelNode operation) throws Operat
                 ModelNode value = ar.response;
                 if (!value.has(FAILURE_DESCRIPTION)) {
                     sortedAttributes.put(entry.getKey(), value.get(RESULT));
+                    if(value.hasDefined(RESPONSE_HEADERS))
+                    context.addResponseWarning(Level.WARNING, value.get(RESPONSE_HEADERS));
```
bstansberry, line 512: Please indent and wrap the body of the if in curly braces.

I think propagation needs to be more complex, i.e. IIRC the headers from the value have structure that includes the level and a message so that info needs to be used to create equivalent output here.

bstansberry:
@parsharma re the LdapTestCase.testLdapKeyStoreCopyRemoveAlias failures if you don't see them locally when running all the manualmode tests I recommend examining the build log from a failed job and see what tests run prior to the one that fails. Maybe CI is running with a different order.

testsuite/manualmode/src/test/java/org/jboss/as/test/manualmode/vault/CustomVaultInModuleTestCase.java
```diff
@@ -113,10 +113,9 @@ public void testStandardRemoteExpressionResolution() throws Exception {
         attrValue = response.get("result", "attribute");
         Assert.assertEquals(response.toString(), ModelType.EXPRESSION, attrValue.getType());
         Assert.assertEquals(response.toString(), "${VAULT::Testing::Stuff::thing}", attrValue.asString());
-        // TODO uncomment when WFCORE-5305 is resolved
-        //responseHeaders = response.get("response-headers").asString();
-        //Assert.assertTrue(response.toString(), responseHeaders.contains("${VAULT::Testing::Stuff::thing}"));
-        //Assert.assertTrue(response.toString(), responseHeaders.contains("WFLYCTL0479"));
+        responseHeaders = response.get("response-headers").asString();
+        Assert.assertTrue(response.toString(), responseHeaders.contains("${VAULT::Testing::Stuff::thing}"));
+        Assert.assertTrue(response.toString(), responseHeaders.contains("WFLYCTL0479"));
```
bstansberry, line 118: To fully test this you'll need to add another resource level to this test extension and put an expression there as well. And then read recursively and confirm the warn for the child resource propagates.

fjuma:
@parsharma re the `LdapTestCase.testLdapKeyStoreCopyRemoveAlias` failure, that's an issue with the most recent version of OpenJDK 8. The following PR fixes it:

https://github.com/wildfly/wildfly-core/pull/4577

parsharma:
@fjuma Thanks :)

testsuite/manualmode/src/test/java/org/jboss/as/test/manualmode/vault/CustomVaultInModuleTestCase.java
parsharma: Please do let me know if more changes needed in the test case except for that scenario which is started in zulip chat.

parsharma: One scenario is still left which I have asked in the zulip chat.

parsharma:
@bstansberry CustomVaultInModuleTestCase.testStandardRemoteExpressionResolution() is not failing in my local, it only failing in CI.

testsuite/manualmode/src/test/java/org/jboss/as/test/manualmode/vault/CustomVaultInModuleTestCase.java
parsharma: Done

parsharma:
@bstansberry this PR is ready for review. Please let me know your review comments.
Thanks

yersan:
@parsharma Could you rebase and resolve the conflicts? thanks!

bstansberry:
I stuck the 'hold' label on this to save reviewers time. This was using an existing vault test to check the behavior but vault is gone so a new test design is needed.

parsharma:
@bstansberry, We won't be able to test this functionality because of this https://issues.redhat.com/browse/WFCORE-5379. Previously, we were testing in Vault, because Vault contains a fixed string and it is gone now. 
We need to wait for the above issue PR.

yersan:
Superseded by #5232

PR #4572, opened 2021-04-24
[WFLY-14431] Do not use WildFly Core dependency management for dependencies that are test-only in core

gaol:
Issue: https://issues.redhat.com/browse/WFLY-14431

That this PR does:

* Create a new module called: `testbom` which holds all test dependencies in core but are production dependencies in WildFly
* Added the dependency of `testbom` to modules that need it as `import` scope
* Move `commons-io`, `xalan`, `jackson-databind` from top level pom to `testbom/pom.xml` because they are test only dependencies in core, but they are production dependencies in WildFly.

### DIFF
--- a/testbom/pom.xml
+++ b/testbom/pom.xml
@@ -1,7 +1,8 @@
 <?xml version="1.0" encoding="UTF-8"?>
 <!--
+  ~
   ~ JBoss, Home of Professional Open Source.
-  ~ Copyright 2010, Red Hat, Inc., and individual contributors
+  ~ Copyright 2021, Red Hat, Inc., and individual contributors
   ~ as indicated by the @author tags. See the copyright.txt file in the
   ~ distribution for a full listing of individual contributors.
   ~
@@ -19,6 +20,7 @@
   ~ License along with this software; if not, write to the Free
   ~ Software Foundation, Inc., 51 Franklin St, Fifth Floor, Boston, MA
   ~ 02110-1301 USA, or see the FSF site: http://www.fsf.org.
+  ~
   -->
 
 <project xmlns="http://maven.apache.org/POM/4.0.0"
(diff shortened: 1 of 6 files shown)


gaol:
* The Full integration tests failure is because we need to [add commons-io dependency in WF root](https://github.com/wildfly/wildfly/compare/master...gaol:update_test_dependencies_inherited_from_core?expand=1)

* I am not sure about the failure of `The Dependency Tree Input Builder / check (pull_request)`, it maybe because the new `testbom` artifact is not recognized, that is why the CI tries to download it from remote repository.

testbom/pom.xml
```diff
@@ -0,0 +1,70 @@
...
+        <version.commons-io>2.5</version.commons-io>
+        <version.xalan>2.7.1.jbossorg-5</version.xalan>
+    </properties>
+
+    <dependencyManagement>
+        <dependencies>
+            <dependency>
+                <groupId>commons-io</groupId>
+                <artifactId>commons-io</artifactId>
+                <version>${version.commons-io}</version>
```
darranl, line 53: Should this pom have a policy that everything has scope test?  That way we know none if it can leak into the other uses.

darranl:
On the first error before we could accept a PR against WildFly Core we would need a PR against WildFly so we don't break CI at the point we merge assuming this PR is merged.

I think the reason for the second failure is there is no <module /> reference to the pom of the bom so it is not included in the build.

Regarding the PR I think it will take a bit of getting used to having test dependencies defined in a second location but our root dependency management sections are now very large so I think it will be good having a clear distinction between what is needed for the server and what is really a test only dependency.  TBH I am surprised the list of test only dependencies is not bigger.

testbom/pom.xml
gaol: Thanks @darranl, I think it makes sense to have every dependency in test scope in this bom, modules that depend on different scope should specify the scope explicitly, like [wildfly-core-testsuite-shared](https://github.com/wildfly/wildfly-core/blob/16.0.0.Beta3/testsuite/shared/pom.xml#L66).

gaol:
> I think it will be good having a clear distinction between what is needed for the server and what is really a test only dependency.

After comparing the dependencies between wildlfy-core and wildfly by checking the dependency scope, I see 3 such dependencies(addressed in this PR). We may also need to have an agreement on where to declare the test only dependencies in wildfly-core and wildfly.

> TBH I am surprised the list of test only dependencies is not bigger

This PR focuses on the dependencies that are used for testing in wildfly-core but are used for production in WildFly full. For dependencies like `junit`, `byteman` etc, are still declared in the root pom, this won't lead too much changes while fixing the problem.

FYI, I also started [a thread on wildfly-dev list](https://lists.jboss.org/archives/list/wildfly-dev@lists.jboss.org/thread/626ZRLCJE5NSNTSZX57J3IK7ICRFS63I/) about this change :)

jmesnil:
@gaol could you open a PR against WildFly for https://github.com/wildfly/wildfly/compare/master...gaol:update_test_dependencies_inherited_from_core?expand=1 please?

gaol:
@jmesnil sorry for late response, I was on holidays. Done at https://github.com/wildfly/wildfly/pull/14279, and I created https://issues.redhat.com/browse/WFLY-14771 to track the change.

.github/workflows/dep-diff-pull_request.yml
```diff
@@ -8,7 +8,7 @@ on:
     - master
 env:
   # The modules to check for dependencies. If there is more than one they are comma separated
-  MODULES_TO_CHECK: core-feature-pack/common
+  MODULES_TO_CHECK: core-feature-pack/common,testbom
```
gaol, line 11: After checking the workflow files, added `testbom` does not make much sense because:
* It contains `test` only dependencies
* It contains only `<dependencyManagement>`, so no dependency:list output to compare at all

As for the CI failure, I think we can ignore it, because this PR introduces a new artifact, but the step of `Build PR` tries to build the modules in `MODULES_TO_CHECK` only. I don't see a good way to solve it. :)

@bstansberry if you agree, I will reset this workflow file and trigger another CI run. :)

PR #4574, opened 2021-04-27
[WFCORE-2592] Special characters escaping in service.bat

lvydra:
https://issues.redhat.com/browse/WFCORE-2592

### DIFF
--- a/core-feature-pack/common/src/main/resources/content/docs/contrib/scripts/service/service.bat
+++ b/core-feature-pack/common/src/main/resources/content/docs/contrib/scripts/service/service.bat
@@ -382,13 +382,24 @@ if /I "%~1"== "/jbosspass" (
   if not "%~2"=="" (
     set T=%~2
     if not "!T:~0,1!"=="/" (
-      set JBOSSPASS=%~2
+      set "JBOSSPASS=%~2"
     )
   )
   if "!JBOSSPASS!" == "" (
     echo ERROR: You need to specify a password for /jbosspass
     goto endBatch
   )
+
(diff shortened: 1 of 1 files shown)


lvydra:
The purpose of this PR is to enable escaping of special characters, which are part of windows command line syntax, for jbosspass argument.
The whole password argument should be put in double quotes, then individual characters can be escaped by:

1.  ; and # which are part of procrun syntax can be escaped by putting them into single quotes (this works even before this PR)
2.  & < > | ^ ( ) ! as part of windows command line syntax can be escaped with ^
3.  other special characters like % , ; \ * ? . = [ ] are ok

core-feature-pack/common/src/main/resources/content/docs/contrib/scripts/service/service.bat
```diff
@@ -382,13 +382,24 @@ if /I "%~1"== "/jbosspass" (
...
+  for %%C in (^^^^ ^^^( ^^^) ^^^> ^^^< ^^^& ^^^|) do (
+    if not "!JBOSSPASS:%%C=!"=="!JBOSSPASS!" set IS_SANITIZED=true
+    set "JBOSSPASS=!JBOSSPASS:%%C=^^%%C!"
+  )
+  if not "!IS_SANITIZED!" == "true" (
+    for %%C in (^^ ^( ^) ^> ^< ^& ^|) do (
+      set "JBOSSPASS=!JBOSSPASS:%%C=^^^%%C!"
+    )
+  )
+
```
jamezp, line 402: I can't say I understand this, but I trust you :)

jamezp (approved):

PR #4575, opened 2021-04-28
WFCORE-5375 Update all projects with a dependency on deprecated Elytron modules to no longer rely on deprecated classes

parsharma:
Issue: https://issues.redhat.com/browse/WFCORE-5375


elytron/src/test/java/org/wildfly/extension/elytron/LdapTestCase.java
```diff
@@ -146,6 +147,7 @@ private static void loadResources() throws Exception {
     public static void startLdapService() throws Exception {
         loadResources();
         TestEnvironment.startLdapService();
+        Security.setProperty("keystore.PKCS12.keyProtectionAlgorithm", "PBEWithHmacSHA256AndAES_128");
```
fjuma, line 150: There's a PR that fixes `LdapTestCase` (https://github.com/wildfly/wildfly-core/pull/4577) so this change can be removed here.

core-feature-pack/common/pom.xml
```diff
@@ -301,10 +301,6 @@
             <groupId>org.wildfly.security</groupId>
             <artifactId>wildfly-elytron-credential</artifactId>
         </dependency>
-        <dependency>
-            <groupId>org.wildfly.security</groupId>
-            <artifactId>wildfly-elytron-credential-source-deprecated</artifactId>
-        </dependency>
```
fjuma, line 307: There are also a few more dependencies on other deprecated Elytron modules that should be removed here as well:
* `wildfly-elytron-auth-server-deprecated`
* `wildfly-elytron-http-deprecated`
* `wildfly-elytron-sasl-deprecated`
* `wildfly-elytron-x500-deprecated`

fjuma: The dependency on `wildfly-elytron-x500-deprecated` in `elytron/pom.xml` should also be removed.

pom.xml
```diff
@@ -1789,11 +1789,6 @@
                 <artifactId>wildfly-elytron-credential</artifactId>
                 <version>${version.org.wildfly.security.elytron}</version>
             </dependency>
-            <dependency>
-                <groupId>org.wildfly.security</groupId>
-                <artifactId>wildfly-elytron-credential-source-deprecated</artifactId>
-                <version>${version.org.wildfly.security.elytron}</version>
-            </dependency>
```
fjuma, line 1796: There are also a few more dependencies on other deprecated Elytron modules that should be removed here as well:
* `wildfly-elytron-auth-server-deprecated`
* `wildfly-elytron-http-deprecated`
* `wildfly-elytron-sasl-deprecated`
* `wildfly-elytron-x500-deprecated`

core-feature-pack/common/pom.xml
fjuma: The `wildfly-elytron-auth-server-deprecated` dependency should also be removed from `protocol/pom.xml` and `remoting/subsystem/pom.xml`.

fjuma (changes requested):
Thanks very much for working on this, @parsharma! I've added some comments about dependencies on other deprecated Elytron modules that should be removed as well.

elytron/pom.xml
```diff
@@ -189,6 +189,11 @@
             <artifactId>wildfly-elytron-sasl</artifactId>
         </dependency>
 
+        <dependency>
+            <groupId>org.wildfly.security</groupId>
+            <artifactId>wildfly-elytron-sasl-auth-util</artifactId>
+        </dependency>
```
fjuma, line 195: Just to check, is anything making use of this? I don't see any imports adjusted related to this one.

fjuma (commented):
@parsharma Thanks very much for working on this and for creating the WFCORE issues for the follow-up work! Just have one minor comment.

elytron/pom.xml
parsharma: Yes, I have checked no imports are dependent on this dependency. As I remember some test case was failing in wildly but I have re-checked everything is fine. I will push the updated one again.

PR #4577, opened 2021-04-29
[WFCORE-5383] Fix test failure in LdapTestCase that happens with openjdk-1.8.0.292

SoniaZaldana:
https://issues.redhat.com/browse/WFCORE-5383

### DIFF
--- a/elytron/src/test/java/org/wildfly/extension/elytron/LdapTestCase.java
+++ b/elytron/src/test/java/org/wildfly/extension/elytron/LdapTestCase.java
@@ -61,6 +61,7 @@ import java.security.KeyPairGenerator;
 import java.security.KeyStore;
 import java.security.MessageDigest;
 import java.security.PublicKey;
+import java.security.Security;
 import java.security.cert.Certificate;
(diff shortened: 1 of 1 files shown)


elytron/src/test/java/org/wildfly/extension/elytron/LdapTestCase.java
```diff
@@ -146,12 +147,14 @@ private static void loadResources() throws Exception {
     public static void startLdapService() throws Exception {
         loadResources();
         TestEnvironment.startLdapService();
+        Security.setProperty("keystore.PKCS12.keyProtectionAlgorithm", "PBEWithHmacSHA256AndAES_128");
```
fjuma, line 150: It would be good to save the old value that gets returned here and then make use of that value when cleaning up (e.g., see KeyStoresTestCase).

SoniaZaldana: Fixed!

fjuma (approved):
Thanks @SoniaZaldana!

PR #4578, opened 2021-05-03
[WFCORE-5386] Increase container shutdown wait time for better CI results stability

ropalka:
https://issues.redhat.com/browse/WFCORE-5386

### DIFF
--- a/controller/src/test/java/org/jboss/as/controller/test/AbstractControllerTestBase.java
+++ b/controller/src/test/java/org/jboss/as/controller/test/AbstractControllerTestBase.java
@@ -178,7 +178,7 @@ public abstract class AbstractControllerTestBase {
         if (container != null) {
             container.shutdown();
             try {
-                container.awaitTermination(5, TimeUnit.SECONDS);
+                container.awaitTermination(30, TimeUnit.SECONDS);
             } catch (InterruptedException e) {
                 e.printStackTrace();
             } finally {
--- a/controller/src/test/java/org/jboss/as/controller/test/AbstractProxyControllerTest.java
+++ b/controller/src/test/java/org/jboss/as/controller/test/AbstractProxyControllerTest.java
@@ -159,7 +159,7 @@ public abstract class AbstractProxyControllerTest {
         if (container != null) {
             container.shutdown();
             try {
-                container.awaitTermination(5, TimeUnit.SECONDS);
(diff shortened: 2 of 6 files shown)


controller/src/test/java/org/jboss/as/controller/test/AbstractControllerTestBase.java
```diff
@@ -178,7 +178,7 @@ public void shutdownServiceContainer() {
         if (container != null) {
             container.shutdown();
             try {
-                container.awaitTermination(5, TimeUnit.SECONDS);
+                container.awaitTermination(30, TimeUnit.SECONDS);
```
yersan, line 181: Did you check the possibility of wrapping the value by using [TimeoutUtil.adjust](https://github.com/wildfly/wildfly-core/blob/21a6c950c785ad24a67cf73c16c07854b5b60234/testsuite/shared/src/main/java/org/jboss/as/test/shared/TimeoutUtil.java#L52)?

It is suitable for timeouts and it will vary the amount based on an environment property adjustable from the caller CI Job

ropalka: I did but this is test suite utility class. The changes I did are in particular components and not in test suite @yersan .

yersan: ahh, I did not realize those tests were component tests, yes, you're right @ropalka , it makes sense to me now, thanks for clearing it up.

`RemotingLegacySubsystemTestCase` is still failing. In case it helps, without a deep understanding, maybe the root could be at jmx test cases, at least those tests are using the same port:

```
$ git grep 27258
jmx/src/test/java/org/jboss/as/jmx/JMXSubsystemTestCase.java:    private static final int JMX_PORT = 27258;
jmx/src/test/java/org/jboss/as/jmx/ModelControllerMBeanTestCase.java:    private static final int JMX_PORT = 27258;
remoting/tests/src/test/java/org/jboss/as/remoting/RemotingLegacySubsystemTestCase.java:                controllerInitializer.addSocketBinding("test", 27258);
remoting/tests/src/test/java/org/jboss/as/remoting/RemotingSubsystemTestCase.java:                controllerInitializer.addSocketBinding("remoting", 27258);
```

PR #4579, opened 2021-05-03
Fix for WFCORE-5384, RemoveManagementRealmTestCase tests intermittently fail

jfdenise:
Issue: https://issues.redhat.com/browse/WFCORE-5384

On some platform, we can enter the case where a read for input is attempted again (in the reload loop) although the console has been closed by a Ctrl-C. The closed state must be checked prior to attempt a read.


jfdenise:
The failures are unrelated to this change. I have run this test on the platform/JDK that showed the problem multiple times and didn't experience test failure:https://ci.wildfly.org/buildConfiguration/WildFlyCore_MasterLinuxS390OpenJ911/254607
No regression observed when running all tests: https://ci.wildfly.org/buildConfiguration/WildFlyCore_MasterLinuxS390OpenJ911/254617

PR #4591, opened 2021-05-15
[WFCORE-5406] Add --add-opens to the SE 9+ VM launch for expected ref…

bstansberry:
…lective access by the server

https://issues.redhat.com/browse/WFCORE-5406

PR #4595, opened 2021-05-17
[WFCORE-5410] Update elytron subsystem to take modular jdk params into account

ropalka:
https://issues.redhat.com/browse/WFCORE-5410

### DIFF
--- a/elytron/pom.xml
+++ b/elytron/pom.xml
@@ -463,7 +463,7 @@
                 <groupId>org.apache.maven.plugins</groupId>
                 <artifactId>maven-surefire-plugin</artifactId>
                 <configuration>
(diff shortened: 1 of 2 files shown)


ropalka:
@bstansberry please review

bstansberry:
@darranl @fjuma WDYT?

In https://github.com/wildfly/wildfly-core/pull/4591/files I did similiar things based on testing I was doing in full WF. But I confess with java.util and java.security I've forgotten the precise thing that led to the need for -add-opens, which is something that should be written down somewhere.

darranl (approved):

darranl:
I have added an approval here, this is just affecting the testsuite run so should be safe.

PR #4594, opened 2021-05-17
[WFCORE-5409]: Upgrade JGit to 5.11.1.202105131744-r

ehsavoie:
* upgrading JGit to 5.11.1.202105131744-r
 * upgrading Apache SSH to 2.6.0

Jira: https://issues.redhat.com/browse/WFCORE-5409

PR #4597, opened 2021-05-18
[WFCORE-5413] Upgrade Apache sshd-common to a version that depends on slf4j 1.7.30

SoniaZaldana:
https://issues.redhat.com/browse/WFCORE-5413


bstansberry:
#4594 also does this upgrade.

PR #4603, opened 2021-05-21
[WFCORE-5423][WFCORE-5424] Upgrade jboss-remoting to 5.0.23.Final and add test dependency from protocol to wildfly-elytron-security-manager

fl4via:
Jiras
https://issues.redhat.com/browse/WFCORE-5423 (see release notes below for further info on dependency upgrade)
https://issues.redhat.com/browse/WFCORE-5424

Fixes CVE-2020-35510


        Release Notes - JBoss Remoting (3+) - Version 5.0.23.Final
                                                            
<h2>        Task
</h2>
<ul>
<li>[<a href='https://issues.redhat.com/browse/REM3-381'>REM3-381</a>] -         Revert REM3-378 in Remoting 5.0.x
</li>
</ul>
                                                                                                                                                                                                                                        


        Release Notes - JBoss Remoting (3+) - Version 5.0.22.Final
                                                        
<h2>        Bug
</h2>
<ul>
<li>[<a href='https://issues.redhat.com/browse/REM3-379'>REM3-379</a>] -         CVE-2020-35510 Threads hold up forever in the EJB server by suppressing the ack from an EJB client
</li>
<li>[<a href='https://issues.redhat.com/browse/REM3-380'>REM3-380</a>] -         Add OutboundMessage info to messages logged by this class
</li>
</ul>
    
<h2>        Task
</h2>
<ul>
<li>[<a href='https://issues.redhat.com/browse/REM3-378'>REM3-378</a>] -         Updated all classes with a dependency on deprecated Elytron modules to no longer rely on deprecated classes
</li>
</ul>

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -194,7 +194,7 @@
         <version.org.jboss.marshalling.jboss-marshalling>2.0.12.Final</version.org.jboss.marshalling.jboss-marshalling>
         <version.org.jboss.modules.jboss-modules>1.11.0.Final</version.org.jboss.modules.jboss-modules>
         <version.org.jboss.msc.jboss-msc>1.4.12.Final</version.org.jboss.msc.jboss-msc>
-        <version.org.jboss.remoting>5.0.21.Final</version.org.jboss.remoting>
+        <version.org.jboss.remoting>5.0.23.Final</version.org.jboss.remoting>
         <version.org.jboss.remotingjmx.remoting-jmx>3.0.4.Final</version.org.jboss.remotingjmx.remoting-jmx>
         <version.org.jboss.shrinkwrap.shrinkwrap>1.2.6</version.org.jboss.shrinkwrap.shrinkwrap>
         <version.org.jboss.slf4j.slf4j-jboss-logmanager>1.1.0.Final</version.org.jboss.slf4j.slf4j-jboss-logmanager>
--- a/protocol/pom.xml
+++ b/protocol/pom.xml
@@ -126,6 +126,11 @@
            <artifactId>junit</artifactId>
            <scope>test</scope>
         </dependency>
+        <dependency>
+            <groupId>org.wildfly.security</groupId>
+            <artifactId>wildfly-elytron-security-manager</artifactId>
+            <scope>test</scope>
+        </dependency>
     </dependencies>
 
 </project>


bstansberry:
@fl4via Test failures seem related.

fl4via:
@bstansberry  the problem is that the CVE fix (REM-379) added a dependency to wildfly security manager. Is that a problem?
If it is, we could remove the configuration of the TIMEOUT, but with that, the fix is incomplete, basically the timeout won't be configurable and someone could have a problem with the read timeout. I added the dependency here to see if the tests pass for the time being, but please let me know what you think about it.

fl4via:
@bstansberry Also, I didn't know if the dependency was required in both galleon and modules, but I assumed so. I didn't have time to fully test this and see what as necessary and what was not.

fl4via:
Well, yeah, another possibility is just read the SystemProperty without the WildflySecurityManager, @bstansberry . For more info, please look at the code here https://github.com/jboss-remoting/jboss-remoting/commit/3d40eb63a7bda63396a731cd08f98f8a890de0d4#diff-a3aec2bf0520a1193f2fecd62304b19d028d180e8c5357baadd2a45d1862d40aR63 and here https://github.com/jboss-remoting/jboss-remoting/commit/3d40eb63a7bda63396a731cd08f98f8a890de0d4#diff-2ac747b8bbadb160b8ce13a4d93c5c0bb0537c36f083498c8b983f6c4e1efb5dR53
If that's the best approach I'll update jboss-remoting and release 5.0.24.Final with that change.

yersan:
Just for the record, `RemotingLegacySubsystemTestCase` is failing intermittently on JDK11, it is already tracked by WFCORE-5386

darranl:
JBoss Remoting has a hard dependency on WildFly Elytron so should be using the WildFlySecurityManager API to access the system properties.

Of the changes made I suspect the only dependency updated needed would be the protocol pom - the reason being poms in WildFly Core have just been updated to use the minimum set of Elytron dependencies.

The galleon layer should not be needed, that should just be about provisioning the subsystem - the modules themselves will be automatically provisioned by Galleon if anything has a dependency on them.

The additional module dependency should also not be needed as the security manager classes are already exposed via the elytron-private module.

I don't however see the original test failures, let me know if you would like one of us to take a close look into the build issues but overall Remoting should be using the Elytron APIs.

yersan:
> Of the changes made I suspect the only dependency updated needed would be the protocol pom 

I believe that even such dependency is also unneeded since the requirement of the security manager is on JBoss Remoting and not in WildFly Core. Such a requirement would be satisfied by JBoss Modules module at runtime.

fl4via:
@yersan Unfortunately, the test did not pass without the change in the protocol change. I'm going to remove the other two changes from the commit, @darranl , thanks for having a look!

fl4via:
@darranl  Btw the original failure is a NoClassDefFoundError thrown by RemoteConnectionHandler, at the line it invokes WildflySecurityManager, pointing out that the class could not be found. It was happening when running RemoteChannelManagementTestCase.

darranl:
@fl4via that error does sound to me like it should be the dependency in the pom on the protocol project - it may be possible that the scope only needs to be "test".

bstansberry (approved):
Thanks @fl4via and @darranl 

The root WF Core pom excludes's remoting's dependency on elytron, so definitely the protocol/pom.xml wouldn't pull in the needed elytron classes transitively, hence the need for the explicit declaration.

fl4via:
Thanks for the feedback, @bstansberry ! And thanks for your help today, @darranl . I hope that now the PR is good to go, as long as CI finishes running and passes :-)

darranl:
Overall just a FYI since we have split the Elytron project into separate artefacts our next step was to update the subsystems to just depend on the artefacts they use.  We wouldn't go to the point where every artefacts is a module but once we have a better understanding of how they are used we may be able to group into some smaller modules which Galleon can provision (or omit) on demand.

PR #4605, opened 2021-05-25
[WFCORE-5427] Avoid ClassCastException when determining whether or not a self-signed certificate should be automatically generated

fjuma:
https://issues.redhat.com/browse/WFCORE-5427


darranl (approved):

fjuma:
Just FYI, I've also added a test case since we didn't have any tests that make use of a `key-manager` that references an `ldap-key-store`.

PR #4615, opened 2021-06-05
[WFCORE-5441] Logic in try/finally blocks hides exceptions

boris-unckel:
Fixes https://issues.redhat.com/browse/WFCORE-5441

### DIFF
--- a/cli/src/main/java/org/jboss/as/cli/impl/CommandContextImpl.java
+++ b/cli/src/main/java/org/jboss/as/cli/impl/CommandContextImpl.java
@@ -1785,6 +1785,8 @@ public class CommandContextImpl implements CommandContext, ModelControllerClient
                 if (opLine != null && !opLine.equals(line)) {
                     resetArgs(opLine);
                 }
+
+                Throwable originalException = null;
                 try {
                     // Could be an operation.
                     if (exec.isOperation()) {
@@ -1797,7 +1799,13 @@ public class CommandContextImpl implements CommandContext, ModelControllerClient
                         handleLegacyCommand(exec.getLine(), handler, false);
                         continue;
                     }
+                } catch (Throwable ex) {
+                    if (ex instanceof InterruptedException) {
+                        Thread.currentThread().interrupt();
+                    }
+                    originalException = ex;
                 } finally {
+                    Throwable suppressed = originalException; //OK to be null
                     // We must close any output redirection, that is automaticaly done
                     // when calling exec.execute something that we are not doing here.
                     if (invContext.getConfiguration().getOutputRedirection() != null) {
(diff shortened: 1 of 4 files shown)


cli/src/main/java/org/jboss/as/cli/impl/CommandContextImpl.java
```diff
@@ -1785,34 +1785,53 @@ private void handleCommand(ParsedCommandLine parsed) throws CommandFormatExcepti
...
+                            handleOperation(parsedCmd);
+                            continue;
+                        }
+                        // Could be a legacy command.
+                        CommandHandler handler = exec.getLegacyHandler();
+                        if (handler != null) {
+                            handleLegacyCommand(exec.getLine(), handler, false);
+                            continue;
+                        }
+                    } catch (CommandLineException e) {
```
jfdenise, line 1806: @boris-unckel , I feel that we are not catching all exceptions there, CommandLineException is one typed exception, there are some more (CommandLineParserException, OptionValidatorException). We could also handle Runtime, Error and Interrupted).
What do you think of these changes: https://github.com/wildfly/wildfly-core/compare/master...jfdenise:WFCORE-5441

boris-unckel: Good Morning @jfdenise thanks for your reply. Two things from my side:
[CommandFormatException](https://github.com/wildfly/wildfly-core/blob/master/cli/src/main/java/org/jboss/as/cli/CommandFormatException.java#L28) extends CommandLineException and OptionValidatorException is not in the throws clause of the method, it should not occur here as it is a checked exception. Your're perfectly right for any RuntimeExceptions / Errors, my fault not considering it.

I'm aware that using a [Throwable originalException](https://github.com/wildfly/wildfly-core/compare/master...jfdenise:WFCORE-5441#diff-f5c14d363990d209745e98164e43fea634bfd7a31cf67f9eb50211b12bf133e9R1788) before the try block and use it in the finally block [originalException.addSuppressed(exception);](https://github.com/wildfly/wildfly-core/compare/master...jfdenise:WFCORE-5441#diff-f5c14d363990d209745e98164e43fea634bfd7a31cf67f9eb50211b12bf133e9R1822) is the regular way to resolve this. I have used the same pattern in all other solutions of this PR. I could reproduce a failing unit test [org.jboss.as.test.integration.management.cli.FilePermissionTestCase.testWriteFileIntoDirWithoutWritePermission()](https://github.com/wildfly/wildfly-core/blob/master/testsuite/standalone/src/test/java/org/jboss/as/test/integration/management/cli/FilePermissionTestCase.java#L63) Every attempt to get a valid reference to the IOException failed. I had to rethrow it. Does your code pass the unit test? Maybe I'm doing something wrong locally.

jfdenise: Perhaps there is a mix there, CommandLineParserException is not CommandFormatException. It extends Exception and is thrown from the exec.getLine() method (sam for OptionValidatorException). So it can occur in this try/finally block that these exceptions are thrown. You will notice that the main try/catch block at the beginning of the method catch those and throws a CommandFormatException that wrap those.  
The unit test passes with my changes.

boris-unckel: Thanks for the clarification. I missed the outer catch in my argumentation. How shall we proceed? I copy your solution, and adopt it to the other classes in the PR? Or I remove CommandContextImpl from my PR?

jfdenise: @boris-unckel , you can include my changes (it was a POC, a start of fix, you are free to iron it out).
And thank-you for finding and handling these issues, I also noticed that you found some invalid casts: https://issues.redhat.com/browse/WFCORE-5442 Well spotted!

domain-management/src/main/java/org/jboss/as/domain/management/security/LdapCacheService.java
```diff
@@ -356,6 +356,19 @@ private BaseSearchCache(final int evictionTime, final boolean cacheFailures, fin
                     }
                 }
             }
+            // Previous logic suppressed exceptions in finally statement,
```
darranl, line 359: I don't believe this part is needed, the change to the finally block to eliminate the return statement should be sufficient.

darranl (changes requested):
I am not really keen on changing LdapCacheService as it is deprecated and to be removed soon so we don't really want ongoing upstream changes but for now the changes seem to go further than is really needed for the original issue.

domain-management/src/main/java/org/jboss/as/domain/management/security/LdapCacheService.java
boris-unckel: Thanks for your feedback, I'll reduce it.

boris-unckel:
Hello @jfdenise @darranl please review again and approve. I think all issues should have been fixed now, checks have passed.

darranl:
@jfdenise I am happy with the security changes, I will let you add final approval as this touches more CLI classes.

jfdenise (approved):

jmesnil:
@boris-unckel thanks

PR #4616, opened 2021-06-05
[WFCORE-5442] Math operands should be cast before assignment

boris-unckel:
Fixes https://issues.redhat.com/browse/WFCORE-5442

### DIFF
--- a/cli/src/main/java/org/jboss/as/cli/impl/CommandContextImpl.java
+++ b/cli/src/main/java/org/jboss/as/cli/impl/CommandContextImpl.java
@@ -1469,7 +1469,7 @@ public class CommandContextImpl implements CommandContext, ModelControllerClient
             builder.addProperty(Util.NAME, Util.NAME);
 
             final long start = System.currentTimeMillis();
-            final long timeoutMillis = config.getConnectionTimeout() + 1000;
+            final long timeoutMillis = config.getConnectionTimeout() + 1000L;
             boolean tryConnection = true;
             while (tryConnection) {
                 final ModelNode response = client.execute(builder.buildRequest());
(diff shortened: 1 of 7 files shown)


cli/src/main/java/org/jboss/as/cli/impl/CommandContextImpl.java
```diff
@@ -1469,7 +1469,7 @@ private void tryConnection(final ModelControllerClient client, ControllerAddress
             builder.addProperty(Util.NAME, Util.NAME);
 
             final long start = System.currentTimeMillis();
-            final long timeoutMillis = config.getConnectionTimeout() + 1000;
+            final long timeoutMillis = config.getConnectionTimeout() + 1000L;
```
yersan, line 1472: To overflow the timeouts adding 1000 to the `config.getConnectionTimeout()` integer value, the user should configure an unreal timeout value

deployment-scanner/src/main/java/org/jboss/as/server/deployment/scanner/ZipCompletionScanner.java
```diff
@@ -283,7 +283,7 @@ private static boolean scanForEndSig(File file, FileChannel channel) throws IOEx
                 break;
             }
             lastChannelPos = channelPos;
-            channelPos -= Math.min(channelPos - bufferPos, CHUNK_SIZE - bufferPos);
+            channelPos -= Math.min(channelPos - bufferPos, (long) CHUNK_SIZE - bufferPos);
```
yersan, line 286: Here the `bufferPos` will never be higher than CHUNK_SIZE, the buffer is created with such capacity, so, it will never overflow the integer that results from the subtraction.

yersan (approved):
I've just added some minor notes, I don't think the changes are necessary, so probably the Jira could be optional. However, they make the code more compliant for code analyzer tools and are harmless changes, so I don't see any reason why they should not be merged.

jmesnil:
@boris-unckel thanks

PR #4618, opened 2021-06-05
[WFCORE-5445] Short circuit logic for boolean eval

boris-unckel:
Fixing https://issues.redhat.com/browse/WFCORE-5445

### DIFF
--- a/patching/src/main/java/org/jboss/as/patching/runner/AbstractPatchingTask.java
+++ b/patching/src/main/java/org/jboss/as/patching/runner/AbstractPatchingTask.java
@@ -152,7 +152,7 @@ abstract class AbstractPatchingTask<T extends ContentItem> implements PatchingTa
             return;
         }
         final PatchContentLoader contentLoader = description.getLoader();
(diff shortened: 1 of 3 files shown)


patching/src/main/java/org/jboss/as/patching/runner/AbstractPatchingTask.java
```diff
@@ -152,7 +152,7 @@ public void execute(final PatchingTaskContext context) throws IOException {
             return;
         }
         final PatchContentLoader contentLoader = description.getLoader();
-        final boolean skip = skipExecution | context.isExcluded(contentItem);
+        final boolean skip = skipExecution || context.isExcluded(contentItem);
```
yersan, line 155: This one could have been the more tricky to understand the possible original intention, but there the logic executed on the right side method is idempotent and it doesn't change the internal behavior, so it should be fine for the change.

yersan (approved):
I haven't found anything that could prevent merging these changes, but I'm not sure if any Wildfly @mergers can see a downside by merging these harmless changes.

jmesnil:
@boris-unckel thanks

PR #4621, opened 2021-06-08
Fix for WFCORE-5453, Include add-opens and add-exports to Bootable JAR runtime manifest

jfdenise:
Issue: https://issues.redhat.com/browse/WFCORE-5453
The set of add-opens and add-exports used in bin/common.sh launch script is added to Bootable JAR manifest file.
Having them in all cases for all JDK doesn't appear to be an issue. On JDK8, these entries are ignored.
Users can still add their own --add-opens --add-exports to the command line and both sources of options are honored.


jfdenise:
@ropalka , manifest file to contain the required options.

ropalka (approved):
LGTM

jmesnil:
@jfdenise thanks

PR #4622, opened 2021-06-08
[WFCORE-1934] Make number of max thread size for ServerService Thread…

yersan:
… Pool configurable

Jira: https://issues.redhat.com/browse/WFCORE-1934

### DIFF
--- a/server/src/main/java/org/jboss/as/server/ServerService.java
+++ b/server/src/main/java/org/jboss/as/server/ServerService.java
@@ -524,8 +524,11 @@ public final class ServerService extends AbstractControllerService {
     private static class ServerExecutorService implements Service<ExecutorService> {
 
         private static final int DEFAULT_CORE_POOL_SIZE = 1;
+        private static final int DEFAULT_MAX_POOL_SIZE = 1024;
         private static final int DEFAULT_DOMAIN_CORE_POOL_SIZE = 3; // keep more threads in a domain server as the intra-process comms use more tasks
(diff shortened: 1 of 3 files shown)


yersan:
@jmesnil In case there is any chance can review this one

server/src/main/java/org/jboss/as/server/ServerService.java
```diff
@@ -544,9 +547,10 @@ public synchronized void start(StartContext context) throws StartException {
             } else {
                 executorService = new EnhancedQueueExecutor.Builder()
                     .setCorePoolSize(getCorePoolSize(forDomain))
-                    .setMaximumPoolSize(1024)
+                    .setMaximumPoolSize(getMaxPoolSize())
                     .setKeepAliveTime(20L, TimeUnit.SECONDS)
                     .setThreadFactory(threadFactory)
+                    .setMBeanName(ENHANCED_EXECUTOR_MBEAN_NAME)
```
jmesnil, line 553: I was not sure why you add this: it allows to have a consistent name for the MBean instead of a autogenerated names with a counter.

jmesnil (approved):

jmesnil:
@yersan thanks!

server/src/main/java/org/jboss/as/server/ServerService.java
yersan: Yes, it facilitates the testing; we can do it because there is only one MBean per server JVM / managed server (domain mode) JVM

PR #4623, opened 2021-06-08
[WFCORE-5450] TlsTestCase is failing on IBM JDK

SoniaZaldana:
https://issues.redhat.com/browse/WFCORE-5450

### DIFF
--- a/elytron/src/test/resources/org/wildfly/extension/elytron/tls-ibm.xml
+++ b/elytron/src/test/resources/org/wildfly/extension/elytron/tls-ibm.xml
@@ -122,14 +122,15 @@
                                 session-timeout="321" maximum-session-cache-size="123"/>
             <server-ssl-context name="ServerSslContextSSLv2Hello" protocols="TLSv1 SSLv2Hello"
                                 key-manager="ServerKeyManager" trust-manager="CaTrustManager" want-client-auth="true" need-client-auth="true"
-                                authentication-optional="false" providers="ManagerProviderLoader" provider-name="IBMJSSE"
+                                authentication-optional="false" providers="ManagerProviderLoader" provider-name="IBMJSSE2"
                                 session-timeout="321" maximum-session-cache-size="123"/>
-            <server-ssl-context name="ServerSslContextDefaultProtocols"
+            <server-ssl-context name="ServerSslContextMultipleProtocolsNoSSLv2Hello"
                                 key-manager="ServerKeyManager" trust-manager="CaTrustManager" want-client-auth="true" need-client-auth="true"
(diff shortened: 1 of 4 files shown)


jmesnil:
@fjuma could you please review?

ropalka (approved):
LGTM

elytron/src/test/resources/org/wildfly/extension/elytron/tls-ibm.xml
```diff
@@ -122,14 +122,15 @@
                                 session-timeout="321" maximum-session-cache-size="123"/>
             <server-ssl-context name="ServerSslContextSSLv2Hello" protocols="TLSv1 SSLv2Hello"
                                 key-manager="ServerKeyManager" trust-manager="CaTrustManager" want-client-auth="true" need-client-auth="true"
-                                authentication-optional="false" providers="ManagerProviderLoader" provider-name="IBMJSSE"
+                                authentication-optional="false" providers="ManagerProviderLoader"
```
fjuma, line 125: To make sure the IBM provider is actually used for this test, instead of removing the `provider-name`, we could update it to `IBMJSSE2`. That seems to be the approach used for the other ssl-contexts configured in this file.

elytron/src/test/resources/org/wildfly/extension/elytron/tls-ibm.xml
```diff
@@ -149,7 +150,7 @@
             <client-ssl-context name="ClientSslContextOnlyTLS1" protocols="TLSv1" key-manager="ClientKeyManager"
                                 trust-manager="CaTrustManager" providers="ManagerProviderLoader" />
             <client-ssl-context name="ClientSslContextDefaultProtocols" key-manager="ClientKeyManager"
-                                trust-manager="CaTrustManager" providers="ManagerProviderLoader" />
+                                trust-manager="CaTrustManager" providers="ManagerProviderLoader" protocols="TLSv1 TLSv1.1 TLSv1.2"/>
```
fjuma, line 153: This change should also be added to the other `tls-*.xml` files to keep them all in sync. Since this `client-ssl-context` has been updated to explicitly configure protocols instead of just relying on the default protocols, I think it should be renamed from "ClientSslContextDefaultProtocols" to something else. It would also be good to double check that only the SSLv2Hello test is making use of this `client-ssl-context`. If another test was relying on the default protocol configuration, we should instead add a new `client-ssl-context` with the necessary protocols.

fjuma (changes requested):

SoniaZaldana:
hi @fjuma, I made the requested updates :)

fjuma (approved):
Thanks @SoniaZaldana!

SoniaZaldana:
@jmesnil I resolved it. Thanks!

PR #4630, opened 2021-06-11
WFCORE-5436: Clean up licenses

michpetrov:
Issue: [WFCORE-5436](https://issues.redhat.com/browse/WFCORE-5436)


core-feature-pack/common/src/main/resources/license/core-feature-pack-common-licenses.xml
```diff
@@ -209,7 +214,7 @@
           <distribution>repo</distribution>
         </license>
         <license>
-          <name>GNU General Public License, Version 2 with the Classpath Exception</name>
+          <name>GNU General Public License v2.0 only, with Classpath exception</name>
```
darranl, line 217: How did you identify this project is using the "only" variant and not the "or later" variant?

michpetrov: As far as I can tell "GPL 2.0 **or later** with Classpath Exception" is not a thing.

jmesnil: @michpetrov Where did you get that name?
I'm not found anything about a "v2.0 only". Eg https://openjdk.java.net/legal/gplv2+ce.html is named "GNU General Public License, version 2, with the Classpath Exception". I'm not sure to understand why we would change that name...

michpetrov: I took it from WildFly proper, e.g. https://github.com/wildfly/wildfly/blob/master/ee-feature-pack/common/src/main/resources/license/ee-feature-pack-common-licenses.xml#L363

the reason I'm changing it is that both of those come with their own license file and the version with "only" is more in line with the other GNU licenses we're using

PR #4636, opened 2021-06-15
[WFCORE-5467] Remove deprecated properties and methods

jasondlee:
https://issues.redhat.com/browse/WFCORE-5467

Remove props and methods
Delete noinspection comments
Fix lingering uses of deprecated property access

@darranl I'd appreciate a careful look at the Elytron-related changes to make sure I got those correct

### DIFF
--- a/elytron/src/main/java/org/wildfly/extension/elytron/AuthenticationClientDefinitions.java
+++ b/elytron/src/main/java/org/wildfly/extension/elytron/AuthenticationClientDefinitions.java
@@ -141,8 +141,7 @@ class AuthenticationClientDefinitions {
             .setFlags(AttributeAccess.Flag.RESTART_RESOURCE_SERVICES)
             .build();
 
-    static final PropertiesAttributeDefinition MECHANISM_PROPERTIES = new PropertiesAttributeDefinition.Builder(ElytronDescriptionConstants.PROPERTIES, true)
-            .setName(ElytronDescriptionConstants.MECHANISM_PROPERTIES)
+    static final PropertiesAttributeDefinition MECHANISM_PROPERTIES = new PropertiesAttributeDefinition.Builder(ElytronDescriptionConstants.MECHANISM_PROPERTIES, true)
             .setXmlName(ElytronDescriptionConstants.MECHANISM_PROPERTIES)
             .setRestartAllServices()
             .build();
--- a/controller/src/main/java/org/jboss/as/controller/AbstractAttributeDefinitionBuilder.java
+++ b/controller/src/main/java/org/jboss/as/controller/AbstractAttributeDefinitionBuilder.java
@@ -48,83 +48,33 @@ import org.jboss.dmr.ModelType;
 @SuppressWarnings("unchecked")
 public abstract class AbstractAttributeDefinitionBuilder<BUILDER extends AbstractAttributeDefinitionBuilder, ATTRIBUTE extends AttributeDefinition> {
 
-    /** @deprecated use {@link #getName()} as this field will be made private in a future release */
-    @Deprecated
-    protected String name;
-    /** @deprecated use {@link #getType()} as this field will be made private in a future release */
-    @Deprecated
-    protected ModelType type;
-    /** @deprecated use {@link #getXmlName()} as this field will be made private in a future release */
-    @Deprecated
-    protected String xmlName;
-    /** @deprecated use {@link #isAllowNull()} as this field will be made private in a future release */
-    @Deprecated
-    protected boolean allowNull;
-    /** @deprecated use {@link #isAllowExpression()} as this field will be made private in a future release */
-    @Deprecated
-    protected boolean allowExpression;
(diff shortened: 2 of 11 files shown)


darranl:
FYI especially related to clean up code to get the reviews through it may be easier to use separate PRs for separate areas.

darranl:
> FYI especially related to clean up code to get the reviews through it may be easier to use separate PRs for separate areas.

Ignore that, I see the Elytron changes are a consequence to the other change.

elytron/src/main/java/org/wildfly/extension/elytron/AuthenticationClientDefinitions.java
```diff
@@ -141,8 +141,7 @@
             .setFlags(AttributeAccess.Flag.RESTART_RESOURCE_SERVICES)
             .build();
 
-    static final PropertiesAttributeDefinition MECHANISM_PROPERTIES = new PropertiesAttributeDefinition.Builder(ElytronDescriptionConstants.PROPERTIES, true)
-            .setName(ElytronDescriptionConstants.MECHANISM_PROPERTIES)
+    static final PropertiesAttributeDefinition MECHANISM_PROPERTIES = new PropertiesAttributeDefinition.Builder(ElytronDescriptionConstants.MECHANISM_PROPERTIES, true)
```
darranl, line 144: I am not sure about this change, is this a case of a referenced attribute loosing it's name anyway otherwise this looks like a management model change to rename the attribute.

darranl:
I have been through the Elytron changes, they seem Ok.

Has any notification been sent for projects that may have been depending on these?  Some other projects add their own subsystems, something to the dev list my be sufficient but they may not watch the PR queues closely.

jmesnil:
@jasondlee the failures in the Full integration jobs show that we still have places in WildFly where these methods are called.

We need to find all their occurrences in WildFly and open WFLY issues against the impacted components to fix their codes before we can proceed with deleting them from your PR.

jasondlee:
> @jasondlee the failures in the Full integration jobs show that we still have places in WildFly where these methods are called.
> 
> We need to find all their occurrences in WildFly and open WFLY issues against the impacted components to fix their codes before we can proceed with deleting them from your PR.

I do have this, https://github.com/wildfly/wildfly/pull/14378, which I see failed overnight, so I need to address that, but, once that passes, does it sufficiently address your concerns?

jasondlee:
> I have been through the Elytron changes, they seem Ok.
> 
> Has any notification been sent for projects that may have been depending on these? Some other projects add their own subsystems, something to the dev list my be sufficient but they may not watch the PR queues closely.

I can definitely send an email to the dev list. Brian, who asked that I make the change, noted that these have been deprecated forever so there "shouldn't" be any existing uses (though I do have a small PR for WF full that belies that :), but another warning certainly couldn't hurt.

subsystem-test/framework/src/main/java/org/jboss/as/subsystem/test/SubsystemTestDelegate.java
```diff
@@ -584,8 +584,8 @@ public KernelServices build() throws Exception {
                         //we check the reject policy simulating a reject before adding it to the list of boot operations
                         if (!transformedOp.rejectOperation(SUCCESS)) {
                             transformedBootOperations.add(transformedOp.getTransformedOperation());
-//                        } else {
-//                            System.out.println(transformedOp.getFailureDescription());
+                        } else {
+                            System.out.println(transformedOp.getFailureDescription());
```
bstansberry, line 588: @jasondlee I don't think this should be uncommented.

jasondlee: You're right. I had a random test failure that I was trying to figure out, but it seems to have worked itself out and I forgot about that. Change incoming.

ide-configs/eclipse/wildfly-cleanup.xml
```diff
@@ -1,56 +1,121 @@
 <?xml version="1.0" encoding="UTF-8" standalone="no"?>
 <profiles version="2">
-<profile kind="CleanUpProfile" name="Wildfly" version="2">
```
jmesnil, line 3: could you remove the ide-configs changes from your PR please?

jmesnil (changes requested):
Code looks good to me. There are just changes in ide-configs that are not related to the issue. Could you please remove them and I'll merge the PR?

ide-configs/eclipse/wildfly-cleanup.xml
jasondlee: Absolutely. Not sure how those changes were made or how they slipped into the PR. :)

jasondlee: That must have been my branch being slightly out-of-date. Updated my fork and that change doesn't show anymore.

jmesnil (approved):

jmesnil:
@jasondlee thanks!

PR #4639, opened 2021-06-25
[WFCORE-5489] As a developer, I want to override management attribute values using environment variables

jmesnil:
When validating and setting the value for an attribute, check if there is an env var
that would override its value before doing any expression resolution or
correction.

JIRA: https://issues.redhat.com/browse/WFCORE-5489

### DIFF
--- a/controller/src/main/java/org/jboss/as/controller/AttributeDefinition.java
+++ b/controller/src/main/java/org/jboss/as/controller/AttributeDefinition.java
@@ -501,10 +501,26 @@ public abstract class AttributeDefinition {
      * @throws OperationFailedException if the value is not valid
      */
     public final void validateAndSet(ModelNode operationObject, final ModelNode model) throws OperationFailedException {
+        final boolean isAttributeBoundToAResource = operationObject.has(ModelDescriptionConstants.OP_ADDR);
+        PathAddress pathAddress = PathAddress.pathAddress(operationObject.get(ModelDescriptionConstants.OP_ADDR));
         if (operationObject.hasDefined(name) && deprecationData != null && deprecationData.isNotificationUseful()) {
-            ControllerLogger.DEPRECATED_LOGGER.attributeDeprecated(getName(),
-                    PathAddress.pathAddress(operationObject.get(ModelDescriptionConstants.OP_ADDR)).toCLIStyleString());
+            ControllerLogger.DEPRECATED_LOGGER.attributeDeprecated(getName(), pathAddress.toCLIStyleString());
         }
+
+        // overriding the value of an attribute using an env var is performed only when the operation
+        // containers a PathAddress. This method is also used for validating parameters (that are not necessary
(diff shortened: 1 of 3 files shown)


jmesnil:
Dev analysis is at https://github.com/wildfly/wildfly-proposals/pull/404

controller/src/main/java/org/jboss/as/controller/AttributeDefinition.java
```diff
@@ -501,10 +501,20 @@ public ModelNode validateOperation(final ModelNode operationObject) throws Opera
      * @throws OperationFailedException if the value is not valid
      */
     public final void validateAndSet(ModelNode operationObject, final ModelNode model) throws OperationFailedException {
+        PathAddress pathAddress = PathAddress.pathAddress(operationObject.get(ModelDescriptionConstants.OP_ADDR));
```
bstansberry, line 504: @jmesnil 

This concerns me a bit, as I think it's possible, maybe likely, that callers will pass in a custom operation object. Perhaps guard this with an 'operationObject.has(ModelDescriptionConstants.OP_ADDR)'.

Not 'hasDefined' as both DomainRootDefinition and ServerRootResourceDefinition have writable attributes 'name' and 'organization'

bstansberry: What I described above is definitely a problem, e.g. the standard write-attribute handling:

https://github.com/wildfly/wildfly-core/blob/main/controller/src/main/java/org/jboss/as/controller/AbstractWriteAttributeHandler.java#L93

ProcessNameWriteAttributeHandler does something similar. And then there is ReadChildrenResourceHandler which is using AttributeDefinition for *parameters* not for config attributes. That's a common use of that class.

jmesnil: I see your point.

This enhancement should not be done for configuring the management model and not for operation parameters.
Guarding against the presence of a `OP_ADDR` will ensure that we only trigger that feature when we have an attribute "bound" to a resource at that address.

jmesnil: @bstansberry I've updated the code to verify the presence of an `OP_ADDR`. I did not change the deprecation log but your comment made me realise that when we are validating a deprecated *parameter* (so without an `OP_ADDR` the message will not provide information on the location of the deprecated parameter).
I let it as it is as most of our deprecated AttributeDefinitions are most likely attributes rather than parameters

jmesnil:
After some thoughts, I'll close this PR and update the design.
This feature is only targeting the use of WildFly inside container images and is too disruptive for other use cases (eg bare metal) as it affects the persistent management model.

This feature should be disabled by default (and enabled by the presence of a env var). This ways, it can be explicitly enabled only when WildFly is running inside an image.

PR #4644, opened 2021-06-30
WFCORE-5493: Invoke DeploymentTransformer in operation "full-replace-…

fanste:
…deployment"

https://issues.redhat.com/browse/WFCORE-5493

### DIFF
--- a/server/src/main/java/org/jboss/as/server/deployment/DeploymentTransformation.java
+++ b/server/src/main/java/org/jboss/as/server/deployment/DeploymentTransformation.java
@@ -32,7 +32,7 @@ import org.jboss.dmr.ModelNode;
 
 import static org.jboss.as.server.deployment.DeploymentHandlerUtils.getInputStream;
 
-public class DeploymentTransformation {
+class DeploymentTransformation {
     @SuppressWarnings("deprecation")
     private final DeploymentTransformer deploymentTransformer;
 
--- a/server/src/main/java/org/jboss/as/server/deployment/DeploymentTransformation.java
+++ b/server/src/main/java/org/jboss/as/server/deployment/DeploymentTransformation.java
@@ -32,11 +32,11 @@ import org.jboss.dmr.ModelNode;
 
 import static org.jboss.as.server.deployment.DeploymentHandlerUtils.getInputStream;
 
-public class DeploymentTransformationUtil {
+public class DeploymentTransformation {
     @SuppressWarnings("deprecation")
     private final DeploymentTransformer deploymentTransformer;
 
-    public DeploymentTransformationUtil() {
+    public DeploymentTransformation() {
         this.deploymentTransformer = loadDeploymentTransformer();
     }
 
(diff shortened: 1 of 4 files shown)


ehsavoie (changes requested):
I think the code should be shared instead of being copied from DeploymentAddHandler

ehsavoie:
@jfdenise  may you take a look please ?

fanste:
> 
> 
> I think the code should be shared instead of being copied from DeploymentAddHandler

New superclass that is used by `DeploymentAddHandler` and `DeploymentFullReplaceHandler`?

`addFromContentAdditionParameter` is slightly different in both handlers. The only methods, that could be extracted are `loadDeploymentTransformer` and `doTransformation` and of course the variable `deploymentTransformer`.

ehsavoie:
I was thinking of all the transformation code: loading the transformer / transforming in some utility class like DeploymentHandlerUtil

jfdenise:
+1 for the Utility class.

fanste:
For the case, that no notification was sent out for my changes: I've moved the common code to a new utility class.

server/src/main/java/org/jboss/as/server/deployment/DeploymentTransformationUtil.java
```diff
@@ -0,0 +1,70 @@
...
+import org.jboss.as.controller.OperationContext;
+import org.jboss.as.controller.OperationFailedException;
+import org.jboss.as.server.controller.resources.DeploymentAttributes;
+import org.jboss.as.server.deployment.transformation.DeploymentTransformer;
+import org.jboss.as.server.logging.ServerLogger;
+import org.jboss.dmr.ModelNode;
+
+import static org.jboss.as.server.deployment.DeploymentHandlerUtils.getInputStream;
+
+public class DeploymentTransformationUtil {
```
jfdenise, line 35: Generally xxxUtil class only expose static methods. I would rename it to DeploymentTransformation, instance of this class has a state (the transformer).

server/src/main/java/org/jboss/as/server/deployment/DeploymentTransformation.java
```diff
@@ -0,0 +1,70 @@
...
+import org.jboss.as.controller.OperationContext;
+import org.jboss.as.controller.OperationFailedException;
+import org.jboss.as.server.controller.resources.DeploymentAttributes;
+import org.jboss.as.server.deployment.transformation.DeploymentTransformer;
+import org.jboss.as.server.logging.ServerLogger;
+import org.jboss.dmr.ModelNode;
+
+import static org.jboss.as.server.deployment.DeploymentHandlerUtils.getInputStream;
+
+public class DeploymentTransformation {
```
bstansberry, line 35: Please make this class package-protected. Thanks :)

fanste:
Rebased the PR and applied the change as requested by @bstansberry

bstansberry (approved):

jmesnil:
@fanste thanks

bstansberry:
@fanste 

The WildFly project would like to change its license from the GNU Lesser General Public License version 2.1 to the Apache Software License 2.0. We feel that ASL 2.0 is a better option for the project and our user community, as we find a project's use of LGPL can be a barrier to adoption in some parts of the software community, while we see no such barriers from use of ASL.

Do you agree to a relicensing of your contributions to the WildFly, WildFly Core or JBoss AS 7 projects from LGPL 2.1 to ASL 2.0?

I'm happy to answer any questions you may have about this, and I hope you are doing well. Thank you for your contributions to WildFly and open source!

Best regards,
Brian Stansberry
WildFly Project Lead

PR #4646, opened 2021-07-02
[WFCORE-5496] Upgrading JBoss MSC from 1.4.12.Final to 1.4.13.Final

ropalka:
https://issues.redhat.com/browse/WFCORE-5496

Brings in:
 * [MSC-256] Fix ServiceBuilder javadoc
 * [MSC-257] Unregistering container's WeakReference from shutdown hook containers set on container shutdown
 * [MSC-258] Use one shutdown hook per container and let JVM keep track of them


ropalka:
This is the proper fix of https://github.com/wildfly/wildfly-core/pull/4646

jmesnil:
@ropalka thanks

PR #4647, opened 2021-07-06
WFCORE-5353 Multiple syntax issues with 'wildfly-init-redhat.sh'

parsharma:
Issue: https://issues.redhat.com/browse/WFCORE-5353

### DIFF
--- a/core-feature-pack/common/src/main/resources/content/docs/contrib/scripts/init.d/wildfly-init-redhat.sh
+++ b/core-feature-pack/common/src/main/resources/content/docs/contrib/scripts/init.d/wildfly-init-redhat.sh
@@ -20,10 +20,10 @@
 # Source function library.
 . /etc/init.d/functions
 
-NAME=$(readlink -f ${0} | xargs basename)
+NAME=$(readlink -f "${0}" | xargs basename)
 
 # Check privileges
-if [ `id -u` -ne 0 ]; then
+if [ "$(id -u)" -ne 0 ]; then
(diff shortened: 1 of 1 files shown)


jfdenise:
@parsharma , I checked your changes, all seems good to me. I don't know if you want to get rid-off 100% of the warnings, https://www.shellcheck.net/ find some more (mainly related to echo -n and echo $").

parsharma:
@jfdenise thanks for reviewing it. I have updated and resolved most of the warnings only 4 warnings are left which are: 
 
Line 21:
. /etc/init.d/functions
  ^-- SC1091: Not following: /etc/init.d/functions was not specified as input (see shellcheck -x).
 
Line 38:
        . "$JBOSS_CONF"
          ^-- SC1090: ShellCheck can't follow non-constant source. Use a directive to specify location.
 
Line 146:
        printf "%s\n" $"Starting $NAME: "
                      ^-- SC3004: In POSIX sh, $".." is undefined.
 
Line 212:
        printf "%s\n" $"Shutting down $NAME: "
                      ^-- SC3004: In POSIX sh, $".." is undefined.
 
Line 245:
                printf "%s\n" $"$NAME is already stopped"
                              ^-- SC3004: In POSIX sh, $".." is undefined.

$

I think we are good to go with it. Please let me know if anything needs to change.

jmesnil (approved):

jmesnil:
@parsharma thanks

PR #4648, opened 2021-07-06
[WFCORE-5480] Bouncyclastle provider module requires a dependency on …

moulalis:
Issue: https://issues.redhat.com/browse/WFCORE-5480

### DIFF
--- a/core-feature-pack/common/src/main/resources/modules/system/layers/base/org/bouncycastle/bcprov/main/module.xml
+++ b/core-feature-pack/common/src/main/resources/modules/system/layers/base/org/bouncycastle/bcprov/main/module.xml
@@ -34,4 +34,8 @@
             <with-class name="org.bouncycastle.pqc.jcajce.provider.BouncyCastlePQCProvider"/>
         </service>
     </provides>
+    <dependencies>
+        <module name="java.naming"/>
+        <module name="java.sql"/>
+    </dependencies>
 </module>


yersan:
Just adding hold label since I would like to avoid this PR create a conflict with the BC upgrade

core-feature-pack/common/src/main/resources/modules/system/layers/base/org/bouncycastle/bcprov/main/module.xml
```diff
@@ -34,5 +34,9 @@
             <with-class name="org.bouncycastle.pqc.jcajce.provider.BouncyCastlePQCProvider"/>
         </service>
     </provides>
+    <dependencies>
+        <module name="java.sql"/>
+        <module name="java.naming"/>
+    </dependencies>
```
yersan, line 40: This is completely a nitpick and should not prevent this to be merged, but better if we keep the dependencies alphabetically when possible

yersan (approved):

jmesnil:
@moulalis could you please rebase your PR, a conflict was introduced after we merged #4628

moulalis (commented):
> @moulalis could you please rebase your PR, a conflict was introduced after we merged #4628

@jmesnil It's done. Thanks

jmesnil:
@moulalis thanks

PR #4650, opened 2021-07-06
[WFLY-5498] Upgrade commons-io to 2.10.0

bstansberry:
https://issues.redhat.com/browse/WFCORE-5498

### DIFF
--- a/testbom/pom.xml
+++ b/testbom/pom.xml
@@ -41,7 +41,7 @@
 
     <properties>
         <version.com.fasterxml.jackson>2.11.3</version.com.fasterxml.jackson>
-        <version.commons-io>2.5</version.commons-io>
+        <version.commons-io>2.10.0</version.commons-io>
         <version.io.netty>4.1.63.Final</version.io.netty>
         <version.org.apache.velocity>2.3</version.org.apache.velocity>
         <version.org.mock-server.mockserver-netty>5.8.1</version.org.mock-server.mockserver-netty>


jmesnil:
The failures look related to the commons.io upgrade (they fail when calling `org.apache.commons.io.FileUtils.forceDelete`)

bstansberry:
@jmesnil Yes, this is why we haven't been upgrading commons-io. I wanted to see if this latest release would have fixed this problem; seems not. But there's a valid need to upgrade in WildFly for production code (https://github.com/wildfly/wildfly/pull/14411) so we'll have to evaluate there whether we care about this behavior change, and either not upgrade anywhere, or just there, or change this test and upgrade here as well. In core it's just a test dep.

jmesnil:
at first glance, it seems commons-io 2.10.0 does not properly handle weird symlinks (that are specific to these tests).

Since we use common-io with the `<scope> test</scope>`, I think we can provide an alternative to correctly delete directories created by the tests and unblock the upgrade in wildfly full.

boris-unckel:
The situation has not change since 2.8.0: It depends on the filesystem: [https://github.com/wildfly/wildfly-core/pull/4400#issuecomment-770276813](https://github.com/wildfly/wildfly-core/pull/4400#issuecomment-770276813) You can check the dir with lsattr

gaol:
> Since we use common-io with the `<scope> test</scope>`, I think we can provide an alternative to correctly delete directories created by the tests and unblock the upgrade in wildfly full.

Agreed, I can take a look on fixing the tests

jmesnil:
this is superseded by #4653

PR #4654, opened 2021-07-09
[WFCORE-5482] Bouncycastle bcpkix module requires a dependency on java logging and java.naming

sudeshnas93:
Issue: https://issues.redhat.com/browse/WFCORE-5482

### DIFF
--- a/core-feature-pack/common/src/main/resources/modules/system/layers/base/org/bouncycastle/bcpkix/main/module.xml
+++ b/core-feature-pack/common/src/main/resources/modules/system/layers/base/org/bouncycastle/bcpkix/main/module.xml
@@ -32,5 +32,7 @@
     <dependencies>
         <module name="org.bouncycastle.bcprov" services="import"/>
         <module name="org.bouncycastle.bcutil"/>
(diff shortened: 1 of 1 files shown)


core-feature-pack/common/src/main/resources/modules/system/layers/base/org/bouncycastle/bcpkix/main/module.xml
```diff
@@ -32,5 +32,7 @@
     <dependencies>
         <module name="org.bouncycastle.bcprov" services="import"/>
         <module name="org.bouncycastle.bcutil"/>
+        <module name="java.naming"/>
+        <module name="java.sql"/>
```
soul2zimate, line 36: shouldn't this be `java.logging` instead of `java.sql`?

sudeshnas93: Sorry for the mistake. Thank you @soul2zimate :)

jmesnil:
@sudeshnas93 thanks

PR #4655, opened 2021-07-09
[WFCORE-5501] Remove legacy feature pack

jmesnil:
* Remove maven modules:
  * build-legacy
  * dist-legacy
  * core-feature-pack/legacy-feature-pack
* Remove subsystem resources and tests that were used only for these
  feature packs

JIRA: https://issues.redhat.com/browse/WFCORE-5501


core-feature-pack/ee-8-api/pom.xml
```diff
@@ -32,6 +32,7 @@
 
     <artifactId>wildfly-core-feature-pack-ee-8-api</artifactId>
     <packaging>pom</packaging>
+    <name>WildFly Core: Feature Pack for EE 8 API</name>
```
bstansberry, line 35: This is more "EE 8 API resources for Galleon Feature Packs"

bstansberry:
At a glance, removing the org.jboss.as.product stuff (module.xml and MANIFEST.MF) seems off topic. But I see the Galleon FP is not including a package for this module, and thus the build/dist doesn't either. So, we're already not providing that module from core, and this PR isn't changing that. So that's fine. (The relevant files you removed are from legacy build or dist, so for sure they would go. The removal of files with those names just caught my eye.

bstansberry:
@jmesnil Minor comment about the EE 8 API maven artifact description aside, this LGTM. I'm not github-approving it because it's still in draft state,, but please feel free to merge it.

core-feature-pack/ee-8-api/pom.xml
jmesnil: fixed, thanks

jmesnil:
@bstansberry thanks for the review. `org.jboss.as.product` was not ported to the Galleon pack (I'm not sure why but a good reason is that it was not need as the module is overriding the default values that are already in the WildFly Core codebase with the same values).

Once the CI passes, I'll merge the PR as it is.
I'll make a 2nd pass at the feature packs as I'd like to merge `wildfly-core-feature-pack-common` into `wildfly-core-feature-pack-galleon-common`. It does not make sense to keep a distinct `wildfly-core-feature-pack-common` module now that we only produce a Galleon feature pack.
However this one implies changes to WildFly (that consumes both modules). I'll do that in a separate PR to minimise the dependencies between the releases and PR to bring that change to WildFly.

PR #4661, opened 2021-07-13
[WFCORE-5468] Autocreate validator

jasondlee:
https://issues.redhat.com/browse/WFCORE-5468

In .setAllowedValues() add validator if one is not already set
Add IntAllowedValuesValidator
Add tests for StringAllowedValuesValidator and IntAllowedValuesValidator


jmesnil:
Failures are related to the PR.

Eg in `KeyStoresTestCase#testRevokeCertificateWithReason`, the operation sets the `reason` parameter to `KeyCompromise`while the attribute specifies the allowed value `KEYCOMPROMISE`.

@jasondlee please check with @fjuma how to handle this. Maybe the `reason` attribute should have an explicit validator that validate the value ignoring case?

fjuma:
@jasondlee The `reason` attribute is case insensitive. Currently, the value of this attribute is converted to an enum value using the following method:

https://github.com/wildfly/wildfly-core/blob/main/elytron/src/main/java/org/wildfly/extension/elytron/AdvancedModifiableKeyStoreDecorator.java#L897-L923

Instead of this approach, we could switch to using an `EnumValidator` instead for this attribute.

PR #4663, opened 2021-07-14
WFCORE-5502 Remove unnecessary unboxing to clean code

RanabirChakraborty:
Issue: https://issues.redhat.com/browse/WFCORE-5502

### DIFF
--- a/elytron/src/main/java/org/wildfly/extension/elytron/SSLDefinitions.java
+++ b/elytron/src/main/java/org/wildfly/extension/elytron/SSLDefinitions.java
@@ -1592,7 +1592,7 @@ class SSLDefinitions {
                     return false;
                 }
 
-                return isFips != null && isFips instanceof Boolean ? ((Boolean) isFips).booleanValue() : false;
+                return isFips instanceof Boolean ? (Boolean) isFips : false;
             };
 
         } catch (ClassNotFoundException | NoSuchMethodException | SecurityException e) {
--- a/cli/src/main/java/org/jboss/as/cli/handlers/BaseOperationCommand.java
+++ b/cli/src/main/java/org/jboss/as/cli/handlers/BaseOperationCommand.java
@@ -158,7 +158,7 @@ public abstract class BaseOperationCommand extends CommandHandlerWithHelp implem
         }
 
         if(available != null) {
-            return available.booleanValue();
+            return available;
         }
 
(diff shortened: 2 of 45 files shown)


RanabirChakraborty:
@boris-unckel Can you please review this

elytron/src/main/java/org/wildfly/extension/elytron/SSLDefinitions.java
```diff
@@ -1592,7 +1592,7 @@ private static BooleanSupplier getFipsSupplier() {
                     return false;
                 }
 
-                return isFips != null && isFips instanceof Boolean ? ((Boolean) isFips).booleanValue() : false;
+                return isFips != null && isFips instanceof Boolean ? (Boolean) isFips : false;
```
boris-unckel, line 1595: There is a double check, instanceof null returns false:
"return isFips instanceof Boolean ? (Boolean) isFips : false;" is the same.

RanabirChakraborty: The double check is intact, I just remove the unboxing.

boris-unckel: This seems to be a misunderstanding. I meant to remove the "isFips != null &&" because it's not necessary.

RanabirChakraborty: Thanks! done now.

boris-unckel: Please check https://ci.wildfly.org (Guest login possible). If you push --force your existing builds will not be canceled and the new ones will enter the queue. If a previous build can show you an error (source code formatter, compile error, unit test fail) it is better to combine both the learning from the build and the recommendations here.
Other builds (maybe more important) have the chance to run and you avoid your build to be canceled.

yersan (commented):
Hello @RanabirChakraborty , I have some difficulties reviewing this PR, and I can see some changes which are not strictly related to the current Jira description.

Could you justify them by adding a better Jira description? That for sure will help to review it.

If you end up with different descriptions, you could split the changes by commit, for example, changes related to avoiding invoking the deprecated constructor for primitive wrappers in one commit, changes to avoid doing redundant unboxing in another commit, changes to improve readability purpose in another commit.

PR #4670, opened 2021-07-15
[WFCORE-5513] Manage the BouncyCastle Jakarta Mail dependency with WildFly Core

yersan:
Jira issue: https://issues.redhat.com/browse/WFCORE-5513

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -1301,6 +1301,11 @@
             </dependency>
             <dependency>
                 <groupId>org.bouncycastle</groupId>
+                <artifactId>bcjmail-jdk15on</artifactId>
+                <version>${version.org.bouncycastle}</version>
+            </dependency>
+            <dependency>
+                <groupId>org.bouncycastle</groupId>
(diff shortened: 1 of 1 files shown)


yersan:
Added hold until getting a valid integration test that verifies wildfly full on wildfly-preview

bstansberry:
Just a note re the issue title / commit message -- this doesn't result in WF Core distributing this artifact. It just results in the WF Core pom attempting to manage the dependency for any other pom that uses its dependencyManagement (as full WF does.)

So "Manage the BouncyCastle Jakarta Mail dependency with WildFly Core" is more accurate.

yersan:
Thanks for the review @bstansberry , I've renamed the title and commit.

An integration Job with the wildfly full counterpart (https://github.com/wildfly/wildfly/pull/14495) is available here:
https://ci.wildfly.org/viewLog.html?buildId=264898&buildTypeId=WF_WildFlyCoreIntegrationExperiments

bstansberry (approved):

PR #4673, opened 2021-07-18
[WFCORE-5508] Add missing static final

boris-unckel:
Fixes https://issues.redhat.com/browse/WFCORE-5508

### DIFF
--- a/cli/src/main/java/org/jboss/as/cli/gui/JConsoleCLIPlugin.java
+++ b/cli/src/main/java/org/jboss/as/cli/gui/JConsoleCLIPlugin.java
@@ -74,7 +74,7 @@ public class JConsoleCLIPlugin extends JConsolePlugin {
 
     // Global count of created pools
     private static final AtomicInteger executorCount = new AtomicInteger();
-    private final Logger log = Logger.getLogger(JConsoleCLIPlugin.class.getName());
+    private static final Logger log = Logger.getLogger(JConsoleCLIPlugin.class.getName());
 
     private ConnectDialog dialog;
     CliGuiContext cliGuiCtx;
--- a/cli/src/main/java/org/jboss/as/cli/handlers/LsHandler.java
+++ b/cli/src/main/java/org/jboss/as/cli/handlers/LsHandler.java
@@ -63,7 +63,7 @@ public class LsHandler extends BaseOperationCommand {
     private final ArgumentWithValue nodePath;
     private final ArgumentWithoutValue l;
     private final ArgumentWithoutValue resolve;
-    private final Logger log = Logger.getLogger(LsHandler.class);
+    private static final Logger log = Logger.getLogger(LsHandler.class);
 
     public LsHandler(CommandContext ctx) {
         this(ctx, "ls");
--- a/cli/src/main/java/org/jboss/as/cli/impl/CommandContextImpl.java
+++ b/cli/src/main/java/org/jboss/as/cli/impl/CommandContextImpl.java
@@ -290,7 +290,7 @@ public class CommandContextImpl implements CommandContext, ModelControllerClient
     /** operation request address prefix */
     private final OperationRequestAddress prefix = new DefaultOperationRequestAddress();
     /** the prefix formatter */
-    private final NodePathFormatter prefixFormatter = DefaultPrefixFormatter.INSTANCE;
+    private static final NodePathFormatter prefixFormatter = DefaultPrefixFormatter.INSTANCE;
     /** provider of operation request candidates for tab-completion */
     private final OperationCandidatesProvider operationCandidatesProvider;
     /** operation request handler */
--- a/cli/src/main/java/org/jboss/as/cli/impl/DeploymentItemCompleter.java
+++ b/cli/src/main/java/org/jboss/as/cli/impl/DeploymentItemCompleter.java
@@ -42,7 +42,7 @@ import org.jboss.logging.Logger;
 public class DeploymentItemCompleter implements CommandLineCompleter {
 
     private final OperationRequestAddress address;
-    private final Logger log = Logger.getLogger(DeploymentItemCompleter.class);
+    private static final Logger log = Logger.getLogger(DeploymentItemCompleter.class);
 
     public DeploymentItemCompleter(OperationRequestAddress address) {
         this.address = address;
--- a/core-model-test/framework/src/main/java/org/jboss/as/core/model/test/TestModelControllerService.java
+++ b/core-model-test/framework/src/main/java/org/jboss/as/core/model/test/TestModelControllerService.java
@@ -408,7 +408,7 @@ class TestModelControllerService extends ModelTestModelControllerService {
     private class ServerInitializer implements Initializer {
         final ExtensibleConfigurationPersister persister = new NullConfigurationPersister();
         final ServerEnvironment environment = createStandaloneServerEnvironment();
-        final boolean parallelBoot = false;
+        static final boolean parallelBoot = false;
 
         public void setRootResourceDefinitionDelegate() {
             rootResourceDefinition.setDelegate(new ServerRootResourceDefinition(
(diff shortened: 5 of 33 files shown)


boris-unckel:
@jmesnil Hello Jeff, tests passed for this PR, failures seem not related. Could you review and approve, please?

cli/src/main/java/org/jboss/as/cli/gui/JConsoleCLIPlugin.java
```diff
@@ -74,7 +74,7 @@
 
     // Global count of created pools
     private static final AtomicInteger executorCount = new AtomicInteger();
-    private final Logger log = Logger.getLogger(JConsoleCLIPlugin.class.getName());
+    private static final Logger log = Logger.getLogger(JConsoleCLIPlugin.class.getName());
```
jmesnil, line 77: I'm not sure this is required: the `log` field is used as an instance field. We do not need to make it static.

If we do, let's rename it to `LOG` as this is our convention for constants.

cli/src/main/java/org/jboss/as/cli/handlers/LsHandler.java
```diff
@@ -63,7 +63,7 @@
     private final ArgumentWithValue nodePath;
     private final ArgumentWithoutValue l;
     private final ArgumentWithoutValue resolve;
-    private final Logger log = Logger.getLogger(LsHandler.class);
+    private static final Logger log = Logger.getLogger(LsHandler.class);
```
jmesnil, line 66: idem

cli/src/main/java/org/jboss/as/cli/impl/CommandContextImpl.java
```diff
@@ -290,7 +290,7 @@
     /** operation request address prefix */
     private final OperationRequestAddress prefix = new DefaultOperationRequestAddress();
     /** the prefix formatter */
-    private final NodePathFormatter prefixFormatter = DefaultPrefixFormatter.INSTANCE;
+    private static final NodePathFormatter prefixFormatter = DefaultPrefixFormatter.INSTANCE;
```
jmesnil, line 293: idem

cli/src/main/java/org/jboss/as/cli/impl/DeploymentItemCompleter.java
```diff
@@ -42,7 +42,7 @@
 public class DeploymentItemCompleter implements CommandLineCompleter {
 
     private final OperationRequestAddress address;
-    private final Logger log = Logger.getLogger(DeploymentItemCompleter.class);
+    private static final Logger log = Logger.getLogger(DeploymentItemCompleter.class);
```
jmesnil, line 45: idem

core-model-test/framework/src/main/java/org/jboss/as/core/model/test/TestModelControllerService.java
```diff
@@ -408,7 +408,7 @@ private void delete(File file) {
     private class ServerInitializer implements Initializer {
         final ExtensibleConfigurationPersister persister = new NullConfigurationPersister();
         final ServerEnvironment environment = createStandaloneServerEnvironment();
-        final boolean parallelBoot = false;
+        static final boolean parallelBoot = false;
```
jmesnil, line 411: I'm not sure we need to flag it as static.

core-model-test/framework/src/main/java/org/jboss/as/core/model/test/TestModelControllerService.java
```diff
@@ -443,7 +443,7 @@ public void initCoreModel(Resource rootResource, ManagementResourceRegistration
     }
 
     private class HostInitializer implements Initializer {
-        final String hostName = "master";
+        static final String hostName = "master";
```
jmesnil, line 446: idem

domain-management/src/main/java/org/jboss/as/domain/management/security/UserLdapCallbackHandler.java
```diff
@@ -88,7 +88,7 @@
 
     private final boolean allowEmptyPassword;
     private final boolean shareConnection;
-    protected final int searchTimeLimit = 10000; // TODO - Maybe make configurable.
+    protected static final int searchTimeLimit = 10000; // TODO - Maybe make configurable.
```
jmesnil, line 91: as far as I can tell, this field is not used at all. I'm not sure we need to make it static as it's not clear what its usage would be.

@darranl could you shed some light on it? Could we delete it completely?

host-controller/src/main/java/org/jboss/as/host/controller/RemoteDomainConnection.java
```diff
@@ -481,7 +481,7 @@ public void execute(ManagementRequestContext<Void> voidManagementRequestContext)
      private class CompleteRegistrationRequest extends AbstractManagementRequest<Void, Void> {
 
          private final byte outcome;
-         private final String message = "yay!"; //
+         private static final String message = "yay!"; //
```
jmesnil, line 484: static is not needed

protocol/src/main/java/org/jboss/as/protocol/mgmt/ManagementClientChannelStrategy.java
```diff
@@ -139,7 +139,7 @@ private static ProtocolConnectionConfiguration createConfiguration(final Protoco
      */
     private static class Establishing extends FutureManagementChannel {
 
-        private final String serviceType = DEFAULT_CHANNEL_SERVICE_TYPE;
+        private static final String serviceType = DEFAULT_CHANNEL_SERVICE_TYPE;
```
jmesnil, line 142: static is not needed

server/src/main/java/org/jboss/as/server/mgmt/domain/HostControllerConnection.java
```diff
@@ -327,7 +327,7 @@ public void execute(ManagementRequestContext<Void> voidManagementRequestContext)
 
     public class ServerStartedRequest extends AbstractManagementRequest<Void, Void> {
 
-        private final String message = ""; // started / failed message
+        private static final String message = ""; // started / failed message
```
jmesnil, line 330: static is not needed

jmesnil (changes requested):

cli/src/main/java/org/jboss/as/cli/gui/JConsoleCLIPlugin.java
boris-unckel: @jmesnil Thanks for your fast feedback. The pattern behind the missing static (both for LOG and the Strings below) is a final together with a direct assignment is indicating a constant. And constants shall be reused per class not per instance.
What is your thought when a immutable Object is assigned final to keep it as instance? Maybe I'm missing something.

domain-management/src/main/java/org/jboss/as/domain/management/security/UserLdapCallbackHandler.java
darranl: I am not going to be looking at this, at the moment my tasks are to remove legacy security.

darranl:
Some of these statics do seem more problematic than they are worth, in some cases it appears now references will be retained as long as the class is loaded instead for as long as there is an instance on the heap.

For these kinds of "clean up" it may be better to have a discussion on the wildfly-dev list - if there are bugs in the code fixes are great but when we are cleaning up based on third party coding conventions and recommendations it is probably worth checking how far this needs to go.

All of this code being changed is in the versions we need to maintain for support purposes with code being ported between these branches now at risk of merge conflicts - if we are not addressing a bug we should be sure of the benefit.

boris-unckel:
@jmesnil @darranl Thanks for feedback and suggestions how to deal with cases like this one in the future, I will follow that. I understand that cost/risk/benefit ratio is not enough to go forward with this PR. I withdraw it.

PR #4682, opened 2021-07-25
[WFCORE-5465] Using local repos used for git configuration persistenc…

bstansberry:
…e respect the default 'master' regardless of local gitconfig defaultbranch settings

https://issues.redhat.com/browse/WFCORE-5465

This is a draft as I'm suspicious we have other bugs; see added TODO in GitRepositoryTestCase. Plus this needs more thought in general to make sure the semantics are correct.

### DIFF
--- a/server/src/test/java/org/jboss/as/controller/persistence/GitPersistenceResourceTestCase.java
+++ b/server/src/test/java/org/jboss/as/controller/persistence/GitPersistenceResourceTestCase.java
@@ -46,7 +46,7 @@ public class GitPersistenceResourceTestCase extends AbstractGitPersistenceResour
         File baseDir = root.toAbsolutePath().toFile();
         File gitDir = new File(baseDir, Constants.DOT_GIT);
         if (!gitDir.exists()) {
-            try (Git git = Git.init().setDirectory(baseDir).setGitDir(gitDir).call()) {
+            try (Git git = Git.init().setDirectory(baseDir).setGitDir(gitDir).setInitialBranch(Constants.MASTER).call()) {
                 StoredConfig config = git.getRepository().getConfig();
                 config.setBoolean(ConfigConstants.CONFIG_COMMIT_SECTION, null, ConfigConstants.CONFIG_KEY_GPGSIGN, false);
                 config.save();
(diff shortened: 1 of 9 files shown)


bstansberry:
This is almost certainly wrong, at least in the production code, as it's doing things to work around WFCORE-5523. My instinct is the production code should not care what the default branch of a repo is; it should always ensure it's using the configured branch.

jmesnil (changes requested):
In case we have to specify explicitly the default branch, I'm not keen on defaulting to `master`.

We are moving away from using terms such as `master` in our codebase and don't want to reintroduce that term in this context. If we ever have to specify a name of a Git branch in our code base, I'd prefer we use `main`.

bstansberry:
I think we talked about this elsewhere but I'll respond here for the history.

First, the changes I made to production code -- i.e. setting the initial branch to master -- are wrong and are just workarounds to https://issues.redhat.com/browse/WFCORE-5523. So those will be removed in favor of ensuring the configured branch is always checked out.

Second, the default value of '--git-branch' is 'master' and changing that would be a breaking change. Note that the branch in question is the user's branch. We should evaluate changing that default but I think that should be handled separately from WFCORE-5465 and WFCORE-5523 work.

server/src/test/java/org/jboss/as/controller/persistence/GitPersistenceResourceTestCase.java
```diff
@@ -46,7 +46,7 @@ public void createDirectoriesAndFiles() throws Exception {
         File baseDir = root.toAbsolutePath().toFile();
         File gitDir = new File(baseDir, Constants.DOT_GIT);
         if (!gitDir.exists()) {
-            try (Git git = Git.init().setDirectory(baseDir).setGitDir(gitDir).call()) {
+            try (Git git = Git.init().setDirectory(baseDir).setGitDir(gitDir).setInitialBranch(Constants.MASTER).call()) {
```
ehsavoie, line 49: we should use the branch set on the command line which will default to master

bstansberry:
Closing this as @petrberan has taken over this work, see #4706

PR #4686, opened 2021-07-28
WFCORE-5527: ensure content folder is deleted on undeploy

michpetrov:
Issue: [WFCORE-5527](https://issues.redhat.com/browse/WFCORE-5527)


bstansberry:
This seems ok but I think we should wait for an approval by @ehsavoie (which may take a while given what month it is). He's the  one who added the test, and it seems quite deliberate, so he might remember a reason / use case why the behavior it's testing is correct.

In that test method I think the 'overlay' var and the path name it points to should be renamed to e.g. 'junk' and 'junk.txt'.  When thinking about this PR I wasted a lot of time examining how this might have something to do with deployment overlays.

This next bit isn't necessary for this PR but if you had time to add a commit doing it that would be nice. Rename src/text/resources overlay.html to something else and then change all the 'overlay' use in this test class. AFAICT nothing in that test class has anything to do with deployment overlays.

ehsavoie (changes requested):
Also I coded this on purpose as I don't usually like to remove stuff I didn't put there. I agree that the user or its OS is doing something 'wrong' so let's delete those files.

ehsavoie (approved):

PR #4692, opened 2021-07-30
[WFCORE-5524] *ConfigurationChangesHistoryTestCase fails on Windows

rmartinc:
Issue: https://issues.redhat.com/browse/WFCORE-5524

The `TreeSet` in `ConfigurationChangesCollector` has issues if there are changes in the same timestamp (this seems to happen more on windows). The [TreeSet](https://docs.oracle.com/javase/8/docs/api/java/util/TreeSet.html) needs the class that implements `Comparable` to be consistent with equals to work properly (and the `ConfigurationChange` class does not fulfill this statement).

The PR substitutes the `TreeSet` with a `Deque` (insertion order). The operation time is assigned when the change is added to the queue to maintain the time coherence in the list. If you prefer something different please let me know.

All tests pass OK in my windows env now.

@ehsavoie Can you please review this PR when you have time?

### DIFF
--- a/controller/src/main/java/org/jboss/as/controller/ConfigurationChangesCollector.java
+++ b/controller/src/main/java/org/jboss/as/controller/ConfigurationChangesCollector.java
@@ -34,9 +34,10 @@ import java.net.InetAddress;
 import java.time.Instant;
 import java.time.format.DateTimeFormatter;
 import java.time.format.DateTimeFormatterBuilder;
+import java.util.ArrayDeque;
+import java.util.Deque;
 import java.util.List;
 import java.util.Locale;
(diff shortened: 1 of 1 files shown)


ehsavoie:
The ConfigurationChange shouldn't be implementing Comparable since this is no longer used.

ehsavoie (changes requested):
The ConfigurationChange shouldn't be implementing Comparable since this is no longer used.

rmartinc:
> The ConfigurationChange shouldn't be implementing Comparable since this is no longer used.

Thanks @ehsavoie! It's done now.

ehsavoie (approved):

PR #4693, opened 2021-08-01
[WFCORE-5520] Avoid Nullpointer (Patching)

boris-unckel:

### DIFF
--- a/patching/src/main/java/org/jboss/as/patching/runner/PatchModuleInvalidationUtils.java
+++ b/patching/src/main/java/org/jboss/as/patching/runner/PatchModuleInvalidationUtils.java
@@ -67,7 +67,9 @@ class PatchModuleInvalidationUtils {
      * End of central directory record marker
      */
     public static final int GOOD_ENDSIG = 0x06054b50; // Good signature
+    public static final boolean FLAG_GOOD_ENDSIG = false;
     public static final int CRIPPLED_ENDSIG = 0x07054b50; // Crippled signature
+    public static final boolean FLAG_CRIPPLED_ENDSIG = true;
 
     /**
      * Length of the fixed portion of a local file header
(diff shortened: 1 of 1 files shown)


patching/src/main/java/org/jboss/as/patching/runner/PatchModuleInvalidationUtils.java
```diff
@@ -189,7 +189,7 @@ private static void updateJar(final File file, final byte[] searchPattern, final
                 if (!validateEndRecord(file, channel, pos, endSig)) {
                     pos = scanForEndSig(file, channel, context);
                 }
-                if (pos == -1) {
+                if (pos == -1 && context != null) {
```
bstansberry, line 192: The scanForEndSig call would fail if context were null.

I think the thing to do with this class is change L186 to throw an IllegalArgumentException. The purpose of this method is to write either CRIPPLED_ENDSIG or GOOD_ENDSIG; if neither is passed in it should fail.

bstansberry (changes requested):

patching/src/main/java/org/jboss/as/patching/runner/PatchModuleInvalidationUtils.java
bstansberry: Better is to replace the newSig param with a boolean and use CRIPPLED_ENDSIG or GOOD_ENDSIG based on that.

boris-unckel: @bstansberry Hello Brian, thanks for your recommendation. It's fixed. Could you review/merge, please?

jmesnil (approved):

bstansberry (approved):

PR #4694, opened 2021-08-01
[WFCORE-5520] Avoid Nullpointer (host-controller)

boris-unckel:

### DIFF
--- a/host-controller/src/main/java/org/jboss/as/host/controller/DirectoryGrouping.java
+++ b/host-controller/src/main/java/org/jboss/as/host/controller/DirectoryGrouping.java
@@ -52,7 +52,10 @@ public enum DirectoryGrouping {
     }
 
     public static DirectoryGrouping forName(String localName) {
-        final DirectoryGrouping directoryGrouping = localName != null ? MAP.get(localName.toLowerCase()) : null;
+        if (localName == null) {
+            return null;
+        }
+        final DirectoryGrouping directoryGrouping = MAP.get(localName.toLowerCase());
         return directoryGrouping == null ? DirectoryGrouping.valueOf(localName.toUpperCase(Locale.ENGLISH)) : directoryGrouping;
     }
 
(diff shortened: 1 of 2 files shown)


host-controller/src/main/java/org/jboss/as/host/controller/DirectoryGrouping.java
```diff
@@ -52,8 +51,10 @@
     }
 
     public static DirectoryGrouping forName(String localName) {
-        final DirectoryGrouping directoryGrouping = localName != null ? MAP.get(localName.toLowerCase()) : null;
-        return directoryGrouping == null ? DirectoryGrouping.valueOf(localName.toUpperCase(Locale.ENGLISH)) : directoryGrouping;
+        if (localName == null) {
+            return null;
+        }
+        return MAP.get(localName.toLowerCase());
```
yersan, line 57: This change is introducing an unexpected behavior change.

For example, of `localName` was something different of `null` and `MAP.get(localName.toLowerCase())` returns null, this method was returning `DirectoryGrouping.valueOf(localName.toUpperCase(Locale.ENGLISH))`, but now it will return null.

yersan (changes requested):

host-controller/src/main/java/org/jboss/as/host/controller/DirectoryGrouping.java
boris-unckel: @yersan Hi Yeray, thanks for your feedback. It's fixed. Could you review and approve, please?

yersan (approved):

host-controller/src/main/java/org/jboss/as/host/controller/DirectoryGrouping.java
yersan: Done, @boris-unckel, thanks!

PR #4695, opened 2021-08-01
[WFCORE-5520] Avoid Nullpointer (Remoting)

boris-unckel:


yersan (commented):
`connectorPropertiesOptionMap` is also used similarly on L179, if finally this change is accepted, the if sentence on L179 should be modified accordingly.

boris-unckel:
> `connectorPropertiesOptionMap` is also used similarly on L179, if finally this change is accepted, the if sentence on L179 should be modified accordingly.

@yersan thanks for your feedback. I have changed the PR accordingly. Please check again.

yersan (commented):
Hello @boris-unckel , looking more carefully at the source code, I don't believe that we are going to have a `connectorPropertiesOptionMap` as `null` in this piece of code. All the callers pass a non-null value.

In any case, preventing for null is fine and there is an if around L152 that does that. Instead of adding a repetitive null check for the same instance as you did on your changes, what about replacing the uses of `connectorPropertiesOptionMap` with `resultingMap` after the initial null check on L152 and avoid the repetitive null checks? What do you think?

If `connectorPropertiesOptionMap` is not null, then `resultingMap` will contain their values. If `connectorPropertiesOptionMap` is null, then `resultingMap` will be empty, so there is no need for further null checks. It seems that would make the code cleaner and will avoid repeating check if `connectorPropertiesOptionMap` when it is used along with this method.

boris-unckel:
Hello @yersan please have a look at this version. It relies on `resultingMap` now.

yersan:
Sorry @boris-unckel , this PR can be closed, we have been working on a piece of code that was going to be removed, see https://github.com/wildfly/wildfly-core/pull/4716

boris-unckel:
@yersan Anyway no problem anymore. Thanks for your ongoing support. :+1:

PR #4699, opened 2021-08-04
[WFCORE-4035] operation-requires-reload message is missing

petrberan:
https://issues.redhat.com/browse/WFCORE-4035


bstansberry (approved):

bstansberry:
RE the WF 24 wildscribe content, we don't update those. They reflect what the API reports from the relevant release, right or wrong.

Thanks, @petrberan

PR #4700, opened 2021-08-04
WFCORE-1562: preserve failure-description in 'resolve' operation result

michpetrov:
Issue: [WFCORE-1562](https://issues.redhat.com/browse/WFCORE-1562)

### DIFF
--- a/controller/src/main/java/org/jboss/as/controller/operations/global/GlobalOperationHandlers.java
+++ b/controller/src/main/java/org/jboss/as/controller/operations/global/GlobalOperationHandlers.java
@@ -332,7 +332,7 @@ public class GlobalOperationHandlers {
     private abstract static class AbstractAddressResolver implements OperationStepHandler {
 
         private static final FilterPredicate DEFAULT_PREDICATE = item -> !item.isDefined()
-                || !item.hasDefined(OP_ADDR);
+                || (!item.hasDefined(OP_ADDR) && !item.hasDefined(FAILURE_DESCRIPTION));
 
 
         private final ModelNode operation;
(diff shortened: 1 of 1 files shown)


controller/src/main/java/org/jboss/as/controller/operations/global/GlobalOperationHandlers.java
```diff
@@ -332,7 +332,7 @@ public void handleResult(OperationContext.ResultAction resultAction, OperationCo
     private abstract static class AbstractAddressResolver implements OperationStepHandler {
 
         private static final FilterPredicate DEFAULT_PREDICATE = item -> !item.isDefined()
-                || !item.hasDefined(OP_ADDR);
+                || (!item.hasDefined(OP_ADDR) && !item.hasDefined(FAILURE_DESCRIPTION));
```
bstansberry, line 335: The original purpose of this predicate was to avoid leaking information in violation of RBAC settings. See https://github.com/wildfly/wildfly/commit/442106fe77f766a8a0a7d7e8921e933dcb3f61de. Changing that logic concerns me.

Why is OP_ADDR not defined in something other than a case where it is being stripped for RBAC reasons?

michpetrov: The address is undefined until the [ResultHandler is executed](https://github.com/wildfly/wildfly-core/blob/a7c39a56decb28118cfa0899b44971bdd963d428/controller/src/main/java/org/jboss/as/controller/operations/global/GlobalOperationHandlers.java#L490)

In the case of this issue the `handler.execute(context, operation)` (line 484) throws OFE and so the address is never set. We could catch the OFE, set the address and rethrow, I assume that wouldn't cause issues with RBAC?

michpetrov: @bstansberry does my comment above make sense or are there still issues with that implementation?

PR #4702, opened 2021-08-05
[WFCORE-5536] list-snapshots() should include snapshot file with a cu…

soul2zimate:
…stom name (not only snapshot with timestamp prefix).

Issue: https://issues.redhat.com/browse/WFCORE-5536

It's possible to take a snapshot with a custom name, but list-snapshots() doesn't include them. I think it should pick files in snapshot directory with a more general filter.


bstansberry (changes requested):
@soul2zimate This LGTM but please add test coverage in org.wildfly.core.test.standalone.mgmt.api.ModelPersistenceTestCase. Thanks.

soul2zimate:
Thank you for the review, I have added a simple test to cover it.

bstansberry (approved):

PR #4708, opened 2021-08-16
[WFCORE-5543] Operation-scoped caching of static module Jandex indices

bstansberry:
https://issues.redhat.com/browse/WFCORE-5543

### DIFF
--- a/server/src/main/java/org/jboss/as/server/deployment/DeploymentHandlerUtil.java
+++ b/server/src/main/java/org/jboss/as/server/deployment/DeploymentHandlerUtil.java
@@ -51,6 +51,7 @@ import org.jboss.as.repository.ContentReference;
 import org.jboss.as.repository.ContentRepository;
 import org.jboss.as.server.ServerEnvironment;
 import org.jboss.as.server.controller.resources.DeploymentAttributes;
+import org.jboss.as.server.deployment.annotation.AnnotationIndexSupport;
 import org.jboss.as.server.deploymentoverlay.DeploymentOverlayIndex;
 import org.jboss.as.server.logging.ServerLogger;
 import org.jboss.dmr.ModelNode;
@@ -71,6 +72,8 @@ import org.jboss.vfs.VirtualFile;
  * @author <a href="mailto:ropalka@redhat.com">Richard Opalka</a>
  */
 public class DeploymentHandlerUtil {
+
+    private static final OperationContext.AttachmentKey<AnnotationIndexSupport> ANNOTATION_INDEX_SUPPORT = OperationContext.AttachmentKey.create(AnnotationIndexSupport.class);
 
     static class ContentItem {
         // either hash or <path, relativeTo, isArchive>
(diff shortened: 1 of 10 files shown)


bstansberry:
This is a draft as it should have some sort of test. Unfortunately the only way I can think of to test it would be to turn on debug login for ServerLogger.DEPLOYMENT_LOGGER and scan the logs.

server/src/main/java/org/jboss/as/server/deployment/annotation/CompositeIndexProcessor.java
```diff
@@ -124,21 +117,10 @@ public void deploy(DeploymentPhaseContext phaseContext) throws DeploymentUnitPro
...
             } else {
-                try {
-                    Module module = moduleLoader.loadModule(moduleIdentifier);
-                    final CompositeIndex additionalIndex = ModuleIndexBuilder.buildCompositeIndex(module);
-                    if (additionalIndex != null) {
-                        indexes.addAll(additionalIndex.indexes);
-                        additionalAnnotationIndexes.put(moduleIdentifier, additionalIndex);
-                    } else {
-                        final Index index = calculateModuleIndex(module);
-                        indexes.add(index);
```
bstansberry, line 135: @tadamski @fl4via @stuartwdouglas 

I don't see any reason why this index shouldn't be made available to WarAnnotationDeploymentProcessor via the additionalAnnotationIndexes map that's attached below via Attachments.ADDITIONAL_ANNOTATION_INDEXES_BY_MODULE. This PR has the effect of including it in the map.  

That DUP is looking for various EE annotations, and I don't see why ones found via processing a jandex.idx file should be relevant while those from class scanning are not.

Any objections? Am I missing anything?

Note that if this enhancement is ported to a maintenance branch I think the distinction should be retained, as we wouldn't want WarAnnotationDeploymentProcessor to start finding additional classes as a result of a performance enhancement.

bstansberry: FYI, I plan to update this PR to retain the existing distinct handling of indexes from jandex.idx vs class scanning. And then unless there's a reason not to unify treatment, file a separate issue/PR to unify. IOW make porting to maintenenance branches easier by not mixing two separate things in one PR.

server/src/main/java/org/jboss/as/server/deployment/annotation/AnnotationIndexSupport.java
```diff
@@ -0,0 +1,170 @@
...
+            Lock lock = getModuleIndexLock(moduleId);
+            try {
+                lock.lockInterruptibly();
+                try {
+                    // Check in case they were cached while we waited for the lock
+                    result = cache.get(moduleId);
+                    if (result == null) {
+                        // Nope. We build and cache the indices
+                        result = indexModule(moduleId, moduleLoader);
+                        cache.put(moduleId, result);
```
stuartwdouglas, line 91: AFAICS this is a HashMap, so the 'per key' lock is not enough, you need to synchronize on all operations, or change it to a CHM.

server/src/main/java/org/jboss/as/server/deployment/annotation/CompositeIndexProcessor.java
stuartwdouglas: Yea, the current behaviour looks like a bug. I don't think there is any reason behind it.

stuartwdouglas (changes requested):

server/src/main/java/org/jboss/as/server/deployment/RootDeploymentUnitService.java
```diff
@@ -46,6 +51,7 @@
     private final String managementName;
     private final DeploymentUnit parent;
     private final DeploymentOverlayIndex deploymentOverlays;
+    private final Map<String, Set<Index>> annotationIndexCache;
```
bstansberry, line 54: Heh; this is certainly wrong as it will prevent gc of the map!

server/src/main/java/org/jboss/as/server/deployment/annotation/AnnotationIndexSupport.java
bstansberry: oh, sure, of course. Thanks!

server/src/main/java/org/jboss/as/server/deployment/annotation/CompositeIndexProcessor.java
bstansberry: Thanks. I filed https://issues.redhat.com/browse/WFCORE-5545 for that.

server/src/main/java/org/jboss/as/server/deployment/DeploymentHandlerUtil.java
```diff
@@ -450,4 +459,13 @@ private static String getFormattedFailureDescription(OperationContext context) {
         }
         return hash;
     }
+
+    private static Map<String, Set<Index>> getAnnotationIndexCache(OperationContext context) {
+        Map<String, Set<Index>> result = context.getAttachment(ANNOTATION_INDEX_CACHE);
+        if (result == null) {
+            result = new HashMap<>();
```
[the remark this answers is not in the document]
bstansberry: I changed this a fair bit since you looked at this, basically to attach a single AnnotationIndexSupport here instead of what should be its internal map.

But I changed that internal map to a CHM; thanks.

stuartwdouglas (approved):

bstansberry:
I have added a test. Ported the log scanning utility from full WildFly.

PR #4710, opened 2021-08-17
WFCORE-5537 Improve error messages in LayersTest

parsharma:
Issue: https://issues.redhat.com/browse/WFCORE-5537

### DIFF
--- a/testsuite/shared/src/main/java/org/jboss/as/test/layers/LayersTest.java
+++ b/testsuite/shared/src/main/java/org/jboss/as/test/layers/LayersTest.java
@@ -105,8 +105,11 @@ public class LayersTest {
                 invalidUnref.add(unref);
             }
         }
+
+        // If an ERROR occurs please refer to unreferenced java doc for more details
         if (!invalidUnref.isEmpty()) {
-            throw new Exception("Some unreferenced modules are un-expected " + invalidUnref);
+            throw new Exception("!!ERROR, some modules have not been provisioned-- " +
+                    " Some unreferenced modules in installation [${project.build.directory}/layers-results/test-all-layers] are unexpected compared to [${project.build.directory}/layers-results/test-standalone-reference] " + invalidUnref);
         }
         StringBuilder builder = new StringBuilder();
         appendResult("REFERENCE", reference, builder, null);


testsuite/shared/src/main/java/org/jboss/as/test/layers/LayersTest.java
```diff
@@ -105,6 +105,7 @@ public static void test(String root, Set<String> unreferenced, Set<String> unuse
                 invalidUnref.add(unref);
             }
         }
+        // If error an occurs please have a look at "unreferenced" java doc
         if (!invalidUnref.isEmpty()) {
             throw new Exception("Some unreferenced modules are un-expected " + invalidUnref);
```
bstansberry, line 110: This should still provide some context, e.g. the path to the 'test-standalone-reference' and 'test-all-layers' installations that were compared, e.g.

"Some unreferenced modules in installation <path to test-all-layers> are unexpected compared to <path to test-standalone-reference> -- "

bstansberry (changes requested):
The "ERROR, some modules have not been provisioned: " message should also be improved.

testsuite/shared/src/main/java/org/jboss/as/test/layers/LayersTest.java
bstansberry: @parsharma I meant in the exception message.

bstansberry: @parsharma I meant in the exception message.

parsharma:
Thanks, @bstansberry.

testsuite/shared/src/main/java/org/jboss/as/test/layers/LayersTest.java
```diff
@@ -105,8 +105,11 @@ public static void test(String root, Set<String> unreferenced, Set<String> unuse
                 invalidUnref.add(unref);
             }
         }
+
+        // If an ERROR occurs please refer to unreferenced java doc for more details
         if (!invalidUnref.isEmpty()) {
-            throw new Exception("Some unreferenced modules are un-expected " + invalidUnref);
+            throw new Exception("!!ERROR, some modules have not been provisioned-- " +
+                    " Some unreferenced modules in installation [path to test-all-layers] are unexpected compared to [path to test-standalone-reference] " + invalidUnref);
```
bstansberry, line 112: Those should be the actually paths.

PR #4711, opened 2021-08-17
[WFCORE-5545] Include module indices computed via class scanning in the ADDITIONAL_ANNOTATION_INDEXES attachment

bstansberry:
https://issues.redhat.com/browse/WFCORE-5545

Builds on #4708


jmesnil:
@bstansberry thanks

PR #4714, opened 2021-08-20
Deprecate ModelTestControllerVersion-s older than EAP_7_4_0

rhusar:
Follow up on https://issues.redhat.com/browse/WFLY-15061


rhusar:
@bstansberry I assume we should deprecate and eventually remove these too?

bstansberry:
Thanks @rhusar! I filed a WFCORE and referenced it in the PR title so it will appear in the git history from the merge commit.

rhusar:
@bstansberry Ok thanks, I wasn't sure if this warrants a Jira but clearly it did, so thanks.

PR #4719, opened 2021-08-25
[WFCORE-4867] Highlight the config name in start message

khermano:
Issue: https://issues.redhat.com/browse/WFCORE-4867

### DIFF
--- a/core-feature-pack/common/src/main/resources/content/bin/domain.ps1
+++ b/core-feature-pack/common/src/main/resources/content/bin/domain.ps1
@@ -44,7 +44,11 @@ foreach($p in $HOST_CONTROLLER_OPTS){
 		$JBOSS_LOG_DIR=$p.Substring('-Djboss.domain.log.dir='.Length)
 	}elseif ($arg.StartsWith('-Djboss.domain.config.dir')){
 		$JBOSS_CONFIG_DIR=$p.Substring('-Djboss.domain.config.dir='.Length)
-	}
+	}elseif ($arg.StartsWith('--host-config')){
+    	$CONFIG_FILE=$p.Substring('--host-config='.Length)
+    }elseif ($arg.StartsWith('--domain-config')){
+    	$CONFIG_FILE=$p.Substring('--domain-config='.Length)
+    }
 }
 
 # If the -Djava.security.manager is found, enable the -secmgr and include a bogus security manager for JBoss Modules to replace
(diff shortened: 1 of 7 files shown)


yersan (commented):
What is the expected behavior if a configuration alias is in place? e.g if this were possible`./standalone.sh -c mp` 

Would not be better to handle this in the server code instead of in the script files?, well, for printing it in the "JBoss Bootstrap Environment" section, the scripts file are the only choice but I have my doubts if it would be better to handle this in the server code.

core-feature-pack/common/src/main/resources/content/bin/domain.ps1
```diff
@@ -44,7 +44,11 @@ foreach($p in $HOST_CONTROLLER_OPTS){
 		$JBOSS_LOG_DIR=$p.Substring('-Djboss.domain.log.dir='.Length)
 	}elseif ($arg.StartsWith('-Djboss.domain.config.dir')){
 		$JBOSS_CONFIG_DIR=$p.Substring('-Djboss.domain.config.dir='.Length)
-	}
+	}elseif ($arg.StartsWith('--host-config')){
+    	$CONFIG_FILE=$p.Substring('--host-config='.Length)
+    }elseif ($arg.StartsWith('--domain-config')){
+    	$CONFIG_FILE=$p.Substring('--domain-config='.Length)
```
yersan, line 50: Hi @khermano Independently of the final implementation, in the server code or in the scripts, I guess this won't work as expected since you can use both options `--host-config` and `--domain-config` together to start the domain.

I've also asked to get a wider description about what is the goal of this enhancement in the Jira tickethttps://issues.redhat.com/browse/WFCORE-4867 . The intention is still a bit unclear to me. I don't see yet the utility of just printing the used config file name when you are starting the server. feel free to share any feedback about its utility in the Jira, thanks!

yersan (changes requested):

core-feature-pack/common/src/main/resources/content/bin/domain.ps1
khermano: Hello @yersan. Thank you for comment I will take a look at that domain situation as soon as I contact Rostislav as I claim in JIRA comment. And sorry about the late response, I somehow managed to missed it completely.

khermano:
Ok, thank you very much. :) I will look at it and create a new PR.

PR #4729, opened 2021-08-31
[WFCORE-5570] Add a profile to disable compilation to facilitate test…

bstansberry:
…ing with a different JDKs

https://issues.redhat.com/browse/WFCORE-5570


jmesnil:
@bstansberry thanks

PR #4740, opened 2021-09-14
[WFCORE-5580] Fix conditional structure (controller)

boris-unckel:

### DIFF
--- a/controller/src/main/java/org/jboss/as/controller/capability/RuntimeCapability.java
+++ b/controller/src/main/java/org/jboss/as/controller/capability/RuntimeCapability.java
@@ -107,7 +107,7 @@ public class RuntimeCapability<T> extends AbstractCapability  {
         this.serviceValueType = serviceValueType;
         if (serviceValueType != null) {
             Assert.assertNotNull(baseServiceName);
-            this.serviceName = dynamicElement == null ? baseServiceName : baseServiceName.append(dynamicElement);
+            this.serviceName = baseServiceName.append(dynamicElement);
         } else {
             assert baseServiceName == null;
         }
--- a/controller/src/main/java/org/jboss/as/controller/parsing/DeferredExtensionContext.java
+++ b/controller/src/main/java/org/jboss/as/controller/parsing/DeferredExtensionContext.java
@@ -60,39 +60,34 @@ public class DeferredExtensionContext {
         if (!loaded) {
             loaded = true;
 
-            final Map<String, Future<XMLStreamException>> loadFutures = bootExecutor != null
-                    ? new LinkedHashMap<String, Future<XMLStreamException>>() : null;
+            final Map<String, Future<Void>> loadFutures = bootExecutor != null
+                    ? new LinkedHashMap<String, Future<Void>>() : null;
 
             for(ExtensionData extension : extensions) {
                 String moduleName = extension.moduleName;
                 XMLMapper xmlMapper = extension.xmlMapper;
                 if (loadFutures != null) {
                     // Load the module asynchronously
-                    Callable<XMLStreamException> callable = new Callable<XMLStreamException>() {
+                    Callable<Void> callable = new Callable<Void>() {
                         @Override
-                        public XMLStreamException call() throws Exception {
-                            return loadModule(moduleName, xmlMapper);
+                        public Void call() throws Exception {
+                            loadModule(moduleName, xmlMapper);
+                            return null;
                         }
                     };
-                    Future<XMLStreamException> future = bootExecutor.submit(callable);
+                    Future<Void> future = bootExecutor.submit(callable);
                     loadFutures.put(moduleName, future);
                 } else {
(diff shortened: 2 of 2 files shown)


controller/src/main/java/org/jboss/as/controller/capability/RuntimeCapability.java
```diff
@@ -106,8 +105,8 @@ private RuntimeCapability(String baseName, Class<?> serviceValueType, ServiceNam
         this.runtimeAPI = runtimeAPI;
         this.serviceValueType = serviceValueType;
         if (serviceValueType != null) {
-            Assert.assertNotNull(baseServiceName);
-            this.serviceName = dynamicElement == null ? baseServiceName : baseServiceName.append(dynamicElement);
+            assert baseServiceName != null;
```
yersan, line 108: What does motivate you to replace the uses of Assert.assertNotNull?

`Assert.assertNotNull` will log a message and it is widely used.

controller/src/main/java/org/jboss/as/controller/parsing/DeferredExtensionContext.java
```diff
@@ -126,7 +126,7 @@ private XMLStreamException loadModule(final String moduleName, final XMLMapper x
             }
             return null;
         } catch (final ModuleLoadException e) {
-            throw ControllerLogger.ROOT_LOGGER.failedToLoadModule(e);
+            return ControllerLogger.ROOT_LOGGER.failedToLoadModule(e);
```
yersan, line 129: Could you also explain the motivation for this change?

The original code intention was to throw the exception instead of returning it as the invocation result, so, I don't get the reason for changing it now. In addition, is a change unrelated to the original issue intention, I don't see it connected.

controller/src/main/java/org/jboss/as/controller/capability/RuntimeCapability.java
boris-unckel: @yersan The second assert was a native one. But you're right both should be Assert.assert based then.

controller/src/main/java/org/jboss/as/controller/parsing/DeferredExtensionContext.java
boris-unckel: @yersan It's related. The caller checks for the existence of the exception and throws it:
* [https://github.com/wildfly/wildfly-core/blob/62da6381bbbd5c5305e36bc25a03834f265d8b4e/controller/src/main/java/org/jboss/as/controller/parsing/DeferredExtensionContext.java#L82](https://github.com/wildfly/wildfly-core/blob/62da6381bbbd5c5305e36bc25a03834f265d8b4e/controller/src/main/java/org/jboss/as/controller/parsing/DeferredExtensionContext.java#L82)
* [https://github.com/wildfly/wildfly-core/blob/62da6381bbbd5c5305e36bc25a03834f265d8b4e/controller/src/main/java/org/jboss/as/controller/parsing/DeferredExtensionContext.java#L93](https://github.com/wildfly/wildfly-core/blob/62da6381bbbd5c5305e36bc25a03834f265d8b4e/controller/src/main/java/org/jboss/as/controller/parsing/DeferredExtensionContext.java#L93)

In the current implementation it is always null and therefor the if evaluates always to false.

controller/src/main/java/org/jboss/as/controller/capability/RuntimeCapability.java
boris-unckel: @yersan I have reverted to Assert.assertNotNull. There is no assertNull equivalent available.

controller/src/main/java/org/jboss/as/controller/parsing/DeferredExtensionContext.java
yersan: I see thanks, so it's a matter of removing the `if` and kept the exception be thrown or return the exception and keeping the `if`

The former looks cleaner to me; less code to read :)

In any case, since it is a private method both options would work fine. If finally, you choose for the second one, returning the exception, you have also to remove the throws clause at the method level.

boris-unckel: @yersan I have switched to the variant removing the `if`, please have a look. Tests passed except win full.

yersan (commented):
Changes look good. 

Windows Job is failing by a different cause (https://wildfly.zulipchat.com/#narrow/stream/174184-wildfly-developers/topic/Clustering.20tests.20Failed.20to.20delete.20server.2Elog), so we are missing around 1100 tests, being conservative, I'll wait to get the Windows one fixed before merging. I don't think these changes could affect windows, but, well, I don't have the proof.

jmesnil:
@yersan thanks.

We can merge this one: The `Full intregation -Linux` job is passing and I don't see changes in that PR that would be Windows-specific.

PR #4744, opened 2021-09-15
[WFCORE-5580] Fix conditional structure (remoting)

boris-unckel:


remoting/subsystem/src/main/java/org/jboss/as/remoting/RemotingSubsystem12Parser.java
```diff
@@ -83,10 +83,12 @@ void parseConnector(final XMLExtendedStreamReader reader, final ModelNode addres
                 }
                 case SERVER_NAME: {
                     ConnectorCommon.SERVER_NAME.parseAndSetParameter(value, connector, reader);
+                    serverName = value;
```
darranl, line 86: In this case as we have the parseAndSetParameter calls it looks like these variables should be removed entirely and the relevant if sections removed from below.

boris-unckel: @darranl I have changed it according to your recommendation.

PR #4745, opened 2021-09-15
[WFCORE-5580] Fix conditional structure (server)

boris-unckel:

### DIFF
--- a/server/src/main/java/org/jboss/as/server/ServerEnvironment.java
+++ b/server/src/main/java/org/jboss/as/server/ServerEnvironment.java
@@ -879,12 +879,8 @@ public class ServerEnvironment extends ProcessEnvironment implements Serializabl
     }
 
     private void configureBundlesDir(String dirPath, Properties providedProperties) {
-        boolean haveDirProperty = dirPath != null;
         File tmp = getFileFromPath(dirPath);
         if (tmp == null) {
-            if (haveDirProperty) {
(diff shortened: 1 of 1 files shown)


server/src/main/java/org/jboss/as/server/ServerEnvironment.java
```diff
@@ -879,12 +879,8 @@ public File getBundlesDir() {
     }
 
     private void configureBundlesDir(String dirPath, Properties providedProperties) {
-        boolean haveDirProperty = dirPath != null;
         File tmp = getFileFromPath(dirPath);
         if (tmp == null) {
-            if (haveDirProperty) {
-                throw ServerLogger.ROOT_LOGGER.bundlesDirectoryDoesNotExist(new File(dirPath).getAbsoluteFile());
```
yersan, line 886: I'm not sure about these changes. We should not lose this trace in case of calling this method with a null dirPath argument. And, in my opinion, although both methods are at the same class and are both private, we should not assume that if `getFileFromPath(dirPath)` returns null it means that `dirPath` was null.

If we remove this change, any possible future change on `getFileFromPath(dirPath)` would have a side effect on this one, and that can be inadvertent by the developer who is modifying the `getFileFromPath(dirPath)`

boris-unckel: @yersan ok, we keep it stay robust here. I'll close the PR.

PR #4746, opened 2021-09-15
[WFCORE-5580] Fix conditional structure (subsystem-test)

boris-unckel:


yersan (approved):
Looks fine, in any case, reading the `openStream()` API doc, my interpretation is it will never return null

PR #4747, opened 2021-09-15
[WFCORE-4867] Highlight the config name in start message

khermano:
Issue: https://issues.redhat.com/browse/WFCORE-4867

related to PR https://github.com/wildfly/wildfly-core/pull/4719

### DIFF
--- a/host-controller/src/main/java/org/jboss/as/host/controller/DomainModelControllerService.java
+++ b/host-controller/src/main/java/org/jboss/as/host/controller/DomainModelControllerService.java
@@ -897,11 +897,26 @@ public class DomainModelControllerService extends AbstractControllerService impl
                     Notification notification = new Notification(ModelDescriptionConstants.BOOT_COMPLETE_NOTIFICATION, PathAddress.pathAddress(PathElement.pathElement(CORE_SERVICE, MANAGEMENT),
                             PathElement.pathElement(SERVICE, MANAGEMENT_OPERATIONS)), ControllerLogger.MGMT_OP_LOGGER.bootComplete());
                     getNotificationSupport().emit(notification);
-                    bootstrapListener.printBootStatistics();
+
+                    String message;
+                    String hostConfig = environment.getHostConfigurationFile().getMainFile().getName();
+                    if (environment.getDomainConfigurationFile() != null) { //for slave HC is null
+                        String domainConfig = environment.getDomainConfigurationFile().getMainFile().getName();
+                        message = ROOT_LOGGER.configFilesInUse(domainConfig, hostConfig);
+                    } else {
+                        message = ROOT_LOGGER.configFileInUse(hostConfig);
+                    }
+                    bootstrapListener.printBootStatistics(message);
                 }
             } else {
                 // Die!
-                String failed = ROOT_LOGGER.unsuccessfulBoot();
+                String message;
+                if (environment.getDomainConfigurationFile() != null) {
+                    message = ROOT_LOGGER.configFilesInUse(environment.getDomainConfigurationFile().getMainFile().getName(), environment.getHostConfigurationFile().getMainFile().getName());
+                } else {
+                    message = ROOT_LOGGER.configFileInUse(environment.getHostConfigurationFile().getMainFile().getName());
+                }
+                String failed = ROOT_LOGGER.unsuccessfulBoot(message);
                 ROOT_LOGGER.fatal(failed);
                 bootstrapListener.bootFailure(failed);
 
--- a/host-controller/src/main/java/org/jboss/as/host/controller/logging/HostControllerLogger.java
+++ b/host-controller/src/main/java/org/jboss/as/host/controller/logging/HostControllerLogger.java
@@ -368,8 +368,8 @@ public interface HostControllerLogger extends BasicLogger {
     @Message(id = 33, value = "Caught exception during boot")
     void caughtExceptionDuringBoot(@Cause Exception e);
 
-    @Message(id = 34, value = "Host Controller boot has failed in an unrecoverable manner; exiting. See previous messages for details.")
-    String unsuccessfulBoot();
+    @Message(id = 34, value = "Host Controller boot has failed in an unrecoverable manner; exiting. See previous messages for details. %s")
+    String unsuccessfulBoot(String append);
 
     @LogMessage(level = Level.ERROR)
     @Message(id = 35, value = "Installation of the domain-wide configuration has failed. Because the running mode of this Host Controller is ADMIN_ONLY boot has been allowed to proceed. If ADMIN_ONLY mode were not in effect the process would be terminated due to a critical boot failure.")
@@ -1463,5 +1463,14 @@ public interface HostControllerLogger extends BasicLogger {
 
     @Message(id = 217, value = "Security realms are no longer supported, please migrate references to them from the configuration.")
     XMLStreamException securityRealmReferencesUnsupported();
+
+    ////////////////////////////////////////////////
+    //Messages without IDs
+
+    @Message(id = Message.NONE, value = "- Host Controller configuration files in use: %s, %s")
+    String configFilesInUse(String domainConfigFile, String hostConfigFile);
+
+    @Message(id = Message.NONE, value = "- Host Controller configuration file in use: %s")
+    String configFileInUse(String hostConfigFile);
 
 }
--- a/server/src/main/java/org/jboss/as/server/BootstrapListener.java
+++ b/server/src/main/java/org/jboss/as/server/BootstrapListener.java
@@ -68,7 +68,7 @@ public final class BootstrapListener {
         return monitor;
     }
 
-    public void printBootStatistics() {
+    public void printBootStatistics(String message) {
         final StabilityStatistics statistics = new StabilityStatistics();
         try {
             monitor.awaitStability(statistics);
@@ -78,7 +78,7 @@ public final class BootstrapListener {
         } finally {
             serviceTarget.removeMonitor(monitor);
             final long bootstrapTime = System.currentTimeMillis() - startTime;
-            done(bootstrapTime, statistics);
+            done(bootstrapTime, statistics, message);
             monitor.clear();
         }
     }
@@ -87,7 +87,7 @@ public final class BootstrapListener {
         futureContainer.failed(new Exception(message));
     }
 
-    private void done(final long bootstrapTime, final StabilityStatistics statistics) {
+    private void done(final long bootstrapTime, final StabilityStatistics statistics, String message) {
         futureContainer.done(serviceContainer);
         if (serviceContainer.isShutdown()) {
             // Do not print boot statistics because server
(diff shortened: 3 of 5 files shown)


rsvoboda (commented):
Having info about domain config after host config was a bit unexpected, but maybe it's the way it should be. Or it was process controller which was started first 🤔, not sure now.

```
[Host Controller] 11:30:11,026 INFO  [org.jboss.as] (MSC service thread 1-2) WFLYSRV0049: WildFly Core 17.0.0.Beta8-SNAPSHOT - host config: host.xml - starting
[Host Controller] 11:30:11,511 INFO  [org.wildfly.security] (Host Controller Service Threads - 4) ELY00001: WildFly Elytron version 1.17.0.Final
[Host Controller] 11:30:11,674 INFO  [org.jboss.as.config] (Controller Boot Thread) WFLYSRV0049: - domain config: domain.xml - starting
```

Overall no concerns around this PR. @yersan can you pls review this PR as you were looking into the initial PR https://github.com/wildfly/wildfly-core/pull/4719.

yersan (changes requested):
I think this still needs some changes, at least for the localization.

It would be great also to have clear first what we want to be printed and when.
Do we really want this information at the beginning and repeat it at the end?

Why not use only the latest information printed when the server or host controller has started (successfully or with an error). For example, for standalone:

`WFLYSRV0025: WildFly Full 26.0.0.Beta1-SNAPSHOT (WildFly Core 18.0.0.Beta1-SNAPSHOT) - started in 1977ms - Started 298 of 538 services (337 services are lazy, passive or on-demand) - Server configuration file in use: standalone.xml`

For a Host controller (which is also a DC):

`INFO  [org.jboss.as] (Controller Boot Thread) WFLYSRV0025: WildFly Full 26.0.0.Beta1-SNAPSHOT (WildFly Core 18.0.0.Beta1-SNAPSHOT) - (Host Controller) started in 5920ms - Started 71 of 72 services (14 services are lazy, passive or on-demand) - Host Controller configuration files in use: domain.xml, host-master.xml`

For a Host controller:
`INFO  [org.jboss.as] (Controller Boot Thread) WFLYSRV0025: WildFly Full 26.0.0.Beta1-SNAPSHOT (WildFly Core 18.0.0.Beta1-SNAPSHOT) - (Host Controller) started in 5920ms - Started 71 of 72 services (14 services are lazy, passive or on-demand) - Host Controller configuration files in use: host-slave.xml`

Other ideas? Thoughs?

host-controller/src/main/java/org/jboss/as/host/controller/DomainModelControllerService.java
```diff
@@ -897,7 +897,8 @@ public void execute(OperationContext context, ModelNode operation) throws Operat
                     Notification notification = new Notification(ModelDescriptionConstants.BOOT_COMPLETE_NOTIFICATION, PathAddress.pathAddress(PathElement.pathElement(CORE_SERVICE, MANAGEMENT),
                             PathElement.pathElement(SERVICE, MANAGEMENT_OPERATIONS)), ControllerLogger.MGMT_OP_LOGGER.bootComplete());
                     getNotificationSupport().emit(notification);
-                    bootstrapListener.printBootStatistics();
+                    bootstrapListener.printBootStatistics(true, environment.getDomainConfigurationFile().getMainFile().getName()
+                            + ", " + environment.getHostConfigurationFile().getMainFile().getName());
```
yersan, line 901: Wouldn't it be better to pass a variable argument and leave the format, in this case, separated by ", " to the logger itself?

I feel how the files will be printed is the responsibility of the logger message and not to the argument passed by to the printBootStatistics()

host-controller/src/main/java/org/jboss/as/host/controller/DomainModelControllerService.java
```diff
@@ -897,7 +897,8 @@ public void execute(OperationContext context, ModelNode operation) throws Operat
                     Notification notification = new Notification(ModelDescriptionConstants.BOOT_COMPLETE_NOTIFICATION, PathAddress.pathAddress(PathElement.pathElement(CORE_SERVICE, MANAGEMENT),
                             PathElement.pathElement(SERVICE, MANAGEMENT_OPERATIONS)), ControllerLogger.MGMT_OP_LOGGER.bootComplete());
                     getNotificationSupport().emit(notification);
-                    bootstrapListener.printBootStatistics();
+                    bootstrapListener.printBootStatistics(true, environment.getDomainConfigurationFile().getMainFile().getName()
```
yersan, line 900: environment.getDomainConfigurationFile() will be null if we are on an slave HC. You have to check this possibility.

host-controller/src/main/java/org/jboss/as/host/controller/DomainModelControllerService.java
```diff
@@ -897,7 +897,8 @@ public void execute(OperationContext context, ModelNode operation) throws Operat
...
+                            + ", " + environment.getHostConfigurationFile().getMainFile().getName());
                 }
             } else {
```
yersan, line 903: What about if there is a problem? It can be considered optional, but It would be nice to get the name of the configuration file in use if your server fails to start. Keep in mind if an HC fails to start before processing the domain configuration file, you will end here with a null environment.getDomainConfigurationFile()

khermano:
@yersan I added some changes earlier but I am not sure if you were notified about that so I rather letting you know. If you could look at that later when you have some spare time it would be great. Thank you and sorry for the delay.

server/src/main/java/org/jboss/as/server/BootstrapListener.java
```diff
@@ -104,10 +105,26 @@ private void done(final long bootstrapTime, final StabilityStatistics statistics
...
         if (failed == 0 && problem == 0) {
-            ServerLogger.AS_ROOT_LOGGER.startedClean(prettyVersion, bootstrapTime, started, active + passive + onDemand + never + lazy, onDemand + passive + lazy);
+            String append = "";
+            if (configFiles != null & !configFiles.isEmpty() & isHostController & isMaster & configFiles.size() == 2) {
+                append = ServerLogger.AS_ROOT_LOGGER.hostControllerMasterConfigFilesInUse(configFiles.get(0), configFiles.get(1));
+            } else if (configFiles != null & !configFiles.isEmpty() & isHostController & !isMaster & configFiles.size() == 1) {
+                append = ServerLogger.AS_ROOT_LOGGER.hostControllerSlaveConfigFileInUse(configFiles.get(0));
+            } else if (configFiles != null & !configFiles.isEmpty() & !isHostController & configFiles.size() == 1) {
+                append = ServerLogger.AS_ROOT_LOGGER.serverConfigFileInUse(configFiles.get(0));
+            }
```
yersan, line 115: This piece of code is repeated in an `if/else`. You can extract it from that flow control structure to avoid redundancy.
In addition, && is preferred here since it will break the shortcut as soon as it detects on operand as false, instead of & that will evaluate all the operands aways.

server/src/main/java/org/jboss/as/server/BootstrapListener.java
```diff
@@ -68,7 +69,7 @@ public StabilityMonitor getStabilityMonitor() {
         return monitor;
     }
 
-    public void printBootStatistics() {
+    public void printBootStatistics(boolean isHostController, boolean isMaster, List<String> configFiles) {
```
yersan, line 72: Remove the `boolean isMaster`, you don't need it, you can infer the behavior by using `configFiles.size() == 2`, (configFiles.size() == 2, then you are in a master DC)

server/src/main/java/org/jboss/as/server/BootstrapListener.java
yersan: Removing the `isMaster` argument, you could arrive at something like the following:

```
String append = "";
if (configFiles != null && !configFiles.isEmpty()) {
    if (isHostController) {
        if (configFiles.size() == 2)
          //  master host contoller message
        else {
          //  slave host controller message
        }
    } else {
        // server message
    }
}
```

That will facilitate the debugging and it is more redeable

server/src/main/java/org/jboss/as/server/ServerService.java
```diff
@@ -424,10 +424,18 @@ public void undeploy(DeploymentUnit context) {
             Notification notification = new Notification(ModelDescriptionConstants.BOOT_COMPLETE_NOTIFICATION, PathAddress.pathAddress(PathElement.pathElement(CORE_SERVICE, MANAGEMENT),
                     PathElement.pathElement(SERVICE, MANAGEMENT_OPERATIONS)), ServerLogger.AS_ROOT_LOGGER.bootComplete());
             getNotificationSupport().emit(notification);
-            bootstrapListener.printBootStatistics();
+            List<String> configFiles = new ArrayList<String>();
```
yersan, line 427: Minor detail, you can use Java Type inference capabilities, so instead of  `new ArrayList<String>()` use `new ArrayList<>()` here.

host-controller/src/main/java/org/jboss/as/host/controller/logging/HostControllerLogger.java
```diff
@@ -1464,4 +1464,13 @@ void noDomainControllerConfigurationProvidedForAdminOnly(String policyAttribute,
     @Message(id = 217, value = "Security realms are no longer supported, please migrate references to them from the configuration.")
     XMLStreamException securityRealmReferencesUnsupported();
 
+    ////////////////////////////////////////////////
+    //Messages without IDs
+
+    @Message(id = Message.NONE, value = " Config files in use: %s, %s")
```
yersan, line 1470: s/Config files/Host Controller configuration files

host-controller/src/main/java/org/jboss/as/host/controller/logging/HostControllerLogger.java
```diff
@@ -1464,4 +1464,13 @@ void noDomainControllerConfigurationProvidedForAdminOnly(String policyAttribute,
...
+    String configFilesInUse(String domainConfigFile, String hostConfigFile);
+
+    @Message(id = Message.NONE, value = " Config file in use: %s")
```
yersan, line 1473: s/Config file/Host Controller configuration file

host-controller/src/main/java/org/jboss/as/host/controller/DomainModelControllerService.java
```diff
@@ -897,11 +897,22 @@ public void execute(OperationContext context, ModelNode operation) throws Operat
                     Notification notification = new Notification(ModelDescriptionConstants.BOOT_COMPLETE_NOTIFICATION, PathAddress.pathAddress(PathElement.pathElement(CORE_SERVICE, MANAGEMENT),
                             PathElement.pathElement(SERVICE, MANAGEMENT_OPERATIONS)), ControllerLogger.MGMT_OP_LOGGER.bootComplete());
                     getNotificationSupport().emit(notification);
-                    bootstrapListener.printBootStatistics();
+                    List<String> configFiles = new ArrayList<String>();
```
yersan, line 900: Minor detail, you can use Java Type inference capabilities, so instead of `new ArrayList<String>()` use `new ArrayList<>()` here.

host-controller/src/main/java/org/jboss/as/host/controller/logging/HostControllerLogger.java
yersan: Remove the extra white space at the beginning

host-controller/src/main/java/org/jboss/as/host/controller/logging/HostControllerLogger.java
yersan: Remove the extra white space at the beginning

yersan (changes requested):

yersan (approved):
@jmesnil / @bstansberry Changes on this PR look good to me, but I would like to get your feedback before merging. With this PR we were going to get these logs:

Standalone:

```
09:10:00,539 INFO  [org.jboss.as] (Controller Boot Thread) WFLYSRV0025: WildFly Core 18.0.0.Beta4-SNAPSHOT started in 1550ms - Started 80 of 82 services (16 services are lazy, passive or on-demand) - Server configuration file in use: standalone.xml
```

Domain Mode:

```
[Host Controller] 09:12:13,221 INFO  [org.jboss.as] (Controller Boot Thread) WFLYSRV0025: WildFly Core 18.0.0.Beta4-SNAPSHOT (Host Controller) started in 4438ms - Started 70 of 72 services (15 services are lazy, passive or on-demand) - Host Controller configuration files in use: domain.xml, host.xml
```

jmesnil (approved):
@yersan we are not losing any existing info and add an important one. I'm fine with that changes.

@bstansberry do you see any issue with adding info to this essential `WFLYSRV0025` log?

jmesnil:
@khermano thanks

PR #4754, opened 2021-09-19
[WFCORE-5575] Refactor Boolean parsing and Bool object use (domain-http)

boris-unckel:
https://issues.redhat.com/browse/WFCORE-5575

### DIFF
--- a/domain-http/interface/src/main/java/org/jboss/as/domain/http/server/security/RealmReadinessHandler.java
+++ b/domain-http/interface/src/main/java/org/jboss/as/domain/http/server/security/RealmReadinessHandler.java
@@ -47,7 +47,7 @@ abstract class RealmReadinessHandler implements HttpHandler {
 
     @Override
     public void handleRequest(HttpServerExchange exchange) throws Exception {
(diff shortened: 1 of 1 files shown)


jmesnil (commented):
I'm not sure the use of `Boolean` here is worth it. The current code reads better

boris-unckel:
@jmesnil thanks for your feedback

> I'm not sure the use of `Boolean` here is worth it. The current code reads better

The current implementation with autoboxing is not null safe, thats the intention of the change. I fully agree the readability is much better.
At the moment it looks like a static (never null) value is here: [o.j.a.d.h.s.ManagementHttpServer.java#L277](https://github.com/wildfly/wildfly-core/blob/17.0.0.Beta7/domain-http/interface/src/main/java/org/jboss/as/domain/http/server/ManagementHttpServer.java#L277) but this might change.

I'll undo all other changes comparable to this one - but weighing null-safety/robustness vs. readability you're sure you want readability in these cases?

boris-unckel:
I'll change the other PRs accordingly to your feedback. This has only one (non-readable) change, therefor I withdraw it.

PR #4766, opened 2021-09-20
[WFCORE-5343]: YAML support for configuration customization.

ehsavoie:


ehsavoie:
supersedes https://github.com/wildfly/wildfly-core/pull/4731 after a wrong command line :)

xstefank (commented):
LGTM overall

controller/src/main/java/org/jboss/as/controller/persistence/yaml/YamlConfigurationExtension.java
```diff
@@ -0,0 +1,687 @@
...
+    public ConfigurationExtension load(Path... files) {
+        this.files = files;
+        load();
+        return this;
+    }
+
+    private void load() {
+        long start = System.currentTimeMillis();
+        for (Path file : files) {
+            if (file != null && Files.exists(file) && Files.isRegularFile(file)) {
```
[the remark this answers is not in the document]
xstefank: actually, I found out later that server won't boot when the yaml file is not found so maybe these checks are not necessary?

controller/src/main/java/org/jboss/as/controller/ModelControllerImpl.java
```diff
@@ -560,7 +564,19 @@ boolean boot(final List<ModelNode> bootList, final OperationMessageHandler handl
                 resultAction = validateContext.executeOperation();
             }
         }
-
+        //forcing the XML persistence after applying the YAML changes
```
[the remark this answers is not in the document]
ehsavoie: Yes, from the 'initial' idea of not being in read-only mode

controller/src/test/resources/org/jboss/as/controller/persistence/yaml/complete.yml
```diff
@@ -0,0 +1,185 @@
...
+        value: bar
+    # Put it before elytron to force re-ordering due to sasl factory being required.
+    core-service:
+      management:
+        management-interface:
+          http-interface:
+              console-enabled: true
+              http-upgrade:
+                enabled: true
+#                sasl-authentication-factory: example-sasl-auth
```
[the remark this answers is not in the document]
ehsavoie: This file is not used for testing. i t was the base file I used to create the extension. it should be removed

controller/src/test/resources/org/jboss/as/controller/persistence/yaml/simple_list.yml
```diff
@@ -0,0 +1,9 @@
+wildfly-configuration:
+    extension:
+      org.jboss.as.failure:
+        module: org.jboss.as.failure
+    list:
+      my-list: 
+        strings:
+            - foo
```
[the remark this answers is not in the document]
ehsavoie: 'list' is the resource with the name 'my-list'.
'strings' is the attribute of 'list' with the name 'strings'.
foo / bar are the elements of the list 'strings'

controller/src/main/java/org/jboss/as/controller/persistence/yaml/YamlConfigurationExtension.java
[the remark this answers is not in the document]
ehsavoie: correct

ehsavoie:
Full PR: https://github.com/wildfly/wildfly/pull/14614

controller/src/main/java/org/jboss/as/controller/ListAttributeDefinition.java
```diff
@@ -61,6 +63,10 @@ public ParameterValidator getElementValidator() {
         return elementValidator;
     }
 
+    public AttributeDefinition getValueType() {
```
jmesnil, line 66: Could you add some javadoc please?

Why would the valuetype of list throws an exception about YAML configuration?

controller/src/main/java/org/jboss/as/controller/PrimitiveListAttributeDefinition.java
```diff
@@ -51,8 +51,9 @@
     }
 
 
-    public ModelType getValueType() {
-        return valueType;
+    @Override
+    public AttributeDefinition getValueType() {
```
jmesnil, line 55: Why is the ValueType now returning an AttributeDefinition instead of a ModelType?

controller/src/main/java/org/jboss/as/controller/logging/ControllerLogger.java
```diff
@@ -3665,4 +3667,45 @@ OperationFailedRuntimeException capabilityAlreadyRegisteredInContext(String capa
...
     @Message(id = 484, value = "Attribute definition of attribute '%s' is null")
     IllegalArgumentException invalidAttributeDefinition(String attributeName);
+
+    @Message(id = 485, value = "Error parsing yaml file %s")
+    IllegalArgumentException failedToParseYamlConfigurationFile(String path, @Cause Throwable throwable);
+
+    @Message(id = 486, value = "Missing yaml file %s")
+    IllegalArgumentException missingYamlFile(String path);
+
+    @LogMessage(level = INFO)
```
jmesnil, line 3677: should be debug?

controller/src/main/java/org/jboss/as/controller/persistence/ConfigurationExtension.java
```diff
@@ -0,0 +1,42 @@
...
+import java.util.List;
+import org.jboss.as.controller.ParsedBootOp;
+import org.jboss.as.controller.RunningMode;
+import org.jboss.as.controller.registry.ImmutableManagementResourceRegistration;
+
+/**
+ * Interface for extensions of the XML boot configuration.
+ * @author Emmanuel Hugonnet (c) 2021 Red Hat, Inc.
+ */
+public interface ConfigurationExtension {
```
jmesnil, line 29: Could you please provide some javadoc?

controller/src/main/java/org/jboss/as/controller/persistence/ConfigurationExtensionFactory.java
```diff
@@ -0,0 +1,70 @@
...
+
+import java.nio.file.Path;
+import java.util.Iterator;
+import java.util.ServiceLoader;
+
+/**
+ *
+ * @author Emmanuel Hugonnet (c) 2021 Red Hat, Inc.
+ */
+public class ConfigurationExtensionFactory {
```
jmesnil, line 26: Could you please provide some javadoc? At least specify that only one ConfigurationExtension will be loaded by the ServiceLoader...

controller/src/main/java/org/jboss/as/controller/persistence/ConfigurationFile.java
```diff
@@ -150,7 +151,7 @@ private boolean isRemoveExisting() {
      *                                    to the configuration history directory
      */
     public ConfigurationFile(final File configurationDir, final String rawName, final String name, final boolean persistOriginal) {
-        this(configurationDir, rawName, name, persistOriginal ? InteractionPolicy.STANDARD : InteractionPolicy.READ_ONLY, false);
+        this(configurationDir, rawName, name, persistOriginal ? InteractionPolicy.STANDARD : InteractionPolicy.READ_ONLY, false, null, (ConfigurationExtension) null);
```
jmesnil, line 154: Why do you need to cast the last `null` parameter?

controller/src/main/java/org/jboss/as/controller/resource/InterfaceDefinition.java
```diff
@@ -342,6 +342,11 @@ protected void addOperationParameterValueTypeDescription(ModelNode node, String
             private void setValueType(ModelNode node) {
                 node.get(ModelDescriptionConstants.VALUE_TYPE).set(ModelType.STRING);
             }
+
+            @Override
+            public AttributeDefinition getValueType() {
```
jmesnil, line 347: This is fishy... the setter accepts a `ModeNode` while the getter returns an `AttributeDefinition`?

protocol/src/main/java/org/jboss/as/protocol/ProtocolConnectionUtils.java
```diff
@@ -192,7 +192,7 @@ public static Connection connectSync(final ProtocolConnectionConfiguration confi
             saslOptions = new HashMap<>(saslOptions);
             // Drop SASL_DISALLOWED_MECHANISMS which we already handled
             saslOptions.remove(Options.SASL_DISALLOWED_MECHANISMS.getName());
-            mergedConfiguration = mergedConfiguration.useMechanismProperties(saslOptions);
+            mergedConfiguration = mergedConfiguration.useSaslMechanismProperties(saslOptions);
```
jmesnil, line 195: what's the relation with the YAML support?

server/src/main/java/org/jboss/as/server/ServerEnvironment.java
```diff
@@ -332,47 +338,54 @@ public ServerEnvironment(final String hostControllerName, final Properties props
...
-
+        ConfigurationFile.InteractionPolicy configInteractionPolicy = configurationInteractionPolicy;
         this.startSuspended = startSuspended;
         this.startGracefully = startGracefully;
 
         this.launchType = launchType;
         this.standalone = launchType != LaunchType.DOMAIN;
 
         this.initialRunningMode = initialRunningMode == null ? RunningMode.NORMAL : initialRunningMode;
+        Path[] yamlFiles = findYamlFiles(yaml);
```
jmesnil, line 384: what is `yaml`? Is the the path to YAML files? or YAML content?

testsuite/manualmode/src/test/java/org/jboss/as/test/manualmode/auditlog/AbstractLogFieldsOfLogTestCase.java
```diff
@@ -101,15 +101,19 @@ void configureUser(final ModelControllerClient client, final CompositeOperationB
...
-            op.get("key").set(DEFAULT_USER_KEY);
             compositeOp.addStep(op.clone());
             op = Operations.createOperation("map-put", userAuthAddress);
             op.get("name").set("properties");
             op.get("key").set(DEFAULT_USER_KEY);
             op.get("value").set("$local");
             compositeOp.addStep(op.clone());
+            op = Operations.createOperation("map-put", userAuthAddress);
+            op.get("name").set("properties");
+            op.get("key").set("wildfly.sasl.local-user.challenge-path");
```
jmesnil, line 114: what's the relation with the YAML configuration?

testsuite/manualmode/src/test/java/org/jboss/as/test/manualmode/management/cli/ReloadOpsTestCase.java
```diff
@@ -94,6 +94,7 @@ public void testWriteAttribvuteWithReload() throws Exception {
         assertTrue(result.isIsOutcomeSuccess());
         value = (String) result.getResult();
         assertThat(value, is("false"));
+        cli.sendLine("/subsystem=logging:write-attribute(name=add-logging-api-dependencies, value=true)");
```
jmesnil, line 97: idem

jmesnil (changes requested):
Is there a WildFly PR that documents the feature?

The feature is being disabled by default but there should still be some documentation for it?

pom.xml
```diff
@@ -2099,6 +2099,11 @@
                     </exclusion>
                 </exclusions>
             </dependency>
+            <dependency>
+                <groupId>org.yaml</groupId>
+                <artifactId>snakeyaml</artifactId>
```
jmesnil, line 2104: the PR is missing the license stuff for this new dependency

jmesnil (changes requested):

server/src/main/java/org/jboss/as/server/ServerEnvironment.java
ehsavoie: path to the yaml file

protocol/src/main/java/org/jboss/as/protocol/ProtocolConnectionUtils.java
ehsavoie: I guess this is a rebase artefact

testsuite/manualmode/src/test/java/org/jboss/as/test/manualmode/management/cli/ReloadOpsTestCase.java
ehsavoie: to reset the configuration to what was here before the test

testsuite/manualmode/src/test/java/org/jboss/as/test/manualmode/auditlog/AbstractLogFieldsOfLogTestCase.java
ehsavoie: to reset the configuration

controller/src/main/java/org/jboss/as/controller/logging/ControllerLogger.java
ehsavoie: ok

controller/src/main/java/org/jboss/as/controller/PrimitiveListAttributeDefinition.java
ehsavoie: because we need more information than just a type which an AttributeDefinition provides

controller/src/main/java/org/jboss/as/controller/ListAttributeDefinition.java
ehsavoie: Ok

controller/src/main/java/org/jboss/as/controller/persistence/ConfigurationFile.java
ehsavoie: Not required

controller/src/main/java/org/jboss/as/controller/resource/InterfaceDefinition.java
ehsavoie: I agree but the setter is private :)

pom.xml
jmesnil: iiuc snakeyaml is already used by WildFly. Could you then create a WFLY issue to remove `snakeyaml` from WildFly when we will eventually upgrade a version of WildFly Core that contains this feature

ehsavoie:
Adding https://issues.redhat.com/browse/WFLY-15816 as a follow up for snakeyaml in WildFly

core-feature-pack/common/src/main/resources/modules/system/layers/base/org/yaml/snakeyaml/main/module.xml
```diff
@@ -0,0 +1,31 @@
...
+    <properties>
+        <property name="jboss.api" value="private"/>
+    </properties>
+
+    <resources>
+        <artifact name="${org.yaml:snakeyaml}"/>
+    </resources>
+
+    <dependencies>
+        <module name="java.logging"/>
```
jmesnil, line 28: should it be identical to https://github.com/wildfly/wildfly/blob/3eaeac0b07b4a03917f94bfb8368b6ee09c12530/ee-feature-pack/common/src/main/resources/modules/system/layers/base/org/yaml/snakeyaml/main/module.xml?

ehsavoie: ok, done

controller/src/main/java/org/jboss/as/controller/ListAttributeDefinition.java
jmesnil: my questions have not been answered afaict?

controller/src/main/java/org/jboss/as/controller/resource/InterfaceDefinition.java
jmesnil: The visibility of the gettter/setter do not matter. I'd expect to get consistent types for the getter and the setter.

controller/src/test/resources/org/jboss/as/controller/persistence/yaml/simple_list.yml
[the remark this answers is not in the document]
jmesnil: Martin is right, why can't it be:

```
list:
  my-list:
    strings:
      - foo
      - bar
```

controller/src/test/resources/org/jboss/as/controller/persistence/yaml/simple_list_add.yml
```diff
@@ -0,0 +1,10 @@
+wildfly-configuration:
+    extension:
+      org.jboss.as.failure:
+        module: org.jboss.as.failure
+    list:
+      my-list: 
+        strings: !list-add
```
jmesnil, line 7: have you defined somewhere all the YAML tags that are supported or do users need to know about https://github.com/wildfly/wildfly-core/pull/4766/files#diff-d50b9b2a6b6cbf891246ead8a48357b7fdd24ef0ba76f35bda6e992afb26c742R535-R537?

testsuite/manualmode/src/test/resources/org/jboss/as/test/manualmode/management/persistence/test.xml
```diff
@@ -0,0 +1,235 @@
+<?xml version="1.0" ?>
+
+<server xmlns="urn:jboss:domain:19.0">
```
jmesnil, line 3: Domain mode was not supported according to https://github.com/wildfly/wildfly-proposals/blob/be6bfece2ab58e46ea2f26c5f30e211a0f4c4e56/core/WFCORE-5343_Support_YAML_files_for_configuration_customization.adoc#non-requirements

Why are you testing the feature with a XML domain configuration?

controller/src/main/java/org/jboss/as/controller/ListAttributeDefinition.java
ehsavoie: It doesn't throw a YAML configuration exception anymore

jmesnil: Please keep in mind that this class is in the public API and can be used by any extension.

I'd not expect to get an unsupported exception from this base class. 
If only specialised subclass can support the YAML feature, let's move the corresponding API to their levels

controller/src/test/resources/org/jboss/as/controller/persistence/yaml/simple_list_add.yml
ehsavoie: In the doc only

controller/src/test/resources/org/jboss/as/controller/persistence/yaml/simple_list.yml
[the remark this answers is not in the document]
ehsavoie: Sorry I misunderstood. No, you don't need that 2nd indent.

controller/src/main/java/org/jboss/as/controller/ListAttributeDefinition.java
jmesnil: Our usual convention to put "experimental" stuff in our public API is to `@Deprecate` the method with a warning that this may be removed in future releases e.g. https://github.com/wildfly/wildfly-core/blob/0d189892dd2a503baed3a38989d2b2c21d0ab5c6/server/src/main/java/org/jboss/as/server/ExternalManagementRequestExecutor.java#L47-L51

ehsavoie: This class is abstract, the gaol was to provide a 'default' behavior

controller/src/main/java/org/jboss/as/controller/SimpleListAttributeDefinition.java
```diff
@@ -54,6 +54,7 @@ protected SimpleListAttributeDefinition(final ListAttributeDefinition.Builder bu
     }
 
 
+    @Override
     public AttributeDefinition getValueType() {
```
ehsavoie, line 58: @jmesnil this was already using the same semantic as I did

ehsavoie: same in ObjectListAttributeDefinition That's the reason why I kept the semantic

controller/src/main/java/org/jboss/as/controller/ListAttributeDefinition.java
jmesnil: sigh... so we have:

* ListAttributeDefinition that now defines `AttributeDefinition getValueType()`
* PrimitiveListAttributeDefinition.java  that used to define `ModelType getValueType()`
* SimpleListAttributeDefinition.java  that used to define `AttributeDefinition getValueType()`

is that correct?

There are issues with these commits:

* PrimitiveListAttributeDefinition#getValueType is a public method. You can't change its return type like that.
* The default behaviour for a list attribute definition is to throw an exception that there is no value type. This is fishy to me...

ehsavoie: for the1st point I agree (you could have added ObjectListAttributeDefinition to the list)
for the 2nd I disagree, throwing an UnsupportedOperationException when a method is not implemented is not something fishy. We have it at several places in our codebase for default methods in interfaces for example

jfdenise:
@ehsavoie , the new --yaml option needs to be recognized by Bootable JAR: https://github.com/wildfly/wildfly-core/blob/main/bootable-jar/runtime/src/main/java/org/wildfly/core/jar/runtime/Arguments.java#L65
In a bootable JAR context the files could be just file names and expected to be packaged in JBOSS_HOME/standalone/configuration directory. The absolute path would be construct when processing arguments.
The new argument would need to be added there: https://github.com/wildfly/wildfly-core/blob/main/bootable-jar/runtime/src/main/java/org/wildfly/core/jar/runtime/_private/BootableJarLogger.java
And finally the help must be updated: https://github.com/wildfly/wildfly-core/blob/main/bootable-jar/runtime/src/main/java/org/wildfly/core/jar/runtime/CmdUsage.java

ehsavoie:
@jfdenise I'll update the bootable jar code in a similar way as bare-metal and i'll see to be able to support relative path for thos configuration files

ehsavoie:
retest this please

controller/src/main/java/org/jboss/as/controller/ObjectListAttributeDefinition.java
```diff
@@ -174,6 +174,10 @@ protected void addAllowedValuesToDescription(ModelNode result, ParameterValidato
     }
 
     public final ObjectTypeAttributeDefinition getValueType() {
+        return getValueAttributeDefinition();
+    }
+
+    public final ObjectTypeAttributeDefinition getValueAttributeDefinition() {
```
jmesnil, line 180: is it correct? Shouldn't this `@Override` the method in ListAttributeDefinition.java

controller/src/main/java/org/jboss/as/controller/PrimitiveListAttributeDefinition.java
```diff
@@ -168,7 +172,7 @@ public Builder(final String name, final ModelType valueType) {
 
         public Builder(final PrimitiveListAttributeDefinition basis) {
             super(basis);
-            this.valueType = basis.getValueType();
+            this.valueType = basis.getValueAttributeDefinition().getType();
```
jmesnil, line 175: is that change requrired? The `getValueType()` method is still there.

controller/src/main/java/org/jboss/as/controller/ListAttributeDefinition.java
```diff
@@ -61,6 +63,14 @@ public ParameterValidator getElementValidator() {
         return elementValidator;
     }
 
+    /**
+     * Returns an AttributeDefinition describing the content of the list.
+     * @return an AttributeDefinition describing the content of the list.
+     */
+    public AttributeDefinition getValueAttributeDefinition() {
+        throw ROOT_LOGGER.missingListAttributeValueType(getName());
```
jmesnil, line 71: The Javadoc must mention when this exception can be raised.

When did you catch it actually? How can the code using a ListAttributeDefinition assert if calling this method is valid or not?

I'm concerned by having this method raises an exception by default. If the fundamental definition of a list can not provide an attributedefinition for its item, it does not belong to this class.

It seems to me that your change would need an intermediate class between `ListAttributeDefinition` and subclasses that *can* provide such an attribute definition (which is the case for all the subclasses defined in wildfly-core, I've not checked in wildfly full).

Having an intermediate `AttributeDefinition-based ListAttributeDefintion` would also simplify a lot the various subclasses we have (Object, Primitive, String, Simple)

controller/src/main/java/org/jboss/as/controller/persistence/yaml/YamlConfigurationExtension.java
```diff
@@ -0,0 +1,713 @@
...
+                        break;
+                }
+            }
+        }
+        return objectNode;
+    }
+
+    @SuppressWarnings("unchecked")
+    private void processListAttribute(ListAttributeDefinition att, ModelNode list, Object value) {
+        AttributeDefinition type = att.getValueAttributeDefinition();
```
jmesnil, line 382: what if this method raises an exception?

jmesnil (changes requested):
@bstansberry Do you have issues with the proposed change in the `ListAttributeDefinition`. class (adding a default method that throws an exception)?

Having a method defaulting to throw an exception means that YAML will not be able to cope with *any* kind of list.
Maybe, it would be cleaner to assert that a list is valid (an attribute definition can be inferred from the list's attribute definition) when handling the YAML.

If that's not the case, the YAML can raise an exception (or ignore the list with a warning).

A valid list would be an instance of any of our known subclass (ObjectListAttributeDefinition, etc.) with the attribute definition inferred from their specific methods (either `ModelType getValueType()` or `ObjectTypeAttributeDefinition getValueType()` or `AttributeDefinition getValueType()`)

Or as mentioned in one of my comments, add a intermediate ListAttributeDefinition subclass that unify all other subclasses based on valueType/attributeDefinition

controller/src/main/java/org/jboss/as/controller/persistence/yaml/YamlConfigurationExtension.java
ehsavoie: Then the configuration would fail and the server wouldn't boot until you fix your YAML configuration or the attribute definition. Basically that means that this attribute is not configurable from YAML.

controller/src/main/java/org/jboss/as/controller/ListAttributeDefinition.java
ehsavoie: The exception is not caught. Trying to configure a list that is not properly defined using YAML will fail.

Note that this only concerns ListAttributeDefinitions that are configurable. So you mean that we have a list which values aren't definable by AttributeDefinition but that we want to be able to configure using YAML ? In such a case it will fail, exception or not, so I think that throwing a standard exception is clearer.
Just FYI we have several such ListAttributeDefinitions in full for runtime attributes.

controller/src/main/java/org/jboss/as/controller/PrimitiveListAttributeDefinition.java
ehsavoie: Yes :) I don't remember why but otherwise it breaks

controller/src/main/java/org/jboss/as/controller/ListAttributeDefinition.java
jmesnil: What make a `list` configurable with YAML?
The presence of a valueType (directly for primitive or though an attributedefinition for object list)?
Is is valid to have a list without a valueType?

controller/src/main/java/org/jboss/as/controller/persistence/yaml/YamlConfigurationExtension.java
jmesnil: I'm splitting hairs here but the only thing you need is to check if the `att` is an instance of `ObjectListAttributeDefinition` or `PrimitiveListAttributeDefinition` or `SimpleListAttributeDefinition` and process accordingly.

jmesnil: or even just check if it is an instance of ObjectListAttributeDefinition and handles all other types of list the same way (with a String item)

controller/src/main/java/org/jboss/as/controller/ListAttributeDefinition.java
bstansberry: If it is not valid to have a list w/o a valid "value definition", I think the thing to do is have this method return null, and then YamlConfigurationExtension processListAttribute can fail if null is returned. IOW move the exception throwing to the code that can't deal with a situation, and don't have it in the ListAD code which doesn't care whether there is a value definition.

From the calling code POV that allows callers who don't know what ListAD subclass they are dealing with to do flow control using a null check instead of a try/catch. If the only place that would ever call this was YamlConfigurationExtension that distinction wouldn't matter much (as it doesn't want to catch) but methods in core API classes like ListAD should be robust when other uses we haven't thought of appear.

controller/src/main/java/org/jboss/as/controller/persistence/yaml/YamlConfigurationExtension.java
ehsavoie: I disagree (I have long hair) as if there is some ListAttributeDefinition in the outer space that wants to manipulate an ObjectAttributeDefinition without extending ObjectListAttributeDefinition this switch over the implementation would 'break'.

controller/src/main/java/org/jboss/as/controller/ListAttributeDefinition.java
ehsavoie: I hate returning 'null' tbh as this leads to NPE (even if i agree that a I could use a checked exception in that case.

controller/src/main/java/org/jboss/as/controller/persistence/yaml/YamlConfigurationExtension.java
rpelisse: @jmesnil for what is worth, it seems better to me to fail "loudly" and not boot, so the config is fixed (and the lack of support for the requested features explicit).

ehsavoie:
@jmesnil are all your concerns addressed ?

controller/src/main/java/org/jboss/as/controller/ListAttributeDefinition.java
```diff
@@ -61,6 +62,15 @@ public ParameterValidator getElementValidator() {
         return elementValidator;
     }
 
+    /**
+     * Returns an AttributeDefinition describing the content of the list.
+     * @return an AttributeDefinition describing the content of the list.
+     * @throws an UnsupportedOperationException if the expected AttributeDefinition of list values is not defined.
```
jmesnil, line 68: I don't think that javadoc is accurate anymore. Instead it should mention that it returns `null` if the list does not define an attribute definition for its item.

jmesnil (changes requested):
@ehsavoie one last fix to the Javadoc and we're good to go :)

jmesnil (approved):
@ehsavoie thanks for the changes. I'll merge this PR after the CI jobs are finished

jmesnil:
@ehsavoie thanks 🍾 !

PR #4771, opened 2021-09-25
[WFCORE-5587] maximum-cert-path does not work with empty certificate-revocation-list in Elytron trust-manager

rmartinc:
Issue: https://issues.redhat.com/browse/WFCORE-5587

A few problems in the `SSLDefinitions` class:

1. The `certificate-revocation-list` can be defined empty and although it exists the current call to `isDefined` returns `false`. This is because no other element is defined inside it (no default value or whatever) and it's resolved to undefined. It is important to detect that this node is set (even empty) because distribution points (DP) should be checked in that case. The call to `isDefined` is changed to `hasDefined` in the parent node that detects its presence correctly. This is changed in the places that DP should be considered.
2. The `CrlFile` is added to the list when the `path` is null, that throws a NPE. For example when setting the `certificate-revocation-list` only with the `maximum-cert-path` set.
3. The `only-leaf-cert` property was never set in the builder.

Tests modified to detect 1 and 2.

@fjuma Take a look to this when you have time.


jmesnil:
@fjuma could you please review?

fjuma (approved):
Thanks @rmartinc!

jmesnil:
@rmartinc thanks

PR #4772, opened 2021-09-25
[WFCORE-5594] Restriction of XML External Entity Reference (XXE)

boris-unckel:
Fixes https://issues.redhat.com/browse/WFCORE-5594

### DIFF
--- a/bootable-jar/runtime/src/main/java/org/wildfly/core/jar/runtime/BootableJar.java
+++ b/bootable-jar/runtime/src/main/java/org/wildfly/core/jar/runtime/BootableJar.java
@@ -39,7 +39,6 @@ import java.util.stream.Stream;
 import javax.xml.parsers.DocumentBuilder;
 import javax.xml.parsers.DocumentBuilderFactory;
 import javax.xml.transform.Transformer;
-import javax.xml.transform.TransformerFactory;
 import javax.xml.transform.dom.DOMSource;
 import javax.xml.transform.stream.StreamResult;
 import org.jboss.as.controller.client.ModelControllerClient;
@@ -75,6 +74,8 @@ import org.w3c.dom.Document;
 import org.w3c.dom.Element;
 import org.w3c.dom.Node;
 import org.w3c.dom.NodeList;
+import org.wildfly.common.xml.DocumentBuilderFactoryUtil;
+import org.wildfly.common.xml.TransformerFactoryUtil;
 import static org.wildfly.core.jar.runtime.Constants.DEPLOYMENTS;
 import static org.wildfly.core.jar.runtime.Constants.LOGGING_PROPERTIES;
 import static org.wildfly.core.jar.runtime.Constants.SERVER_LOG;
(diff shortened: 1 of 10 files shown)


boris-unckel:
@jmesnil I have seen that [WFCORE-5595][WFCORE-5596] are merged and are part of 18.0.0.Beta2. Could you specify a module where this code should be placed? Or do you want me to rebase this after WFLY-15551 has been merged, to inspect the impact?

boris-unckel:
@bstansberry @jmesnil The PR is now ready for review, tests passed, errors seem unrelated. Could you check and approve, please?

dmlloyd:
I think these could make sense in `wildfly-common`, under an `xml` sub-package. :+1:

controller-client/src/main/java/org/jboss/as/controller/client/logging/ControllerClientLogger.java
```diff
@@ -372,6 +374,10 @@
     @Message(id = 37, value = "Failed to parse the configuration file: %s")
     RuntimeException failedToParseAuthenticationConfig(@Cause Throwable cause, URI location);
 
+    @LogMessage(level = WARN)
+    @Message(id = 38, value = "Property or feature %s not supported by %s")
+    void xmlFactoryPropertyNotSupported(@Cause Throwable cause, String property, String factoryName);
```
bstansberry, line 379: Is it expected these would never be logged? Otherwise recurring WARNs would be problematic.

ropalka: I apologize but I somehow unnoticed the ongoing Zulip discussion about this PR @boris-unckel .
I agree with both of you @bstansberry and @dmlloyd this kind of change should be in a more generic
module / library like e.g. wildfly-common, because this PR introduces some "optimized XML processing" settings
and these settings might be reused by other modules / libraries / WS subsystems.

ropalka:
I apologize but I somehow unnoticed the ongoing Zulip discussion about this PR @boris-unckel .
I agree with both of you @bstansberry and @dmlloyd this kind of change should be in a more generic
module / library like e.g. wildfly-common because it introduces some "optimized XML processing" settings
and these settings might be reused by other modules / libraries / WS subsystems.

controller-client/src/main/java/org/jboss/as/controller/client/logging/ControllerClientLogger.java
bstansberry: @boris-unckel Thanks. In the WildFly context, is the expectation that these would never be logged? @ropalka you may know the answer too.

My assumption is we don't expect these to be logged; i.e. the factories we'd wrap will support all of these features. But I don't want to assume. :)

bstansberry (approved):

bstansberry:
Thanks, @boris-unckel

boris-unckel:
@bstansberry I assume it happened accidentally that 18.x has been merged, but this one is still open. If you follow a different procedure please ignore my comment.

yersan:
Thank you for the contribution, @boris-unckel !

PR #4774, opened 2021-09-27
[WFCORE-5597] Remove references to the security-integration module as it is being removed from WildFly.

darranl:
https://issues.redhat.com/browse/WFCORE-5597


darranl:
This one should not depend on changes in WildFly, the test failure is unrelated.

jmesnil:
@darranl thanks

PR #4779, opened 2021-10-01
Roll back Jandex to 2.3.1

bstansberry:
I have a theory the Jandex 2.4.0 upgrade is causing CDI TCK issues so this is an experiment in rolling it back. But not all the way to 2.2.3; just to 2.3.1.

I expect failures in the full integration tests as other libs have been upgraded expecting 2.4.0 to be present. https://github.com/wildfly/wildfly/pull/14743 is an experiment in full with rolling jandex back there. I filed this here mostly to get the core CI jobs to run.


jmesnil:
superset by #4780

PR #4783, opened 2021-10-05
[WFCORE-5602] security enable-http-auth-http-server command fails for the security domain "other"

rmartinc:
Issue: https://issues.redhat.com/browse/WFCORE-5602

When executing a security command over an existing undertow app domain it can happen that it was defined using referenced domain and the command is trying to add an http factory (or vice-versa). In one case the command returns the weird error detected in the JIRA and in the other the domain is not modified at all. This PR just detects that situation and throws an error: `Can't mix mechanism and referenced security domain` (the same error that was already used when the command passed both parameters).

The tests seem to be in wildfly `SecurityAuthCommandsTestCase.java` class. But we need to wait for that test to be re-enabled in WFLY-15057 (PR https://github.com/wildfly/wildfly/pull/14670).

@jfdenise Please review this when you have some time.

Thanks!

### DIFF
--- a/cli/src/main/java/org/jboss/as/cli/impl/aesh/cmd/security/auth/HTTPServerEnableAuthCommand.java
+++ b/cli/src/main/java/org/jboss/as/cli/impl/aesh/cmd/security/auth/HTTPServerEnableAuthCommand.java
@@ -167,11 +167,10 @@ public class HTTPServerEnableAuthCommand extends AbstractEnableAuthenticationCom
     protected void secure(CommandContext ctx, AuthSecurityBuilder builder) throws Exception {
         ApplicationSecurityDomain secDomain = HTTPServer.getSecurityDomain(ctx, securityDomain);
         if (secDomain != null) {
-            if (secDomain.getSecurityDomain() != null) {
-                if (!secDomain.getSecurityDomain().equals(builder.getReferencedSecurityDomain())) {
-                    // re-write the existing security domain
-                    HTTPServer.writeReferencedSecurityDomain(builder, securityDomain, ctx);
-                }
+            if (secDomain.getSecurityDomain() != null && builder.getReferencedSecurityDomain() != null
+                    && !secDomain.getSecurityDomain().equals(builder.getReferencedSecurityDomain())) {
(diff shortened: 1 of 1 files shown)


jfdenise:
@rmartinc , thank-you. The change looks good to me. Did you ran the currently ignored tests with this change? It would be nice to add a testcase to cover it.

rmartinc:
Thanks @jfdenise!

> Did you ran the currently ignored tests with this change? It would be nice to add a testcase to cover it.

Yes, I have worked on top of your PR https://github.com/wildfly/wildfly/pull/14670, tests pass OK and I have tentatively added something like this:

```java
@@ -371,6 +373,16 @@ public class SecurityAuthCommandsTestCase {
     public void testOOBHTTP() throws Exception {
         ctx.handle("security enable-http-auth-http-server --no-reload --security-domain=" + TEST_UNDERTOW_DOMAIN);
         Assert.assertEquals(ElytronUtil.OOTB_APPLICATION_DOMAIN, getReferencedSecurityDomain(ctx, TEST_UNDERTOW_DOMAIN));
+
+        try {
+           // enabling the same domain using mechanism should fail now
+            ctx.handle("security enable-http-auth-http-server --no-reload --mechanism=BASIC "
+                    + "--file-system-realm-name=" + TEST_FS_REALM + " --security-domain=" + TEST_UNDERTOW_DOMAIN);
+            Assert.fail("Enabling using mechanism should have failed");
+            MatcherAssert.assertThat(e.getMessage(), CoreMatchers.containsString("Can't mix mechanism and referenced security domain"));
+        }
+
         ctx.handle("security disable-http-auth-http-server --no-reload --security-domain=" + TEST_UNDERTOW_DOMAIN);
         Assert.assertFalse(domainExists(TEST_UNDERTOW_DOMAIN));
     }
```

The same in the `testOOBHTTP2` to test the other path. I have create [WFLY-15410](https://issues.redhat.com/browse/WFLY-15410) to modify the tests once the activation PR is merged.

jfdenise:
@rmartinc , perfect, thank-you!

jmesnil:
@rmartinc thanks!

PR #4784, opened 2021-10-05
[WFCORE-5609] Fix AbstractOperationContext#getCurrentOperationName()

jmesnil:
The name of the operation is held by the `OP` constant (`"operation"`)
and not by the `NAME` constant (that would contain an eventual `"name"`
parameter of the given operation).

JIRA: https://issues.redhat.com/browse/WFCORE-5609


jmesnil:
@bstansberry I've reviewed WildFly Full `main` branch and there is no usage of this method. I think we can fix its implementation without any impact on WildFly

ehsavoie:
:+1 Let's fix this before starting working on WFLY-15347 Subsystem code standardization

bstansberry (approved):

bstansberry:
Thanks @jmesnil

PR #4786, opened 2021-10-05
Wfcore xxxx parsing cleanup

jmesnil:
Tentative work to cleaning Management parsing to simplify release and maintenance of WildFly Core


darranl:
@jmesnil What is the clean up?  Removing the clones that didn't result in their own changes?

jmesnil:
@darranl there is 2 cleanup:

* removing the clones as you mentioned
* https://github.com/wildfly/wildfly-core/pull/4786/commits/78e9116583e618b3e6640596eb7cb3a4e116ce5d to avoid having to bump the XML schema every time we release WildFly Core. The enum does not bring anything (as we don't validate the XML files when the server are started and we already have code that generate KnownRelease from a string)

It work in progress and I need to create proper WFCORE to get this done eventually

server/src/main/resources/schema/wildfly-config_19_0.xsd
```diff
@@ -3815,231 +3815,18 @@
...
+                            <![CDATA[
+                              The short hand identifier can have to form:
+                              * EAP<major>.<minor> (for example, EAP6.2 or EAP7.3)
+                              * WildFly<major>.<minor> (for example WildFly10.0 or WildFly10.1)
+
+                              The corresponding kernel management API version is defined in the enum
+                              org.jboss.as.domain.controller.resources.HostExcludeResourceDefinition.KnownRelease.
+                            ]]>
+                        </xs:documentation>
+                    </xs:annotation>
```
yersan, line 3829: This change looks good, this is rejected by the attribute validator and not by the XSD schema.

PR #4787, opened 2021-10-06
[WFCORE-5611] Refactor Array copies (Controller)

boris-unckel:
Fixes https://issues.redhat.com/browse/WFCORE-5611
as part of https://issues.redhat.com/browse/WFCORE-5576

### DIFF
--- a/controller/src/main/java/org/jboss/as/controller/AbstractAddStepHandler.java
+++ b/controller/src/main/java/org/jboss/as/controller/AbstractAddStepHandler.java
@@ -452,9 +452,9 @@ public class AbstractAddStepHandler implements OperationStepHandler, OperationDe
          */
         public OrderedResourceCreator(boolean indexedAdd, String... orderedChildTypes) {
             this.indexedAdd = indexedAdd;
-            Set<String> set = new HashSet<String>(orderedChildTypes.length);
-            for (String type : orderedChildTypes) {
-                set.add(type);
+            Set<String> set = Collections.emptySet();
+            if (orderedChildTypes != null && orderedChildTypes.length > 0) {
+                set = new HashSet<String>(Arrays.asList(orderedChildTypes));
             }
             this.orderedChildTypes = set;
         }
--- a/controller/src/main/java/org/jboss/as/controller/operations/validation/TimeUnitValidator.java
+++ b/controller/src/main/java/org/jboss/as/controller/operations/validation/TimeUnitValidator.java
@@ -23,6 +23,7 @@
 package org.jboss.as.controller.operations.validation;
 
 import java.util.ArrayList;
+import java.util.Collections;
 import java.util.EnumSet;
 import java.util.List;
 import java.util.Locale;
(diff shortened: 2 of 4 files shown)


controller/src/main/java/org/jboss/as/controller/AbstractAddStepHandler.java
```diff
@@ -452,9 +452,9 @@ public OrderedResourceCreator(boolean indexedAdd, Set<String> orderedChildTypes)
          */
         public OrderedResourceCreator(boolean indexedAdd, String... orderedChildTypes) {
             this.indexedAdd = indexedAdd;
-            Set<String> set = new HashSet<String>(orderedChildTypes.length);
-            for (String type : orderedChildTypes) {
-                set.add(type);
+            Set<String> set = Collections.emptySet();
```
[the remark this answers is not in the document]
boris-unckel: @yersan Good point. I'll fix it.

controller/src/main/java/org/jboss/as/controller/operations/validation/TimeUnitValidator.java
```diff
@@ -59,8 +60,8 @@ public TimeUnitValidator(final boolean nullable, final boolean allowExpressions)
     public TimeUnitValidator(final boolean nullable, final boolean allowExpressions, final TimeUnit... allowed) {
         super(ModelType.STRING, nullable, allowExpressions);
         allowedValues = EnumSet.noneOf(TimeUnit.class);
-        for (TimeUnit tu : allowed) {
-            allowedValues.add(tu);
+        if (allowed != null && allowed.length > 0) {
+            Collections.addAll(allowedValues, allowed);
```
yersan, line 64: Also, @boris-unckel just out of curiosity, what motivates you to replace the simple for loop by the Collections method?

It is inappreciable, so just to see your point of view, but these Collections.addAll() takes care of an internal flag to return whether all elements were added to the list. We do not need to take care of this flag here, so the previous for loop code is slightly faster than the one implemented on the Collections.addAll, so really I'm not sure how to justify these changes.

yersan: Hello @boris-unckel , if we finally conclude this is just a matter of code beautification, there are reasons to not include it, as you know, we backport changes to several other branches, and each commit slightly adds more difficulties to track the source of a change. These are quite weak arguments, but if we do not have anything against them I think would be better to not include them.

I've tried to also justify the changes based on the null checks, but even in that case, they sound a bit questionable to me too, for example, specifically to this class, throwing an NPE would make the developer use the correct available method, the one that does not use `allowed` as an argument, well, also a bit weak argument, but an argument, would you agree with it?

boris-unckel: Hi @yersan thanks for taking the time to weighten the pros/cons for this PR. I think the bottom line for this JDK util methods is that they are useful for new code, it increases reuse and readability. For existing code it's more code beautifying. I'm going to  withdraw the series of PRs for this reason.

boris-unckel:
Close as "Won't do", see [PR5611](https://github.com/wildfly/wildfly-core/pull/4787#discussion_r729724022)

PR #4792, opened 2021-10-07
[WFCORE-5615] Refactor resource closing (controller)

boris-unckel:
Fixes https://issues.redhat.com/browse/WFCORE-5615
as part of https://issues.redhat.com/browse/WFCORE-5606


boris-unckel:
The proposed change series contains XNIO as dependency not wanted for this utility purpose and the overall value/risk ratio is too low. See [PR14791](https://github.com/wildfly/wildfly/pull/14791#issuecomment-942503121) It's therefore withdrawn.

PR #4814, opened 2021-10-18
[WFCORE-5640][WFCORE-5641] XSD and XML cleanup

jmesnil:
This PR includes:

* [WFCORE-5640] Relax constraint on host-releaseType's id attribute
  Instead of making it an enum that must be updated with every WildFly
  Core release, let's use a string type.

  JIRA: https://issues.redhat.com/browse/WFCORE-5640
* [WFCORE-5641] Clean up DomainXml parser classes
  Remove all WildFly core Xml classes that were identical to their previous
  versions.
  This only leaves classes that introduced actual changes to the
  XML parsing
  
  1 commit per XML class

  JIRA: https://issues.redhat.com/browse/WFCORE-5641


jmesnil:
@yersan thanks for the review

PR #4816, opened 2021-10-18
[WFCORE-5489] As a developer, I want to override management attribute…

jmesnil:
… values using environment variables

When validating and setting the value for an attribute, check if there is an env var
that would override its value before doing any expression resolution or
correction.

This feature is enabled by the existence of the
`WILDFLY_OVERRIDING_ENV_VARS` environment variable.

JIRA: https://issues.redhat.com/browse/WFCORE-5489

Dev analysis is at https://github.com/wildfly/wildfly-proposals/pull/404

### DIFF
--- a/controller/pom.xml
+++ b/controller/pom.xml
@@ -168,4 +168,53 @@
         </dependency>
     </dependencies>
 
+    <build>
+        <plugins>
+            <plugin>
+                <groupId>org.apache.maven.plugins</groupId>
+                <artifactId>maven-surefire-plugin</artifactId>
+                <executions>
+                    <execution>
+                        <id>default-test</id>
+                        <goals>
+                            <goal>test</goal>
+                        </goals>
+                        <configuration>
+                            <environmentVariables>
+                                <!-- Used in EnvVarAttributeOverrideModelTestCase -->
+                                <!-- This corresponds to the /subsystem=custom resource and its my-attr attribute -->
+                                <SUBSYSTEM_CUSTOM__MY_ATTR>5678</SUBSYSTEM_CUSTOM__MY_ATTR>
+                            </environmentVariables>
+                            <excludes>
+                                <exclude>org/jboss/as/controller/EnvVarAttributeOverrideModel$EnabledOverridingEnvVarTestCase</exclude>
+                            </excludes>
+                        </configuration>
+                    </execution>
+                    <execution>
+                        <id>env-var-overrides</id>
+                        <goals>
+                            <goal>test</goal>
+                        </goals>
+                        <configuration>
+                            <environmentVariables>
+                                <!-- Used in EnvVarAttributeOverrideModelTestCase -->
+                                <!-- The presence of this env var is required to trigger the overridding feature -->
+                                <WILDFLY_OVERRIDING_ENV_VARS>1</WILDFLY_OVERRIDING_ENV_VARS>
+                                <!-- This corresponds to the /subsystem=custom resource and its my-attr attribute -->
+                                <SUBSYSTEM_CUSTOM__MY_ATTR>5678</SUBSYSTEM_CUSTOM__MY_ATTR>
+                                <!-- test that a complex attribute is not overidden -->
+                                <SUBSYSTEM_CUSTOM__MY_LIST_ATTR>[a, b, c]</SUBSYSTEM_CUSTOM__MY_LIST_ATTR>
+                            </environmentVariables>
+                            <includes>
+                                <include>org/jboss/as/controller/EnvVarAttributeOverrideModel$EnabledOverridingEnvVarTestCase</include>
+                            </includes>
+                            <excludes>
+                                <exclude></exclude>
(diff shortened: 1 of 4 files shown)


controller/src/test/java/org/jboss/as/controller/EnvVarAttributeOverrideModelTestCase.java
```diff
@@ -0,0 +1,150 @@
...
+        addOp.get(MY_ATTR.getName()).set("value1");
+        executeCheckNoFailure(addOp);
+
+        ModelNode readResource = createOperation(READ_RESOURCE_OPERATION, CUSTOM_RESOURCE_ADDR);
+        executeCheckNoFailure(readResource);
+
+        ModelNode readAttribute = createOperation(READ_ATTRIBUTE_OPERATION, CUSTOM_RESOURCE_ADDR);
+        readAttribute.get(NAME).set(MY_ATTR.getName());
+        ModelNode response = executeCheckNoFailure(readAttribute);
+        if (isFeatureEnabled()) {
```
vsalbaba, line 99: Is this test somehow running automatically for both `true` and `false` branches, or is this designed so that you have to manually change/comment out `controller/pom.xml#186` and then run the testsuite again?

jmesnil:
Community Documentation is in PR https://github.com/wildfly/wildfly/pull/14418

controller/src/test/java/org/jboss/as/controller/EnvVarAttributeOverrideModelTestCase.java
vsalbaba: I believe that even if the enabling functionality was broken and the feature was always on, this would still be green with the feature "disabled", as there is nothing trying to rewrite "value1" to "value2" in `default-test` execution.
Would it be possible to set 
```
                            <environmentVariables>
                                <SUBSYSTEM_CUSTOM__MY_ATTR>value2</SUBSYSTEM_CUSTOM__MY_ATTR>
                            </environmentVariables>
```
in execution `default-test`? Than we would be sure that the on/off switch is working as intended.

jmesnil: @vsalbaba good catch, I've updated the PR based on your suggestion

vsalbaba:
LGTM

controller/pom.xml
```diff
@@ -168,4 +168,45 @@
...
+                        <configuration>
+                            <environmentVariables>
+                                <!-- Used in EnvVarAttributeOverrideModelTestCase -->
+                                <!-- The presence of this env var is required to trigger the overridding feature -->
+                                <WILDFLY_OVERRIDING_ENV_VARS>1</WILDFLY_OVERRIDING_ENV_VARS>
+                                <!-- This corresponds to the /subsystem=custom resource and its my-attr attribute -->
+                                <SUBSYSTEM_CUSTOM__MY_ATTR>value2</SUBSYSTEM_CUSTOM__MY_ATTR>
+                            </environmentVariables>
+                            <includes>
+                                <include>org/jboss/as/controller/EnvVarAttributeOverrideModelTestCase.java</include>
```
bstansberry, line 204: Running the same test in two different executions doesn't work well in CI as failures in the 2nd execution don't get reported by TeamCity. Perhaps duplicate the test and exclude the one for this execution from the default execution?

I wonder if changing the name of EnvVarAttributeOverrideModelTestCase to match the pattern that gets tested would work, and then add 2 trivial public static inner classes. So it all stays in 1 .java file.

controller/src/main/java/org/jboss/as/controller/AttributeDefinition.java
```diff
@@ -501,10 +501,23 @@ public ModelNode validateOperation(final ModelNode operationObject) throws Opera
      * @throws OperationFailedException if the value is not valid
      */
     public final void validateAndSet(ModelNode operationObject, final ModelNode model) throws OperationFailedException {
+        final boolean isAttributeBoundToAResource = operationObject.has(ModelDescriptionConstants.OP_ADDR);
+        PathAddress pathAddress = PathAddress.pathAddress(operationObject.get(ModelDescriptionConstants.OP_ADDR));
```
bstansberry, line 505: Initializing this to null and then creating the first time it's needed seems better, given how often this is called, particularly durning testsuite reloads.

controller/pom.xml
jmesnil: updated

controller/src/main/java/org/jboss/as/controller/AttributeDefinition.java
jmesnil: I've reverted the code so that this is only instantiated if the feature is enabled.
The 2 other instances occurred outside of the main code path

bstansberry (approved):
Thanks, @jmesnil!

vsalbaba:
Looking good to me

PR #4846, opened 2021-10-29
[WFCORE-5675] Avoid NPE sending transformed operation results in Oper…

yersan:
…ationTransformationTestCase and fixing it for JDK 17

Jira issue: https://issues.redhat.com/browse/WFCORE-5675

### DIFF
--- a/controller/src/main/java/org/jboss/as/controller/remote/BlockingQueueOperationListener.java
+++ b/controller/src/main/java/org/jboss/as/controller/remote/BlockingQueueOperationListener.java
@@ -199,4 +199,71 @@ public class BlockingQueueOperationListener<T extends TransactionalProtocolClien
 
     }
 
+    public static class SucceededOperation<T extends TransactionalProtocolClient.Operation>
+            implements TransactionalProtocolClient.PreparedOperation<T> {
+
+        private final T operation;
+        private final ModelNode finalResult;
+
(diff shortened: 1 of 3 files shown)


bstansberry:
Hi @yersan 

Can you add a bit more info in a JIRA comment about how this fails? Like is it on the DC side and it just gets reported as being for 'slave' because it's in some code for handling the interaction with 'slave'? Or is it on the remote HC side.  And then why does JEPS 358 matter? At a glance it seems like that is about changing the NPE message, so I'm not clear why it would change how the NPE is handled.

And then before https://github.com/wildfly/wildfly-core/pull/4846/files#diff-9331f67c4ad3d56f4a5b46209d12c5311dba94eb06455a42f8cb43bb4c3473eeR115 a good code comment would be good, particularly to explain why discarding tranformed == null is the right thing to do.

I expect this is fine, but I don't really understand why (which is ok TBH). But in 5 years some maintainer who needs to deal with those code for some reason will be happy for more info.

yersan:
Hello @bstansberry , I've added some additional information on the code and Jira, let me know if that is enough to understand better the issue and solution, thanks

bstansberry (approved):
Thanks, @yersan -- that's very clear.

bstansberry:
Thanks, @yersan!

PR #4850, opened 2021-11-04
[WFCORE-5685] KerberosHttpMgmtSaslTestCase & KerberosNativeMgmtSaslTestCase tests fail on JDK 18 EA

rmartinc:
Issue: https://issues.redhat.com/browse/WFCORE-5685

The ldap/kerberos tests now also add all the enc-types defined in the keytab to the `permitted_enctypes` option in the `krb5.conf` file. This way the allowed enc-types don't depend on the default JDK values. Besides, previously the enc-types selected by the tests were different for ibm and non-ibm JDKs, this part has been also changed (don't know why but with non-ibm only two types were selected, this was done since the creation of the file).

Both tests pass for me with for jdk-18 and the current JDKs (I tried with a lot of versions and vendors).


rmartinc:
@fjuma Remember this PR is waiting. It's just a minor change in the TS that adds all available mechanism to the kerberos setup. Thanks!

fjuma (approved):
@rmartinc Thanks very much for the reminder and for working on this!

PR #4851, opened 2021-11-04
[WFCORE-1178] Use specific description for reply value types in ListAttributeDefintion and MapAttributeDefinition

petrberan:
https://issues.redhat.com/browse/WFCORE-1178


petrberan:
Changes in WFLY for PR checks to pass are in https://github.com/wildfly/wildfly/pull/14994

PR #4852, opened 2021-11-04
WFCORE-5676: remove incorrect description resolution

michpetrov:
Issue: [WFCORE-5676](https://issues.redhat.com/browse/WFCORE-5676)

This code doesn't seem to serve a purpose, the relevant data are always defined on the ObjectList proper while the valueType is a bare-bones definition so we shouldn't be trying to extract data from it.


bstansberry:
Thanks, @michpetrov

PR #4853, opened 2021-11-05
WFCORE-4041: supply default values for nested attributes

michpetrov:
Issue: [WFCORE-4041](https://issues.redhat.com/browse/WFCORE-4041)

Resubmitting because I cannot reopen the original PR.

### DIFF
--- a/elytron/src/main/java/org/wildfly/extension/elytron/LdapRealmDefinition.java
+++ b/elytron/src/main/java/org/wildfly/extension/elytron/LdapRealmDefinition.java
@@ -101,7 +101,6 @@ class LdapRealmDefinition extends SimpleResourceDefinition {
 
         static final SimpleAttributeDefinition RECURSIVE_SEARCH = new SimpleAttributeDefinitionBuilder(ElytronDescriptionConstants.SEARCH_RECURSIVE, ModelType.BOOLEAN, true)
                 .setRequires(ElytronDescriptionConstants.FILTER)
-                .setDefaultValue(ModelNode.TRUE)
                 .setAllowExpression(true)
                 .setFlags(AttributeAccess.Flag.RESTART_RESOURCE_SERVICES)
                 .build();
@@ -312,7 +311,6 @@ class LdapRealmDefinition extends SimpleResourceDefinition {
 
         static final SimpleAttributeDefinition USE_RECURSIVE_SEARCH = new SimpleAttributeDefinitionBuilder(ElytronDescriptionConstants.USE_RECURSIVE_SEARCH, ModelType.BOOLEAN, true)
                 .setRequires(ElytronDescriptionConstants.SEARCH_BASE_DN)
-                .setDefaultValue(ModelNode.FALSE)
                 .setAllowExpression(true)
                 .setFlags(AttributeAccess.Flag.RESTART_RESOURCE_SERVICES)
                 .build();
@@ -497,7 +495,7 @@ class LdapRealmDefinition extends SimpleResourceDefinition {
             }
 
             ModelNode useRecursiveSearchNode = IdentityMappingObjectDefinition.USE_RECURSIVE_SEARCH.resolveModelAttribute(context, identityMappingNode);
-            if (useRecursiveSearchNode.asBoolean()) {
+            if (useRecursiveSearchNode.isDefined() && useRecursiveSearchNode.asBoolean()) {
                 identityMappingBuilder.searchRecursive();
             }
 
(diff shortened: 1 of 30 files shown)


michpetrov:
Ok, so `list-add` is performing a read with defaults included and that messes up the model.

michpetrov:
@bstansberry it turns out the change has a greater impact, these tests fail because of how some Elytron attribute definitions are set, it's easy to fix but there more tests failing down the line where the models don't align with expected values

elytron/src/main/java/org/wildfly/extension/elytron/LdapRealmDefinition.java
```diff
@@ -113,7 +113,6 @@
 
         static final SimpleAttributeDefinition RECURSIVE_SEARCH = new SimpleAttributeDefinitionBuilder(ElytronDescriptionConstants.SEARCH_RECURSIVE, ModelType.BOOLEAN, true)
                 .setRequires(ElytronDescriptionConstants.FILTER)
-                .setDefaultValue(ModelNode.TRUE)
```
darranl, line 116: What is the reason for making this change?  By having the default defined which also matches our XML schema it means we can avoid complicating it with the isDefined check.

michpetrov: If an attribute A has a default value and requires attribute B then B has to have a default value as well, otherwise it will fail validation. Top-level attributes are being checked for this but I didn't implement it correctly for nested attributes.

If there is an available default value for FILTER and TO I can change that.

michpetrov: @darranl how do we proceed with this?

michpetrov:
@ehsavoie I have deleted the default value from `domain-controller.remote.protocol`, another option is to supply default values for `host` and `port` but I don't know what those should be.

michpetrov:
@darranl this has been open for a while, what can I do to expedite it?

testsuite/domain/src/test/resources/host-configs/admin-only-discovery.xml
```diff
@@ -27,7 +27,7 @@
         <!-- Remote domain controller configuration with a host and port -->
         <remote authentication-context="secondaryHostAContext" ignore-unused-configuration="true" admin-only-policy="${jboss.test.admin-only-policy}">
             <discovery-options>
-                <static-discovery name="primary" host="${jboss.test.host.primary.address}" port="9999"/>
+                <static-discovery name="primary" protocol="remote" host="${jboss.test.host.primary.address}" port="9999"/>
```
darranl, line 30: These additions look problematic, if we need to make configuration changes in test cases as a result of work in the management model this indicates that these changes will potentially break existing  user configurations that don't have these attributes defined.

darranl:
I think this is still going to need some discussion regarding implications for backwards compatibility, IMO preserving backwards compatibility feels like a higher priority than the original issue.

I don't know if this is correct but one strategy might be to use model bumps and update the older parsers to compensate for the attribute with the default not being present.  If we remove defaults we may need to change the schemas as well.  Then transformers likely are impacted as well.

Another option to consider is relaxing the constraints that means defaults could not be used on these attributes, that may reduce the backwards compatibility impact but then would need review of how the attributes are actually used.

michpetrov:
@darranl I'm gonna doublecheck but I don't think this actually causes compatibility issues.

I should also mention that there are two larger issues, which I figure we should be aware of:
 1. the schema and attribute definitions are not always identical, and we currently have no way to ensure that they are
    - also as far as I can tell some relationships are not possible to model in the schema accurately 
 3. the XML files we use for testing are often not validated (this is solvable but would have to be done manually)

Now, for Elytron the default values are present in the underlying classes, so whether `AttributeDefinition#resolveModelAttribute` returns undefined or the default value doesn't make a difference. This would require a schema bump for removing the default values but more so for showing the relationship between the attributes (A requires B). Though we're kinda making a difference between "this attribute has a default value" and "the variable this attribute represents has a default value".

Regarding the config files - per current schema `protocol` is actually a required attribute for `static-discovery`, so I don't think that's a setup anyone uses because it wouldn't pass validation. It is not required for `remote` but according to the description:
https://github.com/wildfly/wildfly-core/blob/f4221cf2fbd76da106cbf224a7e11ad581875bb2/server/src/main/resources/schema/wildfly-config_20_0.xsd#L1443-L1447
the absence of `protocol` doesn't mean we're expecting a default value to be used (the attribute has no default value defined in the schema).

PR #4856, opened 2021-11-11
[WFCORE-2957] read-resource without include-runtime shows runtime-only resources

petrberan:
https://issues.redhat.com/browse/WFCORE-2957

### DIFF
--- a/controller/src/main/java/org/jboss/as/controller/operations/global/ReadResourceHandler.java
+++ b/controller/src/main/java/org/jboss/as/controller/operations/global/ReadResourceHandler.java
@@ -275,9 +275,10 @@ public class ReadResourceHandler extends GlobalOperationHandlers.AbstractMultiTa
                     ModelNode rrOp = Util.createEmptyOperation(READ_RESOURCE_OPERATION, absoluteChildAddr);
                     PathAddress relativeAddr = PathAddress.pathAddress(childPE);
 
+                    ImmutableManagementResourceRegistration childReg = registry == null ? null : registry.getSubModel(relativeAddr);
+
                     if (recursive) {
                         boolean getChild = false;
-                        ImmutableManagementResourceRegistration childReg = registry == null ? null : registry.getSubModel(relativeAddr);
                         if (childReg != null) {
                             // Decide if we want to invoke on this child resource
                             boolean proxy = childReg.isRemote();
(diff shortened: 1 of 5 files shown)


petrberan:
Follow up on https://github.com/wildfly/wildfly-core/pull/4473 which was closed due to inactivity

yersan (commented):
Hello @petrberan test failures are related to this change.

petrberan:
Thank you @yersan , had a false positive tests on my device. Should be fixed in this commit

petrberan:
Hi @yersan just noticed this PR is still open with test failures. However, I'm not getting any failures on my device, could you retest this one please?

yersan:
Hi, @petrberan I've kicked it off again, you can do it by opening the "Details" link and clicking on the "Run" button at the upper corner.

In any case, this PR is still under discussion see https://github.com/wildfly/wildfly-core/pull/4473#issuecomment-804503317

petrberan:
Sorry about that @yersan , got few problems in my branches. Anyway, I pushed the proposed changes with the latest commit. Is there anything I can do to make this PR merged?

yersan (changes requested):
@petrberan I reviewed again this and found a minor detail to address on one of the test cases. Besides that minor detail, it looks good to me and we can merge it after the approvals.


@bstansberry Would you mind taking another look and approve on your side? there were some discussions in the past about it, but the summary is the PR is going to remove runtime resources from reading operations. I've been checking HAL in case of finding any obvious error, but it looks fine.

Since we are making major changes, I think we can proceed with these, after all, not to show runtime values when include-runtime is false is the correct behavior.

petrberan:
Hi @yersan @bstansberry can I ask for review please?

petrberan:
Can I ask for review please @yersan @bstansberry ? The failed test is not related to the PR, got all greens on my part

yersan (approved):
@bstansberry I gave my approval since my last review since my last change request was addressed, but since these changes affect the standard read operations returning something different, my main concern here is HA or any other client.

We can argue that runtime resources should not be shown if you are not enabling them, but since merging this can open the door to bugs on those clients (or corrections needed on those clients) I'm still running into doubts, so I would like to hear your opinion here to also ensure I'm not missing something.

yersan:
In any case, I would defer this to WF29 so we can get some bake

bstansberry (approved):
@yersan @petrberan LGTM.

yersan:
Thanks, @petrberan

PR #4857, opened 2021-11-11
[WFCORE-5694] RealmsTestCase fails on Mac

rmartinc:
Issue: https://issues.redhat.com/browse/WFCORE-5694

Test {{RealmsTestCase.testPropertyRealm}} failed in macos because it seems that the FS sets the modification timestamp with second precision. Just adding a loop to wait two seconds for the time change.


kabir (approved):

jmesnil:
@rmartinc thanks

PR #4866, opened 2021-11-14
[WFCORE-5708] Add a new "org.wildfly.security.http.sfbasic" module.

darranl:
Added the hold label as this will require an update of WildFly Elytron but I need an upstream PR.

https://issues.redhat.com/browse/WFCORE-5708

### DIFF
--- a/core-feature-pack/common/pom.xml
+++ b/core-feature-pack/common/pom.xml
@@ -326,6 +326,10 @@
         </dependency>
         <dependency>
             <groupId>org.wildfly.security</groupId>
+            <artifactId>wildfly-elytron-http-stateful-basic</artifactId>
+        </dependency>
+        <dependency>
+            <groupId>org.wildfly.security</groupId>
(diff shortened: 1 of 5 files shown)


fjuma:
Looks good, an update will also be needed in `core-feature-pack-common-licenses.xml`.

fjuma (approved):
Marking as approved, just needs an update in core-feature-pack-common-licenses.xml.

gaol:
Maybe we need to add `org.wildfly.security.http.sfbasic` to `NOT_USED` in [LayersTestCase of WildFly testsuite](https://github.com/wildfly/wildfly/blob/main/testsuite/layers/src/test/java/org/jboss/as/test/layers/LayersTestCase.java#L71) as well to pass `Galleon Full Integration - Linux` ?

The same for #4879

gaol:
I opened a PR https://github.com/wildfly/wildfly/pull/14970 to address the changes in wildfly

PR #4871, opened 2021-11-19
[WFCORE-5715] Fix permission check failures that occur in the properties realm with the security manager enabled

fjuma:
https://issues.redhat.com/browse/WFCORE-5715

https://github.com/wildfly/wildfly-core/pull/4821 updated the Elytron properties realm to check the last modified time of the properties files to see if they need to be reloaded. This PR fixes the WildFly test failures that started occurring with this change.


bstansberry (approved):

jmesnil:
@fjuma thanks!

PR #4875, opened 2021-11-25
[WFCORE-5725] Upgrade to Byteman 4.0.18

ropalka:
https://issues.redhat.com/browse/WFCORE-5725


darranl:
Just running the Windows job again, in my aggregate PR it also stalled at the same location so just checking if it is a conincidence.

PR #4890, opened 2021-12-09
[WFCORE-5742] Upgrade WildFly Galleon Plugin to 5.2.7.Final

bstansberry:
https://issues.redhat.com/browse/WFCORE-5742


bstansberry:
@jmesnil https://github.com/wildfly/wildfly/pull/15014 does this upstream. That will be merged into WF main for WF 26 today (assuming ci is ok). I don't think this is critical for the WF Core 18 final, but otoh staying in sync is good. So I'm just calling it to your attention in case you want it.

jmesnil (approved):

jmesnil:
@bstansberry thanks

PR #4897, opened 2021-12-13
update pom.xml to use log4j 2.12.1

sumantro93:


jamezp:
Thank you for your contribution @sumantro93. Unfortunately this won't work as we can't simply replace log4j 1 with log4j 2. We already have a log4j 2 dependency.

PR #4899, opened 2021-12-14
Add ModelTestControllerVersion for XP 4

kabir:
https://issues.redhat.com/browse/WFCORE-5751

https://issues.redhat.com/browse/WFCORE-5752

### DIFF
--- a/model-test/src/main/java/org/jboss/as/model/test/ModelTestControllerVersion.java
+++ b/model-test/src/main/java/org/jboss/as/model/test/ModelTestControllerVersion.java
@@ -55,6 +55,13 @@ public enum ModelTestControllerVersion {
     @Deprecated
     EAP_7_3_0("7.3.0.GA-redhat-00004", true, "18.0.0", "10.1.2.Final-redhat-00001", "7.3.0"),
     EAP_7_4_0("7.4.0.GA-redhat-00005", true, "23.0.0", "15.0.2.Final-redhat-00001", "7.4.0"),
+
+    // https://issues.redhat.com/browse/WFCORE-5753
+    // Once EAP XP 4 is out, we need to replace the following live with
+    // EAP_XP_4("4.0.0.GA-redhat-0000x", true, "24.0.0", "18.0.0.Final-redhat-0000x", "xp4")
+    // We will also need to remove the *-wf26 files in wildfly-legacy-test and replace them with *-xp4 ones
+    // generated from EAP 7.4.0 with XP 4 installed
+    EAP_XP_4("26.0.0.Beta1", false, "25.0.0", "18.0.0.Final", "wf26")
     ;
 
     private final String mavenGavVersion;
--- a/pom.xml
+++ b/pom.xml
@@ -226,7 +226,7 @@
         <version.org.wildfly.client.config>1.0.1.Final</version.org.wildfly.client.config>
         <version.org.wildfly.common>1.5.4.Final</version.org.wildfly.common>
         <version.org.wildfly.discovery>1.2.1.Final</version.org.wildfly.discovery>
-        <version.org.wildfly.legacy.test>6.0.1.Final</version.org.wildfly.legacy.test>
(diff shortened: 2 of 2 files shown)


model-test/src/main/java/org/jboss/as/model/test/ModelTestControllerVersion.java
```diff
@@ -55,6 +55,13 @@
     @Deprecated
     EAP_7_3_0("7.3.0.GA-redhat-00004", true, "18.0.0", "10.1.2.Final-redhat-00001", "7.3.0"),
     EAP_7_4_0("7.4.0.GA-redhat-00005", true, "23.0.0", "15.0.2.Final-redhat-00001", "7.4.0"),
+
+    // https://issues.redhat.com/browse/WFCORE-5753
+    // Once EAP XP 4 is out, we need to replace the following live with
+    // EAP_XP_4("4.0.0.GA-redhat-0000x", true, "23.0.0", "18.0.0.Final-redhat-0000x", "xp4")
+    // We will also need to remove the *-wf26 files in wildfly-legacy-test and replace them with *-xp4 ones
+    // generated from EAP 7.4.0 with XP 4 installed
+    EAP_XP_4("26.0.0.Beta1", false, "25.0.0", "18.0.0.Final", "wf26")
```
yersan, line 64: It should be fine, although I don't understand at all why you added the WF Model descriptions for WF26 (https://github.com/wildfly/wildfly-legacy-test/commit/ffab0be196205c88f8721195bf63f4a3fe02ba89), indeed using a Beta1 release instead of a Final, and then you added the legacy controllers for WF25 (https://github.com/wildfly/wildfly-legacy-test/commit/b214f32b3dcf27c2da7a38548c8e660fffbbfa87)

Probably they are the same but seems a bit misleading since we are getting here an enum that uses WF26 with WF25 controllers.

@kabir Is this configuration expected by you?

I would expect to have WF26 model descriptions based on a Final release on wildfly-legacy-test and WF26 subsystems and controllers instead of WF25, but I'm not sure about the intentions behind this.

kabir: Hi @yersan, there are a few parts to this:

* 26.0.0.Beta1 and 18.0.0.Final are the versions which will be used at runtime by the legacy controllers.
* 25.0.0 is the version of the legacy controller adapters to use (i.e. the subfolder in wildfly-legacy-test under core/ and subsystem/). These are compiled against a version of the controller API, but really the versions above specify what is pulled in for the legacy controller at runtime.

We introduce these adapters when there is a breaking change. In this case the breaking change was introduced with WildFly 25.0.0 where the VaultReader was removed from various server module APIs used by these server adapters.

I agree it is slightly confusing, but in the (very unlikely) case we would need transformers tests for WildFly 25 at some stage we can reuse the "25.0.0" versions. And if there are no breaking changes in the next few controller versions, we would be able to use an enum like the following for that:

```
FUTURE_RELEASE("28.0.0.Final", false, "25.0.0", "20.0.0.Final", "wf28")
```
So IMO it is correct.

Although, I see I have a typo on the following line:
```
 // EAP_XP_4("4.0.0.GA-redhat-0000x", true, "23.0.0", "18.0.0.Final-redhat-0000x", "xp4")
```
Here "23.0.0" should be "25.0.0", I'll replace that.

kabir: Or to put it another way, if the VaultReader had not been removed from the server in WildFly 25, we could have continued using the "23.0.0" adapters used in the previous EAP_7_4_0 version :-)

yersan: Ok, I understand. It was more natural to me to think more about even versions that belong all to the referred legacy version.

This legacy test framework still has a lot of unknown details to me, so I only commented on what I saw weirdly. If you are fine with the changes on `wildfly-legacy-test` and how they are brought in, I'm fine too.

yersan (approved):

jmesnil:
@kabir thanks

PR #4910, opened 2021-12-27
WFCORE-5760 Replace deprecated EnumValidator constructors and methods

clara0:
https://issues.redhat.com/browse/WFCORE-5760 

Changes made for equivalent results:
Constructors:

```java
EnumValidator(Class, boolean, E...) --> EnumValidator(Class, E...)
EnumValidator(Class, boolean, boolean) --> create(Class, E...)
EnumValidator(Class, boolean, boolean, E...) --> EnumValidator(Class, E...)
```

Methods:
```java
create(Class, boolean, E...) --> create(Class, E...)
create(Class, boolean, boolean) --> create(Class, E...)
create(Class, boolean, boolean, E...) --> create(Class, E...)
create(Class, boolean, boolean, E...) --> create(Class, EnumSet)
```

### DIFF
--- a/host-controller/src/main/java/org/jboss/as/domain/controller/operations/DomainServerLifecycleHandlers.java
+++ b/host-controller/src/main/java/org/jboss/as/domain/controller/operations/DomainServerLifecycleHandlers.java
@@ -88,7 +88,7 @@ public class DomainServerLifecycleHandlers {
 
     private static final AttributeDefinition START_MODE = SimpleAttributeDefinitionBuilder.create(ModelDescriptionConstants.START_MODE, ModelType.STRING, true)
             .setDefaultValue(new ModelNode(StartMode.NORMAL.toString()))
-            .setValidator(new EnumValidator<>(StartMode.class, true, true))
+            .setValidator(new EnumValidator<>(StartMode.class, EnumSet.allOf(StartMode.class)))
             .build();
 
     @Deprecated
--- a/host-controller/src/main/java/org/jboss/as/domain/controller/operations/DomainServerLifecycleHandlers.java
+++ b/host-controller/src/main/java/org/jboss/as/domain/controller/operations/DomainServerLifecycleHandlers.java
@@ -87,8 +87,9 @@ public class DomainServerLifecycleHandlers {
             .build();
 
     private static final AttributeDefinition START_MODE = SimpleAttributeDefinitionBuilder.create(ModelDescriptionConstants.START_MODE, ModelType.STRING, true)
+            .setAllowExpression(true)
             .setDefaultValue(new ModelNode(StartMode.NORMAL.toString()))
-            .setValidator(new EnumValidator<>(StartMode.class, EnumSet.allOf(StartMode.class)))
+            .setValidator(EnumValidator.create(StartMode.class))
             .build();
 
     @Deprecated
--- a/logging/src/main/java/org/jboss/as/logging/handlers/AsyncHandlerResourceDefinition.java
+++ b/logging/src/main/java/org/jboss/as/logging/handlers/AsyncHandlerResourceDefinition.java
@@ -89,7 +89,7 @@ public class AsyncHandlerResourceDefinition extends AbstractHandlerDefinition {
             .setDefaultValue(new ModelNode(OverflowAction.BLOCK.name()))
             .setPropertyName("overflowAction")
             .setResolver(OverflowActionResolver.INSTANCE)
-            .setValidator(EnumValidator.create(OverflowAction.class, false, false))
+            .setValidator(EnumValidator.create(OverflowAction.class))
             .build();
 
     static final SimpleAttributeDefinition HANDLER = SimpleAttributeDefinitionBuilder.create("handler", ModelType.STRING)
--- a/logging/src/main/java/org/jboss/as/logging/handlers/ConsoleHandlerResourceDefinition.java
+++ b/logging/src/main/java/org/jboss/as/logging/handlers/ConsoleHandlerResourceDefinition.java
@@ -50,7 +50,7 @@ public class ConsoleHandlerResourceDefinition extends AbstractHandlerDefinition 
             .setAttributeMarshaller(ElementAttributeMarshaller.NAME_ATTRIBUTE_MARSHALLER)
             .setDefaultValue(new ModelNode(Target.SYSTEM_OUT.toString()))
             .setResolver(TargetResolver.INSTANCE)
-            .setValidator(EnumValidator.create(Target.class, true, false))
+            .setValidator(EnumValidator.create(Target.class))
             .build();
 
     private static final AttributeDefinition[] ATTRIBUTES = Logging.join(DEFAULT_ATTRIBUTES, AUTOFLUSH, TARGET, NAMED_FORMATTER);
--- a/remoting/subsystem/src/main/java/org/jboss/as/remoting/RemoteOutboundConnectionResourceDefinition.java
+++ b/remoting/subsystem/src/main/java/org/jboss/as/remoting/RemoteOutboundConnectionResourceDefinition.java
@@ -24,6 +24,8 @@ package org.jboss.as.remoting;
 
 import static org.jboss.as.remoting.Capabilities.AUTHENTICATION_CONTEXT_CAPABILITY;
 
+import java.util.EnumSet;
+
 import org.jboss.as.controller.AttributeDefinition;
 import org.jboss.as.controller.ModelVersion;
 import org.jboss.as.controller.OperationStepHandler;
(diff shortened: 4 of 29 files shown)


host-controller/src/main/java/org/jboss/as/domain/controller/operations/DomainServerLifecycleHandlers.java
```diff
@@ -88,7 +88,7 @@
 
     private static final AttributeDefinition START_MODE = SimpleAttributeDefinitionBuilder.create(ModelDescriptionConstants.START_MODE, ModelType.STRING, true)
             .setDefaultValue(new ModelNode(StartMode.NORMAL.toString()))
-            .setValidator(new EnumValidator<>(StartMode.class, true, true))
```
jmesnil, line 91: shouldn't you add `setAllowExpressions(true)` as the 2nd boolean is `true`?

logging/src/main/java/org/jboss/as/logging/handlers/AsyncHandlerResourceDefinition.java
```diff
@@ -89,7 +89,7 @@ public void marshallAsElement(final AttributeDefinition attribute, final ModelNo
             .setDefaultValue(new ModelNode(OverflowAction.BLOCK.name()))
             .setPropertyName("overflowAction")
             .setResolver(OverflowActionResolver.INSTANCE)
-            .setValidator(EnumValidator.create(OverflowAction.class, false, false))
+            .setValidator(EnumValidator.create(OverflowAction.class))
```
jmesnil, line 92: there was a mismatch here. The validator was set with `allow expressions set to false` while the builder itself was using `setAllowExpression(true)`.

AIUI, in the validation code, that means that if there was an expression, it was replaced *before* validation with the validator.

logging/src/main/java/org/jboss/as/logging/handlers/ConsoleHandlerResourceDefinition.java
```diff
@@ -50,7 +50,7 @@
             .setAttributeMarshaller(ElementAttributeMarshaller.NAME_ATTRIBUTE_MARSHALLER)
             .setDefaultValue(new ModelNode(Target.SYSTEM_OUT.toString()))
             .setResolver(TargetResolver.INSTANCE)
-            .setValidator(EnumValidator.create(Target.class, true, false))
```
jmesnil, line 53: same remark, there was a mismatch between `setAllowExpression(true)` and the EnumValidator 2nd boolean parameter

remoting/subsystem/src/main/java/org/jboss/as/remoting/RemoteOutboundConnectionResourceDefinition.java
```diff
@@ -72,8 +72,7 @@
             .build();
 
     public static final SimpleAttributeDefinition PROTOCOL = new SimpleAttributeDefinitionBuilder(
-            CommonAttributes.PROTOCOL, ModelType.STRING, true).setValidator(
-                    new EnumValidator<Protocol>(Protocol.class, true, false))
+            CommonAttributes.PROTOCOL, ModelType.STRING, true).setValidator(EnumValidator.create(Protocol.class))
             .setDefaultValue(new ModelNode(Protocol.HTTP_REMOTING.toString()))
             .setAllowExpression(true)
```
jmesnil, line 77: same remark

jmesnil (changes requested):
@clara0 Could you please double-check the comments I've added?
We also need to wait for the CI jobs to run to catch any incompatible changes.

thanks!

logging/src/main/java/org/jboss/as/logging/handlers/AsyncHandlerResourceDefinition.java
clara0: Hi, I think the `setAllowExpression(true)` is correct because the `validateParameter()` method in the `EnumValidator` class doesn't seem to check whether the value is true or false.

https://github.com/wildfly/wildfly-core/blob/main/controller/src/main/java/org/jboss/as/controller/operations/validation/EnumValidator.java#L220

logging/src/main/java/org/jboss/as/logging/handlers/ConsoleHandlerResourceDefinition.java
clara0: Hi, I think the `setAllowExpression(true)` is correct because the `validateParameter()` method in the `EnumValidator` class doesn't seem to check whether the value is true or false.

https://github.com/wildfly/wildfly-core/blob/main/controller/src/main/java/org/jboss/as/controller/operations/validation/EnumValidator.java#L220

remoting/subsystem/src/main/java/org/jboss/as/remoting/RemoteOutboundConnectionResourceDefinition.java
clara0: Hi, I think the `setAllowExpression(true)` is correct because the `validateParameter()` method in the `EnumValidator` class doesn't seem to check whether the value is true or false.

https://github.com/wildfly/wildfly-core/blob/main/controller/src/main/java/org/jboss/as/controller/operations/validation/EnumValidator.java#L220

host-controller/src/main/java/org/jboss/as/domain/controller/operations/DomainServerLifecycleHandlers.java
clara0: Fixed

logging/src/main/java/org/jboss/as/logging/handlers/AsyncHandlerResourceDefinition.java
bstansberry: @jmesnil @clara0 The reason it's ok to just remove this boolean parameters passed to validator constructors, without worrying about their values is that the validators passed to any .setValidator is always wrapped with an internal validator that handles the nullable and expression allowed checks based on the AttributeDefinition settings.

The validator actually used by an AttributeDefinition is always passed through this:

https://github.com/wildfly/wildfly-core/blob/main/controller/src/main/java/org/jboss/as/controller/AttributeDefinition.java#L111

This is to ensure the settings on the AD itself control without inconsistent settings on a validator resulting in different behavior from what the AD describes in :read-resource-description or read-operation-description.

The EnumValidator constructors that take those params are deprecated because the params are cruft due to the above.

clara0:
@jmesnil any other comments?

server/src/main/java/org/jboss/as/server/controller/resources/DeploymentAttributes.java
```diff
@@ -127,7 +127,7 @@ public void marshallAsAttribute(AttributeDefinition attribute, ModelNode resourc
             .build();
 
     public static final AttributeDefinition STATUS = SimpleAttributeDefinitionBuilder.create(ModelDescriptionConstants.STATUS, ModelType.STRING, true)
-        .setValidator(new EnumValidator<AbstractDeploymentUnitService.DeploymentStatus>(AbstractDeploymentUnitService.DeploymentStatus.class, true))
+        .setValidator(new EnumValidator<AbstractDeploymentUnitService.DeploymentStatus>(AbstractDeploymentUnitService.DeploymentStatus.class))
```
yersan, line 130: For consistency, this should be probably replaced by the EnumValidator.create(..) version. Anyway, it shouldn't block merging this.

yersan (commented):
The changes look good to me. I just added a small optional detail. 
It would need to squash the two commits. 

@jmesnil Do you want to add anything else before asking for squashing the commit?

server/src/main/java/org/jboss/as/server/controller/resources/DeploymentAttributes.java
clara0: Hi, I think that originally a new EnumValidator with no values would be created since there are no varargs. Using the EnumValidator.create() method would create an EnumValidator with all enum values, so I decided to use the constructor in order to keep consistent with the original.

yersan: Hello @clara0 , you're right, thanks for clarifying it. What you have described is correct to me too.

yersan: Although the deprecation message at the [old constructor](https://github.com/wildfly/wildfly-core/blob/0d4059b79850503ec612e7717faeb119a599be57/controller/src/main/java/org/jboss/as/controller/operations/validation/EnumValidator.java#L94) states that this constructor is equivalent to calling it with `EnumSet#allOf(Class)`

I'm not sure yet how one way or the other could affect it, so, my conclusion of this point is your approach is the safest one.

yersan (approved):

bstansberry:
Thanks @clara0

PR #4911, opened 2022-01-04
[WFCORE-5764] Can't write/acquire credentials from a store using the elytron-tool script

gaol:

Issue: https://issues.redhat.com/browse/WFCORE-5764

### DIFF
--- a/core-feature-pack/common/src/main/resources/modules/system/layers/base/org/wildfly/security/elytron-tool/main/module.xml
+++ b/core-feature-pack/common/src/main/resources/modules/system/layers/base/org/wildfly/security/elytron-tool/main/module.xml
@@ -40,7 +40,7 @@
         <module name="org.jboss.logging" />
         <module name="org.jboss.logmanager" />
         <module name="org.slf4j" />
-        <module name="org.wildfly.security.elytron-private" />
+        <module name="org.wildfly.security.elytron-private" services="import"/>
         <module name="org.wildfly.security.elytron-tool-addons" services="import" optional="true"/>
         <module name="org.wildfly.common" />
     </dependencies>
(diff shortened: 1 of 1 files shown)


core-feature-pack/common/src/main/resources/modules/system/layers/base/org/wildfly/security/elytron-tool/main/module.xml
```diff
@@ -40,7 +40,7 @@
         <module name="org.jboss.logging" />
         <module name="org.jboss.logmanager" />
         <module name="org.slf4j" />
-        <module name="org.wildfly.security.elytron-private" />
+        <module name="org.wildfly.security.elytron-private" services="import"/>
```
gaol, line 43: I am not sure if this dependency should be `org.wildfly.security.elytron-base`, for the reference, the new module [org.wildfly.security.http.sfbasic](https://github.com/wildfly/wildfly-core/blob/main/core-feature-pack/common/src/main/resources/modules/system/layers/base/org/wildfly/security/http/sfbasic/main/module.xml#L32) has the elytron-base dependency.

fjuma: This could be changed to `org.wildfly.security.elytron-base` (`elytron-private` depends on `elytron-base` and a few other deps but it doesn't look like those are actually needed by `elytron-tool`).

fjuma (approved):
Thanks very much for working on this @gaol!

gaol:
Thank you @fjuma for the review, I updated to dependent module to `org.wildfly.security.elytron-base`.

It is good that the CI comes back :)

jmesnil:
@gaol thanks

PR #4917, opened 2022-01-05
[WFCORE-5767] Add the Elytron OIDC Client related dependencies to Core

fjuma:
https://issues.redhat.com/browse/WFCORE-5767

Moving the Elytron OIDC Client related dependencies from WildFly to Core. The modules that actually make use of these dependencies will remain in WildFly.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -1845,6 +1845,11 @@
             </dependency>
             <dependency>
                 <groupId>org.wildfly.security</groupId>
+                <artifactId>wildfly-elytron-http-oidc</artifactId>
+                <version>${version.org.wildfly.security.elytron}</version>
+            </dependency>
+            <dependency>
+                <groupId>org.wildfly.security</groupId>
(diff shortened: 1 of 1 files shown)


boris-unckel:
@fjuma Do you need to add it to https://github.com/wildfly/wildfly-core/blob/main/core-feature-pack/common/src/main/resources/license/core-feature-pack-common-licenses.xml ?

fjuma:
@boris-unckel Good question. Since these dependencies aren't actually being used by anything in Core and only get used by the Elytron OIDC layer in WildFly, I think we can leave these in the https://github.com/wildfly/wildfly/blob/main/elytron-oidc-client/galleon-common/src/main/resources/license/elytron-oidc-feature-pack-licenses.xml file.

darranl:
+1 this should not impact the modules provisioned by the core feature pack so should not need listing in the licenses.

jmesnil (approved):

PR #4920, opened 2022-01-08
[WFCORE-5768] Change default behaviour to resolve jboss parent pom

boris-unckel:
Fixes https://issues.redhat.com/browse/WFCORE-5768


bstansberry:
Thanks, @boris-unckel

PR #4924, opened 2022-01-11
[WFCORE-5343]: YAML support for configuration customization.

ehsavoie:
**Disabled by default experimental feature**

* filtering elements to be processed
* moving logs to debug
* adding some perfs traces
* Multi file support
* Adding support for ignoring some elements
* Adding tags to support:
  - deleting a resource
  - undefining an attribute
  - list-add with index support
* trying to be smart to detect cli scripts application to set InteractionPolicy
* support reloading
* Adding support for multiple defintion of the same key.
* Adding support for flattening YAML
* Making the YAML support easily removable

Jira: https://issues.redhat.com/browse/WFCORE-5343


jmesnil:
For the record, usually, we don't accept such big patch in WildFly Core micro releases (that typically includes bug fixes and component upgrades only.
But this feature is experimental, well-contained and disabled by default, so I'm fine with incorporating in the upcoming 18.0.3.Final release

PR #4925, opened 2022-01-11
[WFCORE-5490] Integrate Elytron credential expressions into system-pr…

bstansberry:
…operty resource handling.

As was done with vault expressions, if expressions in system-property resource value attributes cannot be resolved in Stage.MODEL, defer and try in Stage.RUNTIME.

Remove capability-based integration with RuntimeExpressionResolver by Elytron and deprecate the general-use capability based integration. Replace with integration via the ExtensionContext via an API that accounts for the fact that resolution may be attempted in Stage.MODEL.

https://issues.redhat.com/browse/WFCORE-5490


fjuma (approved):
Thanks @bstansberry! This looks good to me.

jmesnil:
@bstansberry thanks!

PR #4934, opened 2022-01-19
[WFCORE-5778] Reset 'java.security.manager' default value back to "allow" by default in launcher

ropalka:
https://issues.redhat.com/browse/WFCORE-5778

### DIFF
--- a/launcher/src/main/java/org/wildfly/core/launcher/DomainCommandBuilder.java
+++ b/launcher/src/main/java/org/wildfly/core/launcher/DomainCommandBuilder.java
@@ -581,7 +581,14 @@ public class DomainCommandBuilder extends AbstractCommandBuilder<DomainCommandBu
         if (arg != null && !arg.trim().isEmpty()) {
             final Argument argument = Arguments.parse(arg);
             if (argument.getKey().equals(SECURITY_MANAGER_PROP)) {
-                setUseSecurityManager(true);
+                if (ALLOW_VALUE.equals(argument.getValue())) {
+                    // [WFCORE-5778] java.security.manager system property with value "allow" detected.
+                    // It doesn't mean SM is going to be installed but it indicates SM can be installed dynamically.
+                    setUseSecurityManager(false);
+                } else {
+                    setUseSecurityManager(true);
+                }
+                setUseSecurityManager(!ALLOW_VALUE.equals(argument.getValue()));
             } else {
                 processControllerJavaOpts.add(argument);
             }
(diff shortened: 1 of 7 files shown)


jmesnil:
@fjuma could you please review this one & #4936? I think we should merge both at the same time

launcher/src/main/java/org/wildfly/core/launcher/DomainCommandBuilder.java
```diff
@@ -581,7 +581,14 @@ public DomainCommandBuilder addProcessControllerJavaOption(final String arg) {
...
             if (argument.getKey().equals(SECURITY_MANAGER_PROP)) {
-                setUseSecurityManager(true);
+                if (ALLOW_VALUE.equals(argument.getValue())) {
+                    // [WFCORE-5778] java.security.manager system property with value "allow" detected.
+                    // It doesn't mean SM is going to be installed but it indicates SM can be installed dynamically.
+                    setUseSecurityManager(false);
+                } else {
+                    setUseSecurityManager(true);
+                }
+                setUseSecurityManager(!ALLOW_VALUE.equals(argument.getValue()));
```
fjuma, line 591: Looks like the `if` block right above this can be removed since `setUseSecurityManager` is being set twice.

launcher/src/main/java/org/wildfly/core/launcher/StandaloneCommandBuilder.java
```diff
@@ -126,7 +126,13 @@ public StandaloneCommandBuilder addJavaOption(final String jvmArg) {
...
                     break;
                 case SECURITY_MANAGER_PROP:
-                    setUseSecurityManager(true);
+                    if (ALLOW_VALUE.equals(argument.getValue())) {
+                        // [WFCORE-5778] java.security.manager system property with value "allow" detected.
+                        // It doesn't mean SM is going to be installed but it indicates SM can be installed dynamically.
+                        setUseSecurityManager(false);
+                    } else {
+                        setUseSecurityManager(true);
+                    }
```
fjuma, line 135: Just a minor comment but this could be simplified to `setUseSecurityManager(!ALLOW_VALUE.equals(argument.getValue()));` like in the `DomainCommandBuilder`.

launcher/src/main/java/org/wildfly/core/launcher/DomainCommandBuilder.java
ropalka: Fixed.

launcher/src/main/java/org/wildfly/core/launcher/StandaloneCommandBuilder.java
ropalka: Fixed.

ropalka:
I closed this PR by mistake @fjuma . The follow up PR with your proposal changes applied is here: https://github.com/wildfly/wildfly-core/pull/4939

PR #4936, opened 2022-01-21
[WFCORE-5782] Reset 'java.security.manager' default value back to "allow" by default in WildFly shell scripts

ropalka:
https://issues.redhat.com/browse/WFCORE-5782


fjuma (approved):
Thanks @ropalka!

fjuma:
@jamezp It would be good if you could take a look at this one as well.

Note that as Jeff pointed out, we should merge this and https://github.com/wildfly/wildfly-core/pull/4934 at the same time.

jamezp (changes requested):
This looks good, but the launcher would need changes too for Arquillian and others to work properly.

ropalka:
Right @jamezp , launcher changes are in separate PR here: https://github.com/wildfly/wildfly-core/pull/4934

jamezp (approved):

jmesnil:
@ropalka thanks

PR #4939, opened 2022-01-26
[WFCORE-5778] Reset 'java.security.manager' default value back to "allow" by default in launcher

ropalka:
https://issues.redhat.com/browse/WFCORE-5778

Supersedes https://github.com/wildfly/wildfly-core/pull/4934 which I closed by mistake :(

### DIFF
--- a/launcher/src/main/java/org/wildfly/core/launcher/DomainCommandBuilder.java
+++ b/launcher/src/main/java/org/wildfly/core/launcher/DomainCommandBuilder.java
@@ -581,7 +581,9 @@ public class DomainCommandBuilder extends AbstractCommandBuilder<DomainCommandBu
         if (arg != null && !arg.trim().isEmpty()) {
             final Argument argument = Arguments.parse(arg);
             if (argument.getKey().equals(SECURITY_MANAGER_PROP)) {
-                setUseSecurityManager(true);
+                // [WFCORE-5778] java.security.manager system property with value "allow" detected.
+                // It doesn't mean SM is going to be installed but it indicates SM can be installed dynamically.
+                setUseSecurityManager(isJavaSecurityManagerConfigured(argument));
             } else {
                 processControllerJavaOpts.add(argument);
             }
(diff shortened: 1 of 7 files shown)


launcher/src/main/java/org/wildfly/core/launcher/DomainCommandBuilder.java
```diff
@@ -581,7 +581,9 @@ public DomainCommandBuilder addProcessControllerJavaOption(final String arg) {
         if (arg != null && !arg.trim().isEmpty()) {
             final Argument argument = Arguments.parse(arg);
             if (argument.getKey().equals(SECURITY_MANAGER_PROP)) {
-                setUseSecurityManager(true);
+                // [WFCORE-5778] java.security.manager system property with value "allow" detected.
+                // It doesn't mean SM is going to be installed but it indicates SM can be installed dynamically.
+                setUseSecurityManager(!ALLOW_VALUE.equals(argument.getValue()));
```
jamezp, line 586: This would be set to `true` if `-Djava.security.manager=disallow` was passed. That seems a bit odd to me.

ropalka: I agree with you @jamezp that it looks weird. But that is caused by the fact that JDK development team decided to reuse -Djava.security.manager system property for configuring JVM runtime behavior. Two special values "allow" and "disallow" are just two values of the switch that is defining if JVM should throw UnsupportedOperationException or not on an attempt to install SecurityManager dynamically.

ropalka: But if either "allow" or "disallow" is set it still means Security Manager wasn't configured.

ropalka: Thanks for pointing that out @jamezp - fixed.

jamezp (approved):

jmesnil:
@ropalka thanks

PR #4942, opened 2022-01-26
[WFCORE-5784] Deprecate the org.jboss.log4j.logmanager module. Also a…

jamezp:
…dded deprecation log messages when a log4j configuration file is used or an appender is used as a custom-handler.

https://issues.redhat.com/browse/WFCORE-5784


jmesnil (approved):

jmesnil:
@jamezp thanks

PR #4728, opened 2021-08-31
[WFCORE-5406] Add --add-opens to the SE 9+ VM launch for expected reflective access by the server (round 2)

ropalka:
https://issues.redhat.com/browse/WFCORE-5406

Further opens clauses are needed after recent upstream changes.

PR #4966, opened 2022-02-09
[WFCORE-4296][WFCORE-5406] Add required --add-opens for java.base/com.sun.net.ssl.internal.ssl when HTTPS is configured

ropalka:
https://issues.redhat.com/browse/WFCORE-4296
https://issues.redhat.com/browse/WFCORE-5406

This PR follows up on:
 * https://github.com/wildfly/wildfly-core/pull/4728
 * https://github.com/wildfly/wildfly-core/pull/4728

This PR replaces:
 * https://github.com/wildfly/wildfly/pull/15208

### DIFF
--- a/host-controller/src/main/java/org/jboss/as/host/controller/jvm/JvmType.java
+++ b/host-controller/src/main/java/org/jboss/as/host/controller/jvm/JvmType.java
@@ -120,6 +120,13 @@ public final class JvmType {
         return isModularJvm ? DEFAULT_MODULAR_JVM_ARGUMENTS : Collections.EMPTY_LIST;
     }
 
+    /**
+     * The expectation is this method is called once per JvmType instance.
+     * In theory we could cache the result for reuse, but if we change how
+     * we use these instances it may not be the case that a cached value
+     * would be correct over multiple invocations.
+     * @return collection of optional (may not be available on all JDK versions) modular JVM arguments
+     */
     public Collection<String> getOptionalDefaultArguments() {
         if (isModularJvm) {
             Collection<String> retVal = null;
@@ -186,7 +193,7 @@ public final class JvmType {
             process = builder.redirectErrorStream(true)
                     .redirectOutput(stdout.toFile()).start();
 
-            if (process.waitFor(1, TimeUnit.SECONDS)) {
+            if (process.waitFor(30, TimeUnit.SECONDS)) {
                 result = process.exitValue() == 0;
             } else {
                 result = false;
--- a/host-controller/src/main/java/org/jboss/as/host/controller/jvm/JvmType.java
+++ b/host-controller/src/main/java/org/jboss/as/host/controller/jvm/JvmType.java
@@ -22,11 +22,18 @@
 
 package org.jboss.as.host.controller.jvm;
 
+import java.io.BufferedReader;
 import java.io.File;
+import java.io.FileInputStream;
+import java.io.InputStreamReader;
+import java.io.IOException;
+import java.nio.file.Files;
+import java.nio.file.Path;
 import java.util.ArrayList;
 import java.util.Collection;
 import java.util.Collections;
 import java.util.Locale;
+import java.util.concurrent.TimeUnit;
 
 import org.jboss.as.host.controller.logging.HostControllerLogger;
 import org.wildfly.security.manager.WildFlySecurityManager;
@@ -42,7 +49,10 @@ public final class JvmType {
     private static final String JAVA_HOME_SYS_PROP = "java.home";
     private static final String JAVA_HOME_ENV_VAR = "JAVA_HOME";
     private static final String OS_NAME_SYS_PROP = "os.name";
+    /** Packages being exported or opened are available on all JVMs */
     private static final Collection<String> DEFAULT_MODULAR_JVM_ARGUMENTS;
+    /** Packages being exported or opened may not be available on all JVMs */
+    private static final Collection<String> DEFAULT_OPTIONAL_MODULAR_JVM_ARGUMENTS;
     private static final String JAVA_EXECUTABLE;
     private static final String JAVA_UNIX_EXECUTABLE = "java";
     private static final String JAVA_WIN_EXECUTABLE = "java.exe";
@@ -63,7 +73,6 @@ public final class JvmType {
         modularJavaOpts.add("--add-exports=java.naming/com.sun.jndi.url.ldap=ALL-UNNAMED");
         modularJavaOpts.add("--add-exports=java.naming/com.sun.jndi.url.ldaps=ALL-UNNAMED");
         modularJavaOpts.add("--add-exports=jdk.naming.dns/com.sun.jndi.dns=ALL-UNNAMED");
-        modularJavaOpts.add("--add-opens=java.base/com.sun.net.ssl.internal.ssl=ALL-UNNAMED");
         modularJavaOpts.add("--add-opens=java.base/java.lang=ALL-UNNAMED");
         modularJavaOpts.add("--add-opens=java.base/java.lang.invoke=ALL-UNNAMED");
         modularJavaOpts.add("--add-opens=java.base/java.lang.reflect=ALL-UNNAMED");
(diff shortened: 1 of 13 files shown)


bstansberry (changes requested):
@ropalka Please update this file as well:

https://github.com/wildfly/wildfly-core/blob/main/bootable-jar/boot/pom.xml#L61

fjuma:
@ropalka Just FYI, when that method doesn't exist, a different approach is used to detect if FIPS is enabled, see:

https://issues.redhat.com/browse/WFCORE-5566

and

https://github.com/wildfly/wildfly-core/blob/a19937eea5a4a51cd3a51cb0a034716c17f87944/elytron/src/main/java/org/wildfly/extension/elytron/SSLDefinitions.java#L1609

ropalka:
> @ropalka Please update this file as well:
> 
> https://github.com/wildfly/wildfly-core/blob/main/bootable-jar/boot/pom.xml#L61

Fixed @bstansberry

ropalka:
> But please leave this open as it's a good place to discuss the tradeoffs.

The main problem is there is currently not available any way to detect the JVM version we are running on. Neither in shell scripts nor in Java source code. This kind of problems would be nicely solvable if we would have such possibility.

soul2zimate:
Is there any decision made on this ?

soul2zimate:
reopen this since the WFCORE issue is still open.

ropalka:
> Is there any decision made on this ?

No @soul2zimate . The problem still persists - no possibility to detect exact running JVM version in shell scripts.

jbliznak:
> no possibility to detect exact running JVM version in shell scripts

@ropalka why do we need exact version? 

Aother idea, can't we just use the exact same trick as here (simply try the problematic argument first and act accordingly later)? https://github.com/wildfly/wildfly-core/blob/main/core-feature-pack/common/src/main/resources/content/bin/common.sh#L27-L29

gaol:
@ropalka does what @jbliznak's suggestion make sense to you ? and  may I ask you to rebase  ? Thanks :)

ropalka:
@bstansberry please review

host-controller/src/main/java/org/jboss/as/host/controller/jvm/JvmType.java
```diff
@@ -108,6 +120,20 @@ public Collection<String> getDefaultArguments() {
         return isModularJvm ? DEFAULT_MODULAR_JVM_ARGUMENTS : Collections.EMPTY_LIST;
     }
 
+    public Collection<String> getOptionalDefaultArguments() {
+        if (isModularJvm) {
+            Collection<String> retVal = null;
```
bstansberry, line 125: Can you add a comment...

```
// The expectation is this method is called once per JvmType instance.
// In theory we could cache the result for reuse, but if we change how
// we use these instances it may not be the case that a cached value
// would be correct over multiple invocations.
```

bstansberry (changes requested):
Hi @ropalka 

I have a few comments; only substantial one is the timeout.

When this is ready please ping me in chat so I don't forget about it.

host-controller/src/main/java/org/jboss/as/host/controller/jvm/JvmType.java
```diff
@@ -148,6 +174,54 @@ private static boolean isModularJvm(final String javaExecutable, final boolean f
...
+        final ProcessBuilder builder = new ProcessBuilder(javaExecutable, optionalModularArgument, "-version");
+        Process process = null;
+        Path stdout = null;
+        try {
+            // Create a temporary file for stdout
+            stdout = Files.createTempFile("stdout", ".txt");
+            process = builder.redirectErrorStream(true)
+                    .redirectOutput(stdout.toFile()).start();
+
+            if (process.waitFor(1, TimeUnit.SECONDS)) {
```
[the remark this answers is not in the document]
ropalka: This method was copied from org.wildfly.core.launcher.JVM class. I adjusted both occurrences of this small value.

host-controller/src/main/java/org/jboss/as/host/controller/jvm/JvmType.java
ropalka: Added. Thanks for feedback.

launcher/src/main/java/org/wildfly/core/launcher/AbstractCommandBuilder.java
```diff
@@ -84,6 +86,12 @@ abstract class AbstractCommandBuilder<T extends AbstractCommandBuilder<T>> imple
             modularJavaOpts.add("--add-modules=java.se");
         }
         DEFAULT_MODULAR_VM_ARGUMENTS = Collections.unmodifiableList(modularJavaOpts);
+
+        final ArrayList<String> optionalModularJavaOpts = new ArrayList<>();
+        if (!Boolean.parseBoolean(System.getProperty("launcher.skip.jpms.properties", "false"))) {
```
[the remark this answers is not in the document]
ropalka: Good point.

ropalka:
Fixed @bstansberry . Thanks for review!

pom.xml
```diff
@@ -164,7 +164,7 @@
              NB: In case an update is made to these exports and opens, make sure that the common.sh script is in sync.
         -->
         <embedding.jar.jpms.exports>java.desktop/sun.awt java.naming/com.sun.jndi.ldap java.naming/com.sun.jndi.url.ldap java.naming/com.sun.jndi.url.ldaps jdk.naming.dns/com.sun.jndi.dns</embedding.jar.jpms.exports>
-        <embedding.jar.jpms.opens>java.base/java.lang java.base/java.lang.invoke java.base/java.lang.reflect java.base/java.io java.base/java.net java.base/java.security java.base/java.util java.base/java.util.concurrent java.management/javax.management java.naming/javax.naming</embedding.jar.jpms.opens>
+        <embedding.jar.jpms.opens>java.base/com.sun.net.ssl.internal.ssl java.base/java.lang java.base/java.lang.invoke java.base/java.lang.reflect java.base/java.io java.base/java.net java.base/java.security java.base/java.util java.base/java.util.concurrent java.management/javax.management java.naming/javax.naming</embedding.jar.jpms.opens>
```
bstansberry, line 167: @ropalka @jfdenise AIUI this will result in a warning if the CLI or a bootable jar is used on SE 17. So in this case it seems we are forced to make a choice, unless there's some fancy MANIFEST.mf, etc thing in SE that allows some kind of conditional behavior.

If not, my inclination is to not add this since

1) Getting the warn with SE 11 is the existing state, whereas introducing a warn for SE 17 would be a breaking change. All things equal, annoying a fresh set of users in order to stop annoying ones who may already be used to the warn seems a bad tradeoff.

2) We'll be moving on from SE 11 soon enough.

jbliznak: > unless there's some fancy MANIFEST.mf, etc thing in SE that allows some kind of conditional behavior.

it is not, but when MANIFEST.MF has add-opens/add-exports on non-existing module, it silently ignores it, so IIUC this could be fine then.

https://openjdk.org/jeps/261#Breaking-encapsulation
```
Two new JDK-specific JAR-file manifest attributes are defined to correspond to the --add-exports and --add-opens command-line options:

Add-Exports: <module>/<package>( <module>/<package>)*
Add-Opens: <module>/<package>( <module>/<package>)*
The value of each attribute is a space-separated list of slash-separated module-name/package-name pairs. A <module>/<package> pair in the value of an Add-Exports attribute has the same meaning as the command-line option --add-exports <module>/<package>=ALL-UNNAMED. A <module>/<package> pair in the value of an Add-Opens attribute has the same meaning as the command-line option --add-opens <module>/<package>=ALL-UNNAMED.

Each attribute can occur at most once, in the main section of a MANIFEST.MF file. A particular pair can be listed more than once. If a specified module was not resolved, or if a specified package does not exist, then the corresponding pair is ignored. These attributes are interpreted only in the main executable JAR file of an application, i.e., in the JAR file specified to the -jar option of the Java run-time launcher; they are ignored in all other JAR files.
```

bstansberry: Thanks, a lot @jbliznak -- that's very helpful.

bstansberry (approved):

bstansberry:
Thanks @ropalka and @jbliznak

ropalka:
@jbliznak is right. Not existenting values of add-opens or add-exports attributes in MANIFEST.MF are silently ignored for executable jar files.

ropalka:
Thanks for review and merge of this PR @bstansberry

ropalka:
I provided a reproducer (for the claim about executable jars) and attached it to the JIRA  @bstansberry

PR #4967, opened 2022-02-11
[WFCORE-5807] Update logback test dependency to 1.2.9 (CVE-2021-42550)

bstansberry:
https://issues.redhat.com/browse/WFCORE-5807

### DIFF
--- a/testsuite/embedded/pom.xml
+++ b/testsuite/embedded/pom.xml
@@ -38,7 +38,7 @@
         <jbossas.ts.dir>${jbossas.ts.integ.dir}/..</jbossas.ts.dir>
         <jbossas.project.dir>${jbossas.ts.dir}</jbossas.project.dir>
         <wildfly.home>${project.basedir}/target/wildfly-core</wildfly.home>
-        <version.ch.qos.logback>1.2.3</version.ch.qos.logback>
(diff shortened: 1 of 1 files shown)


yersan:
It looks like the time given to the HostControllerBootOperationTestCase is not yet enough. The Job execution for this test reveals that the read operations took 34 seconds (under normal circumstances in my local environment with the security manager enabled, they take no more than 10 seconds). I guess this situation is pretty dependent on the current CI load.

At the moment, the byteman rules are defined with a timeout of 25 seconds, which is not enough. We can increment them to 45s, the downside is we overall test will take more than a minimum 1m 30s.

Meanwhile I'm going to kick off this Job again

yersan:
[ExternalJMSDestinationDefinitionLegacyPrefixMessagingDeploymentTestCase.testSendMessage](https://ci.wildfly.org/viewLog.html?buildId=294884&buildTypeId=WildFlyCore_PullRequest_GalleonWildFlyCoreFullIntegrationLinuxJdk11#testNameId-8846704328679989003) is failing intermittently, at first glance, tracked by [WFLY-15881](https://issues.redhat.com/browse/WFLY-15881)

jamezp (approved):

jmesnil:
@bstansberry thanks

PR #4970, opened 2022-02-18
WFCORE-5810 Fix license in wildfly-cli jar.

RanabirChakraborty:
Issue: https://issues.redhat.com/browse/WFCORE-5810

### DIFF
--- a/core-feature-pack/common/src/main/resources/license/core-feature-pack-common-licenses.xml
+++ b/core-feature-pack/common/src/main/resources/license/core-feature-pack-common-licenses.xml
@@ -230,6 +230,11 @@
           <url>http://www.apache.org/licenses/LICENSE-2.0</url>
           <distribution>repo</distribution>
         </license>
+        <license>
+          <name>Eclipse Public License - v 1.0</name>
+          <url>https://www.eclipse.org/legal/epl-v10.html</url>
+          <distribution>repo</distribution>
+        </license>
       </licenses>
     </dependency>
     <dependency>
@@ -583,6 +588,11 @@
         <license>
           <name>Apache License 2.0</name>
           <url>http://www.apache.org/licenses/LICENSE-2.0</url>
+          <distribution>repo</distribution>
+        </license>
+        <license>
+          <name>Eclipse Public License - v 1.0</name>
+          <url>https://www.eclipse.org/legal/epl-v10.html</url>
           <distribution>repo</distribution>
         </license>
       </licenses>


RanabirChakraborty:
@jfdenise Can you please review the PR

core-feature-pack/common/src/main/resources/license/core-feature-pack-common-licenses.xml
```diff
@@ -230,6 +230,11 @@
           <url>http://www.apache.org/licenses/LICENSE-2.0</url>
           <distribution>repo</distribution>
         </license>
+        <license>
```
jfdenise, line 233: @RanabirChakraborty , the same license needs to be added to wildfly-cli artifact in the same file:
https://github.com/wildfly/wildfly-core/blob/dd456d87dcbf03d6bbc3db05d48b1d7f38e92e0f/core-feature-pack/common/src/main/resources/license/core-feature-pack-common-licenses.xml#L582

jfdenise (approved):

pom.xml
```diff
@@ -51,6 +51,11 @@
             <url>http://repository.jboss.org/licenses/apache-2.0.txt</url>
             <distribution>repo</distribution>
         </license>
+        <license>
+            <name>Eclipse Public License - v 1.0</name>
+            <url>https://www.eclipse.org/legal/epl-v10.html</url>
+            <distribution>repo</distribution>
+        </license>
     </licenses>
```
yersan, line 59: I think the licenses on the main pom.xml file are the wildfly-core license and not an aggregation of the licenses used by the modules shipped with it. "Eclipse Public License" should be removed from the main pom.xml

yersan (changes requested):

pom.xml
yersan: Sorry, I should have read the Jira description first. It seems pretty clear we need it here, so, ignore my comment

bstansberry: It's not clear to me, so I've put the 'hold' label on this while I get expert advice. :)

bstansberry: @RanabirChakraborty The feedback I've gotten is this is not required, so please remove it. Thanks for your patience.

bstansberry (changes requested):
I've removed the 'hold' label from this but would want to see the change to the root pom removed.

pom.xml
RanabirChakraborty: Thanks! @bstansberry did the changes :)

yersan (approved):
As per discussions, this looks good to me

jfdenise (approved):

bstansberry (approved):

PR #4978, opened 2022-02-23
[yaml-fix]: Fixing YAML delete operation.

ehsavoie:
* Instead of appending a detele operation we are deleting all the


ehsavoie:
Superseded by https://github.com/wildfly/wildfly-core/pull/4979

PR #4979, opened 2022-02-23
[WFCORE-5818]:  !remove might not work properly with some resources

ehsavoie:
*  Fixing YAML remove operation: instead of appending a detele operation we are deleting all the

Jira: https://issues.redhat.com/browse/WFCORE-5818

### DIFF
--- a/controller/src/main/java/org/jboss/as/controller/ModelControllerImpl.java
+++ b/controller/src/main/java/org/jboss/as/controller/ModelControllerImpl.java
@@ -544,7 +544,16 @@ class ModelControllerImpl implements ModelController {
                     // stop
                     break;
                 } else {
-                    postExtContext.addBootStep(parsedOp);
+                    if(parsedOp.handler instanceof ParallelBootOperationStepHandler &&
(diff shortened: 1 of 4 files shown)


controller/src/main/java/org/jboss/as/controller/ModelControllerImpl.java
```diff
@@ -544,7 +544,16 @@ boolean boot(final List<ModelNode> bootList, final OperationMessageHandler handl
                     // stop
                     break;
                 } else {
-                    postExtContext.addBootStep(parsedOp);
+                    if(parsedOp.handler instanceof ParallelBootOperationStepHandler &&
```
jmesnil, line 547: I'm not sure to understand the condition that triggers this.
This does not apply only to the `!remove` operations but to any operations, right?

ehsavoie: What I meant is that currently ops will be added to the global ops list. 
With delete I'm removing the add/update ops from the list of ops but I also process child ops? It is those child ops that while being updated are not used by the ParallelBootOperationStepHandler which will only use its initial list

jmesnil (approved):
Thanks for the explanation.

Could you also please backport it to the `18.x` branch?

PR #4982, opened 2022-02-28
[WFCORE-5818]:  !remove might not work properly with some resources

ehsavoie:
*  Fixing YAML remove operation: instead of appending a detele operation we are deleting all the

Jira: https://issues.redhat.com/browse/WFCORE-5818


ehsavoie:
Backport of https://github.com/wildfly/wildfly-core/pull/4979

jmesnil (approved):

PR #5002, opened 2022-03-10
WFCORE-5831 Duplicate FastCopyHashMap classes

clara0:
https://issues.redhat.com/browse/WFCORE-5831


bstansberry:
Hi @clara0 -- we should not make this class public.  This class is not part of the API the controller module should be exposing. If there is no general purpose utility library that we're already using that has the equivalent class, then we'll need to continue to duplicate it.

clara0:
Ok, thank you for clarifying!

PR #5003, opened 2022-03-11
[18.x][WFCORE-5798] Restore ReloadRedirectTestCase.testReloadwithRedirect

yersan:
WFCORE-5798 backporting to 18.x. This issue enables a new test, so it makes sense to me to enable it on 18.x as well.

upstream: https://github.com/wildfly/wildfly-core/pull/4955
Jira issue: https://issues.redhat.com/browse/WFCORE-5798


bstansberry (approved):
I'm approving this even though the original commit is mine, because Yeray's port is essentially an approval and I have no concerns about the suitability of this for 18.x. If I'd had time to port it and he approved my port, the effect would be the same.

PR #5005, opened 2022-03-14
[HOLD] Fix for WFCORE-5835, Missing modular export when enabling dns.ping protocol

jfdenise:
Issue: https://issues.redhat.com/browse/WFCORE-5835
This PR is on hold, do not merge. Fix could be applied differently.

### DIFF
--- a/bootable-jar/boot/pom.xml
+++ b/bootable-jar/boot/pom.xml
@@ -57,7 +57,7 @@
                         <manifestEntries>
                             <Multi-Release>true</Multi-Release>
                             <!-- NB: In case an update is made to these exports and opens, make sure that common.sh script is in sync. -->
-                            <Add-Exports>java.desktop/sun.awt java.naming/com.sun.jndi.ldap</Add-Exports>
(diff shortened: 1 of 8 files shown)


jfdenise:
@ropalka , if you could give an eye. Once merged it would conflict with the PR you opened: https://github.com/wildfly/wildfly-core/pull/4966

ropalka (approved):
LGTM @jfdenise , just FYI my PR you mentioned above is not going to be applied soon. It turned into discussion PR for now.

jfdenise:
@ropalka , thank-you. I put this PR on hold for now, there are some discussions for this one too.

jfdenise:
@bstansberry @pferraro should we still hold this PR? Thank-you.

bstansberry:
@jfdenise -- @pferraro can confirm but IIRC this what not a change we wanted to make outside of the cloud context.

jfdenise:
@bstansberry , Ok, that is something that we can handle in the cloud if @pferraro confirms.

bstansberry:
BTW, my concern about this was the reference to the jdk.* module and the impact that might have on users who want to run with a slimmed JVM. Unfortunately you can't add JPMS settings that refer to non-existent modules, so it's not safe to just add stuff.

jfdenise:
Closing, will be handled at the cloud level.

PR #5007, opened 2022-03-15
[WFCORE-5837] AttributeMarshallers.SimpleListAttributeMarshaller needs to flag isMarshallableAsElement as true

gaol:
Issue: https://issues.redhat.com/browse/WFCORE-5837

This PR tries to:
* Implement `AttributeMarshallers.SimpleListAttributeMarshaller.isMarshallableAsElement()` to return `true`.
* Fixes the argument of calling sub element of the list to the `valueType` instead of the `SimpleListAttribute` itself.
* Adds a `SIMPLE_ELEMENT_UNWRAP` in `AttributeMarshallers` to render the simple element from resourceModel.
* Adds tests.


gaol:
@bstansberry would you please review ? thanks

PR #5009, opened 2022-03-15
[WFCORE-5834]: The deployent-scanner needs org.wildfly.common

ehsavoie:
* adding 'org.wildfly.common' to the deployment scanner module.xml

Jira: https://issues.redhat.com/browse/WFCORE-5834


jmesnil:
@ehsavoie thanks

PR #5011, opened 2022-03-15
[WFCORE-5849] Null pointers should not be dereferenced (cli)

boris-unckel:
Fixes https://issues.redhat.com/browse/WFCORE-5849


cli/src/main/java/org/jboss/as/cli/impl/aesh/cmd/security/model/ElytronUtil.java
```diff
@@ -488,6 +490,7 @@ public static ServerSSLContext getServerSSLContext(CommandContext context, Strin
         if (sslContext != null) {
             String kmName = sslContext.get(Util.KEY_MANAGER).asString();
             ModelNode km = getChildResource(kmName, Util.KEY_MANAGER, context);
+            checkNotNullParamWithNullPointerException("ModelNode for " + kmName, km);
```
yersan, line 493: The `"ModelNode for " + kmName` text ends up in the [following log](https://github.com/wildfly/wildfly-common/blob/main/src/main/java/org/wildfly/common/_private/CommonMessages.java#L62)

So, this text should be a parameter name instead. Notice that messages of the loggers are translated into different languages. What about Util.KEY_MANAGER?

The other points look fine to me, CC @jfdenise

yersan (changes requested):

cli/src/main/java/org/jboss/as/cli/impl/aesh/cmd/security/model/ElytronUtil.java
boris-unckel: I would suggest to use a NullPointerException (compatible Exception, same as before, else a IllegalStateException) with a custom text here, because Util.KEY_MANAGER (kmName) itself is not null, but the childResource.

yersan: That sounds like an alternative, although I cannot suggest to you how you can internationalize it, I searched for a valid logger but I'm not sure how this is treated on CLI. In general, if a message arrives at the user, it should be internationalized.

boris-unckel: @yersan The method is not in use currently. Several other exceptions use hardcoded English texts (see some lines above), I have prepared the following:
```
            if (km == null) {
                throw new NullPointerException("The ModelNode of '" + kmName + "' must not be null");
            }
```

yersan (approved):

cli/src/main/java/org/jboss/as/cli/impl/aesh/cmd/security/model/ElytronUtil.java
yersan: ok @boris-unckel , thanks, I've approved it from my side, but this should be approved by the component lead, @jfdenise FYI, not sure if you are interested in adding these changes.

cli/src/main/java/org/jboss/as/cli/impl/aesh/cmd/security/model/ElytronUtil.java
```diff
@@ -488,6 +488,9 @@ public static ServerSSLContext getServerSSLContext(CommandContext context, Strin
         if (sslContext != null) {
             String kmName = sslContext.get(Util.KEY_MANAGER).asString();
             ModelNode km = getChildResource(kmName, Util.KEY_MANAGER, context);
+            if (km == null) {
+                throw new NullPointerException("The ServerSSLContext " + sslContextName + " references the KeyManager " + kmName + " that doesn't exist.");
```
jfdenise, line 492: I am wandering if a IllegalStateException would be a better fit than NPE?

boris-unckel: @jfdenise In earlier fixings regarding NullPointers it was a requirement to have the same Exception as before (IllegalArgumentEx vs NullPointerEx). That was the reason to choose NPE here, too. My point of the fix is to have an "fail early" and a clear message, but I don't have a preference for the type of the exception.

jfdenise: I would vote for IllegalArhumentEx, NPE seems strange to me.

boris-unckel: @jfdenise thanks for your feedback. I'm going to provide an update when the CI is re-opened for everything in Zulip.

boris-unckel: @jfdenise The changes requested are present now. Could you check and approve, please?

jfdenise (approved):

PR #5012, opened 2022-03-15
[WFCORE-5850] Null pointers should not be dereferenced (controller)

boris-unckel:
Fixes https://issues.redhat.com/browse/WFCORE-5850

### DIFF
--- a/controller/src/main/java/org/jboss/as/controller/OperationContextImpl.java
+++ b/controller/src/main/java/org/jboss/as/controller/OperationContextImpl.java
@@ -1029,6 +1029,9 @@ final class OperationContextImpl extends AbstractOperationContext {
 
     private Resource createResourceInternal(PathAddress relativeAddress, int index) throws UnsupportedOperationException {
         ImmutableManagementResourceRegistration current = getResourceRegistration();
+        if (current == null) {
+            throw MGMT_OP_LOGGER.currentResourceRegistrationIsNull();
+        }
         ImmutableManagementResourceRegistration mrr = relativeAddress == PathAddress.EMPTY_ADDRESS ? current : current.getSubModel(relativeAddress);
         final Resource toAdd = Resource.Factory.create(mrr.isRuntimeOnly());
         addResourceInternal(relativeAddress, index, toAdd);
--- a/controller/src/main/java/org/jboss/as/controller/operations/validation/OperationValidator.java
+++ b/controller/src/main/java/org/jboss/as/controller/operations/validation/OperationValidator.java
@@ -130,6 +130,7 @@ public class OperationValidator {
         OperationEntry entry = root.getOperationEntry(address, name);
         if (entry == null) {
             throwOrWarnAboutDescriptorProblem(ControllerLogger.ROOT_LOGGER.noOperationEntry(name, address));
+            return; // in case of warning no further processing possible (entry causes NullPointer)
         }
         //noinspection ConstantConditions
         if (entry.getType() == EntryType.PRIVATE || entry.getFlags().contains(OperationEntry.Flag.HIDDEN)) {
--- a/controller/src/main/java/org/jboss/as/controller/persistence/yaml/YamlConfigurationExtension.java
+++ b/controller/src/main/java/org/jboss/as/controller/persistence/yaml/YamlConfigurationExtension.java
@@ -498,7 +498,9 @@ public class YamlConfigurationExtension implements ConfigurationExtension {
                             if (map != null && map.containsKey("index")) {
                                 op.get("index").set((Integer) map.get("index"));
                             }
-                            op.get(VALUE).set(processObjectAttribute((ObjectTypeAttributeDefinition) type, map));
+                            if (map != null) {
+                                op.get(VALUE).set(processObjectAttribute((ObjectTypeAttributeDefinition) type, map));
+                            }
                             break;
                         case LIST:
                             if (entry instanceof Map) {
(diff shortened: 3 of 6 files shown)


controller/src/main/java/org/jboss/as/controller/registry/OperationTransformerRegistry.java
```diff
@@ -355,7 +355,10 @@ private OperationTransformerRegistry createChild(Iterator<PathElement> iterator,
         }
 
         private void registerTransformer(Iterator<PathElement> iterator, String value, String operationName,  OperationTransformerEntry entry) {
-            get(value).registerTransformer(iterator, operationName, entry);
+            OperationTransformerRegistry operationTransformerRegistry = get(value);
+            if (operationTransformerRegistry != null) {
+                operationTransformerRegistry.registerTransformer(iterator, operationName, entry);
+            }
```
yersan, line 361: I've reviewed the changes but I don't feel capable to approve them as they are. In general, they look harmless; we can argue that if the CI is passing, then we are fine, but there are possibilities for introducing an unexpected bug or a side effect.

I think we should be more cautious with those changes.

Specifically, this one, if `operationTransformerRegistry` is null the old behavior was to get a null pointer exception, but now the method is silently invoked without registering the expected transformer and that can be a wrong result difficult to track and debug on a mixed domain scenario. Maybe it could be better to keep the NPE or throw an IAE if it is considered an invalid state. This argument can be also extrapolated to [YamlConfigurationExtension.java](https://github.com/wildfly/wildfly-core/pull/5012/files#diff-d50b9b2a6b6cbf891246ead8a48357b7fdd24ef0ba76f35bda6e992afb26c742), [OperationValidator.java](https://github.com/wildfly/wildfly-core/pull/5012/files#diff-ef60c464eae6dddb2e6505f8c47b941da01873c707d5ab547349340abc40b647), in which case leaving the code as it is now is also a valid alternative.

controller/src/main/java/org/jboss/as/controller/persistence/yaml/YamlConfigurationExtension.java
```diff
@@ -498,7 +498,9 @@ public void processOperation(ImmutableManagementResourceRegistration rootRegistr
                             if (map != null && map.containsKey("index")) {
                                 op.get("index").set((Integer) map.get("index"));
                             }
-                            op.get(VALUE).set(processObjectAttribute((ObjectTypeAttributeDefinition) type, map));
+                            if (map != null) {
+                                op.get(VALUE).set(processObjectAttribute((ObjectTypeAttributeDefinition) type, map));
```
boris-unckel, line 502: @ehsavoie Could you please have a look at this fix? All other calls prevent a null pointer deref, except this one. Is it intended to break here?

controller/src/main/java/org/jboss/as/controller/operations/validation/OperationValidator.java
```diff
@@ -130,6 +130,7 @@ public void validateOperation(final ModelNode operation) {
         OperationEntry entry = root.getOperationEntry(address, name);
         if (entry == null) {
             throwOrWarnAboutDescriptorProblem(ControllerLogger.ROOT_LOGGER.noOperationEntry(name, address));
+            return; // in case of warning no further processing possible (entry causes NullPointer)
```
boris-unckel, line 133: @kabir Could you have a look at this fix, please? Many other points in the code have the pattern
```
throwOrWarnAboutDescriptorProblem(ControllerLogger....);
return;
```
Here the `return;` is missing. Is it intended to break later with an NPE in case of logging a warning?

controller/src/main/java/org/jboss/as/controller/persistence/yaml/YamlConfigurationExtension.java
ehsavoie: I think the null check used to be before the check from the index

controller/src/main/java/org/jboss/as/controller/operations/validation/OperationValidator.java
kabir: @boris-unckel I think it looks fine. A lot of this is cruft from ancient times before we became a lot stricter about needing definitions for everything

boris-unckel:
@yersan The changes requested are present or clarified with the original authors. Test failures seem unrelated. Could you have a final look at this, please?

controller/src/main/java/org/jboss/as/controller/OperationContextImpl.java
```diff
@@ -1028,7 +1029,7 @@ public Resource createResource(PathAddress relativeAddress) {
     }
 
     private Resource createResourceInternal(PathAddress relativeAddress, int index) throws UnsupportedOperationException {
-        ImmutableManagementResourceRegistration current = getResourceRegistration();
+        ImmutableManagementResourceRegistration current = checkNotNullParamWithNullPointerException("getResourceRegistration", getResourceRegistration());
```
yersan, line 1032: Sorry for the back-and-forth @boris-unckel , I missed this one; the purpose of `checkNotNullParamWithNullPointerException` is to check method params, so I would say we should not use it for any kind of situation. If you still want to avoid deferring the NPE here, we could use a specific logger entry for this one.

controller/src/main/java/org/jboss/as/controller/persistence/yaml/YamlConfigurationExtension.java
yersan: @ehsavoie I'm still in doubt, are you fine with this change then?

controller/src/main/java/org/jboss/as/controller/OperationContextImpl.java
boris-unckel: @yersan NP, thanks for the additional feedback. I have changed accordingly.

PR #5017, opened 2022-03-16
WFCORE-2852/WFCORE-5853: CLI autocomplete suggests non-existent subsystems

ehsavoie:
* Removing the child names for runtime resource definitions when the
  resource is not present for completion.
* Setting the logging deployment submodel to runtime

Jira: https://issues.redhat.com/browse/WFCORE-5852
        https://issues.redhat.com/browse/WFCORE-5853


yersan:
@ehsavoie This is now unblocked, would you mind describing better what did you mean by adding a flag?

ehsavoie:
@yersan I meant adding a boolean parameter like we have for include-runtime / alias etc.

yersan (approved):
Since the WildFly counterpart was merged by using this approach we should be fine with it and we don't need to use a parameter to deal with these resources.

So giving my approval, although we would need a test case covering it.

PR #4984, opened 2022-03-01
[18.x][WFCORE-5820] Add version 15.1 of the WildFly Elytron model and schema

fjuma:
https://issues.redhat.com/browse/WFCORE-5820
Upstream: https://github.com/wildfly/wildfly-core/pull/4980

PR #5020, opened 2022-03-17
[18.x][WFCORE-5490] Integrate Elytron credential expressions into system-property resource handling

bstansberry:
Issue: https://issues.redhat.com/browse/WFCORE-5490
Upstream: https://github.com/wildfly/wildfly-core/pull/4925


jmesnil:
@bstansberry this needs to be rebased as there is a conflict after https://github.com/wildfly/wildfly-core/pull/4984 was merged.

PR #5028, opened 2022-03-24
[WFCORE-5861] Add the standard server JPMS settings to jboss-cli-clie…

bstansberry:
…nt.jar so it can embed the server on SE 16+

Configure the settings in properties in the root pom so they can be shared between the bootable-jar/boot/pom.xml and the cli/pom.xml

https://issues.redhat.com/browse/WFCORE-5861

### DIFF
--- a/bootable-jar/boot/pom.xml
+++ b/bootable-jar/boot/pom.xml
@@ -56,9 +56,9 @@
                         </manifest>
                         <manifestEntries>
                             <Multi-Release>true</Multi-Release>
-                            <!-- NB: In case an update is made to these exports and opens, make sure that common.sh script is in sync. -->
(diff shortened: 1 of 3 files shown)


yersan:
Linux - Security Manager failure relates to https://github.com/wildfly/wildfly-core/pull/5032

yersan:
Hi @tomjenkinson I've just kicked off the failed job again. It must pass, the failure was already identified and it's so far under control; it happens intermittently when the CI is very busy. It's completely unrelated to this PR.

bstansberry:
@marekkopecky The use case for jboss-client.jar is different. Unlike jboss-cli-client.jar it is not intended to be the argument to java -jar, and these MANIFEST.mf settings are used by the JVM if they are present in the jar with the Main class.

PR #4961, opened 2022-02-08
[WFCORE-5805] Fix for HostControllerBootOperationsTestCase intermiten…

yersan:
…t failures

- Incremented the timeouts of the byteman rules
- Fix operations that ensure the test conditions are satisfied
- In addition, to help out with debugging, added traces to get the duration of the read operations

Jira issue: https://issues.redhat.com/browse/WFCORE-5805

PR #5032, opened 2022-03-25
[WFCORE-5805] HostControllerBootOperationsTestCase fails intermittent…

yersan:
…ly when Security Manager is enabled

Jira issue: https://issues.redhat.com/browse/WFCORE-5805

This PR follow-ups on https://github.com/wildfly/wildfly-core/pull/4961. 

It adds 40 seconds more to the overall test to ensure we are not hitting failures when CI is pretty busy :(. Ideally, we should be able to unlock the Byte man rule by issuing `signalWake` when the test condition is satisfied, but it is really difficult to find a condition that can be used to trigger the awake by using the same identifier ($this). So I'm trying this with the hope that this additional time is acceptable. It is a workaround though, it won't ensure we will never going to pass the limit that ensures the test can be run, but for the last failures, it looks enough.


bstansberry:
Thanks, @yersan

PR #5042, opened 2022-04-01
[WFCORE-5871] Replace invalid operator

rosiecki:
Invalid operator while comparing variable SECURITY_MANAGER_SET_TO_ALLOW to "x" caused bash message "unexpected operator".

Jira issue: https://issues.redhat.com/browse/WFCORE-5871

### DIFF
--- a/core-feature-pack/common/src/main/resources/content/bin/standalone.sh
+++ b/core-feature-pack/common/src/main/resources/content/bin/standalone.sh
@@ -309,7 +309,7 @@ fi
 SECURITY_MANAGER_SET=`echo $JAVA_OPTS | $GREP "java\.security\.manager"`
 if [ "x$SECURITY_MANAGER_SET" != "x" ]; then
     SECURITY_MANAGER_SET_TO_ALLOW=`echo $JAVA_OPTS | $GREP "java\.security\.manager=allow"`
-    if [ "x$SECURITY_MANAGER_SET_TO_ALLOW" == "x" ]; then
+    if [ "x$SECURITY_MANAGER_SET_TO_ALLOW" = "x" ]; then
         echo "ERROR: The use of -Djava.security.manager has been removed. Please use the -secmgr command line argument or SECMGR=true environment variable."
         exit 1
     fi


core-feature-pack/common/src/main/resources/content/bin/standalone.sh
```diff
@@ -309,7 +309,7 @@ fi
 SECURITY_MANAGER_SET=`echo $JAVA_OPTS | $GREP "java\.security\.manager"`
 if [ "x$SECURITY_MANAGER_SET" != "x" ]; then
     SECURITY_MANAGER_SET_TO_ALLOW=`echo $JAVA_OPTS | $GREP "java\.security\.manager=allow"`
-    if [ "x$SECURITY_MANAGER_SET_TO_ALLOW" == "x" ]; then
+    if [ "x$SECURITY_MANAGER_SET_TO_ALLOW" = "x" ]; then
```
yersan, line 312: Hi @rosiecki thanks for the contribution but this is a valid string comparison for any POSIX shell; you will also see it used widely in this script. 

In any case, just for your info, if you detect a bug the preferred approach is to report it by creating a Jira describing how to reproduce it.

I'm going to close this pull request, but please, feel free to comment here if you have any questions or notes about this. Thanks!

rosiecki: > you will also see it used widely in this script


let me check:

``
fgrep "==" wildfly-26.1.0.Beta1/bin/standalone.sh
``

the result is only one:

``
    if [ "x$SECURITY_MANAGER_SET_TO_ALLOW" == "x" ]; then
echo "========================================================================="
echo "========================================================================="
``

yersan: Ups, I'm sorry @rosiecki , I got confused with the change, you are removing the double == which should be fine since the script uses a single = in general. 

I'm curious since == and = are both valid operators for a POSIX compatible shell, could you describe on which OS are you hitting the issue?

rosiecki: I found the issue on Ubuntu 18.04.5 LTS.

yersan: @rosiecki thanks, now I got it, and == is not a POSIX compatible operator that causes; I made two mistakes reviewing this.

I've created this Jira to track it: https://issues.redhat.com/browse/WFCORE-5871

Would you mind changing your commit message to include the Jira number? See https://docs.wildfly.org/26/Hacking_On_WildFly.html#contributing_commit_push for getting more info.

yersan: In addition to the above and due to the changes you are doing here, we require that all contributions be made under the terms of the MIT License, http://www.opensource.org/licenses/mit-license.php. Do you agree to these terms?

rosiecki: > In addition to the above and due to the changes you are doing here, we require that all contributions be made under the terms of the MIT License, http://www.opensource.org/licenses/mit-license.php. Do you agree to these terms?

I agree with the terms.

yersan: Thanks, @rosiecki, we need you to follow https://docs.wildfly.org/26/Hacking_On_WildFly.html#contributing_commit_push instructions and change the commit message to include the Jira issue, after that, we will enable CI on this PR, and then we will ready for merging it :)

yersan:
Closed, superseded by https://github.com/wildfly/wildfly-core/pull/5047

PR #5046, opened 2022-04-05
[WFCORE-5839][primary/secondary] Find and replace occurrences used by…

moulalis:
Issue: https://issues.redhat.com/browse/WFCORE-5839


yersan (changes requested):
@moulalis I would say you also have to include here the changes of this file: https://github.com/wildfly/wildfly-core/blob/a3309bd1ce8b267fcef965c69b4ec2412f2e0fda/domain-management/src/main/resources/org/jboss/as/domain/management/_private/LocalDescriptions.properties

@bstansberry allow me to defer the review of this one to you.

bstansberry (approved):

bstansberry:
Thanks @moulalis

PR #5050, opened 2022-04-08
[WFCORE-5861] Fix InaccessibleObjectException with OpenJDK17 and embed-server

yersan:
It builds on top of https://github.com/wildfly/wildfly-core/pull/5028,  it adds a test case to verify it.

Jira issue: https://issues.redhat.com/browse/WFCORE-5861

### DIFF
--- a/bootable-jar/boot/pom.xml
+++ b/bootable-jar/boot/pom.xml
@@ -56,9 +56,9 @@
                         </manifest>
                         <manifestEntries>
                             <Multi-Release>true</Multi-Release>
-                            <!-- NB: In case an update is made to these exports and opens, make sure that common.sh script is in sync. -->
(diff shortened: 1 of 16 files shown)


bstansberry:
@yersan We decided to go ahead and merge #5028 for now using manual testing. But we still want this test. Please check with @xstefank as his plan is to file a separate JIRA to track the test task, so the commit message / PR title here will need to be updated.

yersan:
Tests pass locally by using the default maven local repository, however, Team City CI explicitly uses the non-default maven local repo. That's an issue when we are using the jboss-client.jar; we need to specify the location of the maven repo to be able to run the embedded server.

Closing this in favor of https://issues.redhat.com/browse/WFCORE-5873

PR #5051, opened 2022-04-14
[WFCORE-5874] Upgrade bouncycastle to 1.71

boris-unckel:
Fixes https://issues.redhat.com/browse/WFCORE-5874

### DIFF
--- a/core-feature-pack/common/pom.xml
+++ b/core-feature-pack/common/pom.xml
@@ -248,22 +248,22 @@
 
         <dependency>
             <groupId>org.bouncycastle</groupId>
-            <artifactId>bcpg-jdk15on</artifactId>
+            <artifactId>bcpg-jdk18on</artifactId>
         </dependency>
 
         <dependency>
             <groupId>org.bouncycastle</groupId>
-            <artifactId>bcpkix-jdk15on</artifactId>
+            <artifactId>bcpkix-jdk18on</artifactId>
         </dependency>
 
         <dependency>
             <groupId>org.bouncycastle</groupId>
-            <artifactId>bcprov-jdk15on</artifactId>
+            <artifactId>bcprov-jdk18on</artifactId>
         </dependency>
 
         <dependency>
             <groupId>org.bouncycastle</groupId>
-            <artifactId>bcutil-jdk15on</artifactId>
+            <artifactId>bcutil-jdk18on</artifactId>
          </dependency>
 
         <dependency>
--- a/core-feature-pack/common/src/main/resources/license/core-feature-pack-common-licenses.xml
+++ b/core-feature-pack/common/src/main/resources/license/core-feature-pack-common-licenses.xml
@@ -157,7 +157,7 @@
     </dependency>
     <dependency>
       <groupId>org.bouncycastle</groupId>
-      <artifactId>bcpg-jdk15on</artifactId>
+      <artifactId>bcpg-jdk18on</artifactId>
       <licenses>
         <license>
           <name>MIT License</name>
@@ -168,7 +168,7 @@
     </dependency>
     <dependency>
       <groupId>org.bouncycastle</groupId>
-      <artifactId>bcpkix-jdk15on</artifactId>
+      <artifactId>bcpkix-jdk18on</artifactId>
       <licenses>
         <license>
           <name>MIT License</name>
@@ -179,7 +179,7 @@
     </dependency>
     <dependency>
       <groupId>org.bouncycastle</groupId>
-      <artifactId>bcprov-jdk15on</artifactId>
+      <artifactId>bcprov-jdk18on</artifactId>
       <licenses>
         <license>
           <name>MIT License</name>
@@ -190,7 +190,7 @@
     </dependency>
     <dependency>
       <groupId>org.bouncycastle</groupId>
-      <artifactId>bcutil-jdk15on</artifactId>
+      <artifactId>bcutil-jdk18on</artifactId>
       <licenses>
         <license>
           <name>MIT License</name>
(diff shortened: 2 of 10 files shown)


boris-unckel:
@yersan @fjuma I have prepared the component upgrade of bouncycastle to 1.71. It includes a maven coordinate change, caused by the JDK compatibility goals upstream. Please find all the links in the Jira. Obviously full integration tests fail because https://github.com/wildfly/wildfly/pull/15429 is missing (which will fail due to the lack of the PR here).

Please review and advise how to deal with it.

boris-unckel:
One more thing: If other third party libs have exclusions I did not touch them. If the third party upstream changes their bouncycastle deps this has to be changed here, too. Do we need a comment somewhere, or a Jira pointing at it?

yersan (changes requested):
Hi @boris-unckel , thanks for looking at this.

Due to the uncertainty about what could be brought in by the third parties and when I think it is fine to leave them as they are. When the third party is upgraded, that change should have to deal with the exclusions if required.

WildFly Core specifies the BouncyCastle versions to use, so wildfly-core will not be affected by any transitive dependency, it will use always the explicit ones added on the main pom.xml. However, there is an artifactId renaming, so I think we should add the old ones as banned dependencies to prevent pollution. We could add them here: https://github.com/wildfly/wildfly-core/blob/9864f64ee32a7472c8f21db30daa8f0c9cb7f386/pom.xml#L412

There are rules that ban any transitive dependency coming from wildfly-core when we are building wildfly full, so I believe there won't be dependencies leaked there.

I have doubts whether it will be necessary to also ban the old ones in wildfly full as well, I don't think so because, at the moment, all the BC deps are handled in wildfly by using the ones brought in by wildfly-core. Any comment in that regard is more than welcome.

We need to run integration jobs between the wildfly-core and wildfly topic branches to ensure both test suites pass. I can add them once the PR is in place with the banned deps.

This is going to be problematic as well, we have sweeping it under a carpet and it is going to block this upgrade as well: https://github.com/wildfly/wildfly/blob/1b01f6efd4f69660d5c9b75838b1e9c3e1c66477/testsuite/integration/elytron/pom.xml#L55-L57

boris-unckel:
> > This is going to be problematic as well, we have sweeping it under a carpet and it is going to block this upgrade as well: https://github.com/wildfly/wildfly/blob/1b01f6efd4f69660d5c9b75838b1e9c3e1c66477/testsuite/integration/elytron/pom.xml#L55-L57
> 
> Well, looking more closely, these dependencies are also `-jdk18on` variant, so they should be fine.

@yersan Thanks for your feedback. I'm preparing the PR with the suggested Maven bans. The wildfly change will fail, I have to change it. The new coordinates are not offered for old versions. We cannot leave the pinning on 1.69. For now we can only `@Ignore` the relevant test and remove the maven 1.69 version property. @fjuma is this ok for you, too?

fjuma:
@boris-unckel Just took a quick look at some of the discussion above. Can you explain why we cannot leave the pinning on 1.69 for the Elytron tests? Sorry, I probably missed something above.

boris-unckel:
> I have doubts whether it will be necessary to also ban the old ones in wildfly full as well, I don't think so because, at the moment, all the BC deps are handled in wildfly by using the ones brought in by wildfly-core. Any comment in that regard is more than welcome.

@yersan I think we have to double the bans. The log dependencies are managed by WildFly-Core, but the banned excludes are present in both.

boris-unckel:
> @boris-unckel Just took a quick look at some of the discussion above. Can you explain why we cannot leave the pinning on 1.69 for the Elytron tests? Sorry, I probably missed something above.

@fjuma At the moment we have only one point in WildFly-Core to manage the current `org.bouncycastle:*-jdk15on` dependencies. If we keep the version pinned, the new Maven coordinate (`org.bouncycastle:*-jdk18on`) will be imported from WildFly-Core and have the old 1.69 version assigend by overwriting (the pinned property) - but this combination does not exist: `org.bouncycastle:*-jdk18on:jar:1.69:compile` Sample: https://mvnrepository.com/artifact/org.bouncycastle/bcprov-jdk15on vs https://mvnrepository.com/artifact/org.bouncycastle/bcprov-jdk18on

fjuma:
Could we add test dependencies for the appropriate `org.bouncycastle:*-jdk15on` versions?

boris-unckel:
> Could we add test dependencies for the appropriate `org.bouncycastle:*-jdk15on` versions?

Yes, that should be possible. To sum up the necessary WildFly changes to the PR:
* Double the bans
* Expand the version pinning to full test dependency declarations

boris-unckel:
@yersan @fjuma Now both the WildFly-Core and WildFly PR are updated. A local snapshot integration test worked.

yersan:
Integration Jobs:

https://ci.wildfly.org/viewLog.html?buildId=300951
https://ci.wildfly.org/viewLog.html?buildId=300953

yersan (commented):
Integration Jobs looks fine.

@jamezp Could you approve the upgrade in relation to RESTEasy?
@jimma @rsearls Could you approve the upgrade in relation to jbossws/cxf?
@fjuma Could you approve the upgrade in relation to the security component?

Thanks all!

jamezp (approved):
Approved for RESTEasy.

fjuma (approved):
Approved for security.

jimma:
This looks good to webservice.

yersan (approved):

yersan:
Integration Jobs showed there were no related test failures.

@jmesnil / @darranl , this PR will require some coordination to be integrated on wildfly.

PR #5052, opened 2022-04-15
[WFCORE-5875] Migrated the log manager off javax.json.api to the internal module.

jamezp:
Includes the removal of Apache Geronimo JSON and Apache Johnzon. These do not seem to work well because the use the TCCL quite heavily requiring the `internal.javax.json.api.ee8` to be forced to be leaked. Using the `org.glassfish:jakarta.json` implementation makes this easier as it includes both the API and implementation in the artifact. In WildFly we will be migrating to Eclipse Parsson for the implementation so this will not be a conflict.

* https://issues.redhat.com/browse/WFCORE-5875
* https://issues.redhat.com/browse/WFCORE-5876


jamezp:
This passes for me locally. The failures seem to be an issue with TeamCity that @luck3y is looking at.

luck3y:
@jamezp I still need to update these jobs, should be done later today.

yersan (approved):

PR #5054, opened 2022-04-18
[WFCORE-5873] Add tests for WFCORE-5861 InaccessibleObjectException w…

yersan:
…ith OpenJDK17 and embed-server

Jira issue: https://issues.redhat.com/browse/WFCORE-5873

### DIFF
--- a/testsuite/manualmode/src/test/java/org/jboss/as/test/manualmode/management/cli/JbossCliClientJarJPMSSettingsTestCase.java
+++ b/testsuite/manualmode/src/test/java/org/jboss/as/test/manualmode/management/cli/JbossCliClientJarJPMSSettingsTestCase.java
@@ -1,0 +1,130 @@
+/*
+ * JBoss, Home of Professional Open Source.
+ * Copyright 2022 Red Hat, Inc., and individual contributors
+ * as indicated by the @author tags.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.jboss.as.test.manualmode.management.cli;
+
+import java.io.File;
+import java.nio.file.Files;
+import java.nio.file.Path;
+import java.nio.file.Paths;
+import java.util.PropertyPermission;
+import java.util.jar.Attributes;
+import java.util.jar.JarFile;
+import java.util.jar.Manifest;
+
+import org.jboss.as.test.integration.management.cli.CliProcessWrapper;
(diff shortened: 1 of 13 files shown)


yersan:
JDK-17 job: https://ci.wildfly.org/viewLog.html?buildId=300458&buildTypeId=WildFlyCore_MainLinuxJdk17

yersan:
The new JbossCliClientJarJPMSSettingsTestCase passed with JDK17 Job. This is ready for review.

xstefank:
@bstansberry, this seems to be stuck for a while. Can you please look?

testsuite/manualmode/src/test/java/org/jboss/as/test/manualmode/management/cli/JbossCliClientJarJPMSSettingsTestCase.java
```diff
@@ -0,0 +1,130 @@
...
+    @Test
+    public void testManifestEntries() throws Exception {
+        JarFile clientJar = new JarFile(SERVER_HOME.resolve("bin").resolve("client").resolve("jboss-cli-client.jar").toString());
+        Manifest manifest = clientJar.getManifest();
+        Attributes mainAttribs = manifest.getMainAttributes();
+        String addOpens = mainAttribs.getValue("Add-Opens");
+        Assert.assertTrue("Unexpected value for Manifest Add-Opens. Current: " + addOpens,
+                addOpens.contains("java.base/java.util"));
+        Assert.assertTrue("Unexpected value for Manifest Add-Opens. Current: " + addOpens,
+                addOpens.contains("java.base/java.lang.invoke"));
```
bstansberry, line 75: These can be a lot more extensive, i.e. what the jar contains now is:

Add-Exports: java.desktop/sun.awt java.naming/com.sun.jndi.ldap java.nam
 ing/com.sun.jndi.url.ldap java.naming/com.sun.jndi.url.ldaps
Add-Opens: java.base/java.lang java.base/java.lang.invoke java.base/java
 .lang.reflect java.base/java.io java.base/java.security java.base/java.
 util java.base/java.util.concurrent java.management/javax.management ja
 va.naming/javax.naming

If you pass maven prop embedding.jar.jpms.exports into surefire via -D you could parse it out and dynamically check for all of them.

bstansberry (approved):
I approved this as it good progress, but see my comment for potential further improvement.

testsuite/manualmode/src/test/java/org/jboss/as/test/manualmode/management/cli/JbossCliClientJarJPMSSettingsTestCase.java
yersan: Thanks for reviewing it @bstansberry . Got it, I'll add a follow-up Jira for your suggestion, originally it was not my intention to verify all the settings are in the Jar.

The intention here was just to check if the required settings used by the test were satisfied to ensure we are not getting a false negative. The test case is only verifying a subset of required settings by executing something that requires them.

bstansberry:
@yersan That's a good point, and leaving the two hard coded MANIFEST checks you have makes sense for that reason.

I don't think doing more is urgent. It would kind of just be a regression test in case we mess up how the manifest is built. Not a real strong one though, as the embedding.jar.jpms.exports prop is an implementation detail, so we'd be baking an assumption of how the impl works into a test.

PR #5056, opened 2022-04-19
[WFCORE-5869] standalone.bat script does not parse JAVA_OPTS containing '|' symbol properly

lvydra:
https://issues.redhat.com/browse/WFCORE-5869

### DIFF
--- a/core-feature-pack/common/src/main/resources/content/bin/standalone.bat
+++ b/core-feature-pack/common/src/main/resources/content/bin/standalone.bat
@@ -104,6 +104,22 @@ if exist "%STANDALONE_CONF%" (
    echo Config file not found "%STANDALONE_CONF%"
 )
 
+rem Sanitize JAVA_OPTS
+rem Currently escaping only | characters any other might be added if needed
+setlocal EnableDelayedExpansion
+
+rem If characters are already escaped then IS_SANITIZED is set to true and JAVA__OPTS are left as they are
+for %%C in (^^^|) do (
+  if not "!JAVA_OPTS:%%C=!"=="!JAVA_OPTS!" set IS_SANITIZED=true
+)
+if not "!IS_SANITIZED!" == "true" (
+  for %%C in (^|) do (
+    set "JAVA_OPTS=!JAVA_OPTS:%%C=^%%C!"
+  )
+)
+
+setlocal DisableDelayedExpansion
+
 if NOT "x%DEBUG_PORT%" == "x" (
   set DEBUG_PORT_VAR=%DEBUG_PORT%
 )
@@ -228,7 +244,7 @@ for %%a in (!CONSOLIDATED_OPTS!) do (
 )
 
 rem If the -Djava.security.manager is found, enable the -secmgr and include a bogus security manager for JBoss Modules to replace
-echo(!JAVA_OPTS! | findstr /r /c:"-Djava.security.manager" > nul && (
+echo("!JAVA_OPTS!" | findstr /r /c:"-Djava.security.manager" > nul && (
     echo ERROR: The use of -Djava.security.manager has been removed. Please use the -secmgr command line argument or SECMGR=true environment variable.
     GOTO :EOF
 )
--- a/testsuite/scripts/src/test/java/org/wildfly/scripts/test/StandaloneScriptTestCase.java
+++ b/testsuite/scripts/src/test/java/org/wildfly/scripts/test/StandaloneScriptTestCase.java
@@ -41,10 +41,12 @@ import org.jboss.as.test.shared.TestSuiteEnvironment;
 import org.jboss.dmr.ModelNode;
 import org.junit.Assert;
 import org.junit.Assume;
+import org.junit.Test;
 import org.junit.runner.RunWith;
 import org.junit.runners.Parameterized;
 import org.junit.runners.Parameterized.Parameter;
 import org.junit.runners.Parameterized.Parameters;
+import org.wildfly.common.test.ServerConfigurator;
 import org.wildfly.common.test.ServerHelper;
 import org.wildfly.common.test.LoggingAgent;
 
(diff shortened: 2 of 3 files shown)


core-feature-pack/common/src/main/resources/content/bin/standalone.bat
```diff
@@ -104,6 +104,22 @@ if exist "%STANDALONE_CONF%" (
    echo Config file not found "%STANDALONE_CONF%"
 )
 
+rem Sanitize JAVA_OPTS
+rem Currently escaping only | characters any other might be added if needed
+setlocal EnableDelayedExpansion
+
+rem If characters are already escaped then IS_SANITIZED is set to true and JAVA__OPTS are left as they are
+for %%C in (^^^|) do (
+  if not "!JAVA_OPTS:%%C=!"=="!JAVA_OPTS!" set IS_SANITIZED=true
```
yersan, line 113: This is done character by character, so only the last occurrence will count to know whether it is sanitized. @lvydra Is this intentional?

yersan: Well, as soon as on is scaped it counts as completely sanitized. That should be fine, @lvydra, we are not going to treat scaped and non-scaped characters. If only one is scaped and not the others, the script will fail. Maybe it can be controlled better. I'm approving from my side

yersan (approved):
@jamezp Could you also take a look?

core-feature-pack/common/src/main/resources/content/bin/standalone.bat
yersan: Thanks, @lvydra , let's wait for James' input here. I think we are fine with what we have; now it can accept '|' or '^|' but not both at the same time, that could be a corner case and it is unlikely to happen.

jamezp: I've got no real strong opinion here as I cannot find the documentation as to how this works in batch scripts :) I trust whatever you guys feel is correct.

jamezp (commented):
The only thing I can think of that would be helpful is to add a test for this in the `testsuite/scripts` tests. Something could probably be added here https://github.com/wildfly/wildfly-core/blob/71173c7b8468936d31f41cc11583a82954d58679/testsuite/scripts/src/test/java/org/wildfly/common/test/ServerConfigurator.java#L254.

core-feature-pack/common/src/main/resources/content/bin/standalone.bat
yersan: Thanks, @jamezp , I had no idea we had those kinds of tests. For this one, I found this useful: https://docs.microsoft.com/en-us/windows-server/administration/windows-commands/for

@lvydra Could you take a look at whether we can add a test case for this one?. Code works, but a test case will ensure we will not get a regression. Let us know your findings, thanks!

lvydra: Thanks @jamezp, sure @yersan I will look at the tests.

lvydra:
Hi @jamezp @yersan, tests added.

testsuite/scripts/src/test/java/org/wildfly/scripts/test/StandaloneScriptTestCase.java
```diff
@@ -86,6 +89,28 @@ public static Collection<Object> data() {
         return result;
     }
 
+    @Test
+    public void testBatchScriptJavaOptsToEscape() throws Exception {
+        Assume.assumeTrue(!TestSuiteEnvironment.isWindows() && Shell.BASH.isSupported());
```
yersan, line 94: The assume condition needs to be fixed; it should be: `Assume.assumeTrue(TestSuiteEnvironment.isWindows())`

Although the test should pass on Linux, we don't need to run it there, indeed, looking at the method name, that was not your original intention.

testsuite/scripts/src/test/java/org/wildfly/scripts/test/StandaloneScriptTestCase.java
```diff
@@ -86,6 +89,28 @@ public static Collection<Object> data() {
...
+        ServerConfigurator.appendJavaOpts(ServerHelper.JBOSS_HOME, "standalone", variableToEscape);
+        try (ScriptProcess script = new ScriptProcess(ServerHelper.JBOSS_HOME, STANDALONE_BASE_NAME, Shell.BATCH, ServerHelper.TIMEOUT)) {
+            testScript(script);
+        }
+        ServerConfigurator.removeJavaOpts(ServerHelper.JBOSS_HOME, "standalone", variableToEscape);
+    }
+
+    @Test
+    public void testBatchScriptJavaOptsEscaped() throws Exception {
+        Assume.assumeTrue(!TestSuiteEnvironment.isWindows() && Shell.BASH.isSupported());
```
yersan, line 105: The assume condition needs to be fixed: `Assume.assumeTrue(TestSuiteEnvironment.isWindows())`

yersan (changes requested):

testsuite/scripts/src/test/java/org/wildfly/scripts/test/StandaloneScriptTestCase.java
lvydra: Oh sure, thanks @yersan, I don't know how it got there. :-)

lvydra:
@yersan fixed

testsuite/scripts/src/test/java/org/wildfly/scripts/test/StandaloneScriptTestCase.java
yersan: :) thanks, @lvydra

yersan (approved):

PR #5062, opened 2022-04-20
[WFCORE-5881] Use JPMS properties from launcher optionally

marekkopecky:
Jira: https://issues.redhat.com/browse/WFCORE-5881

The details from jira:

JPMS properties are currently [hardcoded in a launcher](https://github.com/wildfly/wildfly-core/blob/main/launcher/src/main/java/org/wildfly/core/launcher/AbstractCommandBuilder.java#L65-L78). Launcher is used by [arquillian container](https://mvnrepository.com/artifact/org.wildfly.arquillian/wildfly-arquillian-container-managed). But latest version of arquillian container uses wf-core 16.0.0.Final. So if we depend on old wf-core launcher, we can't ensure that proper JPMS settings would be used. Fortunately, it's quite easy to propagate correct JPMS properties to arquillian directly. This would be useful only if hard-coded properties from Launcher won't be used. This task adds option to not use hard-coded properties from launcher.

The default settings will be still the same - hardcoded JPMS properties would be used by default.

cc: @jamezp @bstansberry

### DIFF
--- a/launcher/src/main/java/org/wildfly/core/launcher/AbstractCommandBuilder.java
+++ b/launcher/src/main/java/org/wildfly/core/launcher/AbstractCommandBuilder.java
@@ -62,20 +62,22 @@ abstract class AbstractCommandBuilder<T extends AbstractCommandBuilder<T>> imple
         // Additions to these should include good explanations why in the relevant JIRA
         // Keep them alphabetical to avoid the code history getting confused by reordering commits
         final ArrayList<String> modularJavaOpts = new ArrayList<>();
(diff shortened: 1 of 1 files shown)


marekkopecky:
CI issues doesn't seems to be related with the MR. CI runs on another MRs have similar issues, see for example:
* https://github.com/wildfly/wildfly-core/pull/5063
* https://ci.wildfly.org/viewLog.html?buildId=301158&buildTypeId=WildFlyCore_PullRequest_WildFlyCoreFullIntegrationLinuxJdk11

bstansberry:
Thanks, @marekkopecky

PR #5063, opened 2022-04-20
WFCORE-5862 Excessive output regarding downloads when project built

clara0:
https://issues.redhat.com/browse/WFCORE-5862

### DIFF
--- a/.github/workflows/dep-diff-pull_request.yml
+++ b/.github/workflows/dep-diff-pull_request.yml
@@ -56,7 +56,7 @@ jobs:
       - name: Build base
         working-directory: base
         run: |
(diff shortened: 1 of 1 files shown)


yersan:
Thanks, @clara0  for the details about opening the Jira. 

This change sounds appropriate for the GitHub actions and how we are using it. I usually don't look at the traces on that Job,  but since it is a personal matter, it would be great to hear any other opinion before approving. Maybe @jmesnil / @kabir ?

yersan (approved):
Thanks for the Feedback Kabir, I share the same impression. If someone misses them, we can roll back it.

yersan:
Thank you for the contribution, @clara0 !

PR #5068, opened 2022-04-22
[WFCORE-5842] [primary/secondary] Bump the xsd to 20.0 due to documen…

moulalis:
Issue: https://issues.redhat.com/browse/WFCORE-5842

### DIFF
--- a/server/src/main/resources/schema/wildfly-config_20_0.xsd
+++ b/server/src/main/resources/schema/wildfly-config_20_0.xsd
@@ -1,0 +1,3892 @@
+<?xml version="1.0" encoding="UTF-8"?>
+
+<!--
+  ~ JBoss, Home of Professional Open Source.
+  ~ Copyright 2021 Red Hat, Inc., and individual contributors
+  ~ as indicated by the @author tags.
+  ~
+  ~ Licensed under the Apache License, Version 2.0 (the "License");
+  ~ you may not use this file except in compliance with the License.
+  ~ You may obtain a copy of the License at
+  ~
+  ~     http://www.apache.org/licenses/LICENSE-2.0
+  ~
+  ~ Unless required by applicable law or agreed to in writing, software
+  ~ distributed under the License is distributed on an "AS IS" BASIS,
+  ~ WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+  ~ See the License for the specific language governing permissions and
+  ~ limitations under the License.
+  -->
+
+<xs:schema xmlns:xs="http://www.w3.org/2001/XMLSchema"
+           xmlns="urn:jboss:domain:19.0"
+           xmlns:credential-reference="urn:wildfly:credential-reference:1.1"
+           targetNamespace="urn:jboss:domain:19.0"
+           elementFormDefault="qualified"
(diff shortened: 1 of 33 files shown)


server/src/main/resources/schema/wildfly-config_20_0.xsd
```diff
@@ -0,0 +1,3892 @@
...
+  ~
+  ~ Unless required by applicable law or agreed to in writing, software
+  ~ distributed under the License is distributed on an "AS IS" BASIS,
+  ~ WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+  ~ See the License for the specific language governing permissions and
+  ~ limitations under the License.
+  -->
+
+<xs:schema xmlns:xs="http://www.w3.org/2001/XMLSchema"
+           xmlns="urn:jboss:domain:19.0"
```
moulalis, line 22: @yersan this also be changed to 20.0 in this scope?

jmesnil:
@moulalis could you specify what are the documentation changes? Could you also paste or attached a diff between the XSD files to help figuring out what changed? thanks

darranl:
Actually if we have modifications to the schema can we please split them into their own commit otherwise when using IDEs and command line tools we will not be able to detect what happened.

WildFly Core may need a schema bump for other reasons but a documentation change doesn't sound like a sufficient justification unless there is also a change in behaviour.

yersan:
> WildFly Core may need a schema bump for other reasons but a documentation change doesn't sound like a sufficient justification unless there is also a change in behaviour.

Hi @darranl , thanks for pointing it out. The original intention of this PR was just to adapt the wrong language usage used in the XSD (e.g "fetch-from-master" admin only policy value), however, the analysis doc of the related feature revealed we will need to adapt the parsers to accept "fetch-from-master" and "fetch-from-domain-controller, so it became in a non-only documentation issue.

The changes to adapt the parsers will be tracked on a different Jira. Now I'm wondering if all should be addressed by this one (WFCORE-5842) but using different commits to separate the changes. Feel free to suggest, otherwise, we will continue handling two Jiras.

moulalis:
> @moulalis could you specify what are the documentation changes? Could you also paste or attached a diff between the XSD files to help figuring out what changed? thanks

@jmesnil PFA screenshots of the differences. https://docs.google.com/document/d/10JWg75ASnR5wE2PMytaEp1I0PbHI_UvZfyEyxZEoPks/edit#heading=h.uffl05i85a9a is the reference document.

Screenshot1
Screenshot3
Screenshot 2
Screenshot 4
Screenshot 5
Screenshot 6
Screenshot 7
Screenshot 8

darranl:
@moulalis please split this into two commits so we have access to the history in the future.

jmesnil (changes requested):
@moulalis none of the changes you made impact the schema. There is no need to bump them if only the doc is updated.

The doc can be updated in the existing 19.0 schema and we will bump it when the structure of the schema is effectively changed.

yersan:
As discussed internally with @moulalis , closing this for now, we will open a different one including a replacement of fetch-from-master with fetch-from-domain-controller (WFCORE-5901)

PR #5070, opened 2022-04-22
[WFCORE-5320] ModelTestKernelServices providing BootErrorCollector info

khermano:
Issue: https://issues.redhat.com/browse/WFCORE-5320

### DIFF
--- a/model-test/src/main/java/org/jboss/as/model/test/ModelTestKernelServices.java
+++ b/model-test/src/main/java/org/jboss/as/model/test/ModelTestKernelServices.java
@@ -27,11 +27,17 @@ import java.util.List;
 import org.jboss.as.controller.ModelController.OperationTransactionControl;
 import org.jboss.as.controller.ModelVersion;
 import org.jboss.as.controller.OperationFailedException;
+import org.jboss.as.controller.PathAddress;
+import org.jboss.as.controller.PathElement;
+import org.jboss.as.controller.operations.common.Util;
 import org.jboss.as.controller.registry.ImmutableManagementResourceRegistration;
 import org.jboss.as.controller.transform.OperationTransformer.TransformedOperation;
 import org.jboss.as.controller.transform.TransformerOperationAttachment;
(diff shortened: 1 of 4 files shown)


model-test/src/main/java/org/jboss/as/model/test/ModelTestKernelServices.java
```diff
@@ -183,4 +189,43 @@ TransformedOperation transformOperation(ModelVersion modelVersion, ModelNode ope
...
+        }
+        return result;
+    }
+
+    /**
+     *
+     * @return true if BootErrorCollector collected any failure
+     */
+    default boolean hasBootErrorCollectorFailures() {
+        return getBootErrorCollectorFailures() != null && !getBootErrorCollectorFailures().asString().equals("undefined");
```
yersan, line 229: Just in relation to the above, I'm wondering if we should remove `getBootErrorCollectorFailures().asString().equals("undefined")` here and make `getBootErrorCollectorFailures()` return null if only if there are no error failures. 

Just wondering, if you think we can leave it as it is, I'm fine, but it should be reflected on the Java Doc.

yersan (changes requested):

model-test/src/main/java/org/jboss/as/model/test/ModelTestKernelServices.java
khermano: Of course, you're right. :) There is no point in keeping undefined failure without any further details so I changed it as you suggested. Also, I improved the Java Doc. Hopefully, it looks more suitable now. What do you think? Is it okay like that or did I miss something?

yersan (approved):

model-test/src/main/java/org/jboss/as/model/test/ModelTestKernelServices.java
yersan: Thanks, @khermano , looks better now, merged

PR #5075, opened 2022-04-25
[WFCORE-5679] Generate and export operation

lvydra:
https://issues.redhat.com/browse/WFCORE-5679


fjuma:
@darranl Just wanted to check with you since you had created [WFCORE-5679](https://issues.redhat.com/browse/WFCORE-5679) - this issue is about introducing new runtime operations for the `credential-store` resource. Although it doesn't affect subsystem configuration itself, it does affect documentation. Should we create an RFE to track this?

darranl:
I think lets get all the pieces together for this first and reassess if we want an RFE.  IMO this probably could be an enhancement rather than a full feature.

We should double check we have sufficient tests and we will also need a PR to WildFly for the community documentation.

elytron/src/main/java/org/wildfly/extension/elytron/AbstractCredentialStoreResourceDefinition.java
```diff
@@ -90,10 +100,53 @@ abstract class AbstractCredentialStoreResourceDefinition extends SimpleResourceD
...
+
     static final SimpleAttributeDefinition KEY = new SimpleAttributeDefinitionBuilder(ElytronDescriptionConstants.KEY, ModelType.STRING, false)
             .setMinSize(1)
             .build();
 
+    static final SimpleAttributeDefinition PUBLIC_KEY_STRING = new SimpleAttributeDefinitionBuilder(ElytronDescriptionConstants.PUBLIC_KEY_STRING, ModelType.STRING, true)
+            .setMinSize(1)
+            .build();
+
+    static final SimpleAttributeDefinition PRIVATE_KEY_STRING = new SimpleAttributeDefinitionBuilder(ElytronDescriptionConstants.PRIVATE_KEY_STRING, ModelType.STRING, true)
```
fjuma, line 119: Same here.

elytron/src/main/java/org/wildfly/extension/elytron/AbstractCredentialStoreResourceDefinition.java
```diff
@@ -184,6 +237,109 @@ protected void removeAliasOperation(OperationContext context, ModelNode operatio
...
+            if (credentialStore.exists(alias, KeyPairCredential.class)) {
+                throw ROOT_LOGGER.credentialAlreadyExists(alias, KeyPairCredential.class.getName());
+            }
+
+            String algorithm;
+            ModelNode algorithmModel = ALGORITHM.resolveModelAttribute(context, operation);
+            if (algorithmModel.isDefined()) {
+                algorithm = algorithmModel.asString();
+            } else {
+                algorithm = RSA_ALGORITHM;
```
fjuma, line 253: The default value can be defined above using the `SimpleAttributeDefinitionBuilder` instead of doing it here.

elytron/src/main/java/org/wildfly/extension/elytron/AbstractCredentialStoreResourceDefinition.java
```diff
@@ -184,6 +237,109 @@ protected void removeAliasOperation(OperationContext context, ModelNode operatio
...
+            }
+
+            int size;
+            ModelNode sizeModel = SIZE.resolveModelAttribute(context, operation);
+            if (sizeModel.isDefined()) {
+                size = sizeModel.asInt();
+            } else {
+                size = 0;
```
fjuma, line 261: Same here.

elytron/src/main/java/org/wildfly/extension/elytron/AbstractCredentialStoreResourceDefinition.java
```diff
@@ -90,10 +100,53 @@ abstract class AbstractCredentialStoreResourceDefinition extends SimpleResourceD
...
 
+    static final SimpleAttributeDefinition PUBLIC_KEY_STRING = new SimpleAttributeDefinitionBuilder(ElytronDescriptionConstants.PUBLIC_KEY_STRING, ModelType.STRING, true)
+            .setMinSize(1)
+            .build();
+
+    static final SimpleAttributeDefinition PRIVATE_KEY_STRING = new SimpleAttributeDefinitionBuilder(ElytronDescriptionConstants.PRIVATE_KEY_STRING, ModelType.STRING, true)
+            .setMinSize(1)
+            .build();
+
+    static final SimpleAttributeDefinition PUBLIC_KEY_LOCATION = new SimpleAttributeDefinitionBuilder(ElytronDescriptionConstants.PUBLIC_KEY_LOCATION, ModelType.STRING, true)
```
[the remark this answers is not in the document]
fjuma: Have you checked to verify that an error occurs if neither are specified?

elytron/src/main/java/org/wildfly/extension/elytron/AbstractCredentialStoreResourceDefinition.java
```diff
@@ -184,6 +237,109 @@ protected void removeAliasOperation(OperationContext context, ModelNode operatio
...
+            if (credentialStore.exists(alias, KeyPairCredential.class)) {
+                throw ROOT_LOGGER.credentialAlreadyExists(alias, KeyPairCredential.class.getName());
+            }
+
+            ModelNode privateKeyString = PRIVATE_KEY_STRING.resolveModelAttribute(context, operation);
+            ModelNode publicKeyString = PUBLIC_KEY_STRING.resolveModelAttribute(context, operation);
+            ModelNode privateKeyLocation = PRIVATE_KEY_LOCATION.resolveModelAttribute(context, operation);
+            ModelNode publicKeyLocation = PUBLIC_KEY_LOCATION.resolveModelAttribute(context, operation);
+
+            String passphrase = PASSPHRASE.resolveModelAttribute(context, operation).asString();
```
fjuma, line 306: Looks like the `PASSPHRASE` is optional so it could be null here. Should it be required?

fjuma: In general, would be good to double check which attributes are required for each operation.

elytron/src/main/java/org/wildfly/extension/elytron/KeyPairUtil.java
```diff
@@ -0,0 +1,136 @@
...
+import java.security.KeyPair;
+import java.security.KeyPairGenerator;
+import java.security.NoSuchAlgorithmException;
+import java.security.PrivateKey;
+import java.security.PublicKey;
+import java.security.SecureRandom;
+
+import static org.wildfly.extension.elytron._private.ElytronSubsystemMessages.ROOT_LOGGER;
+
+public class KeyPairUtil {
```
fjuma, line 39: Does this need to be public?

elytron/src/main/java/org/wildfly/extension/elytron/KeyPairUtil.java
```diff
@@ -0,0 +1,136 @@
...
+        } catch (InvalidParameterException e) {
+            throw ROOT_LOGGER.invalidKeyPairKeySize(e.getMessage());
+        }
+        return keyPairGenerator.generateKeyPair();
+    }
+
+    public static KeyPair parseKeyPair(String privateKeyContent, String publicKeyContent, ElytronFilePasswordProvider passwordProvider) {
+        KeyPair keyPair;
+        try {
+            keyPair = Pem.parsePemOpenSSHContent(CodePointIterator.ofString(privateKeyContent), passwordProvider).next().tryCast(KeyPair.class);
```
fjuma, line 93: We might want to consider having a different parameter for specifying OpenSSH format so we don't try to guess the format. As an example, take a look at how this was done here:

https://github.com/wildfly-security/wildfly-elytron/blob/1.x/auth/client/src/main/resources/schema/elytron-client-1_7.xsd

fjuma (changes requested):

fjuma:
@lvydra Thanks for working on this! I've added some comments. Feel free to let me know if you have any questions.

elytron/src/main/java/org/wildfly/extension/elytron/AbstractCredentialStoreResourceDefinition.java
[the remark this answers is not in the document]
lvydra: > Since either PUBLIC_KEY or PUBLIC_KEY_LOCATION should be specified, we should use `setAlternatives` to specify this. For some examples, there are other places in the Elytron subsystem where `setAlternatives` is used.

I have added `setAlternatives`, however, it's not working here how it is supposed to. I'm not sure if I'm not missing something.

elytron/src/main/java/org/wildfly/extension/elytron/AbstractCredentialStoreResourceDefinition.java
lvydra: > Looks like the `PASSPHRASE` is optional so it could be null here. Should it be required?

Passphrase can be null if none was used to encrypt the key, while it is not recommended. I have added a test case to check it.

lvydra:
Hi @fjuma Thanks for the review, I have updated PR and added some comments.

yersan:
hi @lvydra, it looks like errors are related to your changes. Could you take a look?

lvydra:
Hi @yersan, thanks, I will look at it.

elytron/src/test/java/org/wildfly/extension/elytron/CredentialStoreTestCase.java
```diff
@@ -221,13 +222,14 @@ public void testExportPublicKey() {
         Assert.assertTrue(publicKey.contains("ssh-rsa AAAAB3NzaC1yc2EAAAADAQABAAABAQ"));
     }
 
+    @Ignore
```
fjuma, line 225: @lvydra Feel free to let us know when this is ready for another review. Since this test is currently being ignored, it looks like you're working on fixing the test failures, right?

lvydra: Hi @fjuma, yes, I'm working on fixing test failures. Ignored tests are written to test `setAlternatives` behavior, which is currently not working as it's supposed to, so I'm looking for a way how to fix that.

lvydra:
Hi @fjuma, I think that the PR should be ready for another round of review :-) All requests should be addressed, the only persisting problem is  `setAlternatives` validation, which I haven't been able to get working. I have added a custom validation method using alternatives attribute, however, I don't think it's a suitable solution.

elytron/src/test/java/org/wildfly/extension/elytron/CredentialStoreTestCase.java
```diff
@@ -153,6 +158,391 @@ private void testDefaultContent(final String storeType, final String storeName,
...
+        String algorithm = EC_ALGORITHM;
+        String size = "521";
+
+        generateKeyPair(storeType, storeName, aliasName, algorithm, size);
+        exportKeyPairPublicKey(storeType, storeName, aliasName);
+        removeKeyPair(storeType, storeName, aliasName);
+    }
+
+    @Test
+    public void testExportPublicKey() {
```
fjuma, line 212: Would also be good to add a test for the case where the alias doesn't exist.

fjuma:
Thanks for the updates, @lvydra! I've added some comments.

lvydra:
Hi @fjuma, thanks for the review, I have updated PR.

fjuma:
This PR is still active.

@lvydra Looks like there's a conflict that now needs to be resolved.

lvydra:
Hi @fjuma, thanks, resolved.

yersan:
@fjuma / @darranl Is there any decision about handling this as an enhancement vs Feature Request?
Since there is a user controllable action here, probably it is more inclined to a Feature Request, however, it seems it doesn't change the subsystem configuration, so here are my doubts.

fjuma:
@lvydra Thanks for the updates! Sorry for the delayed response, just catching up after being away for a couple weeks.

Just to check, have you already created a PR against WildFly with the corresponding community documentation?

lvydra:
Hi @fjuma, not yet, I will prepare a documentation update and open PR.

lvydra:
Hi @fjuma, I have opened documentation PR: https://github.com/wildfly/wildfly/pull/16139

fjuma:
This PR is still active.

fjuma:
This PR is still active

yersan:
@fjuma / @darranl Is there any decision about handling this as an enhancement vs Feature Request?
Since there is a user controllable action here, probably it is more inclined to a Feature Request, however, it seems it doesn't change the subsystem configuration, so here are my doubts.

elytron/src/test/java/org/wildfly/extension/elytron/CredentialStoreTestCase.java
yersan: This concern seems to be already addressed by `testExportPublicKeyFailNonExistingAlias` test

PR #5077, opened 2022-04-25
WFCORE-5892 IndexOutOfBoundsException in RequirementRegistration.getOldestRegistrationPoint() during sequence of resource removals

pferraro:
https://issues.redhat.com/browse/WFCORE-5892

### DIFF
--- a/controller/src/main/java/org/jboss/as/controller/capability/registry/RequirementRegistration.java
+++ b/controller/src/main/java/org/jboss/as/controller/capability/registry/RequirementRegistration.java
@@ -73,7 +73,9 @@ public class RequirementRegistration {
      */
     public RequirementRegistration(RuntimeRequirementRegistration toCopy) {
         this(toCopy.getRequiredName(), toCopy.getDependentId());
-        this.registrationPoints.putAll(((RequirementRegistration)toCopy).registrationPoints);
+        for (Map.Entry<PathAddress, List<RegistrationPoint>> entry : ((RequirementRegistration) toCopy).registrationPoints.entrySet()) {
+            this.registrationPoints.put(entry.getKey(), new ArrayList<>(entry.getValue()));
+        }
     }
 
     public String getRequiredName() {


controller/src/main/java/org/jboss/as/controller/capability/registry/RequirementRegistration.java
```diff
@@ -73,7 +73,9 @@ protected RequirementRegistration(String requiredName, String dependentName, Cap
      */
     public RequirementRegistration(RuntimeRequirementRegistration toCopy) {
         this(toCopy.getRequiredName(), toCopy.getDependentId());
-        this.registrationPoints.putAll(((RequirementRegistration)toCopy).registrationPoints);
+        for (Map.Entry<PathAddress, List<RegistrationPoint>> entry : ((RequirementRegistration) toCopy).registrationPoints.entrySet()) {
+            this.registrationPoints.put(entry.getKey(), new ArrayList<>(entry.getValue()));
```
yersan, line 77: Hi @pferraro , I have some difficulties understanding how this fixes the error trace described in the Jira.

Please correct me if I am wrong, the error occurs because the code tries to access an element of an empty array list stored in the `registrationPoints` LinkedHashMap.

If `this.registrationPoints.putAll(((RequirementRegistration)toCopy).registrationPoints);` introduces such an empty array list, how `new ArrayList<>(entry.getValue())` will create a non-empty array list?

I have to be missing something. 

Map.putAll Java doc describes the behavior of this operation is undefined if the specified map is modified while the operation is in progress. Is that the reason to replace the Map.putAll by your for loop?

pferraro: @yersan 
> If this.registrationPoints.putAll(((RequirementRegistration)toCopy).registrationPoints); introduces such an empty array list

This is not the cause of the empty list.

Currently, the RequirementRegistration constructor only performs a shallow copy of the provided registration point map, rather than a deep copy.  Consider 2 instances of RequirementRegistration created from the same RuntimeRequirementRegistration (to support 2 successive management batches).  When the first management batch (which contains a remove operation that removes a registration point) executes it will remove a registration point from one of its lists.  However, because each RequirementRegistration references the same list (because these were never copied), this removal is also reflected in the second RequirementRegistration.  When it too tries to remove the same registration point, it fails with an IndexOutOfBoundsException.

See the attached jira (which contains a link to the failing CI job).  With this fix, the IndexOutOfBoundsException no longer occurs.

yersan (approved):
Thanks for the explanation, @pferraro . I missed the shared reference that was indeed in the Jira description.

PR #5085, opened 2022-04-27
[WFCORE-5649] Assertion arguments should be passed correct (cli)

boris-unckel:
Assertion arguments should be passed in the correct order (cli)

Fixes https://issues.redhat.com/browse/WFCORE-5649


darranl (approved):

yersan:
Thanks, @boris-unckel !

PR #5086, opened 2022-04-27
[18.x][WFCORE-5649] Assertion arguments should be passed correct (cli)

boris-unckel:
Assertion arguments should be passed in the correct order (cli)

Fixes https://issues.redhat.com/browse/WFCORE-5649
Upstream https://github.com/wildfly/wildfly-core/pull/5085


yersan:
Thanks, @boris-unckel

PR #5088, opened 2022-04-27
WFCORE-5872: update module generation to use current schema

michpetrov:
Issue: [WFCORE-5872](https://issues.redhat.com/browse/WFCORE-5872)

### DIFF
--- a/cli/src/main/java/org/jboss/as/cli/handlers/module/ASModuleHandler.java
+++ b/cli/src/main/java/org/jboss/as/cli/handlers/module/ASModuleHandler.java
@@ -65,9 +65,12 @@ import org.jboss.as.cli.operation.ParsedCommandLine;
 import org.jboss.as.cli.parsing.ExpressionBaseState;
 import org.jboss.as.cli.parsing.ParsingState;
 import org.jboss.as.cli.parsing.WordCharacterHandler;
+import org.jboss.modules.xml.MXParser;
 import org.jboss.staxmapper.FormattingXMLStreamWriter;
 import org.jboss.staxmapper.XMLExtendedStreamWriter;
 import org.wildfly.security.manager.WildFlySecurityManager;
+
+import static org.jboss.modules.xml.XmlPullParser.FEATURE_PROCESS_NAMESPACES;
 
 /**
  *
@@ -135,6 +138,7 @@ public class ASModuleHandler extends CommandHandlerWithHelp {
     private final ArgumentWithValue moduleRootDir;
     private final ArgumentWithoutValue allowNonExistentResources;
     private File modulesDir;
+    private String detectedSchemaVersion;
 
     public ASModuleHandler(CommandContext ctx) {
         super("module", false);
@@ -486,6 +490,11 @@ public class ASModuleHandler extends CommandHandlerWithHelp {
                 config.setMainClass(mainCls);
             }
 
+            final String schema = getSchemaVersion(ctx,slotVal != null);
+            if (schema != null) {
+                config.setSchemaVersion(schema);
+            }
+
             FileOutputStream fos = null;
             final File moduleFile = new File(moduleDir, "module.xml");
             try {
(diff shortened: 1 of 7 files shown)


cli/src/main/java/org/jboss/as/cli/handlers/module/ASModuleHandler.java
```diff
@@ -707,4 +716,38 @@ public static void copy(final File source, final File target) throws CommandLine
...
+        if (detectedSchemaVersion == null) {
+            MXParser parser = new MXParser();
+            final File wfCommonModule = new File(modulesDir, "system.layers.base.org.wildfly.common.main.".replace('.', File.separatorChar) + "module.xml");
+            try {
+                parser.setFeature(FEATURE_PROCESS_NAMESPACES, true);
+                parser.setInput(new FileInputStream(wfCommonModule), null);
+                parser.nextTag();
+            } catch (Exception ignore) {
+            }
+            detectedSchemaVersion = parser.getNamespace();
```
jfdenise, line 729: If an error occured, seems strange to access the parser. This should be guarded by the exception handling.

cli/src/main/java/org/jboss/as/cli/handlers/module/ASModuleHandler.java
```diff
@@ -707,4 +716,38 @@ public static void copy(final File source, final File target) throws CommandLine
...
-}
+    private String getSchemaVersion(CommandContext ctx, boolean hasSlot) {
+        if (detectedSchemaVersion == null) {
+            MXParser parser = new MXParser();
+            final File wfCommonModule = new File(modulesDir, "system.layers.base.org.wildfly.common.main.".replace('.', File.separatorChar) + "module.xml");
+            try {
+                parser.setFeature(FEATURE_PROCESS_NAMESPACES, true);
+                parser.setInput(new FileInputStream(wfCommonModule), null);
+                parser.nextTag();
+            } catch (Exception ignore) {
```
jfdenise, line 727: You could list the set of exceptions and add a comment saying that it is safe to ignore the exception, we will simply not have a discovered schema.

cli/src/main/java/org/jboss/as/cli/handlers/module/ASModuleHandler.java
```diff
@@ -707,4 +716,38 @@ public static void copy(final File source, final File target) throws CommandLine
...
+            }
+            detectedSchemaVersion = parser.getNamespace();
+        }
+
+        if (hasSlot && detectedSchemaVersion != null) {
```
jfdenise, line 732: I think that we have the opportunity to create a "requiresSlot" function that would try to discover the current version. This function could be used in the slot parameter canAppearNext function (You can check in CLI code, we have a lot of examples) to only display the --slot argument (when pressing TAB) if the server supports it (so version <= 1.5). If any error occurs trying to discover the version, then we should fallback to use slot, the safer.

cli/src/main/resources/help/module.txt
```diff
@@ -49,7 +49,7 @@ ARGUMENTS
  --name          - (required) the name of the module to be added or removed.
                    NOTE: slot is not a part of the module name.
 
- --slot          - (optional) specifies a slot which should be created or
+ --slot          - (deprecated, optional) specifies a slot which should be created or
```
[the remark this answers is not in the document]
michpetrov: I was under the impression that kind of name wasn't allowed but I think I misunderstood how the parser works.

But then this begs a question, if a user calls `module add --name=org.example.module --slot=test` and we detect version 1.9 then instead of downgrading to 1.5 we could make it work as if the user called `module add --name=org.example.module:test` and no downgrade is necessary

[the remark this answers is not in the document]
jfdenise: @michpetrov , yes I thought at it too. I liked the idea but in some way it uses a deprecated concept (slot) to convey the new version information. The JBoss modules documentation: https://jboss-modules.github.io/jboss-modules/manual/#versions explains it well.
I think that we should clearly specify the behavior in all the different cases, a wildfly-proposals would be a nice way to capture the new command logic: https://github.com/wildfly/wildfly-proposals/

cli/src/main/java/org/jboss/as/cli/handlers/module/ASModuleHandler.java
darranl: These kind of catches can lead to painful remote debugging sessions when end users run into issues, I think as @jfdenise says it would be better to restrict to the exceptions expected to be ignored as at the moment this will also catch all Runtime exceptions.

Regardless I think it would still be useful to log the exception at TRACE level then there remains a way to obtain information about what happened.

PR #5095, opened 2022-05-09
[WFCORE-5905] Don't reject EE 10 permissions.xml files

ropalka:
https://issues.redhat.com/browse/WFCORE-5905

### DIFF
--- a/security-manager/src/main/java/org/wildfly/extension/security/manager/deployment/PermissionsParser.java
+++ b/security-manager/src/main/java/org/wildfly/extension/security/manager/deployment/PermissionsParser.java
@@ -90,7 +90,7 @@ public class PermissionsParser {
             switch (attribute) {
                 case VERSION: {
                     String version = reader.getAttributeValue(i);
-                    if (!"7".equals(version) && !"9".equals(version))
+                    if (!"7".equals(version) && !"9".equals(version) && !"10".equals(version))
                         throw SecurityManagerLogger.ROOT_LOGGER.invalidPermissionsXMLVersion(version, "7");
                     break;
                 }


yersan (approved):

security-manager/src/main/java/org/wildfly/extension/security/manager/deployment/PermissionsParser.java
```diff
@@ -90,7 +90,7 @@ private static List<PermissionFactory> parsePermissions(final XMLStreamReader re
             switch (attribute) {
                 case VERSION: {
                     String version = reader.getAttributeValue(i);
-                    if (!"7".equals(version) && !"9".equals(version))
+                    if (!"7".equals(version) && !"9".equals(version) && !"10".equals(version))
```
darranl, line 93: Should there be an "8" in this list as well or did they skip that one entirely?

yersan: I understood there were no 8 version: http://www.oracle.com/webfolder/technetwork/jsc/xml/ns/javaee/index.html#8

[XMLResourceResolver.java](https://github.com/jboss/metadata/pull/146/commits/aabb2978a93872744629ab88b987ee95e7d195fa#diff-d0f31f9160bf60f5a2d3f2a8ff1a25488a7b1d6991cef7723edb4f869cb857ee)

yersan: I believe the jump from 7 to 9 was to align with Jakarta EE 9

darranl (approved):

PR #5098, opened 2022-05-10
WFCORE-5841 [primary/secondary] Review the documentation of affected CLI management operations

moulalis:
Issue: https://issues.redhat.com/browse/WFCORE-5841

### DIFF
--- a/cli/src/main/resources/help/reload.txt
+++ b/cli/src/main/resources/help/reload.txt
@@ -46,8 +46,8 @@ ARGUMENTS
                  when it restarts. An ADMIN_ONLY controller will start any
                  configured management interfaces and accept management
                  requests, but will not start servers or, if this host
-                 controller is the master for the domain, accept incoming
-                 connections from slave host controllers. For embedded host
+                 controller is the primary Host Controller for the domain, accept incoming
+                 connections from secondary host controllers. For embedded host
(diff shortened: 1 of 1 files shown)


cli/src/main/resources/help/reload.txt
```diff
@@ -46,8 +46,8 @@ ARGUMENTS
                  when it restarts. An ADMIN_ONLY controller will start any
                  configured management interfaces and accept management
                  requests, but will not start servers or, if this host
-                 controller is the master for the domain, accept incoming
-                 connections from slave host controllers. For embedded host
+                 controller is the primary for the domain, accept incoming
```
yersan, line 49: Our analysis doc states the following:

"Where ‘master’ was used as a shorthand for ‘master Host Controller’ replace with either  ‘Domain Controller’ or ‘primary Host Controller’."

So I think we should use ‘primary Host Controller’ here. @bstansberry Would you mind confirming? honestly, I always have difficulties with these kinds of subtlety.

moulalis: Same here, waiting for @bstansberry to confirm.

yersan: @moulalis Could you please update the PR accordingly? thanks!

yersan (approved):

PR #5102, opened 2022-05-12
[WFCORE-5912] [blacklist/whitelist] Find and replace occurrences in d…

moulalis:
Issue: https://issues.redhat.com/browse/WFCORE-5912

### DIFF
--- a/remoting/subsystem/src/main/resources/org/jboss/as/remoting/LocalDescriptions.properties
+++ b/remoting/subsystem/src/main/resources/org/jboss/as/remoting/LocalDescriptions.properties
@@ -124,7 +124,7 @@ local-outbound-connection.property=The XNIO Options that will be used during the
 sasl=The "sasl" element contains the SASL authentication configuration for this connector.
 sasl.add=Adds the SASL authentication configuration for its connector
 sasl.remove=Removes the SASL authentication configuration for its connector
-sasl.include-mechanisms=The optional nested "include-mechanisms" element contains a whitelist of allowed SASL mechanism names. No mechanisms will be allowed which are not present in this list.
+sasl.include-mechanisms=The optional nested "include-mechanisms" element contains a list of allowed SASL mechanism names. Only mechanisms that are present in this list will be allowed.
 sasl.qop=The optional nested "qop" element contains a list of quality-of-protection values, in decreasing order of preference.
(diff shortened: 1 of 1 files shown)


remoting/subsystem/src/main/resources/org/jboss/as/remoting/LocalDescriptions.properties
```diff
@@ -124,7 +124,7 @@ local-outbound-connection.property=The XNIO Options that will be used during the
 sasl=The "sasl" element contains the SASL authentication configuration for this connector.
 sasl.add=Adds the SASL authentication configuration for its connector
 sasl.remove=Removes the SASL authentication configuration for its connector
-sasl.include-mechanisms=The optional nested "include-mechanisms" element contains a whitelist of allowed SASL mechanism names. No mechanisms will be allowed which are not present in this list.
+sasl.include-mechanisms=The optional nested "include-mechanisms" element contains a allowlist of allowed SASL mechanism names. No mechanisms will be allowed which are not present in this list.
```
yersan, line 127: s/a allowlist/an allowlist. However, in any case, it results in redundancy.


What about if we use:

"The optional nested "include-mechanisms" element contains a list of allowed SASL mechanism names. No mechanisms will be allowed which are not present in this list"

or

"The optional nested "include-mechanisms" element contains an allowlist of SASL mechanism names. No mechanisms will be allowed which are not present in this list"


@fjuma as a native English speaker, would you mind helping us with this sentence? thanks!

yersan (changes requested):

remoting/subsystem/src/main/resources/org/jboss/as/remoting/LocalDescriptions.properties
fjuma: @yersan The first option you proposed looks good. I wonder if we could reword the second sentence to make it more clear, e.g.,

"The optional nested "include-mechanisms" element contains a list of allowed SASL mechanism names. Only mechanisms that are present in this list will be allowed."

yersan: Thanks, Farah. That sounds also more clear to me too, +1 to your proposal.

moulalis: Thanks @fjuma and @yersan, I've updated the PR accordingly.

PR #5106, opened 2022-05-19
[WFCORE-5914] Elytron - key-size of secret-key-credential-store is not validated

lvydra:
https://issues.redhat.com/browse/WFCORE-5914

### DIFF
--- a/elytron/src/main/java/org/wildfly/extension/elytron/SecretKeyCredentialStoreDefinition.java
+++ b/elytron/src/main/java/org/wildfly/extension/elytron/SecretKeyCredentialStoreDefinition.java
@@ -46,6 +46,7 @@ import org.jboss.as.controller.SimpleAttributeDefinitionBuilder;
 import org.jboss.as.controller.SimpleOperationDefinition;
 import org.jboss.as.controller.SimpleOperationDefinitionBuilder;
 import org.jboss.as.controller.descriptions.StandardResourceDescriptionResolver;
+import org.jboss.as.controller.operations.validation.IntAllowedValuesValidator;
 import org.jboss.as.controller.registry.ManagementResourceRegistration;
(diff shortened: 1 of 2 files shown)


elytron/src/main/java/org/wildfly/extension/elytron/SecretKeyCredentialStoreDefinition.java
```diff
@@ -96,7 +97,7 @@ class SecretKeyCredentialStoreDefinition extends AbstractCredentialStoreResource
             .setAttributeGroup(ElytronDescriptionConstants.IMPLEMENTATION)
             .setAllowExpression(true)
             .setDefaultValue(new ModelNode(256))
-            .setAllowedValues(128, 192, 256)
+            .setValidator(new IntAllowedValuesValidator(128, 192, 256))
```
yersan, line 100: I think we could use both allowed values and the validator. One thing that allowed values does is to show the expected values when the user press <TAB>. For this case makes sense to me since it can help the user to think about which values are allowed, they are difficult to identify. 

@lvydra WDYT?

yersan (changes requested):

elytron/src/main/java/org/wildfly/extension/elytron/SecretKeyCredentialStoreDefinition.java
lvydra: Hi @yersan, as I tested it now, it seems that the validator offers the same functionality in terms of showing expected values.

yersan: Ahh, great then, thanks for checking it

yersan (approved):

PR #5108, opened 2022-05-24
WFCORE-5897 [primary/secondary] Problematic Language usage deprecation and replacement in testsuite

parsharma:
Issue: https://issues.redhat.com/browse/WFCORE-5897


yersan:
Hi @parsharma , I would say we need to split the changes into at least three commits as described in the main issue. We should not add any change unrelated to each point, for example, do not change elytron security user names or auth contexts, or variable names, or whatever is not forced to be changed for each main point, just focus on the three points:

1. Renaming the host-[master/slave].xml configuration files
2. Rename the default DC hostname
3. Rename properties used in the configuration files

That will ensure we are changing the tests accordingly and ensure we are not introducing wrong behaviors.

testsuite/domain/src/test/resources/host-configs/host-secondary.xml
```diff
@@ -23,7 +23,7 @@
 <host xmlns="urn:jboss:domain:20.0"
       xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
       xsi:schemaLocation="urn:jboss:domain:20.0 wildfly-config_20_0.xsd"
-      name="slave">
+      name="secondary">
```
yersan, line 26: Check HcExtensionAndSubsystemManagementTestCase#checkSlaveReconnected, it looks like it is using a wrong host name

yersan (changes requested):
In addition to the above, there are still some uses of unexpected host names in the following classes that should be part of the work done renaming the host names:

```
src/main/java/org/jboss/as/test/integration/management/extension/blocker/BlockerExtension.java:                    name = "master";
src/main/java/org/jboss/as/test/integration/management/extension/error/ErrorExtension.java:                        name = "master";
src/main/java/org/jboss/as/test/integration/management/extension/error/ErrorExtension.java:                                boolean induceOOME = false; // = context.getProcessType().isServer() && !"master".equals(targetHost.asString());
src/main/java/org/jboss/as/test/integration/management/extension/optypes/OpTypesExtension.java:                    String name = hosts.size() > 1 ? "master": hosts.iterator().next();
src/main/java/org/wildfly/test/jmx/ServiceActivatorDeploymentUtil.java:                new FilePermission(targetPath.resolve("domains").resolve("JmxControlledStateNotificationsTestCase").resolve("master").resolve("target").toAbsolutePath().toString(), "read, write"),
src/main/java/org/wildfly/test/jmx/ServiceActivatorDeploymentUtil.java:                new FilePermission(targetPath.resolve("domains").resolve("JmxControlledStateNotificationsTestCase").resolve("master").resolve("target").toAbsolutePath().toString() + File.separatorChar + '-', "read, write"),
```

parsharma:
@yersan, I have removed ModelDescriptionConstants.MASTER and there are 3 classes using MASTER so I have changed it with String "master".

core-model-test/framework/src/main/java/org/jboss/as/core/model/test/CoreModelTestDelegate.java
```diff
@@ -200,7 +199,7 @@ private void validateDescriptionProviders(TestModelType type, KernelServices ker
             //Big big hack to get around the fact that the tests install the host description twice
             //we're only interested in the host model anyway
             //See KnownIssuesValidator.createHostPlatformMBeanAddress
-            model = model.require(CHILDREN).require(HOST).require(MODEL_DESCRIPTION).require(MASTER);
+            model = model.require(CHILDREN).require(HOST).require(MODEL_DESCRIPTION).require("master");
```
yersan, line 202: It looks like you missed my [comment](https://github.com/wildfly/wildfly-core/pull/5108#pullrequestreview-1044858446);  this should be replaced by "primary"

core-model-test/tests/src/test/java/org/jboss/as/core/model/test/host/ServerConfigTestCase.java
```diff
@@ -54,7 +53,7 @@ public void testAddServerConfigBadSocketBindingGroup() throws Exception {
 
         KernelServices kernelServices = createKernelServices("host.xml");
 
-        PathAddress pa = PathAddress.pathAddress(PathElement.pathElement(HOST, MASTER), PathElement.pathElement(SERVER_CONFIG, "server-four"));
+        PathAddress pa = PathAddress.pathAddress(PathElement.pathElement(HOST, "master"), PathElement.pathElement(SERVER_CONFIG, "server-four"));
```
yersan, line 56: It looks like you missed my [comment](https://github.com/wildfly/wildfly-core/pull/5108#pullrequestreview-1044858446);  this and the others on the same class should be replaced by "primary"

host-controller/src/main/java/org/jboss/as/host/controller/discovery/S3Discovery.java
```diff
@@ -130,7 +129,7 @@ public int compare(DomainControllerData data, DomainControllerData otherData) {
                     }
                     return -1;
                 }});
-            writeToFile(data, MASTER);
+            writeToFile(data, "master");
```
yersan, line 132: This one is OK, we will handle S3Discovery.java in a different Jira.

yersan (changes requested):
Hi @parsharma , would you mind reviewing [my last comment](https://github.com/wildfly/wildfly-core/pull/5108#pullrequestreview-1044858446)? it seems you missed it. 

Let me know when it's ready to review; in addition, check if your custom integration Jobs are passing after applying the changes, thanks

parsharma:
> CoreModelTestDelegate
> ServerConfigTestCase

These two classes are not the part of test suite so I think we should not change it to primary.

parsharma:
@yersan PR is ready for review. Only two test cases are failing in the integration jobs [CI Job](https://ci.wildfly.org/buildConfiguration/WildFlyCore_PullRequest_WildFlyCoreFullIntegrationLinuxJdk11_2/316451?hideProblemsFromDependencies=false&hideTestsFromDependencies=false&expandBuildChangesSection=true&expandBuildProblemsSection=true)

yersan:
@parsharma The integration job failures need to be corrected on the WildFly full PR.

parsharma:
Wildfly PR: https://github.com/wildfly/wildfly/pull/15613

yersan (changes requested):
> @yersan
> 
> > CoreModelTestDelegate
> > ServerConfigTestCase
> 
> These two classes are not the part of test suite so I think we should not change it to primary.

These tests are part of `wildfly-core-model-test-framework`, which is used in the testsuite (both in WildFly Full and Core) we have to change them together with this PR to avoid the need of doing another integration between changes in WildFly Core that requires changes in full.

yersan (approved):
LGTM, I'm just only waiting to see the result of an integration Job in Windows. 
@parsharma would you mind adding the Jira number to all the commits? Having a Jira number there is very useful when you are looking for the reason for a change in the GitHub history. 

There is no need to trigger the CI Integration jobs again.

Linux: https://ci.wildfly.org/viewLog.html?buildId=324460
Windows: https://ci.wildfly.org/viewLog.html?buildId=324659

parsharma:
@yersan changes are done. Thanks

PR #5109, opened 2022-05-24
[WFCORE-5920] fix inconsistency in JDK17 JAVA_OPTS from domain.bat an…

soul2zimate:
…d standalone.ps1 scripts

Issue: https://issues.redhat.com/browse/WFCORE-5920

Please check more information about problem and fix inside the issue.

### DIFF
--- a/core-feature-pack/common/src/main/resources/content/bin/domain.bat
+++ b/core-feature-pack/common/src/main/resources/content/bin/domain.bat
@@ -104,8 +104,14 @@ set "PROCESS_CONTROLLER_JAVA_OPTS=!PROCESS_CONTROLLER_JAVA_OPTS! !DEFAULT_MODULA
 call "!DIRNAME!common.bat" :setDefaultModularJvmOptions "!HOST_CONTROLLER_JAVA_OPTS!"
 set "HOST_CONTROLLER_JAVA_OPTS=!HOST_CONTROLLER_JAVA_OPTS! !DEFAULT_MODULAR_JVM_OPTIONS!"
 
+rem If the -Djava.security.manager is found, enable the -secmgr and include a bogus security manager for JBoss Modules to replace
+echo(!PROCESS_CONTROLLER_JAVA_OPTS! | findstr /r /c:"-Djava.security.manager" > nul && (
+    echo "ERROR: Support for using -Djava.security.manager has been removed. Please use -secmgr or set the environment variable SECMGR=true"
+    GOTO :EOF
+)
+
 rem Set default Security Manager configuration value
-call "!DIRNAME!common.bat" :setSecurityManagerDefault"
+call "!DIRNAME!common.bat" :setSecurityManagerDefault
 set "PROCESS_CONTROLLER_JAVA_OPTS=!PROCESS_CONTROLLER_JAVA_OPTS! !SECURITY_MANAGER_CONFIG_OPTION!"
 set "HOST_CONTROLLER_JAVA_OPTS=!HOST_CONTROLLER_JAVA_OPTS! !SECURITY_MANAGER_CONFIG_OPTION!"
 
@@ -180,11 +186,6 @@ for %%a in (!CONSOLIDATED_OPTS!) do (
    )
 )
 
-rem If the -Djava.security.manager is found, enable the -secmgr and include a bogus security manager for JBoss Modules to replace
-echo(!PROCESS_CONTROLLER_JAVA_OPTS! | findstr /r /c:"-Djava.security.manager" > nul && (
-  echo "ERROR: Support for using -Djava.security.manager has been removed. Please use -secmgr or set the environment variable SECMGR=true"
-  GOTO :EOF
-)
 setlocal DisableDelayedExpansion
 
 rem Setup JBoss specific properties
--- a/core-feature-pack/common/src/main/resources/content/bin/common.ps1
+++ b/core-feature-pack/common/src/main/resources/content/bin/common.ps1
@@ -113,7 +113,7 @@ Function Get-Java-Opts {
 
 Function SetEnhancedSecurityManager {
     $ENHANCED_SM = $false
-    & $JAVA -Djava.security.manager=allow -version >$null 2>&1
+    & $JAVA "-Djava.security.manager=allow" -version >$null 2>&1
     if ($LastExitCode -eq 0){
         $ENHANCED_SM = $true
     }


core-feature-pack/common/src/main/resources/content/bin/domain.bat
```diff
@@ -182,8 +182,11 @@ for %%a in (!CONSOLIDATED_OPTS!) do (
 
 rem If the -Djava.security.manager is found, enable the -secmgr and include a bogus security manager for JBoss Modules to replace
 echo(!PROCESS_CONTROLLER_JAVA_OPTS! | findstr /r /c:"-Djava.security.manager" > nul && (
-  echo "ERROR: Support for using -Djava.security.manager has been removed. Please use -secmgr or set the environment variable SECMGR=true"
-  GOTO :EOF
+  echo !PROCESS_CONTROLLER_JAVA_OPTS! | findstr /r /c:"-Djava.security.manager=allow" > nul
```
bstansberry, line 185: @soul2zimate Are lines 184 and 190 now redundant?

soul2zimate: It's not redundant. I believe we still want to fail the script when `-Djava.security.manager` was found, and only skip the case of `-Djava.security.manager=allow`

The same logic is also applied in https://github.com/wildfly/wildfly-core/blob/19.0.0.Beta10/core-feature-pack/common/src/main/resources/content/bin/standalone.sh#L308-L316


I can also keep that part intact and move it up just before setting default Security Manager configuration value at line 107, that means something like:

```
...

rem If the -Djava.security.manager is found, enable the -secmgr and include a bogus security manager for JBoss Modules to replace
echo(!PROCESS_CONTROLLER_JAVA_OPTS! | findstr /r /c:"-Djava.security.manager" > nul && (
  echo "ERROR: Support for using -Djava.security.manager has been removed. Please use -secmgr or set the environment variable SECMGR=true"
  GOTO :EOF
)

rem Set default Security Manager configuration value
call "!DIRNAME!common.bat" :setSecurityManagerDefault
set "PROCESS_CONTROLLER_JAVA_OPTS=!PROCESS_CONTROLLER_JAVA_OPTS! !SECURITY_MANAGER_CONFIG_OPTION!"
set "HOST_CONTROLLER_JAVA_OPTS=!HOST_CONTROLLER_JAVA_OPTS! !SECURITY_MANAGER_CONFIG_OPTION!"
...
```

yersan: Hi @soul2zimate in that case, is not the if condition wrong? I mean, both `findstr` commands are looking for strings and finally it checks the error level; 0 if the string is found, to 1 if it is not. Would not be the final error level = 1 if `-Djava.security.manager=allow` is found?

Based on your feedback if `-Djava.security.manager=allow` we should not raise an error.

soul2zimate: I keep that codes but move them up before setting default Security Manager configuration value.
It at that point only checks if `-Djava.security.manager` is used by user. I think this give better readability.

yersan: Cannot the user add `-Djava.security.manager=allow`? in that case, it will go to error. The Linux variant will continue in that case.

soul2zimate: For Linux variant, you mean `domain.sh` ?  It will fail if user define `-Djava.security.manager=allow`

yersan: Sorry @soul2zimate , when I wrote "linux variant" I had in my mind the `standalone.sh` example you pasted before.

In any case, I think your change still needs to allow `-Djava.security.manager=allow` on `HOST_CONTROLLER_JAVA_OPTS` and reject `-Djava.security.manager`.  Right no my understanding is it is rejecting both.

As a more widely feedback, it looks like we are also failing on domain mode when this setting is configured by using `JAVA_OPTS`. In that case, I would expect a similar behavior; we should allow `-Djava.security.manager=allow` but fail when `-Djava.security.manager` is used.

And even in a more general view, I guess we should only accept `-Djava.security.manager=allow` if the JDK allows it, and reject it otherwise, but it looks like we have never implemented this kind of logic. This could be an optional improvement.

Passing `-Djava.security.manager=allow` or `-Djava.security.manager` directly to the `domain.sh|bat|ps1` or `standalone.sh|bat|ps1` scripts arealways rejected, and that sounds fine because we are forcing the users to use `-secmgr` in that case, which is a setting passed directly to jboss-modules who is going to take care of it. 

WDYT?

yersan: Sorry @soul2zimate , when I wrote "linux variant" I had in my mind the `standalone.sh` example you pasted before.

In any case, I think your change still needs to allow `-Djava.security.manager=allow` on `HOST_CONTROLLER_JAVA_OPTS` and reject `-Djava.security.manager`.  Right no my understanding is it is rejecting both.

As a more widely feedback, it looks like we are also failing on domain mode when this setting is configured by using `JAVA_OPTS`. In that case, I would expect a similar behavior; we should allow `-Djava.security.manager=allow` but fail when `-Djava.security.manager` is used.

And even in a more general view, I guess we should only accept `-Djava.security.manager=allow` if the JDK allows it, and reject it otherwise, but it looks like we have never implemented this kind of logic. This could be an optional improvement.

Passing `-Djava.security.manager=allow` or `-Djava.security.manager` directly to the `domain.sh|bat|ps1` or `standalone.sh|bat|ps1` scripts arealways rejected, and that sounds fine because we are forcing the users to use `-secmgr` in that case, which is a setting passed directly to jboss-modules who is going to take care of it. 

WDYT?

yersan: @soul2zimate ,  my mistake:

> In any case, I think your change still needs to allow -Djava.security.manager=allow on HOST_CONTROLLER_JAVA_OPTS and reject -Djava.security.manager. Right now my understanding is it is rejecting both.

Forget the above, you are controlling the security manager setting for the `PROCESS_CONTROLLER_JAVA_OPTS` and not for the `HOST_CONTROLLER_JAVA_OPTS`. It makes sense it is always rejected for the `PROCESS_CONTROLLER_JAVA_OPTS`

yersan: > As a more widely feedback, it looks like we are also failing on domain mode when this setting is configured by using JAVA_OPTS. In that case, I would expect a similar behavior; we should allow -Djava.security.manager=allow but fail when -Djava.security.manager is used.

This is also incorrect, any JAVA_OPTS configured are passed to the PROCESS_CONTROLLER_JAVA_OPTS and  HOST_CONTROLLER_JAVA_OPTS and it looks like they are well controlled there.

yersan (approved):
@soul2zimate I'm sorry for all the back and forth, changes look good to me, however, it would be great to have another review.

core-feature-pack/common/src/main/resources/content/bin/domain.bat
soul2zimate: >As a more widely feedback, it looks like we are also failing on domain mode when this setting is configured by using `JAVA_OPTS`. In that case, I would expect a similar behavior; we should allow `-Djava.security.manager=allow` but fail when `-Djava.security.manager is used`.

@yersan Thank you for the feedback. `-Djava.security.manager=allow` is rejected because found in `PROCESS_CONTROLLER_JAVA_OPTS` shared from `JAVA_OPTS`.

>And even in a more general view, I guess we should only accept -Djava.security.manager=allow if the JDK allows it, and reject it otherwise, but it looks like we have never implemented this kind of logic. This could be an optional improvement

The JVM will cause an error anyway. i.e. passing -Djava.security.manager=allow in JDK 11 will fail with:

```
chaowan@Chaos-MacBook-Pro wildfly-core % java -Djava.security.manager=allow -version
Error occurred during initialization of VM
java.lang.Error: Could not create SecurityManager
        at java.lang.System.initPhase3(java.base@11.0.15/System.java:2065)
Caused by: java.lang.ClassNotFoundException: allow
        at jdk.internal.loader.BuiltinClassLoader.loadClass(java.base@11.0.15/BuiltinClassLoader.java:581)
        at jdk.internal.loader.ClassLoaders$AppClassLoader.loadClass(java.base@11.0.15/ClassLoaders.java:178)
        at java.lang.ClassLoader.loadClass(java.base@11.0.15/ClassLoader.java:522)
        at java.lang.Class.forName0(java.base@11.0.15/Native Method)
        at java.lang.Class.forName(java.base@11.0.15/Class.java:398)
        at java.lang.System.initPhase3(java.base@11.0.15/System.java:2050)
```

Hi @ropalka, would you mind to review this? Thanks.

yersan: > The JVM will cause an error anyway. i.e. passing -Djava.security.manager=allow in JDK 11

Right, as I commented before, It would be more friendly if this configuration were also rejected if the JDK doesn't support it, it was just a suggestion. We can live with it since it is only valid for JDK17 and above

jamezp (approved):
This looks good to me now too.

PR #5111, opened 2022-05-25
Fix for WFCORE-5911, Upgrade to Galleon 5.0.1.Final and Galleon plugins 6.0.0.Alpha2

jfdenise:
Issue: https://issues.redhat.com/browse/WFCORE-5911


yersan (approved):

yersan:
Added @jmesnil to the review since I still have a minor doubt, are we allowed to use Alpha releases?

yersan:
@jmesnil , ignore this. I've just found WildFly's full counterpart was merged by using an Alpha, which resolves my doubt.

PR #5113, opened 2022-05-26
[WFCORE-785] Improving capability related error messages

khermano:
Issue: https://issues.redhat.com/browse/WFCORE-785

### DIFF
--- a/controller/src/main/java/org/jboss/as/controller/CapabilityRegistry.java
+++ b/controller/src/main/java/org/jboss/as/controller/CapabilityRegistry.java
@@ -943,11 +943,24 @@ public final class CapabilityRegistry implements ImmutableCapabilityRegistry, Po
         if (satisfactoryCapability == null) {
             if (forServer) {
                 throw ControllerLogger.MGMT_OP_LOGGER.unknownCapability(capabilityName);
+            }
+            if (getRegistrationPoints(capabilityName).isEmpty()) {
+                throw ControllerLogger.MGMT_OP_LOGGER.unknownCapability(capabilityName);
             } else {
-                throw ControllerLogger.MGMT_OP_LOGGER.unknownCapabilityInContext(capabilityName, capabilityScope.getName());
+                throw ControllerLogger.MGMT_OP_LOGGER.noSatisfactoryCapability(capabilityName, capabilityScope.getName(), getRegistrationPoints(capabilityName), capabilityScope.getName());
             }
         }
         return capabilities.get(satisfactoryCapability.singleCapability);
+    }
+
+    private Set<RegistrationPoint> getRegistrationPoints(String capabilityName) {
+        Set<RegistrationPoint> registrationPoints = new HashSet<>();
+        for (CapabilityId key: capabilities.keySet()) {
+            if (key.getName().equals(capabilityName)) {
+                registrationPoints = capabilities.get(key).getRegistrationPoints();
+            }
+        }
+        return registrationPoints;
     }
 
     private SatisfactoryCapability findSatisfactoryCapability(String capabilityName, CapabilityScope dependentContext,
--- a/controller/src/main/java/org/jboss/as/controller/logging/ControllerLogger.java
+++ b/controller/src/main/java/org/jboss/as/controller/logging/ControllerLogger.java
@@ -3736,4 +3736,7 @@ public interface ControllerLogger extends BasicLogger {
      */
     @Message(id = Message.NONE, value = "The yaml configuration files for customizing the configuration. Paths can be absolute, relative to the current execution directory or relative to the standalone configuration directory.")
     String argYaml();
+
+    @Message(id = 499, value = "There is no satisfactory capability '%s' available to resources with capability scope '%s'. This capability is registered at address(es) '%s', and are not accessible to resources with scope '%s'.")
+    IllegalStateException noSatisfactoryCapability(String capability, String scopeName, Set<RegistrationPoint> addresses, String scopeNameAgain);
 }


khermano:
Hi @yersan, sure, and thank you for the feedback.

controller/src/main/java/org/jboss/as/controller/CapabilityRegistry.java
```diff
@@ -940,16 +940,33 @@ private static Set<RuntimeRequirementRegistration> findInconsistent(Map<Capabili
...
         SatisfactoryCapability satisfactoryCapability = findSatisfactoryCapability(capabilityName, capabilityScope, false);
+
         if (satisfactoryCapability == null) {
             if (forServer) {
-                throw ControllerLogger.MGMT_OP_LOGGER.unknownCapability(capabilityName);
+                if (isInCapability(capabilityName)) {
+                    throw ControllerLogger.MGMT_OP_LOGGER.cannotGetCapability(capabilityName);
+                } else {
+                    throw ControllerLogger.MGMT_OP_LOGGER.missingCapability(capabilityName);
+                }
```
bstansberry, line 950: This if (forServer) block should just be the existing code, unchanged.

For the 'forServer' case, findSatisfactoryCapability is just a simple lookup. There's no difference between 'capability exists' and 'capability is satisfactory'. The existing message covers this.

controller/src/main/java/org/jboss/as/controller/CapabilityRegistry.java
```diff
@@ -940,16 +940,33 @@ private static Set<RuntimeRequirementRegistration> findInconsistent(Map<Capabili
...
-                throw ControllerLogger.MGMT_OP_LOGGER.unknownCapabilityInContext(capabilityName, capabilityScope.getName());
+                throw ControllerLogger.MGMT_OP_LOGGER.missingCapabilityInContext(capabilityScope.getName(), capabilityName);
             }
         }
         return capabilities.get(satisfactoryCapability.singleCapability);
     }
 
+    private boolean isInCapability(String capabilityName) {
+        for (CapabilityId key: capabilities.keySet()) {
+            if (capabilities.get(key).getCapabilityName().equals(capabilityName)) {
```
bstansberry, line 963: This should just be 

if (key.getCapabilityName().equals(capabilityName)) {

It's the same thing.

Note though that if it were different, you should either iterate over capabilities.values() or capabilities.entrySet(). Getting a key set and then calling get is less efficient.

controller/src/main/java/org/jboss/as/controller/logging/ControllerLogger.java
```diff
@@ -3209,9 +3209,23 @@ XMLStreamException discoveryOptionsMustBeDeclared(String adminOnlyCmd, String po
     @Message(id = 364, value = "Capability '%s' is unknown.")
     IllegalStateException unknownCapability(String capability);
 
+    @Message(id = NONE, value = "It was not possible to get satisfactory capability for capability '%s' even when " +
```
bstansberry, line 3212: This would not be NONE; it's a different message for a different call path and should have different id. Same for the others.

controller/src/main/java/org/jboss/as/controller/CapabilityRegistry.java
```diff
@@ -940,16 +940,33 @@ private static Set<RuntimeRequirementRegistration> findInconsistent(Map<Capabili
...
+                } else {
+                    throw ControllerLogger.MGMT_OP_LOGGER.missingCapability(capabilityName);
+                }
+            }
+            if (isInCapability(capabilityName)) {
+                throw ControllerLogger.MGMT_OP_LOGGER.cannotGetCapabilityInContext(capabilityScope.getName(), capabilityName);
             } else {
-                throw ControllerLogger.MGMT_OP_LOGGER.unknownCapabilityInContext(capabilityName, capabilityScope.getName());
+                throw ControllerLogger.MGMT_OP_LOGGER.missingCapabilityInContext(capabilityScope.getName(), capabilityName);
             }
```
bstansberry, line 956: This if / else is the area where the current information can be improved.

If findSatisfactoryCapability returns null, either there is a CapabilityId in capabilities with the requested name, or there isn't.

If there isn't, 'throw ControllerLogger.MGMT_OP_LOGGER.unknownCapability(capabilityName);' is a better message than the existing 'unknownCapabilityInContext' message. I think it's better than your 'missingCapabilityInContext' as well. Because it's simple.

If there is a CapabilityId with the requested name, then the problem is the cap is registered in a part of the model that the dependent is not allowed to access. So we can improve this message. The whole 'capabilityScope' thing is an internal detail that's not really understandable to the user. So we can provide some info that is understandable.  What's more understandable are the RegistrationPoint instances that are inside the RuntimeCapabilityRegistration values in the capabilities map, as those are basically just resource addresses.

If instead of a 'isInCapabilities' that returns a boolean, you instead had a getRegistrationPoints that returns a Set<RegistrationPoint>, then if the set isn't empty you use its content to construct an exception message that includes those points.

```
@Message(id = ###, value = "There is no satisfactory capability '%s' available to resources with capability scope '%s'. This capability is registered at address(es) '%s', and are not accessible to resources with scope '%s'.")
    IllegalStateException noSatisfactoryCapability(String capability, String scopeName, Set<RegistrationPoint> addresses, String scopeNameAgain);
```

bstansberry (changes requested):

controller/src/main/java/org/jboss/as/controller/CapabilityRegistry.java
khermano: Thank you very much for your feedback. I force-pushed the changes you suggest.
Also, if I may ask regarding the WFCORE-785 comments, I want to ask if it's okay for these changes to stay linked with this issue or if I should create a new one.

khermano:
Hello @bstansberry,
Were you able to take a look at my changes, by chance? :)

khermano:
Hello @bstansberry  ,
I hate to interrupt you, but did you have a chance to look at this? :)

bstansberry (approved):

bstansberry:
Thanks, @khermano -- nice improvement!  Sorry for neglecting this!

khermano:
@bstansberry that's okay, thank you for the feedback.

PR #5115, opened 2022-05-27
WFCORE-5927 Misleading message for embedded server configuration file.

RanabirChakraborty:
Issue: https://issues.redhat.com/browse/WFCORE-5927

### DIFF
--- a/controller/src/main/java/org/jboss/as/controller/logging/ControllerLogger.java
+++ b/controller/src/main/java/org/jboss/as/controller/logging/ControllerLogger.java
@@ -2438,10 +2438,11 @@ public interface ControllerLogger extends BasicLogger {
      *
      * @return an {@link IllegalStateException} for the error.
      */
-    @Message(id = 214, value = "Could not get main file: %s. Specified files must be relative to the configuration dir: %s")
+    @Message(id = 214, value = "Could not load configuration file: %s. The configuration file argument must specify the path to a file located in the configuration directory. The path must be a relative path, and must be relative to the configuration directory %s.")
     IllegalStateException mainFileNotFound(String name, File configurationDir);
 
-    // 215 free
+    @Message(id = 215, value = "Could not load configuration file: %s. The configuration file argument must specify one of the following: 1) an absolute path to an existing file; 2) a relative path to an existing file, relative to the current working directory; or 3) a relative path to a file located in the configuration directory. In the latter case, it must be a path relative to the configuration directory %s.")
+    IllegalStateException absolutePathMainFileNotFound(String name, File configurationDir);
 
     /**
      * Creates an exception indicating a resource cannot be found.
--- a/controller/src/main/java/org/jboss/as/controller/persistence/ConfigurationFile.java
+++ b/controller/src/main/java/org/jboss/as/controller/persistence/ConfigurationFile.java
@@ -379,6 +379,10 @@ public class ConfigurationFile {
             }
         }
 
+        if(interactionPolicy.isReadOnly()){
+            throw ControllerLogger.ROOT_LOGGER.absolutePathMainFileNotFound(name, configurationDir);
+        }
+
         throw ControllerLogger.ROOT_LOGGER.mainFileNotFound(name != null ? name : rawName, configurationDir);
     }
 


controller/src/main/java/org/jboss/as/controller/logging/ControllerLogger.java
```diff
@@ -2438,7 +2438,7 @@ public interface ControllerLogger extends BasicLogger {
      *
      * @return an {@link IllegalStateException} for the error.
      */
-    @Message(id = 214, value = "Could not get main file: %s. Specified files must be relative to the configuration dir: %s")
+    @Message(id = 214, value = "Could not get main file: %s. the configuration dir: %s could be set by the property jboss.server.config.dir during jboss-cli start.")
```
jbliznak, line 2441: Capitalize the first character of the sentence? `the configuration ... `
Is this really limited to `jboss-cli` only? I thought it applies for server start in general.

Anyway, IMHO I would still find this misleading, maybe even more than the previous message which at least tries to tell what is wrong. 

I think the main focus of the message needs to be that `only files directly in configuration dir can be loaded` (I hope I understood the changed implementation correctly).
Of course you can also add the additional info how to change the configuration dir, as an enhancement to the main message.

RanabirChakraborty: Thanks! @jbliznak for the review. I agree with your point but the thing is users already getting the idea as the relative path which was provided by `Could not get main file: %s.` If we can provide a solution to them, that will be a good idea.
maybe we can use this generalized logger message instead:
"Could not get main file: %s. Files directly in configuration dir: %s can be loaded which could be set by the property jboss.server.config.dir"

jbliznak: > users already getting the idea as the relative path which was provided by Could not get main file: %s.

I don't think the users are getting the idea what is wrong:
```
./standalone.sh --server-config=../../docs/examples/configs/standalone-jts.xml

WFLYCTL0214: Could not get main file: ../../docs/examples/configs/standalone-jts.xml. Specified files must be relative to the configuration dir: /tmp/wildfly-26.1.1.Final/standalone/configuration
```
Given the original message I am lost because the `../../docs/examples/configs/standalone-jts.xml` is a valid relative path to the configuration dir pointing to existing configuration so everything seems to be ok. (that was the reason for the jira I guess)
The whole point is that the implementation changed and doesn't allow files outside of configuration dir to be loaded anymore so the message stopped to be correct (it was correct before). So we must tell a user that this is the case.

My proposals would be one of these:
```
// adding the keyword `only` so that the restriction is obvious
@Message(id = 214, value = "Could not get main file: %s. Only files from the configuration dir %s could be loaded.")

// adding also possible solution as you wanted
@Message(id = 214, value = "Could not get main file: %s. Only files from the configuration dir %s could be loaded. Configuration dir location can be changed by setting the property jboss.server.config.dir.")
```

BTW isn't there some constant for `jboss.server.config.dir` to make it safe? I don't know if that is required though.

RanabirChakraborty: Thanks, Got your point and I'm going with your 1st suggestion.
About the constant for `jboss.server.config.dir` I haven't looked into that much as it was not mentioned in Jira. will take a look.

yersan (commented):
If you need it, these constants are available at [ServerEnvironment.java#L176](https://github.com/wildfly/wildfly-core/blob/52f62b95acb8c9dccacf10a4965f2689a80ba20b/server/src/main/java/org/jboss/as/server/ServerEnvironment.java#L176)

I'm fine with the current changes, so I don't see any additional action. However, since the other option gives more information and sounds more aligned with the original issue, let's ask @dcihak for it.

Daniel, would the current message resolve your initial motivation? If so, we can go ahead since all of us are fine with the changes.

The alternative option is:
`@Message(id = 214, value = "Could not get main file: %s. Only files from the configuration dir %s could be loaded. Configuration dir location can be changed by setting the property jboss.server.config.dir.")`

bstansberry:
Removing discussion of the relative nature of the file name should not be done.  The argument is not an absolute path; using an absolute path that ends up pointing to a file that's in the configuration dir does not work.

```
$ cd /Users/bstansberry/dev/wildfly/wildfly/dist/target/wildfly-27.0.0.Alpha2-SNAPSHOT
yingtao:wildfly-27.0.0.Alpha2-SNAPSHOT 
$ bin/jboss-cli.sh 
You are disconnected at the moment. Type 'connect' to connect to the server or 'help' for the list of supported commands.
[disconnected /] embed-server -c=/Users/bstansberry/dev/wildfly/wildfly/dist/target/wildfly-27.0.0.Alpha2-SNAPSHOT/standalone/configuration/standalone.xml
Cannot start embedded server: WFLYEMB0022: Cannot invoke 'start' on embedded process: WFLYCTL0214: Could not get main file: /Users/bstansberry/dev/wildfly/wildfly/di
st/target/wildfly-27.0.0.Alpha2-SNAPSHOT/standalone/configuration/standalone.xml. Specified files must be relative to the configuration dir: /Users/bstansberry/dev/wildfly/wild
fly/dist/target/wildfly-27.0.0.Alpha2-SNAPSHOT/standalone/configuration
```

There are really two cases here, depending on whether ConfigurationFile.interactionPolicy.isReadOnly() returns true.

In the typical case where that is false:

Could not get main file: %s. The configuration file argument must specify the relative path to a file located in the configuration directory. The must be path relative to the configuration directory %s

Where it is true:

Could not get main file: %s. The configuration file argument must specify either an absolute path to an existing file, a relative path to an existing file, relative to the current working directory, or the relative path to a file located in the configuration directory. In the latter case the must be path relative to the configuration directory %s.

jbliznak:
So you propose to introduce new log message to distinguish the cases, right? No problem on my side with that. However, I would mention something about readonly mode in the second message to make.

Another argument is, that `The must be path relative to the configuration directory %s` (BTW there is something missing after The) is not clear enough and it is not the same as the statement `Only files from the configuration dir %s could be loaded.` (which is more restrictive, the relative path here == just filename)

Is it also possible to rename the "main file" in `Could not get main file: %s.` to something more user-facing? Like `Could not get configuration file: %s.`
User doesn't know anything about "main file".

bstansberry:
@jbliznak 

> So you propose to introduce new log message to distinguish the cases, right? No problem on my side with that. However, I would mention something about readonly mode in the second message to make it clear.

Yes, a new message. The way the code is now if the user does --read-only-server-config=/home/bstansberry/stnadalone.xml (a typo) they'd get a message incorrectly telling them they couldn't have used /home/bstansberry/standalone.xml.

> 
> Another argument is, that `The must be path relative to the configuration directory %s` (BTW there is something missing after The) is not clear enough and it is not the same as the statement `Only files from the configuration dir %s could be loaded.` (which is more restrictive, the relative path here == just filename)
> 

Oops; I misordered the words. s/The must be path relative/The path must be relative/g

'Only files from the configuration dir %s could be loaded' is true but does not tell the user how to construct a valid argument. Which is ok, I suppose.  If we go that way, then I think "can be used" is better than "could be loaded." Definitely "can" not "could", as "could" implies the process did load some files.

Note also that if you create dir standalone/configuration/foo and put file bar.xml in it, the -c=foo/bar.xml will work.

> Is it also possible to rename the "main file" in `Could not get main file: %s.` to something more user-facing? Like `Could not get configuration file: %s.` User doesn't know anything about "main file".

+1.  I think 'find' is better than 'get' as well. "Could not find configuration file %s"

jbliznak:
> Note also that if you create dir standalone/configuration/foo and put file bar.xml in it, the -c=foo/bar.xml will work.

OK, that is new, now I see why you phrased it like that initially. However, 'relative to' still seems to allow more than what we need (eg. `../../../standalone.xml` is also relative to configuration directory but not allowed because it leads outside configuration directory). 

TBH I don't know how to phrase that correctly and simply enough at the same time in English, I just wanted not to replace one unclear message with another unclear message.

> +1. I think 'find' is better than 'get' as well. "Could not find configuration file %s"

```
Could not find configuration file: /Users/bstansberry/dev/wildfly/wildfly/dist/target/wildfly-27.0.0.Alpha2-SNAPSHOT/standalone/configuration/standalone.xml. Specified files must be relative to the configuration dir: /Users/bstansberry/dev/wildfly/wildfly/dist/target/wildfly-27.0.0.Alpha2-SNAPSHOT/standalone/configuration
```
I don't know if "find" is better here when we know that file exists - "get" may at least implicate there is some problem with reading/accessing it. What about "load" ?

jbliznak:
@bstansberry your proposal is much better than current state so I would be ok to go with it. 

If there is a chance to enhance it more, I would welcome some notion we are in readonly scenario in the 'true' scenario, however, that is probably not mandatory.

controller/src/main/java/org/jboss/as/controller/persistence/ConfigurationFile.java
```diff
@@ -379,6 +379,10 @@ private File determineMainFile(final String rawName, final String name) {
             }
         }
 
+        if(interactionPolicy.isReadOnly()){
+            throw ControllerLogger.ROOT_LOGGER.absolutePathMainFileNotFound(name, configurationDir);
```
yersan, line 383: Looking at L386 it seems we could arrive at this point with name == null. Looking at the method code I've been unable to determine if the fact of being interaction policy as read-only can prevent this situation.

@RanabirChakraborty Just to confirm, are you sure you don't need to check the variable `name` nullity on this line as it is done on L386?

RanabirChakraborty: @yersan I think there is no need to put a check for the variable `name` nullity, as `name` is always going to be not null. 
And about the read-only scenario as Brian stated I think it'll be good as the specified file will not get updated.

controller/src/main/java/org/jboss/as/controller/logging/ControllerLogger.java
```diff
@@ -2438,10 +2438,11 @@ public interface ControllerLogger extends BasicLogger {
      *
      * @return an {@link IllegalStateException} for the error.
      */
-    @Message(id = 214, value = "Could not get main file: %s. Specified files must be relative to the configuration dir: %s")
+    @Message(id = 214, value = "Could not load configuration file: %s. The configuration file argument must specify the path to a file located in the configuration directory. The path must be a relative path, and must be relative to the configuration directory %s.")
     IllegalStateException mainFileNotFound(String name, File configurationDir);
 
-    // 215 free
+    @Message(id = 215, value = "Could not load configuration file: %s. The configuration file argument must specify one of the following: 1) an absolute path to an existing file; 2) a relative path to an existing file, relative to the current working directory; or 3) a relative path to a file located in the configuration directory. In the latter case the must be path relative to the configuration directory %s.")
```
yersan, line 2444: s/In the latter case the must be path relative to the configuration directory %s./In the latter case, it must be a path relative to the configuration directory %s.

yersan (changes requested):
Thanks @RanabirChakraborty , I've realized there was a minor miswording on the error message, besides that, it looks good to me.

yersan (approved):

yersan:
Thanks to all of you involved in this issue and its review!

PR #5031, opened 2022-03-25
[WFCORE-5844] [primary/secondary] Rename Host Controller model resour…

moulalis:
Issue: https://issues.redhat.com/browse/WFCORE-5844

PR #5116, opened 2022-05-27
[WFCORE-5846][primary/secondary] Adapt S3Discovery mechanism to new l…

yersan:
…anguage usage

Build on top of https://github.com/wildfly/wildfly-core/pull/5031 to avoid conflicts.

It replaces language usage on the S3Discovery discovery option.


yersan:
Closed for further investigation since changes on the discovery option should take into account mixed domains.

PR #5137, opened 2022-06-22
[WFLY-5950] Update client-side modular args to standard set that we s…

bstansberry:
…uggest user's use.

Only use -add-opens=java.base/sun.security.ssl=ALL-UNNAMED in the elytron module which has a test that requires it.

https://issues.redhat.com/browse/WFCORE-5950

### DIFF
--- a/elytron/pom.xml
+++ b/elytron/pom.xml
@@ -30,6 +30,14 @@
 
     <name>WildFly: Elytron Subsystem</name>
 
+    <properties>
+        <!-- TlsTestCase specifies use of sun.security.ssl.SunJSSE in the server config,
+             so make it available. -->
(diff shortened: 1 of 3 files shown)


bstansberry:
TlsTestCase passes for me locally on SE 17 with this change. I don't get why it fails here.

yersan (commented):
@bstansberry the errors are logged for the `wildfly-elytron-integration-jakarta` module, it looks like you need to fix the properties for that module too, look at https://github.com/wildfly/wildfly-core/blob/1e12e2b79954ecc17b931d172f96c59e58c8019f/jakartaee/elytron-jakarta/pom.xml#L289-L296

yersan (approved):

PR #5139, opened 2022-06-23
WFCORE-5953/WFCORE-5954: Fixing several YAML extension issues

ehsavoie:
Two fixes for YAML Extension: empty resources and list attributes

Jira: https://issues.redhat.com/browse/WFCORE-5953
      https://issues.redhat.com/browse/WFCORE-5954


yersan:
@ehsavoie would you mind backporting this to 18.x too? there are other YAML issues backported there so, if finally, we want to incorporate them we can incorporate all related

PR #5141, opened 2022-06-29
[WFCORE-5956] Align Jakarta Authorization and Authentication versions…

yersan:
… on wildfly-core-jakartaee with the ones used on WildFly Preview

Jira issue: https://issues.redhat.com/browse/WFCORE-5956

### DIFF
--- a/jakartaee/pom.xml
+++ b/jakartaee/pom.xml
@@ -36,10 +36,10 @@
     <packaging>pom</packaging>
 
     <properties>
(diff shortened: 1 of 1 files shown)


darranl (approved):
I am adding an approval but maybe elytrron-ee also could be synced?

yersan:
@darranl Right, yes, good point, I updated it as well to sync it with WildFly

PR #5144, opened 2022-06-30
[WFCORE-5960] Fix bug initializing Domain Base Dir for path Manager

yersan:
jira issue: https://issues.redhat.com/browse/WFCORE-5960

### DIFF
--- a/testsuite/domain/src/test/java/org/jboss/as/test/integration/domain/suites/CoreResourceManagementTestCase.java
+++ b/testsuite/domain/src/test/java/org/jboss/as/test/integration/domain/suites/CoreResourceManagementTestCase.java
@@ -80,6 +80,9 @@ import static org.jboss.as.test.integration.domain.management.util.DomainTestUti
 import static org.hamcrest.MatcherAssert.assertThat;
 
 import java.io.IOException;
+import java.nio.file.Files;
+import java.nio.file.Path;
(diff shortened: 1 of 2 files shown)


yersan:
Let's see first how it goes, it could deserve a test case to ensure this never change for any other reason

yersan:
I haven't found a more simple way to test except by adding a real configuration, I was expecting something around the host path manager service, but seems to be pretty hidden for external accesses.

testsuite/domain/src/test/java/org/jboss/as/test/integration/domain/suites/CoreResourceManagementTestCase.java
```diff
@@ -974,6 +978,36 @@ public void testHostRelativeToDomainBaseDir() throws IOException {
 
     }
 
+    /** WFCORE-5960 */
+    @Test
+    public void testHostRelativeToDomainBaseDirConfiguration() throws Exception {
+        final DomainClient masterClient = domainMasterLifecycleUtil.getDomainClient();
+        final PathAddress hostAddress = PathAddress.pathAddress("host", "master");
```
yersan, line 985: @parsharma FYI, in case it gets merged before your work in progress replacing language usage.

bstansberry (approved):

PR #5148, opened 2022-07-13
[WFCORE-5964] Add test utility methods for checking the test environm…

soul2zimate:
…ent that don't throw AssumptionViolationException but instead return a boolean

Issue: https://issues.redhat.com/browse/WFCORE-5964

This just replicates @bstansberry's work in https://github.com/wildfly/wildfly/pull/15759 to wfcore.


yersan (approved):

bstansberry:
Thanks @soul2zimate!

PR #5149, opened 2022-07-13
[WFCORE-5967] Fix Duplicate Module Path argument on Process Controlle…

yersan:
…r command line launcher

Jira issue: https://issues.redhat.com/browse/WFCORE-5967


bstansberry (changes requested):
@yersan This is contrary to the design of org.jboss.as.process.Main:

https://github.com/wildfly/wildfly-core/blob/main/process-controller/src/main/java/org/jboss/as/process/Main.java#L180-L185

Perhaps that TODO comment is out of date, but passing -mp as an arg to main was intentional. If it is out of date I don't think we should remove it without removing that comment and for the benefit of future maintainers the JIRA should explain why the TODO's concern is no longer valid. If it isn't out of date we shouldn't change DomainCommandBuilder.

yersan:
Hi @bstansberry , you're right, this is an invalid bug, closing it, thanks for the review

PR #5152, opened 2022-07-13
[WFCORE-5968] log thread dump upon management timeout

aogburn:
https://issues.redhat.com/browse/WFCORE-5968

### DIFF
--- a/controller/src/main/java/org/jboss/as/controller/OperationContextImpl.java
+++ b/controller/src/main/java/org/jboss/as/controller/OperationContextImpl.java
@@ -501,6 +501,8 @@ final class OperationContextImpl extends AbstractOperationContext {
                 // Deliberate log and throw; we want to log this but the caller method passes a slightly different
                 // message to the user as part of the operation response
                 MGMT_OP_LOGGER.timeoutExecutingOperation(timeout / 1000, containerMonitorStep.operationId.name, containerMonitorStep.address);
+                // Produce and log thread dump for diagnostics
+                ThreadDumpUtil.threadDump();
                 throw te;
             } finally {
                 executionStatus = originalExecutionStatus;
@@ -868,6 +870,8 @@ final class OperationContextImpl extends AbstractOperationContext {
                     // in AbstractOperationContext.executeStep is not what I wanted
                     ControllerLogger.MGMT_OP_LOGGER.timeoutAwaitingInitialStability(timeout / 1000, activeStep.operationId.name, activeStep.operationId.address);
                     setRollbackOnly();
+                    // Produce and log thread dump for diagnostics
+                    ThreadDumpUtil.threadDump();
                     throw new OperationFailedRuntimeException(ControllerLogger.ROOT_LOGGER.timeoutAwaitingInitialStability());
                 } finally {
                     executionStatus = origStatus;
@@ -1233,6 +1237,8 @@ final class OperationContextImpl extends AbstractOperationContext {
                     // it's almost certain we never stabilized during execution or we are rolling back and destabilized there.
                     // Either one means there is already a failure message associated with this op.
                     MGMT_OP_LOGGER.timeoutCompletingOperation(timeout / 1000, activeStep.operationId.name, activeStep.operationId.address);
+                    // Produce and log thread dump for diagnostics
+                    ThreadDumpUtil.threadDump();
                 }
             }
 
--- a/controller/src/main/java/org/jboss/as/controller/ThreadDumpUtil.java
+++ b/controller/src/main/java/org/jboss/as/controller/ThreadDumpUtil.java
@@ -1,0 +1,129 @@
+/*
+ * JBoss, Home of Professional Open Source.
+ *
+ * Copyright 2022 Red Hat, Inc., and individual contributors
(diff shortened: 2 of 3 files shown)


controller/src/main/java/org/jboss/as/controller/ThreadDumpUtil.java
```diff
@@ -0,0 +1,142 @@
...
+import java.io.IOException;
+import java.io.PrintWriter;
+import java.io.StringWriter;
+import java.lang.management.LockInfo;
+import java.lang.management.ManagementFactory;
+import java.lang.management.MonitorInfo;
+import java.lang.management.ThreadInfo;
+import java.lang.management.ThreadMXBean;
+
+public final class ThreadDumpUtil {
```
bstansberry, line 36: This class needs to be package protected.

controller/src/main/java/org/jboss/as/controller/ThreadDumpUtil.java
```diff
@@ -0,0 +1,142 @@
...
+
+      try (
+         StringWriter str = new StringWriter();
+         PrintWriter out = new PrintWriter(str)
+      ) {
+
+         ThreadMXBean threadMXBean = ManagementFactory.getThreadMXBean();
+
+         out.println("*******************************************************************************");
+         out.println("Complete Thread dump " + msg);
```
bstansberry, line 48: This and the other text message in this class need to be i18n via ControllerLogger methods.

bstansberry (changes requested):
Thanks, @aogburn

controller/src/main/java/org/jboss/as/controller/ThreadDumpUtil.java
aogburn: Thanks, changed

controller/src/main/java/org/jboss/as/controller/ThreadDumpUtil.java
aogburn: Thanks, changed

controller/src/main/java/org/jboss/as/controller/ThreadDumpUtil.java
aogburn: Thanks, moved the text to the ControllerLogger

controller/src/main/java/org/jboss/as/controller/OperationContextImpl.java
```diff
@@ -867,6 +869,8 @@ private void ensureWriteLockForRuntime() {
                     // different messages for both so just throwing a RuntimeException to get the automatic handling
                     // in AbstractOperationContext.executeStep is not what I wanted
                     ControllerLogger.MGMT_OP_LOGGER.timeoutAwaitingInitialStability(timeout / 1000, activeStep.operationId.name, activeStep.operationId.address);
+                    // Produce and log thread dump for diagnostics
+                    ThreadDumpUtil.threadDump();
```
yersan, line 873: @aogburn Would you mind moving the call after the `setRollbackOnly();`?

I could be too conservative, but if there is an error propagated from the `ThreadDumpUtil` it could be better to still ensure the `setRollbackOnly();` was called.

controller/src/main/java/org/jboss/as/controller/ThreadDumpUtil.java
```diff
@@ -0,0 +1,133 @@
...
+ * This software is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the GNU
+ * Lesser General Public License for more details.
+ *
+ * You should have received a copy of the GNU Lesser General Public
+ * License along with this software; if not, write to the Free
+ * Software Foundation, Inc., 51 Franklin St, Fifth Floor, Boston, MA
+ * 02110-1301 USA, or see the FSF site: http://www.fsf.org.
+ */
```
yersan, line 21: WildFly Core license header for new files must be ASL 2: https://www.apache.org/licenses/LICENSE-2.0

Fo example: https://github.com/wildfly/wildfly-core/blob/main/logging/src/main/java/org/jboss/as/logging/deployments/LoggingCleanupDeploymentProcessor.java#L1-L18

controller/src/main/java/org/jboss/as/controller/ThreadDumpUtil.java
```diff
@@ -0,0 +1,133 @@
...
+
+         if (deadlockedThreads != null && deadlockedThreads.length > 0) {
+            for (ThreadInfo threadInfo : threadMXBean.getThreadInfo(deadlockedThreads, true, true)) {
+               deadlockout.println(threadInfoToString(threadInfo));
+            }
+
+            ControllerLogger.MGMT_OP_LOGGER.deadLock(deadlockstr.toString());
+         }
+
+      } catch (IOException e) {
```
yersan, line 65: Would it be better to catch `Exception` instead so we can also capture any `RuntimeExcepion`?

yersan (changes requested):

yersan (approved):

PR #5156, opened 2022-07-22
WFCORE-5970 Server does not start when configuration file is a soft link to a file outside of the server configuration directory

parsharma:
Issue: https://issues.redhat.com/browse/WFCORE-5970

### DIFF
--- a/controller/src/main/java/org/jboss/as/controller/persistence/ConfigurationFile.java
+++ b/controller/src/main/java/org/jboss/as/controller/persistence/ConfigurationFile.java
@@ -369,7 +369,8 @@ public class ConfigurationFile {
         if (mainName != null) {
             try {
                 final File mainFile = new File(configurationDir, name != null ? name : mainName);
-                if (mainFile.getCanonicalPath().startsWith(configurationDir.getCanonicalPath())) {
+                if (mainFile.getCanonicalPath().startsWith(configurationDir.getCanonicalPath()) ||
+                            Files.isSymbolicLink(mainFile.toPath())) {
                     return new File(configurationDir, new File(mainName).getName());
                 } else if (interactionPolicy.isReadOnly()) {
                     return mainFile;
--- a/controller/src/test/java/org/jboss/as/controller/persistence/PersistanceResourceTestCase.java
+++ b/controller/src/test/java/org/jboss/as/controller/persistence/PersistanceResourceTestCase.java
@@ -1,23 +1,23 @@
 /*
-* JBoss, Home of Professional Open Source.
-* Copyright 2011, Red Hat Middleware LLC, and individual contributors
-* as indicated by the @author tags. See the copyright.txt file in the
-* distribution for a full listing of individual contributors.
-*
-* This is free software; you can redistribute it and/or modify it
-* under the terms of the GNU Lesser General Public License as
-* published by the Free Software Foundation; either version 2.1 of
-* the License, or (at your option) any later version.
-*
-* This software is distributed in the hope that it will be useful,
-* but WITHOUT ANY WARRANTY; without even the implied warranty of
-* MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the GNU
-* Lesser General Public License for more details.
-*
-* You should have received a copy of the GNU Lesser General Public
-* License along with this software; if not, write to the Free
-* Software Foundation, Inc., 51 Franklin St, Fifth Floor, Boston, MA
-* 02110-1301 USA, or see the FSF site: http://www.fsf.org.
+ * JBoss, Home of Professional Open Source.
+ * Copyright 2011, Red Hat Middleware LLC, and individual contributors
+ * as indicated by the @author tags. See the copyright.txt file in the
+ * distribution for a full listing of individual contributors.
+ *
+ * This is free software; you can redistribute it and/or modify it
+ * under the terms of the GNU Lesser General Public License as
+ * published by the Free Software Foundation; either version 2.1 of
+ * the License, or (at your option) any later version.
+ *
+ * This software is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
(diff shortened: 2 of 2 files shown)


darranl (dismissed):
Please also include the Jira ID in the commit message - in years to come if anyone is working on this code they may be interested in the context of the code change and wish to refer back to the Jira issue to understand the motivation for the change.

controller/src/main/java/org/jboss/as/controller/persistence/ConfigurationFile.java
```diff
@@ -369,7 +370,9 @@ private File determineMainFile(final String rawName, final String name) {
         if (mainName != null) {
             try {
                 final File mainFile = new File(configurationDir, name != null ? name : mainName);
-                if (mainFile.getCanonicalPath().startsWith(configurationDir.getCanonicalPath())) {
+                Path file = Paths.get(mainFile.getPath());
```
yersan, line 373: Thanks @parsharma , just a suggestion, you have already a File object, if you need a Path to check the symbolic link on the right side of || you could just convert a File into a path, that will avoid creating this object if it is not necessary
```
if (mainFile.getCanonicalPath().startsWith(configurationDir.getCanonicalPath()) || Files.isSymbolicLink(mainFile.toPath))
```

A test case for this is more than welcome

parsharma: Done

controller/src/main/java/org/jboss/as/controller/persistence/ConfigurationFile.java
```diff
@@ -369,7 +370,8 @@ private File determineMainFile(final String rawName, final String name) {
         if (mainName != null) {
             try {
                 final File mainFile = new File(configurationDir, name != null ? name : mainName);
-                if (mainFile.getCanonicalPath().startsWith(configurationDir.getCanonicalPath())) {
+                if (mainFile.getCanonicalPath().startsWith(configurationDir.getCanonicalPath()) ||
+                        Files.isSymbolicLink(Paths.get(mainFile.getPath()))) {
```
yersan, line 374: You also have a toPath() method to convert the File into a Path: https://docs.oracle.com/javase/7/docs/api/java/io/File.html#toPath()

controller/src/test/java/org/jboss/as/controller/persistence/PersistanceResourceTestCase.java
```diff
@@ -370,6 +371,15 @@ public void testRelativePathFromOutsideConfigDirectory() throws Exception {
         }
     }
 
+    @Test
+    public void testSymbolicLinkForConfigurationFiles() throws Exception {
+        Path file = Paths.get(createFile(externalDir, "standard.xml", "test").toString());
+        Path symbolicLink = Paths.get(standardDir + "standard.xml");
```
yersan, line 377: The `standardDir` is a file and it is not trailed with a slash, so the `symbolicLink` is not being created correctly.

You should use something like `Path symbolicLink = standardDir.toPath().resolve("standard.xml");`

Another issue with the test is you are never deleting the `standard.xml` file located by default on the `standardDir`. You need to replace this file with the symbolic link

yersan (changes requested):

controller/src/test/java/org/jboss/as/controller/persistence/PersistanceResourceTestCase.java
```diff
@@ -370,6 +371,17 @@ public void testRelativePathFromOutsideConfigDirectory() throws Exception {
...
     }
 
+    @Test
+    public void testSymbolicLinkForConfigurationFiles() throws Exception {
+        Path file = Paths.get(createFile(externalDir, "standard.xml", "test").toString());
+        Path symbolicLink = standardDir.toPath().resolve("standard.xml");
+        if(Files.exists(symbolicLink)){
+            Files.delete(symbolicLink);
+        }
+        Files.createSymbolicLink(symbolicLink, file);
```
yersan, line 381: You have to use absolute paths to create the symbolicLink, otherwise, the link will be created by using the current relative paths and will end up in an invalid link.

controller/src/test/java/org/jboss/as/controller/persistence/PersistanceResourceTestCase.java
```diff
@@ -370,6 +371,17 @@ public void testRelativePathFromOutsideConfigDirectory() throws Exception {
...
+        ConfigurationFile configurationFile = new ConfigurationFile(standardDir, "standard.xml", file.toString(), false);
```
yersan, line 382: You are using here wrong arguments:
* The third argument is expecting the name of the file provided by the user relative to the configuration dir, you are passing `file.toString()`, which is incorrect. Leave it as null or just "standard.xml"
* The fourth argument is `false`, which means the changes are not persisted in the "standard.xml" file by enabling `InteractionPolicy.READ_ONLY`. That will make the code used by ConfigurationFile return before testing changes giving a false negative result. You have to use `true`instead

yersan (changes requested):

yersan:
In addition to the above, @parsharma could you rebase this on top of main? thanks!

parsharma:
I have updated the code and rebased it. Thanks

controller/src/test/java/org/jboss/as/controller/persistence/PersistanceResourceTestCase.java
yersan: @parsharma You have missed this comment, you have to use absolute paths when creating the symbolic link.

yersan (changes requested):
It looks like you have missed one comment on my review.

One useful advice: One thing you can do to ensure your changes are fine is to revert your changes in the code while keeping the test case. Your test case should fail in that case, if your test case still passes, your changes are wrong. Please do that locally before force pushing the branch again, thanks

yersan (approved):

PR #5171, opened 2022-08-11
[WFCORE-5978] Move WildFly Core distribution to Jakarta Namespace Renamed dependencies

yersan:
Jira issue: https://issues.redhat.com/browse/WFCORE-5978

Relates to https://github.com/wildfly/wildfly/pull/15948


core-feature-pack/common/pom.xml
```diff
@@ -44,9 +44,14 @@
...
         </dependency>
 
+        <!-- Legacy dependency for JSonP API and Impl. We still need it for
+             internal.javax.json.api.ee8 used by org.wildfly.event.logger and
+             org.jboss.logmanager JBoss Modules
+        -->
         <dependency>
             <groupId>org.glassfish</groupId>
             <artifactId>jakarta.json</artifactId>
+            <version>${legacy.version.org.glassfish.jakarta.json}</version>
```
bstansberry, line 54: What prevents controlling this in the root pom dependencyManagement? Full WF isn't overriding this artifact so it doesn't harm WF to have it there.

Also, please use 'version.org.glassfish.jakarta.json', no 'legacy'. I'm using 'legacy' for things we intend to remove once we stop needing them for compilation of javax.* stuff. But we don't intend to remove org.glassfish:jakarta.json as we have ongoing need for it.

Note that in full WF I also used 'legacy.version.org.glassfish.jakarta.json'. I *think* I can just remove full WF's override of this dep, as I don't think there's any need for WF to not just use core's depMgmt for it.

yersan: > What prevents controlling this in the root pom dependencyManagement? 

Nothing indeed. It was that my understanding of this dependency is that it is something temporal until we are able to decouple its usage from `event-logger` and `logging` modules, and since it is an API and Implementation of a pre-Jakarta EE 10 I was starting looking at it as something that will be removed in the future.

However, at this stage, there is nothing preventing us to use the root pom for declaring it since it is part of the wildfly-core distribution.

> Note that in full WF I also used 'legacy.version.org.glassfish.jakarta.json'. I think I can just remove full WF's override of this dep, as I don't think there's any need for WF to not just use core's depMgmt for it.

If it's true that this dependency will be decoupled from `event-logger` and `logging` modules and finally removed from core, maybe it could be better to not drive it on full by using the core. That could allow us to remove it from the core when we want without worrying about full. I could be looking at this dependency from the wrong side, to me, it is something that we want to get rid of in the future.

yersan: @bstansberry I've updated the PR accordingly.

> Also, please use 'version.org.glassfish.jakarta.json', no 'legacy'. I'm using 'legacy' for things we intend to remove once we stop needing them for compilation of javax.* stuff. But we don't intend to remove org.glassfish:jakarta.json as we have ongoing need for it.

I read again this comment that that comment already clarifies I was looking at `org.glassfish:jakarta.json` from the wrong side, you asked to remove the 'legacy', that means we don't plan to remove it in the future. So +1 if you want to drive the version in WF full by using the WildFly Core

bstansberry (approved):

bstansberry:
Thanks @yersan!

PR #5145, opened 2022-07-04
WFCORE-5953/WFCORE-5954: Fixing several YAML extension issues

ehsavoie:
Two fixes for YAML Extension: empty resources and list attributes

Jira: https://issues.redhat.com/browse/WFCORE-5953
https://issues.redhat.com/browse/WFCORE-5954

Upstream: #5139

PR #5135, opened 2022-06-21
[WFCORE-5949]: Adding meta-data to json-logger via YamlConfigurationE…

ehsavoie:
…xtension results in ClassCastException.

* Adding support for MapAttributeDefinition in YamlExtension

Jira: https://issues.redhat.com/browse/WFCORE-5949

Signed-off-by: Emmanuel Hugonnet <ehugonne@redhat.com>

PR #5134, opened 2022-06-21
[WFCORE-5918]: Adding SSO to application security domain via YamlConf…

ehsavoie:
…igurationExtension doesn't work.

 * Fixing issue with placeholder resources that should be added instead
   of written.

Jira: https://issues.redhat.com/browse/WFCORE-5918

Signed-off-by: Emmanuel Hugonnet <ehugonne@redhat.com>

PR #5130, opened 2022-06-15
[WFCORE-5931] Upgrade JBoss Remoting to 5.0.25.Final

fl4via:
Thanks for submitting your Pull Request!

Jira:  https://issues.redhat.com/browse/WFCORE-5931

Main PR: #5129 

        Release Notes - JBoss Remoting (3+) - Version 5.0.25.Final
                                                                                                                                                                                
<h2>        Enhancement
</h2>
<ul>
<li>[<a href='https://issues.redhat.com/browse/REM3-391'>REM3-391</a>] -         Remove the lock around Endpoint connection creation
</li>
</ul>

PR #5173, opened 2022-08-14
[18.x] Pulls

bstansberry:
#5130 
#5134 
#5135 
#5145


yersan:
org.jboss.as.test.integration.domain.AdminOnlyPolicyTestCase is making the Jobs hang, I opened it some days ago https://issues.redhat.com/browse/WFCORE-5987 to track it. At first glance, I thought it was only reproducible on Windows Jobs, but this test is now failing also on Linux. I saw it also failing in the `wildfly-core-eap` testsuite

I'm going to take a closer look at that one

PR #5174, opened 2022-08-15
[WFCORE-3373] DomainLifecycleUtil HC reload API

lvydra:
https://issues.redhat.com/browse/WFCORE-3373

Changes in tests are here as proof of functionality and will be removed later.

### DIFF
--- a/testsuite/domain/src/test/java/org/jboss/as/test/integration/domain/AdminOnlyModeTestCase.java
+++ b/testsuite/domain/src/test/java/org/jboss/as/test/integration/domain/AdminOnlyModeTestCase.java
@@ -100,21 +100,12 @@ public class AdminOnlyModeTestCase {
 
     @Test
     public void testAdminOnlyMode() throws Exception {
-
         final DomainClient masterClient = domainMasterLifecycleUtil.getDomainClient();
 
         // restart master HC in admin only mode
+        domainMasterLifecycleUtil.reloadAdminOnly("primary");
+
         ModelNode op = new ModelNode();
-        op.get(OP_ADDR).add(HOST, "primary");
-        op.get(OP).set("reload");
-        op.get("admin-only").set(true);
-        domainMasterLifecycleUtil.executeAwaitConnectionClosed(op);
-
-        // Try to reconnect to the hc
-        domainMasterLifecycleUtil.connect();
-        domainMasterLifecycleUtil.awaitHostController(System.currentTimeMillis());
-
-        op = new ModelNode();
         op.get(OP).set(READ_ATTRIBUTE_OPERATION);
         op.get(OP_ADDR).add(HOST, "primary");
         op.get(NAME).set("running-mode");
@@ -124,18 +115,7 @@ public class AdminOnlyModeTestCase {
         Assert.assertEquals("ADMIN_ONLY", result.asString());
 
         // restart back to normal mode
-        op = new ModelNode();
-        op.get(OP_ADDR).add(HOST, "primary");
-        op.get(OP).set("reload");
-        op.get("admin-only").set(false);
-        domainMasterLifecycleUtil.executeAwaitConnectionClosed(op);
-
-        // Try to reconnect to the hc
-        domainMasterLifecycleUtil.connect();
-        domainMasterLifecycleUtil.awaitHostController(System.currentTimeMillis());
-
-        // check that the servers are up
-        domainMasterLifecycleUtil.awaitServers(System.currentTimeMillis());
+        domainMasterLifecycleUtil.reload("primary", null, true);
 
         // Wait for the slave to reconnect
         waitForHost(masterClient, "secondary");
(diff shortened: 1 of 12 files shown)


testsuite/shared/src/main/java/org/jboss/as/test/integration/domain/management/util/DomainLifecycleUtil.java
```diff
@@ -381,6 +381,54 @@ public void close() {
...
+        }
+
+        executeAwaitConnectionClosed(op);
+        // Try to reconnect to the hc
+        connect();
+        awaitHostController(System.currentTimeMillis(), startupTimeout);
+        // check that the servers are up
+        if (adminOnly == null || !adminOnly) {
+            awaitServers(System.currentTimeMillis(), startupTimeout);
+        }
```
yersan, line 429: One additional suggestion, even if you are not in admin-only mode, is not always required to wait for your servers.

Could you also make this configurable by using a method argument? If not specified, we can assume it is true.

yersan (commented):
@lvydra Would you mind resolving the conflicts? 

Notice that terms like "master"/"slave" are no longer appropriate. We are now using "primary"/"secondary" to refer to the domain hosts.

lvydra:
Hi @yersan, conflicts should be resolved now and the `waitForServers` argument added.

testsuite/domain/src/test/java/org/jboss/as/test/integration/domain/DomainControllerMigrationTestCase.java
```diff
@@ -226,15 +218,7 @@ public void testDCFailover() throws Exception {
...
-        restartOp = new ModelNode();
-        restartOp.get(ModelDescriptionConstants.ADDRESS).add(ModelDescriptionConstants.HOST, HOSTS[2]);
-        restartOp.get(ModelDescriptionConstants.OP).set("reload");
-        restartOp.get(ModelDescriptionConstants.RESTART_SERVERS).set(false);
-
-        log.info("Reloading failover-h3 to register to new domain controller.");
-        hostUtils[2].executeAwaitConnectionClosed(restartOp);
-
-        waitUntilHostControllerReady(hostUtils[2]);
+        hostUtils[2].reload(HOSTS[2], null, false, false);
```
yersan, line 221: waitUntilHostControllerReady also waits for the servers, so this should be `hostUtils[2].reload(HOSTS[2], null, false, true);`

Probably, since we are not restarting the servers here we do not need to wait for them, but, let's try to not add a change here. There are other calls in the test suite that were also waiting for the servers even if they are not restarted. If you prefer to change it, please change it on the others too. otherwise, better if you leave it as it was.

yersan (commented):
Just commenting since this can probably go as it is, but in any case, I found it difficult to track the changes with the overloading constructors, so maybe we can give it another thought to make it simpler for the reader. 

Some thoughts:

1. We are using three values for attributes that will only accept two, for example, admin-only. I understand null means using the default one implemented by the management operation, but I don't believe we are going to change the defaults in the future.
2. There is a small dependency between the arguments, only if you are restarting the servers, you could need to use the `awaitForServers`. If you are using `awaitForServers` you could need to specify a different timeout (although it is shared for the host controller startup timeout as well, although the servers are more susceptible to request a delay due to they are deploying the applications)
    
So, as a suggestion, we could have:

domainMasterLifecycleUtil.reload() -> A reload as usual
domainMasterLifecycleUtil.reloadAdminOnly() -> reload with admin-only=true

And then we can use `restarServers`, `awaitForServers`, and an optional `timeout` arguments. So we ended up with the following:

domainMasterLifecycleUtil.reload(String host)
domainMasterLifecycleUtil.reload(String host, boolean restarServers)
domainMasterLifecycleUtil.reload(String host, boolean restarServers, boolean awaitForServers, Integer... timeout) 

domainMasterLifecycleUtil.reloadAdminOnly(String host)
domainMasterLifecycleUtil.reloadAdminOnly(String host, boolean restarServers)
domainMasterLifecycleUtil.reloadAdminOnly(String host, boolean restarServers, boolean awaitForServers, Integer... timeout)

** Notice reloading in admin-only will never restart the servers, so the third method could be:
domainMasterLifecycleUtil.reloadAdminOnly(String host, boolean restarServers, Integer... hcStartupTimeout)


All of the above methods will call your implemented 
`reload(String host, Boolean adminOnly, Boolean restartServers, Boolean waitForServers, Long timeout) throws IOException, InterruptedException, TimeoutException`

Being this method private.

The above could be a bit more intuitive for the reader and the developer. @lvydra What do you think? Does it sound more intuitive to you too?

lvydra:
Hi @yersan, thank you for your suggestion, I agree that it can get confusing quickly, and will be better to refactor it.

testsuite/domain/src/test/java/org/jboss/as/test/integration/domain/AdminOnlyModeTestCase.java
```diff
@@ -221,14 +172,7 @@ public void testServersCannotStartInAdminOnlyMode() throws Exception {
...
         // restart back to normal mode
-        op = new ModelNode();
-        op.get(OP_ADDR).add(HOST, "primary");
-        op.get(OP).set("reload");
-        op.get("admin-only").set(false);
-        domainMasterLifecycleUtil.executeAwaitConnectionClosed(op);
-
-        // check that the servers are up
-        domainMasterLifecycleUtil.awaitServers(System.currentTimeMillis());
+        domainMasterLifecycleUtil.reload("primary", false, null, true, 300L);
```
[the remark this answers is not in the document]
lvydra: This was just as a proof of the timeout parameter not breaking things.

lvydra:
@yersan updated

yersan (approved):
Thanks, @lvydra, looks good to me, adding just a minor comment to the Java Doc.

As the last thing, could you squash the commits and add the Java Docs to the public methods?

lvydra:
Hi @yersan, I have added Java doc and squashed the commits.

PR #5178, opened 2022-08-16
[WFCORE-5978] Mitigate issues getting the process exist code on AdminOnlyPolicyTestCase

yersan:
Jira issue: https://issues.redhat.com/browse/WFCORE-5987

In addition, removed old code commented

### DIFF
--- a/testsuite/domain/src/test/java/org/jboss/as/test/integration/domain/AdminOnlyPolicyTestCase.java
+++ b/testsuite/domain/src/test/java/org/jboss/as/test/integration/domain/AdminOnlyPolicyTestCase.java
@@ -37,6 +37,7 @@ import java.net.URL;
 import java.util.Arrays;
 import java.util.HashSet;
 import java.util.Set;
+import java.util.concurrent.TimeoutException;
 
 import org.jboss.as.controller.PathAddress;
 import org.jboss.as.controller.PathElement;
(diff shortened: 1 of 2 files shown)


yersan:
We have also seen how this test hangs on Linux Job, I haven't found the reason why this test could hang; this PR is just to try  to make it more resilient on Windows

testsuite/domain/src/test/java/org/jboss/as/test/integration/domain/AdminOnlyPolicyTestCase.java
```diff
@@ -68,6 +68,8 @@ public class AdminOnlyPolicyTestCase {
     private static final long initTime = System.currentTimeMillis();
     private static int slaveCount;
 
+    private static int exitCodeTimeout = TimeoutUtil.adjust(5);
```
bstansberry, line 71: Is there any reason not to make this quite a bit higher, e.g. 20-30? If there's no particular requirement that the process exits quickly and if it doesn't that means there's a bug, then a higher timeout protects against odd environmental problems without leading to longer test runs or tests passing when they shouldn't.

bstansberry:
@yersan I asked a question about the timeout, but don't block on me looking again before merging. This looks fine. I'm going to cancel the SM job though as it is hanging. Unfortunately the TC thread dump feature doesn't seem to show thread info, just the process launch command.

testsuite/domain/src/test/java/org/jboss/as/test/integration/domain/AdminOnlyPolicyTestCase.java
yersan: Hi @bstansberry , there is no specific reason, I just only thought that 5 seconds should be enough waiting for an exit code. I don't see any problems if we use a higher value, so I'm going to update it

yersan:
I've just canceled the integration Job since it is unneeded for this PR. Linux - Security Manager - JDK 11 failed due to the hung, tracked on [WFCORE-6032](https://issues.redhat.com/browse/WFCORE-6032)

PR #5179, opened 2022-08-17
[WFCORE-6032] Ignore AdminOnlyPolicyTestCase

yersan:
Jira issue: https://issues.redhat.com/browse/WFCORE-6032

This test case is failing very often and it is making the testsuite to hang. Ignore it temporarily until further investigation


yersan:
@bstansberry Would you mind taking a quick look and approve if you agree?

yersan:
Canceled the integration Jobs since we do not need them

bstansberry (approved):

PR #5187, opened 2022-08-19
[WFCORE-6034] Upgrade log4j2 2.19.0

boris-unckel:
Fixes https://issues.redhat.com/browse/WFCORE-6034

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -215,7 +215,7 @@
         -->
         <version.org.apache.httpcomponents.httpclient>4.5.13</version.org.apache.httpcomponents.httpclient>
         <version.org.apache.httpcomponents.httpcore>4.4.15</version.org.apache.httpcomponents.httpcore>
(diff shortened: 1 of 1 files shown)


boris-unckel:
This is currently blocked by https://issues.apache.org/jira/browse/LOG4J2-3579

boris-unckel:
The service loader issue has been resolved. Please find all links in the Jira.

yersan (commented):
@jamezp Would you mind approving this component upgrade? I'm not 100% sure if we should proceed with it at this stage, in any case, the testsuite looks good, so in general, I'm fine with it.

jamezp (approved):

jamezp:
@yersan It looks good to me. If we want to wait until the next release that seems reasonable to me. I'm fine with putting it in now though too. Which ever you're more comfortable with.

yersan:
@boris-unckel , thank you !

PR #5193, opened 2022-08-26
[WFCORE-6011] Modify description in logging of subsystem to its corre…

moulalis:
Issue: https://issues.redhat.com/browse/WFCORE-6011


yersan:
Hi @jamezp, would you mind doing a quick review for this one? thanks

jamezp:
@yersan I don't object to this, but it will require a new round of translations at some point for downstream projects. The "Defines" to "It defines" seems to be a minor change that I can go either way on. The other change does make the last sentence use proper punctuation.

yersan (approved):

PR #5194, opened 2022-08-26
[WFCORE-6037] Upgrade to Jandex 3.0.0

Ladicek:
https://issues.redhat.com/browse/WFCORE-6037

PR #5200, opened 2022-09-07
WFCORE-6037 Upgrade jandex to 3.0.0

pferraro:
https://issues.redhat.com/browse/WFCORE-6037

@yersan @Ladicek 
This is an alternate approach to https://github.com/wildfly/wildfly-core/pull/5194 that retains the org.jboss:jandex artifact so that wildfly-core can upgrade without breaking wildfly full integration.
Once wildfly full pull in the requisite wildfly-core version, it can safely upgrade to io.smallrye:jandex.  The org.jboss:jandex artifact can then be removed from the core-feature-pack in a subsequent release.

I think this will be far simpler/faster than doing things in reverse order, as suggested in https://github.com/wildfly/wildfly-core/pull/5194

### DIFF
--- a/core-feature-pack/common/pom.xml
+++ b/core-feature-pack/common/pom.xml
@@ -80,11 +80,16 @@
         </dependency>
 
         <dependency>
+            <groupId>io.smallrye</groupId>
+            <artifactId>jandex</artifactId>
+        </dependency>
+
+        <dependency>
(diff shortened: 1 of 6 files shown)


pferraro:
Nevermind - it looks like code changes are needed in wildfly full.

PR #5202, opened 2022-09-13
WFCORE-5903: fix remote host unregistration

michpetrov:
Issue: [WFCORE-5903](https://issues.redhat.com/browse/WFCORE-5903)

Given what is happening in the [registerRemoteHost method](https://github.com/michpetrov/wildfly-core/blob/373c285655e992c5ed857de055735269063d0516/host-controller/src/main/java/org/jboss/as/host/controller/DomainModelControllerService.java#L390-L398) processing the `slaveHostRegistrations` conditionally seems wrong here.

### DIFF
--- a/host-controller/src/main/java/org/jboss/as/host/controller/DomainModelControllerService.java
+++ b/host-controller/src/main/java/org/jboss/as/host/controller/DomainModelControllerService.java
@@ -413,13 +413,13 @@ public class DomainModelControllerService extends AbstractControllerService impl
                 if (pinger != null) {
                     pinger.cancel();
                 }
+                final String address = hostRegistration.getAddress();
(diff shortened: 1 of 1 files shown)


yersan:
>  conditionally seems wrong here.

Could you elaborate on the reason why you think it is wrong?

michpetrov:
@yersan in the register method `slaveHostRegistrations.registerHost` is always executed, but in the unregister method `slaveHostRegistrations.unregisterHost` is executed only when `hostProxies` contains the hostName we're trying to remove. But conversely `hostProxies` is populated in the register method only when `registerProxyController == true` (and it will be false with admin-only). So the host gets registered (but a proxy isn't) and when you restart it doesn't get unregistered (because no proxy is registered).

yersan (approved):
Thanks for the explanation @michpetrov

PR #5207, opened 2022-09-15
[WFCORE-6052] Don't add identity attribute if it already exists

soul2zimate:
Issue: https://issues.redhat.com/browse/WFCORE-6052

@fjuma  please review


yersan (approved):
LGTM

fjuma (changes requested):
@soul2zimate Thanks for working on this. From a quick look at this, it doesn't seem like this logic should be added in the subsystem. Instead, we should check if this can be fixed directly from the wildfly-elytron code base where the attributes are being set.

soul2zimate:
@fjuma Instead of checking before call `setAttributes` method, do you mean to check around https://github.com/wildfly-security/wildfly-elytron/blob/2.x/auth/realm/base/src/main/java/org/wildfly/security/auth/realm/FileSystemSecurityRealm.java#L898-L906 for duplicate values in `attributes` parameter?

soul2zimate:
as per discussion in https://github.com/wildfly-security/wildfly-elytron/pull/1803, we won't fix this due to backward compatibility concern. closing.

PR #5214, opened 2022-09-20
[WFCORE-4902] Console availability message printing always before the boot statistics

khermano:
Issue: https://issues.redhat.com/browse/WFCORE-4902

### DIFF
--- a/server/src/main/java/org/jboss/as/server/BootstrapListener.java
+++ b/server/src/main/java/org/jboss/as/server/BootstrapListener.java
@@ -53,6 +53,8 @@ public final class BootstrapListener {
     private final String prettyVersion;
     private final FutureServiceContainer futureContainer;
     private final File tempDir;
+    private  String startedCleanMessage;
+    private  String startedWitErrorsMessage;
 
     public BootstrapListener(final ServiceContainer serviceContainer, final long startTime, final ServiceTarget serviceTarget, final FutureServiceContainer futureContainer, final String prettyVersion, final File tempDir) {
         this.serviceContainer = serviceContainer;
@@ -68,7 +70,7 @@ public final class BootstrapListener {
         return monitor;
     }
 
-    public void printBootStatistics(String message) {
+    public void generateBootStatistics(String message) {
         final StabilityStatistics statistics = new StabilityStatistics();
         try {
             monitor.awaitStability(statistics);
@@ -81,6 +83,16 @@ public final class BootstrapListener {
             done(bootstrapTime, statistics, message);
             monitor.clear();
         }
+    }
+
+    public void printBootStatisticsMessage() {
+        if (startedCleanMessage != null) {
+            ServerLogger.AS_ROOT_LOGGER.startedClean(startedCleanMessage);
+        } else if (startedWitErrorsMessage != null) {
+            ServerLogger.AS_ROOT_LOGGER.startedWitErrors(startedWitErrorsMessage);
+        }
+        startedCleanMessage = null;
+        startedWitErrorsMessage = null;
     }
 
     public void bootFailure(String message) {
@@ -104,10 +116,10 @@ public final class BootstrapListener {
         final int problem = statistics.getProblemsCount();
         final int started = statistics.getStartedCount();
         if (failed == 0 && problem == 0) {
-            ServerLogger.AS_ROOT_LOGGER.startedClean(prettyVersion, bootstrapTime, started, active + passive + onDemand + never + lazy, onDemand + passive + lazy, message);
+            startedCleanMessage = ServerLogger.AS_ROOT_LOGGER.startedCleanMessage(prettyVersion, bootstrapTime, started, active + passive + onDemand + never + lazy, onDemand + passive + lazy, message);
             createStartupMarker("success", startTime);
         } else {
-            ServerLogger.AS_ROOT_LOGGER.startedWitErrors(prettyVersion, bootstrapTime, started, active + passive + onDemand + never + lazy, failed + problem, onDemand + passive + lazy, message);
+            startedWitErrorsMessage = ServerLogger.AS_ROOT_LOGGER.startedWitErrorsMessage(prettyVersion, bootstrapTime, started, active + passive + onDemand + never + lazy, failed + problem, onDemand + passive + lazy, message);
             createStartupMarker("error", startTime);
         }
     }
--- a/server/src/main/java/org/jboss/as/server/ServerService.java
+++ b/server/src/main/java/org/jboss/as/server/ServerService.java
@@ -429,7 +429,7 @@ public final class ServerService extends AbstractControllerService {
                 String serverConfig = configuration.getServerEnvironment().getServerConfigurationFile().getMainFile().getName();
                 message = ServerLogger.AS_ROOT_LOGGER.serverConfigFileInUse(serverConfig);
             }
-            bootstrapListener.printBootStatistics(message);
+            bootstrapListener.generateBootStatistics(message);
         } else {
             // Die!
             String messageToAppend = "";
@@ -456,6 +456,7 @@ public final class ServerService extends AbstractControllerService {
 
     @Override
     protected void postBoot() {
+        bootstrapListener.printBootStatisticsMessage();
         executeAdditionalCliBootScript();
     }
 
--- a/host-controller/src/main/java/org/jboss/as/host/controller/DomainModelControllerService.java
+++ b/host-controller/src/main/java/org/jboss/as/host/controller/DomainModelControllerService.java
@@ -908,7 +908,7 @@ public class DomainModelControllerService extends AbstractControllerService impl
                     } else {
                         message = ROOT_LOGGER.configFileInUse(hostConfig);
                     }
-                    bootstrapListener.printBootStatistics(message);
+                    bootstrapListener.generateBootStatistics(message);
                 }
             } else {
                 // Die!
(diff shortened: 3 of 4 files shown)


server/src/main/java/org/jboss/as/server/ServerService.java
```diff
@@ -461,6 +455,12 @@ protected void finishBoot(boolean suspend) throws ConfigurationPersistenceExcept
 
     @Override
     protected void postBoot() {
+        String message = "";
+        if (configuration.getServerEnvironment().getServerConfigurationFile() != null) {
+            String serverConfig = configuration.getServerEnvironment().getServerConfigurationFile().getMainFile().getName();
+            message = ServerLogger.AS_ROOT_LOGGER.serverConfigFileInUse(serverConfig);
+        }
+        bootstrapListener.printBootStatistics(message);
```
yersan, line 463: This is a more disruptive change and we should avoid it.

Notice `bootstrapListener.printBootStatistics()` do more things besides just printing the boot statistics message, for example, this method marks as done the Future that returns the service container:

https://github.com/wildfly/wildfly-core/blob/main/server/src/main/java/org/jboss/as/server/BootstrapListener.java#L91

Moving the invocation of this method to the `postBoot()` will make the Future will return a value after the process transitions to RUNNING. That could have consequences for any thread waiting for the Future. The current behavior is the Future value will be returned before transitioning to RUNNING.

We should avoid changes like this one if we don't have any strong argument to do it. So far, the only motivation is to change the order of the messages, which is weak to support the change.

`postBoot()` looks adequate to fix the issue, but maybe you have to split the `.printBootStatistics()` into two methods (phases to print the message), one method to prepare the message (and to return to the future and do the same it is doing now) and the other to print it on the `postBoot()`.

You should also take a look at the Domain Mode variant since the secondary hosts could be also affected by this unexpected order of messages

yersan (changes requested):

server/src/main/java/org/jboss/as/server/ServerService.java
khermano: I am sorry that it took me so long. 
I force-pushed changes.

yersan: Hi @khermano , sure, don't worry, it is ok if it took time.

server/src/main/java/org/jboss/as/server/BootstrapListener.java
```diff
@@ -68,27 +68,34 @@ public StabilityMonitor getStabilityMonitor() {
...
+            return statistics;
+        }
+    }
+
+    public void printBootStatisticsMessage(StabilityStatistics statistics, String message) {
+        if (statistics == null) {
+            return;
         }
+        final long bootstrapTime = System.currentTimeMillis() - startTime;
+        done(bootstrapTime, statistics, message);
```
yersan, line 91: This has an unwanted side effect since you will be invoking `createStartupMarker` in the `postBoot()` phase. That will occur after the process has transitioned to `RUNNING`.

You have split the methods into two, but you have to just calculate the String message in the first method, store it, and print it in later the second method, which will be invoked on the `postBoot()`. 

All should behave as it was before, it is just you will be calculating the final message first to be printed later. We do not want to introduce any side effects, basically, changing when the message is printed is not enough to support any behavioral change.

yersan (changes requested):
We are still in the same situation as before. We do not want to introduce any side effects due to this Jira. There are not enough arguments to introduce any behavior change beyond when the log message is printed.

server/src/main/java/org/jboss/as/server/BootstrapListener.java
khermano: My apologies. I see what you mean. Thank you.
I force-pushed another change. I didn't want to store the whole message in the BootstrapListener instead of Logger, so I introduced a bit different approach. What do you think about that? Can it stay like that, or the potential of changing the statistics variable is just a no-go?

yersan (changes requested):

server/src/main/java/org/jboss/as/server/BootstrapListener.java
yersan: Hi @khermano , you have progressed with it since now we do not have unexpected side effects that change how they currently behave. However, I was thinking of a slightly different approach I think it is more simple since doesn't expose data outside of the `boostrapListener`:

You can use the `ServerLogger.AS_ROOT_LOGGER.startedClean()` and `ServerLogger.AS_ROOT_LOGGER.startedWitErrors()` to just generate the messages, which could be Strings. You have already done this in the past, just create the loggers without message id and without a log level.

So in the first phase, you will prepare these messages that `printBootStatisticsMessage` will print later invoked from the `postBoot()`.

For example, you can create two different fields for the messages in `boostrapListener`. If you use two strings, then you will know later if the `printBootStatisticsMessage` is dealing with an ERROR or an INFO message. For example, you can create these two loggers by copying the existing ones (these loggers will not use a Message ID and log level):

`ServerLogger.AS_ROOT_LOGGER.prepareStartedClean(prettyVersion, bootstrapTime, started, active + passive + onDemand + never + lazy, onDemand + passive + lazy, message)` -> Will generate the info message, you can store it in a class field variable, e.g  String startedCleanMessage;

`ServerLogger.AS_ROOT_LOGGER.startedWitErrors(prettyVersion, bootstrapTime, started, active + passive + onDemand + never + lazy, failed + problem, onDemand + passive + lazy, message)` -> Will generate the Error message, you can store it in a class field variable, e.g  String startedWitErrorsMessage;

Then in `printBootStatisticsMessage()` print the one that is available and also reset them, e.g:
```
        if (!startedCleanMessage != null) {
            ServerLogger.AS_ROOT_LOGGER.startedClean(startedCleanMessage);
        } else if (!startedWitErrorsMessage.isEmpty()) {
            ServerLogger.AS_ROOT_LOGGER.startedWitErrors(startedWitErrorsMessage);
        }
       startedCleanMessage = null;
       startedWitErrorsMessage = null;
```

You will need to adapt the existing loggers `ServerLogger.AS_ROOT_LOGGER.startedClean(...)` and `ServerLogger.AS_ROOT_LOGGER.startedWitErrors(...)` so they use a single string as a message (the one generated by the prepareXXXX methods described before). These loggers will be the ones that will have a Message ID and a Log level message.

khermano: Hi @yersan 
Thank you for the information and your suggestions. I force-pushed the changes. :)

server/src/main/java/org/jboss/as/server/BootstrapListener.java
```diff
@@ -103,11 +116,23 @@ private void done(final long bootstrapTime, final StabilityStatistics statistics
...
+        String message;
+        if (configFiles == null || configFiles.isEmpty()) {
+            message = "";
+        } else if (configFiles.size() == 2) {
+            message = ServerLogger.AS_ROOT_LOGGER.hostConfigFilesInUse(configFiles.get(0), configFiles.get(1));
+        } else if (configFiles.size() == 1 && isHostController) {
+            message = ServerLogger.AS_ROOT_LOGGER.hostConfigFileInUse(configFiles.get(0));
+        } else {
+            message = ServerLogger.AS_ROOT_LOGGER.serverConfigFileInUse(configFiles.get(0));
+        }
```
yersan, line 129: Hi @khermano , if you don't have a strong argument for doing this, I suggest reverting the changes that force you to add this logic and importing the `ServerLogger` into `DomainModelControllerService.java`. It adds more complexity and now you have related logic split into two areas, one in the `BootstrapListener.java` and one into `ServerService.java` and `DomainModelControllerService.java` (// Die! part). Notice these two classes don't need any logic to detect the different cases, you don't need to check if you are managing the Host Controller vs Server case. All of this is implicitly managed by the location of the code in these classes (`ServerService.java` and `DomainModelControllerService.java`).

However, after moving these cases into  `BootstrapListener.java` we are missing this status provided by the old location and you now need to deal with all the cases inside `BootstrapListener.java`. I don't get the motivation to do it, but share your argument if you think it can be justified.
 
I think we should stick to what we have been discussing: just split the generation of the message into two phases, one phase generates the `generateBootStatistics()`, and the other prints it on `postBoot()` without altering any behavioral change that could affect to the order of when things are done regarding when the process status change to RUNNING.

yersan (changes requested):

server/src/main/java/org/jboss/as/server/BootstrapListener.java
khermano: Hi @yersan,
Yes, I agree it added unwanted complexity and use of ServerLogger in DomainModelControllerService, which is undesirable. I was trying to move all the messages creating responsibility (both boot statistics and configFilesInUse) into BootstrapListener, which had unwanted consequences, so I am splitting it into two parts.
I reverted these changes such that the configFilesInUse is created in ServerService/ DomainModelControllerService as before. Then I send it to the method BootstrapListener.printBootStatisticsMessage, where I combine the boot statistics message into one final one. 
What do you think about that?

yersan (approved):
@khermano , thanks for the commitment and work in getting this PR done

khermano:
Gladly @yersan. Thank you for your patience and all your advice.
However, I have a problem getting these tests to pass. At first glims, it doesn't seem to have something to do with my commit, but now I am wondering. Especially the tests failing for Windows - JDK 11. 
Should I take a closer look at this, or is it some known issue?

yersan:
Hi @khermano , the `Windows - JDK 11` and `Full integration - Linux` have passed previously for this PR, so we should be fine. Lately, we have been getting random test failures since the CI is not stable yet due to the Beta component integration.

Usually, the trick here is to check the test history and verify it has passed before in other builds or if the same failure trace is in another PR. In addition, a kind of intuition with the changes applied vs what was tested on the failed test.

We hope to get better test stability in the incoming weeks when we will be moving out of the component Beta versions.

PR #5215, opened 2022-09-21
[WFCORE-4758]: Community - Simple config export for a domain server.

ehsavoie:
* Adding a new read-config-as-xml-file to replace read-config-as-xml.
* Adding more tests
* 'Fixing' some tests that aren't cleaning up their changes.

Jira: https://issues.redhat.com/browse/WFCORE-4758
Analysis: https://github.com/wildfly/wildfly-proposals/pull/428
Signed-off-by: Emmanuel Hugonnet <ehugonne@redhat.com>

### DIFF
--- a/controller/src/main/resources/org/jboss/as/controller/descriptions/common/LocalDescriptions.properties
+++ b/controller/src/main/resources/org/jboss/as/controller/descriptions/common/LocalDescriptions.properties
@@ -630,7 +630,13 @@ root.composite.reply=A composite operation response that consists of all the ste
 # Common Operations
 read-config-as-xml=Reads the current configuration and returns it in XML format.
 read-config-as-xml.response=The XML form of the persistent configuration.
+read-config-as-xml.deprecated=Use the operation 'read-config-as-xml-file' instead.
+read-config-as-xml-file=Reads the current configuration and returns it as a file in XML format.
+read-config-as-xml-file.response=The file if the XML form of the persistent configuration.
+read-config-as-xml-file.reply.uuid= The uuid of the attachement for the file.
 add-deployer-chains=Adds deployer chain
+
+
 
 #Publish
 publish.publish-configuration=Publish the configuration to some external location.
--- a/testsuite/manualmode/src/test/resources/org/jboss/as/test/manualmode/management/persistence/testWithCli.xml
+++ b/testsuite/manualmode/src/test/resources/org/jboss/as/test/manualmode/management/persistence/testWithCli.xml
@@ -1,256 +1,256 @@
-<?xml version="1.0" encoding="UTF-8"?>
+<?xml version='1.0' encoding='UTF-8'?>
 
-<server xmlns="urn:jboss:domain:20.0">
+<server xmlns='urn:jboss:domain:20.0'>
     <extensions>
-        <extension module="org.jboss.as.jmx"/>
-        <extension module="org.jboss.as.logging"/>
-        <extension module="org.jboss.as.remoting"/>
-        <extension module="org.wildfly.extension.core-management"/>
-        <extension module="org.wildfly.extension.discovery"/>
-        <extension module="org.wildfly.extension.elytron"/>
-        <extension module="org.wildfly.extension.io"/>
-        <extension module="org.wildfly.extension.request-controller"/>
-        <extension module="org.wildfly.extension.security.manager"/>
+        <extension module='org.jboss.as.jmx'/>
+        <extension module='org.jboss.as.logging'/>
+        <extension module='org.jboss.as.remoting'/>
+        <extension module='org.wildfly.extension.core-management'/>
+        <extension module='org.wildfly.extension.discovery'/>
+        <extension module='org.wildfly.extension.elytron'/>
+        <extension module='org.wildfly.extension.io'/>
+        <extension module='org.wildfly.extension.request-controller'/>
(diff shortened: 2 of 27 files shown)


ehsavoie:
add a rbac covering test

ehsavoie:
RBAC covering test added

bstansberry (approved):

testsuite/manualmode/src/test/resources/org/jboss/as/test/manualmode/management/persistence/testWithCli.xml
```diff
@@ -1,256 +1,256 @@
-<?xml version="1.0" encoding="UTF-8"?>
+<?xml version='1.0' encoding='UTF-8'?>
```
yersan, line 1: It looks like these changes are related to the YAML support more than the new operation, @ehsavoie could you describe why they are not required?

yersan (commented):
Added a minor question about some minor changes on the xml files loaded by the yaml test case. Overall looks good and those should not prevent merging this, but I want to be sure of the intention before playing a bit more with them

testsuite/manualmode/src/test/resources/org/jboss/as/test/manualmode/management/persistence/testWithCli.xml
ehsavoie: No, those changes are related to the fact that the cli was escaping " thus the XML was invalid. Replacing " by ' fixed that. It was of the initial issue we were trying to handle here.

controller/src/main/java/org/jboss/as/controller/operations/common/XmlMarshallingHandler.java
```diff
@@ -2,103 +2,45 @@
...
-public class XmlMarshallingHandler implements OperationStepHandler{
+public class XmlMarshallingHandler extends AbstractXmlMarshallingHandler {
 
     private static final String OPERATION_NAME = ModelDescriptionConstants.READ_CONFIG_AS_XML_OPERATION;
 
-    public static final SimpleOperationDefinition DEFINITION = new SimpleOperationDefinitionBuilder(OPERATION_NAME,ControllerResolver.getResolver())
+    public static final SimpleOperationDefinition DEFINITION = new SimpleOperationDefinitionBuilder(OPERATION_NAME, ControllerResolver.getResolver())
             .addAccessConstraint(SensitiveTargetAccessConstraintDefinition.READ_WHOLE_CONFIG)
             .setReplyType(ModelType.STRING)
+            .setDeprecated(ModelVersion.create(17))
```
yersan, line 32: This lags behind the current model version for WF 31, which is 24.

testsuite/manualmode/src/test/resources/org/jboss/as/test/manualmode/management/persistence/testWithCli.xml
yersan: Thanks, it is clear to me now, I see it was intentional for the `read-config-as-xml` (deprecated one), and the new operation continues using `"`

controller/src/main/java/org/jboss/as/controller/operations/common/XmlMarshallingHandler.java
ehsavoie: Since the replacement operation is in COMMUNITY stability only we shouldn't deprecate it

yersan: @pferraro This seems to be something pending to address with the new feature streams, ability to conditionally add a deprecation

controller/src/main/resources/org/jboss/as/controller/descriptions/common/LocalDescriptions.properties
```diff
@@ -630,8 +630,14 @@ root.composite.reply=A composite operation response that consists of all the ste
 # Common Operations
 read-config-as-xml=Reads the current configuration and returns it in XML format.
 read-config-as-xml.response=The XML form of the persistent configuration.
+read-config-as-xml.deprecated=Use the operation 'read-config-as-xml-file' instead.
```
yersan, line 633: Harmless, so it is ok to proceed with it, it will be needed again once we deprecate the `read-config-as`

yersan (approved):

yersan:
Thanks @ehsavoie and @bstansberry

controller/src/main/java/org/jboss/as/controller/operations/common/XmlMarshallingHandler.java
pferraro: Since the stability level of the server cannot be determined statically, you would need to define this during registration.
AIUI, you want to deprecate XmlFileMarshallingHandler.DEFINITION in favor of XmlMarshallingHandler.DEFINITION for Stability.COMMUNITY?
If so, you'll want to do something like:
e.g.
```java
resourceRegistration.registerOperationHandler(XmlFileMarshallingHandler.DEFINITION, new XmlFileMarshallingHandler(configurationPersister));
ModelVersion deprecation = resourceRegistration.enables(XmlFileMarshallingHandler.DEFINITION) ? ModelVersion.create(...) : null;
resourceRegistration.registerOperationHandler(XmlMarshallingHandler.createOperationDefinitionBuilder().setDeprecated(deprecation).build(), xmh);
```

PR #4716, opened 2021-08-23
[WFCORE-5526] Removal of legacy security realms.

darranl:
https://issues.redhat.com/browse/WFCORE-5526

PR #5218, opened 2022-09-23
WFCORE-6061 [Primary/Secondary]]Problematic Language usage deprecation and replacement of unit test cases. 

parsharma:
Issue: https://issues.redhat.com/browse/WFCORE-6061


domain-management/src/test/resources/org/jboss/as/domain/management/security/realms/group-to-principal.ldif
```diff
@@ -200,7 +200,7 @@ cn: Group Seven
 uid: GroupSeven
 uniqueMember: uid=ReferralUserSeven,ou=users,dc=group-to-principal,dc=wildfly,dc=org
 
-# This group uses the DN of the entry on the slave server.
+# This group uses the DN of the entry on the secondary server.
```
yersan, line 203: At this level there are more files with data susceptible to be changed with this PR:

./resources/org/jboss/as/domain/management/security/realms/simple-partition-slave.ldif
./resources/org/jboss/as/domain/management/security/realms/group-to-principal-slave.ldif
./resources/org/jboss/as/domain/management/security/realms/principal-to-group-slave.ldif

Is there any reason for not doing it? The goal is to get the rid of any wrong language, so we should change them

parsharma: Actually, I have not renamed any classes or files as if it is used somewhere else. I will check the files and rename them.

yersan: Well, your changes revealed they are no longer used. As far as I was able to check they should have been deleted together with https://github.com/wildfly/wildfly-core/pull/4716/commits/682b5eeef1c194b83940578183461993937cd7cf#

It's fine to remove the `/resources/org/jboss/as/domain/management/security/realms` directory with its content

parsharma: Done

yersan (approved):

PR #5229, opened 2022-10-03
[WFCORE-6074] Unlock preview features from underlying JDKs by default

ropalka:
https://issues.redhat.com/browse/WFCORE-6074

### DIFF
--- a/core-feature-pack/common/src/main/resources/content/bin/common.bat
+++ b/core-feature-pack/common/src/main/resources/content/bin/common.bat
@@ -37,6 +37,8 @@ goto :eof
     echo "%~1" | findstr /I "\-\-add\-modules" > nul
     if errorlevel == 1 (
       rem Set default modular jdk options
+      rem Needed by the clustering subsystem (uses Virtual threads)
+      set "DEFAULT_MODULAR_JVM_OPTIONS=!DEFAULT_MODULAR_JVM_OPTIONS! --enable-preview"
       rem Needed by the iiop-openjdk subsystem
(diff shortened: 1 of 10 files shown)


core-feature-pack/common/src/main/resources/content/bin/common.bat
```diff
@@ -37,6 +37,7 @@ goto :eof
     echo "%~1" | findstr /I "\-\-add\-modules" > nul
     if errorlevel == 1 (
       rem Set default modular jdk options
+      set "DEFAULT_MODULAR_JVM_OPTIONS=!DEFAULT_MODULAR_JVM_OPTIONS! --enable-preview"
```
jbliznak, line 40: it would be nice to keep the previous format

```
comment explaining need for the option
option
```

ropalka: Fixed @jbliznak

jbliznak:
Just out of curiosity - virtual threads are JDK19+ feature, how this setting affects our runs with JDK11 and JDK17? Is JGroups really not working correctly with current JDKs?

yersan:
Added 20.x label since this is still under discussion inclined to not enable this on WF27, check related WildFly PR

bstansberry:
@ropalka I'm going to close this and https://github.com/wildfly/wildfly/pull/16131 as the JGroups problem has been corrected. If you want to continue to pursue the topic as a general thing to think about, let's do it on the wildfly-dev mail list.

PR #5232, opened 2022-10-06
WFCORE-5305 Response warnings don't propagate up in read-resource response assembly

parsharma:
Issue: https://issues.redhat.com/browse/WFCORE-5305

### DIFF
--- a/controller/src/main/java/org/jboss/as/controller/operations/global/ReadResourceHandler.java
+++ b/controller/src/main/java/org/jboss/as/controller/operations/global/ReadResourceHandler.java
@@ -28,7 +28,11 @@ import static org.jboss.as.controller.descriptions.ModelDescriptionConstants.OP;
 import static org.jboss.as.controller.descriptions.ModelDescriptionConstants.OP_ADDR;
 import static org.jboss.as.controller.descriptions.ModelDescriptionConstants.READ_ATTRIBUTE_OPERATION;
 import static org.jboss.as.controller.descriptions.ModelDescriptionConstants.READ_RESOURCE_OPERATION;
+import static org.jboss.as.controller.descriptions.ModelDescriptionConstants.RESPONSE_HEADERS;
+import static org.jboss.as.controller.descriptions.ModelDescriptionConstants.WARNING;
+import static org.jboss.as.controller.descriptions.ModelDescriptionConstants.WARNINGS;
 import static org.jboss.as.controller.descriptions.ModelDescriptionConstants.RESULT;
+
 import static org.jboss.as.controller.operations.global.GlobalOperationAttributes.INCLUDE_ALIASES;
 import static org.jboss.as.controller.operations.global.GlobalOperationAttributes.INCLUDE_DEFAULTS;
 import static org.jboss.as.controller.operations.global.GlobalOperationAttributes.INCLUDE_RUNTIME;
@@ -46,6 +50,7 @@ import java.util.LinkedHashMap;
 import java.util.Map;
 import java.util.Set;
 import java.util.TreeMap;
+import java.util.logging.Level;
 
 import org.jboss.as.controller.AttributeDefinition;
 import org.jboss.as.controller.OperationContext;
@@ -500,6 +505,7 @@ public class ReadResourceHandler extends GlobalOperationHandlers.AbstractMultiTa
                 ModelNode value = ar.response;
                 if (!value.has(FAILURE_DESCRIPTION)) {
                     sortedAttributes.put(entry.getKey(), value.get(RESULT));
+                    addWarning(value, context);
                 } else if (value.hasDefined(FAILURE_DESCRIPTION)) {
                     context.getFailureDescription().set(value.get(FAILURE_DESCRIPTION));
                     failed = true;
@@ -517,6 +523,7 @@ public class ReadResourceHandler extends GlobalOperationHandlers.AbstractMultiTa
                     ModelNode value = entry.getValue();
                     iter.remove();
                     if (!value.has(FAILURE_DESCRIPTION)) {
+                        addWarning(value, context);
                         if (value.hasDefined(RESULT)) {
                             ModelNode childTypeNode = sortedChildren.get(path.getKey());
                             if (childTypeNode == null) {
@@ -609,6 +616,15 @@ public class ReadResourceHandler extends GlobalOperationHandlers.AbstractMultiTa
 
                 if (reportFilteredData && filteredData.hasFilteredData()) {
                     context.getResponseHeaders().get(ACCESS_CONTROL).set(filteredData.toModelNode());
+                }
+            }
+        }
+
+        private void addWarning(ModelNode value, OperationContext context) {
+            if (value.hasDefined(RESPONSE_HEADERS, WARNINGS)) {
+                for (ModelNode response : value.get(RESPONSE_HEADERS).get(WARNINGS).asList()) {
+                    String level = response.get("level").asString();
+                    context.addResponseWarning(Level.parse(level), response.get(WARNING));
                 }
             }
         }


yersan:
This PR is still actived

parsharma:
@yersan, could you please review this PR?
Thanks

parsharma:
@yersan I have rebased the PR.

controller/src/main/java/org/jboss/as/controller/operations/global/ReadResourceHandler.java
```diff
@@ -500,6 +504,9 @@ public void execute(OperationContext context, ModelNode operation) throws Operat
                 ModelNode value = ar.response;
                 if (!value.has(FAILURE_DESCRIPTION)) {
                     sortedAttributes.put(entry.getKey(), value.get(RESULT));
+                    if (value.hasDefined(RESPONSE_HEADERS)) {
```
yersan, line 507: Since we are only interested in the warnings, you could change the if to be:
```
if (value.hasDefined(RESPONSE_HEADERS, WARNINGS)) {
....                        
}
```

controller/src/main/java/org/jboss/as/controller/operations/global/ReadResourceHandler.java
```diff
@@ -516,7 +523,13 @@ public void execute(OperationContext context, ModelNode operation) throws Operat
                     PathElement path = entry.getKey();
                     ModelNode value = entry.getValue();
                     iter.remove();
+                    ModelNode responseHeaders = entry.getValue().get(RESPONSE_HEADERS).get(WARNINGS);
                     if (!value.has(FAILURE_DESCRIPTION)) {
+                        if (responseHeaders.isDefined()) {
+                            for (ModelNode response : responseHeaders.asList()) {
+                                context.addResponseWarning(Level.WARNING, response.get(WARNING));
```
yersan, line 530: I guess that ideally instead of adding the Level.WARNING, we should take the level added to the warning ModelNode instead: `response.get(LEVEL)`

However, that will force you to parse later the String into a valid Level object, and looking at the code at https://github.com/wildfly/wildfly-core/blob/main/controller/src/main/java/org/jboss/as/controller/AbstractOperationContext.java#L384C5-L386 we only have such a level for a Warning. I'm just adding this in case any one could be asking for the same.

yersan (commented):
Thank @parsharma , it took me some time to understand how to properly reproduce the issue. I've added some suggestions to the current chances.

controller/src/main/java/org/jboss/as/controller/operations/global/ReadResourceHandler.java
yersan: I correct myself, indeed the warning could be logged with a lower level, so we will need to parse it from the `response.get(LEVEL)` value instead of hardcoding it always with `Level.WARNING`

controller/src/main/java/org/jboss/as/controller/operations/global/ReadResourceHandler.java
```diff
@@ -500,6 +504,9 @@ public void execute(OperationContext context, ModelNode operation) throws Operat
                 ModelNode value = ar.response;
                 if (!value.has(FAILURE_DESCRIPTION)) {
                     sortedAttributes.put(entry.getKey(), value.get(RESULT));
+                    if (value.hasDefined(RESPONSE_HEADERS)) {
+                        context.addResponseWarning(Level.WARNING, value.get(RESPONSE_HEADERS).get(WARNINGS).get(0).get(WARNING));
```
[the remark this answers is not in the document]
parsharma: The reason is that we are looking for the 1st warning. Otherwise, we need to handle all the warnings here.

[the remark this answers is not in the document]
yersan: We should handle all the possible warnings here, there is no reason to handle just the first one.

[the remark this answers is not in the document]
parsharma: Done

controller/src/main/java/org/jboss/as/controller/operations/global/ReadResourceHandler.java
parsharma: Done

controller/src/main/java/org/jboss/as/controller/operations/global/ReadResourceHandler.java
```diff
@@ -516,7 +523,13 @@ public void execute(OperationContext context, ModelNode operation) throws Operat
                     PathElement path = entry.getKey();
                     ModelNode value = entry.getValue();
                     iter.remove();
+                    ModelNode responseHeaders = entry.getValue().get(RESPONSE_HEADERS).get(WARNINGS);
```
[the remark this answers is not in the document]
parsharma: Done

controller/src/main/java/org/jboss/as/controller/operations/global/ReadResourceHandler.java
```diff
@@ -516,7 +524,14 @@ public void execute(OperationContext context, ModelNode operation) throws Operat
                     PathElement path = entry.getKey();
                     ModelNode value = entry.getValue();
                     iter.remove();
+                   // ModelNode responseHeaders = entry.getValue().get(RESPONSE_HEADERS).get(WARNINGS);
                     if (!value.has(FAILURE_DESCRIPTION)) {
+                        if (value.hasDefined(RESPONSE_HEADERS, WARNINGS)){
+                            for (ModelNode response : value.get(RESPONSE_HEADERS).get(WARNINGS).asList()) {
+                                String level = response.get("level").toString().replace("\"", "");
```
yersan, line 531: You have to use `response.get("level").asString()` instead of `response.get("level").toString().replace("\"", "");`

In addition, it is always preferred to get the model attribute name from the available `ModelDescriptionConstants` class.  Instead of `response.get("level")` you should use `response.get(ModelDescriptionConstants.LEVEL)`

controller/src/main/java/org/jboss/as/controller/operations/global/ReadResourceHandler.java
```diff
@@ -500,6 +505,9 @@ public void execute(OperationContext context, ModelNode operation) throws Operat
                 ModelNode value = ar.response;
                 if (!value.has(FAILURE_DESCRIPTION)) {
                     sortedAttributes.put(entry.getKey(), value.get(RESULT));
+                    if (value.hasDefined(RESPONSE_HEADERS, WARNINGS)) {
+                        context.addResponseWarning(Level.WARNING, value.get(RESPONSE_HEADERS).get(WARNINGS));
+                    }
```
yersan, line 510: This piece is still hardcoding the Level.WARNING and is not iterating over the potential list. 

If you look at how `context.addResponseWarning()` is done, it is not prepared to add a list of warnings, it adds a single warning per call.

yersan (changes requested):

yersan:
Sadly, this is going to need additional thinking since now using Level.parse(String) we are restricting the available set of Level implementation that any caller can use when context.addResponseWarning() is invoked. 

See for example https://github.com/wildfly/wildfly/blob/main/jaxrs/src/main/java/org/jboss/as/jaxrs/DeploymentRestResourcesDefintion.java#L320 where the warning is added using org.jboss.logmanager.Level

The String representation of Level.WARN is "WARN". This string is unknown for java.util.logging.Level so we will get a llegalArgumentException when we process it.

We should find an alternative to implement these changes or make OperationContext.addResponseWarning more restrictive commenting that the Level string representation must match with java.util.logging.Level representation.

parsharma:
@yersan, thank you for the comments. I will try to find out the other way to resolve this problem.

yersan:
> Sadly, this is going to need additional thinking since now using Level.parse(String) we are restricting the available set of Level implementation that any caller can use when context.addResponseWarning() is invoked.
> 
> See for example https://github.com/wildfly/wildfly/blob/main/jaxrs/src/main/java/org/jboss/as/jaxrs/DeploymentRestResourcesDefintion.java#L320 where the warning is added using org.jboss.logmanager.Level
> 
> The String representation of Level.WARN is "WARN". This string is unknown for java.util.logging.Level so we will get a llegalArgumentException when we process it.
> 
> We should find an alternative to implement these changes or make OperationContext.addResponseWarning more restrictive commenting that the Level string representation must match with java.util.logging.Level representation.

@parsharma  I can see these two possibilities:

- At the point of processing Level.parse(level), we are sure the level ModelNode was already validated against the rules imposed by [isWarningLoggable](https://github.com/wildfly/wildfly-core/blob/main/controller/src/main/java/org/jboss/as/controller/AbstractOperationContext.java#L426) method at the OperationContext. So, we could overload the addResponseWarning method in OperationContext exposing a new `void addResponseWarning(String level, ModelNode warning);` that will add the warning without validating the level. The downside is it opens to any caller the possibility to add warnings without validating a level, which is weak, but it could be supported by a good java doc in the overloaded method.

- Make the required modifications to not only add the level string representation warning Model Node, but also add the Level value (Level.intValue()) which is what the [isWarningLoggable](https://github.com/wildfly/wildfly-core/blob/main/controller/src/main/java/org/jboss/as/controller/AbstractOperationContext.java#L426) uses to validate the Level object. The you could overload `void addResponseWarning(Level level, ModelNode warning);` with `void addResponseWarning(String level, int levelValue, ModelNode warning);` which internally will be using the `isWarningLoggable` warning method and will continue ensuring the warnings are still validated against the same rules.

WDYT?

parsharma:
@yersan 
Thanks for the options. I can see that both solutions are good, but the first one seems better if we provide a good JavaDoc with a proper explanation of that new overloaded method.

yersan:
@parsharma I moved the discussion to Zulip: https://wildfly.zulipchat.com/#narrow/stream/174184-wildfly-developers/topic/WFCORE-5305

parsharma:
@yersan I have tried to incorporate all the points discussed on Zulip. Please let me know if I am missing something.

yersan (changes requested):
Hi @parsharma , you have requested me to review this PR, however, you still have not addressed all the pending points described above. Could you take another look? thanks

parsharma:
@yersan I have updated the code accordingly. Could you please have a look?
Thanks

controller/src/main/java/org/jboss/as/controller/operations/global/ReadResourceHandler.java
```diff
@@ -500,6 +505,9 @@ public void execute(OperationContext context, ModelNode operation) throws Operat
                 ModelNode value = ar.response;
                 if (!value.has(FAILURE_DESCRIPTION)) {
                     sortedAttributes.put(entry.getKey(), value.get(RESULT));
+                    if (value.hasDefined(RESPONSE_HEADERS, WARNINGS)) {
```
yersan, line 508: I think it is better if you add this validation inside of the private method. It is common to both points and that will protect the method to be invoked when the value has not defined the RESPONSE_HEADERS/WARNINGS

controller/src/main/java/org/jboss/as/controller/operations/global/ReadResourceHandler.java
```diff
@@ -516,7 +524,11 @@ public void execute(OperationContext context, ModelNode operation) throws Operat
                     PathElement path = entry.getKey();
                     ModelNode value = entry.getValue();
                     iter.remove();
+                   // ModelNode responseHeaders = entry.getValue().get(RESPONSE_HEADERS).get(WARNINGS);
                     if (!value.has(FAILURE_DESCRIPTION)) {
+                        if (value.hasDefined(RESPONSE_HEADERS, WARNINGS)){
+                                getWarning(value, context);
```
yersan, line 530: It should keep the indentation accordingly to the current rules

controller/src/main/java/org/jboss/as/controller/operations/global/ReadResourceHandler.java
```diff
@@ -613,6 +625,13 @@ public void execute(OperationContext context, ModelNode operation) throws Operat
             }
         }
 
+        private void getWarning(ModelNode value, OperationContext context) {
```
yersan, line 628: Nitpick: `getWarning` suggests this method should return something. I would suggest something like `addResponseWarning` or `addWarning`. Those sound more appropriate to describe what this method does.

yersan (changes requested):

controller/src/main/java/org/jboss/as/controller/operations/global/ReadResourceHandler.java
```diff
@@ -516,7 +524,11 @@ public void execute(OperationContext context, ModelNode operation) throws Operat
                     PathElement path = entry.getKey();
                     ModelNode value = entry.getValue();
                     iter.remove();
+                   // ModelNode responseHeaders = entry.getValue().get(RESPONSE_HEADERS).get(WARNINGS);
```
yersan, line 527: This should be removed, seems it was left from development

yersan (changes requested):

yersan (approved):

controller/src/main/java/org/jboss/as/controller/operations/global/ReadResourceHandler.java
```diff
@@ -46,6 +50,7 @@
 import java.util.Map;
 import java.util.Set;
 import java.util.TreeMap;
+import java.util.logging.Level;
```
yersan, line 53: nitpick: imports are out of order ... can be fixed with another commit.

PR #5236, opened 2022-10-10
[WFCORE-6082] Verify the path when unzipping zip files

yersan:
Jira issue: https://issues.redhat.com/browse/WFCORE-6082

### DIFF
--- a/patching/src/main/java/org/jboss/as/patching/ZipUtils.java
+++ b/patching/src/main/java/org/jboss/as/patching/ZipUtils.java
@@ -118,9 +118,8 @@ public class ZipUtils {
             final ZipEntry entry = entries.nextElement();
             final String name = entry.getName();
             final File current = new File(patchDir, name);
-            final String canonicalDestinationPath = current.getCanonicalPath();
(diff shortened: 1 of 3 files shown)


patching/src/main/java/org/jboss/as/patching/ZipUtils.java
```diff
@@ -119,7 +119,7 @@ private static void unzip(final ZipFile zip, final File patchDir) throws IOExcep
             final String name = entry.getName();
             final File current = new File(patchDir, name);
             final String canonicalDestinationPath = current.getCanonicalPath();
```
jfdenise, line 121: I am wandering if this variable definition could be removed and replaced by `current.getCanonicalPath()` when used in `throw PatchLogger.ROOT_LOGGER.entryOutsideOfPatchDirectory(canonicalDestinationPath);`

yersan: Sounds good to me too

jfdenise (approved):

PR #5237, opened 2022-10-10
[WFCORE-6083] Upgrade Google guava to 31.1-jre

boris-unckel:
Fixes https://issues.redhat.com/browse/WFCORE-6083


yersan (approved):
Notice we do not incorporate upgrades that have moved beyond micro releases at this stage, however, this dependency is used purely for test cases in wildfly-core and we are already using this upgraded version in WildFly full, so we should be fine by doing this.

yersan:
Thanks @boris-unckel !

PR #5244, opened 2022-10-18
WFCORE-6080 Remove AttributeDefinition.isAllowNull

parsharma:
Issue: https://issues.redhat.com/browse/WFCORE-6080

### DIFF
--- a/controller/src/main/java/org/jboss/as/controller/AbstractAttributeDefinitionBuilder.java
+++ b/controller/src/main/java/org/jboss/as/controller/AbstractAttributeDefinitionBuilder.java
@@ -93,7 +93,7 @@ public abstract class AbstractAttributeDefinitionBuilder<BUILDER extends Abstrac
      * Creates a builder for an attribute with the give name and type and nullability setting.
      * @param attributeName the {@link AttributeDefinition#getName() name} of the attribute. Cannot be {@code null}
      * @param type the {@link AttributeDefinition#getType() type} of the attribute. Cannot be {@code null}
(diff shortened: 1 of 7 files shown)


parsharma:
Wildfly Issue: https://issues.redhat.com/browse/WFLY-17164
PR: https://github.com/wildfly/wildfly/pull/16181

Both PR needs to be merged together.

yersan (changes requested):
Hello @parsharma , working with PRs that need to be changed on both sides is time-consuming and more complex. If there is a simple alternative to avoid that, that's preferable.

For this specific case, you can just deprecate the old method in WildFly Core and make it invoke the new one. We can release wildfly-core with both methods. Then you can drop the deprecations in wildfly and invoke directly the new method name available on wildfly-core. Once, the wildfly PR is merged, you can finally remove the deprecated method in wildfly-core. That workflow allows us to keep an eye on the CI on each step, and since every change is atomic, we do not need to be triggering Jobs manually. We can release wildfly-core at any moment to expedite the changes.

Sometimes, coordination is required between WildFly Core and WildFly, but if there is a simple alternative, that's preferable.

parsharma:
@yersan sure. I will change the code accordingly. Thanks for the suggestion.

parsharma:
@yersan I have updated the code. Please let me know your review comments. Thanks

yersan (approved):
Changes look good to me; they go beyond Jira requirements but changing `AbstractAttributeDefinitionBuilder` accordingly sounds logical to me too, after all, it is the builder for `AttributeDefinition`

PR #5245, opened 2022-10-18
[WFCORE-6091] Drop suffix from transformed wildfly-elytron-integration artifact

jbliznak:
https://issues.redhat.com/browse/WFCORE-6091

Drop -jakarta suffix from the transformed artifact and use new suffix for original artifact instead, until it is properly migrated to Jakarta namespace.

This change would require change in wildfly to use the dependency without suffix.

### DIFF
--- a/elytron/pom.xml
+++ b/elytron/pom.xml
@@ -26,7 +26,7 @@
         <version>19.0.0.Final-SNAPSHOT</version>
     </parent>
 
-    <artifactId>wildfly-elytron-integration</artifactId>
+    <artifactId>wildfly-elytron-integration-legacy-namespace</artifactId>
 
     <name>WildFly: Elytron Subsystem</name>
 
--- a/core-feature-pack/common/pom.xml
+++ b/core-feature-pack/common/pom.xml
@@ -262,7 +262,7 @@
 
         <dependency>
             <groupId>org.wildfly.core</groupId>
-            <artifactId>wildfly-elytron-integration-jakarta</artifactId>
+            <artifactId>wildfly-elytron-integration</artifactId>
             <exclusions>
                 <exclusion>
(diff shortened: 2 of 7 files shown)


elytron/pom.xml
```diff
@@ -26,7 +26,7 @@
         <version>19.0.0.Final-SNAPSHOT</version>
     </parent>
 
-    <artifactId>wildfly-elytron-integration</artifactId>
+    <artifactId>wildfly-elytron-integration-legacy-namespace</artifactId>
 
     <name>WildFly: Elytron Subsystem</name>
```
yersan, line 31: Minor, for the shake of completeness, this should be changed to "WildFly: Elytron Subsystem (non-Jakarta Namespace)"

jbliznak: @bstansberry wasn't a fan of such change in wildfly when I pushed the first version of similar change so I didn't introduce it here either (even when I think it makes sense, it is just one more place to look for when doing the migration which is a good argument not to).

I can add it if you want it though.

yersan: Ahh, ok, I thought we wanted them. Then, we are fine as it is, thanks

yersan:
https://ci.wildfly.org/viewLog.html?buildId=330991 (Linux)
 https://ci.wildfly.org/viewLog.html?buildId=330993 (Galleon Layers)
 https://ci.wildfly.org/viewLog.html?buildId=330995 (Microprofile)
 https://ci.wildfly.org/viewLog.html?buildId=330997 (Bootable)
 https://ci.wildfly.org/viewLog.html?buildId=330999 (Preview)
 https://ci.wildfly.org/viewLog.html?buildId=331001 (Windows)

yersan (approved):

PR #5249, opened 2022-10-20
[WFCORE-6094] Increase LayersTest timeout

lvydra:
https://issues.redhat.com/browse/WFCORE-6094

PR depends on https://github.com/wildfly/wildfly/pull/16242

### DIFF
--- a/testsuite/shared/src/main/java/org/jboss/as/test/layers/LayersTest.java
+++ b/testsuite/shared/src/main/java/org/jboss/as/test/layers/LayersTest.java
@@ -41,6 +41,7 @@ import javax.xml.parsers.DocumentBuilder;
 import javax.xml.parsers.DocumentBuilderFactory;
 import javax.xml.parsers.ParserConfigurationException;
 
+import org.jboss.as.test.shared.TimeoutUtil;
 import org.w3c.dom.Document;
 import org.wildfly.core.launcher.StandaloneCommandBuilder;
 import org.xml.sax.SAXException;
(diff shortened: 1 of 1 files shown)


testsuite/shared/src/main/java/org/jboss/as/test/layers/LayersTest.java
```diff
@@ -208,7 +208,7 @@ public void run() {
             }
         };
         try {
-            executor.submit(r).get(1, TimeUnit.MINUTES);
+            executor.submit(r).get(2, TimeUnit.MINUTES);
```
yersan, line 211: Hello @lvydra , would you mind using [TimeoutUtil.java](https://github.com/wildfly/wildfly-core/blob/main/testsuite/shared/src/main/java/org/jboss/as/test/shared/TimeoutUtil.java) for handling this timeout?

This utility class allows timeouts to be modified with a correction factor. Slow CI environments can append the `ts.timeout.factor` property to increase the default timeout, making the testsuite more reliable in slow environments.

Thanks

yersan (changes requested):

yersan (changes requested):
Sadly this is going to need changes in wildfly full before merging this. The test case is complaining about a missing dependency for the test scope for the testsuite/layers module:

org.wildfly.security.manager.WildFlySecurityManager

lvydra:
Hi @yersan, I have opened PR in wildfly which adds the dependency.

yersan (approved):
Thanks @lvydra , it is green now

PR #5263, opened 2022-11-04
[WFCORE-6108] Add a new Phase constant to install global EJB client interceptors

fjuma:
https://issues.redhat.com/browse/WFCORE-6108


darranl (approved):
Tentitively approving provided we accept the WildFly change - this PR is just a phase ID so not much to discuss here.

yersan (approved):

PR #5268, opened 2022-11-10
Trivial spelling/hyphenation fixes around auto-generated self-signed …

rhusar:
…certificate messages/descriptions.


rhusar:
Sorry for a nitpic, but It's a bit silly to have bad English in some of these messages as they are printed on initial boot of the WildFly server and are the first log that users see. No need for Jira.

rhusar:
Also, are are all variants of "KeyStore", "Keystore", "Key store" that would deserve polishing up. Generally, we try not to reference Java classes in log messages, but then that affects other instances so it's going down the rabbit hole...

yersan:
I'm not a native English speaker, but the use of self-signed certificates is pretty common and familiar, isn't it? I don't really see an issue here, but I'm not opposed to the changes, is just it sounds fine to me as it is; it is familiar and understandable.

About the Keystore wording, I'm not familiar with the rules we have regarding the messages, although I could assume we can be quite flexible too. So in any case, don't take this feedback as a strong opinion. From my point of view, replacing them with KeyStore looks fine.

In any case, I've requested @darranl review, for getting his point of view, so I'll defer the decision to him.

rhusar:
> I'm not a native English speaker, but the use of self-signed certificates is pretty common and familiar, isn't it? 

The "self-signed" is the correct spelling (https://en.wikipedia.org/wiki/Self-signed_certificate)

> I don't really see an issue here, but I'm not opposed to the changes, is just it sounds fine to me as it is; it is familiar and understandable.

Au contraire, I have never seen it misspelt outside of WildFly.

> About the Keystore wording, I'm not familiar with the rules we have regarding the messages, although I could assume we can be quite flexible too. So in any case, don't take this feedback as a strong opinion. From my point of view, replacing them with KeyStore looks fine.

I agree - we are flexible, but we should be consistent. 

> In any case, I've requested @darranl review, for getting his point of view, so I'll defer the decision to him.

Perfect.

yersan:
> Yes, the "self-signed" is the correct spelling (https://en.wikipedia.org/wiki/Self-signed_certificate)

Upps, sorry Rado, somehow my brain processed the diff on the reversed side 🤦‍♂️

rhusar:
> > Yes, the "self-signed" is the correct spelling (https://en.wikipedia.org/wiki/Self-signed_certificate)
> 
> Upps, sorry Rado, somehow my brain processed the diff on the reversed side 🤦‍♂️

Haha, no worries :-)

bstansberry (approved):

bstansberry:
Thanks @rhusar

PR #5269, opened 2022-11-10
[WFCORE-5780] adapted the client-old-server tests to newer versions o…

istudens:
…f EAP

https://issues.redhat.com/browse/WFCORE-5780


yersan:
@luck3y , I played a bit with this Job, in order to verify this PR we need EAP 7.4.7 available at `/store/legacy-as/jboss-eap-7.4.7.zip` (or any other path you think it is better).

Let us know when the zip is available, so I can create a custom Job to verify this PR

testsuite/client-old-server/src/test/java/org/wildfly/client/old/server/util/Version.java
```diff
@@ -32,25 +32,7 @@ public class Version {
...
-        WF_9_0_0_FINAL(WILDFLY_CORE, "1.0.0.Final", false, JDK.JDK8, "http-remoting", "9990"),
-        WF_10_0_0_FINAL(WILDFLY_CORE, "2.0.10.Final", false, JDK.JDK8, "http-remoting", "9990"),
-
-        //For these we only had a full distribution
-        EAP_6_2_0(EAP, "6.2.0", true, JDK.JDK6, "remote", "9999"),
-        EAP_6_3_0(EAP, "6.3.0", true, JDK.JDK7, "remote", "9999"),
-        EAP_6_4_0(EAP, "6.4.0", true, JDK.JDK8, "remote", "9999"),
-        //Although EAP >= 7 is split, the full distribution is needed in the mixed domain tests
-        EAP_7_0_0(EAP, "7.0.0", true, JDK.JDK8, "http-remoting", "9990");
+        EAP_7_4_7(EAP, "7.4.7", true, JDK.JDK11, "http-remoting", "9990");
```
yersan, line 35: Hi @istudens, finally we realized that for this test we are only going to have major versions available at the CI, otherwise, we would need to package the major+the patch in a zip file for each patch with the maintenance cost.

So we are required to use 7.4.0 version here and not the 7.4.7. Could you change this version accordingly?

yersan (changes requested):

testsuite/client-old-server/src/test/java/org/wildfly/client/old/server/util/Version.java
istudens: @yersan Sure, I will do it.

istudens: @yersan done

yersan (approved):

testsuite/client-old-server/src/test/java/org/wildfly/client/old/server/util/Version.java
yersan: @istudens , thanks!

PR #5282, opened 2022-11-16
[WFCORE-6125] Stuck server after invoking update-key-pair on filesystem-realm

lvydra:
https://issues.redhat.com/browse/WFCORE-6125

### DIFF
--- a/testsuite/elytron/src/test/java/org/wildfly/test/security/common/elytron/SecurityDomainLockReleaseTest.java
+++ b/testsuite/elytron/src/test/java/org/wildfly/test/security/common/elytron/SecurityDomainLockReleaseTest.java
@@ -1,0 +1,72 @@
+/*
+ * Copyright 2022 Red Hat, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *   http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.wildfly.test.security.common.elytron;
+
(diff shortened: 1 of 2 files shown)


darranl:
Do you have any information regarding how not closing these resources was causing the deadlock?

lvydra:
Hi @darranl, deadlock is happening at [line 42](https://github.com/wildfly-security/wildfly-elytron/blob/1.x/auth/realm/base/src/main/java/org/wildfly/security/auth/realm/IdentitySharedExclusiveLock.java#L43), which is probably caused by read-identity operation not releasing the shared look.

fjuma:
@lvydra Are you able to add a test case for this that fails before your changes and passes with your changes?

lvydra:
Hi @fjuma, sure, I will add it.

lvydra:
Hi @fjuma, I have added a test case.

testsuite/elytron/src/test/java/org/wildfly/test/security/common/elytron/SecurityDomainLockReleaseTest.java
```diff
@@ -0,0 +1,92 @@
...
+    public void testReadIdentityReleasingLock() throws Exception {
+        boolean success = cli.sendLine("/subsystem=elytron/security-domain=exampleSecurityDomain:read-identity(name=user1)", true);
+        Assert.assertTrue(success);
+
+        Thread updateThread = new Thread(() -> {
+            cli.sendLine("/subsystem=elytron/filesystem-realm=exampleSecurityRealm:update-key-pair()", true);
+        });
+        updateThread.start();
+
+        boolean deadlock = false;
```
fjuma, line 69: Thanks for adding this test case @lvydra!

Your analysis is indeed correct, the `read-identity` operation was acquiring the shared lock for the identity but wasn't releasing it, resulting in a hang because the `update-key-pair` operation needs the exclusive lock in order to proceed.

Now that we've confirmed that this test case resulted in a hang before your changes and passes after your changes, I think we can update the test body here to just assert that the `update-key-pair` operation executes successfully.

Might be good to also try calling `read-identity` again after the call to `update-key-pair`.

lvydra:
Thanks @fjuma, I have updated the test case.

fjuma (approved):
Thanks @lvydra!

PR #5288, opened 2022-11-22
[WFCORE-6131] AsyncFutureTask fail ensure Exception presence

boris-unckel:
Fixes https://issues.redhat.com/browse/WFCORE-6131

### DIFF
--- a/host-controller/src/main/java/org/jboss/as/host/controller/mgmt/HostControllerRegistrationHandler.java
+++ b/host-controller/src/main/java/org/jboss/as/host/controller/mgmt/HostControllerRegistrationHandler.java
@@ -340,7 +340,7 @@ public class HostControllerRegistrationHandler implements ManagementRequestHandl
                 boolean as711 = (major == 1 && minor == 1);
                 if (as711) {
                     final OperationFailedException failure = HostControllerLogger.ROOT_LOGGER.unsupportedManagementVersionForHost(major, minor, 1, 2);
-                    registrationContext.failed(SlaveRegistrationException.ErrorCode.INCOMPATIBLE_VERSION, failure.getMessage());
+                    registrationContext.failed(failure, SlaveRegistrationException.ErrorCode.INCOMPATIBLE_VERSION, failure.getMessage());
                     throw failure;
                 }
                 // Initialize the transformers
@@ -427,7 +427,7 @@ public class HostControllerRegistrationHandler implements ManagementRequestHandl
 
         @Override
         public void failed(Exception e) {
-            failed(SlaveRegistrationException.ErrorCode.UNKNOWN, e.getClass().getName() + ":" + e.getMessage());
+            failed(e, SlaveRegistrationException.ErrorCode.UNKNOWN, e.getClass().getName() + ":" + e.getMessage());
         }
 
         @Override
@@ -443,9 +443,9 @@ public class HostControllerRegistrationHandler implements ManagementRequestHandl
                 try {
                     registerHost(transaction, result);
                 } catch (SlaveRegistrationException e) {
-                    failed(e.getErrorCode(), e.getErrorMessage());
+                    failed(e, e.getErrorCode(), e.getErrorMessage());
                 } catch (Exception e) {
-                    failed(SlaveRegistrationException.ErrorCode.UNKNOWN, e.getClass().getName() + ":" + e.getMessage());
+                    failed(e, SlaveRegistrationException.ErrorCode.UNKNOWN, e.getClass().getName() + ":" + e.getMessage());
                 }
                 if(failed) {
                     transaction.rollback();
@@ -587,13 +587,18 @@ public class HostControllerRegistrationHandler implements ManagementRequestHandl
             completeTask(!failed);
         }
 
-        void failed(SlaveRegistrationException.ErrorCode error, String message) {
+        /**
+         * @param t  - the cause of failure, must not be null
+         * @param error code representing the failure cause {@link SlaveRegistrationException.ErrorCode}
+         * @param message - text explaining the failure
+         */
+        void failed(final Throwable t, SlaveRegistrationException.ErrorCode error, String message) {
             byte errorCode = error.getCode();
             if(completed.compareAndSet(false, true)) {
                 failed = true;
                 final IOTask<?> task = this.task;
                 if(task != null) {
-                    task.setFailed();
+                    task.failed(t);
                 }
                 try {
                     sendFailedResponse(responseChannel, errorCode, message);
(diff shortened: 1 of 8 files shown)


boris-unckel:
@yersan thanks for the suggestion, yes, it's better message. Fixed.

yersan:
`Linux - Security Manager - JDK 11 (Pull Request)` and `Linux - JDK 11 (Pull Request)` failures are unrelated

bstansberry:
Except in the BootstrapImpl/FutureServiceContainer case, I don't think we should be creating exceptions in any of these methods. The methods should be documented as not accepting null and perhaps we add an assert to enforce that. AFAICT none of them, except maybe BootstrapImpl/FutureServiceContainer, can get passed a null and I don't think the intent was that they should be given a null.  This is all internal code so we are free to formalize the existing informal contract without worries that it will break something.

There are two calls to FutureServiceContainer.failed that get passed the result of ServiceController.getStartException() when the code handles a LifecycleEvent.FAILED event. It's not clear that that would never return null in that situation, so that's a case where creating an exception if null is passed is valid.

boris-unckel:
@yersan @bstansberry I have switched from placeholder exceptions to asserts. I hope the [HostControllerRegistrationHandler](https://github.com/wildfly/wildfly-core/pull/5288/files#diff-5bffcb114dc39ea546774a706c87e24f4ee7722f58ed94ebbb46768db13b42b5L702-L703) matches the requirement, if not I would undo it. Could you check again, please?

protocol/src/main/java/org/jboss/as/protocol/mgmt/ActiveOperationImpl.java
```diff
@@ -92,7 +93,8 @@ public boolean done(T result) {
             }
 
             @Override
-            public boolean failed(Throwable t) {
+            public boolean failed(final Throwable t) {
```
yersan, line 96: Could you add a Java doc describing the argument cannot be null?

server/src/main/java/org/jboss/as/server/BootstrapImpl.java
```diff
@@ -214,8 +214,9 @@ void done() {
             setResult(container);
         }
 
-        void failed(Throwable t) {
-            setFailed(t);
+        void failed(final Throwable t) {
```
yersan, line 217: Could you add a Java doc describing the argument cannot be null?

host-controller/src/main/java/org/jboss/as/host/controller/mgmt/HostControllerRegistrationHandler.java
```diff
@@ -699,8 +704,9 @@ boolean completeStep(Object result) {
             return setResult((T) result);
         }
 
-        boolean setFailed() {
-            return setFailed(null);
+        boolean failed(Throwable t) {
```
yersan, line 707: Could you add a Java doc describing the argument cannot be null?

host-controller/src/main/java/org/jboss/as/host/controller/mgmt/HostControllerRegistrationHandler.java
```diff
@@ -587,13 +592,13 @@ void completeRegistration(final ManagementRequestContext<RegistrationContext> re
             completeTask(!failed);
         }
 
-        void failed(SlaveRegistrationException.ErrorCode error, String message) {
+        void failed(final Throwable t, SlaveRegistrationException.ErrorCode error, String message) {
```
yersan, line 595: Could you add a Java doc describing the argument cannot be null?

server/src/main/java/org/jboss/as/server/mgmt/domain/ServerBootOperationsService.java
```diff
@@ -153,8 +154,9 @@ public void completed(final ModelNode result) {
         }
 
         @Override
-        public void failed(Exception e) {
-            setFailed(e);
+        public void failed(final Exception e) {
```
yersan, line 157: Could you add a Java doc describing the argument cannot be null?

yersan (changes requested):
Just added minor details, in general, we should document the methods that are not private and I guess optionally those that have package visibility. 

I'm fine with the `HostControllerRegistrationHandler` changes with a minor suggestion, thanks @boris-unckel

boris-unckel:
@yersan Thanks for taking the time for the review and your comments. I'm going to fix things as you suggested or answer directly below the issue.

host-controller/src/main/java/org/jboss/as/host/controller/mgmt/HostControllerRegistrationHandler.java
```diff
@@ -587,13 +592,13 @@ void completeRegistration(final ManagementRequestContext<RegistrationContext> re
...
             byte errorCode = error.getCode();
             if(completed.compareAndSet(false, true)) {
                 failed = true;
                 final IOTask<?> task = this.task;
                 if(task != null) {
-                    task.setFailed();
+                    task.failed(t);
```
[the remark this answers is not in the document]
boris-unckel: I have checked the caller usages: The ISEs created have context messages and not a generic message. We would loose information with a central generated ISE.

protocol/src/main/java/org/jboss/as/protocol/mgmt/ActiveOperationImpl.java
boris-unckel: Fixed.

server/src/main/java/org/jboss/as/server/BootstrapImpl.java
boris-unckel: Fixed.

host-controller/src/main/java/org/jboss/as/host/controller/mgmt/HostControllerRegistrationHandler.java
boris-unckel: Fixed.

host-controller/src/main/java/org/jboss/as/host/controller/mgmt/HostControllerRegistrationHandler.java
boris-unckel: Fixed.

server/src/main/java/org/jboss/as/server/mgmt/domain/ServerBootOperationsService.java
boris-unckel: Fixed.

yersan (changes requested):

host-controller/src/main/java/org/jboss/as/host/controller/mgmt/HostControllerRegistrationHandler.java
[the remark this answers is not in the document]
boris-unckel: Sorry, yes, got it. Fixed with a second method and reverting the callers. It's using a ```Exception``` instead of ISE now.

yersan (approved):
I'm ok with the changes, @bstansberry Can you spot anything else?

bstansberry (approved):

yersan:
@boris-unckel , thanks !

PR #5291, opened 2022-11-23
[WFCORE-6132] Upgrade sshd-common to 2.9.2 to address CVE-2022-45047

Skyllarr:
https://issues.redhat.com/browse/WFCORE-6132

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -216,7 +216,7 @@
         <version.org.apache.logging.log4j>2.19.0</version.org.apache.logging.log4j>
         <version.org.apache.maven.provider>3.5.4</version.org.apache.maven.provider>
         <version.org.apache.maven.resolver>1.1.1</version.org.apache.maven.resolver>
-        <version.org.apache.sshd>2.8.0</version.org.apache.sshd>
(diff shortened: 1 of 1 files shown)


yersan:
The "Linux - Security Manager - JDK 11" can be ignored since it passed for the same PR in a previous execution. Note in general we are getting a number of random test failures, I suspect due to MSC which is still in Beta, but that's pending evaluation.

ropalka:
This failure shouldn't be related to JBoss MSC @yersan - I remember this test was failing even before MSC moved to Beta.
I'm going to have a look to that test if there isn't some race condition there.

ropalka:
Fix [PR](https://github.com/wildfly/wildfly/pull/16311) have been sent @yersan

yersan (approved):

yersan:
@Skyllarr if possible, would you mind adding diff and release notes for this upgrade on the Jira issue? that's always very useful for doing the review

Skyllarr:
@yersan Sure, I have added the diff and RN link to the jira issue

yersan (approved):
Thanks @Skyllarr !

PR #5299, opened 2022-11-29
[WFCORE-6141] Fixing transient test failures

ropalka:
https://issues.redhat.com/browse/WFCORE-6141

### DIFF
--- a/subsystem-test/tests/src/test/java/org/jboss/as/subsystem/test/otherservices/OtherServicesSubsystemTestCase.java
+++ b/subsystem-test/tests/src/test/java/org/jboss/as/subsystem/test/otherservices/OtherServicesSubsystemTestCase.java
@@ -88,7 +88,7 @@ public class OtherServicesSubsystemTestCase extends AbstractSubsystemTest {
 
         Assert.assertNotNull(services.getContainer().getService(OtherService.NAME));
         Assert.assertNotNull(services.getContainer().getService(MyService.NAME));
-        MyService myService = (MyService)services.getContainer().getService(MyService.NAME).getValue();
+        MyService myService = (MyService)services.getContainer().getService(MyService.NAME).awaitValue();
         Assert.assertNotNull(myService.otherValue.getValue());
     }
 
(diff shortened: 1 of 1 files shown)


subsystem-test/tests/src/test/java/org/jboss/as/subsystem/test/otherservices/OtherServicesSubsystemTestCase.java
```diff
@@ -88,7 +88,7 @@ public void testOtherService() throws Exception {
 
         Assert.assertNotNull(services.getContainer().getService(OtherService.NAME));
         Assert.assertNotNull(services.getContainer().getService(MyService.NAME));
-        MyService myService = (MyService)services.getContainer().getService(MyService.NAME).getValue();
+        MyService myService = (MyService)services.getContainer().getService(MyService.NAME).awaitValue();
```
yersan, line 91: @ropalka Wouldn't we need to add a timeout to the wait condition? I'm just wondering if that can prevent any kind of hang running the test suites.

We can also proceed with this as it is and react in case we get these tests hanging the whole test suite. WDYT?

yersan:
'Windows - Bootable JAR - JDK 11' failure can be ignored, it has been reported at https://issues.redhat.com/browse/WFCORE-6069 since 19.0.0.Beta18

"Linux - Security Manager - JDK 11" and "Linux - JDK 17 (Pull Request)" failures, hopefully, will be resolved by https://github.com/wildfly/wildfly-core/pull/5300

subsystem-test/tests/src/test/java/org/jboss/as/subsystem/test/otherservices/OtherServicesSubsystemTestCase.java
ropalka: It is not needed @yersan . There is no async involved in these scenarios so it shouldn't result into hang.

yersan (approved):

PR #5300, opened 2022-11-29
[WFCORE-6144] JmxFacadeRbacEnabledTestCase tests fail intermittently due to an NPE

yersan:
Built on top of #5299

Jira issue: https://issues.redhat.com/browse/WFCORE-6144

Ensures the Service is up before initializing the MBeanServer instance value


yersan:
There are more places that need to be changed for the JMX test cases

ropalka (approved):

PR #5304, opened 2022-11-30
[WFCORE-2218] Ability to ask testrunner ServerController to wait for server process exit and get the exit code

lvydra:
https://issues.redhat.com/browse/WFCORE-2218

### DIFF
--- a/testsuite/manualmode/src/test/java/org/jboss/as/test/manualmode/management/cli/ShutdownRestartTestCase.java
+++ b/testsuite/manualmode/src/test/java/org/jboss/as/test/manualmode/management/cli/ShutdownRestartTestCase.java
@@ -1,0 +1,60 @@
+/*
+ * JBoss, Home of Professional Open Source.
+ * Copyright 2022 Red Hat, Inc., and individual contributors
+ * as indicated by the @author tags.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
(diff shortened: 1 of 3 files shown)


yersan (changes requested):

lvydra:
Hi @yersan, thank you for the review, I have updated PR.

testsuite/manualmode/src/test/java/org/jboss/as/test/manualmode/management/cli/ShutdownRestartTestCase.java
```diff
@@ -0,0 +1,60 @@
...
+
+        Thread restartTrigger = new Thread(() -> {
+            try {
+                ctx.handle("shutdown --restart");
+            } catch (CommandLineException e) {
+                throw new RuntimeException(e);
+            }
+        });
+
+        serverController.handleServerRestart(restartTrigger, TimeoutUtil.adjust(10000));
```
yersan, line 58: We should change `handleServerRestart` method if we want to use it here. `handleServerRestart` restarts the server, something that is not needed for `ShutdownRestartTestCase`, this test doesn't expect to restart the server, it is enough if it is able to verify the server has exited with the expected exit code to restart it from the launch scripts.

These manual mode test cases should end without leaving servers running, which could affect other test cases running after this one.

The simplest way could be to add a boolean argument to `handleServerRestart` so the caller can decide whether the server should be restarted. Additionally, it could be used if the caller needs to pass other arguments to `server.start();` called internally by `handleServerRestart`, since the caller will control the  `server.start(...);`

yersan (changes requested):

lvydra:
Thanks @yersan, updated.

yersan (approved):

PR #5264, opened 2022-11-08
[WFCORE-6097][WFCORE-6098] Upgrade JBoss Class File Writer to 1.3.0.Final and JBoss Invocation to 2.0.0.Final

ropalka:
https://issues.redhat.com/browse/WFCORE-6097
https://issues.redhat.com/browse/WFCORE-6098

PR #5314, opened 2022-12-10
[WFCORE-6163] Upgrade JBoss Invocation 1.7.1.Final

boris-unckel:
Fixes https://issues.redhat.com/browse/WFCORE-6163

Upstream https://github.com/wildfly/wildfly-core/pull/5264

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -229,7 +229,7 @@
         <version.org.glassfish.jakarta.json>1.1.6</version.org.glassfish.jakarta.json>
         <version.org.jboss.byteman>4.0.19</version.org.jboss.byteman>
         <version.org.jboss.classfilewriter>1.2.5.Final</version.org.jboss.classfilewriter>
-        <version.org.jboss.invocation>1.7.0.Final</version.org.jboss.invocation>
+        <version.org.jboss.invocation>1.7.1.Final</version.org.jboss.invocation>
         <version.org.jboss.jboss-dmr>1.6.1.Final</version.org.jboss.jboss-dmr>
         <version.org.jboss.jboss-vfs>3.2.17.Final</version.org.jboss.jboss-vfs>
         <version.org.jboss.logging.commons-logging-jboss-logging>1.0.0.Final</version.org.jboss.logging.commons-logging-jboss-logging>


ropalka:
This [PR](https://issues.redhat.com/browse/WFLY-17227) must be backported first before this one can be applied @boris-unckel

boris-unckel:
> This [PR](https://issues.redhat.com/browse/WFLY-17227) must be backported first before this one can be applied @boris-unckel

@ropalka Just to understand the root cause: 1.7.1.Final is not compatible with WFCORE 19, because it lacks jakarta compatibility? It's a breaking change version compared to 1.7.0.Final?

ropalka:
> @ropalka Just to understand the root cause: 1.7.1.Final is not compatible with WFCORE 19, because it lacks jakarta compatibility? 
Correct, 1.7.1.Final supports Java EE only. 
> It's a breaking change version compared to 1.7.0.Final? 
Yes, it dropped support for Jakarta EE comparing to 1.7.0.Final.
It is not a problem, because 1.7.x branch
was an intermediary step on a way to Jakarta Support.
For Jakarta EE support you need to use 2.0.0.Final.

> Second thing - the 19.x does not have `boms/legacy-ee/pom.xml` - where does 1.7.1.Final go then?
I would need to check. I am not familier with differences between 19.x and 20.x

yersan:
Hi @boris-unckel, the way to integrate the upgrade would be by following what Richard described, by backporting [PR](https://issues.redhat.com/browse/WFLY-17227). However, 19.x is a micro release for WildFly 27, so we do not want to merge major versions.


> Second thing - the 19.x does not have boms/legacy-ee/pom.xml - where does 1.7.1.Final go then?

WildFly Core doesn't use the same set of boms that WildFly uses; `1.7.1.Final` is a release for compiling against Java EE artifacts and we have already switched WildFly Core APIs to the Jakarta namespace. This version is not usable on WildFly Core 19.x since we do not have modules on WildFly core that need to be compiled against the JBoss Invocation that uses Java EE APIs.

This version is at the moment only used on WildFly, to be able to build modules that are still using Java EE APIs, those modules are later transformed to do the transition to Jakarta namespace. If you are explicitly looking for a place to update it, it should be in WildFly under 27.x branch, however, is also unnecessary since WildFly uses it only for compiling, it is not used at runtime.

I'm going to close this PR, since at the moment we don't plan to bump invocation from 1.7.0 to 2.0.0 for WildFly 27, anyway, thank you for the contribution.

PR #5318, opened 2022-12-12
[WFCORE-6167] Excluding tests on JDK20 for now until proper fix is available

ropalka:
https://issues.redhat.com/browse/WFCORE-6167


jamezp:
I'm not sure we should just ignore these tests. We should look at why it's failing because we might want to report that to OpenJDK. I don't have time to look at it right now, but we should look. I'm also not even sure this would be a logging problem. If it is, that means the `ConsoleHandler` is very broken now.

PR #5321, opened 2022-12-13
[WFCORE-6168] Write server configuration files using UTF-8 encoding a…

yersan:
…nd force to write the XML encoding header

Jira issue: https://issues.redhat.com/browse/WFCORE-6168


bstansberry (approved):

testsuite/manualmode/src/test/java/org/jboss/as/test/manualmode/management/persistence/EncodingPersistenceTestCase.java
```diff
@@ -0,0 +1,106 @@
...
+
+        final Path standalone = jbossHome.resolve("standalone").resolve("configuration");
+        try (Stream<Path> file = Files.list(standalone)) {
+            file.filter(p -> p.toString().endsWith(".xml")).forEach(this::checkXmlHeader);
+        }
+
+        final Path domain = jbossHome.resolve("domain").resolve("configuration");
+        try (Stream<Path> file = Files.list(domain)) {
+            file.filter(p -> p.toString().endsWith(".xml")).forEach(this::checkXmlHeader);
+        }
```
bstansberry, line 94: This block needs to be guarded with a Files.exists check.

bstansberry (changes requested):

yersan:
I don't get yet why the Job is complaining about missing artifacts:

"Failed to resolve artifact 'org.wildfly.core:wildfly-jmx:20.0.0.Beta5-SNAPSHOT' (position: END_TAG seen ...tifact name="org.wildfly.core:wildfly-jmx:20.0.0.Beta5-SNAPSHOT"/>"

The `org.wildfly.core:wildfly-jmx:20.0.0.Beta5-SNAPSHOT` artifact is part of the build and when the manual mode tests cases are executed, it should have been already built and installed locally.

testsuite/manualmode/src/test/java/org/jboss/as/test/manualmode/management/persistence/EncodingPersistenceTestCase.java
yersan: @bstansberry Fixed, although I'm still getting failures when `org.wildfly.core:wildfly-jmx:20.0.0.Beta5-SNAPSHOT`is being resolved; I'm now running out of ideas about what could cause the issue. 

@luck3y can you give some feedback about this? It is weird getting a build complaining about an artifact that has been installed into the maven cache before running this test.

yersan: Notice the server for this test case is launched with a different encoding, but I've tested that this is not the reason for not being able to resolve the `org.wildfly.core:wildfly-jmx:20.0.0.Beta5-SNAPSHOT` from the local maven cache

testsuite/manualmode/src/test/java/org/jboss/as/test/manualmode/management/persistence/EncodingPersistenceTestCase.java
```diff
@@ -0,0 +1,115 @@
...
+    private ServerController serverController;
+
+    /**
+     * Tests that we can write and read the persistence XML when the server is launched using an encoding different that UTF-8
+     *
+     * @throws Exception
+     */
+    @Test
+    public void testEncodingManagementClient() throws Exception {
+        String originalArgs = System.getProperty("jboss.args");
```
yersan, line 57: This is the problem, I'm getting the jboss.args here :( I forgot to change it when I was developing the test case

bstansberry (approved):

yersan:
Failed tests are unrelated, they passed on previous builds for the same PR

PR #5322, opened 2022-12-13
[19.x][WFCORE-6168] Write server configuration files using UTF-8 encoding a…

yersan:
…nd force to write the XML encoding header

Jira issue: https://issues.redhat.com/browse/WFCORE-6168

### DIFF
--- a/testsuite/manualmode/src/test/java/org/jboss/as/test/manualmode/management/persistence/EncodingPersistenceTestCase.java
+++ b/testsuite/manualmode/src/test/java/org/jboss/as/test/manualmode/management/persistence/EncodingPersistenceTestCase.java
@@ -1,0 +1,115 @@
+/*
+ * Copyright 2022 Red Hat, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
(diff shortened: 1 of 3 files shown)


testsuite/manualmode/src/test/java/org/jboss/as/test/manualmode/management/persistence/EncodingPersistenceTestCase.java
```diff
@@ -0,0 +1,107 @@
...
+        final Path standalone = jbossHome.resolve("standalone").resolve("configuration");
+        try (Stream<Path> file = Files.list(standalone)) {
+            file.filter(p -> p.toString().endsWith(".xml")).forEach(this::checkXmlHeader);
+        }
+
+        final Path domain = jbossHome.resolve("domain").resolve("configuration");
+
+        try (Stream<Path> file = Files.list(domain)) {
+            file.filter(p -> p.toString().endsWith(".xml")).forEach(this::checkXmlHeader);
+        }
```
bstansberry, line 95: This block needs to be guarded with a Files.exists check.

bstansberry (changes requested):

PR #5323, opened 2022-12-13
[WFCORE-6170] Update SNICombinedWithALPNTestCase to be compliant with…

fl4via:
… RFC6066

Signed-off-by: Flavia Rainone <frainone@redhat.com>

Jira: https://issues.redhat.com/browse/WFCORE-6170


fl4via:
the failures are not expected. I triggered a new run of Full Integration Windows to investigate.

yersan (approved):
This PR only changed a test case on WildFly Core which is not used on the Integration test, all wildfly-core Jobs are green so we can proceed with this

PR #5313, opened 2022-12-09
[WFCORE-6161] Upgrade Undertow to 2.3.2.Final

fl4via:
Signed-off-by: Flavia Rainone <frainone@redhat.com>

Jira: https://issues.redhat.com/browse/WFCORE-6161


        Release Notes - Undertow - Version 2.3.2.Final
                                                        
<h2>        Bug
</h2>
<ul>
<li>[<a href='https://issues.redhat.com/browse/UNDERTOW-2209'>UNDERTOW-2209</a>] -         deny-uncovered-methods grants access to forbidden methods when default security is blank
</li>
</ul>
                                                                                                                                                                                                                                                                        
        Release Notes - Undertow - Version 2.3.1.Final
        
<h2>        Sub-task
</h2>
<ul>
<li>[<a href='https://issues.redhat.com/browse/UNDERTOW-2054'>UNDERTOW-2054</a>] -         Buffer leak errors involving SslConduit on CI
</li>
</ul>
                                        
<h2>        Feature Request
</h2>
<ul>
<li>[<a href='https://issues.redhat.com/browse/UNDERTOW-2198'>UNDERTOW-2198</a>] -         Add Option to force the SNIHostName 
</li>
</ul>
        
<h2>        Bug
</h2>
<ul>
<li>[<a href='https://issues.redhat.com/browse/UNDERTOW-1988'>UNDERTOW-1988</a>] -         undertow-servlet-jakartaee9 has a Java EE 8 dependency
</li>
<li>[<a href='https://issues.redhat.com/browse/UNDERTOW-2123'>UNDERTOW-2123</a>] -         AsyncContextImpl.dispatch  uses empty path sometimes
</li>
<li>[<a href='https://issues.redhat.com/browse/UNDERTOW-2184'>UNDERTOW-2184</a>] -         SessionListenerBridge doesn&#39;t properly implement sessionIdChanged
</li>
<li>[<a href='https://issues.redhat.com/browse/UNDERTOW-2186'>UNDERTOW-2186</a>] -         Application sub directories named WEB-INF or META-INF are no longer served
</li>
<li>[<a href='https://issues.redhat.com/browse/UNDERTOW-2188'>UNDERTOW-2188</a>] -         Adjust corner case of covered methods if no methods are  present 
</li>
<li>[<a href='https://issues.redhat.com/browse/UNDERTOW-2189'>UNDERTOW-2189</a>] -         IOException thrown when trying to close already closed stream
</li>
[...]

PR #5324, opened 2022-12-13
Do not merge -- Combined PR

yersan:
Combines:

#5313
#5323


fl4via:
@yersan the regression in the deny uncovered methods testcase is fixed by Undertow 2.3.2.Final (that includes only https://github.com/undertow-io/undertow/pull/1423 )
the #5313 PR is updated to upgrade to 2.3.2.Final instead of 2.3.1.Final

yersan:
Got it @fl4via, I'm going to update this one and kick it off again. Notice we are also getting constant failures on Windows - JDK 11. Those failures are unrelated to this PR. Lately, the CI has become to be less stable; we are still working around this.

PR #5325, opened 2022-12-13
[WFCORE-6157] Partial revert - enabling failing tests

ropalka:
https://issues.redhat.com/browse/WFCORE-6157

### DIFF
--- a/controller/src/main/java/org/jboss/as/controller/OperationContextImpl.java
+++ b/controller/src/main/java/org/jboss/as/controller/OperationContextImpl.java
@@ -845,7 +845,7 @@ final class OperationContextImpl extends AbstractOperationContext {
                 ExecutionStatus origStatus = executionStatus;
                 try {
                     executionStatus = ExecutionStatus.AWAITING_STABILITY;
-                    modelController.awaitContainerStability(timeout, TimeUnit.MILLISECONDS, false); // interruption forbidden - see WFCORE-6158
+                    modelController.awaitContainerStability(timeout, TimeUnit.MILLISECONDS, respectInterruption);
                     notifyModificationBegun();
                 } catch (InterruptedException e) {
                     if (resultAction != ResultAction.ROLLBACK) {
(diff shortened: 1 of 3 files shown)


yersan:
Windows - JDK 11 Jobs are failing mostly constantly (with this and other PRs). I created #5327 to let the CI without the stability monitor change and see how it goes.

yersan (commented):
@bstansberry Would you mind reviewing this? I'm fine with it, although we need to prioritize the investigation and resolution of failed tests ignored here

ropalka:
@bstansberry @yersan any possibility this PR could be merged? I received a message about 90 days inactivity and warning it will be automatically closed :(

yersan:
@ropalka Could we avoid ignoring the tests by supplying a fix for them together with this PR? 

This is the main issue why this PR is blocked waiting for full approvals. The idea is to not lose test coverage; as soon we merge this one test fixing would be the priority, so much better if we can revert the change without losing the test coverage.

ropalka:
OK @yersan  , I gave it another try. Although domain management and operations execution are out of my expertise, the fix I provided with latest push is how it should work IMHO. Please correct me if I'm wrong @bstansberry .

ropalka:
Hello @bstansberry - what about this PR? Can it get it into WildFly 28 Final release? How about JBoss MSC component. Don't we need final version to be included in the release? Currently we're using MSC candidate release.

controller/src/main/java/org/jboss/as/controller/OperationContextImpl.java
```diff
@@ -1213,12 +1213,12 @@ void releaseStepLocks(AbstractOperationContext.Step step) {
                 // because of minor user impatience can release the controller lock while the
                 // container is unsettled. OTOH, if we don't allow interruption, if the
                 // container can't settle (e.g. a broken service is blocking in start()), the operation
-                // will not be cancellable. I (BES 2012/01/24) chose the former as the lesser evil.
+                // will not be cancellable. We chose the late as the lesser evil.
```
bstansberry, line 1216: I'm going to merge this, but I'm going to send up a patch changing these comments, as my impression is right now we are using false because that improves testsuite stability, which isn't a higher priority.

PR #5330, opened 2022-12-15
[WFCORE-6177] Remove test property added by EncodingPersistenceTestCase

yersan:
Jira issue: https://issues.redhat.com/browse/WFCORE-6177


yersan:
Hi @bstansberry , the integration Jobs are unrelated to this PR, so could you take a look and approve it if you agree so we can unlock the other PRs? I'll file a PR for 19.x too, this only affects the test suite

testsuite/manualmode/src/test/java/org/jboss/as/test/manualmode/management/persistence/EncodingPersistenceTestCase.java
```diff
@@ -72,6 +72,9 @@ public void testEncodingManagementClient() throws Exception {
             op = Util.getReadResourceOperation(PROPERTY_ADDR);
             ModelNode result = client.executeForResult(op);
             Assert.assertEquals("áéíóú", result.get(VALUE).asString());
+
+            op = Util.createRemoveOperation(PROPERTY_ADDR);
+            client.executeForResult(op);
```
bstansberry, line 77: This cleanup should happen before the assertEquals so it happens even if the assert fails.

bstansberry (changes requested):
I have one suggestion. Please don't wait for further review; go ahead and merge when you're happy with the change (even now is ok.)

testsuite/manualmode/src/test/java/org/jboss/as/test/manualmode/management/persistence/EncodingPersistenceTestCase.java
yersan: Good catch!, yes

yersan: It won't cover all the cases, the read there is just a sanity check. If there are encoding issues reading/writing the configuration, the server even won't start on L70 ... let me see if I find a reasonable simple way to do it.

The other option to change the https://github.com/wildfly/wildfly-core/blob/main/testsuite/manualmode/src/test/java/org/jboss/as/test/manualmode/logging/AbstractLoggingTestCase.java#L324 and make it more resilent to read invalid characters

yersan: @bstansberry Using a different configuration file specific to this test will do the trick, it'll leave on the history minor "garbage" due to starting the test with a different config file, but if the tests pass, I think we could be fine with it.

bstansberry (approved):

PR #5346, opened 2023-01-05
[WFCORE-6191] Eliminating usage of deprecated org.jboss.msc.service.ValueService

ropalka:
https://issues.redhat.com/browse/WFCORE-6191

### DIFF
--- a/server/src/main/java/org/jboss/as/server/deployment/module/ModuleSpecProcessor.java
+++ b/server/src/main/java/org/jboss/as/server/deployment/module/ModuleSpecProcessor.java
@@ -31,6 +31,7 @@ import java.util.Enumeration;
 import java.util.HashSet;
 import java.util.List;
 import java.util.PropertyPermission;
+import java.util.function.Consumer;
 
 import org.jboss.as.server.deployment.Attachments;
 import org.jboss.as.server.deployment.DeploymentPhaseContext;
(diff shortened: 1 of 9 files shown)


server/src/main/java/org/jboss/as/server/deployment/module/ModuleSpecProcessor.java
```diff
@@ -299,11 +304,12 @@ private void installAliases(final ModuleSpecification moduleSpecification, final
...
             ModuleDefinition moduleDefinition = new ModuleDefinition(alias, dependencies, spec);
 
-            final ValueService<ModuleDefinition> moduleSpecService = new ValueService<>(new ImmediateValue<>(moduleDefinition));
-            final ServiceBuilder sb = phaseContext.getServiceTarget().addService(moduleSpecServiceName, moduleSpecService);
+            final ServiceBuilder sb = phaseContext.getServiceTarget().addService(moduleSpecServiceName);
+            final Consumer<ModuleDefinition> moduleDefinitionConsumer = sb.provides(moduleSpecServiceName);
             sb.requires(deploymentUnit.getServiceName());
             sb.requires(phaseContext.getPhaseServiceName());
             sb.setInitialMode(Mode.ON_DEMAND);
+            sb.setInstance(new ModuleDefinitionService(moduleDefinitionConsumer, moduleDefinition));
```
yersan, line 312: Why you are not using here the same approach used before? Cannot this be done via `org.jboss.msc.Service.newInstance(consumer, value)` ?

ropalka: It cannot be because ModuleDefinitionService is retrieved by other services via deprecated ServiceController.getValue() method and so it needs to implement legacy org.jboss.msc.service.Service interface.

yersan (approved):

PR #5347, opened 2023-01-05
[WFCORE-6192] Add version 17 of the WildFly Elytron model and schema

fjuma:
https://issues.redhat.com/browse/WFCORE-6192

Creating this as a draft PR for now since we have at least one RFE being worked on right now that needs this. This can be merged once we have an RFE that has been pre-checked.


darranl:
This is still active

PR #5351, opened 2023-01-13
[WFCORE-6199] avoid duplicate module dependencies

aogburn:
https://issues.redhat.com/browse/WFCORE-6199

### DIFF
--- a/server/src/main/java/org/jboss/as/server/deployment/module/ModuleSpecification.java
+++ b/server/src/main/java/org/jboss/as/server/deployment/module/ModuleSpecification.java
@@ -54,6 +54,8 @@ public class ModuleSpecification extends SimpleAttachable {
      * Local dependencies are dependencies on other parts of the deployment, such as class-path entry
      */
     private final List<ModuleDependency> localDependencies = new ArrayList<ModuleDependency>();
+
+    private final Set<ModuleIdentifier> localDependenciesSet = new HashSet<ModuleIdentifier>();
     /**
      * If set to true this indicates that a dependency on this module requires a dependency on all it's local
      * dependencies.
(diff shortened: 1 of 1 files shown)


server/src/main/java/org/jboss/as/server/deployment/module/ModuleSpecification.java
```diff
@@ -150,18 +154,26 @@ public void addSystemDependencies(final Collection<ModuleDependency> dependencie
 
     public void addUserDependency(final ModuleDependency dependency) {
         allDependencies = null;
-        this.userDependencies.add(dependency);
+        if (!userDependenciesSet.contains(dependency.getIdentifier())) {
```
ropalka, line 157: This code can be enhanced. Just use 
this.userDependenciesSet.add(dependency.getIdentifier())
which returns boolean instead of 
if (!userDependenciesSet.contains(dependency.getIdentifier()))

server/src/main/java/org/jboss/as/server/deployment/module/ModuleSpecification.java
```diff
@@ -150,18 +154,26 @@ public void addSystemDependencies(final Collection<ModuleDependency> dependencie
...
+        for (final ModuleDependency dependency : dependencies) {
+            addUserDependency(dependency);
+        }
     }
 
     public void addLocalDependency(final ModuleDependency dependency) {
         allDependencies = null;
         if (!exclusions.contains(dependency.getIdentifier())) {
-            this.localDependencies.add(dependency);
+            if (!localDependenciesSet.contains(dependency.getIdentifier())) {
```
ropalka, line 173: Same as above. Replace contains() with add() method call.

ropalka (changes requested):

server/src/main/java/org/jboss/as/server/deployment/module/ModuleSpecification.java
aogburn: Thanks @ropalka good point! Fixed.

ropalka (approved):

TomasHofman:
I'm going to backport to 15.0.x.

PR #5353, opened 2023-01-17
[WFCORE-6154] Log reasons for implicit module dependencies

khermano:
Issue: https://issues.redhat.com/browse/WFCORE-6154
Related to issue: https://issues.redhat.com/browse/WFLY-2297

### DIFF
--- a/server/src/main/java/org/jboss/as/server/deployment/module/ModuleDependency.java
+++ b/server/src/main/java/org/jboss/as/server/deployment/module/ModuleDependency.java
@@ -25,6 +25,7 @@ package org.jboss.as.server.deployment.module;
 import java.io.Serializable;
 import java.util.ArrayList;
 import java.util.List;
+import java.util.Optional;
 
 import org.jboss.modules.ModuleIdentifier;
 import org.jboss.modules.ModuleLoader;
@@ -45,7 +46,11 @@ public final class ModuleDependency implements Serializable {
         if (moduleLoader != null)
             builder.append("moduleLoader=").append(moduleLoader).append(", ");
         builder.append("export=").append(export).append(", optional=").append(optional).append(", importServices=").append(
-                importServices).append("]");
+                importServices);
+        if (reason != null) {
+            builder.append(", ").append("reason=").append(reason);
+        }
+        builder.append("]");
         return builder.toString();
     }
 
@@ -59,6 +64,7 @@ public final class ModuleDependency implements Serializable {
     private final List<FilterSpecification> exportFilters = new ArrayList<FilterSpecification>();
     private final boolean importServices;
     private final boolean userSpecified;
+    private final String reason;
 
     /**
      * Construct a new instance.
@@ -71,7 +77,22 @@ public final class ModuleDependency implements Serializable {
      * @param userSpecified {@code true} if this dependency was specified by the user, {@code false} if it was automatically added
      */
     public ModuleDependency(final ModuleLoader moduleLoader, final String identifier, final boolean optional, final boolean export, final boolean importServices, final boolean userSpecified) {
-        this(moduleLoader, ModuleIdentifier.create(identifier), optional, export, importServices, userSpecified);
+        this(moduleLoader, ModuleIdentifier.create(identifier), optional, export, importServices, userSpecified, null);
+    }
+
+    /**
+     * Construct a new instance.
+     *
+     * @param moduleLoader the module loader of the dependency (if {@code null}, then use the default server module loader)
+     * @param identifier the module identifier
+     * @param optional {@code true} if this is an optional dependency
+     * @param export {@code true} if resources should be exported by default
+     * @param importServices
+     * @param userSpecified {@code true} if this dependency was specified by the user, {@code false} if it was automatically added
+     * @param reason reason for adding implicit module dependency
+     */
+    public ModuleDependency(final ModuleLoader moduleLoader, final String identifier, final boolean optional, final boolean export, final boolean importServices, final boolean userSpecified, String reason) {
+        this(moduleLoader, ModuleIdentifier.create(identifier), optional, export, importServices, userSpecified, reason);
     }
 
     /**
@@ -85,12 +106,28 @@ public final class ModuleDependency implements Serializable {
      * @param userSpecified {@code true} if this dependency was specified by the user, {@code false} if it was automatically added
      */
     public ModuleDependency(final ModuleLoader moduleLoader, final ModuleIdentifier identifier, final boolean optional, final boolean export, final boolean importServices, final boolean userSpecified) {
+        this(moduleLoader, identifier, optional, export, importServices, userSpecified, null);
+    }
+
+    /**
+     * Construct a new instance.
+     *
+     * @param moduleLoader the module loader of the dependency (if {@code null}, then use the default server module loader)
+     * @param identifier the module identifier
+     * @param optional {@code true} if this is an optional dependency
+     * @param export {@code true} if resources should be exported by default
+     * @param importServices
+     * @param userSpecified {@code true} if this dependency was specified by the user, {@code false} if it was automatically added
+     * @param reason reason for adding implicit module dependency
+     */
+    public ModuleDependency(final ModuleLoader moduleLoader, final ModuleIdentifier identifier, final boolean optional, final boolean export, final boolean importServices, final boolean userSpecified, String reason) {
         this.identifier = identifier;
         this.optional = optional;
         this.export = export;
         this.moduleLoader = moduleLoader;
         this.importServices = importServices;
         this.userSpecified = userSpecified;
+        this.reason = reason;
     }
 
     public ModuleLoader getModuleLoader() {
@@ -99,6 +136,14 @@ public final class ModuleDependency implements Serializable {
 
     public ModuleIdentifier getIdentifier() {
         return identifier;
+    }
+
+    public Optional<String> getReason() {
+        if (reason != null) {
+            return Optional.of(reason);
+        } else {
+            return Optional.empty();
+        }
     }
 
     public boolean isOptional() {


server/src/main/java/org/jboss/as/server/deployment/module/ModuleSpecification.java
```diff
@@ -151,11 +156,15 @@ public void addSystemDependencies(final Collection<ModuleDependency> dependencie
     public void addUserDependency(final ModuleDependency dependency) {
         allDependencies = null;
         this.userDependencies.add(dependency);
+        this.logImplicitDependency(dependency.getIdentifier().getName(), dependency.getIdentifier().getSlot(), dependency.getReason());
```
yersan, line 159: User dependencies are dependencies that the user has specifically added, either via jboss-deployment-structure.xml or via the manifest, so they are not implicit ones, and I believe they will end up without having a reason, it depends on how we want to code them.

Those are dependencies added by the user, so I wouldn't add a reason for them and consequently, I wouldn't try to log them.

yersan (commented):
Just added some questions related to this change so we can discuss it a bit more, I'm going to ask for them in [Zulip](https://wildfly.zulipchat.com/#narrow/stream/174184-wildfly-developers/topic/Adding.20reasons.20behind.20implicit.20dependencies), since we have initiated a conversation there, and makes more sense for a discussion.

I also think we should use [ModuleLoadService](https://github.com/wildfly/wildfly-core/blob/5b2f67e26fe3326b6d1751aded4770352c07669e/server/src/main/java/org/jboss/as/server/moduleservice/ModuleLoadService.java#L78) instead of [ModuleSpecification.java](https://github.com/wildfly/wildfly-core/pull/5353/files#diff-885b96238348931be3c4782bce06419c89e0d57814fbf27eab48f5060fd3b670) to log the traces. 

Does anyone have any objection to it? The ModuleLoadService represents the final effective module that is used by the deployment, ModuleSpecification could be used for other purposes. We will delay the trace only when the module is being loaded, but I think that could be fine.

server/src/main/java/org/jboss/as/server/deployment/module/ModuleDependency.java
```diff
@@ -59,6 +60,7 @@ public String toString() {
     private final List<FilterSpecification> exportFilters = new ArrayList<FilterSpecification>();
     private final boolean importServices;
     private final boolean userSpecified;
+    private Optional<String> reason;
```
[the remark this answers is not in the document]
khermano: Thank you, @yersan, for the feedback.
I am not sure if we mean the same thing, but in the Zulip discussion, Darran mentioned that some DUPs exist outside our codebase in the third-party future packs and subsystems. That is why I think there is a possibility that this getReason could be used there for some reason, and I believe Optional would indicate that they should work with it more carefully than just straight string. 
What do you think about that?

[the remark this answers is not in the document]
yersan: If DUPS are used outside, then an optional could have benefits for the third parties, so, I'm fine with it then.

yersan:
@khermano just for your info, we do not plan to integrate this on WFCore 20; we will revisit it on Wildfly Core 21. This area is a bit sensitive to changes and we need to do a conscious review of those modifications and be sure it is the correct approach. We will release WFCore 20 in the following weeks, so this work will be deferred for WFCore 21.

server/src/main/java/org/jboss/as/server/deployment/module/ModuleDependency.java
```diff
@@ -91,6 +108,28 @@ public ModuleDependency(final ModuleLoader moduleLoader, final ModuleIdentifier
...
+     *
+     * @param moduleLoader the module loader of the dependency (if {@code null}, then use the default server module loader)
+     * @param identifier the module identifier
+     * @param optional {@code true} if this is an optional dependency
+     * @param export {@code true} if resources should be exported by default
+     * @param importServices
+     * @param userSpecified {@code true} if this dependency was specified by the user, {@code false} if it was automatically added
+     * @param reason reason for adding implicit module dependency
+     */
+    public ModuleDependency(final ModuleLoader moduleLoader, final ModuleIdentifier identifier, final boolean optional, final boolean export, final boolean importServices, final boolean userSpecified, Optional<String> reason) {
```
yersan, line 125: Don't use an Optional<String> for the constructor and replace it with a String reason.

The callers don't need to an Optional, the callers will have two constructors, one with an additional argument for the reason, so by choosing one constructor instead of the other, they are already defining whether the reason is optional.

This makes it more concise avoiding callers the need to add think about if they should use Optional.empty() if there is no reason available.

Getting an Optional<String> for the getReason() makes sense for third parties.

server/src/main/java/org/jboss/as/server/logging/ServerLogger.java
```diff
@@ -1481,4 +1481,11 @@ public interface ServerLogger extends BasicLogger {
     @Message(id = Message.NONE, value = "%s started (with errors) in %dms - Started %d of %d services (%d services failed or missing dependencies, %d services are lazy, passive or on-demand) %s")
     String startedWitErrorsMessage(String prettyVersionString, long time, int startedServices, int allServices, int problemServices, int passiveOnDemandServices, String append);
 
+    @LogMessage(level = DEBUG)
+    @Message(id = Message.NONE, value= "Adding module dependency %s to module %s -- reason %s")
+    void logImplicitDependency(String dependency, String module, String reason);
+
+    @LogMessage(level = DEBUG)
+    @Message(id = Message.NONE, value= "Adding module dependency %s to module %s")
+    void logImplicitDependencyWithoutReason(String dependency, String module);
```
yersan, line 1490: The traces will be shown at DEBUG level and will contain a description provided by any DUPs without a proper translation. I would suggest just using the logger to print debug messages for debugging purposes.

server/src/main/java/org/jboss/as/server/moduleservice/ModuleLoadService.java
```diff
@@ -129,6 +129,16 @@ public Module getValue() throws IllegalStateException, IllegalArgumentException
...
 
     private static ServiceName install(final ServiceTarget target, final ModuleIdentifier identifier, ModuleLoadService service) {
+        for (ModuleDependency dependency : service.allDependencies) {
+            if (dependency.getReason().isPresent() && !dependency.isUserSpecified() && dependency.getIdentifier() != null
+                    && dependency.getIdentifier().getName() != null && dependency.getIdentifier().getSlot() != null) {
+                ServerLogger.AS_ROOT_LOGGER.logImplicitDependency(dependency.getIdentifier().getName(), dependency.getIdentifier().getSlot(), dependency.getReason().get());
+            } else if (!dependency.isUserSpecified() && dependency.getIdentifier() != null && dependency.getIdentifier().getName() != null
+                    && dependency.getIdentifier().getSlot() != null) {
+                ServerLogger.AS_ROOT_LOGGER.logImplicitDependencyWithoutReason(dependency.getIdentifier().getName(), dependency.getIdentifier().getSlot());
+            }
```
yersan, line 139: Let's try to avoid this, this is a bit convoluted code that could be even inside of the ModuleDependency code.

I've found we have already logged the dependencies added into the deployments. Check this log:
https://github.com/wildfly/wildfly-core/blob/main/server/src/main/java/org/jboss/as/server/deployment/module/ModuleSpecProcessor.java#L356

We can use it to log also the reasons by using the ModuleDependency.toString().

yersan (changes requested):

server/src/main/java/org/jboss/as/server/deployment/module/ModuleDependency.java
khermano: What do you mean by "it makes sense for third parties"? Can you please clarify this for me? 
Because there is this comment from Darran https://wildfly.zulipchat.com/#narrow/stream/174184-wildfly-developers/topic/Adding.20reasons.20behind.20implicit.20dependencies/near/313017577 saying that many DUPs exists outside of our codebase, and I am not sure now if this somehow touches this or not.

yersan: It follows up on what we discussed here: https://github.com/wildfly/wildfly-core/pull/5353#discussion_r1085073266

I was a bit reluctant to use an Optional, but you gave an argument when a dependency can be used by any third party (third party feature packs and subsystems).

We have two constructors, so we should not use an Optional for the constructor since choosing one or the other will do the optional choice of given a reason. However, we could have third-party code that can invoke the getReason(), so it makes sense to me to leave it returning an Optional as discussed before.

Does it clarify your question?

server/src/main/java/org/jboss/as/server/deployment/module/ModuleDependency.java
```diff
@@ -101,6 +137,10 @@ public ModuleIdentifier getIdentifier() {
         return identifier;
     }
 
+    public String getReason() {
+        return reason;
+    }
```
yersan, line 142: According to the discussion taken in https://github.com/wildfly/wildfly-core/pull/5353#discussion_r1085329304 this could return an Optional<String>. The difference is minimal, but since this can be used by any third party, getting an optional one is more friendly since advertises the optional nature of this field, especially if you are iterating over a stream of an object like this.  It is a nice to have.

server/src/main/java/org/jboss/as/server/deployment/module/ModuleDependency.java
khermano: Yes, very much. :)
Thank you for clarifying.

server/src/main/java/org/jboss/as/server/deployment/module/ModuleDependency.java
```diff
@@ -71,7 +77,22 @@ public String toString() {
...
+     * @param moduleLoader the module loader of the dependency (if {@code null}, then use the default server module loader)
+     * @param identifier the module identifier
+     * @param optional {@code true} if this is an optional dependency
+     * @param export {@code true} if resources should be exported by default
+     * @param importServices
+     * @param userSpecified {@code true} if this dependency was specified by the user, {@code false} if it was automatically added
+     * @param reason reason for adding implicit module dependency
+     */
+    public ModuleDependency(final ModuleLoader moduleLoader, final String identifier, final boolean optional, final boolean export, final boolean importServices, final boolean userSpecified, Optional<String> reason) {
+        this(moduleLoader, ModuleIdentifier.create(identifier), optional, export, importServices, userSpecified, reason);
```
yersan, line 95: Hi @khermano , don't use an Optional for a constructor.

The point is since this class already exposes two different constructors, the caller choosing one or another will be able to add the reason optionally.

We don't want the callers to be creating Optionals with a String. The caller will choose one of the overloaded constructors, the one without "String reason" to choose not reason at all, and the one with "String reason", to specify a reason.

yersan (changes requested):

server/src/main/java/org/jboss/as/server/deployment/module/ModuleDependency.java
khermano: Oh, I am sorry I have misunderstood the previous comment.
So, if I understand it correctly now, the "reason" attribute should be a String, and just and only the getReason() method will be Optional<String>. Am I right?

yersan: > Oh, I am sorry I have misunderstood the previous comment.

No problem at all!

> Am I right?

Yes, right, internally the class will use a String for the reason. This String can be null upon the kind of constructor involved, there won't be any setter, it can be final, and it will be exposed as an Optional<String> from the getter: if it is null, we should return an Empty Optional.

khermano: Oh, good. Thank you.

yersan (approved):
LGTM, thanks for all the work done with this one @khermano

khermano:
Thank you for all the guidance and reviews.

PR #5355, opened 2023-01-19
[WFCORE-6202] Don't set `-server` option for IBM i

ThePrez:
Addresses [WFCORE-6202](https://issues.redhat.com/projects/WFCORE/issues/WFCORE-6202)

### DIFF
--- a/core-feature-pack/common/src/main/resources/content/bin/common.ps1
+++ b/core-feature-pack/common/src/main/resources/content/bin/common.ps1
@@ -373,59 +373,6 @@ Function Start-WildFly-Process {
 	Env-Clean-Up
 }
 
-# Returns java -version output for JDK defined by java opts
-Function Get-Java-Version ($javaOpts) {
-    if ($javaOpts -match '-d64') {
-        $opt_version = '-d64'
-    } elseif ($javaOpts -match '-d32') {
-        $opt_version = '-d32'
-    }
-
-    $result = & $JAVA $opt_version -version 2>&1
-    if ($LASTEXITCODE -eq 1) {
-        $result = & $JAVA -version 2>&1
-    }
-
-    return $result
-}
-
-# Check if can set option to use HotSpot VM
-Function Is-Java-Server-Option ($javaOpts) {
-    if ( -Not ($javaOpts -match '-server') ) {
-        # Check user requested JDK 'data model'
-        $version = Get-Java-Version ($javaOpts)
-
-        # Check if data model is supported by JDK
-        # Check for SUN(tm) JVM w/ HotSpot support
-        # Check for OpenJDK JVM w/server support
-        if ($LASTEXITCODE -eq 1) {
-            Write-Host $version
-        } elseif ($version -match 'HotSpot' -or $version -match 'OpenJDK' -or $version -match 'IBM J9') {
-            return $true
-        }
-    }
-
-    return $false
-}
-
-# For stanalone mode, check if client option could be set, if not try to set server option
-Function Set-Java-Client-Option ($javaOpts) {
-    if ( -Not($javaOpts -match '-server') -and -Not($javaOpts -match '-client') ) {
-        # Check user requested JDK 'data model'
-        $version = Get-Java-Version ($javaOpts)
-
-        if ($LASTEXITCODE -eq 1) {
-            Write-Host $version
-        } elseif ($version -match 'Client VM') {
-            $javaOpts = ,"-client" + $javaOpts
-        } elseif ($version -match 'hotspot' -or $version -match 'openJDK' -or $version -match 'IBM J9') {
-            $javaOpts = ,"-server" + $javaOpts
-        }
-    }
-
-    return $javaOpts
-}
-
 Function Set-Global-Variables {
 PARAM(
 [Parameter(Mandatory=$true)]
--- a/core-feature-pack/common/src/main/resources/content/bin/domain.sh
+++ b/core-feature-pack/common/src/main/resources/content/bin/domain.sh
@@ -98,45 +98,6 @@ if [ "x$JAVA" = "x" ]; then
     fi
 fi
 
-# Check for -d32/-d64 in JAVA_OPTS
-JVM_OPTVERSION="-version"
-JVM_D64_OPTION=`echo $JAVA_OPTS | $GREP "\-d64"`
-JVM_D32_OPTION=`echo $JAVA_OPTS | $GREP "\-d32"`
-test "x$JVM_D64_OPTION" != "x" && JVM_OPTVERSION="-d64 $JVM_OPTVERSION"
-test "x$JVM_D32_OPTION" != "x" && JVM_OPTVERSION="-d32 $JVM_OPTVERSION"
-
-# If -server not set in JAVA_OPTS, set it, if supported
-SERVER_SET=`echo $JAVA_OPTS | $GREP "\-server"`
-if [ "x$SERVER_SET" = "x" ]; then
-
-    # Check for SUN(tm) JVM w/ HotSpot support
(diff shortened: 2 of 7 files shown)


bstansberry:
Hi @ThePrez -- I wonder if we can just drop that setting altogether. See discussion at https://wildfly.zulipchat.com/#narrow/stream/174184-wildfly-developers/topic/JVM.20-server.20option

ThePrez:
@bstansberry, given the Zulip discussion, I cleaned this up to just get rid of the `-server`/`-client` logic altogether.

yersan (approved):
Thanks, @ThePrez , LGTM, added a minor comment, but that's optional in my opinion.


@jamezp Would you mind approving this as well? 

We still have logic regarding "-client" option on the windows and PowerShell variants, but those are not relevant to fix the issue with the IBM I and I haven't spotted any problems with those scripts. Domain mode scripts don't use the "-client" option.

jamezp (dismissed):
This seems okay to me as well. I do wonder if we should update the other scripts, e.g. `standalone.bat`, as well. But as @yersan said, that doesn't really have anything to do with this specific issue.

core-feature-pack/common/src/main/resources/content/bin/standalone.sh
```diff
@@ -249,7 +249,6 @@ if [ "$PRESERVE_JAVA_OPTS" != "true" ]; then
 
     # Check If server or client is specified
```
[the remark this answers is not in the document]
ThePrez: Thanks! Fixed at https://github.com/wildfly/wildfly-core/pull/5355/commits/2e524748767d5e03e769790bf0687fe2416187b1

ThePrez:
Removed `-server` logic from standalone.bat and domain.sh and performed basic tests from Windows and IBM i, respectively

core-feature-pack/common/src/main/resources/content/bin/domain.sh
```diff
@@ -105,38 +105,6 @@ JVM_D32_OPTION=`echo $JAVA_OPTS | $GREP "\-d32"`
...
-        if [ "$darwin" != "true" ]; then
-            PROCESS_CONTROLLER_JAVA_OPTS="-server $PROCESS_CONTROLLER_JAVA_OPTS"
-            HOST_CONTROLLER_JAVA_OPTS="-server $HOST_CONTROLLER_JAVA_OPTS"
-            JVM_OPTVERSION="-server $JVM_OPTVERSION"
-        fi
-    fi
-else
-    JVM_OPTVERSION="-server $JVM_OPTVERSION"
-fi
-
```
yersan, line 139: I think it is harmless, the code is prepared to add the "-server" setting automatically if it is supported. The same is applicable to the `standalone.bat` with -client and -server. 

@ThePrez I could be missing something, what's the reason for the removal?

ThePrez: This code in `domain.sh` also breaks with system Java on IBM i (domain mode is likely very uncommon on IBM i so I didn't do proper testing with it). 

Rather than cleaning it up so it works, removal seemed to make the most sense, given the comments on [Zulip](https://wildfly.zulipchat.com/#narrow/stream/174184-wildfly-developers/topic/JVM.20-server.20option). The `-server` option has lost its relevancy over time.

yersan (changes requested):

ThePrez:
The `-server` logic should be completely removed now, except this blurb that remains in `standalone.sh`

```fortran
    # Check for -d32/-d64 in JAVA_OPTS
    JVM_D64_OPTION=`echo $JAVA_OPTS | $GREP "\-d64"`
    JVM_D32_OPTION=`echo $JAVA_OPTS | $GREP "\-d32"`

    # Check If -server is specified
    SERVER_SET=`echo $JAVA_OPTS | $GREP "\-server"`

    if [ "x$JVM_D32_OPTION" != "x" ]; then
        JVM_OPTVERSION="-d32"
    elif [ "x$JVM_D64_OPTION" != "x" ]; then
        JVM_OPTVERSION="-d64"
    elif $darwin && [ "x$SERVER_SET" = "x" ]; then
        # Use 32-bit on Mac, unless server has been specified or the user opts are incompatible
        "$JAVA" -d32 $JAVA_OPTS -version > /dev/null 2>&1 && PREPEND_JAVA_OPTS="-d32" && JVM_OPTVERSION="-d32"
    fi
```

Seems unnecessary and dated, but harmless. Opinions?

core-feature-pack/common/src/main/resources/content/bin/domain.sh
ThePrez: > Ok. I thought it was not going to fail since the above code only will add the "-server" option for HotSpot/OpenJDK and IBM J9

FWIW, System Java on this platform is IBM J9, but not all flavors of IBM J9 support `-server` 😖

yersan: Ok, I thought we were going to have a different version, then removing that completely is safest for being able to use it on IBM i.

yersan:
@ThePrez I forgot to mention that we need to squash the commits, that help us to move the change across other branches. Thanks!

ThePrez:
> > The -server logic should be completely removed now, except this blurb that remains in standalone.sh
> 
> @ThePrez apologies for the delay in giving a response. The -d32/64 options look outdated and my understanding is they are only valid for systems where you can have a 32/64 JDK installed at the same time and you want to choose the bit mode. This seems only possible on Solaris. I think we can safely remove this check, even more nowadays where WildFly's minimal JDK version is JDK 11.

I agree. Removed at latest commit

ThePrez:
> @ThePrez I forgot to mention that we need to squash the commits, that help us to move the change across other branches. Thanks!

No problem. Squashed!

core-feature-pack/common/src/main/resources/content/bin/common.ps1
```diff
@@ -375,57 +375,14 @@ Function Start-WildFly-Process {
...
-    }
-
-    $result = & $JAVA $opt_version -version 2>&1
+    $result = & $JAVA -version 2>&1
     if ($LASTEXITCODE -eq 1) {
         $result = & $JAVA -version 2>&1
     }
 
     return $result
 }
```
yersan, line 384: This function can be completely removed since it is no longer used by WildFly scripts.

Otherwise, it would need some changes, since now the `$javaOpts` argument is unused by the function, so it is no longer doing what it was expected to do. In addition, now the error if-check is redundant since it repeats the same command that generated the error.

We can safely remove it.

yersan (commented):
Thanks, @ThePrez, I've added one last comment related to your last modification but looks good to me.

@jamezp another two eyes are welcome, those scripts are always a bit tricky, could you also review them?

core-feature-pack/common/src/main/resources/content/bin/common.ps1
ThePrez: Thanks, @yersan . Removed, squashed, and re-pushed. Sorry for needing multiple iterations. 
Remove `-server`/`-client` and `-d32`/`-d64` logic from scripts

yersan (approved):
@ThePrez, Thank you!, LGTM

yersan:
@jamezp , when you have a chance, would you mind reviewing this one?

jamezp (approved):
I think this seems fine. If a user wants to use `-server` it's very easy for them to just place it in their `JAVA_OPTS` in one of the `standalon.conf.*` files.

Side note; as a former IBM i developer I'm very happy to see WildFly running on the IBM i :)

yersan:
@ThePrez , merged. Thank you for contributing to WildFly!

PR #5356, opened 2023-01-20
[WFCORE-6204] Upgrade SLF4J from 1.7.36 to 2.0.6.

jamezp:
https://issues.redhat.com/browse/WFCORE-6204

According to https://www.slf4j.org/manual.html SLF4J should be backwards compatible. This does however require changes to the implementation of `org.jboss.logging:slf4j-jboss-logmanager`. We also need to upgrade logback for testing.

We do not support logback so any user using logback would be required to provide their own version of the `slf4j-api`. We should be able to safely upgrade. Tests passed for me locally and the full test suite of WildFly.

### DIFF
--- a/core-feature-pack/common/src/main/resources/modules/system/layers/base/org/slf4j/main/module.xml
+++ b/core-feature-pack/common/src/main/resources/modules/system/layers/base/org/slf4j/main/module.xml
@@ -28,6 +28,6 @@
     </resources>
 
     <dependencies>
-        <module name="org.slf4j.impl"/>
+        <module name="org.slf4j.impl" services="import"/>
     </dependencies>
 </module>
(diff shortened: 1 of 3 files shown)


core-feature-pack/common/src/main/resources/modules/system/layers/base/org/slf4j/main/module.xml
```diff
@@ -28,6 +28,6 @@
     </resources>
 
     <dependencies>
-        <module name="org.slf4j.impl"/>
+        <module name="org.slf4j.impl" services="import"/>
```
bstansberry, line 31: Why?

jamezp: Previously SLF4J used static binding and you had to name the implementation something like `org.slf4j.StaticLoggerBinder`. In SLF4J 2.0 they switched to using a `ServiceLoader` which is why the services need to be imported from the implementation.

bstansberry (approved):

PR #5357, opened 2023-01-20
WFCORE-6205: allow encrypted expression resolution without explicit r…

michpetrov:
…esolver

Issue: [WFCORE-6205](https://issues.redhat.com/browse/WFCORE-6205)


elytron/src/main/java/org/wildfly/extension/elytron/ExpressionResolverResourceDefinition.java
```diff
@@ -62,7 +62,7 @@
  */
 class ExpressionResolverResourceDefinition extends SimpleResourceDefinition {
 
-    public static final Pattern INITIAL_PATTERN = Pattern.compile("\\$\\{.+::.*:.+}");
+    public static final Pattern INITIAL_PATTERN = Pattern.compile("\\$\\{.+::.+}");;
```
yersan, line 65: There are two semicolon characters ending the `Pattern.compile()`

fjuma: @michpetrov Are you able to provide some more details on what the issue was and about this fix?

michpetrov: @fjuma in the encrypted expression format `${ENC:resolver-name:<encoded-string>}` the "resolver-name" is supposed to be optional, but the INITIAL_PATTERN only accepts expression strings with the resolver name included. The issue only comes up when a user tries to put encrypted expressions in system properties.

bstansberry:
@michpetrov @fjuma I think this regex matches the behavior coded in https://github.com/wildfly/wildfly-core/blob/main/elytron/src/main/java/org/wildfly/extension/elytron/expression/ElytronExpressionResolver.java resolveExpressionInternal:

```
\$\{.+::[^:]*:?[^:]+}
```

(Note: the above is a standard regex string and doesn't have the required java string escaping of the \ chars.)

So, an elytron credential expression inside the ${} consists of

1) 1 or more chars.
2) The chars '::'.
3) 0 or more chars that are not ':'
4) 0 or 1  ':'
4) 1 or more chars that are not ':'

If the expression has 3 colons in a row (e.g. ENC:::abcd) the 3rd colon is ignored and it is treated the same as just 2 colons, ie. use the default resolver. I don't know if that was deliberately done in the parsing, but it is good behavior IMO as it allows people to use and empty string for the resolver name part of the expression, which some may find intuitive.

The regex in the PR allows more than one colon after the ENC:: and doesn't require content after that colon. The resolveExpressionInternal parsing would actually allow more than one colon, but would silently drop part of the expression if that was done, which is not valid. If there's no content after the final : then decrypting will fail with an exception.

michpetrov:
@bstansberry an expression such as `${ENC:::abcd}` will throw an exception because of invalid resolver, or are you describing desired behavior?

The second thing is true, a user can currently enter expression such as `${ENC::resolver:nonsense-string:encoded-string}` and it will correctly resolve but I was operating under the assumption that a user will use only those encrypted expressions that were created by the encryption tool.

I can change the regex, the issue here is that in the case of system properties (and only in that case) the initial resolution always fails (no extensions available) and so this INITIAL_PATTERN is then used to determine if the expression should be tried later when extensions are available. In a normal scenario the expression is passed straight to the resolver which doesn't do any pattern matching, so the problems described above would still occur.

bstansberry:
> @bstansberry an expression such as `${ENC:::abcd}` will throw an exception because of invalid resolver, or are you describing desired behavior?

I'm describing desired behavior. The regex is used to see if the resolver extension expression is one that ElytronExpressionResolver is meant to handle, and it as much as practical should only match strings that ElytronExpressionResolver can handle correctly. Its resolveExpressionInternal method is what defines what  ElytronExpressionResolver can properly handle. I believe it would handle `${ENC:::abcd}` correctly if it were given a chance to do so, and that IMHO is a good behavior. (My apologies if I'm wrong; I've just read the code.)

> 
> The second thing is true, a user can currently enter expression such as `${ENC::resolver:nonsense-string:encoded-string}` and it will correctly resolve but I was operating under the assumption that a user will use only those encrypted expressions that were created by the encryption tool.

We shouldn't assume that. ExpressionResolverExtension is an SPI meant to allow different implementations, not just ElytronExpressionResolver. Such expressions must be of the form ${xxx::<something>} but different extensions can have different requirements for <something>. ElytronExpressionResolver shouldn't process expressions that don't fit its own requirements.

> 
> I can change the regex, the issue here is that in the case of system properties (and only in that case) the initial resolution always fails (no extensions available) and so this INITIAL_PATTERN is then used to determine if the expression should be tried later when extensions are available. In a normal scenario the expression is passed straight to the resolver which doesn't do any pattern matching, so the problems described above would still occur.

This is a good point. The reason I commented yesterday (besides the my main goal of not wanting your good PR to get auto-closed!!) is to try and get a closer match of the regex and the actual parsing. But the parsing itself has a minor flaw in that it silently ignores garbage between the resolver name and the last ':'. Not urgent, but that should be fixed.

michpetrov:
> The regex is used to see if the resolver extension expression is one that ElytronExpressionResolver is meant to handle, and it as much as practical should only match strings that ElytronExpressionResolver can handle correctly

Right, but at the moment the regex is useful only in the specific case of unavailable extensions, [the check happens here](https://github.com/wildfly/wildfly-core/blob/main/controller/src/main/java/org/jboss/as/controller/extension/ExtensionRegistry.java#L1200).

Under "normal" circumstances, ElytronExpressionResolver will be processing every expression and it's up to it to figure out if the expression is valid. If we want the regex to match what the resolver can process then shouldn't we take advantage of it and filter out the invalid expressions before they get to the resolver?

bstansberry:
> 
> Right, but at the moment the regex is useful only in the specific case of unavailable extensions, [the check happens here](https://github.com/wildfly/wildfly-core/blob/main/controller/src/main/java/org/jboss/as/controller/extension/ExtensionRegistry.java#L1200).
> 
> Under "normal" circumstances, ElytronExpressionResolver will be processing every expression and it's up to it to figure out if the expression is valid. If we want the regex to match what the resolver can process then shouldn't we take advantage of it and filter out the invalid expressions before they get to the resolver?

Thanks @michpetrov -- I see what you mean.  I doubt always testing against the regex would be worth it, and having the resolver extension know any input had passed would be worth it. For example it would save ElytronExpressionResolver.FutureExpressionResolverExtension almost nothing.

bstansberry (approved):

bstansberry:
@yersan if CI is on on this it would be good to get into WF Core 20.

yersan (approved):

yersan:
@michpetrov thanks, and all involved in the review

PR #5359, opened 2023-01-24
WFCORE-6203 Move test dependency to testbom

parsharma:
Issue: https://issues.redhat.com/browse/WFCORE-6203

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -178,13 +178,21 @@
 
         <version.com.googlecode.javaewah>1.1.13</version.com.googlecode.javaewah>
         <version.com.google.code.findbugs>3.0.2</version.com.google.code.findbugs>
+        <version.com.google.guava>31.1-jre</version.com.google.guava>
+        <version.com.google.guava.failureaccess>1.0.1</version.com.google.guava.failureaccess>
+        <version.com.fasterxml.jackson>2.13.4</version.com.fasterxml.jackson>
+        <version.com.fasterxml.jackson.databind>${version.com.fasterxml.jackson}.1</version.com.fasterxml.jackson.databind>
         <version.com.jcraft.jsch>0.1.55</version.com.jcraft.jsch>
         <version.commons-cli>1.4</version.commons-cli>
+        <version.commons-collections>3.2.2</version.commons-collections>
         <version.commons-lang3>3.12.0</version.commons-lang3>
+        <version.commons-io>2.10.0</version.commons-io>
+        <version.io.netty>4.1.86.Final</version.io.netty>
         <version.io.smallrye.jandex>3.0.5</version.io.smallrye.jandex>
         <version.io.undertow>2.3.5.Final</version.io.undertow>
         <version.jakarta.json.jakarta-json-api>2.1.1</version.jakarta.json.jakarta-json-api>
         <version.jakarta.interceptor.jakarta-interceptors-api>2.1.0</version.jakarta.interceptor.jakarta-interceptors-api>
+        <version.jakarta.inject.jakarta.inject-api>2.0.1</version.jakarta.inject.jakarta.inject-api>
         <version.jakarta.authentication.jakarta-authentication-api>3.0.0</version.jakarta.authentication.jakarta-authentication-api>
         <version.jakarta.authorization.jakarta-authorization-api>2.1.0</version.jakarta.authorization.jakarta-authorization-api>
         <version.javax.xml.bind.jaxb-api>2.4.0-b180830.0359</version.javax.xml.bind.jaxb-api>
@@ -204,8 +212,10 @@
         <version.org.apache.maven.provider>3.5.4</version.org.apache.maven.provider>
         <version.org.apache.maven.resolver>1.1.1</version.org.apache.maven.resolver>
         <version.org.apache.sshd>2.9.2</version.org.apache.sshd>
+        <version.org.apache.velocity>2.3</version.org.apache.velocity>
         <version.org.bouncycastle>1.72</version.org.bouncycastle>
         <version.org.bouncycastle.bcpg-jdk18on>1.72.2</version.org.bouncycastle.bcpg-jdk18on>
+        <version.org.codehaus.plexus.plexus-utils>3.5.0</version.org.codehaus.plexus.plexus-utils>
         <version.org.eclipse.jgit>6.2.0.202206071550-r</version.org.eclipse.jgit>
         <version.org.eclipse.parsson>1.1.1</version.org.eclipse.parsson>
         <version.org.fusesource.jansi>1.18</version.org.fusesource.jansi>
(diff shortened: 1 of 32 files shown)


parsharma:
@bstansberry could you please review it? 
Thanks

pom.xml
```diff
@@ -2218,31 +2190,13 @@
...
-                <groupId>org.jmockit</groupId>
-                <artifactId>jmockit</artifactId>
-                <version>${version.org.jmockit}</version>
-                <scope>test</scope>
-            </dependency>
             <dependency>
                 <groupId>org.mockito</groupId>
                 <artifactId>mockito-core</artifactId>
                 <version>${version.org.mockito}</version>
                 <scope>test</scope>
```
yersan, line 2197: Hi @parsharma , there are still some dependencies in the main pom with the test scope. Why are those not managed via the test bom? You have moved some of them, but not all, so I wonder what's the reason.

parsharma: Hi @yersan,
Other modules use these dependencies, so I have not moved them.
Thanks

yersan: In that case, the approach should be to add the `testbom` in the dependency management section of the other modules that need these pure test dependencies.

The general idea is to move all the test dependencies outside of the main pom.xml to avoid incorporating pure test dependencies to WildFly from WildFly Core.

parsharma: Yes, I will move above dependencies also. Thanks, @yersan.

pom.xml
```diff
@@ -245,10 +261,14 @@
         <version.org.wildfly.legacy.test>7.0.0.Final</version.org.wildfly.legacy.test>
         <version.org.wildfly.openssl>2.2.5.Final</version.org.wildfly.openssl>
         <version.org.wildfly.openssl.natives>2.2.2.Final</version.org.wildfly.openssl.natives>
-        <version.org.wildfly.openssl.wildfly-openssl-linux-x86_64>2.2.2.SP01</version.org.wildfly.openssl.wildfly-openssl-linux-x86_64>
-        <version.org.wildfly.openssl.wildfly-openssl-linux-s390x>2.2.2.SP01</version.org.wildfly.openssl.wildfly-openssl-linux-s390x>
-        <version.org.wildfly.openssl.wildfly-openssl-macosx-x86_64>${version.org.wildfly.openssl.natives}</version.org.wildfly.openssl.wildfly-openssl-macosx-x86_64>
-        <version.org.wildfly.openssl.wildfly-openssl-windows-x86_64>${version.org.wildfly.openssl.natives}</version.org.wildfly.openssl.wildfly-openssl-windows-x86_64>
+        <version.org.wildfly.openssl.wildfly-openssl-linux-x86_64>2.2.2.SP01
```
yersan, line 264: It is ok but, but I would suggest trying to avoid code re-formats as much as possible if they are not related to the current PR, and if you do them, add them in a separate commit. 

That would prevent possible conflicts from moving changes between branches and helps with the review

testbom/pom.xml
```diff
@@ -270,6 +270,90 @@
...
+                <artifactId>byteman-bmunit</artifactId>
+                <version>${version.org.jboss.byteman}</version>
+                <scope>test</scope>
+            </dependency>
+            <dependency>
+                <groupId>org.jboss.byteman</groupId>
+                <artifactId>byteman-install</artifactId>
+                <version>${version.org.jboss.byteman}</version>
+                <scope>test</scope>
+            </dependency>
```
yersan, line 356: Could you sort them alphabetically GAV?

yersan (changes requested):
Thanks, @parsharma , ideally an integration Job would help to get a previous result without waiting for the wildfly counterpart to be merged

testbom/pom.xml
parsharma: Sure, I will update it.
Thanks

yersan: @parsharma Sorry it was a bad description, GAV stands for Group, Artifact, Version.

pom.xml
parsharma: Sure, I will make sure to keep it in mind.
Thanks

testbom/pom.xml
parsharma: Thanks for the clarification. I will update the code soon.

parsharma:
This is passed custom Integration job https://ci.wildfly.org/buildConfiguration/WF_WildFlyCoreIntegrationExperiments/353558?hideProblemsFromDependencies=false&hideTestsFromDependencies=false&expandBuildChangesSection=true

yersan (changes requested):
Hi @parsharma , there are still some dependencies in the main pom with the test scope.

These test dependencies should be controlled by the `test-bom` module and incorporated by it into the modules that require them. They should not be distributed by using the parent pom.

I'm not sure if you have found an issue by using this strategy, if so, we can discuss them in Zulip. 

What I would expect after this PR is:

1. No pure test dependencies in the parent pom.
2. Modules that require test dependencies should use the test-bom to import them.
3. All pure test dependency version properties should be declared in the test-bom pom.xml, or we should move these properties altogether to the main pom.xml, but we should avoid to have an inconsistencies where some pure test dependency version properties are declared in the main pom.xml and others in the test-bom. Consolidate them would be good.

parsharma:
@yersan 
Thanks for the explanation. I have worked on each point and pushed the updated code. Please let me know if I am missing anything else. Here is the custom integration job https://ci.wildfly.org/buildConfiguration/WF_WildFlyCoreIntegrationExperiments/353558?hideProblemsFromDependencies=false&hideTestsFromDependencies=false&expandBuildChangesSection=true

yersan:
Requires: https://github.com/wildfly/wildfly/pull/16541

testbom/pom.xml.bak
```diff
@@ -0,0 +1,360 @@
+<?xml version="1.0" encoding="UTF-8"?>
```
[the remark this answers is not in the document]
parsharma: Done

pom.xml
```diff
@@ -253,6 +253,19 @@
...
+        <version.com.fasterxml.jackson.databind>${version.com.fasterxml.jackson}.1</version.com.fasterxml.jackson.databind>
+        <version.commons-io>2.10.0</version.commons-io>
+        <version.commons-collections>3.2.2</version.commons-collections>
+        <version.io.netty>4.1.86.Final</version.io.netty>
+        <version.jakarta.inject.jakarta.inject-api>2.0.1</version.jakarta.inject.jakarta.inject-api>
+        <version.org.apache.velocity>2.3</version.org.apache.velocity>
+        <version.org.codehaus.plexus.plexus-utils>3.5.0</version.org.codehaus.plexus.plexus-utils>
+        <version.org.jboss.spec.javax.servlet.jboss-servlet-api_4.0_spec>2.0.0.Final</version.org.jboss.spec.javax.servlet.jboss-servlet-api_4.0_spec>
+        <version.org.mock-server.mockserver-netty>5.8.1</version.org.mock-server.mockserver-netty>
+        <version.xom.xom>1.3.7</version.xom.xom>
```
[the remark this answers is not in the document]
parsharma: I have updated the version in alphabetical order.

yersan:
@parsharma , thanks!

PR #5360, opened 2023-01-24
[WFCORE-6208] Set signing of tags to the same value as signing commits.

jamezp:
https://issues.redhat.com/browse/WFCORE-6208


cam-rod (approved):
Successfully passes the tests when GPG signing is enabled. Doesn't work when SSH signing is configured, but that's expected since JGit doesn't support it yet.

server/src/main/java/org/jboss/as/server/controller/git/GitRepository.java
```diff
@@ -143,7 +144,7 @@ public GitRepository(GitRepositoryConfiguration gitConfig)
                     git.clean().call();
                     git.pull().setRemote(remoteName).setRemoteBranchName(branch).setStrategy(MergeStrategy.RESOLVE).call();
                     if (createGitIgnore(git, basePath)) {
-                        git.commit().setMessage(ServerLogger.ROOT_LOGGER.addingIgnored()).call();
+                        git.commit().setSign(gitConfig.isSign()).setMessage(ServerLogger.ROOT_LOGGER.addingIgnored()).call();
```
ehsavoie, line 147: @jamezp why aren't you calling setting sign to false here ?

yersan (approved):
LGTM, thanks @cam-rod for your review.

@ehsavoie Could you approve on your side?

ehsavoie (approved):

PR #5370, opened 2023-02-06
[WFCORE-5605] Remove the prompt for checking if this is a server to server process.

darranl:
This is no longer configured as a securet in the security realms.

https://issues.redhat.com/browse/WFCORE-5605


fjuma (approved):
Thanks @darranl!

Skyllarr (approved):

PR #5372, opened 2023-02-13
[WFCORE-6224][WFCORE-6225] Upgrade httpcore to 4.4.16 and  Apache HTT…

yersan:
…P Client to 4.5.14

Jira issue: https://issues.redhat.com/browse/WFCORE-6224
Jira issue: https://issues.redhat.com/browse/WFCORE-6225


yersan:
@jamezp Can you give approval? These dependencies are used on resteasy, they are micro and only incorporate bugfixes, so we should be fine

jamezp (approved):

PR #5377, opened 2023-02-14
[WFCORE-5393] Remove the wildlfy-elytron-auth-server-deprecated dependency module to no longer rely on deprecated classes

sudeshnas93:
Issue: https://issues.redhat.com/browse/WFCORE-5393


darranl (changes requested):
This is going to be a bigger task than just a package rename.

These classes are advertised as capabilities, we also have external feature packs / extensions that work with these types that can be broken if we just change them all without some form of compatibility mode.

darranl:
> This is going to be a bigger task than just a package rename.
> 
> These classes are advertised as capabilities, we also have external feature packs / extensions that work with these types that can be broken if we just change them all without some form of compatibility mode.

The compilation failures in WildFly start to illustrate this problem.

sudeshnas93:
This PR can only be merged after merging this jboss-remoting PR https://github.com/jboss-remoting/jboss-remoting/pull/286

PR #5384, opened 2023-02-16
[WFCORE-6237] Eliminating usage of deprecated javax.api aggregation module

ropalka:
https://issues.redhat.com/browse/WFCORE-6237


yersan (approved):
LGTM, I guess we will have another one removing them for WildFly Full and finally, remove the module from the core, correct me if I'm wrong.

ropalka:
Current plan is to remove usage of deprecated modules.
Then we can decide whether we will ship them (for backward compatibility) or we will drop them.

PR #5389, opened 2023-02-21
[WFCORE-6190] Add the ability to configure virtual security domains

fjuma:
https://issues.redhat.com/browse/WFCORE-6190

This PR also contains https://github.com/wildfly/wildfly-core/pull/5347 (the required Elytron subsystem bump) and https://github.com/wildfly/wildfly-core/pull/5388 (the Elytron component upgrade with a needed bug fix) so this can run through CI.


elytron/src/main/java/org/wildfly/extension/elytron/DomainDefinition.java
```diff
@@ -193,6 +195,13 @@ class DomainDefinition extends SimpleResourceDefinition {
             .setCapabilityReference(SECURITY_DOMAIN_CAPABILITY, SECURITY_DOMAIN_CAPABILITY)
             .build();
 
+    static final StringListAttributeDefinition TRUSTED_VIRTUAL_SECURITY_DOMAINS = new StringListAttributeDefinition.Builder(ElytronDescriptionConstants.TRUSTED_VIRTUAL_SECURITY_DOMAINS)
+            .setRequired(false)
+            .setMinSize(1)
+            .setFlags(AttributeAccess.Flag.RESTART_RESOURCE_SERVICES)
+            .setCapabilityReference(VIRTUAL_SECURITY_DOMAIN_CAPABILITY, SECURITY_DOMAIN_CAPABILITY)
```
OndrejKotek, line 202: The CLI autocomplete feature does not work correctly for this attribute. When there is a dot in the name of a virtual security domain just the extension is offered. For example

```
/subsystem=elytron/virtual-security-domain=abc.war:add
/subsystem=elytron/security-domain=ApplicationDomain:write-attribute(name=trusted-virtual-security-domains,value=[<TAB>
->
/subsystem=elytron/security-domain=ApplicationDomain:write-attribute(name=trusted-virtual-security-domains,value=[war
```
It works correctly for names without a dot.

elytron/src/main/java/org/wildfly/extension/elytron/VirtualDomainDefinition.java
```diff
@@ -0,0 +1,207 @@
...
+            .setMinSize(1)
+            .setFlags(AttributeAccess.Flag.RESTART_RESOURCE_SERVICES)
+            .setCapabilityReference(SECURITY_DOMAIN_CAPABILITY, VIRTUAL_SECURITY_DOMAIN_CAPABILITY)
+            .build();
+
+    private static final AttributeDefinition[] ATTRIBUTES = new AttributeDefinition[] { OUTFLOW_SECURITY_DOMAINS, OUTFLOW_ANONYMOUS };
+
+    private static final VirtualDomainAddHandler ADD = new VirtualDomainAddHandler();
+    private static final OperationStepHandler REMOVE = new VirtualDomainRemoveHandler(ADD);
+    private static final WriteAttributeHandler WRITE = new WriteAttributeHandler(ElytronDescriptionConstants.VIRTUAL_SECURITY_DOMAIN);
```
OndrejKotek, line 82: When adding `virtual-security-domain` with `outflow-security-domains`, the resource is created as expected. When adding an empty `virtual-security-domain` and writing `outflow-security-domains`, an error occurs:

```
/subsystem=elytron/virtual-security-domain=abcde.war:add(outflow-security-domains=[ApplicationDomain,ManagementDomain])
{"outcome" => "success"}


/subsystem=elytron/virtual-security-domain=abc.war:add()
/subsystem=elytron/virtual-security-domain=abc.war:write-attribute(name=outflow-security-domains,value=[ApplicationDomain,ManagementDomain])
{
    "outcome" => "failed",
    "failure-description" => "WFLYCTL0158: Operation handler failed: java.lang.IllegalArgumentException: WFLYCTL0394: Capability 'org.wildfly.security.virtual-security-domain.abc.war' does not provide services of type 'class org.wildfly.security.auth.server.SecurityDomain'",
    "rolled-back" => true
}
```

elytron/src/main/resources/org/wildfly/extension/elytron/LocalDescriptions.properties
```diff
@@ -264,9 +264,20 @@ elytron.security-domain.realms.principal-transformer=A principal transformer to
...
 
+###########################
+# Virtual Security Domain #
+###########################
+
+elytron.virtual-security-domain=A virtual security domain definition.
+elytron.virtual-security-domain.add=The add operation for a virtual security domain definition.
+elytron.virtual-security-domain.remove=The remove operation for a virtual security domain definition.
+elytron.virtual-security-domain.outflow-security-domains=The list of security domains that the security identity from this virtual domain should automatically outflow to.
+elytron.virtual-security-domain.outflow-anonymous=When outflowing to a security domain if outflow is not possible should the anonymous identity be used?  Outflowing anonymous has the effect of clearing any identity already established for that domain.
```
OndrejKotek, line 279: > outflow is not possible

Which cases are covered by this? I see two possible and a clarification would help
- There is an outflow security domain specified which is not configured to trust this virtual security domain
- The outflowed identity is not known in the trusting security domain

elytron/src/main/java/org/wildfly/extension/elytron/DomainDefinition.java
jfdenise: The problem is related to capabilities (CLI completion is based on that). The capability-registry returns the last item after the last dot (https://github.com/wildfly/wildfly-core/blob/main/controller/src/main/java/org/jboss/as/controller/CapabilityRegistry.java#L1034). For example, I created a security domain named foo.war. Then call:
/core-service=capability-registry:suggest-capabilities(name=org.wildfly.security.security-domain, dependent-address=[("subsystem" => "elytron"), ("http-authentication-factory" => "test") ])
You will see that the returned list only contains war.
I am wandering if that would be enough to returns the part of the capability name after the provided capability name. For example: "org.wildfly.security.security-domain" is the provided capability name, "org.wildfly.security.security-domain.foo.war" is the capability of a registration point that matches, we would return "foo.war". Perhaps @bstansberry could provide input there.

jfdenise: Actually we have a conflicting case (capability "org.wildfly.undertow.host") with the CLI command: /subsystem=undertow:write-attribute(name=default-virtual-host,value=TAB that returns the capability "org.wildfly.undertow.host.default-server.default-host", in this case the prefix "default-server." must be removed. That was the driver to return only the last part after the dot (See https://issues.redhat.com/browse/WFCORE-2690)

elytron/src/main/java/org/wildfly/extension/elytron/VirtualDomainDefinition.java
fjuma: Good catch, fixed.

elytron/src/main/resources/org/wildfly/extension/elytron/LocalDescriptions.properties
fjuma: Clarified this, thanks!

fjuma:
@OndrejKotek Thanks for the feedback! I've made the suggested changes.

OndrejKotek:
`org.jboss.as.test.manualmode.management.persistence.YamlExtensionTestCase#testSimpleYaml` and `#testSimpleYamlWithReload` are failing on QE CI. The cause seems to be new line at the end of `testsuite/manualmode/src/test/resources/org/jboss/as/test/manualmode/management/persistence/test.xml`.

ehsavoie:
@yersan yes they are. Maybe I should only look for the resulting changes instead of the full files. Please note that i have a rfe to export to a file as opposed to reading the cli ... but it wasn't merged

fjuma:
> `org.jboss.as.test.manualmode.management.persistence.YamlExtensionTestCase#testSimpleYaml` and `#testSimpleYamlWithReload` are failing on QE CI. The cause seems to be new line at the end of `testsuite/manualmode/src/test/resources/org/jboss/as/test/manualmode/management/persistence/test.xml`.
> 
> See [56fb2c7#diff-256bf24674f5d110f0bbc5c1dcbd32fdb5661c3a3fc38d9301d30e05c8d1851dL235](https://github.com/wildfly/wildfly-core/commit/56fb2c7d7da2129e9ea336836f1ce91edc76a4ac#diff-256bf24674f5d110f0bbc5c1dcbd32fdb5661c3a3fc38d9301d30e05c8d1851dL235)

Thanks @OndrejKotek, I've fixed this.

yersan:
I'm not sure the reason, but I had to manually start this Job since there were no results for this PR after the latest push, @luck3y in case you know something about what could make this to happen.

yersan:
> I'm not sure the reason, but I had to manually start this Job since there were no results for this PR after the latest push, @luck3y in case you know something about what could make this happen.

This has been already discussed in [Zulip](https://wildfly.zulipchat.com/#narrow/stream/174177-ci-servers/topic/Pull.20player.20down.3F), so far we can move on by triggering the Jobs manually.

luck3y:
> > I'm not sure the reason, but I had to manually start this Job since there were no results for this PR after the latest push, @luck3y in case you know something about what could make this happen.
> 
> This has been already discussed in [Zulip](https://wildfly.zulipchat.com/#narrow/stream/174177-ci-servers/topic/Pull.20player.20down.3F), so far we can move on by triggering the Jobs manually.

Hi @yersan pull-player should be working for core, I brought up a replacement instance yesterday. I'll have a look at why this didn't start.

luck3y:
@yersan should work now.

yersan (approved):

yersan:
@fjuma / @OndrejKotek merged, thanks!

PR #5048, opened 2022-04-06
[WFCORE-5859] Add Filesystem integrity support

Ashpan:
Jira: https://issues.redhat.com/browse/WFCORE-5859
To pass CI, this will require an Elytron upgrade.

Analysis document: https://github.com/wildfly/wildfly-proposals/pull/462
Requires: https://issues.redhat.com/browse/ELY-2320
Elytron PR with tests: https://github.com/wildfly-security/wildfly-elytron/pull/1709
WildFly PR with tests and documentation updates: https://github.com/wildfly/wildfly/pull/15411

PR #5393, opened 2023-02-27
[WFCORE-6244] Elytron mapped-role-mapper does not support expressions

cam-rod:
https://issues.redhat.com/browse/WFCORE-6244

Rebased onto #5347 to include upgrade to Elytron model version 17

### DIFF
--- a/elytron/src/main/java/org/wildfly/extension/elytron/ElytronSubsystemTransformers.java
+++ b/elytron/src/main/java/org/wildfly/extension/elytron/ElytronSubsystemTransformers.java
@@ -25,20 +25,23 @@ import static org.wildfly.extension.elytron.ElytronDescriptionConstants.AUTHORIZ
 import static org.wildfly.extension.elytron.ElytronDescriptionConstants.AUTOFLUSH;
 import static org.wildfly.extension.elytron.ElytronDescriptionConstants.BCRYPT_MAPPER;
 import static org.wildfly.extension.elytron.ElytronDescriptionConstants.CERTIFICATE_AUTHORITY;
-import static org.wildfly.extension.elytron.ElytronDescriptionConstants.FILESYSTEM_REALM;
 import static org.wildfly.extension.elytron.ElytronDescriptionConstants.CERTIFICATE_REVOCATION_LISTS;
 import static org.wildfly.extension.elytron.ElytronDescriptionConstants.CREDENTIAL_STORE;
+import static org.wildfly.extension.elytron.ElytronDescriptionConstants.FILESYSTEM_REALM;
 import static org.wildfly.extension.elytron.ElytronDescriptionConstants.FILE_AUDIT_LOG;
+import static org.wildfly.extension.elytron.ElytronDescriptionConstants.FROM;
 import static org.wildfly.extension.elytron.ElytronDescriptionConstants.HASH_CHARSET;
 import static org.wildfly.extension.elytron.ElytronDescriptionConstants.HASH_ENCODING;
 import static org.wildfly.extension.elytron.ElytronDescriptionConstants.JDBC_REALM;
 import static org.wildfly.extension.elytron.ElytronDescriptionConstants.KEY_STORE;
 import static org.wildfly.extension.elytron.ElytronDescriptionConstants.KEY_STORE_ALIAS;
 import static org.wildfly.extension.elytron.ElytronDescriptionConstants.LDAP_REALM;
+import static org.wildfly.extension.elytron.ElytronDescriptionConstants.MAPPED_ROLE_MAPPER;
 import static org.wildfly.extension.elytron.ElytronDescriptionConstants.MODULAR_CRYPT_MAPPER;
 import static org.wildfly.extension.elytron.ElytronDescriptionConstants.PERIODIC_ROTATING_FILE_AUDIT_LOG;
 import static org.wildfly.extension.elytron.ElytronDescriptionConstants.PRINCIPAL_TRANSFORMER;
 import static org.wildfly.extension.elytron.ElytronDescriptionConstants.PROPERTIES_REALM;
+import static org.wildfly.extension.elytron.ElytronDescriptionConstants.ROLE_MAP;
 import static org.wildfly.extension.elytron.ElytronDescriptionConstants.SALTED_SIMPLE_DIGEST_MAPPER;
 import static org.wildfly.extension.elytron.ElytronDescriptionConstants.SALT_ENCODING;
 import static org.wildfly.extension.elytron.ElytronDescriptionConstants.SCRAM_MAPPER;
@@ -47,6 +50,7 @@ import static org.wildfly.extension.elytron.ElytronDescriptionConstants.SECURITY
 import static org.wildfly.extension.elytron.ElytronDescriptionConstants.SIMPLE_DIGEST_MAPPER;
 import static org.wildfly.extension.elytron.ElytronDescriptionConstants.SIZE_ROTATING_FILE_AUDIT_LOG;
 import static org.wildfly.extension.elytron.ElytronDescriptionConstants.SYNCHRONIZED;
+import static org.wildfly.extension.elytron.ElytronDescriptionConstants.TO;
 import static org.wildfly.extension.elytron.ElytronDescriptionConstants.TRUSTED_VIRTUAL_SECURITY_DOMAINS;
 import static org.wildfly.extension.elytron.ElytronExtension.ELYTRON_10_0_0;
 import static org.wildfly.extension.elytron.ElytronExtension.ELYTRON_11_0_0;
@@ -88,6 +92,7 @@ import org.jboss.as.controller.transform.description.RejectAttributeChecker;
 import org.jboss.as.controller.transform.description.ResourceTransformationDescriptionBuilder;
 import org.jboss.as.controller.transform.description.TransformationDescriptionBuilder;
 import org.jboss.dmr.ModelNode;
+import org.jboss.dmr.ModelType;
 import org.wildfly.security.password.interfaces.ScramDigestPassword;
 import org.wildfly.security.x500.cert.acme.CertificateAuthority;
 
@@ -153,6 +158,12 @@ public final class ElytronSubsystemTransformers implements ExtensionTransformerR
                 .getAttributeBuilder()
                 .addRejectCheck(RejectAttributeChecker.DEFINED, TRUSTED_VIRTUAL_SECURITY_DOMAINS)
                 .setDiscard(DiscardAttributeChecker.UNDEFINED, TRUSTED_VIRTUAL_SECURITY_DOMAINS);
+
+        builder.addChildResource(PathElement.pathElement(MAPPED_ROLE_MAPPER))
+                .getAttributeBuilder()
+                .setDiscard(DiscardAttributeChecker.UNDEFINED, FROM)
+                .setDiscard(DiscardAttributeChecker.UNDEFINED, TO)
+                .setValueConverter(MAPPED_ROLE_MAPPER_CONVERTER, ROLE_MAP);
     }
 
     private static void from16(ChainedTransformationDescriptionBuilder chainedBuilder) {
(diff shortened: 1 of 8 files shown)


elytron/src/main/resources/org/wildfly/extension/elytron/LocalDescriptions.properties
```diff
@@ -711,7 +711,9 @@ elytron.mapped-role-mapper.remove=The remove operation for the role mapper.
 # Attributes
 elytron.mapped-role-mapper.keep-mapped=When set to 'true' the mapped roles will retain all roles, that have defined mappings.
 elytron.mapped-role-mapper.keep-non-mapped=When set to 'true' the mapped roles will retain all roles, that have no defined mappings.
-elytron.mapped-role-mapper.role-map=A string to string list map for mapping roles.
+elytron.mapped-role-mapper.role-map=A list of mappings to use for mapping roles.
+elytron.mapped-role-mapper.role-map.from=The role to match for mapping roles.
+elytron.mapped-role-mapper.role-map.to=A string list of roles to map to.
```
Skyllarr, line 716: Just a note for other reviewers, these are not new elements, they were present in the XML schema before as well. 

This PR looks great to me. CLI command changed but old format continues to work, so I don't think RFE process is needed, but will let @fjuma to confirm. @cam-rod LGTM

Skyllarr (approved):

yersan (commented):
Since the new format is not understood by WildFly previous versions, even if the change is small, we would need to bump the model and add the required transformers for the mixed domain.

The point is elytron subsystem is under `/profile` resource in a domain mode, so changes on the elytron subsystem under this resource will be sent to the remote hosts in a Domain Mode. We need to transform the format to the old one so the operation is understood by the legacy versions.

cam-rod:
Alright. This is my first update of this kind; should an RFE be opened, and generally where do transformers need to be added?

yersan:
Hi @cam-rod 

> should an RFE be opened

I always have difficulties tracing the line between the need and not need for an RFE. In general, issues impacting how a user configures the server would deserve an RFE since it will imply the required documentation changes will be done. However, if from the beginning it was expected to have those fields supporting expressions, then handling it as a bug looks appropriate to me. Was it expected from the beginning that those fields must support expressions or is it something we are adding now? 

If we do not have a response to this question, maybe @fjuma or @bstansberry can trace the line here and give us a point about how we should treat this one.


> and generally where do transformers need to be added?

Sometimes, configuring the transformers is not a simple strait forward change. You could start looking at the code impacting a similar issue in Elytron, since how and where each subsystem defines the transformers usually is particular to each subsystem. 

In Elytron, I can see the transformers are defined here: [ElytronSubsystemTransformers.java
](https://github.com/wildfly/wildfly-core/blob/main/elytron/src/main/java/org/wildfly/extension/elytron/ElytronSubsystemTransformers.java)

You could take a look at https://issues.redhat.com/browse/WFCORE-5859 which added some transformers to reject the resource that is not understood by the legacy controller.

For your issue, you would need a different transformer, since you don't want to reject the operation, you need to change it to adapt the attributes to the old format when needed. I believe you would need to implement an [OperationTransformer.java](https://github.com/wildfly/wildfly-core/blob/main/controller/src/main/java/org/jboss/as/controller/transform/OperationTransformer.java).

I suggest looking at them and asking in Zulip concrete questions about this stuff.

yersan:
> My assumption was yes, based on the [original Core PR](https://github.com/wildfly/wildfly-core/pull/5048/files#diff-04d6c66e9456cebe5f60c8a28a1e24269021ca166c5c52dca0d782e9dccf4a6eR161-R179) and the [wildscribe documentation](https://docs.wildfly.org/27/wildscribe/subsystem/elytron/mapped-role-mapper/).

Then, my interpretation is it can be handled via a bug. Let's hear other opinions to see if support ours.

cam-rod:
Thanks @darranl

elytron/src/main/java/org/wildfly/extension/elytron/ElytronSubsystemTransformers.java
```diff
@@ -153,6 +158,28 @@ private static void from17(ChainedTransformationDescriptionBuilder chainedBuilde
...
+                     * Attributes formatted as a list of maps (ObjectListAttributeDefinition) should be rejected. */
+                    @Override
+                    protected boolean rejectAttribute(PathAddress address, String attributeName, ModelNode attributeValue, TransformationContext context) {
+                        return attributeValue.getType() != ModelType.OBJECT;
+                    }
+
+                    @Override
+                    public String getRejectionLogMessage(Map<String, ModelNode> attributes) {
+                        return ROOT_LOGGER.mappedRoleMapExpectsObject(
+                                attributes.get("role-map") != null ? attributes.get("role-map").getType().toString() : "not")
```
fjuma, line 179: We shouldn't hardcode the word "not" here (all message strings should be in the messages class. Maybe the error message could be made more generic, perhaps something along the lines of:

"Value for attribute `role-map` is invalid"

elytron/src/main/java/org/wildfly/extension/elytron/RoleMapperDefinitions.java
```diff
@@ -154,15 +181,15 @@ protected ValueSupplier<RoleMapper> getValueSupplier(OperationContext context, M
...
-                    Set<String> set = new LinkedHashSet<>();
-                    for (ModelNode m : currentList) {
-                        set.add(m.asString());
-                    }
-                    roleMappingMap.put(s, set);
+                List<ModelNode> mappings = roleMappingMapNode.asList();
+                final Map<String, Set<String>> roleMappingMap = new LinkedHashMap<>(mappings.size());
+
+                for (ModelNode mapping : mappings) {
+                    String mapFrom = FROM.resolveModelAttribute(context, mapping).asStringOrNull();
```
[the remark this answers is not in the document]
cam-rod: It probably is required, will update.

bstansberry:
@yersan @cam-rod @darranl @fjuma Re the process discussion, I think this is fine to go in as a bug fix.  It isn't really adding "new" user controllable/configurable behavior (which is the simplest definition of a "Feature"); it's just fixing the handling of existing behavior.  And the compatibility work minimizes and impact that might arise from old docs out there somewhere that describe the old approach.

If you haven't done so already, a quick scan of downstream docs looking for coverage of this would be good, with an update issue filed if you find any. But I don't regard that as a blocker for merging this upstream, as any update issue would be of not be critical or higher priority; probably could be minor.

elytron/src/main/java/org/wildfly/extension/elytron/RoleMapperDefinitions.java
```diff
@@ -154,15 +181,15 @@ protected ValueSupplier<RoleMapper> getValueSupplier(OperationContext context, M
...
-                    }
-                    roleMappingMap.put(s, set);
+                List<ModelNode> mappings = roleMappingMapNode.asList();
+                final Map<String, Set<String>> roleMappingMap = new LinkedHashMap<>(mappings.size());
+
+                for (ModelNode mapping : mappings) {
+                    String mapFrom = FROM.resolveModelAttribute(context, mapping).asString();
+                    List<ModelNode> mapTo = TO.resolveModelAttribute(context, mapping).asList();
+                    Set<String> mapToSet = new LinkedHashSet<>();
+                    mapTo.forEach((m) -> mapToSet.add(m.asString()));
```
fjuma, line 191: We can use a normal for loop here.

cam-rod:
Took a quick look, no mentions in downstream docs.

testsuite/elytron/src/test/java/org/wildfly/test/security/common/elytron/MappedRoleMapperTest.java
```diff
@@ -0,0 +1,126 @@
...
+    CLIWrapper cli;
+
+    @Before
+    public void setup() throws Exception {
+        cli = new CLIWrapper(true);
+    }
+
+    @After
+    public void cleanup() throws Exception {
+        cli.sendLine("reload", true);
```
fjuma, line 41: We also need to make sure that all of the resources that were added in this test get removed.

testsuite/elytron/src/test/java/org/wildfly/test/security/common/elytron/MappedRoleMapperTest.java
```diff
@@ -0,0 +1,135 @@
...
+
+        success = cli.sendLine("/subsystem=elytron/mapped-role-mapper=mrm:write-attribute(name=keep-mapped,value=false)", true);
+        Assert.assertTrue(success);
+        cli.sendLine("reload", true);
+
+        success = cli.sendLine("/subsystem=elytron/mapped-role-mapper=mrm:read-attribute(name=keep-mapped)", true);
+        Assert.assertTrue(success);
+        MatcherAssert.assertThat(cli.readOutput(), containsString("\"result\" => false"));
+
+        success = cli.sendLine("/subsystem=elytron/mapped-role-mapper=mrm:remove", true);
```
fjuma, line 88: Thanks for all the updates, @cam-rod! This is looking good. Just one final comment - because the resource removal is being handled in the `cleanup` method, the additional calls to `remove` here and below can be deleted.

cam-rod: Thanks, removed the calls

fjuma (approved):
Thanks very much @cam-rod!

elytron/src/main/java/org/wildfly/extension/elytron/ElytronSubsystemTransformers.java
```diff
@@ -153,6 +158,26 @@ private static void from17(ChainedTransformationDescriptionBuilder chainedBuilde
...
+                    /* [WFCORE-6244] role-map attribute was previously formatted (directly) as a string to string-list mapping of roles.
+                     * Attributes formatted as a list of maps (ObjectListAttributeDefinition) should be rejected. */
+                    @Override
+                    protected boolean rejectAttribute(PathAddress address, String attributeName, ModelNode attributeValue, TransformationContext context) {
+                        return attributeValue.getType() != ModelType.OBJECT;
+                    }
+
+                    @Override
+                    public String getRejectionLogMessage(Map<String, ModelNode> attributes) {
+                        return ROOT_LOGGER.mappedRoleMapExpectsObject().getMessage();
```
yersan, line 178: The rejection message is not the expected one. 

The attribute is rejected when it is a list of Objects but the Rejection message is:

`Value type for attribute `role-map` is invalid. Expected [LIST] of [OBJECT].`

It should be something that describes what is expected, in this case, something like it is expected a [OBJECT] but got a LIST.

yersan (changes requested):

elytron/src/main/java/org/wildfly/extension/elytron/ElytronSubsystemTransformers.java
cam-rod: @fjuma referring back to [this comment](https://github.com/wildfly/wildfly-core/pull/5393#discussion_r1149454732), I'll restore this message to the format I previously had. 

Should I just try to detect the type of object, and allow the `getType` command to throw an Exception if it fails? Or would it be better to just create an if-else statement, where the alternate message without the type is thrown if 'role-map` attribute doesn't exist?

cam-rod: alright, I'll remove the case then, thanks

cam-rod: Talked with @fjuma about the issue, i ended up replacing this method with a ValueConverter, since only the model needs to be updated (the schema hasn't changed).

yersan: ok, that makes more sense and was our original suggestion. I assumed you wanted to reject it instead of doing a conversion.

cam-rod:
The Windows build previously failed due to the issue mentioned [in this comment](https://github.com/wildfly/wildfly-core/pull/5429#pullrequestreview-1366790809), but passed in all other cases. Since I haven't rebased, the rerun may fail in the same spot.

yersan (changes requested):

elytron/src/main/java/org/wildfly/extension/elytron/ElytronSubsystemTransformers.java
```diff
@@ -581,6 +591,24 @@ protected void convertAttribute(PathAddress address, String attributeName, Model
...
+     * back to a direct string to string-list mapping:      {STRING=[STRING,STRING,...],...}
+     *
+     * XML schema remains unchanged. */
+    private static final AttributeConverter MAPPED_ROLE_MAPPER_CONVERTER = new AttributeConverter.DefaultAttributeConverter() {
+        @Override
+        protected void convertAttribute(PathAddress address, String attributeName, ModelNode attributeValue, TransformationContext context) {
+            if (attributeValue.isDefined()) {
+                ModelNode oldRoleMappingMap = new ModelNode();
+                for (ModelNode mapping : attributeValue.asList()) {
+                    oldRoleMappingMap.add(mapping.get(FROM).asString(), mapping.get(TO));
```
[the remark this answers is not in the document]
yersan: I guess you can just add an additional check to the if sentence so `if (attribute is defined && is List) then ...`
The fact that the attribute is defined is not enough in this case since both types will go throughout the conversion.

yersan (approved):

staerion:
Thanks for your work on this!

cam-rod:
No problem! It should arrive in an upcoming WildFly release.

PR #5401, opened 2023-03-04
WFCORE-6257 Refactor Schema related interfaces from wildfly-clustering-common into wildfly-controller

pferraro:
https://issues.redhat.com/browse/WFCORE-6257

Requires https://github.com/wildfly/wildfly-core/pull/5400


yersan (approved):

yersan:
Sorry @pferraro I thought I was going to do it when merging the Pull Request but I'm only able to change the commit message only for the merge commit. 

Having the Jira number to the commit would help track the change from the code in the future.

pferraro:
> Would you mind adding the Jira number to each commit message?

No problem.

pferraro:
> Would you mind adding the Jira number to each commit message?

Done.

PR #5402, opened 2023-03-04
WFCORE-6258 Refactor Model interface from wildfly-clustering-common into wildfly-controller

pferraro:
https://issues.redhat.com/browse/WFCORE-6258

### DIFF
--- a/controller/src/main/java/org/jboss/as/controller/SubsystemModel.java
+++ b/controller/src/main/java/org/jboss/as/controller/SubsystemModel.java
@@ -1,0 +1,37 @@
+/*
+ * Copyright 2023 Red Hat, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *   http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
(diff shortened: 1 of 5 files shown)


yersan (commented):
Is there any reason for not making the existing wildfly core subsystems implement this new interface in this PR?

controller/src/main/java/org/jboss/as/controller/SubsystemModel.java
```diff
@@ -0,0 +1,44 @@
...
+ * This software is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the GNU
+ * Lesser General Public License for more details.
+ *
+ * You should have received a copy of the GNU Lesser General Public
+ * License along with this software; if not, write to the Free
+ * Software Foundation, Inc., 51 Franklin St, Fifth Floor, Boston, MA
+ * 02110-1301 USA, or see the FSF site: http://www.fsf.org.
+ */
```
yersan, line 21: All WildFly Core new files should use Apache license instead of LGPL

yersan (changes requested):

controller/src/main/java/org/jboss/as/controller/SubsystemModel.java
pferraro: Fixed.

pferraro:
> Is there any reason for not making the existing wildfly core subsystems implement this new interface in this PR?

Enumerating supported model versions for every subsystem is probably best done by the corresponding component owner - as this would otherwise require some time to investigate the history of every subsystem.
That said, I should at least do this for the core components that I own, i.e. discovery.

yersan (approved):

yersan:

pferraro:
> Would you mind adding the Jira number to each commit message?, we don't need to wait for a full CI, since it already passed.

Done.

PR #5403, opened 2023-03-07
WFCORE-6259 Refactor SubsystemResourceDescriptionResolver implementation from wildfly-clustering-common to wildfly-controller

pferraro:
https://issues.redhat.com/browse/WFCORE-6259

### DIFF
--- a/controller/pom.xml
+++ b/controller/pom.xml
@@ -147,6 +147,11 @@
            <artifactId>junit</artifactId>
            <scope>test</scope>
         </dependency>
+        <dependency>
+           <groupId>org.mockito</groupId>
+           <artifactId>mockito-core</artifactId>
+           <scope>test</scope>
+        </dependency>
 
         <dependency>
             <groupId>org.jboss.logging</groupId>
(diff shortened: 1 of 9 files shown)


controller/pom.xml
```diff
@@ -147,6 +147,11 @@
            <artifactId>junit</artifactId>
            <scope>test</scope>
         </dependency>
+        <dependency>
+           <groupId>org.mockito</groupId>
+           <artifactId>mockito-core</artifactId>
+           <scope>test</scope>
+        </dependency>
```
yersan, line 154: We are planning to move all test dependencies to the testbom. I think it reasonable to merge this as it is. 

@parsharma We will need to take into account this new test dependency and move it to the testbom. Sadly your PR is still in the queue since we need to get the WildFly Full counterpart merged before merging yours.

yersan (approved):

yersan:

pferraro:
> Would you mind adding the Jira number to each commit message?, we don't need to wait for a full CI, since it already passed.

Done.

PR #5412, opened 2023-03-17
[WFCORE-6264] Add additional methods to DomainTestUtils to work with …

yersan:
…an Operation object

Jira issue: https://issues.redhat.com/browse/WFCORE-6264


yersan:
This is a harmless change, so I'm approving it

PR #5413, opened 2023-03-20
WFCORE-6221 Add formal experimental/preview mode to management model

pferraro:
https://issues.redhat.com/browse/WFCORE-6221

This pull request formalizes the concept of a quality level in the WildFly kernel based on the general strategy described in WFCORE-6221.
In general, a subsystem associate features within its management model with a specific Quality.  During management registration, only those features enabled by the quality level of the server will be registered.

See the test case here for an example of how I imagined this might work: https://github.com/pferraro/wildfly-core/tree/WFCORE-6221/subsystem-test/tests/src/test/java/org/jboss/as/subsystem/test/quality

Please scrutinize.  Tell me what you like and do not like.

### DIFF
--- a/controller/src/main/java/org/jboss/as/controller/registry/ManagementResourceRegistration.java
+++ b/controller/src/main/java/org/jboss/as/controller/registry/ManagementResourceRegistration.java
@@ -67,7 +67,8 @@ public interface ManagementResourceRegistration extends ImmutableManagementResou
      *
      * @param resourceDefinition source for descriptive information describing this
      *                            portion of the model (must not be {@code null})
-     * @return a resource registration which may be used to add attributes, operations, notifications and sub-models
+     * @return a resource registration which may be used to add attributes, operations, notifications and sub-models,
+     *         or null, if this resource definition is not enabled by the current stability level.
      *
      * @throws IllegalArgumentException if a submodel is already registered at {@code address}
      * @throws IllegalStateException if {@link #isRuntimeOnly()} returns {@code true}
@@ -103,7 +104,7 @@ public interface ManagementResourceRegistration extends ImmutableManagementResou
      * @param name the specific name of the resource. Cannot be {@code null} or {@link PathElement#WILDCARD_VALUE}
      * @param descriptionProvider provider for descriptions of the additional attributes or child types
      *
-     * @return a resource registration which may be used to add attributes, operations and sub-models
+     * @return a resource registration which may be used to add attributes, operations and sub-models.
      *
      * @throws IllegalArgumentException if either parameter is null or if there is already a registration under {@code name}
      * @throws IllegalStateException if {@link #isRuntimeOnly()} returns {@code true} or if {@link #isAllowsOverride()} returns false
@@ -118,7 +119,8 @@ public interface ManagementResourceRegistration extends ImmutableManagementResou
      * @param registration the child registration of this registry that should no longer be available
      * @param descriptionProvider provider for descriptions of the additional attributes or child types
      *
-     * @return a resource registration which may be used to add attributes, operations and sub-models
+     * @return a resource registration which may be used to add attributes, operations and sub-models,
+     *         or null, if this resource definition is not enabled by the current stability level.
      *
      * @throws IllegalArgumentException if either parameter is null or if there is already a registration under {@code name}
      * @throws IllegalStateException if {@link #isRuntimeOnly()} returns {@code true} or if {@link #isAllowsOverride()} returns false
--- a/host-controller/src/main/java/org/jboss/as/host/controller/HostControllerEnvironment.java
+++ b/host-controller/src/main/java/org/jboss/as/host/controller/HostControllerEnvironment.java
@@ -485,7 +485,7 @@ public class HostControllerEnvironment extends ProcessEnvironment {
         this.processType = processType;
 
         this.stability = getEnumProperty(hostSystemProperties, STABILITY, this.productConfig.getDefaultStability());
-        if (!this.productConfig.getMinimumStability().enables(this.stability)) {
+        if (!this.productConfig.getStabilitySet().contains(this.stability)) {
             throw HostControllerLogger.ROOT_LOGGER.unsupportedStability(this.stability, this.productConfig.getProductName());
         }
         if (!hostSystemProperties.containsKey(STABILITY)) {
--- a/host-controller/src/main/java/org/jboss/as/host/controller/Main.java
+++ b/host-controller/src/main/java/org/jboss/as/host/controller/Main.java
@@ -118,7 +118,7 @@ public final class Main {
             final long startTime = Module.getStartTime();
             final HostControllerEnvironmentWrapper hostControllerEnvironmentWrapper = determineEnvironment(args, startTime);
             if (hostControllerEnvironmentWrapper.getHostControllerEnvironment() == null) {
-                usage(); // In case there was an error determining the environment print the usage
+                usage(hostControllerEnvironmentWrapper.getProductConfig()); // In case there was an error determining the environment print the usage
                 if (hostControllerEnvironmentWrapper.getHostControllerEnvironmentStatus() == HostControllerEnvironmentWrapper.HostControllerEnvironmentStatus.ERROR) {
                     abort();
                 } else {
@@ -174,8 +174,8 @@ public final class Main {
         SystemExiter.abort(ExitCodes.FAILED);
     }
 
-    private static void usage() {
-        CommandLineArgumentUsageImpl.printUsage(STDOUT);
+    private static void usage(ProductConfig productConfig) {
+        CommandLineArgumentUsageImpl.printUsage(productConfig, STDOUT);
     }
 
     public static HostControllerEnvironmentWrapper determineEnvironment(String[] args, long startTime) {
@@ -196,7 +196,7 @@ public final class Main {
         String initialHostConfig = null;
         RunningMode initialRunningMode = RunningMode.NORMAL;
         Map<String, String> hostSystemProperties = getHostSystemProperties();
-        ProductConfig productConfig;
+        ProductConfig productConfig = ProductConfig.fromFilesystemSlot(Module.getBootModuleLoader(), WildFlySecurityManager.getPropertyPrivileged(HostControllerEnvironment.HOME_DIR, null), null);
         ConfigurationFile.InteractionPolicy hostConfigInteractionPolicy = ConfigurationFile.InteractionPolicy.STANDARD;
         ConfigurationFile.InteractionPolicy domainConfigInteractionPolicy = ConfigurationFile.InteractionPolicy.STANDARD;
         String modulePath = null;
@@ -215,22 +215,22 @@ public final class Main {
                         || CommandLineConstants.SHORT_PROPERTIES.equals(arg)) {
                     // Set system properties from url/file
                     if (!processProperties(arg, args[++i], hostSystemProperties)) {
-                        return new HostControllerEnvironmentWrapper(HostControllerEnvironmentWrapper.HostControllerEnvironmentStatus.ERROR);
+                        return new HostControllerEnvironmentWrapper(HostControllerEnvironmentWrapper.HostControllerEnvironmentStatus.ERROR, productConfig);
                     }
                 } else if (arg.startsWith(CommandLineConstants.PROPERTIES)) {
                     String urlSpec = parseValue(arg, CommandLineConstants.PROPERTIES);
                     if (urlSpec == null || !processProperties(arg, urlSpec, hostSystemProperties)) {
-                        return new HostControllerEnvironmentWrapper(HostControllerEnvironmentWrapper.HostControllerEnvironmentStatus.ERROR);
+                        return new HostControllerEnvironmentWrapper(HostControllerEnvironmentWrapper.HostControllerEnvironmentStatus.ERROR, productConfig);
                     }
                 } else if (arg.startsWith(CommandLineConstants.SHORT_PROPERTIES)) {
                     String urlSpec = parseValue(arg, CommandLineConstants.SHORT_PROPERTIES);
                     if (urlSpec == null || !processProperties(arg, urlSpec, hostSystemProperties)) {
-                        return new HostControllerEnvironmentWrapper(HostControllerEnvironmentWrapper.HostControllerEnvironmentStatus.ERROR);
+                        return new HostControllerEnvironmentWrapper(HostControllerEnvironmentWrapper.HostControllerEnvironmentStatus.ERROR, productConfig);
                     }
                 }  else if (arg.startsWith(CommandLineConstants.OLD_PROPERTIES)) {
                     String urlSpec = parseValue(arg, CommandLineConstants.OLD_PROPERTIES);
                     if (urlSpec == null || !processProperties(arg, urlSpec, hostSystemProperties)) {
-                        return new HostControllerEnvironmentWrapper(HostControllerEnvironmentWrapper.HostControllerEnvironmentStatus.ERROR);
+                        return new HostControllerEnvironmentWrapper(HostControllerEnvironmentWrapper.HostControllerEnvironmentStatus.ERROR, productConfig);
                     }
                 } else if (CommandLineConstants.PROCESS_CONTROLLER_BIND_PORT.equals(arg)) {
                     final String port = args[++i];
@@ -238,16 +238,16 @@ public final class Main {
                         pmPort = Integer.valueOf(port);
                     } catch (NumberFormatException e) {
                         STDERR.println(HostControllerLogger.ROOT_LOGGER.invalidValue(CommandLineConstants.PROCESS_CONTROLLER_BIND_PORT, "Integer", port, usageNote()));
-                        return new HostControllerEnvironmentWrapper(HostControllerEnvironmentWrapper.HostControllerEnvironmentStatus.ERROR);
+                        return new HostControllerEnvironmentWrapper(HostControllerEnvironmentWrapper.HostControllerEnvironmentStatus.ERROR, productConfig);
                     }
                 } else if (arg.startsWith(CommandLineConstants.PROCESS_CONTROLLER_BIND_PORT)) {
                     String val = parseValue(arg, CommandLineConstants.PROCESS_CONTROLLER_BIND_PORT);
                     if (val == null) {
-                        return new HostControllerEnvironmentWrapper(HostControllerEnvironmentWrapper.HostControllerEnvironmentStatus.ERROR);
+                        return new HostControllerEnvironmentWrapper(HostControllerEnvironmentWrapper.HostControllerEnvironmentStatus.ERROR, productConfig);
                     }
                     final Integer port = parsePort(val, CommandLineConstants.PROCESS_CONTROLLER_BIND_PORT);
                     if (port == null) {
-                        return new HostControllerEnvironmentWrapper(HostControllerEnvironmentWrapper.HostControllerEnvironmentStatus.ERROR);
+                        return new HostControllerEnvironmentWrapper(HostControllerEnvironmentWrapper.HostControllerEnvironmentStatus.ERROR, productConfig);
                     }
                     pmPort = port;
                 } else if (CommandLineConstants.PROCESS_CONTROLLER_BIND_ADDR.equals(arg)) {
@@ -256,21 +256,21 @@ public final class Main {
                         pmAddress = InetAddress.getByName(addr);
                     } catch (UnknownHostException e) {
                         STDERR.println(HostControllerLogger.ROOT_LOGGER.unknownHostValue(CommandLineConstants.PROCESS_CONTROLLER_BIND_ADDR, addr, usageNote()));
-                        return new HostControllerEnvironmentWrapper(HostControllerEnvironmentWrapper.HostControllerEnvironmentStatus.ERROR);
+                        return new HostControllerEnvironmentWrapper(HostControllerEnvironmentWrapper.HostControllerEnvironmentStatus.ERROR, productConfig);
                     }
                 } else if (arg.startsWith(CommandLineConstants.PROCESS_CONTROLLER_BIND_ADDR)) {
                     final String val = parseValue(arg, CommandLineConstants.PROCESS_CONTROLLER_BIND_ADDR);
                     if (val == null) {
-                        return new HostControllerEnvironmentWrapper(HostControllerEnvironmentWrapper.HostControllerEnvironmentStatus.ERROR);
+                        return new HostControllerEnvironmentWrapper(HostControllerEnvironmentWrapper.HostControllerEnvironmentStatus.ERROR, productConfig);
                     }
                     final InetAddress addr = parseAddress(val, arg);
                     if (addr == null) {
-                        return new HostControllerEnvironmentWrapper(HostControllerEnvironmentWrapper.HostControllerEnvironmentStatus.ERROR);
+                        return new HostControllerEnvironmentWrapper(HostControllerEnvironmentWrapper.HostControllerEnvironmentStatus.ERROR, productConfig);
                     }
                     pmAddress = addr;
                 } else if (pcSocketConfig.processPCSocketConfigArgument(arg, args, i)) {
                     if (pcSocketConfig.isParseFailed()) {
-                        return new HostControllerEnvironmentWrapper(HostControllerEnvironmentWrapper.HostControllerEnvironmentStatus.ERROR);
+                        return new HostControllerEnvironmentWrapper(HostControllerEnvironmentWrapper.HostControllerEnvironmentStatus.ERROR, productConfig);
                     }
                     i += pcSocketConfig.getArgIncrement();
                 } else if (CommandLineConstants.RESTART_HOST_CONTROLLER.equals(arg)) {
@@ -282,31 +282,31 @@ public final class Main {
                 } else if(CommandLineConstants.DEFAULT_JVM.equals(arg) || CommandLineConstants.OLD_DEFAULT_JVM.equals(arg)) {
                     defaultJVM = checkValueIsNotAnArg(arg, args[++i]);
                     if (defaultJVM == null) {
-                        return new HostControllerEnvironmentWrapper(HostControllerEnvironmentWrapper.HostControllerEnvironmentStatus.ERROR);
+                        return new HostControllerEnvironmentWrapper(HostControllerEnvironmentWrapper.HostControllerEnvironmentStatus.ERROR, productConfig);
                     }
                 } else if (CommandLineConstants.DOMAIN_CONFIG.equals(arg)
                         || CommandLineConstants.SHORT_DOMAIN_CONFIG.equals(arg)
                         || CommandLineConstants.OLD_DOMAIN_CONFIG.equals(arg)) {
                     domainConfig = checkValueIsNotAnArg(arg, args[++i]);
                     if (domainConfig == null) {
-                        return new HostControllerEnvironmentWrapper(HostControllerEnvironmentWrapper.HostControllerEnvironmentStatus.ERROR);
+                        return new HostControllerEnvironmentWrapper(HostControllerEnvironmentWrapper.HostControllerEnvironmentStatus.ERROR, productConfig);
                     }
                 } else if (arg.startsWith(CommandLineConstants.DOMAIN_CONFIG)) {
                     String val = parseValue(arg, CommandLineConstants.DOMAIN_CONFIG);
                     if (val == null) {
-                        return new HostControllerEnvironmentWrapper(HostControllerEnvironmentWrapper.HostControllerEnvironmentStatus.ERROR);
+                        return new HostControllerEnvironmentWrapper(HostControllerEnvironmentWrapper.HostControllerEnvironmentStatus.ERROR, productConfig);
                     }
                     domainConfig = val;
                 } else if (arg.startsWith(CommandLineConstants.SHORT_DOMAIN_CONFIG)) {
                     String val = parseValue(arg, CommandLineConstants.SHORT_DOMAIN_CONFIG);
                     if (val == null) {
-                        return new HostControllerEnvironmentWrapper(HostControllerEnvironmentWrapper.HostControllerEnvironmentStatus.ERROR);
+                        return new HostControllerEnvironmentWrapper(HostControllerEnvironmentWrapper.HostControllerEnvironmentStatus.ERROR, productConfig);
                     }
                     domainConfig = val;
                 } else if (arg.startsWith(CommandLineConstants.OLD_DOMAIN_CONFIG)) {
                     String val = parseValue(arg, CommandLineConstants.OLD_DOMAIN_CONFIG);
                     if (val == null) {
-                        return new HostControllerEnvironmentWrapper(HostControllerEnvironmentWrapper.HostControllerEnvironmentStatus.ERROR);
+                        return new HostControllerEnvironmentWrapper(HostControllerEnvironmentWrapper.HostControllerEnvironmentStatus.ERROR, productConfig);
                     }
                     domainConfig = val;
                 } else if (processType == ProcessType.EMBEDDED_HOST_CONTROLLER && arg.startsWith("--empty-host-config")) {
@@ -326,41 +326,41 @@ public final class Main {
                     initialDomainConfig = parseValue(arg, CommandLineConstants.READ_ONLY_DOMAIN_CONFIG);
                     domainConfigInteractionPolicy = ConfigurationFile.InteractionPolicy.READ_ONLY;
                     if (initialDomainConfig == null) {
-                        return new HostControllerEnvironmentWrapper(HostControllerEnvironmentWrapper.HostControllerEnvironmentStatus.ERROR);
+                        return new HostControllerEnvironmentWrapper(HostControllerEnvironmentWrapper.HostControllerEnvironmentStatus.ERROR, productConfig);
                     }
                 } else if (CommandLineConstants.HOST_CONFIG.equals(arg) || CommandLineConstants.OLD_HOST_CONFIG.equals(arg)) {
                     hostConfig = checkValueIsNotAnArg(arg, args[++i]);
                     if (hostConfig == null) {
-                        return new HostControllerEnvironmentWrapper(HostControllerEnvironmentWrapper.HostControllerEnvironmentStatus.ERROR);
+                        return new HostControllerEnvironmentWrapper(HostControllerEnvironmentWrapper.HostControllerEnvironmentStatus.ERROR, productConfig);
                     }
                 } else if (arg.startsWith(CommandLineConstants.HOST_CONFIG)) {
                     String val = parseValue(arg, CommandLineConstants.HOST_CONFIG);
                     if (val == null) {
-                        return new HostControllerEnvironmentWrapper(HostControllerEnvironmentWrapper.HostControllerEnvironmentStatus.ERROR);
+                        return new HostControllerEnvironmentWrapper(HostControllerEnvironmentWrapper.HostControllerEnvironmentStatus.ERROR, productConfig);
                     }
                     hostConfig = val;
                 } else if (arg.startsWith(CommandLineConstants.OLD_HOST_CONFIG)) {
                     String val = parseValue(arg, CommandLineConstants.OLD_HOST_CONFIG);
                     if (val == null) {
-                        return new HostControllerEnvironmentWrapper(HostControllerEnvironmentWrapper.HostControllerEnvironmentStatus.ERROR);
+                        return new HostControllerEnvironmentWrapper(HostControllerEnvironmentWrapper.HostControllerEnvironmentStatus.ERROR, productConfig);
                     }
                     hostConfig = val;
                 } else if (arg.startsWith(CommandLineConstants.READ_ONLY_HOST_CONFIG)) {
                     initialHostConfig = parseValue(arg, CommandLineConstants.READ_ONLY_HOST_CONFIG);
                     hostConfigInteractionPolicy = ConfigurationFile.InteractionPolicy.READ_ONLY;
                     if (initialHostConfig == null) {
-                        return new HostControllerEnvironmentWrapper(HostControllerEnvironmentWrapper.HostControllerEnvironmentStatus.ERROR);
+                        return new HostControllerEnvironmentWrapper(HostControllerEnvironmentWrapper.HostControllerEnvironmentStatus.ERROR, productConfig);
                     }
                 } else if (arg.startsWith(CommandLineConstants.PRIMARY_ADDRESS)) {
 
                     int idx = arg.indexOf('=');
                     if (idx == arg.length() - 1) {
                         STDERR.println(HostControllerLogger.ROOT_LOGGER.argumentExpected(arg, usageNote()));
-                        return new HostControllerEnvironmentWrapper(HostControllerEnvironmentWrapper.HostControllerEnvironmentStatus.ERROR);
+                        return new HostControllerEnvironmentWrapper(HostControllerEnvironmentWrapper.HostControllerEnvironmentStatus.ERROR, productConfig);
                     }
                     String value = idx > -1 ? arg.substring(idx + 1) : checkValueIsNotAnArg(arg, args[++i]);
                     if (value == null) {
-                        return new HostControllerEnvironmentWrapper(HostControllerEnvironmentWrapper.HostControllerEnvironmentStatus.ERROR);
+                        return new HostControllerEnvironmentWrapper(HostControllerEnvironmentWrapper.HostControllerEnvironmentStatus.ERROR, productConfig);
                     }
                     value = fixPossibleIPv6URL(value);
                     hostSystemProperties.put(HostControllerEnvironment.JBOSS_DOMAIN_PRIMARY_ADDRESS, value);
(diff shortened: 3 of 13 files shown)


controller/src/main/java/org/jboss/as/controller/Stream.java
```diff
@@ -0,0 +1,44 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.jboss.as.controller;
+
+/**
+ * @author Paul Ferraro
+ */
+public enum Stream {
```
jamezp, line 22: This is just kind of a general though. Should we use another name besides stream because of `java.util.stream.Stream`?

controller/src/main/java/org/jboss/as/controller/SimpleOperationDefinitionBuilder.java
```diff
@@ -201,9 +202,13 @@ public SimpleOperationDefinitionBuilder addAccessConstraint(final AccessConstrai
...
     }
 
-    SimpleOperationDefinitionBuilder setDescriptionProvider(DescriptionProvider provider){
+    SimpleOperationDefinitionBuilder setDescriptionProvider(DescriptionProvider provider) {
         this.descriptionProvider = provider;
         return this;
     }
 
+    SimpleOperationDefinitionBuilder setStream(Stream stream) {
+        this.stream = stream;
```
jamezp, line 211: Should this be something like:
```java
this.stream = stream == null ? Stream.DEFAULT : stream;
```

jamezp (commented):
Just a couple minor comments. Overall though I think this looks good.

controller/src/main/java/org/jboss/as/controller/SimpleOperationDefinitionBuilder.java
pferraro: Oops - that code is obsolete.  Originally, I intended to associate Resource/Attribute/OperationDefinition instances with a stream, and conditionally register based on the stream used by the server - but that proved to be super awkward for ResourceDefinition registration, which expects a child MRR in return.

controller/src/main/java/org/jboss/as/controller/Stream.java
pferraro: Honestly, I could think of a better name. Any ideas?

yersan: ModelStream?

rhusar: ModelStream sounds good to me.

controller/src/main/java/org/jboss/as/controller/xml/VersionedURN.java
```diff
@@ -51,12 +62,30 @@ public VersionedURN(String identifier, IntVersion version) {
...
+        this(identifiers, version, Stream.DEFAULT);
+    }
+
+    /**
+     * Constructs a versioned URN-based namespace using the specified URN components and version.
+     * @param components a list of URN components
+     * @param version a version
+     */
+    public VersionedURN(List<String> identifiers, IntVersion version, Stream stream) {
+        super(String.join(URN_DELIMITER, new CompositeIterable<>(List.of(URN), identifiers, (stream == Stream.DEFAULT) ? List.of(version.toString(2)) : List.of(stream.toString(), version.toString(2)))), version);
```
cam-rod, line 89: I feel like it should also the stable stream should also be a valid URN (ex. `urn:wildfly:elytron:17.0:stable`). Default representation still drops that component, but it should be valid.

server/src/main/java/org/jboss/as/server/ServerEnvironment.java
```diff
@@ -385,13 +395,13 @@ public ServerEnvironment(final String hostControllerName, final Properties props
             final ConfigurationFile.InteractionPolicy configInteractionPolicy, final LaunchType launchType,
             final RunningMode initialRunningMode, ProductConfig productConfig, long startTime, boolean startSuspended,
             String gitRepository, String gitBranch, String gitAuthConfiguration, String supplementalConfiguration) {
-        this(hostControllerName, props, env, serverConfig, configInteractionPolicy, launchType, initialRunningMode,
+        this(hostControllerName, props, env, serverConfig, configInteractionPolicy, launchType, initialRunningMode, Stream.STABLE,
```
cam-rod, line 398: Is there a reason to define this as STABLE instead of just DEFAULT?

server/src/main/java/org/jboss/as/server/Main.java
```diff
@@ -392,6 +395,15 @@ public static ServerEnvironmentWrapper determineEnvironment(String[] args, Prope
...
                     }
+                } else if (arg.startsWith(CommandLineConstants.STREAM)) {
+                    String streamValue = parseValue(arg, CommandLineConstants.STREAM);
+                    try {
+                        stream = Stream.valueOf(streamValue.toUpperCase(Locale.ROOT));
+                    } catch (IllegalArgumentException e) {
+                        STDERR.println(ServerLogger.ROOT_LOGGER.failedToParseCommandLineEnum(arg, streamValue, EnumSet.allOf(Stream.class)));
+                        usage();
+                        return new ServerEnvironmentWrapper(ServerEnvironmentWrapper.ServerEnvironmentStatus.ERROR);
+                    }
```
yersan, line 406: Practically the same changes should be done in org.jboss.as.host.controller.Main to initialize a HostControllerEnvironment with the stream passed as an argument.

We also have to verify if the ProcessController is able to pass these arguments to the command that launches the HostController. Is in class:

https://github.com/wildfly/wildfly-core/blob/08cb24dacd9615bb530b53067567c707a6184e16/process-controller/src/main/java/org/jboss/as/process/Main.java#L223-L247C2

The idea is when the process controller is launched with --stream, this argument value arrives at org.jboss.as.host.controller.Main to initialize the HostControllerEnvironment with it, and then, take it to initialize the Managed Servers launch command.

server/src/main/java/org/jboss/as/server/ServerEnvironment.java
pferraro: Yeah - this should use DEFAULT.

controller/src/main/java/org/jboss/as/controller/xml/VersionedURN.java
pferraro: So, a given subsystem (or deployment descriptor) would need to register 2 namespaces with the same parser for the current stable version?  That seems overly cumbersome to me.

server/src/main/java/org/jboss/as/server/ServerEnvironment.java
pferraro: Nice catch.  Fixed.

server/src/main/java/org/jboss/as/server/ServerStartTask.java
```diff
@@ -133,7 +134,7 @@ public AsyncFuture<ServiceContainer> run(final List<ServiceActivator> runService
         // Create server environment on the server, so that the system properties are getting initialized on the right side
         final ServerEnvironment providedEnvironment = new ServerEnvironment(hostControllerName, properties,
                 WildFlySecurityManager.getSystemEnvironmentPrivileged(), null, null, ServerEnvironment.LaunchType.DOMAIN,
-                RunningMode.NORMAL, productConfig, Module.getStartTime(), suspend, gracefulStartup, null, null, null, null);
+                RunningMode.NORMAL, Stream.DEFAULT, productConfig, Module.getStartTime(), suspend, gracefulStartup, null, null, null, null);
```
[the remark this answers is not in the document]
pferraro: OK - that makes sense to me now.  I'll add this today.

controller/src/main/java/org/jboss/as/controller/xml/VersionedURN.java
cam-rod: Yeah, I'm not sure how I feel about that one, just seemed a bit inconsistent

server/src/main/java/org/jboss/as/server/Main.java
pferraro: Sounds good.  I'll add this today.

controller/src/main/java/org/jboss/as/controller/xml/VersionedURN.java
pferraro: Keep in mind that most subsystems will not have any preview or experimental features (at least initially), and thus will not register any preview/experimental namespaces.
It seems odd to me that we would start registering duplicate namespaces for the stable stream only when they eventually incorporate a preview or experimental feature.

cam-rod: Although I don't see a reason for a separate "default" URN to exist, that's just what the unspecified one means.

pferraro: Ultimately, every valid URN needs to be registered with a subsystem parser.  It is unclear to me when subsystem would need to register both a urn:wildfly:foo:X.Y and urn:wildfly:foo:stable:X.Y namespace.  Currently, subsystems only register namespaces using the former pattern.  None use the latter.  Do you propose we create these retroactively for every existing namespace for every subsystem?  What about external subsystems? e.g. those in wildfly-extras?

cam-rod: Good point, leave it as is then :)

server/src/main/java/org/jboss/as/server/ServerStartTask.java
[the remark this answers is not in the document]
pferraro: OK - I've updated this.  Let me know if it looks ok now.

server/src/main/java/org/jboss/as/server/Main.java
pferraro: OK - I've updated this.  Let me know if it looks ok now.

controller/src/main/java/org/jboss/as/controller/Stream.java
pferraro: I've updated this to FeatureStream.  If anyone has a better idea, let me know.

yersan (commented):
Hi @pferraro, I did another review. Regarding domain mode, we also need to reject registrations of host controllers that are not running at the same FeatureStream level as the domain controller.

I think Zulip could be better to discuss this, so I'll add the comments there

process-controller/src/main/java/org/jboss/as/process/CommandLineArgumentUsageImpl.java
```diff
@@ -95,6 +95,9 @@ public static void init(){
 
         addArguments(CommandLineConstants.SECMGR);
         instructions.add(ProcessLogger.ROOT_LOGGER.argSecMgr());
+
+        addArguments(CommandLineConstants.FEATURE_STREAM + "=<value>");
+        instructions.add(ProcessLogger.ROOT_LOGGER.argFeatureStream());
```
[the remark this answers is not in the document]
pferraro: Will do.

process-controller/src/main/java/org/jboss/as/process/logging/ProcessLogger.java
```diff
@@ -460,6 +460,14 @@ public interface ProcessLogger extends BasicLogger {
     @Message(id = Message.NONE, value = "Runs the server with a security manager installed.")
     String argSecMgr();
 
+    /**
+     * Instructions for the {@link CommandLineConstants#FEATURE_STREAM} command line argument.
+     *
+     * @return the message
+     */
+    @Message(id = Message.NONE, value = "Runs the server using a specific feature stream.")
+    String argFeatureStream();
```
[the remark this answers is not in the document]
pferraro: Makes sense.

host-controller/src/main/java/org/jboss/as/host/controller/HostControllerEnvironment.java
```diff
@@ -496,6 +504,10 @@ public HostControllerEnvironment(Map<String, String> hostSystemProperties, boole
         // Note the java.security.manager property shouldn't be set, but we'll check to ensure the security manager should be enabled
         this.securityManagerEnabled = securityManagerEnabled || isJavaSecurityManagerConfigured(hostSystemProperties);
         this.processType = processType;
+
+        String streamName = hostSystemProperties.get(HOST_STREAM);
+        this.stream = (streamName != null) ? FeatureStream.valueOf(streamName.toUpperCase(Locale.ROOT)) : FeatureStream.DEFAULT;
```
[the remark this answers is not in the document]
pferraro: Will do

server/src/main/java/org/jboss/as/server/ServerStartTask.java
```diff
@@ -124,6 +124,7 @@ public ServerStartTask(final String hostControllerName, final String serverName,
         setPropertyIfFound(launchProperties, ServerEnvironment.SERVER_CONTENT_DIR, properties);
         setPropertyIfFound(launchProperties, ServerEnvironment.SERVER_LOG_DIR, properties);
         setPropertyIfFound(launchProperties, ServerEnvironment.SERVER_TEMP_DIR, properties);
+        setPropertyIfFound(launchProperties, ServerEnvironment.SERVER_STREAM, properties);
```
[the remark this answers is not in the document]
pferraro: Understood.  In that case, it probably makes sense to use the same property name for both - as they would never differ.

server/src/main/java/org/jboss/as/server/ServerEnvironment.java
```diff
@@ -306,7 +311,7 @@ public ProcessType getProcessType() {
      * Properties that can only be set via {@link #systemPropertyUpdated(String, String)} during server boot.
      */
     private static final Set<String> BOOT_PROPERTIES = new HashSet<String>(Arrays.asList(BUNDLES_DIR, SERVER_TEMP_DIR,
-            NODE_NAME, SERVER_NAME, HOST_NAME, QUALIFIED_HOST_NAME));
+            NODE_NAME, SERVER_NAME, HOST_NAME, QUALIFIED_HOST_NAME, SERVER_STREAM));
```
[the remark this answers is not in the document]
pferraro: OK - will fix.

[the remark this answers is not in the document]
pferraro: OK - will fix.

controller/src/main/java/org/jboss/as/controller/extension/ExtensionRegistry.java
```diff
@@ -137,21 +139,38 @@ public final class ExtensionRegistry {
...
      *
      * @param processType the type of the process
+     * @param stream the feature stream of the core application server
      * @param runningModeControl the process' running mode
      * @param auditLogger logger for auditing changes
      * @param authorizer hook for exposing access control information to the JMX subsystem
      * @param hostControllerInfoAccessor the host controller
      */
-    public ExtensionRegistry(ProcessType processType, RunningModeControl runningModeControl, ManagedAuditLogger auditLogger, JmxAuthorizer authorizer,
+    public ExtensionRegistry(ProcessType processType, FeatureStream stream, RunningModeControl runningModeControl, ManagedAuditLogger auditLogger, JmxAuthorizer authorizer,
```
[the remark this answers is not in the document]
pferraro: ExtensionRegistry, and similar classes with numerous constructor parameters (e.g. AbstractControllerService), many of them optional, would be better served either by encapsulating the parameters within an interface, e.g. ExtensionRegistryConfiguration, with appropriate default method implementations; or by using a builder pattern.  This allows the class to evolve its constructor parameters with minimal impact to callers.  This avoids constructor bloat, as well as the need to continually deprecate public constructors, as new overloaded constructors with additional optional parameters are added.

pferraro:
Rebased this branch on top of https://github.com/wildfly/wildfly-core/pull/5563 which makes it possible to perform AttributeDefinition registration filtering without impacting add resource and write-attribute OperationStepHandler implementations.  As such, the full integration tests will fail until https://github.com/wildfly/wildfly/pull/17007 is merged.

pferraro:
@yersan FYI, the last commit addresses your final comment above.  Can you have a look?

host-controller/src/main/java/org/jboss/as/host/controller/mgmt/HostControllerRegistrationHandler.java
```diff
@@ -443,12 +443,16 @@ public void operationPrepared(final ModelController.OperationTransaction transac
...
+                if (this.extensionRegistry.enables(this.hostInfo)) {
+                    try {
+                        registerHost(transaction, result);
+                    } catch (SlaveRegistrationException e) {
+                        failed(e, e.getErrorCode(), e.getErrorMessage());
+                    } catch (Exception e) {
+                        failed(e, SlaveRegistrationException.ErrorCode.UNKNOWN, e.getClass().getName() + ":" + e.getMessage());
+                    }
+                } else {
+                    this.failed(SlaveRegistrationException.ErrorCode.INCOMPATIBLE_VERSION, DomainControllerLogger.ROOT_LOGGER.hostFeatureStreamNotEnabled(this.extensionRegistry.getFeatureStream(), this.hostInfo.getFeatureStream()));
```
yersan, line 455: Hi @pferraro , the rejection should be done a bit earlier, more specifically at https://github.com/wildfly/wildfly-core/blob/main/host-controller/src/main/java/org/jboss/as/host/controller/mgmt/HostControllerRegistrationHandler.java#L336-L348

You can get a reference of the ExtensioRegistry using the HostControllerRegistrationHandler constructor.
That's also the place where we reject is a host is incompatible due to its kernel version and we know we can reject it there.

yersan (changes requested):

host-controller/src/main/java/org/jboss/as/host/controller/mgmt/HostControllerRegistrationHandler.java
pferraro: @yersan Makes sense.  Will fix.

pferraro: @yersan I've updated that commit.  Let me know if it looks OK now.

host-controller/src/main/java/org/jboss/as/host/controller/mgmt/HostControllerRegistrationHandler.java
```diff
@@ -346,6 +347,13 @@ public void execute(final OperationContext context, final ModelNode operation) t
                     registrationContext.failed(failure, SlaveRegistrationException.ErrorCode.INCOMPATIBLE_VERSION, failure.getMessage());
                     throw failure;
                 }
+                // Ensure feature stream of host is compatible
+                LocalHostControllerInfo domainInfo = domainController.getLocalHostInfo();
+                if (!domainInfo.enables(hostInfo)) {
```
yersan, line 352: The agreement was to allow only paired domains, I mean, STABLE <->, PREVIEW <-> PREVIEW, EXPERIMENTAL <-> EXPERIMENTAL, we don't want to deal with host exclusions with host using mixed streams. However, looking at the FeatureStream.enables(FeatureStream stream), here we are not rejecting paired streams.

I guess we need to be more restrictive here.

yersan (dismissed):

host-controller/src/main/java/org/jboss/as/host/controller/mgmt/HostControllerRegistrationHandler.java
pferraro: @yersan Forgive my ignorance in this area, but I'm struggling a bit to recall the reasons why host exclusions were specifically a problem WRT feature streams...

AIUI, host exclusions are meant to handle the case where a subsystem's module referenced within the configuration of the domain controller does not exist on a host controller because it is running an earlier version prior to the module's existence.  Does that sound right?
Do host exclusions serve any other purpose?

If this is the case, I do not see why a host supporting PREVIEW features should not be allowed to connect to a domain controller supporting only STABLE features, since the configuration sent from the domain controller will always be valid on the host (and also why I said the logic above is backwards).  So, host1 using the STABLE feature stream, host2 running the PREVIEW feature stream, and host3 running the EXPERIMENTAL feature stream _should_ all be able to safely join a domain controller so long as it is running the STABLE feature stream - since the feature stream does not affect the modules available to a host (unless we also decided to also allow modules to be associated with a feature stream - but we'll ignore that for now).  Am I missing something?

The complexity, I think, comes when the hosts are of mixed versions - since we now require model transformation to the management API version supported by the host, and we don't want to introduce the complexity/maintenance of supporting model transformations for non-STABLE feature streams.  Therefore, it seems to me that our host registration check would need to be more restrictive than your comment above; i.e. we can allow a domain of hosts running different versions, so long as all hosts use the STABLE stream; or a domain of hosts running the same versions, so long as all hosts are using the same feature stream.  Does that make sense?

yersan: > Forgive my ignorance in this area, but I'm struggling to recall the reasons why host exclusions were specifically a problem WRT feature streams...

Of course, there is nothing to forgive @pferraro :)

> AIUI, host exclusions are meant to handle the case where a subsystem's module referenced within the configuration of the domain controller does not exist on a host controller because it is running an earlier version prior to the module's existence. Does that sound right?

Yes, that's right. We can talk about extensions, host exclusions exclude extensions.

> Do host exclusions serve any other purpose?

No, just to exclude extensions (with their subsystem) that are unknown by the secondary host. They need to be ignored on the secondary host. When the secondary host gets a copy of the management model from the domain controller, it needs to know which extensions must be ignored based on its kernel version. Sadly we control the list of those extensions on the Domain Controller side, a smarter approach could be host controller to advise what are the extensions known (that's a different topic we don't need to talk here).

> If this is the case, I do not see why a host supporting PREVIEW features should not be allowed to connect to a domain controller supporting only STABLE features, since the configuration sent from the domain controller will always be valid on the host (and also why I said the logic above is backwards). So, host1 using the STABLE feature stream, host2 running the PREVIEW feature stream, and host3 running the EXPERIMENTAL feature stream should all be able to safely join a domain controller so long as it is running the STABLE feature stream - since the feature stream does not affect the modules available to a host (unless we also decided to also allow modules to be associated with a feature stream - but we'll ignore that for now). Am I missing something?

You're right about the above, let's forget mixed domains here. Mixed domain use case is addressed in your third paragraph. I see your reasoning and agree with it; any host controller will support a superset of the extensions sent by the domain controller, so the secondary hosts always be able to understand any possible extension sent by the Domain Controller. We even don't need to worry about host-exlusions for extensions added only on EXPERIMENTAL/PREVIEW. Got it. 

(**) Theoretically, we could have secondary hosts running on a feature stream level above the Domain Controller feature stream.

> The complexity, I think, comes when the hosts are of mixed versions - since we now require model transformation to the management API version supported by the host, and we don't want to introduce the complexity/maintenance of supporting model transformations for non-STABLE feature streams. Therefore, it seems to me that our host registration check would need to be more restrictive than your comment above; i.e. we can allow a domain of hosts running different versions, so long as all hosts use the STABLE stream; or a domain of hosts running the same versions, so long as all hosts are using the same feature stream. Does that make sense?

Right, that makes sense, and indeed Transformers was what I had in mind but I only talked about host exclusions. Yes, this is what we have discussed in the past, they should be running at the same level, at least from the point of view of different HC kernel versions. We will always have (DC <-> HC) in {STABLE <-> STABLE, PREVIEW <-> PREVIEW, EXPERIMENTAL <-> EXPERIMENTAL}, otherwise we reject the registration.

We can go further and do it only if the kernel versions are different, falling back to the discussion (**) but there is no transformers involved, but, there is no gain with that. I would vote for the most restrictive one: paired feature stream versions.

Let me know if I'm missing something.

pferraro: @yersan Thanks for the explanations.  I think I get it now.
I'll update the last commit to reject legacy hosts if the domain controller is not using the STABLE stream.

pferraro:
FYI - full integration test failures should be resolved by https://github.com/wildfly/wildfly/pull/17028

yersan:
@pferraro FYI, I've noticed that any extension/subsystem that wants to use a feature stream explicitly should get a dependency on `org.jboss.as.version` jBoss Modules module, I'm not sure if it is something we wanted to add by default somehow, or if it should be added on demand.

pferraro:
@yersan Good point.  Initially, the FeatureStream interface resided within org.wildfly.core:wildfly-controller.  However, it also needs to be referenced by org.wildfly.core:wildfly-process-controller, which only has a single internal dependency, i.e. org.wildfly.core:wildfly-version; so this seemed like the obvious option.
Frankly, I don't see why a subsystem having to add a dependency on org.jboss.as.version to include an experimental feature is any more cumbersome than a subsystem needing to add a dependency on org.jboss.as.server to add a DUP to the deployment chain.
IMO, this is a general encapsulation issue with WildFly that has always affected subsystems.  The fact that subsystems need to depend on org.jboss.as.controller (and especially org.jboss.as.server) already expose subsystems to lots of unnecessary server internals.  Ideally, a subsystem module should only need to depend on a single subsystem SPI module.  I do think this is worth addressing at some point - but I think that can be done independently of this proposal.

host-controller/src/main/java/org/jboss/as/host/controller/mgmt/HostInfo.java
```diff
@@ -130,6 +135,7 @@ public static HostInfo fromModelNode(final ModelNode hostInfo, DomainHostExclude
     private final Set<ServerConfigInfo> serverConfigInfos;
     private final Set<String> domainIgnoredExtensions;
     private final boolean hostDeclaredIgnoreUnaffected;
+    private final FeatureStream stream;
```
yersan, line 138: The HostInfo must override the `getFeatureStream()` method and return this steam value instead of the one supplied by default by its parent `org.jboss.as.controller.Feature` class.

server/src/main/java/org/jboss/as/server/Main.java
yersan: Thank you, @pferraro LGTM

host-controller/src/main/java/org/jboss/as/host/controller/mgmt/HostControllerRegistrationHandler.java
yersan: Thank you, @pferraro LGTM

server/src/main/java/org/jboss/as/server/ServerStartTask.java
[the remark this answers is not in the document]
yersan: Thank you, @pferraro LGTM

host-controller/src/main/java/org/jboss/as/host/controller/mgmt/HostControllerRegistrationHandler.java
yersan: Just for the record, this last step rejecting when not using STABLE for legacy hosts has not been done yet

host-controller/src/main/java/org/jboss/as/host/controller/HostControllerEnvironment.java
```diff
@@ -44,7 +46,9 @@
 import org.jboss.as.host.controller.jvm.JvmType;
 import org.jboss.as.host.controller.operations.LocalHostControllerInfoImpl;
 import org.jboss.as.network.NetworkUtils;
+import org.jboss.as.server.ServerEnvironment;
```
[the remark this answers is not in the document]
pferraro: > I think we could try to avoid incorporating server constants into the Host Controller environment.
> 
> If we are going to manage the stream property as a common property for Host Controllers, managed servers and standalone servers, we could move it up to the parent class `org.jboss.as.controller.operations.common.ProcessEnvironment`. That would remove the need to import `ServerEnvironment` into the `HostControllerEnvironment`

Excellent suggestion.  Makes sense to me.

host-controller/src/main/java/org/jboss/as/host/controller/mgmt/HostInfo.java
pferraro: Indeed.  Good catch.

host-controller/src/main/java/org/jboss/as/host/controller/mgmt/HostControllerRegistrationHandler.java
pferraro: Thanks for the reminder... :)  Will do this today.

host-controller/src/main/java/org/jboss/as/host/controller/HostControllerEnvironment.java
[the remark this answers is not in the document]
pferraro: Fixed.

host-controller/src/main/java/org/jboss/as/host/controller/mgmt/HostInfo.java
pferraro: Fixed.

yersan: Thanks!

host-controller/src/main/java/org/jboss/as/host/controller/HostControllerEnvironment.java
```diff
@@ -496,13 +500,29 @@ public HostControllerEnvironment(Map<String, String> hostSystemProperties, boole
         // Note the java.security.manager property shouldn't be set, but we'll check to ensure the security manager should be enabled
         this.securityManagerEnabled = securityManagerEnabled || isJavaSecurityManagerConfigured(hostSystemProperties);
         this.processType = processType;
+
+        this.stream = getEnumProperty(hostSystemProperties, FEATURE_STREAM, FeatureStream.DEFAULT);
+        WildFlySecurityManager.setPropertyPrivileged(FEATURE_STREAM, this.stream.toString());
```
yersan, line 505: Is not this somehow redundant? `hostSystemProperties` should already contain the properties for this process. Maybe you want it explicitly when it got the default value?

host-controller/src/main/java/org/jboss/as/host/controller/HostControllerEnvironment.java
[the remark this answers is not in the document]
yersan: thanks!

host-controller/src/main/java/org/jboss/as/host/controller/mgmt/HostControllerRegistrationHandler.java
```diff
@@ -337,6 +339,7 @@ public void execute(final OperationContext context, final ModelNode operation) t
                 final int major = hostInfo.getManagementMajorVersion();
                 final int minor = hostInfo.getManagementMinorVersion();
                 final int micro = hostInfo.getManagementMicroVersion();
+                ModelVersion hostVersion = ModelVersion.create(major, minor, micro);
```
yersan, line 342: Optional: It could be final ... just to follow the same pattern used in this code and to reflect the intention. I leave it up to your criteria.

host-controller/src/main/java/org/jboss/as/host/controller/mgmt/HostControllerRegistrationHandler.java
```diff
@@ -346,9 +349,22 @@ public void execute(final OperationContext context, final ModelNode operation) t
...
+                    OperationFailedException failure = HostControllerLogger.ROOT_LOGGER.incompatibleFeatureStream(domainInfo.getFeatureStream(), hostInfo.getFeatureStream());
+                    registrationContext.failed(failure, SlaveRegistrationException.ErrorCode.INCOMPATIBLE_VERSION, failure.getMessage());
+                    throw failure;
+                }
+                // Legacy hosts must use DEFAULT stream
+                if ((hostVersion.compareTo(ModelVersion.CURRENT) < 0) && (hostInfo.getFeatureStream() != FeatureStream.DEFAULT)) {
+                    OperationFailedException failure = HostControllerLogger.ROOT_LOGGER.incompatibleFeatureStreamForLegacyHost(hostInfo.getFeatureStream());
+                    registrationContext.failed(failure, SlaveRegistrationException.ErrorCode.INCOMPATIBLE_VERSION, failure.getMessage());
+                    throw failure;
+                }
```
yersan, line 364: Optional: To save one comparison, you could place the if that checks the versions before and add an `else` to verify the other case. It could make both checks more independent, but this is a very minor detail and not strictly neccessary at all. I leave it under your criteria, you can resolve this conversation yourself.

host-controller/src/main/java/org/jboss/as/host/controller/mgmt/HostControllerRegistrationHandler.java
yersan: Looks good, thanks @pferraro , I've just added two minor optional comments related to this but I left them up to your criteria, please resolve this conversation from your side at your convenience.

host-controller/src/main/java/org/jboss/as/host/controller/mgmt/HostControllerRegistrationHandler.java
pferraro: This article best summarizes my opinion on the use of final for local variables: https://brytecode.be/articles/use-of-the-final-keyword-in-java/#final-for-local-variables

host-controller/src/main/java/org/jboss/as/host/controller/mgmt/HostControllerRegistrationHandler.java
pferraro: Hmm - I'm confused.  Which comparison is the redundant one?

host-controller/src/main/java/org/jboss/as/host/controller/HostControllerEnvironment.java
pferraro: Yes - you are right.  I'll fix this to call WildFlySecurityManager.setPropertyPrivileged(...) conditionally.

host-controller/src/main/java/org/jboss/as/host/controller/mgmt/HostControllerRegistrationHandler.java
yersan: This can be resolved now.

host-controller/src/main/java/org/jboss/as/host/controller/HostControllerEnvironment.java
yersan: Looks good, but I was expecting something like this:

```
if(hostSystemProperties.get(FEATURE_STREAM) == null) {
    WildFlySecurityManager.setPropertyPrivileged(...) 
}
```

So we set it with the default value only if it was not set before.

controller/src/main/java/org/jboss/as/controller/registry/ManagementResourceRegistration.java
```diff
@@ -110,6 +111,23 @@ public interface ManagementResourceRegistration extends ImmutableManagementResou
...
     ManagementResourceRegistration registerOverrideModel(final String name, final OverrideDescriptionProvider descriptionProvider);
 
+    /**
+     * Register a specifically named resource that overrides this {@link PathElement#WILDCARD_VALUE wildcard registration}
+     * by adding additional attributes, operations or child types.
+     *
+     * @param registration the child registration of this registry that should no longer be available
+     * @param descriptionProvider provider for descriptions of the additional attributes or child types
+     *
+     * @return a resource registration which may be used to add attributes, operations and sub-models
```
bstansberry, line 121: This should note that it may return null.

server/src/main/java/org/jboss/as/server/Main.java
```diff
@@ -366,6 +367,12 @@ public static ServerEnvironmentWrapper determineEnvironment(String[] args, Prope
                     } else {
                         gitBranch = arg.substring(idx + 1);
                     }
+                } else if (arg.startsWith(CommandLineConstants.STABILITY)) {
```
bstansberry, line 370: For now, if ProductConfig.getDefaultStability() == DEFAULT this should fail. A similar check can be done in CommandLineArgumentUsageImpl to guard allowing the argument.

host-controller/src/main/java/org/jboss/as/host/controller/Main.java
```diff
@@ -445,6 +446,12 @@ public static HostControllerEnvironmentWrapper determineEnvironment(String[] arg
                 } else if (arg.equals(CommandLineConstants.SECMGR)) {
                     // Enable the security manager
                     securityManagerEnabled = true;
+                } else if (arg.startsWith(CommandLineConstants.STABILITY)) {
```
bstansberry, line 449: See my comments in server module Main.

bstansberry:
@pferraro I've made some minor comments but in general looks good.

logging/pom.xml
```diff
@@ -123,26 +123,9 @@
...
         <dependency>
             <groupId>org.wildfly.core</groupId>
             <artifactId>wildfly-server</artifactId>
-            <exclusions>
-                <exclusion>
-                    <groupId>*</groupId>
-                    <artifactId>*</artifactId>
-                </exclusion>
-            </exclusions>
         </dependency>
```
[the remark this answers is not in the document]
pferraro: See: https://github.com/wildfly/wildfly-core/pull/5413/commits/e081907b4ddd21bab25e3d7c0a84937a2ea6f005

host-controller/src/main/java/org/jboss/as/host/controller/mgmt/HostInfo.java
```diff
@@ -127,6 +133,7 @@ private HostInfo(final ModelNode hostInfo, DomainHostExcludeRegistry hostIgnoreR
         productVersion = hostInfo.hasDefined(PRODUCT_VERSION) ? hostInfo.require(PRODUCT_VERSION).asString() : null;
         remoteConnectionId = hostInfo.hasDefined(RemoteDomainConnectionService.DOMAIN_CONNECTION_ID)
                 ? hostInfo.get(RemoteDomainConnectionService.DOMAIN_CONNECTION_ID).asLong() : null;
+        this.stability = Optional.ofNullable(hostInfo.get(ModelDescriptionConstants.STABILITY).asStringOrNull()).map(Stability::valueOf).orElse(productConfig.getDefaultStability());
```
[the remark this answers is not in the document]
pferraro: Added clarifying comment.

server/src/main/java/org/jboss/as/server/Main.java
pferraro: Fixed.  Also removed from command line usage, when stability has only 1 possible value.

host-controller/src/main/java/org/jboss/as/host/controller/Main.java
pferraro: Fixed.  Also removed from command line usage, when stability has only 1 possible value.

host-controller/src/main/java/org/jboss/as/host/controller/Main.java
```diff
@@ -428,33 +429,38 @@ public static HostControllerEnvironmentWrapper determineEnvironment(String[] arg
...
-                    return new HostControllerEnvironmentWrapper(HostControllerEnvironmentWrapper.HostControllerEnvironmentStatus.ERROR);
+                    return new HostControllerEnvironmentWrapper(HostControllerEnvironmentWrapper.HostControllerEnvironmentStatus.ERROR, productConfig);
                 }
             } catch (IndexOutOfBoundsException e) {
                 STDERR.println(HostControllerLogger.ROOT_LOGGER.argumentExpected(arg, usageNote()));
-                return new HostControllerEnvironmentWrapper(HostControllerEnvironmentWrapper.HostControllerEnvironmentStatus.ERROR);
+                return new HostControllerEnvironmentWrapper(HostControllerEnvironmentWrapper.HostControllerEnvironmentStatus.ERROR, productConfig);
             }
         }
-        productConfig = ProductConfig.fromFilesystemSlot(Module.getBootModuleLoader(), WildFlySecurityManager.getPropertyPrivileged(HostControllerEnvironment.HOME_DIR, null), hostSystemProperties);
```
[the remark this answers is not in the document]
pferraro: Fixed

host-controller/src/main/java/org/jboss/as/host/controller/Main.java
```diff
@@ -195,7 +196,7 @@ public static HostControllerEnvironmentWrapper determineEnvironment(String[] arg
         String initialHostConfig = null;
         RunningMode initialRunningMode = RunningMode.NORMAL;
         Map<String, String> hostSystemProperties = getHostSystemProperties();
-        ProductConfig productConfig;
+        ProductConfig productConfig = ProductConfig.fromFilesystemSlot(Module.getBootModuleLoader(), WildFlySecurityManager.getPropertyPrivileged(HostControllerEnvironment.HOME_DIR, null), hostSystemProperties);
```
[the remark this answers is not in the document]
pferraro: Fixed

server/src/main/java/org/jboss/as/server/Main.java
```diff
@@ -376,26 +382,25 @@ public static ServerEnvironmentWrapper determineEnvironment(String[] args, Prope
...
             } catch (IndexOutOfBoundsException e) {
                 STDERR.println(ServerLogger.ROOT_LOGGER.valueExpectedForCommandLineOption(arg));
-                usage();
+                usage(productConfig);
                 return new ServerEnvironmentWrapper(ServerEnvironmentWrapper.ServerEnvironmentStatus.ERROR);
             }
         }
 
         String hostControllerName = null; // No host controller unless in domain mode.
-        productConfig = ProductConfig.fromFilesystemSlot(Module.getBootModuleLoader(), WildFlySecurityManager.getPropertyPrivileged(ServerEnvironment.HOME_DIR, null), systemProperties);
```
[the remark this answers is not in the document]
pferraro: Fixed

server/src/main/java/org/jboss/as/server/Main.java
```diff
@@ -124,7 +125,7 @@ public static ServerEnvironmentWrapper determineEnvironment(String[] args, Prope
         String gitAuthConfiguration = null;
         String supplementalConfiguration = null;
         RunningMode runningMode = RunningMode.NORMAL;
-        ProductConfig productConfig;
+        ProductConfig productConfig = ProductConfig.fromFilesystemSlot(Module.getBootModuleLoader(), WildFlySecurityManager.getPropertyPrivileged(ServerEnvironment.HOME_DIR, null), systemProperties);
```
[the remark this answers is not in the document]
pferraro: Fixed

bstansberry (approved):

bstansberry:
Thanks @pferraro

PR #5415, opened 2023-03-22
[WFCORE-4916] Unclear attribute name completion for LIST type

ChristinaDsl:
https://issues.redhat.com/browse/WFCORE-4916

After types List, Object and String are added ",", if there are more attributes to come or ")" if this is the last one.

### DIFF
--- a/cli/src/main/java/org/jboss/as/cli/impl/AttributeNamePathCompleter.java
+++ b/cli/src/main/java/org/jboss/as/cli/impl/AttributeNamePathCompleter.java
@@ -332,6 +332,11 @@ public class AttributeNamePathCompleter implements CommandLineCompleter {
                             candidateIndex += chunk.length();
                         }
                         candidates.add("[");
+                    } else if (modelType.equals(ModelType.STRING) || modelType.equals(ModelType.INT) || modelType.equals(ModelType.BOOLEAN)){
+                        if (candidates.isEmpty()) {
+                            candidateIndex += chunk.length();
+                        }
+                        candidates.add("");
                     }
                 }
             }
--- a/testsuite/domain/src/test/java/org/jboss/as/test/integration/domain/management/cli/CliCompletionTestCase.java
+++ b/testsuite/domain/src/test/java/org/jboss/as/test/integration/domain/management/cli/CliCompletionTestCase.java
@@ -1140,10 +1140,12 @@ public class CliCompletionTestCase {
                 List<String> candidates = new ArrayList<>();
                 ctx.getDefaultCommandCompleter().complete(ctx, cmd,
                         cmd.length(), candidates);
-                assertTrue(candidates.toString(), candidates.size() == 1);
+                assertTrue(candidates.toString(), candidates.size() == 2);
+                assertTrue(candidates.toString(), candidates.contains(" "));
                 assertTrue(candidates.toString(), candidates.contains("namespaces"));
-                candidates = complete(ctx, cmd, false, cmd.length() - "name".length());
-                assertTrue(candidates.toString(), candidates.size() == 1);
+                candidates = complete(ctx, cmd, false, cmd.length());
+                assertTrue(candidates.toString(), candidates.size() == 2);
+                assertTrue(candidates.toString(), candidates.contains(" "));
                 assertTrue(candidates.toString(), candidates.contains("namespaces"));
             }
 
@@ -1155,6 +1157,17 @@ public class CliCompletionTestCase {
                 assertTrue(candidates.toString(), candidates.isEmpty());
                 candidates = complete(ctx, cmd, null, -1);
                 assertTrue(candidates.toString(), candidates.isEmpty());
+            }
+
+            {
+                String cmd = "read-attribute management-minor-version";
+                List<String> candidates = new ArrayList<>();
+                ctx.getDefaultCommandCompleter().complete(ctx, cmd,
+                        cmd.length(), candidates);
+                assertTrue(candidates.toString(), candidates.size() == 1);
+                assertTrue(candidates.toString(), candidates.contains(" "));
+                candidates = complete(ctx, cmd, false, cmd.length());
+                assertTrue(candidates.toString(), candidates.contains(" "));
             }
 
             {
--- a/cli/src/main/java/org/jboss/as/cli/operation/OperationRequestCompleter.java
+++ b/cli/src/main/java/org/jboss/as/cli/operation/OperationRequestCompleter.java
@@ -23,7 +23,6 @@ import org.jboss.as.cli.parsing.StateParser;
 import org.jboss.as.cli.parsing.StateParser.SubstitutedLine;
 import org.jboss.logging.Logger;
 
-
 /**
  *
  * @author Alexey Loubyansky
@@ -401,13 +400,13 @@ public class OperationRequestCompleter implements CommandLineCompleter {
             // Implies a single candidate to inline, the value is complete.
             // propose the property separator if more properties to come
             // or the propertyListEnd if no more properties.
-            if (suggestionEqualsUserEntry(candidates, chunk, valueResult)) {
+            if (suggestionEqualsUserEntry(candidates, chunk, valueResult)|| areIncludedCandidatesForSpecificValueTypes(candidates)) {
                 final CommandLineFormat format = parsedCmd.getFormat();
                 if (format != null) {
                     for (CommandArgument arg : allArgs) {
                         try {
                             if (arg.canAppearNext(ctx)) {
-                                candidates.set(0, "" + format.getPropertySeparator());
+                                candidates.add("" + format.getPropertySeparator());
                                 return buffer.length();
                             }
                         } catch (CommandFormatException e) {
@@ -415,8 +414,10 @@ public class OperationRequestCompleter implements CommandLineCompleter {
                         }
                     }
                     // inline the end of properties.
-                    candidates.set(0, format.getPropertyListEnd());
                     // at the end of the input.
+                    if((buffer.charAt(buffer.length() - 1))!='='){
+                        candidates.add(format.getPropertyListEnd());
+                    }
                     return buffer.length();
                 }
             }
@@ -717,12 +718,26 @@ public class OperationRequestCompleter implements CommandLineCompleter {
             return false;
         }
 
-        if (suggestionOffset > 0) {
+        if (suggestionOffset > 0 && candidates.get(0)!="") {
             // user entry before suggestionOffset is always the same - compare only part after offset
             return chunk.substring(suggestionOffset).equals(candidates.get(0));
         } else {
-            return chunk.equals(candidates.get(0));
+            if(chunk.equals(candidates.get(0))){
+                candidates.clear();
+                return true;
+            }
+            return false;
         }
+    }
+
+    boolean areIncludedCandidatesForSpecificValueTypes(List<String> candidates){
+        if(candidates.contains("[") || candidates.contains(".")){
+            return true;
+        }else if(candidates.contains("")){
+            candidates.remove("");
+            return true;
+        }
+        return false;
     }
 
     protected CommandLineCompleter getValueCompleter(CommandContext ctx, Iterable<CommandArgument> allArgs, final String argName) {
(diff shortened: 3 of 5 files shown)


yersan (changes requested):
Hi @ChristinaDsl , following the Jira description, the idea here was to complete with ',' not only for List type attributes but also for the others attributes.

CC @jfdenise , please correct me if I am wrong.

jfdenise:
@ChristinaDsl , thank-you for taking over this issue. I would advise to first write down a small analysis document (or write directly a comment in the Issue) with what you plan to do. The CLI completion as a lot of edge cases that it would be good to first identify.

Some examples:
When completing (note that the list attributes is the last in the list of allowed attribute) `/host=primary/server-config=server-one/jvm=default:read-attribute(include-defaults,!include-undefined-metric-values,name=jvm-options` you should get "[" or ")", not "," (there are no more arguments in the operation).

Another one would be: `/subsystem=logging/logger=sun.rmi:write-attribute(name=filter-spec`, in such a case we should propose the ',' character or `/subsystem=logging/logger=sun.rmi:write-attribute(name=filter`, it should propose '.' and ','.

yersan:
Note: Integration Job failures are unrelated to this PR, the current branch had invalid changes merged, and they were already reverted.

yersan (changes requested):
@ChristinaDsl could you add the Jira number to your commit message? that's a requirement for WildfLy/WildFly Core PRs, thanks!

yersan (changes requested):
@ChristinaDsl It looks like you also have to take a look at CliCompletionTestCase.java since it is failing with your changes

ChristinaDsl:
There are some exceptions which seems irrelevant with the changes in the PR.

cli/src/main/java/org/jboss/as/cli/impl/AttributeNamePathCompleter.java
```diff
@@ -236,9 +236,11 @@ public int complete(String buffer, List<String> candidates, final ModelNode desc
             return -1;
         }
         final Collection<String> foundCandidates = handler.getCandidates(descr, writeOnly);
+
```
petrberan, line 239: Try to keep changes not related to your code to a minimum - such as new lines, deleted empty lines or adding some extra space. Those changes are shown and make reviewing your code harder and the file history less clear.

For example, the `ParserUtil.java` is flagged as changed, yet the only change is some extra space on 2 lines. Similar changes are in this class and in `OperationRequestCompleter.java`

yersan:
> There are some exceptions which seems irrelevant with the changes in the PR.

Yes @ChristinaDsl , notice we have been getting some issues with the CI; they were unrelated to this PR; related discussion is available in [Zulip](https://wildfly.zulipchat.com/#narrow/stream/174177-ci-servers/topic/WildFly.20Core.2020.2Ex.20Jobs)

ChristinaDsl:
Sorry for the late reply. Is there anything else that needs to be done from my side?

testsuite/domain/src/test/java/org/jboss/as/test/integration/domain/management/cli/CliCompletionTestCase.java
```diff
@@ -1158,10 +1158,12 @@ public void testCommandsCompletion2() throws Exception {
                 List<String> candidates = new ArrayList<>();
                 ctx.getDefaultCommandCompleter().complete(ctx, cmd,
                         cmd.length(), candidates);
-                assertTrue(candidates.toString(), candidates.size() == 1);
+                assertTrue(candidates.toString(), candidates.size() == 2);
```
jfdenise, line 1161: @ChristinaDsl , why do we have 2 elements there? We should have namespaces only.
For example if you type: read-attribute name<TAB>, you should get read-attribute namespaces, I can see an added "" character that breaks this completion.

ChristinaDsl: If you type read-attribute name, I thought that it would be right to have two choices:
Either you are allowed to leave it as it is "read-attribute name" or
End up in "read-attribute namespaces"
That is why I added empty space as another possible completion.
If there is no such a case I will erase it.

jfdenise: @ChristinaDsl , Thank-you, I see, I feel that we should keep the original behavior, this change would be outside the scope of the fix.

ChristinaDsl: Ok, I will fix it then, thank you too.

ChristinaDsl: @jfdenise 
With my existed modifications, after chunks which has Model Type:
list
string or 
object 
it is possible to add a property separator (if completed chunk concerns operation) or "" (if it is a command). Thus, because "name" lies to the case of a completed string and command, in the same time, an empty space is proposed. This new addition concerns the general case of proposing a comma or empty space after a completed string (which is right behavior) and if it's changed then the issue should be resolved in a completely other way.

jfdenise: @ChristinaDsl , I am not really following your comment there, could you give examples of what the problem is ? Thank-you.

ChristinaDsl: The part of existing code (not mine), which says that after a completed chunk, in possible completions should exist "," or "",  whether chunk is an operation or command accordingly, covers this case also( with "name" chunk). So, when "name" occurs, which is a command (and string in the same time), it is added in candidates the "".
Another example is typing "read-attribute organization" after hitting tab it proposes the empty space "", which, before my modifications, user had to type it by his own to continue.
Also after "read-attribute release-version" or "read-attribute server-state", "read-attribute uuid" and many others, happens the same thing.
Because "name" is a complete chunk by its own and  belongs to the same category as all of above which follows the "read-attribute" it has the same behaviour. Same happens with chunks following "read-operation".

ChristinaDsl: @jfdenise Does this makes sense? Would you like anything more?

jfdenise: @ChristinaDsl , I did a test, although read-attribute organization works as you describe, read-attribute management-minor-version doesn't automatically inline the " ".

ChristinaDsl: @jfdenise You are right. management-minor-version such as management-micro-version and management-major-version in chunkDescr has ModelType "INT". My modifications apply to three types: List, Object and String(in this last type is "name" included). Would you like me to add INT too?

ChristinaDsl: @jfdenise I didn't limit it on purpose. Very few cases are on the INT type,  while most of them are Strings. I was mainly testing it in Strings and that is why I did not take into consideration INT. I need to search it further in order to see if would be unwanted side effects in applying it to INT.

ChristinaDsl: @jfdenise What do you believe I need to do next?

jfdenise: Do you plan to apply to INT?

ChristinaDsl (commented):
I 've just added the case for INT. Please, let me know if it's fine or it needs any modifications.

jfdenise:
@ChristinaDsl , I can see a lot of reformatting in this change. Would you mind remove them? It is better to just have what changed in the commit. Thank-you.

ChristinaDsl:
@jfdenise Is there anything else you need from my side?

jfdenise:
@ChristinaDsl , I gave a quick try to your changes. Thank-you for the cleanup you made in formatting.
I can still see that completing read-attribute name{TAB} doesn't complete with namespaces. We have "" and namespaces.
Another case I observed: 
/host=primary/server-config=server-one/jvm=default:write-attribute(name=jvm-options,value={TAB}
 gives ')' and '[' although it should give only '['
It would be great to add a unit test to cover this case to the existing domain CLICompletionTest.
Thank-you.

ChristinaDsl:
@jfdenise 
Thanks for you comment.
I just tried again the case (both in standalone and domain mode) with completing read-attribute name{TAB}. It only gives as completion "namespaces", not " " and "namespaces..
Also for the second case, tried in 
/host=primary/jvm=default:write-attribute(name=jvm-options,value={TAB} (because the is no jvm=default in server-one) and only completes with '['

jfdenise:
@ChristinaDsl , I pulled this branch: https://github.com/ChristinaDsl/wildfly-core/tree/WFCORE-4916 (BTW you should resync with WildFly main, this branch is now very old).
The last commit I have is 8c235d0679112e75176d2b5227e392560d8a6f6e Are we on the same page?

jfdenise:
@ChristinaDsl , we can see an extra commit from your merge with the main branch. 
You should call: `git pull --rebase upstream main` doing so no extra commit will show-up, your commit will be put on top of the WildFly main branch head.

ChristinaDsl:
@jfdenise Did anything changed now?

ChristinaDsl:
@jfdenise I 've just done it. Thank you.

jfdenise:
@ChristinaDsl , I still see the behavior changes.

The testcase that has been added highlight it: https://github.com/wildfly/wildfly-core/pull/5415/files#diff-ba73312e1be9d81f3a69edc2787bdaf815887f361d0fa42fbe27297a82d072deR1144
When we press TAB on read-attribute name[TAB] we should only have namespaces inlined in the command (the only choice).

For `/host=primary/jvm=default:write-attribute(name=jvm-options,value=TAB` I still see ')' and '['. This testcase change highlights this change of behavior: https://github.com/wildfly/wildfly-core/pull/5415/files#diff-d6fb429ff7719b30c8a05809d02a6aaf1bf77bafdb916e9fc8cfead430b2aad2R139

ChristinaDsl:
@jfdenise thanks for your comment. Finally today I can see the two things you are speaking about locally..
Regardings the TAB after read-attribute name discussion, I thought that we agreed in our previous conversation, few months ago, that there will be two choices ("namespaces" and ""). You can see in our previous comments my arguments about it.
I just fixed the second matter, after my change it should appear only "[" after /host=primary/jvm=default:write-attribute(name=jvm-options,value=TAB

jfdenise:
@ChristinaDsl , it seems that yo ushould fix the ExplodedDeploymentTest, ony '[' is proposed.
For the "name/namespaces" behavior change, we can't avoid it?

ChristinaDsl:
@jfdenise After my modifications, after list, string or object  a property separator(in case last chunk is an operation) can be added, or an empty space (if last chunk is a command). "name" lies both in string & command, so according to above an empty space "" is proposed. This lies to a general pattern which is made and it says that a comma or empty space is proposed after a completed string. 
Another example in the same pattern is typing "read-attribute release-version". After hitting tab it proposes the empty space "", which, before my modifications, user had to type it by his own to continue. Now, lies to command & string case like "name". There are many others like this which follow the same logic/parts of code. If I change this then the issue should be resolved in another way from the beginning and that is why is difficult to avoid it.
As for ExplodedDeploymentTest, I will fix it.
Thank you.

jfdenise (approved):
We have a real usability gain with your changes, I much prefer the new completion behavior. I am approving it. Thank-you very much for your first contribution to the WildFly CLI!

ChristinaDsl:
@jfdenise 
I am grateful for my contribution to CLI. Thank you very much!

yersan:
Thanks, @ChristinaDsl and @jfdenise

PR #5417, opened 2023-03-22
[WFCORE-6268] Remove use of deprecated ModelTestControllerVersion enums

bstansberry:
https://issues.redhat.com/browse/WFCORE-6268


yersan (approved):

bstansberry:
@yersan I put a 'hold' on this as it looks like I missed a use of an enum value in the preparatory work I did for this in full WF. So that would need to be cleaned up first.

yersan (approved):
The failed org.jboss.as.logging.RootSubsystemOperationsTestCase does not use ModelTestControllerVersion, failure is unrelated

bstansberry:
Thanks @yersan

PR #5418, opened 2023-03-22
Fix for WFCORE-6270, Bootable JAR, Security providers are not registered during boot

jfdenise:
Issue: https://issues.redhat.com/browse/WFCORE-6270

### DIFF
--- a/bootable-jar/runtime/src/main/java/org/wildfly/core/jar/runtime/BootableJar.java
+++ b/bootable-jar/runtime/src/main/java/org/wildfly/core/jar/runtime/BootableJar.java
@@ -16,6 +16,8 @@
  */
 package org.wildfly.core.jar.runtime;
 
+import static java.security.AccessController.doPrivileged;
+
 import java.io.BufferedReader;
 import java.io.FileInputStream;
 import java.io.IOException;
@@ -27,10 +29,18 @@ import java.nio.charset.StandardCharsets;
 import java.nio.file.Files;
 import java.nio.file.Path;
 import java.nio.file.StandardOpenOption;
+import java.security.AccessControlContext;
+import java.security.PrivilegedAction;
+import java.security.ProtectionDomain;
+import java.security.Provider;
+import java.security.Security;
 import java.util.ArrayList;
 import java.util.Collections;
+import java.util.Iterator;
 import java.util.List;
 import java.util.Properties;
+import java.util.ServiceConfigurationError;
+import java.util.ServiceLoader;
 import java.util.concurrent.ExecutorService;
 import java.util.concurrent.Executors;
 import java.util.concurrent.TimeUnit;
@@ -331,6 +341,24 @@ public final class BootableJar implements ShutdownHandler {
         // At this point we can configure JMX
         configureJMX(moduleClassLoader, bootableJar.log);
 
+        // Automatic loading of Security providers.
+        // Needed for logic that requires access to providers prior elytron subsystem is configured.
+        final ServiceLoader<Provider> providerServiceLoader = ServiceLoader.load(Provider.class, moduleClassLoader);
+        SecurityManager sm = System.getSecurityManager();
+        Iterator<Provider> iterator = providerServiceLoader.iterator();
+        for (;;) try {
+            if (! (iterator.hasNext())) break;
+            final Provider provider = iterator.next();
+            if (sm == null) {
+                new AddProviderAction(provider).run();
+            } else {
+                final Class<? extends Provider> providerClass = provider.getClass();
+                // each provider needs permission to install itself
+                doPrivileged(new AddProviderAction(provider), getProviderContext(providerClass));
+            }
+        } catch (ServiceConfigurationError | RuntimeException e) {
+            bootableJar.log.securityProviderFailed(e);
+        }
         bootableJar.run();
     }
 
@@ -439,4 +467,22 @@ public final class BootableJar implements ShutdownHandler {
             }
         }
     }
+
+    static final class AddProviderAction implements PrivilegedAction<Void> {
+        private final Provider provider;
+
+        AddProviderAction(final Provider provider) {
+            this.provider = provider;
+        }
+
+        public Void run() {
+            Security.addProvider(provider);
+            return null;
+        }
+    }
+
+    private static AccessControlContext getProviderContext(final Class<? extends Provider> providerClass) {
+        return new AccessControlContext(new ProtectionDomain[]{providerClass.getProtectionDomain()});
+    }
+
 }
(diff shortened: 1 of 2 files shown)


darranl:
I assume we have this loop in the WildFly boot, do you have a reference to it?

bootable-jar/runtime/src/main/java/org/wildfly/core/jar/runtime/BootableJar.java
```diff
@@ -331,6 +340,19 @@ public static void run(Path jbossHome, List<String> args, ModuleLoader moduleLoa
...
+        // Automatic loading of Security providers.
+        // Needed for logic that requires access to providers prior elytron subsystem is configured.
+        final ServiceLoader<Provider> providerServiceLoader = ServiceLoader.load(Provider.class, moduleClassLoader);
+        Iterator<Provider> iterator = providerServiceLoader.iterator();
+        for (;;) try {
+            if (! (iterator.hasNext())) break;
+            final Provider provider = iterator.next();
+            final Class<? extends Provider> providerClass = provider.getClass();
+            // each provider needs permission to install itself
+            doPrivileged(new AddProviderAction(provider), getProviderContext(providerClass));
```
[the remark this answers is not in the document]
jfdenise: Thanks, fixed.

darranl:
@jfdenise Are you seeing us hitting that logic in JBoss Modules?  That looks to be triggered by a command line argument but I don't see that in our scripts.

bootable-jar/runtime/src/main/java/org/wildfly/core/jar/runtime/BootableJar.java
```diff
@@ -331,6 +341,24 @@ public static void run(Path jbossHome, List<String> args, ModuleLoader moduleLoa
         // At this point we can configure JMX
         configureJMX(moduleClassLoader, bootableJar.log);
 
+        // Automatic loading of Security providers.
+        // Needed for logic that requires access to providers prior elytron subsystem is configured.
+        final ServiceLoader<Provider> providerServiceLoader = ServiceLoader.load(Provider.class, moduleClassLoader);
+        Iterator<Provider> iterator = providerServiceLoader.iterator();
+        SecurityManager sm = System.getSecurityManager();
+        for (;;) try {
+            if (! (iterator.hasNext())) break;
```
jamezp, line 350: Any reason not to just use:
```java
for (Provider provider : providerServiceLoader)
````

File: bootable-jar/runtime/src/main/java/org/wildfly/core/jar/runtime/BootableJar.java [hunk 3]
```diff
@@ -331,6 +341,24 @@ public static void run(Path jbossHome, List<String> args, ModuleLoader moduleLoa
...
+            final Provider provider = iterator.next();
+            if (sm == null) {
+                new AddProviderAction(provider).run();
+            } else {
+                final Class<? extends Provider> providerClass = provider.getClass();
+                // each provider needs permission to install itself
+                doPrivileged(new AddProviderAction(provider), getProviderContext(providerClass));
+            }
```
jamezp, line 358: Is there a reason we do escalate to a privileged action here? There are other places in this entry point that would trigger SM failures if enabled. It seems odd to allow it in this specific case.

bootable-jar/runtime/src/main/java/org/wildfly/core/jar/runtime/_private/BootableJarLogger.java
```diff
@@ -165,4 +165,8 @@ public interface BootableJarLogger extends BasicLogger {
 
     @Message(id = Message.NONE, value = "Path to a CLI script to execute when starting the Bootable JAR")
     String argCliScript();
+
+    @LogMessage(level = DEBUG)
+    @Message(id = 25, value = "Failed to initialize a security provider. Exception: %s")
```
jamezp, line 170: Minor nit, but should the last word be "Reason" instead of "Exception"?

bootable-jar/runtime/src/main/java/org/wildfly/core/jar/runtime/BootableJar.java
jfdenise: I understand it this way:  Bootable JAR code base is granted all permissions. When adding a provider, the provider domain is used to restrict the set of permissions. FYI, this is the way it is implemented in WildFly boot sequence.

bootable-jar/runtime/src/main/java/org/wildfly/core/jar/runtime/_private/BootableJarLogger.java
jfdenise: Good! Fixed.

bootable-jar/runtime/src/main/java/org/wildfly/core/jar/runtime/BootableJar.java
darranl: If we are looping I think we do need to do it the way it was written in JBoss Modules, I believe David selected that pattern based on experiences around lazy loading of the classes and problems encountered during loading.

darranl (changes requested):
There was a reason behind the pattern in JBoss Modules.

bootable-jar/runtime/src/main/java/org/wildfly/core/jar/runtime/BootableJar.java
```diff
@@ -331,6 +340,23 @@ public static void run(Path jbossHome, List<String> args, ModuleLoader moduleLoa
         // At this point we can configure JMX
         configureJMX(moduleClassLoader, bootableJar.log);
 
+        // Automatic loading of Security providers.
+        // Needed for logic that requires access to providers prior elytron subsystem is configured.
+        final ServiceLoader<Provider> providerServiceLoader = ServiceLoader.load(Provider.class, moduleClassLoader);
+        SecurityManager sm = System.getSecurityManager();
+        for (Provider provider : providerServiceLoader) {
```
darranl, line 347: I believe this line can throw the RuntimeException.

The pattern in JBoss Modules installs all the providers that can be loaded, this installs all or refuses to boot.

jfdenise:
@darranl I reused the JBoss Modules pattern.

darranl (approved):

bootable-jar/runtime/src/main/java/org/wildfly/core/jar/runtime/BootableJar.java
jamezp: I don't have any argument against it either way, but the byte code ends up being pretty much the same.

darranl: In this version the try / catch block wraps the iteration calls so will try again if one fails, in the for Provider version the try / catch block was inside the for loop so the process would exit on the first error.

The Java code could have changed since but I believe the iterator still moves to the next item even if creating it fails.

fjuma (approved):
Thanks @jfdenise!

yersan:
This PR already passed the CI, there is no need to kick them off again

PR #5426, opened 2023-03-30
[WFCORE-6277] adds wildflyee.api module dependency to java.se

emmartins:
Issue: https://issues.redhat.com/browse/WFCORE-6277

### DIFF
--- a/core-feature-pack/common/src/main/resources/modules/system/layers/base/wildflyee/api/main/module.xml
+++ b/core-feature-pack/common/src/main/resources/modules/system/layers/base/wildflyee/api/main/module.xml
@@ -50,5 +50,7 @@
         <module name="org.glassfish.jakarta.el" export="true" optional="true" />
         <module name="javax.xml.rpc.api" export="true" optional="true" />
         <module name="org.omg.api" export="true" optional="true" />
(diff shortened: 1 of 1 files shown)


yersan:
Added Hold label, since this bug is still under discussion

yersan (commented):
@ropalka  what's your feeling about this?

I'm not fully getting the root cause yet, accordingly to the changes in https://github.com/wildfly/wildfly-core/pull/5384, the `java.se` module should be available to all the deployments since it is added as a dependency to the Deployment module via `ServerDependenciesProcessor`.

jamezp:
Is there a discussion about this somewhere? There is no JIRA description so it's not clear if this is causing issues somewhere.

yersan:
We can follow up the discussion in [Zulip](https://wildfly.zulipchat.com/#narrow/stream/174184-wildfly-developers/topic/WFCORE-6277)

ropalka:
I removed javax.api but in this particular case it is obvious module is also used outside from WildFly runtime so java.se explicit dependency is needed.

yersan:
Thanks for adding your review @ropalka

yersan:
And @emmartins for fix it :)

PR #5427, opened 2023-03-30
WFCORE-6278 Overload EnumValidator.create(...) to avoid redundant set copy

pferraro:
... when validator accepts all enum values.

https://issues.redhat.com/browse/WFCORE-6278


yersan (approved):
RootSubsystemOperationsTestCase failure on Windows is unrelated and should be fixed in https://github.com/wildfly/wildfly-core/pull/5431

yersan:
thanks @pferraro

PR #5429, opened 2023-03-30
[WFCORE-6276] RegexRoleMapperTest does not clean up resources

ivassile:
Issue: https://issues.redhat.com/browse/WFCORE-6276


yersan (approved):
RootSubsystemOperationsTestCase failure on Windows is unrelated and should be fixed in https://github.com/wildfly/wildfly-core/pull/5431

PR #5432, opened 2023-04-02
[WFCORE-6283] Make TestModule.getModulesDirectory public

gaol:
Issue: https://issues.redhat.com/browse/WFCORE-6283

### DIFF
--- a/testsuite/shared/src/main/java/org/jboss/as/test/module/util/TestModule.java
+++ b/testsuite/shared/src/main/java/org/jboss/as/test/module/util/TestModule.java
@@ -270,7 +270,7 @@ public class TestModule {
         return new File(getModulesDirectory(createParent), this.moduleName.replace('.', File.separatorChar));
     }
 
-    static File getModulesDirectory(boolean create) {
+    public static File getModulesDirectory(boolean create) {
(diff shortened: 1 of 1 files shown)


gaol:
@yersan would you please review? thanks

gaol:
@yersan sure, my test case is at: https://github.com/wildfly/wildfly/compare/main...gaol:wildfly:ra_load_from_module?expand=1

I need to create `META-INF/ra.xml` inside of the module, like: `<modules>/org/jboss/ironjacamar/ra16out/main/META-INF/ra.xml`, and there is no such support in TestModule. So I need to know the directory of the module, and create the file there.

gaol:
@yersan no, because the `META-INF/ra.xml` file exists on the side of `ra16out.jar`, not inside of it.  the module structure is like:
```bash
[🎩 lgao@lins-p1 basic]$ tree target/modules
target/modules
└── org
    └── jboss
        └── ironjacamar
            └── ra16out
                └── main
                    ├── META-INF
                    │   └── ra.xml
                    ├── module.xml
                    └── ra16out.jar

```
We have such module structure in `org.jboss.genericjms` module.

PR #5434, opened 2023-04-03
[WFCORE-6279] Replace Deployment Processors Javax dependencies with Jakarta modules

yersan:
And take into account if the user is adding aliases as exclusions and warn if a user dependency has been added via a module-alias

Jira issue: https://issues.redhat.com/browse/WFCORE-6279

### DIFF
--- a/server/src/main/java/org/jboss/as/server/deployment/module/DriverDependenciesProcessor.java
+++ b/server/src/main/java/org/jboss/as/server/deployment/module/DriverDependenciesProcessor.java
@@ -30,7 +30,6 @@ import org.jboss.as.server.deployment.DeploymentUnit;
 import org.jboss.as.server.deployment.DeploymentUnitProcessingException;
 import org.jboss.as.server.deployment.DeploymentUnitProcessor;
 import org.jboss.modules.Module;
-import org.jboss.modules.ModuleIdentifier;
 import org.jboss.modules.ModuleLoader;
 import org.jboss.vfs.VirtualFile;
 
@@ -39,7 +38,7 @@ import org.jboss.vfs.VirtualFile;
  */
 public class DriverDependenciesProcessor implements DeploymentUnitProcessor {
     private static final String SERVICE_FILE_NAME = "META-INF/services/java.sql.Driver";
-    private static final ModuleIdentifier JTA = ModuleIdentifier.create("javax.transaction.api");
+    private static final String JTA = "jakarta.transaction.api";
 
     @Override
     public void deploy(final DeploymentPhaseContext phaseContext) throws DeploymentUnitProcessingException {
--- a/server/src/main/java/org/jboss/as/server/deployment/module/ModuleAliasChecker.java
+++ b/server/src/main/java/org/jboss/as/server/deployment/module/ModuleAliasChecker.java
@@ -1,0 +1,109 @@
+/*
+ * Copyright 2023 Red Hat, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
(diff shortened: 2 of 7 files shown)


server/src/main/java/org/jboss/as/server/deployment/module/DriverDependenciesProcessor.java
```diff
@@ -39,7 +38,7 @@
  */
 public class DriverDependenciesProcessor implements DeploymentUnitProcessor {
     private static final String SERVICE_FILE_NAME = "META-INF/services/java.sql.Driver";
-    private static final ModuleIdentifier JTA = ModuleIdentifier.create("javax.transaction.api");
+    private static final String JTA = "jakarta.transaction.api";
```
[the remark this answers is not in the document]
yersan: Thanks for the review, I left the others as they are because they were public, but after reading your comment I realized the class is package visible, so they can also be replaced from ModuleIdentifier to String.

server/src/main/java/org/jboss/as/server/deployment/module/ModuleAliasChecker.java
```diff
@@ -0,0 +1,85 @@
...
+        final ModuleLoader moduleLoader = Module.getBootModuleLoader();
+        try {
+            Module module = moduleLoader.loadModule(identifier);
+            String moduleName = module.getName();
+            if (!moduleName.equals(identifier)) {
+                return moduleName;
+            }
+        } catch (ModuleLoadException e) {
+            // Explicitly ignored, we are not interested in knowing whether the module cannot be loaded at this point.
+        }
```
bstansberry, line 62: I think we should wrap L53-62 in a 'if (identifier.startsWith("javax.")) {' block. Just to be conservative at the start.

bstansberry: if (identifier.startsWith("javax." && identifier.contains(".api")) {

yersan: That will be ok, but we need to be sure all the work replacing DUPs only affect to javax modules

server/src/main/java/org/jboss/as/server/deployment/module/ModuleAliasChecker.java
```diff
@@ -0,0 +1,85 @@
...
+     */
+    static String getTargetModule(String identifier) {
+        final ModuleLoader moduleLoader = Module.getBootModuleLoader();
+        try {
+            Module module = moduleLoader.loadModule(identifier);
+            String moduleName = module.getName();
+            if (!moduleName.equals(identifier)) {
+                return moduleName;
+            }
+        } catch (ModuleLoadException e) {
```
bstansberry, line 60: I'd prefer 'catch (ModuleLoadException | RuntimeException e)'

Again, just to be conservative.

bstansberry (commented):
@yersan I've made a couple comments aimed at making this more conservative. If you are ok with those and they pass CI, go ahead and merge this for WFC 20.

server/src/main/java/org/jboss/as/server/deployment/module/ModuleAliasChecker.java
yersan: @bstansberry done

server/src/main/java/org/jboss/as/server/deployment/module/ModuleAliasChecker.java
yersan: Done

bstansberry (approved):

PR #5437, opened 2023-04-04
[WFCORE-6288] Manifest class path processor patch

huuskart:
Avoid entering duplicated dependencies from class path entries, to improve deployment speed.

### DIFF
--- a/server/src/main/java/org/jboss/as/server/deployment/module/ManifestClassPathProcessor.java
+++ b/server/src/main/java/org/jboss/as/server/deployment/module/ManifestClassPathProcessor.java
@@ -141,6 +141,7 @@ public final class ManifestClassPathProcessor implements DeploymentUnitProcessor
             }
         }
 
+        Set<VirtualFile> processedClassPathFiles = new HashSet<>();
         //additional resource roots may be added as
         while (!resourceRoots.isEmpty()) {
             final RootEntry entry = resourceRoots.pop();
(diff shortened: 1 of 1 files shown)


yersan:
Hello @huuskart , thanks for contributing to WildFly Core. Notice that the following message is shown when you are going to create a Pull Request:

```
```

We are required to add the link to the Jira to the Pull Request description and the Jira ID to each commit. In addition, please squash your commits in a single one, that helps when changes need to be backported to other branches if required.

server/src/main/java/org/jboss/as/server/deployment/module/ManifestClassPathProcessor.java
```diff
@@ -141,11 +141,17 @@ public synchronized void deploy(final DeploymentPhaseContext phaseContext) throw
...
+        Attachable lastTarget = null;
         //additional resource roots may be added as
         while (!resourceRoots.isEmpty()) {
             final RootEntry entry = resourceRoots.pop();
             final ResourceRoot resourceRoot = entry.resourceRoot;
             final Attachable target = entry.target;
+            if (target != lastTarget) {
+                processedClassPathFiles.clear();
+                lastTarget = target;
+            }
```
bstansberry, line 154: Please add a comment explaining this logic, as this class is already really complex and it's not intuitive why this is done instead of just creating processedClassPathFiles inside the loop. Reviewing this carefully I get that the idea is the same Attachable can be attached to multiple resource roots, but it's not clear why it's only appropriate to check 'lastTarget'.

Perhaps this should use a Map<Attachable, Set<VirtualFile>>> for this tracking?

huuskart: Yes, you are right, Map<Attachable, Set>> is correct.

target != lastTarget worked for me by chance in my deployment, and in some other case it could have failed to achieve anything useful.

PR #5441, opened 2023-04-04
[WFCORE-6157] Allow OperationCancellationUnitTestCase to complete in a reasonable time.

bstansberry:
Also adds an another assert, and cleans up a bunch of IDE warns in the class.


bstansberry:
Thanks @ropalka

PR #5443, opened 2023-04-05
[WFCORE-6290] Failure to handle errors loading process-uuid file

lvydra:
https://issues.redhat.com/browse/WFCORE-6290
https://issues.redhat.com/browse/JBEAP-24716

### DIFF
--- a/controller/src/main/java/org/jboss/as/controller/logging/ControllerLogger.java
+++ b/controller/src/main/java/org/jboss/as/controller/logging/ControllerLogger.java
@@ -3739,4 +3739,12 @@ public interface ControllerLogger extends BasicLogger {
 
     @Message(id = 499, value = "There is no satisfactory capability '%s' available to resources with capability scope '%s'. This capability is registered at address(es) '%s', and are not accessible to resources with scope '%s'.")
     IllegalStateException noSatisfactoryCapability(String capability, String scopeName, Set<RegistrationPoint> addresses, String scopeNameAgain);
+
+    @LogMessage(level = WARN)
+    @Message(id = 500, value = "There is no UUID string at '%s'. A new value will be generated.")
+    void uuidIsEmpty(String path);
+
+    @LogMessage(level = WARN)
+    @Message(id = 501, value = "An invalid UUID string '%s' was found at '%s'. A new value will be generated.")
+    void uuidNotValid(String corruptedUuid, String path);
 }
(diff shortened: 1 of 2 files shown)


yersan (commented):
I added some comments about the warnings, but I'm not sure whether the current approach is the correct one.

Adding the ability to auto-recover from a corrupted `process-uuid` content looks good, but there is a consequence where restaring a server with a corrupted `process-uuid` could make the server start back with a different process id. I don't think that could have consequences on standalone but I'm not sure if that could be an issue in domain mode.

In such a case, maybe logging a good warning trace with a better description of why a server / managed server did not start could be more secure than auto-cover generating a new UUID. That moves the responsibility of fixing the corrupted file issue to the user.

I would like to hear other opinions about this; if you think domain mode won't be an issue if a server is restarted without a different `process-uuid`, I'm fine with these changes after improving the message content. This case is a corner case as, so I'm being pretty conservative here, which could not be so good.

bstansberry (approved):

yersan:
@bstansberry not really, I initially thought it could be related with something kept in the server inventory, but after looking at it, we do not kept this information, we kept the process name among others. This field seems to be only used by the product info, to report an instance identifier, so I guess we should be fine going ahead with this.

controller/src/main/java/org/jboss/as/controller/logging/ControllerLogger.java
```diff
@@ -3739,4 +3739,12 @@ OperationFailedRuntimeException capabilityAlreadyRegisteredInContext(String capa
 
     @Message(id = 499, value = "There is no satisfactory capability '%s' available to resources with capability scope '%s'. This capability is registered at address(es) '%s', and are not accessible to resources with scope '%s'.")
     IllegalStateException noSatisfactoryCapability(String capability, String scopeName, Set<RegistrationPoint> addresses, String scopeNameAgain);
+
+    @LogMessage(level = WARN)
+    @Message(id = 500, value = "There is no UUID string at '%s'. A new value will be generated.")
+    void uuidIsEmpty(String path, @Cause Throwable throwable);
```
yersan, line 3745: @lvydra Would you mind removing the @Cause from the output on both traces?

That will make the warning less scaring for the users, the error doesn't show anything relevant for the users and the trace already contains the file location and the invalid file and or the invalid value.

It should not prevent merging this, but I think it is better if we hide the stack trace here.

yersan (changes requested):

controller/src/main/java/org/jboss/as/controller/logging/ControllerLogger.java
lvydra: @yersan updated

yersan (approved):

yersan:
Thanks @lvydra

PR #5472, opened 2023-04-25
[WFCORE-6326] Revert the removal of the org.jboss.vfs module on deplo…

jamezp:
…yments.

https://issues.redhat.com/browse/WFCORE-6326

### DIFF
--- a/server/src/main/java/org/jboss/as/server/deployment/module/ServerDependenciesProcessor.java
+++ b/server/src/main/java/org/jboss/as/server/deployment/module/ServerDependenciesProcessor.java
@@ -40,17 +40,25 @@ import org.jboss.modules.ModuleLoader;
  */
 public class ServerDependenciesProcessor implements DeploymentUnitProcessor {
 
-    private static final String JAVA_SE = "java.se";
+    private static final String[] DEFAULT_MODULES = {
+            "java.se",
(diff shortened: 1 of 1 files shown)


kabir:
@jamezp I'm totally unfamiliar with this, but since https://issues.redhat.com/browse/WFLY-17921 mentions RestEasy Spring deployments, I am curious if this could be added to the subsystem rather than server wide. 

Apart from that, it looks good

jamezp:
> @jamezp I'm totally unfamiliar with this, but since https://issues.redhat.com/browse/WFLY-17921 mentions RestEasy Spring deployments, I am curious if this could be added to the subsystem rather than server wide.
> 
> Apart from that, it looks good

@kabir It could, but this affects all Spring deployments. My initial thought was to add it there because the test failures came up in the resteasy-spring project during an upgrade. However, when I looked into it seems it would affect any deployment using Spring.

bstansberry (approved):

bstansberry:
@jamezp Can you send up a PR against 20.x as well? Thanks.

yersan (approved):
This need to be documented back in [WFLY-17841](https://issues.redhat.com/browse/WFLY-17841)

jamezp:
Upstream PR #5476

PR #5476, opened 2023-04-26
[20.x][WFCORE-6326] Revert the removal of the org.jboss.vfs module on deplo…

jamezp:
…yments.

https://issues.redhat.com/browse/WFCORE-6326

Upstream #5472

### DIFF
--- a/server/src/main/java/org/jboss/as/server/deployment/module/ServerDependenciesProcessor.java
+++ b/server/src/main/java/org/jboss/as/server/deployment/module/ServerDependenciesProcessor.java
@@ -43,6 +43,9 @@ public class ServerDependenciesProcessor implements DeploymentUnitProcessor {
 
     private static ModuleIdentifier[] DEFAULT_MODULES = new ModuleIdentifier[] {
         ModuleIdentifier.create("java.se"),
+        // Currently this is required for Spring deployments. Spring identifies the resource protocol as "vfs" and
+        // attempts to use VFS to search for configuration files within the deployment.
+        ModuleIdentifier.create("org.jboss.vfs"),
     };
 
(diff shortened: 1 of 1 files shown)


server/src/main/java/org/jboss/as/server/deployment/module/ServerDependenciesProcessor.java
```diff
@@ -49,6 +49,9 @@ public class ServerDependenciesProcessor implements DeploymentUnitProcessor {
             // The Sun JDK is added as a dependency with service import = true since it's required for JSR-223 Javascript engine to be available.
             // @see https://issues.jboss.org/browse/AS7-1116 and https://issues.jboss.org/browse/WFLY-1373
             ModuleIdentifier.create("sun.jdk"),
+            // Currently this is required for Spring deployments. Spring identifies the resource protocol as "vfs" and
+            // attempts to use VFS to search for configuration files within the deployment.
+            ModuleIdentifier.create"org.jboss.vfs"),
```
yersan, line 54: @jamezp missing "("

Originally this "org.jboss.vfs" module was added via the `DEFAULT_MODULES` array, but now it will be added with service import, if this now required for recent Spring versions?

yersan (changes requested):

server/src/main/java/org/jboss/as/server/deployment/module/ServerDependenciesProcessor.java
bstansberry: +1. This should be added to the DEFAULT_MODULES array, where it was before, not DEFAULT_MODULES_WITH_SERVICE_IMPORTS. That's what the fix in main does.

jamezp: Sorry, that's what I get for trying to be quick and using the git mergetool :) Should be fixed now.

yersan (approved):

PR #5479, opened 2023-04-28
[WFCORE-6329] Remove deprecated ServerEnvironment API

bstansberry:
https://issues.redhat.com/browse/WFCORE-6329

Requires https://github.com/wildfly/wildfly/pull/16796 first

### DIFF
--- a/server/src/main/java/org/jboss/as/server/ServerEnvironment.java
+++ b/server/src/main/java/org/jboss/as/server/ServerEnvironment.java
@@ -74,7 +74,7 @@ public class ServerEnvironment extends ProcessEnvironment implements Serializabl
     /**
      * The manner in which a server can be launched
      */
-    public static enum LaunchType {
+    public enum LaunchType {
         /**
          * Launched by a Host Controller in a managed domain
          */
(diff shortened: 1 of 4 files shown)


server/src/main/java/org/jboss/as/server/ServerEnvironment.java
```diff
@@ -928,52 +845,6 @@ public File getHomeDir() {
...
-    public File getBundlesDir() {
-        return bundlesDir;
-    }
-
-    private void configureBundlesDir(String dirPath, Properties providedProperties) {
-        boolean haveDirProperty = dirPath != null;
-        File tmp = getFileFromPath(dirPath);
-        if (tmp == null) {
-            if (haveDirProperty) {
-                throw ServerLogger.ROOT_LOGGER.bundlesDirectoryDoesNotExist(new File(dirPath).getAbsoluteFile());
```
yersan, line 967: ServerLogger.ROOT_LOGGER.bundlesDirectoryDoesNotExist can be commented out after this.

server/src/main/java/org/jboss/as/server/ServerEnvironment.java
```diff
@@ -121,37 +121,6 @@ public ProcessType getProcessType() {
...
-    public static final String MODULES_DIR = "jboss.modules.dir";
-
-    /**
-     * Constant that holds the name of the system property for specifying
-     * {@link #getBundlesDir() the bundles directory}.
-     *
-     * <p>
-     * Defaults to <tt><em>HOME_DIR</em>/bundles</tt>/
-     */
-    public static final String BUNDLES_DIR = "jboss.bundles.dir";
```
yersan, line 153: I understand we can remove this without the previous deprecation because this directory for OSGi bundles was used only internally, is that correct?

server/src/main/java/org/jboss/as/server/ServerEnvironment.java
bstansberry: Thanks @yersan -- I commented that out. Some more metaspace saved!

server/src/main/java/org/jboss/as/server/ServerEnvironment.java
bstansberry: That said, the constant itself is internal API. The server module is private.

yersan (approved):

yersan:
Thanks @bstansberry

PR #5480, opened 2023-04-28
WFCORE-6321 Expose ServerEnvironment via a capability

parsharma:
Issue: https://issues.redhat.com/browse/WFCORE-6321

### DIFF
--- a/server/src/main/java/org/jboss/as/server/ServerEnvironmentService.java
+++ b/server/src/main/java/org/jboss/as/server/ServerEnvironmentService.java
@@ -22,6 +22,7 @@
 
 package org.jboss.as.server;
 
+import static org.jboss.as.server.ServerService.SERVER_ENVIRONMENT_CAPABILITY;
 import org.jboss.msc.service.Service;
 import org.jboss.msc.service.ServiceName;
 import org.jboss.msc.service.ServiceTarget;
@@ -54,7 +55,10 @@ import org.jboss.msc.service.StopContext;
  */
 public class ServerEnvironmentService implements Service<ServerEnvironment> {
 
+
     /** Standard ServiceName under which a ServerEnvironmentService would be registered */
+    /** @deprecated Use {@link #SERVER_ENVIRONMENT_CAPABILITY.getCapabilityServiceName() instead} */
+    @Deprecated
     public static final ServiceName SERVICE_NAME = ServiceName.JBOSS.append("server", "environment");
 
     /**
@@ -65,8 +69,9 @@ public class ServerEnvironmentService implements Service<ServerEnvironment> {
      * @param target the batch builder. Cannot be {@code null}
      */
     public static void addService(ServerEnvironment serverEnvironment, ServiceTarget target) {
-        target.addService(SERVICE_NAME, new ServerEnvironmentService(serverEnvironment))
-            .install();
+        target.addService(SERVER_ENVIRONMENT_CAPABILITY.getCapabilityServiceName(), new ServerEnvironmentService(serverEnvironment))
+                .addAliases(SERVICE_NAME)
+                .install();
     }
 
     private final ServerEnvironment serverEnvironment;
--- a/server/src/main/java/org/jboss/as/server/ServerService.java
+++ b/server/src/main/java/org/jboss/as/server/ServerService.java
@@ -178,6 +178,7 @@ public final class ServerService extends AbstractControllerService {
 
     static final String SUSPEND_CONTROLLER_CAPABILITY_NAME = "org.wildfly.server.suspend-controller";
     static final String EXTERNAL_MODULE_CAPABILITY_NAME = "org.wildfly.management.external-module";
+    public static final String SERVER_ENVIRONMENT_CAPABILITY_NAME = "org.wildfly.server.environment";
 
     static final RuntimeCapability<Void> SUSPEND_CONTROLLER_CAPABILITY =
             RuntimeCapability.Builder.of(SUSPEND_CONTROLLER_CAPABILITY_NAME, SuspendController.class)
@@ -186,6 +187,8 @@ public final class ServerService extends AbstractControllerService {
     static final RuntimeCapability<Void> EXTERNAL_MODULE_CAPABILITY =
             RuntimeCapability.Builder.of(EXTERNAL_MODULE_CAPABILITY_NAME, ExternalModule.class)
                     .build();
+   static final RuntimeCapability<Void> SERVER_ENVIRONMENT_CAPABILITY = RuntimeCapability.Builder.of(SERVER_ENVIRONMENT_CAPABILITY_NAME, ServerEnvironment.class).build();
+
 
     /**
      * Construct a new instance.
@@ -504,6 +507,8 @@ public final class ServerService extends AbstractControllerService {
                 new RuntimeCapabilityRegistration(EXTERNAL_MODULE_CAPABILITY, CapabilityScope.GLOBAL,new RegistrationPoint(PathAddress.EMPTY_ADDRESS, null)));
         capabilityRegistry.registerCapability(
                 new RuntimeCapabilityRegistration(CONSOLE_AVAILABILITY_CAPABILITY, CapabilityScope.GLOBAL, new RegistrationPoint(PathAddress.EMPTY_ADDRESS, null)));
+        capabilityRegistry.registerCapability(
+                new RuntimeCapabilityRegistration(SERVER_ENVIRONMENT_CAPABILITY, CapabilityScope.GLOBAL, new RegistrationPoint(PathAddress.EMPTY_ADDRESS, null)));
 
         // Record the core capabilities with the root MRR so reads of it will show it as their provider
         // This also gets them recorded as 'possible capabilities' in the capability registry
@@ -514,6 +519,7 @@ public final class ServerService extends AbstractControllerService {
         rootRegistration.registerCapability(PROCESS_STATE_NOTIFIER_CAPABILITY);
         rootRegistration.registerCapability(EXTERNAL_MODULE_CAPABILITY);
         rootRegistration.registerCapability(CONSOLE_AVAILABILITY_CAPABILITY);
+        rootRegistration.registerCapability(SERVER_ENVIRONMENT_CAPABILITY);
     }
 
     @Override
--- a/server/src/main/java/org/jboss/as/server/deployment/ManagedExplodedContentServitor.java
+++ b/server/src/main/java/org/jboss/as/server/deployment/ManagedExplodedContentServitor.java
@@ -16,6 +16,7 @@ limitations under the License.
 
 package org.jboss.as.server.deployment;
 
+import static org.jboss.as.server.ServerService.SERVER_ENVIRONMENT_CAPABILITY_NAME;
 import static org.jboss.as.server.Services.requireServerExecutor;
 
 import java.io.IOException;
@@ -29,15 +30,14 @@ import java.util.function.Consumer;
 import java.util.function.Supplier;
 import java.util.stream.Stream;
 
+import org.jboss.as.controller.OperationContext;
 import org.jboss.as.repository.ContentRepository;
 import org.jboss.as.repository.ExplodedContentException;
 import org.jboss.as.server.ServerEnvironment;
-import org.jboss.as.server.ServerEnvironmentService;
 import org.jboss.msc.Service;
 import org.jboss.msc.service.ServiceBuilder;
 import org.jboss.msc.service.ServiceController;
 import org.jboss.msc.service.ServiceName;
-import org.jboss.msc.service.ServiceTarget;
 import org.jboss.msc.service.StartContext;
 import org.jboss.msc.service.StartException;
 import org.jboss.msc.service.StopContext;
@@ -61,11 +61,11 @@ class ManagedExplodedContentServitor implements Service {
     private final byte[] hash;
     private volatile Path deploymentRoot;
 
-    static ServiceController<?> addService(final ServiceTarget serviceTarget, final ServiceName serviceName, final String managementName, final byte[] hash) {
-        final ServiceBuilder<?> sb = serviceTarget.addService(serviceName);
+    static ServiceController<?> addService(OperationContext context, final ServiceName serviceName, final String managementName, final byte[] hash) {
+        final ServiceBuilder<?> sb = context.getCapabilityServiceTarget().addService(serviceName);
         final Consumer<VirtualFile> vfConsumer = sb.provides(serviceName);
         final Supplier<ContentRepository> crSupplier = sb.requires(ContentRepository.SERVICE_NAME);
-        final Supplier<ServerEnvironment> seSupplier = sb.requires(ServerEnvironmentService.SERVICE_NAME);
+        final Supplier<ServerEnvironment> seSupplier = sb.requires(context.getCapabilityServiceName(SERVER_ENVIRONMENT_CAPABILITY_NAME, ServerEnvironment.class));
         final Supplier<ExecutorService> esSupplier = requireServerExecutor(sb);
         sb.setInstance(new ManagedExplodedContentServitor(managementName, hash, vfConsumer, crSupplier, seSupplier, esSupplier));
         return sb.install();
--- a/server/src/main/java/org/jboss/as/server/deployment/DeploymentHandlerUtil.java
+++ b/server/src/main/java/org/jboss/as/server/deployment/DeploymentHandlerUtil.java
@@ -201,7 +201,7 @@ public class DeploymentHandlerUtil {
                 contentService = ContentServitor.addService(serviceTarget, contentsServiceName, contents[0].hash);
             } else {
                 isExplodedContent = true;
-                contentService = ManagedExplodedContentServitor.addService(serviceTarget, contentsServiceName, managementName, contents[0].hash);
+                contentService = ManagedExplodedContentServitor.addService(context, contentsServiceName, managementName, contents[0].hash);
             }
         }
         else {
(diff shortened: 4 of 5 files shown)


server/src/main/java/org/jboss/as/server/ServerEnvironmentService.java
```diff
@@ -69,6 +72,9 @@ public static void addService(ServerEnvironment serverEnvironment, ServiceTarget
             .install();
     }
 
+    private static final String SERVER_ENVIRONMENT_SERVICE_NAME = "jboss.server.environment";
```
yersan, line 75: @bstansberry I'm not sure if we are following any specific rule for the capability names that replace the service usage, I think we should use something under `org.wildfly` name space. Could you add your feedback regarding it?

server/src/main/java/org/jboss/as/server/ServerEnvironmentService.java
```diff
@@ -69,6 +72,9 @@ public static void addService(ServerEnvironment serverEnvironment, ServiceTarget
...
+    static final RuntimeCapability<Void> SERVICE_ENVIRONMENTT_CAPABILITY = RuntimeCapability.Builder.of(SERVER_ENVIRONMENT_SERVICE_NAME, ServerEnvironment.class).build();
```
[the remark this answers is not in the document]
yersan: We also need to register the capability, I think the place for this one would be under the root resource. See https://github.com/wildfly/wildfly-core/blob/main/server/src/main/java/org/jboss/as/server/ServerService.java#L508-L509

yersan (changes requested):

yersan:
@parsharma we need the Jira Issue on the commit message and PR title.

parsharma:
@yersan sorry, I was busy on 7.4.12 GA. I will update the PR soon.

server/src/main/java/org/jboss/as/server/ServerEnvironmentService.java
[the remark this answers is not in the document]
pferraro: > It must be public so it can be accessed from other classes to build the service based on the capability or capability service name

The RuntimeCapability itself should not be public, as this is a registration detail that is not relevant/appropriate for consuming subsystems.  Only the name and its service interface need to be public.

server/src/main/java/org/jboss/as/server/ServerEnvironmentService.java
pferraro: @yersan I agree. I suggest using "org.wildfly.server.environment" or similar, where the exposed ServerEnvironment is also available via its legacy ServiceName for compatibility purposes.

yersan: Thanks for confirming it @pferraro 

> where the exposed ServerEnvironment is also available via its legacy ServiceName for compatibility purposes.

Right, we have to add an alias in this case

server/src/main/java/org/jboss/as/server/ServerEnvironmentService.java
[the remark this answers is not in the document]
parsharma: If we change the access modifier to private of capability https://github.com/wildfly/wildfly-core/pull/5480#discussion_r1259766713 then how it will access here https://github.com/wildfly/wildfly-core/blob/main/server/src/main/java/org/jboss/as/server/ServerService.java#L508-L509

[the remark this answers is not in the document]
yersan: Good point, I think this capability falls into the same type as the Path Manager for example, and instead of declaring it on the ServerEnvironmentService.java, we should declare and register it in the ServerService.java.

The path manager capability is declared in the parent class (AbstractControllerService), just because the same capability is shared across the other Model Controller services. 

Indeed, we have a Host Environment Service that should follow the same pattern (in a different Jira) and instead of declaring it on a ServerService.java, the Host Environment would be declared on DomainModelControllerService.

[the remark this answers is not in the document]
parsharma: Got it. Thanks

yersan:
@parsharma errors are legit.

You still need to address all the pending points commented on above: add a Jira number to the PR title, use an alias for the service, add a comment to the @Deprecarte annotation, review any usage of the current service server/src/main/java/org/jboss/as/server/ServerEnvironmentService.java#SERVICE_NAME and replace it with the capability, ensure the capability is correctly registered. In addition, when done, please squash the commits in a single one.

If you are pushing changes but you still think there is pending work to do, please, move the PR to a draft so we can ignore its review until is done from your side, thanks!

parsharma:
@yersan I will change it to draft because I am still working on it. Thanks for the review comments. I will update the code soon.

yersan (changes requested):

server/src/main/java/org/jboss/as/server/ServerService.java
```diff
@@ -514,6 +517,7 @@ protected void initModel(ManagementModel managementModel, Resource modelControll
         rootRegistration.registerCapability(PROCESS_STATE_NOTIFIER_CAPABILITY);
         rootRegistration.registerCapability(EXTERNAL_MODULE_CAPABILITY);
         rootRegistration.registerCapability(CONSOLE_AVAILABILITY_CAPABILITY);
+        rootRegistration.registerCapability(SERVER_ENVIRONMENT_CAPABILITY);
```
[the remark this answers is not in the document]
parsharma: Done

server/src/main/java/org/jboss/as/server/ServerService.java
```diff
@@ -178,6 +178,7 @@ public final class ServerService extends AbstractControllerService {
 
     static final String SUSPEND_CONTROLLER_CAPABILITY_NAME = "org.wildfly.server.suspend-controller";
     static final String EXTERNAL_MODULE_CAPABILITY_NAME = "org.wildfly.management.external-module";
+    static final String SERVER_ENVIRONMENT_SERVICE_NAME = "org.wildfly.server.environment";
```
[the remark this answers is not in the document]
parsharma: Done

server/src/main/java/org/jboss/as/server/ServerEnvironmentService.java
```diff
@@ -54,9 +56,14 @@
  */
 public class ServerEnvironmentService implements Service<ServerEnvironment> {
 
+
     /** Standard ServiceName under which a ServerEnvironmentService would be registered */
+    /** SERVICE_NAME is deprecated so use Runtime capability ServerService SERVER_ENVIRONMENT_CAPABILITY */
+    @Deprecated
```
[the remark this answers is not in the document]
parsharma: Done

server/src/main/java/org/jboss/as/server/ServerEnvironmentService.java
```diff
@@ -22,13 +22,16 @@
...
+
 import org.jboss.msc.service.Service;
 import org.jboss.msc.service.ServiceName;
 import org.jboss.msc.service.ServiceTarget;
 import org.jboss.msc.service.StartContext;
 import org.jboss.msc.service.StartException;
 import org.jboss.msc.service.StopContext;
 
+import static org.jboss.as.server.ServerService.SERVER_ENVIRONMENT_CAPABILITY;
+
```
yersan, line 34: Sorry @parsharma , I missed this one, static imports go before, take a look at your IDE settings and ensure you are using the Eclipse Formatter with the WildFly profile. The ide configs can be found here herehttps://github.com/wildfly/wildfly-core/tree/main/ide-configs

yersan (changes requested):
Minor detail about the import, but looks good to me

server/src/main/java/org/jboss/as/server/ServerEnvironmentService.java
parsharma: Done, thanks

yersan (approved):

server/src/main/java/org/jboss/as/server/ServerEnvironmentService.java
```diff
@@ -54,9 +55,14 @@
  */
 public class ServerEnvironmentService implements Service<ServerEnvironment> {
 
+
     /** Standard ServiceName under which a ServerEnvironmentService would be registered */
+    /** @deprecated Use {@link #SERVER_ENVIRONMENT_SERVICE_NAME instead} */
+    @Deprecated
     public static final ServiceName SERVICE_NAME = ServiceName.JBOSS.append("server", "environment");
 
+    public static final ServiceName SERVER_ENVIRONMENT_SERVICE_NAME = SERVER_ENVIRONMENT_CAPABILITY.getCapabilityServiceName();
```
pferraro, line 64: We don't want to create yet another public ServiceName.  Consumers are meant to acquire the ServerEnvironment via a capability requirement (using its name).

pferraro (changes requested):

server/src/main/java/org/jboss/as/server/ServerEnvironmentService.java
parsharma: In that case, capability should be public. It is the default right now. 
For example here https://github.com/wildfly/wildfly-core/blob/main/server/src/main/java/org/jboss/as/server/operations/HttpManagementAddHandler.java#L148

yersan: @pferraro I guess the issue without using a service name created from the capability is that it is required a capability service target to use the capability name for the server environment service. 

That's the reason I gave an LGTM as it is now, we have used the same approach for several services created from the capabilities, see for example the [MANAGEMENT_EXECUTOR](https://github.com/wildfly/wildfly-core/blob/main/server/src/main/java/org/jboss/as/server/ServerService.java#L160) at ServerService.

Maybe there is another way to do it, and if there is one, is unknown to me.

yersan: @pferraro , right!, I missed that link, thanks.

yersan: @pferraro I misunderstood the approach and made a mistake with my previous comment (thanks for ping me @parsharma). We cannot access the RuntimeCapability from [HttpManagementAddHandler.java#L148](https://github.com/wildfly/wildfly-core/blob/main/server/src/main/java/org/jboss/as/server/operations/HttpManagementAddHandler.java#L148) because the capability was requested to be declared with package visibility. 

If we do not expose the service name publicly in [ServerEnvironmentService.java](https://github.com/wildfly/wildfly-core/pull/5480/files/70d5b038e370e5029b8f07b683b423d2fed02bde#diff-1e87d3842cfaee888f1f6c3a1aa1d55132dd410d7577cfeb60179637f9e39dba), I understand we have to declare the capability with public visibility at ServerService so we can get the service name in HttpManagementAddHandler.java#L148 and for any service target that cannot manage capability names. Otherwise, if there is any other approach, please, describe it here since we do not see it yet.

pferraro: The logic in [HttpManagementAddHandler](https://github.com/wildfly/wildfly-core/blob/main/server/src/main/java/org/jboss/as/server/operations/HttpManagementAddHandler.java#L148) is incorrect.  See: https://github.com/wildfly/wildfly-core/pull/5516/commits/c9580133c7984edf3d1edaeeebd51d57ea4f8d8f
... taken from https://github.com/wildfly/wildfly-core/pull/5516 which incorporates this PR.

Perhaps it would best to pull that change into a separate JIRA.  I'll create a pull request for that momentarily, then @parsharma can base her work on that branch.

server/src/main/java/org/jboss/as/server/ServerEnvironmentService.java
```diff
@@ -54,9 +55,14 @@
  */
 public class ServerEnvironmentService implements Service<ServerEnvironment> {
```
gaol, line 56: Shouldn't we use `org.jboss.msc.Service` as `org.jboss.msc.service.Service` is deprecated already ? or we need to address that in a different issue and pr ? @pferraro @yersan

server/src/main/java/org/jboss/as/server/ServerEnvironmentService.java
parsharma: @pferraro I have one more doubt here https://github.com/wildfly/wildfly-core/pull/5480/files#diff-700a2581dad9cd3005f25b8c3a70acc323cd43d385ed8f8014a52b01a557605dR68 we are using ServiceBuilder so we need ServiceName. Do you think we need to change it to capabiltyBuilder?

server/src/main/java/org/jboss/as/server/ServerEnvironmentService.java
pferraro: Rewriting the ServerEnvironmentService class is not the goal of WFCORE-6321.

That said, I did address this deprecated MSC usage as part of https://github.com/wildfly/wildfly-core/pull/5516.
See: https://github.com/wildfly/wildfly-core/pull/5516/commits/8ff268a719349678b45b18f5973f1f68a9858ad6
Hopefully, someone will review this (and its dependencies) soon.

gaol: OK, I see. thanks!

server/src/main/java/org/jboss/as/server/ServerEnvironmentService.java
pferraro: > @pferraro I have one more doubt here https://github.com/wildfly/wildfly-core/pull/5480/files#diff-700a2581dad9cd3005f25b8c3a70acc323cd43d385ed8f8014a52b01a557605dR68 we are using ServiceBuilder so we need ServiceName. Do you think we need to change it to capabiltyBuilder?

Normally, ManagedExplodedContentServitor would indeed use the service target returned via OperationContext.getCapabilityServiceTarget(), however the ServiceBuilder currently returned via CapabilityServiceTarget.addService(ServiceName) does not support capability resolution.  This is being addressed by https://github.com/wildfly/wildfly-core/pull/5588
Until then, you can generate the requisite ServiceName via OperationContext.getCapabilityServiceName("org.wildfly.server.environment")

server/src/main/java/org/jboss/as/server/ServerEnvironmentService.java
```diff
@@ -65,11 +69,11 @@ public class ServerEnvironmentService implements Service<ServerEnvironment> {
...
     public static void addService(ServerEnvironment serverEnvironment, ServiceTarget target) {
-        target.addService(SERVICE_NAME, new ServerEnvironmentService(serverEnvironment))
-            .install();
+        target.addService(SERVER_ENVIRONMENT_CAPABILITY.getCapabilityServiceName(), new ServerEnvironmentService(serverEnvironment))
+                .addAliases(SERVICE_NAME)
+                .install();
     }
-
-    private final ServerEnvironment serverEnvironment;
+        private final ServerEnvironment serverEnvironment;
```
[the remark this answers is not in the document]
parsharma: Updated.

server/src/main/java/org/jboss/as/server/ServerService.java
```diff
@@ -186,6 +187,8 @@ public final class ServerService extends AbstractControllerService {
     static final RuntimeCapability<Void> EXTERNAL_MODULE_CAPABILITY =
             RuntimeCapability.Builder.of(EXTERNAL_MODULE_CAPABILITY_NAME, ExternalModule.class)
                     .build();
+   public static final RuntimeCapability<Void> SERVER_ENVIRONMENT_CAPABILITY = RuntimeCapability.Builder.of(SERVER_ENVIRONMENT_CAPABILITY_NAME, ServerEnvironment.class).build();
```
pferraro, line 190: We are still planning to reduce the visibility of this, correct?

parsharma: Maybe I missing something but when you mentioned `OperationContext.getCapabilityServiceName("org.wildfly.server.environment")` I thought you mean to your capability.getCapabilityServiceName()

pferraro: I meant that OperationContext.getCapabilityServiceName("org.wildfly.server.environment", ServerEnvironment.class) can be used to generate a ServiceName for use with a standard ServiceTarget.

parsharma: `getCapabilityServiceName("org.wildfly.server.environment", ServerEnvironment.class)` is not the static method. We can't use it like this.

parsharma: Got it. Thanks

server/src/main/java/org/jboss/as/server/deployment/ManagedExplodedContentServitor.java
```diff
@@ -43,6 +43,7 @@
 import org.jboss.msc.service.StopContext;
 import org.jboss.vfs.VFS;
 import org.jboss.vfs.VirtualFile;
+import org.jboss.as.controller.OperationContext;
```
yersan, line 46: This import is out of order.

server/src/main/java/org/jboss/as/server/deployment/ManagedExplodedContentServitor.java
```diff
@@ -61,11 +62,11 @@ class ManagedExplodedContentServitor implements Service {
     private final byte[] hash;
     private volatile Path deploymentRoot;
 
-    static ServiceController<?> addService(final ServiceTarget serviceTarget, final ServiceName serviceName, final String managementName, final byte[] hash) {
+    static ServiceController<?> addService(final ServiceTarget serviceTarget, final ServiceName serviceName, final String managementName, final byte[] hash, OperationContext context) {
```
pferraro, line 65: Why not replace the ServiceTarget argument with OperationContext? Passing both seems redundant.

pferraro (changes requested):

server/src/main/java/org/jboss/as/server/deployment/ManagedExplodedContentServitor.java
parsharma: Right.

pferraro (approved):
All looks good now.  Thank you for persevering @parsharma !

yersan:
Thanks, @parsharma and @pferraro!

PR #5481, opened 2023-04-28
WFCORE-6332: update license config files

michpetrov:
Issue: [WFCORE-6332](https://issues.redhat.com/browse/WFCORE-6332)

A port of wildfly/wildfly#16693 but Core has a single feature pack so I'm only updating the merger.xsl and moving licenses.xsl to the same relative location.


yersan (commented):
@michpetrov Could you describe here where the merger.xsl file is used?
I've noticed we are not using the `docs.licenses.merge` in then wildfly-core feature pack, that could be fine, but even using it, I don't get where the merger.xsl is used for the transformation.

michpetrov:
@yersan there's no explicit reference to the file but the `xml-merge` directive (in the [tasks.xml](https://github.com/wildfly/wildfly-core/blob/main/core-feature-pack/galleon-common/src/main/resources/packages/core.docs.licenses.merge/pm/wildfly/tasks.xml) file) requires a `merger.xml` file to exist

core-feature-pack/galleon-common/src/main/resources/packages/core.docs.licenses.merge/pm/wildfly/merger.xsl
```diff
@@ -59,36 +62,19 @@
...
-    </xsl:text>
             <xsl:copy>
-                <xsl:apply-templates select="@* | node()"/><xsl:text disable-output-escaping="no">  </xsl:text><source><xsl:call-template name="substring-after-last">
-                    <xsl:with-param name="string" select="$fileName"/>
-                    <xsl:with-param name="delimiter" select="$fileSeparator"/>
-                </xsl:call-template></source><xsl:text disable-output-escaping="no">
-    </xsl:text>
+                <xsl:apply-templates select="@* | node()"/>
+                <xsl:element name="source">
+                    <xsl:attribute name="from">core-feature-pack/galleon-common</xsl:attribute>
```
yersan, line 72: The attribute from was not added to the WildFly Full variant, is it expected here? it looks like even less relevant since WildFly Core only uses a single feature pack.

I don't think an additional attribute could affect the license processes done before, but well, just wanted to know the reason for this difference.

yersan (commented):
@michpetrov ok, found its usage inside of galleon-plugins at https://github.com/wildfly/galleon-plugins/blob/main/galleon-plugins/src/main/java/org/wildfly/galleon/plugin/config/XmlMerge.java#L110

core-feature-pack/galleon-common/src/main/resources/packages/core.docs.licenses.merge/pm/wildfly/merger.xsl
michpetrov: I've deleted it, I think that was a leftover from me testing how the licenses propagate to full

PR #5486, opened 2023-04-30
[WFCORE-6337] Remove deprecated PersistentResourceXmlDescription API;…

bstansberry:
… reduce member visibility

I also deprecated a couple unused builder methods the use cases for which are better handled in other ways.

https://issues.redhat.com/browse/WFCORE-6337


yersan:
Thanks @bstansberry

PR #5489, opened 2023-05-02
[WFCORE-6338]: A resource name can't match an attribute name of this resource.

ehsavoie:
* Fixing the case when the resource name matches a resource attribute name.

Jira: https://issues.redhat.com/browse/WFCORE-6338


yersan (commented):
Let's wait for the CI. In any case, looks the Jira description is not in sync with what this PR resolves, maybe the Jira description would deserve a better explanation.

yersan (approved):

yersan:
Thanks @ehsavoie

PR #5498, opened 2023-05-09
[WFCORE-6346] Module java.base must open java.net package to all unnamed modules

ropalka:
https://issues.redhat.com/browse/WFCORE-6346

### DIFF
--- a/core-feature-pack/common/src/main/resources/content/bin/common.bat
+++ b/core-feature-pack/common/src/main/resources/content/bin/common.bat
@@ -50,10 +50,12 @@ goto :eof
       set "DEFAULT_MODULAR_JVM_OPTIONS=!DEFAULT_MODULAR_JVM_OPTIONS! --add-opens=java.base/java.lang=ALL-UNNAMED"
       rem Needed by the MicroProfile REST Client subsystem
       set "DEFAULT_MODULAR_JVM_OPTIONS=!DEFAULT_MODULAR_JVM_OPTIONS! --add-opens=java.base/java.lang.invoke=ALL-UNNAMED"
-      rem Needed by JBoss Marshalling
-      set "DEFAULT_MODULAR_JVM_OPTIONS=!DEFAULT_MODULAR_JVM_OPTIONS! --add-opens=java.base/java.io=ALL-UNNAMED"
       rem Needed for marshalling of proxies
       set "DEFAULT_MODULAR_JVM_OPTIONS=!DEFAULT_MODULAR_JVM_OPTIONS! --add-opens=java.base/java.lang.reflect=ALL-UNNAMED"
+      rem Needed by JBoss Marshalling
+      set "DEFAULT_MODULAR_JVM_OPTIONS=!DEFAULT_MODULAR_JVM_OPTIONS! --add-opens=java.base/java.io=ALL-UNNAMED"
+      rem Needed by WildFly Http Client
+      set "DEFAULT_MODULAR_JVM_OPTIONS=!DEFAULT_MODULAR_JVM_OPTIONS! --add-opens=java.base/java.net=ALL-UNNAMED"
       rem Needed by WildFly Security Manager
       set "DEFAULT_MODULAR_JVM_OPTIONS=!DEFAULT_MODULAR_JVM_OPTIONS! --add-opens=java.base/java.security=ALL-UNNAMED"
       rem Needed for marshalling of collections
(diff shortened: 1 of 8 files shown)


bstansberry:
@ropalka @fjuma This seems like something to fix in the 20.x branch as well, for this month's WF 28.0.1, as it seems like a regression in WF 28.0.0 via the upgrade that brought in https://issues.redhat.com/browse/WEJBHTTP-108.

fjuma:
Good point, thanks @bstansberry.

ropalka:
> @ropalka @fjuma This seems like something to fix in the 20.x branch as well, for this month's WF 28.0.1, as it seems like a regression in WF 28.0.0 via the upgrade that brought in https://issues.redhat.com/browse/WEJBHTTP-108.

Yes, I agree.

launcher/src/test/java/org/wildfly/core/launcher/CommandBuilderTest.java
```diff
@@ -273,9 +273,11 @@ private void testJPMSArguments(final Collection<String> command, final int expec
         assertArgumentExists(command, "--add-exports=java.naming/com.sun.jndi.url.ldap=ALL-UNNAMED", expectedCount);
         assertArgumentExists(command, "--add-exports=java.naming/com.sun.jndi.url.ldaps=ALL-UNNAMED", expectedCount);
         assertArgumentExists(command, "--add-exports=jdk.naming.dns/com.sun.jndi.dns=ALL-UNNAMED", expectedCount);
+        assertArgumentExists(command, "--add-opens=java.base/java.lang=ALL-UNNAMED", expectedCount);
```
mchoma, line 276: Seems as out of scope of this change.

Double checking if this is intentional.

core-feature-pack/common/src/main/resources/content/bin/common.bat
```diff
@@ -50,10 +50,12 @@ goto :eof
       set "DEFAULT_MODULAR_JVM_OPTIONS=!DEFAULT_MODULAR_JVM_OPTIONS! --add-opens=java.base/java.lang=ALL-UNNAMED"
       rem Needed by the MicroProfile REST Client subsystem
       set "DEFAULT_MODULAR_JVM_OPTIONS=!DEFAULT_MODULAR_JVM_OPTIONS! --add-opens=java.base/java.lang.invoke=ALL-UNNAMED"
-      rem Needed by JBoss Marshalling
-      set "DEFAULT_MODULAR_JVM_OPTIONS=!DEFAULT_MODULAR_JVM_OPTIONS! --add-opens=java.base/java.io=ALL-UNNAMED"
       rem Needed for marshalling of proxies
       set "DEFAULT_MODULAR_JVM_OPTIONS=!DEFAULT_MODULAR_JVM_OPTIONS! --add-opens=java.base/java.lang.reflect=ALL-UNNAMED"
+      rem Needed by JBoss Marshalling
+      set "DEFAULT_MODULAR_JVM_OPTIONS=!DEFAULT_MODULAR_JVM_OPTIONS! --add-opens=java.base/java.io=ALL-UNNAMED"
```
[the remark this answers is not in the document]
yersan: @mchoma This one should be fine, it was already there and down from L53 to keep it in the same order as it was declared in the other files

launcher/src/test/java/org/wildfly/core/launcher/CommandBuilderTest.java
yersan: I think this is fine; this test is testing that the expected JVM modular options are also used when launching jar files or the server for the tests

yersan (approved):
LGTM, so approving from my side

launcher/src/test/java/org/wildfly/core/launcher/CommandBuilderTest.java
bstansberry: Yeah, this was just a missing assert that Rio took the opportunity to correct.

jamezp (approved):

fjuma:
@ropalka When you get a chance, please create a PR against the 20.x branch. Thanks!

ropalka:
Here you go @fjuma : https://github.com/wildfly/wildfly-core/pull/5500

bstansberry (approved):
@ropalka This LGTM but if there isn't one already, please file a JIRA to update https://github.com/wildfly/wildfly/blob/main/pom.xml#L175-L207

yersan:
Thanks, @ropalka and all involved in the review

PR #5500, opened 2023-05-10
[WFCORE-6346] Module java.base must open java.net package to all unnamed modules

ropalka:
https://issues.redhat.com/browse/WFCORE-6346

Upstream PR: https://github.com/wildfly/wildfly-core/pull/5498


bstansberry (approved):
@ropalka This LGTM but please include 28.x to update https://github.com/wildfly/wildfly/blob/28.x/pom.xml#L175-L207

PR #5494, opened 2023-05-05
[WFCORE-6341]: YAML: MapAttributeDefinition not properly processed for existing resources.

ehsavoie:
* A MapAttributeDefinition or an ObjectAttributeDefinition wasn't
   properly handled.

Jira: https://issues.redhat.com/browse/WFCORE-6341

Please note that this commit also integrates https://github.com/wildfly/wildfly-core/pull/5489

PR #5502, opened 2023-05-11
[WFCORE-6348]: Invalid YAML configuration fails silently.

ehsavoie:
* If the YAML doesn't match a resource we are now failing.

Jira: https://issues.redhat.com/browse/WFCORE-6348


yersan:
@ehsavoie Notice the WFCORE-6341 hash does not match the current hash of https://github.com/wildfly/wildfly-core/pull/5494

I don't know how smart Git is in this situation, but it is likely this PR will be a conflict, we will see it in some minutes :)

ehsavoie:
@yersan rebased and fully green :)

darranl (approved):

yersan:
Thanks @ehsavoie

PR #5511, opened 2023-05-22
[WFCORE-5544] Servers unable to connect in domain mode with Elytron without local auth

darranl:
This will be dependent on the following PR in WildFly Elytron  https://github.com/wildfly-security/wildfly-elytron/pull/1905, also I am still doing some allTests runs locally but I am submitting to start getting some reviews in.

https://issues.redhat.com/browse/WFCORE-5544

This is a refactoring of some quite old code now we no longer support legacy security as well so some reviewers may be because you worked at this code at some point in time - sorry but the commit history does not lie so you are experts for life :smile: 

Fundamentally we have always tried to get the domain servers to authenticate back using the same authentication mechanisms as the users use, this has always been problematic as once users change the enabled mechanisms we still need to support the selected set which is not always possible.  With WildFly Elytron is also a lot harder to intercept the username / password mechanisms to validate the token.

The original solution borrowed the token used for process controller authentication, within this PR I have split this out so the process controller loopback token is completely independent of the authentication token used by the domain server to connect to it's host controller.  For now this is just a 246 bit token generated using SecureRandom however as most of the process just considers it a variable length opaque token later on Elytron could also issue alternate tokens if we felt this was appropriate.  In the future we may also be able to consider a solution where the domain server does not need to have a new token issued just because the host controller restarts.

To avoid the interaction with the mechanisms used by the users this uses a new mechanism called JBOSS-DOMAIN-SERVER, for all of our clients if they are offered a mechanism they don't support they just ignore it and use the mechanisms they do understand, this means that the only clients that will understand this mechanism are the domain servers.  Additionally the domain servers have all other mechanisms disabled so they will no longer accidentally use any other mechanism (including JBOSS-LOCAL-USER).  When the new mechanism successfully authenticates a server it also grants the JBOSS-DOMAIN-SERVER role so this can be checked for domain server calls, again making it impossible to use the wrong mechanism.

As before we have needed in intermediary service as the management interfaces and the related security services need to be up before we bring up the server inventory.  I believe the inventory only comes up late so we can confirm the address of the management endpoint so maybe that could be handled in a different way but for now this uses an intermediary to bridge the gap.


darranl:
Running -DallTests on WildFly Core and WildFly locally is green for me.

embedded/src/main/java/org/wildfly/core/embedded/EmbeddedHostControllerFactory.java
```diff
@@ -271,10 +270,10 @@ public void exit(int status) {
                     HostControllerEnvironment environment = createHostControllerEnvironment(jbossHomeDir, cmdargs, startTime);
 
                     FutureServiceContainer futureContainer = new FutureServiceContainer();
-                    final byte[] authBytes = new byte[ProcessController.AUTH_BYTES_LENGTH];
```
yersan, line 274: What's the reason for not continuing to use `ProcessController.AUTH_BYTES_LENGTH here`?

If we want as it is now, then the compile-time maven dependency on wildfly-process-controller could be also removed: https://github.com/wildfly/wildfly-core/blob/main/embedded/pom.xml#L198-L211

host-controller/src/main/java/org/jboss/as/host/controller/ServerInventoryImpl.java
```diff
@@ -227,6 +223,20 @@ public ServerStatus startServer(final String serverName, final ModelNode domainM
...
+    private String createServerAuthToken(final String serverName) {
+        // For now this is hard coded but at a later point Elytron may start to issue
+        // a different token so if we plug in an alternative approach it can come through
+        // this method.
+
+        // Create a new serverAuthToken
+        final byte[] tokenBytes = new byte[16]; // 16 for now but will be making independent.
+        new Random(new SecureRandom().nextLong()).nextBytes(tokenBytes);
+        String serverAuthToken = Base64.getEncoder().encodeToString(tokenBytes);
+        serverAuthToken = "SIK" + serverAuthToken.substring(3);
```
fjuma: Just curious, what does "SIK" stand for?

host-controller/src/main/java/org/jboss/as/host/controller/ServerInventoryImpl.java
```diff
@@ -229,10 +228,11 @@ private String createServerAuthToken(final String serverName) {
         // this method.
 
         // Create a new serverAuthToken
-        final byte[] tokenBytes = new byte[16]; // 16 for now but will be making independent.
-        new Random(new SecureRandom().nextLong()).nextBytes(tokenBytes);
-        String serverAuthToken = Base64.getEncoder().encodeToString(tokenBytes);
-        serverAuthToken = "SIK" + serverAuthToken.substring(3);
```
fjuma: Ignore my previous comment, I see now that "SIK" was renamed to "SAT" here.

process-controller/src/main/java/org/jboss/as/process/ProcessController.java
```diff
@@ -342,7 +340,9 @@ void sendInventory() {
                         StreamUtils.writeInt(os, processCollection.size());
                         for (ManagedProcess process : processCollection) {
                             StreamUtils.writeUTFZBytes(os, process.getProcessName());
-                            os.write(process.getAuthKey().getBytes(StandardCharsets.US_ASCII));
+                            // TODO Do we really need to send this?  The HostController generates new tokens anyway
```
fjuma, line 343: Do we need any follow up issues created for any of the TODOs in this PR?

server/src/main/java/org/jboss/as/server/security/sasl/DomainServerSaslClient.java
```diff
@@ -0,0 +1,127 @@
...
+        } catch (SaslException e) {
+            throw e;
+        } catch (IOException | UnsupportedCallbackException e) {
+            throw saslPlain.mechCallbackHandlerFailedForUnknownReason(e).toSaslException();
+        }
+        final String name = nameCallback.getName();
+        if (name == null) {
+            throw saslPlain.mechNoLoginNameGiven().toSaslException();
+        }
+        final String token = credentialCallback.getCredential(DomainServerCredential.class).getToken();
```
fjuma, line 81: Do we need a null check in case `getCredential` returns `null`?

host-controller/src/main/java/org/jboss/as/host/controller/ServerInventoryImpl.java
darranl: +1 it started as Server Inventory Token so I could work out where these tokens were flying around, then became Server Authentication Token - it is not strickly needed but when you see the token it is reassuring it is the intended one.

process-controller/src/main/java/org/jboss/as/process/ProcessController.java
darranl: @fjuma I am not sure it feels like one of this issues that if we create it we will never come back to it - this was more acknowledging this is redundant but I think it has been redundant for a decade !

yersan (approved):
I haven't found any other details, just minor comments that should not prevent merging this, so I adding my LGTM

darranl:
@fjuma if you would not mind having another look I have corrected the null handling.

I have corrected the bits you requested @yersan but I don't think that invalidates your review.

process-controller/src/main/java/org/jboss/as/process/ProcessController.java
fjuma: Ok that's fine then :)

fjuma (approved):
LGTM. Thanks @darranl!

luck3y (approved):
It has been quite a while since I worked on this code, but LGTM :)

darranl:
Thanks @luck3y it has been even longer since I really worked on it - going back to 2021 :smile:

PR #5516, opened 2023-05-23
WFCORE-6311 & WFCORE-6347 Port service/subsystem constructs from wildfly-clustering-service and wildfly-clustering-common to wildfly-core

pferraro:
https://issues.redhat.com/browse/WFCORE-6311
https://issues.redhat.com/browse/WFCORE-6347

This is largely a direct port, but with some refinements (detailed below), renamed interfaces, and some minimal integration with wildfly-controller.

To summarize:

* org.wildfly:wildfly-clustering-service ported to org.wildfly.core:wildfly-service
  * ServiceConfigurator abstraction generalized as ServiceInstaller and structured to accommodate a CapabilityServiceInstaller specialization (see below)
    * Replace generic ServiceConfigurator/Service implementations with ServiceInstaller factory methods
  * Replace Dependency/DependencySupplier implementations with interface factory methods
  * Rename Requirement -> ServiceDescriptor, including subtypes
  * Integrate ServiceDescriptor into wildfly-controller via RuntimeCapability factory methods and CapabilityServiceBuilder.require(...) method variants.
  * Demonstrate ServiceDescriptor usage using org.wildfly.core:wildfly-network capabilities and path capabilitites from org.wildfly.core:wildfly-server
* Majority of org.wildfly:wildfly-clustering-common ported to org.wildfly.core:wildfly-subsystem
  * Create CapabilityServiceInstaller specialization of Installer to support service installation via CapabilityServiceTarget.
  * Add Dependency specializations for use with CapabilityServiceBuilder 
  * Generalized ResourceServiceConfigurator to accommodate ServiceTarget or CapabilityServiceTarget-based service installation
  * Add builder pattern to ResourceDescriptor
  * Replace ResourceServiceHandler implementation classes with factory methods.
  * Misc renaming, e.g. ComplexResource -> DynamicResource, ResourceServiceHandler -> ResourceOperationRuntimeHandler, etc.
* Migrated discovery subsystem to use org.wildfly.core:wildfly-subsystem to demonstrate how this all works.

Rebased on top of https://github.com/wildfly/wildfly-core/pull/5563


pferraro:
*sigh* This is why excluding transitive dependencies within project poms is a bad idea...  I'll submit a fix to wildfly-full.

service/src/main/java/org/wildfly/service/Installer.java
```diff
@@ -0,0 +1,393 @@
...
+        private final Consumer<T> stopTask;
+
+        private volatile T value;
+
+        /**
+         * Constructs a generic service that provides a mapped value from the specified factory to the specified consumer on {@link Service#start(StartContext)} and consumed by the specified destroyer on {@link Service#stop(StopContext)}.
+         * @param consumer the consumer that accepts the provided value
+         * @param mapper a function that maps the supplied value
+         * @param factory a supplier of this service's provided value
+         * @param destroyer a consumer that destroys the provided value on {@link Service#stop(StopContext)}.
```
bstansberry, line 364: This param doc doesn't match the code.

pferraro: Will fix.
FYI - I'm in the process of revising this branch to better leverage the changes introduced by WFCORE-6400.  I'll add the hold label for now.

subsystem/src/main/java/org/wildfly/subsystem/resource/operation/AddResourceOperationStepHandler.java
```diff
@@ -0,0 +1,292 @@
...
+                    OperationStepHandler writeHandler = attribute.getWriteHandler();
+                    if ((writeHandler != null) && (writeHandler.getClass() != WriteAttributeOperationStepHandler.class)) {
+                        // If attribute has custom handling, perform a separate write-attribute operation
+                        ModelNode writeAttributeOperation = Util.getWriteAttributeOperation(currentAddress, parameter.getName(), parameter.validateOperation(operation));
+                        context.addStep(writeAttributeOperation, writeHandler, OperationContext.Stage.MODEL);
+                    } else {
+                        // Auto-populate add resource operation parameters that correspond to resource attributes
+                        parameter.validateAndSet(operation, model);
+                    }
+                    parameter.validateAndSet(operation, model);
```
bstansberry, line 178: This doesn't seem right.

pferraro: Hmm.  No, it isn't.  Not sure how that snuck in there.  Will fix.

controller/src/main/java/org/jboss/as/controller/security/SecurityServiceDescriptor.java
```diff
@@ -0,0 +1,43 @@
...
+import org.wildfly.service.descriptor.UnaryServiceDescriptor;
+
+/**
+ * Enumerates security related service descriptors.
+ */
+public class SecurityServiceDescriptor {
+
+    public static final UnaryServiceDescriptor<KeyStore> KEY_STORE = UnaryServiceDescriptor.of("org.wildfly.security.key-store", KeyStore.class);
+
+    public static final UnaryServiceDescriptor<Policy> JACC_POLICY = UnaryServiceDescriptor.of("org.wildfly.security.jacc-policy", Policy.class);
```
bstansberry, line 26: I haven't looked at all the items here, but this one seems off-topic for this PR. I'd want to get an ok from @darranl and @fjuma before moving anything into controller that isn't provided by or used by the server kernel.

Assume the same comment applies to any of these that are for capabilities not provided or used by the kernel.

Scrolling further I realize I want an approval from one of them for this particular commit regardless.

To clarify my general concern: if subsystem X needs a service value of some type found in Java SE, it doesn't care where it comes from, and if it resolves that value using a String capability name it doesn't need to care where it comes from. If X starts relying on the kernel to tell it how to resolve it (via a constant here) now there is a coupling that only exists for the convenience of provided by that constant. The controller module starts to become a kind of registry of capabilities provided by different subsystems.

bstansberry (commented):
I started going through yesterday and have finished now. My questions/comments are below.

Also, a couple general points:

1) I noticed at least one case where a new file had an old-style license header (Copyright Red Hat etc, but it was ASL.) So please do a scan for that and correct as needed.

2) It would be good for some of us to get together and do a project to add class javadoc to the stuff in the service and subsystem modules. Not you; you could just be a resource for answering questions and code review.

You have appropriate class javadoc here; thank you for that; this isn't a criticism. But one weakness of functional programming styles is type and method names tend to lose descriptive value, so it can quickly get confusing what the purpose of things is. One mitigation I can think of is to add to the class level javadoc use-case focused content. How something is expected to be used, why it solves the use case the way it does, etc.

Probably similar for some methods.

My impression from your PR description is you've renamed things from what was in clustering; my sense is these names are better.

subsystem/src/main/java/org/wildfly/subsystem/resource/operation/AddResourceOperationStepHandler.java
pferraro: Fixed.

subsystem/src/main/java/org/wildfly/subsystem/resource/operation/WriteAttributeTranslationOperationStepHandlerHandler.java
```diff
@@ -0,0 +1,50 @@
...
+
+    private final AttributeTranslation translation;
+
+    public WriteAttributeTranslationOperationStepHandlerHandler(AttributeTranslation translation) {
+        this.translation = translation;
+    }
+
+    @Override
+    public void execute(OperationContext context, ModelNode operation) throws OperationFailedException {
+        ModelNode value = context.resolveExpressions(operation.get(ModelDescriptionConstants.VALUE));
```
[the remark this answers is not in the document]
pferraro: Good point.  For expressions, we ought to skip any attempt translate the value, i.e pass-through.

subsystem/src/main/java/org/wildfly/subsystem/service/capture/ResourceServiceValueExecutorRegistry.java
```diff
@@ -0,0 +1,50 @@
...
+ */
+package org.wildfly.subsystem.service.capture;
+
+import java.util.function.Consumer;
+
+import org.wildfly.subsystem.service.ServiceDependency;
+import org.wildfly.subsystem.service.ServiceInstaller;
+
+/**
+ * A registry of captured service value executors.
```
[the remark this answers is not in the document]
pferraro: Correct.  This commit is a port of: https://github.com/wildfly/wildfly/pull/12676

core-feature-pack/common/src/main/resources/modules/system/layers/base/org/wildfly/extension/discovery/main/module.xml
```diff
@@ -18,7 +18,7 @@
         <module name="org.jboss.staxmapper"/>
         <module name="org.jboss.as.controller"/>
         <module name="org.jboss.logging"/>
-        <module name="org.jboss.msc"/>
+        <!--module name="org.jboss.msc"/-->
```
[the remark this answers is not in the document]
pferraro: Indeed.

controller/src/main/java/org/jboss/as/controller/security/SecurityServiceDescriptor.java
pferraro: I'll extract this commit to a separate jira and PR so this can be discussed separately.

pferraro:
> I noticed at least one case where a new file had an old-style license header (Copyright Red Hat etc, but it was ASL.) So please do a scan for that and correct as needed.

Will do.

pferraro:
> It would be good for some of us to get together and do a project to add class javadoc to the stuff in the service and subsystem modules.
> Not you; you could just be a resource for answering questions and code review.

:+1:

pferraro:
> You have appropriate class javadoc here; thank you for that; this isn't a criticism. But one weakness of functional programming styles is type and method names tend to lose descriptive value, so it can quickly get confusing what the purpose of things is. One mitigation I can think of is to add to the class level javadoc use-case focused content. How something is expected to be used, why it solves the use case the way it does, etc.
> Probably similar for some methods.
> My impression from your PR description is you've renamed things from what was in clustering; my sense is these names are better.

Naming is not one of my strengths, so if you have better names in mind for any of the public interfaces/methods, let me know.

pferraro:
> > I noticed at least one case where a new file had an old-style license header (Copyright Red Hat etc, but it was ASL.) So please do a scan for that and correct as needed.
> 
> Will do.

Ah - it looks like I neglected to scan the XML files.

controller/src/main/java/org/jboss/as/controller/security/SecurityServiceDescriptor.java
pferraro: I opened https://issues.redhat.com/browse/WFCORE-6580 to track the security-related service descriptors
and https://issues.redhat.com/browse/WFCORE-6581 to track the jmx-related service descriptors
and dropped these commits from this PR.

bstansberry:
Thanks, @pferraro!

PR #5517, opened 2023-05-23
WFCORE-6365 RestartParent OperationStepHandler implementations are in compatible with service built via CapabilityServiceTarget

pferraro:
https://issues.redhat.com/browse/WFCORE-6365

### DIFF
--- a/controller/src/main/java/org/jboss/as/controller/RestartParentResourceHandlerBase.java
+++ b/controller/src/main/java/org/jboss/as/controller/RestartParentResourceHandlerBase.java
@@ -21,7 +21,7 @@
  */
 package org.jboss.as.controller;
 
-import org.jboss.as.controller.logging.ControllerLogger;
+import org.jboss.as.controller.OperationContext.Stage;
 import org.jboss.as.controller.operations.common.Util;
 import org.jboss.as.controller.registry.Resource;
 import org.jboss.dmr.ModelNode;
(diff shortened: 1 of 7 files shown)


pferraro:
Failures in MailSubsystemTestCase are due to https://github.com/wildfly/wildfly/pull/16871

controller/src/main/java/org/jboss/as/controller/RestartParentResourceHandlerBase.java
```diff
@@ -57,19 +57,24 @@ public void execute(OperationContext context, ModelNode operation) throws Operat
...
-                            recreateParentService(context, address, parentModel);
+                            // Remove/recreate parent services in separate step using an OperationContext relative to the parent resource
+                            // This is necessary to support service installation via CapabilityServiceTarget
+                            context.addStep(Util.getReadResourceOperation(address), new OperationStepHandler() {
+                                @Override
+                                public void execute(OperationContext context, ModelNode operation) throws OperationFailedException {
+                                    removeServices(context, serviceName, parentModel);
+                                    recreateParentService(context, parentModel);
+                                }
+                            }, OperationContext.Stage.RUNTIME);
```
bstansberry, line 77: @pferraro I *think* these steps should be added at the top of the queue so they execute prior to any other steps, i.e. in the same relative position as the original code executed them. I pushed up #5528 to get CI going on that while we discuss.

Note that this commit is already merged into main, so please don't update it in this branch; if any change is appropriate it needs to be via a new commit.

PR #5528, opened 2023-05-30
[WFCORE-6365] RestartParent OperationStepHandler implementations are incompatible with service built via CapabilityServiceTarget

bstansberry:
https://issues.redhat.com/browse/WFCORE-6365

This builds on #5517 but tweaks where in the step queue the added steps are inserted.


pferraro (approved):
Yeah - I think you're right - we want these extra steps to run first.
And thanks for the clarifying comment re: the use of the read-resource operation.  I realize in retrospect that this might be confusing otherwise.

bstansberry:
Thanks, @pferraro!

PR #5530, opened 2023-05-31
[WFCORE-6290] Tests if the server starts when the process UUID is invalid or removed.

dcihak:
Tests if the server starts when the process UUID is invalid or removed.
Tests if the correct warning messages are logged.

Test case for [WFCORE-6290](https://issues.redhat.com/browse/WFCORE-6290)

### DIFF
--- a/testsuite/manualmode/src/test/java/org/jboss/as/test/manualmode/logging/ProcessUuidTestCase.java
+++ b/testsuite/manualmode/src/test/java/org/jboss/as/test/manualmode/logging/ProcessUuidTestCase.java
@@ -1,0 +1,87 @@
+/*
+ * JBoss, Home of Professional Open Source.
+ * Copyright 2023 Red Hat, Inc., and individual contributors
+ * as indicated by the @author tags.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.jboss.as.test.manualmode.logging;
+
+import jakarta.inject.Inject;
+import org.apache.commons.io.FileUtils;
+import org.junit.After;
+import org.junit.Test;
+import org.junit.runner.RunWith;
+import org.wildfly.core.testrunner.ServerControl;
+import org.wildfly.core.testrunner.ServerController;
+import org.wildfly.core.testrunner.WildFlyRunner;
+
+import java.io.File;
+import java.nio.charset.StandardCharsets;
+import java.nio.file.Path;
+
+/**
+ * Tests if the server starts when the process UUID is invalid or removed.
+ * Tests if the correct warning messages are logged.
+ *
+ * @author <a href="mailto:dcihak@redhat.com">Daniel Cihak</a>
+ * @see <a href="https://issues.redhat.com/browse/WFCORE-6290">WFCORE-6290</a>
+ */
+@RunWith(WildFlyRunner.class)
+@ServerControl(manual = true)
+public class ProcessUuidTestCase extends AbstractLoggingTestCase {
+
+    @Inject
+    protected static ServerController container;
+
+    @After
+    public void stopContainer() {
+        container.stop();
+    }
+
+    /**
+     * Server should start with the invalid process UUID.
+     *
+     * @throws Exception
+     */
+    @Test
+    public void testInvalidProcessUuid() throws Exception {
+        FileUtils.write(this.getProcessUuidFile(), "INVALID_UUID", StandardCharsets.UTF_8);
+
+        container.start();
+        Path serverLog = AbstractLoggingTestCase.getAbsoluteLogFilePath("server.log");
+        AbstractLoggingTestCase.checkLogs("WFLYCTL0501", serverLog, true);
+    }
+
+    /**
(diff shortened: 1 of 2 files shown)


yersan (changes requested):
Hi @dcihak , could you please use the standard WildFly Core style format and apply it to this class? It can be found at https://github.com/wildfly/wildfly-core/tree/main/ide-configs

Also, please add a copyright header (Apache License), thanks!

testsuite/manualmode/src/test/java/org/jboss/as/test/manualmode/logging/ProcessUuidTestCase.java
```diff
@@ -0,0 +1,65 @@
+package org.jboss.as.test.manualmode.logging;
```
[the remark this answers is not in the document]
dcihak: Copyright added

testsuite/manualmode/src/test/java/org/jboss/as/test/manualmode/logging/ProcessUuidTestCase.java
```diff
@@ -0,0 +1,82 @@
...
+     *
+     * @throws Exception
+     */
+    @Test
+    public void testInvalidProcessUuid() throws Exception {
+        FileUtils.write(this.getProcessUuidFile(), "INVALID_UUID", StandardCharsets.UTF_8);
+
+        container.start();
+        Path serverLog = AbstractLoggingTestCase.getAbsoluteLogFilePath("server.log");
+        AbstractLoggingTestCase.checkLogs("An invalid UUID string 'INVALID_UUID' was found at", serverLog, true);
```
yersan, line 58: Could you replace this with the expected message id? We should not assert log messages in the tests cases.

For example for this one you can check whether the log trace contains `WFLYCTL0501`

testsuite/manualmode/src/test/java/org/jboss/as/test/manualmode/logging/ProcessUuidTestCase.java
```diff
@@ -0,0 +1,82 @@
...
+     *
+     * @throws Exception
+     */
+    @Test
+    public void testEmptyProcessUuid() throws Exception {
+        FileUtils.write(this.getProcessUuidFile(), "", StandardCharsets.UTF_8);
+
+        container.start();
+        Path serverLog = AbstractLoggingTestCase.getAbsoluteLogFilePath("server.log");
+        AbstractLoggingTestCase.checkLogs("There is no UUID string at", serverLog, true);
```
yersan, line 73: Check the message ID (`WFLYCTL0500`) for this log trace instead of the message test.

yersan (changes requested):

testsuite/manualmode/src/test/java/org/jboss/as/test/manualmode/logging/ProcessUuidTestCase.java
dcihak: Fixed

testsuite/manualmode/src/test/java/org/jboss/as/test/manualmode/logging/ProcessUuidTestCase.java
dcihak: Fixed

yersan (approved):
Thanks @dcihak

testsuite/manualmode/src/test/java/org/jboss/as/test/manualmode/logging/ProcessUuidTestCase.java
```diff
@@ -0,0 +1,82 @@
...
+     * @throws Exception
+     */
+    @Test
+    public void testInvalidProcessUuid() throws Exception {
+        FileUtils.write(this.getProcessUuidFile(), "INVALID_UUID", StandardCharsets.UTF_8);
+
+        container.start();
+        Path serverLog = AbstractLoggingTestCase.getAbsoluteLogFilePath("server.log");
+        AbstractLoggingTestCase.checkLogs("WFLYCTL0501", serverLog, true);
+        container.stop();
```
yersan, line 59: Could you add `container.stop();` in a @After method? That will ensure the container is stopped if there is an error.

yersan (changes requested):
Hi @dcihak , CI revealed this test fails when using it with a Bootable Jar profile.

Since it is not suitable for Bootable Jar, you can exclude it from the surefire configuration activated when we are using a Bootable JAR https://github.com/wildfly/wildfly-core/blob/e67bb474036a23dc8da57d5651940214230e4ee1/testsuite/manualmode/pom.xml#L606. The reason for the exclusion is it requires modification of the process-uuid file inside the Jar.

testsuite/manualmode/src/test/java/org/jboss/as/test/manualmode/logging/ProcessUuidTestCase.java
dcihak: Test excluded for the bootable jar profile, stopping the container moved into the @after method.

testsuite/manualmode/src/test/java/org/jboss/as/test/manualmode/logging/ProcessUuidTestCase.java
```diff
@@ -0,0 +1,87 @@
...
+     * @throws Exception
+     */
+    @Test
+    public void testEmptyProcessUuid() throws Exception {
+        FileUtils.write(this.getProcessUuidFile(), "", StandardCharsets.UTF_8);
+
+        container.start();
+        Path serverLog = AbstractLoggingTestCase.getAbsoluteLogFilePath("server.log");
+        AbstractLoggingTestCase.checkLogs("WFLYCTL0500", serverLog, true);
+        container.stop();
```
yersan, line 79: It is harmless, but this should have been also removed from this method since now the container is stopped in the `@After` method. We don't need to remove it with this PR and can be done after merging this.

yersan (approved):

yersan:
Thanks @dcihak

PR #5532, opened 2023-06-02
Experiment to reduce memory retention of read-only ops with large num…

bstansberry:
…bers of steps

@ehsavoie fyi


bstansberry:
Replaced by #5540

PR #5535, opened 2023-06-05
WFCORE-6369 Eliminate unnecessary collection copying in AbstractAddStepHandler constructors

pferraro:
https://issues.redhat.com/browse/WFCORE-6369

Fix improper use of Set in AbstractAddStepHandler.Parameters, since AttributeDefinition is not hashable.

### DIFF
--- a/controller/src/main/java/org/jboss/as/controller/AbstractAddStepHandler.java
+++ b/controller/src/main/java/org/jboss/as/controller/AbstractAddStepHandler.java
@@ -82,7 +82,7 @@ public class AbstractAddStepHandler implements OperationStepHandler, OperationDe
         return this.attributes;
     }
 
-    /** {@inheritDoc */
+    @Override
     public void execute(final OperationContext context, final ModelNode operation) throws OperationFailedException {
         final Resource resource = createResource(context, operation);
         populateModel(context, operation, resource);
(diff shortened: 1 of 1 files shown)


controller/src/main/java/org/jboss/as/controller/AbstractAddStepHandler.java
```diff
@@ -391,27 +380,29 @@ public Resource createResource(OperationContext context, ModelNode operation) {
     }
 
     public static class Parameters {
-        protected Set<AttributeDefinition> attributes = null;
+        protected Collection<AttributeDefinition> attributes = List.of();
```
bstansberry, line 383: @pferraro This is a breaking change, so why not go for it and make this private as well?

pferraro: Ugh - I missed that this was protected instead of private...  I did not intend for this to introduce breaking changes.
Frankly, this PR is mostly obsoleted by WFCORE-6407, which proposes to deprecate the AbstractAddStepHandler.Parameters class, as well as all non-default constructors of AbstractAddStepHandler and AbstractWriteAttributeHandler.

pferraro: Fixed.

yersan (approved):
LGTM, @bstansberry you already did a review, anything else from your side?

bstansberry (approved):
This approach changes the semantics of the protected 'attributes' field by making it immutable, but that's fine with me; as I'd have been ok with making it private.

We should follow this up by deprecating the field an adding a 'protected final Collection<AttributeDefinition> getAttributes()' javadoced as returning an immutable collection. But that doesn't need to block merging this.

yersan:
I created https://issues.redhat.com/browse/WFCORE-6419 to follow up on the last comment

yersan:
Thanks @pferraro / @bstansberry

PR #5537, opened 2023-06-06
[WFCORE-6384] When shuting down to perform an installation CLI is dis…

yersan:
…connected from DC

Jira issue: https://issues.redhat.com/browse/WFCORE-6384

Instead of comparing Paths to understand whether we are using a local CLI client instance when performing an installation, now we use a CLI client marker file. The content of this client marker file is returned by the `shutdown --perform-installation` management operation. If the content doesn't match, we are not using a local CLI instance. Otherwise, we use the most restrictive approach assuming assuming the CLI is local.

### DIFF
--- a/cli/src/main/java/org/jboss/as/cli/CommandLineMain.java
+++ b/cli/src/main/java/org/jboss/as/cli/CommandLineMain.java
@@ -21,7 +21,12 @@
 */
 package org.jboss.as.cli;
 
+import java.io.File;
+import java.io.FileWriter;
 import java.io.IOException;
+import java.nio.file.Files;
+import java.nio.file.Path;
+import java.nio.file.Paths;
 import java.security.AccessController;
 import java.security.PrivilegedAction;
 import java.util.Locale;
@@ -29,19 +34,21 @@ import java.util.Properties;
 import java.util.logging.LogManager;
 
 import org.jboss.as.cli.impl.CliLauncher;
+import org.jboss.logging.Logger;
 import org.jboss.logmanager.Configurator;
 import org.jboss.logmanager.LogContext;
 import org.jboss.logmanager.PropertyConfigurator;
-
 
 /**
 *
 * @author Alexey Loubyansky
 */
 public class CommandLineMain {
+    private static final Logger LOG = Logger.getLogger(CommandLineMain.class);
 
     public static void main(String[] args) throws Exception {
         configureLogManager(args);
+        createClientMarker();
         CliLauncher.main(args);
     }
 
@@ -150,5 +157,32 @@ public class CommandLineMain {
         properties.setProperty("formatter.PATTERN.pattern", "%d{yyyy-MM-dd HH:mm:ss,SSS} %-5p [%c] (%t) %s%e%n");
         properties.setProperty("formatter.PATTERN.properties", "pattern");
         return properties;
+    }
+
+    /**
+     * Creates a file marker under $JBOSS_HOME/bin directory. This is used by the shutdown operation to understand from where we have
+     * launched CLI instance when we are performing an update of a remote installation.
+     */
+    private static void createClientMarker() {
+        try {
+            final String jbossHome = System.getenv("JBOSS_HOME");
+            if (jbossHome != null) {
+                final Path cliMarkerPath = Paths.get(jbossHome).resolve("bin").resolve(Util.CLI_MARKER);
+                Files.deleteIfExists(cliMarkerPath);
+                Files.createFile(cliMarkerPath);
+
+                final File cliMarkerFile = cliMarkerPath.toFile();
+                cliMarkerFile.deleteOnExit();
+                final String data = String.valueOf(System.currentTimeMillis());
+                try (FileWriter writer = new FileWriter(cliMarkerFile, true)) {
+                    writer.write(data);
+                }
+                System.setProperty(Util.CLI_MARKER_VALUE, data);
+            }
+        } catch (SecurityException e) {
+            throw e;
+        } catch (Exception e) {
+            LOG.debug("Creation of cli marker failed and will be ignored.", e);
+        }
     }
 }
--- a/cli/src/main/java/org/jboss/as/cli/handlers/ShutdownHandler.java
+++ b/cli/src/main/java/org/jboss/as/cli/handlers/ShutdownHandler.java
@@ -23,7 +23,6 @@
 package org.jboss.as.cli.handlers;
 
 import java.io.IOException;
-import java.nio.file.Paths;
 import java.util.Collection;
 import java.util.concurrent.TimeUnit;
 import java.util.concurrent.atomic.AtomicReference;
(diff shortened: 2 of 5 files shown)


host-controller/src/main/java/org/jboss/as/host/controller/operations/HostShutdownHandler.java
```diff
@@ -128,6 +131,19 @@ public void execute(OperationContext context, ModelNode operation) throws Operat
             } catch (Exception e) {
                 throw HostControllerLogger.ROOT_LOGGER.noServerInstallationPrepared(productName);
             }
+
+            // check the presence of a client marker in our server installation and if so, returns its value so the client
+            // can detect whether he was launched from the same installation dir.
+            final Path cliMarker = environment.getHomeDir().toPath().resolve("bin").resolve("cli-marker");
```
jfdenise, line 137: We should test that the Files.exist

cli/src/main/java/org/jboss/as/cli/handlers/ShutdownHandler.java
```diff
@@ -172,28 +171,40 @@ protected void doHandle(CommandContext ctx) throws CommandLineException {
...
-            // This will also force the user to relaunch the CLI session using the most recent updates once the server/host has
-            // been updated.
-            boolean localClientLaunch = isLocalClientLaunch(ctx);
-            executeOperation(client, cliClient, op, !localClientLaunch);
-            if (localClientLaunch) {
+            boolean isLocalClient = true;
+            if (clientMarker != null) {
+                final String jbossHome = WildFlySecurityManager.getEnvPropertyPrivileged("JBOSS_HOME", null);
+                final String clientMarkerData = WildFlySecurityManager.getPropertyPrivileged(Util.CLI_MARKER_VALUE, null);
+                if (jbossHome != null && clientMarkerData != null) {
```
jfdenise, line 187: Not sure why we need to check for the presence of JBOSS_HOME env variable?

cli/src/main/java/org/jboss/as/cli/CommandLineMain.java
```diff
@@ -151,4 +158,29 @@ private static Properties createLogManagerConfig(final String level) {
...
+                final File cliMarkerFile = cliMarkerPath.toFile();
+                cliMarkerFile.deleteOnExit();
+                final String data = String.valueOf(System.currentTimeMillis());
+                try (FileWriter writer = new FileWriter(cliMarkerFile, true)) {
+                    writer.write(data);
+                }
+                WildFlySecurityManager.setPropertyPrivileged(Util.CLI_MARKER_VALUE, data);
+            }
+        } catch (Throwable e) {
+            // explicitly ignore this.
```
jfdenise, line 183: Could we have a warning displayed. I imagine the case where part of the JBOSS_HOME sub directories would be readonly.

host-controller/src/main/java/org/jboss/as/host/controller/operations/HostShutdownHandler.java
yersan: I think we can do it without checking the existence since we can build Path objects for unknown files since Path doesn't access to the filesystem, do you agree?

cli/src/main/java/org/jboss/as/cli/handlers/ShutdownHandler.java
yersan: Right, it was left over from a previous approach, we can remove the JBOSS_HOME check here

cli/src/main/java/org/jboss/as/cli/CommandLineMain.java
yersan: Yes, this is explicitly ignored because of the read-only use case (or any other exception).

We could log a WARN (we should move the method invocation after the one that setups the logging) but I considered it as a corner case that could go silently; if that occurs the logic will assume we are using a local CLI instance and it will get closed when the update is in place just because we won't be able to detect whether it is local.

We can add a WARN, but, shouldn't it be just noise since it is a very particular case only related to the shutdown when performing an installation? WDYT?

darranl: Or at the very least is any logging possible?  Catch blocks that drop the Throwable somehow end up being the place you need more information from when remote debugging an issue.

cli/src/main/java/org/jboss/as/cli/CommandLineMain.java
```diff
@@ -151,4 +158,29 @@ private static Properties createLogManagerConfig(final String level) {
...
         return properties;
     }
+
+    /**
+     * Creates a file marker under $JBOSS_HOME/bin directory. This is used by the shutdown operation to understand from where we have
+     * launched CLI instance when we are performing an update of a remote installation.
+     */
+    private static void createClientMarker() {
+        try {
+            final String jbossHome = WildFlySecurityManager.getEnvPropertyPrivileged("JBOSS_HOME", null);
```
darranl, line 168: I am not sure we need the SecurityManager wrappers here, in this case as this is the entry point there should not be other protection domains on the call stack.

cli/src/main/java/org/jboss/as/cli/CommandLineMain.java
```diff
@@ -151,4 +158,29 @@ private static Properties createLogManagerConfig(final String level) {
...
+            if (jbossHome != null) {
+                final Path cliMarkerPath = Paths.get(jbossHome).resolve("bin").resolve(Util.CLI_MARKER);
+                Files.deleteIfExists(cliMarkerPath);
```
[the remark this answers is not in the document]
yersan: The RFE contemplates this situation where we have stated that is the customer the one that has to ensure no other processes are launched when they are updating the server.

In this case, the marker value returned by the customer won't match the current CLI instance, so it will assume it is not a local instance and won't be automatically closed.

The user will end up in the same situation as the current Patch tool, where an update/patch can update the underlying jboss-modules.jar. Since the client CLI session is not automatically closed, the user will be using an obsolete CLI process launched from a patched jboss-modules.jar. Probably he will get errors using the CLI and he will be forced to close it and open it again.

cli/src/main/java/org/jboss/as/cli/CommandLineMain.java
darranl: I am suggesting neither, just access the property directly.

darranl: The security manager being possibly enabled is never a reason to add a doPrivileged.

When you add a doPrivileged call what you are effectively doing is dropping the protection domains of any callers currently on the call stack and setting up the permission check to be in the context of the protection domain of the class that is calling the method with the permissions check.

As this is a main method / entry point - what callers are being dropped from the call stack and what measures have been put in place to verify they can safely call this method.

The way these properties are being used the safety is probably fine but what callers are being dropped from the call stack?

darranl: > I just wanted to avoid the need to require specific permission to load an environment variable that is indeed configured by the launch script. We know that this is safe to do because it is the launch script the one that invokes this Main class.

This is not the purpose of a doPrivileged call, it is not possible to prevent the permission check.  The ProtectionDomain for the CommandLineMain class will still need the relevant permissions to be granted.

When a permission check is performed it loops through the PermissionDomains of all classes on the call stack, and each and every one of those must be granted that permission or the check will fail.  What doPrivileged does is it says ignore the classes that called me but it does not get away from the "me" class being required the permission.

If this class is always intended to be called as the main entry point to the program it does not need any doPrivileged calls as it will always be the starting point of the call stack so there will be no protection domains to drop.

On the other hand if this is expected that other classes from other protection domains will call this class then this class should only be using doPrivileged if those are trusted called / it has been designed so the doPrivileged calls can not be abused.

TBH this pattern that creeps in where if there is going to be a permission check a doPrivileged is added is one of the reasons that adds to the justification of the removal of the security manager in Java.

yersan: I'm starting to understand the tricky details of the security manager when, for good reasons, it was going to be removed :)...never is late, even if it going to be an obsolete knowledge

yersan:
@jfdenise / @darranl Could you take another look and approve if you think we can proceed as it is? Thanks!

marekkopecky:
I confirm that the original issue described in https://issues.redhat.com/browse/WFCORE-6384 is fixed by this MR! Thank you!

cli/src/main/java/org/jboss/as/cli/CommandLineMain.java
```diff
@@ -151,4 +158,31 @@ private static Properties createLogManagerConfig(final String level) {
...
         return properties;
     }
+
+    /**
+     * Creates a file marker under $JBOSS_HOME/bin directory. This is used by the shutdown operation to understand from where we have
+     * launched CLI instance when we are performing an update of a remote installation.
+     */
+    private static void createClientMarker() {
+        try {
+            final String jbossHome = System.getenv("JBOSS_HOME");
```
jfdenise, line 168: @yersan when CLI is started as a fat jar (bin/client/jboss-cli-client.jar), we don't have this env set. I am wandering if we should attempt to find the jar location (class.getProtectionDomain().getCodeSource().getLocation()) and determinate if that is inside an installation?

yersan: @jfdenise We should be fine there since Installation Manager high level commands are not available for the CLI if the CLI is not running on a modular environment.

cli/src/main/java/org/jboss/as/cli/handlers/ShutdownHandler.java
```diff
@@ -172,28 +171,39 @@ protected void doHandle(CommandContext ctx) throws CommandLineException {
...
                 try {
                     TimeUnit.SECONDS.sleep(3);
                 } catch (InterruptedException e) {
                     // Ignored
                 }
-
                 // We are using a CLI which was launched from the server installation we have requested to be updated.
                 // In order to prevent keeping using a jboss-modules.jar that could have been updated, we finish the CLI process
                 // Once the server has been restarted the user will launch again the CLI that will use the most recent updates
                 ctx.terminateSession();
```
jfdenise, line 205: Shouldn't you return after this call?

yersan: I've never found an issue, but yes, looks correct returning here.

darranl (approved):
Probably still needs review from @jfdenise but no objections from me.

jfdenise (approved):

cli/src/main/java/org/jboss/as/cli/CommandLineMain.java
```diff
@@ -21,27 +21,34 @@
...
 import org.jboss.logmanager.LogContext;
 import org.jboss.logmanager.PropertyConfigurator;
 
-
 /**
 *
 * @author Alexey Loubyansky
 */
 public class CommandLineMain {
+    private static final Logger LOG = Logger.getLogger(CommandLineMain.class);
```
jamezp, line 47: We're going to have to re-think this part and not use a static logger. This will be created before the `configureLogMangaer()` is invoked and could create issues.

jamezp: https://github.com/wildfly/wildfly-core/pull/5550

PR #5547, opened 2023-06-13
[WFCORE-6395] Workaround for WFCORE-6395

yersan:
Jira issue: https://issues.redhat.com/browse/WFCORE-6395

This is just a workaround until find a better fix for this. I haven't found yet a better way to do this, as soon as we have a ReadOnlyContext and its delegated is closed in a different thread, we could face this issue.

We have hit it when the managed servers are being registered, so we could even be more fine and only avoid the main OperationContext clean up only for those cases, but I'm not sure if there could be other cases for other management operations.


yersan:
closed, looking for another approach

PR #5552, opened 2023-06-15
[WFCORE-3792] Make the org.jboss.remoting and org.jboss.xnio modules …

bstansberry:
…private

https://issues.redhat.com/browse/WFCORE-3792


fl4via (approved):
Hi @bstansberry and @yersan ! I agree with this change. My only question is how does this affect the Undertow API from private/public pov, since it extends and uses XNIO. I hope that's not a problem and that's why I'm approving this. Please let me know otherwise.

PR #5556, opened 2023-06-16
[WFCORE-6313] fixed the package name of NativeManagementRemoveHandler…

istudens:
… used in NativeManagementResourceDefinition

Except of RBAC checking, the code is the same as the prevous one. According to local tests, both WildFly Core and WildFly test-suites pass successfully.

Issue: https://issues.redhat.com/browse/WFCORE-6313


istudens:
@bstansberry can you review this please? thanks

bstansberry (approved):

PR #3285, opened 2018-05-11
[WFCORE-3819][WFCORE-3877] added custom-security-event-listener into Elytron

hkalina:
https://issues.jboss.org/browse/WFCORE-3819 (custom audit, https://issues.jboss.org/browse/EAP7-1007, analysis: https://github.com/wildfly/wildfly-proposals/pull/68)
https://issues.jboss.org/browse/WFCORE-3877 (Configurable optional, https://issues.jboss.org/browse/EAP7-1005, analysis: https://github.com/wildfly/wildfly-proposals/pull/69)

test + doc of this: https://github.com/wildfly/wildfly/pull/11367 (https://issues.jboss.org/browse/WFLY-10611)

* added custom-security-event-listener
* reworked CustomComponentDefinition to allow passing wrapper (to allow custom-security-event-listener to implement Consumer<SecurityEvent> and not to depend on subsystem-specific SecurityEventListener)
* as requested by @darranl, removed need to implement Configurable interface requirement - method initialize(Map) sufficient
* in reworked component removed dependencies on deprecated ServiceBuilder methods

PR #5557, opened 2023-06-16
[WFCORE-6405] Make the org.wildfly.extension.elytron module private

fjuma:
https://issues.redhat.com/browse/WFCORE-6405

The `org.wildfly.extension.elytron` module was originally public as there were some classes in the Elytron subsystem that custom components needed access to in order to implement them.

There are currently still two Elytron custom components that rely on needing to implement interfaces from the Elytron subsystem:

* `custom-principal-transformer` relies on a `org.wildfly.extension.elytron.capabilities.PrincipalTransformer` implementation
* `custom-credential-security-factory` relies on a `org.wildfly.extension.elytron.capabilities.CredentialSecurityFactory` implementation

This PR updates `custom-principal-transformer` and `custom-credential-security-factory` so they can be used without needing to implement classes from the Elytron subsystem. 

Note that this follows the same approach that was used to update `custom-security-event-listener` so that it no longer relied on implementing a class from the Elytron subsystem (see https://github.com/wildfly/wildfly-core/pull/3285).

With this change, we can mark the `org.wildfly.extension.module` as private.

### DIFF
--- a/core-feature-pack/common/src/main/resources/modules/system/layers/base/org/wildfly/extension/elytron/main/module.xml
+++ b/core-feature-pack/common/src/main/resources/modules/system/layers/base/org/wildfly/extension/elytron/main/module.xml
@@ -23,6 +23,9 @@
   -->
 
 <module xmlns="urn:jboss:module:1.9" name="org.wildfly.extension.elytron">
+    <properties>
+        <property name="jboss.api" value="private"/>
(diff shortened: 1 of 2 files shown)


bstansberry (approved):

bstansberry:
Thanks @fjuma

PR #5561, opened 2023-06-22
[WFCORE-5279] Split Elytron SSL into its own subsystem and layer

cam-rod:
JIRA: https://issues.redhat.com/browse/WFCORE-5279

Some future commits may be needed for the Elytron TLS subsystem (ex. missing TLS classes, switching to an `ssl-base` artifact from Elytron), but the major changes are included in this PR. It splits the Elytron subsystem into two modules, ~~under a new parent `wildfly-core-elytron-parent`~~ _(deleted as per [this comment](https://github.com/wildfly/wildfly-core/pull/5561#discussion_r1275324957))_:

- `wildfly-elytron-integration-common` **(`/elytron-common`)**: A base folder with components needed to make an Elytron-compatible subsystem. Currently contains components needed for TLS connections. Uses the new Java package `org.wildfly.extension.elytron.common`
- `wildfly-elytron-integration` **(`/elytron`)**: The implementation that provides the Elytron subsystem. Contains some wrapper functions to minimize the number of class name changes. Keeps the original Java package `org.wildfly.extension.elytron`

Commits:
 
1. Cherry-pick from #5619
2. Creates the _common_ and _subsystem_ modules ~~under the `elytron/` folder~~; keeps all code in the subsystem, and adds a placeholder dependency on _common_
3. ~~Creates the `wildfly-core-elytron-parent` module, and move _common_ to the parent's `<dependencyManagement>`~~
4. Moves all needed classes to the _common_ module, and adds `ElytronCommonDefinitions`, a collection of methods that vary depending on subsystem name (ex. `ElytronDefinition.getRequirements()` and `ElytronExtension.getResourceDescriptionResolver()`).
    - Classes that calls these methods, or are otherwise extended by the subsystem, get the prefix `ElytronCommon`. The matching calls in the subsystem are renamed in this commit, and then reversed in the next one.
5.  Wrappers are created in _subsystem_ to replace the moved methods and maintain the structure of the calls. Methods in both modules are updated to set variables based on the subsystem name (ex. resources from different subsystems will be registered to different packages in LocalDescriptions.properties)
    - Methods/fields not needed by _common_ (ex. creating security realms, log messages, capabilities) are also moved back to _subsystem_
6. The _common_ module is renamed to the new ~~`org.wildfly.extension.elytron_common`~~ `org.wildfly.extension.elytron.common` package.
7. A JBoss Module `org.wildfly.extension.elytron-common` is added, and `org.wildfly.extension.elytron` lists it as a dependent module.
8. Create the dependency on the Elytron changes - `wildfly-elytron-ssl-base` from wildfly-security/wildfly-elytron#1729


cam-rod:
From the test failure:

```
[INFO] -------------------------------------------------------------
[ERROR] COMPILATION ERROR :
[INFO] -------------------------------------------------------------
[ERROR] /opt/buildAgent/work/e8e0dd9c7c4ba60/full/testsuite/integration/elytron/src/test/java/org/wildfly/test/integration/elytron/securityapi/TestCustomPrincipalTransformer.java:[20,50] package org.wildfly.extension.elytron.capabilities does not exist
[ERROR] /opt/buildAgent/work/e8e0dd9c7c4ba60/full/testsuite/integration/elytron/src/test/java/org/wildfly/test/integration/elytron/securityapi/TestCustomPrincipalTransformer.java:[27,56] cannot find symbol
  symbol: class PrincipalTransformer
[ERROR] /opt/buildAgent/work/e8e0dd9c7c4ba60/full/testsuite/integration/elytron/src/test/java/org/wildfly/test/integration/elytron/securityapi/TestCustomPrincipalTransformer.java:[29,5] method does not override or implement a method from a supertype
[INFO] 3 errors
[INFO] -------------------------------------------------------------
Compiler
  /opt/buildAgent/work/e8e0dd9c7c4ba60/full/testsuite/integration/elytron/src/test/java/org/wildfly/test/integration/elytron/securityapi/TestCustomPrincipalTransformer.java:[20,50] package org.wildfly.extension.elytron.capabilities does not exist
    /opt/buildAgent/work/e8e0dd9c7c4ba60/full/testsuite/integration/elytron/src/test/java/org/wildfly/test/integration/elytron/securityapi/TestCustomPrincipalTransformer.java:[27,56] cannot find symbol
      symbol: class PrincipalTransformer
    /opt/buildAgent/work/e8e0dd9c7c4ba60/full/testsuite/integration/elytron/src/test/java/org/wildfly/test/integration/elytron/securityapi/TestCustomPrincipalTransformer.java:[29,5] method does not override or implement a method from a supertype
```

@fjuma The module split moved the `PrincipalTransformer` interface to the `org.wildfly.extension.elytron.common` package, so the Java import has changed. I could create a wrapper in the `subsystem/` folder to handle this case, but I'm worried that it may lead to issues down the line, since `elytron.PrincipalTransformer` and `elytron_common.PrincipalTransformer` would be different classes. I don't really see another way around it, since the interface is needed by `SSLDefinitions`.

I can try this if you think it makes sense, but otherwise I'll modify the imports in the WildFly server repo.

elytron/common/src/main/java/org/wildfly/extension/elytron/CertificateAuthorityDefinition.java
```diff
@@ -92,15 +90,23 @@ public void validateParameter(String parameterName, ModelNode value) throws Oper
     }
 
     private static final AttributeDefinition[] ATTRIBUTES = new AttributeDefinition[]{URL, STAGING_URL};
-    private static final AbstractAddStepHandler ADD = new CertificateAuthorityAddHandler();
+    private static final ElytronReloadRequiredWriteAttributeHandler WRITE = new ElytronReloadRequiredWriteAttributeHandler(ATTRIBUTES);
 
-    private static final OperationStepHandler REMOVE = new TrivialCapabilityServiceRemoveHandler(ADD, CERTIFICATE_AUTHORITY_RUNTIME_CAPABILITY);
-    private static final AbstractWriteAttributeHandler WRITE = new ElytronReloadRequiredWriteAttributeHandler(ATTRIBUTES);
+    static CertificateAuthorityDefinition configure(final String subsystemName, final Class<?> extensionClass) {
```
fjuma: Since the `ElytronExtension` class contains the `SUBSYSTEM_NAME`, just wondering if we can make use of that directly instead of passing the subsystem name as a parameter here and in other places. Perhaps we could introduce a common Elytron extension class with a `getSubsystemName()` method and then the Elytron and Elytron TLS extensions could extend the common class and implement the `Extension` interface?

elytron/common/pom.xml
```diff
@@ -0,0 +1,31 @@
...
+ -->
+<project xmlns="http://maven.apache.org/POM/4.0.0"
+         xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+         xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/xsd/maven-4.0.0.xsd">
+    <modelVersion>4.0.0</modelVersion>
+
+    <parent>
+        <groupId>org.wildfly.core</groupId>
+        <artifactId>wildfly-core-parent</artifactId>
+        <version>21.0.0.Beta5-SNAPSHOT</version>
```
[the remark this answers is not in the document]
cam-rod: yep, I'll rebase a bit before the changes are merged

elytron/common/pom.xml
```diff
@@ -0,0 +1,31 @@
...
+    </parent>
+
+    <artifactId>wildfly-elytron-common</artifactId>
```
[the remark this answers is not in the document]
cam-rod: yeah that makes sense, it would make it easier to draw the connection to the subsystem. I'll update the name

elytron/common/src/main/java/org/wildfly/extension/elytron/CertificateAuthorityDefinition.java
cam-rod: Yeah this might work. Right now the methods that were pulled from ElytronExtension are also in `ElytronCommonDefinitions`, so I can try moving it to a new `ElytronCommonExtension`

cam-rod: Did some testing, the main issue is that since the method `ElytronCommonExtension.getSubsystemName()` would be abstract, it can't be static, which also applies to anything using the subsystem name (like `configure()`). I can't think of another straightforward way to require that the `getSubsystemName()` method be overridden.

elytron/common/src/main/java/org/wildfly/extension/elytron_common/AcmeAccountService.java
```diff
@@ -16,9 +16,9 @@
  * limitations under the License.
  */
 
-package org.wildfly.extension.elytron;
+package org.wildfly.extension.elytron_common;
```
fjuma, line 19: To be consistent with other similar package names, just wondering if this should be:

`org.wildfly.extension.elytron.common`

core-feature-pack/common/src/main/resources/modules/system/layers/base/org/wildfly/extension/elytron-common/main/module.xml
```diff
@@ -0,0 +1,37 @@
...
+ ~
+ ~ Unless required by applicable law or agreed to in writing, software
+ ~ distributed under the License is distributed on an "AS IS" BASIS,
+ ~ WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ ~ See the License for the specific language governing permissions and
+ ~ limitations under the License.
+ -->
+
+<module xmlns="urn:jboss:module:1.9" name="org.wildfly.extension.elytron-common">
+
```
fjuma, line 19: We should mark this as private API (e.g., see https://github.com/wildfly/wildfly-core/blob/main/core-feature-pack/common/src/main/resources/modules/system/layers/base/org/wildfly/extension/elytron/main/module.xml#L26-L27).

core-feature-pack/common/src/main/resources/modules/system/layers/base/org/wildfly/extension/elytron/main/module.xml
```diff
@@ -31,7 +31,6 @@
         <exclude path="org/wildfly/extension/elytron/ElytronExtension"/>
         <exclude path="org/wildfly/extension/elytron/ElytronSubsystemTransformers"/>
         <exclude path="org/wildfly/extension/elytron/_private"/>
-        <exclude path="org/wildfly/extension/elytron/capabilities/_private"/>
```
fjuma, line 34: Just to check, why was this removed?

fjuma: Ah, probably because that doesn't exist anymore.

elytron/subsystem/src/main/java/org/wildfly/extension/elytron/ElytronSubsystemParser17_0.java
```diff
@@ -19,8 +19,8 @@
 package org.wildfly.extension.elytron;
 
 import static org.jboss.as.controller.PersistentResourceXMLDescription.decorator;
-import static org.wildfly.extension.elytron.ElytronCommonConstants.SECURITY_DOMAIN;
```
fjuma: Just a note that we'll need to make sure we pull in the latest updates from Core since there's now also an 18 version of the parser.

elytron/common/src/main/java/org/wildfly/extension/elytron_common/ElytronCommonDefinitions.java
```diff
@@ -0,0 +1,121 @@
...
+    }
+
+    /* Extension methods */
+
+    public static StandardResourceDescriptionResolver getResourceDescriptionResolver(final String subsystemName, Class<?> extensionClass,
+                                                                              final String... keyPrefixes) {
+
+        String resourceName = extensionClass.getPackage().getName() + ".LocalDescriptions";
+
+        StringBuilder sb = new StringBuilder(subsystemName);
```
fjuma, line 98: It seems like the subsystem name could be derived from `extensionClass.getPackage().getName()`. That might be an option that would prevent the need to pass the subsystem name everywhere.

elytron/subsystem/src/main/java/org/wildfly/extension/elytron/TlsParser.java
```diff
@@ -58,6 +58,13 @@
...
+import org.wildfly.extension.elytron_common.CertificateAuthorityAccountDefinition;
+import org.wildfly.extension.elytron_common.CertificateAuthorityDefinition;
+import org.wildfly.extension.elytron_common.FileAttributeDefinitions;
+import org.wildfly.extension.elytron_common.FilteringKeyStoreDefinition;
+import org.wildfly.extension.elytron_common.KeyStoreDefinition;
+import org.wildfly.extension.elytron_common.LdapKeyStoreDefinition;
+import org.wildfly.extension.elytron_common.SSLDefinitions;
 
 /**
  * A parser for the TLS related definitions.
```
Skyllarr, line 70: @cam-rod The simplified TLS subsystem configuration should be placed to the `elytron.common` module in the future?

.gitleaks.toml
```diff
@@ -4,7 +4,7 @@
   description = "Allowlist"
   # ignore secrets present in the below test folders
   paths = [
-	'''elytron\/src\/test\/resources\/org\/wildfly\/extension\/elytron''',
+	'''elytron\/subsystem\/src\/test\/resources\/org\/wildfly\/extension\/elytron''',
```
Skyllarr, line 7: @cam-rod  I think we can add a `test` folder of the `elytron_common` here as well so it does not notify people in the future

elytron/common/src/main/java/org/wildfly/extension/elytron_common/CertificateAuthorityDefinition.java
```diff
@@ -139,13 +148,17 @@ protected TrivialService.ValueSupplier<CertificateAuthority> getValueSupplier(Op
...
                 String certificateAuthorityResourceName = context.getCurrentAddress().getLastElement().getValue();
                 return new CertificateAuthority(certificateAuthorityResourceName,
-                        model.get(ElytronDescriptionConstants.URL).asString(),
-                        model.get(ElytronDescriptionConstants.STAGING_URL).asStringOrNull());
+                        model.get(ElytronCommonConstants.URL).asString(),
+                        model.get(ElytronCommonConstants.STAGING_URL).asStringOrNull());
             };
         }
+
+        public String getSubsystemName() {
```
Skyllarr, line 156: Just a minor, this is currently missing Override annotation

elytron/common/src/main/java/org/wildfly/extension/elytron_common/CertificateAuthorityAccountDefinition.java
```diff
@@ -203,30 +206,45 @@ protected void performRuntime(OperationContext context, ModelNode operation, Res
...
         }
 
         @Override
         protected void rollbackRuntime(OperationContext context, final ModelNode operation, final Resource resource) {
             rollbackCredentialStoreUpdate(CREDENTIAL_REFERENCE, context, resource);
         }
-    }
 
-    private static final OperationStepHandler REMOVE = new TrivialCapabilityServiceRemoveHandler(ADD, CERTIFICATE_AUTHORITY_ACCOUNT_RUNTIME_CAPABILITY);
+        public String getSubsystemName() {
```
Skyllarr, line 221: Just a minor, this is currently missing Override annotation

elytron/common/src/main/java/org/wildfly/extension/elytron_common/AcmeAccountService.java
cam-rod: Yeah that should be fine. I think there used to be issues with having the paths nested, but those shouldn't be present anymore.

core-feature-pack/common/src/main/resources/modules/system/layers/base/org/wildfly/extension/elytron/main/module.xml
fjuma: This is the `elytron` module.

pom.xml
```diff
@@ -1501,6 +1507,17 @@
...
+            <dependency>
+                <groupId>org.wildfly.core</groupId>
+                <artifactId>wildfly-elytron-integration-common</artifactId>
+                <version>${project.version}</version>
+            </dependency>
+            <dependency>
+                <groupId>org.wildfly.core</groupId>
+                <artifactId>wildfly-elytron-integration-common</artifactId>
+                <version>${project.version}</version>
+                <type>test-jar</type>
```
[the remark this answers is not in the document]
cam-rod: oh yeah, moved

core-feature-pack/common/src/main/resources/modules/system/layers/base/org/wildfly/extension/elytron/main/module.xml
cam-rod: Yeah, I'm a bit confused by what you mean. Those files were [moved to elytron-common](https://github.com/cam-rod/wildfly-core/tree/27350b0f43fbb9efa590769e042048a2fe48c191/elytron/common/src/main/java/org/wildfly/extension/elytron_common/capabilities/_private), so I removed the exclusion from ` elytron`.

elytron/common/src/main/java/org/wildfly/extension/elytron_common/CertificateAuthorityDefinition.java
cam-rod: added

core-feature-pack/common/src/main/resources/modules/system/layers/base/org/wildfly/extension/elytron/main/module.xml
fjuma: Ah, I think when I was looking before, it looked like things were moved from `src/main/java/org/wildfly/extension/elytron/capabilities/_private` to `src/main/java/org/wildfly/extension/elytron` so was wondering why the corresponding path in the `elytron` module wasn't updated but perhaps I missed that things were actually moved to `elytron-common` instead?

cam-rod: I did move around some of those folders after the first review, so it might have changed a bit

elytron-common/pom.xml
```diff
@@ -28,4 +28,165 @@
...
+                <type>pom</type>
+                <scope>import</scope>
+            </dependency>
+        </dependencies>
+    </dependencyManagement>
+
+    <dependencies>
+        <dependency>
+            <groupId>jakarta.authentication</groupId>
+            <artifactId>jakarta.authentication-api</artifactId>
```
fjuma, line 47: Just to check, where is this needed?

elytron-common/pom.xml
```diff
@@ -28,4 +28,165 @@
...
+
+        <dependency>
+            <groupId>org.jboss.msc</groupId>
+            <artifactId>jboss-msc</artifactId>
+            <scope>provided</scope>
+        </dependency>
+
+        <dependency>
+            <groupId>org.wildfly.security</groupId>
+            <artifactId>wildfly-elytron-realm-ldap</artifactId>
```
fjuma, line 72: Is this needed for elytron-common?

fjuma: Would be godo to check each dependency here to confirm it's needed.

elytron-common/src/main/java/org/wildfly/extension/elytron/KeyStoreDefinition.java
```diff
@@ -226,16 +238,36 @@ protected void performRuntime(ModelNode result, ModelNode operation, KeyStoreSer
...
         super.registerOperations(resourceRegistration);
-        resourceRegistration.registerOperationHandler(LOAD, PersistanceHandler.INSTANCE);
+        resourceRegistration.registerOperationHandler(loadOperation, PersistanceHandler.INSTANCE);
         if (isServerOrHostController(resourceRegistration)) {
-            resourceRegistration.registerOperationHandler(STORE, PersistanceHandler.INSTANCE);
+            resourceRegistration.registerOperationHandler(storeOperation, PersistanceHandler.INSTANCE);
+        }
+    }
+
+    private static class KeyStoreOperations {
```
fjuma: Why is this class needed?

elytron-common/pom.xml
cam-rod: It's only used by [AbstractElytronSubsystemBaseTest](https://github.com/cam-rod/wildfly-core/blob/194a812265cf11165c061572121f2c9fc631d392/elytron-common/src/test/java/org/wildfly/extension/elytron/common/AbstractElytronSubsystemBaseTest.java#L19) for the `DEFAULT_SECURITY_FACTORY_PROPERTY`. I'll relabel it as a test dependency.

elytron-common/pom.xml
cam-rod: I'll take a second look. The dependencies I included were a subset of the original subsystem, so I might've reduced the number of dependencies by pulling out the wrong ones.

elytron-common/src/main/java/org/wildfly/extension/elytron/KeyStoreDefinition.java
cam-rod: I think I added it at some point to deal with static method issues I was running into. It's not needed anymore so I'll remove it

elytron-common/src/main/java/org/wildfly/extension/elytron/common/AcmeAccountService.java
```diff
@@ -55,7 +55,7 @@ class AcmeAccountService implements Service<AcmeAccount> {
     private final String keyStoreName;
     private volatile AcmeAccount acmeAccount;
 
-    AcmeAccountService(String certificateAuthorityName, List<String> contactUrlsList, String alias, String keyStoreName) {
+    public AcmeAccountService(String certificateAuthorityName, List<String> contactUrlsList, String alias, String keyStoreName) {
```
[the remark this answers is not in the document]
cam-rod: I don't think so, I probably changed the access level before resolving the AcmeClientSpi issues

elytron-common/pom.xml
cam-rod: Working through the Elytron dependencies, I've updated the list to this (pushed changes):

- `wildfly-elytron-auth-server`
- `wildfly-elytron-credential-store`
- `wildfly-elytron-keystore`
- `wildfly-elytron-ssl` (ideally will replace with `wildfly-elytron-ssl-common`)
- `wildfly-elytron-x500-cert-acme`
- with `org.wildfly.core:wildfly-server`

and then the transitive dependencies from this set:

- `wildfly-elytron-base` (from any module on the list)
- `wildfly-elytron-credential` (from `credential-store`)
- `wildfly-elytron-provider-util` (from `keystore`)
- `wildfly-elytron-security-manager` (from `wildfly-server`)
- `wildfly-elytron-x500-cert` and `wildfly-elytron-x500` (both from `x500-cert-acme`)

There are two main blockers to removing `wildfly-elytron-auth-server`:

1. `SecurityDomain` and `RealmMapper` are values that can be injected into a [server-ssl-context](https://github.com/cam-rod/wildfly-core/blob/194a812265cf11165c061572121f2c9fc631d392/elytron-common/src/main/java/org/wildfly/extension/elytron/common/SSLDefinitions.java#L1292-L1298). This one isn't too bad, I can move the implementation of the `TrivialAddHandler` outside of the common module.
2. The `CredentialSource` interface itself is part of the module. It ties back to the empty object [IdentityCredentials.NONE](https://github.com/wildfly-security/wildfly-elytron/blob/ab0080b71564640a889860f33ae462372e9e5d56/auth/server/base/src/main/java/org/wildfly/security/auth/server/IdentityCredentials.java#L404) (which is also directly used by [CredentialStoreUtility](https://github.com/cam-rod/wildfly-core/blob/194a812265cf11165c061572121f2c9fc631d392/elytron-common/src/test/java/org/wildfly/extension/elytron/common/CredentialStoreUtility.java#L92)). I don't see any obvious reason either of those classes need to be part of `auth-server`, but they both have to be moved to take out the dependency.

cam-rod: Made similar changes in `wildfly-elytron-integration`, based on dependencies that aren't listed in the `wildfly-elytron-integration-common` POM. Listed Elytron dependencies:

- `wildfly-elytron-audit`
- `wildfly-elytron-auth-server` (since eventually we want to remove this) and `wildfly-elytron-auth-server-deprecated`
- `wildfly-elytron-auth-util`
- `wildfly-elytron-client`
- `wildfly-elytron-http-util`
- `wildfly-elytron-realm-ldap`, `-realm-jdbc`, and `-realm-token`
- `wildfly-elytron-sasl`
- `wildfly-elytron-ssl`

And transitive deps:

- `wildfly-elytron-asn1`, `-x500` and `-x500-cert` (from `-x500-cert-acme` via **common**)
- `wildfly-elytron-encryption` (from `-credential-store` via **common**)
- `wildfly-elytron-auth`,  `-base`, `-credential`, `-permission`, and `-util` (from `-auth-server-common`)
- `wildfly-elytron-mechanism-gssapi` and `-security-manager-action` (from `-auth-util`)
- `wildfly-elytron-http` (from `-http-util`)
- `wildfly-elytron-provider-util` (from `-client`)
- `wildfly-elytron-realm` (from any of the `-realm-*`)

cam-rod:
Added a commit creating the dependency of `elytron-common` on `org.wildfly.security:wildfly-elytron-ssl-base`. The PR now needs to be built on top of wildfly-security/wildfly-elytron#1729

elytron-common/src/main/java/org/wildfly/extension/elytron/common/ElytronCommonCapabilities.java
```diff
@@ -0,0 +1,118 @@
...
+            .Builder.of(CERTIFICATE_AUTHORITY_ACCOUNT_CAPABILITY, true, AcmeAccount.class)
+            .build();
+
+    public static final String CERTIFICATE_AUTHORITY_CAPABILITY = CAPABILITY_BASE + "certificate-authority";
+
+    public static final RuntimeCapability<Void> CERTIFICATE_AUTHORITY_RUNTIME_CAPABILITY =  RuntimeCapability
+            .Builder.of(CERTIFICATE_AUTHORITY_CAPABILITY, true, CertificateAuthority.class)
+            .build();
+
+    public static final String ELYTRON_CAPABILITY = CAPABILITY_BASE + "elytron";
```
cam-rod, line 61: @fjuma As I was working on the TLS subsystem, I went over the uses of this capability again, and I don't think it should be in `elytron-common`. There are a few mentions in the basic add and remove handlers (ex. [ElytronCommonBaseAddHandler](https://github.com/cam-rod/wildfly-core/blob/782f37e11f8335e2b4c9813d5db1cfe6e0e764aa/elytron-common/src/main/java/org/wildfly/extension/elytron/common/ElytronCommonBaseAddHandler.java#L83)), but I'm thinking that instead I should be passing the capability of the implementing subsystem as an argument (ex. `ELYTRON_TLS_CAPABILITY`). Is this correct?

fjuma: @cam-rod Yes, for this particular capability, I think it makes sense to not have this in `elytron-common`.

PR #5563, opened 2023-06-25
WFCORE-6407 Deprecate non-default contructors from AbstractAddStepHandler and AbstractWriteAttributeHandler

pferraro:
https://issues.redhat.com/browse/WFCORE-6407

Essentially, this change avoids the need for subsystems to specify AttributeDefinition instances to their add resource and write-attribute operation handlers.
See jira for full rationale.
This change will require minor unit test updates to wildfly, so expect failures in full integration test runs until those fixes are available and merged.
Given that this proposed change affects every add/write-attribute operation, I wanted to make this available for review/feedback as soon as possible.

This should simplify the creation and registration of add and write-attribute operation handlers as well as further reduce our memory footprint, since many operation handler implementations can now be consolidated both within and across subsystems.

~~Requires https://github.com/wildfly/wildfly-core/pull/5644~~ Merged.

### DIFF
--- a/remoting/subsystem/src/main/java/org/jboss/as/remoting/RemotingSubsystemRootResource.java
+++ b/remoting/subsystem/src/main/java/org/jboss/as/remoting/RemotingSubsystemRootResource.java
@@ -16,7 +16,6 @@ import java.util.stream.Stream;
 import org.jboss.as.controller.AttributeDefinition;
 import org.jboss.as.controller.OperationContext;
 import org.jboss.as.controller.OperationFailedException;
-import org.jboss.as.controller.OperationStepHandler;
 import org.jboss.as.controller.PathElement;
 import org.jboss.as.controller.ReloadRequiredRemoveStepHandler;
 import org.jboss.as.controller.ReloadRequiredWriteAttributeHandler;
@@ -102,9 +101,8 @@ public class RemotingSubsystemRootResource extends SimpleResourceDefinition {
     public void registerAttributes(ManagementResourceRegistration resourceRegistration) {
         resourceRegistration.registerReadWriteAttribute(WORKER, null, new WorkerAttributeWriteHandler());
 
-        OperationStepHandler writeHandler = new ReloadRequiredWriteAttributeHandler(OPTIONS);
         for (final AttributeDefinition attribute: OPTIONS) {
-            resourceRegistration.registerReadWriteAttribute(attribute, null, writeHandler);
+            resourceRegistration.registerReadWriteAttribute(attribute, null, ReloadRequiredWriteAttributeHandler.INSTANCE);
         }
     }
 
@@ -120,10 +118,6 @@ public class RemotingSubsystemRootResource extends SimpleResourceDefinition {
     }
 
     private static class WorkerAttributeWriteHandler extends ReloadRequiredWriteAttributeHandler {
-
-        WorkerAttributeWriteHandler() {
-            super(WORKER);
-        }
 
         @Override
         protected void finishModelStage(OperationContext context, ModelNode operation, String attributeName, ModelNode newValue,
--- a/controller/src/main/java/org/jboss/as/controller/AbstractAddStepHandler.java
+++ b/controller/src/main/java/org/jboss/as/controller/AbstractAddStepHandler.java
@@ -6,6 +6,8 @@
 package org.jboss.as.controller;
 
 import org.jboss.as.controller.capability.RuntimeCapability;
+import org.jboss.as.controller.descriptions.ModelDescriptionConstants;
+import org.jboss.as.controller.registry.AttributeAccess;
 import org.jboss.as.controller.registry.ImmutableManagementResourceRegistration;
 import org.jboss.as.controller.registry.Resource;
 import org.jboss.dmr.ModelNode;
@@ -16,6 +18,7 @@ import java.util.Collection;
 import java.util.Collections;
 import java.util.Iterator;
 import java.util.List;
+import java.util.Map;
 import java.util.Set;
 
 import static org.jboss.as.controller.descriptions.ModelDescriptionConstants.ADD_INDEX;
@@ -24,23 +27,32 @@ import static org.jboss.as.controller.descriptions.ModelDescriptionConstants.ADD
  * Base class for {@link OperationStepHandler} implementations that add managed resource.
  *
  * @author John Bailey
+ * @author Paul Ferraro
  */
 public class AbstractAddStepHandler implements OperationStepHandler, OperationDescriptor {
 
+    private static final String AUTO_POPULATE_MODEL = "auto-populate-model";
+
+    /**
+     * This is only retained to support {@link #getAttributes()} and for those subclasses that reference this protected attribute directly.
+     */
+    @Deprecated(forRemoval = true) // This is referenceable by subclasses
     protected final Collection<? extends AttributeDefinition> attributes;
 
     /**
      * Constructs an add handler.
      */
-    public AbstractAddStepHandler() { //default constructor to preserve backward compatibility
+    public AbstractAddStepHandler() {
         this.attributes = List.of();
     }
 
     /**
      * Constructs an add handler
      * @param attributes attributes to use in {@link #populateModel(OperationContext, org.jboss.dmr.ModelNode, org.jboss.as.controller.registry.Resource)}
+     * @deprecated Use default constructor instead. Resource model auto-population logic relies on the {@link OperationDefinition#getParameters()} method of the operation definition with which this handler was registered.
      */
     @SuppressWarnings("unchecked")
+    @Deprecated(forRemoval = true)
     public AbstractAddStepHandler(Collection<? extends AttributeDefinition> attributes) {
         // Create defensive copy, if collection was not already immutable
         this.attributes = (attributes instanceof Set) ? Set.copyOf((Set<AttributeDefinition>) attributes) : List.copyOf(attributes);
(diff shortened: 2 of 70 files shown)


pferraro:
Added hold label until I sort out some missing domain testsuite changes.

pferraro:
Full-integration tests failures are addressed by https://github.com/wildfly/wildfly/pull/17007

yersan (commented):
@pferraro , it looks like there are compilation errors combining the WildFly counterpart with this PR:

https://ci.wildfly.org/buildConfiguration/WF_WildFlyCoreIntegrationExperiments/380109?hideProblemsFromDependencies=false&hideTestsFromDependencies=false

```
16:40:39] :	 [Step 4/5] [INFO] 100 warnings 
[16:40:39] :	 [Step 4/5] [INFO] -------------------------------------------------------------
[16:40:39] :	 [Step 4/5] [INFO] -------------------------------------------------------------
[16:40:39]W:	 [Step 4/5] [ERROR] COMPILATION ERROR : 
[16:40:39] :	 [Step 4/5] [INFO] -------------------------------------------------------------
[16:40:39]W:	 [Step 4/5] [ERROR] /opt/buildAgent/work/5732051f49542de7/full/ee/src/main/java/org/jboss/as/ee/subsystem/GlobalDirectoryResourceDefinition.java:[76,61] ReloadRequiredRemoveStepHandler() has protected access in org.jboss.as.controller.ReloadRequiredRemoveStepHandler
[16:40:39] :	 [Step 4/5] [INFO] 1 error
[16:40:39] :	 [Step 4/5] [INFO] -------------------------------------------------------------
[16:40:39] :	 [Step 4/5] [Maven Watcher] 
[16:40:39]E:	 [Step 4/5] Compiler
```

Ideally, we should have the commit squashed or at least have the Jira issue number on each commit.

remoting/subsystem/src/main/java/org/jboss/as/remoting/RemotingEndpointResource.java
```diff
@@ -54,14 +56,18 @@ class RemotingEndpointResource extends SimpleResourceDefinition {
     static {
         Map<String, AttributeDefinition> attrs = new LinkedHashMap<>();
         assert RemotingSubsystemRootResource.WORKER.getName().equals(RemotingSubsystemRootResource.WORKER.getXmlName());
-        attrs.put(RemotingSubsystemRootResource.WORKER.getName(), RemotingSubsystemRootResource.WORKER);
-        for (AttributeDefinition ad : RemotingSubsystemRootResource.OPTIONS) {
+        attrs.put(RemotingSubsystemRootResource.WORKER.getName(), alias(RemotingSubsystemRootResource.WORKER));
```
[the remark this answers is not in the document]
pferraro: > @pferraro , it looks like there are compilation errors combining the WildFly counterpart with this PR:
> 
> https://ci.wildfly.org/buildConfiguration/WF_WildFlyCoreIntegrationExperiments/380109?hideProblemsFromDependencies=false&hideTestsFromDependencies=false

Blech.  Restricting the visibility of this constructor was meant to be temporary - to detect cases that should be using the static INSTANCE instead.
The intention is to @Deprecate the public constructor.  I'll add the HOLD label while I clean this up.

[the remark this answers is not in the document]
pferraro: > Why is the alias created here and not handled directly on the `RemotingSubsystemRootResource.WORKER` with `.addFlag(AttributeAccess.Flag.ALIAS)` ?

I don't know the full history of this subsystem, but according to the existing code, the WORKER attribute of the /subsystem=remoting resource is not an alias, while all of the attributes of the /subsystem=remoting/configuration=endpoint resource are aliases referencing the parent.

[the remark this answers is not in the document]
pferraro: > Ideally, we should have the commit squashed or at least have the Jira issue number on each commit.

I don't generally add the jira issue number to a commit if the change is not directly related the commit and can be a source of confusion.

pferraro:
Rebasing against main.

pferraro:
Rebased against main for hopefully the final time.  All dependent changes on which this branch was based have been merged.
@bstansberry I think you might be the only person who can reasonably review this.

remoting/subsystem/src/main/java/org/jboss/as/remoting/RemotingSubsystemRootResource.java
```diff
@@ -227,10 +226,8 @@ private Attributes(ProcessType processType, boolean forDomain) {
                             .build();
                 }
             } else if (forDomain) {
-                this.worker = SimpleAttributeDefinitionBuilder.create(WORKER).setCapabilityReference((CapabilityReferenceRecorder) null).build();
```
bstansberry, line 230: @pferraro Fyi re https://issues.redhat.com/browse/WFCORE-6473 and https://issues.redhat.com/browse/WFCORE-6472

remoting/tests/src/test/java/org/jboss/as/remoting/RemotingLegacySubsystemTestCase.java
```diff
@@ -178,12 +178,6 @@ public void testSubsystemWithConnectorProperties() throws Exception {
         for (AttributeDefinition ad : RemotingSubsystemRootResource.LEGACY_ATTRIBUTES) {
             assertFalse(ad.getName(), subsystem.hasDefined(ad.getName()));
         }
-        ModelNode endpoint = subsystem.get(RemotingEndpointResource.ENDPOINT_PATH.getKey(), RemotingEndpointResource.ENDPOINT_PATH.getValue());
-        for (AttributeDefinition ad : RemotingEndpointResource.ATTRIBUTES.values()) {
-            ModelNode dflt = ad.getDefaultValue();
-            assertEquals(ad.getName(), dflt == null ? new ModelNode() : dflt, endpoint.require(ad.getName()));
-        }
```
bstansberry, line 185: Note to self: This needs to be removed because changing to aliases means the values no longer show up in a read-resource.

bstansberry: @pferraro is this commit absolutely required for this work?

This Flag.ALIAS addition results in the attributes not appearing in a normal read-resource, which makes the resource basically pointless, as it only exists for compatibility reasons. And it is pointless anyway, and is deprecated. But really we should have removed it in 29 Alpha1 and probably should now (and port that downstream.)

IOW if we must break it we should probably break it all the way and port that downstream. That's a bigger task though. @darranl FYI.

bstansberry:
@pferraro Other than my concern about the remoting endpoint alias thing this all LGTM. And that need not block this while I'm away, so long as there's a discussion/decision about what to do with that resource here and downstream. Basically...

1) If we're going to remove the resource, what you did here is fine; we'll just follow it up with the removal later in WF 30 work.
2) If we're going to keep the resource, but what you did here is with the alias flag required for your overall work, then fine. It's a correct change and we need to move forward.
3) If we're going to keep the resource but what you did with the alias flag is not required, then lets not do it and come back to it later.

remoting/tests/src/test/java/org/jboss/as/remoting/RemotingLegacySubsystemTestCase.java
pferraro: > @pferraro is this commit absolutely required for this work?

Yes.  The current /subsystem=remoting/configuration=endpoint resource has 2 fundamental issues:

- The runtime behavior of its attributes do not comply with their management description.
- The worker attribute definition is not consistent between domain vs server wrt capability references - which, once the correct attribute definition flags are applied, otherwise breaks galleon.

WFCORE-6407 ensures that the parameters of the OperationDefinition for the add operation of a given resource are consistent with the parameters used the OperationStepHandler handling the add operation.  Without this commit, add operation parameters as generated by https://github.com/pferraro/wildfly-core/blob/WFCORE-6407/controller/src/main/java/org/jboss/as/controller/SimpleResourceDefinition.java#L245 will not be correct, and lots of model compatibility tests will fail.

> This Flag.ALIAS addition results in the attributes not appearing in a normal read-resource, which makes the resource basically pointless, as it only exists for compatibility reasons.

Presumably, retaining compatibility is not pointless, unless this deprecation is sufficiently old.

> But really we should have removed it in 29 Alpha1 and probably should now (and port that downstream.)

If this child resource can be removed, all the better.

pferraro:
Adding hold label.  This will be rebased on WFCORE-6472/WFCORE-6473 work.

pferraro:
> @pferraro Other than my concern about the remoting endpoint alias thing this all LGTM. And that need not block this while I'm away, so long as there's a discussion/decision about what to do with that resource here and downstream. Basically...
> 
> 1. If we're going to remove the resource, what you did here is fine; we'll just follow it up with the removal later in WF 30 work.
> 2. If we're going to keep the resource, but what you did here is with the alias flag required for your overall work, then fine. It's a correct change and we need to move forward.
> 3. If we're going to keep the resource but what you did with the alias flag is not required, then lets not do it and come back to it later.

Provided that https://github.com/wildfly/wildfly-core/pull/5643 looks ok, I will rebase this branch on top of that work.

pferraro:
Rebased on top of https://github.com/wildfly/wildfly-core/pull/5644

pferraro:
Rebased against main.

bstansberry (approved):

bstansberry:
Thanks @pferraro

PR #5564, opened 2023-06-26
Next is 21.1.0.Beta1

bstansberry:
@yersan The core release for WildFly 29 Beta1 should be 21.1.0.Beta1.


yersan:
Thanks @bstansberry

PR #5574, opened 2023-07-05
Fix for WFCORE-6424, Generic command argument value issue with List containing Object

jfdenise:
Issue: https://issues.redhat.com/browse/WFCORE-6424


yersan:
Thanks @jfdenise

PR #5576, opened 2023-07-07
[WFCORE-6423] Allow reload operation to send an early response when u…

yersan:
…sing a remote proxy controller

Jira issue: https://issues.redhat.com/browse/WFCORE-6423

### DIFF
--- a/controller/src/main/java/org/jboss/as/controller/remote/TransactionalProtocolOperationHandler.java
+++ b/controller/src/main/java/org/jboss/as/controller/remote/TransactionalProtocolOperationHandler.java
@@ -23,10 +23,17 @@
 package org.jboss.as.controller.remote;
 
 import static java.security.AccessController.doPrivileged;
+import static org.jboss.as.controller.client.helpers.ClientConstants.RESULT;
 import static org.jboss.as.controller.descriptions.ModelDescriptionConstants.CANCELLED;
+import static org.jboss.as.controller.descriptions.ModelDescriptionConstants.COMPOSITE;
 import static org.jboss.as.controller.descriptions.ModelDescriptionConstants.FAILED;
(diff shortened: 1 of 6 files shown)


yersan:
@bstansberry I've added some documentation about how is the TransactionalProtocolClient implementation as suggested in the Jira. 

From my point of view, `TransactionalProtocolHandlers#executeBlocking` should be more of a TransactionalProtocolClient responsibility instead of `TransactionalProtocolHandlers`

controller/src/main/java/org/jboss/as/controller/remote/TransactionalProtocolOperationHandler.java
```diff
@@ -75,6 +88,7 @@
  * @author <a href="mailto:darran.lofthouse@jboss.com">Darran Lofthouse</a>
  */
 public class TransactionalProtocolOperationHandler implements ManagementRequestHandlerFactory {
+    private static final Set<String> PREPARED_RESPONSE_OPERATIONS = new HashSet<>(List.of(RELOAD));
```
bstansberry, line 91: This should be an immutable set.

bstansberry (changes requested):
I have one minor change suggestion.

This seems like something that should wait for WF 30, i.e. merge it next week after the core 21.1 release. I think it's fine but it doesn't seem like something so urgent that it should be added to 29 at the last moment.

controller/src/main/java/org/jboss/as/controller/remote/TransactionalProtocolOperationHandler.java
yersan: Done

yersan: @bstansberry Could you take another look?

bstansberry (approved):

yersan:
Tnanks @bstansberry

PR #5584, opened 2023-07-14
[WFCORE-6419] make the direct usage of attributes, which is a protect…

istudens:
…ed field, deprecated

Issue: https://issues.redhat.com/browse/WFCORE-6419


pferraro:
FYI, this pull request will become redundant following: https://github.com/wildfly/wildfly-core/pull/5563

istudens:
Superseded with https://github.com/wildfly/wildfly-core/pull/5563

PR #5588, opened 2023-07-20
WFCORE-6400 Update DeploymentPhaseContext to return a ServiceTarget capable of resolving capability requirements.

pferraro:
https://issues.redhat.com/browse/WFCORE-6400

N.B. The interfaces extracted from CapabilityServiceTarget/CapabilityServiceBuilder intentionally override only non-deprecated MSC ServiceTarget/ServiceBuilder methods.

Technically, the proposed modification to the DeploymentPhaseContext.getServiceTarget() return type, while source-compatible, is not binary-compatibile.
I assumed this would be an acceptable change, since the org.wildfly.core:wildfly-server module is private, and this would only affect external subsystems that perform service installation from a custom DeploymentUnitProcessor.

### DIFF
--- a/controller/src/main/java/org/jboss/as/controller/CapabilityServiceBuilder.java
+++ b/controller/src/main/java/org/jboss/as/controller/CapabilityServiceBuilder.java
@@ -19,25 +19,22 @@
 package org.jboss.as.controller;
 
 import java.util.function.Consumer;
-import java.util.function.Supplier;
 
 import org.jboss.as.controller.capability.RuntimeCapability;
 import org.jboss.msc.Service;
 import org.jboss.msc.service.LifecycleListener;
-import org.jboss.msc.service.ServiceBuilder;
 import org.jboss.msc.service.ServiceController;
 import org.jboss.msc.service.ServiceName;
 
 /**
  * A builder for an individual service in a {@code CapabilityServiceTarget}.
- * Create an instance via the {@link CapabilityServiceTarget#addCapability(RuntimeCapability)}.
- * Builder also adds support to define capability requirement via {@link #requiresCapability(String, Class, String...)}.
+ * Create an instance via the {@link CapabilityServiceTarget#addService()} method.
  *
  * @param <T> the service type
  * @author Tomaz Cerar (c) 2017 Red Hat Inc.
  * @author <a href="mailto:ropalka@redhat.com">Richard Opalka</a>
  */
-public interface CapabilityServiceBuilder<T> extends ServiceBuilder<T> {
+public interface CapabilityServiceBuilder<T> extends RequirementServiceBuilder<T> {
 
     @Override
     CapabilityServiceBuilder<T> setInitialMode(ServiceController.Mode mode);
@@ -86,15 +83,4 @@ public interface CapabilityServiceBuilder<T> extends ServiceBuilder<T> {
      * @return consumer providing value
      */
     <V> Consumer<V> provides(final RuntimeCapability<?>[] capabilities, final ServiceName[] aliases);
-
-    /**
-     * Capability requirement.
-     *
-     * @param capabilityName name of capability requirement
-     * @param dependencyType the class of the value of the dependency
-     * @param <V>            the type of the value of the dependency
-     * @param referenceNames dynamic part(s) of capability name, only useful when using dynamic named capabilities
-     * @return readonly dependency reference
-     */
-    <V> Supplier<V> requiresCapability(String capabilityName, Class<V> dependencyType, String... referenceNames);
 }
--- a/server/src/main/java/org/jboss/as/server/deployment/DeploymentUnitPhaseService.java
+++ b/server/src/main/java/org/jboss/as/server/deployment/DeploymentUnitPhaseService.java
@@ -29,13 +29,19 @@ import java.util.List;
 import java.util.ListIterator;
 import java.util.Set;
 import java.util.concurrent.atomic.AtomicBoolean;
+import java.util.function.Supplier;
 
+import org.jboss.as.controller.RequirementServiceBuilder;
+import org.jboss.as.controller.RequirementServiceTarget;
+import org.jboss.as.controller.capability.CapabilityServiceSupport;
 import org.jboss.as.server.deployment.module.ModuleSpecification;
 import org.jboss.as.server.logging.ServerLogger;
 import org.jboss.modules.ModuleIdentifier;
 import org.jboss.msc.service.LifecycleEvent;
 import org.jboss.msc.service.LifecycleListener;
+import org.jboss.msc.service.DelegatingServiceBuilder;
 import org.jboss.msc.service.DelegatingServiceRegistry;
+import org.jboss.msc.service.DelegatingServiceTarget;
 import org.jboss.msc.service.Service;
 import org.jboss.msc.service.ServiceBuilder;
 import org.jboss.msc.service.ServiceContainer;
@@ -119,7 +125,7 @@ final class DeploymentUnitPhaseService<T> implements Service<T> {
         final List<RegisteredDeploymentUnitProcessor> list = chains.getChain(phase);
         final ListIterator<RegisteredDeploymentUnitProcessor> iterator = list.listIterator();
         final ServiceContainer container = context.getController().getServiceContainer();
-        final ServiceTarget serviceTarget = context.getChildTarget().subTarget();
+        final RequirementServiceTarget serviceTarget = new DeploymentUnitServiceTarget(context.getChildTarget().subTarget(), deploymentUnit.getAttachment(Attachments.CAPABILITY_SERVICE_SUPPORT));
         final DeploymentUnit parent = deploymentUnit.getParent();
 
         final List<DeploymentUnitPhaseDependency> dependencies = new LinkedList<>();
@@ -276,5 +282,68 @@ final class DeploymentUnitPhaseService<T> implements Service<T> {
             }
         }
         return !shouldNotRun.contains(deployer.getSubsystemName());
+    }
+
+    static class DeploymentUnitServiceTarget extends DelegatingServiceTarget implements RequirementServiceTarget {
+        private final CapabilityServiceSupport support;
+
+        DeploymentUnitServiceTarget(ServiceTarget target, CapabilityServiceSupport support) {
+            super(target);
+            this.support = support;
+        }
+
+        @Override
+        public RequirementServiceTarget addListener(LifecycleListener listener) {
+            super.addListener(listener);
+            return this;
+        }
+
+        @Override
+        public RequirementServiceTarget removeListener(LifecycleListener listener) {
+            super.removeListener(listener);
+            return this;
+        }
+
+        @Override
+        public RequirementServiceTarget subTarget() {
+            return new DeploymentUnitServiceTarget(super.subTarget(), this.support);
+        }
+
+        @Override
+        public RequirementServiceBuilder<?> addService() {
+            return new DeploymentUnitServiceBuilder<>(super.addService(), this.support);
+        }
+    }
+
+    static class DeploymentUnitServiceBuilder<T> extends DelegatingServiceBuilder<T> implements RequirementServiceBuilder<T> {
+        private final CapabilityServiceSupport support;
+
+        DeploymentUnitServiceBuilder(ServiceBuilder<T> builder, CapabilityServiceSupport support) {
+            super(builder);
+            this.support = support;
+        }
+
+        @Override
+        public RequirementServiceBuilder<T> setInitialMode(Mode mode) {
+            super.setInitialMode(mode);
+            return this;
+        }
+
+        @Override
+        public RequirementServiceBuilder<T> setInstance(org.jboss.msc.Service service) {
+            super.setInstance(service);
+            return this;
+        }
+
+        @Override
+        public RequirementServiceBuilder<T> addListener(LifecycleListener listener) {
+            super.addListener(listener);
+            return this;
+        }
+
+        @Override
+        public <V> Supplier<V> requiresCapability(String capabilityName, Class<V> dependencyType, String... referenceNames) {
+            return super.requires(this.support.getCapabilityServiceName(capabilityName, referenceNames));
+        }
     }
 }
--- a/controller/src/main/java/org/jboss/as/controller/CapabilityServiceTarget.java
+++ b/controller/src/main/java/org/jboss/as/controller/CapabilityServiceTarget.java
@@ -20,6 +20,7 @@ package org.jboss.as.controller;
 
 import org.jboss.as.controller.capability.RuntimeCapability;
 import org.jboss.msc.service.LifecycleListener;
+import org.jboss.msc.service.ServiceName;
 
 /**
  * The target of ServiceBuilder for capability installations.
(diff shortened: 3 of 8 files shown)


pferraro:
LOL... My "source compatible" statement above assumed that the nobody would re-implement DeploymentUnitPhaseContext, and of course someone is doing exactly this in a unit test.  I'll add an upstream fix that changes this test to use a dynamic mock.

controller/src/main/java/org/jboss/as/controller/ServiceTarget.java
```diff
@@ -0,0 +1,38 @@
...
+
+package org.jboss.as.controller;
+
+import org.jboss.msc.service.LifecycleListener;
+
+/**
+ * {@link org.jboss.msc.service.ServiceTarget} whose builders support capability requirements.
+ * @author Paul Ferraro
+ */
+public interface ServiceTarget extends org.jboss.msc.service.ServiceTarget {
```
yersan, line 25: This interface is practically a kind of mirror of `CapabilityServiceTarget`. I'm wondering if we should try to avoid these name clashes introduced via this interface (I mean `ServiceTarget` vs `org.jboss.msc.service.ServiceTarget`), and consequently the binary incompatible issue, and try to find a solution by just adding `CapabilityServiceTarget getCapabilityServiceTarget()` in `DeploymentPhaseContext`. I don't know if that is possible or whether it was already checked, so this is just a comment.

This seems to be aligned with what `OperationContext` does. There we have deprecated the uses of `ServiceTarget getServiceTarget()` in favor of `CapabilityServiceTarget getCapabilityServiceTarget()`.

Would that make sense?

pferraro: > This interface is practically a kind of mirror of `CapabilityServiceTarget`.

No. It is a generalization of CapabilityServiceTarget.  CapabilityServiceTarget is used to create services for RuntimeCapability instances.  This new interface is used to create standard MSC services.

> I'm wondering if we should try to avoid these name clashes introduced via this interface (I mean `ServiceTarget` vs `org.jboss.msc.service.ServiceTarget`)

Why?

> ... , and consequently the binary incompatible issue.

The binary compatibility issue is not a consequence of the name clash of this interface with org.jboss.msc.service.ServiceTarget; it is a consequence of narrowing the return type of a public interface method (regardless of its name).

> ... and try to find a solution by just adding `CapabilityServiceTarget getCapabilityServiceTarget()` in `DeploymentPhaseContext`.

This would be bad interface design, since the object returned by DeploymentPhaseContext.getServiceTarget() cannot implement all of the methods of CapabilityServiceTarget.  More specifically, this object cannot create objects that implement all of the methods of CapabilityServiceBuilder, since this new object explicitly supports only standard MSC service installation.

Frankly, the question presented in the PR description needs to be answered first, i.e. are the deployment-related interfaces in org.wildfly.core:wildfly-server constrained by binary compatibility?

> This seems to be aligned with what `OperationContext` does. There we have deprecated the uses of `ServiceTarget getServiceTarget()` in favor of `CapabilityServiceTarget getCapabilityServiceTarget()`.
> Would that make sense?

If we decide that binary compatibility is a requirement for interfaces within the wildfly-server module, then, yes, adding a new method is essential.  However, we would still need the same generalization hierarchy proposed by this PR, i.e. CapabilityServiceTarget -> FooServiceTarget -> org.jboss.msc.service.ServiceTarget, and figure out a better name.

However, if we decide that binary compatibility is _not_  a requirement for the interfaces within the wildfly-server module (the case for which I am arguing), then we should avoid adding a new DeploymentPhaseContext.getFooServiceTarget() method and should prefer a source-compatible approach, such as what is being proposed by this PR.

pferraro: > In any case, these changes for sure need more eyes on it, since I'm not fully familiar with these interfaces and compositions.

Yes. I am hoping @bstansberry can provide his perspective on the binary-compatibility question -and that @ropalka can provide his MSC perspective.

yersan: > No. It is a generalization of CapabilityServiceTarget. CapabilityServiceTarget is used to create services for RuntimeCapability instances. This new interface is used to create standard MSC services.

Yes right, CapabilityServiceTarget is used to create services for RuntimeCapability, but it also extends org.jboss.msc.service.ServiceTarget. So it looked to me quite similar to your new interface.


> Why?

No technical reasons, but if the two interfaces add different things, I would think that adding them with different names could make sense. Not a strong reason. I understand that creating org.jboss.as.controller.ServiceTarget, will allow us to avoid refactoring the callers; they can continue using the parent interface org.jboss.msc.service.ServiceTarget, but I got the impression that was going to be more convoluted than standardizing the CapabilityServiceTarget usage if possible.

OperationContext deprecated the uses of org.jboss.msc.service.ServiceTarget, which also made me think we wanted to eliminate the ServiceTarget usages. However we are now adding a new ServiceTarget interface, yes an interface in a new package, but I'm more talking about what I thought there were the future intentions: I assumed CapabilityServiceTarget in favor of ServiceTarget.

pferraro: > Yes right, CapabilityServiceTarget is used to create services for RuntimeCapability, but it also extends org.jboss.msc.service.ServiceTarget. So it looked to me quite similar to your new interface.

This change makes CapabilityServiceTarget a _specialization_ of the new interface.

> No technical reasons, but if the two interfaces add different things, I would think that adding them with different names could make sense.

That's the point.  They produce the same thing, but WildFly's ServiceTarget returns a ServiceBuilder specialization that adds resolution of capability requirements.

> OperationContext deprecated the uses of org.jboss.msc.service.ServiceTarget, which also made me think we wanted to eliminate the ServiceTarget usages.

This was done because the two methods were redundant. OperationContext.getServiceTarget() and Operation.getCapabilityServiceTarget() returned the exact same object.

> However we are now adding a new ServiceTarget interface, yes an interface in a new package, but I'm more talking about what I thought there were the future intentions: I assumed CapabilityServiceTarget in favor of ServiceTarget.

AIUI, CapabilityServiceTarget is not appropriate for use within a DeploymentPhaseContext.  From what I can tell, the concept of a RuntimeCapability is only really relevant to the server management model, not to a deployment management model, where there is no such thing as deployment model validation independent of runtime.  Thus, a DeploymentUnitProcessor would have no use for the complete functionality of CapabilityServiceTarget.  @bstansberry Am I mistaken?  If so, then you are right, and we could just expose CapabilityServiceTarget from DeploymentPhaseContext.

bstansberry: @pferraro You are correct; a deployment cannot provide a capability. The deployment resource definition cannot register at server boot that the resource *could* provide a capability, and when the concrete resource instance is added the add op cannot register at Stage.MODEL the fact that the deployment provides the capability. So all the deployment could do at runtime would be to install a service under the name of a capability registered by some 'foreign' resource. Which would be wrong. Perhaps if some day we implemented the old idea of a 'system deployment' this could somehow be something such a thing does, but that's pure speculation.

Re source and binary compatibility, IMHO most packages in the wildfly-controller module and the wildfly-server module's org.jboss.as.server.deployment package tree should *try very hard* to maintain binary compatibility but are only *required* to maintain source compatibility. Except for removal of long-deprecated API. Binary compatibility breakage should be infrequent. Ideally we batch such breakage up.

Binary incompatibility in these packages forces external extensions to rebuild and release, and perhaps to branch.

pferraro: > most packages in the wildfly-controller module and the wildfly-server module's org.jboss.as.server.deployment package tree should try very hard to maintain binary compatibility

@bstansberry ok - I will try harder... :)  IOW, I'll leave the existing getServiceTarget() intact and add a separate method.

server/src/main/java/org/jboss/as/server/deployment/DeploymentUnitPhaseService.java
```diff
@@ -277,4 +281,67 @@ private static boolean shouldRun(final DeploymentUnit unit, final RegisteredDepl
...
+    static class DeploymentUnitServiceTarget extends DelegatingServiceTarget implements ServiceTarget {
+        private final CapabilityServiceSupport support;
+
+        DeploymentUnitServiceTarget(org.jboss.msc.service.ServiceTarget target, CapabilityServiceSupport support) {
+            super(target);
+            this.support = support;
+        }
+
+        @Override
+        public ServiceTarget addListener(LifecycleListener listener) {
```
ropalka, line 294: Why returning ServiceTarget instead of DeploymentUnitServiceTarget? Is it intentional?

server/src/main/java/org/jboss/as/server/deployment/DeploymentUnitPhaseService.java
```diff
@@ -277,4 +281,67 @@ private static boolean shouldRun(final DeploymentUnit unit, final RegisteredDepl
...
+    static class DeploymentUnitServiceBuilder<T> extends DelegatingServiceBuilder<T> implements ServiceBuilder<T> {
+        private final CapabilityServiceSupport support;
+
+        DeploymentUnitServiceBuilder(org.jboss.msc.service.ServiceBuilder<T> builder, CapabilityServiceSupport support) {
+            super(builder);
+            this.support = support;
+        }
+
+        @Override
+        public ServiceBuilder<T> setInitialMode(Mode mode) {
```
ropalka, line 325: Same here. Why returning ServiceBuilder instead of DeploymentUnitServiceBuilder?

ropalka (changes requested):
I reviewed this PR from JBoss MSC perspective only @pferraro .
It looks good to me, I just see minor enhancements there.
Overall it is a good solution for recent WildFly targeting EAP8.
In the future it would be great to decouple WildFly API from JBoss MSC API,
implement some kind of decorator patter on top of JBoss MSC API and hide its usages.
This implies all recently introduced classes plus others exposing JBoss MSC API should be deprecated for removal.

controller/src/main/java/org/jboss/as/controller/CapabilityServiceBuilder.java
```diff
@@ -86,15 +84,4 @@ public interface CapabilityServiceBuilder<T> extends ServiceBuilder<T> {
      * @return consumer providing value
      */
     <V> Consumer<V> provides(final RuntimeCapability<?>[] capabilities, final ServiceName[] aliases);
```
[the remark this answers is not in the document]
pferraro: The RequirementServiceBuilder interface (formerly o.j.a.c.ServiceBuilder) is incapable of providing values for a RuntimeCapability.  See response from Brian above.

server/src/main/java/org/jboss/as/server/deployment/DeploymentUnitPhaseService.java
pferraro: Because there is nothing to be gained by overloading the return type in the implementation, which is not visible to consumers.

server/src/main/java/org/jboss/as/server/deployment/DeploymentUnitPhaseService.java
pferraro: Because there is nothing to be gained by overloading the return type in the implementation, which is not visible to consumers.

controller/src/main/java/org/jboss/as/controller/ServiceTarget.java
pferraro: @yersan @bstansberry  OK - I renamed these interfaces accordingly:
org.jboss.as.controller.ServiceTarget -> RequirementServiceTarget
org.jboss.as.controller.ServiceBuilder -> RequirementServiceBuilder

I restored the original DeploymentPhaseContext.getServiceTarget() to return its original type and added a new DeploymentPhaseContext.getRequirementServiceTarget() method.
Feel free to suggest better names for these interfaces.

server/src/main/java/org/jboss/as/server/deployment/DeploymentUnitPhaseService.java
ropalka: Makes sense. Thanks for clarification!

server/src/main/java/org/jboss/as/server/deployment/DeploymentUnitPhaseService.java
ropalka: Makes sense. Thanks for clarification.

controller/src/main/java/org/jboss/as/controller/CapabilityServiceBuilder.java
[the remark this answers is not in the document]
ropalka: Aha, now I see.

ropalka (approved):

ropalka:
I'm approving this PR from JBoss MSC PoV @pferraro . From WildFly side I expect it to be approved by @bstansberry .

bstansberry:
> 
In the future it would be great to decouple WildFly API from JBoss MSC API,
implement some kind of decorator patter on top of JBoss MSC API and hide its usages.
>

Just a note for discussion clarity; I don't think you mean a decorator pattern as AIUI what's here now is a decorator pattern.

>
This implies all recently introduced classes plus others exposing JBoss MSC API should be deprecated for removal
if we will go this path.
>

If this is the direction we are going to go, should we start now? @pferraro has introduced new interfaces, but they are exposing MSC API. Which means if we later do what he suggests we'll break these new ones as well. Adding something and then immediately deprecating it is a sign you're off track. :)

There are big downsides of course, when it comes to this particular JIRA. Doing it would mean use of the single method Paul wants to introduce would require callers to switch everything to the new API. That would just be import changes to a source file the user would be editing anyway, so maybe that's not so terrible.

pferraro:
> In the future it would be great to decouple WildFly API from JBoss MSC API

For starters, there is no "WildFly API".  Subsystem currently use interfaces and classes from both wildfly-controller and wildfly-server, which exposes subsystem code, not just to the MSC API, but all matter of server internals.

> If this is the direction we are going to go, should we start now?

Not with this PR, no.

However, might I suggest using https://github.com/wildfly/wildfly-core/pull/5516 as a starting point for this discussion?

Decouple subsystem development from the MSC API was not a specific goal of that enhancement, however, if you look at the sample changes made to the discovery subsystem [1] (my usual test subject), the modified subsystem code contains no direct usage the MSC API.  This was meant to be a simple port of code from wildfly full (wildfly-clustering-common, to be specific), but I can certainly begin look into MSC decoupling as part of its goals.

[1] https://github.com/pferraro/wildfly-core/tree/WFCORE-6311/discovery/src/main/java/org/wildfly/extension/discovery

ropalka:
> > 
> 
> In the future it would be great to decouple WildFly API from JBoss MSC API, implement some kind of decorator patter on top of JBoss MSC API and hide its usages.
> 
> > 
> 
> Just a note for discussion clarity; I don't think you mean a decorator pattern as AIUI what's here now is a decorator pattern.
> 
> > 
Sorry for confusion. You're right @bstansberry . I meant WildFly subsystem API should provide a simple facade that would hide internal JBoss MSC usages (adapter pattern).
> 
> This implies all recently introduced classes plus others exposing JBoss MSC API should be deprecated for removal if we will go this path.
> 
> > 
> 
> If this is the direction we are going to go, should we start now? @pferraro has introduced new interfaces, but they are exposing MSC API. Which means if we later do what he suggests we'll break these new ones as well. Adding something and then immediately deprecating it is a sign you're off track. :)
> 
> There are big downsides of course, when it comes to this particular JIRA. Doing it would mean use of the single method Paul wants to introduce would require callers to switch everything to the new API. That would just be import changes to a source file the user would be editing anyway, so maybe that's not so terrible.

I didn't mean this PR should start rewriting this. I just wanted to say what I don't like about current status quo and what should change in near future.

bstansberry (approved):

bstansberry:
Thanks @pferraro, @yersan and @ropalka

PR #5589, opened 2023-07-24
[WFCORE-6375][WFCORE-6434] Do not synchronize stopped servers when a HC reconnects and allow reload/restart required when starting

yersan:
Jira issue: https://issues.redhat.com/browse/WFCORE-6375
Jira issue: https://issues.redhat.com/browse/WFCORE-6434

### DIFF
--- a/controller/src/main/java/org/jboss/as/controller/ControlledProcessState.java
+++ b/controller/src/main/java/org/jboss/as/controller/ControlledProcessState.java
@@ -94,6 +94,9 @@ public class ControlledProcessState {
 
     private boolean restartRequiredFlag = false;
 
+    private boolean reloadRequiredOnStarting = false;
+    private boolean restartRequiredOnStarting = false;
+
     public ControlledProcessState(final boolean reloadSupported) {
         this.reloadSupported = reloadSupported;
         service = new ControlledProcessStateService(State.STOPPED);
@@ -117,6 +120,7 @@ public class ControlledProcessState {
 
     public void setRunning() {
         AtomicStampedReference<State> stateRef = state;
+        int currentStamp = stamp.get();
         int newStamp = stamp.incrementAndGet();
         int[] receiver = new int[1];
         // Keep trying until stateRef is set with our stamp
@@ -126,8 +130,13 @@ public class ControlledProcessState {
                 break;
             }
             synchronized (service) {
-                State newState = restartRequiredFlag ? State.RESTART_REQUIRED : State.RUNNING;
-                if (state.compareAndSet(was, newState, receiver[0], newStamp)) {
+                State newState =  restartRequiredOnStarting ? State.RESTART_REQUIRED
+                        : reloadRequiredOnStarting ? State.RELOAD_REQUIRED
+                        : restartRequiredFlag ? State.RESTART_REQUIRED : State.RUNNING;
+                int stamp = restartRequiredOnStarting || reloadRequiredOnStarting ? currentStamp : newStamp;
+                if (state.compareAndSet(was, newState, receiver[0], stamp)) {
+                    restartRequiredOnStarting = false;
+                    reloadRequiredOnStarting = false;
                     service.stateChanged(newState);
                     break;
                 }
@@ -137,6 +146,8 @@ public class ControlledProcessState {
 
     public void setStopping() {
         synchronized (service) {
+            restartRequiredOnStarting = false;
+            reloadRequiredOnStarting = false;
             state.set(State.STOPPING, stamp.incrementAndGet());
             service.stateChanged(State.STOPPING);
         }
@@ -144,6 +155,8 @@ public class ControlledProcessState {
 
     public void setStopped() {
         synchronized (service) {
+            restartRequiredOnStarting = false;
+            reloadRequiredOnStarting = false;
             state.set(State.STOPPED, stamp.incrementAndGet());
             service.stateChanged(State.STOPPED);
         }
@@ -159,6 +172,9 @@ public class ControlledProcessState {
         // Keep trying until stateRef is RELOAD_REQUIRED with our stamp
         for (;;) {
             State was = stateRef.get(receiver);
+            synchronized (service) {
+                reloadRequiredOnStarting = was == State.STARTING;
+            }
             if (was == State.STARTING || was == State.STOPPING || was == State.RESTART_REQUIRED) {
                 break;
             }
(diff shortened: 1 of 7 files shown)


host-controller/src/main/java/org/jboss/as/domain/controller/operations/SyncServerStateOperationHandler.java
```diff
@@ -117,8 +123,12 @@ public void execute(OperationContext context, ModelNode operation) throws Operat
...
                             ServerProcessStateHandler.REQUIRE_RESTART_OPERATION : ServerProcessStateHandler.REQUIRE_RELOAD_OPERATION;
                     final OperationStepHandler handler = context.getResourceRegistration().getOperationHandler(serverAddress, opName);
-                    final ModelNode op = Util.createEmptyOperation(opName, serverAddress);
-                    context.addStep(op, handler, OperationContext.Stage.MODEL);
+                    if (handler != null) {
+                        final ModelNode op = Util.createEmptyOperation(opName, serverAddress);
+                        context.addStep(op, handler, OperationContext.Stage.MODEL);
+                    } else {
+                        throw HostControllerLogger.ROOT_LOGGER.failedToSyncServerStatus(opName, serverAddress.toString());
+                    }
```
bstansberry, line 131: In our June 1 chat we talked about logging a WARN here, not throwing an exception. I think this is fine though; if this is null it's a ??? situation and we should probably fail.

controller/src/main/java/org/jboss/as/controller/ControlledProcessState.java
```diff
@@ -126,8 +130,13 @@ public void setRunning() {
                 break;
             }
             synchronized (service) {
-                State newState = restartRequiredFlag ? State.RESTART_REQUIRED : State.RUNNING;
+                State newState =  restartRequiredOnStarting ? State.RESTART_REQUIRED
+                        : reloadRequiredOnStarting ? State.RELOAD_REQUIRED
+                        : restartRequiredFlag ? State.RESTART_REQUIRED : State.RUNNING;
+                newStamp = restartRequiredOnStarting || reloadRequiredOnStarting ? currentStamp : newStamp;
```
bstansberry, line 136: Changing the value of newStamp instead of using a new variable makes me nervous, as this executes in a loop, so the reassignment can happen more than once. It might be fine, e.g. because of the synchronization on service, but it feels fragile / easily broken by a future maintainer.

bstansberry (changes requested):
Hi @yersan. Good job; I like the ControlledProcessState approach. I have a couple minor change requests but once you've handled those feel free to merge.

yersan:
@bstansberry I think the transition on starting still required a bit more thinking since the states of the newly added flags are not synched in the same way as the reference. The state of these flags is part of the state kept in the reference. I moved it to a draft until getting more confident with this

host-controller/src/main/java/org/jboss/as/domain/controller/operations/SyncServerStateOperationHandler.java
bstansberry: 👍

controller/src/main/java/org/jboss/as/controller/ControlledProcessState.java
```diff
@@ -117,6 +120,7 @@ public void setStarting() {
 
     public void setRunning() {
         AtomicStampedReference<State> stateRef = state;
+        int currentStamp = stamp.get();
         int newStamp = stamp.incrementAndGet();
```
bstansberry, line 124: I just noticed a race here. I think this should be:

int currentStamp = stamp.getAndIncrement();
int newStamp = currentStamp + 1; // the value from the increment we just did is "ours"

bstansberry (changes requested):

controller/src/main/java/org/jboss/as/controller/ControlledProcessState.java
yersan: If we increment it, we won't be able to revert from a restart/reload required. The issue is the stamp saved in the step that executed the setRestartRequired()/setReloadRequired() does not change, because it is an artificial state derived from a flag variable

yersan: If you want to see it in a different way, the `stamp.getAndIncrement()` you are requesting to do here, was already done when setRestartRequired()/setReloadRequired() was called: [ControlledProcessState.java#L192](https://github.com/wildfly/wildfly-core/blob/757e3012d9a14e3090fc129c22a82b15514d3e58/controller/src/main/java/org/jboss/as/controller/ControlledProcessState.java#L192)

yersan:
Hi @bstansberry , I'm fine with the current status, could you take a look again? thanks!

bstansberry (changes requested):
@yersan Below I point out 3 places where I'd like changes. They all revolve around L123 in your current branch, which has the effect of using 'stamp' as part of the state machine and not solely as a unique value generator. That works but it's harder to understand and thus is fragile.

To review this I checked out your branch to do my reviewing in my IDE, and while doing that I edited some to help me understand. In the end doing that led to a proposed change that addresses the issues I mention below, plus adds comments and makes some other misc adjustments. I've sent you a PR with that. It passes the controller module tests and testsuite/domain. 

https://github.com/yersan/wildfly-core/pull/7

controller/src/main/java/org/jboss/as/controller/ControlledProcessState.java
```diff
@@ -126,8 +130,13 @@ public void setRunning() {
                 break;
             }
             synchronized (service) {
-                State newState = restartRequiredFlag ? State.RESTART_REQUIRED : State.RUNNING;
-                if (state.compareAndSet(was, newState, receiver[0], newStamp)) {
+                State newState =  restartRequiredOnStarting ? State.RESTART_REQUIRED
+                        : reloadRequiredOnStarting ? State.RELOAD_REQUIRED
+                        : restartRequiredFlag ? State.RESTART_REQUIRED : State.RUNNING;
+                int stamp = restartRequiredOnStarting || reloadRequiredOnStarting ? currentStamp : newStamp;
```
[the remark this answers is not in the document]
yersan: > This use of currentStamp treats 'stamp' as part of the state machine.

Yes, thanks @bstansberry , I understand where my approach was failing, due to the thread scheduling, my approach could make it impossible to revert the restart/reload required.

Your suggestion prevents the stamp stored in the step could be different than the one expected to revert.

bstansberry (approved):
Thanks @yersan LGTM.

I had a thought this morning for further improvement in the future. Create a new InternalState enum that encapsulates State, reloadRequiredOnStarting and restartRequiredOnStarting. Then the 'state' var stores InternalState, not State. Then we just do state transitions, instead of manipulating multiple fields.  Besides probably being cleaner looking over all, that eliminates the somewhat 'magic' behavior in my commit from yesterday where setReload/RestartRequired return the current stamp when was == STARTING. Instead any caller gets a unique stamp and there's no fragility where someday a maintainer breaks the current magic.

PR #5595, opened 2023-08-02
[WFCORE-6442] Make ModuleSpecification retains distinct ModuleDepende…

bstansberry:
…ncy instances that have the same id.

https://issues.redhat.com/browse/WFCORE-6442

Note that an alternate approach to solving this problem would be to introduce some sort of ModuleDependency.merge(ModuleDependency other) method, such that a ModuleSpecification would end up with just one entry per module id. That would require more consideration though, particularly in how filter rules should be combined. The approach here is conservative in that it basically goes back to the longstanding pre-WFCORE-6199 (i.e. WF 27 and earlier) behavior of potential having multiple dependencies entries for a given module, but still tries to avoid pointless duplication.


yersan (approved):

yersan:
Thank you, @bstansberry

PR #5599, opened 2023-08-04
[WFCORE-6446] Enable better test failure output when the queue for sy…

jamezp:
…slog messages is not empty.

https://issues.redhat.com/browse/WFCORE-6446


yersan (approved):

yersan:
Merged @jamezp , I hope this helps with the final resolution ... random failures are always tricky

PR #5600, opened 2023-08-07
[WFCORE-5691] Allow setting timeout for token introspection

lvydra:
https://issues.redhat.com/browse/WFCORE-5691


yersan (commented):
Hi @lvydra , this PR is going to introduce a set of new fields under an Elytron resource visible via a profile in a Domain mode configuration. It should be based on an Elytron model bump and should include the required transformers for mixed-domain configurations

PR #5605, opened 2023-08-09
[WFCORE-6454] Use TCP/IP process communication for surefire plugin

jstourac:
https://issues.redhat.com/browse/WFCORE-6454

This change utilizes new TCP/IP communication for surefire execution [1].

By default, the communication writes to STDOUT which may lead to similar messages in the execution output if something else also writes there at wrong time:

```
[WARNING] Corrupted STDOUT by directly writing to native stream in forked JVM 1.
See FAQ web page and the dump file ... .dumpstream.
```

This change should mittigate such warnings and also should reduce number of open files during execution which may lead to crash on some systems with `nofile` limit set quite low.

Truth is that this issue hasn't been seen in this upstream WildFly Core project, still I believe this change should be beneficial avoiding eventual issue in the future.

[1] https://maven.apache.org/surefire/maven-surefire-plugin/examples/process-communication.html

### DIFF
--- a/testsuite/pom.xml
+++ b/testsuite/pom.xml
@@ -323,6 +323,7 @@
                     <artifactId>maven-surefire-plugin</artifactId>
                     <configuration>
                         <failIfNoTests>false</failIfNoTests>
+			<forkNode implementation="org.apache.maven.plugin.surefire.extensions.SurefireForkNodeFactory"/>
                         <redirectTestOutputToFile>${testLogToFile}</redirectTestOutputToFile>
                         <argLine>-Djava.util.logging.manager=org.jboss.logmanager.LogManager</argLine>
                         <systemPropertyVariables>
--- a/pom.xml
+++ b/pom.xml
@@ -135,6 +135,10 @@
         </surefire.system.args>
         <test.level>INFO</test.level>
 
+        <!-- This overrides default surefire plugin version specified in parent pom - reason is to enable TCP/IP communication,
+             see https://maven.apache.org/surefire/maven-surefire-plugin/examples/process-communication.html -->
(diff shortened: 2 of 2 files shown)


testsuite/pom.xml
```diff
@@ -154,6 +154,10 @@
              See https://issues.apache.org/jira/browse/MNG-2496 -->
         <version.org.apache.xalan>2.7.3</version.org.apache.xalan>
 
+	<!-- This overrides default version specified in parent pom - reason is to enable TCP/IP communication,
+	     see https://maven.apache.org/surefire/maven-surefire-plugin/examples/process-communication.html -->
+        <version.surefire.plugin>3.1.2</version.surefire.plugin>
```
bstansberry, line 159: I think this should be in the root pom, as surefire is used outside the testsuite tree, and we'll likely lose track of this separate setting here.

The xalan version above should perhaps move to the root pom too (I'm not asking you to do that; I'm just explaining what may look like an inconsistency.) But it's a somewhat different case, as there the dependency is only used in the testsuite module. I think part of the idea with xalan was to keep the property out of places where people could misuse it to configure production code versions.

jstourac: @bstansberry thank you for your review! I've moved the version definition to the root pom as you suggested. I also included `forkNode` definition in the surefire plugin configuration there too.

Let me know if you see some more issues, thanks! :slightly_smiling_face:

jstourac:
Just for the record - surefire plugin version is gonna be updated by https://github.com/jboss/jboss-parent-pom/pull/163. So we can remove the version overriding introduced in this PR once we move to the relevant parent pom in the future.

bstansberry (approved):

PR #5611, opened 2023-08-11
[WFCORE-6456] Add checks to LayersTest to confirm expected unused or …

bstansberry:
…unreferenced modules are indeed unused/unreferenced

Also reorganize the reporting logic and add a few asserts confirming the expected installations are there.

https://issues.redhat.com/browse/WFCORE-6456
https://issues.redhat.com/browse/WFCORE-6459
https://issues.redhat.com/browse/WFCORE-6465

Also, split the 'unused' testing from the 'unreferenced' as they are orthogonal.

https://issues.redhat.com/browse/WFCORE-6481

Also improves testing of alias modules:

https://issues.redhat.com/browse/WFCORE-6498

### DIFF
--- a/testsuite/layers/src/test/java/org/jboss/as/test/layers/LayersTestCase.java
+++ b/testsuite/layers/src/test/java/org/jboss/as/test/layers/LayersTestCase.java
@@ -38,11 +38,7 @@ public class LayersTestCase {
         // Not currently used internally
         "org.wildfly.event.logger",
         // wildfly-elytron-http-stateful-basic
-        "org.wildfly.security.http.sfbasic",
-        // wildfly-elytron-tool
-        "org.apache.commons.cli",
-        "org.apache.commons.lang3",
-        "org.wildfly.security.elytron-tool",
+        "org.wildfly.security.http.sfbasic"
     };
     // Packages that are not referenced from the module graph but needed.
     // This is the expected set of un-referenced modules found when scanning
@@ -70,6 +66,10 @@ public class LayersTestCase {
         "wildflyee.api",
         // bootable jar runtime
         "org.wildfly.bootable-jar",
+        // wildfly-elytron-tool
+        "org.apache.commons.cli",
+        "org.apache.commons.lang3",
+        "org.wildfly.security.elytron-tool",
     };
 
     /**
--- a/testsuite/layers/src/test/java/org/jboss/as/test/layers/LayersTestCase.java
+++ b/testsuite/layers/src/test/java/org/jboss/as/test/layers/LayersTestCase.java
@@ -5,9 +5,7 @@
 package org.jboss.as.test.layers;
 
 import java.io.File;
-import java.util.Arrays;
 import java.util.HashMap;
-import java.util.HashSet;
 import java.util.List;
 
 import org.junit.AfterClass;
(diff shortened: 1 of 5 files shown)


testsuite/shared/src/main/java/org/jboss/as/test/layers/LayersTest.java
```diff
@@ -321,6 +360,27 @@ private static Set<String> appendResult(String title, Result result, StringBuild
...
     }
 
+    private static String appendExceptionMsg(String existing, String toAppend) {
+        if (!existing.isEmpty()) {
+            existing += '\n';
+        }
+        return existing + toAppend;
+    }
+
+    private static String listModules(Set<String> set, Predicate<String> predicate) {
```
yersan, line 370: To improve the readability, this can be replaced by Collectors.joining(","):

```
return set.stream()
                .filter(s -> {
                    assert s != null;
                    return predicate.test(s);
                })
                .collect(Collectors.joining(","));
```

testsuite/shared/src/main/java/org/jboss/as/test/layers/LayersTest.java
```diff
@@ -321,6 +360,27 @@ private static Set<String> appendResult(String title, Result result, StringBuild
         return deltaModules;
     }
 
+    private static String appendExceptionMsg(String existing, String toAppend) {
+        if (!existing.isEmpty()) {
+            existing += '\n';
+        }
+        return existing + toAppend;
+    }
```
yersan, line 368: What do you think if instead of using Strings this is replaced by using a StreamBuilder()? This would reduce memory consumptions

testsuite/shared/src/main/java/org/jboss/as/test/layers/LayersTest.java
bstansberry: Done

bstansberry:
@jfdenise FYI re this. Testing against a snapshot of this led to a lot of improvements in how the test is used in full WF -- see https://github.com/wildfly/wildfly/pull/17148.

testsuite/layers/src/test/java/org/jboss/as/test/layers/LayersTestCase.java
```diff
@@ -32,17 +32,13 @@ public class LayersTestCase {
         "org.jboss.as.threads",
         "sun.jdk",
         "sun.scripting",
-        // No patching modules in layers
+        // TODO why doesn't core-tools provision these given that passive+ is used?
```
jfdenise, line 26: @bstansberry , org.jboss.as.cli JBoss Modules module has an optional dependency on patching. Optional dependencies in module.xml are ignored to identify the Galleon package dependencies tree. No layer brings patching, so patching is not provisioned.

bstansberry: Thanks @jfdenise. We talked about that but I forgot to come back and remove the comment.

bstansberry: @jfdenise Fixed

yersan:
Ok, @bstansberry, I've added the following integration tests. @jfdenise if you are fine with all the changes, would you mind approving it?

Linux Integration JDK 11
https://ci.wildfly.org/buildConfiguration/WF_WildFlyCoreIntegrationExperiments/404910

Galleon Integration Job
https://ci.wildfly.org/buildConfiguration/WF_WildFlyCoreIntegrationExperiments/404912

Windows Integration JDK 11
https://ci.wildfly.org/buildConfiguration/WF_WildFlyCoreIntegrationExperimentsWindows/404916

jfdenise (approved):

bstansberry:
Hopefully the https://github.com/wildfly/wildfly/pull/17618 fix will mean this passes and can be merged.

PR #5628, opened 2023-08-17
[WFCORE-6471] Upgrade FasterXML Jackson to 2.14.3

bstansberry:
https://issues.redhat.com/browse/WFCORE-6471

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -185,8 +185,8 @@
         <version.com.google.code.findbugs>3.0.2</version.com.google.code.findbugs>
         <version.com.google.guava>31.1-jre</version.com.google.guava>
         <version.com.google.guava.failureaccess>1.0.1</version.com.google.guava.failureaccess>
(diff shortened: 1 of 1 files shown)


bstansberry:
@fjuma @Skyllarr Is this ok?

This version prop is only used in testbom/pom.xml, and I'm pretty sure that's just to control the Jackson version used by mockserver for AcmeMockServerBuilder. We started directly controlling it instead of relying on transitives when we started getting scanner reports about vulnerabilties in the transitive versions.

fjuma (approved):
Thanks @bstansberry!

Just FYI, Elytron is being upgraded to FasterXML Jackson 2.15.2 due to another CVE, see:

https://github.com/wildfly-security/wildfly-elytron/pull/1951

bstansberry:
Thanks @fjuma -- I expect another upgrade will be coming soon then.

PR #5639, opened 2023-08-22
[WFCORE-6466] Split LayersTest#test into two separate methods

The-Huginn:
JIRA: https://issues.redhat.com/browse/WFCORE-6466

### DIFF
--- a/testsuite/shared/src/main/java/org/jboss/as/test/layers/LayersTest.java
+++ b/testsuite/shared/src/main/java/org/jboss/as/test/layers/LayersTest.java
@@ -71,31 +71,46 @@ public class LayersTest {
      * @param unused The set of modules that are OK to not be provisioned when all layers are provisioned.
      * If more modules than this set are not provisioned, it means that we are missing some modules and
      * an error occurs.
-     * @throws Exception
+     * @throws Exception on failure
      */
+    @Deprecated
     public static void test(String root, Set<String> unreferenced, Set<String> unused) throws Exception {
+        testExecution(root);
+        testDeployedModules(root, unreferenced, unused);
+    }
+
+    /**
+     * Checks only allowed modules were loaded with the installation
+     * @param root Path to installation
+     * @param unreferenced The set of modules that are present in a default installation (with all modules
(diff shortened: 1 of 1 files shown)


testsuite/shared/src/main/java/org/jboss/as/test/layers/LayersTest.java
```diff
@@ -71,31 +71,68 @@ public class LayersTest {
      * @param unused The set of modules that are OK to not be provisioned when all layers are provisioned.
      * If more modules than this set are not provisioned, it means that we are missing some modules and
      * an error occurs.
-     * @throws Exception
+     * @throws Exception on failure
      */
+    @Deprecated
     public static void test(String root, Set<String> unreferenced, Set<String> unused) throws Exception {
+        testBootInstallations(root);
```
bstansberry, line 78: This should be a call to the existing testExecutions(String) as the new testBootInstallations does the same thing.

testsuite/shared/src/main/java/org/jboss/as/test/layers/LayersTest.java
```diff
@@ -71,31 +71,68 @@ public class LayersTest {
...
-                    layers.put(f.getName(), res);
                 }
             }
         } finally {
             executor.shutdownNow();
         }
+    }
+
+    /**
+     * Checks only allowed modules were loaded with the installation
```
bstansberry, line 105: I'd say

"Checks the expected modules were provisioned with the @{code test-standalone-reference} and @{test-all-layers} installations."

bstansberry (changes requested):
Thanks for looking at this @The-Huginn. I think it can be simplified a bit, but this is what I had in mind.

testsuite/shared/src/main/java/org/jboss/as/test/layers/LayersTest.java
The-Huginn: Sorry, I didn't realize that

bstansberry (approved):

bstansberry:
Thanks @The-Huginn

PR #5642, opened 2023-08-30
[WFCORE-6255] Deployment scanner did not undeploy failed archives

khermano:
Issue: https://issues.redhat.com/browse/WFCORE-6255

### DIFF
--- a/deployment-scanner/src/main/java/org/jboss/as/server/deployment/scanner/FileSystemDeploymentService.java
+++ b/deployment-scanner/src/main/java/org/jboss/as/server/deployment/scanner/FileSystemDeploymentService.java
@@ -41,6 +41,7 @@ import java.nio.file.DirectoryStream.Filter;
 import java.nio.file.Files;
 import java.nio.file.Path;
 import java.util.ArrayList;
+import java.util.Collections;
 import java.util.HashMap;
 import java.util.HashSet;
 import java.util.Iterator;
@@ -152,6 +153,12 @@ class FileSystemDeploymentService implements DeploymentScanner, NotificationHand
     @SuppressWarnings("deprecation")
     private final DeploymentTransformer deploymentTransformer;
 
+    /**
+     * Property have to be guarded by scanLock.
+     * Reads or writes of this field can be done only in code executing
+     * between a call to {@link #acquireScanLock()} and {@link #releaseScanLock()}
+     */
+    private ScanContext scanContext;
 
     @Override
     public void handleNotification(Notification notification) {
@@ -505,6 +512,23 @@ class FileSystemDeploymentService implements DeploymentScanner, NotificationHand
     }
 
     /**
+     * This method exists only for testing purposes
+     *
+     * @return unmodifiable collection of registered deployments, null otherwise
+     */
+    Map<String,Boolean> getScanContextRegisteredDeployments() {
+        if (acquireScanLock()) {
+            try {
+                assert scanContext != null;
+                return Collections.unmodifiableMap(scanContext.registeredDeployments);
+            } finally {
+                releaseScanLock();
+            }
+        }
+        return null;
+    }
+
+    /**
      * Perform a post-boot scan to remove any deployments added during boot that failed to deploy properly.
      * This method isn't private solely to allow a unit test in the same package to call it.
      */
@@ -513,7 +537,7 @@ class FileSystemDeploymentService implements DeploymentScanner, NotificationHand
         if (acquireScanLock()) {
             try {
                 ROOT_LOGGER.tracef("Performing a post-boot forced undeploy scan for scan directory %s", deploymentDir.getAbsolutePath());
-                ScanContext scanContext = new ScanContext(deploymentOperations);
+                scanContext = new ScanContext(deploymentOperations);
 
                 // Add remove actions to the plan for anything we count as
                 // deployed that we didn't find on the scan
@@ -540,6 +564,7 @@ class FileSystemDeploymentService implements DeploymentScanner, NotificationHand
             } catch (Exception e) {
                 ROOT_LOGGER.scanException(e, deploymentDir.getAbsolutePath());
             } finally {
+                scanContext = null;
                 releaseScanLock();
             }
         }
@@ -581,7 +606,7 @@ class FileSystemDeploymentService implements DeploymentScanner, NotificationHand
                 deployedContentEstablished = true;
             }
 
-            ScanContext scanContext = null;
+            scanContext = null;
             try {
                 scanContext = new ScanContext(deploymentOperations);
             } catch (RuntimeException ex) {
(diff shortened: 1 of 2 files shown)


deployment-scanner/src/main/java/org/jboss/as/server/deployment/scanner/FileSystemDeploymentService.java
```diff
@@ -931,6 +931,7 @@ private void scanDirectory(final File directory, final String relativePath, fina
                 final String deploymentName = fileName.substring(0, fileName.length() - FAILED_DEPLOY.length());
                 scanContext.toRemove.remove(deploymentName);
                 if (!deployed.containsKey(deploymentName) && !(new File(child.getParent(), deploymentName).exists())) {
+                    scanContext.scannerTasks.add(new UndeployTask(deploymentName, deploymentDir, child.lastModified(), true));
```
bstansberry, line 934: I believe child.lastModified() should be scanContext.scanStartTime. If this actually does an undeploy, the undeploy didn't happen when the FAILED_DEPLOY marker was written.

It doesn't really matter because this param ends up unused since you pass 'true' as the forcedUndeploy param, but maybe after some future change it will matter.

bstansberry: A more substantive problem is there needs to be a check that this isn't going to undeploy something not associated with this scanner; e.g. a scanned deployment failed, and admin removed the file, and then the admin used the CLI to force-deploy.

I believe scanContext.persistentDeployments is meant to contain the set of deployments not associated with this scanner, so that could be how to check.

Tests of this change should cover this scenario in addition to the basic one described in the JIRA.

deployment-scanner/src/test/java/org/jboss/as/server/deployment/scanner/FileSystemDeploymentServiceUnitTestCase.java
```diff
@@ -495,6 +495,7 @@ public void testCleanSpuriousMarkers() throws Exception {
         f1 = createFile("spurious" + FileSystemDeploymentService.DEPLOYED);
         f2 = createFile(new File(tmpDir, "nested"), "nested" + FileSystemDeploymentService.DEPLOYED);
 
+        ts.controller.addCompositeSuccessResponse(1);
```
bstansberry, line 498: I would prefer this test being left as is and one or two next test methods specifically targeting this JIRA be added. It's too hard to reason about how this test change means the fix is being tested.

If you had to do this just to get the existing test to pass, a code comment explaining why it's needed would be good.

bstansberry (changes requested):
Thanks @khermano -- the approach seems reasonable, although I mention a scenario that needs to be covered, plus this needs specific testing.

deployment-scanner/src/test/java/org/jboss/as/server/deployment/scanner/FileSystemDeploymentServiceUnitTestCase.java
khermano: Hello @bstansberry. Thank you very much for the review and detailed description of the problem.
Yes, I just tried to get the existing test pass.
If I understand it correctly, the test was failing because I added UndeploymentTask in FileSystemDeploymentService which wasn't considered by the mocked controller. I guess the addCompositeSuccessResponse method only adds one success response to MockServerController.responses to use it in the processOp method somehow as a counter of requests.
Should I also try to create some other test?
For now I will force-push my changes.

bstansberry: Hi @khermano -- yes, please add another test or two, e.g. one that specifically covers the bug and another that validates that a deployment not associated with the scanner being tested isn't undeployed just because a cruft FAILED_DEPLOY marker file is left in the scanner's directory. I *think* the test infrastructure used by FileSystemDeploymentServiceTestCase can handle that scenario, or at least is close to being able to.

It sounds like the line I commented on here needs to remain, which is fine, but preceding it with a brief comment would be good as it's not obvious why it's there.

khermano: Hello @bstansberry , 
so I added the comment and the test for the bug you recommended, but I am unsure about the other one.
I tried a couple of things to deploy the failing kitchensink.war through CLI and console, but none of them create any content in the scanned standalone/deployments directory, not even `deploy ~/../kitchensink.war --force`.
And I think that maybe creating markers like ".failed", ".deployed" and others is only a deployment-scanner thing, or am I wrong about that?

khermano: Hi @bstansberry ,
I am sorry for the delay. I got stuck with this and needed to work on something else, but I think I finally have it. 
So again, I'm really sorry for the inconvenience. Hopefully, it looks okay.

deployment-scanner/src/main/java/org/jboss/as/server/deployment/scanner/FileSystemDeploymentService.java
```diff
@@ -152,6 +153,11 @@ class FileSystemDeploymentService implements DeploymentScanner, NotificationHand
     @SuppressWarnings("deprecation")
     private final DeploymentTransformer deploymentTransformer;
 
+    private ScanContext scanContext;
```
bstansberry, line 156: Making this an instance variable is a bit dangerous for maintenance, so it needs a good usage comment. Something like

// Guarded by scanLock. 
// Reads or writes of this field can only be done in code executing between a call to {@link #acquireScanLock()} and {@link #releaseScanLock()}

deployment-scanner/src/main/java/org/jboss/as/server/deployment/scanner/FileSystemDeploymentService.java
```diff
@@ -152,6 +153,11 @@ class FileSystemDeploymentService implements DeploymentScanner, NotificationHand
...
+
+    public Map<String,Boolean> getScanContextRegisteredDeployments() {
```
bstansberry, line 158: This should be package-protected and moved down in the file to a less prominent position. Please add method javadoc stating the method only exists for testing.  The test is in the same package, so package-protected works. That kind of thing is why it's in the same package. :)

deployment-scanner/src/main/java/org/jboss/as/server/deployment/scanner/FileSystemDeploymentService.java
```diff
@@ -152,6 +153,11 @@ class FileSystemDeploymentService implements DeploymentScanner, NotificationHand
...
+        return Collections.unmodifiableMap(scanContext.registeredDeployments);
```
bstansberry, line 159: Per my comment above about the new scanContext field, please wrap this in acquireScanLock() try { ... } finally { releaseScanLock(); }

an 'assert scanContext != null' would also be good.

Doing all this doesn't really matter for your test but it reduces the risk of future misuse of this method by having the test code follow the concurrency rules that we create for the production code.

bstansberry (changes requested):
Hi @khermano 

The tests look good. Converting scanContext to an instance field brings some maintenance risk related to thread safety, so I've requested a few changes related to reducing that risk. But the basic fix looks fine.

bstansberry (approved):

bstansberry:
Thanks @khermano

PR #5643, opened 2023-08-31
WFCORE-6472 Remove the remoting subsystem transformers and long deprecated resources/attributes

pferraro:
https://issues.redhat.com/browse/WFCORE-6472
https://issues.redhat.com/browse/WFCORE-6473 (implicitly addressed)

Resources and attributes deprecated since WF23 have been dropped.  This includes the weird handling described in WFCORE-6473.
I've added an enumeration of supported model versions, to be referenced by future transformers.

This PR also adds new 6.0 version of schema that prunes obsolete elements/attributes for which there is no longer any model in which to persist.
The parser/writer for version 6.0 is now based on PersistentResourceXMLDescription, and the 6.0 schema simplified accordingly.
Subsystem tests have also been parameterized to accommodate future schema versions, so that parsing of legacy schemas tests do not get dropped.


pferraro:
@bstansberry FYI - full integration failures are due to these lines that need to be removed in main wildfly:
https://github.com/wildfly/wildfly/blob/main/ee-feature-pack/galleon-shared/src/main/resources/feature_groups/remoting.xml#L5
https://github.com/wildfly/wildfly/blob/main/ee-feature-pack/galleon-shared/src/main/resources/feature_groups/unsecured-basic-profile.xml#L6

It is unclear to me whether those changes can be made independently from this PR.
Side note: are we ready to merge wildfly-core back into wildfly yet?

pferraro:
Closing for now.  I seem to have screwed something up.

PR #2148, opened 2017-02-03
[WFCORE-2262] Latest WildFly Elytron and Invocation Changes

darranl:
:bangbang: Commits from this branch are used in numerous other topic branches so please do not cherry-pick or rebase. :bangbang:

The following issues are resolved by this PR: -

https://issues.jboss.org/browse/WFCORE-1984
https://issues.jboss.org/browse/WFCORE-2078
https://issues.jboss.org/browse/WFCORE-2115
https://issues.jboss.org/browse/WFCORE-2239
https://issues.jboss.org/browse/WFCORE-1891
https://issues.jboss.org/browse/WFCORE-2261
https://issues.jboss.org/browse/WFCORE-2262
https://issues.jboss.org/browse/WFCORE-2263

WildFly PR https://github.com/wildfly/wildfly/pull/9595

Experimental Builds: -
Linux https://ci.wildfly.org/viewLog.html?buildId=44066&
Windows https://ci.wildfly.org/viewLog.html?buildId=44068&

Linux experimental build with disabled tests enabled: -
https://ci.wildfly.org/viewLog.html?buildId=44067&
(Expect approximately 91 failures)

PR #5644, opened 2023-09-01
WFCORE-6472 Remove obsolete remoting subsystem transformers and long deprecated resources/attributes

pferraro:
https://issues.redhat.com/browse/WFCORE-6472
https://issues.redhat.com/browse/WFCORE-6473

Resources and attributes deprecated since WF23 have been dropped. This includes the weird handling described in WFCORE-6473.
I've added an enumeration of supported model versions, to be referenced by future transformers.

This PR also adds new 6.0 version of schema that prunes obsolete elements/attributes for which there is no longer any model in which to persist.
The parser/writer for version 6.0 is now based on PersistentResourceXMLDescription, and the 6.0 schema simplified accordingly.
Subsystem tests have also been parameterized to accommodate future schema versions, so that testing of legacy schema parsing does not get dropped.
The worker attribute has been made required, rather than a hidden default, since this is a capability reference.

~~Requires https://github.com/wildfly/wildfly/pull/17145 to fix wildfly full integration test failures.~~ Merged.


pferraro:
wildfly full integration tests will likely fail, requiring https://github.com/wildfly/wildfly/pull/17145

remoting/subsystem/src/main/java/org/jboss/as/remoting/RemotingSubsystemAdd.java
```diff
@@ -72,26 +71,33 @@ protected void performRuntime(OperationContext context, ModelNode operation, Res
...
+        Consumer<Endpoint> endpointConsumer = builder.provides(REMOTING_ENDPOINT_CAPABILITY);
+
+        if (workerName != null) {
+            OptionMap map = EndpointConfigFactory.populate(context, endpointModel);
+            String nodeName = WildFlySecurityManager.getPropertyPrivileged(RemotingExtension.NODE_NAME_PROPERTY, null);
+
+            Supplier<XnioWorker> workerSupplier = builder.requiresCapability(IO_WORKER_CAPABILITY_NAME, XnioWorker.class, workerName);
+            builder.setInstance(new EndpointService(endpointConsumer, workerSupplier, nodeName, EndpointService.EndpointType.SUBSYSTEM, map));
+        } else {
+            // Otherwise, reference endpoint created by DomainServerCommunicationServices
```
bstansberry, line 86: Perhaps

"// This is a domain server as WORKER is a required attribute otherwise. Reference endpoint created by DomainServerCommunicationServices"

When looking at this it took me a while to figure out why the else block just assumed a domain server.

remoting/subsystem/src/main/java/org/jboss/as/remoting/RemotingSubsystemRootResource.java
```diff
@@ -106,65 +102,30 @@ public class RemotingSubsystemRootResource extends SimpleResourceDefinition {
...
-                .setAddHandler(new RemotingSubsystemAdd(attributes))
+                .setAddHandler(new RemotingSubsystemAdd())
                 .setAddRestartLevel(OperationEntry.Flag.RESTART_NONE)
                 .setRemoveHandler(new ReloadRequiredRemoveStepHandler())
                 .setRemoveRestartLevel(OperationEntry.Flag.RESTART_ALL_SERVICES)
                 .setCapabilities(REMOTING_ENDPOINT_CAPABILITY, HTTP_LISTENER_REGISTRY_CAPABILITY)
         );
-        this.attributes = attributes;
+        // Worker is not required if endpoint is meant to be supplied by DomainServerCommunicationServices
+        this.workerAttribute = new SimpleAttributeDefinitionBuilder(WORKER).setRequired(processType != ProcessType.DOMAIN_SERVER).build();
```
bstansberry, line 120: I think that means it's not required on an HC either; i.e. domain.xml doesn't need to specify it if the server doesn't need it.

core-feature-pack/galleon-common/src/main/resources/layers/standalone/remoting/layer-spec.xml
```diff
@@ -5,10 +5,4 @@
         <layer name="elytron"/>
     </dependencies>
     <feature-group name="remoting"/>
-
-    <feature spec="subsystem.remoting.http-connector">
-        <param name="http-connector" value="http-remoting-connector"/>
-        <param name="connector-ref" value="default"/>
-        <param name="sasl-authentication-factory" value="application-sasl-authentication"/>
-    </feature>
```
bstansberry, line 13: Note to self:

This change requires https://github.com/wildfly/wildfly/pull/17145/files as that is where the remoting feature group used in full WF that provides the http-connector is configured. The 'remoting' feature group file here is not relevant in full WF.

bstansberry (changes requested):
@pferraro @fl4via 

I've made some minor review comments below. I think the biggest issue to discuss are the various breaking changes.

1) The original WFCORE-6472 request; the legacy endpoint resource. As we discussed starting at https://github.com/wildfly/wildfly-core/pull/5563#discussion_r1298984656 I think this is fine. It just removes an alternate view with no difference in runtime behavior.

2) The remote outbound connection removal stuff. I think that should have a different JIRA. This I think needs good discussion as it is fairly large and is a behavior change. When  GenericOutboundConnectionResourceDefinition and LocalOutboundConnectionResourceDefinition were deprecated the advice was to replace with RemoteOutboundConnectionResourceDefinition with the protocol configured. But now that is not possible as 'protocol' is removed, so users have to configure an authentication context. Looking at how that works I don't get the impression it is possible to set the protocol. I think we may be limited to remote+http.

3) The removal of the default value for WORKER. You've handled that well with respect to domain mode and legacy schema parsing; the question here is if we're ok with breaking CLI scripts.


Thanks, Paul, for doing this as this extension was due for a general update.

remoting/subsystem/src/main/java/org/jboss/as/remoting/RemotingSubsystemModel.java
```diff
@@ -0,0 +1,42 @@
...
+    VERSION_5_0_0(5, 0, 0), // WildFly 12 - 26, EAP 7.4
+    VERSION_6_0_0(6, 0, 0), // WildFly 27 - present, EAP 8.0
+    VERSION_7_0_0(6, 0, 0), // WildFly 30
+    ;
+    static final RemotingSubsystemModel CURRENT = VERSION_7_0_0;
+
+    private final ModelVersion version;
+
+    RemotingSubsystemModel(int major, int minor, int micro) {
+        this.version = ModelVersion.create(major);
```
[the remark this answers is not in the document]
pferraro: Indeed.

remoting/subsystem/src/main/java/org/jboss/as/remoting/RemotingSubsystemAdd.java
pferraro: I will clarify accordingly.  Thanks for the feedback.

remoting/subsystem/src/main/java/org/jboss/as/remoting/RemotingSubsystemRootResource.java
pferraro: This logic was based on https://github.com/wildfly/wildfly-core/blob/main/remoting/subsystem/src/main/java/org/jboss/as/remoting/RemotingSubsystemAdd.java#L83
https://github.com/wildfly/wildfly-core/blob/main/remoting/subsystem/src/main/java/org/jboss/as/remoting/RemotingSubsystemAdd.java#L83

But, to your point, ProcessType.HOST_CONTROLLER would not execute any runtime phase, so the model should not otherwise require this, I think.

core-feature-pack/galleon-common/src/main/resources/layers/standalone/remoting/layer-spec.xml
pferraro: Correct.

pferraro:
Thanks for the detailed feedback, @bstansberry 

> @pferraro @fl4via
> 
> I've made some minor review comments below. I think the biggest issue to discuss are the various breaking changes.
> 
> 1. The original WFCORE-6472 request; the legacy endpoint resource. As we discussed starting at [WFCORE-6407 Deprecate non-default contructors from AbstractAddStepHandler and AbstractWriteAttributeHandler #5563 (comment)](https://github.com/wildfly/wildfly-core/pull/5563#discussion_r1298984656) I think this is fine. It just removes an alternate view with no difference in runtime behavior.

OK.

> 2. The remote outbound connection removal stuff. I think that should have a different JIRA. This I think needs good discussion as it is fairly large and is a behavior change. When  GenericOutboundConnectionResourceDefinition and LocalOutboundConnectionResourceDefinition were deprecated the advice was to replace with RemoteOutboundConnectionResourceDefinition with the protocol configured. But now that is not possible as 'protocol' is removed, so users have to configure an authentication context. Looking at how that works I don't get the impression it is possible to set the protocol. I think we may be limited to remote+http.

I did a fair bit of archaeology for this work and found that the `protocol` and `security-realm` attributes of the `remote-outbound-connection` resource [1], as well as the `generic-outbound-connection` and `local-outbound-connection` resources were deprecated in wildfly-core 3.0.0.Final, last used in WF11 (yep, 11!).  Given that this proposed change removes everything deprecated since WF23, including the legacy configuration=endpoint resource (deprecated in WF12), I am a little surprised that removing these would be at all controversial.

That said, it would be no trouble to restore these resources if their removal is indeed problematic.

[1] https://github.com/wildfly/wildfly-core/blob/3.0.0.Final/remoting/subsystem/src/main/java/org/jboss/as/remoting/RemoteOutboundConnectionResourceDefinition.java#L81
[2] https://github.com/wildfly/wildfly-core/blob/3.0.0.Final/remoting/subsystem/src/main/java/org/jboss/as/remoting/GenericOutboundConnectionResourceDefinition.java#L55
[3] https://github.com/wildfly/wildfly-core/blob/3.0.0.Final/remoting/subsystem/src/main/java/org/jboss/as/remoting/LocalOutboundConnectionResourceDefinition.java#L59

> 3. The removal of the default value for WORKER. You've handled that well with respect to domain mode and legacy schema parsing; the question here is if we're ok with breaking CLI scripts.

Good point.  Who can answer that definitely?

remoting/subsystem/src/main/java/org/jboss/as/remoting/RemotingSubsystemRootResource.java
pferraro: Replaced conditional with ProcessType.isManagedDomain(), which should cover all use cases, I think.

bstansberry:
> 
> I did a fair bit of archaeology for this work and found that the `protocol` and `security-realm` attributes of the `remote-outbound-connection` resource [1], as well as the `generic-outbound-connection` and `local-outbound-connection` resources were deprecated in wildfly-core 3.0.0.Final, last used in WF11 (yep, 11!). 

Ok, that's good. My archaelogy skills appear to have rusted as from my own research I had the impression it was later than that!

> Given that this proposed change removes everything deprecated since WF23, including the legacy configuration=endpoint resource (deprecated in WF12), I am a little surprised that removing these would be at all controversial.
> 

We had a nice window to lots of deprecation cleanup breaking changes in WF 29, particularly 29 Alpha1. This would have been a no-brainer then.

Outside that window it needs more thought.

> That said, it would be no trouble to restore these resources if their removal is indeed problematic.

Ok, we'll see where the discussion goes.

> 
> [1] https://github.com/wildfly/wildfly-core/blob/3.0.0.Final/remoting/subsystem/src/main/java/org/jboss/as/remoting/RemoteOutboundConnectionResourceDefinition.java#L81 [2] https://github.com/wildfly/wildfly-core/blob/3.0.0.Final/remoting/subsystem/src/main/java/org/jboss/as/remoting/GenericOutboundConnectionResourceDefinition.java#L55 [3] https://github.com/wildfly/wildfly-core/blob/3.0.0.Final/remoting/subsystem/src/main/java/org/jboss/as/remoting/LocalOutboundConnectionResourceDefinition.java#L59
> 
> > 3. The removal of the default value for WORKER. You've handled that well with respect to domain mode and legacy schema parsing; the question here is if we're ok with breaking CLI scripts.
> 
> Good point. Who can answer that definitely?

@fl4via, me, @darranl, other knowledgable engineers particularly including yourself.

fl4via (approved):

pferraro:
@bstansberry @fl4via As discussed, I've narrowed the scope of this PR and have retained the existing local-outbound-connection and generic outbound-connection resources [1] and restored the worker attribute of the subsystem resource to be optional [2], where its magic default value is applied conditionally (only for server processes).
[1] https://issues.redhat.com/browse/WFCORE-6489
[2] https://issues.redhat.com/browse/WFCORE-6490

pferraro:
Rebased against main.

remoting/subsystem/src/main/java/org/jboss/as/remoting/GenericOutboundConnectionResourceDefinition.java
```diff
@@ -40,39 +44,35 @@ class GenericOutboundConnectionResourceDefinition extends AbstractOutboundConnec
...
-                .setRemoveHandler(new ServiceRemoveStepHandler(OUTBOUND_CONNECTION_CAPABILITY.getCapabilityServiceName(), GenericOutboundConnectionAdd.INSTANCE))
+                .setAddHandler(addHandler)
+                .setRemoveHandler(new ServiceRemoveStepHandler(OUTBOUND_CONNECTION_CAPABILITY.getCapabilityServiceName(), addHandler))
                 .setDeprecatedSince(RemotingSubsystemModel.VERSION_4_0_0.getVersion())
         );
     }
 
     @Override
     public void registerChildren(ManagementResourceRegistration resourceRegistration) {
+        // TODO why does this resource have properties that are never referenced?
```
bstansberry, line 67: @pferraro @fl4via This dates to:

https://github.com/wildfly/wildfly-core/commit/7281132dee1f595f165720dcd4c91c7ee9dd7256

which came in as part of the big Elytron integration PR https://github.com/wildfly/wildfly-core/pull/2148

due to its incorporation of https://github.com/wildfly-security-incubator/wildfly-core/pull/37

The child is kind of deprecated because this resource is deprecated but I suppose it should be deprecatd on its own. And if we decided to undeprecate this and local-outbound-connection for sure we should deprecate the child.

remoting/subsystem/src/main/java/org/jboss/as/remoting/LocalOutboundConnectionResourceDefinition.java
```diff
@@ -40,38 +44,39 @@ class LocalOutboundConnectionResourceDefinition extends AbstractOutboundConnecti
 
     static final PathElement ADDRESS = PathElement.pathElement(CommonAttributes.LOCAL_OUTBOUND_CONNECTION);
 
-    public static final SimpleAttributeDefinition OUTBOUND_SOCKET_BINDING_REF = new SimpleAttributeDefinitionBuilder(CommonAttributes.OUTBOUND_SOCKET_BINDING_REF, ModelType.STRING, false)
+    // TODO This is never used - why is it required?
```
bstansberry, line 47: @pferraro @fl4via This dates to:

https://github.com/wildfly/wildfly-core/commit/7281132dee1f595f165720dcd4c91c7ee9dd7256

which came in as part of the big Elytron integration PR https://github.com/wildfly/wildfly-core/pull/2148

due to its incorporation of https://github.com/wildfly-security-incubator/wildfly-core/pull/37

I guess the local: protocol used to involve a host name in its URIs, but no longer does.

If this resource survives this attribute should be deprecated and made optional.

bstansberry: It also shouldn't allow expressions :)

remoting/subsystem/src/main/java/org/jboss/as/remoting/LocalOutboundConnectionWriteHandler.java
```diff
@@ -1,78 +0,0 @@
...
-        final ServiceRegistry registry = context.getServiceRegistry(true);
-        ServiceController sc = registry.getService(serviceName);
-        if (sc != null && sc.getState() == ServiceController.State.UP) {
-            reloadRequired = true;
-        } else {
-            // Service isn't up so we can bounce it
-            context.removeService(serviceName); // safe even if the service doesn't exist
-            // install the service with new values
-            LocalOutboundConnectionAdd.INSTANCE.installRuntimeService(context, fullModel);
-        }
```
bstansberry, line 75: Big +1 to getting rid of this.

remoting/subsystem/src/main/java/org/jboss/as/remoting/RemoteOutboundConnectionWriteHandler.java
```diff
@@ -1,82 +0,0 @@
...
-            reloadRequired = true;
-        } else {
-            // Service isn't up so we can bounce it
-            context.removeService(serviceName); // safe even if the service doesn't exist
-            // install the service with new values
-            RemoteOutboundConnectionAdd.INSTANCE.installRuntimeService(context, fullModel);
-        }
-
-        return reloadRequired;
-    }
```
bstansberry, line 81: Big +1 to getting rid of this.

remoting/subsystem/src/main/java/org/jboss/as/remoting/RemotingExtension.java
```diff
@@ -104,9 +104,9 @@ public void initialize(ExtensionContext context) {
         httpSasl.registerSubModel(PropertyResource.INSTANCE_HTTP_CONNECTOR);
 
         // remote outbound connection
-        subsystem.registerSubModel(RemoteOutboundConnectionResourceDefinition.INSTANCE);
+        subsystem.registerSubModel(new RemoteOutboundConnectionResourceDefinition());
         // local outbound connection
-        subsystem.registerSubModel(LocalOutboundConnectionResourceDefinition.INSTANCE);
+        subsystem.registerSubModel(new LocalOutboundConnectionResourceDefinition());
         // (generic) outbound connection
         subsystem.registerSubModel(new GenericOutboundConnectionResourceDefinition());
```
bstansberry, line 114: TODO -- this children wiring should be moved out of the Extension impl.

bstansberry (approved):

PR #5645, opened 2023-09-04
[WFCORE-6477]: Possible NPE in YAMLExtension for some resource without a

ehsavoie:
add operation.

 * If an ADD operation isn't defined and the resource doesn't have attributes then we will get a NPE

Issue: https://issues.redhat.com/browse/WFCORE-6477.


bstansberry (approved):

bstansberry:
Thanks @ehsavoie

PR #5659, opened 2023-09-14
[WFCORE-6383] throwing boot errors if super.boot fails

istudens:
Issue: https://issues.redhat.com/browse/WFCORE-6383

With this fix the `ModelTestControllerService.boot()` always throws new `BootOperationFailedException` if `super.boot` returns `false`.
Note that a few test-cases had to be changed according to it.
Regarding "BTW, I think the entire class can and should be package protected, but if there's some outside use I've missed, then never mind.", it is not possible as `BootErrorCollector` is also used outside the package.
Note 2: This will need a fix in WildFly code base in two PicketLink Subsystem tests. I will create a separate Jira for it.

### DIFF
--- a/controller/src/main/java/org/jboss/as/controller/AbstractControllerService.java
+++ b/controller/src/main/java/org/jboss/as/controller/AbstractControllerService.java
@@ -591,6 +591,10 @@ public abstract class AbstractControllerService implements Service<ModelControll
         return bootErrorCollector;
     }
 
+    protected final ModelNode getBootErrors() {
+        return bootErrorCollector.getErrors();
+    }
+
     protected OperationStepHandler createExtraValidationStepHandler() {
         return null;
     }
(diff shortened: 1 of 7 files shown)


istudens:
@bstansberry can you review this please? thanks

yersan (commented):
Could you prepare a custom integration Job using your topic branches and link them to this PR?

WildFly requires this PR and the other way around, so an integration Job picking up your topic branches will allow us to test both PRs in conjunction before merging them.

Integration Jobs are available at https://ci.wildfly.org/buildConfiguration/WF_WildFlyCoreIntegrationExperiments

istudens:
@yersan Ok, I gave it a try, it is #1192.

istudens:
@yersan The integration job passed successfully, see https://ci.wildfly.org/buildConfiguration/WF_WildFlyCoreIntegrationExperiments/392051

yersan (approved):
Thanks @istudens , looks good to me, we will merge this one as the last PR for the final release.

Could you rebase and fix the conflicts?

istudens:
> Thanks @istudens , looks good to me, we will merge this one as the last PR for the final release.
> 
> Could you rebase and fix the conflicts?

done

PR #5660, opened 2023-09-15
[WFCORE-6493] Add Guava as a test dependency to the Test Suite.

ivassile:
Issue: https://issues.redhat.com/browse/WFCORE-6493


bstansberry:
Thanks @ivassile

PR #5670, opened 2023-09-19
[WFCORE-6502] Fix variable configuration on common.conf.bat are ignored

yersan:
Jira issue: https://issues.redhat.com/browse/WFCORE-6502

Remove the usage of setLocal/endLocal on common.bat that makes any variable configuration on common.conf.bat are ignored since the environment is restored to the original one after loading common.conf.bat.


ropalka (approved):

yersan:
Thanks for reviewing this, Richard !

PR #5672, opened 2023-09-20
WFCORE-6505 Avoid creating duplicate thread groups on server reload

TomasHofman:
https://issues.redhat.com/browse/WFCORE-6505

This is wildfly-core part of eliminating duplicate thread groups creation.

I tried to remove the thread groups and instead inject the group name string straight into the name pattern, where it was used by ("%G - ...").

### DIFF
--- a/cli/src/main/java/org/jboss/as/cli/gui/JConsoleCLIPlugin.java
+++ b/cli/src/main/java/org/jboss/as/cli/gui/JConsoleCLIPlugin.java
@@ -187,10 +187,9 @@ public class JConsoleCLIPlugin extends JConsolePlugin {
     }
 
     private ExecutorService createExecutor() {
-        final ThreadGroup group = new ThreadGroup("management-client-thread");
         final ThreadFactory threadFactory = doPrivileged(new PrivilegedAction<JBossThreadFactory>() {
             public JBossThreadFactory run() {
-                return new JBossThreadFactory(group, Boolean.FALSE, null, "%G " + executorCount.incrementAndGet() + "-%t", null, null);
+                return new JBossThreadFactory(ThreadGroupHolder.THREAD_GROUP, Boolean.FALSE, null, "%G " + executorCount.incrementAndGet() + "-%t", null, null);
             }
         });
         return EnhancedQueueExecutor.DISABLE_HINT ?
@@ -245,5 +244,10 @@ public class JConsoleCLIPlugin extends JConsolePlugin {
                 return;
             }
         }
+    }
+
+    // Wrapper class to delay thread group creation until when it's needed.
+    private static class ThreadGroupHolder {
+        private static final ThreadGroup THREAD_GROUP = new ThreadGroup("management-client-thread");
     }
 }
--- a/controller/src/main/java/org/jboss/as/controller/remote/AbstractModelControllerOperationHandlerFactoryService.java
+++ b/controller/src/main/java/org/jboss/as/controller/remote/AbstractModelControllerOperationHandlerFactoryService.java
@@ -88,7 +88,7 @@ public abstract class AbstractModelControllerOperationHandlerFactoryService impl
 
         final ThreadFactory threadFactory = doPrivileged(new PrivilegedAction<JBossThreadFactory>() {
             public JBossThreadFactory run() {
-                return new JBossThreadFactory(new ThreadGroup("management-handler-thread"), Boolean.FALSE, null, "%G - %t", null, null);
+                return new JBossThreadFactory(ThreadGroupHolder.THREAD_GROUP, Boolean.FALSE, null, "%G - %t", null, null);
             }
         });
         if (EnhancedQueueExecutor.DISABLE_HINT) {
@@ -159,6 +159,11 @@ public abstract class AbstractModelControllerOperationHandlerFactoryService impl
 
     protected final ExecutorService getClientRequestExecutor() {
         return clientRequestExecutor;
+    }
+
+    // Wrapper class to delay thread group creation until when it's needed.
+    private static class ThreadGroupHolder {
+        private static final ThreadGroup THREAD_GROUP = new ThreadGroup("management-handler-thread");
     }
 
 }
--- a/deployment-scanner/src/main/java/org/jboss/as/server/deployment/scanner/DeploymentScannerAdd.java
+++ b/deployment-scanner/src/main/java/org/jboss/as/server/deployment/scanner/DeploymentScannerAdd.java
@@ -233,7 +233,7 @@ class DeploymentScannerAdd implements OperationStepHandler {
     static ScheduledExecutorService createScannerExecutorService() {
         final ThreadFactory threadFactory = doPrivileged(new PrivilegedAction<ThreadFactory>() {
             public ThreadFactory run() {
-                return new JBossThreadFactory(new ThreadGroup("DeploymentScanner-threads"), Boolean.FALSE, null, "%G - %t", null, null);
+                return new JBossThreadFactory(ThreadGroupHolder.THREAD_GROUP, Boolean.FALSE, null, "%G - %t", null, null);
             }
         });
         return Executors.newScheduledThreadPool(2, threadFactory);
@@ -302,5 +302,10 @@ class DeploymentScannerAdd implements OperationStepHandler {
         public Set<String> getUnrelatedDeployments(ModelNode owner) {
             return Collections.emptySet();
         }
+    }
+
+    // Wrapper class to delay thread group creation until when it's needed.
+    private static class ThreadGroupHolder {
+        private static final ThreadGroup THREAD_GROUP = new ThreadGroup("DeploymentScanner-threads");
     }
 }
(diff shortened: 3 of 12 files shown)


threads/src/main/java/org/jboss/as/threads/ThreadFactoryResolver.java
```diff
@@ -140,8 +140,12 @@ private ServiceName resolveDefaultThreadFactory(String threadPoolName, ServiceNa
...
             final ServiceName threadFactoryServiceName = threadPoolServiceName.append("thread-factory");
             final ThreadFactoryService service = new ThreadFactoryService();
-            service.setThreadGroupName(getThreadGroupName(threadPoolName));
-            service.setNamePattern("%G - %t");
+
+            // Do not set service#threadGroupName, make the group name part of the pattern instead. Setting the
+            // service#threadGroupName would lead to a ThreadGroup being created each time the service is restarted.
+            String threadGroupName = getThreadGroupName(threadPoolName);
+            service.setNamePattern(threadGroupName + " - %t");
+
```
TomasHofman, line 148: I think this is the most problematic part - related to the ThreadFactoryResourceDefinition / ThreadFactoryService.

This resource allows to create custom thread factory, and allows user to define:
* thread group name,
* thread priority
* thread name pattern (which can use %G to emit thread group name).

The thread priority is not in fact set through a created ThreadGroup instance but directly on newly created Threads, so the ThreadGroup still doesn't have much meaning.

Above is a workaround - do not set the threadGroupName on the ThreadFactoryService, so no ThreadGroup is going to be created, instead use the thread group name as part of the name pattern.

If we have a freedom change the resource model, we could remove the threadGroupName attribute from the ThreadFactoryResource and just leave the name pattern and priority attributes, which still allows for the same end effect (thread group name can be "hard-coded" part of the pattern).

yersan:
An ongoing discussion about this is available at: https://lists.jboss.org/archives/list/wildfly-dev@lists.jboss.org/thread/2NICNYJE4NO2HK67DOFTPZUGNAKD5RRM/

threads/src/main/java/org/jboss/as/threads/ThreadFactoryResolver.java
bstansberry: @TomasHofman What you've done here is fine, except please add @Deprecated(forRemoval="true") to AbstractThreadFactoryResolver.getThreadGroupName

Also, I think ThreadFactoryService should cache any ThreadGroup created in start in an instance field and call ThreadGroup.destroy() on it in stop().  Catch and ignore or DEBUG log any RuntimeException thrown from destroy(). If whatever service injected the ThreadFactoryService didn't clean up in its own stop() the threads it created, the destroy() call will fail, but oh well. Lets not fail or make log noise but we can debug log. If they did clean up destroy() will clean up the parent TG ref to the group.

The various 'xxxThreadPoolService' classes in this module clean up their threads correctly.

yersan: @bstansberry @TomasHofman `threadGroup.destroy()` is deprecated for removal on JDK 17 so I discourage their usage. I agree with moving it to a  variable instance since there is a 1:1 relationship between the service and the ThreadGroup, at first glance, I am expecting that a service reload can reuse the same ThreadGroup.

I commented on the JBoss List as well, the removal of the ThreadsGroups, although it looks like the way to follow since this class has been degraded in JDK, seems that it could be helpful for debugging purposes, so I am more inclined to keep them on the services that require it as a static field instead of remove them altogether. At least, don't do it until there is any other available option on the JDK to group them.

Thoughts?

TomasHofman: Hi @yersan, I did see the ticket you referred on the mailing list (https://bugs.openjdk.org/browse/JDK-8252885), it mentions some debugging related APIs like JDWP, JVM TI and JDI which I think is what you meant. It's sort of like these APIs contains some methods that lets you enumerate threads and subgroups belonging to a ThreadGroup.

My perspective is that we would not loose anything that much important. If you can't filter by groups you can still filter by name. The tools that I use don't support this anyway, or I didn't notice because I never felt the need to use it... But I know this is very subjective statement.

Regarding keeping an instance field in the `ThreadFactoryService` for created group, I think that only makes sense if we want to call the destroy method on it. The `ThreadFactoryService` instances are created fresh after each app server reload, so it won't help with reusing the original groups. If we want to reuse the original groups we would have to keep them in some some static map with service name or something like that as a key, or we would need to keep the `ThreadFactoryService` instance itself in a static field wherever they are created (and then have the instance field for the group).

The destroy method is still present in the JDK 21, I'm now checking, but it is an empty method. I'm slightly in favor of not using it, but I will go by what you guys like most.

yersan:
> I did see the ticket you referred on the mailing list (https://bugs.openjdk.org/browse/JDK-8252885), it mentions some debugging related APIs like JDWP, JVM TI and JDI which I think is what you meant. 

Right @TomasHofman, I wanted to highlight such feedback since developers could still see useful continue getting the threads grouped.

> The ThreadFactoryService instances are created fresh after each app server reload, so it won't help with reusing the original groups. If we want to reuse the original groups we would have to keep them in some static map with service name or something like that as a key, or we would need to keep the ThreadFactoryService instance itself in a static field wherever they are created (and then have the instance field for the group).

Thanks, right, I see what you mean. Looking at it I have also noticed that the Threads extension was deprecated and converted into a read-only extension to give support on mixed domains. However, that was even much before WF23, and we do not support mixed domains below WF23. 


@bstansberry Should not we remove completely this extension?

If the above is true, by moving the Threads extension outside of the discussion, we can:
1. If the others can use static ThreadGroups to avoid the memory leak, use static versions to keep the possible usability for debugging purposes.
2. We can go ahead with removing them completely as proposed, and just use the ThreadGroup name to identify the threads.

So far I am still in favor of 1. if possible. I have not checked each of the affected services yet.

TomasHofman:
@yersan what do you mean exactly by "Threads extension" and it being deprecated? Where can I see that info?

As far as I can see the classes (ThreadFactoryResolver erc) are being actively used by several subsystems (Batch, EJB), even in standalone mode.

TomasHofman:
I probably need to retract my proposal to keep thread groups statically :).

The thread-factory resource allows to modify thread group name, in which case you need to create new ThreadGroups...

```
/subsystem=batch-jberet/thread-factory=tf:add(group-name=name1)
/subsystem=batch-jberet/thread-pool=batch:write-attribute(name=thread-factory, value=tf)
/subsystem=batch-jberet/thread-factory=tf:write-attribute(name=group-name, value=name2)
```

yersan:
> @yersan what do you mean exactly by "Threads extension" and it being deprecated? Where can I see that info?

See [ThreadsExtension](https://github.com/wildfly/wildfly-core/blob/035d41260346840f0073a90875997f2b8f16df17/threads/src/main/java/org/jboss/as/threads/ThreadsExtension.java#L37)

> As far as I can see the classes (ThreadFactoryResolver erc) are being actively used by several subsystems (Batch, EJB), even in standalone mode.

Ok, I need to be more familiarized with this subsystem to give a more accurate opinion; I only saw that `ThreadFactoryResolver` was inside of the `org.jboss.as.threads` and I related it directly with the Threads Extension supplied by this module.

threads/src/main/java/org/jboss/as/threads/ThreadFactoryResolver.java
bstansberry: I'll defer to @yersan re the static ThreadGroups.

I don't see anything useful IDEA does with them; it just displays the TG name as part of its string representation of a thread. I don't see any filtering or sorting feature.  I pinged @ehsavoie to ask if NetBeans does anything.

JConsole also doesn't do anything (i.e. the platform mbeans don't show TG data.)

If destroy() is a no-op in SE 21, then +1 to not using it. It being deprecated in 17 is good enough reason to drop it. And I agree if we don't call destroy() using an instance field does nothing. So for that we either live with the leak or add a static WeakHashMap<String, ThreadGroup> to cache them.

ehsavoie: NetBeans shows threads grouped under their thread group in the debug window

yersan: In eclipse IDE you can group the threads by ThreadGroup in your debug view, so hence the utility for keeping the ThreadGroups if possible but removing the leak.

TomasHofman:
PR updated - thread groups are now kept in static fields.

bstansberry:
@yersan The threads module at this point is essentially a utility library used by other subsystems (Batch, EJB) to provide configuration of thread pools and factories within their own config. ThreadFactoryService and its ThreadGroup creation is relevant for that reason.

Some of what's in the module could be removed, yes, i.e. anything not related to its use as a utility library.

yersan (approved):

yersan:
Thanks @TomasHofman and @bstansberry / @ehsavoie for helping with this

PR #5677, opened 2023-09-22
[WFCORE-6519] Enabling logging tests because they no longer fail neither on JDK17 not on JDK21

ropalka:
https://issues.redhat.com/browse/WFCORE-6519


yersan (approved):

yersan:
I've approved this since the change is pretty clear: it just removed the condition to skip the test under JDK12. However, I've noticed the surefire plugin is configured to [include only](https://github.com/wildfly/wildfly-core/blob/3fd7d3ab65630fda59bd6dca9a56c6128fcb229d/testsuite/manualmode/pom.xml#L258-L261) tests with the TestCase suffix, so `AuditLogBootingSyslogTest` is never executed here.

I'm going to file a new Jira to rename this test, so far I don't see any condition that requires it and it looks like it was unintentional.

yersan:
Thanks @ropalka

PR #5688, opened 2023-09-26
[WFCORE-6527]: Upgrade JGit from 6.6.0.202305301015-r to 6.7.0.202309050840-r

ehsavoie:
* Upgrading jgit version:https://projects.eclipse.org/projects/technology.jgit/releases/6.7.0
* Explicit dependency on common-codec

Issue: https://issues.redhat.com/browse/WFCORE-6527

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -198,7 +198,7 @@
         <version.org.apache.velocity>2.3</version.org.apache.velocity>
         <version.org.bouncycastle>1.76</version.org.bouncycastle>
         <version.org.codehaus.plexus.plexus-utils>3.5.1</version.org.codehaus.plexus.plexus-utils>
-        <version.org.eclipse.jgit>6.6.1.202309021850-r</version.org.eclipse.jgit>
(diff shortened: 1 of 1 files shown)


pom.xml
```diff
@@ -700,6 +701,11 @@
                 <artifactId>commons-cli</artifactId>
                 <version>${version.commons-cli}</version>
             </dependency>
+            <dependency>
+                <groupId>commons-codec</groupId>
+                <artifactId>commons-codec</artifactId>
+                <version>${version.commons-codec}</version>
+            </dependency>
```
yersan, line 708: @ehsavoie Have you tested whether it is just possible to exclude `commons-codec` transitive dependency from  `org.eclipse.jgit`? It seems it is not used at runtime (it has not been declared for org.eclipse.jgit module.xml) and we do not compile `org.eclipse.jgit`, so I guess excluding it would be enough.

yersan (approved):

PR #5702, opened 2023-10-05
Fix for WFCORE-6486, CLIEmbedHostControllerTestCase fails randomly with JDK 17 Jobs

jfdenise:
Issue: https://issues.redhat.com/browse/WFCORE-6486

The test seems the culprit there, we should introduce some timeout prior to assert that the trace has not been printed.

### DIFF
--- a/testsuite/manualmode/src/test/java/org/jboss/as/test/manualmode/management/cli/CLIEmbedHostControllerTestCase.java
+++ b/testsuite/manualmode/src/test/java/org/jboss/as/test/manualmode/management/cli/CLIEmbedHostControllerTestCase.java
@@ -758,8 +758,27 @@ public class CLIEmbedHostControllerTestCase extends AbstractCliTestBase {
     }
 
     private void checkLogging(String line) throws IOException {
-        String logOutput = readLogOut();
-        assertTrue(logOutput, checkLogging(logOutput, line));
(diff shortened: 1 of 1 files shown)


testsuite/manualmode/src/test/java/org/jboss/as/test/manualmode/management/cli/CLIEmbedHostControllerTestCase.java
```diff
@@ -758,8 +758,27 @@ private String readLogOut() {
     }
 
     private void checkLogging(String line) throws IOException {
-        String logOutput = readLogOut();
-        assertTrue(logOutput, checkLogging(logOutput, line));
+        long delay = System.currentTimeMillis() + 10000;
```
yersan, line 761: Would you mind using `TimeoutUtil.adjust(1000)`? 

That's more useful since it allows us to control all timeouts from a system property and use more time in a slow system.

jfdenise: Thanks, fixed.

yersan (approved):

yersan:
Thanks @jfdenise

PR #5705, opened 2023-10-10
[WFCORE-6548] Unify test logging in elytron module

OndrejKotek:
https://issues.redhat.com/browse/WFCORE-6548

### DIFF
--- a/elytron/src/test/resources/logging.properties
+++ b/elytron/src/test/resources/logging.properties
@@ -7,7 +7,7 @@
 loggers=org.wildfly.extension.elytron,org.apache.directory,net.sf.ehcache,org.apache.directory.api.ldap,javax.security.sasl,org.wildfly.security,org.wildfly.security.sasl,org.wildfly.security.xml,org.wildfly.security.tls
 
 # Root logger configuration
-logger.level=${test.level:TRACE}
+logger.level=INFO
 logger.org.apache.directory.level=INFO
 logger.org.apache.directory.api.ldap.level=OFF
 logger.net.sf.ehcache.level=INFO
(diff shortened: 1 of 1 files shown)


elytron/src/test/resources/logging.properties
```diff
@@ -7,7 +7,7 @@
 loggers=org.wildfly.extension.elytron,org.apache.directory,net.sf.ehcache,org.apache.directory.api.ldap,javax.security.sasl,org.wildfly.security,org.wildfly.security.sasl,org.wildfly.security.xml,org.wildfly.security.tls
 
 # Root logger configuration
-logger.level=${test.level:TRACE}
+logger.level=INFO
```
yersan, line 10: Maybe the Elytron team is used to using this config `-Dtest.level`

@OndrejKotek is there any reason for not using `logger.level=${test.level:INFO}` instead?

OndrejKotek: I have taken other similar logging configurations into account and tried to find the best solution. I think the levels of the most important loggers are set below and the logging level for Elytron tests is handled through `handler.CONSOLE.level=${test.level:INFO}`.

@fjuma @Skyllarr WDYT please?

fjuma: @OndrejKotek Thanks for the explanation. This seems ok then. @yersan WDYT?

yersan: Thanks for reviewing @fjuma. I am fine with this too then.

The change here is now only the loggers declared in the `loggers` configuration will be taken into account by the `CONSOLE` handler when the level changes to something different than `INFO`. The root logger will be always `INFO`. That's more clear handling since it takes into account only the loggers you want for your tests avoiding the root logger.

yersan (approved):

elytron/src/test/resources/logging.properties
yersan: Threads and installation-manager could use the same approach, that's a different Jira, I will create one

yersan:
Thanks @fjuma @OndrejKotek

PR #5706, opened 2023-10-10
[WFCORE-6532] Add security manager to script tests

xjusko:
https://issues.redhat.com/browse/WFCORE-6532

### DIFF
--- a/testsuite/scripts/src/test/java/org/wildfly/scripts/test/DomainScriptTestCase.java
+++ b/testsuite/scripts/src/test/java/org/wildfly/scripts/test/DomainScriptTestCase.java
@@ -6,6 +6,9 @@
 package org.wildfly.scripts.test;
 
 import java.io.IOException;
+import java.util.Collection;
+import java.util.List;
+import java.util.Map;
 import java.util.concurrent.Callable;
 import java.util.concurrent.TimeoutException;
 import java.util.function.Function;
@@ -15,22 +18,32 @@ import org.jboss.as.controller.client.helpers.Operations;
 import org.jboss.as.test.shared.TestSuiteEnvironment;
 import org.jboss.dmr.ModelNode;
 import org.junit.Assert;
+import org.junit.runner.RunWith;
+import org.junit.runners.Parameterized;
 import org.wildfly.common.test.ServerHelper;
 
 /**
  * @author <a href="mailto:jperkins@redhat.com">James R. Perkins</a>
  */
+@RunWith(Parameterized.class)
 public class DomainScriptTestCase extends ScriptTestCase {
 
     private static final Function<ModelControllerClient, Boolean> HOST_CONTROLLER_CHECK = ServerHelper::isDomainRunning;
 
+    @Parameterized.Parameter
+    public Map<String, String> env;
     public DomainScriptTestCase() {
         super("domain");
     }
 
+    @Parameterized.Parameters
+    public static Collection<Object> data() {
+        return List.of(Map.of(), Map.of("SECMGR", "true"));
+    }
+
     @Override
     void testScript(final ScriptProcess script) throws InterruptedException, TimeoutException, IOException {
-        script.start(HOST_CONTROLLER_CHECK, ServerHelper.DEFAULT_SERVER_JAVA_OPTS);
+        script.start(HOST_CONTROLLER_CHECK, env, ServerHelper.DEFAULT_SERVER_JAVA_OPTS);
 
         Assert.assertNotNull("The process is null and may have failed to start.", script);
         Assert.assertTrue("The process is not running and should be", script.isAlive());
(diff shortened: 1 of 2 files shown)


testsuite/scripts/src/test/java/org/wildfly/scripts/test/DomainScriptTestCase.java
```diff
@@ -15,22 +20,42 @@
...
+        final Collection<Object> result = new ArrayList<>(2);
+        result.add(Collections.emptyMap());
+        result.add(Collections.singletonMap("SECMGR", "true"));
+        return result;
+    }
+
     @Override
     void testScript(final ScriptProcess script) throws InterruptedException, TimeoutException, IOException {
-        script.start(HOST_CONTROLLER_CHECK, ServerHelper.DEFAULT_SERVER_JAVA_OPTS);
+        boolean isSecurityManagerEnabled = Boolean.parseBoolean(System.getProperty("security.manager"));
```
jamezp, line 51: IMO we should skip this check and just always test with the security manager enabled in one run. Currently we don't run security manager tests on Windows so we'd lose that testing.

testsuite/scripts/src/test/java/org/wildfly/scripts/test/DomainScriptTestCase.java
```diff
@@ -15,22 +20,42 @@
...
     public DomainScriptTestCase() {
         super("domain");
     }
 
+    @Parameterized.Parameters
+    public static Collection<Object> data() {
+        final Collection<Object> result = new ArrayList<>(2);
+        result.add(Collections.emptyMap());
+        result.add(Collections.singletonMap("SECMGR", "true"));
+        return result;
```
jamezp, line 46: This could simply be `return List.of(Map.of(), Map.of("SECMGR", "true"));`.

testsuite/scripts/src/test/java/org/wildfly/scripts/test/StandaloneScriptTestCase.java
```diff
@@ -89,7 +91,13 @@ void testScript(final ScriptProcess script) throws InterruptedException, Timeout
         // seem to work when a directory has a space. An error indicating the trailing quote cannot be found. Removing
         // the `\ parts and just keeping quotes ends in the error shown in JDK-8215398.
         Assume.assumeFalse(TestSuiteEnvironment.isWindows() && env.containsKey("GC_LOG") && script.getScript().toString().contains(" "));
-        script.start(STANDALONE_CHECK, env, ServerHelper.DEFAULT_SERVER_JAVA_OPTS);
+
+        boolean isSecurityManagerEnabled = Boolean.parseBoolean(System.getProperty("security.manager"));
```
jamezp, line 95: Same here, we should always do a run with the security manager enabled.

jamezp (changes requested):

testsuite/scripts/src/test/java/org/wildfly/scripts/test/DomainScriptTestCase.java
```diff
@@ -15,22 +19,34 @@
...
+    @Parameterized.Parameters
+    public static Collection<Object> data() {
+        return List.of(Map.of(), Map.of("SECMGR", "true"));
+    }
+
     @Override
     void testScript(final ScriptProcess script) throws InterruptedException, TimeoutException, IOException {
-        script.start(HOST_CONTROLLER_CHECK, ServerHelper.DEFAULT_SERVER_JAVA_OPTS);
+        String[] modifiedJavaOpts = Arrays.copyOf(ServerHelper.DEFAULT_SERVER_JAVA_OPTS, ServerHelper.DEFAULT_SERVER_JAVA_OPTS.length + 1);
+        modifiedJavaOpts[modifiedJavaOpts.length - 1] = "-secmgr";
```
jamezp, line 48: Sorry, I was clear with my wording here. We shouldn't be adding the `-secmgr` parameter to the server arguments if we've already got it set in the environment variable. Essentially we could just do this:
```java
script.start(HOST_CONTROLLER_CHECK, env, ServerHelper.DEFAULT_SERVER_JAVA_OPTS);
```

jamezp (changes requested):

xjusko:
I see, no problem, it should be fixed now.
Thanks for your help.

testsuite/scripts/src/test/java/org/wildfly/scripts/test/StandaloneScriptTestCase.java
```diff
@@ -8,9 +8,8 @@
 import java.io.IOException;
 import java.nio.file.Files;
 import java.nio.file.Path;
-import java.util.ArrayList;
+import java.util.Arrays;
```
jamezp, line 11: This is also causing a checkstyle issue and should be removed.

jamezp (changes requested):

jamezp (approved):

yersan:
Thanks @xjusko @jamezp

PR #5721, opened 2023-10-12
[WFCORE-6559] Allow Remove-Item command to delete directories with lo…

yersan:
…ng paths

Jira issue: https://issues.redhat.com/browse/WFCORE-6559

### DIFF
--- a/core-feature-pack/common/src/main/resources/content/bin/installation-manager.ps1
+++ b/core-feature-pack/common/src/main/resources/content/bin/installation-manager.ps1
@@ -7,6 +7,21 @@ param (
     [string]$instMgrLogProperties
 )
 
+# Sanitizes URLs to avoid issues with unsupported long paths
+function Sanitize-Path {
+    param(
+        [Parameter(Mandatory=$true)]
+        [string]$inputPath
+    )
+
+    # Check if the path starts with double backslashes and contains at least one additional character
+    if ($inputPath -match '^\\\\[^\\]+\\.*') {
+        $removedBackslash = $inputPath.Substring(2)
+        return "\\?\UNC\$removedBackslash"
+    } else {
+        return "\\?\$inputPath"
+    }
+}
 
 # For security, reset the environment variables first
 Set-Variable -Name INST_MGR_COMMAND -Scope Script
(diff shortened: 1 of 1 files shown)


bstansberry:
@yersan What about UNC?

From https://learn.microsoft.com/en-ca/windows/win32/fileio/maximum-file-path-limitation?tabs=registry

```
The "\\?\" prefix can also be used with paths constructed according to the universal naming convention (UNC). To specify such a path using UNC, use the "\\?\UNC\" prefix. For example, "\\?\UNC\server\share", where "server" is the name of the computer and "share" is the name of the shared folder.
```

yersan:
Hi @bstansberry It could be a possibility and if so, we can include it here. The `INST_MGR_PREPARED_SERVER_DIR` used by the script is computed from the result of `pathManager.getPathEntry("jboss.controller.temp.dir").resolvePath()`.

If the PathManager is able to return such a kind of path, then we could include the UNC here. I am going to verify this possibility first.

yersan:
It is possible to use UNC paths for configuring the temp dir, so we should allow deleting UNC paths as well. Moved to draft

darranl:
@jamezp just mentioning you as a reviewer as I think you said you could look at this one, other reviews still welcome.

yersan:
Maybe @spyrkob ? I did manual tests here and it worked on my end, so it is a matter of putting other eyes on it in case we are missing something.

spyrkob:
@yersan what about `Test-Path -Path $INST_MGR_PREPARED_SERVER_DIR -PathType Container` and `Get-ChildItem -Path $INST_MGR_PREPARED_SERVER_DIR` earlier on in the script? Do they also need handle long paths?

yersan:
Hi @spyrkob , thanks for the review. It is unlikely this case could occur just with the `$INST_MGR_PREPARED_SERVER_DIR` and its immediate children, the issue is more probable when we have created a full server directory structure over the `$INST_MGR_PREPARED_SERVER_DIR`. However, the corner case could be there. I haven't tested if `Test-Path -Path` or `Get-ChildItem -Path` could hit the same error as `Remove-Item`. I need to check this case.

yersan:
@spyrkob I took into account your point, could you review it again?

spyrkob (approved):
Thanks @yersan, looks good as far as I can tell

marekkopecky:
MR LGTM, see more details on jira issue

yersan:
Thanks to all for collaborating on the review!

PR #5724, opened 2023-10-16
[WFCORE-6560] Log installation provisioning information at boot

bstansberry:
https://issues.redhat.com/browse/WFCORE-6560

Draft for now what we log is expected to change. This just dumps what's currently available with no particular attempt at formatting.

@yersan @spyrkob FYI

### DIFF
--- a/installation-manager/src/main/java/org/wildfly/core/instmgr/InstMgrService.java
+++ b/installation-manager/src/main/java/org/wildfly/core/instmgr/InstMgrService.java
@@ -5,11 +5,14 @@
 
 package org.wildfly.core.instmgr;
 
+import static org.wildfly.core.instmgr.logging.InstMgrLogger.ROOT_LOGGER;
+
 import java.io.File;
 import java.io.IOException;
 import java.nio.file.Files;
 import java.nio.file.Path;
 import java.nio.file.Paths;
+import java.util.Collection;
 import java.util.Comparator;
 import java.util.Iterator;
 import java.util.Map;
@@ -28,12 +31,18 @@ import org.jboss.msc.service.StartContext;
 import org.jboss.msc.service.StartException;
 import org.jboss.msc.service.StopContext;
 import org.wildfly.core.instmgr.logging.InstMgrLogger;
+import org.wildfly.installationmanager.ManifestVersion;
+import org.wildfly.installationmanager.MavenOptions;
+import org.wildfly.installationmanager.spi.InstallationManager;
+import org.wildfly.installationmanager.spi.InstallationManagerFactory;
 
 /**
  * This is the main service used by the installation manager management operation handlers.
  */
 class InstMgrService implements Service {
     private static final Logger LOG = Logger.getLogger(InstMgrService.class);
+
+    private final InstallationManagerFactory imf;
     private final Supplier<PathManager> pathManagerSupplier;
     private final Consumer<InstMgrService> consumer;
     private final Supplier<ExecutorService> executorSupplier;
@@ -48,7 +57,8 @@ class InstMgrService implements Service {
     private final InstMgrCandidateStatus candidateStatus;
     private ExecutorService executor;
 
-    InstMgrService(Supplier<PathManager> pathManagerSupplier, Supplier<ExecutorService> executorSupplier, Consumer<InstMgrService> consumer) {
+    InstMgrService(InstallationManagerFactory imf, Supplier<PathManager> pathManagerSupplier, Supplier<ExecutorService> executorSupplier, Consumer<InstMgrService> consumer) {
+        this.imf = imf;
         this.pathManagerSupplier = pathManagerSupplier;
         this.candidateStatus = new InstMgrCandidateStatus();
         this.executorSupplier = executorSupplier;
@@ -83,6 +93,9 @@ class InstMgrService implements Service {
         } catch (IOException e) {
             throw new StartException(e);
         }
+
+        logInstallationStatus();
+
         started.set(true);
         this.consumer.accept(this);
     }
@@ -194,5 +207,26 @@ class InstMgrService implements Service {
 
     public ExecutorService getMgmtExecutor() {
         return executor;
+    }
+
+
+
+    private void logInstallationStatus() {
+        if (ROOT_LOGGER.isInfoEnabled()) {
+            MavenOptions mavenOptions = new MavenOptions(null, false);
+            InstallationManager installationManager = null;
+            Collection<ManifestVersion> channels = null;
+            try {
+                installationManager = imf.create(homeDir, mavenOptions);
+                channels = installationManager.getInstalledVersions();
+                ROOT_LOGGER.provisioningChannels(channels);
+            } catch (Exception e) {
+                if (installationManager == null) {
+                    ROOT_LOGGER.failedToCreateInstallationManager(homeDir, e);
+                } else if (channels == null) {
+                    ROOT_LOGGER.failedToFindInstallationChannels(e);
+                }
+            }
+        }
     }
 }
--- a/installation-manager/src/main/java/org/wildfly/core/instmgr/logging/InstMgrLogger.java
+++ b/installation-manager/src/main/java/org/wildfly/core/instmgr/logging/InstMgrLogger.java
@@ -5,13 +5,17 @@
 
 package org.wildfly.core.instmgr.logging;
 
+import java.nio.file.Path;
+import java.util.Collection;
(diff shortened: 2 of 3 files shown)


spyrkob:
I also added an SPI operation to expose the latest manifest name in https://github.com/wildfly-extras/wildfly-installation-manager-api/pull/18 and https://github.com/wildfly-extras/prospero/pull/481

bstansberry:
@yersan Will this stuff work in embedded servers / hosts? I have a vague recollection you disabled it for embedded just to narrow the scope of your work on this.

I'm not asking for any change; I'm just refreshing my memory. :) Yesterday I thought about installing the service even on embedded and just not registering the Resource and ResourceDefinition. But then I decided I wanted to keep the scope narrow for this logging as well. Logging this for embedded is a corner case we can come back to if there is demand.

installation-manager/src/main/java/org/wildfly/core/instmgr/InstMgrService.java
```diff
@@ -195,4 +211,32 @@ Path getControllerTempDir() {
...
+            InstallationManager installationManager = null;
+            Collection<Channel> channels = null;
+            List<HistoryResult> history = null;
+            try {
+                installationManager = imf.create(homeDir, mavenOptions);
+                channels = installationManager.listChannels();
+                CONFIG_LOGGER.provisioningChannels(channels);
+                if (CONFIG_LOGGER.isDebugEnabled()) {
+                    history = installationManager.history();
+                    CONFIG_LOGGER.provisioningHistory(history);
```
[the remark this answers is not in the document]
bstansberry: @yersan I added a commit dropping that history logging. And without that verbose logging don't see any need for using the org.jboss.as.config logger.

bstansberry:
> I also added an SPI operation to expose the latest manifest name in [wildfly-extras/wildfly-installation-manager-api#18](https://github.com/wildfly-extras/wildfly-installation-manager-api/pull/18) and [wildfly-extras/prospero#481](https://github.com/wildfly-extras/prospero/pull/481)

@spyrkob I didn't approve those as they are draft PRs, but they LGTM.

yersan:
> Will this stuff work in embedded servers / hosts? I have a vague recollection you disabled it for embedded just to narrow the scope of your work on this.

Hi @bstansberry , good point there. The tool is not available there, narrowing the scope was one reason. In general:

1. The focus of the integration has been always the update/revert of a remote installation via CLI, we do not use embedded server/hosts to configure a remote installation.
2. For any kind of local installation, you should use Prospero.
3. You need to restart the server to perform the installation, but we do not have such a possibility in an embedded server.

Although it could be possible to circumvent the above points and allow the use of the CLI integration, for simplicity and the first cut, we disabled the usability of the installation manager in the embedded server/host.

Thinking on the embedded, as soon as we are running in a modular environment and Prospero is installed with the server, we should be able to use the factory to create an instance to get any kind of version.

We can continue avoiding registering the management resources when using the embedded process type to continue being compliant with the above points.

bstansberry:
Thanks @yersan. I think for this part I'll describe having this logging in an embedded process as a 'Nice to Have', and move to 'Future Work' if we don't get to it (which is likely.)

It's easy to tweak this to support embedded, but testing it is more work, and if we don't test it having it adds risk.

bstansberry:
FYI, it turns out testing this will have to happen in full WF, as the org.wildfly.prospero:wildfly-galleon-pack depends on Jackson modules only provided by the wildfly-ee FP. So I'll submit a draft component-upgrade-needed wildfly/wildfly PR for those tests, link it here and link some custom integration test runs. I'll move it out of draft when that is done.

yersan:
ok, in case it helps and if you see it useful, our testing strategy for Installation Manager in WildFly Core has been using a mocked integration. It means tests in WildFly Core for the installation manager are testing until the API method invocation . Prospero also has an extended test suite that also tests the internal invocation of their own implementation. You could just use a mocked version and return something to test the functionality. Not sure if that would cope with your requirements, but that would avoid changes in WildFly full.

bstansberry:
@spyrkob To test this I'll need a release of Prospero with ProsperoInstallationManager.getInstalledVersions() implemented.

spyrkob:
@bstansberry prospero 1.2.1.Final with the updated InstallationManager has just been released

bstansberry:
I've added the test in full WF

https://github.com/wildfly/wildfly/pull/17470

Here are custom runs exercising this PR and the test PR in combination:

https://ci.wildfly.org/viewLog.html?buildId=406756&tab=buildResultsDiv&buildTypeId=WF_WildFlyCoreIntegrationExperiments

https://ci.wildfly.org/viewLog.html?buildId=406758&tab=buildResultsDiv&buildTypeId=WF_WildFlyCoreIntegrationExperimentsWindows

bstansberry:
> ok, in case it helps and if you see it useful, our testing strategy for Installation Manager in WildFly Core has been using a mocked integration. It means tests in WildFly Core for the installation manager are testing until the API method invocation. Prospero also has an extended test suite that also tests the internal invocation of their own implementation. You could just use a mocked version and return something to test the functionality. Not sure if that would cope with your requirements, but that would avoid changes in WildFly full.

Thanks, @yersan -- that's very helpful. I've done exactly that in this PR, so there's no need for a test in full WF. I missed your comment and wrote a test in full WF anyway, but I don't want it merged anymore, as core is where such tests belong.

So, assuming the current CI runs are ok, this is ready to go.

installation-manager/src/main/java/org/wildfly/core/instmgr/logging/InstMgrLogger.java
```diff
@@ -81,6 +86,18 @@ public interface InstMgrLogger extends BasicLogger {
     @Message(id = 20, value = "No custom patches installed found for the specified manifest maven coordinates: '%s'")
     OperationFailedException noCustomPatchFound(String manifestGA);
 
+    @LogMessage(level = Logger.Level.INFO)
+    @Message(id = 21, value = "Installation provisioning is using the following channels: '%s'")
```
spyrkob, line 90: Nitpick, but I wonder if this should read something like "Installation provisioned using the following channel states/(versions?):". The ManifestVersions provide information about the latest used versions from each channel rather then channel information like manifest location, repositories etc.

testsuite/manualmode/pom.xml
```diff
@@ -632,6 +633,8 @@
                                         <exclude>org.wildfly.core.test.standalone.mgmt.events.*TestCase</exclude>
                                         <!-- Requires modification of the process-uuid file inside the Jar -->
                                         <exclude>org.jboss.as.test.manualmode.logging.ProcessUuidTestCase</exclude>
+                                        <!-- Validates content of traditional unzipped installation. (Works, but is irrelevant to bootable jar testing.) -->
+                                        <exclude>org.jboss.as.test.manualmode.provisioning.ProvisioningConsistencyTestCase</exclude>
```
yersan, line 637: This looks unrelated to these changes, isn't it? I am not against it, just trying to link the points.

yersan (commented):
I'm just waiting for the minor pending questions and getting the feature progressing more before approving, but overall LGTM

testsuite/manualmode/pom.xml
bstansberry: Yes, it's unrelated. I have a bunch of branches in this general 'test provisioning' topic area and I got tired of switching back and forth. :)

yersan: Ok, thank you for confirming @bstansberry , I also wanted to be sure it was not unintentional

installation-manager/src/main/java/org/wildfly/core/instmgr/logging/InstMgrLogger.java
bstansberry: Thanks @spyrkob -- good point. I went with "Installation was provisioned using the following channel versions:".

bstansberry:
@yersan @spyrkob This change to WF isn't really a feature and doesn't need to be handled as such. So I changed the JIRA to Enhancement and removed the 'Feature' etc labels from this PR. So if you all are ok with approving it we can get it merged and out of the WIP queue.

Downstream it's part of a larger requirement set, but that's not relevant to how we need to handle it in WF.

spyrkob (approved):
LGTM

yersan:
Thanks @bstansberry @spyrkob

bstansberry:
A test I meant to add to this PR but neglected to push can be found at #5967

PR #5733, opened 2023-10-20
[WFCORE-6476-fix]: Reducing logging level for failed internal read-on…

ehsavoie:
…ly operation steps.

Jira issue: https://issues.redhat.com/browse/WFCORE-6476

 * Fixing missing case

### DIFF
--- a/controller/src/main/java/org/jboss/as/controller/AbstractOperationContext.java
+++ b/controller/src/main/java/org/jboss/as/controller/AbstractOperationContext.java
@@ -1072,7 +1072,11 @@ abstract class AbstractOperationContext implements OperationContext, AutoCloseab
             }
         } catch (Throwable t) {
             // Handling for throwables that don't implement OperationClientException marker interface
-            MGMT_OP_LOGGER.operationFailed(t, step.operation.get(OP), step.operation.get(OP_ADDR));
+            if (!step.requiresDoneStage && !isExternalClient()) {
+                logStepFailure(step, false);
+            } else {
+                MGMT_OP_LOGGER.operationFailed(t, step.operation.get(OP), step.operation.get(OP_ADDR));
+            }
 
             // Provide a failure description if there isn't one already
             if (!step.hasFailed()) {


controller/src/main/java/org/jboss/as/controller/AbstractOperationContext.java
```diff
@@ -1072,7 +1072,7 @@ private void executeStep(final Step step) {
             }
         } catch (Throwable t) {
             // Handling for throwables that don't implement OperationClientException marker interface
-            MGMT_OP_LOGGER.operationFailed(t, step.operation.get(OP), step.operation.get(OP_ADDR));
+            logStepFailure(step, false);
```
yersan, line 1075: I am not sure why this case was left from the previous initial work. However, it makes sense to me. 

I can only see two consequences:

1. Errors arriving at this point when the stage is `Stage.DOMAIN`, will no longer logged, but @brian added this comment on his previous change https://github.com/wildfly/wildfly-core/blob/035d41260346840f0073a90875997f2b8f16df17/controller/src/main/java/org/jboss/as/controller/AbstractOperationContext.java#L1132-L1133, so probably ignoring them would be fine.

If even with that we are not sure and want to be conservative, we could ask for the stage here and log them in such a case. For example
```
if (currentStage != Stage.DOMAIN) {
   logStepFailure(step, false);
} else {
MGMT_OP_LOGGER.operationFailed(t, step.operation.get(OP), 
step.operation.get(OP_ADDR));
}
```

2. In the worst case, this change will change some specific logs from ERROR to DEBUG and will not change specific server behavior.

If there are no other thoughts, I would approve it as it is.

yersan: Another possible more conservative alternative is to use the new method only for the internal callers and left anything else as it was before:

```
if (!isExternalClient()) {
    logStepFailure(step, false);
} else {
    MGMT_OP_LOGGER.operationFailed(t, step.operation.get(OP), step.operation.get(OP_ADDR));
}
```

@ehsavoie Do the above resolve your issue? It seems it could be a more secure option to avoid miss or change any log-level

ehsavoie: I would also ensure that the op is read-only

yersan:
Thanks @ehsavoie

PR #5768, opened 2023-11-15
[WFCORE-6601][WFCORE-6607][WFCORE-6608] ModuleSpecification dependency tracking fixes

bstansberry:
https://issues.redhat.com/browse/WFCORE-6601
https://issues.redhat.com/browse/WFCORE-6607
https://issues.redhat.com/browse/WFCORE-6608

The expectation is this will also fix https://issues.redhat.com/browse/WFCORE-6606, but resolving that should depend upon a successful TCK run.

### DIFF
--- a/server/src/main/java/org/jboss/as/server/deployment/module/ModuleSpecProcessor.java
+++ b/server/src/main/java/org/jboss/as/server/deployment/module/ModuleSpecProcessor.java
@@ -9,11 +9,13 @@ import java.io.IOException;
 import java.security.Permission;
 import java.security.Permissions;
 import java.util.ArrayList;
+import java.util.Collection;
 import java.util.Collections;
 import java.util.Enumeration;
 import java.util.HashSet;
 import java.util.List;
 import java.util.PropertyPermission;
+import java.util.Set;
 import java.util.function.Consumer;
 
 import org.jboss.as.server.deployment.Attachments;
@@ -146,9 +148,9 @@ public class ModuleSpecProcessor implements DeploymentUnitProcessor {
      * @param module              The additional module
      */
     private void addAllDependenciesAndPermissions(final ModuleSpecification moduleSpecification, final AdditionalModuleSpecification module) {
-        module.addSystemDependencies(moduleSpecification.getSystemDependencies());
-        module.addLocalDependencies(moduleSpecification.getLocalDependencies());
-        for(ModuleDependency dep : moduleSpecification.getUserDependencies()) {
+        module.addSystemDependencies(moduleSpecification.getSystemDependenciesSet());
+        module.addLocalDependencies(moduleSpecification.getLocalDependenciesSet());
+        for(ModuleDependency dep : moduleSpecification.getUserDependenciesSet()) {
             if(!dep.getIdentifier().equals(module.getModuleIdentifier())) {
                 module.addUserDependency(dep);
             }
@@ -202,9 +204,9 @@ public class ModuleSpecProcessor implements DeploymentUnitProcessor {
         for (final DependencySpec dep : moduleSpecification.getModuleSystemDependencies()) {
             specBuilder.addDependency(dep);
         }
-        final List<ModuleDependency> dependencies = moduleSpecification.getSystemDependencies();
-        final List<ModuleDependency> localDependencies = moduleSpecification.getLocalDependencies();
-        final List<ModuleDependency> userDependencies = moduleSpecification.getUserDependencies();
+        final Set<ModuleDependency> dependencies = moduleSpecification.getSystemDependenciesSet();
+        final Set<ModuleDependency> localDependencies = moduleSpecification.getLocalDependenciesSet();
+        final Set<ModuleDependency> userDependencies = moduleSpecification.getUserDependenciesSet();
 
         final List<PermissionFactory> permFactories = moduleSpecification.getPermissionFactories();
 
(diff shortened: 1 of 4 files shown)


server/src/main/java/org/jboss/as/server/deployment/module/ModuleSpecification.java
```diff
@@ -130,9 +130,11 @@ public class ModuleSpecification extends SimpleAttachable {
...
 
     public void addSystemDependency(final ModuleDependency dependency) {
-        if (!exclusions.contains(dependency.getIdentifier()) && !systemDependenciesSet.contains(dependency.getIdentifier())) {
-            this.systemDependencies.add(dependency);
-            this.systemDependenciesSet.add(dependency.getIdentifier());
+        if (!exclusions.contains(dependency.getIdentifier())) {
+            if (!systemDependenciesSet.contains(dependency.getIdentifier())) {
+                this.systemDependencies.add(dependency);
+                this.systemDependenciesSet.add(dependency.getIdentifier());
+            }
```
yersan: just as a side note, I was expecting a different change here for this commit what WFCORE-6607 does is nothing regarding code change so not sure why we wanted it tracked in a different Jira. The Jira title suggests a change but there is no change at all, so, it could be quite confusing. I'm not asking for a change, just adding my observation.

server/src/main/java/org/jboss/as/server/deployment/module/ModuleSpecification.java
```diff
@@ -32,15 +36,17 @@ public class ModuleSpecification extends SimpleAttachable {
...
      * System dependencies are dependencies that are added automatically by the container.
      */
-    private final List<ModuleDependency> systemDependencies = new ArrayList<>();
+    private final Set<ModuleDependency> systemDependenciesSet = new LinkedHashSet<>();
 
-    /** Unique names of system dependencies, used to ensure only one system ModuleDependency per name is added. */
-    private final Set<ModuleIdentifier> systemDependenciesSet = new HashSet<>();
+    /**
+     * List view of {@link #systemDependenciesSet}.
+     */
```
yersan, line 43: Optional: Add a `@Deprecated(forRemoval = true)` too? I understand we will have a follow-up to fully clean up the dual view of the dependencies kept in the Set and List. The same applies to `userDependencies`.

yersan (approved):
Added minor optional comments that, in my opinion, should not prevent merging it. This change has been confirmed that fix WFCORE-6606.

I also wondering 
1. Does it deserve a test case using EARs (it can be added later)?
2. There is no follow up for the deprecation, I guess, we can get it done for the next WildFly release, but not sure what is our cadence regarding deprecations and removals.

scottmarlow (approved):

server/src/main/java/org/jboss/as/server/deployment/module/ModuleSpecification.java
bstansberry: Done

bstansberry:
@yersan and @scottmarlow 

You've approved this but I just updated it a fair bit. The changes don't affect how the existing WF code interact with this though, so the good TCK results won't be affected.

1) I added @Deprecated annotations to a couple fields per Yeray's input.
2) I cleaned up some other javadoc and IDE warning stuff.
3) I added a test case.
4) Most substantially I changed the new 'remove' method I'd added, from:

```
removeUserDependency(ModuleDependency)
```
to 

```
removeUserDependencies(Predicate<ModuleDependency>)
```

That's a much better API to replace the intended use of getMutableUserDependencies(), which was to allow ApplicationClientDependencyProcessor to remove some dependencies that match its rules. Now it can pass in the rules and let ModuleSpecification deal with the internal state.

This last change shouldn't affect any existing  behavior, as nothing calls this remove method yet (other than the new test).

scottmarlow:
> @yersan and @scottmarlow
> 
> You've approved this but I just updated it a fair bit. The changes don't affect how the existing WF code interact with this though, so the good TCK results won't be affected.
> 
>     1. I added @deprecated annotations to a couple fields per Yeray's input.
> 
>     2. I cleaned up some other javadoc and IDE warning stuff.
> 
>     3. I added a test case.
> 
>     4. Most substantially I changed the new 'remove' method I'd added, from:
> 
> 
> ```
> removeUserDependency(ModuleDependency)
> ```
> 
> to
> 
> ```
> removeUserDependencies(Predicate<ModuleDependency>)
> ```
> 
> That's a much better API to replace the intended use of getMutableUserDependencies(), which was to allow ApplicationClientDependencyProcessor to remove some dependencies that match its rules. Now it can pass in the rules and let ModuleSpecification deal with the internal state.
> 
> This last change shouldn't affect any existing behavior, as nothing calls this remove method yet (other than the new test).

I really like this`Predicate<ModuleDependency>` change.  I was thinking about the `contract` issue and this is a nice solution.

server/src/main/java/org/jboss/as/server/deployment/module/ModuleSpecProcessor.java
```diff
@@ -146,9 +148,9 @@ private void deployModuleSpec(final DeploymentPhaseContext phaseContext) throws
      * @param module              The additional module
      */
     private void addAllDependenciesAndPermissions(final ModuleSpecification moduleSpecification, final AdditionalModuleSpecification module) {
-        module.addSystemDependencies(moduleSpecification.getSystemDependencies());
-        module.addLocalDependencies(moduleSpecification.getLocalDependencies());
-        for(ModuleDependency dep : moduleSpecification.getUserDependencies()) {
+        module.addSystemDependencies(moduleSpecification.getSystemDependenciesSet());
```
scottmarlow, line 151: I didn't mention this before but I really like the reduction in memory use by using Set with these changes.

yersan (approved):

yersan:
Thanks @bstansberry and @scottmarlow !

PR #5771, opened 2023-11-17
[WFCORE-6617] Clean out ts.elytron profile cruft

soul2zimate:
Issue: https://issues.redhat.com/browse/WFCORE-6617 [WFCORE-6617] Clean out ts.elytron profile cruft


yersan (approved):
Thanks @soul2zimate I've found minor cruft here: https://github.com/wildfly/wildfly-core/blob/035d41260346840f0073a90875997f2b8f16df17/testsuite/standalone/src/test/java/org/jboss/as/test/integration/management/cli/ReloadSASLFactoryTestCase.java#L95

Ideally, we could remove it with this PR, but we can also append an additional commit after merging it, if that's not done before I will remove it before emerging it.

We do not need to wait for CI after such comment removal.

yersan:
Thanks @soul2zimate !

PR #5776, opened 2023-11-21
[WFCORE-6624] Add Locale parameter while calling toUpperCase and toLowerCase methods

soul2zimate:
Issue: https://issues.redhat.com/browse/WFCORE-6624

 Add Locale parameter while calling `toUpperCase` and `toLowerCase` methods

This is similar to https://github.com/wildfly/wildfly/pull/17422


host-controller/src/main/java/org/jboss/as/host/controller/HostControllerEnvironment.java
```diff
@@ -307,7 +308,7 @@ public HostControllerEnvironment(Map<String, String> hostSystemProperties, boole
                 // Give up
                 qualifiedHostName = "unknown-host.unknown-domain";
             } else {
-                qualifiedHostName = qualifiedHostName.trim().toLowerCase();
+                qualifiedHostName = qualifiedHostName.trim().toLowerCase(Locale.ENGLISH);
```
bstansberry, line 311: @soul2zimate this is user input so I think the system locale is appropriate here.

soul2zimate: Thank you for the review, I updated it to `Locale.getDefault()`

server/src/main/java/org/jboss/as/server/ServerEnvironment.java
```diff
@@ -757,7 +758,7 @@ private void configureQualifiedHostName(String qualifiedHostName, String provide
                 // Give up
                 qualifiedHostName = "unknown-host.unknown-domain";
             } else {
-                qualifiedHostName = qualifiedHostName.trim().toLowerCase();
+                qualifiedHostName = qualifiedHostName.trim().toLowerCase(Locale.ENGLISH);
```
yersan, line 761: This should be removed or, alternatively, use Locale.getDefault(). It is the same case as the Host Controller variant.

soul2zimate: 👍 Fixed.

yersan (approved):

PR #5784, opened 2023-11-30
[WFCORE-6635] Add a base JBossModulesCommandBuilder for launching oth…

jamezp:
…er types like the appclient or other JBoss Module containers in a new process.

https://issues.redhat.com/browse/WFCORE-6635

### DIFF
--- a/launcher/src/main/java/org/wildfly/core/launcher/AbstractCommandBuilder.java
+++ b/launcher/src/main/java/org/wildfly/core/launcher/AbstractCommandBuilder.java
@@ -8,194 +8,63 @@ package org.wildfly.core.launcher;
 import java.nio.file.Files;
 import java.nio.file.Path;
 import java.nio.file.Paths;
-import java.util.ArrayList;
-import java.util.Collection;
-import java.util.Collections;
 import java.util.List;
 
-import org.wildfly.core.launcher.Arguments.Argument;
 import org.wildfly.core.launcher.logger.LauncherMessages;
 
 /**
  * @author <a href="mailto:jperkins@redhat.com">James R. Perkins</a>
  */
-@SuppressWarnings("unused")
-abstract class AbstractCommandBuilder<T extends AbstractCommandBuilder<T>> implements CommandBuilder {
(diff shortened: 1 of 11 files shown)


launcher/src/main/java/org/wildfly/core/launcher/AbstractCommandBuilder.java
```diff
@@ -8,196 +8,65 @@
...
-     * @param moduleDirs the collection of module directories to use
-     *
-     * @return the builder
-     *
-     * @throws java.lang.IllegalArgumentException if any of the module paths are invalid or {@code null}
-     */
+    @Override
     public T setModuleDirs(final Iterable<String> moduleDirs) {
-        environment.setModuleDirs(moduleDirs);
+        super.addModuleDirs(moduleDirs);
```
yersan, line 60: This should be `setModuleDirs()` instead of the add variant, otherwise the module dirs will be duplicated, and that's the root cause of the ModuleLoadingManagementTestCase failed tests

launcher/src/main/java/org/wildfly/core/launcher/AbstractCommandBuilder.java
```diff
@@ -8,196 +8,65 @@
...
-     * @param moduleDirs the array of module directories to use
-     *
-     * @return the builder
-     *
-     * @throws java.lang.IllegalArgumentException if any of the module paths are invalid or {@code null}
-     */
+    @Override
     public T setModuleDirs(final String... moduleDirs) {
-        environment.setModuleDirs(moduleDirs);
+        super.addModuleDirs(moduleDirs);
```
yersan, line 66: This should be `setModuleDirs()` instead

yersan (changes requested):

launcher/src/main/java/org/wildfly/core/launcher/AbstractCommandBuilder.java
jamezp: Nice catch, fixed.

yersan (approved):

yersan:
Thanks @jamezp

PR #5789, opened 2023-12-04
[WFCORE-6579] Allow process controller log file capture initial manag…

yersan:
…ed process Standard Out and Err

Jira issue: https://issues.redhat.com/browse/WFCORE-6579

### DIFF
--- a/host-controller/src/main/java/org/jboss/as/host/controller/Main.java
+++ b/host-controller/src/main/java/org/jboss/as/host/controller/Main.java
@@ -89,6 +89,16 @@ public final class Main {
         } catch (Throwable ignored) {
         }
 
+        // This message is not used to display any information on the stdout and stderr of this process, it signals the
+        // Process Controller to switch the stdout and stderr of this managed process from the process controller log to
+        // the process controller stdout and stderr.
+        // Once the StdioContext gets installed, both the stdout and stderr of this managed process will be captured,
+        // aggregated and handled by this process logging framework.
+        STDOUT.println(ProcessController.STDIO_ABOUT_TO_INSTALL_MSG);
+        STDOUT.flush();
+        STDERR.println(ProcessController.STDIO_ABOUT_TO_INSTALL_MSG);
+        STDERR.flush();
+
         // Install JBoss Stdio to avoid any nasty crosstalk.
         StdioContext.install();
         final StdioContext context = StdioContext.create(
--- a/process-controller/src/main/java/org/jboss/as/process/CommandLineConstants.java
+++ b/process-controller/src/main/java/org/jboss/as/process/CommandLineConstants.java
@@ -121,6 +121,9 @@ public class CommandLineConstants {
     public static final String PREFER_IPV4_STACK = "java.net.preferIPv4Stack";
     public static final String PREFER_IPV6_ADDRESSES = "java.net.preferIPv6Addresses";
 
+    public static final String MANAGED_PROCESS_SYSTEM_ERROR_TO_LOG = "jboss.domain.managed-process.system.stderr.to.process-controller.log";
+    public static final String MANAGED_PROCESS_SYSTEM_OUT_TO_LOG = "jboss.domain.managed-process.system.stdout.to.process-controller.log";
+
     /** Additional class path items, used only by app client*/
     public static final String APPCLIENT_CONFIG = "--appclient-config";
     public static final String SHORT_HOST = "-H";
--- a/process-controller/src/main/java/org/jboss/as/process/ManagedProcess.java
+++ b/process-controller/src/main/java/org/jboss/as/process/ManagedProcess.java
@@ -6,6 +6,8 @@
 package org.jboss.as.process;
 
 import static java.lang.Thread.holdsLock;
+import static org.jboss.as.process.CommandLineConstants.MANAGED_PROCESS_SYSTEM_ERROR_TO_LOG;
+import static org.jboss.as.process.CommandLineConstants.MANAGED_PROCESS_SYSTEM_OUT_TO_LOG;
 import static org.jboss.as.process.protocol.StreamUtils.copyStream;
 import static org.jboss.as.process.protocol.StreamUtils.safeClose;
 import static org.jboss.as.process.protocol.StreamUtils.writeBoolean;
@@ -27,11 +29,13 @@ import java.util.ArrayList;
 import java.util.List;
 import java.util.Map;
 import java.util.concurrent.atomic.AtomicInteger;
+import java.util.function.Consumer;
 
 import org.jboss.as.process.logging.ProcessLogger;
 import org.jboss.as.process.stdin.Base64OutputStream;
 import org.jboss.logging.Logger;
 import org.wildfly.common.Assert;
+import org.wildfly.security.manager.WildFlySecurityManager;
 
 /**
  * A managed process.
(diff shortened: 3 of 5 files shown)


yersan:
CC @bstansberry / @jamezp I ended up with the above, let me know your thoughts, thanks!

server/src/main/java/org/jboss/as/server/DomainServerMain.java
```diff
@@ -81,6 +81,15 @@ public static void main(String[] args) {
         } catch (Throwable ignored) {
         }
 
+        // This message is not used to display any information on the standard output, it signals the Process Controller
+        // to switch the standard output of this managed process from the process controller log to the process controller
+        // standard output.
+        // Once the StdioContext gets installed, both the stout and stderr of this managed process will be captured,
+        // aggregated and handled by this process logging framework.
+        // Signalling the Process Controller to switch also the System.err is optional since after installing StdioContext,
+        // there won't be more logs sent to the standard Err.
```
yersan, line 90: thinking a bit more, although optional, the aggregation of the StdError and Stout and where they are redirected will be handled by the `logging.configuration` file, so assuming that both will be redirected to the Stout is weak, since I guess that can be changed via the logging configuration file and instead of using `handler.CONSOLE.target=SYSTEM_OUT` use `handler.CONSOLE.target=SYSTEM_ERR`, so probably this should be changed.

process-controller/src/main/java/org/jboss/as/process/ManagedProcess.java
```diff
@@ -107,7 +125,9 @@ enum State {
         this.pcAuthKey = pcAuthKey;
         isPrivileged = privileged;
         respawnPolicy = respawn ? RespawnPolicy.RESPAWN : RespawnPolicy.NONE;
-        log = Logger.getMessageLogger(ProcessLogger.class, "org.jboss.as.process." + processName + ".status");
+        logStatus = Logger.getMessageLogger(ProcessLogger.class, "org.jboss.as.process." + processName + ".status");
+        logSystemErr = Logger.getMessageLogger(ProcessLogger.class, "org.jboss.as.process." + processName + ".system.err");
+        logSystemOut = Logger.getMessageLogger(ProcessLogger.class, "org.jboss.as.process." + processName + ".system.out");
```
jamezp, line 130: Not critical, should we use `stdout` and `stderr` instead? I don't have a strong opinion, I'm just curious :)

We also might want to consider the order to have the process name last to make filtering easier. For example you could just skip `stdout` with `logger.org.jboss.as.process.system.out.level=WARN`

yersan: > Not critical, should we use stdout and stderr instead?

I am fine with both. `xxx.system.out` and `xxx.system.err` are very intuitive to me due to the similitude with the Java methods, if `xxx.system.stdout` and `xxx.system.stderr` are more related to logging world, I am fine with them too. I do not have any preference.
 

> We also might want to consider the order to have the process name last to make filtering easier. For example you could just skip stdout with logger.org.jboss.as.process.system.out.level=WARN

Thanks, that's a smarter choice. I agree with you.

We already have the `.status` category defined as `"org.jboss.as.process." + processName + ".status"`. Should we change it too? 

That would make all the related categories consistent, although would affect existing configurations. However, since are logging categories, any existing configuration can be changed to adapt to the new version and that would not affect server behaviour, so I am in favor of adapting those three.

server/src/main/java/org/jboss/as/server/DomainServerMain.java
yersan: Done

host-controller/src/main/java/org/jboss/as/host/controller/Main.java
```diff
@@ -88,6 +88,14 @@ public static void main(String[] args) throws IOException {
         } catch (Throwable ignored) {
         }
 
+        // This message is not used to display any information on the standard output, it signals the Process Controller
+        // to switch the standard output of this managed process from the process controller log to the process controller
+        // standard output.
+        // Once the StdioContext gets installed, both the stout and stderr of this managed process will be captured,
```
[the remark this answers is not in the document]
yersan: Done

process-controller/src/main/java/org/jboss/as/process/ManagedProcess.java
yersan: @jamezp 

I've changed the categories to `xxx.system.stdout` and `xxx.system.stderr`.

However, I left the `"org.jboss.as.process." + processName + ".system.sterr"` and `"org.jboss.as.process." + processName + ".system.stout"` as proposed originally since we already have `"org.jboss.as.process." + processName + ".status"` and users could be familiar with that category. 

I prefer being conservative here since this is an enhancement and I don't want to introduce a change on the `.status` category or have categories working differently than the status one. I think the usage of them is unlikely, but if they are needed I think is preferable a consistent with what we previously have.

process-controller/src/main/java/org/jboss/as/process/ManagedProcess.java
```diff
@@ -107,7 +125,9 @@ enum State {
         this.pcAuthKey = pcAuthKey;
         isPrivileged = privileged;
         respawnPolicy = respawn ? RespawnPolicy.RESPAWN : RespawnPolicy.NONE;
-        log = Logger.getMessageLogger(ProcessLogger.class, "org.jboss.as.process." + processName + ".status");
+        logStatus = Logger.getMessageLogger(ProcessLogger.class, "org.jboss.as.process." + processName + ".status");
+        logSystemErr = Logger.getMessageLogger(ProcessLogger.class, "org.jboss.as.process." + processName + ".system.sterr");
```
bstansberry, line 129: s/sterr/stderr

process-controller/src/main/java/org/jboss/as/process/ManagedProcess.java
```diff
@@ -107,7 +125,9 @@ enum State {
...
+        logSystemOut = Logger.getMessageLogger(ProcessLogger.class, "org.jboss.as.process." + processName + ".system.stout");
```
bstansberry, line 130: s/stout/stdout

bstansberry: +1 to 

"org.jboss.as.process." + processName + "<suffix>"

We don't want to mess with the existing 'status' and these should be consistent.

bstansberry (changes requested):
I'll happily defer to you and @jamezp re the 'system.stdxxx' thing, but my input is just 'stdxxx' is best. Admins consume logs, and they may not even know about the Java System class. OTOH I expect a Java developer interested in domain logging would understand what stdout and stderr mean.

process-controller/src/main/java/org/jboss/as/process/CommandLineConstants.java
```diff
@@ -121,6 +121,9 @@ public class CommandLineConstants {
     public static final String PREFER_IPV4_STACK = "java.net.preferIPv4Stack";
     public static final String PREFER_IPV6_ADDRESSES = "java.net.preferIPv6Addresses";
 
+    public static final String MANAGED_PROCESS_SYSTEM_ERROR_TO_LOG = "jboss.domain.managed-process.system.sterr.to.process-controller.log";
```
jamezp, line 124: sterr should be stderr

process-controller/src/main/java/org/jboss/as/process/CommandLineConstants.java
```diff
@@ -121,6 +121,9 @@ public class CommandLineConstants {
...
+    public static final String MANAGED_PROCESS_SYSTEM_OUT_TO_LOG = "jboss.domain.managed-process.system.stout.to.process-controller.log";
```
jamezp, line 125: stout should be stdout

process-controller/src/main/java/org/jboss/as/process/ManagedProcess.java
yersan: Done, thanks!

process-controller/src/main/java/org/jboss/as/process/CommandLineConstants.java
yersan: Done, thanks!

process-controller/src/main/java/org/jboss/as/process/CommandLineConstants.java
yersan: Done, thanks!

jamezp (approved):

yersan:
Thanks @jamezp and @bstansberry !

PR #5793, opened 2023-12-11
Minor fix next release version

yersan:


yersan:
There is no code change, so we can safely merge it without using a Jira.

PR #5794, opened 2023-12-11
[DO NOT MERGE] ** [WFCORE-6446] Testing if the syslog address was set.

jamezp:
This is a test to see if we can get more information from a test failure.


jamezp:
I'm going to close this. It's not seeming to fail.

PR #5798, opened 2023-12-14
[WFCORE-6643] Upgrade JGit from 6.7.0.202309050840-r to 6.8.0.2023112…

yersan:
…91450-r

Jira issue: https://issues.redhat.com/browse/WFCORE-6643


ehsavoie (approved):
We don't have to upgrade the dependencies as we don't use org.eclipse.jgit:org.eclipse.jgit.archive or jgit cli which requires commons-compress and we already have the other ones up-to-date

yersan:
Thanks @ehsavoie

PR #5805, opened 2023-12-19
[WFCORE-6648]: Mark "read-config-as-xml" as deprecated on Community stream.

ehsavoie:
* Making te operation deprecated if run on 'community' stability.

Issue: https://issues.redhat.com/browse/WFCORE-6648

### DIFF
--- a/controller/src/main/java/org/jboss/as/controller/operations/common/XmlMarshallingHandler.java
+++ b/controller/src/main/java/org/jboss/as/controller/operations/common/XmlMarshallingHandler.java
@@ -8,7 +8,6 @@ import java.io.ByteArrayOutputStream;
 import java.nio.charset.StandardCharsets;
 
 import org.jboss.as.controller.OperationContext;
-import org.jboss.as.controller.SimpleOperationDefinition;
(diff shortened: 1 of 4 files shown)


host-controller/src/main/java/org/jboss/as/domain/controller/transformers/KernelAPIVersion.java
```diff
@@ -75,6 +75,8 @@ public enum KernelAPIVersion {
     VERSION_22_0(22, 0, 0),
     // WildFly 30.0.0
     VERSION_23_0(23, 0, 0),
+    // WildFly 31.0.0
+    VERSION_24_0(24, 0, 0),
```
yersan, line 79: Do not specify the version here, right now 24 is in CURRENT. You can use `ModelVersion.create(24, 0, 0 )`  or even `ModelVersion.create(24)` without sticking it to the KernelAPIVersion.java

yersan (changes requested):

yersan (approved):

yersan:
Thanks @ehsavoie

PR #5810, opened 2023-12-21
[WFCORE-6649][WFCORE-6653] Deprecate --no-resol-local-cache and replace it with --…

yersan:
…use-default-local-cache. Fix minor help installer descriptions

Jira issue: https://issues.redhat.com/browse/WFCORE-6649
Jira issue: https://issues.redhat.com/browse/WFCORE-6653

--no-resol-local-cache gets deprecated and changes its default value to true.

### DIFF
--- a/installation-manager/src/main/java/org/wildfly/core/instmgr/cli/UpdateCommand.java
+++ b/installation-manager/src/main/java/org/wildfly/core/instmgr/cli/UpdateCommand.java
@@ -20,6 +20,7 @@ import org.jboss.as.cli.CommandContext;
 import org.jboss.as.cli.Util;
 import org.jboss.as.cli.impl.aesh.cmd.HeadersCompleter;
 import org.jboss.as.cli.impl.aesh.cmd.HeadersConverter;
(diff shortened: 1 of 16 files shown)


installation-manager/src/main/java/org/wildfly/core/instmgr/cli/UpdateCommand.java
```diff
@@ -38,8 +39,10 @@ public class UpdateCommand extends AbstractInstMgrCommand {
     private List<String> repositories;
     @Option(name = "local-cache")
     private File localCache;
-    @Option(name = "no-resolve-local-cache", hasValue = false)
+    @Option(name = NO_RESOLVE_LOCAL_CACHE_OPTION, hasValue = false, activator = AbstractInstMgrCommand.NoResolveLocalCacheActivator.class, defaultValue = "true")
```
jfdenise, line 42: A minor comment, this option ha no value, setting the default value to true is not needed.

jfdenise (approved):
I made a comment but this shouldn't block the merge.

darranl:
Thank you for the review @jfdenise

PR #5818, opened 2024-01-02
WFCORE-6347 Add ServiceDescriptor variants of OperationContext.hasOptionalCapability(...) and CapabilityServiceSupport.hasCapability(...)

pferraro:
… for convenience.
https://issues.redhat.com/browse/WFCORE-6347

### DIFF
--- a/controller/src/main/java/org/jboss/as/controller/OperationContext.java
+++ b/controller/src/main/java/org/jboss/as/controller/OperationContext.java
@@ -712,6 +712,7 @@ public interface OperationContext extends ExpressionResolver, CapabilityServiceD
      *            manager exists and its {@link SecurityManager#checkPermission checkPermission} method doesn't allow
      *            access to the relevant system property or environment variable
      */
+    @Override
     ModelNode resolveExpressions(ModelNode node) throws OperationFailedException;
 
     /**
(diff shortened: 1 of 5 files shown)


pferraro:
@bstansberry The full integration test failures are due to a case in wildfly full where the caller invokes OperationContext.hasOptionalCapability(...) with a null "dependent" parameter resulting in method signature ambiguity.  The javadoc indicates that this is forbidden, but git blame indicates that this there was an intentional effort to temporarily allow a null dependent via https://issues.redhat.com/browse/WFCORE-900.
Can you can elaborate on the current state of this?  Why would a operation handler need to call this method with a null "dependent" parameter rather than use  OperationContext.getCapabilityServiceSupport().hasCapability(String)?

pferraro:
@bstansberry Wondering if you had a moment to look at this.

bstansberry:
> @bstansberry The full integration test failures are due to a number of instances in wildfly full where the caller invokes OperationContext.hasOptionalCapability(...) with a null "dependent" parameter resulting in method signature ambiguity. The javadoc indicates that this is forbidden, but git blame indicates that this there was an intentional effort to temporarily allow a null dependent via https://issues.redhat.com/browse/WFCORE-900. Can you can elaborate on the current state of this? Why would a operation handler need to call this method with a null "dependent" parameter rather than use OperationContext.getCapabilityServiceSupport().hasCapability(String)?

I don't see any need to continue to be permissive. In the WFCORE-900 description I was pretty clear that the contract (the javadoc) doesn't allow null, so that can only be a temporary thing. That was in Aug 2015 when capabilities and requirements were really new and the bulk of the server didn't use them. So we needed flexibility to let us iterate toward doing it right. 8+ years later any code that isn't doing it right needs to be fixed and the permissiveness can go away.

pferraro:
@bstansberry FYI - the downstream conflicts have been resolved and this is ready for review.

yersan:
Thanks @pferraro

PR #5821, opened 2024-01-05
[WFCORE-4917]: Experimental - Provide a cool boot banner.

ehsavoie:
* Add support for a cool ASCII art banner via a banner.txt in the $JBOSS_HOME/bin directory.

Jira: https://issues.redhat.com/browse/WFCORE-4917
      https://issues.redhat.com/browse/WFLY-19116

Required for https://github.com/wildfly/wildfly/pull/17703


core-feature-pack/galleon-feature-pack/src/main/resources/packages/product.conf/content/modules/system/layers/base/org/jboss/as/product/wildfly-core/module.xml
```diff
@@ -4,8 +4,7 @@
   ~ Copyright The WildFly Authors
   ~ SPDX-License-Identifier: Apache-2.0
   -->
-
-<module xmlns="urn:jboss:module:1.9" name="org.jboss.as.product" slot="wildfly-core">
+<module xmlns="urn:jboss:module:1.5" name="org.jboss.as.product" slot="wildfly-core">
```
yersan, line 7: Downgrading the JBoss Modules module version is unexpected here. We should keep 1.9

version/src/main/java/org/jboss/as/version/ProductConfig.java
```diff
@@ -81,10 +84,18 @@ private ProductConfig(ModuleLoader loader, ProductConfProps productConfProps, Ma
...
+                        projectBanner = System.lineSeparator() + Files.readString(bannerPath);
+                    } catch (Exception e) {
+                        // Don't care
+                    }
+                }
             }
-
             setSystemProperties(productConfProps.miscProperties, providedProperties);
         } catch (Exception e) {
+            e.printStackTrace();
```
yersan, line 98: If we want the trace, I would say we have to use a logger. Printing the trace on the standard error doesn't look correct to me.

version/src/main/java/org/jboss/as/version/ProductConfig.java
```diff
@@ -167,6 +187,14 @@ public Set<Stability> getStabilitySet() {
         return this.stabilities;
     }
 
+    /**
+     * The product ASCII banner defined in MANIFEST.MF using \n as line ending.
+     * @return the ASCII banner.
+     */
```
yersan, line 193: This is not accurate since the banner is being retrieved from a file and not from the MANIFEST.MF in this class

yersan (changes requested):

ehsavoie:
@yersan all should be fixed :)

version/src/main/java/org/jboss/as/version/ProductConfig.java
```diff
@@ -34,21 +35,23 @@ public class ProductConfig implements Serializable {
...
 
     public static ProductConfig fromKnownSlot(String slot, ModuleLoader loader, Map<?, ?> providedProperties) {
-        return new ProductConfig(loader, new ProductConfProps(slot), providedProperties);
+        return new ProductConfig(loader, null, new ProductConfProps(slot), providedProperties);
     }
-    private ProductConfig(ModuleLoader loader, ProductConfProps productConfProps, Map<?, ?> providedProperties) {
+    private ProductConfig(ModuleLoader loader, String home, ProductConfProps productConfProps, Map<?, ?> providedProperties) {
         String productName = null;
         String projectName = null;
+        String projectBanner = "";
```
yersan, line 54: This is now no longer used.

core-feature-pack/galleon-common/src/main/resources/configs/standalone/model.xml
```diff
@@ -15,6 +15,7 @@
         <prop name="--internal-remove-config" value=""/>
     </props>
     <packages>
+        <package name="banner" optional="true"/>
```
yersan, line 18: My initial concern here was more about how this will affect WildFly full. If this package were enabled in full, we would get the WildFly Core banner in WildFly full. I don't know what are the plans for a banner in WildFly full, if there is one, I guess it will overwrite the banner.txt with its banner and it will use the same package.

yersan (commented):
Changes look good to me, just a minor variable removal nit. However, they would need also a consensus about how WildFly full is going to handle a banner. I am assuming full is going to overwrite both `banner.txt` file and the galleon package name

@bstansberry / @pferraro / @ehsavoie could you add your feedback here about how we plan to handle it in WildFly full? thanks

core-feature-pack/galleon-common/src/main/resources/configs/standalone/model.xml
ehsavoie: We can use a WildFly banner instead

ehsavoie: Yes the idea is to provide the same package with a WildFly Full banner

ehsavoie: @bstansberry Do yo agree with that ?

ehsavoie: https://github.com/wildfly/wildfly/pull/17703

ehsavoie: https://issues.redhat.com/browse/WFLY-19116

version/src/main/java/org/jboss/as/version/ProductConfig.java
```diff
@@ -97,6 +100,19 @@ private ProductConfig(ModuleLoader loader, ProductConfProps productConfProps, Ma
         this.stabilities = EnumSet.range(maxStability, minStability);
     }
 
+    public static void getBannerFile(ProductConfig config, String home) {
+        if (home != null) {
+            Path bannerPath = Paths.get(home, "bin", "banner.txt");
```
bstansberry, line 105: IMO this should be read as a classloader resource from the org.jboss.as.version module. I don't want to start putting a file in bin in every server we provision that's only meant for an experimental feature[1]. Having it in bin also makes the file a kind of API. That might be intentional, but the WFCORE-4917 description doesn't say anything about that being the intent.

[1] WFLY-19021 will provide a hook that can be used to disable provisioning it, but users would need to go out of their way to be aware of and take advantage of that.

bstansberry (changes requested):

version/src/main/java/org/jboss/as/version/ProductConfig.java
ehsavoie: Fixed, updating the Jira accordingly

bstansberry (approved):

bstansberry:
@yersan the handling in full WF is fine; I've approved https://github.com/wildfly/wildfly/pull/17703

yersan (approved):

yersan:
Thanks @ehsavoie / @bstansberry

PR #5824, opened 2024-01-07
[WFCORE-6658] Convert ControlledProcessStateService to the current MS…

bstansberry:
…C API

https://issues.redhat.com/browse/WFCORE-6658

### DIFF
--- a/bootable-jar/runtime/src/main/java/org/wildfly/core/jar/runtime/Server.java
+++ b/bootable-jar/runtime/src/main/java/org/wildfly/core/jar/runtime/Server.java
@@ -17,6 +17,9 @@ import java.util.concurrent.ExecutorService;
 import java.util.concurrent.Executors;
 import java.util.concurrent.Future;
 import java.util.concurrent.TimeUnit;
+import java.util.concurrent.atomic.AtomicReference;
+import java.util.function.Supplier;
+
 import org.jboss.as.controller.ControlledProcessState;
 import org.jboss.as.controller.ControlledProcessStateService;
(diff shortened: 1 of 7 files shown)


bstansberry:
@ropalka FYI.

bootable-jar/runtime/src/main/java/org/wildfly/core/jar/runtime/Server.java
```diff
@@ -221,6 +230,7 @@ private synchronized void establishModelControllerClient(ControlledProcessState.
         if (state != ControlledProcessState.State.STOPPING && state != ControlledProcessState.State.STOPPED && serviceContainer != null) {
             ModelControllerClientFactory clientFactory;
             try {
+                // TODO replace this in start() with the ServiceActivator approach we used to capture the ProcessStateNotifier
```
yersan, line 233: I've just added WFCORE-6662 as a follow up to cover this.

yersan (approved):
LGTM, @ropalka could you also take a quick look?

My very minor detail is that now the new SimpleService that provides the ProcessStateNotifier is not invoking `consumer.accept(null)` on the stop method as we did in ControlledProcessStateService:61, I don't think it could be an issue, but I am not certainly sure if that could cause any unexpected side effect. It is done when the service is being stopped, so it looks like safe.

yersan:
@ropalka FYI in case you have missed it.

ropalka (approved):
LGTM

yersan:
Thanks @bstansberry and @ropalka

PR #5827, opened 2024-01-10
[WFCORE-6661] More accurate, resettable start time calculation

bstansberry:
https://issues.redhat.com/browse/WFCORE-6661

This encapsulates the existing and possible future (WFCORE-6657) behavior around calculating and resetting 'start time' in a class, instead of passing around primitive 'startTime' longs and doing math. It uses RuntimeMXBean for the original 'start time' of a non-embedded process, instead of using Module.getStartTime(), which is a time when JBoss Modules started, not when the JVM did.

### DIFF
--- a/host-controller/src/main/java/org/jboss/as/host/controller/HostControllerService.java
+++ b/host-controller/src/main/java/org/jboss/as/host/controller/HostControllerService.java
@@ -36,6 +36,7 @@ import org.jboss.as.host.controller.logging.HostControllerLogger;
 import org.jboss.as.remoting.HttpListenerRegistryService;
 import org.jboss.as.remoting.management.ManagementRemotingServices;
 import org.jboss.as.server.BootstrapListener;
+import org.jboss.as.server.ElapsedTime;
 import org.jboss.as.server.FutureServiceContainer;
 import org.jboss.as.server.Services;
 import org.jboss.as.server.logging.ServerLogger;
@@ -78,8 +79,9 @@ public class HostControllerService implements Service<AsyncFuture<ServiceContain
     private final ControlledProcessState processState;
     private final String authCode;
     private final CapabilityRegistry capabilityRegistry;
+    private final ElapsedTime elapsedTime;
     private volatile FutureServiceContainer futureContainer;
-    private volatile long startTime;
+    private volatile boolean everStopped;
 
     public HostControllerService(final HostControllerEnvironment environment, final HostRunningModeControl runningModeControl,
                           final String authCode, final ControlledProcessState processState, FutureServiceContainer futureContainer) {
@@ -87,7 +89,7 @@ public class HostControllerService implements Service<AsyncFuture<ServiceContain
         this.runningModeControl = runningModeControl;
         this.authCode = authCode;
         this.processState = processState;
-        this.startTime = environment.getStartTime();
+        this.elapsedTime = environment.getElapsedTime();
         this.futureContainer = futureContainer;
         this.capabilityRegistry = new CapabilityRegistry(false);
     }
@@ -99,8 +101,10 @@ public class HostControllerService implements Service<AsyncFuture<ServiceContain
 
     @Override
     public void start(StartContext context) throws StartException {
-        //Moved to AbstractControllerService.start()
-        // processState.setStarting();
+
+        // If this is a reload, track start time independently of the overall process elapsed time
+        ElapsedTime startupTime = everStopped ? elapsedTime.checkpoint() : elapsedTime;
+
         final ProductConfig config = environment.getProductConfig();
         final String prettyVersion = config.getPrettyVersionString();
         final String banner = environment.getStability() == org.jboss.as.version.Stability.EXPERIMENTAL ? config.getBanner() : "";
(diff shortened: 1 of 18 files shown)


yersan:
Hello @bstansberry are you interested in getting this for WF23?

bstansberry:
@yersan Thanks for asking. No, this not urgent and shouldn't go in WF 31. It's a small foundation piece that's important for my prototyping at https://github.com/bstansberry/wildfly-core/commits/WildFly_CRaC, but that is just a research project. It's valid on it's own (it uses a more accurate JVM start time) so I went ahead and sent up the PR, but that benefit is quite minor and is not enough to justify doing it so late in the 31 work.

host-controller/src/main/java/org/jboss/as/host/controller/HostControllerService.java
```diff
@@ -148,14 +152,7 @@ public void start(StartContext context) throws StartException {
...
 
-        long startTime = this.startTime;
-        if (startTime == -1) {
-            startTime = System.currentTimeMillis();
-        } else {
-            this.startTime = -1;
-        }
-
-        final BootstrapListener bootstrapListener = new BootstrapListener(serviceContainer, startTime, serviceTarget, futureContainer, prettyVersion + " (Host Controller)", environment.getDomainTempDir());
+        final BootstrapListener bootstrapListener = new BootstrapListener(serviceContainer, elapsedTime, serviceTarget, futureContainer, prettyVersion + " (Host Controller)", environment.getDomainTempDir());
```
bstansberry, line 155: oops; this is wrong! It should be 'startupTime'.

server/src/main/java/org/jboss/as/server/ElapsedTime.java
```diff
@@ -0,0 +1,98 @@
...
+
+import java.lang.management.ManagementFactory;
+import java.lang.management.RuntimeMXBean;
+import java.util.Map;
+import java.util.WeakHashMap;
+
+/**
+ * Tracks elapsed time since either JVM start or a moment of initialization. The start point
+ * for the calculation can be 'reset', allowing the calculation provided by an instance to
+ * account for a conceptual 'restart'.
```
bstansberry, line 16: This needs updating as it describes an earlier form of behavior before I saw the need to switch to the 'checkpoint()' approach.

server/src/main/java/org/jboss/as/server/ElapsedTime.java
```diff
@@ -0,0 +1,98 @@
...
+    /**
+     * Creates a tracker that tracks elapsed time since whatever point the given {@code toCopy} tracker does.
+     * This is a cloning operation; the created tracker shares no state with the {@code toCopy} tracker.
+     *
+     * @param toCopy tracker to copy. Cannot be {@code null}
+     * @return the tracker. Will not return {@code null}.
+     */
+    public static ElapsedTime from(ElapsedTime toCopy) {
+        return new ElapsedTime(toCopy.startTime);
+    }
```
bstansberry, line 45: I believe I can remove this now as it was for the use case that 'checkpoint()' handles now.

server/src/main/java/org/jboss/as/server/ElapsedTime.java
```diff
@@ -0,0 +1,98 @@
...
+     * Any ElapsedTime objects that were returned from this object's {@link #checkpoint()} method will also be reset.
+     * Meant for cases where the 'origin' moment may have changed and this tracker should be updated accordingly --
+     * for example, in a 'restored' JVM that supports some form of checkpoint and restore behavior.
+     */
+    public synchronized void reset() {
+        startTime = null;
+        for (ElapsedTime checkpoint : checkpoints.keySet()) {
+            checkpoint.reset();
+        }
+    }
```
bstansberry, line 85: oops, forgot to remove this.

host-controller/src/main/java/org/jboss/as/host/controller/HostControllerService.java
```diff
@@ -99,8 +101,10 @@ public HostControllerService(final HostControllerEnvironment environment, final
 
     @Override
     public void start(StartContext context) throws StartException {
-        //Moved to AbstractControllerService.start()
-        // processState.setStarting();
+
+        // If this is a reload, track start time independently of the overall process elapsed time
+        ElapsedTime startupTime = everStopped ? elapsedTime.checkpoint() : elapsedTime;
```
yersan, line 106: It looks like `everStopped` should be flagged as true when HostControllerService is stopped, otherwise this ternary operator makes no sense.

yersan (changes requested):

bstansberry:
I think I'm going to close this soonish as it's fairly complex and the use case for it isn't immediate. Before I do I'll address Yeray's point above so that doesn't fall in the crack if this is ever reopened.

host-controller/src/main/java/org/jboss/as/host/controller/HostControllerService.java
bstansberry: Fixed

yersan:
Upss, I made a mistake by resolving the conflicts directly.

bstansberry:
@yersan I rebased.

yersan (approved):

yersan:
Thanks @bstansberry

PR #5829, opened 2024-01-11
WFCORE-6176 Generalize capability filter to cover more complex scenarios.

pferraro:
Expands context of capability filter in wildfly-subsystem to support more complex use cases, e.g. https://github.com/wildfly/wildfly/blob/30.0.1.Final/undertow/src/main/java/org/wildfly/extension/undertow/ServerAdd.java#L78

https://issues.redhat.com/browse/WFCORE-6176


bstansberry:
I put the Blocker label on this as otherwise this change would be a breaking change in WF 32. It's new API in 31 so it's not too late to change it.

bstansberry:
Thanks, @pferraro

PR #5831, opened 2024-01-15
[WFCORE-6664] Upgrade SLF4J from 2.0.9 to 2.0.11

yersan:
Jira issue: https://issues.redhat.com/browse/WFCORE-6664

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -232,7 +232,7 @@
         <version.org.mock-server.mockserver-netty>5.8.1</version.org.mock-server.mockserver-netty>
         <version.org.mockito>3.10.0</version.org.mockito>
         <version.org.projectodd.vdx>1.1.6</version.org.projectodd.vdx>
-        <version.org.slf4j>2.0.9</version.org.slf4j>
+        <version.org.slf4j>2.0.11</version.org.slf4j>
         <version.org.syslog4j>0.9.30</version.org.syslog4j>
         <version.org.wildfly.client.config>1.0.1.Final</version.org.wildfly.client.config>
         <version.org.wildfly.common>1.7.0.Final</version.org.wildfly.common>


jamezp (approved):
I think this is okay. There is one commit that is not backwards compatible, but it's on a utility that should not be used. See https://github.com/qos-ch/slf4j/commit/316b5d1727d647250ff791565650070094f5b85e and https://github.com/qos-ch/slf4j/issues/361

yersan:
@jamezp thanks for looking at this. You are talking about the `Util.report()` method removal, isn't it? 

The class itself states that [it is an internal utility class](https://github.com/qos-ch/slf4j/blob/316b5d1727d647250ff791565650070094f5b85e/slf4j-api/src/main/java/org/slf4j/helpers/Util.java#L28C2-L28C30), so I think we are safe. The `slf4j-api` module is distributed in the modules directory, so potentially there could be application using it, but again, it states that it is an internal utility class, so if someone is using it, its usage should be removed.

jamezp:
> @jamezp thanks for looking at this. You are talking about the `Util.report()` method removal, isn't it?
> 
> The class itself states that [it is an internal utility class](https://github.com/qos-ch/slf4j/blob/316b5d1727d647250ff791565650070094f5b85e/slf4j-api/src/main/java/org/slf4j/helpers/Util.java#L28C2-L28C30), so I think we are safe. The `slf4j-api` module is distributed in the modules directory, so potentially there could be user applications using it, but again, it states that it is an internal utility class, so if someone is using it, its usage should be removed. Looking at what the Report method does, its usage doesn't look useful.

@yersan Yes, that is what I was referring to. I think it's okay too, I just wanted to make sure to point out that we saw it and we think it's okay :)

yersan:
Thanks @jamezp

PR #5834, opened 2024-01-15
[WFCORE-6662] Remove deprecated usage of ServiceController.getValue() by capturing ModelControllerClientFactory ...

yersan:
...via ServiceActivator starting the Bootable JAR and Embedded server

This PR follows up on #5824, also built on top of that one.

Jira issue: https://issues.redhat.com/browse/WFCORE-6662

### DIFF
--- a/bootable-jar/runtime/src/main/java/org/wildfly/core/jar/runtime/Server.java
+++ b/bootable-jar/runtime/src/main/java/org/wildfly/core/jar/runtime/Server.java
@@ -17,6 +17,9 @@ import java.util.concurrent.ExecutorService;
 import java.util.concurrent.Executors;
 import java.util.concurrent.Future;
 import java.util.concurrent.TimeUnit;
+import java.util.concurrent.atomic.AtomicReference;
+import java.util.function.Supplier;
+
 import org.jboss.as.controller.ControlledProcessState;
 import org.jboss.as.controller.ControlledProcessStateService;
 import org.jboss.as.controller.ModelControllerClientFactory;
@@ -31,10 +34,14 @@ import org.jboss.as.server.SystemExiter;
 import org.jboss.modules.Module;
 import org.jboss.modules.ModuleLoadException;
 import org.jboss.modules.ModuleLoader;
+import org.jboss.msc.Service;
 import org.jboss.msc.service.ServiceActivator;
+import org.jboss.msc.service.ServiceActivatorContext;
+import org.jboss.msc.service.ServiceBuilder;
 import org.jboss.msc.service.ServiceContainer;
 import org.jboss.msc.service.ServiceController;
-import org.jboss.msc.value.Value;
+import org.jboss.msc.service.StartContext;
+import org.jboss.msc.service.StopContext;
 import org.wildfly.core.jar.runtime._private.BootableJarLogger;
 
 /**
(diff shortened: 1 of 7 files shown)


bstansberry:
@yersan Cool!  I filed https://issues.redhat.com/browse/WFCORE-6659 for this and gave it a shot in https://github.com/wildfly/wildfly-core/commit/303beed34ce451e7f75a95ebb88d95603d6a8e8b, but it failed CI so I shelved it. It looks like yours here works. :)

I'll review it after I get unburied a bit from PTO.

yersan:
@bstansberry oohh, I missed the https://issues.redhat.com/browse/WFCORE-6659 notification after my PTOs :(

It looks like there is still some stuff to do with this PR, after taking a quick look at your changes, this one is pretty similar, although this one uses the null value on the notifier to establish when the dependent simple services that provide the (ProcessStateNotifier or ModelControllerClientFactory) are up. In your PR, you are using a Future.

The `ProcessStateNotifier` for the `Server` and `EmbeddedServer` is not accessed via the `notifierRef` when the service that provides its value has not been started yet, but the ModelControllerClientFactory could be accessed under this circumstance on the `establishModelControllerClient` private method.

This PR also assumes that there are no service reloads in between once the `Server` and `EmbeddedServer` have started. That could be an issue, but I think only for the `modelControllerClient`.

This PR has the issue that it is not correctly cleaning up the `clientFactorySvcCaptureRef` / `notifierRef` when the dependent service is down. For the `notifierRef` I think it is optional, but that could be an issue for `clientFactorySvcCaptureRef`.

I think I can fix it by using the intermediate service that provides the value of those `AtomicReference` notifiers, setting them to null if the dependent service goes down. That would require keeping this intermediate service up until the dependent ones are still up.

bootable-jar/runtime/src/main/java/org/wildfly/core/jar/runtime/Server.java
```diff
@@ -80,7 +81,8 @@ private Server(String[] cmdargs, Properties systemProps,
         this.systemEnv = systemEnv;
         this.moduleLoader = moduleLoader;
         this.shutdownHandler = shutdownHandler;
-        this.clientFactorySvcCaptureRef = new AtomicReference<>(null);
+        this.clientFactorySvcCaptureRef = new AtomicReference<>();
+        this.notifierRef = new AtomicReference<>();
```
bstansberry, line 85: This shouldn't need to be a field, as its use is confined to one method and the anonymous Service impl created by a call from that method.

bstansberry:
> The `ProcessStateNotifier` for the `Server` and `EmbeddedServer` is not accessed via the `notifierRef` when the service that provides its value has not been started yet, but the ModelControllerClientFactory could be accessed under this circumstance on the `establishModelControllerClient` private method.

I initially did mine without the Future and got test failures because the getModelControllerClient() call  in EmbedServerHandler.doHandle would return null. That was because the `establishModelControllerClient` call in StandaloneServerImpl.start would not create a client, because the clientFactorySvcCaptureRef.get() call would return null. That's what led to my comment at https://issues.redhat.com/browse/WFCORE-6659?focusedId=23738156&page=com.atlassian.jira.plugin.system.issuetabpanels%3Acomment-tabpanel#comment-23738156

bstansberry:
> Partially resolved, it is still racy between the stop method and getting the values from the notifiers (it was also racy before with the original code where `createSuperUserClient` could be called when the service was down, I guess there are low probabilities, but that's racy). This version also does not remove the intermediate services as soon as the value from the dependents is received. So, it still needs some work.

Yes, the createSuperUserClient call should only happen if the clientFactorySvcCaptureRef.get() doesn't return null.

Note that it's ok for ModelControllerClient use to fail in the middle of a reload. If you create a client to a remote server and someone (including yourself) reloads the server, that client isn't expected to work through that. The CLI uses a client that is designed to handle that. The client we return from getModelControllerClient does as well.

PR #5836, opened 2024-01-16
WFCORE-6669 Various tests fail when env specifies JDK_JAVA_OPTIONS or JAVA_TOOL_OPTIONS

pferraro:
https://issues.redhat.com/browse/WFCORE-6669

### DIFF
--- a/testsuite/scripts/src/test/java/org/wildfly/scripts/test/CliScriptTestCase.java
+++ b/testsuite/scripts/src/test/java/org/wildfly/scripts/test/CliScriptTestCase.java
@@ -50,8 +50,7 @@ public class CliScriptTestCase extends ScriptTestCase {
         StringBuilder builder = new StringBuilder();
         // Read the output lines which should be valid DMR
         for (String line : script.getStdout()) {
-            // Skip lines like: "Picked up _JAVA_OPTIONS: ..."
-            if (line.startsWith("Picked up _JAVA_") || line.startsWith("WARNING")) {
+            if (line.contains("Picked up JDK_JAVA_OPTIONS:") || line.contains("Picked up JAVA_TOOL_OPTIONS:") || line.startsWith("WARNING")) {
                 continue;
             }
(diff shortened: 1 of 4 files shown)


testsuite/scripts/src/test/java/org/wildfly/scripts/test/CliScriptTestCase.java
```diff
@@ -50,8 +50,7 @@ private static ModelNode outputToModelNode(final ScriptProcess script) throws IO
         StringBuilder builder = new StringBuilder();
         // Read the output lines which should be valid DMR
         for (String line : script.getStdout()) {
-            // Skip lines like: "Picked up _JAVA_OPTIONS: ..."
-            if (line.startsWith("Picked up _JAVA_") || line.startsWith("WARNING")) {
```
yersan, line 54: _JAVA_OPTIONS is still supported on JDK11, JDK17 and JDK21. I have not found the JEP that is removing it, as far as I understand is not supported, but in any case

Are we sure that removing it from these tests is not going to break any other test environment? Wouldn't be better if we also keep "Picked up _JAVA_" as a valid possibility?

pferraro: Per https://bugs.openjdk.org/browse/JDK-8185446:
> "JDK 9 supports a new environment variable `JDK_JAVA_OPTIONS` ... The new environment variable has several advantages over the legacy/_unsupported_ `_JAVA_OPTIONS` environment variable ..."

> Are we sure that removing it from these tests is not going to break any other test environment?

As far as I am aware, I am the only dev who ever had an issue with these tests failing on JDK8.
See: https://issues.redhat.com/browse/WFCORE-5043

yersan (approved):
No other feedback/change, so approving as it is now.

yersan:
Thanks @pferraro

PR #5838, opened 2024-01-17
WFCORE-6668 read-feature-description operation needs to include stability per feature for use by WildFly Glow

pferraro:
https://issues.redhat.com/browse/WFCORE-6668

Adds stability level to the descriptions for resources, capabilities, operations, attributes, and notifications.

### DIFF
--- a/controller/src/main/java/org/jboss/as/controller/descriptions/DefaultNotificationDescriptionProvider.java
+++ b/controller/src/main/java/org/jboss/as/controller/descriptions/DefaultNotificationDescriptionProvider.java
@@ -9,10 +9,13 @@ import static org.jboss.as.controller.NotificationDefinition.DataValueDescriptor
 import static org.jboss.as.controller.descriptions.ModelDescriptionConstants.NOTIFICATION_DATA_TYPE;
 import static org.jboss.as.controller.descriptions.ModelDescriptionConstants.DESCRIPTION;
 import static org.jboss.as.controller.descriptions.ModelDescriptionConstants.NOTIFICATION_TYPE;
+import static org.jboss.as.controller.descriptions.ModelDescriptionConstants.STABILITY;
 
 import java.util.Locale;
 import java.util.ResourceBundle;
 
+import org.jboss.as.controller.NotificationDefinition;
+import org.jboss.as.version.Stability;
 import org.jboss.dmr.ModelNode;
 
 /**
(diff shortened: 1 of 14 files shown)


controller/src/main/java/org/jboss/as/controller/descriptions/DefaultOperationDescriptionProvider.java
```diff
@@ -126,9 +130,22 @@ public DefaultOperationDescriptionProvider(final String operationName,
         this.deprecationData = deprecationData;
         this.replyParameters = replyParameters;
         this.accessConstraints = accessConstraints != null ? accessConstraints : Collections.<AccessConstraintDefinition>emptyList();
+        this.stability = Stability.DEFAULT;
```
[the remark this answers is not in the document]
pferraro: > I think this could be suspicious, this description provider is somehow tight to a specific operation definition; the constructor is using an operation name. An Operation definition is Feature aware, since it implements it.

Frankly, I would normally have deprecated this constructor (and the others).  However, there seems to be a least use of this constructor in org.wildfly.core:wildfly-host-controller, but its usage seem dubious.  I'm in the process of tracking down the author to verify.

> However, in this constructor, the Operation definition feature stream is unknown because it uses an Operation name. I am wondering if that could lead to problems where an Operation running on a different stability level above the default one, could be using this constructor.

The same could be said for any aspect of the operation description.  Using any of the existing constructors, a user could describe their operation in a manner inconsistent with the OperationDefinition.

controller/src/main/java/org/jboss/as/controller/descriptions/DefaultNotificationDescriptionProvider.java
```diff
@@ -23,13 +26,15 @@
 public class DefaultNotificationDescriptionProvider implements DescriptionProvider {
 
     private final String notificationType;
+    private final Stability stability;
     private final ResourceDescriptionResolver descriptionResolver;
     private final DataValueDescriptor dataValueDescriptor;
 
-    public DefaultNotificationDescriptionProvider(final String notificationType,
+    public DefaultNotificationDescriptionProvider(final NotificationDefinition definition,
```
bstansberry, line 33: This is public API so to be cautious we should add a constructor and deprecate the existing one.

Hmm, probably better is to not add or change a constructor here and just statically write DEFAULT in the description. Then deprecate the entire class and copy the code into a private class in NotificationDefinition, leaving this as temporary cruft. I don't see any hook that would allow a user to use any other class than this as the DescriptionProvider.

jfdenise:
@pferraro , I gave a try to the branch, I notice that what galleon is using (the `read-feature-description` operation) doesn't contain the stability level.

The "read-resource-description" contains it, but I am observing that the advertised stability level by the subsystem is the one that has been set when starting the server (--stability=xxx) and not the one declared by the extension. So when starting without any stability level, I got "community" for all extensions. When starting with experimental, I got experimental for all.

Another observation, the attributes contained in the subsystem always declare the stability as default.

pferraro:
@jfdenise yes - you are right. I did not realize that the read-feature-description only populates its result from specific contents of a given DescriptionProvider.  Give me a few minutes to complete this.

pferraro:
@jfdenise Can you give this another look?  The read-feature-description results should now report the "stability" per feature and param.
I've modify MRR.getStability() to report the stability of the associated ResourceDefinition, rather than the stability of the process, which should address the behavior you reported above.

pferraro:
Looks like org.jboss.as.test.integration.management.api.ReadFeatureDescriptionTestCase test in downstream testsuite needs to be updated to accommodate this change.

controller/src/main/java/org/jboss/as/controller/descriptions/DefaultNotificationDescriptionProvider.java
pferraro: Point taken.  I'll clean this up.

pferraro:
Adding the hold label.  I need to think a bit more about the consequence of the changes to MRR.getStability().

pferraro:
this needs a bit of rework.

PR #5840, opened 2024-01-18
[WFCORE-5917] Use variable expansion replacement instead of pattern r…

yersan:
…emoving to avoid duplicate jboss.server.base.dir options

Jira issue: https://issues.redhat.com/browse/WFCORE-5917

### DIFF
--- a/core-feature-pack/common/src/main/resources/content/bin/standalone.sh
+++ b/core-feature-pack/common/src/main/resources/content/bin/standalone.sh
@@ -212,8 +212,9 @@ fi
 # clean server opts
 for var in $SERVER_OPTS
 do
-   if [ "${var//"-Djboss.server.base.dir"/}" != "${var}" ]; then
-     SERVER_OPTS="${SERVER_OPTS//$var/}"
+   v=`echo "${var}" | sed 's/-Djboss.server.base.dir//g'`
+   if [ "${v}" != "${var}" ]; then
+     SERVER_OPTS=`echo "${SERVER_OPTS}" | sed "s|${var}||g"`
    fi
 done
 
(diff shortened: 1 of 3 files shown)


jamezp (changes requested):
I'm not sure this is correct. This will remove the `-Djboss.server.base.dir` from the `SERVER_OPTS`. While the original way this was done doesn't seem correct either, I don't think this is.

Really having it defined twice isn't a huge issue. However, we need to be able to override it from the command line.

yersan:
>This will remove the `-Djboss.server.base.dir` from the `SERVER_OPTS`.

This variable is not exported so is not visible to the server process. After removing the option, which is duplicated later in the server command launch (see that it explicitly uses `-Djboss.server.base.dir=\""$JBOSS_BASE_DIR"\" \`, the server needs to be stopped and started again to use it, which will initialize it back to the original value. I mean, it is unused after removing the duplicated option on this line.

The domain.sh has the same issue.

> 
> Really having it defined twice isn't a huge issue.

Yes, I agree, it is just the point of having it twice for the same server launch command line. 


>  However, we need to be able to override it from the command line.

This change does not affect this calculation. This is done when the options are consolidated above this change, L140.

lvydra (approved):
Thanks @yersan, now it's working as it was formerly intended. @jamezp At the point of removing `-Djboss.server.base.dir` from the `SERVER_OPTS,` that one should be already processed as part of opts consolidation.

jamezp (approved):
Sorry, I must have been looking at something wrong in my testing. This does seem work, my apologies.

yersan:
converting to draft, since although the approach is approved, we found we need to apply the same to `domain.sh` and potentially, look at batch ps1 scripts too, and or check if we can add a validation test to it.

PR #5841, opened 2024-01-18
[WFCORE-6531] Wrap server options in double quotes

yersan:
Jira issue: https://issues.redhat.com/browse/WFCORE-6531

The SERVER_OPTS were being wrapped in single quotes, which does not prevent a misuse using eval. Wrap them in double quotes.

### DIFF
--- a/core-feature-pack/common/src/main/resources/content/bin/standalone.sh
+++ b/core-feature-pack/common/src/main/resources/content/bin/standalone.sh
@@ -34,7 +34,7 @@ do
           shift
           break;;
       *)
-          SERVER_OPTS="$SERVER_OPTS '$1'"
+          SERVER_OPTS="$SERVER_OPTS \"$1\""
           ;;
     esac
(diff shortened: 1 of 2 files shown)


core-feature-pack/common/src/main/resources/content/bin/standalone.sh
```diff
@@ -144,7 +144,7 @@ if $linux; then
     for var in $CONSOLIDATED_OPTS
     do
        # Remove quotes
-       p=`echo $var | tr -d "'"`
+       p=`echo $var | tr -d "'"` | tr -d "\""
```
yersan, line 147: Ups, this is wrong, it was fixed in the domain.sh version but I forgot to correct it on the standalone one.

The last `tr` must be inside the backticks

jamezp (approved):

yersan:
Thanks @jamezp

PR #5842, opened 2024-01-19
WFCORE-6668 Ensure read-feature-description results contain stability as reported by associated description for use by WildFly Glow

pferraro:
https://issues.redhat.com/browse/WFCORE-6668

### DIFF
--- a/controller/src/main/java/org/jboss/as/controller/AbstractControllerService.java
+++ b/controller/src/main/java/org/jboss/as/controller/AbstractControllerService.java
@@ -317,7 +317,7 @@ public abstract class AbstractControllerService implements Service<ModelControll
         final ModelControllerImpl controller = new ModelControllerImpl(container, target,
                 rootResourceRegistration,
                 new ContainerStateMonitor(container),
-                configurationPersister, processType, runningModeControl, prepareStep,
+                configurationPersister, processType, stability, runningModeControl, prepareStep,
                 processState, executorService, expressionResolver, authorizer, securityIdentitySupplier, auditLogger, notificationSupport,
                 bootErrorCollector, createExtraValidationStepHandler(), capabilityRegistry, getPartialModelIndicator(),
                 instabilityListener != null ? instabilityListener.get() : null);
--- a/controller/src/main/java/org/jboss/as/controller/AbstractOperationContext.java
+++ b/controller/src/main/java/org/jboss/as/controller/AbstractOperationContext.java
@@ -82,6 +82,7 @@ import org.jboss.as.controller.registry.OperationEntry;
 import org.jboss.as.controller.registry.Resource;
 import org.jboss.as.core.security.AccessMechanism;
 import org.jboss.as.protocol.StreamUtils;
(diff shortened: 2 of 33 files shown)


jfdenise:
@pferraro , I tried and I don't see what I was expecting.
I have batch-jberet to be a fake Prevew feature:

    @Override
    public Stability getStability() {
        return Stability.PREVIEW;
    }

I am building the feature-pack with preview as minimum, provisioning with preview. Starting the server with --stability=preview, the reported stability level for `/subsystem=batch-jberet:read-feature-description` is `default`. I was expecting it to be `preview`

I am perhaps missing something.

pferraro:
> @Override
> public Stability getStability() {
>     return Stability.PREVIEW;
> }

Which class did you override?

pferraro:
Ah - you must have overridden the Stability of the extension.  I'll need to ensure that the resource definition of the extension reflects its stability level so that the corresponding MMR.getStability() returns the expected value.

jfdenise:
@pferraro, yes, the extension. In such a case, should all defined resources/attributes located in the extension be at least at the extension stability level? It could apply to any hierarchy, the stability, if not redefined locally, being inherited from the parent.

pferraro:
Latest changes seem to have caused the feature spec generator to hang.  I will nvestigate/
A complication with the read-feature-description: The /extension=* resource description will always have DEFAULT stability - since this is a wildcard resource definition.
I can experiment with creating override definitions for extensions with non-default stability, but I'm not sure how well this will reflected in the read-feature-description operation.

> @pferraro, yes, the extension. In such a case, should all defined resources/attributes located in the extension be at least at the extension stability level? It could apply to any hierarchy, the stability, if not redefined locally, being inherited from the parent.

That said, with the above changes, the corresponding /subsystem=* resource definitions and their child resources, will all inherit the max stability of the extension that created them.

Unfortunately, for objects associated with resource definitions; e.g. like AttributeDefinitions, OperationDefinitions, Notifications, etc.; cannot get the same treatment, since none of these abstractions are encapsulated by interfaces that can be reasonably decorated.

pferraro:
@jfdenise OK - this should be working now.

pferraro:
I'm sorting through the test failures, but at least we are no longer hanging while generating feature specs...

jfdenise:
@pferraro , thank-you! I can see the stability level at the extension level and the sub resources inheriting it.

jfdenise (approved):
Fine with me with respect to Galleon provisioning.

yersan (approved):

yersan:
@pferraro / @jfdenise Are we fine treating this as a task and not as a feature request? This is going to add a new field exposing additional information and its requirement is driven by a feature request, so I ran into the doubt when I was going to merge it.

Adding a hold label until it gets confirmed.

jfdenise:
@yersan , I would say yes, this operation is not exposed, it is there to be consumed by Galleon.

yersan:
Thanks @pferraro @jfdenise @bstansberry

PR #5847, opened 2024-01-23
WFCORE-6490 worker attribute of /subsystem=remoting should not have a default value

pferraro:
https://issues.redhat.com/browse/WFCORE-6490

Includes prerequisites:
WFCORE-6673 Create SPI module for IO subsystem
https://issues.redhat.com/browse/WFCORE-6673

WFCORE-6670 Drop io subsystem parsers for 1.0 version
https://issues.redhat.com/browse/WFCORE-6670

WFCORE-6671 Drop obsolete subsystem transformations from IO subsystem
https://issues.redhat.com/browse/WFCORE-6671


yersan:
Hello @fl4via , this is a large PR related to remoting/io subsystems, I don't think the changes are likely to conflict, but in any case, it would require feedback from your side, an overall point of view would be enough. Just pinging you here in case you have missed the review request. Thanks!

pferraro:
I need to rebase this branch and drop the last 2 commits now that these test issues have since been resolved.

fl4via:
I am reviewing this PR now. @ropalka is also having a look

testsuite/domain/src/test/java/org/jboss/as/test/integration/domain/suites/ManagementReadXmlTestCase.java
```diff
@@ -52,6 +52,7 @@
  *
  * @author Emmanuel Hugonnet (c) 2022 Red Hat, Inc.
  */
+@org.junit.Ignore("WFCORE-6679")
```
fl4via, line 55: @pferraro please rebase as this test is fixed

testsuite/manualmode/src/test/java/org/jboss/as/test/manualmode/management/persistence/YamlExtensionTestCase.java
```diff
@@ -42,6 +42,7 @@
  *
  * @author Emmanuel Hugonnet (c) 2021 Red Hat, Inc.
  */
+@org.junit.Ignore("WFCORE-6680")
```
fl4via, line 45: @pferraro please rebase as this test is fixed

io/subsystem/src/main/java/org/wildfly/extension/io/IOSubsystemAdd.java
```diff
@@ -23,11 +23,6 @@
  */
 class IOSubsystemAdd extends AbstractAddStepHandler {
 
-    static final IOSubsystemAdd INSTANCE = new IOSubsystemAdd();
```
fl4via, line 26: My question might be bogus, since I'm not familiar with the latest improvements in WF code... but anyway, why exactly are we removing INSTANCE here?

fl4via (changes requested):
I am not sure I agree this is a bug. IMO, this is a new feature or an enhancement, as it is going to require explicit configuration of worker. So, it changes how you are configuring the server. You may affirm that the drive behind it is that you don't agree it should have a magic default worker, but in the end, still an enhancement.

testsuite/domain/src/test/java/org/jboss/as/test/integration/domain/suites/ManagementReadXmlTestCase.java
pferraro: Will do.

testsuite/manualmode/src/test/java/org/jboss/as/test/manualmode/management/persistence/YamlExtensionTestCase.java
pferraro: Will do.

io/subsystem/src/main/java/org/wildfly/extension/io/IOSubsystemAdd.java
pferraro: Because it is never referenced, nor would it need to be.

pferraro:
> I am not sure I agree this is a bug. IMO, this is a new feature or an enhancement, as it is going to require explicit configuration of worker. So, it changes how you are configuring the server.

The server *already* requires explicit configuration of a worker.  That has not changed.

>  You may affirm that the drive behind it is that you don't agree it should have a magic default worker, but in the end, still an enhancement.

It is incorrect for a capability reference to use a default value.  That is a bug.  This otherwise misconfiguration causes failures at runtime that should have been detected during model validation.

fl4via:
> > I am not sure I agree this is a bug. IMO, this is a new feature or an enhancement, as it is going to require explicit configuration of worker. So, it changes how you are configuring the server.
> 
> The server _already_ requires explicit configuration of a worker. That has not changed.
> 
Okay, that clears things up for me.

> > You may affirm that the drive behind it is that you don't agree it should have a magic default worker, but in the end, still an enhancement.
> 
> It is incorrect for a capability reference to use a default value. That is a bug. This otherwise misconfiguration causes failures at runtime that should have been detected during model validation.

Ditto. Thanks for the clarification.

fl4via (approved):
There are some CI failures, I know you will look into that. As far as I'm concerned regarding the code, it looks very good, and all my questions were clarified.
Thanks for the hard work @pferraro !

pferraro:
Thank you so much for reviewing, @fl4via 
The newly failing test was previous ignored due to WFCORE-6679.  I suspect the test inputs just need to be updated to reflect the changes made here, but I will investigate further later today.

yersan:
@ropalka You were requested to review this one, is this ok to go? It was already approved by @fl4via but I don't know if it needs any consensus between you both.

ropalka (approved):
Thank you very much for this big enhancement @pferraro !

yersan:
Thanks @fl4via @pferraro @ropalka

PR #5851, opened 2024-01-29
[WFCORE-6679]: ManagementReadXmlTestCase inputs are too fragile.

ehsavoie:
* Generating the servers configuration through scripts to make them less fragile.

Jira: https://issues.redhat.com/browse/WFCORE-6679


testsuite/domain/src/test/java/org/jboss/as/test/integration/domain/ManagementReadXmlTestCase.java
```diff
@@ -59,8 +61,89 @@ public class ManagementReadXmlTestCase {
     private static DomainLifecycleUtil domainPrimaryLifecycleUtil;
     private static DomainLifecycleUtil domainSecondaryLifecycleUtil;
 
+    private static void prepareExpectedServerConfiguration(String server, boolean propertyOne, int offset) throws Exception {
+        String serverConfigFileName = "expected-main-" + server + ".xml";
+        Path serverHome = new File(WildFlySecurityManager.getPropertyPrivileged("jboss.home", "toto")).toPath();
+        Path configurationDir = serverHome.resolve("standalone").resolve("configuration");
+        Files.createDirectories(configurationDir);
+        Path referenceConfiguration = configurationDir.resolve(serverConfigFileName);
+        Files.copy(new File("target/test-classes", "base-server-config.xml").toPath(), referenceConfiguration, REPLACE_EXISTING);
```
yersan, line 70: Mmmhh, I'm not sure if it will work always. Should not we separate the path names by using the default path name separator instead of "/"?

I mean, something like:
`Files.copy(Paths.get("target").resolve("test-classes").resolve("base-server-config.xml"), referenceConfiguration, REPLACE_EXISTING);`

ehsavoie: oups sorry, fixing that

ehsavoie:
@yersan done

testsuite/domain/src/test/java/org/jboss/as/test/integration/domain/suites/ManagementReadXmlTestCase.java
```diff
@@ -58,8 +62,89 @@ public class ManagementReadXmlTestCase {
     private static DomainLifecycleUtil domainPrimaryLifecycleUtil;
     private static DomainLifecycleUtil domainSecondaryLifecycleUtil;
 
+    private static void prepareExpectedServerConfiguration(String server, boolean propertyOne, int offset) throws Exception {
+        String serverConfigFileName = "expected-main-" + server + ".xml";
+        Path serverHome = new File(WildFlySecurityManager.getPropertyPrivileged("jboss.home", "toto")).toPath();
+        Path configurationDir = serverHome.resolve("standalone").resolve("configuration");
```
yersan, line 68: We are running domain mode based tests, so better if you use the domain configuration directories instead of the standalone one.

In addition, you should find out how to do it via `domainPrimaryLifecycleUtil.getConfiguration().getJbossHome()` for example, to use the currently running instance and to avoid any side effect by using the shared target/wildfly-core server.

testsuite/domain/src/test/java/org/jboss/as/test/integration/domain/suites/ManagementReadXmlTestCase.java
```diff
@@ -58,8 +62,89 @@ public class ManagementReadXmlTestCase {
...
+        Files.createDirectories(configurationDir);
+        Path referenceConfiguration = configurationDir.resolve(serverConfigFileName);
+        Files.copy(new File("target").toPath().resolve("test-classes").resolve("base-server-config.xml"), referenceConfiguration, REPLACE_EXISTING);
+        try (CLIWrapper cli = new CLIWrapper(false)) {
+            cli.sendLine("embed-server --admin-only --server-config=" + serverConfigFileName + " --jboss-home=" + serverHome.toAbsolutePath().toString());
```
yersan, line 73: serverHome.toAbsolutePath().toString()

the toString() is redundant here.

testsuite/domain/src/test/resources/base-server-config.xml
```diff
@@ -0,0 +1,71 @@
+<?xml version="1.0" encoding="UTF-8"?>
```
yersan, line 1: I think this could be removed and used as a base for the domain-configs/domain-standard.xml which is indeed the base configuration used to start the domain for DomainTestSupport

yersan (changes requested):
Added some comments, but in any case, it seems current failures are relevant to this PR

yersan:
@ehsavoie ManagementReadXmlTestCase has failed repeatedly in Windows, could you take a look?

yersan (approved):

yersan:
Thanks @ehsavoie

PR #5860, opened 2024-02-14
[WFCORE-6503]:Add support for unmanaged deployments with YAML extension.

ehsavoie:
* checking that the YAML deployment is unmanaged.
* adding the unmanaged deployment to the list of operations
* adding some light esting on this

Jira: https://issues.redhat.com/browse/WFCORE-6503
      https://issues.redhat.com/browse/WFLY-19442
Proposal: https://github.com/wildfly/wildfly-proposals/pull/554
Documenation PR: https://github.com/wildfly/wildfly/pull/17970

### DIFF
--- a/controller/src/main/java/org/jboss/as/controller/persistence/yaml/YamlConfigurationExtension.java
+++ b/controller/src/main/java/org/jboss/as/controller/persistence/yaml/YamlConfigurationExtension.java
@@ -6,11 +6,17 @@ package org.jboss.as.controller.persistence.yaml;
 
 import static org.jboss.as.controller.client.impl.AdditionalBootCliScriptInvoker.CLI_SCRIPT_PROPERTY;
 import static org.jboss.as.controller.descriptions.ModelDescriptionConstants.ADD;
+import static org.jboss.as.controller.descriptions.ModelDescriptionConstants.BYTES;
+import static org.jboss.as.controller.descriptions.ModelDescriptionConstants.DEPLOYMENT;
+import static org.jboss.as.controller.descriptions.ModelDescriptionConstants.EMPTY;
+import static org.jboss.as.controller.descriptions.ModelDescriptionConstants.HASH;
+import static org.jboss.as.controller.descriptions.ModelDescriptionConstants.INPUT_STREAM_INDEX;
 import static org.jboss.as.controller.descriptions.ModelDescriptionConstants.NAME;
 import static org.jboss.as.controller.descriptions.ModelDescriptionConstants.OP;
 import static org.jboss.as.controller.descriptions.ModelDescriptionConstants.OP_ADDR;
 import static org.jboss.as.controller.descriptions.ModelDescriptionConstants.REMOVE;
 import static org.jboss.as.controller.descriptions.ModelDescriptionConstants.UNDEFINE_ATTRIBUTE_OPERATION;
+import static org.jboss.as.controller.descriptions.ModelDescriptionConstants.URL;
 import static org.jboss.as.controller.descriptions.ModelDescriptionConstants.VALUE;
 import static org.jboss.as.controller.descriptions.ModelDescriptionConstants.WRITE_ATTRIBUTE_OPERATION;
 import static org.jboss.as.controller.logging.ControllerLogger.MGMT_OP_LOGGER;
@@ -28,6 +34,7 @@ import java.util.Collections;
 import java.util.HashMap;
 import java.util.HashSet;
 import java.util.Iterator;
+import java.util.LinkedHashMap;
 import java.util.List;
 import java.util.ListIterator;
 import java.util.Map;
@@ -77,7 +84,9 @@ public class YamlConfigurationExtension implements ConfigurationExtension {
     private boolean needReload;
     private Path[] files;
     private final List<Map<String, Object>> configs = new ArrayList<>();
+    private final Map<String, Object> deployments = new LinkedHashMap<>();
     private static final String[] EXCLUDED_ELEMENTS = {"deployment", "extension", "deployment-overlay", "path"};
+    public static final Set<String> MANAGED_CONTENT_ATTRIBUTES = Set.of(INPUT_STREAM_INDEX, HASH, BYTES, URL, EMPTY);
 
     @SuppressWarnings("unchecked")
     public YamlConfigurationExtension() {
@@ -114,11 +123,15 @@ public class YamlConfigurationExtension implements ConfigurationExtension {
                         boolean isPresent = config.containsKey(excluded);
                         if (isPresent) {
                             Object value = config.remove(excluded);
-                            String message = MGMT_OP_LOGGER.ignoreYamlElement(excluded);
-                            if (value != null) {
-                                message = message + MGMT_OP_LOGGER.ignoreYamlSubElement(yaml.dump(value).trim());
+                            if (value instanceof Map && DEPLOYMENT.equals(excluded)) {
+                                deployments.putAll((Map<String, Object>) value);
+                            } else {
+                                String message = MGMT_OP_LOGGER.ignoreYamlElement(excluded);
+                                if (value != null) {
+                                    message = message + MGMT_OP_LOGGER.ignoreYamlSubElement(yaml.dump(value).trim());
+                                }
+                                MGMT_OP_LOGGER.warn(message);
                             }
-                            MGMT_OP_LOGGER.warn(message);
                         }
                     }
                     parsedFiles.add(file.toAbsolutePath().toString());
@@ -159,6 +172,9 @@ public class YamlConfigurationExtension implements ConfigurationExtension {
         }
         for (Map<String, Object> config : configs) {
             processResource(PathAddress.EMPTY_ADDRESS, new HashMap<>(config), rootRegistration, rootRegistration, xmlOperations, postExtensionOps, false);
+        }
+        for (Map.Entry<String, Object> deployment : deployments.entrySet()) {
+            processUnmanagedDeployments(rootRegistration, deployment, xmlOperations, postExtensionOps);
         }
         this.configs.clear();
         needReload = true;
@@ -395,13 +411,22 @@ public class YamlConfigurationExtension implements ConfigurationExtension {
     @SuppressWarnings("unchecked")
     private void processAttributes(PathAddress address, ImmutableManagementResourceRegistration rootRegistration, OperationEntry operationEntry, Map<String, Object> map, List<ParsedBootOp> postExtensionOps, Map<PathAddress, ParsedBootOp> xmlOperations) {
         Set<AttributeDefinition> attributes = new HashSet<>();
+        Set<String> attributeNames = new HashSet<>();
         for (AttributeAccess attributeAccess : rootRegistration.getAttributes(address).values()) {
             if (attributeAccess.getStorageType() == AttributeAccess.Storage.CONFIGURATION) {
                 AttributeDefinition def = attributeAccess.getAttributeDefinition();
                 if (def != null) {
                     if (!def.isResourceOnly()) {
                         attributes.add(def);
+                        attributeNames.add(def.getName());
                     }
+                }
+            }
+        }
+        for (AttributeDefinition def : operationEntry.getOperationDefinition().getParameters()) {
+            if (def != null && ! attributeNames.contains(def.getName())) {
+                if (!def.isResourceOnly()) {
+                    attributes.add(def);
                 }
             }
         }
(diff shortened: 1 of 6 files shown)


mnovak1:
For WARN message it seems to be logging some null values. For example:  
```09:46:09,662 WARN  [org.jboss.as.controller.management-operation] (main) WFLYCTL0506: The yaml element null
 is ignored
09:46:09,663 WARN  [org.jboss.as.controller.management-operation] (main) WFLYCTL0506: The yaml element org.jboss.as.failure: {module: org.jboss.as.failure}
 is ignored
09:46:09,664 WARN  [org.jboss.as.controller.management-operation] (main) WFLYCTL0506: The yaml element null
 is ignored
```  
for yaml:  

```
wildfly-configuration:
    extension:
      org.jboss.as.failure:
        module: org.jboss.as.failure
    socket-binding-group:
        standard-sockets:
          default-interface: public
          port-offset: ${jboss.socket.binding.port-offset:0}
          socket-binding:
              http:
                interface: public
              https:
          remote-destination-outbound-socket-binding:
            mail-snmt:  
              host: foo
              port: 8081
            foo2:
              host: foo2
              port: 8082
```  
I should log the name of yaml tag which is ignored.

controller/src/main/java/org/jboss/as/controller/logging/ControllerLogger.java
```diff
@@ -3752,4 +3752,13 @@ OperationFailedRuntimeException capabilityAlreadyRegisteredInContext(String capa
 
     @Message(id = 504, value = "The operation %s is not defined for resource %s.")
     UnsupportedOperationException missingOperationForResource(String op, String address);
+
+    @Message(id = 505, value = "Unsuported deployment yaml file %s with attributes %s")
+    IllegalArgumentException unsupportedDeployment(String deployment, Set<String> attribues);
+
+    @Message(id = 506, value = "The yaml element \"%s\" and its sub-elements are ignored.")
```
mnovak1, line 3759: Dummy issue but can be space added at the end of message like:
`@Message(id = 506, value = "The yaml element \"%s\" and its sub-elements are ignored. ")`

to avoid: `ignored.Thus ignoring`  in:  
```
12:59:42,222 WARN  [org.jboss.as.controller.management-operation] (main) WFLYCTL0506: The yaml element "extension" and its sub-elements are ignored.Thus ignoring element "org.jboss.as.failure: {module: org.jboss.as.failure}".
```

controller/src/main/java/org/jboss/as/controller/logging/ControllerLogger.java
```diff
@@ -3682,7 +3682,7 @@ OperationFailedRuntimeException capabilityAlreadyRegisteredInContext(String capa
     IllegalArgumentException noResourceForUndefiningAttribute(String attribute, String address);
 
     @LogMessage(level = WARN)
-    @Message(id = 490, value = "You have defined a resource for address %s without any attributes, doing nothing")
+    @Message(id = 490, value = " A YAML resource has been defined for the address %s without any attribute. No actions will be taken.")
```
yersan, line 3685: I guess this white space at the beginning is not intentional for this message that contains an ID, could you confirm @ehsavoie

ehsavoie: correct

yersan:
Hello @bstansberry, you added "Feature" and "missing-reqs" labels to this PR. So far we are treating this as an enhancement, do you want this to be treated as a Feature Request instead? There is more information about this on the Jira.

controller/src/main/java/org/jboss/as/controller/ModelControllerImpl.java
```diff
@@ -518,7 +518,7 @@ boolean boot(final List<ModelNode> bootList, final OperationMessageHandler handl
                     headers, handler, null, managementModel.get(), control, processState, auditLogger,
                             bootingFlag.get(), true, hostServerGroupTracker, null, notificationSupport, true,
                             extraValidationStepHandler, partialModel, securityIdentitySupplier);
-            if (configExtension != null && configExtension.shouldProcessOperations(runningModeControl.getRunningMode())) {
+            if (configExtension != null && configExtension.shouldProcessOperations(runningModeControl.getRunningMode()) && (!runningModeControl.isReloaded() || runningModeControl.isApplyConfigurationExtension())) {
```
yersan, line 521: I think we should simply do this, it is difficult to track down and it took me some time to understand its meaning.

My feeling is this logic should be encapsulated inside  `configExtension.shouldProcessOperations`. I think the additional `&& (!runningModeControl.isReloaded() || runningModeControl.isApplyConfigurationExtension())` complements it by adding a new case to shouldProcessOperations.

Maybe shouldProcessOperations to take into account the whole `runningModeControl` instead of only the `runningModeControl.getRunningMode()`

controller/src/main/java/org/jboss/as/controller/RunningModeControl.java
```diff
@@ -17,6 +17,7 @@ public class RunningModeControl {
     private volatile boolean useCurrentConfig;
     private volatile String newBootFileName;
     private volatile Boolean suspend;
+    private volatile boolean applyConfigurationExtension;
```
yersan, line 20: RunningModeControl has evolved keeping more stuff than only the required information to provide control over the server's current RunningMode. However, instead of adding information on whether the configuration extensions should be applied, wouldn't be easier if we just save here the value of extensibleConfigurationPersister.hasStored()?

With this information and the other field values we could derive in  `ConfigurationExtension.shouldProcessOperations()` whether the yaml configuration should be applied again after a reload.

yersan (commented):
Added some comments to think about whether some stuff that could be moved to simply a bit the changes on the reload handlers and ModelControllerImpl, they can be evolved later, but it took me some time to figure out why we needed them.

In general, it looks pretty isolated to yaml work, so, except the additional trailing white space on the --yaml option, looks good and I think we can go with them. However, if possible, we should move the conditions added on reload handlers and ModelControllerImpl to the ConfigurationExtension.shouldProcessOperations.

testsuite/test-runner/src/main/java/org/wildfly/core/testrunner/Server.java
```diff
@@ -255,6 +261,10 @@ protected void start(PrintStream out) {
                 }
 
                 commandBuilder.setYamlFiles(yamlFiles);
+                //Forcing this to be able to test failure when no yaml file is provided
+                if(yamlFiles != null && yamlFiles.length == 0) {
+                    commandBuilder.addServerArgument("--yaml= ");
```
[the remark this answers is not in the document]
ehsavoie: It is on purpose to provoke the failure since we don't have a 'proper' command line.

controller/src/main/java/org/jboss/as/controller/RunningModeControl.java
ehsavoie: I don't really understand  what you are proposing there.

testsuite/test-runner/src/main/java/org/wildfly/core/testrunner/Server.java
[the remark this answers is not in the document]
ehsavoie: Well I removed the space

ehsavoie:
@bstansberry @yersan I've rebased on https://github.com/wildfly/wildfly-core/pull/5934 so only the feature part where unmanaged deployments are supported is here now.

controller/src/main/java/org/jboss/as/controller/RunningModeControl.java
yersan: @ehsavoie I mean, move all the logic that calculates the `applyConfigurationExtension` from the reload handler to `ConfigExtension.shouldProcessOperations()` method and instead save in the RunningModeControl the flag that indicates the persistent storage has been stored. e.g extensibleConfigurationPersister.hasStored() value.

With the RunningModeControl class field values you should be able to calculate on `ConfigExtension.shouldProcessOperations()` whether the YAML operations should be loaded again.

Understanding the logic and purpose of `applyConfigurationExtension` calculation by looking at the handlers is a bit unclear, if you prefer to keep what you currently have, at least write a java doc explaining the applyConfigurationExtension purpose and meaning on RunningModeControl accessors.

Out of curiosity, I am not sure how the stored method is invoked even when we are using a read-only configuration file policy, but, if the user writes something after reloading, all the configuration, including those derived from the YAML files, will be available from the persistent storage? I understand that's the reason why you don't want to apply the YAML config again to do not overwrite any modification done after the previous reloading

yersan:
@bstansberry I removed the feature label due to the https://issues.redhat.com/browse/WFCORE-6503 comments. That probably was a mistake since you added them after the Jira discussions. 

Just to be clear, do you want to treat this as a pure feature request instead of an enhancement that adds more value to a released feature request?

controller/src/main/java/org/jboss/as/controller/RunningModeControl.java
ehsavoie: When we are in read-only mode the changes are persisted to the standalone-boot.xml so that when you reload you keep them. That's when the store method is invoked.
When you start or restart the standalone-boot.xml is overwritten by the standalone.xml.

controller/src/main/java/org/jboss/as/controller/persistence/yaml/YamlConfigurationExtension.java
```diff
@@ -524,6 +549,25 @@ public String getCommandLineInstructions() {
...
+        OperationEntry operationEntry = rootRegistration.getOperationEntry(PathAddress.pathAddress("deployment", name), ADD);
+        if (deployment.getValue() != null && deployment.getValue() instanceof Map) {
+            Map<String, Object> attributes = (Map<String, Object>) deployment.getValue();
+            Map<String, Object> content = (Map<String, Object>) (((Iterable<? extends Object>) attributes.get("content")).iterator().next());
+            Set<String> result = content.keySet().stream().distinct().filter(MANAGED_CONTENT_ATTRIBUTES::contains).collect(Collectors.toSet());
+            if (!result.isEmpty()) {
+                throw MGMT_OP_LOGGER.unsupportedDeployment(name, result);
+            }
+            if(attributes.containsKey("enabled")) {
+                boolean enabled = (Boolean) attributes.getOrDefault("enabled", "false");
```
bstansberry, line 564: Is L563-565 cruft? This seems to be validation logic and I don't see why 'enabled' needs special handling.

L564 is confusing, as String "false" will result in ClassCastException when cast to Boolean.

L563 has already established that getOrDefault will never use the default value though, so it doesn't break anything, it's just fragile.

Cleaner is replacing the entire L563-565 block with

// Confirm we have a valid boolean 'enabled' setting, if present
(Boolean) attributes.getOrDefault("enabled", Boolean.FALSE);

But.... why is that needed, and if validation is needed a validation check that uses CCE to signal failure isn't very user friendly.

controller/src/main/java/org/jboss/as/controller/persistence/yaml/YamlConfigurationExtension.java
```diff
@@ -114,11 +123,15 @@ private void load() {
                         boolean isPresent = config.containsKey(excluded);
                         if (isPresent) {
                             Object value = config.remove(excluded);
-                            String message = MGMT_OP_LOGGER.ignoreYamlElement(excluded);
-                            if (value != null) {
-                                message = message + MGMT_OP_LOGGER.ignoreYamlSubElement(yaml.dump(value).trim());
+                            if (value != null && value instanceof Map && DEPLOYMENT.equals(excluded)) {
```
bstansberry, line 126: Nit:

value != null && value instanceof Map

can be reduced to

value instanceof Map

controller/src/main/java/org/jboss/as/controller/persistence/yaml/YamlConfigurationExtension.java
```diff
@@ -524,6 +549,25 @@ public String getCommandLineInstructions() {
         return MGMT_OP_LOGGER.argYaml();
     }
 
+    @SuppressWarnings("unchecked")
+    private void processUnmanagedDeployments(ImmutableManagementResourceRegistration rootRegistration, Map.Entry<String, Object> deployment, Map<PathAddress, ParsedBootOp> xmlOperations, List<ParsedBootOp> postExtensionOps) {
+        String name = deployment.getKey();
+        OperationEntry operationEntry = rootRegistration.getOperationEntry(PathAddress.pathAddress("deployment", name), ADD);
+        if (deployment.getValue() != null && deployment.getValue() instanceof Map) {
```
bstansberry, line 556: Nit:

This can be

if (deployment.getValue() instanceof Map) {

or, since this check was already performed in load(), simply

assert deployment.getValue() instanceof Map;

server/src/main/java/org/jboss/as/server/ServerEnvironment.java
```diff
@@ -347,11 +347,6 @@ public ServerEnvironment(final String hostControllerName, final Properties props
         javaExtDirs = getFilesFromProperty(JAVA_EXT_DIRS, props);
 
         if (launchType.equals(LaunchType.SELF_CONTAINED)) {
-            Path[] supplementalConfigurationFiles = findSupplementalConfigurationFiles(null, supplementalConfiguration);
-            ConfigurationExtension configurationExtension = ConfigurationExtensionFactory.createConfigurationExtension(supplementalConfigurationFiles);
-            if (configurationExtension != null) {
-                configInteractionPolicy = configurationExtension.shouldProcessOperations(runningModeControl) ? ConfigurationFile.InteractionPolicy.READ_ONLY : configInteractionPolicy;
```
bstansberry, line 353: Note to reviewers:

The configInteractionPolicy var is never read in this branch, making the assignment here pointless and thus the other removed lines pointless as well.

There are uses of the var in other branches.

bstansberry (changes requested):

bstansberry:
@yersan Sorry for not answering your questions until now. This is new user controllable behavior so it's a feature. We should communicate that to WF users by using Feature Request issue type. It's been validated as part of validating the earlier YAML work, so once code review stuff comments here are addressed by @ehsavoie this can be merged.

controller/src/main/java/org/jboss/as/controller/persistence/yaml/YamlConfigurationExtension.java
ehsavoie: Yes correct, those lies are cruft.

controller/src/main/java/org/jboss/as/controller/persistence/yaml/YamlConfigurationExtension.java
```diff
@@ -524,6 +549,25 @@ public String getCommandLineInstructions() {
...
+            Map<String, Object> attributes = (Map<String, Object>) deployment.getValue();
+            Map<String, Object> content = (Map<String, Object>) (((Iterable<? extends Object>) attributes.get("content")).iterator().next());
```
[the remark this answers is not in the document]
ehsavoie: Fixing that as it is only for validation purpose

bstansberry (approved):
@yersan This LGTM.

yersan (approved):
Thanks for helping with the review @bstansberry . 
The points I reviewed were finally moved to a different issue and were already merged. I've checked it again and I have no more questions.

There is still pending paperwork on the RFE Jira that I guess, still prevents merging this.

yersan:
Thanks @ehsavoie @mnovak1 @bstansberry

PR #5869, opened 2024-02-16
WFCORE-6697: fix resource location check

michpetrov:
Issue: [WFCORE-6697](https://issues.redhat.com/browse/WFCORE-6697)

The operation is deprecated so I don't know if we want to fix this.

### DIFF
--- a/server/src/main/java/org/jboss/as/server/controller/resources/ModuleLoadingResourceDefinition.java
+++ b/server/src/main/java/org/jboss/as/server/controller/resources/ModuleLoadingResourceDefinition.java
@@ -211,8 +211,9 @@ public class ModuleLoadingResourceDefinition extends SimpleResourceDefinition {
 
         List<String> result = new LinkedList<>();
         for (ResourceLoaderInfo rl : loader.getResourceLoaders(moduleName)){
-            if (rl.getLocation() != null) {
-                URL url = new URL(rl.getLocation());
+            final String location = rl.getLocation();
+            if (location != null && !location.equals("null")) {
+                URL url = new URL(location);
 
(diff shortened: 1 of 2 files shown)


server/src/main/java/org/jboss/as/server/controller/resources/ModuleLoadingResourceDefinition.java
```diff
@@ -211,7 +211,7 @@ private static List<String> findResourcePaths(String moduleName) throws ModuleLo
 
         List<String> result = new LinkedList<>();
         for (ResourceLoaderInfo rl : loader.getResourceLoaders(moduleName)){
-            if (rl.getLocation() != null) {
+            if (!rl.getLocation().equals("null")) {
```
yersan, line 214: The [ResourceLoaderInfo](https://github.com/jboss-modules/jboss-modules/blob/be392c3854eb6390be7fa2e6b405f944219f908a/src/main/java/org/jboss/modules/management/ResourceLoaderInfo.java#L38) constructor and `getLocation()` method don't say too much about the nullity returned in case there is no location, but in any case, checking for a null value before comparing with "null" string would make it more resilient to JBoss Modules changes.

What do you think @michpetrov ?

michpetrov: The initialization was changed to avoid dealing with null values - https://github.com/jboss-modules/jboss-modules/commit/a357326bea4e08c13048f4f4cd406a0094eded69 - so I don't think it's likely, but sure I'll change it.

yersan (approved):

PR #5877, opened 2024-02-21
[WFCORE-6685] CLIEmbedServerTestCase.testStdOutEcho fails intermittently

lvydra:
https://issues.redhat.com/browse/WFCORE-6685

### DIFF
--- a/testsuite/manualmode/src/test/java/org/jboss/as/test/manualmode/management/cli/CLIEmbedServerTestCase.java
+++ b/testsuite/manualmode/src/test/java/org/jboss/as/test/manualmode/management/cli/CLIEmbedServerTestCase.java
@@ -32,6 +32,7 @@ import java.nio.charset.StandardCharsets;
 import java.nio.file.Files;
 import java.util.Collections;
 import java.util.List;
+import java.util.concurrent.TimeUnit;
 
 import org.codehaus.plexus.util.FileUtils;
 import org.jboss.as.controller.client.ModelControllerClient;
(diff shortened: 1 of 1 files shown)


testsuite/manualmode/src/test/java/org/jboss/as/test/manualmode/management/cli/CLIEmbedServerTestCase.java
```diff
@@ -725,8 +725,20 @@ private String readLogOut() {
...
+
+    private void checkLogging(String line, int timeout) throws IOException, InterruptedException {
         String logOutput = readLogOut();
+        long done = System.currentTimeMillis() + timeout;
+        while (timeout > 0 && System.currentTimeMillis() < done) {
+            if (checkLogging(logOutput, line)) {
+                break;
+            } else {
+                Thread.sleep(20);
+            }
```
yersan, line 740: Thanks for adding this, although I would suggest a longer pause, let's say 1 second or half one. Using TimeUnit makes it more readable, but that's also completely optional e.g `TimeUnit.MILLISECONDS.sleep(500)`

The else can be removed since the break will do the work, that's optional too.

None of these notes would prevent merging it as it is, so I leave it up to you to apply the changes.

lvydra: Thanks for review @yersan, I have added suggested changes.

yersan (approved):

yersan:
Thanks @lvydra

PR #5882, opened 2024-02-23
[WFCORE-6712] Ensure the JBoss Modules ModuleLoggerFinder is activate…

jamezp:
…d after the log manager is configured.

https://issues.redhat.com/browse/WFCORE-6712

Please note we could wait for https://github.com/jboss-modules/jboss-modules/pull/328 to be merged, then upgrade JBoss Modules and simply replace the reflection calls with `ModuleLoggerFinder.activate()`. However, I figured I'd send this for now just in case we want a fix sooner or one that is backwards compatible with other versions of JBoss Modules.

### DIFF
--- a/bootable-jar/runtime/src/main/java/org/wildfly/core/jar/runtime/BootableJar.java
+++ b/bootable-jar/runtime/src/main/java/org/wildfly/core/jar/runtime/BootableJar.java
@@ -13,6 +13,8 @@ import java.io.InputStream;
 import java.io.InputStreamReader;
 import java.io.UncheckedIOException;
 import java.lang.management.ManagementFactory;
+import java.lang.reflect.InvocationTargetException;
+import java.lang.reflect.Method;
 import java.nio.charset.StandardCharsets;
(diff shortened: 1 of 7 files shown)


yersan:
Ok, @jamezp , let's wait a bit for MODULES-447 decision

jamezp:
> Ok, @jamezp , let's wait a bit for MODULES-447 decision

@yersan It looks like for now, [reflection](https://github.com/jboss-modules/jboss-modules/pull/329#discussion_r1502761692) is the way to go.

bstansberry (changes requested):
@jamezp  The change looks fine but I think this should have a regression test.

jamezp:
> @jamezp The change looks fine but I think this should have a regression test.

That's a good point. Let me see if I can think of a way to do that.

jamezp:
Regression tests added.

yersan (approved):

yersan:
Thanks @jamezp @bstansberry

PR #5884, opened 2024-02-26
[WFCORE-6739] Allow ordering of ServerActivity execution by registerin…

bstansberry:
…g them in distinct, ordered, 'execution groups'

https://issues.redhat.com/browse/WFCORE-6739

This is a draft pending further discussion, particularly of how suitable this is for the transaction subsystem's implementation of https://issues.redhat.com/browse/WFLY-17742.

@jmfinelli FYI.

### DIFF
--- a/server/src/main/java/org/jboss/as/server/suspend/SuspendController.java
+++ b/server/src/main/java/org/jboss/as/server/suspend/SuspendController.java
@@ -6,9 +6,14 @@
 package org.jboss.as.server.suspend;
 
 import java.util.ArrayList;
+import java.util.Iterator;
 import java.util.List;
+import java.util.NavigableMap;
 import java.util.Timer;
 import java.util.TimerTask;
+import java.util.TreeMap;
+import java.util.function.BiConsumer;
+
 import org.jboss.as.controller.notification.NotificationHandlerRegistry;
 import org.jboss.as.server.logging.ServerLogger;
 import org.jboss.msc.service.Service;
@@ -16,6 +21,7 @@ import org.jboss.msc.service.StartContext;
 import org.jboss.msc.service.StartException;
 import org.jboss.msc.service.StopContext;
 import org.jboss.msc.value.InjectedValue;
+import org.wildfly.common.Assert;
 
 /**
  * The graceful shutdown controller. This class co-ordinates the graceful shutdown and pause/resume of a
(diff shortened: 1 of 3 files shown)


server/src/main/java/org/jboss/as/server/suspend/SuspendController.java
```diff
@@ -75,20 +81,32 @@ public synchronized void suspend(long timeoutMillis) {
         for(OperationListener listener: new ArrayList<>(operationListeners)) {
             listener.suspendStarted();
         }
-        outstandingCount = activities.size();
-        if (outstandingCount == 0) {
+        groupsCount = activities.size();
```
jmfinelli, line 84: Shouldn't this be
```suggestion
        // Work out the number of ServerActivity implementations 
        activities.forEach((ranking,executionGroup) -> groupsCount =+ executionGroup.size());
```

Or, we could also add a `groupsCount++` in `registerActivity(ServerActivity)`

jmfinelli (commented):
I'm not sure why this is a DRAFT PR :-) It looks super good to me.
I just left a comment in the code.

Thank you very much @bstansberry for taking care of this

server/src/main/java/org/jboss/as/server/suspend/SuspendController.java
bstansberry: I don't think so. This var is meant to track the number of groups, not the number of activities. When each *group* completes preSuspend work it signals preSuspendGroupCallBack. When preSuspendGroupCallBack get groupsCount signals, it proceeds with the 'suspend' phase.

I should probably rename the 'activities' field to 'activitiesByGroup'.

bstansberry:
Thanks for the review, @jmfinelli. 

It's draft for a couple reasons. :-)

The biggest is that we're doing this to meet your use case, so before having other folks waste time looking at it, I wanted to make sure you thought it would. When we talked I thought we understood each other and something like this should solve your problem, but maybe I misunderstood or you thought more and don't think it will help etc. Sounds like it's ok though, so that's good.

The other is that I should write a unit test. This works fine it that it doesn't break existing tests but that doesn't prove it actually works. :)

Looking at SuspendController it seems very unit-testable, and the basic behavior that needs testing is straightforward (activities are invoked in the expected order regardless of the order of calls to registerActivity).

server/src/main/java/org/jboss/as/server/suspend/SuspendController.java
jmfinelli: Yep, I see it now. Sorry for the noise.

jmfinelli:
Yep, it look super good to me and it reflects exactly what we talked about during the meeting :-) I think this is the last piece of the puzzle :partying_face: I also declared a dependency in the [proposal](https://github.com/wildfly/wildfly-proposals/pull/520/files#diff-e4492f568858140c27f70bb443a285510631fb5836e49b4a8684b89dcd2cec2fR104).

bstansberry:
@jmfinelli I added the test and moved this out of draft state.

bstansberry:
Thanks @jmfinelli

PR #5885, opened 2024-02-28
[WFCORE-6717]: Incorrectly set options on the command line should fail

ehsavoie:
to start a server.
 * If no value is defined for yaml o git command line options we should fail.

Jira: https://issues.redhat.com/browse/WFCORE-6717


yersan:
Thanks @ehsavoie

PR #5887, opened 2024-02-28
[WFCORE-6167] Workaround for known issue (enabled on JDK22+)

ropalka:
https://issues.redhat.com/browse/WFCORE-6167


jamezp (approved):
I almost wonder if this is more of a fix than a workaround. Test manually the modes don't break. However, in the test itself if I run without this fix I see the output on `System.out` no matter what is done.

ropalka:
The fundamental question what was asked in WFLY-6167 JIRA is whether when console is available we want logging to be so verbose or not on System.out. To compare how much verbose it is just compare WF Core 'mvn clean install' output on JDK21 vs JDK22 @jamezp .

yersan:
@ropalka Could you add a more descriptive commit message?

ropalka:
Fixed @yersan - it was an oversight

yersan:
Thanks @ropalka @jamezp

PR #5888, opened 2024-02-29
[WFCORE-6719] Upgrade JBoss Invocation to 2.0.1.Final

ropalka:
https://issues.redhat.com/browse/WFCORE-6719


yersan (approved):
Going to conflict with classfilewriter upgrade, but we can manually edit it before merging

PR #5894, opened 2024-03-06
[WFCORE-6723] Add header 'X-Content-Type-Options' to http management handlers

soul2zimate:
Issue: https://issues.redhat.com/browse/WFCORE-6723

[WFCORE-6723] Add header 'X-Content-Type-Options' to http management handlers

### DIFF
--- a/domain-http/interface/src/main/java/org/jboss/as/domain/http/server/DomainUtil.java
+++ b/domain-http/interface/src/main/java/org/jboss/as/domain/http/server/DomainUtil.java
@@ -202,7 +202,7 @@ public class DomainUtil {
                 .setCachable(not(suffixes(NOCACHE_JS, APP_HTML, INDEX_HTML)));
 
         // avoid clickjacking attacks: console must not be included in (i)frames
-        SetHeaderHandler frameHandler = new SetHeaderHandler(handler, "X-Frame-Options", "SAMEORIGIN");
+        SetHeaderHandler frameHandler = new SetHeaderHandler (new SetHeaderHandler(handler, "X-Frame-Options", "SAMEORIGIN"), "X-Content-Type-Options", "nosniff");
         // we also need to setup the default resource redirect
         PredicateHandler predicateHandler = new PredicateHandler(path("/"), new RedirectHandler(ExchangeAttributes.constant(context + DEFAULT_RESOURCE)), frameHandler);
         return new ResourceHandlerDefinition(context, DEFAULT_RESOURCE, predicateHandler);
--- a/domain-http/interface/src/main/java/org/jboss/as/domain/http/server/ManagementHttpServer.java
+++ b/domain-http/interface/src/main/java/org/jboss/as/domain/http/server/ManagementHttpServer.java
@@ -260,7 +260,7 @@ public class ManagementHttpServer {
     }
 
     private static HttpHandler addDmrRedinessHandler(PathHandler pathHandler, HttpHandler domainApiHandler, Function<HttpServerExchange, Boolean> readinessFunction) {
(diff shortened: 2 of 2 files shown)


yersan (approved):
Looks good to me, although I am not sure how we can ensure HAL is setting the expected content type for CSS and JS files. I assume HAL is doing it correctly.

@hpehl Could you provide any feedback about it and is so, approve this PR on your side?

hpehl (approved):
Looks good to me! 

As a pure client, HAL itself is not involved in setting the `X-Content-Type-Options` response header.

bstansberry:
@yersan @soul2zimate Harald is correct that setting the correct MIME type is not a HAL concern. I think we should add some tests that we are doing it correctly though.

OR

We look at the cases where the management interface is returning static css and js content, look at the Undertow classes used, and we find or add tests to Undertow to confirm the behavior of those classes.

I think we are talking about 

1) ConsoleMode:

```        
    static ResourceHandlerDefinition createConsoleHandler(String skin) throws ModuleLoadException {
            final ClassPathResourceManager resource = new ClassPathResourceManager(findConsoleClassLoader(Module.getCallerModuleLoader(), skin), "");
            return DomainUtil.createStaticContentHandler(resource, CONTEXT);

        }
```

leading to DomainUtil:

```    
    static ResourceHandlerDefinition createStaticContentHandler(ResourceManager resource, String context) {
        final io.undertow.server.handlers.resource.ResourceHandler handler = new io.undertow.server.handlers.resource.ResourceHandler(resource)
                .setCacheTime(60 * 60 * 24 * 31)
                .setAllowed(not(path("META-INF")))
                .setResourceManager(resource)
                .setDirectoryListingEnabled(false)
                .setCachable(not(suffixes(NOCACHE_JS, APP_HTML, INDEX_HTML)));

        // avoid clickjacking attacks: console must not be included in (i)frames
        SetHeaderHandler frameHandler = new SetHeaderHandler(handler, "X-Frame-Options", "SAMEORIGIN");
        // we also need to setup the default resource redirect
        PredicateHandler predicateHandler = new PredicateHandler(path("/"), new RedirectHandler(ExchangeAttributes.constant(context + DEFAULT_RESOURCE)), frameHandler);
        return new ResourceHandlerDefinition(context, DEFAULT_RESOURCE, predicateHandler);
    }

```

2) ErrorContextHandler

```
    public static HttpHandler createErrorContext(final String slot) throws ModuleLoadException {
        final ClassPathResourceManager cpresource = new ClassPathResourceManager(getClassLoader(Module.getCallerModuleLoader(), ERROR_MODULE, slot), "");
        final io.undertow.server.handlers.resource.ResourceHandler handler = new io.undertow.server.handlers.resource.ResourceHandler(cpresource)
                .setAllowed(not(path("META-INF")))
                .setDirectoryListingEnabled(false)
                .setCachable(Predicates.<HttpServerExchange>falsePredicate());

        //we also need to setup the default resource redirect
        return new PredicateHandler(path("/"), new RedirectHandler(ExchangeAttributes.constant(ERROR_CONTEXT + DEFAULT_RESOURCE)), handler);
    }
```

bstansberry:
Thanks @soul2zimate, @yersan and @hpehl.

I suggested adding a test that we set the mime type correctly, but really that kind of test belongs in Undertow. I don't know if there is one, but looking at the code it looks to be handled correctly, and is done in a way that they'd likely have had many bug reports if it was wrong.

soul2zimate:
There is similar test in WFLY https://github.com/wildfly/wildfly/blob/main/testsuite/integration/basic/src/test/java/org/jboss/as/test/integration/management/console/XFrameOptionsHeaderTestCase.java I am about to add a new one via https://issues.redhat.com/browse/WFLY-19101 for this change as well once wildfly core is upgraded there

PR #5896, opened 2024-03-08
Fix for WFCORE-6729, Upgrade Galleon to 6.0.0.Beta3, Galleon plugins to 7.0.0.Beta4, and provisioning tooling

jfdenise:
Issue: https://issues.redhat.com/browse/WFCORE-6729


yersan:
@jfdenise Although the core-feature-pack is not directly used in WildFly, shouldn't it already define the default minimum stability level now that Galleon is upgraded?

yersan (approved):

yersan:
Thanks @jfdenise

PR #5897, opened 2024-03-08
WFCORE-6730 A registered operation should never contain parameters or reply parameters not enabled by stability-level of process

pferraro:
https://issues.redhat.com/browse/WFCORE-6730

### DIFF
--- a/controller/src/main/java/org/jboss/as/controller/SimpleOperationDefinitionBuilder.java
+++ b/controller/src/main/java/org/jboss/as/controller/SimpleOperationDefinitionBuilder.java
@@ -8,6 +8,7 @@ package org.jboss.as.controller;
 import java.util.Arrays;
 import java.util.Collections;
 import java.util.EnumSet;
+import java.util.Set;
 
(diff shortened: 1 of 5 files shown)


controller/src/main/java/org/jboss/as/controller/SimpleOperationDefinitionBuilder.java
```diff
@@ -23,17 +24,57 @@
...
+                .setReplyType(basis.getReplyType())
+                .setReplyValueType(basis.getReplyValueType())
+                .setStability(basis.getStability())
+                .withFlags(basis.getFlags())
+                ;
+        DeprecationData deprecation = basis.getDeprecationData();
+        if (deprecation != null) {
+            builder.setDeprecated(deprecation.getSince(), deprecation.isNotificationUseful());
+        }
+        return builder;
```
yersan, line 70: It looks like this would also require the initialization of `replyAllowNull` from the `SimpleOperationDefinition` passed as argument.

yersan (approved):

yersan:
Thanks @pferraro

PR #5899, opened 2024-03-11
[WFCORE-6741] [WFCORE-6652] Improve user experience when using extension or namespace of unsupported stability.

pferraro:
https://issues.redhat.com/browse/WFCORE-6652
https://issues.redhat.com/browse/WFCORE-6741

Rather than skip parser registration of unsupported extensions (or unsupported schemas of a supported extension) entirely, register its namespaces using a placeholder parser throwing an informative message.

### DIFF
--- a/controller/src/main/java/org/jboss/as/controller/extension/ExtensionRegistry.java
+++ b/controller/src/main/java/org/jboss/as/controller/extension/ExtensionRegistry.java
@@ -776,7 +776,7 @@ public final class ExtensionRegistry implements FeatureRegistry {
 
         @Override
         public Stability getStability() {
-            return this.profileRegistration.getStability();
+            return ExtensionRegistry.this.getStability();
         }
 
         @Override
--- a/controller/src/main/java/org/jboss/as/controller/extension/ExtensionRegistry.java
+++ b/controller/src/main/java/org/jboss/as/controller/extension/ExtensionRegistry.java
@@ -348,9 +348,9 @@ public final class ExtensionRegistry implements FeatureRegistry {
      *                   be any actual parsing (e.g. in a slave Host Controller or in a server in a managed domain)
      */
     public void initializeParsers(final Extension extension, final String moduleName, final XMLMapper xmlMapper) {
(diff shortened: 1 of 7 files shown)


controller/src/main/java/org/jboss/as/controller/extension/ExtensionRegistry.java
```diff
@@ -348,9 +348,9 @@ public ExtensionParsingContext getExtensionParsingContext(final String moduleNam
      *                   be any actual parsing (e.g. in a slave Host Controller or in a server in a managed domain)
      */
     public void initializeParsers(final Extension extension, final String moduleName, final XMLMapper xmlMapper) {
-        ExtensionParsingContextImpl parsingContext = new ExtensionParsingContextImpl(moduleName, xmlMapper);
-        extension.initializeParsers(parsingContext);
-        parsingContext.attemptCurrentParserInitialization();
+        try (AbstractExtensionParsingContext context = this.enables(extension) ? new ExtensionParsingContextImpl(moduleName, xmlMapper) : new UnstableExtensionParsingContext(moduleName, xmlMapper)) {
+            extension.initializeParsers(context);
+        }
```
yersan, line 353: Autocloseable interface is used to mark objects that could hold resources, here it is used to invoke the ExtensionParsingContext#attemptCurrentParserInitialization(), since the abstract class does nothing on its close() method.

Is that usage correct? It looks inappropriate. I don't see where ExtensionParsingContext#attemptCurrentParserInitialization() was trying to release resources that can justify the movement to an Autocloseable.close() implementation.

pferraro: * An extension parsing context *could* hold resources.
* The JDK itself implements AutoCloseable for objects that do not "hold resources" in the traditional sense, but that have a defined lifecycle, e.g. DoubleStream, IntStream, LongStream, Stream, etc.
* The usage here is an internal implementation detail of the private class and is not publicly exposed.

kabir (approved):

PR #5903, opened 2024-03-15
WFCORE-4868: implement aliases for standalone configuration files

michpetrov:
Issue: [WFCORE-4868](https://issues.redhat.com/browse/WFCORE-4868)

Documentation PR: wildfly/wildfly#17751

### DIFF
--- a/server/src/main/java/org/jboss/as/server/ServerEnvironment.java
+++ b/server/src/main/java/org/jboss/as/server/ServerEnvironment.java
@@ -496,7 +496,10 @@ public class ServerEnvironment extends ProcessEnvironment implements Serializabl
             } else {
                 repository = null;
             }
-            serverConfigurationFile = standalone ? new ConfigurationFile(serverConfigurationDir, defaultServerConfig, serverConfig, configInteractionPolicy, repository != null, serverTempDir, configurationExtension) : null;
+
+            this.stability = getEnumProperty(props, ProcessEnvironment.STABILITY, productConfig.getDefaultStability());
+            final String translatedConfig = translateFileAlias(serverConfig, stability);
+            serverConfigurationFile = standalone ? new ConfigurationFile(serverConfigurationDir, defaultServerConfig, translatedConfig, configInteractionPolicy, repository != null, serverTempDir, configurationExtension) : null;
             // Adds a system property to indicate whether or not the server configuration should be persisted
             @SuppressWarnings("deprecation")
             final String propertyKey = JBOSS_PERSIST_SERVER_CONFIG;
@@ -522,7 +525,6 @@ public class ServerEnvironment extends ProcessEnvironment implements Serializabl
                 this.domainConfigurationDir = null;
             }
 
-            this.stability = getEnumProperty(props, ProcessEnvironment.STABILITY, productConfig.getDefaultStability());
             if (!productConfig.getStabilitySet().contains(this.stability)) {
                 throw ServerLogger.ROOT_LOGGER.unsupportedStability(this.stability, productConfig.getProductName());
             }
@@ -1241,5 +1243,26 @@ public class ServerEnvironment extends ProcessEnvironment implements Serializabl
 
     ManagedAuditLogger createAuditLogger() {
         return new ManagedAuditLoggerImpl(getProductConfig().resolveVersion(), true);
+    }
+
+    public static String translateFileAlias(String alias, Stability stability) {
+        if (!stability.enables(Stability.COMMUNITY) || alias == null) {
+            return alias;
+        }
+        switch (alias) {
+            case "full":
+            case "ha":
+            case "full-ha":
+            case "load-balancer":
+            case "microprofile":
+            case "microprofile-ha":
+                break;
+            case "fha":   alias = "full-ha"; break;
+            case "lb":    alias = "load-balancer"; break;
+            case "mp":    alias = "microprofile"; break;
+            case "mpha":  alias = "microprofile-ha"; break;
+            default:      return alias;
+        }
+        return "standalone-" + alias + ".xml";
     }
 }
--- a/server/src/test/java/org/jboss/as/server/ServerEnvironmentTestCase.java
+++ b/server/src/test/java/org/jboss/as/server/ServerEnvironmentTestCase.java
@@ -91,4 +91,42 @@ public class ServerEnvironmentTestCase {
         assertThat(uuids.get(0), is(not(uuid)));
         Files.delete(uuidPath);
     }
+
+    @Test
+    public void testAliasFunctionality() throws IOException {
+        Properties props = new Properties();
+        Path standaloneDir = homeDir.resolve("standalone");
+        Files.createDirectories(standaloneDir.resolve("configuration"));
+        Files.createFile(standaloneDir.resolve("configuration").resolve("standalone.xml"));
+        Files.createFile(standaloneDir.resolve("configuration").resolve("standalone-load-balancer.xml"));
+        Files.createFile(standaloneDir.resolve("configuration").resolve("custom.xml"));
+        props.put(HOME_DIR, homeDir.toAbsolutePath().toString());
+
+        // default stability = COMMUNITY
+        ProductConfig productConfig = ProductConfig.fromFilesystemSlot(null, "", props);
+
+        ServerEnvironment serverEnvironment = createServerEnvironment(props, null, productConfig);
+        assertThat(serverEnvironment.getServerConfigurationFile().getBootFile().getName(), is("standalone.xml"));
+
+        serverEnvironment = createServerEnvironment(props, "lb", productConfig);
+        assertThat(serverEnvironment.getServerConfigurationFile().getBootFile().getName(), is("standalone-load-balancer.xml"));
+
+        serverEnvironment = createServerEnvironment(props, "custom.xml", productConfig);
+        assertThat(serverEnvironment.getServerConfigurationFile().getBootFile().getName(), is("custom.xml"));
+    }
+
+    @Test(expected = IllegalStateException.class)
+    public void testAliasNotWorkingInDefaultStability() {
+        Properties props = new Properties();
+        props.put(HOME_DIR, homeDir.toAbsolutePath().toString());
+
+        // default stability = DEFAULT
+        ProductConfig productConfig = new ProductConfig(null, null, null);
+        createServerEnvironment(props, "lb", productConfig);
+    }
+
+    private ServerEnvironment createServerEnvironment(Properties props, String serverConfig, ProductConfig productConfig) {
+        return new ServerEnvironment(null, props, System.getenv(), serverConfig,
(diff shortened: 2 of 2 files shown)


server/src/main/java/org/jboss/as/server/ServerEnvironment.java
```diff
@@ -496,7 +496,16 @@ public ServerEnvironment(final String hostControllerName, final Properties props
...
+
+            this.stability = getEnumProperty(props, ProcessEnvironment.STABILITY, productConfig.getDefaultStability());
+            if (!productConfig.getStabilitySet().contains(this.stability)) {
+                throw ServerLogger.ROOT_LOGGER.unsupportedStability(this.stability, productConfig.getProductName());
+            }
+
+            serverConfigurationFile = standalone
+                    ? new ConfigurationFile(serverConfigurationDir, defaultServerConfig, serverConfig,
+                        configInteractionPolicy, repository != null, serverTempDir, configurationExtension, stability)
+                    : null;
```
yersan, line 508: Could you translate the alias here outside of the ConfigurationFile?

controller/src/main/java/org/jboss/as/controller/persistence/ConfigurationFile.java
```diff
@@ -339,8 +361,13 @@ private File determineMainFile(final String rawName, final String name) {
         if (mainName == null) {
             // Try the basic case, where name is the name
             final File directoryFile = new File(configurationDir, name);
+            // Translate alias if possible
+            String aliasedName = translateFileAlias(name);
```
yersan, line 365: After looking at the changes implemented, I think we should extract the `translateFileAlias` method outside of the ConfigurationFile to avoid any side effects, for example, at this moment the `bootFileName` field class will get the alias name, instead of the translated file name.

If we extract the `translateFileAlias` method and move it to ServerEnvironment.java, we can leave any domain mode-related classes untouched and have this only applied to standalone files. That would also avoid the need to pass Stability to the `ConfigurationFile` constructors. The Stabilitily.enables(...) invocation could be also invoked in this `translateFileAlias`, which can be a static method so it can be easily tested in a Unit Test.

controller/src/main/java/org/jboss/as/controller/persistence/ConfigurationFile.java
```diff
@@ -373,6 +400,18 @@ private File determineMainFile(final String rawName, final String name) {
         throw ControllerLogger.ROOT_LOGGER.mainFileNotFound(name != null ? name : rawName, configurationDir);
     }
 
+    private String translateFileAlias(String alias) {
+        final int pos = rawFileName.lastIndexOf('.');
+        final String base = pos != -1 ? rawFileName.substring(0, pos) : rawFileName;
```
yersan, line 405: If you make the final name dependent on the `rawFileName` you would need to clarify it on the proposal analysis doc.
As far as I understand, the `rawFileName` is the name default name for configuration files. This default name can be modified by `jboss.server.default.config`. If the base of the translate alias depends on that, we need to describe this dependency on the analysis doc file to don't add any unadvertised behaviour.

controller/src/main/java/org/jboss/as/controller/persistence/ConfigurationFile.java
michpetrov: Yes, the `bootFileName` would be the alias, that's the way it behaves with values like "last" or "boot" which are kind of aliases already, except they don't look in the base folder.

And these work in domain too, so I don't see why not to make this available in domain as well.

controller/src/main/java/org/jboss/as/controller/persistence/ConfigurationFile.java
michpetrov: I'll clarify in the proposal then, making it only work with files specifically named "standalone-" seems needlessly limiting.

controller/src/main/java/org/jboss/as/controller/persistence/ConfigurationFile.java
yersan: > Yes, the bootFileName would be the alias, that's the way it behaves with values like "last" or "boot" which are kind of aliases already, except they don't look in the base folder.

Thanks for discussing it. Right, but concepts like last / boot / initial and so on are already known concepts, so the code should be already prepared to handle any corner cases using the value of `bootFileName` with these values. However, we are now introducing something new, that will be deactivated on the default stability level. I think minimizing the changes could be a good approach, even more, when we have had undiscovered product issues after modifying this ConfigurationFile. I can see logic related to this `bootFileName` class field and don't control it at all, so being conservative is my first approach to this. I could be wrong.


> And these work in domain too, so I don't see why not to make this available in domain as well.

This is a feature that is not available in domain mode. Also, I'm not sure how useful would be for that, maybe having an alias for host-secondary.xml.

The ConfigurationFile class is used on both, standalone and domain mode. I agree it is the logical class to place details related to aliases for server configuration files, but again, since it has shown in the past to be a bit delicate, this is not a domain mode RFE and it would be in community stability level, I'm suggesting to get it out of this class to minimize the changes.

@bstansberry / @darranl An external opinion about what could be the best approach would be very useful for us. Maybe I am being too conservative with this. I am trying to minimize the changes extracting any related logic outside of the ConfigurationFile, however, this class is the natural place for this, so I could be wrong. What's your feeling about this?

michpetrov: @yersan Bear in mind that if I extract this to `ServerEnvironment` I have to make sure not to interfere with the special cases, so I have to duplicate all the checks from `ConfigurationFile`:
 https://github.com/wildfly/wildfly-core/blob/c9d07f4835f0695e9e333a6c32e3019d3193f4a1/controller/src/main/java/org/jboss/as/controller/persistence/ConfigurationFile.java#L346-#L360
to make sure this an alias to be translated. Or I would have to change method to work only with the files that we distribute.

yersan: @michpetrov I think you won't need to extract any logic. The ConfigurationFile uses the argument `name` as the user's provided name of the configuration file to use. My understanding is we need to translate the alias to the expected name and pass this translated name in this argument. The ConfigurationFile will receive the translated name, so from its point of view, there is no alias in place.

yersan (changes requested):
This RFE also needs the community documentation PR linked and ready

controller/src/main/java/org/jboss/as/controller/persistence/ConfigurationFile.java
michpetrov: @yersan in case of e.g. "boot" I have to pass "boot" as the `name` but my implementation would translate it to "standalone-boot.xml".

yersan: @michpetrov I would prefer to keep the same meaning of these suffixes, they are suffixes with a special meaning (the same for initla, last, v* and so on), so they should not be treated as possible aliases. You would need to fix the log to take them into account.  If the user uses `boot`, this value should be translated to `boot` and passed as an argument to the ConfigurationFile using the `name` argument.

michpetrov: That's what I meant, I have to copy over the logic that checks for these values so that I can skip them (only for `ConfigurationFile` to run the same checks again). And I'll have to expose a few of the constants from `ConfigurationFile`, or change the checks into static methods.

yersan: @michpetrov Sorry, but I am not following you with your approach.

The logic should take into account the user-provided configuration file name, and if the name is an alias that represents a configuration name we distribute (standalone-full-ha.xml, standalone-microprofile.xml and so on), return that one, otherwise return whatever the has provided.

The ConfiguarionFile will be in charge of managing all the boot, initial, last details as it is doing now. Your logic only would translate aliases in specific defined configuration files. I don't think you need to extract any logic / constants to do that, a simple switch should be able to do all the work.

michpetrov: @yersan the basic logic is for the method to take the a string like "custom" and return "standalone-custom.xml", except for a few special strings that translate to the files we provide. But this means a string like "boot" would get turned into "standalone-boot.xml" and I have to make sure that doesn't happen.

yersan: @michpetrov Ok, I see now. I've been all the time thinking only on the simple approach where we were going to add aliases only for the files we distribute (standalone-full-ha.xml, standalone-microprofile.xml and so on).

You already clarified it in the proposal: https://github.com/wildfly/wildfly-proposals/pull/390#issuecomment-2009841010, but I did not pay the correct attention to that.

If the proposal gets approved as it is, I am fine by adding the changes inside of ConfigurationFile, because you would need to process the boot, last, initial, and so on before doing any possible alias translation. I was just expecting this only for the files we distribute.

server/src/test/java/org/jboss/as/server/ServerEnvironmentTestCase.java
```diff
@@ -91,4 +92,12 @@ public void testUUIDLifeCycle() throws IOException {
...
         Files.delete(uuidPath);
     }
+
+    @Test
+    public void testAliasFunctionality() {
+        assertThat(ServerEnvironment.translateFileAlias("lb", Stability.DEFAULT), is("lb"));
+        assertThat(ServerEnvironment.translateFileAlias(null, Stability.DEFAULT), is((String) null));
+        assertThat(ServerEnvironment.translateFileAlias("lb", Stability.COMMUNITY), is("standalone-load-balancer.xml"));
+        assertThat(ServerEnvironment.translateFileAlias("custom.xml", Stability.COMMUNITY), is("custom.xml"));
+    }
```
yersan, line 102: Looks good, but I wonder if we should test the functionality using a ServerEnvironent instance instead of the static method.

ServerEnvironent has a getServerConfigurationFile() method that can be used to assert the final file that will be in use.

yersan: That would increase the reliability of the change, and the testUUIDLifeCycle() test has all the pieces required to initialize a ServerEnvironment, so looks very simple to add it.

yersan (approved):
LGTM, we still need the documentation PR linked and full approval of the proposal by all the parties, but the changes here look good to me, thanks @michpetrov !

yersan:
@darranl FYI proposal has been approved by all the parties, so I would say this is ready to go in, just need the documentation PR

michpetrov:
@yersan the PR is linked in the top comment

darranl:
I have just kicked off the CI jib that had an error again, looks like I can proceed with the merge once this completes.

darranl:
Thank you @michpetrov

PR #5905, opened 2024-03-18
[WFCORE-6745] Upgrade log4j2 from 2.22.1 to 2.23.1

yersan:
Jira issue: https://issues.redhat.com/browse/WFCORE-6745


jamezp:
I'm placing a hold on this for now because there were some breaking changes to the `StatusLogger` with regards to the `log4j2-jboss-logmanager`. This might not be critical as users likely don't use the status logger, but it's something I'd like to look at.

PR #5907, opened 2024-03-19
Fix for WFCORE-6747, Upgrade to Galleon 6.0.0.Beta4 and WFGP 7.0.0.Beta5

jfdenise:
Issue: https://issues.redhat.com/browse/WFCORE-6747
Align with Stability handling in provisioning as specified in https://github.com/wildfly/wildfly-proposals/pull/555

### DIFF
--- a/core-feature-pack/galleon-feature-pack/pom.xml
+++ b/core-feature-pack/galleon-feature-pack/pom.xml
@@ -233,6 +233,8 @@
                             <fork-embedded>${galleon.fork.embedded}</fork-embedded>
                             <generate-channel-manifest>true</generate-channel-manifest>
                             <minimum-stability>experimental</minimum-stability>
+                            <config-stability>community</config-stability>
+                            <package-stability>experimental</package-stability>
                         </configuration>
                     </execution>
                 </executions>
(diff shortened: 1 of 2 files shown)


core-feature-pack/galleon-feature-pack/pom.xml
```diff
@@ -233,6 +233,8 @@
                             <fork-embedded>${galleon.fork.embedded}</fork-embedded>
                             <generate-channel-manifest>true</generate-channel-manifest>
                             <minimum-stability>experimental</minimum-stability>
+                            <config-stability>community</config-stability>
+                            <package-stability>experimental</package-stability>
```
bstansberry, line 237: @jfdenise In a future iteration these 3 flags seem like things that should be part of wildfly-feature-back-build.xml.

The mojo options can remain as overrides (and for compatibility), but the other file seems like the right place for configuring the FP content, as opposed to the tool.

bstansberry (approved):

bstansberry:
Thanks @jfdenise

core-feature-pack/galleon-feature-pack/pom.xml
jfdenise: @bstansberry , thank-you I logged https://issues.redhat.com/browse/WFGP-275 to track it.

PR #5908, opened 2024-03-19
[WFCORE-6731] Expose supported stability levels via the management API

yersan:
Jira issue: https://issues.redhat.com/browse/WFCORE-6731

### DIFF
--- a/host-controller/src/main/java/org/jboss/as/host/controller/HostControllerEnvironment.java
+++ b/host-controller/src/main/java/org/jboss/as/host/controller/HostControllerEnvironment.java
@@ -16,6 +16,7 @@ import java.util.Collections;
 import java.util.EnumSet;
 import java.util.Locale;
 import java.util.Map;
+import java.util.Set;
 import java.util.UUID;
 
 import org.jboss.as.controller.OperationContext;
@@ -234,6 +235,7 @@ public class HostControllerEnvironment extends ProcessEnvironment {
 
     private final RunningMode initialRunningMode;
     private final Stability stability;
+    private final Set<Stability> stabilities;
     private final ProductConfig productConfig;
     private final String qualifiedHostName;
     private final String hostName;
@@ -485,6 +487,7 @@ public class HostControllerEnvironment extends ProcessEnvironment {
         this.processType = processType;
 
         this.stability = getEnumProperty(hostSystemProperties, STABILITY, this.productConfig.getDefaultStability());
+        this.stabilities = productConfig.getStabilitySet();
         if (!this.productConfig.getStabilitySet().contains(this.stability)) {
             throw HostControllerLogger.ROOT_LOGGER.unsupportedStability(this.stability, this.productConfig.getProductName());
         }
@@ -806,6 +809,11 @@ public class HostControllerEnvironment extends ProcessEnvironment {
     @Override
     public Stability getStability() {
         return this.stability;
+    }
+
+    @Override
+    public Set<Stability> getStabilities() {
+        return this.stabilities;
     }
 
     @Override
--- a/server/src/main/java/org/jboss/as/server/ServerEnvironment.java
+++ b/server/src/main/java/org/jboss/as/server/ServerEnvironment.java
@@ -304,6 +304,7 @@ public class ServerEnvironment extends ProcessEnvironment implements Serializabl
     private final boolean startGracefully;
     private final GitRepository repository;
     private final Stability stability;
+    private final Set<Stability> stabilities;
 
     public ServerEnvironment(final String hostControllerName, final Properties props, final Map<String, String> env, final String serverConfig,
             final ConfigurationFile.InteractionPolicy configInteractionPolicy, final LaunchType launchType,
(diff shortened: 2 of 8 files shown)


host-controller/src/main/java/org/jboss/as/host/controller/HostControllerEnvironment.java
```diff
@@ -485,6 +487,7 @@ public HostControllerEnvironment(Map<String, String> hostSystemProperties, boole
         this.processType = processType;
 
         this.stability = getEnumProperty(hostSystemProperties, STABILITY, this.productConfig.getDefaultStability());
+        this.stabilities = Collections.unmodifiableSet(productConfig.getStabilitySet());
```
pferraro, line 490: Making this set unmodifiable is the right thing to do, but better to change that at the source: https://github.com/wildfly/wildfly-core/blob/main/version/src/main/java/org/jboss/as/version/ProductConfig.java#L97
That way every process env does not need to do the same wrapping.

host-controller/src/main/java/org/jboss/as/host/controller/HostControllerEnvironment.java
```diff
@@ -808,6 +811,15 @@ public Stability getStability() {
         return this.stability;
     }
 
+    /**
+     * Returns an unmodifiable set of all the permissible stability levels.
+     *
+     * @return a set of stability levels
+     */
+    public Set<Stability> getStabilities() {
+        return this.stabilities;
```
pferraro, line 820: ... that way this method (and t can just delegate to ProductConfig.getStabilities().

I would also consider defining this method in ProcessEnvironment to enforce consistency across subclasses.

pferraro (changes requested):
Very minor suggestions only.

darranl:
@hpehl Could you please also review, this exposed information will be useful to management clients.

host-controller/src/main/java/org/jboss/as/host/controller/HostControllerEnvironment.java
yersan: @pferraro Done

host-controller/src/main/java/org/jboss/as/host/controller/HostControllerEnvironment.java
yersan: done

hpehl (approved):
In general, LGTM. Thanks @yersan 

AIUI, this PR adds two properties to the management model: 

1. current stability level
2. possible stability levels

Some questions: 

- Can the user change the stability level at runtime (e.g., using an operation), or is this only possible at startup time when using a command-line option/system property?
- How can I, as a client, detect whether a specific resource, operation, or attribute is only supported for a specific stability level?

yersan:
@hpehl thanks for reviewing it.

> Can the user change the stability level at runtime (e.g., using an operation), or is this only possible at startup time when using a command-line option/system property?

There is an ongoing Feature that would allow restarting the server at a different stability level: https://github.com/wildfly/wildfly-core/pull/5895


> How can I, as a client, detect whether a specific resource, operation, or attribute is only supported for a specific stability level?


When you describe a resource/operation, now the resources provide a field that specifies the higher stability lever where this resource is available. This attribute name is "stability".

If the resource / attribute / operation is not suitable to run in the current server stability level, then the resource will not be available.

For example, if the operation `A` is configured to work on "preview" stability level, it won't be available if you run the server in "community" stability level. "community" has a higher stability level than "preview".

The operation `A` should report "preview" as its higher stability level where it can work, and you have to run the server in preview to be able to describe it.

server/src/main/java/org/jboss/as/server/ServerEnvironment.java
```diff
@@ -304,6 +304,7 @@ public ProcessType getProcessType() {
     private final boolean startGracefully;
     private final GitRepository repository;
     private final Stability stability;
+    private final Set<Stability> stabilities;
```
bstansberry, line 307: This field shouldn't be needed as this.productConfig can be called to get the stabilities set.

host-controller/src/main/java/org/jboss/as/host/controller/HostControllerEnvironment.java
```diff
@@ -234,6 +235,7 @@ public class HostControllerEnvironment extends ProcessEnvironment {
 
     private final RunningMode initialRunningMode;
     private final Stability stability;
+    private final Set<Stability> stabilities;
```
bstansberry, line 238: This field shouldn't be needed as this.productConfig can be called to get the stabilities set.

bstansberry (changes requested):
@yersan Just a couple minor points.

host-controller/src/main/java/org/jboss/as/host/controller/HostControllerEnvironment.java
bstansberry: > I would also consider defining this method in ProcessEnvironment to enforce consistency across subclasses.

@pferraro @yersan This is a good idea but I think it's out of scope for this JIRA. Just because when I look at ProcessEnvironment it's an abstract class (not an interface) shares a lot of functionality with both of its subclasses, but stores and exposes no state itself and thus forces both subclasses to do so.  If we're going to change that for this we should revisit all of it, which is out of scope.

I have no idea why it's done this way. My only guess is ProcessEnvironment was originally stateless and over time the two subclasses evolved independently and the kind of refactoring I'm discussing here was never in scope. :)

I filed https://issues.redhat.com/browse/WFCORE-6760

bstansberry:
Superseded by #5917 which resolves the conflict and deals with my minor requests.

bstansberry:
Thanks @yersan @pferraro and @hpehl

PR #5911, opened 2024-03-20
[WFCORE-6749] Fix incorrect output from standalone.sh -v

soul2zimate:
Issue: https://issues.redhat.com/browse/WFCORE-6749 Fix incorrect output from standalone.sh -v

This handles the argument `version` late and recreate `productConfig` using updated properties

### DIFF
--- a/server/src/main/java/org/jboss/as/server/Main.java
+++ b/server/src/main/java/org/jboss/as/server/Main.java
@@ -212,6 +212,10 @@ public final class Main {
                         value = arg.substring(idx + 1);
                     }
                     systemProperties.setProperty(name, value);
+                    if (ServerEnvironment.HOME_DIR.equals(name)) {
+                        // Re-create using updated properties
+                        productConfig = ProductConfig.fromFilesystemSlot(Module.getBootModuleLoader(), value, systemProperties);
+                    }
                 } else if (arg.startsWith(CommandLineConstants.PUBLIC_BIND_ADDRESS)) {
 
                     int idx = arg.indexOf('=');
@@ -448,6 +452,10 @@ public final class Main {
          try {
              url = makeURL(urlSpec);
              systemProperties.load(url.openConnection().getInputStream());
+             if (systemProperties.getProperty(ServerEnvironment.HOME_DIR) != null) {
+                 // Re-create using updated properties
+                 productConfig = ProductConfig.fromFilesystemSlot(Module.getBootModuleLoader(), systemProperties.getProperty(ServerEnvironment.HOME_DIR), systemProperties);
+             }
              return true;
          } catch (MalformedURLException e) {
              STDERR.println(ServerLogger.ROOT_LOGGER.malformedCommandLineURL(urlSpec, arg));


server/src/main/java/org/jboss/as/server/Main.java
```diff
@@ -392,6 +388,12 @@ public static ServerEnvironmentWrapper determineEnvironment(String[] args, Prope
                     } else {
                         supplementalConfiguration = arg.substring(idx + 1);
                     }
+                } else if (CommandLineConstants.VERSION.equals(arg) || CommandLineConstants.SHORT_VERSION.equals(arg)
+                        || CommandLineConstants.OLD_VERSION.equals(arg) || CommandLineConstants.OLD_SHORT_VERSION.equals(arg)) {
+                    // Re-create using updated properties
+                    productConfig = ProductConfig.fromFilesystemSlot(Module.getBootModuleLoader(), WildFlySecurityManager.getPropertyPrivileged(ServerEnvironment.HOME_DIR, null), systemProperties);
+                    STDOUT.println(productConfig.getPrettyVersionString());
+                    return new ServerEnvironmentWrapper(ServerEnvironmentWrapper.ServerEnvironmentStatus.NORMAL);
```
bstansberry, line 396: @soul2zimate @pferraro This tells me the handling of productConfig in this loop is wrong in general. This fixes it for -v but not for anything else.

The -v handling can remain at the top. It should so the change that fixes this doesn't imply it's needed to move it.

I think we need to add logic to all the else blocks that process system properties (-p and -D blocks) that check if HOME_DIR is set and if so recreate productConfig.

Probably leave it as null until that happens (or if HOME_DIR is known where we assign productConfig now), and the CommandLineUsageImpl stuff deals with a null ProductConfig by not adding the --stability arg to the output (i.e. default to it not being available until we know whether it is by loading the ProductConfig

bstansberry (changes requested):

server/src/main/java/org/jboss/as/server/Main.java
soul2zimate: > I think we need to add logic to all the else blocks that process system properties (-p and -D blocks) that check if HOME_DIR is set and if so recreate productConfig.

Can we just move the handling of  system properties (-p and -D blocks) to the first blocks, and recreate `productConfig` if `jboss.home.dir` is found ?

soul2zimate: I misread your comment, don't bother with my previous comment via email(deleted).

I updated this PR, now it recreates `proudctConfig` if `jboss.home.dir` is found in system properties.

server/src/main/java/org/jboss/as/server/Main.java
```diff
@@ -212,6 +212,10 @@ public static ServerEnvironmentWrapper determineEnvironment(String[] args, Prope
                         value = arg.substring(idx + 1);
                     }
                     systemProperties.setProperty(name, value);
+                    if (ServerEnvironment.HOME_DIR.equals(name)) {
+                        // Re-create using updated properties
+                        productConfig = ProductConfig.fromFilesystemSlot(Module.getBootModuleLoader(), WildFlySecurityManager.getPropertyPrivileged(ServerEnvironment.HOME_DIR, null), systemProperties);
+                    }
```
yersan, line 218: I think the `WildFlySecurityManager.getPropertyPrivileged` usage here would be unnecessary, indeed it could be just replaced with `value` at this point and I don't see the need to read it again from System.properties using a privileged block.

server/src/main/java/org/jboss/as/server/Main.java
```diff
@@ -448,6 +452,10 @@ private static boolean processProperties(ProductConfig productConfig, final Stri
          try {
              url = makeURL(urlSpec);
              systemProperties.load(url.openConnection().getInputStream());
+             if (systemProperties.getProperty(ServerEnvironment.HOME_DIR) != null) {
+                 // Re-create using updated properties
+                 productConfig = ProductConfig.fromFilesystemSlot(Module.getBootModuleLoader(), WildFlySecurityManager.getPropertyPrivileged(ServerEnvironment.HOME_DIR, null), systemProperties);
+             }
```
yersan, line 458: I think the `WildFlySecurityManager.getPropertyPrivileged(ServerEnvironment.HOME_DIR, null)` can be replaced here with `systemProperties.getProperty(ServerEnvironment.HOME_DIR)` and retrieved without a privileged block.

darranl: +1

soul2zimate: Fixed.

yersan (approved):

PR #5941, opened 2024-04-08
[WFCORE-6756] Subsystem model bump and transformer adjustments needed for the dynamic SSLContext support

Skyllarr:
https://issues.redhat.com/browse/WFCORE-6756

### DIFF
--- a/elytron/src/main/java/org/wildfly/extension/elytron/ElytronExtension.java
+++ b/elytron/src/main/java/org/wildfly/extension/elytron/ElytronExtension.java
@@ -75,8 +75,9 @@ public class ElytronExtension implements Extension {
     static final ModelVersion ELYTRON_16_0_0 = ModelVersion.create(16);
     static final ModelVersion ELYTRON_17_0_0 = ModelVersion.create(17);
     static final ModelVersion ELYTRON_18_0_0 = ModelVersion.create(18);
+    static final ModelVersion ELYTRON_19_0_0 = ModelVersion.create(19);
 
-    private static final ModelVersion ELYTRON_CURRENT = ELYTRON_18_0_0;
+    private static final ModelVersion ELYTRON_CURRENT = ELYTRON_19_0_0;
 
     static final String ISO_8601_FORMAT = "yyyy-MM-dd'T'HH:mm:ss.SSSZ";
 
--- a/elytron/src/main/java/org/wildfly/extension/elytron/ElytronSubsystemTransformers.java
+++ b/elytron/src/main/java/org/wildfly/extension/elytron/ElytronSubsystemTransformers.java
@@ -54,6 +54,7 @@ import static org.wildfly.extension.elytron.ElytronExtension.ELYTRON_15_1_0;
 import static org.wildfly.extension.elytron.ElytronExtension.ELYTRON_16_0_0;
 import static org.wildfly.extension.elytron.ElytronExtension.ELYTRON_17_0_0;
 import static org.wildfly.extension.elytron.ElytronExtension.ELYTRON_18_0_0;
+import static org.wildfly.extension.elytron.ElytronExtension.ELYTRON_19_0_0;
 import static org.wildfly.extension.elytron.ElytronExtension.ELYTRON_1_2_0;
 import static org.wildfly.extension.elytron.ElytronExtension.ELYTRON_2_0_0;
 import static org.wildfly.extension.elytron.ElytronExtension.ELYTRON_3_0_0;
@@ -105,6 +106,8 @@ public final class ElytronSubsystemTransformers implements ExtensionTransformerR
     public void registerTransformers(SubsystemTransformerRegistration registration) {
         ChainedTransformationDescriptionBuilder chainedBuilder = TransformationDescriptionBuilder.Factory.createChainedSubystemInstance(registration.getCurrentSubsystemVersion());
 
+        // 19.0.0 (WildFly 32) to 18.0.0 (WildFly 29)
+        from19(chainedBuilder);
         // 18.0.0 (WildFly 29) to 17.0.0 (WildFly 28)
         from18(chainedBuilder);
         // 17.0.0 (WildFly 28) to 16.0.0 (WildFly 27)
@@ -142,8 +145,15 @@ public final class ElytronSubsystemTransformers implements ExtensionTransformerR
         // 2.0.0 (WildFly 12) to 1.2.0, (WildFly 11 and EAP 7.1.0)
         from2(chainedBuilder);
 
-        chainedBuilder.buildAndRegister(registration, new ModelVersion[] { ELYTRON_17_0_0, ELYTRON_16_0_0, ELYTRON_15_1_0, ELYTRON_15_0_0, ELYTRON_14_0_0, ELYTRON_13_0_0, ELYTRON_12_0_0, ELYTRON_11_0_0, ELYTRON_10_0_0, ELYTRON_9_0_0,
+        chainedBuilder.buildAndRegister(registration, new ModelVersion[] { ELYTRON_18_0_0, ELYTRON_17_0_0, ELYTRON_16_0_0, ELYTRON_15_1_0, ELYTRON_15_0_0, ELYTRON_14_0_0, ELYTRON_13_0_0, ELYTRON_12_0_0, ELYTRON_11_0_0, ELYTRON_10_0_0, ELYTRON_9_0_0,
                 ELYTRON_8_0_0, ELYTRON_7_0_0, ELYTRON_6_0_0, ELYTRON_5_0_0, ELYTRON_4_0_0, ELYTRON_3_0_0, ELYTRON_2_0_0, ELYTRON_1_2_0 });
+    }
+
+    private static void from19(ChainedTransformationDescriptionBuilder chainedBuilder) {
+        ResourceTransformationDescriptionBuilder builder = chainedBuilder.createBuilder(ELYTRON_19_0_0, ELYTRON_18_0_0);
+
+        builder.rejectChildResource(PathElement.pathElement(ElytronDescriptionConstants.DYNAMIC_CLIENT_SSL_CONTEXTS));
+        builder.rejectChildResource(PathElement.pathElement(ElytronDescriptionConstants.DYNAMIC_CLIENT_SSL_CONTEXT));
     }
 
     private static void from18(ChainedTransformationDescriptionBuilder chainedBuilder) {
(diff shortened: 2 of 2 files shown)


elytron/src/main/java/org/wildfly/extension/elytron/ElytronSubsystemTransformers.java
```diff
@@ -142,10 +145,17 @@ public void registerTransformers(SubsystemTransformerRegistration registration)
         // 2.0.0 (WildFly 12) to 1.2.0, (WildFly 11 and EAP 7.1.0)
         from2(chainedBuilder);
 
-        chainedBuilder.buildAndRegister(registration, new ModelVersion[] { ELYTRON_17_0_0, ELYTRON_16_0_0, ELYTRON_15_1_0, ELYTRON_15_0_0, ELYTRON_14_0_0, ELYTRON_13_0_0, ELYTRON_12_0_0, ELYTRON_11_0_0, ELYTRON_10_0_0, ELYTRON_9_0_0,
+        chainedBuilder.buildAndRegister(registration, new ModelVersion[] { ELYTRON_18_0_0, ELYTRON_17_0_0, ELYTRON_16_0_0, ELYTRON_15_1_0, ELYTRON_15_0_0, ELYTRON_14_0_0, ELYTRON_13_0_0, ELYTRON_12_0_0, ELYTRON_11_0_0, ELYTRON_10_0_0, ELYTRON_9_0_0,
                 ELYTRON_8_0_0, ELYTRON_7_0_0, ELYTRON_6_0_0, ELYTRON_5_0_0, ELYTRON_4_0_0, ELYTRON_3_0_0, ELYTRON_2_0_0, ELYTRON_1_2_0 });
     }
 
+    private static void from19(ChainedTransformationDescriptionBuilder chainedBuilder) {
```
Skyllarr, line 152: @pferraro Following from your comment [here](https://github.com/wildfly/wildfly-core/pull/4311#discussion_r1534498261), I have added a new version 19 of the elytron subsystem, and added schemas for both the DEFAULT and the COMMUNITY stability for this version.  I also added a transformer from the new version 19 to version 18, that you can see here. 

I have ignored the stability levels when I added this transformer. And I did not add any transformers between different stability levels of the same version. Just wanted to double check that this is correct, please? Thank you!

Skyllarr:
@fjuma Just a note that there is a failure in the org.wildfly.test.manual.management.ManagementOnlyModeTestCase.java that has this cause:

`Caused by: javax.xml.stream.XMLStreamException: WFLYCTL0505: elytron-oidc-client subsystem namespace urn:wildfly:elytron-oidc-client:preview:2.0 is not enabled by the current stability level`

but it is not related to this PR and it is occuring in other PRs as well

yersan:
@Skyllarr @fjuma I have not investigated it yet, but the Windows Job passed, and also the nightly Jobs that run the integration between WIldFly and WildFly Core, so I'm going to kick it off again to see how it goes.

yersan:
> Windows Job passed, and also the nightly Jobs

This is because the test ordering when the OIDC tests are executed before ManagementOnlyModeTestCase, the latter will fail

yersan:
I've just opened https://issues.redhat.com/browse/WFLY-19219 to track it

Skyllarr:
> I've just opened https://issues.redhat.com/browse/WFLY-19219 to track it

Thank you @yersan !

elytron/src/main/java/org/wildfly/extension/elytron/ElytronSubsystemSchema.java
```diff
@@ -53,9 +53,10 @@ public enum ElytronSubsystemSchema implements PersistentSubsystemSchema<ElytronS
     VERSION_16_0(16),
     VERSION_17_0(17),
     VERSION_18_0(18),
-    VERSION_18_0_COMMUNITY(18, Stability.COMMUNITY),
+    VERSION_19_0(19),
+    VERSION_19_0_COMMUNITY(19, Stability.COMMUNITY),
```
pferraro, line 57: WFCORE-6756 indicates the need to increment the subsystem model.  Why is the XML schema version changing?

elytron/src/main/java/org/wildfly/extension/elytron/ElytronSubsystemSchema.java
```diff
@@ -192,8 +193,8 @@ private void addCredentialStoreParser(PersistentResourceXMLDescription.Persisten
 
     private void addTlsParser(PersistentResourceXMLDescription.PersistentResourceXMLBuilder builder) {
         TlsParser tlsParser = new TlsParser();
-        if (this.since(ElytronSubsystemSchema.VERSION_18_0_COMMUNITY) && this.enables(getDynamicClientSSLContextDefinition())) {
-            builder.addChild(tlsParser.tlsParserCommunity_18_0);
+        if (this.since(ElytronSubsystemSchema.VERSION_19_0_COMMUNITY) && this.enables(getDynamicClientSSLContextDefinition())) {
+            builder.addChild(tlsParser.tlsParserCommunity_19_0);
```
pferraro, line 197: Why is this changing?

elytron/src/main/java/org/wildfly/extension/elytron/ElytronSubsystemSchema.java
Skyllarr: @pferraro we are always bumping the schema when incrementing the subsystem model, see for example previous bump jira https://issues.redhat.com/browse/WFCORE-6312 . 
So I have moved the COMMUNITY resource dynamic-ssl-context from the community_18 schema to the new schema community_19, if that is okay for this Final release? Maybe that's too late now, let me know. @fjuma Do you agree we should increment both the model and the XML schema?

pferraro: > we are always bumping the schema when incrementing the subsystem model

That is not correct.  Schema versions and model versions are not inherently related.  A new version of a subsystem schema is only required if there are changes to the XML representation of the model.  A new version of a subsystem model is only required when there are changes to the subsystem's resource description (read recursively).  The XML representation change change without the model, just as the model can change without affecting the XML representation.

> see for example previous bump jira https://issues.redhat.com/browse/WFCORE-6312

As a general note, it is not advisable to create JIRA issues for this.
Changes like WFCORE-6312 are not supposed to be merged independently from a schema and/or model change that requires it - otherwise this can easily result in redundant versions.  While redundant model versions have no signficant cost (other than being a potential source of confusion), redundant schema versions require parser registration and are otherwise a waste of heap space.  Tracking this creates a completely avoidable burden to the release coordinator, who would otherwise need to keep track of version bumps for every subsystem, and rollback changes if there were no associated subsystem changes.  Consequently, version bumps to a subsystem model and/or schema are not appropriate as a standalone pull request.

> So I have moved the COMMUNITY resource dynamic-ssl-context from the community_18 schema to the new schema community_19

Unless the XML representation of the subsystem's model has changed, there is reason to change this.

elytron/src/main/java/org/wildfly/extension/elytron/ElytronSubsystemSchema.java
pferraro: > @pferraro So you think I should just bump the model version and don't touch the schema 18 at this point, right?

Right.

elytron/src/main/java/org/wildfly/extension/elytron/ElytronExtension.java
```diff
@@ -75,8 +75,9 @@ public class ElytronExtension implements Extension {
     static final ModelVersion ELYTRON_16_0_0 = ModelVersion.create(16);
     static final ModelVersion ELYTRON_17_0_0 = ModelVersion.create(17);
     static final ModelVersion ELYTRON_18_0_0 = ModelVersion.create(18);
+    static final ModelVersion ELYTRON_19_0_0 = ModelVersion.create(19);
```
pferraro, line 78: OK - it looks like ModelVersion 18.0.0 was associated with WF 29-31, meaning that we need a new ModelVersion 19.0.0 to identify the model changes for WF32 and later, correct?
(I'm trying to make sure I understand the context of this change)

elytron/src/main/java/org/wildfly/extension/elytron/ElytronSubsystemSchema.java
Skyllarr: @pferraro I see, thank you very much for explanation! That makes sense, I will remove the schema update from this PR

elytron/src/main/java/org/wildfly/extension/elytron/ElytronExtension.java
Skyllarr: @pferraro Yes that's correct. The context of this change is just to fix the fact that I did not do the subsystem model bump when I added a resource to the community stability level previously.

Skyllarr:
@pferraro I removed the schema bump, thank you!

Skyllarr:
@darranl  / @fjuma Can you please review this one, thanks!

elytron/src/main/java/org/wildfly/extension/elytron/ElytronSubsystemSchema.java
darranl: On the general note I would say it is preferred that we do use a Jira issue for the model bump - where we have multiple engineers working on features in a subsystem independently we do want a coordinated approach so one does the work handling the bump(s) and then the individual features can depend on the common commits.

Yes we do need to ensure the bump only goes in as a feature goes in and generally this can happen when the first of any are merged.  Working at different stability levels is going to confuse that a bit more now so we may need to think some more on the coordination.

darranl (approved):

elytron/src/main/java/org/wildfly/extension/elytron/ElytronSubsystemSchema.java
fjuma: +1, I agree that it's good to create a Jira issue for the model bump. Typically, what we do is get a draft PR open with the changes needed so that others working on new features can make use of the common commits. We make sure that the PR is only merged once we have approved the PRs for at least one of the features that make use of it.

fjuma (approved):
Thanks @Skyllarr!

yersan:
This seems to be ready to go, @pferraro do you have any additional requests? otherwise, I'll mark it ready-for-merge

elytron/src/main/java/org/wildfly/extension/elytron/ElytronSubsystemTransformers.java
pferraro: Since the transformer is rejection only, this should function correctly for either stability level (rejecting a child resource with no registration should just be no-op).
Additionally, stability level of the target host is not yet exposed via SubsystemTransformerRegistration - so this would not yet be possible anyway.

pferraro (approved):

yersan:
Thanks @Skyllarr and all the reviewers!

PR #5944, opened 2024-04-08
Fix for WFCORE-6776, Can't specify a stability level when starting a bootable JAR.

jfdenise:
Issue: https://issues.redhat.com/browse/WFCORE-6776

### DIFF
--- a/bootable-jar/runtime/src/main/java/org/wildfly/core/jar/runtime/CmdUsage.java
+++ b/bootable-jar/runtime/src/main/java/org/wildfly/core/jar/runtime/CmdUsage.java
@@ -9,6 +9,7 @@ import org.jboss.as.controller.persistence.ConfigurationExtensionFactory;
 
 import org.jboss.as.process.CommandLineArgumentUsage;
 import org.jboss.as.process.CommandLineConstants;
+import org.jboss.as.version.ProductConfig;
 import org.wildfly.core.jar.runtime._private.BootableJarLogger;
 
 /**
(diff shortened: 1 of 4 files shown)


yersan:
@jfdenise FYI: ManagementOnlyModeTestCase is unrelated and tracked on a different Jira https://issues.redhat.com/browse/WFLY-19219

bootable-jar/runtime/src/main/java/org/wildfly/core/jar/runtime/CmdUsage.java
```diff
@@ -48,6 +48,9 @@ public static void init() {
         addArguments(CommandLineConstants.SECMGR);
         instructions.add(BootableJarLogger.ROOT_LOGGER.argSecurityManager());
 
+        addArguments(CommandLineConstants.STABILITY + "=<value>");
+        instructions.add(BootableJarLogger.ROOT_LOGGER.argStability());
```
yersan, line 52: Would it be possible for the Bootable JAR to get the available stability levels from ProductConfig.getStabilitySet()?

In addition, we have a kind of restriction here where the stability levels should be hidden when there is only one permissible stability level. See https://github.com/wildfly/wildfly-core/blob/b20d4c06b41542316af51ff213bb564f2574d3d3/server/src/main/java/org/jboss/as/server/CommandLineArgumentUsageImpl.java#L83-L86 as a reference. CC @pferraro

jfdenise: @yersan , thank-you. I adjusted to comply with what we do in the server.

yersan (approved):

yersan:
Thanks @jfdenise

PR #5947, opened 2024-04-09
[WFCORE-6755] Move the org.wildfly.security:wildfly-elytron-dynamic-ssl artifact into its own module

Skyllarr:
https://issues.redhat.com/browse/WFCORE-6755


yersan:
Hello @Skyllarr , we need a WFLY Jira/PR in place to fix the layers test case in WildFly. We can merge this WFCORE PR as the latest PR for this release, but we need the Jira/PR to run manually the integration Job.

An alternative is to hack the WidlFly LayersTestCase to ignore this specific layer, and then remove the hack once Core is bumped on WildFly

Skyllarr:
> Hello @Skyllarr , we need a WFLY Jira/PR in place to fix the layers test case in WildFly. We can merge this WFCORE PR as the latest PR for this release, but we need the Jira/PR to run manually the integration Job.
> 
> An alternative is to hack the WidlFly LayersTestCase to ignore this specific layer, and then remove the hack once Core is bumped on WildFly

Thanks @yersan , I've created https://issues.redhat.com/browse/WFLY-19223 and I'm working on fixing the test

yersan:
Integration Jobs:

https://ci.wildfly.org/buildConfiguration/WF_WildFlyCoreIntegrationExperiments/429188 (Galleon Integration)
https://ci.wildfly.org/buildConfiguration/WF_WildFlyCoreIntegrationExperiments/429204 (WildFly Preview Integration JDK 17)

core-feature-pack/common/src/main/resources/modules/system/layers/base/org/wildfly/security/elytron-dynamic-ssl/main/module.xml
```diff
@@ -0,0 +1,25 @@
...
+
+<!--
+  ~ Copyright The WildFly Authors
+  ~ SPDX-License-Identifier: Apache-2.0
+  -->
+<module xmlns="urn:jboss:module:1.9" name="org.wildfly.security.elytron-dynamic-ssl">
+
+    <properties>
+        <property name="jboss.api" value="private"/>
+    </properties>
```
yersan, line 11: This module is for community stability level, so it requires jboss.stability=community property here. Otherwise, it will take the default stability

testsuite/layers/src/test/java/org/jboss/as/test/layers/LayersTestCase.java
```diff
@@ -26,6 +26,7 @@ public class LayersTestCase {
         // No patching modules in layers
         "org.jboss.as.patching",
         "org.jboss.as.patching.cli",
+        "org.wildfly.security.elytron-dynamic-ssl"
```
yersan, line 29: If the intention is to leave the usage of this module when the server is trimmed including the Elytron layer, this change could be fine but we need to add a comment above it explaining the reason why this module is not provided when using layers. 

I have doubts we want that, it looks like the intention could be to also provide the capabilities of the dynamic SSL context even when we are using the Elytron layer. If that's the case, I guess you have to register the module with https://github.com/wildfly/wildfly-core/blob/d5456e9f7565513984b539ba494f9805863a36c6/controller/src/main/java/org/jboss/as/controller/ResourceDefinition.java#L127, which in turns should be invoked depending on the current Stability level

yersan (changes requested):
@Skyllarr The integration test failed, if we really don't want to provision this module when we are trimming the server, the fix in WildFly Full should be to add it at https://github.com/wildfly/wildfly/blob/a53eba00bde46400befd053c4a104e7107da0071/testsuite/shared/src/main/java/org/jboss/as/test/shared/LayersTestBase.java#L45 instead.

But I have doubts whether this is want we want to do, I would expect this feature also when we are trimming the server but we are including the Elytron layer

testsuite/layers/src/test/java/org/jboss/as/test/layers/LayersTestCase.java
yersan: > should be invoked depending on the current Stability level

or as a different way to see it, be invoked when the management resource associated with this Elytron Dynamic SSL is being registered, which I would assume is registered for community only stability level

Skyllarr: @yersan  yes, this module should be provided when using elytron layer and the community stability. I'll fix this, thank you!

Skyllarr:
@yersan I've addressed your feedback. I closed the affiliated wildfly PR since these latest changes do not introduce a failure in wildfly testsuite. Thank you!

Skyllarr:
@darranl / @fjuma Please review, thanks!

yersan (approved):
LGTM, Integration full failure is unrelated. @fjuma / @darranl please review when you have a chance, thanks. I'll wait for any of your approvals before going with this.

elytron/src/main/java/org/wildfly/extension/elytron/DynamicSSLContextHelper.java
```diff
@@ -0,0 +1,34 @@
...
+class DynamicSSLContextHelper {
+
+    /**
+     * Get DynamicSSLContext instance from the provided authentication context
+     * @param authenticationContext authentication context to use with the DynamicSSLContext
+     * @return DynamicSSLContext instance
+     */
+    static SSLContext getDynamicSSLContextInstance(AuthenticationContext authenticationContext) {
+        try {
+            return new DynamicSSLContext(new DynamicSSLContextImpl(authenticationContext));
```
darranl, line 29: How about if they provision at the Default stability level but then try and run at community level - do we need to handle an error for when the module does not exist?

Skyllarr: @darranl Hm maybe I can use Class.forName instead of imports, and catch the ClassNotFoundException if that sounds good?

yersan: Would `getDynamicSSLContextInstance` get invoked if the dynamic SSL resource management resource is not registered?

elytron/src/main/java/org/wildfly/extension/elytron/ElytronDefinition.java
```diff
@@ -363,7 +363,7 @@ protected void revertUpdateToRuntime(OperationContext context, ModelNode operati
 
     @Override
     public void registerAdditionalRuntimePackages(ManagementResourceRegistration resourceRegistration) {
-        resourceRegistration.registerAdditionalRuntimePackages(RuntimePackageDependency.required("org.wildfly.security.elytron"));
+        resourceRegistration.registerAdditionalRuntimePackages(RuntimePackageDependency.required("org.wildfly.security.elytron"), RuntimePackageDependency.required("org.wildfly.security.elytron-dynamic-ssl"));
```
yersan, line 366: I missed that one, `RuntimePackageDependency.required` is not capability aware, so it is going to require the "org.wildfly.security.elytron-dynamic-ssl" when we are running in a higher level capability than community, e.g default.

Should register this module conditionally based on the current capability level where this module is needed. 
We could evolve the RuntimePackageDependency.required() to accept a Stability level and implement the dynamic check on that method, but that is out of the scope of this PR

yersan (changes requested):

elytron/src/main/java/org/wildfly/extension/elytron/DynamicSSLContextHelper.java
yersan: My understanding is this method won't be called when you are running on `default` since this is supplied by the add handler that provides the Dynamic SSL context, but that handler will never be used because the resource won't be available on a lower stability level.

Skyllarr: @yersan That was my original understanding as well. That the class DynamicSSLContextHelper would not be loaded if dynamic SSL resource is not registered... but was not completely sure

elytron/src/main/java/org/wildfly/extension/elytron/ElytronDefinition.java
Skyllarr: @yersan I fixed this with:

```
if (resourceRegistration.getStability().enables(Stability.COMMUNITY)) {
            resourceRegistration.registerAdditionalRuntimePackages(RuntimePackageDependency.required("org.wildfly.security.elytron-dynamic-ssl"));
 }
```

Thank you!

yersan: More on this, since it has been discussed also in Zulip: https://wildfly.zulipchat.com/#narrow/stream/174184-wildfly-developers/topic/Disable.20Specific.20Stability.20Levels.20at.20Provisioning.20Time

As originally mentioned at https://github.com/wildfly/wildfly-core/pull/5947#discussion_r1560868361, the requirement for this module should be done when the  Dynamic SSL resource is being added. That will remove the Stability level conditional check, since the Dynamic SSL resource will be only added when the server is launched at `community` level or lower levels.

elytron/src/main/java/org/wildfly/extension/elytron/DynamicSSLContextHelper.java
darranl: FYI there is a thread in Zulip with some more context but the general scenario is, what happens when a user provisions a server at the default stability level but then runs at the community level and tries to use this resource:

https://wildfly.zulipchat.com/#narrow/stream/174184-wildfly-developers/topic/Disable.20Specific.20Stability.20Levels.20at.20Provisioning.20Time/near/432858502

core-feature-pack/common/src/main/resources/modules/system/layers/base/org/wildfly/security/elytron-base/main/module.xml
```diff
@@ -112,5 +111,6 @@
         modules use the parser, they need to have visibility to this module.
         -->
         <module name="org.wildfly.client.config" export="true"/>
+        <module name="org.wildfly.security.elytron-dynamic-ssl" export="true" optional="true"/>
```
darranl, line 114: I am considering if this should not be here.

Maybe the subsystem should have it's own dependency on this new module specified and then in a deployment unit processor associate it with the deployment if running at community level or less.

elytron/src/main/java/org/wildfly/extension/elytron/ElytronDefinition.java
darranl: I think there is still more discussion to go - I think we are nearing the point the module is unlikely to be provisioned at all unless the provisioned configuration contains a definition of the resource but that is not what we are trying to achieve.

yersan: In any case, @Skyllarr notice that what you are checking there is if the resourceRegistration enables the `community` stability. That's not what we would need to check, we would need to check whether the resourceRegistration enables the current stability where the server is running at runtime.

So, if you launch your server with `default` stability, your additional package won't be required to be registered.

yersan: In any case, as discussed in Zulip, we should avoid this check and move the additional package registration to the child resource that requires it.

fjuma:
> There are 2 options that can be configured in galleon: `config-stability-level` and `package-stability-level` ( https://github.com/wildfly/wildfly-proposals/pull/555/files ) . From reading it seems to me that when `package-stability-level` option is configured to be `community` then the module is not provisioned / not written to disk when we provision for `default`, because it would contain `<property name="jboss.stability" value="community"/>` stability? But I am not sure about this at all and this has to be investigated more.
> 
> @fjuma @darranl @yersan Do you all agree to move this issue to the WF33 cycle? The added classes in elytron are private API, and changes in code in WF are internal. This small package might be included in installation when it should not be, but that might not be a blocker for this release

Yes, I'm ok with having this done for WildFly 33 instead. Thanks!

Skyllarr:
Thank you @yersan , I will work on the changes you mentioned in https://github.com/wildfly/wildfly-core/pull/5947#discussion_r1565387696 . 

Since this PR will require more discussion and review, we should skip this PR for this release, see comment from @fjuma above. It will be worked on in the WF33 cycle

yersan:
@Skyllarr there are still chances to get this one in for the current release since we are still waiting for others, but I am also OK with having this done for WildFly 33. I have no strong opinions against that.

@darranl would you share your feedback here? If you agree, we can downgrade the priority and see if it gets done before the release.

darranl:
If we have a couple more days and it is possible to conclude this I think it is fine to go in but I know this is also a busy week, if we are happy we have covered the scenarios no reason not to merge but if we are still testing etc lets not force it through.

Skyllarr:
> Downgrade the priority and see if it gets done before the release.

@yersan yes this approach sounds good! I will work on required changes and we'll see if it's done in time, but as I'm not that familiar with provisioning etc. it might take me some time to update this PR

elytron/src/main/java/org/wildfly/extension/elytron/ElytronDefinition.java
```diff
@@ -364,6 +365,9 @@ protected void revertUpdateToRuntime(OperationContext context, ModelNode operati
     @Override
     public void registerAdditionalRuntimePackages(ManagementResourceRegistration resourceRegistration) {
         resourceRegistration.registerAdditionalRuntimePackages(RuntimePackageDependency.required("org.wildfly.security.elytron"));
+        if (resourceRegistration.getStability().enables(Stability.COMMUNITY)) {
+            resourceRegistration.registerAdditionalRuntimePackages(RuntimePackageDependency.required("org.wildfly.security.elytron-dynamic-ssl"));
+        }
```
bstansberry, line 370: @Skyllarr Why is this here? This seems like something the ResourceDefinition returned from SSLDefinitions.getDynamicClientSSLContextDefinition() should handle. And there'd be no conditional logic as the availability of that ResourceDefinition is already conditional on the stability setting.

@jfdenise I think we should discuss this in zulip, not here, but I'm interested in your opinion re this kind of logic in the general sense, i.e. in cases where the relevant RD is always present but the need for a particular additional runtime package depends on the stability level. I *think* it's ok. The module would only be depended upon optionally, so something is needed to tell Galleon to provision it. We don't want to always tell Galleon to provision it, because the module.xml will have a stability setting, and if we always tell Galleon to provision it in RD. registerAdditionalRuntimePackages, but then the module.xml indicates that Galleon should not provision it, we have a conflict.

bstansberry:
@Skyllarr @yersan I think on Wednesday we should move this to WF 33, i.e. change the JIRA and switch the 24.x label on this PR to 25.x. But leave it as a blocker.

I think there are still things to discuss about it and I agree the presence of this jar on the system is not a blocker, as it's not usable except via the community-stability ResourceDefinition.

elytron/src/main/java/org/wildfly/extension/elytron/ElytronDefinition.java
Skyllarr: Thanks @yersan , I've moved the module registration  to the `registerAdditionalRuntimePackages` method when registering dynamic ssl context

elytron/src/main/java/org/wildfly/extension/elytron/ElytronDefinition.java
Skyllarr: @bstansberry Thanks, I moved the package registration to the registerAdditionalRuntimePackages method so it is registered when getDynamicClientSSLContextDefinition is invoked

core-feature-pack/common/src/main/resources/modules/system/layers/base/org/wildfly/security/elytron-base/main/module.xml
Skyllarr: CC @jfdenise , what do you think about this module registration? It should be provisioned only when provisioning with COMMUNITY stability

yersan (commented):
In my opinion, this is completed, although we are waiting for https://issues.redhat.com/browse/WFCORE-6805 to do more testing and see if the approach is the correct one. We are waiting to have the ability to limit the available stability levels based on the provisioning stability level

yersan:
> In my opinion, this is completed, although we are waiting for https://issues.redhat.com/browse/WFCORE-6805 to do more testing and see if the approach is the correct one. We are waiting to have the ability to limit the available stability levels based on the provisioning stability level

Well, current failures on integration tests are not expected, so it still requires investigation into what is causing them.

Skyllarr:
> In my opinion, this is completed, although we are waiting for https://issues.redhat.com/browse/WFCORE-6805 to do more testing and see if the approach is the correct one. We are waiting to have the ability to limit the available stability levels based on the provisioning stability level

I see, thanks @yersan ! I will look into the  integration tests failures

Skyllarr:
> > In my opinion, this is completed, although we are waiting for https://issues.redhat.com/browse/WFCORE-6805 to do more testing and see if the approach is the correct one. We are waiting to have the ability to limit the available stability levels based on the provisioning stability level
> 
> Well, current failures on integration tests are not expected, so it still requires investigation into what is causing them.

@yersan  This PR should solve the issues with LayersTestCase https://github.com/wildfly/wildfly/pull/17862 , locally it works for me

yersan:
@Skyllarr There is a relevant error on the "Galleon Linux - JDK 11 (Pull Request)" Job:

```
java.lang.AssertionError: Some expected modules have not been provisioned in test-all-layers: org.wildfly.security.elytron-dynamic-ssl
  at org.jboss.as.test.layers.LayersTest.testLayersModuleUse(LayersTest.java:172)
  at org.jboss.as.test.layers.LayersTestCase.testLayersModuleUse(LayersTestCase.java:111)
  at java.base/jdk.internal.reflect.NativeMethodAccessorImpl.invoke0(Native Method)
  at java.base/jdk.internal.reflect.NativeMethodAccessorImpl.invoke(NativeMethodAccessorImpl.java:62)
  at java.base/jdk.internal.reflect.DelegatingMethodAccessorImpl.invoke(DelegatingMethodAccessorImpl.java:43)
  at java.base/java.lang.reflect.Method.invoke(Method.java:566)
```

I don't know exactly what is causing that, but the problem seems to be that `elytron` galleon layer is not provisioning the `org.wildfly.security.elytron-dynamic-ssl`

I've just taken a look but I don't see the root cause, since Elytron subsystem is advertising to Galleon the requirement of this package at https://github.com/wildfly/wildfly-core/pull/5947/files#diff-263be4d2ab5108fc2de76f44766487703d2444e7bcc70bac0aaa4be6a27a9089R114

@jfdenise Could you shed some light on this?

I've just done a quick test and, although the relevant package is correctly advertised with `resourceRegistration.registerAdditionalRuntimePackages(RuntimePackageDependency.required(dependencyPackageName))`, the expected set of provisioned packages for the Elytron galleon layer does not includes the `org.wildfly.security.elytron-dynamic-ssl` package name.

What we are missing here?

yersan:
@jfdenise a friendly reminder, do you have any points about why Galleon does not provision the advertised package?

yersan:
Hello @jfdenise this is still a blocker for us, could you provide your feedback about why Galleon is not working as we expect for this case? I am not sure if we are missing something here.

yersan:
@bstansberry

> Is the resource part of the elytron subsystem configuration provided by the 'elytron' layer? Or, less likely, some other layer included in the test-all-layers set?

The resource is just a subresource of the standard Elytron subsystem, however, it is distributed via an optional module, which, in turn, is never provisioned by any other Galleon Layer.

> The resourceRegistration.registerAdditionalRuntimePackages stuff is only relevant if the resource is part of the configuration. It doesn't tell Galleon to provision things that are not in the configuration just in case they user might want to add them later.

Ok, thanks for clarifying it. Now I understand the issue is that since not configured by the Elytron subsystem when it is provisioned as a Galleon layer, then the source code will never advertise it requires such a package.

However, this subresource does not require any Elytron subsystem configuration, so, I don't know how it can be added to the Elytron layer as an empty feature.

I can only think of adding it as a required package for the Elytron layer at:

https://github.com/wildfly/wildfly-core/blob/bb0806bacc370a49141f0c9cd9be4fb12066a0bb/core-feature-pack/galleon-common/src/main/resources/layers/standalone/elytron/layer-spec.xml#L16-L20

But not sure if that is the correct approach. It fixes the problem, since it makes it provisioned always when Elytron is configured. but not sure if that is the approach or even if we add it there, whether we would still need the resourceRegistration.registerAdditionalRuntimePackages stuff

darranl:
I think to invert some of the discussions the two times we would expect that this module is NOT provisioned are:

 - provisioning a server that does not contain the elytron extension
 - provisioning a server that only supports default stability level

In all other cases this module should be present in the installation

yersan:
Superseded by #6066

PR #5948, opened 2024-04-09
[WFCORE-6777] Do not allow mixed domains for preview and experimental…

yersan:
… stability levels

Jira issue: https://issues.redhat.com/browse/WFCORE-6777

### DIFF
--- a/host-controller/src/main/java/org/jboss/as/host/controller/mgmt/HostControllerRegistrationHandler.java
+++ b/host-controller/src/main/java/org/jboss/as/host/controller/mgmt/HostControllerRegistrationHandler.java
@@ -61,6 +61,7 @@ import org.jboss.as.protocol.mgmt.ManagementRequestHandler;
 import org.jboss.as.protocol.mgmt.ManagementRequestHandlerFactory;
 import org.jboss.as.protocol.mgmt.ManagementRequestHeader;
 import org.jboss.as.protocol.mgmt.ManagementResponseHeader;
+import org.jboss.as.version.Version;
 import org.jboss.dmr.ModelNode;
 import org.jboss.dmr.Property;
 import org.jboss.remoting3.Channel;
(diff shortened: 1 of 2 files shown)


host-controller/src/main/java/org/jboss/as/host/controller/mgmt/HostControllerRegistrationHandler.java
```diff
@@ -338,6 +340,13 @@ public void execute(final OperationContext context, final ModelNode operation) t
                     registrationContext.failed(failure, SlaveRegistrationException.ErrorCode.INCOMPATIBLE_VERSION, failure.getMessage());
                     throw failure;
                 }
+                // Ensure mixed domains are not allowed for preview and experimental stability levels
+                if (domainInfo.getStability().enables(Stability.PREVIEW) && (Version.MANAGEMENT_MAJOR_VERSION != major || Version.MANAGEMENT_MINOR_VERSION != minor)) {
```
pferraro, line 344: Rather than hard code this, why not base this threshold on ProductConfig?
e.g.
```java
if (!domainInfo.getDefaultStability().enables(domainInfo.getStability()) && ...)

yersan: @pferraro looking at your snipped, I am not following your concern here:
- There is no `domainInfo.getDefaultStability()` method, and if it were there, that would provide the default stability level.
- I don't see how checking that the default stability level does not enable the current configured stability level can help with this case.

Could you expand your concept about using ProductConfig here?

Are you suggesting hiding the logic of this check inside of a `ProductConfig` static method that receives as parameters the major/minor version of the HC being registered and the current stability level where we are running? well, it could make sense, but I don't think any other parties would be interested in knowing when this condition is satisfied outside of the registration handler. To use it outside of the registration handler you would need to have beforehand the major/minor version of the remote host you want to verify: I mean you have to read them from the remote host and bring them to the current JVM.

pferraro: Sorry for the confusion, I've updated the snippet above - hopefully that makes more sense now.
By generalizing the logic, we make it more resilient to future changes.  In this case, disallowing mixed domains for any stability level less than the product default, i.e. COMMUNITY for WildFly, and DEFAULT for downstream.

yersan: Np, ok, now I see what you mean. 

In your snipped you are not even comparing with the remote host stability, but at this point both are equals, so we can use both `domainInfo.getStability()` or `hostInfo.getStability()`. Using the hostInfo here looks a bit more clear to me.

pferraro: The whole point of this jira is to validate that the stability level of the _domain_ controller should allows any mixed domains (independent of the stability level of the host).  
 Changing this to use hostInfo seems more confusing to me, but, to your point, they are the same.  However, if the logic in this method get reordered the original semantics would not be preserved.

yersan: @pferraro maybe we are not on the same page, or I am not understanding you well:

>The whole point of this jira is to validate that the stability level of the domain controller should allows any mixed domains (independent of the stability level of the host - which is a separate concern).

The whole point is to reject mixed domains for preview and experimental, so the changes here are related to the stability levels on mixed domains

> Changing this to use hostInfo seems more confusing to me, but, to your point, they are the same. However, if the logic in this method get reordered the original semantics would not be preserved.

Well, I finally left it as you originally suggested by using the domain info. And yes, reordering the original would break the semantics, I was counting on that too.

pferraro (approved):

yersan:
Thanks for the review, @pferraro

PR #5964, opened 2024-04-15
WFCORE-6791 Tidy wildfly-discovery code to better use as a reference for wildfly-subsystem usage

pferraro:
Consolidate ResourceDefinitionRegistrar and ResourceServiceConfigurator to single class.
Reduce visibility of static fields where sensible.

https://issues.redhat.com/browse/WFCORE-6791


yersan:
@pferraro what's your confidence whether these classes are not used outside of WildFly?, more formally, do you think it is worth for WF32 without deprecating the old ones?

pferraro:
There are zero consumers within WildFly let alone outside...

yersan (approved):

bstansberry:
Thanks @pferraro and @yersan

PR #5975, opened 2024-04-22
[WFCORE-6799] Upgrade SLF4J from 2.0.12 to 2.0.13

yersan:
Jira issue: https://issues.redhat.com/browse/WFCORE-6799

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -230,7 +230,7 @@
         <version.org.mockito>3.10.0</version.org.mockito>
         <version.org.jmockit>1.49</version.org.jmockit>
         <version.org.projectodd.vdx>1.1.6</version.org.projectodd.vdx>
-        <version.org.slf4j>2.0.12</version.org.slf4j>
+        <version.org.slf4j>2.0.13</version.org.slf4j>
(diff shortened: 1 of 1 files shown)


jamezp (approved):
This is only used for testing. I can't open up the failed CI job for some reason, but I don't think it would be related.

yersan:
It is unrelated has failed with a similar stack trace for other PRs. In addition, since it is a test only dependency, it is not being driven to WildFly via WildFly Core, so any failure on integration tests should be unrelated.

CC @rhusar in relation to https://issues.redhat.com/browse/WFLY-14064, it looks like the test case is still failing

rhusar:
> It is unrelated has failed with a similar stack trace for other PRs. In addition, since it is a test only dependency, it is not being driven to WildFly via WildFly Core, so any failure on integration tests should be unrelated.
> 
> CC @rhusar in relation to https://issues.redhat.com/browse/WFLY-14064, it looks like the test case is still failing

Thanks, I ll have a look (and yes, definitely unrelated)

yersan:
Thanks @jamezp @rhusar

PR #5976, opened 2024-04-22
[WFCORE-6801] Use maven.compiler.release to ensure the desired binary…

yersan:
… compilation level

It also removes the `maven.compiler.target` and `maven.compiler.source` properties since they will be replaced by the `--release` option supported by the maven compiler plugin 3.6. The maven compiler plugin version is inherited from the jboss-parent maven module, so `maven.compiler.release` will be available for the project. Behind scenes, the maven compiler plugging should be ignoring the `--target` and `--source` options when `--release` is in place since `target/source` is not compatible with `release`

Jira issue: https://issues.redhat.com/browse/WFCORE-6801


yersan:
Thanks @jamezp

PR #5985, opened 2024-04-28
WFCORE-6808 Add capability-resolving variant of child ServiceTarget-based ServiceInstaller.Builder

pferraro:
https://issues.redhat.com/browse/WFCORE-6808

### DIFF
--- a/controller/src/main/java/org/jboss/as/controller/RequirementServiceTarget.java
+++ b/controller/src/main/java/org/jboss/as/controller/RequirementServiceTarget.java
@@ -5,7 +5,15 @@
 
 package org.jboss.as.controller;
 
+import java.util.function.Supplier;
+
+import org.jboss.as.controller.capability.CapabilityServiceSupport;
+import org.jboss.msc.Service;
+import org.jboss.msc.service.DelegatingServiceBuilder;
+import org.jboss.msc.service.DelegatingServiceTarget;
 import org.jboss.msc.service.LifecycleListener;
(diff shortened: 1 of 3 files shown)


yersan:
@pferraro even for simple refactorings, I would say we should have them properly tracked in a valid Jira that, in a nutshell, explains the motivation/reason of the refactor. That would help to understand the change and track it down in the future.

What's your opinion about that?

yersan:
@pferraro WFCORE-6808 or WFCORE-6806 (linked to this PR)? there are two commits, could you fix the message on both of them? 

If this PR is for two Jiras I would expect:
-  two Jira links on the PR description
-  two Jira IDs in the PR title
-  each commit message with its Jira ID in the commit message.

pferraro:
There is only 1 jira - but I did transcribe it all kinds of wrong. :/ Fixing...

yersan:
@pferraro Would you agree by squashing both commits or adding the WFCORE-6808 Jira ID to the first commit?

That would help to follow up on the changes in the future. Having a single commit without Jira ID makes finding the Jira that caused the change more difficult. 

We recommend that each commit has the [Jira ID on its message](https://docs.wildfly.org/32/Hacking_On_WildFly.html#commit-message) and I think it is a good practice even for simple refactorings.

pferraro:
Done.

yersan (approved):

yersan:
Thanks @pferraro

PR #5990, opened 2024-05-07
WFCORE-6817 Operation parameters and reply attributes not enabled by the current stability level still appear in the resource description

pferraro:
https://issues.redhat.com/browse/WFCORE-6817


yersan (approved):

yersan:
Thanks @pferraro

PR #5993, opened 2024-05-09
[WFCORE-6823] Don't register the Jakarta AuthConfigFactory in the host controller.

darranl:
https://issues.redhat.com/browse/WFCORE-6823


yersan:
@darranl just asking, shouldn't this change be driven by the RFE that would move Jakarta Authentication to its own subsystem?

bstansberry:
@darranl The JIRA description could use an update then, as the 'bug' aspect is unclear.

darranl:
@yersan / @bstansberry I have updated the description in the issue to say it is to fix the bug,

yersan (approved):

yersan:
Thanks @darranl @bstansberry

PR #5999, opened 2024-05-15
WFCORE-6802 [Preview] OCSP stapling support

PrarthonaPaul:
Issue: https://issues.redhat.com/browse/WFCORE-6802
Bump issue: https://issues.redhat.com/browse/WFCORE-6879
Depends on: https://github.com/wildfly-security/wildfly-elytron/pull/2144
Proposal Doc: https://github.com/wildfly/wildfly-proposals/pull/564
Todo: 
fix client-ssl-context subsystem test (please see comment https://github.com/wildfly/wildfly-core/pull/5999#discussion_r1626050478)

### DIFF
--- a/elytron/src/main/java/org/wildfly/extension/elytron/SSLDefinitions.java
+++ b/elytron/src/main/java/org/wildfly/extension/elytron/SSLDefinitions.java
@@ -26,6 +26,7 @@ import static org.wildfly.extension.elytron.FileAttributeDefinitions.PATH;
 import static org.wildfly.extension.elytron.FileAttributeDefinitions.RELATIVE_TO;
 import static org.wildfly.extension.elytron.FileAttributeDefinitions.pathName;
 import static org.wildfly.extension.elytron._private.ElytronSubsystemMessages.ROOT_LOGGER;
+import static org.wildfly.security.provider.util.ProviderUtil.findProvider;
 
 import java.io.File;
 import java.io.FileInputStream;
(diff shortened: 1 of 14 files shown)


elytron/src/main/java/org/wildfly/extension/elytron/SSLDefinitions.java
```diff
@@ -1689,4 +1754,40 @@ public InjectedValue<PathManager> getPathManagerInjector() {
...
+    public static KeyStore getKeyStoreFromTrustManager(X509ExtendedTrustManager trustManager) throws Exception {
+        // TODO: proporly extract the keystore from the trustmanager
+        KeyStore trustStore = KeyStore.getInstance(KeyStore.getDefaultType());
+        trustStore.load(null, null);
+        X509Certificate[] trustedCerts = trustManager.getAcceptedIssuers();
+        for (X509Certificate certificate : trustedCerts) {
+            trustStore.setCertificateEntry(certificate.getSerialNumber().toString(), certificate);
+        }
+        return trustStore;
+    }
```
PrarthonaPaul, line 1791: I don't think this is valid. 
I could not find a more valid way of getting the truststore from a trustmanager, which is needed to build a X509RevocationTrustManager.

elytron/src/test/java/org/wildfly/extension/elytron/TlsTestCase.java
```diff
@@ -644,6 +648,53 @@ public void testOcspSimple() {
...
+        operation.get(ClientConstants.OP).set(ClientConstants.WRITE_ATTRIBUTE_OPERATION);
+        operation.get(ClientConstants.NAME).set(ElytronDescriptionConstants.OCSP_STAPLING + "." + ElytronDescriptionConstants.RESPONDER_OVERRIDE);
+        operation.get(ClientConstants.VALUE).set(true);
+        operation.get(ClientConstants.OP).set(ClientConstants.WRITE_ATTRIBUTE_OPERATION);
+        operation.get(ClientConstants.NAME).set(ElytronDescriptionConstants.OCSP_STAPLING + "." + ElytronDescriptionConstants.IGNORE_EXTENSIONS);
+        operation.get(ClientConstants.VALUE).set(true);
+        Assert.assertEquals(SUCCESS, services.executeOperation(operation).get(OUTCOME).asString());
+    }
+
+//    @Test
```
PrarthonaPaul, line 686: This is commented out because when executing these commands from a test, the following line fails when creating the client context. 
https://github.com/wildfly/wildfly-core/blob/e409c5b6392b21b4f6732f74df4f2c4b398343ab/controller/src/main/java/org/jboss/as/controller/OperationContextImpl.java#L629
However, when running this add command from the management CLI, this does not fail. I am not sure if that failure is expected due to the set up of the tests.

PR #6000, opened 2024-05-16
[WFCORE-6819] Show current stability level with the server boot message

yersan:
Jira issue: https://issues.redhat.com/browse/WFCORE-6819

### DIFF
--- a/server/src/main/java/org/jboss/as/server/logging/ServerLogger.java
+++ b/server/src/main/java/org/jboss/as/server/logging/ServerLogger.java
@@ -1475,6 +1475,9 @@ public interface ServerLogger extends BasicLogger {
     @Message(id = Message.NONE, value = "- Server configuration file in use: %s")
     String serverConfigFileInUse(String serverConfigFile);
 
+    @Message(id = Message.NONE, value = "- Minimum feature stability level: %s")
+    String serverStabilityInUse(Stability stability);
+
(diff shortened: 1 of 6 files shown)


server/src/main/java/org/jboss/as/server/logging/ServerLogger.java
```diff
@@ -1475,6 +1475,9 @@ public interface ServerLogger extends BasicLogger {
     @Message(id = Message.NONE, value = "- Server configuration file in use: %s")
     String serverConfigFileInUse(String serverConfigFile);
 
+    @Message(id = Message.NONE, value = "- Server Stability level: %s")
```
bstansberry, line 1478: I think something like

"Minimum feature stability level: %s"

This setting doesn't describe the overall stability of the server or even most of its features. And many readers of the message will have no idea what this means, so it can't be alarming.

bstansberry (changes requested):

server/src/main/java/org/jboss/as/server/logging/ServerLogger.java
yersan: @bstansberry Done, using "Minimum feature stability level: %s" for server and host controller. I also removed the Server and Host Controller words from the sentence since the previous message already establishes such a context, let me know otherwise.

bstansberry (approved):

yersan:
Thanks @bstansberry

PR #6014, opened 2024-05-27
[WFCORE-6834] wildfly-elytron-integration jar duplicated in server modules

ivassile:
Issue: https://issues.redhat.com/browse/WFCORE-6834

Requires [WFLY-19378](https://issues.redhat.com/browse/WFLY-19378) https://github.com/wildfly/wildfly/pull/17928 to pass CI.

### DIFF
--- a/core-feature-pack/common/src/main/resources/modules/system/layers/base/org/wildfly/extension/elytron/jaas-realm/main/module.xml
+++ b/core-feature-pack/common/src/main/resources/modules/system/layers/base/org/wildfly/extension/elytron/jaas-realm/main/module.xml
@@ -33,10 +33,6 @@
         <exclude path="org/wildfly/extension/elytron/JaasCustomSecurityRealmWrapper"/>
     </exports>
 
-    <resources>
-        <artifact name="${org.wildfly.core:wildfly-elytron-integration}"/>
-    </resources>
-
     <dependencies>
         <module name="java.logging"/>
         <module name="java.xml"/>
@@ -48,5 +44,6 @@
         <module name="org.jboss.logmanager"/>
         <module name="org.wildfly.common"/>
         <module name="org.wildfly.security.elytron-private" export="true"/>
+        <module name="org.wildfly.extension.elytron"/>
     </dependencies>
 </module>


soul2zimate:
@ivassile The failure seems related

```
org.jboss.as.test.integration.domain.HostExcludesTestCase
testHostExcludes
Stacktrace
java.lang.AssertionError: The following extensions [org.wildfly.extension.elytron.jaas-realm] have been removed on the current release. Remove them on ExtensionConf.CURRENT object defined in this test. The following extensions [] have been added on the current release. Add them to ExtensionConf.CURRENT object defined in this test.
```

yersan (commented):
Looking at this case, I realised that `org.wildfly.extension.elytron.jaas-realm` is not really an extension, it is just a JBoss Modules module that exposes org/wildfly/extension/elytron/JaasCustomSecurityRealmWrapper.

My naive understanding of why it was implemented this way is that Elytron expects the use of a JBoss Modules module to create a custom realm. @fjuma, correct me if I am wrong here.

In such a case, it seems that naming it starting with `org.wildfly.extension.` was not correct at all, as it makes it to look like a regular WildFly extension instead of a login module that can be used to provide a custom realm. @fjuma, are there any other reasons for naming it with the `org.wildfly.extension.` prefix? Is there a requirement that forces us to keep the elytron integration artifact as a resource instead of a dependency?


@ivassile In any case, `org.wildfly.extension.elytron.jaas-realm` is not an extension so we should remove it from `HostExcludesTestCase` and host exclusions. `HostExcludesTestCase` could be adapted to understand more these cases since what it does is to search for JBoss Modules modules that supplies a `META-INF/services/org.jboss.as.controller.Extension` resource. It assumes if a JBoss Modules module supplies this resource, then the module is supplying an extension, but that's not the case for `org.wildfly.extension.elytron.jaas-realm` module. 

I've just opened https://issues.redhat.com/browse/WFLY-19378 to adress this.

ivassile:
@fjuma The jbeap issue is on EAP 8 Update 3 payload. Can you please review this PR and comment on the @yersan questions? Thanks!

bstansberry:
@yersan @ivassile @fjuma The org.wildfly.extension.elytron.jaas-realm module is jboss.api=deprecated, which means we are a long time away from being able to rename it. So, while I agree that the name is unfortunate and it would be nice to rename it, I think the best we can do is make sure there's an issue to eventually remove the module, and even having that JIRA shouldn't block getting any fix done. 

Removing and renaming are both breaking changes so there's not much point doing two breaking changes when one will suffice.

core-feature-pack/common/src/main/resources/modules/system/layers/base/org/wildfly/extension/elytron/jaas-realm/main/module.xml
```diff
@@ -48,5 +44,6 @@
         <module name="org.jboss.logmanager"/>
         <module name="org.wildfly.common"/>
         <module name="org.wildfly.security.elytron-private" export="true"/>
+        <module name="org.wildfly.extension.elytron"/>
```
yersan, line 47: The most equivalent configuration would be by re-exporting `org.wildfly.extension.elytron` from this module. We would like to want to limit it to the `org.wildfly.extension.elytron` package only, so use the following:

```
<module name="org.wildfly.extension.elytron" export="true">
            <exports>
                <exclude path="org/wildfly/extension/elytron/*"/>
            </exports>
        </module>
```

core-feature-pack/common/src/main/resources/modules/system/layers/base/org/wildfly/extension/elytron/jaas-realm/main/module.xml
```diff
@@ -33,10 +33,6 @@
         <exclude path="org/wildfly/extension/elytron/JaasCustomSecurityRealmWrapper"/>
     </exports>
```
yersan, line 34: This export was not excluding `JaasCustomSecurityRealmWrapper` as seems to be the intention, exports/exclude works only with paths and not with specific class resources, but JaasCustomSecurityRealmWrapper is not under a specific package so we cannot easily exclude it from this module export. We can safely remove it from here.

yersan (changes requested):

yersan:
Hello @ivassile FYI, take a look at Zulip where we have been discussing these points

PR #6015, opened 2024-05-28
[WFCORE-6821] Use a builder to construct the different attributes for the domain mode reload test suite operation

lvydra:
https://issues.redhat.com/browse/WFCORE-6821

### DIFF
--- a/testsuite/shared/src/main/java/org/jboss/as/test/integration/domain/management/util/DomainLifecycleUtil.java
+++ b/testsuite/shared/src/main/java/org/jboss/as/test/integration/domain/management/util/DomainLifecycleUtil.java
@@ -405,43 +405,7 @@ public class DomainLifecycleUtil implements AutoCloseable {
      *
      */
     public void reload(String host) throws IOException, InterruptedException, TimeoutException {
-        reload(host, false, null, null, null);
-    }
-
-    /**
-     * Executes a {@code reload} operation.
-     *
-     * @param host the host to use for the request. Cannot be {@code null}
-     * @param restartServers if {@code true}, servers will be restarted
-     *
-     */
-    public void reload(String host, Boolean restartServers) throws IOException, InterruptedException, TimeoutException {
-        reload(host, false, restartServers, null, null);
-    }
-
-    /**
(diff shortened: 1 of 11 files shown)


testsuite/shared/src/main/java/org/jboss/as/test/integration/domain/management/util/DomainLifecycleUtil.java
```diff
@@ -462,48 +429,106 @@ public void reloadAdminOnly(String host) throws IOException, InterruptedExceptio
...
+
+        Boolean waitForServers = null;
+
+        Long timeout = null;
+
+        /**
+         * Sets the adminOnly
+         *
+         * @param adminOnly if {@code true}, the Host Controller will be reloaded in admin-only mode
+         * @return this ReloadParameters object
```
yersan, line 491: The Java doc would need to be completed to describe the other cases. I mean, if it is true, it gets reloadaed with admin only, but what happends if it is null or false? What is the difference bewtween them?

This applies to all the parameters.

Looking at it now, I think we could even using just primitives and avoid the ambiguity between null/false, but in that case, we have to review all the callers since, for example, for restart servers, the default value at management operation level is `true`. So callers should be changed to invoke the reload with the correct parameters yo keep the semantic.

@lvydra what do you think? Would it be simpler to use primitives? in any case, the Java doc needs fixing, as it is now, it is a bit ambiguous.

yersan (changes requested):

testsuite/shared/src/main/java/org/jboss/as/test/integration/domain/management/util/DomainLifecycleUtil.java
lvydra: Hi @yersan, thanks for the review.

I have updated the Java doc.

Using primitives seems to be more straightforward and it makes the reload method cleaner for sure. Keeping default values on ReloadParameters the same as default values on the management operation level should ensure consistency, if I'm not mistaken.

yersan (approved):

yersan:
Thanks @lvydra

PR #6017, opened 2024-05-29
[WFCORE-6732] WildFlyRunner doesn't call tearDown and completed ServerSetupTasks if a later one fails in setup

gaol:
Issue: https://issues.redhat.com/browse/WFCORE-6732

Not sure if a test case needs to be added and where to add if so, please let me know :)

### DIFF
--- a/testsuite/test-runner/src/main/java/org/wildfly/core/testrunner/WildFlyRunner.java
+++ b/testsuite/test-runner/src/main/java/org/wildfly/core/testrunner/WildFlyRunner.java
@@ -184,17 +184,34 @@ public class WildFlyRunner extends BlockJUnit4ClassRunner {
             Security.insertProviderAt(ELYTRON_PROVIDER, 0);
             providerInstalled = true;
         }
+        RuntimeException setupException = null;
         try {
             if (automaticServerControl) {
                 runSetupTasks();
             }
         } catch (AssumptionViolatedException e) {
             assumptionFailedInServerSetup = e;
+        } catch (RuntimeException e) {
+            setupException = e;
         }
-        super.run(notifier);
+        if (setupException == null) {
+            super.run(notifier);
+        }
 
         if (automaticServerControl) {
-            runTearDownTasks();
+            RuntimeException tearDownException = null;
+            try {
+                runTearDownTasks();
+            } catch (RuntimeException e) {
+                tearDownException = e;
+            } finally {
+                if (setupException != null) {
+                    throw setupException;
+                }
+                if (tearDownException != null) {
+                    throw tearDownException;
+                }
+            }
         }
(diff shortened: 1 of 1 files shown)


testsuite/test-runner/src/main/java/org/wildfly/core/testrunner/WildFlyRunner.java
```diff
@@ -184,34 +184,50 @@ public void testRunFinished(Result result) throws Exception {
...
-                runSetupTasks();
+                setupException = runSetupTasks();
             }
         } catch (AssumptionViolatedException e) {
             assumptionFailedInServerSetup = e;
         }
-        super.run(notifier);
+        if (setupException == null) {
+            super.run(notifier);
+        }
```
yersan, line 197: Before this change, if an AssumptionViolatedException gets thrown, the `super.run(notifier);` was executed. Now, we are introducing a change here. I'm not familiar with JUInit runners, are there any side effects?

testsuite/test-runner/src/main/java/org/wildfly/core/testrunner/WildFlyRunner.java
```diff
@@ -184,34 +184,50 @@ public void testRunFinished(Result result) throws Exception {
...
                 task.setup(controller.getClient());
                 runSetupTasks++;
             } catch (AssumptionViolatedException e) {
                 throw e;
             } catch (Exception e) {
-                throw new RuntimeException(String.format("Could not run setup task '%s'", task), e);
+                return new RuntimeException(String.format("Could not run setup task '%s'", task), e);
             }
         }
+        return null;
```
yersan, line 230: Just out of curiosity, why not continue to control the behaviour by throwing the RuntimeException instead of returning it as a value? At first glance, it looked more natural to me thrown an exception instead of returning it as a value.

testsuite/test-runner/src/main/java/org/wildfly/core/testrunner/WildFlyRunner.java
gaol: IMO, it won't introduce a change here. With the change, the `super.run(notifier)` will also be executed when AssumptionViolatedException was thrown from `runSetupTasks()`, the `setupException` is `null` in that case.

testsuite/test-runner/src/main/java/org/wildfly/core/testrunner/WildFlyRunner.java
gaol: thanks, I think it makes sense to throw the exception , will update it. :)

yersan (approved):

yersan:
Thanks @gaol

PR #6019, opened 2024-05-29
Fix for WFCORE-6069, PermissionsDeploymentTestCase fails intermittently on Windows - Bootable JAR Job

jfdenise:
Issue: https://issues.redhat.com/browse/WFCORE-6069

@jamezp , what do you think of this possible scenario?

### DIFF
--- a/bootable-jar/runtime/src/main/java/org/wildfly/core/jar/runtime/BootableJar.java
+++ b/bootable-jar/runtime/src/main/java/org/wildfly/core/jar/runtime/BootableJar.java
@@ -446,13 +446,14 @@ public final class BootableJar implements ShutdownHandler {
             executor.shutdown();
             try {
                 if (!executor.awaitTermination(environment.getTimeout(), TimeUnit.SECONDS)) {
-                    // For some reason we've timed out. The deletion should likely be executing, but let's force it to
-                    // be safe.
-                    cleaner.cleanup();
+                    // For some reason we've timed out. The deletion should likely be executing.
+                    // We can't start a new cleanup to force it. On Windows we would have the side effect to have 2 cleaner processes to
+                    // be executed, with the risk that a new installation has been installed and the new cleaner cleaning the new installation
+                    log.cleanupTimeout(environment.getTimeout(), environment.getJBossHome());
                 }
-            } catch (IOException | InterruptedException e) {
-                // Possibly already logged, but we should log again to be safe
-                log.failedToStartCleanupProcess(e, environment.getJBossHome());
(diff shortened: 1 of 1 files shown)


bootable-jar/runtime/src/main/java/org/wildfly/core/jar/runtime/BootableJar.java
```diff
@@ -446,13 +446,13 @@ public void run() {
             executor.shutdown();
             try {
                 if (!executor.awaitTermination(environment.getTimeout(), TimeUnit.SECONDS)) {
-                    // For some reason we've timed out. The deletion should likely be executing, but let's force it to
-                    // be safe.
-                    cleaner.cleanup();
+                    // For some reason we've timed out. The deletion should likely be executing.
+                    // We can't start a new cleanup to force it. On Windows we would have the side effect to have 2 cleaner processes to
+                    // be executed, with the risk that a new installation has been installed and the new cleaner cleaning the new installation
+                    log.cleanupTimeout(environment.getTimeout(), environment.getJBossHome());
```
jamezp, line 452: Maybe we should just execute the `cleaner.cleanup()` in a try/catch where we simply ignore the IOException and, like you do here, log the exception.

jfdenise: If we do call cleaner.cleanup, even in a try/catch, it will fork a new process and leave. With very special scheduling, in which the installation dir has been already populated with some content (because the initial cleaner actually cleaned everything), the new process will delete the new installation. I suspect that the random failure is caused by this scenario,

bootable-jar/runtime/src/main/java/org/wildfly/core/jar/runtime/BootableJar.java
```diff
@@ -446,13 +446,13 @@ public void run() {
...
                 }
-            } catch (IOException | InterruptedException e) {
-                // Possibly already logged, but we should log again to be safe
-                log.failedToStartCleanupProcess(e, environment.getJBossHome());
+            } catch (InterruptedException e) {
+                // The task has veen interrupted, leaving
```
jamezp, line 455: I do think we should leave the log message here too, but the above change makes sense as well.

jfdenise: Good! I added it back.

jamezp (approved):

yersan:
Thanks @jfdenise @jamezp

PR #6021, opened 2024-05-29
[WFCORE-3502]: Fix debug options in test-runner.

ehsavoie:
* Excluding the jpda profile surefire parameters from the jmv.args system property so it doesn't interfer with the server args.

Jira: https://issues.redhat.com/browse/WFCORE-3502


yersan (changes requested):

yersan (approved):

yersan:
Thanks @ehsavoie

PR #6033, opened 2024-06-11
[WFCORE-5691][Preview] Allow setting timeout for token introspection

lvydra:
https://issues.redhat.com/browse/EAP7-1856
https://issues.redhat.com/browse/WFCORE-5691
https://issues.redhat.com/browse/ELY-2189

### DIFF
--- a/elytron/src/main/java/org/wildfly/extension/elytron/ElytronSubsystemSchema.java
+++ b/elytron/src/main/java/org/wildfly/extension/elytron/ElytronSubsystemSchema.java
@@ -232,7 +232,9 @@ public enum ElytronSubsystemSchema implements PersistentSubsystemSchema<ElytronS
 
     private void addRealmParser(PersistentResourceXMLDescription.PersistentResourceXMLBuilder builder) {
         RealmParser realmParser = new RealmParser();
-        if (this.since(ElytronSubsystemSchema.VERSION_18_0)) {
+        if (this.since(ElytronSubsystemSchema.VERSION_18_0_PREVIEW)) {
+            builder.addChild(realmParser.realmParserPreview_18_0);
+        } else if (this.since(ElytronSubsystemSchema.VERSION_18_0)) {
             builder.addChild(realmParser.realmParser_18);
         } else if (this.since(ElytronSubsystemSchema.VERSION_16_0)) {
             builder.addChild(realmParser.realmParser_16);
--- a/elytron/src/main/java/org/wildfly/extension/elytron/ElytronSubsystemSchema.java
+++ b/elytron/src/main/java/org/wildfly/extension/elytron/ElytronSubsystemSchema.java
@@ -54,8 +54,9 @@ public enum ElytronSubsystemSchema implements PersistentSubsystemSchema<ElytronS
     VERSION_17_0(17),
     VERSION_18_0(18),
     VERSION_18_0_COMMUNITY(18, Stability.COMMUNITY),
+    VERSION_18_0_PREVIEW(18, Stability.PREVIEW),
     ;
-    static final Map<Stability, ElytronSubsystemSchema> CURRENT = Feature.map(EnumSet.of(VERSION_18_0, VERSION_18_0_COMMUNITY));
+    static final Map<Stability, ElytronSubsystemSchema> CURRENT = Feature.map(EnumSet.of(VERSION_18_0, VERSION_18_0_COMMUNITY, VERSION_18_0_PREVIEW));
 
     private final VersionedNamespace<IntVersion, ElytronSubsystemSchema> namespace;
 
--- a/elytron/src/main/java/org/wildfly/extension/elytron/RealmParser.java
+++ b/elytron/src/main/java/org/wildfly/extension/elytron/RealmParser.java
@@ -112,6 +112,9 @@ class RealmParser {
     private final PersistentResourceXMLDescription tokenRealmParser = builder(PathElement.pathElement(ElytronDescriptionConstants.TOKEN_REALM))
             .addAttributes(TokenRealmDefinition.ATTRIBUTES)
             .build();
+    private final PersistentResourceXMLDescription tokenRealmParserPreview_18_0 = builder(PathElement.pathElement(ElytronDescriptionConstants.TOKEN_REALM))
+            .addAttributes(TokenRealmDefinition.ATTRIBUTES)
+            .build();
     private final PersistentResourceXMLDescription cachingRealmParser = builder(PathElement.pathElement(ElytronDescriptionConstants.CACHING_REALM))
             .addAttributes(CachingRealmDefinition.ATTRIBUTES)
             .build();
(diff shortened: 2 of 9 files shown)


yersan:
Not sure if this would be a necessary fix that would prevent merging this, but in general model and schema bumps go in an independent commit, and on top of that one, any system change. This is up to the Elytron component leads. CC @fjuma

yersan:
@lvydra There is nothing written, but for nondefault stability level, we generally append the stability level at the PR title, e.g "[WFCORE-XYZ][Community] ...."

yersan:
This is likely to conflict with https://github.com/wildfly/wildfly-core/pull/5999

fjuma:
> Not sure if this would be a necessary fix that would prevent merging this, but in general model and schema bumps go in an independent commit, and on top of that one, any system change. This is up to the Elytron component leads. CC @fjuma

Yes, model and schema bumps should be part of a separate commit and should also have a separate WFCORE issue associated with it.

elytron/src/main/java/org/wildfly/extension/elytron/ElytronSubsystemSchema.java
```diff
@@ -54,8 +54,9 @@ public enum ElytronSubsystemSchema implements PersistentSubsystemSchema<ElytronS
     VERSION_17_0(17),
     VERSION_18_0(18),
     VERSION_18_0_COMMUNITY(18, Stability.COMMUNITY),
+    VERSION_19_0_PREVIEW(19, Stability.PREVIEW),
```
pferraro, line 57: There is no reason to increment the numeric version.
The next schema version would just be 18.0:preview.

elytron/src/main/java/org/wildfly/extension/elytron/RealmParser.java
```diff
@@ -272,6 +275,23 @@ class RealmParser {
...
+            .addChild(keyStoreRealmParser)
+            .addChild(propertiesRealmParser_14_0)
+            .addChild(ldapRealmParser)
+            .addChild(filesystemRealmParser_16)
+            .addChild(tokenRealmParserPreview_19_0)
+            .addChild(cachingRealmParser)
+            .addChild(distributedRealmParser_18)
+            .addChild(failoverRealmParser)
+            .addChild(jaasRealmParser)
+            .build();
```
pferraro, line 293: This is not a sustainable way to manage version-specific parsing logic.  Why not add children conditionally based on schema version?

elytron/src/main/java/org/wildfly/extension/elytron/TokenRealmDefinition.java
```diff
@@ -148,6 +149,18 @@ static class JwtValidatorAttributes {
...
+                .setValidator(new IntRangeValidator(0))
+                .setRestartAllServices()
+                .build();
+
+        static final SimpleAttributeDefinition READ_TIMEOUT = new SimpleAttributeDefinitionBuilder(ElytronDescriptionConstants.READ_TIMEOUT, ModelType.INT, true)
+                .setAllowExpression(true)
+                .setValidator(new IntRangeValidator(0))
+                .setRestartAllServices()
+                .build();
+
```
pferraro, line 163: Are these attributes only meant to be registered when the server supports PREVIEW features?  If so, you'll want to specify this.
e.g.
```java
.setStability(Stability.PREVIEW)
```

elytron/src/main/java/org/wildfly/extension/elytron/TokenRealmDefinition.java
```diff
@@ -198,9 +211,21 @@ static class OAuth2IntrospectionValidatorAttributes {
...
+        static final SimpleAttributeDefinition READ_TIMEOUT = new SimpleAttributeDefinitionBuilder(ElytronDescriptionConstants.READ_TIMEOUT, ModelType.INT, true)
+                .setAllowExpression(true)
+                .setValidator(new IntRangeValidator(0))
+                .setRestartAllServices()
+                .build();
```
pferraro, line 224: Same as above.

elytron/src/test/java/org/wildfly/extension/elytron/ElytronMixedStabilitySubsystemParsingTestCase.java
```diff
@@ -49,10 +49,12 @@ private static void mockReadResourceWithValidSubsystemTestFilePaths() {
...
-                    return ModelTestUtils.readResource(getClass(), name.replace("elytron", "legacy-elytron-subsystem"));
-                } else {
+                String previewNamespaceUri = ElytronSubsystemSchema.CURRENT.get(Stability.PREVIEW).getNamespace().getUri();
+                String previewVersion = previewNamespaceUri.substring(previewNamespaceUri.lastIndexOf(':') + 1);
+                if (name.contains(version + ".xml") || name.contains(previewVersion + ".xml")) {
                     return ModelTestUtils.readResource(getClass(), name.replace("elytron", "elytron-subsystem"));
+                } else {
+                    return ModelTestUtils.readResource(getClass(), name.replace("elytron", "legacy-elytron-subsystem"));
                 }
             }
```
pferraro, line 59: I don't understand the purpose of mockReadResourceWithValidSubsystemTestFilePaths().  If the goal is simply to override the default file names of the subsystem test files, would it not be far simpler to just override the method used to generate them?
e.g.
```java
@Override
protected String getSubsystemXmlPathPattern() {
    // Same as super impl, but with "-subsystem-" thrown in
    String pattern = (this.schema.getStability() == Stability.DEFAULT) ? "%1$s-subsystem-%2$d.%3$d.xml" : "%1$s-subsystem-%4$s-%2$d.%3$d.xml";
    // If not a current schema, prefix with "legacy-"
    return !ElytronSubsystemSchema.CURRENT.values().contains(this.schema) ? "legacy-" + pattern : pattern;
}
```

pferraro (changes requested):

PrarthonaPaul:
> > Not sure if this would be a necessary fix that would prevent merging this, but in general model and schema bumps go in an independent commit, and on top of that one, any system change. This is up to the Elytron component leads. CC @fjuma
> 
> Yes, model and schema bumps should be part of a separate commit and should also have a separate WFCORE issue associated with it. Please coordinate with @PrarthonaPaul since both of you will need this.

Here is the commit for the subsystem bump for elytron from community:18.0 to preview:18.0 : https://github.com/wildfly/wildfly-core/pull/5999/commits/22825d1cbf5632934f32026283dbd11e2e2f2271

darranl:
Just one small change on the comment from @PrarthonaPaul - @lvydra please rebase your changes on top of that commit so you both preserve the same SHA - where we use a common commit we are trying to ensure it happens only once.

darranl:
Also the Jira issue and this PR seem to be referencing the Community stability level but the code seems to be targeting Preview.

I would suggest we continue with Preview as we can then promote to Community or Default in a later step.

lvydra:
Thanks @PrarthonaPaul, I have rebased my changes on top of your subsystem bump commit.

bstansberry:
@pferraro Does this look ok now?

darranl:
I will take this one and get this finished off.

PR #6045, opened 2024-06-21
WFCORE-6807 Prune deprecations from wildfly-service and wildfly-subsystem

pferraro:
https://issues.redhat.com/browse/WFCORE-6807

### DIFF
--- a/service/src/main/java/org/wildfly/service/Installer.java
+++ b/service/src/main/java/org/wildfly/service/Installer.java
@@ -88,34 +88,15 @@ public interface Installer<ST extends ServiceTarget> {
     }
 
     /**
-     * Implemented by builds with asynchronous service support.
-     * @param <B> the builder type
-     */
-    @Deprecated(forRemoval = true)
-    interface AsyncBuilder<B> {
-        /**
-         * Indicates that the installed service should start and, if a stop task is configured, stop asynchronously.
-         * @return a reference to this builder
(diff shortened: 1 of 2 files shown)


darranl:
As this module is relatively new it is probably too new that these removals will cause issues but how do we determine when it is safe to remove deprecations like this?

Subsystems and feature packs can be developed outside of the WildFly and WildFly Core repos and may not be keeping up to date at the same cadence as WildFly.

pferraro:
> As this module is relatively new it is probably too new that these removals will cause issues but how do we determine when it is safe to remove deprecations like this?
> 
> Subsystems and feature packs can be developed outside of the WildFly and WildFly Core repos and may not be keeping up to date at the same cadence as WildFly.

FWIW, the methods being removed here were introduced in 23.0.0.Final and deprecated for removal in 24.0.0.Final.
There is only one external repo of which I am aware that uses wildfly-service/subsystem (https://github.com/wildfly-extras/mvc-krazo-feature-pack), and it is unaffected.
I think it would be unwise to delay these removals too long lest they end up in a downstream release.

yersan (approved):
I agreed with Paul on this, so I approved from my side

@darranl Do you still have concerns with these removals?

darranl (approved):

yersan:
Thanks @pferraro @darranl

PR #6048, opened 2024-06-25
WFCORE-6866 Create ServiceDescriptor for management executor

pferraro:
https://issues.redhat.com/browse/WFCORE-6866

If someone has a better idea of where to locate this service descriptor, let me know.


yersan:
> If someone has a better idea of where to locate this service descriptor, let me know.

I have no better suggestions than the Capabilities class. I understand the natural choice would be to be declared on the service value class, but for this one, `Executor`, is impossible.

yersan (approved):
In any case, I'll wait some days more in case there is any other suggestion

bstansberry (approved):

bstansberry:
I don't have any better locations either.

bstansberry:
Thanks @pferraro and @yersan

PR #6049, opened 2024-06-26
[WFCORE-6750] Iterate resources rather than writing to a shared index…

kabir:
….html

Follow-up on 
https://issues.redhat.com/browse/WFCORE-6750


bstansberry (approved):

bstansberry:
Thanks @kabir

PR #6051, opened 2024-06-27
WFCORE-6812 Allow validating multiple XML files

parsharma:
Issue: https://issues.redhat.com/browse/WFCORE-6812

### DIFF
--- a/subsystem-test/framework/src/main/java/org/jboss/as/subsystem/test/AbstractSubsystemBaseTest.java
+++ b/subsystem-test/framework/src/main/java/org/jboss/as/subsystem/test/AbstractSubsystemBaseTest.java
@@ -6,11 +6,13 @@
 package org.jboss.as.subsystem.test;
 
 import java.io.IOException;
+import java.io.UncheckedIOException;
 import java.util.Collections;
 import java.util.Comparator;
 import java.util.List;
 import java.util.Properties;
 import java.util.Set;
+import java.util.stream.Collectors;
 
 import org.jboss.as.controller.Extension;
 import org.jboss.as.controller.PathAddress;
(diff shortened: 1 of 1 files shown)


yersan (commented):
Changes look good to me, but what this PR does is create a new method to load multiple XML files as a single one.

The Jira description does not clarify too much, at least to me, about what is the final goal here. Is it to allow the validation of multiple XML files against a common XSD schema? Notice this class is used for subsystems, so I would understand the intention here is to concatenate multiple XML subsystem files into a single one, and if that's the intention, the final one will have multiple xmls namespaces, so, I don't get how this is expected to work.

@parsharma / @michpetrov Could you describe the Jira intention better and how this method would allow validating multiple XML files?


My only bet is that this is here to allow later to do the following:

`SchemaValidator.validateXML(getSubsystemXml(List<String>), schemaPath, getResolvedProperties());`

And, if that's the case, exposing a method that does this validation is the task to do here, otherwise this PR is to allow composing multiple XML files in a single one. And also, I cannot figure out what list of XML would be passed to this method, to do the above the resulting XML file should only have a single XML NS ... correct me if I am wrong, I think I need to refresh my XML knowledge.

michpetrov (changes requested):
This won't work unfortunately, if you concatenate multiple files it's no longer a valid XML and the parser will fail. Bear in mind that `getSubsystemXml()` is used in other methods, so we can't change its output.

I don't think you can get  around processing the contents of each file separately.

michpetrov:
@yersan The goal is to validate multiple files against a single schema, yes.

I'm assuming the expected usage was to do
```
    return getSubsystemXml(List.of("a.xml", "b.xml", "c.xml"));  // previously return readResource("a.xml");
}
```
which is kinda what I had in mind in terms of what's required to do in a given subsystem test. But putting it all into a single file doesn't work for the parser.

yersan:
thanks @michpetrov, then do you think we should close the Jira as won't do?

Otherwise, I would need a real use case since I don't know yet what kind of subsystem XML files we would need to compose and validate against a single xsd schema. I am also reluctant to add any piece of code if it is not going to be used yet, adding something for the future for someone who would need it doesn't look good, better let's add it when there is an immediate usage need.

michpetrov:
@yersan there are subsystems like datasources that use multiple files for testing - https://github.com/wildfly/wildfly/tree/main/connector/src/test/resources/org/jboss/as/connector/subsystems/datasources

currently only `datasources-minimal.xml` is being checked but `datasources-full.xml` is invalid. If we want to ensure everything is valid we would have to create a separate test class for each of the files, so I want a better way.

PR #6055, opened 2024-07-01
[WFCORE-6873] Remoting: use subsystem type endpoint for outbound comm…

tadamski:
…unication

https://issues.redhat.com/browse/WFCORE-6873

### DIFF
--- a/remoting/subsystem/src/main/java/org/jboss/as/remoting/EndpointService.java
+++ b/remoting/subsystem/src/main/java/org/jboss/as/remoting/EndpointService.java
@@ -29,6 +29,7 @@ public class EndpointService implements Service {
 
     protected final String endpointName;
     protected volatile Endpoint endpoint;
+    private final EndpointType type;
     protected final OptionMap optionMap;
     private final Consumer<Endpoint> endpointConsumer;
     private final Supplier<XnioWorker> workerSupplier;
@@ -40,6 +41,7 @@ public class EndpointService implements Service {
             nodeName = "remote";
         }
         endpointName = type == EndpointType.SUBSYSTEM ? nodeName : nodeName + ":" + type;
+        this.type = type;
         this.optionMap = optionMap;
     }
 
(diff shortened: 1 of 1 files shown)


yersan:
Hello @fl4via / @ropalka Could anyone of you review this? I am not familiar with the ContextManager and what this `setGlobalDefault` implies, thanks

baranowb:
Generally would be good to point why this change will affect outbound comm.(given comment above alone) @tadamski  ^^

fl4via (approved):

ropalka (approved):

bstansberry:
@tadamski @fl4via How is this tested?

rachmatowicz:
It seems to me that there might be a problem here. The Contextual<Endpoint> is used by all client code (naming client, EJB client, Transaction client) to initiate connections from the client to a server. The wildfly-config.xml file is used to configure this contextual as well as the others that are available to the client side code (EJBClientContext, Discovery, AuthenticationContext, and others). Generally, these various Contextuals are initialized statically by parsing the wildfly-config.xml file and setting a global default in the context manager, whether the client code is running on a standalone client or within a deployment on a server. This change will cause the Contextual<Endpoint> to point to the Endpoint defined in the Remoting subsystem, and not the usual Contextual<Endpoint>. The problems I see are:
(1) now to configure the client-side connection properties for any Wildfly client code, we cannot use the wildfly-client.xml file and need to configure the client-side properties in the Remoting subsystem Endpoint configuration, which will have to be the same for all clients
(2)

ropalka:
@dmlloyd WDYT about this PR proposal?

dmlloyd:
Generally I'm in favor of this idea. It's what I always wanted to do but didn't quite manage at the time.

To address (partially) @rachmatowicz's comments:

It is true that XNIO is more centrally configured compared to Remoting. This actually is by design though: it was always intended that one XNIO instance would support the needs of all networking I/O, and this has guided the design of XNIO from the beginning. So XNIO remains a foundational piece, whereas other things (like Remoting) are meant to be added atop it.

As for the loss of per-deployment client configuration, I agree this may end up being a problem in practice. But the solution to this problem is not general (e.g. to replicate just the Endpoint). The better approach would be to determine what configuration makes sense for individual deployments based on the underlying abstractions which are available and the needs of the user. Exposing every possible knob has always been a weakness of this architecture, not a strength. I'd say it's most prudent to address specific problems that are expected on a one-by-one basis, and to handle other problems as they arise, *if* they arise.

tadamski:
I'm putting the issue on hold, because of the ongoing discussion and the need to add tests before the PR is merged.

PR #6057, opened 2024-07-03
[WFCORE-6874] Fix logging.properties configuration file when apply se…

yersan:
…rver phase is used by the installer command

Jira issue: https://issues.redhat.com/browse/WFCORE-6874

@spyrkob Could you take also a look? thanks!

### DIFF
--- a/core-feature-pack/common/src/main/resources/content/bin/domain.ps1
+++ b/core-feature-pack/common/src/main/resources/content/bin/domain.ps1
@@ -113,4 +113,4 @@ if ($SERVER_OPTS -ne $null){
 $backgroundProcess = Get-Env LAUNCH_JBOSS_IN_BACKGROUND 'false'
 $runInBackGround = $global:RUN_IN_BACKGROUND -or ($backgroundProcess -eq 'true')
 
-Start-WildFly-Process -programArguments $PROG_ARGS -runInBackground $runInBackGround
+Start-WildFly-Process -programArguments $PROG_ARGS -runInBackground $runInBackGround -logFileProperties "$JBOSS_CONFIG_DIR\logging.properties" -logFile "$JBOSS_LOG_DIR\host-controller.log"
--- a/core-feature-pack/common/src/main/resources/content/bin/installation-manager.bat
+++ b/core-feature-pack/common/src/main/resources/content/bin/installation-manager.bat
@@ -1,4 +1,4 @@
-setlocal
+setlocal DisableDelayedExpansion
 rem This script is only for internal usage and should not be invoked directly by users from the command line.
 rem This script launches the operation to apply a candidate server installation to update or revert.
 rem The server JVM writes the required values into the installation-manager.properties file by using InstMgrCandidateStatus.java
@@ -11,14 +11,20 @@ if "%INST_MGR_SCRIPT_DEBUG%"=="true" (
 
 set INSTALLATION_HOME=%~1
 set INST_MGR_LOG_PROPERTIES=%~2
+set INST_MGR_LOG_FILE=%~3
+
+rem Prepare variables for log traces
+set "LOG_NAME=[management-cli-installer]"
 
 rem For security, reset the environment variables first
 set INST_MGR_COMMAND=
 set INST_MGR_STATUS=
 
-set PROPS_FILE=%INSTALLATION_HOME%\bin\installation-manager.properties
+echo %date% %time% INFO %LOG_NAME% - Executing Management CLI Installer script.
+
+set "PROPS_FILE=%INSTALLATION_HOME%\bin\installation-manager.properties"
 if not exist "%PROPS_FILE%" (
-    echo ERROR: Installation Manager properties file not found at %PROPS_FILE%.
+    echo %date% %time% ERROR %LOG_NAME% - Installation Manager properties file not found at %PROPS_FILE%.
 
     goto EOF
 )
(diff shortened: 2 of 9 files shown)


core-feature-pack/common/src/main/resources/content/bin/installation-manager.bat
```diff
@@ -51,24 +58,30 @@ set "INST_MGR_COMMAND=!INST_MGR_COMMAND:\:=:!"
 set "INST_MGR_COMMAND=!INST_MGR_COMMAND:\\=\!"
 setlocal DisableDelayedExpansion
 
-set JAVA_OPTS=-Dlogging.configuration=file:"%INST_MGR_LOG_PROPERTIES%" %JAVA_OPTS%
+set JAVA_OPTS=-Dlogging.configuration=file:"%INST_MGR_LOG_PROPERTIES%" -Dorg.jboss.boot.log.file="%INST_MGR_LOG_FILE%" -Dorg.wildfly.prospero.log.file %JAVA_OPTS%
```
spyrkob, line 61: @yersan shouldn't the `org.wildfly.prospero.log.file` be set to `INST_MGR_LOG_FILE` ?

core-feature-pack/common/src/main/resources/content/bin/domain.ps1
```diff
@@ -113,4 +113,4 @@ if ($SERVER_OPTS -ne $null){
 $backgroundProcess = Get-Env LAUNCH_JBOSS_IN_BACKGROUND 'false'
 $runInBackGround = $global:RUN_IN_BACKGROUND -or ($backgroundProcess -eq 'true')
 
-Start-WildFly-Process -programArguments $PROG_ARGS -runInBackground $runInBackGround
+Start-WildFly-Process -programArguments $PROG_ARGS -runInBackground $runInBackGround -logFileProperties "$JBOSS_CONFIG_DIR\logging.properties" -logFile "$JBOSS_LOG_DIR\host-controller.log"
```
spyrkob, line 116: Does this need to be added to the standalone.ps1 as well?

core-feature-pack/common/src/main/resources/content/bin/installation-manager.bat
yersan: @spyrkob It could, but the value of `org.wildfly.prospero.log.file` is only used by Prospero logging configuration file and here we are using the server logging configuration file available at `JBOSS_CONFIG_DIR\logging.properties`

This `org.wildfly.prospero.log.file` empty initialization here is just a trick to avoid that `prospero.sh[bat,ps1]` do any configuration that could affect the logging framework. When this property is initialized at the JAVA_OPTS, `prospero.sh[bat,ps1]` will ignore it and won't add any additional JAVA_OPTS configuration, see: https://github.com/wildfly-extras/prospero/blob/main/dist/common/src/main/resources/packages/org.jboss.prospero/content/bin/prospero.sh#L94 

As you said, it could be initialized to "%INST_MGR_LOG_FILE%", but really, it is an unknown property from the point of view of `JBOSS_CONFIG_DIR\logging.properties`. Maybe it is better to give it a value, but it is not used at all.

core-feature-pack/common/src/main/resources/content/bin/domain.ps1
yersan: @spyrkob Not really, both domain.ps1 and standalone.ps1 use a common method to kick off the server. This common method gets the argument with default values when standalone.ps1 invokes it. So, we only overwrite the defaults when using domain.ps1

See: https://github.com/wildfly/wildfly-core/blob/3aa4a9ee0720e1a4c07dfd1b5b9fd1def0355a00/core-feature-pack/common/src/main/resources/content/bin/common.ps1#L351-L353

spyrkob: ah I missed that, thanks!

spyrkob (approved):
LGTM

core-feature-pack/common/src/main/resources/content/bin/installation-manager.sh
```diff
@@ -8,15 +8,25 @@ fi
...
 
 # For security, reset the environment variables first
 unset INST_MGR_COMMAND
 unset INST_MGR_STATUS
 
+TIMESTAMP=$(date "+%Y-%m-%d %H:%M:%S,%3N")
+LOG_NAME="[management-cli-installer]"
+
+log() {
+    echo "$TIMESTAMP ${1} $LOG_NAME - ${2}"
```
jamezp, line 21: This will log with the same timestamp for each entry. The `TIMESTAMP` variable should probably be removed and the `$(date "+%Y-%m-%d %H:%M:%S,%3N")` should be explicitly used here.

core-feature-pack/common/src/main/resources/content/bin/installation-manager.bat
```diff
@@ -11,14 +11,21 @@ if "%INST_MGR_SCRIPT_DEBUG%"=="true" (
 
 set INSTALLATION_HOME=%~1
 set INST_MGR_LOG_PROPERTIES=%~2
+set INST_MGR_LOG_FILE=%~3
+
+rem Prepare variables for log traces
+set "TIMESTAMP=%date% %time%"
```
jamezp, line 17: I think this sets the `TIMESTAMP` to the same value each time it is used. IIRC one way around that is by using the `DelayedExpansion`, however it's explicitly enabled here. I don't know if we can use a batch call here or not, but that might also be an option.

core-feature-pack/common/src/main/resources/content/bin/installation-manager.ps1
```diff
@@ -4,33 +4,23 @@
...
-
-    # Check if the path starts with double backslashes and contains at least one additional character
-    if ($inputPath -match '^\\\\[^\\]+\\.*') {
-        $removedBackslash = $inputPath.Substring(2)
-        return "\\?\UNC\$removedBackslash"
-    } else {
-        return "\\?\$inputPath"
-    }
-}
+$TimeStamp = Get-Date -Format "yyyy-MM-dd HH:mm:ss"
```
jamezp, line 11: Same here as above, won't this always be set to the same timestamp?

core-feature-pack/common/src/main/resources/content/bin/installation-manager.sh
jamezp: Just a thought, but would `printf` work here too? Something like:

```bash
printf "%s %5s ${LOG_NAME} - %s\n" "$(date "+%Y-%m-%d %H:%M:%S,%3N")" "${1}" "${2}"
```

yersan: @jamezp I guess it should work, my doubt comes from whether `printf` is available on all the shells, would it be? I tested it is available in dash in my local environment but I have not found how to validate it will be in all shells

core-feature-pack/common/src/main/resources/content/bin/installation-manager.bat
yersan: Right, it was a mistake, fixed

core-feature-pack/common/src/main/resources/content/bin/installation-manager.ps1
yersan: Fixed

jamezp (approved):
Thanks @yersan. LGTM.

yersan:
Thanks @spyrkob @jamezp !

PR #6061, opened 2024-07-08
[WFCORE-6887] Introducing new phase STRUCTURE_NAMING_JDK_DEPENDENCIES

ropalka:
https://issues.redhat.com/browse/WFCORE-6887

### DIFF
--- a/server/src/main/java/org/jboss/as/server/deployment/Phase.java
+++ b/server/src/main/java/org/jboss/as/server/deployment/Phase.java
@@ -234,6 +234,7 @@ public enum Phase {
     public static final int STRUCTURE_GLOBAL_MODULES                    = 0x1E00;
     public static final int STRUCTURE_GLOBAL_DIRECTORIES                = 0x1E01;
     public static final int STRUCTURE_NAMING_EXTERNAL_CONTEXTS          = 0x1F00;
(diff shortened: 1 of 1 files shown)


server/src/main/java/org/jboss/as/server/deployment/Phase.java
```diff
@@ -234,6 +234,7 @@ public AttachmentKey<?> getPhaseKey() {
     public static final int STRUCTURE_GLOBAL_MODULES                    = 0x1E00;
     public static final int STRUCTURE_GLOBAL_DIRECTORIES                = 0x1E01;
     public static final int STRUCTURE_NAMING_EXTERNAL_CONTEXTS          = 0x1F00;
+    public static final int STRUCTURE_NAMING_JDK_DEPENDENCIES           = 0x1F01;
```
darranl, line 237: This should probably be between the existing two constants rather than incrementing the prior constant by one as this prevents inserting another constant between STRUCTURE_NAMING_EXTERNAL_CONTEXT and STRUCTURE_NAMING_JDK_DEPENDENCIES if the need arose in the future.

darranl (changes requested):

server/src/main/java/org/jboss/as/server/deployment/Phase.java
ropalka: OK

darranl (approved):

PR #6062, opened 2024-07-08
[WFCORE-6888] Upgrade WildFly Elytron to 2.5.0.Final

fjuma:
https://issues.redhat.com/browse/WFCORE-6888

        Release Notes - WildFly Elytron - Version 2.5.0.Final
                                                                
<h2>        Task
</h2>
<ul>
<li>[<a href='https://issues.redhat.com/browse/ELY-2776'>ELY-2776</a>] -         Add tests to the OIDC testsuite to test for cases where request/request_uri parameters are not supported by the OpenID provider
</li>
</ul>
                
<h2>        Release
</h2>
<ul>
<li>[<a href='https://issues.redhat.com/browse/ELY-2782'>ELY-2782</a>] -         Release Elytron 2.5.0.Final
</li>
</ul>
                                                                                                                                                    
<h2>        Enhancement
</h2>
<ul>
<li>[<a href='https://issues.redhat.com/browse/ELY-2774'>ELY-2774</a>] -         Move CAGenerationTool to PKCS#12
</li>
</ul>

PR #6064, opened 2024-07-09
[WFCORE-6869] Get the KeyStore File and type from the Defined id entities for the OpenSslTlsTestCase

darranl:
https://issues.redhat.com/browse/WFCORE-6869

Also incorporates https://github.com/wildfly/wildfly-core/pull/6062


yersan (approved):

yersan:
Thanks @darranl

PR #6066, opened 2024-07-10
[WFCORE-6755] Move the org.wildfly.security:wildfly-elytron-dynamic-ssl artifact into its own module

yersan:
Supersedes https://github.com/wildfly/wildfly-core/pull/5947

Jira issue: https://issues.redhat.com/browse/WFCORE-6755

See https://wildfly.zulipchat.com/#narrow/stream/174184-wildfly-developers/topic/Layers.20and.20optional.20packages.

This PR includes #5947 but removes the community stability property configuration from `org.wildfly.security.elytron-dynamic-ssl` JBoss Modules module and adds its package as required for the Elytron layer. Until getting a better approach for this case, this will fix the issue reported by the test case where this module was not provisioned when using Elytron Galleon Layer and ensure we can use `default` stability level out of the box when the server is built targetting this stability level.

### DIFF
--- a/core-feature-pack/common/src/main/resources/modules/system/layers/base/org/wildfly/security/elytron-base/main/module.xml
+++ b/core-feature-pack/common/src/main/resources/modules/system/layers/base/org/wildfly/security/elytron-base/main/module.xml
@@ -34,7 +34,6 @@
         <artifact name="${org.wildfly.security:wildfly-elytron-credential-source-impl}"/>
         <artifact name="${org.wildfly.security:wildfly-elytron-credential-store}"/>
         <artifact name="${org.wildfly.security:wildfly-elytron-digest}"/>
-        <artifact name="${org.wildfly.security:wildfly-elytron-dynamic-ssl}"/>
         <artifact name="${org.wildfly.security:wildfly-elytron-encryption}"/>
         <artifact name="${org.wildfly.security:wildfly-elytron-http}"/>
         <artifact name="${org.wildfly.security:wildfly-elytron-http-basic}"/>
@@ -112,5 +111,6 @@
         modules use the parser, they need to have visibility to this module.
         -->
         <module name="org.wildfly.client.config" export="true"/>
+        <module name="org.wildfly.security.elytron-dynamic-ssl" export="true" optional="true"/>
     </dependencies>
 </module>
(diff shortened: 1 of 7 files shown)


yersan:
Hello @fjuma / @darranl , a friendly reminder that I requested any of you to review this one, thanks!

yersan:
@fjuma / @darranl, sorry for the noise for get my last comment, I have to review the current status first since I lost track of this one, I'm going to remove you from the review until this one gets refreshed

darranl:
@yersan FYI I will be working on a PR to promote this to default soon, although would be good to know we could solve this situation if it comes up again so I think still worth proceeding for now.

darranl:
@yersan are we proceeding with this one?

yersan:
@darranl I want to back to this once we have passed the feature freeze, I plan to take a closer look next week

yersan:
@darranl kicked it off again, I've added `valid-for-stability` to the package dependency for Elytron layer, does mean, the package will not be provisioned for an stability level with higher stability than "community"

yersan:
@darranl do you want to take a look? I am fine with merging this as it is

yersan:
@darranl friendly reminder, when you have a chance, take a look at this one and approve from your side, thanks!

darranl:
@yersan we may need to wait on a full solution, at this stage this module must not be provisioned at lower stability levels.

yersan:
> we may need to wait on a full solution, at this stage this module must not be provisioned at lower stability levels.

@darranl the `valid-for-stability` to the package dependency on Elytron layer does that

yersan:
> we may need to wait on a full solution, at this stage this module must not be provisioned at lower stability levels.

Also, "lower stability", I assume you are talking about higher stability levels (e.g default)

darranl:
Sorry yes "higher"

yersan:
@darranl ok, and then, since we are using `valid-for-stability`, would you agree to resolve this as it is now?

darranl:
@yersan What we need is a generic solution so that the module is correctly provisioned at the various stability levels and not provisioned when we only support default - do we have that today?

yersan:
@darranl yes, we are using the "valid-for-stability" attribute at a package level to decide when we want this packaged provisioned, so we are not provisioning it for any stability level higher than community.  At conversion time branches we can add a simple maven verify configuration to verify this module is not provisioned so we be more confident about this. I can add that after merging this and rebasing the conversion.

darranl:
@yersan so both commits on this PR complete that?

yersan:
@darranl just a friendly remainder about this, I would like to release it the sooner the better since it affects the conversion branches, indeed we are already using it there, so if you need to do any manual verification to approve this, you can use the conversion branches

yersan:
@darranl Do you have any questions regarding this PR? Do you plan to review it?

darranl (approved):

yersan:
Thanks @Skyllarr and @darranl

PR #6070, opened 2024-07-11
[WFCORE-6893] Add information to product-info

spyrkob:
Issue: https://issues.redhat.com/browse/WFCORE-6893

Note: Setting as draft, because it requires https://github.com/wildfly-extras/wildfly-installation-manager-api/pull/21 
Related: https://github.com/wildfly-extras/prospero/pull/720


spyrkob:
@yersan could you take a look at this? I created a hidden/private operation handler for the installation manager history that is called from both the `:product-info` and `installer history` operation handlers.

installation-manager/src/main/java/org/wildfly/core/instmgr/InstMgrHistoryHandler.java
```diff
@@ -38,6 +43,12 @@ public class InstMgrHistoryHandler extends InstMgrOperationStepHandler {
 
     @Override
     public void execute(OperationContext context, ModelNode operation) throws OperationFailedException {
+        System.out.println("!! InstMgrHistoryRuntimeHandler");
+        System.out.println(context.getCurrentStage());
+        PathAddress installerAddress = PathAddress.pathAddress(PathElement.pathElement(CORE_SERVICE, "installer"));
+        OperationEntry installerOpEntry = context.getRootResourceRegistration().getOperationEntry(installerAddress, "history-runtime");
+        context.addStep(installerOpEntry.getOperationHandler(), OperationContext.Stage.RUNTIME);
```
yersan, line 50: We have to find a better approach that does not require splitting the history handle into "history-runtime" and "history", both handlers do the same, in addition, the InstallationReportHandler.java host version uses history and the server version the history-handler. I'll take a closed look to find a better approach.

spyrkob:
@yersan should Il keep this as draft until the installation-manager 1.0.3.Final is out and update it in this PR?

yersan:
@spyrkob Ideally yes, we can release installation-manager once we test this PR together with Prospero. Then we bump it on wildfly core and this would be rebased so we can proceed with the merge. We would need to test the operation when Prospero (or the installation manager test version) gets provisioned

yersan:
@spyrkob  Talking about testing strategy, this is what we have so far:

* Management Operation unit tests: https://github.com/wildfly/wildfly-core/blob/main/installation-manager/src/test/java/org/wildfly/core/instmgr/InstMgrResourceTestCase.java 
* High-level commands using domain mode. See https://github.com/wildfly/wildfly-core/blob/main/testsuite/domain/src/test/java/org/jboss/as/test/integration/domain/installationmanager/InstallationManagerIntegrationTestCase.java#L104-L110, we create a specific module to make the mock available in the server. We could do the same for a standalone instance so we can test the `:product-info` management operation on both standalone and Domain Mode.

Share your feedback if you see a better testing strategy, real provisioning with Prospero for testing seems more tricky since this is only WildFly Core and we do not want to be tight to one Prospero version, here we are only testing until the Prospero API invocation.

spyrkob:
@yersan I added tests to check that the correct information is added to the product info and installer history. I based the integration tests for standalone mode on the existing domain tests. I'm not sure if there's a better way to avoid the code duplication between the two

yersan:
> I'm not sure if there's a better way to avoid code duplication between the two

@spyrkob I think it could have been enough by just adding the `:product-info` operation test to the standalone version. The code path for domain mode and standalone is the same, the only difference is that the standalone address is `/core-service=installer` instead of `/host=name/core-service=installer`, but server handlers are the same, CLI handlers just add the host name.

It is good to get both covered though, but if we do it, I guess we should remove the duplication (which can be done later). We can use this directory https://github.com/wildfly/wildfly-core/tree/main/testsuite/shared/src/main/java/org/wildfly/test/installationmanager to place a base class that accepts a list of host names and inherit both test claasses from that base class, the domain one will add the hosts to the test methods.

It is ok to me if this is done after this fix, I leave it up to you. If not we can create a Jira to do it later.

server/src/main/java/org/jboss/as/server/operations/AbstractInstallationReporter.java
```diff
@@ -310,11 +316,23 @@ String getInstallationDir() {
...
-            if (patchingInfo.isDefined()) {
-                List<ModelNode> result = Operations.readResult(patchingInfo).asList();
-                for (ModelNode patchAtt : result) {
-                    if (patchAtt.has("applied-at")) {
-                        return patchAtt.get("applied-at").asString();
+            if (installerInfo.get("result").isDefined()) {
+                List<ModelNode> result = Operations.readResult(installerInfo).asList();
+                for (ModelNode installerAtt : result) {
+                    if (installerAtt.has("timestamp")) {
+                        return installerAtt.get("timestamp").asString();
```
yersan, line 323: We should keep the format given by the legacy tool:

https://github.com/wildfly/wildfly-core/blob/69e4d2a6856d4dd909ac02400a77a1576ef4e667/patching/src/main/java/org/jboss/as/patching/runner/PatchUtils.java#L200

yersan (changes requested):
Looks good to me, there are two minor details, the one about test duplicity, up to you can be done later, and the minor about the date format of the timestamp.

I'm going to merge and release the one of the installation manager API, to get it later available here

server/src/main/java/org/jboss/as/server/operations/AbstractInstallationReporter.java
spyrkob: OK that probably should happen in [InstMgrHistoryHandler](https://github.com/wildfly/wildfly-core/blob/main/installation-manager/src/main/java/org/wildfly/core/instmgr/InstMgrHistoryHandler.java#L54), in the Reporter it's too late - we already have a text value

yersan:
@spyrkob Reviewing your PR I found that there is an issue with the installation manager integration tests running on Windows. The failures here are not related to your PR. I opened https://issues.redhat.com/browse/WFCORE-7047 to fix them

spyrkob:
@yersan I'm happy to do either one, whatever is easier for you/makes more sense from process perspective

yersan:
@spyrkob Since there were no strong preferences, I've just merged https://github.com/wildfly/wildfly-core/pull/6240, you can rebase and incorporate the two minimal changes done in such a PR on the current one and the failures in Windows should disappear

spyrkob:
@yersan I think the error in the Windows test is not related to this change, but I don't have the rights to rerun the test

yersan:
@spyrkob yes, unrelated. In any case, anyone with a user different from anonymous in ci.wildfly.org can re-trigger a job.
I'll do a final review for this one today

PR #6071, opened 2024-07-15
[WFCORE-6895] Add a grace period to Management CLI Installer Windows …

yersan:
…script

Jira issue: https://issues.redhat.com/browse/WFCORE-6895

This just adds a configurable grace period to help prevent locking file errors on Windows when using Management CLI Installer

### DIFF
--- a/cli/src/main/java/org/jboss/as/cli/handlers/ShutdownHandler.java
+++ b/cli/src/main/java/org/jboss/as/cli/handlers/ShutdownHandler.java
@@ -7,7 +7,6 @@ package org.jboss.as.cli.handlers;
 
 import java.io.IOException;
 import java.util.Collection;
-import java.util.concurrent.TimeUnit;
 import java.util.concurrent.atomic.AtomicReference;
 
 import org.jboss.as.cli.AwaiterModelControllerClient;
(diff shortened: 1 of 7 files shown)


core-feature-pack/common/src/main/resources/content/bin/domain.conf.ps1
```diff
@@ -89,3 +89,6 @@ if ($HOST_CONTROLLER_JAVA_OPTS -eq $null) {
 # Sample JPDA settings for shared memory debugging
 #PROCESS_CONTROLLER_JAVA_OPTS="$PROCESS_CONTROLLER_JAVA_OPTS -agentlib:jdwp=transport=dt_shmem,server=y,suspend=n,address=jboss"
 #HOST_CONTROLLER_JAVA_OPTS="$HOST_CONTROLLER_JAVA_OPTS -agentlib:jdwp=transport=dt_shmem,server=y,suspend=n,address=jboss"
```
spyrkob, line 91: Not connected to this PR, but should those assignments start with `$sign` (e.g. `$HOST_CONTROLLER_JAVA_OPTS`)?

cli/src/main/java/org/jboss/as/cli/handlers/ShutdownHandler.java
```diff
@@ -177,11 +176,6 @@ protected void doHandle(CommandContext ctx) throws CommandLineException {
 
             if(isLocalClient) {
                 ctx.printLine("The JBoss CLI session will be closed automatically to allow the server be updated. Once the server has been restarted, you can relaunch the JBoss CLI session.", false);
-                try {
```
spyrkob, line 180: The wait in the startup scripts only applies to Windows scripts, but this wait happened also on Linux - is that OK to remove it?

spyrkob (approved):
Couple of questions below, otherwise LGTM

cli/src/main/java/org/jboss/as/cli/handlers/ShutdownHandler.java
yersan: @spyrkob thanks for commenting on it. It does nothing, it just waits a bit to have a better UX; you execute the command and see how it waits a bit with the message printed, that's all. It is safe to close it immediately.

core-feature-pack/common/src/main/resources/content/bin/domain.conf.ps1
yersan: Good catch!, variables in Powershell are defined starting the with $, I'll append a commit

yersan:
Thanks @spyrkob

PR #6084, opened 2024-07-18
[WFCORE-6907] Remove deprecated SimpleResourceDefinition.registerAddO…

bstansberry:
…peration variant

https://issues.redhat.com/browse/WFCORE-6907

### DIFF
--- a/controller/src/main/java/org/jboss/as/controller/SimpleResourceDefinition.java
+++ b/controller/src/main/java/org/jboss/as/controller/SimpleResourceDefinition.java
@@ -242,19 +242,6 @@ public class SimpleResourceDefinition extends ResourceDefinition.MinimalResource
     }
 
     /**
-     * Registers add operation
-     *
-     * @param registration resource on which to register
-     * @param handler      operation handler to register
-     * @param flags        with flags
-     * @deprecated Redundant with {@link #registerAddOperation(ManagementResourceRegistration, OperationStepHandler, org.jboss.as.controller.registry.OperationEntry.Flag...)
-     */
-    @Deprecated(forRemoval = true)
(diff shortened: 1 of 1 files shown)


jamezp:
I manually re-ran the security manager test as I cannot reproduce the failure.

jamezp:
The failure is now different.

controller/src/main/java/org/jboss/as/controller/SimpleResourceDefinition.java
```diff
@@ -241,19 +241,6 @@ protected void registerRemoveOperation(final ManagementResourceRegistration regi
...
 
-    /**
-     * Registers add operation
-     *
-     * @param registration resource on which to register
-     * @param handler      operation handler to register
-     * @param flags        with flags
-     * @deprecated Redundant with {@link #registerAddOperation(ManagementResourceRegistration, OperationStepHandler, org.jboss.as.controller.registry.OperationEntry.Flag...)
-     */
-    @Deprecated(forRemoval = true)
```
darranl, line 252: This comment is not so much about this PR but for future deprecations it would be useful to add the since attribute so we can gauge how long people have had to migrate.

bstansberry: Hmm, actually the WFCORE-6407 change last year undeprecated the method we now want, and deprecated this one that the previous deprecation said to use. So I think even WF 35 is too quick for this.

The issue with leaving this around is code that wishes to use the correct method but is passing in an AbstractAddStepHandler needs to cast the argument to get the compiler to use the desired method. A clunky change like that in full WF is what inspired me to get rid of this.

PR #6090, opened 2024-07-19
WFCORE-6911 Remove usage of deprecated AbstractAddStepHandler constructors in elytron subsystem

pferraro:
https://issues.redhat.com/browse/WFCORE-6911


darranl (approved):
Our add operations to have extensive testing, probably every test cases that needs to define security resources exercises the add handlers so I think we are good here.

PR #6092, opened 2024-07-19
WFCORE-6913 Remove usage of deprecated AbstractAddStepHandler constructors in logging subsystem

pferraro:
https://issues.redhat.com/browse/WFCORE-6913

### DIFF
--- a/logging/src/main/java/org/jboss/as/logging/formatters/StructuredFormatterResourceDefinition.java
+++ b/logging/src/main/java/org/jboss/as/logging/formatters/StructuredFormatterResourceDefinition.java
@@ -16,6 +16,7 @@ import org.jboss.as.controller.ObjectTypeAttributeDefinition;
 import org.jboss.as.controller.OperationContext;
 import org.jboss.as.controller.OperationFailedException;
 import org.jboss.as.controller.OperationStepHandler;
+import org.jboss.as.controller.PathAddress;
 import org.jboss.as.controller.PathElement;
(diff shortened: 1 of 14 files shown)


logging/src/main/java/org/jboss/as/logging/formatters/StructuredFormatterResourceDefinition.java
```diff
@@ -349,10 +349,11 @@ private static String modelValueToMetaData(final ModelNode metaData) {
 
     private static class AddStructuredFormatterStepHandler extends LoggingOperations.LoggingAddOperationStepHandler {
         private final Class<? extends StructuredFormatter> type;
+        private final AttributeDefinition[] attributes;
 
         private AddStructuredFormatterStepHandler(final Class<? extends StructuredFormatter> type, final AttributeDefinition[] attributes) {
-            super(attributes);
             this.type = type;
+            this.attributes = attributes;
```
yersan, line 356: Cannot these attributes be retrieved from the context.getRegistration instead in the performRuntime method?

pferraro: I originally kept this in order to keep the changes minimal, but yes, with some more work this can be eliminated.
I've updated this PR with more extensive changes required handle this removal.

jamezp (approved):
This LGTM thank you @pferraro

PR #6100, opened 2024-07-19
WFCORE-6921 Remove usage of deprecated AbstractWriteAttributeHandler constructors in domain-management

pferraro:
https://issues.redhat.com/browse/WFCORE-6921

### DIFF
--- a/domain-management/src/main/java/org/jboss/as/domain/management/audit/SyslogAuditLogHandlerResourceDefinition.java
+++ b/domain-management/src/main/java/org/jboss/as/domain/management/audit/SyslogAuditLogHandlerResourceDefinition.java
@@ -132,7 +132,7 @@ public class SyslogAuditLogHandlerResourceDefinition extends AuditLogHandlerReso
     @Override
     public void registerAttributes(ManagementResourceRegistration resourceRegistration) {
         super.registerAttributes(resourceRegistration);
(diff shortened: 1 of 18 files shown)


domain-management/src/main/java/org/jboss/as/domain/management/audit/SyslogAuditLogHandlerResourceDefinition.java
```diff
@@ -412,7 +412,7 @@ protected void revertUpdateToRuntime(OperationContext context, ModelNode operati
                 addr = addr.subAddress(0, addr.size() - 1);
                 auditLogger.updateSyslogHandlerReconnectTimeout(Util.getNameFromAddress(addr), valueToRestore.asInt());
             } else if (attributeName.equals(KEYSTORE_PASSWORD_CREDENTIAL_REFERENCE_NAME) || attributeName.equals(KEY_PASSWORD_CREDENTIAL_REFERENCE_NAME)) {
-                rollbackCredentialStoreUpdate(getAttributeDefinition(attributeName), context, valueToRevert);
+                rollbackCredentialStoreUpdate(context.getResourceRegistration().getAttributeAccess(PathAddress.EMPTY_ADDRESS, attributeName).getAttributeDefinition(), context, valueToRevert);
```
yersan, line 415: +1 The usage of this method is tricky one since it is still deprecated but not removed with this PR because it needs chances in WildFly too, however, we have to prevent that nobody uses it if we are removing the constructor with the attributes, that's my understanding.

yersan (approved):

PR #6102, opened 2024-07-19
WFCORE-6923 Remove usage of deprecated AbstractWriteAttributeHandler constructors in host-controller

pferraro:
https://issues.redhat.com/browse/WFCORE-6923


host-controller/src/main/java/org/jboss/as/domain/controller/resources/ProfileResourceDefinition.java
```diff
@@ -98,6 +98,6 @@ public void registerAttributes(ManagementResourceRegistration resourceRegistrati
     }
 
     public static OperationStepHandler createIncludesValidationHandler() {
-        return new DomainIncludesValidationWriteAttributeHandler(INCLUDES);
+        return new DomainIncludesValidationWriteAttributeHandler();
```
yersan, line 101: We could use a `DomainIncludesValidationWriteAttributeHandler.INSTANCE ` since it is used in several places, I leave it up to you @pferraro

yersan (approved):

PR #6109, opened 2024-07-19
WFCORE-6930 Remove usage of deprecated AbstractWriteAttributeHandler constructors in testsuite

pferraro:
https://issues.redhat.com/browse/WFCORE-6930


subsystem-test/tests/src/test/java/org/jboss/as/subsystem/test/transformers/subsystem/simple/VersionedExtensionCommon.java
```diff
@@ -128,6 +128,10 @@ protected void performRuntime(OperationContext context, ModelNode operation, Mod
 
     private static class BasicAttributeWriteHandler extends AbstractWriteAttributeHandler<Void> {
 
+        protected BasicAttributeWriteHandler() {
+            super(List.of());
+        }
+
```
bstansberry, line 134: @pferraro This class seems wrong; it seems to be adding a new call to a deprecated c'tor instead of removing the existing one.

bstansberry (changes requested):

subsystem-test/tests/src/test/java/org/jboss/as/subsystem/test/transformers/subsystem/simple/VersionedExtensionCommon.java
pferraro: Hmm - not sure what I was thinking here.  Will fix.

bstansberry (approved):

PR #6110, opened 2024-07-19
WFCORE-6931 Remove usage of deprecated AbstractWriteAttributeHandler constructors in threads subsystem

pferraro:
https://issues.redhat.com/browse/WFCORE-6931

### DIFF
--- a/threads/src/main/java/org/jboss/as/threads/ThreadSubsystemResourceDefinition.java
+++ b/threads/src/main/java/org/jboss/as/threads/ThreadSubsystemResourceDefinition.java
@@ -5,16 +5,11 @@
 
 package org.jboss.as.threads;
 
-import org.jboss.as.controller.AttributeDefinition;
-import org.jboss.as.controller.PersistentResourceDefinition;
 import org.jboss.as.controller.ReloadRequiredRemoveStepHandler;
+import org.jboss.as.controller.SimpleResourceDefinition;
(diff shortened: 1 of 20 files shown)


threads/src/main/java/org/jboss/as/threads/ThreadSubsystemResourceDefinition.java
```diff
@@ -5,53 +5,38 @@
...
-
 /**
  * {@link org.jboss.as.controller.ResourceDefinition} for the root resource of the threads subsystem.
  *
  * @author Brian Stansberry (c) 2011 Red Hat Inc.
  */
 @Deprecated(forRemoval = true)
 @SuppressWarnings({"removal", "DeprecatedIsStillUsed"})
-class ThreadSubsystemResourceDefinition extends PersistentResourceDefinition {
+class ThreadSubsystemResourceDefinition extends SimpleResourceDefinition {
```
yersan, line 21: @pferraro 

I would like to understand the usage of PersistentResourceDefinition better.
Could you explain when a resource does not need to be managed as a PersistentResourceDefinition? I don't get the reason for this change yet, which in turn cascades it to the other resources

pferraro: See https://issues.redhat.com/browse/WFCORE-6864

yersan (approved):

yersan:
Thanks @pferraro

PR #6114, opened 2024-07-23
[WFCORE-6936] Allow the maven.test.redirectTestOutputToFile system pr…

jamezp:
…operty work as well as the testLogToFile property. Remove the duplication of the <redirectTestOutputToFile/> setting in the various POM's.

https://issues.redhat.com/browse/WFCORE-6936

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -111,6 +111,12 @@
             ${surefire.jacoco.args} -Djava.io.tmpdir=${project.build.directory}</surefire.jvm.args>
         <surefire.system.args>${surefire.jpda.args} ${surefire.jvm.args}</surefire.system.args>
         <test.level>INFO</test.level>
+        <!-- The maven.test.redirectTestOutputToFile is used by the surefire-maven-plugin and failsafe-maven-plugin to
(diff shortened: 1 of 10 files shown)


pom.xml
```diff
@@ -339,7 +341,7 @@
                     </dependencies>
                     <configuration>
                         <forkNode implementation="org.apache.maven.plugin.surefire.extensions.SurefireForkNodeFactory"/>
-                        <redirectTestOutputToFile>true</redirectTestOutputToFile>
+                        <redirectTestOutputToFile>${maven.test.redirectTestOutputToFile}</redirectTestOutputToFile>
```
yersan, line 344: `maven.test.redirectTestOutputToFile` is the user property for `redirectTestOutputToFile`. Just out of curiosity @jamezp , it didn't work well by having it defined at the property section above in this `pom.xml`?

Without playing with it, looks like it could be optional to define it here

jamezp: You're right Yeray. That works, I'll update the PR now.

bstansberry (approved):

PR #6115, opened 2024-07-23
[WFCORE-6933] Add the required permissions for the Log4j2ServiceActiv…

jamezp:
…ator when deploying.

https://issues.redhat.com/browse/WFCORE-6933


jamezp:
> @jamezp I have no idea how this relates to the stack trace in the JIRA description, but it seems the right thing to do regardless and I don't see the 'HttpString' failure in the SM CI job server log, so LGTM. :-)

@bstansberry yes, the stack trace I saw did not match. I did have get some SM exceptions. Once those were fixed, it seemed to work okay.

bstansberry:
@jamezp Thinking about it, that stack trace is odd. It fails in clinit of the packaged-in-a-module io.undertow.util.HttpString class. It's kind of surprising that class wouldn't be init'd by the server itself prior to the deployment using it. It is a manualmode test though, so the call paths exercised in the server process are more limited.

If it crops up again we'll have a head start.

PR #6123, opened 2024-08-06
WFCORE-6663 Introduce AttributeDefinition implementations to wildfly-subsystem that can resolve directly to a ServiceDependency

pferraro:
https://issues.redhat.com/browse/WFCORE-6663

e.g.
```java
static CapabilityReferenceAttributeDefinition<Foo> FOO = new CapabilityReferenceAttributeDefinition.Builder("foo", CapabilityReference.builder(CAPABILITY, Foo.SERVICE_DESCRIPTOR).build()).build();

static CapabilityReferenceListAttributeDefinition<Bar> BAR = new CapabilityReferenceListAttributeDefinition.Builder("bar", CapabilityReference.builder(CAPABILITY, Bar.SERVICE_DESCRIPTOR).build()).build();
```

Within an operation runtime handler, we can use the context of the capability reference to resolve our attribute from the model directly to a ServiceDependency:

```java
ServiceDependency<Foo> foo = FOO.resolve(context, model);

ServiceDependency<List<Bar>> bars = BAR.resolve(context, model);
```

If FOO is not required, the resulting ServiceDependency will either return a null, or, if the service descriptor provides alternate default resolution, will result to the default value.
Similarly, as a multi-value attribute, BAR resolves to a service dependency providing an empty list if not required.

### DIFF
--- a/controller/src/main/java/org/jboss/as/controller/StringListAttributeDefinition.java
+++ b/controller/src/main/java/org/jboss/as/controller/StringListAttributeDefinition.java
@@ -86,10 +86,12 @@ public final class StringListAttributeDefinition extends PrimitiveListAttributeD
             setElementValidator(new ModelTypeValidator(ModelType.STRING));
         }
 
-        public Builder(final StringListAttributeDefinition basic) {
-            super(basic);
-            setAttributeParser(AttributeParser.STRING_LIST);
-            setAttributeMarshaller(AttributeMarshaller.STRING_LIST);
+        public Builder(String name, StringListAttributeDefinition basis) {
+            super(name, basis);
+        }
+
+        public Builder(final StringListAttributeDefinition basis) {
+            this(basis.getName(), basis);
         }
 
         @Override
--- a/controller/src/main/java/org/jboss/as/controller/StringListAttributeDefinition.java
+++ b/controller/src/main/java/org/jboss/as/controller/StringListAttributeDefinition.java
@@ -17,9 +17,9 @@ import org.jboss.dmr.ModelType;
 /**
  * @author <a href="mailto:tomaz.cerar@redhat.com">Tomaz Cerar</a>
  */
-public final class StringListAttributeDefinition extends PrimitiveListAttributeDefinition {
+public class StringListAttributeDefinition extends PrimitiveListAttributeDefinition {
 
-    private StringListAttributeDefinition(Builder builder) {
+    protected <B extends ListAttributeDefinition.Builder<B, A>, A extends ListAttributeDefinition> StringListAttributeDefinition(ListAttributeDefinition.Builder<B, A> builder) {
         super(builder, ModelType.STRING);
     }
 
@@ -91,14 +91,12 @@ public final class StringListAttributeDefinition extends PrimitiveListAttributeD
         }
 
         public Builder(final StringListAttributeDefinition basis) {
-            this(basis.getName(), basis);
+            super(basis);
         }
 
         @Override
         public StringListAttributeDefinition build() {
             return new StringListAttributeDefinition(this);
         }
-
     }
-
 }
--- a/io/subsystem/src/main/java/org/wildfly/extension/io/IOSubsystemRegistrar.java
+++ b/io/subsystem/src/main/java/org/wildfly/extension/io/IOSubsystemRegistrar.java
@@ -9,21 +9,17 @@ import java.util.ArrayList;
 import java.util.List;
 import java.util.concurrent.atomic.AtomicInteger;
 
-import org.jboss.as.controller.AttributeDefinition;
 import org.jboss.as.controller.OperationContext;
 import org.jboss.as.controller.OperationFailedException;
 import org.jboss.as.controller.PathElement;
 import org.jboss.as.controller.ResourceDefinition;
 import org.jboss.as.controller.ResourceRegistration;
-import org.jboss.as.controller.SimpleAttributeDefinitionBuilder;
 import org.jboss.as.controller.SubsystemRegistration;
 import org.jboss.as.controller.capability.RuntimeCapability;
 import org.jboss.as.controller.descriptions.ParentResourceDescriptionResolver;
 import org.jboss.as.controller.descriptions.SubsystemResourceDescriptionResolver;
-import org.jboss.as.controller.registry.AttributeAccess;
 import org.jboss.as.controller.registry.ManagementResourceRegistration;
 import org.jboss.dmr.ModelNode;
-import org.jboss.dmr.ModelType;
 import org.wildfly.common.function.Functions;
 import org.wildfly.io.IOServiceDescriptor;
 import org.wildfly.subsystem.resource.ManagementResourceRegistrar;
@@ -31,11 +27,13 @@ import org.wildfly.subsystem.resource.ManagementResourceRegistrationContext;
 import org.wildfly.subsystem.resource.ResourceDescriptor;
 import org.wildfly.subsystem.resource.SubsystemResourceDefinitionRegistrar;
 import org.wildfly.subsystem.resource.capability.CapabilityReference;
+import org.wildfly.subsystem.resource.capability.CapabilityReferenceAttributeDefinition;
 import org.wildfly.subsystem.resource.operation.ResourceOperationRuntimeHandler;
 import org.wildfly.subsystem.service.ResourceServiceConfigurator;
 import org.wildfly.subsystem.service.ResourceServiceInstaller;
 import org.wildfly.subsystem.service.ServiceDependency;
 import org.wildfly.subsystem.service.capability.CapabilityServiceInstaller;
+import org.xnio.XnioWorker;
 
 /**
  * @author <a href="mailto:tomaz.cerar@redhat.com">Tomaz Cerar</a> (c) 2013 Red Hat Inc.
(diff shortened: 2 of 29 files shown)


yersan (commented):
@bstansberry do you plan to review this one? I just added minimal comments, but I think it is good to go, most of the relevant changes go into server and subsystem modules.

It is blocking work on WildFly, so I am in favor of releasing it the sooner, the better, I you plan to review this I will differ it for the next release, otherwise it has chance to be ready for a release for this week.

subsystem/src/main/java/org/wildfly/subsystem/resource/capability/CapabilityReferenceListAttributeDefinition.java
```diff
@@ -0,0 +1,99 @@
...
+import org.jboss.dmr.ModelNode;
+import org.jboss.dmr.ModelType;
+import org.wildfly.subsystem.resource.SimpleResource;
+import org.wildfly.subsystem.resource.ResourceModelResolver;
+import org.wildfly.subsystem.service.ServiceDependency;
+
+/**
+ * An list attribute definition referencing a capability.
+ * Resolves directly to a list of {@link ServiceDependency} via {@link #resolve(OperationContext, ModelNode)}.
+ */
```
yersan, line 30: Maybe specify here that is only valid for a String List attribute definition? Otherwise the ModelType could be extracted as a generic value so any code using it can specify the model type, or maybe the intention was to make this extends `StringListAttributeDefinition`?

service/src/main/java/org/wildfly/service/descriptor/UnaryServiceDescriptor.java
```diff
@@ -21,9 +19,7 @@ public interface UnaryServiceDescriptor<T> extends ServiceDescriptor<T> {
      * @return a tuple containing the resolved name and dynamic segments
      */
     default Map.Entry<String, String[]> resolve(String reference) {
-        return Map.entry(this.getName(), new String[] {
-                Assert.checkNotNullParamWithNullPointerException("reference", reference)
-        });
+        return (reference != null) ? Map.entry(this.getName(), new String[] { reference }) : null;
```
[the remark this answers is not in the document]
pferraro: This change intentional and is required by the CapabilityReferenceAttributeDefinition for null handling.
However, to your point, I should update the javadoc to clarify this.

subsystem/src/main/java/org/wildfly/subsystem/resource/capability/CapabilityReferenceListAttributeDefinition.java
pferraro: The element ModelType is meant to be an implementation detail.
I would have extended StringListAttributeDefinition, but it (or rather its builder) is not extensible.

controller/src/main/java/org/jboss/as/controller/StringListAttributeDefinition.java
```diff
@@ -17,9 +17,9 @@
 /**
  * @author <a href="mailto:tomaz.cerar@redhat.com">Tomaz Cerar</a>
  */
-public final class StringListAttributeDefinition extends PrimitiveListAttributeDefinition {
+public class StringListAttributeDefinition extends PrimitiveListAttributeDefinition {
```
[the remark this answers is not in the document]
pferraro: I originally modified this with the intention of subclassing it, but later resorted to extending PrimitiveListAttributeDefinition since it was easier to extend that class's builder.  However, CapabilityReferenceListAttributeDefinition can still subclass StringListAttributeDefinition, even if its builder does not.  I will sort that out.

subsystem/src/main/java/org/wildfly/subsystem/resource/capability/CapabilityReferenceListAttributeDefinition.java
pferraro: This extends StringListAttributeDefinition now.

controller/src/main/java/org/jboss/as/controller/StringListAttributeDefinition.java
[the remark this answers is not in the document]
pferraro: CapabilityReferenceListAttributeDefinition now extends StringListAttributeDefinition.

subsystem/src/main/java/org/wildfly/subsystem/resource/capability/CapabilityReferenceAttributeDefinition.java
```diff
@@ -0,0 +1,86 @@
...
+        public Builder(String attributeName, CapabilityReferenceAttributeDefinition<T> basis) {
+            super(attributeName, basis);
+            this.resolver = basis.resolver;
+        }
+
+        @Override
+        public Builder<T> setDefaultValue(ModelNode defaultValue) {
+            // A capability reference must not specify a default value
+            if ((defaultValue != null) && defaultValue.isDefined()) {
+                throw new UnsupportedOperationException();
```
bstansberry, line 67: @pferraro This should be explained in javadoc.

subsystem/src/main/java/org/wildfly/subsystem/resource/capability/CapabilityReferenceAttributeDefinition.java
```diff
@@ -0,0 +1,86 @@
...
+            }
+            return this;
+        }
+
+        @Override
+        public Builder<T> setAllowExpression(boolean allowExpression) {
```
bstansberry, line 73: @pferraro This should be explained in javadoc.

bstansberry: This meaning L76 :)

subsystem/src/main/java/org/wildfly/subsystem/resource/capability/CapabilityReferenceListAttributeDefinition.java
```diff
@@ -0,0 +1,118 @@
...
+        public Builder(String attributeName, CapabilityReferenceListAttributeDefinition<T> basis) {
+            super(attributeName, basis);
+            this.resolver = basis.resolver;
+        }
+
+        @Override
+        public Builder<T> setDefaultValue(ModelNode defaultValue) {
+            // A capability reference must not specify a default value
+            if ((defaultValue != null) && defaultValue.isDefined()) {
+                throw new UnsupportedOperationException();
```
bstansberry, line 99: javadoc

io/subsystem/src/main/java/org/wildfly/extension/io/IOSubsystemRegistrar.java
```diff
@@ -52,10 +50,8 @@ class IOSubsystemRegistrar implements SubsystemResourceDefinitionRegistrar, Reso
 
     static final ModelNode LEGACY_DEFAULT_WORKER = new ModelNode("default");
 
-    static final AttributeDefinition DEFAULT_WORKER = new SimpleAttributeDefinitionBuilder("default-worker", ModelType.STRING)
+    static final CapabilityReferenceAttributeDefinition<XnioWorker> DEFAULT_WORKER = new CapabilityReferenceAttributeDefinition.Builder<>("default-worker", CapabilityReference.builder(DEFAULT_WORKER_CAPABILITY, IOServiceDescriptor.WORKER).build())
```
bstansberry, line 53: @pferraro Shouldn't there be more to this, i.e. a change to IOSubsystemRegistar.configure?

bstansberry:
@pferraro This looks fine in general (and I know Yeray reviewed it carefully). I made some minor javadoc comments, and then a somewhat more substantial one about IOSubsystemRegistrar, which, if valid should be simple to correct.

subsystem/src/main/java/org/wildfly/subsystem/resource/capability/CapabilityReferenceAttributeDefinition.java
pferraro: Fixed.

subsystem/src/main/java/org/wildfly/subsystem/resource/capability/CapabilityReferenceListAttributeDefinition.java
pferraro: Fixed.

subsystem/src/main/java/org/wildfly/subsystem/resource/capability/CapabilityReferenceListAttributeDefinition.java
```diff
@@ -0,0 +1,118 @@
...
+            }
+            return this;
+        }
+
+        @Override
+        public Builder<T> setAllowExpression(boolean allowExpression) {
+            // A capability reference must not allow expressions
+            if (allowExpression) {
+                throw new UnsupportedOperationException();
```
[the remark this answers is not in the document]
pferraro: Fixed.

subsystem/src/main/java/org/wildfly/subsystem/resource/capability/CapabilityReferenceAttributeDefinition.java
pferraro: Fixed.

io/subsystem/src/main/java/org/wildfly/extension/io/IOSubsystemRegistrar.java
pferraro: Hmm - indeed.  I had added a ServiceDependency.isEmpty() method to better handle the optional capability reference case, but it looks like I stashed that work without committing it.  Will fix.

pferraro: Fixed.

pferraro:
Rebased against main.

yersan (approved):

PR #6126, opened 2024-08-09
WFCORE-6945 Required child resource logic in wildfly-subsystem is not stability-level aware

pferraro:
https://issues.redhat.com/browse/WFCORE-6945

### DIFF
--- a/subsystem/src/main/java/org/wildfly/subsystem/resource/PathElementProvider.java
+++ b/subsystem/src/main/java/org/wildfly/subsystem/resource/PathElementProvider.java
@@ -13,7 +13,7 @@ import org.jboss.as.controller.PathElement;
  * @author Paul Ferraro
  * @deprecated To be removed without replacement
  */
-@Deprecated(forRemoval = true)
+@Deprecated(forRemoval = true, since = "26.0.0")
 public interface PathElementProvider extends Supplier<PathElement> {
 
     /**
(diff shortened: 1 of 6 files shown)


subsystem/src/main/java/org/wildfly/subsystem/resource/PathElementProvider.java
```diff
@@ -11,7 +11,9 @@
 /**
  * Provides a {@link PathElement}.
  * @author Paul Ferraro
+ * @deprecated To be removed without replacement
  */
+@Deprecated(forRemoval = true)
```
yersan, line 16: @pferraro Could you add the since attribute for this deprecation?

@darranl added a relevant comment about its usage at https://github.com/wildfly/wildfly-core/pull/6084#discussion_r1696625630, I also think it is worth using it since it will provide information for future removal, although you can get it from the commit history the "since" makes more sense. Do you also agree? We can start getting it as a rule

pferraro: Done.

yersan: @pferraro FYI, there are more than one @Deprecated annotation that needs to be adjusted with the "since", could you take another look?

yersan (approved):

yersan:
Thanks @pferraro

PR #6134, opened 2024-08-15
[WFCORE-6881] Support Stability Levels for the Management Parsers

darranl:
I am starting off submitting this as a draft PR as technically it is not "needed" until our first new schema is added, but it may make sense to get this in in preparation but that can be discussed.  One challenge is until the first schema is added I can not add a test to test turning the stability levels on and off but the remainder of our testsuites do rely on a large amount of paring just by starting test servers.

As much as I can I have tried to work with the APIs / patterns from Paul where support for stability level based parsing was added to the subsystems, as the core model parsers have a lot of historic code there were some limits where I have needed to adapt here.

https://issues.redhat.com/browse/WFCORE-6881

I am going to be working on the following issues next which will add the new schema and some attributes which may put this in a better context:
   https://issues.redhat.com/browse/WFCORE-6829
   https://issues.redhat.com/browse/WFCORE-6830


core-model-test/framework/src/main/java/org/jboss/as/core/model/test/TestParser.java
```diff
@@ -46,44 +50,115 @@ public TestParser(TestModelType type, XMLElementReader<List<ModelNode>> reader,
...
+        return testParser;
+    }
+
+
+    public static TestParser create(ExtensionRegistry registry, XMLMapper xmlMapper, TestModelType type) {
+        if (getModelMajorVersion() < 27) {
+            try {
+                return createLegacy(registry, xmlMapper, type);
+            } catch (Exception e) {
+                throw new IllegalStateException("Failed to create the parser", e);
```
bstansberry, line 156: i18n

bstansberry (changes requested):
Other than the one i18n thing this seems ok.

core-model-test/framework/src/main/java/org/jboss/as/core/model/test/TestParser.java
darranl: @bstansberry This one is just a part of the core-model-test framework so I think we can stick with English only?

darranl:
I have updated this to be ready for merge, this PR does not add any PRs so also does not need to be held back.

core-model-test/framework/src/main/java/org/jboss/as/core/model/test/TestParser.java
bstansberry: ah, right. No problem.

bstansberry (approved):
@darranl LGTM.

controller/src/main/java/org/jboss/as/controller/parsing/ExtensionXml.java
```diff
@@ -72,7 +72,12 @@ public void writeExtensions(final XMLExtendedStreamWriter writer, final ModelNod
         }
     }
 
-    public void parseExtensions(final XMLExtendedStreamReader reader, final ModelNode address, final Namespace expectedNs, final List<ModelNode> list)
+    @Deprecated
```
yersan, line 75: All the `@Deprecated` should use the "since" argument

yersan (changes requested):

controller/src/main/java/org/jboss/as/controller/parsing/ExtensionXml.java
yersan: This is almost done, and CI looks good. We can add the since argument later. Since this is blocking some other work in WildFly full, let's move on with it now so it can be included in today's release.

yersan:
Thanks @darranl @bstansberry

controller/src/main/java/org/jboss/as/controller/parsing/ExtensionXml.java
darranl: @yersan FYI these deprecated methods will be gone fairly shortly, I just needed them to hang around long enough to not break WildFly.

yersan: Ok, I won't create any follow-up PR then, thanks!

PR #6137, opened 2024-08-21
[WFCORE-6956] Wrong endpoint-name if the JBoss server name and 'jboss…

jbaesner:
https://issues.redhat.com/browse/WFCORE-6956


yersan (approved):

yersan:
Thanks @jbaesner

yersan:
@rachmatowicz makes sense to me, I've created this https://issues.redhat.com/browse/WFCORE-6959 to track it.

My feeling is the behavior of these values is correctly managed and the original bug and resolution were pretty clear to me, but yes, I agree we could have added a test case to cover (and explain) it for the future, thanks for pointing it out.

PR #6142, opened 2024-08-25
[WFCORE-6722] Standardize the error message shown when the server is launched with an valid but empty option

arsenalzp:
Hello,
This PR is fixing the following issue:

https://issues.redhat.com/browse/WFCORE-6722

It is trying to unify error messages when no value was specified for an option.


yersan (approved):

yersan:
@arsenalzp merged, thanks for contributing to WildFly Core!

PR #6145, opened 2024-08-27
[WFCORE-6963] improve AbstractModelResource.hasChildren efficiency

aogburn:
https://issues.redhat.com/browse/WFCORE-6963

### DIFF
--- a/controller/src/main/java/org/jboss/as/controller/registry/AbstractModelResource.java
+++ b/controller/src/main/java/org/jboss/as/controller/registry/AbstractModelResource.java
@@ -274,7 +274,9 @@ public abstract class AbstractModelResource extends ResourceProvider.ResourcePro
 
         @Override
         public boolean hasChildren() {
-            return ! children().isEmpty();
+            synchronized (children) {
+                return ! children.isEmpty();
+            }
         }
 
         @Override


controller/src/main/java/org/jboss/as/controller/registry/AbstractModelResource.java
```diff
@@ -274,7 +274,7 @@ public Resource get(String name) {
 
         @Override
         public boolean hasChildren() {
-            return ! children().isEmpty();
+            return ! children.isEmpty();
```
yersan, line 277: I understand that the bottleneck here is due to the synchronized keyword used by the `children()` method, but the original intention of this `DefaultResourceProvider` seems to be to make it threadsafe.

Calculating whether it is empty without using the synchronized keyword will break this original intention.

@pferraro / @bstansberry What do you think about this and can you figure out why a resource provider should be thread-safe managing its children's resources?

When I think about it, I end up in a possible concurrency where multiple management operations are mutating the model at the same time, is that possible?

bstansberry: It seems reasonable to replace the call to children() with a direct read of children, but in a synchronized block. I don't see why copying out the keys is needed just to see if there are any.

aogburn: Thanks I added the synchronized block

yersan (approved):

yersan:
Thanks @aogburn @bstansberry

PR #6058, opened 2024-07-04
WFCORE-6882 Upgrade to wildfly-common 2.0.0 / WFCORE-6883 Upgrade to jboss-logging 3.6.0 / jboss-logging-tools 3.0.0

marko-bekhta:
https://issues.redhat.com/browse/WFCORE-6882
https://issues.redhat.com/browse/WFCORE-6883

Hey @jamezp 👋🏻 😃 

It's probably easier to look at the patch by commit.

- the new generator seems to fail on `JmxLogger` where the exception requires at least an exception as a parameter, so I've updated the interface to include it. 
- wildfly-common has a bunch of `io.smallrye.common` dependencies, so I've added them, and also created a `io.smallrye.common` module, let me know if that's the way to go.
- The `wildfly-maven-plugin` requires overriding the logging dependency as it pulls the previous one, and with all the changes for the new API, the build fails if the dependency is not overridden...

PR #6146, opened 2024-08-28
WFCORE-6883 Upgrade to jboss-logging 3.6.0 / jboss-logging-tools 3.0.0 

jamezp:
Testing #6058  with some minor changes and better logging output.


server/src/main/java/org/jboss/as/server/ServerService.java
```diff
@@ -433,6 +433,9 @@ public void undeploy(DeploymentUnit context) {
             }
             String message = ServerLogger.ROOT_LOGGER.unsuccessfulBoot(String.join(" ", messages));
             bootstrapListener.bootFailure(new Exception(message, cause));
+            if (cause != null) {
+                cause.printStackTrace();
+            }
```
jamezp, line 438: I'm pretty sure this fixes the test failure, but I need to understand why still. I'll have to do some testing against the main branch with logging 3.5.3 to see why this doesn't cause an issue there.

PR #6147, opened 2024-08-28
[WFCORE-6960]: Promote Simple config export for a server as an attachment from COMMUNITY to DEFAULT.

ehsavoie:
* Adding more test using the CLI.
* Moving from COMMUNITY to DEFAULT
* Renaming the ManagementReadXmlCommunityTestCase to ManagementReadXmlAsFileTestCase.

Jira: https://issues.redhat.com/browse/WFCORE-6960
Proposal: https://github.com/wildfly/wildfly-proposals/pull/594

### DIFF
--- a/controller/src/main/java/org/jboss/as/controller/operations/common/XmlFileMarshallingHandler.java
+++ b/controller/src/main/java/org/jboss/as/controller/operations/common/XmlFileMarshallingHandler.java
@@ -9,6 +9,7 @@ import static org.jboss.as.controller.descriptions.ModelDescriptionConstants.UUI
 
 import java.io.ByteArrayInputStream;
 import java.io.ByteArrayOutputStream;
+import java.util.UUID;
 
 import org.jboss.as.controller.OperationContext;
 import org.jboss.as.controller.SimpleAttributeDefinitionBuilder;
@@ -18,7 +19,6 @@ import org.jboss.as.controller.access.management.SensitiveTargetAccessConstraint
 import org.jboss.as.controller.descriptions.ModelDescriptionConstants;
 import org.jboss.as.controller.descriptions.common.ControllerResolver;
 import org.jboss.as.controller.persistence.ConfigurationPersister;
-import org.jboss.as.version.Stability;
 import org.jboss.dmr.ModelType;
 
 /**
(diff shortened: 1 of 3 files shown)


testsuite/domain/src/test/java/org/jboss/as/test/integration/domain/suites/ManagementReadXmlAsFileTestCase.java
```diff
@@ -25,47 +23,33 @@
...
-import org.wildfly.test.stability.StabilityDomainSetupSnapshotRestoreTasks;
 
 /**
  * Test of various read-config-as-xml-file against the domain controller.
  *
  * @author Emmanuel Hugonnet (c) 2022 Red Hat, Inc.
  */
-public class ManagementReadXmlCommunityTestCase extends ManagementReadXmlTestCase{
-    static StabilityDomainSetupSnapshotRestoreTasks tasks;
+public class ManagementReadXmlAsFileTestCase extends ManagementReadXmlTestCase {
```
darranl, line 43: Just from our general discussion today, as this is not referenced by the Suite it may not be being executed in CI so worth a double check.

ehsavoie: I noticed that ManagementReadXmlTestCase wasn't referenced either.

fabiobrz (approved):
Tests LGTM, and they reflect what in the community proposal and related feature test plan.
I see CI checks are failing but it seems that's about an unrelated test. Approving from QA perspective

controller/src/main/java/org/jboss/as/controller/operations/common/XmlFileMarshallingHandler.java
```diff
@@ -36,7 +36,7 @@ public class XmlFileMarshallingHandler extends AbstractXmlMarshallingHandler {
             .setReplyParameters(new SimpleAttributeDefinitionBuilder(UUID, ModelType.STRING, false).build())
             .setReadOnly()
             .setRuntimeOnly()
-            .setStability(Stability.COMMUNITY)
+            .setStability(Stability.DEFAULT)
```
darranl, line 39: I think we can remove the setStability call here, DEFAULT is the default when none is set so we can probably have one line less.

yersan (approved):

yersan:
Thanks @ehsavoie @fabiobrz @darranl

PR #6148, opened 2024-08-28
Wfcore 6967 intermittent failures jwt identity propagation test case

rsearls:
https://issues.redhat.com/browse/WFCORE-6967

This is a temporary code in order to assist in debugging a test failure in wildfly.  
A "custom" team city build appears to only allow use of wildfly-core main branch blds.
The code change will be removed in the future.


yersan:
@rsearls 

> No longer needed. Using custom builds in team city

In such a case, please close it, or if you still need to use it to debug CI Jobs, you can use a draft instead of a PR. That will provide better information to the reviewers

PR #6156, opened 2024-09-02
WFCORE-6974 Add convenience factory method to create a ServiceDependency from a Supplier.

pferraro:
https://issues.redhat.com/browse/WFCORE-6974


pferraro:
@yersan FYI - I updated this to additionally validate a non-null supplier, which would otherwise defer the NPE until the returned dependency is used.

bstansberry:
Thanks @pferraro and @yersan

PR #6157, opened 2024-09-02
WFCORE-6975 Add convenience method to combine a ServiceDependency with another.

pferraro:
https://issues.redhat.com/browse/WFCORE-6975

Based on https://github.com/wildfly/wildfly-core/pull/6156 to avoid conflicts.

### DIFF
--- a/service/src/main/java/org/wildfly/service/Dependency.java
+++ b/service/src/main/java/org/wildfly/service/Dependency.java
@@ -51,6 +51,25 @@ public interface Dependency<B extends ServiceBuilder<?>, V> extends Consumer<B>,
         }
     }
 
+    class SuppliedDependency<B extends ServiceBuilder<?>, V> implements Dependency<B, V> {
+
+        private final Supplier<V> supplier;
+
+        protected SuppliedDependency(Supplier<V> supplier) {
+            this.supplier = supplier;
+        }
+
+        @Override
+        public V get() {
+            return this.supplier.get();
+        }
+
+        @Override
+        public void accept(B builder) {
+            // Nothing to register
+        }
+    }
+
     abstract class AbstractDependency<B extends ServiceBuilder<?>, V> implements Dependency<B, V>, Function<B, Supplier<V>> {
 
         private volatile Supplier<V> supplier = Functions.constantSupplier(null);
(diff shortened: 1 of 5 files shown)


service/src/main/java/org/wildfly/service/Dependency.java
```diff
@@ -18,6 +19,28 @@
  */
 public interface Dependency<B extends ServiceBuilder<?>, V> extends Consumer<B>, Supplier<V> {
 
+    @Override
+    default Dependency<B, V> andThen(Consumer<? super B> after) {
+        return new Dependency<>() {
```
yersan, line 24: Should not do a sanity check `Objects.requireNonNull(after);`?

service/src/main/java/org/wildfly/service/ServiceDependency.java
```diff
@@ -15,6 +18,22 @@
  */
 public interface ServiceDependency<V> extends Dependency<ServiceBuilder<?>, V> {
 
+    @Override
+    default ServiceDependency<V> andThen(Consumer<? super ServiceBuilder<?>> after) {
+        return new ServiceDependency<>() {
```
yersan, line 23: Missed Objects.requireNonNull(after); sanity check?

subsystem/src/main/java/org/wildfly/subsystem/service/ServiceDependency.java
```diff
@@ -25,6 +27,22 @@
  */
 public interface ServiceDependency<V> extends Dependency<RequirementServiceBuilder<?>, V> {
 
+    @Override
+    default ServiceDependency<V> andThen(Consumer<? super RequirementServiceBuilder<?>> after) {
+        return new ServiceDependency<>() {
```
yersan, line 32: Missed Objects.requireNonNull(after); sanity check?

service/src/main/java/org/wildfly/service/Dependency.java
pferraro: Sounds good.

yersan (approved):

yersan:
Thanks @pferraro

PR #6165, opened 2024-09-06
[WFCORE-6982] Turn on compiler annotation processing on SE 23+

ropalka:
https://issues.redhat.com/browse/WFCORE-6982


yersan:
Thanks @ropalka @bstansberry

PR #6152, opened 2024-08-29
[WFCORE-6829] Add a new 'urn:jboss:domain:community:20.0' schema.

darranl:
Also update the SupportedNamespaceParsingTestCases to include stability level tests.

https://issues.redhat.com/browse/WFCORE-6829

I will close after CI and submit along with the PR that adds new attributes, first I want to check if there are any unintended consequences across the wider testsuites - a simple local build passes fine.

PR #6166, opened 2024-09-06
[COMMUNITY] [WFCORE-6830] Add four new attributes to the HTTP management interface

darranl:
https://issues.redhat.com/browse/WFCORE-6830

This is a continuation of the addition of the new schema added under:
    https://github.com/wildfly/wildfly-core/pull/6152

https://issues.redhat.com/browse/WFCORE-6829

This is also a continuation of https://github.com/wildfly/wildfly-core/pull/6004 which is the actual implementation.  This PR focuses on adding the four configuration options to the HTTP management interface.

     - backlog
     - no-request-timeout
     - connection-high-water
     - connection-low-water

WildFly Proposal: https://github.com/wildfly/wildfly-proposals/pull/595


controller/src/main/java/org/jboss/as/controller/management/BaseHttpInterfaceResourceDefinition.java
```diff
@@ -181,8 +182,37 @@ public void validateParameter(String parameterName, ModelNode value) throws Oper
...
+    public static final SimpleAttributeDefinition BACKLOG = new SimpleAttributeDefinitionBuilder(ModelDescriptionConstants.BACKLOG, ModelType.INT, true)
+            .setAllowExpression(true)
+            .setDefaultValue(new ModelNode(50))
+            .setFlags(AttributeAccess.Flag.RESTART_RESOURCE_SERVICES)
+            .setStability(Stability.COMMUNITY)
+            .build();
+
+    public static final SimpleAttributeDefinition NO_REQUEST_TIMEOUT = new SimpleAttributeDefinitionBuilder(ModelDescriptionConstants.NO_REQUEST_TIMEOUT, ModelType.INT, true)
+            .setAllowExpression(true)
+            .setDefaultValue(new ModelNode(60000))
```
bstansberry, line 194: Please add a setMeasurementUnit

domain-http/interface/src/main/java/org/jboss/as/domain/http/server/ManagementHttpServer.java
```diff
@@ -252,6 +252,9 @@ private static ManagementHttpServer create(Builder builder) {
             }
         }
 
+        ROOT_LOGGER.debugf("Resource Constraints backlog=%d, noRequestTimeout=%d, connectionHighWater=%d, connectionLowWater=%d",
```
bstansberry, line 255: "Resource Constraints" is pretty vague in a message from a broadly scoped thing like ROOT_LOGGER.

bstansberry (changes requested):
@darranl I've only got a couple minor things.

testsuite/domain/src/test/resources/base-server-config.xml
```diff
@@ -1,6 +1,6 @@
 <?xml version="1.0" encoding="UTF-8"?>
 
-<server xmlns="urn:jboss:domain:20.0" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xsi:schemaLocation="urn:jboss:domain:20.0 wildfly-config_20_0.xsd">
+<server xmlns="urn:jboss:domain:community:20.0" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xsi:schemaLocation="urn:jboss:domain:community:20.0 wildfly-config_community_20_0.xsd">
```
yersan, line 3: I think our last agreement was not to have community schemas in these configuration files. If we have specific community stability level tests, we should reload the server into that stability level before running those tests. Otherwise, we will need to convert these configuration files back into default schema when creating the product branches and we want to avoid that.

yersan (changes requested):

bstansberry:
I've merged the creaper upgrade in full WF and manually started new runs of the 3 "Full integration" jobs that failed due to the need for that.

testsuite/domain/src/test/resources/base-server-config.xml
darranl: @yersan FYI we have the following Jira issue to come back and follow up on this https://issues.redhat.com/browse/WFCORE-6972 this is why the commit message says it is a temporary change.  I have bumped the follow up Jira issue to Critical so we can come back to these as these configuration files already contain community schema references so this is not something new.

Added to this this is becoming a real rabbit hole which is intertwined by tests @ehsavoie is also evolving further so I think we need to get this PR in and Emmanuel's PR in then come back and fix up this testsuite.

There is also further information on https://github.com/wildfly/wildfly-core/pull/6153 regarding the types of failure we see if we try to just use the default stability level.

domain-http/interface/src/main/java/org/jboss/as/domain/http/server/ManagementHttpServer.java
darranl: @bstansberry although this is ROOT_LOGGER the category logged at is actually "org.jboss.as.domain.http.api.undertow" - I could change the phrasing slightly to something like "Concurrent connection limits" but I think the log category does cover that this is the Undertow HTTP API.

testsuite/domain/src/test/resources/base-server-config.xml
darranl: FYI this was my closing conclusion on the chat thread on the 2nd Sept:

> On the original discussion for this thread, as these tests are also being refactored for their own feature promotion I think I am going to take a step back on that and temporarily just add another community namespace to the configuration schema just so I can progress with my feature.
> 
> Once my feature and Emmanual's promotion is ready we can look at how to combine at the end to complete some of the thoughts here.

domain-http/interface/src/main/java/org/jboss/as/domain/http/server/ManagementHttpServer.java
darranl: Actually may go for "HTTP Management API Connection Constraints"

controller/src/main/java/org/jboss/as/controller/management/BaseHttpInterfaceResourceDefinition.java
darranl: This is added, BTW @bstansberry GitHub says FAHRENHEIGHT was your contribution? :smile:

domain-http/interface/src/main/java/org/jboss/as/domain/http/server/ManagementHttpServer.java
darranl: Added, I went for the second option to add some clarity.

controller/src/main/java/org/jboss/as/controller/management/BaseHttpInterfaceResourceDefinition.java
bstansberry: Somebody hacked me!

bstansberry: Not guilty! Except of careless cutting and pasting!

https://access.redhat.com/webassets/avalon/d/Red_Hat_JBoss_Operations_Network-3.1-API-en-US/files/domain/index.html?org/rhq/core/domain/measurement/MeasurementUnits.html

domain-http/interface/src/main/java/org/jboss/as/domain/http/server/ManagementHttpServer.java
bstansberry: Oh sorry; for some reason I thought this was in controller or server or something. Anyway, the change is an improvement.

bstansberry (commented):
@darranl You've addressed my concerns, but from the CI results it seems the changed ManagmentInterfaceResourcesTestCase is unstable.

darranl:
The test failures are odd, checking the line numbers that is actually the existing version of the test failing not the new management model configured version.

testsuite/domain/src/test/resources/base-server-config.xml
yersan: > Added to this this is becoming a real rabbit hole which is intertwined by tests @ehsavoie is also evolving further so I think we need to get this PR in and Emmanuel's PR in then come back and fix up this testsuite.

Ok, got it, I mistakenly thought that the changes here were to accommodate the `ManagementInterfaceResourcesTestCase`, but they are just because after incorporating the community schema we need to bypass the flaws in the `ManagementReadXmlTestCase` test case, which is not able to work with schemas in default stability level.

However, independently of that, since this `ManagementInterfaceResourcesTestCase` now is testing changes that are only in community stability level, should not be already prepared to reload the server into the expected stability level? That would make it to skip these tests if the community stability level is not available and the server was provisioned to default stability level.

I am fine if we postpone this to later, but just wanted to crosscheck if you agreed with that plan.

testsuite/manualmode/src/test/java/org/wildfly/core/test/standalone/mgmt/ManagementInterfaceResourcesTestCase.java
```diff
@@ -128,12 +146,31 @@ public void testTimeout() throws Exception {
...
-            cli.sendLine(String.format("/system-property=%s:add(value=%d)", CONNECTION_HIGH_WATER_PROPERTY, 6));
-            cli.sendLine(String.format("/system-property=%s:add(value=%d)", CONNECTION_LOW_WATER_PROPERTY, 3));
-            cli.sendLine(String.format("/system-property=%s:add(value=%d)", NO_REQUEST_TIMEOUT_PROPERTY, noRequestTimeout));
+            if (useSystemProperties) {
+                cli.sendLine(String.format(SYSTEM_PROPERTY_ADD, BACKLOG_PROPERTY, 2));
+                cli.sendLine(String.format(SYSTEM_PROPERTY_ADD, CONNECTION_HIGH_WATER_PROPERTY, 6));
+                cli.sendLine(String.format(SYSTEM_PROPERTY_ADD, CONNECTION_LOW_WATER_PROPERTY, 3));
+                cli.sendLine(String.format(SYSTEM_PROPERTY_ADD, NO_REQUEST_TIMEOUT_PROPERTY, noRequestTimeout));
+            } else {
+                cli.sendLine(String.format(HTTP_INTERFACE_READ_ATTRIBUTE, BACKLOG_ATTRIBUTE), true);
```
darranl, line 162: @yersan FYI at this point we do not need to reload the server for the test as our testsuites already run the server at the default stability level of the feature pack so here it is already running at the community level.

This is going to be a part of a bigger challenge, our WildFly and WildFly Core testsuites are going to need to remain runnable with the default configuration for the feature pack under test and for both these projects that will mean we will have community schemas in the configuration and we can't assume to run using the stability level of default.

To bridge this gap I have added this test to check if the attribute is present in the management model, if the attribute does not exist we assume we are running at a stability level that does not support it and skip that part of the test.

testsuite/domain/src/test/resources/base-server-config.xml
darranl: FYI I just added an inline comment where the test checks the attributes are available to decide if that part of the test should run / be skipped.

testsuite/manualmode/src/test/java/org/wildfly/core/test/standalone/mgmt/ManagementInterfaceResourcesTestCase.java
yersan: > This is going to be a part of a bigger challenge, our WildFly and WildFly Core testsuites are going to need to remain runnable with the default configuration for the feature pack under test and for both these projects that will mean we will have community schemas in the configuration and we can't assume to run using the stability level of default.

My understanding about the initial approach to challenge this was that community schemas were not going to be present (they were merged but the plan is to rollback them) and any test which is not meant to run for "default" stability level should reload the server to the stability level they need. 

So, hence, if we provision to "community" stability level, the test will reload the server to the "community" stability level (if it is not already running on it) and the test will run. If we provision to "default" then the test will be skipped because the server will not support a reload to "community" stability level. This approach is what I had in mind that we were going to follow. If that is not correct or we require a different approach we should discuss what to do in Zulip.

> To bridge this gap I have added this test to check if the attribute is present in the management model

Ok, that's another approach we can use for this test.

testsuite/manualmode/src/test/java/org/wildfly/core/test/standalone/mgmt/ManagementInterfaceResourcesTestCase.java
```diff
@@ -144,10 +181,17 @@ private void runTest(int noRequestTimeout, ExceptionRunnable<Exception> test) th
...
-                cli.sendLine(String.format("/system-property=%s:remove()", CONNECTION_HIGH_WATER_PROPERTY));
-                cli.sendLine(String.format("/system-property=%s:remove()", CONNECTION_LOW_WATER_PROPERTY));
-                cli.sendLine(String.format("/system-property=%s:remove()", NO_REQUEST_TIMEOUT_PROPERTY));
+                if (useSystemProperties) {
+                    cli.sendLine(String.format(SYSTEM_PROPERTY_REMOVE, BACKLOG_PROPERTY));
+                    cli.sendLine(String.format(SYSTEM_PROPERTY_REMOVE, CONNECTION_HIGH_WATER_PROPERTY));
+                    cli.sendLine(String.format(SYSTEM_PROPERTY_REMOVE, CONNECTION_LOW_WATER_PROPERTY));
+                    cli.sendLine(String.format(SYSTEM_PROPERTY_REMOVE, NO_REQUEST_TIMEOUT_PROPERTY));
+                } else {
+                    cli.sendLine(String.format(HTTP_INTERFACE_UNDEFINE_ATTRIBUTE, BACKLOG_ATTRIBUTE));
```
bstansberry, line 190: I think this needs handling similar to the L163/172 block. In default stability this will fail and the controller.stop will not be called.

bstansberry (changes requested):

testsuite/manualmode/src/test/java/org/wildfly/core/test/standalone/mgmt/ManagementInterfaceResourcesTestCase.java
darranl: @yersan this test is unrelated to the issues with the existing schemas, this is a test that already exists in the manualmode testsuite and already tests using a server provisioned at the default stability level of the feature pack i.e. community.

testsuite/manualmode/src/test/java/org/wildfly/core/test/standalone/mgmt/ManagementInterfaceResourcesTestCase.java
darranl: Ah good catch.

testsuite/manualmode/src/test/java/org/wildfly/core/test/standalone/mgmt/ManagementInterfaceResourcesTestCase.java
yersan: @darranl I think we are talking about different things. Now, with your latest changes skipping the test if the attribute is not present, the potential issue this PR could have introduced when we are converting to product branches is gone. That's fine. However, I was talking about what was the approach we planned to follow to face these cases: reload the server to the desired stability level if that stability level is not "default".

yersan: > I have not added anything to this test to get the server to run at the Commnity stability level, in this module the server already runs at the default stability level of the feature pack there is nothing to need a reload to community as it is already at that stability level.

@darranl Without your change checking whether the attribute is present, we should have tweaked this test as soon as we provision the server to "default" stability level to create the product branches. However, the approach we defined to face this case was to reload the server to the desired stability level if that stability level is not "default". This PR is now approaching this problem differently, which is fine, but I only wanted to explain what we planned to do.

darranl: @yersan I am still not following the place reloading to the community stability level would have a place on that workflow.

In this repository the server is provisioned using the community stability level and runs using the community stability level - reloading to the community stability level would be a noop as we end up where we started.

Once we are in conversion branches and beyond the default stability level doesn't just become the default stability level it becomes the only stability level so the reload to community again serves no purpose.

testsuite/manualmode/src/test/java/org/wildfly/core/test/standalone/mgmt/ManagementInterfaceResourcesTestCase.java
darranl: @bstansberry actual if you scroll up slightly to line 166 I do call controller.stop() the moment the decision is made to not proceed with that test.

bstansberry: Ok, and you return.

bstansberry (approved):

testsuite/manualmode/src/test/java/org/wildfly/core/test/standalone/mgmt/ManagementInterfaceResourcesTestCase.java
yersan: @darranl Yes, let me try to explain this. If we use our current tasks to reload to a desired stability level, see https://issues.redhat.com/browse/WFCORE-6728:

1. The reload will be a noop if the target server is already running on the desired stability level; so in the end there won't be any overhead if we try to reload to "community" stability level a server that is already running at "community" level. Hence, in a nonproduct branch, the test will be executed as we were "not doing" a reload.
2. If the server does not support the desired stability level, for example, at product branches reload to "community" stability level won't be possible because the running server will only support "default" stability level; the test case will be ignored.

The final effect will be that the same code will make the test be ignored at product branches but be executed in nonproduct branches. Test configuration files should be kept using "default" stability level schemas. This "default" schemas can be read by a server that is running on "community" stability level or in "default" stability level.

Ignoring the test if the attributes don't exist is also a valid approach, in the end, what we should use will depend a lot on the test case itself. Having specific tests for community-only tests could make the code easier to promote to a higher stability level, since it promotion would be removing the task that reloads the server to the desired stability level, but I guess that should be done on more complex test cases. Its implementation is more complex ... and I don't know yet if it will cover all the cases or if we are missing something in that puzzle.

darranl: @yersan thank you for the clarification, I understand where you are coming from.  I can move to that pattern in a follow up PR.

We maybe should add something to the table in the wildfly-proposals repo for the feature process describing the testing requirements (I need to check what we already have).

I think we should probably avoid going into the reload detail too much but I can see we should add some guidance discussing that all new tests should use the appropriate server setup task for the target stability level.

yersan:
Thanks @darranl @bstansberry @pedro-hos

testsuite/manualmode/src/test/java/org/wildfly/core/test/standalone/mgmt/ManagementInterfaceResourcesTestCase.java
yersan: > We maybe should add something to the table in the wildfly-proposals repo for the feature process describing the testing requirements (I need to check what we already have).

yes, that maybe could help

PR #6172, opened 2024-09-11
[WFCORE-4395] Move the JDBCRealm Mapper validation to model time check

pedro-hos:
https://issues.redhat.com/browse/WFCORE-4395

### DIFF
--- a/elytron/src/main/java/org/wildfly/extension/elytron/JdbcRealmDefinition.java
+++ b/elytron/src/main/java/org/wildfly/extension/elytron/JdbcRealmDefinition.java
@@ -19,7 +19,6 @@ import static org.wildfly.extension.elytron.ElytronDescriptionConstants.UTF_8;
 import static org.wildfly.extension.elytron._private.ElytronSubsystemMessages.ROOT_LOGGER;
 
 import java.nio.charset.Charset;
-import java.security.InvalidKeyException;
 import java.util.ArrayList;
 import java.util.Collections;
 import java.util.HashMap;
@@ -33,8 +32,10 @@ import org.jboss.as.controller.AttributeDefinition;
 import org.jboss.as.controller.ObjectListAttributeDefinition;
 import org.jboss.as.controller.ObjectTypeAttributeDefinition;
 import org.jboss.as.controller.OperationContext;
+import org.jboss.as.controller.OperationContext.Stage;
 import org.jboss.as.controller.OperationFailedException;
 import org.jboss.as.controller.OperationStepHandler;
+import org.jboss.as.controller.PathAddress;
 import org.jboss.as.controller.PathElement;
 import org.jboss.as.controller.ResourceDefinition;
 import org.jboss.as.controller.RunningMode;
@@ -43,11 +44,12 @@ import org.jboss.as.controller.SimpleAttributeDefinitionBuilder;
 import org.jboss.as.controller.SimpleResourceDefinition;
 import org.jboss.as.controller.capability.RuntimeCapability;
 import org.jboss.as.controller.operations.validation.CharsetValidator;
-import org.jboss.as.controller.operations.validation.StringAllowedValuesValidator;
 import org.jboss.as.controller.operations.validation.IntRangeValidator;
+import org.jboss.as.controller.operations.validation.StringAllowedValuesValidator;
 import org.jboss.as.controller.registry.AttributeAccess;
 import org.jboss.as.controller.registry.ManagementResourceRegistration;
 import org.jboss.as.controller.registry.OperationEntry;
+import org.jboss.as.controller.registry.Resource;
 import org.jboss.dmr.ModelNode;
 import org.jboss.dmr.ModelType;
 import org.jboss.msc.inject.InjectionException;
@@ -455,7 +457,7 @@ class JdbcRealmDefinition extends SimpleResourceDefinition {
     }
 
     interface PasswordMapperObjectDefinition {
-        PasswordKeyMapper toPasswordKeyMapper(OperationContext context, ModelNode propertyNode) throws OperationFailedException, InvalidKeyException;
+        PasswordKeyMapper toPasswordKeyMapper(OperationContext context, ModelNode propertyNode) throws OperationFailedException;
     }
 
     static class AttributeMappingObjectDefinition {
@@ -592,6 +594,13 @@ class JdbcRealmDefinition extends SimpleResourceDefinition {
         }
 
         @Override
+        protected void populateModel(OperationContext context, ModelNode operation, Resource resource)
+                throws OperationFailedException {
+            super.populateModel(context, operation, resource);
+            context.addStep(new JdbcRealmDefinitionValidation(), Stage.MODEL);
+        }
+
+        @Override
         protected void performRuntime(OperationContext context, ModelNode operation, ModelNode model)
                 throws OperationFailedException {
             ServiceTarget serviceTarget = context.getServiceTarget();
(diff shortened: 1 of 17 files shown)


elytron/src/main/java/org/wildfly/extension/elytron/JdbcRealmDefinition.java
```diff
@@ -635,6 +644,32 @@ public void uninject() {
...
+                try {
+                    keyMapper = mapperResource.toPasswordKeyMapper(context, propertyNode);
+                } catch (InvalidKeyException e) {
+                    throw new OperationFailedException("Invalid key type.", e);
+                }
+            }
+
+            return keyMapper;
+        }
+
```
pedro-hos, line 672: I added the Mapper validation, but I don't know if we still need to have this code on we can just remove the `InvalidKeyException` from the interface, and call this without those validations (both `if`). I guess that we can keep the `throw new OperationFailedException("Invalid key type.", e);` only at the `checkKeyMappers`. Or do we need those validations on Runtime time as well?

pedro-hos:
This is the error when try to startup with more than one mapper:

```{bash}
12:04:39,177 INFO  [org.jboss.as] (MSC service thread 1-4) WFLYSRV0049: WildFly 34.0.0.Beta1-SNAPSHOT (WildFly Core 26.0.0.Beta4-SNAPSHOT) starting
12:04:39,210 ERROR [org.jboss.as.controller.management-operation] (ServerService Thread Pool -- 10) WFLYCTL0013: Operation ("add") failed - address: ([
    ("subsystem" => "elytron"),
    ("jdbc-realm" => "servlet-security-jdbc-realm")
]) - failure description: "WFLYELY00034: A principal query can only have a single key mapper"
12:04:39,212 ERROR [org.jboss.as.controller.management-operation] (Controller Boot Thread) "WFLYCTL0193: Failed executing subsystem elytron boot operations"
12:04:39,212 ERROR [org.jboss.as.controller.management-operation] (Controller Boot Thread) WFLYCTL0013: Operation ("parallel-subsystem-boot") failed - address: ([]) - failure description: "\"WFLYCTL0193: Failed executing subsystem elytron boot operations\""
12:04:39,214 FATAL [org.jboss.as.server] (Controller Boot Thread) WFLYSRV0056: Server boot has failed in an unrecoverable manner; exiting. See previous messages for details. - Server configuration file in use: standalone.xml - Minimum feature stability level: community

```

elytron/src/main/java/org/wildfly/extension/elytron/JdbcRealmDefinition.java
```diff
@@ -653,33 +688,46 @@ private AttributeMapper[] resolveAttributeMappers(OperationContext context, Mode
...
+                if (mapperResource == null) {
+                    continue;
+                }
+
+                if (keyMapper != null) {
+                    throw ROOT_LOGGER.jdbcRealmOnlySingleKeyMapperAllowed();
+                }
+
+                try {
+                    keyMapper = mapperResource.toPasswordKeyMapper(context, propertyNode);
```
bstansberry, line 723: This is not allowed. The toPasswordKeyMapper implementations do work that is not meant for Stage.MODEL (all the resolveModelAttribute calls.

elytron/src/main/java/org/wildfly/extension/elytron/JdbcRealmDefinition.java
bstansberry: Can you move this method back where it was? AFAICT it doesn't need to move, and moving it makes this diff very hard to read.

elytron/src/main/java/org/wildfly/extension/elytron/JdbcRealmDefinition.java
bstansberry: That said, my impression reading this is that it's not necessary to call this in Stage.MODEL. This entire try/catch can go away. Looking (quickly) at all the impls of this interface, none of them do anything that is relevant to Stage.MODEL.

There's separate logic called in populateModel that validates the individual mapper configs, so the various things done in the toPasswordKeyMapper impls is not a necessary part of the Stage.MODEL validation. All that's needed here is checking that there is only config element.

pedro-hos: Hello Brian! Thank you for your revision. You mean, that I need do this validation without using the `toPassword...` method?

elytron/src/main/java/org/wildfly/extension/elytron/JdbcRealmDefinition.java
bstansberry: Unless @fjuma disagrees, +1 to getting rid of 'throws InvalidKeyException'. None of the impls throw it.

If we keep it we should fix the i18 of this OFE. Exceptions should not have hard coded English messages.

pedro-hos:
@bstansberry I have made the changes to check that we have only one mapper by `principal-query` during the Model stage.   I keep the `InvalidKeyException` so far, until we decide to keep or remove it.

elytron/src/main/java/org/wildfly/extension/elytron/JdbcRealmDefinition.java
fjuma: +1 for getting rid of the `throws InvalidKeyException`.

pedro-hos:
@bstansberry / @fjuma I have made the changes. Please, let me know if I need to make any other improvements. Thanks

elytron/src/main/java/org/wildfly/extension/elytron/JdbcRealmDefinition.java
```diff
@@ -653,6 +661,28 @@ private AttributeMapper[] resolveAttributeMappers(OperationContext context, Mode
         }
     }
 
+    private static class JdbcRealmDefinitionValidation implements OperationStepHandler {
+
+        @Override
+        public void execute(OperationContext context, ModelNode operation) throws OperationFailedException {
+            List<ModelNode> principalQueries = PrincipalQueryAttributes.PRINCIPAL_QUERIES_7_0.resolveModelAttribute(context, operation).asList();
```
bstansberry, line 668: @pedro-hos Unfortunately 'resolveModelAttribute' shouldn't be called in Stage.MODEL. Expressions meant to be resolved from secure sources (i.e. Elytron credential stores) can't be resolved in MODEL. We don't ban their use in MODEL (i.e. throw an exception) because there are a couple valid cases, but this isn't one. (The odds of such an expression being used here are probably minuscule, but we shouldn't rely on that.)

I think the thing to do is:

ModelNode model = context.readResource(PathAddress.EMPTY_ADDRESS, false).getModel();
ModelNode queries = model.get(PrincipalQueryAttributes.PRINCIPAL_QUERIES_7_0.getName());

And then inspect 'queries'.

The 'ModelNode model = context.readResource(PathAddress.EMPTY_ADDRESS, false).getModel();' is because by the time this runs the 'operation' contents have already been processed and stored into the model. It is better to use the results of that work (which may have somehow transformed things) that to rely on the original operation.

bstansberry (changes requested):

elytron/src/main/java/org/wildfly/extension/elytron/JdbcRealmDefinition.java
pedro-hos: Sure, I have made the changes. Thank you for the feedback

bstansberry (approved):
LGTM. Thanks @pedro-hos

pedro-hos:
Thank you

yersan:
Thanks @fjuma @pedro-hos @bstansberry

PR #6179, opened 2024-09-18
[WFCORE-6935] Improve systemd service units

yersan:
Jira issue: https://issues.redhat.com/browse/WFCORE-6935

### DIFF
--- a/core-feature-pack/common/src/main/resources/content/bin/systemd/README
+++ b/core-feature-pack/common/src/main/resources/content/bin/systemd/README
@@ -1,0 +1,81 @@
+= How to configure WildFly as a systemd service
+
+The following steps describe the process to configure WildFly as a systemd service.
+You can use the provided wildfly-[standalone,domain].service file as a template for your own systemd unit file.
+It can be adjusted to your needs, for example, you can manually change the user that runs the service,
+the location of the WildFly installation, logs, etc.
+Alternatively, you can use the generate_systemd_unit.sh script to automatically generate a new systemd unit file
+using this server installation as the WildFly home.
+
+If you want to install WildFly as a systemd service from scratch as a systemd service, follow the steps below.
+
+== Install WildFly and initialize the WILDFLY_HOME variable
+
+    # unzip -d /opt wildfly-34.0.0.Final.zip
+    # ln -s /opt/wildfly-34.0.0.Final /opt/wildfly
+    # WILDFLY_HOME=/opt/wildfly
+
+== Standalone or Domain mode ?
+
+   # MODE=standalone
+
+== Create a wildfly user and group
+
+    # groupadd -r wildfly
+    # useradd -r -g wildfly -d "${WILDFLY_HOME}" -s /sbin/nologin wildfly
+
+== Configure systemd
+
+    # cd "${WILDFLY_HOME}/bin/systemd"
+    # cp "wildfly-${MODE}.conf" /etc/sysconfig/
+    # cp "wildfly-${MODE}.service" $(pkg-config systemd --variable=systemdsystemunitdir)
+    # chown -R wildfly:wildfly "${WILDFLY_HOME}"/
+    # chmod +x "${WILDFLY_HOME}/bin/systemd/launch.sh"
+    # systemctl daemon-reload
+
+=== If SELinux is enabled, you need to set the appropriate context for the launch.sh script:
+
+    # sudo chcon -R -t bin_t "${WILDFLY_HOME}/"
+    # sudo semanage fcontext -a -t systemd_unit_file_t '${WILDFLY_HOME}(/.*)?'
+
+== Start and enable WildFly
+
+    # systemctl start "wildfly-${MODE}.service"
+    # systemctl enable "wildfly-${MODE}.service"
+
+== Check the status of WildFly
+
+    # systemctl status "wildfly-${MODE}.service"
+
+
+
+= How to remove WildFly as a systemd service
+
+== Standalone or Domain mode ?
+
+   # MODE=standalone
+
+== Stop and disable WildFly
+
+    # systemctl stop "wildfly-${MODE}.service"
+    # systemctl disable "wildfly-${MODE}.service"
+
+== If you are using SELinux, remove the custom context added
+
+    sudo semanage fcontext -d '${WILDFLY_HOME}(/.*)?'
+
+== Remove WildFly systemd service
+
+    # rm -f "$(pkg-config systemd --variable=systemdsystemunitdir)/wildfly-${MODE}.service"
+    # rm -f "/etc/sysconfig/wildfly-${MODE}.conf"
+    # systemctl daemon-reload
+
+== Remove WildFly installation
+
+    # rm -rf $(readlink "${WILDFLY_HOME}")
+    # rm -rf "${WILDFLY_HOME}"
+
+== Remove WildFly user and group
+
+    # userdel wildfly
+    # groupdel wildfly
--- a/core-feature-pack/common/src/main/resources/content/bin/systemd/generate_systemd_unit.sh
+++ b/core-feature-pack/common/src/main/resources/content/bin/systemd/generate_systemd_unit.sh
@@ -1,0 +1,108 @@
+#!/bin/sh
+
+# This script is just a helper to generate the systemd unit files for WildFly
+# assuming that the server home will be the one that hold this script file.
+# It also allows to specify the user and group that will run the server.
+
+function print_usage() {
+  echo "INFO: Usage: "
+  echo "${SCRIPT_NAME} -m,--mode standalone|domain [-u,--user user] [-g,--group group] [-c,--confirm y|n]"
+
+  exit 1
+}
+
+SCRIPT_NAME=$(basename "${0}")
+SCRIPT_DIR=$(dirname $(realpath "${0}"))
+JBOSS_HOME=$(realpath "${SCRIPT_DIR}/../..")
+JBOSS_SYSTEMD_MODE="${1}"
+
+JBOSS_SYSTEMD_MODE=""
+JBOSS_SYSTEMD_USER="wildfly"
+JBOSS_SYSTEMD_GROUP="wildfly"
+JBOSS_CONFIRM="n"
+
+while [ "${#}" -gt 0 ]
+do
+# For each argument option, checks whether the argument length is zero or starts with hyphen(s)
+    case "${1}" in
+        -m|--mode)
+            if [ -z "${2}" ] || [ ! "${2}" = `echo "${2}" | sed 's/^-*//'` ]; then
+                echo "Error: Missing argument for ${1}" >&2
+                print_usage
+            fi
+            JBOSS_SYSTEMD_MODE="${2}"
+            shift 2
+            ;;
+        -u|--user)
+            if [ -z "${2}" ] || [ ! "${2}" = `echo "${2}" | sed 's/^-*//'` ]; then
+                echo "Error: Missing argument for ${1}" >&2
+                print_usage
+            fi
+            JBOSS_SYSTEMD_USER="${2}"
+            shift 2
+            ;;
+        -g|--group)
+            if [ -z "${2}" ] || [ ! "${2}" = `echo "${2}" | sed 's/^-*//'` ]; then
+                echo "Error: Missing argument for ${1}" >&2
+                print_usage
+            fi
+            JBOSS_SYSTEMD_GROUP="${2}"
+            shift 2
+            ;;
+        -c|--confirm)
+            if [ -z "${2}" ] || [ ! "${2}" = `echo "${2}" | sed 's/^-*//'` ]; then
+                echo "Error: Missing argument for ${1}" >&2
+                print_usage
+            fi
+            JBOSS_CONFIRM="${2}"
+            shift 2
+            ;;
+        *)
+            echo "Error: Unknown argument ${1}" >&2
+            print_usage
+            ;;
+    esac
+done
+
+if [ "${JBOSS_SYSTEMD_MODE}" != 'standalone' ] && [ "${JBOSS_SYSTEMD_MODE}" != 'domain' ]; then
+  print_usage
+fi
+
+SYSTEMD_FILE="${SCRIPT_DIR}/wildfly-${JBOSS_SYSTEMD_MODE}.service"
+
+echo "INFO: systemd unit file to generate: ${SYSTEMD_FILE}"
+echo "INFO: Using JBOSS_HOME: ${JBOSS_HOME}"
+echo "INFO: User: ${JBOSS_SYSTEMD_USER}"
+echo "INFO: Group: ${JBOSS_SYSTEMD_GROUP}"
+
+if [ 'y' != "${JBOSS_CONFIRM}" ] && [ 'Y' != "${JBOSS_CONFIRM}" ];then
+  while true; do
+    read -p "Do you want to generate ${SYSTEMD_FILE}? (y/n): " choice
+    case "${choice}" in
+        y|Y ) break;;
+        n|N ) echo "Operation cancelled."; exit 1;;
+        * ) ;;
+    esac
+  done
+fi
+
+sed -e "s|@@@JBOSS_SYSTEMD_SERVER_HOME@@@|${JBOSS_HOME}|g" \
+    -e "s|@@@JBOSS_SYSTEMD_USER@@@|${JBOSS_SYSTEMD_USER}|g" \
+    -e "s|@@@JBOSS_SYSTEMD_GROUP@@@|${JBOSS_SYSTEMD_GROUP}|g" \
+    "${SYSTEMD_FILE}.template" > "${SYSTEMD_FILE}"
+
+systemd-analyze verify --recursive-errors=no "${SYSTEMD_FILE}"
+
+echo ""
+echo "INFO: systemd unit file generated."
+echo "INFO: The ${JBOSS_SYSTEMD_USER}:${JBOSS_SYSTEMD_GROUP} are the user:group configured to launch the server. You have to ensure this user and group exist and have the necessary permissions to read and launch the server."
+echo "INFO: Use ${JBOSS_HOME}/bin/systemd/wildfly-${JBOSS_SYSTEMD_MODE}.conf to configure the server environment."
+echo "INFO: After configuring your environment, to install the server as a systemd service do the following:"
+echo ""
+echo "sudo cp ${SYSTEMD_FILE} $(pkg-config systemd --variable=systemdsystemunitdir)"
+echo "sudo cp ${JBOSS_HOME}/bin/systemd/wildfly-${JBOSS_SYSTEMD_MODE}.conf /etc/sysconfig"
+echo "sudo systemctl daemon-reload"
+echo "sudo systemctl start $(basename "${SYSTEMD_FILE}")"
+echo ""
+echo "INFO: In case of issues, you can check the service logs with:"
+echo "sudo journalctl -u $(basename "${SYSTEMD_FILE}")"
--- a/core-feature-pack/common/src/main/resources/content/bin/systemd/launch.sh
+++ b/core-feature-pack/common/src/main/resources/content/bin/systemd/launch.sh
@@ -1,0 +1,26 @@
+#!/bin/sh
+if [ "${WILDFLY_SYSTEMD_DEBUG}" == "true" ]; then
+    set -x
+fi
+
+echo "INFO: Systemd Unit File server launch script"
+
+# Disable color output for the standard output
+export JAVA_OPTS="${JAVA_OPTS} -Dorg.jboss.logmanager.nocolor=true"
+export PATH="${JAVA_HOME}/bin:${PATH}"
+
+logDir=$(dirname "${WILDFLY_CONSOLE_LOG}")
+if [ ! -d "${logDir}" ]; then
+    mkdir -p "${logDir}"
+fi
+
+wildflyOpts="${WILDFLY_OPTS}"
+if [ -n "${WILDFLY_SUSPEND_TIMEOUT}" ]; then
+   wildflyOpts="-Dorg.wildfly.sigterm.suspend.timeout=${WILDFLY_SUSPEND_TIMEOUT} ${wildflyOpts}"
+fi
+
+if [ -z "${WILDFLY_HOST_CONFIG}" ]; then
+   "${WILDFLY_SH}" -c "${WILDFLY_SERVER_CONFIG}" -b "${WILDFLY_BIND}" ${wildflyOpts} > "${WILDFLY_CONSOLE_LOG}" 2>&1
+else
+   "${WILDFLY_SH}" --host-config="${WILDFLY_HOST_CONFIG}" -c "${WILDFLY_SERVER_CONFIG}" -b "${WILDFLY_BIND}" ${wildflyOpts} > "${WILDFLY_CONSOLE_LOG}" 2>&1
+fi
--- a/core-feature-pack/common/src/main/resources/content/bin/systemd/wildfly-domain.service
+++ b/core-feature-pack/common/src/main/resources/content/bin/systemd/wildfly-domain.service
@@ -1,0 +1,24 @@
+[Unit]
+Description=WildFly (domain mode)
+After=syslog.target network.target
+Before=httpd.service
+
+[Service]
+Environment=LAUNCH_JBOSS_IN_BACKGROUND=1
+Environment="JAVA_HOME=/usr/lib/jvm/jre"
+
+Environment="WILDFLY_SH=/opt/wildfly/bin/domain.sh"
+Environment="WILDFLY_SERVER_CONFIG=domain.xml"
+Environment="WILDFLY_CONSOLE_LOG=/opt/wildfly/domain/log/systemd-console.log"
+Environment="WILDFLY_HOST_CONFIG="
+Environment="WILDFLY_BIND=0.0.0.0"
+EnvironmentFile=-/etc/sysconfig/wildfly-domain.conf
+
+User=wildfly
+Group=wildfly
+LimitNOFILE=102642
+ExecStart=/bin/sh -c "/opt/wildfly/bin/systemd/launch.sh"
+TimeoutSec=600
+
+[Install]
+WantedBy=multi-user.target
--- a/core-feature-pack/common/src/main/resources/content/bin/systemd/wildfly-domain.service.template
+++ b/core-feature-pack/common/src/main/resources/content/bin/systemd/wildfly-domain.service.template
@@ -1,0 +1,24 @@
+[Unit]
+Description=WildFly (domain mode)
+After=syslog.target network.target
+Before=httpd.service
+
+[Service]
+Environment=LAUNCH_JBOSS_IN_BACKGROUND=1
+Environment="JAVA_HOME=/usr/lib/jvm/jre"
+
+Environment="WILDFLY_SH=@@@JBOSS_SYSTEMD_SERVER_HOME@@@/bin/domain.sh"
+Environment="WILDFLY_SERVER_CONFIG=domain.xml"
+Environment="WILDFLY_CONSOLE_LOG=@@@JBOSS_SYSTEMD_SERVER_HOME@@@/domain/log/systemd-console.log"
+Environment="WILDFLY_HOST_CONFIG="
+Environment="WILDFLY_BIND=0.0.0.0"
+EnvironmentFile=-/etc/sysconfig/wildfly-domain.conf
+
+User=@@@JBOSS_SYSTEMD_USER@@@
+Group=@@@JBOSS_SYSTEMD_GROUP@@@
+LimitNOFILE=102642
+ExecStart=/bin/sh -c "@@@JBOSS_SYSTEMD_SERVER_HOME@@@/bin/systemd/launch.sh"
+TimeoutSec=600
+
+[Install]
+WantedBy=multi-user.target
--- a/core-feature-pack/common/src/main/resources/content/bin/systemd/wildfly-standalone.service
+++ b/core-feature-pack/common/src/main/resources/content/bin/systemd/wildfly-standalone.service
@@ -1,0 +1,23 @@
+[Unit]
+Description=WildFly (standalone mode)
+After=syslog.target network.target
+Before=httpd.service
+
+[Service]
+Environment=LAUNCH_JBOSS_IN_BACKGROUND=1
+Environment="JAVA_HOME=/usr/lib/jvm/jre"
+
+Environment="WILDFLY_SH=/opt/wildfly/bin/standalone.sh"
+Environment="WILDFLY_SERVER_CONFIG=standalone.xml"
+Environment="WILDFLY_CONSOLE_LOG=/opt/wildfly/standalone/log/systemd-console.log"
+Environment="WILDFLY_BIND=0.0.0.0"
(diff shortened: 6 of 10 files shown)


core-feature-pack/common/src/main/resources/content/bin/systemd/wildfly-domain.service
```diff
@@ -0,0 +1,23 @@
...
+[Service]
+Environment=LAUNCH_JBOSS_IN_BACKGROUND=1
+Environment="JAVA_HOME=/usr/lib/jvm/jre"
+
+Environment="WILDFLY_SH=/opt/wildfly/bin/domain.sh"
+Environment="WILDFLY_SERVER_CONFIG=domain.xml"
+Environment="WILDFLY_CONSOLE_LOG=/opt/wildfly/domain/log/systemd-console.log"
+Environment="WILDFLY_HOST_CONFIG="
+Environment="WILDFLY_BIND=0.0.0.0"
+EnvironmentFile=-/etc/sysconfig/wildfly-domain.conf
```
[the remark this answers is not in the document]
yersan: This is just to make `wildfly-domain.conf` an optional file since the Systemd unit is enough to make the server run. `wildfly-domain.conf` can be optionally used to replace the default values but it is not mandatory. A '-' prefix is not used to validate whether the content of the file is correct, it is just to validate if the file exists.

Notice the documentation link is: https://www.freedesktop.org/software/systemd/man/latest/systemd.exec.html#EnvironmentFile=

"""
The argument passed should be an absolute filename or wildcard expression, optionally prefixed with "-", which indicates that if the file does not exist, it will not be read and no error or warning message is logged.
"""

I have no strong opinions about this, right now it aligns with the RPM installations done by the product, so the question is more whether we want to make `wildfly-domain.conf` mandatory to start the server even if the system unit is enough to be able to start the server and you have not added any variable declaration.

One possible advantage of making it mandatory is that you will be sure from the beginning that the file you pretend to use for configuration is read. Otherwise, a weird effect is to be modifying it and see how your server starts but you don't see (or understand) why your environment configuration is not being applied. That could occur if you incorrectly place it in an incorrect place.

core-feature-pack/common/src/main/resources/content/bin/systemd/README
```diff
@@ -0,0 +1,80 @@
...
+the location of the WildFly installation, logs, etc.
+Alternatively, you can use the generate_systemd_unit.sh script to automatically generate a new Systemd Unit File
+using this server installation as the WildFly home.
+
+If you want to install WildFly as a Systemd service from scratch as a Systemd service, follow the steps below.
+
+== Install WildFly
+
+    # unzip -d /opt wildfly-34.0.0.Final.zip
+    # ln -s /opt/wildfly-34.0.0.Final /opt/wildfly
```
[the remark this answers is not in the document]
yersan: I need to check but not sure if that would be fine taking into account that this file contains some other linux commands, I don't want to make them be replaced by maven applying environment properties across this file

yersan (commented):
Thanks for the review @jamezp , added some comments related

core-feature-pack/common/src/main/resources/content/bin/systemd/README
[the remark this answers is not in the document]
jamezp: That is a good point. I don't see any others with the pattern `${property.name}`, but it's definitely something to consider since that is the linux way.

I was mostly trying to think of a way where we don't have to keep updating the version, but maybe it doesn't matter.

core-feature-pack/common/src/main/resources/content/bin/systemd/wildfly-domain.service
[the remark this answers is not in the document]
jamezp: This makes sense to me. I was mostly curious on the intent, but you explained it.

jamezp (approved):
This looks good to me. We can figure out the version thing later and really maybe it doesn't make sense to keep it up to date. After all that is the full version and this is core.

core-feature-pack/common/src/main/resources/content/bin/systemd/README
```diff
@@ -0,0 +1,80 @@
...
+    # systemctl daemon-reload
+
+=== If SELinux is enabled, you need to set the appropriate context for the launch.sh script:
+
+    # sudo chcon -R -t bin_t /opt/wildfly/
+    # sudo semanage fcontext -a -t systemd_unit_file_t '/opt/wildfly(/.*)?'
+
+== Start and enable WildFly
+
+    # systemctl start wildfly-$MODE.service
```
rpelisse, line 42: To respect Bash's best practises:
`# systemctl start "wildfly-${MODE}.service"`

core-feature-pack/common/src/main/resources/content/bin/systemd/README
```diff
@@ -0,0 +1,80 @@
...
+   # MODE=standalone
+
+== Create a wildfly user and group
+
+    # groupadd -r wildfly
+    # useradd -r -g wildfly -d /opt/wildfly -s /sbin/nologin wildfly
+
+== Configure systemd
+
+    # cd /opt/wildfly/bin/systemd
```
rpelisse, line 28: To avoid hard coding /opt/wildfly everywhere, I would suggest defining a WILDFLY_HOME=/opt/wildfly at the beginning.

core-feature-pack/common/src/main/resources/content/bin/systemd/README
```diff
@@ -0,0 +1,80 @@
...
+    # cp wildfly-$MODE.conf /etc/sysconfig/
+    # cp wildfly-$MODE.service $(pkg-config systemd --variable=systemdsystemunitdir)
```
rpelisse, line 30: To follow bash's best practices:
```
    # cp "wildfly-${MODE}.conf" /etc/sysconfig/
    # cp "wildfly-${MODE}.service" $(pkg-config systemd --variable=systemdsystemunitdir)
```

core-feature-pack/common/src/main/resources/content/bin/systemd/README
```diff
@@ -0,0 +1,80 @@
...
+    # systemctl stop wildfly-$MODE.service
+    # systemctl disable wildfly-$MODE.service
+
+== If you are using SELinux, remove the custom context added
+
+    sudo semanage fcontext -d '/opt/wildfly(/.*)?'
+
+== Remove WildFly systemd service
+
+    # rm -f $(pkg-config systemd --variable=systemdsystemunitdir)/wildfly-$MODE.service
```
rpelisse, line 68: # rm -f "$(pkg-config systemd --variable=systemdsystemunitdir)/wildfly-${MODE}.service"

core-feature-pack/common/src/main/resources/content/bin/systemd/README
```diff
@@ -0,0 +1,80 @@
...
+    # rm -f /etc/sysconfig/wildfly-$MODE.conf
+    # systemctl daemon-reload
```
rpelisse, line 70: Should we not add systemctl status wildfy and systemctl start wildfly with an example of the output?

core-feature-pack/common/src/main/resources/content/bin/systemd/generate_systemd_unit.sh
```diff
@@ -0,0 +1,55 @@
...
+# This script is just a helper to generate the Systemd Unit Files for WildFly
+# assuming that the server home will be the one that hold this script file.
+# It also allows to specify the user and group that will run the server.
+
+SCRIPT_NAME=$(basename "$0")
+SCRIPT_DIR="$(dirname "$(realpath "$0")")"
+JBOSS_HOME="$(realpath "$SCRIPT_DIR/../..")"
+JBOSS_SYSTEMD_MODE="${1}"
+
+if [ -z "$JBOSS_SYSTEMD_MODE" ]; then
```
rpelisse, line 12: `if [ -z "${JBOSS_SYSTEMD_MODE}" ]; then`

for consistency on top of best practice

core-feature-pack/common/src/main/resources/content/bin/systemd/generate_systemd_unit.sh
```diff
@@ -0,0 +1,55 @@
...
+  echo "INFO: Usage: "
+  echo "${SCRIPT_NAME} standalone|domain [user name] [group name]"
+  exit 0;
+fi
+
+SYSTEMD_FILE="$SCRIPT_DIR/wildfly-${JBOSS_SYSTEMD_MODE}.service"
```
rpelisse, line 18: `SYSTEMD_FILE="${SCRIPT_DIR}/wildfly-${JBOSS_SYSTEMD_MODE}.service"`

core-feature-pack/common/src/main/resources/content/bin/systemd/generate_systemd_unit.sh
```diff
@@ -0,0 +1,55 @@
...
+JBOSS_SYSTEMD_USER="${2:-wildfly}"
+JBOSS_SYSTEMD_GROUP="${3:-wildfly}"
```
rpelisse, line 20: My preference is to add single quote around literals to make it clear to the reader that it's supposed to be fixed characters strings:
`JBOSS_SYSTEMD_GROUP="${3:-'wildfly'}"`

core-feature-pack/common/src/main/resources/content/bin/systemd/generate_systemd_unit.sh
```diff
@@ -0,0 +1,55 @@
...
+
+echo "INFO: Systemd Unit File to generate: ${SYSTEMD_FILE}"
+echo "INFO: Using JBOSS_HOME: $JBOSS_HOME"
```
rpelisse, line 23: As we have JBOSS_HOME here, we should use it at the beginning, to not repeat /opt/wildfly all the time.

Also: 
`echo "INFO: Using JBOSS_HOME: ${JBOSS_HOME}"`

core-feature-pack/common/src/main/resources/content/bin/systemd/generate_systemd_unit.sh
```diff
@@ -0,0 +1,55 @@
...
+echo "INFO: User: ${JBOSS_SYSTEMD_USER}"
+echo "INFO: Group: ${JBOSS_SYSTEMD_GROUP}"
+
+while true; do
```
rpelisse, line 27: Interactive shell is a bad idea. I would suggest using opts so that the user can pass a parameters if he wants to generate the systemd file.
We should also aimed at making this script idempotent. (I'll do another review in this regard, after this one)

core-feature-pack/common/src/main/resources/content/bin/systemd/generate_systemd_unit.sh
```diff
@@ -0,0 +1,55 @@
...
+  read -p "Do you want to generate ${SYSTEMD_FILE}? (y/n): " choice
+  case "$choice" in
+      y|Y ) break;;
+      n|N ) echo "Operation cancelled."; exit 1;;
+      * ) ;;
+  esac
+done
+
+sed -e "s|@@@JBOSS_SYSTEMD_SERVER_HOME@@@|$JBOSS_HOME|g" \
```
rpelisse, line 36:
```
sed -e "s|@@@JBOSS_SYSTEMD_SERVER_HOME@@@|${JBOSS_HOME}|g" \
    -e "s|@@@JBOSS_SYSTEMD_USER@@@|${JBOSS_SYSTEMD_USER}|g" \
    -e "s|@@@JBOSS_SYSTEMD_GROUP@@@|${JBOSS_SYSTEMD_GROUP}|g" \
    "${SYSTEMD_FILE}.template" > "${SYSTEMD_FILE}"
```

core-feature-pack/common/src/main/resources/content/bin/systemd/generate_systemd_unit.sh
```diff
@@ -0,0 +1,55 @@
...
+    -e "s|@@@JBOSS_SYSTEMD_USER@@@|$JBOSS_SYSTEMD_USER|g" \
+    -e "s|@@@JBOSS_SYSTEMD_GROUP@@@|$JBOSS_SYSTEMD_GROUP|g" \
+    "${SYSTEMD_FILE}.template" > "$SYSTEMD_FILE"
+
+systemd-analyze verify --recursive-errors=no "$SYSTEMD_FILE"
+
+echo ""
+echo "INFO: Systemd Unit File generated."
+echo "INFO: The ${JBOSS_SYSTEMD_USER}:${JBOSS_SYSTEMD_GROUP} are the user:group configured to launch the server. You have to ensure this user and group exist and have the necessary permissions to read and launch the server."
+echo "INFO: Use $JBOSS_HOME/bin/systemd/wildfly-${JBOSS_SYSTEMD_MODE}.conf to configure the server environment."
```
rpelisse, line 46: `echo "INFO: Use ${JBOSS_HOME}/bin/systemd/wildfly-${JBOSS_SYSTEMD_MODE}.conf to configure the server environment."`

core-feature-pack/common/src/main/resources/content/bin/systemd/generate_systemd_unit.sh
```diff
@@ -0,0 +1,55 @@
...
+echo "INFO: After configuring your environment, to install the server as a Systemd service do the following:"
+echo ""
+echo "sudo cp $SYSTEMD_FILE $(pkg-config systemd --variable=systemdsystemunitdir)"
+echo "sudo cp $JBOSS_HOME/bin/systemd/wildfly-${JBOSS_SYSTEMD_MODE}.conf /etc/sysconfig/wildfly-${JBOSS_SYSTEMD_MODE}.conf"
+echo "sudo systemctl daemon-reload"
+echo "sudo systemctl start $(basename $SYSTEMD_FILE)"
+echo ""
+echo "INFO: In case of issues, you can check the service logs with:"
+echo "sudo journalctl -u $(basename $SYSTEMD_FILE)"
```
rpelisse, line 55: Be mindful of where to use quote:
`echo sudo journalctl -u $(basename "${SYSTEMD_FILE}")`
This approach will yield a more comprehensible error message if, somehow, a space has been inserted in the variable.

core-feature-pack/common/src/main/resources/content/bin/systemd/launch.sh
```diff
@@ -0,0 +1,26 @@
+#!/bin/sh
```
rpelisse, line 1: I would suggest /bin/bash so that we can use bash specific mechanism. It also better because with /bin/sh we should test the script with all sort of shell interpreter.

core-feature-pack/common/src/main/resources/content/bin/systemd/README
yersan: Done

core-feature-pack/common/src/main/resources/content/bin/systemd/README
yersan: Done

core-feature-pack/common/src/main/resources/content/bin/systemd/README
yersan: Done

core-feature-pack/common/src/main/resources/content/bin/systemd/README
yersan: Done

core-feature-pack/common/src/main/resources/content/bin/systemd/README
yersan: At this point, this is the section that describes how to remove the service.

The start / enable / status commands are described in the previous section. There is no sample of status output, though

core-feature-pack/common/src/main/resources/content/bin/systemd/generate_systemd_unit.sh
yersan: Done

core-feature-pack/common/src/main/resources/content/bin/systemd/generate_systemd_unit.sh
yersan: That would move the quotes to the Systemd unit file as well, is that still valid?

core-feature-pack/common/src/main/resources/content/bin/systemd/generate_systemd_unit.sh
yersan: > As we have JBOSS_HOME here, we should use it at the beginning, to not repeat /opt/wildfly all the time.

JBOSS_HOME is initialized at the beginning of the script and there is no "/opt/wildfly" path hardcoded in this script, so I don't follow you here.

> Also:
> echo "INFO: Using JBOSS_HOME: ${JBOSS_HOME}"

Done

core-feature-pack/common/src/main/resources/content/bin/systemd/generate_systemd_unit.sh
yersan: > Interactive shell is a bad idea.

It is not expected this file be invoked by a third-party tool, this is just a utility that is expected to be used by a user to make his life easier when he needs to replace the JBOSS_HOME, user:group in the systemd unit files. Asking interactively for confirmation here allows the user to review what he is going to configure, so I think it is useful here.

Would that convince you to leave it as it is?

core-feature-pack/common/src/main/resources/content/bin/systemd/generate_systemd_unit.sh
yersan: Done

core-feature-pack/common/src/main/resources/content/bin/systemd/generate_systemd_unit.sh
```diff
@@ -0,0 +1,55 @@
...
+      * ) ;;
+  esac
+done
+
+sed -e "s|@@@JBOSS_SYSTEMD_SERVER_HOME@@@|$JBOSS_HOME|g" \
+    -e "s|@@@JBOSS_SYSTEMD_USER@@@|$JBOSS_SYSTEMD_USER|g" \
+    -e "s|@@@JBOSS_SYSTEMD_GROUP@@@|$JBOSS_SYSTEMD_GROUP|g" \
+    "${SYSTEMD_FILE}.template" > "$SYSTEMD_FILE"
+
+systemd-analyze verify --recursive-errors=no "$SYSTEMD_FILE"
```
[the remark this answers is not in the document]
yersan: Done

core-feature-pack/common/src/main/resources/content/bin/systemd/generate_systemd_unit.sh
yersan: Done

core-feature-pack/common/src/main/resources/content/bin/systemd/generate_systemd_unit.sh
yersan: Done

core-feature-pack/common/src/main/resources/content/bin/systemd/launch.sh
yersan: We use /sh to maximize the portability now and in the future, it is the same shell used by the launch scripts. Using bash opens the door to not be compatible in the future.

yersan (commented):
Thanks for the review @rpelisse , I fixed most of the suggestion and added some minor comments

core-feature-pack/common/src/main/resources/content/bin/systemd/launch.sh
rpelisse: Yeah, but do you ensure that CI test the script with different flavor of shell? Unlikely... By using /sh you are opening the door to people filling issue because it does not work with "not Bash". It's generally consider best practises to use explicit /bin/bash for this very reason. As this is a new script, I would take the chance to reduce the scope of incoming issue. That being said, it's not a big deal.

core-feature-pack/common/src/main/resources/content/bin/systemd/generate_systemd_unit.sh
rpelisse: No :) I really think it's a bad idea. 

Bear in mind that this will be used by Ops (or DevOps), they won't expect an interactive shell and will be quite annoyed with it. Also, they are most likely going to automate around this (I've done so in the past, at customers). They may want to automate the deployment of Wildfly based and this and have script (or a tool like Ansible) automatically generate the file on the target system.... 

And right now, it's impossible to because the script is interactive by default. So if you really want to keep it, at least you should provide a mecanism around it.

IMHO, the best approach, that would match what people do in this kind of script, would be to test if the var is defined and act according to it:

```
if  [ -z ${SYSTEMD_FILE} ]; then
   echo "No systemd file provided, not generating one."
   exit 0
fi
```

(last point if somehow you really, really want to keep this interaction thing, I would move into a function, because it's big chunk of code)

yersan: @rpelisse  Ok, you bring more experience to this from Ansible automation, so I'll remove the interactive question or add a confirm flag for skipping it

yersan: @rpelisse I've added a new argument c,--confirm to avoid the interactive mode.

yersan:
Hello @rpelisse could you review this again?

rpelisse (changes requested):
@yersan I'm super sorry to have not reviewed this PR (again) more quickly... I did find a couple of areas I think we should look into before we merge.

yersan (commented):
@rpelisse I'm done with this, if there is anything else critital, we would like to merge this as it is now, please take another look, thank you!

core-feature-pack/common/src/main/resources/content/bin/systemd/generate_systemd_unit.sh
```diff
@@ -0,0 +1,108 @@
...
+
+function print_usage() {
+  echo "INFO: Usage: "
+  echo "${SCRIPT_NAME} -m,--mode standalone|domain [-u,--user user] [-g,--group group] [-c,--confirm y|n]"
+
+  exit 1
+}
+
+SCRIPT_NAME=$(basename "${0}")
+SCRIPT_DIR="$(dirname "$(realpath "${0}")")"
```
rpelisse, line 15: I think this might be broken:

`"$(dirname "$(realpath "${0}")")"`

I count 2 opening " and 3 closing one. Also, with command substitution double quotes can be omitted (by spec the $() does the same job as ""). So this should be enough, as the most important "" are around the variable replacement, to ensure the result is treated as one arguments (so a whitespace in the file path does NOT break the script):

`$(dirname $(realpath "${0}"))`

core-feature-pack/common/src/main/resources/content/bin/systemd/generate_systemd_unit.sh
```diff
@@ -0,0 +1,109 @@
...
+
+JBOSS_SYSTEMD_MODE=""
+JBOSS_SYSTEMD_USER="wildfly"
+JBOSS_SYSTEMD_GROUP="wildfly"
+JBOSS_CONFIRM="n"
+
+while [ "$#" -gt 0 ]
+do
+# For each argument option, checks whether the argument length is zero or starts with hyphen(s)
+    case "$1" in
```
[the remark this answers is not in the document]
rpelisse: Somehow, I can't close this as resolved, but please do!

core-feature-pack/common/src/main/resources/content/bin/systemd/README
rpelisse: Somehow, I can't close this as resolved, but please do!

core-feature-pack/common/src/main/resources/content/bin/systemd/generate_systemd_unit.sh
```diff
@@ -0,0 +1,108 @@
...
+SCRIPT_DIR="$(dirname "$(realpath "${0}")")"
+JBOSS_HOME="$(realpath "${SCRIPT_DIR}/../..")"
+JBOSS_SYSTEMD_MODE="${1}"
+
+JBOSS_SYSTEMD_MODE=""
+JBOSS_SYSTEMD_USER="wildfly"
+JBOSS_SYSTEMD_GROUP="wildfly"
+JBOSS_CONFIRM="n"
+
+while [ "$#" -gt 0 ]
```
rpelisse, line 24: I did not saw it on the first pass, but all the variables in the args treatement are "$XXX" instead of "${XXX}". For consistency purpose and respecting best practices, it might be better to fix that.

core-feature-pack/common/src/main/resources/content/bin/systemd/generate_systemd_unit.sh
```diff
@@ -0,0 +1,109 @@
...
+    esac
+  done
+fi
+
+sed -e "s|@@@JBOSS_SYSTEMD_SERVER_HOME@@@|${JBOSS_HOME}|g" \
+    -e "s|@@@JBOSS_SYSTEMD_USER@@@|${JBOSS_SYSTEMD_USER}|g" \
+    -e "s|@@@JBOSS_SYSTEMD_GROUP@@@|${JBOSS_SYSTEMD_GROUP}|g" \
+    "${SYSTEMD_FILE}.template" > "${SYSTEMD_FILE}"
+
+systemd-analyze verify --recursive-errors=no "${SYSTEMD_FILE}"
```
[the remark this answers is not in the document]
rpelisse: Somehow, I can't close this as resolved, but please do!

core-feature-pack/common/src/main/resources/content/bin/systemd/launch.sh
rpelisse: Somehow, I can't close this as resolved, but please do!

core-feature-pack/common/src/main/resources/content/bin/systemd/generate_systemd_unit.sh
```diff
@@ -0,0 +1,109 @@
...
+echo "INFO: Use ${JBOSS_HOME}/bin/systemd/wildfly-${JBOSS_SYSTEMD_MODE}.conf to configure the server environment."
+echo "INFO: After configuring your environment, to install the server as a systemd service do the following:"
+echo ""
+echo "sudo cp ${SYSTEMD_FILE} $(pkg-config systemd --variable=systemdsystemunitdir)"
+echo "sudo cp ${JBOSS_HOME}/bin/systemd/wildfly-${JBOSS_SYSTEMD_MODE}.conf /etc/sysconfig"
+echo "sudo systemctl daemon-reload"
+echo "sudo systemctl start $(basename "${SYSTEMD_FILE}")"
+echo ""
+echo "INFO: In case of issues, you can check the service logs with:"
+echo "sudo journalctl -u $(basename "${SYSTEMD_FILE}")"
```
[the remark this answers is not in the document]
rpelisse: I think I did not get my point across, but that's fine, it's not a big deal. We can fine tuned that later. I may simply do a PR with my changes and we can discuss directly. For now, good enough, feel free to close it!

core-feature-pack/common/src/main/resources/content/bin/systemd/launch.sh
```diff
@@ -0,0 +1,26 @@
...
+if [ "$WILDFLY_SYSTEMD_DEBUG" == "true" ]; then
```
rpelisse, line 2: Another missing "${" that I failed to spot the first time :(

core-feature-pack/common/src/main/resources/content/bin/systemd/wildfly-domain.service.template
```diff
@@ -0,0 +1,23 @@
...
+Description=WildFly (domain mode)
+After=syslog.target network.target
+Before=httpd.service
+
+[Service]
+Environment=LAUNCH_JBOSS_IN_BACKGROUND=1
+Environment="JAVA_HOME=/usr/lib/jvm/jre"
+
+Environment="WILDFLY_SH=@@@JBOSS_SYSTEMD_SERVER_HOME@@@/bin/domain.sh"
+Environment="WILDFLY_SERVER_CONFIG=domain.xml"
```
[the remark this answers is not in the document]
rpelisse: Yeah, we just need to be careful to not update only ONE of them whenever there is a fix to do in them. You can close!

core-feature-pack/common/src/main/resources/content/bin/systemd/wildfly-standalone.service
```diff
@@ -0,0 +1,23 @@
...
+[Service]
+Environment=LAUNCH_JBOSS_IN_BACKGROUND=1
+Environment="JAVA_HOME=/usr/lib/jvm/jre"
+
+Environment="WILDFLY_SH=/opt/wildfly/bin/standalone.sh"
+Environment="WILDFLY_SERVER_CONFIG=standalone.xml"
+Environment="WILDFLY_CONSOLE_LOG=/opt/wildfly/standalone/log/systemd-console.log"
+Environment="WILDFLY_BIND=0.0.0.0"
+EnvironmentFile=-/etc/sysconfig/wildfly-standalone.conf
+
```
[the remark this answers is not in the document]
rpelisse: It's more to be aligned with the default we picked for the Ansible collection, but you are right, it's fine! You can close this!

core-feature-pack/common/src/main/resources/content/bin/systemd/wildfly-standalone.service.template
```diff
@@ -0,0 +1,22 @@
...
+
+[Service]
+Environment=LAUNCH_JBOSS_IN_BACKGROUND=1
+Environment="JAVA_HOME=/usr/lib/jvm/jre"
+
+Environment="WILDFLY_SH=@@@JBOSS_SYSTEMD_SERVER_HOME@@@/bin/standalone.sh"
+Environment="WILDFLY_SERVER_CONFIG=standalone.xml"
+Environment="WILDFLY_CONSOLE_LOG=@@@JBOSS_SYSTEMD_SERVER_HOME@@@/standalone/log/systemd-console.log"
+Environment="WILDFLY_BIND=0.0.0.0"
+EnvironmentFile=-/etc/sysconfig/wildfly-standalone.conf
```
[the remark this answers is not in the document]
rpelisse: Somehow, I can't close this as resolved, but please do!

rpelisse (changes requested):
I think I spot a potential bug, that's why I'm putting this in "Request Changes". I can run shellcheck right now, but I suspect if you do, it will confirm the issue I found. Apart from that (and a couple missing "${xxx}" around variables), I think we're good!

core-feature-pack/common/src/main/resources/content/bin/systemd/generate_systemd_unit.sh
yersan: Done

core-feature-pack/common/src/main/resources/content/bin/systemd/generate_systemd_unit.sh
yersan: Done

core-feature-pack/common/src/main/resources/content/bin/systemd/launch.sh
yersan: Done

yersan (commented):
@rpelisse I've followed your latest review, Could you check again and approve if you are fine? thansk!

rpelisse (approved):
LGTM!

yersan:
Thanks @rpelisse and @jamezp

PR #6180, opened 2024-09-18
[DRAFT] Shaded jar handling

jfdenise:
For testing purpose.
Relies on a released SNAPSHOT of galleon-plugins: https://repository.jboss.org/nexus/content/repositories/snapshots/org/wildfly/galleon-plugins/wildfly-galleon-plugins/7.1.3.Final-SNAPSHOT/


jfdenise:
The integration tests are expected to fail. We would need to synchronize changes in WildFly that are required: https://github.com/wildfly/wildfly/compare/main...jfdenise:wildfly:shaded-jar-wildfly

A custom integration (with core and full updated) CI job show that tests are passing : https://ci.wildfly.org/buildConfiguration/WF_WildFlyCoreIntegrationExperiments/457864 (The test failures are due to the wildfly-galleon-plugins being a SNAPSHOT (glow maven config has snapshot disabled). Local tests are passing).

@bstansberry , the wildfly-cli:client shaded jar is built as before, it is just not used when provisioning it in the WildFly installation. It is still referenced in the bom: 

The wildfly-elytron-tool-wrapper.jar is still built but is no more a shaded jar. It only contains the classes it defines. The shaded jar is built at provisioning time.

jfdenise:
This is superseded by https://github.com/wildfly/wildfly-core/pull/6213

PR #6199, opened 2024-09-23
[WFCORE-7008] Split ManagementInterfaceResourcesTestCase for different stability levels.

darranl:
https://issues.redhat.com/browse/WFCORE-7008

### DIFF
--- a/testsuite/manualmode/src/test/java/org/wildfly/core/test/standalone/mgmt/AbstractManagementInterfaceResourcesTestCase.java
+++ b/testsuite/manualmode/src/test/java/org/wildfly/core/test/standalone/mgmt/AbstractManagementInterfaceResourcesTestCase.java
@@ -1,0 +1,127 @@
+/*
+ * Copyright The WildFly Authors
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+package org.wildfly.core.test.standalone.mgmt;
+
+import static org.jboss.as.test.shared.TimeoutUtil.adjust;
+import static org.junit.Assert.assertTrue;
+
+import java.io.IOException;
+import java.net.InetSocketAddress;
+import java.net.Socket;
+import java.net.SocketAddress;
+import java.util.logging.Level;
+import java.util.logging.Logger;
+
+import org.jboss.as.test.shared.TestSuiteEnvironment;
+import org.junit.Test;
+import org.junit.runner.RunWith;
+import org.wildfly.common.function.ExceptionRunnable;
+import org.wildfly.core.testrunner.ServerControl;
+import org.wildfly.core.testrunner.ServerController;
+import org.wildfly.core.testrunner.WildFlyRunner;
+
+import jakarta.inject.Inject;
+
+/**
+ * Test case to test resource limits and clean up of management interface connections.
+ *
+ * @author <a href="mailto:darran.lofthouse@jboss.com">Darran Lofthouse</a>
+ */
+@RunWith(WildFlyRunner.class)
+@ServerControl(manual = true)
+public abstract class AbstractManagementInterfaceResourcesTestCase {
+    protected static final Logger LOG = Logger.getLogger(AbstractManagementInterfaceResourcesTestCase.class.getName());
+
+
+    @Inject
+    protected static ServerController controller;
+
+    /**
+     * Test that the management interface will not accept new connections when the number of active connections reaches the
+     * high water mark.  After the number of open connections has been reduced to the low watermark it will test that connections
+     * are accepted again.
+     */
+    @Test
+    public void testWatermarks() throws Exception {
+        runTest(60000, () -> {
+            String mgmtAddress = TestSuiteEnvironment.getServerAddress();
+            int mgmtPort = TestSuiteEnvironment.getServerPort();
+            LOG.info(mgmtAddress + ":" + mgmtPort);
+            SocketAddress targetAddress = new InetSocketAddress(mgmtAddress, mgmtPort);
+
+            int socketsOpened = 0;
+            boolean oneFailed = false;
+            Socket[] sockets = new Socket[9];
+            for (int i = 0 ; i < 9 ; i++) {
+                LOG.info("Opening socket " + i + " socketsOpened=" + socketsOpened);
+                try {
+                    sockets[i] = new Socket();
+                    sockets[i].connect(targetAddress, 5000);
+                    socketsOpened++;
+                } catch (IOException e) {
+                    LOG.log(Level.SEVERE, "Probably an expected exception trying to open a new connection", e);
+                    assertTrue("Less sockets than low watermark opened.", socketsOpened > 3);
+                    oneFailed = true;
+                }
+            }
+            assertTrue("Opening of one socket was expected to fail.", oneFailed);
+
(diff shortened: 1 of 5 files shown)


darranl:
All tests run locally on Linux, just added a test that does nothing to check if the setup and cleanup works.

kabir (approved):

testsuite/manualmode/src/test/java/org/wildfly/core/test/standalone/mgmt/ManagementInterfaceResourcesCommunityTestCase.java
```diff
@@ -0,0 +1,81 @@
...
+    protected void runTest(int noRequestTimeout, ExceptionRunnable<Exception> test) throws Exception {
+        controller.start();
+        ManagementClient client = controller.getClient();
+
+        ServerSetupTask task = new ManagementInterfaceSetUpTask(noRequestTimeout);
+
+        try {
+            task.setup(client);
+            test.run();
+            controller.reload();
```
yersan, line 48: This reload here seems unnecessary, doesn't it?

testsuite/manualmode/src/test/java/org/wildfly/core/test/standalone/mgmt/ManagementInterfaceResourcesCommunityTestCase.java
```diff
@@ -0,0 +1,81 @@
...
+import org.wildfly.test.stability.StabilityServerSetupSnapshotRestoreTasks;
+
+/**
+ * Test case to test resource limits and clean up of management interface connections.
+ *
+ * This test case uses attributes defined directly on the HTTP management interface resource for configuration.
+ *
+ * @author <a href="mailto:darran.lofthouse@jboss.com">Darran Lofthouse</a>
+ */
+public class ManagementInterfaceResourcesCommunityTestCase extends AbstractManagementInterfaceResourcesTestCase {
```
yersan, line 27: Should not this test be excluded from the Bootable JAR set? I guess saving a snapshot in a Bootable JAR would not be a valid operation

testsuite/manualmode/src/test/java/org/wildfly/core/test/standalone/mgmt/AbstractManagementInterfaceResourcesTestCase.java
```diff
@@ -0,0 +1,132 @@
...
+
+/**
+ * Test case to test resource limits and clean up of management interface connections.
+ *
+ * @author <a href="mailto:darran.lofthouse@jboss.com">Darran Lofthouse</a>
+ */
+@RunWith(WildFlyRunner.class)
+@ServerControl(manual = true)
+public abstract class AbstractManagementInterfaceResourcesTestCase {
+    protected static final Logger LOG = Logger.getLogger(ManagementInterfaceResourcesTestCase.class.getName());
```
yersan, line 36: One minor, this log should be in the child's classes instead of the parent's.

yersan (changes requested):

testsuite/manualmode/src/test/java/org/wildfly/core/test/standalone/mgmt/ManagementInterfaceResourcesCommunityTestCase.java
darranl: I am not sure at the moment, the original tests also had this reload.

The problem with this test is we deliberately exhaust the http management interface connection to test that this configuration is working and then we try and use the connection to perform further updates.

testsuite/manualmode/src/test/java/org/wildfly/core/test/standalone/mgmt/ManagementInterfaceResourcesCommunityTestCase.java
darranl: TBH I am not sure, as the process is only running once it would make sense if we could run bootable jar tests otherwise we will never be able to test features with a stability level lower than default.

testsuite/manualmode/src/test/java/org/wildfly/core/test/standalone/mgmt/AbstractManagementInterfaceResourcesTestCase.java
darranl: +1

yersan:
Weird, Linux - Security Manager - JDK 11 has failed twice on the test developed for this PR

testsuite/manualmode/src/test/java/org/wildfly/core/test/standalone/mgmt/AbstractManagementInterfaceResourcesTestCase.java
darranl: Fixed

yersan: Ok, I was expecting a log on each child class so it can be easier to track which one is being executed, but in case of a stack trace, we can inferred the origin class from the stack, so I am fine with whatever you prefer

yersan:
I am still getting intermittent failures with these tests running locally, I am not sure why, so let's see how the CI behaves

darranl:
It does look to be something security manager related quite consistently failing this test.

yersan:
@darranl  I wonder if somehow not all the sockets, or at least a number that would allow open a new one to properly do the reload, could be the issue. We close them from the point of view of one pair, but we proceed to the reload without being sure they have been fully closed on the other pair (server side), just an idea in case it sounds your bell

darranl:
Being only security manager affected now feels like there is a specific target to debug.

The other kind of brute force option here would be to also define the native management interface on the server under test and use that for the server management ops so even if we exhaust the resources on the http management interface we can always access the native one for server control.

A big issue with this kind of test is it is also heavily dependent on the behaviour of the underlying TCP stack on the target system handling a lot of clean up.  Where resource constrained I don't know if the target is more or less likely to proactively clean up half closed sockets.

yersan:
@darranl 

>Where resource constrained I don't know if the target is more or less likely to proactively clean up half closed sockets.

I tried with `Socket.setSoLinger(true, 0)` without luck. In any case, they should be closed on the server side automatically when the defined timeout occurs, right? If we have exhausted the resources on the http management interface, they should eventually cleaned up once the timeout occurs. So why this could not be happening when we are sending a reload request? Why the connection already established for the client used to do the reload is not able to do the work? Would open a new client connection work in this case?

darranl:
I have just rebased on the off chance it may help.

yersan:
@darranl the test case is still failing, not sure why. Just for the record, the current alternative of checking whether the resource exists at the current stability level is enough to skip the test, so if trying to get it to work using the stability level setup task gets complicated, I think we can live with the current approach.

darranl:
I may have a look at if I can reproduce locally again but am getting close to giving up on this - my only other option is to introduce a native management interface for configuration to keep it separate from the one being tested as I think that is the underlying issue.

darranl:
@yersan I am going to close this one, this one has become way more problematic than it is worth.

PR #6202, opened 2024-09-26
[WFCORE-6959] test validating final endpoint name with different jboss.server.name and jboss.node.name configurations

simkam:
https://issues.redhat.com/browse/WFCORE-6959

I tried to run this test case with wildfly-core 25.0.2 and it failed due to [WFCORE-6956](https://issues.redhat.com/browse/WFCORE-6956) as expected. 

I find interesting that jboss.server.name configured in server configuration (`<server xmlns="urn:jboss:domain:20.0" name="custom-server-name">`) has precedence over configuration via `-Djboss.server.name` system property.


testsuite/manualmode/src/test/java/org/jboss/as/test/manualmode/management/JBossNodeNameTestCase.java
```diff
@@ -0,0 +1,191 @@
+/*
+ * Copyright The WildFly Authors
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+package org.jboss.as.test.manualmode.management;
+
+import static org.jboss.as.test.manualmode.expressions.module.TestSecureExpressionsExtension.PARAM_EXPRESSION;
```
yersan, line 8: I would say it is better to keep tests independent; `TestSecureExpressionsExtension` is unrelated to what we are doing here, so I would not use a constant defined in an external class, instead it would be better to define it in this class.

Do you agree?

yersan:
> I find interesting that jboss.server.name configured in server configuration (<server xmlns="urn:jboss:domain:20.0" name="custom-server-name">) has precedence over configuration via -Djboss.server.name system property.

I see what you mean, and I agree that the natural behavior would be the system property to override the model value. It has been working as it has since the beginning, so I'm not sure if it makes sense to change this behavior now, especially since we haven't heard any complaints (I guess the domain mode host controller name has the same issue and it cannot be overwritten via a System Property). However, I agree that it is not a natural behavior. CC @bstansberry , maybe he knows a reason for that, also it would be great to heard your opinion about this.

testsuite/manualmode/src/test/java/org/jboss/as/test/manualmode/management/JBossNodeNameTestCase.java
simkam: fixed

yersan (approved):

yersan:
@simkam thanks!

yersan:
@simkam In relation to the above discussion about the precedence order, I've just read this:

https://docs.wildfly.org/34/Admin_Guide.html#host-controller-configuration

""
If the name attribute is not set, the default name for the host will be the value of the jboss.host.name system property
""

So the value of the configuration file precedes the value of System property and is intentional. Management Operation wins versus a System Property.

PR #6201, opened 2024-09-26
[WFCORE-7010] Allow starting Domain Mode tests under 'default' stability…

yersan:
… level

Jira issue: https://issues.redhat.com/browse/WFCORE-7010

PR #6204, opened 2024-09-27
[WFCORE-7011] Use 'default' stability level when the remote host controller do not advertise its current stability level

yersan:
Built on top of https://github.com/wildfly/wildfly-core/pull/6201

Jira issue: https://issues.redhat.com/browse/WFCORE-7011

It is expected that mixed domain tests fail with this change. I opened it as a draft since this change still requires some discussions to decide whether we want to allow DC -- HC mixed domains under community -- default stability levels

### DIFF
--- a/version/src/main/java/org/jboss/as/version/ProductConfig.java
+++ b/version/src/main/java/org/jboss/as/version/ProductConfig.java
@@ -23,6 +23,7 @@ import java.util.Map;
 import java.util.Properties;
 import java.util.Set;
 import java.util.jar.Manifest;
+import java.util.stream.Collectors;
 
 import org.jboss.modules.Module;
 import org.jboss.modules.ModuleIdentifier;
@@ -150,8 +151,10 @@ public class ProductConfig implements Serializable {
         this.name = productName;
         this.version = productVersion;
         this.consoleSlot = consoleSlot;
-        this.defaultStability = Stability.DEFAULT;
-        this.stabilities = EnumSet.of(this.defaultStability);
+        this.defaultStability = Stability.COMMUNITY;
+        this.stabilities = EnumSet.allOf(Stability.class).stream()
+                .filter(this.defaultStability::enables)
+                .collect(Collectors.toUnmodifiableSet());
         this.banner = null;
     }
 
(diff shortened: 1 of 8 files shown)


yersan:
Just rebased on top of the current main. @pferraro Could you approve from your side? I would like to release this this week to unlock pending work with the mixed domains in upstream. This one will be merged as the last PR for the release since it breaks the TS until the WildFly counterpart gets merged

After that, we will continue adding a WildFly for the mixed domains and follow with https://github.com/wildfly/wildfly-core/pull/6214

pferraro:
Are you sure this branch is up to date?  The git diff looks odd.

yersan:
Hi @pferraro , it is now. We do not need to get the ProductConf stability level at the Host registration process, so most of the "weird" changes shown in the diff are to remove the need to pass that variable through all the method calls to get it at the registration phase.

Then, some adjustments on some tests because I moved the default stability level for the tests to Community, which apparently makes more sense. We can continue using DEFAULT and let each test choose whatever it needs. Maybe that makes more sense.

Still in draft since I have not gotten the full CI feedback yet, but great if you can take a look at it is

yersan:
@pferraro I think this is ready to go with your review, galleon full failure is unrelated

version/src/main/java/org/jboss/as/version/ProductConfig.java
```diff
@@ -150,8 +150,8 @@ public ProductConfig(final String productName, final String productVersion, fina
         this.name = productName;
         this.version = productVersion;
         this.consoleSlot = consoleSlot;
-        this.defaultStability = Stability.DEFAULT;
-        this.stabilities = EnumSet.of(this.defaultStability);
+        this.defaultStability = Stability.COMMUNITY;
+        this.stabilities =  Collections.unmodifiableSet(EnumSet.range(Stability.DEFAULT, Stability.COMMUNITY));
```
pferraro, line 154: I would keep this logic internal to the Stability object.
e.g.
```java
this.stabilities = EnumSet.allOf(Stability.class).stream().filter(this.defaultStability::enables).collect(Collectors.toUnmodifiableSet());
```

yersan: @pferraro thanks for reviewing, done

Since that was the only pending fix, I am going to move on with it so I can prepare a release today

yersan: Well, wait .... now I am wondering, I would expect to see the mixed domain integration tests failing with this PR in, since we will be assuming EAP 7.4 legacy host will be running in "default" stability level, hence the mixed domains should have failed because WildFly will be running by default in "community" stability level.

Looking ....

yersan: Ok, got it, I've realized today that full integration Jobs are ignoring the mixed domain tests because of a misconfiguration of those two Jobs (they are using JDK 17 but legacy JDK to launch the legacy servers is not configured). I'll fix them once we have integrated both this PR and the WildFly counterpart, so we can continue without breaking WildFly Core CI until the WildFly full gets merged.

yersan:
Thanks @pferraro

version/src/main/java/org/jboss/as/version/ProductConfig.java
yersan: WildFly Core integration Jobs have been fixed to include the expected JDK to be able to run the mixed domain tests

PR #6206, opened 2024-09-27
WFCORE-7004: use custom parser over default

michpetrov:
Issue: [WFCORE-7004](https://issues.redhat.com/browse/WFCORE-7004)

The default `parseElement` method should ideally never be accessed after this fix but in case it is I assume it's better to throw an exception rather then having one thrown by the XML reader because that makes it look like something's wrong with the XML file.


yersan:
@michpetrov Would it be possible to add a test case for the use case described in the Jira?

michpetrov:
@yersan you mean a test for the parser? Do we have something like that already? I'd rather not have to mock up a subsystem from scratch.

michpetrov:
@yersan added the test

yersan (approved):

yersan:
Thanks @michpetrov for the fix and adding a test case, CC @pferraro since you are now refactorizing this class

PR #6214, opened 2024-10-11
WFCORE-6995 Allow stability-specific resource transformations for mixed domains

pferraro:
https://issues.redhat.com/browse/WFCORE-6995

~~Submitting as a draft while I collect feedback and add tests.~~

@yersan Until https://github.com/wildfly/wildfly-legacy-test publishes releases for WildFly versions, it will be impossible to create model transformation tests for COMMUNITY stability.


controller/src/main/java/org/jboss/as/controller/transform/SubsystemModelTransformerRegistration.java
```diff
@@ -0,0 +1,52 @@
...
+
+    @Override
+    public void registerTransformers(SubsystemTransformerRegistration registration) {
+        // Build and register transformation descriptions for all but the current subsystem model version
+        for (E model : EnumSet.complementOf(EnumSet.of(this.currentSubsystemModel))) {
+            ModelVersion version = model.getVersion();
+            ResourceTransformationDescriptionBuilder builder = registration.createResourceTransformationDescriptionBuilder();
+            this.transformation.accept(builder, version);
+            TransformationDescription.Tools.register(builder.build(), registration, version);
+        }
```
yersan, line 50: Do we have an example of how to use this class?

I'm not a SME in these transformers and it is too big to digest it easily, but I guess what I cannot see is how this new way to register the transformers will be able to match what we currently have. At this moment, transformers are registered as a chain, with a common builder that registers the transformation from one version to another, however here we are only registering a single version with a specific builder, so I get lost on this.

controller/src/main/java/org/jboss/as/controller/transform/TransformerRegistry.java
```diff
@@ -275,11 +288,16 @@ public static class Factory {
          * Create a new Transformer registry.
          *
          * @return the created transformer registry
+         * @deprecated Superseded by {@link #create(Stability)}.
          */
+        @Deprecated(forRemoval = true)
         public static TransformerRegistry create() {
-            return new TransformerRegistry();
+            return create(Stability.DEFAULT);
         }
```
yersan, line 296: What happens with the current usage of this method in our code? Shouldn't we at least force the creation of a transformer registry using a specific Stability level? Otherwise, we will continue using the wrong registry and we won't be able yet to add specific transformations to the subsystems.

host-controller/src/main/java/org/jboss/as/domain/controller/transformers/DomainTransformers.java
```diff
@@ -66,11 +66,12 @@ public static void initializeDomainRegistry(final TransformerRegistry registry)
...
-        registerChainedManagementTransformers(registry);
-        registerChainedServerGroupTransformers(registry);
+        ChainedTransformationDescriptionBuilderFactory factory = KernelAPIVersion.createChainedTransformationDescriptionBuilderFactory(registry);
+        registerChainedManagementTransformers(registry, factory);
+        registerChainedServerGroupTransformers(registry, factory);
         registerProfileTransformers(registry);
-        registerSocketBindingGroupTransformers(registry);
-        registerDeploymentTransformers(registry);
+        registerSocketBindingGroupTransformers(registry, factory);
+        registerDeploymentTransformers(registry, factory);
```
yersan, line 74: Why we are only changing the Domain Transformers? Shouldn't we also change and adapt in this PR how we are registering the other transformers?

For example, at ElytronSubsystemTransformers we have:

```
@Override
```

What would be the equivalent code to create `ChainedTransformationDescriptionBuilder` using the current Stability level?

yersan (commented):
Just added comments, I do not have the expertise on transformers API to validate the work done here, it should be fine since the `BasicResourceTestCase` explicitly tests all of those transformations, but I got the impression there is still pending work to do to enable the server  to use the expected stability level when it is registering the transformers, that;s still a bit unclear to me

pferraro:
Bump (to prevent github autoclose)

pferraro:
Bump (to prevent github autoclose).  @bstansberry Are we still interested in this?  If so, can you take a look?

controller/src/main/java/org/jboss/as/controller/transform/description/AttributeTransformationDescriptionBuilderImpl.java
```diff
@@ -29,74 +30,72 @@ abstract class AttributeTransformationDescriptionBuilderImpl<T extends BaseAttri
...
+        return this.thisBuilder();
     }
 
     @Override
     public T addRejectCheck(RejectAttributeChecker rejectChecker, String... rejectedAttributes) {
-        for (String attribute : rejectedAttributes) {
-            registry.addAttributeCheck(attribute, rejectChecker);
+        for (String rejectedAttribute : rejectedAttributes) {
+            this.registry.addAttributeCheck(rejectedAttribute, rejectChecker);
         }
```
bstansberry, line 74: Should this method be deprecated? Just reading the diff it seems like it's unable to apply the desired 'enables' check so it's just assuming. Which is ok temporarily but perhaps should be deprecated.

It seems like the javadoc in the interface should explain the behavior re the 'enables' call for all the methods.

controller/src/main/java/org/jboss/as/controller/transform/description/AttributeTransformationDescriptionBuilderImpl.java
```diff
@@ -29,74 +30,72 @@ abstract class AttributeTransformationDescriptionBuilderImpl<T extends BaseAttri
...
+            this.addRejectCheck(rejectChecker, rejectedAttributes);
         }
         return thisBuilder();
     }
 
     @Override
     public T addRejectChecks(List<RejectAttributeChecker> rejectCheckers, String... rejectedAttributes) {
         for (RejectAttributeChecker rejectChecker : rejectCheckers) {
-            addRejectCheck(rejectChecker, rejectedAttributes);
+            this.addRejectCheck(rejectChecker, rejectedAttributes);
```
bstansberry, line 89: Same comment re deprecation.

controller/src/main/java/org/jboss/as/controller/transform/description/AttributeTransformationDescriptionBuilderImpl.java
```diff
@@ -106,19 +105,21 @@ public T addRename(String attributeName, String newName) {
         return thisBuilder();
     }
 
+    @Override
     public T addRenames(Map<String, String> renames) {
         for (Map.Entry<String, String> rename : renames.entrySet()) {
-            registry.addRenamedAttribute(rename.getKey(), rename.getValue());
+            this.addRename(rename.getKey(), rename.getValue());
```
bstansberry, line 111: Same comment re deprecation.

controller/src/main/java/org/jboss/as/controller/transform/description/AttributeTransformationDescriptionBuilderImpl.java
```diff
@@ -131,10 +132,6 @@ public T setValueConverter(AttributeConverter attributeConverter, String... conv
         return thisBuilder();
```
bstansberry, line 132: Same comment re deprecation.

controller/src/main/java/org/jboss/as/controller/transform/description/ResourceTransformationDescriptionBuilder.java
```diff
@@ -86,8 +114,13 @@ public interface ResourceTransformationDescriptionBuilder extends Transformation
      *
      * @param definition the resource definition
      * @return the builder for the child resource
+     * @deprecated Use {@link ResourceTransformationDescriptionBuilder#addChildResource(ResourceRegistration)} instead.
      */
-    ResourceTransformationDescriptionBuilder addChildResource(ResourceDefinition definition);
+    @Deprecated(forRemoval = true)
```
bstansberry, line 119: Note to reviewers: I only see 5 uses of this in WF, all in the legacy Picketbox extensions.

controller/src/main/java/org/jboss/as/controller/transform/description/ResourceTransformationDescriptionBuilder.java
```diff
@@ -97,25 +130,49 @@ public interface ResourceTransformationDescriptionBuilder extends Transformation
      * @param definition the resource definition
      * @param dynamicDiscardPolicy a checker to decide whether the child should be added or not
      * @return the builder for the child resource
+     * @deprecated Use {@link ResourceTransformationDescriptionBuilder#addChildResource(ResourceRegistration, DynamicDiscardPolicy)} instead.
      */
-    ResourceTransformationDescriptionBuilder addChildResource(ResourceDefinition definition, DynamicDiscardPolicy dynamicDiscardPolicy);
-
+    @Deprecated(forRemoval = true)
```
bstansberry, line 135: Note to reviewers: I don't see any uses of this in WF.

controller/src/main/java/org/jboss/as/controller/transform/description/ResourceTransformationDescriptionBuilder.java
```diff
@@ -165,14 +276,29 @@ public interface ResourceTransformationDescriptionBuilder extends Transformation
...
+    ResourceTransformationDescriptionBuilder addChildRedirection(ResourceRegistration registration, PathAddressTransformer pathAddressTransformer, DynamicDiscardPolicy dynamicDiscardPolicy);
 
     /**
      * Add an already created {@link TransformationDescriptionBuilder} as a child of this builder.
      *
      * @param builder the builder
      * @return the builder for this resource
+     * @deprecated Use other methods for creating child transformation descriptions
      */
+    @Deprecated(forRemoval = true)
```
bstansberry, line 301: Note to reviewers: I see no use in WF.

bstansberry (commented):
@pferraro I had a look and made comments. I think Yeray's earlier are the more critical ones.

Re whether we are interested in this, TBH I'm relying on you and @yersan to tell me. I *think* we are because in all the discussions about it we keep working on it and I think we'd stop if we decided not to do it. ;) But the details of the back-and-forth are flushed from memory, partly because I'm relying on you guys.

yersan:
> @pferraro I had a look and made comments. I think Yeray's earlier are the more critical ones.
> 
> Re whether we are interested in this, TBH I'm relying on you and @yersan to tell me. I _think_ we are because in all the discussions about it we keep working on it and I think we'd stop if we decided not to do it. ;) But the details of the back-and-forth are flushed from memory, partly because I'm relying on you guys.

This is something good to have, since it will allow us to test community-community paired stability levels in mixed domains. I don't see it as something urgent to do; I have never read issues with mixed domains on the community side, so, probably something that is not widely used. It brings up more maintenance, and the fix indeed looks complex. However, having those tests in practice is good and necessary. It will also help later to future promotions by moving a feature from community stability to a higher one.

pferraro:
Bump to keep alive.

controller/src/main/java/org/jboss/as/controller/transform/description/AttributeTransformationDescriptionBuilderImpl.java
bstansberry: #6589 adds the deprecation.

controller/src/main/java/org/jboss/as/controller/transform/description/AttributeTransformationDescriptionBuilderImpl.java
bstansberry: #6589 adds the deprecation.

controller/src/main/java/org/jboss/as/controller/transform/description/AttributeTransformationDescriptionBuilderImpl.java
bstansberry: #6589 adds the deprecation.

controller/src/main/java/org/jboss/as/controller/transform/description/AttributeTransformationDescriptionBuilderImpl.java
bstansberry: #6589 adds the deprecation.

controller/src/main/java/org/jboss/as/controller/transform/description/AttributeTransformationDescriptionBuilderImpl.java
pferraro: Thanks for that - I'll rebase this branch on that change next week.

controller/src/main/java/org/jboss/as/controller/transform/SubsystemModelTransformerRegistration.java
pferraro: > At this moment, transformers are registered as a chain...

That depends on the subsystem.
Defining resource transformations via chains is only necessary when the set of subsystem model versions is not well defined.
For subsystems that enumerate their model versions (via SubsystemModel), building transformation descriptions via chains is actually counter-productive.

By leveraging the SubsystemModel#requiresTransformation(ModelVersion) method, you can easily compose transformation descriptions based on the subsystem model version that introduced a given change.

For example, say I have a resource that added a new attribute in VERSION_2_0, and then changed the default value to a different value in VERSION_3_0.
My transformation registration method for this resource (invoked for each version of the subsystem model) might look like:

```java
static void registerTransformation(ResourceTransformationDescriptionBuilder parent, ModelVersion version) {
    ResourceTransformationDescriptionBuilder builder = parent.addChildResource(MyResourceDefinition.REGISTRATION.getPathElement());

    if (MySubsystemModel.VERSION_2_0.requiresTransformation(version)) {
        builder.getAttributeBuilder()
                .setDiscard(DiscardAttributeChecker.UNDEFINED, MyResourceDefinition.ATTRIBUTE)
                .addRejectChecker(RejectAttributeChecker.DEFINED, MyResourceDefinition.ATTRIBUTE)
                ;
    }
    if (MySubsystemModel.VERSION_3_0.requiresTransformation(version)) {
        builder.getAttributeBuilder().setValueConverter(AttributeConverter.DEFAULT_VALUE, MyResourceDefinition.ATTRIBUTE);
    }
}
```

When defined this way, the evolutionary history of a given resource becomes easy to comprehend.

By leveraging SubsystemModelTransformerRegistration, my ExtensionTransformerRegistration implementation becomes:

```java
public class MySubsystemModelTransformerRegistration extends ExtensionTransformerRegistration {
    public MySubsystemModelTransformerRegistration() {
        super(MySubsystemResourceDefinition.REGISTRATION, MySubsystemModel.CURRENT, MySubsystemResourceDefinition::registerTransformation);
    }
}
```

In contrast, the ExtensionTransformerRegistration implementation for same subsystem, with the same transformer registration method, but implemented using chains, would look like:

```java
public class MySubsystemExtensionTransformerRegistration implements ExtensionTransformerRegistration {
    
    @Override
    public String getSubsystemName() {
        return MySubsystemResourceDefinition.SUBSYSTEM_NAME;
    }

    @Override
        ChainedResourceTransformationDescriptionBuilder chainBuilder = registration.createChainedResourceTransformationDescriptionBuilder();
        MySubsystemResourceDefinition.registerTransformation(chainBuilder.createBuilder(VERSION_3_0, VERSION_2_0), VERSION_3_0);
        MySubsystemResourceDefinition.registerTransformation(chainBuilder.createBuilder(VERSION_2_0, VERSION_1_0), VERSION_2_0);
        chainedBuilder.buildAndRegister(registration, new ModelVersion[] { VERSION_3_0, VERSION_2_0, VERSION_1_0 });
    }
}
```

Both generate identical transformer descriptions.

controller/src/main/java/org/jboss/as/controller/transform/TransformerRegistry.java
pferraro: Ideally, there should not be any - other than in test code.
A quick peek indicates that this PR needs more updates, as this method is currently be used by ExtensionRegistry.

host-controller/src/main/java/org/jboss/as/domain/controller/transformers/DomainTransformers.java
pferraro: I didn't want to modify specific subsystems within the context of this proposal, especially since they are constantly changing.
But to answer your question, the ElytronSubsystemTransformers code above would become:

```java
@Override
    ChainedTransformationDescriptionBuilder chainedBuilder = registration.createChainedTransformationDescriptionBuilder(); 
    //...
}
```

pferraro:
Rebased on main.

PR #6218, opened 2024-10-16
[WFCORE-7031] Do not provision unstable annotation module for stabili…

yersan:
…ty levels higher than preview

Jira issue: https://issues.redhat.com/browse/WFCORE-7031

### DIFF
--- a/core-feature-pack/common/src/main/resources/modules/system/layers/base/org/wildfly/extension/core-management/main/module.xml
+++ b/core-feature-pack/common/src/main/resources/modules/system/layers/base/org/wildfly/extension/core-management/main/module.xml
@@ -30,7 +30,7 @@
         <module name="org.wildfly.extension.core-management-client"/>
         <module name="org.wildfly.security.elytron-private"/>
         <module name="org.wildfly.service"/>
-        <module name="org.wildfly.unstable.annotation.api.indexer"/>
-        <module name="org.wildfly._internal.unstable-api-annotation-index"/>
+        <module name="org.wildfly.unstable.annotation.api.indexer" optional="true"/>
+        <module name="org.wildfly._internal.unstable-api-annotation-index" optional="true"/>
     </dependencies>
 </module>
--- a/core-feature-pack/common/src/main/resources/modules/system/layers/base/org/wildfly/_internal/unstable-api-annotation-index/main/module.xml
+++ b/core-feature-pack/common/src/main/resources/modules/system/layers/base/org/wildfly/_internal/unstable-api-annotation-index/main/module.xml
@@ -9,6 +9,7 @@
 
     <properties>
         <property name="jboss.api" value="private"/>
+        <property name="jboss.stability" value="preview"/>
     </properties>
 
     <resources>
--- a/core-feature-pack/common/src/main/resources/modules/system/layers/base/org/wildfly/unstable/annotation/api/indexer/main/module.xml
+++ b/core-feature-pack/common/src/main/resources/modules/system/layers/base/org/wildfly/unstable/annotation/api/indexer/main/module.xml
@@ -8,6 +8,7 @@
 <module xmlns="urn:jboss:module:1.9" name="org.wildfly.unstable.annotation.api.indexer">
     <properties>
         <property name="jboss.api" value="private"/>
+        <property name="jboss.stability" value="preview"/>
     </properties>
 
     <resources>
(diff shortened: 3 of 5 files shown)


core-feature-pack/common/src/main/resources/modules/system/layers/base/org/wildfly/extension/core-management/main/module.xml
```diff
@@ -30,7 +30,7 @@
         <module name="org.wildfly.extension.core-management-client"/>
         <module name="org.wildfly.security.elytron-private"/>
         <module name="org.wildfly.service"/>
-        <module name="org.wildfly.unstable.annotation.api.indexer"/>
-        <module name="org.wildfly._internal.unstable-api-annotation-index"/>
+        <module name="org.wildfly.unstable.annotation.api.indexer" optional="true"/>
```
kabir, line 33: I'm not totally familiar with how this works.
Does being optional AND the dependency having `jboss.stability=preview` mean it gets provisioned?

jfdenise: When core-management is transformed onto a Galleon package, dependencies on optional modules are ignored. An optional module is expected to be provisioned from another root. For example, a layer, a subsystem, ...

kabir: Looking for 'org.wildfly._internal.unstable-api-annotation-index' and 'org.wildfly.unstable.annotation.api.indexer' in the wildfly-core sources, I don't see those being referenced anywhere that could make a difference apart from  via this module dependency

yersan: I've added the packages via model.xml, let's see how it goes

kabir:
@yersan Failures seem related

jfdenise:
@yersan I think that the inclusion thanks to model.xml of these 2 modules need to be replicated in WildFly Full repo.
https://github.com/wildfly/wildfly/blob/main/ee-feature-pack/galleon-local/src/main/resources/configs/standalone/model.xml
https://github.com/wildfly/wildfly/blob/main/ee-feature-pack/galleon-local/src/main/resources/configs/domain/model.xml

yersan:
@jfdenise @kabir I am missing something here. 

If we add packages via model.xml, are those packages provisioned when we are trimming the server using layers?

I mean, the error:

```
java.lang.AssertionError: Some expected modules have not been provisioned in test-all-layers: org.wildfly._internal.unstable-api-annotation-index,org.wildfly.unstable.annotation.api.indexer
  at org.jboss.as.test.layers.LayersTest.testLayersModuleUse(LayersTest.java:172)
  at org.jboss.as.test.shared.LayersTestBase.testLayersModuleUse(LayersTestBase.java:413)
  at org.jboss.as.test.layers.base.LayersTestCase.testLayersModuleUse(LayersTestCase.java:37)
```

Looks legit to me, but I guess adding them to model.xml won't fix it in WildFly. Also, I don't get why the Galleon Jobs passed under the current situation for WildFly Core, I would expect them to fail due the above assumption where packages are not added to layers if they are declared only in model.xml

What I am missing?

jfdenise:
@yersan , when included in standalone/model.xml, layers will benefit from them. Layers ased provisioning inherit from model.xml content.

yersan:
Integration Jobs (until the WildFly Counterpart gets merged)

https://ci.wildfly.org/viewLog.html?buildId=462644 (Linux integration)
https://ci.wildfly.org/viewLog.html?buildId=462646 (Bootable JAR)
https://ci.wildfly.org/viewLog.html?buildId=462642 (Galleon)
https://ci.wildfly.org/viewLog.html?buildId=462648 (Microprofile)
https://ci.wildfly.org/viewLog.html?buildId=462659 (WildFly Preview)
https://ci.wildfly.org/viewLog.html?buildId=462632 (Windows)

yersan:
This one needs https://github.com/wildfly/wildfly/pull/18287 merged first

yersan:
https://github.com/wildfly/wildfly/pull/18287 was merged, so kicking off the integration Jobs again

kabir:
Hi,

According to my understanding it now looks correct.

From @jfdenise  previous comment:

* The original dependencies is now optional
* Then the model.xml parts should count as being referenced from a root (and only kick in when preview or below)

yersan:
> From @jfdenise previous comment:
> 
> The original dependencies is now optional
> Then the model.xml parts should count as being referenced from a root (and only kick in when preview or below)

Yes, this is what we did here and in WildFly with the `"optional="true" valid-for-stability="preview"` configuration on the model.xml.

@kabir based on your feedback, I assume this is approved from your side (I don't see the green tick under your review though)

kabir (approved):

yersan:
Thanks @kabir @jfdenise

PR #6222, opened 2024-10-17
[WFCORE-7033] signature verification

spyrkob:
JIRA: https://issues.redhat.com/browse/WFCORE-7033


spyrkob:
@yersan this is not a complete PR yet, but could you take a quick look at this and let me know if you're happy with this approach?

installation-manager/src/main/java/org/wildfly/core/instmgr/cli/AddCertificatesCommand.java
```diff
@@ -0,0 +1,90 @@
...
+
+        return operationBuilder.build();
+    }
+
+    protected Operation buildParseOperation() {
+        final ModelNode op = new ModelNode();
+        final OperationBuilder operationBuilder = OperationBuilder.create(op);
+
+        op.get(OP).set(InstMgrCertificateParseHandler.DEFINITION.getName());
+        op.get("cert-file").set(0);
```
yersan, line 62: There is a constant for "cert-file"

installation-manager/src/main/java/org/wildfly/core/instmgr/cli/AddCertificatesCommand.java
```diff
@@ -0,0 +1,90 @@
...
+        final CommandContext ctx = commandInvocation.getCommandContext();
+
+        final ModelNode modelNode = this.executeOp(buildParseOperation(), ctx, this.host).get(RESULT);
+
+        commandInvocation.println("Components signed with following certificate will be considered trusted:");
+        commandInvocation.println("key-id: " + modelNode.get(InstMgrConstants.CERT_KEY_ID));
+        commandInvocation.println("fingerprint: " + modelNode.get(InstMgrConstants.CERT_FINGERPRINT));
+        commandInvocation.println("description: " + modelNode.get(InstMgrConstants.CERT_DESCRIPTION));
+
+        final String input = commandInvocation.inputLine(new Prompt(String.format("Import this certificate y/N")));
```
yersan, line 79: String.format is redundant since there is no arguments for it

installation-manager/src/main/java/org/wildfly/core/instmgr/cli/AddCertificatesCommand.java
```diff
@@ -0,0 +1,90 @@
...
+        if (nonInteractive || "y".equals(input)) {
+            commandInvocation.print("Importing a trusted certificate");
+        } else {
+            commandInvocation.print("Importing canceled.");
+        }
+
+        this.executeOp(ctx, this.host);
```
yersan, line 86: This should not be executed if the import is canceled

installation-manager/src/main/java/org/wildfly/core/instmgr/cli/AddCertificatesCommand.java
```diff
@@ -0,0 +1,90 @@
...
+    }
+
+    @Override
+    protected Operation buildOperation() {
+        final ModelNode op = new ModelNode();
+        final OperationBuilder operationBuilder = OperationBuilder.create(op);
+
+        op.get(OP).set(InstMgrCertificateImportHandler.DEFINITION.getName());
+        op.get(CERT_FILE).set(0);
+        operationBuilder.addFileAsAttachment(certFile);
```
yersan, line 52: Both, buildParseOperation() and buildOperation() will send the same file as an attachment. A possible improvement would be to save the attachment somehow in the Installation Manager service so it can be referenced by the second operation without sending it again. It would be a minor improvement that will require a method to clean up the attachments. Consider it only if the attachment sizes are big

yersan (commented):
Added some comments with a quick look, I have not played with this yet, but overall approach looks good

pom.xml
```diff
@@ -240,7 +240,7 @@
         <version.org.wildfly.client.config>1.0.1.Final</version.org.wildfly.client.config>
         <version.org.wildfly.common>1.7.0.Final</version.org.wildfly.common>
         <version.org.wildfly.discovery>1.3.0.Final</version.org.wildfly.discovery>
-        <version.org.wildfly.installation-manager.installation-manager-api>1.0.3.Final</version.org.wildfly.installation-manager.installation-manager-api>
+        <version.org.wildfly.installation-manager.installation-manager-api>2.0.0.Beta1-SNAPSHOT</version.org.wildfly.installation-manager.installation-manager-api>
```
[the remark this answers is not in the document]
spyrkob: speaking of which I think the version in the main branch of https://github.com/wildfly-extras/wildfly-installation-manager-api is not correct - it should be `2.0.0.Beta2-SNAPSHOT` not `2.0.1.Beta1-SNAPSHOT`?

spyrkob:
Thanks @yersan! I updated the PR with your comments and made a few changes to work with the Web Console a bit easier. I'll create the wildfly-proposal and mark this as ready for review when that's ready

pom.xml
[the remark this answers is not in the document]
yersan: @spyrkob  Right, I've corrected the version in `main`, thanks for reporting it

PR #6228, opened 2024-10-29
[WFCORE-6981] fix incorrect alternatives in reply-properties inherited from basis attribute

soul2zimate:
Issue: https://issues.redhat.com/browse/WFCORE-6981 

[WFCORE-6981] fix incorrect alternatives in `reply-properties `inherited from basis attribute

### DIFF
--- a/server/src/main/java/org/jboss/as/server/controller/resources/DeploymentAttributes.java
+++ b/server/src/main/java/org/jboss/as/server/controller/resources/DeploymentAttributes.java
@@ -395,6 +395,7 @@ public class DeploymentAttributes {
     /** Return type for the upload-deployment-xxx operaions */
     private static final SimpleAttributeDefinition UPLOAD_HASH_REPLY = SimpleAttributeDefinitionBuilder.create(CONTENT_HASH)
             .setRequired(true)
+            .setAlternatives()
             .build();
 
 


yersan (changes requested):
Hello @soul2zimate 

This is a partial fix, there are several operations that have this issue. We are getting these alternatives because the CONTENT_HASH attribute is already built with those alternatives in place:

https://github.com/wildfly/wildfly-core/blob/e4bedf571fe2a01a0234279afc5562108d2ad1e8/server/src/main/java/org/jboss/as/server/controller/resources/DeploymentAttributes.java#L155-L159

If we look at that file, we can see several attributes that are created with those alternatives in place. So, this PR is a partial fix only for the usage of CONTENT_HASH in the UPLOAD_HASH_REPLY operation only/

The question is more why those attributes were created with all of those alternatives if they can end up as a single attribute in one operation. I don't know, but in such a case, the alternatives should be removed. Maybe @ehsavoie could remember.

So, I guess we need to perform a full check of those attributes and remove the alternatives when required in the Operation handlers where they are used, or, maybe fix their definition by adding them without alternatives and adding them when more than one is added to the same Operation.

soul2zimate:
> So, this PR is a partial fix only for the usage of CONTENT_HASH in the UPLOAD_HASH_REPLY operation only/

Hello @yersan, `UPLOAD_HASH_REPLY` is attribute, the fix applies to all related operations `UPLOAD_BYTES_DEFINITION`, `DOMAIN_UPLOAD_BYTES_DEFINITION`, `UPLOAD_URL_DEFINITION`, `DOMAIN_UPLOAD_URL_DEFINITION`, `UPLOAD_STREAM_ATTACHMENT_DEFINITION` and `DOMAIN_UPLOAD_STREAM_ATTACHMENT_DEFINITION` as `UPLOAD_HASH_REPLY` is the reply parameter. It doesn't make sense to provide alternatives for `reply-properties`.

AFAIU, I need to check the usage of this [`create`](https://github.com/wildfly/wildfly-core/blob/27.0.0.Beta1/controller/src/main/java/org/jboss/as/controller/SimpleAttributeDefinitionBuilder.java#L27) method and make sure there are no incorrect alternatives inherited from basis attributes. In these places, where the basic attribute defines alternatives, it has already been taken care of. 

e.g. https://github.com/wildfly/wildfly-core/blob/27.0.0.Beta1/logging/src/main/java/org/jboss/as/logging/handlers/SocketHandlerResourceDefinition.java#L84-L89 and https://github.com/wildfly/wildfly-core/blob/27.0.0.Beta1/logging/src/main/java/org/jboss/as/logging/handlers/SyslogHandlerResourceDefinition.java#L116

ehsavoie:
You need all the alternative in the operation parameters but I think the issue was in the reply attribute that was created based on the parameter instead of the parameter name + type

yersan (approved):
I see it now @soul2zimate , the UPLOAD_HASH_REPLY is used as a common structure for the reply parameters of all the upload_XXXX operations, but it was inheriting the alternative attributes from the CONTENT_HASH, hence the problem with the alternatives being leaked into the reply parameters

yersan:
Thanks @soul2zimate and @ehsavoie

PR #6233, opened 2024-11-02
[WFCORE-7040] Move to SE 17 as the minimum SE version

bstansberry:
https://issues.redhat.com/browse/WFCORE-7040

This is just a draft until we're sure we're ready.

### DIFF
--- a/.github/workflows/dep-diff-pull_request.yml
+++ b/.github/workflows/dep-diff-pull_request.yml
@@ -38,7 +38,7 @@ jobs:
         uses: actions/setup-java@v4
         with:
           distribution: 'temurin'
-          java-version: 11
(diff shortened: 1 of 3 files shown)


bstansberry:
One reason I've let this sit is there's benefit in leaving the 'client/tools' modules on SE 11 so tools can continue to use the latest while still supporting SE 11 for use with older releases.

Client/tools == controller-client, protocol, cli, embedded

bstansberry:
We need to move on to SE17-based deps like JBoss Threads 3.9, so it's time to move on with this.

jamezp (approved):

yersan:
Thanks @bstansberry @jamezp

PR #6242, opened 2024-11-18
[WFCORE-7044] Remove the launcher module; use the wildfly-launcher pr…

bstansberry:
…oject artifact instead

https://issues.redhat.com/browse/WFCORE-7044

Note: CI of integration with full WF will fail until https://github.com/wildfly/wildfly/pull/18400 is merged.


jamezp (approved):

bstansberry:
Note that this is upgrading to a Beta, but that's because integrating the output of that project is the best way to validate the initial release. I assume @jamezp would be fine releasing a .Final as soon as we're satisfied it's ok.

jfdenise (approved):

yersan (approved):

PR #6245, opened 2024-11-19
[WFCORE-7058]: CachingRealm should start lazily in admin only mode.

ehsavoie:
* updating the start mode to LAZY if the server is in admin-mode
 * adding a proper error message if the caching realm service is not UP when the user tries to clear the cache.

Issue: https://issues.redhat.com/browse/WFCORE-7058


darranl (approved):
Thank you @ehsavoie

yersan:
Thanks @ehsavoie @darranl

PR #6247, opened 2024-11-20
[WFCORE-3720] Update ServiceModuleLoader to use non-deprecated JBoss Modules APIs

lvydra:
https://issues.redhat.com/browse/WFCORE-3720

### DIFF
--- a/server/src/main/java/org/jboss/as/server/deployment/DeploymentListModulesHandler.java
+++ b/server/src/main/java/org/jboss/as/server/deployment/DeploymentListModulesHandler.java
@@ -74,7 +74,7 @@ public class DeploymentListModulesHandler implements OperationStepHandler {
                         moduleIdentifier = deploymentUnit.getAttachment(Attachments.MODULE_IDENTIFIER);
                     }
 
-                    final ServiceController<?> moduleLoadServiceController = sr.getService(ServiceModuleLoader.moduleServiceName(moduleIdentifier));
+                    final ServiceController<?> moduleLoadServiceController = sr.getService(ServiceModuleLoader.moduleServiceName(moduleIdentifier.toString()));
                     final ModuleLoadService moduleLoadService = (ModuleLoadService) moduleLoadServiceController.getService();
 
                     final ModelNode result = new ModelNode();
--- a/server/src/main/java/org/jboss/as/server/moduleservice/ServiceModuleLoader.java
+++ b/server/src/main/java/org/jboss/as/server/moduleservice/ServiceModuleLoader.java
@@ -128,9 +128,19 @@ public class ServiceModuleLoader extends ModuleLoader implements Service<Service
         }
     }
 
+    /**
+     * @deprecated Will be made protected in line with this method in the parent class Use {@link ServiceModuleLoader#findModule(String)}
+     */
+    @Deprecated(forRemoval = true, since = "27.0.0")
     @SuppressWarnings("unchecked")
     @Override
     public ModuleSpec findModule(ModuleIdentifier identifier) throws ModuleLoadException {
+        return findModule(identifier.toString());
+    }
+
+    @SuppressWarnings("unchecked")
+    @Override
+    protected ModuleSpec findModule(String identifier) throws ModuleLoadException {
         ServiceController<ModuleDefinition> controller = (ServiceController<ModuleDefinition>) serviceContainer.getService(moduleSpecServiceName(identifier));
         if (controller == null) {
             ServerLogger.MODULE_SERVICE_LOGGER.debugf("Could not load module '%s' as corresponding module spec service '%s' was not found", identifier, identifier);
@@ -193,18 +203,42 @@ public class ServiceModuleLoader extends ModuleLoader implements Service<Service
      *
      * @param identifier The module identifier
      * @return The service name of the ModuleSpec service
+     * @deprecated Use {@link ServiceModuleLoader#moduleSpecServiceName(String)}
      */
+    @Deprecated(forRemoval = true, since = "27.0.0")
     public static ServiceName moduleSpecServiceName(ModuleIdentifier identifier) {
+        return moduleSpecServiceName(identifier.toString());
+    }
+
+    /**
+     * Returns the corresponding ModuleSpec service name for the given module.
+     *
+     * @param identifier The module identifier string, method does not support non-canonized identifiers
+     * @return The service name of the ModuleSpec service
+     */
+    public static ServiceName moduleSpecServiceName(String identifier) {
         if (!isDynamicModule(identifier)) {
             throw ServerLogger.ROOT_LOGGER.missingModulePrefix(identifier, MODULE_PREFIX);
         }
-        return MODULE_SPEC_SERVICE_PREFIX.append(identifier.getName()).append(identifier.getSlot());
+        return MODULE_SPEC_SERVICE_PREFIX.append(identifier);
     }
 
+    /**
+     * @deprecated Use {@link ServiceModuleLoader#installModuleResolvedService(ServiceTarget, String)}
+     */
+    @Deprecated(forRemoval = true, since = "27.0.0")
     public static void installModuleResolvedService(ServiceTarget serviceTarget, ModuleIdentifier identifier) {
+        installModuleResolvedService(serviceTarget, identifier.toString());
+    }
+
+    /**
+     * @param serviceTarget service target to use to install the service. Cannot be {@code null}.
+     * @param identifier The module identifier string, method does not support non-canonized identifiers
+     */
+    public static void installModuleResolvedService(ServiceTarget serviceTarget, String identifier) {
         final ServiceName sn = ServiceModuleLoader.moduleResolvedServiceName(identifier);
         final ServiceBuilder<?> sb = serviceTarget.addService(sn);
-        final Consumer<ModuleIdentifier> moduleIdConsumer = sb.provides(sn);
+        final Consumer<String> moduleIdConsumer = sb.provides(sn);
         sb.requires(moduleSpecServiceName(identifier));
         final org.jboss.msc.Service resolvedService = org.jboss.msc.Service.newInstance(moduleIdConsumer, identifier);
         sb.setInstance(resolvedService);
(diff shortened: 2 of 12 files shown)


bstansberry:
In general ModuleIdentifier.getName() is not a safe replacement for ModuleIdentifier. It does not include any slot value. ModuleIdentifier.toString() provides the String variant of the module id.

server/src/main/java/org/jboss/as/server/deployment/DeploymentListModulesHandler.java
```diff
@@ -74,7 +74,7 @@ public void execute(OperationContext context, ModelNode operation) {
                         moduleIdentifier = deploymentUnit.getAttachment(Attachments.MODULE_IDENTIFIER);
                     }
 
-                    final ServiceController<?> moduleLoadServiceController = sr.getService(ServiceModuleLoader.moduleServiceName(moduleIdentifier));
+                    final ServiceController<?> moduleLoadServiceController = sr.getService(ServiceModuleLoader.moduleServiceName(moduleIdentifier.getName()));
```
bstansberry, line 77: All the changes like this need to be updated to use toString(), not getName().

server/src/main/java/org/jboss/as/server/moduleservice/ServiceModuleLoader.java
```diff
@@ -218,33 +218,38 @@ public static void installModuleResolvedService(ServiceTarget serviceTarget, Mod
...
     /**
      * Returns true if the module identifier is a dynamic module that will be loaded by this module loader
      */
-    public static boolean isDynamicModule(ModuleIdentifier identifier) {
-        return identifier.getName().startsWith(MODULE_PREFIX);
+    public static boolean isDynamicModule(String name) {
+        return name.startsWith(MODULE_PREFIX);
+    }
+
+    @Deprecated
```
bstansberry, line 238: Please add forRemoval=true and add javadoc @deprecated pointing to the new method.

bstansberry (changes requested):

lvydra:
Hi @bstansberry, thank you for the review, change requests should be addressed now.

server/src/main/java/org/jboss/as/server/moduleservice/ServiceModuleLoader.java
```diff
@@ -130,10 +130,10 @@ protected Module preloadModule(final String name) throws ModuleLoadException {
 
     @SuppressWarnings("unchecked")
     @Override
-    public ModuleSpec findModule(ModuleIdentifier identifier) throws ModuleLoadException {
-        ServiceController<ModuleDefinition> controller = (ServiceController<ModuleDefinition>) serviceContainer.getService(moduleSpecServiceName(identifier));
+    public ModuleSpec findModule(String name) throws ModuleLoadException {
+        ServiceController<ModuleDefinition> controller = (ServiceController<ModuleDefinition>) serviceContainer.getService(moduleSpecServiceName(name));
```
yersan, line 134: I think this method should have been deprecated for removal calling internally to the one using the String as an argument. Otherwise, any caller currently using the old one will now be using the one provided by the parent class.

server/src/main/java/org/jboss/as/server/moduleservice/ServiceModuleLoader.java
```diff
@@ -218,33 +218,45 @@ public static void installModuleResolvedService(ServiceTarget serviceTarget, Mod
...
      * come up.
      *
-     * @param identifier The module identifier
+     * @param name The module name
      * @return The service name of the ModuleSpec service
      */
-    public static ServiceName moduleResolvedServiceName(ModuleIdentifier identifier) {
-        if (!isDynamicModule(identifier)) {
-            throw ServerLogger.ROOT_LOGGER.missingModulePrefix(identifier, MODULE_PREFIX);
+    public static ServiceName moduleResolvedServiceName(String name) {
```
yersan, line 224: This falls into the same category as what was done with `ServiceName moduleServiceName(ModuleIdentifier identifier)` and should be deprecated as well, although better to get other feedback since I don't know exactly what was the use case where this can be called from other parties than WildFly

server/src/main/java/org/jboss/as/server/moduleservice/ServiceModuleLoader.java
```diff
@@ -218,33 +218,45 @@ public static void installModuleResolvedService(ServiceTarget serviceTarget, Mod
...
+        return MODULE_RESOLVED_SERVICE_PREFIX.append(name);
     }
 
     /**
      * Returns true if the module identifier is a dynamic module that will be loaded by this module loader
      */
-    public static boolean isDynamicModule(ModuleIdentifier identifier) {
-        return identifier.getName().startsWith(MODULE_PREFIX);
+    public static boolean isDynamicModule(String name) {
+        return name.startsWith(MODULE_PREFIX);
```
yersan, line 235: This falls into the same category as what was done with `ServiceName moduleServiceName(ModuleIdentifier identifier)` and should be deprecated as well. Although better to get other feedback since I don't know exactly what was the use case where this can be called from other parties than WildFly

lvydra:
Thanks @yersan, updated.

server/src/main/java/org/jboss/as/server/moduleservice/ServiceModuleLoader.java
```diff
@@ -191,22 +201,22 @@ public static void addService(final ServiceTarget serviceTarget, final Bootstrap
...
-    public static ServiceName moduleSpecServiceName(ModuleIdentifier identifier) {
-        if (!isDynamicModule(identifier)) {
-            throw ServerLogger.ROOT_LOGGER.missingModulePrefix(identifier, MODULE_PREFIX);
+    public static ServiceName moduleSpecServiceName(String name) {
+        if (!isDynamicModule(name)) {
+            throw ServerLogger.ROOT_LOGGER.missingModulePrefix(name, MODULE_PREFIX);
         }
-        return MODULE_SPEC_SERVICE_PREFIX.append(identifier.getName()).append(identifier.getSlot());
+        return MODULE_SPEC_SERVICE_PREFIX.append(name);
     }
```
yersan, line 212: I would say we should Keep the `ServiceName moduleSpecServiceName(ModuleIdentifier identifier)` variant and deprecate it for removal.

server/src/main/java/org/jboss/as/server/moduleservice/ServiceModuleLoader.java
```diff
@@ -191,22 +201,22 @@ public static void addService(final ServiceTarget serviceTarget, final Bootstrap
...
         final ServiceBuilder<?> sb = serviceTarget.addService(sn);
-        final Consumer<ModuleIdentifier> moduleIdConsumer = sb.provides(sn);
-        sb.requires(moduleSpecServiceName(identifier));
-        final org.jboss.msc.Service resolvedService = org.jboss.msc.Service.newInstance(moduleIdConsumer, identifier);
+        final Consumer<String> moduleIdConsumer = sb.provides(sn);
+        sb.requires(moduleSpecServiceName(name));
+        final org.jboss.msc.Service resolvedService = org.jboss.msc.Service.newInstance(moduleIdConsumer, name);
         sb.setInstance(resolvedService);
         sb.install();
     }
```
yersan, line 222: The same here, do not remove the previous method without deprecating it first.

server/src/main/java/org/jboss/as/server/moduleservice/ServiceModuleLoader.java
yersan: nitpick: I would also keep the argument name as `identifier`. Even being a string, it will be an identifier (name+slot). It is not a hard requirement, but I think it is more coherent with what it is. The ModuleIdentifier class keeps the two fields (name and slot). It also has a method named moduleIdentifier.getName() which will return only the name, without a slot. So, `name` the argument as `name` here could be misleading and make someone think that it should be invoked with `moduleIdentifier.getName()`.

yersan (changes requested):

lvydra:
Thanks @yersan, updated.

bstansberry:
FYI re https://github.com/wildfly/wildfly-core/pull/6270

I don't know if that will be the ultimate approach, and I haven't looked at the current state of this to see if it's applicable to the change here. But FYI.

My expected use of that utility is that in any place where we are taking external input as the string naming a module, we run it through that utility and then use the canonicalized string. Internal places where we know any passed in string has already been through that canonicalization can just use the provided string. Understanding what case is what requires care.

External inputs include things like management model 'module' attribute values, management model ee subsystem global directory resource names, data parsed from deployment descriptors and data parsed from MANIFEST files. Perhaps others but I think that's the list.

If what I'm talking about *is* applicable here, please hold this until we get something like https://github.com/wildfly/wildfly-core/pull/6270 merged and then this PR should use it.

yersan:
@bstansberry 

This is tricky. The new String variants have been created from the `ModuleIdentifed` versions, it means that we can shift the responsibility to invoke them with a proper canonized module identifier when we are adapting the callers. In that case, we should at least add a proper Java Doc on those new public methods stating that.

However, it extends the `ModuleLoader`, and there are other methods using a string in the ModuleLocader parent class. Should not they be protected by JBoss Modules, either by accepting a non-canonized version or providing proper Java Doc if that is not acceptable?

yersan:
> However, it extends the ModuleLoader, and there are other methods using a string in the ModuleLoader parent class. Should not they be protected by JBoss Modules, either by accepting a non-canonized version or providing proper Java Doc if that is not acceptable?

Well, the only one inherited from ModuleLoader is `findModule(String)`, so we should ensure we can manage non-canonized indentifier in that method, although the parent Java Doc is a bit misleading about that.

The others I guess are up to us, but if they cannot manage a canonized version, we should document them as such since they are public methods.

I think the problem here is that they are public methods so we cannot control who can invoke them outside of our code. @bstansberry / @lvydra any thoughts on this?

yersan:
@lvydra In addition to my previous comment, could you also add a "since=27.0.0" attribute to the @deprecated annotations? it will help us track when they were deprecated without looking at the Git history, thanks!

server/src/main/java/org/jboss/as/server/moduleservice/ServiceModuleLoader.java
```diff
@@ -128,9 +130,20 @@ protected Module preloadModule(final String name) throws ModuleLoadException {
         }
     }
 
+    /**
+     * @deprecated Use {@link ServiceModuleLoader#findModule(String)}
```
bstansberry, line 134: Please change this to:

@deprecated Will be made protected in line with this method in the parent class

There may have been a reason for this being public in 2011 when it was changed to public, but I don't see it now. It should be an internal call within ModuleLoader, not used by outside code.  (WF and WF Core code don't call this method directly.)

I doubt any code is calling this and if they are they are misusing something. Removing ModuleIdentifier will be a breaking change to whatever the code might be so they might as well stop doing whatever that is.

We'll be removing this method later

server/src/main/java/org/jboss/as/server/moduleservice/ServiceModuleLoader.java
```diff
@@ -128,9 +130,20 @@ protected Module preloadModule(final String name) throws ModuleLoadException {
...
+    @Deprecated(forRemoval = true, since = "27.0.0")
     @SuppressWarnings("unchecked")
     @Override
     public ModuleSpec findModule(ModuleIdentifier identifier) throws ModuleLoadException {
+        return findModule(identifier.toString());
+    }
+
+    @SuppressWarnings("unchecked")
+    @Override
+    public ModuleSpec findModule(String identifier) throws ModuleLoadException {
```
bstansberry, line 145: I think this should be protected, in line with the superclass.

bstansberry:
> If there is no other feedback, just to sum up what is pending:
> 
> * `findModule(String)` should be able to manage non-canonized identifiers, as it is supposed the same method on the parent class does

I don't think it's our responsibility to handle this. JBoss Modules itself doesn't, so I don't think we should. As I commented on the new method, I think it should be protected, as it's not meant to be called by arbitrary code. (And isn't).

Note that nothing in WF calls findModule. It's called by ModuleLoader.loadModuleLocal. If the identifier passed from there is non-canonical then lots of other code in JBM has already been trying to deal with that string.

> * The other new variants created by us are up to us, if we want to put the responsibility to the callers, we have to explicitly document the public methods stating that they don't support non-canonized identifiers, otherwise, we have to ensure inside of each of those public methods that the identified passed as a String is canonized before working with it. I am more in favor of shifting the responsibility to the caller to avoid computing the same more than once internally, although it is a bit less robust and the computing time will be inappreciable, so open to any other feedback.

The responsibility for using canonicalModuleIdentifier should belong to the code that is accepting external inputs for module names. If we start using it inside tons methods that take a String passed in by other code we'll end up canonicalizing the same String multiple times.

+1 to javadoc; the way it's handled in the PR looks good.

lvydra:
Thanks @bstansberry, updated.

bstansberry (approved):

yersan:
Thanks @lvydra and @bstansberry

PR #6250, opened 2024-11-21
[WFCORE-7063] Removing useless code related to JDK version detection

ropalka:
https://issues.redhat.com/browse/WFCORE-7063


bstansberry (approved):

ropalka:
I added short description to the JIRA as you suggested @yersan and yes those checks were irrelevant because minimum JDK that is required is version 11 so assumption was always true.

yersan:
Thanks @ropalka @bstansberry

PR #6252, opened 2024-11-21
[WFCORE-7064] Do not set SM options on JDK24+

ropalka:
https://issues.redhat.com/browse/WFCORE-7064

### DIFF
--- a/testsuite/scripts/src/test/java/org/wildfly/launcher/test/LauncherTestCase.java
+++ b/testsuite/scripts/src/test/java/org/wildfly/launcher/test/LauncherTestCase.java
@@ -45,10 +45,13 @@ public class LauncherTestCase {
     @Test
     public void testStandaloneWithAgent() throws Exception {
         final StandaloneCommandBuilder builder = StandaloneCommandBuilder.of(ServerHelper.JBOSS_HOME)
-                .addJavaOptions(ServerHelper.DEFAULT_SERVER_JAVA_OPTS)
-                .setUseSecurityManager(parseProperty("security.manager"))
-                // Add the test logging agent to the jboss-modules arguments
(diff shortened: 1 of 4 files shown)


bstansberry:
The test changes look good, but I'm wondering about the production code changes. 

We don't want to be setting -Djava.security.manager=allow _on our own_ on SE 24+, but that would be done in things like the scripts or the launcher. The code here seems to be about processing external input. (It may be input that's been through various layers and transformations, but ultimately it's derived from external input.) If we get input to enable the SM, I don't think we should ignore it on 24; we should process it and let things fail.

host-controller/src/main/java/org/jboss/as/host/controller/Main.java
```diff
@@ -212,7 +212,8 @@ public static HostControllerEnvironmentWrapper determineEnvironment(String[] arg
         String modulePath = null;
 
         // Note the java.security.manager property shouldn't be set, but we'll check to ensure the security manager should be enabled
-        boolean securityManagerEnabled = System.getSecurityManager() != null || isJavaSecurityManagerConfigured(hostSystemProperties);
+        // [WFCORE-7064] Setting SM is not allowed on JDK24+
+        boolean securityManagerEnabled = Runtime.version().feature() < 24 && (System.getSecurityManager() != null || isJavaSecurityManagerConfigured(hostSystemProperties));
```
bstansberry, line 216: If this was as follows it would guard/future-proof the System.getSecurityManager() call, while not ignoring the processing of external inputs:

```
boolean securityManagerEnabled = (Runtime.version().feature() < 24 && (System.getSecurityManager() != null) || isJavaSecurityManagerConfigured(hostSystemProperties);
```

ropalka: Fixed.

ropalka:
Thinking about it you are of course right @bstansberry ! I updated PR to include only test suite changes.

bstansberry (approved):

bstansberry:
Thanks, @ropalka!

PR #6260, opened 2024-11-26
[WFCORE-7074] Use the ModuleIdentifier.fromString() where a slot may …

jamezp:
…be used in a string module name.

https://issues.redhat.com/browse/WFCORE-7074


bstansberry:
Perhaps this should also fix ExternalModuleService.addExternalModule. I believe the usage via GlobalDirectoryDeploymentService could incorporate a user-defined module name with a slot. It would be an edge case but I think the module name is the value part of a PathElement key/value pair, and thus could have a value like "foo:slot" or "foo:main".

bstansberry:
Thanks @jamezp. This passed CI so we can deal with ExternalModuleService.addExternalModule separately.

PR #6261, opened 2024-11-26
WFCORE-6779 Replace PersistentResourceXMLDescription with more robust API

pferraro:
https://issues.redhat.com/browse/WFCORE-6779

Feature set includes everything from PersistentResourceXMLDescription, plus:
* Better support for non-default stability-levels
* More intuitive implementation, including consistent entry/exit criteria for reader/writer methods
* Correct XML namespace/prefix support
* Support for (and strict enforcement of) element and element group cardinality during parsing
* Support for override resource registrations, e.g. /name=*, /name=value1, /name=value2, etc.
* Support for (and strict enforcement of) xs:choice, xs:sequence, and xs:all semantics.
* Explicit element content definition (via xs:choice, xs:sequence, or xs:all).
* Better operation manipulation capability
* Improved attribute/element name and parser/marshaller overrides for legacy schema versions

N.B. This proposed change no longer attempts to reimplement PersistentResourceXMLDescription using the new implementation, as this is made impossible (without degrading existing schema validation logic) by existing illegal subsystem schemas (see WFLY-20446 and WFCORE-7181).  Instead, PersistentSubsystemSchema now extends SubsystemResourceXMLSchema by adapting the PersistentResourceXMLDescription for the subsystem resource to a SubsystemResourceXMLElement.

Requires:
* https://github.com/wildfly/wildfly-core/pull/6377
* https://github.com/wildfly/wildfly-core/pull/6375


controller/src/main/java/org/jboss/as/controller/PersistentResourceXMLDescriptionReader.java
```diff
@@ -17,7 +17,9 @@
 /**
  * An {@link XMLElementReader} based on a {@link PersistentResourceXMLDescription}.
  * @author Paul Ferraro
+ * @deprecated Superseded by {@link org.jboss.as.controller.persistence.xml.SubsystemResourceXMLElementReader}.
  */
+@Deprecated(forRemoval = true)
```
yersan, line 22: Hello @pferraro Do not forget our policy about adding the "since" attribute to any deprecated method. thanks!

pferraro: Fixed.

pferraro:
Re: above elytron-oidc-client subsystem parsing failures, see https://wildfly.zulipchat.com/#narrow/channel/174184-wildfly-developers/topic/elytron-oidc-client.20subsystem.20schema.20vs.20model.20discrepancies
FYI @fjuma

pferraro:
N.B. Undertow subsystem test failures are due to https://issues.redhat.com/browse/WFLY-20439

controller/src/main/java/org/jboss/as/controller/logging/ControllerLogger.java
```diff
@@ -3813,4 +3815,45 @@ OperationFailedRuntimeException capabilityAlreadyRegisteredInContext(String capa
     @Message(id = 518, value = "Resource /extension=%s uses a non-canonical module name; use the canonical form %s. " +
             "The canonical representation includes slot information in the module name only if the is slot is not 'main'.")
     String nonCanonicalExtensionName(String nonCanonical, String canonical);
+
+    @LogMessage(level = WARN)
+    @Message(id = 519, value = "Attribute '%2$s' of element '%1$s' is no longer supported and will be ignored")
+    void attributeIgnored(QName elementName, QName attributeName);
```
yersan, line 3821: Optional. Maybe it is easier / clearer if instead of specifying the argument number, the argument methods are put in orden as they are expected in the log message

discovery/src/main/java/org/wildfly/extension/discovery/AggregateDiscoveryProviderRegistrar.java
```diff
@@ -1,51 +0,0 @@
...
-import org.wildfly.subsystem.resource.capability.CapabilityReferenceListAttributeDefinition;
-import org.wildfly.subsystem.service.ResourceServiceInstaller;
-import org.wildfly.subsystem.service.ServiceDependency;
-import org.wildfly.subsystem.service.capability.CapabilityServiceInstaller;
-
-/**
- * Registers the aggregate discovery provider resource definition.
- * @author Paul Ferraro
- */
-public class AggregateDiscoveryProviderRegistrar extends DiscoveryProviderRegistrar {
```
yersan, line 29: Should not this be first deprecated and then removed?

yersan (commented):
@pferraro Is this already ready for review? I've been monitoring and has been pushed several times with new changes. If not, move it to draft instead.

This is also addressing some issues under Elytron and IO subsystems, my feeling is those changes should be tracked in different Jiras so relevant component leads can have a better way to track them for future maintenance.

If they agree to manage those changes here, I'm fine with it (but seems more logical to have a specific Jira to track them), in any case CC @fjuma / @fl4via so they can confirm if they are fine with the current approach.

discovery/src/main/java/org/wildfly/extension/discovery/AggregateDiscoveryProviderRegistrar.java
pferraro: This is not a public module.

controller/src/main/java/org/jboss/as/controller/logging/ControllerLogger.java
pferraro: What aspect of argument indices is not clear?

Frankly, we should probably always use argument indices in the message text, since the text itself can always change, not to mention that the incident order may change during translation.

In this instance, I find it more natural to specify the element name before the attribute name in the method signature (just as you would encounter them in the XML itself).

pferraro:
@yersan
> Is this already ready for review? I've been monitoring and has been pushed several times with new changes.

Yes, this is ready for review now.

> This is also addressing some issues under Elytron and IO subsystems, my feeling is those changes should be tracked in different Jiras so relevant component leads can have a better way to track them for future maintenance.

Good point. Originally, I planned to reimplement PersistentResourceXMLDescription to delegate to this API - and the fixes to which you refer were prerequisites.
Now that this is no longer the case (due to WFLY-20446 and WFCORE-7181), I can easily split these into separate pull requests (with distinct jira issues).

bstansberry:
Regarding the JIRA issues, please add WFCORE-6779 to the 5th commit's message, unless you intend to squash it into the first.  I mean the 'Generalize PersistentSubsystemSchema as a SubsystemResourceXMLSchema' commit. If that change pops up in a git blame it should be possible to find the JIRA directly from the commit message.

Maybe the same for the map parser commit.

controller/src/main/java/org/jboss/as/controller/logging/ControllerLogger.java
yersan: > What aspect of argument indices is not clear?

Well, it is that it just results more readable since none of the arguments for the loggers use the positional notation; they use the orden that are defined. Anyway, I commented about it as an optional change, but for sure, it is ok to leave it as it is.

yersan:
yes, in general, I am always more inclined to use the Jira number on each commit, it makes it easier to navigate and evaluate the reason for that change more quickly. Adding the Jira link to the commit message also helps but this is always a personal preference

pferraro:
@bstansberry
> Regarding the JIRA issues, please add WFCORE-6779 to the 5th commit's message, unless you intend to squash it into the first. I mean the 'Generalize PersistentSubsystemSchema as a SubsystemResourceXMLSchema' commit. If that change pops up in a git blame it should be possible to find the JIRA directly from the commit message.

Done.

> Maybe the same for the map parser commit.

I've extracted this to a separate JIRA and PR to be reviewed/merged separately: https://github.com/wildfly/wildfly-core/pull/6375

@yersan
> This is also addressing some issues under Elytron and IO subsystems, my feeling is those changes should be tracked in different Jiras so relevant component leads can have a better way to track them for future maintenance.

Done. See referenced pull requests.

pferraro:
Rebased against main.

yersan (approved):
Thanks @pferraro , according to the CI, this looks ready for merge, one last thing, could you please fix the commit message that points to WFCORE-7220 ? It should be WFCORE-7200 instead

pferraro:
@yersan Whoops - will fix that commit message.

yersan:
Thanks @pferraro

PR #6262, opened 2024-11-27
[WFCORE-7074] Use fromString instead of create since moduleName in th…

bstansberry:
...eory could include a slot

https://issues.redhat.com/browse/WFCORE-7074

Follow up to #6260 to pick up https://github.com/wildfly/wildfly-core/pull/6260#issuecomment-2501944640


jamezp (approved):
Thank you @bstansberry

yersan:
Thanks @jamezp @bstansberry

PR #6263, opened 2024-11-27
[WFCORE-6497] - Move JVMServerPropertiesTestCase from WildFly to Wild…

pedro-hos:
…Fly Core

https://issues.redhat.com/browse/WFCORE-6497

### DIFF
--- a/testsuite/domain/src/test/resources/host-configs/host-primary-jvm-properties.xml
+++ b/testsuite/domain/src/test/resources/host-configs/host-primary-jvm-properties.xml
@@ -1,0 +1,206 @@
+<?xml version='1.0' encoding='UTF-8'?>
+<!--
+  ~ Copyright The WildFly Authors
+  ~ SPDX-License-Identifier: Apache-2.0
+  -->
+
+<host xmlns="urn:jboss:domain:20.0"
+      name="primary" organization="core-primary">
+
+    <extensions>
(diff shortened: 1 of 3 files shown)


testsuite/domain/src/test/resources/host-configs/host-primary-jvm-properties.xml
```diff
@@ -0,0 +1,215 @@
...
+            </environment-variables>
+            <!--
+            <jvm-options>
+                <option value="-Xdebug"/>
+                <option value="-Xnoagent"/>
+                <option value="-Djava.compiler=NONE"/>
+                <option value="-agentlib:jdwp=transport=dt_socket,address=5005,server=y,suspend=y"/>
+
+            </jvm-options>
+            -->
```
yersan, line 84: Would you mind removing this comment from the file? It will tend to keep it there and even copy pasted without any good reason

yersan (approved):
LGTM, only minor optional detail that should not prevent merging this

testsuite/domain/src/test/resources/host-configs/host-primary-jvm-properties.xml
pedro-hos: Sure, I have just removed and squashed the commit. Thanks

yersan:
Thanks @pedro-hos

PR #6265, opened 2024-11-28
WFCORE-7050 - Remove MetaspaceSize settings

pedro-hos:
https://issues.redhat.com/browse/WFCORE-7050


yersan:
Any reason for not removing it also from the testsuite/pom.xml? https://github.com/wildfly/wildfly-core/blob/main/testsuite/pom.xml#L120 

I know it is harmless, and testsuite is not the main focus for this PR, but, just to be consistent, those arguments will be used by Surefire JVM

pedro-hos:
> Any reason for not removing it also from the testsuite/pom.xml? https://github.com/wildfly/wildfly-core/blob/main/testsuite/pom.xml#L120
> 
> I know it is harmless, and testsuite is not the main focus for this PR, but, just to be consistent, those arguments will be used by Surefire JVM

The test failed on Wildfly without this parameter, so I didn't remove it from the wildfly-core either. However, I have just tested here, and on core, all tests passed. So, removing this on pom.xml as well.

yersan (approved):

yersan:
Thanks @pedro-hos

PR #6266, opened 2024-11-29
[WFCORE-7076] Check required installation metadata is present before registering installation-manager subsystem

spyrkob:
Issue: https://issues.redhat.com/browse/WFCORE-7076


spyrkob:
@yersan I tried to implement your suggestion about changing `ModelControllerServiceInitialization` to pass the `PathManager` to the initializer. It ended up with a lot of other changes - could you take a look at the PR if that's roughly what you had in mind?

PR #6267, opened 2024-11-30
WFCORE-7077 SuspendController relies on blocking ServiceActivity behaviour for ordering

pferraro:
https://issues.redhat.com/browse/WFCORE-7077

### DIFF
--- a/server/src/main/java/org/jboss/as/server/suspend/SuspendController.java
+++ b/server/src/main/java/org/jboss/as/server/suspend/SuspendController.java
@@ -90,13 +90,13 @@ public class SuspendController implements ServerSuspendController, SuspendableAc
         }
         CompletableFuture<Void> result = new CompletableFuture<>();
         // Prepare activity groups in priority order, i.e. first -> last
-        this.phaseStage(this.activityGroups, SuspendableActivity::prepare, context, Functions.discardingBiConsumer()).whenComplete((ignored, prepareException) -> {
+        phaseStage(this.activityGroups, SuspendableActivity::prepare, context, Functions.discardingBiConsumer()).whenComplete((ignored, prepareException) -> {
             if (prepareException != null) {
                 result.completeExceptionally(prepareException);
             } else {
                 this.state = State.SUSPENDING;
                 // Suspend activity groups in priority order, i.e. first -> last order
-                this.phaseStage(this.activityGroups, SuspendableActivity::suspend, context, Functions.discardingBiConsumer()).whenComplete((ignore, suspendException) -> {
+                phaseStage(this.activityGroups, SuspendableActivity::suspend, context, Functions.discardingBiConsumer()).whenComplete((ignore, suspendException) -> {
                     if (suspendException != null) {
                         result.completeExceptionally(suspendException);
                     } else {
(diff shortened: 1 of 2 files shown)


server/src/main/java/org/jboss/as/server/suspend/SuspendController.java
```diff
@@ -143,56 +143,57 @@ private Iterator<List<SuspendableActivity>> resumeIterator() {
...
-                    try {
-                        phase.apply(activity, context).whenComplete(activityCompleter);
-                    } catch (Throwable e) {
-                        activityCompleter.accept(null, e);
                     }
                 }
             }
-        }
+        };
+        groupCompleter.accept(null, null);
```
yersan, line 196: Since `groupCompleter.accept` is being invoked here, wouldn't it be clearer if we get the rid of this BiConsumer and just transform it on a method that is directly invoked?

I mean, put everything inside `groupCompleter.accept` in its own method and avoid the usage of the BiConsumer

yersan: Well, I tried simplifying it since it looks complex. I was just thinking whether it would be possible to split it into clearer chunks, my understanding is we are processing groups, and per each group, the activities in that group are in order so I was trying to at least get it split in those two methods to trying to see it more clear, but it does not look simple.

If it is not possible, I have no complaints moving on with it, although looks complex and difficult to understand, at least to me :)

pferraro: I expect that this looks complex because it uses recursion (i.e. the completion handler of each group stage registers itself as the completion handler of the next group stage).
I can replace this with a more verbose iterative implementation if that would be more intuitive.

yersan: @pferraro feel free to ignore the above, now I got the issue and the resolution, in the old code we were moving to the next group even if the list of activities of the previous group has not been completed yet. Now this is fixed since the next group is not processed until all the activities have been completed.

server/src/main/java/org/jboss/as/server/suspend/SuspendController.java
```diff
@@ -143,56 +143,57 @@ private Iterator<List<SuspendableActivity>> resumeIterator() {
...
+                    // Create stage for next group
+                    List<SuspendableActivity> activities = List.copyOf(groups.next());
+                    CompletableFuture<Void> groupStage = new CompletableFuture<>();
+                    groupStage.whenComplete(this);
+                    if (activities.isEmpty()) {
+                        // No activities, complete immediately
+                        groupStage.complete(null);
+                    } else {
+                        // Counter used to determine when to complete group stage
+                        AtomicInteger groupCounter = new AtomicInteger(activities.size());
```
yersan, line 169: Nitpick: This should be named `activityCounter` instead `groupCounter` due to what it represents.

yersan (approved):

server/src/main/java/org/jboss/as/server/suspend/SuspendController.java
pferraro: Let me fix that - surely that's a potential source of confusion.

PR #6268, opened 2024-12-03
[WFCORE-7079] Remove ModuleIdentifier usage from Version module

yersan:
Jira issue: https://issues.redhat.com/browse/WFCORE-7079

### DIFF
--- a/version/src/main/java/org/jboss/as/version/ProductConfig.java
+++ b/version/src/main/java/org/jboss/as/version/ProductConfig.java
@@ -26,7 +26,6 @@ import java.util.jar.Manifest;
 import java.util.stream.Collectors;
 
 import org.jboss.modules.Module;
-import org.jboss.modules.ModuleIdentifier;
 import org.jboss.modules.ModuleLoader;
 
 /**
@@ -224,7 +223,7 @@ public class ProductConfig implements Serializable {
     }
 
     private void setSystemProperties(final Properties propConfProps, final Map providedProperties) {
-        if (propConfProps.size() == 0) {
+        if (propConfProps.isEmpty()) {
             return;
         }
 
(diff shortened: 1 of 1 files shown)


version/src/main/java/org/jboss/as/version/ProductConfig.java
```diff
@@ -262,10 +261,10 @@ private static void safeClose(Closeable c) {
 
     private static class ProductConfProps {
         private final Properties miscProperties;
-        private final ModuleIdentifier productModuleId;
+        private final String productModuleId;
 
         private ProductConfProps(String slot) {
-            this.productModuleId = slot == null ? null : ModuleIdentifier.create("org.jboss.as.product", slot);
+            this.productModuleId = slot == null ? null : "org.jboss.as.product:" + slot;
```
bstansberry, line 267: @yersan This needs to be:

this.productModuleId = slot == null ? null : ("main",equals(slot) ? "org.jboss.as.product" : "org.jboss.as.product:" + slot);

The canonical string representation of a module with slot 'main' is to NOT include ":main".  The ModuleIdentifier class implements that logic. Where we discard its use we need to implement it ourselves.

I want to add utility methods that we can call throughout WF to handle this kind of thing, somewhere in the 'controller' or 'server' modules, but for the 'version' module I think I just needs to be manually coded.

bstansberry (changes requested):

version/src/main/java/org/jboss/as/version/ProductConfig.java
bstansberry: #6270is the proposed utility. But, not usable here. Not really needed either, as the use case here is simple. That PR has a lot of code related to escaping arbitrary inputs but I think this code can assume that any slot value is simple text that doesn't need escaping.

yersan: @bstansberry Done

bstansberry (approved):

bstansberry:
Thanks @yersan

PR #6270, opened 2024-12-04
[WFCORE-7082] Provide utilities to produce canonical representations …

bstansberry:
…of module names

https://issues.redhat.com/browse/WFCORE-7082

I'm not super thrilled with the class name, method names, or putting this in the controller module, so thoughts are welcome.

This shouldn't be in the server module 'deployment' packages as the use case goes beyond DUPs. And if it's not in that package I'd prefer not to add new classes to 'server' that extensions might use.

### DIFF
--- a/controller/src/main/java/org/jboss/as/controller/ModuleIdentifierUtil.java
+++ b/controller/src/main/java/org/jboss/as/controller/ModuleIdentifierUtil.java
@@ -1,0 +1,154 @@
+/*
+ * Copyright The WildFly Authors
+ * SPDX-License-Identifier: Apache-2.0
+ */
+package org.jboss.as.controller;
+
+import org.jboss.dmr.ModelNode;
+import org.jboss.dmr.ModelType;
+
+/**
+ * Provides utilities related to working with names of JBoss Modules modules.
+ */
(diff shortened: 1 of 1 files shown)


controller/src/main/java/org/jboss/as/controller/ModuleIdentifierUtil.java
```diff
@@ -0,0 +1,154 @@
...
+        } else if (moduleSpec.isEmpty()) {
+            throw new IllegalArgumentException("Empty module specification");
+        } else {
+            StringBuilder b = new StringBuilder();
+
+            int c;
+            int i;
+            for(i = 0; i < moduleSpec.length(); i = moduleSpec.offsetByCodePoints(i, 1)) {
+                c = moduleSpec.codePointAt(i);
+                if (c == 92) {
```
scottmarlow, line 37: Does 92 == `\`?
Does  58 == `:`?

bstansberry: Yes

jamezp (approved):

yersan:
Thanks @bstansberry @jamezp

PR #6272, opened 2024-12-05
[WFCORE-7084] Upgrade Installation Manager API to 1.1.0.Final

yersan:
Jira issue: https://issues.redhat.com/browse/WFCORE-7084


yersan:
I forgot we cannot move the JDK version of the installation manager API yet

PR #6274, opened 2024-12-06
[WFCORE-7076] Upgrade maven dependencies to 3.9.9

spyrkob:
Issue: https://issues.redhat.com/browse/WFCORE-7085


spyrkob:
I'm not really sure why those integration tests are failing. I rerun those locally and they seem to be passing fine

PR #6275, opened 2024-12-12
[WFCORE-7087] Support resolution of extension expressions during prep…

bstansberry:
…aration of a domain server launch command

https://issues.redhat.com/browse/WFCORE-7087

Draft; needs a test


bstansberry:
I closed this; I got the CI data I wanted.

bstansberry:
The new test fails because some other test is leaving the secondary HC in restart-required, so when the new test checks that it's properly reloaded into 'running' that fails.

I'm looking into adding a utility to DomainTestSupport to check for such a thing and fail the stop of the naughty test.

PR #6279, opened 2024-12-13
[WFCORE-7089] When setting a policy in the PolicyDefinitions do not a…

jamezp:
…ttempt to refresh the policy if it's null.

https://issues.redhat.com/browse/WFCORE-7089

### DIFF
--- a/elytron/src/main/java/org/wildfly/extension/elytron/PolicyDefinitions.java
+++ b/elytron/src/main/java/org/wildfly/extension/elytron/PolicyDefinitions.java
@@ -199,7 +199,9 @@ class PolicyDefinitions {
                     }
 
                     private void setPolicy(Policy policy) {
-                        policy.refresh();
+                        if (policy != null) {
+                            policy.refresh();
+                        }
                         try {
                             if (WildFlySecurityManager.isChecking()) {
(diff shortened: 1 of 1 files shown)


bstansberry:
@darranl or @fjuma Is this ok?

The NPE fix is clearly fine; I just want to make sure you don't see anything wrong with 'policy' being null in the first place.

bstansberry:
Well, I asked so I'll be patient a bit for an answer, but thinking more this seems fine. The 'policy' param is null because it's what was read from getPolicy() in service start, and on SE 24 there is no particular reason for that call to return anything but null.

darranl (approved):

darranl:
@bstansberry yes this is fine, before Java 24 we would have always had a policy but 24 onwards we will never have a Policy - this service is sufficiently self contained there are not other reasons this would become null.

darranl:
The test failures look unrelated, I was going to kick off CI but I see we have a big queue at the moment.

yersan:
@darranl test failures are old friends and unrelated, they are also failing with similar traces for other PR, the xerces one is recently, but unrelated. I adding a "ready-to-merge"

PR #6281, opened 2024-12-14
[WFCORE-7094] Fix remaining incorrect calls to ModuleIdentifier.getNa…

bstansberry:
…me() from WFCORE-6871

https://issues.redhat.com/browse/WFCORE-7094

PR #6282, opened 2024-12-15
[WFCORE-7095] Deprecate ModuleDependency.getIdentifier() for removal

bstansberry:
https://issues.redhat.com/browse/WFCORE-7095

This builds on #6281


bstansberry:
I put the 'hold' label on this because in the joy of refactoring I forgot about #6247, which should take precedence for that piece of this. This is mostly about different things.

I do prefer not changing the current base.name.slot structure of the service names though, mostly because we don't have to and you never know if someone out would be disrupted in some small way by a change. I'll send a PR to @lvydra about that part.

yersan (approved):

PR #6283, opened 2024-12-16
Fix for WFCORE-7097 and Fix for WFCORE-7098

jfdenise:
ISSUES: 
* https://issues.redhat.com/browse/WFCORE-7097
* https://issues.redhat.com/browse/WFCORE-7098

### DIFF
--- a/testsuite/manualmode/src/test/java/org/jboss/as/test/manualmode/provisioning/InstallationManagerBootTestCase.java
+++ b/testsuite/manualmode/src/test/java/org/jboss/as/test/manualmode/provisioning/InstallationManagerBootTestCase.java
@@ -66,7 +66,10 @@ public class InstallationManagerBootTestCase {
     public static void removeTestModule() {
         if (testModule != null) {
             controller.stop();
-            testModule.remove();
+            // Bootable JAR fully delete the installation
+            if (!AssumeTestGroupUtil.isBootableJar()) {
+                testModule.remove();
+            }
         }
     }
 


bstansberry:
@jamezp Please review as you're kind of an SME on this.

@jfdenise @yersan @jamezp I put the 27.x label on this mostly to get your attention so you can think whether this needs to be in WF 35 or not. I *suspect* the only urgency around this is the intermittent failure WFCORE-7097 mentions, and then the bootable jar failures we are seeing in full WF in ts/int/elytron-oidc-client. But those don't force us to do something quickly if we don't think that's the right thing to do; those both may have workarounds.

bootable-jar/boot/src/main/java/org/wildfly/core/jar/boot/CleanupProcessor.java
```diff
@@ -34,6 +34,9 @@ public static void main(final String[] args) throws Exception {
         final Path installDir = Paths.get(args[0]);
         final int retries = Integer.parseInt(args[1]);
         final Path cleanupMarker = installDir.resolve("wildfly-cleanup-marker");
+        if (Files.notExists(cleanupMarker)) {
+            return;
+        }
```
jamezp, line 39: I'm not sure I follow this. Can you explain why we were seeing an issue if the file does not exist here?

bootable-jar/runtime/src/main/java/org/wildfly/core/jar/runtime/InstallationCleaner.java
```diff
@@ -104,6 +93,23 @@ void cleanup() throws IOException {
...
+        }
+        // In case we have a running process, kill it to avoid it to continue any cleanup.
+        if (newProcess && process != null) {
+            process.destroyForcibly();
+            process.waitFor(environment.getTimeout(), TimeUnit.SECONDS);
+            process = null;
+        }
+        // Do a last cleanup, in case the cleanupMarker still exists (could have been deleted by running process).
+        if (Files.exists(cleanupMarker)) {
+            cleanup();
```
jamezp, line 109: This could potentially launch another process. We should probably just invoke the `deleteDirectory()` at this point.

jamezp (changes requested):

bootable-jar/runtime/src/main/java/org/wildfly/core/jar/runtime/InstallationCleaner.java
jfdenise: Yes, that is done in purpose. On Windows we need the the external process. The cleanup waits until the process terminate (with a timeout).

bootable-jar/runtime/src/main/java/org/wildfly/core/jar/runtime/InstallationCleaner.java
```diff
@@ -83,7 +69,10 @@ public void run() {
      *
      * @throws IOException if an error occurs deleting the directory
      */
-    void cleanup() throws IOException {
+    private synchronized void cleanup() throws InterruptedException, IOException {
```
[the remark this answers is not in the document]
jfdenise: Ok, yes, removing it.

bootable-jar/runtime/src/main/java/org/wildfly/core/jar/runtime/InstallationCleaner.java
```diff
@@ -45,28 +46,13 @@ class InstallationCleaner implements Runnable {
     }
 
     @Override
-    public void run() {
+    public synchronized void run() {
```
yersan, line 49: I am missing something; this task is submitted by a `SingleThreadExecutor` in the BootableJar shutdown hook.
If we are marking it as synchronized, it only means that we could have more than one Bootable Jar instance for the same server home, right?
If that's true, the same marked file is also shared as a marked across the multiple Bootable JAar instances launched from the same home, wouldn't that be an issue at all?

yersan: > this task is submitted by a SingleThreadExecutor in the BootableJar shutdown hook.
If we are marking it as synchronized, it only means that we could have more than one Bootable Jar instance for the same server home, right?

Well, even in that case, we are creating new instances of this InstallationCleaner on each shuftdown hook, so I don't get why the synchronized is required (or nice to have ) at this point.

jfdenise: @yersan , we could have a timeout on the calling thread. The task (running in its own thread) is not yet done, and the calling thread will attempt to do a cleanup again. We need to synchronize at this point to avoid multiple cleanup in //. synchronize enforces it.

yersan: @jfdenise ok, so it is not to allow dealing with muiltiple Bootable JARs from the same server home.

Ok, in that case, should not be InstallationCleaner.cleanup() method be the one that should be synchronized?

That's the method in common with the `InstallationCleaner.run()` and `InstallationCleaner.cleanupTimeout()` which are the entry points for the submitted task and the explicit `cleaner.cleanupTimeout()`

jfdenise: @yersan The key piece is: Files.notExists(cleanupMarker)
We need to have all threads to share a common view on it. So all entry points to it should be synchronized.

bootable-jar/runtime/src/main/java/org/wildfly/core/jar/runtime/InstallationCleaner.java
jfdenise: That is in case, the previous process didn't complete the deletion for some reason (timeout and process forcibly terminated from the caller thread), we start a new process to finalize.

jfdenise:
@yersan , 10 green runs on each platform. I will stop testing.

yersan:
@jamezp Can you review again? Thanks!

jamezp (approved):
I'm approving, but I think we might need to revisit this. I don't want to block fixing CI though.

I don't think this necessarily wrong, I think we're just overly using resources. We have a shutdown hook which attempts to cleanup the resources. On Windows these typically can't be cleaned up until the server process is ended. However, we attempt to wait for the process to end, then we launch another process as a final clean up. I think we could stream line this a bit, but I'd need to think it through a little more.

bootable-jar/runtime/src/main/java/org/wildfly/core/jar/runtime/InstallationCleaner.java
```diff
@@ -153,7 +159,9 @@ private void newProcess() throws IOException {
                 .redirectError(ProcessBuilder.Redirect.INHERIT)
                 .redirectOutput(ProcessBuilder.Redirect.INHERIT)
                 .directory(new File(System.getProperty("user.dir")));
-        builder.start();
+        process = builder.start();
+        process.waitFor(environment.getTimeout(), TimeUnit.SECONDS);
```
yersan, line 163: This also sounds a bit inappropriate, since this method could be invoked directly from the shutdown hook thread. The shutdown hook API says that is inadvisable to attempt any user interaction or to perform a long-running computation in a shutdown hook.

I guess my question would be if this is somehow killed completely by the JVM, if there could be possibilities of having the started process running around.

In any case, if decided, we can move on and see how it behaves in the CI

jfdenise: @yersan , I was thinking to it more, and I think that we shouldn't merge it. Although I am confident on the Linux fix, it requires more on Windows front.

jfdenise:
@jamezp , When testing with a lot of corner cases (with complex scheduling scenarii) on Windows, I came to the conclusion that we need, in the forked process, to wait for the server process to terminate prior to delete the installation. That is the only way to ensure that nothing is left behind and the installation is actually deleted.

jamezp:
> @jamezp , When testing with a lot of corner cases (with complex scheduling scenarii) on Windows, I came to the conclusion that we need, in the forked process, to wait for the server process to terminate prior to delete the installation. That is the only way to ensure that nothing is left behind and the installation is actually deleted.

@jfdenise Yes. That is what it's supposed to be doing currently. I guess I should read the Jira's to see what problem we're trying to solve.

One thing I'm not sure why I'd originally done, is create an `Executor` in the shutdown hook and launch the deletion in a new thread. That seems odd to me. However, the new process is correct. It's the only way on Windows that file locks will be removed.

jamezp:
Looking at the `PermissionsDeploymentTestCase` test, my guess is the `testWithConfiguredMaxBootThreads` is failing because it runs last on Windows. What is likely happening is on the `ServerController.stop()`, the new process is launched and the bootable JAR is being deleted when the new test starts. The new process is deleting files the second test is starting to extract.

The `ServerController` has some specific deleting of files for the bootable JAR. I think this is likely a timing issue.

PR #6287, opened 2024-12-17
[WFCORE-7102]: AccessDeniedException on Windows when using a read-only

ehsavoie:
configuration dir.

* Replacing java.io.File.canWrite() by java.nio.file.FIles.isWritable(Path).

Jira: https://issues.redhat.com/browse/WFCORE-7102

### DIFF
--- a/testsuite/manualmode/src/test/java/org/jboss/as/test/manualmode/management/ReadOnlyModeTestCase.java
+++ b/testsuite/manualmode/src/test/java/org/jboss/as/test/manualmode/management/ReadOnlyModeTestCase.java
@@ -15,6 +15,14 @@ import java.nio.file.attribute.PosixFileAttributeView;
 import java.nio.file.attribute.PosixFilePermission;
 import java.util.HashSet;
 import java.util.Set;
+import java.io.File;
+import java.nio.file.attribute.AclEntry;
+import java.nio.file.attribute.AclEntryPermission;
+import java.nio.file.attribute.AclEntryType;
+import java.nio.file.attribute.AclFileAttributeView;
+import java.nio.file.attribute.FileOwnerAttributeView;
+import java.nio.file.attribute.UserPrincipal;
+import java.util.List;
 
 import jakarta.inject.Inject;
 
@@ -28,7 +36,6 @@ import org.jboss.as.test.shared.TestSuiteEnvironment;
 import org.jboss.dmr.ModelNode;
 import org.junit.After;
 import org.junit.Assert;
-import org.junit.Assume;
 import org.junit.Test;
 import org.junit.runner.RunWith;
 import org.wildfly.core.testrunner.ServerControl;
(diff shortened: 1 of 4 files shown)


yersan:
thanks @ehsavoie , Ideally, we should enable this test for Windows with this PR: https://github.com/wildfly/wildfly-core/blob/main/testsuite/manualmode/src/test/java/org/jboss/as/test/manualmode/management/ReadOnlyModeTestCase.java#L72-L74

But I know there are currently issues setting a read-only directory in the code since the CI won't be able to delete it. We can discuss it with @luck3y if there is any way to enable that permissions for Windows OS

luck3y:
Hi Yeray,

You should be able to create a directory in Windows and mark it read only,
ideally reversing the read only bit when the test is done. I don't think
any special permissions are required as long as you stay within the
checkout directory in CI.

Ken

yersan:
Weird, test still appears as ignored in Windows: https://ci.wildfly.org/buildConfiguration/WildFlyCore_PullRequest_WindowsJdk11/476232?buildTab=tests&name=ReadOnlyModeTestCase&suite=&package=org.jboss.as.test.manualmode.management

WDYT @ehsavoie ?

ehsavoie:
@yersan This is properly fixed now and it works

server/src/main/java/org/jboss/as/server/operations/ServerProcessReloadHandler.java
```diff
@@ -159,6 +159,7 @@ protected ProcessReloadHandler.ReloadContext<RunningModeControl> initializeReloa
         if (serverConfig != null && !environment.getServerConfigurationFile().checkCanFindNewBootFile(serverConfig)) {
             throw ServerLogger.ROOT_LOGGER.serverConfigForReloadNotFound(serverConfig);
         }
+        ServerLogger.ROOT_LOGGER.warn("Reloading with config file " + serverConfig);
```
bstansberry, line 162: @ehsavoie Please remove. I assume this is test cruft.

testsuite/manualmode/src/test/java/org/jboss/as/test/manualmode/management/ReadOnlyModeTestCase.java
```diff
@@ -17,6 +17,14 @@
 import java.util.Set;
 
 import jakarta.inject.Inject;
```
bstansberry, line 19: This goes after java.xxx

testsuite/manualmode/src/test/java/org/jboss/as/test/manualmode/management/ReadOnlyModeTestCase.java
```diff
@@ -120,13 +135,34 @@ public void testReadOnlyConfigurationDirectory() throws Exception {
...
-
+            perms.add(PosixFilePermission.OWNER_READ);
+            perms.add(PosixFilePermission.OWNER_EXECUTE);
+            perms.add(PosixFilePermission.GROUP_READ);
+            perms.add(PosixFilePermission.GROUP_EXECUTE);
+            perms.add(PosixFilePermission.OTHERS_READ);
+            perms.add(PosixFilePermission.OTHERS_EXECUTE);
             Files.getFileAttributeView(roConfigDir, PosixFileAttributeView.class).setPermissions(perms);
-
+        }
```
bstansberry, line 161: The identation from R150-R161 is off.

bstansberry (changes requested):
I've made some trivial change requests. I've not tried to understand the meat of this, so please don't regard fixing my requests as an approval; rely on the others who are reviewing.

OTOH, mergers please don't block on my re-reviewing my trivial items before merging.

bstansberry:
@luck3y Please review.

testsuite/manualmode/src/test/java/org/jboss/as/test/manualmode/management/ReadOnlyModeTestCase.java
```diff
@@ -69,29 +76,37 @@ public void testConfigurationNotUpdated() throws Exception {
...
-        perms.add(PosixFilePermission.OWNER_READ);
-        perms.add(PosixFilePermission.OWNER_EXECUTE);
-        perms.add(PosixFilePermission.GROUP_READ);
-        perms.add(PosixFilePermission.GROUP_EXECUTE);
-        perms.add(PosixFilePermission.OTHERS_READ);
-        perms.add(PosixFilePermission.OTHERS_EXECUTE);
-
-        Files.getFileAttributeView(roConfigDir, PosixFileAttributeView.class).setPermissions(perms);
-
+        if (TestSuiteEnvironment.isWindows()) {
```
luck3y, line 89: This seems odd to me, for Windows an ACL is constructed that denies the listed permissions, which seems correct, but on all other OSes are added as allowed. I'm probably missing something here though?

testsuite/manualmode/src/test/java/org/jboss/as/test/manualmode/management/ReadOnlyModeTestCase.java
```diff
@@ -120,13 +135,34 @@ public void testReadOnlyConfigurationDirectory() throws Exception {
...
+            perms.add(PosixFilePermission.OWNER_EXECUTE);
+            perms.add(PosixFilePermission.GROUP_READ);
+            perms.add(PosixFilePermission.GROUP_EXECUTE);
+            perms.add(PosixFilePermission.OTHERS_READ);
+            perms.add(PosixFilePermission.OTHERS_EXECUTE);
             Files.getFileAttributeView(roConfigDir, PosixFileAttributeView.class).setPermissions(perms);
-
+        }
             PathUtil.deleteRecursively(roConfigDir);
+            if( TestSuiteEnvironment.isWindows()) {
```
luck3y, line 163: I think the delete should be left out, as long as the permissions are reset this will get deleted by normal cleanup and shouldn't have any issues, and a recursive delete floating around on windows gives me the heebee-jeebies :)

luck3y (changes requested):

testsuite/manualmode/src/test/java/org/jboss/as/test/manualmode/management/ReadOnlyModeTestCase.java
ehsavoie: ok, removed

testsuite/manualmode/src/test/java/org/jboss/as/test/manualmode/management/ReadOnlyModeTestCase.java
ehsavoie: fixed

testsuite/manualmode/src/test/java/org/jboss/as/test/manualmode/management/ReadOnlyModeTestCase.java
ehsavoie: ok

server/src/main/java/org/jboss/as/server/operations/ServerProcessReloadHandler.java
ehsavoie: correct

testsuite/manualmode/src/test/java/org/jboss/as/test/manualmode/management/ReadOnlyModeTestCase.java
yersan: @ehsavoie FYI ^^^

ehsavoie: Well 1st, I only looked at Windows but the problem is that the permissions are quite complex with a lot of principals so I reduced the change to the owner only and removing a permission isn't sufficient in Windows you have to DENY it.

yersan (approved):
LGTM, @luck3y Do you have any additional review from your side?

luck3y:
> LGTM, @luck3y Do you have any additional review from your side?

LGTM too, I looked at the code again, its only adding read and execute, no write, so I must have misread it the first time.

luck3y (approved):

yersan:
Thanks @ehsavoie @luck3y @bstansberry

PR #6289, opened 2024-12-18
[WFCORE-7104] Use ModuleDependency.Builder instead of deprecated ModuleDependency ctor in Elytron

ropalka:
https://issues.redhat.com/browse/WFCORE-7104


bstansberry:
@ropalka Can you deal with this as well:

{code}
public class PermissionsParser {

    public static List<PermissionFactory> parse(final XMLStreamReader reader, final ModuleLoader loader, final ModuleIdentifier identifier)
{code}

I mean add a new method that takes String and then @Deprecate(forRemoval=true) the current one.

There is code in WF that calls this method so if we add an alternative we can migrate that code to the alternative and then delete the deprecated one. I don't see any need to keep the deprecated one longer than it takes to change the WF use; it's not API meant for use outside our security layer.

bstansberry (approved):

yersan:
Thanks @ropalka @bstansberry

PR #6292, opened 2024-12-18
[WFCORE-7105] Use ModuleDependency.Builder instead of deprecated ModuleDependency ctor in Server

ropalka:
https://issues.redhat.com/browse/WFCORE-7107

### DIFF
--- a/server/src/main/java/org/jboss/as/server/deployment/module/ManifestDependencyProcessor.java
+++ b/server/src/main/java/org/jboss/as/server/deployment/module/ManifestDependencyProcessor.java
@@ -104,7 +104,7 @@ public final class ManifestDependencyProcessor implements DeploymentUnitProcesso
                     }
                 }
 
-                final ModuleDependency dependency = new ModuleDependency(dependencyLoader, dependencyId, optional, export, services, true);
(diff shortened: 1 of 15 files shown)


bstansberry:
Hi @ropalka Please reopen when nearly all the 35.x PRs in WildFly are merged.

server/src/main/java/org/jboss/as/server/deployment/module/ManifestDependencyProcessor.java
```diff
@@ -104,7 +104,7 @@ public void deploy(final DeploymentPhaseContext phaseContext) throws DeploymentU
                     }
                 }
 
-                final ModuleDependency dependency = new ModuleDependency(dependencyLoader, dependencyId, optional, export, services, true);
+                final ModuleDependency dependency = ModuleDependency.Builder.of(dependencyLoader, dependencyId.toString()).setOptional(optional).setExport(export).setImportServices(services).build();
```
bstansberry, line 107: This needs .setUserSpecified(true)

bstansberry (changes requested):

server/src/main/java/org/jboss/as/server/deployment/module/ManifestDependencyProcessor.java
ropalka: Good catch! It is fixed now @bstansberry .

bstansberry (approved):

PR #6293, opened 2024-12-19
[WFCORE-7108] Add a ModuleIdentifierUtil method to parse a module ide…

bstansberry:
…ntifier into name and slot and process the tuple

https://issues.redhat.com/browse/WFCORE-7108

Needed for https://github.com/wildfly/wildfly/pull/18561

### DIFF
--- a/controller/src/main/java/org/jboss/as/controller/ModuleIdentifierUtil.java
+++ b/controller/src/main/java/org/jboss/as/controller/ModuleIdentifierUtil.java
@@ -4,6 +4,8 @@
  */
 package org.jboss.as.controller;
 
+import java.util.function.BiFunction;
+
 import org.jboss.dmr.ModelNode;
 import org.jboss.dmr.ModelType;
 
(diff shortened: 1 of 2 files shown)


controller/src/main/java/org/jboss/as/controller/ModuleIdentifierUtil.java
```diff
@@ -21,32 +23,88 @@ public final class ModuleIdentifierUtil {
...
+
+
+    /**
+     * Parses the given module identifier into name and optional slot elements, passing those to the given
+     * function and returning the result of that function.
+     * <p/>
+     *
+     * @param moduleIdentifier an identifier for a module. Cannot be {@code null}
+     * @param function a function to apply to the module's name and optional slot. Cannot be {@code null}.
+     *                 The slot value passed the function may be null if the identifier does not contain one.
```
ropalka, line 53: "passed the function"  should be "passed to the function"

controller/src/main/java/org/jboss/as/controller/ModuleIdentifierUtil.java
```diff
@@ -21,32 +23,88 @@ public final class ModuleIdentifierUtil {
...
```
ropalka, line 70: same as above

ropalka (approved):
Except minor wording issue in javadoc it looks good to me.

controller/src/main/java/org/jboss/as/controller/ModuleIdentifierUtil.java
bstansberry: Thanks; fixed.

controller/src/main/java/org/jboss/as/controller/ModuleIdentifierUtil.java
bstansberry: Thanks; fixed.

PR #6301, opened 2024-12-22
[WFCORE-7095][WFCORE-7108][WFCORE-7111][WFCORE-7112][WFCORE-7113][WFCORE-7114] Module identifier in api fixes

bstansberry:
This is an aggregate of:

#6282 
#6293
#6298 
#6299 
#6300 
#6302

In total this deals with what I think are all the places where WF Core exposes ModuleIdentifier in API that full WildFly or an external extension might use. It deprecates all such places and provides an alternative for all but two things that I don't think external modules will use.[1]

I realize this is a lot to deal with at the last minute for WF Core 27, and **if it's too much that's ok**. Doing this though will give us more freedom to change things right away in Core 28 / WF 36, and gets the deprecation out of the way for downstream. (These APIs are not supported for external use downstream, but we'd likely want to deprecate them anyway before removing them in some future release.)

#6282 is the most substantive. #6293 is already approved. The others are new but pretty simple.

[1] Those two things deprecated without alternatives provided in this PR are:
Attachments.CLASS_PATH_ENTRIES which we should just break (by changing its list element type) with no migration path. It's not used outside the kernel and shouldn't be.
Attachments.ADDITIONAL_ANNOTATION_INDEXES for which we will need to do a simple migration dance in a couple core Betas to let the one 'ee' subsystem use migrate.

@yersan @ropalka @dmlloy FYI.

### DIFF
--- a/controller/src/main/java/org/jboss/as/controller/ModuleIdentifierUtil.java
+++ b/controller/src/main/java/org/jboss/as/controller/ModuleIdentifierUtil.java
@@ -4,6 +4,8 @@
  */
 package org.jboss.as.controller;
 
+import java.util.function.BiFunction;
+
 import org.jboss.dmr.ModelNode;
 import org.jboss.dmr.ModelType;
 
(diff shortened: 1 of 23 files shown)


yersan:
"Windows - JDK 11 (Pull Request) - merge" failure seems unrelated, same error trace in other Jobs

yersan:
Thanks @bstansberry

PR #6302, opened 2024-12-22
[WFCORE-7114] Deprecate use of ModuleIdentifier in ModuleDefinition

bstansberry:
https://issues.redhat.com/browse/WFCORE-7114

### DIFF
--- a/server/src/main/java/org/jboss/as/server/moduleservice/ModuleDefinition.java
+++ b/server/src/main/java/org/jboss/as/server/moduleservice/ModuleDefinition.java
@@ -25,14 +25,29 @@ public class ModuleDefinition {
     private final Set<ModuleDependency> dependencies;
     private final ModuleSpec moduleSpec;
 
+
+    /** @deprecated use {@link #ModuleDefinition(String, Set, ModuleSpec)} */
+    @Deprecated(forRemoval = true)
     public ModuleDefinition(final ModuleIdentifier moduleIdentifier, final Set<ModuleDependency> dependencies, final ModuleSpec moduleSpec) {
         this.moduleIdentifier = moduleIdentifier;
         this.dependencies = dependencies;
         this.moduleSpec = moduleSpec;
     }
 
+    public ModuleDefinition(final String moduleIdentifier, final Set<ModuleDependency> dependencies, final ModuleSpec moduleSpec) {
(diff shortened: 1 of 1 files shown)


server/src/main/java/org/jboss/as/server/moduleservice/ModuleDefinition.java
```diff
@@ -25,16 +25,31 @@ public class ModuleDefinition {
...
     }
 
+    public ModuleDefinition(final String moduleIdentifier, final Set<ModuleDependency> dependencies, final ModuleSpec moduleSpec) {
+        this.moduleIdentifier = ModuleIdentifier.fromString(moduleIdentifier);  // inefficient but this is unused. When we switch the use to this we'll store the string.
+        this.dependencies = dependencies;
+        this.moduleSpec = moduleSpec;
+    }
+
+    /** @deprecated use {@link #getModuleName()}  */
+    @Deprecated(forRemoval = true)
```
yersan, line 44: As a rule of thumb, we are always asking for the since attribute on all @Deprecated annotations so we can make an idea without looking at the Git history when it was deprecated.

This can be done later though

server/src/main/java/org/jboss/as/server/moduleservice/ModuleDefinition.java
```diff
@@ -25,16 +25,31 @@ public class ModuleDefinition {
...
     public ModuleIdentifier getModuleIdentifier() {
         return moduleIdentifier;
     }
 
+    public String getModuleName() {
+        return moduleIdentifier.toString(); // inefficient but this is unused. When we switch the use to this we'll store the string.
+    }
```
yersan, line 51: I wonder if getModuleName is what we would want. The ModuleIdentifier is composed by `name` and `slot` attributes. So `getModuleName()` returning the result of `moduleIdentifier.toString()` maybe is not what we want since it could drive into confusion since someone could be expecting to get just the ModuleIdentifier name. 

As additional detail, the JBoss Modules documentation still makes a distinction about what a module name is (https://jboss-modules.github.io/jboss-modules/manual/#names) and what are module identifiers https://jboss-modules.github.io/jboss-modules/manual/#slots

yersan: getModuleIdentifier() is already taken, so ....

yersan (approved):

PR #6304, opened 2024-12-23
[WFCORE-7116] Remove ModuleIdentifier from Extension handling

bstansberry:
https://issues.redhat.com/browse/WFCORE-7116

### DIFF
--- a/cli/src/main/java/org/jboss/as/cli/impl/ExtensionsLoader.java
+++ b/cli/src/main/java/org/jboss/as/cli/impl/ExtensionsLoader.java
@@ -26,10 +26,10 @@ import org.jboss.as.cli.Util;
 import org.jboss.as.cli.handlers.CommandHandlerWithHelp;
 import org.jboss.as.cli.impl.aesh.CLICommandRegistry;
 import org.jboss.as.controller.client.ModelControllerClient;
+import org.jboss.as.controller.client.helpers.JBossModulesNameUtil;
 import org.jboss.dmr.ModelNode;
 import org.jboss.dmr.Property;
 import org.jboss.modules.ModuleClassLoader;
-import org.jboss.modules.ModuleIdentifier;
 import org.jboss.modules.ModuleLoadException;
 import org.jboss.modules.ModuleLoader;
 
@@ -102,9 +102,7 @@ class ExtensionsLoader {
      * Using the client, iterates through the available domain management model extensions
      * and tries to load CLI command handlers from their modules.
      *
-     * @param registry
      * @param address
-     * @param client
      */
     void loadHandlers(ControllerAddress address) throws CommandLineException, CommandLineParserException {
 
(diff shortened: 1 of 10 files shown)


controller/src/main/java/org/jboss/as/controller/logging/ControllerLogger.java
```diff
@@ -3809,4 +3809,7 @@ OperationFailedRuntimeException capabilityAlreadyRegisteredInContext(String capa
     @LogMessage(level = WARN)
     @Message(id = 517, value = "There are multiple Parallel Boot Operations.")
     void multipleParallelBootOperation();
+
+    @Message(id = 518, value = "Resource /extension=%s uses a non-canonical module name; use the canonical form %s")
```
yersan, line 3813: Perhaps, I would add in the message a short hint about what a canonical form is. Something like:

"Resource /extension=%s uses a non-canonical module name; use the canonical form %s. The canonical representation includes slot information in the module name if the slot is not 'main'"

cli/src/main/java/org/jboss/as/cli/impl/ExtensionsLoader.java
```diff
@@ -213,6 +210,12 @@ private void addError(String msg) {
         }
     }
 
+    private static String canonicalize(String moduleName) {
+        // This isn't a full canonicalization, but extension module names are unlikely to be non-canonical
+        // in any way beyond maybe a trailing ":main". So just check for and strip any trailing ":main"
```
yersan, line 215: question: Did you consider adding a ModuleIdentifierUtil version to the wildfly-controller-client module so we can have the utility available on the client code?

host-controller/src/main/java/org/jboss/as/domain/controller/operations/ApplyExtensionsHandler.java
```diff
@@ -116,6 +115,8 @@ public void execute(OperationContext context, ModelNode operation) throws Operat
             if (resourceAddress.size() == 1 && resourceAddress.getElement(0).getKey().equals(EXTENSION)) {
                 final String module = resourceAddress.getElement(0).getValue();
                 if (appliedExtensions.add(module)) {
+                    // We know 'module' is a canonical module name as the primary is our version or later
+                    // and we require canonical names for extension resources
```
yersan, line 119: +1

controller/src/main/java/org/jboss/as/controller/logging/ControllerLogger.java
bstansberry: Done.

cli/src/main/java/org/jboss/as/cli/impl/ExtensionsLoader.java
bstansberry: I hadn't no, but I updated this to move the main code to controller-client and have the controller module class just delegate to the client class.

I wouldn't want to do too much of this kind of thing, as its kind of an edge case for controller-client. But after almost 15 years we have hardly any of these, so I think it's ok. ;-)

yersan (approved):

yersan:
Thanks @bstansberry

PR #6309, opened 2024-12-27
[WFCORE-7124] Remove use of server-side Stability type from the CLI

bstansberry:
https://issues.redhat.com/browse/WFCORE-7124

### DIFF
--- a/cli/pom.xml
+++ b/cli/pom.xml
@@ -274,10 +274,6 @@
             </exclusions>
         </dependency>
         <dependency>
-            <groupId>org.wildfly.core</groupId>
-            <artifactId>wildfly-version</artifactId>
-        </dependency>
-        <dependency>
             <groupId>org.wildfly.security</groupId>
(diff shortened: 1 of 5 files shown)


bstansberry:
@jfdenise and @yersan If we don't like this approach we can just drop the tab completion suggestion and the user must know what to type. The protocol. controller-client, cli and embedded modules are client-side modules and can't have deps on server-side modules. Those four should be able to have an independent lifecycle (e.g. we should be able to put them in a separate code base from all the server side code.) Also, as the TODO I removed hints, suggesting all options is incorrect as the process being embedded may only support 'default'. But EmbedServerHandler and EmbedHostController handler can't know when making suggestions what the process will support if the CLI isn't running as a modular app in the same installation as the eventual embedded process. The only way to know is to accept the user's jboss-home input and find the target installation.

jfdenise (approved):

yersan (approved):

yersan:
Thanks @bstansberry @jfdenise

PR #6312, opened 2024-12-28
[WFCORE-7127] Replace Attachments.CLASS_PATH_ENTRIES

bstansberry:
https://issues.redhat.com/browse/WFCORE-7127

Note that I added a commit to this to remove a bunch of IDE warns.

### DIFF
--- a/server/src/main/java/org/jboss/as/server/deployment/Attachments.java
+++ b/server/src/main/java/org/jboss/as/server/deployment/Attachments.java
@@ -38,7 +38,6 @@ import org.jboss.vfs.VirtualFile;
 /**
  * @author <a href="mailto:david.lloyd@redhat.com">David M. Lloyd</a>
  */
-@SuppressWarnings("deprecation")
 public final class Attachments {
 
     //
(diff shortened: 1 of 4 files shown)


server/src/main/java/org/jboss/as/server/deployment/Attachments.java
```diff
@@ -125,14 +124,12 @@ public final class Attachments {
      * <p/>
      * <strong>Note: This is only meant for use within the server kernel.</strong>
      *
-     * @deprecated this will either be changed incompatibly (to provide a string) or removed altogether in the next WildFly Core major.
      */
-    @Deprecated(forRemoval = true)
-    public static final AttachmentKey<AttachmentList<ModuleIdentifier>> CLASS_PATH_ENTRIES = AttachmentKey.createList(ModuleIdentifier.class);
+    public static final AttachmentKey<AttachmentList<String>> CLASS_PATH_MODULES = AttachmentKey.createList(String.class);
```
bstansberry, line 128: I've realized this doesn't need to be in this class, so I'll move it to the only package where it's used.

ropalka (approved):

bstansberry:
Thanks @ropalka

PR #6324, opened 2025-01-20
[WFCORE-7142] Upgrade BouncyCastle from 1.79 to 1.80

yersan:
Jira issue: https://issues.redhat.com/browse/WFCORE-7142

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -200,7 +200,7 @@
         <version.org.apache.maven.resolver>1.1.1</version.org.apache.maven.resolver>
         <version.org.apache.sshd>2.14.0</version.org.apache.sshd>
         <version.org.apache.velocity>2.3</version.org.apache.velocity>
-        <version.org.bouncycastle>1.79</version.org.bouncycastle>
+        <version.org.bouncycastle>1.80</version.org.bouncycastle>
         <version.org.codehaus.plexus.plexus-utils>3.5.1</version.org.codehaus.plexus.plexus-utils>
         <version.org.eclipse.jgit>6.10.0.202406032230-r</version.org.eclipse.jgit>
         <version.org.eclipse.parsson>1.1.7</version.org.eclipse.parsson>


yersan:
@jamezp @jimma Could you approve from your side?

darranl:
I think we need to pick this up again this year and ideally look at how to separate BouncyCastle from the project that calls it .

If to start we can wrap it behind some form of SPI maybe we can write sufficient tests against the SPI to demonstrate it has not changed.

jamezp:
> I think we need to pick this up again this year and ideally look at how to separate BouncyCastle from the project that calls it .
> 
> If to start we can wrap it behind some form of SPI maybe we can write sufficient tests against the SPI to demonstrate it has not changed.

This makes sense to be @darranl

jamezp (approved):

jamezp (approved):

yersan:
@jimma Could you review this?

yersan:
> > I think we need to pick this up again this year and ideally look at how to separate BouncyCastle from the project that calls it .
> > If to start we can wrap it behind some form of SPI maybe we can write sufficient tests against the SPI to demonstrate it has not changed.
> 
> This makes sense to be @darranl

I'm not sure if I am following correctly. Are the affected components that use Bouncy Castle getting it from a WildFly Core version or are they getting Bouncy Castle as an independent dependency?

My understanding is we want WildFly Core to provide an SPI with a set of tests in the WildFly Core testsuite to verify the blackguard compatibility of that SPI, and then use that SPI in the third-party components (RESTEasy and Webservices). The ultimate goal is to avoid wanting to get third-party components' test suite results when we are upgrading Bouncy Castle in WildFly Core. Is that the plan you suggest?

jamezp:
@yersan I can only speak for RESTEasy and my personal preference there is to remove Bouncy Castle all together. I don't think RESTEasy can have a dependency on WildFly Core, so if the SPI lived there it wouldn't likely be able to use it.

yersan:
> so if the SPI lived there it wouldn't likely be able to use it.

Ok, in such a case, I don't get how an SPI can resolve this requirement of asking whether this upgrade could affect third party testsuites, I am missing a piece of the puzzle.


@jimma in case you missed the other ping, could you approve this from your component point of view?

jimma (approved):

yersan:
Thanks @jamezp @darranl @jimma

PR #6325, opened 2025-01-21
[WFCORE-7143] Consolidate HostControllerBootstrap and EmbeddedHostCon…

bstansberry:
…trollerBootstrap

https://issues.redhat.com/browse/WFCORE-7143

### DIFF
--- a/host-controller/src/main/java/org/jboss/as/host/controller/AbstractHostControllerBootstrap.java
+++ b/host-controller/src/main/java/org/jboss/as/host/controller/AbstractHostControllerBootstrap.java
@@ -1,0 +1,164 @@
+/*
+ * Copyright The WildFly Authors
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+package org.jboss.as.host.controller;
+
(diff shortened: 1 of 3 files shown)


bstansberry:
@yersan @luck3y @jamezp FYI.

jamezp:
The failure seems like it might be related.

luck3y (approved):
This looks like it should be fine to me, don't see anything obvious :)

bstansberry:
> The failure seems like it might be related.

@jamezp This PR only touches host controller code and none of the failures are in tests running a domain; they are use standalone servers. They all fail intermittently.

bstansberry:
I put the 'hold' label on this as I just realized the existing HostControllerBootstrap doesn't have a try/catch block in bootstrap. That's only in EmbeddedHostControllerBootstrap. So I'll need to fix that.

bstansberry:
I fixed that so this is ready to go.

yersan (approved):

yersan:
Thanks @bstansberry @jamezp

PR #6327, opened 2025-01-27
[WFCORE-7151] Upgrade WildFly Galleon Plugins to 7.3.1.Final

yersan:
Jira issue: https://issues.redhat.com/browse/WFCORE-7151


yersan:
This upgrade is the current version used in WildFly, so to me this is just an implicit approval for a major from the component lead

PR #6331, opened 2025-01-30
[WFCORE-7153] Don't test reload-enhanced if it's not an available operation

bstansberry:
This builds on #6329, adding a commit to allow the tests to work downstream.

https://issues.redhat.com/browse/WFCORE-7153


yersan:
Thanks @bstansberry

PR #6333, opened 2025-02-06
[WFCORE-6846] Upgrade commons-daemon to 1.4.0

lvydra:
https://issues.redhat.com/browse/WFCORE-6846

Verified that WildFly standalone / WildFly domain can be installed as a Windows Service and started/stopped successfully with commons-daemon updated to 1.4.0


yersan:
Thanks @lvydra

PR #6340, opened 2025-02-14
[WFCORE-6843] Logging a WARN if a deployment's runtime name doesn't have an extension

khermano:
Issue: https://issues.redhat.com/browse/WFCORE-6843

### DIFF
--- a/host-controller/src/main/java/org/jboss/as/domain/controller/operations/deployment/ServerGroupDeploymentAddHandler.java
+++ b/host-controller/src/main/java/org/jboss/as/domain/controller/operations/deployment/ServerGroupDeploymentAddHandler.java
@@ -27,6 +27,7 @@ import org.jboss.as.repository.ContentRepository;
 import org.jboss.as.repository.HostFileRepository;
 import org.jboss.as.server.deployment.ModelContentReference;
 import org.jboss.dmr.ModelNode;
+import org.jboss.logging.Logger;
 
 /**
  * Handles addition of a deployment to a server group.
@@ -39,6 +40,7 @@ public class ServerGroupDeploymentAddHandler implements OperationStepHandler {
 
     private final HostFileRepository fileRepository;
     private final ContentRepository contentRepository;
+    private static final Logger logger = Logger.getLogger("org.jboss.as.domain.controller.operations.deployment");
 
     public ServerGroupDeploymentAddHandler(final HostFileRepository fileRepository, ContentRepository contentRepository) {
         this.fileRepository = fileRepository;
@@ -94,6 +96,9 @@ public class ServerGroupDeploymentAddHandler implements OperationStepHandler {
             Resource root = context.readResourceFromRoot(PathAddress.EMPTY_ADDRESS);
             ModelNode domainDeployment = root.getChild(PathElement.pathElement(DEPLOYMENT, deploymentName)).getModel();
             String runtimeName = getRuntimeName(deploymentName, deployment, domainDeployment);
+            if (!runtimeName.contains(".")) {
+                logger.warnf("Deployment '%s' has a runtime name '%s' without an extension. This may cause issues in some environments.", deploymentName, runtimeName);
+            }
             PathAddress sgAddress = address.subAddress(0, address.size() - 1);
             Resource serverGroup = root.navigate(sgAddress);
             for (Resource.ResourceEntry re : serverGroup.getChildren(DEPLOYMENT)) {
--- a/server/src/main/java/org/jboss/as/server/deployment/DeploymentHandlerUtil.java
+++ b/server/src/main/java/org/jboss/as/server/deployment/DeploymentHandlerUtil.java
@@ -49,6 +49,7 @@ import org.jboss.as.server.deployment.transformation.DeploymentTransformer;
 import org.jboss.as.server.deploymentoverlay.DeploymentOverlayIndex;
 import org.jboss.as.server.logging.ServerLogger;
 import org.jboss.dmr.ModelNode;
+import org.jboss.logging.Logger;
 import org.jboss.msc.service.LifecycleEvent;
 import org.jboss.msc.service.LifecycleListener;
 import org.jboss.msc.service.ServiceBuilder;
@@ -69,6 +70,7 @@ import org.jboss.vfs.VirtualFile;
 public class DeploymentHandlerUtil {
 
     private static final OperationContext.AttachmentKey<AnnotationIndexSupport> ANNOTATION_INDEX_SUPPORT = OperationContext.AttachmentKey.create(AnnotationIndexSupport.class);
+    private static final Logger logger = Logger.getLogger("org.jboss.as.server.deployment");
 
     static class ContentItem {
         // either hash or <path, relativeTo, isArchive>
@@ -142,6 +144,9 @@ public class DeploymentHandlerUtil {
 
             context.addStep(new OperationStepHandler() {
                 public void execute(OperationContext context, ModelNode operation) {
+                    if (!deploymentUnitName.contains(".")) {
+                        logger.warnf("Deployment '%s' has a runtime name '%s' without an extension. This may cause issues in some environments.", managementName, deploymentUnitName);
+                    }
                     final ServiceName deploymentUnitServiceName = Services.deploymentUnitName(deploymentUnitName);
                     final ServiceRegistry serviceRegistry = context.getServiceRegistry(true);
                     final ServiceController<?> deploymentController = serviceRegistry.getService(deploymentUnitServiceName);


khermano:
Thank you for your very quick response and sorry for the late response, I overlooked that. I will look at it now. Thanks for the information. :)

server/src/main/java/org/jboss/as/server/logging/ServerLogger.java
```diff
@@ -115,6 +115,12 @@ public interface ServerLogger extends BasicLogger {
      */
     ServerLogger DEPRECATED_DEP_LOGGER = Logger.getMessageLogger(MethodHandles.lookup(), ServerLogger.class, "org.jboss.as.dependency.deprecated");
 
+    /**
+     * A logger with the category of {@code org.jboss.as.server.deployment.warning}.
+     * <strong>Usage:</strong> Use this in code related to deployment's warnings
+     */
+    ServerLogger DEPLOYMENT_WARN_LOGGER = Logger.getMessageLogger(MethodHandles.lookup(), ServerLogger.class, "org.jboss.as.server.deployment.warning");
+
```
yersan, line 123: We should not create specific root loggers for specific levels. We create root loggers and then we configure the levels from outside, for example in a logger configuration file.

In this file there is already a logger for deployment, use it instead.

host-controller/src/main/java/org/jboss/as/domain/controller/logging/DomainControllerLogger.java
```diff
@@ -56,6 +56,12 @@ public interface DomainControllerLogger extends BasicLogger {
      */
     DomainControllerLogger HOST_CONTROLLER_LOGGER = Logger.getMessageLogger(MethodHandles.lookup(), DomainControllerLogger.class, "org.jboss.as.host.controller");
 
+    /**
+     * A logger with the category of {@code org.jboss.as.domain.controller.operations.deployment.warning}.
+     * <strong>Usage:</strong> Use this in code related to deployment's warnings
+     */
+    DomainControllerLogger DEPLOYMENT_WARN_LOGGER = Logger.getMessageLogger(MethodHandles.lookup(), DomainControllerLogger.class, "org.jboss.as.domain.controller.operations.deployment.warning");
+
```
yersan, line 64: We should not create specific root loggers for specific levels. We create root loggers and then we configure the levels from outside, for example in a logger configuration file.

host-controller/src/main/java/org/jboss/as/domain/controller/operations/deployment/ServerGroupDeploymentAddHandler.java
```diff
@@ -94,6 +95,9 @@ static void validateRuntimeNames(String deploymentName, OperationContext context
             Resource root = context.readResourceFromRoot(PathAddress.EMPTY_ADDRESS);
             ModelNode domainDeployment = root.getChild(PathElement.pathElement(DEPLOYMENT, deploymentName)).getModel();
             String runtimeName = getRuntimeName(deploymentName, deployment, domainDeployment);
+            if (!runtimeName.contains(".")) {
+                DEPLOYMENT_WARN_LOGGER.deploymentsRuntimeNameWithoutExtension(deploymentName, runtimeName);
```
yersan, line 99: Use `DomainControllerLogger.ROOT_LOGGER` here instead.

yersan (changes requested):

host-controller/src/main/java/org/jboss/as/domain/controller/logging/DomainControllerLogger.java
khermano: I understand, I just wasn't sure how to implement this correctly since the description of the issue was talking about using distinct logger. So, can I just use the ROOT_LOGGER for the domain and DEPLOYMENT_LOGGER for the standalone @yersan ?

yersan: > I just wasn't sure how to implement this correctly since the description of the issue was talking about using distinct logger for this

... and I did not properly read the Jira description :(

Got it, it was requested to use a different logger so users can turn off the logger under demand.

In such a case, I would suggest using the following (for sure open to any other idea for the name):

"org.jboss.as.server.deployment.runtime.namecheck"
"org.jboss.as.domain.controller.operations.deployment.runtime.namecheck"

yersan: DEPLOYMENT_NAMECHECK_LOGGER is ok, variable name is no an issue, if you read it and identify its meaning, that's ok

server/src/main/java/org/jboss/as/server/deployment/DeploymentHandlerUtil.java
```diff
@@ -142,6 +143,9 @@ public static void deploy(final OperationContext context, final ModelNode operat
 
             context.addStep(new OperationStepHandler() {
                 public void execute(OperationContext context, ModelNode operation) {
+                    if (!deploymentUnitName.contains(".")) {
+                        DEPLOYMENT_NAMECHECK_LOGGER.deploymentsRuntimeNameWithoutExtension(managementName, deploymentUnitName);
+                    }
```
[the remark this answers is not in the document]
khermano: Hi @yersan,
I looked at it, and I pushed my changes, but I would like to ask about redeploy in domain mode. 
I am not entirely sure if I got that right, but I assume that in domain mode, when talking about redeployment, we mean operation deployment-overlay, and it has to do with DomainDeploymentOverlayRedeployLinksHandler, right? 
If that is the case, I think it is not possible to change the runtime name of the deployment, or am I mistaken?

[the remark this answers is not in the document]
yersan: > when talking about redeployment, we mean operation deployment-overlay, and it has to do with DomainDeploymentOverlayRedeployLinksHandler, right?

It is not related to the overlay functionality. We are talking here about the `redeploy` operation under `/server-group=*/deployment=*` resource. The handler is `ServerGroupDeploymentRedeployHandler` but according to the code, it seems that the operations there are redirected directly to the managed servers, so I think we are not going to be able to do something easy with it. Being a redeploy that does not change the runtime name, I think we will be ok with it. In addition, the server handler will be already prepared to log the trace in the managed server JVM, so, that's fine with me.

yersan (changes requested):
@khermano I took anothe look today, domain mode is a bit more tricky, I've added some comments related. 

In addition, you also have to check the `ServerGroupDeploymentReplaceHandler`, which is in charge of `/server-group=*:replace-deployment` operation and acepts a runtime-name as an attribute.

khermano:
Hello @yersan ,
I looked into it and I changed the warning (the commit is not here yet) so it also contains the name of the server group, but I am not sure about removing one of the warnings.
I know it looks like a mistake if the same warning is printed twice in the row, but it's because it is associated with two different operations - ADD and DEPLOY, and I do not think I can remove it from either of them without risking in some cases the runtime name will not be checked at all.
Should I try to implement some counter or check the output or something?
Regarding REPLACE operation, the runtime name attribute in this operation exists, but I can't see it used anywhere. Even when I tried to use this operation with a new runtime-name, the operation for replacement was successful but the new runtime name was never used.
I would suggest that I will aplicate the check anyway and create an issue regarding this unused attribute. What do you think? Or is there something that I overlooked regarding all of this?
Thank you in advance for your help and I am really sorry that it took so long for me to look into it.

yersan:
@khermano:

> I know it looks like a mistake if the same warning is printed twice in the row, but it's because it is associated with two different operations - ADD and DEPLOY, and I do not think I can remove it from either of them without risking in some cases the runtime name will not be checked at all.

The Add operation adds the deployment as a server group resource, but the actual deployment on the servers will occur only if you add the resource with `enable=true`. I think that's the key to distinguishing when we have to show the log trace: the add handler should display the log only if it is executed with `enable=true`. The deploy handler should always display the log.

> Should I try to implement some counter or check the output or something?

You shouldn't, those approaches tend to be more fragile and complex to cope with all the cases, in addition, we do not keep any memory facility in the operation context, and the handlers should work independently.

> Regarding REPLACE operation, the runtime name attribute in this operation exists, but I can't see it used anywhere. Even when I tried to use this operation with a new runtime-name, the operation for replacement was successful but the new runtime name was never used.
I would suggest that I will aplicate the check anyway and create an issue regarding this unused attribute. What do you think? Or is there something that I overlooked regarding all of this?

Ideally, we have to create an issue describing what you have found with the replace operation, could you do that? it should have a short description and how to reproduce it. 

[WFCORE-6843](https://issues.redhat.com/browse/WFCORE-6843) is an enhancement, and the one you found is a bug. The bug should be blocking the enhancement, we should fix the bug first, and then add the WFCORE-6843 changes on top of the changes that fix the bug. That can be done in a single PR with two commits, or fix the bug, wait until gets resolved and then continue with WFCORE-6843

PR #6357, opened 2025-02-28
[WFCORE-7175] Remove ModuleIdentifier usage from server/moduleservice

yersan:
Jira issue: https://issues.redhat.com/browse/WFCORE-7175

### DIFF
--- a/server/src/main/java/org/jboss/as/server/deployment/module/ModuleExtensionListProcessor.java
+++ b/server/src/main/java/org/jboss/as/server/deployment/module/ModuleExtensionListProcessor.java
@@ -10,7 +10,6 @@ import java.util.LinkedHashSet;
 import java.util.List;
 import java.util.Set;
 
-import org.jboss.as.server.logging.ServerLogger;
 import org.jboss.as.server.deployment.AttachmentList;
 import org.jboss.as.server.deployment.Attachments;
 import org.jboss.as.server.deployment.DeploymentPhaseContext;
@@ -19,9 +18,9 @@ import org.jboss.as.server.deployment.DeploymentUnitProcessingException;
 import org.jboss.as.server.deployment.DeploymentUnitProcessor;
 import org.jboss.as.server.deployment.DeploymentUtils;
 import org.jboss.as.server.deployment.Services;
+import org.jboss.as.server.logging.ServerLogger;
 import org.jboss.as.server.moduleservice.ExtensionIndex;
 import org.jboss.as.server.moduleservice.ServiceModuleLoader;
-import org.jboss.modules.ModuleIdentifier;
 import org.jboss.modules.ModuleLoader;
 import org.jboss.modules.filter.PathFilters;
 import org.jboss.msc.service.ServiceController;
@@ -47,15 +46,15 @@ public final class ModuleExtensionListProcessor implements DeploymentUnitProcess
         final ServiceController<?> controller = phaseContext.getServiceRegistry().getRequiredService(Services.JBOSS_DEPLOYMENT_EXTENSION_INDEX);
         final ExtensionIndex index = (ExtensionIndex) controller.getValue();
         final List<ResourceRoot> allResourceRoots = DeploymentUtils.allResourceRoots(deploymentUnit);
-        final Set<ServiceName> nextPhaseDeps = new HashSet<ServiceName>();
+        final Set<ServiceName> nextPhaseDeps = new HashSet<>();
 
-        final Set<ModuleIdentifier> allExtensionListEntries = new LinkedHashSet<>();
+        final Set<String> allExtensionListEntries = new LinkedHashSet<>();
 
         for (ResourceRoot resourceRoot : allResourceRoots) {
             final AttachmentList<ExtensionListEntry> entries = resourceRoot.getAttachment(Attachments.EXTENSION_LIST_ENTRIES);
             if (entries != null) {
                 for (ExtensionListEntry entry : entries) {
-                    final ModuleIdentifier extension = index.findExtension(entry.getName(), entry.getSpecificationVersion(),
+                    final String extension = index.findExtensionAsString(entry.getName(), entry.getSpecificationVersion(),
                             entry.getImplementationVersion(), entry.getImplementationVendorId());
                     if (extension != null) {
                         allExtensionListEntries.add(extension);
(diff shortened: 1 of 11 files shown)


server/src/main/java/org/jboss/as/server/deployment/module/ModuleExtensionListProcessor.java
```diff
@@ -83,12 +83,12 @@ public void deploy(final DeploymentPhaseContext phaseContext) throws DeploymentU
...
 
-        for (ModuleIdentifier extension : allExtensionListEntries) {
-            ModuleDependency dependency = ModuleDependency.Builder.of(moduleLoader, extension.toString()).setImportServices(true).setUserSpecified(true).build();
+        for (String extension : allExtensionListEntries) {
+            ModuleDependency dependency = ModuleDependency.Builder.of(moduleLoader, extension).setImportServices(true).setUserSpecified(true).build();
             dependency.addImportFilter(PathFilters.getMetaInfSubdirectoriesFilter(), true);
             dependency.addImportFilter(PathFilters.getMetaInfFilter(), true);
             moduleSpecification.addLocalDependency(dependency);
-            nextPhaseDeps.add(ServiceModuleLoader.moduleSpecServiceName(extension.toString()));
+            nextPhaseDeps.add(ServiceModuleLoader.moduleSpecServiceName(JBossModulesNameUtil.parseCanonicalModuleIdentifier(extension)));
```
yersan, line 91: Note to the reviewer:

I think parsing the extension here to ensure a canonicalized name is optional, at this stage, the "extension" string should be something without a slot, since it comes from Attachments.EXTENSION_LIST_ENTRIES which is populated at https://github.com/wildfly/wildfly-core/blob/035d41260346840f0073a90875997f2b8f16df17/server/src/main/java/org/jboss/as/server/deployment/module/ManifestExtensionListProcessor.java#L76-L77

So, I am not sure this parse here will never do anything. I left it because it sounds more robust, but I think we can also remove it.

Looks like a minor detail though, so if there is no objection, I'm fine going as it is.

bstansberry: My preference is to remove it. I'd prefer we be rigorous about canonicalizing as close to the questionable external input as possible, javadoc as requiring canonical names any APIs that are called from the class that canonicalized external input, and otherwise regard names as canonical. 

We can end up recanonicalizing all over the place. It's not a super-cheap operation.

server/src/main/java/org/jboss/as/server/moduleservice/ExtensionIndex.java
```diff
@@ -18,28 +18,63 @@ public interface ExtensionIndex {
...
     ModuleIdentifier findExtension(String name, String minSpecVersion, String minImplVersion, String requiredVendorId);
 
+    /**
+     * Find an extension, returning {@code null} if no matching extension is available.
+     *
+     * @param name             the extension name
+     * @param minSpecVersion   the minimum spec version to match, or {@code null} to match any
+     * @param minImplVersion   the minimum implementation version to match, or {@code null} to match any
+     * @param requiredVendorId the vendor ID to require, or {@code null} to match any
+     * @return the resource root of the first matched extension
```
bstansberry, line 38: The javadoc of this and the method it replaces is confusing.  It doesn't seem to be returning a 'resource root'. And the top level javadoc says 'Find an extension', which is unclear.  Something like "Find the [identifier|name] of an extension module...' is better.

TBH I'm not sure what an extension module is in this deployment context. I can figure it out based on how it's used, and have done so probably 5 times over 15 years. If it's clear to you now what it is it would be good to improve the javadoc in a few key places (maybe ExtensionInfo, ExtensionIndex, any relevant constant in Attachments), or file a WFCORE to do so.

bstansberry (changes requested):

server/src/main/java/org/jboss/as/server/moduleservice/ExtensionIndex.java
yersan: > The javadoc of this and the method it replaces is confusing. It doesn't seem to be returning a 'resource root'. And the top level javadoc says 'Find an extension', which is unclear. Something like "Find the [identifier|name] of an extension module...' is better.

Done.

> TBH I'm not sure what an extension module is in this deployment context. I can figure it out based on how it's used, and have done so probably 5 times over 15 years. If it's clear to you now what it is it would be good to improve the javadoc in a few key places (maybe ExtensionInfo, ExtensionIndex, any relevant constant in Attachments), or file a WFCORE to do so.

I'll file a WFCORE. So far I have just an overall idea after looking a bit more at the code related, but i will need a bit more time to play a bit with real examples and understand it well. This entire extension module functionality is related to the Java Extension Mechanism, which is indeed no longer supported. JAR files placed at WFLY_HOME/standalone/lib/ext will be identified as available Java Extensions that your deployment (using entries in the manifest) can declare as needed.

server/src/main/java/org/jboss/as/server/moduleservice/ExtensionIndexService.java
```diff
@@ -203,8 +223,14 @@ public synchronized ModuleIdentifier findExtension(final String name, final Stri
         return null;
     }
 
+    @Deprecated(forRemoval = true, since = "28.0.0")
     public static ModuleIdentifier moduleIdentifier(final String name, final String minSpecVersion,
             final String minImplVersion, final String requiredVendorId) {
+        return ModuleIdentifier.create(moduleIdentifierStr(name, minSpecVersion, minImplVersion, requiredVendorId));
+    }
+
+    public static String moduleIdentifierStr(final String name, final String minSpecVersion,
```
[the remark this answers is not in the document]
yersan: Done.

server/src/main/java/org/jboss/as/server/moduleservice/ExtensionIndex.java
yersan: Opened https://issues.redhat.com/browse/WFCORE-7182 to tack it, however, I guss we should revisit this functionality since Java Extension Mechanism was removed since JDK 9 https://openjdk.org/jeps/220

bstansberry: @yersan Thanks for the reminder and the JIRA. Note that while JEPS 220 did remove most aspects of the extension mechanism, the bits we rely upon (basically some constants in the java.util.jar.Attributes class) still exist. And Jakarta EE 10 specifies the required support around them:

https://jakarta.ee/specifications/platform/10/jakarta-platform-spec-10.0#a2945

Particularly section 8.2.2.

I expect we'll be supporting EE 10 in one way or another for a long time, so I don't think this kernel code will be going anywhere soon!

bstansberry (approved):

PR #6364, opened 2025-03-12
[WFCORE-7183] Upgrade SLF4J from 2.0.16 to 2.0.17

yersan:
Jira issue: https://issues.redhat.com/browse/WFCORE-7183

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -235,7 +235,7 @@
         <version.org.mockito>3.10.0</version.org.mockito>
         <version.org.jmockit>1.49</version.org.jmockit>
         <version.org.projectodd.vdx>1.1.6</version.org.projectodd.vdx>
-        <version.org.slf4j>2.0.16</version.org.slf4j>
+        <version.org.slf4j>2.0.17</version.org.slf4j>
         <version.org.syslog4j>0.9.30</version.org.syslog4j>
         <version.org.wildfly.client.config>1.0.1.Final</version.org.wildfly.client.config>
         <version.org.wildfly.common>1.7.0.Final</version.org.wildfly.common>


jamezp:
This looks like some kind of license issue. This may have come from https://github.com/qos-ch/slf4j/commit/0def25ebfa0e546525fb90aa8d5946d16f26c561. I'll have to dig into how the license plugin works and see what the issue is.

bstansberry:
@jamezp This is handled via the KnownLicenses class in the plugin, which has this stuff for MIT:

```
    private static final String MIT_0_NAME = "MIT-0";
    private static final String MIT_0_URL = "https://spdx.org/licenses/MIT-0.html";
    private static final License MIT_0_LICENSE = createLicense(MIT_0_NAME, MIT_0_URL);

    private static final String MIT_NAME = "MIT License";
    private static final String MIT_URL = "https://www.opensource.org/licenses/MIT";
    private static final License MIT_LICENSE = createLicense(MIT_NAME, MIT_URL);
```

and

```
        KNOWN_LICENSES.put(MIT_0_NAME.toLowerCase(), MIT_0_LICENSE);

        KNOWN_LICENSES.put(MIT_NAME.toLowerCase(), MIT_LICENSE);
```

So, if there is some new, reasonable, way a pom expresses a license it can be added there.

But, the direct way to deal with this is via an entry in https://github.com/wildfly/wildfly-core/blob/main/core-feature-pack/common/src/main/resources/license/core-feature-pack-common-licenses.xml

jamezp:
I guess I'll have to look deeper at the SLF4J license, because it's POM shows it being ASLv2.

jamezp:
Oh wait, not the `jcl-over-slf4j` and `log4j-ver-slf4j` are ASLv2. The others are MIT, sorry. I'll look at this today.

PR #6370, opened 2025-03-19
WFCORE-7188 Fix required child detection.

pferraro:
https://issues.redhat.com/browse/WFCORE-7188

### DIFF
--- a/subsystem/src/main/java/org/wildfly/subsystem/resource/ResourceDescriptor.java
+++ b/subsystem/src/main/java/org/wildfly/subsystem/resource/ResourceDescriptor.java
@@ -22,6 +22,7 @@ import java.util.stream.Stream;
 
 import org.jboss.as.controller.AttributeDefinition;
 import org.jboss.as.controller.OperationContext;
+import org.jboss.as.controller.OperationFailedException;
 import org.jboss.as.controller.OperationStepHandler;
 import org.jboss.as.controller.PathElement;
 import org.jboss.as.controller.ResourceRegistration;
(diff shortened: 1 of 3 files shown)


yersan (commented):
The Jira description states the following:

"This affects the ability of the add operation to act as a replace operation."

Does it make sense to add a test case that reproduces the issue, or, otherwise, add a reproduce step in the Jira so anyone can get a clear picture about the issue this is trying to resolve?

pferraro:
There is a reproducer here (if you remove the workaround commit): https://github.com/wildfly/wildfly/pull/18823

pferraro:
To explain the general issue:
If a ResourceDescriptor of some resource defines any require child resources (or required singleton child resources), then its add operation will auto-create add operation for the required child resources.  Add operations for required child resources (or singleton child resources) are meant to behave as a replace operation.
This was implemented in wildfly-clustering-common and is utilised in a number of subsystems.
While migrating these subsystem to wildfly-subsystem, I noticed that the second part was not happening correctly (somewhere, in the renaming of classes, this got botched).

For example:
* /subsystem=distributable-web/infinispan-session-management=foo:add(...) will automatically add an affinity=primary-owner child resource if it is missing, since its ResourceDescriptor defines this as one of its required singleton children.
* If a subsubsequent add operation arrives for the same resource type, e.g. /subsystem=distributable-web/infinispan-session-management=foo/affinity=ranked:add(...), then the existing affinity=primary-owner child (added in the previous step) is meant to be auto-removed prior to executing the add operation.  This is what I meant by an add operation behaving like a replace operation.

yersan:
Thanks @pferraro

PR #6372, opened 2025-03-20
[WFCORE-7152] Add RBAC enforcement information to the OperationContext javadoc

yersan:
... and harmless code cleanup and fix typos

Jira issue: https://issues.redhat.com/browse/WFCORE-7152

### DIFF
--- a/controller/src/main/java/org/jboss/as/controller/OperationContext.java
+++ b/controller/src/main/java/org/jboss/as/controller/OperationContext.java
@@ -414,7 +414,10 @@ public interface OperationContext extends ExpressionResolver, CapabilityServiceD
     ModelNode getCurrentOperationParameter(String name, boolean nullable);
 
     /**
-     * Get a read only view of the managed resource registration.  The registration is relative to the operation address.
+     * Get a read only view of the managed resource registration. The registration is relative to the operation address.
+     * <p>
+     * This method validates that the caller has the necessary permissions to read the resource registration associated
+     * with the currently executing step.
      *
      * @return the model node registration
      */
@@ -422,6 +425,9 @@ public interface OperationContext extends ExpressionResolver, CapabilityServiceD
 
     /**
      * Get a mutable view of the managed resource registration.  The registration is relative to the operation address.
+     * <p>
+     * This method validates that the caller has the necessary permissions to read and modify the resource registration associated
+     * with the currently executing step.
      *
      * @return the model node registration
      */
@@ -444,6 +450,8 @@ public interface OperationContext extends ExpressionResolver, CapabilityServiceD
      * caller intends to make any modifications to any object reachable, directly or indirectly, from the returned
      * {@link ServiceRegistry}. This includes modifying any {@link ServiceController}, {@link org.jboss.msc.Service},
      * {@link org.jboss.msc.service.Service}, {@link org.jboss.msc.value.Value} or any object reachable from a value.
+     * <p>
+     * This method validates that the caller has the necessary permissions to read or modify the runtime state.
      *
      * @param modify {@code true} if the operation may be modifying any object reachable directly or indirectly from
      *                           the returned {@link ServiceRegistry}, {@code false} otherwise
@@ -456,6 +464,9 @@ public interface OperationContext extends ExpressionResolver, CapabilityServiceD
      * Initiate a service removal.  If the step is not a runtime operation handler step, an exception will be thrown.  Any
      * subsequent step which attempts to add a service with the same name will block until the service removal completes.
      * The returned controller may be used to attempt to cancel a removal in progress.
+     * <p>
+     * This method performs an authorization check to ensure that the caller has the necessary permissions
+     * to modify the runtime state.
      *
      * @param name the service to remove
      * @return the controller of the service to be removed if service of given name exists; null otherwise
@@ -466,6 +477,9 @@ public interface OperationContext extends ExpressionResolver, CapabilityServiceD
     /**
      * Initiate a service removal.  If the step is not a runtime operation handler step, an exception will be thrown.  Any
      * subsequent step which attempts to add a service with the same name will block until the service removal completes.
+     * <p>
+     * This method performs an authorization check to ensure that the caller has the necessary permissions
+     * to modify the runtime state.
      *
      * @param controller the service controller to remove
      * @throws UnsupportedOperationException if the calling step is not a runtime operation step
@@ -514,6 +528,9 @@ public interface OperationContext extends ExpressionResolver, CapabilityServiceD
     /**
      * Create a new resource, relative to the executed operation address.  Since only one operation
      * may write at a time, this operation may block until other writing operations have completed.
+     * <p>
+     * This method performs an authorization check to ensure that the caller has the necessary permissions
+     * to create the resource.
      *
      * @param address the (possibly empty) address where the resource should be created. Address is relative to the
      *                address of the operation being executed
@@ -526,7 +543,9 @@ public interface OperationContext extends ExpressionResolver, CapabilityServiceD
     /**
      * Add a new resource, at the executed operation address.  Since only one operation
      * may write at a time, this operation may block until other writing operations have completed.
-     *
+     * <p>
+     * This method performs an authorization check to ensure that the caller has the necessary permissions
+     * to add the resource.
      * @param address the (possibly empty) address where the resource should be added. Address is relative to the
      *                address of the operation being executed
      * @param toAdd the new resource
@@ -538,6 +557,9 @@ public interface OperationContext extends ExpressionResolver, CapabilityServiceD
     /**
      * Add a new resource, at to the executed operation address.  Since only one operation
      * may write at a time, this operation may block until other writing operations have completed.
+     * <p>
+     * This method performs an authorization check to ensure that the caller has the necessary permissions
+     * to add the resource.
      *
      * @param address the (possibly empty) address where the resource should be added. Address is relative to the
      *                address of the operation being executed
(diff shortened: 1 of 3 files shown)


controller/src/main/java/org/jboss/as/controller/OperationContext.java
```diff
@@ -414,15 +414,20 @@ public interface OperationContext extends ExpressionResolver, CapabilityServiceD
     ModelNode getCurrentOperationParameter(String name, boolean nullable);
 
     /**
-     * Get a read only view of the managed resource registration.  The registration is relative to the operation address.
-     *
+     * Get a read only view of the managed resource registration. The registration is relative to the operation address.
+     * <p>
+     * This method validates that the caller has the necessary permissions to read the resource registration associated
+     * with currently executing step.
+     * <p>
```
rhusar: This `<p>` is unnecessary. There is no more content. Applies multiple times.

rhusar (approved):
LGTM, left a nitpick.

controller/src/main/java/org/jboss/as/controller/OperationContext.java
```diff
@@ -514,7 +527,9 @@ default ServiceTarget getServiceTarget() throws UnsupportedOperationException {
     /**
      * Create a new resource, relative to the executed operation address.  Since only one operation
      * may write at a time, this operation may block until other writing operations have completed.
-     *
+     * <p>
+     * This method performs an authorization check to ensure that the caller has the necessary permissions
+     * to create the resource.
```
rhusar, line 533: Nitpick: we usually leave an empty space in Javadoc between text and @ params.

yersan: Fixed

controller/src/main/java/org/jboss/as/controller/OperationContext.java
yersan: Fixed

controller/src/main/java/org/jboss/as/controller/OperationContext.java
rhusar: There are 4-5 more occurrences of that in this change still...

controller/src/main/java/org/jboss/as/controller/OperationContext.java
```diff
@@ -237,7 +237,7 @@ public interface OperationContext extends ExpressionResolver, CapabilityServiceD
     boolean hasFailureDescription();
 
     /**
-     * Get the node into which the details of server results in a multi-server managed domain operation should be written.
+     * Get the node into which the details of server result in a multi-server managed domain operation should be written.
```
bstansberry: 'results' is correct.

IDEA thinks it's a verb and says to make it 'result' but it's a noun and a plural one -- there are results from multiple servers.

controller/src/main/java/org/jboss/as/controller/OperationContext.java
```diff
@@ -414,15 +414,20 @@ public interface OperationContext extends ExpressionResolver, CapabilityServiceD
     ModelNode getCurrentOperationParameter(String name, boolean nullable);
 
     /**
-     * Get a read only view of the managed resource registration.  The registration is relative to the operation address.
+     * Get a read only view of the managed resource registration. The registration is relative to the operation address.
+     * <p>
+     * This method validates that the caller has the necessary permissions to read the resource registration associated
+     * with currently executing step.
```
bstansberry: s/with currently/with the currently/g

controller/src/main/java/org/jboss/as/controller/OperationContext.java
```diff
@@ -414,15 +414,20 @@ public interface OperationContext extends ExpressionResolver, CapabilityServiceD
...
      * @return the model node registration
      */
     ImmutableManagementResourceRegistration getResourceRegistration();
 
     /**
      * Get a mutable view of the managed resource registration.  The registration is relative to the operation address.
-     *
+     * <p>
+     * This method validates that the caller has the necessary permissions to read and modify the resource registration associated
+     * with currently executing step.
```
bstansberry: s/with currently/with the currently/g

controller/src/main/java/org/jboss/as/controller/OperationContext.java
```diff
@@ -456,6 +463,9 @@ public interface OperationContext extends ExpressionResolver, CapabilityServiceD
      * Initiate a service removal.  If the step is not a runtime operation handler step, an exception will be thrown.  Any
      * subsequent step which attempts to add a service with the same name will block until the service removal completes.
      * The returned controller may be used to attempt to cancel a removal in progress.
+     * <p>
+     * This method performs an authorization check to ensure that the caller has the necessary permissions
+     * to modify the runtime state before removing the service.
```
bstansberry: I suggest dropping 'before removing the service'.  People will reasonably assume that checks are done before the action.

Either that or put it at the start "Before removing the service, this method...".  Having it here reads as if what's required is permission to "modify the runtime state before removing the service" as if someone might not have that but would have permission to "modify the runtime state *after* removing the service".

controller/src/main/java/org/jboss/as/controller/OperationContext.java
```diff
@@ -466,6 +476,9 @@ public interface OperationContext extends ExpressionResolver, CapabilityServiceD
     /**
      * Initiate a service removal.  If the step is not a runtime operation handler step, an exception will be thrown.  Any
      * subsequent step which attempts to add a service with the same name will block until the service removal completes.
+     * <p>
+     * This method performs an authorization check to ensure that the caller has the necessary permissions
+     * to modify the runtime state before performing the service removal.
```
bstansberry: Move or remove 'before performing the service removal'.

controller/src/main/java/org/jboss/as/controller/OperationContext.java
```diff
@@ -550,8 +571,10 @@ default ServiceTarget getServiceTarget() throws UnsupportedOperationException {
 
     /**
      * Get the resource for read only operations, relative to the executed operation address. Reads never block.
-     * If a write action was previously performed, the value read will be from an uncommitted copy of the the management model.<br/>
-     *
+     * If a write action was previously performed, the value read will be from an uncommitted copy of the management model.<br/>
```
bstansberry: The br is not needed.

controller/src/main/java/org/jboss/as/controller/OperationContext.java
```diff
@@ -550,8 +571,10 @@ default ServiceTarget getServiceTarget() throws UnsupportedOperationException {
...
+     * <p>
+     * This method performs an authorization check to ensure that the caller has the necessary permissions
+     * to add the resource.
```
bstansberry, line 578: Add a ```<p>```

controller/src/main/java/org/jboss/as/controller/OperationContext.java
```diff
@@ -700,7 +736,7 @@ default ServiceTarget getServiceTarget() throws UnsupportedOperationException {
     boolean revertResourceRestarted(PathAddress resource, Object owner);
 
     /**
-     * Resolves any expressions in the passed in ModelNode.
+     * Resolves any expressions passed in the ModelNode.
      * Expressions may either represent system properties or vaulted date. For vaulted data the format is
      * ${VAULT::vault_block::attribute_name::sharedKey}
```
bstansberry: Since you are touching this...

Expressions may either represent system properties or environment variables or can be extension expressions. Extension expressions must be of the form {@code ${extensionidentifier::someextensionspecificdetails}}, where
 * {@code extensionidentifier} is some string that identifies the desired extension.

BUT

Perhaps just remove these two lines as this isn't really the place to explain expression grammar. The 'represent system properties or environment variables' bit is hardly complete, and then there are unmentioned things like nested expressions etc.

controller/src/main/java/org/jboss/as/controller/OperationContextImpl.java
```diff
@@ -569,6 +569,11 @@ public boolean isResourceServiceRestartAllowed() {
         return operationHeaders.getContextFlags().contains(ContextFlag.ALLOW_RESOURCE_SERVICE_RESTART);
     }
 
+    /**
+     * Gets the resource registration for the address of the currently executing step.
+     * This method validates that the caller has the necessary permissions to modify the model.
+     * @return the resource registration
+     */
```
bstansberry: This is ok, but can it simple be ```@Override```? In any case ```@Override``` should be there.

controller/src/main/java/org/jboss/as/controller/OperationContext.java
yersan: Fixed

controller/src/main/java/org/jboss/as/controller/OperationContext.java
yersan: Fixed

controller/src/main/java/org/jboss/as/controller/OperationContext.java
yersan: Fixed

controller/src/main/java/org/jboss/as/controller/OperationContext.java
yersan: Fixed. Removed, now seems much better without it, makes it simpler, and yes, something the people would assume, there is no need to add it.

controller/src/main/java/org/jboss/as/controller/OperationContext.java
yersan: Fixed

controller/src/main/java/org/jboss/as/controller/OperationContext.java
yersan: Fixed

controller/src/main/java/org/jboss/as/controller/OperationContext.java
yersan: Fixed

controller/src/main/java/org/jboss/as/controller/OperationContext.java
yersan: > this isn't really the place to explain expression grammar

I agree with that reasoning, I've just removed them

controller/src/main/java/org/jboss/as/controller/OperationContextImpl.java
yersan: Yes, I just missed that this was an overrided method

PR #6376, opened 2025-03-21
WFCORE-7196 Fix illegal duplicate children/attributes in XML descriptions of elytron subsystem.

pferraro:
https://issues.redhat.com/browse/WFCORE-7196

Extracted from https://github.com/wildfly/wildfly-core/pull/6261

### DIFF
--- a/elytron/src/main/java/org/wildfly/extension/elytron/PolicyParser.java
+++ b/elytron/src/main/java/org/wildfly/extension/elytron/PolicyParser.java
@@ -19,15 +19,14 @@ class PolicyParser {
 
     PersistentResourceXMLDescription parser_1_0 = PersistentResourceXMLDescription.builder(PathElement.pathElement(POLICY))
             .setNameAttributeName(PolicyDefinitions.DEFAULT_POLICY.getName())
-            .addAttribute(PolicyDefinitions.DEFAULT_POLICY)
             .addAttribute(JaccPolicyDefinition.POLICIES, AttributeParsers.UNWRAPPED_OBJECT_LIST_PARSER, AttributeMarshallers.OBJECT_LIST_UNWRAPPED)
             .addAttribute(CustomPolicyDefinition.POLICIES, AttributeParsers.UNWRAPPED_OBJECT_LIST_PARSER, AttributeMarshallers.OBJECT_LIST_UNWRAPPED)
             .build();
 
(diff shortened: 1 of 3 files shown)


elytron/src/main/java/org/wildfly/extension/elytron/PolicyParser.java
```diff
@@ -19,15 +19,14 @@ class PolicyParser {
 
     PersistentResourceXMLDescription parser_1_0 = PersistentResourceXMLDescription.builder(PathElement.pathElement(POLICY))
             .setNameAttributeName(PolicyDefinitions.DEFAULT_POLICY.getName())
-            .addAttribute(PolicyDefinitions.DEFAULT_POLICY)
```
yersan, line 22: Based on the Jira description, it is said that PolicyParser defines DEFAULT_POLICY attribute twice. However, I can only see it added once, indeed the one you are removing.

So, what does this `.addAttribute()` removal pretend to do?

pferraro: The duplication is in the previous line.

pferraro: It is parsed once to determine the resource path value, and again as if it were an normal attribute.

pferraro: This kind of misuse should be flagged by PersistentResourceXMLDescription - see jira description of original issue.

yersan: @pferraro thanks for the explanations!

> This effectively duplicates the previous line.

Yes, but the previous line only sets the default value for the "name" field when used for wildcard resources.
So, you are saying that `.setNameAttributeName()` should not be used together with `.addAttribute()` for wildcard resources when they are talking about the same attribute?

> It is parsed once to determine the resource path value, and again as if it were an normal attribute.

ok, probably that's the part I have not identified in the code yet

pferraro: @yersan
> So, you are saying that .setNameAttributeName() should not be used together with .addAttribute() for wildcard resources when they are talking about the same attribute?

This particular resource uses DEFAULT_POLICY as its resource path value, not as an attribute, so the latter usage is incorrect.

yersan (approved):

PR #6384, opened 2025-04-01
[WFCORE-7203] Restore the original behaviour of RoleMapping login mod…

pedro-hos:
…ule in the mapped role mapper

Jira: https://issues.redhat.com/browse/WFCORE-7203
Requires: https://issues.redhat.com/browse/ELY-2892


darranl:
@pedro-hos thank you very much for your pull requests, I think we should find some time to catch up as there are some additional pieces we are going to need to cover.

I think we need to pick a lower stability level either Preview or Community and use that to bring it into WildFly.
This will also need some work bumping one of the schemas so we can include it at the correct stability level, as we maintain schemas for backwards compatibility between releases we can't modify the existing ones.

PR #6388, opened 2025-04-03
[WFCORE-5917][JBEAP-29581] Windows standalone scripts test, powershell script fix

lvydra:
https://issues.redhat.com/browse/JBEAP-29581
https://issues.redhat.com/browse/WFCORE-5917

Related to: https://github.com/wildfly/wildfly-core/pull/5840


lvydra:
@yersan I have added a test functionality for Windows standalone scripts and fixed powershell script, standalone.bat seems to be unaffected.

lvydra:
Thanks @yersan, done

yersan:
@lvydra This one got a conflict due to #6391, it needs a rebase from your side, thanks

lvydra:
Thanks @yersan, rebased.

core-feature-pack/common/src/main/resources/content/bin/common.ps1
```diff
@@ -254,7 +254,9 @@ Param(
   	$PROG_ARGS += "-Dlogging.configuration=file:$logFileProperties"
   }
   $PROG_ARGS += "-Djboss.home.dir=$JBOSS_HOME"
-  $PROG_ARGS += "-Djboss.server.base.dir=$global:JBOSS_BASE_DIR"
+  if (-not($SERVER_OPTS -match "-Djboss.server.base.dir")) {
+    $PROG_ARGS += "-Djboss.server.base.dir=$global:JBOSS_BASE_DIR"
+  }
```
yersan, line 259: That's ok, maybe the alternative would have been to just completely drop this line, since I think  JBOSS_BASE_DIR will always contain the default value. Just FYI, as a finding doing the review.

lvydra:
Hi @yersan ,
I have looked at it a bit more, if there is way to use JDK API, and it seems that in Windows, command line arguments are not available via a non-privileged API for other processes (https://bugs.openjdk.org/browse/JDK-8176725).
As for batch, it seems that the former batch solutions have been deprecated in favor of PowerShell (https://learn.microsoft.com/en-us/windows/win32/wmisdk/wmic).
So, at least I have added a check if PowerShell is available.

testsuite/scripts/src/test/java/org/wildfly/scripts/test/StandaloneScriptTestCase.java
```diff
@@ -100,6 +119,46 @@ void testScript(final ScriptProcess script) throws InterruptedException, Timeout
...
         }
 
+        // test WFCORE-5917
+        if (!script.getContainerHome().toString().contains(" ")) {
+            ProcessHandle handle = script.toHandle();
+            List<ProcessHandle> children = handle.children().collect(Collectors.toList());
+
+            String launchCommand = "";
+            if (TestSuiteEnvironment.isWindows()) {
+                if (isPowerShellSupported()) {
```
yersan, line 129: Use `Shell.POWERSHELL.isSupported()` instead, it is not necessary to repeat the same logic in this class by creating `isPowerShellSupported` method

testsuite/scripts/src/test/java/org/wildfly/scripts/test/StandaloneScriptTestCase.java
```diff
@@ -100,6 +119,46 @@ void testScript(final ScriptProcess script) throws InterruptedException, Timeout
             Assert.assertFalse("Did not expect to find -Djava.security.manager=allow in the JVM parameters.", stdout.contains("-Djava.security.manager=allow"));
         }
 
+        // test WFCORE-5917
+        if (!script.getContainerHome().toString().contains(" ")) {
```
yersan, line 123: Better if you trim before checking the blank space:

`!script.getContainerHome().toString().trim().contains(" ")`

testsuite/scripts/src/test/java/org/wildfly/scripts/test/StandaloneScriptTestCase.java
```diff
@@ -132,4 +191,50 @@ public ModelNode call() throws Exception {
...
+                throw new RuntimeException(e);
+            } finally {
+                if (process != null) {
+                    process.destroyForcibly();
+                }
+            }
+        } catch (IOException e) {
+            throw new UncheckedIOException(e);
+        }
+    }
```
yersan, line 239: Can be removed if you use `Shell.POWERSHELL.isSupported()`

testsuite/scripts/src/test/java/org/wildfly/scripts/test/StandaloneScriptTestCase.java
yersan, line 129: Could you also add an @Test method with a very short explanation that assumes that if we are running on Windows PowerShell is available? The test will only contain that assumption and won't execute the scripts, but we can use it as a way to report whether WFCORE-5917 was verified. At this moment, if we are in Windows and PowerShell is not enabled, the code path will not go inside this `if`, the `launchCommand` will never be asserted, but the test will report green.

Probably the best approach here, instead of adding this @Test would have been to have a separate test case to verify WFCORE-5917, but that could be even more convoluted since we have to repeat it for each shell; the main problem is that this `testScript` method is already being used to test multiple cases. So, my feeling is it will be enough if we have a new @Test method that assumes if we are in Windows, PowerShell is enabled, that can be used to report whether WFCORE-5917 was verified.

Do you see any other idea @lvydra to report this case?

yersan (changes requested):

core-feature-pack/common/src/main/resources/content/bin/common.ps1
yersan: Ok, @lvydra , based on your comment about this, I resolve this thread then. Thanks for investigating whether there is any possibility by using the JDK API

lvydra:
Thanks @yersan, updated, is this how you meant it?

testsuite/scripts/src/test/java/org/wildfly/scripts/test/StandaloneScriptTestCase.java
```diff
@@ -80,6 +87,15 @@ public void testBatchScriptJavaOptsEscaped() throws Exception {
         ServerConfigurator.removeJavaOpts(ServerHelper.JBOSS_HOME, STANDALONE_BASE_NAME, escapedVariable);
     }
 
+    @Test
+    public void testPowershellSupportedOnWindows() throws Exception {
```
yersan, line 91: Maybe `testWFCORE5917InWindows` instead?

testsuite/scripts/src/test/java/org/wildfly/scripts/test/StandaloneScriptTestCase.java
```diff
@@ -80,6 +87,15 @@ public void testBatchScriptJavaOptsEscaped() throws Exception {
...
+        // WFCORE-5917
+        // On Windows, command-line arguments of the process can be obtained only via PowerShell.
+        // For Windows, where PowerShell is not available, testScript WFCORE-5917 related assertions are skipped.
+        // The purpose of this test is to determine whether WFCORE-5917 has been verified.
+        Assume.assumeTrue(TestSuiteEnvironment.isWindows() && Shell.POWERSHELL.isSupported());
+    }
```
yersan, line 97: Well, it will be more:
```
if (TestSuiteEnvironment.isWindows()) {
Assume.assumeTrue(Shell.POWERSHELL.isSupported())
}
```

So the test will report ignored only when the testsuite is running in Windows and PowerShell is not enabled. If the testsuite is running in Linux, we don't need to report WFCORE-5917 verification was ignored.

lvydra:
Thanks @yersan, updated.

yersan:
Thanks @lvydra

PR #6397, opened 2025-04-15
WFCORE-7224 LoopbackAddressInterfaceCriteria needs to be more robust

rhusar:
https://issues.redhat.com/browse/WFCORE-7224

### DIFF
--- a/controller/src/main/java/org/jboss/as/controller/interfaces/LoopbackAddressInterfaceCriteria.java
+++ b/controller/src/main/java/org/jboss/as/controller/interfaces/LoopbackAddressInterfaceCriteria.java
@@ -23,7 +23,7 @@ public class LoopbackAddressInterfaceCriteria extends AbstractInterfaceCriteria 
 
     private static final long serialVersionUID = 1L;
 
-    private String address;
+    private final String address;
     private InetAddress resolved;
     private boolean unknownHostLogged;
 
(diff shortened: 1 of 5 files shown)


rhusar:
ROFL a test case that verifies a log message is misspelled... hilarious.

> Expected: a string containing "WARN  [org.jboss.as.controller.management-operation] (Controller Boot Thread) WFLYCTL0512: No resource exists at address '/socket-binding-group=standard-sockets/remote-destination-outbound-socket-binding=non-existent-binding'. Ignoring the remove opr

bstansberry (approved):

luck3y (approved):

bstansberry:
Thanks @rhusar and @luck3y

controller/src/main/java/org/jboss/as/controller/interfaces/LoopbackAddressInterfaceCriteria.java
```diff
@@ -61,18 +61,19 @@ public synchronized InetAddress getAddress() throws UnknownHostException {
...
     }
 
-        /**
+    /**
      * {@inheritDoc}
      *
      * @return <code>{@link #getAddress()}()</code> if {@link NetworkInterface#isLoopback()} is true, null otherwise.
      */
     @Override
     protected InetAddress isAcceptable(NetworkInterface networkInterface, InetAddress address) throws SocketException {
```
pferraro, line 70: This should no longer throw a SocketException, as this is now handled.

rhusar: @pferraro Well spotted! Opened https://github.com/wildfly/wildfly-core/pull/6398 as a fixup.

rhusar: I missed this because it was NOT documented in the Javadoc...

PR #6398, opened 2025-04-16
LoopbackAddressInterfaceCriteria.isAcceptable now handles SocketExcep…

rhusar:
…tion, update method declaration.

Fixup https://github.com/wildfly/wildfly-core/pull/6397.


yersan:
As it is now, my feeling is that https://github.com/wildfly/wildfly-core/pull/6397 fix is not correct, since it is handling the SocketException when the interface Java doc explicitly states that any implementation should throw a SocketException if evaluating the state of the network interface results in such an error.

If what we want to to make it more robust, then I agree with you @rhusar , and in such a case, `AbstractInterfaceCriteria#isAcceptable` signature should be modified according by removing from there the possibility of throwing a SocketException, and consequently, modify all the `AbstractInterfaceCriteria` implementations in a similar way as was done in https://github.com/wildfly/wildfly-core/pull/6397: logging the error and returning `null`

rhusar:
Thanks for the comment @yersan, work in progress, I will close this in the meantime.

PR #6399, opened 2025-04-16
WFCORE-7222 Improve service start condition configuration of Installer.Builder

pferraro:
https://issues.redhat.com/browse/WFCORE-7222

Adds javadoc clarifying the default StartWhen semantics of each Installer.builder(...) use case.

### DIFF
--- a/service/src/main/java/org/wildfly/service/Installer.java
+++ b/service/src/main/java/org/wildfly/service/Installer.java
@@ -43,6 +43,43 @@ public interface Installer<ST extends ServiceTarget> {
     ServiceController<?> install(ST target);
 
     /**
+     * Enumerates the possible start conditions of a service.
+     */
+    enum StartWhen {
+        /**
+         * Indicates that a service and its dependencies should start when required by another service that wants to start.
+         * Once started, the service will automatically stop when there are no longer any services requiring it.
+         * This is the default start condition.
+         * @see {@link ServiceController.Mode#ON_DEMAND}
+         */
+        REQUIRED(ServiceController.Mode.ON_DEMAND),
+        /**
+         * Indicates that a service should start automatically after all of its dependencies have started.
+         * Once started, the service will automatically stop when any of its dependencies wants to stop.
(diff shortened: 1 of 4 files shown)


service/src/main/java/org/wildfly/service/Installer.java
```diff
@@ -73,18 +110,33 @@ public void accept(SB builder) {
...
+        B startWhen(StartWhen condition);
+
         /**
          * Configures the installed service to automatically start when all of its dependencies are available
          * and to automatically stop when any of its dependencies are no longer available.
          * @return a reference to this builder
+         * @deprecated Superseded by {@link #withStartMode(StartWhen)}
          */
-        B asPassive();
+        @Deprecated(forRemoval = true)
```
yersan, line 126: Could you add the since attribute?

service/src/main/java/org/wildfly/service/Installer.java
```diff
@@ -73,18 +110,33 @@ public void accept(SB builder) {
...
+            return this.startWhen(StartWhen.AVAILABLE);
+        }
 
         /**
          * Configures the installed service to start immediately after installation, forcing any dependencies to start.
          * @return a reference to this builder
+         * @deprecated Superseded by {@link #withStartMode(StartWhen)}
          */
-        B asActive();
+        @Deprecated(forRemoval = true)
```
yersan, line 136: Could you add the since attribute?

yersan (changes requested):

service/src/main/java/org/wildfly/service/Installer.java
pferraro: Fixed.

service/src/main/java/org/wildfly/service/Installer.java
pferraro: Fixed.

yersan (approved):

yersan:
Thanks @pferraro

PR #6405, opened 2025-04-23
[WFCORE-7230] Upgrade JBoss Parent to 49

yersan:
Jira issue: https://issues.redhat.com/browse/WFCORE-7230


yersan:
@jamezp do you have any concerns about this? ... I would prefer to get it updated now until we know better the following steps related to nexus3 using the jboss parent, so probably there will be another update soon.

jamezp (approved):
No concern from me @yersan

yersan:
Thanks @jamezp

PR #6410, opened 2025-04-25
WFCORE-7234 ParallelBootOperationContext.getResourceRegistration() can throw OperationFailedRuntimeException during operation rollback

pferraro:
https://issues.redhat.com/browse/WFCORE-7234

### DIFF
--- a/controller/src/main/java/org/jboss/as/controller/ParallelBootOperationContext.java
+++ b/controller/src/main/java/org/jboss/as/controller/ParallelBootOperationContext.java
@@ -144,14 +144,14 @@ class ParallelBootOperationContext extends AbstractOperationContext {
 
     @Override
     public ImmutableManagementResourceRegistration getResourceRegistration() {
-        ImmutableManagementResourceRegistration parent = primaryContext.getResourceRegistration();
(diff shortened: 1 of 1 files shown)


bstansberry:
@pferraro I don't understand what's going on here. The address of the activeStep of the primaryContext should be the empty address, so I'd like to better understand why it's not.

pferraro:
@bstansberry

> The address of the activeStep of the primaryContext should be the empty address, so I'd like to better understand why it's not.

I'm not exactly sure.  The only clue I can provide is that the exception posted in the jira was thrown during a rollback of boot operations (for unrelated reasons) seen during a local execution of the NongracefulStartTestCase (from the manualmode testsuite).

Is there anything technically _incorrect_ about the proposed change?

bstansberry (approved):

bstansberry:
@pferraro I filed https://issues.redhat.com/browse/WFCORE-7243 for any follow up work re my broader concern. I see why this happens; it's explained there. Thanks for raising this.

yersan:
Thanks @pferraro @bstansberry

PR #6411, opened 2025-04-29
[Community] Fix for WFCORE-7235, evolve platform MBeans with JDK11 com.sun.management content and JDK11 java.lang new content.

jfdenise:
Issue: https://issues.redhat.com/browse/WFCORE-7235

Proposal: https://github.com/wildfly/wildfly-proposals/pull/729


platform-mbean/src/main/resources/org/jboss/as/platform/mbean/LocalDescriptions.properties
```diff
@@ -191,6 +191,8 @@ threading.dump-all-threads.suspended=Whether the thread is suspended.
 threading.dump-all-threads.in-native=Whether the thread associated is executing native code via the Java Native Interface (JNI).
 threading.dump-all-threads.locked-monitors=A list of detyped representations java.lang.management.MonitorInfo objects, each of which represents an object monitor currently locked by the thread.
 threading.dump-all-threads.locked-synchronizers=A list of detyped representations java.lang.management.LockInfo objects, each of which represents an ownable synchronizer currently locked by the thread.
+threading.dump-all-threads.daemon=Tests if the thread associated with this ThreadInfo is a daemon thread.
```
bstansberry, line 227: "Whether the thread is a daemon thread."

^^^ is to make the style of this consistent with the other fields.

platform-mbean/src/main/resources/org/jboss/as/platform/mbean/LocalDescriptions.properties
```diff
@@ -191,6 +191,8 @@ threading.dump-all-threads.suspended=Whether the thread is suspended.
...
+threading.dump-all-threads.priority=The thread priority of the thread associated with this ThreadInfo.
```
bstansberry, line 228: "The thread priority of the thread."

^^^ is to make the style of this consistent with the other fields.

platform-mbean/src/main/resources/org/jboss/as/platform/mbean/LocalDescriptions.properties
```diff
@@ -220,6 +222,8 @@ threading.get-thread-info.suspended=Whether the thread is suspended.
 threading.get-thread-info.in-native=Whether the thread associated is executing native code via the Java Native Interface (JNI).
 threading.get-thread-info.locked-monitors=A list of detyped representations java.lang.management.MonitorInfo objects, each of which represents an object monitor currently locked by the thread.
 threading.get-thread-info.locked-synchronizers=A list of detyped representations java.lang.management.LockInfo objects, each of which represents an ownable synchronizer currently locked by the thread.
+threading.get-thread-info.daemon=Tests if the thread associated with this ThreadInfo is a daemon thread.
+threading.get-thread-info.priority=The thread priority of the thread associated with this ThreadInfo.
```
bstansberry, line 259: Same as above for these two.

platform-mbean/src/main/resources/org/jboss/as/platform/mbean/LocalDescriptions.properties
```diff
@@ -249,6 +253,8 @@ threading.get-thread-infos.suspended=Whether the thread is suspended.
 threading.get-thread-infos.in-native=Whether the thread associated is executing native code via the Java Native Interface (JNI).
 threading.get-thread-infos.locked-monitors=A list of detyped representations java.lang.management.MonitorInfo objects, each of which represents an object monitor currently locked by the thread.
 threading.get-thread-infos.locked-synchronizers=A list of detyped representations java.lang.management.LockInfo objects, each of which represents an ownable synchronizer currently locked by the thread.
+threading.get-thread-infos.daemon=Tests if the thread associated with this ThreadInfo is a daemon thread.
+threading.get-thread-infos.priority=The thread priority of the thread associated with this ThreadInfo.
```
bstansberry, line 290: Same as above for these two.

platform-mbean/src/main/java/org/jboss/as/platform/mbean/PlatformLoggingMXBeanAttributeHandler.java
```diff
@@ -0,0 +1,67 @@
...
+            throw new OperationFailedException(e.toString());
+        }
+
+    }
+
+    static void storeResult(final String name, final ModelNode store) {
+
+        if (PlatformMBeanConstants.OBJECT_NAME.getName().equals(name)) {
+            store.set(PlatformMBeanConstants.PLATFORM_LOGGING_MXBEAN_NAME);
+        } else if (PlatformMBeanConstants.LOGGER_NAMES.equals(name)) {
```
bstansberry, line 57: @jamezp is PlatformLoggingMXBean.getLoggerNames a complete list in the WF context, i.e. the log manager is from JBoss Logging?

bstansberry (commented):
I made some trivial comments.

I'm not sure how I feel about the overall direction of this. I understand now what you meant when talking about com.sun.management. You weren't discussing an implementation detail; you are using that package as a source of the API we offer.

I don't know what different SE variants offer from that package. Maybe it's consistent for anything based on OpenJDK. I don't think it's entirely consistent though as you are using UnixOperatingSystemMXBean. I presume there are things like WindowsOperatingSystemMXBean if you were using a native Windows JDK.

platform-mbean/src/main/java/org/jboss/as/platform/mbean/PlatformLoggingMXBeanAttributeHandler.java
jamezp: It should be, yes. It should return any logger created with:
* JUL
* JBoss Logging
* JBoss Log Manager
* SLF4J
* Log4j 2.x
* Apache Commons Logging

jamezp: Maybe I didn't answer completely. This should be a list of all loggers created, not just the ones defined in the logging subsystem which contain the configuration for the logger.

jfdenise: @jamezp , that is what I am observing. We have the list of all loggers. That would provide a source of information of what are the loggers that could be enabled.

jfdenise:
@bstansberry , only a Unix specific MXBean has been defined. No Windows specific one. 
The com.sun.management API is present, as you said in  the openjdk based JDK. I checked: semuru, Azul zulu, RH openjdk, temurin, openjdk, oracle JDK. 

If we want to be safe, I could use reflection to cast the MXBean actual instance in order to not break when running on a JDK where the classes would be not available.

@bstansberry do you have a feeling on introducing JDK11 MBean attributes that have been deprecated in JDK17?

platform-mbean/src/main/java/org/jboss/as/platform/mbean/PlatformLoggingMXBeanAttributeHandler.java
bstansberry: Great; thanks!

jfdenise:
@bstansberry , I have implemented with this last commit: https://github.com/wildfly/wildfly-core/pull/6411/commits/b89d8b1cfffde65c6be61d1d64e22d8a55faa335
a flexible approach based on the JMX lower level reflective API. I removed the dependency on com.sun.management classes and jdk.management module.

We check that the MBean metadata defines the attributes/operations prior to invoke them. The WildFly management model is static with those attributes being nillable (in case not defined on the platform). Invoking a non supported operation would lead to the expected error.

I feel that it offers a much better portability. WildFly will be able to run on any JDK, adapting to the underlying JMX model.

I also added some unit tests (in the spirit of the existing ones).

@bstansberry another question, should I open a proper WildFly proposal for it? This could be seen as new features.
Thank-you.

platform-mbean/src/main/java/org/jboss/as/platform/mbean/AbstractExtendedMBean.java
```diff
@@ -0,0 +1,116 @@
...
+
+    Object getAttribute(String attributeName) throws OperationFailedException {
+        try {
+            return ManagementFactory.getPlatformMBeanServer().
+                    getAttribute(name, attributeName);
+        } catch (AttributeNotFoundException
+                | InstanceNotFoundException
+                | MBeanException
+                | ReflectionException ex) {
+            throw new OperationFailedException(ex);
```
bstansberry, line 90: This comment applies to this commit overall; I just randomly picked this line.

We should only throw OFE or any other exception type that implements OperationClientException if the failure indicates an end user mistake. For any other cases, an OperationStepHandler should throw a runtime exception of some sort.

The server kernel exception handling relies on this, i.e. perhaps dealing with user mistakes one way while handling other failures differently.

bstansberry:
@jfdenise I made a comment re exception handling in your reflection commit; otherwise that seems fine.

Following are responses to various things you've said:

> I had to name get-threads-allocated-bytes the method that takes an array of ids. get-thread-allocated-bytes being used for the operation taking an id as parameter. 

OK. I'm to lazy to look but if you didn't already, perhaps note that mapping in the LocalDescriptions.properties text.

> The attribute `threading.current-thread-allocated-bytes` has been introduced in JDK 11.0.10. I added it commented, to be un-commented when we evolve to JDK17

FWIW, I think it's fine to enable now. The real target for this is WF, and it requires 17. So it's only build/test of Core where this is relevant. I'm not particularly diligent about updating my SE installs on one machine and even there it's on 11.0.22. 

If this is an extended one now handled by reflection, for sure add it now.

> Some new attributes I have added from `com.sun.management.OperatingSystemMXBean` are already deprecated in more recent JDK17. I added them, but I feel that we shouldn't.
 
Perhaps  add the replacements, which at a glance came in SE 14. Your reflection stuff will mean this will just work when run on SE 11, which isn't a real target environment anyway.

> * I am unsure on how to name Map field of the new GcInfo complex attribute. I named it like [that](https://github.com/jfdenise/wildfly-core/blob/platform-mbeans/platform-mbean/src/main/resources/org/jboss/as/platform/mbean/LocalDescriptions.properties#L49) but its name is unrelated to the resource/attribute:

I'm not sure, but if it gives you the expected content in read-resource-description I think it's fine. :-)

> should I open a proper WildFly proposal for it? This could be seen as new features.
Thank-you.

I think so, yes. If this had just been updating the existing stuff to add some newer content in the existing MBeans, or perhaps adding some other MBeans defined in java.lang.management treating it as an Enhancement would have seemed appropriate to me. But I think the expansion of the API to include stuff from com.sun.management gets us into Feature territory.

I'll be your SME on the feature team.

When you announce the feature in the wildfly-dev post, please highlight how the API will now include what those beans provide. That's the aspect where people may have opinions.

jfdenise:
@bstansberry , thank-you for your feedback, it really helped. 
I fixed exceptions, added the attribute defined in JDK11.0.10 and used JDK14 replacements instead of the deprecated ones.
I am just not clear on your reply for get-thread-allocated-bytes(long[]). If I follow the WildFly naming pattern used for others it should be: get-thread-allocated-bytes**s**. I feel it strange to have double s, it looks like a typo. I preferred to name it get-thread**s**-allocated-bytes.

platform-mbean/src/main/resources/org/jboss/as/platform/mbean/LocalDescriptions.properties
```diff
@@ -140,6 +168,21 @@ threading.thread-cpu-time-enabled=Whether thread CPU time measurement is enabled
 threading.object-monitor-usage-supported=Whether the Java virtual machine supports monitoring of object monitor usage.
 threading.synchronizer-usage-supported=Whether the Java virtual machine supports monitoring of ownable synchronizer usage.
 
+threading.current-thread-allocated-bytes=An approximation of the total amount of memory, in bytes, allocated in heap memory for the current thread.
+threading.get-thread-allocated-bytes=An approximation of the total amount of memory, in bytes, allocated in heap memory for the thread with the specified ID.
+threading.get-thread-allocated-bytes.id=A thread id.
+threading.get-thread-allocated-bytes.reply=An approximation of the total memory allocated, in bytes, in heap memory for the thread with the specified ID if the thread with the specified ID exists, the thread is alive, and thread memory allocation measurement is enabled; -1 otherwise.
+threading.get-threads-allocated-bytes=An approximation of the total amount of memory, in bytes, allocated in heap memory for each thread whose ID is in the input array ids.
```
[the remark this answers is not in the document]
jfdenise: Thank-you @bstansberry , fixed.

ehsavoie (approved):

jfdenise:
@yersan , do you think that this could be there for WF 37? Thank-you.

platform-mbean/src/main/java/org/jboss/as/platform/mbean/GarbageCollectorResourceDefinition.java
```diff
@@ -43,6 +55,58 @@ class GarbageCollectorResourceDefinition extends SimpleResourceDefinition {
...
+            .setStorageRuntime()
+            .setRuntimeServiceNotRequired()
+            .setMeasurementUnit(MeasurementUnit.MILLISECONDS)
+            .setStability(Stability.COMMUNITY)
+            .build();
+    private static final AttributeDefinition MEMORY_USAGE_AFTER_GC_ATTRIBUTE = ObjectMapAttributeDefinition.Builder.of(
+            PlatformMBeanConstants.MEMORY_USAGE_AFTER_GC,
+            ObjectTypeAttributeDefinition.Builder.of(PlatformMBeanConstants.MEMORY_USAGE_AFTER_GC, MEMORY_INIT, MEMORY_USED, MEMORY_COMMITTED, MEMORY_MAX)
+                    .build())
+            .setFlags(AttributeAccess.Flag.STORAGE_RUNTIME)
```
yersan: Just out of curiosity, why not `.setStorageRuntime()` as the others?

platform-mbean/src/main/java/org/jboss/as/platform/mbean/PlatformLoggingResourceDefinition.java
```diff
@@ -0,0 +1,62 @@
...
+import org.jboss.as.controller.AttributeDefinition;
+import org.jboss.as.controller.PrimitiveListAttributeDefinition;
+import org.jboss.as.controller.SimpleResourceDefinition;
+import org.jboss.as.controller.operations.global.ReadResourceHandler;
+import org.jboss.as.controller.registry.ManagementResourceRegistration;
+import static org.jboss.as.platform.mbean.PlatformMBeanConstants.PLATFORM_LOGGING_PATH;
+import org.jboss.as.version.Stability;
+import org.jboss.dmr.ModelType;
+
+class PlatformLoggingResourceDefinition extends SimpleResourceDefinition {
```
yersan, line 21: This resource needs to override the `getStability()` and provide the community stability level, otherwise it will be available in higher stabilities.

yersan (changes requested):

platform-mbean/src/main/java/org/jboss/as/platform/mbean/PlatformLoggingResourceDefinition.java
jfdenise: Thank-you @yersan , good catch. Fixed.

platform-mbean/src/main/java/org/jboss/as/platform/mbean/GarbageCollectorResourceDefinition.java
jfdenise: @yersan , a leftover. I fixed it.

yersan:
Thanks @jfdenise @ehsavoie @bstansberry

PR #6419, opened 2025-05-08
[WFCORE-7249] Startup error traces related to serialization errors ar…

yersan:
…e not captured in the log files

Jira issue: https://issues.redhat.com/browse/WFCORE-7249

### DIFF
--- a/host-controller/src/main/java/org/jboss/as/host/controller/Main.java
+++ b/host-controller/src/main/java/org/jboss/as/host/controller/Main.java
@@ -71,11 +71,8 @@ public final class Main {
     public static void main(String[] args) throws IOException {
         MDC.put("process", "host controller");
 
-
-        // Grab copies of our streams.
+        // Get a ref to the original stdin before we install JBoss Stdio
         final InputStream in = System.in;
-        //final PrintStream out = System.out;
-        //final PrintStream err = System.err;
 
         byte[] pcAuthKey = new byte[ProcessController.AUTH_BYTES_ENCODED_LENGTH];
         try {
--- a/server/src/main/java/org/jboss/as/server/DomainServerMain.java
+++ b/server/src/main/java/org/jboss/as/server/DomainServerMain.java
@@ -75,8 +75,8 @@ public final class DomainServerMain {
      */
     public static void main(String[] args) {
 
+        // Get a ref to the original stdin before we install JBoss Stdio
         final InputStream initialInput = new Base64InputStream(System.in);
-        final PrintStream initialError = System.err;
 
         // Make sure our original stdio is properly captured.
         try {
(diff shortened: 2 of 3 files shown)


host-controller/src/main/java/org/jboss/as/host/controller/Main.java
```diff
@@ -74,8 +74,6 @@ public static void main(String[] args) throws IOException {
 
         // Grab copies of our streams.
         final InputStream in = System.in;
-        //final PrintStream out = System.out;
-        //final PrintStream err = System.err;
```
jamezp, line 78: Should we add a comment here as to why we're no longer capturing these or potentially just delete them?

yersan:
@jamezp Do you have any concerns using structure loggers at this stage? My only doubt is that they will change a little bit how the error traces are captured in the log files; if something external to WildFly is inspecting them and assumes a specific pattern, it could be a nuisance. I guess the probability is too low, after all, these errors mean the server didn't start. Well, I guess I would like to gather some feedback, otherwise, please approve from your side.

host-controller/src/main/java/org/jboss/as/host/controller/Main.java
yersan: :) We crossed the messages, those were only the removal of two lines already commented that don't provide anything about what they are commented

bstansberry (approved):
LGTM. Thanks for the really clear JIRA description; it makes it obvious that this is the right fix.

host-controller/src/main/java/org/jboss/as/host/controller/Main.java
bstansberry: Looking at https://github.com/wildfly/wildfly-core/commit/008be9c6#diff-c2b144af158144bf34cc44a678172b1370f626d5a23e0d0e50f4cc124b516fed, I'm pretty sure those vars had been ultimately unused. AFAICT that commit didn't remove any ultimate use; it just stopped passing the values through method calls. 

I probably wanted to leave the var declarations commented as a kind of tombstone because the code would no longer "Grab copIES of our streamS." Bad choice on my part; it ultimately was just confusing.

That said, @yersan perhaps change that "// Grab copies of our streams" comment to "// Get a ref to the original stdin before we install JBoss Stdio"

jamezp: lol I definitely read the diff wrong there :)

jamezp:
> @jamezp Do you have any concerns using structure loggers at this stage? My only doubt is that they will change a little bit how the error traces are captured in the log files; if something external to WildFly is inspecting them and assumes a specific pattern, it could be a nuisance. I guess the probability is too low, after all, these errors mean the server didn't start. Well, I guess I would like to gather some feedback, otherwise, please approve from your side.

I think that's okay and I don't think it's too early to start logging. In the domain scripts we pass the logging configuration so logging should be configured based on the `logging.properties`.

jamezp (approved):

PR #6420, opened 2025-05-09
Wfcore 5744 29.x

baranowb:
Issue: https://issues.redhat.com/browse/WFCORE-5744
Proposal: https://issues.redhat.com/browse/EAP7-1863

### DIFF
--- a/elytron/src/main/java/org/wildfly/extension/elytron/CertificateChainAttributeDefinitions.java
+++ b/elytron/src/main/java/org/wildfly/extension/elytron/CertificateChainAttributeDefinitions.java
@@ -19,6 +19,7 @@ import org.jboss.as.controller.ObjectListAttributeDefinition;
 import org.jboss.as.controller.ObjectTypeAttributeDefinition;
 import org.jboss.as.controller.SimpleAttributeDefinition;
 import org.jboss.as.controller.SimpleAttributeDefinitionBuilder;
+import org.jboss.as.version.Stability;
 import org.jboss.dmr.ModelNode;
 import org.jboss.dmr.ModelType;
 import org.wildfly.common.iteration.ByteIterator;
@@ -88,9 +89,15 @@ class CertificateChainAttributeDefinitions {
 
     // TODO - Consider adding some of the more detailed fields from X509Certificate
 
+    // derivative attributes
+    private static final SimpleAttributeDefinition VALIDITY = new SimpleAttributeDefinitionBuilder(ElytronDescriptionConstants.VALIDITY, ModelType.STRING)
+            .setRequired(false)
+            .setStability(Stability.COMMUNITY)
+            .build();
+
     static final ObjectTypeAttributeDefinition CERTIFICATE = new ObjectTypeAttributeDefinition.Builder(ElytronDescriptionConstants.CERTIFICATE,
             TYPE, ALGORITHM, FORMAT, PUBLIC_KEY, SHA_1_DIGEST, SHA_256_DIGEST, ENCODED,
-            SUBJECT, ISSUER, NOT_BEFORE, NOT_AFTER, SERIAL_NUMBER, SIGNATURE_ALGORITHM, SIGNATURE, VERSION)
+            SUBJECT, ISSUER, NOT_BEFORE, NOT_AFTER, SERIAL_NUMBER, SIGNATURE_ALGORITHM, SIGNATURE, VERSION, VALIDITY)
         .setStorageRuntime()
         .build();
 
@@ -101,11 +108,11 @@ class CertificateChainAttributeDefinitions {
                 .build();
     }
 
-    static void writeCertificate(final ModelNode certificateModel, final Certificate certificate) throws CertificateEncodingException, NoSuchAlgorithmException {
-        writeCertificate(certificateModel, certificate, true);
+    static void writeCertificate(final ModelNode certificateModel, final Certificate certificate, final Stability stability) throws CertificateEncodingException, NoSuchAlgorithmException {
+        writeCertificate(certificateModel, certificate, true, stability);
     }
 
-    static void writeCertificate(final ModelNode certificateModel, final Certificate certificate, final boolean verbose) throws CertificateEncodingException, NoSuchAlgorithmException {
+    static void writeCertificate(final ModelNode certificateModel, final Certificate certificate, final boolean verbose, final Stability stability) throws CertificateEncodingException, NoSuchAlgorithmException {
        certificateModel.get(ElytronDescriptionConstants.TYPE).set(certificate.getType());
 
         PublicKey publicKey = certificate.getPublicKey();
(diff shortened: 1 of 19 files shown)


elytron/src/main/resources/schema/wildfly-elytron_community_18_1.xsd
```diff
@@ -0,0 +1,6450 @@
+<?xml version="1.0" encoding="UTF-8"?>
+
+<!--
+  ~ Copyright The WildFly Authors
+  ~ SPDX-License-Identifier: Apache-2.0
+  -->
+
+<xs:schema xmlns:xs="http://www.w3.org/2001/XMLSchema"
+           targetNamespace="urn:wildfly:elytron:community:18.1"
```
darranl, line 9: Need to think about what is the correct updates needed here but we avoid minor bumps.

elytron/src/main/java/org/wildfly/extension/elytron/CertificateChainAttributeDefinitions.java
```diff
@@ -88,9 +89,15 @@ class CertificateChainAttributeDefinitions {
 
     // TODO - Consider adding some of the more detailed fields from X509Certificate
 
+    // derivative attributes
+    private static final SimpleAttributeDefinition VALIDITY = new SimpleAttributeDefinitionBuilder(ElytronDescriptionConstants.VALIDITY, ModelType.STRING)
```
darranl, line 93: I think we can also specify the units in the definition instead of in the text, I would propose we use units that are more user friendly.

I don't think it is reasonable to check more frequently than once per second, even that sounds too frequent and maybe the granularity should be in minutes.

elytron/src/main/java/org/wildfly/extension/elytron/ElytronExtension.java
```diff
@@ -75,6 +75,7 @@ public class ElytronExtension implements Extension {
     static final ModelVersion ELYTRON_16_0_0 = ModelVersion.create(16);
     static final ModelVersion ELYTRON_17_0_0 = ModelVersion.create(17);
     static final ModelVersion ELYTRON_18_0_0 = ModelVersion.create(18);
+    static final ModelVersion ELYTRON_18_1_0 = ModelVersion.create(18, 1);
```
darranl, line 78: The model itself should just bump to 20, the subsystem only registers one version at a time so would never register as 18.1

elytron/src/main/java/org/wildfly/extension/elytron/KeyStoreService.java
```diff
@@ -68,13 +74,20 @@ class KeyStoreService implements ModifiableKeyStoreService {
...
     private final String provider;
     private final String type;
     private final String path;
     private final String relativeTo;
     private final boolean required;
     private final String aliasFilter;
+    //delay in ms between service start and periodical check of certificate health
+    //if set to '0', this will mean one-time off warning, as prior to RFE
+    private long expirationCheckDelay = DEFAULT_DELAY;
+    private ScheduledExecutorService scheduledExecutorService;
```
darranl, line 90: FYI as part of the CVE I am working on I am adding the scheduled executor service to the root of the Elytron subsystem so all Elytron services can share the same executor.

elytron/src/main/java/org/wildfly/extension/elytron/KeyStoreService.java
```diff
@@ -206,35 +227,89 @@ private AtomicLoadKeyStore.LoadKey load(AtomicLoadKeyStore keyStore) throws Exce
...
+            cancelHealthCheck(true);
+            keyStore = null;
+            unmodifiableKeyStore = null;
+            trackingKeyStore = null;
+            if (pathResolver != null) {
+                pathResolver.clear();
+                pathResolver = null;
+            }
+        } finally {
+            this.scheduledExecutorService.shutdownNow();
```
darranl, line 312: Looking at how this is handled for the executors elsewhere a more complete shutdown solution is needed but the shared executor I am adding will have that,

elytron/src/main/java/org/wildfly/extension/elytron/KeyStoreService.java
```diff
@@ -206,35 +227,89 @@ private AtomicLoadKeyStore.LoadKey load(AtomicLoadKeyStore keyStore) throws Exce
...
                 if (certificate != null && certificate instanceof X509Certificate) {
-                    try {
-                        ((X509Certificate) certificate).checkValidity();
-                    } catch (CertificateExpiredException | CertificateNotYetValidException e) {
-                        ROOT_LOGGER.certificateNotValid(alias, e);
+                    final X509Certificate xCertificate = (X509Certificate) certificate;
+                    final CertificateValidity certificateValidity = CertificateValidity.getValidity(xCertificate.getNotBefore(), xCertificate.getNotAfter());
+                    switch(certificateValidity) {
+                        case ABOUT_TO_EXPIRE:
+                            ROOT_LOGGER.certificateAboutToExpire(keyStoreName, alias);
```
darranl, line 241: We probably need a mechanism to avoid spamming the logs for the same certificate, especially if they want a frequent check.

elytron/src/main/java/org/wildfly/extension/elytron/CertificateValidity.java
```diff
@@ -0,0 +1,60 @@
...
+        final long WEEK_IN_MS = 7 * 24 * 60 * 60 * 1000;
+        final long DAY_IN_MS = 4 * 60 * 60 * 1000;
+        final Date currentDate = Calendar.getInstance().getTime();
+        if (currentDate.before(notBefore)) {
+            return NOT_YET;
+        } else if (currentDate.after(notBefore) && currentDate.before(notAfter)) {
+            // valid or about to expire
+            // once x509 is up, switch to java.time.temporal.ChronoUnit.DAYS.between ?
+            long diff = (notAfter.getTime() - notBefore.getTime());
+            if (diff < WEEK_IN_MS) {
```
darranl, line 40: The warning interval feels like it is something that should be configuration, I can imagine many operational and financial processes that may influence when an administrator needs to be notified.

darranl (changes requested):
Thank you for the pull request, I have added some comments specific to this pull request but I think we also need to step back and get the WildFly feature development process started so we can begin to discuss the general proposed solution to the problem.

elytron/src/main/java/org/wildfly/extension/elytron/CertificateValidity.java
baranowb: Yeah, that is something that I did think about, but chose to KISS it. Shouldnt be a problem.

elytron/src/main/java/org/wildfly/extension/elytron/CertificateChainAttributeDefinitions.java
baranowb: Oh, honestly I just forgot thats possible...

baranowb: Ach, actually its by design, its string representation not a numeric value(TTL).

darranl: Thank you, I think my comment is on the wrong attribute - I see this one is the info from the certifcate.

elytron/src/main/java/org/wildfly/extension/elytron/CertificateValidity.java
darranl: Thank you, in that case I think this needs further discussion in the analysis.

elytron/src/main/java/org/wildfly/extension/elytron/KeyStoreService.java
baranowb: hmm, whats the point of check than ?

yersan:
Hello @baranowb , you added the `hold` and `29` labels for this community feature, but feature freeze is this Wednesday, so it looks like this one won't be ready for `29`. Otherwise, if you are actively working on this and plan to get it by this Wednesday, you should remove the `hold` label.

baranowb:
@yersan I think it was because at that time it wasnt in good shape.
Im reviewing and testing locally RN.

PR #6426, opened 2025-05-20
[WFCORE-5898] Intermittent failures in DeploymentScannerUnitTestCase

yersan:
Jira issue: https://issues.redhat.com/browse/WFCORE-5898


ehsavoie:
Shouldn't we use TimeoutUtil.adjust to set those values ?

yersan:
> Shouldn't we use TimeoutUtil.adjust to set those values ?

Yes, I guess not too relevant being 2 seconds, but it makes sense since we have to use it for better accommodation on slow CI systems

Thanks @ehsavoie , done, could you review again when you have a chance?

ehsavoie (approved):

yersan:
Thanks @ehsavoie

PR #6438, opened 2025-06-09
[WFCORE-7273] Remove JBoss Threads classes from the ModelControllerClient API; use SE's CompletableFuture

bstansberry:
https://issues.redhat.com/browse/WFCORE-7273

Also https://issues.redhat.com/browse/WFCORE-7275 and https://issues.redhat.com/browse/WFCORE-7278

This is definitely one to review commit-by-commit. The 3rd one is the key one. The first one is there in case we want to port it downstream; the deprecation gets reverted by the 3rd commit.

### DIFF
--- a/protocol/src/main/java/org/jboss/as/protocol/mgmt/AsyncToCompletableFutureAdapter.java
+++ b/protocol/src/main/java/org/jboss/as/protocol/mgmt/AsyncToCompletableFutureAdapter.java
@@ -11,6 +11,7 @@ import java.util.concurrent.atomic.AtomicBoolean;
 import java.util.function.Consumer;
 import java.util.function.Function;
 
+import org.jboss.as.protocol.logging.ProtocolLogger;
 import org.jboss.threads.AsyncFuture;
 
 /**
@@ -116,20 +117,23 @@ public final class AsyncToCompletableFutureAdapter {
                 asyncCancelConsumer.accept(mayInterruptIfRunning);
                 boolean afCancelled = asyncFuture.awaitUninterruptibly() == AsyncFuture.Status.CANCELLED;
 
-                if (!afCancelled) {
+                if (afCancelled) {
+                    ProtocolLogger.ROOT_LOGGER.tracef("%s: AsyncFuture is cancelled; proceeding to cancel ourself",
+                            getClass().getEnclosingClass().getSimpleName(), getClass().getSimpleName());
+                } else {
                     // Our AsyncFuture.Listener completes this future asynchronously,
                     // so block until it has done so before providing the resulting
                     // state by calling super.cancel.
                     //
                     // We do this because canceling the async future may result in
-                    // the canceled management op returning with either outcome=cancelled
-                    // or outcome=failed. We're in this block, so we know it's the latter.
+                    // the canceled management op returning with outcome=success,
+                    // outcome=failed or outcome=cancelled.
                     // We know this will result in the listener, in another
-                    // thread, eventually trying to complete this exceptionally.
+                    // thread, eventually trying to complete or cancel this CompletableFuture.
                     // We don't want the determination of whether this method returns
                     // true or false to be decided by a race between our call below to
-                    // super.cancel and the listener thread completing the
-                    // future as failed. So we block to let the listener thread win the race.
+                    // super.cancel and the listener thread completing or cancelling the
+                    // future. So we block to let the listener thread win the race.
                     //
                     // An alternative approach would be to call super.cancel first before
                     // cancelling the asyncFuture. But that could result in inconsistency
@@ -137,13 +141,15 @@ public final class AsyncToCompletableFutureAdapter {
                     // executed op. When ModelControllerClient async methods formerly returned an
                     // AsyncFuture these results were consistent, so we're maintaining
                     // that behavior with the CompletableFuture we now return.
+                    ProtocolLogger.ROOT_LOGGER.tracef("%s.%s: Awaiting AsyncFuture.Listener before proceeding to cancel ourself",
+                            getClass().getEnclosingClass().getSimpleName(), getClass().getSimpleName());
                     awaitLatch();
                 }
                 return super.cancel(mayInterruptIfRunning);
             } finally {
                 // Code hardening:
                 // If for some reason asyncFuture.awaitUninterruptibly() returns CANCELLED
-                // without the listener getting invoked, now that we're cancelled make sure
+                // without the listener getting invoked (yet), now that we're cancelled make sure
                 // the latch is counted down so other calls to cancel don't block.
                 latch.countDown();
             }
--- a/controller-client/src/main/java/org/jboss/as/controller/client/impl/AbstractModelControllerClient.java
+++ b/controller-client/src/main/java/org/jboss/as/controller/client/impl/AbstractModelControllerClient.java
@@ -24,6 +24,7 @@ import org.jboss.as.controller.client.Operation;
 import org.jboss.as.controller.client.OperationBuilder;
 import org.jboss.as.controller.client.OperationMessageHandler;
 import org.jboss.as.controller.client.OperationResponse;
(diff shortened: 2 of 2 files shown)


protocol/src/main/java/org/jboss/as/protocol/mgmt/ActiveOperation.java
```diff
@@ -48,12 +51,48 @@ public interface ActiveOperation<T, A> {
...
+     * @param transformer function to transformer the operation result. Cannot be {@code null}
+     * @return the future result. Will not return {@code null}
+     */
+    <U> CompletableFuture<U> getCompletableFuture(Function<T, U> transformer);
+
+    /**
+     * Add a cancellation handler.
+     * @param runnable task to execute if the operation is cancelled
+     */
+    default void addCancellable(Runnable runnable) {
```
bstansberry: Note: I need to remove this. It's cruft from an earlier iteration.

protocol/src/main/java/org/jboss/as/protocol/mgmt/ActiveOperationImpl.java
```diff
@@ -121,8 +127,21 @@ public AsyncFuture<T> getResult() {
         return this;
     }
 
+    public <U>CompletableFuture<U> getCompletableFuture(Function<T, U> transformer) {
+        return AsyncToCompletableFutureAdapter.adapt(this, transformer);
+    }
+
     @Override
     public void asyncCancel(boolean interruptionDesired) {
```
bstansberry, line 133: TODO I need to add a javadoc warning about not calling this except from superclass AsyncFutureTask.cancel(boolean). It's meant for handling end user cancellation of the future via Future.cancel.

I should probably have this throw UnsupportedOperationException to prevent its use and override AsyncFuture,cancel(boolean).

jamezp (approved):

jamezp:
I tested this with combinations of 29.0.0.Beta4-SNAPSHOT of core and 37.0.0.Beta1-SNAPSHOT with 29.0.0.Beta4-SNAPSHOT versions of core as well as just the core SNAPSHOT with older versions of WildFly on the following projects:
- https://github.com/wildfly/wildfly-maven-plugin
- https://github.com/wildfly/wildfly-plugin-tools
- https://github.com/wildfly/wildfly-arquillian

All combinations passed.

protocol/src/main/java/org/jboss/as/protocol/mgmt/ActiveOperationImpl.java
bstansberry: This last comment is now out of date as I significantly changed how this works.

protocol/src/main/java/org/jboss/as/protocol/mgmt/AsyncToCompletableFutureAdapter.java
```diff
@@ -0,0 +1,164 @@
...
+        }
+
+        @Override
+        public boolean cancel(boolean mayInterruptIfRunning) {
+
+            // Only try and cancel our asyncFuture once
+            if (!cancellable.compareAndSet(false, true)) {
+                // If another thread is trying to cancel the asyncFuture, await their work
+                awaitLatch();
+                return super.cancel(mayInterruptIfRunning);
```
yersan, line 107: We could arrive at this point even with `asyncCancelConsumer` not being invoked at all, isn't it? if so, we could be returning the result of the `CompletableFuture` cancellation even without invoking the  `asyncCancelConsumer`. Wouldn't this be possible?

For example, suppose the AsyncFuture listener has completed successfully and the thread which is already cancelling the CompletableFuture has not reached L116 yet.

protocol/src/main/java/org/jboss/as/protocol/mgmt/AsyncToCompletableFutureAdapter.java
```diff
@@ -0,0 +1,164 @@
...
+                boolean afCancelled = asyncFuture.awaitUninterruptibly() == AsyncFuture.Status.CANCELLED;
+
+                if (!afCancelled) {
+                    // Our AsyncFuture.Listener completes this future asynchronously,
+                    // so block until it has done so before providing the resulting
+                    // state by calling super.cancel.
+                    //
+                    // We do this because canceling the async future may result in
+                    // the canceled management op returning with either outcome=cancelled
+                    // or outcome=failed. We're in this block, so we know it's the latter.
```
yersan: > We're in this block, so we know it's the latter

I am missing something here, looking at the possible results returned by `asyncFuture.awaitUninterruptibly()`, one of them is `AsyncFuture.Status.COMPLETE`

Can we not arrive here with the AsyncFuture completed successfully?

yersan: Well, I guess it would be perfectly fine to try to cancel something that could have already been completed ... so, just out of curiosity

protocol/src/main/java/org/jboss/as/protocol/mgmt/AsyncToCompletableFutureAdapter.java
yersan: Well, I guess it will also be fine, since in the end, soon or later, will be invoked by the other thread which is already cancelling, so the question would be more if this semantic is the expected one, I mean, if we can return the result of the cancellation even without have started the `asyncCancelConsumer` task.

protocol/src/main/java/org/jboss/as/protocol/mgmt/AsyncToCompletableFutureAdapter.java
bstansberry: Right; that comment should be improved.

protocol/src/main/java/org/jboss/as/protocol/mgmt/AsyncToCompletableFutureAdapter.java
bstansberry: @yersan Thanks for asking! Your question points to a flaw in the existing code, which this PR actually improves a bit. But it can be better.

The flaw in the existing code is AbstractModelControllerClient returns AbstractModelControllerClient.DelegatingCancellableAsyncFuture, which has asyncCancel logic that will send the CancelSyncRequest regardless of the state of the delegate future. And its inherited cancel(boolean) impl doesn't check either (which is fine). This differs from the wrapped future's asyncCancel impl (in ActiveOperationImpl), which can safely be invoked multiple times.

This PR improves things a bit by its use of the 'cancellable' AtomicBoolean -- only one thread can call asyncCancelConsumer. And if asyncCancelConsumer is the variant that calls ActiveOperationImpl.asyncCancel, all is well. But if it's the one from AbstractModelControllerClient that executes CancelAsyncRequest, we still have the potential to send that request when the ActiveOperationImpl is already completed.

I'm confident (but haven't confirmed by close analysis) that trying to send the CancelAsyncRequest when the ActiveOperationImpl is already done will result in an exception. I think we can improve this in AbstractModelControllerClient.executeAsyncCancelRequest by:

1) Check that activeOperation.getStatus() is Status.WAITING before trying to execute CancelAsyncRequest.
2) There's still a race, so in the existing catch block check activeOperation.getStatus() again and only throw RuntimeException if it's Status.WAITING.

bstansberry: @yersan I filed https://issues.redhat.com/browse/WFCORE-7278 for this and added a commit to improve that.

protocol/src/main/java/org/jboss/as/protocol/mgmt/AsyncToCompletableFutureAdapter.java
bstansberry: I updated the comment and added some trace logging.

PR #6439, opened 2025-06-10
[WFCORE-7257] Windows standalone.bat: Missing String delimiters -> JAVA_OPTS interpreted as commands

lvydra:
https://issues.redhat.com/browse/WFCORE-7257

### DIFF
--- a/core-feature-pack/common/src/main/resources/content/bin/common.bat
+++ b/core-feature-pack/common/src/main/resources/content/bin/common.bat
@@ -40,46 +40,41 @@ goto :eof
 :setDefaultModularJvmOptions
   call :setModularJdk
   if "!MODULAR_JDK!" == "true" (
-    echo "%~1" | findstr /I "\-\-add\-modules" > nul
-    if errorlevel == 1 (
-      rem Set default modular jdk options
-      rem Needed by the iiop-openjdk subsystem
-      set "DEFAULT_MODULAR_JVM_OPTIONS=!DEFAULT_MODULAR_JVM_OPTIONS! --add-exports=java.desktop/sun.awt=ALL-UNNAMED"
-      rem Needed to instantiate the default InitialContextFactory implementation used by the
-      rem Elytron subsystem dir-context and core management ldap-connection resources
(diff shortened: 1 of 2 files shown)


core-feature-pack/common/src/main/resources/content/bin/common.bat
```diff
@@ -40,46 +40,41 @@ goto :eof
 :setDefaultModularJvmOptions
   call :setModularJdk
   if "!MODULAR_JDK!" == "true" (
-    echo "%~1" | findstr /I "\-\-add\-modules" > nul
```
lvydra, line 43: This seems to no longer serve its purpose.
And it's probably not even working as supposed, currently (for some time) it's checking only against the first argument from JAVA_OPTS, and so it's always passing.

yersan: Thanks @lvydra , I'm not sure about the changes, It looks like the intention here was not to set the default modular JVM if we are running on a modular JDK, but there is no `--add-modules` setting on JAVA_OPTS.

With your change, we will no longer able to check whether  `--add-modules` is in JAVA_OPTS.

In addition, if the argument is passed with double quotes, this label will detect a single argument. So maybe to fix the problem where "%~1" is only taking the first JAVA_OPTS value, the fix should be in core-feature-pack/common/src/main/resources/content/bin/standalone.bat#L292 and invoking it with:

`call "!DIRNAME!common.bat" :setDefaultModularJvmOptions "!JAVA_OPTS!"`

Other invocations to this label should be revised and verified that all uses double quotes for the argument passed in. WDYT?

yersan (changes requested):

core-feature-pack/common/src/main/resources/content/bin/common.bat
lvydra: Thanks @yersan 

> With your change, we will no longer able to check whether --add-modules is in JAVA_OPTS.

I have been too hasty with this one.

It seems that adding only quotation marks is not sufficient, as JAVA_OPTS is even then broken into multiple arguments, and then the check is not always relevant.
`echo %* | findstr /I "\-\-add\-modules" > nul`
should fix it.

PR #6440, opened 2025-06-11
[WFCORE-7045][WFCORE-7238] Upgrade JBoss Threads to 3.9.1

bstansberry:
https://issues.redhat.com/browse/WFCORE-7045
https://issues.redhat.com/browse/WFCORE-7238

Builds on #6438


bstansberry:
The test results indicate I must have missed something in https://github.com/wildfly/wildfly/pull/18909

bstansberry:
https://github.com/wildfly/wildfly/pull/19006 should file the CI compile problems.

threads/src/main/java/org/jboss/as/threads/ManagedExecutorServiceImpl.java
```diff
@@ -0,0 +1,121 @@
+package org.jboss.as.threads;
```
yersan, line 5: License header is missing

threads/src/main/java/org/jboss/as/threads/ManagedJBossThreadPoolExecutorService.java
```diff
@@ -4,101 +4,12 @@
...
-    }
-
-    public int getQueueSize() {
-        return executor.getQueueSize();
-    }
-
-    <A> void addShutdownListener(final EventListener<A> shutdownListener, final A attachment) {
-        executor.addShutdownListener(shutdownListener, attachment);
-    }
+@Deprecated(forRemoval = true)
```
yersan: Ideally, add a `since` attribute to each `@deprecated` to easy know when it was deprecated, not mandatory, but this is what we have been requested with all deprecation annotations

bstansberry: Thanks, @yersan -- I've added that on this class and ManagedQueuelessExecutorService.

threads/src/main/java/org/jboss/as/threads/ManagedExecutorServiceImpl.java
bstansberry: Fixed

PR #6441, opened 2025-06-12
[WFCORE-7276] InterfaceAddHandler should consider updates made by lat…

bstansberry:
…er ops in the composite when creating its ParsedInterfaceCriteria

https://issues.redhat.com/browse/WFCORE-7276


ehsavoie:
@bstansberry what is strange is that there is already a test for testing this exact YAML

ehsavoie:
The test is good, to be honest I didn't see it when I fetched and thought it was an initial one ;)

yersan (approved):

yersan:
@ehsavoie could you approve from your side? thanks!

ehsavoie (approved):

PR #6445, opened 2025-06-21
[WFCORE-7281] Work with JDK versions with differing CompositeData structures for GarbageCollectorMXBean lastGcInfo

bstansberry:
https://issues.redhat.com/browse/WFCORE-7281

### DIFF
--- a/testsuite/standalone/src/test/java/org/wildfly/core/test/standalone/mgmt/ReadResourceTestCase.java
+++ b/testsuite/standalone/src/test/java/org/wildfly/core/test/standalone/mgmt/ReadResourceTestCase.java
@@ -1,0 +1,34 @@
+package org.wildfly.core.test.standalone.mgmt;
+
+import java.io.IOException;
+
+import jakarta.inject.Inject;
+import org.jboss.as.controller.descriptions.ModelDescriptionConstants;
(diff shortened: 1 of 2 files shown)


testsuite/standalone/src/test/java/org/wildfly/core/test/standalone/mgmt/ReadResourceTestCase.java
```diff
@@ -0,0 +1,34 @@
+package org.wildfly.core.test.standalone.mgmt;
```
yersan, line 5: It misses a license header, I'll add it after merging it. Maybe we can have a GitHub action to check this

yersan (approved):

yersan:
RemoteHttpGitRepositoryTestCase failures on Windows JDK 17 are flaky tests, tracked on WFCORE-7282, WFCORE-7283

testsuite/standalone/src/test/java/org/wildfly/core/test/standalone/mgmt/ReadResourceTestCase.java
bstansberry: Thanks @yersan. I need to update the IDE on that machine.

PR #6450, opened 2025-07-02
Fix for WFCORE-7290 and WFCORE-7291

jfdenise:
Issues:
https://issues.redhat.com/browse/WFCORE-7290
https://issues.redhat.com/browse/WFCORE-7291

* Add a new test that only runs when we are running tests with the minimal JDK version supported that is the version in sync with the platform MBeans integration.
* Fix platform MBeans that are not registered in the correct parent + Fix objectName value.


platform-mbean/src/main/java/org/jboss/as/platform/mbean/GarbageCollectorMXBeanAttributeHandler.java
```diff
@@ -53,7 +53,7 @@ protected void executeReadAttribute(OperationContext context, ModelNode operatio
             storeExtendedResult(gcName, name, context.getResult());
         } else {
             if (PlatformMBeanConstants.OBJECT_NAME.getName().equals(name)) {
-                final String objName = PlatformMBeanUtil.getObjectNameStringWithNameKey(ManagementFactory.GARBAGE_COLLECTOR_MXBEAN_DOMAIN_TYPE, gcName);
+                final String objName = gcMBean.getObjectName().toString();
```
bstansberry: Is this necessary? The PlatformMBeanUtil code puts the ObjectName key/value pairs in a fixed order. There's no guarantee of ordering from ObjectName.toString().

Related...prior to what GH is showing in the diff is a ```for (GarbageCollectorMXBean mbean : ManagementFactory.getGarbageCollectorMXBeans()) {``` loop. It needs the type check as well. AIUI if that type check passes we know we're in the right kind of mbean.

platform-mbean/src/main/java/org/jboss/as/platform/mbean/MemoryManagerMXBeanAttributeHandler.java
```diff
@@ -49,7 +49,7 @@ protected void executeReadAttribute(OperationContext context, ModelNode operatio
         }
 
         if (PlatformMBeanConstants.OBJECT_NAME.getName().equals(name)) {
-            final String objName = PlatformMBeanUtil.getObjectNameStringWithNameKey(ManagementFactory.MEMORY_MANAGER_MXBEAN_DOMAIN_TYPE, mmName);
+            final String objName = memoryManagerMXBean.getObjectName().toString();
```
bstansberry: Same comment as GarbageCollectorMXBeanAttributeHandler.

platform-mbean/src/main/java/org/jboss/as/platform/mbean/MemoryPoolMXBeanAttributeHandler.java
```diff
@@ -92,7 +92,7 @@ protected void executeWriteAttribute(OperationContext context, ModelNode operati
     static void storeResult(final String name, final ModelNode store, final MemoryPoolMXBean memoryPoolMXBean, final String memPoolName) {
 
         if (PlatformMBeanConstants.OBJECT_NAME.getName().equals(name)) {
-            final String objName = PlatformMBeanUtil.getObjectNameStringWithNameKey(ManagementFactory.MEMORY_POOL_MXBEAN_DOMAIN_TYPE, memPoolName);
+            final String objName = memoryPoolMXBean.getObjectName().toString();
```
bstansberry: Same comment as GarbageCollectorMXBeanAttributeHandler

platform-mbean/src/main/java/org/jboss/as/platform/mbean/GarbageCollectorMXBeanAttributeHandler.java
jfdenise: You are correct, I reverted this change.

bstansberry: @jfdenise There's still this block that I think needs the type check:

https://github.com/wildfly/wildfly-core/blob/main/platform-mbean/src/main/java/org/jboss/as/platform/mbean/GarbageCollectorMXBeanAttributeHandler.java#L43-L46

It's probably not *really* needed as the GarbageCollectorMXBeanResource change would stop the mgmt API reporting the existence of a resource with the wrong type. But if someone somehow knew of an mbean name that's a GarbageCollectorMXBean but the wrong type and typed it in as the address of a :read-attribute operation, then this OSH would try and provide info for it.

platform-mbean/src/main/java/org/jboss/as/platform/mbean/MemoryManagerMXBeanAttributeHandler.java
bstansberry: https://github.com/wildfly/wildfly-core/blob/main/platform-mbean/src/main/java/org/jboss/as/platform/mbean/MemoryManagerMXBeanAttributeHandler.java#L41-L45

platform-mbean/src/main/java/org/jboss/as/platform/mbean/MemoryPoolMXBeanAttributeHandler.java
bstansberry: https://github.com/wildfly/wildfly-core/blob/main/platform-mbean/src/main/java/org/jboss/as/platform/mbean/MemoryManagerMXBeanAttributeHandler.java#L41-L45

platform-mbean/src/main/java/org/jboss/as/platform/mbean/GarbageCollectorMXBeanAttributeHandler.java
jfdenise: @bstansberry Thanks. Fixed.

platform-mbean/src/main/java/org/jboss/as/platform/mbean/MemoryPoolMXBeanAttributeHandler.java
jfdenise: @bstansberry Thanks. Fixed.

platform-mbean/src/main/java/org/jboss/as/platform/mbean/MemoryManagerMXBeanAttributeHandler.java
jfdenise: @bstansberry Thanks. Fixed.

PR #6452, opened 2025-07-04
Fix for WFCORE-7284, Potential mismatch between provisioning of HAL bits and config enablement of the console

jfdenise:
Issue: https://issues.redhat.com/browse/WFCORE-7284

* management layer now disables the console
* introduce a feature-group to disable the console-enabled attribute (that will be used by WildFly to update default configs that need it).
* default standalone config, reset console-enabled attribute

A successful run of WildFly with the fix for WFLY-20748 based on this PR: https://ci.wildfly.org/buildConfiguration/WF_WildFlyCoreIntegrationExperiments/509259

### DIFF
--- a/core-feature-pack/galleon-common/src/main/resources/feature_groups/default-management-console.xml
+++ b/core-feature-pack/galleon-common/src/main/resources/feature_groups/default-management-console.xml
@@ -1,0 +1,11 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<!--
+  ~ Copyright The WildFly Authors
+  ~ SPDX-License-Identifier: Apache-2.0
+  -->
+
+<feature-group-spec name="default-management-console" xmlns="urn:jboss:galleon:feature-group:1.0">
+  <feature spec="core-service.management.management-interface.http-interface">
+    <unset param="console-enabled"/>
+  </feature>
+</feature-group-spec>
--- a/core-feature-pack/galleon-common/src/main/resources/layers/standalone/management/layer-spec.xml
+++ b/core-feature-pack/galleon-common/src/main/resources/layers/standalone/management/layer-spec.xml
@@ -12,11 +12,12 @@
     <feature-group name="unsecured-management"/>
 
     <feature spec="core-service.management.management-interface.http-interface">
+        <param name="console-enabled" value="false"/>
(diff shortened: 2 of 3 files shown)


core-feature-pack/galleon-common/src/main/resources/feature_groups/default-management-console.xml
```diff
@@ -0,0 +1,11 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<!--
+  ~ Copyright The WildFly Authors
+  ~ SPDX-License-Identifier: Apache-2.0
+  -->
+
+<feature-group-spec name="default-management-console" xmlns="urn:jboss:galleon:feature-group:1.0">
+  <feature spec="core-service.management.management-interface.http-interface">
+    <unset param="console-enabled"/>
```
bstansberry: @jfdenise I think setting it to true seems better. It seems cleaner if the configs we generate only produce 2 possible values, not 3. If they produce 3 that leads to the question of "Why is it sometimes 'true' and sometimes 'undefined'?" The answer to that is not something particularly understandable to a user, or even a WF developer.

bstansberry: I've been looking at https://github.com/jfdenise/wildfly/commit/fa53ca4c297efa750442da31ea8fa6ba80b9143e#r161707941 and between that commit and this PR it seems like all uses of this feature-group except https://github.com/wildfly/wildfly-core/pull/6452/files#diff-e2aa14846fc451db46b137aff0de6af2dd82d2c28c3fb9fd70d51dc7300a5802R20 are ones that could use the web-console layer. All the uses include layers asa part of their configuration. So I'm not sure why we need this.

jfdenise: @bstansberry , agreed. If we have only the true and false cases, then we can simplify. I followed your suggestion.

jfdenise:
@bstansberry , I opened a PR for full: https://github.com/wildfly/wildfly/pull/19066 that contains the fix + the overrides. So self contained fix with no dependency on core.

bstansberry (approved):

PR #6453, opened 2025-07-07
[WFCORE-7294] embed-server can delete the same tmp/vfs directory for running server

lvydra:
https://issues.redhat.com/browse/WFCORE-7294

### DIFF
--- a/embedded/src/main/java/org/wildfly/core/embedded/StandaloneSystemPropertyContext.java
+++ b/embedded/src/main/java/org/wildfly/core/embedded/StandaloneSystemPropertyContext.java
@@ -41,6 +41,6 @@ final class StandaloneSystemPropertyContext extends SystemPropertyContext {
         addPropertyIfAbsent(SERVER_CONTENT_DIR, resolvePath(dataDir, "content"));
         addPropertyIfAbsent(SERVER_DEPLOY_DIR, resolvePath(dataDir, "content"));
         addPropertyIfAbsent(SERVER_LOG_DIR, resolvePath(baseDir, "log"));
-        addPropertyIfAbsent(SERVER_TEMP_DIR, resolvePath(baseDir, "tmp"));
+        addPropertyIfAbsent(SERVER_TEMP_DIR, resolvePath(baseDir, "tmp/embedded-server"));
     }
 }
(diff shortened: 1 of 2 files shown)


embedded/src/main/java/org/wildfly/core/embedded/StandaloneSystemPropertyContext.java
```diff
@@ -41,6 +41,6 @@ void configureProperties() {
         addPropertyIfAbsent(SERVER_CONTENT_DIR, resolvePath(dataDir, "content"));
         addPropertyIfAbsent(SERVER_DEPLOY_DIR, resolvePath(dataDir, "content"));
         addPropertyIfAbsent(SERVER_LOG_DIR, resolvePath(baseDir, "log"));
-        addPropertyIfAbsent(SERVER_TEMP_DIR, resolvePath(baseDir, "tmp"));
+        addPropertyIfAbsent(SERVER_TEMP_DIR, resolvePath(baseDir, "tmp-embed-server"));
```
jamezp: I think we should use a sub-directory of `tmp`. Something like `tmp/embedded-server`.

lvydra:
Thank you @jamezp, updated.

jamezp (approved):
This looks good. In some ways I think we should use "tmp/embedded-server", but I don't have a strong opinion on that.

embedded/src/main/java/org/wildfly/core/embedded/StandaloneSystemPropertyContext.java
```diff
@@ -41,6 +41,6 @@ void configureProperties() {
         addPropertyIfAbsent(SERVER_CONTENT_DIR, resolvePath(dataDir, "content"));
         addPropertyIfAbsent(SERVER_DEPLOY_DIR, resolvePath(dataDir, "content"));
         addPropertyIfAbsent(SERVER_LOG_DIR, resolvePath(baseDir, "log"));
-        addPropertyIfAbsent(SERVER_TEMP_DIR, resolvePath(baseDir, "tmp"));
+        addPropertyIfAbsent(SERVER_TEMP_DIR, resolvePath(baseDir, "tmp/embed-server"));
```
bstansberry: s/embed/embedded

It's a nit but a directory name is usually a noun and embedded-server is a noun phrase.

yersan:
Thanks @lvydra @jamezp @bstansberry

PR #6454, opened 2025-07-08
WFLY-20048 Shutdown operation should not block (potentially indefinitely) since it holds the ModelController lock.

pferraro:
This otherwise prevents the user from issuing a resume operation to cancel a prolonged suspend.
https://issues.redhat.com/browse/WFLY-20048

Also, WFCORE-7296 Ensure shutdown operations that wait on server suspension are interruptable
https://issues.redhat.com/browse/WFCORE-7296

As discussed, the cancellation of the future returned by ServerSuspendController.suspend/resume(...) now propagates back to ServerActivities, so that they can respond accordingly.  I added unit tests to validate this logic.

### DIFF
--- a/server/src/main/java/org/jboss/as/server/operations/ServerSuspendHandler.java
+++ b/server/src/main/java/org/jboss/as/server/operations/ServerSuspendHandler.java
@@ -13,6 +13,7 @@ import static org.jboss.as.server.controller.resources.ServerRootResourceDefinit
 import static org.jboss.as.server.controller.resources.ServerRootResourceDefinition.renameTimeoutToSuspendTimeout;
 
 import java.util.EnumSet;
+import java.util.concurrent.CancellationException;
 import java.util.concurrent.CompletionStage;
 import java.util.concurrent.ExecutionException;
 import java.util.concurrent.TimeUnit;
@@ -83,8 +84,9 @@ public class ServerSuspendHandler implements OperationStepHandler {
                         }
                     } catch (InterruptedException e) {
                         Thread.currentThread().interrupt();
+                        context.setRollbackOnly();
                     } catch (TimeoutException e) {
-                        // Done waiting
+                        // Stop waiting, but do not rollback
                     } catch (ExecutionException e) {
                         context.getFailureDescription().set(e.getCause().getLocalizedMessage());
                         context.setRollbackOnly();
@@ -97,7 +99,7 @@ public class ServerSuspendHandler implements OperationStepHandler {
     }
 
     static void suspendComplete(Void result, Throwable exception) {
-        if (exception != null) {
+        if ((exception != null) && !(exception instanceof CancellationException)) {
             ServerLogger.ROOT_LOGGER.suspendFailed(exception);
         }
     }
--- a/server/src/main/java/org/jboss/as/server/operations/ServerResumeHandler.java
+++ b/server/src/main/java/org/jboss/as/server/operations/ServerResumeHandler.java
@@ -11,6 +11,7 @@ import static org.jboss.as.controller.descriptions.ModelDescriptionConstants.RES
 import static org.jboss.as.controller.descriptions.ModelDescriptionConstants.RUNNING_SERVER;
 
 import java.util.EnumSet;
+import java.util.concurrent.CancellationException;
 import java.util.concurrent.CompletionStage;
 import java.util.concurrent.ExecutionException;
 
@@ -64,11 +65,13 @@ public class ServerResumeHandler implements OperationStepHandler {
                     public void handleResult(OperationContext.ResultAction resultAction, OperationContext context, ModelNode operation) {
                         if (resultAction == OperationContext.ResultAction.KEEP) {
                             ServerLogger.ROOT_LOGGER.resumingServer();
-                            CompletionStage<Void> resume = ServerResumeHandler.this.suspendController.resume(ServerSuspendController.Context.RUNNING).whenComplete(ServerResumeHandler::resumeComplete);
+                            CompletionStage<Void> resume = ServerResumeHandler.this.suspendController.resume(ServerSuspendController.Context.RUNNING);
+                            resume.whenComplete(ServerResumeHandler::resumeComplete);
                             try {
                                 resume.toCompletableFuture().get();
                             } catch (InterruptedException e) {
                                 Thread.currentThread().interrupt();
+                                resume.toCompletableFuture().cancel(false);
                             } catch (ExecutionException e) {
                                 context.getFailureDescription().set(e.getCause().getLocalizedMessage());
                             }
@@ -80,7 +83,7 @@ public class ServerResumeHandler implements OperationStepHandler {
     }
 
     static void resumeComplete(Void result, Throwable exception) {
-        if (exception != null) {
+        if ((exception != null) && !(exception instanceof CancellationException)) {
             ServerLogger.ROOT_LOGGER.resumeFailed(exception);
         }
     }


bstansberry (dismissed):

server/src/main/java/org/jboss/as/server/operations/ServerShutdownHandler.java
```diff
@@ -144,21 +144,15 @@ public void execute(OperationContext context, ModelNode operation) throws Operat
...
-                            try {
-                                suspend.join();
-                                shutdown.shutdown();
-                            } catch (CancellationException e) {
-                                shutdown.cancel();
+                                // Auto-complete suspend if not complete with the configured timeout
+                                suspend.toCompletableFuture().completeOnTimeout(null, seconds, TimeUnit.SECONDS);
                             }
+                            // Shutdown when complete
+                            suspend.whenComplete(new ShutdownAction(getOperationName(operation), restart, performInstallation));
```
yersan, line 155: Adding this complete here will invoke the shutdown action always, independently of whether there is an exception during the suspension.

This should be moved up to line 149 and invoke the shutdown only if the exception is null

server/src/main/java/org/jboss/as/server/suspend/SuspendController.java
```diff
@@ -72,60 +75,67 @@ public class SuspendController implements ServerSuspendController, SuspendableAc
...
-        this.state = State.PRE_SUSPEND;
+        CompletableFuture<Void> result = new CompletableFuture<>();
+        this.activeSuspend = result;
         for (OperationListener listener: this.listeners) {
             listener.suspendStarted();
         }
-        CompletableFuture<Void> result = new CompletableFuture<>();
+        // Collect stages in case we need to cancel them
+        List<CompletionStage<Void>> phaseStages = new ArrayList<>(2);
+        result.whenComplete(propagateCancellation(phaseStages));
```
yersan, line 99: Optional, but would `result.exceptionally(...)` fit better here?

server/src/main/java/org/jboss/as/server/operations/ServerSuspendHandler.java
```diff
@@ -71,21 +72,36 @@ public void execute(final OperationContext context, ModelNode operation) {
...
+                if (seconds != 0) {
+                    try {
+                        if (seconds < 0) {
+                            // Wait indefinitely
+                            suspend.toCompletableFuture().get();
+                        } else {
+                            suspend.toCompletableFuture().get(seconds, TimeUnit.SECONDS);
+                        }
+                    } catch (InterruptedException e) {
+                        Thread.currentThread().interrupt();
```
yersan, line 86: Out of curiosity, should we not somehow invoke here suspend.toCompletableFuture().cancel()? ... Otherwise, I do not get how after this thread gets interrupted make the suspension be cancelled with a CancellationException

server/src/main/java/org/jboss/as/server/operations/ServerShutdownHandler.java
yersan: Well, now I saw it, maybe another option is to change L182 and check the exception nullity there

yersan (changes requested):

server/src/main/java/org/jboss/as/server/operations/ServerShutdownHandler.java
pferraro: > Adding this complete here will invoke the shutdown action always, independently of whether there is an exception during the suspension.

This merely registers ShutdownAction as a completion handler.  Whether an exceptional completion should shutdown or not is determined by its accept(...) implementation.

In this proposed change, we only abort shutdown due to a cancellation.  I considered skipping shutdown for all exceptional completions, but this would not have been consistent with the previous implementation (where there was no mechanism to communicate exceptions back to the operation).  If you are ok with making that change, I agree that it is preferrable.

server/src/main/java/org/jboss/as/server/suspend/SuspendController.java
pferraro: exceptionally(...) is used to translate completion outcomes, i.e. to "handle" the exception by returing a value.  Technically, since we do not use the translated stage, we could use exceptionally(...) instead, but I thought that might be a source of confusion, since exceptions in the returned stage are intentionally _not_ handled.

server/src/main/java/org/jboss/as/server/operations/ServerSuspendHandler.java
pferraro: Can we be sure that the intent of the interrupt was to cancel the suspend?  A more conservative interpretation is that the caller did not want to wait any longer for suspend to complete.

yersan: > Can we be sure that the intent of the interrupt was to cancel the suspend? 

If that was not the intention, we could be allowing the suspending task to continue doing whatever was doing but we would return from the main suspend operation being interrupted, so I think it could be more fragile.

If we do not explicitly cancel the task, would the suspension be rollbacked to the original status in a controlled manner?

I think the same discussion applies to not only receiving a `InterruptedException`, but also in case of getting a `TimeoutException`

server/src/main/java/org/jboss/as/server/operations/ServerShutdownHandler.java
yersan: Ok, I follow you. I am more inclined to not shutdown if the suspending tasks did not complete correctly, otherwise we can run into the risk of shutting down the server when the suspension, for some reason failed, and if the suspension failed I would consider the shutdown operation failed, since this is not a :suspend, it is a :shutdown

Although if that was the old behaviour, we can be more conservative here. Maybe gather additional povs .... @bstansberry WDYT?

server/src/main/java/org/jboss/as/server/operations/ServerSuspendHandler.java
pferraro: > If that was not the intention, we could be allowing the suspending task to continue doing whatever was doing but we would return from the main suspend operation being interrupted, so I think it could be more fragile.

How so?  The interrupt handling here is equivalent to truncating the suspend-timeout.

> If we do not explicitly cancel the task, would the suspension be rollbacked to the original status in a controller manner?

In this context, cancelling the task does not rollback anything.  An exceptionally completed suspend can only be "fixed" by issuing a resume.  If we want to abort shutdown in the event of an non-cancellation exceptional response from the suspend controller (again, this would be new behavior - as this is not how the shutdown handler used to work), then we would also want to add a rollback handler that issues a resume (in the same manner as the suspend operation).

However, I am not convinced that this behavior is warranted.  The user issues the shutdown operaton because they want to shutdown the server.  Even if some suspend activity completes exceptionally, the intent of the user is clear: that the server should shutdown.  We already have a separate suspend operation that covers the suspend w/resume rollback use case.

> I think the same discussion applies to not only receiving a InterruptedException, but also in case of getting a TimeoutException

In this context, a timeout exception means that our suspend-timeout has elapsed.  The behavior of is already well defined, i.e. we proceed with shutdown: https://github.com/wildfly/wildfly-core/blob/main/server/src/main/resources/org/jboss/as/server/controller/descriptions/LocalDescriptions.properties#L103

server/src/main/java/org/jboss/as/server/operations/ServerShutdownHandler.java
pferraro: > If you are ok with making that change, I agree that it is preferrable.

Clearly, I am waffling...

One this is certain - exceptional completion during the prepare phase should cancel shutdown.  This is consistent with the old behavior.
I just realised this was not happening.  See the last commit.

server/src/main/java/org/jboss/as/server/operations/ServerSuspendHandler.java
yersan: > > I think the same discussion applies to not only receiving a InterruptedException, but also in case of getting a TimeoutException

> In this context, a timeout exception means that our suspend-timeout has elapsed. The behavior of is already well defined, i.e. we proceed with shutdown: https://github.com/wildfly/wildfly-core/blob/main/server/src/main/resources/org/jboss/as/server/controller/descriptions/LocalDescriptions.properties#L103

Upss, sorry, yes, forget I mentioned `TimeoutException`. It does not make sense what I said.

yersan: > > If that was not the intention, we could be allowing the suspending task to continue doing whatever was doing but we would return from the main suspend operation being interrupted, so I think it could be more fragile.

> How so? 

I've just now realized that what I have been asking here is what is implemented. After getting the InterruptedException in the thread, which is waiting to get the result of the suspend task, the RollbackHandler of the ServerSuspendHandler enters in action, and is the one that triggers the resume which at the end is the one that cancels the suspend task. I did not realized about it when I started the discussion.

I was asking to cancel the suspending task after getting the InterruptedException, but this is done by issuing the resume of the suspend controller. I want to do an additional review to finally get all the details, but ok, seems to be good from me

yersan (approved):
Approved meaning I do not have additional concerns, it still needs to be verified with the CI results, but looks good to me

yersan:
Thanks @pferraro @bstansberry

PR #6455, opened 2025-07-09
WFCORE-7297 ModelTestControllerVersion#getMavenGav produces invalid coordinates for core artifacts + WFCORE-7298 Replace legacy ModelTestControllerVersion#getMavenGav with proper methods to format Maven coordinates

rhusar:
Resolves
https://issues.redhat.com/browse/WFCORE-7297
https://issues.redhat.com/browse/WFCORE-7298

The method org.jboss.as.model.test.ModelTestControllerVersion#getMavenGav incorrectly uses version of the main artifact even when specified as core artifact resulting in unusable maven coordinates.
The issue was never uncovered as this is not used, but it is required for EAP 8.1 clustering transformers.
The org.jboss.as.model.test.ModelTestControllerVersion#getMavenGav is legacy and problematic.
First of all, the method is incorrectly named - it's not a getter of any sorts.
Using both 'maven' and 'gav' in the same method is redundant, GAV is a maven concept.
Moreover, GAV should be all caps as it's already an abbreviation.
It makes use of irrelevant org.jboss.as.model.test.ModelTestControllerVersion#getArtifactIdPrefix which was last used in EAP 6.

Also... runs formatter on this horrible mess.

### DIFF
--- a/model-test/src/main/java/org/jboss/as/model/test/ModelTestControllerVersion.java
+++ b/model-test/src/main/java/org/jboss/as/model/test/ModelTestControllerVersion.java
@@ -18,7 +18,7 @@ import org.wildfly.legacy.version.LegacyVersions;
  */
 public enum ModelTestControllerVersion {
     // Release version under test
-    MASTER (CurrentVersion.VERSION, false, null, "master" ),
+    MASTER(CurrentVersion.VERSION, false, null, "master"),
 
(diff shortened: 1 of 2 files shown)


yersan (approved):

yersan:
Thanks @rhusar

PR #6456, opened 2025-07-11
[WFCORE-7299] Add a dco.txt file

bstansberry:
Commonhaus requires that projects have either a Developer Certificate of Origin system or a CLA system. A dco.txt file is part of adding a Developer Certificate of Origin system. (The other part is linking too it from a CONTRIBUTING.md, which will be a separate commit.)

https://issues.redhat.com/browse/WFCORE-7299

### DIFF
--- a/dco.txt
+++ b/dco.txt
@@ -1,0 +1,34 @@
+Developer Certificate of Origin
+Version 1.1
+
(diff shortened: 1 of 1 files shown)


yersan:
Thanks @bstansberry

PR #6465, opened 2025-07-22
[WFCORE-7303]: GitRepository.clearExistingFiles removes file in the 'ignored' set

ehsavoie:
* Using gitignore rules to define which file/directory to clean

Issue: https://issues.redhat.com/browse/WFCORE-7303

### DIFF
--- a/testsuite/manualmode/src/test/java/org/jboss/as/test/manualmode/management/persistence/AbstractGitRepositoryTestCase.java
+++ b/testsuite/manualmode/src/test/java/org/jboss/as/test/manualmode/management/persistence/AbstractGitRepositoryTestCase.java
@@ -82,7 +82,7 @@ public class AbstractGitRepositoryTestCase {
         if (emptyRemoteRepository != null) {
             emptyRemoteRepository.close();
         }
-        FileUtils.delete(emptyRemoteRoot.getParent().toFile(), FileUtils.RECURSIVE | FileUtils.RETRY);
+        FileUtils.delete(emptyRemoteRoot.getParent().toFile(), FileUtils.RECURSIVE | FileUtils.RETRY | FileUtils.SKIP_MISSING);
     }
 
     protected Repository createRepository() throws IOException {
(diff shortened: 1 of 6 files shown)


testsuite/manualmode/src/test/java/org/jboss/as/test/manualmode/management/persistence/AbstractGitRepositoryTestCase.java
```diff
@@ -82,7 +82,7 @@ protected void closeEmptyRemoteRepository() throws Exception {
         if (emptyRemoteRepository != null) {
             emptyRemoteRepository.close();
         }
-        FileUtils.delete(emptyRemoteRoot.getParent().toFile(), FileUtils.RECURSIVE | FileUtils.RETRY);
+        FileUtils.delete(emptyRemoteRoot.getParent().toFile(), FileUtils.RECURSIVE | FileUtils.RETRY | FileUtils.SKIP_MISSING | FileUtils.IGNORE_ERRORS);
```
yersan: `FileUtils.IGNORE_ERRORS`, do we really want to silently ignore any errors if the file is not deleted?I guess allowing to fail in these cases would make it more robust and help detect unexpected code behaviors ...

ehsavoie: This is about cleaning the files after the tests, having a test to fail on Windows because the file handle wasn't properly released at the end of the test is going to create intermittent errors without any real value in my opinion

yersan: > Windows because the file handle wasn't properly released at the end of the test is going to create intermittent errors

The question then is why the file handler is not properly released at the end of the test, is it because the the test code? it is because the server code? and, if it is not properly released, would it affect to other tests in the testsuite? For this last question, I guess no because the files are in a temp directory now, but my first impression is that ignore errors, even after the test has finished, can hide other problems and it is not a good practice. I'm only commenting about this to be sure it is want we want.

ehsavoie: i don't know the internal of Windows file handles but sometimes the handle is not effectively released at the time we try to delete the file even if emptyRemoteRepository.close(); has been called. Also this might be environmental (like having a github desktop or IDE running with this folder open)

yersan: I would feel more confortable if at least we can detect which one was not deleted and at least get a warn trace with such an information. In any case, in general, it would be nice to hear more opinions about how to properly dealing with this.

yersan (approved):

PR #6467, opened 2025-07-28
[WFCORE-7310] ear-exclusions-cascaded-to-subdeployments doesn't work correctly

mskacelik:
https://issues.redhat.com/browse/WFCORE-7310

The fix is in that I used `subDeploymentMap`, which holds info for all the "global" sub-deployments, instead of `result.getSubDeploymentSpecifications()`, which only contains the explicitly set sub-deployments. I also removed the unnecessary helper methods.

Tested with the reproducer in the issue:
image
And also tested with integration tests within WF: 
https://github.com/wildfly/wildfly/compare/main...mskacelik:wildfly:WFCORE-7310-TEST (WIP, Proof of Concept)
The test alone might spark another discussion if there is even a need for the original explicit subsystem test.


yersan:
Hello @mskacelik , thanks for the PR. We need to have the testcase in place in order to fully verify this. I suggest to go ahead by creating the testcase PR (without the wildfly core version bump) and let it to fail as expected. Then, we can use our CI to run an Integration Job that verifies both of your topic branches, the one for core and the one for wildfly, and that one should pass.

mskacelik:
@yersan hopefully fixed now, and created a PR for the tests as you wanted https://github.com/wildfly/wildfly/pull/19095
The error for the test failures was my fault, due to my not understanding a trivial for loop and its implications. Some tests seem to still fail, but they seem to be failing even before the given commit.

yersan (approved):

mskacelik:
> Linux Integration Job: https://ci.wildfly.org/buildConfiguration/WF_WildFlyCoreIntegrationExperiments/512668
> 
> Windows Integration Job: https://ci.wildfly.org/buildConfiguration/WF_WildFlyCoreIntegrationExperimentsWindows/512670

Linux TC jobs were successful. But the Windows ones, some tests regarding permissions seem to fail, but I am not really sure why that is.

yersan:
Thanks @mskacelik

PR #6472, opened 2025-08-01
[WFCORE-7316] Server: don't use relative content path

tadamski:
https://issues.redhat.com/browse/WFCORE-7316


yersan:
Adding hold label since this issue is still under discussion in Jira

tadamski:
@bstansberry  ok, closing it then

PR #6482, opened 2025-08-11
[WFCORE-7325] Upgrade smallrye-common to 2.13.8

yersan:
Jira issue: https://issues.redhat.com/browse/WFCORE-7325

PR #6492, opened 2025-08-18
[WFCORE-7337] Upgrade wildfly-common to 2.0.1

yersan:
built on https://github.com/wildfly/wildfly-core/pull/6482

Jira issue: https://issues.redhat.com/browse/WFCORE-7337

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -272,7 +272,7 @@
         <version.org.slf4j>2.0.17</version.org.slf4j>
         <version.org.syslog4j>0.9.30</version.org.syslog4j>
         <version.org.wildfly.client.config>1.0.1.Final</version.org.wildfly.client.config>
-        <version.org.wildfly.common>1.7.0.Final</version.org.wildfly.common>
+        <version.org.wildfly.common>2.0.1</version.org.wildfly.common>
         <version.org.wildfly.discovery>1.3.0.Final</version.org.wildfly.discovery>
         <version.org.wildfly.installation-manager.installation-manager-api>1.1.1.Final</version.org.wildfly.installation-manager.installation-manager-api>
         <version.org.wildfly.launcher>1.0.2.Final</version.org.wildfly.launcher>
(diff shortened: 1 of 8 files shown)


pom.xml
```diff
@@ -787,6 +787,16 @@
...
+            <dependency>
+                <groupId>io.smallrye.common</groupId>
+                <artifactId>smallrye-common-net</artifactId>
+                <version>${version.io.smallrye.smallrye-common}</version>
+            </dependency>
+            <dependency>
+                <groupId>io.smallrye.common</groupId>
+                <artifactId>smallrye-common-os</artifactId>
+                <version>${version.io.smallrye.smallrye-common}</version>
+            </dependency>
```
pferraro: Why not replace all of these with a single bom import?

```xml
<dependency>
    <groupId>io.smallrye.common</groupId>
    <artifactId>smallrye-common-bom</artifactId>
    <version>${version.io.smallrye.smallrye-common}</version>
    <type>pom</type>
    <scope>import</scope>
</dependency>
```

pferraro (changes requested):

pom.xml
pferraro: Oh - I suppose this comment belongs here: https://github.com/wildfly/wildfly-core/pull/6482

yersan: Hi @pferraro , thanks for the review. I understand it will bring more dependencies that we do not use to the dependency management section, which potentially could warn if there is a CVE on any of them, although since we are not declaring them, maybe it is more than OK since they won't be even downloaded by maven. So, I think that will be OK, going to change it

yersan: although ... bring more without needing is also an argument against it ... so, it will be a balance between it and parent pom organization/simplicity

pferraro: WDYM by "bring more without needing"?

yersan: > WDYM by "bring more without needing"?

I mean that there are smallrye-common-bom dependencies that are not used at all by the server. I think building the project with them being on the parent declaration using the bom will make them be resolved and downloaded even if they are not used at all ... well really, I need to test it

yersan: Well, ok, I guess the correct approach is to declare the smallrye common bom into the dependency management section, instead of the dependencies section to avoid that issue. And declare what we need into the dependencies section instead

yersan: Upps sorry, my mistake; it is already on the management dependencies section, so it will bring only what's needed. Forget my initial comment.

yersan: @pferraro I think we are fine with this now, do you have any other concerns?

yersan:
Thanks for the review @pferraro

PR #6498, opened 2025-09-02
WIP: [WFCORE-7335] Block deployments if the security manager subsyste…

mskacelik:
/cc @darranl 
issue: https://issues.redhat.com/browse/WFCORE-7335

This is only a draft PR to showcase a possible solution.

When `standalone.xml` security manager subsystem configuration is invalid, f.g: 
```xml
<subsystem xmlns="urn:jboss:domain:security-manager:1.0">
    <deployment-permissions>
        <maximum-set>
            <permission class="java.io.FilePermission" name="${badExpression}" actions="write,delete"/>
        </maximum-set>
    </deployment-permissions>
</subsystem>
```
Upon WF instance startup, the `ExpressionResolverImpl.java` will log and throw an exception of type `OperationClientException`
https://github.com/wildfly/wildfly-core/blob/70bb2a1953d7e8cd9d8f0dbbe542f3574741b6e5/controller/src/main/java/org/jboss/as/controller/ExpressionResolverImpl.java#L282
> 
`OperationClientException` means that:
```java
This class implements {@link OperationClientException}, so if it is thrown during execution of an {@code OperationStepHandler}, the management kernel will adequately handle the exception as a user mistake, not a server fault.
````
This exception is then handled in the `AbstractOperationContext`:
https://github.com/wildfly/wildfly-core/blob/70bb2a1953d7e8cd9d8f0dbbe542f3574741b6e5/controller/src/main/java/org/jboss/as/controller/AbstractOperationContext.java#L1074-L1084

So here is the fundamental problem.

### Possible Solution
I have come up with a solution by wrapping the `OperationClientException` of the expression resolver in the `SecurityManagerSubsystemAdd`, which, from my understanding, is only executed during the start-up of WildFly (boot). This wrapped exception is not handled in the `AbstractOperationContext`, making the WildFly startup fail due to the exception.

So, in the current implementation with invalid configuration:
 - using CLI => roll back the configuration (with or without `-secmgr`)
 - booting without `-secmgr` => logs the error, but the boot won't fail
 - booting with `-secmgr` => logs the error, but boot will fail
 - `--admin-only` mode =>  both CLI and booting (invalid `XML`) won't fail (with or without `-secmgr`)
 
> [!NOTE]
> - Is the `--admin-only` mode behavior valid in this case?
> - PR is missing the test, one because this is a Draft PR, so given that this solution might not be ideally implemented, I did not implement the tests yet, and secondly, I was not sure where to put these tests, if in `wildfly-core` repository or in the `wildfly` (repository) integration tests.
> - Due to this PR being a PoC, I only used `RuntimeException`, but maybe other exceptions would be suited better.

### DIFF
--- a/security-manager/src/main/java/org/wildfly/extension/security/manager/SecurityManagerSubsystemAdd.java
+++ b/security-manager/src/main/java/org/wildfly/extension/security/manager/SecurityManagerSubsystemAdd.java
@@ -22,6 +22,7 @@ import java.util.ArrayList;
 import java.util.List;
 
 import org.jboss.as.controller.AbstractBoottimeAddStepHandler;
+import org.jboss.as.controller.ExpressionResolver;
 import org.jboss.as.controller.OperationContext;
 import org.jboss.as.controller.OperationFailedException;
 import org.jboss.as.controller.PathAddress;
@@ -62,14 +63,21 @@ class SecurityManagerSubsystemAdd extends AbstractBoottimeAddStepHandler {
         final ModelNode node = Resource.Tools.readModel(resource);
         // get the minimum set of deployment permissions.
         final ModelNode deploymentPermissionsModel = node.get(DEPLOYMENT_PERMISSIONS_PATH.getKeyValuePair());
-        final ModelNode minimumPermissionsNode = MINIMUM_PERMISSIONS.resolveModelAttribute(context, deploymentPermissionsModel);
-        final List<PermissionFactory> minimumSet = this.retrievePermissionSet(DeferredPermissionFactory.Type.MINIMUM_SET, context, minimumPermissionsNode);
-
-        // get the maximum set of deployment permissions.
-        ModelNode maximumPermissionsNode = MAXIMUM_PERMISSIONS.resolveModelAttribute(context, deploymentPermissionsModel);
+        ModelNode minimumPermissionsNode = null;
+        ModelNode maximumPermissionsNode = null;
+        try {
+            minimumPermissionsNode = MINIMUM_PERMISSIONS.resolveModelAttribute(context, deploymentPermissionsModel);
+            maximumPermissionsNode = MAXIMUM_PERMISSIONS.resolveModelAttribute(context, deploymentPermissionsModel);
+        } catch (ExpressionResolver.ExpressionResolutionUserException ex) {
+            // TODO: better exception choice ?
+            throw (System.getSecurityManager() != null) ? new RuntimeException(ex) : ex;
+        }
         if (!maximumPermissionsNode.isDefined())
             maximumPermissionsNode = DEFAULT_MAXIMUM_SET;
-        final List<PermissionFactory> maximumSet = this.retrievePermissionSet(DeferredPermissionFactory.Type.MAXIMUM_SET, context, maximumPermissionsNode);
+        final List<PermissionFactory> minimumSet = this
+                .retrievePermissionSet(DeferredPermissionFactory.Type.MINIMUM_SET, context, minimumPermissionsNode);
+        final List<PermissionFactory> maximumSet = this
+                .retrievePermissionSet(DeferredPermissionFactory.Type.MAXIMUM_SET, context, maximumPermissionsNode);
 
         // validate the configured permissions - the minimum set must be implied by the maximum set.
         final FactoryPermissionCollection maxPermissionCollection = new FactoryPermissionCollection(maximumSet.toArray(new PermissionFactory[maximumSet.size()]));

### CODE REVIEW

File: security-manager/src/main/java/org/wildfly/extension/security/manager/SecurityManagerSubsystemAdd.java [hunk 1]
```diff
@@ -62,14 +63,21 @@ protected void performBoottime(final OperationContext context, final ModelNode o
...
-        // get the maximum set of deployment permissions.
-        ModelNode maximumPermissionsNode = MAXIMUM_PERMISSIONS.resolveModelAttribute(context, deploymentPermissionsModel);
+        ModelNode minimumPermissionsNode = null;
+        ModelNode maximumPermissionsNode = null;
+        try {
+            minimumPermissionsNode = MINIMUM_PERMISSIONS.resolveModelAttribute(context, deploymentPermissionsModel);
+            maximumPermissionsNode = MAXIMUM_PERMISSIONS.resolveModelAttribute(context, deploymentPermissionsModel);
+        } catch (ExpressionResolver.ExpressionResolutionUserException ex) {
+            // TODO: better exception choice ?
+            throw (System.getSecurityManager() != null) ? new RuntimeException(ex) : ex;
```
bstansberry: Just throw the original exception.

Don't use an exception type to try and influence the behavior of the operation; use OperationContext.setRollbackOnly().  

I also don't think throwing a RuntimeException would abort the boot anyway. :) It would roll back a change after boot, but would not abort boot. OperationContext.setRollbackOnly() aborts the boot.

Please include a quick comment explaining the rationale for doing that; i.e. don't make people use git blame and research the JIRA.

Also, the current code at L84 throws an exception in a similar situation (detection of invalid config), so if the goal is to abort the boot, the same SM check + OperationContext.setRollbackOnly() seems applicable.

mskacelik: Thanks for the info, I was not aware of the `OperationContext.setRollbackOnly()`, which seems to solve the issue.

With that said, yesterday I was trying to come up with an Integration test, something similar to https://github.com/wildfly/wildfly/blob/68bc534d9ef0adf349c10b278d7a310d429763ab/testsuite/integration/vdx/src/test/java/org/wildfly/test/integration/vdx/standalone/SmokeStandaloneTestCase.java. (groovy file for modifying the `standalone.xml` with invalid config, checking logs...)

But it seems that I had some issues regarding setting up `-secmgr` and pretty much checking halting/non-halting asserts. Do you think I am on the right track (vdx/standalone), or is there a better test solution that you can direct me to?

bstansberry: A test wouldn't belong in full testsuite/vdx. That testsuite module is only about testing handling of reporting of XML errors. This isn't really a matter of an XML error.

This seems like something that can be tested in WildFly Core's testsuite/manualmode. It needs to be in manualmode because that module is the place for tests that need to control the entire lifecycle of the server. I assume you need to do that as the test will cause the server VM to exit.

Have a look at the org.jboss.as.test.shared.AssumeTestGroupUtil class, which allows a test to see if it is running with the security manager enabled. You don't need to worry about turning on the SM. WF Core runs different test jobs, and PRs get tested with a job that has the SM one. (See the status block below.) There are also nightly jobs with the SM enabled. So just write the test to always run and use AssumeTestGroupUtil to control the behavior of the test based on whether the SM is enabled. If there is nothing useful to test if the SM is not enabled, then throw an AssumptionViolatedException if AssumeTestGroupUtil.isSecurityManagerDisabled() == true.

It's possible you don't need a broken config in the initial boot of the server. If not, that may give you more flexibility.  The test can reload an existing server into admin-only mode, add a config that will fail in a normal boot, and then reload into normal mode. And then check what happened (plus clean up). What you are adding here are checks that will not execute in an admin-only server. The performBoottime method should not execute in admin-only.

See also the org.jboss.as.test.shared.logging package in testsuite/shared for utilities that can be used to capture logging data. I assume what the test would need to do is look at logging data. The reload will fail but I'm pretty sure you can't get useful data from any failure response. It would be in the log.

security-manager/src/main/java/org/wildfly/extension/security/manager/SecurityManagerSubsystemAdd.java
```diff
@@ -62,14 +63,25 @@ protected void performBoottime(final OperationContext context, final ModelNode o
         final ModelNode node = Resource.Tools.readModel(resource);
         // get the minimum set of deployment permissions.
         final ModelNode deploymentPermissionsModel = node.get(DEPLOYMENT_PERMISSIONS_PATH.getKeyValuePair());
-        final ModelNode minimumPermissionsNode = MINIMUM_PERMISSIONS.resolveModelAttribute(context, deploymentPermissionsModel);
-        final List<PermissionFactory> minimumSet = this.retrievePermissionSet(DeferredPermissionFactory.Type.MINIMUM_SET, context, minimumPermissionsNode);
-
-        // get the maximum set of deployment permissions.
-        ModelNode maximumPermissionsNode = MAXIMUM_PERMISSIONS.resolveModelAttribute(context, deploymentPermissionsModel);
+        ModelNode minimumPermissionsNode = null;
+        ModelNode maximumPermissionsNode = null;
```
yersan: optional, null assignements can be removed, those variables will never be used without being initialized, so can be just:

```
        ModelNode minimumPermissionsNode;
        ModelNode maximumPermissionsNode;
```

testsuite/manualmode/src/test/java/org/jboss/as/test/manualmode/secman/UnresolvedExpressionSecurityManagerTestCase.java
```diff
@@ -0,0 +1,194 @@
...
+    private void triggerFailedReloadToNormalMode(ModelControllerClient client) throws Exception {
+        ModelNode reloadOp = Operations.createOperation("reload");
+        reloadOp.get("admin-only").set(false);
+        try {
+            client.execute(reloadOp);
+        } catch (IOException e) {
+            // connection closes exception
+        }
+        // flush logs
+        Thread.sleep(2000);
```
yersan: I think you will need to manage two cases here:

1. If sec mgr was enabled, this reload will end up in a server stopped.
In this case, you can send the reload operation, assert that the reload operation was executed succesfully and then call the container.stop(), to ensure it waits for the process to be fully stopped.

2. If sec mgr is not enabled, the reload will success.
So, in this case, you should be able to use `ServerReload.executeReloadAndWaitForCompletion(client, false);`

The idea is to get rid of that `Thread.sleep(2000);` and be more deterministic in the status after this method ends.

testsuite/manualmode/src/test/java/org/jboss/as/test/manualmode/secman/UnresolvedExpressionSecurityManagerTestCase.java
```diff
@@ -0,0 +1,194 @@
...
+        }
+        assertTrue("Expected server boot failure (" + BOOT_FAILURE_ID + ") in " + LOG_FILE_NAME,
+                foundBootFailure);
+
+    }
+
+    private void cleanupSecurityManagerSubsystem(ModelControllerClient client,
+            boolean isMaximumPermissions) throws Exception {
+        container.stop();
+        container.startInAdminMode();
```
yersan: If the container is not already stopped here, you could use a reload instead:

```
ServerReload.executeReloadAndWaitForCompletion(client, true);
```

If the container was already stopped, you will need to start it in admin-only mode and re-initialize the client variable after starting it:

```
container.startInAdminMode()
client = TestSuiteEnvironment.getModelControllerClient();
```

testsuite/manualmode/src/test/java/org/jboss/as/test/manualmode/secman/UnresolvedExpressionSecurityManagerTestCase.java
mskacelik: Interesting that the failed reload is considered as a successful outcome. But the test seems to support it :).

testsuite/manualmode/src/test/java/org/jboss/as/test/manualmode/secman/UnresolvedExpressionSecurityManagerTestCase.java
```diff
@@ -0,0 +1,230 @@
+/*
+ * Copyright The WildFly Authors SPDX-License-Identifier: Apache-2.0
+ */
```
yersan, line 4: For consistency, please use the standard format of the license header, two lines:

```
/*
 * Copyright The WildFly Authors
 * SPDX-License-Identifier: Apache-2.0
 */
```

testsuite/manualmode/src/test/java/org/jboss/as/test/manualmode/secman/UnresolvedExpressionSecurityManagerTestCase.java
```diff
@@ -0,0 +1,230 @@
...
+        setUpLogDirPath();
+    }
+
+    @After
+    public void tearDown() throws Exception {
+        try {
+            if (client != null) {
+                logHandlerSetup.tearDown(client);
+                client.close();
+            }
```
yersan, line 68: Regarding the problems on Windows, see this note in the TestLogHandlerSetupTask

https://github.com/wildfly/wildfly-core/blob/035d41260346840f0073a90875997f2b8f16df17/testsuite/shared/src/main/java/org/jboss/as/test/shared/logging/TestLogHandlerSetupTask.java#L95-L97

So, it looks like the test could not be correctly triggering down the TestLogHandlerSetupTask. So, maybe the test is arriving here with an invalid  `client` variable, which drives us to the code at L167:

```
client = TestSuiteEnvironment.getModelControllerClient();
```
You are modifying the local client reference, but not the one defined as class field instance.

Not sure if that would resolve the issue you currently have, but it seems it could be related. 

I suggest not to pass a `client` variable to the `cleanupSecurityManagerSubsystem` method; always use the class instance `client` variable instead to ensure that the `client` variable that arrives at `UnresolvedExpressionSecurityManagerTestCase#tearDown()` has the expected value so you can tear down the `TestLogHandlerSetupTask`

mskacelik: Oh yeah, that parameter should not be there; that is from a previous implementation using try-with-resources.

I also changed a test a bit (checking a WFLYCTL0193 instead of checking number of occurrences of start log)

yersan: Sadly, it did not resolve the Windows issue

yersan: I've been taking a look, but I have not found anything at first glance that can give me a hint about why in your case, when the log is torn down, Windows complains about the log file being in use.

The InstallationManagerBootTestCase is using the same setup to configure the logs, and there are no issues there. This needs a closer look, maybe configuring and removing the log on each test instead of at the end of the all tests ... but I'm not sure why it is complaining on Windows

mskacelik: > maybe configuring and removing the log on each test instead of at the end of the all tests ...

 Is that not what the current tests are doing? (i.e., JUnit4's `@Before` and `@After`).
 
 But it seems that outside of the Windows tests even the Linux tests failed, due to `java.io.IOException: WFLYPRT0054: Channel closed`.

> The InstallationManagerBootTestCase is using the same setup to configure the logs, and there are no issues there

Maybe since the runtime is halting, reloading, and changing from admin to non-admin mode, it has some weird testing behaviors. The same way I find weird that there is a need for the `setUpLogDirPath` and `tearLogDirPath`... (unlike the other tests).

But there is the `* new PropertyPermission("jboss.server.log.dir", "read")`  in the note.

yersan: > Is that not what the current tests are doing? (i.e., JUnit4's @Before and @After)

Yes that's right, I commented on it, but I did not realize that's what the test was already doing, so ignore my comment about that

mskacelik: I reworked the logging handling mechanism a bit + change the step order:
from: wrong expression config -> trigger boot -> verify logs -> fix config
to: wrong expression config -> trigger boot -> fix config -> verify logs

So lets see what it will do.

mskacelik: Well, at least all the other jobs passed, but the Windows job failure is rather unusual. Not really sure why it's happening and why it's not happening to other tests that are similar to this. To the best of my understanding, Windows tests are executed without a security manager, so there should be no unusual behavior in the test results...

testsuite/shared/src/main/java/org/jboss/as/test/shared/logging/TestLogHandlerSetupTask.java
```diff
@@ -94,7 +94,33 @@ public void tearDown(ModelControllerClient managementClient) throws Exception {
 
         // remove file, note this needs to be done after the operations have been executed as we need to ensure that
         // no FD's are open. This can be an issue on Windows.
-        Files.deleteIfExists(logPath);
+        // small retry/backoff loop to handle lingering OS file locks.
```
yersan: It is a good try, but this is not going to work. It seems the main issue here is that the file is still in use by the running server. You are invoking `logHandlerSetup.tearDown(client)` before the container gets stopped, so delaying the deletion here is not going to take any effect. Calling it before the container gets stopped is expected, indeed you need to connect to the server and rollback the changes. We need to understand why in other tests, the file is deleted without issues even if the server is running, and in this test, it is not.

If we try to workaround it, there are chances where we could be putting problems under the carpet that in the future will arise again, so I would avoid it.

I understand this is a test issue, and it is not related to the fix itself, so if we do not find the root cause, we could assume the error by catching the exception in your test case and creating a follow up Jira with the intention of being fixed sooner than later.

mskacelik: It was just an attempt, but yes I agree with you. Going to investigate further.

mskacelik:
I disabled the test for the Windows environment for now and created a separate issue: https://issues.redhat.com/projects/WFCORE/issues/WFCORE-7500.
I created the issue in advance so that I could include its issue key (WFCORE-7500) in the comment (`// FIXME: ...`).

I restarted the jobs, because some of them failed randomly (esp. Windows one).

In the description of WFCORE-7500, I noted that WFCORE-7335 has not yet been merged, so people should not work on it yet.

Could you take one last look at the PR, @yersan? Hopefully, no major changes are needed (if any) since you checked the PR before Christmas.

If you wanna remove/close the issue (7500) let me know, so people would not start working on it (in case this PR takes a another few days to change/fix some stuff).

testsuite/manualmode/src/test/java/org/jboss/as/test/manualmode/secman/UnresolvedExpressionSecurityManagerTestCase.java
```diff
@@ -0,0 +1,219 @@
...
+        container.start();
+        client = TestSuiteEnvironment.getModelControllerClient();
+        logHandlerSetup.setup(client);
+    }
+
+    @After
+    public void tearDown() throws Exception {
+        try {
+            if (client != null) {
+                logHandlerSetup.tearDown(client);
```
yersan, line 66: @mskacelik I played a bit with this PR on my Windows VM, and I understand now the issue with the lock in Windows. The problem seems to be that the server is arriving at this point in admin-only mode; it goes out in this mode after `cleanupSecurityManagerSubsystem`.

In such a case, the `logHandlerSetup.tearDown()` logic, although it is removing the resource, does not fully release the resources if the server is in admin-only, hence the file lock issue in Windows.

We should add a ` ServerReload.executeReloadAndWaitForCompletion(client, false)` before tearing down the log resources, so the files can be deleted without issues in Windows. It could be just before this line, or, much better, at the end of the `cleanupSecurityManagerSubsystem` method.

testsuite/manualmode/src/test/java/org/jboss/as/test/manualmode/secman/UnresolvedExpressionSecurityManagerTestCase.java
```diff
@@ -0,0 +1,219 @@
...
+    @BeforeClass
+    public static void beforeClass() {
+        AssumeTestGroupUtil.assumeNotBootableJar();
+
+        // FIXME: [WFCORE-7500] - UnresolvedExpressionSecurityManagerTestCase fails in
+        // Windows environments
+        if (TestSuiteEnvironment.isWindows()) {
+            throw new AssumptionViolatedException(
+                    "FIXME: [WFCORE-7500] - UnresolvedExpressionSecurityManagerTestCase fails in Windows environments");
+        }
```
yersan: As per the review below, this should not be necessary any longer. The issue can be resolved as we won't do or together with this PR

testsuite/manualmode/src/test/java/org/jboss/as/test/manualmode/secman/UnresolvedExpressionSecurityManagerTestCase.java
yersan: > In such a case, the `logHandlerSetup.tearDown()` logic, although it is removing the resource, does not fully release the resources if the server is in admin-only, hence the file lock issue in Windows.

Adding a note about this case along https://github.com/wildfly/wildfly-core/blob/035d41260346840f0073a90875997f2b8f16df17/testsuite/shared/src/main/java/org/jboss/as/test/shared/logging/TestLogHandlerSetupTask.java#L95-L96 lines would be helpful

mskacelik: Wow, thank you so much :), I will try to change it ASAP.

PR #6510, opened 2025-09-10
[WFCORE-7343] Fix FQDN host name calculation when HOSTNAME/COMPUTERNA…

yersan:
…ME are not set

Jira issue: https://issues.redhat.com/browse/WFCORE-7343

### DIFF
--- a/network/src/main/java/org/jboss/as/network/NetworkUtils.java
+++ b/network/src/main/java/org/jboss/as/network/NetworkUtils.java
@@ -49,11 +49,11 @@ public class NetworkUtils {
     }
 
     /**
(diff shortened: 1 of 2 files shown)


yersan:
Still a draft since I am still thinking how to add a good test case for this, but seems a bit difficult since it has to be launch in a controlled environment without setting HOSTNAME/COMPUTERNAME and even in that case, the host resolution could be be depenendent on the host configuration, so it seems quite challenging to make it determinist ... if you are reading this and has hint, that would be great

yersan:
@bstansberry Any thoughts on this? thanks!

PR #6518, opened 2025-09-17
WFCORE-7350: Elytron schema bump to 19 Community

theashiot:
https://issues.redhat.com/browse/WFCORE-7350

### DIFF
--- a/elytron/src/main/java/org/wildfly/extension/elytron/ElytronExtension.java
+++ b/elytron/src/main/java/org/wildfly/extension/elytron/ElytronExtension.java
@@ -76,8 +76,9 @@ public class ElytronExtension implements Extension {
     static final ModelVersion ELYTRON_17_0_0 = ModelVersion.create(17);
     static final ModelVersion ELYTRON_18_0_0 = ModelVersion.create(18);
     static final ModelVersion ELYTRON_19_0_0 = ModelVersion.create(19);
+    static final ModelVersion ELYTRON_20_0_0 = ModelVersion.create(20);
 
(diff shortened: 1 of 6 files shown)


rhusar:
There is very little value in the PR like this as this is not mergeable on it's own. You need a change to go along with the schema bump.

theashiot:
Thanks, @rhusar, @darranl! I've removed an unnecessary newline from my previous changes and marked this PR as ready for review.

best,
ashwin

pferraro:
> @rhusar +1 to the comment from @theashiot we have multiple developers who may work on this subsystem concurrently, the approach we use is one coordinates in chat that they will handle the schema bump and then all other changes will build upon that bump.

Topic branches?

If there are multiple developers working on branches for the same model and/or schema version, they would typically base their respective locally branches on some common branch, e.g. theashiot:WFCORE-7350.
It's fine to use the pull request system as a means of publicising your topic branch, but such a pull request should be submitted as a draft - and should never be merged independently.

darranl:
> > @theashiot At this point please don't change the SHAs of either of these commits so other PRs can depend upon them
> 
> Just to mention, git is smart enough to skip the same commit even if the SHA differs. Also, even then, people can easily rebase last commit on top of the branch with no breakage, e.g. `git rebase --onto theashiot/WFCORE-7350 HEAD~1`

FYI I also want to avoid the commit appearing twice in the history and trying to avoid other developers commits getting modified by others.

The dynamic SSL work is close to being merged so closing this one.

PR #6519, opened 2025-09-17
WFCORE-7361 Suspend/resume operation handlers can throw IllegalArgumentException if failed future has no cause

pferraro:
https://issues.redhat.com/browse/WFCORE-7361

### DIFF
--- a/server/src/main/java/org/jboss/as/server/operations/ServerResumeHandler.java
+++ b/server/src/main/java/org/jboss/as/server/operations/ServerResumeHandler.java
@@ -11,6 +11,7 @@ import static org.jboss.as.controller.descriptions.ModelDescriptionConstants.RES
 import static org.jboss.as.controller.descriptions.ModelDescriptionConstants.RUNNING_SERVER;
 
 import java.util.EnumSet;
+import java.util.Optional;
 import java.util.concurrent.CancellationException;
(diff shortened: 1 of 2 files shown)


server/src/main/java/org/jboss/as/server/operations/ServerResumeHandler.java
```diff
@@ -73,7 +74,7 @@ public void handleResult(OperationContext.ResultAction resultAction, OperationCo
                                 Thread.currentThread().interrupt();
                                 resume.toCompletableFuture().cancel(false);
                             } catch (ExecutionException e) {
-                                context.getFailureDescription().set(e.getCause().getLocalizedMessage());
+                                context.getFailureDescription().set(Optional.ofNullable(e.getCause()).orElse(e).getLocalizedMessage());
```
yersan: If the message is also null, the resulting description will not help too much. 

What about:

`context.getFailureDescription().set(Optional.ofNullable(e.getCause()).orElse(e).toString());`

To ensure that at least something helpful will be returned?

pferraro: Good point.  Will change.

yersan (approved):

PR #6521, opened 2025-09-18
[WFCORE-7362] Add CSP header to management endpoints

baranowb:
Issue: https://issues.redhat.com/browse/WFCORE-7362

### DIFF
--- a/domain-http/interface/src/main/java/org/jboss/as/domain/http/server/ContentSecurityPolicyHeaderHandler.java
+++ b/domain-http/interface/src/main/java/org/jboss/as/domain/http/server/ContentSecurityPolicyHeaderHandler.java
@@ -1,0 +1,28 @@
+/*
+ * Copyright The WildFly Authors
+ * SPDX-License-Identifier: Apache-2.0
+ */
+package org.jboss.as.domain.http.server;
+
+import io.undertow.server.HttpHandler;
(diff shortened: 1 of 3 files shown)


baranowb:
Im not sure if this should be done for Common#sendError and DomainUtil as well ?

domain-http/interface/src/main/java/org/jboss/as/domain/http/server/ContentSecurityPolicyHeaderHandler.java
```diff
@@ -0,0 +1,35 @@
...
+
+import io.undertow.server.HttpHandler;
+import io.undertow.server.HttpServerExchange;
+import io.undertow.util.Headers;
+
+public class ContentSecurityPolicyHeaderHandler implements HttpHandler {
+
+    public static final String DEFAULT_CSP_HEADER = "default-src 'self';";
+    private final HttpHandler next;
+    private String headerContent = DEFAULT_CSP_HEADER;
```
bstansberry: This field and the 2 arg constructor should be dropped and we just use the constant.

If we ever have a situation where we actually need to make this configurable we can add the logic for it then.

This class's maven module is not meant for use outside of the server kernel so we should optimize it for that use.

bstansberry (changes requested):

PR #6524, opened 2025-09-22
[WFCORE-7312] Use jboss-release to enable the assembly plugin for dis…

yersan:
…t/build modules

Jira issue: https://issues.redhat.com/browse/WFCORE-7312


yersan:
Changes were simple enough and harmless, so I went ahead and merged

PR #6525, opened 2025-09-23
[WFCORE-7366] / [WFCORE-7367] / [WFCORE-7322] Elytron Component Upgrades and Jakarta Authorization Activation

darranl:
https://issues.redhat.com/browse/WFCORE-7366
https://issues.redhat.com/browse/WFCORE-7367
https://issues.redhat.com/browse/WFCORE-7322

### DIFF
--- a/elytron/src/main/java/org/wildfly/extension/elytron/Capabilities.java
+++ b/elytron/src/main/java/org/wildfly/extension/elytron/Capabilities.java
@@ -269,7 +269,13 @@ class Capabilities {
             .Builder.of(JACC_POLICY_CAPABILITY, false, Policy.class)
             .build();
 
+    static final String JAKARTA_AUTHORIZATION_CAPABILITY = CAPABILITY_BASE + "jakarta-authorization";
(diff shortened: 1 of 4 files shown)


elytron/src/main/java/org/wildfly/extension/elytron/Capabilities.java
```diff
@@ -269,8 +269,14 @@ public void accept(final ServiceBuilder serviceBuilder) {
             .Builder.of(JACC_POLICY_CAPABILITY, false, Policy.class)
             .build();
 
+    static final String JAKARTA_AUTHORIZATION_CAPABILITY = CAPABILITY_BASE + "jakarta-authorization";
```
jamezp, line 272: Just a question. The capability names tend to be . (dot) delimited. Should we do the same here?

darranl: I think we are Ok, this class already contains 30 hyphenated capability names - I hope we are Ok to make it 31 :smile:

jamezp: lol totally fair. I was mostly just curious as I'd not seen it before.

PR #6527, opened 2025-09-27
[WFCORE-7369] Remove runtime dependency on JaccDelegatingPolicy class.

darranl:
https://issues.redhat.com/browse/WFCORE-7369


yersan:
Thanks @darranl @jamezp

PR #6532, opened 2025-10-06
[WFCORE-7372]: Feature-spec generation treats all host parameters as bound to domain mode.

ehsavoie:
Renaming host and profile address parameters to __host and __profile when they are the 1st elements of the address to avoid confusion
Jira: https://issues.redhat.com/browse/WFCORE-7372

For integration use https://github.com/wildfly/wildfly/pull/19253


yersan:
@jfdenise would it be helpful if you also review this since it is pretty related with Galleon tooling

PR #6533, opened 2025-10-07
WFLY-21013 eliminate testAddingNewResources non determinism

18chanp1:
https://issues.redhat.com/browse/WFLY-21013

YamlConfigurationExtensionTest is indeterministic because there is no guarantee of the order of operations after processOperations is run. 

The fix is to sort the list after running processOperations and then fixing the order instead.


controller/src/test/java/org/jboss/as/controller/persistence/yaml/YamlConfigurationExtensionTest.java
```diff
@@ -210,20 +211,21 @@ public void testAddingNewResources() throws URISyntaxException {
...
+        assertEquals(PathAddress.pathAddress("basic", "test"), postExtensionOps.get(0).address);
+        assertFalse(postExtensionOps.get(0).operation.hasDefined("value"));
         assertEquals(ADD, postExtensionOps.get(1).operationName);
-        assertEquals(PathAddress.pathAddress("basic", "test"), postExtensionOps.get(1).address);
-        assertFalse(postExtensionOps.get(1).operation.hasDefined("value"));
-        assertEquals(ADD, postExtensionOps.get(2).operationName);
-        assertEquals(PathAddress.pathAddress("classpath", "runtime"), postExtensionOps.get(2).address);
-        ModelNode objectMap = postExtensionOps.get(2).operation.get("complex-map").asObject();
+        assertEquals(PathAddress.pathAddress("classpath", "runtime"), postExtensionOps.get(1).address);
+        ModelNode objectMap = postExtensionOps.get(1).operation.get("complex-map").asObject();
```
yersan: Also if you want, you can improve the readability of this test. For example, by sorting the code by the index of each postExtensionOps element and in groups separated by a blank line. I mean:

```
assertEquals(ADD, postExtensionOps.get(0).operationName);
assertEquals(PathAddress.pathAddress("basic", "test"), postExtensionOps.get(0).address);
assertFalse(postExtensionOps.get(0).operation.hasDefined("value"));

assertEquals(ADD, postExtensionOps.get(1).operationName);
assertEquals(PathAddress.pathAddress("classpath", "runtime"), postExtensionOps.get(1).address);
ModelNode objectMap = postExtensionOps.get(1).operation.get("complex-map").asObject();

... then the group that is asserting on  postExtensionOps.get(2) ... and so on
```

yersan:
Thanks for contributing to WildFly Core @18chanp1 !

This Pull Request (PR) and #6534 are targeting the same issue and the same testcase, so it is preferable a single PR  instead of two.

Could you add all your changes in a single PR instead? 

We do not need two commits, but that's up to you. A single PR with a single commit is even better here since we can later easily backport it to other branches when required.

Also, notice that the commit message for each commit should [also reference a JIRA number](https://docs.wildfly.org/27/Hacking_On_WildFly.html#commit-message).

18chanp1:
Hi @yersan, thanks for the suggestions

I've squashed the commits into 1 commit, combined it into this PR, added gaps. Please let me know if there are any additional changes to be made.

Thanks

PR #6557, opened 2025-11-03
[Community] WFCORE-5744 - improve keystore certificate health check

baranowb:
Issue: https://issues.redhat.com/browse/WFCORE-5744
Proposal: https://github.com/wildfly/wildfly-proposals/pull/724
Replace: https://github.com/wildfly/wildfly-core/pull/6420

NOTE: had to redo due to conflict and weird quirk - git indicate its filed against main( OG was 29.x). Rebase cause conflicts everywhere.

### DIFF
--- a/elytron/src/main/java/org/wildfly/extension/elytron/CertificateValidity.java
+++ b/elytron/src/main/java/org/wildfly/extension/elytron/CertificateValidity.java
@@ -1,0 +1,60 @@
+/*
+ * Copyright The WildFly Authors
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+package org.wildfly.extension.elytron;
+
+import java.util.Calendar;
+import java.util.Date;
+
+enum CertificateValidity {
+    NOT_YET("not yet"), VALID("valid"), ABOUT_TO_EXPIRE("about to expire"), EXPIRED("EXPIRED");
+
+    private String token;
+
+    CertificateValidity(String string) {
+        this.token = string;
+    }
+
+    /*
+     * (non-Javadoc)
+     *
+     * @see java.lang.Enum#toString()
+     */
+    @Override
+    public String toString() {
+        return token;
+    }
+
+    static CertificateValidity getValidity(final Date notBefore, final Date notAfter) {
+        final long WEEK_IN_MS = 7 * 24 * 60 * 60 * 1000;
+        final long DAY_IN_MS = 4 * 60 * 60 * 1000;
+        final Date currentDate = Calendar.getInstance().getTime();
+        if (currentDate.before(notBefore)) {
+            return NOT_YET;
+        } else if (currentDate.after(notBefore) && currentDate.before(notAfter)) {
+            // valid or about to expire
+            // once x509 is up, switch to java.time.temporal.ChronoUnit.DAYS.between ?
+            long diff = (notAfter.getTime() - notBefore.getTime());
+            if (diff < WEEK_IN_MS) {
+                // cert is very short lived, lets settle for 1 day or just mark is as about to expire
+                if (diff < DAY_IN_MS) {
+                    return ABOUT_TO_EXPIRE;
+                } else {
+                    diff = DAY_IN_MS;
+                }
+            } else {
+                diff = WEEK_IN_MS;
+            }
+            final Date interimBoundryDate = new Date(notAfter.getTime() - diff);
+            if (currentDate.after(interimBoundryDate)) {
+                return ABOUT_TO_EXPIRE;
+            } else {
+                return VALID;
+            }
+        }
+        // all is left is above notAfter
+        return EXPIRED;
+    }
+}
--- a/elytron/src/main/java/org/wildfly/extension/elytron/CertificateValidity.java
+++ b/elytron/src/main/java/org/wildfly/extension/elytron/CertificateValidity.java
@@ -5,6 +5,8 @@
 
 package org.wildfly.extension.elytron;
 
+import java.time.Duration;
+import java.time.temporal.ChronoUnit;
 import java.util.Calendar;
 import java.util.Date;
 
@@ -27,34 +29,56 @@ enum CertificateValidity {
         return token;
     }
 
-    static CertificateValidity getValidity(final Date notBefore, final Date notAfter) {
-        final long WEEK_IN_MS = 7 * 24 * 60 * 60 * 1000;
-        final long DAY_IN_MS = 4 * 60 * 60 * 1000;
+    /**
+     * Fetch representation of health based on provided factors.
+     * @param notBefore - certificate start date
+     * @param notAfter - certificate expiration date
+     * @param expirationWaterMark - watermark expressed in minutes.
+     * @return
+     * <ul>
+     * <li>{@linkplain #NOT_YET} - if currentDate.before(notBefore)</li>
+     * <li>{@linkplain #EXPIRE} - if currentDate.after(notAfter)</li>
+     * <li>{@linkplain #ABOUT_TO_EXPIRE} - if (currentDate+waterMark).after(notAfter)</li>
+     * <li>{@linkplain #VALID} - otherwise</li>
+     * </ul>
+     */
+    static CertificateValidity getValidity(final Date notBefore, final Date notAfter, final long expirationWaterMark) {
         final Date currentDate = Calendar.getInstance().getTime();
         if (currentDate.before(notBefore)) {
             return NOT_YET;
-        } else if (currentDate.after(notBefore) && currentDate.before(notAfter)) {
-            // valid or about to expire
-            // once x509 is up, switch to java.time.temporal.ChronoUnit.DAYS.between ?
-            long diff = (notAfter.getTime() - notBefore.getTime());
-            if (diff < WEEK_IN_MS) {
-                // cert is very short lived, lets settle for 1 day or just mark is as about to expire
-                if (diff < DAY_IN_MS) {
+        } else if(currentDate.after(notBefore) && currentDate.before(notAfter)) {
+            if (expirationWaterMark > 0) {
+                //if current_date+watermark>expiration_date
+                final Date interimBoundryDate = new Date(currentDate.getTime() + Duration.of(expirationWaterMark, ChronoUnit.MINUTES).toMillis());
+                if(interimBoundryDate.after(notAfter)) {
                     return ABOUT_TO_EXPIRE;
                 } else {
-                    diff = DAY_IN_MS;
+                    return VALID;
                 }
             } else {
-                diff = WEEK_IN_MS;
+                //watermark should be n+, but just in case someone abuse it lets have some fallback
+                final long WEEK_IN_MS = Duration.of(1, ChronoUnit.WEEKS).toMillis();
+                final long DAY_IN_MS = Duration.of(1, ChronoUnit.DAYS).toMillis();
+                long diff = (notAfter.getTime() - notBefore.getTime());
+                if (diff < WEEK_IN_MS) {
+                    // cert is very short lived, lets settle for 1 day or just mark is as about to expire
+                    if (diff < DAY_IN_MS) {
+                        return ABOUT_TO_EXPIRE;
+                    } else {
+                        diff = DAY_IN_MS;
+                    }
+                } else {
+                    diff = WEEK_IN_MS;
+                }
+                final Date interimBoundryDate = new Date(notAfter.getTime() - diff);
+                if (currentDate.after(interimBoundryDate)) {
+                    return ABOUT_TO_EXPIRE;
+                } else {
+                    return VALID;
+                }
             }
-            final Date interimBoundryDate = new Date(notAfter.getTime() - diff);
-            if (currentDate.after(interimBoundryDate)) {
-                return ABOUT_TO_EXPIRE;
-            } else {
-                return VALID;
-            }
+        } else /*if(currentDate.after(notAfter))*/ {
+            return EXPIRED;
         }
-        // all is left is above notAfter
-        return EXPIRED;
     }
 }
--- a/elytron/src/main/java/org/wildfly/extension/elytron/ElytronExtension.java
+++ b/elytron/src/main/java/org/wildfly/extension/elytron/ElytronExtension.java
@@ -76,8 +76,9 @@ public class ElytronExtension implements Extension {
     static final ModelVersion ELYTRON_17_0_0 = ModelVersion.create(17);
     static final ModelVersion ELYTRON_18_0_0 = ModelVersion.create(18);
     static final ModelVersion ELYTRON_19_0_0 = ModelVersion.create(19);
+    static final ModelVersion ELYTRON_20_0_0 = ModelVersion.create(20);
 
-    private static final ModelVersion ELYTRON_CURRENT = ELYTRON_19_0_0;
+    private static final ModelVersion ELYTRON_CURRENT = ELYTRON_20_0_0;
 
     static final String ISO_8601_FORMAT = "yyyy-MM-dd'T'HH:mm:ss.SSSZ";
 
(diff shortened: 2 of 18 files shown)


elytron/src/main/java/org/wildfly/extension/elytron/ElytronExtension.java
```diff
@@ -41,9 +41,9 @@ public class ElytronExtension implements Extension {
     /**
      * The current name space used for the {@code subsystem} element
      */
-    static final String NAMESPACE_18_0 = "urn:wildfly:elytron:18.0";
+    static final String NAMESPACE_19_0 = "urn:wildfly:elytron:19.0";
```
darranl: We probably need to get rid of this, it doesn't feel correct that the Extensions is mapping a schema namespace at this level.

elytron/src/main/java/org/wildfly/extension/elytron/ElytronSubsystemSchema.java
```diff
@@ -54,21 +54,23 @@ public enum ElytronSubsystemSchema implements PersistentSubsystemSchema<ElytronS
     VERSION_17_0(17),
     VERSION_18_0(18),
     VERSION_18_0_COMMUNITY(18, Stability.COMMUNITY),
+    VERSION_19_0_COMMUNITY(19, Stability.COMMUNITY),
+    VERSION_19_0(19),
```
darranl, line 57: We should not be adding a new default stability namespace at this time - it is only the community schema that needs a bump.

elytron/src/main/java/org/wildfly/extension/elytron/TlsParser.java
```diff
@@ -85,6 +85,19 @@ class TlsParser {
             .addAttribute(FileAttributeDefinitions.RELATIVE_TO)
             ;
 
+    private PersistentResourceXMLBuilder keyStoreParser_19_0 = PersistentResourceXMLDescription.builder(PathElement.pathElement(KEY_STORE))
```
darranl: As above - now is not the time to add a new default schema.

elytron/src/main/java/org/wildfly/extension/elytron/KeyStoreService.java
```diff
@@ -112,6 +136,7 @@ static KeyStoreService createFileBasedKeyStoreService(String provider, String ty
     @Override
     public void start(StartContext startContext) throws StartException {
         try {
+            this.scheduledExecutorService = createScannerExecutorService();
```
darranl: Before we can merge we will need to align on the other work adding the executor to the Elytron subystsem itself so it can be injected.

elytron/src/main/java/org/wildfly/extension/elytron/KeyStoreService.java
```diff
@@ -206,35 +232,89 @@ private AtomicLoadKeyStore.LoadKey load(AtomicLoadKeyStore keyStore) throws Exce
...
+        //JIC
+        cancelHealthCheck(false);
+        //Check if its one-time or periodic.
+        //Schedule in executor, to have consistent log output(prefix).
+        // simple fixed rate
+        final Runnable task =() -> {
+            try {
+                checkCertificatesValidity(keyStore);
+            } catch (KeyStoreException e) {
+                ROOT_LOGGER.periodicKeyStoreCheckFailed(this.keyStoreName, e);
```
darranl, line 270: We need to be cautious with this, we don't know why we may be getting a failure but for a repeated task this is logging at WARN so could flood the logs.

If we need something logged at WARN we may be better to log at WARN once but then endure the exceptions are captured at DEBUG level.

darranl (changes requested):
Thank you, generally looking good.

FYI for schema bumps we use a common commit to avoid the situation where multiple people are working on the same bump at the same time - the bump to 19 community is being coordinated in the PR https://github.com/wildfly/wildfly-core/pull/6518 - I am not sure if Ashwin is able to wrap that one up but we have a thread in Zulip to coordinate schema related work.

We had quite a lengthy discussion in Zulip about how to interpret schema versions etc.. The conclusion was just bump the community schema and don't try and assign any additional meaning to the namespace in the schema.

We are still going to need to depend on the work to add the scheduled executor to the Elytron subsystem - the main reason being that we need to ensure it shuts down cleanly - I can possibly split that piece into it's own PR so it can be merged and used.

elytron/src/main/java/org/wildfly/extension/elytron/ElytronSubsystemSchema.java
baranowb: AFAIR original PR was before this became rule or even something that was discussed.

elytron/src/main/java/org/wildfly/extension/elytron/KeyStoreService.java
baranowb: That should happen only once per check, which is controlled by user/admin. Only scenario when this could flood is when its low or set to 0( ie, default 10s). Having it as DEBUG level defeat purpose of this change IMHO. Having it flip flop between level would require arbitrary rule to flip back as well.

What could be done:
- -1 delay - turn it off? ( would happen only on server startup)
- separate logger category which is either off or DEBUG - unless its well documented/exposed I kind of am not a fan of this, as it would defeat purpose of monitoring.

elytron/src/main/java/org/wildfly/extension/elytron/CertificateValidity.java
```diff
@@ -0,0 +1,82 @@
...
+                final Date interimBoundryDate = new Date(currentDate.getTime() + expirationWaterMark*60L*1000L);
+                if(interimBoundryDate.after(notAfter)) {
+                    return ABOUT_TO_EXPIRE;
+                } else {
+                    return VALID;
+                }
+            } else {
+                //watermark should be n+, but just in case someone abuse it lets have some fallback
+                final long WEEK_IN_MS = 7 * 24 * 60 * 60 * 1000;
+                final long DAY_IN_MS = 4 * 60 * 60 * 1000;
```
skyllarr: @baranowb Just a minor, but this is not a day but 4 hours, we should update to 24 * 60 * 60 * 1000

core-feature-pack/common/src/main/resources/modules/system/layers/base/org/wildfly/extension/elytron/main/module.xml
```diff
@@ -41,6 +41,7 @@
         <module name="org.jboss.threads"/>
         <module name="org.jboss.vfs"/>
         <module name="org.jboss.logmanager"/>
+        <module name="org.jboss.threads"/>
```
skyllarr: @baranowb Just a minor, this module is already on line 41

skyllarr (commented):
@baranowb ust a minor, this should be rebased on the the community schema bump that is here https://github.com/wildfly/wildfly-core/pull/6518

elytron/src/main/java/org/wildfly/extension/elytron/KeyStoreService.java
skyllarr: I think that since this is community stability it is okay as is now, and QE can discuss this again if they wish during the promotion to DEFAULT

elytron/src/main/java/org/wildfly/extension/elytron/CertificateValidity.java
rhusar: Please don't and as @Skyllarr points out it's also wrong. Just use Duration.of(...).toMillis().

rhusar (changes requested):

elytron/src/main/java/org/wildfly/extension/elytron/KeyStoreService.java
```diff
@@ -206,29 +226,71 @@ private AtomicLoadKeyStore.LoadKey load(AtomicLoadKeyStore keyStore) throws Exce
...
+            }
+        };
+
+        if(this.expirationCheckDelay == 0) {
+            this.certificateTaskFuture = scheduledExecutorService.schedule(task, 10, TimeUnit.SECONDS);
+        } else {
+            this.certificateTaskFuture = scheduledExecutorService.scheduleAtFixedRate(task, 0, this.expirationCheckDelay, TimeUnit.MILLISECONDS);
```
skyllarr: @baranowb EXPIRATION_CHECK_DELAY attribute is defined with `.setMeasurementUnit(MeasurementUnit.MINUTES)` , but here we are passing it with TimeUnit.MILLISECONDS , IMO here it should be TimeUnit.MINUTES

elytron/src/main/java/org/wildfly/extension/elytron/ElytronSubsystemSchema.java
```diff
@@ -193,9 +194,11 @@ private void addCredentialStoreParser(PersistentResourceXMLDescription.Persisten
 
     private void addTlsParser(PersistentResourceXMLDescription.PersistentResourceXMLBuilder builder) {
         TlsParser tlsParser = new TlsParser();
-        if (this.since(ElytronSubsystemSchema.VERSION_19_0)) {
+        if (this.since(ElytronSubsystemSchema.VERSION_19_0_COMMUNITY)) {
+            builder.addChild(tlsParser.tlsParserCommunity_19_0);
+        } else if (this.since(ElytronSubsystemSchema.VERSION_19_0)) {
             builder.addChild(tlsParser.tlsParser_19_0);
-        } else if (this.since(ElytronSubsystemSchema.VERSION_18_0_COMMUNITY) && this.enables(getDynamicClientSSLContextDefinition())) {
+        } else  if (this.since(ElytronSubsystemSchema.VERSION_18_0_COMMUNITY) && this.enables(getDynamicClientSSLContextDefinition())) {
```
rhusar: Stray unnecessary change.

baranowb:
ding.

elytron/src/main/java/org/wildfly/extension/elytron/KeyStoreService.java
darranl: Sorry, I still want us to address this- I don't agree with us logging an exception at WARN on a schedule.

If we really want to log I suggest we log a more simple message at WARN but log the exception at DEBUG.

elytron/src/main/java/org/wildfly/extension/elytron/KeyStoreDefinition.java
```diff
@@ -271,6 +295,8 @@ protected void performRuntime(OperationContext context, ModelNode operation, Res
             keyStoreService.getCredentialSourceSupplierInjector()
                     .inject(CredentialReference.getCredentialSourceSupplier(context, KeyStoreDefinition.CREDENTIAL_REFERENCE, model, serviceBuilder));
 
+            serviceBuilder.addDependency(SCHEDULED_EXECUTOR_RUNTIME_CAPABILITY.getCapabilityServiceName(), ScheduledExecutorService.class, keyStoreService.getScheduledExecutorInjector());
```
darranl, line 298: We should only inject this if the stability level is COMMUNITY or lower.

elytron/src/main/java/org/wildfly/extension/elytron/KeyStoreService.java
```diff
@@ -206,29 +228,71 @@ private AtomicLoadKeyStore.LoadKey load(AtomicLoadKeyStore keyStore) throws Exce
...
+                        default:
+                            break;
                     }
                 }
             }
         }
     }
 
+    private void scheduleCertificateHealthCheck() {
+        final ScheduledExecutorService scheduledExecutorService = executoreServiceSupplier.getValue();
```
darranl, line 260: This activation should only be happening at the COMMUNITY stability level or lower.

elytron/src/main/java/org/wildfly/extension/elytron/KeyStoreService.java
```diff
@@ -68,17 +71,28 @@ class KeyStoreService implements ModifiableKeyStoreService {
...
     private final String type;
     private final String path;
     private final String relativeTo;
     private final boolean required;
     private final String aliasFilter;
+    //delay in minutes between service start and periodical check of certificate health
+    //if set to '0', this will mean one-time off warning, as prior to RFE
+    private long expirationCheckDelay = DEFAULT_DELAY;
+    //minutes value between TTL of certificate and its expiration date, if TTL-expiration<water_mark, it will mean warning is warranted
+    private long expirationWaterMark = DEFAULT_EXPIRATION_WATERMARK;
```
darranl, line 90: I really think in addition to controlling the delay users should be able to turn this off, if this has any issues with any Keystore implementation users will be stuck with the behaviour.

elytron/src/main/java/org/wildfly/extension/elytron/KeyStoreService.java
```diff
@@ -206,29 +228,71 @@ private AtomicLoadKeyStore.LoadKey load(AtomicLoadKeyStore keyStore) throws Exce
...
+                final String alias = aliases.nextElement();
+                final Certificate certificate = keyStore.getCertificate(alias);
                 if (certificate != null && certificate instanceof X509Certificate) {
-                    try {
-                        ((X509Certificate) certificate).checkValidity();
-                    } catch (CertificateExpiredException | CertificateNotYetValidException e) {
-                        ROOT_LOGGER.certificateNotValid(alias, e);
+                    final X509Certificate xCertificate = (X509Certificate) certificate;
+                    final CertificateValidity certificateValidity = CertificateValidity.getValidity(xCertificate.getNotBefore(), xCertificate.getNotAfter(), this.expirationWaterMark);
+                    switch(certificateValidity) {
```
darranl, line 240: I think this logging should be optimised, after we have logged for a specific state I don't see we would need to log again until the certificate has actually expired.

elytron/src/main/java/org/wildfly/extension/elytron/KeyStoreService.java
```diff
@@ -206,29 +228,71 @@ private AtomicLoadKeyStore.LoadKey load(AtomicLoadKeyStore keyStore) throws Exce
...
 
-    private void checkCertificatesValidity(KeyStore keyStore) throws KeyStoreException {
+    private void checkCertificatesValidity(final KeyStore keyStore) throws KeyStoreException {
         if (ROOT_LOGGER.isEnabled(Logger.Level.WARN)) {
-            Enumeration<String> aliases = keyStore.aliases();
+            final Enumeration<String> aliases = keyStore.aliases();
             while (aliases.hasMoreElements()) {
-                String alias = aliases.nextElement();
-                Certificate certificate = keyStore.getCertificate(alias);
+                final String alias = aliases.nextElement();
```
darranl, line 235: As this is going to be a scheduled event this is where I really feel some control needs to be possible to turn it off - at this point we don't even know why this Keystore is defined, this could be backed by remote storage and the store not even being used for it's certificates.

Having a way to turn it off gives us a safe workaround should it cause anyone issues.

elytron/src/main/java/org/wildfly/extension/elytron/ElytronSubsystemTransformers.java
```diff
@@ -149,6 +151,18 @@ public void registerTransformers(SubsystemTransformerRegistration registration)
                 ELYTRON_8_0_0, ELYTRON_7_0_0, ELYTRON_6_0_0, ELYTRON_5_0_0, ELYTRON_4_0_0, ELYTRON_3_0_0, ELYTRON_2_0_0, ELYTRON_1_2_0 });
     }
 
+    private static void from20(ChainedTransformationDescriptionBuilder chainedBuilder) {
+        ResourceTransformationDescriptionBuilder builder = chainedBuilder.createBuilder(ELYTRON_20_0_0, ELYTRON_19_0_0);
+
+        builder.addChildResource(PathElement.pathElement(ElytronDescriptionConstants.KEY_STORE))
```
darranl, line 157: We can drop the transformer additions, after reviewing other changes these only make sense for the default stability level.

darranl (changes requested):
I have added some further review re how much we are logging and also about disabling the feature, I think we need to address both of those.

Also please rebase on top of https://github.com/wildfly-security-incubator/wildfly-core/tree/WFCORE-7690 

We have a few people working on community level features for WildFly 42, that branch will be the common base for all of them. 

If we can clean up the logging / make it possible to disable I think we are close to getting merged for WildFly 42 (Also community docs looks like the other missing thing).

PR #6584, opened 2025-11-27
[WFCORE-7258] Make use of ModuleIdentifierUtil.MODULE_NAME_CORRECTOR on the attributes that could require it

mskacelik:
issue: https://issues.redhat.com/browse/WFCORE-7258

Hopefully I applied correctly all the correctors.

### DIFF
--- a/elytron/src/main/java/org/wildfly/extension/elytron/ClassLoadingAttributeDefinitions.java
+++ b/elytron/src/main/java/org/wildfly/extension/elytron/ClassLoadingAttributeDefinitions.java
@@ -22,6 +22,7 @@ class ClassLoadingAttributeDefinitions {
     static final SimpleAttributeDefinition MODULE = new SimpleAttributeDefinitionBuilder(ElytronDescriptionConstants.MODULE, ModelType.STRING, true)
         .setAttributeGroup(ElytronDescriptionConstants.CLASS_LOADING)
         .setAllowExpression(false)
+        .setCorrector(ModuleIdentifierUtil.MODULE_NAME_CORRECTOR)
         .setMinSize(1)
         .setRestartAllServices()
         .build();
@@ -43,7 +44,7 @@ class ClassLoadingAttributeDefinitions {
     static ClassLoader resolveClassLoader(String module) throws ModuleLoadException {
         Module current = Module.getCallerModule();
         if (module != null && current != null) {
-            current = current.getModule(ModuleIdentifierUtil.parseCanonicalModuleIdentifier(module));
+            current = current.getModule(module);
         }
 
         return current != null ? current.getClassLoader() : ClassLoadingAttributeDefinitions.class.getClassLoader();
--- a/elytron/src/main/java/org/wildfly/extension/elytron/DirContextDefinition.java
+++ b/elytron/src/main/java/org/wildfly/extension/elytron/DirContextDefinition.java
@@ -126,6 +126,7 @@ class DirContextDefinition extends SimpleResourceDefinition {
 
     static final SimpleAttributeDefinition MODULE = new SimpleAttributeDefinitionBuilder(ElytronDescriptionConstants.MODULE, ModelType.STRING, true)
             .setAllowExpression(true)
+            .setCorrector(ModuleIdentifierUtil.MODULE_NAME_CORRECTOR)
             .setRestartAllServices()
             .build();
 
@@ -160,7 +161,7 @@ class DirContextDefinition extends SimpleResourceDefinition {
         if(moduleName != null && !moduleName.isEmpty()){
             try {
                 Module cm = Module.getCallerModule();
-                module = cm.getModule(ModuleIdentifierUtil.parseCanonicalModuleIdentifier(moduleName));
+                module = cm.getModule(moduleName);
             } catch (ModuleLoadException e) {
                 throw ElytronSubsystemMessages.ROOT_LOGGER.unableToLoadModule(moduleName, e);
             }
--- a/host-controller/src/main/java/org/jboss/as/host/controller/discovery/DiscoveryOptionResourceDefinition.java
+++ b/host-controller/src/main/java/org/jboss/as/host/controller/discovery/DiscoveryOptionResourceDefinition.java
@@ -11,6 +11,7 @@ import static org.jboss.as.host.controller.discovery.Constants.DEFAULT_MODULE;
 import org.jboss.as.controller.AttributeDefinition;
 import org.jboss.as.controller.MapAttributeDefinition;
 import org.jboss.as.controller.ModelOnlyWriteAttributeHandler;
+import org.jboss.as.controller.ModuleIdentifierUtil;
 import org.jboss.as.controller.PathElement;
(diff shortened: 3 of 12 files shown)


elytron/src/main/java/org/wildfly/extension/elytron/ClassLoadingAttributeDefinitions.java
```diff
@@ -22,6 +22,7 @@ class ClassLoadingAttributeDefinitions {
     static final SimpleAttributeDefinition MODULE = new SimpleAttributeDefinitionBuilder(ElytronDescriptionConstants.MODULE, ModelType.STRING, true)
         .setAttributeGroup(ElytronDescriptionConstants.CLASS_LOADING)
         .setAllowExpression(false)
+        .setCorrector(ModuleIdentifierUtil.MODULE_NAME_CORRECTOR)
```
yersan, line 25: Since the module will be provided now in a canonicalized form, apparently, we can now remove the canonicalization request done at https://github.com/wildfly/wildfly-core/blob/481cbde8dc6ac6a2a8b29353f6b85879db54a09d/elytron/src/main/java/org/wildfly/extension/elytron/ClassLoadingAttributeDefinitions.java#L46


The only one I have in doubt is the invocation of that method from https://github.com/wildfly/wildfly-core/blob/bb0806bacc370a49141f0c9cd9be4fb12066a0bb/elytron/src/main/java/org/wildfly/extension/elytron/JaasCustomSecurityRealmWrapper.java#L63

@darranl What's your feeling about this? Would this `configuration.get("module")` always be in a canonicalized form? I've tried by finding the usages of the enclosing method, but it returns that there are no usages at all, so not sure from where this method gets invoked.

elytron/src/main/java/org/wildfly/extension/elytron/DirContextDefinition.java
```diff
@@ -126,6 +126,7 @@ class DirContextDefinition extends SimpleResourceDefinition {
 
     static final SimpleAttributeDefinition MODULE = new SimpleAttributeDefinitionBuilder(ElytronDescriptionConstants.MODULE, ModelType.STRING, true)
             .setAllowExpression(true)
+            .setCorrector(ModuleIdentifierUtil.MODULE_NAME_CORRECTOR)
```
yersan, line 129: We can safely now remove the request to canononicalize the module at: 
https://github.com/wildfly/wildfly-core/blob/481cbde8dc6ac6a2a8b29353f6b85879db54a09d/elytron/src/main/java/org/wildfly/extension/elytron/DirContextDefinition.java#L163

This is a private method, and it will use the value of the module attribute already canonicalized

host-controller/src/main/java/org/jboss/as/host/controller/discovery/DiscoveryOptionResourceDefinition.java
```diff
@@ -44,6 +45,7 @@ public class DiscoveryOptionResourceDefinition extends SimpleResourceDefinition
     public static final SimpleAttributeDefinition MODULE = new SimpleAttributeDefinitionBuilder(ModelDescriptionConstants.MODULE, ModelType.STRING, true)
         .setDefaultValue(new ModelNode(DEFAULT_MODULE))
         .setValidator(new StringLengthValidator(1))
+        .setCorrector(ModuleIdentifierUtil.MODULE_NAME_CORRECTOR)
```
yersan, line 48: With this, we can remove https://github.com/wildfly/wildfly-core/blob/dd7a7d713f508c5e1d14f77f3978496b9e73797b/host-controller/src/main/java/org/jboss/as/host/controller/operations/DiscoveryOptionAddHandler.java#L79 and avoid canonicalizing again a value already canonicalized.

server/src/main/java/org/jboss/as/server/controller/resources/ModuleLoadingResourceDefinition.java
```diff
@@ -55,7 +56,8 @@
  */
 public class ModuleLoadingResourceDefinition extends SimpleResourceDefinition {
 
-    private static final AttributeDefinition MODULE_NAME = new SimpleAttributeDefinitionBuilder(MODULE, ModelType.STRING).build();
+    private static final AttributeDefinition MODULE_NAME = new SimpleAttributeDefinitionBuilder(MODULE,
+            ModelType.STRING).setCorrector(ModuleIdentifierUtil.MODULE_NAME_CORRECTOR).build();
```
yersan, line 59: With this in place, we can now remove the canonicalization at:

https://github.com/wildfly/wildfly-core/blob/de60453b8a5c0213c6781093e1b1b1460b8f08bd/server/src/main/java/org/jboss/as/server/controller/resources/ModuleLoadingResourceDefinition.java#L210

yersan (commented):
@mskacelik Added some comments, notice that the ultimate goal here by moving the module name corrector to the attributes, is also to avoid parsing to the canonize form twice ... it is a bit tricky, so we need to closely review this

elytron/src/main/java/org/wildfly/extension/elytron/ClassLoadingAttributeDefinitions.java
darranl: Unless I am misreading this comment this sounds like a breaking change we can not make to the management model.

Is the call to parseCanonicalModuleIdentifier more permissive than ModuleIdentifierUtil.MODULE_NAME_CORRECTOR?

darranl (changes requested):
This sounds at risk of being a breaking model change.

I think I would like to see test cases that demonstrate that working values that can be specified today will continue to work without error even where we previously relied on passing them through parseCanonicalModuleIdentifier

elytron/src/main/java/org/wildfly/extension/elytron/ClassLoadingAttributeDefinitions.java
yersan: @darranl  The `ModuleIdentifierUtil.MODULE_NAME_CORRECTOR` as an attribute corrector just moves the `#parseCanonicalModuleIdentifier(String)` to the attribute, so it will always return the canonicalized form of the module, see https://github.com/wildfly/wildfly-core/blob/b5ab64888b0df01259762f0b9d575ef2fb08b750/controller/src/main/java/org/jboss/as/controller/ModuleIdentifierUtil.java#L94-L97. 

They are equivalents, but having it at the attribute level is preferred since it is supposed to be the entry of the user values.

yersan: The idea has always been to place this sort of conversions/adjustments as close as possible to the external inputs, hence on the attribute definitions via `ModuleIdentifierUtil.MODULE_NAME_CORRECTOR`, however, for this [JaasCustomSecurityRealmWrapper.java](https://github.com/wildfly/wildfly-core/blob/bb0806bacc370a49141f0c9cd9be4fb12066a0bb/elytron/src/main/java/org/wildfly/extension/elytron/JaasCustomSecurityRealmWrapper.java) class, which is indeed deprecated, I'm not sure what the external input is for this `configuration` object

PR #6631, opened 2026-01-28
[WFCORE-7484] add-user should be able to create user credentials in a separate directory

rafaeldarosagit:
Link  Jira issue: https://issues.redhat.com/browse/WFCORE-7484


yersan (dismissed):
Hello @rafaeldarosagit , WildFly Core has its own Jira project, hence if the issue you are resolving is against wildfly core base code, it should be moved from WFLY to WFCORE.

Could you address it and change PR title and description accordingly? thanks!

rafaeldarosagit:
@yersan fixed.

yersan (commented):
Thanks for moving it @rafaeldarosagit 

This is an RFE, and as such, it should follow the standard development process based on the stability level it is targeting.

PR #6706, opened 2026-03-27
[WFCORE-7544] Fix usability issue renaming installation-manager.sh to…

yersan:
… _installation-manager_helper.sh

Jira issue: https://redhat.atlassian.net/browse/WFCORE-7544


core-feature-pack/galleon-common/src/main/resources/configs/domain/model.xml
```diff
@@ -29,5 +29,6 @@
         <!-- Unstable API Annotation Index -->
         <package name="org.wildfly._internal.unstable-api-annotation-index" optional="true" valid-for-stability="preview" />
         <package name="org.wildfly.unstable.annotation.api.indexer" optional="true" valid-for-stability="preview" />
+        <package name=".installation"/>
```
yersan, line 32: @jfdenise We are adding this new .installation directory under the standard galleon content directory, so I understand there is an auto discovery done by Galleon and a new package, named ".installation" is created out of the box.

My question is more about design than functionality. Under which circumstances should we count on this auto discovery of package names instead of creating our own dedicated package name?

I only can think of having a maybe better package name, but I am not sure

yersan:
Full Integration Linux:
https://ci.wildfly.org/buildConfiguration/WF_WildFlyCoreIntegrationExperiments/564203
Full Integration Windows:
https://ci.wildfly.org/buildConfiguration/WF_WildFlyCoreIntegrationExperimentsWindows/564234
Preview Integration Linux:
https://ci.wildfly.org/buildConfiguration/WF_WildFlyCoreIntegrationExperiments/564232

PR #6726, opened 2026-04-17
[WFCORE-7552] GC_LOG broken for Semeru JDK on Windows with space in path

lvydra:
https://redhat.atlassian.net/browse/WFCORE-7552

### DIFF
--- a/core-feature-pack/common/src/main/resources/content/bin/standalone.bat
+++ b/core-feature-pack/common/src/main/resources/content/bin/standalone.bat
@@ -272,14 +272,24 @@ if not "%PRESERVE_JAVA_OPTS%" == "true" (
             "!JAVA!" -Xverbosegclog:"!JBOSS_LOG_DIR!\gc.log" -version >nul 2>&1 && (set OPEN_J9_JDK=true) || (set OPEN_J9_JDK=false)
             if "!OPEN_J9_JDK!" == "true" (
                 set TMP_PARAM=-Xverbosegclog:"!JBOSS_LOG_DIR!\gc.log"
+                "!JAVA!" !TMP_PARAM! -version > nul 2>&1
+                if not errorlevel == 1 (
+                   set "GC_OPTS=!TMP_PARAM!"
+                ) else (
+                   set "GC_OPTS="
+                )
             ) else if "!MODULAR_JDK!" == "true" (
                 set TMP_PARAM=-Xlog:gc*:file="\"!JBOSS_LOG_DIR!\gc.log\"":time,uptimemillis:filecount=5,filesize=3M
+                "!JAVA!" !TMP_PARAM! -version > nul 2>&1
+                if not errorlevel == 1 (
+                   set "JAVA_OPTS=!JAVA_OPTS! !TMP_PARAM!"
+                )
             ) else (
                 set TMP_PARAM=-verbose:gc -Xloggc:"!JBOSS_LOG_DIR!\gc.log" -XX:+PrintGCDetails -XX:+PrintGCDateStamps -XX:+UseGCLogFileRotation -XX:NumberOfGCLogFiles=5 -XX:GCLogFileSize=3M -XX:-TraceClassUnloading
-            )
-            "!JAVA!" !TMP_PARAM! -version > nul 2>&1
-            if not errorlevel == 1 (
-               set "JAVA_OPTS=!JAVA_OPTS! !TMP_PARAM!"
+                "!JAVA!" !TMP_PARAM! -version > nul 2>&1
+                if not errorlevel == 1 (
+                   set "JAVA_OPTS=!JAVA_OPTS! !TMP_PARAM!"
+                )
             )
             rem Remove the gc.log file from the -version check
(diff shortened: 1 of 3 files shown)


yersan:
@lvydra a just quick comment without reviewing the PR changes, https://github.com/wildfly/wildfly-core/blob/main/testsuite/scripts/src/test/java/org/wildfly/scripts/test/StandaloneScriptTestCase.java#L62 is testing this, have you checked whether would it be possible to exercise this with a JBOSS_HOME containing white spaces?

lvydra:
@yersan  not yet, I will look at it.

core-feature-pack/common/src/main/resources/content/bin/standalone.bat
```diff
@@ -272,14 +272,24 @@ if not "%PRESERVE_JAVA_OPTS%" == "true" (
             "!JAVA!" -Xverbosegclog:"!JBOSS_LOG_DIR!\gc.log" -version >nul 2>&1 && (set OPEN_J9_JDK=true) || (set OPEN_J9_JDK=false)
             if "!OPEN_J9_JDK!" == "true" (
                 set TMP_PARAM=-Xverbosegclog:"!JBOSS_LOG_DIR!\gc.log"
+                "!JAVA!" !TMP_PARAM! -version > nul 2>&1
+                if not errorlevel == 1 (
+                   set "GC_OPTS=!TMP_PARAM!"
```
jamezp, line 277: Is there a reason we use the `GC_OPTS` instead of just adding to the `JAVA_OPTS` using the `!TMP_PARAM!`? I know Windows is a bit finicky about this.

jamezp (commented):
We should look to see if we can now re-enable the GC_LOGS testing on Windows given we no longer support Java 11. It might be fixed now.

core-feature-pack/common/src/main/resources/content/bin/standalone.bat
lvydra: Hi @jamezp, the combination of parentheses in path ( _(test)_ ), quotes needed for spaces, and batch re-parsing variables during expansion seems to create an unsolvable conflict when trying to merge gc log opts into JAVA_OPTS.
Keeping GC_OPTS separate and passing it directly to the Java command is the only working solution I have found.

lvydra:
@yersan @jamezp  I have removed  https://github.com/wildfly/wildfly-core/blob/285a59abd9a513f97c709f6fe688af8088ccff79/testsuite/scripts/src/test/java/org/wildfly/scripts/test/StandaloneScriptTestCase.java#L108 as it disabled GC_LOG testing on Windows for these cases. This way, I have been able to reproduce the error and also detected similar .ps1 errors.

core-feature-pack/common/src/main/resources/content/bin/standalone.bat
jamezp: Interesting because it seems to work below. However, I understand batch processes things oddly :)

jamezp (approved):
Sorry, I forgot to press submit on my review :)

This seems fine to me. The only minor concern I had was the `GC_OPT` could be set by the user for other JVM's besides Semeru, but that's like not a big deal.

core-feature-pack/common/src/main/resources/content/bin/standalone.bat
```diff
@@ -272,14 +272,24 @@ if not "%PRESERVE_JAVA_OPTS%" == "true" (
             "!JAVA!" -Xverbosegclog:"!JBOSS_LOG_DIR!\gc.log" -version >nul 2>&1 && (set OPEN_J9_JDK=true) || (set OPEN_J9_JDK=false)
             if "!OPEN_J9_JDK!" == "true" (
                 set TMP_PARAM=-Xverbosegclog:"!JBOSS_LOG_DIR!\gc.log"
```
yersan, line 274: Should not the path of JBOSS_LOG_DIR be surrounder by scaping the `"`?

I mean, the MODULAR_JDK variant uses `"\"!JBOSS_LOG_DIR!\gc.log\""`, but not the other two. It seems it  potentially could be misinterpreted by the Windows batch parser.

yersan (commented):
I have not got the chance to give it a try in a Windows machine, my only question about this is why it only fails on OpenJDK J9 and not in the other cases.

core-feature-pack/common/src/main/resources/content/bin/standalone.bat
lvydra: I did try that, but while it resolves the server starting issue, the log path is not validly propagated to the JVM. It seems that, in contrast to the MODULAR_JDK variant, it's not able to deal with `"` surrounding the log path.

You can see here the related part of the JAVA_OPTS echo:

`-Xverbosegclog:"\"E:\wildfly-38.0.1.Final (test)\standalone\log\gc.log\""`

results in:

`JVMJ9GC063E Unable to open file '"E:\wildfly-38.0.1.Final (test)\standalone\log\gc.log"' for writing.`

validly passed MODULAR_JDK variant echo output for comparison:

`-Xlog:gc*:file="\"E:\wildfly-38.0.1.Final (test)\standalone\log\gc.log\""`

PR #6730, opened 2026-04-20
[WFCORE-7569] Skip additional activation types when parsing the profiles.

darranl:
https://redhat.atlassian.net/browse/WFCORE-7569

### DIFF
--- a/model-test/src/test/java/org/jboss/as/model/test/MavenSettingsTest.java
+++ b/model-test/src/test/java/org/jboss/as/model/test/MavenSettingsTest.java
@@ -113,6 +113,33 @@ public class MavenSettingsTest {
         Assert.assertFalse(settings.getServers().containsKey("non-authenticated-repo"));
     }
 
+    @Test
+    public void testActivationTypes() throws Exception {
+        MavenSettings settings = new MavenSettings();
+        MavenSettings.parseSettingsXml(Paths.get(MavenSettingsTest.class.getResource("settings-with-various-activations.xml").toURI()), settings);
+        settings.resolveActiveSettings();
+
+        List<String> repoIds = settings.getRemoteRepositories().stream()
+                .map(MavenSettings.Repository::getId)
+                .toList();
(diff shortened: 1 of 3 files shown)


bstansberry (approved):

model-test/src/test/java/org/jboss/as/model/test/MavenSettingsTest.java
```diff
@@ -113,6 +113,33 @@ public void testServersConfig() throws Exception {
...
+    public void testActivationTypes() throws Exception {
+        MavenSettings settings = new MavenSettings();
+        MavenSettings.parseSettingsXml(Paths.get(MavenSettingsTest.class.getResource("settings-with-various-activations.xml").toURI()), settings);
+        settings.resolveActiveSettings();
+
+        List<String> repoIds = settings.getRemoteRepositories().stream()
+                .map(MavenSettings.Repository::getId)
+                .toList();
+
+        // Only profiles with activeByDefault=true should be activated
```
jbliznak, line 126: TBH this is so counter-intuitive that it is worth to be documented somehow - at least javadoc on MavenSettings?

bstansberry: Fair enough, but I don't think this should block this PR, since it doesn't introduce this behavior and fixes peoples' ability to build.

I figure the reason for this is trying to parse all the other possible activation mechanism, check the env to see if they are relevant and then maintain that as Maven evolves is too much.

jbliznak: yeah, we need this class fixed too ASAP ..

darranl: +1 this test case is only testing the other permutations to verify no other surprise permutations break builds for people.

jbliznak (approved):

PR #6732, opened 2026-04-21
[WFCORE-7563] Upgrade Elytron EE to 3.2.1.CR3

darranl:
Jira issue: https://redhat.atlassian.net/browse/WFCORE-7563

https://github.com/wildfly-security/wildfly-elytron-ee/compare/3.2.0.Final...3.2.1.CR3

This change also adds some optional module.xml dependencies. These are able to be optional as we know if these methods are called the caller had access to this module.


yersan:
@darranl FYI At integration Jobs for WildFly preview, the same set of errors failed twice, is this PR waiting for some actions to be taken in WildFly Preview?

darranl:
It is weird, those aren't tests this should influence

yersan (approved):

PR #6741, opened 2026-04-29
[WFCORE-7576] do not use JAVA_OPTS or GC_LOG options with standalone.sh version command

aogburn:
https://redhat.atlassian.net/browse/WFCORE-7576

### DIFF
--- a/core-feature-pack/common/src/main/resources/content/bin/domain.sh
+++ b/core-feature-pack/common/src/main/resources/content/bin/domain.sh
@@ -25,6 +25,7 @@ do
           ;;
       -v|-V|--version|-h|--help)
           PROCESS_CONTROLLER_JAVA_OPTS="-Xmx16m"
+          SKIP_CONF="true"
           SERVER_OPTS="$SERVER_OPTS \"$1\""
           ;;
       *)
@@ -88,12 +89,14 @@ else
 fi
 export JBOSS_HOME
 
-# Read an optional running configuration file
-if [ "x$DOMAIN_CONF" = "x" ]; then
-    DOMAIN_CONF="$DIRNAME/domain.conf"
-fi
-if [ -r "$DOMAIN_CONF" ]; then
-    . "$DOMAIN_CONF"
+# Read an optional running configuration file - skip for version/help commands
+if [ "$SKIP_CONF" != "true" ]; then
+    if [ "x$DOMAIN_CONF" = "x" ]; then
+        DOMAIN_CONF="$DIRNAME/domain.conf"
+    fi
+    if [ -r "$DOMAIN_CONF" ]; then
+        . "$DOMAIN_CONF"
+    fi
 fi
 
 # Setup the JVM
--- a/core-feature-pack/common/src/main/resources/content/bin/domain.sh
+++ b/core-feature-pack/common/src/main/resources/content/bin/domain.sh
@@ -23,6 +23,10 @@ do
           echo "ERROR: The use of -Djava.security.manager has been removed. Please use the -secmgr command line argument or SECMGR=true environment variable."
           exit 1
           ;;
+      -v|-V|--version)
+          PROCESS_CONTROLLER_JAVA_OPTS="-Xmx16m"
+          SERVER_OPTS="$SERVER_OPTS \"$1\""
+          ;;
       *)
           SERVER_OPTS="$SERVER_OPTS \"$1\""
           ;;
--- a/core-feature-pack/common/src/main/resources/content/bin/domain.sh
+++ b/core-feature-pack/common/src/main/resources/content/bin/domain.sh
@@ -23,7 +23,7 @@ do
           echo "ERROR: The use of -Djava.security.manager has been removed. Please use the -secmgr command line argument or SECMGR=true environment variable."
           exit 1
           ;;
-      -v|-V|--version)
+      -v|-V|--version|-h|--help)
           PROCESS_CONTROLLER_JAVA_OPTS="-Xmx16m"
           SERVER_OPTS="$SERVER_OPTS \"$1\""
           ;;
--- a/core-feature-pack/common/src/main/resources/content/bin/standalone.bat
+++ b/core-feature-pack/common/src/main/resources/content/bin/standalone.bat
@@ -55,6 +55,12 @@ if "%~1" == "" (
    goto READ-DEBUG-PORT
 ) else if "%~1" == "-secmgr" (
    set SECMGR=true
+) else if "%~1" == "-v" (
+   goto READ-VERSION
+) else if "%~1" == "-V" (
+   goto READ-VERSION
+) else if "%~1" == "--version" (
+   goto READ-VERSION
 )
 shift
 goto READ-ARGS
@@ -70,6 +76,12 @@ if not %DEBUG_ARG% == "" (
    shift
    goto READ-ARGS
 )
+
+:READ-VERSION
+set "JAVA_OPTS=-Xmx16m"
+set "PRESERVE_JAVA_OPTS=true"
+shift
+goto READ-ARGS
 
 :MAIN
 rem $Id$
--- a/core-feature-pack/common/src/main/resources/content/bin/standalone.bat
+++ b/core-feature-pack/common/src/main/resources/content/bin/standalone.bat
@@ -84,6 +84,7 @@ if not %DEBUG_ARG% == "" (
 :READ-VERSION
 set "JAVA_OPTS=-Xmx16m"
 set "PRESERVE_JAVA_OPTS=true"
+set SKIP_CONF=true
 shift
 goto READ-ARGS
 
@@ -111,15 +112,17 @@ if /i "%RESOLVED_JBOSS_HOME%" NEQ "%SANITIZED_JBOSS_HOME%" (
    echo.
 )
 
-rem Read an optional configuration file.
-if "x%STANDALONE_CONF%" == "x" (
-   set "STANDALONE_CONF=%DIRNAME%standalone.conf.bat"
-)
-if exist "%STANDALONE_CONF%" (
-   echo Calling "%STANDALONE_CONF%"
-   call "%STANDALONE_CONF%" %*
-) else (
-   echo Config file not found "%STANDALONE_CONF%"
+rem Read an optional configuration file - skip for version/help commands
+if not "%SKIP_CONF%" == "true" (
+   if "x%STANDALONE_CONF%" == "x" (
+      set "STANDALONE_CONF=%DIRNAME%standalone.conf.bat"
+   )
+   if exist "%STANDALONE_CONF%" (
+      echo Calling "%STANDALONE_CONF%"
+      call "%STANDALONE_CONF%" %*
+   ) else (
+      echo Config file not found "%STANDALONE_CONF%"
+   )
 )
 
 rem Sanitize JAVA_OPTS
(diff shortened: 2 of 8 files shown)


yersan:
Thanks @aogburn the change looks reasonable to me, I assume that we will need to do the same for all the script variants: .bat|.ps1, also, probably for the domain mode ones. 

Do you plan to continue with this or prefer us to take it?

aogburn:
Thanks for the feedback @yersan.  I can expand this to do the same for other scripts as you mentioned.

aogburn:
@yersan, updated for the other sh|bat|ps1 version commands.  Let me know if you see any needed changes.  Thanks!

core-feature-pack/common/src/main/resources/content/bin/domain.sh
```diff
@@ -23,6 +23,10 @@ do
           echo "ERROR: The use of -Djava.security.manager has been removed. Please use the -secmgr command line argument or SECMGR=true environment variable."
           exit 1
           ;;
+      -v|-V|--version)
+          PROCESS_CONTROLLER_JAVA_OPTS="-Xmx16m"
+          SERVER_OPTS="$SERVER_OPTS \"$1\""
```
yersan, line 29: I think it is ok since it makes the script more aligned for any future change, but `SERVER_OPTS` will never be in use since the process controller will exit as soon as it has read the version from the product conf.

Do you have any specific motivation to change it here? I could be missing something ....

core-feature-pack/common/src/main/resources/content/bin/domain.sh
```diff
@@ -23,6 +23,10 @@ do
           echo "ERROR: The use of -Djava.security.manager has been removed. Please use the -secmgr command line argument or SECMGR=true environment variable."
           exit 1
           ;;
+      -v|-V|--version)
+          PROCESS_CONTROLLER_JAVA_OPTS="-Xmx16m"
```
yersan, line 27: This also prevents to populate it from the domain.conf, but users could modify the domain.conf as they want so there is still possibilities to have side effects when we are just trying to read the version.

Maybe better to add a kind of preservate process controller java opts flag and avoid reading the domain.conf in that case?. This also applies to domain.bat

yersan (commented):
Thanks @aogburn , added two comments and want to get your feedback

Optionally, we could also add a test at testsuite/scripts/src/test/java/org/wildfly/scripts/test/ to verify the server starts with the version argument and verifies the JAVA_OPTS or PROCESS_CONTROLLER_JAVA_OPTS are not changed, but it is optional to me, I am not sure if the `ScriptTestCase` base test class is prepared for a test like this one. 

@jamezp could you also add your feedback to this PR?

core-feature-pack/common/src/main/resources/content/bin/domain.sh
aogburn: Before this added version block, the -v flag is added to the `SERVER_OPTS` at https://github.com/aogburn/wildfly-core/blob/f7fb31149d7b2ba2bb6798b97f7def1f4db3f9fb/core-feature-pack/common/src/main/resources/content/bin/domain.sh#L31 to be passed to the launched process controller along with any other `SERVER_OPTS` if they were passed in as well (but I'm not aware of a legit use case for that currently).  So my decision/motivation here was just minimizing change and to keep this the same to pass on -v and any/all other command line args through the `SERVER_OPTS` as dropping any other arguments from the command seemed to be adding another "behavior change" of the script beyond what's needed for this issue fix.  Dropping them wouldn't seem to change anything currently (nor would keeping them currently as you mentioned), but keeping them could future-proof this in case there's some reason why other arguments would still be passed and used with a version command.

We could certainly drop the extra `SERVER_OPTS` as a part of this change if you prefer.

core-feature-pack/common/src/main/resources/content/bin/domain.sh
aogburn: Yeah, I thought about that some and wasn't sure if we'd want to keep some residual capability to more forcefully change JVM options to a version command through the .conf files.  So since the .confs are still called, an option could still be forced to the version command if desired.  We could skip the .confs and the trade-off is any needed option change has to be made to the -v block JAVA_OPTS above and that may be unclear.  But any legit need to change a version command's options is likely few and far between though so we probably could skip the .confs altogether if you think that's best.

core-feature-pack/common/src/main/resources/content/bin/standalone.bat
```diff
@@ -55,6 +55,12 @@ if "%~1" == "" (
    goto READ-DEBUG-PORT
 ) else if "%~1" == "-secmgr" (
    set SECMGR=true
+) else if "%~1" == "-v" (
```
aogburn, line 58: On a further look, I realized the -h/--help command would re-use the JVM options for the same potential impact as the version commands so I should probably actually expand this logic in all the scripts to skip .conf JAVA_OPT sets for both version and help commands.

core-feature-pack/common/src/main/resources/content/bin/domain.sh
yersan: Ok thanks, since it does nothing, let's remove the line then; we do not need to add the version argument for `SERVER_OPTS`

core-feature-pack/common/src/main/resources/content/bin/domain.sh
yersan: > But any legit need to change a version command's options is likely few and far between though so we probably could skip the .confs altogether if you think that's best.

yes, I think it could help; my concern is more to avoid passing any java option via .conf files that could reset anything in the gc logs or so, for example "-Xlog:gc*:file=/path/to/gc.log" injected via .conf files would overwrite the gc logs when we are getting the version. Users, counterintuitively, could do that manipulating the conf files. A corner case, but skipping the .conf files load seems simple if `-v|-V|--version` is in use

aogburn: Thanks @yersan added a change to fully skip the confs for version and help commands.

core-feature-pack/common/src/main/resources/content/bin/standalone.bat
aogburn: Added a change to to address this for help commands as well.

core-feature-pack/common/src/main/resources/content/bin/domain.sh
aogburn: We'd actually need the -v flag in the `SERVER_OPTS` as that is how it is currently passed on to the java process that is launched by the script.  So if we altogether exclude this line, the -v flag is dropped from the launched process and it attempts a normal Jboss process launch instead.

aogburn:
> Optionally, we could also add a test at testsuite/scripts/src/test/java/org/wildfly/scripts/test/ to verify the server starts with the version argument and verifies the JAVA_OPTS or PROCESS_CONTROLLER_JAVA_OPTS are not changed

@yersan There doesn't seem to be any test checking the version command currently.  But I could add one so far for standalone.sh pretty easily as an example in `StandaloneScriptTestCase` with the newest commit here.  Would this be worth keeping and then also expanding to other bash and batch scripts?  And if so, would we care to test across all the variations of the -v,-V,--version,-h,--help flags for full coverage?

core-feature-pack/common/src/main/resources/content/bin/domain.sh
yersan: You're right, somehow I was assuming that these arguments were being passed via PROCESS_CONTROLLER_JAVA_OPTS. Let's keep it as it is then. I'm resolving this.

core-feature-pack/common/src/main/resources/content/bin/standalone.bat
yersan: Nice !, thanks

testsuite/scripts/src/test/java/org/wildfly/scripts/test/StandaloneScriptTestCase.java
```diff
@@ -98,6 +98,21 @@ public void testWFCORE5917InWindows() throws Exception {
...
+        ScriptProcess script = new ScriptProcess(ServerHelper.JBOSS_HOME, STANDALONE_BASE_NAME, Shell.BASH, ServerHelper.TIMEOUT);
+        // Test WFCORE-7576
+        script.start(null, Map.of("GC_LOG", "true"), new String [] {"-v"});
+
+        Assert.assertNotNull("The process is null and may have failed to start.", script);
+        validateProcess(script);
+        final var stdout = script.getStdoutAsString();
+        Assert.assertTrue("Expected to find -Xmx16m in the JVM parameters for a version command.", stdout.contains("-Xmx16m\n"));
+        Assert.assertFalse("Did not expect to find gc.log in the JVM parameters.", stdout.contains("gc.log"));
+    }
```
yersan, line 114: Sadly, this is not going to work. We cannot re-use this StandaloneScriptTest, its `testScript` is already messed up testing different cases and the use case to test the -v, -h and friends does not fit well here out of the box.

The ScriptTestCase method  is annotated with @BeforeClass invokes ServerConfigurator.configure(), which is expecting to have a fully started server to configure the log and other stuff, but our use case starting the server with -v or -h will never start a running server, it will just start the server to get some information, so the serve never will never be configured and the test will fail. It passes now because of the order of executions of the other tests.

I look closely at how these tests behave and sketched one as a guide about what I think we will need to do:

```
@RunWith(Parameterized.class)
public class StandaloneHelpVersionScriptTestCase extends ScriptTestCase {
    private static final String STANDALONE_BASE_NAME = "standalone";

    // skip execution of @BeforeClass parent, needs to make the field public at ServerConfigurator
    static {
        ServerConfigurator.CONFIGURED.set(true);
    }

    @Parameter
    public String arg;

    public StandaloneHelpVersionScriptTestCase() {
        super(STANDALONE_BASE_NAME);
    }

    @Parameters
    public static Collection<Object> data() {
        return List.of("-v", "-h"); /// add the remainder ones 
    }

    @Override
        script.start(null, Map.of(), new String [] {arg});
        final var stdout = script.getStdoutAsString();
        Assert.assertTrue("Expected to find -Xmx16m in the JVM parameters for a version command.", stdout.contains("-Xmx16m\n"));
    }
}
```
I've not checked the domain mode variant, but I guess it will be a similar approach. The idea will be to create those two: StandaloneHelpVersionScriptTestCase and DomainHelpVersionScriptTestCase.

The tests will be parametrized with all the arguments modified in this PR (-v, -V, --version, -h ... and so on). The ScriptTestCase will take care of running each test on each shell if it is available in the current environment.

These scripts are tricky, and having a test case covering them in all the shells helps a lot since we will be exercising them across Windows and Linux when they are running in our CI.

I hope this can guide you, fell free to open a new Jira if you want someone else to take care of the test case. 

Without manually testing your changes across all the different environments, my feeling is the issue is already well done and fixed, adding the test will allow us to remove this manual testing, but we can take care of it if that's your preference (I don't know how complex the domain one will be since I have not checked it)

aogburn: Thanks @yersan for the feedback and suggestion.  I'll be out next week but I can follow your direction here take a stab at the separate *HelpVersionScriptTestCase classes after that.

yersan: @aogburn do you want us to finish this issue? thanks for all the work on this so far

yersan: @aogburn We would like to deliver this this week, so I took the liberty to move on with the pending test case and created #6819 

Your fix looks ok, and the test case verified it on Linux, let's wait for the Windows variants

PR #6756, opened 2026-05-12
[WFCORE-6934] Server startup marker file is not a reliable method to detect when the server is running (POC)

elguardian:
modify the logic to include the pid of the server and check some possibilities.

### DIFF
--- a/process-controller/src/main/java/org/jboss/as/process/Main.java
+++ b/process-controller/src/main/java/org/jboss/as/process/Main.java
@@ -207,7 +207,7 @@ public final class Main {
 
         final ProcessController processController = new ProcessController(configuration, System.out, System.err);
 
-        final FileChannel lockChannel = acquireRunningLock(jbossHome);
+        processController.setRunningLockChannel(acquireRunningLock(jbossHome));
 
         final InetSocketAddress boundAddress = processController.getServer().getBoundAddress();
 
@@ -243,7 +243,6 @@ public final class Main {
         final Thread shutdownThread = new Thread(new Runnable() {
             public void run() {
                 processController.shutdown();
-                releaseRunningLock(lockChannel);
             }
         }, "Shutdown thread");
         shutdownThread.setDaemon(false);
@@ -267,16 +266,6 @@ public final class Main {
             // ignore
         }
         return null;
-    }
-
-    private static void releaseRunningLock(FileChannel channel) {
-        if (channel != null) {
-            try {
-                channel.close();
-            } catch (IOException e) {
-                // ignore
-            }
-        }
     }
 
     private static boolean isJavaSecurityManagerConfigured(final String arg) {
--- a/process-controller/src/main/java/org/jboss/as/process/Main.java
+++ b/process-controller/src/main/java/org/jboss/as/process/Main.java
@@ -11,6 +11,11 @@ import java.io.File;
 import java.io.IOException;
 import java.net.InetAddress;
 import java.net.InetSocketAddress;
+import java.nio.channels.FileChannel;
+import java.nio.channels.FileLock;
+import java.nio.file.Path;
+import java.nio.file.Paths;
+import java.nio.file.StandardOpenOption;
 import java.security.PrivilegedAction;
 import java.util.ArrayList;
 import java.util.Collections;
@@ -201,6 +206,9 @@ public final class Main {
         configuration.setReadExecutor(Executors.newCachedThreadPool(threadFactory));
 
         final ProcessController processController = new ProcessController(configuration, System.out, System.err);
+
+        final FileChannel lockChannel = acquireRunningLock(jbossHome);
+
         final InetSocketAddress boundAddress = processController.getServer().getBoundAddress();
 
         final List<String> initialCommand = new ArrayList<String>();
@@ -235,12 +243,40 @@ public final class Main {
         final Thread shutdownThread = new Thread(new Runnable() {
             public void run() {
                 processController.shutdown();
+                releaseRunningLock(lockChannel);
             }
         }, "Shutdown thread");
         shutdownThread.setDaemon(false);
         Runtime.getRuntime().addShutdownHook(shutdownThread);
 
         return processController;
+    }
+
+    private static FileChannel acquireRunningLock(String jbossHome) {
+        try {
+            Path lockPath = Paths.get(jbossHome, ".installation", "running.lock");
+            java.nio.file.Files.createDirectories(lockPath.getParent());
+            FileChannel channel = FileChannel.open(lockPath,
+                    StandardOpenOption.CREATE, StandardOpenOption.WRITE, StandardOpenOption.READ);
+            FileLock lock = channel.tryLock();
+            if (lock != null) {
+                return channel;
+            }
+            channel.close();
+        } catch (IOException e) {
+            // ignore
+        }
+        return null;
+    }
+
+    private static void releaseRunningLock(FileChannel channel) {
+        if (channel != null) {
+            try {
+                channel.close();
+            } catch (IOException e) {
+                // ignore
+            }
+        }
     }
 
     private static boolean isJavaSecurityManagerConfigured(final String arg) {
--- a/process-controller/src/main/java/org/jboss/as/process/Main.java
+++ b/process-controller/src/main/java/org/jboss/as/process/Main.java
@@ -11,11 +11,6 @@ import java.io.File;
 import java.io.IOException;
 import java.net.InetAddress;
 import java.net.InetSocketAddress;
-import java.nio.channels.FileChannel;
-import java.nio.channels.FileLock;
-import java.nio.file.Path;
-import java.nio.file.Paths;
-import java.nio.file.StandardOpenOption;
 import java.security.PrivilegedAction;
 import java.util.ArrayList;
 import java.util.Collections;
@@ -207,7 +202,7 @@ public final class Main {
 
         final ProcessController processController = new ProcessController(configuration, System.out, System.err);
 
-        processController.setRunningLockChannel(acquireRunningLock(jbossHome));
+        processController.acquireRunningLock(jbossHome);
 
         final InetSocketAddress boundAddress = processController.getServer().getBoundAddress();
 
@@ -249,23 +244,6 @@ public final class Main {
         Runtime.getRuntime().addShutdownHook(shutdownThread);
 
         return processController;
-    }
-
-    private static FileChannel acquireRunningLock(String jbossHome) {
-        try {
-            Path lockPath = Paths.get(jbossHome, ".installation", "running.lock");
-            java.nio.file.Files.createDirectories(lockPath.getParent());
-            FileChannel channel = FileChannel.open(lockPath,
-                    StandardOpenOption.CREATE, StandardOpenOption.WRITE, StandardOpenOption.READ);
-            FileLock lock = channel.tryLock();
-            if (lock != null) {
-                return channel;
-            }
-            channel.close();
-        } catch (IOException e) {
-            // ignore
-        }
-        return null;
     }
 
     private static boolean isJavaSecurityManagerConfigured(final String arg) {
--- a/process-controller/src/main/java/org/jboss/as/process/ProcessController.java
+++ b/process-controller/src/main/java/org/jboss/as/process/ProcessController.java
@@ -9,6 +9,7 @@ import java.io.IOException;
 import java.io.InputStream;
 import java.io.OutputStream;
 import java.io.PrintStream;
+import java.nio.channels.FileChannel;
 import java.nio.charset.StandardCharsets;
 import java.security.SecureRandom;
 import java.util.Arrays;
@@ -46,6 +47,7 @@ public final class ProcessController {
     private final Set<Connection> managedConnections = new CopyOnWriteArraySet<Connection>();
 
     private volatile boolean shutdown;
+    private volatile FileChannel runningLockChannel;
 
     private static final short AUTH_BYTES_LENGTH = 16;
     public static final short AUTH_BYTES_ENCODED_LENGTH = 24;
@@ -224,7 +226,20 @@ public final class ProcessController {
                 }
             }
             ProcessLogger.ROOT_LOGGER.shutdownComplete();
+            final FileChannel channel = runningLockChannel;
+            if (channel != null) {
+                runningLockChannel = null;
+                try {
+                    channel.close();
+                } catch (IOException e) {
+                    // ignore
+                }
+            }
         }
+    }
+
+    void setRunningLockChannel(FileChannel channel) {
+        this.runningLockChannel = channel;
     }
 
     public ManagedProcess getServerByAuthCode(final byte[] code) {
--- a/process-controller/src/main/java/org/jboss/as/process/ProcessController.java
+++ b/process-controller/src/main/java/org/jboss/as/process/ProcessController.java
@@ -10,7 +10,11 @@ import java.io.InputStream;
 import java.io.OutputStream;
 import java.io.PrintStream;
 import java.nio.channels.FileChannel;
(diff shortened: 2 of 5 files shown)


yersan:
@elguardian This PR adds some additional checks to ensure the PID is written in the marker file and the marker file is handled properly. However, the main problem with the issue is what happens when the JVM is killed or crashed. 

Also, my understanding is that checking the PID will not help to identify which process wrote it. Notice that the installation manager tool would be running in a complete different JVM, even with a different user than the server.

Our initial conclusion was that using marker files will not be a reliable method to detect whether the server is running or not, even if a shutdown hook is added, that shutdown hook won't be invoked at all when the server receives a SIGKILL, which is indeed the specific case that is triggering this.

yersan:
Well, I've just realized you also added some comments on the Jira I have not processed yet, so maybe what I said before is redundant. Let's continue any discussion in the Jira and not in the PR.

Moved the PR to draft to state it is in progress

elguardian:
@yersan  I did simplify this to hold the file lock during the lifecycle. it should be more align with your thoughts... Regarding IM still pending how to handle that one. In the end the file is in standalone/tmp/<marker> or domain/tmp<marker> that is something we can know when we instantiate the IM. as for the cli we can always add add a parameter to determine the tmp folder
wdyt ?

elguardian:
retest this please

yersan:
Sorry @elguardian I missed your latest notification. I'll take a closer look at this...

server/src/main/java/org/jboss/as/server/BootstrapListener.java
```diff
@@ -116,29 +127,41 @@ private void done(final long bootstrapTime, final StabilityStatistics statistics
...
     }
 
-    public static void deleteStartupMarker(File tempDir) {
-        File file = new File(tempDir, BootstrapListener.MARKER_FILE);
+    public void releaseStartupMarker() {
         try {
-            Files.deleteIfExists(file.toPath());
+            if (markerFileLock != null && markerFileLock.isValid()) {
+                markerFileLock.release();
+            }
```
yersan: We should not change this behaviour, I don't know which other stuff outside of the server is relaying on this marker and we should continue deleting the marker if possible.

host-controller/src/main/java/org/jboss/as/host/controller/HostControllerService.java
```diff
@@ -161,8 +162,8 @@ public void start(StartContext context) throws StartException {
 
         final ServiceContainer serviceContainer = myController.getServiceContainer();
 
-        final BootstrapListener bootstrapListener = new BootstrapListener(serviceContainer, startupTime, serviceTarget, futureContainer, prettyVersion + " (Host Controller)", environment.getDomainTempDir());
-        bootstrapListener.getStabilityMonitor().addController(myController);
+        this.bootstrapListener = new BootstrapListener(serviceContainer, startupTime, serviceTarget, futureContainer, prettyVersion + " (Host Controller)", environment.getDomainTempDir());
+        this.bootstrapListener.getStabilityMonitor().addController(myController);
```
yersan: I would say this is not going to work as we expect. In domain mode, we have to look up the Process Controller; the process controller should be the one that takes the lock, here the lock is being held by the Host Controller. On weird situations, the Host Controller could die but the managed servers could be still running, we want to release the lock only when the Process Controller finishes.

server/src/main/java/org/jboss/as/server/BootstrapListener.java
```diff
@@ -116,29 +127,41 @@ private void done(final long bootstrapTime, final StabilityStatistics statistics
...
             // ignore
         }
+    }
 
+    private void openMarkerFile(File file, String content) throws IOException {
+        Files.deleteIfExists(file.toPath());
+        markerFileChannel = FileChannel.open(file.toPath(),
+                StandardOpenOption.CREATE_NEW, StandardOpenOption.WRITE, StandardOpenOption.READ);
+        markerFileChannel.write(ByteBuffer.wrap(content.getBytes(StandardCharsets.UTF_8)));
+        markerFileLock = markerFileChannel.lock();
```
yersan: This is a blocking operation, we should use tryLock(). We have to paid attention to what happens with the embedded server. The embedded server must not acquire the lock

yersan (commented):
Hi @elguardian changes could be simplified, we should not change the current logic done by the bootstrap listener, I commented about the reason why I think it. In addition, we should not use the Host Controller.

We even don't need to create a marker file in the server JVM or write data on it, indeed creating the marker file from the server JVM opens the door to handle what happens if the file is not created, data are not written or any other issue. I believe we can avoid its creation, and also, avoid using the temp dir. That will remove the requirement to tell to IM (Installation Manager) where the marker is.

I guess the main problem here is where we should take the lock, and decide what to do if there is any problem acquiring it; that needs some thought.

My suggestions:
1. We can create the file for the lock when the server is provisioned; we can instruct galleon to create an empty file as if were part of the server files, for example at WFLY_HOME/.installation/running.lock. That will be a stable, fixed path; no need to tell to the IM where the file is, we don't need to deal with what happen if the file is not deleted, is not created and so on. It will be provisioned when the server is installed or updated.
2. Acquire an advisory lock, where?
2.1 For standalone server, it could be in the ApplicationServerService, that's the server root service for standalone, it is shutdown when a reload, but I think that's ok. Acquire the lock on the start(), release it on stop(). That service knows the location of the home dir, so it can access to WFLY_HOME/.installation/running.lock
2.2 On Domain Mode, that one is more tricky, I think it could be below https://github.com/wildfly/wildfly-core/blob/035d41260346840f0073a90875997f2b8f16df17/process-controller/src/main/java/org/jboss/as/process/protocol/ProtocolServer.java#L60 line. We would need to pass the jboss home. That thread is the thread that keeps the Process Controller socket.

IM only have to try to acquire a lock for the same file WFLY_HOME/.installation/running.lock. IM needs the server home, so there is no need to send information from the server to IM. If the lock is already held by another process, it should not apply the candidate server, otherwise it can proceed.

We would need to add a test case to verify how it works on Windows/Linux.

Something to check is if the embedded server acquires the lock, if so, we have to think how we can skip that acquisition when the embedded server starts. An embedded Host Controller will not acquire it because we are placing the logic on the Process Controller, that's good

yersan:
hold on @elguardian ... this is turning into a fragile approach ....I wonder if we can overcome it by using a marker handled by the launch scripts ... I'm going to open a wider discussion in Zulip about this

yersan:
See https://wildfly.zulipchat.com/#narrow/channel/174184-wildfly-developers/topic/Detecting.20running.20server/near/596684370

process-controller/src/main/java/org/jboss/as/process/Main.java
```diff
@@ -243,6 +252,33 @@ public void run() {
         return processController;
     }
 
+    private static FileChannel acquireRunningLock(String jbossHome) {
+        try {
+            Path lockPath = Paths.get(jbossHome, ".installation", "running.lock");
+            java.nio.file.Files.createDirectories(lockPath.getParent());
```
yersan: Why creating the file and the directory? 

It would be better to assume the `.installation/running.lock` will be always there if we create it with galleon when the server is provisioned. If we assume the file is always created at provisioning time, we don't want to worry about creating it here or dealing with any potential IO error thrown in this line. That will make the behaviour less prone to errors. I would encourage you to avoid creating the file if that file can be part of the server provisioning to reduce possible points of failures.

process-controller/src/main/java/org/jboss/as/process/Main.java
```diff
@@ -243,6 +252,33 @@ public void run() {
...
+            FileChannel channel = FileChannel.open(lockPath,
+                    StandardOpenOption.CREATE, StandardOpenOption.WRITE, StandardOpenOption.READ);
+            FileLock lock = channel.tryLock();
+            if (lock != null) {
+                return channel;
+            }
```
yersan: We have to decide what we want to do in this case. 

If the lock is not acquired, what does it mean? Do we want to start the server as usual? (I would say yes) If so, Prospero should not try to upgrade the server since we will never know whether the lock was legitimately acquired to signal that the server is running. 

We need to sort out this case and control it. If the server does not acquired the lock, by whatever reason, how Prospero would understand whether the server is still running?

It seems to me we need a fallback for this case. That fallback case can be the old which uses temp directory, although moving it out of the temp directory. How to indicate to Prospero whether the server has legitimately acquired the log is still open. As first glance it could be a flag on the Installation Manager service that is written down into the properties file used to communicate to Prospero any context for the apply operation.

process-controller/src/main/java/org/jboss/as/process/Main.java
```diff
@@ -201,6 +206,9 @@ public JBossThreadFactory run() {
         configuration.setReadExecutor(Executors.newCachedThreadPool(threadFactory));
 
         final ProcessController processController = new ProcessController(configuration, System.out, System.err);
+
+        processController.setRunningLockChannel(acquireRunningLock(jbossHome));
```
yersan: Why the indirection if acquireRunningLock is private in main? Would it make more sense to hide the acquireRunningLock login inside of the ProcessController?

process-controller/src/main/java/org/jboss/as/process/Main.java
elguardian: this just is for a test.... jvm has some sort of reentry for locks so it does not allow to acquire the lock twice (make sense), so if the lock is already taken... it should be by this jvm.

process-controller/src/main/java/org/jboss/as/process/Main.java
elguardian: moved to PC

process-controller/src/main/java/org/jboss/as/process/Main.java
elguardian: removed

process-controller/src/main/java/org/jboss/as/process/Main.java
yersan: ok, yes, we should not have other JVMs trying to get a lock ther .... just keep an eye on the embedded server / embedded Host Controller, I guess they won't start from main methods .. but just in case

elguardian:
retest this please

process-controller/src/main/java/org/jboss/as/process/ProcessController.java
```diff
@@ -224,6 +230,31 @@ public void shutdown() {
                 }
             }
             ProcessLogger.ROOT_LOGGER.shutdownComplete();
+            final FileChannel channel = runningLockChannel;
+            if (channel != null) {
+                runningLockChannel = null;
+                try {
+                    channel.close();
+                } catch (IOException e) {
+                    // ignore
```
yersan: Do not silently ignore them, it could help on debugging, could we at least add a debug trace?

`ProcessLogger.ROOT_LOGGER.debugf("Failed to close process controller lock channel");`

Something like this.

process-controller/src/main/java/org/jboss/as/process/ProcessController.java
```diff
@@ -224,6 +230,31 @@ public void shutdown() {
...
+            FileChannel channel = FileChannel.open(lockPath,
+                    StandardOpenOption.CREATE, StandardOpenOption.WRITE, StandardOpenOption.READ);
+            FileLock lock = channel.tryLock();
+            if (lock != null) {
+                this.runningLockChannel = channel;
+            } else {
+                channel.close();
+            }
+        } catch (IOException e) {
+            // ignore
```
yersan: Do not silently ignore them, it could help on debugging. Write at least a debug log

server/src/main/java/org/jboss/as/server/BootstrapListener.java
```diff
@@ -142,6 +154,46 @@ public static void deleteStartupMarker(File tempDir) {
...
+            channel.close();
+        }
+    }
+
+    public void releaseRunningLock() {
+        if (lockFileChannel != null) {
+            try {
+                lockFileChannel.close();
+            } catch (IOException e) {
+                // ignore
```
yersan: Do not silently ignore it

process-controller/src/main/java/org/jboss/as/process/ProcessController.java
```diff
@@ -224,6 +230,31 @@ public void shutdown() {
...
+                }
+            }
+        }
+    }
+
+    void acquireRunningLock(String jbossHome) {
+        try {
+            Path lockPath = Paths.get(jbossHome, ".installation", "running.lock");
+            FileChannel channel = FileChannel.open(lockPath,
+                    StandardOpenOption.CREATE, StandardOpenOption.WRITE, StandardOpenOption.READ);
```
yersan: Get the rid off CREATE option. The file can be provisioned by Galleon, there is no need to run into the problem of not creating it, this will reduce the problem surface and will make Prospero to assume the file is there

server/src/main/java/org/jboss/as/server/BootstrapListener.java
```diff
@@ -36,8 +47,10 @@ public final class BootstrapListener {
     private final String prettyVersion;
     private final FutureServiceContainer futureContainer;
     private final File tempDir;
-    private  String startedCleanMessage;
-    private  String startedWitErrorsMessage;
+    private String startedCleanMessage;
+    private String startedWitErrorsMessage;
+    private volatile FileChannel lockFileChannel;
```
yersan: There is still an unnecessary indirection between the BootstrapListener and the ApplicationServerService for this. 

From my point if view, changes could be simplified if all the logic to acquire / release the lock is in the ApplicationServerService. Right now the ApplicationServerService uses the BootstrapListener for acquire/releasing the lock on each server start/reload and looks unnecessary.

This is still a risk since during the reload, the advisory lock will be released, meanwhile the server is still started.

server/src/main/java/org/jboss/as/server/BootstrapListener.java
```diff
@@ -142,6 +154,46 @@ public static void deleteStartupMarker(File tempDir) {
...
+        try {
+            if (WildFlySecurityManager.isChecking()) {
+                doPrivileged((PrivilegedExceptionAction<Void>) () -> {
+                    openLockFile(lockPath);
+                    return null;
+                });
+            } else {
+                openLockFile(lockPath);
+            }
+        } catch (PrivilegedActionException | IOException e) {
```
yersan: We no longer use security manager, so we do not need to check for permissions.

process-controller/src/main/java/org/jboss/as/process/ProcessController.java
```diff
@@ -224,6 +230,31 @@ public void shutdown() {
...
+        try {
+            Path lockPath = Paths.get(jbossHome, ".installation", "running.lock");
+            FileChannel channel = FileChannel.open(lockPath,
+                    StandardOpenOption.WRITE, StandardOpenOption.READ);
+            FileLock lock = channel.tryLock();
+            if (lock != null) {
+                this.runningLockChannel = channel;
+            } else {
+                channel.close();
+            }
```
yersan: Fall back method notification to prospero is still missing

process-controller/src/main/java/org/jboss/as/process/ProcessController.java
```diff
@@ -224,6 +230,31 @@ public void shutdown() {
...
+    void acquireRunningLock(String jbossHome) {
+        try {
+            Path lockPath = Paths.get(jbossHome, ".installation", "running.lock");
+            FileChannel channel = FileChannel.open(lockPath,
+                    StandardOpenOption.WRITE, StandardOpenOption.READ);
+            FileLock lock = channel.tryLock();
+            if (lock != null) {
+                this.runningLockChannel = channel;
+            } else {
+                channel.close();
```
yersan: This is also a Failure to acquire process controller running lock

PR #6703, opened 2026-03-27
[WFCORE-7541] Upgrade log4j2 from 2.23.1 to 2.25.3

yersan:
Jira issue: https://redhat.atlassian.net/browse/WFCORE-7541

PR #6773, opened 2026-05-19
[WFCORE-7541] upgrade log4j2 from 2.23.1 to 2.25.4

istudens:
This PR supersedes https://github.com/wildfly/wildfly-core/pull/6703.

Jira issue: 
https://redhat.atlassian.net/browse/WFCORE-7541

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -255,7 +255,7 @@
         <version.org.apache.kerby>2.0.3</version.org.apache.kerby>
         <version.org.apache.httpcomponents.httpclient>4.5.14</version.org.apache.httpcomponents.httpclient>
         <version.org.apache.httpcomponents.httpcore>4.4.16</version.org.apache.httpcomponents.httpcore>
-        <version.org.apache.logging.log4j>2.23.1</version.org.apache.logging.log4j>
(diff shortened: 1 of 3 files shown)


istudens:
Looking at the failures.

yersan:
@jamezp Could you also take a look? it is a minor jump so not sure if we want it for `32.x` branch too

yersan:
@istudens I think some could be flaky tests; the bootable JAR and Galleon ones are old friends, the one with Security  Manager seems relevant at first glance

darranl (changes requested):
If we are going to list CVEs please make it clear that these are only used in dependencies used by the testsuite and not the WildFly distribution itself which does not ship these components.

[#wildfly-developers > Current Log4j CVE Reports](https://wildfly.zulipchat.com/#narrow/channel/174184-wildfly-developers/topic/Current.20Log4j.20CVE.20Reports/with/596415750)

jamezp:
Yes, these CVE's do not affect WildFly. We only ship the log4j-api, which is not affected.

bstansberry:
Please see the note I added to the WFCORE-7541 issue description and modify if necessary.

yersan (approved):

PR #6778, opened 2026-05-27
WFCORE-7602 Required attributes with alternatives should be treated as optional during XML parsing.

pferraro:
https://redhat.atlassian.net/browse/WFCORE-7602


pferraro:
@yersan FYI - this fix is blocking a series of WFLY issues on which I am working.

yersan (approved):

PR #6782, opened 2026-05-28
WFCORE-7605 XMLAll/Choice/Sequence should be optional, regardless of cardinality, if all children are optional.

pferraro:
https://redhat.atlassian.net/browse/WFCORE-7605


pferraro:
@yersan FYI - this fix is blocking a series of WFLY issues on which I am working.

yersan:
It got stuck due to some details with the CI, but seems now ready to review

yersan (approved):

yersan:
Thanks @pferraro

PR #6787, opened 2026-06-01
[WFCORE-7607] Upgrade commons-daemon to 1.6.0

lvydra:
https://redhat.atlassian.net/browse/WFCORE-7607

Verified that WildFly standalone / WildFly domain can be installed as a Windows Service and started/stopped successfully with commons-daemon upgraded to 1.6.0


yersan:
Thanks @lvydra  ... lately we had some issues in the CI that prevents Windows jobs to run, so the current failures should be unrelated so forget about them, in any case, I prefer to kick them off again before merging

yersan (approved):

yersan:
Thanks @lvydra

PR #6788, opened 2026-06-03
[WFCORE-7611] Remote Outbound Connection HeartBeat settings not working

elguardian:
Issue: https://redhat.atlassian.net/browse/WFCORE-7611

### DIFF
--- a/remoting/subsystem/src/main/java/org/jboss/as/remoting/ConnectionInfo.java
+++ b/remoting/subsystem/src/main/java/org/jboss/as/remoting/ConnectionInfo.java
@@ -1,0 +1,24 @@
+/*
+ * Copyright The WildFly Authors
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+package org.jboss.as.remoting;
+
+import java.net.URI;
+import org.wildfly.service.descriptor.NullaryServiceDescriptor;
+import org.xnio.OptionMap;
+
+/**
+ * Provides connection information for an outbound remoting connection.
+ */
+public interface ConnectionInfo {
+
+    NullaryServiceDescriptor<ConnectionInfo> CONNECTION_INFO_CAPABILITY = NullaryServiceDescriptor.of("org.wildfly.remoting.connection-info", ConnectionInfo.class);
+
+    URI getDestinationUri();
+
+    String getUsername();
+
+    OptionMap getConnectionCreationOptions();
+}
--- a/remoting/subsystem/src/main/java/org/jboss/as/remoting/EndpointService.java
+++ b/remoting/subsystem/src/main/java/org/jboss/as/remoting/EndpointService.java
@@ -40,7 +40,7 @@ public class EndpointService implements Service {
                            String nodeName, EndpointType type, final OptionMap optionMap) {
         this.endpointConsumer = endpointConsumer;
         this.workerSupplier = workerSupplier;
-        this.connectionInfoSupplier = connectionInfoSupplier;
+        this.connectionInfoSupplier = connectionInfoSupplier != null ? connectionInfoSupplier : () -> null;
         if (nodeName == null) {
             nodeName = "remote";
         }
--- a/remoting/subsystem/src/main/java/org/jboss/as/remoting/EndpointService.java
+++ b/remoting/subsystem/src/main/java/org/jboss/as/remoting/EndpointService.java
@@ -16,6 +16,7 @@ import org.jboss.msc.service.StartException;
 import org.jboss.msc.service.StopContext;
 import org.jboss.remoting3.Endpoint;
 import org.jboss.remoting3.EndpointBuilder;
+import org.jboss.remoting3.RemotingOptions;
 import org.xnio.OptionMap;
 import org.xnio.XnioWorker;
 
@@ -32,10 +33,14 @@ public class EndpointService implements Service {
     protected final OptionMap optionMap;
     private final Consumer<Endpoint> endpointConsumer;
     private final Supplier<XnioWorker> workerSupplier;
+    private final Supplier<ConnectionInfo> connectionInfoSupplier;
 
-    public EndpointService(final Consumer<Endpoint> endpointConsumer, final Supplier<XnioWorker> workerSupplier, String nodeName, EndpointType type, final OptionMap optionMap) {
+    public EndpointService(final Consumer<Endpoint> endpointConsumer, final Supplier<XnioWorker> workerSupplier,
+                           final Supplier<ConnectionInfo> connectionInfoSupplier,
+                           String nodeName, EndpointType type, final OptionMap optionMap) {
         this.endpointConsumer = endpointConsumer;
         this.workerSupplier = workerSupplier;
+        this.connectionInfoSupplier = connectionInfoSupplier;
         if (nodeName == null) {
             nodeName = "remote";
         }
@@ -50,6 +55,16 @@ public class EndpointService implements Service {
         builder.setEndpointName(endpointName);
         builder.setXnioWorker(workerSupplier.get());
         builder.setDefaultConnectionsOptionMap(optionMap);
+
+        final ConnectionInfo info = connectionInfoSupplier.get();
+        if (info != null && info.getDestinationUri() != null) {
+            final var connectionBuilder = builder.addConnection(info.getDestinationUri());
+            final OptionMap connectionOptions = info.getConnectionCreationOptions();
+            if (connectionOptions.contains(RemotingOptions.HEARTBEAT_INTERVAL)) {
+                connectionBuilder.setHeartbeatInterval(connectionOptions.get(RemotingOptions.HEARTBEAT_INTERVAL));
+            }
+        }
+
         try {
             endpoint = builder.build();
         } catch (IOException e) {
--- a/remoting/subsystem/src/main/java/org/jboss/as/remoting/RemoteOutboundConnectionAdd.java
+++ b/remoting/subsystem/src/main/java/org/jboss/as/remoting/RemoteOutboundConnectionAdd.java
@@ -41,12 +41,20 @@ class RemoteOutboundConnectionAdd extends AbstractAddStepHandler {
         final String authenticationContext = RemoteOutboundConnectionResourceDefinition.AUTHENTICATION_CONTEXT.resolveModelAttribute(context, fullModel).asStringOrNull();
         final String protocol = authenticationContext != null ? null : RemoteOutboundConnectionResourceDefinition.PROTOCOL.resolveModelAttribute(context, fullModel).asString();
 
+        // Install the RemoteOutboundConnectionInfoService
+        final CapabilityServiceBuilder<?> infoBuilder = context.getCapabilityServiceTarget().addService();
+        final Consumer<ConnectionInfo> infoConsumer = infoBuilder.provides(RemoteOutboundConnectionResourceDefinition.CONNECTION_INFO_CAPABILITY);
+        final Supplier<OutboundSocketBinding> osbSupplier = infoBuilder.requires(OutboundSocketBinding.SERVICE_DESCRIPTOR, outboundSocketBindingRef);
+        infoBuilder.setInstance(new RemoteOutboundConnectionInfoService(infoConsumer, osbSupplier, connOpts, username, protocol));
+        infoBuilder.install();
+
+        // Install the RemoteOutboundConnectionService
         final CapabilityServiceBuilder<?> builder = context.getCapabilityServiceTarget().addCapability(OUTBOUND_CONNECTION_CAPABILITY);
         final Consumer<RemoteOutboundConnectionService> serviceConsumer = builder.provides(OUTBOUND_CONNECTION_CAPABILITY);
-        final Supplier<OutboundSocketBinding> osbSupplier = builder.requires(OutboundSocketBinding.SERVICE_DESCRIPTOR, outboundSocketBindingRef);
-        final Supplier<AuthenticationContext> acSupplier = (authenticationContext != null) ? builder.requiresCapability(AUTHENTICATION_CONTEXT_CAPABILITY, AuthenticationContext.class, authenticationContext) : null;
+        final Supplier<ConnectionInfo> connectionInfoSupplier = builder.requires(ConnectionInfo.CONNECTION_INFO_CAPABILITY);
         builder.requiresCapability(RemotingSubsystemRootResource.REMOTING_ENDPOINT_CAPABILITY.getName(), Endpoint.class);
-        builder.setInstance(new RemoteOutboundConnectionService(serviceConsumer, osbSupplier, acSupplier, connOpts, username, protocol));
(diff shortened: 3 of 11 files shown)


yersan:
Hello @elguardian this is still a draft but notice we need to have a WFCORE issue to properly track it down in this repo, so I suggest you to move WFLY-16388 to the WildFly Core Jira project.  If part of the work to do is spread across WildFly and WildFly Core projects, then clone and move instead, thanks!

remoting/subsystem/src/main/java/org/jboss/as/remoting/RemotingSubsystemAdd.java
```diff
@@ -57,7 +57,13 @@ protected void performRuntime(OperationContext context, ModelNode operation, Res
...
 
             Supplier<XnioWorker> workerSupplier = builder.requires(IOServiceDescriptor.WORKER, workerName);
-            builder.setInstance(new EndpointService(endpointConsumer, workerSupplier, nodeName, EndpointService.EndpointType.SUBSYSTEM, map)).install();
+
+            Supplier<ConnectionInfo> connectionInfoSupplier = () -> null;
+            if (context.hasOptionalCapability(ConnectionInfo.CONNECTION_INFO_CAPABILITY.getName(), REMOTING_ENDPOINT_CAPABILITY.getName(), null)) {
+                connectionInfoSupplier = builder.requires(ConnectionInfo.CONNECTION_INFO_CAPABILITY);
+            }
+
+            builder.setInstance(new EndpointService(endpointConsumer, workerSupplier, connectionInfoSupplier, nodeName, EndpointService.EndpointType.SUBSYSTEM, map)).install();
```
yersan: The endpoint element already accepts heartbeat interval, for example we can have this:

```
<subsystem xmlns="urn:jboss:domain:remoting:8.0">
    <endpoint worker="default-remoting" heartbeat-interval="2000"/>
    <outbound-connections>
        <remote-outbound-connection name="remote-ejb-connection" outbound-socket-binding-ref="dummy-outbound-socket" protocol="remote+http">
            <properties>
                <property name="SASL_POLICY_NOANONYMOUS" value="false"/>
                <property name="SSL_ENABLED" value="false"/>
                <property name="KEEP_ALIVE" value="true"/>
                <property name="READ_TIMEOUT" value="4000"/>
            </properties>
        </remote-outbound-connection>
    </outbound-connections>
</subsystem>
```

So what's the point of adding `org.jboss.remoting3.RemotingOptions.HEARTBEAT_INTERVAL` to configure the endpoint if it already has  a heartbeat-interval?

I am not sure if those properties should be applied to the remoting endpoint here, do you know where those other SASL_POLICY_NOANONYMOUS, SSL_ENABLED ... XNIO properties are applied to?

yersan: I do not see how those properties are used to configure XNIO, also the description of that resource explicitly says they are XNIO only properties, so something like `org.jboss.remoting3.RemotingOptions.XYZ` does not fit well there.

I do not have too much knowledge on how remoting services and subsystem work, so @ropalka or @fl4via could you help us providing some feedback on this issue?

yersan: @ropalka / @fl4via friendly reminder, could you provide some input for this one? thanks!

elguardian: yersan ... this are related to Endpoint. the point has property based on properties. so anytime the endpoint connect should take into account those IIRC.

elguardian:
retest this please.

remoting/subsystem/src/main/java/org/jboss/as/remoting/RemotingSubsystemAdd.java
yersan: @elguardian  Ok, I took a closer look, I understand it a bit better now; you indeed wrote about it on the test case.

Remoting multiplexes the connections per uri on a single Endpoint, so in the case where there is already a heartbeat-interval configured at the endpoint level, this is applied as the default value.
With the changes introduced in this PR, the default value would be overridden by the new connection properties configured on the remote-outbound-connection/properties stanza. Correct?

I still have some concerns about the properties, for example the resource documentation says these are XNIO only options, but here we are adding an org.jboss.remoting3.RemotingOptions property. The documentation looks wrong in that regards, I don't know if there is an XNIO equivalent for the HEARTBEAT one though. Maybe something worth to check. Independently of that, what is documented is not enforced and we do not want to enforce it, we need to fix the documentation in any case.

remoting/tests/src/test/resources/org/jboss/as/remoting/remoting-with-heartbeat-property.xml
```diff
@@ -0,0 +1,18 @@
+<!--
+  ~ Copyright The WildFly Authors
+  ~ SPDX-License-Identifier: Apache-2.0
+  -->
+<subsystem xmlns="urn:jboss:domain:remoting:8.0">
+    <endpoint worker="default-remoting"/>
```
yersan: For the shake of completeness, What do you think if we also add a default heartbeat-interval at the enddpoint level and more than one remote-outbound-connection with different heartbeat-interval and properties? just to make a test more robust and ensure we are managing the default and the heartbeat-interval per uri.

remoting/tests/src/test/java/org/jboss/as/remoting/RemoteOutboundConnectionPropertiesTestCase.java
```diff
@@ -0,0 +1,137 @@
...
+
+        // After the fix, per-URI transport options (HEARTBEAT_INTERVAL, KEEP_ALIVE, READ_TIMEOUT)
+        // must be registered in EndpointImpl.connectionOptions for the destination URI.
+        Map<URI, OptionMap> perUriOptions = EndpointWrapper.getOptionMap(endpoint);
+        assertTrue(
+                "Per-URI transport options for " + destUri + " must be registered in the endpoint "
+                        + "(WFLY-16388: non-SASL <properties> are silently dropped)",
+                perUriOptions.containsKey(destUri));
+        assertEquals(
+                "HEARTBEAT_INTERVAL must be 2000ms as configured on remote-ejb-connection <properties>",
```
yersan: KEEP_ALIVE, READ_TIMEOUT are not asserted at all.

remoting/subsystem/src/main/java/org/jboss/as/remoting/EndpointService.java
```diff
@@ -50,6 +55,16 @@ public void start(final StartContext context) throws StartException {
         builder.setEndpointName(endpointName);
         builder.setXnioWorker(workerSupplier.get());
         builder.setDefaultConnectionsOptionMap(optionMap);
+
+        final ConnectionInfo info = connectionInfoSupplier.get();
+        if (info != null && info.getDestinationUri() != null) {
+            final var connectionBuilder = builder.addConnection(info.getDestinationUri());
+            final OptionMap connectionOptions = info.getConnectionCreationOptions();
+            if (connectionOptions.contains(RemotingOptions.HEARTBEAT_INTERVAL)) {
```
yersan, line 69: What about the other properties, for example KEEP_ALIVE and READ_TIMEOUT, they are not added per uri here. Are more candidate properties available?

remoting/tests/src/test/java/org/jboss/as/remoting/RemoteOutboundConnectionPropertiesTestCase.java
```diff
@@ -0,0 +1,137 @@
...
+                Integer.valueOf(2000),
+                perUriOptions.get(destUri).get(RemotingOptions.HEARTBEAT_INTERVAL));
```
yersan: KEEP_ALIVE, READ_TIMEOUT are not asserted at all; they should.

remoting/tests/src/test/resources/org/jboss/as/remoting/remoting-with-heartbeat-property.xml
elguardian: added to the test case.

remoting/subsystem/src/main/java/org/jboss/as/remoting/EndpointService.java
elguardian: added to the endpiont service per connection builder

remoting/tests/src/test/java/org/jboss/as/remoting/RemoteOutboundConnectionPropertiesTestCase.java
elguardian: added those.

remoting/subsystem/src/main/resources/org/jboss/as/remoting/LocalDescriptions.properties
```diff
@@ -117,7 +117,7 @@ remote-outbound-connection.security-realm.deprecated=Outbound security settings
 remote-outbound-connection.protocol=The protocol to use for the remote connection.
 remote-outbound-connection.protocol.deprecated=Outbound security settings should be migrated to an authentication-context definition.
 remote-outbound-connection.authentication-context=Reference to the authentication context instance containing the configuration for outbound connections.
-remote-outbound-connection.property=The XNIO Options that will be used during the connection creation.
+remote-outbound-connection.property=The transport options (XNIO and org.jboss.remoting3.RemotingOptions) that will be used during the connection creation. Options such as HEARTBEAT_INTERVAL, READ_TIMEOUT, WRITE_TIMEOUT, KEEP_ALIVE, and IP_TRAFFIC_CLASS are applied per-connection URI at the remoting endpoint level.
```
yersan, line 120: +1

remoting/subsystem/src/main/java/org/jboss/as/remoting/RemoteOutboundConnectionAdd.java
```diff
@@ -31,22 +31,30 @@ class RemoteOutboundConnectionAdd extends AbstractAddStepHandler {
...
+    static void installRuntimeService(final OperationContext context, final ModelNode fullModel, final String connectionName) throws OperationFailedException {
         final String outboundSocketBindingRef = RemoteOutboundConnectionResourceDefinition.OUTBOUND_SOCKET_BINDING_REF.resolveModelAttribute(context, fullModel).asString();
         final OptionMap connOpts = ConnectorUtils.getOptions(context, fullModel.get(CommonAttributes.PROPERTY));
         final String username = RemoteOutboundConnectionResourceDefinition.USERNAME.resolveModelAttribute(context, fullModel).asStringOrNull();
         final String authenticationContext = RemoteOutboundConnectionResourceDefinition.AUTHENTICATION_CONTEXT.resolveModelAttribute(context, fullModel).asStringOrNull();
         final String protocol = authenticationContext != null ? null : RemoteOutboundConnectionResourceDefinition.PROTOCOL.resolveModelAttribute(context, fullModel).asString();
 
+        // Install the RemoteOutboundConnectionInfoService
+        final CapabilityServiceBuilder<?> infoBuilder = context.getCapabilityServiceTarget().addCapability(RemoteOutboundConnectionResourceDefinition.CONNECTION_INFO_CAPABILITY);
```
yersan: addCapability() is deprecated, use addService(), the provides method call will add the capability.

remoting/subsystem/src/main/java/org/jboss/as/remoting/RemoteOutboundConnectionAdd.java
```diff
@@ -31,22 +31,30 @@ class RemoteOutboundConnectionAdd extends AbstractAddStepHandler {
...
+        installRuntimeService(context, fullModel, context.getCurrentAddressValue());
     }
 
-    static void installRuntimeService(final OperationContext context, final ModelNode fullModel) throws OperationFailedException {
+    static void installRuntimeService(final OperationContext context, final ModelNode fullModel, final String connectionName) throws OperationFailedException {
         final String outboundSocketBindingRef = RemoteOutboundConnectionResourceDefinition.OUTBOUND_SOCKET_BINDING_REF.resolveModelAttribute(context, fullModel).asString();
         final OptionMap connOpts = ConnectorUtils.getOptions(context, fullModel.get(CommonAttributes.PROPERTY));
         final String username = RemoteOutboundConnectionResourceDefinition.USERNAME.resolveModelAttribute(context, fullModel).asStringOrNull();
         final String authenticationContext = RemoteOutboundConnectionResourceDefinition.AUTHENTICATION_CONTEXT.resolveModelAttribute(context, fullModel).asStringOrNull();
         final String protocol = authenticationContext != null ? null : RemoteOutboundConnectionResourceDefinition.PROTOCOL.resolveModelAttribute(context, fullModel).asString();
```
yersan: if `authenticationContext != null` then the protocol will be `null`

This protocol with null value is passed to the `RemoteOutboundConnectionInfoService`, which will create an URI for this service without scheme. What would be its consequence?

yersan (changes requested):

remoting/subsystem/src/main/java/org/jboss/as/remoting/RemotingSubsystemAdd.java
yersan: Property description was already addressed, so I am resolving this

remoting/subsystem/src/main/java/org/jboss/as/remoting/RemoteOutboundConnectionAdd.java
elguardian: changed

remoting/subsystem/src/main/java/org/jboss/as/remoting/RemoteOutboundConnectionAdd.java
elguardian: actually it is not needed as we are reading the protocol from the conf.

remoting/subsystem/src/main/java/org/jboss/as/remoting/ConnectionInfo.java
```diff
@@ -0,0 +1,25 @@
...
+package org.jboss.as.remoting;
+
+import java.net.URI;
+import org.wildfly.service.descriptor.UnaryServiceDescriptor;
+import org.xnio.OptionMap;
+
+/**
+ * Provides connection information for an outbound remoting connection.
+ */
+public interface ConnectionInfo {
```
yersan, line 15: This can be package protected

yersan (approved):
LGTM, @elguardian just one minor about visibility of ConnectionInfo class

yersan (approved):
@elguardian In addition to make the interface package protected, would you mind squashing the commits in a single one?

Thanks!

elguardian:
I did create a problem during squash and push... 
superseeded by https://github.com/wildfly/wildfly-core/pull/6913

PR #6794, opened 2026-06-09
[WFCORE-7172] Fix intermittent SyslogHandlerTestCase failures

elguardian:
Issue: https://redhat.atlassian.net/browse/WFCORE-7172

just send the the logs we need/want and filter deployment ones..


testsuite/standalone/src/test/java/org/jboss/as/test/integration/logging/syslog/SyslogHandlerTestCase.java
```diff
@@ -250,11 +252,15 @@ public void setup(final ManagementClient managementClient) throws Exception {
...
 
-            op = Operations.createAddOperation(SYSLOG_PROFILE_ROOT_LOGGER_ADDR);
+            op = Operations.createAddOperation(SYSLOG_PROFILE_LOGGER_ADDR);
             op.get("level").set("TRACE");
             op.get("handlers").add("SYSLOG");
             builder.addStep(op);
 
+            op = Operations.createAddOperation(SYSLOG_PROFILE_ROOT_LOGGER_ADDR);
+            op.get("level").set("TRACE");
+            builder.addStep(op);
```
jamezp: If we don't want all messages, we don't need to add the root logger at all. It won't do anything without handlers anyway.

elguardian: removed

yersan:
@elguardian  as a good practice to follow, squash the commits or add the Jira number on all of them; that helps to easily identify the original Jira when we are looking at the git history

elguardian:
@yersan squashed

yersan (approved):
LGTM following the latest @jamezp review.

@jamezp Could you approve from your side?

jamezp (approved):
LGTM, thank you @elguardian

PR #6797, opened 2026-06-10
[WFCORE-7616] Deprecate the /subsystem=elytron/policy=* resource.

darranl:
This resource will be removed at a later date once no longer supported.

Additionally for Jakarta EE 11 don't fail the service start up, instead log a WARN.

https://redhat.atlassian.net/browse/WFCORE-7616

### DIFF
--- a/elytron/src/main/java/org/wildfly/extension/elytron/PolicyDefinitions.java
+++ b/elytron/src/main/java/org/wildfly/extension/elytron/PolicyDefinitions.java
@@ -174,10 +174,18 @@ class PolicyDefinitions {
             private Service<Policy> createPolicyService(Consumer<Consumer<Policy>> policyProvider) {
                 return new Service<Policy>() {
                     volatile Policy original;
+                    volatile boolean restorePolicy;
 
                     @Override
                     public void start(StartContext context) throws StartException {
-                        original = getPolicy();
+                        try {
+                            original = getPolicy();
+                            restorePolicy = true;
+                        } catch (UnsupportedOperationException e) {
+                            restorePolicy = false;
+                            ElytronSubsystemMessages.ROOT_LOGGER.settingPolicyNotSupported();
+                        }
+
 
                         try {
                             policyProvider.accept(this::setPolicy);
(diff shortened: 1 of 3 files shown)


elytron/src/main/java/org/wildfly/extension/elytron/_private/ElytronSubsystemMessages.java
```diff
@@ -671,6 +671,10 @@ public interface ElytronSubsystemMessages extends BasicLogger {
     @LogMessage(level = WARN)
     void noAllowedJkuValuesSpecifiedForTokenRealm(String realmName, String systemPropertyName);
 
+    @Message(id = 1091, value = "From Jakarta EE 11 setting a Policy is no longer required, the /subsystem=elytron/policy=* resource(s) can be removed.")
```
bstansberry:
```suggestion
    @Message(id = 1091, value = "In an EE 11 or later environment setting a Policy is no longer required; the /subsystem=elytron/policy=* resource(s) can be removed.")
```

elytron/src/main/resources/org/wildfly/extension/elytron/LocalDescriptions.properties
```diff
@@ -1663,6 +1663,7 @@ elytron.dir-context.properties=The additional connection properties for the DirC
 #######################
 
 elytron.policy=A definition that sets up a policy provider.
+elytron.policy.deprecated=Support for security Policy implementations has been removed from Java 25, setting a JACC Policy will no longer be required from Jakarta EE 11.
```
bstansberry:
```suggestion
elytron.policy.deprecated=Support for security Policy implementations has been removed from Java 25, and setting a JACC Policy is no longer required in an EE 11 or later environment.
```

bstansberry (changes requested):

elytron/src/main/java/org/wildfly/extension/elytron/PolicyDefinitions.java
```diff
@@ -174,10 +174,18 @@ protected void performRuntime(OperationContext context, ModelNode operation, Mod
             private Service<Policy> createPolicyService(Consumer<Consumer<Policy>> policyProvider) {
                 return new Service<Policy>() {
                     volatile Policy original;
+                    volatile boolean restorePolicy;
```
yersan, line 177: `restorePolicy` is a local variable and it is not used at all. It is harmless, but I don't get the point of having it here.

I do not think this will prevent merging it, but just in case CC @darranl

yersan (approved):
`org.jboss.as.test.clustering.cluster.web.remote.FineHotRodSessionExpirationTestCase.test(URL, URL)` is a flaky test, so this looks good for main. @bstansberry review was already addressed.

rhusar:
> `org.jboss.as.test.clustering.cluster.web.remote.FineHotRodSessionExpirationTestCase.test(URL, URL)` is a flaky test, so this looks good for main. @bstansberry review was already addressed.

n.b. the fix for the test is available at https://github.com/wildfly/wildfly/pull/20137

yersan:
Thanks for the info @rhusar !

PR #6798, opened 2026-06-10
[WFCORE-7616] Deprecate the /subsystem=elytron/policy=* resource.

darranl:
This resource will be removed at a later date once no longer supported.

Additionally for Jakarta EE 11 don't fail the service start up, instead log a WARN.

https://redhat.atlassian.net/browse/WFCORE-7616

Upstream PR: https://redhat.atlassian.net/browse/WFCORE-7616

### DIFF
--- a/elytron/src/main/java/org/wildfly/extension/elytron/PolicyDefinitions.java
+++ b/elytron/src/main/java/org/wildfly/extension/elytron/PolicyDefinitions.java
@@ -174,10 +174,18 @@ class PolicyDefinitions {
             private Service<Policy> createPolicyService(Consumer<Consumer<Policy>> policyProvider) {
                 return new Service<Policy>() {
                     volatile Policy original;
+                    volatile boolean restorePolicy;
 
                     @Override
                     public void start(StartContext context) throws StartException {
-                        original = getPolicy();
+                        try {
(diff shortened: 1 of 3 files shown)


darranl:
Optional but this might help where users are getting start up errors due to their configuration containing the JACC Policy as they move over from an older WildFly version using Jakarta EE 10.

bstansberry (changes requested):
See my comments on the PR against main.

I support getting this into WF 40.0.1.Final, since currently using a config with this element that worked in WF 39 or earlier now dumps a scary stack trace in the log.

yersan:
Something went wrong with this CI report, I can see here Job results from PR 6797. Regarding the ones executed for this PR, CI seems OK, the failed tests reported by [WildFly Core -> Full Integration - Windows - JDK 17 (Pull Request 32.x (WildFly 40))](https://ci.wildfly.org/buildConfiguration/WildFlyCore_PullRequest32xWildFly40_WildFlyCoreFullIntegrationWindowsJdk17/567679) passed with the same commit in previous runs. 

We still need to merge main first, and change the message text as requested by review. Changing the message text should not trigger a new CI cycle, so we can do that on Monday before preparing the release.

PR #6802, opened 2026-06-11
Add skip-teamcity label to Dependabot PRs

yersan:
No Jira needed. 

Just add default labels for dependabot upgrade PR

See https://github.com/wildfly/wildfly/pull/20136


yersan:
ahh, too late for adding the label ... let me start over again ....

PR #6803, opened 2026-06-11
Add skip-teamcity label to Dependabot PRs

yersan:
No Jira needed.

Just add default labels for dependabot upgrade PR

See https://github.com/wildfly/wildfly/pull/20136

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -11,6 +11,9 @@ registries:
 updates:
   - package-ecosystem: "maven"
     directory: "/"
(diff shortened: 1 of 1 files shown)


rhusar:
Hmhm, the label still doesn't work?

yersan:
yes, is seems it doesn't

bstansberry:
Thanks @rhusar 

If the pull-player for core hasn't been upgraded yet I'm sure it will be soon, and this will start doing what's wanted.

PR #6804, opened 2026-06-11
[WFCORE-7619] verbosegclog java opt is missing from JAVA_OPTS echo output on standalone bat script on Semeru JDKs

lvydra:
https://redhat.atlassian.net/browse/WFCORE-7619

Related to: https://github.com/wildfly/wildfly-core/pull/6726

### DIFF
--- a/core-feature-pack/common/src/main/resources/content/bin/standalone.bat
+++ b/core-feature-pack/common/src/main/resources/content/bin/standalone.bat
@@ -274,6 +274,7 @@ if not "%PRESERVE_JAVA_OPTS%" == "true" (
                 set TMP_PARAM=-Xverbosegclog:"!JBOSS_LOG_DIR!\gc.log"
                 "!JAVA!" !TMP_PARAM! -version > nul 2>&1
                 if not errorlevel == 1 (
+                   rem Combination of Semeru JDK with a space in the path needs keeping GC_OPTS separate and passing it directly to the Java command
                    set "GC_OPTS=!TMP_PARAM!"
                 ) else (
                    set "GC_OPTS="
@@ -335,7 +336,7 @@ echo   JBOSS_HOME: "%JBOSS_HOME%"
 echo.
 echo   JAVA: "%JAVA%"
 echo.
-echo   JAVA_OPTS: "%JAVA_OPTS%"
+echo   JAVA_OPTS: "%JAVA_OPTS% %GC_OPTS%"
 echo.
 echo ===============================================================================
 echo.


core-feature-pack/common/src/main/resources/content/bin/standalone.bat
```diff
@@ -335,7 +335,7 @@ echo   JBOSS_HOME: "%JBOSS_HOME%"
 echo.
 echo   JAVA: "%JAVA%"
 echo.
-echo   JAVA_OPTS: "%JAVA_OPTS%"
+echo   JAVA_OPTS: "%JAVA_OPTS% %GC_OPTS%"
```
yersan, line 339: More than append the GC_OPTS here, we need to add the GC_OPTS to the JAVA_OPTS after initializing it.
So any code after initializing the GC_OPTS can work directly with the final JAVA_OPTS.

I think we could get the rid of GC_OPTS and work directly with JAVA_OPTS. I mean, remove the `else` and GC_OPTS from https://github.com/wildfly/wildfly-core/blob/main/core-feature-pack/common/src/main/resources/content/bin/standalone.bat#L274-L280  and work directly with ` set "JAVA_OPTS=!JAVA_OPTS! !TMP_PARAM!"`

That seems more consistent and improves the maintenance

yersan (changes requested):

core-feature-pack/common/src/main/resources/content/bin/standalone.bat
yersan: I am not sure if there were any issue doing it in that way, but looking at the code now, that approach looks more consistent, I do not see any justification to creating a GC_OPTS at this time

lvydra: Hi @yersan, thank you for looking at it. Using directly `set "JAVA_OPTS=!JAVA_OPTS! !TMP_PARAM!"` would be ideal, but unfortunately that doesn't work for a combination of Semeru JDK with a space in the path (https://github.com/wildfly/wildfly-core/pull/6726)

yersan: Ahh, ok thanks for clarifying it @lvydra .... that's pretty nasty issue from using Semeru JDK

We should be fine then, but could you at least a comment about that in the code? 

I guess when someone in the future check again the line where we are adding `"%JAVA_OPTS% %GC_OPTS%"` the same question will arise, thanks! ....  we should have a better approach, but that's not a priority

lvydra: @yersan  I have added a comment to the GC_OPTS.

yersan: Ok, just to see if I understand the main problem with this. According with https://github.com/wildfly/wildfly-core/pull/6726

"" the combination of parentheses in path ( (test) ), quotes needed for spaces, and batch re-parsing variables during expansion seems to create an unsolvable conflict when trying to merge gc log opts into JAVA_OPTS.""

For non Semeru J9 JDKs, that was solvable by adding quotes across all the path, however on Semeru, Xverbosegclog: does not accept quoting the path:

`-Xverbosegclog:"\"E:\wildfly-38.0.1.Final (test)\standalone\log\gc.log\""`

results in:

`JVMJ9GC063E Unable to open file '"E:\wildfly-38.0.1.Final (test)\standalone\log\gc.log"' for writing.`

Is that correct @lvydra ?

lvydra: Yes, it is.

yersan: ok, thanks for confirming, I've created https://redhat.atlassian.net/browse/WFCORE-7627 to track it

yersan (approved):

core-feature-pack/common/src/main/resources/content/bin/standalone.bat
lvydra: Great, thanks. I have taken it. I will do some more research to see what can be done.

PR #6805, opened 2026-06-11
Fix for WFCORE-7618, [Bootable JAR] Support for jdk.serialFilter is missing

jfdenise:
ISSUE: https://redhat.atlassian.net/browse/WFCORE-7618

### DIFF
--- a/bootable-jar/runtime/src/main/java/org/wildfly/core/jar/runtime/Arguments.java
+++ b/bootable-jar/runtime/src/main/java/org/wildfly/core/jar/runtime/Arguments.java
@@ -8,6 +8,7 @@ import static org.wildfly.core.jar.runtime.Constants.DEPLOYMENT_ARG;
 
 import java.io.IOException;
 import java.io.InputStream;
+import java.io.ObjectInputFilter;
 import java.net.MalformedURLException;
 import java.net.URL;
 import java.nio.file.Files;
@@ -32,6 +33,11 @@ import org.wildfly.core.jar.runtime._private.BootableJarLogger;
  * @author <a href="mailto:jperkins@redhat.com">James R. Perkins</a>
  */
 final class Arguments {
+    private static final String JDK_SERIAL_FILTER = "jdk.serialFilter";
+    private static final String DISABLE_JDK_SERIAL_FILTER_ENV = "DISABLE_JDK_SERIAL_FILTER";
+    private static final String JDK_SERIAL_FILTER_ENV = "JDK_SERIAL_FILTER";
+    // This default value comes from the standalone.conf file, in case it is changes, make sure to update this value.
+    private static final String DEFAULT_SERIAL_FILTER = "maxbytes=10485760;maxdepth=128;maxarray=100000;maxrefs=300000";
 
     private Arguments(final BootableEnvironment environment) {
         this.environment = environment;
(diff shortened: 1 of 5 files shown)


bootable-jar/runtime/src/main/java/org/wildfly/core/jar/runtime/Arguments.java
```diff
@@ -105,6 +116,17 @@ private void handleArguments(List<String> args) throws Exception {
...
         }
+        // Special handling of jdk.serialFilter
+        if (!jdkSerialFilterSet) {
+            String disabled = System.getenv(DISABLE_JDK_SERIAL_FILTER_ENV);
+            if (disabled == null || !disabled.equalsIgnoreCase("true")) {
+                String filter = System.getenv(JDK_SERIAL_FILTER_ENV);
+                if (filter == null) {
+                    filter = DEFAULT_SERIAL_FILTER;
+                }
+                systemProperties.put(JDK_SERIAL_FILTER, filter);
```
bstansberry: This is a bit fragile since it assumes the static init of ObjectInputFilter.Config has not happened yet. That's probably the case, but it's out of our control.

Also calling ObjectInputFilter.Config.createFilter and ObjectInputFilter.Config.setSerialFilter will ensure the desired global filter is the one that's wanted.

jfdenise: @bstansberry , thank-you, you are right. 
I should generalize the usage of the API call even in the case we have a filter explicitly set in the user arguments (-D property or --properties). The class could have been init and setting the property (that we do in the same method) would have no effect. 
I have handled all cases with the explicit API call and handled the case where the system property is actually set in the Java args, not in the bootable JAR args.

bootable-jar/runtime/src/main/java/org/wildfly/core/jar/runtime/Arguments.java
```diff
@@ -105,6 +116,27 @@ private void handleArguments(List<String> args) throws Exception {
...
+                    }
+                    serialFilter = filter;
+                }
+            } else {
+                // The filter has been explicitly set, don't keep the system property.
+                // The filter will be set by the Bootable JAR runtime.
+                serialFilter = jdkSerialFilterSet;
+                systemProperties.remove(JDK_SERIAL_FILTER);
+            }
+        }
```
bstansberry, line 143: An 'else' block would be good here. Pseudo code being:

else (if !DISABLE_JDK_SERIAL_FILTER_ENV && JDK_SERIAL_FILTER_ENV != null) {
   log.info("The JDK_SERIAL_FILTER_ENV environment variable is being ignored because the global ObjectInputFilter was already configured by the jdk.serialFilter system or security property");
}

bstansberry (changes requested):
Looks good except for some possible corner-case INFO logging.

bootable-jar/runtime/src/main/java/org/wildfly/core/jar/runtime/Arguments.java
jfdenise: @bstansberry , thank-you, I added such log, a bit delayed due to log configuration, but logged the way you suggested.

bstansberry (approved):

PR #6812, opened 2026-06-15
[WFCORE-6343] Add management endpoint capability requirement to remoting-connector

Ohifeme:
issue: https://redhat.atlassian.net/browse/WFCORE-6343


yersan (changes requested):
Hello @Ohifeme , always open you topic branch from latest `main`, this PR is for WFCORE-6343 but is it based on WFCORE-5985

Could you fix that? we need independent PRs, moreover when they are completely unrelated, thanks!

yersan (changes requested):
@Ohifeme  Errors on the CI are related to this change, "org.wildfly.management.endpoint" is not a valid capability.

I did a quick look and after reading the Jira I ws not able to find the service pointed out in the error description, to later check what capability is related with that service or resource.

@darranl Could you help us with this by carifying a bit more on the Jira what management capability were you talking about?

PR #6832, opened 2026-06-26
WFCORE-7625 Deprecate TimeoutUtil.adjust(int) and add Duration-based variant of the method

rhusar:
The integer-based adjust method loses precision very quickly; deprecate it in favor of a new adjust(Duration) overload that operates at sufficient, nanosecond, granularity.

Migrate all callers from TimeoutUtil.adjust(int) to adjust(Duration) and make all timeouts proper Durations

Replace all usages of the deprecated adjust(int) with the new Duration-based one.
Timeout fields are stored as Duration for clarity and flexibility.

Note that this is 95% mechanical with no intention to fix pre-existing issues.

Resolves
https://redhat.atlassian.net/browse/WFCORE-7625

### DIFF
--- a/testsuite/client-old-server/src/test/java/org/wildfly/client/old/server/ClientAgainstOldServerTestCase.java
+++ b/testsuite/client-old-server/src/test/java/org/wildfly/client/old/server/ClientAgainstOldServerTestCase.java
@@ -11,6 +11,7 @@ import static org.jboss.as.controller.descriptions.ModelDescriptionConstants.PRO
 import static org.jboss.as.controller.descriptions.ModelDescriptionConstants.READ_RESOURCE_OPERATION;
 import static org.jboss.as.controller.descriptions.ModelDescriptionConstants.RELEASE_VERSION;
 
+import java.time.Duration;
 import java.util.List;
 import java.util.concurrent.ExecutionException;
 import java.util.concurrent.Future;
@@ -45,7 +46,7 @@ import org.wildfly.core.testrunner.ManagementClient;
 public class ClientAgainstOldServerTestCase {
 
     // Max time to wait for some action to complete, in s
-    private static final int TIMEOUT = TimeoutUtil.adjust(240);
+    private static final Duration TIMEOUT = TimeoutUtil.adjust(Duration.ofMinutes(4));
 
     private final Version.AsVersion version;
 
@@ -103,7 +104,7 @@ public class ClientAgainstOldServerTestCase {
 
     private void awaitDeploymentExecution(Future<?> future) {
         try {
-            future.get(TIMEOUT, TimeUnit.SECONDS);
+            future.get(TIMEOUT.toNanos(), TimeUnit.NANOSECONDS);
         } catch (InterruptedException e) {
             Thread.currentThread().interrupt();
             throw new RuntimeException(e);
--- a/testsuite/manualmode/src/test/java/org/jboss/as/test/manualmode/deployment/DeploymentScannerUnitTestCase.java
+++ b/testsuite/manualmode/src/test/java/org/jboss/as/test/manualmode/deployment/DeploymentScannerUnitTestCase.java
@@ -20,6 +20,7 @@ import static org.junit.Assert.assertEquals;
 import java.io.IOException;
 import java.nio.file.Files;
 import java.nio.file.Path;
+import java.time.Duration;
 import java.time.LocalDateTime;
 import java.time.format.DateTimeFormatter;
 import java.util.concurrent.TimeUnit;
@@ -60,7 +61,7 @@ public class DeploymentScannerUnitTestCase extends AbstractDeploymentScannerBase
     private static final String JAR_TWO = "deployment-startup-two.jar";
     private static final PathAddress DEPLOYMENT_ONE = PathAddress.pathAddress(DEPLOYMENT, JAR_ONE);
     private static final PathAddress DEPLOYMENT_TWO = PathAddress.pathAddress(DEPLOYMENT, JAR_TWO);
-    private static final int TIMEOUT = TimeoutUtil.adjust(30000);
+    private static final Duration TIMEOUT = TimeoutUtil.adjust(Duration.ofSeconds(30));
     private static final DateTimeFormatter TIME_FORMATTER = DateTimeFormatter.ofPattern("uuuu-MM-dd HH:mm:ss,SSS");
 
     @SuppressWarnings("unused")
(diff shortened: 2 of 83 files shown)


jasondlee (approved):
API change is great. Spot checked quite a few usage updates. Looks good.

testsuite/client-old-server/src/test/java/org/wildfly/client/old/server/ClientAgainstOldServerTestCase.java
```diff
@@ -45,7 +46,7 @@
 public class ClientAgainstOldServerTestCase {
 
     // Max time to wait for some action to complete, in s
-    private static final int TIMEOUT = TimeoutUtil.adjust(240);
+    private static final Duration TIMEOUT = TimeoutUtil.adjust(Duration.ofSeconds(240));
```
jasondlee: I said everything looks good, then noticed this. Everywhere else, `.adjust(x)` becomes `.adjust(Duration.ofSeconds(x/1000))`. Here, though, `240` becomes 240 seconds, which makes sense, but why are the others treated as millis to seconds, while this one (and others?) is treated as seconds -> seconds?

rhusar: @jasondlee I think I see what you are getting at.

So the problem with current way people do things, they create an int/long field and add a number in ms precision, which is not well human-redable, as you need to parse out the zeroes out - and people aren't using _ jdk separator. Apart from the fact how inflexible that is if you need to use it in different context like Future.get, Thread.sleep, some waitForXyz API, etc.

What this PR also does is demonstrating how to do this in convenient human-readable way and unit is explicit. So in this case, 240 is unnecessary as that's 4 minutes. 

I have updated this and other fields to also do the `ofMinutes` that I initially didn't do as that reduces granularity (e.g. going from 60 seconds to 90 seconds is more complicated when using `.ofMinutes(1)`).

jasondlee: Ah, ok. So the units the number represents is dependent on what the value is then passed to. `TimeoutUtil.adjust` merely fudges the number based on some logic without any concern on what it actually represents. I think I was assuming the value passed in was a specific unit of measurement, but it's not, so I think the change is good and I understand the API correctly now. :)

jasondlee (approved):

testsuite/client-old-server/src/test/java/org/wildfly/client/old/server/ClientAgainstOldServerTestCase.java
rhusar: @jasondlee https://download.java.net/java/early_access/valhalla/docs/api/java.base/java/time/Duration.html :-)

rhusar:
> **Core -> Full Integration** [Build 15237](https://ci.wildfly.org/buildConfiguration/WildFlyCore_PullRequest_GalleonWildFlyCoreFullIntegrationLinuxJdk21/571629) outcome was **FAILURE** using a merge of [05d4351](https://github.com/wildfly/wildfly-core/commit/05d43510e282fe233ea06344e2eb4f86b99a115c) Summary: Tests passed: 292, ignored: 12; exit code 1 (Step: Build and test full with JDK 21 (Maven)) (new) Build time: 00:44:06

Oh no, this is a [bug](https://github.com/rhusar/wildfly-glow/blob/main/core/src/main/java/org/wildfly/glow/DeploymentScanner.java#L662) in glow as it's non-determistic... 

```
17:47:24   [ERROR] Failed to execute goal org.wildfly.glow:wildfly-glow-arquillian-plugin:2.1.0.Final:scan (scan-legacy-sar-layers) on project wildfly-ts-integ-basic: Error in glow discovery.
17:47:24   [ERROR] -Expected: [ee-integration, mail, naming, sar, transactions]==>ee-core-profile-server,mail,sar,transactions
17:47:24   [ERROR] -Found   : [ee-integration, mail, naming, resource-adapters, sar, transactions]==>ee-core-profile-server,mail,resource-adapters,sar
17:47:24   [ERROR] -> [Help 1]
```

rhusar:
Uh... this is probably an issue in WF - opened https://github.com/wildfly/wildfly/pull/20213

So disregard that failure for this PR.

yersan (commented):
I'm just waiting for the Galleon Job, but maybe for a less verbose api, we could add the following methods to the TimeUtil:

TimeoutUtil.adjustOfMinutes(int)
TimeoutUtil.adjustOfSeconds(int)
TimeoutUtil.adjustOfMilliseconds(int)

@rhusar WDYT? Makes sense to you?

rhusar:
> I'm just waiting for the Galleon Job, but maybe for a less verbose api, we could add the following methods to the TimeUtil:
> 
> TimeoutUtil.adjustOfMinutes(int) TimeoutUtil.adjustOfSeconds(int) TimeoutUtil.adjustOfMilliseconds(int)
> 
> @rhusar WDYT? Makes sense to you?

While I see where you are coming from, I would much rather keep the single adjust(Duration) method.

The Duration.ofSeconds/ofMinutes/ofMillis factories are standard JDK API and self-documenting where the "verbosity" becomes the readability. Adding adjustOfSeconds(int) etc. would reintroduce unit-untyped int parameters which is what we're moving away from, and we'd need one wrapper per unit while still not covering everything Duration already offers (Duration.of(n, ChronoUnit), Duration.parse(...), etc.).

I think the current API is the right one. As they say, KISS. :-)

Also, it's not necessary at all if verbosity is a concern if you use static imports, if you import those 2 methiods satically you get even simpler and complety standard:

```
Duration ... = adjust(ofSeconds(30));
```

yersan:
Ok @rhusar , fine with me, thanks

~Weird, the Galleon Job got stuck testing the layers, going to kill it now and trigger it again ...~

Seems ok, it is moving on ... I got the impression this Job was quicker

testsuite/manualmode/src/test/java/org/jboss/as/test/manualmode/deployment/DeploymentScannerUnitTestCase.java
```diff
@@ -352,9 +353,9 @@ private PathAddress getLoggerConsoleHandler() {
 
 
     private void waitFor(String message, ExceptionWrappingSupplier<Boolean> condition) throws Exception {
-        long timeout = System.currentTimeMillis() + TimeoutUtil.adjust(TIMEOUT);
+        long timeout = System.currentTimeMillis() + TIMEOUT.toMillis();
```
yersan, line 356: Great!, it was being adjusted twice before this

yersan (approved):

testsuite/manualmode/src/test/java/org/jboss/as/test/manualmode/deployment/DeploymentScannerUnitTestCase.java
rhusar: Oh right, kind of accidental fix in the cleanup, but goes to show this is more readable.

yersan:
Thanks @rhusar and @jasondlee !

rhusar:
@yersan I will prepare at some point analogous change for WF, tracked as https://redhat.atlassian.net/browse/WFLY-22059, just to state the obvious, this won't interfere with core upgrade in WF in any way (apart from logging a new deprecation warnings).

PR #6842, opened 2026-07-08
[WFCORE-7380] Remove JDK 23 specific profile, add per-module annotationProcessorPaths

javideros:
Fixes: https://issues.redhat.com/browse/WFCORE-7380

Removes the jdk23 profile containing the global -proc:full compiler argument and configures annotationProcessorPaths explicitly in each module that requires annotation processing.

Implementation by @rhusar, adopted here as agreed on this PR, plus two follow-up fixes discussed in the thread:
- removed the leftover jboss-logging-processor dependency from embedded, now redundant since the processor path covers it (jdeparser and the processor's other transitive dependencies reach the compiler plugin through that path)
- removed the annotationProcessorPaths block from core-security, which has no logging-annotated types and generated nothing

Verified with a full clean build: annotation processing produces output across all 30 modules that need it, and core-security builds cleanly generating nothing.

Counterpart of WFLY-21910 on the WildFly side.

Relates to https://github.com/wildfly/wildfly/pull/20205

### DIFF
--- a/bootable-jar/runtime/pom.xml
+++ b/bootable-jar/runtime/pom.xml
@@ -90,14 +90,6 @@
             <optional>true</optional>
         </dependency>
         <dependency>
-            <groupId>org.jboss.logging</groupId>
-            <artifactId>jboss-logging-processor</artifactId>
-            <!-- This is a compile-time dependency of this project, but is not needed at compile or runtime by other
-            projects that depend on this project.-->
-            <scope>provided</scope>
-            <optional>true</optional>
-        </dependency>
-        <dependency>
             <groupId>org.jboss.stdio</groupId>
             <artifactId>jboss-stdio</artifactId>
             <scope>provided</scope>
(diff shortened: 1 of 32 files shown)


yersan:
Just for the record, we can forget the failures on the "Integration" Jobs, we need the WildFly counterpart PR merged first to proceed with this one.

yersan (approved):

yersan:
Thanks for all the work on this one @javideros !! I know it has been a long standing task with several discussions to get it finally merged, thanks for keeping with it until the end!

javideros:
Thanks @yersan & @rhusar! Glad we finally got this across the finish line. Really appreciate all the helpful feedback and guidance.

rhusar:
> Thanks @yersan & @rhusar! Glad we finally got this across the finish line. Really appreciate all the helpful feedback and guidance.

Cool! Happy to see this finished :-)

PR #6845, opened 2026-07-09
[WFCORE-7638] all server process types, other than domain server, sho…

emmartins:
Issue: https://redhat.atlassian.net/browse/WFCORE-7638

### DIFF
--- a/controller/src/main/java/org/jboss/as/controller/extension/AbstractLegacyExtension.java
+++ b/controller/src/main/java/org/jboss/as/controller/extension/AbstractLegacyExtension.java
@@ -43,7 +43,7 @@ public abstract class AbstractLegacyExtension implements Extension {
             // to legacy servers to work
             ControllerLogger.MGMT_OP_LOGGER.ignoringUnsupportedLegacyExtension(subsystemNames, extensionName);
             return;
-        } else if (context.getProcessType() == ProcessType.STANDALONE_SERVER) {
+        } else if (context.getProcessType().isServer()) {
             if (context.getRunningMode() == RunningMode.ADMIN_ONLY) {
                 //log a message, but fall through and register the model
                 ControllerLogger.MGMT_OP_LOGGER.removeUnsupportedLegacyExtension(subsystemNames, extensionName);
@@ -66,7 +66,7 @@ public abstract class AbstractLegacyExtension implements Extension {
             // Do nothing. This allows the extension=cmp:add op that's really targeted
             // to legacy servers to work
             return;
-        } else if (context.getProcessType() == ProcessType.STANDALONE_SERVER && context.getRunningMode() != RunningMode.ADMIN_ONLY) {
+        } else if (context.getProcessType().isServer() && context.getRunningMode() != RunningMode.ADMIN_ONLY) {
             throw new UnsupportedOperationException(ControllerLogger.ROOT_LOGGER.unsupportedLegacyExtension(extensionName));
         }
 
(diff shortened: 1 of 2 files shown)


controller/src/main/java/org/jboss/as/controller/extension/AbstractLegacyExtension.java
```diff
@@ -43,7 +43,7 @@ public void initialize(ExtensionContext context) {
             // to legacy servers to work
             ControllerLogger.MGMT_OP_LOGGER.ignoringUnsupportedLegacyExtension(subsystemNames, extensionName);
             return;
-        } else if (context.getProcessType() == ProcessType.STANDALONE_SERVER) {
+        } else if (context.getProcessType().isServer()) {
```
yersan, line 46: I suspect that `initializeParsers()` should follow the same pattern. It could be a conner case, but I think it is better to be consistent.

The addExtensionHandler initializes the parsers before initializing the extension (for extensions that are not configured in our standard config files):

https://github.com/wildfly/wildfly-core/blob/main/controller/src/main/java/org/jboss/as/controller/extension/ExtensionAddHandler.java#L126

So, if at the end the extension won't be initialize for anything except DOMAIN_SERVER, HOST_CONTROLLER and EMBEDDED_HOST_CONTROLLER, we don't need to initialize its parsers.

yersan (changes requested):
Seems there could be a corner case, do you agree @emmartins ?

subsystem-test/framework/src/main/java/org/jboss/as/subsystem/test/SubsystemTestDelegate.java
```diff
@@ -155,7 +156,9 @@ String getMainSubsystemName() {
     void initializeParser() throws Exception {
         //Initialize the parser
         xmlMapper = XMLMapper.Factory.create();
-        extensionParsingRegistry = ExtensionRegistry.builder(this.getProcessType()).withStability(this.stability).build();
+        // legacy extensions require ADMIN-ONLY to initialize parsers on servers other than domain server
```
emmartins, line 159: This is needed to prevent all legacy extension subsystem tests (extending this) from failing. Parsers from such extensions now can only initialize on admin-mode running mode, when on server processes other than domain server.

emmartins: The JSSE subsystem tests that failed will require similar changes, cause those cloned the code changed here...

yersan (commented):
This one is going to require also changes in WildFly, maybe those changes can go before this one.

@emmartins tthis one arrived a bit late to be included on 33.x (WF41) so would have to to leave this for WildFly Core 34.x (WF42), sounds fine with you?

emmartins:
@yersan it would be nice to have this on 41, cause there is no workaround for what this fix, need it to be able to use the new features on the migration tool that auto discover a server's supported extensions and subsystems. But of course I can do as before and input those manually.

subsystem-test/framework/src/main/java/org/jboss/as/subsystem/test/SubsystemTestDelegate.java
yersan: Ok, we can use EMBEDDED_HOST_CONTROLLER as process type or run them in ADMIN_ONLY mode

yersan (approved):

PR #6847, opened 2026-07-10
[WFCORE-6447] Beef up code documentation of ModelTestModelControllerService.ControllerXXx

Ohifeme:
issue: https://redhat.atlassian.net/browse/WFCORE-6447

### DIFF
--- a/model-test/src/main/java/org/jboss/as/model/test/ModelTestModelControllerService.java
+++ b/model-test/src/main/java/org/jboss/as/model/test/ModelTestModelControllerService.java
@@ -574,8 +574,18 @@ public abstract class ModelTestModelControllerService extends AbstractController
         }
     }
 
-    //These are here to overload the constructor used for the different legacy controllers
-
+    /**
+     * Marker classes used solely to select the correct {@link ModelTestModelControllerService} constructor
+     * for a given legacy WildFly Core controller version. Because Java does not allow distinguishing
+     * overloaded constructors by a plain version value alone (the signatures would be ambiguous), each
+     * supported legacy version has its own distinct marker type. Passing {@code ControllerXXx.INSTANCE}
+     * as the final constructor argument causes the compiler to resolve the intended overload unambiguously.
+     *
+     * <p>The actual callers of these overloaded constructors that use these marker classes as discriminators
+     * are in the <a href="https://github.com/wildfly/wildfly-legacy-test">wildfly/wildfly-legacy-test</a>
+     * module controller implementations, where a concrete {@link ModelTestModelControllerService}
+     * implementation is instantiated for each legacy controller under test.
+     */
 
     @SuppressWarnings("InstantiationOfUtilityClass")
     public static class Controller23x {
--- a/model-test/src/main/java/org/jboss/as/model/test/ModelTestModelControllerService.java
+++ b/model-test/src/main/java/org/jboss/as/model/test/ModelTestModelControllerService.java
@@ -630,8 +630,14 @@ public abstract class ModelTestModelControllerService extends AbstractController
         }
     }
 
-    //These are here to overload the constructor used for the different legacy controllers
-
+    // Marker classes used solely to select the correct ModelTestModelControllerService constructor
+    // for a given legacy WildFly Core controller version. Because Java does not allow distinguishing
+    // overloaded constructors by a plain version value alone (the signatures would be ambiguous), each
+    // supported legacy version has its own distinct marker type. Passing ControllerXXx.INSTANCE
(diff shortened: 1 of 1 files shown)


model-test/src/main/java/org/jboss/as/model/test/ModelTestModelControllerService.java
```diff
@@ -574,8 +574,23 @@ public BootOperationFailedException(final String msg, final ModelNode descriptio
...
-
+    /**
+     * Marker classes used solely to select the correct {@link ModelTestModelControllerService} constructor
+     * for a given legacy WildFly Core controller version. Because Java does not support calling an overloaded
+     * constructor based on a version enum or integer alone (all overloads would have the same erasure), each
+     * supported legacy version has its own distinct marker type. Passing {@code ControllerXXx.INSTANCE} as the
+     * final constructor argument causes the compiler to resolve the intended overload unambiguously.
+     *
+     * <p>Within this module the {@code INSTANCE} constants are never read; their only purpose is to carry
+     * type information at the call site. The actual callers that supply these markers are in the external
```
yersan: @Ohifeme

```....  The actual callers that supply these markers are in the external
     * WildFly test framework project (wildfly/wildfly), specifically in the {@code LegacyKernelServicesImpl}
     * classes under {@code subsystem-test} and {@code core-model-test}" ....
```    
This is not correct.  WildFly test framework project is not in wildfly/wildfly, it is in this repo wildfly/wildfly-core.

There are no callers in LegacyKernelServicesImpl invoking these ModelTestModelControllerService.java‎
 overloaded constructors. This looks like as an AI hallucination that seems was not properly reviewed, do not forget to review and understand its output.


Probably the way to get this would have been to check from Git history on which Jira were those methods added, which could have driven you to https://redhat.atlassian.net/browse/WFCORE-5267. From there to https://redhat.atlassian.net/browse/WFCORE-5346 and it PR https://github.com/wildfly/wildfly-legacy-test/pull/11

Really, tricky to realize it.

Those constructors are being called from (for example): https://github.com/wildfly/wildfly-legacy-test/blob/main/subsystem/35.0.0/src/main/java/org/wildfly/legacy/test/controller/subsystem_35_0_0/TestModelControllerService35_0_0.java#L87

We need to mention that the actual callers of these overloaded constructors that use these enums as discriminators are in the wildfly-legacy-test module controller implementations where a concrete
 {@code ModelTestModelControllerService} implementation is instantiated for each legacy controller under test.

yersan (changes requested):

model-test/src/main/java/org/jboss/as/model/test/ModelTestModelControllerService.java
Ohifeme: Okay, i see now
Thanks

model-test/src/main/java/org/jboss/as/model/test/ModelTestModelControllerService.java
```diff
@@ -574,8 +574,18 @@ public BootOperationFailedException(final String msg, final ModelNode descriptio
...
+     * for a given legacy WildFly Core controller version. Because Java does not allow distinguishing
+     * overloaded constructors by a plain version value alone (the signatures would be ambiguous), each
+     * supported legacy version has its own distinct marker type. Passing {@code ControllerXXx.INSTANCE}
+     * as the final constructor argument causes the compiler to resolve the intended overload unambiguously.
+     *
+     * <p>The actual callers of these overloaded constructors that use these marker classes as discriminators
+     * are in the <a href="https://github.com/wildfly/wildfly-legacy-test">wildfly/wildfly-legacy-test</a>
+     * module controller implementations, where a concrete {@link ModelTestModelControllerService}
+     * implementation is instantiated for each legacy controller under test.
+     */
```
yersan: @Ohifeme sorry, I didn't catch this before.

We should not be adding a Java doc here; the Javadoc is attached to Controller23x, but the explanation applies to all the marker classes, a comment would be more appropiated. We do not want this to appear in public API docs, it is just a hint for the developer.

So, basically, remote the /** ... **/  and use // instead. Also remove the `@link`, `@code` and all markup used in Java doc.

yersan (changes requested):

model-test/src/main/java/org/jboss/as/model/test/ModelTestModelControllerService.java
Ohifeme: Thank you Yeray, I've noted this

yersan (approved):

PR #6856, opened 2026-07-21
[WFCORE-7547] Make sasl-authentication-factory attribute mandatory in model, since its already required at runtime

baranowb:
Issue: https://redhat.atlassian.net/browse/WFCORE-7547


baranowb:
This will fail without galleon bump: https://redhat.atlassian.net/browse/WFGP-318

yersan:
@baranowb better open it as a draft with the skip-teamcity label to avoid waste resources running the CI that is expected to fail. Draft will hint the maintainers to do not worry about this since it is in progress. skip-teamcity  will hint the CI to do not run using this PR

PR #6861, opened 2026-07-23
WFCORE-7649 ResourceDefinition deprecations are always useful

pferraro:
https://redhat.atlassian.net/browse/WFCORE-7649

### DIFF
--- a/controller/src/main/java/org/jboss/as/controller/ResourceDefinition.java
+++ b/controller/src/main/java/org/jboss/as/controller/ResourceDefinition.java
@@ -199,7 +199,7 @@ public interface ResourceDefinition extends ResourceRegistration {
     static Builder builder(ResourceRegistration registration, ResourceDescriptionResolver descriptionResolver, SubsystemModel deprecation) {
         Objects.requireNonNull(registration);
         Objects.requireNonNull(descriptionResolver);
-        return new MinimalBuilder(registration, descriptionResolver, (deprecation != null) ? new DeprecationData(deprecation.getVersion()) : null);
+        return new MinimalBuilder(registration, descriptionResolver, (deprecation != null) ? new DeprecationData(deprecation.getVersion(), true) : null);
     }
 
     /**


rhusar (approved):

yersan (commented):
I also cannot image the use case where we want a management resource deprecated but without showing a warning about it, although sounds weird that this has been there since long time ago.

@bstansberry Could you figure out a use case where we would want to have this as it was without the warning?

bstansberry:
@yersan The DeprecationData class javadoc explains the use case. Logging a WARN when there is nothing the user can do about it is not particularly helpful. A user who has a mandate to eliminate WARNs has no alternative but to suppress the log level, and then they will not learn of newer things where they actually can take action.

I have no objection to changing the default behavior to warning; we just need to check whether we're going to be emitting a lot of annoying warns.

There was a period, IIRC in the early EAP 7 dev days, where a bunch of stuff got deprecated because people had a vague plan about how some day soon we would replace them. That kind of thing we should not be logging a WARN about. We should not have deprecated those things at all! I think that kind of thing is mostly gone by now. I know you helped get rid of some of them earlier this year. :-)

bstansberry (approved):
Regarding checking for what would be affected by a change in default behavior, there's not much:

```
Method
    builder(ResourceRegistration, ResourceDescriptionResolver, SubsystemModel)
Usages or usages of base method in Project and Libraries  (6 usages found)
    Unclassified  (6 usages found)
        wildfly-clustering-infinispan-extension  (3 usages found)
            org.jboss.as.clustering.infinispan.subsystem  (2 usages found)
                CacheResourceDefinitionRegistrar  (1 usage found)
                    register(ManagementResourceRegistration, ManagementResourceRegistrationContext)  (1 usage found)
                        115 ManagementResourceRegistration registration = parent.registerSubModel(ResourceDefinition.builder(this.registration, resolver, this.registration.getDeprecation()).build());
                ConfigurationResourceDefinitionRegistrar  (1 usage found)
                    register(ManagementResourceRegistration, ManagementResourceRegistrationContext)  (1 usage found)
                        63 ResourceDefinition definition = ResourceDefinition.builder(this.registration, resolver, this.deprecation).build();
            org.jboss.as.clustering.infinispan.subsystem.remote  (1 usage found)
                ConnectionPoolResourceDefinitionRegistrar  (1 usage found)
                    register(ManagementResourceRegistration, ManagementResourceRegistrationContext)  (1 usage found)
                        80 ResourceDefinition definition = ResourceDefinition.builder(RemoteComponentResourceRegistration.CONNECTION_POOL, resolver, InfinispanSubsystemModel.VERSION_20_0_0).build();
        wildfly-clustering-jgroups-extension  (1 usage found)
            org.jboss.as.clustering.jgroups.subsystem  (1 usage found)
                ProtocolChildResourceDefinitionRegistrar  (1 usage found)
                    register(ManagementResourceRegistration, ManagementResourceRegistrationContext)  (1 usage found)
                        73 getResourceDefinitionConfigurator().apply(ResourceDefinition.builder(this.configurator.getResourceRegistration(), resolver, this.configurator.getDeprecation())).build(
        wildfly-undertow  (1 usage found)
            org.wildfly.extension.undertow  (1 usage found)
                HostSingleSignOnDefinition  (1 usage found)
                    HostSingleSignOnDefinition()  (1 usage found)
                        37 return ResourceDefinition.builder(registration, resolver, UndertowSubsystemModel.VERSION_12_0_0);
        Maven: org.wildfly.core:wildfly-controller:34.0.0.Beta1  (1 usage found)
            org.jboss.as.controller  (1 usage found)
                ResourceDefinition  (1 usage found)
                    builder(ResourceRegistration, ResourceDescriptionResolver)  (1 usage found)
                        188 return builder(registration, descriptionResolver, null);
```

Paul filed this and Rado approved so I'll just assume the clustering ones are ok. (A quick poke at a couple makes me think they pass in null anyway.) 

HostSingleSignOnDefinition has been deprecated since WFLY-15732, has no Stage.RUNTIME function, and relates to legacy picketbox security. So any use of it should log a WARN.

bstansberry:
Thanks @pferraro, @rhusar and @yersan

PR #6862, opened 2026-07-24
[WFCORE-7653] standalone.bat --debug (without port) falls through to READ-VERSION, causing OOM and boot failure

lvydra:
https://redhat.atlassian.net/browse/WFCORE-7653

### DIFF
--- a/testsuite/scripts/src/test/java/org/wildfly/scripts/test/StandaloneDebugScriptTestCase.java
+++ b/testsuite/scripts/src/test/java/org/wildfly/scripts/test/StandaloneDebugScriptTestCase.java
@@ -1,0 +1,62 @@
+/*
+ * Copyright The WildFly Authors
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+package org.wildfly.scripts.test;
+
+import java.io.IOException;
+import java.util.ArrayList;
+import java.util.Arrays;
+import java.util.Collection;
+import java.util.List;
+import java.util.concurrent.TimeoutException;
+import java.util.function.Function;
+
+import org.jboss.as.controller.client.ModelControllerClient;
+import org.junit.Assert;
+import org.junit.runner.RunWith;
+import org.junit.runners.Parameterized;
+import org.junit.runners.Parameterized.Parameter;
+import org.junit.runners.Parameterized.Parameters;
+import org.wildfly.common.test.ServerHelper;
+
+@RunWith(Parameterized.class)
+public class StandaloneDebugScriptTestCase extends ScriptTestCase {
+    private static final Function<ModelControllerClient, Boolean> STANDALONE_CHECK = ServerHelper::isStandaloneRunning;
+    private static final String DEBUG_LISTEN_PREFIX = "Listening for transport dt_socket at address: ";
+
+    @Parameter(0)
+    public String[] debugArgs;
+
+    @Parameter(1)
+    public String expectedPort;
+
+    public StandaloneDebugScriptTestCase() {
+        super("standalone");
+    }
+
+    @Parameters(name = "{0} -> port {1}")
+    public static Collection<Object[]> data() {
+        return List.of(
+                new Object[] {new String[] {"--debug"}, "8787"},
+                new Object[] {new String[] {"--debug", "9797"}, "9797"},
+                new Object[] {new String[] {"-Dtest=test", "--debug"}, "8787"},
+                new Object[] {new String[] {"-Dtest=test", "--debug", "9797"}, "9797"}
+        );
+    }
+
+    @Override
+    void testScript(final ScriptProcess script) throws InterruptedException, TimeoutException, IOException {
+        final List<String> args = new ArrayList<>(Arrays.asList(ServerHelper.DEFAULT_SERVER_JAVA_OPTS));
+        args.addAll(Arrays.asList(debugArgs));
+        script.start(STANDALONE_CHECK, args.toArray(String[]::new));
+
+        Assert.assertTrue("The process is not running and should be", script.isAlive());
+
+        final var stdout = script.getStdoutAsString();
+        final String expectedMessage = DEBUG_LISTEN_PREFIX + expectedPort;
+        Assert.assertTrue("Expected to find '" + expectedMessage + "' for a server started with " + script.getLastExecutedCmd() + "\nThe server output was:\n" + stdout, stdout.contains(expectedMessage));
+    }
+}
+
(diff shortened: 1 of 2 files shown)


yersan:
Ideally, we should have a test case for this under testsuite/scripts folder

yersan:
@lvydra in addition of the test case, we should also backport this 33.x branch, let me know if you do not plan to work on the test case so we can take a look.

lvydra:
@yersan Thanks, I have added test case.

testsuite/scripts/src/test/java/org/wildfly/scripts/test/StandaloneDebugScriptTestCase.java
```diff
@@ -0,0 +1,53 @@
...
+        super("standalone");
+    }
+
+    @Override
+    void testScript(final ScriptProcess script) throws InterruptedException, TimeoutException, IOException {
+        final List<String> args = new ArrayList<>(Arrays.asList(ServerHelper.DEFAULT_SERVER_JAVA_OPTS));
+        args.add("--debug");
+        script.start(STANDALONE_CHECK, MAVEN_JAVA_OPTS, args.toArray(String[]::new));
+
+        Assert.assertNotNull("The process is null and may have failed to start.", script);
```
yersan: I've noticed script will be always not null at this point, otherwise it will fail the line before

testsuite/scripts/src/test/java/org/wildfly/scripts/test/StandaloneDebugScriptTestCase.java
```diff
@@ -0,0 +1,53 @@
...
+        Assert.assertTrue("The process is not running and should be", script.isAlive());
+
+        final var stdout = script.getStdoutAsString();
+        Assert.assertFalse("Found -Xmx16m in JVM parameters, for a server started with " + script.getLastExecutedCmd() + "\nThe server output was:\n" + stdout, stdout.contains("-Xmx16m"));
```
yersan: I would say that what we should assert is that the server was started in debug mode, hence that the output contains

`Listening for transport dt_socket at address: `

testsuite/scripts/src/test/java/org/wildfly/scripts/test/StandaloneDebugScriptTestCase.java
```diff
@@ -0,0 +1,53 @@
...
+        final Callable<ModelNode> callable = new Callable<>() {
+            @Override
+            public ModelNode call() throws Exception {
+                try (ModelControllerClient client = TestSuiteEnvironment.getModelControllerClient()) {
+                    return executeOperation(client, Operations.createOperation("shutdown"));
+                }
+            }
+        };
+        execute(callable);
+        validateProcess(script);
```
yersan: This is not necessary, the ScriptProcess implements AutoCloseable and the ScriptTestCase will take care of closing it.

I think it could be harmless, but completely unnecessary, I would vote for removing it and avoid any headaches for the future, but I let you the decision @lvydra 

Thanks for adding the test, those scripts have demonstrated to be no very developer friendly and had lot of details, so the more test coverage, the better

yersan (changes requested):

testsuite/scripts/src/test/java/org/wildfly/scripts/test/StandaloneDebugScriptTestCase.java
lvydra: @yersan Thanks for looking at it, test should be updated accordingly now.

testsuite/scripts/src/test/java/org/wildfly/scripts/test/StandaloneDebugScriptTestCase.java
```diff
@@ -0,0 +1,37 @@
...
+    private static final Function<ModelControllerClient, Boolean> STANDALONE_CHECK = ServerHelper::isStandaloneRunning;
+
+    public StandaloneDebugScriptTestCase() {
+        super("standalone");
+    }
+
+    @Override
+    void testScript(final ScriptProcess script) throws InterruptedException, TimeoutException, IOException {
+        final List<String> args = new ArrayList<>(Arrays.asList(ServerHelper.DEFAULT_SERVER_JAVA_OPTS));
+        args.add("--debug");
```
yersan: @lvydra something more, to get a better coverage, would you mind adding additional variants using a specific port, and using an additional property after specifying the debug argument? I mean, additional tests cases where the script is started with:

 --debug --> debug at port 8787, already verified but without explicitly reading the port number
 --debug 9797 --> debug at port 9797
 --debug -Dtest=test --> debug at port 8787, pass -Dtest=test to server
 --debug 9797 -Dtest=test --> debug at port 9797, pass -Dtest=test to server
  
What do you think?, we don't need to assert that the property test is visible to the server JVM, I mean the server can resolve it; just with STANDALONE_CHECK we will assert the server started. For the debug port, we can easily assert its value extending the "Listening for transport dt_socket at address:" value.

I've checked the standalone.bat code and looks it will work fine with those combinations, so this will be only to have more coverage and do not necessarily be required for this PR, but better if we do all in one commit.

yersan: @lvydra This PR is enough as it is, since we need to move on with downstreams if you think we will need more time, we can backport this as it is now and create a follow up Jira to add additional coverage. The point is to do not block downstreams unnecessarily.

We would not need to create different tests classes, we would need to start the test with @Parameters to supply the different arguments we want to use to start the server.

So the question more for you and based on your availability, do you prefer to have a follow up Jira or do you expect to increase the coverage quickly so it can be backported to downstreams soon?

lvydra: @yersan I think I will be able to add it quickly

lvydra: @yersan Updated

yersan: That's great!, thanks @lvydra , let's wait for a CI cycle

yersan (approved):

yersan:
Thanks @lvydra , could you backport this to 33.x branch?

lvydra:
@yersan Thanks, sure, to wildfly-core or wildfly-core-eap only?

PR #6872, opened 2026-07-30
[WFCORE-7663] Update the ParameterCorrector javadoc

bstansberry:
General cleanup to reflect that the primary relevant concept is operation parameter values, and not the attributes where such values often end up being stored.

Add a discussion of how expression values are handled.

Add example use cases

Add discussion of the need to confirm nodes meet expectations before correcting and how to handle any that don't

https://redhat.atlassian.net/browse/WFCORE-7663


yersan (approved):

yersan:
Thanks @bstansberry

PR #6873, opened 2026-07-30
WFCORE-7664 Add Dependency.thenAccept(...) methods to simplify ServiceInstaller composition

pferraro:
https://redhat.atlassian.net/browse/WFCORE-7664

### DIFF
--- a/service/src/main/java/org/wildfly/service/Dependency.java
+++ b/service/src/main/java/org/wildfly/service/Dependency.java
@@ -5,6 +5,7 @@
 package org.wildfly.service;
 
 import java.util.Objects;
+import java.util.function.BiConsumer;
 import java.util.function.BiFunction;
 import java.util.function.Consumer;
 import java.util.function.Function;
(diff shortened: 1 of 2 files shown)


service/src/main/java/org/wildfly/service/Dependency.java
```diff
@@ -85,6 +86,35 @@ public R get() {
         };
     }
 
+    /**
+     * Returns a runnable task that consumes the provided dependency value.
+     * @param consumer a consumer of the dependency value.
+     * @return a runnable task that consumes the provided dependency value.
+     */
+    default Runnable thenAccept(Consumer<? super V> consumer) {
+        return new Runnable() {
```
bstansberry, line 95: Why a Runnable? The BlockingLifecycle.compose use case in the Jira seems to need a Consumer.

pferraro: Why not?  Runnable is a fairly common parameter type.  In particular, Installer.Builder itself has a number of methods for describing service behaviour via a Runnable.  I should think any opportunity to consolidate redundant anonymous classes would be worthwhile.

bstansberry: @pferraro I don't have any intrinsic objection to Runnable. It's just that AFAICT the example use case coded out in the JIRA description will not work because there is no BlockingLifecycle.compose method that takes Runnable as a param type. And, presuming the 'Server' in the example is org.wildfly.extension.undertow.Server, the Server::registerHost and Server:unregisterHost methods are not method references to BiFunctions, so the  other new 'thenAccept' in this PR is not relevant.

So, this PR might be fine but in that case the JIRA description should be updated to some example that better relates to the actual fix.

pferraro: Added requested example to jira description.

yersan: @pferraro I still do not understand how `thenAccept` returning a Runneable is used in the Jira example:

```
ServiceDependency<Parent> parent = …;
Runnable refresh = parent.compose(Parent::refresh);
Supplier<Child> factory = …;
ServiceInstaller.BlockingBuilder.of(factory).requires(parent).onStart(refresh).onStop(refresh).build().install(context);
```

Where is `thenAccept` returnning a Runneable being used in that example?

pferraro: The jira description should have read:
```java
Runnable refresh = parent.thenAccept(Parent::refresh);
```

yersan (approved):

PR #6881, opened 2026-08-10
[WFCORE-7668] ParameterArgumentValidationErrorEmptyStringNotAllowed on domain.ps1

lvydra:
https://redhat.atlassian.net/browse/WFCORE-7668

### DIFF
--- a/testsuite/scripts/src/test/java/org/wildfly/scripts/test/DomainHelpVersionScriptTestCase.java
+++ b/testsuite/scripts/src/test/java/org/wildfly/scripts/test/DomainHelpVersionScriptTestCase.java
@@ -48,5 +48,10 @@ public class DomainHelpVersionScriptTestCase extends ScriptTestCase {
         boolean ok = stdout.contains("-Xmx16m") && !stdout.replaceFirst("-Xmx", "").contains("-Xmx");
         Assert.assertTrue("Expected to find -Xmx16m in the JVM parameters for a server started with " + script.getLastExecutedCmd() + "\nThe server output was: \n" + stdout, ok);
         Assert.assertFalse("Did not expect to find gc.log in the JVM parameters for a server started with " + script.getLastExecutedCmd() + "\nThe server output was: \n" + stdout, stdout.contains("gc.log"));
+        if ("--help".equals(arg) || "-h".equals(arg)) {
+            Assert.assertTrue("Expected help output containing 'Usage:' for " + script.getLastExecutedCmd() + "\nThe server output was: \n" + stdout, stdout.contains("Usage:"));
+        } else if ("-v".equals(arg) || "-V".equals(arg) || "--version".equals(arg)) {
+            Assert.assertTrue("Expected version output containing 'WildFly Core' for " + script.getLastExecutedCmd() + "\nThe server output was: \n" + stdout, stdout.contains("WildFly Core"));
+        }
     }
 }
--- a/testsuite/scripts/src/test/java/org/wildfly/scripts/test/StandaloneHelpVersionScriptTestCase.java
+++ b/testsuite/scripts/src/test/java/org/wildfly/scripts/test/StandaloneHelpVersionScriptTestCase.java
@@ -48,5 +48,10 @@ public class StandaloneHelpVersionScriptTestCase extends ScriptTestCase {
         boolean ok = stdout.contains("-Xmx16m") && !stdout.replaceFirst("-Xmx", "").contains("-Xmx");
         Assert.assertTrue("Expected to find -Xmx16m in the JVM parameters for a server started with " + script.getLastExecutedCmd() + "\nThe server output was: \n" + stdout, ok);
         Assert.assertFalse("Did not expect to find gc.log in the JVM parameters for a server started with " + script.getLastExecutedCmd() + "\nThe server output was: \n" + stdout, stdout.contains("gc.log"));
+        if ("--help".equals(arg) || "-h".equals(arg)) {
+            Assert.assertTrue("Expected help output containing 'Usage:' for " + script.getLastExecutedCmd() + "\nThe server output was: \n" + stdout, stdout.contains("Usage:"));
+        } else if ("-v".equals(arg) || "-V".equals(arg) || "--version".equals(arg)) {
+            Assert.assertTrue("Expected version output containing 'WildFly Core' for " + script.getLastExecutedCmd() + "\nThe server output was: \n" + stdout, stdout.contains("WildFly Core"));
+        }
     }
 }


core-feature-pack/common/src/main/resources/content/bin/domain.ps1
```diff
@@ -108,7 +108,9 @@ $PROG_ARGS += "$JBOSS_HOME"
 $PROG_ARGS += "--"  
 $PROG_ARGS += "-Dorg.jboss.boot.log.file=$JBOSS_LOG_DIR\host-controller.log"
 $PROG_ARGS += "-Dlogging.configuration=file:$JBOSS_CONFIG_DIR\logging.properties"
-$PROG_ARGS += $HOST_CONTROLLER_JAVA_OPTS
+if ($HOST_CONTROLLER_JAVA_OPTS -ne $null){
```
yersan: @lvydra do we know exactly where the HOST_CONTROLLER_JAVA_OPTS is initialized to null when using --version/--help or its variants?

lvydra: @yersan Originally, HOST_CONTROLLER JAVA_OPTS is initialized at https://github.com/wildfly/wildfly-core/blob/31551b88ba900abacf628e2aa05d031192e31155/core-feature-pack/common/src/main/resources/content/bin/domain.conf.ps1#L84, but after changes, when the script is run using --version/--help, initialization is skipped https://github.com/wildfly/wildfly-core/blob/31551b88ba900abacf628e2aa05d031192e31155/core-feature-pack/common/src/main/resources/content/bin/domain.ps1#L13, and it remains undeclared, implicitly null.

yersan: Ok, so PROCESS_CONTROLLER_JAVA_OPTS has the same problem. inst't it?

I suggest to initialize them at https://github.com/wildfly/wildfly-core/blob/main/core-feature-pack/common/src/main/resources/content/bin/domain.ps1#L23-L28 with the following:

```
  if ((-Not(Test-Path variable:PROCESS_CONTROLLER_JAVA_OPTS)) -or (-Not($PROCESS_CONTROLLER_JAVA_OPTS))) {
      $PROCESS_CONTROLLER_JAVA_OPTS = Get-Java-Opts
  }
  if ((-Not(Test-Path variable:HOST_CONTROLLER_JAVA_OPTS)) -or (-Not($HOST_CONTROLLER_JAVA_OPTS))) {
      $HOST_CONTROLLER_JAVA_OPTS = Get-Java-Opts
  }
```

I would expect them to be initialized with the existing Java options

yersan (changes requested):
@lvydra  scripts are failing with this change, could you take a closer look?

lvydra:
@yersan  Sure, I will look at it.

core-feature-pack/common/src/main/resources/content/bin/domain.ps1
lvydra: @yersan I have rebased the commit, and it seems that the changes in https://github.com/wildfly/wildfly-core/pull/6884/changes/e6ee67965ddd9717a572f83dbec02c886069e053 are enough to resolve the issue. I have left an update of the tests in the commit to verify that.

yersan (commented):
Weird, it looks like those test are already silently failing, correct me if I am wrong @lvydra I assume you are looking closely to them

lvydra:
@yersan Yes, that's true. There is only a check for the expected arguments in the JAVA_OPTS output and not for successful execution.

core-feature-pack/common/src/main/resources/content/bin/domain.ps1
yersan: ok @lvydra , resolving this

yersan:
> @yersan Yes, that's true. There is only a check for the expected arguments in the JAVA_OPTS output and not for successful execution.

Hi @lvydra , just catching up with the status. Last CI run shows the Jobs failing. Have you been able to check this and verify what is going wrong with them?

lvydra:
> > @yersan Yes, that's true. There is only a check for the expected arguments in the JAVA_OPTS output and not for successful execution.
> 
> Hi @lvydra , just catching up with the status. Last CI run shows the Jobs failing. Have you been able to check this and verify what is going wrong with them?

Hi @yersan, it seems to be caused by the JAVA_OPTS rewrite on:
https://github.com/wildfly/wildfly-core/blob/40eab6f84ee42ea33eb7e8a6e5c0ee6c28f65463/core-feature-pack/common/src/main/resources/content/bin/standalone.bat#L87
thus not propagating _maven.repo.local_:
https://github.com/lvydra/wildfly-core/blob/cbac858c6f5431a5ef0164d6bf9db3bfb61b740e/testsuite/scripts/src/test/java/org/wildfly/scripts/test/ScriptTestCase.java#L64 

I will try to verify that by reproducing it locally.

lvydra:
@yersan I have bypassed maven.repo.local parameter propagation via JAVA_OPTS for batch scripts, and the tests are passing now. 

https://github.com/wildfly/wildfly-core/blob/40eab6f84ee42ea33eb7e8a6e5c0ee6c28f65463/core-feature-pack/common/src/main/resources/content/bin/standalone.bat#L87 
JAVA_OPTS overwrite prevents _maven.repo.local_ from being passed. 

This is probably silently failing now.

core-feature-pack/common/src/main/resources/content/bin/standalone.bat
```diff
@@ -84,7 +84,11 @@ shift
 goto READ-ARGS
 
 :READ-VERSION
-set "JAVA_OPTS=-Xmx16m"
+if not "x%MAVEN_REPO_LOCAL%" == "x" (
+    set "JAVA_OPTS=-Xmx16m -Dmaven.repo.local=%MAVEN_REPO_LOCAL%"
+) else (
+    set "JAVA_OPTS=-Xmx16m"
+)
```
yersan: We should not circumbent test cases issues modifying end user files.

In addition to that, the current issue should be affecting to all the scripts, not only to the .bat ones.

testsuite/scripts/src/test/java/org/wildfly/scripts/test/DomainHelpVersionScriptTestCase.java
```diff
@@ -39,7 +39,12 @@ public static Collection<Object> data() {
 
     @Override
     void testScript(final ScriptProcess script) throws InterruptedException, TimeoutException, IOException {
-        script.start(null, MAVEN_JAVA_OPTS, new String [] {arg});
+        final String localRepo = System.getProperty("maven.repo.local");
+        final java.util.Map<String, String> env = new java.util.HashMap<>();
+        if (localRepo != null) {
+            env.put("MAVEN_REPO_LOCAL", localRepo);
+        }
+        script.start(null, env, new String [] {arg});
```
yersan: @lvydra The MAVEN_JAVA_OPTS should be injected into the JAVA_OPTS environment variable before the scripts are launched.

Why this is not working for the .bat scripts?

yersan (changes requested):
@lvydra thanks for continue digging into this, but we need a different solution here.

ok so the problem is that with the version/help arguments, we are restricting the usage of the JAVA_OPTS from the outside, hence make the tests to silently fail when the server is built in a non default maven local repo environment, since the `maven.repo.local` should be made visible to JBoss Modules. 

For these tests only, we should use JDK_JAVA_OPTIONS  to set the maven local repo if it is available and do not modify the launch scripts; the content of that variable will be automatically visible to JVM launched via the scripts.

core-feature-pack/common/src/main/resources/content/bin/standalone.bat
lvydra: @yersan Thanks for the reply and suggestions. The workaround was meant more as proof of the JAVA_OPTS rewriting. Now it should be more aligned with what you have suggested.

testsuite/scripts/src/test/java/org/wildfly/scripts/test/ScriptTestCase.java
```diff
@@ -62,6 +63,7 @@ public static void configureEnvironment() throws Exception {
         final String localRepo = System.getProperty("maven.repo.local");
         if (localRepo != null) {
             MAVEN_JAVA_OPTS.put("JAVA_OPTS", "-Dmaven.repo.local=" + localRepo);
+            JDK_JAVA_OPTIONS.put("JDK_JAVA_OPTIONS", "-Dmaven.repo.local=" + localRepo);
```
yersan, line 66: @lvydra The fix should be only applied to the help/version tests, see my previous review. Do not modify the  ScriptTestCase test, this is the base class used by all the tests. We want to use the JDK_JAVA_OPTIONS environment variable approach only for the help/Version tests.

The reasoning behind this is that those are the only tests that test launch scripts that don't process the JAVA_OPTS arguments specified by the user for their configuration. We want to continue testing JAVA_OPTS usage on the other launch scripts though. If we add JDK_JAVA_OPTIONS here, we won't know if the other launch scripts have any issues processing the JAVA_OPTS. 

Just calculate the JDK_JAVA_OPTIONS value on the Help/Version tests and pass it as environment variable from there, thanks!

yersan (changes requested):

PR #6884, opened 2026-08-12
[WFCORE-7670] Add deny-list of well-known gadget chains to default jdk.serialFilter system property

bstansberry:
https://redhat.atlassian.net/browse/WFCORE-7670

I assume there will be full WF integration test failures until https://github.com/wildfly/wildfly/pull/20322 is resolved.


yersan:
> I assume there will be full WF integration test failures until https://github.com/wildfly/wildfly/pull/20322 is resolved.

Let's see how it behaves with:
https://ci.wildfly.org/buildConfiguration/WF_WildFlyCoreIntegrationExperimentsWindows/577895

yersan (commented):
Integration Job failed, there are still issues with the powersshell scripts that are not yet resolved

```
java.lang.AssertionError:
11:51:17   Expected 2 lines
11:51:17   Command Executed:
11:51:17   powershell -ExecutionPolicy Unrestricted -NonInteractive -File C:\BuildAgent\work\6e61adf388bc437e\full\testsuite\scripts\target\wildfly\bin\appclient.ps1 "-v"
11:51:17   Environment:
11:51:17   Output:
11:51:17   Exception in thread "main" org.jboss.modules.ModuleLoadException: Error loading module from C:\BuildAgent\work\6e61adf388bc437e\full\testsuite\scripts\target\wildfly\modules\system\layers\base\org\jboss\as\appclient\main\module.xml
11:51:17     at org.jboss.modules.xml.ModuleXmlParser.parseModuleXml(ModuleXmlParser.java:336)
11:51:17     at org.jboss.modules.xml.ModuleXmlParser.parseModuleXml(ModuleXmlParser.java:292)
11:51:17     at org.jboss.modules.xml.ModuleXmlParser.parseModuleXml(ModuleXmlParser.java:253)
11:51:17     at org.jboss.modules.LocalModuleFinder.parseModuleXmlFile(LocalModuleFinder.java:250)
11:51:17     at org.jboss.modules.LocalModuleFinder.lambda$findModule$1(LocalModuleFinder.java:195)
11:51:17     at java.base/java.security.AccessController.doPrivileged(AccessController.java:712)
11:51:17     at org.jboss.modules.LocalModuleFinder.findModule(LocalModuleFinder.java:195)
11:51:17     at org.jboss.modules.ModuleLoader.findModule0(ModuleLoader.java:700)
11:51:17     at org.jboss.modules.ModuleLoader.findModule(ModuleLoader.java:693)
11:51:17     at org.jboss.modules.ModuleLoader.loadModuleLocal(ModuleLoader.java:503)
11:51:17     at org.jboss.modules.DelegatingModuleLoader.preloadModule(DelegatingModuleLoader.java:57)
11:51:17     at org.jboss.modules.ModuleLoader.loadModule(ModuleLoader.java:299)
11:51:17     at org.jboss.modules.Main.main(Main.java:361)
11:51:17   Caused by: org.jboss.modules.xml.XmlPullParserException: Failed to resolve artifact 'org.wildfly:wildfly-appclient:42.0.0.Beta1-SNAPSHOT' (position: END_TAG seen ...ifact name="org.wildfly:wildfly-appclient:42.0.0.Beta1-SNAPSHOT"/>... @20:79)
11:51:17     at org.jboss.modules.xml.ModuleXmlParser.parseArtifact(ModuleXmlParser.java:1049)
11:51:17     at org.jboss.modules.xml.ModuleXmlParser.parseResources(ModuleXmlParser.java:936)
11:51:17     at org.jboss.modules.xml.ModuleXmlParser.parseModuleContents(ModuleXmlParser.java:696)
11:51:17     at org.jboss.modules.xml.ModuleXmlParser.parseDocument(ModuleXmlParser.java:468)
11:51:17     at org.jboss.modules.xml.ModuleXmlParser.parseModuleXml(ModuleXmlParser.java:334)
11:51:17     ... 12 more
```

JBoss Modules is unable to resolve the dependency likely due to a wrong local maven repo configuration. Will take a closer look

core-feature-pack/common/src/main/resources/content/bin/domain.bat
```diff
@@ -217,6 +210,22 @@ if "%SECMGR%" == "true" (
     set "MODULE_OPTS=-secmgr"
 )
 
+rem Add -Djdk.serialFilter if not specified
+echo "%JAVA_OPTS% %SERVER_OPTS% %JDK_JAVA_OPTIONS%" | findstr /I "\-Djdk.serialFilter" > nul
+if errorlevel == 1 (
```
rhusar, line 215: This is invalid syntax IIUC

See https://learn.microsoft.com/en-us/windows-server/administration/windows-commands/if

Correct is:

```suggestion
if ERRORLEVEL 1 (
```

rhusar (changes requested):

bootable-jar/runtime/src/main/java/org/wildfly/core/jar/runtime/Arguments.java
```diff
@@ -128,7 +131,15 @@ private void handleArguments(List<String> args) throws Exception {
...
                 if (filterEnabled) {
                     if (filter == null) {
-                        filter = DEFAULT_SERIAL_FILTER;
+                        URL defaultFilterURL = loader.getResource(DEFAULT_JDK_SERIAL_FILTER_RESOURCE);
+                        if (defaultFilterURL != null) {
+                            String defaultFilterParameter = Files.readString(Path.of(defaultFilterURL.toURI()));
+                            Matcher matcher = DEFAULT_JDK_SERIAL_FILTER_PATTERN.matcher(defaultFilterParameter);
+                            if (matcher.find()) {
+                                // Join line continuations
+                                filter = matcher.group(1).replaceAll("\\\\\\v\\h*", "");
```
rhusar, line 140: n.b. if a `filter` is never found there is no logging, fails silently. Supposedly that should never happen so I think that is fine.

core-feature-pack/common/src/main/resources/content/bin/domain.conf.ps1
```diff
@@ -27,10 +27,17 @@ if (-Not(test-path env:JBOSS_MODULES_SYSTEM_PKGS )) {
...
-    $JAVA_OPTS = @()
+if (Test-Path env:JAVA_OPTS) {
+    $opts = $env:JAVA_OPTS.split()
+    ForEach ($opt in $opts) {
+        $JAVA_OPTS += $opt
+    }
+}
+
+# Set default values if none have been set by the user
+if ((!$JAVA_OPTS) -or (!$PRESERVE_JAVA_OPTS)) {
```
rhusar, line 40: The condition does not seem correct here "if java opts are not set OR we were asked not to preserve them then override them".

This should rather be: "unless java opts are set and we were asked to preserve them then override them":

Fix is:

```suggestion
if (!($JAVA_OPTS -or $PRESERVE_JAVA_OPTS)) {
```

rhusar (changes requested):

core-feature-pack/common/src/main/resources/content/bin/domain.conf.ps1
rhusar: Example "JAVA_OPTS=-Xmx2g PRESERVE_JAVA_OPTS=false (or unset)" => default's aren't applied.

core-feature-pack/common/src/main/resources/content/bin/common.bat
```diff
@@ -39,45 +39,45 @@ goto :eof
...
 :setDefaultModularJvmOptions
   call :setModularJdk
-  if "!MODULAR_JDK!" == "true" (
+  if "%MODULAR_JDK%" == "true" (
     echo %* | findstr /I "\-\-add\-modules" > nul
     if errorlevel == 1 (
       rem Set default modular jdk options
       rem Needed by the iiop-openjdk subsystem
-      set "DEFAULT_MODULAR_JVM_OPTIONS=!DEFAULT_MODULAR_JVM_OPTIONS! --add-exports=java.desktop/sun.awt=ALL-UNNAMED"
+      set "DEFAULT_MODULAR_JVM_OPTIONS=%DEFAULT_MODULAR_JVM_OPTIONS% --add-exports=java.desktop/sun.awt=ALL-UNNAMED"
```
rhusar: Replacing !DEFAULT_MODULAR_JVM_OPTIONS! with %DEFAULT_MODULAR_JVM_OPTIONS% inside a parenthesized if (...) block means all 14 accumulating set lines expand at parse time and each one overwrites the previous, so the variable ends up holding only `--add-opens=java.naming/javax.naming=ALL-UNNAMED` as seen on server boot (in other branch)

```
===============================================================================
  JBoss Bootstrap Environment
  JBOSS_HOME: "C:\path\to\wf"
  JAVA: "C:\path\to\java\bin\java"
  JAVA_OPTS: "-Dprogram.name=standalone.bat -Xms1G -Xmx1G -Djava.net.preferIPv4Stack=true -Djboss.modules.system.pkgs=org.jboss.byteman -Djava.awt.headless=true "-Xlog:gc*:file=C:\path\to\wf\standalone\log\gc.log:time,uptimemillis:filecount=5,filesize=3M"  
--add-opens=java.naming/javax.naming=ALL-UNNAMED "@C:\path\to\bin\jdk.serialFilter" "
```

rhusar (changes requested):

bstansberry:
> Integration Job failed, there are still issues with the powersshell scripts that are not yet resolved

@yersan Please note that https://github.com/wildfly/wildfly/pull/20322 is unmerged and touches the scripts that related to the failures in the "WildFly Core -> Full Integration - Windows - JDK 17" job.

yersan:
@bstansberry look at https://ci.wildfly.org/buildConfiguration/WF_WildFlyCoreIntegrationExperimentsWindows/577895, it uses your topic branches for core and wildfly

core-feature-pack/common/src/main/resources/content/bin/domain.conf.ps1
yersan: The `PRESERVE_JAVA_OPTS` is a flag to indicate to the scripts whether they are allowed to manipulate the user `JAVA_OPTS`. 

The tricky case is what the script should do if user has set JAVA_OPTS, but PRESERVE_JAVA_OPTS is not set at all or if it is false, should the user JAVA_OPTS be merged with the ones added by the script or should we just preservate the JAVA_OPTS even if PRESERVE_JAVA_OPTS is not set? .... I need to check what other scripts (bat, sh, do in this case)

yersan: Looks clear to me that `PRESERVE_JAVA_OPTS` should not be in the equation at the .conf.ps1 level for both domain and standalone.

That looks matching with what .sh, and .bat do at:

https://github.com/wildfly/wildfly-core/blob/0aca5b842bda91265168f210e422b2f32571dd25/core-feature-pack/common/src/main/resources/content/bin/standalone.conf#L59-L64

https://github.com/wildfly/wildfly-core/blob/0aca5b842bda91265168f210e422b2f32571dd25/core-feature-pack/common/src/main/resources/content/bin/standalone.conf.bat#L18-L21

core-feature-pack/common/src/main/resources/content/bin/domain.bat
yersan: I agree with this. Seems that's an existing bug since it is already present in other parts even before these PR. I would suggest to track it in a different Jira

rhusar: OK agreed - though i now assume this really checks whether String equal i.e. "errorlevel" == "1" which would have alway been false, odd.

rhusar: The condition always evaluates to FALSE and will never execute the code inside the block, regardless of the actual exit code, so how come this worked..

yersan: Will be tracked on https://redhat.atlassian.net/browse/WFCORE-7673

core-feature-pack/common/src/main/resources/content/bin/common.bat
rhusar: Opened https://redhat.atlassian.net/browse/WFCORE-7674

core-feature-pack/common/src/main/resources/content/bin/common.ps1
```diff
@@ -58,18 +58,18 @@ Function Get-String {
 }
 
 Function String-To-Array($value) {
-  $res = @()
-  if (!$value){
```
rhusar, line 62: Hidden in the refactoring, the function for filtering out empty spaces, was removed....

Edit: only added commas.

rhusar:
Opened https://github.com/bstansberry/wildfly-core/pull/10 against this branch to add fix for https://redhat.atlassian.net/browse/WFCORE-7674.

core-feature-pack/common/src/main/resources/content/bin/common.bat
rhusar: Opened PR https://github.com/bstansberry/wildfly-core/pull/10

core-feature-pack/common/src/main/resources/content/bin/common.ps1
bstansberry: @rhusar I don't understand.

yersan: @rhusar / @bstansberry the missing change I guess is needed here:

```
  foreach ($element in $value.split()) {
        $result += $element
   }
```

Should be 

```
  foreach ($element in $value.split()) {
        if ($element) {
              $result += $element
        }
   }
```

Otherwise, when the input is just `" "`, `" ".split()` will produce two empty elements, so `$result` will be the following array `{"",""}`

Rado spotted that this was introduced with the latest changes on ps1 scripts added in this PR, so we would like them fixed.

If it is easier for tracking / merging purposes we can use a different Jira. I would avoid the paperwork by just changing it here

core-feature-pack/common/src/main/resources/content/bin/common.bat
yersan: This was already merged in this topic branch, so resolving this

core-feature-pack/common/src/main/resources/content/bin/common.ps1
rhusar: Sorry @bstansberry I should have linked https://github.com/wildfly/wildfly/pull/20330#pullrequestreview-4920299584 - I didn't realize it wouldn't reference.

bstansberry: ok, thanks. I'll add a commit

bstansberry: @rhusar @yersan I added the commit.

bstansberry:
Next I'll send up a 33.x variant of this.

yersan (approved):

yersan:
@rhusar I think we are ready for merging this, your concerns about the missing check in the string to array function was already fixed, sure we still have things to do with to properly review the possibilities of returning an string when we wanted an array, but as discussed in a different Jira we want to change that in different PR after merging this

Could you approve from your side if you are ok with this?

yersan:
I've realized Rado is not available this week, since his concern was already addressed, I'm moving on merging this.

yersan:
thanks @bstansberry @rhusar

rhusar:
> I've realized Rado is not available this week, since his concern was already addressed, I'm moving on merging this.

Yeah @yersan good call not waiting for me on this :) LGTM!

PR #6886, opened 2026-08-12
[WFCORE-7671] Add boot errors and expected input args checks to Domai…

bstansberry:
…nScriptTestCase and StandaloneScriptTestCase

https://redhat.atlassian.net/browse/WFCORE-7671

### DIFF
--- a/testsuite/scripts/src/test/java/org/wildfly/common/test/ServerHelper.java
+++ b/testsuite/scripts/src/test/java/org/wildfly/common/test/ServerHelper.java
@@ -25,10 +25,13 @@ import java.util.Map;
 import java.util.concurrent.TimeUnit;
 import java.util.function.BooleanSupplier;
 import java.util.function.Supplier;
+import java.util.logging.Logger;
+
 import jakarta.json.Json;
 import jakarta.json.JsonObject;
 import jakarta.json.JsonReader;
 
+import org.jboss.as.controller.PathAddress;
 import org.jboss.as.controller.client.ModelControllerClient;
 import org.jboss.as.controller.client.Operation;
 import org.jboss.as.controller.client.OperationMessageHandler;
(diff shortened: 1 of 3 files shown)


bstansberry:
This is temporarily a draft while I see how CI like it on Windows and I use it on top of other branches to see if it reports the uncaught problems we noticed on those branches.

testsuite/scripts/src/test/java/org/wildfly/common/test/ServerHelper.java
```diff
@@ -88,15 +107,38 @@ public static void shutdownStandalone(final ModelControllerClient client) throws
...
+            response = client.execute(Operations.createOperation("read-boot-errors", bootErrorAddress));
+            if (Operations.isSuccessfulOutcome(response)) {
+                final ModelNode errors = Operations.readResult(response);
+                if (errors.isDefined() && !errors.asList().isEmpty()) {
+                    logger.severe("Server has boot errors -- " + response);
+                    return false;
+                }
+            } else {
+                logger.severe("Server read-boot-errors failed -- " + response);
+            }
```
yersan: Out of curiosity, Why we do not return false in this case?

`isStandaloneRunning` will be running in a loop, so it will try again in case an error and if it is never possible to execute "read-boot-errors", test will fail. I'm not sure why we wouldn't want such a behaviour.

testsuite/scripts/src/test/java/org/wildfly/common/test/ServerHelper.java
```diff
@@ -146,17 +188,33 @@ public static boolean isDomainRunning(final ModelControllerClient client, final
...
+                return false;
+            }
+
+            ModelNode result = Operations.readResult(response);
+
+            // If we're admin-only we should be fully started by now, so check we are.
+            // With servers, maybe not, so we don't check
+            if ("ADMIN_ONLY".equals(Operations.readResult(result.get("step-1")).asString())) {
+                ModelNode step2Response = result.get("step-2");
+                if (Operations.isSuccessfulOutcome(step2Response)) {  // todo wouldn't this always be true if the overall op succeeded?
```
yersan, line 190: > // todo wouldn't this always be true if the overall op succeeded?

I think yes, and at this point the operation was successfull, looks quite redundant but harmless so, looks fine to me

yersan (dismissed):
LGTM, only raised one question out of curiosity

testsuite/scripts/src/test/java/org/wildfly/scripts/test/StandaloneScriptTestCase.java
```diff
@@ -114,6 +117,10 @@ void testScript(final ScriptProcess script) throws InterruptedException, Timeout
         Assert.assertNotNull("The process is null and may have failed to start.", script);
         Assert.assertTrue("The process is not running and should be", script.isAlive());
 
+        ModelControllerClient client = TestSuiteEnvironment.getModelControllerClient();
+        ServerHelper.checkBootErrors(client, PathAddress.EMPTY_ADDRESS);
```
yersan, line 121: nit: STANDALONE_CHECK is used before, we are checking with `ServerHelper.checkBootErrors` again here without need

yersan (commented):
`isDomainRunning` variant does not read the boot errors, I don't get the reasoning behind it.

- The two modified test cases explicitly assert on the boot errors by using the ServerHelper.checkBootErrors
- The standalone STANDALONE_CHECK, used by the standalone test case, already read the boot errors, so we will be reading them twice.
- The HOST_CONTROLLER_CHECK doesn't.

Seems we should have removed the `ServerHelper.checkBootErrors` call from both tests and ensure that both checks reads the boot errors, so anyone using the aforementioned checks will assert on boot errors without explicitly invoke `ServerHelper.checkBootErrors`

testsuite/scripts/src/test/java/org/wildfly/common/test/ServerHelper.java
bstansberry: @yersan Thanks! This entire if/else should be removed. I thought I had done that. This is cruft from an initial version before I realized this method is repeatedly tested in a loop and the boot errors and input-args tests should be done separately

testsuite/scripts/src/test/java/org/wildfly/common/test/ServerHelper.java
bstansberry: +1

Since this needs to get done quickly I decided to just not change this in case there was something I wasn't thinking about.

testsuite/scripts/src/test/java/org/wildfly/scripts/test/StandaloneScriptTestCase.java
bstansberry: The STANDALONE_CHECK one shouldn't be testing boot errors -- https://github.com/wildfly/wildfly-core/pull/6886/changes#r3776786786

yersan: ok, that explains all

testsuite/scripts/src/test/java/org/wildfly/common/test/ServerHelper.java
bstansberry: Fixed.

yersan (approved):

PR #6887, opened 2026-08-13
[WFCORE-7453] Expose ordered child types in read-resource-description

rhusar:
Resources now report their ordered child types in the management resource description under the 'ordered-children' key, as a list of child type names. This allows management clients to represent children accurately. Akin to other description keys, it is always present, and is an empty list for resources with no ordered child types. The registration stores the ordered child types sorted by name, so that the description is stable across invocations without having to sort on every call.

Resolves
https://redhat.atlassian.net/browse/WFCORE-7453

Today's discussion on the call @hpehl @bstansberry @yersan reminded me that I had a branch for this. Upon further inspection I have redone it completely – rather than the child exposing an ordered? flag – it should be resposibility of the parent, to describe its ordered children. 

Not a priority, just throwing it out there if people have time while I will be on PTO to think about this.


hpehl:
Thanks @rhusar 
This will definitely make it easier to add ordered resources in the console.

bstansberry:
@rhusar Any update for this will also need an update to https://docs.wildfly.org/41/Admin_Guide#description-of-the-wildfly-managed-resources

PR #6891, opened 2026-08-13
[WFCORE-7670][WFCORE-7671] Update default jdk.serialFilter system property + script test improvements

bstansberry:
Combines:

#6884
#6886

The latter test improvement was created to detect an earlier flaw with the first PR so might as well combine them. They can be merged independently though.


bstansberry:
#6886 has been merged so this adds no value vs #6884

PR #6892, opened 2026-08-13
[WFCORE-7677] The core-feature-pack/galleon-feature-pack module shoul…

bstansberry:
…d always clean before building

https://redhat.atlassian.net/browse/WFCORE-7677


yersan (approved):

rhusar (approved):
Nifty

PR #6896, opened 2026-08-14
[WFCORE-7347] Add public-key-url attribute to token realm jwt configuration

Ohifeme:
issue: https://redhat.atlassian.net/browse/WFCORE-7347

The performRuntime wiring to JwtValidator.Builder.publicKeyUrl() is left as a TODO comment in TokenRealmDefinition. This will be fixed once the companion library change [ELY-3066](https://github.com/wildfly-security/wildfly-elytron/pull/2493) ships in a new wildfly-elytron release.

### DIFF
--- a/elytron/pom.xml
+++ b/elytron/pom.xml
@@ -390,8 +390,9 @@
                                 <exclude>jacc-with-providers.xml</exclude>
                                 <exclude>legacy*.xml</exclude>
                                 <exclude>elytron-subsystem-community*.xml</exclude>
+                                <exclude>elytron-subsystem-19.0.xml</exclude>
                             </excludes>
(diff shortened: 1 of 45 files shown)


yersan (changes requested):
@Ohifeme  for feature requests that requires subsystem bump, we need to have a separated commit for the model bump, and on top of it the changes for the new feature.

Also, this feature is targeting to "default" stability level, hence the process for it is different, the feature itself should be tracked in a different project and you need to form a team with an engineer as a tester and another one as a writer.

Generally, the process is we start upstream delivering the feature in a lower stability level, in such a case the process for it is described at https://docs.google.com/document/d/15_yKhW74-X9s2zUhs_ZUuZ3h-RlMfH5xWmHHsfYf1AA/edit?tab=t.0

PR #6908, opened 2026-08-21
WFCORE-7685 fix for CompositeOperationTestCase intermittent failures

laDok8:
JIRRA: https://redhat.atlassian.net/browse/WFCORE-7685
- Propagate `ts.timeout.factor` to managed domain servers so deployed applications can use TimeoutUtil

### DIFF
--- a/testsuite/domain/src/test/java/org/jboss/as/test/integration/domain/suites/CompositeOperationTestCase.java
+++ b/testsuite/domain/src/test/java/org/jboss/as/test/integration/domain/suites/CompositeOperationTestCase.java
@@ -122,7 +122,7 @@ public class CompositeOperationTestCase {
     }
 
     @Before
-    public void setup() throws IOException {
+    public void setup() throws IOException, MgmtOperationException {
         sysPropVal = 0;
         ModelNode op = Util.createAddOperation(PathAddress.pathAddress(SYS_PROP_ELEMENT));
         op.get(VALUE).set(sysPropVal);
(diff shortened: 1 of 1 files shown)


testsuite/domain/src/test/java/org/jboss/as/test/integration/domain/suites/CompositeOperationTestCase.java
```diff
@@ -112,6 +112,12 @@ public static void setupDomain() throws Exception {
         testSupport = DomainTestSuite.createSupport(CompositeOperationTestCase.class.getSimpleName());
         domainPrimaryLifecycleUtil = testSupport.getDomainPrimaryLifecycleUtil();
         primaryClient = domainPrimaryLifecycleUtil.getDomainClient();
+
+        // Propagate ts.timeout.factor to the server group so managed servers receive it.
+        // This is necessary because ts.timeout.factor is not passed to managed server JVMs.
+        ModelNode op = Util.createAddOperation(SERVER_GROUP_MAIN_SERVER_GROUP.append(SYSTEM_PROPERTY, TimeoutUtil.FACTOR_SYS_PROP));
+        op.get(VALUE).set(Integer.toString(TimeoutUtil.getRawFactor()));
+        DomainTestUtils.executeForResult(op, primaryClient);
```
yersan: This must be a cleanup at `tearDownDomain()` method to avoid any side effects when the domain is shared with other tests.

Or, for consistency, be added in `setup()` method and removed in `tearDown()`. It is a value that never change in the test, so would be ok to leave it under `setupDomain()`, it is just adding on setup seems to be more clear. That's up to you.

yersan (changes requested):

yersan (approved):

PR #6912, opened 2026-09-02
WFCORE-7683 Fixup - Deprecate setStartupTimeoutInSeconds and replace …

rhusar:
…with Duration-based API; i.e. setStartupTimeout(Duration)

https://redhat.atlassian.net/browse/WFCORE-7683

Certainly, this can cleanup the messy DomainLifecycleUtil, but that's for another day.

### DIFF
--- a/testsuite/shared/src/main/java/org/jboss/as/test/integration/domain/management/util/WildFlyManagedConfiguration.java
+++ b/testsuite/shared/src/main/java/org/jboss/as/test/integration/domain/management/util/WildFlyManagedConfiguration.java
@@ -58,7 +58,7 @@ public class WildFlyManagedConfiguration {
 
     private String javaVmArguments = System.getProperty("server.jvm.args", "-Xmx512m");
 
-    private int startupTimeoutInSeconds = (int) TimeoutUtil.adjust(Duration.ofMinutes(2)).toSeconds();
+    private Duration startupTimeout = TimeoutUtil.adjust(Duration.ofMinutes(2));
 
     private boolean outputToConsole = true;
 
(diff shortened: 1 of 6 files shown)


testsuite/shared/src/main/java/org/jboss/as/test/integration/domain/management/util/WildFlyManagedConfiguration.java
```diff
@@ -181,18 +181,35 @@ public WildFlyManagedConfiguration setJavaVmArguments(String javaVmArguments) {
...
+     * @return the maximum time to wait for startup
+     */
+    public Duration getStartupTimeout() {
+        return startupTimeout;
+    }
+
+    /**
+     * @deprecated Use {@link #setStartupTimeout(Duration)} instead.
+     */
+    @Deprecated(forRemoval = true)
```
yersan, line 201: Add a since attribute is always very helpful with the @deprecated, so we know when it was deprecated, e.g "since = 35.0". This is not base code, so it is ok

yersan (approved):

testsuite/shared/src/main/java/org/jboss/as/test/integration/domain/management/util/WildFlyManagedConfiguration.java
rhusar: @yersan Noted, I think you are right – the main use case for it  I see for maintainer is "can i remove this, given how this has been deprecated?" without having to do git blame.

yersan:
Thanks @rhusar

PR #2495, opened 2017-06-02
WFCORE-2426: Use correct restart-required attribute access flag in Elytron resources

yersan:
This PR finishes the work done in JBEAP-10168, setting the correct restart-required attribute flags for some pending Elytron resources.

Jira issues:
https://issues.jboss.org/browse/WFCORE-2426
https://issues.jboss.org/browse/JBEAP-6922

PR #6915, opened 2026-09-07
[WFCORE-7694] Wire jose dependencies into elytron-base

rawojeda:
## Solution
Adds `wildfly-elytron-jose-jwk`, `wildfly-elytron-jose-jwks`, and `wildfly-elytron-jose-util` as direct dependencies of `core-feature-pack-common`, wired into `elytron-base` as `<resources>` artifacts.

Jackson is explicitly kept out of wildfly-core's dependency footprint.
An initial version of this PR wired Jackson (`jackson-core`, `jackson-annotations`, `jackson-databind`) directly into `wildfly-core` as well, but this was flagged during review ([Zulip discussion][zulip]): all Jackson-consuming production code currently lives in `wildfly/wildfly`, which is the authority for the Jackson version used in production.

## Related
* [WFCORE-7694](https://redhat.atlassian.net/browse/WFCORE-7694) (this ticket)
* [ELY-2924](https://redhat.atlassian.net/browse/ELY-2924) => [wildfly-security/wildfly-elytron PR#2495](https://github.com/wildfly-security/wildfly-elytron/pull/2495) (the change that requires this)
* [Zulip](https://wildfly.zulipchat.com/#narrow/channel/173102-wildfly-elytron/topic/JKS.20Refactoring/with/622924964) (Conversation around the PR)

## Testing (with Elytron SNAPSHOT (with PR #2495 changes))
- `mvn enforcer:enforce -pl core-feature-pack/common` =>`BanTransitiveDependencies` passes.
- `mvn clean install -DskipTests`.
- `mvn test -pl testsuite/layers -Dts.layers` => checks a core only provisioned install is unaffected by the optional Jackson references.
- `ls dist/target/wildfly-core-*/modules/system/layers/base/org/wildfly/security/elytron-base/main/ | grep wildfly-elytron-jose` => jars are present.


rawojeda:
Just wanted to flag that currently, wildfly also declares a module.xml for jose-jwk/jose-util(ee-feature-pack/galleon-shared), separate from how wildfly-core bundles them in elytron-base. Does not look as a version divergence as wildfly inherits Elytron from wildfly-core. So its single-sourced but organizationally duplicated (same module content declared in two places).

rawojeda:
@darranl about whether any tests need to move to wildfly. I checked for tests breaking because of the dependency changes and none do.

Note: JwtSecurityRealmTestCase(only WFCore test touching the area) test runs on a flat unit-test classpath where Jackson is present via unrelated test-only deps, so it wouldn't catch a regression if the optional flag on Jackson is removed. Its not a real problem from my perspective but do we want to add a test for that?.

PR #6917, opened 2026-09-09
[WFCORE-7695] Prune unused wildfly-patching code

bstansberry:
https://redhat.atlassian.net/browse/WFCORE-7695


bstansberry:
A test in full WF uses a method in the removed ZipUtil class, so I'll likely restore that method, deprecate the class and file issues to stop using ZipUtil and then remove it.

yersan (approved):

